PR #7792, opened 2021-04-06
fix subquery limit limit bug have signed-off

yangxuanjia:
…ess into yxj_subquery_limit_limit

Signed-off-by: yangxuanjia <yangxuanjia@jd.com>

### DIFF
--- a/go/vt/vtgate/engine/pullout_subquery.go
+++ b/go/vt/vtgate/engine/pullout_subquery.go
@@ -111,7 +111,11 @@ var (
 )
 
 func (ps *PulloutSubquery) execSubquery(vcursor VCursor, bindVars map[string]*querypb.BindVariable) (map[string]*querypb.BindVariable, error) {
-	result, err := ps.Subquery.Execute(vcursor, bindVars, false)
+	subqueryBindVars := make(map[string]*querypb.BindVariable, len(bindVars))
+	for k, v := range bindVars {
+		subqueryBindVars[k] = v
+	}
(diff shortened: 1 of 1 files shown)


systay:
Code looks good. Nice catch.

Would you mind adding a test showing the issue?

yangxuanjia:
> Code looks good. Nice catch.
> 
> Would you mind adding a test showing the issue?

This is my question, which test file should I add this test case?
I see the pullout_subquery_test.go, I can't find any test case fit this bug's scene.

harshit-gangal:
@yangxuanjia If you could write the issue description we can help you with adding test at the right place.

yangxuanjia:
> @yangxuanjia If you could write the issue description we can help you with adding test at the right place.

Lucky day. @harshit-gangal,
This issue is in shard env.
Outer layer select's limit and inner layer subquery's limit,  all use a bindVars["__upper_limit"] to set the number.

So like this sql:
```sql
select customer_id,email from customer where email="aaa11" and customer_id >= (select customer_id from customer where email="aaa11" order by customer_id limit 0,1) limit 0,100;
```
To build bindVars["__upper_limit"], outer layer limit number 100 will be set to this "__upper_limit" key.
But when code run into subquery, limit number 1 will be set to this "__upper_limit" key too. 
So we will see outer layer limit will be replace by subquery limit.

This will cause run the above sql, will return 2 rows, but should be 100.
So I use another copy var to save outer layer bindVars map, when run subquery end, outer layer will see the "__upper_limit" key is 100, not 1.

PR #7793, opened 2021-04-06
Do Not Merge: VReplication: instrumentation and logs for debugging data issues

rohit-nayak-ps:
The following functionality has been added to be able to debug data issues found in a recent resharding workflow. Based off of fb0b16a0f2b7964f5924c17ba3b69678ea936793

### Sequence number
We add a sequence number to each event sent by vstreamer. On the vreplication side we check that the sequence number is monotonically increasing. This ensures that all vstreamer events reach the apply logic. This validates that events are not dropped either in the network or during some buffering.
Any missing event results in the workflow being stopped and the stream's message updated to an error message with the current event.

### GTID validations
On both the vstreamer and vreplication, we check that gtids are monotonically increasing. If not we log these as warnings: we are not entirely sure if gtids will always be monotonically increasing, hence we don't stop the workflow.

### Check RowsAffected 
For each update and insert we verify the RowsAffected by the applied query. During the copy phase we could have inserts and updates for PKs which have not already been copied and hence don't need to be applied. For these RowsAffected should be 0, for others non-zero. We log warnings if the validations failed.

Additionally info logs have been added as part of the apply process. For ease of grepping the logs each log added in this PR contains "dbg:". Errors are logged as edbg:, warnings as wdbg: and info as idbg:

Also includes changes to the rowlog tool to handle sharded keyspaces .

_Note_
GTID distance measurement is currently limited to the MySQL56 flavor and uses a naive algorithm that works for this specific use case.

## Related Issue(s)

- 

## Checklist

## Impacted Areas in Vitess

- [ ]  Query Serving
- [ ]  VReplication
- [ ]  Cluster Management
- [ ]  Build/CI
- [ ]  VTAdmin


rohit-nayak-ps:
We will not be merging this into main. We can use the approach in this PR in case we need to debug any issues in the future.

PR #7803, opened 2021-04-07
Use actions/setup-go@v2 and go 1.15.11 or newer

dctrwatson:

## Description
actions/setup-go v2 contains fixes for version matching and logging of go env

Also, rather than relying on whatever latest Go version is cached on the runner, explicitly ask for Go 1.15.11 or newer patch release

## Related Issue(s)

- 

## Checklist

## Impacted Areas in Vitess

- [ ]  Query Serving
- [ ]  VReplication
- [ ]  Cluster Management
- [x]  Build/CI
- [ ]  VTAdmin

### DIFF
--- a/.github/workflows/check_formatting.yml
+++ b/.github/workflows/check_formatting.yml
@@ -8,9 +8,9 @@ jobs:
     steps:
 
     - name: Set up Go
-      uses: actions/setup-go@v1
+      uses: actions/setup-go@v2
       with:
-        go-version: 1.15
+        go-version: ~1.15.11
 
     - name: Check out code
       uses: actions/checkout@v2
--- a/.github/workflows/check_make_parser.yml
+++ b/.github/workflows/check_make_parser.yml
@@ -8,9 +8,9 @@ jobs:
     steps:
 
     - name: Set up Go
-      uses: actions/setup-go@v1
+      uses: actions/setup-go@v2
       with:
-        go-version: 1.15
+        go-version: ~1.15.11
 
     - name: Check out code
       uses: actions/checkout@v2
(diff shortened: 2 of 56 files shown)


deepthi:
Questions for reviewers: Is this a good idea? Do you see any downsides to requiring a specific minor version in the build?
This is what we would use for the next release build (11.0 since 10.0 has branched).

dctrwatson:
> Do you see any downsides to requiring a specific minor version in the build?

This just sets the minimum patch version.

For example, currently the Github runner image only has go1.15.10, so this forces the usage of go1.15.11 now.

If go1.15.12 is released, it would automatically upgrade as soon as https://github.com/actions/go-versions/blob/main/versions-manifest.json is updated or if a new runner image is released with it pre-isntalled.

dkhenry (approved):
I am approving assuming there isn't a known issue forcing the 1.15.11+ minor version. 

If there is something causing you to specifically want 1.15.11+ then you should also update the main Docker images to require that as well. As it stands now they only require 1.15

dctrwatson:
> then you should also update the main Docker images to require that as well. As it stands now they only require 1.15

oh interesting, the docker images are pinned to whatever the 1.15.x release was at the time the bootstrap images are built.

So this really doesn't change that mismatch.

> If there is something causing you to specifically want 1.15.11+

there is not

dkhenry:
It should be fine then, to my knowledge there isn't any incompatibilities in 1.15 that would cause CI to pass, but cause failures running in the docker images.

deepthi:
I was looking at some failed tests and it looks like we are getting 1.15.11 already:
https://github.com/vitessio/vitess/runs/2300484018?check_suite_focus=true

dctrwatson:
> I was looking at some failed tests and it looks like we are getting 1.15.11 already:

yay, github finally released a new runner image

PR #7805, opened 2021-04-08
Update `GetSchema` filtering to exclude shards where `IsMasterServing` but no `MasterAlias`

ajm188:

## Description

This effectively accounts for corner cases where there is an empty shard which happens to also be marked as serving.

## Related Issue(s)

- 

## Checklist

## Impacted Areas in Vitess

- [ ]  Query Serving
- [ ]  VReplication
- [ ]  Cluster Management
- [ ]  Build/CI
- [x]  VTAdmin

### DIFF
--- a/go/vt/vtadmin/cluster/cluster.go
+++ b/go/vt/vtadmin/cluster/cluster.go
@@ -633,7 +633,7 @@ func (c *Cluster) getTabletsToQueryForSchemas(ctx context.Context, keyspace stri
 		tabletsToQuery := make([]*vtadminpb.Tablet, 0, len(shards))
 
 		for _, shard := range shards {
-			if !shard.Shard.IsMasterServing {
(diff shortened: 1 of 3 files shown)


go/vt/vtadmin/cluster/cluster.go
```diff
@@ -633,7 +633,7 @@ func (c *Cluster) getTabletsToQueryForSchemas(ctx context.Context, keyspace stri
 		tabletsToQuery := make([]*vtadminpb.Tablet, 0, len(shards))
 
 		for _, shard := range shards {
-			if !shard.Shard.IsMasterServing {
+			if !(shard.Shard.IsMasterServing && shard.Shard.MasterAlias != nil) {
```
rafael, line 636: Could you add comment for why we need to do this check? Our future selves will be thankful for that :D

rafael (approved):
LGTM

rafael:
Looks good, let's merge when the tests pass.

Issue #7804, opened 2021-04-07
Small typo fix request


hallaroo:
We got a report of a typo [here](https://github.com/vitessio/vitess/blob/master/go/vt/vttablet/tabletmanager/vreplication/stats.go#L291) in our stats.go doc where 'the tags should be connected by “dot”, otherwise seems prometheus will drop the metrics'.

Could we please get this fixed? Thanks!

hallaroo:
Added the confirmation of a fix to the meeting notes with the reporter. Closing this out. Thank you so much!!!!!

PR #7807, opened 2021-04-08
VReplicationErrors metric: use . as delimiter instead of _ to behave well with Prometheus

rohit-nayak-ps:
## Description

Prometheus requires the . (period) separator between dimensions and was dropping the VReplicationErrors metrics due to the use of _ as a delimiter at one place.

## Related Issue(s)
- #7804 

## Checklist

## Impacted Areas in Vitess

- [ ]  Query Serving
- [X]  VReplication
- [ ]  Cluster Management
- [ ]  Build/CI
- [ ]  VTAdmin


hallaroo (approved):
Looks perfect. Thank you!!

PR #7811, opened 2021-04-08
Bump MySQL version for docker builds to 8.0.23

aquarapid:
since the xtrabackup issue has been resolved.

* Remove comment implying we do not support 8.0

### DIFF
--- a/docker/lite/install_dependencies.sh
+++ b/docker/lite/install_dependencies.sh
@@ -92,7 +92,7 @@ mysql57)
     )
     ;;
 mysql80)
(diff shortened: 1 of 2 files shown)


aquarapid:
Resolves #7812

dkhenry (approved):
I notice on this we don't pin the version of xtra backup. Not going to block the PR on it, but since this is only fixed in newer versions of xtra backup it may be good to pin the version to a known good one

deepthi (approved):
LGTM.
Test failures look unrelated.

PR #5930, opened 2020-03-17
Pass in cell to EnsureVSchema and GetOrCreateShard

zmagg:
### Description
Modifies `EnsureVSchema` so that we always pass in a cell and it rebuilds SrvVSchema only in the provided cell. This removes its auto-rebuild of all cells automatically when you're creating a new shard or keyspace.

### Background
During a setup of a new region, we had a misconfigured topology and spinning up a new shard caused us to corrupt **all cells in the topology**, just as part of booting up a new tablet. For safety reasons, it seems better to decouple spinning up a new shard from rebuilding SrvVSchema on all cells. As talked about in person a few weeks ago--we'll keep the rebuild for the simple case where there's only one cell.

Signed-off-by: Maggie Zhou <mzhou@slack-corp.com>

PR #7815, opened 2021-04-10
tm init: if srvKeyspace exists but is empty, rebuild it

### LINKED ISSUE #7855: Upgrading from Vitess 9 to 10 needs RebuildKeyspaceGraph
When upgrading Vitess in staging a cluster ran into an issue where it's shards disappeared. We got a "no shard in keyspace" error, but all the shards are showing as serving and the topo looks good.

```
mysql> show vitess_keyspaces; 
+---------------+
| Database      |
+---------------+
| testshrinking |
+---------------+
1 row in set (0.07 sec)
mysql> show vitess_tablets;  
+--------------+---------------+-------+------------+---------+-------------------------+--------------+----------------------+
| Cell         | Keyspace      | Shard | TabletType | State   | Alias                   | Hostname     | MasterTermStartTime  |
+--------------+---------------+-------+------------+---------+-------------------------+--------------+----------------------+
| awsuseast1a1 | testshrinking | -80   | MASTER     | SERVING | XXXX| XXXX   | 2021-04-07T22:49:56Z |
| awsuseast1a1 | testshrinking | 80-   | MASTER     | SERVING | XXXX| XXXX  | 2021-04-07T22:47:39Z |
| awsuseast1c1 | testshrinking | -80   | REPLICA    | SERVING | XXXX | XXXX  |                      |
| awsuseast1c1 | testshrinking | 80-   | REPLICA    | SERVING | XXXX | XXXX |                      |
+--------------+---------------+-------+------------+---------+-------------------------+--------------+----------------------+
4 rows in set (0.07 sec)
```

But:
```
mysql> show vitess_shards;
I0407 23:36:39.940341       1 vtgate.go:466] Execute: Code: UNAVAILABLE
no shard in keyspace
, request: map[BindVariables:map[] Session:autocommit:true target_string:"testshrinking" options:<included_fields:ALL workload:OLTP > found_rows:4 row_count:-1 DDLStrategy:"direct" SessionUUID:"XXXX" Sql:show tables]
```

And we found that a RebuildKeyspaceGraph fixed it.

deepthi:
## Description
Sometimes clusters get into a situation where vtgate is unable to serve and only a `RebuildKeyspaceGraph` will fix it.
This PR detects the condition on tablet startup and rebuilds the SrvKeyspace if necessary.

## Related Issue(s)
Fixes #7855 

## Checklist
- [x] Should this PR be backported? (Only for 10.0)

## Impacted Areas in Vitess

- [ ]  Query Serving
- [ ]  VReplication
- [x]  Cluster Management
- [ ]  Build/CI
- [ ]  VTAdmin

### DIFF
--- a/go/vt/vttablet/tabletmanager/tm_init.go
+++ b/go/vt/vttablet/tabletmanager/tm_init.go
@@ -41,6 +41,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/golang/protobuf/proto"
+
 	"vitess.io/vitess/go/flagutil"
 	"vitess.io/vitess/go/sync2"
 	"vitess.io/vitess/go/vt/vterrors"
@@ -384,8 +386,17 @@ func (tm *TabletManager) createKeyspaceShard(ctx context.Context) (*topo.ShardIn
 	srvKeyspace, err := tm.TopoServer.GetSrvKeyspace(ctx, tm.tabletAlias.Cell, tablet.Keyspace)
 	switch {
 	case err == nil:
-		tm.tmState.RefreshFromTopoInfo(ctx, nil, srvKeyspace)
+		if proto.Equal(srvKeyspace, &topodatapb.SrvKeyspace{}) {
+			// If srvKeyspace exists but is empty we need to rebuild it.
+			// RebuildKeyspace will fail until at least one tablet is up for every shard.
+			// So we call this in a goroutine and retry until it succeeds.
+			go tm.rebuildKeyspace(tablet.Keyspace, rebuildKeyspaceRetryInterval)
+		} else {
+			tm.tmState.RefreshFromTopoInfo(ctx, nil, srvKeyspace)
+		}
 	case topo.IsErrType(err, topo.NoNode):
+		// RebuildKeyspace will fail until at least one tablet is up for every shard.
+		// So we call this in a goroutine and retry until it succeeds.
 		go tm.rebuildKeyspace(tablet.Keyspace, rebuildKeyspaceRetryInterval)
 	default:
 		return nil, vterrors.Wrap(err, "initeKeyspaceShardTopo: failed to read SrvKeyspace")
@@ -403,6 +414,7 @@ func (tm *TabletManager) createKeyspaceShard(ctx context.Context) (*topo.ShardIn
 		}
 	case topo.IsErrType(err, topo.NoNode):
 		// There is no SrvSchema in this cell at all, so we definitely need to rebuild.
+		// No need to call this in a goroutine because it will succeed as soon as the keyspace is present.
 		if err := tm.TopoServer.RebuildSrvVSchema(ctx, []string{tm.tabletAlias.Cell}); err != nil {
 			return nil, vterrors.Wrap(err, "initeKeyspaceShardTopo: failed to RebuildSrvVSchema")
 		}
--- a/go/vt/vttablet/tabletmanager/tm_init_test.go
+++ b/go/vt/vttablet/tabletmanager/tm_init_test.go
@@ -20,6 +20,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/golang/protobuf/proto"
+
 	"context"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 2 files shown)


go/vt/vttablet/tabletmanager/tm_init.go
```diff
@@ -384,8 +386,17 @@ func (tm *TabletManager) createKeyspaceShard(ctx context.Context) (*topo.ShardIn
 	srvKeyspace, err := tm.TopoServer.GetSrvKeyspace(ctx, tm.tabletAlias.Cell, tablet.Keyspace)
 	switch {
 	case err == nil:
-		tm.tmState.RefreshFromTopoInfo(ctx, nil, srvKeyspace)
+		if proto.Equal(srvKeyspace, &topodatapb.SrvKeyspace{}) {
```
sougou, line 389: Better to check if the current keyspace exists. The SrvKeyspace could be non-empty, but not have the current keyspace.

deepthi: We are already checking that (line 386). There is a file per-keyspace and we will get a topo.NoNode error if that file doesn't exist.

sougou: This is a different situation: let's say SrvKeyspace has been rebuilt across all cells for ks1. Now, if we create ks2, it will already exist, and this code path will get skipped. But we actually need to rebuild to get the ks2 part of the SrvKeyspace.

sougou: nvm. SrvKeyspace is the keyspace we're looking for.

sougou (approved):

go/vt/vttablet/tabletmanager/tm_init.go
deepthi: To clarify for anyone else following along:
- SrvKeyspace is stored at .../keyspaces/\<keyspace\>/SrvKeyspace
- We already detect if that file doesn't exist and call Rebuild
- Sometimes (for unknown reasons), the file exists but is empty, which leads to vtgate throwing a `no shard in keyspace` error.
- This PR is intended to detect that situation.

rafael: At Slack, we once ran into a very bad incident due to this current logic of rebuilding the keyspace automatically. Since then, I'm nervous of this approach. I would advocate against adding more of those in the codebase. 

In my current view of the world, this kind of errors should fold into the category where an operator intervention is expected. If that unknown reason manifest, the operator should manually call RebuildKeyspaceGraph to fix the issue. 

I'm going to request changes to this PR since Sugu approved, so we can talk more about this :D

rafael (changes requested):

rafael:
Looking at #7855, I think we should look deeper into what changed that is causing this problem. We have never seen this issue before, so there must be a regression. I think this fix would be a workaround for the underlying regression.

deepthi:
There are 3 cells 1a, 1b and 1c. Each shard has only 2 tablets and they are in 1a and 1c. There are two vtgates, one each in 1a and 1c. The vtgate in 1c throws a `no shards in keyspace` error, which happens if the SrvKeyspace file exists but is empty. 1c has only replica tablets, no master tablets.
The vtgate in 1a does not throw the same error.
@aquarapid is this accurate?

deepthi:
> At Slack, we once ran into a very bad incident due to this current logic of rebuilding the keyspace automatically. Since then, I'm nervous of this approach. I would advocate against adding more of those in the codebase.

Was that `RebuildKeyspaceGraph` or `RebuildVSchemaGraph`? (Not that it really matters, the concern is the same).
Note that if the SrvKeyspace or SrvVSchema doesn't exist for the keyspace, then it _will_ be rebuilt already by existing code. The only assurance is that the rebuild is done only by the very first tablet started in that cell/keyspace combination and it affects only the cell the tablet belongs to. Other cells are not affected (for better or worse).

rafael:
In our case was with: `RebuildKeyspaceGraph`. Like you mention I think the concern is the same.

deepthi:
> In our case was with: `RebuildKeyspaceGraph`. Like you mention I think the concern is the same.

Can you share more details about what happened?

rafael:
I actually, I think our problem was more with `EnsureVSchema`, this PR has some context: https://github.com/vitessio/vitess/pull/5930/files

deepthi:
I dug a little deeper after discussing with @rafael.
It turns out that an empty SrvKeyspace produces a different error from what is reported in #7855 
When SrvKeyspace is empty (and you can get there by calling `CreateKeyspace` followed by `RebuildKeyspaceGraph` with no shards or tablets), the error message produced is 
```
ERROR 1105 (HY000): No partition found for tabletType master in keyspace customer
```
In order to get the error reported in #7855, SrvKeyspace has to exist, but the desired partition (for a specific tablet_type) has no shard_references. Right now we don't know how we can get into that situation.
This PR does not detect that condition so it is unlikely to really fix #7855 
We should probably try to reproduce #7855 and find the root cause before attempting to fix it.

PR #7817, opened 2021-04-10
[vtadmin-web] Add DataCell component

doeg:

## Description

A couple of straightforward styling changes that I keep re-implementing in my bigger branches. 😎  Figured I should just pull them out into their own thing:

- Adds a DataCell component for consistent styling. (A separate component will make more sense when we do rowspan grouping... I admit right now it's kinda overkill, haha. But! Still convenient.) 

- Renders cluster as secondary text where it "makes sense", which is (for now) when the cardinality of clusters is << the cardinality of whatever the table is rendering. Every cluster has many tablets, for example, so a separate "cluster" column is quite redundant when every row on a page is for the same cluster. Some tables, though, only have one or two columns right now, and in those cases I left the "cluster" column as a space filler. :) 

Before:
Screen Shot 2021-04-10 at 10 50 55 AM

After: 
Screen Shot 2021-04-10 at 11 19 52 AM

## Related Issue(s)

N/A

## Checklist
- [ ] Should this PR be backported? No

## Deployment Notes

N/A

## Impacted Areas in Vitess

- [ ]  Query Serving
- [ ]  VReplication
- [ ]  Cluster Management
- [ ]  Build/CI
- [x]  VTAdmin

### DIFF
--- a/web/vtadmin/src/components/dataTable/DataCell.tsx
+++ b/web/vtadmin/src/components/dataTable/DataCell.tsx
@@ -1,0 +1,28 @@
+/**
+ * Copyright 2021 The Vitess Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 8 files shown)


ajm188 (approved):

doeg:
I had to sneak in a small change. 😊

ajm188 (approved):

PR #7828, opened 2021-04-12
perf: optimize bind var generation

vmg:
## Description

This is 3 optimizations in one:

1. `sqlparser: do not split statements that don't contain a semicolon`: the insight is simple: the MySQL server implementation tries to split all incoming query packets into their individual statements, but the vast majority of queries do not contain more than one statement. If a query doesn't contain a `;`, we don't bother to do a full tokenization phase to split it.

2. Optimize reserved variable generation: the normalization/rewriting code uses a map to keep track of the bind variable names that are already in use, as to ensure we don't insert duplicated variables. This becomes very slow in queries with many binds, such as bulk inserts. This optimization implements a new `ReservedVars` type that can detect duplicate variables and generate unique variable names more efficiently.
    - The var generator has a "fast" mode: when creating a new `ReservedVars` instance, we check whether any of the existing variable names has the same prefix as the Vitess-specific variables we're going to generate. If that's not the case, we can generate unique variables incrementally, ensuring that they'll never collide. We don't need to check the existing variables map when generating.
    - Furthermore, the most common use case for query normalization uses the `"vtg"` prefix for all variable names (this is the normalization that is performed for all incoming queries in VTGate), so we don't need to allocate new variable names dynamically. This PR introduces a static array of variable names that can be reused between all the incoming queries to a VTGate, drastically reducing memory allocations, particularly in bulk insert queries.

3. Remove memory allocations in `Argument` and `ListArg`: this is the larger refactoring in this PR. These two SQL AST nodes hold an invariant: that the name of the argument or list argument must always begin with `:` or `::`, respectively. However, the vast majority of places in Vitess where we're using bind variables as arguments, they're **not** prefixed with a colon. This leads to a pattern that repeats dozens of times in the codebase, where we allocate `Argument` instances by creating superfluous string allocations, e.g. `NewArgument(":" + bindVar)`. If we simply store the name of the argument _without_ the colon, we never have to prepend colons -- an operation that happens hundreds of times during a normal `vtgate` request. **This refactoring reduces the individual memory allocations by 5%**, and this is a _global_ reduction. Very significant.

A synthetic benchmark also shows very impressive results. Note that this synthetic benchmark includes parsing (which is required before normalization), and roughly 50% of the runtime of the benchmark is spent on _parsing_, while this is a perf improvement for _normalization_. The point improvement, if it could be measured in isolation, would be closer to `40%`.

```
name                    old time/op    new time/op    delta
NormalizeTPCCInsert-16     255ms ± 1%     203ms ± 1%  -20.29%  (p=0.000 n=9+10)

name                    old alloc/op   new alloc/op   delta
NormalizeTPCCInsert-16    68.5MB ± 0%    61.7MB ± 0%   -9.91%  (p=0.000 n=10+10)

name                    old allocs/op  new allocs/op  delta
NormalizeTPCCInsert-16     1.31M ± 0%     1.06M ± 0%  -19.21%  (p=0.000 n=10+10)
``` 

## Related Issue(s)

- 

## Checklist

## Impacted Areas in Vitess

- [x]  Query Serving
- [x]  VReplication
- [ ]  Cluster Management
- [ ]  Build/CI
- [ ]  VTAdmin

### DIFF
--- a/go/vt/sqlparser/ast_rewriting.go
+++ b/go/vt/sqlparser/ast_rewriting.go
@@ -17,6 +17,8 @@ limitations under the License.
 package sqlparser
 
 import (
+	"strconv"
+
 	querypb "vitess.io/vitess/go/vt/proto/query"
 	vtrpcpb "vitess.io/vitess/go/vt/proto/vtrpc"
 	"vitess.io/vitess/go/vt/vterrors"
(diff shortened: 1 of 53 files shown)


go/vt/sqlparser/ast_rewriting.go
```diff
@@ -32,10 +34,109 @@ type RewriteASTResult struct {
 	AST Statement // The rewritten AST
 }
 
+type ReservedVars struct {
```
systay, line 37: comments for the public members in this file would be ⭐

vmg: Done!

harshit-gangal (approved):

PR #7840, opened 2021-04-13
Increase the timeout for StartActionLoop

harshit-gangal:

## Description

Recently this function was added with vreplication test engine and may be due to that it would be taken more time to come up. Increasing the timeout to see if that makes it less flaky.

## Related Issue(s)

- 

## Impacted Areas in Vitess

- [X]  Build/CI


rohit-nayak-ps (approved):
This change is worth trying . Though I wonder if the issue is localized to this: I often see networking/connection errors in other tests that are run before it reaches this point.

You mentioned in the PR comment that this has been added recently, but it seems to have been around since 2015?

harshit-gangal:
I think this does not seem to work.

The recent change on this function was around adding `binlogplayer.NewFakeDBClient` to `vreplication.NewTestEngine`

harshit-gangal:
As discussed, this does not seem to be a problem. Closing this

Issue #7859, opened 2021-04-14
Feature Request: log unsupported queries


systay:
In order to get a sense of which Vitess limitations are being hit upon in the query path, it would be nice to log these on vtgate and vttablet

harshit-gangal:
I think logging at vtgate would be sufficient as first there will be only few queries if they reach vttablet will throw error as Vitess limitations, second even if there are some will get caught at VTGate as limitations only.

deepthi:
Do we need to worry about redacting bind variable values?

PR #7865, opened 2021-04-14
log unsupported queries

harshit-gangal:

## Related Issue(s)

- resolves https://github.com/vitessio/vitess/issues/7859

## Checklist

## Impacted Areas in Vitess

- [X]  Query Serving


deepthi:
Do we need to worry about redacting bind variable values?

harshit-gangal:
> Do we need to worry about redacting bind variable values?

Yes, I was quite surprised that we were not doing it already. I did not change it before taking with @systay

systay (approved):

PR #7868, opened 2021-04-14
CI: run some tests inside docker

deepthi:
## Description
Attempt to work-around GitHub Actions issues by running tests inside docker.
* Unit tests can be run with the correct docker image because we publish bootstrap images for various flavors
* There is a mismatch between flavors and docker images. The percona bootstrap image has now been tagged with percona56 and the mariadb image with mariadb101 to allow us to still run tests by flavor.
* mariadb102 test has been temporarily removed. If docker becomes a permanent solution, we will need to build a new bootstrap image for this
* Other failing tests (onlineddl, examples) etc. will run using the mysql57 image
* Test run time goes up when running in docker. For example, unit tests take 9-10 minutes without docker, and 10-12 minutes with.
* Increase timeout inside the test runner to 30 minutes. unit_test_race takes longer than 10 minutes.
* By default the test runner runs each test 3 times, turn that off by setting retry=1
* onlineddl_ghost and onlineddl_declarative will be overwritten if we run `make generate_ci_workflows` again, so we need to watch out for that.

## Impacted Areas in Vitess

- [ ]  Query Serving
- [ ]  VReplication
- [ ]  Cluster Management
- [x]  Build/CI
- [ ]  VTAdmin


rohit-nayak-ps (approved):
lgtm
brilliant idea :-)

PR #7873, opened 2021-04-15
Fix + integration test for `keyspaces_to_watch` routing regression [fixes #7882]

### LINKED ISSUE #7882: [vtgate] Fails to route any query when keyspace filtering is enabled
#### Overview of the Issue

When we started working on online DDL support as a first-party action we introduced a dependency in the vtgate to be able to modify the topology. This conflicts with limitations enforced by a KeyspaceFilteringServer which has access to only a subset of the topo.

As a protective measure when attempting to gain access to the Write-enabled topo.Server a filtering server throws an error. Since this is now happening as part of the vcursor setup to support schema changes all vtgate routing was failing if `-keyspaces_to_watch` was passed to the vtgate.

#### Reproduction Steps

The summary:
1. start vtgate with `-keyspaces_to_watch commerce`
2. attempt to perform a query

I ended up validating my change by modifying `examples/local/scripts/vttablet-up.sh` to pass the keyspace filtering flag and then using the sample `examples/local/101_initial_cluster.sh` to bootstrap everything.

When making a query you get a result back like:

```
[~/vitess/examples/local] (vtgate_keyspaces_to_watch)
sa@docker $ mysql -S /tmp/mysql.sock
Welcome to the MariaDB monitor.  Commands end with ; or \g.
Your MySQL connection id is 3
Server version: 5.7.9-vitess-11.0.0-SNAPSHOT

Copyright (c) 2000, 2018, Oracle, MariaDB Corporation Ab and others.

Type 'help;' or '\h' for help. Type '\c' to clear the current input statement.

MySQL [(none)]> select 1 from dual;
ERROR 1105 (HY000): cannot access underlying topology server when keyspace filtering is enabled
```

#### Binary version

This is present in all releases >= 8.0, I don't recall off hand which PR actually introduced it.

setassociative:
## Description

When we started working on online DDL support as a first-party action we introduced a dependency in the vtgate to be able to modify the topology. This conflicts with limitations enforced by a [KeyspaceFilteringServer](https://github.com/vitessio/vitess/blob/master/go/vt/srvtopo/keyspace_filtering_server.go#L39-L45) which has a access to only a subset of the topo.

As a protective measure when attempting to gain access to the Write-enabled topo.Server a filtering server throws an error. Since this happens as part of the vcursor setup to support schema changes all vtgate routing was failing if `-keyspaces_to_watch` was passed to the vtgate.

I fixed this, added a new endtoend integration test to prevent future regressions, and updated how we deal with the flag value to more closely match an approach @ajm188 and I have schemed up to get flags into a more manageable state.

### Unit vs integration
This is done as an integration test because adding it as a unit test relies on a fixed assumption around how keyspaces_to_watch / keyspace filtering works which I didn't want to introduce into the vcursor test suite. It also is somewhat fragile as changing the type of topo.Server used in the test would make the test invalid.

In the end all we really care about is that vtgates can route queries and so it seemed most reasonable to just make it an integration test.

## Related Issue(s)
n/a

## Checklist
Nominally this should be backported to 8.0-10.0 but I believe most users aren't in a position to need access to this flag so it's buyer's choice here.

## Deployment Notes
This is a non-breaking change for folks who were successfully using vtgates >= 8.0.

## Impacted Areas in Vitess

- [x]  Query Serving
- [ ]  VReplication
- [ ]  Cluster Management
- [ ]  Build/CI
- [ ]  VTAdmin

### DIFF
--- a/go/vt/vtgate/vcursor_impl.go
+++ b/go/vt/vtgate/vcursor_impl.go
@@ -42,6 +42,7 @@ import (
 	"context"
 
 	"vitess.io/vitess/go/sqltypes"
+	"vitess.io/vitess/go/vt/discovery"
 	"vitess.io/vitess/go/vt/key"
 	"vitess.io/vitess/go/vt/sqlparser"
 	"vitess.io/vitess/go/vt/srvtopo"
@@ -163,7 +164,8 @@ func newVCursorImpl(
 	vschema *vindexes.VSchema,
 	resolver *srvtopo.Resolver,
 	serv srvtopo.Server,
-	warnShardedOnly bool) (*vcursorImpl, error) {
+	warnShardedOnly bool,
+) (*vcursorImpl, error) {
 	keyspace, tabletType, destination, err := parseDestinationTarget(safeSession.TargetString, vschema)
 	if err != nil {
 		return nil, err
@@ -174,7 +176,9 @@ func newVCursorImpl(
 		return nil, vterrors.Errorf(vtrpcpb.Code_UNIMPLEMENTED, "transaction is supported only for master tablet type, current type: %v", tabletType)
 	}
 	var ts *topo.Server
-	if serv != nil {
+	// We don't have access to the underlying TopoServer if this vtgate is
+	// filtering keyspaces because we don't have an accurate view of the topo.
+	if serv != nil && !discovery.FilteringKeyspaces() {
 		ts, err = serv.GetTopoServer()
 		if err != nil {
 			return nil, err
@@ -588,6 +592,9 @@ func (vc *vcursorImpl) TabletType() topodatapb.TabletType {
 
 // SubmitOnlineDDL implements the VCursor interface
 func (vc *vcursorImpl) SubmitOnlineDDL(onlineDDl *schema.OnlineDDL) error {
+	if vc.topoServer == nil {
+		return vterrors.New(vtrpcpb.Code_INTERNAL, "Unable to apply DDL toposerver unavailable, ensure this vtgate is not using filtered keyspaces")
+	}
 	conn, err := vc.topoServer.ConnForCell(vc.ctx, topo.GlobalCell)
 	if err != nil {
 		return err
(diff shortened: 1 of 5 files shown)


deepthi:
We should get this into 10.0 branch so that it can go into the release.
It _is_ a regression. @setassociative have others not seen it in 8.0/9.0 only because they don't use `keyspaces_to_watch`?

setassociative:
> @setassociative have others not seen it in 8.0/9.0 only because they don't use `keyspaces_to_watch`?

Yea, I think Slack may be the only org utilizing that flag bc of how we deploy gates/tablets.

> Can you create an issue that documents the error message you were getting? It will be useful for future reference.

Sure thing, will do tomorrow.

setassociative:
Filed https://github.com/vitessio/vitess/issues/7882

go/vt/vtgate/vcursor_impl.go
```diff
@@ -588,6 +592,9 @@ func (vc *vcursorImpl) TabletType() topodatapb.TabletType {
 
 // SubmitOnlineDDL implements the VCursor interface
 func (vc *vcursorImpl) SubmitOnlineDDL(onlineDDl *schema.OnlineDDL) error {
+	if vc.topoServer == nil {
+		return vterrors.New(vtrpcpb.Code_INTERNAL, "Unable to apply DDL toposerver unavailable, ensure this vtgate is not using filtered keyspaces")
```
deepthi, line 596: nit: phrasing.
"Unable to apply DDL because toposerver is read-only..." might be better?

deepthi:
Note: this means that online DDL cannot be used along with keyspace filtering.
However, we have started work on migrating away from the topo dependency. Once that is complete, it should be possible to use online DDL even in the presence of keyspace filtering.

shlomi-noach:
> we have started work on migrating away from the topo dependency. Once that is complete, it should be possible to use online DDL even in the presence of keyspace filtering.

Reference: https://github.com/vitessio/vitess/pull/7785, https://github.com/vitessio/vitess/issues/6926#issuecomment-817619302

deepthi:
I have merged this as-is so that we can get it into 10.0. The error message can be fixed in another PR.

PR #7879, opened 2021-04-16
Fix bug with reserved connections to stale tablets

### LINKED ISSUE #7875: vtgate `enable_system_settings` reserved connections do not respect re-mastering
Scenario:
  * Tablets running against external (remote) MySQL instances.
  * Vitess is not managing the replication/failover for the MySQL instances (e.g. Aurora/RDS, etc.)
  * Multiple vttablets are pointed to the same external MySQL instance (e.g. the master);  one running as MASTER, one as SPARE.  This allows quick failover if the host running the MASTER vttablet dies.
  * In such a scenario, a `TabletExternallyReparented` is used to promote the SPARE tablet to MASTER (and demote the MASTER).

However, the reserved connections used by the vtgate `enable_system_settings` feature are not aware of such a failover, and subsequent requests from the client application to existing (reserved) connections will fail (and continue failing) with an error like:

```
ERROR 1105 (HY000): vtgate: http://localhost:15001/: vttablet: rpc error: code = FailedPrecondition desc = operation not allowed in state NOT_SERVING
```

The client will be "stuck" and keep receiving errors, unless it disconnects and reconnects.

Repro:

 * Use Vitess version:  `Version: daa6085 (Git branch 'HEAD') built on Tue Jan 26 09:37:50 UTC`  (i.e. 9.0;  but I believe this will behave the same on current master)
 * Start 2 tablets (with `-enable_system_settings true`), both pointing to the same external MySQL tablet
 * Use `vtctlclient ChangeTabletType` (if necessary) to make the one a `SPARE` tablet.
 * This is the state of the tablets:
  ```
zone1-0000000400 keyspace3 0 master 192.168.0.139:15400 192.168.0.139:17300 [] 2021-04-15T13:37:56Z
zone1-0000000402 keyspace3 0 spare 192.168.0.139:15402 192.168.0.139:17300 [] <null>
  ```
  * Connect with a MySQL client to keyspace via vtgate, then run a `SET` command that will reserve the connection (`set sql_mode=''` in this case):
  ```
mysql> use keyspace3
Database changed
mysql> set sql_mode='';

mysql> select * from t2;
+----+------+
| c1 | c2   |
+----+------+
|  1 | NULL |
+----+------+
1 row in set (0.00 sec)
  ```
  * Everything is still fine.  Leave this session running.  Now do a TER to the spare tablet:
  ```
$ vtctlclient -server 127.0.0.1:15999 TabletExternallyReparented zone1-0000000402
$ vtctlclient -server 127.0.0.1:15999 ListAllTablets     
zone1-0000000400 keyspace3 0 replica 192.168.0.139:15400 192.168.0.139:17300 [] <null>
zone1-0000000402 keyspace3 0 master 192.168.0.139:15402 192.168.0.139:17300 [] 2021-04-15T13:41:44Z
  ```
  * Depending on the exact configuration the original master (`zone1-0000000400`) might become a `REPLICA` or a `SPARE`;  but either way; it will be non-serving, as can be seen from `show vitess_tablets;`:
  ```
mysql> show vitess_tablets;
+-------+--------------+-------+------------+-------------+------------------+---------------+----------------------+
| Cell  | Keyspace     | Shard | TabletType | State       | Alias            | Hostname      | MasterTermStartTime  |
+-------+--------------+-------+------------+-------------+------------------+---------------+----------------------+
| zone1 | keyspace3    | 0     | MASTER     | SERVING     | zone1-0000000402 | 192.168.0.139 | 2021-04-15T13:41:44Z |           |
| zone1 | keyspace3    | 0     | REPLICA    | NOT_SERVING | zone1-0000000400 | 192.168.0.139 |                      |                  |
+-------+--------------+-------+------------+-------------+------------------+---------------+----------------------+
  ```
  * Now, in the same mysql session as you used to issue the `select * from t2;` above, try it again:
  ```
mysql> select * from t2;
ERROR 1105 (HY000): vtgate: http://localhost:15001/: vttablet: rpc error: code = FailedPrecondition desc = operation not allowed in state NOT_SERVING
mysql> select * from t2;
ERROR 1105 (HY000): vtgate: http://localhost:15001/: vttablet: rpc error: code = FailedPrecondition desc = operation not allowed in state NOT_SERVING
  ```
  * Note the error, and the fact that it persists for that connection.
  * If you disconnect the mysql client and try again, it will work:
  ```
mysql> use keyspace3;
Database changed
mysql> select * from t2;
+----+------+
| c1 | c2   |
+----+------+
|  1 | NULL |
+----+------+
1 row in set (0.00 sec)
  ```

systay:
## Description
Make sure to reconnect reserved connections to tablets that are not answering any more.

## Related Issue(s)
Fixes #7875 

## Checklist
- [x] Should this PR be backported? `Yes!`

## Impacted Areas in Vitess

- [x]  Query Serving


go/test/endtoend/vtgate/reservedconn/reconnect1/main_test.go
```diff
@@ -0,0 +1,151 @@
+/*
+Copyright 2020 The Vitess Authors.
```
deepthi, line 2: Copyright year 2021

go/vt/discovery/healthcheck.go
```diff
@@ -45,19 +45,19 @@ import (
...
 
 	"vitess.io/vitess/go/flagutil"
-
-	"vitess.io/vitess/go/vt/topo"
-
 	"vitess.io/vitess/go/stats"
 	"vitess.io/vitess/go/vt/log"
-	"vitess.io/vitess/go/vt/proto/query"
-	"vitess.io/vitess/go/vt/proto/topodata"
-	"vitess.io/vitess/go/vt/topo/topoproto"
```
deepthi, line 60: Can we undo the imports changes? We will be backporting this PR.

deepthi (approved):
LGTM, a couple of nits

go/vt/vterrors/constants.go
```diff
@@ -0,0 +1,34 @@
...
+)
+
+// RxOp regex for operation not allowed error
+var RxOp = regexp.MustCompile("operation not allowed in state (NOT_SERVING|SHUTTING_DOWN)")
+
+// WrongTablet for invalid tablet type error
+const WrongTablet = "wrong tablet type"
+
+// RxWrongTablet regex for invalid tablet type error
+var RxWrongTablet = regexp.MustCompile(WrongTablet)
```
harshit-gangal, line 34: error message cannot change as it would force an upgrade order. We should keep the error message same as what the tablet returns today. If we want to change the error message we would do it over 2 releases.

systay: can we change the error message and look for both the old and new message?

systay: could we change the message and check for both messages?

PR #7885, opened 2021-04-18
[vtadmin] Promote ErrNoSchema to a TypedError which returns http 404

### LINKED ISSUE #7820: [vtadmin-api] /api/schema route should return an error for schemas that don't exist
#### Overview of the Issue

Currently, vtadmin-api returns a "successful-looking" response for `/api/schema/...` requests, even when no schema exists for the request parameters. 

At a minimum, the response should have `ok: false`. A 404 status code would be better than a 500 (although in practice it doesn't make much difference to the front-end.) 

We should also update the `/schema` view on the front-end, since [the current `is404` mechanism](https://github.com/vitessio/vitess/blob/ea0c13e/web/vtadmin/src/components/routes/Schema.tsx#L44) is not ideal for tables that don't have a `table_definition` property. 

#### Reproduction Steps

Running vtadmin-api locally, `http://localhost:14200/api/schema/id1/commerce/does-not-exist` returns:

```json
{"result":{"cluster":{"id":"id1","name":"cluster1"},"keyspace":"commerce"},"ok":true}
```

Compare with `http://localhost:14200/api/tablet/does-not-exist`, which returns a `500` status code and the following data:

```json
{"error":{"message":"no such tablet: does-not-exist, searched clusters = [id1]","code":"unknown"},"ok":false}
```

ajm188:

## Related Issue(s)

- Closes #7820 

## Checklist
- [ ] Should this PR be backported? **no**
- [ ] Tests were added or are not required **n/a**

## Impacted Areas in Vitess

- [ ]  Query Serving
- [ ]  VReplication
- [ ]  Cluster Management
- [ ]  Build/CI
- [x]  VTAdmin

### DIFF
--- a/go/vt/vtadmin/errors/typed_error.go
+++ b/go/vt/vtadmin/errors/typed_error.go
@@ -18,6 +18,7 @@ package errors
 
 import (
 	"fmt"
+	"net/http"
 	"strings"
 )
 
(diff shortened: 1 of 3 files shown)


go/vt/vtadmin/errors/typed_error.go
```diff
@@ -77,3 +78,29 @@ func (e *MissingParams) Error() string {
...
+
+	if e.Clusters != nil {
+		details["clusters"] = e.Clusters
+	}
+
+	return details
+}
+
+func (e *NoSuchSchema) Code() string    { return "no such schema" }
+func (e *NoSuchSchema) HTTPStatus() int { return http.StatusNotFound }
```
ajm188, line 106: Open TODO for anyone to swap the rest of our magic numbers in this file to reference the constants in `net/http`

doeg: Do you mean in this PR, or a separate one?

doeg (approved):
Thanks for doing this!

go/vt/vtadmin/errors/typed_error.go
ajm188: Separate one! It's not really related to this change, but something we should do. https://github.com/vitessio/vitess/projects/12#card-59345039

PR #7886, opened 2021-04-18
[vtadmin] gate/tablet/vtctld FQDNs

ajm188:
## Description

This PR introduces an optional FQDN field to gates/tablets/vtctld proto message types, along with options to specify a go template for generating an FQDN from the rest of the struct fields in a gate/tablet/vtctld.

For gates and vtctlds, it's up to the particular discovery implementations to populate those fields, and for tablets it is handled by the `Cluster`. Consequently, the gate/vtctld template flags are pushed down into the `ConsulDiscovery` implementation (and `StaticFileDiscovery` can just read the "FQDN" field directly from the static file, so I did not include an option), while the tablet template flag is parsed as part of the cluster DSN.

**One note**: I named the field "FQDN" rather than "fqdn" in the proto file, so that it would generate the horrid "Fqdn" Go field name; the side effect of me doing this is that when marshaling to JSON, the key is in all-caps rather than all-lowercase like the rest of the fields. **If we feel that json field casing consistency is more valuable**, then I can bear with the `Fqdn`.

I also added a generic function to `package textutil` to remove the duplicated "create buffer; execute template into buffer; return resulting string" that we have in like 6 places now.

_Final note_: I'm really bad at typing "fqdn" apparently, and kept making the typo for "fdqn", so something to keep an eye out for when reviewing.

## Related Issue(s)

- https://github.com/vitessio/vitess/projects/12#card-57032043

## Checklist
- [ ] Tests were added or are not required **N/A**

## Impacted Areas in Vitess

- [ ]  Query Serving
- [ ]  VReplication
- [ ]  Cluster Management
- [ ]  Build/CI
- [x]  VTAdmin


go/vt/vtadmin/cluster/discovery/discovery_consul.go
```diff
@@ -91,45 +93,69 @@ func NewConsul(cluster *vtadminpb.Cluster, flags *pflag.FlagSet, args []string)
...
 		disco.vtgateDatacenter, err = generateConsulDatacenter("vtgate", cluster, *vtgateDatacenterTmplStr)
 		if err != nil {
 			return nil, fmt.Errorf("failed to generate vtgate consul datacenter from template: %w", err)
 		}
 	}
 
+	if *vtgateFQDNTmplStr != "" {
+		disco.vtgateFQDNTmpl, err = template.New("consul-vtgate-fqdn-template-" + cluster.Id).Parse(*vtgateFQDNTmplStr)
+		if err != nil {
+			return nil, fmt.Errorf("failde to parse vtgate FQDN template %s: %w", *vtgateFQDNTmplStr, err)
```
doeg, line 135: Typo: "failde" (there's another instance lower in the file)

go/vt/vtadmin/cluster/discovery/discovery_consul.go
```diff
@@ -252,6 +283,15 @@ func (c *ConsulDiscovery) discoverVTGates(_ context.Context, tags []string) ([]*
 			vtgate.Keyspaces = strings.Split(keyspaces, ",")
 		}
 
+		if executeFQDNTemplate {
+			if c.vtgateFQDNTmpl != nil {
+				vtgate.FQDN, err = textutil.ExecuteTemplate(c.vtgateFQDNTmpl, vtgate)
+				if err != nil {
+					return nil, err // TODO: wrap
```
doeg, line 290: 🔪 TODO?

go/vt/vtadmin/cluster/discovery/discovery_consul.go
```diff
@@ -184,28 +211,32 @@ func (c *ConsulDiscovery) DiscoverVTGateAddr(ctx context.Context, tags []string)
...
 	defer span.Finish()
 
-	return c.discoverVTGates(ctx, tags)
+	executeFQDNTemplate := true
+
+	return c.discoverVTGates(ctx, tags, executeFQDNTemplate)
 }
 
-func (c *ConsulDiscovery) discoverVTGates(_ context.Context, tags []string) ([]*vtadminpb.VTGate, error) {
+func (c *ConsulDiscovery) discoverVTGates(_ context.Context, tags []string, executeFQDNTemplate bool) ([]*vtadminpb.VTGate, error) {
```
doeg, line 239: Despite looking all at the callsites, I'm not sure I follow why the `executeFQDNTemplate` boolean is necessary, or when I'd want to use true vs. false. 🤔 Maybe a comment would help? (Same comment about `discoverVtctld`'s parameter.)

go/vt/vtadmin/cluster/cluster.go
```diff
@@ -388,6 +406,26 @@ func (c *Cluster) findWorkflows(ctx context.Context, keyspaces []string, opts Fi
...
+// GetGates returns the list of all VTGates in the cluster.
+func (c *Cluster) GetGates(ctx context.Context) ([]*vtadminpb.VTGate, error) {
+	// (TODO|@ajm188) Support tags in the vtadmin RPC request and pass them
+	// through here.
+	gates, err := c.Discovery.DiscoverVTGates(ctx, []string{})
+	if err != nil {
+		return nil, fmt.Errorf("DiscoverVTGates(cluster = %s): %w", c.ID, err)
+	}
+
+	// Not every discovery implementation will necessarily populate the Cluster
```
doeg, line 418: This is pretty nitpicky so feel free to leave it, but a more straightforward phrasing might be:

```
// This overwrites any "Cluster" field populated by the discovery implementation
```

I know that the code is _right there_, but the current comment (to me) more readily implies that the "Cluster" field will be populated iff it doesn't already exist from the discovery implementation. Like I said... nitpicky. 😎

doeg (commented):
Thanks so much for doing this. It's going to be so useful, especially for "internal deprecation" reasons. 😼 

Just a couple of minor comments + questions but otherwise this looks great! Go templates make this so much nicer than the equivalent front-end implementation.

go/vt/vtadmin/cluster/discovery/discovery_consul.go
ajm188: it's the Olde English of "failed" (thank you!)

go/vt/vtadmin/cluster/cluster.go
ajm188: No this distinction makes sense to me! Thank you!

go/vt/vtadmin/cluster/discovery/discovery_consul.go
ajm188: In practice it's an optimization for `Discover{Vtctld,VTGate}Addr`. If we didn't have this flag to thread through, then we would make one `ExecuteTemplate` call per gate (read also "vtctld") as part of the discovery (to generate the FQDN), pick one gate, and make a final `ExecuteTemplate` call to generate the dialable address, for a total of N+1 calls to `ExecuteTemplate`.

Since everything that's available to `ExecuteTemplate` when we're generating the address is also available when generating the FQDN, we can force the user to (potentially*) be a little redundant in the config and save the N calls for generating FQDNs and make a single `ExecuteTemplate` call for the final address.

\* "potentially" only because it's redundant if your dial address is something like "FQDN + gRPC port", then you would need a config of:

```
-fqdn-tmpl "{{ .Hostname }}.example.com"
-addr-tmpl "{{ .Hostname }}.example.com:15000"
```

whereas if we didn't do this optimization then you could specify something like

```
-fqdn-tmpl "{{ .Hostname }}.example.com"
-addr-tmpl "{{ .FQDN }}:15000"
```

doeg (approved):

PR #7782, opened 2021-04-01
Fuzzing: Fixup oss-fuzz build script

AdamKorcz:
This PR adds two more fuzzers to the build script. 

The coverage build is also deactivated since there are issues with getting it to work.

PR #7902, opened 2021-04-20
Planbuilder: Add fuzzer

AdamKorcz:
This PR adds a fuzzer for vtgate/planbuilder.

An update in the OSS-fuzz build script is required, but I will add that in a separate PR since https://github.com/vitessio/vitess/pull/7782 is currently a blocker for that.


systay (approved):
this is not really what we talked about - this is not at all an endpoint.

but I'm really curious about what you find, so why not, right? :)

PR #7904, opened 2021-04-20
ci: fix all racy tests

vmg:
## Description

This PR fixes all the racy tests in our unit test suite that I've managed to reproduce locally. @shlomi-noach: these are all legit and should get the build closer to consistently green, but if you can think of any flaky race that I've missed, please let me know.

## Related Issue(s)

- 

## Checklist

## Impacted Areas in Vitess

- [ ]  Query Serving
- [ ]  VReplication
- [ ]  Cluster Management
- [x]  Build/CI
- [ ]  VTAdmin

### DIFF
--- a/go/vt/vtadmin/cluster/discovery/discovery.go
+++ b/go/vt/vtadmin/cluster/discovery/discovery.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"sync"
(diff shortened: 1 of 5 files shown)


shlomi-noach (approved):
looks real good. Awesome work on all fronts!

deepthi (approved):
:100:

Issue #7906, opened 2021-04-20
Update health should not nuke out all healthy tablet when the primary tablet returned error


narcsfz:
#### Overview of the Issue
I think the change we made here https://github.com/vitessio/vitess/pull/7654/files#diff-b52bf818696494288a93bba0e91b8ff75fb2fd109bb1c80c77d166ea8349b722R452 might leads to some unexpected behavior during an unplanned external reparent failover.

For example, there are 3 tablets t1 (primary), t2, t3. Now t1 crashed and t2 becomes the new primary through an external reparent process. t2 will notify VTGate via the stream connection so that VTGate knows there is a new primary.

However, because t1 crashed, the stream connection also closed, we will jump into updateHealth with `isPrimary = true; isPrimaryUp = false` from the check conn [here](https://github.com/vitessio/vitess/pull/7654/files#diff-c87b92e62c8d14accef2560231f8abeb6e54f66b98fe5d519b15ad44b2b74821R309) and nuke out all the primary tablet -- including the updated one from the healthcheck stream and VTGate needs to wait the poll from topo server to recover from this case.

IMO, we might just want to do delete the crashed tablet from `hc.healthy[targetKey]` here instead of nuking it out completely. 

#### Reproduction Steps
See the description above

#### Binary version
We are on top of `fe6d095ccae0f50356d4691b0171bbeb8778e911`

#### Operating system and Environment details
```
$ cat /etc/os-release
NAME="Ubuntu"
VERSION="20.04.2 LTS (Focal Fossa)"
```

cc @deepthi

deepthi:
We will probably almost always receive an update from t2 before t1 times out, so this needs to be fixed.

wangmeng99:
The health check for the crashing primary and one for the newly appointed primary could always race. We can use `MasterTermStartTime` to differentiate which check wins.

narcsfz:
I think we already use ` MasterTermStartTime` when we do the insertion on newly appointed primary case: https://github.com/vitessio/vitess/blob/master/go/vt/discovery/healthcheck.go#L438

When the old primary crashed, we just need to remove the tablet from healthy tablet map (without insertion). So I think we don't need compare `MasterTermStartTime` in this case bc we don't want to update/insert the master record

wangmeng99:
> I think we already use ` MasterTermStartTime` when we do the insertion on newly appointed primary case: https://github.com/vitessio/vitess/blob/master/go/vt/discovery/healthcheck.go#L438
> 
[...]

PR #7915, opened 2021-04-21
Check tablet alias before removing after error stream

narcsfz:

## Description

When VTGate gets an error from the primary tablet connection, it should not just nuke out all the healthy tablets. Because an external reparent process might have already changed the primary tablet.

Instead we should only try to remove the tablet with stream error from the map

## Related Issue(s)

- https://github.com/vitessio/vitess/issues/7906

## Checklist

## Impacted Areas in Vitess

- [ ]  Query Serving
- [ ]  VReplication
- [X]  Cluster Management
- [ ]  Build/CI
- [ ]  VTAdmin

### DIFF
--- a/go/vt/discovery/healthcheck.go
+++ b/go/vt/discovery/healthcheck.go
@@ -391,6 +391,11 @@ func (hc *HealthCheckImpl) deleteTablet(tablet *topodata.Tablet) {
 	th.cancelFunc()
 	delete(hc.healthByAlias, tabletAlias)
 	// delete from map by keyspace.shard.tabletType
+	hc.deleteHealthDateLocked(key, tabletAlias)
+}
+
+func (hc *HealthCheckImpl) deleteHealthDateLocked(key keyspaceShardTabletType, tabletAlias tabletAliasString) {
+	// delete from map by keyspace.shard.tabletType
 	ths, ok := hc.healthData[key]
 	if !ok {
 		log.Warningf("We have no health data for target: %v", key)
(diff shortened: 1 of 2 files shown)


go/vt/discovery/healthcheck.go
```diff
@@ -448,8 +453,7 @@ func (hc *HealthCheckImpl) updateHealth(th *TabletHealth, prevTarget *query.Targ
 			}
 		}
 	case isPrimary && !isPrimaryUp:
-		// No healthy master tablet
-		hc.healthy[targetKey] = []*TabletHealth{}
+		hc.deleteHealthDateLocked(targetKey, tabletAlias)
```
deepthi, line 456: I don't believe this is correct. We want to delete it from `healthy` but not from `healthData`.

narcsfz: Updated to keep it same as the original logic.

deepthi (approved):
LGTM

deepthi:
This should also go into 10.0 (release-10.0 branch)

go/vt/discovery/healthcheck.go
```diff
@@ -448,8 +448,16 @@ func (hc *HealthCheckImpl) updateHealth(th *TabletHealth, prevTarget *query.Targ
 			}
 		}
 	case isPrimary && !isPrimaryUp:
-		// No healthy master tablet
-		hc.healthy[targetKey] = []*TabletHealth{}
+		if healthy, ok := hc.healthy[targetKey]; ok && len(healthy) > 0 {
+			var newHealthy []*TabletHealth
```
narcsfz, line 452: i don't know if this is the proper way to do the deletion in this case, please let me know if you have better suggestions

deepthi: This is fine too.

PR #7923, opened 2021-04-22
vtctl: return error on invalid ddl_strategy

shlomi-noach:
## Description

Quick fix for a minor bug: if you run `vtctl ApplySchema -ddl_strategy=foobar ...`, then `vtctl` must return error because `foobar` is an invalid strategy. Right now it skips the migration but does not report an error.

Will supply a backport for v10

## Checklist

## Impacted Areas in Vitess

- [x]  Query Serving
- [ ]  VReplication
- [ ]  Cluster Management
- [ ]  Build/CI
- [ ]  VTAdmin


ajm188 (approved):
lgtm

PR #7929, opened 2021-04-22
vttablet/tabletmanager: add isInSRVKeyspace/isShardServing

guidoiaquinti:
Signed-off-by: Guido Iaquinti <giaquinti@slack-corp.com>

## Description
Add a new metric in `vttablet` to keep track if the keyspace partition it belongs to is serving or not. This metric can be useful for external systems to automatically enable/disable alerts.

In order to don't add additional watches to the topology service (as we don't want those to linearly scale with the number of vttablets), we refresh the serving keyspace view in TabletManager only when `RefreshFromTopoInfo` is called.

Note: this PR is still a WIP but I'll appreciate an early feedback. 

## Related Issue(s)
N/A

## Checklist

## Deployment Notes
N/A

## Impacted Areas in Vitess

- [ ]  Query Serving
- [ ]  VReplication
- [X]  Cluster Management
- [ ]  Build/CI
- [ ]  VTAdmin

### DIFF
--- a/go/vt/vttablet/tabletmanager/tm_init.go
+++ b/go/vt/vttablet/tabletmanager/tm_init.go
@@ -94,6 +94,9 @@ var (
 	// statsBackupIsRunning is set to 1 (true) if a backup is running.
 	statsBackupIsRunning *stats.GaugesWithMultiLabels
 
+	// statsIsInSRVKeyspace is set to 1 (true), 0 (false) whether the tablet is in the serving keyspace
+	statsIsInSRVKeyspace *stats.Gauge
+
 	statsKeyspace      = stats.NewString("TabletKeyspace")
 	statsShard         = stats.NewString("TabletShard")
 	statsKeyRangeStart = stats.NewString("TabletKeyRangeStart")
@@ -115,6 +118,7 @@ func init() {
 	statsTabletType = stats.NewString("TabletType")
 	statsTabletTypeCount = stats.NewCountersWithSingleLabel("TabletTypeCount", "Number of times the tablet changed to the labeled type", "type")
 	statsBackupIsRunning = stats.NewGaugesWithMultiLabels("BackupIsRunning", "Whether a backup is running", []string{"mode"})
+	statsIsInSRVKeyspace = stats.NewGauge("IsInSRVKeyspace", "Whether the vttablet is in the serving keyspace (1 = true / 0 = false)")
 }
 
 // TabletManager is the main class for the tablet manager.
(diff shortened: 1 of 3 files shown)


go/vt/vttablet/tabletmanager/tm_init.go
```diff
@@ -115,6 +118,7 @@ func init() {
 	statsTabletType = stats.NewString("TabletType")
 	statsTabletTypeCount = stats.NewCountersWithSingleLabel("TabletTypeCount", "Number of times the tablet changed to the labeled type", "type")
 	statsBackupIsRunning = stats.NewGaugesWithMultiLabels("BackupIsRunning", "Whether a backup is running", []string{"mode"})
+	statsIsInSRVKeyspace = stats.NewGauge("IsInSRVKeyspace", "Whether the vttablet is in the serving keyspace (1 = true / 0 = false)")
```
deepthi, line 121: should is `IsInSrvKeyspace` otherwise the prometheus metric might turn out to be `vttablet_is_in_s_r_v_keyspace`

guidoiaquinti: > Should be `IsInSrvKeyspace` otherwise the prometheus metric might turn out to be `vttablet_is_in_s_r_v_keyspace`

Actually I think it is rendered correctly:
```
# HELP vttablet_is_in_srv_keyspace Whether the vttablet is in the serving keyspace (1 = true / 0 = false)
# TYPE vttablet_is_in_srv_keyspace gauge
vttablet_is_in_srv_keyspace 0
```

guidoiaquinti:
I found out that with this implementation the topology info is not propagated to the tablet manager state after operations like `SwitchReads`/`SwitchWrites` are completed.

go/vt/vttablet/tabletmanager/tm_init.go
deepthi: TIL :)

guidoiaquinti:
> I found out that with this implementation the topology info is not propagated to the tablet manager state after operations like `SwitchReads`/`SwitchWrites` are completed. I need to take a better look at this.

I think I found the underlying issue: in few places of the codebase we call `MigrateServedType` that _removes/adds shards from srvKeyspace when migrating a served type_ but then we never refresh the tablet manager state.

I think the right approach here to don't end up with stale info in the tablet manager state is to make sure we call `RefreshTabletsByShard` every time the service keyspace view changes (despite this will add additional topo lookup).

rafael (approved):
Guido this looks great! I added a small comment, but I think that could be part of a different PR. 

Going to merge this one as is.

go/vt/vttablet/tabletmanager/tm_state_test.go
```diff
@@ -154,6 +155,158 @@ func TestStateTabletControls(t *testing.T) {
 	assert.False(t, qsc.IsServing())
 }
 
+func TestStateIsShardServingisInSrvKeyspace(t *testing.T) {
```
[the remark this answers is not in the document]
guidoiaquinti: Make sense, let me add that. Thanks!

PR #7810, opened 2021-04-08
docker/vttestserver:  Set max_connections in default-fast.cnf at container build time

mcronce:
## Description
In both `vttestserver` container images, this sets `max_connections` to 1000 on the underlying MySQL daemons

## Related Issue(s)
N/A

## Checklist
- [ ] Should this PR be backported?
- [X] Tests were added or are not required
- [X] Documentation was added or is not required

## Deployment Notes
N/A

## Impacted Areas in Vitess
Components that this PR will affect:

- [ ]  Query Serving
- [ ]  VReplication
- [ ]  Cluster Management
- [ ]  Build/CI
- [ ]  VTAdmin

PR #7806, opened 2021-04-08
Makefile: add cross-build target for cross-compiling client binaries

gedgar:
Signed-off-by: Gary Edgar <gary@planetscale.com>

## Description
Adds a `cross-build` target which allows you to cross-compile Vitess binaries.
This makes it easier build a slim client tarball (namely, vtexplain + vtctl{,d}client) for OSX from Linux, which we can start including in PlanetScale's biweekly Vitess release. 

## Related Issue(s)

- 

## Checklist
- [ ] Should this PR be backported?
- [x] Tests were added or are not required
- [x] Documentation was added or is not required

## Impacted Areas in Vitess
Components that this PR will affect:

- [ ]  Query Serving
- [ ]  VReplication
- [ ]  Cluster Management
- [x]  Build/CI
- [ ]  VTAdmin

PR #7796, opened 2021-04-06
Removing Linux/amd64 specific subdependencies

shlomi-noach:
## Description

Reported by @systay, the merging of https://github.com/vitessio/vitess/pull/7725 introduced a subdependency `github.com/moby/term` which does not build on MacOS:

```
> make build
Tue Apr 6 14:14:12 CEST 2021: Building source tree
# github.com/moby/term
../go/pkg/mod/github.com/moby/term@v0.0.0-20200429084858-129dac9f73f6/tc.go:13:28: undefined: unix.SYS_IOCTL
../go/pkg/mod/github.com/moby/term@v0.0.0-20200429084858-129dac9f73f6/tc.go:18:28: undefined: unix.SYS_IOCTL
../go/pkg/mod/github.com/moby/term@v0.0.0-20200429084858-129dac9f73f6/termios_bsd.go:24:31: undefined: unix.SYS_IOCTL
../go/pkg/mod/github.com/moby/term@v0.0.0-20200429084858-129dac9f73f6/termios_bsd.go:37:31: undefined: unix.SYS_IOCTL
make: *** [build] Error 2
```

This is a subdependency of `github.com/fsouza/go-dockerclient`, which is in itself a dependency of `tengo`.

We already user our own [fork](https://github.com/planetscale/tengo) of `tengo`. I've submitted and merged https://github.com/planetscale/tengo/pull/5, which removes any `docker` references, none of which we use. This removes the dependencies  on aforementioned libraries. I tagged as `v0.9.6-ps.v2`

This PR instructs `go.mod` to use `v0.9.6-ps.v2`.

## Checklist
- [x] Should this PR be backported? (to `v10`)
- [x] Tests were added or are not required
- [x] Documentation was added or is not required

## Impacted Areas in Vitess
Components that this PR will affect:

- [ ]  Query Serving
- [ ]  VReplication
- [ ]  Cluster Management
- [x]  Build/CI
- [ ]  VTAdmin

PR #7932, opened 2021-04-22
Release notes generation

systay:
This PR changes how Vitess creates release-notes for releases.

We'll keep track of which PRs get merged by looking at the git log, and produce the release notes from that.
To group PRs in useful buckets, we'll use labels a bit more than today.

The command looks something like this:
```
make FROM="upstream/release-10.0" TO="upstream/master" release-notes
```
This would generate the release notes for all changes that have gone into the `master` branch since `release-10.0` was forked from it.

The produced release notes look something like this:

_____
### Query Serving
 - Routing of Information Schema Queries #7841
 - Panic when EOF after @ symbol #7925
### Cluster management
 - make: build vitess as static binaries by default #7795 
### Other
 - update moby/term to fix darwin build issue #7787
 - Removing Linux/amd64 specific subdependencies #7796
 - Makefile: add cross-build target for cross-compiling client binaries #7806
 - Bump MySQL version for docker builds to 8.0.23 #7811 
### vttestserver
 - docker/vttestserver:  Set max_connections in default-fast.cnf at container build time #7810
_____

First PRs are grouped by type, and then by component. If a PR is missing either, "Other" will be used instead.


frouioui:
@systay - 93d1906 introduces the rendering of `prPerType` using go text templates. A sample output is:

```
## Other 
### Component: Query Serving
 - Fix bug with reserved connections to stale tablets #7879 
### Other
 - Minor cleanups around errors on the vtgate #7864
 - vreplication: performance & benchmarks (table copying) #7881
 - Planbuilder: Add fuzzer #7902
 - Memory Sort to close the goroutines when callback returns error  #7903
 - Check tablet alias before removing after error stream #7915
## Type: Bug 
### Component: Query Serving
 - Panic when EOF after @ symbol #7925
```

go/tools/release-notes/release_notes.go
```diff
@@ -0,0 +1,200 @@
+package main
```
systay, line 1: license

frouioui:
Note: we were originally planning on sorting PR information (inside each component) based on their merged date. However, the issue described here https://github.com/cli/cli/issues/3497 occurs.

.github/pull_request_template.md
```diff
@@ -10,23 +10,19 @@
 <!-- A few sentences describing the overall goals of the pull request's commits. -->
 
 ## Related Issue(s)
-<!-- List related issues and pull requests: -->
+<!-- List related issues and pull requests. If this PR fixes an issue, please add it using Fixes #????  -->
 
-- 
 
 ## Checklist
-- [ ] Should this PR be backported?
```
deepthi, line 18: Do we use the `Backport me` label instead?

deepthi (commented):
This is awesome!
For anyone else following along, we will use this to generate an initial release notes which will then be edited by hand (curated) to produce a final version.

.github/pull_request_template.md
frouioui: Correct. The intention is to get rid of these true/false checklists in pull requests, and instead, use them as Todos.

deepthi: Then let us add that guidance to the PR template as well.

frouioui: Added that guidance in c1f6d5b.

deepthi (approved):
:tada:

.github/pull_request_template.md
```diff
@@ -10,23 +10,21 @@
...
-- [ ]  VReplication
-- [ ]  Cluster Management
-- [ ]  Build/CI
-- [ ]  VTAdmin
+For each PR, try to find one `Type:` label and one `Component:` label. 
+This will help in finding a reviewer, and in creating release notes for the next release.
+
+If your PR needs to be backported, add the label `Backport me!`.
+
+-->
```
shlomi-noach, line 30: Please consider that only people with `Write` privileges on our GitHub repo, are able to create labels. Which means the casual contributor is unable to classify their PR by labeling.

I'm not sure if this undermines the purpose of this PR...

go/tools/release-notes/release_notes.go
```diff
@@ -0,0 +1,264 @@
...
+		})
+		data = append(data, newPrType)
+	}
+	sort.Slice(data, func(i, j int) bool {
+		return data[i].Name < data[j].Name
+	})
+	return data
+}
+
+func writePrInfos(fileout string, prPerType map[string]map[string][]prInfo) (err error) {
```
shlomi-noach, line 222: > map[string]map[string][]prInfo

Can we create cascading `tyepdef` declarations so that we have something like `PullRequests`, `PullRequestLabels` or whatever it is that is implied by these `map`s?

shlomi-noach (changes requested):

.github/pull_request_template.md
systay: I don't think it does. We as maintainers should be able to handle labelling before PRs get merged.
And about backporting - deciding what gets backported and what does not has in my experience always been a discussion between maintainers. 
Maybe we should just remove this comment altogether from the PR template.

shlomi-noach (approved):

.github/pull_request_template.md
systay: And I just removed it. @shlomi-noach raised the valid objection that you need write access to the repo to be able to assign labels. I tweaked the text so it's aimed at new contributors without write access, and not aimed at maintainers.

My suggestion is that we maintainers do it for our own PRs and for any PRs we merge.

Alternatively, we could preface the label section with something that this only applies persons with write access to the repo. Not sure which is the better option `¯\_(ツ)_/¯`

shlomi-noach (commented):
perhaps add a HTML comment `<!-- -->` in the PR template to explain how we should go about labeling? Not sure if it's a great idea, because it's aimed for us the maintainers rather than the contributors. At your discretion.

PR #7933, opened 2021-04-22
vreplication: dynamic packet sizing

vmg:
## Description

:skull_and_crossbones: **Warning: Spicy PR ahead** :skull_and_crossbones:

As discussed on last week's VReplication performance improvements, I noticed that table copying performance was basically being bottlenecked by MySQL bulk insert performance. After some testing on staging environments, I found out that the simplest way to improve MySQL insert performance was tweaking the packet size (`-vstream_packet_size`) on the vttablet source: **depending on the layout and size of the tables being copied, careful tuning of this configuration option can make table copying up to 30% faster**.

This is good, because it means advanced Vitess users can configure their clusters to squeeze more performance out of VReplication, but it is also bad, because it means you _need_ be an advanced user and be able to understand how to tweak the packet sizes based on your use case. So, in the spirit of PlanetScale, it occurred to me: **can we make Vitess more magical so it always does the right thing?**

And it turns out we can! This PR introduces a new configuration option `-vstream_dynamic_packet_size`, which turns on an extension to the VReplication controller that **measures the throughput of the replication stream and uses statistical analysis to adjust its packet size for optimal performance**. The idea is good, but the devil is in the details, of course: to ensure this real-time optimization is as effective as it can be, the dynamic packet sizer performs repeated experiments to sample the effects of packet size changes. It keeps track of the response times from previous stream calls and aggregates them into statistical samples by removing any outliers and performing a two-sample Welch T-Test to ensure changes in packet size are _actually_ faster. Tthis is the same statistical analysis that I usually perform when analysing benchmark results, to ensure optimizations are meaningfully faster.

To accomplish all this, I've had to vendor some of the maths logic from `golang.org/x/perf/benchstat` into Vitess; the meaningful math code is in an `internal` package so it cannot be used as a dependency directly. I believe this is fine because these kinds of statistic analysis functions are probably not being changed or improved anytime soon. Relatedly: I decided to use a Welch T-Test instead of the default U-test that the `benchstat` package uses because the later runs in `O(n^2)` memory, while the T-Test uses constant memory. I don't think the accuracy difference between the tests is meaningful, or more accurately, Wikipedia thinks this is fine. Both tests should be statistically meaningful and the T-test doesn't allocate, sooo...

### Testing

I've started by verifying in the end-to-end performance harness that this change meaningfully improves table copying performance. It does for all testing I've been able to do locally -- it seems to consistently improve on the default 25000 packet size setting from Vitess.

Besides that, as you may guess, this is a complex feature to unit test. In order to give some sanity to the implementation, I've implemented a simulation as a unit test: basically, I've graphed the performance of several MySQL bulk insert behaviors and curve-fitted the response time graphs into several polynomials. The unit tests are using this curve fitting to simulate the response times of a MySQL server and to ensure that the packet size adjustments give a meaningful performance impact.

## Checklist

## Impacted Areas in Vitess

- [ ]  Query Serving
- [x]  VReplication
- [ ]  Cluster Management
- [ ]  Build/CI
- [ ]  VTAdmin


vmg:
To note: right now the performance measures are fully client side. It measures the roundtrip time for each insert batch as seen for the client. This could easily be improved to make the server collaborate and report the actual insert timing, but I think there's value in this client-side performance metric, because it takes into account network throughput and latency, something that I believe is also meaningful when tuning the packet sizes.

go/vt/vttablet/tabletserver/vstreamer/packet_size.go
```diff
@@ -0,0 +1,184 @@
...
+			return
+		}
+
+		change := ps.changeInThroughput()
+		switch change {
+		case notChanging, gettingSlower:
+			if len(ps.candidate.Xs) >= SettleCandidateLen {
+				ps.target = ps.last
+				ps.settled = true
+				log.Infof("packetSize: not changing for %d, settled at %d", len(ps.candidate.Xs), ps.target)
```
rohit-nayak-ps, line 167: How often will these statements log? If it is going to happen a lot we should probably comment them out since they can spam the logs.

go/vt/vttablet/tabletserver/vstreamer/packet_size_test.go
```diff
@@ -0,0 +1,86 @@
...
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			ps1 := newDynamicPacketSizer(tc.baseSize)
+			elapsed1, sent1 := simulate(t, ps1, tc.baseSize, tc.baseSize*1000, tc.p.fit)
+
+			ps2 := newFixedPacketSize(tc.baseSize)
+			elapsed2, sent2 := simulate(t, ps2, tc.baseSize, tc.baseSize*1000, tc.p.fit)
+
+			t.Logf("dynamic = (%v, %d), fixed = (%v, %d)", elapsed1, sent1, elapsed2, sent2)
+			require.True(t, elapsed1 < elapsed2)
```
rohit-nayak-ps, line 82: Can you add a comment here what we are comparing here and why the assertions in the test are valid?

rohit-nayak-ps (changes requested):
Magical PR as usual :)

I don't understand the statistical stuff but the changes to the vreplication code look good.

I suggest adding some per-stream vstreamer metrics to externalize the packet size (current, max, average). This will help provide more insights into how the algorithm is working in production setups both for users and (maybe) for us to tune it in specific circumstances. Examples here: go/vt/vttablet/tabletserver/vstreamer/engine.go:108

go/vt/vttablet/tabletserver/vstreamer/packet_size.go
```diff
@@ -0,0 +1,184 @@
+package vstreamer
```
systay, line 1: license missing

XuanThuLe:
Ok

XuanThuLe (approved):

XuanThuLe:
Ok

go/vt/vttablet/tabletserver/vstreamer/packet_size.go
XuanThuLe: Ok

XuanThuLe: Ok

go/vt/vttablet/tabletserver/vstreamer/packet_size.go
```diff
@@ -0,0 +1,210 @@
...
+}
+
+// Record records the total duration it took to send the given byte count while streaming
+func (ps *fixedPacketSizer) Record(_ int, _ time.Duration) {}
+
+type dynamicPacketSizer struct {
+	// currentSize is the last size for the packet that is safe to use
+	currentSize int
+
+	// current are the performance metrics for the last
```
rohit-nayak-ps, line 88: minor: current => currentMetrics

rohit-nayak-ps (approved):
lgtm

go/vt/vttablet/tabletserver/vstreamer/rowstreamer.go
```diff
@@ -237,11 +240,9 @@ func (rs *rowStreamer) streamQuery(conn *snapshotConn, send func(*binlogdatapb.V
 	byteCount := 0
 	for {
 		//log.Infof("StreamResponse for loop iteration starts")
-		select {
```
rohit-nayak-ps, line 240: Will we always get ctx.Err() as not nil when the context expires? We just call ctx.Done() on timeouts, for example.

vmg: @rohit-nayak-ps: Yes, you can verify the implementation on https://golang.org/src/context/context.go#L394 -- this is a known performance fix. Doing a `default` read on the `Done()` channel is always equivalent to checking `Err()`, because the only implementation for this functionality is in the `cancelCtx` which behaves this way. It doesn't matter if this is a manually canceled context, timeout or duration context.

rohit-nayak-ps: Ah, yes of course, it sets the err to Cancelled

Issue #7931, opened 2021-04-22
Migrate VDiff rpc to VtctldServer


ajm188:
#### Feature Description

We should have a gRPC endpoint to call VDiff on the new VtctldServer, similar to the `vtctlclient VDiff` vtctl command.

#### Use Case(s)

Programmatically calling Vdiff from the new clients.

#### Notes

This is going to involve a series of preperatory refactors to enable us to call TrafficSwitcher/VDiff code (vdiffer relies on traffic switcher) from outside of `package wrangler` and _then_ it should be a simple final change to add the RPC endpoint.

ajm188:
closing as dupe of #12155

PR #7934, opened 2021-04-22
[workflow] extract migration targets from wrangler

ajm188:
## Description

This is the first in a series of PRs to make VDiff runnable from the new `grpcvtctldserver` (in order to do that we eventually need the `vdiff` struct to be defined outside of `package wrangler` so that both `wrangler` and `grpcvtctldserver` can use it, so that's where this is headed).

To start, I extracted `buildTargets` as a publicly exported function, and then updated all callsites to work on the new types. These should be functionally equivalent.

In the next PR, I'll be looking at moving parts of `trafficSwitcher` struct to `package workflow`

## Related Issue(s)

- #7931 

## Checklist
- [x] Tests were added or are not required (other tests should be covered by other wrangler/vrep tests)

## Impacted Areas in Vitess

- [ ]  Query Serving
- [x]  VReplication
- [ ]  Cluster Management
- [ ]  Build/CI
- [ ]  VTAdmin

### DIFF
--- a/go/vt/vtctl/workflow/traffic_switcher.go
+++ b/go/vt/vtctl/workflow/traffic_switcher.go
@@ -140,18 +140,10 @@ func BuildTargets(ctx context.Context, ts *topo.Server, tmc tmclient.TabletManag
 			return nil, err
 		}
 
-		query := fmt.Sprintf(
-			`SELECT
-				id,
-				source,
-				message,
-				cell,
-				tablet_types
-			FROM
-				_vt.vreplication
-			WHERE
-				workflow=%s AND
-				db_name=%s`, encodeString(workflow), encodeString(primary.DbName()))
+		// NB: changing the whitespace of this query breaks tests for now.
+		// (TODO:@ajm188) extend FakeDBClient to be less whitespace-sensitive on
+		// expected queries.
+		query := fmt.Sprintf("select id, source, message, cell, tablet_types from _vt.vreplication where workflow=%s and db_name=%s", encodeString(workflow), encodeString(primary.DbName()))
 		p3qr, err := tmc.VReplicationExec(ctx, primary.Tablet, query)
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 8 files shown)


ajm188:
Going to set this back to draft while I take time to look into the tests. I thought it was just a query formatting issue, but there might be something else going on.

ajm188:
oooookay that was fun. re-opening!!

rafael (approved):
This is really straightforward from my perspective and makes sense.  I think we should get @rohit-nayak-ps eyes just to make sure we get his input on this approach as well.

ajm188:
Organizational note: I'm planning on squashing this one. The commits are mainly there to break up the diff into reasonably-small chunks but this change really only makes sense logistically as one atomic commit.

XuanThuLe (approved):

ajm188:
I'm going to rebase on master to try to get that `tabletmanager_tablegc` check to actually start (there's not even a button for me to retrigger it from what i can tell)

go/vt/vtctl/workflow/traffic_switcher.go
```diff
@@ -0,0 +1,242 @@
...
+	si        *topo.ShardInfo
+	primary   *topo.TabletInfo
+	Position  string
+	Journaled bool
+}
+
+// NewMigrationSource returns a MigrationSource for the given shard and primary.
+//
+// (TODO|@ajm188): do we always want to start with (position:"", journaled:false)?
+func NewMigrationSource(si *topo.ShardInfo, primary *topo.TabletInfo) *MigrationSource {
```
rohit-nayak-ps, line 69: I see that there is no corresponding NewMigrationTarget()  ...  Is it because it only used in this file?

rohit-nayak-ps (approved):
lgtm
nice set of refactors

go/vt/vtctl/workflow/traffic_switcher.go
ajm188: Yeah, there has to be some temporarily public functions so that I can move things piecemeal (I think this is going to be temporary anyway, and if not I can come back and add the corresponding NewMigrationTarget!)

PR #7941, opened 2021-04-23
Plan StreamExecute Queries

harshit-gangal:
## Description

This is about refactoring the StreamExecute method in executor where the queries received with go through planning phase and then will be executed based on the plan outcome.

## Checklist

## Impacted Areas in Vitess

- [X]  Query Serving


go/vt/vtgate/engine/vstream.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+// GetTableName implements the Primitive interface
+func (v *VStream) GetTableName() string {
+	return v.TableName
+}
+
+// Execute implements the Primitive interface
+func (v *VStream) Execute(_ VCursor, _ map[string]*querypb.BindVariable, wantfields bool) (*sqltypes.Result, error) {
+	return nil, vterrors.New(vtrpcpb.Code_UNIMPLEMENTED, "'Execute' in VStream is not supported")
```
systay, line 65: This should be `vtrpcpb.Code_INTERNAL`, since we'll never implement Execute on vstream, right?

go/vt/vtgate/planbuilder/stream.go
```diff
@@ -0,0 +1,94 @@
+package planbuilder
```
systay, line 1: license header

go/vt/vtgate/planbuilder/stream.go
```diff
@@ -0,0 +1,94 @@
...
+		dest = key.DestinationAllShards{}
+	}
+	var pos string
+	if stmt.Where != nil {
+		pos, err = getVStreamStartPos(stmt)
+		if err != nil {
+			return nil, err
+		}
+	}
+	limit := 100
```
systay, line 52: shouldn't this be a const instead?

go/vt/vtgate/planbuilder/stream.go
```diff
@@ -0,0 +1,94 @@
...
+			if v.Operator == sqlparser.GreaterThanOp {
+				switch c := v.Left.(type) {
+				case *sqlparser.ColName:
+					switch val := v.Right.(type) {
+					case *sqlparser.Literal:
+						pos = string(val.Val)
+					}
+					colName = strings.ToLower(c.Name.String())
+					if colName != "pos" {
+						return "", fmt.Errorf("can only use pos in vstream where clause ")
```
systay, line 83: shouldn't this be a vterror instead?

systay (approved):
LGTM. Only found a couple of minor things

indera-shsp:
Is there an RFC doc for this change?

Issue #7812, opened 2021-04-08
Build docker images with MySQL 8.0.23


aquarapid:
We have not been building the docker images with MySQL > 8.0.21;  because initially xtrabackup would not work with 8.0.22.  This has been resolved in xtrabackup, so we should be able to go ahead and bump the images.

PR #7943, opened 2021-04-23
vitess/lite docker build fails for mysql80

mcronce:
## Description
Currently, the `vitess/lite:mysql80` image is throwing the following error a bunch of times during its build:
```
Reading package lists...
Building dependency tree...
Reading state information...
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 libmysqlclient21 : Depends: mysql-community-client-plugins (= 8.0.23-1debian10) but it is not going to be installed
 mysql-community-client-core : Depends: mysql-community-client-plugins (= 8.0.23-1debian10) but it is not going to be installed
E: Unable to correct problems, you have held broken packages.
```

The build eventually succeeds, but trying to start a cluster with `docker-compose`, the vttablet containers never come up healthy.

It looks like there are dependency issues with those packages.  Downgrading is the first solution that came to mind, although upgrading to 8.0.24 _also_ seems to work fine.  Super open to ideas other than just changing the version.

## Related Issue(s)
#7811 #7812

## Checklist

## Deployment Notes
N/A

## Impacted Areas in Vitess
- [ ]  Query Serving
- [ ]  VReplication
- [ ]  Cluster Management
- [x]  Build/CI
- [ ]  VTAdmin

### DIFF
--- a/docker/lite/install_dependencies.sh
+++ b/docker/lite/install_dependencies.sh
@@ -92,7 +92,7 @@ mysql57)
     )
     ;;
 mysql80)
-    mysql8_version=8.0.21
+    mysql8_version=8.0.24
     do_fetch https://repo.mysql.com/apt/debian/pool/mysql-8.0/m/mysql-community/libmysqlclient21_${mysql8_version}-1debian10_amd64.deb /tmp/libmysqlclient21_${mysql8_version}-1debian10_amd64.deb
     do_fetch https://repo.mysql.com/apt/debian/pool/mysql-8.0/m/mysql-community/mysql-community-client-core_${mysql8_version}-1debian10_amd64.deb /tmp/mysql-community-client-core_${mysql8_version}-1debian10_amd64.deb
     do_fetch https://repo.mysql.com/apt/debian/pool/mysql-8.0/m/mysql-community/mysql-community-client_${mysql8_version}-1debian10_amd64.deb /tmp/mysql-community-client_${mysql8_version}-1debian10_amd64.deb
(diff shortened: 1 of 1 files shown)


aquarapid:
My view is that we should fix the build unless there's a (functionality) problem with the newer version of MySQL.  8.0.21 is getting long in the tooth.  I'll take a look.

mcronce:
@aquarapid going up to 8.0.24 also works

aquarapid:
yeah 8.0.24 is brand new, won't chance it, yet :-)  I've pushed a fix to keep it on 8.0.23

deepthi (approved):
@sougou has verified that this works with a local docker build.

deepthi:
Ideally the docker build should fail and stop on errors. However, because we are doing this in a loop (to tolerate transient errors), we ignore the exit status 
image

```
for i in $(seq 1 $MAX_RETRY); do apt-get install -y --no-install-recommends "${PACKAGES[@]}" && break; done
```

dkhenry:
> Ideally the docker build should fail and stop on errors. However, because we are doing this in a loop (to tolerate transient errors), we ignore the exit status
> image
> 
> ```
> for i in $(seq 1 $MAX_RETRY); do apt-get install -y --no-install-recommends "${PACKAGES[@]}" && break; done
> ```

You can fix that by having a check after the break `if [[ $i -eq  $MAX_RETRY ]]; exit 1; fi`

Issue #7667, opened 2021-03-12
MoveTables fails due to reserved keyword in table column name


tokikanno:
#### Overview of the Issue

MoveTables operation will fail if any moved table has a reserved keyword in its column name.

#### Reproduction Steps

Having a table with the following schema

```sql
CREATE TABLE profile (
    `uid` VARCHAR(32) NOT NULL PRIMARY KEY,
    `rank` VARCHAR(64) NOT NULL DEFAULT 'general',  -- rank is a reversed keyword in MySQL8
    `created` TIMESTAMP DEFAULT 0
) ENGINE=InnoDB CHARACTER SET=utf8mb4 COLLATE=utf8mb4_bin;
```

insert some random rows and execute MoveTables workflow on it

```sh
vtctlclient MoveTables -workflow test-move-profile -tablet_types=master pinkoi default '{"profile": {}}'
```

It will fail while copy insert into the new KeySpace because the auto-generated insert SQL statement didn't quote the reserved keyword in the column name ( `rank` ).

Can also see following slack link for more detail:
https://vitess.slack.com/archives/C0PQY0PTK/p1615454636258200

Ref:
MySQL keyword list
https://dev.mysql.com/doc/refman/8.0/en/keywords.html

#### Binary version

Docker image: `vitess/lite:mysql80  00485232f586` 
(built from Vitess source with git commit `d508d333632e59838310af6d356c0f62a0809002`)

#### Operating system and Environment details

OS, Architecture, and any other information you can provide
about the environment.

- Operating system (output of `cat /etc/os-release`):

```
NAME="Ubuntu"
VERSION="18.04.4 LTS (Bionic Beaver)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 18.04.4 LTS"
VERSION_ID="18.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=bionic
UBUNTU_CODENAME=bionic
```

- Kernel version (output of `uname -sr`):

`Linux 4.15.0-136-generic`

- Architecture (output of `uname -m`):

`x86_64`

#### Log Fragments

```
[...]

PR #7944, opened 2021-04-24
Add rank as reserved keyword

tokikanno:
## Description
Make `rank` as a reserved keyword.

`rank` has become a new reserved keyword in MySQL8, I've seen some PR added it into the [sql.y](https://github.com/vitessio/vitess/blob/482a7008117ee3215663aeb33cad981e5242a88a/go/vt/sqlparser/sql.y#L259)  and [sql.go](https://github.com/vitessio/vitess/blob/482a7008117ee3215663aeb33cad981e5242a88a/go/vt/sqlparser/sql.go#L402), but not in the `keywords.go`.

If you're running the move table process and having `rank` in your table column name, it will cause the move table process to fail due to the generated insert statement doesn't quote the rank keyword properly.

## Related Issue(s)
- https://github.com/vitessio/vitess/issues/7667

## Checklist

Can't be very sure about the above checklist item, how many tests do I need to add? (or I don't?)

## Impacted Areas in Vitess

- [x]  Query Serving
- [x]  VReplication
- [ ]  Cluster Management
- [ ]  Build/CI
- [ ]  VTAdmin


GuptaManan100 (approved):

systay:
Hi @tokikanno!

Welcome as contributor! ❤️ 

Seems like we have added new CI tests that are not on your branch. Would you mind rebasing or merging master into your branch and pushing? All CI tests have to pass, and that won't happen since this old branch doesn't have all required tests yet.

tokikanno:
@systay : sure, branch synced.

tokikanno:
@systay : I've pushed the newly merged branch contents, but the CI flow seems wasn't triggered. Do I need to do anything special for triggering the CI flow? Or this can only be manually triggered?

harshit-gangal (approved):

PR #7945, opened 2021-04-25
ddl_strategy: -singleton scope is now migration context

shlomi-noach:
## Description

Followup to #7785 

`-singleton` now runs in migration-request context. That is, a `-singleton` migration is rejected if there's a pending migration from another context. It's fine to submit multiple `-singleton` migrations within the same context.

e.g. the following submittion of 3 migration is valid: 
```
vtctlclient ApplySchema -ddl_strategy="online -singleton" -sql "drop table if exists t1; drop table if exists t2; drop table if exists t3;" commerce
```

If you submit multiple `-singleton`  migrations within the same context, either all are accepted or all are rejected.

I believe there is yet a race condition to tackle here, looking into.

## Related Issue(s)

- #7785 
- #6926  

## Checklist

## Impacted Areas in Vitess

- [x]  Query Serving
- [ ]  VReplication
- [ ]  Cluster Management
- [ ]  Build/CI
- [ ]  VTAdmin

### DIFF
--- a/go/vt/schema/ddl_strategy.go
+++ b/go/vt/schema/ddl_strategy.go
@@ -28,9 +28,10 @@ var (
 )
 
 const (
-	declarativeFlag = "declarative"
-	skipTopoFlag    = "skip-topo"
-	singletonFlag   = "singleton"
+	declarativeFlag      = "declarative"
+	skipTopoFlag         = "skip-topo"
+	singletonFlag        = "singleton"
+	singletonContextFlag = "singleton-context"
 )
 
 // DDLStrategy suggests how an ALTER TABLE should run (e.g. "direct", "online", "gh-ost" or "pt-osc")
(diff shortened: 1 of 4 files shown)


shlomi-noach:
> I believe there is yet a race condition to tackle here, looking into.

There's no race condition running via `vtctl ApplySchema`.

When running from `vtgate`, then the user may submit migrations at different times, which means every migration gets evaluated at time of submission. So, say the user tries to submit one migration and gets rejected (because some other migration is pending). 2 minutes later, the user tries a 2nd, different submission, and gets successful (because that previous migration was completed). This is a non-atomic situation.

I think it's fair to say `-singleton` makes sense in a `vtctl` setup, and does not make sense in `vtgate` setup.

Thinking about this a little bit more.

shlomi-noach:
Replaced by https://github.com/vitessio/vitess/pull/7946

PR #7946, opened 2021-04-25
Online DDL: introducing ddl_strategy `-singleton-context` flag

shlomi-noach:
## Description

Followup to #7785

This PR introduces `ddl_strategy` now supports `-singleton-context` flag.

- `-singleton` rejects a migration if there's any other pending migration
- `-singleton-context` rejects a migration is there's any other pending migration that has a different migration context.
With `-singleton-context`, it is possible to submit multiple migrations which all share the same migration context. For example, the following is allowed:
```
vtctlclient ApplySchema -ddl_strategy="online -singleton-context" -sql "drop table if exists t1; drop table if exists t2; drop table if exists t3;" commerce
```

The above submits three migrations. All three will be accepted. For comparison, running the above with `-singleton` would accept the first migration (`drop table if exists t1`) and reject the other two.

## Related Issue(s)

- #7785 
- #6926  

## Checklist

## Impacted Areas in Vitess

- [x]  Query Serving
- [ ]  VReplication
- [ ]  Cluster Management
- [ ]  Build/CI
- [ ]  VTAdmin

### DIFF
--- a/go/vt/vttablet/onlineddl/executor.go
+++ b/go/vt/vttablet/onlineddl/executor.go
@@ -1167,16 +1167,17 @@ func (e *Executor) readMigration(ctx context.Context, uuid string) (onlineDDL *s
 		return nil, nil, ErrMigrationNotFound
 	}
 	onlineDDL = &schema.OnlineDDL{
-		Keyspace:    row["keyspace"].ToString(),
-		Table:       row["mysql_table"].ToString(),
-		Schema:      row["mysql_schema"].ToString(),
-		SQL:         row["migration_statement"].ToString(),
-		UUID:        row["migration_uuid"].ToString(),
-		Strategy:    schema.DDLStrategy(row["strategy"].ToString()),
-		Options:     row["options"].ToString(),
-		Status:      schema.OnlineDDLStatus(row["migration_status"].ToString()),
-		Retries:     row.AsInt64("retries", 0),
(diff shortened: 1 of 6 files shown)


shlomi-noach:
tests added and looking good.

go/vt/vttablet/onlineddl/executor.go
```diff
@@ -2543,16 +2544,36 @@ func (e *Executor) SubmitMigration(
...
-			return result, err
+			return nil, err
 		}
-		if len(uuids) > 0 {
-			return result, fmt.Errorf("singleton migration rejected: found pending migrations [%s]", strings.Join(uuids, ", "))
+		switch {
+		case onlineDDL.StrategySetting().IsSingleton():
+			// We will reject this migration if there's any pending migration
+			if len(pendingUUIDs) > 0 {
+				return result, fmt.Errorf("singleton migration rejected: found pending migrations [%s]", strings.Join(pendingUUIDs, ", "))
```
systay, line 2564: `vterror` should be used, right?

systay (approved):
LGTM

go/vt/vttablet/onlineddl/executor.go
shlomi-noach: Right :stuck_out_tongue:

shlomi-noach: fixed

Issue #7632, opened 2021-03-07
VReplication: evaluate "progress" and "ETA" for online DDL


shlomi-noach:
#### Feature Description

We have a request for VReplication/OnlineDDL to provide similar functionality found in `gh-ost`:

- be able to evaluate progress percent (float64, 0%-100%)
- be able to estimate time till completion (int64, units in seconds)

Both metrics are to evaluate based on the `VCopier` progress, and ignoring `VPlayer` work, based on the heuristic that binlog events are generally applied while migration is running, and that the majority of time spent is in copying existing data.

#### Use Case(s)

Visibility to the user

cc @rohit-nayak-ps FYI

rohit-nayak-ps:
We have had requests for ETA for MoveTables/Reshard vreplication workflows as well. 

The V2 VReplication Workflows have a `progress` command (https://vitess.io/docs/reference/vreplication/v2/progress/) that shows copy progress as a percentage and aggregate, using the number of rows and table size from `information_schema.columns`.  This is run on-demand when user runs a vtctl command. We don't store the duration of the copy or a snapshot of ongoing progress, at the moment, which we will need for an eta.

We have discussed maintaining a separate vreplication workflows store  to maintain a top-level view as well as workflow metadata.  We can also record progress and eta there.

shlomi-noach:
@rohit-nayak-ps can we store progress and ETA in `_vt.vreplication`, in same transaction that updates `pos`?

rohit-nayak-ps:
Fixed by #8015

PR #7950, opened 2021-04-26
VReplication: populate rows_estimate and rows_copied

shlomi-noach:
# Work in Progress

## Description

This PR adds th efollowing columns to `_vt.vreplication`:

- `rows_estimate`
- `rows_copied`
- `time_started`
- `eta_seconds`

The purpose is for a VReplication stream (not a complete workflow) to be able to track its own progress, and evaluate % complete as well as ETA.
 
## Related Issue(s)

- https://github.com/vitessio/vitess/issues/7632

## Checklist

## Impacted Areas in Vitess

- [ ]  Query Serving
- [x]  VReplication
- [ ]  Cluster Management
- [ ]  Build/CI
- [ ]  VTAdmin

cc @rohit-nayak-ps


shlomi-noach:
@rohit-nayak-ps while this PR is in `Draft`, can you give an initial review to see I'm on the correct path?

One thing I notice (testing with Online DDL, thought this is pure VReplication logic) is that sometimes I get `rows_estimate=0` in `vreplication` table. This is weird to me, since the value comes from `information_schema.tables.table_rows`. I can verify that `VStreamRowsResponse` has the value `0` for `TableRows` which means the value `0` comes from the streamer. Right now I'm unsure what makes it so.

shlomi-noach:
`table_rows=0` tracked all the way down to MySQL itself, so nothing to look for in Vitess, it merely reflects MySQL.

shlomi-noach:
`table_rows=0` tracked all the way down to MySQL itself, so nothing to look for in Vitess, it merely reflects MySQL.

shlomi-noach:
Another question: I notice `tabletserver/engine`  actually querying **all** keyspace tables, even though it only needs to read from a single table. Is this intentional?

go/vt/vttablet/tabletserver/vstreamer/vstreamer.go
```diff
@@ -80,9 +80,6 @@ type vstreamer struct {
 	vse   *Engine
 }
 
-// CopyState contains the last PK for tables to be copied
-type CopyState map[string][]*sqltypes.Result
-
```
shlomi-noach, line 85: I just found that `CopyState` was never used anywhere. Removed it as cleanup.

shlomi-noach:
Replaced by https://github.com/vitessio/vitess/pull/7963

PR #7961, opened 2021-04-27
OnlineDDL: more migration check ticks upon migration start

shlomi-noach:
## Description

Online DDL: tablets do periodic checks to see whether new migrations were submitted, and also, in the case of VReplication, they do periodic checks to see if a migration is completed. The normal check interval is `1 minute`, and there's already some boosting whenever we heuristically expect more work to come.

This PR adds a bit more boost: whenever new migration_ starts_, we schedule more _ticks_ tun run within the next X seconds. These help us with very quick migrations, in particular VReplication, which runs independently and asycnhronously.

Meaning, we will detect sonner, rather than later, that a VReplication migration is complete, and will move on, sooner rather than later, to the next migraiton.

## Related Issue(s)

- Tracking: #6926 

## Checklist


shlomi-noach:
Request for review :pray:

systay (approved):

PR #7963, opened 2021-04-27
VReplication: tracking `rows_copied`

shlomi-noach:
## Description

`_vt.vreplication` has new column, `rows_copied`, which is updated periodically during `vreplication` work, to reflect the actual number of rows written on the target table.

The value is written together with `time_updated`, so it does not get transactionally synced with each copy iteration, it follows closely though, and makes a good estimate.

This will be later used by Online DDL to track Vreplication-based migration progress & ETA. 

## Related Issue(s)

- Replaces https://github.com/vitessio/vitess/pull/7950
- Tracking: #6926 

## Checklist

### DIFF
--- a/go/vt/binlog/binlogplayer/binlog_player.go
+++ b/go/vt/binlog/binlogplayer/binlog_player.go
@@ -476,7 +476,7 @@ func (blp *BinlogPlayer) writeRecoveryPosition(tx *binlogdatapb.BinlogTransactio
 	}
 
 	now := time.Now().Unix()
-	updateRecovery := GenerateUpdatePos(blp.uid, position, now, tx.EventToken.Timestamp)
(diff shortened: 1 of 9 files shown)


shlomi-noach:
@rohit-nayak-ps I could use some assistance in understanding all the test failure here. I'm very confused. Almost everything that has to do with VReplication fails.

I started by looking into the failing unit tests, failing on `expectDBClientQueries`, which made some sense because I added `withDDL` to multiple queries. But it still does not make sense.

I then turned over to various other tests, and all report timeouts or switch errors.

The addition of `withDDL` is absolutely required, code fails to run if I try upgrade from `v10` without `withDDL`, due to the added column `table_rows`. But otherwise I did not change logic, per se. Just added that column, and wrapped in `withDDL`.

BTW in a  local environment, Online DDL via vreplication works just fine.

shlomi-noach:
Replaced by https://github.com/vitessio/vitess/pull/7980

PR #7965, opened 2021-04-27
[wrangler|topotools] Migrate `UpdateShardRecords`, `RefreshTabletsByShard`, and `{Get,Save}RoutingRules`

ajm188:
## Description

I am moving these to `package topotools` to reuse this code in both
`package grpcvtctldserver` (which wrangler imports) and `package
wrangler`.

## Related Issue(s)

- #7931 

## Checklist

### DIFF
--- a/go/vt/topotools/keyspace.go
+++ b/go/vt/topotools/keyspace.go
@@ -1,0 +1,119 @@
+/*
+Copyright 2021 The Vitess Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package topotools
+
+import (
+	"context"
+	"sync"
+	"time"
+
+	"vitess.io/vitess/go/vt/logutil"
+	"vitess.io/vitess/go/vt/topo"
+	"vitess.io/vitess/go/vt/vttablet/tmclient"
+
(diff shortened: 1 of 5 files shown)


go/vt/topotools/keyspace.go
```diff
@@ -0,0 +1,119 @@
...
+		if err != nil {
+			return err
+		}
+
+		shards[i] = updatedShard
+
+		// For 'to' shards, refresh to make them serve. The 'from' shards will
+		// be refreshed after traffic has migrated.
+		if !isFrom {
+			_ = RefreshTabletsByShard(ctx, ts, tmc, si, cells, logger)
```
rafael, line 114: I like this pattern of being explicit that we are ignoring the error. However, is not a pattern that I've seen in the codebase.  

I wonder if start adding here will create confusion.

rafael (commented):
This looks good. Left a small comment. Let me know what you think.

go/vt/topotools/keyspace.go
ajm188: My opinion is being that explicit about ignoring the error is a better (or at least more clear) pattern, and we've got to start somewhere! 😂  I'll do a quick temp check to see what some other folks think about this

rafael: That sounds like a plan. If we can get thumbs-up that we would like to make this a standard pattern moving forward, let's start here!

ajm188: Consensus seemed to be:
* yes, explicitly acknowledging that we're ignoring errors is better than implicitly doing so
* we've had lots of trouble debugging tests specifically due to errors being discarded
* we should log the ignored error in this specific case, and whether to log or not is probably a case-by-case basis
* i'm also going to work on [adding `errcheck` to the linter](https://github.com/vitessio/vitess/pull/7996#issuecomment-829241274) so we can start to enforce this as a pattern going forward

rafael (approved):
LGTM

go/vt/topotools/keyspace.go
rafael: Great, thanks for following up on this. All makes sense.

Issue #3984, opened 2018-05-27
RBR replication works incorrectly for 'BINARY' type


sougou:
#### Overview of the Issue

MySQL pads `binary` columns with zeroes. Here's an example script that shows the problem:

```
mysql> create table d(id int, val binary(5), primary key(id));
Query OK, 0 rows affected (0.02 sec)

mysql> insert into d values(1, 'a');
Query OK, 1 row affected (0.01 sec)

mysql> select * from d where val='a';
Empty set (0.00 sec)

mysql> select * from d where val='a\0\0\0\0';
+----+-------+
| id | val   |
+----+-------+
|  1 | a     |
+----+-------+
1 row in set (0.00 sec)
```

When a row like this goes through our binlog streamer, DMLs get generated as `val='a'`, and that does not match the target row. This fails filtered replication.

#### Reproduction Steps

Create a table like the above, start running a split clone, and perform a DML that changes any row on that table. The filtered replication will fail to affect the target row.

sougou:
@alainjobart Let's discuss this when you have the time.

sougou:
Did some binlog mining:
```
mysql> create table sbt(id int, val char(5), primary key(id));
Query OK, 0 rows affected (0.03 sec)

mysql> insert into sbt values(1, 'a');
Query OK, 1 row affected (0.02 sec)

mysql> update sbt set val='b' where id = 1;
Query OK, 1 row affected (0.01 sec)
Rows matched: 1  Changed: 1  Warnings: 0

mysql> update sbt set val='cc' where id = 1;
Query OK, 1 row affected (0.02 sec)
Rows matched: 1  Changed: 1  Warnings: 0
```

Char codes:
```
'a' =0x61
'b'=0x62
'c'=0x63
```

For first update ('a'->'b'):
```
mysql.binlogEvent{0xaa, 0x2f, 0xa, 0x5b, 0x1f, 0x96, 0xe9, 0xf5, 0x40, 0x32, 0x0, 0x0, 0x0, 0x48, 0xf, 0x0, 0x0, 0x0, 0x0, 0x72, 0x0, 0x0, 0x0, 0x0, 0x0, 0x1, 0x0, 0x2, 0x0, 0x2, 0xff, 0xff, 0xfc, 0x1, 0x0, 0x0, 0x0, 0x1, 0x61, 0xfc, 0x1, 0x0, 0x0, 0x0, 0x1, 0x62, 0xe4, 0xbf, 0x63, 0x5a}
```
[...]

PR #7969, opened 2021-04-27
VReplication: Pad binlog values for binary() columns to match the value returned by mysql selects

rohit-nayak-ps:

## Description
For fixed length binary columns, mysql internally pads the values on the right will nulls. However the binlogs contain the original value. This leads to a problem in vreplication workflows if a binary column is used as a sharding key.

During the copy phase the row is read using a mysql select and the column has the padded value. However while replicating we get the unpadded value. This means that we get different keyspace_ids. We found bugs where the copy phase inserted the row in one shard but the update during the catchup phase routed this row to another shard where, of course, the update failed.

The PR pads the value found in the binlog to match the value returned my mysql to fix this issue.

## Related Issue(s)

https://github.com/vitessio/vitess/issues/3984

## Checklist

### DIFF
--- a/go/mysql/binlog_event_rbr.go
+++ b/go/mysql/binlog_event_rbr.go
@@ -895,6 +895,18 @@ func CellValue(data []byte, pos int, typ byte, metadata uint16, styp querypb.Typ
 		l := int(data[pos])
 		mdata := data[pos+1 : pos+1+l]
 		if sqltypes.IsBinary(styp) {
+			// For binary(n) column types, mysql pads the data on the right with nulls. However the binlog event contains
+			// the data without this padding. This causes several issues:
+			//    * if a binary(n) column is part of the sharding key, the keyspace_id() returned during the copy phase
+			//      (where the value is the result of a mysql query) is different from the one during replication
+			//      (where the value is the one from the binlogs)
+			//    * mysql where clause comparisons do not do the right thing without padding
+			// So for fixed length binary() columns we right-pad it with nulls if necessary
+			if l < max {
+				paddedData := make([]byte, max)
+				copy(paddedData[:l], mdata)
+				mdata = paddedData
+			}
 			return sqltypes.MakeTrusted(querypb.Type_BINARY, mdata), l + 1, nil
 		}
 		return sqltypes.MakeTrusted(querypb.Type_VARCHAR, mdata), l + 1, nil
--- a/go/test/endtoend/vreplication/config.go
+++ b/go/test/endtoend/vreplication/config.go
@@ -10,6 +10,7 @@ create table orders(oid int, cid int, pid int, mname varchar(128), price int, pr
 create table order_seq(id int, next_id bigint, cache bigint, primary key(id)) comment 'vitess_sequence';
 create table customer2(cid int, name varbinary(128), typ enum('individual','soho','enterprise'), sport set('football','cricket','baseball'),ts timestamp not null default current_timestamp, primary key(cid));
 create table customer_seq2(id int, next_id bigint, cache bigint, primary key(id)) comment 'vitess_sequence';
+create table tenant(tenant_id binary(16), name varbinary(16), primary key (tenant_id));
 `
 
 	initialProductVSchema = `
@@ -28,7 +29,8 @@ create table customer_seq2(id int, next_id bigint, cache bigint, primary key(id)
 	},
 	"order_seq": {
 		"type": "sequence"
-	}
+	},
+	"tenant": {}
   }
 }
 `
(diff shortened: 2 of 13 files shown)


go/vt/vttablet/tabletserver/vstreamer/vstreamer.go
```diff
@@ -874,7 +874,9 @@ func (vs *vstreamer) extractRowAndFilter(plan *streamerPlan, data []byte, dataCo
 			return false, nil, err
 		}
 		pos += l
+
```
deepthi, line 877: nit: why the extra whitespace?

deepthi (approved):
LGTM

go/vt/vttablet/tabletserver/vstreamer/planbuilder.go
```diff
@@ -157,10 +157,22 @@ func (plan *Plan) filter(values []sqltypes.Value) (bool, []sqltypes.Value, error
...
+func getKeyspaceID(values []sqltypes.Value, vindex vindexes.Vindex, vindexColumns []int, fields []*querypb.Field) (key.DestinationKeyspaceID, error) {
 	vindexValues := make([]sqltypes.Value, 0, len(vindexColumns))
-	for _, col := range vindexColumns {
-		vindexValues = append(vindexValues, values[col])
+	for colNum, col := range vindexColumns {
+		// For binary(n) column types, mysql pads the data on the right with nulls. However the binlog event contains
+		// the data without this padding. In particular, this causes an issue if a binary(n) column is part of the
+		// sharding key: the keyspace_id() returned during the copy phase (where the value is the result of a mysql query)
+		// is different from the one during replication (where the value is the one from the binlogs)
+		// Hence we need to add the padding here
```
rafael, line 167: What a catch this one!!

rohit-nayak-ps:
Just reviewed with Sugu and he recommended that the padding should be done right when it is read from the binlog and not just for computing the keyspace id. That way the padded value, which is also the result of a mysql select, will be sent to all consumers of the vstream. 

So moving it to draft for now.

rohit-nayak-ps:
@sougou I was able to pad the binary columns in the binlog reader. One benefit with this fix is that we could remove the cast we had added earlier to cast binary pks in sql which should improve workflow performance too.

@deepthi, please re-review

go/mysql/binlog_event_rbr.go
```diff
@@ -895,6 +895,18 @@ func CellValue(data []byte, pos int, typ byte, metadata uint16, styp querypb.Typ
...
+			//    * if a binary(n) column is part of the sharding key, the keyspace_id() returned during the copy phase
+			//      (where the value is the result of a mysql query) is different from the one during replication
+			//      (where the value is the one from the binlogs)
+			//    * mysql where clause comparisons do not do the right thing without padding
+			// So for fixed length binary() columns we right-pad it with nulls if necessary
+			if l < max {
+				paddedData := make([]byte, max)
+				copy(paddedData[:l], mdata)
+				mdata = paddedData
+			}
```
shlomi-noach, line 909: :+1:

shlomi-noach (approved):
It's painful to see how small the change is :)

sougou (approved):

sougou (commented):
I feel like a similar bug may exist with CHAR(N), where we may need to pad with spaces.

go/test/endtoend/vreplication/config.go
```diff
@@ -39,9 +41,12 @@ create table customer_seq2(id int, next_id bigint, cache bigint, primary key(id)
...
 	    "reverse_bits": {
 	      "type": "reverse_bits"
-	    }
+	    },
+		"binary_md5": {
+          "type": "binary_md5"
+		}
 	  },
-   "tables": {
+   "tables":  {
```
rafael, line 49: nit nit - extra whitespace.

rafael (approved):
LGTM

deepthi (approved):
LGTM

deepthi:
Let us plan to backport this to 10.0, 9.0 and 8.0.
We also need to document it as a "Known Issue" in the 10.0 release notes.

shlomi-noach:
Also, just realized there was an almost identical recent contribution to `gh-ost`. I should be more wary.

go/mysql/binlog_event_rbr.go
mattlord: There's a bug here for CHAR columns using binary collations. If you have e.g. `CHAR (3) COLLATE UTF8MB4_BIN` then the padding goes out to the max **byte length** of 12 and you get this in the vreplication stream: 
`FOO\0\0\0\0\0\0\0\0\0` and the SQL fails because it's too long for the column.

Issue #7942, opened 2021-04-23
vttestserver: default charset


deepthi:
#### Feature Description
vttestserver defaults the charset to utf8. Either we should change this to utf8mb4 or provide a way to override it on the docker container (or both).
https://github.com/vitessio/vitess/blob/master/go/vt/vttest/vtprocess.go#L179

#### Use Case(s)
Frameworks set charset. This is required to automate testing.

harshit-gangal:
I am of the opinion that we make it as default on vttablet.

deepthi:
Fixed in #7970

PR #7970, opened 2021-04-27
docker/vttestserver/run.sh:  Add $CHARSET environment variable

mcronce:
## Description
This adds an environment variable to pass through to vttestserver's `-charset` command-line flag, and defaults it to `utf8mb4`.  This, among other things, enables a portion Prisma's integration test suite that validates certain Unicode support.

## Related Issue(s)
#7942

## Checklist

## Deployment Notes
This only changes the vttestserver containers.  @harshit-gangal mentioned in #7942 making it the default in vttablet.  This PR does _not_ do that, so I don't disagree with leaving that issue open if we consider that a requirement.

### DIFF
--- a/docker/vttestserver/run.sh
+++ b/docker/vttestserver/run.sh
@@ -17,4 +17,4 @@
 # Setup the Vschema Folder
 /vt/setup_vschema_folder.sh "$KEYSPACES" "$NUM_SHARDS"
 # Run the vttestserver binary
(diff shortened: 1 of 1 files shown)


deepthi:
> This only changes the vttestserver containers. @harshit-gangal mentioned in #7942 making it the default in vttablet. This PR does not do that, so I don't disagree with leaving that issue open if we consider that a requirement.

Let's leave it open. We can do a separate PR for the default change (once we decide it is safe to do so).

mcronce:
@deepthi SGTM :)

deepthi (approved):

PR #7976, opened 2021-04-28
docker/lite/install_dependencies.sh:  If the dependency install loop reaches its max number of retries, consider that a failure; exit the script nonzero so the build halts.

mcronce:
## Description
In #7943, we fixed the `vitess/lite:mysql80` build; however, we observed that a failure to install dependencies was still "successfully" building a container image.

This fixes that - if installing dependencies fails enough times to exhaust the retry loop, we will now exit nonzero.

Tested manually by reverting #7943 in my local copy and attempting a build.  It failed, as expected.

## Related Issue(s)
* #7943

## Checklist


deepthi (approved):
:+1:

Issue #7620, opened 2021-03-05
Reshard Workflow VSchema validation


rafael:
#### Feature Description

Sometimes it is possible to have tables that are not part of the VSchema in VTtablets. Most notably when doing schema changes with tools like ghost. In our environment these tables are kept for sometime in case we need to run some audits or we have to rollback a change. 

Currently, Reshard workflows don't validate if all the tables in the tablet are part of the VSchema. We should add a validation for this. If a table is in the tablet but not in the VSchema, the workflow can't be created. 

At the moment, 
#### Use Case(s)

The use case here is to provide a better experience to Reshard workflows. Let the user early in the process that it needs to do some cleanup, otherwise the overall process will fail.

rohit-nayak-ps:
Fixed by @makinje16  in https://github.com/vitessio/vitess/pull/7977

PR #7977, opened 2021-04-28
VSchema Validation on ReshardWorkflow Creation

makinje16:

## Description

This PR adds in vschema validation during creation of a ReshardWorkflow. If any of the source shards have tables in their `_vt` database that are not in the vschema then the creation of the ReshardWorkflow will fail. We will then return which shards failed their vschema checks and what tables caused their respective failures.
## Related Issue(s)

#7620 

## Checklist

### DIFF
--- a/go/vt/wrangler/schema.go
+++ b/go/vt/wrangler/schema.go
@@ -143,7 +143,7 @@ func (wr *Wrangler) diffSchema(ctx context.Context, masterSchema *tabletmanagerd
 }
 
 // ValidateSchemaShard will diff the schema from all the tablets in the shard.
-func (wr *Wrangler) ValidateSchemaShard(ctx context.Context, keyspace, shard string, excludeTables []string, includeViews bool) error {
+func (wr *Wrangler) ValidateSchemaShard(ctx context.Context, keyspace, shard string, excludeTables []string, includeViews bool, includeVSchema bool) error {
 	si, err := wr.ts.GetShard(ctx, keyspace, shard)
 	if err != nil {
 		return fmt.Errorf("GetShard(%v, %v) failed: %v", keyspace, shard, err)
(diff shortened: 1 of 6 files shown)


rafael (approved):
Great work Malcolm! 

This is looking good. One thing that I was discussing with @rohit-nayak-ps is that we should make sure to ignore ghost schema changes from these validations. 

However, there are a few other places we need to update before fully adopting those conventions (e.g VDiff should ignore tables that are related to ghost). 

We can work on that as separate PR.

deepthi:
> Great work Malcolm!
> 
> This is looking good. One thing that I was discussing with @rohit-nayak-ps is that we should make sure to ignore ghost schema changes from these validations.
> 
> However, there are a few other places we need to update before fully adopting those conventions (e.g VDiff should ignore tables that are related to ghost).
> 
> We can work on that as separate PR.

Does this mean that in the presence of gh-ost artifact tables workflow creation will always fail (with the changes in this PR)?

rafael:
> Does this mean that in the presence of gh-ost artifact tables workflow creation will always fail (with the changes in this PR)?
@deepthi  that is correct. But I think that is better right now than the existent behavior. Today, we will create the resharding workflow, but then VDiff will consistently fail because it doesn't know how to handle this case yet.

rohit-nayak-ps (approved):
lgtm

go/vt/wrangler/schema.go
```diff
@@ -199,7 +217,7 @@ func (wr *Wrangler) ValidateSchemaKeyspace(ctx context.Context, keyspace string,
 	}
 	sort.Strings(shards)
 	if len(shards) == 1 {
-		return wr.ValidateSchemaShard(ctx, keyspace, shards[0], excludeTables, includeViews)
+		return wr.ValidateSchemaShard(ctx, keyspace, shards[0], excludeTables, includeViews, false /*includeVSchema*/)
```
ajm188, line 220: As a follow-up, I think it would be great to add the VSchema validation to `ValidateSchemaKeyspace` as well

ajm188 (approved):
LGTM. One comment that I think can be done as a follow-up change.

Issue #7802, opened 2021-04-07
Backup/restore performance issues


aquarapid:
Writing up a general/container issue, to enumerate the problem areas:

## Intro

For backup/restore, the preferred solution in Vitess today is to use the xtrabackup integration.  In general, the way this works is:

### Backup
 * vttablet launches xtrabackup against the local MySQL instance
 * xtrabackup streams the backup output to vttablet (potentially multiple streams)
 * vttablet compresses the stream(s)
 * vttablet then directs the stream(s) to a data store (upload to S3, upload to GCS, store to local file, send to Ceph, etc.)

### Restore
  * vttablet reads the backup data from a data store
  * vttablet decompresses the stream
  * vttablet streams the decompressed data into xtrabackup, which restores the database in a temporary directory locally
  * xtrabackup then performs a "prepare" on the backup (replaying the logged changes that accumulated during the backup and was stored along with the MySQL data files).  Depending on how long the backup took, how large it is, and how many writes the database took;  this can take a significant amount of time
  * When the prepare is done, the files are moved into position, and MySQL can be started on top of it, completing the restore.

Note that for purposes of this discussion, I am going to ignore any point-in-time restore by applying binlogs after the above process is complete.

## Performance bottlenecks

 * General:  Because much of the streaming/compression/decompression work happens in vttablet, which also might be serving queries;  this can be a problem, both from a CPU and GC point of view.  We have no limiter/throttler in the pipeline to limit impact on the serving tablet (only really a problem for backups;  restores obviously should go as fast as possible)
 * Backup:  compression phase.  This has been problematic (from a correctness point of view) in the past, we have used a few parallel gzip/zlib libraries.  There is also not much knobs for the user to tune to make tradeoffs (e.g. compression level).  One specific issue here is that with the parallel compression and multiple threads, it is easy to just blow out the local host/container CPU allocation completely (maybe just something to document?)
 * Backup:  decompression phase.  zlib/gzip decompression is single-threaded, and not very fast (at least with the library we use), so we are limited to a maximum decompression speed around 100-200 MB/sec (depending on the data).  For small tablets, this isn't a big deal, but we can do better.  By definition, this is probably the most performance critical stage.
   * In general, it might be advisable to move to zstd compression, it compresses much faster (or with less CPU), supports multi-threaded compression out of the box, and it's single thread decompression is easily 4x as fast as our current gzip/zlib library.  As a bonus, depending on the compression level you use, it can compress significantly better than gzip/zlib.
[...]

PR #7978, opened 2021-04-28
Support for additional compressors/decompressors

rvrangel:
Signed-off-by: Renan Rangel renan@slack-corp.com

## Description

This change will allow us to use an external decompressor when available as opposed to the internal `pgzip` currently used. We ran some tests using `pigz -d -c` and we get a nice speed bump on decompression, around 30% or more faster decompressing when restoring from a backup in S3. While decompressing is still single-threaded in `pigz`, it uses separate threads for reading, writing and checksum, and it results in faster performance. 

I wanted to get an initial feedback on this change, as we would like to also to plug our own compressor to use other encryption algorithms (like `lz4` or `zstd`) with an external binary as this is a more flexible setup for users, but also work on the change to add built-in support for one of these compression algorithms.

## Related Issue(s)

https://github.com/vitessio/vitess/issues/7802

## Checklist

## Deployment Notes

This should not have any affect on current deployments. To take advantage of this, one needs to pass the correct flags to `vttablet` and the command also needs to exist in the OS, otherwise it will revert to the builtin decompressor.


rafael (commented):
Let's get @deepthi input on this.

deepthi:
Sorry for the delay in responding. I somehow missed the notification for reviewing this PR.
I am not opposed to introducing more options for decompressing the backup file. However, it will be good to make the UX slightly different from what is proposed in this PR.
We can assume that we will have two classes of decompressors
1. Pure golang which can be plugged in directly. We currently use pgzip but there is a proposal to either replace it with zstd or provide zstd as an alternative.
2. External binaries for which we need a path and we use command.Exec to run them.

To start with we only need one flag - let us call it `xtrabackup_external_decompressor`. If that is specified, we use it, otherwise we use the builtin. In future if we decide to support multiple `builtin` flavors, we can add another flag.
Some thought should be given to whether to expect the binary to exist on the path, or to expect the flag value to contain the full path.

deepthi:
Things have changed a little bit now (see detail in #8175), so the design will need to change to accommodate a choice of builtin decompressors also.

rvrangel:
@deepthi I saw that, I will update this PR to take it into account, thanks!

rvrangel:
hi @deepthi, I have update the code with support for the two classes of compressors and decompressors:

- external compressor/decompressor:
  - as you mentioned, I updated it to use `xtrabackup_external_decompressor` and also added two: `xtrabackup_external_compressor` and `xtrabackup_external_compressor_extension` to control which command to use for external compressor (along with the file extension to use)
- builtin compressors/decompressors:
  - users should be able to use `xtrabackup_builtin_compressor` to choose a different compressor (defaults to `pgzip`) and they can also specify a decompressor with `xtrabackup_builtin_decompressor` but by default it will try to figure out which decompressor to use if not specified 
  - I added support for `lz4` and `zstd`, both in pure Go. I noticed there is a [cgo version](https://github.com/DataDog/zstd) of the `zstd` library which was faster in my tests, we can also look into adding support for it.

deepthi:
> hi @deepthi, I have update the code with support for the two classes of compressors and decompressors:
> 
> * external compressor/decompressor:
>   
>   * as you mentioned, I updated it to use `xtrabackup_external_decompressor` and also added two: `xtrabackup_external_compressor` and `xtrabackup_external_compressor_extension` to control which command to use for external compressor (along with the file extension to use)
> * builtin compressors/decompressors:
>   
>   * users should be able to use `xtrabackup_builtin_compressor` to choose a different compressor (defaults to `pgzip`) and they can also specify a decompressor with `xtrabackup_builtin_decompressor` but by default it will try to figure out which decompressor to use if not specified
>   * I added support for `lz4` and `zstd`, both in pure Go. I noticed there is a [cgo version](https://github.com/DataDog/zstd) of the `zstd` library which was faster in my tests, we can also look into adding support for it.

This sounds good. Let us _not_ add cgo right now, we want to avoid that if possible.
LMK when the PR is ready to review, and at that time I can also run the CI for you.

rvrangel:
@deepthi  It should be ready for review now 👍, I would love to get your feedback on it

go/vt/mysqlctl/compression.go
```diff
@@ -0,0 +1,254 @@
...
+)
+
+func getEngineFromExtension(extension string) (string, error) {
+	for ext, eng := range engineExtensions {
+		if ext == extension {
+			return eng[0], nil // we select the first supported engine in auto mode
+		}
+	}
+
+	return "", fmt.Errorf("%w \"%s\"", errUnsupportedCompressionExtension, extension)
```
vmg, line 44: Anywhere where you have `\"%s\"` in a format string, you can replace it with `%q`, which adds the quotes automatically _and_ escapes the contents of the string so they always look proper inside the quotes. :ok_hand:

vmg:
Hiiii @rvrangel! Been looking at this PR. I'm afraid that your usage of `Exec` and pipes in general is not correct here and may lead to corruption.

From the `exec` package docs:

>  StdoutPipe returns a pipe that will be connected to the command's standard output when the command starts.
>
> Wait will close the pipe after seeing the command exit, so most callers need not close the pipe themselves. It is thus incorrect to call Wait before all reads from the pipe have completed. For the same reason, it is incorrect to call Run when using StdoutPipe.

You can see that this API in the stdlib is very nuanced. `Wait` will close the pipe as soon as the program has exited, and if we haven't read everything that has been buffered so far, we'd be dropping data.

I would recommend refactoring this implementation so that all `io.ReadCloser` and `io.WriteCloser` that are returned when creating a compression/decompression engine, whether it's builtin or external, have identical semantics. For the builtin types, the returned types can be the (de)compressor for their respective implementations, but for the externals, you probably want to wrap the `Command` as a whole plus its pipes, so that reads/writes are going through the pipe, and `Close` actually performs the `Wait` on the subprocess. This will also cleanup the callsite for this interfaces: right now the (de)compressors are always `Closer`s, but the `Close` method must only be called for builtins, which makes the interface unsafe to use.

Let me know if you get stuck!

go/vt/mysqlctl/compression.go
```diff
@@ -0,0 +1,254 @@
...
+func validateExternalCmd(cmd string) (string, error) {
+	if cmd == "" {
+		return "", errors.New("external command is empty")
+	}
+
+	return exec.LookPath(cmd)
+}
+
+func prepareExternalCompressionCmd(ctx context.Context, cmdStr string) (*exec.Cmd, error) {
+	cmdArgs := strings.Split(cmdStr, " ")
```
vmg, line 69: This isn't perfectly cromulent; it will corrupt argument strings with quotes or escaped whitespace. I would suggest bringing https://pkg.go.dev/github.com/google/shlex as a (tiny) dependency to parse the commandline properly.

go/vt/mysqlctl/compression.go
rvrangel: neat, I will change that!

go/vt/mysqlctl/compression.go
rvrangel: Sounds good, I will take a look at it and make the changes 👍

rvrangel:
hi @vmg, thanks for the feedback. It is correct that we should not call `Wait()` when using `StdoutPipe`, in this code it only happens for the compressor, which uses `StdinPipe`, not `StdoutPipe` and is required to make sure everything we have written to the external process has been processed and written to is `Stdout` to avoid corrupting data. Nevertheless, wrapping everything so they are threaded in the same way makes sense, so I will make the changes and let you know

vmg:
I see now, I was a bit confused with the `WaitGroup` usage. I think if you wrap the `Command` and pipes, you should be able to run both compression and decompression with just a single goroutine (the logger), which should make it much easier to follow. Let's see what it looks like!

rvrangel:
@vmg I updated the PR, let me know if that covers most of what you thought and if you have any additional feedback

vmg (approved):
Very nice! Looks like a good start! Once @deepthi approves this we'll merge and I'll run some profiles to see how it performs in practice. Thank you!

rvrangel:
and thanks for the feedback @vmg! As a heads up, one limiting factor we encountered is the download speed for the restores when using S3, due to the fact `GetObject` is used, so we can't avail of multipart download the same way we do for uploads (which limits download speeds to <100MB/s):
https://github.com/vitessio/vitess/blob/440d2e68bb420c892842f3f9e177e20bd6424e76/go/vt/mysqlctl/s3backupstorage/s3.go#L191-L208

I have some code that implements multipart download as a buffer in memory, but still needs some polishing, but will soon create another PR. We saw some improvements on the 2-4x range when using other algorithms since `pgzip` is our main limiting factor at the moment. It can be helpful to test this PR with another backend that as of now can provide a better download speed (or maybe the `file` storage?) when you are testing it.

go/vt/mysqlctl/compression.go
```diff
@@ -0,0 +1,281 @@
+package mysqlctl
```
ajm188, line 1: this file is missing the copyright notice

go/vt/mysqlctl/compression_test.go
```diff
@@ -0,0 +1,199 @@
+package mysqlctl
```
ajm188, line 1: ditto on copyright notice

ajm188 (approved):
just need the copyright notices, then good from my end.

go/vt/mysqlctl/xtrabackupengine.go
```diff
@@ -63,6 +61,13 @@ var (
 	// striping mode
 	xtrabackupStripes         = flag.Uint("xtrabackup_stripes", 0, "If greater than 0, use data striping across this many destination files to parallelize data transfer and decompression")
 	xtrabackupStripeBlockSize = flag.Uint("xtrabackup_stripe_block_size", 102400, "Size in bytes of each block that gets sent to a given stripe before rotating to the next stripe")
+	// switch which compressor/decompressor to use
+	builtinCompressor   = flag.String("xtrabackup_builtin_compressor", "pgzip", "which builtin compressor engine to use")
+	builtinDecompressor = flag.String("xtrabackup_builtin_decompressor", "auto", "which builtin decompressor engine to use")
+	// use and external command to decompress the backups
```
deepthi, line 67: typo: and -> an

go/vt/mysqlctl/xtrabackupengine.go
```diff
@@ -63,6 +61,13 @@ var (
 	// striping mode
 	xtrabackupStripes         = flag.Uint("xtrabackup_stripes", 0, "If greater than 0, use data striping across this many destination files to parallelize data transfer and decompression")
 	xtrabackupStripeBlockSize = flag.Uint("xtrabackup_stripe_block_size", 102400, "Size in bytes of each block that gets sent to a given stripe before rotating to the next stripe")
+	// switch which compressor/decompressor to use
+	builtinCompressor   = flag.String("xtrabackup_builtin_compressor", "pgzip", "which builtin compressor engine to use")
+	builtinDecompressor = flag.String("xtrabackup_builtin_decompressor", "auto", "which builtin decompressor engine to use")
```
deepthi, line 66: It will be nice to add some help text on what "auto" means.

deepthi (commented):
Is there a reason the same support is not being added to the builtin engine? The compression options should be independent of the engine being used.
Also, more tests are needed. It should be fine to overload existing backup tests so that different tests use different options.

go/vt/mysqlctl/xtrabackupengine.go
deepthi: While adding support to `builtinbackupengine`, all these flags should be renamed so that they are not xtrabackup specific.

rvrangel: hi @deepthi thanks for the feedback. I will add support for it this week in the builtin engine and for the tests 👍

deepthi:
Superseded by #10558 which has been merged.

Issue #7688, opened 2021-03-15
Move TabletControls into TabletManager Context


makinje16:
Please search the existing issues for relevant feature requests, and use the [reaction feature](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to add upvotes to pre-existing requests.

### **Overview:**
This modification done to the Vitess codebase would **move TabletControls into the context of the tabletmanager, allow users to push  changes down to the tablet level using grpc commands, and separate the tablet control information which currently lives in both the global and local topos at times**.

Utilizing the grpc interface that Vitess currently uses we want to create new messages, methods, and implementation that allows us to use both grpc and vtctl commands to interact with a TabletControl data-structure that will reside within the context of the tabletmanager. This data structure will be persisted between restarts and refreshes by adding a table to the _vt database in mysql. `_vt.tablet_controls` is a possible table name.

This table will hold all information regarding a tablets tablet_control. The table rows can be created, updated, and or cleared by using the grpc methods mentioned earlier. On first initialization or when RefreshState is called, we will end up reading from this table to get starting values.

**[PR](https://github.com/vitessio/vitess/pull/7621) started here.**

</details>

### **Implementation Details**
These change will end up adding new grpc methods, responses, requests, and a `TabletControl` data-structure in the proto files for the tabletmanager:

<details open>

```
/* proto/tabletmanagerservice.proto */
    rpc UpdateTabletControls(tabletmanagerdata.UpdateTabletControlsRequest) returns (tabletmanagerdata.UpdateTabletControlsResponse) {};

/* proto/tabletmanagerdata.proto */
    message TabletControl {
      // Tablet type
      topodata.TabletType tablet_type = 1;

      string cell = 2;

      // should query service be enabled on this shard
      bool query_service = 3;

      repeated string denylist_tables = 4;

      bool frozen = 5;
    }
    
    message UpdateTabletControlsRequest {
      TabletControl tc = 1;
    }

    message UpdateTabletControlsResponse {
    }
``` 
</details>

The flow of Updating TabletControls would then be as follows:

- `UpdateTabletControls` is called on the primaries grpc server.
-  The primary tablet will update its `_vt.tablet_control` table and then call `WaitForPosition` on the non-primary tablets. The arg passed to WaitForPosition will be the position right after the primary successfully updates its tablet_control table.
- If all replicas are able to catch up the primary will return success to the grpc client meaning it is safe to refresh state
[...]

PR #7979, opened 2021-04-28
wrangler - refresh the tabletmanager view after SrvKeyspace changes

guidoiaquinti:
## Description
While working on #7929 I found few places where we don't refresh the tabletmanager state view after changes to the `SrvKeyspace` are made. This PR should address few of those.   

1. `MigrateServedTypes`: make sure we refresh the tabletmanager state view after we update the `SrvKeyspace` **for both the source and destination shard**.
1. `replicaMigrateServedType`: make sure we refresh the tabletmanager state view after we update the `SrvKeyspace`.
1. `masterMigrateServedType`: make sure we refresh the tabletmanager state view after we update the `SrvKeyspace` **also for the source shards**.
1. `switchShardReads`: make sure we refresh the tabletmanager state view after we update the `SrvKeyspace` in `MigrateServedType`.
1. `changeShardRouting`: make sure we refresh the tabletmanager state view after we update the `SrvKeyspace` in `MigrateServedType`.

## Checklist

## Deployment Notes
N/A

### DIFF
--- a/go/vt/wrangler/keyspace.go
+++ b/go/vt/wrangler/keyspace.go
@@ -461,14 +461,15 @@ func (wr *Wrangler) MigrateServedTypes(ctx context.Context, keyspace, shard stri
 	wr.Logger().Infof("WaitForDrain: Sleeping finished. Shutting down queryservice on old tablets now.")
 
 	rec := concurrency.AllErrorRecorder{}
-	refreshShards := sourceShards
-	if reverse {
-		// For a backwards migration, we should refresh (disable) destination shards instead.
-		refreshShards = destinationShards
+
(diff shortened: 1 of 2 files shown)


rafael:
Unfortunately, this change is not as straightforward. @makinje16 already tried to do something similar and that's one of the motivations behind: https://github.com/vitessio/vitess/issues/7688

What you are doing here is the expected behavior, however there are some edge cases where existent code relies on the fact that we are not always refreshing the topo. 

I think we should wait for#7688 to land and then revisit this.

guidoiaquinti:
Ack, I'll close this one so. Thanks!

PR #7980, opened 2021-04-28
VReplication: tracking `rows_copied`

shlomi-noach:
## Description

`_vt.vreplication` has new column, `rows_copied`, which is updated periodically during `vreplication` work, to reflect the actual number of rows written on the target table.

The value is written together with `time_updated`, so it does not get transactionally synced with each copy iteration, it follows closely though, and makes a good estimate.

This will be later used by Online DDL to track Vreplication-based migration progress & ETA. 

## Related Issue(s)

- Replaces https://github.com/vitessio/vitess/pull/7963
- Tracking: #6926 

## Checklist

### DIFF
--- a/go/vt/binlog/binlogplayer/binlog_player.go
+++ b/go/vt/binlog/binlogplayer/binlog_player.go
@@ -476,7 +476,7 @@ func (blp *BinlogPlayer) writeRecoveryPosition(tx *binlogdatapb.BinlogTransactio
 	}
 
 	now := time.Now().Unix()
-	updateRecovery := GenerateUpdatePos(blp.uid, position, now, tx.EventToken.Timestamp)
+	updateRecovery := GenerateUpdatePos(blp.uid, position, now, tx.EventToken.Timestamp, blp.blplStats.CopyRowCount.Get())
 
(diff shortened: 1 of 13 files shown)


shlomi-noach:
Good to review!

rohit-nayak-ps (approved):
LGTM

Just noting here that after the copy phase the rows_copied will never change.

shlomi-noach:
@rohit-nayak-ps currently - that's the objective. Just count the number of rows copied by `vcopier`. There is no adjustment by `vplayer`'s work. That'll come slightly later, and I see some larger things at play, I'll write more tomorrow. So anyway, leaving this to a different PR.

While this PR is good, I do notice that on a very long running vreplication stream, and with near zero traffic, `GenerateUpdatePos` either doesn't get called or is not applied. I see 25 minutes go by without an update to `time_updated`, nor `transaction_timestamp`, and thereby obviously not to `rows_copied`.

That's an unexpected behavior. It also relates to what I'll write more about tomorrow. So this is another thing to iterate on: I'd like to see `rows_copied` on the fly and frequently (at least once per minute).

PR #7983, opened 2021-04-28
Change vreplication error metric name to start with a string to pass prometheus validations

rohit-nayak-ps:

## Description

The vreplication error metric started with a number. Fixed by prefixing with the workflow name. Per prometheus docs the required naming convention is:

```
Label names may contain ASCII letters, numbers, as well as underscores. They must match the regex [a-zA-Z_][a-zA-Z0-9_]*. Label names beginning with __ are reserved for internal use.
```

## Checklist


ajm188 (approved):
LGTM

PR #7996, opened 2021-04-29
[WIP | Do not merge] add `errcheck` to golangci-lint

ajm188:

## Description

⚠️ I'm opening this explicitly not as a draft so I can see the CI checks run. It's not ready for review yet ⚠️ 

## Checklist


harshit-gangal:
@ajm188 AFAIK, in draft mode also the CI tests run :)

ajm188:
huh! i have a vague memory of turning a PR back to draft mode to debug some tests and none of the CIs were running. happy to be wrong though!

go/cmd/rulesctl/cmd/list.go
```diff
@@ -40,11 +40,13 @@ func List() *cobra.Command {
 				out = rules
 			}
 		} else {
-			out = rules.Find(listOptName)
-			if listOptNamesOnly && out != nil {
+			rule := rules.Find(listOptName)
```
ajm188, line 43: See https://github.com/vitessio/vitess/pull/7996/commits/6ab6b47bb907f8815bfee48a2a7aab2250aa534f for why this change

go/vt/binlog/binlogplayer/binlog_player.go
```diff
@@ -733,7 +733,9 @@ func MysqlUncompress(input string) []byte {
 		return nil
 	}
 	var outputBytes bytes.Buffer
-	io.Copy(&outputBytes, reader)
+	if _, err := io.Copy(&outputBytes, reader); err != nil {
```
ajm188, line 736: @rohit-nayak-ps I returned `nil` instead of error here following along the `outputBytes.Len()` check immediately below returning `nil` instead of error. Let me know if you think this should return an error instead.

go/vt/orchestrator/collection/collection.go
```diff
@@ -283,7 +283,7 @@ func (c *Collection) Append(m Metric) error {
 	c.Lock()         //nolint SA5011: possible nil pointer dereference
 	defer c.Unlock() //nolint SA5011: possible nil pointer dereference
 	// we don't want to add nil metrics
-	if c == nil {
+	if c == nil { //nolint SA5011: redundant nil pointer check - maybe this was supposed to check if m == nil ?
```
ajm188, line 286: @shlomi-noach calling your particular attention to this line; I think this should be checking if `m == nil` (since we checked `c == nil` just before locking above) but wanted to confirm with you first.

shlomi-noach: @ajm188 I believe you're right, this should be `m == nil`. Took me a while to understand why I have no recollection of this code, it was a contribution and I'm not too familiar with it.

shlomi-noach (commented):
LGTM

go/cmd/vtctl/vtctl.go
```diff
@@ -85,7 +85,7 @@ func main() {
 	startMsg := fmt.Sprintf("USER=%v SUDO_USER=%v %v", os.Getenv("USER"), os.Getenv("SUDO_USER"), strings.Join(os.Args, " "))
 
 	if syslogger, err := syslog.New(syslog.LOG_INFO, "vtctl "); err == nil {
-		syslogger.Info(startMsg)
+		syslogger.Info(startMsg) // nolint:errcheck
```
deepthi, line 88: nit: why not `_ = ...` instead of `nolint`?

deepthi (approved):
LGTM.
A couple of things:
* Review the `nolint` directives, perhaps some of them should change to ignore the return value instead
* It will be nice to update the version in https://github.com/vitessio/vitess/blob/main/misc/git/hooks/golangci-lint

go/vt/binlog/binlogplayer/binlog_player.go
rohit-nayak-ps: that is fine

Issue #4905, opened 2019-06-04
Add a way to run a restore from a specific backup


makmanalp:
#### Feature Description

Following discussion here: https://vitess.slack.com/archives/C0PQY0PTK/p1559677665019600

`-restorefrombackup` currently takes no argument but grabs only the latest backup. It would be nice to pass in the option of which backup to restore.

#### Use Case(s)

- Disaster recovery: it would basically give us point in time recovery for free, though as Deepthi pointed out, for a single shard unless you do some magic with replaying binlogs too. But also even on a single shard this is useful for e.g. a customer getting hacked and all their stuff being deleted. Or e.g. some mishap where a server took writes that it shouldn’t and we’re hand-fixing stuff. Or someone did a `delete from foo` without specifying a where clause, some job had bad logic and corrupted data, etc etc.
- Integrity checks: we’re interested in building an automated rig to fully verify specific backups by actually doing a restore, catching up to replication, and running a checksum, and if that works then marking that backup as "good".

semi-related: https://github.com/vitessio/vitess/issues/4886

PR #7998, opened 2021-04-29
vttablet - add '-restore_from_backup_ts' flag

guidoiaquinti:
## Description
Add a new `-restore_from_backup_ts` flag in `vttablet` to restore the last backup taken at or before the provided timestamp.

Note: in a future iteration we could also add this functionality to the `RestoreFromBackup` gRPC call. 

## Related Issue(s)
This should address #4905.

## Checklist

## Deployment Notes
N/A

### DIFF
--- a/go/vt/mysqlctl/backupengine.go
+++ b/go/vt/mysqlctl/backupengine.go
@@ -217,8 +217,8 @@ func FindBackupToRestore(ctx context.Context, params RestoreParams, bhs []backup
 				continue
 			}
 		}
-		if !checkBackupTime /* not snapshot */ || backupTime.Equal(params.StartTime) || backupTime.Before(params.StartTime) {
-			params.Logger.Infof("Restore: found backup %v %v to restore", bh.Directory(), bh.Name())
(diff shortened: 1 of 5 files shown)


deepthi (commented):
The main objection to this is that after restoring from the backup, the replica will try to catch-up using binlogs and if binlogs have been rotated out, then the replica will fail to go into serving.
With the current restore functionality (using the most recent backup), we can reasonably assume that binlogs are available.
The first usecase in #4905 is now covered by PITR. For the second use-case of backup validation, the way the issue is written, I assume it is expected to fail if binlogs are not available?

Re

guidoiaquinti:
> The main objection to this is that after restoring from the backup, the replica will try to catch-up using binlogs and if binlogs have been rotated out, then the replica will fail to go into serving.

Yes and I think we can consider this "by design" and it depends on each setup (how often do you run backups and what's the binlog retention policy).

deepthi (commented):
Mostly LGTM, but needs an actual test case.

mattlord:
Closing this and carrying it forward here: https://github.com/vitessio/vitess/pull/8824/

Thank you for the PR @guidoiaquinti ! I'll make sure that you retain the proper credit/attribution. 😃

PR #7999, opened 2021-04-29
[vtadmin-web] Display shard state on Tablets view + extract tablet utilities 

doeg:
## Description

A few more small changes that will come in handy for the Keyspace + Workflow views:

- Extracts some Tablet formatters + other utils to `utils/tablets.ts`
- Displays shard serving status on the Tablets view

Before:

Screen Shot 2021-04-29 at 4 00 40 PM

After:

Screen Shot 2021-04-29 at 4 00 49 PM

## Checklist

## Deployment Notes

N/A

### DIFF
--- a/web/vtadmin/src/components/pips/ShardServingPip.tsx
+++ b/web/vtadmin/src/components/pips/ShardServingPip.tsx
@@ -1,0 +1,32 @@
+/**
+ * Copyright 2021 The Vitess Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Pip, PipState } from './Pip';
+
+interface Props {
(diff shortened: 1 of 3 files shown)


web/vtadmin/src/components/pips/ShardServingPip.tsx
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+import { Pip, PipState } from './Pip';
+
+interface Props {
+    // If shard status is still pending, a neutral pip is used.
+    // The distinction between "is loading" and "is_master_serving is undefined"
+    // is useful, as it's common for the shard `shard.is_master_serving` to be
+    // excluded from API responses for non-serving shards (instead of being
+    // explicitly false.)
```
ajm188, line 24: If it will make your life easier (or more pleasant!) I'm pretty sure we can change the JSON marshaling in the http adapter to not omit fields with zero values

ajm188 (approved):
lgtm

web/vtadmin/src/components/pips/ShardServingPip.tsx
doeg: Oh that's good to know! Definitely not a blocker but seems useful for boolean + numeric values... perhaps one day I will change it, but in the meantime it doesn't really matter much. I made a card anyway: https://github.com/vitessio/vitess/projects/12#card-60142070 😎

doeg: (That said, this made me think of a small change I want to make, so I'll do that real quick.)

doeg: ......... okay, so I did the change and it turns out I like it way less, so I think this is fine to merge as-is. Once this unit test passes anyway, haha.

PR #8003, opened 2021-04-30
[vtadmin] experimental tabletdebug

ajm188:
## Description

This PR adds an experimental, http-only endpoint to vtadmin for making passthrough requests to a tablet's `/debug/vars` endpoint.

Example:

```
amason@some-host:~/vitess$ curl -s "${VTADMIN_URL}/api/experimental/tablet/${TABLET_HOSTNAME}/debug/vars | jq '.result.BuildGitBranch' 
HEAD
```

## Checklist
- [x] Tests were added or are not required – not required


doeg (approved):
I love it 😈 

We talked about this previously, but for posterity: is there anything I should keep in mind in querying this endpoint? Beyond the usual considerations like... y'know, "do not query this in an infinite loop". 

A reasonable front-end implementation would poll this endpoint around once every second (per client) -- I believe we arrived at "this should not be a problem, and if it is a problem, we can tune it later", which works for me.

ajm188:
🤔 I don't think so!! It's still experimental, so let's just keep going with "this should not be a problem, and if it is a problem, we can tune it later"

Issue #7994, opened 2021-04-29
Schema tracking v1


systay:
See #7995 for background

# Suggested solution

## 1. vttablet schema version tracking
Primary tablets will keep track of the schema of all tables of the underlying MySQL database. 
This includes column names, types and nullability.
Keeping track will be done by copying the relevant information_schema data to a new table in the `_vt` database on the primary.
Whenever a change is noticed, we can push out an event to the vtgates, informing them that they need to reload schema.

_ Questions: _ 
How do we notice that the underlying schema has changed?

## 2. vtgate queries a primary tablet
In this schema tracking V1, vtgate will simply talk with the first shard primary and keep the schema from the tablet mirrored in its local state.

The schema data fetched this way will be applied as authoritative columns for tables that the user has not already done so for. This was, users will be able to override the actual mysql schema, when that is needed.

- [x] At tablet startup - check if the necessary `_vt.schema` tables exist, if it is missing, it needs to be created.
- [x] When doing schema check, we can join the `_vt.schema` table with the `information_schema.columns` to figure out what has changed, and if something has changed, publish in the health check and update the `_vt.schema`
- [x] On the vtgate side, receive these events and trigger a schema update
- [x] Update the vtgate vschema copy with authoritative columns.

systay:
Edited this and #7995 to align them and reflect our current understanding

PR #8005, opened 2021-04-30
Detect and signal schema changes on vttablets

systay:
# Detect schema changes on the tablet.

With this PR, a vttablet that finds itself as a `master`, will periodically check if schema has changed on the underlying mysqld. If any schema changes are found, the changed table names will be sent to any listening vtgates using the pre existing health check mechanism.

If this feature is enabled, a database will be created on the underlying MySQL instance named `_vt`, where meta-information about the schema will be stored.

Related to #7994


shlomi-noach:
> If this feature is enabled, a database will be created on the underlying MySQL instance named _vt

I think `_vt` is _always_ created, if only for the heartbeat mechanism.

shlomi-noach:
Elaborating on my previous comment: `_vt` schema is always created, within a few seconds of tablet startup. There's multiple parties that would create it, but `onlineddl/Executor` is one I'm most familiar with.

Creation: 

https://github.com/vitessio/vitess/blob/0ce71c727e5883062d87ac333e28a226b585e14e/go/vt/vttablet/onlineddl/executor.go#L2292

Called by tablet state manager, `Open()`, right after tablet is ready.

Which runs these queries:

https://github.com/vitessio/vitess/blob/0ce71c727e5883062d87ac333e28a226b585e14e/go/vt/vttablet/onlineddl/schema.go#L345-L359

Which, in particular, include: 

https://github.com/vitessio/vitess/blob/0ce71c727e5883062d87ac333e28a226b585e14e/go/vt/vttablet/onlineddl/schema.go#L21

go/mysql/schema.go
```diff
@@ -34,8 +34,74 @@ const (
...
+	PRIMARY KEY (table_schema, table_name, ordinal_position))`
+
+	detectNewColumns = `
+select ISC.table_name
+from information_schema.columns as ISC
+	 left join _vt.schemacopy as c on 
+		ISC.table_name = c.table_name and 
+		ISC.table_schema=c.table_schema and 
+		ISC.ordinal_position = c.ordinal_position
+where ISC.table_schema = database() AND c.table_schema is null`
```
shlomi-noach, line 61: :+1:

go/mysql/schema.go
```diff
@@ -34,8 +34,74 @@ const (
...
+
+	detectChangeColumns = `
+select ISC.table_name
+from information_schema.columns as ISC
+	  join _vt.schemacopy as c on 
+		ISC.table_name = c.table_name and 
+		ISC.table_schema=c.table_schema and 
+		ISC.ordinal_position = c.ordinal_position
```
shlomi-noach, line 69: :+1:

go/mysql/schema.go
```diff
@@ -34,8 +34,74 @@ const (
...
+where ISC.table_schema = database() 
+	AND (not(c.column_name <=> ISC.column_name) 
+	OR not(ISC.character_set_name <=> c.character_set_name) 
+	OR not(ISC.collation_name <=> c.collation_name) 
+	OR not(ISC.column_type <=> c.column_type) 
+	OR not(ISC.column_key <=> c.column_key))`
```
shlomi-noach, line 75: :+1:

go/vt/vttablet/tabletserver/health_streamer.go
```diff
@@ -255,3 +297,82 @@ func (hs *healthStreamer) SetUnhealthyThreshold(v time.Duration) {
...
+	defer conn.Recycle()
+
+	if !hs.initSuccess {
+		hs.initSuccess, err = hs.InitSchemaLocked(conn)
+		if err != nil {
+			return err
+		}
+	}
+
+	qr, err := conn.Exec(ctx, mysql.DetectSchemaChange, 5, false)
```
shlomi-noach, line 325: anything significant about the number `5`? Or is it enough that it's nonzero? If the latter, then `1` is a better magic number than `5`.

go/vt/vttablet/tabletserver/tabletenv/config.go
```diff
@@ -107,6 +107,7 @@ func init() {
 	flag.Int64Var(&currentConfig.QueryCacheMemory, "queryserver-config-query-cache-memory", defaultConfig.QueryCacheMemory, "query server query cache size in bytes, maximum amount of memory to be used for caching. vttablet analyzes every incoming query and generate a query plan, these plans are being cached in a lru cache. This config controls the capacity of the lru cache.")
 	flag.BoolVar(&currentConfig.QueryCacheLFU, "queryserver-config-query-cache-lfu", defaultConfig.QueryCacheLFU, "query server cache algorithm. when set to true, a new cache algorithm based on a TinyLFU admission policy will be used to improve cache behavior and prevent pollution from sparse queries")
 	SecondsVar(&currentConfig.SchemaReloadIntervalSeconds, "queryserver-config-schema-reload-time", defaultConfig.SchemaReloadIntervalSeconds, "query server schema reload time, how often vttablet reloads schemas from underlying MySQL instance in seconds. vttablet keeps table schemas in its own memory and periodically refreshes it from MySQL. This config controls the reload time.")
```
shlomi-noach, line 109: I guess this code already exists. But it's unclear from the name `queryserver-config-schema-reload-time` what the units are. I'd assume it was a `Duration`, but it's in fact seconds. I guess too later to change the name,

go/vt/vttablet/tabletserver/health_streamer.go
```diff
@@ -61,9 +67,25 @@ type healthStreamer struct {
...
+	conns       *connpool.Pool
+	initSuccess bool
 }
 
 func newHealthStreamer(env tabletenv.Env, alias topodatapb.TabletAlias) *healthStreamer {
+	var newTimer *timer.Timer
+	var pool *connpool.Pool
+	if env.Config().SignalWhenSchemaChange {
+		reloadTime := env.Config().SchemaReloadIntervalSeconds.Get()
+		newTimer = timer.NewTimer(reloadTime)
```
shlomi-noach, line 82: Should be `time.Duration(reloadTime) * time.Second` otherwise reload takes place every `30 * 60` nanoseconds.

Tip/opinion: when a variable denotes a time, and when it's not a `Duration` type, I always name it with the units. ie. I'd do:
```go
reloadTimeSeconds := env.Config().SchemaReloadIntervalSeconds.Get()
```
IMO this avoids much confusion.

shlomi-noach (changes requested):
one bug found.

Fear mongering alert: I guess my reservation is that every X minutes (`30` in this code) we're going to query information_schema for all tables in a schema, which can be tens of thousands of tables in some environments. I actually never tried querying I_S for that many tables, and I fear this might be a risky operation.

go/vt/vttablet/tabletserver/health_streamer.go
```diff
@@ -96,14 +121,27 @@ func (hs *healthStreamer) Open() {
...
 
 func (hs *healthStreamer) Close() {
 	hs.mu.Lock()
 	defer hs.mu.Unlock()
 
 	if hs.cancel != nil {
+		if hs.ticks != nil {
+			hs.ticks.Stop()
+		}
 		hs.cancel()
```
harshit-gangal, line 145: conn pool also needs to be closed. `hs.conns.Close()`

go/vt/vttablet/tabletserver/health_streamer.go
systay: In this particular case, the returned object is a `time.Duration`, and not an integer as you might have thought. The duration is already in seconds. `SchemaReloadIntervalSeconds.Get()` looks like this:

```golang
return time.Duration(s * Seconds(1*time.Second))
```

systay:
> Fear mongering alert: I guess my reservation is that every X minutes (`30` in this code) we're going to query information_schema for all tables in a schema, which can be tens of thousands of tables in some environments. I actually never tried querying I_S for that many tables, and I fear this might be a risky operation.

This one of the reasons we are adding a flag for this feature. We don't think that it will be something everyone uses. The use case you mention sounds like a situation where you might not want to keep track of schema like this.

go/vt/vttablet/tabletserver/health_streamer.go
shlomi-noach: Oh, good. Thanks for explaining!

shlomi-noach (approved):

go/vt/vttablet/endtoend/framework/server.go
```diff
@@ -114,7 +114,7 @@ func StartServer(connParams, connAppDebugParams mysql.ConnParams, dbName string)
 	config.HotRowProtection.Mode = tabletenv.Enable
 	config.TrackSchemaVersions = true
 	config.GracePeriods.ShutdownSeconds = 2
-	config.SchemaReloadIntervalSeconds = tabletenv.Seconds(3)
+	config.SchemaReloadIntervalSeconds = tabletenv.Seconds(2.1)
```
shlomi-noach, line 117: Just curious why `2.1` :smile:

harshit-gangal: 2 seconds check related to no schema change was added by @vmg to avoid query cache test to fail. To make the test faster made it to `2.1`, i wanted to make it `2.0000001` but that would be too ambitious

shlomi-noach: Then I think `2.1` might also be too ambitious. Might introduce flakyness?

PR #8008, opened 2021-04-30
[wrangler|workflow] extract `*wrangler.streamMigrater` to `workflow.StreamMigrator`

ajm188:
## Description

This PR moves the bulk of the `stream_migrater.go` code in `package wrangler` to `package workflow`, and then replaces the guts of the `*wrangler.streamMigrater` struct with a pointer to the new type, and delegates all other function calls through to the new type (in order to minimize the number of callsites I needed to update, which I will come back to in the next PR).

**Note**: the private `*wrangler.trafficSwitcher` and `*wrangler.streamMigrater` types depend on each other, so in order to not have to move both at once, and create a huge diff, I've temporarily defined a `workflow.ITrafficSwitcher` interface, which covers every function of `*workflow.trafficSwitcher` that the StreamMigrator code depends on. This is _extremely_ temporary, and, in the next next PR, I will migrate `*trafficSwitcher`, at which point we won't need that interface anymore.

This also moves the private `streamType` enum.

## Related Issue(s)

- #7931 

## Checklist


rohit-nayak-ps (approved):
lgtm

Issue #8010, opened 2021-04-30
`srv_topo_timeout` of 1s too aggressive


aquarapid:
A recent change (#7278), introduced a timeout for srv_topo operations.  The default value was set at 1 second.  This seems reasonable, but it has had the unintended side-effect of causing topo timeouts, especially during vtgate startup (tablet wait/healthcheck), for users that have topology spread across multiple DCs.

In the interest of sane defaults, I suggest we lift the default to 5 seconds; and those who need a lower value can set it down.

For discoverability, here is the type of error you would see:

```
E0430 16:34:01.359471       1 resilient_server.go:311] GetSrvKeyspaceNames(context.Background.WithDeadline(2021-04-30 16:34:30.359023101 +0000 UTC m=+30.040465073 [28.999518338s]), xxxx4a1) failed: deadline exceeded: /vitess/tst006/global/cells/xxxx4a1/CellInfo (no cached value, caching and returning error)
F0430 16:34:01.359543       1 vtgate.go:157] gateway.WaitForTablets failed: deadline exceeded: /vitess/tst006/global/cells/xxxx4a1/CellInfo
```

deepthi:
Fixed by #8011

PR #7278, opened 2021-01-10
fix hang when local topo was down

Johnny-Three:
when local cell topo  server is down, all commands related with function GetSrvKeyspaceNames are hanging
Added topo timeout and when local cell topo is broken , will use cache instead , make sure not too much effect on vtgate

Signed-off-by: JohnnyThree <whereshallyoube@gmail.com>

## Related Issue(s)

- 

## Checklist
- [ ] Should this PR be backported?
- [x] Tests were added or are not required
- [x] Documentation was added or is not required

## Impacted Areas in Vitess
Components that this PR will affect:

- [ ]  Query Serving
- [ ]  VReplication
- [x]  Cluster Management
- [ ]  Build 
- [ ]  VTAdmin

PR #8011, opened 2021-04-30
Increase the default srv_topo_timeout

aquarapid:
Addresses #8010 ;  allowing defaults to work in more cases, and bringing us somewhat closer to the effective value (infinite) before #7278.  Seems like a reasonable compromise.


deepthi (approved):
I would have made it 30 seconds, but this is fine.

PR #8012, opened 2021-04-30
Added includeVSchema to ValidateSchemaKeyspace

makinje16:

## Description

This PR adds in `includeVSchema` to `ValidateSchemaKeyspace`. The goal is to give the same functionality that #7977 brings to `ValidateSchemaShard`. ValidateSchemaShard in fact, now utilizes ValidateSchemaShard when includeVSchema is true.

## Checklist


go/vt/vtctl/vtctl.go
```diff
@@ -2802,7 +2802,7 @@ func commandValidateSchemaKeyspace(ctx context.Context, wr *wrangler.Wrangler, s
 	if *excludeTables != "" {
 		excludeTableArray = strings.Split(*excludeTables, ",")
 	}
-	return wr.ValidateSchemaKeyspace(ctx, keyspace, excludeTableArray, *includeViews, *skipNoMaster)
+	return wr.ValidateSchemaKeyspace(ctx, keyspace, excludeTableArray, *includeViews, *skipNoMaster, false /*includeVSchema*/)
```
ajm188, line 2805: We should add a flag to this command (and probably `ValidateSchemaShard`, sorry for not pointing that out earlier) to be able to pass `true` through here.

go/vt/wrangler/schema.go
```diff
@@ -227,6 +233,25 @@ func (wr *Wrangler) ValidateSchemaKeyspace(ctx context.Context, keyspace string,
 	er := concurrency.AllErrorRecorder{}
 	wg := sync.WaitGroup{}
 
+	// If we are checking against the vschema then all shards
+	// should just be validated individually against it
+	if includeVSchema {
+		for _, shard := range shards {
+			wg.Add(1)
+			err := wr.validateSchemaShardConcurrent(ctx, keyspace, shard, excludeTables, includeViews, includeVSchema, &wg)
```
ajm188, line 241: the vschema object exists at the keyspace, not shard, level. i think this isn't quite the refactor we want.

i think it should be more like:

```go
func (wr *Wrangler) ValidateSchemaKeyspace(ctx, ....) {
    // other stuff
    if includeVSchema {
        wr.validateVSchema(ctx, keyspace)
}

func (wr *Wrangler) ValidateSchemaShard(ctx ....., includeVSchema bool) {
    // other stuff
    if includeVSchema {
        wr.validateVSchema(ctx, keyspace)
    }
}
```

where `validateVSchema` is this [this bit](https://github.com/vitessio/vitess/pull/8012/files#diff-325df117ec2cfabd09085e4914cd447d075be627b12d7aec1fcb74a1b038bda5R169-R183)

ajm188 (changes requested):
Almost there, I think. I think the current refactor makes the `ValidateSchemaShard` a little overloaded, and pulling out the vschema validation to its own function will make this a bit clearer.

go/vt/wrangler/schema.go
```diff
@@ -273,6 +277,60 @@ func (wr *Wrangler) ValidateSchemaKeyspace(ctx context.Context, keyspace string,
...
+	if err != nil {
+		return fmt.Errorf("GetVSchema(%s) failed: %v", keyspace, err)
+	}
+
+	shardFailures := concurrency.AllErrorRecorder{}
+	var wg sync.WaitGroup
+	wg.Add(len(shards))
+
+	for _, shard := range shards {
+		go func(shard string, shardFailures *concurrency.AllErrorRecorder) {
```
ajm188, line 292: You shouldn't need to pass a reference to the recorder here; it's in scope in the closure of the goroutine (similar to how you're using the wait group).

go/vt/wrangler/schema.go
```diff
@@ -273,6 +277,60 @@ func (wr *Wrangler) ValidateSchemaKeyspace(ctx context.Context, keyspace string,
...
+	}
+	wg.Wait()
+	if shardFailures.HasErrors() {
+		return fmt.Errorf("ValidateVSchema(%v, %v, %v, %v) failed: %v", keyspace, shards, excludeTables, includeViews, shardFailures.Error().Error())
+	}
+	return nil
+}
+
+// ValidateVSchemaKeyspace compares the schema of each primary tablet in "keyspace" and errs if there are differences
+func (wr *Wrangler) ValidateVSchemaKeyspace(ctx context.Context, keyspace string, excludeTables []string, includeViews bool) error {
```
ajm188, line 326: I think this is unused?

go/vt/wrangler/schema.go
```diff
@@ -142,6 +142,12 @@ func (wr *Wrangler) diffSchema(ctx context.Context, masterSchema *tabletmanagerd
 	tmutils.DiffSchema(topoproto.TabletAliasString(masterTabletAlias), masterSchema, topoproto.TabletAliasString(alias), replicaSchema, er)
 }
 
+// helper method to asynchronously validate a shards schema
+func (wr *Wrangler) validateSchemaShardConcurrent(ctx context.Context, keyspace, shard string, excludeTables []string, includeViews bool, includeVSchema bool, wg *sync.WaitGroup) error {
```
ajm188, line 146: unused?

ajm188 (commented):
Everything I have is pretty minor changes; I think Rafa was saying he had other thoughts so we should wait for him as well.

go/vt/vtctl/vtctl.go
```diff
@@ -2783,13 +2784,14 @@ func commandValidateSchemaShard(ctx context.Context, wr *wrangler.Wrangler, subF
...
 	}
-	return wr.ValidateSchemaShard(ctx, keyspace, shard, excludeTableArray, *includeViews, false /*includeVSchema*/)
+	return wr.ValidateSchemaShard(ctx, keyspace, shard, excludeTableArray, *includeViews, *includeVSchema)
 }
 
 func commandValidateSchemaKeyspace(ctx context.Context, wr *wrangler.Wrangler, subFlags *flag.FlagSet, args []string) error {
 	excludeTables := subFlags.String("exclude_tables", "", "Specifies a comma-separated list of tables to exclude. Each is either an exact match, or a regular expression of the form /regexp/")
 	includeViews := subFlags.Bool("include-views", false, "Includes views in the validation")
 	skipNoMaster := subFlags.Bool("skip-no-master", false, "Skip shards that don't have master when performing validation")
+	includeVSchema := subFlags.Bool("include-vschema", true, "Validate schemas against the vschema")
```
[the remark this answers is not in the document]
makinje16: changed 👍

go/vt/wrangler/schema.go
makinje16: I actually got linting errors when I did not pass in the recorder or shard which is why I passed them in

```
Linting go/vt/wrangler
go/vt/wrangler/schema.go:289:45: loopclosure: loop variable shard captured by func literal (govet)
			si, err := wr.ts.GetShard(ctx, keyspace, shard)
			                                         ^
go/vt/wrangler/schema.go:291:83: loopclosure: loop variable shard captured by func literal (govet)
				shardFailures.RecordError(fmt.Errorf("GetShard(%v, %v) failed: %v", keyspace, shard, err))
				                                                                              ^
go/vt/wrangler/schema.go:297:45: loopclosure: loop variable shard captured by func literal (govet)
					excludeTables, includeViews, keyspace, shard, err,
					                                       ^
```

go/vt/wrangler/schema.go
makinje16: removed

go/vt/wrangler/schema.go
ajm188: Yeah I think if we don't need it now (and we don't expect to immediately need it in the short-term), then we should hold off. We can always add it back later!

Issue #8001, opened 2021-04-29
DDL using explicit keyspaceID syntax


rafael:
#### Overview of the Issue

DDL statements should respect target shard/kid that has been specified by the user in the mysql session. 

#### Reproduction Steps

1. Create a  keyspace with multiple shards and no vschema.
2. Try to perform an ALTER targeting specific kid:
```
use keyspace[kid]
alter statement ...
```
3. This alter is not respecting the target and it will try to apply it to all shards on this keyspace. Because there is no vschema, you will get an error like this:
```
 Keyspace does not have exactly one shard
```

#### Expected behavior

* This worked prior to 9.0. It is a weird use case, but at Slack we rely on this behavior. 
* Given that the target is being explicit about kid/shard that the DDL should be run, Vitess shouldn't error and apply the DDL.

PR #8013, opened 2021-05-01
DDL bypass plan

harshit-gangal:

## Description

If there is destination set on the session it will bypass the DDL planner.

## Related Issue(s)

Fixes: https://github.com/vitessio/vitess/issues/8001

## Checklist


rafael (approved):
Thanks for the quick fix @harshit-gangal. LGTM

PR #8014, opened 2021-05-01
Update vreplicator docs

kirs:
In a few references in code and in the docs website I found that `Only "in_keyrange" expressions are supported in the where clause` - while that's not true and `vreplication` also supports plain equality comparisons like `WHERE somecol='blah'` (here's [the code](https://github.com/vitessio/vitess/blob/172fac7dec8b11937a4efb26ebf4bedf1771f189/go/vt/vttablet/tabletserver/vstreamer/planbuilder.go#L374); it was shipped as part of https://github.com/vitessio/vitess/pull/5949).

I'd like this to be clearly documented for others.


rohit-nayak-ps (approved):
lgtm
thanks, appreciate the documentation fix

PR #8015, opened 2021-05-02
Online DDL: progress & ETA for Vreplication migrations

shlomi-noach:
## Description

https://github.com/vitessio/vitess/pull/7980 introduced VReplicaiton tracking for `rows_copied`. This PR both intensifies that tracking, and follows up in Online DDL: the online DDL executor periodically checks `rows_copied` from `_vt.vstream`, compares with estimated `table_rows`, looks at `started_timestamp`, and generates values for `progress` and for `eta_seconds`.

Online DDL does so every 1 minute, normally, which means the valuee `progress` and `eta_seconds` are estimates only. Plus, `vcopier` itself only updates `rows_copied` every `30s`, so we could get almost a `90s` stale value for `progress` and `eta_seconds`. Which is acceptable to us at this tim.

## Related Issue(s)

- https://github.com/vitessio/vitess/issues/7632
- #6926 
- followup to https://github.com/vitessio/vitess/pull/7980

## Checklist


shlomi-noach:
This PR is generally good to go. I'm struggling if/whether to add `endtoend` test that checks this specific behavior.

Obviously it would be nice to test the behavior. But then, I'm not sure how to go about it. It would take a relatively long migration to be able to test this. We'd need a migration that is absolutely known to run for over `90sec`, and that we can ensure will show a non-zero, non-100% progress value, at some point. This can be done, at a cost. It will take long minutes to populate and test such a migrated table, and I'm concerned about impact to CI.

shlomi-noach:
Now also complements #8255 and updates `rows_copied` for VReplication migrations. Note that while migration is running `rows_copied` is only updated once per minute, and based on `vreplication`'s `rows_copied` which in itself can be delayed.

Wen the migration completes, `rows_copied` is ensured to be the final & exact value.

shlomi-noach:
This is now ready for review.

rohit-nayak-ps (approved):
lgtm

Issue #7997, opened 2021-04-29
Initial thoughts about VReplication optimization


shlomi-noach:
At this point in time I'm starting to look more in depth into Online DDL via VReplication, to make sure it's on par, and above, `gh-ost`'s capabilities. One of the first things I wanted to do is compare migration speed. I got my hands on an external server, and here goes.

My initial tests are on a `example/local` (unsharded, one primary, two replicas) cluster, and a large table (20M rows, ~1.7GB), with no traffic. So my intention was to first check the speed of `vcopier`. This follows an internal discussion with @rohit-nayak-ps and @vmg about how we should address packet size. I won't produce the wonderful graphs @vmg does, first because I'm still not so good at benchmarking, and secondly because things took a turn.

## The good news

With the default `vstream_packet_size` (`250000`), and with even smaller packet sizes (`65536`) and marginally as low as (`16384`), Vreplication migration is _faster_ than `gh-ost`:

- `gh-ost` migration: consistently `310s` - `320s`
- `online` (vreplication): migration: consistently `~200s`

With lower `vstream_packet_size` (e.g `4096`) VReplication runtime for same table jumps as high as `~20min` == `1200sec`

## Problem statement

The above is really great, but only in isolation. I see two immediate problems with VReplication's implementation that creates a problem in a non-isolated environment:

### 1. Packet size <=> copy size

This was discussed internally, as well: VReplication fetches data up to `n` bytes (`vstream_packet_size`) from the streamer, then applies that data via `insert` onto the target table. This means a `500KB` write per transaction. For `50byte` rows that means `10,000` rows; for `1KB` rows it means `500` rows.

The numbers are not unreasonable. However:

- Those are still big writes. We need to make sure [`binlog_cache_size`](https://dev.mysql.com/doc/refman/5.7/en/replication-options-binary-log.html#sysvar_binlog_cache_size) is at least as high as the packet size, or else every write is guaranteed to exceed the binlog cache size, which means the write goes down to disk o na temporary file associated with the session, _before_ being read back and actually written (again, to disk) into the binlog.
- For comparison, at my previous place, and where we developed `gh-ost`, we set on smaller sizes. We counted by rows, not by bytes, and limited ourselves to `50` rows at a time, no more than `100`. Experimenting with higher values rarely produced any speedup, but did immediately impact performance.

[...]

PR #8017, opened 2021-05-02
EXPERIMENTAL: breaking up vstreamer source query

shlomi-noach:
# DO NOT MERGE. Known to be breaking

## Description

This PR is an **EXPERIMENT** in breaking up `vstreamer` source query into multiple chunks. Instead of a single `select <column-list> from <table-name> order by <pk-columns>` query/transaction, we run multiple queries, transactions, of the form `select <column-list> from <table-name> order by <pk-columns> where <pk-columns > :lastPK> limit 1000` -- notice **limit 1000**, hard coded for now.

This is known to be breaking and I see the following issues:

1. Currently all queries still in same transaction. They need to be in different transactions.
2. Snapshot. In this experiment we only take snapshot & GTID at first transaction/query. But this means streaming of 2nd chunk and onwards have a different GTID context.
3. Chasing the end: we need to first measure lastPK on table, and limit query by that PK, otherwise we run into a scenario where rows are added and added during the operation, and row copy keeps chasing them and never completes.
4. In this simplified experiment there's a lot of iterative code that can probably be executed just once for all shards, or otherwise share memory better between invocations.

## Related Issue(s)

#7997 

## Checklist

## Deployment Notes

cc @rohit-nayak-ps

### DIFF
--- a/go/vt/vttablet/tabletserver/vstreamer/rowstreamer.go
+++ b/go/vt/vttablet/tabletserver/vstreamer/rowstreamer.go
@@ -32,6 +32,8 @@ import (
 	querypb "vitess.io/vitess/go/vt/proto/query"
 )
 
+const streamRowsLimit = 1000
+
 // RowStreamer exposes an externally usable interface to rowStreamer.
 type RowStreamer interface {
 	Stream() error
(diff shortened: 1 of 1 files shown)


shlomi-noach:
superseded by https://github.com/vitessio/vitess/pull/8044

PR #8020, opened 2021-05-03
Pass the provided keyspace through to `UpdateDisableQueryService` rather than hard-coding the sourceKeyspace

### LINKED ISSUE #8019: VReplication: trafficswitcher's `changeShardsAccess` always operates on the source keyspace
In `changeShardsAccess`, we always operate on the `sourceKeyspace` of the workflow [here](https://github.com/vitessio/vitess/blob/2544e5ab674fb87bd7cf35db8742e8b77b9831fb/go/vt/wrangler/traffic_switcher.go#L1363-L1368), ignoring the `keyspace` parameter:

```go
func (ts *trafficSwitcher) changeShardsAccess(ctx context.Context, keyspace string, shards []*topo.ShardInfo, access accessType) error {
	if err := ts.wr.ts.UpdateDisableQueryService(ctx, ts.sourceKeyspace, shards, topodatapb.TabletType_MASTER, nil, access == disallowWrites /* disable */); err != nil {
		return err
	}
	return ts.wr.refreshMasters(ctx, shards)
}
```

However in `allowTargetWrites`, we pass `targetKeyspace` as the `keyspace` parameter [here](https://github.com/vitessio/vitess/blob/2544e5ab674fb87bd7cf35db8742e8b77b9831fb/go/vt/wrangler/traffic_switcher.go#L1274-L1279):

```go
func (ts *trafficSwitcher) allowTargetWrites(ctx context.Context) error {
	if ts.migrationType == binlogdatapb.MigrationType_TABLES {
		return ts.allowTableTargetWrites(ctx)
	}
	return ts.changeShardsAccess(ctx, ts.targetKeyspace, ts.targetShards(), allowWrites)
}
```

Now, _technically_, this is currently fine, because we don't do any reshard workflows across keyspaces (it doesn't make sense), but that can't really be expressed in the code, and if we were to ever change that design constraint, we could get surprised by this.

ajm188:

## Description

See #8019 for details.

## Related Issue(s)

- Fixes #8019 

## Checklist
- [ ] ~~Tests were added or are not required~~
- [ ] ~Documentation was added or is not required~

### DIFF
--- a/go/vt/wrangler/traffic_switcher.go
+++ b/go/vt/wrangler/traffic_switcher.go
@@ -1361,7 +1361,7 @@ func (ts *trafficSwitcher) startReverseVReplication(ctx context.Context) error {
 }
 
 func (ts *trafficSwitcher) changeShardsAccess(ctx context.Context, keyspace string, shards []*topo.ShardInfo, access accessType) error {
-	if err := ts.wr.ts.UpdateDisableQueryService(ctx, ts.sourceKeyspace, shards, topodatapb.TabletType_MASTER, nil, access == disallowWrites /* disable */); err != nil {
+	if err := ts.wr.ts.UpdateDisableQueryService(ctx, keyspace, shards, topodatapb.TabletType_MASTER, nil, access == disallowWrites /* disable */); err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


rohit-nayak-ps (approved):
lgtm

PR #8024, opened 2021-05-03
vreplication: fix vreplication timing metrics

dyv:
## Description
* Bug Fix

Several of the vreplication metrics were recently updated to no longer be
correct. Specifically time.Now() is no longer executed when the function
starts and is instead executed when the defer statement is called after the end
of the function.

This updates the metrics to properly execute time.Now() at the start
of the functions.


shlomi-noach (approved):
LGTM. Just to explain it a bit in `golang` terms:

in
```
defer func() {
		vc.vr.stats.PhaseTimings.Record("catchup", time.Now())
	}()
```
the value of `time.Now()` is evaluated when the code inside the defer function runs, which is at the end of the `catchup` function.

In: 
```
defer vc.vr.stats.PhaseTimings.Record("catchup", time.Now())
```
the value of `time.Now()` is evaluated immediately, and cached until the deferred function is actually invoked.

rohit-nayak-ps (approved):
lgtm

PR #8031, opened 2021-05-04
Release 10.0.1

askdba:


systay (approved):

go/vt/servenv/version.go
```diff
@@ -1,3 +1,3 @@
 package servenv
 
-const versionName = "10.0.1-SNAPSHOT"
+const versionName = "10.0.2-SNAPSHOT"
```
deepthi, line 3: Do we change this for every patch release? I thought it was going to stay at 10.0.0-SNAPSHOT.

deepthi: Nvm, this makes sense.

deepthi (approved):

PR #8038, opened 2021-05-04
Handle online DDL user creation if we do not have SUPER privs (e.g. AWS Aurora)

aquarapid:

### DIFF
--- a/go/vt/vttablet/onlineddl/executor.go
+++ b/go/vt/vttablet/onlineddl/executor.go
@@ -361,7 +361,8 @@ func (e *Executor) readMySQLVariables(ctx context.Context) (variables *mysqlVari
 	return variables, nil
 }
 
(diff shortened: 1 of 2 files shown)


shlomi-noach (commented):
I see `REPLICATION SLAVE` is listed in both `sqlGrantOnlineDDLSuper` and `sqlGrantOnlineDDLUser`. Is that intentional?

aquarapid:
> I see `REPLICATION SLAVE` is listed in both `sqlGrantOnlineDDLSuper` and `sqlGrantOnlineDDLUser`. Is that intentional?

That was to minimize the textual diff, but you're right, it can just appear in the 2nd one.  Fixing.

shlomi-noach (approved):
For context, `gh-ost` user requirements: https://github.com/github/gh-ost/blob/master/doc/requirements-and-limitations.md#requirements, show s that we can do without `SUPER` in Vitess.

PR #8048, opened 2021-05-05
Make TwoPC engine initialization async.

aquarapid:
There was a "deadlock" here if you use semi-sync **and**
twopc during a failover, because we open the twopc
engine first, and it immediately tries to create its 
metadata tables in `_vt`.  On the PRS target, that
then hangs if using semi-sync, and since the code
to do the repointing of the replicas to the new master
sits "behind" the twopc engine open, the PRS would
never complete. 

### DIFF
--- a/go/vt/vttablet/tabletserver/tabletserver.go
+++ b/go/vt/vttablet/tabletserver/tabletserver.go
@@ -453,6 +453,11 @@ func (tsv *TabletServer) TableGC() *gc.TableGC {
 	return tsv.tableGC
 }
 
+// TwoPCEngineWait waits until the TwoPC engine has been opened, and the redo read
+func (tsv *TabletServer) TwoPCEngineWait() {
+	tsv.te.twoPCReady.Wait()
+}
+
 // SchemaEngine returns the SchemaEngine part of TabletServer.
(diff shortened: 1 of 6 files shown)


harshit-gangal:
Some of the TwoPC tests are failing now.

aquarapid:
> Some of the TwoPC tests are failing now.

Yeah, I guess since they start issuing TXes immediately.  I'll take a look.

go/vt/vttablet/tabletserver/tabletserver.go
```diff
@@ -453,6 +453,11 @@ func (tsv *TabletServer) TableGC() *gc.TableGC {
 	return tsv.tableGC
 }
 
+// TwoPCEngineWait waits until the TwoPC engine has been opened, and the redo read
+func (tsv *TabletServer) TwoPCEngineWait() {
```
deepthi, line 457: nit: rename to `WaitForTwoPCEngine`

deepthi (commented):
This is a clever idea. I have only one nit.
LGTM but let's have @sougou review as well.

harshit-gangal (approved):

deepthi:
Also need a corresponding issue.

deepthi (approved):

PR #8049, opened 2021-05-05
Add optional TLS feature to gRPC servers

noxiouz:
## Description

Implement TransportCredentials that allows a gRPC server to accept both
TLS and plain-text connections at the same time on the same endpoint.

This feature is vital to migrating Vitess cluster in production from plain-text to TLS **without** downtime.

As Vitess components talk to each other via gRPC, currently, it's not possible to do a rolling update: 
- not possible to update any gRPC server to listen to TLS as gRPC clients will not be able to connect to it
- not possible to update gRPC clients first as they won't be able to talk to any gRPC server

With the optional TLS feature this update is possible in 3 steps without downtime:
1. Restart gRPC servers to accept both TLS & plain-text
2. Update clients to use TLS only
3. Restart gRPC servers again to accept TLS only 

To detect a connection type TransportCredentials inspects the first 6 bytes read from a socket and compares them with TLS prefix. TLS prefix is different from plain-text HTTP2 prefix.

Originally implemented as a separate gRPC package: [go-grpc-optionaltls-creds](https://github.com/noxiouz/go-grpc-optionaltls-creds)

## Checklist


deepthi:
@noxiouz Thank you for the contribution. We can run CI once a reviewer has had a chance to look at the change.
In the meantime, can you fix the signoff? DCO check is failing.

noxiouz:
> DCO check is failing.
fixed :)

go/vt/grpcoptionaltls/certificates_test.go
```diff
@@ -0,0 +1,53 @@
...
+*/
+package grpcoptionaltls
+
+import "strings"
+
+// LocalhostCert is a PEM-encoded TLS cert with SAN IPs
+// "127.0.0.1" and "[::1]", expiring at Jan 29 16:00:00 2084 GMT.
+// generated from src/crypto/tls:
+// go run generate_cert.go  --rsa-bits 1024 --host 127.0.0.1,::1,example.com --ca --start-date "Jan 1 00:00:00 1970" --duration=1000000h
+var localhostCert = []byte(`-----BEGIN CERTIFICATE-----
```
noxiouz, line 21: I have not found, but maybe there is a better way to create certificates for tests. That one I copied from golang. I am happy to replace it

dkhenry: we use the `vt/tlstest` pacakge to generate certificates

go/vt/servenv/grpc_server.go
```diff
@@ -135,6 +138,9 @@ func createGRPCServer() {
 
 		// create the creds server options
 		creds := credentials.NewTLS(config)
+		if *GRPCEnableOptionalTLS {
```
dkhenry, line 141: Can you log a message here at the warning level to indicate that plan-text connections will be accepted.

dkhenry (approved):
LGTM, I would prefer if we could log that insecure connections would be accepted so that users would know this is a potentially unsafe flag

noxiouz:
Please, let me know if you need more details, tests, docs, etc from my side. I am happy to give any :)

dkhenry:
I think you should be good to go. I have just approved the CI to run for this PR

PR #8053, opened 2021-05-06
Fix update query plan with table alias and owned vindex

### LINKED ISSUE #8050: update to owned vindex query plan has errors when there's a table alias
Here's a test case that could be added to go/vt/vtgate/planbuilder/testdata/dml_cases.txt to demonstrate the problem. It's a slightly modified copy of an existing test case:

```
# update by primary keyspace id, changing one vindex column, with a table alias
"update user_metadata as um set email = 'juan@vitess.io' where um.user_id = 1"
{
  "QueryType": "UPDATE",
  "Original": "update user_metadata as um set email = 'juan@vitess.io' where um.user_id = 1",
  "Instructions": {
    "OperatorType": "Update",
    "Variant": "Equal",
    "Keyspace": {
      "Name": "user",
      "Sharded": true
    },
    "TargetTabletType": "MASTER",
    "ChangedVindexValues": [
      "email_user_map:3"
    ],
    "KsidVindex": "user_index",
    "MultiShardAutocommit": false,
    "OwnedVindexQuery": "select user_id, email, address, email = 'juan@vitess.io' from user_metadata as um where um.user_id = 1 for update",
    "Query": "update user_metadata as um set email = 'juan@vitess.io' where um.user_id = 1",
    "Table": "user_metadata",
    "Values": [
      1
    ],
    "Vindex": "user_index"
  }
}
```

This test case fails with this error:

```
--- FAIL: TestPlan (0.17s)
    --- FAIL: TestPlan/dml_cases.txt (0.03s)
        --- FAIL: TestPlan/dml_cases.txt/339_V3:_#_update_by_primary_keyspace_id,_changing_one_vindex_column,_with_a_table_alias_ (0.00s)
            plan_test.go:458: V3 - File: dml_cases.txt, Line: 339
                Diff:
                  (
                        """
                        ... // 14 identical lines
                            "KsidVindex": "user_index",
                            "MultiShardAutocommit": false,
                -           "OwnedVindexQuery": "select user_id, email, address, email = 'juan@vitess.io' from user_metadata as um where um.user_id = 1 for update",
                +           "OwnedVindexQuery": "select user_id, email, address, email = 'juan@vitess.io' from user_metadata where um.user_id = 1 for update",
                            "Query": "update user_metadata as um set email = 'juan@vitess.io' where um.user_id = 1",
                            "Table": "user_metadata",
                        ... // 6 identical lines
                        """
                  )
```

You can see that the `OwnedVindexQuery` currently built into the query plan is not a valid query. It doesn't keep the table alias in the `from` clause but still has references to it in the `where` clause

The net effect is that you get an error like this: `execUpdateEqual: updateVindexEntries: vttablet: rpc error: code = NotFound desc = Unknown column 'um.user_id' in 'where clause'` when this query plan is used

dweitzman:
Resolves #8050

Signed-off-by: David Weitzman <dweitzman@pinterest.com>

## Description

This is a bug fix for table aliases + owned vindexes, which might be a codepath that hasn't been exercised before

## Checklist
- [x] Tests were added

## Deployment Notes
n/a

### DIFF
--- a/go/vt/vtgate/planbuilder/update.go
+++ b/go/vt/vtgate/planbuilder/update.go
@@ -112,8 +112,25 @@ func buildChangedVindexesValues(update *sqlparser.Update, table *vindexes.Table,
 	if len(changedVindexes) == 0 {
 		return nil, "", nil
 	}
+
+	var tableExpr sqlparser.TableExpr
+	if len(update.TableExprs) != 1 {
+		return nil, "", vterrors.Errorf(vtrpcpb.Code_UNIMPLEMENTED, "multi-table update statement is not supported in sharded database")
+	}
+	for _, expr := range update.TableExprs {
+		// There is only one table.
+		tableExpr = expr
+	}
+
+	aliasedTableExpr, isAliased := tableExpr.(*sqlparser.AliasedTableExpr)
+	if isAliased && !aliasedTableExpr.As.IsEmpty() {
+		buf.Myprintf(" from %v as %v", table.Name, aliasedTableExpr.As)
+	} else {
+		buf.Myprintf(" from %v", table.Name)
+	}
+
 	// generate rest of the owned vindex query.
-	buf.Myprintf(" from %v%v%v%v for update", table.Name, update.Where, update.OrderBy, update.Limit)
+	buf.Myprintf("%v%v%v for update", update.Where, update.OrderBy, update.Limit)
 	return changedVindexes, buf.String(), nil
 }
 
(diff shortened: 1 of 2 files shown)


go/vt/vtgate/planbuilder/update.go
```diff
@@ -112,8 +112,25 @@ func buildChangedVindexesValues(update *sqlparser.Update, table *vindexes.Table,
 	if len(changedVindexes) == 0 {
 		return nil, "", nil
 	}
+
+	var tableExpr sqlparser.TableExpr
```
dweitzman, line 116: This may not be the most elegant approach. Initially I tried just printing out the entire TableExprs but that changed a bunch of query plans that use things like shard targeting to include the shard targeting in this owned vindex query.

I'm not sure if including shard targeting in the owned vindex query is ok, so I tried to avoid impacting any existing query plans.

Issue #8055, opened 2021-05-06
Online DDL: explicitly reject FOREIGN KEY DDLs


shlomi-noach:
Online DDL does not support `FOREIGN KEY`s at this time, and this is documented.

Right now a schema migration that has a `FOREIGN KEY` clause fails with an obscure parsing error. e.g.:

```sql
mysql> set @@ddl_strategy='online';
Query OK, 0 rows affected (0.00 sec)

mysql> alter table corder add FOREIGN KEY my_fk(customer_id) references customer(customer_id);
+--------------------------------------+
| uuid                                 |
+--------------------------------------+
| a8c5cd57_ae2d_11eb_9dfd_0a43f95f28a3 |
+--------------------------------------+

show vitess_migrations \G
*************************** 1. row ***************************
                 id: 1
     migration_uuid: a8c5cd57_ae2d_11eb_9dfd_0a43f95f28a3
           keyspace: commerce
              shard: 0
       mysql_schema: vt_commerce
        mysql_table: corder
migration_statement: alter table corder
           strategy: online
            options: -skip-topo
    added_timestamp: 2021-05-06 05:41:09
requested_timestamp: 2021-05-06 05:41:09
    ready_timestamp: 2021-05-06 05:41:10
  started_timestamp: 2021-05-06 05:41:10
 liveness_timestamp: 2021-05-06 05:41:10
completed_timestamp: NULL
  cleanup_timestamp: NULL
   migration_status: failed
           log_path: 
          artifacts: 
            retries: 0
             tablet: zone1-0000000100
     tablet_failure: 0
           progress: 0
  migration_context: vtgate:78e4ef9d-ae2d-11eb-9dfd-0a43f95f28a3
         ddl_action: alter
            message: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'table corder' at line 1 (errno 1064) (sqlstate 42000) during query: ALTER TABLE `_a8c5cd57_ae2d_11eb_9dfd_0a43f95f28a3_20210506054110_vrepl` alter table corder
        eta_seconds: -1
```

We should do better analysis of the query/schema to determine that it contains `FOREIGN KEY` constraints and clearly indicate these are not supported.

shlomi-noach:
Our parsing for `FOREIGN KEY`s do not take into account the optional index name. Thus, we are able to parse:
```sql
alter table corder add FOREIGN KEY (customer_id) references customer(customer_id);
```
[...]

Issue #8057, opened 2021-05-06
Parser does not fail on invalid query


shlomi-noach:
Based on https://github.com/vitessio/vitess/issues/8055

The following query:
```sql
alter table corder zzzz zzzz zzzz
```
should fail parsing. However, it parses without error, and on top of that, claims `isFullyParsed() == true`.

shlomi-noach:
On vitess:
```sql
mysql> set @@ddl_strategy='direct';
Query OK, 0 rows affected (0.00 sec)

mysql> alter table corder zzzz zzzz zzzz;
Query OK, 0 rows affected (0.01 sec)
```

On MySQL:
```sql
mysql> alter table demo zzzz zzzz zzzz;
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'zzzz zzzz zzzz' at line 1
```

shlomi-noach:
I traced the problem down to this path:

https://github.com/vitessio/vitess/blob/9cc166fabc05239edd1de4c135fd21c4f34f7ce9/go/vt/sqlparser/sql.y#L1869-L1871

shlomi-noach:
`alter_commands_list` allows just about anything. I'm not sure why that happens. The following query is also valid:

```sql
alter table corder zzzz zzzz zzzz remove partitioning;
```

in this path:

https://github.com/vitessio/vitess/blob/9cc166fabc05239edd1de4c135fd21c4f34f7ce9/go/vt/sqlparser/sql.y#L2052-L2058

shlomi-noach:
I also notice this line gets executed on such invalid SQL:

https://github.com/vitessio/vitess/blob/9cc166fabc05239edd1de4c135fd21c4f34f7ce9/go/vt/sqlparser/parser.go#L84

Other samples of wrongly accepted syntax:

```sql
alter table corder engine=innodb zzzz zzzz;
alter table corder drop column c zzzz zzzz;
```

GuptaManan100:
`alter_command_list` does not allow anything, it allows empty values. In the input `alter table corder zzzz zzzz zzzz` only `alter table corder` part is used in parsing and the remaining `zzzz zzzz zzzz` is not. alter_commands_list takes an empty value. 

[...]

PR #8058, opened 2021-05-06
parser: support index name in FOREIGN KEY clause; Online DDL to reject FK clauses

shlomi-noach:
## Description

Fixes the specific comment https://github.com/vitessio/vitess/issues/8055#issuecomment-833333689
The parser now supports queries of the form:
```sql
alter table corder add FOREIGN KEY my_fk (customer_id) references customer(customer_id);
```
- we have added support for the `index_name` part of the clause (`my_fk`) in the above.

Online DDL now outright rejects any query that has to do with foreign keys.

## Related Issue(s)

- #8055 

## Checklist


shlomi-noach:
Menawhile, https://github.com/vitessio/vitess/pull/8058/commits/f6d9ca7842c860ed5b9c9217c686f1c054c824c7 adds a bit of sanity to prevent _some cases_ of invalid `ALTER TABLE` queries, presented in #8057

go/vt/sqlparser/parse_test.go
```diff
@@ -1034,6 +1034,9 @@ var (
 		input: "alter table a add fulltext index idx (id)",
 	}, {
 		input: "alter table a add foreign key (id) references f (id)",
+	}, {
+		input:  "alter table a add foreign key the_idx (id) references f (id)",
+		output: "alter table a add foreign key (id) references f (id)",
```
GuptaManan100, line 1039: Should we also change the Format method of foreign key constraint to also print the constraint name if it is passed? If we do that then the expected output would be the same as the input

GuptaManan100 (approved):

go/vt/sqlparser/parse_test.go
shlomi-noach: > to also print the constraint name if it is passed?

I considered this, but I think not, because MySQL's `SHOW CREATE TABLE` does not show it.

GuptaManan100: This is the output I get running on MyCLI 
```sql
MySQL root@(none):test> show create table testing2;
+----------+-----------------------------------------------------------------------------------------------------------------------------------+
| Table    | Create Table                                                                                                                      |
+----------+-----------------------------------------------------------------------------------------------------------------------------------+
| testing2 | CREATE TABLE `testing2` (\n  `id` bigint(20) DEFAULT NULL,\n  `fk` bigint(20) DEFAULT NULL\n) ENGINE=InnoDB DEFAULT CHARSET=utf16 |
+----------+-----------------------------------------------------------------------------------------------------------------------------------+
1 row in set
Time: 0.007s
MySQL root@(none):test> alter table testing2 add foreign key new_fk_const (fk) references product112s (id);
You're about to run a destructive command.
Do you want to proceed? (y/n): y
Your call!
Query OK, 0 rows affected
Time: 0.040s
MySQL root@(none):test> show create table testing2;
+----------+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Table    | Create Table                                                                                                                                                                                                                                        |
+----------+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| testing2 | CREATE TABLE `testing2` (\n  `id` bigint(20) DEFAULT NULL,\n  `fk` bigint(20) DEFAULT NULL,\n  KEY `new_fk_const` (`fk`),\n  CONSTRAINT `testing2_ibfk_1` FOREIGN KEY (`fk`) REFERENCES `product112s` (`id`)\n) ENGINE=InnoDB DEFAULT CHARSET=utf16 |
+----------+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
1 row in set
Time: 0.007s
```
there is a key with the name `new_fk_const` which otherwise would be lost if we do not include the key name in the alter table statement. Am I missing something?

systay: I think our AST should be as minimally surprising as possible. If a constraint name is parsed, it should also be outputted in `Format`.
If we want to remove the constraint name, we should alter the struct instead of changing the `Format` method.

shlomi-noach: index name added in `Format()`

go/vt/schema/online_ddl.go
```diff
@@ -163,10 +177,30 @@ func NewOnlineDDLs(keyspace string, ddlStmt sqlparser.DDLStatement, ddlStrategyS
...
+				return nil, vterrors.Errorf(vtrpcpb.Code_INVALID_ARGUMENT, "NewOnlineDDL: cannot parse statement: %v", sqlparser.String(ddlStmt))
+			}
+		}
+
+		if !ddlStmt.IsFullyParsed() {
+			return nil, vterrors.Errorf(vtrpcpb.Code_INVALID_ARGUMENT, "NewOnlineDDL: cannot parse statement: %v", sqlparser.String(ddlStmt))
+		}
+
+		fk := &fkContraint{}
+		_ = sqlparser.Walk(fk.FkWalk, ddlStmt)
```
systay, line 198: nit: You can return the error from `fk.FkWalk` directly, which would also abort the traversal of the AST.

go/vt/schema/online_ddl.go
```diff
@@ -49,6 +49,21 @@ const (
...
+	found bool
+}
+
+func (fk *fkContraint) FkWalk(node sqlparser.SQLNode) (kontinue bool, err error) {
+	switch node.(type) {
+	case *sqlparser.CreateTable, *sqlparser.AlterTable,
+		*sqlparser.TableSpec, *sqlparser.AddConstraintDefinition, *sqlparser.ConstraintDefinition:
+		return true, nil
+	case *sqlparser.ForeignKeyDefinition:
+		fk.found = true
```
systay, line 62: nit: here you could return an error

shlomi-noach: Oh I like this! Updated code.

go/vt/schema/online_ddl.go
shlomi-noach: done!

systay (approved):

go/vt/schema/online_ddl.go
```diff
@@ -216,7 +250,7 @@ func NewOnlineDDL(keyspace string, table string, sql string, ddlStrategySetting
 			switch stmt := stmt.(type) {
 			case sqlparser.DDLStatement:
 				if !stmt.IsFullyParsed() {
-					return nil, vterrors.Errorf(vtrpcpb.Code_INVALID_ARGUMENT, "NewOnlineDDL: cannot fully parse statement %v", sqlparser.String(stmt))
+					return nil, vterrors.Errorf(vtrpcpb.Code_INVALID_ARGUMENT, "NewOnlineDDL: cannot parse statement: %v", sqlparser.String(stmt))
```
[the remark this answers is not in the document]
shlomi-noach: done

PR #8060, opened 2021-05-06
vreplication: fix vreplication timing metrics

systay:
Backport of #8024 

Several of the vreplication metrics were recently updated to no longer be
correct. Specifically time.Now() is no longer executed when the function
starts and is instead executed when the defer statement is called after the end
of the function.

This updates the metrics to properly execute time.Now() at the start
of the functions.

Signed-off-by: Dylan Visher <dvisher@figma.com>

## Checklist


rohit-nayak-ps (approved):
lgtm

PR #8061, opened 2021-05-06
VReplication: Pad binlog values for binary() columns to match the value returned by mysql selects

systay:
Backport of #7969

* Pad binlog values for binary() columns to match the value returned by a select query. This also ensures that if such columns are used as sharding keys we get the same keyspace_id

* Pad binary() values in the binlog reader directly so that all consumers see the padded value instead of doing it later in vstreamer or doint it just for keyspace id computation


rohit-nayak-ps (approved):
lgtm

PR #8064, opened 2021-05-06
Add image and base image arguments to build.sh

alexrs:
## Description

`docker/bootstrap/build.sh` accepts two arguments: flavor, and version. These arguments are used to decide what image to build.

This PR adds two new arguments: 
- `base_image`: The base image name. All Dockerfiles (except `common`) in `docker/bootstrap` accept an `image` argument. This argument can't be set when using `build.sh`.
- `image`: The image name to build. Defaults to `vitess/bootstrap:$version-$flavor$arch_ext`

This is useful if we are building our own images, and we want to specify a different name.

This change is backwards compatible.

## Checklist


shlomi-noach (commented):
This looks good. Could you please add comments in the script itself to explain what the arguments mean? I recognize the script currently has no documentation; but ideally all scripts should have some simple `usage` section.

alexrs:
Done @shlomi-noach!

shlomi-noach (approved):

PR #4163, opened 2018-08-28
Flush binlogs after reparent

rafael:
### Desc

Addresses #4161 

### Test

I tested this in our dev environment.  @sougou  do you think we should add some sort of integration/unit tests for this? Not sure how to go about it. Do you have ideas on that regard?

Issue #4161, opened 2018-08-28
Slow writes after PlannedReparent


rafael:
### Description

We have been debugging for a very **long** time an issue where after we PlannedReparent a shard, we get a spike of errors with the following symptoms:

1.    New master is accepting writes, but it's taking a long time to commit this transactions.
1.    Transaction pool get's exhausted and errors are propagated to the app. 
1.    After a few seconds, new master **warms** up and starts accepting writes. 

For context, a little bit about environment. We have some shards that have high write voulme. We use RBR with semi-sync in MySQL. Due to RBR and how we rotate our binlog files, they can grow to considerable large amounts (hundreds of MB) in a relative short period of time.

We've been having a theory where we think that the newly promoted master is not fully ready to receive semi-sync acks. @sougou had an explanation for this in: https://vitess.slack.com/archives/D6L0T2UAV/p1535427769000200. The problem could be because the master has to scan the binlogs to find the requested GTID.

Today @demmer had a really good idea and proposed that before doing a **PlannedReparent** we should **`FLUSH BINARY LOGS`** in the **replica** (soon to be master) so MySQL closes current binlog file and creates a new one. With this change, if there a writes that happen after the flush (which will always be the case in Vitess, due to PlannedReparent inserting rows in `_vt.reparent_journal`), scanning that file should be fast because it will be small. 

### Tests 

I did the following test in our dev environment with some promising results: 

1.  Generate tons of writes. 
1.  Wait for the binlog to grow (300/400 MB)
1.  Perform a `PlannedReparent`.
1.  Time how long it takes to perform this operation.
1.  Repeat the test, but run `FLUSH BINARY BINLOGS` in the target replica before reparenting.

Results:
1. Reparenting without flushing binlogs: ***20 to 30 seconds***. This is consistent with what we see in prod.
1. Reparenting with flushing binlogs: ***less than second***

### Proposal

We should enhance *PlannedReparent* to  `FLUSH BINARY BINLOGS` before promoting the target replica to master.

sougou:
@shlomi-noach any thoughts on this?

shlomi-noach:
Yeah, that makes sense to me. It is one of the unfortunate designs of GTID where you need to scan the entire binlog to make the first connection (with Pseudo-GTID you get markers throughout the binlog file and quick jump into the end of file).

[...]

PR #8066, opened 2021-05-07
reparent: flush binary logs on candidate primary

deepthi:
## Description

The problem described in #4161 has resurfaced. This is not actually a regression, it seems like the fix in #4179 doesn't necessarily help in a situation where the binlog file is large and the cluster is taking a lot of write traffic.

This PR tries to address that by flushing the binary logs **before** demoting the original primary. The problem with doing it **after** is that if no writes have happened after the binlog is flushed, then you still need to scan the previous file to find the GTID you are looking for.

The idea here is that in a high-traffic system, you are almost guaranteed to have some writes between flushing the binlogs and demoting the primary. In a low-traffic system it doesn't really matter.

## Related Issue(s)

#4161 #4163 #4179

## Checklist


go/vt/mysqlctl/reparent.go
```diff
@@ -118,3 +118,24 @@ func (mysqld *Mysqld) Promote(hookExtraEnv map[string]string) (mysql.Position, e
...
+	defer conn.Recycle()
+
+	// Since we handle replication, just stop it.
+	cmds := []string{
+		// When using semi-sync and GTID, a replica first connects to the new master with a given GTID set,
+		// it can take a long time to scan the current binlog file to find the corresponding position.
+		// This can cause commits that occur soon after the master is promoted to take a long time waiting
+		// for a semi-sync ACK, since replication is not fully set up.
+		// More details in: https://github.com/vitessio/vitess/issues/4161
+		"FLUSH BINARY LOGS",
```
deepthi, line 137: @rafael do you think we should delete this from `Promote`? Lines 108-113.

rafael:
I'm still puzzled by this issue. 

>The problem with doing it after is that if no writes have happened after the binlog is flushed, then you still need to scan the previous file to find the GTID you are looking for.
If the goal is to solve this for high write throughput systems, it seems that this shouldn't be a concern. 

I spoke with @aquarapid about this issue last week and I tried to reproduce this problem in a system with large binlogs and also high throughput writes. I couldn't reproduce and the existent solution was sufficient. I also tested reverting the original PR and ran into issues where PlannedReparents took a very long time. 

One more thought. I don't love the idea of having to add one more rpc call in the PlannedReparent codepath. This is something that we also tried to avoid the first time we did this. I was thinking that another way of handling this issue is by recommending users to tweak `max_binlog_size` so the binlog files never grow too large. I'm curious if this will be sufficient.

aquarapid:
I have reproduced the issue, to some extent.  I could get up to PRS up 20-30 seconds failover time (instead of the usual < 2-3 secs), not quite enough to time the PRS out;  but that may have been due to the shape of my workload, storage IO, etc.

I am not a huge fan of recommending a max_binlog_size change, since keeping it large is recommended by many MySQL experts, and I think we want a "good" experience regardless?

go/vt/mysqlctl/reparent.go
rafael: Yup. If we decide with this new approach, we should remove this.

rafael:
> I have reproduced the issue, to some extent. I could get up to PRS up 20-30 seconds failover time (instead of the usual < 2-3 secs), not quite enough to time the PRS out; but that may have been due to the shape of my workload, storage IO, etc.
>
Were you able to reproduce in MySQL 5.7?

> I am not a huge fan of recommending a max_binlog_size change, since keeping it large is recommended by many MySQL experts,

Do you know what's the rationale behind that recommendation? For different reasons, we flush binary logs regularly and our binlog files are relatively small (below 100mb). We haven't seen any issues with this.

>and I think we want a "good" experience regardless?

Yeah I share this as well.

@deepthi  - If this is a burning issue and the change here fixes the problem you are seeing, I don't have objections going with this approach. Expanding on why I'm hesitant: 

a) I couldn't reproduce this problem with the same conditions that the PR mentions will fix. I worry we a re missing something. This is the setup that I had:

1) Test table: 
```
CREATE TABLE `test_table` (
  `i` int(11) NOT NULL,
  `c` char(10) DEFAULT NULL,
  `counter` int(11) DEFAULT NULL
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4
```

2) Generating 6K QPS inserts like the following:
```
INSERT INTO test_table(i) VALUES (1)
```
3) Let the binlog files grow to 4GB of data.

4) Perform a PlannedReparent and assert that it finishes within few seconds. 

@aquarapid  - Is this test similar to the one you are doing? Could you try a similar test in your end and check if you get the same results as I did? 

b)  It seems we will be using a somewhat big hammer (the new sync rpc call in PRS), for something that is ultimately a MySQL bug.

aquarapid:
@rafael my test is similar;  but more specifically;  I let the current binlog file grow until it's near it's per file maximum (which is 1 GB; 
 the mysql default, and also maximum I believe).

RE the MySQL binlog size recommendation;  this is the MySQL default, and many DBAs are resistant to change it because they don't want to deal with the increase in number of binlogs that goes with this.  There's also the perception that switch binlogs has a performance impact when it occurs, something that is probably not true anymore, but has been in the past.

deepthi:
We now think the root cause was errant gtids.

PR #8067, opened 2021-05-07
Update the schema copy with minimal changes

systay:
Instead of deleting everything and refilling the table on every change, we only remove the rows for the updated tables and fill those in again.

### DIFF
--- a/go/vt/vttablet/tabletserver/health_streamer.go
+++ b/go/vt/vttablet/tabletserver/health_streamer.go
@@ -20,8 +20,11 @@ import (
 	"flag"
 	"fmt"
 	"io"
+	"strings"
 	"sync"
 	"time"
+
+	"vitess.io/vitess/go/vt/sqlparser"
 
 	"vitess.io/vitess/go/vt/dbconfigs"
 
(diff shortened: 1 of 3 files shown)


go/vt/vttablet/tabletserver/health_streamer.go
```diff
@@ -334,19 +337,32 @@ func (hs *healthStreamer) reload() error {
...
+	var tables []string
+	var tablePredicates []string
+	for _, row := range qr.Rows {
+		table := row[0].ToString()
+		tables = append(tables, table)
+
+		tableName := sqlparser.NewStrLiteral(table)
+		tablePredicates = append(tablePredicates, "table_name = "+sqlparser.String(tableName))
+	}
+	tableNamePredicates := strings.Join(tablePredicates, " OR ")
```
harshit-gangal, line 349: we can rewrite as `table_name in (<t1>, <t2>, ... )`

go/vt/vttablet/tabletserver/health_streamer.go
```diff
@@ -323,8 +326,8 @@ func (hs *healthStreamer) reload() error {
 		}
 	}
 
-	// TODO: fix the maxrows from using a magic number.
-	qr, err := conn.Exec(ctx, mysql.DetectSchemaChange, 10000, false)
+	maxrows := 10000
+	qr, err := conn.Exec(ctx, mysql.DetectSchemaChange, maxrows, false)
```
harshit-gangal, line 330: I meant we should, stream the results than having maxrows

systay: Why do we want to stream? I don't follow the logic

harshit-gangal (approved):

PR #8069, opened 2021-05-07
Materialize: Add additional comparison operators in Materialize and fix bug where they not applied for sharded keyspaces

rohit-nayak-ps:
## Description

This PR fixes a bug where comparison filters were being ignored in Materialize workflows if the source keyspace was sharded.

In addition, it adds support for additional operators (`>, >=, <, <=, <>`) . The filter expression can chain multiple comparison operations using `and` only.

Examples:
```
select * from t1 where id = 1
select * from t1 where id <> 1
select * from t1 where val > 'abc'
select * from t1 where id >= 1
select * from t1 where id < 2 and val <= 'xyz'
```
## Checklist

### DIFF
--- a/go/vt/vttablet/tabletserver/vstreamer/planbuilder.go
+++ b/go/vt/vttablet/tabletserver/vstreamer/planbuilder.go
@@ -145,14 +145,39 @@ func getOpcode(comparison *sqlparser.ComparisonExpr) (Opcode, error) {
 
 // compare returns true after applying the comparison specified in the Filter to the actual data in the column
 func compare(comparison Opcode, columnValue, filterValue sqltypes.Value) (bool, error) {
+	// NullsafeCompare returns 0 if values match, -1 if columnValue < filterValue, 1 if columnValue > filterValue
 	result, err := evalengine.NullsafeCompare(columnValue, filterValue)
 	if err != nil {
 		return false, err
 	}
-	if result == 0 && (comparison == Equal || comparison == LessThanEqual || comparison == GreaterThanEqual) ||
-		result < 0 && (comparison == LessThan || comparison == LessThanEqual || comparison == NotEqual) ||
-		result > 0 && (comparison == GreaterThan || comparison == GreaterThanEqual || comparison == NotEqual) {
-		return true, nil
+
+	switch comparison {
+	case Equal:
+		if result == 0 {
+			return true, nil
+		}
+	case NotEqual:
+		if result != 0 {
+			return true, nil
+		}
+	case LessThan:
+		if result == -1 {
+			return true, nil
+		}
+	case LessThanEqual:
+		if result <= 0 {
+			return true, nil
+		}
+	case GreaterThan:
+		if result == 1 {
+			return true, nil
+		}
+	case GreaterThanEqual:
+		if result >= 0 {
+			return true, nil
+		}
+	default:
+		return false, fmt.Errorf("comparison operator %d not supported", comparison)
 	}
 	return false, nil
 }
--- a/go/vt/vttablet/tabletserver/vstreamer/planbuilder.go
+++ b/go/vt/vttablet/tabletserver/vstreamer/planbuilder.go
@@ -145,6 +145,11 @@ func getOpcode(comparison *sqlparser.ComparisonExpr) (Opcode, error) {
 
 // compare returns true after applying the comparison specified in the Filter to the actual data in the column
 func compare(comparison Opcode, columnValue, filterValue sqltypes.Value) (bool, error) {
+	// use null semantics: return false if either value is null
+	if columnValue.IsNull() || filterValue.IsNull() {
+		return false, nil
+	}
+	// at this point neither values can be null
 	// NullsafeCompare returns 0 if values match, -1 if columnValue < filterValue, 1 if columnValue > filterValue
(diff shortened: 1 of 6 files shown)


go/vt/vttablet/tabletserver/vstreamer/planbuilder.go
```diff
@@ -111,6 +121,42 @@ func (plan *Plan) fields() []*querypb.Field {
...
+// compare returns true after applying the comparison specified in the Filter to the actual data in the column
+func compare(comparison Opcode, columnValue, filterValue sqltypes.Value) (bool, error) {
+	result, err := evalengine.NullsafeCompare(columnValue, filterValue)
+	if err != nil {
+		return false, err
+	}
+	if result == 0 && (comparison == Equal || comparison == LessThanEqual || comparison == GreaterThanEqual) ||
+		result < 0 && (comparison == LessThan || comparison == LessThanEqual || comparison == NotEqual) ||
+		result > 0 && (comparison == GreaterThan || comparison == GreaterThanEqual || comparison == NotEqual) {
+		return true, nil
```
shlomi-noach, line 155: Personal preference: break this `if` into 3 `if` statements (`== 0`, `< 0`, `> 0`), each of which ends with `return true, nil`.
Alternatively, a `switch` statement with 3 cases. 
There's just too much `&&` and `||` going on.

shlomi-noach (commented):
I'm not sure what in this PR adds support for more Opcodes other than the parser part. I'm guessing they are automatically handled later on by the same existing code?

Otherwise I confess I don't understand the code nor the changes :/

go/vt/vttablet/tabletserver/vstreamer/planbuilder.go
systay: another alternative is to use bitmasks for the comparisons, so you could write something like:

```
if result == 0  {
  return comparison =| 1 << Equal, nil
}
if result < 0  {
  return comparison =| 1 << LessThan, nil
}
if result > 0  {
  return comparison =| 1 << GreaterThan, nil, nil
}
```

shlomi-noach: I find this syntax confusing. Didn't ever see `=|` used before in `golang`

systay: how are you used to checking bitmasks?

shlomi-noach: C-style: `if my_var | 0x0040 { do_something }`

systay: sorry for the confusing comment. I was thinking of something like:

```golang
	const (
		NotEqual = 0
		Equal = 1 << iota
		LessThan
		GreaterThan
	)
	var cmp int
	var x sqlparser.ComparisonExpr
	switch x.Operator {
	case sqlparser.EqualOp:
		cmp = Equal
	case sqlparser.LessThanOp:
		cmp = LessThan
	case sqlparser.GreaterThanOp:
		cmp = GreaterThan
	case sqlparser.LessEqualOp:
		cmp = LessThan | Equal
	case sqlparser.GreaterEqualOp:
		cmp = GreaterThan | Equal
	case sqlparser.NotEqualOp:
		cmp = NotEqual
	}
	var result int
	if cmp == NotEqual {
		return result != 0, nil
	}
	if result == 0 {
		return cmp&Equal != 0
	}
	if result < 1 {
		return cmp&LessThan != 0
	}
	if result > 0 {
		return cmp&GreaterThan != 0
	}
```

rohit-nayak-ps: Simplifed the checks to a switch-based one. @shlomi-noach, `newVStreamer()` in vstreamer.go has an explanatory comment on `Filters`.

shlomi-noach (approved):

go/vt/vttablet/tabletserver/vstreamer/planbuilder.go
```diff
@@ -111,6 +121,67 @@ func (plan *Plan) fields() []*querypb.Field {
...
+	default:
+		return -1, fmt.Errorf("comparison operator %s not supported", comparison.Operator.ToString())
+	}
+	return opcode, nil
+}
+
+// compare returns true after applying the comparison specified in the Filter to the actual data in the column
+func compare(comparison Opcode, columnValue, filterValue sqltypes.Value) (bool, error) {
+	// NullsafeCompare returns 0 if values match, -1 if columnValue < filterValue, 1 if columnValue > filterValue
+	result, err := evalengine.NullsafeCompare(columnValue, filterValue)
```
sougou, line 149: Don't we need a compare that handles null values semantics here?

rohit-nayak-ps:
Rebasing against main seems to have messed up this PR. Researcing.

go/vt/vttablet/tabletserver/vstreamer/planbuilder.go
rohit-nayak-ps: Add null value semantics

rohit-nayak-ps:
Recreated PR at https://github.com/vitessio/vitess/pull/8247 since this PR got messed up while rebasing with main.

PR #8089, opened 2021-05-10
Queryservice fix

harshit-gangal:
## Checklist

### DIFF
--- a/go/vt/vttablet/sandboxconn/sandboxconn.go
+++ b/go/vt/vttablet/sandboxconn/sandboxconn.go
@@ -114,6 +114,8 @@ type SandboxConn struct {
 
 	// this error will only happen once
 	EphemeralShardErr error
+
+	NotServing bool
 }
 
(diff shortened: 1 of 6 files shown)


harshit-gangal:
The fix for the old issue is still inplace. It just removed the additional check done to retrieve queryservice using tablet alias.

go/vt/vttablet/sandboxconn/sandboxconn.go
```diff
@@ -557,6 +565,11 @@ func (sbc *SandboxConn) Tablet() *topodatapb.Tablet {
 	return sbc.tablet
 }
 
+// ChangeTabletType changes the tablet type.
+func (sbc *SandboxConn) ChangeTabletType(typ topodatapb.TabletType) {
+	sbc.tablet.Type = typ
+}
+
```
deepthi, line 572: Where is this being used?

PR #8094, opened 2021-05-11
Set fully parsed to false when ignoring errors

GuptaManan100:
## Description
For queries like 
```sql
alter table corder zzzz zzzz zzzz
```
`fullyParsed` should be set to false but that was not the case until now.

## Related Issue(s)
#8057 

## Checklist


shlomi-noach (approved):
:+1: :rocket:

PR #8097, opened 2021-05-11
Fix for query and sub query with limits

systay:
## Description
An outer query with a sub query, where both have `LIMIT` declared, sometimes will use the wrong limit for the outer query.

Example query:
```sql
SELECT * 
FROM t
WHERE x > (SELECT x FROM y ORDER BY col LIMIT 1)
LIMIT 100
```

In this query, the outer query should return 100 rows, but will before this change instead return a single row.

## Related Issue(s)
Replaces #7792

## Checklist


yangxuanjia:
@systay , my pleasure. 
Wonderful test case, It's hard to me.

harshit-gangal (approved):

PR #8098, opened 2021-05-11
[9.0] Queryservice fix

systay:
An issue was introduced In a recent fix (#8041) which was a backport of #7879 

Before sending queries to a tablet, #7879 changed the behaviour to check if the tablet is ready to answer, by checking it's ServingStatus and that the tablet type hasn't changed.

If PlannedReparentShard is going on, this check should not be done for transactions in flight.
The vttablet waits for inflight transactions to get commit/rollback i.e. we want queries to existing transaction to be sent down to get the transaction completed, even if the tablet is currently saying it is NotServing.

The test that exposed this issue was already in the code base: go/test/endtoend/tabletgateway/buffer/buffer_test.go became flaky after #7879 was merged.

So, to fix the issue, the pre-check is removed from Gateway when getting the tablet connection for existing active shard_sessions with vttablet.

This also means that the reserved connection that used to reset based on this pre-check logic will have to hit the vttablet first and then only will reset the shard session on receiving the expected error making it two round trips.

## Related Issue(s)
Backport of #8089

### DIFF
--- a/go/vt/discovery/fake_healthcheck.go
+++ b/go/vt/discovery/fake_healthcheck.go
@@ -150,13 +150,6 @@ func (fhc *FakeHealthCheck) TabletConnection(alias *topodatapb.TabletAlias, targ
 	defer fhc.mu.RUnlock()
 	for _, item := range fhc.items {
 		if proto.Equal(alias, item.ts.Tablet.Alias) {
-			if !item.ts.Serving {
-				return nil, vterrors.Errorf(vtrpcpb.Code_FAILED_PRECONDITION, vterrors.NotServing)
-			}
-			if target != nil && !proto.Equal(item.ts.Target, target) {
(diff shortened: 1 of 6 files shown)


harshit-gangal:
some unit tests are failing.

deepthi:
As discussed offline, let us create an issue the describes the regression and link to the backport PRs.

deepthi (dismissed):

deepthi:
> some unit tests are failing.

We should have listened to @harshit-gangal. One of the perils of doing backports is that we may pull in changes that don't make sense on an older branch. That is what seems to have happened here. Tests should pass with the latest commit that reverts the unwanted changes.

deepthi (approved):
I reviewed the diff carefully against the diff in #8089 to ensure that we are not pulling in anything extraneous.
LGTM

PR #8099, opened 2021-05-11
[10.0] Queryservice fix

systay:
If PRS is going on, then Query serving will fail with the recent change in reserved connection fix. Before sending to tablet there was check introduce to see if the tablet is serving and the target has not changed then send the query to vttablet.
This is not true for transaction as we want the query which is in transaction to be sent down to get the transaction completed, also the PRS is waiting for any outstanding transaction to get completed.

So, this change is to remove the logic around it and the reserved connection will reset the shard session only when the error is received from the vttablet and there is no pre-check for it.

---

This is a combination of 3 commits.

* remove precheck of tablet serving and target
* remove the additional logic and return error if queryservice not found to serve query
* fix test as per new change

## Related Issue(s)
Backport of #8089

### DIFF
--- a/go/vt/vttablet/sandboxconn/sandboxconn.go
+++ b/go/vt/vttablet/sandboxconn/sandboxconn.go
@@ -114,6 +114,8 @@ type SandboxConn struct {
 
 	// this error will only happen once
 	EphemeralShardErr error
+
+	NotServing bool
 }
 
 var _ queryservice.QueryService = (*SandboxConn)(nil) // compile-time interface check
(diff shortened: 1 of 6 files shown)


harshit-gangal (approved):

go/vt/vttablet/sandboxconn/sandboxconn.go
```diff
@@ -557,6 +565,11 @@ func (sbc *SandboxConn) Tablet() *topodatapb.Tablet {
 	return sbc.tablet
 }
 
+// ChangeTabletType changes the tablet type.
+func (sbc *SandboxConn) ChangeTabletType(typ topodatapb.TabletType) {
+	sbc.tablet.Type = typ
+}
+
```
deepthi, line 572: Where is this being used?

PR #8100, opened 2021-05-11
[vtadmin-web] Do not parse numbers/booleans in URL query parameters by default

doeg:

## Description

Currently, VTAdmin by default parses numeric URL query parameters as numbers. This resulted in unintended behaviour with shard filtering, where filtering by a string like "000" (to, say, find shard `0001-0002`) would be parsed as `0` (no leading zeroes) and would _also_ evaluate as falsey, which caused a host of other unfun side-effects. 

In this screenshot, I tried to filter by "00". Note the `?filter=0` in the URL and the empty text input:

Screen Shot 2021-05-11 at 1 48 50 PM

It turns out that _not_ parsing booleans/numbers is the more common VTAdmin use case. The one exception so far is pagination, were we do want `page=1` to parse as an integer. 

Here's what it looks like with the fix:

Screen Shot 2021-05-11 at 1 48 42 PM

## Related Issue(s)

N/A

## Checklist

## Deployment Notes

N/A


ajm188 (approved):
nice find, and great writeup!

PR #8111, opened 2021-05-12
[vtadmin-web] Add Keyspace detail view

doeg:

## Description

This PR adds the foundation for the Keyspace detail view, starting with a listing of shards + tablets.

A caveat: grouping table rows by shard complicates the filtering, since `filterNouns` wasn't written to handle nested properties. For now, "kinda gross but exhaustively commented hacks that work 90% of the time in this one place" seemed a better trade-off than "don't add filtering until we can handle this in a nicely generalized way". (Especially since using a library like https://fusejs.io is a better long-term filtering solution _anyway_ ... who wants to maintain regexes? (Not me!)). I ticketed https://github.com/vitessio/vitess/projects/12#card-60968004 to investigate a more robust solution, when we get time.

Here's what it looks like:

Screen Shot 2021-05-12 at 1 54 18 PM
Screen Shot 2021-05-12 at 1 53 17 PM
Screen Shot 2021-05-12 at 12 50 14 PM

And a weird, semi-error state just for fun:
Screen Shot 2021-05-12 at 1 53 41 PM

## Related Issue(s)

N/A

## Checklist

## Deployment Notes

N/A

### DIFF
--- a/web/vtadmin/src/components/routes/keyspace/KeyspaceShards.tsx
+++ b/web/vtadmin/src/components/routes/keyspace/KeyspaceShards.tsx
@@ -1,0 +1,229 @@
+/**
+ * Copyright 2021 The Vitess Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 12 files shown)


web/vtadmin/src/components/routes/keyspace/KeyspaceShards.tsx
```diff
@@ -0,0 +1,229 @@
...
+
+// Filtering data is complicated by how we group rows by their shard,
+// since we want the filtering to apply to both shard-level properties,
+// ("shard:-80") as well as nested tablet-level properties ("hostname:some-tablet-123").
+//
+// This gets surprisingly complex: what do you do when fuzzy filtering,
+// and you match a shard, and also _some_ tablets in the shard -- do you show all
+// of the tablets (since the shard matches), or (more likely) just the filtered ones?
+//
+// Instead of doing a complicated-and-hacky solution, this approach is a
```
ajm188, line 146: This makes sense to me!

ajm188 (approved):
LGTM

PR #8113, opened 2021-05-12
WIP: implement ApplyVSchema in grpc-vtctld

acharisshopify:
@ajm188 am i on the right track here?  if so, i'll add tests.
context: https://vitess.slack.com/archives/C0PQY0PTK/p1620754861183700

The goal here is to add the ApplyVSchema api call to the grpc-vtctld service.

Signed-off-by: Alex Charis <alex.charis@shopify.com>

## Checklist


acharisshopify:
CC: @alexrs

go/vt/vtctl/grpcvtctldserver/server.go
```diff
@@ -793,6 +793,15 @@ func (s *VtctldServer) GetVSchema(ctx context.Context, req *vtctldatapb.GetVSche
 	}, nil
 }
 
+// ApplyVSchema is part of the vtctlservicepb.VtctldServer interface.
+func (s *VtctldServer) ApplyVSchema(ctx context.Context, req *vtctldatapb.ApplyVSchemaRequest) (*vtctldatapb.ApplyVSchemaResponse, error) {
+	if err := s.ts.SaveVSchema(ctx, req.Keyspace, req.VSchema); err != nil {
```
ajm188, line 798: looks like you're missing the ["keyspace existence" check](https://github.com/vitessio/vitess/blob/61e812af4546c7609b9359f0bd081f3f498f7844/go/vt/vtctl/vtctl.go#L3231-L3236)

proto/vtctldata.proto
```diff
@@ -414,6 +414,14 @@ message GetVSchemaResponse {
   vschema.Keyspace v_schema = 1;
 }
 
+message ApplyVSchemaRequest {
+  string keyspace = 1;
+  vschema.Keyspace v_schema = 2;
+}
+
+message ApplyVSchemaResponse {
```
ajm188, line 422: I wonder if it will be useful to include the final, saved vschema here (after reading it back out of the topo, similar to how [`ChangeTabletType`](https://github.com/vitessio/vitess/blob/master/go/vt/vtctl/grpcvtctldserver/server.go#L110-L123) works

ajm188 (commented):
This is definitely on the right track. In addition to the testing (there are many examples of how to use the helpers in `server_test.go`), you'll also need to add a cli entrypoint, and I think that's where we should include the `sql-mode/json-mode` behavior of the [original vtctl command](https://github.com/vitessio/vitess/blob/61e812af4546c7609b9359f0bd081f3f498f7844/go/vt/vtctl/vtctl.go#L3156-L3217)

proto/vtctldata.proto
acharisshopify: and similar to how the current vtctlclient command works

acharisshopify: done

go/vt/vtctl/grpcvtctldserver/server.go
acharisshopify: done

proto/vtctldata.proto
```diff
@@ -414,6 +414,14 @@ message GetVSchemaResponse {
   vschema.Keyspace v_schema = 1;
 }
 
+message ApplyVSchemaRequest {
+  string keyspace = 1;
+  vschema.Keyspace v_schema = 2;
```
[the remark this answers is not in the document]
acharisshopify: done

proto/vtctlservice.proto
```diff
@@ -91,6 +91,8 @@ service Vtctld {
   rpc GetTablets(vtctldata.GetTabletsRequest) returns (vtctldata.GetTabletsResponse) {};
   // GetVSchema returns the vschema for a keyspace.
   rpc GetVSchema(vtctldata.GetVSchemaRequest) returns (vtctldata.GetVSchemaResponse) {};
+  // ApplyVSchema applies a vschema to a keyspace.
+  rpc ApplyVSchema(vtctldata.ApplyVSchemaRequest) returns (vtctldata.ApplyVSchemaResponse) {};
```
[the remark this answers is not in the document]
acharisshopify: done

proto/vtctldata.proto
```diff
@@ -414,6 +414,14 @@ message GetVSchemaResponse {
   vschema.Keyspace v_schema = 1;
 }
 
+message ApplyVSchemaRequest {
```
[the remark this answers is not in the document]
acharisshopify: done

go/vt/vtctl/grpcvtctldserver/server.go
```diff
@@ -793,6 +793,15 @@ func (s *VtctldServer) GetVSchema(ctx context.Context, req *vtctldatapb.GetVSche
 	}, nil
 }
 
+// ApplyVSchema is part of the vtctlservicepb.VtctldServer interface.
```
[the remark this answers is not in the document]
acharisshopify: done

acharisshopify:
still needs tests and cli entrypoint, but i wanted to get eyes on this again before i take it out of draft.
and i wanted to see if the failing check here in github was a flake or if it failed again.

go/vt/vtctl/grpcvtctldserver/server.go
```diff
@@ -77,6 +78,30 @@ func NewVtctldServer(ts *topo.Server) *VtctldServer {
 	}
 }
 
+// ApplyVSchema is part of the vtctlservicepb.VtctldServer interface.
+func (s *VtctldServer) ApplyVSchema(ctx context.Context, req *vtctldatapb.ApplyVSchemaRequest) (*vtctldatapb.ApplyVSchemaResponse, error) {
+	if _, err := s.ts.GetKeyspace(ctx, req.Keyspace); err != nil {
+		if strings.Contains(err.Error(), "node doesn't exist") {
```
ajm188, line 84: this should be `if topo.IsErrType(err, topo.NoNode)`

go/vt/vtctl/grpcvtctldserver/server.go
```diff
@@ -77,6 +78,30 @@ func NewVtctldServer(ts *topo.Server) *VtctldServer {
...
+
+	if !req.SkipRebuild {
+		if err := s.ts.RebuildSrvVSchema(ctx, nil); err != nil {
+			return nil, err
+		}
+	}
+	updatedVS, err := s.ts.GetVSchema(ctx, req.Keyspace)
+	if err != nil {
+		return nil, err
+	}
```
ajm188, line 101: we should wrap these errors to give some extra context of which step failed

ajm188 (commented):
This is looking great, just some minor feedback that I noticed now, but I think you're good to take it out of draft and work on testing/cli

acharisshopify:
i found some more time to circle back on this.
it still needs tests, but i got the CLI done.
i'll rebase to fix the proto conflicts.

go/cmd/vtctldclient/internal/command/vschemas.go
```diff
@@ -33,8 +39,117 @@ var (
...
+	VSchemaFile string
+	SQL         string
+	SQLFile     string
+	DryRun      bool
+	SkipRebuild bool
+	Cells       []string
+}{}
+
+func commandApplyVSchema(cmd *cobra.Command, args []string) error {
+	keyspace := cmd.Flags().Arg(0) // validated on the server-side
```
acharisshopify, line 61: happy to validate here as well. wdyt?

acharisshopify:
not sure why these are being marked as conflicting, i just rebased and recompiled protos. i'll take another look.

acharisshopify:
yep, another pull and rebase (and proto compile) seems to have gotten rid of the conflicts. thanks

go/vt/vtctl/grpcvtctldserver/server.go
```diff
@@ -128,6 +129,30 @@ func (s *VtctldServer) ApplyRoutingRules(ctx context.Context, req *vtctldatapb.A
 	return resp, nil
 }
 
+// ApplyVSchema is part of the vtctlservicepb.VtctldServer interface.
+func (s *VtctldServer) ApplyVSchema(ctx context.Context, req *vtctldatapb.ApplyVSchemaRequest) (*vtctldatapb.ApplyVSchemaResponse, error) {
+	if _, err := s.ts.GetKeyspace(ctx, req.Keyspace); err != nil {
+		if topo.IsErrType(err, topo.NoNode) {
+			return nil, fmt.Errorf("keyspace(%s) doesn't exist, check if the keyspace is initialized", req.Keyspace)
```
ajm188, line 136: I would make this an `vterrors.Wrapf(vtrpc.FAILED_PRECONDITION, ...)`

go/vt/vtctl/grpcvtctldserver/server.go
```diff
@@ -128,6 +129,30 @@ func (s *VtctldServer) ApplyRoutingRules(ctx context.Context, req *vtctldatapb.A
...
+	}
+
+	if !req.SkipRebuild {
+		if err := s.ts.RebuildSrvVSchema(ctx, req.Cells); err != nil {
+			return nil, fmt.Errorf("RebuildSrvVSchema = %w", err)
+		}
+	}
+	updatedVS, err := s.ts.GetVSchema(ctx, req.Keyspace)
+	if err != nil {
+		return nil, fmt.Errorf("GetKeyspace(%s) = %w", req.Keyspace, err)
```
ajm188, line 151: nit: wrong error string (probably copy-paste from above?)

go/cmd/vtctldclient/internal/command/vschemas.go
ajm188: imo we should validate as much server-side as possible, that way the implementation can be shared between vtctldserver and the legacy vtctl commands

go/cmd/vtctldclient/internal/command/vschemas.go
```diff
@@ -33,8 +39,117 @@ var (
 		Args: cobra.ExactArgs(1),
 		RunE: commandGetVSchema,
 	}
+	// ApplyVSchema makes an ApplyVSchema gRPC call to a vtctld.
+	ApplyVSchema = &cobra.Command{
```
ajm188, line 43: A net-new thing (that you could not have known!) I'm trying to introduce is:
* Set `DisableFlagsInUseLine: true` on every command. This will change the help from `{{ .Use }} [flags]` to just `{{ .Use }}` (which is better, because you include the actual named flags in the Use string).
* Include a `.Short` help text, which in most cases I just copy from the old vtctl command.

ajm188 (commented):
This is looking great. I think we just need to move that sqlparsing/topotools logic from the client to the server and we're good to go. Really sorry that didn't click for me earlier

go/cmd/vtctldclient/internal/command/vschemas.go
```diff
@@ -33,8 +39,117 @@ var (
...
+		} else {
+			schema = []byte(applyVSchemaOptions.VSchema)
+		}
+
+		vs = &vschemapb.Keyspace{}
+		err = json2.Unmarshal(schema, vs)
+		if err != nil {
+			return err
+		}
+	}
```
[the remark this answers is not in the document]
acharisshopify: interesting, i can definitely imagine someone calling the new api with only an sql string rather than a vschema object. so i can see where you're coming from.
i think i liked the idea of keeping the flag validation on the client side up above the call to FinishedParsing.
(though generally i do agree with what you say below about doing as much validation on the server side as possible.)

i do need to keep the file reading parts on the client side though, obviously. so it's not going to be a complete cut+paste job. but i'm definitely happy to shift the bulk of the logic up to the server side.

acharisshopify:
> (in general my view is that the client commands should do the bare minimum to convert CLI args into a proto message and then invoke the corresponding RPC).

do you want the dry run logic on the server side as well, then?

go/vt/vtctl/grpcvtctldserver/server.go
acharisshopify: not totally sure what you meant here.
Wrap/Wrapf don't take an error code like failed precondition.
New and Errorf do take a code, but those are for making a new error, whereas here i already have an error.

acharisshopify:
with the shift of that logic, i've got to expand the testing a bit i think.

go/vt/vtctl/grpcvtctldserver/server_test.go
```diff
@@ -1132,7 +1287,7 @@ func TestDeleteCellsAlias(t *testing.T) {
 		name      string
 		ts        *topo.Server
 		setup     func(ts *topo.Server) error
-		req       *vtctldata.DeleteCellsAliasRequest
+		req       *vtctldatapb.DeleteCellsAliasRequest
```
ajm188, line 1290: thanks for catching this one, i must have missed this in another change

ajm188 (approved):
some minor changes, then lgtm! thanks for doing this!

go/cmd/vtctldclient/internal/command/vschemas.go
```diff
@@ -58,5 +142,14 @@ func commandGetVSchema(cmd *cobra.Command, args []string) error {
 }
 
 func init() {
+	ApplyVSchema.Flags().StringVar(&applyVSchemaOptions.VSchema, "vschema", "", "VSchema")
+	ApplyVSchema.Flags().StringVar(&applyVSchemaOptions.VSchemaFile, "vschema_file", "", "VSchema File")
```
ajm188, line 146: we're using dashes instead of underscores in the new client

go/cmd/vtctldclient/internal/command/vschemas.go
```diff
@@ -58,5 +142,14 @@ func commandGetVSchema(cmd *cobra.Command, args []string) error {
...
+	ApplyVSchema.Flags().StringVar(&applyVSchemaOptions.SQL, "sql", "", "A VSchema DDL SQL statement, e.g. `alter table t add vindex hash(id)`")
+	ApplyVSchema.Flags().StringVar(&applyVSchemaOptions.SQLFile, "sql_file", "", "A file containing VSchema DDL SQL")
```
ajm188, line 148: ditto

go/cmd/vtctldclient/internal/command/vschemas.go
```diff
@@ -58,5 +142,14 @@ func commandGetVSchema(cmd *cobra.Command, args []string) error {
...
+	ApplyVSchema.Flags().BoolVar(&applyVSchemaOptions.DryRun, "dry-run", false, "If set, do not save the altered vschema, simply echo to console.")
+	ApplyVSchema.Flags().BoolVar(&applyVSchemaOptions.SkipRebuild, "skip_rebuild", false, "If set, do no rebuild the SrvSchema objects.")
```
ajm188, line 150: ditto

ajm188 (commented):
whoops, meant to include these comments in my last review

acharisshopify:
@ajm188 alright, tests passing. are we good to go?

systay:
Since we are creating the release notes automatically from PRs, I was wondering if we could tweak the title a bit? Should the WIP demarcation stay around?

ajm188:
Nope! I updated it just now, let me know if that's good for the release notes tooling or if I should tweak it differently

acharisshopify:
thanks!

PR #8114, opened 2021-05-12
[9.0] restore: check disable_active_reparents properly before waiting for position update

### LINKED ISSUE #8108: [9.0.1] Backup/restore operations do not respect -disable_active_reparents flag
#7657 is broken on 9.0.1.
The fix in #7576 was backported to 9.0 branch as #8063. However, that fix is incorrect.
The correct fix is in #7703 and needs to be backported.

deepthi:
## Description
Backport of #7703 
Fixes #8108 

## Checklist


harshit-gangal (approved):
Not sure what the changes are.
Approving because it is a backport and CI is green.

PR #8116, opened 2021-05-13
Support for vtgate -enable_direct_ddl flag

shlomi-noach:
## Description

We add a `-enable_direct_ddl` (bool) flag to `vtgate`. The default is `true`.
It is now possible to disable direct DDL via `vtgate.` If you run `vtgate -enable_direct_ddl=false`, then, for example:

```
$ mysql
Welcome to the MySQL monitor.  Commands end with ; or \g.
Your MySQL connection id is 1

mysql> select @@ddl_strategy;
+----------------+
| @@ddl_strategy |
+----------------+
| direct         |
+----------------+
1 row in set (0.00 sec)

mysql> create table my_test(id int primary key);
ERROR 1105 (HY000): direct DDL is disabled
```

This variable only controls `direct` (non-online) DDLs. As reminder, online DDL can be controlled via `-enable_online_ddl`. It is possible to enable one and disable another, enable both, or disable both.

## Checklist

cc @piki @deepthi @mavenraven

### DIFF
--- a/go/vt/schema/online_ddl.go
+++ b/go/vt/schema/online_ddl.go
@@ -41,6 +41,8 @@ var (
 )
 
 var (
+	// ErrDirectDDLDisabled is returned when direct DDL is disabled, and a user attempts to run a DDL statement
+	ErrDirectDDLDisabled = errors.New("direct DDL is disabled")
 	// ErrOnlineDDLDisabled is returned when online DDL is disabled, and a user attempts to run an online DDL operation (submit, review, control)
 	ErrOnlineDDLDisabled = errors.New("online DDL is disabled")
 	ErrForeignKeyFound   = errors.New("Foreign key found")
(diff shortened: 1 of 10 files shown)


shlomi-noach:
This change only affects `vtgate`. Schema changes via `vtctl ApplySchema` are unaffected by this flag.

systay:
Would be nice if you could add a test

shlomi-noach:
re: test, I agree; but the only way I can see it is if I spawn a dedicated new `endtoend` test just for this, becaus eit entails starting `vtgate` with the new flag. Any thoughts?

systay:
well, how about the planner not directly depending on the flag, and have this value be given to the planner from the outside. 
this makes the dependency explicit and makes it easier to test. WDYT?

systay (approved):

PR #8043, opened 2021-05-05
VReplication Online DDL: fix classification of virtual columns

shlomi-noach:
Signed-off-by: Shlomi Noach <2607934+shlomi-noach@users.noreply.github.com>

## Description

A small fix. While reviewing some work @fatih was doing  I realized that the way OnlineDDL/Vrepl identifies virtual columns is incorrect. Per documentation: https://dev.mysql.com/doc/refman/8.0/en/show-columns.html, a virtual columns is identified by either `VIRTUAL GENERATED` or `VIRTUAL STORED`.

The existing code only identifies `VIRTUAL GENERATED` columns; this PR identifies both types of virtual columns.

## Related Issue(s)

- tracking: #6926 

## Checklist
- [ ] Tests were added or are not required
- [ ] Documentation was added or is not required

PR #8129, opened 2021-05-14
VReplication: ignore generated columns in workflows

rohit-nayak-ps:
## Description

Currently generated columns are not supported during a resharding workflow. MySQL does not allow values to be provided for generated columns either for insert or update. Today, generated columns are treated as regular columns and hence they attempt to populate such columns. 

This PR identifies these columns by looking at the `Extra` column in `information_schema.columns` and drops them from insert/update statements on the target. 

*_Note:_* We continue to send these columns as part of the vstreamer events so the downstream consumers will see the actual values. We need to do this since vstreams might be consumed by CDC pipelines which should be expected to recompute the generated column data.

## Checklist


go/vt/vttablet/tabletmanager/vreplication/table_plan_builder.go
```diff
@@ -745,6 +771,17 @@ func (tpb *tablePlanBuilder) generatePKConstraint(buf *sqlparser.TrackedBuffer,
 	buf.WriteString(")")
 }
 
+func (tpb *tablePlanBuilder) isColumnGenerated(col sqlparser.ColIdent) bool {
+	isGenerated := false
+	for _, colInfo := range tpb.colInfos {
+		if col.EqualString(colInfo.Name) && colInfo.IsGenerated {
+			isGenerated = true
+			break
```
shlomi-noach, line 779: just my coding style opinion: `return true`

go/vt/vttablet/tabletmanager/vreplication/table_plan_builder.go
```diff
@@ -745,6 +771,17 @@ func (tpb *tablePlanBuilder) generatePKConstraint(buf *sqlparser.TrackedBuffer,
...
+		}
+	}
+	return isGenerated
```
shlomi-noach, line 782: just my coding style opinion: `return false`

go/vt/vttablet/tabletmanager/vreplication/vreplicator.go
```diff
@@ -224,22 +224,24 @@ func (vr *vreplicator) replicate(ctx context.Context) error {
...
 }
 
-// PrimaryKeyInfo is used to store charset and collation for primary keys where applicable
-type PrimaryKeyInfo struct {
-	Name       string
-	CharSet    string
-	Collation  string
-	DataType   string
-	ColumnType string
+// ColumnInfo is used to store charset and collation for primary keys where applicable
```
shlomi-noach, line 227: change "for primary keys"

go/vt/vttablet/tabletmanager/vreplication/vreplicator.go
```diff
@@ -224,22 +224,24 @@ func (vr *vreplicator) replicate(ctx context.Context) error {
...
+type ColumnInfo struct {
+	Name        string
+	CharSet     string
+	Collation   string
+	DataType    string
+	ColumnType  string
+	IsPK        bool
+	IsGenerated bool
```
shlomi-noach, line 235: this becomes very parallel to https://github.com/vitessio/vitess/blob/50365bb24460e919e9875a5fd204e3714a94066d/go/vt/vttablet/onlineddl/vrepl/types.go#L35-L58. I see we are solving the same kind of problems in different places. I'll defer to yours anyway.

go/vt/vttablet/tabletmanager/vreplication/vreplicator.go
```diff
@@ -257,47 +259,56 @@ func (vr *vreplicator) buildPkInfoMap(ctx context.Context) (map[string][]*Primar
...
-				return nil, fmt.Errorf("no dataType/columnType found in information_schema.columns for table %s, column %s", td.Name, pk)
+				return nil, fmt.Errorf("no dataType/columnType found in information_schema.columns for table %s, column %s", td.Name, columnName)
+			}
+			for _, pk := range pks {
+				if columnName == pk {
+					isPK = true
+				}
+			}
+			extra := row[5].ToString()
+			if strings.Contains(strings.ToLower(extra), "generated") {
```
shlomi-noach, line 296: also check for `"virtual"` because it could be `"virtual stored"`. See https://github.com/vitessio/vitess/pull/8043/files

shlomi-noach (commented):
initial comments, not thorough review.

go/vt/vttablet/tabletmanager/vreplication/vreplicator.go
```diff
@@ -78,8 +78,8 @@ type vreplicator struct {
 	state           string
 	stats           *binlogplayer.Stats
 	// mysqld is used to fetch the local schema.
-	mysqld    mysqlctl.MysqlDaemon
-	pkInfoMap map[string][]*PrimaryKeyInfo
+	mysqld     mysqlctl.MysqlDaemon
+	colInfoMap map[string][]*ColumnInfo
```
shlomi-noach, line 82: `map[string][]*ColumnInfo` confuses me. Is the key a table-name? Let's consider making a new `type` to simplify the syntax.

shlomi-noach (approved):

go/vt/vttablet/tabletmanager/vreplication/vreplicator.go
systay: I sometimes use type aliases for this. 

```golang
type tableNameStr = string
colInfoMap := map[tableNameStr][]*ColumnInfo{}
``

PR #8134, opened 2021-05-17
vtctl.generateShardRanges -> key.GenerateShardRanges

guidoiaquinti:
Signed-off-by: Guido Iaquinti <giaquinti@slack-corp.com>

## Description
Export `vtctl.generateShardRanges` to `key.GenerateShardRanges` so it can be used also in other places (I'm not sure `key` is the best package where to put it but I didn't find a better one).
 

## Checklist


go/vt/key/key.go
```diff
@@ -346,3 +347,74 @@ var krRegexp = regexp.MustCompile(`^[0-9a-fA-F]*-[0-9a-fA-F]*$`)
...
+	case shards <= 0:
+		return nil, errors.New("shards must be greater than zero")
+	case shards <= 256:
+		format = "%02x"
+		maxShards = 256
+	case shards <= 65536:
+		format = "%04x"
+		maxShards = 65536
+	default:
+		return nil, errors.New("this function does not support more than 65336 shards in a single keyspace")
```
guidoiaquinti, line 366: The error has been renamed

go/vt/key/key_test.go
```diff
@@ -579,3 +579,75 @@ func TestIsKeyRange(t *testing.T) {
...
+			false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := GenerateShardRanges(tt.args.shards)
+			if tt.wantErr {
+				assert.Error(t, err)
```
guidoiaquinti, line 636: Before it was `require.Error(t, err)` but I don't think this was really needed (and we can avoid importing `require`)

go/vt/key/key_test.go
```diff
@@ -579,3 +579,75 @@ func TestIsKeyRange(t *testing.T) {
...
+			} else {
+				assert.NoError(t, err)
```
guidoiaquinti, line 638: Ditto

go/vt/key/key_test.go
ajm188: We actually want the behavior of `require` in the cases where `tt.wantErr=false` in that it will stop execution of the test if the assertion fails (this is the main difference between the `assert` and `require` packages. It makes the intent of the test assertion much clearer in my view

ajm188 (commented):
Just the one comment about `require` and then lgtm

go/vt/key/key_test.go
guidoiaquinti: > Package require implements the same assertions as the `assert` package but stops test execution when a test fails.
I don't see a big difference between the two but if you think it's cleaner I revert my change (see next commit).

go/vt/key/key_test.go
```diff
@@ -633,9 +634,9 @@ func TestGenerateShardRanges(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			got, err := GenerateShardRanges(tt.args.shards)
 			if tt.wantErr {
-				assert.Error(t, err)
+				require.Error(t, err)
 			} else {
```
ajm188, line 638: you no longer need the `else` here (this is the benefit of `require`)

this should read:

```go
if tt.wantErr {
    assert.Error(t, err)
    return
}

require.NoError(t, err)
assert.Equal(t, got, tt.want)
```

guidoiaquinti: Fair but I've blindly copy pasted what was there before:
https://github.com/vitessio/vitess/blob/7c532a533d40d049d81c4cc9676ff921064e5700/go/vt/vtctl/vtctl_test.go#L59-L70

ajm188 (approved):

go/vt/key/key_test.go
ajm188: 🤦 haha sorry, and thank you for cleaning up my original mistake

PR #8139, opened 2021-05-17
Improve ScatterErrorsAsWarnings functionality

harshit-gangal:
## Description
This PR improves `ScatterErrorsAsWarnings` by doing three things:

- Make sure that it works in both OLAP and OLTP mode. It was previously not working in OLAP mode
- When all shards fail, the query will still fail. `ScatterErrorsAsWarnings` should only kick in when a single or a few shards fail.
- Adds better end to end tests to assert that everything works as expected
- Support `SHOW WARNINGS` in OLAP mode

**Note: this is a breaking change. Queries that failed on all shards were not returning an error earlier, but after this change, scatter errors are turned into warnings only if at least one shard was successful.**

## Checklist


systay (approved):

harshit-gangal:
Marked as breaking change. As earlier all the errors coming from executing the plan will be consumed as error, now at least 1 shard should return results then warnings are logged otherwise throws an error.

go/vt/vtgate/engine/route.go
```diff
@@ -375,8 +386,9 @@ func (route *Route) mergeSort(vcursor VCursor, bindVars map[string]*querypb.Bind
 		})
 	}
 	ms := MergeSort{
-		Primitives: prims,
-		OrderBy:    route.OrderBy,
+		Primitives:              prims,
+		OrderBy:                 route.OrderBy,
+		ScatterErrorsAsWarnings: route.ScatterErrorsAsWarnings,
 	}
 	err := ms.StreamExecute(vcursor, bindVars, wantfields, func(qr *sqltypes.Result) error {
```
harshit-gangal, line 393: for merge sort as well same/similar check is needed 
```
if !route.ScatterErrorsAsWarnings || len(errs) == len(rss) {
	return vterrors.Aggregate(errs)
}
```

go/vt/vtgate/engine/session_primitive.go
```diff
@@ -0,0 +1,83 @@
...
+	qr, err := s.action(vcursor.Session())
+	if err != nil {
+		return err
+	}
+	return callback(qr)
+}
+
+// GetFields implements the Primitive interface
+func (s *SessionPrimitive) GetFields(_ VCursor, _ map[string]*querypb.BindVariable) (*sqltypes.Result, error) {
+	panic("implement me")
```
harshit-gangal, line 75: we should change this to `internal`  error

go/vt/vtgate/executor_select_test.go
```diff
@@ -1131,12 +1131,113 @@ func TestSelectScatterPartial(t *testing.T) {
 	conns[6].MustFailCodes[vtrpcpb.Code_RESOURCE_EXHAUSTED] = 1000
 	conns[7].MustFailCodes[vtrpcpb.Code_RESOURCE_EXHAUSTED] = 1000
 
-	results, err = executorExec(executor, "select /*vt+ SCATTER_ERRORS_AS_WARNINGS=1 */ id from user", nil)
+	_, err = executorExec(executor, "select /*vt+ SCATTER_ERRORS_AS_WARNINGS=1 */ id from user", nil)
+	require.Error(t, err)
+
+	testQueryLog(t, logChan, "TestExecute", "SELECT", "select /*vt+ SCATTER_ERRORS_AS_WARNINGS=1 */ id from user", 8)
```
harshit-gangal, line 1137: add this test with order by query as well.

go/vt/vtgate/executor_select_test.go
```diff
@@ -1131,12 +1131,113 @@ func TestSelectScatterPartial(t *testing.T) {
...
+	conns[1].MustFailCodes[vtrpcpb.Code_RESOURCE_EXHAUSTED] = 1000
+	conns[3].MustFailCodes[vtrpcpb.Code_RESOURCE_EXHAUSTED] = 1000
+	conns[4].MustFailCodes[vtrpcpb.Code_RESOURCE_EXHAUSTED] = 1000
+	conns[5].MustFailCodes[vtrpcpb.Code_RESOURCE_EXHAUSTED] = 1000
+	conns[6].MustFailCodes[vtrpcpb.Code_RESOURCE_EXHAUSTED] = 1000
+	conns[7].MustFailCodes[vtrpcpb.Code_RESOURCE_EXHAUSTED] = 1000
+
+	_, err = executorStream(executor, "select /*vt+ SCATTER_ERRORS_AS_WARNINGS=1 */ id from user")
+	require.Error(t, err)
+	testQueryLog(t, logChan, "TestExecuteStream", "SELECT", "select /*vt+ SCATTER_ERRORS_AS_WARNINGS=1 */ id from user", 8)
```
harshit-gangal, line 1185: add this test with order by query as well.

PR #8142, opened 2021-05-18
Switchwrites: error if no tablets available on target for reverse replication

rohit-nayak-ps:
## Description
This PR addresses an edge case while switching writes. 
* VReplication Workflow is setup with `-tablet_types` selected as `RDONLY`
* Source keyspace has RDONLY tablets and the forward replication is completed
* Reads are switched to the target keyspace
* `SwitchWrites` is called next to switch all traffic to the target, with the `-reverse_replication true` flag
* However there are no RDONLY tablets on the target, so the reverse replication workflow never starts
* User only finds out after the writes are switched. In the meanwhile due to problems on the target 
the user tries to rollback by reversing the writes. 
* This fails with a timeout waiting for the source to catchup with the target

This PR adds a check on invoking SwitchWrites that there are tablets available to source data from the target keyspace if reverse replication is enabled. Note that this only checks for tablets being available when SwitchWrites is called. If, in this usecase, the RDONLY tablets disappear in the future then the workflow will stop replicating resulting in the same problem as outlined above. 

## Checklist

### DIFF
--- a/go/vt/discovery/tablet_picker.go
+++ b/go/vt/discovery/tablet_picker.go
@@ -107,7 +107,7 @@ func (tp *TabletPicker) PickForStreaming(ctx context.Context) (*topodatapb.Table
 			return nil, vterrors.Errorf(vtrpcpb.Code_CANCELED, "context has expired")
 		default:
 		}
-		candidates := tp.getMatchingTablets(ctx)
+		candidates := tp.GetMatchingTablets(ctx)
 
 		if len(candidates) == 0 {
 			// if no candidates were found, sleep and try again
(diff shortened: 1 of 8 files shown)


sougou (commented):
This is looking good. I feel like we should do this check while creating any stream, for all workflows.

rohit-nayak-ps:
> This is looking good. I feel like we should do this check while creating any stream, for all workflows.

The case fixed here is the one that caused the user issue recently where they were not able to reverse after detecting an issue after cutover causing an unexpected outage.

I agree we should validate for available tablets while creating streams as well. For those use cases though, the workflow would just not start and the error would be visible in `Workflow Show` or by looking at the vreplication table. 

Adding other validations will involve some time writing the tests for it, so I would like to move ahead with this for now and implement those later.

deepthi (approved):

PR #8147, opened 2021-05-19
CI: fail PR if it does not have required labels

### LINKED ISSUE #8146: Add CI job to validate PR labels
As of recently, we label each PR with one `Component: ...` label and one `Type: ...` label. We want to ensure&require that these labels exists on every PR.

Suggestion: add a CI job that fails when such labels do not exist.

Caveat: say a PR does not have said labels, and CI fails. Someone adds the labels. We need to trigger CI. That's doable, but I'm thinking can we only trigger CI for _those_ labels, and avoid retriggering CI any time any label is added/removed?

shlomi-noach:
## Description

Fixes #8146 

Adding a ci job to fail a PR if it does not have at least one `Component: ...` label and at least one `Type: ...` label.

## Related Issue(s)

- https://github.com/vitessio/vitess/issues/8146

## Checklist

### DIFF
--- a/.github/workflows/pr-labels.yml
+++ b/.github/workflows/pr-labels.yml
@@ -23,7 +23,7 @@ jobs:
         LABELS_JSON="/tmp/labels.json"
 
         # Get labels for this pull request
-        curl \
+        curl -s \
           -H 'authorization: Bearer ${{ secrets.GITHUB_TOKEN }}' \
           -H "Accept: application/vnd.github.v3+json" \
           -H "Content-type: application/json" \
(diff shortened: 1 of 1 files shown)


shlomi-noach:
Right now this PR fails because it does not have a `Type: ...` label: https://github.com/vitessio/vitess/runs/2617745882?check_suite_focus=true

Screenshot from 2021-05-19 10-24-46

So i'll now add a `Type: ...` label and see if CI passes. Note: I will have to trigger the CI manually because the workflow file does not exist in the default branch (`master`) because that's how it works; when the PR is merged, then CI gets retriggered upon any label change, automatically.

shlomi-noach:
yay, as expected, CI passes: https://github.com/vitessio/vitess/runs/2617806212?check_suite_focus=true

Screenshot from 2021-05-19 10-27-03

systay (approved):
❤️

PR #8107, opened 2021-05-12
[tm_state] updateLocked should re-populate local metadata tables to reflect promotion rule changes

ajm188:
Signed-off-by: Andrew Mason <amason@slack-corp.com>

## Description

TODO.

This means we call `PopulateMetadataTables` on:
* `(*tmState).ChangeTabletType`
* `(*tmState).RefreshFromTopoInfo`
* `(*tmState).RefreshFromTopo` (via `RefreshFromTopoInfo`)
* `(*tmState).Open`

### See it in action

```
❯ vtctldclient --server "localhost:15999" GetTablets
zone1-0000000100 commerce 0 master SFO-M-AMASON02:15100 SFO-M-AMASON02:17100 [] 2021-05-12T15:48:34Z
zone1-0000000101 commerce 0 replica SFO-M-AMASON02:15101 SFO-M-AMASON02:17101 [] <null>
zone1-0000000102 commerce 0 rdonly SFO-M-AMASON02:15102 SFO-M-AMASON02:17102 [] <null>

# on zone1-101
mysql> select * from _vt.local_metadata;
+---------------+------------------+-------------+
| name          | value            | db_name     |
+---------------+------------------+-------------+
| Alias         | zone1-0000000101 | vt_commerce |
| ClusterAlias  | commerce.0       | vt_commerce |
| DataCenter    | zone1            | vt_commerce |
| PromotionRule | neutral          | vt_commerce |
+---------------+------------------+-------------+
4 rows in set (0.00 sec)

# now, the tablet type change
❯ vtctldclient --server "localhost:15999" ChangeTabletType zone1-0000000101 rdonly
- zone1-0000000101 commerce 0 replica SFO-M-AMASON02:15101 SFO-M-AMASON02:17101 [] <null>
+ zone1-0000000101 commerce 0 rdonly SFO-M-AMASON02:15101 SFO-M-AMASON02:17101 [] <null>

mysql> mysql> select * from _vt.local_metadata;
+---------------+------------------+-------------+
| name          | value            | db_name     |
+---------------+------------------+-------------+
| Alias         | zone1-0000000101 | vt_commerce |
| ClusterAlias  | commerce.0       | vt_commerce |
| DataCenter    | zone1            | vt_commerce |
| PromotionRule | must_not         | vt_commerce |
+---------------+------------------+-------------+
4 rows in set (0.01 sec)
```

## Checklist
- [ ] Tests were added or are not required
- [ ] Documentation was added or is not required

PR #8152, opened 2021-05-20
[mysqlctl] Restructure `MetadataManager` to reduce public API surface area

ajm188:

## Description

This is a follow-up PR to address some comments on #8107, just to address the public API surface of `mysqlctl.MetadataManager`.

I need to think through a bit more about how to work through the behaviors of `-init_populate_metadata` and `-restore_from_backup` to simplify (and hopefully clarify) the checks we currently do in `populateLocalMetadatLocked`, so I'm going to do that in a second PR.

## Related Issue(s)

- #8107 

## Checklist
- [x] Tests were added or are not required -- N/A


deepthi (approved):
LGTM

PR #8154, opened 2021-05-20
[Java] JDBC mysql driver test

### LINKED ISSUE #7916: jdbc:mysql @replica is not working
When we were trying to use third-party tool TDDL to automatically separate READ transactions and WRITE transactions, to dispatch WRITE to master tablet only and READ to other replica tablets, we found the jdbc:mysql connection on replica does not work. For example: "jdbc:mysql://192.168.20.231:15306/dbname@replica" is actually connecting to the master tablet.

harshit-gangal:
## Description

Added MySQL JDBC driver to test it with Vitess and also shows that @<tablet_type> works in the connection string.

## Related Issue(s)

Closes #7916 

## Checklist

### DIFF
--- a/java/example/src/main/java/io/vitess/example/MysqlJDBCExample.java
+++ b/java/example/src/main/java/io/vitess/example/MysqlJDBCExample.java
@@ -1,0 +1,179 @@
+/*
+ * Copyright 2021 The Vitess Authors.
+
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+
+ *     http://www.apache.org/licenses/LICENSE-2.0
+
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


java/example/src/main/java/io/vitess/example/MysqlJDBCExample.java
```diff
@@ -0,0 +1,179 @@
...
+      if (null != updateCounts) {
+        evalBatchResult(updateCounts);
+      }
+    }
+
+    try (Statement stmt = conn.createStatement()) {
+      Instant timeCreated = Instant.now();
+      int page = rand.nextInt(100) + 1;
+      System.out.println("Page selected for all dml operation: " + page);
+      stmt.addBatch("INSERT INTO messages (page,time_created_ns,message) VALUES (" + page + ","
```
systay, line 147: nit: `String.Format` is your friend

systay (approved):

java/example/src/main/java/io/vitess/example/MysqlJDBCExample.java
harshit-gangal: I want to spend less time doing java.

PR #8163, opened 2021-05-20
[vtadmin-api] Update GetTablet to use alias instead of hostname

doeg:

## Description

Since tablet hostnames alone aren't guaranteed to be unique even in the same cluster (as with the local example), this updates the `/api/tablet/{tablet}` and `/experimental/tablet/{tablet}/debug/vars` endpoints to use tablet alias as an identifier instead. 

No changes required to vtadmin-web for this; it doesn't use these endpoints yet. 

An example:

```bash
$ curl "http://localhost:14200/api/tablet/zone1-102" | jq .
{
  "result": {
    "cluster": {
      "id": "id1",
      "name": "cluster1"
    },
    "tablet": {
      "alias": {
        "cell": "zone1",
        "uid": 102
      },
      "hostname": "a4c88ebf1fe6",
      "keyspace": "commerce",
      "shard": "0",
      "type": 3
    },
    "state": 1
  },
  "ok": true
}
```

...and an example of an error when an invalid alias is given (which the existing tests already cover): 

```bash
$ curl "http://localhost:14200/api/tablet/does-not-exist" | jq .
{
  "error": {
    "message": "failed to parse tablet_alias does-not-exist: invalid tablet alias: 'does-not-exist', expecting format: '<cell>-<uid>'",
    "code": "unknown"
  },
  "ok": false
}
```

## Related Issue(s)

N/A

## Checklist

## Deployment Notes

N/A


rafael (approved):
Makes sense to me. LGTM

ajm188 (approved):

Issue #6041, opened 2020-04-09
vttablet fails to parse cell names with dashes


gedgar:
### Overview of the Issue
vttablet fails to parse tablet-aliases with more than one `-` in the name, as it's expecting the format `<cell>-<uid>`. 

I'm not sure if dashes break anything else; we should either fix the parsing of tablet-aliases or disallow adding cells with dashes in the name. 

#### Reproduction Steps
Add a cell with a dash in the name, eg `commerce-zone1`.
Try starting a tablet in that cell:
`F0409 16:05:14.759550   11595 vttablet.go:69] failed to parse -tablet-path: invalid tablet alias: 'commerce-zone1-0000000100', expecting format: '<cell>-<uid>'`

PR #8167, opened 2021-05-21
Correctly parse cell names with dashes in tablet aliases

hkdsun:
Signed-off-by: Hormoz Kheradmand <hormoz.kheradmand@shopify.com>

## Description

Closes https://github.com/vitessio/vitess/issues/6041

vttablet fails to parse tablet-aliases with more than one `-` in the cell name.

This is especially problematic because cells are often named after GCP/AWS zones, which are often valid DNS-1123 labels (consists of lower case alphanumeric characters or '-', and must start and end with an alphanumeric character. [see similar validation in the `kubernetes/kubernetes` repository](https://github.com/kubernetes/kubernetes/blob/7f23a743e8c23ac6489340bbb34fa6f1d392db9d/staging/src/k8s.io/apimachinery/pkg/util/validation/validation.go#L107-L124))

Since this parser has been out in the wild for a while, I am not sure how strict we want to make the format. I chose to include the capital alphabetic characters in addition to `.`, `_`, and of course `-`.

I am happy to change the expected format to be any more or any less strict, as maintainers see fit.

## Checklist

## Deployment Notes

There are chances of invalid cell names out in the wild the more strict we make the expected format.

@deepthi @gedgar @harshit-gangal 
cc @jeremycole @acharis

### DIFF
--- a/go/vt/topo/topoproto/tablet.go
+++ b/go/vt/topo/topoproto/tablet.go
@@ -21,6 +21,7 @@ package topoproto
 import (
 	"fmt"
 	"net"
+	"regexp"
 	"sort"
 	"strconv"
 	"strings"
@@ -75,19 +76,23 @@ func TabletAliasUIDStr(ta *topodatapb.TabletAlias) string {
 	return fmt.Sprintf("%010d", ta.Uid)
 }
 
+const tabletAliasFormat = "^(?P<cell>[-_.a-zA-Z0-9]+)-(?P<uid>[0-9]+)$"
+
+var tabletAliasRegexp = regexp.MustCompile(tabletAliasFormat)
+
 // ParseTabletAlias returns a TabletAlias for the input string,
 // of the form <cell>-<uid>
 func ParseTabletAlias(aliasStr string) (*topodatapb.TabletAlias, error) {
-	nameParts := strings.Split(aliasStr, "-")
-	if len(nameParts) != 2 {
-		return nil, fmt.Errorf("invalid tablet alias: '%s', expecting format: '<cell>-<uid>'", aliasStr)
+	nameParts := tabletAliasRegexp.FindStringSubmatch(aliasStr)
+	if len(nameParts) != 3 {
+		return nil, fmt.Errorf("invalid tablet alias: '%s', expecting format: '%s'", aliasStr, tabletAliasFormat)
 	}
-	uid, err := ParseUID(nameParts[1])
+	uid, err := ParseUID(nameParts[tabletAliasRegexp.SubexpIndex("uid")])
 	if err != nil {
 		return nil, vterrors.Wrapf(err, "invalid tablet uid in alias '%s'", aliasStr)
 	}
 	return &topodatapb.TabletAlias{
-		Cell: nameParts[0],
+		Cell: nameParts[tabletAliasRegexp.SubexpIndex("cell")],
 		Uid:  uid,
 	}, nil
 }
--- a/go/vt/topo/topoproto/tablet_test.go
+++ b/go/vt/topo/topoproto/tablet_test.go
@@ -1,0 +1,72 @@
+/*
+Copyright 2019 The Vitess Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+		http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package topoproto
+
+import (
+	"fmt"
+	"testing"
+
+	topodatapb "vitess.io/vitess/go/vt/proto/topodata"
+)
+
+func TestParseTabletAlias(t *testing.T) {
+	for aliasStr, expectedAlias := range map[string]*topodatapb.TabletAlias{
+		// valid cases
+		"cell1-42":                         {Cell: "cell1", Uid: 42},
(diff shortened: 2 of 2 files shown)


go/vt/topo/topoproto/tablet.go
```diff
@@ -75,19 +76,23 @@ func TabletAliasUIDStr(ta *topodatapb.TabletAlias) string {
 	return fmt.Sprintf("%010d", ta.Uid)
 }
 
+const tabletAliasFormat string = "([-_.a-zA-Z0-9]+)-([0-9]+)"
```
jeremycole, line 79: This could be made a little nicer with named capture groups and `SubexpIndex`:

```suggestion
const tabletAliasFormat string = "(?P<cell>[-_.a-zA-Z0-9]+)-(?P<uid>[0-9]+)"
```

go/vt/topo/topoproto/tablet.go
```diff
@@ -75,19 +76,23 @@ func TabletAliasUIDStr(ta *topodatapb.TabletAlias) string {
...
 // of the form <cell>-<uid>
 func ParseTabletAlias(aliasStr string) (*topodatapb.TabletAlias, error) {
-	nameParts := strings.Split(aliasStr, "-")
-	if len(nameParts) != 2 {
+	nameParts := tabletAliasRegexp.FindStringSubmatch(aliasStr)
+	if len(nameParts) != 3 {
 		return nil, fmt.Errorf("invalid tablet alias: '%s', expecting format: '<cell>-<uid>'", aliasStr)
 	}
-	uid, err := ParseUID(nameParts[1])
+	uid, err := ParseUID(nameParts[2])
```
jeremycole, line 90:
```suggestion
	uid, err := ParseUID(nameParts[tabletAliasRegexp.SubexpIndex("uid")])
```

go/vt/topo/topoproto/tablet.go
```diff
@@ -75,19 +76,23 @@ func TabletAliasUIDStr(ta *topodatapb.TabletAlias) string {
...
 	if err != nil {
 		return nil, vterrors.Wrapf(err, "invalid tablet uid in alias '%s'", aliasStr)
 	}
 	return &topodatapb.TabletAlias{
-		Cell: nameParts[0],
+		Cell: nameParts[1],
```
jeremycole, line 95:
```suggestion
		Cell: nameParts[tabletAliasRegexp.SubexpIndex("cell")],
```

go/vt/topo/topoproto/tablet.go
ajm188: In addition, is there a reason you don't add the anchors here instead of adding them inline when you create `tabletAliasRegexp`?

rafael:
Hello @hkdsun , @jeremycole ! Good to see you in Vitess land 🎉  

How cell is inferred in tablets is something that has bugged for a long time. Thank you for fixing this. In my view, this will work and after addressing that outstanding comments we proceed with the approach. 

Having said this, there is an alternative approach that we could consider that I would like us to consider. All other vitess components are explicit about which `cell` they are registering to. I can't think of a good reason why that shouldn't be the case here as well. Curious if @deepthi or @sougou remember something that I might be missing. If we can make it explicit, I think we can do something as follows:

1) Add two flags to the tablet (`-cell` and `-uid`).
2) If those attributes are not set, default to existent behavior.
3) If they are set, they take precedence and we don't have to rely on processing the tablet_path to infer what portion is the cell and which one is the uid. 
4) Eventually deprecate `tablet_path`. 

In the long term, this might be a better approach. Of course, that will increase the scope of what you are actually trying to fix. It will not only mean updating o tablet, but other places where this assumption is also being made (e.g some of the vtctld operation like reparents). 

I think we merge this, but I would be curious of people thoughts on the alternate approach.

go/vt/topo/topoproto/tablet.go
hkdsun: The original intention was to print out the `tabletAliasFormat` in the error message, in which case dropping the anchors is more friendly to human eyes. I do realize I never executed my plan, hence the confusion 😄 


If everyone thinks this is a welcome change, I'm happy to commit it.

* old error message: `invalid tablet alias: '%s', expecting format: '<cell>-<uid>'`
* new error message `invalid tablet alias: '%s', expecting format: '(?P<cell>[-_.a-zA-Z0-9]+)-(?P<uid>[0-9]+)'`

ajm188: Ah that thinking makes sense. My one concern is that if we're matching on an anchored regex, and we produce an error that suggests we're not that that could lead to confusion. For an (admittedly trivial) example `invalid tablet alias: ',zone1-100', expecting format '(?P<cell>[-_.a-zA-Z0-9+)-(?P<uid>[0-9]+)'` would still require me to read the code to confirm the regexp is anchored (although in this example it wouldn't be too hard to guess that), whereas if we include the anchors then that question is already answered for the user.

Either way, including the regexp in any form in the error string seems like a great idea to me!

hkdsun: Change committed! ✅

hkdsun:
Thanks for the comments everyone and the excellent context, Rafael.

Using separate `-uid` and `-cell` arguments makes a lot of sense to me, though I can only commit to this simple fix at the moment for the stage of the project we are in – happy to reconsider in the future.

I would still love to have this PR be merged if we agree on it as a good short-term fix.

deepthi:
> Thanks for the comments everyone and the excellent context, Rafael.
> 
> Using separate `-uid` and `-cell` arguments makes a lot of sense to me, though I can only commit to this simple fix at the moment for the stage of the project we are in – happy to reconsider in the future.
> 
> I would still love to have this PR be merged if we agree on it as a good short-term fix.

I like the long-term approach proposed by @rafael but understand your position too. We can merge this as-is and revisit later.

deepthi (approved):

rafael:
I will create an issue to keep track of the long term fix.

PR #7519, opened 2021-02-19
perf: RPC Serialization

vmg:
## Description

Hellooooo computer fiends. Coming up next in my series of performance improvements, we're tracking **RPC performance**, starting with the lowest hanging fruit: **serialization**. Right now all the RPC communication for Vitess happens through GRPC (between vtgate and the tablets, for the admin API, etc), and GRPC kind of forces us towards using Protocol Buffers for (de)serialization. This is actually not a hard requirement on the GRPC API, as it supports custom serializers, but this is not a particularly well tested interface, so in the spirit of incremental, non-breaking changes, let us not go down that road (at least for the time being).

So, since we're stuck with Protobuf, what options do we have to improve its performance? The most obvious answer is [`gogo/protobuf`](https://github.com/gogo/protobuf). For those who've never heard of the Gogo generator, it's an alternative implementation of the `protoc-gen-go` plugin that comes with Google's `go/protobuf` package. The new generator supports many new features, the most relevant (for us) being the full unfolding of the generated serialization and deserialization code for all our Protobuf messages. The default implementation of `protoc-gen-go` generates code that uses **reflection** at runtime to (de)serialize messages, and that is a very significant performance overhead.

By simply switching the generator for `protobuf` files from Google's to the `gogo` fork, we can scrape a significant amount of performance when (de)serializing Protobuf messages from/to the wire in GRPC **without changing at all the actual shape of the messages on the wire** (i.e. this is a fully transparent and backwards compatible change).

Now, relevant questions:

### If this is a magical performance improvement, why isn't everybody using it?

[...]

PR #8173, opened 2021-05-24
proto: Generate faster code using vtprotobuf

vmg:
## Description

Following up on last weeks upgrade to the ProtoBuf APIv2 and the newest version of GRPC, observing minds will have probably noticed a significant regression on CPU usage and performance during RPC since the upgrade.

This performance regression has multiple sources, but the most notable of them is the removal of the `gogo/protobuf` generated code that we were using with the ProtoBuf APIv1. In https://github.com/vitessio/vitess/pull/7519, I replaced the serialization and de-serialization code in Vitess with code autogenerated by the Gogo compiler; this was a significant performance improvement with very little cost, although the change itself was contentious: the Gogo compiler has been abandoned and it's looking for new ownership, so adding it as a dependency to Vitess was a questionable choice, particularly since as soon as we upgraded to ProtoBuf API v2, the generated Gogo compiler was no longer usable.

Lo and behold, a couple months later, and we've upgraded to ProtoBuf APIv2, with the consequent performance regression because the autogenerated code from `gogo/protobuf` is no longer usable. So what do we do about this?

### Introducing `protoc-gen-go-vtproto`

We're replacing the now-deprecated `gogo/protobuf` compiler with a proper implementation that has been written from scratch for ProtoBuf APIv2. The new compiler, available at https://github.com/planetscale/vtprotobuf, has been designed as a helper generator that is run alongside the upstream `protoc-gen-go`, instead of a replacement, and it focuses on generating high-performance (un)marshalling code -- which is the core feature that Vitess used from the old compiler.

There are still more improvements on the pipeline for the new compiler, including, huh, a README, but for now the generated code fully passes both the required and optional conformance suites for ProtoBuf and brings back most of the performance we've missed since the upgrade.

Let's see how it behaves in practice!

## Related Issue(s)
- https://github.com/vitessio/vitess/pull/7519
- https://github.com/vitessio/vitess/pull/8075

## Checklist


vmg:
I've updated the repo for the new protobuf compiler with a README and examples: https://github.com/planetscale/vtprotobuf

go/vt/servenv/grpc_codec.go
```diff
@@ -0,0 +1,56 @@
+package servenv
```
systay, line 1: license missing

Issue #8179, opened 2021-05-25
Tracking: import gh-ost test suite to Vitess/VReplication


shlomi-noach:
As we continue to invest in VReplication based migrations, we want to have parity with `gh-ost` functionality. To achieve parity, we will import the `gh-ost` testing quite: https://github.com/openark/gh-ost/tree/master/localtests

This test suite examines various types of `alter table` statements and in various table schemas. E.g. migrating an `enum` to a `varchar`, or changing a character set type, or adding a unique key on a generated column, ... Some tests are expecitng a failure (validating that an unsupported feature is reported as such).

Some of the tests will outright fail, for sure. I know at least about one capability that is not supported in `VReplication`, which is the lack of a `PRIMARY KEY`. So once we've imported the suite and have adapted the code to support it, we will for sure run into failures. Then, we will work one by one to fix those.

We will need code changes to support the suite. The `gh-ost` tests work by modifying the table on a replica, then stopping replication so as to freeze the tables, then compare contents of the _before_ and _after_ tables. With VReplication we don't run on replicas; we can cheat by tweaking the `RENAME TABLE` statement. PR to follow.

The main logic in the `gh-ost` suite is in `bash`: https://github.com/openark/gh-ost/blob/master/localtests/test.sh

I'll begin with directly importing that `bash` script; as needed, I'll translate pieces of it to `golang`.

shlomi-noach:
#8181 is in good shape and we have the testing framework up and running. #8181 includes just two tests, as proof of concept: one that is expected to pass, another that is expected to fail. The PR contains dozens of other tests, imported from `gh-ost`, but kept aside, disabled.

We can then enable those tests one by one, and fix VReplication/OnlineDDL for any failing test.

shlomi-noach:
This import is complete.

PR #8181, opened 2021-05-25
Towards a VReplication/OnlineDDL testing suite

shlomi-noach:
Setting up a VReplication testing suite.

Initial commit: support for ddl_strategy '-vreplication-test-suite', which renames both tables on cut-over to names inaccessible to the app

e.g. if table name is `mytable`, then we end up with `mytable_before` and `mytable_after`. The app (and the scripts executed by the test suite) knows nothing about those tables and therefore cannot make changes to those tabes. This means both tables are frozen in time at the moment of cut-over, and this means they can be compared without interference. And that's how the suite validates that the migration is correct: by comparing table data.

## Related Issue(s)
- https://github.com/vitessio/vitess/issues/8179

## Checklist


shlomi-noach:
# Ready for review

Suite is now working: https://github.com/vitessio/vitess/pull/8181/checks?check_run_id=2672882181

All tests files are under `testdata`: https://github.com/planetscale/vitess/tree/online-ddl-vreplication-test-suite/go/test/endtoend/onlineddl/vrepl_suite/testdata. Right now I only added two tests: one trivial that has to pass, and one that is expected to fail.

The rest of the tests are temporarily in `untestdata`: https://github.com/planetscale/vitess/tree/online-ddl-vreplication-test-suite/go/test/endtoend/onlineddl/vrepl_suite/untestdata. There's dozens of them, and I expect many to fail, if tested.

The objective is:

- Approve and merge this PR with proof of concept of two working tests
- Later, iterate the tests in `untestdata` and slowly move them to `testdata`. Anything that passes, great! Anything that fails: fix VReplication/OnlineDDL to make it pass...

go/test/endtoend/onlineddl/vtgate_util.go
```diff
@@ -106,6 +106,12 @@ func CheckCancelAllMigrations(t *testing.T, vtParams *mysql.ConnParams, expectCo
 	assert.Equal(t, expectCount, int(r.RowsAffected))
 }
 
+// ReadMigrations reads migration entries
+func ReadMigrations(t *testing.T, vtParams *mysql.ConnParams, like string) *sqltypes.Result {
+	showQuery := fmt.Sprintf("show vitess_migrations like '%s'", like)
```
vmg, line 111: Can I talk you into using `sqlparser.TrackedBuffer` for this? I don't think anybody is gonna do SQL injection in our integration test suite buuuut it's a good consistency habit to pick up :p

shlomi-noach: Sure thing

shlomi-noach: I'm looking at Vitess code and I don't see `sqlparser.TrackedBuffer` widely used for encapsulating values; I see construction using `sqlparser.NewStrLiteral`, like in:

https://github.com/vitessio/vitess/blob/bb15f60111dd984fcee83a392ac5b97d260ba125/go/vt/vtgate/planbuilder/show.go#L217-L226

Any thoughts on how to use `TrackedBuffer` to avoid SQL injection?

shlomi-noach: Rewritten to use `sqlparser.ParseAndBind()`

vmg: That's the one! Sorry I didn't make it explicit.

shlomi-noach: Good good, thank you!

shlomi-noach:
Request for review :pray: I have more to come on this topic but don't want to overwhelm with a giant PR

go/test/endtoend/onlineddl/vrepl_suite/onlineddl_vrepl_suite_test.go
```diff
@@ -0,0 +1,376 @@
+/*
+Copyright 2019 The Vitess Authors.
```
rohit-nayak-ps, line 2: 2021?

go/test/endtoend/onlineddl/vrepl_suite/onlineddl_vrepl_suite_test.go
```diff
@@ -0,0 +1,376 @@
...
+	defaultSQLMode = "ONLY_FULL_GROUP_BY,STRICT_TRANS_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO,NO_AUTO_CREATE_USER,NO_ENGINE_SUBSTITUTION"
+)
+
+func TestMain(m *testing.M) {
+	defer cluster.PanicHandler(nil)
+	flag.Parse()
+
+	exitcode, err := func() (int, error) {
+		clusterInstance = cluster.NewCluster(cell, hostname)
+		schemaChangeDirectory = path.Join("/tmp", fmt.Sprintf("schema_change_dir_%d", clusterInstance.GetAndReserveTabletUID()))
```
rohit-nayak-ps, line 73: `ClusterInstance` has a `TmpDirectory` parameter (based off of VTDATAROOT). Maybe using that here might be more consistent with the other tests. I have found that helps while debugging: if you comment out the teardown temporarily all transient files are then in the same root dir.

rohit-nayak-ps (approved):
some nits, otherwise lgtm

go/test/endtoend/onlineddl/vrepl_suite/onlineddl_vrepl_suite_test.go
shlomi-noach: Actually, I inherited this `schemaChangeDirectory` variable, and am unsure how Vitess even uses it. Having said that, there's a problem with using suggested `TmpDirectory`:

- `TmpDirectory` is only set in `StartTopo()`
- `StartTopo()` requires that `VtctldExtraArgs` are set
- `VtctldExtraArgs` rely on `schemaChangeDirectory`

so if we introduce `schemaChangeDirectory  = cluster.TmpDirectory` there's a dependency loop.

I have no strong opinions (and really no opinions at all) on how this should be resolved. Even after reviewing the code I don't understand how `SchemaChangeDirName` is being used: it's used by `LocalController` -- I don't really understand what LocalController is?

go/test/endtoend/onlineddl/vrepl_suite/onlineddl_vrepl_suite_test.go
shlomi-noach: fixed

go/test/endtoend/onlineddl/vrepl_suite/onlineddl_vrepl_suite_test.go
shlomi-noach: I'm gonna merge this PR with the nit still outstanding, and resolve `TmpDirectory` in a future PR :pray:

PR #8184, opened 2021-05-25
Update do_release to work in the 9.0 branch

systay:
## Description
In the 9.0 branch, we are not using the `version.go` file, so it should not be created by the release script


askdba (approved):
LGTM

PR #8189, opened 2021-05-25
Signed int parse

### LINKED ISSUE #8153: Error parsing DDL: `int signed`
The following valid query:
```sql
alter table t add column iii int signed not null;
```
works, but generates this in the error log:
```
ignoring error parsing DDL 'alter table t add column iii int signed not null': syntax error at position 40 near 'signed'
```
Again, it works, but I'm concerned about that warning; there shouldn't be one.

GuptaManan100:
## Description
The following query does not work correctly in the parser
```sql
alter table t add column iii int signed not null;
```
This PR fixes it by adding the optional SIGNED token after the numeric type names.

## Related Issue(s)
Fixes #8153 

## Checklist


shlomi-noach:
Does this need support in `ast_print.go`?

systay (changes requested):
We should not drop the signed/unsigned part in the AST

GuptaManan100:
We are already storing the signed/unsigned part in the AST. Its just that signed is the default so it is dropped. Since we already had the functionality of storing this in AST, we do not require any changes to `ast_format.go`. I have added a test which checks that we indeed do not ignore `unsigned`.

shlomi-noach (approved):
sounds good

systay (approved):

PR #8190, opened 2021-05-25
[workflow] Migrate `getCellsWith{Shard,Table}ReadsSwitched`, `TrafficSwitchDirection` and `TableRemovalType` to package workflow

ajm188:
## Description

Finally picking up this workstream again. This does pretty much what it says in the title, with a couple important notes to add here:
* In the original implementation of `getCellsWithShardReadsSwitched`, there was never a case where `noControls` would be set to false (because it's initialized to `true`, and there's only one branch where we say `if len(parition.GetShardTabletControls()) == 0) { noControls = true }`. I believe this to be a (apparently harmless) bug, so I've changed the implementation to initialize `noControls` to false. **No tests broke when I made this change.**
* I updated the `getCellsWith{Shard,Table}ReadsSwitched` functions to take a `topodatapb.TabletType` argument rather than a string representing the tablet type. This made a call to `strings.EqualFold` from `get....ShardReadsSwitched` unnecessary, but made a call to `strings.ToLower(ttype.String())` required. Just pushing the case-insensitive comparisons around, I guess.
* When moving the enum types (`TrafficSwitchDirection` / `TableRemovalType`), I needed to change a bunch of local variables from `workflow` to `workflowName`, so that the identifier `workflow` could refer to the package. This made the diff much larger as a result. I also took the opportunity to reformat our imports to be a little cleaner since I was already here.

## Related Issue(s)

- #7931 

## Checklist
- [x] Tests were added or are not required -- N/a

### DIFF
--- a/go/vt/vtctl/workflow/server.go
+++ b/go/vt/vtctl/workflow/server.go
@@ -118,7 +118,7 @@ func (s *Server) GetCellsWithShardReadsSwitched(
 	ctx context.Context,
 	keyspace string,
 	si *topo.ShardInfo,
-	ttype topodatapb.TabletType,
+	tabletType topodatapb.TabletType,
 ) (cellsSwitched []string, cellsNotSwitched []string, err error) {
 	cells, err := s.ts.GetCellInfoNames(ctx)
 	if err != nil {
@@ -139,11 +139,11 @@ func (s *Server) GetCellsWithShardReadsSwitched(
 		)
 
 		for _, partition := range srvks.GetPartitions() {
-			if ttype != partition.GetServedType() {
+			if tabletType != partition.GetServedType() {
 				continue
 			}
 
-			// If reads and writes are both switched it is passible that the
+			// If reads and writes are both switched it is possible that the
 			// shard is not in the partition table.
 			for _, shardReference := range partition.GetShardReferences() {
 				if key.KeyRangeEqual(shardReference.GetKeyRange(), si.GetKeyRange()) {
@@ -193,7 +193,7 @@ func (s *Server) GetCellsWithTableReadsSwitched(
 	ctx context.Context,
 	keyspace string,
 	table string,
-	ttype topodatapb.TabletType,
+	tabletType topodatapb.TabletType,
 ) (cellsSwitched []string, cellsNotSwitched []string, err error) {
 	cells, err := s.ts.GetCellInfoNames(ctx)
 	if err != nil {
@@ -221,7 +221,7 @@ func (s *Server) GetCellsWithTableReadsSwitched(
 		)
 
 		for _, rule := range srvVSchema.RoutingRules.Rules {
-			ruleName := fmt.Sprintf("%s.%s@%s", keyspace, table, strings.ToLower(ttype.String()))
+			ruleName := fmt.Sprintf("%s.%s@%s", keyspace, table, strings.ToLower(tabletType.String()))
 			if rule.FromTable == ruleName {
 				found = true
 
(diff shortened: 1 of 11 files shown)


go/vt/vtctl/workflow/server.go
```diff
@@ -73,6 +75,154 @@ func NewServer(ts *topo.Server, tmc tmclient.TabletManagerClient) *Server {
...
+// for the given tablet type.
+//
+// This function is for use in Reshard, and "switched reads" is defined as if
+// any one of the source shards has the query service disabled in its tablet
+// control record.
+func (s *Server) GetCellsWithShardReadsSwitched(
+	ctx context.Context,
+	keyspace string,
+	si *topo.ShardInfo,
+	ttype topodatapb.TabletType,
```
rafael, line 90: nit - I think across the codebase we use `tabletType` for this variable.

go/vt/wrangler/traffic_switcher.go
```diff
@@ -156,107 +127,27 @@ func (ts *trafficSwitcher) ForAllTargets(f func(source *workflow.MigrationTarget
 func (wr *Wrangler) getCellsWithShardReadsSwitched(ctx context.Context, targetKeyspace string, si *topo.ShardInfo, tabletType string) (
 	cellsSwitched, cellsNotSwitched []string, err error) {
 
-	cells, err := wr.ts.GetCellInfoNames(ctx)
+	ttype, err := topoproto.ParseTabletType(tabletType)
```
rafael, line 130: If you follow my suggestion from above, here we can:
* Rename to tabletType to be tabletTypeStr
* Just this variable stays `ttype`.

go/vt/vtctl/workflow/server.go
```diff
@@ -73,6 +75,154 @@ func NewServer(ts *topo.Server, tmc tmclient.TabletManagerClient) *Server {
...
+			found            bool
+			noControls       bool
+		)
+
+		for _, partition := range srvks.GetPartitions() {
+			if ttype != partition.GetServedType() {
+				continue
+			}
+
+			// If reads and writes are both switched it is passible that the
```
rafael, line 115: r/passible/possible

go/vt/vtctl/workflow/server.go
```diff
@@ -73,6 +75,154 @@ func NewServer(ts *topo.Server, tmc tmclient.TabletManagerClient) *Server {
...
+					found = true
+					break
+				}
+			}
+
+			// It is possible that there are no tablet controls if the target
+			// shards are not yet serving, or once reads and writes are both
+			// switched.
+			if len(partition.GetShardTabletControls()) == 0 {
+				noControls = true
```
rafael, line 128: Good catch on the `noControls` thing. Let's check with @sougou. I think there are other problems with the checks here. My understanding is a follows:
* A serving type is migrated iff:
a) The tablet_type / shard if found in the serving graph.
b) There are no tablet controls set for this tablet_type / shard.

First problem I see is that the noControls check seems tautological: Having found=true and noControls, means the shard is migrate.

The second problem I see is the following condition:

```
if !tabletControl.GetQueryServiceDisabled() {
						shardServedTypes = append(shardServedTypes, si.ShardName())
}
```
Isn't this adding to the list shards that actually have QueryService Enabled?


Furthermore, the presence of TabletControls should indicate that a migration is in progress, but ultimately the routing is determined by what is the serving graph. The more I think about this, it seems that we should just remove all these checks.

rafael (commented):
The change looks good. Good catch around the `noControls`. Let's take this opportunity to review that. I added one comment around that.

go/vt/vtctl/workflow/server.go
rafael: @rohit-nayak-ps  do you have a different understanding here? I'm more familiarized with the legacy MigrateServedTypes and maybe the logic changed.

rafael (approved):
@ajm188 this looks good to me. There is an open question, but is actually outside the scope of this PR. We can revisit that offline with @rohit-nayak-ps.

I think we can merge this as is.

PR #8193, opened 2021-05-26
[workflow] Cleanup wrangler wrappers, migrate `checkIfJournalExistsOnTablet` to package workflow

ajm188:
## Description

What it says in the title. I think long-term that the (newly-named) `CheckReshardingJournalExistsOnTablet` should stop being exported, but we need to access it from package wrangler for now, so I left a TODO comment.

Also, the previous implementation had an unnecessary check on `if !exists` which would always be true, so I removed the check.

[Edit] Oh, also, this is almost certainly going to run into merge conflicts with #8190 so 🤷 I'll just rebase whichever one loses

## Related Issue(s)

#7931 

## Checklist

### DIFF
--- a/go/vt/vtctl/workflow/server.go
+++ b/go/vt/vtctl/workflow/server.go
@@ -24,7 +24,6 @@ import (
 	"time"
 
 	"google.golang.org/protobuf/encoding/prototext"
-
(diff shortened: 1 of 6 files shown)


rafael (approved):
LGTM

PR #8210, opened 2021-05-31
Adds an environment variable to set the MySQL max connections limit in vttestserver docker image

GuptaManan100:
## Description
This PR adds an environment variable to set the MySQL max connections limit in vttestserver docker image. This will allow the users to set the MySQL `max_connections` even in the docker image of vttestserver. 

## Checklist


harshit-gangal:
The new tests added is failing. @GuptaManan100 you need to look into it.

GuptaManan100:
@harshit-gangal cluster 12 seems to already have sufficient tests running on it, adding docker tests to it makes it fail with 11-minute timeout. I have moved the tests to run on cluster 26 instead since they had the least amount of tests currently.

GuptaManan100:
reverted the tests for now. Will add them as a separate PR is needed

harshit-gangal (approved):

deepthi (commented):
Niiice!

GuptaManan100:
Tests are added in #8253

Issue #3955, opened 2018-05-21
vtgate doesn't report missing cell properly


sjmudd:
This won't happen much but vtgate doesn't validate the cell it's part of to actually exists. If this misconfiguration happens it just fails miserably:

```
Log file created at: 2018/05/21 15:43:45
Running on machine: xxxxx
Binary: Built with gc go1.9.4 for linux/amd64
Log line format: [IWEF]mmdd hh:mm:ss.uuuuuu threadid file:line] msg
I0521 15:43:45.855968   22162 buffer.go:144] vtgate buffer not enabled.
I0521 15:43:45.856162   22162 discoverygateway.go:105] loading tablets for cells: MYCELL
I0521 15:43:45.856188   22162 gateway.go:98] Gateway waiting for serving tablets...
I0521 15:43:45.858707   22162 logutil.go:31] log: Connected to ....
I0521 15:43:45.885258   22162 logutil.go:31] log: Authenticated: id=9999, timeout=10000
I0521 15:43:45.885310   22162 zk_conn.go:296] zk conn: session for addr ... event: {EventSession StateHasSession  <nil> 10.a.b.c:pppp}
E0521 15:43:45.886888   22162 topology_watcher.go:175] cannot get tablets for cell: MYCELL: node doesn't exist
E0521 15:43:45.886950   22162 resilient_server.go:290] GetSrvKeyspaceNames(context.Background.WithDeadline(2018-05-21 15:44:15.856190465 +0200 CEST m=+30.013509912 [29.969211722s]), MYCELL) failed: node doesn't exist (no cached value, caching and returning error)
F0521 15:43:45.887232   22162 vtgate.go:169] gateway.WaitForTablets failed: node doesn't exist
...
following this is a goroutine stack trace for each go routine
...
```
```
[root@some-host tmp]# grep -c "^goroutine " vtgate.INFO
12
```

I think it would be better to explicitly check the cell exists and if it does not report this as an error and exit cleanly.  Something like `cell <MYCELL> is not configured in the topology. Please check your configuration`

So this looks like it's in `go/vt/vtgate/vtgate.go:169`

So maybe not critical but it makes it clearer what's going on if someone misconfigures vitess. A clean check that the cell is valid prior to trying to use it seems worthwhile.

sjmudd:
Later I fixed this with:

```
$ zk -server myserver:2181 touch -p /vitess/MYCELL
```

So this could even be suggested too, if that seems appropriate.

PR #7251, opened 2021-01-05
Add key flags valid check when gates creating

Johnny-Three:
1.modify proper annotations for gates
2.add CheckCellFlags function to give valid check before gates creating
3.CheckCellFlags will  help to avoid strange behaviors when there are invalid parameters of cell related set 
vtgate brings up but actually does not work correctly ..

Before  : 
when you are setting cell wrong you will get some confusing error outputs like :
image
now you will see it a bit clear:
image

you could also set cell flag correct but cells_to_watch flag with two does not existing cells but vtgate will could also brings up 
but just will serves not as expectingly. like below:
image
image

also if you Accidentally  configed -tablet_types_to_wait flag wrongly like below:
image
also vtgate will runs but when you are using command like : use @replica
and then you send a query like "select * from table1" it will send qps to all tablet with all types .. 

CheckCellFlags will help to make a in advance valid check and give hints to make people to check and be familar on the key parameters on cell related flags

## Related Issue(s)

https://github.com/vitessio/vitess/issues/3955
- 

## Checklist
- [ ] Should this PR be backported?
- [ ] Tests were added or are not required
- [ ] Documentation was added or is not required

## Impacted Areas in Vitess
Components that this PR will affect:

- [ ]  Query Serving
- [ ]  VReplication
- [ ]  Cluster Management
- [ ]  Build

PR #8218, opened 2021-05-31
vtgate: validate important flags during startup

harshit-gangal:
Taken from https://github.com/vitessio/vitess/pull/7251 to get it merged.

1.modify proper annotations for gates
2.add CheckCellFlags function to give valid check before gates creating
3.CheckCellFlags will  help to avoid strange behaviors when there are invalid parameters of the cell-related set 
vtgate brings up but actually does not work correctly.

Before  : 
when you are setting cell wrong you will get some confusing error outputs like :
image
now you will see it a bit clear:
image

You could also set cell flag correct but cells_to_watch flag with two does not existing cells but vtgate will could also bring up but just will serve not as expectingly. like below:
image
image

also if you Accidentally  configured -tablet_types_to_wait flag wrongly like below:
image
also, vtgate will runs but when you are using a command like use @replica
and then you send a query like "select * from table1" it will send qps to all tablets with all types. 

CheckCellFlags will help to make an in advance valid check and give hints to make people check and be familiar with the key parameters on cell-related flags

## Related Issue(s)

Closes https://github.com/vitessio/vitess/issues/3955

## Checklist

### DIFF
--- a/go/cmd/vtgate/vtgate.go
+++ b/go/cmd/vtgate/vtgate.go
@@ -17,12 +17,14 @@ limitations under the License.
 package main
 
 import (
+	"context"
 	"flag"
 	"math/rand"
 	"strings"
 	"time"
 
-	"context"
+	"vitess.io/vitess/go/vt/proto/vtrpc"
+	"vitess.io/vitess/go/vt/vterrors"
 
 	"vitess.io/vitess/go/exit"
 	"vitess.io/vitess/go/vt/discovery"
@@ -32,6 +34,7 @@ import (
 	"vitess.io/vitess/go/vt/topo"
 	"vitess.io/vitess/go/vt/topo/topoproto"
 	"vitess.io/vitess/go/vt/vtgate"
+	"vitess.io/vitess/go/vt/vttablet/tabletserver"
 
 	topodatapb "vitess.io/vitess/go/vt/proto/topodata"
 )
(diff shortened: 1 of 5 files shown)


go/cmd/vtgate/vtgate.go
```diff
@@ -49,6 +52,63 @@ func init() {
...
+	//cell valid check
+	if cell == "" || cell == "test_nj" {
+		return vterrors.Errorf(vtrpc.Code_INVALID_ARGUMENT, "cell flag must be set")
+	}
+	hasCell := false
+	for _, v := range cellsInTopo {
+		if v == cell {
+			hasCell = true
+		}
+	}
```
frouioui, line 86: nit: you can add a `break` when the condition `v == cell` is reached to minimize the number of operation

go/cmd/vtgate/vtgate.go
```diff
@@ -49,6 +52,63 @@ func init() {
...
+		if c == "" {
+			continue
+		}
+		//cell should be contained in cellsInTopo
+		exists := topo.InCellList(c, cellsInTopo)
+		if exists {
+			cells = append(cells, c)
+		} else {
+			return vterrors.Errorf(vtrpc.Code_INVALID_ARGUMENT, "cell: [%v] is not valid. Available cells: [%v]", c, strings.Join(cellsInTopo, ","))
+		}
```
frouioui, line 103: nit: can be simplified to the following snippet

```go
if exists := topo.InCellList(c, cellsInTopo); !exists {
   return vterrors.Errorf(vtrpc.Code_INVALID_ARGUMENT, "cell: [%v] is not valid. Available cells: [%v]", c, strings.Join(cellsInTopo, ","))
}
cells = append(cells, c)
```

frouioui (approved):
Just a couple of nits! Looking good @Johnny-Three

go/cmd/vtgate/vtgate.go
frouioui: resolved via 26a5bba6475e2065a3fd5ce0853683e53bdca92d

PR #8224, opened 2021-06-01
Schema tracking: One schema load at a time per keyspace

frouioui:
## Description

Allow one schema load at a time per keyspace.

## Related Issue(s)

Depends on #8074.

## Checklist


go/vt/vtgate/schema/uptate_controller.go
```diff
@@ -0,0 +1,105 @@
...
+					break
+				}
+			}
+			if !found {
+				item.TablesUpdated = append(item.TablesUpdated, table)
+			}
+		}
+	}
+	// emptying queue's items as all items from 0 to i (length of the queue) are merged
+	u.queue.items = u.queue.items[i:]
```
systay, line 89: nit: why not use `nil` or `[]*discovery.TabletHealth{}` instead of `[i:]`?

systay (approved):

go/vt/vtgate/schema/uptate_controller.go
frouioui: This is intended to reduce memory consumption by keeping an empty `[]*discovery.TabletHealth{}` in memory instead of creating a new one every time

frouioui:
#7994 for tracking.

PR #8228, opened 2021-06-01
Online DDL/VReplication: fail on existence of FOREIGN KEY

shlomi-noach:
## Description

Importing `fail-fk` test from the `gh-ost` suite:

This PR ensures the migrated table does not participate in a foreign key relationship.

Today, we reject any `ALTER/CREATE` statements that include any `FOREIGN KEY` clause. But what if the user created a table with foreign keys directly on MySQL?

With this PR, a `online` strategy migration validates that no FKs exist on the table, child side or parent side.

It's noteworth ythat `gh-ost` runs this same test internally; so I've only added the logic to `online`/vreplication strategy.

## Related Issue(s)

- #8181 

## Checklist


GuptaManan100 (approved):
LGTM 😃

PR #8177, opened 2021-05-24
Add options to vttestserver to pass -foreign_key_mode, -enable_online_ddl, and -enable_direct_ddl through to vtcombo

mcronce:
## Description
* Add `-foreign_key_mode`, `enable_online_ddl`, and `-enable_direct_ddl` to `vttestserver`
* Pass all three through to underlying `vtcombo` process
* Add `$FOREIGN_KEY_MODE`, `$ENABLE_ONLINE_DDL`, and `$ENABLE_DIRECT_DDL` to `vttestserver` container image

## Related Issue(s)
N/A

## Checklist
- [x] Tests were added or are not required
- [x] Documentation was added or is not required

## Deployment Notes
N/A

PR #8229, opened 2021-06-01
Speedup new vttestserver tests

mcronce:
## Description
The tests added in #8177 substantially increases the time required to run that test suite (on the order of 20-40 seconds) because of all the additional spinup/teardown cycles; this combines the new tests into a single spinup/teardown cycle.  It reduces readability a bit (IMO) but given the amount of time it saves, I think it's a worthwhile trade.

## Related Issue(s)
#8177 

## Checklist

## Deployment Notes
N/A


deepthi (approved):
LGTM

dkhenry (approved):

PR #7784, opened 2021-04-02
[tablet, queryrules] Extend query rules to check comments

setassociative:
## Description
TODO: fill out PR

## Related Issue(s)

- 

## Checklist
- [ ] Should this PR be backported?
- [ ] Tests were added or are not required
- [ ] Documentation was added or is not required

## Impacted Areas in Vitess
Components that this PR will affect:

- [ ]  Query Serving
- [ ]  VReplication
- [ ]  Cluster Management
- [ ]  Build/CI
- [ ]  VTAdmin

PR #8233, opened 2021-06-02
[tablet, queryrules] Extend query rules to check comments

falun:
:warning: Caveat is that I pulled the PR and description from https://github.com/vitessio/vitess/pull/7784 since it was accidentally merged that into upstream instead of the TS branch so it's a little non-trivial to just reopen it due to the revert (afaik). I did _not_ do any additional retesting relying on the unit tests etc to do their thing.

## Description

This PR makes three changes

1. It introduces an optional us of `fsnotify` to watch for changes vs the rules file
2. it extends rules evaluation to support comparison vs leading/trailing comments
3. It adds support to specify leading/trailing comments as part of rule creation in `rulesctl`

### Details: Watching the file

We use [fsnotify](https://pkg.go.dev/github.com/fsnotify/fsnotify) which seems to be the generally accepted approach. There are some gymnastics necessary as watching the specific file is prone to losing track of it in the event that the file is deleted+recreated. To handle this case we watch the directory containing the file and react only when the node we receive events for matches the file we are watching.

### Details: Filtering on MarginComments

This was mostly plumbing within the `Rule` itself + the application callstack as the `tabletserver/query_executor::QueryExecutor` has already baked margin comments into it during the initial phase of [the Execute call](https://github.com/vitessio/vitess/blob/master/go/vt/vttablet/tabletserver/tabletserver.go#L729).

Evaluation is _similar to_ but not the same as the Query regexp because that is stable across query shapes (otherwise it wouldn't be the same query). For margin comments we need to defer evaluation until after the plan cache which is why Query & comments take different paths.

### Test failures
The only listed failure is `endtoend/tabletmanager/tablegc` which seems unlikely to be related afaict.

## Related Issue(s)

n/a

## Checklist
- [ ] Documentation was added or is not required.

## Deployment Notes
No changes necessary for existing deployments.

## Impacted Areas in Vitess

- [x]  Query Serving
- [ ]  VReplication
- [ ]  Cluster Management
- [ ]  Build/CI
- [ ]  VTAdmin

### DIFF
--- a/go/cmd/rulesctl/cmd/add.go
+++ b/go/cmd/rulesctl/cmd/add.go
@@ -13,13 +13,15 @@ import (
 )
 
 var (
-	addOptDryrun      bool
-	addOptName        string
-	addOptDescription string
-	addOptAction      string
-	addOptPlans       []string
-	addOptTables      []string
-	addOptQueryRE     string
+	addOptDryrun            bool
+	addOptName              string
+	addOptDescription       string
+	addOptAction            string
+	addOptPlans             []string
+	addOptTables            []string
+	addOptQueryRE           string
+	addOptLeadingCommentRE  string
+	addOptTrailingCommentRE string
 	// TODO: other stuff, bind vars etc
 )
 
@@ -36,8 +38,20 @@ func runAdd(cmd *cobra.Command, args []string) {
 		rule.AddTableCond(t)
 	}
 
-	if err := rule.SetQueryCond(addOptQueryRE); err != nil {
-		log.Fatalf("Query condition invalid '%v': %v", addOptQueryRE, err)
+	if addOptQueryRE != "" {
+		if err := rule.SetQueryCond(addOptQueryRE); err != nil {
+			log.Fatalf("Query condition invalid '%v': %v", addOptQueryRE, err)
+		}
+	}
+	if addOptLeadingCommentRE != "" {
+		if err := rule.SetLeadingCommentCond(addOptLeadingCommentRE); err != nil {
+			log.Fatalf("Leading comment condition invalid '%v': %v", addOptLeadingCommentRE, err)
+		}
+	}
+	if addOptTrailingCommentRE != "" {
+		if err := rule.SetTrailingCommentCond(addOptTrailingCommentRE); err != nil {
+			log.Fatalf("Trailing comment condition invalid '%v': %v", addOptTrailingCommentRE, err)
+		}
 	}
 
 	var rules *vtrules.Rules
@@ -141,6 +155,16 @@ func Add() *cobra.Command {
 		"query", "q",
 		"",
 		"A regexp that will be applied to a query in order to determine if it matches")
+	addCmd.Flags().StringVarP(
+		&addOptLeadingCommentRE,
+		"leading-comment", "l",
+		"",
+		"A regexp that will be applied to comments prefacing a SQL statement")
+	addCmd.Flags().StringVarP(
+		&addOptTrailingCommentRE,
+		"trailing-comment", "r",
+		"",
+		"A regexp that will be applied to comments after a SQL statement")
 
 	for _, f := range []string{"name", "action"} {
 		addCmd.MarkFlagRequired(f)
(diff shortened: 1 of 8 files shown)


go/vt/vttablet/tabletserver/query_executor.go.orig
```diff
@@ -0,0 +1,939 @@
+/*
```
harshit-gangal, line 1: looks like this file got accidentally pushed.

go/vt/vttablet/tabletserver/rules/rules_test.go
```diff
@@ -496,31 +497,68 @@ func TestAction(t *testing.T) {
...
+	newQrs := qrs.Copy()
+	newQrs.Add(qr4)
+
+	action, desc = newQrs.GetAction("1234", "user1", bv, mc)
+	if action != QRFail {
+		t.Errorf("want fail")
+	}
+	if desc != "rule 4" {
+		t.Errorf("want rule 4, got %s", desc)
+	}
```
harshit-gangal, line 549: change all the assertions with `asset,NotEqual` or other function checks over IF condition followed by `t.Error`

harshit-gangal (commented):
overall looks good. added some comments.

go/cmd/rulesctl/cmd/add.go
```diff
@@ -141,6 +155,16 @@ func Add() *cobra.Command {
...
+	addCmd.Flags().StringVarP(
+		&addOptLeadingCommentRE,
+		"leading-comment", "l",
+		"",
+		"A regexp that will be applied to comments prefacing a SQL statement")
+	addCmd.Flags().StringVarP(
+		&addOptTrailingCommentRE,
+		"trailing-comment", "r",
+		"",
+		"A regexp that will be applied to comments after a SQL statement")
```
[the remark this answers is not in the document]
falun: posix short flags are single character; the cobra toolkit enforces this when [registering new flags](https://github.com/spf13/pflag/blob/v1.0.5/flag.go#L861-L865)

go/vt/vttablet/tabletserver/query_executor.go.orig
falun: :facepalm:

go/vt/vttablet/tabletserver/rules/rules_test.go
falun: sure thing -- I'll update the test cases I edit so that they'll have a consistent check style but will not expand to the full file if that seems reasonable?

harshit-gangal: that's what I also do :+1:

falun: :white_check_mark: done

go/vt/vttablet/tabletserver/query_executor.go.orig
falun: removed

go/vt/vttablet/customrule/filecustomrule/filecustomrule.go
```diff
@@ -102,6 +108,45 @@ func ActivateFileCustomRules(qsc tabletserver.Controller) {
 	if *fileRulePath != "" {
 		qsc.RegisterQueryRuleSource(FileCustomRuleSource)
 		fileCustomRule.Open(qsc, *fileRulePath)
+
+		if *fileRuleShouldWatch {
+			baseDir := path.Dir(*fileRulePath)
+			ruleFileName := path.Base(*fileRulePath)
+
```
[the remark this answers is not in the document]
falun: :white_check_mark: done

go/vt/vttablet/customrule/filecustomrule/filecustomrule.go
```diff
@@ -102,6 +108,47 @@ func ActivateFileCustomRules(qsc tabletserver.Controller) {
 	if *fileRulePath != "" {
 		qsc.RegisterQueryRuleSource(FileCustomRuleSource)
 		fileCustomRule.Open(qsc, *fileRulePath)
+
+		if !*fileRuleShouldWatch {
+			return
+		}
+
```
harshit-gangal, line 115: nit: same change could be made for `fileRulePath` as was done for `fileRuleShouldWatch`

harshit-gangal (approved):

PR #8244, opened 2021-06-02
VStream API: add heartbeat for idle streams

rohit-nayak-ps:
## Description
 Currently if all source shards have 0 QPS the VStream API will not send any events. So if a client doesn't receive events it is not clear if the connection is broken or idle. This PR adds an optional heartbeat which is sent if there are no events within a specified time period.

Earlier streams used to just invoke the passed callback. Now they send events to a shared channel. We initiate an idle timer here to send heartbeats if no events are received in `flags.HeartbeatInterval`. 

### Misc

* The test for "vstream * from" construct was removed. This construct was implemented as a POC and it no longer works due to recent vtgate refactoring
* A sample client for streaming using the vstream API is added. It is setup to work with the local example.

## Checklist


go/vt/vtgate/vstream_manager.go
```diff
@@ -214,6 +221,58 @@ func (vs *vstream) stream(ctx context.Context) error {
...
+		case evs := <-vs.eventCh:
+			if err := send(evs); err != nil {
+				vs.once.Do(func() {
+					vs.err = err
+				})
+				return
+			}
+			timer.Reset(heartbeatDuration)
+		case <-timer.C:
+			now := time.Now().UnixNano()
```
vmg, line 260: This could be made simpler!

```suggestion
		case t := <-timer.C:
			now := t.UnixNano()
```

go/vt/vtgate/vstream_manager.go
```diff
@@ -214,6 +221,58 @@ func (vs *vstream) stream(ctx context.Context) error {
...
+func (vs *vstream) sendEvents(ctx context.Context) {
+	var heartbeatDuration time.Duration
+	if vs.heartbeatInterval == 0 {
+		heartbeatDuration = time.Duration(math.MaxInt64) * time.Nanosecond // max time.Duration => 290 years
+	} else {
+		heartbeatDuration = time.Duration(vs.heartbeatInterval) * time.Second
+	}
+
+	timer := time.NewTicker(heartbeatDuration)
+	defer timer.Stop()
```
vmg, line 233: So, I'm a bit spooked about keeping a timer with `MaxInt64` for the case where we're not supposed to heartbeat. I don't know how well the runtime deals with such large periods and whether it could overflow or lead to weird behavior.

What do you think about this refactoring?

```suggestion
	var heartbeat <-chan time.Time
	var resetHeartbeat func()
	
	if vs.heartbeatInterval == 0 {
		heartbeat = make(chan time.Time)
		resetHeartbeat = func() {}
	} else {
		d := time.Duration(vs.heartbeatInterval) * time.Second
		timer := time.NewTicker(d)
		defer timer.Stop()
		
		heartbeat = timer.C
		resetHeartbeat = func() { timer.Reset(d) }
	}

```

rohit-nayak-ps: Elegant solution, thanks! Implemented. I was not so crazy about that MaxInt solution as well!

vmg (approved):
neat!

PR #8247, opened 2021-06-02
Materialize: Add additional comparison operators in Materialize and fix bug where they not applied for sharded keyspaces

rohit-nayak-ps:
## Description

This PR fixes a bug where comparison filters were being ignored in Materialize workflows if the source keyspace was sharded.

In addition, it adds support for additional operators (`>, >=, <, <=, <>`) . The filter expression can chain multiple comparison operations using `and` only.

Examples:
```
select * from t1 where id = 1
select * from t1 where id <> 1
select * from t1 where val > 'abc'
select * from t1 where id >= 1
select * from t1 where id < 2 and val <= 'xyz'
```
## Checklist

### DIFF
--- a/go/test/endtoend/vreplication/materialize_test.go
+++ b/go/test/endtoend/vreplication/materialize_test.go
@@ -1,0 +1,94 @@
+/*
+Copyright 2020 The Vitess Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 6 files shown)


rohit-nayak-ps:
https://github.com/vitessio/vitess/pull/8069 has been moved over here due to issues with rebasing against main

deepthi:
@shlomi-noach @sougou can you re-review? It seems like all feedback from #8069 has been addressed.

harshit-gangal (approved):

PR #8255, opened 2021-06-03
Online DDL: report rows_copied for migration (initial support in gh-ost)

shlomi-noach:
## Description
This PR introduces a new column `rows_copied` in `_vt.schema_migrations` table and in the output for `SHOW VITESS_MIGRATIONS ...` queries.

`rows_copied` is either an exact or close estimate of the number of rows copied from the original table into the ghost table during a migration.

- If the migration is successfully complete then `rows_copied` is the exact number of rows.
- If the migration is failed then `rows_copied` value can be slightly below the actual number of rows copied, depending on the type of failure.

The value gets populated in this PR by `gh-ost` migrations. Work on https://github.com/vitessio/vitess/pull/8015 will make the value available in `online` (vreplication) migrations, too.

There's an `endtoend` test that verifies the behavior in a sharded cluster.

## Related Issue(s)

- #6926 

## Checklist

### DIFF
--- a/go/vt/vttablet/onlineddl/executor.go
+++ b/go/vt/vttablet/onlineddl/executor.go
@@ -117,6 +117,7 @@ const (
 	progressPctFull                  float64 = 100.0
 	etaSecondsUnknown                        = -1
 	etaSecondsNow                            = 0
+	rowsCopiedUnknown                        = 0
 	databasePoolSize                         = 3
 	cutOverThreshold                         = 3 * time.Second
 	vreplicationTestSuiteWaitSeconds         = 5
@@ -452,7 +453,7 @@ func (e *Executor) executeDirectly(ctx context.Context, onlineDDL *schema.Online
 	}
 	defer conn.Close()
 
-	_ = e.onSchemaMigrationStatus(ctx, onlineDDL.UUID, schema.OnlineDDLStatusRunning, false, progressPctStarted, etaSecondsUnknown)
+	_ = e.onSchemaMigrationStatus(ctx, onlineDDL.UUID, schema.OnlineDDLStatusRunning, false, progressPctStarted, etaSecondsUnknown, rowsCopiedUnknown)
 	_, err = conn.ExecuteFetch(onlineDDL.SQL, 0, false)
 
 	if err != nil {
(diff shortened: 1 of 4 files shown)


go/vt/vttablet/onlineddl/executor.go
```diff
@@ -2535,6 +2536,23 @@ func (e *Executor) updateMigrationProgress(ctx context.Context, uuid string, pro
 	return err
 }
 
+func (e *Executor) updateRowsCopied(ctx context.Context, uuid string, rowsCopied int64) error {
+	if rowsCopied <= 0 {
+		// Number of rows can only be positive. Zero or negative must mean "no information" and
+		// we don't update the table value.
+		return nil
+	}
+	query, err := sqlparser.ParseAndBind(sqlUpdateMigrationRowsCopied,
```
vmg, line 2545: My only concern is that this `ParseAndBind` could be optimized by only parsing the statement once and re-binding on each update, but I see the same pattern is used throughput the executor, so optimizing only for `updateRowsCopied` seems wrong. I'll revisit this down the road and refactor the whole file.

go/vt/vttablet/onlineddl/executor.go
```diff
@@ -2703,8 +2725,12 @@ func (e *Executor) OnSchemaMigrationStatus(ctx context.Context, uuidParam, statu
 	if eta, err := strconv.ParseInt(etaParam, 10, 64); err == nil {
 		etaSeconds = eta
 	}
+	var rowsCopied int64 = 0
```
vmg, line 2728: Weird that the linter is not complaining about this. I thought we had a lint for not zero-initializing variables that are already zero-initialized.

vmg (approved):

go/vt/vttablet/onlineddl/executor.go
shlomi-noach: Agreed, weird. I'll clean it nonetheless.

go/vt/vttablet/onlineddl/executor.go
shlomi-noach: This statement is expected to run once or twice a minute. All over the Online DDL code, I don't try to optimize query performance. Do feel free to rewrite if you want to have uniform coding style through the code, but optimization-wise this is not an interesting IMHO.

PR #8256, opened 2021-06-03
Schema tracking: new tables in sharded keyspace

### LINKED ISSUE #7994: Schema tracking v1

frouioui:
## Description

Add new end-to-end tests to ensure schema tracking does not affect the creation of new tables in a sharded keyspace.

This pull request also adds the `endtoend/vtgate/schematracker` package to the CI builds.

## Related Issue(s)

Closes #7994 

## Checklist


frouioui:
For tracking: #7994

systay (approved):

PR #8257, opened 2021-06-03
Tracker/VStreamer: only reload schema for tables in current database and not for internal table artifacts

rohit-nayak-ps:
## Description

A user had a problem migrating from Vitess 7.0 to Vitess 10.0. Their vreplication workflows were not making any forward progress in the replication mode after the migration. The cause was a custom heartbeat that was running at 2 qps which was implementing a `pseudo-gtid` similar to http://code.openark.org/blog/mysql/pseudo-gtid-ascending. The query would look like `DROP VIEW IF EXISTS pseudo_gtid._pseudo_gtid_hint__asc:55B364E3:0000000000056EE2:6DD57B85`

As part of DDL processing the VStreamer reloaded the schema for each heartbeat. In addition the tracker would also have tried to update the schema version table. Reloading schemas is expensive and doing it twice a second broke vstreamer processing.

This PR changes the logic to only reload schemas if a ddl is for a table in the current database. This automatically results in the pseudo-gtid ddls to be noops for the reloading logic. In addition we also ignore temporary tables generated by `online ddl` because vstreamer never sends events for those.

## Checklist

### DIFF
--- a/go/vt/vttablet/tabletserver/schema/tracker.go
+++ b/go/vt/vttablet/tabletserver/schema/tracker.go
@@ -25,6 +25,8 @@ import (
 
 	"google.golang.org/protobuf/proto"
 
+	"vitess.io/vitess/go/vt/schema"
+
 	"vitess.io/vitess/go/mysql"
 
 	"vitess.io/vitess/go/vt/sqlparser"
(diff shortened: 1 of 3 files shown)


rafael (approved):
LGTM

deepthi (approved):
Nice work figuring this out!

PR #8261, opened 2021-06-04
[workflow] Add vreplication_log data to workflow protos, and `VtctldServer.GetWorkflows` method

### LINKED ISSUE #8260: [vtctld] Add `vreplication_log` data to VtctldServer.GetWorkflows
#### Feature Description

We should enhance the `GetWorkflows` vtctld rpc to fetch data from the new `_vt.vreplication_log` table added in #7831. Since `vtadmin-api` is essentially forwarding the responses from the vtctlds for these endpoints, adding this to vtctld should be sufficient for us to get that for free.

ajm188:
## Description

This enhances the `GetWorkflows` endpoint in VtcltdServer to also, on a best-effort basis, query the `_vt.vreplication_log` table for all streams involved in each workflow. On tablets that are running with older versions of vitess (which do not have the log table), then the log fetch queries should fail, and we will set `LogFetchError` on each stream in each workflow in the response, but the overall RPC will still be non-error.

Here is this change running, accessed via vtadmin, on the `303_reshard` local example:

Screen Shot 2021-06-03 at 6 33 22 PM

## Related Issue(s)

Closes #8260 

## Checklist


go/vt/vtctl/workflow/server.go
```diff
@@ -447,6 +449,147 @@ func (s *Server) GetWorkflows(ctx context.Context, req *vtctldatapb.GetWorkflows
...
+					},
+					Message: message,
+					Count:   count,
+				}
+
+				// Earlier, in the main loop where we called scanWorkflow for
+				// each _vt.vreplication row, we also sorted each ShardStreams
+				// slice by ascending id, and our _vt.vreplication_log query
+				// ordered by (stream_id ASC, id ASC), so we can walk the
+				// streams in index order in O(n) amortized over all the rows
```
doeg, line 574: 🤓

doeg (approved):
This looks really really great to me. I messed with it a bit locally and the data model is purrfect for what I need. 😈 

The vexec/query planner parts seem straightforward to me, but you might want another set of eyes on the particularly Vitess-y parts. Up to you!

Thank youuuu!!!

rohit-nayak-ps (approved):
lgtm

doeg (approved):

PR #8262, opened 2021-06-04
[vtadmin] Ensure we log any errors when closing the tracer

ajm188:

## Description

This is a very small improvement to the current tracing setup in vtadmin. Rather than calling `.Close()` on the traceCloser directly, we call the wrapper in `package trace` which logs, and therefore gives us visibility into, errors that happen during the close.

## Checklist
- [ ] Tests were added or are not required -- N/A
- [ ] Documentation was added or is not required -- N/A


doeg (approved):
image

Issue #8171, opened 2021-05-24
Stats: Rates object  leaks goroutines


rohit-nayak-ps:
#### Overview of the Issue
When a new Rates object `go/stats/rates.go' is created it launches a goroutine `track()`. However this is never cleaned up. Most callers (vtgate, tabletserver) instantiate this once during their lifetime so they never go out of scope.

The VReplication controller handling VReplicationExec (within the tabletmanager) also instantiates this (via the `NewStats()`) call. Controllers get created when a stream is updated or deleted. The previous controller for this stream then go out of scope and the rates goroutine is orphaned and never cleaned up. Normally is not a major issue today since stream update/deletion is rare. 

A recent refactor in the vreplication engine as part of https://github.com/vitessio/vitess/pull/7818 resulted in a new stats object being created for every VReplicationExec call (including select). There was one use case where users were using VReplicationExec to poll workflows every few seconds. The leaked goroutines then resulted in OOMs in the corresponding VTTablets. 

While this specific use case was fixed via https://github.com/vitessio/vitess/pull/8166, we should provide a mechanism for cleaning up the rates object properly.

#### Reproduction Steps

* Use VReplicationExec to update an existing stream multiple times 
* Inspect goroutines in the corresponding vttablet's debugging profiles (eg: `http://rohit-ubuntu:15200/debug/pprof/goroutine?debug=1` and see the count for track() goroutine grow:
```
4 @ 0x43a965 0x4067cf 0x40640b 0xcb2117 0x470821
#	0xcb2116	vitess.io/vitess/go/stats.(*Rates).track+0x56	/home/rohit/vitess-ps/go/stats/rates.go:99
```

rohit-nayak-ps:
Fixed via https://github.com/vitessio/vitess/pull/13824

PR #8263, opened 2021-06-04
[vtadmin-web] Add QPS and VReplicationQPS charts to Tablet view

doeg:

## Description

As the title says! This updates the Tablet view with QPS and VReplication QPS charts, which query the /debug/vars tablet endpoint. 

I will almost certainly + indefinitely shuffle around the layout of the Tablet view but what else is new! 

Here's what a cutover looks like. 🔪 ✨ 

Screen Shot 2021-06-04 at 9 51 31 AM

Screen Shot 2021-06-04 at 9 51 26 AM

## Related Issue(s)

N/A

## Checklist

## Deployment Notes

N/A


web/vtadmin/src/util/tabletDebugVars.ts
```diff
@@ -0,0 +1,117 @@
...
+
+            if (!Array.isArray(data[planTypes[j]])) {
+                data[planTypes[j]] = [];
+            }
+
+            data[planTypes[j]].unshift({ x, y });
+        }
+    }
+
+    return data;
```
doeg, line 116: Apart from changing a couple of variable names and adding tests, I lifted this logic from the vtctld2 UI here: https://github.com/vitessio/vitess/blob/09e74af/go/vt/vttablet/tabletmanager/vreplication/stats.go#L478-L505

I came across two potential bugs/inconsistencies that I'll investigate further (outside this PR). In both cases, I decided to match the existing behaviour since at least that's consistent, for now.

- The vtctld2 logic + comments state that the charts span a 15 minute time range, with data points at 5 second intervals. However, the x-axis on the existing charts (and therefore these new VTAdmin charts) spans ~ 35 minutes:

    Screen Shot 2021-06-04 at 9 55 45 AM

    I _think_ its the x-axis that is correct, and that the data is indeed doing a 15 minute span with a 5 second sampling interval, however I'll need to more closely investigate how rates work.

- I _think_ there is a lurking bug in the aggregation of VReplicationQPS stats. It's tedious to replicate, but the VRep QPS data seems to flip-flop between individual stream rates every few (five?) seconds, rather than aggregating them. This behaviour has a similar feel to https://github.com/vitessio/vitess/issues/8171, although I'm not sure if it's related. Will follow up. :)

ajm188: https://github.com/vitessio/vitess/projects/12#card-62535183

ajm188 (approved):
lgtm

Issue #7984, opened 2021-04-28
Flaky test: TestPurge (TableGC)


rohit-nayak-ps:
```
=== RUN   TestPurge
    tablegc_test.go:331: 
        	Error Trace:	tablegc_test.go:331
        	Error:      	Should be false
        	Test:       	TestPurge
--- FAIL: TestPurge (125.03s)
FAIL
FAIL	vitess.io/vitess/go/test/endtoend/tabletmanager/tablegc	206.970s
FAIL

shlomi-noach:
https://github.com/vitessio/vitess/pull/7947 should have fixed this. But apparently didn't. https://github.com/vitessio/vitess/runs/2475285377?check_suite_focus=true

deepthi:
Happened again on #8250

shlomi-noach:
:+1:  will look into it.

shlomi-noach:
another workflow failure url: https://github.com/vitessio/vitess/pull/8255/checks?check_run_id=2736778722

shlomi-noach:
yeah this now fails more often than not :scream:

shlomi-noach:
hopefully resolved by #8270

shlomi-noach:
Nooooo! Just happened again. What is going on?!

https://github.com/vitessio/vitess/pull/8278/checks?check_run_id=2765885600

Only *ever* happens on GitHub CI and *never* on local tests

deepthi:
Right now there are 3 flaky tests in this suite.
![tablegc](https://user-images.githubusercontent.com/388311/159538507-1a47caa4-e824-4134-9562-262b4f083b22.png)

shlomi-noach:
re-investigating

PR #8270, opened 2021-06-06
Fix TableGC flaky test by reducing check interval

shlomi-noach:
## Description
Fixes https://github.com/vitessio/vitess/issues/7984

https://github.com/vitessio/vitess/issues/7984 only happens on GitHub CI, never locally. This is the 2nd time I'm looking into it, with same conclusion: it's a timing/timeout issue. In this PR I make the parge check interval configurable, and set it to `5s` as opposed to `1min`, expecting the test to now pass.

## Related Issue(s)
https://github.com/vitessio/vitess/issues/7984

## Checklist

### DIFF
--- a/go/test/endtoend/tabletmanager/tablegc/tablegc_test.go
+++ b/go/test/endtoend/tabletmanager/tablegc/tablegc_test.go
@@ -88,6 +88,7 @@ func TestMain(m *testing.M) {
 			"-heartbeat_enable",
 			"-heartbeat_interval", "250ms",
 			"-gc_check_interval", "5s",
+			"-gc_purge_check_interval", "5s",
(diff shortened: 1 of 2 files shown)


shlomi-noach:
`TableGC` passes: https://github.com/vitessio/vitess/pull/8270/checks?check_run_id=2756624032
but what I'll do is manually restart the workflow for a few times.

shlomi-noach:
I've manually restarted a dozen or so tabletmanager_tablegc` tests and they seem well. There was one failure that seemed unrelated and affected multiple steps of the test.

rohit-nayak-ps (approved):
lgtm

shlomi-noach:
:crossed_fingers:

PR #8273, opened 2021-06-07
proto: enable pooling for vreplication

vmg:
## Description

This is the optimization that is fully described in [this blog post](https://vitess.io/blog/2021-06-03-a-new-protobuf-generator-for-go/). The blog post also includes a performance graph so this issue description is rather light. :)

I'm actively looking at more places where memory pooling makes sense, but for now VReplication is a clear winner and I think we can merge this right away.

## Checklist

### DIFF
--- a/go/vt/vttablet/grpctabletconn/conn.go
+++ b/go/vt/vttablet/grpctabletconn/conn.go
@@ -675,18 +675,18 @@ func (conn *gRPCQueryClient) VStreamRows(ctx context.Context, target *querypb.Ta
 		return err
 	}
 	for {
-		r, err := stream.Recv()
+		r := binlogdatapb.VStreamRowsResponseFromVTPool()
+		err := stream.RecvMsg(r)
 		if err != nil {
 			return tabletconn.ErrorFromGRPC(err)
 		}
-		select {
-		case <-ctx.Done():
+		if ctx.Err() != nil {
 			return ctx.Err()
-		default:
 		}
 		if err := send(r); err != nil {
 			return err
 		}
+		r.ReturnToVTPool()
 	}
 }
 
--- a/go/vt/vttablet/tabletmanager/vreplication/vcopier.go
+++ b/go/vt/vttablet/tabletmanager/vreplication/vcopier.go
@@ -244,13 +244,13 @@ func (vc *vcopier) copyTable(ctx context.Context, tableName string, copyState ma
 			}
 			fieldEvent := &binlogdatapb.FieldEvent{
 				TableName: initialPlan.SendRule.Match,
-				Fields:    rows.Fields,
 			}
+			fieldEvent.Fields = append(fieldEvent.Fields, rows.Fields...)
 			vc.tablePlan, err = plan.buildExecutionPlan(fieldEvent)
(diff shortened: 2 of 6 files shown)


go/vt/vttablet/grpctabletconn/conn.go
```diff
@@ -675,18 +675,18 @@ func (conn *gRPCQueryClient) VStreamRows(ctx context.Context, target *querypb.Ta
...
-		select {
-		case <-ctx.Done():
+		if ctx.Err() != nil {
 			return ctx.Err()
-		default:
 		}
 		if err := send(r); err != nil {
 			return err
 		}
+		r.ReturnToVTPool()
```
systay, line 689: should the `r.ReturnToVTPool()` not be done in a defer so we don't miss the two other exit points from this method?

go/vt/vttablet/tabletmanager/vreplication/vcopier.go
```diff
@@ -244,13 +244,13 @@ func (vc *vcopier) copyTable(ctx context.Context, tableName string, copyState ma
 			}
 			fieldEvent := &binlogdatapb.FieldEvent{
 				TableName: initialPlan.SendRule.Match,
-				Fields:    rows.Fields,
 			}
+			fieldEvent.Fields = append(fieldEvent.Fields, rows.Fields...)
```
systay, line 248: I'm not following. isn't `fieldEvent.Fields` guaranteed to be nil since we just created the struct? Why append to the field instead of just setting it?

vmg: This is an idiomatic (maybe not that idiomatic, since you haven't heard about it! :sweat_smile:) way of copying `rows.Fields`. Since `fieldEvent.Fields` is always `nil`, but `nil` is a valid slice, this is doing the same thing as:

```go
fieldEvent.Fields = make([]pb.Field, len(rows.Fields)
copy(fieldEvent.Fields, rows.Fields)
```

...Which is something we need to do since now we're using memory pooling. We cannot set it directly to `rows.Fields` or that slice will change when we return the rows to the pool.

go/vt/vttablet/grpctabletconn/conn.go
vmg: You cannot `defer` inside of a `for` loop: each `defer` call pushes the cleanup callback to the stack, so you'd overflow very quickly. It's perfectly fine _not_ to return to the pool the message if we don't have a clean exit: not returning objects to the pool is always safe, they get GC'ed all the same. So we can afford to only return the message to the pool in the happy path. :+1:

go/vt/vttablet/tabletmanager/vreplication/vcopier.go
systay: That's what I guessed it was for, but I was uncertain.

If it said `fieldEvent.Fields = append(nil, rows.Fields...)` instead I would probably have not reacted, but here it looked weird.

systay (approved):

yangxuanjia:
when this optimizer attach to DML? It's a wonderful thing.

PR #8274, opened 2021-06-07
ci: upgrade to Go 1.16

vmg:
## Description

Heyyyoo. 1.16 is stable enough to warrant an upgrade. No breakages are expected here.

## Checklist


dkhenry:
This will require you to rebuild the bootstrap images as well. I don't think this specific upgrade will have any dependency conflicts, but we added in a `bootstrap_version` flag so we could safely upgrade the bootstrap image without impacting previous builds. You should update the flag in the main Makefile as part of this PR so that all images prior to this PR continue to use 1.15 and only new builds use 1.16 

https://github.com/vitessio/vitess/blob/696e1f51d0d2ebdd80b442e1d904287a696a923f/Makefile#L247

dkhenry (changes requested):
bump bootstrap_version as well to ensure old versions continue to use 1.15

vmg:
PR is currently blocked until I get access to Dockerhub to push the new images.

vmg:
@dkhenry I've bumped the version number and uploaded new Docker images. Can you approve?

vmg:
This PR is now green. Sorry for the delay, we had serious issues with DockerHub that have now been worked around. The Go 1.16 upgrade is ready to merge.

harshit-gangal (approved):

PR #8275, opened 2021-06-07
Online DDL/Vreplication suite: support ENUM->VARCHAR/TEXT type change

shlomi-noach:
## Description

~~This is a difficult one to solve and this PR is not ready yet.~~ solved.

See https://github.com/github/gh-ost/issues/642 for issue description in `gh-ost` and see https://github.com/openark/gh-ost/pull/18 for `gh-ost` solution.

When the user converts a column's type from from `ENUM` to `VARCHAR` they expect the new column to contain the textual representation of the `enum` value. And that's what happens for rows copied via `VCopier`. But `VPlayer` copies the _numeric_ value.

Problem is `VPlayer` (or table plan builder) do not _have_ the information about what the `enum` looks like. WIP.

## Related Issue(s)

#8181 

## Checklist


shlomi-noach:
This is now ready to review. Noteworthy:

- converting ENUM to TEXT requires changes to VReplication logic. This is because VReplication needs to know how to interpret an enum valu in the binary logs.
- OnlineDDL/vrepl is responsible for identifying an Online DDL which converts an `enum` to `varchar`/`text`
- The, it parses the enum definition
- And stores it in VReplication's Rule: a mapping between column name to its enum values. The existence of the mapping will indicate to VReplication that it must use the mapping.
- To that effect, I've added `enum_text` in `rule`'s proto. This has to happen, because VReplicatio nitself needs to be able to tell what to do with an enum value.
- We go the way down to `TablePlan` (replicator_plan.go). In `applyChange`, TablePlan looks at the field it read from the binary log. If it is an enum and is non-NULL and there is a mapping for this column, then instead of binding the `enum`'s numeric value, it binds the textual value.
- This PR does not change behavior for `enum`s otherwise. e.g. if there's a table with `enum` and there's Online DDL or MoveTables or whatever, without modifying the `enum`, then behavior is unaffected by this PR.

shlomi-noach:
Hmmm, I just added a new complex test: `enum-to-varchar-rename` which both converts a column from `enum` to test _and_ renames it. This fails. Looking into.

shlomi-noach:
Converting back to draft while checking the new test

shlomi-noach:
Ready for review, final.

Fixed the issue failing the added test.
Also, refactored code to match changes in another PR, #8322. Both PRs touch replicator plan in similar way.
Whiever PR gets merged first, will cause merge conflicts on the other, but now they're much more aligned.

rohit-nayak-ps (approved):
lgtm.

Issue #6769, opened 2020-09-22
Task: vtorc endtoend tests


deepthi:
Now that we have a vitess-native implementation of [orchestrator](https://github.com/openark/orchestrator) aka vtorc, we need to add endtoend tests for the following use cases.
- [x] Cluster has no master, master should be elected automatically
- [x] Cluster has no master, RDONLY should not be elected as master
- [x] Bring master down, a replica should be elected, replication from other instances should be re-configured to point to new master
- [ ] Bring up old master, should rejoin cluster properly
- [x] Graceful takeover
- [x] TER to a non-master, wait for repair
- [x] Make Master read-only, wait for repair
- [x] stop replication on a replica, wait for repair
- [x] point replica elsewhere, wait for repair
- [x] make replica read-write, wait for repair
- [x] Create a co-master situation (instance A replicates from B and B from A), wait for repair
- [x] Basic testcases for `-clusters_to_watch` (default, single keyspace, single keyspace/shard)
- [x] In a multi-cluster environment, each orc instance acts on specified clusters (`-clusters_to_watch` flag)

Infrastructure tasks:
- [x] basic test infrastructure
- [x] enable in CI by adding to `test/config.json`
- [x] configuration changes to speed up tests
- [x] generate port number instead of using default

EDIT June 1, 2021:
Replace Decider with vtorc.

GuptaManan100:
The only remaining test left in the list is `Bring up old master, should rejoin cluster properly`. I am not sure if that is a good idea. We don't recommend people bring up their old primaries back after they have been killed. If we decide to add this test, then it would be inherently flaky wherein the old primary could end up having errant-GTIDs causing the test to fail. If we decide we want to add that test, let's open a new issue for it. Meanwhile, I am closing this issue.

PR #8276, opened 2021-06-07
Alter tests to reuse cluster components

GuptaManan100:
## Description
This PR alters the vtorc tests to reuse the cluster components to speed them up. It also adds an additional test that checks for the failure case of circular replication. i.e. Primary and replica are replicating from each other.

Before the PR
BeforeChange

After the changes
AfterChange

The test time goes from almost 7 minutes to almost 4 minutes which is a 42.8% decrease in runtime.

## Related Issue(s)
 - #6769

## Checklist


go/test/endtoend/cluster/vtorc_process.go
```diff
@@ -60,7 +61,7 @@ func (orc *VtorcProcess) Setup() (err error) {
 	orc.proc.Args = append(orc.proc.Args, orc.ExtraArgs...)
 	orc.proc.Args = append(orc.proc.Args, "-alsologtostderr", "http")
 
-	errFile, _ := os.Create(path.Join(orc.LogDir, "orc-stderr.txt"))
+	errFile, _ := os.Create(path.Join(orc.LogDir, "orc-stderr.txt"+time.Now().String()))
```
shlomi-noach, line 64: - nit: keep `.txt` as file extension. Inject the timestamp before the extension.
- nit: `time.Now().String()` has spaces. Perhaps use unix time converted to string, which is only numeric.

go/test/endtoend/cluster/vttablet_process.go
```diff
@@ -231,7 +245,27 @@ func (vttablet *VttabletProcess) WaitForTabletTypesForTimeout(expectedTypes []st
 		}
 	}
 	return fmt.Errorf("Vttablet %s, current status = %s, expected status [%s] not reached, details: %v",
-		vttablet.TabletPath, status, strings.Join(expectedTypes, ","), vttablet.GetStatusDetails())
+		vttablet.TabletPath, status, strings.Join(expectedStatuses, ","), vttablet.GetStatusDetails())
+}
+
+// WaitForTabletTypesForTimeout waits till the tablet reaches to any of the provided types
+func (vttablet *VttabletProcess) WaitForTabletTypesForTimeout(expectedTypes []string, timeout time.Duration) error {
+	timeToWait := time.Now().Add(timeout)
```
shlomi-noach, line 253: nit: `timeToWait` is a confusing name. Perhaps rename to `waitUntil` or `expierTime` or similar?

go/vt/vtctl/vtctl.go
```diff
@@ -798,6 +801,26 @@ func commandDeleteTablet(ctx context.Context, wr *wrangler.Wrangler, subFlags *f
...
+		return err
+	}
+	for _, tabletAlias := range tabletAliases {
+		if err := wr.DemoteMasterTablet(ctx, tabletAlias); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
```
shlomi-noach, line 823: I assume this new command is only used for testing? But this still exposes the command to everyone and can confuse people.

shlomi-noach (commented):
Good work! Some nits inside

go/vt/vtctl/vtctl.go
GuptaManan100: Yes, we had a discussion regarding this and have decided to remove this function. We will create a separate function which demotes a primary, which isn't exposed for end users.

go/test/endtoend/cluster/vtorc_process.go
GuptaManan100: Changed

go/test/endtoend/cluster/vttablet_process.go
GuptaManan100: Changed

go/vt/vtctl/vtctl.go
GuptaManan100: Moved this functionality to a local function in the testing environment

shlomi-noach (approved):
Nice work! :rocket:

go/vt/wrangler/tablet.go
```diff
@@ -151,7 +151,7 @@ func (wr *Wrangler) DeleteTablet(ctx context.Context, tabletAlias *topodatapb.Ta
 		return err
 	}
 
-	return nil
+	return
```
deepthi, line 154: nit: this seems unnecessary.

deepthi (approved):
Love the reduction in test run time :)

go/vt/wrangler/tablet.go
GuptaManan100: Fixed

PR #8278, opened 2021-06-07
Online DDL/VReplication suite: reject non utf8* charsets

shlomi-noach:
## Description

Right now Vitess only supports `utf8*` characters sets, per https://vitess.io/docs/reference/compatibility/mysql-compatibility/#character-set-and-collation

While schemas pass through, schema changes with non `utf8*` charsets will not work in VReplication. For this reason, we rejects changes where tables have non `utf8*` textual columns. This PR imports and modifies a online DDL/VReplication  suite test, from the `gh-ost` suite.

Will possibly try to fix that in VReplication later on.

## Related Issue(s)

#8181 

## Checklist


shlomi-noach:
This is not ready yet

shlomi-noach:
I'm actually working to make non-utf8 work. Only blocker right now is conversion (`modify column`) from non-utf to utf column

PR #8281, opened 2021-06-08
change git hooks to use main branch

harshit-gangal:
## Description

git hooks to use main as the git_main_branch

## Checklist


harshit-gangal:
looks like these files are removed. closing it.

PR #8286, opened 2021-06-08
Copy parameter data from byte buffer in COM_STMT_SEND_LONG_DATA

harshit-gangal:
## Checklist


systay:
Let's re-open against `release-8.0`

PR #8287, opened 2021-06-08
[vtadmin-web] Add client-side error handling interface + Bugsnag implementation

doeg:

## Description

- Adds a generic + extensible interface for observability into client-side errors.
- Implements the `ErrorHandler` interface for [Bugsnag](https://www.bugsnag.com), for deployments that use it.

Implementations for other monitoring services (common ones being [Sentry](https://sentry.io/welcome/), [Rollbar](https://rollbar.com/guides/javascript-exception-handling/), [Datadog](https://docs.datadoghq.com/logs/log_collection/javascript/), etc.) are left as future enhancements. 😸 (I briefly considered implementing one other, to prove the concept, but... it didn't really make sense to add an integration that no one uses!) I looked into the JavaScript libraries for the aforementioned monitoring services and all of them fit nicely into the `ErrorHandler` interface proposed here. 

Also unimplemented for now + worth mentioning are [sourcemaps](https://docs.bugsnag.com/platforms/javascript/source-maps/), which provide proper stack-traces for errors (since our client-side code is minified). "Undocumented" is perhaps a better term than "unimplemented", as source map uploads will (unfortunately) be the responsibility of the operator. I do promise to document how we do it, once we do it. 🙏 

Here's an example of what this (specifically, the metadata) looks like in Bugsnag:

Screen Shot 2021-06-09 at 8 05 53 AM

And here's what it looks like with the Bugsnag/Slack integration:

Screen Shot 2021-06-09 at 8 05 33 AM

Many thanks to David + Anders from PlanetScale for pointing me at https://vitess.io/docs/user-guides/configuration-basic/monitoring/#tools, which was super useful in terms of precedent + documentation! 

## Related Issue(s)

N/A

## Checklist

## Deployment Notes

For any deployments that want to integrate VTAdmin with Bugsnag, you'll need to set the `REACT_APP_BUGSNAG_API_KEY` environment variable.

### DIFF
--- a/web/vtadmin/src/api/http.ts
+++ b/web/vtadmin/src/api/http.ts
@@ -15,68 +15,66 @@
  */
 
 import { vtadmin as pb } from '../proto/vtadmin';
+import * as errorHandler from '../errors/errorHandler';
+import { HttpFetchError, HttpResponseNotOkError, MalformedHttpResponseError } from '../errors/errorTypes';
+import { HttpOkResponse } from './responseTypes';
 
-interface HttpOkResponse {
-    ok: true;
-    result: any;
-}
-
-interface HttpErrorResponse {
-    ok: false;
-}
-
-export const MALFORMED_HTTP_RESPONSE_ERROR = 'MalformedHttpResponseError';
-
-// MalformedHttpResponseError is thrown when the JSON response envelope
-// is an unexpected shape.
-class MalformedHttpResponseError extends Error {
-    responseJson: object;
-
-    constructor(message: string, responseJson: object) {
-        super(message);
-        this.name = MALFORMED_HTTP_RESPONSE_ERROR;
-        this.responseJson = responseJson;
-    }
-}
-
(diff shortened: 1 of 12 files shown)


web/vtadmin/src/api/http.ts
```diff
@@ -15,68 +15,65 @@
...
-
-    const json = await response.json();
-    if (!('ok' in json)) throw new MalformedHttpResponseError('invalid http envelope', json);
-
-    // Throw "not ok" responses so that react-query correctly interprets them as errors.
-    // See https://react-query.tanstack.com/guides/query-functions#handling-and-throwing-errors
-    if (!json.ok) throw new HttpResponseNotOkError(endpoint, json);
-
-    return json as HttpOkResponse;
+    try {
```
doeg, line 37: Unfortunately this diff looks much worse than it is, since I had to surround everything in a `try/catch`. Logic changes here are minimal and well-tested. 🙏

ajm188 (approved):
🐛 🐍

PR #8291, opened 2021-06-09
[topo] Refactor `ExpandCells` to not error on valid aliases

### LINKED ISSUE #8290: [topo] ExpandCells does not correctly expand aliases
#### Overview of the Issue

`(*topo.Server).ExpandCells` returns an error when expanding a cell list that includes a cells alias.

The problem is that in [this block](https://github.com/vitessio/vitess/blob/4449aa9ccde7d007383f3699588394a0fe58f291/go/vt/topo/cell_info.go#L194-L205), we don't nil out `err` if `err2` is nil, so even if there's a valid alias, we still return the error from trying to look up the alias as pure cell.

This is affecting the ability of anyone to use Reshard/MoveTables (or any other operation that calls `ValidateSrvKeyspace`) when using a cells alias rather than explicitly specifying a list of pure cells.

#### Reproduction Steps

1. Create a cell alias
2. run `vtctlclient ... Reshard -cells $cell_alias_name ...`

This will fail during the `ValidateSrvKeyspace` call with:

```
Reshard Error: rpc error: code = Unknown desc = SrvKeyspace for keyspace${keyspace_name} is corrupt in cell ${cell_alias_name}: node doesn't exist: vitess/global/cells/${cell_alias_name}/CellInfo
```

#### Binary version

This is happening in 9.0, but the code is the same in 10.0 as well as in HEAD

ajm188:
## Description

A couple other changes here, in addition to the title:
- Store the output cells in a set to dedupe the final slice. This covers
the case where a caller passes `"alias,cell1"` where `cell1` is in the
cell list for `alias`. Without a set, it would appear twice in the final
result.
- Add an inner function to correct `defer` semantics on the cancel
funcs. Defering within the scope of a for loop is generally not what we
want, since all of the deferred functions will pile up until the total
function is done.

## Related Issue(s)

Fixes #8290

## Checklist

### DIFF
--- a/go/vt/topo/cell_info.go
+++ b/go/vt/topo/cell_info.go
@@ -17,12 +17,12 @@ limitations under the License.
 package topo
 
 import (
+	"context"
 	"path"
 	"strings"
 
 	"google.golang.org/protobuf/proto"
-
-	"context"
+	"k8s.io/apimachinery/pkg/util/sets"
 
 	"vitess.io/vitess/go/vt/vterrors"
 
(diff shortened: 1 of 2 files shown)


go/vt/topo/cell_info.go
```diff
@@ -176,37 +176,52 @@ func (ts *Server) GetKnownCells(ctx context.Context) ([]string, error) {
...
-			// not a valid cell, check whether it is a cell alias
+			// Not a valid cell name. Check whether it is an alias.
 			shortCtx, cancel := context.WithTimeout(ctx, *RemoteOperationTimeout)
 			defer cancel()
-			alias, err2 := ts.GetCellsAlias(shortCtx, cell2, false)
-			// if we get an error, either cellAlias doesn't exist or it isn't a cell alias at all. Ignore and continue
-			if err2 == nil {
-				outputCells = append(outputCells, alias.Cells...)
-			}
-			if err != nil {
```
rafael, line 203: For my understanding, here is the bug in the current code. Per conversation with @ajm188 
> we don't nil out err if err2 is nil, so even if there's a valid alias, we still return the error from trying to look up the alias as pure cell

rafael (approved):
LGTM

PR #8292, opened 2021-06-09
Add /debug/env for vtgate

aquarapid:
Similar to the vttablet /debug/env handler, borrow the code to make one for vtgate.

Initial use-case is to enable changing new healthcheck flags without a vtgate restart.  Having the getter/setters for the various flags in `replicationlag.go` is not idiomatic, but it minimizes changes.  Open to suggestions.

Not putting for this on the vtgate homepage (yet).


deepthi (approved):
LGTM.
Can you post a screenshot of what this looks like?
Approving for now, LMK if you decide to add it to the homepage and need re-review.

aquarapid:
Screenshot:  
vtgate_debugenv

PR #8293, opened 2021-06-09
The DRPC Branch

vmg:
## Description

This branch adds support for DRPC as a replacement for GRPC in the main interface for tablet connections. This is a work in progress meant to figure out the performance implications of an alternative RPC system.

## Checklist


vmg:
I've managed to get this branch far enough that we can run benchmarks on it. Here are the preliminary results:

image
image

As you can see, the improvement for point queries is significant and measurable. The regression for streaming queries is likewise significant and measurable. Despite the major decrease in CPU and memory usage, streaming queries have regressed, particularly latency wise. This seems to have a clear culprit: DRPC makes a conscious design decision to flush the stream after every write, something that GRPC does not do. I am currently investigating the rationale behind this decision.

vmg:
Update: the regression in TPCC seems to be artificial. There just seems to be a lot of variance in the benchmarking environment. @frouioui is looking into it. Here's an example of a TPCC benchmark that gives much better results:

PR #8294, opened 2021-06-09
Prepared statement buffer concurrent bug fix

systay:
## Description
During high load, if prepared statements are used, the parameter values could accidentally get overwritten with dirty data from other concurrent connections.

This PR fixes this by copying the data before releasing the buffer.

## Checklist

### DIFF
--- a/go/mysql/conn_test.go
+++ b/go/mysql/conn_test.go
@@ -18,7 +18,9 @@ package mysql
 
 import (
 	"bytes"
+	"context"
 	crypto_rand "crypto/rand"
+	"encoding/binary"
 	"fmt"
 	"math/rand"
 	"net"
@@ -28,6 +30,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"vitess.io/vitess/go/sqltypes"
 	querypb "vitess.io/vitess/go/vt/proto/query"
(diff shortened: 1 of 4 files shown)


go/mysql/conn_test.go
```diff
@@ -500,3 +504,145 @@ func (m mockAddress) String() string {
...
+func TestPrepareAndExecute(t *testing.T) {
+	// this test starts a lot of clients that all send prepared statement parameter values
+	// and check that the handler received the correct input
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	for i := 0; i < 1000; i++ {
+		startGoRoutine(ctx, t, randSeq(i))
+	}
+
+	ctx2, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
harshit-gangal, line 527: nit: can be changed to use a single context with timeout. and that can be passed to all the goroutines.

go/mysql/conn_test.go
```diff
@@ -500,3 +504,145 @@ func (m mockAddress) String() string {
...
+		require.EqualValues(t, 0, resp[0])
+
+		for count := 0; ; count++ {
+			select {
+			case <-ctx.Done():
+				return
+			default:
+				if t.Failed() {
+					return
+				}
```
harshit-gangal, line 577: this check is not interesting. as soon as this test fails. The context is cancelled and all the goroutines will start returning.

harshit-gangal (approved):

Issue #6803, opened 2020-09-30
SHOW FULL FIELDS doesn't follow routing rules


mdkent:
#### Feature Description

Allow SHOW FULL FIELDS to follow routing rules like SELECTs. This is more predictable than the current behaviour:

```
mysql> use mainunsharded;
Database changed
mysql> select * from entries limit 1;
+----+------------+----------+----------------+--------------+------------+--------+---------------------+---------------------+---------------------+------------------+-------------------------+
| id | account_id | topic_id | entryable_type | entryable_id | creator_id | status | created_at          | updated_at          | active_at           | replied_entry_id | alternative_sender_name |
+----+------------+----------+----------------+--------------+------------+--------+---------------------+---------------------+---------------------+------------------+-------------------------+
|  1 |          1 |        1 | Message        |            1 |          1 |      1 | 2019-04-08 12:49:06 | 2019-04-08 12:49:06 | 2019-04-08 12:49:06 |             NULL | NULL                    |
+----+------------+----------+----------------+--------------+------------+--------+---------------------+---------------------+---------------------+------------------+-------------------------+
1 row in set (0.10 sec)

mysql> SHOW FULL FIELDS FROM `entries`;
ERROR 1146 (42S02): vtgate: http://foo-useast1-vtgate-7ca66dd7-54c7898d68-g97m7:15000/: target: mainunsharded.-.master, used tablet: useast1-876642563 (10.119.7.187): vttablet: rpc error: code = NotFound desc = Table 'foo.entries' doesn't exist (errno 1146) (sqlstate 42S02) (CallerID: foo-vitess): Sql: "SHOW FULL FIELDS FROM `entries`", BindVars: {}
```
and will fix an issue with Rails where it expects to be able to run [SHOW FULL FIELDS](https://github.com/rails/rails/blob/5cfd58bbfb8425ab1931c618d98b649bab059ce6/activerecord/lib/active_record/connection_adapters/abstract_mysql_adapter.rb#L811-L815) on any ActiveRecord powered tables at boot time.

#### Use Case(s)

We have both unsharded and sharded tables in different keyspaces, which we can link together with routing rules in the same database, but rails expects to be able to run SHOW FULL FIELDS on all of them.

Without this we're forced to use a workaround in [schema caching](https://blog.saeloun.com/2020/03/18/rails-schema-cache-path-config.html) which avoids the SHOW FULL FIELDS entirely at the expense of some complexity.

PR #8303, opened 2021-06-10
check for read_only flag to turn off before draining buffered requests

harshit-gangal:
## Checklist


aquarapid:
Looks like it's reasonable.  Do we know why the tests are failing?

GuptaManan100:
This bug got fixed already in #6803

PR #8304, opened 2021-06-10
Turn off percona 5.6 tests

systay:
## Description
Turns off percona 5.6 tests, since the binaries for it are no longer available.

## Checklist


rohit-nayak-ps (approved):
lgtm

deepthi:
How is CI passing on the main branch? We still seem to have percona56 tests there.
https://github.com/vitessio/vitess/blob/main/.github/workflows/unit_test_percona56.yml

deepthi (changes requested):
Needs further investigation as to why these tests pass on main.

harshit-gangal:
This change is not required, breaking the CI workflows and using ubuntu-18.04 is sufficient to unlock the other PR merged.
https://github.com/vitessio/vitess/pull/8361

Issue #8309, opened 2021-06-11
Consul lock does not really work with timeout


narcsfz:
When filing a bug, please include the following headings if
possible. Any example text in this template can be deleted.

#### Overview of the Issue
Vitess assumes the process gets the lock if the error is nil https://github.com/vitessio/vitess/blob/main/go/vt/topo/locks.go

However, consul will return nil error when the context timeout’d before the process get the lock: https://github.com/hashicorp/consul/blob/master/api/lock.go#L185 - which means anyone will get the lock after the context timeout.

This means after the context timeout, everyone that is waiting for the lock will get one

#### Reproduction Steps
Process A grabbed a lock and then have another process B grab the lock, while A is holding the lock, B will also get one after the context timeout

#### Operating system and Environment details

Does not matter

PR #8310, opened 2021-06-11
Use consul lock properly

narcsfz:

## Description

Vitess assumes the process gets the lock if the error is nil https://github.com/vitessio/vitess/blob/main/go/vt/topo/locks.go

However, consul will return nil error when the context timeout’d before the process get the lock: https://github.com/hashicorp/consul/blob/master/api/lock.go#L185 - which means anyone will get the lock after the context timeout.

This means after the context timeout, everyone that is waiting for the lock will get one

This PR make sure we will raise error if the lost chan returned is nil, we should not assume we have the lock

## Related Issue(s)

https://github.com/vitessio/vitess/issues/8309

## Checklist

### DIFF
--- a/go/vt/topo/consultopo/lock.go
+++ b/go/vt/topo/consultopo/lock.go
@@ -86,14 +86,18 @@ func (s *Server) Lock(ctx context.Context, dirPath, contents string) (topo.LockD
 
 	// We are the only ones trying to lock now.
 	lost, err := l.Lock(ctx.Done())
-	if err != nil {
+	if err != nil || lost == nil {
 		// Failed to lock, give up our slot in locks map.
 		// Close the channel to unblock anyone else.
 		s.mu.Lock()
 		delete(s.locks, lockPath)
 		s.mu.Unlock()
 		close(li.done)
-
+		// Consul will return empty leaderCh with nil error if we cannot get lock before the timeout
+		// therefore we return a timeout error here
+		if lost == nil {
+			return nil, topo.NewError(topo.Timeout, lockPath)
+		}
 		return nil, err
 	}
 


narcsfz:
Gentle reminder, thanks!

deepthi (commented):
Change looks good.
Is there a way to simulate the failure condition in a unit test?

narcsfz:
@deepthi it is a little bit to simulate the failure in unit test. 

iiuc the test we have for topo lock are here: https://github.com/vitessio/vitess/blob/main/go/vt/topo/test/lock.go#L136

we are using same process that tries to grab the lock, the check here will consume all the time in the context: https://github.com/vitessio/vitess/blob/main/go/vt/topo/consultopo/lock.go#L72 before we jump into the line that causes the real issue here https://github.com/vitessio/vitess/blob/main/go/vt/topo/consultopo/lock.go#L88

any suggestions to set this up properly?

deepthi:
> @deepthi it is a little bit to simulate the failure in unit test.
> 
> iiuc the test we have for topo lock are here: https://github.com/vitessio/vitess/blob/main/go/vt/topo/test/lock.go#L136
> 
> we are using same process that tries to grab the lock, the check here will consume all the time in the context: https://github.com/vitessio/vitess/blob/main/go/vt/topo/consultopo/lock.go#L72 before we jump into the line that causes the real issue here https://github.com/vitessio/vitess/blob/main/go/vt/topo/consultopo/lock.go#L88
> 
> any suggestions to set this up properly?

Yeah I see the problem. I think it's ok to merge this as-is.

deepthi (approved):

PR #8311, opened 2021-06-11
[vtadmin-web] Small bugfix where stream source displayed instead of target

doeg:
## Description

A small bugfix to display target shard, instead of source shard, for workflow streams. 😊 

## Related Issue(s)

N/A

## Checklist

## Deployment Notes

N/A


rohit-nayak-ps (approved):
lgtm

PR #8322, opened 2021-06-13
Online DDL/VReplication: support non-UTF8 character sets

shlomi-noach:
## Description

Supporting non UTF8 character sets in OnlineDDL/VReplication:

- Table can be created in any character set
- Online DDL supports copying table data even if character set is non utf8
- Supports converting from utf8 to non-utf8
- Supports converting from non-utf8 to utf8
- Supports oncverting from non-utf8 to non-utf8 (e.g. `latin1` to `latin2`)
- Limited set of character sets supports (but the major ones are supported)

Implementation:

- VReplication's filter->rule now has a new field, `CharsetConversion`, which tells Vreplication:
  - for every textual column
  - which has seen a change in character set
  - what was the original charset and what is the new charset
  VReplication plan will programmatically convert binlog text values according to the above.

- VCopier adds a `convert(<column> using utf8mb4)` to any textual column where charset/collation has changed
- VStreamer identifies the `convert()` function and similarly runs a `convert(... using utf8mb4)` on same column as VCopier hinted, so that we convert the columns on the source side

## Related Issue(s)

- #8179 
- #8181 

## Checklist

## Notes

Submitting code changes with a lot of commented code and with debug info, as `draft pull request`. Will clean it up before making it ready for review.


shlomi-noach:
Ready for review!

Notable:

- online DDL instructs vstreamer to read all textual columns with `convert(? using utf8mb4)`
- online DDL tells vreplication/rule whether a column has a from-to charset conversion
- binary logs text values are always in utf8, so this lines up both
- replication plan examines textual column and sees if there's a charset conversion, and decodes programmatically as needed

Tests are good and present multiple interesting cases (from/to utf8, from/to latin1, latin1-to-latin2, nullable, ...)

shlomi-noach:
Tests added in this PR:

- `alter-charset-non-utf8-80`
- `alter-charset-non-utf8`
- `convert-utf8mb4`

shlomi-noach:
Resolved conflicts having merged https://github.com/vitessio/vitess/pull/8275, and cleaned up the code a bit.

rohit-nayak-ps (approved):
nice! lgtm

PR #8334, opened 2021-06-15
Online DDL/Vreplication suite: fix test for no shared UK

shlomi-noach:
## Description

In this PR we add and fix the test `fail-no-shared-uk` where the migration fails if there's no exact match between PK columns in before/after tables.

Note that we currently expect both before/after tables to have same set of columns covered by `PRIMARY KEY`. In the future, we will relax the constraint and allow for any shared non-null `UNIQUE KEY`.

## Related Issue(s)

- #8179 
- 

## Checklist


shlomi-noach:
converting to draft while investigating impact onto another test

shlomi-noach:
issue fixed. Ready for review.

shlomi-noach:
ping for review

vmg (approved):

PR #8335, opened 2021-06-15
VReplication (and by product, Online DDL): support GENERATED column as part of PRIMARY KEY

shlomi-noach:
## Description

Worked to add/fix Online DDL/VReplication suite test `generated-columns57-unique`, an dended up fixing something more generic to VReplication:

1. In MySQL it is valid to have a generated column as part of the `PRIMARY KEY`.
2. In VReplication, we skip reading `GENERATED` columns
3. In VReplication, we validate we've read all `PRIMARY KEY` columns

The combination of (2+3) contradicts (1).

With this PR, VReplication does not require (3) for `GENERATED` columns.

## Related Issue(s)

- #8179 

## Checklist


rohit-nayak-ps:
This is nice! Another edge case in vreplication handled :-) 

The online ddl test tests both the copy and replication phase, right? So we don't need to add explicit tests for this case as part of the  vcopier/vplayer testsuite, right?

shlomi-noach:
> The online ddl test tests both the copy and replication phase, right?

Correct!

rohit-nayak-ps (approved):
lgtm

PR #8336, opened 2021-06-15
Online DDL/VReplication: test PK column case change

shlomi-noach:
## Description

Add test `modify-change-case-pk` which validates everything works well when changing the case of a `PRIMARY KEY` column, eg. from `id` to `ID`.

This just works with current logic and no changes were required to make this test pass.
BTW `gh-ost` actually fails that kind of change.

## Related Issue(s)

- #8179 

## Checklist


shlomi-noach:
ping for review

vmg (approved):

PR #8337, opened 2021-06-15
Online DDL/VReplication: support DROP+ADD column of same name

shlomi-noach:
## Description

It is valid (though not good practice, probably) to `DROP` a column `c` and `ADD` a new column also called `c`. In this scenario, the column `c` in the _before_ table has nothing to do with `c` in the _after_ table.

In this PR we import logic from `gh-ost` to make sure Vreplication does not try to copy data between such column.

We add the test `drop-null-add-not-null` to OnlineDDL/vrpelication suite.

## Related Issue(s)

- #8179

## Checklist


rohit-nayak-ps (approved):
lgtm

Issue #8323, opened 2021-06-13
Flaky TestPlayerIdleUpdate test


ajm188:
```
--- FAIL: TestPlayerIdleUpdate (0.14s)
Error:     vplayer_flaky_test.go:1900: query:
        "update _vt.vreplication set pos='MySQL56/6f566dda-cc97-11eb-a30c-000d3aff91e3:1-897', time_updated=1623623653, transaction_timestamp=1623623653, rows_copied=0, message='' where id=45", does not match query 0:
        "begin"
Error:     vplayer_flaky_test.go:1900: query:
        "begin", does not match query 1:
        "insert into t1(id,val) values (1,'aaa')"
Error:     vplayer_flaky_test.go:1900: query:
        "insert into t1(id,val) values (1,'aaa')", does not match query 2:
        "/update _vt.vreplication set pos="
Error:     vplayer_flaky_test.go:1900: query:
        "update _vt.vreplication set pos='MySQL56/6f566dda-cc97-11eb-a30c-000d3aff91e3:1-902', time_updated=1623623653, transaction_timestamp=1623623653, rows_copied=0, message='' where id=45", does not match query 3:
        "commit"
Error:     vplayer_flaky_test.go:1911: query:
        "commit", does not match query 0:
        "/update _vt.vreplication set pos="
E0613 22:34:20.451543   32307 vreplicator.go:148] Replicate error: vstream ended
E0613 22:34:20.452607   32307 controller.go:163] stream 52: vstream ended, retrying after 1ms
FAIL
FAIL	vitess.io/vitess/go/vt/vttablet/tabletmanager/vreplication	28.481s
FAIL
```

Logs before the failure:

```
E0613 23:18:56.375369   32587 query.go:83] ExecuteFetch(ALTER TABLE _vt.vreplication ADD COLUMN db_name VARBINARY(255) NOT NULL) failed: Duplicate column name 'db_name' (errno 1060) (sqlstate 42S21) during query: ALTER TABLE _vt.vreplication ADD COLUMN db_name VARBINARY(255) NOT NULL
E0613 23:18:56.424948   32587 binlog_player.go:242] error (expected error) in selecting vreplication settings select pos, stop_pos, max_tps, max_replication_lag, state from _vt.vreplication where id=1
[...]

PR #8339, opened 2021-06-15
Heuristic fix for a flaky test: allow for some noise to pass through

shlomi-noach:
## Description

An attempt at fixing #8323 

What happens in #8323 is that a `update _vt.vreplication set pos=` query is encountered unexpectedly. This query can appear due to heartbeat leftovers from a previous test or from a heartbeat race condition.

This PR allows for just a little bit of noise to pass through: `expectDBClientQueries` supports `skippableOnce` optional list of queries. It's OK to hit _at most_ one of given skippable queries, _once_, while iterating expected queries.

## Related Issue(s)

- https://github.com/vitessio/vitess/issues/8323
- 

## Checklist

### DIFF
--- a/go/vt/vttablet/tabletmanager/vreplication/framework_test.go
+++ b/go/vt/vttablet/tabletmanager/vreplication/framework_test.go
@@ -481,12 +481,24 @@ func shouldIgnoreQuery(query string) bool {
 	return heartbeatRe.MatchString(query)
 }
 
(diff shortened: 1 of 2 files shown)


shlomi-noach:
3 successive successful completions for `unit_race`. Triggering more runs.

shlomi-noach:
OK this continuously passes.

rohit-nayak-ps (approved):
lgtm

PR #8345, opened 2021-06-16
Online DDL/VReplication test suite: support ENUM as part of PRIMARY KEY

shlomi-noach:
## Description

It's valid to have an `enum` as part of a `PRIMARY KEY`. In such scenario, we must treat the `enum` as text and not as integer, to handle a scenario where a query may actually update the enum value (bad practice to update a `PRIMARY KEY` value, but yet valid).

Added tests:

- `enum-pk` - tests that enum PK works, that updates to enum (as part of PK) values work
- `update-pk` - tests that varchar PK works, that updates to varchar (as part of PK) values work 

## Related Issue(s)

- #8179 

## Checklist


rohit-nayak-ps (approved):
lgtm

PR #8350, opened 2021-06-18
[topo] Optimize `GetTabletMap` to only need one conn lookup per cell

ajm188:
## Description

Currently, `GetTabletMap` makes one call to `GetTablet` per tablet
alias. This is good, as that's what the function is supposed to do.
However, `GetTablet` is a high-level function, and _it_ makes a call to
`ConnForCell`. If we have multiple aliases in the same cell, this means
that they will contend with each other for the lock in `ConnForCell`,
when they are going to end up using the same cached connection anyway.

This refactor lets us skip that synchronization, by moving the bulk of
`GetTablet` into a helper function that takes a `Conn`. Therefore we can
perform exactly one Conn lookup per cell during `GetTabletMap`, and
reuse that without needing to synchronize.

## Related Issue(s)

None yet, I will create one if this turns out to be useful.

## Checklist

### DIFF
--- a/go/vt/topo/tablet.go
+++ b/go/vt/topo/tablet.go
@@ -19,6 +19,7 @@ package topo
 import (
 	"fmt"
 	"path"
+	"sort"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


deepthi:
Perhaps this PR is now obsolete since we did https://github.com/vitessio/vitess/pull/9560?

ajm188:
Yeah, let's close this. If we need further topo optimizations we can revisit.

PR #8353, opened 2021-06-18
Add a sema to bound the number of inflight vexec queries when getting workflows

ajm188:

## Description

What it says in the title. Current defaults are somewhat arbitrary 1000 concurrent queries with a 50ms timeout to get a conn.

## Checklist
- [ ] Tests were added or are not required -- n/a
- [ ] Documentation was added or is not required -- n/a

### DIFF
--- a/go/vt/vtctl/workflow/server.go
+++ b/go/vt/vtctl/workflow/server.go
@@ -19,6 +19,7 @@ package workflow
 import (
 	"context"
 	"errors"
+	"flag"
 	"fmt"
 	"sort"
 	"strings"
@@ -29,6 +30,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 
 	"vitess.io/vitess/go/sqltypes"
+	"vitess.io/vitess/go/sync2"
 	"vitess.io/vitess/go/trace"
 	"vitess.io/vitess/go/vt/concurrency"
 	"vitess.io/vitess/go/vt/key"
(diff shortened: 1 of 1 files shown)


doeg (approved):

go/vt/vtctl/workflow/server.go
```diff
@@ -39,11 +41,17 @@ import (
...
 	binlogdatapb "vitess.io/vitess/go/vt/proto/binlogdata"
+	querypb "vitess.io/vitess/go/vt/proto/query"
 	topodatapb "vitess.io/vitess/go/vt/proto/topodata"
 	vtctldatapb "vitess.io/vitess/go/vt/proto/vtctldata"
 	"vitess.io/vitess/go/vt/proto/vttime"
 )
 
+var (
+	vExecPoolSize    = flag.Uint("workflow_server_vexec_pool_size", 1000, "maximum number of concurrent vexec queries to allow")
+	vExecPoolTimeout = flag.Duration("workflow_server_vexec_pool_default_timeout", time.Millisecond*50, "default timeout to wait acquiring a connection from the vexec pool. zero implies no timeout")
```
setassociative, line 52: we should update release notes with this and if we expect to roll out limits per call (imo) we should have a better approach than a flag for concurrency-per-call

setassociative (approved):
Ideally we'd have a little more information around the concurrent query utilization for normal operations workflow before setting a limit. Basically - if this is a problem we should tune to control the problem; if it's not then why introduce it? 

That said I believe this is safe. If there is a larger thing we're chasing let's try to get it into an issue.

No major objections to merging this as is though -- it seems like a good pattern for endpoints that can stand throttling. (🗒️ Implicitly: GetWorkflows is not in the hot path anywhere afaik, does that vibe with your understanding)

setassociative:
From unit tests the e2e is worth looking into; 
> --- FAIL: TestCellAliasVreplicationWorkflow (360.09s)
>   --- FAIL: TestCellAliasVreplicationWorkflow/shardCustomer (300.99s)

The 300s reads like a timeout but that's adjacent to "we exhausted our pool and dead locked somehow." I don't see how that'd happen given code but 🤷. I kicked off a rerun in case it was transient.

ajm188:
#8368 is looking like a more promising (and holistic) fix. Going to close this

Issue #8351, opened 2021-06-18
Flaky test: onlineddl_vrepl_suite_test.go


systay:
`onlineddl_vrepl_suite_test.go` keeps failing on unrelated PRs. 

Two examples:
https://github.com/vitessio/vitess/pull/8312/checks?check_run_id=2855474705
https://github.com/vitessio/vitess/pull/8325/checks?check_run_id=2855475377

Saved a log output:
https://gist.github.com/systay/35275ed8337a8c7a2e9dbc4cbd8c3053

shlomi-noach:
For immediate relief I made the `vrepl_suite` test non-`required`. Looking into.

shlomi-noach:
The failing test is `update-pk`, added in https://github.com/vitessio/vitess/pull/8345

shlomi-noach:
Hmmm, _also_, and unrelated, there's a `10m` timeout.

shlomi-noach:
Fixedin https://github.com/vitessio/vitess/pull/8355

PR #8355, opened 2021-06-20
Bugfix: assign cexpr.references for column CONVERTed to utf8mb4

shlomi-noach:
## Description
Fixes https://github.com/vitessio/vitess/issues/8351

These two PRs:

- https://github.com/vitessio/vitess/pull/8322
- https://github.com/vitessio/vitess/pull/8345

Introduced a conflict. Because both were open at the same time and both merged to `main` without merging the 1st into the 2nd, this conflict was only seen in `main` after both merged.

The nature of the bug:
- In #8322 we introduce casting to `utf8mb4` for texts. 
- In #8345 we introduce a test where a text column is part of a primary key, and is modified via `UPDATE`
Now, when the text is `CONVERT`ed, then it becomes a function and ceases to be a text, and is not found to be part of the primary key anymore.

This PR takes care of that edge case.

## Related Issue(s)

https://github.com/vitessio/vitess/issues/8351

## Checklist


systay (approved):
👍

Issue #7280, opened 2021-01-11
Gen4 Tracking


systay:
This issue is meant to track the work going on on the Gen4 planner.

The Gen4 planner is a new planner in Vitess that explores many different join alternatives and uses a little bit of cost analysis to pick the cheapest plan. In contrast, the V3 planner merges and join tables from left to right, and this made it important for the user to list tables in a good order so that the planner could produce an efficient route.

The gen4 algorithm that we will start implementing is based on the GOO paper (http://citeseerx.ist.psu.edu/viewdoc/summary?doi=10.1.1.31.737), but the infrastructure for it can be reused for other models. 

This is a larger rewrite of the vtgate planner. It introduces new passes and intermediate representations of the query.

The old code used these passes over the query:

| *Pass*                    | *Struct transformation*         |
|---------------------------|---------------------------------|
| Parsing                   | String -> AST                   |
| Rewriting (normalization) | AST -> AST                      |
| Planning                  | AST -> logicalPlan (builder)    |
| WireUp                    | logicalPlan -> engine.Primitive |

This refactored planner now uses the following passes:

| *Pass*                    | *Struct transformation*         |
|---------------------------|---------------------------------|
| Parsing                   | String -> AST                   |
| Rewriting (normalization) | AST -> AST                      |
| Semantic Analysis         | AST -> AST"                     |
| Build Operator Tree       | AST" -> OperatorTree                  |
| Optimize/Merge            | OperatorTree -> joinTree          |
| Horizon Planning          | joinTree -> logicalPlan         |
| WireUp                    | logicalPlan -> engine.Primitive |

By splitting the planning process into smaller pieces, each part can be simplified and extended to do more.

Here follows a short description of each new pass.

### Semantic Analysis

*Responsibilities: Scoping, Binding*

Walks the AST and does scoping and binding, so whenever a column name is found, the planner has information about which tables is being referenced. Tables are given a `TableSet` identifier - a bitmask struct that allows the planner to quickly find what dependencies every expression has.

### Extract Query Graph

*Responsibilities: Extract Subqueries, Create Query Graph*

The query graph is an intermediate representation that is designed to allow the route planner to quickly consider many different solutions for the query. Instead of keeping the query in the AST, which is limited by the tree structure it has, we produce a graphy representation with all used tables (nodes) in one list, and edges between them in a separate list.

[...]

PR #8359, opened 2021-06-21
gen4: Fail all queries not handled well by gen4

systay:
## Description
This PR will make sure that the gen4 planner does one of the following:
- Produces a correct query (correct here means "does what the user wants", not necessarily the same thing that V3 did)
- Produces a correct error (if the query is wrong, it should be detected and failed correctly)
- Produce a Gen4NotSupported error. This makes it easy to fall back on V3 when gen4 does not yet support the query.

## Related Issue(s)
#7280

## Checklist


systay:
Not yet ready to be accepted, but feel free to review what is already there

go/vt/vtgate/planbuilder/testdata/ddl_cases_no_default_keyspace.txt
```diff
@@ -436,6 +500,18 @@
...
+{
+  "QueryType": "DDL",
+  "Original": "create view user.view_a as select sql_calc_found_rows * from music where user_id = 1",
+  "Instructions": {
+    "OperatorType": "DDL",
+    "Keyspace": {
+      "Name": "user",
+      "Sharded": true
+    },
+    "Query": "create view view_a as select sql_calc_found_rows * from music where user_id = 1"
```
harshit-gangal, line 512: Does Gen4 handles sql_calc_found_rows? This query can be handled better.

harshit-gangal (approved):

PR #8362, opened 2021-06-22
tm state: don't populate metadata in updateLocked 

deepthi:
## Description

This is a partial revert of #8107. Calling `populateMetadata` in `updateLocked` causes `publishState` to fail when semi-sync is enabled.

## Related Issue(s)

TBD

## Checklist

### DIFF
--- a/go/test/endtoend/reparent/utils_test.go
+++ b/go/test/endtoend/reparent/utils_test.go
@@ -116,6 +116,7 @@ func setupCluster(ctx context.Context, t *testing.T, shardName string, cells []s
 	clusterInstance.VtTabletExtraArgs = []string{
 		"-lock_tables_timeout", "5s",
 		"-enable_semi_sync",
+		"-init_populate_metadata",
 		"-track_schema_versions=true",
 	}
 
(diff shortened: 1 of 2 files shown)


deepthi:
@ajm188 it's possible that the right fix is as simple as changing these conditions:
https://github.com/vitessio/vitess/pull/8107/files#diff-8b21ba72e9fc7850bc4516d90ebf0639ac39fbef7e7b0ac534e2f470574c5a77R290-R297

```
	if ts.tm.MetadataManager == nil {
		return
	}

	if ts.isOpening && !*initPopulateMetadata {
		return
	}
```
During  PRS isOpening is false and the code gets executed even when initPopulateMetadata is false. I believe the check really needs to be
```
	if ts.tm.MetadataManager == nil || !*initPopulateMetadata {
		return
	}
```

deepthi:
> @ajm188 it's possible that the right fix is as simple as changing these conditions:
> https://github.com/vitessio/vitess/pull/8107/files#diff-8b21ba72e9fc7850bc4516d90ebf0639ac39fbef7e7b0ac534e2f470574c5a77R290-R297

On second thought, populateMetadata should _NEVER_ be called in a code path that runs during a reparent.
No DB writes can happen during that process because they will hang when semi-sync is on.

ajm188:
I am thinking we should add a flag to the section of `ChangeTabletType` [here](https://github.com/vitessio/vitess/blob/1bfab0d3a4c18421a3db66fb54a2f58ca6c37910/go/vt/vttablet/tabletmanager/tm_state.go#L178-L199), and then only call populateMetadata if we're not transition from/to MASTER

ajm188 (approved):

deepthi:
I'm merging this for now to fix the regression. It will need to be fixed properly later (with appropriate tests).

PR #8363, opened 2021-06-22
sks-keyservers.net seems to be finally dead, replace with

aquarapid:
keyserver.ubuntu.com

See https://sks-keyservers.net/:

```
This service is deprecated. This means it is no longer maintained, and new HKPS certificates will not be issued. Service reliability should not be expected.

Update 2021-06-21: Due to even more GDPR takedown requests, the DNS records for the pool will no longer be provided at all.
```


aquarapid:
Full docker builds (`make docker_base_all` and `make docker_lite_all`) validated locally

deepthi (approved):
LGTM

PR #8364, opened 2021-06-22
VReplication support for non-PRIMARY KEY (Online DDL context)

shlomi-noach:
## Description

Today, VReplication assumes:

- All tables have a `PRIMARY KEY`
- source table and target table have _same_ `PRIMARY KEY` (same columns, same order of columns)

Online DDL requires that this restriction is lifted. A user should be allowed to change a `PRIMARY KEY`. Online DDL-wise, the requirement is that both source and target tables have at least one shared not-nullable `UNIQUE KEY`. A valid example:

```sql
create table t {
  uuid varchar(64) not null primary key,
  some_other_column int not null
}
```
with the following change:
```sql
alter table t drop primary key add column id bigint unsigned not null auto_increment primary key, add unique key uuid_uidx(uuid)
```

In the above scenario both before/after tables have a non-nullable unique key on `uuid` column. In the source table, it is `PRIMARY`, and in target table, it is `uuid_idx`. The fact that the name of the index is different is merely a technicality. VReplication should be able to iterate tables by `uuid` order on both source and target.

Today, the assumption that we only ever use a `PRIMARY KEY` goes very deep. The `reload()` function in vreplication pre-evaluates all tables in a schema according to their PK, irrespective of any VReplication stream. In today's assumption, when a new VReplication stream kicks in, it _just knows_ which table columns comprise the PRIMARY KEY, because everything is already evaluated beforehand.

However, we now want to choose a unique key per operation. Maybe the source/target share the exact same `PRIMARY KEY`, as is normally the case, and there's nothing to do. But maybe they don't, and there's multiple options. So we need to evaluate which unique keys are shared between the two tables, pick the "best" shared key, and iterate by that key. But we will only know this once a online DDL is requested!

Again, the assumption that we iterate on `PRIMARY KEY` is deeply rooted. I want to avoid rewriting huge sections of code. Instead, I propose an "override". We keep such terms as `lastPK` or `pkColumns`, but we understand that they may refer to non-PK unique keys. We introduce unique key hints in VReplication's filter/rule. Online DDL evaluates shared unique key and writes down its name into the rule. VReplication overrides `pkColumns` by reading the unique key columns in key order (ie how columns are ordered withing the unique key, not how they are ordered in the table). Most other information remains intact.

Importing `gh-ost` suite tests to validate correctness, and adding more tests.

This is work in progress.

## Related Issue(s)

#8179 

## Checklist


shlomi-noach:
While this is still WIP, I'm happy to get eyes on the proposed solution. For your convenience, this is the outline:

### proto
- vreplication's filter/rule has 2 new fields:
  - `source_unique_key` (mapped to `SourceUniqueKey` variable)
  - `target_unique_key` (mapped to `TargetUniqueKey` variable)
  - These are optional, and are only populated by Online DDL. Normally both will have the string value `PRIMARY`
  - But it's possible one or both will have the name of another unique key. The two names don't have to be identical, but Online DDL ensures that both will cover the exact same columns in same order (though a column is also allowed to be renamed, but that's just a nuance)
 
### Online DDL
- In `vrepl.go`, the part of Online DDL that analyzes a VReplication-based migration, we analyze both source and target table for all existing unique keys. We rank unique keys from best to worst. `PRIMARY` is always "best" because it's the clustered index hence by definition most performant key. Then, we prefer keys that use integers, over keys that use strings, etc.
- `vrepl.go` then intersects unique keys from source table and unique keys from target table and determines which keys are _shared_. It's possible that a UK on the source table and a UK on the target table have different names. That's fine, as long as they cover same columns in same order.
- Based on ranking, and based on shared keys, `vrepl.go` determines the _chosen_ UK on source table, and _chosen_ UK on target table. This is what gets to aforementioned proto/rule.

### Streamer
- rowstreamer receives the streaming query. The Unique Key name for source table is encoded as a comment directive `/*vt+ ukName="name_of_the_unqiue_key" */`
- If the unique key is not specificed, or is `PRIMARY`, we fallback to standard behavior.
- Otherwise, rowstreamer investigates the unique key by querying `information_schema`, and gets the orderd list of columns in that unique key.
- rowstreamer from there on assumes those columns to be the PK columns. It still calls them `pkColumns` (I wanted to minimize code changes)
- From there on, everything happens implicitly: `SendQuery` is calculated by `pkColumns` which means rowstreamer reads the table by UK order, not PK order, and also sends the UK columns as `Fields`, and also calculates `lastPK` by UK columns (again, the name remains `lastPK` but reall ythis could be any UK).

### VReplicator / table plan builder
- Detects `SourceUniqueKey` in Rule, generates a comment directive `/*vt+ ukName="name_of_the_unqiue_key" */` which rowstreamer will later consume
- Detects `TargetUniqueKey` in Rule
- if `TargetUniqueKey` is trivial (ie `PRIMARY`) or unspecified, then nothing happens and we fallback to standard behavior (tables ar ealready analyzed with PK info)
- otherwise, for each stream/rule:
  - table plan builder uses vreplicator's `recalculatePKColsInfo` function to:
    - read list of columns in specified unique key
    - re-evaluate colsInfo: make a copy of colsInfo, reorder columns by UK order, reassign `IsPK`
  - using the modified colsInfo, run `analyzePK` to evaluate `pkCols`. They're still called `pkCols`, but they may now identify the columns of any UK.
  - `pkCols` are then used, as usual and without change, to generate queries.

### Notes

- I believe I have modified the minimal set of code required for supporting non-PK columns
- The use of comment directives is temporary and is my easiest way to communicate hints from vreplication to streamer. We can change that later.
- I've added some tests from `gh-ost` test suite, which pass. I've added tests to `vrepl_stress` endtoend test, which pass. But I'm still unsure if I've generate da strong enough test case.
- All thoughts welcome. I certainly need this logic reviewed.

cc @rohit-nayak-ps @sougou @deepthi

shlomi-noach:
Added a new `endtoend` test: `vrepl_stress_suite`:

- Tests a VRreplication based migration
- uses `online -vreplication-test-suite -skip-topo` strategy, so that we end up with `_before` and `_after` tables, just like `vrepl_suite`
- Runs multiple `ALTER TABLE` cases, under "stress" (concurrent INSERT/DELETE/UPDATE operations) for a few seconds
- Specifically checks for migrations based on non-PRIMARY KEY. Uses various unique keys in various combinations of columns, such that order is either guaranteed to be consistent with original PRIMARY, guaranteed to be opposite (negative values), guaranteed to be random (text checksum values), compound keys, etc.
- Expects before/after tables to be completely identical.

shlomi-noach:
I'm satisfied that the new test suite validates the new non-PK behavior.

shlomi-noach:
`endtoend` test suite now use a low value for `-vstream_packet_size` (4096) to ensure we run multiple vstream iterations per table. This ensures we use the codepath in vcopier, that examines lastPK.

shlomi-noach:
I increased concurrency and load on the `endtoend` tests and am now seeing (at least locally) failures on text-based unique key replication.

shlomi-noach:
Failure on text-based `UNIQUE KEY`s seems to reproduce, which is a good thing I guess (bummer, I thought I had it solved).

shlomi-noach:
Further increasing the stress load, I also see failures with negative order UK. This is inline with the text-based (random order) UK failures.

shlomi-noach:
Hmmmm! Thankfully the first test I added was "trivial PK" where we just test normal PK without changes -- _and that fails, too_.

So we're looking at a more colossal error, not necessarily related to PK/UK isues.

shlomi-noach:
Sample failure for "trivial PK" test:
```
                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -30,3 +30,3 @@
                                 58     -58     4aab36bd_58     748398  NULL            2021-06-27 09:57:32     1
                                -60     -60     9584c1d6_60     516033  NULL            2021-06-27 09:57:32     1
                                +60     -60     9584c1d6_60     516033  NULL            2021-06-27 09:57:32     0
                                 62     -62     8a551dcf_62     745565  NULL            2021-06-27 09:57:31     2
                                @@ -78,3 +78,2 @@
                                 149    -149    e22231b3_149    220438  NULL            2021-06-27 09:57:33     0
                                -151    -151    00fb4427_151    258663  NULL            2021-06-27 09:57:34     0
                                 152    -152    655e36d2_152    621677  NULL            2021-06-27 09:57:32     0
                                @@ -336,3 +335,3 @@
                                 647    -647    5688e253_647    356355  NULL            2021-06-27 09:57:31     0
                                -651    -651    c74c2e89_651    281217  NULL            2021-06-27 09:57:25     4
                                +651    -651    c74c2e89_651    281217  NULL            2021-06-27 09:57:25     3
                                 652    -652    a896c975_652    546040  NULL            2021-06-27 09:57:28     3
                                @@ -858,3 +857,2 @@
                                 1660   -1660   f567bcc2_1660   5034    NULL            2021-06-27 09:57:34     0
                                -1663   -1663   98d3a4d1_1663   755896  NULL            2021-06-27 09:57:34     0
                                 1664   -1664   5db405cd_1664   973915  NULL            2021-06-27 09:57:32     0
                                @@ -1008,2 +1006,3 @@
                                 1964   -1964   9ee1701d_1964   643588  NULL            2021-06-27 09:57:34     0
                                +1968   -1968   8c96939f_1968   20687   NULL            2021-06-27 09:57:33     0
                                 1970   -1970   dd6f7137_1970   452018  NULL            2021-06-27 09:57:30     0
                                @@ -1038,3 +1037,3 @@
                                 2033   -2033   8c6cf656_2033   807633  NULL            2021-06-27 09:57:31     1
                                -2037   -2037   dd94e2b0_2037   699751  NULL            2021-06-27 09:57:31     2
                                +2037   -2037   dd94e2b0_2037   699751  NULL            2021-06-27 09:57:31     1
                                 2038   -2038   5af08eae_2038   717591  NULL            2021-06-27 09:57:31     2
                                @@ -1076,3 +1075,2 @@
                                 2120   -2120   b2a1e74e_2120   47155   NULL            2021-06-27 09:57:32     0
                                -2121   -2121   aacd07af_2121   104928  NULL            2021-06-27 09:57:34     0
                                 2122   -2122   8901f05a_2122   793351  NULL            2021-06-27 09:57:33     0
                                @@ -1148,3 +1146,3 @@
                                 2268   -2268   59462375_2268   697931  NULL            2021-06-27 09:57:24     2
                                -2270   -2270   256519b3_2270   268536  NULL            2021-06-27 09:57:32     1
                                +2270   -2270   256519b3_2270   268536  NULL            2021-06-27 09:57:32     0
                                 2275   -2275   d41a4c9a_2275   393016  NULL            2021-06-27 09:57:28     2
                                @@ -1325,2 +1323,3 @@
                                 2600   -2600   2c7dc50a_2600   977047  NULL            2021-06-27 09:57:31     3
                                +2601   -2601   623a4048_2601   68058   NULL            2021-06-27 09:57:32     0
                                 2602   -2602   48eb2c7b_2602   744662  NULL            2021-06-27 09:57:29     3
                                @@ -1330,3 +1329,3 @@
                                 2609   -2609   85e01d5a_2609   765881  NULL            2021-06-27 09:57:30     0
                                -2610   -2610   cc622662_2610   498893  NULL            2021-06-27 09:57:31     2
                                +2610   -2610   cc622662_2610   498893  NULL            2021-06-27 09:57:31     1
                                 2612   -2612   2fce3d1f_2612   251477  NULL            2021-06-27 09:57:32     0
                                @@ -1349,3 +1348,3 @@
                                 2651   -2651   d41eebe3_2651   232602  NULL            2021-06-27 09:57:33     0
                                -2652   -2652   0f8a83dd_2652   610513  NULL            2021-06-27 09:57:24     4
                                +2652   -2652   0f8a83dd_2652   610513  NULL            2021-06-27 09:57:24     3
                                 2653   -2653   c95c5411_2653   437884  NULL            2021-06-27 09:57:32     0
                                @@ -1857,3 +1856,2 @@
                                 3649   -3649   3f9811a9_3649   950012  NULL            2021-06-27 09:57:28     2
                                -3650   -3650   c325f0ed_3650   345566  NULL            2021-06-27 09:57:34     0
                                 3651   -3651   8f6c850b_3651   424790  NULL            2021-06-27 09:57:33     0
                                @@ -2064,3 +2062,3 @@
                                 4051   -4051   2a71ac6b_4051   11497   NULL            2021-06-27 09:57:31     1
                                -4053   -4053   6d887410_4053   417799  NULL            2021-06-27 09:57:33     1
                                +4053   -4053   6d887410_4053   417799  NULL            2021-06-27 09:57:33     0
                                 4056   -4056   23f82d02_4056   495518  NULL            2021-06-27 09:57:26     0
                Test:           TestSchemaChange/trivial_PK/migrate
                Messages:       results mismatch: (
                                                SELECT * FROM stress_test_before order by id, id_negative, rand_text, rand_num
                                        ) and (
                                                SELECT * FROM stress_test_after order by id, id_negative, rand_text, rand_num
                                        )
```

shlomi-noach:
Confirmed this is a general issue with vreplication/online ddl that has not been reliably detected until now. Opening a new PR with `vrepl_stress` failing tests to observe and fix.

shlomi-noach:
Incorporated suggested fix from https://github.com/vitessio/vitess/pull/8390

shlomi-noach:
With the fix from #8390 the only failures seen now are from text-based keys; and the reason is actually because we decorate text columns with `convert(the_column using utf8)`, so that's a bit tangential to the PR -- fixing.

shlomi-noach:
`convert`... issue mentioned above is fixed. Now I see increased lag when the unique key is textual.

shlomi-noach:
text-column based unique keys show significant vreplication lag; the reason they fail is on timeouts. With enough timesots the tests pass.

`convert(col using utf8)` seems to be the reason. If I comment out the `convert()` part, then text columns pass tests just fine.

So I'm looking at how queries might perform poorly with `convert()`. Thing is queries look fine. Examples:

```sql
delete from _6a1302b0_d81c_11eb_97d6_0a43f95f28a3_20210628142332_vrepl where rand_text=(convert('df89b4ba_187' using utf8mb4))

update _6a1302b0_d81c_11eb_97d6_0a43f95f28a3_20210628142332_vrepl set id=165, id_negative=-165, rand_num=295042, nullable_num=null, op_order=29074, hint_col=convert('' using utf8mb4), created_timestamp='2021-06-28 14:23:43', updates=1 where rand_text=(convert('9a9ab0ee_165' using utf8mb4)):

insert into _6a1302b0_d81c_11eb_97d6_0a43f95f28a3_20210628142332_vrepl(id,id_negative,rand_text,rand_num,nullable_num,op_order,hint_col,created_timestamp,updates) values (3235,-3235,convert('51942956_3235' using utf8mb4),293698,null,29063,convert('' using utf8mb4),'2021-06-28 14:23:44',0)
```

All these queries should not be affected by the `convert` function because the function only operates on constants, never on columns.

shlomi-noach:
ahhhhh, the clause `where rand_text=(convert('38c76882_1775' using utf8mb4))` actually forces a hidden casting on `rand_text` thereby causing full table scan. OMG OMG. Looking into it.

shlomi-noach:
And that's because the tested column is `utf8`, not `utf8mb4`.

AHHHH! The mystery is finally solved.

shlomi-noach:
All fixes are done, and this PR should be good.

shlomi-noach:
While keeping this in `Draft`, I wish to solicit reviews at least for the thought process and the design. See both initial comment as well as https://github.com/vitessio/vitess/pull/8364#issuecomment-865840490 for breakdown.

shlomi-noach:
Ping for review

shlomi-noach:
Question: as mentioned above:

> vreplication's filter/rule has 2 new fields:
>  source_unique_key (mapped to SourceUniqueKey variable)
>  target_unique_key (mapped to TargetUniqueKey variable)

- when `source_unique_key` is non empty and non-`PRIMARY`, that means `rowstreamer` needs to read `information_schema.columns` to see what columns are covered by the unique key and in what order.
- similarly, when `target_unique_key` is non empty and non-`PRIMARY`, that means `vreplicator` needs to read `information_schema.columns` to see what columns are covered by the unique key and in what order.

The module that populates these two values is `onlineddl/vrepl`, which already knows the column identities and order for both keys. I wonder if instead of `source_unique_key` (string) and `target_unique_key` (string) we should instead have `source_unique_key_columns` (string array) and `target_unique_key_columns` (string array).

This way the one who own the decision to use a specific unique key can already dictate the columns; the vreplication components don't need to read `information_schema`. Downside is that vreplication will need to _trust_ that the given columns are, indeed, covered by a unique key. Also, it loses context to the name/identity of the key it actually uses.

Thoughts?

shlomi-noach:
Merged https://github.com/vitessio/vitess/pull/8420 into this PR. From #8420's main comment:

--- 

## Description

A take/twist on https://github.com/vitessio/vitess/pull/8364/, per [comment](https://github.com/vitessio/vitess/pull/8364/#issuecomment-874625804)

Mostly exactly the same as #8364 , except:

- `Rule`'s proto has `source_unique_key_columns` instead of `source_unique_key`, and `target_unique_key_columns` instead of `target_unique_key`.
- rowstreamer doesn't need to inspect `information_schema.columns`, since it relies on  `source_unique_key_columns`
- vreplicator likewise doesn't need to inspect `information_schema.columns`, since it relies on `target_unique_key_columns`

rohit-nayak-ps (approved):
lgtm.

Very nice. I like the fact there are minimal changes to the core workflow due to this!

rohit-nayak-ps:
Just observing here that this feature is only implemented for Online DDL, though internally vreplication supports it via this PR. If we want to extend this to  vreplication workflows (movetables/reshard/...) we will need to generate the filter hints that Online DDL generates while creating those workflows.

shlomi-noach:
> Just observing here that this feature is only implemented for Online DDL, though internally vreplication supports it via this PR.

Correct. There has to be some logic above VReplication to set those fields.

PR #8366, opened 2021-06-22
Ignore Internal Operation Tables From Vdiff

makinje16:

## Description

This PR modifies VDiff by now ignoring tables that pass the `IsInternalOperationTableName` method.
## Checklist

### DIFF
--- a/go/vt/wrangler/vdiff.go
+++ b/go/vt/wrangler/vdiff.go
@@ -37,6 +37,7 @@ import (
 	"vitess.io/vitess/go/vt/key"
 	"vitess.io/vitess/go/vt/log"
 	"vitess.io/vitess/go/vt/logutil"
+	vtschema "vitess.io/vitess/go/vt/schema"
 	"vitess.io/vitess/go/vt/sqlparser"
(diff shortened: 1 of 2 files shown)


go/vt/wrangler/vdiff.go
```diff
@@ -37,6 +37,7 @@ import (
 	"vitess.io/vitess/go/vt/key"
 	"vitess.io/vitess/go/vt/log"
 	"vitess.io/vitess/go/vt/logutil"
+	vtschema "vitess.io/vitess/go/vt/schema"
```
ajm188, line 40: I don't think we should rename this import. I read it as `vschema` several times and kept wondering why IsInteralOperationTableName` was in that package.

If there's a local variable in this file preventing us from using this name, I think we should rename that variable.

ajm188 (commented):
very excited for this!

go/vt/wrangler/vdiff.go
makinje16: unrenamed and should be in next commit

rohit-nayak-ps (approved):
lgtm

PR #8378, opened 2021-06-23
[consultopo] Add flag to limit MaxConnsPerHost in consultopo servers

ajm188:

## Description

This PR introduces a new flag to control the `MaxConnsPerHost` setting in the consul client's http.Transport, to prevent unbounded connection growth during highly concurrent topo calls

## Related Issue(s)

None yet, will file after testing

## Checklist

### DIFF
--- a/go/vt/topo/consultopo/server.go
+++ b/go/vt/topo/consultopo/server.go
@@ -36,6 +36,7 @@ import (
 
 var (
 	consulAuthClientStaticFile = flag.String("consul_auth_static_file", "", "JSON File to read the topos/tokens from.")
+	maxConnsPerHost            = flag.Int("consultopo_max_conns_per_host", 0, "limit the MaxConnsPerHost used in the consul client's http Transport. Zero means no limit.")
 )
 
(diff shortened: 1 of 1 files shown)


go/vt/topo/consultopo/server.go
```diff
@@ -36,6 +36,7 @@ import (
 
 var (
 	consulAuthClientStaticFile = flag.String("consul_auth_static_file", "", "JSON File to read the topos/tokens from.")
+	maxConnsPerHost            = flag.Int("consultopo_max_conns_per_host", 0, "limit the MaxConnsPerHost used in the consul client's http Transport. Zero means no limit.")
```
brirams, line 39: might be worth adding a comment around why this exists. If i were to see this without our context, I wouldn't be sure if i need to set it or not.


general Q: do we have any idea how big this gets in practice(i.e. what number of conns tips us over the edge)?

brirams (commented):
my gut says this is too entangled to write any sort of meaningful test but would be awesome if it were a drop in the bucket!

Issue #8259, opened 2021-06-03
MoveTables / SwitchWrites vreplication error: "ambiguous vindex reference: hash"


dasl-:
#### Overview of the Issue
When certain conditions are met, vreplication can fail during a MoveTables workflow with an error message: `ambiguous vindex reference: hash`. (see: https://github.com/vitessio/vitess/blob/4c33b2b5a90dc200e6cc649681cbace2520725bd/go/vt/vtgate/vindexes/vschema.go#L559 )

#### Reproduction Steps
1. Set up three keyspaces: `test_unsharded`, `test_sharded1`, `test_sharded2`.
2. Each keyspace will have the same tables, although the tables in `test_sharded2` are not important. Deploy this sql to all keyspaces:
```sql
CREATE TABLE `ledger` (
  `ledger_id` bigint(20) unsigned NOT NULL DEFAULT '0',
  `user_id` bigint(20) unsigned NOT NULL DEFAULT '0',
  `ledger_type` tinyint(3) unsigned NOT NULL DEFAULT '0',
  `create_date` int(11) NOT NULL DEFAULT '0',
  `update_date` int(11) NOT NULL DEFAULT '0',
  PRIMARY KEY (`ledger_id`),
  UNIQUE KEY `key_userid_ledgertype` (`user_id`,`ledger_type`),
  KEY `key_update_date` (`update_date`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci ROW_FORMAT=COMPRESSED KEY_BLOCK_SIZE=8
```
(note: this example could be modified to have a more minimal set of columns)

3. Setup a vschema on both `test_sharded1` and `test_sharded2`:
```javascript
{
    "sharded": true,
    "vindexes": {
        "hash": {
            "type": "hash"
        }
    },
    "tables": {
        "ledger": { "column_vindexes": [ { "name": "hash", "column": "ledger_id" } ] }
    }
}
```
4. Start a MoveTables workflow from `test_unsharded` -> `test_sharded1`:
```
vtctl MoveTables -tablet_types=replica -workflow=billing2payments test_unsharded test_sharded1 'ledger'
```

5. Insert some data into `test_unsharded` via vitess:
```sql
insert into ledger (ledger_id, user_id) values (1,1);
insert into ledger (select ledger_id+1, user_id+1, ledger_type, create_date, update_date from ledger order by user_id desc limit 1);
```
[...]

PR #8385, opened 2021-06-25
VReplication Reverse Workflows: add keyspace scope to vindex while creating reverse vreplication streams

rohit-nayak-ps:
## Description

For sharded streams we add filters like `select * from customer where in_keyrange(customer_id, 'keyspace1.hash', '-7f')` where the vindex is scoped by the source keyspace. This is needed since other keyspaces might also contain vindexes of the same name and using just the vindex name would create an ambiguity. 

This logic was missing while creating reverse streams which this PR adds.

## Related Issue(s)
https://github.com/vitessio/vitess/issues/8259

## Checklist


deepthi (approved):
LGTM

Issue #8386, opened 2021-06-25
Vitess + MySQL group replication


sonne5:
### Problem
Vitess currently supports async and semi-sync replication, both technologies have been battle-tested in production. There is another mysql replication solution: group replication. It relies on Paxos to coordinate among servers and provide consistency guarantees. It handles the failover automatically within the group.

It is subjective to choose between semi-sync and group replication. Some tooling might need some tweaks, but there is nothing fundamentally blocks Vitess to adopt group replication - it can be an additional feature for Vitess. 

The current Vitess’s implementation does not recognize the group concept. Hence Vitess cannot route requests properly or report lag for tablet healthiness checks.

To address the gap between Vitess and Group Replication, we propose a new component, VTGR (Vitess Group Replication), to do the orchestration and reconciliation between Vitess and MySQL group.

### Requirements
1. VTTablet needs to understand and report the lag under MySQL group replication
2. VTGR handles group bootstrap and maintain the group memberships
3. VTGR handles necessary Vitess and MySQL failover and reconcile between the two

### Solution
To satisfy the requirements, the high level solutions are listed below.

#### Group lag
There is a flavor flag that can be used to override the current status reporting behavior on VTTablet ( ``-db_flavor MysqlGR``). Everything else can remain unchanged.

#### Group orchestration (VTGR)
Every Vitess shard (`TabletType_MASTER`, `TabletType_REPLICA`, `TabletType_RDONLY`) composes a MySQL group.

VTGR is the component that is responsible for group orchestration. It relies on shard lock from topo server for coordination. 

VTGR is running in a loop, where it first does a ``diagnose`` on the shard with MySQL group and Vitess topology information. Based on the diagnose result, VTGR will execute ``repair`` with a shard level lock.

The basic functionalities are the followings:
##### Group maintenance
VTGR will pull hosts for a shard through topo server and fanout to all nodes to collect MySQL information:
- If there is no active MySQL group, VTGR will bootstrap one.
- If there is a replica in the shard that is not in the group, VTGR will add the node into the group.

##### Failover
Since MySQL group will do failover automatically, VTGR is responsible for reconciling Vitess and MySQL group to make sure the primary VTTablet is pointing to a writable primary sql node. To do that, VTGR uses group information to identify primary sql nodes and compare with the primary tablet from topo server:
- If there is no primary tablet, VTGR will set the primary according to the MySQL group.
[...]

PR #8387, opened 2021-06-26
VTGR: Vitess + MySQL group replication

narcsfz:
## Description

This adds single primary mysql group replication support for Vitess. Specifically it does the followings:
- VTTablet needs to understand and report the lag under MySQL group replication
- VTGR handles group bootstrap and maintain the group memberships
- VTGR handles necessary Vitess and MySQL failover and reconcile between the two

The structure of the PR is:

|directory| purpose|
|-----|-----|
|vtgr/config | The configurations for VTGR
vtgr/controller/diagnose.go | VTGR diagnose the status for a Vitess shard
vtgr/controller/repair.go | VTGR repair the shard based on diagnose result
vtgr/controller/group.go | VTGR store MySQL group information
vtgr/controller/refresh.go | VTGR refresh tablet records from topo server
vtgr/db | DB client that talks to MySQL

## Related Issue(s)

https://github.com/vitessio/vitess/issues/8386

## Checklist


go/vt/vtgr/controller/diagnose.go
```diff
@@ -0,0 +1,561 @@
...
+	select {
+	case <-pingCtx.Done():
+		log.Errorf("Ping abort timeout %v", *pingTabletTimeout)
+		return false
+	case err := <-c:
+		if err != nil {
+			log.Errorf("Ping error host=%v: %v", instance.instanceKey.Hostname, err)
+		}
+		return err == nil
+	}
```
vmg, line 348: This is bad but it's actually our fault. Once https://github.com/vitessio/vitess/pull/8368 is merged we should be able tof ix the TMC API so that the initial dial is context-aware and we don't need this hack.

ajm188: I don't think switching to DialContext there will actually fix this issue, though. From [the docs](https://pkg.go.dev/google.golang.org/grpc#DialContext)

> By default, it's a non-blocking dial (the function won't wait for connections to be established, and connecting happens in the background). To make it a blocking dial, use WithBlock() dial option.
>
> In the non-blocking case, the ctx does not act against the connection. It only controls the setup steps.

I don't think we should unilaterally use `WithBlock` as that will result in slowdowns in the rest of the tmc use-cases, so what I think we should do is a follow-up refactor to the tmc api (and merge both #8368 (pending review / further changes) and this PR as-is) to allow callers to specify dial options to use, something like:

```go
tmc := tmclient.TabletManagerClient().WithDialOptions(grpc.WithBlock())
```

vmg: I think this sounds about right: the comment that @narcsfz added to this method was a bit confusing, because it complains about dial/retry timeouts when (in theory) we shouldn't be seeing dial timeouts at all, since this API is not called with `WithBlock` and hence the actual connection will only happen during the actual call to `Ping`, and will be context-aware. I think what @narcsfz was seeing are the basic setup steps which by default are not controlled by context but _will be_ if we switch to `DialContext`.

narcsfz: Thanks @vmg @ajm188. I'm happy to leave a TODO here and refactor it later if that makes sense. Since it is hard to reproduce in unit test, I also need to set up the test case that can reproduce the problem to test the change out.

go/cmd/vtgr/main.go
```diff
@@ -0,0 +1,24 @@
+package main
```
systay, line 1: license header missing

go/mysql/flavor_mysqlgr.go
```diff
@@ -0,0 +1,222 @@
+/*
+Copyright 2019 The Vitess Authors.
```
systay, line 2: should say 2021, right?

go/mysql/flavor_mysqlgr_test.go
```diff
@@ -0,0 +1,38 @@
+package mysql
```
systay, line 1: license header missing

go/mysql/flavor_mysqlgr.go
```diff
@@ -0,0 +1,222 @@
...
+	qr, err := c.ExecuteFetch(query, 100, true /* wantfields */)
+	if err != nil {
+		return err
+	}
+	// if group replication related query returns 0 rows, it means the group replication is not set up
+	if len(qr.Rows) == 0 {
+		return ErrNoGroupStatus
+	}
+	if len(qr.Rows) > 1 {
+		return fmt.Errorf("unexpected results for %v: %v", query, qr.Rows)
```
systay, line 205: we try to use `vterrors.Errors` instead of `fmt`

go/mysql/flavor_mysqlgr.go
narcsfz: yep, updated

go/mysql/flavor_mysqlgr.go
narcsfz: changed to `vterrors.Errorf`

go/mysql/flavor_mysqlgr_test.go
narcsfz: Added

narcsfz:
Addressed your comments ptal @systay

go/vt/vtgr/db/mysql.go
```diff
@@ -0,0 +1,535 @@
...
+package db
+
+import (
+	"errors"
+	"flag"
+	"fmt"
+	"strconv"
+	"strings"
+
+	gouuid "github.com/pborman/uuid"
```
deepthi, line 26: We are using `github.com/google/uuid` elsewhere in the code. Any reason you can't use that?

go/vt/vtgr/db/mysql.go
```diff
@@ -0,0 +1,535 @@
...
+	"vitess.io/vitess/go/mysql"
+	"vitess.io/vitess/go/vt/log"
+	"vitess.io/vitess/go/vt/orchestrator/config"
+	"vitess.io/vitess/go/vt/orchestrator/db"
+	"vitess.io/vitess/go/vt/orchestrator/external/golib/sqlutils"
+	"vitess.io/vitess/go/vt/orchestrator/inst"
+)
+
+var (
+	configFile     = flag.String("db_config", "", "db config file name")
```
deepthi, line 37: naming suggestion: flag name should be clear that it will interpreted as a file name. Is the full path expected? If yes, it should be documented as such.

go/vt/vtgr/db/mysql.go
```diff
@@ -0,0 +1,535 @@
...
+type GroupView struct {
+	TabletAlias       string
+	MySQLHost         string
+	MySQLPort         int
+	GroupName         string
+	UnresolvedMembers []*GroupMember
+}
+
+// SQLAgentImp implements sqlAgent
+type SQLAgentImp struct {
```
deepthi, line 120: naming suggestion: `SQLAgentImpl` or `SqlAgentImpl`.
Also looks like the interface has been renamed to `Agent` but comment still refers to `sqlAgent`.

go/vt/vtgr/vtgr.go
```diff
@@ -0,0 +1,172 @@
...
+var (
+	refreshInterval      = flag.Duration("refresh_interval", 10*time.Second, "refresh interval to load tablets")
+	scanInterval         = flag.Duration("scan_interval", 3*time.Second, "scan interval to diagnose and repair")
+	scanAndRepairTimeout = flag.Duration("scan_repair_timeout", 3*time.Second, "time to wait for a Diagnose and repair operation")
+	vtgrConfigFile       = flag.String("vtgr_config", "", "config file for vtgr")
+
+	localDbPort = flag.Int("db_port", 0, "local mysql port, set this to enable local fast check")
+)
+
+// VTGR manages VTGR
```
deepthi, line 46: A more descriptive comment will be good.

go/vt/vtgr/vtgr.go
```diff
@@ -0,0 +1,172 @@
...
+type VTGR struct {
+	// Shards are all the shards that a VTGR is monitored
```
deepthi, line 48:
```suggestion
	// Shards are all the shards that a VTGR is monitoring.
```

go/vt/vtgr/vtgr.go
```diff
@@ -0,0 +1,172 @@
...
+	// caller can choose to iterate the shards to scan and repair for more granule control (e.g., stats report)
```
deepthi, line 49:
```suggestion
	// Caller can choose to iterate the shards to scan and repair for more granular control (e.g., stats report)
```

go/vt/vtgr/vtgr.go
```diff
@@ -0,0 +1,172 @@
...
+	// instead of calling ScanAndRepair() directly
```
deepthi, line 50:
```suggestion
	// instead of calling ScanAndRepair() directly.
```

go/vt/vtgr/vtgr.go
```diff
@@ -0,0 +1,172 @@
...
+	for _, ks := range clustersToWatch {
+		if strings.Contains(ks, "/") {
+			// This is a keyspace/shard specification
+			input := strings.Split(ks, "/")
+			shards = append(shards, controller.NewGRShard(input[0], input[1], cellsToWatch, vtgr.tmc, vtgr.topo, db.NewVTGRSqlAgent(), config, *localDbPort))
+		} else {
+			// Assume this is a keyspace and find all shards in keyspace
+			shardNames, err := vtgr.topo.GetShardNames(ctx, ks)
+			if err != nil {
+				// Log the errr and continue
```
deepthi, line 92:
```suggestion
				// Log the error and continue
```

go/vt/vtgr/vtgr.go
```diff
@@ -0,0 +1,172 @@
...
+func newVTGR(ctx context.Context, ts controller.GRTopo, tmc tmclient.TabletManagerClient) *VTGR {
+	return &VTGR{
+		topo: ts,
+		tmc:  tmc,
+		ctx:  ctx,
+	}
+}
+
+// OpenTabletDiscovery opens connection with topo server
+// and triggers the first round of controller based on parameter
```
deepthi, line 66:
```suggestion
// and triggers the first round of controller based on specified cells and keyspace/shards.
```

deepthi (commented):
Well-written and well-documented contribution.
These comments are from a first pass. I will be doing another pass over the PR.

go/mysql/flavor_mysqlgr.go
```diff
@@ -0,0 +1,226 @@
...
+}
+
+// restartReplicationCommands returns the commands to stop, reset and start the replication.
+// for mysqlGRFlavor we don't need this functionality
+func (mysqlGRFlavor) restartReplicationCommands() []string {
+	return []string{}
+}
+
+// startReplicationUntilAfter will restart replication, but only allow it
+// to run until `pos` is reached. After reaching pos, replication will be stopped again
```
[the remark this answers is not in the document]
narcsfz: Updated to clarify those are disabled in mysqlGRFlavor

go/mysql/flavor_mysqlgr.go
```diff
@@ -0,0 +1,226 @@
...
+func (mysqlGRFlavor) startReplicationUntilAfter(pos Position) string {
+	return ""
+}
+
+// stopReplicationCommand returns the command to stop the replication.
+func (mysqlGRFlavor) stopReplicationCommand() string {
+	return "STOP GROUP_REPLICATION"
```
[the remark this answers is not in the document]
narcsfz: Updated to be same as `startReplicationCommand`

go/mysql/flavor.go
```diff
@@ -36,6 +36,9 @@ var (
 
 	// ErrNoPrimaryStatus means no status was returned by ShowPrimaryStatus().
 	ErrNoPrimaryStatus = errors.New("no master status")
+
+	// ErrNoGroupStatus means no status for group replication.
+	ErrNoGroupStatus = errors.New("no group status")
```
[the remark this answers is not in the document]
narcsfz: We can - adding it here because this seems to be the place where we declare all the errors. Happy to move into flavor_mysqlgr.go if that's better.

go/test/endtoend/vtgr/test_config.json
```diff
@@ -0,0 +1,9 @@
+{
+  "MySQLTopologyUser": "orc_client_user",
+  "MySQLTopologyPassword": "orc_client_user_password",
```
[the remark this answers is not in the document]
narcsfz: We reuse how `vtorc` setup DB connections, and we need a user/password to set up the connection. Since VTGR is also an orchestration component and `orc_client_user` has all the privileges we need - that's why we use `orc_client` here instead of making a new one.

go/vt/vtgr/controller/diagnose.go
```diff
@@ -0,0 +1,577 @@
...
+
+func (shard *GRShard) primaryTabletAlias() string {
+	primary := shard.findShardPrimaryTablet()
+	if primary == nil {
+		return "UNKNOWN"
+	}
+	return primary.alias
+}
+
+// disconnectedInstance iterate all known the replica records
```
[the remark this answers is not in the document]
narcsfz: Fixed

go/vt/vtgr/controller/diagnose.go
```diff
@@ -0,0 +1,577 @@
...
+
+// We use fanout in two cases:
+// 1. FetchGroupView GTIDs to find a candidate for failover.
+// If a node is not healthy it should not be considered as a failover candidate
+//
+// 2. FetchGroupView group member status to see if we need to bootstrap a group,
+// either for the first time or rebuild a group after all the nodes are died.
+//
+// caller will be responsible to decide if they want to tolerate errors from the fanout call
+func (shard *GRShard) fanout(task func(instance *grInstance, wg *sync.WaitGroup, er concurrency.ErrorRecorder)) *concurrency.AllErrorRecorder {
```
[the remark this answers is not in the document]
narcsfz: done

go/vt/vtgr/controller/error.go
```diff
@@ -0,0 +1,25 @@
...
+limitations under the License.
+*/
+
+package controller
+
+import "errors"
+
+var (
+	errMissingPrimaryTablet = errors.New("no_primary_tablet_available")
+	errMissingGroup         = errors.New("no_mysql_group")
```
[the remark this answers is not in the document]
narcsfz: Updated

go/vt/vtgr/db/mysql.go
narcsfz: https://github.com/vitessio/vitess/blob/main/go/vt/workflow/manager.go#L28 is using github.com/pborman/uuid as well :-)

I will change to google/uuid since they generate same results for our purpose: https://play.golang.org/p/dAEaWBqxlgT

go/vt/vtgr/db/mysql.go
narcsfz: Updated

go/vt/vtgr/vtgr.go
narcsfz: Added more comment

go/mysql/flavor.go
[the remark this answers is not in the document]
deepthi: yes please.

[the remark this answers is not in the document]
narcsfz: Sure, done

narcsfz:
There is a test failure after a lot of retries. Looks like it's related to some network issue: https://github.com/vitessio/vitess/pull/8387/checks?check_run_id=3023339855

I will try to rerun it later, but it should not affect the review cc @deepthi

go/mysql/flavor_mysqlgr.go
```diff
@@ -0,0 +1,226 @@
...
+	}
+	if len(qr.Rows) > 1 {
+		return vterrors.Errorf(vtrpc.Code_INTERNAL, "unexpected results for %v: %v", query, qr.Rows)
+	}
+	return onResult(qr.Rows[0])
+}
+
+// masterStatus returns the result of 'SHOW MASTER STATUS',
+// with parsed executed position.
+func (mysqlGRFlavor) masterStatus(c *Conn) (PrimaryStatus, error) {
```
deepthi, line 218: This has been renamed to `primaryStatus` in the interface.

go/vt/vtgr/controller/diagnose_test.go
```diff
@@ -0,0 +1,615 @@
...
+	}
+	var sqltests = []struct {
+		name          string
+		expected      DiagnoseType
+		errMessage    string
+		config        *config.VTGRConfig
+		inputs        []data
+		removeTablets []string // to simulate missing tablet in topology
+	}{
+		{name: "healthy shard", expected: DiagnoseTypeHealthy, errMessage: "", inputs: []data{
```
deepthi, line 238: I didn't verify this, but I'm assuming all DiagnoseTypes are covered by this unit test.

go/vt/vtgr/controller/refresh_test.go
```diff
@@ -0,0 +1,145 @@
...
+	"vitess.io/vitess/go/vt/logutil"
+
+	"github.com/stretchr/testify/assert"
+
+	topodatapb "vitess.io/vitess/go/vt/proto/topodata"
+
+	"vitess.io/vitess/go/vt/topo"
+
+	"github.com/golang/mock/gomock"
+)
```
deepthi, line 41: Why is this formatted like this? Can you run `goimports` on all the new files?
I'm concerned that this may cause problems for someone who does have the pre-commit hooks installed.
As an aside we probably need a goimports CI check.

go/vt/vtgr/controller/repair_test.go
```diff
@@ -0,0 +1,1242 @@
...
+	"github.com/stretchr/testify/assert"
+
+	"vitess.io/vitess/go/vt/orchestrator/inst"
+
+	topodatapb "vitess.io/vitess/go/vt/proto/topodata"
+	"vitess.io/vitess/go/vt/topo"
+
+	gomock "github.com/golang/mock/gomock"
+
+	db2 "vitess.io/vitess/go/vt/vtgr/db"
```
deepthi, line 47: super nit-pick: db2 -> db

deepthi (commented):
Mostly LGTM. We can merge once feedback has been addressed and tests are passing.

go/cmd/vtgr/main.go
```diff
@@ -0,0 +1,37 @@
...
+	flag.Parse()
+
+	// openTabletDiscovery will open up a connection to topo server
+	// and populate the tablets in memory
+	vtgr := vtgr.OpenTabletDiscovery(context.Background(), nil, strings.Split(*clustersToWatch, ","))
+	vtgr.RefreshCluster()
+	vtgr.ScanAndRepair()
+
+	// block here so that we don't exit directly
+	select {}
```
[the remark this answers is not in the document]
narcsfz: we can SIGTERM the process
my linter asked me to use `select {}` here

go/mysql/flavor_mysqlgr.go
narcsfz: good catch, fixed

go/vt/vtgr/controller/diagnose_test.go
narcsfz: yeah, we have all 10 types (expect unknown) covered:
```
$ grep "expected: DiagnoseType" go/vt/vtgr/controller/diagnose_test.go | awk -F','   '{print $2}' | sort | uniq | wc -l
      10
```

go/vt/vtgr/controller/repair_test.go
narcsfz: Done

go/vt/vtgr/controller/refresh_test.go
narcsfz: I am not sure.. goimports seems to be happy with it... anyway i fixed the order here

go/cmd/vtgr/main.go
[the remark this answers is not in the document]
deepthi: ok, that's fine.

deepthi (approved):
Nice work!

PR #8388, opened 2021-06-26
VReplication: Add ability to tag workflows

rohit-nayak-ps:
## Description
Users have requested the ability to group workflows. Some use cases:
* Sharded->sharded Reshards that are done incrementally, one per source shard, for reducing the risk when workflows have fatal failure and need to be restarted from scratch
* Materialize streams that require custom filters (for PII implementations, say) for each table
* When multiple tables are to be moved and these are split across workflows for performance reasons to work around the current limitation where tables are only copied one at a time per workflow

This PR adds the ability to tag workflows in a simple and extensible way. It adds a `tags` column to the `_vt.vreplication` table. Adding multiple tags is left as an implementation detail: the most obvious way would be to use a list of comma-separated tags in that column. 

This implementation also allows users to tag each stream in a workflow separately by specifying separate list of tags for each stream in the workflow on each target.

To add tags at the workflow level use the `Workflow` command thus: `vtctlclient Workflow customer.cust2cust tags 'a,b'` to set the tags. Specify an empty string to reset the tags.

To set tags at the stream level use `VReplicationExec` thus: `vtctlclient VReplicationExec zone1-100 "update _vt.vreplication set tags = 'a,b' where workflow = 'cust2cust' and keyspace = 'customer'"`

The wrangler function `wr.WorkflowTagAction(ctx, keyspace, workflow, tags)` can be used for programmatically setting it in, say, `vtadmin`.

Note that tags do not affect the core functionality of vreplication workflows.

## Checklist


go/vt/vtctl/vtctl.go
```diff
@@ -489,7 +489,7 @@ var commands = []commandGroup{
 		"Workflow", []command{
 			{"Workflow", commandWorkflow,
```
ajm188, line 490: I think it would be a nice touch to be able to tag workflows directly as they're created with `Reshard/MoveTables` as well

go/vt/vtctl/vtctl.go
```diff
@@ -3506,14 +3506,29 @@ func commandWorkflow(ctx context.Context, wr *wrangler.Wrangler, subFlags *flag.
...
-	}
-	if action == "show" || action == "listall" {
-		return nil
+	var results map[*topo.TabletInfo]*sqltypes.Result
+	if action == "tags" {
+		tags := ""
+		if subFlags.NArg() == 3 {
+			tags = strings.ToLower(subFlags.Arg(2))
+		}
+		results, err = wr.WorkflowTagAction(ctx, keyspace, workflow, tags)
```
ajm188, line 3515: I think we should either allow multiple tags like `<ks.workflow> [tags [tags ...]]` or do some extra linting here. There's a subtle problem where if I were to do `vtctlclient Workflow ks.workflow tags tag1 tag2` then we wouldn't enter the `if subflags.NArg() == 3` branch, and it would silently delete all the tags for the workflow.

go/vt/wrangler/vexec.go
```diff
@@ -385,6 +396,8 @@ type ReplicationStatus struct {
 	TimeUpdated int64
 	// Message represents the message column from the _vt.vreplication table.
 	Message string
+	// Tags contain the tags specified for this stream
+	Tags string
```
ajm188, line 400: We'll also need to update the protos for this field and the corresponding impl in `vtctl/workflow/server.go`

ajm188 (commented):
this is looking good to me! a comment about cli, and then also about keeping the new apis in sync (which, yes, we should unify soon, but for now)

go/vt/vtctl/vtctl.go
rohit-nayak-ps: I wanted to keep the code footprint small to start with : otherwise we need to modify MoveTables/Reshard/Materialize for both v1 and v2 versions. In any case we would need the Workflow command for those who want to tag or retag it later. Is it usable for you with the current implementation?

go/vt/vtctl/vtctl.go
rohit-nayak-ps: For multiple tags, I had thought of keeping it simple (as mentioned in the PR description): you would implement it by joining the tags as `tag1,tag2` before providing it to the `Workflow` command. This limits the CLI requirement to just setting the tag value, either to a single value, csv for multiple tags or empty to clear it. I can fix the validation so that it will error if more tags are provided in the command, so that there is no silent deletion.

go/vt/wrangler/vexec.go
rohit-nayak-ps: Added and verified that the tags are shown in the JSON response for the workflow streams in vtadmin.

proto/vtctldata.proto
```diff
@@ -127,6 +127,8 @@ message Workflow {
     // if there are N logs that exist for the stream, and we fail to fetch the
     // ith log, we will still return logs in [0, i) + (i, N].
     string log_fetch_error = 14;
+    // Note that there is a single tags column: multiple tags can be "stuffed" into this as a CSV, for example
+    string tags = 15;
```
ajm188, line 131: imo this should be `repeated string tags = 15`, and then we do the conversion once in workflow/server from csv string to string slice, and then everyone downstream from that gets the nicely-structured list rather than needing to re-split

go/vt/vtctl/vtctl.go
ajm188: Yeah this is usable now; okay to come back to this later!

go/vt/vtctl/vtctl.go
ajm188: (just closing the loop on this): we chatted and your implementation, plus the fix for silent deletion, makes sense!

ajm188 (commented):
Just one small thing about the protobuf type change and then this lgtm whenever you want to move out of draft

proto/vtctldata.proto
rohit-nayak-ps: Good idea: done.

rohit-nayak-ps:
> Just one small thing about the protobuf type change and then this lgtm whenever you want to move out of draft

Ready for review

ajm188 (approved):
lgtm

PR #8392, opened 2021-06-28
Remove tablet record from Topo server when Vttablet shuts down

GuptaManan100:
## Description
This PR adds the support for removing the tablet record from the topo server whenever a non-primary vttablet shuts down.

## Checklist


ajm188:
What's the rationale behind this change? I'm concerned this is going to break any users that have both checks/alerts that depend on the topo records and also restart vttablet from time to time as part of deploys.

deepthi:
Why would restarts be a problem? The tablet will publish a new tablet record when it comes up.

deepthi:
@GuptaManan100  I believe we have worked around this in the tests for now.
Ideally we do want to merge this change, but based on community feedback let us defer it indefinitely. I'll close the PR but we can keep the branch around for a future where we are able to make this change.

PR #8395, opened 2021-06-29
resolve parser conflicts and add keyword compatibility test case

### LINKED ISSUE #8358: MySQL incompatibility: `BEGIN` and `END` should be `non_reserved_keyword`
For a query like:

```sql
SELECT * FROM test_table WHERE end >= '2001-02-03 04:05:06' limit 1;
```

VTGate as of v10.0.2 requires backticks around `end`, while vanilla MySQL [does not](https://dev.mysql.com/doc/refman/8.0/en/keywords.html).

Specifically:
> `BEGIN` and `END` are keywords but not reserved, so their use as identifiers does not require quoting.

It seems like these should be moved to `non_reserved_keyword` in `sql.y`. I am happy to propose this change but i do not know what knock-on effects this change would have, or where else i would need to do some plumbing to make it work (if anything).

oops-lgtm:
Signed-off-by: longfei <lf.x.0720@gmail.com>

## Description

1. fix conflicts in parser
2. add some Identifier to non_reserved_keyword

## Related Issue(s)

Fixes #8358

## Checklist


GuptaManan100 (changes requested):
Could you also add tests for the changes you made to the parser. We already had the correct parsing of `UNIQUE KEY` because shift/reduce conflicts resolve to shifting. Also I am not sure what replacing `table_id_opt` is accomplishing. The tests for the parser live in `go/vt/schema/parser_test.go`

GuptaManan100:
Yes, you are right, it is `go/vt/sqlparser/parser_test.go`. Could you add tests for this change there? Also, you should run `make parser` to regenerate the `sql.go` file since the test `check_make_parser` is failing

GuptaManan100 (approved):
This looks amazing to me! Also verified that MySQL handles `UNIQUE KEY` exactly as proposed in this PR. Thank you for contributing to Vitess! 🐎  😃

harshit-gangal (approved):

PR #8396, opened 2021-06-29
Ignore SBR statements from pt-table-checksum

rohit-nayak-ps:

## Checklist


shlomi-noach (approved):
In the future, let's extract the `if` logic into a function

PR #8406, opened 2021-06-30
Initial support for stress testing in end-to-end tests

frouioui:
## Description

This pull request starts the implementation of a stress-testing package for the end-to-end tests of Vitess. The package aims to be easily "pluggable" in any existing end-to-end tests and to simulate a load on the test's cluster.

### The `Stresser`

`Stresser` is a new struct that is at the core of this pull request. It is responsible for stressing a cluster which includes: creating fake clients, generating fake data, keeping track of the queries that failed/succeeded, etc. The `Stresser` struct is entirely configurable through the `Config` struct, which has a default (`DefaultConfig`).

A stresser can be created using the `stress.New()` method:

```go
func TestMyEndToEnd(t *testing.T) {
    cfg := stress.DefaultConfig
    cfg.ConnParams = &mysql.ConnParams{Port: 8080, Host: "localhost", DbName: "my_keyspace"}
    s := stress.New(t, cfg)

    ...
}
```

### The load

When a `Stresser` starts its task, it will spawn one or several MySQL clients that continuously query the cluster. The load each client sends to the cluster is composed of: `SELECT`, `INSERT` and `DELETE` statements. These queries are made against tables that are dynamically generated by the Stresser when calling `Stresser.Start()`.

### How to stress?

Below is an example of how to add stressing capabilities to an end-to-end test:

```go
// Assume the cluster is initialized in a TestMain method.

var (
    keyspace = "my_keyspace"    
    vtgatePort = 4242
)

func TestMyEndToEnd(t *testing.T) {

    // Copy the default configuration and alter it with your own settings.
    cfg := stress.DefaultConfig
    cfg.ConnParams = &mysql.ConnParams{Port: vtgatePort, Host: "localhost", DbName: keyspace}
    s := stress.New(t, cfg)

    // Starts the stress. This will start sending a multitude of queries to our cluster.
    s.Start()

    // Pseudo-code to simulate an end-to-end test where we send DML queries to our cluster.
    ApplyDMLToCluster()
    ValidateClusterHealth()

    // Stop the stress test here and asserts that every query was successful, otherwise fail the test.
    s.Stop()
}
```

The example above is a rather simple one and assumes that all queries will be successful. In fact, when we call `s.Stop()`, or its variant `s.StopAfter(time.Duration)`, the `Stresser` will assert that every query was successful.

If an end-to-end test will likely fail queries, for instance, a `reparenting` end-to-end test where we teardown tablets, we can use the `AllowFailure` mode of the `Stresser`. By default `AllowFailure` is set to `false`. When `AllowFailure` is set to `true`, all queries that fail will not be counted in the final assertion when calling `s.Stop()` or `s.StopAfter(time.Duration)`. We can change the value of `AllowFailure` at run-time using `s.AllowFailure(bool)`, enabling more dynamic stress-testing strategies.  

### Reparenting

This pull request adds stress capabilities to some of the reparenting end-to-end tests.

### Summary

This initial support for stress-testing is still simple for now, but is meant to grow larger in the future. All feedback is more than welcome.

## Checklist


go/test/stress/stress.go
```diff
@@ -0,0 +1,404 @@
...
+	} else {
+		s.StopAfter(s.cfg.MinimumDuration - time.Since(s.start))
+	}
+}
+
+// StopAfter stops the Stresser after the given duration. The function will then
+// assert that all the results are successful, and finally prints them to the standard
+// output.
+func (s *Stresser) StopAfter(after time.Duration) {
+	if s.start.Second() == 0 {
```
vmg, line 176: `s.start.IsZero()`

go/test/stress/stress.go
```diff
@@ -0,0 +1,404 @@
...
+// These clients will be responsible for stressing the cluster until Config.MaximumDuration
+// is reached, or until Stresser.Stop() or Stresser.StopAfter() are called.
+//
+// This method returns a pointer to its Stresser to allow chained function call, like:
+//
+// 		s := stress.New(t, cfg).Start()
+//		s.Stop()
+//
+func (s *Stresser) Start() *Stresser {
+	fmt.Println("Starting load testing ...")
```
vmg, line 220: `s.t.Log` here

go/test/stress/stress.go
```diff
@@ -0,0 +1,404 @@
...
+	defer conn.Close()
+
+	var res result
+
+	// Create a timeout based on the Stresser maximum duration and the time
+	// that has already elapsed since the Stresser was started.
+	timeout := time.After(s.cfg.MaximumDuration - time.Since(s.start))
+
+outer:
+	for !s.finish {
```
vmg, line 301: This is not atomic, and the writes to `s.finish` from `StopAfter` are not atomic either. For sanity's sake, I would recommend making `s.finish` an `uint32` and using `atomic.{Load,Store}Uint32` to perform this check.

vmg (approved):
Looks pretty good to me. Just some minor nits.

frouioui:
> Looks pretty good to me. Just some minor nits.

@vmg - Thank you, resolved all the nits via 254d472

shlomi-noach (approved):
This is really awesome and I would have used it

PR #8408, opened 2021-07-02
Replace t.Logf with log.Infof, to avoid confusing red colored logs in CI logs

rohit-nayak-ps:

## Checklist


rohit-nayak-ps:
Incorporating these changes into https://github.com/vitessio/vitess/pull/8404

PR #8414, opened 2021-07-06
v11: deprecation message for a couple Online DDL related commands

shlomi-noach:
## Description
Per https://github.com/vitessio/vitess/issues/6926#issuecomment-817619302, we are deprecating:

- `vtctl OnlineDDL revert ...`
- `vtctl VExec`

Deprecation will spawn two versions (and code removal following that). In v11 we merely log a warning message that these commands will be deprecated in next version.

## Related Issue(s)

## Checklist


askdba (approved):
LGTM

PR #8416, opened 2021-07-06
gomod: do not replace GRPC

vmg:
## Description

I landed https://github.com/grpc/grpc-go/pull/4561 in upstream GRPC last week, which fixes a corner-case bug that was breaking our ETCD client. Since 1.39.0 now exists with the bugfix, we no longer need to pin our version of GRPC with a replacement.

## Checklist


derekperkins (approved):
LGTM

PR #8417, opened 2021-07-06
consolidation: Fix flaky test

vmg:
## Description

This test is flaky and I'm debugging it on CI

## Checklist

### DIFF
--- a/go/vt/vttablet/tabletserver/stream_consolidator_test.go
+++ b/go/vt/vttablet/tabletserver/stream_consolidator_test.go
@@ -53,7 +53,7 @@ type consolidationTest struct {
 	streamItems     []*sqltypes.Result
 
 	leaderCallback func(StreamCallback) error
-	leaderCalls    int
+	leaderCalls    uint64
 
 	results []*consolidationResult
 }
@@ -71,7 +71,7 @@ func generateResultSizes(size, count int) (r []*sqltypes.Result) {
 }
 
 func (ct *consolidationTest) leader(stream StreamCallback) error {
-	ct.leaderCalls++
+	atomic.AddUint64(&ct.leaderCalls, 1)
 	if ct.leaderCallback != nil {
 		return ct.leaderCallback(stream)
 	}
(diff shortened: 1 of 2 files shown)


go/vt/vttablet/tabletserver/stream_consolidator_test.go
```diff
@@ -330,14 +329,14 @@ func TestConsolidatorMemoryLimits(t *testing.T) {
```
dbussink, line 329: Does `streamItemDelay` need to be 10ms above or could that be much lower at 1ms to also have a faster text execution? And the sleep window below could be kept the same, or even lower with more of a time buffer?

dbussink:
```patch
diff --git i/go/vt/vttablet/tabletserver/stream_consolidator_test.go w/go/vt/vttablet/tabletserver/stream_consolidator_test.go
index 97c68aeac3..f3516999a1 100644
--- i/go/vt/vttablet/tabletserver/stream_consolidator_test.go
+++ w/go/vt/vttablet/tabletserver/stream_consolidator_test.go
@@ -323,13 +323,13 @@ func TestConsolidatorMemoryLimits(t *testing.T) {
        t.Run("two-phase consolidation (time)", func(t *testing.T) {
                ct := consolidationTest{
                        cc:              NewStreamConsolidator(128*1024, 2*1024, nocleanup),
-                       streamItemDelay: 10 * time.Millisecond,
+                       streamItemDelay: 1 * time.Millisecond,
                        streamItemCount: 10,
                }
 
                ct.run(10, func(worker int) (string, StreamCallback) {
                        if worker > 4 {
-                               time.Sleep(150 * time.Millisecond)
+                               time.Sleep(50 * time.Millisecond)
                        }
                        return "select 1", func(_ *sqltypes.Result) error {
                                return nil
```

dbussink (approved):
Paired with @vmg on this and the changes make sense and address the CI issue here at hand.

shlomi-noach (approved):

PR #8418, opened 2021-07-06
Small build improvements

dbussink:
## Description

This improves a few bits for the Go build setup. With the recent move to Go 1.16, there's no need anymore to specify GO111MODULE as the default value is now that it is on.

Additionally, it enables `-trimpath` to remove build specific paths from the final release artifacts.

Lastly, this cleans up a bunch of usage of GOPATH. Go itself now falls back to `~/go` if it is not set and we can depend on this default for commands and don't need to set anything manually.

`go env GOPATH` still works and returns this fallback if GOPATH is not set. This further simplifies setup and means people don't need to configure GOPATH anymore separately either.

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -59,7 +59,7 @@ ifndef NOBANNER
 	echo $$(date): Building source tree
 endif
 	bash ./build.env
-	go install $(EXTRA_BUILD_FLAGS) $(VT_GO_PARALLEL) -ldflags "$(shell tools/build_version_flags.sh)" ./go/...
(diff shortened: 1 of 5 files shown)


vmg (approved):

vmg:
Yeeey first @dbussink PR!

shlomi-noach:
@dbussink each PR needs at least one label `Component: *` and one label `Type: *`. For this PR, best match is `Component: build/CI` and `Type: CI/Build`

PR #8420, opened 2021-07-06
VReplication support for non-PRIMARY KEY (Online DDL context): alternate PR

shlomi-noach:
## Description

A take/twist on https://github.com/vitessio/vitess/pull/8364/, per [comment](https://github.com/vitessio/vitess/pull/8364/#issuecomment-874625804)

Mostly exactly the same as #8364 , except:

- `Rule`'s proto has `source_unique_key_columns` instead of `source_unique_key`, and `target_unique_key_columns` instead of `target_unique_key`.
- rowstreamer doesn't need to inspect `information_schema.columns`, since it relies on  `source_unique_key_columns`
- vreplicator likewise doesn't need to inspect `information_schema.columns`, since it relies on `target_unique_key_columns`

## Related Issue(s)

- #6926 
- #8179 

## Checklist


shlomi-noach:
This is now merged into #8364

Issue #7866, opened 2021-04-14
[vtadmin] Provide configuration options to bound the number of inflight tablet/topo RPCs


ajm188:
Please search the existing issues for relevant feature requests, and use the [reaction feature](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to add upvotes to pre-existing requests.

#### Feature Description

From https://github.com/vitessio/vitess/pull/7751#discussion_r602531460

#### Use Case(s)

Anyone running vtadmin against large cluster(s)

PR #8421, opened 2021-07-06
[vtadmin] cluster rpc pools

ajm188:
## Description

This adds an RPCPool specialization on top of the generic pools.ResourcePool, and then adds several read pools (named this way so we don't need to rename (and rename the flags) when we add the r/w pools later). The main reason I'm not using sync2.Semaphore is because the timeout/context semantics don't work for our use case. With the semaphore, you can _either_ use the semaphore's timeout _or_ the context timeout, but what I want is to always use the minimum of the two. RPCPool does exactly that, plus we get all the stats/metrics of ResourcePool "for free".

From there, we add 4 rpc pools:
* topoReadPool - for making general requests that end up reading from the topo, things like GetKeyspace, GetVSchema, etc
* backupReadPool - for making requests that hit the backups endpoints, this is separate in case you need to control request rates to your backing store (s3, gcp, azure, etc)
* schemaReadPool
* workflowReadPool

## Related Issue(s)

- https://github.com/vitessio/vitess/issues/7866

## Checklist

### DIFF
--- a/go/pools/rpc_pool.go
+++ b/go/pools/rpc_pool.go
@@ -1,0 +1,91 @@
+/*
+Copyright 2021 The Vitess Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package pools
+
+import (
+	"context"
+	"time"
+)
+
+// RPCPool is a specialized version of the ResourcePool, for bounding concurrent
+// access to making RPC calls.
+//
+// Whether you use this, or a sync2.Semaphore to gate RPCs (or shared access to
+// any shared resource) depends on what timeout semantics you need. A sync2.Semaphore
+// takes a global timeout, which applies to calls to Acquire(); if you need to
+// respect a context's deadline, you can call AcquireContext(context.Context),
+// but that ignores the global timeout. Conversely, an RPCPool provides only
+// one method of acquisition, Acquire(context.Context), which always uses the
+// lower of the pool-global timeout or the context deadline.
+type RPCPool struct {
+	rp          *ResourcePool
+	waitTimeout time.Duration
+}
+
+// NewRPCPool returns an RPCPool with the given size and wait timeout. A zero
+// timeout will be ignored on calls to Acquire, and only the context deadline
+// will be used. If a logWait function is provided, it will be called whenever
+// a call to Acquire has to wait for a resource, but only after a successful
+// wait (meaning if we hit a timeout before a resource becomes available, it
+// will not be called).
+func NewRPCPool(size int, waitTimeout time.Duration, logWait func(time.Time)) *RPCPool {
+	return &RPCPool{
+		rp:          NewResourcePool(rpcResourceFactory, size, size, 0, size, logWait),
+		waitTimeout: waitTimeout,
+	}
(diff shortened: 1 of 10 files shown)


go/vt/vtadmin/cluster/cluster.go
```diff
@@ -1030,17 +1103,21 @@ func (c *Cluster) GetSrvVSchemas(ctx context.Context, cells []string) ([]*vtadmi
...
 	}
 
-	res, err := c.Vtctld.GetSrvVSchemas(ctx, &vtctldatapb.GetSrvVSchemasRequest{
+	if err := c.topoReadPool.Acquire(ctx); err != nil {
+		return nil, fmt.Errorf("GetSrvVSchema(cluster = %s, cells = %v) failed to acquire topoReadPool: %w", c.ID, cells, err)
+	}
+	resp, err := c.Vtctld.GetSrvVSchemas(ctx, &vtctldatapb.GetSrvVSchemasRequest{
 		Cells: cells,
 	})
+	c.topoReadPool.Release()
```
doeg, line 1112: Non-blocking question -- in some cases, we use `defer c.topoReadPool.Relase()` and in others, like here, we Acquire and the Release sequentially (for lack of a better term). 

In this case, it would be equivalent to `defer`, since it's the only Acquire/Release in the function, right? Is there much of a difference? (You've kinda sold me on the `defer` way since it seems easy (for me) to forget, otherwise, so I'm curious. 😈)

doeg (commented):
Two questions but only one of them is quote-unquote blocking. 🌻

go/pools/rpc_pool.go
```diff
@@ -0,0 +1,91 @@
...
+//
+// It returns nil on successful acquisition, and an error if a timeout occurred
+// before a slot became available.
+//
+// Note: For every successful call to Acquire, the caller must make a
+// corresponding call to Release.
+func (pool *RPCPool) Acquire(ctx context.Context) error {
+	if pool.waitTimeout > 0 {
+		var cancel context.CancelFunc
+		ctx, cancel = context.WithTimeout(ctx, pool.waitTimeout)
```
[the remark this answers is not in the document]
derekperkins: Yes, you can't extend a timeout in a context, only shorten it.

go/vt/vtadmin/cluster/cluster.go
derekperkins: This is a general comment not directed at this specific usage. Until Go 1.14, defer had a significant penalty for hot paths, so you saw a lot of people avoiding it. That has mostly been eliminated, so except for rare circumstances, defer is the safer choice. IMO, it's easier to read, plus eliminates the possibility of adding a return somewhere without handling the unlock.
https://rakyll.org/inlined-defers/

go/pools/rpc_pool.go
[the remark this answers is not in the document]
ajm188: Yeah, it's in the [docs for WithDeadline](https://golang.org/pkg/context/#WithDeadline):

> WithDeadline returns a copy of the parent context with the deadline adjusted to be no later than d. If the parent's deadline is already earlier than d, WithDeadline(parent, d) is semantically equivalent to parent.

go/vt/vtadmin/cluster/cluster.go
ajm188: Yeah, that's exactly the intent I had behind this. If we're just making the rpc, and moving the result from one struct to another (so no other function calls, or iterating over elements in a response, etc), then it's "fine" to just defer the release, but if we're going to do any other "meaningful" work (which i am loosely defining as function calls or iterations) then we should release immediately after finishing the rpc to free up other waiting goroutines.

ajm188: Ah somehow ... github did not show me Derek's comment. Derek's completely correct, but there's another consideration in that we are not _just_ making the rpc call, so it doesn't make sense to be holding onto a slot in the pool for us to do unrelated work.

doeg (approved):

PR #8423, opened 2021-07-07
VReplication: support different keys on source and target tables (different covered columns)

shlomi-noach:
This PR extends #8420 and relies on its functionality

### Background

VReplication works by iterating a source table in key order, and by applying rows and events on target table, based on key order from source and by constructing queries based on target's key.

Historically, VReplication only allowed:
- Iterating via `PRIMARY KEY` on source
- Iterating & applying via `PRIMARY KEY` on target
- both source and target `PRIMARY KEY`s must be identical (same covered columns)

In #8420 we allow source and target tables to use any non-NULLable `UNIQUE KEY`. This is useful for Online DDL migrations where the user may wish to change keys. The idea is to identify a _shared key_, which isn't necessarily the `PRIMARY KEY`, which covers the same set of columns, in same order, in both source and target tables. If such key exists, then the user may, for example, modify the `PRIMARY KEY`. We trick both rowstreamer and vreplicator to believe they're iterating `PRIMARY KEY`, while in fact we give them a list of columns covered by the _shared key_.

### This PR

As of this PR, keys on source and target don't have to cover the same columns. They could cover the exact same set of columns in same order (ideal situation, preferable to other options), or they could cover same columns in different order, or only share subset of columns, or share no columns at all!

There's still a set of restrictions, which are way more relaxed than existing restrictions:

- The source table must have _some_ not-nullable unique key (ideally it will have a `PRIMARY KEY`)
- Likewise, the target table must have _some_ not-nullable unique key (ideally it will have a `PRIMARY KEY`)
- There has to be at least one non-nullable unique key on source table, whose covered set of columns fully exists in target table.
- There has to be at least one non-nullable unique key on target table, whose covered set of columns fully exists in source table.

To illustrate:

```sql
create table this_is_the_source_table (
  uuid varchar(60) not null primary key,
  some_other_unique_identifier bigint not null,
  other columns ...
)

create table this_is_the_target_table (
  uuid varchar(60) not null,
  some_other_unique_identifier bigint not null PRIMARY KEY,
  other columns...
)
```

In the above, it is possible to VReplicate from source table to target table. Each table has some form of non-nullable unique key (in fact, each has a `PRIMARY KEY`). They keys on source and target have no columns shared (source table's unique key covers `uuid`, target tabe's unique key covers `some_other_unique_identifier`). The conditions that make this vreplication still possible is the fact that the _column_ `uuid` exists in target table, and that the _column_ `some_other_unique_identifier` exists in the source table

### What's complicated about this?

Two things:

1. The fact that source and target tables have different keys on different columns, means the tables may be sorted altogether in different order.
2. The way we build & apply DML queries on target (in `table_plan_builder.go`, `replicator_plan.go`) generates `WHERE` conditions that assume a shared key, and also does not anticipate certain conflicts.

### Why does this PR work?

1. Thanks to both some existing logic and to #8420, on target, we iterate the table by the source unique key. That is, vstreamer sends rows to target and identifies which fields comprise the unique key. The target just accepts it as-is, and does not care whether those columns do or do not equal the target's key's columns.
2. We introduce a new field to proto's Rule: `source_unique_key_target_columns`. We educate vreplication about which columns are covered by the unique key on source table (and we accommodate for any possible column renames). `table_plan_builder` then evaluates which of these columns are "extra". On a normal day, both source and target tables will use same exact unique key, and there's no "extra" columns. But e.g. in the above example, `uuid` is marked as an "extra" column.
We use "extra" columns when we construct the `WHERE` clause for DLM events played by `VPlayer`. We add `... AND extra_col=<value>` to the `WHERE` condition for all extra columns. This way the `WHERE` clause covers the columns for both source unique key and target unique key, and thereby row matching is precise.
3. While the mechanism is generic to VReplication, it is _someone's_ responsibility to populate the correct value of `source_unique_key_target_columns` in the vreplication rule. Current implementation is for online DDL's `vrepl.go` coordinator. I tevaluates source and target unique keys, attempts to find a shared key; if impossible, picks best on of the target table, and one from the source table whose covered columns all exist in the target table.

### How to review this PR?
This PR is still Draft, so not strict need to review it. But I'd appreciate it nonetheless!
Because it extends #8420, and because #8420 is not merged at this time, it make sense to review by comparing to `onlineddl-vrepl-source-target-columns` branch, as follows:

https://github.com/planetscale/vitess/compare/onlineddl-vrepl-source-target-columns...planetscale:vrepl-any-uk-source-target?w=1

## Related Issue(s)

- https://github.com/vitessio/vitess/pull/8420
- https://github.com/vitessio/vitess/issues/8179
- #6926 

## Checklist


shlomi-noach:
Added support for yet another interesting scenario: chosen target's unqiue key may contain columns which do not exist in source table...! This is depicted in the following scenario:

```sql
create table onlineddl_test (
  uu varchar(64) not null,
  ts timestamp,
  primary key(uu)
);

alter table onlineddl_test drop primary key, add column id int unsigned auto_increment primary key;
```

in the above there's only one key choice on source table (PK over `uu`) and only one choice for key on target table (PK over `id`). But `id` does not exist in source table. Still, we can make the magic happen!

shlomi-noach:
> chosen target's unqiue key may contain columns which do not exist in source table

That actually not advisable as proven by further test case scenario. When one `UPDATEs` a row on source table in such scenario, it's impracticle (index/performance wise) to identify the row on the target table. It would take a full table scan in best case scenario for any `UPDATE` or `DELETE` because we have no idea what the target's PK should be for associated row. It only works for `INSERT`s. Long term, if we insist, we can perhaps figure something out; ut it would also require us to not use the PK on when generating `WHERE` clause o ntarget table.

Anyway, not for now. 

So for now, the condition is:
- source and target table _may_ use completely different primary keys or otherwise unique keys, which may cover completely disjoint columns.
- The source key must cover columns all of which are present in the target table.
- The target key must cover columns all of which are present in the source table.

shlomi-noach:
Seeing that https://github.com/vitessio/vitess/pull/8364 is merged, this PR is now ready for review

rohit-nayak-ps (approved):
looks good!

The example in the description is incorrect, right? The `id` column is only on the target.

shlomi-noach:
> The example in the description is incorrect, right? The id column is only on the target.

You're right. I began in one trajectory and then had to back down. I'll edit the original comment to clarify.

PR #8427, opened 2021-07-07
[vtadmin] cluster debug pages

ajm188:
## Description

What it says. Some notable callouts:
* clusters, vtctld proxies, and vtgate proxies now take the config they were created from, so they can show this in the debug.
* vtctld and vtgate proxies now track the time their connection was established, as well as the last time that connection was pinged (this is _kind_ of a lie for vtctlds, but that is being fixed separately)
* for credentials, we will **never** show passwords, only an indication of whether there's a password being used (eight asterisks) or not (empty string).

## Checklist


go/vt/vtadmin/debug/debug_test.go
```diff
@@ -0,0 +1,79 @@
...
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestSanitizeString(t *testing.T) {
+	t.Parallel()
+
+	t.Run("empty string", func(t *testing.T) {
```
doeg, line 31: Huh, TIL this `t.Parallel` construction. Very useful!

doeg (approved):

PR #8428, opened 2021-07-07
multishard autocommit should work for bypass routing

demmer:
## Description
Parse and extract the multishard autocommit query directive in the bypass routing plan builder. 

We (Slack) noticed that routing a DML to the keyrange target `[-]`, performance timings implied that the multishard autocommit query directive wasn't being honored when routing the query.

This change makes the query directive take effect when DMLs are force routed to a particular shard or keyrange.

## Checklist
- [x] Verified manually in our dev environment
- [x] Added tests for both the engine and the planner

## Deployment Notes
None


rafael:
This change looks good to me. @demmer  could you look at the conflicts and push again?

Also, extra pair of eyes from @GuptaManan100  or @systay  would be great.

demmer:
I fixed the conflicts and re-pushed. Not sure why all the CI is failing.

systay (approved):
LGTM

systay:
> I fixed the conflicts and re-pushed. Not sure why all the CI is failing.

Guessing that was the "pr-labels" check that was failing. Manan added labels to the PR, so nothing to do 👍

harshit-gangal (approved):

PR #8429, opened 2021-07-08
Gen4 cost calculation tweak

systay:
## Description
When evaluating the cost of plans, gen4 was not comparing route costs correctly. This PR updates how this is done.

## Related Issue(s)
#7280

## Checklist

### DIFF
--- a/go/vt/vtgate/planbuilder/jointree.go
+++ b/go/vt/vtgate/planbuilder/jointree.go
@@ -61,6 +61,13 @@ type (
 		pred  sqlparser.Expr
 	}
 
+	// cost is used to make it easy to compare the cost of two plans with each other
+	cost struct {
+		vindexCost int
+		isUnique   bool
+		opCode     engine.RouteOpcode
+	}
+
 	routePlan struct {
 		routeOpCode engine.RouteOpcode
 		solved      semantics.TableSet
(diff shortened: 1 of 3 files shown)


go/vt/vtgate/planbuilder/jointree.go
```diff
@@ -633,3 +643,48 @@ func (jp *joinPlan) pushOutputColumns(columns []*sqlparser.ColName, semTable *se
...
+	precedence2 := opCodePrecedence[c2.opCode]
+	switch {
+	case precedence1 != precedence2:
+		return precedence1 < precedence2
+	case c1.isUnique == c2.isUnique:
+		return c1.vindexCost <= c2.vindexCost
+	default:
+		return c1.isUnique
+	}
+}
```
frouioui, line 690: I think this whole chunk can be simplified. The `engine.RouteOpcode` is an enumeration and follows the same increment as `opCodePrecedence`. The plan tests were still passing after using directly `c1.opCode` and `c2.opCode`, instead of `precedence1` and `precedence2`.


```suggestion
// less compares two costs and returns true if the first cost is cheaper than the second
func less(c1, c2 cost) bool {
	switch {
	case c1.opCode != c2.opCode:
		return c1.opCode < c2.opCode
	case c1.isUnique == c2.isUnique:
		return c1.vindexCost <= c2.vindexCost
	default:
		return c1.isUnique
	}
}
```

systay: hmm... that is not a bad idea. :+1:

frouioui (approved):

PR #8433, opened 2021-07-08
Adds flag to vttablet to disallow online DDL statements

GuptaManan100:
## Description
Adds flag to vttablet to disallow online DDL statements. It is used in vtexplain.

## Checklist

### DIFF
--- a/go/vt/vtexplain/vtexplain_vttablet.go
+++ b/go/vt/vtexplain/vtexplain_vttablet.go
@@ -103,6 +103,7 @@ func newTablet(opts *Options, t *topodatapb.Tablet) *explainTablet {
 		config.TwoPCAbandonAge = 1.0
 		config.TwoPCEnable = true
 	}
+	config.EnableOnlineDDL = false
 
 	// XXX much of this is cloned from the tabletserver tests
 	tsv := tabletserver.NewTabletServer(topoproto.TabletAliasString(t.Alias), config, memorytopo.NewServer(""), t.Alias)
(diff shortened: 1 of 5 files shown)


deepthi:
What is the problem this PR is trying to fix? Is there an issue that illustrates the bug/gap/feature request?

harshit-gangal:
It is trying to fix the vtexplain flaky tests as it sometimes shows `sql` queries from online ddl which are not expected otherwise.

harshit-gangal:
one of the failed run: https://pipelines.actions.githubusercontent.com/TBP6MbgxaNRkq1k6DamNZYdWLId6AliZAF5kl9JXndalqVBili/_apis/pipelines/1/runs/281272/signedlogcontent/3?urlExpires=2021-07-08T16%3A29%3A59.2289423Z&urlSigningMethod=HMACV1&urlSignature=%2ByrMEDriJG9YAfBqpqsqCABcQwFjdqWujuyuRRcbUO0%3D

shlomi-noach:
This circles back to how we test. Instead of testing a state (read data from tables, or read table definition, see that it matches expectation), we read in-transit events. 

This method slaps us again and again in the face. Now it's online DDL. Previous time it was a changed query in vreplication's WithDDL. In a WIP I have it's the change of sequence of queries in rowstreamer. Tomorrow it will be tableGC or any other new query we may introduce. 

What I'm suggesting is that our testing method is brittle and we will find ourselves turning off more and more modules just so some unrelated test passes. 

I don't mean to block the progress on this PR and no strings attached regarding Online DDL 😁, and for now this may be the fastest way to solve the issue. 

But in the greater scope we need to rewrite our tests in a different way. At the time I told Rohit I'd look into the same, for vreplication related tests. I looked and blacked out. It's very complicated. I'm not sure where to begin.

My 2c

harshit-gangal (approved):

harshit-gangal:
We can change the testing when we re-write vtexplain, for now, this is the fastest way forward.

PR #8439, opened 2021-07-08
Add "no-scatter" flag to prohibit the use of scatter queries

systay:
## Description
Adds a flag, `no-scatter` that makes the planner fail if scatter queries are produced by the planner.

## TODO
- [x] Add query hint that allows the use of scatter queries even when the `no-scatter` flag is on.

## Checklist


GuptaManan100 (approved):
LGTM! 🏎️

frouioui (approved):
That looks good to me!

harshit-gangal (approved):

PR #8440, opened 2021-07-09
Merge 9.0 into 10.0

systay:
## Description
Merge 9.0 into 10.0


shlomi-noach:
@systay looking at why `onlineddl_revert` and `onlineddl_declarative` fail, I notice neither of these two existed in `release-9.0`. They do exist in `release-10.0`.

Since I can't see anything specific to this PR that's related to online DDL and why it might affect the two tests, I tend to believe the reason the tests are failing is due to not existing in 9.0. Does that make sense? Or since this ports from 9.0 to 10.0 they should pass? I'm really not sure.

PR #8445, opened 2021-07-10
VStream API: handle reparenting and unhealthy tablets

rohit-nayak-ps:
## Description

Currently the VStream API can specify a tablet type (MASTER or REPLICA) to stream from. In cases where the tablet stops serving (crash/manually shut down) and there are no more tablets of that type then it errors out and the client has to retry. While this is an inconvenience it is not a huge issue since, in any case, users have to handle network errors and maintain state to resume the streaming. 

However there are a couple of issues not handled correctly:
* when there is a reparent (say we are streaming from the primary and it is now deemed to be a replica, or vice-versa), but the tablet itself is  healthy and its type has changed.
* when there is a large underlying replication lag and the tablet is unhealthy but still serving

In such cases we continue to stream events from the current source tablet, which is incorrect. 

This PR changes the functionality to poll for tablet health and change in state and switches over to a different valid tablet.

### Changes

Instead of streaming using vtgate's ResolveDestination where it is not possible to identify the tablet being used, we use the tablet picker to select the tablet explicitly. The picker is already used in vreplication workflows. Streaming from a specific tablet let's us also subscribe to a tablet's health check to identify when that tablet has become unhealthy or its tablet type has changed. We can then switch to a healthy tablet matching the tablet type specified by the client.

### Test related changes

* The new functionality needs a new e2e test since we need different components to be available
* Changed RetryMax to 1 for a vrep test (default is 3). It is usually pointless to retry a vrep test thrice, since errors occur if there is a bug, flakiness or an issue with the CI's VM. In which case user will be notified earlier and if it thought to be flaky we can start fresh faster.
* The mock topo functionality has been enhanced for the unit tests to work with the new approach

## Checklist

### DIFF
--- a/go/vt/vtgate/vstream_manager.go
+++ b/go/vt/vtgate/vstream_manager.go
@@ -23,11 +23,14 @@ import (
 	"sync"
 	"time"
 
+	"vitess.io/vitess/go/vt/discovery"
+	querypb "vitess.io/vitess/go/vt/proto/query"
+	"vitess.io/vitess/go/vt/topo"
+
 	vtgatepb "vitess.io/vitess/go/vt/proto/vtgate"
 
 	"google.golang.org/protobuf/proto"
 
-	"vitess.io/vitess/go/vt/key"
 	"vitess.io/vitess/go/vt/log"
 	binlogdatapb "vitess.io/vitess/go/vt/proto/binlogdata"
 	topodatapb "vitess.io/vitess/go/vt/proto/topodata"
@@ -92,6 +95,7 @@ type vstream struct {
 
 	eventCh           chan []*binlogdatapb.VEvent
 	heartbeatInterval uint32
+	ts                *topo.Server
 }
 
 type journalEvent struct {
@@ -114,6 +118,14 @@ func (vsm *vstreamManager) VStream(ctx context.Context, tabletType topodatapb.Ta
 	if err != nil {
 		return err
 	}
+	ts, err := vsm.toposerv.GetTopoServer()
+	if err != nil {
+		return err
+	}
+	if ts == nil {
+		log.Errorf("unable to get topo server in VStream()")
+		return fmt.Errorf("unable to get topo server")
+	}
 	vs := &vstream{
 		vgtid:              vgtid,
 		tabletType:         tabletType,
@@ -127,6 +139,7 @@ func (vsm *vstreamManager) VStream(ctx context.Context, tabletType topodatapb.Ta
 		vsm:                vsm,
 		eventCh:            make(chan []*binlogdatapb.VEvent),
 		heartbeatInterval:  flags.GetHeartbeatInterval(),
+		ts:                 ts,
 	}
 	return vs.stream(ctx)
 }
@@ -286,6 +299,7 @@ func (vs *vstream) startOneStream(ctx context.Context, sgtid *binlogdatapb.Shard
 
 		// Set the error on exit. First one wins.
 		if err != nil {
+			log.Errorf("Error in vstream for %+v: %s", sgtid, err)
 			vs.once.Do(func() {
 				vs.err = err
 				vs.cancel()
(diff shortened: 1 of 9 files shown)


rafael:
I took an initial pass at this PR and the overall approach makes sense. I will take one more pass tomorrow looking at the details.

go/vt/vtgate/vstream_manager.go
```diff
@@ -404,25 +418,66 @@ func (vs *vstream) streamFromTablet(ctx context.Context, sgtid *binlogdatapb.Sha
...
+			tabletConn.StreamHealth(ctx, func(shr *querypb.StreamHealthResponse) error {
+				var errString string
+				if ctx.Err() != nil {
+					return fmt.Errorf("context has ended")
+				}
+				if shr == nil || shr.RealtimeStats == nil || shr.Target == nil {
+					return fmt.Errorf("health check failed")
+				}
+				if vs.tabletType != shr.Target.TabletType {
+					errString = fmt.Sprintf("tablet %s is no longer healthy: %s, restarting vstream",
```
rafael, line 455: Shouldn't the error string here be the one below?
```
errString = fmt.Sprintf("tablet type has changed from %s to %s, restarting vstream",
```

go/vt/vtgate/vstream_manager.go
```diff
@@ -404,25 +418,66 @@ func (vs *vstream) streamFromTablet(ctx context.Context, sgtid *binlogdatapb.Sha
...
+						tablet.Alias, shr.RealtimeStats.HealthError)
+				} else if shr.RealtimeStats.HealthError != "" {
+					errString = fmt.Sprintf("tablet type has changed from %s to %s, restarting vstream",
```
rafael, line 458: Same as above, it seems that the error strings are mixed:
```
errString = fmt.Sprintf("tablet %s is no longer healthy: %s, restarting vstream",
```

rafael (approved):
Rohit - second pass from my end. Sorry it took longer. This looks great. I think there is only a small nit the error string that is returned. Left comments around that. Let me know your thoughts. I'm going to merge this as is, so it can make it to 11.0. The error string can be fixed in a different PR. 

One thing that I was reflecting is that it seems that even in this case there where will be a small window where we will stream events from potential incorrect tablet types. The window between the type changing and vtgate getting that information via the healthcheck. I don't think we can do anything about that, but curious if this something we should be very explicit in the documentation, so clients can reason about it.

go/vt/vtgate/vstream_manager.go
```diff
@@ -404,25 +418,66 @@ func (vs *vstream) streamFromTablet(ctx context.Context, sgtid *binlogdatapb.Sha
...
+		}
+		tabletConn, err := vs.vsm.resolver.GetGateway().QueryServiceByAlias(tablet.Alias, target)
+		if err != nil {
+			log.Errorf(err.Error())
+			return err
+		}
+
+		errCh := make(chan string, 1)
+		go func() {
+			tabletConn.StreamHealth(ctx, func(shr *querypb.StreamHealthResponse) error {
```
sonne5, line 446: Just an idea: this seems something that is useful for all stream use cases. Would it make sense to move this logic inside `PickForStreaming` and change the interface of it to
```
PickForStreaming(ctx context.Context) (*topodatapb.Tablet, chan string, error)
```

go/vt/vtgate/vstream_manager.go
```diff
@@ -404,25 +418,66 @@ func (vs *vstream) streamFromTablet(ctx context.Context, sgtid *binlogdatapb.Sha
...
+				var errString string
+				if ctx.Err() != nil {
+					return fmt.Errorf("context has ended")
+				}
+				if shr == nil || shr.RealtimeStats == nil || shr.Target == nil {
+					return fmt.Errorf("health check failed")
+				}
```
sonne5, line 453: Should we also pass the error message to `errCh` here so that vsm can pick a new tablet in these two cases

go/vt/vtgate/vstream_manager.go
```diff
@@ -404,25 +418,66 @@ func (vs *vstream) streamFromTablet(ctx context.Context, sgtid *binlogdatapb.Sha
...
+			TabletType: vs.tabletType,
+			Cell:       vs.vsm.cell,
+		}
+		tabletConn, err := vs.vsm.resolver.GetGateway().QueryServiceByAlias(tablet.Alias, target)
+		if err != nil {
+			log.Errorf(err.Error())
+			return err
+		}
+
+		errCh := make(chan string, 1)
```
sonne5, line 444: nit: we could make this `chan error` then we don't need to do the conversion between string and error

sonne5 (commented):
Thanks for working on this!

go/vt/vtgate/vstream_manager.go
rohit-nayak-ps: > nit: we could make this `chan error` then we don't need to do the conversion between string and error

yes. I am touching this file for a PR this/next week and will make this change.

go/vt/vtgate/vstream_manager.go
rohit-nayak-ps: yes, will do

go/vt/vtgate/vstream_manager.go
rohit-nayak-ps: ouch, will fix this

go/vt/vtgate/vstream_manager.go
rohit-nayak-ps: I get your point: for example vreplication streams should also track tablet changes. 

However the tablet picker is currently a passive module. Making it track the health checks of the tablets will add more complexity and break the usage pattern in other modules. I will keep this in mind when we update vreplication to use this logic and see if this can be wrapped meaningfully in the picker or elsewhere.

PR #8450, opened 2021-07-11
OnlineDDL: -skip-topo is always 'true'

shlomi-noach:
## Description

Per https://github.com/vitessio/vitess/issues/6926#issuecomment-817619302:

`release-11.0` is now frozen. Moving forward, we want `skip-topo` to be always `true`. The user may choose to use `set @@ddl_strategy='gh-ost -skip-topo'` or not use it, e.g `set @@ddl_strategy='gh-ost'` and the result is the same: we will skip topo. We will merge that in `v12`. We intend to eventually remove the logic behind topo-based schema migration requests. But that will only be done in `v13`.

## Related Issue(s)

## Checklist


deepthi (approved):
LGTM

PR #8382, opened 2021-06-24
Gen4 fail more2

systay:
## Description
This PR will make sure that the gen4 planner does one of the following:
- Produces a correct query (correct here means "does what the user wants", not necessarily the same thing that V3 did)
- Produces a correct error (if the query is wrong, it should be detected and failed correctly)
- Produce a Gen4NotSupported error. This makes it easy to fall back on V3 when gen4 does not yet support the query.

## Related Issue(s)
#7280
Continuation of #8359

## Checklist
- [x] Tests were added or are not required
- [x] Documentation was added or is not required

PR #8451, opened 2021-07-12
Gen4: fail more unsupported queries

frouioui:
## Description

This pull request is the continuity of https://github.com/vitessio/vitess/pull/8382. The goal is to support and/or fail more queries in gen4. This pull request ensures we error out queries that are not supposed to be supported, and return a `Gen4NotSupported` error for those that are not yet handled by the gen4 planner.

## Checklist


go/vt/vtgate/planbuilder/jointree_transformers.go
```diff
@@ -112,8 +112,9 @@ func transformRoutePlan(n *routePlan) (*route, error) {
 			LeftExpr:  lft,
 		}
 		tablesForSelect = sqlparser.TableExprs{joinExpr}
-		// todo: add table
-		// tableNameMap[sqlparser.String()] = nil
+		for _, tblNames := range leftJoin.right.tableNames() {
+			tableNameMap[tblNames] = nil
+		}
```
frouioui, line 117: Awesome 👏

frouioui (commented):
Looking good to me

systay (approved):

PR #8456, opened 2021-07-12
Remove vtrpc.LegacyErrorCode

ajm188:
It has been "legacy" for 4 years now, so it should be safe to just drop.

## Checklist
- [x] Tests were added or are not required - n/a
- [x] Documentation was added or is not required - n/a

I wasn't sure what component to pick since it pretty much affects all of them (since they all use vtrpc.RPCError in some form or fashion)


vmg (approved):
The changes look correct to me. You're gonna need to `make proto` again because I merged an updated version of the compiler this morning.

Besides that, I don't know whether the `LegacyErrorCode` is legacy enough to be removed already!

ajm188:
Hmm, it looks like this is failing on the java stuff, do you have any idea how I update that to stop expecting these types?

```
Error:  Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.1:compile (default-compile) on project vitess-client: Compilation failure: Compilation failure: 
Error:  /vt/src/vitess.io/vitess/java/client/src/main/java/io/vitess/client/Proto.java:[96,20] cannot find symbol
Error:    symbol:   method getLegacyCode()
Error:    location: variable error of type io.vitess.proto.Vtrpc.RPCError
... (more like this)
```

vmg:
@systay maybe you know how this Java stuff is generated?

systay:
hmm... if it's not created with `make proto`, I don't know

harshit-gangal:
The only thing to change is removing the switch condition in `java/client/src/main/java/io/vitess/client/Proto.java`
which uses the LegacyErrorCode.

Issue #4254, opened 2018-10-09
FR: Improve TLS monitoring and maybe configuration


sjmudd:
#### Feature Description

Default metrics which are shown by `vtgate` or `vttablet` do not indicate much about TLS usage. The default settings do not use TLS so this leaves you with a setup which is by default easier to configure but somewhat less secure.

If using TLS it would be convenient to be able to view some TLS metrics for each process, at least on the receiving end for the active connections which version of TLS is being used. 

For vtgate that would be
* grpc port incoming connection counts for each TLS version (including TLS not being used)
* mysql port incoming connection counts for each TLS version (including TLS not being used)

For vttablet:
* grpc port incoming connection counts for each TLS version (including TLS not being used)

I guess the same should be visible on other processes.

#### Use Case(s)

This facilitates migration from non-TLS to TLS settings using grpc and the MySQL protocols.
It also would allow the administrator to validate if TLS is being used and which version is used for incoming and outgoing connections.  Given TLS security requirements there may be a need to have a minimum TLS version and to verify that this is actually being used.

I haven't checked the Vitess configuration but it would be convenient to be able to:
* optionally require TLS usage
* be able to select / reject / choose TLS versions

I would guess this is already available for grpc usage. Something similar if not available for the MySQL port access would be good too (if not already there).   Similarly outgoing grpc configuration should have similar options.

I am not sure exactly what is available at the moment but including documentation on what's required to migrate a running Vitess cluster to use TLS first optionally and then for TLS usage would be good.

dbussink:
With https://github.com/vitessio/vitess/pull/8460 it's possible to configure the minimally required TLS version.

PR #8460, opened 2021-07-13
Allow for configuration of the minimal TLS version

dbussink:
TLS 1.0 & TLS 1.1 are deprecated and shouldn't be used anymore. There are however many older MySQL versions out there where the latest 5.6, 5.7 or 8.0 patch release isn't used which means they don't have a build against OpenSSL with latest TLS support.

This means we can't easily change the minimum version to always be TLS 1.2, but the best possible option is to create flag instead.

The changes here add support for that flag. The default still is TLS 1.2 as the minimum version, but people who run against an older MySQL can use a new flag to override this and still allow TLS 1.0 or TLS 1.1 if desired.

## Related Issue(s)

Related to #4254 since this implements a minimal TLS version configuration option. The option is only added / used for the MySQL Vitess connects to and the LDAP server used.

Internally for Vitess and the GRPC layer, TLS 1.2 is set as flag. I don't think this changes anything in practice, since afaik those services already would have TLS 1.2 as the minimum themselves anyway. 

## Checklist

Not entire sure on tests here, if new enough MySQL versions are used everywhere nothing would break in testing. I'm going to let CI take a look at that. If older versions are in the text matrix, we can use the new option to let those still pass. 

As for documentation, the new option does need to be documented but that's in a separate repository. 

## Deployment Notes

This changes the default minimum to TLS 1.2. I'd consider that a potentially breaking change for people so we'd need to make sure it's called out in the release notes that folks using an older MySQL version might need the option to lower the minimal TLS version.

### DIFF
--- a/go/vt/vttls/vttls.go
+++ b/go/vt/vttls/vttls.go
@@ -30,28 +30,48 @@ import (
 // Updated list of acceptable cipher suits to address
 // Fixed upstream in https://github.com/golang/go/issues/13385
 // This removed CBC mode ciphers that are suseptiable to Lucky13 style attacks
-func newTLSConfig() *tls.Config {
-	return &tls.Config{
-		// MySQL Community edition has some problems with TLS1.2
-		// TODO: Validate this will not break servers using mysql community edition < 5.7.10
-		// MinVersion: tls.VersionTLS12,
+func newTLSConfig(minVersion uint16) *tls.Config {
 
-		// Default ordering taken from
-		// go 1.11 crypto/tls/cipher_suites.go
-		CipherSuites: []uint16{
-			tls.TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305,
-			tls.TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305,
-			tls.TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,
-			tls.TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,
-			tls.TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,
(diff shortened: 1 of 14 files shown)


go/vt/vttls/vttls.go
```diff
@@ -30,37 +30,57 @@ import (
...
+	case "TLSv1.3":
+		return tls.VersionTLS13
+	case "", "TLSv1.2":
+		return tls.VersionTLS12
+	case "TLSv1.1":
+		return tls.VersionTLS11
+	case "TLSv1.0":
+		return tls.VersionTLS10
+	default:
+		return tls.VersionTLS12
```
vmg, line 74: Maybe a more sensible option here would be to return an error when we don't recognize a TLS string, instead of defaulting to 1.2. It's also better for future proofing!

go/vt/vttls/vttls.go
```diff
@@ -30,37 +30,57 @@ import (
...
+	return &tls.Config{
+		MinVersion:   minVersion,
+		CipherSuites: ciphers,
+	}
+}
+
+// TLSVersionToNumber converts a text description of the TLS protocol
+// to the internal Go number representation.
+func TLSVersionToNumber(tlsVersion string) uint16 {
+	switch tlsVersion {
```
vmg, line 64: If you match on `strings.ToLower` that's good enough to make it case insensitive.

go/vt/vttls/vttls.go
dbussink: Done!

go/vt/vttls/vttls.go
dbussink: Also done!

deepthi (approved):
LGTM

PR #8461, opened 2021-07-13
[vtadmin] Add Options struct to wrap grpc/http options

ajm188:
For some other branches, the API is going to need more options that
don't fit neatly into either the grpcserver.Options or http.Options
structs, so this will give us a more general scope to put them in.

I'm getting this refactor in now because the diff is noisy, and I don't
want it in the way of those other changes.

## Checklist


doeg (approved):
i like dis

PR #8463, opened 2021-07-13
Remove lingering mysqld sock files on vttestserver container startup

ilikeorangutans:
## Description

When restarting vttestserver after an ungraceful shutdown mysql doesn't start up due to the lingering sock files. Took the snippet from `vitess/examples/compose/vttablet-up.sh` and added it to `vttestserver`'s `run.sh`: https://github.com/vitessio/vitess/blob/345d30a061c2d30a88ab986728bbdda44230b9ac/examples/compose/vttablet-up.sh#L90-L94 

## Related Issue(s)

- https://github.com/vitessio/vitess/pull/5115 

## Checklist

### DIFF
--- a/docker/vttestserver/run.sh
+++ b/docker/vttestserver/run.sh
@@ -24,6 +24,11 @@ if [[ -z $MYSQL_MAX_CONNECTIONS ]]; then
 fi
 echo "max_connections = $MYSQL_MAX_CONNECTIONS" >> /vt/config/mycnf/default-fast.cnf
 
+# Delete socket files before running mysqlctld if exists.
+# This is the primary reason for unhealthy state on restart.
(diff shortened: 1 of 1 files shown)


dkhenry (approved):
LGTM

mcronce (approved):
I think this looks good, the only thing I'd change would be to not `rm` with `-r`, since the things we expect to delete aren't directories

ilikeorangutans:
> I think this looks good, the only thing I'd change would be to not `rm` with `-r`, since the things we expect to delete aren't directories

Good call. I'll update. 

I'm surprised by all the failing tests. Could there be tests that rely on the sock file to remain?

GuptaManan100 (approved):
LGTM

Issue #7114, opened 2020-12-04
Task: Replace master with primary (except TabletType)


deepthi:
All usages of master in variable names / function parameters can be replaced.
vtctl commands will need to be deprecated and equivalent new ones provided. e.g. `SetShardIsMasterServing`
vtctl command arguments will need to go through a deprecation cycle. For one release, we accept both old and new arguments and then remove the old ones. e.g. `-include_master`
RPCs will need to be deprecated and new ones created. e.g. `DemoteMaster`

There is a separate task for TabletType #7113 

Update website docs at the same time to document new commands and flags and mark deprecated ones.

derekperkins:
Cross posting from #7113 https://github.com/vitessio/vitess/issues/7113#issuecomment-738495107

While I personally prefer primary, I'd vote to match MySQL and call it source
https://mysqlhighavailability.com/mysql-terminology-updates/
https://dev.mysql.com/doc/refman/8.0/en/replication.html

deepthi:
x-post: https://github.com/vitessio/vitess/issues/7109#issuecomment-738498969

deepthi:
- [x] mysqld functions and structs #8251
- [x] tabletmanager interface and implementations #8251
- [ ] tabletmanager RPCs
- [ ] ReplicationStatus Protobuf

After the next release
- [ ] Change wrangler code to use new RPCs
- [ ] Deprecate old RPCs

After one more release
- [ ] Delete deprecated RPCs

deepthi:
I have opened separate issues for the outstanding work (which can only happen one release at a time in order to maintain compatibility across releases). Closing this because the rest of the work is complete.
Ref: #8837 #8838 and #8839

PR #8466, opened 2021-07-13
naming: rename proto fields that refer to master

deepthi:
## Description
Rename fields in ReplicationStatus message.
Add PrimaryStatus message (unused right now), eventually to replace MasterStatus message.

## Related Issue(s)

#7114 

## Checklist


deepthi:
@vmg I think this doesn't violate wire compatibility but your :eyes: will be appreciated.
@rohit-nayak-ps this might break vtctld UI, because the json returned by the APIs is now different. Instead of `seconds_behind_master`, the json now contains `replication_lag_seconds`. What do you think is the best way to move forward?
@ajm188 @doeg not sure how extensive the use of SBM is in vtadmin code but it will be good to fix them now versus later.

vmg (approved):
Always safe to rename fields in Proto, they don't appear on the wire. :+1:

rohit-nayak-ps:
> @rohit-nayak-ps this might break vtctld UI, because the json returned by the APIs is now different. In fact api_test now fails but I haven't changed it to expect the new json. Instead of `seconds_behind_master`, the json now contains `replication_lag_seconds`. What do you think is the best way to move forward?

That field is rendered by vttablet itself (the debug/status endpoint) and not the vtctld web UI. I fixed that test and the rendering.

PR #8467, opened 2021-07-14
Add `rows` as keyword in sqlparser

bnu0:
## Description

Rows is a reserved word as of [MariaDB 10.2.4](https://mariadb.com/kb/en/mariadb-1024-release-notes/) and [MySQL 8.0.2](https://dev.mysql.com/doc/mysqld-version-reference/en/keywords-8-0.html). If a table has a column named `rows` (an unfortunate legacy choice) then, for example, VReplication will fail to copy data from the table with:

```
Error: vttablet: rpc error: code = Unknown desc = You have an error in your SQL syntax; check the 
manual that corresponds to your MariaDB server version for the right syntax to use near 'rows, othercol1,
othercol2' at line 1 (errno 1064) (sqlstate 42000) during query: select id, rows, othercol1, othercol2 
from table_xyz order by id
```

## Notes

I am not very clear on the responsibility of these keywords, listed in `sqlparser/keywords.go` vs. the `[non_]reserved_keywords` in `sqlparser/sql.y`. If i made the change in the wrong place, or need to do it in both, please let me know. I was inspred by https://github.com/vitessio/vitess/pull/8086.

## Checklist

### DIFF
--- a/go/vt/sqlparser/keywords.go
+++ b/go/vt/sqlparser/keywords.go
@@ -401,6 +401,7 @@ var keywords = []keyword{
 	{"right", RIGHT},
 	{"rlike", REGEXP},
 	{"rollback", ROLLBACK},
+	{"row", UNUSED},
 	{"row_format", ROW_FORMAT},
 	{"row_number", UNUSED},
 	{"rows", UNUSED},
--- a/go/vt/sqlparser/keywords.go
+++ b/go/vt/sqlparser/keywords.go
@@ -403,6 +403,7 @@ var keywords = []keyword{
 	{"rollback", ROLLBACK},
 	{"row_format", ROW_FORMAT},
 	{"row_number", UNUSED},
+	{"rows", UNUSED},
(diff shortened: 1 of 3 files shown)


GuptaManan100 (approved):
This is great! Thankyou for contributing to Vitess! 💕

GuptaManan100:
There is a failing unit test in `plan_test.go`. The test is in `show_cases.txt, Line: 71`. It is using the `rows` keyword in the query so the output needs to be updated. @bnu0 could you make that change as well?

harshit-gangal:
As you are adding `rows`, can you also add `row` ?

bnu0:
@harshit-gangal: `row` added in  28ed759.
@GuptaManan100: Test fixed in 79330bf.

harshit-gangal (approved):

GuptaManan100 (approved):
Thankyou for the changes! 🥇

PR #8471, opened 2021-07-14
Automatically use the latest tag of Vitess for cluster upgrade E2E test

frouioui:
## Description

This pull request changes the `cluster_endtoend_upgrade` workflow (`Cluster (upgrade)`) by dynamically choosing the Vitess tag on which to downgrade/upgrade the cluster.

## Checklist

### DIFF
--- a/.github/workflows/cluster_endtoend_upgrade.yml
+++ b/.github/workflows/cluster_endtoend_upgrade.yml
@@ -20,7 +20,7 @@ jobs:
 
     - name: Get latest Vitess tag
       run: |
-        echo "latest_release=$(git describe --tags $(git rev-list --tags --max-count=1))" >> $GITHUB_ENV
+        echo "latest_release=$(git show-ref --tags | grep -e 'refs/tags/v[0-9]' | sed 's/[a-z0-9]* refs\/tags\/v//' | sort -nr | head -n1)" >> $GITHUB_ENV
 
     - name: Tune the OS
       run: |
--- a/.github/workflows/cluster_endtoend_upgrade.yml
+++ b/.github/workflows/cluster_endtoend_upgrade.yml
@@ -15,7 +15,7 @@ jobs:
 
     - name: Get latest Vitess tag
       run: |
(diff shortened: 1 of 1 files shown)


deepthi:
The idea is good but the implementation is incorrect.
We need to test against the most recent "released" version which right now means 10.0.
Once 11.0 is released we want to be testing against that.

frouioui:
@deepthi - I changed the script via 006ff45 to only include GA releases, thus removing RCs. The workflow is now running against the latest GA release (`10.0.2` as of today)

deepthi:
> @deepthi - I changed the script via [006ff45](https://github.com/vitessio/vitess/commit/006ff452f8708e1c5cc7f45b328735df2f64b25e) to only include GA releases, thus excluding RCs. The workflow is now running against the latest GA release (`10.0.2` as of today)

What do people think about pinning this to a minor version like this rather than requiring it to be the most recent major version? Any opinions?
cc @rafael

frouioui:
@deepthi - We can have a matrix in the workflow that uses the latest minor and major versions of Vitess. And if we want to test against each supported major version of Vitess (the last 3-4 releases), the matrix can do it too

rafael:
>@deepthi - We can have a matrix in the workflow that uses the latest minor and major versions of Vitess. And if we want to test against each supported major version of Vitess (the last 3-4 releases), the matrix can do it too

This approach makes sense to me!   Love the idea of also running the tests against the major version of supported releases.

frouioui:
Hello @deepthi and @rafael, hope you're well. I have made the workflow more dynamic and extended its coverage. After these changes the workflow runs on:
- The last three major releases of Vitess (this arbitrary value can be changed)
- The last minor release of the current major release of Vitess (right now it will be `10.0.2`, but once we release `11.0.0` it will be none since there is no minor release)

deepthi:
This is nice!
Technically we only support upgrade from the last major release, so we don't need to do three releases.
I'm ok with a follow up PR to fix that or we can wait until upgrade from the oldest release breaks and fix it then.

deepthi (approved):

PR #8472, opened 2021-07-14
Handle lock release with SIGHUP in VTGR

sonne5:

## Description

VTGR relies on the lock from topo server. Most of the topo lock uses health-check of session to determine if a process is still holding the lock. This is not ideal for lock release process during a restart with SIGHUP. Take consul as an example, the session check by default uses serfHealth, which is a health check on the node level - even if the process restart because the node is healthy, consul will think the lock is still being held until the TTL. As a result, during a deploy, there could be TTL period of time that VTGR cannot grab the lock and fix the cluster.

This PR handle the SIGHUP signal by explicitly release the lock that was held to avoid the situation above.

## Related Issue(s)

https://github.com/vitessio/vitess/issues/8386

## Checklist

### DIFF
--- a/go/vt/vtgr/controller/refresh.go
+++ b/go/vt/vtgr/controller/refresh.go
@@ -63,9 +63,6 @@ type GRTmcClient interface {
 	Ping(ctx context.Context, tablet *topodatapb.Tablet) error
 }
 
-// HostnameGetter is used to get local hostname
-type HostnameGetter func() (string, error)
-
 // GRShard stores the information about a Vitess shard that's running MySQL GR
 type GRShard struct {
 	KeyspaceShard        *topo.KeyspaceShard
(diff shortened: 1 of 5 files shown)


deepthi (commented):
Code looks fine, but it seems to be handling SIGHUP, not SIGTERM.
Which one should it be? or both?

narcsfz:
> Code looks fine, but it seems to be handling SIGHUP, not SIGTERM.
> Which one should it be? or both?

It depends on what signal ppl use for restart, I think we can start with `SIGHUP` and call it out in the document. Later add supports for `SIGTERM` if required. wdyt?

deepthi:
> > Code looks fine, but it seems to be handling SIGHUP, not SIGTERM.
> > Which one should it be? or both?
> 
> It depends on what signal ppl use for restart, I think we can start with `SIGHUP` and call it out in the document. Later add supports for `SIGTERM` if required. wdyt?

For safety it is better to handle `SIGTERM` as well. I don't have a preference on when this should be added, I will leave that up to you.
LMK if that is planned to be a separate PR so that I can approve and merge this one.

sonne5:
Thanks @deepthi, I will try to address SIGTERM separately (hopefully soon :-))

sonne5:
Gentle nudge

deepthi (approved):

PR #8474, opened 2021-07-14
CI: Move cluster and unit tests to docker to see if they run more reliably than in the native CI VM

rohit-nayak-ps:

## Checklist

### DIFF
--- a/.github/workflows/cluster_endtoend_11.yml
+++ b/.github/workflows/cluster_endtoend_11.yml
@@ -47,4 +47,7 @@ jobs:
       timeout-minutes: 30
       run: |
         source build.env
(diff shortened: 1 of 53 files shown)


deepthi:
Failures:
unit mysql80: flaky
unit percona56: I pushed a docker tag for this (same as percona).
unit mariadb102: we aren't currently building a bootstrap image for mariadb102. We might need to start doing that.

For the rest, it seems like a couple of them timed out. It does take longer to run with docker.

rohit-nayak-ps:
Still finding flaky tests using docker and in addition there are a few tests that fail because take longer than 10 minutes in docker. So not a viable option.

PR #8480, opened 2021-07-15
Gen4: extracting routing information from predicates for system tables

harshit-gangal:
## Description

Add support for information_schema queries in Gen4. This also works if CNF rewriting is required.

## Related Issue(s)

#7280 

## Checklist


go/vt/vtgate/planbuilder/plan_test.go
```diff
@@ -468,6 +469,10 @@ func escapeNewLines(in string) string {
 }
 
 func testFile(t *testing.T, filename, tempDir string, vschema *vschemaWrapper, checkGen4equalPlan bool) {
+	ksName := ""
+	if vschema.keyspace != nil {
+		ksName = vschema.keyspace.Name
+	}
```
frouioui, line 475: We can probably extract this. We use the same snippet in three different places

frouioui (approved):
Just a nit. Let's double-approve with @systay though

go/vt/vtgate/planbuilder/jointree.go
```diff
@@ -95,6 +97,10 @@ type (
 
 		// columns needed to feed other plans`
 		columns []*sqlparser.ColName
+
+		// The following two fields are used when routing information_schema queries
+		SysTableTableSchema []evalengine.Expr
```
systay, line 102: hmm... we are using `sqlparser.Expr` for predicates here. feels icky to mix both parser stuff and evalengine stuff here. would it make sense to keep this as AST elements and turn them into evalengine thingies in the `jointree_transformers`?

systay (approved):

harshit-gangal:
Will address review comments in following PR

PR #8483, opened 2021-07-16
Fix vreplication error metric

rohit-nayak-ps:
## Description
Metric was setup incorrectly with two dimensions instead of three causing Prometheus errors.


aquarapid (approved):
LGTM

PR #8485, opened 2021-07-16
improve the Gen4NotSupportedF error

systay:
## Description
Very small change to make it easier to find where an error is produced.


frouioui (approved):
Awesome. That will make finding error easier

PR #8494, opened 2021-07-19
Fix Cluster 14 flakiness

frouioui:
## Description

This pull request removes stress testing from cluster 14 end-to-end tests to reduce its flakiness. Stress testing implies bringing up and tearing down more components of our test clusters; however, in most of our E2E tests, the tearing down phase get often "stuck" when a lot of vttablets/vtgates are created/destroyed within the same cluster. Stress testing can be reintroduced once we migrate to a less flaky environment.

## Checklist


deepthi (approved):

deepthi:
@frouioui if this flakiness is happening on release-11.0 branch we may want to port this change over.

frouioui:
@deepthi, noted, stress testing is not part of release 11

Issue #5399, opened 2019-11-04
Support caching_sha2_password authentication plugin


morgo:
#### Feature Description

Vitess only supports mysql_native_password (which was the default up until MySQL 8.0). This is based on sha1, which is now considered cryptographically weak.

This issue is to track adding caching_sha2_password.

aquarapid:
Note that while this FR is (implicitly) for supporting caching_sha2_password in the server (vtgate) code, we should also support it in our client code used by vttablet.

aquarapid:
Another wrinkle is that when talking to (real) MySQL, caching_sha2_password is only supported when the connection is encrypted, unless it's over a local unix_socket.  Something to keep in mind for vttablet supporting "external" MySQL instances.

harshit-gangal:
This will be resolved by https://github.com/vitessio/vitess/pull/6716

deepthi:
Fixed in #6716

PR #8503, opened 2021-07-20
Refactor authentication server plugin mechanism

dbussink:
This change implements a large refactor of the plugin authentication mechanism. This refactor needed for a number of reasons:

- The current AuthServer interface is not well designed. Many of the AuthServer implementations implement at least one if not multiple methods as a panic() as they are unused or they implement something never really called (like the salt generation). This indicates that the current interface doesn't abstract things well.
- Adding `caching_sha2_password` as a server side mechanism would make this interface problem worse. It would add even more methods that would have emply implementations / panic().
- It is impossible right now for a plugin to support multiple authentication mechanisms. The caching_sha2_password plugin allows for caching of passwords and could be used to improve for example the LDAP  plugin. But we can't then still keep the `mysql_clear_password` as a fallback for such an improvement in the LDAP plugin.

Based on these issues, I'm proposing this refactor. It allows for a number of things that also make auth server implementations easier.

- An auth server can implement multiple auth methods. This allows for the above mentioned optimization for the LDAP plugin and many other improvements. Note that these improvements are not implemented here, but it opens up the opportunity.
- There are helpers for all the already supported authentication mechanisms plus basic support for caching_sha2_password. The last only supports TLS connections or Unix sockets right now, since the full authentication protocol is much complicated. The value of the latter is also debatable in the context of Vitess. It depends on a TLS certificate and private key also for the plain text full auth cycle and in that case, TLS would already be available as well anyway. For clients that want caching_sha2_password, switching those to enable TLS seems like a much simpler solution.
- With the helpers, specific interfaces are provided that plugins can implement. Those interfaces provide a simpler way than the complex Negotiate from before which might need to read packets etc. For these helpers, all reading / writing of the protocol is already dealt with and not something auth server plugin writers need to be concerned with.

Given the above advantages, I think the current setup is a better design and better fits what we'd want to use for authentication plugins.

Lastly, this does break any existing plugin implementations that are written in Go and that people might have done themselves in their own Vitess forks etc. I don't know if that's an expected interface / something that needs a larger announcement.

The functionality of all the built in plugins is exactly the same. They have been switched to the new internal APIs, but the external functionality, config files, configuration flags etc. have all stayed exactly the same. 

## Related Issue(s)

https://github.com/vitessio/vitess/issues/5399 is related, but that was closed after only client side support for `caching_sha2_password` was implemented.

With the refactors here, it also becomes possible to implement `caching_sha2_password` for the server side component in Vitess.

## Checklist

Not really sure about the documentation. I don't think we document the internal Go interfaces, only things like configuration flags etc. for plugins that all stay the same?


go/mysql/auth_server.go
```diff
@@ -243,34 +395,207 @@ func EncryptPasswordWithPublicKey(salt []byte, password []byte, pub *rsa.PublicK
...
+func (n *mysqlNativePasswordAuthMethod) HandleUser(conn *Conn, user string) bool {
+	return n.validator.HandleUser(user)
+}
+
+func (n *mysqlNativePasswordAuthMethod) AuthPluginData() ([]byte, error) {
+	salt, err := newSalt()
+	if err != nil {
+		return nil, err
+	}
+	return append(salt, 0), nil
```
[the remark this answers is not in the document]
dbussink: @vmg The protocol defines the auth plugin data as a String<EOF> which means it's the last data in a packet and it's defined as ending when the packet ends, see also https://dev.mysql.com/doc/internals/en/connection-phase-packets.html & https://dev.mysql.com/doc/internals/en/string.html#packet-Protocol::RestOfPacketString.

In the handshake packet it's defined as `String<len>` so it means that it has an explicit length. This means that the `0` byte is not part of the wire protocol and an implementation detail of the specific `mysql_native_password` & `caching_sha2_password` auth methods. So I think that means it's correct here that it's considered part of `AuthPluginData()`.

go/mysql/auth_server.go
```diff
@@ -243,34 +395,207 @@ func EncryptPasswordWithPublicKey(salt []byte, password []byte, pub *rsa.PublicK
...
+func (n *mysqlClearAuthMethod) Name() AuthMethodDescription {
+	return MysqlClearPassword
+}
+
+func (n *mysqlClearAuthMethod) HandleUser(conn *Conn, user string) bool {
+	return n.validator.HandleUser(user)
+}
+
+func (n *mysqlClearAuthMethod) AuthPluginData() ([]byte, error) {
+	return make([]byte, 0), nil
```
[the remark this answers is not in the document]
dbussink: Fixed!

go/mysql/auth_server.go
```diff
@@ -243,34 +395,207 @@ func EncryptPasswordWithPublicKey(salt []byte, password []byte, pub *rsa.PublicK
...
+	password, err := readPacketPasswordString(c)
+	if err != nil {
+		return nil, err
+	}
+
+	return n.storage.UserEntryWithPassword(c.GetTLSClientCerts(), user, password, remoteAddr)
+}
+
+// authServers is a registry of AuthServer implementations.
+var authServers = make(map[string]AuthServer)
```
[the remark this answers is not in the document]
dbussink: @vmg Ah, this code isn't new but code that moved around location wise only. I can still add it, but don't know if there was a reason this never has seen a lock so far?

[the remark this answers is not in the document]
dbussink: Added locking!

go/mysql/auth_server.go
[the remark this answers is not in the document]
vmg: :+1:

vmg (approved):
This looks good to me. It's complex code but I don't see how to make it significantly simpler. It just has to do a lot!

go/mysql/auth_server.go
```diff
@@ -17,109 +17,233 @@ limitations under the License.
...
+	// the initial server handshake. This needs to be either `mysql_native_password`
+	// or `caching_sha2_password` as those are the only supported auth
+	// methods during the initial handshake.
+	//
+	// It's not needed to also support those methods in the AuthMethods(),
+	// in fact, if you want to only support for example clear text passwords,
+	// you must still return `mysql_native_password` or `caching_sha2_password`
+	// here and the auth switch protocol will be used to switch to clear text.
+	DefaultAuthMethodDescription() AuthMethodDescription
+}
```
harshit-gangal, line 63: nit: comment to start with function name `DefaultAuthMethodDescription`

go/mysql/auth_server.go
```diff
@@ -17,109 +17,233 @@ limitations under the License.
...
+// that with AuthNeedMoreData.
+type CacheState int
+
+const (
+	// AuthRejected is used when the cache knows the request can be rejected.
+	AuthRejected CacheState = iota
+	// AuthAccepted is used when the cache knows the request can be accepted.
+	AuthAccepted
+	// AuthNeedMoreData is used when the cache doesn't know the answer and more data is needed.
+	AuthNeedMoreData
```
harshit-gangal, line 117: not able to find the usages of `AuthNeedMoreData`

go/mysql/auth_server_vault.go
```diff
@@ -140,15 +136,66 @@ func newAuthServerVault(addr string, timeout time.Duration, caCertPath string, p
...
+}
+
+// UserEntryWithHash is called when mysql_native_password is used.
+func (a *AuthServerVault) UserEntryWithHash(userCerts []*x509.Certificate, salt []byte, user string, authResponse []byte, remoteAddr net.Addr) (Getter, error) {
+	a.mu.Lock()
+	userEntries, ok := a.entries[user]
+	a.mu.Unlock()
+
+	if !ok {
+		return &StaticUserData{}, NewSQLError(ERAccessDeniedError, SSAccessDeniedError, "Access denied for user '%v'", user)
```
harshit-gangal, line 175: nit: can return nil instead of `&StaticUserData{}`. Is there a reason to still return other than nil, this is also at other places as well.

harshit-gangal (approved):

go/mysql/auth_server.go
dbussink: @harshit-gangal Ah yeah, that's because there's 3 possible states and the other two get checked and the fallback then equates this one. I can also change this to a switch statement if that's preferred with a default that returns an error on an unknown state?

go/mysql/auth_server_vault.go
dbussink: @harshit-gangal I copied this from how it was already working and assumed that it would need an object to be returned and not `nil`.

go/mysql/auth_server.go
dbussink: Will fix!

go/mysql/auth_server.go
dbussink: I've updated it to a switch statement to be more explicit here and reference the value.

PR #8504, opened 2021-07-20
mysql/server: abort connection if secure transport is required

vmg:
## Description

Found while reviewing https://github.com/vitessio/vitess/pull/8503 (by @dbussink). This is an actual error in the auth code, that should abort right away if TLS is required and the connection doesn't have it. The logic error has been there for a while.

I'm thinking this is probably worth a backport just for safety, even though it may not be exploitable. cc @deepthi 

## Checklist


dbussink (approved):
Thanks for catching this one.

deepthi (approved):
LGTM
Let's backport to all currently supported releases, and let's also get this into the 11.0 branch.

systay (approved):

PR #8509, opened 2021-07-20
naming: remove MasterStatus proto and use PrimaryStatus everywhere

deepthi:
## Description

In #8446 we introduced PrimaryStatus protobuf message. In this PR, all uses of MasterStatus proto have been replaced with PrimaryStatus.
IIUC protobuf rules correctly, this should be fully backwards-compatible.

## Related Issue(s)

#7114 

## Checklist


vmg (approved):
Still compatible on the wire :+1:

Issue #8462, opened 2021-07-13
Request Buffering during Sharding


vmg:
I have started working on this functionality for Vitess. To ellaborate on the problem, which is already listed in https://github.com/vitessio/vitess/issues/7061: we want to improve the way we perform buffering to be able to handle more complex resharding situations.

Currently, the main entrypoint for buffering is in `tabletgateway.go`, in the `TabletGateway.withRetry` helper that wraps all the method calls for our `QueryService`:

https://github.com/vitessio/vitess/blob/4bebb8e5a6097772bcb8a4d1c187f14782158069/go/vt/vtgate/tabletgateway.go#L216-L234

Here's a few important things to note:

1. There is another buffering entrypoint in `DiscoveryGateway`, but this whole API has been deprecated in favor of `TabletGateway`, so we can ignore it. :ok_hand:
1. Buffering does not happen for queries in an existing transaction because that's never safe to do
2. Buffering only happens for queries that target a `master` tablet
3. **Critically**: buffering only happens lazily, once a query has failed at least once and is being retried, which is a behavior we also want to change
4. **Also critically**: buffering is scoped for each keyspace and shard pair.

Let's look at the two critical issues which we intend to fix:

The way buffering works right now is by keeping a table of "buffers" **for every keyspace and shard pair**. This "shard buffer" is created lazily the first time a query for a given keyspace+shard arrives to the `vtgate`, and by default it is **disabled**, meaning that requests pass through directly and are not buffered. The first time that a request to a given keyspace/shard returns an error, the retry code in `TabletGateway` will retry the buffering before retrying the request, this time passing the upstream error. If the buffering code detects that the upstream error is a failover error (this is explicitly checked at https://github.com/vitessio/vitess/blob/30250f51afd98dc1629b233959b60a916d88f7ff/go/vt/vtgate/buffer/buffer.go#L263-L290), then the buffer for the keyspace+shard will be marked as "enabled", and the current query will start buffering. Any subsequent queries that target the same keyspace+shard will _also_ start buffering once they reach the buffering check, because their buffer is now marked as "enabled".

This is what we mean by "lazy buffering": the only way a request will start buffering is once it has failed at least once for its keyspace+shard. The retry for this request, and any follow up requests, will then buffer. In an ideal design for this system, **the buffering code will be proactively health checking the different shards in the cluster** so that we can buffer a request right away once we detect a failover, without having to use an error response as the detection mechanism for the failover.

[...]

PR #8514, opened 2021-07-21
topo: add topo events watcher

vmg:
## Description

Yay! Finally some actual code for https://github.com/vitessio/vitess/issues/8462!

The fist step, as suggested by @harshit-gangal, is updating our topology server to keep track of any new events that affect cluster availability, such as resharding or reparenting.

To accomplish this, I've modeled a new `topodatapb.TopoEvent` with all the information that can could possibly be needed to keep track of these events. Then, I've created a new `TopoServer` API to write & watch these events.

Regarding the Topo API: ideally every individual event would go under its own path, but the generic `topo.Conn` doesn't support watching arbitrary prefixes in the K/V store (it works for `etcd`, but just by coincidence), so receiving these events in real time would be expensive. To work around this, we're writing a ProtoBuf _list_ to a single key. Since every event is uniquely identified by an UUID, it's very easy to watch and stream to the callers any new events between each individual update.

cc @harshit-gangal @deepthi @rohit-nayak-ps 

## Related Issue(s)

## Checklist


sougou:
Depending on the topo event mechanism has burnt us in the past. This is why the direct health-check mechanism was implemented between vtgates and vttablets. The previous code used to wait for topo to be updated after a failover. At massive scale, you essentially can't rely on the topo being highly available.

However, we already have code that depends on topo being available for handling resharding cut-overs. So, this approach doesn't make it worse right now.

At some point, we have to come up with a better overall cutover design that doesn't depend on the topo, and we can transition to that method at that point.

In other words, this is ok for now.

proto/topodata.proto
```diff
@@ -421,3 +421,38 @@ message ExternalVitessCluster {
...
+  repeated TopoEvent log = 1;
+}
+
+message TopoEvent {
+  string cell = 1;
+  vttime.Time started_at = 2;
+  string uuid = 3;
+  string keyspace = 4;
+
+  message EmergencyReparenting {
```
deepthi, line 435: Can we leave off the `ing` from these names?

go/vt/topo/cluster_event.go
```diff
@@ -0,0 +1,143 @@
...
+import (
+	"context"
+	"sort"
+
+	"google.golang.org/protobuf/proto"
+	topodatapb "vitess.io/vitess/go/vt/proto/topodata"
+	"vitess.io/vitess/go/vt/vterrors"
+)
+
+type WatchTopoEventData struct {
```
deepthi, line 12: A top-level comment on the struct will be good. What it encapsulates + how it is intended to be used.
(I know most source files in vitess don't currently have this, but we have to start somewhere).

deepthi (commented):
LGTM, a couple of comments.

vmg:
Green with tests & comments :ok_hand:

vmg:
I've started working on reading the data from the topo server and noticed that keeping the data in a list is not particularly useful. When a VTGate spins up and joins the topo server, it gets the list of all the topology events, but it doesn't know which ones are still active.

We discussed with @harshit-gangal that it's better if the gates poll the state of the tablets manually to figure out when a given resharding/reparenting operation is over, but I think it's important that we are also able to mark events as resolved/finished, so that VTGates that join a cluster late don't have to re-try complex health checks to figure out whether _all_ of the events are truly finished.

I've implemented this functionality by storing the events in the topo server in a map.

harshit-gangal:
My thinking was that only active operation will be available to discovery, rest will be removed once they are completed (finished / failed).

harshit-gangal:
> My thinking was that only active operation will be available to discovery, rest will be removed once they are completed (finished / failed).

Also, it can have timestamp and completion time, post that it can be assumed that it is a non-active operation.

harshit-gangal (approved):
Overall looks good. Left comment about having timeout value in the event so as to resolve the events if finished_time is not updated.

rafael:
The code changes look good. But I have similar concerns as @sougou. The overall direction we have been thinking for these kinds of operations is to not rely on the topology. I would argue that the approach here makes it slightly worse as we add more dependencies depending on the topo. It seems like we will be accruing some technical debt. 

Another angle is that I'm not 100% if this approach will be that effective in practice. It seems to me that the propagation of the signal will be competing with the actual cutover. By the time some of the vtgates get the topo event, the cutover could have finished. Am I missing something here? 

For practical purposes, we have solved this problem with retries on the client. It's not super elegant, but it's straightforward. We have code in our application that retries error `1105`. That error code is the one returned by Vitess when it's in the middle of a cutover. 

What do y'all think about using a similar approach in the open source? We could return an explicit error code that means: we are cutting over, retry shortly. It might not be ideal in the sense that clients will need to be aware of this Vitess specific error code, but achieving perfect buffering seems complicated to maintain and guarantee.

rafael:
Well, after reading your issue and not just this PR, the intent of your design is exactly not have to use error codes as the signal and be proactive :D 

However - I'm not sure if that ideal design is something that we can guarantee.

I think what we are trying to do here is doable with error codes and I would still push back on the actual benefits of the new approach.

harshit-gangal:
> Well, after reading your issue and not just this PR, the intent of your design is exactly not have to use error codes as the signal and be proactive :D
> 
> However - I'm not sure if that ideal design is something that we can guarantee.
> 
> I think what we are trying to do here is doable with error codes and I would still push back on the actual benefits of the new approach.

The problem here is in re-sharding, as the actually query needs to be replanned and not the query which has reached the vttablet. 

The query is broken down into multiple query (may also have lookup query), these query even on retry will not work and has to be buffered at higher level which needs proactive detection.

deepthi (changes requested):
I'm going to block this PR until feedback is addressed just so that someone doesn't hit the green button accidentally.

rafael:
Ah I see. I was discussing that bit about replanning with @deepthi yesterday. I see the complications with the retry inside the vtgate.

It would be great if we can provide a reliable proactive mechanism, but I don't think we are achieving that here. Expanding a bit more on the concerns:

* **Scale and reliability:** Adding watches to the global topo from vtgates, will add significant pressure to the global topology. This won't scale for large installations. We could address this by pushing the events through the local cells. However, that will come with its tradeoffs as well. As far as I know, there is no single watch in the global topo from vtgates. 
* **Latency in events**: My intuition (which could be wrong) is that by the time these events are propagated through the topology, the cutover already happened. It seems possible that many queries will fail because of these delays.

At the cost of not solving this problem in the vtgate, I'm still thinking that propagating an error that clients can use to decide if they retry seems a more effective and straightforward way of achieving a similar objective here.  

Maybe there are other approaches we can think that don't require the topo and achieve the proactive goal.

deepthi:
@vmg we can close this PR now, correct?

PR #8520, opened 2021-07-22
Naming: tabletmanagerservice Master->Primary terminology

shlomi-noach:
## Description

Renaming (backwards compatible) the remaining `Master` termed functions and messages.
## Related Issue(s)

#7114 

## Checklist


proto/tabletmanagerservice.proto
```diff
@@ -153,7 +159,10 @@ service TabletManager {
   rpc ReplicaWasPromoted(tabletmanagerdata.ReplicaWasPromotedRequest) returns (tabletmanagerdata.ReplicaWasPromotedResponse) {};
 
   // SetMaster tells the replica to reparent
-  rpc SetMaster(tabletmanagerdata.SetMasterRequest) returns (tabletmanagerdata.SetMasterResponse) {};
+  rpc SetMaster(tabletmanagerdata.SetPrimaryRequest) returns (tabletmanagerdata.SetPrimaryResponse) {};
+
+  // SetMaster tells the replica to reparent
+  rpc SetPrimary(tabletmanagerdata.SetPrimaryRequest) returns (tabletmanagerdata.SetPrimaryResponse) {};
```
deepthi, line 165: Over on the tabletmanager side I chose to call this `SetReplicationSource` instead of `SetPrimary`. That seemed more apt in the context of replication.
So.. cluster management level master -> primary
MySQL replication level master -> replication source.
Can we change this to be consistent?

deepthi (approved):
Mostly looks good except for the one suggestion.
Approving provisionally. Can be merged after feedback is addressed.

proto/tabletmanagerservice.proto
shlomi-noach: done

shlomi-noach:
Renamed `SetPrimary[Request,Response]` as `SetReplicationSource[Request,Reponse]`

PR #8524, opened 2021-07-23
[DO NOT MERGE] Change planner version to Gen4 as default

harshit-gangal:
Trying to see how the e2e tests behave with Gen4 as the query planner.


systay:
replaced by #8858

PR #8525, opened 2021-07-23
go/mysql/auth: fix error message for wrong auth method.

fatih:
Should be `mysql_clientcert_auth_method`.


dbussink (approved):
Thanks for catching this!

harshit-gangal (approved):

PR #8526, opened 2021-07-23
Initial pass at some vtadmin docs

ajm188:

## Description

This is all I had time to put together before a vacation, and something is better than nothing!

## Checklist


doeg (approved):
This looks so great!!! Thank you so much for writing all this down. I'll do a documentation pass soon, too, especially to consolidate some of the front-end docs.

At some point (per our conversation with @deepthi) we could consider moving these to https://vitess.io, which might make them more generally discoverable (at the expense of colocating it with the code, I guess). 

Thank you again!

PR #8527, opened 2021-07-24
Change local example to use v2 vreplication flows

rohit-nayak-ps:
## Description

This PR 
* makes v2 flows as the default and requires `-v1` to be specified to use the v1 `MoveTables` or `Reshard`
* changes the local examples to use the v2 vreplication CLI UX. 
* The vitess.io documentation has been updated at [https://github.com/vitessio/website/pull/784](https://github.com/vitessio/website/pull/784). It will be merged once 10.0 docs are archived.
* V1 tests have been modified to add the `v1` parameter to the `MoveTables` and `Reshard` vtctl commands
* V2 tests have been modified to remove the `v2` parameter to the vtctl commands
* A deprecation statement is displayed if the v1 commands are used

## Notes 
* The v2 flows have been around for over six months know. Since they are not the default, they have not seen a lot of usage in the field. So we continue to call the v2 flow as experimental and make it the default to push it forward.
* This will be first merged  in 11.0 and then later into main 

## Checklist

### DIFF
--- a/examples/local/202_move_tables.sh
+++ b/examples/local/202_move_tables.sh
@@ -19,4 +19,4 @@
 
 source ./env.sh
 
-vtctlclient MoveTables -workflow=commerce2customer commerce customer '{"customer":{}, "corder":{}}'
+vtctlclient MoveTables -source commerce -tables 'customer,corder' Create customer.commerce2customer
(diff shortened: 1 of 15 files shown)


shlomi-noach (approved):
Looks good

PR #8529, opened 2021-07-26
Enhance PRS error message

### LINKED ISSUE #8528: PRS will not report why a new primary cannot be elected if no eligible tablets in the current cell
Scenario:
  * If you have a keyspace with tablets spread across multiple cells, and no replicas in the same cell as the current primary, a PRS without an explicit `-new_master`/`new_primary` directive will fail like this:
  ```
$ vtctlclient -server 127.0.0.1:16999 PlannedReparentShard -keyspace_shard=keyspace1/-2a 
PlannedReparentShard Error: rpc error: code = Unknown desc = cannot find a tablet to reparent to
E0706 21:31:25.627705 1067782 main.go:72] remote error: rpc error: code = Unknown desc = cannot find a tablet to reparent to
  ```
  * This message does not help the user to understand why the PRS failed, since there are eligible tablets available for promotion, just not in the primary's cell.

Suggestion:
  * Update the error message

We already updated the documentation at https://vitess.io/docs/reference/programs/vtctl/shards/#plannedreparentshard previously to explain this behavior.

aquarapid:
Report an informative error message if there is no replica in primary's cell to PRS to.
Fixes #8528 

### DIFF
--- a/go/test/endtoend/reparent/reparent_test.go
+++ b/go/test/endtoend/reparent/reparent_test.go
@@ -276,7 +276,7 @@ func TestReparentAvoid(t *testing.T) {
 	stopTablet(t, tab1, true)
 	out, err := prsAvoid(t, tab2)
 	require.Error(t, err)
-	assert.Contains(t, out, "cannot find a tablet to reparent to")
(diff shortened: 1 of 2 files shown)


deepthi (approved):
:+1:

PR #8535, opened 2021-07-26
Adds some more orchestrator tests to vtorc

GuptaManan100:
## Description
This PR ports some more of the tests from orchestrator to vtorc. The following tests have been added - 

- A replica from the same cell must be promoted instead of one from another cell if `PreventCrossDataCenterMasterFailover` is set to true in the configuration
- A lagging replica is promoted and not a rdonly instance. We choose to lose the transactions that the rdonly tablet has acknowledged.
- Test that recovery from primary failure succeeds if the time lag promotion rule is satisfied
- Test that recovery from primary failure fails if the time lag promotion rule is not satisfied
- Test that we promote the tablet marked as preferred in case of a failure
- Test that we promote the tablet marked as preferred even if it is lagging in case of failure. Also it should have caught up before it is promoted
- Test that we do not promote a tablet marked as preferred if it lives in a different cell and we have set `PreventCrossDataCenterMasterFailover` to true.

## Related Issue(s)

- #6769

## Checklist


go/vt/orchestrator/inst/durability.go
```diff
@@ -161,3 +177,46 @@ func (d *durabilityCrossCell) replicaSemiSync(master, replica *topodatapb.Tablet
 	}
 	return false
 }
+
+//=======================================================================
+
+type durabilitySpecified struct {
+	promotionRules map[string]CandidatePromotionRule
+}
+
```
deepthi, line 186: Can you add some comments describing what this does?

deepthi (approved):
Mostly LGTM. You may merge after adding the requested comments.

go/vt/orchestrator/inst/durability.go
GuptaManan100: Added comments in the last commit.

PR #8537, opened 2021-07-26
Fix handshake protocol with MySQL client

dbussink:
This change fixes the handshake protocol with the MySQL client. The 8.0.25 and earlier MySQL client will always send `caching_sha2_password` as the default (if not configured otherwise), but it won't yet send the actual scramble.

That handshake indicates that the client does support that method, but it doesn't send any yet sensitive information over the wire. Our current logic assumes that if a certain method is sent by the client, it also sends a complete authentication plugin data, but that's not the case.

The way to request the client to send the full data is to return an auth switch packet with the same method again.

This change fixes any auth server that tries to implement multiple methods at once like mysql_native_password and caching_sha2_password so it doesn't error out on clients like MySQL <= 8.0.25.

On 8.0.26 they changed the client behavior to use the server provided auth method as the default and in that case it sends the complete data already so this problem is not an issue.

Additionally I added a small refactor to remove the plain text implementation awareness for dialog and clear authentication to an interface for the AuthMethod instead so the server protocol needs to know less implementation details.

### DIFF
--- a/go/mysql/auth_server.go
+++ b/go/mysql/auth_server.go
@@ -78,6 +78,11 @@ type AuthMethod interface {
 	// depends on what you have.
 	HandleUser(conn *Conn, user string) bool
 
+	// AllowClearTextWithoutTLS identifies if an auth method is allowed
+	// on a plain text connection. This check is only enforced
+	// if the listener has AllowClearTextWithoutTLS() disabled.
+	AllowClearTextWithoutTLS() bool
+
 	// AuthPluginData generates the information for the auth plugin.
 	// This is included in for example the auth switch request. This
 	// is auth plugin specific and opaque to the Mysql handshake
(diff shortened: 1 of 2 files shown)


go/mysql/auth_server.go
```diff
@@ -417,6 +431,14 @@ func (n *mysqlNativePasswordAuthMethod) AuthPluginData() ([]byte, error) {
 	return append(salt, 0), nil
 }
 
+func (n *mysqlNativePasswordAuthMethod) HasCompleteAuthPluginData(authResponse []byte) bool {
+	return len(authResponse) > 0
```
vmg, line 435: All implementations for this method are identical, `len(data) > 0`. Do you think that means this method is superfluous and instead should just be an `if len(data) > 0` inline?

go/mysql/server.go
```diff
@@ -378,14 +378,24 @@ func (l *Listener) handle(conn net.Conn, connectionID uint32, acceptTime time.Ti
...
-			if m.HandleUser(c, user) {
-				negotiatedAuthMethod = m
-				break
+
+	// We need to send down an additional packet if we either have no negotiated method
+	// at all or incomplete authentication data.
+	//
+	// The latter case happens for example for MySQL 8.0 clients until 8.0.25 who advertise
+	// support for caching_sha2_password by default but with no plugin data.
+	if err != nil || !negotiatedAuthMethod.HasCompleteAuthPluginData(clientAuthResponse) {
```
vmg, line 387: I would just do a `len` check here.

dbussink: Doing the length check here now. We can always add back an abstraction if there's contextual knowledge needed.

vmg (approved):
Semms gucci :gucci:

harshit-gangal (approved):

PR #8543, opened 2021-07-27
Gen4: Count Distinct support

harshit-gangal:
## Description

Add support for count and sum distinct

## Related Issue(s)

#7280 

## Checklist


go/vt/vtgate/engine/ordered_aggregate_test.go
```diff
@@ -790,3 +790,185 @@ func TestOrderedAggregateExecuteGtid(t *testing.T) {
 	)
 	assert.Equal(t, wantResult, result)
 }
+
```
systay, line 793: the tests in this file are a little difficult to read. grouping columns and aggregating columns are expected to have very similar results, which makes it hard to overview and understand

go/vt/vtgate/planbuilder/ordered_aggregate.go
```diff
@@ -303,6 +303,31 @@ func (oa *orderedAggregate) needDistinctHandling(pb *primitiveBuilder, funcExpr
...
+func (oa *orderedAggregate) needDistinctHandlingGen4(funcExpr *sqlparser.FuncExpr, opcode engine.AggregateOpcode, semTable *semantics.SemTable, vschema ContextVSchema) (bool, *sqlparser.AliasedExpr, error) {
+	if !funcExpr.Distinct {
+		return false, nil, nil
+	}
+	if opcode != engine.AggregateCount && opcode != engine.AggregateSum {
+		return false, nil, nil
+	}
+	innerAliased, ok := funcExpr.Exprs[0].(*sqlparser.AliasedExpr)
+	if !ok {
+		return false, nil, fmt.Errorf("syntax error: %s", sqlparser.String(funcExpr))
```
systay, line 318: `fmt.Errorf`? should this not be a vterrors error instead?

go/vt/vtgate/planbuilder/ordered_aggregate.go
```diff
@@ -303,6 +303,31 @@ func (oa *orderedAggregate) needDistinctHandling(pb *primitiveBuilder, funcExpr
 	return true, innerAliased, nil
 }
 
+// needDistinctHandling returns true if oa needs to handle the distinct clause.
+// If true, it will also return the aliased expression that needs to be pushed
+// down into the underlying route.
+func (oa *orderedAggregate) needDistinctHandlingGen4(funcExpr *sqlparser.FuncExpr, opcode engine.AggregateOpcode, semTable *semantics.SemTable, vschema ContextVSchema) (bool, *sqlparser.AliasedExpr, error) {
```
frouioui, line 309: Comment should be `// needDistinctHandlingGen4  ...`

frouioui (approved):

go/vt/vtgate/planbuilder/horizon_planning.go
```diff
@@ -145,19 +145,61 @@ func (hp *horizonPlanning) planAggregations() error {
...
+		// push all expression if they are non-aggregating or the plan is not ordered aggregated plan.
+		if !e.Aggr || oa == nil {
+			_, _, err := pushProjection(e.Col, hp.plan, hp.semTable, true, false)
+			if err != nil {
+				return err
+			}
+			continue
 		}
+
 		if e.Aggr && oa != nil {
```
systay, line 157: should this if not be removed altogether? we should always do this, if the previous `if` wasn't executed, right?

go/vt/vtgate/planbuilder/horizon_planning.go
```diff
@@ -145,19 +145,61 @@ func (hp *horizonPlanning) planAggregations() error {
...
-			oa.eaggr.Aggregates = append(oa.eaggr.Aggregates, engine.AggregateParams{
+			handleDistinct, innerAliased, err := oa.needDistinctHandlingGen4(fExpr, opcode, hp.semTable, hp.vschema)
+			if err != nil {
+				return err
+			}
+
+			// Currently the OA engine primitive is able to handle only one distinct aggregation function.
+			// PreProcess being true tells that it is already handling it.
+			if oa.eaggr.PreProcess && handleDistinct {
+				return semantics.Gen4NotSupportedF("multiple distinct aggregation function")
```
systay, line 171: this is a limitation with the engine, and not the planner, if I understand it correctly. `Gen4NotSupportedF` should only be used to mark queries the gen4 planner should handle but doesn't. this is different, IMHO

systay (changes requested):
Left a few comments

go/vt/vtgate/planbuilder/ordered_aggregate.go
systay: we have very little gen4 code in these files. I think this method should live on `horizon_planning.go` and not here

systay (approved):

PR #8546, opened 2021-07-27
Enable GitHub Actions concurrency feature

frouioui:
## Description

This pull request is the second attempt at introducing GitHub Actions [concurrency feature](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#concurrency) into our builds. The goal is to limit the overall number of queued and in-progress builds for this repository.

## How does it work?

When a new commit is added to a pull request, GitHub Actions will add each workflow to groups. Groups are defined like this: `branch_name + workflow_name`, thus every workflow of the pull request's branch will be in their own unique group. GitHub Actions concurrency feature uses these groups to cancel workflow when new workflows (due to a new commit being pushed) are added to the queue. If for a given group there is a more recent workflow in the queue, the older workflows of the same group will be canceled by GitHub Actions, and only the most recent one will be executed.

## Demo

In this pull request commits, the commit 7f3acfb was pushed, and very soon after the empty commit 89ff009 was also pushed. In the build status of 7f3acfb, we can see that most of the builds are marked as canceled, this is due to 89ff009 being pushed.

## Related Issue(s)
#8432

### DIFF
--- a/.github/workflows/cluster_endtoend_11.yml
+++ b/.github/workflows/cluster_endtoend_11.yml
@@ -2,6 +2,10 @@
 
 name: Cluster (11)
 on: [push, pull_request]
+concurrency:
+  group: ${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
 
   build:
--- a/.github/workflows/cluster_endtoend_11.yml
+++ b/.github/workflows/cluster_endtoend_11.yml
@@ -3,11 +3,10 @@
 name: Cluster (11)
 on: [push, pull_request]
 concurrency:
-  group: ${{ github.ref }}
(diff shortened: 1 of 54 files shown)


shlomi-noach (approved):
Duplicate of https://github.com/vitessio/vitess/pull/8432? Anyway, go for it and let's see if that makes things better or worse. I see @GuptaManan100 is meanwhile importing BuildKite FYI

frouioui:
@shlomi-noach - Right, this will be temporary while we are on GH Actions. Regarding the duplicate, I thought it would be better to start over in a new pull request since the builds seemed to be hanging forever, though the "more complex group names" that got added by `7f3acfb` allowed me to understand why the builds were "stuck"

PR #8548, opened 2021-07-27
Disable allowing set statements on system settings by default

harshit-gangal:
Disable changing system settings without allowing it explicitly.

### DIFF
--- a/go/vt/vtgate/vtgate.go
+++ b/go/vt/vtgate/vtgate.go
@@ -80,7 +80,7 @@ var (
 	warnPayloadSize    = flag.Int("warn_payload_size", 0, "The warning threshold for query payloads in bytes. A payload greater than this threshold will cause the VtGateWarnings.WarnPayloadSizeExceeded counter to be incremented.")
 
 	// Put set-passthrough under a flag.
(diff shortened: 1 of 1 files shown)


frouioui (approved):

systay (approved):

harshit-gangal:
Closing this as we want to make this default `on`

deepthi:
A bit more context on the decision:
* When system settings was introduced, it was felt that because it was new, the feature should be OFF on release branches, but ON on `main`.
* Over time we have discovered that frameworks or clients that always use session-level settings lose the advantages of vitess connection pooling (because every connection has different settings from the "defaults" for the system).
* Because of this, we debated whether to turn it off by default on `main` as well as releases/
* The final outcome of the discussion was that this feature is in fact stable now, so we do want to leave it ON on `main`, and also change the default to ON on the next release (12.0).
* This leads to a better user experience because more queries will work as expected with this feature enabled.
* Users who run into issues at scale because of the non-use of connection pools are advised to change the defaults at the MySQL level such that most client connections are uniform and any SET statements issued by frameworks turn into no-ops.

PR #8565, opened 2021-07-29
Adds support for resolving columns using column vindexes in Gen4

systay:
## Description
Adds support for resolving columns using column vindexes in Gen4

## Related Issue(s)
#7280 

## Checklist

### DIFF
--- a/go/vt/vtgate/semantics/semantic_state.go
+++ b/go/vt/vtgate/semantics/semantic_state.go
@@ -305,6 +305,10 @@ func vindexTableToColumnInfo(tbl *vindexes.Table) []ColumnInfo {
 		})
 		nameMap[col.Name.String()] = nil
 	}
+	// If table is authoritative, we do not need ColumnVindexes to help in resolving the unqualified columns.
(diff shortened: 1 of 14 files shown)


go/vt/vtgate/semantics/semantic_state.go
```diff
@@ -296,12 +296,26 @@ func vindexTableToColumnInfo(tbl *vindexes.Table) []ColumnInfo {
...
+		for _, column := range vindex.Columns {
+			name := column.String()
+			if _, exists := nameMap[name]; exists {
+				continue
+			}
+			cols = append(cols, ColumnInfo{
+				Name: name,
+			})
+			nameMap[name] = nil
+		}
```
harshit-gangal, line 318: optimization: this is not required if table is authoritative.

harshit-gangal (approved):

go/vt/vtgate/semantics/semantic_state.go
systay: ah, right. yeah, we could avoid reading the vindexes in that situation

PR #8566, opened 2021-07-29
topodata: remove deprecated field 'ServedTypes'

vmg:
## Description

We found this during some code review yesterday. This field is not needed anymore.

cc @deepthi @rohit-nayak-ps 

## Checklist


vmg:
@deepthi: I also regenerated the Javascript/Typescript ProtoBuf files, that hadn't been regenerated in a long while. They still had all the old master/primary naming.

rafael:
Ah, this is great — long-overdue cleanup. How to rollout this change is something we will need to address. This field is not required anymore, but it is there for backward compatibility. Clusters created before we introduced`IsMasterServing`,  will have ServedTypes set up correctly, but `isMasterServing` won't have the correct information. 

If we merge this as is, we will break old clusters when rebuilding the [serving keyspace graph](https://github.com/vitessio/vitess/pull/8566/files#diff-5df7a2712c34f0a6e3cd88c46e1ee4439dc8d0ee105dca266470f6c1e5a5f446L116). This will leave a keyspace with no valid shards to route to. 

Two options that we could explore:
* We can say that this will be a non backward compatible change and that folks need to update `IsMasterServing` before upgrading to the version where we include this change. We can provide instructions on how to perform such an update. This approah is what we did at Slack. 
* We can make it transparent to Vitess users and split this cleanup into two parts. First, we add logic to "upgrade" when we detect a cluster is using the deprecated field (i.e ServedTypes is set). After that goes out in a release, we merge the cleanup. 

The first approach is more straightforward from the release management perspective, but it could be really bad if someone misses the instructions to upgrade correctly. 

The second approach will be slower as it will require multiple releases to do the cleanup. But it will be a better experience. 

What are your thoughts on this?

vmg:
Thanks for that heads up on the upgrade. That makes sense to me. I think a clear explanation on the release notes that this is a breaking change is much more straightforward. The second option that you suggest is not going to work well in practice for people who skip Vitess releases: not everybody upgrades into each Vitess release, and people who skip the specific version that upgrades clusters with the deprecated field will crash badly when they eventually upgrade to the newest version. 

Given the amount of people who skip Vitess releases, and our release frequency, I think the second option will actually be a worse experience with more chances to break for users. @deepthi what do you think?

rafael:
I don't have strong opinions on this. So both approaches sound reasonable to me. 

However, I don't agree that the second option will be worse experience. To your point, if users are skipping versions (which they shouldn't in my opinion),it will be even more likely that they will miss the release notes.

We should encourage folks to not skip versions. It should be standard practice to rollout deprecations across multiple major releases and provide clean upgrades that way. If users decide to skip versions, then they will be on their own. 

Again, for this particular case, I think release notes will be fine. This is now very old code, so likeness of users with keyspaces created this way, I think is very low. But adding these thoughts, because I think it's an interesting discussion that we should align on how we handle them.

deepthi:
I agree with everything @rafael says here: https://github.com/vitessio/vitess/pull/8566#issuecomment-889249373

In general we do want to take the proper deprecation cycle approach. We can be a bit lazier for proto field name changes because they don't have any meaningful impact. RPC Request/Response field changes also fall into this category.

In this case, because it affects data stored in the topo (which in turn affects whether a running vitess system will continue to serve queries during an upgrade), the approach has to include some sort of automated or manual migration.

However, I'm ok with us taking a release notes / manual migration approach for this change because it _has_ been several releases since the field was deprecated, and because the author of the deprecation helpfully created a vtctl command that can be used to perform the necessary repair.

deepthi:
@doeg can you take a look at the .js/.ts changes and give us a :+1: ?

rafael (approved):

deepthi:
> @doeg can you take a look at the .js/.ts changes and give us a  ?

We can do a follow up PR if necessary. Merging this now.

PR #8568, opened 2021-07-30
gen4: Refactoring of semantic analysis

systay:
## Description
Splits up the big complex `analyzer.go` into smaller pieces that should be easier to understand.

## Related Issue(s)
#7280 

## Checklist

### DIFF
--- a/go/vt/vtgate/semantics/typer.go
+++ b/go/vt/vtgate/semantics/typer.go
@@ -1,0 +1,59 @@
+/*
+Copyright 2021 The Vitess Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package semantics
+
+import (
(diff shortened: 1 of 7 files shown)


go/vt/vtgate/semantics/typer.go
```diff
@@ -0,0 +1,59 @@
...
+			t.exprTypes[node] = sqltypes.Int32
+		case sqlparser.StrVal:
+			t.exprTypes[node] = sqltypes.VarChar
+		case sqlparser.FloatVal:
+			t.exprTypes[node] = sqltypes.Decimal
+		}
+	case *sqlparser.FuncExpr:
+		if node.Name.EqualString("count") {
+			t.exprTypes[node] = sqltypes.Int32
+		}
```
harshit-gangal, line 51: I think we should change the expr type for `count` to be uint64

harshit-gangal: OA uses
```
		AggregateCountDistinct: sqltypes.Int64,
		AggregateSumDistinct:   sqltypes.Decimal,
```

go/vt/vtgate/semantics/typer.go
```diff
@@ -20,17 +20,18 @@ import (
 	"vitess.io/vitess/go/sqltypes"
 	querypb "vitess.io/vitess/go/vt/proto/query"
 	"vitess.io/vitess/go/vt/sqlparser"
+	"vitess.io/vitess/go/vt/vtgate/engine"
 )
 
+// typer is responsible for setting the type for expressions
+// it does it's work after visiting the children (up), since the children types is often needed to type a node.
 type typer struct {
```
harshit-gangal: vtgate engine should have a dependency on semantics. But we can change this if we get into circular dependency.

harshit-gangal (approved):

Issue #8571, opened 2021-07-30
Feature Request: Allowing the Option for CreateLookUpVindex to Continue Vreplication Stream after Backfill With Owner


michaelhzh:
Feature Description
CreateLookUpVindex uses Vreplication stream to backfill the newly created Vindex. When an owner is provided, the Vindex will stop after the backfill completes. CreateLookUpVindex can also be used to create Vindex for a table migrating from unsharded keyspace to a sharded one with the Movetable operation. This can reduce the need for tables from unsharded keyspace to the need to go through an intermediary single sharded keyspace in order to be moved to a multi-shard keyspace, which can greatly reduce the time and operation needed. However, this will mean that the Vreplication stream from Vindex will need to continue until rerouting is finished. However, this is not possible with an owner, and it is not realistic to initially not provide an owner and later add one in for the vindex. Therefore, it will be best to have the option to continue the Vreplication stream even with an owner and then later manually delete it from _vt.vreplication table. Therefore, I hope to include a feature that can force Vreplication stream to continue even with an owner.  

Use Case(s)
This feature can be used during move table operations from unsharded keyspace to sharded keyspace in order to reduce the operation and time cost.

deepthi:
Fixed in #8572

PR #8572, opened 2021-07-30
Customized CreateLookUpVindex to Include a Flag to Continue Vreplication After Backfill with an Owner Provided

michaelhzh:
## Description

This is PR for issue  #8571 , which is to allow forces vreplication to continue after backfill with an owner being provided.

## Related Issue(s)
#8571 

## Checklist


rohit-nayak-ps (approved):
lgtm

PR #8583, opened 2021-08-03
srvtopo: resilient watcher

vmg:
## Description

Making progress on https://github.com/vitessio/vitess/issues/8462

After several meetings with the team, we've decided that the only reliable way to detect is whether a failover or a resharding operation is over is by watching _both_ the `SrvKeyspace` and `SrvVSchema` changes in real time.

The current `ResilientServer` design does not allow for this, however.

- `SrvKeyspace` can only be accessed with a getter (i.e. there is no way to subscribe to changes in the keyspace), even though this functionality is actually watching the topology server in its underlying implementation.
- `SrvVSchema` can be subscribed for event changes, however, this is very expensive because the `Watch` call will open a watcher to the underlying topo server for each call, and we already have a couple of those calls in the codebase.

This refactoring fixes these two issues by merging the two implementations into a single `ResilientWatcher` implementation that:

- Keeps an underlying watch open to the topology server, and retries the watch whenever it fails
- Allows accessing the latest value of the watched event with a getter, like the previous implementation
- Allows subscribing to any changes on the watched event in real time using a callback

Although the implementations have been merged, I've taken special care to ensure the functionality remains identical to the previous versions. The unit tests have not been changed.

cc @harshit-gangal @deepthi @rohit-nayak-ps 

## Checklist

### DIFF
--- a/go/vt/srvtopo/query.go
+++ b/go/vt/srvtopo/query.go
@@ -1,3 +1,19 @@
+/*
+Copyright 2021 The Vitess Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
 package srvtopo
 
 import (
@@ -31,9 +47,9 @@ type queryEntry struct {
 type resilientQuery struct {
 	query func(ctx context.Context, entry *queryEntry) (interface{}, error)
 
-	counts       *stats.CountersWithSingleLabel
-	cacheRefresh time.Duration
-	cacheTTL     time.Duration
+	counts               *stats.CountersWithSingleLabel
+	cacheRefreshInterval time.Duration
+	cacheTTL             time.Duration
 
 	mutex   sync.Mutex
 	entries map[string]*queryEntry
@@ -65,9 +81,9 @@ func (q *resilientQuery) getCurrentValue(ctx context.Context, wkey fmt.Stringer,
 	cacheValid := entry.value != nil && (time.Since(entry.insertionTime) < q.cacheTTL)
 	if !cacheValid && staleOK {
 		// Only allow stale results for a bounded period
-		cacheValid = entry.value != nil && (time.Since(entry.insertionTime) < (q.cacheTTL + 2*q.cacheRefresh))
+		cacheValid = entry.value != nil && (time.Since(entry.insertionTime) < (q.cacheTTL + 2*q.cacheRefreshInterval))
 	}
-	shouldRefresh := time.Since(entry.lastQueryTime) > q.cacheRefresh
+	shouldRefresh := time.Since(entry.lastQueryTime) > q.cacheRefreshInterval
 
 	// If it is not time to check again, then return either the cached
 	// value or the cached error but don't ask topo again.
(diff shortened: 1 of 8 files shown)


deepthi (commented):
All new files need a copyright notice with the current year.

go/vt/srvtopo/query.go
```diff
@@ -0,0 +1,162 @@
...
+	value         interface{}
+	lastError     error
+	lastErrorCtx  context.Context
+}
+
+type resilientQuery struct {
+	query func(ctx context.Context, entry *queryEntry) (interface{}, error)
+
+	counts       *stats.CountersWithSingleLabel
+	cacheRefresh time.Duration
```
deepthi, line 35: cacheRefreshInterval? or duration?

go/vt/srvtopo/query.go
```diff
@@ -0,0 +1,162 @@
...
+	// value or the cached error but don't ask topo again.
+	if !shouldRefresh {
+		if cacheValid {
+			return entry.value, nil
+		}
+		return nil, entry.lastError
+	}
+
+	// Refresh the state in a background goroutine if no refresh is already
+	// in progress none is already running. This way queries are not blocked
```
deepthi, line 82:
```suggestion
	// Refresh the state in a background goroutine if no refresh is already
	// in progress. This way queries are not blocked
```

go/vt/srvtopo/query.go
```diff
@@ -0,0 +1,162 @@
...
+	// for the refresh attempt to complete to get a more up to date
+	// response.
+	//
+	// In the event that the topo service is slow or unresponsive either
+	// on the initial fetch or if the cache TTL expires, then several
+	// requests could be blocked on refreshingCond waiting for the response
+	// to come back.
+	if cacheValid {
+		return entry.value, nil
+	}
```
deepthi, line 145: This could very well be done before the check for refreshingChan.
Also refreshingCond -> refreshingChan in the comment.

go/vt/srvtopo/query.go
```diff
@@ -0,0 +1,162 @@
...
+
+	refreshingChan := entry.refreshingChan
+	entry.mutex.Unlock()
+	select {
+	case <-refreshingChan:
+	case <-ctx.Done():
+		entry.mutex.Lock()
+		return nil, fmt.Errorf("timed out waiting for keyspace names")
```
deepthi, line 153: This error message probably needs to be more generic, and refer to `wkey`.

go/vt/srvtopo/resilient_server_test.go
```diff
@@ -404,7 +403,7 @@ func TestGetSrvKeyspaceCreated(t *testing.T) {
 }
 
 func TestWatchSrvVSchema(t *testing.T) {
-	watchSrvVSchemaSleepTime = 10 * time.Millisecond
```
deepthi, line 407: I see this wasn't being used anywhere (and the test was using 10ms explicitly everywhere).
Do you feel like replacing 10ms with this var and 5ms with another one (or this one / 2)?
While you are at it, it may also be worth replacing `t.Fatalf` with `require`.

deepthi (commented):
Mostly looks fine to me.
Do you think we have sufficient test coverage? Are more tests needed?

go/vt/srvtopo/query.go
vmg: No, this is properly placed here. If `shouldRefresh` is true, we need to launch the background refresh goroutine _even if the cached value is freshe enough to return_, so this check needs to happen after.

vmg:
OK I've expanded the tests for keyspace event watching. I'd like to merge this tomorrow and get started with the changes to VTTablet error returns.

go/vt/srvtopo/query.go
deepthi: ah, makes sense.

deepthi (approved):
Nice work! +1 for adding new tests.

PR #8586, opened 2021-08-04
Trivial:  fix the GenerateShardRanges vtctl[client] help

aquarapid:
This was incorrect;  if you just pass a number (without `-num_shards`), you get the default 2 shard ranges printed.


rohit-nayak-ps (approved):
lgtm

PR #8588, opened 2021-08-04
Add additional options for configuring SSL modes as a client

dbussink:
In the case of for example vttablets connecting to an external MySQL, the current TLS / SSL options don't provide the sometimes needed flexibility.

The only way to provide any option(s) is through the `db_flags` and setting that to the magic value if 2048 (the `CapabilityClientSSL` bit). In this mode, it immediately moves to the strictest mode possible.

These changes make options available for Vitess users to more granularly configure the SSL settings. It mimics the MySQL client with an SSL mode flag that can be set to various values matching MySQL.

This flags replaces the magic constant for `db_flags` and allows more fine grained control.

## Checklist


go/mysql/server_test.go
```diff
@@ -917,22 +912,19 @@ func TestTLSRequired(t *testing.T) {
 	// Below, we are enabling --ssl-verify-server-cert, which adds
 	// a check that the common name of the certificate matches the
 	// server host name we connect to.
-	l, err := NewListener("tcp", ":0", authServer, th, 0, 0, false)
+	l, err := NewListener("tcp", "127.0.0.1:", authServer, th, 0, 0, false)
```
dbussink, line 915: I've updated quite few tests to this style. Binding on the local ip means we're less environment dependent and it removes the need further down to use the hostname. We use the option to pass in a separate `ServerName` so we still have TLS validation working, but don't depend as much on environment things that could be more brittle.

go/mysql/conn_params.go
```diff
@@ -28,15 +30,19 @@ type ConnParams struct {
...
+	// The following SSL flags control the SSL behavior.
+	//
+	// Not setting this value implies preferred mode unless
+	// the CapabilityClientSSL bit is set in db_flags. In the
+	// flag is set, it ends up equivalent to verify_identity mode.
+	SslMode          vttls.SslMode `json:"ssl_mode"`
+	SslCa            string        `json:"ssl_ca"`
+	SslCaPath        string        `json:"ssl_ca_path"`
+	SslCert          string        `json:"ssl_cert"`
+	SslKey           string        `json:"ssl_key"`
```
fatih, line 42: Nit: what do you think if we make the `Ssl` prefix all uppercase `SSL`?

dbussink: @fatih I think in Go only TLS gets capitalized like that by default. So if I'd update these, the linter will complain about them which is why I kept them as is (and we use this capitalization in other places too).

fatih: Gotcha, that makes sense 👍🏼

go/mysql/conn_params_test.go
```diff
@@ -0,0 +1,150 @@
...
+	assert.True(p.SslRequired())
+}
+
+func TestConnParams_SslRequiredVerifyIdentity(t *testing.T) {
+	p := ConnParams{
+		SslMode: vttls.VerifyIdentity,
+	}
+	assert := assert.New(t)
+	assert.True(p.SslRequired())
+}
```
vmg, line 150: This whole file could be a textbook example for Golang table-driven tests, but since you already took the effort to write individual tests for everything I wouldn't bother refactoring it. :)

vmg (approved):

go/mysql/conn_params_test.go
deepthi: It's a trade-off. table-driven tests are compact code-wise and easier to maintain, but harder to debug :)

go/vt/dbconfigs/dbconfigs.go
```diff
@@ -152,6 +156,7 @@ func registerPerUserFlags(userKey string, uc *UserConfig, cp *mysql.ConnParams)
 	flag.StringVar(&cp.UnixSocket, "db-config-"+userKey+"-unixsocket", "", "deprecated: use db_socket")
 	flag.StringVar(&cp.Charset, "db-config-"+userKey+"-charset", "utf8", "deprecated: use db_charset")
 	flag.Uint64Var(&cp.Flags, "db-config-"+userKey+"-flags", 0, "deprecated: use db_flags")
+	flag.Var(&cp.SslMode, "db-config-"+userKey+"-ssl-mode", "deprecated: use db_ssl_mode")
```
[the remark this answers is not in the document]
dbussink: @deepthi I can also drop it, but I kept it for consistency. If it's preferred, I can remove this specific one :smile:.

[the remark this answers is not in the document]
dbussink: I think this implies a potentially bigger thing though maybe for a separate PR, which is that the `UseSSL` flag for a specific user config might also need to be extended? 

And I think that then the client certificate / private key flags that exist in the global config now probably more belong to the specific user config? 

That all seems like a bigger / separate change though then?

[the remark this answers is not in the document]
dbussink: I removed this legacy flag, since there is I think a question around more specific per user SSL config settings, but that can be separate from these changes.

harshit-gangal (approved):

PR #8590, opened 2021-08-04
Fix passing the wrong cell/cells variable to CreateLookupVindex

aquarapid:
Found this by inspection after a user reported an issue.  We are using 
the cells variable as the new usage;  but we were still passing the old
cell variable to the lower code.  As a result, if you pass the `-cells`
directive to CreateLookupVindex, it was effectively ignored. 


rohit-nayak-ps (approved):
lgtm

PR #8598, opened 2021-08-05
VStreamer Field/Row Events: add Keyspace/Shard

rohit-nayak-ps:
## Description

When DDLs are run on a sharded keyspace it is possible that one shard has run a DDL before the other. Thus the schema and hence FieldEvents from each shard might be different. So when a RowEvent is seen it is currently ambiguous which is the corresponding FieldEvent. 

This PR adds both Keyspace and Shard so that consumers of the VStream API can map the RowEvent to the correct FieldEvent.

## Checklist


shlomi-noach (approved):
Adding keyspace/shard looks good. What I'm not sure is where it is being read & used? Is this PR just a partial preface to a later functionality?

rohit-nayak-ps:
> Adding keyspace/shard looks good. What I'm not sure is where it is being read & used? Is this PR just a partial preface to a later functionality?

You are right, this PR doesn't solve anything by itself. It is for consumers of the VStream API. Since DDLs can occur independently on shards it is possible that, for a short duration, the field events from each shard describe different schemas. So if the user just saves the last FieldEvent for a table and tries to map a future RowEvent to it, the columns may not match. Adding keyspace/shard allows users to map row events correctly to the corresponding field event of the right keyspace/shard.

##### Related work
We have also talked internally about detecting this interval within the vstream manager itself (as a plugin option, say). Then (optionally) if vstream detects the inconsistent schema between shards, it assumes a online ddl is in progress and waits for all shards to cutover. This is a larger project though. Also we may be able to build on top of the other work being done of buffering queries during a reshard cutover.

shlomi-noach:
Makes sense. Thanks for explaining!

deepthi:
Do website docs need to be updated?

PR #8603, opened 2021-08-09
OnlineDDL: better scheduling/cancellation logic

shlomi-noach:
## Description

This PR improves the logic around tracking running migrations and cancelling of stale/broken migrations.

Previously, we tracked:
- last `gh-ost` executed migration, and
- last `pt-osc` executed migration, and
- last `vrepl` executed migration
This led to some spaghetti code because `vrepl` doesn't behave the same way as `gh-ost` and `pt-osc`: it doesn't start and end in the same function. It actually doesn't have to start and end in the same tablet. 

This PR introduces `ownedRunningMigrations`: a map whose keys are UUID, which indicate "what migrations this executor expects to be running right now and is happy to own", irrespective of the strategy. In the new logic:

- A migration that is found to be running, and is not expected to be running, is auto-terminated
- executor is good to "adopt" a `vrepl` that is found to be running (this is in preparation for better handling of PRS/ERS)

This PR also produces better error messages around cancelled migrations. What used to be `"auto cancel"` is now a reasoned justified report.

We also cleanup some boiletplate code: previously `CancelMigration` accepted a `bool` flag indicating whether a running migration should be force-terminated. We find that we always pass `true` for this flag, so we remove the flag and just always terminate the migration if found to be running.

Existing test suite covers the logic _reasonably_ well. Last week we had a production issue where a couple migrations were `auto cancel`led without good reason. At the very least this PR will give us more information, but I believe it will also solve the issue. I'm not sure how to reproduce that production issue, so this is just to confess some cases must elude the existing tests.

## Related Issue(s)

#6926 

## Checklist


shlomi-noach:
As it turns out, this improvement to the scheduler also unlocked the would-be next step: automatic recoveries for NativeDDL. 

This seems to work now! You may begin a NativeDDL (OnlineDDL/VReplication) miration, promote a new primary halfway through, and migration will auto-recover, resume and _complete_ on the new primary.

`endtoend` tests needed.

deepthi (commented):
Nice work! I can approve once the endtoend test has been added.

shlomi-noach:
PRS tests are WIP. Meanwhile turning this PR into draft

shlomi-noach:
Added `endtoend` test to confirm vrepl migration survives PRS. The test works as follows:

- force throttling
- kick vrepl migration
- verify it is throttled
- verify vreplication is in `Copying` state (it's not doing anything because it's throttled)
- PRS on one shard
- unthrottle, expect migrations to come through to completion
- verify that on the PRS-d shard, the migration did indeed complete on the newly promoted tablet
- verify that on non-PRS-d shard, everything just executed as normal
- do the same, again, reinstating the original primary for consistency

This... works!

Ready for review

go/vt/vttablet/onlineddl/executor.go
```diff
@@ -1329,7 +1336,12 @@ func (e *Executor) readPendingMigrationsUUIDs(ctx context.Context) (uuids []stri
 }
 
 // terminateMigration attempts to interrupt and hard-stop a running migration
-func (e *Executor) terminateMigration(ctx context.Context, onlineDDL *schema.OnlineDDL, lastMigrationUUID string) (foundRunning bool, err error) {
+func (e *Executor) terminateMigration(ctx context.Context, onlineDDL *schema.OnlineDDL) (foundRunning bool, err error) {
+	// It's possible the killing the migratoin fails for whatever reason, in which case
```
deepthi, line 1340:
```suggestion
	// It's possible the killing the migration fails for whatever reason, in which case
```

deepthi (approved):
All I found is a typo in a comment. Otherwise LGTM

go/vt/vttablet/onlineddl/executor.go

PR #8608, opened 2021-08-09
Add `-full_flush` parameter for vtbackup, which performs

aquarapid:
a clean MySQL shutdown & startup cycle to work around xtrabackup
DDL bugs (see https://jira.percona.com/browse/PXB-2205).  xtrabackup
`-lock-ddl` (which is the default) is not sufficient to fix this, and we
have run into this in real-world scenarios.  See #8610 

### DIFF
--- a/go/cmd/vtbackup/vtbackup.go
+++ b/go/cmd/vtbackup/vtbackup.go
@@ -113,6 +113,7 @@ var (
 
 	initialBackup    = flag.Bool("initial_backup", false, "Instead of restoring from backup, initialize an empty database with the provided init_db_sql_file and upload a backup of that for the shard, if the shard has no backups yet. This can be used to seed a brand new shard with an initial, empty backup. If any backups already exist for the shard, this will be considered a successful no-op. This can only be done before the shard exists in topology (i.e. before any tablets are deployed).")
 	allowFirstBackup = flag.Bool("allow_first_backup", false, "Allow this job to take the first backup of an existing shard.")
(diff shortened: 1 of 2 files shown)


deepthi:
vtbackup can be run with either `builtin` or `xtrabackup` backup engine. A better place for this flag and functionality is in the xtrabackup backup engine code. That gives us two benefits:
- This functionality will become available to all uses of xtrabackup, not just via vtbackup.
- Someone using builtin backup method won't be able to make the mistake of stopping and restarting mysql twice using this flag.

aquarapid:
Not sure I agree with this;  limiting this to vtbackup only avoids someone doing the wrong thing when running xtrabackup against a live instance, and killing MySQL

deepthi:
> Not sure I agree with this; limiting this to vtbackup only avoids someone doing the wrong thing when running xtrabackup against a live instance, and killing MySQL

They would have to start vttablets with the flag for that to happen. We would still default it to false.

deepthi (commented):
Based on an offline discussion, I'm ok with adding this flag just to vtbackup and not vttablet.
The reasoning is that for anyone using vitess-operator, there's currently no way to pass differing xtrabackup flags through vtbackup than the standard vttablet ones;  so you wouldn't be able to vary this flag depending on whether you take an offline (vtbackup) backup or an online one (vttablet). We can revisit a vttablet flag when someone is able to make that change in concert with a vitess-operator change.

@aquarapid Can you rename the flag to reflect that we are doing a stop/start and not actually executing any `FLUSH` statements?

deepthi (approved):
LGTM

PR #8611, opened 2021-08-09
fix copyright

deepthi:
## Description
A few files still had the old copyright from pre-CNCF days.

## Related Issue(s)
None

## Checklist
- [x] Should this PR be backported? Yes, only if a patch release is planned.


rohit-nayak-ps (approved):
lgtm

Issue #8612, opened 2021-08-09
Vreplication fails when gtid_executed is > 10000 bytes


gtowey:
#### Overview of the Issue

Vreplication fails with `unable to parse mysql 5.6 GTID set` when the length of the gtid_executed string exceeds the column size of 10,000 bytes.

__Note__: This is a pretty simple fix and I can submit a PR along with this issue. 

#### Reproduction Steps

1. Create a large gtid_executed set by performing many replication changes across many mysql instances with different server UUIDs
1. Run a `Reshard` workflow
1. View the error in vreplication on the chosen tablet's status page.

#### Operating system and Environment details

n/a

shlomi-noach:
@gtowey I assume the problem is here:

https://github.com/vitessio/vitess/blob/600cca789bad1fb5f312f4666b1678ba926608b6/go/vt/binlog/binlogplayer/binlog_player.go#L544-L545

?

If you are able to, we welcome a PR to increase this length.

mattlord:
I'm going to close this as done for now. Please note that we've since added compression for this value as well: https://github.com/vitessio/vitess/pull/7877

PR #6421, opened 2020-07-08
VReplication: _vt.vreplication source column VARBINARY->BLOB

teejae:
Signed-off-by: Toliver Jue <toliver@planetscale.com>

Change the `source` column to have unlimited size (4gb), to allow for larger vreplication source content, such as many tables.
- Should cause no issues with current code.
- Ideally has no performance impact for same-size `source` column data.

PR #8613, opened 2021-08-09
Increase replication pos column size

gtowey:
## Description

This change fixes errors using vreplication when the gtid_executed set is very large.  

This can happen as part of normal MySQL server operation if there simply happens to be many changes in primary instances in a cluster.  There doesn't seem to be any defined maximum size on the string values returned so it's critical that larger values are handled.  The column type has been changed from varbinary to blob because larger varbinary values cause an error when the maximum row size for the table exceeds 64k

## Related Issue(s)
https://github.com/vitessio/vitess/issues/8612

## Checklist

### DIFF
--- a/go/vt/binlog/binlogplayer/binlog_player.go
+++ b/go/vt/binlog/binlogplayer/binlog_player.go
@@ -541,7 +541,7 @@ func CreateVReplicationTable() []string {
   id INT AUTO_INCREMENT,
   workflow VARBINARY(1000),
   source VARBINARY(10000) NOT NULL,
-  pos VARBINARY(10000) NOT NULL,
+  pos BLOB NOT NULL,
   stop_pos VARBINARY(10000) DEFAULT NULL,
   max_tps BIGINT(20) NOT NULL,
   max_replication_lag BIGINT(20) NOT NULL,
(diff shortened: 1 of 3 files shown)


go/vt/binlog/binlogplayer/binlog_player.go
```diff
@@ -541,7 +541,7 @@ func CreateVReplicationTable() []string {
   id INT AUTO_INCREMENT,
   workflow VARBINARY(1000),
   source VARBINARY(10000) NOT NULL,
-  pos VARBINARY(10000) NOT NULL,
+  pos BLOB NOT NULL,
```
deepthi, line 544: In order for in-place upgrades to also get the new column definition, we need to add an `ALTER TABLE ... MODIFY ...`
(lines 561-566).
I see we did that for `source` but not sure why the definition here still says `VARBINARY`.

deepthi:
Can you fix the DCO? Once this is ready for review we can run CI.

rohit-nayak-ps (changes requested):
Thanks for the PR. CI actions have been enabled.

As @deepthi noted, we should use the ALTER approach so that existing installations are also upgraded with this change.

You should add to the alter DDL list at `go/vt/binlog/binlogplayer/binlog_player.go:561` and also to the test ddls list at `go/vt/vttablet/tabletmanager/vreplication/engine_test.go:498`. 

This PR does a similar thing: https://github.com/vitessio/vitess/pull/6421

We also have a feature to compress the pos before storing: https://github.com/vitessio/vitess/pull/7877. Since you have large GTIDs you may want encounter similar issues described here.

mattlord:
I'm closing this for now as it appears to have gone stale and I believe the underlying issue has been addressed.

Please let me know if I missed or misunderstood something and we can re-open it. Thanks!

Issue #8585, opened 2021-08-03
Trim cached schema version for schema tracking


5antelope:
#### Feature Description
Based on my understanding, when we enable the schema tracking feature, the historian store all the DDL for _all_ the tables in this shard into `_vt.schema_version` table (it seems like we are storing the final `CREATE TABLE` statement instead of the `ALTER TABLE`, but I think they both apply to the issue here)

When there is a DDL, historian will load top 10K schemas from `_vt.schema_version` table and keep appending to `schemas` slice: https://github.com/vitessio/vitess/blob/cb054384781ce1b89b49e68c29d23bfb0d8a190e/go/vt/vttablet/tabletserver/schema/historian.go#L167-L181

Now image a shard has multiple schema migration, the schema slice can easily fill in all the memories. There seems to be 2 optimization we can do:
1. instead of blindly append into the slice, we can do some deduping to same space
2. we don't really need _all_ the schemas, we just need the schema that is as old as `gtid_purged` - anything older than that is no longer available in binlog already

sonne5:
cc @aquarapid this is the issue we discussed today, please correct me if I missed something and this turns out to be a red herring. thanks!

aquarapid:
We actually only store the DDL for the schema update in the `schema_version.ddl` field, so from a memory point of view that is probably not a big deal.  However, we do store the complete schema in protobuf format in the schema_version.schemax` field.   I don't think this is a huge deal in most situations, unless your schema is truly large (many thousands of tables).

However, your point that this table can be cleaned of schemas older that gtid_purged is well taken;  doing this might be somewhat complicated;  I guess as a stopgap;  a user can put an external process/script in place to purge the file of all rows older than X days, where X is well above their binlog retention period?

sonne5:
thanks @aquarapid. Yeah, `schemax` is basically the db schema. 

re: memory usage: putting `gtid_purged` aside, the memory usage also depends on the frequency of schema migration. On every DDL, the current logic will load 10K row and append. this means we will have a lot of duplicated schema in the `schema` slice over the time (e.g., if I do 5 DDL, the slice will keep 1 + 2 + 3 + 4 + 5 = 15 items instead of 5). This might make the search by gtid/position in the slice slower

aquarapid:
Point taken;  but you would also need to deal with subtractive DDL;  i.e. how do you represent a drop table in the `schemax`?  So you would need to implement a diffing setup.

sonne5:
[...]

PR #8617, opened 2021-08-10
historian - keep N versions in memory

indera-shsp:
Signed-off-by: Andrei Sura <andrei.sura@sharpspring.com>

## Description

When the number of rows in `schema_version` table is very large and keeps growing constantly due cronjobs dropping and re-creating tables we observed increase in memory usage.

It is not guaranteed this is the only issue but I thought this PR can help starting a discussion.

Here is what vttablet RAM looks like for ~1500 schema changes (we have more than 300 tables so every row has a large `schemax` column) :
Screen Shot 2021-08-16 at 11 07 09 AM
 

```
desc schema_version;
+--------------+------------------+------+-----+---------+----------------+
| Field        | Type             | Null | Key | Default | Extra          |
+--------------+------------------+------+-----+---------+----------------+
| id           | int(11)          | NO   | PRI | NULL    | auto_increment |
| pos          | varbinary(10000) | NO   |     | NULL    |                |
| time_updated | bigint(20)       | NO   |     | NULL    |                |
| ddl          | blob             | NO   |     | NULL    |                |
| schemax      | longblob         | NO   |     | NULL    |                |
+--------------+------------------+------+-----+---------+----------------+

-- query used to count the version rows

select count(*) from _vt.schema_version;
+----------+
| count(*) |
+----------+
|     2117 |
+----------+

select max(length(schemax)) from schema_version;
+----------------------+
| max(length(schemax)) |
+----------------------+
|               366580 |
+----------------------+

Total MB loaded in RAM
select 2117 * 366580/ (1024 * 1024);
+------------------------------+
| 2117 * 366580/ (1024 * 1024) |
+------------------------------+
|                     740.0988 |
+------------------------------+
```

I did some benchmarking using a mock of the code and measured the memory usage of the two blocks of code, and the numbers look promising to me
```
== old code

Memory usage in loadFromDB: {
allocated:1226 MiB 
totalAllocated:%!f(uint64=1857) MiB 
system:%!f(uint64=1358) MiB
heapAlloc:%!f(uint64=1226) MiB}

== new code	
Memory usage in loadFromDB: {
allocated:745 MiB
totalAllocated:%!f(uint64=1891) MiB
system:%!f(uint64=1153) MiB 
heapAlloc:%!f(uint64=745) MiB}
```

## Checklist

### DIFF
--- a/go/vt/vttablet/tabletserver/schema/historian.go
+++ b/go/vt/vttablet/tabletserver/schema/historian.go
@@ -35,6 +35,9 @@ import (
 	"vitess.io/vitess/go/vt/sqlparser"
 )
 
+// maxVersions - how many entries to keep in memory
+// TODO: should this be exposed as a startup flag?
+const maxVersions = 1000
 const getSchemaVersions = "select id, pos, ddl, time_updated, schemax from _vt.schema_version where id > %d order by id asc"
 
 // vl defines the glog verbosity level for the package
@@ -164,11 +167,12 @@ func (h *historian) loadFromDB(ctx context.Context) error {
 		return err
 	}
 	defer conn.Recycle()
-	tableData, err := conn.Exec(ctx, fmt.Sprintf(getSchemaVersions, h.lastID), 10000, true)
+	tableData, err := conn.Exec(ctx, fmt.Sprintf(getSchemaVersions, h.lastID), maxVersions, true)
 	if err != nil {
 		log.Infof("Error reading schema_tracking table %v, will operate with the latest available schema", err)
 		return nil
 	}
+
 	for _, row := range tableData.Rows {
 		trackedSchema, id, err := h.readRow(row)
 		if err != nil {
@@ -178,6 +182,19 @@ func (h *historian) loadFromDB(ctx context.Context) error {
 		h.lastID = id
 	}
 	h.sortSchemas()
+
+	loadedSchemas := len(h.schemas)
+	if loadedSchemas > maxVersions {
+		cutOff := loadedSchemas - maxVersions
+		for i := 0; i < cutOff; i++ {
+			// discard references to other memory
+			h.schemas[i] = nil
+		}
+
+		// keep only last N versions
+		h.schemas = h.schemas[cutOff:loadedSchemas]
+	}
+
 	return nil
 }
 


go/vt/vttablet/tabletserver/schema/historian.go
```diff
@@ -178,6 +182,13 @@ func (h *historian) loadFromDB(ctx context.Context) error {
 		h.lastID = id
 	}
 	h.sortSchemas()
+
+	loadedSchemas := len(h.schemas)
```
indera-shsp, line 186: loadFromDB is called every time schema changes and the number of entries in `h.schemas` can grow pretty fast

go/vt/vttablet/tabletserver/schema/historian.go
```diff
@@ -178,6 +182,13 @@ func (h *historian) loadFromDB(ctx context.Context) error {
...
+	if loadedSchemas > maxVersions {
+		// keep only last N versions
+		h.schemas = h.schemas[loadedSchemas-maxVersions : loadedSchemas]
```
indera-shsp, line 189: @vmg If you have time to read this... can this cause a memory leak?

derekperkins: I'm pretty sure that isn't saving memory. It's still a slice on the same backing array and this is just moving the window.

indera-shsp: From the article linked the fix is to
```
 is to nil out the trailing pointer(s) before you truncate the slice.
```

indera-shsp: @derekperkins Is the new version acceptable?

indera-shsp:
@deepthi Can you please recommend somebody to review this PR? 

Thank you!

indera-shsp:
I wrote this to emulate the fix

```
package main

import (
	"fmt"
	"sync"
)

func main() {
	print("\n==> Old code")
	testHistorianOrig()
	print("\n==> New code")
	testHistorian()
}

type trackedSchema struct {
	//schema map[string]*binlogdatapb.MinimalTable
	//schema map[string]*string
	ddl    string
}

type historian struct {
	schemas []*trackedSchema
	mu      sync.Mutex
	enabled bool
	isOpen  bool
}

func newTrackedSchema(ddl string) *trackedSchema {
	r := &trackedSchema{
		ddl: ddl,
	}
	return r
}

func newHistorian(enabled bool) *historian {
	sh := historian{
		enabled: enabled,
	}
	return &sh
}


func (h *historian) loadFromDbOriginal(round int) error {
	loadVersions := 3

	for i := 1; i <= loadVersions; i++ {
		// strconv.Itoa(i)
		ddl := fmt.Sprintf("create tbl x_%d_%d", round, i)
		h.schemas = append(h.schemas, newTrackedSchema(ddl))
	}

	return nil
}

func (h *historian) loadFromDB(round int) error {
	loadVersions := 3

	for i := 1; i <= loadVersions; i++ {
		ddl := fmt.Sprintf("create tbl x_%d_%d", round, i)
		h.schemas = append(h.schemas, newTrackedSchema(ddl))
	}

	// fix
	maxVersions := loadVersions + 2
	loadedSchemas := len(h.schemas)
	if loadedSchemas > maxVersions {
		cutOff := loadedSchemas-maxVersions
		for i := 0; i < cutOff; i++ {
			h.schemas[i] = nil
		}
		h.schemas = h.schemas[cutOff : loadedSchemas]
	}

	return nil
}

func (h *historian) printMe() {
	loadedSchemas := len(h.schemas)

	str := fmt.Sprintf("\nHistorian - loadedSchemas: %d", loadedSchemas)
	str += fmt.Sprintf("\n\t %s", h.schemas)
	print(str)
}

func testHistorianOrig() {
	h := newHistorian(true)
	h.loadFromDbOriginal(1)
	h.printMe()

	h.loadFromDbOriginal(2)
	h.printMe()

	h.loadFromDbOriginal(3)
	h.printMe()
}

func testHistorian() {
	h := newHistorian(true)
	h.loadFromDB(1)
	h.printMe()

	h.loadFromDB(2)
	h.printMe()

	h.loadFromDB(3)
	h.printMe()
}

```

sonne5:
Having `maxVersions` requires ppl know how many schema migrations they want for all the binlogs, otherwise we might remove the an old schema so is required for vstream's schema tracking. 

We are having similar discussions, and I think a more ergonomic fix should be just load schema version that is enough to cover all the binlogs https://github.com/vitessio/vitess/issues/8585

indera-shsp:
> Having `maxVersions` requires ppl know how many schema migrations they want for all the binlogs, otherwise we might remove the an old schema so is required for vstream's schema tracking.
> 
> We are having similar discussions, and I think a more ergonomic fix should be just load schema version that is enough to cover all the binlogs #8585

@sonne5 
Does your comment assume that covering all the binlogs _always_ leads to a small and safe number of rows loaded to RAM?

sonne5:
> > Having `maxVersions` requires ppl know how many schema migrations they want for all the binlogs, otherwise we might remove the an old schema so is required for vstream's schema tracking.
> > We are having similar discussions, and I think a more ergonomic fix should be just load schema version that is enough to cover all the binlogs #8585
> 
> @sonne5
> Does your comment assume that covering all the binlogs _always_ leads to a small and safe number of rows loaded to RAM?

Whether if is small depends on how frequent the schema change is, but it should be safe (as in not break the existing functionalities)

indera-shsp:
> > > Having `maxVersions` requires ppl know how many schema migrations they want for all the binlogs, otherwise we might remove the an old schema so is required for vstream's schema tracking.
> > > We are having similar discussions, and I think a more ergonomic fix should be just load schema version that is enough to cover all the binlogs #8585
> > 
> > 
> > @sonne5
> > Does your comment assume that covering all the binlogs _always_ leads to a small and safe number of rows loaded to RAM?
> 
> I think whether if is small depends on how frequent the schema change is, but it should be safe (as in not break the existing functionalities)

It sounds like you are advocating for this to be configurable?

deepthi (commented):
Apologies that reviewing this took so long.
The code itself looks fine. You seem to have done a lot of due diligence on this PR, which is definitely appreciated.

I believe to address the concerns brought up so far:
- maxVersions should be configurable via a flag and should default to the current number of 10000
- a test case is required to ensure the flag is being respected. you can simply convert your standalone test into a unit test.

Auto-tuning this number to cover the binlog period as described in #8585 can be a future improvement.

PR #8623, opened 2021-08-11
Inclusive naming: more master-> primary changes

deepthi:
## Description
More changes to docs, config files, tests, error messages and miscellaneous functions and variables.

Approximately 29% reduction from the previous baseline
```
$ git checkout main
$ git grep -i master | wc -l
4900
$ git checkout ds-naming-m2p9
$ git grep -i master | wc -l
3494
```

## Related Issue(s)
#7114

## Checklist

## Deployment Notes
A lot of error message text has changed. Any tooling that depends on error messages returned from vtctl commands, or alerts based on specific error messages in logs may need to change to adapt.


deepthi:
@ajm188 @doeg Did I break vtadmin with the proto changes (BackupRequest and PopulateReparentJournalRequest). If yes, we need an integration test that catches such things.

rohit-nayak-ps (approved):
nice. lgtm. 

I assume the few references to `source` instead of `primary` is because they refer to mysql primary instances and we are following the mysql `source` nomenclature

deepthi:
> nice. lgtm.
> 
> I assume the few references to `source` instead of `primary` is because they refer to mysql primary instances and we are following the mysql `source` nomenclature

That is correct. It is possible that there are few places that are ambiguous and we went one way versus another. As long as we get it right in the RPCs (which are hard to change), other places can be changed again if someone feels strongly about particular functions or variables.

PR #8625, opened 2021-08-12
Fixing vtorc naming convention

### LINKED ISSUE #7110: Task: Cleanup references to master and slave in vt/orchestrator
There are several references in the code under `vt/orchestrator`. Some of them can probably be deleted outright because we don't have to worry about backwards-compatibility yet.

For example:
```
SlaveLagQuery                              string   // Synonym to ReplicationLagQuery
```
Variables, functions and parameters should be renamed.
Every occurrence should be evaluated to determine whether it can be removed. There might be exceptions like the exact text of MySQL error messages and low-level MySQL commands.

GuptaManan100:
## Description
Cleanup references to master and slave in vt/orchestrator

## Related Issue(s)
- Fixes #7110

## Checklist


go/vt/orchestrator/config/config_test.go
```diff
@@ -15,104 +15,52 @@ func init() {
...
-		c.ReplicationLagQuery = ""
-		err := c.postReadAdjustments()
-		test.S(t).ExpectNil(err)
-		test.S(t).ExpectEquals(c.ReplicationLagQuery, "select 3")
-	}
 }
 
 func TestPostponeReplicaRecoveryOnLagMinutes(t *testing.T) {
 	{
 		c := newConfiguration()
```
deepthi, line 26: I'm wondering if it's even worth keeping these tests. Their intent seems to be test copying of the value from the old param to the new one. Since we are deleting the old param there doesn't seem to be much left to test.

go/vt/orchestrator/app/cli.go
```diff
@@ -51,29 +51,10 @@ func (a stringSlice) Swap(i, j int)      { a[i], a[j] = a[j], a[i] }
...
-	"which-cluster-gh-ost-slaves": "which-cluster-gh-ost-replicas",
-	"which-slaves":                "which-replicas",
-	"detach-slave":                "detach-replica-master-host",
-	"detach-replica":              "detach-replica-master-host",
-	"detach-slave-master-host":    "detach-replica-master-host",
-	"reattach-slave":              "reattach-replica-master-host",
-	"reattach-replica":            "reattach-replica-master-host",
-	"reattach-slave-master-host":  "reattach-replica-master-host",
+	"enslave-siblings": "take-siblings",
+	"enslave-primary":  "take-primary",
```
deepthi, line 55: do we need these mappings, or can we remove the `enslave` versions altogether?

go/vt/orchestrator/http/api.go
```diff
@@ -51,26 +51,8 @@ const (
...
-	"reattach-slave-master-host": "reattach-replica-master-host",
-	"cluster-osc-slaves":         "cluster-osc-replicas",
-	"start-slave":                "start-replica",
-	"restart-slave":              "restart-replica",
-	"stop-slave":                 "stop-replica",
-	"stop-slave-nice":            "stop-replica-nice",
-	"reset-slave":                "reset-replica",
-	"restart-slave-statements":   "restart-replica-statements",
+	"enslave-siblings": "take-siblings",
+	"enslave-primary":  "take-primary",
```
deepthi, line 55: remove synonyms and only support new versions?

go/vt/orchestrator/inst/instance_dao.go
```diff
@@ -984,36 +984,36 @@ func readInstanceRow(m sqlutils.RowMap) *Instance {
...
 	instance.IsDetached, _ = instance.ExecBinlogCoordinates.ExtractDetachedCoordinates()
 	instance.RelaylogCoordinates.LogFile = m.GetString("relay_log_file")
 	instance.RelaylogCoordinates.LogPos = m.GetInt64("relay_log_pos")
 	instance.RelaylogCoordinates.Type = RelayLog
 	instance.LastSQLError = m.GetString("last_sql_error")
 	instance.LastIOError = m.GetString("last_io_error")
-	instance.SecondsBehindMaster = m.GetNullInt64("seconds_behind_master")
-	instance.ReplicationLagSeconds = m.GetNullInt64("slave_lag_seconds")
+	instance.SecondsBehindPrimary = m.GetNullInt64("seconds_behind_primary")
+	instance.ReplicationLagSeconds = m.GetNullInt64("replica_lag_seconds")
```
deepthi, line 998: Are these two independent values or basically the same thing?

deepthi (commented):
:100:  
Nice work! Just a few comments.

go/vt/orchestrator/config/config_test.go
shlomi-noach: agreed

go/vt/orchestrator/app/cli.go
shlomi-noach: those mappings were made for backwards-compatibility, which is a non-issue in vitess. These should be removed.

go/vt/orchestrator/http/api.go
shlomi-noach: agree

go/vt/orchestrator/inst/instance_dao.go
shlomi-noach: They are not. `seconds_behind_master` is output from `show slave status`, and is always there.
`ReplicationLagSeconds` is:
- computed by `ReplicationLagQuery` if present, or
- fallback to `seconds_behind_master`
it's good to have both values.

go/vt/orchestrator/db/generate_base.go
```diff
@@ -28,21 +28,21 @@ var generateSQLBase = []string{
...
-          slave_sql_running tinyint(3) unsigned NOT NULL,
-          slave_io_running tinyint(3) unsigned NOT NULL,
-          master_log_file varchar(128) CHARACTER SET ascii NOT NULL,
-          read_master_log_pos bigint(20) unsigned NOT NULL,
-          relay_master_log_file varchar(128) CHARACTER SET ascii NOT NULL,
-          exec_master_log_pos bigint(20) unsigned NOT NULL,
-          seconds_behind_master bigint(20) unsigned DEFAULT NULL,
-          slave_lag_seconds bigint(20) unsigned DEFAULT NULL,
-          num_slave_hosts int(10) unsigned NOT NULL,
-          slave_hosts text CHARACTER SET ascii NOT NULL,
```
shlomi-noach, line 45: These changes are a hard break: what happens to already existing `vtorc` setups? They will break and fail. Their table schema will not change, and suddenly the app expects different columns in the database.

Is there a migration path for those users? At the very least, they should be told they must delete their existing SQLite database. There should be docs explaining how to do so.

go/vt/orchestrator/db/generate_patches.go
```diff
@@ -549,27 +549,27 @@ var generateSQLPatches = []string{
 	`
 		ALTER TABLE
 			database_instance
-			ADD COLUMN semi_sync_master_timeout INT UNSIGNED NOT NULL DEFAULT 0 AFTER semi_sync_master_enabled
+			ADD COLUMN semi_sync_primary_timeout INT UNSIGNED NOT NULL DEFAULT 0 AFTER semi_sync_primary_enabled
 	`,
 	`
 		ALTER TABLE
 			database_instance
-			ADD COLUMN semi_sync_master_wait_for_slave_count INT UNSIGNED NOT NULL DEFAULT 0 AFTER semi_sync_master_timeout
```
shlomi-noach, line 557: these variable names map directly to the mysql global variables of same name, which is the reason why the master/slave terminology appears here in the first place. Changing these column names will break that implicit connection (which is fine, just pointing this out).

shlomi-noach (commented):
Concern over hard break of database schema for whatever existing setups (even if users are not actually _using_ vtorc, the schema existsm and it will not upgrade with these changes)

go/vt/orchestrator/inst/instance_dao.go
```diff
@@ -351,23 +351,23 @@ func ReadTopologyInstanceBufferable(instanceKey *InstanceKey, bufferWrites bool,
...
 				semiSyncReplicaPluginLoaded := false
 				err := sqlutils.QueryRowsMap(db, "show global variables like 'rpl_semi_sync_%'", func(m sqlutils.RowMap) error {
-					if m.GetString("Variable_name") == "rpl_semi_sync_master_enabled" {
-						instance.SemiSyncMasterEnabled = (m.GetString("Value") == "ON")
-						semiSyncMasterPluginLoaded = true
-					} else if m.GetString("Variable_name") == "rpl_semi_sync_master_timeout" {
-						instance.SemiSyncMasterTimeout = m.GetUint64("Value")
-					} else if m.GetString("Variable_name") == "rpl_semi_sync_master_wait_for_slave_count" {
-						instance.SemiSyncMasterWaitForReplicaCount = m.GetUint("Value")
+					if m.GetString("Variable_name") == "rpl_semi_sync_primary_enabled" {
```
shlomi-noach, line 357: :exclamation: :exclamation: :exclamation: This is wrong. MySQL's variables are unaffected by any changes to this PR. This code reads a MySQL variable, and its name is `rpl_semi_sync_master_enabled`

go/vt/orchestrator/inst/instance_dao.go
```diff
@@ -351,23 +351,23 @@ func ReadTopologyInstanceBufferable(instanceKey *InstanceKey, bufferWrites bool,
...
+						instance.SemiSyncPrimaryEnabled = (m.GetString("Value") == "ON")
+						semiSyncPrimaryPluginLoaded = true
+					} else if m.GetString("Variable_name") == "rpl_semi_sync_primary_timeout" {
```
shlomi-noach, line 360: wrong. MySQL variable name unaffected by this PR

go/vt/orchestrator/inst/instance_dao.go
```diff
@@ -351,23 +351,23 @@ func ReadTopologyInstanceBufferable(instanceKey *InstanceKey, bufferWrites bool,
...
+						instance.SemiSyncPrimaryTimeout = m.GetUint64("Value")
+					} else if m.GetString("Variable_name") == "rpl_semi_sync_primary_wait_for_replica_count" {
```
shlomi-noach, line 362: wrong. MySQL variable name unaffected by this PR

go/vt/orchestrator/inst/instance_dao.go
```diff
@@ -376,10 +376,10 @@ func ReadTopologyInstanceBufferable(instanceKey *InstanceKey, bufferWrites bool,
 			go func() {
 				defer waitGroup.Done()
 				err := sqlutils.QueryRowsMap(db, "show global status like 'rpl_semi_sync_%'", func(m sqlutils.RowMap) error {
-					if m.GetString("Variable_name") == "Rpl_semi_sync_master_status" {
-						instance.SemiSyncMasterStatus = (m.GetString("Value") == "ON")
-					} else if m.GetString("Variable_name") == "Rpl_semi_sync_master_clients" {
-						instance.SemiSyncMasterClients = m.GetUint("Value")
+					if m.GetString("Variable_name") == "Rpl_semi_sync_primary_status" {
```
shlomi-noach, line 379: wrong. MySQL variable name unaffected by this PR

go/vt/orchestrator/inst/instance_dao.go
```diff
@@ -376,10 +376,10 @@ func ReadTopologyInstanceBufferable(instanceKey *InstanceKey, bufferWrites bool,
...
+						instance.SemiSyncPrimaryStatus = (m.GetString("Value") == "ON")
```
shlomi-noach, line 380: wrong. MySQL variable name unaffected by this PR

go/vt/orchestrator/inst/instance_dao.go
```diff
@@ -376,10 +376,10 @@ func ReadTopologyInstanceBufferable(instanceKey *InstanceKey, bufferWrites bool,
...
+					} else if m.GetString("Variable_name") == "Rpl_semi_sync_primary_clients" {
```
shlomi-noach, line 381: wrong. MySQL variable name unaffected by this PR

go/vt/orchestrator/inst/instance_dao.go
```diff
@@ -812,21 +812,21 @@ func ReadReplicationGroupPrimary(instance *Instance) (err error) {
...
 
 	// Read the cluster_name of the _primary_ or _group_primary_ of our instance, derive it from there.
 	query := `
 			select
 					cluster_name,
 					suggested_cluster_alias,
 					replication_depth,
-					master_host,
-					master_port,
+					primary_host,
```
shlomi-noach, line 828: should we use `source_host`, `source_port` terminology? Unsure.

go/vt/orchestrator/inst/instance_dao.go
```diff
@@ -2274,57 +2274,57 @@ func mkInsertOdkuForInstances(instances []*Instance, instanceWasActuallyFound bo
...
+		"primary_port",
+		"replica_sql_running",
+		"replica_io_running",
 		"replication_sql_thread_state",
 		"replication_io_thread_state",
 		"has_replication_filters",
 		"supports_oracle_gtid",
 		"oracle_gtid",
-		"master_uuid",
+		"replica_uuid",
```
shlomi-noach, line 2289: :exclamation: :exclamation: :exclamation: this should be `primary_uuid` or `source_uuid`, not `replica_uuid`

go/vt/orchestrator/inst/instance_dao.go
```diff
@@ -2274,57 +2274,57 @@ func mkInsertOdkuForInstances(instances []*Instance, instanceWasActuallyFound bo
...
 		"mariadb_gtid",
 		"pseudo_gtid",
-		"master_log_file",
-		"read_master_log_pos",
-		"relay_master_log_file",
-		"exec_master_log_pos",
+		"primary_log_file",
+		"read_primary_log_pos",
+		"relay_primary_log_file",
+		"exec_primary_log_pos",
```
shlomi-noach, line 2300: should this use `source` terminology? unsure.

shlomi-noach (changes requested):
Found some alarming changes that would break behavior.

- Please consider that MySQL variable names still have `master` and `slave` terminology, we must still read those variables by their name.
- I found one place that renamed `master_uuid` to `replica_uuid`, which is wrong. I didn't follow the entire code graph to verify how this column is used.

I'm a bit intimidated by the amount of changes in this single PR. It's going to be difficult to ensure everything is done right, and the tests do not have good coverage for these changes.

go/vt/orchestrator/inst/instance_dao.go
GuptaManan100: changed

go/vt/orchestrator/inst/instance_dao.go
GuptaManan100: Fixed

go/vt/orchestrator/inst/instance_dao.go
GuptaManan100: Fixed

go/vt/orchestrator/inst/instance_dao.go
GuptaManan100: Fixed

go/vt/orchestrator/inst/instance_dao.go
GuptaManan100: Fixed

go/vt/orchestrator/inst/instance_dao.go
GuptaManan100: Fixed

go/vt/orchestrator/inst/instance_dao.go
GuptaManan100: Fixed

go/vt/orchestrator/db/generate_patches.go
GuptaManan100: Yes I saw that. I thought we can cleanup the nomenclature from our end, though this would break the one to one mapping between the same names. Should I change it back?

go/vt/orchestrator/db/generate_base.go
GuptaManan100: I agree. We have not done any general release so far, so I thought we could do these changes without worrying about backward compatibility. I can change them to `ALTER TABLE` statements instead? We will need to add a documentation page anyways, since some of the configurations and the API endpoints have changed too

go/vt/orchestrator/http/api.go
GuptaManan100: Changed

go/vt/orchestrator/config/config_test.go
GuptaManan100: Removed

go/vt/orchestrator/app/cli.go
GuptaManan100: Removed

GuptaManan100:
@shlomi-noach I tried to be careful as to not change any MySQL commands but I see that i did make a few mistakes. Would it be preferable if I split the PR into smaller components so that reviewing changes is easier?

go/vt/orchestrator/db/generate_base.go
shlomi-noach: >  I can change them to ALTER TABLE statements instead? 

FYI sqlite does not support `DROP COLUMN`. You can only add new columns

go/vt/orchestrator/db/generate_patches.go
shlomi-noach: this is fine, just pointing out

shlomi-noach:
> Would it be preferable if I split the PR into smaller components so that reviewing changes is easier and we don't accidentally end up introducing bugs?

I don't mean to deter you from this hard work; it does seem like test coverage has no chance of catching e.g. wrong variable names in mysql. Fortunately not many of these. I guess, the scope is very large and it's hard to verify everything is in order. 

I'd for example 1st only refactor golang names, which is a massive change but is safe if done with proper refactoring tool.

Entirely at your discretion.

GuptaManan100:
Yes that seems like a good idea. Closing this PR and will create smaller PRs as decided

go/vt/orchestrator/db/generate_base.go
deepthi: We will need to write the proper release notes to warn people, but because vtorc is not GA, we are free to make breaking changes.
That is one of the reasons to announce a feature as experimental, it frees us from having to carry-forward code for backwards compatibility.

Issue #8595, opened 2021-08-05
Provide an option to not migrate vstream on shard split


sonne5:
#### Feature Description
Right now VStream will be automatically migrated based on a journal event after a shard split. 

However, this is assume there is only one VStream client for all the shards. If we want to partition VStream and distribute the workload, we should just signal the VStream client that there is a shard split and pass along the journal event and some worker will pick up the new shard.

We need to pass along the journal event to VStream client here because the journal event includes the start position for the new shards, and the VStream client need to know that in order to distribute workload. 

The way workload distribution works is the following:
- there is a centralized table that stores (ks, shard, pos)
- every worker is assigned to certain shards based on the mod of hash(shard)
- workers will check the table to see if there is any shard that belongs to it but is not processed locally yet

sonne5:
@rohit-nayak-ps assigned this to you for now as we just discussed about this. Please let me know if the proposed solution makes sense to you. Thanks.

PR #8628, opened 2021-08-14
VStream API: Add flag to stop streaming on a reshard

rohit-nayak-ps:
## Description

A new flag StopOnReshard is added as an option that the VStream client can provide. Usually on a reshard of a shard that is participating in the stream, the VStream API sends all the events on the original shard and then automatically starts streaming from the new shard set. In this case the journal event is not sent on the stream.

This flag, if set, stops the stream after sending all the streams upto the reshard and includes the journal event pertaining to that reshard as the last event. 

## Related Issue(s)
https://github.com/vitessio/vitess/issues/8595

## Checklist


test/ci_workflow_gen.go
```diff
@@ -56,6 +56,8 @@ var (
 		"vreplication_cellalias",
 		"vstream_failover",
 		"vreplication_v2",
+		"vstream_stoponreshard_true",
```
deepthi, line 59: Just for readability, you might consider adding more underscores.

deepthi (commented):
LGTM, but let's have @sougou review this too.

sougou (approved):

test/ci_workflow_gen.go
rohit-nayak-ps: Will fix in my next PR, don't feel like babysitting CI for a couple of hours for this ...

PR #8630, opened 2021-08-16
Inclusive Naming: change terminology for blacklist

rohit-nayak-ps:
## Description

This PR changes the terms around blacklisted tables. We use `denylist` and `denied` depending on whether we are referring to a list or it is a prefix to a noun like `deniedTables` as suggested in https://inclusivenaming.org/language/word-list/

Incompatible Changes:
* GetShardTabletControl  will return `denied_tables` instead of `blacklisted_tables`

Backward Compatibility:
* SetShardTabletControl will continue to accept `blacklisted_tables` in addition to `denied_tables` in 12.0

## Checklist


deepthi:
@rohit-nayak-ps can you squash the commits so that we don't have a commit with merge conflicts in the history?

deepthi (approved):

PR #8631, opened 2021-08-16
Fixing vtorc naming convention - Functions and Variables

### LINKED ISSUE #7110: Task: Cleanup references to master and slave in vt/orchestrator

GuptaManan100:
## Description
Cleanup references to master and slave in `vt/orchestrator` in functions and variables 

## Related Issue(s)
- Fixes #7110

## Checklist

### DIFF
--- a/go/vt/orchestrator/inst/instance.go
+++ b/go/vt/orchestrator/inst/instance.go
@@ -48,10 +48,10 @@ type Instance struct {
 	LogSlaveUpdatesEnabled       bool // for API backwards compatibility. Equals `LogReplicationUpdatesEnabled`
 	LogReplicationUpdatesEnabled bool
 	SelfBinlogCoordinates        BinlogCoordinates
-	MasterKey                    InstanceKey
-	MasterUUID                   string
+	PrimaryKey                   InstanceKey
+	PrimaryUUID                  string
 	AncestryUUID                 string
-	IsDetachedMaster             bool
+	IsDetachedPrimary            bool
 
 	Slave_SQL_Running          bool // for API backwards compatibility. Equals `ReplicationSQLThreadRuning`
 	ReplicationSQLThreadRuning bool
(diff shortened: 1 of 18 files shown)


go/vt/orchestrator/inst/instance.go
```diff
@@ -48,10 +48,10 @@ type Instance struct {
 	LogSlaveUpdatesEnabled       bool // for API backwards compatibility. Equals `LogReplicationUpdatesEnabled`
 	LogReplicationUpdatesEnabled bool
 	SelfBinlogCoordinates        BinlogCoordinates
-	MasterKey                    InstanceKey
-	MasterUUID                   string
+	PrimaryKey                   InstanceKey
```
shlomi-noach, line 51: It's a minor thing, but it's confusing: `PrimaryKey` can be confused with `PRIMARY KEY`. What do you think?

shlomi-noach (approved):
This looks safe

go/vt/orchestrator/inst/instance.go
GuptaManan100: Should I make it `PrimaryInstanceKey`, I think that would clear it up

shlomi-noach: yeah, maybe? Or `SourceKey`? I'm really not sure.

GuptaManan100: We can make it `SourceKey` too. That also makes sense. Then we should also rename to `SourceUUID` for consistency. WDYT?

shlomi-noach: let's ask @deepthi

deepthi: Either is fine. In general using Source for replication related params makes sense.

GuptaManan100: Alright i'll do that

GuptaManan100: Done

PR #8594, opened 2021-08-05
srvtopo: allow unwatching from watch callbacks

vmg:
## Description

Here's a feature I forgot to implement in my previous PR. We're gonna need to stop watching specific keyspaces once we detect their cluster event is finished.

cc @deepthi @harshit-gangal @rohit-nayak-ps 

## Related Issue(s)

#8462 

## Checklist
- [ ] Should this PR be backported?
- [x] Tests were added or are not required
- [x] Documentation was added or is not required

PR #8633, opened 2021-08-16
srvtopo: allow unwatching from watch callbacks

harshit-gangal:
## Description

From @vmg: 
"Here's a feature I forgot to implement in my previous PR. We're gonna need to stop watching specific keyspaces once we detect their cluster event is finished."

cc @deepthi @harshit-gangal @rohit-nayak-ps 

## Related Issue(s)

#8462 

## Checklist

### DIFF
--- a/go/test/endtoend/vtgate/vschema/vschema_test.go
+++ b/go/test/endtoend/vtgate/vschema/vschema_test.go
@@ -23,7 +23,8 @@ import (
 	"os"
 	"testing"
 
-	"github.com/stretchr/testify/assert"
+	"github.com/google/go-cmp/cmp"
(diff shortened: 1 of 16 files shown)


harshit-gangal:
Recreated the PR from https://github.com/vitessio/vitess/pull/8594 as that PR was created on personal fork and do not permission to push changes to trigger CI tests.

deepthi:
My comment is the same as before - we need an example of how to use the callback to unwatch.
Please let me know if I'm missing that in my reading of the code.
It is acceptable if that is coming in a future PR.

deepthi (approved):

PR #8640, opened 2021-08-17
Upgrade to Go 1.17

frouioui:
## Description

Bumping the Go version of the project to [Go 1.17](https://golang.org/doc/go1.17).

## Checklist

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module vitess.io/vitess
 
-go 1.16
+go 1.17
 
 require (
 	cloud.google.com/go/storage v1.10.0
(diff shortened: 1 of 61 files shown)


go.mod
```diff
@@ -126,3 +126,69 @@ require (
 	k8s.io/code-generator v0.17.3
 	sigs.k8s.io/yaml v1.2.0
 )
+
+require (
```
deepthi, line 130: It's weird to have two `require` blocks. Will a `go mod tidy` clean this up?

ajm188: No, this is the "New Way" ™️ in [1.17](https://golang.org/doc/go1.17#go-command), specifically:

>  If a module specifies go 1.17 or higher in its go.mod file, its go.mod file now contains an explicit require directive for every module that provides a transitively-imported package. (In previous versions, the go.mod file typically only included explicit requirements for directly-imported packages.)

> Because the number of explicit requirements may be substantially larger in an expanded Go 1.17 go.mod file, the newly-added requirements on indirect dependencies in a go 1.17 module are maintained in a separate require block from the block containing direct dependencies.

.github/workflows/golangci-linter.yml
```diff
@@ -8,7 +8,7 @@ jobs:
     - name: Set up Go 1.16
```
guidoiaquinti, line 8: `Set up Go 1.16` -> `Set up Go`

go.mod
guidoiaquinti: The Mandalorian

.github/workflows/golangci-linter.yml
frouioui: Addressed via d103122

frouioui:
Tests are all green after building the new docker bootstrap image. Waiting for tomorrow morning to see the benchmarking results.

frouioui:
The upgrade presents no significant performance improvement or regression: https://benchmark.vitess.io/compare?r=d10312297ffd33b144d742aba62fe5365c6b5019&c=645f06caf8336f2ab25f8fab101b11776914ed3b

vmg (approved):
I'm a bit disappointed that the new register allocation engine in the compiler doesn't appear to have a measurable effect in the performance of the whole system. I managed to see some improvements on micro-benchmarks, so it's a bit of a bummer that this is not being propagated to macros.

guidoiaquinti (approved):

PR #8642, opened 2021-08-18
Fixes for reparent endtoend test flakiness

eeSeeGee:
Extended timeout for replication restoration, which sometimes seems to
take a while in CI.

Additionally made test insert rows truly unique, which I'm not sure
fixed anything but makes it much easier to check the state of
replication.

Signed-off-by: Aaron Young <young@squareup.com>

## Checklist

### DIFF
--- a/go/test/endtoend/reparent/reparent_test.go
+++ b/go/test/endtoend/reparent/reparent_test.go
@@ -111,7 +111,7 @@ func TestReparentNoChoiceDownPrimary(t *testing.T) {
 	require.NotEqual(t, newPrimary.Alias, tab1.Alias)
 
 	// Check new primary has latest transaction.
-	err = checkInsertedValues(ctx, t, newPrimary, 2)
+	err = checkInsertedValues(ctx, t, newPrimary, insertVal)
 	require.NoError(t, err)
 
 	// bring back the old primary as a replica, check that it catches up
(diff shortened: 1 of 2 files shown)


deepthi:
Did this help at all? Which tests were you trying to stabilize?

eeSeeGee:
The specific test we were having trouble with was `TestReparentWithDownReplica`, which ran ok locally but would usually fail in CI. We have some pretty pokey CI machines so that seemed pretty suspicious. With this change it seems to be stable.

deepthi:
> The specific test we were having trouble with was `TestReparentWithDownReplica`, which ran ok locally but would usually fail in CI. We have some pretty pokey CI machines so that seemed pretty suspicious. With this change it seems to be stable.

Ok, then it is worth the changes.. Do you have the necessary permissions to re-run the failing tests? (Which all seem like flaky ones).

eeSeeGee:
I don't seem to have permissions to re-run tests

deepthi:
> I don't seem to have permissions to re-run tests

I invited you to the contributor team. That will give you the necessary permissions.
I have re-run all the failing tests for now. Let's hope they pass on this iteration.

deepthi (approved):
LGTM

eeSeeGee:
Looks like we're good! I also can't merge apparently :)

PR #8646, opened 2021-08-19
introduced cluster_operation as a new error code in vtrpc

GuptaManan100:
## Description

Added `CLUSTER_EVENT` as return error code that will be mapped to be used as `cause for buffering` and also to `re-establish reserved connection` if outside of a transaction.

## Related Issue(s)

#8462 

## Checklist


go/mysql/sql_error.go
```diff
@@ -130,9 +130,6 @@ func mapToSQLErrorFromErrorCode(err error, msg string) *SQLError {
 	case vtrpcpb.Code_CANCELED, vtrpcpb.Code_DEADLINE_EXCEEDED, vtrpcpb.Code_ABORTED:
 		num = ERQueryInterrupted
 		ss = SSQueryInterrupted
-	case vtrpcpb.Code_UNKNOWN, vtrpcpb.Code_INVALID_ARGUMENT, vtrpcpb.Code_NOT_FOUND, vtrpcpb.Code_ALREADY_EXISTS,
-		vtrpcpb.Code_FAILED_PRECONDITION, vtrpcpb.Code_OUT_OF_RANGE, vtrpcpb.Code_UNAVAILABLE, vtrpcpb.Code_DATA_LOSS:
-		num = ERUnknownError
```
deepthi, line 135: Is this code block not needed any more? Do we never produce these error codes from query-serving?

deepthi (approved):
One question, the rest LGTM

go/mysql/sql_error.go
GuptaManan100: Actually we initialise num with `ERUnknownError`, so that switch case is redundant.

vmg:
Unless I'm mistaken, this is exactly what we agreed _not_ to do during our planning meeting. 😅 

We discussed there that right now the `vtrpcpb` error codes were a direct 1:1 mapping between the upstream GRPC error codes, a design choice that was made, according to @sougou, because we were planning to make Vitess a GRPC-oriented system back in the day, as opposed to our current design direction which aims for drop-in MySQL compatibility.

Because of this, we discussed that adding a new error code to the `vtrpcpb` error list was **not** a good design choice because it breaks the mapping with the upstream GRPC error codes while still keeping us tied to a GRPC abstraction which doesn't help much. We decided that instead it'd be better to start gradually returning MySQL errors from the vttablet, starting with the specific error for failovers, and making sure that these errors are always kept between `vtgate` and `vttablet`, and never returned to the client. This is a direction that would make implementing a lot of MySQL-compatible features much more straightforward.

@deepthi @sougou: am I going gaga here? Either way, the changes in this PR seem to be working so I intend to implement the buffering stuff on top of them, and potentially changing this again once the buffering code itself is stable. I'd rather not revert this for now.

deepthi:
> @deepthi @sougou: am I going gaga here? Either way, the changes in this PR seem to be working so I intend to implement the buffering stuff on top of them, and potentially changing this again once the buffering code itself is stable. I'd rather not revert this for now.

Obviously I did not go back and look at the discussion notes before approving this PR. Here are the notes for reference:
  * we discussed improving this situation, but there’s no good mapping between the GRPC error codes that we currently return that would allow us to identify these messages explicitly
  * grpc error code decisions were made before vitess moved towards mysql 
  * Let's go back to custom error codes so that we can stop parsing strings in buffer code
  * vtgate can never return a non-standard error code
Let's do what you proposed (move forward and come back to this as a refactor afterwards).

vmg:
Great, we're in aligment. Thanks @deepthi!

PR #8647, opened 2021-08-19
Gen4: Removed default case when dealing with weird/non-unique vindex predicate

frouioui:
## Description
This pull request removes the `Gen4 unsupported error` when dealing with vindex predicates other than `equal`, `in`, `not in`, `like`.

## Checklist


harshit-gangal:
There should be some existing cases as well that should get supported. right?

harshit-gangal (approved):

systay:
> There should be some existing cases as well that should get supported. right?

We thought there should be, but we found none

PR #8649, opened 2021-08-20
increase conn killer check to double the tx timeout value

harshit-gangal:
## Description

Updated the assert checking for tx killer to double the time for tx timeout value to cover up any delayed case.

## Checklist


GuptaManan100 (approved):
LGTM! 💎

PR #8651, opened 2021-08-20
servenv: add `--onclose_timeout` flag

fatih:
This PR adds a new `--onclose_timeout` flag that adds a timeout for
`onClose` hooks. It's very similar to the existing `--onterm_timeout`
flag.

The flag will be helpful for services that use the `--lameduck-period` flag
in conjunction with `onClose` hooks. In this scenario, a service can't add 
timeouts to `onClose` hooks; hence blocking the service shutdown.


deepthi (approved):
LGTM

PR #8652, opened 2021-08-20
vtexplain examples

hallaroo:
Adding files to the local example folder for https://github.com/vitessio/website/pull/803/


deepthi (commented):
Sorry to be picky, but it will be better to put all these files in a separate directory called vtexplain.
Then you could remove the `vtexplain_` prefix from the file names.
Otherwise it might be confusing for new users.

deepthi (approved):
Nice work! Thank you for adding this.

PR #8654, opened 2021-08-20
inclusive naming: fix a few more master references that were missed earlier

deepthi:
## Description
- Deprecate `master_connect_retry` flag
- Change default values in tablet config yaml file
- Misc. error messages, func parameters, tests
- Java files
- vitessdriver
- example client
- rowlog tool
- vagrant startup script

## Related Issue(s)

#7114 

## Checklist


deepthi:
@harshit-gangal can you review the Java changes?

config/tablet/default.yaml
```diff
@@ -98,7 +98,7 @@ hotRowProtection:
   maxGlobalQueueSize: 1000    # hot_row_protection_max_global_queue_size
   maxConcurrency: 5           # hot_row_protection_concurrent_transactions
 
-consolidator: enable|disable|notOnMaster # enable-consolidator, enable-consolidator-replicas
+consolidator: enable|disable|notOnPrimary # enable-consolidator, enable-consolidator-replicas
```
rohit-nayak-ps, line 101: Is this a breaking change? Users could have specified `notOnMaster` in a config somewhere?

java/example/src/main/java/io/vitess/example/VitessClientExample.java
```diff
@@ -52,9 +52,9 @@ public static void main(String[] args) {
     try (RpcClient client = new GrpcClientFactory()
         .create(ctx, args[0]); VTGateBlockingConnection conn = new VTGateBlockingConnection(
         client)) {
-      VTSession session = new VTSession("@master", ExecuteOptions.getDefaultInstance());
+      VTSession session = new VTSession("@primary", ExecuteOptions.getDefaultInstance());
```
rohit-nayak-ps, line 55: Actually I had left some of these references to `@master` to confirm backward compatibility. But since those PR tests passed, guess it is fine to change these now.

rohit-nayak-ps (approved):
lgtm

config/tablet/default.yaml
deepthi: That will still work. We are deprecating the old value but will still respect it if provided.
https://github.com/vitessio/vitess/blob/main/go/vt/vttablet/tabletserver/tabletserver.go#L1817-L1819

Issue #8576, opened 2021-08-02
Gen4: Filtering on top of aggregation - HAVING support


GuptaManan100:
#### Overview of the Issue

The following query does not work - 
```sql
SELECT COUNT(*) AS count_all, `order2s`.`status` AS order2s_status FROM `order2s` GROUP BY `order2s`.`status` HAVING (COUNT(*) > 4)
```

Gen4 does not support `Having` for now. v3 returns the following error
```
unsupported: filtering on results of aggregates
```

Supporting this query will require addition of a new primitive for filtering

#### Binary version
main

PR #8683, opened 2021-08-25
Gen4: Having clause support

harshit-gangal:
## Description

Add support for having clause in Gen4 plan.

## Related Issue(s)

- #7280 
- Closes https://github.com/vitessio/vitess/issues/8576

## Checklist

### DIFF
--- a/go/vt/vtgate/planbuilder/rewrite.go
+++ b/go/vt/vtgate/planbuilder/rewrite.go
@@ -124,6 +124,33 @@ func (r *rewriter) rewrite(cursor *sqlparser.Cursor) bool {
 		if node.GroupBy != nil || node.Having == nil {
 			break
 		}
+
+		selectExprMap := map[string]sqlparser.Expr{}
+		for _, selectExpr := range node.SelectExprs {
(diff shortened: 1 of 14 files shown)


systay (approved):

go/vt/vtgate/planbuilder/rewrite.go
```diff
@@ -109,48 +107,96 @@ func expandTableColumns(tables []semantics.TableInfo, starExpr *sqlparser.StarEx
...
+		rewriteHavingClause(node)
 	case *sqlparser.ExistsExpr:
-		semTableSQ, found := r.semTable.SubqueryRef[node.Subquery]
-		if !found {
-			// should never happen
-			return false
-		}
+		return r.rewriteExistsSubquery(cursor, node)
+	case *sqlparser.Subquery:
+		rewriteSubquery(cursor, r, node)
```
frouioui, line 126: I think we should either have both functions be methods of `rewriter`, or have them both take `r rewriter` as function parameter

frouioui (approved):

go/vt/vtgate/planbuilder/rewrite.go
systay: I fix in the next PR

PR #8699, opened 2021-08-27
gen4: Make sure that column names are correct

systay:
## Description
A bug had crept in that made it so column names were not always respected. This fixes it.

## Related Issue(s)
#7280 

## Checklist


GuptaManan100 (approved):
LGTM! 💕

PR #8700, opened 2021-08-27
gateway: use keyspace events when buffering requests

### LINKED ISSUE #7059: while doing cutover for resharding, an exception is thrown to app
#### Overview of the Issue

Two issues of cutover for reshard: 
`One` is an exception is thrown to App and `the other` is VTGate does not detect the end of cutover. 
To drain the buffered requested is only done by the time out of buffering duration. 

You can find the second issue as #7061 .

#### Reproduction Steps

Version: vitess/lite:v8.0.0 images and percona 8.0 image. 

```yaml
    mysqld:
      mysql80Compatible: percona:8.0
```

I don't think percona:8.0 matters. I just wanted to exploit parallel replication options of mysql 8.0 and chose percona:8.0 which is one of the easiest options. 

I also use Vitess native JDBC driver(v8.0.0) for App(JDBC).

Steps to reproduce this issue:

1. Prepare for the '-' shard of 3 VTTablet replicas:

Let's say the database is "ycsb". 

For my case, I use 128 VTGate. VTGate options are:
```yaml
        enable_buffer: "true"
        gateway_implementation: "tabletgateway"
        buffer_size: "50000"
```

No special options for VTTablet except mysqld:
```yaml
            mysqld:
              configOverrides: |
                [mysqld]
                slave_parallel_workers=20
                slave_parallel_type=LOGICAL_CLOCK
                transaction_write_set_extraction=XXHASH64
                binlog_transaction_dependency_tracking=WRITESET
```

2. Deploy the following `vschema` and `schema`:

```sh
vtctlclient ApplySchema -sql='CREATE TABLE usertable ( YCSB_KEY VARCHAR(255) PRIMARY KEY, FIELD0 TEXT, FIELD1 TEXT, FIELD2 TEXT, FIELD3 TEXT, FIELD4 TEXT, FIELD5 TEXT, FIELD6 TEXT, FIELD7 TEXT, FIELD8 TEXT, FIELD9 TEXT );' ycsb
vtctlclient ApplyVSchema -vschema='{ "sharded": true, "vindexes": { "hash": { "type": "unicode_loose_md5" } }, "tables": { "usertable": { "column_vindexes": [ { "column": "YCSB_KEY", "name": "hash" } ] } } }' ycsb
```

3. Prepare for YCSB client.

You can find it from https://github.com/brianfrankcooper/YCSB

4. Load initial data

Execute ycsb load workload to load some data. You can do it by something like this:

```sh
bin/ycsb load jdbc -P workloads/workloada -p db.driver=io.vitess.jdbc.VitessDriver -p db.url=jdbc:vitess://<VTGate>:15999/ycsb -p db.user=user -p maxexecutiontime=0 -p recordcount=500000 -threads 16 -s"
```

Please modify `<VTGate>` as yours. 

Then, it will load initial 500,000 rows. You may change the number of records. 

5. Create '-80,80-' VTTablets.

```yaml
    - equal:
        parts: 2
```

6. Reshard '-' to '-80,80-'.

```sh
vtctlclient Reshard ycsb.SplitAs2 '-' '-80,80-'
```

7. Once reshard completes copying, execute workload. 

```sh
bin/ycsb run jdbc -P workloads/workloada -p db.driver=io.vitess.jdbc.VitessDriver -p db.url=jdbc:vitess://<VTGate>:15999/ycsb -p db.user=user -p maxexecutiontime=7200 -p recordcount=500000 -p operationcount=2147483627 -threads 128 -target 1500 -s
```

It will generate 1500 ops of which 50% of read and 50% of update workload. It will print stats for every 10 seconds.

8. While workload is executing, do cutover. 

```sh
vtctlclient SwitchReads -tablet_type=rdonly ycsb.SplitAs2
vtctlclient SwitchReads -tablet_type=replica ycsb.SplitAs2
vtctlclient SwitchWrites ycsb.SplitAs2
```

8. View error

There are 2 two issues I see. 

#### issue 1.

The first one is the app is exposed with exceptions. You will see a bunch of errors like this:
 
```
Error in processing read of table usertable: java.sql.SQLNonTransientException: Vitess RPC error: legacy_code: QUERY_NOT_SERVED_LEGACY
message: "vtgate: http://my-zone1-vtgate-2f3311c0-6499d75796-4b7wp:15000/: target: ycsb.-.master, used tablet: zone1-868377885 (10.171.174.217): vttablet: rpc error: code = FailedPrecondition desc = operation not allowed in state NOT_SERVING"
code: FAILED_PRECONDITION

...
```

```
2020-11-20 11:45:46:270 250 sec: 331099 operations; 1322.3 current ops/sec; est completion in 18 days 18 hours [READ: Count=6611, Max=177407, Min=4712, Avg=61428.42, 90=102015, 99=164223, 99.9=170879, 99.99=174975] [UPDATE-FAILED: Count=32, Max=10895359, Min=10076160, Avg=10694912, 90=10895359, 99=10895359, 99.9=10895359, 99.99=10895359] [UPDATE: Count=6547, Max=177663, Min=4124, Avg=66844.16, 90=105407, 99=168703, 99.9=176639, 99.99=177663] [READ-FAILED: Count=33, Max=10903551, Min=10076160, Avg=10721466.18, 90=10895359, 99=10903551, 99.9=10903551, 99.99=10903551]
```

Please notice the failed requests: 65(32 of READ and 33 of UPDATE) for this case.  

I expected SQLNonTransientException won't be thrown to App since the failed requests are buffered and retried after cutover. 

#### issue 2. 

You can find it from #7061 .

#### Binary version

```
vitess/lite:v8.0.0 images
```

### LINKED ISSUE #7061: while doing cutover for resharding, the end of cutover is not automatically detected
#### Overview of the issue

This is another part of #7059.

VTGate does not detect the end of cutover.
To drain the buffered requested is only done by the time out of buffering duration.

#### Reproduction Steps

please find #7059 for repro steps. Repro is same as #7059.

#### View error
Once you cutover, you will see VTGate logs like this:
```
I1120 02:45:32.307383       7 shard_buffer.go:280] Starting buffering for shard: ycsb/- (window: 10s, size: 5000000, max failover duration: 10s) (A failover was detected by this seen error: Code: FAILED_PRECONDITION
vttablet: rpc error: code = FailedPrecondition desc = operation not allowed in state NOT_SERVING
.)
I1120 02:45:32.314849       7 tablet_health_check.go:110] HealthCheckUpdate(Serving State): tablet: zone1-868377885 (10.171.174.217) serving true => false for ycsb/- (MASTER) reason: healthCheck update
I1120 02:45:42.307964       7 shard_buffer.go:545] Stopping buffering for shard: ycsb/- after: 10.0 seconds due to: stopping buffering because failover did not finish in time (10s). Draining 65 buffered requests now.
I1120 02:45:43.128362       7 shard_buffer.go:565] Draining finished for shard: ycsb/- Took: 820.334657ms for: 65 requests.
```

Please notice the log:
```
stopping buffering because failover did not finish in time (10s). Draining 65 buffered requests now.
```

However, switch usually takes only 2-3 seconds. 

#### Observations

I don't know `recordExternallyReparentedTimestamp` works for resharding case.
```go
	if timestamp <= sb.externallyReparented {
		return
	}

        // where timestamp is `th.MasterTermStartTime`.
```

When I experimentally delete the primary pod, vtgate correctly detects the end of cutover.

vmg:
## Description

Fixes https://github.com/vitessio/vitess/issues/8462 (lol, hopefully).
Fixes #7059
Fixes #7061

Alright, this is the first draft of a solution for the dreaded request buffering issue. After a lot of investigation, we've arrived at this approach to fix the issue.

Summarizing:

- We're still performing the request buffering at the shard level, like we were before, but we've replaced the old `HealthCheck`-based approach to find out when a failover is finished with a brand new `KeyspaceEventWatcher`.
- The `KeyspaceEventWatcher` is a new implementation that augments `HealthCheck` events with metadata from the topology server for keyspace changes. This allows it to process any primary promotion events wholistically for the whole keyspace, so when these events are part of a resharding operation, they are only reported once the whole keyspace has been properly sharded and all the new shards are healthy and serving.
- Because of this, we can now distinguish in the Buffer code between plained failovers for a single primary (whose buffered requests can be retried) and primary promotions as part of a resharding operation (whose buffered requests cannot be retried because the shard they were targetting is now gone as part of the resharding process).
- When the latter situation arises, the buffering code reports a special error code that is handled in the execution engine. This handling has been implemented **at the primitive level** (as opposed to the initial approach that intended to handle this at the plan generation level). This means that the buffered query is only re-executed in the new shard for the primitive subquery, and not for the whole plan -- this is required for handling some corner cases, see https://github.com/vitessio/vitess/pull/8681 for examples.

As far as we can tell (and as far as I can test manually), this implementation can resume buffering after any kind of resharding event. Notably however, it does not support MoveTable events, on which we're punting for now.

The implementation has a bit of duplicated code on `Buffer` because I wanted to leave in the old implementation to make them swappable while we gain confidence on the new `KeyspaceEvent`-based buffering code.

I need a drink.

cc @deepthi @harshit-gangal @sougou 

## Related Issue(s)

## Checklist


vmg:
OK, I think this PR is ready for review merge. I spent all of today implementing tests for the new functionality and I think I was successful. I've split the existing `tabletgateway/buffer` end-to-end tests into "reparenting" and "resharding" tests, to verify both behaviors.

The new tests ensure that queries are buffered and do not error out when a Vitess cluster is reparented or resharded. They both use the new Keyspace Events Watcher API for event detection. The reparenting tests are green when using the KEW and the old HealthCheck API, but I decided to run them only with KEW because we're deprecating the old API and these are expensive tests. The resharding tests are green with the KEW and faill with the HealthCheck API, which is, huh, kinda the whole point of this project.

One last thing I noticed while stress testing the resharding operation is another failure case that we never detected and never discussed before: in a high-traffic Vitess cluster with high QPS, it's possible that a query arrives to the VTGate during a resharding operation such as that the query is planned for shard `0` but such shard has been marked as unhealthy by the resharding operation by the time the query is executed. This would usually cause `TabletGateway` to fail fast with an `UNAVAILABLE` error that is not retried, since the topology server would return no shards capable of serving the query, and the buffering code would not be aware (yet) of any buffering events for the shard (as the shard was removed before any queries could fail on it and hence start the buffering process).

This is easy to reproduce with high enough QPS, and in practice results in a very short burst of instantly failed queries from the VTGate in the exact same moment than the reparenting starts. Fortunately, the new Keyspace Event Watcher is smart enough to notice this new corner case, so I've updated `TabletGateway` to handle it explicitly: if there are no shards available to serve the query, but the KEW knows that no shards are healthy because the keyspace is currently being resharded, we'll buffer the query all the same and re-try it at the end of the event so that it lands in the new shard. :ok_hand:

go/vt/vtgate/tabletgateway.go
```diff
@@ -231,10 +258,22 @@ func (gw *TabletGateway) withRetry(ctx context.Context, target *querypb.Target,
...
+			}
 		}
 
 		tablets := gw.hc.GetHealthyTabletStats(target)
 		if len(tablets) == 0 {
+			// if we have a keyspace event watcher, check if the reason why our primary is not available is that it's currently being resharded
+			if target.TabletType == topodatapb.TabletType_PRIMARY && gw.kev != nil && gw.kev.TargetIsBeingResharded(target.Keyspace, target.Shard) {
+				err = vterrors.Errorf(vtrpcpb.Code_CLUSTER_EVENT, "current keyspace is being resharded")
+				continue
+			}
```
vmg, line 276: This is the special handling I was talking about.

go/vt/vtgate/buffer/buffer.go
```diff
@@ -61,6 +60,20 @@ const (
...
 
+// RetryDoneFunc will be returned for each buffered request and must be called
+// after the buffered request was retried.
+// Without this signal, the buffer would not know how many buffered requests are
+// currently retried.
+type RetryDoneFunc context.CancelFunc
+
+// CausedByFailover returns true if "err" was supposedly caused by a failover.
+// To simplify things, we've merged the detection for different MySQL flavors
+// in one function. Supported flavors: MariaDB, MySQL, Google internal.
```
deepthi, line 71: I don't think we need to support Google internal any more.

go/vt/vtgate/buffer/flags.go
```diff
@@ -23,60 +23,50 @@ import (
...
 	"time"
 
+	"vitess.io/vitess/go/vt/log"
 	"vitess.io/vitess/go/vt/topo/topoproto"
 )
 
 var (
-	enabled       = flag.Bool("enable_buffer", false, "Enable buffering (stalling) of primary traffic during failovers.")
-	enabledDryRun = flag.Bool("enable_buffer_dry_run", false, "Detect and log failover events, but do not actually buffer requests.")
+	f_enabled       = flag.Bool("enable_buffer", false, "Enable buffering (stalling) of primary traffic during failovers.")
```
deepthi, line 31: This seems to introduce a new naming convention for flags. I'd prefer we don't do that in this PR. If we agree on such a convention we should change all flags in one go.
Feel free to rename them to be more descriptive, e.g. window -> bufferWindow.

go/vt/vtgate/gateway.go
```diff
@@ -38,6 +37,7 @@ import (
 var (
 	// GatewayImplementation allows you to choose which gateway to use for vtgate routing. Defaults to tabletgateway, other option is discoverygateway
 	GatewayImplementation = flag.String("gateway_implementation", "tabletgateway", "Allowed values: discoverygateway (deprecated), tabletgateway (default)")
+	bufferImplementation  = flag.String("buffer_implementation", "healthcheck", "Allowed values: healthcheck (deprecated), keyspace_events (default)")
```
deepthi, line 40: Docstring says `keyspace_events` is the default but the actual default value has been specified as `healthcheck`.

deepthi (approved):
Nice work. Very clean as usual.
I had a couple of nits, otherwise LGTM.

deepthi:
Cluster test `tabletgateway_buffer_reshard` is failing right now, that needs to be resolved.

vmg:
Fixed all the tests :green_apple:, I'll merge at the end of the day in case somebody else wants to review.

deepthi:
Use the new flag `buffer_implementation=keyspace_events` to enable this feature.

go/vt/vtgate/tabletgateway.go
```diff
@@ -231,10 +258,22 @@ func (gw *TabletGateway) withRetry(ctx context.Context, target *querypb.Target,
 				defer retryDone()
 				bufferedOnce = true
 			}
+
+			if bufferErr != nil {
```
rafael, line 262: Could you provide some context on this change in the order of the logic here? We are checking retryDone, before checking the error. 

I'm not familiarized with the details of the logic here, but this chance caught my attention.

go/vt/vtgate/tabletgateway.go
rafael: Question to make sure I'm following correctly this change: Now, in addition to previous existent logic, it will use these events to retry. 

One follow up question for these retries: other errors in this block, were not that time sensitive. For instance, if a connection to a tablet failed, in the next iteration of the block another tablet would be used for the retry. In this case, there is some amount of time that needs to elapsed before new primaries will become available.  Would it make sense to start thinking of exponential backoff between retries?

rafael (commented):
Arriving late to this party. This looks good to me as well. I left couple comments that are more for my own curiosity and understanding of the change.

go/vt/vtgate/tabletgateway.go
vmg: There is no need for exponential backoff here. In this specific case, we're setting an explicit buffering error, so that the next iteration of the loop will block directly on the buffering code and will not resume until the primary becomes available. This is not a busy loop, so we don't need to backoff. :ok_hand:

go/vt/vtgate/tabletgateway.go
vmg: This is a small corner case fix: there are now cases when the buffering code returns both an error and a cancelation function. It's important to `defer` the cancelation function whenever it's returned, even if we also have an error and must exit the function right away -- not defering the function would cause a (tiny) memory leak.

go/vt/vtgate/tabletgateway.go
rafael: Nice, this makes sense.

go/vt/vtgate/tabletgateway.go
rafael: Ah cool. Makes sense as well.

go/vt/vtgate/tabletgateway.go
sonne5: Am I right that the "buffer" request here is not really put a "request" into a queue, we are queueing the entry and the actaul request goroutine just gets block until the shard split/failover finished?

go/vt/vtgate/tabletgateway.go
```diff
@@ -105,31 +105,65 @@ func NewTabletGateway(ctx context.Context, hc discovery.HealthCheck, serv srvtop
...
-		for {
-			select {
-			case <-ctx.Done():
-				return
-			case result := <-hcChan:
-				if result == nil {
-					// If result is nil it must mean the channel has been closed. Stop goroutine in that case
-					bufferCancel()
+	gw.setupBuffering(ctx)
+	gw.QueryService = queryservice.Wrap(nil, gw.withRetry)
```
sonne5, line 110: I see we rely on srv keyspace to detect shard split events, how would a failover being handled here? Say we have a failover in a shard from A to B:
t1: A became problematic / not responding
t2: Orchestrator detects the problem and do an external reparent from A to B
t3: healthcheck detect the reparent event and set primary of the shard to B

After t3, requests should be handled properly. I'm thinking before t3, what is the mechanism to handle / buffer requests (if it exists)? I'm asking because if we can buffer as much requests as possible between t1 and t3, in theory we should have higher availability. 

Is there some coordination we are doing at vttablet to trigger the buffer logic here for a reparent?

go/vt/vtgate/buffer/buffer.go
```diff
@@ -233,6 +162,15 @@ func (b *Buffer) ProcessPrimaryHealth(th *discovery.TabletHealth) {
 	sb.recordExternallyReparentedTimestamp(timestamp, th.Tablet.Alias)
 }
 
+func (b *Buffer) HandleKeyspaceEvent(ksevent *discovery.KeyspaceEvent) {
+	for _, shard := range ksevent.Shards {
+		sb := b.getOrCreateBuffer(shard.Target.Keyspace, shard.Target.Shard)
+		if sb != nil {
+			sb.recordKeyspaceEvent(shard.Tablet, shard.Serving)
```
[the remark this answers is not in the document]
vmg: We actually don't trigger any buffering operations when we receive the healcheck fail for a primary! That's because these healthchecks are processed by the topology engine, which often lags behind the actual availability issue. Instead, what we do it start buffering once the `vtgate` itself fails to reach the primary (i.e. when we get an error return during a request), and the Keyspace Events handler is designed to detect when the availability incident is over _cluster-wide_.

go/vt/discovery/keyspace_events.go
```diff
@@ -0,0 +1,359 @@
...
+	for _, shard := range primary.ShardReferences {
+		sstate := kss.shards[shard.Name]
+		if sstate == nil || !sstate.serving {
+			return
+		}
+		activeShardsInPartition[shard.Name] = true
+	}
+
+	for shard, sstate := range kss.shards {
+		if sstate.serving && !activeShardsInPartition[shard] {
```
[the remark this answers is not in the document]
vmg: That's not accurate: the first check in line 209 is iterating through all the shards that the topology service knows and making sure we already know about them and that we know them to be healthy. This second loop is iterating through all the shards that we know about, to make sure there are no healthy shards that we know about but the topology service doesn't.

go/vt/vtgate/buffer/buffer.go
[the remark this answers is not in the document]
sonne5: Thanks @vmg that makes sense. 
> start buffering once the vtgate itself fails to reach the primary (i.e. when we get an error return during a request)

it looks like the retry is triggered on these 3 kinds of error code: https://github.com/vitessio/vitess/blob/main/go/vt/vttablet/queryservice/wrapped.go#L75

My current understanding is vttablet maps to those error code here: [Code_FAILED_PRECONDITION](https://github.com/vitessio/vitess/blob/61ecf0d30c5be5e98f1465b86d0397339b66e52e/go/vt/vttablet/tabletserver/tabletserver.go#L1432) here, Code_UNAVAILABLE [here](https://github.com/vitessio/vitess/blob/61ecf0d30c5be5e98f1465b86d0397339b66e52e/go/vt/vttablet/tabletserver/tabletserver.go#L1421) and Code_CLUSTER_EVENT [here](https://github.com/vitessio/vitess/blob/61ecf0d30c5be5e98f1465b86d0397339b66e52e/go/vt/vttablet/tabletserver/tabletserver.go#L1434) - they seem not related to a primary failure. The other place where we return `Code_CLUSTER_EVENT` is when the primary tablet has !serving servingState, which only happen _after_ a reparent event.

I think I'm missing some details here on the failover behavior, could you shed some lights on it? Thanks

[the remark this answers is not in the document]
vmg: The error handling on the tablet was wired up by @harshit-gangal, so I don't know all the details, but the actual buffering code is triggered with this check, so the only relevant error code when it comes to buffering is `CLUSTER_EVENT`:

https://github.com/vitessio/vitess/blob/ca5a27ed5aacbb523d678d910cda8d5571e653f0/go/vt/vtgate/buffer/buffer.go#L72-L75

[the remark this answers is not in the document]
vmg: ...In retrospect, this function could use a different name since `CLUSTER_EVENT` can also be caused by a resharding operation. :sweat_smile:

go/vt/discovery/keyspace_events.go
```diff
@@ -0,0 +1,359 @@
...
+
+	primary := topoproto.SrvKeyspaceGetPartition(kss.lastKeyspace, topodatapb.TabletType_PRIMARY)
+	if primary == nil || len(primary.ShardTabletControls) > 0 {
+		return
+	}
+
+	activeShardsInPartition := make(map[string]bool)
+	for _, shard := range primary.ShardReferences {
+		sstate := kss.shards[shard.Name]
+		if sstate == nil || !sstate.serving {
```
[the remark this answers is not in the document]
vmg: I've documented this behavior in #8890 -- it explains all the different consistency checks.

go/vt/vtgate/buffer/buffer.go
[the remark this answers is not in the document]
sonne5: Got it, looking at the check [here](https://github.com/vitessio/vitess/blob/main/go/vt/vttablet/tabletserver/state_manager.go#L351), looks like the buffering logic only covers _after_ a reparent (but before vtgate get notified via health check), a vttablet would return Code_CLUSTER_EVENT. Since `replHealthy` is always true for primary tablet and serving state only changes via a reparent event. Does that sound right to you?

Issue #7336, opened 2021-01-21
Feature Request: Ignore all write_only vindexes when query planning


jmoldow:
#### Feature Description

Right now, the Vitess query planner / `EXPLAIN` information has no way to distinguish between a vindex that, for a given parametrized query:

- Happens to be hitting all shards for the requested parameter values and for the current data in the lookup tables; vs.
- Is always guaranteed to scatter to all shards, no matter the parameter values and the data in the lookup tables

This explains the behavior observed in #7328:

- `EXPLAIN FORMAT=vitess` reported `SelectEqualUnique` because, as far as it knew, a unique vindex was being used on a single value.
- `vtexplain` reported that the query would go to all shards, because `Map` for the write_only lookup vindex returned a list containing all the shards.

If instead the Vitess query planner had special support for understanding and handling `write_only`, then such vindexes could be ignored entirely during query planning. Then, rather than being unable to tell between the two cases above, the vindex would fall into a third category. And these problematic queries would fall through to all-shards scatter queries, which would be notated appropriately in `EXPLAIN FORMAT=vitess`, `vtexplain`, and `/debug/scatter_stats`.

#### Use Case(s)

Fix for #7328.

jmoldow:
One other interesting thing here.

Suppose there is a query with `WHERE a=? AND b=?`. Suppose this query currently utilizes a `Lookup NonUnique` vindex on column `a`.

If one runs `CreateLookupVindex` to create a `Lookup Unique` on column `b`, then the new vindex will have a lower cost https://vitess.io/zh/docs/schema-management/vschema/#cost , and will therefore be chosen.

But because the new vindex is `write_only`, the `Map` function will return all shards, leading to an all-shards scatter. Whereas if the original `Lookup NonUnique` vindex on column `a` was chosen, it wouldn't necessarily scatter to all-shards.

Whereas if `write_only` vindexes were ignored by the planner, the original `Lookup NonUnique` vindex will continue to be used, until `ExternalizeVindex` is run on the new vindex.

I haven't tried to repro this, but it makes sense, based on what I know, that this bug would manifest as described.

PR #8701, opened 2021-08-27
Fixing error on deletes from owning table for a lookup being populated by CreateLookupVindex after a SwitchWrites

frouioui:
## Description

After creating two keyspaces and a lookup Vindex using `CreateLookupVindex `, every delete query that uses the same lookup Vindex starts failing. Below are: a sample query using the column `c2` (which is our lookup Vindex); and the error that we got after executing the query.

```
mysql> delete from table where c2 = 1;
ERROR 1105 (HY000): cannot map vindex to unique keyspace id: DestinationKeyRange(-)
```

The reason behind this error is the following: the engine tries to execute a `delete equal` however, the lookup Vindex used to resolve the shard is in `write_only` mode, meaning we cannot resolve to which shard this query needs to be routed. The default behavior, when this happened, was to return a slice of `DestinationKeyRange` keyspace IDs, which is not a single shard, thus failing for `delete equal`. This pull request changes this behavior by transforming `delete unique` into `delete scatter` when the query is made using a lookup Vindex on `write_only` mode.

## Checklist


deepthi:
Does this also fix #7336?

harshit-gangal:
> Does this also fix #7336?

It does not fix the issue mentioned.

frouioui:
@harshit-gangal, as we talked about offline: I have modified the code to change the DML's opcode in the planbuilder instead of doing it in the engine

go/vt/vtgate/planbuilder/dml.go
```diff
@@ -51,6 +51,11 @@ func getDMLRouting(where *sqlparser.Where, table *vindexes.Table) (engine.DMLOpc
 			opcode := engine.Equal
 			if pv.IsList() {
 				opcode = engine.In
+			} else if lu, isLu := single.(vindexes.LookupBackfill); isLu && lu.IsBackfilling() {
+				// Checking if the Vindex is currently backfilling or not, if it isn't we can read from the vindex table
+				// and we will be able to do a delete equal. Otherwise, we do a delete scatter since we will not be able
+				// to know in which shard the row is.
+				opcode = engine.Scatter
 			}
```
harshit-gangal, line 59: we should `continue` here so that the planner checks for the next vindex in the where predicates for a match.

harshit-gangal (commented):
we can remove the changes in the engine test as the changes are associated with the planner.
Also, add unit test at executor level.

go/vt/vtgate/planbuilder/dml.go
frouioui: good call

frouioui:
@harshit-gangal, thank you for the review! I have applied the recommended changes

go/vt/vtgate/executor_dml_test.go
```diff
@@ -493,6 +493,44 @@ func TestDeleteScatter(t *testing.T) {
...
+				"keyspace_id": {Type: querypb.Type_VARBINARY, Value: []byte("\x16k@\xb4J\xbaK\xd6")},
+				"wo_lu_col":   {Type: querypb.Type_INT64, Value: []byte("1")},
+			},
+		}, {
+			Sql:           "delete from t2_wo_lookup where wo_lu_col = 1",
+			BindVariables: map[string]*querypb.BindVariable{},
+		}}
+	if !reflect.DeepEqual(sbc1.Queries, wantQueries) {
+		t.Errorf("sbc.Queries:\n%+v, want\n%+v\n", sbc1.Queries, wantQueries)
+	}
```
harshit-gangal, line 531: should also check that the query is also executed in other shards as well i.e. `sbc2` in this case.

go/vt/vtgate/planbuilder/dml.go
```diff
@@ -51,6 +51,11 @@ func getDMLRouting(where *sqlparser.Where, table *vindexes.Table) (engine.DMLOpc
 			opcode := engine.Equal
 			if pv.IsList() {
 				opcode = engine.In
+			} else if lu, isLu := single.(vindexes.LookupBackfill); isLu && lu.IsBackfilling() {
+				// Checking if the Vindex is currently backfilling or not, if it isn't we can read from the vindex table
+				// and we will be able to do a delete equal. Otherwise, we do a delete scatter since we will not be able
+				// to know in which shard the row is.
```
harshit-gangal, line 57: nit: change `Otherwise, we do a delete scatter since we will not be able to know in which shard the row is.` - > `Otherwise, we continue to look for next best vindex`.

go/vt/vtgate/vindexes/lookup_vindex.go
```diff
@@ -0,0 +1 @@
+package vindexes
```
harshit-gangal, line 1: this file can be removed.

harshit-gangal (commented):
changes look good.
More tests are needed in executor and planner where a higher cost Vindex is selected as the lower cost vindex was in backfill.

go/vt/vtgate/engine/update.go
```diff
@@ -232,16 +230,6 @@ func (upd *Update) updateVindexEntries(vcursor VCursor, bindVars map[string]*que
...
-				offset := updColValues.Offset
-				if !row[offset].IsNull() {
-					val, err := evalengine.ToInt64(row[offset])
-					if err != nil {
-						return err
-					}
-					if val == int64(1) { // 1 means that the old and new value are same and vindex update is not required.
-						continue
-					}
-				}
```
frouioui, line 244: This snippet started fails the following test case: `TestUpdateEqualWithWriteOnlyLookupUniqueVindex`. We were checking if the new value we want to assign is equal to `int64(1)` in which case we continue and go to the next updated value. This logic is incorrect when the vindex's original value is equal to `1`. The test `TestUpdateEqualWithWriteOnlyLookupUniqueVindex` represents this issue by having the `lu_col` vindex originally set to `1` and updating it to `5`.

frouioui: The right-most select expressions of the initial update statement are here to tell us whether we are updating a column to the same value or not. The value of those right-most fields is either `0` or `1` (`0` if updating to a different value, `1` if updating to the same value). The snippet deletion will be reverted, resolving this conversation.

harshit-gangal (approved):

PR #8721, opened 2021-08-30
Adding GitHub Self Hosted Runner tests

GuptaManan100:
## Description

Adds infrastructure for shifting end to end tests to self-hosted runners. It also adds the documentation to how to use the self-hosted runners to find all the debug information.

## Checklist


test/ci_workflow_gen.go
```diff
@@ -151,6 +179,81 @@ func parseList(csvList string) []string {
...
+		}
+		err := setupTestDockerFile(test)
+		if err != nil {
+			return err
+		}
+		filePath := fmt.Sprintf("%s/unit_test_%s_self_hosted.yml", workflowConfigDir, platform)
+		err = writeFileFromTemplate(unitTestSelfHostedTemplate, filePath, test)
+		if err != nil {
+			log.Print(err)
+		}
```
frouioui, line 203: Should we return the error here?

.github/dockerFiles/cluster_test_vreplication_basic/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+# Set environment variables
+ENV VTROOT /vt/src/vitess.io/vitess
+# Set the vtdataroot such that it uses the volume mount
+ENV VTDATAROOT /vt/vtdataroot
+
+# create the vtdataroot directory
+RUN mkdir -p $VTDATAROOT
+
+  # sleep for 50 minutes
+CMD sleep 3000
```
frouioui, line 39: Why sleep 3000?

.github/workflows/cluster_endtoend_vreplication_basic_self_hosted.yml
```diff
@@ -0,0 +1,37 @@
...
+
+      - name: Build Docker Image
+        run: docker build -f ./.github/dockerFiles/cluster_test_vreplication_basic/Dockerfile -t cluster_test_vreplication_basic:$GITHUB_SHA  .
+
+      - name: Run test
+        timeout-minutes: 30
+        run: docker run --name "cluster_test_vreplication_basic_$GITHUB_SHA" cluster_test_vreplication_basic:$GITHUB_SHA /bin/bash -c 'source build.env && go run test.go -keep-data=true -docker=false -print-log -follow -shard vreplication_basic -- -- --keep-data=true'
+
+      - name: Print Volume Used
+        if: ${{ always() }}
```
frouioui, line 22: nit: Can we omit the `always` condition?

frouioui (approved):
This is amazing! Looks very good to me. I left several suggestions and questions.
Can't wait to get rid of this flakiness, thanks a bunch @GuptaManan100 🚀

frouioui:
By the way, @GuptaManan100, do you have an approximate idea of how flaky the tests are, _(if they are)_, with this new configuration?

.github/dockerFiles/cluster_test_vreplication_basic/Dockerfile
```diff
@@ -0,0 +1,39 @@
+# DO NOT MODIFY: THIS FILE IS GENERATED USING "make generate_ci_workflows"
```
deepthi, line 1: nit: can you rename the directory? dockerFiles -> docker

deepthi (approved):
one suggestion, the rest LGTM.
We can merge once all feedback is addressed.

test/ci_workflow_gen.go
GuptaManan100: Actually no. That is the error that we got when we were trying to write self-hosted unit test. In case of an error we still want to continue with the other files

.github/dockerFiles/cluster_test_vreplication_basic/Dockerfile
GuptaManan100: This isn't important, since we overwrite it when we run the docker like in `docker run --name "{{.ImageName}}_$GITHUB_SHA" {{.ImageName}}:$GITHUB_SHA /bin/bash -c 'source build.env && go run test.go -keep-data=true -docker=false -print-log -follow -shard {{.Shard}} -- -- --keep-data=true`. I added it so that in case no command is specified, then the default is to sleep for 50 minutes.

.github/workflows/cluster_endtoend_vreplication_basic_self_hosted.yml
GuptaManan100: No we cannot omit the `always` condition. We want to print out the volume the docker container used irrespective of any failure in the previous step. If we remove the `always` condition then this would not run in case of testing failure and that would defeat the purpose of this step. Since it will be used only when someone tries to examine the logs which will largely happen in case of failures. Similarly we want to cleanup the docker container irrespective of failure

.github/workflows/cluster_endtoend_vreplication_basic_self_hosted.yml
```diff
@@ -0,0 +1,37 @@
+# DO NOT MODIFY: THIS FILE IS GENERATED USING "make generate_ci_workflows"
+
+name: Cluster Self Hosted (vreplication_basic)
+on: [push, pull_request]
+
```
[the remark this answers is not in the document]
GuptaManan100: I think that it would not help at all. Since we have only 4 runners setup, in case there are more jobs they will be queued up. Also, I salted the name of the docker container and image with the `GITHUB_SHA` for the reason that it allows us to run the same workflow with two different SHAs in parallel. Am I missing some use case that concurrency will allow us to solve?

GuptaManan100:
> By the way, @GuptaManan100, do you have an approximate idea of how flaky the tests are, _(if they are)_, with this new configuration?

Yes, @frouioui The unit tests fail about once in 3 runs and vreplication tests fail once in 4 runs. It is less flakier than GitHub runners and faster too. The major motivation of this however, is to be able to look at the logs of the vtdataroot directory and fix the flakiness. With this change we will be able to easily move a test to the self-hosted runners if it is flaky and fix the issue

.github/dockerFiles/cluster_test_vreplication_basic/Dockerfile
GuptaManan100: Yes, Ill do that

.github/workflows/cluster_endtoend_vreplication_basic_self_hosted.yml
frouioui: Got it!

frouioui:
> > By the way, @GuptaManan100, do you have an approximate idea of how flaky the tests are, _(if they are)_, with this new configuration?
> 
> Yes, @frouioui The unit tests fail about once in 3 runs and vreplication tests fail once in 4 runs. It is less flakier than GitHub runners and faster too. The major motivation of this however, is to be able to look at the logs of the vtdataroot directory and fix the flakiness. With this change we will be able to easily move a test to the self-hosted runners if it is flaky and fix the issue

This is amazing

.github/workflows/cluster_endtoend_vreplication_basic_self_hosted.yml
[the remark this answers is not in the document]
frouioui: The concurrency feature will allow us to limit the number of jobs in the queue and thus speed up the build process. Let's assume a new commit triggers `n` new builds that run across our fleet of `4` servers. We create a push a new commit on a pull request `p`, this triggers `n` builds, if we create a push a second new commit on top of the first, this will trigger an additional `n` builds. Now our queue is `2n` long, minus the builds that might have already finished. To alleviate the size of the queue getting too long as we add a new commit to `p`, we can use the concurrency feature. The concurrency feature will cancel all existing builds related to `p` in the queue if new builds are added.
We already integrate this feature for most of our other builds: https://github.com/vitessio/vitess/pull/8546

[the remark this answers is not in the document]
GuptaManan100: Yes this seems like a great idea. Thankyou! I'll do that too

frouioui (approved):
🚀

PR #8749, opened 2021-08-31
VReplication: testing CHAR with binary collation

shlomi-noach:
## Description
VPlayer seems to be overpadding chharacter fields with binary collation, resulting with incorrect data. Existing tests don't seem to cover this case.

This PR introduces a test that fails while this bug exists. 

## Related Issue(s)
Reproducing the bug reported in https://github.com/vitessio/vitess/pull/8730

## Checklist

cc @mattlord


mattlord:
This was incorporated into https://github.com/vitessio/vitess/pull/8730. We can close this unless you wanted to do something special here. Thanks!

shlomi-noach:
Thank you!

PR #8757, opened 2021-09-01
Allow min TLS version for tablet to mysqld conns

mattlord:

## Description
When tablets do not connect to mysqld via a unix domain socket, TLS is a concern. And given the version of MySQL involved and the specifics of the key materials used you may need the tablet to [support older ciphers](https://github.com/vitessio/vitess/blob/da96136fe5f430ee8f83b0d447286ba1b9f45341/go/vt/vttls/vttls.go#L33-L59). 

This was seen when using RDS/Aurora instances as the vttablet then uses TCP/IP to communicate with mysqld and you likely want to use TLS to protect that network communication. The [RDS CA bundle](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.SSL.html) does not seem to support ECDHE ciphers. And you may simply be using an [older version of MySQL that does not have TLS 1.2 support](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/CHAP_MySQL.html#MySQL.Concepts.SSLSupport).

Whatever the cause, there may be times where you need to support pre 1.2 TLS for one reason or another and while [vtgate supports that](https://github.com/vitessio/vitess/blob/de7f133dbe2dd6a7910f13c682910a4f5c0ac0df/go/vt/vtgate/plugin_mysql_server.go#L67), vttablet did not. This PR adds a new vttablet flag to support it: `-db_tls_min_version=TLSv1.1`

## Checklist

### DIFF
--- a/go/vt/dbconfigs/dbconfigs.go
+++ b/go/vt/dbconfigs/dbconfigs.go
@@ -82,6 +82,7 @@ type DBConfigs struct {
 	SslCaPath                  string        `json:"sslCaPath,omitempty"`
 	SslCert                    string        `json:"sslCert,omitempty"`
 	SslKey                     string        `json:"sslKey,omitempty"`
+	TLSMinVersion              string        `json:"tlsMinVersion,omitempty"`
 	ServerName                 string        `json:"serverName,omitempty"`
 	ConnectTimeoutMilliseconds int           `json:"connectTimeoutMilliseconds,omitempty"`
 	DBName                     string        `json:"dbName,omitempty"`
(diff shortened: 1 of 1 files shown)


dbussink (approved):
> The [RDS CA bundle](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.SSL.html) does not seem to support ECDHE ciphers.

The CA and which algorithm it uses (RSA or ECDSA) doesn't have anything to do with whether ECDHE is supported or not. RSA-ECDHE is also a very commonly used handshake, also with RSA certificates then. This shouldn't be confused with ECDSA certificates which is something else. 

ECDHE stands for ephemeral Diffie-Hellman negotiation using elliptic curves and is what provides [perfect forward secrecy](https://en.wikipedia.org/wiki/Forward_secrecy). It's often used with RSA and perfectly compatible with that. 

I don't think this makes any difference for the changes we want to allow here, but I wanted to make sure to call out the correct terminology since the wrong terms can be very confusing and put people onto the wrong path. 

Next to this change, I think we also should consider updating the cipher list we have here (or perhaps even remove it and fall back to Go defaults), but that's something for a different PR.

I think that due to sadly there still being many outdated MySQL versions that don't even get patched to the latest 5.6 or 5.7 releases (which support TLS properly), we're basically forced to add this for backwards compatibility for now. 

I do also want to call out https://github.com/golang/go/issues/45428 which doesn't have an immediate impact, as with Go 1.18 the only change is that TLS 1.2 is the default minimum if not configured (but we already configure the default minimum anyway). 

But that does show the long term plan that TLS 1.0 & 1.1 support will be removed at some point from Go itself at which point we can't support it anymore in Vitess either. So we really do need to push folks where we can towards at least upgrading to the latest patch releases of their MySQL servers so that they can handle this.

PR #8762, opened 2021-09-03
Upgrade to k8s 1.18 client

dctrwatson:
## Description
To begin upgrading downstream dependencies (like https://github.com/planetscale/vitess-operator) we need to also upgrade vitess' k8s client.

Currently, we're only trying to upgrade vitess-operator to the 1.18 client, so just matching that for now.


Makefile
```diff
@@ -373,7 +373,7 @@ minimaltools:
 dependency_check:
 	./tools/dependency_check.sh
 
-install_k8s-code-generator: tools.go go.mod
+install_k8s-code-generator: tools/tools.go go.mod
```
dctrwatson, line 376: This was the real change

carsonoid (approved):
The whitespace changes in the `Makefile` seem like they should be a dedicated PR or  commit. But overall The change looks good. The topo uses pretty basic k8s functionality so it should be fine with an upgraded client package.

PR #8770, opened 2021-09-06
Fix bug in parsing of version comments

### LINKED ISSUE #8753: Version-specific MySQL comments can fail even if vtgate `-mysql_server_version` is set correctly
This also happens to break 8.0.x mysqldump with an error like:
```
mysqldump: Couldn't execute '/*!80000 SET SESSION information_schema_stats_expiry=0 */': query was empty (1105)
```

Scenario:
  * Recent main `4358c6a85e`;  compiled with go 1.17
  * MySQL Community 8.0.23 running behind Vitess
  * vtgate running with flag:  `-mysql_server_version=8.0.23-Vitess`
  * Connect to vtgate, and try some MySQL version specific comments:
    * This works (as expected):
  ```
mysql> /*!40001 SET SESSION information_schema_stats_expiry=0 */;
  ```
  * However, this should work, but does not:
  ```
mysql> /*!80000 SET SESSION information_schema_stats_expiry=0 */;
ERROR 1065 (42000): query was empty
   ```
  * Some instrumentation reveals the problem in `go/vt/servenv/buildinfo.go`, where we set the `sqlparser.MySQLVersion` variable that the parser uses to parse version-specific comments in the `init()` function.  However, we populate it with a value derived from the flag `-mysql_server_version`.  But at this point, `flag.Parse()` hasn't been called to populate the flag variable value;  so the version for the purposes of the parser defaults to `5.7.9`.  This can be seen by trying some boundary values:
   ```
mysql> /*!50709 SET SESSION information_schema_stats_expiry=0 */; 
Query OK, 0 rows affected (0.01 sec)                    <---- works

mysql> /*!50710 SET SESSION information_schema_stats_expiry=0 */;
ERROR 1065 (42000): query was empty                     <---- fails
```

Proving that the parser thinks the MySQL version is 5.7.9, ignoring the commandline argument.

systay:
## Description
We were reading the version flag before having parsed the flags, and that does not work, so the running version was always set to 5.7

## Related Issue(s)
Fixed #8753

## Checklist

### DIFF
--- a/go/test/endtoend/vtgate/mysql80/main_test.go
+++ b/go/test/endtoend/vtgate/mysql80/main_test.go
@@ -56,8 +56,8 @@ func TestMain(m *testing.M) {
 		}
 
 		clusterInstance.VtGateExtraArgs = append(clusterInstance.VtGateExtraArgs,
-			"-enable_system_settings", "true",
-			"-mysql_server_version", "8.0.16-7",
+			"-enable_system_settings=true",
+			"-mysql_server_version=8.0.16-7",
 		)
 		// Start vtgate
 		err = clusterInstance.StartVtgate()
--- a/go/test/endtoend/vtgate/mysql80/main_test.go
+++ b/go/test/endtoend/vtgate/mysql80/main_test.go
@@ -55,7 +55,10 @@ func TestMain(m *testing.M) {
 			return 1
 		}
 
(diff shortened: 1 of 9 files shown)


GuptaManan100 (approved):
Good catch!

GuptaManan100:
End to end tests seem to be failing though @systay

harshit-gangal (approved):

PR #8778, opened 2021-09-08
Print actual current datetime for vtctl dry run commands

mattlord:

## Description
Before this, we were printing out [the constant defined here](https://pkg.go.dev/time#pkg-constants) rather than using that as a formatting directive applied to the current time. For example:
```
$ vtctlclient -server localhost:15999 SwitchReads -tablet_type=rdonly -reverse -dry_run targetkeyspace.workflowname
Dry Run results for SwitchReads run at 02 Jan 06 15:04 MST
Parameters: -tablet_type=rdonly -reverse -dry_run targetkeyspace.workflowname
```

This PR simply corrects that (and also adjusts the print formatting parameter ordering in the generic output).

## Checklist

### DIFF
--- a/go/vt/vtctl/vtctl.go
+++ b/go/vt/vtctl/vtctl.go
@@ -2288,7 +2288,7 @@ func commandVRWorkflow(ctx context.Context, wr *wrangler.Wrangler, subFlags *fla
 	}
 	if *dryRun {
 		if len(*dryRunResults) > 0 {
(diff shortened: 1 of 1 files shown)


ajm188 (approved):

ajm188:
I skimmed a couple of the vrep test suites and they seem like flakes, so I'm going to retrigger all of the failed ones so we can merge this.

Issue #8779, opened 2021-09-08
Multiregion support for VTGR


sonne5:
#### Feature Description
VTGR expects a mysql group under the hood. When we do a multi-region deploy, we usually do a group per region to avoid x-region paxos round. 

However, there should be only one primary vttablet x-region. In order to support that, we need to add a flag to indicate if a VTGR is running in an "active" region or "inactive" region -- only VTGR in "active" region will modify the topology records to do a reparent.

The high-level set up is the following:
![image](https://user-images.githubusercontent.com/7486790/132451663-cdf8a80d-12ee-4cb6-8b1b-92176fda4ead.png)

Only active region will trigger the reparent for Vitess, the inactive region just replicate data and maintain the group membership.

PR #8780, opened 2021-09-08
Add isActive flag to vtgr to support multi-cell topology

sonne5:

## Description

When we deploy VTGR in more than one region, we usually deploy one mysql group per region to avoid cross-region paxos round. 

Because of that, we need to define an "active" region for VTGR: only VTGR in the active region will do the `tmc.ChangeType` call. VTGR in the inactive region will just maintain the mysql group without modifying vttablet records in topo server

See the diagram in #8779 for more illustration

## Related Issue(s)

Resolves https://github.com/vitessio/vitess/issues/8779

## Checklist

### DIFF
--- a/go/vt/vtgr/controller/repair.go
+++ b/go/vt/vtgr/controller/repair.go
@@ -308,7 +308,7 @@ func (shard *GRShard) getGTIDSetFromAll(skipPrimary bool) (*groupGTIDRecorder, *
 	primary := shard.findShardPrimaryTablet()
 	var mysqlPrimaryHost string
 	var mysqlPrimaryPort int
-	// skipMaster is true when we manual failover or if there is a unreachalbe primary tablet
+	// skipPrimary is true when we manual failover or if there is a unreachalbe primary tablet
 	// in both case, there should be a reconciled primary tablet
 	if skipPrimary && primary != nil {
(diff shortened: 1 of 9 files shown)


sonne5:
Reviewers: please review with `Hide whitespace changes` enabled

go/vt/vtgr/controller/repair.go
```diff
@@ -308,7 +308,7 @@ func (shard *GRShard) getGTIDSetFromAll(skipPrimary bool) (*groupGTIDRecorder, *
 	primary := shard.findShardPrimaryTablet()
 	var mysqlPrimaryHost string
 	var mysqlPrimaryPort int
-	// skipPrimary is true when we manual failover or if there is a unreachalbe primary tablet
+	// skipMaster is true when we manual failover or if there is a unreachalbe primary tablet
```
deepthi, line 311: have to change this back.

deepthi (commented):
LGTM. We can merge after the review is addressed.

go/vt/vtgr/controller/repair.go
sonne5: Yep, done

deepthi (approved):

PR #8791, opened 2021-09-09
fix parsing of MySQL Server Version flag

GuptaManan100:
## Description
We should only parse the MySQL Server Version flag if it is provided

## Checklist

### DIFF
--- a/go/vt/sqlparser/parser.go
+++ b/go/vt/sqlparser/parser.go
@@ -47,7 +47,7 @@ var parserPool = sync.Pool{
 var zeroParser yyParserImpl
 
 // MySQLVersion is the version of MySQL that the parser would emulate
-var MySQLVersion string
+var MySQLVersion = "50709" // default version if nothing else is stated
 
 // yyParsePooled is a wrapper around yyParse that pools the parser objects. There isn't a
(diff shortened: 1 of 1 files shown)


frouioui (approved):
Looking good!

go/vt/sqlparser/parser.go
```diff
@@ -110,12 +110,14 @@ func Parse2(sql string) (Statement, BindVars, error) {
 func checkParserVersionFlag() {
 	if flag.Parsed() {
 		versionFlagSync.Do(func() {
-			convVersion, err := convertMySQLVersionToCommentVersion(*MySQLServerVersion)
-			if err != nil {
-				log.Error(err)
-				MySQLVersion = "50709" // default version if nothing else is stated
-			} else {
-				MySQLVersion = convVersion
+			MySQLVersion = "50709" // default version if nothing else is stated
```
systay, line 113: shouldn't we move this to the var declaration, so that before flags are parsed, we at least behave as 5.7?

GuptaManan100: yes, that seems like a good idea. I'll do that

frouioui (approved):

PR #8786, opened 2021-09-09
Upgrade deps to get critical vuln fixes (CVE/NVD)

mattlord:
Signed-off-by: Matt Lord <mattalord@gmail.com>

## Checklist
- [ ] Should this PR be backported?
- [ ] Tests were added or are not required
- [ ] Documentation was added or is not required

Issue #8785, opened 2021-09-09
Upgrade consul/api to at least 1.10.1


ajm188:
#### Overview of the Issue

The current version of consul has a few open CVEs:
- https://nvd.nist.gov/vuln/detail/CVE-2020-25201
- https://nvd.nist.gov/vuln/detail/CVE-2020-7219 
- https://nvd.nist.gov/vuln/detail/CVE-2021-28156
- https://nvd.nist.gov/vuln/detail/CVE-2021-32574
- https://nvd.nist.gov/vuln/detail/CVE-2021-36213
- https://nvd.nist.gov/vuln/detail/CVE-2019-12291
- https://nvd.nist.gov/vuln/detail/CVE-2020-13250

According to the NVD, upgrading to at least 1.10.1 should address all of these (even the few that technically don't apply but still show up in scans).

PR #8784, opened 2021-09-08
Upgrade consul api: `go get github.com/hashicorp/consul/api@v1.10.1`

ajm188:
Signed-off-by: Andrew Mason <amason@slack-corp.com>

## Description

This bumps consul past a version that contains fixes for several CVEs (I will file an issue with more details soon)

## Related Issue(s)

Fixes #8785

## Checklist
- [ ] Should this PR be backported?
- [ ] Tests were added or are not required
- [ ] Documentation was added or is not required

PR #8794, opened 2021-09-09
Remove consul-api usage in favor of official consul/api

ajm188:

## Description

This is a subset of the code changes made in https://github.com/vitessio/vitess/pull/8786, specifically the changes to `go/vt/orchestrator` so that we can remove `armon/consul-api` as a dependency.

This is a follow-up to #8784.

## Related Issue(s)

#8785 

## Checklist

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,7 +13,6 @@ require (
 	github.com/HdrHistogram/hdrhistogram-go v0.9.0 // indirect
 	github.com/PuerkitoBio/goquery v1.5.1
 	github.com/aquarapid/vaultlib v0.5.1
-	github.com/armon/consul-api v0.0.0-20180202201655-eb2c6b5be1b6
 	github.com/armon/go-metrics v0.0.0-20190430140413-ec5e00d3c878 // indirect
 	github.com/aws/aws-sdk-go v1.34.2
 	github.com/buger/jsonparser v0.0.0-20200322175846-f7e751efca13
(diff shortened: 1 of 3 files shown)


mattlord (approved):

mattlord:
Just noting that the reason the usage of `this` was replaced is due to the linter warnings/suggestions that occur when committing the import change within the file:
```
Linting go/vt/orchestrator/kv
       3 suggestions for:
go run golang.org/x/lint/golint go/vt/orchestrator/kv/consul.go

Lint suggestions were found. They're not enforced, but we're pausing
to let you know before they get clobbered in the scrollback buffer.

Press enter to cancel, "s" to step through the warnings or type "ack" to continue: s

go/vt/orchestrator/kv/consul.go:68:1: receiver name should be a reflection of its identity; don't use generic names such as "this" or "self"
go/vt/orchestrator/kv/consul.go:77:1: receiver name should be a reflection of its identity; don't use generic names such as "this" or "self"
go/vt/orchestrator/kv/consul.go:91:1: receiver name should be a reflection of its identity; don't use generic names such as "this" or "self"
```

The warning/suggestion comes from the golang best practice noted here: https://github.com/golang/go/wiki/CodeReviewComments#receiver-names

We could just leave that usage in `go/vt/orchestrator/kv/consul.go` alone since it looks like orchestrator uses `this` as the method receiver name all over the place.

harshit-gangal (approved):

mattlord:
Split from https://github.com/vitessio/vitess/pull/8784

PR #8798, opened 2021-09-13
Disable GitHub runners

GuptaManan100:
## Description
Disable GitHub runners for the tests that are being run on the self-hosted runners. 

## Checklist


deepthi (approved):
:100:

PR #8800, opened 2021-09-13
gen4: Simplify handling of ORDER BY 1 and GROUP BY 1

systay:
## Description
Refactoring of gen4. Instead of handling `ORDER BY 1` and `GROUP BY 1` in a few different places, we rewrite the query early on, so the rest of the planning process can deal with a simpler query.

## Related Issue(s)
#7280 

## Checklist


frouioui (approved):

go/vt/vtgate/semantics/early_rewriter.go
```diff
@@ -0,0 +1,152 @@
...
+				r.err = vterrors.Errorf(vtrpcpb.Code_UNIMPLEMENTED, "cannot use column offsets in %s when using `%s`", r.clause, sqlparser.String(expr))
+				return true
+			}
+		}
+
+		aliasedExpr, ok := currScope.selectStmt.SelectExprs[num-1].(*sqlparser.AliasedExpr)
+		if !ok {
+			r.err = vterrors.Errorf(vtrpcpb.Code_INTERNAL, "don't know how to handle %s", sqlparser.String(node))
+			break
+		}
```
harshit-gangal, line 101: the aliasExpr can be stored and used later.

harshit-gangal (approved):

PR #8805, opened 2021-09-13
Slight improvement of the ACL error message

frouioui:
## Description

This pull request improves the ACL error message. The original error message was constructed as follows:

```
"table acl error: <USERNAME> [<GROUPS>] cannot run <QUERY TYPE> on table <TABLE NAME>"
```

The first suggestion made by @deepthi, and pushed to this PR via d0db975, looks as follows:

```
"authorization error: user <USERNAME> is not authorized to run <QUERY TYPE> on table <TABLE NAME>"
```

Another suggestion which was to use MySQL's error message would look as follows:

```
"<QUERY TYPE> command denied to user '<USERNAME>' for table '<TABLE NAME>'"
```

This pull request is open to suggestions.

## Checklist


mattlord:
IMO, the more closely we mimic mysqld from the client's perspective the better -- so I'd vote for doing that here.

frouioui:
@mattlord, that's a good idea. I've applied the suggestion through 498f1dd.

Minor difference with MySQL is that the plan name is printed in pascal case instead of upper case. This is meant to avoid long and odd strings like `SHOWMIGRATIONLOGS`, and instead, print `ShowMigrationLogs`.

go/vt/vttablet/tabletserver/query_executor.go
```diff
@@ -444,7 +444,7 @@ func (qre *QueryExecutor) checkAccess(authorized *tableacl.ACLResult, tableName
 		}
 
 		if qre.tsv.qe.strictTableACL {
-			errStr := fmt.Sprintf("table acl error: %q %v cannot run %v on table %q", callerID.Username, callerID.Groups, qre.plan.PlanID, tableName)
+			errStr := fmt.Sprintf("%v command denied to user '%s' for table '%s' (ACL check error)", qre.plan.PlanID.String(), callerID.Username, tableName)
```
harshit-gangal, line 447: formatter can be `%s` instead of `%v` as it is string type.

harshit-gangal (approved):

deepthi:
@frouioui `unit_race` seems to be flaky. Can we move it to self-hosted runner?
cc @GuptaManan100

go/vt/vttablet/tabletserver/query_executor.go
```diff
@@ -444,7 +444,7 @@ func (qre *QueryExecutor) checkAccess(authorized *tableacl.ACLResult, tableName
 		}
 
 		if qre.tsv.qe.strictTableACL {
-			errStr := fmt.Sprintf("table acl error: %q %v cannot run %v on table %q", callerID.Username, callerID.Groups, qre.plan.PlanID, tableName)
+			errStr := fmt.Sprintf("%s command denied to user '%s' for table '%s' (ACL check error)", qre.plan.PlanID.String(), callerID.Username, tableName)
```
Phanatic, line 447: This is dropping the `groups` information for a given user when the ACL check fails. 
We should consider adding a user's groups if its not an empty set. 

```suggestion
			errStr := fmt.Sprintf("%s command denied to user '%s' for table '%s' (ACL check error)", qre.plan.PlanID.String(), callerID.Username, tableName)
			if len(callerID.Groups)) > 0 {
						errStr := fmt.Sprintf("%s command denied to user '%s' , in groups [%s], for table '%s' (ACL check error)", qre.plan.PlanID.String(), callerID.Username,  strings.Join(callerID.Groups, ","), tableName)

			}
```

Phanatic (approved):
Looks great! I left a comment about retaining `Groups` information in the error message.

go/vt/vttablet/tabletserver/query_executor.go
frouioui: Applied the change via aa90f25

frouioui:
> @frouioui `unit_race` seems to be flaky. Can we move it to self-hosted runner?
> cc @GuptaManan100

Created https://github.com/vitessio/vitess/pull/8854 to migrate this test to our self-hosted runners.

harshit-gangal (approved):

PR #8807, opened 2021-09-13
Add support for TLS certification revocation list (CRL) files

hkdsun:
Signed-off-by: Hormoz Kheradmand <hormoz.kheradmand@shopify.com>

### Background

Certificate Revocation Lists (CRLs) are lists of certificates that have been revoked by the Certificate Authority (CA) before they were due to expire. Support for CRL files were added to mysql many years ago (see https://bugs.mysql.com/bug.php?id=31224)

A CRL lists certificates by their uniquely identifiable serial number. The CA is responsible for tracking this and issuing the CRL. For example, in our case, the [Hashicorp Vault PKI secrets engine](https://www.vaultproject.io/docs/secrets/pki) issue these CRL files and we can upload the resulting PEM file to our infrastructure. 

We have this implemented for our MySQL container, proxysql containers, and we would like to add this feature to Vitess, so that we are prepared for the event where we need to revoke any of our certificates.

### Changes in PR

This PR adds the CRL option to various network links in Vitess:

* Implement a CRL check in `vttls.go`, using the `tls` package's `VerifyPeerCertificate` extension hook.

```
	// VerifyPeerCertificate, if not nil, is called after normal
	// certificate verification by either a TLS client or server. It
	// receives the raw ASN.1 certificates provided by the peer and also
	// any verified chains that normal processing found. If it returns a
	// non-nil error, the handshake is aborted and that error results.
	//
	// If normal verification fails then the handshake will abort before
	// considering this callback. If normal verification is disabled by
	// setting InsecureSkipVerify, or (for a server) when ClientAuth is
	// RequestClientCert or RequireAnyClientCert, then this callback will
	// be considered but the verifiedChains argument will always be nil.
	VerifyPeerCertificate func(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error
```

* Wire up a path-to-CRL-file option (e.g. `grpc_crl`, `tablet_grpc_crl`, ...)  for various usages:
  * Servers (listeners) verifying incoming connections (`ServerConfig` callers)
    * Vtgate MySQL server
    * GRPC server
  * Clients verifying downstream servers (`ClientConfig` callers)
    * LDAP_auth_server clients
    * GRPC clients
    * MySQL clients

Since I am not very familiar with running a "fully managed" tablet through `mysqlctl`, I am not sure if my work covers the communication links between various vttablets.. In our infrastructure, we had to modify our replication setup statement (add the `MASTER_SSL_CRL = 'crl_file_name'` to our `CHANGE MASTER TO` statement).

### Offline vs online CRL checks

Please note that this PR only adds the ability to specify a path to CRL files.

Out in the open internet, CRLs are fetched from "distribution endpoints" (URLs) that are specified as a field on the client/server's certificate file. However online checks come with interesting challenges, such as "what do you do when the endpoint itself is down?" or "can an attacker DoS your CRL endpoint, rendering your CRL checks ineffective?"

Note that MySQL itself is also lacking the "online check" (see https://bugs.mysql.com/bug.php?id=75387)

### Q&A

* **What happens if the CRL file doesn't exist or cannot be read?** In that case, the Vitess process will most likely fail to start up since the CRL file is read as soon as either `vttls.ServerConfig` or `vttls.ClientConfig` are called.

* **What happens if the CRL file itself needs to be updated (is past its `updateAfter` field)?** In that case, we choose to still use the CRL. i.e. we will reject revoked certificates but log a warning for the user to update their CRL file.

## Checklist

## Deployment Notes

All new code here is behind optional arguments – I don't think we need any special considerations.

### DIFF
--- a/go/vt/vttls/crl.go
+++ b/go/vt/vttls/crl.go
@@ -1,0 +1,65 @@
+/*
+Copyright 2021 The Vitess Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package vttls
+
+import (
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"fmt"
+	"io/ioutil"
+	"time"
+
+	"vitess.io/vitess/go/vt/log"
+)
+
+type verifyPeerCertificateFunc func([][]byte, [][]*x509.Certificate) error
+
+func certIsRevoked(cert *x509.Certificate, crl *pkix.CertificateList) bool {
+	if crl.HasExpired(time.Now()) {
+		log.Warningf("The current Certificate Revocation List (CRL) is past expiry date and must be updated. Revoked certificates will still be rejected in this state.")
+	}
+
+	for _, revoked := range crl.TBSCertList.RevokedCertificates {
+		if cert.SerialNumber.Cmp(revoked.SerialNumber) == 0 {
+			return true
+		}
+	}
+	return false
+}
+
+func verifyPeerCertificateAgainstCRL(crl string) (verifyPeerCertificateFunc, error) {
+	body, err := ioutil.ReadFile(crl)
+	if err != nil {
+		return nil, err
+	}
+
+	parsedCRL, err := x509.ParseCRL(body)
+	if err != nil {
+		return nil, err
+	}
+
+	return func(_ [][]byte, verifiedChains [][]*x509.Certificate) error {
+		for _, chain := range verifiedChains {
+			for _, cert := range chain {
+				if certIsRevoked(cert, parsedCRL) {
+					return fmt.Errorf("Certificate revoked: CommonName=%v", cert.Subject.CommonName)
+				}
+			}
+		}
+		return nil
+	}, nil
+}
--- a/go/vt/vttls/vttls.go
+++ b/go/vt/vttls/vttls.go
@@ -128,7 +128,7 @@ var onceByKeys = sync.Map{}
 
 // ClientConfig returns the TLS config to use for a client to
 // connect to a server with the provided parameters.
-func ClientConfig(mode SslMode, cert, key, ca, name string, minTLSVersion uint16) (*tls.Config, error) {
+func ClientConfig(mode SslMode, cert, key, ca, crl, name string, minTLSVersion uint16) (*tls.Config, error) {
 	config := newTLSConfig(minTLSVersion)
 
 	// Load the client-side cert & key if any.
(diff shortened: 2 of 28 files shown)


aquarapid (approved):
This looks nice and clean, thank you.

The only comment I have is that this does not add specific revocation checks for the CRL addition to the LDAP auth server and the binlog server code, but since we do not do specific TLS tests for those today, that is probably fine.

go/vt/vttls/crl.go
```diff
@@ -0,0 +1,65 @@
...
+	"fmt"
+	"io/ioutil"
+	"time"
+
+	"vitess.io/vitess/go/vt/log"
+)
+
+type verifyPeerCertificateFunc func([][]byte, [][]*x509.Certificate) error
+
+func certIsRevoked(cert *x509.Certificate, crl *pkix.CertificateList) bool {
```
dbussink, line 31: Should this also validate that whoever signed `cert` here also has signed the CRL? See also https://pkg.go.dev/crypto/x509#Certificate.CheckCRLSignature.

go/vt/vttls/vttls.go
```diff
@@ -128,7 +128,7 @@ var onceByKeys = sync.Map{}
 
 // ClientConfig returns the TLS config to use for a client to
 // connect to a server with the provided parameters.
-func ClientConfig(mode SslMode, cert, key, ca, name string, minTLSVersion uint16) (*tls.Config, error) {
+func ClientConfig(mode SslMode, cert, key, ca, crl, name string, minTLSVersion uint16) (*tls.Config, error) {
```
dbussink, line 131: It's not really clear here from the signature, but the `ca` field here supports a list of PEM encoded CA certificates as a root store with multiple certificates. 

That might be confusing for the `crl` argument here if that only supports one CRL? What if `ca` contains 2 roots for validation (for example when changing the root CA, or with a larger system list), then it would not be possible to add multiple `crl`s for each CA? 

Or does this already work when multiple CRLs are PEM encoded in the given file?

go/vt/vttls/crl.go
hkdsun: good idea! i can take a look at this improvement

go/vt/vttls/vttls.go
hkdsun: You are totally right. I did leave this undefined in the interest of time.

Currently, only the first PEM-encoded CRL is parsed from the file.

Happy to look at this, at the risk of delaying this PR since I have more priority work upcoming unfortunately

dbussink: > Currently, only the first PEM-encoded CRL is parsed from the file.

How hard would it be to read each PEM encoded CRL from the file so that that would provide a way to load multiple? This question is also specifically relevant combined with the other one around validating the signature of the CRL so it can be associated with the right CA then.

hkdsun: I wouldn't say this is hard at all, just very intricate. It takes a lot of time to test and think through.

I have this commit as a proposal: https://github.com/Shopify/vitess/commit/cd042505330383d6410d4501baf9582db6a98569

Two notes that made me hesitant bringing it over here just yet:

1. I haven't had time to dig into what exactly we get as the `verifiedChains` argument. So I don't know if my assumption about `cert := chain[i], issuerCert := chain[i+1]` is entirely correct. It does achieve the validation of "whoever signed cert here also has signed the CRL".

2. Parsing multiple CRLs from a single file is gross.. I suppose it could use a contribution to `x509.ParseCRL(...)` method to parse multiple PEMs, similarly to `func (s *CertPool) AppendCertsFromPEM(pemCerts []byte)`

I'd love your thoughts/help 🙏

hkdsun: Hey @dbussink, I did some more research and I'm happy enough to bring that commit in.

It seems like (1) is a safe assumption (for example, I see a team of security experts making a similar assumption in [this codebase](https://github.com/cossacklabs/acra/blob/a02704aff992b5b45b5b723960e58173f8483d87/network/crl.go#L527-L537)) and (2) can't easily be avoided.

Would you kindly review again?

PR #8812, opened 2021-09-14
Gen4: added all valid plan output

harshit-gangal:
## Description

Updated all valid plan output

## Related Issue(s)

#7280 

## Checklist

### DIFF
--- a/go/vt/vtgate/planbuilder/testdata/unsupported_cases.txt
+++ b/go/vt/vtgate/planbuilder/testdata/unsupported_cases.txt
@@ -24,18 +24,11 @@ Gen4 plan same as above
 # scatter order by with * expression
 "select * from user order by id"
 "unsupported: in scatter query: order by must reference a column in the select list: id asc"
+Gen4 error: unsupported: '*' expression in cross-shard query
 
 # order by rand on a cross-shard subquery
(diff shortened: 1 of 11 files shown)


go/vt/vtgate/planbuilder/testdata/unsupported_cases.txt
```diff
@@ -354,6 +349,7 @@ Gen4 plan same as above
 "unsupported: where clause for vindex function must be of the form id = <val> (rhs is not a value)"
 Gen4 plan same as above
 
+# vindex func without where condition
 "select keyspace_id from user_index"
 "unsupported: where clause for vindex function must be of the form id = <val> (where clause missing)"
```
frouioui, line 355: Gen4 does not produce a plan for this?

frouioui (approved):

go/vt/vtgate/planbuilder/testdata/unsupported_cases.txt
harshit-gangal: It should be not supported

frouioui: We should produce the same error as v3 though

PR #8821, opened 2021-09-15
Changed parsing for unions

ritwizsinha:
## Description

This pull request, makes changes to the select and union grammar in Vitess to match what is implemented in the [MYSQL](https://github.com/mysql/mysql-server/blob/8.0/sql/sql_yacc.yy) grammar. Here are the changes that have been made

- The parser no longer accepts queries with limit/lock/order by in select with unions without parenthesis for example
```
select 1 order by 1 union select 2 order by 2
```
is a parsing error and this type of query is only accepted if it is parenthesized
```
(select 1 order by 1) union (select 2 order by 2)
```
- We no longer need to have parenthesis in any other select statement apart from the ones mentioned above when working with unions
- We no longer use ParenSelect struct to put parenthesis around statements but we check if the statement needs parenthesis in the ast_format file, and put parenthesis when needed. All utility functions for ParenSelect have been removed
- The Select struct now accepts an additional field of into clause which is checked and passed when present as specified in the grammar
- The union is now stored recursively with first statement field specifying the left select statement and the right side of union being specified by the UnionSelects field
- Introduced precedence symbols SUBQUERY_AS_EXPR and EMPTY_FROM_CLAUSE to avoid ambiguity in grammar while parsing subqueries and into selects respectively. The same trick has been used and explained in the MYSQL grammar
- Updated the parser and planner tests to reflect the current output of parser.

## Checklist


go/vt/sqlparser/ast_funcs.go
```diff
@@ -872,6 +853,11 @@ func (node *Union) SetLock(lock Lock) {
 	node.Lock = lock
 }
 
+// SetInto sets the into clause
+func (node *Union) SetInto(into *SelectInto) {
```
systay, line 857: this does not look correct. there is no INTO on the UNION clause, is there?

GuptaManan100:
```sql
MySQL root@(none):test> (select 1) union (select 2) limit 1 into @myvar
Query OK, 1 row affected
Time: 0.001s
```
MySQL allows into in a union clause too

systay: TIL there is INTO on UNION

GuptaManan100 (approved):
LGTM! 🤟

go/vt/sqlparser/sql.y
```diff
@@ -513,59 +517,127 @@ load_statement:
...
+  }
+| openb query_expression locking_clause closeb
+  {
+    setLockInSelect($2, $3)
+    $$ = $2
+  }
 
-select_statement:
-  base_select order_by_opt limit_opt lock_opt into_option
+// TODO; (Manan, Ritwiz) : Use this in create, insert statements
```
systay, line 535: should this comment be addressed before merging the PR?

GuptaManan100: This PR is already too big, so I don't think we want to do it as part of this PR, but we should do it later. Basically, right now we use the `select_statement` grammar rule in create table or insert into type queries, but `select_statement` also includes selects with into clause which is wrong and should not be passed. So right now we are parsing more in create and insert than we should. We added the TODO to keep track of this, but we will do this change when we also look at the parsing of the create and insert statements and match MySQL's behaviour

systay (approved):

PR #8822, opened 2021-09-15
Gen4: improvement of `joinTree` creation

frouioui:
## Description

This pull request contains a bug fix for #8813. The bug happens when we join two `queryTree`s to create a `joinTree`.

When pushing a join predicate on a `joinTree` (`pushJoinPredicate`), we're looking in the predicate for a `ColName` that `LHS` solves, once we find the `ColName`, we add it to `LHS`'s `columns`, and we add it into our `joinTree`'s `JoinVars`.

Then, we create an argument for this new column, along with a new predicate containing the argument and the second column of our join predicate, which is solved by `RHS`. This new predicate is pushed to `RHS` using the same function `pushJoinPredicate`.

However, if `RHS` is also a `joinTree` we try to push the predicate as an output column, meaning that we will create new output columns in `RHS`, and this is done in a recursive fashion (i.e if RHS's RHS is also a `joinTree`). Instead, we can add the predicate to the first route we find in `RHS` that solves it, avoiding the creation of unrequired output columns.

## Related Issue(s)
- #7280
- Fixes #8813 

## Checklist
- [x] Tests were added or are not required
- [x] Documentation was added or is not required

PR #8828, opened 2021-09-16
Apply goimports before checking make sizegen

frouioui:
## Description

After having issues with `check_make_sizegen` due to a difference between the source file (on which `goimports` was applied) and the file generated by Jennifer in the import statement, [CI builds](https://github.com/planetscale/vitess/runs/3617929606) started failing in [this pull request](https://github.com/vitessio/vitess/pull/8822) with the following error:

```
--- FAIL: TestFullGeneration (1.66s)
    sizegen_test.go:31: 
        	Error Trace:	sizegen_test.go:31
        	Error:      	Should be empty, but was ['/home/runner/work/vitess/vitess/go/tools/sizegen/integration/cached_size.go' has changed]
        	Test:       	TestFullGeneration
```

To alleviate this issue, this pull request changes how the `TestFullGeneration` test compares the local and generated files. After Jennifer has generated the file, we write its content to a temporary file, execute `goimports` on it, and finally, we run the comparison between the local and generated files. This ensures that we compare local files against `goimports`'ed files. And to make sure our local files are always `goimports`'ed, this pull request adds a new workflow (`check_imports`) that fails if a file is not `goimports`'ed.

Adding this new workflow requires the installation of `goimports`, to do that, it was necessary to upgrade `setup-go` action from `v1` to `v2` as explained in this issue: https://github.com/actions/setup-go/issues/27.

## Checklist


.github/workflows/check_imports.yml
```diff
@@ -0,0 +1,25 @@
+name: check_imports
+on: [pull_request]
```
GuptaManan100, line 2: We should run this on push too like the other tests

GuptaManan100 (approved):
Everything else LGTM! 🍰

.github/workflows/check_imports.yml
frouioui: Definitely, let me add that right away

GuptaManan100 (approved):

vmg (approved):

PR #8830, opened 2021-09-16
Gen4: improve analyzer tests

frouioui:
## Description

This pull request adds new tests and benchmarks to the analyzer.

## Related Issue(s)
- #7280 

## Checklist

### DIFF
--- a/go/vt/vtgate/semantics/analyzer_test.go
+++ b/go/vt/vtgate/semantics/analyzer_test.go
@@ -590,9 +590,6 @@ func TestOrderByBindingTable(t *testing.T) {
 	}, {
 		"select id from t1 union select uid from t2 union (select name from t) order by 1",
 		T1 | T2 | T3,
-	}, {
-		"select a.id from t1 as a union (select uid from t2) order by 1",
-		T1 | T2,
 	}}
 	for _, tc := range tcases {
 		t.Run(tc.sql, func(t *testing.T) {
--- a/go/vt/vtgate/semantics/analyzer_test.go
+++ b/go/vt/vtgate/semantics/analyzer_test.go
@@ -778,8 +778,8 @@ func TestUnionWithOrderBy(t *testing.T) {
 
 	stmt, semTable := parseAndAnalyze(t, query, "")
 	union, _ := stmt.(*sqlparser.Union)
-	sel1 := union.FirstStatement.(*sqlparser.Select)
-	sel2 := union.UnionSelects[0].Statement.(*sqlparser.ParenSelect).Select.(*sqlparser.Select)
+	sel1 := sqlparser.GetFirstSelect(union)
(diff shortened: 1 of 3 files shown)


go/vt/vtgate/planbuilder/testdata/aggr_cases.txt
```diff
@@ -2150,7 +2150,7 @@ Gen4 error: In aggregated query without GROUP BY, expression of SELECT list cont
 # Group by invalid column number (code is duplicated from symab).
 "select id from user group by 1.1"
 "column number is not an int"
-Gen4 error: Expression of SELECT list is not in GROUP BY clause and contains nonaggregated column 'id' which is not functionally dependent on columns in GROUP BY clause; this is incompatible with sql_mode=only_full_group_by
+Gen4 error: error parsing column number: 1.1
```
systay, line 2153: Why is this failing? It's a valid (although stupid) query, right?

go/vt/vtgate/semantics/analyzer_test.go
```diff
@@ -337,385 +745,62 @@ func TestUnion(t *testing.T) {
...
-		Name: sqlparser.NewTableIdent("b"),
-	}
+func TestUnionWithOrderBy(t *testing.T) {
+	query := "select col1 from tabl1 union (select col2 from tabl2) order by 1"
 
-	parse, _ := sqlparser.Parse(query)
+	stmt, semTable := parseAndAnalyze(t, query, "")
+	union, _ := stmt.(*sqlparser.Union)
+	sel1 := union.FirstStatement.(*sqlparser.Select)
+	sel2 := union.UnionSelects[0].Statement.(*sqlparser.ParenSelect).Select.(*sqlparser.Select)
```
systay, line 791: `sqlparser.GetFirstSelect()`

frouioui: good catch

go/vt/vtgate/planbuilder/testdata/aggr_cases.txt
frouioui: We had an unused if statement to deal with this kind of error in the analyzer's `earlyRewrite`. I thought it would be more efficient to fail early instead of delegating the work to later parts of the planner. V3 fails with the same kind of error

systay (changes requested):
Working on it with @frouioui

go/vt/vtgate/planbuilder/testdata/aggr_cases.txt
systay: The old error is the correct one. `order by 1.1` is perfectly valid and should not fail. the missing `group by` is the real error here

systay (approved):

PR #8836, opened 2021-09-16
Use vt_dba user for online schema migration cutover phase

### LINKED ISSUE #8835: Online schema change cutover phase table rename done with vt_app instead of vt_dba user
#### Overview of the Issue

In online schema change in the cutover phase, the table rename is currently done using the vt_app user instead of vt_dba. In our setup at Stripe (and I imagine others as well), we do not grant the ALTER TABLE permission to this user, that being reserved for vt_dba.

Can we run this step using vt_dba instead?

utk9:
## Description
In online migration cutover phase, run the `RENAME TABLE` operation using the vt_dba user instead of vt_app.

## Related Issue(s)
Fixes #8835 

## Checklist


deepthi (approved):
LGTM. Thank you for fixing this.

PR #8841, opened 2021-09-17
add weight_string function to group by if it is added to select list

harshit-gangal:
## Description

This change to make the addition of weight_string function to select list to be compatible with sql_mode = full_group_by
by adding to the group by list.

## Related Issue(s)

#7280 

## Checklist


GuptaManan100 (approved):
Should we add an end to end test for the query which wasn't working earlier?

PR #8844, opened 2021-09-17
Re-implement vttablet periodic refresh of DNS hostnames 

aquarapid:
This feature was implemented in 2019 via the vttablet option `-pool_hostname_resolve_interval`.  However, this only applied to pools maintained via `go/vt/dbconnpool`;  not to the pools using the `go/pools/resource_pool` abstraction.

This PR reworks that mechanism to generalize `resource_pool` to allow supplying a refresh check function, which can then be used by clients to achieve this functionality.  Also reworked the original code in `go/vt/dbconnpool` to use the same shared scaffolding.

One large difference in behavior is that previously we would reconnect to the new IP only when a connection is fetched from the pool after a DNS change was detected.  We now drain and re-open the pool when an IP change is detected instead.  Arguably,  this can be more disruptive, but is simpler both conceptually and in implementation.  Lastly, since some pools just don't have any refreshing today, this is still an improvement over that.


systay:
@aquarapid let us know when you want us to look at this

aquarapid:
Ready for review.  I couldn't think of a way to add proper tests for this short of introducing some test-only fixtures into the actual implementation.  It has been manually tested quite a bit though;  and it is definitely no worse than the current situation (where you don't get any effective DNS-based failover).

Just a few comments here for future explorers RE Aurora's failover mechanism (which is the major use-case for this functionality):
  * When you "failover" in an Aurora writer endpoint/instance (manually, since that's all that's really available to test), the following happens:
    * Initially, there is some delay while Aurora does some stuff internally.  This delay can be quite variable;  as short as a few seconds, as long as a minute or more.  During this time, processing against the Aurora cluster is unaffected.
    * Next, all connections to the writer endpoint are terminated.  As this point, your application (or Vitess, if you have Vitess in front of Aurora) will see write errors complaining about the inability to connect to the database.
    * After a few seconds (again, seems to be somewhat variable), the writer endpoint will become available again.  At this point your application (or Vitess) can reconnect and try write transactions again.  However, the instance is now read-only (having gone from primary to replica), and your writes will get errors to that effect (`Error Code: 1290. The MySQL server is running with the –read-only option so it cannot execute this statement`).
    * At this point, the DNS record for the writer endpoint has not been (visibly) updated.  So even with a working DNS refresh implementation, the client (or Vitess) is not aware of the failover yet.
    * Some time later (5-15 seconds);  the DNS change becomes visible.

So, with the changes in this PR, the following phases of behavior will be seen on failover of the primary/writer instance with Vitess in front of it:
  1. Application queries will fail with vttablet errors about being unable to connect to the database
  1. Shortly after this, the connection pools will reconnect, but still see write errors (database in read-only)
  1. Then, the DNS refresh will kick in, draining all the pools, opening new connections to the newly promoted writer instance, and the write errors will subside.

Note that a "real" failover (e.g. if the primary/writer instance dies) is likely to behave differently, but we can't really test this.  I would assume that you might not see phase 2 above, and just go straight from phase 1 to 3.

go/netutil/netutil.go
```diff
@@ -190,6 +191,64 @@ func ResolveIPv4Addrs(addr string) ([]string, error) {
 	return result, nil
 }
 
+func dnsLookup(host string) ([]net.IP, error) {
+	addrs, err := net.LookupHost(host)
+	if err != nil {
+		return nil, fmt.Errorf("Error looking up dns name [%v]: (%v)", host, err)
```
systay, line 197: `vterrors.Errorf` instead of `fmt.Error` please

go/netutil/netutil.go
```diff
@@ -190,6 +191,64 @@ func ResolveIPv4Addrs(addr string) ([]string, error) {
...
+		return bytes.Compare(naddr[i], naddr[j]) < 0
+	})
+	return naddr, nil
+}
+
+// DNSTracker is a closure that persists state for
+//  tracking changes in the DNS resolution of a target dns name
+//  returns true if the DNS name resolution has changed
+//  If there is a lookup problem, we pretend nothing has changed
+func DNSTracker(host string) func() (bool, error) {
```
systay, line 213: shouldn't we fail if the host we get is empty? looks to me like we would just always return false from the returned function

go/vt/vttablet/tabletserver/connpool/pool.go
```diff
@@ -109,7 +113,15 @@ func (cp *Pool) Open(appParams, dbaParams, appDebugParams dbconfigs.Connector) {
...
 		return NewDBConn(ctx, cp, appParams)
 	}
-	cp.connections = pools.NewResourcePool(f, cp.capacity, cp.capacity, cp.idleTimeout, cp.prefillParallelism, cp.getLogWaitCallback())
+
+	var refreshCheck pools.RefreshCheck
+	if net.ParseIP(appParams.Host()) == nil {
+		refreshCheck = netutil.DNSTracker(appParams.Host())
+	} else {
+		refreshCheck = nil
+	}
```
frouioui, line 122: nit: we can remove this `else` statement

go/pools/resource_pool.go
```diff
@@ -142,9 +156,40 @@ func NewResourcePool(factory Factory, capacity, maxCap int, idleTimeout time.Dur
...
+		defer rp.refreshWg.Done()
+		for {
+			select {
+			case <-rp.refreshTicker.C:
+				val, err := rp.refreshCheck()
+				if err != nil {
+					log.Info(err)
+				}
+				if val {
+					go rp.reopen()
```
systay, line 183: I don't see anything stopping two goroutines both running `rp.reopen()` at the same time. one could start the reopen, and later a second attempt might make it in before the first one is finished

should we protect against this with a mutex or something?

go/pools/resource_pool_flaky_test.go
```diff
@@ -427,30 +427,67 @@ func TestClosing(t *testing.T) {
...
+	lastID.Set(0)
+	count.Set(0)
+	refreshCheck := func() (bool, error) {
+		return true, nil
+	}
+	p := NewResourcePool(PoolFactory, 5, 5, time.Second, 0, logWait, refreshCheck, 500*time.Millisecond)
+	var resources [10]Resource
+	for i := 0; i < 5; i++ {
+		r, err := p.Get(ctx)
+		if err != nil {
```
systay, line 454: we like testify. 
`require.NoError(t, err)`

go/pools/resource_pool_flaky_test.go
```diff
@@ -427,30 +427,67 @@ func TestClosing(t *testing.T) {
...
+			t.Errorf("Get failed: %v", err)
+		}
+		resources[i] = r
+	}
+
+	time.Sleep(10 * time.Millisecond)
+	stats := p.StatsJSON()
+	expected := `{"Capacity": 5, "Available": 0, "Active": 5, "InUse": 5, "MaxCapacity": 5, "WaitCount": 0, "WaitTime": 0, "IdleTimeout": 1000000000, "IdleClosed": 0, "Exhausted": 1}`
+	if stats != expected {
```
systay, line 463: `assert.Equal(t, expected, stats)`

systay (approved):
Some tweaks we could add to make this even more stable, but it could go in as is and then get improved on. It's definitely better than what we had before

go/netutil/netutil.go
aquarapid: That is by design,  we would not refresh the DNS in that case;  and accordingly some type of DNS issue that results in an empty record would not result in us reopening the pool, possibly ending up with a non-reachable backend DB.

go/pools/resource_pool.go
aquarapid: Yeah, it's possible in theory;  but since reopen() itself shuts down the refresh ticker almost immediately, it's probably highly unlikely to get executed again until reopen restarts the ticker and exits.  I'll think about it, I guess a local mutex in this closure will do the trick.

go/vt/vttablet/tabletserver/connpool/pool.go
aquarapid: yup.  For some reason my IDE insisted on this check;  I must have had some setting enabled.  I will yank it.

systay (approved):
Nice 👍

PR #6630, opened 2020-08-26
Change GRPC message limit SQL error type

setassociative:
## Summary

There are a collection of errors bundled under the generic GRPC ResourceExhausted error. Several of them are not necessarily related to resource exhausted but to limits that are being exceeded in a way that backing off and waiting for reduced load will not resolve.

This change introduces a processing step in the GRPC -> mysql error conversion that rewrites those errors from the current `1203 - ERTooManyUserConnections` to a (somewhat) more appropriate `1153 - ERNetPacketTooLarge`. Two other errors are also recategorized in the vtgate executor to be treated as the 1153 instead of 1203.

The three errors being changed:

- When a GRPC message coming back to the vtgate exceeds the configured size limit (configured via `grpc_max_message_size`)
- If a write is attempting to make a change with a payload that is larger than the configured allowable size
- if a scatter query is attempting to collect process too many rows in memory within the vtgate to construct a response

Before this change each of these would the SQL error `ERTooManyUserConnections` when using the MySQL interface. This is a fairly unfortunate conflation of that error and doesn't make sense to be consumed as "too many users" when the actual error is various flavors of "too much data."

Further in a situation where you want to use error codes to apply back pressure to callers having these two classes of error combined means that it's not a great signal.

We discussed briefly whether we should _also_ change the GRPC error so that ResourceExhausted isn't overloaded but it seems that those errors need to match GRPC error set ([ref](https://github.com/vitessio/vitess/blob/f886cc2681844f3bec3dd3e9b248e22b24c30a92/proto/vtrpc.proto#L53-L55)). Additionally it would be a much larger change because ResourceExhausted is used for some session management. If we decide later to revisit this decision that will be fine as the GRPC and SQL errors are independent (also, even if it does feel like what we'd want to do I would prefer it to be a separate change from this from a pragmatic standpoint).

### Misc

1️⃣  This change exceeds our initial plan which covered only the GRPC message limit case and I'm open for pushback on the other two but I do think that all these cases are better suited as 1153 than 1203.

2️⃣ The error message checks are _extremely factored_ and I'm happy to pull them out of `vterrors` for another home or deal with that differently if folks have opinions. I didn't know what was the most reasonable location since they needed to be shared between sql_errors and the vtgate

## Testing

- Added unit tests for the method that processes the resource exhausted errors
- Built and deployed a modified vtgate and observed expected (new) error code when running with an artificially low grpc message limit (3k). Note `ERROR 1153` in the following:

```
mysql> select id, data from test_table limit 100;
[...]

PR #8853, opened 2021-09-20
Handle mapping of gRPC packet overflows to `ERTooManyUserConnections` for SELECT cases too

### LINKED ISSUE #8852: Selecting too much data that overflows the gRPC packet limit still reports `ERTooManyUserConnections`
#6630 was supposed to address this, but the regex used to match and map the gRPC error message seems to have only handled the case where too much data was being inserted, **not** selected.

As a result, on current main, if you have a table with wide columns, I see:

```
mysql> select * from t1 limit 900;      
ERROR 1203 (42000): target: keyspace1.0.primary: vttablet: rpc error: code = ResourceExhausted desc = grpc: trying to send message larger than max (18015277 vs. 16777216)
mysql> 
```

Error 1203 is `ERTooManyUserConnections`;  we want Error 1153 (`ERNetPacketTooLarge`).

aquarapid:
Fixes #8852 

See original PR #6630

This is a potential breaking change for those that depended on the old behavior of `ERTooManyUserConnections` being returned on a too-large `SELECT`, and mapped that to some application-side action.

Not addressed in this PR, but we should probably add an end-to-end test here to catch if/when the upstream gRPC error message is ever changed, since we track upstream gRPC now after having been frozen for years.

### DIFF
--- a/go/mysql/sql_error.go
+++ b/go/mysql/sql_error.go
@@ -214,7 +214,7 @@ func convertToMysqlError(err error) error {
 	return NewSQLError(mysqlCode.num, mysqlCode.state, err.Error())
 }
 
-var isGRPCOverflowRE = regexp.MustCompile(`.*grpc: received message larger than max \(\d+ vs. \d+\)`)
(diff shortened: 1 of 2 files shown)


deepthi (approved):
LGTM

systay:
Is this ready to merge?

PR #8854, opened 2021-09-21
Move unit test race to self hosted runners

frouioui:
## Description

This pull request moves the `unit_race` CI workflow to the self-hosted runners in order to reduce its flakiness.

## Checklist


GuptaManan100 (approved):
LGTM! Unit race is not generated via ci_workflow_gen.go, so you had to do it by hand. Sorry about that...

PR #8858, opened 2021-09-21
Gen4: VTGate's end-to-end tests to use `gen4`

frouioui:
## Description

This pull request introduces a new query planner, `Gen4CompareV3`, whose only goal is to execute queries using both `Gen4` and `V3` planners and compare their results. Following the comparison, an error is returned if the result sets are different, otherwise, results are returned normally to the client. Enabling the use of `Gen4CompareV3` is done by setting VTGate's `-planner_version` flag to `Gen4CompareV3`.

This new feature is now enabled on all end-to-end tests of the `./go/test/endtoend/vtgate/...` packages. While adding this new feature, a few tests started failing due to minor issues which are detailed in the next sub-section. Unsupported cases are detailed in the `Unsupported cases` sub-section.

### Bug fixes

While establishing this pull request, a couple of issues were found in both the AST-clone and `queryTree`, they all got fixed as part of this pull request. Those issues were:

- In the AST clone, cloning a `nil` slice resulted in the creation of a new slice of length 0 (non-nil), leading to invalid behaviors when the slice's emptiness was checked using `if slice == nil { }` instead of the `len(slice)` method.
- In the AST clone, cloning a named type with its underlying type being a slice of literals was done using `copy(dest, src)`, which does not work when passing a named type as one of the arguments. This was fixed by using a for loop instead.
- Using joins and derived tables, when a join needs to push output columns to its child `simpleProjection`, columns were not pushed correctly, leading to the `Join` trying to access a column from `SimpleProjection` that does not exist in the result set.

### Mismatch between `Gen4` and `V3`

If the results of the two executions are unequal, their difference is logged as a warning by VTGate and the client receives the following SQL error: `"results did not match, see VTGate's logs for more information"`. Below is a sample of the difference output:

```
W0923 09:06:58.416658   24777 gen4_compare_v3.go:130] Results of Gen4 and V3 are not equal. Displaying diff.
W0923 09:06:58.417063   24777 gen4_compare_v3.go:137] Gen4's plan:
{
  "QueryType": "SELECT",
  "Instructions": {
    "OperatorType": "Route",
    "Variant": "SelectNext",
    "Keyspace": {
      "Name": "uks",
      "Sharded": false
    },
    "FieldQuery": "select next :vtg1 values from sequence_test_seq where 1 != 1",
    "Query": "select next :vtg1 values from sequence_test_seq",
    "Table": "sequence_test_seq"
  }
}
W0923 09:06:58.417091   24777 gen4_compare_v3.go:144] V3's plan:
{
  "QueryType": "SELECT",
  "Instructions": {
    "OperatorType": "Route",
    "Variant": "SelectNext",
    "Keyspace": {
      "Name": "uks",
      "Sharded": false
    },
    "FieldQuery": "select next :vtg1 values from sequence_test_seq where 1 != 1",
    "Query": "select next :vtg1 values from sequence_test_seq",
    "Table": "sequence_test_seq"
  }
}
W0923 09:06:58.417102   24777 gen4_compare_v3.go:146] Gen4's results:
W0923 09:06:58.417111   24777 gen4_compare_v3.go:147]   [rows affected: 0]
W0923 09:06:58.417120   24777 gen4_compare_v3.go:149] 	[INT64(5)]
W0923 09:06:58.417132   24777 gen4_compare_v3.go:151] V3's results:
W0923 09:06:58.417140   24777 gen4_compare_v3.go:152] 	[rows affected: 0]
W0923 09:06:58.417147   24777 gen4_compare_v3.go:154] 	[INT64(6)]
W0923 09:06:58.417157   24777 gen4_compare_v3.go:156] End of diff.
```

### Unsupported cases

Only one end-to-end test is failing due to Gen4 not supporting `SQL_CALC_FOUND_ROWS` yet. This test has been skipped for now.

> **to do:**
> - [x] Add a note to remove the `t.Skip` for `SQL_CALC_FOUND_ROWS` tests in the corresponding card/issue.

## Related Issue(s)

- Tracking #7280 
- Documentation https://github.com/vitessio/website/pull/832

## Checklist


go/test/endtoend/vtgate/errors_as_warnings/main_test.go
```diff
@@ -90,6 +90,7 @@ func TestMain(m *testing.M) {
 		}
 
 		// Start vtgate
+		clusterInstance.VtGateExtraArgs = []string{"-planner_version", "Gen4CompareV3"}
```
systay, line 93: instead of setting this manually everywhere, how about making this the default?

go/vt/vtgate/engine/gen4_compare_v3.go
```diff
@@ -0,0 +1,183 @@
...
+// Gen4CompareV3 is a Primitive used to compare V3 and Gen4's plans.
+type Gen4CompareV3 struct {
+	V3, Gen4   Primitive
+	HasOrderBy bool
+}
+
+var _ Primitive = (*Gen4CompareV3)(nil)
+var _ Gen4Comparer = (*Gen4CompareV3)(nil)
+
+// GetGen4Primitive implements the Gen4Comparer interface
```
systay, line 38: nit: since the field is public, and this is not an interface, might as well access the field directly, right?

systay (approved):
😍 this is really awesome!

go/test/endtoend/vtgate/errors_as_warnings/main_test.go
harshit-gangal: I belive you mean default for test only.

harshit-gangal: sequence tests will fail, because of the reason, it will be the next value always.

frouioui: Queries that use `next` (sequences) are ran against Gen4 only by design

go/test/endtoend/tabletgateway/buffer/buffer_test_helpers.go
```diff
@@ -246,6 +246,7 @@ func (bt *BufferingTest) createCluster() (*cluster.LocalProcessCluster, int) {
 	clusterInstance.VtGateExtraArgs = append(clusterInstance.VtGateExtraArgs, bt.VtGateExtraArgs...)
 
 	// Start vtgate
+	clusterInstance.VtGatePlannerVersion = 0
```
systay, line 249: Eeek. Why not use `planbuilder.V3` instead?

frouioui: `clusterInstance.VtGatePlannerVersion = 0` results in using VTGate's default, once we upgrade to `Gen4` as the default planner, we won't have to change this

frouioui:
```go
if vtgate.PlannerVersion > 0 {
 	args = append(args, "-planner_version", vtgate.PlannerVersion.String())
}
```

Issue #4456, opened 2018-12-14
what different between SplitCloneWorker and LegacySplitCloneWorker


justlooks:
i can not see any difference between them

```
// SplitCloneWorker will clone the data within a keyspace from a
// source set of shards to a destination set of shards.

// LegacySplitCloneWorker will clone the data within a keyspace from a
// source set of shards to a destination set of shards.
```

mattlord:
I'm closing this as the `LegacySplitCloneWorker` is no longer used in any code paths and we're going to remove that (dead) code.

PR #8867, opened 2021-09-22
LegacySplitCloneWorker is no longer used in any code paths

mattlord:

## Description
Keeping things tidy...

## Related Issue(s)
https://github.com/vitessio/vitess/issues/4456

## Checklist


rafael:
Wow - end of an era! Excited to see this go. For completeness, we should not allow `LegacySlitClone` as an option from [here](https://github.com/vitessio/vitess/blob/main/go/vt/workflow/resharding/workflow.go#L77). 

Also, we should delete the resharding workflow entirely. But that could be a different PR.

mattlord:
> Wow - end of an era! Excited to see this go. For completeness, we should not allow `LegacySlitClone` as an option from [here](https://github.com/vitessio/vitess/blob/main/go/vt/workflow/resharding/workflow.go#L77).

I tried to remove the remaining references in https://github.com/vitessio/vitess/pull/8867/commits/bfcfcb9e53f2ecd3a77fc6f23cbfc3638af06568

Let me know if you had something else in mind. 
 
> Also, we should delete the resharding workflow entirely. But that could be a different PR.

I'm actually not sure what you mean here and how it's related. 

Thanks!

mattlord:
> Also, we should delete the resharding workflow entirely. But that could be a different PR.

@Rafael Were you saying that we should remove the non-legacy `SplitClone` code too [now that you should be using `MoveTables` isntead now after 6.0](https://vitess.io/docs/user-guides/historical/vertical-split/)? If so, then I think we probably should but can be done in a separate PR.

rafael (approved):
LGTM

PR #8870, opened 2021-09-23
gen4: Union rewrite

systay:
## Description
This PR changes a few things around UNION.

First the AST struct for UNION changes to only have two children. The parser was already producing UNIONs with two children, but the struct used a slice for the right hand side.

The operator and query tree for UNION have also been simplified.

Gen4 now correctly fails early for queries that are not supported

## Related Issue(s)
#7280 

## Checklist


GuptaManan100 (approved):
looks good to me!

frouioui (approved):
Looking good

Issue #8729, opened 2021-08-30
Flyway query on information schema stopped working


nmonserrato:
#### Overview of the Issue

In my company we're using the `vttestserver` docker image with testcontainers and flyway for the acceptance test (a simple 1 shard 1 keyspace approach).
Since a few days ago (we noticed it on Friday 27 Aug 2021), one of the queries Flyway performs to verify if our database exists fails all the time.

```
SELECT (SELECT 1 FROM information_schema.schemata WHERE schema_name='MyDatabase' LIMIT 1);
```

#### Reproduction Steps

Steps to reproduce this issue:

1. `docker run -p 3306:33577 -e PORT=33574 -e KEYSPACES="MyDatabase" -e NUM_SHARDS="1" -e MYSQL_BIND_HOST=0.0.0.0 vitess/vttestserver:mysql80`
2. execute the query above

(fails also with `vitess/vttestserver:mysql57`, works fine with vanilla `mysql:8.0`.

#### Operating system and Environment details

OS, Architecture, and any other information you can provide about the environment.

- Operating system (output of `cat /etc/os-release`): tried on ubuntu 21, OSX 11 and CentOS
- Kernel version (output of `uname -sr`): Linux 5.11.0-31-generic
- Architecture (output of `uname -m`): x86_64

#### Log Fragments

on mysql:
```
mysql> SELECT (SELECT 1 FROM information_schema.schemata WHERE schema_name='MyDatabase' LIMIT 1);
ERROR 1105 (HY000): target: MyDatabase.0.primary: vttablet: rpc error: code = InvalidArgument desc = missing bind var __vtschemaname (CallerID: userData1)
```

from the container:
```
: Sql: "select (select :vtg1 from information_schema.schemata where schema_name = :__vtschemaname limit :vtg1) from dual", BindVars: {#maxLimit: "type:INT64 value:\"10001\""vtg1: "type:INT64 value:\"1\""vtg2: "type:VARBINARY value:\"MyDatabase\""}
```

PR #8871, opened 2021-09-23
fix merging dba query as a subquery both in v3 and gen4

GuptaManan100:
## Description
SelectDBA queries were not being merged properly when occurring as subqueries. This PR fixes that both for V3 and Gen4 planners.

## Related Issue(s)
- Fixes https://github.com/vitessio/vitess/issues/8729

## Checklist
- [x] Tests were added or are not required Added
- [x] Documentation was added or is not required Not required


harshit-gangal (approved):
can you add a case for dba query as derived table, if it does not exists

GuptaManan100:
Sure, I added it in the commit - 5887a34

frouioui (approved):

PR #8872, opened 2021-09-23
Backtick DB and table names in RoutingRules

msolters:
Signed-off-by: Mark Solters <msolters@gmail.com>

## Description
For some data storage engines, such as Oracle MySQL 8, users will encounter errors if their tables are named after reserved keywords (e.g. `lead`).  This is not an issue if references to these tables are properly escaped.

Currently, the materialize code paths which mutate routing rules do not respect the correct escaping (backticks).  We have independently verified that we get query routing failures when e.g. migrate commands are run, and un-backticked schema/tables are put into RoutingRules.  Manually pulling the RoutingRules as JSON, adding the backticks where appropriate, and manually re-applying fixes these problems.

These changes fix the problem at the source by forcing the Materialize code paths to backtick all database names or table names.

## Checklist

## Deployment Notes

### DIFF
--- a/go/vt/wrangler/materializer.go
+++ b/go/vt/wrangler/materializer.go
@@ -165,16 +165,16 @@ func (wr *Wrangler) MoveTables(ctx context.Context, workflow, sourceKeyspace, ta
 			return err
 		}
 		for _, table := range tables {
-			toSource := []string{sourceKeyspace + "." + table}
-			rules[table] = toSource
-			rules[table+"@replica"] = toSource
-			rules[table+"@rdonly"] = toSource
-			rules[targetKeyspace+"."+table] = toSource
-			rules[targetKeyspace+"."+table+"@replica"] = toSource
-			rules[targetKeyspace+"."+table+"@rdonly"] = toSource
-			rules[targetKeyspace+"."+table] = toSource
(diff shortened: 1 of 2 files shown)


mattlord (changes requested):
The tests will also need to be updated here to include backticks around the entities: https://github.com/vitessio/vitess/blob/main/go/vt/vtctl/grpcvtctldserver/server_test.go#L211-L324

deepthi:
@sougou is there a cleaner way to do this?

deepthi:
> @sougou is there a cleaner way to do this?

Found it. We have a util function that does the backticking:
https://github.com/vitessio/vitess/blob/main/go/vt/mysqlctl/schema.go#L79

sougou:
This is my first pass. I'll need to do another one. But there is already code for handling backticks. You can see the test for it here: https://github.com/vitessio/vitess/blob/c225a6d4ae56dea823e2dd22c49562619bacc286/go/vt/vtgate/vindexes/vschema_test.go#L668-L669

And the PR for it is here https://github.com/vitessio/vitess/pull/7073

I'm guessing that it's missing the handling of your specific corner case. I'll take a look at what that is on my next pass.

PR #8873, opened 2021-09-23
Fix for padding in OrderAndCheckPartitions 

rafael:
## Description
We ran into an issue similar to the one in https://github.com/vitessio/vitess/pull/8296. While trying to switch traffic we found the following error:
```
E0922 08:33:05.346443    2565 main.go:67] E0922 15:33:05.346176 traffic_switcher.go:470] switchShardReads failed: partial result: non-contiguous KeyRange values for RDONLY 
```

This PR has three changes:
* adds a new function to the `key` package so it can check for contagious key ranges. It makes sure this function accounts for padding (in the same way as other key ranges comparison functions). 
* Starts using that function in `OrderAndCheckPartitions`.
* Refactors some code duplication in the tests.

## Related Issue(s)

* https://github.com/vitessio/vitess/pull/8296

## Checklist
- [] Should this PR be backported?
I don't know. Maybe?? I think this will only affect Slack. So I don't think there is a big pressure to back port this change. 

### DIFF
--- a/go/vt/key/key.go
+++ b/go/vt/key/key.go
@@ -242,9 +242,9 @@ func KeyRangeStartEqual(left, right *topodatapb.KeyRange) bool {
 	return bytes.Equal(addPadding(left.Start), addPadding(right.Start))
 }
 
-// KeyRangeIsContiguous returns true if the end of the left key range exactly
+// KeyRangeContiguous returns true if the end of the left key range exactly
 // matches the start of the right key range (i.e they are contigious)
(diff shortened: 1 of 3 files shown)


ajm188 (commented):
one minor copy-paste fix, then lgtm

ajm188 (approved):

deepthi (approved):
LGTM.
Is it possible that there are other places we could be doing this wrong? Wondering if you audited the code.

PR #8878, opened 2021-09-24
Gen4: SQL_CALC_FOUND_ROWS support

systay:
## Description
Adds support in the gen4 planner for SQL_CALC_FOUND_ROWS.
It does this by sharing most of the logic from the V3 planner.

## Related Issue(s)
#7280 

## Checklist

### DIFF
--- a/go/vt/vtgate/planbuilder/gen4_planner.go
+++ b/go/vt/vtgate/planbuilder/gen4_planner.go
@@ -42,27 +42,11 @@ func gen4Planner(query string) func(sqlparser.Statement, *sqlparser.ReservedVars
 				return p, err
 			}
 
-			if sel.SQLCalcFoundRows {
-				if sel.Limit != nil {
-					ksName := ""
-					if ks, _ := vschema.DefaultKeyspace(); ks != nil {
-						ksName = ks.Name
-					}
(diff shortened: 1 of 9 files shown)


frouioui (approved):
Looking good to me, that was the last unsupported feature that v3 supports, yay 🎊

go/vt/vtgate/planbuilder/gen4_planner.go
```diff
@@ -63,6 +69,34 @@ func gen4Planner(_ string) func(sqlparser.Statement, *sqlparser.ReservedVars, Co
...
+	return plan.Primitive(), nil
+}
+
+func planSelectGen4(reservedVars *sqlparser.ReservedVars, vschema ContextVSchema, sel *sqlparser.Select) (*jointab, logicalPlan, error) {
+	plan, err := newBuildSelectPlan(sel, reservedVars, vschema)
+	if err != nil {
+		return nil, nil, err
+	}
+	return nil, plan, nil
+}
```
GuptaManan100, line 98: This function is always returning nil for the joinTable, should we remove that return value?

GuptaManan100: nvm found why we have it this way

GuptaManan100 (approved):

Issue #8881, opened 2021-09-24
Add ability to run new RPCs/commands via the `vtctl` binary.


ajm188:
So far, the migration has been focused on the `vtctld` daemon's server api implementation, and the `vtctldclient` binary, which makes gRPC requests over a network to a running `vtctld` server.

The other mode of operation supported by the legacy api is to run a `vtctl` (no "d") binary with a single command. This mode would work as a sort of one-shot cli: it would spin up connections to the topo, and everything else needed to execute a single command, in-process (no network), and then exit. Logically (but not practically) equivalent to doing:
* `pid=$(vtctld <args> &)`
* `vtctlclient -server <addr_from_above> <command> <args...>`
* `kill "$pid"`

(Note these steps are meant to illustrate the concept; they are far too cumbersome to be an acceptable solution)

We need to support this mode as well.

PR #8882, opened 2021-09-24
[vtctld] localvtctldclient

ajm188:
## Description

This is the first chunk of work required for #8881. We are extending the codegen for the new vtctldclient interface to also support a "local" (inprocess, no network) client, which takes a vtctldserver implementation and calls methods directly on it, disregarding any `grpc.CallOptions` the caller passed.

The client methods all look like: `func (...) SomeRPC(ctx context.Context, req *vtctldatapb.RequestTypeT, opts ...grpc.CallOption) (*vtctldatapb.ResponseTypeT, error)` and the server methods all look like `func (...) SomeRPC(ctx context.Context, req *vtctldatapb.RequestTypeT) (*vtctldatapb.ResponseTypeT, error)`, so the adaptation from "client" to "server" is pretty straightforward.

After merging this, I'll open a second PR that updates the commands in `go/cmd/vtctldclient` to switch between `grpc` and `local` clients, which will let us use this code from the `vtctl` binary entrypoint.

**NOTE**: this PR changes the generator make target from `make grpcvtctldclient` to `make vtctldclient`, which generates both the `grpc` and `local` clients at once. I filed https://github.com/vitessio/vitess/projects/9#card-69491663 to follow up on the CI side to catch any issues related to these code generators.

## Related Issue(s)

#8881 

## Checklist
- [x] Tests were added or are not required **No**


deepthi (approved):
LGTM

PR #8884, opened 2021-09-25
Add parsing support for column list with subquery alias

### LINKED ISSUE #8883: Parser doesn't support column list with subquery alias
#### Overview of the Issue

The parser  tests fail when we give column list to the subquery alias. When we have a query such as 
```
select `name`, numbers from (select * from users) as x(`name`, numbers);
```
with the users table schema as
```
+---------+--------------+------+-----+---------+-------+
| Field   | Type         | Null | Key | Default | Extra |
+---------+--------------+------+-----+---------+-------+
| id      | varchar(255) | YES  |     | NULL    |       |
| rollnos | int          | YES  |     | NULL    |       |
+---------+--------------+------+-----+---------+-------+
```
while the same query is valid in MySQL

ritwizsinha:
## Description
 - Added support for parsing column list with subquery alias as specified [here](https://dev.mysql.com/doc/refman/8.0/en/join.html)
 - Added tests to check  the behaviour

## Related Issue(s)

Fixes #8883 

## Checklist


GuptaManan100 (approved):
LGTM!

systay (approved):

Issue #8145, opened 2021-05-19
flexible connecting pool


Johnny-Three:
#### Feature Description

flexible connection pool , that means , vitess can enlarge or shrink the mysql connecting pool size according the real  QPS load , which will make the pool size at a proper level.

PR #8885, opened 2021-09-25
RFC: Dynamic resource pool implementation

bnu0:
# Overview

Currently, the vitess connection pools hold open a constant number of connections (the `.Capacity`).  In a transitional environment where VTGate is not the primary means of connectivity to MySQL, this can be quite a lot of connections.  It can be helpful to allow the number of connections from vitess components to "float" between 0 and the `.Capacity` as needed.

Relates to:
- #8145 

# What does it do?

- This PR adds a different connection pool implementation, the `DynamicResourcePool`. 
- It refactors the base `ConnectionPool` to an interface and renames the old `ConnectionPool` to `StaticConnectionPool`. Both of these satisfy the interface.
- It adds the config/flags needed to turn on this implementation (off by default) for each pool.

# It's a draft...

I am testing this modification internally but wanted to open it up for comments.


bnu0:
I've marked this RFR because we've deployed it at scale (>1,000 clusters) and it seems to work and has been helpful for us, so wanted to solicit feedback if it interests the upstream community.

systay:
Thanks for the submission, @bnu0. Some of these files have changed in `main`. Mind rebasing or merging main?

harshit-gangal:
We should add minimum pool size ( or we can have minimum idle connections), taking pool connections all the way down to 0 does not seem a good idea, there will be a hit in latency for initial traffic load on every surge.

PR #8886, opened 2021-09-27
Fix Vtorc Flakiness in test suite

GuptaManan100:
Signed-off-by: Manan Gupta <manan@planetscale.com>

## Checklist


GuptaManan100:
Vtorc test only failed once in 6 runs which was a failure due to vttablet not starting within 10 seconds. Closing the PR since there is nothing to fix

PR #8895, opened 2021-09-28
Gen4: Keep track of left and right predicates in outer joins

frouioui:
## Related Issue(s)

- #7280 

## Checklist


systay:
These changes have been moved to #8876

PR #8901, opened 2021-09-29
Clean up Primitive interface implementations

systay:
## Description
Pure clean up. Addresses a lot of linting errors from the engine package.

## Checklist

### DIFF
--- a/go/vt/vtgate/engine/dbddl.go
+++ b/go/vt/vtgate/engine/dbddl.go
@@ -94,7 +94,7 @@ func (c *DBDDL) GetTableName() string {
 	return ""
 }
 
-// Execute implements the Primitive interface
+// TryExecute implements the Primitive interface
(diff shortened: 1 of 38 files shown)


go/vt/vtgate/engine/dbddl.go
```diff
@@ -195,7 +195,7 @@ func (c *DBDDL) GetFields(VCursor, map[string]*querypb.BindVariable) (*sqltypes.
 }
 
 // Description implements the Primitive interface
```
frouioui, line 197: nit: `Description` -> `description`

frouioui (approved):

go/vt/vtgate/engine/replace_variables.go
```diff
@@ -80,7 +80,7 @@ func (r *ReplaceVariables) Inputs() []Primitive {
 }
 
 // Description implements the Primitive interface
```
GuptaManan100, line 82: Some places we still have this comment, in others we do not. I do not have strong opinions about this. I am okay with it being the way it is or having it the same everywhere.

GuptaManan100 (approved):

PR #8902, opened 2021-09-29
perf: try reverting #8871

vmg:
## Description

There is a noticeable perf regression in AreWeFastYet:

image

The diff between the two measures is this: https://github.com/vitessio/vitess/compare/b04726d...7175f65

It looks like #8871 is the only thing that touches actual query serving, so let's attempt a revert of this.

## Checklist


frouioui:
Comparison for b6eabf1 is available here: https://benchmark.vitess.io/compare?r=b6eabf11d4eecac018caf7d8029891d7aa13ee3a&c=2991090def634f76c0c8617a65fb3b50daa24d75

PR #8905, opened 2021-09-29
gen4: implement a growable semantics.TableSet

vmg:
## Description

This was requested by @systay on Slack. We need a `TableSet` data structure that can support more than 64 tables. We're doing this right here in the smallest possible space: `TableSet` is now 16 bytes large and packs a small, single-word bitmap that is used when the set contains less than 64 tables, and a pointer to an implementation of an arbitrarily large set.

This implementation is pretty straightforward -- I don't believe it can be optimized further like we'd do in C/C++/Rust. For completeness, these kind of implementations are usually done like this in non-GC'ed languages:

Instead of the 16-byte `struct` we have here with `struct TableSet{ small uint64; large *largeTableSet }`, we'd have a single-field struct with a single word in it: `struct TableSet { set uintptr }`. When trying to read or write to this set, we'd check `if ts.set & 0x1 != 0`; if this is true, then we have a 63 bit tableset by doing `ts.set >> 1`. If it's false, then we'd have a large tableset by doing `(*largeTableSet)(ts.set)`. This works in practice because memory allocators will never return pointers with a 1-off alignment (their minimal alignment is at the very least the system's word size, 4/8), so we can store a bitmap in the pointer as long as the last bit of the bitmap is always `1`. :ok_hand: 

...Of course this doesn't work in a garbage collected language like Go because if you leave a `*largeTableSet` pointed at by an `uintptr`, the GC won't find it and it'll be freed on a later GC pass. So we don't get to to pack pointers and we need to run with 16-byte `TableSet`s. 

Good enough!

cc @systay @frouioui  

## Related Issue(s)
#7280 

## Checklist

### DIFF
--- a/go/vt/vtgate/semantics/tabletset.go
+++ b/go/vt/vtgate/semantics/tabletset.go
@@ -1,6 +1,23 @@
+/*
+Copyright 2021 The Vitess Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
 package semantics
 
 import (
+	"fmt"
 	"math/bits"
 )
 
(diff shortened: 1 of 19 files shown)


go/vt/vtgate/semantics/tabletset.go
```diff
@@ -0,0 +1,288 @@
+package semantics
+
+import (
```
frouioui, line 3: The header is missing

frouioui (commented):
The following errors appear when compiling:

```
abstract/concatenate.go:41:12: invalid operation: tableSet |= source.TableID() (operator | not defined on struct)
abstract/querygraph.go:81:29: invalid operation: lhs | rhs (operator | not defined on struct)
```

systay (approved):
Love it! Thank you thank you thank you

PR #8913, opened 2021-10-01
Gen4: derived table with aliased column names

frouioui:
## Description

The changes introduced by https://github.com/vitessio/vitess/pull/8884 allow us to parse derived tables' aliased column names, like: `select t.i from (select id from user) t (i)`, here the `id` inner column becomes `i` on the outside.

This pull request adds supports for this kind of queries in Gen4.

## Related Issue(s)

- #7280 
- https://github.com/vitessio/vitess/pull/8884

## Checklist


systay (approved):
Nice stuff

PR #8921, opened 2021-10-04
Reduce `Cluster 17` flakiness

frouioui:
## Description

This pull request tries to reduce the flakiness of `Cluster 17`'s tests. Instead of living in the `misc_test.go` file (at the root of vtgate's E2E tests package), tests using specific queries (like `UNION` queries) are moved to a separate package `vtgate/queries/{query type}`, this allows tests to have their own `clusterInstance` and not be impacted by other tests.

To enable faster builds since we create more `clusterInstance` the `cluster 17` has been moved to our self-hosted runners.

## Checklist


deepthi (approved):
:100: 
This was no doubt tedious. Thank you for doing it!

PR #8925, opened 2021-10-04
Fix `TestTopoCustomRule` flakiness

frouioui:
## Description

This pull request tries to remove the flakiness of `TestTopoCustomRule`. This test was failing due to port already being used, to reduce this error the topology E2E tests are now split across several CI clusters.

## Checklist


test/config.json
```diff
@@ -791,7 +791,7 @@
 			"Args": ["vitess.io/vitess/go/test/endtoend/topotest/zk2", "--topo-flavor=zk2"],
 			"Command": [],
 			"Manual": false,
-			"Shard": "25",
+			"Shard": "vtgate_topo_sk2",
```
deepthi, line 794: tiny typo: sk2 -> zk2

deepthi: Also seems like there should be a workflow file for this?

frouioui: sorry, I did not intend to do this change, I will revert it

frouioui: @deepthi, actually `Cluster 25` does not exist, is this normal?

deepthi (approved):

PR #8927, opened 2021-10-04
Renamed docker_test workflows

frouioui:
## Description

This pull request renames the `docker_test_1|2|3` GHA workflows to `docker_test_cluster_10|25|26` to improve clarity. Additionally a note regarding the GHA clusters numeration has been added in the CI Workflows generation file.

## Checklist

### DIFF
--- a/test/ci_workflow_gen.go
+++ b/test/ci_workflow_gen.go
@@ -41,6 +41,7 @@ const (
 )
 
 var (
+	// Clusters 10, 25 and 26 are executed by the docker_test_cluster 10, 25, 26 workflows.
 	clusterList = []string{
 		"11",
 		"12",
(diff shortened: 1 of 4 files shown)


frouioui:
@deepthi, we probably need to update the list of required checks on the project's settings

test/ci_workflow_gen.go
```diff
@@ -41,6 +41,7 @@ const (
 )
 
 var (
+	// Clusters 10, 25 and 26 are executed by the docker_test_cluster 10, 25, 26 workflows.
```
deepthi, line 44: should change 26 -> 27 in the list so that we don't have two clusters called 26 - docker_test_cluster_26 and `Cluster 26`

PR #8929, opened 2021-10-04
Add restore_extra_restart_args for restore

sonne5:

## Description

When we do a `RestoreFromBackup` mysqld is bring up with "--skip-grant-tables", "--skip-networking". However, depends on the settings in my.cnf, these two might not be enough. 

For example, some people might choose to set mysql with super_read_only, which means with current setup, the restore won't go through because we need to call `create table` in PopulateMetadataTables

This PR adds `restore_extra_restart_args` parameter, so that ppl can pass in any extra parameters to bring up mysqld (e.g., `--super-read-only=OFF`) 

## Related Issue(s)

^

## Checklist

### DIFF
--- a/go/vt/mysqlctl/backup.go
+++ b/go/vt/mysqlctl/backup.go
@@ -25,6 +25,8 @@ import (
 	"strings"
 	"time"
 
+	"vitess.io/vitess/go/flagutil"
+
 	"context"
 
 	"vitess.io/vitess/go/mysql"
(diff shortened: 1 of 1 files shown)


sonne5:
Gentle nudge on this PR, thanks! /cc @aquarapid

GuptaManan100:
Me and @deepthi discussed the use case for this PR and we think that we should always add `--super-read-only=OFF` like we do for `skip-grant-tables`. If the user has not set `super-read-only` to true, then adding that flag does not change anything, but if they have it to true like you do, then we will need to set it to false. So we should always do it and we don't need the extra flag for it.

The code for metadata tables in restore is future-proof in the sense that if there are new tables added to the metadata in a future release, having this code makes it so that you will able to use backups from a previous release without worrying about these metadata tables. This makes the process of upgrades much smoother, so we want to keep that code.

@sonne5 could you change the PR to add that argument directly instead of adding a new flag?

aquarapid:
@sonne5 we've moved this over to #9240 , so I'm going to close this one.

PR #8931, opened 2021-10-05
Avoid some pollution of CLI argument namespace across vtgate and vttablet

aquarapid:
This does not fix all instances of this problem, but it does improve the situation:

Before:

```
$ bin/vtgate -h 2>&1 | grep '  -' | wc -l
343
$ bin/vttablet -h 2>&1 | grep '  -' | wc -l
491
```

After:

```
$ bin/vtgate -h 2>&1 | grep '  -' | wc -l
196
$ bin/vttablet -h 2>&1 | grep '  -' | wc -l
484
```

The vttablet situation is still not great, but the remaining vttablet flags that are really vtgate-only are not easy to remove, e.g. the ones from `go/mysql`, which is not easy to break into sub-packages without import cycles;  and will need greater surgery.

### DIFF
--- a/go/vt/vttablet/tabletserver/tabletserver.go
+++ b/go/vt/vttablet/tabletserver/tabletserver.go
@@ -67,6 +67,7 @@ import (
 	"vitess.io/vitess/go/vt/vttablet/tabletserver/throttle"
 	"vitess.io/vitess/go/vt/vttablet/tabletserver/txserializer"
 	"vitess.io/vitess/go/vt/vttablet/tabletserver/txthrottler"
+	"vitess.io/vitess/go/vt/vttablet/tabletserver/utils"
 	"vitess.io/vitess/go/vt/vttablet/tabletserver/vstreamer"
 	"vitess.io/vitess/go/vt/vttablet/vexec"
 )
@@ -370,7 +371,7 @@ func (tsv *TabletServer) StopService() {
 // connect to the database and serving traffic), or an error explaining
 // the unhealthiness otherwise.
 func (tsv *TabletServer) IsHealthy() error {
-	if IsServingType(tsv.sm.Target().TabletType) {
+	if utils.IsServingType(tsv.sm.Target().TabletType) {
 		_, err := tsv.Execute(
 			tabletenv.LocalContext(),
 			nil,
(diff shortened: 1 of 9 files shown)


go/vt/vttablet/tabletserver/tabletserver.go
```diff
@@ -385,17 +386,6 @@ func (tsv *TabletServer) IsHealthy() error {
 	return nil
 }
 
-// IsServingType returns true if the tablet type is one that should be serving to be healthy, or false if the tablet type
-// should not be serving in it's healthy state.
-func IsServingType(tabletType topodatapb.TabletType) bool {
```
ajm188, line 390: This particular function feels like it belongs better in `topoproto` than in a subpackage of `vttablet/tabletserver`

ajm188 (commented):
I know this is still in draft, but in case it's helpful, I wonder if it would be better to move the flag definitions into subpackages rather than the other way around? For example, what I'm starting to do here: https://github.com/vitessio/vitess/pull/8915/commits/f556f7dbd320c553f03711f9a6cb5baedfc50499

go/vt/vttablet/tabletserver/tabletserver.go
aquarapid: Just moved this into topoproto itself.

aquarapid:
> I know this is still in draft, but in case it's helpful, I wonder if it would be better to move the flag definitions into subpackages rather than the other way around? For example, what I'm starting to do here: [f556f7d](https://github.com/vitessio/vitess/commit/f556f7dbd320c553f03711f9a6cb5baedfc50499)

Yeah, that probably fair enough.  But more surgery than I'm willing to commit to right now.

aquarapid:
For release notes:

```
It has been possible for several Vitess versions to pass certain vttablet flags to vtgate and certain vtgate flags to vttablet.  This did not have any effect on the operation of vtgate or vttablet, but did confuse users when looking at the commandline help.  This change fixes some instances of this problem.  As a result, users might have vtgate and vttablet startup scripts that use flags that will no longer be accepted, and vtgate or vttablet startup will fail.  Please test your vtgate and vttablet startup with the flags you use before rolling this version out in production."

PR #8940, opened 2021-10-06
Fix sqlparser.Walk compilation error when passing Select's From field

Serpentiel:
## Description
Fixes the following compilation error:
```
river/config.go:863:39: cannot use e.From (type []sqlparser.TableExpr) as type sqlparser.SQLNode in argument to sqlparser.Walk:
        []sqlparser.TableExpr does not implement sqlparser.SQLNode (missing Format method)
```

## Related Issue(s)
None

## Checklist

## Deployment Notes
None

### DIFF
--- a/go/vt/sqlparser/ast.go
+++ b/go/vt/sqlparser/ast.go
@@ -207,7 +207,7 @@ type (
 		StraightJoinHint bool
 		SQLCalcFoundRows bool
 		// The From field must be the first AST element of this struct so the rewriter sees it first
-		From        []TableExpr
+		From        TableExprs
 		Comments    Comments
 		SelectExprs SelectExprs
 		Where       *Where
--- a/go/vt/sqlparser/ast_clone.go
+++ b/go/vt/sqlparser/ast_clone.go
@@ -1364,7 +1364,7 @@ func CloneRefOfSelect(n *Select) *Select {
 	}
 	out := *n
 	out.Cache = CloneRefOfBool(n.Cache)
-	out.From = CloneSliceOfTableExpr(n.From)
+	out.From = CloneTableExprs(n.From)
 	out.Comments = CloneComments(n.Comments)
 	out.SelectExprs = CloneSelectExprs(n.SelectExprs)
 	out.Where = CloneRefOfWhere(n.Where)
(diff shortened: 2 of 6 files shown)


systay (changes requested):
The PR is changing the AST without changing most uses of the AST

Serpentiel:
Hey, @systay!

I use it the following way:
```go
func tableMapFromSelectQuery(e *sqlparser.Select) (map[string]string, error) {
	tableMap := map[string]string{}

	tableVisitor := func(n sqlparser.SQLNode) (bool, error) {
		switch t := n.(type) {
		case sqlparser.JoinCondition:
			return false, nil
		case *sqlparser.AliasedTableExpr:
			tName, ok := t.Expr.(sqlparser.TableName)
			if !ok {
				return false, errors.Errorf("expected TableName, but got %s", spew.Sdump(t.Expr))
			}
			alias := t.As.String()
			name := sqlExprToString(tName)
			if alias == "" {
				return false, errors.Errorf("no alias for table '%s'", name)
			}
			otherName, aliasExists := tableMap[alias]
			if aliasExists {
				return false, errors.Errorf("duplicate alias '%s' for tables '%s' and '%s'", alias, otherName, name)
			}
			tableMap[alias] = name
		}
		return true, nil
	}

	// TODO: https://github.com/vitessio/vitess/pull/8940
	err := sqlparser.Walk(tableVisitor, e.From)
	if err != nil {
		return nil, errors.Trace(err)
	}

	return tableMap, nil
}
```

The faulty line is `err := sqlparser.Walk(tableVisitor, e.From)` which is easily fixable by replacing it with `err := sqlparser.Walk(tableVisitor, sqlparser.TableExprs(e.From))`.

Currently `e.From` doesn't satisfies `SQLNode` interface, but the `TableExprs` alias of `[]TableExpr` does, hence the PR.

You can also [track down](https://github.com/vitessio/vitess/commit/65e6f14f35f1c71e0e081e05a1da9316634f3040#diff-700292c6158fc802e3ca4107b1560bf8b702f684a5ebc1d482154e917a5078d3) with `git blame` that it once was `TableExprs` in `ast.go` but was changed.

systay (approved):
Ah, I missed the *s* in `TableExprs`. Now it all makes more sense :)

systay:
Feel free to re-open if anything new is added

PR #8965, opened 2021-10-09
messaging: try to resolve deadlocks

### LINKED ISSUE #8909: vttablet: apparent deadlock when running PlannedReparentShard
#### Overview of the Issue

We have successfully executed thousands of PlannedReparentShard commands before upgrading to v11, and now they are very unreliable (25-100% failure rate with 0% recovery on successive attempts) on tablets with many concurrent vstreams. I can't prove that VStream is causing it, but my guess is that's where we're different from other Vitess clusters. We also have several Materialize workflows running cross-keyspaces.

I have linked to full log output (https://gist.github.com/derekperkins/299519b0bd9da3618645de908a37ca0c), but this seems to be where things are getting locked:

```
I0929 03:47:08.552644 1897126 rpc_replication.go:404] DemotePrimary disabling query service
I0929 03:47:08.552668 1897126 state_manager.go:214] Starting transition to MASTER Not Serving, timestamp: 2021-09-28 23:56:35.082672802 +0000 UTC
I0929 03:47:08.552753 1897126 tablegc.go:193] TableGC: closing
```

When I do a reparent on another keyspace that doesn't have so many streams, this is what I see immediately following `TableGC: closing`. It looks like there must be some race condition with how the Messager is closing, leading to a deadlock

```
I0929 23:50:08.869083   24777 engine.go:102] Messager: closed
I0929 23:50:08.869135   24777 tx_engine.go:333] No grace period specified: performing normal wait.
I0929 23:50:08.871506   24777 tx_engine.go:204] TxEngine: closed
I0929 23:50:08.871536   24777 state_manager.go:504] Killing all OLAP queries.
I0929 23:50:08.904262   24777 repltracker.go:86] Replication Tracker: going into master mode
I0929 23:50:08.904311   24777 state_manager.go:563] TabletServer transition: MASTER: Serving, Sep 29, 2021 at 04:20:33 (UTC) -> MASTER: Not Serving, Sep 29, 2021 at 04:20:33 (UTC) for tablet :companies/0
I0929 23:50:08.904330   24777 state_manager.go:663] State: exiting lameduck
I0929 23:50:08.926629   24777 query.go:81] exec SET GLOBAL read_only = ON
```

I turned off 3 Materialize workflows (16 shards for 48 vreplication rows) and things seem to have improved. I do need those running and have many others that I plan on starting, so I am really hoping to get this resolved in time for v12.

#### Operating system and Environment details

GKE 1.20.9
vttablet: 11.0.0
vtctld: 11.0.0

derekperkins:
## Description

I did an audit of the various lock acquisitions and wait groups in messaging to see if I could isolate reasons for deadlocking in #8909. I can't confirm that this solves that problem, and none of these seem likely to have caused significant problems, but should be a good cleanup and cause no harm. The commits are small and atomic, so reviewing each commit individually should be relatively simple. I've repeated the commits and messages here for convenience.

- **messager: call stopVStream with mu lock** - stopVStream is called from unsubscribe with the mm.mu lock, so this matches that behavior
- **messager: make wg.(Add|Done) pairs more obvious**
- ~**messager: defer Unlock in Engine.Open** - the comment about sharing the same lock appears to be out of date, and it seems more reliable to set isOpen and registering the notifier while holding the lock. UnregisterNotifier is called while holding the same lock during Close()~
- **messager: defer Broadcast when leaving empty state** - since this is currently called before actually adding the message to the cache, it introduces a race condition where the woken up goroutines might still see an empty cache before the add completes. It seems unlikely that this would happen in practice, since both Add and IsEmpty acquire the cache mutex, but shouldn't cause any problems either

## Related Issue(s)
May fix #8909 

## Checklist
- [ ] Backport to v12.0.0

### DIFF
--- a/go/vt/vttablet/tabletserver/messager/engine.go
+++ b/go/vt/vttablet/tabletserver/messager/engine.go
@@ -17,9 +17,8 @@ limitations under the License.
 package messager
 
 import (
-	"sync"
-
 	"context"
+	"sync"
 
 	"vitess.io/vitess/go/sqltypes"
 	"vitess.io/vitess/go/sync2"
@@ -78,12 +77,12 @@ func (me *Engine) Open() {
 		me.mu.Unlock()
 		return
 	}
+	me.isOpen = true
 	me.mu.Unlock()
 	log.Info("Messager: opening")
 	// Unlock before invoking RegisterNotifier because it
 	// obtains the same lock.
 	me.se.RegisterNotifier("messages", me.schemaChanged)
-	me.isOpen = true
 }
 
 // Close closes the Engine service.


derekperkins:
Apparently I was wrong about the comment being out of date for the shared lock in the `defer Unlock in Engine.Open` commit. Somewhere around 70% of the tests deadlocked with that there. I did move the `isOpen = true` inside the mutex lock, but now I'm not sure if that should be done at all.

PR #8978, opened 2021-10-11
Update CLI help for vtgate/vttablet for schema tracking to clarify that

aquarapid:
options need to be enabled on both sides for schema tracking to be
functional.


frouioui (approved):
Looks good to me!

PR #8981, opened 2021-10-12
Report Export BuildGitRev in statsd backend

sonne5:

## Description

BuildGitRev is often a combination of numbers and characters, therefore there is no good way report them in statsdbackend. However, it is helpful to expose this metrics for operation purposes.

This PR export BuildGitRev as a checksum value, so that we can monitoring the BuildGitRev

## Related Issue(s)

^

## Checklist


sonne5:
gentle nudge on this PR, thanks!

aquarapid (approved):

Issue #6754, opened 2020-09-21
SAVEPOINT does not exist: RELEASE SAVEPOINT sa_savepoint_1 (errno 1305)


Smita8081:
#### Overview of the Issue
Hi, I am working onto replace usage of MYSQL with vitess cluster for Openstack's core services such as Nova, Neutron, Glance etc.
On integrating Vitess cluster with Nova service of Openstack, I am facing below mentioned error

`Execute: SAVEPOINT does not exist: RELEASE SAVEPOINT sa_savepoint_1 (errno 1305) (sqlstate 42000), request: map[BindVariables:map[] Session:autocommit:true target_string:"nova" options:<included_fields:ALL client_found_rows:true > last_insert_id:1 found_rows:1 row_count:-1 Sql:RELEASE SAVEPOINT sa_savepoint_1]`

#### Reproduction Steps

Steps to reproduce this issue, example:

1. Deployed below mentioned Vitess cluster for Nova keyspace:

```
apiVersion: planetscale.com/v2
kind: VitessCluster
metadata:
  name: nova
spec:
  images:
    vtctld: docker.io/vitess/lite:v7.0.1
    vtgate: docker.io/vitess/lite:v7.0.1
    vttablet: docker.io/vitess/lite:v7.0.1
    vtbackup: docker.io/vitess/lite:v7.0.1
  cells:
  - name: novazone
    gateway:
      authentication:
        static:
          secret:
            name: nova-cluster-config
            key: users.json
      replicas: 1
      resources:
        requests:
          cpu: 100m
          memory: 256Mi
        limits:
          memory: 256Mi
  vitessDashboard:
    cells:
    - novazone
    extraFlags:
      security_policy: read-only
    replicas: 1
    resources:
      limits:
        memory: 128Mi
      requests:
        cpu: 100m
        memory: 128Mi

  keyspaces:
  - name: nova
    turndownPolicy: Immediate
    partitionings:
    - custom:
        shards:
        - databaseInitScriptSecret:
            name: nova-cluster-config
            key: init_db.sql
          keyRange: {}
          replication:
            enforceSemiSync: false
          tabletPools:
          - cell: novazone
            replicas: 1
            vttablet:
              extraFlags:
                db_charset: utf8
                queryserver-config-pool-size: "5"
                queryserver-config-stream-pool-size: "5"
                queryserver-config-transaction-cap: "5"
              resources:
                requests:
                  cpu: 100m
                  memory: 256Mi
            externalDatastore:
[...]

PR #8983, opened 2021-10-12
Add a testcase that shows savepoints not working correctly after a change to `sql_mode`.

arthurschreiber:
## Description

This adds a failing end2end test case that shows that save point support breaks when `-enable_system_settings` is enabled (which it is by default) and a change to `sql_mode` is detected (even if there is no "real" change to the mode).

This also causes the connection to be stuck and stick around until it gets killed by the transaction killer, causing MySQL to hit the connection limit if many of these savepoint errors are encountered in a short timeframe.

This is especially problematic e.g. in Rails application test suites, where transactions and savepoints are used to manage database state during test execution.

## Related Issue(s)

This might be the same issue as described in https://github.com/vitessio/vitess/issues/6754.

### DIFF
--- a/go/test/endtoend/vtgate/system_settings_and_savepoints/main_test.go
+++ b/go/test/endtoend/vtgate/system_settings_and_savepoints/main_test.go
@@ -1,0 +1,123 @@
+/*
+Copyright 2019 The Vitess Authors.
+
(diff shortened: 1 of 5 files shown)


harshit-gangal:
I have fixed the part where no actual change in sql_mode causes reserved connection. #9019

PR #8984, opened 2021-10-12
Add a failing test case showing that a unique key error can cause Vitess to lose transaction state.

arthurschreiber:
## Description

We've encountered an issue where writing to a sharded table with a lookup VIndex can fail and cause the surrounding transaction state to be lost.

In our case, writing to the original table failed due to a `unique key` constraint not being met. Because this failure happens after Vitess has written information to the Lookup Vindex's table, Vitess performs a rollback of the transaction.

When outside of a user-initiated transaction, this is the correct behaviour to revert all changes that were performed. But when inside of a user-initiated transaction, this will actually abort the user's transaction, and can leave the application that is performing the operation in an undefined state.

I think the correct behaviour here would be to check whether a transaction was already started. If no transaction was started yet, the current behaviour is correct and should be used. If a transaction was already started, Vitess should create a new savepoint that it can roll back to in case of a later failure, or release the savepoint if the insert operation on the table succeeds.

Is https://github.com/arthurschreiber/vitess/blob/f10fa1eb0060f58e1088726a7330a61340dfabe9/go/vt/vtgate/plan_execute.go#L118-L123 the location in the code where that logic would have to be implemented / modified?

### DIFF
--- a/go/test/endtoend/vtgate/unique_key_indexes/main_test.go
+++ b/go/test/endtoend/vtgate/unique_key_indexes/main_test.go
@@ -1,0 +1,145 @@
+/*
+Copyright 2019 The Vitess Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


harshit-gangal:
The approach looks good, savepoint seems reasonable, to achieve this.

harshit-gangal:
some write up on the issue https://github.com/vitessio/vitess/issues/9266

harshit-gangal:
closing this in favor of https://github.com/vitessio/vitess/pull/9269

I will copy over the discussion to RFC

Issue #8772, opened 2021-09-06
VStream does not work when keyspaces_to_watch is set on vtgate


sonne5:
#### Overview of the Issue
VStream relies on topo server to create a tablet picker: https://github.com/vitessio/vitess/blob/990d49ea576ed480a61f415be855fdebf25d99b5/go/vt/vtgate/vstream_manager.go#L431

After https://github.com/vitessio/vitess/pull/7873, when we pass `keyspaces_to_watch` flag, VTGate will initialize a `keyspaceFilteringServer` which will return error when we try to get topo server out of it: https://github.com/vitessio/vitess/blob/c1c9276bcfd6b5191fc493a11b6dc523d3179066/go/vt/srvtopo/keyspace_filtering_server.go#L67

This could be problematic if there is a lot of of keyspaces, and the vtgate has to keep connections to _all_ of them in order to support vstream() endpoint.

@setassociative suggested providing a readonly topo interface in keyspaceFilteringServer, instead of raising error directly - which I think makes sense, it can:
- avoid the online DDL issue that we were trying to address initially (i don't fully grok the problem yet, but a readonly topo interface that avoids writes seems to be good enough for the specific use case)
- allow vstream with `keyspaces_to_watch` flag

#### Reproduction Steps
Launch VTGate with `keyspaces_to_watch` flag and hit VStream endpoint on vtgate

#### Binary version
I'm on top of `4358c6`
```
commit 4358c6a85e607c2676135599f36ef3b53705f004 (upstream/main)
Merge: 6afc58d9df 92e07c2293
Author: Harshit Gangal <harshit@planetscale.com>
Date:   Wed Aug 25 00:25:56 2021 +0530
```

setassociative:
Minor nit -- #7873 didn't introduce this; errors have been thrown when trying to access a writable topo server since the introduction of `keyspaces_to_watch` in https://github.com/vitessio/vitess/pull/4420. If Vstream every worked on a vtgate with this flag it'll be because the method of selecting a tablet at some point didn't rely on a writable topo server ref.

PR #8988, opened 2021-10-13
Support VStream with keyspaces_to_watch

mattlord:
## Description
In order to safely support the [`VStream`](https://vitess.io/docs/design-docs/vreplication/vstream/stream-migration/) api against VTGates running with the [`-keyspaces_to_watch`](https://vitess.io/docs/reference/programs/vtgate/) flag we needed to add a read-only [`VTTopo`](https://vitess.io/docs/reference/features/topology-service/) server interface to ensure that no changes are made to the topology from the (potentially) explicitly limited view of the topology provided via this `vtgate`.

## Related Issue(s)
 - https://github.com/vitessio/vitess/issues/8772

## Checklist
- [x] Should this PR be backported? No. IMO this should bake in 13.0-snapshot
- [x] Tests were added
- [x] Documentation is not required (we never documented that these two features did NOT work together)

### DIFF
--- a/go/test/endtoend/vtgate/keyspace_watches/keyspace_watch_test.go
+++ b/go/test/endtoend/vtgate/keyspace_watches/keyspace_watch_test.go
@@ -48,7 +48,7 @@ var (
 		PRIMARY KEY (id)
 	) Engine=InnoDB;`
 	vschemaDDL      = "alter vschema create vindex test_vdx using hash"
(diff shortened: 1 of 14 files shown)


mattlord:
The new test is failing as expected: https://github.com/vitessio/vitess/pull/8988/checks?check_run_id=3879443572
```
...
I1013 07:14:19.558124   18763 vstream_test.go:125] 2021-10-13 07:14:19.558123471 +0000 UTC m=+39.257232583:: remote error: Code: UNKNOWN
173
rpc error: code = Unknown desc = cannot access underlying topology server when keyspace filtering is enabled
...
```

PR #9017, opened 2021-10-19
Gen4: Filter primitive and planning

frouioui:
## Description

This pull request adds the `Filter` engine primitive along with the `filter` logical plan. A basic version of the `ComparisonExpr` (implementation of EvalEngine's `Expr`) is added through this pull request too. `ComparisonExpr` supports basic operations (`=`, `!=`, `<`, `<=`, `>`, `>=`) on values that can be transformed to numerical values, i.e support for type coercion will come in the following pull request.

## Related Issue(s)

- #7280 

## Checklist


go/vt/vtgate/planbuilder/rewrite_test.go
```diff
@@ -120,7 +120,10 @@ func TestHavingRewrite(t *testing.T) {
 		output: "select 1 from t1 where (x = 1 or y = 2) and a = 1 having count(*) = 1",
 	}, {
 		input:  "select count(*) k from t1 where x = 1 or y = 2 having a = 1 and k = 1",
-		output: "select count(*) as k from t1 where (x = 1 or y = 2) and a = 1 having count(*) = 1",
+		output: "select count(*) as k from t1 where (x = 1 or y = 2) and a = 1 having k = 1",
+	}, {
+		input:  "select count(*) k from t1 having k = 10",
+		output: "select count(*) as k from t1 having k = 10",
```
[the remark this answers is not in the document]
frouioui: we will add a test for this when we support this kind of rewriting

go/vt/vtgate/engine/route_test.go
```diff
@@ -82,7 +82,7 @@ func TestSelectInformationSchemaWithTableAndSchemaWithRoutedTables(t *testing.T)
 	stringListToExprList := func(in []string) []evalengine.Expr {
 		var schema []evalengine.Expr
 		for _, s := range in {
-			schema = append(schema, evalengine.NewLiteralString([]byte(s)))
+			schema = append(schema, evalengine.NewLiteralString([]byte(s), 0))
```
systay, line 85: all the magic 0 in this files should be `collations.Unknown`, right?

go/vt/vtgate/planbuilder/filter.go
```diff
@@ -0,0 +1,83 @@
...
+	simpleConverterLookup struct {
+		semTable *semantics.SemTable
+		plan     logicalPlan
+	}
+)
+
+var _ logicalPlan = (*filter)(nil)
+var _ evalengine.ConverterLookup = (*simpleConverterLookup)(nil)
+
+func (s *simpleConverterLookup) ColumnLookup(col *sqlparser.ColName) (int, error) {
```
systay, line 45: should this be public? if so, it needs a comment

systay (approved):
Looks great. Just two minor comments that we could address in a future PR as well

go/vt/vtgate/engine/route_test.go
frouioui: yes it should

go/vt/vtgate/planbuilder/filter.go
frouioui: The interface has this method in public, I will add descriptive comments to the interface and its implementations

frouioui:
> Looks great. Just two minor comments that we could address in a future PR as well

I will address the two comments via https://github.com/vitessio/vitess/pull/9097

go/vt/vtgate/evalengine/evalengine.go
```diff
@@ -312,3 +338,113 @@ func compareNumeric(v1, v2 EvalResult) (int, error) {
...
+		t, err = time.Parse("2006-01-02 15:04:05", string(expr.bytes))
+	case sqltypes.Time:
+		t, err = time.Parse("15:04:05", string(expr.bytes))
+		if err == nil {
+			now := time.Now()
+			// setting the date to today's date, because we use AddDate on t
+			// which is "0000-01-01 xx:xx:xx", we do minus one on the month
+			// and day to take into account the 01 in both month and day of t
+			t = t.AddDate(now.Year(), int(now.Month()-1), now.Day()-1)
+		}
```
harshit-gangal, line 356: do you have a test case for this? looks like there can be corner cases with timezone (set on MySQL vs VTGate instance)

go/vt/vtgate/engine/ordered_aggregate_test.go
```diff
@@ -496,14 +497,16 @@ func TestOrderedAggregateSumDistinctGood(t *testing.T) {
 		"a|1|3",
 		"b|1|1",
 		"c|7|2",
-		"d|null|1",
+		"d|1|2",
```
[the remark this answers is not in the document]
systay: we wanted to test more cases

go/vt/vtgate/evalengine/evalengine.go
systay: what type of tests do you feel would be helpful here?

frouioui: this is just an internal behavior of Golang where `time.Time`s with no date have their dates set by default to `0000-01-01`

go/vt/vtgate/engine/ordered_aggregate_test.go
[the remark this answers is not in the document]
frouioui: I agree with @harshit-gangal, naming is now confusing, in the test input we are saying `// Single null` even though we have two rows with only one of them being `null`. I'll fix the naming and add another row to test `single null` behavior.

go/vt/vtgate/evalengine/evalengine.go
harshit-gangal: filtering on column with time column on left and date column on right.
not sure if evalengine supports that, that would lead the test to come to this code path.

frouioui: the EvalEngine supports it thanks to this code snippet you highlighted. we also have tests for it:
https://github.com/vitessio/vitess/blob/1152eccae4919bacd3436cec151f4b8b3afa0ff8/go/vt/vtgate/evalengine/comparisons_test.go#L804-L815

Issue #8606, opened 2021-08-09
Collation and Character set support


GuptaManan100:
## Feature Description

Vitess does not yet have support for collations and character-sets. So, to compare varchar strings Vitess needs to rely on [WEIGHT_STRING](https://dev.mysql.com/doc/refman/5.7/en/string-functions.html#function_weight-string) function for now. As per MySQL documentation, WEIGHT_STRING is a debugging function, meant only for internal use. We can move away from this implementation after we have collation and character-set support in Vitess

## Use Case(s)
- Having the ability to compare strings using collation and character set support we will be able to better implement ORDER BY, GROUP BY, JOIN.
- It will also allow us to leverage more advanced join techniques than what we currently implement.

king-11:
Hey, @GuptaManan100 I would like to apply for this project under LFX. I am not much familiar with Vitess as of now can you point me somewhere I can look at current implementation so that I can design the changes that need to be made. Thanks 😄

GuptaManan100:
Hello, @king-11 https://github.com/vitessio/vitess/tree/main/go/vt/vtgate/planbuilder is the package which has the code for building the plans of how queries are to be executed in Vitess. You can look at this https://github.com/vitessio/vitess/blob/main/go/vt/vtgate/planbuilder/testdata/aggr_cases.txt file which contains the test data for aggregation queries. You will find weight_string function added to a lot of those. You can run the tests locally from https://github.com/vitessio/vitess/blob/main/go/vt/vtgate/planbuilder/plan_test.go file.

rgrupesh:
Hello @GuptaManan100 

I went through the issue and I feel like I can accomplish this task. I am working on my cover letter, will submit that along with my resume through LFX portal.

Is there else that I should know of?

rgrupesh:
> Hello, @king-11 https://github.com/vitessio/vitess/tree/main/go/vt/vtgate/planbuilder is the package which has the code for building the plans of how queries are to be executed in Vitess. You can look at this https://github.com/vitessio/vitess/blob/main/go/vt/vtgate/planbuilder/testdata/aggr_cases.txt file which contains the test data for aggregation queries. You will find weight_string function added to a lot of those. You can run the tests locally from https://github.com/vitessio/vitess/blob/main/go/vt/vtgate/planbuilder/plan_test.go file.

Thank you for the links. This was helpful.

GuptaManan100:
[...]

PR #9018, opened 2021-10-19
Collations Module Integration

king-11:
### Integration of Collation Module into Vitess Engine

- #8991

Adds in the collation module into the Vitess Code which can then be used to replace instances of the `weight_string` function used in the vitess codebase. The major change that this PR makes is with the `NullSafeComparator`

https://github.com/vitessio/vitess/blob/315d90751886b5901672a7278623de7d48aa4ec2/go/vt/vtgate/evalengine/arithmetic.go#L187

which now takes in additional collation info parameter which is used by the `collations.LookupByID` to find out the collation and then use it to collate the `varchar` strings based on their collation set.

### Update Function Calls

So given the `NullSafeComparator` needs access to collation info we intend to provide the information available to it by updating the function calls and trying to fetch charset info from the `result` or `Fields` available at the least distant parent function calls and trickle it down the succeeding calls.

## Related Issue(s)

- #8606 

## Checklist

### DIFF
--- a/go/vt/vtgate/engine/ordered_aggregate.go
+++ b/go/vt/vtgate/engine/ordered_aggregate.go
@@ -186,6 +186,15 @@ func (oa *OrderedAggregate) GetTableName() string {
 	return oa.Input.GetTableName()
 }
 
+// getCollations specifies the collation ID value for columns.
+func (oa *OrderedAggregate) getCollations() map[int]collations.ID {
+	colls := make(map[int]collations.ID)
+	for _, key := range oa.GroupByKeys {
+		colls[key.KeyCol] = key.CollationID
+	}
+	return colls
+}
+
 // SetTruncateColumnCount sets the truncate column count.
 func (oa *OrderedAggregate) SetTruncateColumnCount(count int) {
 	oa.TruncateColumnCount = count
@@ -209,6 +218,7 @@ func (oa *OrderedAggregate) execute(vcursor VCursor, bindVars map[string]*queryp
 		Fields: oa.convertFields(result.Fields),
 		Rows:   make([][]sqltypes.Value, 0, len(result.Rows)),
 	}
+	colls := oa.getCollations()
 	// This code is similar to the one in StreamExecute.
 	var current []sqltypes.Value
 	var curDistincts []sqltypes.Value
@@ -217,14 +227,13 @@ func (oa *OrderedAggregate) execute(vcursor VCursor, bindVars map[string]*queryp
 			current, curDistincts = oa.convertRow(row)
 			continue
 		}
-
-		equal, err := oa.keysEqual(current, row)
+		equal, err := oa.keysEqual(current, row, colls)
 		if err != nil {
 			return nil, err
 		}
 
 		if equal {
-			current, curDistincts, err = oa.merge(result.Fields, current, row, curDistincts)
+			current, curDistincts, err = oa.merge(result.Fields, current, row, curDistincts, colls)
 			if err != nil {
 				return nil, err
 			}
@@ -271,6 +280,7 @@ func (oa *OrderedAggregate) TryStreamExecute(vcursor VCursor, bindVars map[strin
 				return err
 			}
 		}
+		colls := oa.getCollations()
 		// This code is similar to the one in Execute.
 		for _, row := range qr.Rows {
 			if current == nil {
@@ -278,13 +288,13 @@ func (oa *OrderedAggregate) TryStreamExecute(vcursor VCursor, bindVars map[strin
 				continue
 			}
 
-			equal, err := oa.keysEqual(current, row)
+			equal, err := oa.keysEqual(current, row, colls)
 			if err != nil {
 				return err
 			}
 
 			if equal {
-				current, curDistincts, err = oa.merge(fields, current, row, curDistincts)
+				current, curDistincts, err = oa.merge(fields, current, row, curDistincts, colls)
 				if err != nil {
 					return err
 				}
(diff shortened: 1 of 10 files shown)


systay:
Another general comment - since we need to do planner work to make sure that the engine primitives and evalengine get the right info, what about focusing this PR to make it work well with grouping (`orderedAggregate`)? We can work on the rest of the primitives once we are happy with grouping.

vmg:
@systay: Yes, but do note that the field in `Request.Fields` that contains the numeric identifier for the _collation_ is actually called `charsetnr`, even though it's a _collation identifier_. MySQL is silly like that.

vmg:
> what about focusing this PR to make it work well with grouping (orderedAggregate)? We can work on the rest of the primitives once we are happy with grouping.

This sounds like a great idea. Let's make grouping work perfectly and leave the other callsites for another time. This means that everywhere else we don't pass a collation ID to the comparison function. That should be an easy way to unblock @king-11.

king-11:
@systay @frouioui awaiting your review made the changes and added in tests

go/vt/vtgate/engine/ordered_aggregate.go
```diff
@@ -186,6 +186,17 @@ func (oa *OrderedAggregate) GetTableName() string {
 	return oa.Input.GetTableName()
 }
 
+// GetCollations specifies the collation ID value for columns.
+func (oa *OrderedAggregate) GetCollations() map[int]collations.ID {
```
systay, line 190: this method is only used from this file and it's tests, so I don't think we need to make it public

go/vt/vtgate/engine/ordered_aggregate_test.go
```diff
@@ -1050,3 +1053,137 @@ func TestMultiDistinct(t *testing.T) {
...
+		fields,
+		"a|1",
+		"A|1",
+		"Ǎ|1",
+		"b|2",
+		"c|7",
+		"\xE3\x83\x8F\xE3\x81\xAF|2",
+		"\xE3\x83\x8F\xE3\x83\x8F|1",
+	)
+	fmt.Println(oa.GetCollations())
```
systay, line 1187: forgot debug code?

go/vt/vtgate/engine/ordered_aggregate_test.go
```diff
@@ -652,13 +653,15 @@ func TestMerge(t *testing.T) {
 		"1|3|2.8|2|bc",
 	)
 
-	merged, _, err := oa.merge(fields, r.Rows[0], r.Rows[1], nil)
+	// TODO(king-11) make collation aware
```
systay, line 656: shouldn't this work if we are adding collations to ordered_aggregate?

go/vt/vtgate/evalengine/arithmetic.go
```diff
@@ -212,6 +223,22 @@ func NullsafeCompare(v1, v2 sqltypes.Value) (int, error) {
...
 		return bytes.Compare(v1.ToBytes(), v2.ToBytes()), nil
 	}
+	if v1.IsText() && v2.IsText() && collationID != collations.Unknown {
+		collation := collations.FromID(collationID)
+		if collation == nil {
+			return 0, UnsupportedCollationError{
+				ID: collationID,
+			}
+		}
+		switch result := collation.Collate(v1.ToBytes(), v2.ToBytes(), false); {
```
systay, line 233: we don't really need to switch on this, right? we should be able to just return the value from `collation.Collate` here, I think

go/vt/vtgate/vindexes/consistent_lookup.go
```diff
@@ -314,7 +315,7 @@ func (lu *clCommon) Delete(vcursor VCursor, rowsColValues [][]sqltypes.Value, ks
 func (lu *clCommon) Update(vcursor VCursor, oldValues []sqltypes.Value, ksid []byte, newValues []sqltypes.Value) error {
 	equal := true
 	for i := range oldValues {
-		result, err := evalengine.NullsafeCompare(oldValues[i], newValues[i])
+		result, err := evalengine.NullsafeCompare(oldValues[i], newValues[i], collations.Unknown)
```
systay, line 318: `TODO`?

frouioui (approved):
Nothing more than what @systay wrote to add. Otherwise lgtm

go/vt/vtgate/evalengine/arithmetic.go
king-11: `Collate` returns a positive or negative or zero value and `NullSafeCompare` needs to return -1,0 or 1.

go/vt/vtgate/engine/ordered_aggregate.go
king-11: made the change

go/vt/vtgate/vindexes/consistent_lookup.go
king-11: added `TODO`

go/vt/vtgate/engine/ordered_aggregate_test.go
king-11: for this test we don't need to pass in collation I added `TODO` by mistake this works as is supposed to.

king-11: made the change removed `TODO`

king-11:
@systay made the changes as per your review.

shlomi-noach:
I'm not sure _how_, but this seems to have broken `main`. The following line: 

https://github.com/vitessio/vitess/blob/ab2f0b442e81cc6b5700ffaedc9d8b755898b6b5/go/vt/vtgate/evalengine/arithmetic.go#L227

Does not compile:
```
# vitess.io/vitess/go/vt/vtgate/evalengine
go/vt/vtgate/evalengine/arithmetic.go:227:16: undefined: collations.FromID
```

right now unit tests and some vreplication tests are failing.

systay:
#9131 contains the fix for main :point_up:

Issue #8959, opened 2021-10-08
Update the region_sharding examples for v2 commands


hallaroo:
#### Feature Description

Currently the examples in https://github.com/vitessio/vitess/tree/main/examples/region_sharding still reference SwitchReads and SwitchWrites, rather than SwitchTraffic.

We should update the following examples to correctly reflect the v2 command:

- https://github.com/vitessio/vitess/blob/main/examples/region_sharding/204_switch_reads.sh 
- https://github.com/vitessio/vitess/blob/main/examples/region_sharding/205_switch_writes.sh

#### Docs updates

Once the updates are made the following docs section will also need to be updated:

https://vitess.io/docs/user-guides/configuration-advanced/region-sharding/#cutover

PR #9031, opened 2021-10-19
switchtraffic updates for region sharding examples

hallaroo:
## Description
This PR adds adjusts the region sharding examples to reference `SwitchTraffic' rather than the -v1 vreplication commands.

## Related Issue(s)
This PR addresses: https://github.com/vitessio/vitess/issues/8959

## Checklist
- [X] Documentation was added or is not required : https://github.com/vitessio/website/pull/847

### DIFF
--- a/examples/region_sharding/205_switch_writes.sh
+++ b/examples/region_sharding/205_switch_writes.sh
@@ -18,7 +18,7 @@
 
 source ./env.sh
 
-vtctlclient Reshard -tablet_types=PRIMARY SwitchTraffic main.main2regions
+vtctlclient Reshard -tablet_types=primary SwitchTraffic main.main2regions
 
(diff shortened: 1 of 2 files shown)


deepthi:
test is failing, probably need to fix `203_reshard.sh` as well:
```
+ ./203_reshard.sh
*** The Reshard v1 flow is deprecated, consider using v2 commands instead, see https://vitess.io/docs/reference/vreplication/v2/ ***
```

deepthi:
Turns out that `SwitchTraffic` expects tablet_types to be lowercase, though `Reshard ... Create` accepts uppercase. They seem to be going through different code paths.
For now, we should document this, but we also need to fix it, otherwise it will be super confusing to users.

PR #9049, opened 2021-10-21
[RFC] git: deprecate golint checker

vmg:
## Description

As per https://github.com/golang/go/issues/38968, `golint` has been **frozen and deprecated** for over a year. The project hasn't been updated since 2018, and since it cannot be configured, it leads to a lot of noise and false positives. Right now running `golint` through the codebase provides thousands of warnings, with 90%+ of them being not relevant at all. It is also extremely slow, greatly increasing the time it takes us to `git commit`.

I propose deprecating the use of `golint` and instead enabling https://github.com/mgechev/revive, a replacement for `golint` with more sensible rules which already ships as part of `golangci-lint`. It's more customizable and should let us start working through the thousands of old warnings from `golint` to cleanup the codebase.

## Checklist


shlomi-noach:
I'm in favor of removing `golint` given that it's abandoned. However, I'm also wary of a random linter. Way back, before `golint` existed, there were a bunch of competing linters in the ecosystem, and every single one of them gave me _incorrect_ error messages on legit code (tested in `openark/orchestrator`).

That being said, it's a nice list of software that uses that kinter.

systay:
I like it. Linters are not always easy to work with, but I prefer living with them than without them.

harshit-gangal (approved):
There is merge conflict

shlomi-noach:
I took the liberty of resolving the conflict seeing that it was just a spacing issue.

deepthi:
> I took the liberty of resolving the conflict seeing that it was just a spacing issue.

The actual lint errors have to be resolved before we can merge this.

vmg:
OK, so in order to get this merged we need the `golangci-lint` build to be green -- which is unfeasible because we've never ran `golint`/`revive` in CI before, so there's like 500 different linting errors reported by the tool.

What I've done for this PR is disabling `revive` in the CI run and enabling it for the `git` post-commit hook -- this is the exact same behavior we have right now, so CI is green and after we merge this PR **it will not be possible to commit code that doesn't pass the `revive` checks**. I think this is a good thing which will help us grow the codebase in a cleaner way until eventually we can enable `revive` in CI too.

Also, while I was working on this tooling, I noticed we had a pretty complex post-commit hook script to run `golangci-lint` while only checking the files that have changed. I removed all that code and we're now using the `--new` flag in `golangci-lint`, which does the exact same thing but more efficiently (this flag was not available when we first introduced `golangci-lint`).

ajm188:
👋 @vmg one of the sideeffects of this PR is that locally, `golangci-lint` now runs on every package on every commit, rather than just packages changed by the commit.

Was this known/intended? It now takes nearly a minute and hundreds of % cpu (and fans) to commit changes on my laptop.

ajm188:
`--new` doesn't really work well for local setups where we might (and I do, which I think is making this worse for me) have untracked changes from time to time; we probably want `new-from-rev`:

```
  -n, --new                            Show only new issues: if there are unstaged changes or untracked files, only those changes are analyzed, else only changes in HEAD~ are analyzed.
                                       It's a super-useful option for integration of golangci-lint into existing large codebase.
                                       It's not practical to fix all existing issues at the moment of integration: much better to not allow issues in new code.
                                       For CI setups, prefer --new-from-rev=HEAD~, as --new can skip linting the current patch if any scripts generate unstaged files before golangci-lint runs.
      --new-from-rev REV               Show only new issues created after git revision REV
```

PR #9054, opened 2021-10-21
Backoff on inconsistent member state

sonne5:

## Description

We had an AZ gameday between 19:11 to 19:21 UTC, around the same time there seems to be a network blip that leads to a lost of quorum in the group:

2021-10-14T19:12:19.552500Z 0 [ERROR] [MY-011496] [Repl] Plugin group_replication reported: 'This server is not able to reach a majority of members in the group. This server will now block all updates. The server will remain blocked for the next 5 seconds. Unless contact with the majority is restored, after this time the member will error out and leave the group. It is possible to use group_replication_force_members to force a new group membership.'
There is something fishy happened inside mysql that makes different node reported different group view:

on primary:
```
mysql> select * from performance_schema.replication_group_members;
+---------------------------+--------------------------------------+---------------------------------------------------------+-------------+--------------+-------------+----------------+
| CHANNEL_NAME              | MEMBER_ID                             | MEMBER_ROLE | MEMBER_VERSION |
+---------------------------+--------------------------------------+---------------------------------------------------------+-------------+--------------+-------------+----------------+
| group_replication_applier | 0d2f5008-2c71-11ec-9588-0208f3fcc021  | ONLINE       | PRIMARY     | 8.0.22         |
| group_replication_applier | 4c13646c-2c70-11ec-991a-06731dd33837  | ONLINE       | SECONDARY   | 8.0.22         |
| group_replication_applier | dcee6acb-2c70-11ec-ae59-0a8fcea3b2db  | ONLINE       | SECONDARY   | 8.0.22         |
+---------------------------+--------------------------------------+---------------------------------------------------------+-------------+--------------+-------------+----------------+
```
on secondary that appears to be online from above:
```
mysql> select * from performance_schema.replication_group_members;
+---------------------------+--------------------------------------+---------------------------------------------------------+-------------+--------------+-------------+----------------+
| CHANNEL_NAME              | MEMBER_ID                             | MEMBER_STATE | MEMBER_ROLE | MEMBER_VERSION |
+---------------------------+--------------------------------------+---------------------------------------------------------+-------------+--------------+-------------+----------------+
| group_replication_applier | dcee6acb-2c70-11ec-ae59-0a8fcea3b2db | OFFLINE      |             |                |
+---------------------------+--------------------------------------+---------------------------------------------------------+-------------+--------------+-------------+----------------+
1 row in set (0.00 sec)
```
As a result, the group is stuck. The only way to get away this situation is to stop the group and rebootstrap.

This PR handles this situation by transit into back off mode when there is a discrepancy in member_state reported by different nodes, if it happens for longer than a threshold VTGR will reboot the group.

And to make it more conservative, it also adds an option to integrate with Vitess's heartbeat table: only execute backoff if there is no heartbeat for a certain period of time.

## Related Issue(s)

^

## Checklist

### DIFF
--- a/go/vt/vtgr/controller/diagnose.go
+++ b/go/vt/vtgr/controller/diagnose.go
@@ -121,7 +121,7 @@ func (shard *GRShard) Diagnose(ctx context.Context) (DiagnoseType, error) {
 	shard.shardStatusCollector.recordDiagnoseResult(diagnoseResult)
 	shard.populateVTGRStatusLocked()
 	if diagnoseResult != DiagnoseTypeHealthy {
-		shard.logger.Warningf(`VTGR diagnose shard as unhealthy for %s/%s: result=%v | last_result=%v | instances=%v | primary=%v | primary_tablet=%v | problematics=%v | unreachables=%v | SQL group=%v`,
+		shard.logger.Warningf(`VTGR diagnose shard as unhealthy for %s/%s: result=%v, last_result=%v, instances=%v, primary=%v, primary_tablet=%v, problematics=%v, unreachables=%v,\n%v`,
 			shard.KeyspaceShard.Keyspace, shard.KeyspaceShard.Shard,
 			shard.shardStatusCollector.status.DiagnoseResult,
 			shard.lastDiagnoseResult,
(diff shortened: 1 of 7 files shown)


sonne5:
Close this in favor of https://github.com/vitessio/vitess/pull/9292

PR #9075, opened 2021-10-25
Changed Collations and Expression grammar

### LINKED ISSUE #8490: Incorrect Handling of COLLATE
#### Overview of the Issue

Currently, Vitess does not parse `COLLATE` correctly. 

The following statement gets parsed correctly:-
```sql
create table x (e enum('red','yellow') collate 'utf8_bin' null default null)
```

but this one does not :-
```sql
create table x (e enum('red','yellow') null collate 'utf8_bin'
```

ritwizsinha:
## Description

This PR fixed collation and expression parsing and matched them with what MySQL was doing
## Related Issue(s)

- Fixes #8490 
- Fixes parsing for exists expressions for #9241

## Checklist


go/vt/sqlparser/parse_test.go
```diff
@@ -2757,15 +2757,13 @@ func TestSubStr(t *testing.T) {
...
 
 func TestLoadData(t *testing.T) {
-	//validSQL := []string{
-	//	"load data from s3 'x.txt'",
-	//	"load data from s3 manifest 'x.txt'",
-	//	"load data from s3 file 'x.txt'",
-	//	"load data infile 'x.txt' into table 'c'",
-	//	"load data from s3 'x.txt' into table x"}
 	validSQL := []string{
-		"select 5 between 0 and 10 between 0 and 1,(5 between 0 and 10) between 0 and 1",
```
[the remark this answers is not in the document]
ritwizsinha: The same test is already present in select_cases.txt, it was added here just to debug

go/vt/sqlparser/parse_test.go
```diff
@@ -121,48 +127,24 @@ var (
...
-	}, {
-		input:  "create table x(location MULTIPOLYGON DEFAULT POINT(7.0, 3.0))",
-		output: "create table x (\n\tlocation MULTIPOLYGON default (POINT(7.0, 3.0))\n)",
 	}, {
 		input:  "create table x(location GEOMETRYCOLLECTION DEFAULT (POINT(7.0, 3.0)))",
 		output: "create table x (\n\tlocation GEOMETRYCOLLECTION default (POINT(7.0, 3.0))\n)",
-	}, {
-		input:  "create table x(location GEOMETRYCOLLECTION DEFAULT POINT(7.0, 3.0))",
-		output: "create table x (\n\tlocation GEOMETRYCOLLECTION default (POINT(7.0, 3.0))\n)",
 	}, {
```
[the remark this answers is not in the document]
ritwizsinha: Currently the parser still parses this but not correctly, the output is 
```
create table x
```
and not syntax error

MySQL has some code to check the ast if these symbols are present and then throw the error, we would have to do the same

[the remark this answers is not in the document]
GuptaManan100: No that isn't needed, MySQL will throw the error for us. Then keep the test and set partialDDL to true

go/vt/sqlparser/ast.go
```diff
@@ -1802,9 +1802,9 @@ type (
 
 	// BetweenExpr represents a BETWEEN or a NOT BETWEEN expression.
 	BetweenExpr struct {
-		Operator RangeCondOperator
-		Left     Expr
-		From, To Expr
+		isBetween bool
```
systay, line 1805: this should be a public field, so we can inspect it from outside the parser package

go/vt/sqlparser/ast.go
```diff
@@ -1935,7 +1935,7 @@ type (
 	// SubstrExpr('static string value', value_expression, value_expression)
 	// In this case StrVal will be set instead of Name.
 	SubstrExpr struct {
```
systay, line 1937: I think SubstrExpr is wrong
my understanding is that you can feed it arbitrary expression
```sql
mysql> select substring(1002010,3);
+----------------------+
| substring(1002010,3) |
+----------------------+
| 02010                |
+----------------------+
1 row in set (0.00 sec)
```

here I tried a literal int, that gets coerced into a string

```
mysql> select substring(lower("ALL CAPS"),3);
+--------------------------------+
| substring(lower("ALL CAPS"),3) |
+--------------------------------+
| l caps                         |
+--------------------------------+
1 row in set (0.01 sec)
```

We should remove `StrVal *Literal` and rename `Name` to `Expr`

go/vt/sqlparser/ast.go
```diff
@@ -1986,7 +1986,7 @@ type (
 	// supported functions are documented in the grammar
 	CurTimeFuncExpr struct {
 		Name ColIdent
-		Fsp  Expr // fractional seconds precision, integer from 0 to 6
+		Fsp  *Literal // fractional seconds precision, integer from 0 to 6
```
systay, line 1989: I think this should be an int instead. no need to wrap it in a Literal

ritwizsinha: Regarding having a pointer for fsp, the fsp also accepts nil and otherwise an IntLiteral thus we had a pointer Literal and not and int

ritwizsinha: The nil is required because, sometimes there may be no arguments to the function

go/vt/sqlparser/ast.go
```diff
@@ -1968,8 +1968,8 @@ type (
 	// SubstrExpr('static string value', value_expression, value_expression)
```
GuptaManan100, line 1968: This comment needs to change now

go/vt/sqlparser/ast.go
```diff
@@ -1968,8 +1968,8 @@ type (
...
 	// In this case StrVal will be set instead of Name.
 	SubstrExpr struct {
-		Name   *ColName
-		StrVal *Literal
```
GuptaManan100, line 1972: StrVal is unused and should be removed

GuptaManan100 (approved):

go/vt/sqlparser/ast.go
GuptaManan100: Fixed

go/vt/sqlparser/ast.go
GuptaManan100: Fixed

PR #9084, opened 2021-10-26
Add CIFuzz

AdamKorcz:
## Description
Adds CIFuzz to Vitess's OSS-fuzz integration. In short, CIFuzz is a service offered by OSS-fuzz to run Vitess's fuzzers during the CI to prevent bugs from being introduced.
In this PR the fuzz time is set to 600 seconds but can be changed.

More about CIFuzz can be found here: https://google.github.io/oss-fuzz/getting-started/continuous-integration/

## Checklist

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,27 @@
+# CIFuzz runs the fuzzers
+name: CIFuzz
+on: [pull_request]
+jobs:
+  Fuzzing:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Build Fuzzers
+      id: build
+      uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'vitess'
+        dry-run: false
+        language: go
+    - name: Run Fuzzers
+      uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
+      with:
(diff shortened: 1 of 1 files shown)


systay:
Hi @AdamKorcz!

Seems like something is unhappy about CIFuzz. 
```
Running go-fuzz -tags gofuzz -func FuzzTestBuilder -o fuzz_test_builder.a vitess.io/vitess/go/vt/vtgate/planbuilder
# vitess.io/vitess/go/vt/vtgate/planbuilder/abstract
go/vt/vtgate/planbuilder/abstract/fuzz.go:31:5: cannot use (*fakeFuzzSI)(nil) (type *fakeFuzzSI) as type semantics.SchemaInformation in assignment:
	*fakeFuzzSI does not implement semantics.SchemaInformation (missing ConnCollation method)
go/vt/vtgate/planbuilder/abstract/fuzz.go:59:48: cannot use &fakeFuzzSI{} (type *fakeFuzzSI) as type semantics.SchemaInformation in argument to semantics.Analyze:
	*fakeFuzzSI does not implement semantics.SchemaInformation (missing ConnCollation method)
```

AdamKorcz:
> Seems like something is unhappy about CIFuzz.

This is because the fuzzing build is broken. https://github.com/vitessio/vitess/pull/9289 should fix that.

systay:
Merged #9289. Should we merge master into this branch?

PR #9089, opened 2021-10-26
Online DDL: support --allow-zero-in-date in 'online' strategy

shlomi-noach:
`gh-ost` recently added support for `--allow-zero-in-date`, letting the user create columns such as `dt datetime default '0000-00-00 00:00:00'`. See https://github.com/openark/gh-ost/pull/31 and https://github.com/openark/gh-ost/pull/32.

Vitess recently also included `gh-ost` binary to support the above: #9052 and #9087. This is important for Vitess users because they sometimes work with frameworks that use such defaults.

This PR adds similar support for `online` (vreplication) migrations.

- `ddl_strategy` now "officially" supports `--allow-zero-in-date` flag
- for `gh-ost`, it just passes the same flag (conveniently under exact same name)
- for `online`, onlineddl/Executor notices the flag and sets `sql_mode` accordingly

Added `vrepl` suite tests to observe behavior:

- Fail operations when `--allow-zero-in-date` is not provided.
- Succeed same operations when `--allow-zero-in-date` is provided.

General tracking issue: #6926


go/vt/vttablet/onlineddl/executor.go
```diff
@@ -716,7 +716,44 @@ func (e *Executor) cutOverVReplMigration(ctx context.Context, s *VReplStream) er
...
+	// Grab current sql_mode value
+	rs, err := conn.ExecuteFetch(`select @@session.sql_mode as sql_mode`, 1, true)
+	if err != nil {
+		return deferFunc, vterrors.Errorf(vtrpcpb.Code_UNKNOWN, "could not read sql_mode: %v", err)
+	}
+	sqlMode, err := rs.Named().Row().ToString("sql_mode")
+	if err != nil {
+		return deferFunc, vterrors.Errorf(vtrpcpb.Code_UNKNOWN, "could not read sql_mode: %v", err)
+	}
+	// Pre-calculate estore function
```
dbussink, line 737: Typo, `restore` here I assume?

dbussink (approved):
All I found was a little typo in a comment, if you can fix that up that would be great.

go/vt/vttablet/onlineddl/executor.go
shlomi-noach: Fixed! Will merge once tests pass

PR #9097, opened 2021-10-27
Add default collation to VTGate and VTTablet

frouioui:
## Description

This pull request does two things:
- Changing the default charset from `utf8` to `utf8mb4`
- Adding a default collation to both VTGate and VTTablet

We are changing the default charset from `utf8` (`utf8mb3`) to `utf8mb4` because `utf8mb4` is way more optimized than its predecessor `utf8mb3`. Another reason, `utf8mb4` supports more characters. With `utf8mb4` being a superset of `utf8mb3`, queries that were previously relying on `utf8mb3` should not have a problem with this new charset.

### Default Collation
We are adding a default collation to both VTGate and VTTablet, this is done to let us do string comparisons at the VTGate level. It also ensures that we are using the same `connection collation` as the one between VTTablet and MySQL. For both components, the collation can be set using the `-collation` (for VTGate) and `-db_collation` (for VTTablet). If the flag is not specified, it will default to the default collation of `utf8mb4` in regards to the underlying database version we are running (MySQL57, MySQL80, etc).

The default collation for certain charset varies depending on the MySQL version we are running, for this reason, the VTTablet is responsible for telling VTGate which MySQL version is being used.

#### VTGate
At start time, VTGate subscribes to a health check and waits until a VTTablet indicates which MySQL version is used. Once the MySQL version is known in VTGate, VTGate will create its own collation environment that contains the default collation to use and the `collations.Environment` that will allow us to lookup other collations when planning/executing queries.

Before each execution, the default collation is added to the `ExecuteOptions` so the VTTablet can make sure collations are all the same from the client to MySQL.

#### VTTablet
When opening a new connection with a MySQL server, VTTablet is responsible for setting the connection collation properly using the collation/charset the user has defined, and again, in regards to the underlying MySQL version we are connecting to.

## Documentation
- https://github.com/vitessio/website/pull/879

## Checklist


go/mysql/conn_params.go
```diff
@@ -16,19 +16,24 @@ limitations under the License.
...
+	Port                 int                     `json:"port"`
+	Uname                string                  `json:"uname"`
+	Pass                 string                  `json:"pass"`
+	DbName               string                  `json:"dbname"`
+	UnixSocket           string                  `json:"unix_socket"`
+	Charset              string                  `json:"charset"`
+	Collation            string                  `json:"collation"`
+	Flags                uint64                  `json:"flags"`
+	Flavor               string                  `json:"flavor,omitempty"`
+	CollationEnvironment *collations.Environment `json:"collation_environment,omitempty"`
```
vmg, line 36: I am not sure this is the right place to store a collation environment. Among other semantic questions, the environment cannot be serialized as JSON at all so...

go/mysql/fakesqldb/server.go
```diff
@@ -406,6 +406,13 @@ func (db *DB) comQueryOrdered(query string) (*sqltypes.Result, error) {
 	db.mu.Lock()
 	defer db.mu.Unlock()
 
+	// when creating a connection to the database, we send an initial query to set the connection's
+	// collation, we want to skip the query check if we get such initial query.
+	// this is done to ease the test readability.
+	if query == "SET collation_connection = utf8mb4_general_ci" {
```
vmg, line 412:
```suggestion
	if strings.HasPrefix(query, "SET collation_connection =") {
```

vmg (changes requested):
This is looking great Floflo. I don't like the collation environment in the Conn Params though. It doesn't feel right, again, because it doesn't serialize as JSON, which the conn params must do. Let's come up with a different place to store it. Hit me up in Slack if you need help.

frouioui:
> This is looking great Floflo. I don't like the collation environment in the Conn Params though. It doesn't feel right, again, because it doesn't serialize as JSON, which the conn params must do. Let's come up with a different place to store it. Hit me up in Slack if you need help.

Your two comments are addressed via a56fe1d

go/mysql/conn_params.go
frouioui: After pairing up, we decided to move `CollationEnvironment` to the `mysql.Conn` instead.

go/mysql/conn.go
```diff
@@ -1511,3 +1524,24 @@ func (c *Conn) IsUnixSocket() bool {
 func (c *Conn) GetRawConn() net.Conn {
 	return c.conn
 }
+
+// MatchCollation returns nil if the given collations.ID matches with the connection's
+// collation, otherwise it returns an error explaining why it does not match.
+// We do the comparison all the way down in the Connector to use mysql.ConnParams
+// collations environment to achieve the collation lookup using the same server version.
```
vmg, line 1531: Outdated comment

vmg (approved):
Paired on this with @frouioui this morning. Lookin' gucci. :fr::crown:

Issue #8465, opened 2021-07-13
vtgate HealthCheck Tablet Cache incorrect information


grsmith-projects:
Hello!

I've noticed that when there is a topology change, sometimes the `HealthCheck Tablet Cache`
data for vtgate becomes stale and it misreports TabletHealth.

I've been able to repro this a few times in a cluster that is:

1 x 2 shard keyspace, with 3 replicas each
1 x unsharded keyspace, with 3 replicas

There are 40 vtgate instances (2 cpu), we are pushing ~15k req/s with varying payload sizes and mix of different query plans.

to repro, I just kill any vttablet pod - there is a chance that some number of vtgate `HealthCheck Tablet Cache` will become stale.

I can determine this by vtgate having an entry for a tablet 1234, that is showing as a failed master -- but recently has been changed to a replica. The tablet ID will be the same for when it changes state between master/replica, but I think the pod IP changes.  The newly created tablet with a different IP will set itself healthy, but the old tablet will remain in the table as unhealthy for some reason (on a % of existing vtgate instances). The only way I've found to clear it is by restarting the vtgate pod. I've attached a screenshot.

![Screenshot from 2021-07-13 12-32-58](https://user-images.githubusercontent.com/1625735/125522557-ddaa7e94-c1c6-4cb9-b7e9-6fa50ebbf6b8.png)

The stale data shows the tablet alias as red, with an error message that says 

```(Code: CANCELED vttablet: rpc error: code = Canceled desc = context canceled )```

I found this in the logs which appears to correlate with the old IP address of the tablet showing a stale HealthCheck:

> [core] grpc: addrConn.createTransport failed to connect to {10.0.0.1:15999 10.0.0.1:15999 <nil> 0 <nil>}. Err: connection error: desc = "transport: Error while dialing dial tcp 10.0.0.1:15999: operation was canceled". Reconnecting...

```/vt/bin/vtgate --version
Version: 12.0.0-SNAPSHOT (Git revision 9fa6ff4400 branch 'main') built on Thu Jul  8 18:16:08 UTC 2021 by vitess@buildkitsandbox using go1.16.5 linux/amd64
```

This is using kubernetes.

Thanks!

mattlord:
Hi @grsmith-projects !

Thank you for reporting this.

I have not yet been able to repeat this when running loops like this which, AFAICT, are doing what you mention (using the [docker_local](https://vitess.io/docs/get-started/local-docker/) container):
```
mysql> show vitess_tablets;
[...]

PR #9106, opened 2021-10-28
Remove tablet healthcheck cache record on error

mattlord:
## Description
If a tablet's health check runs concurrently with its deletion then you can end up with a timing issue where the goroutine getting the status from `vtgate`->`vttablet` via an RPC call has a connection related error (e.g. context cancelled or EOF) but a copy of the tablet's healthcheck cache record is made and that copy's updated state is stored in the healthcheck cache *after* the tablet has gone away. Thus you can end up with a zombie tablet record in the healthcheck cache.

The shorter your [`-health_check_interval` tablet flag](https://vitess.io/docs/reference/programs/vttablet/) is (default is 30s), the more likely you will encounter this. I was able to [repeat it quite easily and regularly](https://github.com/vitessio/vitess/issues/8465#issuecomment-951433088) in the [`docker_local` container](https://vitess.io/docs/get-started/local-docker/) in part because [the interval is set to 5s for those tablets](https://github.com/vitessio/vitess/blob/c3c3c213acf15ebc0d93c223227420cee59ba051/examples/compose/vttablet-up.sh#L152) and the `DeleteTablet` calls made in the test case ([./307_delete_shard_0.sh](https://github.com/vitessio/vitess/blob/c3c3c213acf15ebc0d93c223227420cee59ba051/examples/local/307_delete_shard_0.sh#L21)) are somewhat slow given the execution context.

I was no longer able to repeat the bug using [my test case](https://github.com/vitessio/vitess/issues/8465#issuecomment-951433088) with this patch. 

> Note: w/o this patch, the only way to remove these zombie tablet healthcheck records is to restart the vtgate

## Related Issue(s)
https://github.com/vitessio/vitess/issues/8465

## Checklist
- [x] Should this PR be backported? Thoughts?
- [x] Tests were added
  - This demonstrates the test failing (as expected) on a branch that only has the test: https://github.com/vitessio/vitess/pull/9109

### DIFF
--- a/.github/workflows/cluster_endtoend_vtgate_tablet_healthcheck_cache.yml
+++ b/.github/workflows/cluster_endtoend_vtgate_tablet_healthcheck_cache.yml
@@ -1,0 +1,53 @@
+# DO NOT MODIFY: THIS FILE IS GENERATED USING "make generate_ci_workflows"
+
+ name: Cluster (vtgate_tablet_healthcheck_cache)
+ on: [push, pull_request]
+ concurrency:
+   group: format('{0}-{1}', ${{ github.ref }}, 'Cluster (vtgate_tablet_healthcheck_cache)')
+   cancel-in-progress: true
+
+ jobs:
+   build:
+     name: Run endtoend tests on Cluster (vtgate_tablet_healthcheck_cache)
+     runs-on: ubuntu-18.04
+
+     steps:
+     - name: Set up Go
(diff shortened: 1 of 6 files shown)


deepthi (approved):
:100: on the tests. They give us confidence in this change :smiley:

sougou (approved):

deepthi:
The new changes look good. Thank you for fixing WaitForStatus.

PR #9108, opened 2021-10-28
Collations 1.3

vmg:
## Description

More collations work! I'll add a few more patches to this between today and tomorrow.

- @frouioui requested this functionality: https://github.com/vitessio/vitess/commit/1c665d8896f40e981791af9b1ed837755abb3ca9

## Related Issue(s)
- https://github.com/vitessio/vitess/issues/8606

## Checklist


go/mysql/collations/env.go
```diff
@@ -0,0 +1,166 @@
...
+	all = make([]Collation, 0, len(env.byID))
+	for _, col := range env.byID {
+		col.Init()
+		all = append(all, col)
+	}
+	return
+}
+
+// NewEnvironment creates a collation Environment for the given MySQL Version
+func NewEnvironment(version Version) *Environment {
```
frouioui, line 103: Can we have a global `*Environment` here that gets created when calling this method or another one, so we don't have to keep a `*Environment` in the caller, we can just use `collations.GlobalEnvironment`, i.e:

```go
// instantiate the global environment
collations.NewEnvironment(MySQL5.7)

// use the global environment without having to store it
collations.GlobalEnvironment.DefaultCollationForCharset("utf8")
```

go/mysql/collations/internal/codegen/codegen.go
```diff
@@ -0,0 +1,35 @@
+package codegen
+
```
frouioui, line 2: header is missing

go/mysql/collations/tools/makemysqlversions/makemysqlversions.go
```diff
@@ -0,0 +1,115 @@
+package main
+
+import (
```
frouioui, line 3: missing header too

frouioui (approved):

go/mysql/collations/env.go
vmg: Yikes, that's extremely nasty API design. That's not happening, sorry. If you need a global collation environment, you can keep it on the `gen4` planner package... But I suspect @systay will also scream a you a little bit. :)

Spend some time thinking more carefully where the environment needs to live. I'm confident "globally for the whole process" is not the right answer.

PR #9110, opened 2021-10-29
Update to the latest Tengo version

dbussink:
This updates to Tengo which is based on latest upstream which includes a fix for MySQL8 & handling of utf8 (mb3) collations.

The large go.mod change happens because it updated also minimum dependency versions.

## Checklist


shlomi-noach (approved):
:shipit:

PR #9112, opened 2021-10-29
Fixes buffering while reparenting

### LINKED ISSUE #9071: Buffering while reparenting doesn't work if there is no serving primary when the first query arrives
#### Overview of the Issue

This issue was found while investigating the flakiness in `TestBufferReparenting` test. Specifically the sub tests dealing with PRS.

Consider the following sequence of operations - 
1. Let the initial primary be tablet 1. Initially, everything is working fine and write queries are being served by this tablet.
2. PRS is issued to promote tablet 2 over tablet 1.
3. DemotePrimary is called on tablet 1 as part of PRS. This sets the query serving status on the tablet to false.
```
I1022 19:45:58.599250   89927 rpc_replication.go:364] DemotePrimary
I1022 19:45:58.608283   89927 rpc_replication.go:414] DemotePrimary disabling query service
I1022 19:45:58.608304   89927 state_manager.go:214] Starting transition to PRIMARY Not Serving, timestamp: 2021-10-22 14:15:53.742713 +0000 UTC
```
4. The next health check which is sent out from this tablet has serving status set to false.
5. On vtgate this health check is received and the map of healthy tablets is updated
```
I1022 19:45:58.808826   89966 tablet_health_check.go:111] HealthCheckUpdate(Serving State): tablet: zone1-2478 (localhost) serving true => false for ks1/0 (PRIMARY) reason: healthCheck update
```
6. At this point there are no serving primary tablets as seen by the vtgate. A write query arrives on the vtgate which results in this error - 
```
I1022 19:45:59.466439   89966 vtgate.go:548] Execute: Code: UNAVAILABLE
no healthy tablet available for 'keyspace:"ks1" shard:"0" tablet_type:PRIMARY'

target: ks1.0.primary, request: map[BindVariables:map[vtg1:type:VARBINARY value:"update 4" vtg2:type:INT64 value:"2"] Session:in_transaction:true autocommit:true options:{included_fields:ALL workload:OLTP} system_variables:{key:"default_week_format" value:"1"} row_count:-1 in_reserved_conn:true DDLStrategy:"direct" SessionUUID:"93353972-3342-11ec-a141-acde48001122" enable_system_settings:true Sql:UPDATE buffer SET msg='update 4' WHERE id = 2]
```
7. The new primary tablet 2 gets promoted and sends its own health stream - 
```
### NOTE - health stream updates are not logged, on the vttablet in general. This log was added specifically for gathering intel.
E1022 19:46:20.860753   89953 health_streamer.go:245] heath stream sent - target:{keyspace:"ks1" shard:"0" tablet_type:PRIMARY} serving:true tablet_externally_reparented_timestamp:1634912179 realtime_stats:{} tablet_alias:{cell:"zone1" uid:2479}
```
8. Any query arriving from this point forward are served correctly.

The problem here is that the queries returned errors and were not buffered while the PRS was in progress.

This situation however arises, less frequently. In most cases, before the health check update on vtgate is processed, a write query gets already sent to the previous primary, which fails with a cluster event error triggering the start of buffering of requests. Any query which arrives after it also gets buffered, before the check for the number of serving tablets.
```
I1022 19:42:25.987773   83638 shard_buffer.go:280] Starting buffering for shard: ks1/0 (window: 10m0s, size: 10, max failover duration: 10m0s) (A failover was detected by this seen error: Code: CLUSTER_EVENT
vttablet: rpc error: code = Code(17) desc = operation not allowed in state SHUTTING_DOWN
.)
I1022 19:42:26.253427   83638 tablet_health_check.go:111] HealthCheckUpdate(Serving State): tablet: zone1-1382 (localhost) serving true => false for ks1/0 (PRIMARY) reason: healthCheck update
E1022
```

Therefore, the buffering during PRS will fail if the first query which needs to be bufferred, arrives between the short duration when the initial primary stops serving and the new primary has not started serving. If a query arrives after PRS starts but before demotePrimary is called, then all requests get buffered correctly.

#### Reproduction Steps

1. Run the test `TestBufferReparenting` multiple times and the test will fail eventually.

#### Binary version
main

GuptaManan100:
## Description

This PR adds an additional case to start buffering requests on the vtgate when a primary was available but went non-serving. Please look at the linked issue as to how we can arrive at this situation.

Serving state is only set using the type information of the tablet and during failover operations, Planned Reparent Shard and Emergency Reparent Shard.

So, if a primary tablet gets to a non-serving state, we can conclude that one of the two operations is in progress and can start buffering requests. When either of the two succeed, then the new primary sends its health check which stops the buffering.

The code is still agnostic to the usage of serving in the vttablet in the sense that if a primary goes to a non-serving state for whatever reason and becomes serving again, the buffering logic will work correctly and stop buffering requests as soon as the next healthcheck arrives. A unit test for that has also been added to gain confidence in this decoupled behaviour.
This is possible because of the amazing implementation of the new keyspace events watcher by @vmg which recognises the change in shard serving state and sends a keyspace event when everything is back to normal. 

## Related Issue(s)

- Fixes #9071 

## Checklist


go/vt/vtgate/tabletgateway_flaky_test.go
```diff
@@ -0,0 +1,219 @@
+/*
```
deepthi, line 1: Is this test still flaky? Or should these tests be merged into `tabletgateway_test.go`?

go/vt/vtgate/buffer/flags.go
```diff
@@ -226,3 +226,8 @@ func (cfg *Config) bufferingMode(keyspace, shard string) bufferMode {
 
 	return bufferModeDisabled
 }
+
+// SetBufferingMode should only be used from testing code to change the flag default value
+func SetBufferingMode(enabled bool) {
+	*bufferEnabled = enabled
```
deepthi, line 232: I don't like this being an exported func in a common place like this. Maybe move it to a test_util.go and name it explicitly so that no one is tempted to use it outside of tests?

deepthi (commented):
Lovely piece of work! Just a couple of comments/suggestions.

go/vt/vtgate/tabletgateway_flaky_test.go
harshit-gangal: this looks new tests to me, why will this be flaky?

harshit-gangal (approved):

go/vt/vtgate/tabletgateway_flaky_test.go
GuptaManan100: There is a fifteen second timeout in the end, it hasn't failed on my machine until now, but I am not certain whether it will work perfectly on the CI

go/vt/vtgate/buffer/flags.go
GuptaManan100: Okay, done!

harshit-gangal (approved):

PR #9116, opened 2021-10-29
[grpcshim] Fix select race and move to internal package

ajm188:
## Description

Fix race condition in bidiStream

In the original implementation, bidiStream used one channel, `errch`, to
receive both mid-stream errors and "the stream is closed now" errors.
This meant that, due to go's pseudo-random channel selection, if the
stream was closed and there were still messages in the stream (so, both
`ch` and `errch` were select-able), then there was a chance that an
adapter's `Recv()` would return `io.EOF` even though there were still
responses to pull off the stream.

This version moves to a more context-cancel like paradigm, using a
separate context to check if the stream is closed. Then, we adjust the
template for `Recv` to select:

1. Is stream context done
2. Is streamclosed context done
4. Does mid-stream error channel have an error
3. Does message channel have a message

And the template for `Send` to:

1. If streamclosed context done, return `errStreamClosed`; otherwise
2. Send to the message channel and return nil

<hr/>

Note that this commit also makes a more public/exported API for
bidiStream, to set up a simpler (diff-wise) filemove in a later change.

## Related Issue(s)

#9062 : you can see the codegen [here](https://github.com/vitessio/vitess/pull/9062/files#diff-9621a0f2f20e59135318750c24b496b6c22980027073095fd457df6e449f223a) and the reuse in the testing [here](https://github.com/vitessio/vitess/pull/9062/files#diff-2c51195b080b991d1ddaba9a30afbb1aef36225981435991acc77439e49ea1b0)

## Checklist
- [x] Should this PR be backported? no
- [x] Tests were added or are not required n/a
- [x] Documentation was added or is not required ohhhh yeah


go/vt/vtctl/internal/grpcshim/bidi_stream.go
```diff
@@ -0,0 +1,186 @@
...
+	// ErrCh receives errors mid-stream, and should be selected on with the
+	// same priority as stream.ch and stream/send contexts' cancellations in
+	// Recv methods.
+	ErrCh chan error
+
+	ctx              context.Context
+	sendClosedCtx    context.Context
+	sendClosedCancel context.CancelFunc
+
+	m        sync.RWMutex
```
deepthi, line 99: Quoting @sougou
" Programmatic RWLocks are different from the equivalent db functionality. They're actually more expensive than a normal mutex. They're only useful if you want multiple readers to hold them concurrently and for long periods, and then occasionally want to use a write lock to change the values. Also, mixing RW locking with write locks can easily lead to deadlocks if one doesn't think through the scenarios carefully.
So, in most cases, it's simpler and more efficient to use shortly held write locks even if RW lock seems functionally more appropriate,"

ajm188: Following up from our other conversation:

> looking again at the code, we don't really have many concurrent readers. Only if you passed the stream shim to multiple goroutines, which we currently don't do and probably never would do in practice. I'll switch to a plain-old Mutex

go/vt/vtctl/localvtctldclient/client.go
```diff
@@ -26,9 +26,8 @@ import (
 )
 
 var (
-	m               sync.RWMutex
-	server          vtctlservicepb.VtctldServer
-	errStreamClosed = errors.New("stream is closed for sending") // nolint (TODO:@ajm188) this will be used in a future PR, and the codegen will produce invalid code for streaming rpcs without this
+	m      sync.RWMutex
```
deepthi, line 29: Sorry I didn't notice this other `RWMutex` in the first review.
Should this also change to a plain Mutex?

ajm188: yeah, we can change that as well

deepthi (commented):
Code changes look fine. Is it possible to unit test this at all?

deepthi:
> > Code changes look fine. Is it possible to unit test this at all?
> 
> So because this is necessarily racy, I think the effort might be pretty high to get a test that's not flaky. _however!_, the whole reason I became aware of this was this bug was preventing me from writing working tests for a grpcvtctldserver migration ([branch](https://github.com/ajm188/vitess/tree/vtctldserver-backup)), so I'd prefer to let those tests indirectly exercise this code path.
> 
> What do you think?

That is acceptable.

deepthi (approved):

Issue #9101, opened 2021-10-27
Hex VARBINARY values not getting normalized by query planner


Anders-PlanetScale:
#### Overview of the Issue

An INSERT of a hex `0x` or `X'` notated value into a VARBINARY column  is not normalized by the vtgate query planner.

This means cached query plans for these types of INSERT queries are unique and won't be usable by similar subsequent INSERT statements.

#### Reproduction Steps

Steps to reproduce this issue, example:

Create a Vitess cluster with a 2 or more shard keyspace in it. For ease of repro I used the vitess/examples/local schema/vschema files. I created a keyspace1, which is unsharded and used for sequences. And a keyspace2, which has 2 shards and contains the schema for the table we will be running inserts on.

Apply schemas and vschemas to the cluster.

`vtctlclient -server=localhost:15999 ApplySchema -sql-file ~/vitess/examples/local/create_commerce_seq.sql keyspace1; vtctlclient -server=localhost:15999 ApplySchema -sql-file ~/vitess/examples/local/create_commerce_schema.sql keyspace2; vtctlclient -server=localhost:15999 ApplyvSchema -vschema_file ~/vitess/examples/local/vschema_commerce_seq.json keyspace1; vtctlclient -server=localhost:15999 ApplyvSchema -vschema_file ~/vitess/examples/local/vschema_customer_sharded.json keyspace2`

Once the schemas and vschemas are applied run the following test insert.
```

cat repro.sql
INSERT INTO corder(order_id, customer_id, sku, price) VALUES (100, 290, 0x676F, 1000) ,
(101, 400, 0x7265, 2000) ,
(102, 598, 0x6D65, 500);

```

`mysql -h 127.0.0.1 -P 15306 -u vitess -p < repro.sql`

browse to the vtgate web UI and look at the query_plans. On my setup this is located at: http://localhost:15001/debug/query_plans
Notice the query plan for our repro.sql INSERT shows normalized, bind var,  values for all but the hex values;

```
  "Key": "insert into corder(order_id, customer_id, sku, price) values (:vtg1, :vtg2, 0x676F, :vtg3), (:vtg4, :vtg5, 0x7265, :vtg6), (:vtg7, :vtg8, 0x6D65, :vtg9)",
  "Value": {
   "QueryType": "INSERT",
[...]

PR #9118, opened 2021-10-29
Ensure that hex query predicates are normalized for planner cache

mattlord:
## Description
We were not normalizing hex values of either format:
- `0x01`
- `x'73757265'`

This had a major performance impact in workloads where hex query predicates are used in common queries as we could not leverage the query planner cache. This work normalizes them.

> Note: the only explicitly non-normalized values are now bit types of the format `b'1'`:
```
mysql> select b'011101110111010101110100' /* wut */;
+----------------------------------------------------------+
| b'011101110111010101110100'                              |
+----------------------------------------------------------+
| 0x777574                                                 |
+----------------------------------------------------------+
1 row in set (0.00 sec)
```

Should we try and handle those here as well???

## Related Issue(s)
https://github.com/vitessio/vitess/issues/9101

## Checklist
- [x] Should this PR be backported? Yes, at least to 12.0 IMO
- [x] Tests were added


mattlord:
Making some progress: https://gist.github.com/mattlord/41575a480b667438e5e55ea3e740917b

go/test/endtoend/vtgate/queries/normalize/normalize_test.go
```diff
@@ -0,0 +1,82 @@
...
+	require.Nil(t, err)
+	found := false
+	for _, record := range results {
+		key := record["Key"].(string)
+		if key == normalizedInsertQuery {
+			found = true
+			break
+		}
+	}
+	assert.Equal(t, found, true, "correctly normalized record not found in planner cache")
```
systay, line 58: nit: `assert.True` is nice

mattlord: That is nicer, will switch to that. Thanks!

go/vt/sqlparser/ast_funcs.go
```diff
@@ -480,6 +481,26 @@ func (node *Literal) HexDecode() ([]byte, error) {
 	return hex.DecodeString(node.Val)
 }
 
+// EncodeHexValToMySQLQueryFormat encodes the hexval back into the query format
+// for passing on to MySQL as a bind var
+func (node *Literal) EncodeHexValToMySQLQueryFormat() ([]byte, error) {
```
systay, line 486: does this really need to be a public method?

systay (approved):
Nice stuff

go/vt/sqlparser/ast_funcs.go
mattlord: Good point. And, no, it doesn't. :) Done. Thanks!

java/client/pom.xml
```diff
@@ -5,7 +5,7 @@
   <parent>
     <groupId>io.vitess</groupId>
     <artifactId>vitess-parent</artifactId>
-    <version>12.0.0-SNAPSHOT</version>
+    <version>13.0.0-SNAPSHOT</version>
```
frouioui, line 8: aren't these changes already applied by https://github.com/vitessio/vitess/pull/9066?

frouioui (approved):
this is very nice!

java/client/pom.xml
mattlord: Yeah, not sure why it still shows them in the diff. I can rebase on main, but in the end this should not change either way when merged.

mattlord: rebased on main and they're gone from the diff.

PR #8843, opened 2021-09-17
[vtctld] migrate more util rpcs

ajm188:
## Description

This PR migrates the following RPCs to the new VtctldServer:

- `SetShardIsPrimaryServing`
- `SetShardTabletControl`
- `RebuildKeyspaceGraph`

### examples

```
❯ vtctldclient --server ":15999" SetShardIsPrimaryServing -h 2>&1 | less
❯ vtctldclient --server ":15999" GetTablets
zone1-0000000100 commerce 0 primary localhost:15100 localhost:17100 [] 2021-09-19T19:57:38Z
zone1-0000000101 commerce 0 replica localhost:15101 localhost:17101 [] <null>
zone1-0000000102 commerce 0 rdonly localhost:15102 localhost:17102 [] <null>
❯ vtctldclient --server ":15999" SetShardIsPrimaryServing commerce/0 false
{
  "primary_alias": {
    "cell": "zone1",
    "uid": 100
  },
  "primary_term_start_time": {
    "seconds": "1632081458",
    "nanoseconds": 536266000
  },
  "key_range": null,
  "source_shards": [],
  "tablet_controls": [],
  "is_primary_serving": false
}
❯ vtctldclient --server ":15999" SetShardIsPrimaryServing commerce/0 true
{
  "primary_alias": {
    "cell": "zone1",
    "uid": 100
  },
  "primary_term_start_time": {
    "seconds": "1632081458",
    "nanoseconds": 536266000
  },
  "key_range": null,
  "source_shards": [],
  "tablet_controls": [],
  "is_primary_serving": true
}
❯ vtctldclient --server ":15999" SetShardIsPrimaryServing junkdata 42
Usage:
   SetShardIsPrimaryServing <keyspace/shard> <true/false>
[more help text elided]

E0919 16:01:03.800499   81103 main.go:42] cannot parse keyspace/shard: invalid shard path: junkdata
❯ vtctldclient --server ":15999" SetShardIsPrimaryServing commerce/0 42
Usage:
   SetShardIsPrimaryServing <keyspace/shard> <true/false>

Flags:
  -h, --help   help for SetShardIsPrimaryServing

[more help text elided]
E0919 16:01:14.238557   81115 main.go:42] cannot parse is_serving as bool: strconv.ParseBool: parsing "42": invalid syntax
```

## Related Issue(s)

Closes https://github.com/vitessio/vitess/issues/8850
Closes https://github.com/vitessio/vitess/issues/8851

## Checklist
[...]

PR #9120, opened 2021-10-29
Vertical and horizontal split usage updates

hallaroo:
## Description

Did an audit on the usage of 'horizontal split' and 'vertical split' to see if there were further references after the following commit:

https://github.com/vitessio/vitess/commit/685c221f91142097831bd821993581f1c98c6968

Hopefully this should remove the rest of the references that aren't in split_clone go files.

## Related PR(s)

https://github.com/vitessio/vitess/pull/8843

## Checklist
- [x ] Documentation was added or is not required


hallaroo:
Definitely did something wrong here - closing this out.

PR #9130, opened 2021-11-03
vtcombo: `CREATE DATABASE` creates shards as specified by the topology

arthurschreiber:
## Description

The development workflow in many frameworks (e.g. Rails) relies heavily on the ability to drop and (re)create databases. The `vttest` dbddl plugin for `vtcombo` supports these statements, but dropping and recreating a sharded keyspace via these statements results in an unsharded keyspace.

This changes the code in `vtcombo` to check the original topology configuration to determine the number of shards to use when recreating a keyspace.

## Related Issue(s)

N/A

## Checklist
- [x] Tests were added

### DIFF
--- a/go/cmd/vtcombo/main.go
+++ b/go/cmd/vtcombo/main.go
@@ -29,6 +29,7 @@ import (
 	"time"
 
 	"google.golang.org/protobuf/encoding/prototext"
+	"google.golang.org/protobuf/proto"
 
 	"vitess.io/vitess/go/exit"
 	"vitess.io/vitess/go/vt/dbconfigs"
@@ -120,6 +121,12 @@ func main() {
 		exit.Return(1)
 	}
 
+	// Stash away a copy of the topology that vtcombo was started with.
+	//
+	// We will use this to determine the shard structure when keyspaces
+	// get recreated.
+	originalTopology := proto.Clone(tpb).(*vttestpb.VTTestTopology)
+
 	// default cell to "test" if unspecified
 	if len(tpb.Cells) == 0 {
 		tpb.Cells = append(tpb.Cells, "test")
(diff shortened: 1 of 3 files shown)


harshit-gangal:
have you tried this locally?

My understanding was dropping the database will delete records from the topo server, so next time creating the same database will not get any information from the topo server.

arthurschreiber:
> have you tried this locally?

Yes, of course. 😅 

> My understanding was dropping the database will delete records from the topo server, so next time creating the same database will not get any information from the topo server.

Yes, this is how VTCombo behaves _without_ the changes in this pull request. When starting up, the keyspaces are created as specified in the topology. When `DROP DATABASE` is called, the keyspace is removed and the in-memory topology is updated to reflect this. When `CREATE DATABASE` is called, the keyspaces is created, but as a completely new keyspaces without any knowledge of whatever sharding options were set when VTCombo was started, which means that it will always create an unsharded keyspace.

With this pull request, we store a copy of the original topology information, and consult that copy when `CREATE DATABASE` is executed to determine how many shards need to be started up.

---

As mentioned in the original body, a very common workflow when doing Rails development is to drop and recreate development/test databases to ensure they're in a fresh state. When doing so through VTCombo, the keyspace gets recreated but with an unsharded keyspace configuration, which then requires a restart of VTCombo to pick up the sharded configuration again. With this patch, things "just work" correctly, and only when the topology is changed a restart is required.

go/test/endtoend/vtcombo/recreate/recreate_test.go
```diff
@@ -0,0 +1,158 @@
+/*
+Copyright 2019 The Vitess Authors.
```
mattlord, line 2: For new files we should use the current year.

mattlord (approved):
Had two minor requests, otherwise looks good. Thank you!

harshit-gangal (commented):
LGTM once you address Matt's comments.

go/cmd/vtcombo/main.go
```diff
@@ -169,6 +172,12 @@ func main() {
 	}
 
 	globalCreateDb = func(ctx context.Context, ks *vttestpb.Keyspace) error {
+		for _, originalKs := range originalTopology.Keyspaces {
```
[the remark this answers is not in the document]
arthurschreiber: Good idea! I updated the changes accordingly.

go/test/endtoend/vtcombo/recreate/recreate_test.go
arthurschreiber: Done!

harshit-gangal (approved):

deepthi:
Looks like all feedback has been addressed, hence merged.
Thank you @arthurschreiber, nice work!

Issue #9148, opened 2021-11-04
Add make target and ci hook for running `mockgen` generators


ajm188:
#### Feature Description

tl;dr have CI catch us not running this when we should.

https://vitess.slack.com/archives/CMDJ2KFEZ/p1636055907040300

PR #9149, opened 2021-11-04
[mockgen] update generator directives and regenerate

ajm188:
## Description

This updates our `go:generate` directives to work with the version of mockgen we upgraded to in May, and then runs the generators, which produced some mostly comment-only diffs.

## Related Issue(s)

Follow-up tasks: https://github.com/vitessio/vitess/issues/9148

## Checklist
- [x] Tests were added or are not required **no**
- [x] Documentation was added or is not required **no**


deepthi (approved):
I had no idea these existed!
LGTM

PR #9152, opened 2021-11-05
Fix race condition when rapidly starting/stopping health check

choo-stripe:
## Description
There's an interesting race condition when starting/stopping a healthcheck.
1. Create `NewHealthCheck`
2. `NewHealthCheck` spawns topo-watchers in goroutines
3. `hc.Close()` is called
4. `Stop()` is called on the topo-watchers
5. The topo-watcher cancels the ctx and calls `Wait()` on the wait-group. The `Wait()` returns immediately, because the `Start()` goroutine hasn't even started yet. So nothing was ever added to the wait-group.
6. You close the topo-srv, which should be safe, because the health-check has already been closed
7. The `Start()` goroutine now fires, which calls `tw.loadTablets()` (before even checking ctx!). This will NPE because the topo-srv has already been closed.

I actually managed to reproduce this with a test.

There's another potential fix here:
- Move `loadTablets` under `case <-ticker.C`. Note that [`cancelFunc` is called before `Wait()`](https://github.com/vitessio/vitess/blob/f085f12b1cd38a9bcf907604a747af308e93c4fd/go/vt/discovery/topology_watcher.go#L142-L146). When `Stop()` is called, if we've already added to the wait-group, than `wg.Wait()` will make sure the goroutine actually exits. If we _haven't_ yet added to the wait-group, `tw.ctx.Done()` will fire before `ticker.C`, and again, we won't load tablets.

The last, unrelated fix I made is to check if the cell to watch is `""`. Splitting empty string cannot yield length 0: https://play.golang.org/p/5mKKN9TkMh5, so that check was not doing anything.

Lastly, I validated there were no other call-sites of `Start()` that we'd have to get rid of `go` in front:
```
rg 'go .*\.Start()'
```

## Related Issue(s)
- N/A

## Checklist

## Deployment Notes
N/A

### DIFF
--- a/go/vt/discovery/topology_watcher_test.go
+++ b/go/vt/discovery/topology_watcher_test.go
@@ -59,6 +59,52 @@ func checkChecksum(t *testing.T, tw *TopologyWatcher, want uint32) {
 	}
 }
 
+func TestStartAndCloseTopoWatcher(t *testing.T) {
+	ts := memorytopo.NewServer("aa")
+	fhc := NewFakeHealthCheck(nil)
+	topologyWatcherOperations.ZeroAll()
+	tw := NewCellTabletsWatcher(context.Background(), ts, fhc, nil, "aa", 100*time.Microsecond, true, 5)
+
+	done := make(chan bool, 3)
+	result := make(chan bool, 1)
+	go func() {
+		// We wait for the done channel three times since we execute three
+		// topo-watcher actions (Start, Stop and Wait), once we have read
+		// from the done channel three times we know we have completed all
+		// the actions, the test is then successful.
+		// Each action has a one-second timeout after which the test will be
+		// marked as failed.
+		for i := 0; i < 3; i++ {
+			select {
+			case <-time.After(1 * time.Second):
+				close(result)
+				return
+			case <-done:
+				break
+			}
+		}
+		result <- true
+	}()
+
+	tw.Start()
+	done <- true
+
+	// This sleep gives enough time to the topo-watcher to do 10 iterations
+	// The topo-watcher's refresh interval is set to 100 microseconds.
+	time.Sleep(1 * time.Millisecond)
+
+	tw.Stop()
+	done <- true
+
+	tw.wg.Wait()
+	done <- true
+
+	_, ok := <-result
+	if !ok {
+		t.Fatal("timed out")
+	}
+}
+
 func TestCellTabletsWatcher(t *testing.T) {
 	checkWatcher(t, true)
 }
--- a/go/vt/vtgate/tabletgateway_test.go
+++ b/go/vt/vtgate/tabletgateway_test.go
@@ -104,7 +104,8 @@ func TestTabletGatewayBeginExecuteBatch(t *testing.T) {
 }
 
 func TestTabletGatewayShuffleTablets(t *testing.T) {
-	tg := NewTabletGateway(context.Background(), nil, nil, "local")
+	hc := discovery.NewFakeHealthCheck(nil)
+	tg := NewTabletGateway(context.Background(), hc, nil, "local")
 
 	ts1 := &discovery.TabletHealth{
 		Tablet:  topo.NewTablet(1, "cell1", "host1"),
--- a/go/vt/discovery/healthcheck.go
+++ b/go/vt/discovery/healthcheck.go
@@ -279,7 +279,7 @@ func NewHealthCheck(ctx context.Context, retryDelay, healthCheckTimeout time.Dur
 	var topoWatchers []*TopologyWatcher
 	var filter TabletFilter
 	cells := strings.Split(cellsToWatch, ",")
-	if len(cells) == 0 {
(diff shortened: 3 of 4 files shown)


go/vt/discovery/healthcheck_test.go
```diff
@@ -505,6 +505,16 @@ func TestHealthCheckCloseWaitsForGoRoutines(t *testing.T) {
...
 }
 
+// TestHealthCheckStartCloseRace tests that rapidly calling Start()/Close()
+// still cleanly shuts down.
+func TestHealthCheckStartCloseRace(t *testing.T) {
+	ts := memorytopo.NewServer("cell")
+	hc := createTestHc(ts)
+	assert.Nil(t, hc.Close(), "Close returned error")
+	ts.Close()
+	time.Sleep(50 * time.Millisecond)
```
systay, line 515: I don't get it. why sleep at the end of the method like this?

choo-stripe: Without it, the race/NPE was not triggered. Basically, we need to give the health-check time to actually run the goroutine on a closed topo-server in order to get an NPE. This test-case is for a super-niche bug, so I'd be fine removing if you want.

frouioui: Is there a way we can add an assertion to assert the health check is correctly shut down on line 515?

frouioui: note that the `time.Sleep` might be a bit sensitive as it can introduce flakiness in the CI builds

frouioui (approved):
I left a comment to add, if it is possible, an assertion in the test you've added. Otherwise, it looks good to me.

deepthi (commented):
Nice fix! The code change itself looks good to me. Can you do the following?
- fix DCO, we cannot merge until that is fixed
- address the other feedback
- squash the commits when you fix the DCO (that will anyway require a force-push)

choo-stripe:
@frouioui I actually couldn't arrive at a good strategy to test this. Thoughts:
- It doesn't make sense to test this from the healthchecker, because it actually did nothing wrong. It was correctly calling `Stop` on the topo-watcher
- So I tried to create a test for the topo-watcher. But! There's a potential race in tests. If we:
    - create topo-watcher
    - `tw.Start()`
    - `tw.Stop()`
    The topo-watcher may or may not have already refreshed tablets. So we can't e.g. check for tablet refresh, because both a refresh and no refresh are acceptable. Also! Before the fix, we called `go tw.Start()`. We have no way to know if the topo-watcher would have added to the wait-group or not by the time we `Stop()`. So we can't check that. And we don't get signaled when the topo-watcher goroutines exit (that was the bug, it didn't exit). So without mucking around with the topo-watcher struct and adding test-only status flags, I don't have good ideas to validate that `Stop()` actually closes the goroutines.
- With the changes in this pr, we could create a simple test:
    ```
    // create topo-watcher
    tw.Start()
    tw.Stop()
    tw.wg.Wait()
    ```
    This would pass, and verify that all the `wg` goroutines actually exit. But it wouldn't fail with the old implementation, it would hang because `tw.Start()` is blocking. I'm having trouble coming up with a test that fails in the old implementation, and now passes.

Any ideas?

I did sign off on the commit and rebase. I also deleted the test.

go/vt/vtgate/tabletgateway_test.go
```diff
@@ -104,7 +104,8 @@ func TestTabletGatewayBeginExecuteBatch(t *testing.T) {
 }
 
 func TestTabletGatewayShuffleTablets(t *testing.T) {
-	tg := NewTabletGateway(context.Background(), nil, nil, "local")
+	hc := discovery.NewFakeHealthCheck(nil)
+	tg := NewTabletGateway(context.Background(), hc, nil, "local")
```
deepthi, line 108: Is this change really required?

deepthi:
@frouioui what are your thoughts on testing?

frouioui:
With this new implementation, where the main thread is blocked until all the topo watchers are added to the WaitGroup, I find it hard to create a test to assert a possible race condition. I think it is okay to leave it as it is now, aka with no additional tests. The change looks safe enough to me.

frouioui:
Hello @choo-stripe and @deepthi, I took the liberty to add a test case. The test ensures that rapidly starting and stopping the topo watcher works as intended.

go/vt/discovery/topology_watcher_test.go
```diff
@@ -59,6 +59,52 @@ func checkChecksum(t *testing.T, tw *TopologyWatcher, want uint32) {
...
+	tw.Stop()
+	done <- true
+
+	tw.wg.Wait()
+	done <- true
+
+	_, ok := <-result
+	if !ok {
+		t.Fatal("timed out")
+	}
```
deepthi, line 105: nit: we should use `require` instead of `t.Fatal`.
The rest LGTM.

deepthi (approved):

PR #9155, opened 2021-11-06
Collations Integration Ordering

king-11:
## Engine Update

This PR adds collation support to the `engine` module specifically focused on making `comparer` struct collation aware by including collation ID in it which can then be passed down to the `NullSafeCompare` responsible for actually comparing values.

https://github.com/vitessio/vitess/blob/f085f12b1cd38a9bcf907604a747af308e93c4fd/go/vt/vtgate/engine/comparer.go#L26

To be able to pass down collation ID into the `comparer` struct we need to contain it further up the module level i.e. in `OrderByParams`.

https://github.com/vitessio/vitess/blob/f085f12b1cd38a9bcf907604a747af308e93c4fd/go/vt/vtgate/engine/route.go#L128

## Plan Builder Updates

- As the params are populated in `planBuilder` it becomes trivial to extract collation ID in most cases.

### Memory Sort on Aggregation

The plan builder tests were updated so that testing can be done for `createMemorySortPlanOnAggregation`. For that, the `OrderByParams.String()` was updated to include the collation name with `COLLATE` keyword. This changes the` description` of `route` for which tests were updated this also allows us to test for `createMemorySortPlanOnAggregation` using the plan builder tests.

## v3

We don't have the semantic table for the v3 planner code and we intend to only add collations support to Gen4 only, after discussing with @systay @frouioui. Hence the TODOs are left in `ordering.go` and `memory_sort.go`.

## Tests Added

- Route Sort with Collation
- Memory Sort with Collation
- Merge Sort with Collation

## Related Issue(s)

- #8991

## Checklist

### DIFF
--- a/go/vt/vtgate/engine/route_test.go
+++ b/go/vt/vtgate/engine/route_test.go
@@ -22,6 +22,7 @@ import (
 
 	"github.com/stretchr/testify/assert"
 
+	"vitess.io/vitess/go/mysql/collations"
 	vtrpcpb "vitess.io/vitess/go/vt/proto/vtrpc"
 	"vitess.io/vitess/go/vt/vterrors"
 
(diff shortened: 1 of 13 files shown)


go/vt/vtgate/planbuilder/horizon_planning.go
```diff
@@ -837,11 +838,13 @@ func createMemorySortPlanOnAggregation(plan *orderedAggregate, orderExprs []abst
 		if !found {
 			return nil, vterrors.Errorf(vtrpcpb.Code_INTERNAL, "expected to find the order by expression (%s) in orderedAggregate", sqlparser.String(order.Inner))
 		}
+		// TODO(king-11) not sure whether below will work or not.
```
systay, line 841: you should add a unit test and an end-to-end test if you are uncertain

king-11: Sure next task to do in my list for this PR.

king-11: I am not able to find where should I add in tests for this particular change. I added in tests for the `routeSort` change though.

vmg:
The test failures in the unit tests are not flukes. Looks like you need to tweak the code -- a good starting point would be the slice access that is causing a panic.

king-11:
@vmg the issue is resolved i was trying to access KeyCol instead of actual index

go/vt/vtgate/planbuilder/horizon_planning.go
```diff
@@ -833,32 +834,38 @@ func createMemorySortPlanOnAggregation(plan *orderedAggregate, orderExprs []abst
...
 			StarColFixedIndex: offset,
+			CollationID:       collationID,
 		})
 	}
 	return ms, nil
 }
 
-func findExprInOrderedAggr(plan *orderedAggregate, order abstract.OrderBy) (int, int, bool) {
-	for _, key := range plan.eaggr.GroupByKeys {
+func findExprInOrderedAggr(plan *orderedAggregate, order abstract.OrderBy) (int, int, int, bool) {
```
frouioui, line 857: we should name the variables we are returning so we can easily understand what's their purpose

go/vt/vtgate/engine/route_test.go
```diff
@@ -981,6 +982,137 @@ func TestRouteSortWeightStrings(t *testing.T) {
...
+					"c",
+					"d",
+					"cs",
+					"cs",
+					"c",
+				),
+			},
+		}
+		_, err = sel.TryExecute(vc, map[string]*querypb.BindVariable{}, false)
+		require.EqualError(t, err, "types are not comparable: VARCHAR vs VARCHAR")
```
frouioui, line 1086: it would be nice to have a more explicit error, something like:
```
cannot compare strings, collation is unknown or unsupported (collation ID: %d)
```
with the collation ID in `%d`

go/vt/vtgate/planbuilder/horizon_planning.go
king-11: resolved

go/vt/vtgate/engine/route_test.go
king-11: @frouioui  I can make a separate PR addressing the same for `NullSafeCompare` or should I do it in this one only?

frouioui: However works best for you :)

king-11: Alright I will start working on a separate PR as it seems to be needing more changes than I anticipated

frouioui (approved):

PR #9157, opened 2021-11-08
Topo write error in change tablet type

GuptaManan100:
## Description
Topo write error during TabletManager.ChangeType resulting in actual successful topo write led to a bug where the vttablet did not update its state to reflect the change and did not converge with shard sync.

Consider the case where we try to upgrade a replica to a primary tablet. Before we change the internal state of the tablet, we try to write to the topo server.
Assume that the write succeeds but an error is returned from the topo server. In this case, the tablet does not update its local information and continues to believe it is a replica. All the other tablets and vtgate register this change because of shard-sync and make changes accordingly. 
Since we do not start shard-sync for replica tablets, the tablet does not converge to the correct state. 
We are left in a state where the tablet thinks that it is a replica, but everyone else including the topo think that it is a primary.

This issue is addressed in this PR. After a discussion with @sougou, we converged on the solution to wait indefinitely in case of an error while writing to a topo server, until we are able to read the data from it successfully. If the write was successful, we continue with the operations, if the write wasn't successful, we fail the operation like before.

## Checklist

### DIFF
--- a/go/vt/vttablet/tabletmanager/tm_state.go
+++ b/go/vt/vttablet/tabletmanager/tm_state.go
@@ -181,7 +181,24 @@ func (ts *tmState) ChangeTabletType(ctx context.Context, tabletType topodatapb.T
 		// Update the tablet record first.
 		_, err := topotools.ChangeType(ctx, ts.tm.TopoServer, ts.tm.tabletAlias, tabletType, PrimaryTermStartTime)
 		if err != nil {
-			return err
+			log.Errorf("error in changing type in tablet record in topo for tablet %s :- %v", topoproto.TabletAliasString(ts.tm.tabletAlias), err)
+			log.Errorf("will keep trying to read from the topo server")
+			// In case of a topo error, we aren't sure if the data has been written or not.
+			// We must read the data again and verify whether the previous write succeeded or not.
+			// The only way to guarantee safety is to keep retrying read until we succeed
+			for {
+				ti, errInReading := ts.tm.TopoServer.GetTablet(ctx, ts.tm.tabletAlias)
+				if errInReading != nil {
+					time.Sleep(100 * time.Millisecond)
+					continue
+				}
+				if ti.Type == tabletType && proto.Equal(ti.PrimaryTermStartTime, PrimaryTermStartTime) {
+					log.Errorf("read successful. data written to topo-server, continuing operation")
+					break
+				}
+				log.Errorf("read successful. data not written to topo-server, canceling operation")
+				return err
+			}
 		}
 		if action == DBActionSetReadWrite {
 			// We call SetReadOnly only after the topo has been updated to avoid
--- a/go/vt/vttablet/tabletmanager/tm_state.go
+++ b/go/vt/vttablet/tabletmanager/tm_state.go
@@ -181,22 +181,21 @@ func (ts *tmState) ChangeTabletType(ctx context.Context, tabletType topodatapb.T
 		// Update the tablet record first.
 		_, err := topotools.ChangeType(ctx, ts.tm.TopoServer, ts.tm.tabletAlias, tabletType, PrimaryTermStartTime)
 		if err != nil {
-			log.Errorf("error in changing type in tablet record in topo for tablet %s :- %v", topoproto.TabletAliasString(ts.tm.tabletAlias), err)
-			log.Errorf("will keep trying to read from the topo server")
+			log.Errorf("Error changing type in topo record for tablet %s :- %v\nWill keep trying to read from the toposerver", topoproto.TabletAliasString(ts.tm.tabletAlias), err)
 			// In case of a topo error, we aren't sure if the data has been written or not.
 			// We must read the data again and verify whether the previous write succeeded or not.
 			// The only way to guarantee safety is to keep retrying read until we succeed
 			for {
 				ti, errInReading := ts.tm.TopoServer.GetTablet(ctx, ts.tm.tabletAlias)
 				if errInReading != nil {
-					time.Sleep(100 * time.Millisecond)
+					<-time.After(100 * time.Millisecond)
 					continue
 				}
 				if ti.Type == tabletType && proto.Equal(ti.PrimaryTermStartTime, PrimaryTermStartTime) {
-					log.Errorf("read successful. data written to topo-server, continuing operation")
+					log.Infof("Tablet record in toposerver matches, continuing operation")
 					break
 				}
-				log.Errorf("read successful. data not written to topo-server, canceling operation")
+				log.Errorf("Tablet record read from toposerver does not match what we attempted to write, canceling operation")
 				return err
 			}
 		}
--- a/go/vt/vttablet/tabletmanager/tm_state_test.go
+++ b/go/vt/vttablet/tabletmanager/tm_state_test.go
@@ -22,6 +22,8 @@ import (
 	"testing"
 	"time"
 
+	"vitess.io/vitess/go/vt/topo/faketopo"
+
 	"vitess.io/vitess/go/test/utils"
 
 	"vitess.io/vitess/go/vt/key"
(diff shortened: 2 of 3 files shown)


go/vt/vttablet/tabletmanager/tm_state_test.go
```diff
@@ -380,6 +382,76 @@ func TestStateChangeTabletType(t *testing.T) {
...
+			expectedTabletType: topodatapb.TabletType_REPLICA,
+			expectedError:      "deadline exceeded: tablets/cell1-0000000002/Tablet",
+		},
+	}
+
+	for _, testcase := range testcases {
+		t.Run(testcase.name, func(t *testing.T) {
+			factory := faketopo.NewFakeTopoFactory()
+			// add cell1 to the factory. Make it so that the fourth call to update fails with an error. This will be the call from ChangeTabletType.
+			factory.AddCell("cell1", getErrorsList(testcase.numberOfReadErrors) /* update errors */, []bool{false, false, false, true} /* write Persists */, []bool{true, true, true, testcase.writePersists})
```
deepthi, line 412: nit: I believe the comments need to go after the commas.

go/vt/vttablet/tabletmanager/tm_state.go
```diff
@@ -181,7 +181,24 @@ func (ts *tmState) ChangeTabletType(ctx context.Context, tabletType topodatapb.T
 		// Update the tablet record first.
 		_, err := topotools.ChangeType(ctx, ts.tm.TopoServer, ts.tm.tabletAlias, tabletType, PrimaryTermStartTime)
 		if err != nil {
-			return err
+			log.Errorf("error in changing type in tablet record in topo for tablet %s :- %v", topoproto.TabletAliasString(ts.tm.tabletAlias), err)
```
deepthi, line 184: Sorry to be nitpicky, but I'm trying to rephrase the errors to be more concise. Maybe "error changing type in tablet record for tablet"..

deepthi: Or even "error changing type in topo record for tablet"

go/vt/vttablet/tabletmanager/tm_state.go
```diff
@@ -181,7 +181,24 @@ func (ts *tmState) ChangeTabletType(ctx context.Context, tabletType topodatapb.T
...
+			log.Errorf("will keep trying to read from the topo server")
```
deepthi, line 185: This can be combined into the previous log line with a comma or something similar separating the two parts of the error message.

go/vt/vttablet/tabletmanager/tm_state.go
```diff
@@ -181,7 +181,24 @@ func (ts *tmState) ChangeTabletType(ctx context.Context, tabletType topodatapb.T
...
+			// In case of a topo error, we aren't sure if the data has been written or not.
+			// We must read the data again and verify whether the previous write succeeded or not.
+			// The only way to guarantee safety is to keep retrying read until we succeed
+			for {
+				ti, errInReading := ts.tm.TopoServer.GetTablet(ctx, ts.tm.tabletAlias)
+				if errInReading != nil {
+					time.Sleep(100 * time.Millisecond)
```
deepthi, line 192: As we discussed when we looked at this together, this should be replaced with a timer.

go/vt/vttablet/tabletmanager/tm_state.go
```diff
@@ -181,7 +181,24 @@ func (ts *tmState) ChangeTabletType(ctx context.Context, tabletType topodatapb.T
...
+					continue
+				}
+				if ti.Type == tabletType && proto.Equal(ti.PrimaryTermStartTime, PrimaryTermStartTime) {
+					log.Errorf("read successful. data written to topo-server, continuing operation")
```
deepthi, line 196: Info seems the appropriate log level for this.
also topo-server -> toposerver

deepthi: Do we really need to say "read successful"? that is an implementation detail, right? what matters is that the topo state matches what we expect.
"Tablet record in toposerver matches, continuing .."??

go/vt/vttablet/tabletmanager/tm_state.go
```diff
@@ -181,7 +181,24 @@ func (ts *tmState) ChangeTabletType(ctx context.Context, tabletType topodatapb.T
...
+					break
+				}
+				log.Errorf("read successful. data not written to topo-server, canceling operation")
```
deepthi, line 199: Here too, it could be something like "Tablet record read from toposerver does not match what we attempted to write, canceling ..."

deepthi (commented):
The fix looks good except for the sleep vs timer. Everything else is UX :smile:

go/vt/vttablet/tabletmanager/tm_state_test.go
GuptaManan100: Fixed

go/vt/vttablet/tabletmanager/tm_state.go
GuptaManan100: Fixed

go/vt/vttablet/tabletmanager/tm_state.go
GuptaManan100: Fixed

go/vt/vttablet/tabletmanager/tm_state.go
GuptaManan100: Fixed

go/vt/vttablet/tabletmanager/tm_state.go
GuptaManan100: Fixed

go/vt/vttablet/tabletmanager/tm_state.go
GuptaManan100: Fixed

go/vt/vttablet/tabletmanager/tm_state.go
```diff
@@ -181,7 +181,23 @@ func (ts *tmState) ChangeTabletType(ctx context.Context, tabletType topodatapb.T
...
 		if err != nil {
-			return err
+			log.Errorf("Error changing type in topo record for tablet %s :- %v\nWill keep trying to read from the toposerver", topoproto.TabletAliasString(ts.tm.tabletAlias), err)
+			// In case of a topo error, we aren't sure if the data has been written or not.
+			// We must read the data again and verify whether the previous write succeeded or not.
+			// The only way to guarantee safety is to keep retrying read until we succeed
+			for {
+				ti, errInReading := ts.tm.TopoServer.GetTablet(ctx, ts.tm.tabletAlias)
+				if errInReading != nil {
+					<-time.After(100 * time.Millisecond)
```
deepthi, line 191: This is even more compact than my original suggestion. +1

deepthi (approved):

go/vt/vttablet/tabletmanager/tm_state.go
GuptaManan100: 💕

go/vt/vttablet/tabletmanager/tm_state.go
```diff
@@ -181,7 +181,23 @@ func (ts *tmState) ChangeTabletType(ctx context.Context, tabletType topodatapb.T
...
+					continue
+				}
```
sonne5, line 193: Thanks for fixing this issue @GuptaManan100 @deepthi! 

One small comment: we usually run into this when topo server is unhealthy, the loop here would add more loads on the leader of topo server. Not sure if there is any existing pattern that we can borrow to do an exponential backoff until certain max wait time? 

(cc @aquarapid for visibility)

GuptaManan100: Yes, that sounds like a good idea to me!

PR #9158, opened 2021-11-08
OnlineDDL: formalize artifact cleanup time via retain_artifacts_seconds column

shlomi-noach:
We add the column `retain_artifacts_seconds` to `schema_migrations` table. This will allow us to control, per migration, how long we must retain the artifacts after migration is complete.
In turn, this implies: how long after migration completion are we still able to [revert](https://vitess.io/docs/user-guides/schema-changes/revertible-migrations/) it?

Cleanup now checks the value of `retain_artifacts_seconds`. For backwards compatibility (in-place upgrades of vttablets) the value `0` implies "use the flag `-retainOnlineDDLTables`". Any non-zero value overrides that (e.g. a negative `-1` really means "yes this migration can be cleaned up right now").

In later steps (separate PRs) we can:

- Support a `ALTER VITESS_MIGRATION '<uuid>' CLEANUP` to suggest/request an earlier cleanup than desired
- Allow custom retain time, e.g. as a new flag in `ddl_strategy`

In both cases the effect will be achieved by updating the value of `retain_artifacts_seconds` in the table.

## Checklist


dbussink (approved):

shlomi-noach:
Tracking issue: #6926

PR #9166, opened 2021-11-09
go dependency: updated containerd to 1.5.7

shlomi-noach:
## Description

This CVE: https://github.com/advisories/GHSA-c2h3-6mxw-7mvq gives warnings for `containerd 1.5.5` and recommended upgrading to `1.5.7`. This PR updates to `1.5.7`.
The dependency on `containerd` comes from `planetscale/tengo`, but really is only used by `tengo`'s CI/testing, and has no operational/production significance.

## Related Issue(s)

https://github.com/advisories/GHSA-c2h3-6mxw-7mvq

## Checklist


go.sum
```diff
@@ -236,8 +237,9 @@ github.com/containerd/containerd v1.5.0-beta.3/go.mod h1:/wr9AVtEM7x9c+n0+stptlo
 github.com/containerd/containerd v1.5.0-beta.4/go.mod h1:GmdgZd2zA2GYIBZ0w09ZvgqEq8EfBp/m3lcVZIvPHhI=
 github.com/containerd/containerd v1.5.0-rc.0/go.mod h1:V/IXoMqNGgBlabz3tHD2TWDoTJseu1FGOKuoA4nNb2s=
 github.com/containerd/containerd v1.5.1/go.mod h1:0DOxVqwDy2iZvrZp2JUx/E+hS0UNTVn7dJnIOwtYR4g=
-github.com/containerd/containerd v1.5.5 h1:q1gxsZsGZ8ddVe98yO6pR21b5xQSMiR61lD0W96pgQo=
 github.com/containerd/containerd v1.5.5/go.mod h1:oSTh0QpT1w6jYcGmbiSbxv9OSQYaa88mPyWIuU79zyo=
+github.com/containerd/containerd v1.5.7 h1:rQyoYtj4KddB3bxG6SAqd4+08gePNyJjRqvOIfV3rkM=
+github.com/containerd/containerd v1.5.7/go.mod h1:gyvv6+ugqY25TiXxcZC3L5yOeYgEw0QMhscqVp1AR9c=
```
deepthi, line 242: any idea why the previous version is still present in `go.sum`?
I see something similar for the other changed modules too.

shlomi-noach: It seems like it comes as a dependency by `github.com/fsouza/go-dockerclient@1.7.4`. In `main`, that package does actually now refer to `github.com/containerd/containerd@v1.5.7`, but there's no release yet.

In th epast, what we did was to strip out the entire docker stuff from `planetsclae/tengo`, https://github.com/planetscale/tengo/pull/5

We recently updated from `upstream` and re-acquired the docker files; they're useful for testing, but not for production, which is a bummer because why lose good testing, I wish we could do both. Let me look into it some more.

shlomi-noach: OK, what I'm gonna do is create a new tag in `planetscale/tengo` which is specific to vitess, and has docker stripped out.

shlomi-noach: Live blogging. I see there' already a `v0.10.3-ps.v1-vitess`

shlomi-noach:
Change of plans: I see @dbussink actually created https://github.com/planetscale/tengo/releases/tag/v0.10.3-ps.v1-vitess, a version stripped of docker/containerd - which we now use, eliminating imports for `containerd` altogether.

shlomi-noach:
@deepthi I believe this is good to go.

deepthi (approved):

PR #9172, opened 2021-11-09
Improve self hosted runner docs

GuptaManan100:
## Description
While using the self hosted runners today to check some flaky tests, I realised that there are a few steps that I am using frequently like copying the files from the docker container to the server and then finally scp them to my local system for ease of use. These steps aren't documented and are really useful for anyone trying to use the self-hosted runners to fix flaky tests.

This PR adds those missing steps and improves the docs for using self-hosted runners.

## Checklist


systay (approved):

frouioui (approved):
LGTM. You should add maintainers' SSH keys to the self hosted runners as well, so we can access them

GuptaManan100:
Well they also need access to the equinx dashboard to get the IP address to ssh to. IIRC, the policy that we had opted for was to add people when they ask for that access.

harshit-gangal (approved):

PR #9176, opened 2021-11-09
Add option to run vtcombo and vttest local cluster with real topo server

zhongr3n:
Signed-off-by: Zhong Ren <zhong@zhongren.me>

## Description

Add option to run vtcombo and vttest local cluster with real topo server
- We want to share the topo server across multiple components for local functional tests. Because vtcombo starts its own process, there's no easy way to share the in-memory topo server across different components
- This change allow vtcombo to use an actual topo server endpoint
- Tested with local consul server

```
consul agent -dev

go run ~/stripe/vitess/go/cmd/vttestserver/main.go -remote_topo_implementation=consul -remote_topo_global_server_address=localhost:8500 -remote_topo_global_root=root
```

## Related Issue(s)

N/A

## Checklist

## Deployment Notes

N/A

### DIFF
--- a/go/cmd/vtcombo/main.go
+++ b/go/cmd/vtcombo/main.go
@@ -59,6 +59,9 @@ var (
 
 	mysqlPort = flag.Int("mysql_port", 3306, "mysql port")
 
+	remoteTopoServer = flag.Bool("remote_topo_server", false, "Should vtcombo use a remote topology server instead of starting its own in-memory topology server. "+
+		"If true, vtcombo will use the flags defined in topo/server.go to open topo server")
+
 	ts              *topo.Server
 	resilientServer *srvtopo.ResilientServer
 )
@@ -135,8 +138,14 @@ func main() {
 		flag.Set("log_dir", "$VTDATAROOT/tmp")
 	}
 
-	// Create topo server. We use a 'memorytopo' implementation.
-	ts = memorytopo.NewServer(tpb.Cells...)
+	if *remoteTopoServer {
+		// Open topo server based on the command line flags defined at topo/server.go
+		// do not create cell info as it should be done by whoever sets up the remote topo server
+		ts = topo.Open()
+	} else {
+		// Create topo server. We use a 'memorytopo' implementation.
+		ts = memorytopo.NewServer(tpb.Cells...)
+	}
 	servenv.Init()
 	tabletenv.Init()
 
--- a/go/cmd/vtcombo/main.go
+++ b/go/cmd/vtcombo/main.go
@@ -59,7 +59,7 @@ var (
 
 	mysqlPort = flag.Int("mysql_port", 3306, "mysql port")
 
-	remoteTopoServer = flag.Bool("remote_topo_server", false, "Should vtcombo use a remote topology server instead of starting its own in-memory topology server. "+
+	externalTopoServer = flag.Bool("external_topo_server", false, "Should vtcombo use an external topology server instead of starting its own in-memory topology server. "+
 		"If true, vtcombo will use the flags defined in topo/server.go to open topo server")
 
 	ts              *topo.Server
@@ -138,9 +138,9 @@ func main() {
 		flag.Set("log_dir", "$VTDATAROOT/tmp")
 	}
 
-	if *remoteTopoServer {
+	if *externalTopoServer {
 		// Open topo server based on the command line flags defined at topo/server.go
-		// do not create cell info as it should be done by whoever sets up the remote topo server
+		// do not create cell info as it should be done by whoever sets up the external topo server
 		ts = topo.Open()
 	} else {
 		// Create topo server. We use a 'memorytopo' implementation.
--- a/go/cmd/vttestserver/vttestserver_test.go
+++ b/go/cmd/vttestserver/vttestserver_test.go
@@ -22,15 +22,16 @@ import (
 	"io"
 	"math/rand"
 	"os"
+	"os/exec"
 	"path"
 	"strings"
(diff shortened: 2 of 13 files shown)


go/vt/vttest/local_cluster.go
```diff
@@ -237,6 +245,16 @@ func (db *LocalCluster) Setup() error {
 
 	log.Infof("LocalCluster environment: %+v", db.Env)
 
+	// Set up topo manager if we are using a remote topo server
+	if db.RemoteTopoImplementation != "" {
+		db.topo = db.Env.TopoManager(db.RemoteTopoImplementation, db.RemoteTopoGlobalServerAddress, db.RemoteTopoGlobalRoot, db.Topology)
+		log.Infof("Initializing Topo Manager (%T)...", db.topo)
```
mattlord, line 251: This will print the golang type. Is that what we want here? I would think this will always be `TopoManager` but we are interested in the implementation details?

go/vt/vttest/local_cluster.go
```diff
@@ -237,6 +245,16 @@ func (db *LocalCluster) Setup() error {
...
+		if err := db.topo.Setup(); err != nil {
+			log.Errorf("Failed to set up Topo Manager: %s", err)
```
mattlord, line 253: Nit, but I think it's better to use the %v formatter with err or with %s use `err.Error()`

mattlord (approved):
Thank you for doing this! I can definitely see how this will be useful. I could put it to good use myself for some consul specific testing :-) 

I had a few nits, but all very minor things. Do you think it's worth adding a unit test to be sure this doesn't break down the road?

go/vt/vttest/local_cluster.go
deepthi: +1

go/cmd/vtcombo/main.go
```diff
@@ -135,8 +138,14 @@ func main() {
 		flag.Set("log_dir", "$VTDATAROOT/tmp")
 	}
 
-	// Create topo server. We use a 'memorytopo' implementation.
-	ts = memorytopo.NewServer(tpb.Cells...)
+	if *remoteTopoServer {
+		// Open topo server based on the command line flags defined at topo/server.go
```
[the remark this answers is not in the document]
deepthi: I think this is ok. All flags are in one place which makes it easier to find them.

deepthi (commented):
Looks good overall.
I might be reading this wrong, but I don't see any actual tests exercising the new options. We need at least one.

go/cmd/vtcombo/main.go
```diff
@@ -59,6 +59,9 @@ var (
 
 	mysqlPort = flag.Int("mysql_port", 3306, "mysql port")
 
+	remoteTopoServer = flag.Bool("remote_topo_server", false, "Should vtcombo use a remote topology server instead of starting its own in-memory topology server. "+
```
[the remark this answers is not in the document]
zhongr3n: yeah sure. updating

go/vt/vttest/local_cluster.go
zhongr3n: good catch. fixed to use `%+v`

zhongr3n:
Thanks for reviewing folks. Added a unit test using consul external topo, and addressed other feedback.

go/vt/topo/test/testing.go
```diff
@@ -112,3 +122,90 @@ func TopoServerTestSuite(t *testing.T, factory func() *topo.Server) {
...
+			"serf_wan": port + 3,
+		},
+	}
+
+	if authToken != "" {
+		config["datacenter"] = "vitess"
+		config["acl_datacenter"] = "vitess"
+		config["acl_master_token"] = authToken
+		config["acl_default_policy"] = "deny"
+		config["acl_down_policy"] = "extend-cache"
```
sonne5, line 161: these are all deprecated flags, should we consider keeping `startConsul` as is (since it's testing some acl configs) and start consul via a separated minimum cli e.g., `exec.Command("consul", "agent", "-dev")`

go/cmd/vttestserver/vttestserver_test.go
```diff
@@ -308,6 +311,39 @@ func TestMtlsAuthUnauthorizedFails(t *testing.T) {
 	assert.Contains(t, err.Error(), "code = Unauthenticated desc = client certificate not authorized")
 }
 
+func TestExternalTopoServerConsul(t *testing.T) {
+	// Start a single consul in the background.
+	cmd, configFilename, serverAddr := test.StartConsul(t, "")
+	defer func() {
+		// Alerts command did not run successful
+		if err := cmd.Process.Kill(); err != nil {
```
sonne5, line 319: if we use `consul agent -dev` i think we could do `consul leave` for a graceful shutdown and don't need to call `os.Remove`

go/vt/topo/test/testing.go
zhongr3n: updated to use its own simplified `startConsul`

go/cmd/vttestserver/vttestserver_test.go
zhongr3n: tried `consul leave` and it adds 10 seconds for grace shutdown. keeping process.kill

zhongr3n:
looking into test failure

deepthi (approved):

sonne5:
LGTM

sonne5 (approved):

deepthi:
Thank you for the contribution!

PR #9175, opened 2021-11-09
Fix boolean parameter order in DropSources call for v2 flows

ajm188:
Signed-off-by: Andrew Mason <amason@slack-corp.com>

## Description

This corrects the boolean param passing order to DropSources made by v2 vreplication flows

## Related Issue(s)

Fixes #9174

## Checklist
- [ ] Should this PR be backported? **yes**
- [x] Tests were added or are not required
- [x] Documentation was added or is not required

Issue #9174, opened 2021-11-09
-keep_data is ignored in v2 vreplication workflows


ajm188:
#### Overview of the Issue

The v2 workflows have swapped the `-force` and `-keep_data` arguments passed to `wr.DropSources` meaning that in order to keep data, you need to pass `-force`, and in order to force, you need to pass `-keep_data`.

[v1 method call][v1_dropsources]: `wr.DropSources(ctx, keyspace, workflowName, removalType, *keepData, false, *dryRun)`
[v2 method call][v2_dropsources]: `wr.DropSources(vrw.ctx, vrw.ws.TargetKeyspace, vrw.ws.Workflow, renameTable, false, vrw.params.KeepData, vrw.params.DryRun)`

[v1_dropsources]: https://github.com/vitessio/vitess/blob/131148d3afb277d553451e93c47e3ed22490be57/go/vt/vtctl/vtctl.go#L2841
[v2_dropsources]: https://github.com/vitessio/vitess/blob/131148d3afb277d553451e93c47e3ed22490be57/go/vt/wrangler/workflow.go#L326-L327

#### Reproduction Steps

```
vtctlclient -server "$server" Reshard -v2 -dry_run -keep_data Complete "mykeyspace.reshard"
Dry Run results for 02 Jan 06 15:04 MST run at Complete
Parameters: -v2 -dry_run -keep_data Complete mykeyspace.reshard

Lock keyspace mykeyspace
Deleting following shards (and all related tablets):
		[tablet aliases omitted]
	Cell mycell Keyspace mykeyspace Shard
$shard
Delete reverse vreplication streams on source:
[more output deleting streams, etc]
```

#### Binary version

We're experiencing this in 10.0, but git history suggests that this is still an issue in `main`, and therefore probably also 11.0 and 12.0.

PR #9178, opened 2021-11-10
[backport] Fix boolean parameter order in DropSources call for v2 flows

ajm188:
## Description

This PR backports #9175 to the 12.0 release branch

## Related Issue(s)

#9174 

## Checklist
- [x] Should this PR be backported? **no** (this is the backport)


deepthi (commented):
Can you state a justification for the backport?

ajm188:
I figured we should backport because anyone using the v2 flows (which are now the default) can't use the `-keep_data` option for `MoveTables/Reshard Complete`. Switching back to the v1 flows is a completely valid workaround, but right now the default commands are broken for this mode of operation.

deepthi (approved):
Approving the back port.
This affects migrations to vitess from existing prod systems. For any number of reasons, you might want to keep the existing system around and not drop it.

PR #9182, opened 2021-11-10
Inclusive naming: master->primary in topo package

### LINKED ISSUE #8620: Task: replace master->primary in topo package
The `Conn` interface has functions with `Master` in the name. They should all be renamed.
Creating this as a separate task so that we can investigate whether there are any external dependencies to worry about before doing the rename.

deepthi:
## Description
- Rename functions in topo package
- Deprecate some stats - `NewMasterParticipation`
- Add new stats as replacements - `NewLeaderParticipation`

Note: I verified that vitess-operator does not use any of the functions that are being renamed.

## Related Issue(s)

Fixes #8620 

## Checklist

### DIFF
--- a/go/vt/topo/consultopo/server_flaky_test.go
+++ b/go/vt/topo/consultopo/server_flaky_test.go
@@ -68,6 +68,9 @@ func startConsul(t *testing.T, authToken string) (*exec.Cmd, string, string) {
 		},
 	}
 
+	// TODO(deepthi): this is the legacy ACL format. We run v1.4.0 by default in which this has been deprecated.
+	// We should start using the new format
+	// https://learn.hashicorp.com/tutorials/consul/access-control-replication-multiple-datacenters?in=consul/security-operations
 	if authToken != "" {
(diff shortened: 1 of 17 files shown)


go/vt/topo/consultopo/server_flaky_test.go
```diff
@@ -68,6 +68,9 @@ func startConsul(t *testing.T, authToken string) (*exec.Cmd, string, string) {
 		},
 	}
 
+	// TODO(deepthi): this is the legacy ACL format. We run v1.4.0 by default in which this has been deprecated.
+	// We should start using the new format
+	// https://learn.hashicorp.com/tutorials/consul/access-control-replication-multiple-datacenters?in=consul/security-operations
 	if authToken != "" {
 		config["datacenter"] = "vitess"
 		config["acl_datacenter"] = "vitess"
```
deepthi, line 76: @gedgar @sonne5 is this something one of you can fix?

mattlord (approved):
❤️

PR #9189, opened 2021-11-11
Do keyspace filtering when possible with `show vitess_shards`

aquarapid:
When using a `show vitess_shards like 'keyspace1/%'`
filter, actually use the keyspace part of the query to limit the
keyspaces for which we fetch all shards, instead of getting all
shards from all keyspaces the vtgate knows about and then
filtering them out. This probably does not matter unless you
have a lot of shards and/or are running `show vitess_shards`
like this frequently.

### DIFF
--- a/go/vt/vtgate/executor_test.go
+++ b/go/vt/vtgate/executor_test.go
@@ -716,7 +716,7 @@ func TestExecutorShow(t *testing.T) {
 	}
 	utils.MustMatch(t, wantqr, qr, query)
 
-	query = "show vitess_shards like 'TestSharded%/%'"
+	query = "show vitess_shards like 'TestShard%/%'"
 	qr, err = executor.Execute(ctx, "TestExecute", session, query, nil)
 	require.NoError(t, err)
 
--- a/go/vt/vtgate/executor_test.go
+++ b/go/vt/vtgate/executor_test.go
@@ -701,6 +701,36 @@ func TestExecutorShow(t *testing.T) {
 	}
 	utils.MustMatch(t, wantqr, qr, query)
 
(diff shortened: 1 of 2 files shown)


zhongr3n:
Thanks for quick fix. This looks good!

One minor note: can you add a test case for something like `show vitess_shards like 'keyspace%/%'`. think it should work as expected but would be good to verify

go/vt/vtgate/executor_test.go
```diff
@@ -701,6 +701,36 @@ func TestExecutorShow(t *testing.T) {
...
+	wantqr = &sqltypes.Result{
+		Fields: buildVarCharFields("Shards"),
+		Rows: [][]sqltypes.Value{
+			buildVarCharRow("TestSharded/-20"),
+			buildVarCharRow("TestSharded/e0-"),
+		},
+	}
+	utils.MustMatch(t, wantqr, qr, query)
+
+	query = "show vitess_shards like 'TestSharded%/%'"
```
deepthi, line 719: nit: the test will be better if you used 'TestShard%/%' to exercise the regex.

deepthi (approved):

systay (approved):

go/vt/vtgate/executor_test.go
zhongr3n: +1

zhongr3n (approved):

frouioui (approved):

go/vt/vtgate/executor_test.go
aquarapid: Made that change.

PR #9191, opened 2021-11-11
make sure to copy bindvars when using them concurrently

### LINKED ISSUE #9139: Panic (concurrent map iteration and map write) in CopyBindVariables using vtcombo
A user reported this problem running against vttestserver/vtcombo (from the upstream docker containers) from a few weeks ago.

The scenario was running a load of queries like:

```
(SELECT c1 AS name FROM t1 LIMIT 1)
UNION
(SELECT c2 AS name FROM t2 LIMIT 1)
UNION
(SELECT c3 AS name FROM t3 LIMIT 1)
UNION
(SELECT c4 AS name FROM t4 LIMIT 1)
UNION
(SELECT c5 AS name FROM t5 LIMIT 1)
```

where t1,t2,t3,t4,t5 are all sharded tables.

Panic attached.

Brief look seems like we are somehow modifying the bind variables involved while copying them internally between the vtgate and vttablet "parts" of vtcombo.  It is therefore likely that this does not affect Vitess itself.

[panic.txt](https://github.com/vitessio/vitess/files/7471633/panic.txt)

systay:
## Description
When executing `UNION` queries, the individual queries are sent concurrently to the tablets. For `information_schema`  queries, the bind variables are changed before sending them down to the tablet. These two features together lead to one map being iterated over while being updated, which is a no-no that will panic.

The solution is to make a copy of the bindvars before executing the individual queries of the `UNION`.

## Related Issue(s)
Fixes #9139

### DIFF
--- a/go/test/endtoend/vtgate/misc_test.go
+++ b/go/test/endtoend/vtgate/misc_test.go
@@ -637,3 +637,159 @@ func TestSavepointInReservedConn(t *testing.T) {
 	defer utils.Exec(t, conn, `delete from t7_xxhash`)
 	utils.AssertMatches(t, conn, "select uid from t7_xxhash", `[[VARCHAR("2")]]`)
 }
+
+func TestUnionWithManyInfSchemaQueries(t *testing.T) {
+	// trying to reproduce the problems in https://github.com/vitessio/vitess/issues/9139
+	defer cluster.PanicHandler(t)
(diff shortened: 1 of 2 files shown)


GuptaManan100 (approved):

arthurschreiber:
Oh, thanks, I think this fixes one of the issues we've been running into! ❤️

PR #9171, opened 2021-11-09
OnlineDDL: implementing -postpone-completion, ALTER VITESS_MIGRATION ... COMPLETE

shlomi-noach:
This PR implements:

## -postpone-completion
`-postpone-completion` flag in `ddl_strategy`. This flag indicates that the migration should not auto-complete. This applies for:
- any `CREATE TABLE`
- any `DROP TABLE`
- `ALTER` table in `online` strategy
- `ALTER` table in `gh-ost` strategy

Note that this flag is not supported for `pt-osc` strategy.

Behavior of migrations with this flag:
- an `ALTER` table begins, runs, but does not cut-over.
- `CREATE` or `DROP` migrations are silently not even scheduled

Complimenting this flag is:

## ALTER VITESS_MIGRATION '-uuid-' COMPLETE

This command indicates that a migration executed with `-postpone-completion` is good to complete. Behavior:

- For running `ALTER`s (`online` and `gh-ost`) which are ready to cut-over: cut-over imminently (though not immediately - cut-over depends on polling interval, replication lag, etc)
- For running `ALTER`s (`online` and `gh-ost`) which are only partly through the migration: they will cut-over automatically when they complete their work, as if `-postpone-completion` wasn't indicated
- For queued `CREATE` and `DROP` migrations: "unblock" them from being scheduled. They'll be scheduled at the schedulers discretion. there is no guarantee that they will be scheduled to run immediately.
---

tests added:

- `vrepl` endtoend
- `gh-ost` endtoend
---

This PR extends https://github.com/vitessio/vitess/pull/9160 (not yet merged at this time)
Tracking issue: #6926

PR #9192, opened 2021-11-11
Concurrent migrations

shlomi-noach:
**See updated https://github.com/vitessio/vitess/pull/9192#issuecomment-968706668**

## Description

Ongoing work towards supporting concurrent REVERT migrations.

Up till now, Vitess would only ever run a single migration at a time, for a given keyspace.

- This isn't 100% correct, as a tablet failure could end up resurrecting an already vreplication stream.

Anyway, we wish to support the following:

- REVERT migrations, which are by nature high priority migrations, should be able to run concurrently to any other migration
- We will require an additional flag in `ddl_strategy`, something like `-allow-concurrent` or similar
- There can be an unlimited number of concurrent REVERT migrations with this flag
- But still only one single "normal" migration at a time.

This changes the scheduler logic, which up till now assumed that if _any_ migration was running, then no need to schedule the next one. From now on, it will need to consider these scenarios:

- if a migration is queued, it can be made ready if nothing else is running
- if a "allowed concurrent" migration is queued, it can be made ready is no other migration operates _on the same table_
- if any migration is queued, it can be made ready if only other running migrations are "allowed concurrent" and none of them operates on _same table_

This means a more programmatic approach to scheduling next migration (where today some of the logic is a simple SQL)

## Related Issue(s)

- This PR extends https://github.com/vitessio/vitess/pull/9171 (unmerged yet)
- tracking issue: #6926

## Checklist


shlomi-noach:
While #9171 is still unmerged, this is the diff for this branch: https://github.com/planetscale/vitess/compare/online-ddl-postpone-completion...planetscale:online-ddl-concurrent-revert-migrations

shlomi-noach:
This PR has evolved a bit. We now support concurrent migrations as follows:

- A migration needs to opt in to run concurrently via `-allow-concurrent` flag in `ddl_strategy`
- A migration is scheduled and executed concurrently if:
  - It specifies `-allow-concurrent`
  - Is `CREATE`, or `DROP`, or `REVERT` (`REVERT` works for `CREATE`, `DROP` and for `online`/vrplication `ALTER` migrations)
  - There is no other running migration operating on same table.

With the new scheduler:

- There can be as many concurrent migrations running at any given time
- And there can be at most one non-concurrent ("regular") migration running at any time
- And never two migrations operating on the same table

Also:

- There can be multiple migrations in `ready` state (previously only one)
- A migration can be "blocked" from running, if it conflicts with a running migration (operating on same table, or non-concurrent)
- While some `ready` migrations can be blocked, new migrations are still able to proceed to `ready` state an then to `running`, i.e. actually execute.

Just some possible scenarios:

- If all migrations are non-concurrent, i.e. are regular, then they can only run sequentially.
- If a regular migration is running, it's possible for a `-allow-concurrent` migration to also kick in and run
  - and then it's also possible for yet another N `-allow-concurrent` migrations to kick in and run
- If one or more `-allow-concurrent` migrations are running, it is possible for a regular migration to kick in and run
-

shlomi-noach:
This is now ready to review, given https://github.com/vitessio/vitess/pull/9171 is merged.

go/vt/vttablet/onlineddl/executor.go
```diff
@@ -2179,45 +2235,67 @@ func (e *Executor) executeMigration(ctx context.Context, onlineDDL *schema.Onlin
 	return nil
 }
 
+// runNextMigration picks one 'ready' migration that is able to run, and executes it.
```
shlomi-noach, line 2238: This is at the heart of the PR: the scheduler here gets to decide whether there's a migration thta can be executed, concurrently or not concurrently, based on existence or non-existence of running migrations.

go/vt/vttablet/onlineddl/executor.go
```diff
@@ -306,15 +308,86 @@ func (e *Executor) triggerNextCheckInterval() {
...
+		}
+		return true
 	})
-	return migrationFound
+	return sameTableMigrationFound
+}
+
+// isAnyNonConcurrentMigrationRunning sees if there's any migration running right now
+// that does not have -allow-concurrent.
+// such a running migration will for example prevent a new non-concurrent migration from running.
```
deepthi, line 380: copy-pasta. Need the correct description of this function instead.

go/vt/vttablet/onlineddl/executor.go
```diff
@@ -1561,36 +1634,15 @@ func (e *Executor) CancelPendingMigrations(ctx context.Context, message string)
 }
 
 // scheduleNextMigration attemps to schedule a single migration to run next.
-// possibly there's no migrations to run. Possibly there's a migration running right now,
-// in which cases nothing happens.
+// possibly there's no migrations to run.
```
deepthi, line 1637: nit: there are

go/vt/vttablet/onlineddl/executor.go
```diff
@@ -1561,36 +1634,15 @@ func (e *Executor) CancelPendingMigrations(ctx context.Context, message string)
...
+// The effect of this function is to move a migration from 'queued' state to 'ready' state, is all.
+// Notice that the query sqlScheduleSingleMigration embeds some logic inside. We may choose
+// to refactor the logic into the app
```
deepthi, line 1640: hmm.. which app is this referring to?

go/vt/vttablet/onlineddl/executor.go
```diff
@@ -2230,48 +2286,67 @@ func (e *Executor) executeMigration(ctx context.Context, onlineDDL *schema.Onlin
 	return nil
 }
 
+// runNextMigration picks one 'ready' migration that is able to run, and executes it.
+// Possible scenarios:
+// - no migration is in 'ready' state -- nothing to be done
+// - a migration is 'ready', but conflicts with other running migrations -- try another 'ready' migration
+// It is therefore possible that there is a 'ready' migration, and still this function only handles one.
```
deepthi, line 2293: This sentence is confusing. I would have thought that it is possible that there is a "ready" migration and still this function chooses none.

go/vt/vttablet/onlineddl/executor.go
```diff
@@ -2230,48 +2286,67 @@ func (e *Executor) executeMigration(ctx context.Context, onlineDDL *schema.Onlin
...
 
-	if e.isAnyMigrationRunning() {
-		return ErrExecutorMigrationAlreadyRunning
+	if !e.reviewedRunningMigrationsFlag {
+		// Since Open(), we havent's once executed reviewRunningMigrations() successfully.
+		// This means we may not have a good picture of what is actually running. Perhaps there's
+		// a vreplication migration from a pre-PRS/ERS that we still need to learn about?
+		// We're going to be careful here, and avoid running new migrations until we have
+		// a better picture. It will likely take a couple seconds till next iteration.
+		// execution. This delay ony takes place shortly after Open().
```
deepthi, line 2305:
```suggestion
		// This delay only takes place shortly after Open().
```

go/vt/vttablet/onlineddl/executor.go
```diff
@@ -2230,48 +2286,67 @@ func (e *Executor) executeMigration(ctx context.Context, onlineDDL *schema.Onlin
...
-			UUID:           row["migration_uuid"].ToString(),
-			Strategy:       schema.DDLStrategy(row["strategy"].ToString()),
-			Options:        row["options"].ToString(),
-			Status:         schema.OnlineDDLStatus(row["migration_status"].ToString()),
-			Retries:        row.AsInt64("retries", 0),
-			TabletAlias:    row["tablet"].ToString(),
-			RequestContext: row["migration_context"].ToString(),
+	// getNonConflictingMigration finds a single 'ready' migration which does not conflict with running migrations.
+	// Conflicts are:
+	// - a migration is 'ready' but is not set to run _concurrnetly_, and there's a running migration that is also non-concurrent
```
deepthi, line 2311:
```suggestion
	// - a migration is 'ready' but is not set to run _concurrently_, and there's a running migration that is also non-concurrent
```

go/vt/vttablet/onlineddl/executor.go
```diff
@@ -2230,48 +2286,67 @@ func (e *Executor) executeMigration(ctx context.Context, onlineDDL *schema.Onlin
...
+				return onlineDDL, err
 			}
 		}
-		e.executeMigration(ctx, onlineDDL)
-		// the query should only ever return a single row at the most
-		// but let's make it also explicit here that we only run a single migration
-		if i == 0 {
-			break
+		// no non-conflicting migration found...
+		// Either all ready migrations are conflicting, or there's no ready migrations...
```
deepthi, line 2330: there's no ready migration OR there are no ready migrations.

deepthi (approved):
Feedback is all on code comments.
May merge after fixing them.

go/vt/vttablet/onlineddl/executor.go
shlomi-noach: Clarified, moved explanation into the function itself. This is more a developer's note for future me/us.

go/vt/vttablet/onlineddl/executor.go
shlomi-noach: Oh wow the wording was really bad. Changed to:

```
// Note that per the above breakdown, and due to potential conflicts, it is possible to have one or
// more 'ready' migration, and still none is executed.
```

shlomi-noach:
Addressed all comments.

shlomi-noach:
Woohoo! That's an important one merged.

PR #9205, opened 2021-11-12
Explicit Collation Error

king-11:
## Description

While working on #9155 it was suggested by @frouioui that we can use better error formatting to explicitly let the user know about the error being caused due to missing collation information or it is an unsupported collation.

Earlier the case for missing collation information was being suggested as a `UnsupportedComparisonError` error but given collation support is inbound we should update the error message to explicitly tell in case of `text` values that collation information is missing using `UnsupportedCollationError`

To support this the following was done:
- Error message was updated in `UnsupportedCollationError`
- `if` check with `text` types now also handles the case for `collation.Unknown`
- Accommodate the return of `UnsupportedCollationError` in `ordered_aggregate.go` and `comparer.go` so that they can use weight string in case `UnsupportedCollationError` or `UnsupportedComparisonError` is being returned
- Update the tests to compare with newly formated string

Commits Made:
- feat: explicit error for collation with text
- fix: accommodate both support and collation error
- fix: update tests to compare with new error str

## Related Issue(s)

- #8606 

## Checklist

### DIFF
--- a/go/vt/vtgate/engine/comparer.go
+++ b/go/vt/vtgate/engine/comparer.go
@@ -44,10 +44,10 @@ func (c *comparer) compare(r1, r2 []sqltypes.Value) (int, error) {
 	if err != nil {
 		_, isComparisonErr := err.(evalengine.UnsupportedComparisonError)
 		_, isCollationErr := err.(evalengine.UnsupportedCollationError)
-		if !((isComparisonErr || isCollationErr) && c.weightString != -1) {
+		if !isComparisonErr && !isCollationErr || c.weightString == -1 {
 			return 0, err
 		}
-		// in case of a comparison error switch to using the weight string column for ordering
+		// in case of a comparison or collation error switch to using the weight string column for ordering
 		c.orderBy = c.weightString
 		c.weightString = -1
 		cmp, err = evalengine.NullsafeCompare(r1[c.orderBy], r2[c.orderBy], c.collationID)
--- a/go/vt/vtgate/engine/comparer.go
+++ b/go/vt/vtgate/engine/comparer.go
@@ -43,7 +43,8 @@ func (c *comparer) compare(r1, r2 []sqltypes.Value) (int, error) {
 	cmp, err := evalengine.NullsafeCompare(r1[colIndex], r2[colIndex], c.collationID)
 	if err != nil {
 		_, isComparisonErr := err.(evalengine.UnsupportedComparisonError)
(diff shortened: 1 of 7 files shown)


king-11:
@frouioui I have updated the tests which were already present in the main branch. I will wait for us to merge #9155. After that, I will update the tests that will be added through it. Then we can merge this PR.

frouioui:
Sounds good to me, I'll keep myself posted

go/vt/vtgate/engine/comparer.go
```diff
@@ -43,7 +43,8 @@ func (c *comparer) compare(r1, r2 []sqltypes.Value) (int, error) {
 	cmp, err := evalengine.NullsafeCompare(r1[colIndex], r2[colIndex], c.collationID)
 	if err != nil {
 		_, isComparisonErr := err.(evalengine.UnsupportedComparisonError)
-		if !(isComparisonErr && c.weightString != -1) {
+		_, isCollationErr := err.(evalengine.UnsupportedCollationError)
+		if !((isComparisonErr || isCollationErr) && c.weightString != -1) {
```
frouioui, line 47: This if condition can be simplified to:
```suggestion
		if !isComparisonErr && !isCollationErr && c.weightString == -1) {
```

go/vt/vtgate/engine/ordered_aggregate.go
```diff
@@ -412,7 +412,8 @@ func (oa *OrderedAggregate) keysEqual(row1, row2 []sqltypes.Value, colls map[int
 		cmp, err := evalengine.NullsafeCompare(row1[key.KeyCol], row2[key.KeyCol], colls[key.KeyCol])
 		if err != nil {
 			_, isComparisonErr := err.(evalengine.UnsupportedComparisonError)
-			if !(isComparisonErr && key.WeightStringCol != -1) {
+			_, isCollationErr := err.(evalengine.UnsupportedCollationError)
+			if !((isComparisonErr || isCollationErr) && key.WeightStringCol != -1) {
```
frouioui, line 416: same thing here, the if condition can be simplified

frouioui (approved):
I left two nits, but overall it looks good to me.

king-11:
@frouioui I have made the suggested changes.

systay (approved):

king-11:
@vmg awaiting your check on this PR

vmg (approved):

Issue #9000, opened 2021-10-14
`vtctlclient Reshard Create` fails on timeout, and flag usage is misleading or wrong


ajm188:
#### Overview of the Issue

When creating a reshard workflow with a large amount of initial copying to do, the vtctl command unsurprisingly exceeds the 30s default timeout, leading to:

```
Waiting for workflow to start:
0% ... 0% ... 0% ... 0% ... 0% ... 0% ... 0% ... 0% ... 0% ... 0% ... 0% ... 0% ... 0% ... 0% ... 0% ... 0% ... 0% ... 0% ... 0% ... 0% ... 0% ... 0% ... 0% ... 0% ... 0% ... 0% ... 0% ... 0% ... 0% ... 0% ... E1014 11:57:43.690052   98269 main.go:67] E1014 18:57:43.687240 vtctl.go:2273] workflow did not start within 30s
Reshard Error: rpc error: code = Unknown desc = workflow did not start within 30s
E1014 11:57:43.703769   98269 main.go:76] remote error: rpc error: code = Unknown desc = workflow did not start within 30s
```

However, after the copy finishes, the workflow proceeds to the Running state as if nothing went wrong:

```
❯ vtctlclient -server ":15999" VReplicationExec -json local1-300 "select * from _vt.vreplication" | jq -c '[{key: .fields[11].name, value: .rows[0][11]}] | from_entries'
{"state":"Running"}
```

Looking at the [`timeout` usage](https://github.com/vitessio/vitess/blob/d09e8d10c501395dbb6e312c5235c60532a4f70e/go/vt/vtctl/vtctl.go#L1988) does not clarify things, either:

> Specifies the maximum time to wait, in seconds, for vreplication to catch up on primary migrations. The migration will be cancelled on a timeout. -timeout is only supported for SwitchTraffic and ReverseTraffic.

### Discussion

I'm not sure what the correct thing to do here is. In my view, a stream that is "Copying" has not failed to start, and it feels bad to require the user to know how long the initial copy will take ahead of time to set the `-timeout` appropriately, but as things currently stand I think this is pretty confusing. Unfortunately I don't have a good suggestion for how to make it better.

rohit-nayak-ps:
Fixed by #9206

PR #9206, opened 2021-11-12
VReplication Workflow Create: Check for Copying state while waiting for streams to start

rohit-nayak-ps:
## Description
Workflow Create polls the status of all vreplication streams for _timeout_ seconds (default:30) and reports an error if all streams are deemed not to have started. 

Currently we poll for each stream to reach the Running state. This is incorrect we start with the copy phase which can take a significant amount of time for large data. This PR also checks for the Copying state.

It is expected that, if the workflow is configured correctly, the default 30 seconds will be enough for all streams to find source tablets to start the copy phase.

## Related Issue(s)
#9000 

## Checklist


deepthi (approved):
:+1:

Issue #9195, opened 2021-11-11
Backup fails with gRPC GOAWAY


deepthi:
#### Overview of the Issue
While running a backup with the builtinbackupengine, if there is a large file (200G+ in this case) that has to be uploaded to s3, it can take a while. In the meantime the client (`vtctlclient`) gets a `GOAWAY` from the vttablet that is running the backup.

#### Reproduction Steps
Execute `vtctlclient ... BackupShard ...` on a keyspace/shard on a large-ish database. I haven't figured out a lower bound on how long the backup should be running, but the error shows up within 2 minutes of the last log message received from the tablet.

#### Binary version
Built from commit SHA 9ebc060dc91285c0ec62aafa0d2d9578c36f406c, so basically 12.0

#### Operating system and Environment details
EKS+s3 running `vitess/lite:mysql80` image

#### Workaround
The workaround is to run the backup with the flag value for `grpc_keepalive_time` set to greater than the expected backup time. For example:
```
$ vtctlclient ... -grpc_keepalive_time=2h -action_timeout=2h BackupShard ...
```

#### Log Fragments
```
E1111 02:18:33.644275       1 main.go:76] remote error: rpc error: code = Unavailable desc = closing transport due to: connection error: desc = "error reading from server: EOF", received prior goaway: code: ENHANCE_YOUR_CALM, debug data: too_many_pings
BackupShard Error: rpc error: code = Unavailable desc = closing transport due to: connection error: desc = "error reading from server: EOF", received prior goaway: code: ENHANCE_YOUR_CALM, debug data: too_many_pings
I1111 02:16:52.799998       1 main.go:67] I1111 02:16:52.799517 backup.go:187] I1111 02:16:52.830390 builtinbackupengine.go:376] Backing up file: table_name_redacted.ibd
```

deepthi:
Even though there is a workaround, we should fix this properly.
This seems like a regression from when we updated the gRPC dependency version. The majority of keepalive options are being left at default values in the vitess client/server implementations, but we do set the client side keepalive_time to 10s.

Suggestion from @vmg is to rewrite the gRPC call as a streaming RPC so that the server sends updates back to the client.

Refs:
https://grpc.github.io/grpc/core/md_doc_keepalive.html
https://github.com/grpc/proposal/blob/master/A8-client-side-keepalive.md
https://github.com/vitessio/vitess/blob/main/go/vt/grpcclient/client.go#L42
[...]

PR #9212, opened 2021-11-12
backupengine: send periodic progress updates

vmg:
## Description

Fixes https://github.com/vitessio/vitess/issues/9195

As explained on the linked issue, whenever a very large database is being backed up, the GRPC connection can time out. This is despite the fact that the GRPC request we're performing has been designed to stream back the progress of a backup. The reason why the timeout occurs is because the copying of individual files in a backup is performed via `io.Copy`, which is essentially a busy-loop for copying data, so there is no actual progress report being sent via GRPC.

To work around this, I've implemented a custom `io.Writer` called `backupWriter`, which takes care of both hashing the backup as it is being written, and of keeping tack of how much of the file has been written, so it is capable of sending progress reports via GRPC while the copy phase is happening.

This applies both when performing a backup and when restoring one, so we're killing two birds with one stone in this PR.

## Related Issue(s)
- https://github.com/vitessio/vitess/issues/9195

## Checklist


go/vt/mysqlctl/builtinbackupengine.go
```diff
@@ -602,13 +664,13 @@ func (be *BuiltinBackupEngine) restoreFile(ctx context.Context, params RestorePa
...
 	// Check the hash.
-	hash := hasher.HashString()
+	hash := bw.HashString()
 	if hash != fe.Hash {
 		return vterrors.Errorf(vtrpc.Code_INTERNAL, "hash mismatch for %v, got %v expected %v", fe.Name, hash, fe.Hash)
 	}
 
 	// Flush the buffer.
-	if err := dst.Flush(); err != nil {
+	if err := bw.Close(); err != nil {
```
deepthi, line 673: nit: fix error message below to match the code change.

deepthi (approved):
Looks fine to me, but some of the backup/restore tests are failing.

vmg:
@deepthi Green now :ok_hand:

go/vt/mysqlctl/builtinbackupengine.go
deepthi: Sorry to be nitpicky, but I don't think this was addressed? It's line 540 now.

vmg: It had been addressed when you wrote the comment, I think you may have been looking at the wrong line. And you didn't approve the PR. :sweat:

go/vt/mysqlctl/builtinbackupengine.go
```diff
@@ -446,13 +535,13 @@ func (be *BuiltinBackupEngine) backupFile(ctx context.Context, params BackupPara
 		}
 	}
 
-	// Flush the buffer to finish writing on destination.
-	if err = dst.Flush(); err != nil {
+	// Close the backupPipe to finish writing on destination.
+	if err = bw.Close(); err != nil {
 		return vterrors.Wrapf(err, "cannot flush destination: %v", name)
 	}
```
deepthi, line 541: This is what I was referring to.
The comment says `Close the backupPipe` and the error message says `cannot flush destination`.
I _want_ to approve and merge this PR!
I'll approve provisionally and you can fix the error message and merge it.

deepthi (approved):
Nice fix!

PR #9226, opened 2021-11-14
restore 'vtctl VExec' command, remove Onlind DDL VExec functionality (main branch)

### LINKED ISSUE #9193: Restore/resurrect vtctl VExec command

cc @deepthi @hallaroo

shlomi-noach:
## Description

Fixes #9193 
Backport PR: #9225 

This PR now re-enabled `vtctl VExec`, but then strips out any Online DDL -specific logic from VExec. This means removing the `schema_migrations` table references, planner, templates; demoting `onlineddl.Executor` from being a `VExec`-utor, and removing some boilerplate code.

## Checklist

cc @deepthi @hallaroo

### DIFF
--- a/doc/releasenotes/12_0_0_summary.md
+++ b/doc/releasenotes/12_0_0_summary.md
@@ -38,6 +38,21 @@ If you have reserved connections disabled, you will get the `old` Vitess behavio
 
 ## Deprecations
 
-The command `vtctl VExec` is deprecated and removed. All Online DDL commands should run through `vtctl OnlineDDL`.
+`vtctl VExec` does not support Online DDL queries. All Online DDL commands should be run through `OnlineDDL`.
 
-The command `vtctl OnlineDDL revert` is deprecated. Use `REVERT VITESS_MIGRATION '...'` SQL command either via `vtctl ApplySchema` or via `vtgate`.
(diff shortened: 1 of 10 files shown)


shlomi-noach:
Heh. Not so fast. `vtctl ApplySchema` still relies on `VExec`. Looking into.

shlomi-noach:
OK, further dependency resolved.

shlomi-noach:
Ugh sorry, yet another dependency in `vtctl OnlineDDL ... show ...` command. Resolving, and meanwhile marking as draft.

shlomi-noach:
This gets more complicated. `vtctl OnlineDDL` commands heavily rely on `VExec`. We cannot just delegate commands to the tablets because there is a syntax that is invalid to MySQL, e.g. `SHOW VITESS_MIGRATIONS...`. Normally VTGate takes care of the syntax. On `vtctl` side, we used `VExec` as a parallel path to handle the syntax.

I need to think about this some more.

shlomi-noach:
For now I'm going to close this PR and create a new PR which merely reverts https://github.com/vitessio/vitess/pull/8972, until I figure out the best way to deprecate VExec/OnlineDDL.

PR #9227, opened 2021-11-14
restore 'vtctl VExec' command (backport v12)

### LINKED ISSUE #9193: Restore/resurrect vtctl VExec command

shlomi-noach:
## Description

Fixes #9193 

For now, we restore the full functionality, although we intend to deprecate use of VExec for Online DDL. See some discussion in https://github.com/vitessio/vitess/pull/8972

## Checklist

cc @deepthi @hallaroo @mattlord

### DIFF
--- a/doc/releasenotes/12_0_0_release_notes.md
+++ b/doc/releasenotes/12_0_0_release_notes.md
@@ -31,8 +31,6 @@ zone1-0000000103 sourcekeyspace 0 rdonly 192.168.0.134:15103 192.168.0.134:17103
 
 ## Deprecations
 
-The command `vtctl VExec` is deprecated and removed. All Online DDL commands should run through `vtctl OnlineDDL`.
-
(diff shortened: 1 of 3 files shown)


doc/releasenotes/12_0_0_release_notes.md
```diff
@@ -31,8 +31,6 @@ zone1-0000000103 sourcekeyspace 0 rdonly 192.168.0.134:15103 192.168.0.134:17103
 
 ## Deprecations
 
-The command `vtctl VExec` is deprecated and removed. All Online DDL commands should run through `vtctl OnlineDDL`.
-
```
deepthi, line 35: This should be reverted. See https://github.com/vitessio/vitess/pull/9228#discussion_r750829720

deepthi: Can you start a 12_0_1_release_notes.md and add a line about reverting the deprecation?

shlomi-noach: Done.

deepthi (approved):

PR #9228, opened 2021-11-14
restore 'vtctl VExec' command (main branch)

### LINKED ISSUE #9193: Restore/resurrect vtctl VExec command

shlomi-noach:
## Description

Fixes #9193 
Backport PR: #9227

For now, we restore the full functionality, although we intend to deprecate use of VExec for Online DDL. See some discussion in https://github.com/vitessio/vitess/pull/8972

## Checklist

cc @deepthi @hallaroo @mattlord


doc/releasenotes/12_0_0_release_notes.md
```diff
@@ -31,8 +31,6 @@ zone1-0000000103 sourcekeyspace 0 rdonly 192.168.0.134:15103 192.168.0.134:17103
 
 ## Deprecations
 
-The command `vtctl VExec` is deprecated and removed. All Online DDL commands should run through `vtctl OnlineDDL`.
-
```
deepthi, line 35: These are the release notes that shipped with 12.0.0. Since that is out the door we should not change them.
In 12.0.1 we can write new release notes that we are undo-ing the deprecation.

deepthi (approved):
Looks good.
Can you revert the change to the release notes file and then merge?

doc/releasenotes/12_0_0_release_notes.md
shlomi-noach: Reverted

PR #9232, opened 2021-11-15
Online DDL: evaluate REVERTibility of a migration

shlomi-noach:
## Description

In this WIP PR we evaluate whether a migration is `REVERT`ible. A migration may, for example, possibly be irrevertible if:

- It drops a unique constraint (what happens if a user INSERTs a new row on the new table, which would create a conflict in the old schema?)
- Id expands a data type (e.g. `INT` to `BIGINT`; what happens if a user INSERTs a value greater than `INT` on the new table? This value cannot be propagated back to the old table)
- dropping a `NOT NULL` column with no `DEFAULT` value, then INSERTing any row on the new table. How can we populate the dropped column on the old table?
- other?

We cannot say for certain that a migration is _not_ revertible. It depends on how the user uses the new table in the interim.
However, assuming we cover all the cases, we can _predict_ that a migration _is in fact_ revertible. If we see nothing to block revertibility, then we have high confidence that we will make it.

As a side note, most cases illustrated above are IMO _unlikely_. It's _unlikely_ that immediately after cut-over the user will suddenly populate a BIGINT value: their app needs to first realize it _can_ populate such values. But of course we can't say for sure. We cannot predict an app's behavior in the generic case.

## Related Issue(s)

Followup to #8495 
Tracking: #6926

## Checklist


shlomi-noach:
Examples of expanded data types, which we need to check for revertibility:

- [ ] Examples of expanded data types:
- [ ] Any NOT NULL to NULLable
- [ ] varchar(10)->varchar(20)
- [ ] expansion/change of character set (e.g. going utf8 to utf8mb4)
- [ ] Any increase in INT type (eg INT->BIGINT)
- [ ] Any unsigned to signed integer (because -1 is an overflow)
- [ ] Some signed to unsigned, where new datatype equals or greater than previous datatype
- [ ] E.g. INT SIGNED to INT UNSIGNED, INT SIGNED to BIGINT UNSIGNED
- [ ] But no problem with BIGINT SIGNED to INT UNSIGNED
- [ ] Increase in BIT length
- [ ] FLOAT to DOUBLE
- [ ] Integer to FLOAT/DOUBLE
- [ ] Increase in DECIMAL length
- [ ] DATE->DATETIME, TIME->DATETIME
- [ ] DATE->TIMESTAMP, TIME->TIMESTAMP
- [ ] TIMESTAMP->DATE (because DATE  can be before 1970-01-01)
- [ ] TIMESTAMP->DATETIME
- [ ] Temporal without fraction to temporal with fraction (e.g. TIMESTAMP to TIMESTAMP(3), TIMESTAMP(3) to TIMESTAMP(6))
- [ ] Adding new ENUM/SET value

shlomi-noach:
It occurs to me that a user may change a column's type from anything to anything, really. e.g. `INT` to `VARCHAR`, `CHAR` to `DATETIME` etc. We can't (won't?) review all possible transformations, just the most sensible ones (between numerics, between texts, etc.)

go/vt/vttablet/onlineddl/vrepl/unique_key.go
```diff
@@ -0,0 +1,161 @@
+/*
```
shlomi-noach, line 1: This was refactored out of `vrepl.go` without code changes

go/vt/vttablet/onlineddl/vrepl/columns.go
```diff
@@ -0,0 +1,192 @@
+/*
```
shlomi-noach, line 1: This file was refactored out of `vrepl.go` with changes as noted below.

go/vt/vttablet/onlineddl/vrepl/columns.go
```diff
@@ -0,0 +1,192 @@
...
+	"time:datetime":      true,
+	"date:datetime":      true,
+	"timestamp:datetime": true,
+	"time:timestamp":     true,
+	"date:timestamp":     true,
+	"timestamp:date":     true,
+}
+
+// GetSharedColumns returns the intersection of two lists of columns in same order as the first list
+func GetSharedColumns(
```
shlomi-noach, line 28: This function is unchanged, refactored out of `vrepl.go`

go/vt/vttablet/onlineddl/vrepl/columns.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+	mappedSharedColumnNames := []string{}
+	for _, columnName := range sharedColumnNames {
+		mappedSharedColumnNames = append(mappedSharedColumnNames, sharedColumnsMap[columnName])
+	}
+	return NewColumnList(sharedColumnNames), NewColumnList(mappedSharedColumnNames), NewColumnList(droppedSourceNonGeneratedColumnsNames), sharedColumnsMap
+}
+
+// isExpandedColumn sees if target column has any value set/range that is impossible in source column. See GetExpandedColumns comment for examples
+func isExpandedColumn(sourceColumn *Column, targetColumn *Column) (bool, string) {
```
shlomi-noach, line 96: New functionality, and at the heart of this PR

go/vt/vttablet/onlineddl/vrepl/columns.go
```diff
@@ -0,0 +1,192 @@
...
+}
+
+// GetExpandedColumnNames is given source and target shared columns, and returns the list of columns whose data type is expanded.
+// An expanded data type is one where the target can have a value which the source does not. Examples:
+// - any NOT NULL to NULLable (a NULL in the target cannot appear on source)
+// - INT -> BIGINT (obvious)
+// - BIGINT UNSIGNED -> INT SIGNED (negative values)
+// - TIMESTAMP -> TIMESTAMP(3)
+// etc.
+func GetExpandedColumnNames(
```
shlomi-noach, line 162: new functionality

go/vt/vttablet/onlineddl/vrepl/columns.go
```diff
@@ -0,0 +1,192 @@
...
+		if isExpanded, description := isExpandedColumn(&sourceColumn, &targetColumn); isExpanded {
+			expandedColumnNames = append(expandedColumnNames, sourceColumn.Name)
+			expandedDescriptions[sourceColumn.Name] = description
+		}
+	}
+	return expandedColumnNames, expandedDescriptions
+}
+
+// GetNoDefaultColumnNames returns names of columns which have no default value, out of given list of columns
+func GetNoDefaultColumnNames(columns *ColumnList) (names []string) {
```
shlomi-noach, line 184: new functionality. Though small, it's at the heart of this PR

go/vt/vttablet/onlineddl/schema.go
```diff
@@ -558,4 +565,8 @@ var ApplyDDL = []string{
 	alterSchemaMigrationsTableLogFile,
 	alterSchemaMigrationsTableRetainArtifacts,
 	alterSchemaMigrationsTablePostponeCompletion,
+	alterSchemaMigrationsTableRemovedUniqueNames,
+	alterSchemaMigrationsTableRemovedNoDefaultColNames,
+	alterSchemaMigrationsTableExpandedColNames,
+	alterSchemaMigrationsTableRevertibleNotes,
```
shlomi-noach, line 571: These new column in `schema_migrations` table help us to understand better the nature of the schema change, and in particular inform us what may be blockers/issues to reverting the migration.

go/vt/vttablet/onlineddl/schema.go
```diff
@@ -45,24 +45,28 @@ const (
...
+	alterSchemaMigrationsTableRemovedUniqueKeys        = "ALTER TABLE _vt.schema_migrations add column removed_unique_keys int unsigned NOT NULL DEFAULT 0"
+	alterSchemaMigrationsTableLogFile                  = "ALTER TABLE _vt.schema_migrations add column log_file varchar(1024) NOT NULL DEFAULT ''"
+	alterSchemaMigrationsTableRetainArtifacts          = "ALTER TABLE _vt.schema_migrations add column retain_artifacts_seconds bigint NOT NULL DEFAULT 0"
+	alterSchemaMigrationsTablePostponeCompletion       = "ALTER TABLE _vt.schema_migrations add column postpone_completion tinyint unsigned NOT NULL DEFAULT 0"
+	alterSchemaMigrationsTableRemovedUniqueNames       = "ALTER TABLE _vt.schema_migrations add column removed_unique_key_names text NOT NULL"
+	alterSchemaMigrationsTableRemovedNoDefaultColNames = "ALTER TABLE _vt.schema_migrations add column dropped_no_default_column_names text NOT NULL"
+	alterSchemaMigrationsTableExpandedColNames         = "ALTER TABLE _vt.schema_migrations add column expanded_column_names text NOT NULL"
+	alterSchemaMigrationsTableRevertibleNotes          = "ALTER TABLE _vt.schema_migrations add column revertible_notes text NOT NULL"
```
shlomi-noach, line 69: These new columns in `schema_migrations` table help us to understand better the nature of the schema change, and in particular inform us what may be blockers/issues to reverting the migration.

shlomi-noach:
Ready for review!

Notes notes for the reviewers:

- Added endtoend tests in `go/test/endtoend/onlineddl/revertible/onlineddl_revertible_test.go`. 
  This is a suite of tests. Each test has two table definitions: "from" and "to", and runs an actual migrations between the two table. We then validate the analysis of the change: which unique constraints were dropped, which columns dropped which had no default value, which column changed type into a borader scope.

- `go/vt/vttablet/onlineddl/vrepl/columns.go` refactored out side of `vrepl.go`, and then also added the functions `isExpandedColumn`, `GetExpandedColumnNames`
- `go/vt/vttablet/onlineddl/vrepl/unique_key.go` refactored out side of `vrepl.go` without code changes
- likewise, tests were refactored outside `vrepl_test.go`.

go/vt/schema/parser.go
```diff
@@ -54,7 +54,7 @@ var (
 	createTableRegexp     = regexp.MustCompile(`(?s)(?i)(CREATE\s+TABLE\s+)` + "`" + `([^` + "`" + `]+)` + "`" + `(\s*[(].*$)`)
 	revertStatementRegexp = regexp.MustCompile(`(?i)^revert\s+([\S]*)$`)
 
-	enumValuesRegexp = regexp.MustCompile("(?i)^enum[(](.*)[)]$")
+	enumValuesRegexp = regexp.MustCompile("(?i)^(enum|set)[(](.*)[)]$")
```
deepthi, line 57: Does this change affect anything besides online DDL?

go/vt/schema/parser.go
```diff
@@ -109,7 +109,7 @@ func legacyParseRevertUUID(sql string) (uuid string, err error) {
 // ParseEnumValues parses the comma delimited part of an enum column definition
 func ParseEnumValues(enumColumnType string) string {
 	if submatch := enumValuesRegexp.FindStringSubmatch(enumColumnType); len(submatch) > 0 {
-		return submatch[1]
+		return submatch[2]
```
deepthi, line 112: Is this fixing a bug?

go/vt/vttablet/onlineddl/executor.go
```diff
@@ -890,7 +890,18 @@ func (e *Executor) ExecuteWithVReplication(ctx context.Context, onlineDDL *schem
...
 	}
-	if err := e.updateMigrationAddedRemovedUniqueKeys(ctx, onlineDDL.UUID, len(v.addedUniqueKeys), len(v.removedUniqueKeys)); err != nil {
+	removedUniqueKeyNames := []string{}
+	for _, uniqueKey := range v.removedUniqueKeys {
+		removedUniqueKeyNames = append(removedUniqueKeyNames, uniqueKey.Name)
+	}
+	if err := e.updateSchemaAnalysis(ctx, onlineDDL.UUID,
+		len(v.addedUniqueKeys),
+		len(v.removedUniqueKeys),
+		strings.Join(removedUniqueKeyNames, ","),
```
deepthi, line 900: Is it safe to concatenate like this? MySQL lets you use commas in column names.

go/vt/vttablet/onlineddl/vrepl/columns.go
```diff
@@ -0,0 +1,192 @@
+/*
+   Copyright 2016 GitHub Inc.
+	 See https://github.com/github/gh-ost/blob/master/LICENSE
+*/
+
```
deepthi, line 5: Has this been copied from `gh-ost` with no changes?

go/vt/vttablet/onlineddl/vrepl/unique_key.go
```diff
@@ -0,0 +1,161 @@
...
+		return false
+	}
+	if uniqueKey.HasFloat {
+		// float & double data types are imprecise and we cannot use them while iterating unique keys
+		return false
+	}
+	return true // good to go!
+}
+
+// GetSharedUniqueKeys returns the unique keys shared between the two source&target tables
```
deepthi, line 27: nit: it seems redundant to say "two source&target". Just "source and target"?

deepthi (commented):
Very nicely written PR (as usual). The online DDL stuff blows my mind.
A few questions/clarification requests inline.

go/vt/schema/parser.go
shlomi-noach: This isn't fixing a bu. This is because the regex changed from `enum` to `(enum|set)`, which adds a new `(...)` submatch to the pattern. There are unit tests to validate the correctness of the submatch in this function.

go/vt/vttablet/onlineddl/executor.go
shlomi-noach: You are right -- but this is going to be informational only. That is, we're not going to then re-parse this. This is about us being able to tell the user "there is a chance we can't revert this migration due to these columns: a,b,c". So in a way, I consider this to be free text.
However, you're right and why no make the extra effort of qualifying the names. I'll proceed to do that.

go/vt/vttablet/onlineddl/vrepl/columns.go
shlomi-noach: Right! It was copied from `gh-ost` and there have been changes. I'll append the Vitess copyright.

go/vt/schema/parser.go
shlomi-noach: Double checking.

shlomi-noach:
Addressed all review issues:

- Copyright notices updated
- Separate parsing for `enum` and for `set` data types
- Qualifying column names in schema analysis columns
- Typos & grammar

deepthi (approved):
👍

Issue #9222, opened 2021-11-13
Task: delete automationservice and related code


ajm188:
The `go/vt/automation` code (and related protobufs) was used to run automated tasks as part of integration tests and is not needed any longer.

- [ ] Delete AutomationServer and `cmd` entrypoint
- [ ] Delete AutomationClient and `cmd` entrypoint
- [ ] Delete `go/vt/automation` and anything now-unused related to it
    - This is dependent on #9219, as all the legacy resharding Tasks (SplitClone, VerticalSplitClone, and so on) live here
- [ ] Delete `automation.proto`
- [ ] Remove golangci rules for `go/vt/automation`

Here is my git grep:

```
.golangci.yml:    - path: '^go/vt/automation/'
go/cmd/automation_server/automation_server.go:  "vitess.io/vitess/go/vt/automation"
go/vt/workflow/resharding/tasks.go:     "vitess.io/vitess/go/vt/automation"
```

PR #9234, opened 2021-11-15
Delete `go/cmd/automation_{client,server}`

ajm188:

## Description

Just a `git rm -rf ./go/cmd/automation_{client,server}`. I've never actually deleted a binary from vitess so I'm not sure if this is the right process (i.e. should we "deprecate" the binary by having it print a warning whenever you try to run it?)

## Related Issue(s)

#9222 

## Checklist


deepthi:
> I've never actually deleted a binary from vitess so I'm not sure if this is the right process (i.e. should we "deprecate" the binary by having it print a warning whenever you try to run it?)

In this case I think we are sure these binaries are defunct.
For something like `vtctlclient` we will go through a proper deprecation cycle.

Why not delete the `go/vt/automation` package along with `cmd/automation_*`?

ajm188:
> Why not delete the `go/vt/automation` package along with `cmd/automation_*`?

It's used in a few other places, namely vtworker (which is also being deleted but will be a much larger diff):

```
❯ go list -f '{{ .ImportPath }}:{{ .Deps }}' ./... | grep automation | cut -d ':' -f1
vitess.io/vitess/go/cmd/automation_client
vitess.io/vitess/go/cmd/automation_server
vitess.io/vitess/go/cmd/vtctld
vitess.io/vitess/go/vt/automation
vitess.io/vitess/go/vt/proto/automation
vitess.io/vitess/go/vt/proto/automationservice
vitess.io/vitess/go/vt/vtctld
vitess.io/vitess/go/vt/workflow/resharding
vitess.io/vitess/go/vt/workflow/reshardingworkflowgen
```

deepthi (approved):

PR #9235, opened 2021-11-15
Ensure that vtgate's two tablet caches are in sync

mattlord:

## Checklist


mattlord:
Going with a different approach after going through the problem space in depth with @sougou and @deepthi

PR #9237, opened 2021-11-15
Prevent Race Conditions Between Tablet Deletes and Updates

mattlord:
## Description
There was a race condition between [deleting a tablet's healthcheck record](https://github.com/vitessio/vitess/blob/693c5dbdeacdd7a705b46ebce6776a5256c8cfef/go/vt/discovery/healthcheck.go#L379-L408) from the authoritative map ([`hc.healthByAlias`](https://github.com/vitessio/vitess/blob/693c5dbdeacdd7a705b46ebce6776a5256c8cfef/go/vt/discovery/healthcheck.go#L227)) and [updating the same tablet's health data record](https://github.com/vitessio/vitess/blob/693c5dbdeacdd7a705b46ebce6776a5256c8cfef/go/vt/discovery/healthcheck.go#L410-L485) ([`hc.healthData`](https://github.com/vitessio/vitess/blob/693c5dbdeacdd7a705b46ebce6776a5256c8cfef/go/vt/discovery/healthcheck.go#L227)).

This could cause us to effectively re-add an updated copy of the tablet healthcheck record after it's been deleted [due to the open stream with the tablet encounting an error](https://github.com/vitessio/vitess/blob/693c5dbdeacdd7a705b46ebce6776a5256c8cfef/go/vt/discovery/tablet_health_check.go#L275-L304) -- often a context being cancelled or a gRPC goaway error seen due to the tablet shutting down. This then leads to "zombie" tablet records in the `SHOW VITESS_TABLETS` output as it is based on what is in the `hc.healthData` map and it had records for tablets that were no longer in `hc.healthByAlias` (so the authoritative map was no longer authoritative):
https://github.com/vitessio/vitess/blob/693c5dbdeacdd7a705b46ebce6776a5256c8cfef/go/vt/vtgate/executor.go#L1108
...
https://github.com/vitessio/vitess/blob/693c5dbdeacdd7a705b46ebce6776a5256c8cfef/go/vt/discovery/healthcheck.go#L537-L557

## Changes
 - **_KEY:_** The healthcheck will **_ONLY_** delete from its own lower level cache directly -- via `hc.deleteTablet()` -- if there's mismatching endpoints for the tablet indicating that the healthcheck record is no longer valid (this was the case before https://github.com/vitessio/vitess/pull/9106) 
   - This is the **_only_** safe exception to the rule that the healthcheck cache be updated (add,replace,delete) via topo server -> topology watcher -> healthcheck [here](https://github.com/vitessio/vitess/blob/693c5dbdeacdd7a705b46ebce6776a5256c8cfef/go/vt/discovery/topology_watcher.go#L217-L240), as in this case the record is simply not valid and will get corrected in the noted path via the topo server tablet records in the next refresh
 - **_KEY:_** We avoid the race condition between`hc.deleteTablet()` and `hc.updateHealth()` by confirming that the tablet record still exists in the authoritative `hc.healthByAlias` map before updating the health of the tablet in the sibling `hc.healthData` map because the mutex does not enforce order of operations. If it does NOT exist there then the update is a no-op (the old tablet health data record copy made [here](https://github.com/vitessio/vitess/blob/693c5dbdeacdd7a705b46ebce6776a5256c8cfef/go/vt/discovery/tablet_health_check.go#L303) or [here](https://github.com/vitessio/vitess/blob/693c5dbdeacdd7a705b46ebce6776a5256c8cfef/go/vt/discovery/tablet_health_check.go#L314) will go out of scope and be GC'd).
 - The `vtgate/tablet_healthcheck_cache/correctness` end2end test was updated to verify the correct behavior (unfortunately the test originally added in https://github.com/vitessio/vitess/pull/9106 was simply verifying bad behavior)
 - We create the record in the sibling `hc.healthData` map -- if it doesn't exist -- when updating the health record for a tablet that we have confirmed exists in the authoritative `hc.healthByAlias` map 

With these changes, I could no longer repeat the zombie tablet record using [the steps shown here](https://github.com/vitessio/vitess/issues/8465#issuecomment-951433088), and we see the expected log messages indicating that we've avoided the race condition:
```
/vt/vtdataroot/tmp/vtgate.INFO:I1115 23:31:18.311915    2021 healthcheck.go:424] Tablet zone1-0000000201 has been deleted, skipping health update
/vt/vtdataroot/tmp/vtgate.INFO:I1115 23:31:18.311976    2021 healthcheck.go:424] Tablet zone1-0000000200 has been deleted, skipping health update
/vt/vtdataroot/tmp/vtgate.INFO:I1115 23:31:18.311994    2021 healthcheck.go:424] Tablet zone1-0000000202 has been deleted, skipping health update
```

## Related Issue(s)

Fixes: https://github.com/vitessio/vitess/issues/9238
**_Properly_** Fixes: https://github.com/vitessio/vitess/issues/8465

## Checklist
- [x] Tests were added/updated

### DIFF
--- a/go/vt/discovery/healthcheck.go
+++ b/go/vt/discovery/healthcheck.go
@@ -410,7 +410,7 @@ func (hc *HealthCheckImpl) updateHealth(th *TabletHealth, prevTarget *query.Targ
 	defer hc.mu.Unlock()
 
 	tabletAlias := tabletAliasString(topoproto.TabletAliasString(th.Tablet.Alias))
-	// let's be sure that this tablet hasn't been deleted from the authortative map
+	// let's be sure that this tablet hasn't been deleted from the authoritative map
 	// so that we're not racing to update it and in effect re-adding a copy of the
 	// tablet record that was deleted
 	if _, ok := hc.healthByAlias[tabletAlias]; !ok {
--- a/go/vt/discovery/healthcheck.go
+++ b/go/vt/discovery/healthcheck.go
@@ -351,13 +351,10 @@ func (hc *HealthCheckImpl) AddTablet(tablet *topodata.Tablet) {
 	}
 	hc.healthByAlias[tabletAliasString(tabletAlias)] = thc
 	res := thc.SimpleCopy()
-	if ths, ok := hc.healthData[key]; !ok {
+	if _, ok := hc.healthData[key]; !ok {
 		hc.healthData[key] = make(map[tabletAliasString]*TabletHealth)
-		hc.healthData[key][tabletAliasString(tabletAlias)] = res
-	} else {
-		// just overwrite it if it exists already
-		ths[tabletAliasString(tabletAlias)] = res
 	}
+	hc.healthData[key][tabletAliasString(tabletAlias)] = res
 
 	hc.broadcast(res)
 	hc.connsWG.Add(1)
@@ -381,6 +378,7 @@ func (hc *HealthCheckImpl) deleteTablet(tablet *topodata.Tablet) {
 	hc.mu.Lock()
 	defer hc.mu.Unlock()
 
+	key := hc.keyFromTablet(tablet)
 	tabletAlias := tabletAliasString(topoproto.TabletAliasString(tablet.Alias))
 	// delete from authoritative map
 	th, ok := hc.healthByAlias[tabletAlias]
@@ -392,18 +390,14 @@ func (hc *HealthCheckImpl) deleteTablet(tablet *topodata.Tablet) {
 	// which will call finalizeConn, which will close the connection.
 	th.cancelFunc()
 	delete(hc.healthByAlias, tabletAlias)
-
-	// the tablet has been deleted from the authoritative healthByAlias map so let's ensure it's deleted
-	// from the healthData map as well, which means we need to delete any existing combinations of
-	// keyspace.shard.tabletType we may have had for the tablet
-	for _, key := range hc.keysFromTablet(tablet) {
-		if ths, ok := hc.healthData[key]; ok {
-			delete(ths, tabletAlias)
-		}
+	// delete from map by keyspace.shard.tabletType
+	ths, ok := hc.healthData[key]
+	if !ok {
+		log.Warningf("We have no health data for target: %v", key)
+		return
 	}
-
-	// We only need to recompute the healthy record for the current state of the tablet
-	key := hc.keyFromTablet(tablet)
+	delete(ths, tabletAlias)
+	// delete from healthy list
 	healthy, ok := hc.healthy[key]
 	if ok && len(healthy) > 0 {
 		hc.recomputeHealthy(key)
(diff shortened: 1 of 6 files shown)


go/vt/discovery/healthcheck.go
```diff
@@ -413,6 +416,15 @@ func (hc *HealthCheckImpl) updateHealth(th *TabletHealth, prevTarget *query.Targ
 	defer hc.mu.Unlock()
 
 	tabletAlias := tabletAliasString(topoproto.TabletAliasString(th.Tablet.Alias))
+	// let's be sure that this tablet hasn't been deleted from the authortative map
+	// so that we're not racing to update it and in effect re-adding a copy of the
+	// tablet record that was deleted
+	_, ok := hc.healthByAlias[tabletAlias]
+	if !ok {
+		log.Infof("Tablet %s has been deleted, skipping health update", tabletAlias)
```
deepthi, line 424: We should always print the full tablet info instead of just the alias. Makes debugging easier.

go/vt/discovery/tablet_health_check.go
```diff
@@ -290,11 +290,9 @@ func (thc *tabletHealthCheck) checkConn(hc *HealthCheckImpl) {
 		if err != nil {
 			hcErrorCounters.Add([]string{thc.Target.Keyspace, thc.Target.Shard, topoproto.TabletTypeLString(thc.Target.TabletType)}, 1)
 			// We have reason to suspect the tablet healthcheck record is corrupted or invalid so let's remove the tablet's record
-			// from the healthcheck cache and it will get re-added again if the tablet is reachable
-			if strings.Contains(err.Error(), "health stats mismatch") ||
-				strings.HasSuffix(err.Error(), context.Canceled.Error()) ||
-				strings.Contains(err.Error(), `"error reading from server: EOF", received prior goaway`) {
-				log.Warningf("tablet %s had a suspect healthcheck error: %s -- clearing cache record", thc.Tablet.Alias, err.Error())
+			// from the healthcheck cache and the corrected tablet record will be fetched from the topology and added to the
+			// healthcheck cache again via the topology watcher.
```
deepthi, line 294: Let us update these comments.
```
This means that another tablet has taken over the host:port that we were connected to. So let's remove the tablet's data from the healthcheck, and if it still a part of the cluster, the new tablet record will be fetched from the topology and added to the healthcheck cache again via the topology watcher.
WARNING: Under no other circumstances should we be deleting the tablet here.
```

go/vt/discovery/tablet_health_check.go
```diff
@@ -290,11 +290,9 @@ func (thc *tabletHealthCheck) checkConn(hc *HealthCheckImpl) {
...
+			if strings.Contains(err.Error(), "health stats mismatch") {
 				hc.deleteTablet(thc.Tablet)
```
deepthi, line 296: Let's add a log line before this `deleteTablet`

go/vt/discovery/healthcheck.go
```diff
@@ -413,6 +410,14 @@ func (hc *HealthCheckImpl) updateHealth(th *TabletHealth, prevTarget *query.Targ
 	defer hc.mu.Unlock()
 
 	tabletAlias := tabletAliasString(topoproto.TabletAliasString(th.Tablet.Alias))
+	// let's be sure that this tablet hasn't been deleted from the authortative map
```
deepthi, line 413:
```suggestion
	// let's be sure that this tablet hasn't been deleted from the authoritative map
```

deepthi (approved):
Found a tiny typo in comments, otherwise LGTM.
Nice work!

go/vt/discovery/healthcheck.go
mattlord: Nice catch! :-) Corrected

sougou (approved):

PR #9239, opened 2021-11-15
Migrate jQuery (pt. 1)

notfelineit:
## Description
This PR addresses an issue on the private vitess fork to upgrade jQuery to the latest version.

It completely removes jQuery dependencies from the vtgate and vttablet web UIs, as well as updates the plotly.js snippet in vtctld.

### Surface Area
Orchestrator still needs to be looked at, but these UIs have been identified via [this list.](https://vitess.io/docs/user-guides/configuration-basic/ports/)
Screen Shot 2021-11-15 at 4 40 55 PM

#### vtgate
It looks like jQuery in the vtgate UI was primarily used to mount Google's chart library. The chart library has been updated to use modern import and jQuery was swapped out here for a simple `document.getElementById` and the browser native `fetch` api. Confirmed that the charts still render and there is no jQuery anymore:
Screen Shot 2021-11-15 at 4 05 12 PM

#### vttablet
vttablet uses jQuery for the same reasons as vtgate, but in two places. Both have been updated to use `document.getElementById` and the `fetch` api. Confirmed that the charts still render and there is no jQuery:
Screen Shot 2021-11-15 at 5 10 54 PM

#### vtctld
vtctld does not actually import jQuery and it is not loaded on the vtctld dashboard. The only place that references jQuery is the old plotly.js snippet. I updated this snippet and confirmed the graph still looks the same:
Screen Shot 2021-11-12 at 9 55 10 AM

## Related Issue(s)
N/A

## Checklist

## Deployment Notes
No extra deployment steps necessary for these changes. They do not add any new libraries.


doeg:
Ah YAY great work (and great write-up)! 🎉 Is this ready for review? If so, I can look first thing tomorrow.

notfelineit:
@doeg Yes! Gotta get the tests to past though 😅

.gitignore
```diff
@@ -73,6 +73,8 @@ releases
 /web/vtctld2/bower.json~
 /web/vtctld2/public/bower_components/
 
+# Local examples
+/examples/local/vtdataroot
```
doeg, line 77: Thanks for adding this!

doeg:
@notfelineit I can re-run the tests for you this morning. :') The failures are definitely unrelated to your PR. 

The only test that needs attention is the one that checks for the release note PR labels [here](https://github.com/vitessio/vitess/runs/4219459248?check_suite_focus=true). (This seems like a good one to mention in the release notes, fwiw!)

doeg (approved):
YAYyy the best jQuery upgrade is one that removes jQuery entirely! 😈 

Such great work! Updating these dusty UIs is no easy feat. :) I pulled your branch + ran it locally and everything looks good to me! Thanks again for such a thorough write-up with screenshots.

.gitignore
notfelineit: No problemo! :-)

deepthi (approved):
@doeg's review is the one that matters, but :+1: 
We can discuss which release(s) this actually needs to be backported to.

PR #9240, opened 2021-11-15
Set super_read_only off during restore;  achieves same as PR #8929

aquarapid:
In short, we are blindly setting `super_read_only` off in the restore flow, in case the default my.cnf params has `super_read_only=ON`.  This allows us to populate our metadata tables without running into write issues.

Tested manually and by the current DB restore e2e flow.

### DIFF
--- a/go/vt/mysqlctl/backup.go
+++ b/go/vt/mysqlctl/backup.go
@@ -334,6 +334,16 @@ func Restore(ctx context.Context, params RestoreParams) (*BackupManifest, error)
 		return nil, err
 	}
 
+	// We disable super_read_only, in case it is in the default MySQL startup
+	// parameters and will be blocking the writes we need to do in
+	// PopulateMetadataTables().  We do it blindly, since
+	// this will fail on MariaDB, which doesn't have super_read_only
+	// This is safe, since we're restarting MySQL after the restore anyway
+	params.Logger.Infof("Restore: disabling super_read_only")
+	if err := params.Mysqld.SetSuperReadOnly(false); err != nil {
+		params.Logger.Warningf("Restore: setting super_read_only failed, continuing anyway: %v", err)
+	}
+
 	params.Logger.Infof("Restore: running mysql_upgrade")
 	if err := params.Mysqld.RunMysqlUpgrade(); err != nil {
 		return nil, vterrors.Wrap(err, "mysql_upgrade failed")
--- a/go/vt/mysqlctl/backup.go
+++ b/go/vt/mysqlctl/backup.go
@@ -58,6 +58,8 @@ const (
 const (
 	// replicationStartDeadline is the deadline for starting replication
 	replicationStartDeadline = 30
+
+	Error1193 = "Unknown system variable"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


go/vt/mysqlctl/backup.go
```diff
@@ -334,6 +334,16 @@ func Restore(ctx context.Context, params RestoreParams) (*BackupManifest, error)
 		return nil, err
 	}
 
+	// We disable super_read_only, in case it is in the default MySQL startup
+	// parameters and will be blocking the writes we need to do in
+	// PopulateMetadataTables().  We do it blindly, since
+	// this will fail on MariaDB, which doesn't have super_read_only
+	// This is safe, since we're restarting MySQL after the restore anyway
+	params.Logger.Infof("Restore: disabling super_read_only")
+	if err := params.Mysqld.SetSuperReadOnly(false); err != nil {
```
deepthi, line 343: It will be better to check for a specific error to ignore for mariadb (and log warning and move on), and fail the operation if there is some other unexpected error.

aquarapid: One of the problems here is that it is not trivial to know if it is MariaDB or not, without writing a whole set of version parsing code;  I was initially going to do that and then just try to set super_read_only for non-mariadb.  I'll do a best-effort error without having to dig that out.

go/vt/mysqlctl/backup.go
```diff
@@ -58,6 +58,8 @@ const (
 const (
 	// replicationStartDeadline is the deadline for starting replication
 	replicationStartDeadline = 30
+
+	Error1193 = "Unknown system variable"
```
deepthi, line 62: There is a constant defined for this error code (not string). Can you use that instead, or should we worry about false positive hits?
It's defined here https://github.com/vitessio/vitess/blob/main/go/mysql/constants.go#L495

go/vt/mysqlctl/backup.go
```diff
@@ -334,6 +336,21 @@ func Restore(ctx context.Context, params RestoreParams) (*BackupManifest, error)
...
 	}
 
+	// We disable super_read_only, in case it is in the default MySQL startup
+	// parameters and will be blocking the writes we need to do in
+	// PopulateMetadataTables().  We do it blindly, since
+	// this will fail on MariaDB, which doesn't have super_read_only
+	// This is safe, since we're restarting MySQL after the restore anyway
+	params.Logger.Infof("Restore: disabling super_read_only")
+	if err := params.Mysqld.SetSuperReadOnly(false); err != nil {
+		if strings.Contains(err.Error(), Error1193) {
```
deepthi, line 346: @systay I think we have a bigger problem here..
Ideally we would like people to be able to do something like this
```
	errCode := vterrors.Code(err)
	sqlErr, ok := err.(*mysql.SQLError)
```
However, the funcs in the `mysqlctl` package were never migrated to `vterrors`, which means you can't use any of the good stuff you get from that package.
I won't hold up this PR, but something for us to tackle as a cleanup/tech debt task.

go/vt/mysqlctl/backup.go
aquarapid: I've migrated it to SQLError (as you have below) to align.

aquarapid:
Ah nevermind;  converting to SQLError won't work as you noted.  I'll leave this as-is for now.

systay (approved):

go/vt/mysqlctl/backup.go
deepthi: Created #9311

PR #9242, opened 2021-11-16
Support EXISTS subquery with conditional

### LINKED ISSUE #9241: EXISTS subquery with a conditional not supported
Scenario:
  * Unsharded keyspace
  * Recent `main` branch (`Version: 13.0.0-SNAPSHOT (Git revision d9fd31e3ae branch 'main') built on Mon Nov 15 17:33:56 PST 2021 by jacques@dhoomdeskw.localdomain using go1.17.1 linux/amd64`)
  * The following `EXISTS` subqueries work without issue:
  ```
mysql> create table t1 (c1 bigint not null, c2 bigint not null, c3 bigint, primary key(c1));                     
Query OK, 0 rows affected (0.03 sec)

mysql> insert into t1 (c1,c2,c3) values (2, 1, 1);                                                           
Query OK, 1 row affected (0.00 sec)

mysql> select * from t1 where EXISTS (select * from t1 where c1 = 9);                          
Empty set (0.01 sec)

mysql> select * from t1 where EXISTS (select * from t1 where c1 = 2);
+----+----+------+
| c1 | c2 | c3   |
+----+----+------+
|  2 |  1 |    1 |
+----+----+------+
1 row in set (0.00 sec)
  ```
  * However, the equivalent `EXISTS` subquery with a conditional gives a syntax error:
  ```
mysql> select * from t1 where EXISTS (select * from t1 where c1 = 2) = TRUE;
ERROR 1105 (HY000): syntax error at position 64
  ```
  * Meanwhile, when running the last case directly against the underlying MySQL, this works fine:
  ```
mysql> select * from t1 where EXISTS (select * from t1 where c1 = 2) = TRUE;
+----+----+------+
| c1 | c2 | c3   |
+----+----+------+
|  2 |  1 |    1 |
+----+----+------+
1 row in set (0.00 sec)
  ```

We should support this use-case, some ORMs seem to generate this (e.g. older versions of Django).

aquarapid:
Fixes #9241

Not 100% sure what/where additional tests should be added.


systay:
I think @GuptaManan100 and @ritwizsinha had a plan for how to change the parser so these queries would be parsed correctly. Adding the comparison this way is a little hacky, and I think we should avoid it unless we have to get it in now-now-now

ritwizsinha:
I tested these queries on #9075 and it works though not quite matching the output given in these tests

GuptaManan100:
Closing this in favour of #9075 which has added support for the parsing of exists expressions and have booleans after it

ritwizsinha:
Should I add the tests added here ?

GuptaManan100:
I already did to that PR

PR #9243, opened 2021-11-16
Gen4: allow group by queries with sharding keys to pass through when it is a single route

### LINKED ISSUE #9221: Query Planner ERROR 1105 (HY000): symbol %s not found when running SELECT FROM keyspacename.tablename
#### Overview of the Issue

Running the following query - 
```sql
SELECT user.* FROM user.user INNER JOIN user_extra ue ON ue.user_id = user.id GROUP BY user.id
```
results in the following plans - 
V3 
```json
"symbol `user`.id not found"
```
Gen4
```json
Expression of SELECT list is not in GROUP BY clause and contains nonaggregated column '`user`.*' which is not functionally dependent on columns in GROUP BY clause; this is incompatible with sql_mode=only_full_group_by
```

And if we remove the user keyspace from the query to - 
```sql
SELECT user.* FROM user INNER JOIN user_extra ue ON ue.user_id = user.id GROUP BY user.id
```
We get the following plans - 
V3 
```json
{
  "QueryType": "SELECT",
  "Original": "SELECT user.* FROM user INNER JOIN user_extra ue ON ue.user_id = user.id GROUP BY user.id;",
  "Instructions": {
    "OperatorType": "Route",
    "Variant": "SelectScatter",
    "Keyspace": {
      "Name": "user",
      "Sharded": true
    },
    "FieldQuery": "select `user`.* from `user` join user_extra as ue on ue.user_id = `user`.id where 1 != 1 group by `user`.id",
    "Query": "select `user`.* from `user` join user_extra as ue on ue.user_id = `user`.id group by `user`.id",
    "Table": "`user`, user_extra"
  }
}
```
Gen4
```json
Expression of SELECT list is not in GROUP BY clause and contains nonaggregated column '`user`.*' which is not functionally dependent on columns in GROUP BY clause; this is incompatible with sql_mode=only_full_group_by
```

So from the user perspective, the query doesn't work initially with `Gen4Fallback` but starts working after removing the keyspace name.

#### Reproduction Steps

1. Copy the query to `onecase.txt` and run the `TestOne` test.
2. View the results

#### Binary version
main

frouioui:
## Description

This pull request allows queries with a group by on a sharding key to pass through when it is a single route.

## Related Issue(s)

- fixes #9221 

## Checklist

### DIFF
--- a/go/vt/vtgate/planbuilder/abstract/queryprojection.go
+++ b/go/vt/vtgate/planbuilder/abstract/queryprojection.go
@@ -43,6 +43,7 @@ type (
 		OrderExprs         []OrderBy
 		CanPushDownSorting bool
 		HasStar            bool
+		ProjectionError    error
 	}
 
 	// OrderBy contains the expression to used in order by and also if ordering is needed at VTGate level then what the weight_string function expression to be sent down for evaluation.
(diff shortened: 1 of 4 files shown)


systay (approved):
LGTM

PR #9244, opened 2021-11-16
Fix owned vindex query to keep alias in table expression

### LINKED ISSUE #4257: OwnedVindexQuery keeps alias in where clause but drops alias in table expression
This test case in dml_cases.txt will pass, but you can see that it's generating an invalid `OwnedVindexQuery` query that references aliases without setting them:

```
# update by primary keyspace id, changing one vindex column, with alias
"update user_metadata as um set um.email = 'juan@vitess.io' where um.user_id = 1"
{
  "Original": "update user_metadata as um set um.email = 'juan@vitess.io' where um.user_id = 1",
  "Instructions": {
    "Opcode": "UpdateEqual",
    "Keyspace": {
      "Name": "user",
      "Sharded": true
    },
    "Query": "update user_metadata as um set um.email = 'juan@vitess.io' where um.user_id = 1",
    "Vindex": "user_index",
    "Values": [1],
    "ChangedVindexValues": {
      "email_user_map": ["juan@vitess.io"]
    },
    "Table": "user_metadata",
    "OwnedVindexQuery": "select email, address from user_metadata where um.user_id = 1 for update"
  }
}
```

harshit-gangal:
## Description

This PR fixes the table expression to keep table alias. Also, add table alias support in delete statement parsing.

## Related Issue(s)

- Fixes #4257

## Checklist


go/vt/vtgate/planbuilder/dml.go
```diff
@@ -179,15 +179,15 @@ func buildDMLPlan(vschema ContextVSchema, dmlType string, stmt sqlparser.Stateme
 	return edml, ksidVindex, ksidCol, nil
 }
 
-func generateDMLSubquery(where *sqlparser.Where, orderBy sqlparser.OrderBy, limit *sqlparser.Limit, table *vindexes.Table, ksidCol string) string {
+func generateDMLSubquery(tblExpr sqlparser.TableExpr, where *sqlparser.Where, orderBy sqlparser.OrderBy, limit *sqlparser.Limit, table *vindexes.Table, ksidCol string) string {
```
GuptaManan100, line 182: `table` is an unused argument now, isn't it?

GuptaManan100 (approved):

go/vt/vtgate/planbuilder/dml.go
harshit-gangal: it is used to get vindex columns for projection columns

PR #9245, opened 2021-11-16
git: optimize golangci-lint post-commit hook

vmg:
## Description

As suggested by @ajm188 in https://github.com/vitessio/vitess/pull/9049, the `--new` flag to `golangci-lint` that we're running in our post-commit hook doesn't work great if you have untracked files, if you're doing a partial commit, etc.

I'm switching the post-commit hook to work only on the packages that have actually been touched by the last commit. This code looks a lot like the one we had before except:

- We're running `revive` to check our style
- We're not running `golangci-lint` once for each package. We can just pass all the packages to a single invocation to significantly speed up processing time.

## Related Issue(s)
- https://github.com/vitessio/vitess/pull/9049

## Checklist


harshit-gangal (approved):

ajm188 (commented):
belated +1 of enthusiasm!! thank you! ❤️

PR #9248, opened 2021-11-16
Match collation between VTGate and VTTablet

frouioui:
## Description

This pull request enforces the collation match between VTGate and VTTablet.

## Related Issue(s)

- Follow up of #9097 

## Checklist


vmg (approved):
I paired with @frouioui to design this and I'm happy with the result.

PR #9177, opened 2021-11-10
[backport] Fix boolean parameter order in DropSources call for v2 flows

ajm188:
## Description

This PR backports #9175 to the 12.0 release branch

## Related Issue(s)

#9174 

## Checklist
- [x] Should this PR be backported? **no** (this is the backport)
- [x] Tests were added or are not required **n/a**
- [x] Documentation was added or is not required **n/a**

PR #9250, opened 2021-11-16
changing -dry-run to -dry_run

hallaroo:
Signed-off-by: hallaroo 

## Description
@aquarapid found a couple places in the help where [-dry_run] was replaced with [-dry-run]. This PR changes this back as [-dry-run] is not the correct usage, it should be the underscore one.

e.g.:

`SwitchReads Error: rpc error: code = Unknown desc = flag provided but not defined: -dry-run
E1116 15:14:07.388982   98844 main.go:76] remote error: rpc error: code = Unknown desc = flag provided but not defined: -dry-run
`

## Related Issue(s)
Not 100% sure when this was introduced but found the same issue in https://github.com/vitessio/vitess/pull/9177/ 

## Checklist
- [ - ] Tests were added or are not required
- [ - ] Documentation was added or is not required

### DIFF
--- a/go/vt/vtctl/vtctl.go
+++ b/go/vt/vtctl/vtctl.go
@@ -217,7 +217,7 @@ var commands = []commandGroup{
 			{
 				name:   "ChangeTabletType",
 				method: commandChangeTabletType,
-				params: "[-dry_run] <tablet alias> <tablet type>",
+				params: "[-dry-run] <tablet alias> <tablet type>",
 				help: "Changes the db type for the specified tablet, if possible. This command is used primarily to arrange replicas, and it will not convert a primary.\n" +
 					"NOTE: This command automatically updates the serving graph.\n",
 			},
@@ -702,7 +702,7 @@ var commands = []commandGroup{
 			{
 				name:   "ApplyVSchema",
 				method: commandApplyVSchema,
-				params: "{-vschema=<vschema> || -vschema_file=<vschema file> || -sql=<sql> || -sql_file=<sql file>} [-cells=c1,c2,...] [-skip_rebuild] [-dry_run] <keyspace>",
+				params: "{-vschema=<vschema> || -vschema_file=<vschema file> || -sql=<sql> || -sql_file=<sql file>} [-cells=c1,c2,...] [-skip_rebuild] [-dry-run] <keyspace>",
 				help:   "Applies the VTGate routing schema to the provided keyspace. Shows the result after application.",
 			},
 			{
--- a/go/vt/vtctl/vtctl.go
+++ b/go/vt/vtctl/vtctl.go
@@ -714,7 +714,7 @@ var commands = []commandGroup{
 			{
 				name:   "ApplyRoutingRules",
 				method: commandApplyRoutingRules,
-				params: "{-rules=<rules> || -rules_file=<rules_file>} [-cells=c1,c2,...] [-skip_rebuild] [-dry_run]",
+				params: "{-rules=<rules> || -rules_file=<rules_file>} [-cells=c1,c2,...] [-skip_rebuild] [-dry-run]",
 				help:   "Applies the VSchema routing rules.",
 			},
 			{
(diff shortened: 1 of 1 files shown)


go/vt/vtctl/vtctl.go
```diff
@@ -217,7 +217,7 @@ var commands = []commandGroup{
 			{
 				name:   "ChangeTabletType",
 				method: commandChangeTabletType,
-				params: "[-dry-run] <tablet alias> <tablet type>",
+				params: "[-dry_run] <tablet alias> <tablet type>",
```
deepthi, line 220: Should be reverted

go/vt/vtctl/vtctl.go
```diff
@@ -522,13 +522,13 @@ var commands = []commandGroup{
 			{
 				name:   "SwitchReads",
 				method: commandSwitchReads,
-				params: "[-cells=c1,c2,...] [-reverse] -tablet_type={replica|rdonly} [-dry-run] <keyspace.workflow>",
+				params: "[-cells=c1,c2,...] [-reverse] -tablet_type={replica|rdonly} [-dry_run] <keyspace.workflow>",
```
deepthi, line 525: :heavy_check_mark:

go/vt/vtctl/vtctl.go
```diff
@@ -522,13 +522,13 @@ var commands = []commandGroup{
...
 				help:   "Switch read traffic for the specified workflow.",
 			},
 			{
 				name:   "SwitchWrites",
 				method: commandSwitchWrites,
-				params: "[-timeout=30s] [-reverse] [-reverse_replication=true] [-dry-run] <keyspace.workflow>",
+				params: "[-timeout=30s] [-reverse] [-reverse_replication=true] [-dry_run] <keyspace.workflow>",
```
deepthi, line 531: :heavy_check_mark:

go/vt/vtctl/vtctl.go
```diff
@@ -702,7 +702,7 @@ var commands = []commandGroup{
 			{
 				name:   "ApplyVSchema",
 				method: commandApplyVSchema,
-				params: "{-vschema=<vschema> || -vschema_file=<vschema file> || -sql=<sql> || -sql_file=<sql file>} [-cells=c1,c2,...] [-skip_rebuild] [-dry-run] <keyspace>",
+				params: "{-vschema=<vschema> || -vschema_file=<vschema file> || -sql=<sql> || -sql_file=<sql file>} [-cells=c1,c2,...] [-skip_rebuild] [-dry_run] <keyspace>",
```
deepthi, line 705: Should be reverted

go/vt/vtctl/vtctl.go
```diff
@@ -714,7 +714,7 @@ var commands = []commandGroup{
 			{
 				name:   "ApplyRoutingRules",
 				method: commandApplyRoutingRules,
-				params: "{-rules=<rules> || -rules_file=<rules_file>} [-cells=c1,c2,...] [-skip_rebuild] [-dry-run]",
+				params: "{-rules=<rules> || -rules_file=<rules_file>} [-cells=c1,c2,...] [-skip_rebuild] [-dry_run]",
```
deepthi, line 717: Should be reverted

deepthi (changes requested):
I looked through the history of the file, and only two of the changes are needed.
The others are consistent between the help and actual command syntax, so should stay the same.
It is unfortunate that the earlier commands used hyphens but the newer ones use underscores (and have already been deprecated).

ajm188 (commented):
any reason we don't just switch all these to `DualFormatStringVar`? https://github.com/vitessio/vitess/blob/main/go/flagutil/flagutil.go#L140-L149

hallaroo:
> any reason we don't just switch all these to `DualFormatStringVar`? https://github.com/vitessio/vitess/blob/main/go/flagutil/flagutil.go#L140-L149

Just to get the help changed out would you be willing to move this to be its own issue? Or let me know and I can get one added. Thanks!

deepthi:
> any reason we don't just switch all these to `DualFormatStringVar`? https://github.com/vitessio/vitess/blob/main/go/flagutil/flagutil.go#L140-L149

No reason not to, but let's make that a separate task and not block this PR.

deepthi (approved):

Issue #8975, opened 2021-10-11
RFC - Durability and consensus in Vtorc


GuptaManan100:
This issue describes the feature for pluggable durability requirements and the consensus approach needed to fulfil those. It is also used to keep track of the work done and the pending things left to do before a GA release.

## Feature Description

Vtorc should be used in Vitess instead of orchestrator and it should support pluggable durability requirements that the user can specify according to their use case. For example, let A, B, C, D, E and F be Vitess `replica` type servers for a single shard. Then the user should be able to specify the servers that each of them need ACKs from to accept writes. They can also specify servers which are not capable of becoming primaries.
Example: [A-(B,C), D-(E), F] be one such set of rules. Here A needs acks from either B or C to accept writes. Similarly, D needs ACKs from E and F can be a primary on its own.
possible quorums from inferred from these rules - ([A,B], [A,C], [D,E], [F])

### Technical Details - 

There are three steps that each consensus algorithm must follow, namely - 
1. Revocation - revoking access from the previous leader to accept writes.
2. Propagation - propagate any transactions already accepted by the previous leader.
3. Establishment - reach enough servers to establish leadership by fulfilling quorum rules.

We now outline each of these phases in greater detail. 

#### Revocation
The first phase is to revoke access from the previous primary so that it can no longer accepts writes. This can be done either by directly contacting the primary or reaching enough of its replicas such that it cannot accept writes. Since we will use MySQL semi-sync, if we remove enough replicas, the primary will not accept any writes and will block.

The servers needed to contact to revoke access will be defined by the durability policy rules. For example, in the case described in the feature description, the possible server sets needed to be reached for revocation are ([A,E,F], [A,D,F], [B,C,E,F], [B,C,D,F]). 

#### Propagation
The next phase is propagation where we will need to find and propagate any transactions that the previous primary accepted or might have accepted if we are unable to determine conclusively that it was rejected. There will be at-most one GTID range which had been accepted and needs to be propagated. This GTID range will the one that is found from the servers during the revoke phase with the largest timestamp.

Here we discuss multiple failure scenarios to establish the rules that need to be followed. For all these cases we assume that the primary needs 2 acknowledgements to commit a transaction.

[...]

PR #9259, opened 2021-11-18
Make PRS use Durability Rules

GuptaManan100:
## Description

In the PR #8492, ERS was changed to use the durability policies to find the ideal candidate for being the primary. This PR makes that change to PRS. The only functionality which has been changed pertains to choosing the next primary. Instead of using the position sorter to find the most advanced tablet, we will now use the same function that ERS does for sorting the tablets and take the best alternative.

## Related Issue(s)

- #8975 

## Checklist


deepthi (approved):
LGTM, but maybe good to wait for @ajm188 to take a look as well.

go/vt/vtctl/reparentutil/reparent_sorter.go
```diff
@@ -26,68 +26,69 @@ import (
...
-	iPromotionRule := PromotionRule(ersISSorter.tablets[i])
+	jPromotionRule := PromotionRule(rs.tablets[j])
+	iPromotionRule := PromotionRule(rs.tablets[i])
 	return !jPromotionRule.BetterThan(iPromotionRule)
 }
 
-// sortTabletsForERS sorts the tablets, given their positions for emergency reparent shard
-func sortTabletsForERS(tablets []*topodatapb.Tablet, positions []mysql.Position) error {
+// sortTabletsForReparent sorts the tablets, given their positions for emergency reparent shard and planned reparent shard
+// primary sorting is via the tablet positions, and ties are broken by the promotion rules.
```
ajm188, line 84: nit: this is an unfortunate name collision, because you mean "primary" as in "the main sorting" and not the tablet type.

suggestion:

> Tablets are sorted first by their replication positions, with ties broken by the promotion rules.

(also can you add some punctuation to the previous line? i had to read it a few times to realize this wasn't all one sentence)

ajm188 (approved):
One comment wording nitpick, then good to merge after that!

go/vt/vtctl/reparentutil/reparent_sorter.go
GuptaManan100: Yes, that's an unfortunate conflict 😅. I didn't realise it when I wrote it. Thankyou for pointing it out. I have taken your suggestion

PR #9260, opened 2021-11-18
vtexplain: correct handling of column types for queries involving multiple tables

arthurschreiber:
## Description

When running `vtexplain` with a query that joins tables that have matching column names with differing types, it would fail with an error like `unknown error: column type mismatch for column`. This was caused by having similarly named columns (but with different types) on the different tables.

`vtexplain` builds an internal map of column names to their respective types, build only builds a single map for all tables in the query - so if the tables have columns that share the name but different types, the code would fail with an error.

This pull request changes `vtexplain` to build a map per table, which allows `vtexplain` to handle these types of queries correctly.

## Checklist

### DIFF
--- a/go/vt/vtexplain/vtexplain_vttablet_test.go
+++ b/go/vt/vtexplain/vtexplain_vttablet_test.go
@@ -25,6 +25,61 @@ import (
 	topodatapb "vitess.io/vitess/go/vt/proto/topodata"
 )
 
+func TestRun(t *testing.T) {
+	testVSchema := `
+{
+	"test_keyspace": {
+		"sharded": false,
+		"tables": {
+			"t1": {
+				"columns": [
+					{ "name": "id", "type": "INT64" }
+				],
+				"column_list_authoritative": true
+			},
+			"t2": {
+				"columns": [
+					{ "name": "id", "type": "INT32" }
+				],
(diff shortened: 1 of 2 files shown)


harshit-gangal (approved):
Thank you for contributing.

LGTM

arthurschreiber:
Looks like there's a bunch of test cases that are failing. I'll take another look! 🙇‍♂️

arthurschreiber:
@harshit-gangal Thanks for the quick approval. I just pushed an updated set of changes that actually passes all tests. The builds that are failing are flaky builds I guess?

go/vt/vtexplain/vtexplain_vttablet_test.go
```diff
@@ -25,6 +25,61 @@ import (
 	topodatapb "vitess.io/vitess/go/vt/proto/topodata"
 )
 
+func TestRun(t *testing.T) {
```
arthurschreiber, line 28: Should I replace this test with extending the tests in `go/vt/vtexplain/vtexplain_flaky_test.go` instead?

deepthi:
Merging based on harshit's approval.
I think the current test is fine. If another one is required in vtexplain_flaky_test, that can be added later.

Issue #9214, opened 2021-11-12
typeorm: add support for information schema query coming from framework


chreniuc:
I wanted to use [typeorm](https://typeorm.io/#/) with vitess, but I'm encountering multiple problems. One was fixed by this issue: https://github.com/vitessio/vitess/issues/9139#issuecomment-965417512

After that fix was integrated I've encountered another issue.

This is an introduction, it the same as from the previous issue:

>I'm using mysql driver for typeorm, from nodejs. Typeorm has an option: [`synchronize`](https://typeorm.io/#/undefined/creating-a-connection-to-the-database), if that is set to true, it will start a transaction and it will query the `INFORMATION_SCHEMA` to see if the current db is up to date, if not, it will alter the tables.
>This was an empty database with only one shard, it was running on a k8 cluster using the operator.

When I start the app and it does those queries, the vtgate crashes and it logs the following:

```bash
[...]

PR #9263, opened 2021-11-18
Gen4: filter on derived union system table with star projection

harshit-gangal:
## Description

This PR add support for filter on derived union system table with star projection.

## Related Issue(s)

- Closes https://github.com/vitessio/vitess/issues/9214

## Checklist

### DIFF
--- a/go/vt/vtgate/planbuilder/abstract/concatenate.go
+++ b/go/vt/vtgate/planbuilder/abstract/concatenate.go
@@ -44,8 +44,20 @@ func (c *Concatenate) TableID() semantics.TableSet {
 }
 
 // PushPredicate implements the Operator interface
-func (c *Concatenate) PushPredicate(sqlparser.Expr, *semantics.SemTable) error {
-	return vterrors.Errorf(vtrpcpb.Code_UNIMPLEMENTED, "can't push predicates on concatenate")
+func (c *Concatenate) PushPredicate(expr sqlparser.Expr, semTable *semantics.SemTable) error {
+	for index, source := range c.Sources {
+		if len(c.SelectStmts[index].SelectExprs) != 1 {
+			return vterrors.Errorf(vtrpcpb.Code_UNIMPLEMENTED, "can't push predicates on concatenate")
+		}
+		if _, isStarExpr := c.SelectStmts[index].SelectExprs[0].(*sqlparser.StarExpr); !isStarExpr {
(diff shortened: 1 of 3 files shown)


go/vt/vtgate/planbuilder/abstract/concatenate.go
```diff
@@ -44,8 +44,20 @@ func (c *Concatenate) TableID() semantics.TableSet {
...
-func (c *Concatenate) PushPredicate(sqlparser.Expr, *semantics.SemTable) error {
-	return vterrors.Errorf(vtrpcpb.Code_UNIMPLEMENTED, "can't push predicates on concatenate")
+func (c *Concatenate) PushPredicate(expr sqlparser.Expr, semTable *semantics.SemTable) error {
+	for index, source := range c.Sources {
+		if len(c.SelectStmts[index].SelectExprs) != 1 {
+			return vterrors.Errorf(vtrpcpb.Code_UNIMPLEMENTED, "can't push predicates on concatenate")
+		}
+		if _, isStarExpr := c.SelectStmts[index].SelectExprs[0].(*sqlparser.StarExpr); !isStarExpr {
+			return vterrors.Errorf(vtrpcpb.Code_UNIMPLEMENTED, "can't push predicates on concatenate")
+		}
```
GuptaManan100, line 54: Is there a reason why we only allow pushing predicates in case of `*` expressions? Why don't we allow pushing it always to all the sources?

GuptaManan100 (approved):

frouioui (approved):

GuptaManan100 (commented):
Also, how does this change fix the panic? The planbuilding would have failed if we tried pushing predicate on the concatenate, or are we ignoring the error from this somewhere?

go/vt/vtgate/planbuilder/abstract/concatenate.go
harshit-gangal: the problem is how we resolve dependency for the predicate expression, it will only be pushed to first statement in the union.
for star expression, the dependency is none so it gets pushed to all the select statements in the union.

harshit-gangal:
> Also, how does this change fix the panic? The planbuilding would have failed if we tried pushing predicate on the concatenate, or are we ignoring the error from this somewhere?

panic was fixed by other PR, this is a follow-up issue.

PR #9264, opened 2021-11-18
mysql: Pass mysql.Conn through {Hash,PlainText,Caching}Storage interfaces

mattrobenolt:
The TLS client certificates can simply be derived based on the
mysql.Conn passed through, but passing through the actual Conn, gives us
access to Conn.ClientData or anything else an Auth{Method,Server} might need to
implement.

It's arguable that if we need access to the lower level Conn, we could
simply implement the actual HandleAuthPluginData instead of these
*Storage interfaces, and that's true. But not all AuthMethods can get
away with this because they need to write to the Conn's connection using
private functions.

For example, the `mysqlCachingSha2AuthMethod` needs to write more data
on the socket after calling `UserEntryWithCacheHash`, so to write a new
`UserEntryWithCacheHash` that needs data from `mysql.Conn`, I'm left
wtih no other options since I can't vendor or reimplement the rest of
the logic without having access to the private functions/methods within
the `mysql` package.

## Checklist
- [ ] Should this PR be backported? (no)
- [x] Tests were added or are not required (no, existing tests were modified)
- [ ] Documentation was added or is not required (In changelog probably?)

### DIFF
--- a/go/mysql/auth_server.go
+++ b/go/mysql/auth_server.go
@@ -22,7 +22,6 @@ import (
 	"crypto/sha1"
 	"crypto/sha256"
 	"crypto/subtle"
-	"crypto/x509"
 	"encoding/hex"
 	"net"
 	"sync"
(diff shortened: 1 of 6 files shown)


deepthi (approved):

deepthi:
@mattrobenolt welcome to vitess and thank you for the contribution!

PR #9267, opened 2021-11-19
collations: Wildcard matching for the 'LIKE' operator

vmg:
## Description

I've added support to the `collations` package to perform collation-aware wildcard matching. I've written an API optimized for our evaluation engine, where you pass in a pattern (i.e. the right side of the `LIKE` operator) and it compiles it to a `WildcardPattern` matcher that can be repeatedly be called on different values for the left side to see if they match.

The wildcard patterns are pre-processed and optimized: any wildcards that can be matched using `Collate` directly (i.e. wildcards without any special matching characters, or wildcards where the only special character is at the end, `"foobar%"`) will not use the wildcard matcher.

I've added comprehensive tests for all collations and many different patterns against a live MySQL instance.

cc @systay @frouioui 

## Related Issue(s)
- https://github.com/vitessio/vitess/issues/8606

## Checklist

### DIFF
--- a/go/mysql/collations/8bit.go
+++ b/go/mysql/collations/8bit.go
@@ -20,7 +20,13 @@ import (
 	"vitess.io/vitess/go/mysql/collations/internal/charset"
 )
 
+var sortOrderIdentity [256]byte
+
 func init() {
+	for i := range sortOrderIdentity {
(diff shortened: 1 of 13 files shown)


frouioui (approved):
Looks 👌🏻 to me!

PR #9269, opened 2021-11-19
Better handling of partial dml query exec in transaction

### LINKED ISSUE #9266: RFC: Rollback of partial failed query in an explicit transaction
# Problem Statement
In a transaction, when a single DML (data manipulation language) query is broken into sub-queries like a scatter query or a vindex query. Then, any error in executing those queries leads to the rollback of the complete transaction.

This helps in guaranteeing that no unexpected data is committed on the transaction but it also leads to a bad user experience on handling distributed query errors inside transactions.

# Proposed Solution
In an explicit transaction created by the user, any DML query that executes on the multi-shard or needs a vindex update will execute in the scope of a savepoint. If there is any partial failure then the effect of that DML query will be rollback to the savepoint. In case the query completes successfully, savepoint will not be released, and only on a commit/rollback by the user, it will be cleared.

In an auto-commit mode, when a DML goes to multi-shard or vindex is updated, the implicit transaction created at the VTGate level will still have the same behavior of rollback of the transaction i.e. no impact of the failed DML query on the database.

# Test Cases

| S.No. | Test Case | Expectation |
| ----- | ------------- | ------------- |
| 1 | Begin -> partial failing query | transaction remains open; partial change rolled back |
| 2 | Begin -> partial vindex failing query | transaction remains open; partial change rolled back |
| 3 | Autocommit false -> partial failing query | transaction remains open; partial change rolled back|
| 4 | Autocommit false -> partial vindex failing query | transaction remains open; partial change rolled back |
| 5 | Partial change rollback failed (rollback to savepoint) | transaction is rolled back |
| 6 | Autcommit true -> partial failed query | Transaction is rolled back; No change on the database |

All 4 the cases in OLTP, OLTP-reserved, OLAP, OLAP-reserved

harshit-gangal:
## Description

This PR reverses the effect of partial dml query execution instead of rolling back the complete transaction when the transaction is explicitly created by the user.

## Related Issue(s)

- Closes #9266 

## Checklist

### DIFF
--- a/go/test/endtoend/vtgate/partialfailure/main_test.go
+++ b/go/test/endtoend/vtgate/partialfailure/main_test.go
@@ -1,0 +1,170 @@
+/*
+Copyright 2020 The Vitess Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package reservedconn
+
+import (
+	"context"
+	"flag"
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"vitess.io/vitess/go/test/endtoend/vtgate/utils"
+
+	"vitess.io/vitess/go/mysql"
+	"vitess.io/vitess/go/test/endtoend/cluster"
+)
+
+var (
+	clusterInstance *cluster.LocalProcessCluster
+	vtParams        mysql.ConnParams
+	keyspaceName    = "ks"
+	cell            = "zone1"
+	hostname        = "localhost"
+	sqlSchema       = `
+	create table test(
+		id bigint,
+		val1 varchar(16),
+		val2 int,
+		val3 float,
+		primary key(id)
+	)Engine=InnoDB;
+
+CREATE TABLE test_vdx (
+    val1 varchar(16) NOT NULL,
+    keyspace_id binary(8),
+    UNIQUE KEY (val1)
+) ENGINE=Innodb;
+`
+
+	vSchema = `
+		{	
+			"sharded":true,
+			"vindexes": {
+				"hash_index": {
+					"type": "hash"
+				},
+				"lookup1": {
+					"type": "consistent_lookup",
+					"params": {
+						"table": "test_vdx",
+						"from": "val1",
+						"to": "keyspace_id",
+						"ignore_nulls": "true"
+					},
+					"owner": "test"
+				},
+				"unicode_vdx":{
+					"type": "unicode_loose_md5"
+                }
+			},	
+			"tables": {
+				"test":{
+					"column_vindexes": [
+						{
+							"column": "id",
+							"name": "hash_index"
+						},
+						{
+							"column": "val1",
+							"name": "lookup1"
+						}
+					]
+				},
+				"test_vdx":{
+					"column_vindexes": [
+						{
+							"column": "val1",
+							"name": "unicode_vdx"
+						}
+					]
+				}
+			}
+		}
+	`
+)
+
+func TestMain(m *testing.M) {
+	defer cluster.PanicHandler(nil)
+	flag.Parse()
+
+	exitCode := func() int {
+		clusterInstance = cluster.NewCluster(cell, hostname)
+		defer clusterInstance.Teardown()
+
+		// Start topo server
+		if err := clusterInstance.StartTopo(); err != nil {
+			return 1
+		}
+
+		// Start keyspace
+		keyspace := &cluster.Keyspace{
+			Name:      keyspaceName,
+			SchemaSQL: sqlSchema,
+			VSchema:   vSchema,
+		}
+		if err := clusterInstance.StartKeyspace(*keyspace, []string{"-40", "40-80", "80-c0", "c0-"}, 0, false); err != nil {
+
+			return 1
+		}
+
+		// Start vtgate
+		if err := clusterInstance.StartVtgate(); err != nil {
+			return 1
+		}
+
+		vtParams = mysql.ConnParams{
+			Host: clusterInstance.Hostname,
+			Port: clusterInstance.VtgateMySQLPort,
+		}
+		return m.Run()
+	}()
+	os.Exit(exitCode)
+}
+
+func TestPartialQueryFailure(t *testing.T) {
+	tcases := []struct {
+		mode string
+	}{
+		{"set workload = oltp"},
+		{"set workload = oltp; set sql_mode = ''"},
+		{"set workload = olap"},
+		{"set workload = olap; set sql_mode = ''"},
+	}
+
+	for _, tc := range tcases {
+		t.Run(tc.mode, func(t *testing.T) {
+			conn, err := mysql.Connect(context.Background(), &vtParams)
+			require.NoError(t, err)
+			defer conn.Close()
+
+			// setup the mode
+			utils.Exec(t, conn, tc.mode)
+
+			// cleanup previous run data from table.
+			utils.Exec(t, conn, `delete from test`)
+
+			utils.Exec(t, conn, `begin`)
+			utils.Exec(t, conn, `insert into test(id, val1) values (1,'A'),(2,'B'),(3,'C')`)
+			utils.AssertContainsError(t, conn, `insert into test(id, val1) values (1,'D'),(4,'E')`, `AlreadyExists`)
+			utils.AssertMatches(t, conn, `select id, val1 from test order by id`, `[[INT64(1) VARCHAR("A")] [INT64(2) VARCHAR("B")] [INT64(3) VARCHAR("C")]]`)
+			utils.Exec(t, conn, `commit`)
+			utils.AssertMatches(t, conn, `select id, val1 from test order by id`, `[[INT64(1) VARCHAR("A")] [INT64(2) VARCHAR("B")] [INT64(3) VARCHAR("C")]]`)
+		})
+	}
+}
--- a/go/test/endtoend/vtgate/partialfailure/main_test.go
+++ b/go/test/endtoend/vtgate/partialfailure/main_test.go
@@ -164,7 +164,7 @@ func TestPartialQueryFailure(t *testing.T) {
 
 			utils.Exec(t, conn, `begin`)
 			utils.Exec(t, conn, `insert into test(id, val1) values (1,'A'),(2,'B'),(3,'C')`)
-			utils.AssertContainsError(t, conn, `insert into test(id, val1) values (1,'D'),(4,'E')`, `failed to execute due to partial DML execution`)
+			utils.AssertContainsError(t, conn, `insert into test(id, val1) values (1,'D'),(4,'E')`, `revered partial DML execution failure`)
 			utils.AssertMatches(t, conn, `select id, val1 from test order by id`, `[[INT64(1) VARCHAR("A")] [INT64(2) VARCHAR("B")] [INT64(3) VARCHAR("C")]]`)
 			utils.Exec(t, conn, `commit`)
 			utils.AssertMatches(t, conn, `select id, val1 from test order by id`, `[[INT64(1) VARCHAR("A")] [INT64(2) VARCHAR("B")] [INT64(3) VARCHAR("C")]]`)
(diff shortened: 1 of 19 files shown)


go/test/endtoend/vtgate/partialfailure/main_test.go
```diff
@@ -0,0 +1,561 @@
...
+			utils.AssertContainsError(t, conn, `insert into test(id, val1) values (1,'D'),(4,'E')`, `transaction rolled back to reverse changes of partial DML execution`)
+			utils.AssertMatches(t, conn, `select id, val1 from test order by id`, `[[INT64(1) VARCHAR("A")] [INT64(2) VARCHAR("B")] [INT64(3) VARCHAR("C")]]`)
+			utils.Exec(t, conn, `rollback`)
+			utils.AssertMatches(t, conn, `select id, val1 from test order by id`, `[[INT64(1) VARCHAR("A")] [INT64(2) VARCHAR("B")] [INT64(3) VARCHAR("C")]]`)
+		})
+	}
+}
+
+func TestPartialVindexQueryFailureAutoCommitRollback(t *testing.T) {
+	tcases := []struct {
```
systay, line 529: nit: this pattern is copy pasted over and over again in this file. I think we should extract to a single place

go/vt/vtexplain/vtexplain.go
```diff
@@ -336,6 +341,26 @@ func ExplainsAsText(explains []*Explain) string {
 	return b.String()
 }
 
+func specialHandlingOfSavepoints(q *MysqlQuery) {
+	if strings.Contains(q.SQL, "savepoint") {
+		stmt, _ := sqlparser.Parse(q.SQL)
```
systay, line 346: we should not ignore the error, and we should use a checked type conversion so we don't panic if we got something other than a savepoint

go/vt/vtgate/executor.go
```diff
@@ -582,34 +582,51 @@ func (e *Executor) handleSavepoint(ctx context.Context, safeSession *SafeSession
...
+	qr, err := e.executeSPInAllSessions(ctx, safeSession, sql, ignoreMaxMemoryRows)
+	safeSession.SetCommitOrder(orig)
 	if err != nil {
 		return nil, err
 	}
 	safeSession.StoreSavepoint(sql)
 	return qr, nil
 }
 
+func (e *Executor) executeSPInAllSessions(ctx context.Context, safeSession *SafeSession, sql string, ignoreMaxMemoryRows bool) (*sqltypes.Result, error) {
```
systay, line 603: maybe it's too early, but I'm not sure what SPI means here. Would you mind adding a comment explaining?

go/vt/vtgate/executor.go
```diff
@@ -1432,6 +1449,18 @@ func (e *Executor) getPlan(vcursor *vcursorImpl, sql string, comments sqlparser.
 	return e.checkThatPlanIsValid(stmt, plan)
 }
 
+func ignoreMaxPayloadSize(stmt sqlparser.Statement) bool {
```
systay, line 1452: why not add this to `sqlparser.IgnoreMaxPayloadSizeDirective` instead?

go/vt/vtgate/executor_dml_test.go
```diff
@@ -53,12 +55,8 @@ func TestUpdateEqual(t *testing.T) {
 		Sql:           "update `user` set a = 2 where id = 1",
 		BindVariables: map[string]*querypb.BindVariable{},
 	}}
-	if !reflect.DeepEqual(sbc1.Queries, wantQueries) {
-		t.Errorf("sbc1.Queries: %+v, want %+v\n", sbc1.Queries, wantQueries)
-	}
-	if sbc2.Queries != nil {
-		t.Errorf("sbc2.Queries: %+v, want nil\n", sbc2.Queries)
-	}
+	testQueries(t, sbc1, wantQueries)
```
systay, line 58: should we rename this to `assertQueries` or something? this method is just asserting that we got the right queries back, right?

go/vt/vtgate/executor_framework_test.go
```diff
@@ -572,10 +571,47 @@ func executorStream(executor *Executor, sql string) (qr *sqltypes.Result, err er
...
+		}
+		require.Equal(t, query.BindVariables, wantQueries[idx].BindVariables)
+		got := query.Sql
+		expected := wantQueries[idx].Sql
+		idx++
+		assert.Equal(t, expected, got)
+	}
+}
+
+func testQueriesWithSavepoint(t *testing.T, sbc *sandboxconn.SandboxConn, wantQueries []*querypb.BoundQuery) {
```
systay, line 592: same this as above - since this method is just asserting state, could we name it something with assert, check or require?

go/vt/vtgate/safe_session.go
```diff
@@ -40,6 +40,11 @@ type SafeSession struct {
 	mustRollback    bool
 	autocommitState autocommitState
 	commitOrder     vtgatepb.CommitOrder
+	savepointState  savepointState
+	// rollbackOnPartialExec is set if any DML was successfully
```
systay, line 44: set to what? why is this a string?

go/vt/vtgate/vcursor_impl.go
```diff
@@ -423,6 +421,8 @@ func (vc *vcursorImpl) StreamExecutePrimitive(primitive engine.Primitive, bindVa
 	return vterrors.New(vtrpcpb.Code_UNAVAILABLE, "upstream shards are not available")
 }
 
+const txRollback = "Rollback Transaction"
```
systay, line 424: nit: weird placement for a const

go/vt/vtgate/executor.go
frouioui: I think the function's name translates to "Execute Save Points In All Sessions", but we could add a comment explaining what "sp" means. In `vtexplain.go:302` I also got confused the first time I saw "sp".

go/vt/vtgate/executor_framework_test.go
frouioui: I am thinking we could probably move these two methods in the `sandboxconn` package and make them public so we can reuse them everywhere.

go/vt/vtgate/plan_execute.go
```diff
@@ -179,6 +179,10 @@ func (e *Executor) insideTransaction(ctx context.Context, safeSession *SafeSessi
 	// at the beginning, but never after.
 	safeSession.SetAutocommittable(mustCommit)
 
+	// If we to instantly commit the query, then there is no need to add savepoints.
```
frouioui, line 182: `// If we to instantly` => `// If we want to instantly`

go/vt/vtgate/safe_session.go
```diff
@@ -166,6 +187,33 @@ func (session *SafeSession) AutocommitApproval() bool {
 	return false
 }
 
+// SetSavepointState sets the state to insertSavepoints if true and noInsertSavepoints if false
```
frouioui, line 190:
```suggestion
// SetSavepointState sets the state to insertSavepoints if true and noInsertSavepoints is false
```

go/vt/vttablet/sandboxconn/sandboxconn.go
```diff
@@ -55,6 +55,9 @@ type SandboxConn struct {
 	MustFailStartCommit         int
 	MustFailSetRollback         int
 	MustFailConcludeTransaction int
+	// MustFailExecute is keyed by the statement type and stores the number
+	// of errors to return for that statement type.
+	MustFailExecute map[sqlparser.StatementType]int
```
frouioui, line 60: The name is a little confusing considering the comment

go/vt/vtgate/executor_framework_test.go
harshit-gangal: they are used only for executor tests, if it is used/required anywhere else we can move it later.

go/vt/vtgate/safe_session.go
harshit-gangal: rollbackOnPartialExec will store the `rollback to <savepoint>` sql query which is generated randomly when doing MarkSavepoint.

go/vt/vtgate/vcursor_impl.go
harshit-gangal: earlier it was near to usage, then the code the moved,
now moved it back to the usage.

go/vt/vtgate/safe_session.go
harshit-gangal: the statement is correct, suggestion means that it is not clear, will rephrase it.

go/vt/vttablet/sandboxconn/sandboxconn.go
harshit-gangal: it is similar to the naming convention of other MustFails in the struct.

harshit-gangal: added more comment for clarity

go/vt/vtgate/autocommit_test.go
```diff
@@ -385,10 +385,10 @@ func TestAutocommitTransactionStarted(t *testing.T) {
 	_, err := executor.Execute(context.Background(), "TestExecute", NewSafeSession(session), sql, map[string]*querypb.BindVariable{})
 	require.NoError(t, err)
 
-	testQueries(t, "sbc1", sbc1, []*querypb.BoundQuery{{
-		Sql:           "update `user` set a = 2 where id = 1",
-		BindVariables: map[string]*querypb.BindVariable{},
-	}})
+	require.Equal(t, 2, len(sbc1.Queries))
+	require.Contains(t, sbc1.Queries[0].Sql, "savepoint")
+	require.Equal(t, sbc1.Queries[1].Sql, "update `user` set a = 2 where id = 1")
```
frouioui, line 390: the expected value is the second argument and the actual value is the third argument of `require.Equal`
```suggestion
	require.Equal(t, "update `user` set a = 2 where id = 1", sbc1.Queries[1].Sql)
```

go/vt/vttablet/sandboxconn/sandboxconn.go
frouioui: okay got it

frouioui (approved):
Looks good to me. Agreed with @systay on https://github.com/vitessio/vitess/pull/9269#discussion_r758064993, we should avoid the repetition in the test suite to make adding/changing tests easier

go/vt/vtexplain/vtexplain.go
```diff
@@ -333,7 +341,40 @@ func ExplainsAsText(explains []*Explain) string {
 		fmt.Fprintf(&b, "\n")
 	}
 	fmt.Fprintf(&b, "----------------------------------------------------------------------\n")
-	return b.String()
+	return b.String(), nil
+}
+
+func specialHandlingOfSavepoints(q *MysqlQuery) error {
+	if !strings.Contains(q.SQL, "savepoint") {
```
systay, line 348: nit: can't we just skip this string check since we are looking at the type of the AST below?

harshit-gangal: This is just to avoid parsing overhead.

go/vt/vtgate/safe_session.go
systay: we should add :point_up: to the comment

go/vt/vtgate/safe_session.go
```diff
@@ -166,6 +187,34 @@ func (session *SafeSession) AutocommitApproval() bool {
 	return false
 }
 
+// SetSavepointState sets the state only once for the complete query execution life.
+// Calling the function multiple times will have no effect, only the first call would be used.
+// Default state is savepointStateNotSet,
+// if savepoint needed (spNeed true) then it will be set to savepointNeeded otherwise savepointNotNeeded.
+
+func (session *SafeSession) SetSavepointState(spNeed bool) {
```
systay, line 195: comment missing

go/vt/vtgate/safe_session.go
```diff
@@ -576,3 +625,24 @@ func (session *SafeSession) getSelectLimit() int {
...
 }
+
+func (session *SafeSession) isTxOpen() bool {
+	session.mu.Lock()
+	defer session.mu.Unlock()
+
+	return len(session.ShardSessions) > 0 || len(session.PreSessions) > 0 || len(session.PostSessions) > 0
+}
+
+func (session *SafeSession) GetSessions() []*vtgatepb.Session_ShardSession {
```
systay, line 636: comment missing

systay (commented):
I was too late to accept it, but I would have if given the chance :)

LGTM :+1:

go/vt/vtexplain/vtexplain.go
systay: in vtexplain? doesn't feel like it's part of the code where we need to squeeze extra drops of performance. ¯\_(ツ)_/¯

go/vt/vtgate/safe_session.go
harshit-gangal: there was extra line, comments are there :)

go/vt/vtgate/safe_session.go
harshit-gangal: added in #9294

PR #9270, opened 2021-11-20
Use relay log position in PRS while choosing the new primary

### LINKED ISSUE #5991: PlannedReparent improvements
#### Overview of the Issue
During a planned reparent, it is possible that `PromoteSlaveWhenCaughtUp` succeeds without an error but the outer context has been canceled. In that situation, we still call `UndoDemoteMaster` with a background context. 
By design, the old master discovers that it is no longer master and changes its type to REPLICA. However, the call to UndoDemoteMaster is ~unnecessary~ insufficient in this situation (as explained further below).
When the outer context is canceled at a point where `PromoteSlaveWhenCaughtUp` has completed, other replicas are still pointing to the old master, so we can end up with a situation like this:

```
               New Master <---- Old Master <---- Other replicas
```
To prevent this, we should consider two additional safeguards:
- add logic in `PromoteSlaveWhenCaughtUp` to revert any partial changes in case of errors.
- create an `UndoPromoteSlave` that is called along with `UndoDemoteMaster`.

Code snippet for reference:
https://github.com/vitessio/vitess/blame/master/go/vt/wrangler/reparent.go#L605
```
		rp, err = wr.tmc.PromoteSlaveWhenCaughtUp(promoteCtx, masterElectTabletInfo.Tablet, rp)
		if err != nil || (ctx.Err() != nil && ctx.Err() == context.DeadlineExceeded) {
			// If we fail to promote the new master, try to roll back to the
			// original master before aborting.
			// It is possible that we have used up the original context, or that
			// not enough time is left on it before it times out.
			// But at this point we really need to be able to Undo so as not to
			// leave the cluster in a bad state.
			// So we create a fresh context based on context.Background().
			undoCtx, undoCancel := context.WithTimeout(context.Background(), *topo.RemoteOperationTimeout)
			defer undoCancel()
			if err1 := wr.tmc.UndoDemoteMaster(undoCtx, oldMasterTabletInfo.Tablet); err1 != nil {
				log.Warningf("Encountered error %v while trying to undo DemoteMaster", err1)
			}
			return fmt.Errorf("master-elect tablet %v failed to catch up with replication or be upgraded to master: %v", masterElectTabletAliasStr, err)
		}
```

GuptaManan100:
## Description

While choosing the new primary in PRS, we should be using the relay log position, if it is available. This PR makes that change and adds tests for it.

## Related Issue(s)

- Closes #5991

## Checklist

## Deployment Notes

This PR has been created on top of #9259, and it should be merged after it.

### DIFF
--- a/go/vt/vtctl/reparentutil/util.go
+++ b/go/vt/vtctl/reparentutil/util.go
@@ -125,7 +125,7 @@ func findPositionForTablet(ctx context.Context, tablet *topodatapb.Tablet, logge
 		return mysql.Position{}, err
 	}
 
-	// Use the relay log position if available, otherwise use the executed GTID set executed.
+	// Use the relay log position if available, otherwise use the executed GTID set (binary log position).
 	positionString := status.Position
 	if status.RelayLogPosition != "" {
 		positionString = status.RelayLogPosition
--- a/go/vt/vtctl/reparentutil/util.go
+++ b/go/vt/vtctl/reparentutil/util.go
@@ -125,9 +125,14 @@ func findPositionForTablet(ctx context.Context, tablet *topodatapb.Tablet, logge
 		return mysql.Position{}, err
 	}
 
-	pos, err := mysql.DecodePosition(status.Position)
+	// Use the relay log position if available, otherwise use the executed GTID set executed.
+	positionString := status.Position
+	if status.RelayLogPosition != "" {
+		positionString = status.RelayLogPosition
+	}
+	pos, err := mysql.DecodePosition(positionString)
 	if err != nil {
-		logger.Warningf("cannot decode replica position %v for tablet %v, ignoring tablet: %v", status.Position, topoproto.TabletAliasString(tablet.Alias), err)
+		logger.Warningf("cannot decode replica position %v for tablet %v, ignoring tablet: %v", positionString, topoproto.TabletAliasString(tablet.Alias), err)
 		return mysql.Position{}, err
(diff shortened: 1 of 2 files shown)


go/vt/vtctl/reparentutil/util.go
```diff
@@ -81,17 +84,59 @@ func ChooseNewPrimary(
...
+	ctx, cancel := context.WithTimeout(ctx, waitTimeout)
+	defer cancel()
+
+	status, err := tmc.ReplicationStatus(ctx, tablet)
+	if err != nil {
+		logger.Warningf("failed to get replication status from %v, ignoring tablet: %v", topoproto.TabletAliasString(tablet.Alias), err)
+		return mysql.Position{}, err
+	}
+
+	// Use the relay log position if available, otherwise use the executed GTID set executed.
```
deepthi, line 128:
```suggestion
	// Use the relay log position if available, otherwise use the executed GTID set (binary log position).
```

deepthi (approved):
LGTM except for a nit.
I assume this needs to be rebased once #9259 is merged.

go/vt/vtctl/reparentutil/util.go
GuptaManan100: Fixed

GuptaManan100:
Yes, I'll rebase it once that PR is merged

ajm188 (approved):
LGTM

PR #9276, opened 2021-11-23
Merge InitShardPrimary functionality into PlannedReparentShard

GuptaManan100:
## Description

This PR merges the functionality of InitShardPrimary and PlannedReparentShard, so that just calling PRS even for initialisation would be sufficient.

However, PlannedReparentShard would not do the destructive step of copying the executed gtid set from the new primary to all the other replicas. PlannedReparentShard is meant to be used for non-destructive repairable operations. For this use case, it is still advised that users use InitShardPrimary.

## Related Issue(s)

- #8975 

## Checklist
- [ ] Documentation was added or is not required Documentation is required to say that PRS should be goto option to initialise the cluster


go/test/endtoend/vreplication/cluster.go
```diff
@@ -340,20 +340,15 @@ func (vc *VitessCluster) AddShards(t testing.TB, cells []*Cell, keyspace *Keyspa
...
 				log.Infof("Running Setup() for vttablet %s", tablets[ind].Name)
 				if err := tablet.Vttablet.Setup(); err != nil {
 					t.Fatalf(err.Error())
 				}
 			}
 		}
 		require.NotEqual(t, 0, primaryTabletUID, "Should have created a primary tablet")
-		log.Infof("InitShardPrimary for %d", primaryTabletUID)
-		require.NoError(t, vc.VtctlClient.InitShardPrimary(keyspace.Name, shardName, cells[0].Name, primaryTabletUID))
+		log.Infof("PlannedReparentShard for %d", primaryTabletUID)
```
deepthi, line 350: Should we log InitializeShard instead of PRS?

go/vt/vtctl/reparentutil/planned_reparenter.go
```diff
@@ -490,22 +517,36 @@ func (pr *PlannedReparenter) reparentShardLocked(
 
 	currentPrimary := FindCurrentPrimary(tabletMap, pr.logger)
 	reparentJournalPos := ""
+	// needsReload is used to keep track whether we need to refresh the state
```
deepthi, line 520:
```suggestion
	// needsReload is used to keep track of whether we need to refresh the state
```
```suggestion
	// needsReload is used to keep track whether we need to refresh the state
```

go/vt/vtctl/reparentutil/planned_reparenter.go
```diff
@@ -490,22 +517,36 @@ func (pr *PlannedReparenter) reparentShardLocked(
...
+	// of the new primary tablet. The only case that we need to reload the state
+	// is when we are initializing the new primary. The reason is that the first
+	// time we try to setup all the components like vreplication.Engine, they fail
+	// since the database isn't created until we setServing.
+	// A call to Refresh state fixes all the components. This isn't strictly necessary
+	// in the sense that all the components will retry initialization anyways after some
+	// time, so even without a call to RefreshState, they all converge correctly.
+	needsReload := false
```
deepthi, line 528: Let's rename this to `needsRefresh`. reload makes me think we are reloading topo information.

deepthi (approved):
LGTM. Nice work!
Let's wait for @sougou to give it a once-over before we merge.
Can you fix the little things I found in the meantime?

go/vt/vtctl/reparentutil/planned_reparenter.go
GuptaManan100: fixed

go/vt/vtctl/reparentutil/planned_reparenter.go
GuptaManan100: fixed

go/test/endtoend/vreplication/cluster.go
GuptaManan100: fixed

sougou (approved):

Issue #9275, opened 2021-11-22
Binary Hash VIndex Function With HexVals Always Uses First Shard


mattlord:
#### Overview of the Issue

In https://github.com/vitessio/vitess/pull/9118 we began to normalize hex values -- Base16 Numbers (HexNum) such as `0xA1` and Binary as Hex (HexVal) such as `x'A1'` -- in queries so that the planner cache is used for them.

For HexVals we had use the original string literal of the form `x'A1'` in the BindVar and pass that original user value on to mysqld (in the golang parser we decode the string literal to the hex value of `A1`).

The vindex functions, however, use the BindVar values when mapping the column value to a keyspace ID and so should use _the binary value actually stored in the mysql column after handling the string literal_: `A1`. Currently we are not doing that and thus the binary hash is done on `x'A1'` when it should be done on `A1`.

#### Reproduction Steps

Using the docker_local container from main:
```
make docker_local && ./docker/local/run.sh

./201_customer_tablets.sh ; ./202_move_tables.sh ; sleep 10; ./203_switch_reads.sh ; ./204_switch_writes.sh ; ./205_clean_commerce.sh ; ./301_customer_sharded.sh ; ./302_new_shards.sh ; ./303_reshard.sh ; sleep 10; ./304_switch_reads.sh ; ./305_switch_writes.sh ; ./306_down_shard_0.sh

vtctlclient GetVSchema customer > /tmp/customer.json

vtctlclient ApplyVSchema -vschema "$(jq '. | .vindexes += {"binary":{"type":"binary"}}' < /tmp/customer.json)" customer

mysql customer -e "select shard from customer.binary where id = x'7901'"
mysql customer -e "select shard from customer.binary where id = x'997901'"
```

#### Test results
**13.0-SNAPSHOT:**
```
$ mysql customer -e "select shard from customer.binary where id = x'7901'"
+-------+
| shard |
+-------+
| -80   |
+-------+
$ mysql customer -e "select shard from customer.binary where id = x'997901'"
+-------+
| shard |
+-------+
| -80   |
+-------+
```

**11.0.2:**
```
$ mysql customer -e "select shard from customer.binary where id = x'7901'"
+-------+
| shard |
+-------+
| -80   |
+-------+
$ mysql customer -e "select shard from customer.binary where id = x'997901'"
+-------+
| shard |
+-------+
| 80-   |
+-------+
```

#### Binary version
Example:

```sh
$ vtgate --version
[...]

PR #9145, opened 2021-11-04
Ensure that hex query predicates are normalized for planner cache

mattlord:
## Description

We were not normalizing hex values of either format:
- `0x01`
- `x'73757265'`

This had a major performance impact in workloads where hex query predicates are used in common queries as we could not leverage the query planner cache. This work normalizes them.

> Note: the only explicitly non-normalized values are now bit types of the format `b'1'`:
```
mysql> select b'011101110111010101110100' /* wut */;
+----------------------------------------------------------+
| b'011101110111010101110100'                              |
+----------------------------------------------------------+
| 0x777574                                                 |
+----------------------------------------------------------+
1 row in set (0.00 sec)
```

## Related Issue(s)
https://github.com/vitessio/vitess/issues/9101
https://github.com/vitessio/vitess/pull/9118

## Checklist
- [x] Should this PR be backported? Not beyond 12.0 IMO
- [x] Tests were added
- [x] Documentation is not required

PR #9277, opened 2021-11-23
Use decoded hex string when calculating the keyspace ID

mattlord:
## Description
This is a follow-up to https://github.com/vitessio/vitess/pull/9118

There we [normalized the incoming `x'A1'` or `0xA1` string literal token](https://github.com/planetscale/vitess/blob/94c55f08328a6d13a1edec46b8a188e0caae24cb/go/vt/sqlparser/normalizer.go#L201-L209) that represents binary data as hex so that we can pass it on to MySQL as a BindVar.

However, this means that when calculating the Keyspace ID in the vindex functions we need to decode the value again ([as we do when parsing the incoming query](https://github.com/vitessio/vitess/blob/main/go/vt/sqlparser/analyzer.go#L472-L477)) so that we're doing the keyspace range check using a decoded representation of the bytes that end up being stored in the column in MySQL.

This is accomplished [here](https://github.com/vitessio/vitess/pull/9277/commits/ef89fb53a89884b418104ebc27f126f37fd1ef47) and [here](https://github.com/vitessio/vitess/pull/9277/commits/55dcac1036806ada44d655f1e59157c08d6d2ac7) by properly managing the HexVal and HexNum BindVars:
  - The "raw" value (`Value.val` / `Value.Raw()` and its variants) is "internal" and represents what was seen in the SQL query (e.g. `x'997902'` or `0x997902`)
  - The ToBytes value (`Value.ToBytes()`) returns the bytes that mysql would return when querying the column (e.g. `�y`).
    For example:
    ```
    $ mysql -u root -h 127.0.0.1 --binary-as-hex=false mydb -e "drop table if exists varbt; create table varbt (c1 varbinary(100)); insert into varbt values (x'997902'),(0x997902); select c1, hex(c1) from varbt"
    +------+---------+
    | c1   | hex(c1) |
    +------+---------+
    | �y   | 997902  |
    | �y   | 997902  |
    +------+---------+
    ```

We then see the exact same [results for the vindex functions using the `x'A1'` format seen before normalizing hex values](https://github.com/vitessio/vitess/pull/9277#issuecomment-976209654).

🆕 With this PR we also add support for using the `0xA1` format with vindexes which did NOT work before:
  - #### 11.0:
    ```
    $ mysql customer -e "select id, hex_keyspace_id, shard from customer.binary where id = 0x997901"
    ERROR 1105 (HY000) at line 1: unsupported: where clause for vindex function must be of the form id = <val> (rhs is not a value)

    $ mysql customer -e "select id, hex_keyspace_id, shard from customer.binary_md5 where id = 0x997901"
    ERROR 1105 (HY000) at line 1: unsupported: where clause for vindex function must be of the form id = <val> (rhs is not a value)

    $ mysql customer -e "select id, hex_keyspace_id, shard from customer.xxhash where id = 0x997901"
    ERROR 1105 (HY000) at line 1: unsupported: where clause for vindex function must be of the form id = <val> (rhs is not a value)
    ```
  - #### This branch:
    ```
    $ mysql customer -e "select id, hex_keyspace_id, shard from customer.binary where id = 0x997901"
    +------+-----------------+-------+
    | id   | hex_keyspace_id | shard |
    +------+-----------------+-------+
    | �y  | 997901          | 80-   |
    +------+-----------------+-------+

    $ mysql customer -e "select id, hex_keyspace_id, shard from customer.binary_md5 where id = 0x997901"
    +------+----------------------------------+-------+
    | id   | hex_keyspace_id                  | shard |
    +------+----------------------------------+-------+
    | �y  | 2d2da332b59f56aeb235ecb36d639440 | -80   |
    +------+----------------------------------+-------+

    $ mysql customer -e "select id, hex_keyspace_id, shard from customer.xxhash where id = 0x997901"
    +------+------------------+-------+
    | id   | hex_keyspace_id  | shard |
    +------+------------------+-------+
    | �y  | ac29dcaa4a0b4527 | 80-   |
    +------+------------------+-------+
    ```

So now for vindexes we are properly treating `x'A1'` and `0xA1` as equivalent. This is proper because while the string literal in the SQL query differs the same bytes are stored in MySQL for both. You can see a complete user test [here](https://github.com/vitessio/vitess/pull/9277#issuecomment-978168444).

## Related Issue(s)
Fixes: https://github.com/vitessio/vitess/issues/9275

## Checklist
- [x] This PR should be backported to 12.0 for the next patch release or we need to revert https://github.com/vitessio/vitess/pull/9145
- [x] Tests were added [here](https://github.com/vitessio/vitess/pull/9277/commits/d63bb050c57e51e03992268ae1a683e265337d00)

### DIFF
--- a/go/sqltypes/value.go
+++ b/go/sqltypes/value.go
@@ -251,6 +251,10 @@ func (v Value) ToBytes() ([]byte, error) {
 		dv, err := v.decodeHexVal()
 		return dv, err
 	}
+	if v.typ == HexNum {
+		dv, err := v.decodeHexNum()
+		return dv, err
+	}
 	return v.val, nil
 }
 
@@ -487,6 +491,22 @@ func (v *Value) decodeHexVal() ([]byte, error) {
 		return nil, vterrors.Errorf(vtrpc.Code_INVALID_ARGUMENT, "invalid hex value: %v", v.val)
 	}
 	hexBytes := v.val[2 : len(v.val)-1]
+	decodedHexBytes, err := hex.DecodeString(string(hexBytes))
+	if err != nil {
+		return nil, err
+	}
+	return decodedHexBytes, nil
+}
+
+// decodeHexNum decodes the SQL hex value of the form 0xA1 into a byte
+// array matching what MySQL would return when querying the column where
+// an INSERT was performed with 0xA1 having been specified as a value
+func (v *Value) decodeHexNum() ([]byte, error) {
+	match, err := regexp.Match("^0x.*$", v.val)
+	if !match || err != nil {
+		return nil, vterrors.Errorf(vtrpc.Code_INVALID_ARGUMENT, "invalid hex number: %v", v.val)
+	}
+	hexBytes := v.val[2:]
 	decodedHexBytes, err := hex.DecodeString(string(hexBytes))
 	if err != nil {
 		return nil, err
--- a/go/sqltypes/value.go
+++ b/go/sqltypes/value.go
@@ -30,9 +30,9 @@ import (
 	"vitess.io/vitess/go/bytes2"
 	"vitess.io/vitess/go/hack"
 
-	"vitess.io/vitess/go/vt/log"
 	querypb "vitess.io/vitess/go/vt/proto/query"
 	"vitess.io/vitess/go/vt/proto/vtrpc"
+	vtrpcpb "vitess.io/vitess/go/vt/proto/vtrpc"
 	"vitess.io/vitess/go/vt/vterrors"
 )
 
@@ -241,20 +241,17 @@ func (v Value) RawStr() string {
 
 // ToBytes returns the value as MySQL would return it as []byte.
 // In contrast, Raw returns the internal representation of the Value, which may not
-// match MySQL's representation for newer types.
-// If the value is not convertible like in the case of Expression, it returns nil.
-func (v Value) ToBytes() []byte {
+// match MySQL's representation for hex encoded binary data or newer types.
+// If the value is not convertible like in the case of Expression, it returns an error.
+func (v Value) ToBytes() ([]byte, error) {
 	if v.typ == Expression {
-		return nil
+		return nil, vterrors.New(vtrpcpb.Code_INVALID_ARGUMENT, "expression cannot be converted to bytes")
 	}
 	if v.typ == HexVal {
 		dv, err := v.decodeHexVal()
-		if err != nil {
-			log.Errorf("Unexpected error seen when returning MySQL representation of SQL Hex value: %v", err)
-		}
-		return dv
+		return dv, err
 	}
-	return v.val
+	return v.val, nil
 }
 
 // Len returns the length.
--- a/go/sqltypes/value.go
+++ b/go/sqltypes/value.go
@@ -19,16 +19,21 @@ package sqltypes
 
 import (
 	"encoding/base64"
+	"encoding/hex"
 	"encoding/json"
 	"errors"
 	"fmt"
+	"regexp"
 	"strconv"
 	"strings"
 
 	"vitess.io/vitess/go/bytes2"
 	"vitess.io/vitess/go/hack"
 
+	"vitess.io/vitess/go/vt/log"
 	querypb "vitess.io/vitess/go/vt/proto/query"
+	"vitess.io/vitess/go/vt/proto/vtrpc"
+	"vitess.io/vitess/go/vt/vterrors"
 )
 
 var (
@@ -241,6 +246,13 @@ func (v Value) RawStr() string {
 func (v Value) ToBytes() []byte {
 	if v.typ == Expression {
 		return nil
+	}
+	if v.typ == HexVal {
+		dv, err := v.decodeHexVal()
+		if err != nil {
+			log.Errorf("Unexpected error seen when returning MySQL representation of SQL Hex value: %v", err)
+		}
+		return dv
 	}
 	return v.val
 }
@@ -467,6 +479,22 @@ func (v *Value) UnmarshalJSON(b []byte) error {
 	}
 	*v, err = InterfaceToValue(val)
 	return err
+}
+
+// decodeHexVal decodes the SQL hex value of the form x'A1' into a byte
+// array matching what MySQL would return when querying the column where
+// an INSERT was performed with x'A1' having been specified as a value
+func (v *Value) decodeHexVal() ([]byte, error) {
+	match, err := regexp.Match("^x'.*'$", v.val)
+	if !match || err != nil {
+		return nil, vterrors.Errorf(vtrpc.Code_INVALID_ARGUMENT, "invalid hex value: %v", v.val)
+	}
+	hexBytes := v.val[2 : len(v.val)-1]
+	decodedHexBytes, err := hex.DecodeString(string(hexBytes))
+	if err != nil {
+		return nil, err
+	}
+	return decodedHexBytes, nil
 }
 
 func encodeBytesSQL(val []byte, b BinWriter) {
--- a/go/sqltypes/value_test.go
+++ b/go/sqltypes/value_test.go
@@ -22,6 +22,8 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/stretchr/testify/require"
+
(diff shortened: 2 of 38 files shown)


mattlord:
Example vindex function results before normalizing hex values:
```
$ mysql customer -e "select id, hex_keyspace_id, shard from customer.xxhash where id = x'997901'"
+------+------------------+-------+
| id   | hex_keyspace_id  | shard |
+------+------------------+-------+
| �y  | ac29dcaa4a0b4527 | 80-   |
+------+------------------+-------+

$ mysql customer -e "select id, hex_keyspace_id, shard from customer.binary where id = x'997901'"
+------+-----------------+-------+
| id   | hex_keyspace_id | shard |
+------+-----------------+-------+
| �y  | 997901          | 80-   |
+------+-----------------+-------+

$ mysql customer -e "select id, hex_keyspace_id, shard from customer.binary_md5 where id = x'997901'"
+------+----------------------------------+-------+
| id   | hex_keyspace_id                  | shard |
+------+----------------------------------+-------+
| �y  | 2d2da332b59f56aeb235ecb36d639440 | -80   |
+------+----------------------------------+-------+

$ mysql customer -e "select id, hex_keyspace_id, shard from customer.xxhash where id = x'997901'"
+------+------------------+-------+
| id   | hex_keyspace_id  | shard |
+------+------------------+-------+
| �y  | ac29dcaa4a0b4527 | 80-   |
+------+------------------+-------+
```

go/sqltypes/value.go
```diff
@@ -235,6 +237,13 @@ func (v Value) ToBytes() []byte {
 	if v.typ == Expression {
 		return nil
 	}
+	if v.typ == HexVal {
+		dv, err := v.decodeHexVal()
+		if err != nil {
+			log.Errorf("Unexpected error seen when returning MySQL representation of SQL Hex value: %v", err)
+		}
+		return dv
```
harshit-gangal, line 245: if not able to decode hex val, then should it not return error then returning dv (which will contain nil value).

go/sqltypes/value.go
```diff
@@ -462,6 +471,21 @@ func (v *Value) UnmarshalJSON(b []byte) error {
 	return err
 }
 
+func (v *Value) decodeHexVal() ([]byte, error) {
+	if len(v.val) == 0 {
+		return nil, fmt.Errorf("empty hex value")
+	}
+	if len(v.val) < 3 || v.val[0] != 'x' || v.val[1] != '\'' || v.val[len(v.val)-1] != '\'' {
+		return nil, fmt.Errorf("invalid hex value: %v", v.val)
```
harshit-gangal, line 482: should use vterrors.New / vterrors.Errorf instead of fmt.Errorf

go/sqltypes/value.go
```diff
@@ -462,6 +471,24 @@ func (v *Value) UnmarshalJSON(b []byte) error {
...
 
+// decodeHexVal decodes the SQL hex value of the form x'A1' into a byte
+// array matching what MySQL would return when querying the column where
+// an INSERT was performed with x'A1' having been specified as a value
+func (v *Value) decodeHexVal() ([]byte, error) {
+	if len(v.val) == 0 {
+		return nil, fmt.Errorf("empty hex value")
+	}
+	if len(v.val) < 3 || v.val[0] != 'x' || v.val[1] != '\'' || v.val[len(v.val)-1] != '\'' {
+		return nil, fmt.Errorf("invalid hex value: %v", v.val)
```
harshit-gangal, line 482: nit: can use regex pattern matching here.

go/vt/sqlparser/ast_funcs.go
```diff
@@ -490,7 +490,7 @@ func (node *Literal) encodeHexValToMySQLQueryFormat() ([]byte, error) {
 	}
 
 	// Let's make this idempotent in case it's called more than once
-	if nb[0] == 'x' && nb[1] == '0' && nb[len(nb)-1] == '\'' {
+	if len(nb) > 2 && nb[0] == 'x' && nb[1] == '\'' && nb[len(nb)-1] == '\'' {
```
harshit-gangal, line 493: nit: can use regex pattern matching.

go/sqltypes/value.go
mattlord: Yeah, I feel like we should... but that method never returned an error so for now I just logged it. If you think it's OK to modify all the `Value.ToBytes()` call sites to handle errors then I'll do that.

go/vt/sqlparser/ast_funcs.go
mattlord: I thought that might be less efficient but it's certainly simpler so I'll switch to that.

go/sqltypes/value.go
mattlord: I think/hope this is what you mean/wanted: https://github.com/vitessio/vitess/pull/9277/commits/99d065a3f6e2c8d46f6672c37f604d6d4686d079

If not, at least we can easily revert the single commit.

frouioui (approved):
It looks good to me! :)

Before merging, let's see if https://github.com/vitessio/vitess/pull/9277/commits/61cb238ff304d7d6e7ef1ebafb497682d5086ce6 resolves @harshit-gangal 's comment

go/sqltypes/value.go
```diff
@@ -238,11 +243,15 @@ func (v Value) RawStr() string {
 // In contrast, Raw returns the internal representation of the Value, which may not
 // match MySQL's representation for newer types.
 // If the value is not convertible like in the case of Expression, it returns nil.
-func (v Value) ToBytes() []byte {
+func (v Value) ToBytes() ([]byte, error) {
```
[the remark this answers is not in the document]
mattlord: Good catch! Will do that now

go/sqltypes/value_test.go
```diff
@@ -412,7 +418,12 @@ func TestToBytesAndString(t *testing.T) {
 	}
 
 	tv := TestValue(Expression, "aa")
-	if b := tv.ToBytes(); b != nil {
+	tvBytes, err := tv.ToBytes()
+	expectedErr := vterrors.New(vtrpcpb.Code_INVALID_ARGUMENT, "expression cannot be converted to bytes")
+	if err.Error() != expectedErr.Error() {
+		t.Errorf("Expected: %v, got: %v", expectedErr, err)
+	}
```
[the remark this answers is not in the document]
mattlord: Thanks! Will update this now

go/sqltypes/value_test.go
```diff
@@ -403,7 +405,11 @@ func TestToBytesAndString(t *testing.T) {
 		TestValue(Int64, "1"),
 		TestValue(Int64, "12"),
 	} {
-		if b := v.ToBytes(); !bytes.Equal(b, v.Raw()) {
+		vBytes, err := v.ToBytes()
+		if err != nil {
+			t.Error(err)
+		}
```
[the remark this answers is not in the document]
mattlord: Done

mattlord:
Here's a full user test:
```
make docker_local && ./docker/local/run.sh

./201_customer_tablets.sh ; ./202_move_tables.sh ; sleep 10; ./203_switch_reads.sh ; ./204_switch_writes.sh ; ./205_clean_commerce.sh ; ./301_customer_sharded.sh ; ./302_new_shards.sh ; ./303_reshard.sh ; sleep 10; ./304_switch_reads.sh ; ./305_switch_writes.sh ; ./306_down_shard_0.sh

vtctlclient GetVSchema customer > /tmp/customer.json

vtctlclient ApplyVSchema -vschema "$(jq '. | .vindexes += {"binary":{"type":"binary"}} | .tables += {"btest":{"columnVindexes":[{"column":"id","name":"binary"}]}}' < /tmp/customer.json)" customer

mysql customer -e "insert into btest (id, name) values (0xF901, 'test'),(0x1901, 'test'),(x'1801', 'test'),(x'F801', 'test')"

mysql customer -e "select hex(id) from btest"
mysql customer/-80 -e "select hex(id) from btest"
mysql customer/80- -e "select hex(id) from btest"
```

harshit-gangal (approved):
LGTM

deepthi:
Very well-written description. Thank you for taking the time to write it up along with the code changes.

PR #9281, opened 2021-11-23
Add Tuples and InOp implementation to EvalEngine

GuptaManan100:
## Description
As a long term goal, we want to replace the usage of `PlanValue` everywhere. To accomplish this change, we need to add an implementation to `evalengine.Expr` which can store tuples and evaluate them. This PR accomplishes that change. It introduces a new implementation of `evalengine.Expr` which stores multiple values and also changes the `EvalResult` struct to be able to store multiple results for intermediate steps.

## Related Issue(s)

- #7280 

## Checklist

### DIFF
--- a/go/vt/sqlparser/parse_test.go
+++ b/go/vt/sqlparser/parse_test.go
@@ -3255,6 +3255,27 @@ func TestCreateTable(t *testing.T) {
 	}
 }
 
+func TestOne(t *testing.T) {
+	testOne := struct {
+		input, output string
+	}{
+		input:  "",
+		output: "",
+	}
+	if testOne.input == "" {
+		return
+	}
+	sql := strings.TrimSpace(testOne.input)
+	tree, err := Parse(sql)
+	require.NoError(t, err)
+	got := String(tree)
+	expected := testOne.output
+	if expected == "" {
+		expected = sql
+	}
+	require.Equal(t, expected, got)
+}
+
 func TestCreateTableLike(t *testing.T) {
 	normal := "create table a like b"
 	testCases := []struct {
--- a/go/vt/vtgate/evalengine/comparisons.go
+++ b/go/vt/vtgate/evalengine/comparisons.go
@@ -183,24 +183,6 @@ func foldSingleLenTuples(val EvalResult) EvalResult {
 	return val
 }
 
-func compareTuples(lVal EvalResult, rVal EvalResult) (int, bool, error) {
-	if len(lVal.tupleResults) != len(rVal.tupleResults) {
-		return 0, false, vterrors.NewErrorf(vtrpcpb.Code_INVALID_ARGUMENT, vterrors.OperandColumns, "Operand should contain %d column(s)", len(lVal.tupleResults))
-	}
-	hasSeenNull := false
-	for idx, lResult := range lVal.tupleResults {
-		rResult := rVal.tupleResults[idx]
-		res, isNull, err := nullSafeExecuteComparison(lResult, rResult)
-		if isNull {
-			hasSeenNull = true
-		}
-		if res != 0 || err != nil {
-			return res, false, err
-		}
-	}
-	return 0, hasSeenNull, nil
-}
-
 // Evaluate implements the Expr interface
 func (c *ComparisonExpr) Evaluate(env ExpressionEnv) (EvalResult, error) {
 	if c.Op == nil {
--- a/go/vt/vtgate/evalengine/comparisons.go
+++ b/go/vt/vtgate/evalengine/comparisons.go
@@ -423,12 +423,15 @@ func (i *InOp) String() string {
 
 // Evaluate implements the ComparisonOp interface
 func (n *NotInOp) Evaluate(left, right EvalResult) (EvalResult, error) {
-	panic("implement me")
+	res, err := (&InOp{}).Evaluate(left, right)
+	res.ival = 1 - res.ival
+	return res, err
 }
 
 // IsTrue implements the ComparisonOp interface
 func (n *NotInOp) IsTrue(left, right EvalResult) (bool, error) {
-	return false, nil
+	res, err := (&InOp{}).IsTrue(left, right)
+	return !res, err
 }
 
 // Type implements the ComparisonOp interface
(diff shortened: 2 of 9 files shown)


go/vt/vtgate/evalengine/expressions.go
```diff
@@ -67,14 +69,41 @@ type (
 		Offset    int
 		Collation collations.ID
 	}
+	Tuple []Expr
 )
 
+func (t Tuple) Evaluate(env ExpressionEnv) (EvalResult, error) {
```
systay, line 75: don't we want comments on public methods? these are all interface implementations, so I don't personally miss them, but the linter should have complained, no?

go/vt/vtgate/evalengine/comparisons.go
```diff
@@ -28,7 +28,6 @@ type (
 	// when evaluating the whole comparison
 	ComparisonOp interface {
 		Evaluate(left, right EvalResult) (EvalResult, error)
-		IsTrue(left, right EvalResult) (bool, error)
```
frouioui, line 31: Why removing `IsTrue`?

go/vt/vtgate/evalengine/comparisons.go
```diff
@@ -126,27 +125,43 @@ func evalResultsAreDateAndNumeric(l, r EvalResult) bool {
 
 // For more details on comparison expression evaluation and type conversion:
 // 		- https://dev.mysql.com/doc/refman/8.0/en/type-conversion.html
-func executeComparison(lVal, rVal EvalResult) (int, error) {
+func nullSafeExecuteComparison(lVal, rVal EvalResult) (comp int, isNull bool, err error) {
```
frouioui, line 128: I think we should keep this method as it was before, and add another method (`nullSafeExecuteComparison`) that calls (`executeComparison`). The new method (`nullSafeExecuteComparison`) will encapsulate the logic that was previously in `ComparisonExpr.Evaluate`:
```go
if hasNullEvalResult(lVal, rVal) {
	return 0, true, nil
}
```

This will allow us to have a simpler function for comparisons.

go/vt/vtgate/evalengine/expressions.go
GuptaManan100: For some reason it didn't complain. I'll fix this

go/vt/vtgate/evalengine/comparisons.go
```diff
@@ -156,8 +171,34 @@ func executeComparison(lVal, rVal EvalResult) (int, error) {
...
+		res, isNull, err := nullSafeExecuteComparison(lResult, rResult)
+		if isNull {
+			hasSeenNull = true
+		}
+		if res != 0 || err != nil {
+			return res, false, err
+		}
+	}
+	return 0, hasSeenNull, nil
 }
```
frouioui, line 202: let's move this to `evalengine.go` with the other compare functions

go/vt/vtgate/evalengine/expressions.go
```diff
@@ -67,14 +69,44 @@ type (
...
+}
+
+// String implements the Expr interface
+func (t Tuple) String() string {
+	var stringSlice []string
+	for _, expr := range t {
+		stringSlice = append(stringSlice, expr.String())
+	}
+	return "(" + strings.Join(stringSlice, ",") + ")"
+}
```
frouioui, line 101: small nit, but can we also move this down in the file, where the other implementations are

go/vt/vtgate/evalengine/comparisons.go
GuptaManan100: The reason we had it was that it had lesser memory pressure in the sense that EvalResult was not needed to be created, but once we added comparisons for tuples, we needed to do null safe comparisons on each step leading to the IsTrue calling Evaluate instead of it being the other way around. So there is nothing new that it offers. If we want to use this someplace, then we can add a function for getting a boolean value from the output of the evaluate function

go/vt/vtgate/evalengine/comparisons.go
GuptaManan100: After adding comparisons for tuples, we need nullSafeComparisons in all the places. Since there is no place where we would call the executeComparison from we haven't extracted out yet.

go/vt/vtgate/evalengine/expressions.go
GuptaManan100: Sure

go/vt/vtgate/evalengine/comparisons.go
GuptaManan100: Sure

frouioui (approved):

vmg (approved):
Looking good! I'm confused about collations propagation for comparisons where the right side is a tuple, but that's something I gotta figure out myself in my branch. :ok_hand:

go/vt/sqlparser/parse_test.go
```diff
@@ -3255,6 +3255,27 @@ func TestCreateTable(t *testing.T) {
...
+		input, output string
+	}{
+		input:  "",
+		output: "",
+	}
+	if testOne.input == "" {
+		return
+	}
+	sql := strings.TrimSpace(testOne.input)
+	tree, err := Parse(sql)
```
harshit-gangal, line 3269: looks like the test is skipped with the if condition.

go/vt/vtgate/evalengine/comparisons_test.go
```diff
@@ -894,3 +893,79 @@ func TestCompareStrings(t *testing.T) {
...
 }
+
+// TestInOp tests the In operator comparisons
+func TestInOp(t *testing.T) {
+	tests := []testCase{
+		{
+			name: "integer In tuple",
+			v1:   NewLiteralInt(52), v2: Tuple{NewLiteralInt(52), NewLiteralInt(54)},
+			out: &T,
+			op:  &InOp{},
```
harshit-gangal, line 904: nit: add cases for 
1. single value 
2. only null value
3. no value

in tuple

harshit-gangal (approved):

go/vt/sqlparser/parse_test.go
GuptaManan100: Yes that is the intention. It is supposed to be used for testing one query while working and shouldn't have any other tests. Kinda like how we have testOne in the plan_tests

go/vt/vtgate/evalengine/comparisons_test.go
GuptaManan100: Fixed

harshit-gangal (approved):

PR #9286, opened 2021-11-24
evalengine: Coercion, LIKE, IN optimization

vmg:
## Description

I've been working on the eval engine this week. Fun times!

I've already shown this code to @systay @frouioui and @GuptaManan100, but here's the PR. It's split in individual commits that should be easier to review.

## Checklist


frouioui (approved):
This is great!

GuptaManan100 (approved):
Love It! Especially the IN operator optimisation

Issue #9002, opened 2021-10-15
MoveTables doesn't migrate sequence in VSchema definition


derekperkins:
#### Overview of the Issue

When migrating from a sharded keyspace to an unsharded keyspace, my expectation was that the sequence would continue to be applied afterwards, but it wasn't in the vschema definition. I quickly re-added it, so I didn't do any testing to see how it actually behaved.

Before MoveTables (sharded keyspace)

```javascript
    "semetrics_grepwords_pipeline": {
      "autoIncrement": {
        "column": "pipeline_id",
        "sequence": "semetrics_grepwords_pipeline_seq"
      }
    },
```

After MoveTables (unsharded keyspace)

```javascript
    "semetrics_grepwords_pipeline": {}
```

#### Operating system and Environment details

vttablet: 11.0.1
vtctld: 11.0.1
vtgate: 10.0.0

DeathBorn:
I think for Movetables, one needs to provide vindexes or/and sequence definitions for target. In v8, when moving unsharded to sharded, I had to provide both.

Another note, when Movetables and  switchwrites is performed on sequence table, tables which use that sequence table do not  respect routing rules at all. Maybe sequence tables should not be migrated in that way.

rohit-nayak-ps:
Closed via https://github.com/vitessio/vitess/pull/9288

PR #9288, opened 2021-11-26
MoveTables: update vschema while moving tables with autoincrement from sharded to unsharded

rohit-nayak-ps:
## Description

When a sharded table with autoincrement  is moved using MoveTables to an unsharded keyspace, currently the related sequence information is not carried over into the target (unsharded) keyspace's vschema.

This PR fixes that by copying over the sequence info from the source schema for that table. This is verified by adding this use-case to the v2 end-to-end test.

## Related Issue(s)
#9002 

## Checklist

### DIFF
--- a/go/test/endtoend/vreplication/resharding_workflows_v2_test.go
+++ b/go/test/endtoend/vreplication/resharding_workflows_v2_test.go
@@ -22,6 +22,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/tidwall/gjson"
+
 	"github.com/stretchr/testify/assert"
 
 	"vitess.io/vitess/go/vt/log"
@@ -246,24 +248,43 @@ func TestBasicV2Workflows(t *testing.T) {
 	log.Flush()
 }
 
+const workflowStartTimeout = 5 * time.Second
+
 func waitForWorkflowToStart(t *testing.T, ksWorkflow string) {
 	done := false
-	ticker := time.NewTicker(10 * time.Millisecond)
+	ticker := time.NewTicker(100 * time.Millisecond)
+	timer := time.NewTimer(workflowStartTimeout)
 	log.Infof("Waiting for workflow %s to start", ksWorkflow)
 	for {
 		select {
 		case <-ticker.C:
 			if done {
+				log.Infof("Workflow %s has started", ksWorkflow)
 				return
 			}
 			output, err := vc.VtctlClient.ExecuteCommandWithOutput("Workflow", ksWorkflow, "show")
 			require.NoError(t, err)
-			if strings.Contains(output, "\"State\": \"Running\"") {
-				done = true
-				log.Infof("Workflow %s has started", ksWorkflow)
-			}
-		case <-time.After(5 * time.Second):
-			require.FailNow(t, "workflow %s not yet started", ksWorkflow)
+			done = true
+			state := ""
+			result := gjson.Get(output, "ShardStatuses")
+			result.ForEach(func(tabletId, tabletStreams gjson.Result) bool { // for each participating tablet
+				tabletStreams.ForEach(func(streamId, streamInfos gjson.Result) bool { // for each stream
+					if streamId.String() == "PrimaryReplicationStatuses" {
+						streamInfos.ForEach(func(attributeKey, attributeValue gjson.Result) bool { // for each attribute in the stream
+							state = attributeValue.Get("State").String()
+							if state != "Running" {
+								done = false // we need to wait for all streams to start
+							}
+							return true
+						})
+					}
+					return true
+				})
+				return true
+			})
+
+		case <-timer.C:
+			require.FailNowf(t, "workflow %s not yet started", ksWorkflow)
 		}
 	}
 }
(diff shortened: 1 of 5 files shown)


go/test/endtoend/vreplication/resharding_workflows_v2_test.go
```diff
@@ -244,6 +246,73 @@ func TestBasicV2Workflows(t *testing.T) {
...
+*/
+func testVSchemaForSequenceAfterMoveTables(t *testing.T) {
+	// at this point the unsharded product and sharded customer keyspaces are created by previous tests
+
+	// use MoveTables to move customer2 from product to customer using
+	currentWorkflowType = wrangler.MoveTablesWorkflow
+	err := tstWorkflowExec(t, defaultCellName, "wf2", sourceKs, targetKs,
+		"customer2", workflowActionCreate, "", "", "")
+	require.NoError(t, err)
+	time.Sleep(1 * time.Second)
```
mattlord, line 261: I wonder if this will introduce some flakiness? In my experience the new Workflow Create call will return an error because it does not start within the default wait timeout in some percentage of cases. In these 2 lines, we just want to confirm that it has successfully started, right? If so, can't we check the Show output instead to reduce flakiness?

go/test/endtoend/vreplication/resharding_workflows_v2_test.go
```diff
@@ -244,6 +246,73 @@ func TestBasicV2Workflows(t *testing.T) {
...
+	// check that customer2 still has the sequence tag
+	output, err = vc.VtctlClient.ExecuteCommandWithOutput("GetVSchema", "product")
+	require.NoError(t, err)
+	assert.Contains(t, output, "\"sequence\": \"customer_seq2\"", "customer2 still found in keyspace product")
+
+	// ensure sequence is available to vtgate
+	for i := 0; i < num; i++ {
+		execVtgateQuery(t, vtgateConn, "product", "insert into customer2(name) values('a')")
+	}
+	validateCount(t, vtgateConn, "product", "customer2", 3+num+num)
```
mattlord, line 313: Is it worth checking the values to be sure they are actually incrementing integers as we expect? Not sure, but came to mind

go/vt/wrangler/materializer.go
```diff
@@ -65,6 +65,37 @@ const (
...
+	}
+
+	if copyAttributes { // if source keyspace is provided, copy over the sequence info.
+		srcVSchema, err := wr.ts.GetVSchema(ctx, sourceKeyspace)
+		if err != nil {
+			return err
+		}
+		if srcVSchema == nil {
+			return fmt.Errorf("no vschema found for source keyspace %s", sourceKeyspace)
+		}
```
mattlord, line 87: I wonder if this has to be an error? If you're using shard targeting everywhere then you probably don't have a vschema and there would be no sequence info to copy either. 

I'm not sure if this method would even be called in that case though as this is only called if `if !vschema.Sharded`

mattlord (approved):
I had a nit about test flakiness and a couple of questions but nothing blocking IMO. Thanks!

deepthi (commented):
Matt's review looks good. You may merge once it is addressed.

go/test/endtoend/vreplication/resharding_workflows_v2_test.go
```diff
@@ -244,6 +246,73 @@ func TestBasicV2Workflows(t *testing.T) {
...
+	for i := 0; i < num; i++ {
+		execVtgateQuery(t, vtgateConn, "customer", "insert into customer2(name) values('a')")
+	}
+	validateCount(t, vtgateConn, "customer", "customer2", 3+num)
+
+	// use MoveTables to move customer2 back to product. Note that now the table has an associated sequence
+	err = tstWorkflowExec(t, defaultCellName, "wf3", targetKs, sourceKs,
+		"customer2", workflowActionCreate, "", "", "")
+	require.NoError(t, err)
+	time.Sleep(1 * time.Second)
```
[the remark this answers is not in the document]
rohit-nayak-ps: Good idea about removing the sleeps, got too lazy. Added an explicit check for workflow to start.

go/vt/wrangler/materializer.go
rohit-nayak-ps: Removed the check.

go/test/endtoend/vreplication/resharding_workflows_v2_test.go
rohit-nayak-ps: Added a check to ensure that the sequence values are being correctly used.

go/test/endtoend/vreplication/resharding_workflows_v2_test.go
```diff
@@ -246,6 +246,28 @@ func TestBasicV2Workflows(t *testing.T) {
...
+	log.Infof("Waiting for workflow %s to start", ksWorkflow)
+	for {
+		select {
+		case <-ticker.C:
+			if done {
+				return
+			}
+			output, err := vc.VtctlClient.ExecuteCommandWithOutput("Workflow", ksWorkflow, "show")
+			require.NoError(t, err)
+			if strings.Contains(output, "\"State\": \"Running\"") {
```
mattlord, line 261: Minor nit, maybe better to use a RegEx or JSON parsing so that we are checking for what we really care about while resilient to whitespace and other minor format changes? Full JSON parsing is probably overkill, maybe RegEx is too :-) 

Feel free to ignore.

mattlord (approved):
Nice work! I had two minor nits/questions, but you can ignore them. Thanks!

PR #9292, opened 2021-11-29
Backoff on inconsistent state

sonne5:
## Description

There are cases after a network partition, different member will report different state for the same node. For example, we observe a node reported itself as "OFFLINE" whereas the other node reported the same node as "ONLINE". The only solution we have is to rebootstrap the group. 

For example, during an AZ outage between 19:11 to 19:21 UTC, around the same time there seems to be a network blip that leads to a lost of quorum in the group:

2021-10-14T19:12:19.552500Z 0 [ERROR] [MY-011496] [Repl] Plugin group_replication reported: 'This server is not able to reach a majority of members in the group. This server will now block all updates. The server will remain blocked for the next 5 seconds. Unless contact with the majority is restored, after this time the member will error out and leave the group. It is possible to use group_replication_force_members to force a new group membership.'
There is something fishy happened inside mysql that makes different node reported different group view:

on primary:
```
mysql> select * from performance_schema.replication_group_members;
+---------------------------+--------------------------------------+---------------------------------------------------------+-------------+--------------+-------------+----------------+
| CHANNEL_NAME              | MEMBER_ID                             | MEMBER_ROLE | MEMBER_VERSION |
+---------------------------+--------------------------------------+---------------------------------------------------------+-------------+--------------+-------------+----------------+
| group_replication_applier | 0d2f5008-2c71-11ec-9588-0208f3fcc021  | ONLINE       | PRIMARY     | 8.0.22         |
| group_replication_applier | 4c13646c-2c70-11ec-991a-06731dd33837  | ONLINE       | SECONDARY   | 8.0.22         |
| group_replication_applier | dcee6acb-2c70-11ec-ae59-0a8fcea3b2db  | ONLINE       | SECONDARY   | 8.0.22         |
+---------------------------+--------------------------------------+---------------------------------------------------------+-------------+--------------+-------------+----------------+
```
on secondary that appears to be online from above:
```
mysql> select * from performance_schema.replication_group_members;
+---------------------------+--------------------------------------+---------------------------------------------------------+-------------+--------------+-------------+----------------+
| CHANNEL_NAME              | MEMBER_ID                             | MEMBER_STATE | MEMBER_ROLE | MEMBER_VERSION |
+---------------------------+--------------------------------------+---------------------------------------------------------+-------------+--------------+-------------+----------------+
| group_replication_applier | dcee6acb-2c70-11ec-ae59-0a8fcea3b2db | OFFLINE      |             |                |
+---------------------------+--------------------------------------+---------------------------------------------------------+-------------+--------------+-------------+----------------+
1 row in set (0.00 sec)
```
As a result, the group is stuck. The only way to get away this situation is to stop the group and rebootstrap.

This PR adds a check in VTGR to detect the situation when a node is reported differently in a group (with one of them is "ONLINE"), then VTGR will trigger to backoff logic to rebootstrap the group.

To make it more conservative, we integrated with the Vitess's heartbeat table - so that the logic is triggered after certain threshold.

## Related Issue(s)

See above

## Checklist

### DIFF
--- a/go/vt/vtgr/controller/group.go
+++ b/go/vt/vtgr/controller/group.go
@@ -34,7 +34,7 @@ var (
 	groupOnlineSize = stats.NewGaugesWithMultiLabels("MysqlGroupOnlineSize", "Online MySQL server in the group", []string{"Keyspace", "Shard"})
 	isLostQuorum    = stats.NewGaugesWithMultiLabels("MysqlGroupLostQuorum", "If MySQL group lost quorum", []string{"Keyspace", "Shard"})
 
-	heartbeatThreshold = flag.Int("group_heartbeat_threshold_sec", 0, "VTGR will trigger backoff on inconsistent state and the group heartbeat staleness exceeds the threshold here. Need to set together with -enable_heartbeat_check")
+	heartbeatThreshold = flag.Int("group_heartbeat_threshold", 0, "VTGR will trigger backoff on inconsistent state if the group heartbeat staleness exceeds this threshold (in seconds). Should be used along with -enable_heartbeat_check")
 )
 
 // SQLGroup contains views from all the nodes within the shard
--- a/go/vt/vtgr/controller/group.go
+++ b/go/vt/vtgr/controller/group.go
@@ -34,29 +34,29 @@ var (
 	groupOnlineSize = stats.NewGaugesWithMultiLabels("MysqlGroupOnlineSize", "Online MySQL server in the group", []string{"Keyspace", "Shard"})
 	isLostQuorum    = stats.NewGaugesWithMultiLabels("MysqlGroupLostQuorum", "If MySQL group lost quorum", []string{"Keyspace", "Shard"})
 
-	heartbeatThreshold = flag.Int("group_heartbeat_threshold", 0, "Group heartbeat staleness threshold. Need to set together with -enable_heartbeat_check")
+	heartbeatThreshold = flag.Int("group_heartbeat_threshold_sec", 0, "VTGR will trigger backoff on inconsistent state and the group heartbeat staleness exceeds the threshold here. Need to set together with -enable_heartbeat_check")
 )
 
 // SQLGroup contains views from all the nodes within the shard
 type SQLGroup struct {
-	views                       []*db.GroupView
-	resolvedView                *ResolvedView
-	logger                      *log.Logger
-	size                        int
-	singlePrimary               bool
-	heartbeatStalenessThreshold int
-	statsTags                   []string
+	views              []*db.GroupView
+	resolvedView       *ResolvedView
+	logger             *log.Logger
+	size               int
+	singlePrimary      bool
+	heartbeatThreshold int
+	statsTags          []string
 	sync.Mutex
 }
 
 // NewSQLGroup creates a new SQLGroup
 func NewSQLGroup(size int, singlePrimary bool, keyspace, shard string) *SQLGroup {
 	return &SQLGroup{
-		size:                        size,
-		singlePrimary:               singlePrimary,
-		statsTags:                   []string{keyspace, shard},
-		logger:                      log.NewVTGRLogger(keyspace, shard),
-		heartbeatStalenessThreshold: *heartbeatThreshold,
+		size:               size,
+		singlePrimary:      singlePrimary,
+		statsTags:          []string{keyspace, shard},
+		logger:             log.NewVTGRLogger(keyspace, shard),
+		heartbeatThreshold: *heartbeatThreshold,
 	}
 }
 
@@ -254,19 +254,19 @@ func (group *SQLGroup) resolveLocked() error {
 			}
 			// Members in a group should eventually converge on a state
 			// if there is a partition, then a node should be removed from
-			// a group. If we a node is reported as ONLINE together with
-			// some other staet, we back off if we see a node with diverged state
+			// a group. If a node is reported as ONLINE together with
+			// some other state, we back off if we see a node with diverged state
 			if memberState != db.UNKNOWNSTATE &&
 				st.State != db.UNKNOWNSTATE &&
 				st.State != memberState &&
 				(st.State == db.ONLINE || memberState == db.ONLINE) {
 				group.logger.Warningf("found inconsistent member state for %v: %v vs %v", instance.Hostname, st.State, memberState)
-				if group.heartbeatStalenessThreshold != 0 &&
+				if group.heartbeatThreshold != 0 &&
 					// Check minStalenessResult among the group is not math.MaxInt32
 					// which means at least one node returns the lag from _vt.heartbeat table
 					// otherwise we don't trigger backoff on inconsistent state
 					minStalenessResult != math.MaxInt32 &&
-					minStalenessResult >= group.heartbeatStalenessThreshold {
+					minStalenessResult >= group.heartbeatThreshold {
 					group.logger.Warningf("ErrGroupBackoffError by staled heartbeat check %v", minStalenessResult)
 					var sb strings.Builder
 					for _, view := range group.views {
--- a/go/vt/vtgr/controller/group.go
+++ b/go/vt/vtgr/controller/group.go
@@ -17,7 +17,9 @@ limitations under the License.
 package controller
 
 import (
+	"flag"
 	"fmt"
+	"math"
 	"sort"
 	"strings"
 	"sync"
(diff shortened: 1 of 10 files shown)


mattlord:
Just noting that group membership changes go through the consensus protocol driven by the primary (paxos leader). So if the primary cannot communicate with enough peers/members to form a quorum -- it cannot change the membership as that is a state change and no state changes can be made. So in effect you are stuck with the last agreed upon cluster membership view on the primary. While the secondaries cannot talk to anyone so only report their local view  -- themselves, unable to communicate with the group.

Not sure if this helps or not, but it's something to keep in mind when considering how to deal with this situation. You are correct that when you lose quorum, you have to re-bootstrap the cluster.

sonne5:
> it cannot change the membership as that is a state change and no state changes can be made.

There is [group_replication_unreachable_majority_timeout](https://dev.mysql.com/doc/refman/8.0/en/group-replication-options.html#sysvar_group_replication_unreachable_majority_timeout) that can be set to unblock the group from the situation - after the timeout, the group would be exit automatically. What we observe is the group is stuck there even with this setting, and that's the place the logic in this PR could be helpful.

mattlord:
> > it cannot change the membership as that is a state change and no state changes can be made.
> 
> There is [group_replication_unreachable_majority_timeout](https://dev.mysql.com/doc/refman/8.0/en/group-replication-options.html#sysvar_group_replication_unreachable_majority_timeout) that can be set to unblock the group from the situation - after the timeout, the group would be exit automatically. What we observe is the group is stuck there even with this setting, and that's the place the logic in this PR could be helpful.

I don't see how `group_replication_unreachable_majority_timeout` is relevant here. The local view on those secondary nodes shows that they've left the group -- they are only aware of themselves. The problem here AFAICT is that we've lost quorum so we cannot make progress. You can only make progress -- which includes membership changes -- when you can form a quorum (simple majority of the last membership view which is exposed in the GR layer -- there is the GR layer, GCS, and XCom layers).

Cluster membership is a cluster state change initiated via the failure detector and is managed in the GCS (group communication system)/paxos implementation layer -- XCom -- and it cannot make progress in this case to change the membership as it requires consensus protocol rounds. The membership is a cluster state change, but not a MySQL state change, which is why it's only done solely in the XCom layer and then exposed via the `performance_schema.group_replication_members` table (you can see lower level XCom messages in the mysql error log):
  - https://dev.mysql.com/doc/refman/8.0/en/group-replication-group-membership.html
  - https://dev.mysql.com/blog-archive/the-king-is-dead-long-live-the-king-our-homegrown-paxos-based-consensus/
  - https://dev.mysql.com/blog-archive/group-replication-coping-with-unreliable-failure-detection/
  - https://dev.mysql.com/blog-archive/order-from-chaos-member-coordination-in-group-replication/
  - https://dev.mysql.com/blog-archive/mysql-group-replication-transaction-life-cycle-explained/

I apologize if you know all of this and I'm just missing something here. As noted, in any event, when the group cannot make progress because of a loss of quorum, the only thing you can do is re-bootstrap the cluster so I think your solution still fits.

sonne5:
> > > it cannot change the membership as that is a state change and no state changes can be made.
> > 
> > 
> > There is [group_replication_unreachable_majority_timeout](https://dev.mysql.com/doc/refman/8.0/en/group-replication-options.html#sysvar_group_replication_unreachable_majority_timeout) that can be set to unblock the group from the situation - after the timeout, the group would be exit automatically. What we observe is the group is stuck there even with this setting, and that's the place the logic in this PR could be helpful.
> 
> I don't see how `group_replication_unreachable_majority_timeout` is relevant here. The local view on those secondary nodes shows that they've left the group -- they are only aware of themselves. The problem here AFAICT is that we've lost quorum so we cannot make progress. You can only make progress -- which includes membership changes -- when you can form a quorum (simple majority of the last membership view which is exposed in the GR layer; there is the GR layer, GCS, and XCom layers).
> 
> Cluster membership is a cluster state change initiated via the failure detector and is managed in the GCS (group communication system)/paxos implementation layer -- XCom -- and it cannot make progress in this case to change the membership as it requires consensus protocol rounds. The membership is a cluster state change, but not a MySQL state change, which is why it's only done solely in the XCom layer and then exposed via the `performance_schema.group_replication_members` table (you can see lower level XCom messages in the mysql error log):
> 
> * https://dev.mysql.com/doc/refman/8.0/en/group-replication-group-membership.html
> * https://dev.mysql.com/blog-archive/the-king-is-dead-long-live-the-king-our-homegrown-paxos-based-consensus/
> * https://dev.mysql.com/blog-archive/group-replication-coping-with-unreliable-failure-detection/
> * https://dev.mysql.com/blog-archive/order-from-chaos-member-coordination-in-group-replication/
> * https://dev.mysql.com/blog-archive/mysql-group-replication-transaction-life-cycle-explained/
> 
> I apologize if you know all of this and I'm just missing something here. As noted, in any event, when the group cannot make progress because of a loss of quorum, the only thing you can do is re-bootstrap the cluster so I think your solution still fits.

@mattlord thanks for all the pointers. I think you are right that the group cannot make progress after a network partition. But if `group_replication_unreachable_majority_timeout` is set, the timeout should kick in when a node cannot reach the majority. In this case, a replica gets partitioned and report OFFLINE locally, the primary cannot reach out to others and cannot update it's view - after the timeout, the primary is supposed to exit the group as well and become OFFLINE instead of stuck with a staled group view. What we observed here is even with `group_replication_unreachable_majority_timeout`, the primary still reported a staled view, hence the logic in this PR could be triggered to help with the situation.

go/vt/vtgr/controller/group.go
```diff
@@ -31,22 +33,31 @@ import (
 var (
 	groupOnlineSize = stats.NewGaugesWithMultiLabels("MysqlGroupOnlineSize", "Online MySQL server in the group", []string{"Keyspace", "Shard"})
 	isLostQuorum    = stats.NewGaugesWithMultiLabels("MysqlGroupLostQuorum", "If MySQL group lost quorum", []string{"Keyspace", "Shard"})
+
+	heartbeatThreshold = flag.Int("group_heartbeat_threshold", 0, "Group heartbeat staleness threshold. Need to set together with -enable_heartbeat_check")
```
deepthi, line 37: It will be good to write a bit more description for the flag. It's unclear what this flag really affects.

go/vt/vtgr/controller/group.go
```diff
@@ -233,6 +252,30 @@ func (group *SQLGroup) resolveLocked() error {
 			if st.State == memberState && st.Role == memberRole && st.ReadOnly == isReadOnly {
 				continue
 			}
+			// Members in a group should eventually converge on a state
+			// if there is a partition, then a node should be removed from
+			// a group. If we a node is reported as ONLINE together with
```
deepthi, line 257: If we **see** a node?

go/vt/vtgr/controller/group.go
```diff
@@ -233,6 +252,30 @@ func (group *SQLGroup) resolveLocked() error {
...
+			// some other staet, we back off if we see a node with diverged state
```
deepthi, line 258: typo: is this supposed to be state?

go/vt/vtgr/controller/group.go
deepthi: Also is that supposed to be seconds? The unit is unclear too.

go/vt/vtgr/controller/group.go
```diff
@@ -31,22 +33,31 @@ import (
...
-	logger        *log.Logger
-	size          int
-	singlePrimary bool
-	statsTags     []string
+	views                       []*db.GroupView
+	resolvedView                *ResolvedView
+	logger                      *log.Logger
+	size                        int
+	singlePrimary               bool
+	heartbeatStalenessThreshold int
```
deepthi, line 47: Not a big fan of long field names, but if this is required to express the meaning, that's fine.

go/vt/vtgr/controller/refresh.go
```diff
@@ -44,10 +44,10 @@ var (
 // grInstance represents an instance that's running MySQL GR
 // it wraps a InstanceKey plus some tablet related information
 type grInstance struct {
-	instanceKey      *inst.InstanceKey
-	tablet           *topodatapb.Tablet
-	primaryTimeStamp time.Time
-	alias            string
+	instanceKey     *inst.InstanceKey
+	tablet          *topodatapb.Tablet
+	masterTimeStamp time.Time
```
deepthi, line 49: oops, change this back please :)

deepthi: and all other instances in this file.

go/vt/vtgr/controller/refresh_test.go
```diff
@@ -61,7 +61,7 @@ func TestRefreshTabletsInShard(t *testing.T) {
 	})
 	assert.Equal(t, testHost, instances[0].tablet.Hostname)
 	assert.Equal(t, int32(testPort0), instances[0].tablet.MysqlPort)
-	assert.Equal(t, topodatapb.TabletType_PRIMARY, instances[0].tablet.Type)
+	assert.Equal(t, topodatapb.TabletType_MASTER, instances[0].tablet.Type)
```
deepthi, line 64: all master->primary in this file too.. in fact you may be able to simply revert the changes to these two files?

go/vt/vtgr/controller/repair.go
```diff
@@ -301,7 +301,7 @@ func (shard *GRShard) stopAndRebootstrap(ctx context.Context) error {
 	return shard.dbAgent.RebootstrapGroupLocked(candidate.instanceKey, uuid)
 }
 
-func (shard *GRShard) getGTIDSetFromAll(skipPrimary bool) (*groupGTIDRecorder, *concurrency.AllErrorRecorder, error) {
+func (shard *GRShard) getGTIDSetFromAll(skipMaster bool) (*groupGTIDRecorder, *concurrency.AllErrorRecorder, error) {
```
deepthi, line 304: more of the same.

go/vt/vtgr/controller/group.go
sonne5: Updated the description and added `_sec` suffix.

go/vt/vtgr/controller/group.go
sonne5: Fixed

go/vt/vtgr/controller/group.go
sonne5: yep, fixed

go/vt/vtgr/controller/group.go
sonne5: i renamed it to `heartbeatThreshold`, i can change it back if it feels less clear

go/vt/vtgr/controller/refresh_test.go
sonne5: good catch, fixed

go/vt/vtgr/controller/group.go
```diff
@@ -31,22 +33,31 @@ import (
 var (
 	groupOnlineSize = stats.NewGaugesWithMultiLabels("MysqlGroupOnlineSize", "Online MySQL server in the group", []string{"Keyspace", "Shard"})
 	isLostQuorum    = stats.NewGaugesWithMultiLabels("MysqlGroupLostQuorum", "If MySQL group lost quorum", []string{"Keyspace", "Shard"})
+
+	heartbeatThreshold = flag.Int("group_heartbeat_threshold_sec", 0, "VTGR will trigger backoff on inconsistent state and the group heartbeat staleness exceeds the threshold here. Need to set together with -enable_heartbeat_check")
```
deepthi, line 37:
```suggestion
	heartbeatThreshold = flag.Int("group_heartbeat_threshold", 0, "VTGR will trigger backoff on inconsistent state if the group heartbeat staleness exceeds this threshold (in seconds). Should be used along with -enable_heartbeat_check")
```

deepthi: I prefer the old flag name to this one. Adding "(in seconds)" should help.
Sorry to nitpick on this, but flags are painful to change, and together with their description are part of the interface we present to users, so I tend to be picky about them.

deepthi (approved):
We can merge once you make one last change.
Thank you for your patience.

PR #9293, opened 2021-11-29
Backport12.0: Use decoded hex string when calculating the keyspace ID

mattlord:
## Description
This is a backport of https://github.com/vitessio/vitess/pull/9277, please see there for details.

The combination of https://github.com/vitessio/vitess/pull/9118 and https://github.com/vitessio/vitess/pull/9277 add more complete support for hex values of the form `x'e19a'` and `0xe19a` in queries within vitess, specifically:
1. The hex predicates/input values are normalized in the query -- without this the query planner cache is unused and each query is seen as unique, which has a very large impact on performance when some of your primary queries use hex values
2. The columns where you insert these hex values can (still) be used in primary vindexes as the keyspace ID is calculated using the bytes stored for the column in MySQL rather than the "raw" BindVar value which is the string literal representation used in the query

## Related Issue(s)
Backports: https://github.com/vitessio/vitess/pull/9277

## Checklist
- [x] Should this PR be backported? No, not beyond 12.0
- [x] Tests were added

### DIFF
--- a/go/mysql/endtoend/replication_test.go
+++ b/go/mysql/endtoend/replication_test.go
@@ -25,6 +25,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/require"
+
 	"vitess.io/vitess/go/vt/vtgate/evalengine"
 
 	"context"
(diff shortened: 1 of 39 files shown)


frouioui (approved):
Reviewed the original pull request (https://github.com/vitessio/vitess/pull/9277), this back port looks good to me

deepthi (approved):
Approving the backport.

PR #9295, opened 2021-11-29
only grab gh-ost exec if not overridden

choo-stripe:
## Description
[`ricebox`](https://github.com/GeertJohan/go.rice) is an interesting tool to embed assets in binaries. As best I can tell, it does some hackery with symbol tables and embedding extraneous zips in binaries to make assets available at runtime.

This is not really compatible with golang's standard tooling (or bazel), it's manipulation that must be done after-build (at least for `append`).

Notably, the ricebox embedded assets don't actually seem to be critical to run-time. At least, with `gh-ost`, there's a flag to override the bin.

Can we eliminate the dependence on this `ricebox` if the `gh-ost` bin is overridden?

## Related Issue(s)
- None

## Checklist

## Deployment Notes
None


mattlord (approved):
Makes sense and seems safe to me. Thanks!

deepthi (approved):
LGTM

PR #9296, opened 2021-11-30
Add support for changing Planner Version in vttestserver docker image

GuptaManan100:
## Description

This PR adds support for changing Planner Version in vttestserver docker image by providing the environment variable. This is needed for the rails sharded blog, so that the readers can follow along with the code examples if they want to without bringing up the entire Vitess cluster, but just using the docker image for vttestserver. However, rails only works with Gen4, so we need the functionality for changing the planner version.

## Checklist
- [x] Tests were added or are not required - Tested manually

### DIFF
--- a/go/cmd/vttestserver/main.go
+++ b/go/cmd/vttestserver/main.go
@@ -137,6 +137,9 @@ func init() {
 
 	flag.StringVar(&config.Charset, "charset", "utf8mb4", "MySQL charset")
 	flag.StringVar(&config.Collation, "collation", "", "MySQL collation")
+
+	flag.StringVar(&config.PlannerVersion, "planner_version", "v3", "Sets the default planner to use when the session has not changed it. Valid values are: V3, Gen4, Gen4Greedy and Gen4Fallback. Gen4Fallback tries the new gen4 planner and falls back to the V3 planner if the gen4 fails. All Gen4 versions should be considered experimental!")
+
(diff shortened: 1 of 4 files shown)


go/cmd/vttestserver/main.go
```diff
@@ -137,6 +137,9 @@ func init() {
 
 	flag.StringVar(&config.Charset, "charset", "utf8mb4", "MySQL charset")
 	flag.StringVar(&config.Collation, "collation", "", "MySQL collation")
+
+	flag.StringVar(&config.PlannerVersion, "planner_version", "v3", "Sets the default planner to use when the session has not changed it. Valid values are: V3, Gen4, Gen4Greedy and Gen4Fallback. Gen4Fallback tries the new gen4 planner and falls back to the V3 planner if the gen4 fails. All Gen4 versions should be considered experimental!")
```
mattlord, line 141: Since this is in main/13.0, should we still say Gen4 is experimental?

mattlord (approved):

frouioui (approved):

go/cmd/vttestserver/main.go
GuptaManan100: Gen4 is still marked experimental on main and in the planner_version help of vtgate. I replicated it here. When we make that the default, we should also do it for this.

Issue #7344, opened 2021-01-21
Migration testing in CI


deepthi:
#### Feature Description
In #7294, we automated upgrading and downgrading the build version by 1 major version. There are some additional scenarios that should be tested to ensure that we are complying with [VEP-3](https://github.com/vitessio/enhancements/blob/master/veps/vep-3.md)
* Ability to restore backups taken with the previous version after upgrade/downgrade
* Ability for a cluster to operate correctly in the midst of an upgrade / downgrade. E.g. vtctld and vttablets are still running but vtgate is restarted with the new version.
 
#### Use Case(s)
* Documented in VEP-3

shlomi-noach:
I will work on the 2nd bullet, ie halfway-upgrade.

> E.g. vtctld and vttablets are still running but vtgate is restarted with the new version.

What's a complete list (`order by importance desc`) for half-upgrade setups I should be investigating?

deepthi:
1. Upgrade vtgate only
1. Upgrade vtctld only
1. Upgrade all vttablets
1. Upgrade a subset of vttablets
1. Combination of 1 or 2 with 3 or 4

shlomi-noach:
Cool, thank you! The approach I will take is to always take the cluster down, then to start on top of the existing data with mixed binaries. Thus, the approach will **not** test e.g. just taking down `vtgate` and bringing up a newer `vtgate`. We can strive to do that, too, at some point in the future. For now the approach I'm suggesting is easiest.

I _think_ we may be able to squeeze in all of these tests in a single CI and within a  reasonable amount of runtime, seeing that most of the time is spent on setup/build. But we may need to split to two different CI workflows.

deepthi:
That approach should be fine.

shlomi-noach:
See https://github.com/vitessio/vitess/pull/7368 for partial upgrade tests

frouioui:
Pull request https://github.com/vitessio/vitess/pull/8471 relates to this issue.

frouioui:
I have opened #9300, I am starting the implementation of the matrix detailed in https://github.com/vitessio/vitess/issues/7344#issuecomment-949297139.

PR #9300, opened 2021-11-30
Upgrade Downgrade Testing

frouioui:
## Description

This pull request introduces a more comprehensive upgrade/downgrade test strategy in order to comply with [VEP-3](https://github.com/vitessio/enhancements/blob/main/veps/vep-3.md). The new strategy is thoroughly discussed in https://github.com/vitessio/vitess/issues/7344 (specifically in this comment: https://github.com/vitessio/vitess/issues/7344#issuecomment-949297139).

This pull request focuses on testing the `query serving` part of Vitess. The `backup` and `cluster management` parts will follow the same approach as the one introduced by this pull request.

The methodology used to perform the upgrade downgrade test is the following:

1. Build Vitess binaries for versions `n` _(the current commit)_ **and** `n-1` _(the latest major release)_ 
2. Move the binaries to a `n_bin` and `n-1_bin` directory, so we can reuse them later
3. We use a `bin` directory with all the binaries that our test will use
4. Keep the binaries from `n` in the `bin` directory
5. Start the test suite once. This will test with both vtgate and vttablet using version `n`
6. Swap the binaries. We want to use vtgate version `n-1` now. Copy the `vtgate` binary from the `n-1_bin` we created earlier into the `bin` directory
7. Start the test suite once. This will test with vtgate using version `n-1` and vttablet using version `n`
8. Swap the binaries one more time. We want to use vtgate version `n` and vttablet version `n-1`. Use the binaries stored in the `n_bin` and `n-1_bin` directories.
9. Start the test suite once. This will test with vtgate using version `n` and vttablet using version `n-1`

This methodology is similar to the one detailed in @deepthi's [comment](https://github.com/vitessio/vitess/issues/7344#issuecomment-949297139), but instead of creating a global cluster, we use the already-existing end-to-end tests.

The tests ran in this new workflow are defined using the `config.json` file. All the tests in this file that are tagged with `upgrade_downgrade_query_serving` will automatically be tested by the new workflow. Example below:

```json
"vtgate_schematracker_unsharded": {
	"File": "unused.go",
	"Args": ["vitess.io/vitess/go/test/endtoend/vtgate/schematracker/unsharded"],
	"Command": [],
	"Manual": false,
	"Shard": "vtgate_schema_tracker",
	"RetryMax": 1,
	"Tags": ["upgrade_downgrade_query_serving"]
},
```

## Trigger upgrade downgrade testing

The label `Upgrade Downgrade` enables the run of the workflow. Without the label on your pull request, the workflow will not run. This is done to limit the number of long and not-always-necessary workflows.

cc @deepthi @GuptaManan100 ; What do you think about this label check?

## Checklist


deepthi:
> What do you think about this label check?

In general, opt-in CI is not a good idea. PR authors may not realize that they are making a change that can break upgrade/downgrade.

test/config.json
```diff
@@ -748,7 +748,7 @@
 			"Manual": false,
 			"Shard": "vtgate_schema_tracker",
 			"RetryMax": 1,
-			"Tags": []
+			"Tags": ["upgrade_downgrade_query_serving"]
```
deepthi, line 751: Is it expected that there will be more tags of this kind?

GuptaManan100:
I think that the opt-in CI would be useful, in the sense that for smaller PRs or for PRs which do not require running the upgrade downgrade testing, we can avoid doing so. But this would mean that it is the responsibility of the maintainer merging the PR to add the label and run this CI before merging if deemed necessary and not already present

.github/workflows/upgrade_downgrade_test.yml
```diff
@@ -0,0 +1,189 @@
+name: Upgrade Downgrade Testing
```
GuptaManan100, line 1: Should the name also include Query Serving since this test only runs the Query serving upgrade downgrade tests?

GuptaManan100 (approved):

test/config.json
frouioui: I am expecting to have at least two more, one for backups and one for cluster management

.github/workflows/upgrade_downgrade_test.yml
frouioui: Yes it should

frouioui:
@deepthi @GuptaManan100 How about an opt-out CI strategy? With the query serving's upgrade downgrade testing workflow taking approximately an hour, I can see the case where we want to skip the workflow.

.github/workflows/upgrade_downgrade_test.yml
frouioui: This is fixed via 903b2c3

GuptaManan100:
I think an opt-out strategy might be okay. For the PRs which we don't want to run them, then we can specifically prevent running them

PR #9304, opened 2021-11-30
[vtadmin-web] Consolidate font sizing into tailwind.config.js

doeg:
## Description

This consolidates all of our font sizing variables into the tailwind.config.js file. Since we use a 10px basis, as opposed to the tailwind/browser default of 16px, our font scale differs from [the tailwind docs](https://tailwindcss.com/docs/font-size), hence the overrides. (This will also make it 1000x easier to adjust font sizes if we do ever decide to change the rem basis.) 

Also added some lorem ipsum to the /debug page until we get Storybook set up. 

## Related Issue(s)

- #9215 

## Checklist

## Deployment Notes

N/A

### DIFF
--- a/web/vtadmin/tailwind.config.js
+++ b/web/vtadmin/tailwind.config.js
@@ -25,6 +25,14 @@ module.exports = {
                 'sans-serif',
             ],
         },
+        fontSize: {
+            sm: '1.2rem',
+            base: '1.4rem',
+            lg: '1.6rem',
+            xl: '2rem',
+            '2xl': '2.8rem',
+            '3xl': '3.6rem',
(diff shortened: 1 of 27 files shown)


web/vtadmin/tailwind.config.js
```diff
@@ -25,6 +25,16 @@ module.exports = {
                 'sans-serif',
             ],
         },
+        // Presently, we use a 1rem = 10px basis, which diverges from
```
notfelineit, line 28: 👍 but switch will be SO EASY when we have to!!

notfelineit (approved):

Issue #9307, opened 2021-12-01
vtgate serves queries from replicas when its source is unavailable and replication is unhealthy


mattlord:
#### Overview of the Issue

You can use vitess for read scale out by explicitly routing queries to shard replicas using `<keyspace>/<shard>@replica` shard targeting. In order to limit the potential window of inconsistency for these reads and bound how stale the reads can be, you tell the tablet to reflect the tablet as being unhealthy when the replication lag gets beyond [`-unhealthy_threshold`](https://vitess.io/docs/reference/programs/vttablet/). 

The problem is that when the replica cannot talk to its source -- because the source mysqld is down or otherwise unreachable -- then mysqld reports [`seconds_behind_master`](https://dev.mysql.com/doc/refman/5.7/en/show-slave-status.html) as `NULL` because it's unknown. We then silently convert that to 0 and thus the replica tablet passes its healthcheck with flying colors and is never marked as unhealthy even though it has not been able to talk to the source. This can lead to replicas serving VERY stale reads that can in turn cause a cascade of downstream issues.

#### Reproduction Steps

Steps to reproduce this issue:
```
$ make docker_local && ./docker/local/run.sh

# create 1 record to make it easier to see when we're serving queries
mysql commerce -e "insert into customer values (1, 'you@planetscale.com')"

# kill the current replica tablet
kill $(ps auxww | grep vttablet | grep 101 | awk '{print $2}'); sleep 2;

# restart it with a 10 second replication lag limit for being considered healthy
vttablet -topo_implementation etcd2 -topo_global_server_address localhost:2379 -topo_global_root /vitess/global -log_dir /vt/vtdataroot/tmp -log_queries_to_file /vt/vtdataroot/tmp/vttablet_0000000101_querylog.txt -tablet-path zone1-0000000101 -init_keyspace commerce -init_shard 0 -init_tablet_type replica -health_check_interval 5s -enable_semi_sync -enable_replication_reporter -backup_storage_implementation file -file_backup_storage_root /vt/vtdataroot/backups -restore_from_backup -port 15101 -grpc_port 16101 -service_map grpc-queryservice,grpc-tabletmanager,grpc-updatestream -pid_file /vt/vtdataroot/vt_0000000101/vttablet.pid -vtctld_addr http://2a35755b0ef2:15000/ -unhealthy_threshold=10s &

sleep 5

# kill its replication source mysqld_safe and mysqld processes so that replication is now broken
[...]

PR #6665, opened 2020-09-03
vttablet: heartbeat always enabled

shlomi-noach:
Closes https://github.com/vitessio/vitess/issues/6662

- `-heartbeat_enable` flag now always assumed to be enabled
- `-heartbeat_interval` is now always set to some value in the range `(0..1]` seconds, default `250ms`

Issue #6662, opened 2020-09-02
Suggestion: always enable heartbeat on vttablet


shlomi-noach:
`vttablet` has these flags: 

- `-heartbeat_enable`
- `-heartbeat_interval`

The use of heartbeat is so important, that I suggest it must always be enabled. Plus, it should have a maximum interval that is subsecond.

Heartbeats are to be used by [throttling services](https://github.com/vitessio/vitess/issues/6661), by `orchestrator`, etc.

Is there any reason why we woudln't always enable heartbeat?

sougou:
It was badly implemented and had bugs. But we've fixed them all over time. We should encourage using this feature now.

deepthi:
We document that only one of `-heartbeat_enable` and `-enable_replication_reporter` should be used. Right now they both default to false. From that POV it is safe to default heartbeat to true. However, it should be documented in the release notes as a possibly breaking change.

shlomi-noach:
> We document that only one of -heartbeat_enable and -enable_replication_reporter should be used

I couldn't find where this is documented, searched in `vitessio/vitess` and in `vitessio/website` repos. The one related thing I found in `vitessio/website` is:

> `-heartbeat_enable` and `-heartbeat interval duration`: cause vttablet to write heartbeats to the sidecar database. This information is also used by the replication reporter to assess replica lag.

shlomi-noach:
#6665 was created to close this issue. However, #6668, the throttler PR, implicitly enables heartbeat without changing configuration. So #6665 was closed without merging.

We're still in experimental mode for the tablet throttler. It's debatable whether we can call this issue "implicitly implemented" or "won't fix"... Opting for "won't fix" for now and closing.

PR #9308, opened 2021-12-01
Estimate replica lag when seconds behind from mysqld is unknown

mattlord:
## Description
When mysqld tells us that [`seconds_behind_master`](https://dev.mysql.com/doc/refman/5.7/en/show-slave-status.html) is unknown (`NULL`) — or we get back any other value that cannot be converted to a uint — let's estimate the replication lag seconds value ([`seconds_behind_master` itself is fuzzy](https://dev.mysql.com/doc/refman/5.7/en/replication-administration-status.html)) in vttablet's healthcheck stats using the last seen value and the time elapsed since. This gives us the most accurate estimate possible of the replica lag when the replica cannot talk to its source and is constantly retrying and in a state where `Slave_IO_Running` == `Connecting`, [which we consider to be equivalent to running](https://github.com/vitessio/vitess/blob/5c7ee04b64ecf2cbe7b7430091f7e14d16f39723/go/mysql/flavor.go#L308-L313) as reconnects happen regularly anyway e.g. due to [`slave_net_timeout`](https://dev.mysql.com/doc/refman/5.7/en/replication-options-replica.html#sysvar_slave_net_timeout), and allows us to correctly support the defined [`-unhealthy_threshold`](https://vitess.io/docs/reference/programs/vttablet/) behavior for the replica tablet — preventing us from serving queries from a replica that may have very stale data.

With these changes, the results of [the test here](https://github.com/vitessio/vitess/issues/9307#issue-1067845996) seem to be expected and correct:
```
vitess@fb5036c15bc9:/vt/local$ for i in {1..5}; do
>   echo -n "mysqld says:"
>   command mysql -u root --socket=/vt/vtdataroot/vt_0000000101/mysql.sock -e "show slave status\G" | grep Seconds | cut -d: -f2
>   echo -n "vttablet says: "
>   curl -s localhost:15101/debug/status_details | jq -r '.[1].Value'
>   echo
>   sleep 5
> done
mysqld says: 0
vttablet says: 0s

mysqld says: NULL
vttablet says: 5s

mysqld says: NULL
vttablet says: 10s

mysqld says: NULL
vttablet says: 15s

mysqld says: NULL
vttablet says: 20s


vitess@fb5036c15bc9:/vt/local$ mysql commerce/0@replica -e "select * from customer"
ERROR 1105 (HY000) at line 1: target: commerce.0.replica: no healthy tablet available for 'keyspace:"commerce" shard:"0" tablet_type:REPLICA'

vitess@fb5036c15bc9:/vt/local$ curl -s localhost:15101/debug/status_details
[
 {
  "Key": "Current State",
  "Class": "healthy",
  "Value": "REPLICA: Serving"
 },
 {
  "Key": "Replication Lag",
  "Class": "unhealthy",
  "Value": "25s"
 }
]

vitess@fb5036c15bc9:/vt/local$ mysql -e "show vitess_tablets"
+-------+----------+-------+------------+-------------+------------------+--------------+----------------------+
| Cell  | Keyspace | Shard | TabletType | State       | Alias            | Hostname     | PrimaryTermStartTime |
+-------+----------+-------+------------+-------------+------------------+--------------+----------------------+
| zone1 | commerce | 0     | PRIMARY    | NOT_SERVING | zone1-0000000100 | fb5036c15bc9 | 2021-12-03T01:10:04Z |
| zone1 | commerce | 0     | REPLICA    | NOT_SERVING | zone1-0000000101 | fb5036c15bc9 |                      |
| zone1 | commerce | 0     | RDONLY     | SERVING     | zone1-0000000102 | fb5036c15bc9 |                      |
+-------+----------+-------+------------+-------------+------------------+--------------+----------------------+
```

## Related Issue(s)

 - Fixes: https://github.com/vitessio/vitess/issues/9307

## Checklist
- [x] Tests were added
- [x] Documentation is not required (**_[I don't think?](https://github.com/vitessio/vitess/pull/9308#issuecomment-985178985)_**)

### DIFF
--- a/go/mysql/flavor.go
+++ b/go/mysql/flavor.go
@@ -19,14 +19,12 @@ package mysql
 import (
 	"errors"
 	"fmt"
-	"math"
 	"strconv"
 	"strings"
 
 	"context"
 
 	"vitess.io/vitess/go/sqltypes"
-	"vitess.io/vitess/go/vt/log"
 	"vitess.io/vitess/go/vt/proto/vtrpc"
 	"vitess.io/vitess/go/vt/vterrors"
 )
@@ -319,10 +317,13 @@ func parseReplicationStatus(fields map[string]string) ReplicationStatus {
 	status.ConnectRetry = int(parseInt)
 	parseUint, err := strconv.ParseUint(fields["Seconds_Behind_Master"], 10, 0)
 	if err != nil {
-		log.Errorf("Could not compute replica lag from seconds_behind_master value of '%s', this means that replication is unhealthy; setting replica lag to max value to prevent the serving of queries", fields["Seconds_Behind_Master"])
-		parseUint = math.MaxUint32
+		// we could not parse the value into a valid uint -- most commonly because the value is NULL from the
+		// database -- so let's reflect that the underlying value was unknown on our last check
+		status.ReplicationLagUnknown = true
+	} else {
+		status.ReplicationLagUnknown = false
+		status.ReplicationLagSeconds = uint(parseUint)
 	}
-	status.ReplicationLagSeconds = uint(parseUint)
 	parseUint, _ = strconv.ParseUint(fields["Master_Server_Id"], 10, 0)
 	status.SourceServerID = uint(parseUint)
 
--- a/go/mysql/flavor.go
+++ b/go/mysql/flavor.go
@@ -19,12 +19,14 @@ package mysql
 import (
 	"errors"
 	"fmt"
+	"math"
 	"strconv"
 	"strings"
 
 	"context"
 
 	"vitess.io/vitess/go/sqltypes"
+	"vitess.io/vitess/go/vt/log"
 	"vitess.io/vitess/go/vt/proto/vtrpc"
 	"vitess.io/vitess/go/vt/vterrors"
 )
@@ -315,7 +317,11 @@ func parseReplicationStatus(fields map[string]string) ReplicationStatus {
 	status.SourcePort = int(parseInt)
 	parseInt, _ = strconv.ParseInt(fields["Connect_Retry"], 10, 0)
 	status.ConnectRetry = int(parseInt)
-	parseUint, _ := strconv.ParseUint(fields["Seconds_Behind_Master"], 10, 0)
+	parseUint, err := strconv.ParseUint(fields["Seconds_Behind_Master"], 10, 0)
+	if err != nil {
+		log.Errorf("Could not compute replica lag from seconds_behind_master value of '%s', this means that replication is unhealthy; setting replica lag to max value to prevent the serving of queries", fields["Seconds_Behind_Master"])
+		parseUint = math.MaxUint32
+	}
 	status.ReplicationLagSeconds = uint(parseUint)
 	parseUint, _ = strconv.ParseUint(fields["Master_Server_Id"], 10, 0)
 	status.SourceServerID = uint(parseUint)
--- a/go/test/endtoend/tabletmanager/tablet_health_test.go
+++ b/go/test/endtoend/tabletmanager/tablet_health_test.go
@@ -24,6 +24,7 @@ import (
 	"strings"
 	"sync"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/require"
 
@@ -134,7 +135,7 @@ func TestHealthCheck(t *testing.T) {
 	// make sure the health stream is updated
 	result, err := clusterInstance.VtctlclientProcess.ExecuteCommandWithOutput("VtTabletStreamHealth", "-count", "1", rTablet.Alias)
 	require.NoError(t, err)
-	verifyStreamHealth(t, result)
+	verifyStreamHealth(t, result, true)
 
 	// then restart replication, make sure we stay healthy
 	err = clusterInstance.VtctlclientProcess.ExecuteCommand("StartReplication", rTablet.Alias)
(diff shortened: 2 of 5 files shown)


go/mysql/flavor.go
```diff
@@ -315,7 +317,11 @@ func parseReplicationStatus(fields map[string]string) ReplicationStatus {
 	status.SourcePort = int(parseInt)
 	parseInt, _ = strconv.ParseInt(fields["Connect_Retry"], 10, 0)
 	status.ConnectRetry = int(parseInt)
-	parseUint, _ := strconv.ParseUint(fields["Seconds_Behind_Master"], 10, 0)
+	parseUint, err := strconv.ParseUint(fields["Seconds_Behind_Master"], 10, 0)
+	if err != nil {
+		log.Errorf("Could note compute replica lag from seconds_behind_master value of %s, this means that the replica is unhealthy -- setting replica lag to the max value to prevent the serving of queries when replication is not functioning properly", fields["Seconds_Behind_Master"])
```
deepthi, line 322: typo: note -> not

deepthi (approved):
Looks fine to me.
I thought @aquarapid wanted us to add more error checking though..

mattlord:
I don't think that I like this — better to err on the side of safety and not provide incorrect results, IMO — but saving the rough idea here in case everyone else disagrees:
```
diff --git a/go/mysql/flavor.go b/go/mysql/flavor.go
index 97265888be..37436db22e 100644
--- a/go/mysql/flavor.go
+++ b/go/mysql/flavor.go
@@ -38,6 +38,9 @@ var (

+
+	// Keep track of consecutive errors calculating replica lag
+	consecutiveReplicaLagCalculationErrors = 0
 )

 const (
@@ -319,8 +322,16 @@ func parseReplicationStatus(fields map[string]string) ReplicationStatus {
 	status.ConnectRetry = int(parseInt)
 	parseUint, err := strconv.ParseUint(fields["Seconds_Behind_Master"], 10, 0)
 	if err != nil {
-		log.Errorf("Could not compute replica lag from seconds_behind_master value of '%s', this means that replication is unhealthy; setting replica lag to max value to prevent the serving of queries", fields["Seconds_Behind_Master"])
-		parseUint = math.MaxUint32
+		// in order to avoid flapping on recently caught up replicas when the network
+		// is unstable, let's only take action if the lag was > 0 last time we checked
+		// OR we've had two or more consecutive checks where the lag was unknown
+		if status.ReplicationLagSeconds != 0 || consecutiveReplicaLagCalculationErrors > 1 {
+			log.Errorf("Could not compute replica lag from seconds_behind_master value of '%s', this means that replication is unhealthy; setting replica lag to max value to prevent the serving of queries", fields["Seconds_Behind_Master"])
+			parseUint = math.MaxUint32
+		}
+		consecutiveReplicaLagCalculationErrors++
+	} else {
+		consecutiveReplicaLagCalculationErrors = 0
 	}
 	status.ReplicationLagSeconds = uint(parseUint)
 	parseUint, _ = strconv.ParseUint(fields["Master_Server_Id"], 10, 0)
```

mattlord:
@deepthi, while I don't think that this PR as a stand-alone change warrants doc updates... after having gone through this I feel like we could probably use a new docs page on "Using Vitess for Read Scale Out" that documents how to do `@replica` shard targeting and how the `vtgate`, `vttablet`, and `mysqld` coordinate to prevent potentially stale reads beyond a configured window of inconsistency (walking through how that's configured). What does everyone else think? If everyone thinks that's a good idea then I can create a docs PR for that.

shlomi-noach (commented):
We do have a heartbeat mechanism running in tablets. It's still not always-on, though I've advocated for it to be, and perhaps this is a good opportunity to revive that discussion.
We use the heartbeat value in the throttler service and in vreplication. I think we should explore reusing it here, too.

mattlord:
> We do have a heartbeat mechanism running in tablets. It's still not always-on, though I've advocated for it to be, and perhaps this is a good opportunity to revive that discussion.
> We use the heartbeat value in the throttler service and in vreplication. I think we should explore reusing it here, too.

@shlomi-noach We did not add another mechanism here. Are you talking about something other than what I’m using in the PR? It’s existing the tablet healthcheck. If there’s another one then I’m not aware of it. Can you share details about what you’re referring to?

shlomi-noach:
@mattlord I misunderstood and thought there was a new functionality here.

I think we might still want to first consult with our heartbeat mechanism. I don't know whether this PR should change right now, but I think ideally all our lag calculations should use the heartbeat,

Examples:

- tablet throttler uses this query to determine lag: https://github.com/vitessio/vitess/blob/ea40f2d1e8ccbf165c6ccc1c86d5529ddaa30b0a/go/vt/vttablet/tabletserver/throttle/throttler.go#L82
- tablet throttler is still not enabled by default. It takes a `-enable-lag-throttler` flag to enable it, in which case it enabled the heartbeat mechanism.
- This discussion about whether/not to always enable heartbeat: https://github.com/vitessio/vitess/issues/6662#event-3844112294

Anyway, I think this digresses from the PR. LGTM and perhaps keep the discussion going :pray:

mattlord:
Hi @shlomi-noach ! 

Today we have the vttablet healthcheck stream that vtgates subscribe to which helps the gateway determine if the tablet should be serving queries or not. That is exposed to users via:
```
curl -s localhost:15101/debug/status_details
[
 {
  "Key": "Current State",
  "Class": "healthy",
  "Value": "REPLICA: Serving"
 },
 {
  "Key": "Replication Lag",
  "Class": "healthy",
  "Value": "0s"
 }
```

(vtgate exposes resulting info in `SHOW VITESS_TABLETS` and `/debug/status`.)

The replication lag portion of that is enabled by default via `-enable_replication_reporter` and what we do with the data can be configured via `-health_check_interval` and `-unhealthy_threshold` in the tablet, `-discovery_low_replication_lag` and `-discovery_high_replication_lag_minimum_serving` in the gateway. We're using that here in this PR but correctly handling `seconds_behind_master` when it's NULL or unknown.

We _**also**_ have the heartbeat mechanism that IS NOT enabled by default but can be controlled via `-heartbeat_enable` and `-heartbeat_interval` in the tablet. This then causes the tablet to store a heartbeat timestamp info to its mysqld in the `_vt.heartbeat` table.

We then have a throttler interface that is NOT enabled by default but can be enabled in vttablet via `-enable-lag-throttler` and `-enable-tx-throttler` and how we generated the data and what we do with it can be controlled via `-throttle_metrics_query`, `-throttle_threshold`, `-throttle_metrics_threshold`, `-throttle_check_as_check_self`. Those results are exposed to the user via `/throttler/check` and `/throttler/check-self` in the tablet's web UI. Example underlying protobuf info:
```
target_replication_lag_sec: 2,
max_replication_lag_sec: 10,
initial_rate: 100,
max_increase: 1,
emergency_decrease: 0.5,
min_duration_between_increases_sec: 40,
max_duration_between_increases_sec: 62,
min_duration_between_decreases_sec: 20,
spread_backlog_across_sec: 20,
age_bad_rate_after_sec: 180,
bad_rate_increase: 0.1,
max_rate_approach_threshold: 0.9,
```

As a matter of principle, I would agree that we should consolidate these things to have a single shared mechanism for determining the properties of a tablet: liveness, responsiveness (ability to handle future load), and consistency. Perhaps this is something we can discuss over dinner.  I do, however, think it's orthogonal to this PR as this PR is a bug fix that is simply correcting behavior related to the healthcheck.

Thanks!

shlomi-noach:
@mattlord agreed that this PR should proceed and no need to address hearbeat/throttler right now.

> We also have the heartbeat mechanism that IS NOT enabled by default but can be controlled via -heartbeat_enable and -heartbeat_interval in the tablet. This then causes the tablet to store a heartbeat timestamp info to its mysqld in the _vt.heartbeat table.
> We then have a throttler interface that is NOT enabled by default but can be enabled in vttablet via -enable-lag-throttler and -enable-tx-throttler with what and how we generate the data and what we do with it can be controlled via -throttle_metrics_query, -throttle_threshold, -throttle_metrics_threshold, -throttle_check_as_check_self. Those results are exposed to the user via /throttler/check and /throttler/check-self in the tablet's web UI. Example underlying protobuf info:

Agreed that this looks like quite a lot of configuration - though `-enable-lag-throttler` is the only thing the user would have to worry about. By enabling lag throttler, Vitess implicitly enables the heartbeat mechanism. The metrics query is by default the correct query to compare timestamps on `_vt.heartbeat` and with a `1s` throttle threshold. So the idea is that the user just needs to enable the throttler and not worry about anything else; the defaults should be good for all but the most technical users who want greater control.

deepthi (approved):

go/test/endtoend/tabletmanager/tablet_health_test.go
```diff
@@ -183,18 +225,22 @@ func checkTabletType(t *testing.T, tabletAlias string, typeWant string) {
...
-	assert.True(t, serving, "Tablet should be in serving state")
 	assert.True(t, UID > 0, "Tablet should contain uid")
-	// replicationLagSeconds varies till 7200 so setting safe limit
-	assert.True(t, replicationLagSeconds < 10000, "replica should not be behind primary")
+	if expectHealthy {
+		assert.True(t, serving, "Tablet should be in serving state")
+		// replicationLagSeconds varies till 7200 so setting safe limit
+		assert.True(t, replicationLagSeconds < 10000, "replica should not be behind primary")
+	} else {
+		assert.True(t, (!serving || replicationLagSeconds >= uint32(tabletUnhealthyThreshold.Seconds())), "Tablet should not be in serving and healthy state")
```
mattlord, line 242: To help explain this assertion, as far as the healthcheck is concerned the tablet can be SERVING while unhealthy with regards to replica lag. For example:
```
vitess@fb5036c15bc9:/vt/local$ curl -s localhost:15101/debug/status_details
[
 {
  "Key": "Current State",
  "Class": "healthy",
  "Value": "REPLICA: Serving"
 },
 {
  "Key": "Replication Lag",
  "Class": "unhealthy",
  "Value": "25s"
 }
]
```

deepthi:
> We do have a heartbeat mechanism running in tablets. It's still not always-on, though I've advocated for it to be, and perhaps this is a good opportunity to revive that discussion.
> We use the heartbeat value in the throttler service and in vreplication. I think we should explore reusing it here, too.

We have discussed this in the past, and I still think it is desirable to at least make the heartbeat the default.
We have some work to do before we can make that change though as discussed here:
https://github.com/vitessio/vitess/pull/6665#issuecomment-701634139

PR #9309, opened 2021-12-01
Adds parsing support for string introducers

GuptaManan100:
## Description
Vitess only had parsing support for some of the string introducers. This PR adds it for the rest of them.

This PR also introduces an additional struct `IntroducerExpr` which stores the introducers and the character set requested, instead of piggybacking on the `UnaryExpr` struct. This will make it easier to use.

## Checklist


go/vt/sqlparser/constants.go
```diff
@@ -396,11 +433,52 @@ const (
 	TildaOp
 	BangOp
 	BinaryOp
+	NStringOp
+)
+
+// Constant for Enum Type - IntroducerCharacterSet
+const (
+	Armscii8Op IntroducerCharacterSet = iota
```
vmg, line 441: Could you rename the introducer character sets so they have a `Charset` suffix, instead of `Op`? They're no longer unary ops. :)

**edit**: actually -- thinking about it: since we're assigning these manually from the grammar, it'd be easier to work with the charsets if the actual values in the AST were lowercase strings. We could use them as-is with the `collations` package, which would be super convenient.

So we could just remove this enum altogether and just assign a string to the `IntroducerExpr.CharacterSet` node.

vmg (changes requested):

vmg (approved):
Looking good!

PR #9310, opened 2021-12-01
evalengine: Fuzz!

vmg:
## Description

As explained previously to @frouioui @systay and @GuptaManan100, I've implemented a simple fuzzer for the evaluation engine that generates random expressions and compares their evaluation result's against MySQL's to see if we're being consistent.

It resulted in a _lot_ of small fixes, which have been implemented in this branch. The most significant one, which is not small at all, is the introduction of a `decimal` package to handle decimal precision numbers. We're using a fork of https://github.com/ericlagergren/decimal (which is license compatible with Vitess) with several small changes to make it behave closer to how the decimal implementation works in MySQL. In order to simplify integration and development, the relevant parts of the library (i.e. the functionality we use) has been vendored inside the `evalengine` package.
 
I'm going to keep pushing a few more bug fixes to this branch for the remaining issues found by the fuzzer.

## Checklist


go/sqltypes/bind_variables.go
```diff
@@ -123,6 +128,8 @@ func ValueBindVariable(v Value) *querypb.BindVariable {
 	return &querypb.BindVariable{Type: v.typ, Value: v.val}
 }
 
+type DecimalFloat float64
```
frouioui, line 131: I think this should live at the top of the file

go/vt/vtgate/evalengine/comparisons.go
```diff
@@ -126,63 +150,104 @@ func evalResultsAreDates(l, r EvalResult) bool {
...
+	if hasNullEvalResult(lVal, rVal) {
+		if lVal.typ == rVal.typ {
+			return 0, nil
+		}
+		return 1, nil
+	}
+	return evalCompare1(lVal, rVal)
+}
+
+func evalCompare(lVal, rVal EvalResult) (comp int, isNull bool, err error) {
```
frouioui, line 210: We should rename this to `evalCompareAll` as it compares all types of `EvalResult` (tuples, non-tuples, nulls)

go/vt/vtgate/evalengine/comparisons.go
```diff
@@ -126,63 +150,104 @@ func evalResultsAreDates(l, r EvalResult) bool {
...
 	if hasNullEvalResult(lVal, rVal) {
 		return 0, true, nil
 	}
+	comp, err = evalCompare1(lVal, rVal)
+	return comp, false, err
+}
+
+// For more details on comparison expression evaluation and type conversion:
+// 		- https://dev.mysql.com/doc/refman/8.0/en/type-conversion.html
+func evalCompare1(lVal, rVal EvalResult) (comp int, err error) {
```
frouioui, line 227: once the above method `evalCompare` is renamed, we can rename this to `evalCompare`

go/vt/vtgate/evalengine/comparisons.go
```diff
@@ -331,24 +449,49 @@ func (i *InOp) String() string {
...
-		matched = l.Match.Match(left.bytes)
+		return l.Match.Match(left)
 	} else {
-		coll := collations.Local().LookupByID(left.collation.Collation)
-		wc := coll.Wildcard(right.bytes, 0, 0, 0)
-		matched = wc.Match(left.bytes)
+		coll := collations.Local().LookupByID(coll)
+		wc := coll.Wildcard(right, 0, 0, 0)
+		return wc.Match(left)
 	}
```
frouioui, line 459: the else statement can be omitted

go/vt/vtgate/evalengine/mysql_test.go
```diff
@@ -0,0 +1,75 @@
+package evalengine
+
```
frouioui, line 2: Add header

go/vt/vtgate/evalengine/integration/fuzz_test.go
```diff
@@ -0,0 +1,204 @@
+package integration
+
```
frouioui, line 2: Add header

go/vt/vtgate/evalengine/integration/fuzz_test.go
```diff
@@ -0,0 +1,204 @@
...
+	rand         *rand.Rand
+	ratioTuple   int
+	ratioSubexpr int
+	tupleLen     int
+
+	operators  []string
+	primitives []string
+}
+
+type dummyCollation collations.ID
```
frouioui, line 46: lets move these at the top of the file

go/vt/vtgate/evalengine/integration/main_test.go
```diff
@@ -0,0 +1 @@
+../../../../mysql/collations/integration/main_test.go
```
frouioui, line 1: @vmg, is this a way to import the content of `mysql/collations/integration/main_test.go` to this file?

frouioui (approved):
I love it! thank you @vmg, looks good to me. I will take care of dealing with the few comments I left

go/vt/vtgate/evalengine/integration/fuzz_test.go
```diff
@@ -0,0 +1,222 @@
...
+func (d dummyCollation) CollationIDLookup(_ sqlparser.Expr) collations.ID {
+	return collations.ID(d)
+}
+
+func TestSelect(t *testing.T) {
+	var conn = mysqlconn(t)
+	defer conn.Close()
+
+	select {}
+}
```
frouioui, line 79: This needs to be changed

frouioui (changes requested):

go/vt/vtgate/evalengine/integration/fuzz_test.go
frouioui: @vmg, I removed the test via 1c2ef13

frouioui (commented):
The evalengine's fuzzer is failing on my side when I run it using MySQL80. Here is a snippet of the failures:
```
mismatch for query "SELECT 0 NOT LIKE \"foo\"": local=UINT64(1), remote=INT64(1)
mismatch for query "SELECT NULL NOT LIKE \"FOO\"": local=UINT64(1), remote=NULL
mismatch for query "SELECT -1 = \"fOo\"": local=UINT64(0), remote=INT64(0)
mismatch for query "SELECT 1 <=> 1": local=UINT64(1), remote=INT64(1)
```

It seems that we are returning different types than what MySQL80 does

systay:
out of interest - why fork https://github.com/ericlagergren/decimal instead of contributing fixes and depending on it?

AdamKorcz:
Very cool. Would be great to get this running continuously on OSS-fuzz which would require a rewrite to go-fuzz or waiting until OSS-fuzz supports native fuzzing which will happen in 2022.

vmg:
> why fork https://github.com/ericlagergren/decimal instead of contributing fixes and depending on it?

I don't think the maintainer of that package will accept fixes that implement MySQL's specific behavior. The package implements the ISO decimal standard, and MySQL doesn't so...

frouioui (approved):
awesome work 🏆

Issue #9290, opened 2021-11-29
TestVstreamFailover Flaky Test


GuptaManan100:
#### Overview of the Issue

The test `TestVstreamFailover` has been flaky on the main branch and me and @rohit-nayak-ps  started investigating this when it started blocking merging open PRs.

@rohit-nayak-ps tried running the test on different commits until he found the one that introduced the issue. This commit turned out to be https://github.com/vitessio/vitess/pull/9033/

The error that we were seeing from the test looked like -
```
I1123 13:53:59.032337     611 vtctlclient_process.go:175] Executing vtctlclient with command: vtctlclient -server localhost:15999 PlannedReparentShard -keyspace_shard=product/0 -new_primary=zone1-101
I1123 13:54:04.067374     611 vstream_test.go:125] 2021-11-23 13:54:04.067368416 +0100 CET m=+41.924175818:: remote error: Code: INVALID_ARGUMENT
rpc error: code = InvalidArgument desc = vttablet: rpc error: code = InvalidArgument desc = GTIDSet Mismatch: requested source position:MySQL56/57e7b78b-4c5c-11ec-a6f2-04ed332e05c2:1-1352, current target vrep position: MySQL56/57e7162a-4c5c-11ec-a75f-04ed332e05c2:1-29,57e7b78b-4c5c-11ec-a6f2-04ed332e05c2:1-1332
```

It wasn't immediately obvious what the issue was. It seemed that PRS allowed the new primary to be promoted even before it was caught up. The logs show that the old primary was more advanced than the new primary promoted. However, looking at the code this seemed like an impossible scenario. So after finding the commit which introduced this issue, I talked it over with @harshit-gangal, and we came to a hypothesis. The problem and the logs associated are written here.

This issue has arisen because of the interaction of 3 components of Vitess - VStreamer(@rohit-nayak-ps), PlannedReparentShard(@deepthi, @GuptaManan100) and Heartbeat Writer (@aquarapid), which has been changed in the commit. 

Consider the following scenario - 
Let our current primary be `zone1-0000000100`, and the vstream is running.

1. Everything is functioning normally and the vstream is running.
2. Planned Reparent Shard is issued to change the primary to `zone1-0000000101`.
3. In the PRS code-path we follow the following steps - 
  a. Set `zone1-0000000100` to read only
  b. Find the executed gtid set on `zone1-0000000100`, let us call this set A.
[...]

Issue #9311, opened 2021-12-01
Task: mysqlctl package cleanup


deepthi:
- [ ] migrate errors to use vterrors
- [ ] replace context.TODO with real contexts

PR #9312, opened 2021-12-01
Change default for super_read_only to true

deepthi:
## Description

While looking into #9290 we discovered that the heartbeat writer was creating errant transactions during a failover (PlannedReparent). While we need to fix that properly, it has been suggested that we should run with `super_read_only` on MySQL versions that support it.
This PR does two things.
- change default for use_super_read_only to true
- set super-read-only in cnf files

## Related Issue(s)

#9290 

## Checklist
- [ ] Documentation was added or is not required - needs to be added in website docs

## Deployment Notes

It is expected that this change is safe and backwards-compatible. Anyone who is relying on the current behavior should pass `-use_super_read_only=false` on the vttablet command line, and make sure they are using a custom my.cnf instead of the one provided as the default by Vitess.

### DIFF
--- a/config/mycnf/mysql57.cnf
+++ b/config/mycnf/mysql57.cnf
@@ -12,6 +12,9 @@ relay_log_info_repository = TABLE
 relay_log_purge = 1
 relay_log_recovery = 1
 
+# we should never need super privileges
+super-read-only
+
 # In MySQL 5.7 the default charset is latin1
 
 character_set_server = utf8
--- a/go/vt/mysqlctl/backup.go
+++ b/go/vt/mysqlctl/backup.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"time"
 
@@ -58,8 +59,6 @@ const (
 const (
 	// replicationStartDeadline is the deadline for starting replication
 	replicationStartDeadline = 30
-
-	Error1193 = "Unknown system variable"
 )
 
 var (
@@ -343,8 +342,8 @@ func Restore(ctx context.Context, params RestoreParams) (*BackupManifest, error)
 	// This is safe, since we're restarting MySQL after the restore anyway
 	params.Logger.Infof("Restore: disabling super_read_only")
 	if err := params.Mysqld.SetSuperReadOnly(false); err != nil {
-		if strings.Contains(err.Error(), Error1193) {
-			params.Logger.Warningf("Restore: server does not know about super_read_only; maybe MariaDB? Continuing anyway.")
+		if strings.Contains(err.Error(), strconv.Itoa(mysql.ERUnknownSystemVariable)) {
+			params.Logger.Warningf("Restore: server does not know about super_read_only, continuing anyway...")
 		} else {
 			params.Logger.Errorf("Restore: unexpected error while trying to set super_read_only: %v", err)
 			return nil, err
--- a/go/vt/vttablet/tabletmanager/rpc_replication.go
+++ b/go/vt/vttablet/tabletmanager/rpc_replication.go
@@ -433,7 +433,7 @@ func (tm *TabletManager) demotePrimary(ctx context.Context, revertPartialFailure
 		// Setting super_read_only also sets read_only
 		if err := tm.MysqlDaemon.SetSuperReadOnly(true); err != nil {
 			if strings.Contains(err.Error(), strconv.Itoa(mysql.ERUnknownSystemVariable)) {
-				log.Warningf("server does not know about super_read_only; maybe MariaDB? Continuing anyway.")
+				log.Warningf("server does not know about super_read_only, continuing anyway...")
 			} else {
 				return nil, err
 			}
--- a/go/vt/vttablet/tabletmanager/rpc_replication.go
+++ b/go/vt/vttablet/tabletmanager/rpc_replication.go
@@ -19,6 +19,7 @@ package tabletmanager
 import (
 	"flag"
 	"fmt"
+	"strconv"
 	"strings"
 	"time"
 
@@ -39,7 +40,7 @@ import (
 
 var (
 	enableSemiSync   = flag.Bool("enable_semi_sync", false, "Enable semi-sync when configuring replication, on primary and replica tablets only (rdonly tablets will not ack).")
-	setSuperReadOnly = flag.Bool("use_super_read_only", false, "Set super_read_only flag when performing planned failover.")
+	setSuperReadOnly = flag.Bool("use_super_read_only", true, "Set super_read_only flag when performing planned failover.")
 )
 
 // ReplicationStatus returns the replication status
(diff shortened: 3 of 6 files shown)


config/mycnf/mysql57.cnf
```diff
@@ -12,6 +12,9 @@ relay_log_info_repository = TABLE
 relay_log_purge = 1
 relay_log_recovery = 1
 
+# we should never need super privileges
+super-read-only
```
mattlord, line 16: While I think this is a good idea — always starting mysqld in super-read-only mode — it does mean that it's something that we'd need to turn off on the primary tablet during the tablet repair. I vaguely recall that we don't do that today?

go/vt/vttablet/tabletmanager/rpc_replication.go
```diff
@@ -431,7 +432,11 @@ func (tm *TabletManager) demotePrimary(ctx context.Context, revertPartialFailure
 	if *setSuperReadOnly {
 		// Setting super_read_only also sets read_only
 		if err := tm.MysqlDaemon.SetSuperReadOnly(true); err != nil {
-			return nil, err
+			if strings.Contains(err.Error(), strconv.Itoa(mysql.ERUnknownSystemVariable)) {
+				log.Warningf("server does not know about super_read_only; maybe MariaDB? Continuing anyway.")
```
mattlord, line 436: I'd remove the maybe MariaDB part as the user could be using MySQL 5.6 or who knows what. MariaDB may eventually add the option too. 

Although I still think that we should end any official support for MariaDB as it's so different now and we don't test with it. That's another discussion though :-)

config/mycnf/mysql57.cnf
mattlord: Yeah, we don't touch it in repair: https://github.com/vitessio/vitess/blob/93c811b9f9356e6407c4564cb95bfff9a402f0da/go/vt/vttablet/tabletmanager/rpc_replication.go#L554-L669

Only in init primary: https://github.com/vitessio/vitess/blob/93c811b9f9356e6407c4564cb95bfff9a402f0da/go/vt/vttablet/tabletmanager/rpc_replication.go#L245-L282

And demote primary: https://github.com/vitessio/vitess/blob/93c811b9f9356e6407c4564cb95bfff9a402f0da/go/vt/vttablet/tabletmanager/rpc_replication.go#L378-L469

When this would cause problems is when the mysqld instance for a primary tablet crashes or is otherwise restarted. Ironically, I happen to be doing that in a test I just added here: https://github.com/vitessio/vitess/pull/9308/commits/ef39eb8ed4084c7aa06d3d1740745c9ac4a6b645

mattlord: I'm unsure what the historical reasons were for not fixing read-only in tablet repair, but IMO it would be a good idea.

deepthi: I see what you mean. Tablet restarts and demotes itself to replica through a path that doesn't go through `demotePrimary`. I don't know if there was an actual reason for not doing it. Probably a miss from when `use_super_read_only` was implemented.

go/vt/vttablet/tabletmanager/rpc_replication.go
```diff
@@ -39,7 +40,7 @@ import (
 
 var (
 	enableSemiSync   = flag.Bool("enable_semi_sync", false, "Enable semi-sync when configuring replication, on primary and replica tablets only (rdonly tablets will not ack).")
-	setSuperReadOnly = flag.Bool("use_super_read_only", false, "Set super_read_only flag when performing planned failover.")
+	setSuperReadOnly = flag.Bool("use_super_read_only", true, "Set super_read_only flag when performing planned failover.")
```
GuptaManan100, line 43: do we also want to deprecate this flag? We can even keep it around so that anyone using Vitess with mysql 5.6 will just have to pass this as a false parameter

GuptaManan100 (approved):

config/mycnf/mysql57.cnf
deepthi: I looked at the code some more. When the tablet demotes itself, it first calls `demotePrimary` which takes care of read_only (and super_read_only) before `SetReplicationSource` is called. So that path is fine.
The other case where tablet is up but mysqld is down or restarted is something that vtorc will handle. It has never been handled properly prior to vtorc (and replication_reporter was not designed to do that either).

deepthi: As far as the primary goes, we call `SetReadOnly(false)` which also turns off super_read_only (per MySQL docs).

go/vt/vttablet/tabletmanager/rpc_replication.go
deepthi: We should keep it for one release so that users can fallback if this causes any unforeseen problems.

mattlord (approved):
❤️

deepthi:
Slack discussion: https://vitess.slack.com/archives/C0PQY0PTK/p1638915929375700
There is a path for that user, so it should be fine to merge this now.

go/vt/mysqlctl/backup.go
```diff
@@ -343,8 +342,8 @@ func Restore(ctx context.Context, params RestoreParams) (*BackupManifest, error)
 	// This is safe, since we're restarting MySQL after the restore anyway
 	params.Logger.Infof("Restore: disabling super_read_only")
 	if err := params.Mysqld.SetSuperReadOnly(false); err != nil {
-		if strings.Contains(err.Error(), Error1193) {
-			params.Logger.Warningf("Restore: server does not know about super_read_only; maybe MariaDB? Continuing anyway.")
+		if strings.Contains(err.Error(), strconv.Itoa(mysql.ERUnknownSystemVariable)) {
```
mattlord, line 345: This feels a little brittle to me, but it's not new in this PR. This should be an INVALID_ARGUMENT vitess error right? https://github.com/vitessio/vitess/blob/f238c0bb9ab5757b2f9d1771081e899ea06f2df6/go/vt/vttablet/tabletserver/tabletserver.go#L1575-L1592

So can we leverage this here? https://pkg.go.dev/errors#Is 

```
if err.Is(err, mysql.ERUnknownSystemVariable) {
```
Or if that doesn't work for some reason:
```
if err.Is(err, vtrpcpb.Code_INVALID_ARGUMENT) {
```

This is non-blocking though, as we were already doing the substring search.

deepthi: This is a good idea. We should combine the Code + substring. I'll make the change before I merge.

shlomi-noach (approved):
LGTM

GuptaManan100 (approved):

go/vt/mysqlctl/backup.go
deepthi: I forgot that we can't do this yet. See https://github.com/vitessio/vitess/pull/9240#discussion_r754662749
Here's where we lose the type information
https://github.com/vitessio/vitess/blob/d74de0c575396dcc0fdc4e4fc9dc00f799638652/go/vt/mysqlctl/query.go#L84
I opened #9311 for this. Once that is done, we can clean up error handling in these files.

Issue #9010, opened 2021-10-18
VTTablet heartbeat write should use DBA connection


sonne5:
#### Feature Description
When vttablet creates a heartbeatWriter, it takes connection parameters for both vt_app and vt_dba. However, when execute CREATE TABLE statement, it always uses vt_app: https://github.com/vitessio/vitess/blob/main/go/vt/vttablet/tabletserver/repltracker/writer.go#L173

We should use vt_dba for the `CREATE TABLE` statement, and vt_app user for the actual insert/update

sonne5:
cc @aquarapid @mattlord

GuptaManan100:
@sonne5 The PR that made the change #9033, actually created another issue with PRS. The issue is described in detail in #9290. TL;DR, the `vt_dba` user has the super privilege which allows it to run commands even when the underlying MySQL instance is set to read only. This causes the old primary to diverge from the new primary when we perform a PRS.

Keeping this in mind, we have reverted the change to the user in the heartbeat writer for now back to `vt_app`. The PR that makes that change is #9299. 

Is there a specific reason why you wanted to use the `vt_dba` user for the creation of table and database? `vt_app` has the CREATE privilege as seen in the `init_db.sql` file.

Also, as far as the pool was concerned, it was only using the `vt_dba` to kill the queries which were running longer than the set timeout

sonne5:
We don't grant `CREATE` privilege to `vt_app` user hence the change in https://github.com/vitessio/vitess/pull/9299 will break our integration.

+1 to what @deepthi suggested here https://github.com/vitessio/vitess/issues/9290#issuecomment-982813381 , it seems a bit risky to assume the privilege for `vt_app` user here - for example, i'm under the impression (incorrectly) that `vt_app` was meant to used by requests from vtgate and not being used by heart beat by design :-) I think the short term workaround might be using `vt_allprivs` to do the heartbeat write here?

cc @aquarapid

deepthi:
thanks @sonne5 we'll go back to the drawing board and figure out how to fix this properly.

GuptaManan100:
@sonne5 I have made the required changes in https://github.com/vitessio/vitess/pull/9320

Issue #6474, opened 2020-07-27
RFC: Standardizing on db privileges


sougou:
Vitess currently requires you to configure the following privileges:
* app: for serving application queries
* dba: has SUPER privileges, for executing configuration changes to the database
* filtered: for vreplication to apply changes
* repl: for replica credentials
* allprivs: has all privileges except SUPER, currently unused.
* erepl: deprecated.

As features were organically added, we haven't followed much discipline on how we used these privileges internally. It is time we trim and standardize on these. I propose the following changes:
* Deprecate dba for vttablet. mysqlctl may still need it. Using DBA privileges has mostly led to problems because we've accidentally written to read-only masters. I believe allprivs is sufficient to perform any maintenance operation. There is one specific use case that will need to be addressed: the replica metadata. We will find a different way to solve this problem.
* Reduce privileges for app. The app user currently has too many privileges. These should ideally be reduced to operations performed by the app, whatever they may be. If we want to prevent the app from issuing DDLs, we should be able to enforce that by revoking these from the app user.
* Use allprivs for vreplication: Since vreplication replays all binlog events, these also include DDLs and other administrative commands that were applied to the source.
* repl must remain unchanged with its own set of privileges because of the mysql-mysql connections.

In summary, we'll be left with app, allprivs, and repl for vttablet.

PR #9313, opened 2021-12-01
DO NOT MERGE: remove CREATE permission from vt_app user

deepthi:
## Description

There are places where we are creating DB/tables using the vt_app user, and we are implicitly assuming that this user always has CREATE permissions.
Some people are trying to run without the CREATE perms so this is an attempt to see what all breaks in that case.

## Related Issue(s)

#9010 

## Checklist


deepthi:
Closing this for now. We now know that fixing #6474 is non-trivial 😞

PR #9315, opened 2021-12-02
Remove query trees

systay:
## Description
This PR is a rewrite of parts of the gen4 planner. 

A lot of the heavy lifting in the planner, where the join order is decided on, was planned using query trees. Query trees worked for what we wanted, but it's not the typical abstraction to use. After reading a couple of papers on plan rewriting, we realized that we needed to use an abstraction close to what academia uses.

An example should help. Here is a query:

```sql
SELECT 1
FROM A 
  JOIN B on A.id = B.id
  JOIN C on B.id = C.id
```

`A` and `B` can me merged into a single route, but we have to evaluate the join with `C` on the vtgate side. A plan using query trees for this query would look something like this:

```
            join
          /     \
   route(AxB)   route(C)
```

The content inside the route is in a different data structure, so pushing something down under the route requires changing from operator to relations.

With the new physical operators, the plan would instead look like this:

```
               join
             /     \
         route    route
           |        |
         join     Tbl(C)
         /   \
    Tbl(A)  Tbl(B)
```

The join above and under the route are of the same type - nothing has to change just because we push it under or over a route. This is the biggest benefit with this new way of looking at the plans - it's easier to change the plan and push things down under the route.

When planning is done, everything under the route gets turned into an AST object representing the query, and this is the query that we then send down to vttablet/mysql.

## Checklist


go/vt/vtgate/planbuilder/operator_to_query_test.go
```diff
@@ -0,0 +1,23 @@
...
+limitations under the License.
+*/
+
+package planbuilder
+
+import "testing"
+
+func TestDerivedTable(t *testing.T) {
+
+}
```
GuptaManan100, line 23: We should remove this file

go/vt/vtgate/planbuilder/operator_transformers.go
```diff
@@ -0,0 +1,564 @@
...
+
+	vtrpcpb "vitess.io/vitess/go/vt/proto/vtrpc"
+	"vitess.io/vitess/go/vt/sqlparser"
+	"vitess.io/vitess/go/vt/vtgate/engine"
+	"vitess.io/vitess/go/vt/vtgate/vindexes"
+
+	"vitess.io/vitess/go/vt/vterrors"
+)
+
+func transformOpToLogicalPlan(ctx *plancontext.PlanningContext, op abstract.PhysicalOperator) (logicalPlan, error) {
```
GuptaManan100, line 42: This function name can go back to being `transformToLogicalPlan`

go/vt/vtgate/planbuilder/physical/operator_funcs.go
```diff
@@ -0,0 +1,363 @@
...
+	default:
+		return nil, vterrors.Errorf(vtrpcpb.Code_INTERNAL, "this should not happen - tried to remove predicate from table op")
+	}
+}
+
+func breakExpressionInLHSandRHS(
+	ctx *plancontext.PlanningContext,
+	expr sqlparser.Expr,
+	lhs semantics.TableSet,
+) (bvNames []string, columns []*sqlparser.ColName, rewrittenExpr sqlparser.Expr, err error) {
```
GuptaManan100, line 334: This function exists in horizon_planning too. We should export this one and use it there too instead of having 2 copies.

go/vt/vtgate/planbuilder/horizon_planning.go
```diff
@@ -1205,3 +1207,128 @@ func isJoin(plan logicalPlan) bool {
...
+	}
+	return nil
+}
+
+func breakExpressionInLHSandRHS(
+	ctx *plancontext.PlanningContext,
+	expr sqlparser.Expr,
+	lhs semantics.TableSet,
+) (bvNames []string, columns []*sqlparser.ColName, rewrittenExpr sqlparser.Expr, err error) {
+	rewrittenExpr = sqlparser.CloneExpr(expr)
```
GuptaManan100, line 1306: Copy of a function in operator_funcs

go/vt/vtgate/planbuilder/plan_test.go
```diff
@@ -243,6 +245,7 @@ func TestSysVarSetDisabled(t *testing.T) {
 }
 
 func TestOne(t *testing.T) {
+	// runGen4New = false
```
GuptaManan100, line 248: Cleanup

go/vt/vtgate/planbuilder/plan_test.go
```diff
@@ -669,9 +672,7 @@ func testFile(t *testing.T, filename, tempDir string, vschema *vschemaWrapper) {
 			//       produces the same plan as the V3 planner does
 			testName := fmt.Sprintf("%d Gen4: %s", tcase.lineno, tcase.comments)
 			t.Run(testName, func(t *testing.T) {
-				if out != tcase.output2ndPlanner {
-					t.Errorf("Gen4 - %s:%d\nDiff:\n%s\n[%s] \n[%s]", filename, tcase.lineno, cmp.Diff(tcase.output2ndPlanner, out), tcase.output2ndPlanner, out)
-				}
```
GuptaManan100, line 674: We should add this back

GuptaManan100 (approved):

go/vt/vtgate/planbuilder/horizon_planning.go
systay: yeah, we need both for now.
I want to move horizon planning to the physical operators, and when we do that, we can get rid of this code

go/vt/vtgate/planbuilder/plan_test.go
systay: 👍

go/vt/vtgate/planbuilder/plan_test.go
systay: 👍

go/vt/vtgate/planbuilder/horizon_planning.go
systay: actually, I removed this one

go/vt/vtgate/planbuilder/plancontext/planning_context.go
```diff
@@ -0,0 +1,59 @@
...
+		SemTable:       semTable,
+		VSchema:        vschema,
+		JoinPredicates: map[sqlparser.Expr][]sqlparser.Expr{},
+		SkipPredicates: map[sqlparser.Expr]interface{}{},
+	}
+	return ctx
+}
+
+func (c PlanningContext) IsSubQueryToReplace(e sqlparser.Expr) bool {
+	ext, ok := e.(*sqlparser.Subquery)
```
harshit-gangal, line 49: nit: should PlanningContext be passed as a reference?

harshit-gangal (approved):

go/vt/vtgate/planbuilder/plancontext/planning_context.go
systay: I only moved the type, didn't change uses of it. you might be correct though, not sure

PR #9316, opened 2021-12-02
Fuzzing: Disable logging for mysql fuzzers

AdamKorcz:
## Description

The mysql fuzzers print out lots of error messages. This disables the `Errorf` logging.

## Checklist
- [ ] Should this PR be backported?
- [ ] Tests were added or are not required
- [ ] Documentation was added or is not required

PR #9317, opened 2021-12-02
Fuzzing: Add 2 collations fuzzers

AdamKorcz:
## Description

Adds two collations fuzzers and instructs OSS-fuzz to include them.

## Checklist

### DIFF
--- a/go/mysql/collations/fuzz.go
+++ b/go/mysql/collations/fuzz.go
@@ -1,3 +1,6 @@
+//go:build gofuzz
+// +build gofuzz
+
 /*
 Copyright 2021 The Vitess Authors.
 Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


go/test/fuzzing/collate_fuzzer.go
```diff
@@ -0,0 +1,63 @@
...
+package fuzzing
+
+import (
+	fuzz "github.com/AdaLogics/go-fuzz-headers"
+
+	"vitess.io/vitess/go/mysql/collations"
+)
+
+func FuzzCollateUnicode(data []byte) int {
+	c := &collations.Collation_unicode_general_ci{}
```
vmg, line 23: This is not correct, you're not supposed to instantiate collation objects directly: they must be looked up directly from an Environment. I guess this is not properly explained in the API docs. :)

go/test/fuzzing/collate_fuzzer.go
```diff
@@ -0,0 +1,63 @@
...
+	isPrefix, err := f.GetBool()
+	if err != nil {
+		return 0
+	}
+	_ = c.Collate(left, right, isPrefix)
+	return 1
+}
+
+func FuzzCollateWildcard(data []byte) int {
+	c := &collations.Collation_multibyte{}
```
vmg, line 46: Likewise, this is not a valid collation.

go/test/fuzzing/collate_fuzzer.go
AdamKorcz: Thanks for the heads up. I rewrote it a bit to initialize the same way as [the existing fuzzer](https://github.com/vitessio/vitess/blob/main/go/mysql/collations/uca_test.go#L53).

go/mysql/collations/fuzz.go
```diff
@@ -0,0 +1,52 @@
...
+	left, err := f.GetBytes()
+	if err != nil {
+		return 0
+	}
+	right, err := f.GetBytes()
+	if err != nil {
+		return 0
+	}
+	_ = coll.Collate(left, right, false)
+	return 1
```
frouioui, line 51: are we asserting anything or simply making sure it does not crash?

systay: fuzzing is all about the latter. we are just making sure that nothing panics

vmg (approved):

AdamKorcz:
This one alone with https://github.com/vitessio/vitess/pull/9316 and https://github.com/vitessio/vitess/pull/9339 should be ready.

systay:
Replaced by #9475

PR #9319, opened 2021-12-03
Re-enable default Make behavior of failing target if any command fails

dctrwatson:

## Description
With `.ONESHELL` enabled in the Makefile, it disables the default behavior of failing targets if any command fails. As the default is only `.SHELLFLAGS = -c`.

## Checklist


deepthi (approved):
Yay, thank you!
Verified locally that this works as we need it to.

PR #9320, opened 2021-12-03
Make the heartbeat writer use 2 pools

### LINKED ISSUE #9010: VTTablet heartbeat write should use DBA connection

GuptaManan100:
## Description

Currently the heartbeat writer uses `vt_app` user to create the database, table and also run the upsert commands. This isn't correct since `vt_app` might not have the `CREATE` privilege, causing the create database and table commands to fail. This PR addresses this issue by creating 2 app pools, one for `vt_app` to run the insert commands and `vt_allprivs` to run the create commands.

As part of this PR, `WithDDL` has also changed, and now takes two functions instead of one in its `Execute` method. The second function, if provided, is used for running the DDL commands. This allows the caller of this function to have two different execution functions for DDLs and other queries. This change will be helpful in the future too, in all the other usages of `WithDDL` which are using the `vt_app` user for running the DDLs too.

## Related Issue(s)

- Fixes #9010 
- Implements the discussion https://github.com/vitessio/vitess/issues/9290#issuecomment-982127317

## Checklist

### DIFF
--- a/go/vt/withddl/withddl.go
+++ b/go/vt/withddl/withddl.go
@@ -55,8 +55,6 @@ func (wd *WithDDL) DDLs() []string {
 // It takes 2 functions, one to run the query and the other to run the
 // DDL commands. This is generally needed so that different users can be used
 // to run the commands. i.e. AllPrivs user for DDLs and App user for query commands.
-// If only 1 function is provided, then that is the function used to run
-// both the type of queries.
 // Funcs can be any of these types:
 // func(query string) (*sqltypes.Result, error)
 // func(query string, maxrows int) (*sqltypes.Result, error)
@@ -67,12 +65,9 @@ func (wd *WithDDL) Exec(ctx context.Context, query string, fQuery interface{}, f
 	if err != nil {
 		return nil, err
 	}
-	execDDL := execQuery
-	if fDDL != nil {
-		execDDL, err = wd.unify(ctx, fDDL)
-		if err != nil {
-			return nil, err
-		}
+	execDDL, err := wd.unify(ctx, fDDL)
+	if err != nil {
+		return nil, err
 	}
 	qr, err := execQuery(query)
 	if err == nil {
--- a/go/vt/withddl/withddl.go
+++ b/go/vt/withddl/withddl.go
@@ -52,17 +52,29 @@ func (wd *WithDDL) DDLs() []string {
 
 // Exec executes the query using the supplied function.
 // If there are any schema errors, it applies the DDLs and retries.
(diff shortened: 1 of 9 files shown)


go/vt/withddl/withddl.go
```diff
@@ -52,17 +52,29 @@ func (wd *WithDDL) DDLs() []string {
...
 	}
-	qr, err := exec(query)
+	execDDL := execQuery
+	if fDDL != nil {
+		execDDL, err = wd.unify(ctx, fQuery)
+		if err != nil {
+			return nil, err
+		}
+	}
+	qr, err := execQuery(query)
```
harshit-gangal, line 77: looks like here fDDL needs to be used inside.
Also, can you add a unit test that passes both functions?

sonne5 (commented):
Thanks for working on this!

I wonder if it makes sense to do a round of auditing to make sure the create table ddl for the internal tables (e.g., heartbeat, schema migration, shard split, etc) are all using `vt_allprivs` instead of `vt_app`? We ran into a few other privilege cases where we end up switching vt_app to vt_dba for schema migration as well.

Related: idk if it would be an overkill, if we decide `vt_allprivs` will be responsible for creating all the internal meta tables, would it be helpful to add a check and raise errors if user end up not granting `CREATE` privilege to it?

GuptaManan100:
@sonne5 you are right in that we need to go over and look at all the uses of `vt_app`. There are other places as well where we use vt_app for create statements. We will fix it too but it's going to be a larger refactor

go/vt/withddl/withddl.go
GuptaManan100: Yes, fixed both now

deepthi:
@sonne5 right now a whole bunch of things [fail](https://github.com/vitessio/vitess/pull/9313) if I remove `CREATE` from `vt_app`, so it's clear that we are relying on that. Also Manan and I went through the code and `vt_allprivs` is being used in only one place.
There is already a tracking [issue](https://github.com/vitessio/vitess/issues/6474) for this. We just need to do what that issue is saying.

harshit-gangal (approved):

deepthi (approved):
Nice work!

sougou (approved):

PR #9330, opened 2021-12-07
Fix error checking in ERS

### LINKED ISSUE #9329: Bug in ERS checking the pointer of the error received
#### Overview of the Issue

The `StopReplicationAndBuildStatusMaps` function calls the `StopReplicationAndGetStatus` rpc and proceeds to check whether the received error is `ErrNotReplica`. We use this information to know that the tablet in question considered itself to be the primary or not. While working on a different issue, I realised that we were never executing this switch case. On further testing, it was found that the error message that we receive is actually a GRPC error which contains the `ErrNotReplica` error. This leads to the condition always being evaluated to false.

This also meant that if we had 2 primary tablets, ERS would not be able to fix it because of this issue, even though it should be able to.

This also means that ERS cannot fix the replication if there are more than 1 replicas with no replication status, either due to startup or because someone cleared them via `RESET REPLICA ALL`

#### Binary version

main

GuptaManan100:
## Description

The `StopReplicationAndBuildStatusMaps` function calls the `StopReplicationAndGetStatus` rpc and proceeds to check whether the received error is `ErrNotReplica`. We use this information to know that the tablet in question considered itself to be the primary or not. 

However, we should not directly compare the error pointers since we receive an RPC error and the original error we sent from the vttablet. This PR changes one of the test to send the correct error message and fixes the underlying problem.

Also, as part of this PR, the two constant errors have been moved to the vterrors package from the mysql package.

## Related Issue(s)

- Fixes #9329 

## Checklist


GuptaManan100:
@harshit-gangal @deepthi To fix the issue currently we have changed the type of `ErrNoReplica` from a normal error to a `vterrors` error, we are sending it as a grpc error and then converting back to a vterror to check its type. But since type check is insufficient, we also need to do a substring match for the exact error message.

There is another approach that we can take. For now in the mysql package we have the constants for the mysql error numbers. We can add a new error number representing `ErrNoReplica` which does not collide with MySQL error numbers. Then we can send an error with that error message and use that to check the type. There are a couple of pros and cons associated with this approach

Pro - Approach is easier to understand by looking at the code.
Con - Blurs the boundary between Vitess and MySQL errors. We would now have Vitess-specific MySQL errors apart from vterrors.

WDYT?

go/test/endtoend/reparent/emergencyreparent/ers_test.go
```diff
@@ -202,3 +202,21 @@ func TestPullFromRdonly(t *testing.T) {
 	err = checkInsertedValues(ctx, t, newPrimary, insertVal)
 	require.NoError(t, err)
 }
+
+// TestTwoReplicaNoReplicationStatus checks that ERS is able to fix
+// two replicas which do not have any replication status
+func TestTwoReplicaNoReplicationStatus(t *testing.T) {
```
deepthi, line 208: nit: `TwoReplicas`

go/vt/vterrors/constants.go
```diff
@@ -38,3 +42,12 @@ const (
 	// PrimaryVindexNotSet is the error message to be used when there is no primary vindex found on a table
 	PrimaryVindexNotSet = "table '%s' does not have a primary vindex"
 )
+
+const NoReplicationStatusStr = "no replication status"
```
deepthi, line 46: Can you group all the `const` together in one block?

go/vt/vtctl/reparentutil/replication.go
```diff
@@ -181,8 +182,9 @@ func StopReplicationAndBuildStatusMaps(
 		logger.Infof("getting replication position from %v", alias)
 
 		_, stopReplicationStatus, err := tmc.StopReplicationAndGetStatus(groupCtx, tabletInfo.Tablet, replicationdatapb.StopReplicationMode_IOTHREADONLY)
-		switch err {
-		case mysql.ErrNotReplica:
+
+		switch {
+		case vterrors.Code(err) == vtrpc.Code_INTERNAL && strings.Contains(err.Error(), vterrors.NoReplicationStatusStr):
```
deepthi, line 187: @harshit-gangal @vmg this is the line Manan is referring to in his question.
We seem to be already calling `FromGRPC` and `ToGRPC` in several places so that is not a concern.

go/test/endtoend/reparent/emergencyreparent/ers_test.go
GuptaManan100: fixed

go/vt/vterrors/constants.go
GuptaManan100: fixed

vmg:
@GuptaManan100 upon further thought, I think this is an important enough error case that we should add a new error code (while ensuring that it doesn't collide with MySQL's error codes). There's enough gaps in the MySQL error codes space to add a section of Vitess-specific errors -- keeping them close together will allow us to not blur the difference between MySQL errors and ours.

GuptaManan100:
Fixing the bug in ERS has led to a new issue. Consider the following scenario -
1. There are 4 tablets, let's call them tab1, tab2, tab3 and tab4. 
2. Let tab1 be the primary originally.
3. The cluster is working correctly.
4. The user executes ERS requesting tab4 to be the new primary
5. ERS calls `StopReplicationAndBuildStatusMap`. We run the `fillStatus` function for all the 4 tablets simultaneously.
6. tab1, tab2 and tab3 return their results before tab4, maybe due to momentary congestion in the network.
7. We cancel the context as soon as we receive 3 successful tablets. (n-1 check in the wait group).
8. As a result, tab4 is not added to the list of candidates for promotion! 

There doesn't seem to be a way to fix this problem without further deliberation, since we can't wait on all the n tablets too, if we did then if the primary was actually down, we would exhaust all our time waiting for it to return!

GuptaManan100:
Solution for the above problem is to always wait for the newPrimary that the user has asked for and any other n-2 tablets. Proceeding with this change

go/mysql/flavor.go
```diff
@@ -32,7 +32,7 @@ import (
 var (
 	// ErrNotReplica means there is no replication status.
 	// Returned by ShowReplicationStatus().
-	ErrNotReplica = errors.New("no replication status")
+	ErrNotReplica = NewSQLError(ERNotReplica, SSUnknownSQLState, "no replication status")
 
 	// ErrNoPrimaryStatus means no status was returned by ShowPrimaryStatus().
 	ErrNoPrimaryStatus = errors.New("no master status")
```
deepthi, line 38: Should we invent an error code for this and make this a SQLError too?

go/vt/concurrency/error_group.go
```diff
@@ -19,60 +19,80 @@ package concurrency
...
 	NumRequiredSuccesses int
 	NumAllowedErrors     int
+	NumErrorsToWaitFor   int
+}
+
+// Error is used in ErrGroup.Wait function
+// It contains the error that was received along
+// with the information of whether the received error
+// originated from a tablet that we must wait for
+type Error struct {
```
deepthi, line 67: I don't know whether it is good practice to use the same name as a builtin type, but just exported instead of not..
@sougou @vmg WDYT?

go/vt/vtctl/reparentutil/replication.go
```diff
@@ -168,21 +170,27 @@ func StopReplicationAndBuildStatusMaps(
...
 
 		logger.Infof("getting replication position from %v", alias)
 
 		_, stopReplicationStatus, err := tmc.StopReplicationAndGetStatus(groupCtx, tabletInfo.Tablet, replicationdatapb.StopReplicationMode_IOTHREADONLY)
-		switch err {
-		case mysql.ErrNotReplica:
+		sqlErr, isSQLErr := mysql.NewSQLErrorFromError(err).(*mysql.SQLError)
+		switch {
+		case isSQLErr && sqlErr != nil && sqlErr.Number() == mysql.ERNotReplica:
 			var primaryStatus *replicationdatapb.PrimaryStatus
```
deepthi, line 194: I think it will be more readable to rewrite this switch..case as if..else instead.
```
if err != nil {
  // sql error creation and handling
  if isSQLErr && .... {
     ...
  } else {
    // default case goes here
  }
} else {
// nil error case goes here
}
```
WDYT?

go/vt/vtctl/reparentutil/replication.go
```diff
@@ -208,16 +216,26 @@ func StopReplicationAndBuildStatusMaps(
 		}
 	}
 
+	tabletAliasToWaitFor := ""
+	numErrosToWaitFor := 0
```
deepthi, line 220:
```suggestion
	numErrorsToWaitFor := 0
```

go/vt/wrangler/testlib/emergency_reparent_shard_test.go
```diff
@@ -244,22 +235,12 @@ func TestEmergencyReparentShardPrimaryElectNotBest(t *testing.T) {
 	defer newPrimary.StopActionLoop(t)
 
 	// old primary, will be scrapped
-	oldPrimary.FakeMysqlDaemon.ReplicationStatusError = mysql.ErrNotReplica
+	oldPrimary.FakeMysqlDaemon.ReplicationStatusError = fmt.Errorf("old primary stopped working")
```
deepthi, line 238: 👍

go/vt/vtctl/reparentutil/replication.go
```diff
@@ -208,16 +216,26 @@ func StopReplicationAndBuildStatusMaps(
...
+	if tabletToWaitFor != nil {
+		tabletAliasToWaitFor = topoproto.TabletAliasString(tabletToWaitFor)
```
deepthi, line 222: Contortions because of the different types. Not much we can do about it I guess.

deepthi (commented):
Looks fine overall, but maybe a few small changes needed.
Good work 💯

go/mysql/flavor.go
GuptaManan100: For now, we are only returning that error and not checking for it anywhere which is what the motivation to add the code for the error was in the first place. If you want, I can add it to this error too, but it won't be used anywhere except for returning this error

go/vt/concurrency/error_group.go
GuptaManan100: I thought of this for a while, but couldn't come up with a meaningful name for it. Looking for suggestions. The ones that I considered where `SpecialError`, `ErrorGroupError`, `MustWaitForError`. None of these alternatives seemed better than just `Error`

go/vt/vtctl/reparentutil/replication.go
GuptaManan100: I am fine it being either way. If you think this is more readable, I am making this change

go/vt/vtctl/reparentutil/replication.go
GuptaManan100: Unfortunately, we can't. I tried sending a string into the function instead of the tabletAlias for the new parameter, but that wasn't helpful, since we would have to do the nil check outside in the `reparentShardLocked` function, which seemed an even worse design choice

go/vt/vtctl/reparentutil/replication.go
GuptaManan100: fixed

go/vt/vtctl/reparentutil/replication.go
GuptaManan100: changed

go/vt/wrangler/testlib/emergency_reparent_shard_test.go
GuptaManan100: The remainder of the test was written with the expectation that the oldPrimary was down, but after the bug fix, the test started failing. So I changed the returned error which is commensurate to the primary failing

go/vt/concurrency/error_group.go
vmg: That's perfectly OK, the builtin type is lowercase `error`.

deepthi (approved):

Issue #9207, opened 2021-11-12
VReplication discrepancy between copy vs catchup phases for binary columns


jakon89:
#### Overview of the Issue

This is somewhat related to https://github.com/vitessio/vitess/pull/7969

In the MoveTables workflow there is a difference in how vreplication handles binary columns with null values that come from different phases:
- `copy` phase: column value is being read without any padding and put into target in that form 
- `catchup`: during reading from binlog, we right pad the value with `0` as implemented here https://github.com/vitessio/vitess/pull/7969

This also breaks VDiff as source vs target rows aren't equal as events from binlog get padded with `0`.

Please see below steps to get a better understanding of the issue:

#### Reproduction Steps

Steps to reproduce this issue, example:

```
# git clone git@github.com:vitessio/vitess.git && cd vitess

git checkout main && git pull

make docker_local

./docker/local/run.sh

vtctlclient -server localhost:15999 ApplySchema -sql "ALTER TABLE customer ADD md5 CHAR(32) COLLATE UTF8MB4_BIN NOT NULL DEFAULT ''" commerce

mysql -h 127.0.0.1 -P 15306 commerce -e "insert into customer (customer_id, email) values (1, 'test1@testing.com')"

mysql -h 127.0.0.1 -P 15306 commerce -e "insert into customer (customer_id, email, md5) values (2, 'test2@testing.com', 'a2a7396a2ff9ce39074aa92d23f39c3b')"

./201_customer_tablets.sh
./202_move_tables.sh

mysql -h 127.0.0.1 -P 15306 commerce -e "insert into customer (customer_id, email) values (3, 'test3@testing.com')"
mysql -h 127.0.0.1 -P 15306 commerce -e "insert into customer (customer_id, email, md5) values (4, 'test4@testing.com', 'a2a7396a2ff9ce39074aa92d23f39c3b')"

vtctlclient -server localhost:15999 ApplyRoutingRules -rules='{"rules": []}'
mysql -h 127.0.0.1 -P 15306 commerce -e "select customer_id, hex(md5) from customer;"
mysql -h 127.0.0.1 -P 15306 customer -e "select customer_id, hex(md5) from customer;"
```

```
vitess@6616c92b3fda:/vt/local$ vtctlclient -server localhost:15999 VDiff customer.commerce2customer
Summary for table corder:
	ProcessedRows: 0
	MatchingRows: 0
	MismatchedRows: 0
	ExtraRowsSource: 0
	ExtraRowsTarget: 0
[...]

PR #9331, opened 2021-12-07
Take MySQL Column Type Into Account in VStreamer

mattlord:
## Description
In https://github.com/vitessio/vitess/pull/7969 we added right side null-byte padding to `BINARY` columns after processing the binlog event in order to match the [MySQL behavior](https://dev.mysql.com/doc/refman/8.0/en/binary-varbinary.html) and value so that we correctly calculate the keyspace IDs for `BINARY` columns in vindex functions like binary_md5 and correctly apply vreplication filters using those columns.

It turns out, however, that row based binlog events make no distinction between a `BINARY(4)` column and a `CHAR(4)` column with a binary collation like `utf8mb4_bin` (you can see a [detailed discussion here](https://github.com/vitessio/vitess/pull/8730)). So after https://github.com/vitessio/vitess/pull/7969 we were also, incorrectly, adding right side null-byte padding to `CHAR` columns with binary collations.

Although we ensured we did not add more padding than the actual column would hold in https://github.com/vitessio/vitess/pull/8730, we really shouldn't be adding any padding at all to `CHAR` columns in order to match the [MySQL behavior](https://dev.mysql.com/doc/refman/8.0/en/char.html) (otherwise we have discrepancies as [described here](https://github.com/vitessio/vitess/issues/9207)). But in order to do this we needed to thread the target MySQL column type info through the vstreamer and RBR binlog event processing components so that we ONLY add the padding to values for actual `BINARY` columns and nothing else (`CHAR` columns being the known problematic case today).

**Now we _only_ add the padding if the _underlying MySQL column on the target is `BINARY`_ and then it’s just bytes, not chars made of up N bytes, so [the subsequent pad trimming based on the charset](https://github.com/vitessio/vitess/pull/8730,) was removed.**

The [manual test case here](https://github.com/vitessio/vitess/issues/9207) also now passes in this branch:
```
vitess@163dba62a8fb:/vt/local$ mysql -h 127.0.0.1 -P 15306 customer -e "select customer_id, hex(md5) from customer;"
+-------------+------------------------------------------------------------------+
| customer_id | hex(md5)                                                         |
+-------------+------------------------------------------------------------------+
|           1 |                                                                  |
|           2 | 6132613733393661326666396365333930373461613932643233663339633362 |
|           3 |                                                                  |
|           4 | 6132613733393661326666396365333930373461613932643233663339633362 |
+-------------+------------------------------------------------------------------+

vitess@163dba62a8fb:/vt/local$ vtctlclient -server localhost:15999 VDiff customer.commerce2customer
Summary for table customer:
	ProcessedRows: 4
	MatchingRows: 4
	MismatchedRows: 0
	ExtraRowsSource: 0
	ExtraRowsTarget: 0
Summary for table corder:
	ProcessedRows: 0
	MatchingRows: 0
	MismatchedRows: 0
	ExtraRowsSource: 0
	ExtraRowsTarget: 0

vitess@163dba62a8fb:/vt/local$ vtgate --version
Version: 13.0.0-SNAPSHOT (Git revision 3ff9fbdad7 branch 'NullBytePaddingFinalFinal') built on Thu Dec  9 04:04:01 UTC 2021 by vitess@09daf1423251 using go1.17 linux/amd6
```

## Related Issue(s)
Fixes: https://github.com/vitessio/vitess/issues/9207

## Checklist
- [x] Should this PR be backported? Yes, to 12.0
- [x] Tests were added


rohit-nayak-ps (approved):
lgtm.
Very nice! An elegant solution to another binlog parser edge case.

deepthi (approved):
Very thorough code comments and nice test case. 💯

Issue #9213, opened 2021-11-12
Shard metrics collector for throttler should be able to config TLS or disabled


sonne5:
#### Feature Description
When we enable throttler on VTTablet, it automatically enables both `Shard` and `Self`. For self, it just use DBA connection pool to query mysql metrics. However, with `Shard` build connection via formatted string: https://github.com/vitessio/vitess/blob/main/go/vt/vttablet/tabletserver/throttle/mysql/probe.go#L85

This is problematic if mysql is bring up with `--require_secure_transport=ON`. As a result:
```
$ curl -s http://localhost:19000/throttler/status | jq .
{
  "Keyspace": "sbtest",
  "Shard": "0",
  "IsLeader": true,
  "IsOpen": true,
  "IsDormant": false,
  "AggregatedMetrics": {
    "mysql/self": {
      "Value": 0.456644
    },
    "mysql/shard": {
      "ClusterName": "shard",
      "Key": {
        "Hostname": "xxxxx",
        "Port": 17000
      },
      "Value": 0,
      "Err": {
        "Number": 3159,
        "Message": "Connections using insecure transport are prohibited while --require_secure_transport=ON."
      }
    }
  },
  "MetricsHealth": {
    "mysql/self": {
      "LastHealthyAt": "2021-11-12T17:04:17.018064082Z",
      "SecondsSinceLastHealthy": 0
    }
  }
```

I think we should either:
- allow adding `tls` in the url connection
- allow disable shard throttler check, especially if `throttle_check_as_check_self` is set

#### Use Case(s)
Any one that's using throttler with `--require_secure_transport=ON` might run into this issue

sonne5:
@aquarapid Could you help and take a look to see if this feature makes sense? If so, mind helping triage this one? Thanks!

mattlord:
Hi @sonne5,

Thank you for the helpful issue! I see a bigger underlying problem here around how these connections are managed and we're discussing that internally. The issue you mentioned here is simply 1 of many potential problems one could encounter.

But in this specific case, I would expect using `-throttle_check_as_check_self` would effectively disable the shard level check (https://vitess.io/docs/reference/features/tablet-throttler/#configuration):
> Use -throttle_check_as_check_self to implicitly reroute any /throttler/check call into /throttler/check-self. This makes sense when the user supplies a custom query, and where the user wishes to throttle writes to the cluster based on the primary tablet’s health, rather than the overall health of the cluster.

I have confirmed this behavior:
```
[...]

PR #9334, opened 2021-12-07
Tablet throttler: check replicas via HTTP rather than via MySQL

shlomi-noach:
## Description

In this PR we override the existing throttler behavior for shard health checks.
A shard is considered healthy if all replicas are lagging within a certain threshold. 

Today, the primary tablet acquires the identities of all shard's tablets, filters the relevant tablets, then records the hostname.port of the **MySQL** servers of those tablets. It continuously polls these MySQL servers directly for replication lag.

In this PR, the primary tablet no longer accesses those MySQL servers directly (the code to do that is still here, but never used). Instead, it accesses the tablet's HTTP interface for `/throttler/check-self`. On each replica tablet, that check already reflects the replication lag status of the tablet. So now the primary tablet acquires lag information indirectly via HTTP requests.

## Related Issue(s)

- https://github.com/vitessio/vitess/issues/9213

## Checklist

### DIFF
--- a/go/vt/vttablet/tabletserver/throttle/throttler.go
+++ b/go/vt/vttablet/tabletserver/throttle/throttler.go
@@ -550,7 +550,7 @@ func (throttler *Throttler) generateTabletHTTPProbeFunction(ctx context.Context,
 		}
 		mySQLThrottleMetric.Value = checkResult.Value
 
-		if checkResult.StatusCode != http.StatusOK {
+		if checkResult.StatusCode == http.StatusInternalServerError {
 			mySQLThrottleMetric.Err = fmt.Errorf("Status code: %d", checkResult.StatusCode)
 		}
 		return mySQLThrottleMetric
--- a/go/vt/vttablet/tabletserver/throttle/throttler.go
+++ b/go/vt/vttablet/tabletserver/throttle/throttler.go
@@ -40,10 +40,10 @@ import (
 
 const (
 	leaderCheckInterval         = 5 * time.Second
-	mysqlCollectInterval        = 100 * time.Millisecond
+	mysqlCollectInterval        = 250 * time.Millisecond
 	mysqlDormantCollectInterval = 5 * time.Second
 	mysqlRefreshInterval        = 10 * time.Second
-	mysqlAggregateInterval      = 100 * time.Millisecond
+	mysqlAggregateInterval      = 125 * time.Millisecond
 
 	aggregatedMetricsExpiration   = 5 * time.Second
 	aggregatedMetricsCleanup      = 10 * time.Second
--- a/go/vt/vttablet/tabletserver/throttle/throttler.go
+++ b/go/vt/vttablet/tabletserver/throttle/throttler.go
@@ -8,8 +8,10 @@ package throttle
 
 import (
 	"context"
+	"encoding/json"
 	"flag"
 	"fmt"
+	"io"
 	"math"
 	"math/rand"
 	"net/http"
@@ -83,7 +85,7 @@ var (
 )
 
 // ThrottleCheckType allows a client to indicate what type of check it wants to issue. See available types below.
-type ThrottleCheckType int
+type ThrottleCheckType int // nolint:revive
 
 const (
 	// ThrottleCheckPrimaryWrite indicates a check before making a write on a primary server
@@ -186,7 +188,7 @@ func NewThrottler(env tabletenv.Env, ts *topo.Server, tabletTypeFunc func() topo
 		throttler.tickers = [](*timer.SuspendableTicker){}
 		throttler.nonLowPriorityAppRequestsThrottled = cache.New(nonDeprioritizedAppMapExpiration, nonDeprioritizedAppMapInterval)
 
-		throttler.httpClient = base.SetupHTTPClient(0)
+		throttler.httpClient = base.SetupHTTPClient(2 * mysqlCollectInterval)
 		throttler.initThrottleTabletTypes()
 		throttler.ThrottleApp("abusing-app", time.Now().Add(time.Hour*24*365*10), defaultThrottleRatio)
 		throttler.check = NewThrottlerCheck(throttler)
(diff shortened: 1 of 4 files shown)


go/vt/vttablet/tabletserver/throttle/throttler.go
```diff
@@ -518,6 +520,43 @@ func (throttler *Throttler) Operate(ctx context.Context) {
...
+		// connection to mysql and run a query
+		return nil
+	}
+	return func() *mysql.MySQLThrottleMetric {
+		// Hit a tablet's `check-self` via HTTP, and convert its CheckResult JSON output into a MySQLThrottleMetric
+		mySQLThrottleMetric := mysql.NewMySQLThrottleMetric()
+		mySQLThrottleMetric.ClusterName = clusterName
+		mySQLThrottleMetric.Key = probe.Key
+
+		tabletCheckSelfURL := fmt.Sprintf("http://%s:%d/throttler/check-self?app=vitess", probe.TabletHost, probe.TabletPort)
```
shlomi-noach, line 535: Can a tablet serve `https`? Can it expect `TLS`? I'm sense this `http://` will break sometime.

shlomi-noach:
Assuming this PR is accepted and goes well, we will follow up with another PR to remove existing direct-mysql-access code, and probably refactor a bit as cleanup.

go/vt/vttablet/tabletserver/throttle/throttler.go
mattlord: It could support https but it doesn't today. I don't see `ListenAndServeTLS` or `ServeTLS` used anywhere today in the codebase ([except for orchestrator](https://github.com/vitessio/vitess/search?q=ListenAndServeTLS)). It's a valid point though — at some point I think it's worth doing an "https everywhere" project that at least adds support for using it everywhere and makes the protocol used a variable when we form it like this, which we do in many other places already:
```
$ grep -R "http://" go/vt | grep -Ev "apache|colorbar|test.go|rice|Ref|stack|blogs|Advert|foo|orc|refman|google"
go/vt/vtgate/executor.go:	resp, err := client.Get(fmt.Sprintf("http://%s/throttler/check?app=vtgate", tabletHostPort))
go/vt/vtgate/status.go:    <td><a href="http://{{$status.Addr}}">{{$status.Name}}</a></td>
go/vt/vttablet/onlineddl/executor.go:	curl --max-time 10 -s 'http://localhost:%d/schema-migration/report-status?uuid=%s&status=%s&dryrun='"$GH_OST_DRY_RUN"'&progress='"$GH_OST_PROGRESS"'&eta='"$GH_OST_ETA_SECONDS"'&rowscopied='"$GH_OST_COPIED_ROWS"
go/vt/vttablet/onlineddl/executor.go:			fmt.Sprintf(`--throttle-http=http://localhost:%d/throttler/check?app=online-ddl:gh-ost:%s&p=low`, *servenv.Port, onlineDDL.UUID),
go/vt/vttablet/onlineddl/executor.go:	  get("http://localhost:{{VTTABLET_PORT}}/schema-migration/report-status?uuid={{MIGRATION_UUID}}&status={{OnlineDDLStatusRunning}}&dryrun={{DRYRUN}}");
go/vt/vttablet/onlineddl/executor.go:	    get("http://localhost:{{VTTABLET_PORT}}/schema-migration/report-status?uuid={{MIGRATION_UUID}}&status={{OnlineDDLStatusComplete}}&dryrun={{DRYRUN}}");
go/vt/vttablet/onlineddl/executor.go:	    get("http://localhost:{{VTTABLET_PORT}}/schema-migration/report-status?uuid={{MIGRATION_UUID}}&status={{OnlineDDLStatusFailed}}&dryrun={{DRYRUN}}");
go/vt/vttablet/onlineddl/executor.go:			if (head("http://localhost:{{VTTABLET_PORT}}/throttler/check?app=online-ddl:pt-osc:{{MIGRATION_UUID}}&p=low")) {
go/vt/vttablet/endtoend/framework/server.go:	ServerAddress = fmt.Sprintf("http://%s", ln.Addr().String())
go/vt/wrangler/keyspace.go:		webURL = fmt.Sprintf("http://%v:%d/", ts.Tablet.Hostname, webPort)
go/vt/wrangler/version.go:	resp, err := http.Get("http://" + tabletAddr + "/debug/vars")
go/vt/vtctld/api.go:		tablet.URL = "http://" + netutil.JoinHostPort(t.Hostname, t.PortMap["vt"])
go/vt/discovery/legacy_healthcheck.go:// http://{{.GetTabletHostPort}} -> http://host.dc.domain:22
go/vt/discovery/healthcheck.go:	TabletURLTemplateString = flag.String("tablet_url_template", "http://{{.GetTabletHostPort}}", "format string describing debug tablet url formatting. See the Go code for getTabletDebugURL() how to customize this.")
go/vt/discovery/tablet_health.go:// http://{{.GetTabletHostPort}} -> http://host.dc.domain:22
go/vt/vitessdriver/doc.go:For more information, visit http://www.vitess.io.
go/vt/vttest/vtprocess.go:	url := fmt.Sprintf("http://%s/debug/vars", addr)
go/vt/vtadmin/README.md:# "http://127.0.0.1:14200".
```

go/vt/vttablet/tabletserver/throttle/throttler.go
```diff
@@ -518,6 +520,43 @@ func (throttler *Throttler) Operate(ctx context.Context) {
...
+		resp, err := throttler.httpClient.Get(tabletCheckSelfURL)
```
mattlord, line 536: Should we have a timeout? Maybe interval * 0.5 or something?

go/vt/vttablet/tabletserver/throttle/throttler.go
```diff
@@ -518,6 +520,43 @@ func (throttler *Throttler) Operate(ctx context.Context) {
...
+			return mySQLThrottleMetric
+		}
+		checkResult := &CheckResult{}
+		if err := json.Unmarshal(b, checkResult); err != nil {
+			mySQLThrottleMetric.Err = err
+			return mySQLThrottleMetric
+		}
+		mySQLThrottleMetric.Value = checkResult.Value
+
+		if checkResult.StatusCode == http.StatusInternalServerError {
```
mattlord, line 553: Would we ever get here? I would think we would have gone into one of the earlier err blocks. Doesn't hurt, but if we do handle http codes we could handle any of the err codes, 4XX and 5XX or even anything not 2XX.

mattlord (approved):
Had a couple of minor questions but nothing blocking in any way. 

Thank you for working on this!

go/vt/vttablet/tabletserver/throttle/throttler.go
shlomi-noach: The timeout is here:

https://github.com/vitessio/vitess/pull/9334/files#diff-aad1c740b216e391106ae8f24a0b47c524a4ee8c7c9d92e1936325712e26c1b2R191

go/vt/vttablet/tabletserver/throttle/throttler.go
shlomi-noach: Probably 5XX would be handled earlier and still I'd want to surface that. As for the other error codes, these are app-level codes that are generated by the underlying check. There is no need to propagate the error to the upper level, it's enough that we propagate the `Value`. In fact, it hurts if we report back an error, where there is no error per-se, just an app-level decision that "this check is a no-go". It's the upper layer, the collector's responsibility to collect all values and decide if the total state is "go" or "no-go".
The introduction of this mid HTTP layer complicates things because we now only have implicit/secondary understanding of the nature of errors on the MySQL side.
As it stands, the current logic passes all the tests correctly, which means it complies with existing behavior.

deepthi (approved):
I was wondering about the performance impact, but convinced myself that it should be fine.

shlomi-noach:
> I was wondering about the performance impact, but convinced myself that it should be fine.

Now you made me worried. There is indeed a difference between a persistent MySQL connection and a non-persistent HTTP connection. I am also concerned about the impact. I'm going to increase poll interval from `100ms` to `250ms` ie reduce frequency by `x2.5`. Pushing that commit then merging.

Issue #9314, opened 2021-12-01
materialize: query plan error when referencing same column multiple times


derekperkins:
I'm trying to run a materialize command and getting this error: `wrong number of fields: got 9 fields for 12 bind locations`. I’m selecting 9 unique fields from the source table. 1 of those is referenced 2 times and another 3 times, for 12 total references, which appears to be the error. The raw query runs just fine against vtgate, but fails when executed from `Materialize`. I'm sure it's much easier to reproduce than I have here, this is just the exact setup where it failed.

https://github.com/vitessio/vitess/blob/ccdb44645c0aa87a0ff96350c0cc364603b1139d/go/vt/sqlparser/parsed_query.go#L88-L98

### Raw query
```sql
select
   event_id as id,
   user_id as keyspace_id,
   unix_timestamp(created_at)*1e9 as time_next,
   unix_timestamp(created_at)*1e9 as time_scheduled,
   unix_timestamp(created_at)*1e9 as time_created,
   if(workspace_id=0,null,workspace_id) as workspace_id,
   team_id,
   kind,
   kind_id,
   kind_version_id,
   type
from iam_events
```

### Materialize command
```sh
vtctlclient -server localhost:15999 Materialize '{
  "workflow": "iam__invite_email__msgs_test",
  "source_keyspace": "iam",
  "target_keyspace": "iam",
  "table_settings": [
    {
      "target_table": "iam__invite_email__msgs_test",
      "source_expression": "select event_id as id, user_id as keyspace_id, unix_timestamp(created_at)*1e9 as time_next, unix_timestamp(created_at)*1e9 as time_scheduled, unix_timestamp(created_at)*1e9 as time_created, if(workspace_id=0,null,workspace_id) as workspace_id, team_id, kind, kind_id, kind_version_id, type from iam_events"
    }
  ],
  "tablet_types": "PRIMARY"
}'
```

### Source table

```sql
CREATE TABLE `iam_events` (
  `event_id` bigint NOT NULL,
  `workspace_id` bigint NOT NULL,
  `team_id` bigint DEFAULT NULL,
  `kind` varbinary(15) NOT NULL COMMENT 'string representation of a crud object name: workspace, team, etc.',
  `kind_id` bigint NOT NULL COMMENT 'the id for the kind - used for joins back to the original object',
  `kind_version_id` bigint NOT NULL COMMENT 'the version id for the kind - used for joins back to the original object',
[...]

PR #9335, opened 2021-12-07
Allow source columns to be specified multiple times in a vreplication rule filter

rohit-nayak-ps:
## Description
This PR fixes a bug where a column cannot be referred to multiple times in a Filter. There was an incorrect optimization in vstreamer where a column was being ignored while creating the source query if it was already seen.

So a materialize `source_expression` like `select id, val, ts, dayofmonth(ts) as day from mat` would fail since `ts` was mentioned twice in the select list. 

In addition, a new test has been added that confirms that Materialize filters can contain user-defined functions: we can define a `source_expression` like `select id, val, ts, customFunc1(id, val) as x from mat`. This is more an observation rather than a new feature, since, to our query parser, a custom function looks like just a pre-defined mysql function.  It is being stated explicitly since we had earlier assumed we did not yet support custom functions because a user had earlier reported failures with such a filter. It now looks as if that report was caused by this same bug with multiple columns. 

## Related Issue(s)
#9314 

## Checklist

### DIFF
--- a/go/test/endtoend/vreplication/materialize_test.go
+++ b/go/test/endtoend/vreplication/materialize_test.go
@@ -96,3 +96,136 @@ func TestShardedMaterialize(t *testing.T) {
 	validateQuery(t, vtgateConn, "ks2:-", "select id, val from tx",
 		`[[INT64(3) VARBINARY("def")] [INT64(5) VARBINARY("def")]]`)
 }
+
+/*
+ * The following sections are related to TestMaterialize. TestMaterialize is intended to test some edge cases. Currently
+ * it tests
+ * - the case where the same column is referred to multiple times
+ * - use of mysql functions in the filter
+ * - use of a custom function in the filter
+ */
+
+const smMaterializeSchemaSource = `
+	CREATE TABLE mat (
+	id bigint NOT NULL,
+	val varbinary(10) NOT NULL,
+	ts timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+	PRIMARY KEY (id)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci;
+`
+
+const smMaterializeVSchemaSource = `
+{
+  "sharded": true,
+  "tables": {
+      "mat": {
+          "column_vindexes": [
+              {
+                  "column": "id",
+                  "name": "hash"
+              }
+          ]
+      }
+  },
+  "vindexes": {
+      "hash": {
+          "type": "hash"
+      }
+  }
+}
+`
+const smMaterializeSchemaTarget = `
+	CREATE TABLE mat2 (
+	id bigint NOT NULL,
+	val varbinary(10) NOT NULL,
+	ts timestamp NOT NULL,
+	day int NOT NULL,
+	month int NOT NULL,
+    x int not null,
+	PRIMARY KEY (id)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci;
+`
+
+const smMaterializeVSchemaTarget = `
+{
+  "sharded": true,
+  "tables": {
+      "mat2": {
+          "column_vindexes": [
+              {
+                  "column": "id",
+                  "name": "hash"
+              }
+          ]
+      }
+  },
+  "vindexes": {
+      "hash": {
+          "type": "hash"
+      }
+  }
+}
+`
+const smMaterializeSpec2 = `{"workflow": "wf1", "source_keyspace": "ks1", "target_keyspace": "ks2", "table_settings": [ {"target_table": "mat2", "source_expression": "select id, val, ts, dayofmonth(ts) as day, month(ts) as month, custom1(id, val) as x from mat"  }] }`
+
+const materializeInitDataQuery = `insert into ks1.mat(id, val, ts) values (1, 'abc', '2021-10-9 16:17:36'), (2, 'def', '2021-11-10 16:17:36')`
+
+const customFunc = `
+CREATE FUNCTION custom1(id int, val varbinary(10))
+RETURNS int
+DETERMINISTIC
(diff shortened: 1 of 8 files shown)


deepthi:
There was some discussion offline. table_plan_builder is probably not the right place for this, we should be handling it on the other end (vcopier/vplayer).

rohit-nayak-ps:
> There was some discussion offline. `tablePlanBuilder` is probably not the right place for this, we should be handling it on the other end (vcopier/vplayer).

Iiirc we had discussed this in Open Hour, but decided to go with the current implementation: moving it to vplayer/vcopier will require a bigger refactor. Currently the target expects that vstreamer is sending *all* the columns that the vplayer/vcopier need. Changing that will touch a few different parts of code. The current approach has minimal code impact and will fix the existing bug (albeit not optimally). We will address the optimization in a future PR.

go/test/endtoend/vreplication/materialize_test.go
```diff
@@ -96,3 +96,144 @@ func TestShardedMaterialize(t *testing.T) {
...
+	defaultReplicas = 0 // because of CI resource constraints we can only run this test with primary tablets
+	defer func() { defaultReplicas = 1 }()
+
+	defer vc.TearDown(t)
+
+	defaultCell = vc.Cells[defaultCellName]
+	vc.AddKeyspace(t, []*Cell{defaultCell}, sourceKs, "-", smMaterializeVSchemaSource, smMaterializeSchemaSource, defaultReplicas, defaultRdonly, 300)
+	vtgate = defaultCell.Vtgates[0]
+	require.NotNil(t, vtgate)
+	vtgate.WaitForStatusOfTabletInShard(fmt.Sprintf("%s.%s.primary", sourceKs, "0"), 1)
```
mattlord, line 198: They seem to be synonyms but is there any reason to use "-" above and "0" here?

go/test/endtoend/vreplication/materialize_test.go
```diff
@@ -96,3 +96,144 @@ func TestShardedMaterialize(t *testing.T) {
...
+
+	vtgateConn = getConnection(t, vc.ClusterConfig.hostname, vc.ClusterConfig.vtgateMySQLPort)
+	defer vtgateConn.Close()
+	verifyClusterHealth(t, vc)
+
+	_, err := vtgateConn.ExecuteFetch(materializeInitDataQuery, 0, false)
+	require.NoError(t, err)
+
+	ks2Keyspace := vc.Cells[defaultCell.Name].Keyspaces[targetKs]
+	ks2Primary := ks2Keyspace.Shards["-"].Tablets["zone1-400"].Vttablet
```
mattlord, line 211: I think we can do this instead of relying on a hardcoded tablet UID: 
```
ks2Primary := vc.getVttabletsInKeyspace(t, defaultCellName, targetKs, "primary")
```

go/test/endtoend/vreplication/materialize_test.go
rohit-nayak-ps: We had a regression (a long while ago) where '-' and '0' were giving different results which wasn't caught by the tests at that time! So as a reflex  I have used the two interchangeably at times, but it is not strictly necessary for this test.

go/vt/vttablet/tabletmanager/vreplication/vcopier_test.go
```diff
@@ -464,7 +464,7 @@ func TestPlayerCopyTables(t *testing.T) {
 	filter := &binlogdatapb.Filter{
 		Rules: []*binlogdatapb.Rule{{
 			Match:  "dst1",
-			Filter: "select * from src1",
+			Filter: "select id, val, val as val2 from src1",
```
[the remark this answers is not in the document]
rohit-nayak-ps: `val2` is an extra column in the target. source does not have it. The filter requests `val` twice, second time as `val2` to test that the same column can be referred to multiple times. The sql generated on the target uses the field name as the column name so it needs to match the column name on the target.

go/test/endtoend/vreplication/materialize_test.go
rohit-nayak-ps: `getVttabletsInKeyspace` returns a map with tablet name as the key. In this case we get a map with a single entry since we only have one shard and hence only one primary, but in general we could have multiple entries for each shard. I added another helper `getPrimaryTablet(keyspace, shard)` for this purpose.

go/test/endtoend/vreplication/materialize_test.go
```diff
@@ -61,7 +61,7 @@ const smMaterializeSpec = `{"workflow": "wf1", "source_keyspace": "ks1", "target
 const initDataQuery = `insert into ks1.tx(id, typ, val) values (1, 1, 'abc'), (2, 1, 'def'), (3, 2, 'def'), (4, 2, 'abc'), (5, 3, 'def'), (6, 3, 'abc')`
 
 // TestShardedMaterialize tests a materialize from a sharded (single shard) using comparison filters
```
deepthi, line 63: nit: comment should be changed to reflect function name change.

go/test/endtoend/vreplication/materialize_test.go
```diff
@@ -89,10 +89,150 @@ func TestShardedMaterialize(t *testing.T) {
...
+const materializeInitDataQuery = `insert into mat(id, val, ts) values (1, 'abc', '2021-10-9 16:17:36'), (2, 'def', '2021-11-10 16:17:36')`
+
+const customFunc = `
+CREATE FUNCTION custom1(id int, val varbinary(10))
+RETURNS int
+DETERMINISTIC
+RETURN id * length(val);
+`
+
+// TestMaterialize: details mentioned above
```
deepthi, line 180: comment should be changed: Test -> test

deepthi: or can possibly be deleted altogether.

deepthi (approved):
LGTM.
It seems like all of Matt's feedback has been addressed.

PR #9337, opened 2021-12-08
Throttler: cleaning up legacy remote MySQL connection logic

shlomi-noach:
## Description

Followup to #9334 
Since #9334 is merged, the throttler no longer directly accesses MySQL on other tablets. As such, there's no longer need in:

- A special throttler user account 
- Any user/password configuration
- logic to open a remote connection
- any special logic upon promotion to leader (other than generate a Tick)
This PR cleans up all the above with minor refactoring/changes.

## Related Issue(s)

#9334 

## Checklist


mattlord (approved):
Thank you! 🥇

deepthi (approved):
💯

PR #9340, opened 2021-12-08
[vtadmin-web] Add colours to tailwind.config.js

doeg:

## Description

This extends tailwind's colour palette with our own, which handily generates a bunch of utility classes for us for background colour, etc. (Eventually, we may want to replace tailwind's default palette entirely, for simplicity... but this is fine for now.) 

I also renamed "vitess-blue" to "vtblue" since it saves a few characters. :) 

At some point I'll do a pass to update everything that's presently using bespoke CSS to instead use the generated tailwind utility classes.

Screen Shot 2021-12-08 at 12 17 54 PM

Screen Shot 2021-12-08 at 12 17 58 PM

## Related Issue(s)

N/A

## Checklist

## Deployment Notes

N/A


notfelineit (approved):
This is so exciting!!

PR #9342, opened 2021-12-08
[vtadmin-web] Add .btn class + remove Button component

doeg:
## Description

This :hocho:s the button component and instead replaces it with a Tailwind [component class](https://tailwindcss.com/docs/extracting-components). One day we might (?) once again need a full-fledged Button component, but for now I _really_ like this simplification. 

It's a good time to do this clean-up since I can no longer avoid adding a link-that-looks-like-a-button "component". :) With the previous Button component implementation, I was debating whether I should conflate buttons and link-buttons in the Button class (bad) or adding a separate LinkButton component (also cumbersome). This way is sooo much easier.

I also added a whole bunch of colour variants, since we get all those utility classes for free after merging in #9340. 🌈 ✨ 

Screen Shot 2021-12-08 at 2 49 43 PM

Here's before, for comparison. Not sure when the alignment got so busted, haha. Oh well, fixed now. 

Screen Shot 2021-12-08 at 2 41 17 PM

## Related Issue(s)

- Requires #9340

## Checklist

## Deployment Notes

N/A


web/vtadmin/src/index.css
```diff
@@ -51,8 +54,8 @@
 
     /* Inputs + other form controls */
     --inputHeightSmall: 2.4rem;
-    --inputHeightMedium: 3.6rem;
-    --inputHeightLarge: 4.8rem;
+    --inputHeightMedium: 3.7rem;
+    --inputHeightLarge: 4.6rem;
```
doeg, line 58: These adjustments are a gross little hack to get text inputs + buttons to line up vertically, since the inputs don't use tailwind util classes (yet) and are therefore off by a few pixels. Will revisit the inputs later!

notfelineit (approved):
I'm excited to use these! :-)

PR #9351, opened 2021-12-10
Move more code to use evalengine

systay:
## Description
`sqltypes.PlanValue` has reached the end of its usefulness, and is being replaced by `evalengine`.
This PR updates engine primitives to use evalengine.


go/vt/vtgate/evalengine/convert.go
```diff
@@ -350,6 +350,23 @@ func convertExpr(e sqlparser.Expr, lookup ConverterLookup) (Expr, error) {
...
+		}
+		lit, ok := expr.(*Literal)
+		if !ok {
+			return nil, vterrors.Errorf(vtrpcpb.Code_INVALID_ARGUMENT, "character set introducers are only supported for literals")
+		}
+		coll := collations.Local().DefaultCollationForCharset(node.CharacterSet[1:])
+		if coll == nil {
+			return nil, vterrors.Errorf(vtrpcpb.Code_INTERNAL, "unknown character set")
+		}
+		lit.Val.collation = collations.TypedCollation{
```
vmg, line 366: You should not override the whole `TypedCollation`, just assign to `lit.Val.collation.Collation` so it keeps the original coercibility. :ok_hand:

vmg (changes requested):

vmg (approved):

go/vt/vtgate/engine/limit.go
```diff
@@ -162,58 +154,56 @@ func (l *Limit) Inputs() []Primitive {
...
 
-	if !l.Count.IsNull() {
-		other["Count"] = l.Count.Value
+	if l.Count != nil {
+		other["Count"] = evalengine.FormatExpr(l.Count)
 	}
-	if !l.Offset.IsNull() {
-		other["Offset"] = l.Offset.Value
+	if l.Offset != nil {
+		other["Offset"] = evalengine.FormatExpr(l.Offset)
```
frouioui, line 206: It would be nicer to print the raw value instead of printing type+value, like: print `1` instead of `INT64(1)`

frouioui (approved):
awesome work, this is really exciting

harshit-gangal (approved):

Issue #9356, opened 2021-12-10
Buffering concerns with vtgate


FancyFane:
### Overview of the Issue

The buffering behavior of vtgate during a PRS event is not consistent; and only functions when specific traffic conditions are met.

Problem Scenarios:
1.) With the right amount and type of traffic buffering will work, however, it will slow down the PRS process.
2.) Too few request hitting vtgate causes the buffering to not trigger; resulting in dropped SQL statements.
3.) Too many request hitting vtgate causes the buffering to timeout; the PRS fails.

End Goal:
To get consistent buffering results when issuing a PRS on a buffered vtgate. 

### Reproduction Steps
To create traffic I used the [gateslap](https://github.com/FancyFane/gateslap) utility. This allowed control of  the type of connection: oneoff, persistent, pooled; and allowed for the control of how much and what type of traffic to send.

```sh
git clone git@github.com:FancyFane/gateslap.git
cd gateslap
virtualenv venv
source venv/bin/activate
python3 setup.py install
```

The `gateslap` utility is leveraging `mysqlslap` to generate semi-random SQL files, if you do not have it installed you will have to install a mysql client (which includes the mysqlslap utility). 

```sh
sudo apt-get install mysql-client
```

#### Scenario 1

1.) Bring up the vtgate with buffering flags enabled (my vtgate flags below):
```sh
jason@melvin:~$ vtgate -topo_implementation etcd2 -topo_global_server_address 192.168.1.36:2379 -topo_global_root /vitess/global -log_dir=/home/jason/vt_data/logs -log_queries_to_file=/home/jason/vt_data/logs/vtgate_querylog.txt -port=15001 -grpc_port=15991 -mysql_server_port=15306 -cell=zone1 -cells_to_watch=zone1 -tablet_types_to_wait=MASTER,REPLICA -service_map=grpc-vtgateservice -pid_file=/home/jason/vt_data/logs/vtgate.15306.pid -mysql_auth_server_impl=none -gateway_initial_tablet_timeout=5s -enable_buffer=true -buffer_size=2000
```

2.) Create Load and perform a PRS 

2.a) In the terminal window where the gateslap utility is located (you must be in the gateslap directory) run the gateslap command:
```sh
(venv) jason@melvin:~/Github/gateslap$ gateslap
```

2.b) In a second terminal window right after the first loading bar completes, issue a PRS command
```sh
[...]

Issue #9346, opened 2021-12-09
vtgate does not buffer queries when primary is unhealthy and awaiting a primary failover


tslo-stripe:
#### Overview of the Issue

Vitess vtgate has an option (-enable_buffer) which enable buffering queries during a primary failover.  During testing, I found that when the primary vttablet becomes unhealthy and before the primary has failed over to another tablet, queries are not buffer and number of query failures did not drop as expected .  Looking at the code, it looks like we skip buffering when no healthy primary is found - https://github.com/vitessio/vitess/blob/main/go/vt/vtgate/tabletgateway.go#L288.

Is this expected?  Is it possible to buffer queries in vtgate when the primary becomes unhealthy and primary failover is pending?

#### Reproduction Steps

1. Enable buffering by setting -enable_buffer in vtgate
2. Start generating load to a shard in vitess through the vtgate with buffer enabled
3. Restart the primary tablet of that shard
4. Observe is queries are getting buffered and note the number of query failures.
5. Repeat steps 2-4 with buffer disabled (leave enable_buffer unset) and compare the results.

GuptaManan100:
Hello @tslo-stripe, the buffering only happens for failover scenarios. So the buffering will happen if `PlannedReparentShard` is called. The buffering does not happen when the primary tablet is restarted for now.

There is also a new implementation of buffering, controlled by the flag `buffer_implementation`. The new implementation uses `keyspace_events` and it also buffers when a resharding operation is in progress. There was also a bug fix https://github.com/vitessio/vitess/pull/9112 that went into the newer implementation for buffering while PRS was in progress.

GuptaManan100:
cc @deepthi @harshit-gangal @vmg

vmg:
Yes, you should try the new keyspace events watcher: `-buffer_implementation=keyspace_events`. It has heuristics to detect more cluster states where buffering is safe to perform, including some cases where the primary is down.

tslo-stripe:
Thanks.  It sounds like the reason is that even though vttablet is restarted and reparent will happen soon, we do not buffer requests as we are not sure if it is safe to do so and will rely on client to restart, is that correct?

I saw the buffer_implementation flag and it says "for testing" for "keyspace_events", so I was not sure if it is safe to be used in prod environment.

vmg:
[...]

PR #9360, opened 2021-12-11
Change default buffer implementation #9359

### LINKED ISSUE #9359: Change default Vitess buffering implementation
#### Feature Description

The current default buffering implementation leads to inconsistent behavior as seen in github issues #9356 and #9346. We may want to change the default buffering implementation to use `keyspace events` which appears to be more reliable. Currently, this has to be set manually using the flag `-buffer_implementation=keyspace_events`.

FancyFane:
Signed-off-by: FancyFane <fane@planetscale.com>

## Description
The current buffer implementation leads to inconsistent behavior during testing. This pull request purposes changing the default buffering implementation after seeing these inconsistencies resolved with the `keyspace_events` implementation. 

## Related Issue(s)
Fixes #9359 #9346 #9356 

## Checklist

## Deployment Notes
n/a

### DIFF
--- a/doc/releasenotes/13_0_0_summary.md
+++ b/doc/releasenotes/13_0_0_summary.md
@@ -9,6 +9,9 @@ The default value used to be false. What this means is that during a failover, w
 In addition, all Vitess-managed databases will be started with `super-read-only` in the cnf file.
 It is expected that this change is safe and backwards-compatible. Anyone who is relying on the current behavior should pass `-use_super_read_only=false` on the vttablet command line, and make sure they are using a custom my.cnf instead of the one provided as the default by Vitess.
 
+### vtgate -buffer_implementation now defaults to keyspace_events
+The default value used to be `healthcheck`. The new `keyspace_events` implementation has been tested in production with good results and shows more consistent buffering behavior during PlannedReparentShard operations. The `keyspace_events` implementation utilizes heuristics to detect additional cluster states where buffering is safe to perform, including cases where the primary may be down. If there is a need to revert back to the previous buffer implementation, ensure buffering is enabled in vtgate and pass the flag `-buffer_implementation=healthcheck`. 
+
 ### ddl_strategy: -postpone-completion flag
 
 `ddl_strategy` (either `@@ddl_strategy` in VtGate or `-ddl_strategy` in `vtctlclient ApplySchema`) supports the flag `-postpone-completion`
--- a/go/vt/vtgate/buffer/buffer_test.go
+++ b/go/vt/vtgate/buffer/buffer_test.go
@@ -271,15 +271,15 @@ func testPassthrough1(t *testing.T, fail failover) {
 	}
 }
 
-// TestLastReparentTooRecent_BufferingSkipped tests that buffering is skipped if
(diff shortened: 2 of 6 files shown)


vmg:
@deepthi: cool with this? I think everybody who's tried the new implementation has seen good results

@FancyFane: a couple of unit tests were expecting the old implementation to be the default and are upset now. Can you give them a look?

GuptaManan100:
I personally think, this should be safe. We eventually wanted to make `keyspace_events` the default implementation anyways. Also it was experimental in the previous release, so it should be okay to make it GA for release-13

deepthi:
> @deepthi: cool with this? I think everybody who's tried the new implementation has seen good results
> 
:+1:

deepthi:
@FancyFane can you add a release note into this PR? They go into https://github.com/vitessio/vitess/blob/main/doc/releasenotes/13_0_0_summary.md

mattlord:
@vmg it seems that we only implement `WatchSrvKeyspace` in the keyspace filtering server currently: https://github.com/vitessio/vitess/blob/main/go/vt/srvtopo/keyspace_filtering_server.go#L115

Meaning that the buffering will not work unless you specify `-keyspaces_to_watch`, no? Which in turn makes it more involved to add keyspaces in your cluster and the keyspace filtering server may have other limitations beyond the one I recently addressed in https://github.com/vitessio/vitess/pull/8988.

Or am I misunderstanding something? I just want to raise this as one potential reason we may not want to change the default.

FancyFane:
I've added the configuration change for buffer_size moving from 10 to 1000. The go linter also asked that we rename the functions to remove "_" in a related file, so those changes have been included as well.

doc/releasenotes/13_0_0_summary.md
```diff
@@ -4,6 +4,14 @@
...
+The default value used to be false. What this means is that during a failover, we will set `super_read_only` on database flavors that support them (MySQL 5.7+ and Percona 5.7+).
+In addition, all Vitess-managed databases will be started with `super-read-only` in the cnf file.
+It is expected that this change is safe and backwards-compatible. Anyone who is relying on the current behavior should pass `-use_super_read_only=false` on the vttablet command line, and make sure they are using a custom my.cnf instead of the one provided as the default by Vitess.
+
+### vtgate -buffer_implementation	now defaults to keyspace_events
+The default value used to be `healthcheck`. The keyspace_events implementation has shown more consistent buffering behavior during PlannedReparentShard operations. The keyspace_events implementation also, utilizes heuristics to detect additional cluster states where buffering is safe to perform; including cases where the primary may be down. If there is a need to revert back to the previous buffer implementation, ensure buffering is enabled in vtgate and pass the flag `-buffer_implementation=healthcheck`.
```
deepthi, line 13:
```suggestion
The default value used to be `healthcheck`. The `keyspace_events` implementation has shown more consistent buffering behavior during `PlannedReparentShard` operations. The `keyspace_events` implementation also utilizes heuristics to detect additional cluster states where buffering is safe to perform; including cases where the primary may be down. If there is a need to revert back to the previous buffer implementation, ensure buffering is enabled in vtgate and pass the flag `-buffer_implementation=healthcheck`. 
```

vmg (approved):

PR #9362, opened 2021-12-13
Test: reduce flakiness in reparent tests

GuptaManan100:
## Description

This PR refactors the reparent tests to reduce the flakiness of this testing suite.

## Checklist


GuptaManan100:
There are two changes to reduce flakiness in the tests -
1. Moved 2 tests from `plannedreparent` package to `emergencyreparent`. This was done since the package `plannedreparent` was taking on average about 9 minutes to run. In some cases it would exceed the 10 minute mark and cause the test to fail. The two tests that are moved are - `TestReparentDownPrimary` and `TestReparentNoChoiceDownPrimary` which should have been in the `emergencyreparent` tests in the first place.
2. Online ddl is disabled in these reparent tests by setting the flag `queryserver_enable_online_ddl` to false on the vttablets. Comments explaining why this change reduces flakiness have also been documented. 
```
// disabling online-ddl for reparent tests. This is done to reduce flakiness.
// All the tests in this package reparent frequently between different tablets
// This means that Promoting a tablet to primary is sometimes immediately followed by a DemotePrimary call.
// In this case, the close method and initSchema method of the onlineDDL executor race.
// If the initSchema acquires the lock, then it takes about 30 seconds for it to run during which time the
// DemotePrimary rpc is stalled!
```

deepthi (approved):

PR #9365, opened 2021-12-13
vttablet: Better use of vtprotobuf memory pool

mattrobenolt:
In the previous implementation, the last iteration of the for loop
leaves behind an object to get GC'd since it never gets released back to
the pool when the function returns.

Also should be a minor optimization by re-using the VStreamRowsResponse
rather than returning it to the pool each iteration.

## Checklist


deepthi (approved):
👍

PR #7850, opened 2021-04-13
WIP Running CI workflows on Ubuntu 20.04

shlomi-noach:
## Description

Resubmission of https://github.com/vitessio/vitess/pull/7613 (which went a bit stale). This is an attempt to see whether ubuntu-latest (`20.04` at this time) reduces CI flakiness.
## Checklist
- [ ] Should this PR be backported?
- [ ] Tests were added or are not required
- [ ] Documentation was added or is not required

## Impacted Areas in Vitess
Components that this PR will affect:

- [ ]  Query Serving
- [ ]  VReplication
- [ ]  Cluster Management
- [x]  Build/CI
- [ ]  VTAdmin

PR #9368, opened 2021-12-13
CI: ubuntu-latest now has MySQL 8.0.26, let us override it with latest 8.0.x

deepthi:
## Description

Actions workflows that run on `ubuntu-latest` have been failing for ~12 hours now.
This is because of changes made to the images
https://github.com/actions/virtual-environments/pull/4708
https://github.com/actions/virtual-environments/pull/4674

~Since the image already includes a mysql install, perhaps we can stop trying to install it ourselves.~
Let's try to install the latest 8.0.x version ourselves.

## Checklist

### DIFF
--- a/.github/workflows/cluster_endtoend_mysql80.yml
+++ b/.github/workflows/cluster_endtoend_mysql80.yml
@@ -34,12 +34,33 @@ jobs:
 
     - name: Get dependencies
       run: |
+
+        sudo DEBIAN_FRONTEND="noninteractive" apt-get update
+
+        # Uninstall any previously installed MySQL first
+        sudo systemctl stop apparmor
+        sudo DEBIAN_FRONTEND="noninteractive" apt-get remove -y --purge mysql-server mysql-client mysql-common
+        sudo apt-get -y autoremove
+        sudo apt-get -y autoclean
+        sudo deluser mysql
+        sudo rm -rf /var/lib/mysql
+        sudo rm -rf /etc/mysql
+
+        # Install mysql80
+        wget -c https://dev.mysql.com/get/mysql-apt-config_0.8.20-1_all.deb
+        echo mysql-apt-config mysql-apt-config/select-server select mysql-8.0 | sudo debconf-set-selections
+        sudo DEBIAN_FRONTEND="noninteractive" dpkg -i mysql-apt-config*
         sudo apt-get update
-        sudo apt-get install -y mysql-server mysql-client make unzip g++ etcd curl git wget eatmydata
+        sudo DEBIAN_FRONTEND="noninteractive" apt-get install -y mysql-server mysql-client
+
+        # Install everything else we need, and configure
+        sudo apt-get install -y make unzip g++ etcd curl git wget eatmydata
         sudo service mysql stop
         sudo service etcd stop
+        sudo bash -c "echo '/usr/sbin/mysqld { }' > /etc/apparmor.d/usr.sbin.mysqld" # https://bugs.launchpad.net/ubuntu/+source/mariadb-10.1/+bug/1806263
         sudo ln -s /etc/apparmor.d/usr.sbin.mysqld /etc/apparmor.d/disable/
-        sudo apparmor_parser -R /etc/apparmor.d/usr.sbin.mysqld
+        sudo apparmor_parser -R /etc/apparmor.d/usr.sbin.mysqld || echo "could not remove mysqld profile"
+
         go mod download
 
         # install JUnit report formatter
--- a/.github/workflows/check_make_parser.yml
+++ b/.github/workflows/check_make_parser.yml
@@ -16,23 +16,14 @@ jobs:
       run: |
         echo '1024 65535' | sudo tee -a /proc/sys/net/ipv4/ip_local_port_range
 
-    # TEMPORARY WHILE GITHUB FIXES THIS https://github.com/actions/virtual-environments/issues/3185
-    - name: Add the current IP address, long hostname and short hostname record to /etc/hosts file
-      run: |
-        echo -e "$(ip addr show eth0 | grep "inet\b" | awk '{print $2}' | cut -d/ -f1)\t$(hostname -f) $(hostname -s)" | sudo tee -a /etc/hosts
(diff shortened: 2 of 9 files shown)


mattlord:
I'm OK with this as a temporary workaround but it breaks a reasonable assumption here, that is: you install the latest official MySQL version with `apt-get install`. That's what we normally want to do as we try and regularly upgrade to stay up to date. We do also in some cases alter behavior based on the patch version.

deepthi:
> This is one potential short-term option though, for sure. Others may be:
>
> * Specifying a specific image that is fine, e.g. 18.04 or the specific version latest pointed to before this change
> * Explicitly adding the upstream MySQL APT repository (potential long-term solution if we need it)
> * Pinning and managing usage of a specific MySQL version on our side and then installing via dpkg directly (potential long-term solution if we need it)

I'm reluctant to move everything back to 18.04, we had tried to actually move everything forward to 20.04 at some point but didn't complete that work, see #7850
The others are good options, this was just the fastest, smallest change to get us unblocked.
I can try the second option and see if I can do it correctly :)

.github/workflows/cluster_endtoend_mysql80.yml
```diff
@@ -34,12 +34,33 @@ jobs:
...
+
+        # Uninstall any previously installed MySQL first
+        sudo systemctl stop apparmor
+        sudo DEBIAN_FRONTEND="noninteractive" apt-get remove -y --purge mysql-server mysql-client mysql-common
+        sudo apt-get -y autoremove
+        sudo apt-get -y autoclean
+        sudo deluser mysql
+        sudo rm -rf /var/lib/mysql
+        sudo rm -rf /etc/mysql
+
```
mattlord, line 48: I don't any of this is necessary, is it? Any previous installation and its files don't matter as we use non-default values anyway and this will slow down the CI jobs further.

.github/workflows/cluster_endtoend_mysql80.yml
```diff
@@ -34,12 +34,33 @@ jobs:
...
+        sudo DEBIAN_FRONTEND="noninteractive" dpkg -i mysql-apt-config*
         sudo apt-get update
-        sudo apt-get install -y mysql-server mysql-client make unzip g++ etcd curl git wget eatmydata
+        sudo DEBIAN_FRONTEND="noninteractive" apt-get install -y mysql-server mysql-client
+
+        # Install everything else we need, and configure
+        sudo apt-get install -y make unzip g++ etcd curl git wget eatmydata
         sudo service mysql stop
         sudo service etcd stop
+        sudo bash -c "echo '/usr/sbin/mysqld { }' > /etc/apparmor.d/usr.sbin.mysqld" # https://bugs.launchpad.net/ubuntu/+source/mariadb-10.1/+bug/1806263
```
mattlord, line 60: I would remove this. This was about running MariaDB after MySQL was previously installed -- which is not relevant for us here.

.github/workflows/cluster_endtoend_mysql80.yml
```diff
@@ -34,12 +34,33 @@ jobs:
...
+        sudo rm -rf /var/lib/mysql
+        sudo rm -rf /etc/mysql
+
+        # Install mysql80
+        wget -c https://dev.mysql.com/get/mysql-apt-config_0.8.20-1_all.deb
+        echo mysql-apt-config mysql-apt-config/select-server select mysql-8.0 | sudo debconf-set-selections
+        sudo DEBIAN_FRONTEND="noninteractive" dpkg -i mysql-apt-config*
         sudo apt-get update
-        sudo apt-get install -y mysql-server mysql-client make unzip g++ etcd curl git wget eatmydata
+        sudo DEBIAN_FRONTEND="noninteractive" apt-get install -y mysql-server mysql-client
```
mattlord, line 54: I would remove this and just keep the original apt-get install for all of our packages.

.github/workflows/cluster_endtoend_mysql80.yml
```diff
@@ -34,12 +34,33 @@ jobs:
...
+
+        # Install everything else we need, and configure
+        sudo apt-get install -y make unzip g++ etcd curl git wget eatmydata
         sudo service mysql stop
         sudo service etcd stop
+        sudo bash -c "echo '/usr/sbin/mysqld { }' > /etc/apparmor.d/usr.sbin.mysqld" # https://bugs.launchpad.net/ubuntu/+source/mariadb-10.1/+bug/1806263
         sudo ln -s /etc/apparmor.d/usr.sbin.mysqld /etc/apparmor.d/disable/
-        sudo apparmor_parser -R /etc/apparmor.d/usr.sbin.mysqld
+        sudo apparmor_parser -R /etc/apparmor.d/usr.sbin.mysqld || echo "could not remove mysqld profile"
```
mattlord, line 62: If this step failing is not fatal, then why do it at all?

mattlord (approved):
I had a few comments/nits, but want to unblock you and thus everyone else. :-) 

I can make another commit, or another PR, if you want.

deepthi:
I'm going to merge this as-is to unblock everyone, and @mattlord will do a follow up PR cleaning up how we are installing mysql in some of the tests.

.github/workflows/cluster_endtoend_mysql80.yml
```diff
@@ -34,12 +34,33 @@ jobs:
...
+        sudo DEBIAN_FRONTEND="noninteractive" apt-get update
+
+        # Uninstall any previously installed MySQL first
+        sudo systemctl stop apparmor
+        sudo DEBIAN_FRONTEND="noninteractive" apt-get remove -y --purge mysql-server mysql-client mysql-common
+        sudo apt-get -y autoremove
+        sudo apt-get -y autoclean
+        sudo deluser mysql
+        sudo rm -rf /var/lib/mysql
+        sudo rm -rf /etc/mysql
```
GuptaManan100, line 47: This is a generated file, so we should be changing the code generation and not the file directly. The file that generates this is `ci_workflow_gen.go`

doeg:
Thank you both for looking into this! ❤️

PR #9370, opened 2021-12-14
CI: Simplify Changes to Explicitly Install and Configure MySQL8

mattlord:
## Description
Limiting the new/additional work we do in the CI in order to limit the time and effort required to meet the primary objective: ensure that we deterministically install the latest MySQL 8.0 release from the official MySQL APT repo. 

## Related Issue(s)
Follow-up on: https://github.com/vitessio/vitess/pull/9368
Caused by: https://github.com/actions/virtual-environments/pull/4674

## Checklist
- [x] Should this PR be backported? 12.0 maybe?

### DIFF
--- a/.github/workflows/cluster_endtoend_mysql80.yml
+++ b/.github/workflows/cluster_endtoend_mysql80.yml
@@ -34,33 +34,16 @@ jobs:
 
     - name: Get dependencies
       run: |
-
-        sudo DEBIAN_FRONTEND="noninteractive" apt-get update
-
-        # Uninstall any previously installed MySQL first
-        sudo systemctl stop apparmor
-        sudo DEBIAN_FRONTEND="noninteractive" apt-get remove -y --purge mysql-server mysql-client mysql-common
-        sudo apt-get -y autoremove
-        sudo apt-get -y autoclean
-        sudo deluser mysql
-        sudo rm -rf /var/lib/mysql
-        sudo rm -rf /etc/mysql
-
-        # Install mysql80
+        # Setup MySQL 8.0
         wget -c https://dev.mysql.com/get/mysql-apt-config_0.8.20-1_all.deb
         echo mysql-apt-config mysql-apt-config/select-server select mysql-8.0 | sudo debconf-set-selections
         sudo DEBIAN_FRONTEND="noninteractive" dpkg -i mysql-apt-config*
         sudo apt-get update
-        sudo DEBIAN_FRONTEND="noninteractive" apt-get install -y mysql-server mysql-client
 
         # Install everything else we need, and configure
-        sudo apt-get install -y make unzip g++ etcd curl git wget eatmydata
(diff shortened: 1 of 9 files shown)


.github/workflows/cluster_endtoend_mysql80.yml
```diff
@@ -34,33 +34,16 @@ jobs:
 
     - name: Get dependencies
       run: |
-
```
deepthi, line 37: @GuptaManan100 pointed out that this file is actually generated (see Line 1) using https://github.com/vitessio/vitess/blob/main/test/ci_workflow_gen.go.
I'm not sure what happens to other tests if we change the template that is being used by the `make` target to generate yaml files.
I guess you will need to try and move these changes into the template and if it breaks other tests, then maybe create a new template for mysql80?

mattlord: Yeah, will need to do something there. Running `make generate_ci_workflow` puts `cluster_endtoend_mysql80.yml` back into it's original state where it breaks after the ubuntu-latest image change.

mattlord: Added support here: https://github.com/vitessio/vitess/pull/9370/commits/85494ba94bb67375ce7f810f233f48bc21e88d4b

So now `make generage_ci_workflow` produces what I had previously done in the generated file.

GuptaManan100 (approved):

GuptaManan100 (approved):
LGTM!
We can even remove the ubuntu version check. Like make all the tests run on 18.04. As far as I know, we were only using 20.04 to get the mysql 8.0 pre-installed

test/ci_workflow_gen.go
```diff
@@ -125,7 +127,7 @@ type unitTest struct {
 }
 
 type clusterTest struct {
-	Name, Shard                  string
+	Name, Shard, Platform        string
```
deepthi, line 130: This is elegant.
Only one nit, `Platform` doesn't seem like the right term for database (hat is what we are changing here).
How about using `Flavor` since we use that elsewhere in the code base?

deepthi (approved):

test/ci_workflow_gen.go
mattlord: I had similar thoughts, but:
1. We already use Platform this way (e.g. mysql57 or mysql8) for the unit tests and self hosted [in the same file](https://github.com/vitessio/vitess/blob/9f6bd06b5142ef4faa9b909879a448368b4d1657/test/ci_workflow_gen.go) 
2. We already use OS, and Flavor is used for the GTID type so the Flavor is mysql56 for 5.7 and 8.0 in those contexts

I'm fine changing it to whatever though.

deepthi: alright, you've convinced me :)
If we are already using Platform, it makes sense to keep it consistent.

Issue #9332, opened 2021-12-07
Return the exact keyspace/shard in VEvent


5antelope:
#### Problem
Right now the VStream() endpoint returns VEvent with a [VGtid](https://github.com/vitessio/vitess/blob/main/proto/binlogdata.proto#L380). This filed contains all the target shards for the VStream endpoint. AFAICT, we do this as the checkpoint returns to the client so that they can store it and resume vstream if needed. 

However, this makes it hard identify which shard is the event coming from. We do have shard info inside FieldEvent and RowEvent https://github.com/vitessio/vitess/blob/main/proto/binlogdata.proto#L285-L298 but there are VEvent without those fields set as well.

#### Proposal
Would it make sense to return exact shard info as the top level field in `VEvent` (e.g., `ShardGtid`) so that the client always know which shard the event is coming from (the `VGtid` will be returned as is for checkpoint purpose)? 

This would help Vitess be more compatible with Debezium, where people usually expect a deterministic id for events. 

One might think why now use `VGtid` as the id directly? There are two problems with that: 1. the order of VGtid might changed depends on VStream API; 2. it is a unbounded list that is as long as number of targeted shard.

sonne5:
cc @rohit-nayak-ps @deepthi would love to hear your thoughts on this! Thanks

sonne5:
cc @aquarapid for visibility as well

rohit-nayak-ps:
I get your point. We could definitely add keyspace/shard at the top level. Code change is easy, just a lot of tests need to be updated. 

However I would also like @keweishang's input on this since he was involved with writing the Vitess Debezium adapter: I wonder if they faced the same issue and if so, what approach they took.

sonne5:
I think until recent, Vitess Debezium connector only emit field event/row event as is, without thinking about the transaction.

Therefore it wasn't a concern so far.

keweishang:
>I think until recent, Vitess Debezium connector only emit field event/row event as is, without thinking about the transaction.

Correct, it isn't a concert for us so far. If adding keyspace/shard at the top level can be easily done, it would be much more consistent and Debezium friendly 👍

sonne5:
Thanks @rohit-nayak-ps, I will close this ticket.

cc @shichao-an @jeffchao @gunnarmorling for visibility:

[...]

PR #9386, opened 2021-12-14
VEvent: Add Keyspace/Shard properties to proto

rohit-nayak-ps:
## Description

Currently not all types of VEvents have the originating Keyspace/Shard defined: they are present in Row and Field events but we also have other special events like Journal, LastPK, Version, VGTID etc where they are not.

This makes it difficult for VStream API consumers to identify the source of some events. This PR adds the Keyspace and Shard attributes to every VEvent.

## Related Issue(s)
https://github.com/vitessio/vitess/issues/9332

## Checklist


mattlord (approved):
I like it! Event context always comes in handy :-)

sonne5 (approved):
Awesome, thanks for working on this @rohit-nayak-ps! 🎉 

This would help a lot for the integration with CDC projects like Debezium, I think it might worth mentioning this on vstream documentations if there is any or mention this in the next release.

deepthi (approved):

deepthi:
As @sonne5 said we should document this.

PR #9388, opened 2021-12-15
vtctld: use new healthcheck for realtime stats

deepthi:
Signed-off-by: deepthi <deepthi@planetscale.com>

## Checklist


deepthi:
We discussed this and decided that we can actually remove `tabletStatsCache` completely and simply call `HealthCheck.CacheStatus` whenever the related APIs are called.
@frouioui will create a replacement PR.

Issue #3481, opened 2017-12-18
MultiColumn Vindexes 


rafael:
## Overview

The following project aims to provide a more flexible interface to Vindexes. Current [implementation](https://github.com/youtube/vitess/blob/master/doc/V3VindexDesign.md) requires vindexes to be defined on a single **column**. This proposal will extend this interface so a vindex can be defined from a collection of **columns**.

## Motivation

The following are the main motivations to make this change:

* Given that you can have at most one column, there is an implicit requirement where the tuple **(column, keyspace_id)** must be unique for a secondary vindex to work properly. This creates a limitation when you want to add secondary vindex on a column where that property does not hold. 
* When using a multi column vindex, depending on the values available in the context of a given query you could use different parts of the index. 
   For instance, given a vindex on columns (**a,b,c**) and then a query with only **(a,b)** available in the context, we could use the multicolumn as non unique vindex on **(a,b)**. 
* It will be possible to define composite columns as the sharding key for a given table. 
* Unique secondary vindexes on more than 3 columns (current implementation only supports at most two columns).
## Implementation

This feature will be implemented by redefining the main interfaces that define the vindexes: 

### Map Interfaces for Unique and NonUnique

This interface will be updated to be: 

```
type Unique interface {
	Map(cursor VCursor, fromIds [][]sqltypes.Value) ([][]byte, error)
}
type NonUnique interface {
	Map(cursor VCursor, fromIds [][]sqltypes.Value) ([][][]byte, error)
}
```

### Lookup interfaces

It will be changed to: 

```
type Lookup interface {
	Create(vc VCursor, fromIds [][]sqltypes.Value, ksids [][]byte, ignoreMode bool) error
	Delete(VCursor, [][]sqltypes.Value, []byte) error
}
```

### Key concepts to keep in mind during implementation 

This will require a somewhat big refactor of the building and execution engine. There are many implementation details that are left outside of the scope of this design document. These details will become obvious in the implementation PR and could be discussed in that context.  The following are list of the most important pieces so we don't miss them during implementation. 
 
* A multi column vindex could be Unique or NonUnique depending on how many columns are available in the `where` clause. For this purpose, the Vindex interface should also be extended to reflect this fact. This is very important for join plans. If we have an unique vindex in the context, it is possible to push the entire query down to the shard. 

* If a value is not supplied, we use the `Reverse` function to back-compute the value of a column. With multi-columns, this will be harder. 

[...]

PR #9390, opened 2021-12-15
Multicolumn partial vindex selection

harshit-gangal:
## Description

This PR adds a new function to MultiColumn Vindex interface to provide information that the multi-column vindex supports partial vindex selection (i.e. fanout to subset of shards if part of the column values are provided).

Eg.
```
cola, colb, colc forms a multicolumn vindex
then providing values for cola or (cola and colb) the vindex should be able to fan out the query to subset of shards.
```

The user does not have to provide any additional information in the VSchema. Based on the table that uses multi-column vindex which allows partial vindex column selection, Vitess will handle it internally and the planner will be able to utilize those columns for vindex selection.

This PR also modifies the existing `region_experimental` vindex to handle partial column values i.e. if only `region_id` is provided to Map function.

## Related Issue(s)

- https://github.com/vitessio/vitess/issues/3481

## Checklist

### DIFF
--- a/go/test/endtoend/vtgate/gen4/gen4_test.go
+++ b/go/test/endtoend/vtgate/gen4/gen4_test.go
@@ -19,6 +19,7 @@ package vtgate
 import (
 	"context"
 	"fmt"
+	"strconv"
 	"testing"
 
 	"vitess.io/vitess/go/test/endtoend/vtgate/utils"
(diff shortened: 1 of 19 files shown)


go/test/endtoend/vtgate/gen4/gen4_test.go
```diff
@@ -276,3 +277,55 @@ func TestMultiColumnVindex(t *testing.T) {
...
+
+	for _, workload := range []string{"olap", "oltp"} {
+		utils.Exec(t, newConn, fmt.Sprintf(`set workload = %s`, workload))
+		for _, tcase := range tcases {
+			t.Run(workload+strconv.Itoa(tcase.regionID), func(t *testing.T) {
+				sql := fmt.Sprintf("select rg, uid, msg from region_tbl where rg = %d order by uid", tcase.regionID)
+				assert.Equal(t, tcase.exp, fmt.Sprintf("%v", utils.Exec(t, newConn, sql).Rows))
+			})
+		}
+	}
```
GuptaManan100, line 330: If I understand correctly, we are inserting rows in all the shards for a given regionID, and we are checking which all shard will be reached by a `select *` query. Love this test 😮

GuptaManan100 (approved):
LGTM!

go/test/endtoend/vtgate/gen4/gen4_test.go
harshit-gangal: that is correct. :)

PR #9391, opened 2021-12-15
evalengine: better compat v2

vmg:
## Description

Here's another iteration on the compatibility changes I've been making to the `evalengine` based on fuzzing expressions againsts a live MySQL instance.

The `evalengine` package has been significantly refactored, in preparation for some further changes that haven't been made to this PR yet. The types for the different expression nodes in the evaluation AST have been cleaned up are now using inheritance so their relationships look more similar to what they do in MySQL.

The main (new) functionality implemented in this PR is:

- support for all logical operators in SQL (AND, OR, XOR NOT)
- proper tuple comparisons, which have been implemented by decomposing the comparisons into logical steps like MySQL does
- introduction of an eager typecheck step which verifies the cardinality of tuples in the expressions, and which behaves similarly to what MySQL does
- several small fixes for evaluation results. A new `mysql_golden_1639145523.json` has been checked into the repository with 100+ test cases that were previously failing and are now green

## Checklist


go/vt/vtgate/evalengine/collate.go
```diff
@@ -0,0 +1,52 @@
+package evalengine
```
systay, line 1: license missing

go/vt/vtgate/evalengine/comparisons.go
```diff
@@ -261,39 +313,21 @@ func evalCompare(lVal, rVal EvalResult) (comp int, err error) {
 	}
 }
 
-func evalCompareTuplesNullSafe(lVal EvalResult, rVal EvalResult) (int, error) {
+func evalCompareTuplesNullSafe(lVal EvalResult, rVal EvalResult) (bool, error) {
 	if len(*lVal.tuple) != len(*rVal.tuple) {
-		return 0, vterrors.NewErrorf(vtrpcpb.Code_INVALID_ARGUMENT, vterrors.OperandColumns, "Operand should contain %d column(s)", len(*lVal.tuple))
+		panic("did not typecheck cardinality")
```
systay, line 318: should we not just return an error here?

go/vt/vtgate/evalengine/comparisons.go
```diff
@@ -309,106 +343,53 @@ func evalResultBool(b bool) EvalResult {
...
-// Type implements the ComparisonOp interface
-func (e *EqualOp) Type() querypb.Type {
-	return querypb.Type_UINT64
+	return cmp.evalResult(), nil
 }
 
-// String implements the ComparisonOp interface
-func (e *EqualOp) String() string {
-	return e.Operator
+// Type implements the Expr interface
```
systay, line 359: nit: comment should say `typeof`

go/vt/vtgate/evalengine/comparisons.go
```diff
@@ -492,28 +473,10 @@ func (l *LikeOp) Evaluate(left, right EvalResult) (EvalResult, error) {
...
-}
-
-func (r *RegexpOp) Evaluate(left, right EvalResult) (EvalResult, error) {
-	panic("implement me")
-}
-
-// Type implements the ComparisonOp interface
-func (r *RegexpOp) Type() querypb.Type {
-	return querypb.Type_UINT64
+func (l *LikeExpr) typeof(env *ExpressionEnv) (querypb.Type, error) {
```
systay, line 476: nit: same as above - comment should use method name

go/vt/vtgate/evalengine/convert.go
```diff
@@ -33,41 +34,95 @@ type (
...
-	case sqlparser.NotRegexpOp:
-		return &RegexpOp{Negate: true}
+		return &LikeExpr{BinaryCoercedExpr: coercedExpr, Negate: true}, nil
 	default:
-		return nil
+		return nil, vterrors.Errorf(vtrpcpb.Code_UNIMPLEMENTED, "%s: %s", ErrConvertExprNotSupported, op.ToString())
 	}
 }
 
+func convertLogicalExpr(opname string, left, right sqlparser.Expr, lookup ConverterLookup) (Expr, error) {
```
systay, line 94: I'm not quite following why we are sending the op as a string here. wouldn't the original AST struct work?

go/vt/vtgate/planbuilder/plan_test.go
```diff
@@ -722,13 +726,37 @@ type testCase struct {
 	comments         string
 }
 
+var (
```
systay, line 729: we usually do this using the `TestOne` method and `onecate.txt` file. do we need both these approaches to test a single query?

vmg: Oh I didn't know that. This just a made fixing bugs easier for me. Would you prefer I revert this?

go/vt/vtgate/evalengine/comparisons.go
vmg: This is on purpose: to match MySQL's behavior, we need to perform the cardinality checks early. If by the time we've reached this function the cardinalities do not match, then that's a bug, not an error.

go/vt/vtgate/evalengine/convert.go
vmg: The Logical Expressions in the AST have a different type for each logical expression, so by having this method we can build the resulting Expr node directly. If we pass in the AST node directly, we'd have to switch on it anyway to get the `opname` string, as it is needed when printing the resulting Expr.

systay (approved):

go/vt/vtgate/planbuilder/plan_test.go
vmg: Reverted :+1:

Issue #6675, opened 2020-09-05
Experimental: Mysql Protocol support for VStream


rohit-nayak-ps:
**Note: This is a quick experimental POC and NOT for production use**

We add a new vstream command in vtgate that uses the VStream API to stream data
from a table.

To try it out:

* run local_example upto (and not including) the 401_cleanup.sh
* start the mysql client pointing to vtgate
```mysql -u root --host=127.0.0.1 -P 15306 --quick```

```
mysql> set workload=olap;
Query OK, 0 rows affected (0.00 sec)

mysql> vstream * from product \G
*************************** 1. row ***************************
         op: +
        sku: SKU-1001
description: Monitor
      price: 100
*************************** 2. row ***************************
         op: +
        sku: SKU-1002
description: Keyboard
      price: 30

/* In another mysql client run "update product set price=200 where sku = 'SKU-1001';", gives: */
*************************** 3. row ***************************
         op: *
        sku: SKU-1001
description: Monitor
      price: 200

/* In another mysql client run "delete from product where sku = 'SKU-1001';", gives: */
*************************** 4. row ***************************
         op: -
        sku: SKU-1001
description: Monitor
      price: 200

```

Notes
* The first column contains + for inserts, - for deletes and * for updates
* Without any where clause all binlog events for that table are streamed
* We can get just new event using "vstream * from product where pos > 'current';"
* The pos in the where clause can be any gtid to start streaming from
* There is a limit clause to only stream a few rows "vstream * from product limit 10;"
* Currently data is streamed from all shards in the keyspace
* Only one table can be streamed at a time
* This is separate from the "stream * from <table>" support which uses the message subsystem

rohit-nayak-ps:
The functionality detailed above has been implemented in #6670.

PR #9392, opened 2021-12-15
Initial commit to add the 'vstream * from' functionality back to vtgate

rohit-nayak-ps:
## Description
The experimental `vstream * from` vtgate query functionality (#6675) was temporarily removed in #9252  because it was crashing vtgate after a recent vtgate refactor.

This PR changes the original functionality to be in line with the refactoring and also adds unit and e2e tests to alert us of such issues in the future.

As previous mentioned, the functionality is experimental and is intended to be provide a POC of what is possible. Once we have a firm requirement it will be enhanced.

## Related Issues

#6675
#9252

## Checklist


GuptaManan100 (approved):

harshit-gangal:
Currently, this uses a separate health check and picker logic, may be we can change this use the tabletgateway which also has the similar logic, may be it can make the logic simpler.

If the refactor makes sense, this can be taken up in separate PR.

harshit-gangal (approved):

PR #9393, opened 2021-12-15
Add -mysqlctl_mycnf_retain_file to vttablet backups

jakon89:
Signed-off-by: Adam Jedro <adamjedro@gmail.com>

## Description

This adds a new option `-mysqlctl_mycnf_retain_file` to vttablet. Now if we are restoring from backup, vttablet wipes all logs and uses default`my.cnf` file template (unless `mysqlctl_mycnf_template` is provided) before doing the restore. 

In my case the `my.cnf` is managed by an external service (puppet) and I don't want this file to be changed by vitess. This also allows me to hold `my.cnf` definition/template outside of `vttablet` configuration and manage it externally.

If `-mysqlctl_mycnf_retain_file` is set, it's up to user to provide non-coliding `server-id`.

## Checklist
-- manual tests performed

### DIFF
--- a/doc/releasenotes/13_0_0_summary.md
+++ b/doc/releasenotes/13_0_0_summary.md
@@ -2,6 +2,8 @@
 
 - `vtctl/vtctlclient ApplySchema` now respects `-allow-zero-in-date` for `direct` strategy. For example, the following statement is now accepted: `vtctlclient ApplySchema -skip_preflight -ddl_strategy='direct -allow-zero-in-date' -sql "create table if not exists t2(id int primary key, dt datetime default '0000-00-00 00:00:00')" commerce`
 
+- `-mysqlctl_mycnf_retain_file` was added to skip recreating `my.cnf` while restoring backup
+
 ## Major Changes
 
 ### vttablet -use_super_read_only flag now defaults to true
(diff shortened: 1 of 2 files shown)


deepthi:
@aquarapid do you see any problems with this?

aquarapid (approved):
I have no specific issue with this, since it is optional and a valid use case. However, I'd like to ask the general question of what type of config changes you are applying via puppet that you could not generally use the mysqlctl template for?  I assume these would be host/environment specific overrides/variations?

jakon89:
> I have no specific issue with this, since it is optional and a valid use case. However, I'd like to ask the general question of what type of config changes you are applying via puppet that you could not generally use the mysqlctl template for? I assume these would be host/environment specific overrides/variations?

Exactly - I want some nodes to have slightly different config. The other reason is that I have separate modules for vitess and MySQL so ideally, I want MySQL to be separated as much as possible from vitess.

deepthi:
@jakon89 did you test this manually? Doesn't seem worth adding a NEW endtoend test for it, but adding on to an existing test is an option. Is that something you can do?

jakon89:
@deepthi yeah, I did test it manually locally by building vitess using `make docker_local` and then doing backup & restore. I am happy to add a test case for that - can you please point me to the file that tests a similar use case? I would like to get it merged to 13.0.0 if possible - it's an important feature to us and in fact, it blocks our prod rollout.

deepthi:
@jakon89 there are a number of backup/restore endtoend tests here https://github.com/vitessio/vitess/blob/main/go/test/endtoend/backup/vtctlbackup/backup_utils.go

PR #9399, opened 2021-12-16
Allow GCM cipher without ECHDE for TLS 1.2

dbussink:
There are various databases such as Amazon AWS Aurora that do support TLS 1.2, but no ciphers that handle ECDHE.

Right now it means needing to downgrade to TLS 1.0 / 1.1 for a tablet, because there's otherwise no cipher overlap.

There is support however for GCM mode with no ECHDE and the known issues with CBC don't apply there.

This still means there's no perfect forward secrecy, but Aurora only supports that with DHE ciphers which Go doesn't support at all. So the best we can do is allowing this cipher as well so we can use at least TLS 1.2 and get the best possible setup.

With TLS 1.0 enabled, we allow all CBC ciphers including perfect forward secrecy ones. This ensures that in case of TLS 1.0 with only CBC support we still select forward secrecy if available.
 
Also removes the previously duplicated cipher.


Phanatic (approved):
LGTM

systay (approved):

Issue #7433, opened 2021-02-02
Consolidate / fix dockerfiles and docker builds


deepthi:
#### Overview of the Issue
We have several different docker images we publish on dockerhub which are supposed to be "full" or comprehensive vitess installs: `base`, `lite`, `k8s`.
In addition, we use `k8s` as the base and build small images for individual binaries.

We have a few problems right now
- `k8s` images are used by helm charts, but everyone else uses `lite` images. It will be nice to eliminate the k8s image altogether and use `lite` as the base for all the smaller images.
- `k8s` images are not auto-built with releases unlike the `lite` image
- When we make a release only the `mysql57` version of the `lite` image is being built.
- If we want to auto-build any other images (e.g. `vtexplain`) with a release, we currently don't have a way of making that wait until it's base image with the same tag has been built.
- Somehow each push to master results in **two** docker builds with the same SHA. Given the number of variants we build, this slows the pipeline down quite a bit.
- There are some other issues like `vtexplain` is installed in a different place from `vtctlclient` which is confusing to users.

derekperkins:
How committed are we to sticking with the name `lite`? There have been a number of questions about it in the past. If it is going to contain all the binaries, can we just call it `vitess`?

deepthi:
`vitess/vitess`? It would be a breaking change for people who are already using `lite`. That is the main concern.
@enisoc do you want to comment?

derekperkins:
Yes, `vitess/vitess`. We can always push the same image to `lite` too. `vitess/lite` isn't memorable and doesn't inspire confidence that you've downloaded the production version of Vitess.

https://hub.docker.com/r/cockroachdb/cockroach
https://hub.docker.com/r/mysql/mysql-server

On a similar note, we're artificially inflating the amount of installation and operational complexity by having all the separate binaries, since the majority of each binary is all the same code and a ton of dependencies. If we had a single binary that "shelled" out to all the existing binaries: e.g. `vitess vtgate`, `vitess vttablet`, etc, you'd have everything you needed in a single 100 MB binary instead of 10 80 MB files. There's a lot of value on the Cockroach side of having the install be a single binary. 

```
curl https://binaries.cockroachdb.com/cockroach-v20.2.4.darwin-10.9-amd64.tgz | tar -xJ
```

That may be more than we want to tackle here, but I really think we could significantly simplify things.

enisoc:
[...]

PR #9405, opened 2021-12-16
Add vtadmin binary to release and Docker images

jakon89:
Signed-off-by: Adam Jedro <adamjedro@gmail.com>

## Description

This adds `vtadmin` binary (server) to the release package and Docker images. I would like to run `vtadmin` for qa/prod clusters but unless binary is in the official package I am not allowed to do it.

## Checklist

### DIFF
--- a/docker/k8s/Dockerfile
+++ b/docker/k8s/Dockerfile
@@ -50,6 +50,7 @@ COPY --from=base /vt/bin/vtgate /vt/bin/
 COPY --from=base /vt/bin/vttablet /vt/bin/
 COPY --from=base /vt/bin/vtworker /vt/bin/
 COPY --from=base /vt/bin/vtbackup /vt/bin/
+COPY --from=base /vt/bin/vtadmin /vt/bin/
 
 # copy web admin files
 COPY --from=base $VTROOT/web /vt/web/
--- a/docker/lite/Dockerfile.mysql57
+++ b/docker/lite/Dockerfile.mysql57
@@ -51,6 +51,7 @@ ENV PATH $VTROOT/bin:$PATH
 # Copy artifacts from builder layer.
 COPY --from=builder --chown=vitess:vitess /vt/install /vt
 COPY --from=builder --chown=vitess:vitess /vt/src/vitess.io/vitess/web/orchestrator /vt/web/orchestrator
+COPY --from=builder --chown=vitess:vitess /vt/src/vitess.io/vitess/web/vtadmin /vt/web/vtadmin
 
 # Create mount point for actual data (e.g. MySQL data dir)
 VOLUME /vt/vtdataroot
--- a/Makefile
+++ b/Makefile
@@ -104,14 +104,14 @@ endif
 install: build
 	# binaries
 	mkdir -p "$${PREFIX}/bin"
-	cp "$${VTROOT}/bin/"{mysqlctl,mysqlctld,vtorc,vtctld,vtctlclient,vtctldclient,vtgate,vttablet,vtworker,vtbackup} "$${PREFIX}/bin/"
+	cp "$${VTROOT}/bin/"{mysqlctl,mysqlctld,vtorc,vtadmin,vtctld,vtctlclient,vtctldclient,vtgate,vttablet,vtworker,vtbackup} "$${PREFIX}/bin/"
 
 # Install local install the binaries needed to run vitess locally
 # Usage: make install-local PREFIX=/path/to/install/root
 install-local: build
 	# binaries
 	mkdir -p "$${PREFIX}/bin"
-	cp "$${VTROOT}/bin/"{mysqlctl,mysqlctld,vtorc,vtctl,vtctld,vtctlclient,vtctldclient,vtgate,vttablet,vtworker,vtbackup} "$${PREFIX}/bin/"
+	cp "$${VTROOT}/bin/"{mysqlctl,mysqlctld,vtorc,vtadmin,vtctl,vtctld,vtctlclient,vtctldclient,vtgate,vttablet,vtworker,vtbackup} "$${PREFIX}/bin/"
 
 
 # install copies the files needed to run test Vitess using vtcombo into the given directory tree.
--- a/tools/make-release-packages.sh
+++ b/tools/make-release-packages.sh
@@ -35,7 +35,7 @@ mkdir -p releases
 
 # Copy a subset of binaries from issue #5421
 mkdir -p "${RELEASE_DIR}/bin"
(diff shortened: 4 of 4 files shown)


doeg:
This looks good to me, and I appreciate you taking the time to do this! 💖 

That said, I'm definitely not the person to sign off on go releases -- @ajm188 @deepthi @rohit-nayak-ps would one of you mind signing off on this?

docker/k8s/Dockerfile
```diff
@@ -50,6 +50,7 @@ COPY --from=base /vt/bin/vtgate /vt/bin/
 COPY --from=base /vt/bin/vttablet /vt/bin/
 COPY --from=base /vt/bin/vtworker /vt/bin/
 COPY --from=base /vt/bin/vtbackup /vt/bin/
+COPY --from=base /vt/bin/vtadmin /vt/bin/
```
deepthi, line 53: Are you using the k8s image? The standard image is `vitess/lite`. You are already adding `vtadmin` to the `make install` target, that gets it into all the `lite` images.

deepthi (commented):
For all docker image changes we like to see a before and after size to make sure it is not getting too bloated. vtadmin will be a core component so we eventually need to get it in regardless of size, but it will be nice to see how much bigger the image gets.

docker/k8s/Dockerfile
derekperkins: I publish all the individual component images from the `k8s` image, so IMO this should stay. There is probably an additional step required to copy out the files needed for the web. This step `COPY --from=base $VTROOT/web /vt/web/` is what copies them for the existing UI, so I'd expect to see something similar for vtadmin.

derekperkins: Once this gets merged and we come up with an appropriate Dockerfile, I'll add a new directory and vtadmin image repo on Docker Hub that approximates the existing vtctld. https://github.com/vitessio/vitess/blob/main/docker/k8s/vtctld/Dockerfile

Side note: there's an open issue for consolidating images, but for now we still need to support people using `lite` and the component images. https://github.com/vitessio/vitess/issues/7433

jakon89:
Hey @deepthi

The binary alone is 52M on linux:
```
vitess@5432da5b3fcb:/vt/bin$ ls -lh | grep vtadmin
-rwxr-xr-x 1 vitess vitess  52M Dec 16 20:14 vtadmin
```

Docker image before:
```
REPOSITORY                                              TAG            IMAGE ID       CREATED          SIZE
vitess/lite                                             latest         60d5bedea288   15 seconds ago   1.19GB
```

and after:

```
vitess/lite                                             latest         84a8cde24a65   About a minute ago   1.13GB
```

docker/k8s/Dockerfile
jakon89: @derekperkins 
I have added `/web/vtadmin` to the lite image.

jakon89: btw there is a difference between vtadmin and orchestrator web files: orchestrator stores js/css files in final form in the repo, however, for `vtadmin` we copy the whole frontend project that needs to be built by `npm run build` - only then we get the final js/css files.

This is not a problem since in my project I use `vitess/lite` image as a base `builder` and then copy binaries/web files to final images (vtadmin-web, vtorc, vtgate, vtadmin-server).
Example of how I do it for vtorc:
```
ARG BASE_REGISTRY=gcr.io
ARG BASE_IMAGE=distroless/base
ARG BASE_TAG=nonroot

FROM vitess/lite:v12.0.1 AS builder

FROM ${BASE_REGISTRY}/${BASE_IMAGE}:${BASE_TAG}

COPY --from=builder vt/bin/vtorc /vtorc
COPY --from=builder vt/web/orchestrator /web/orchestrator

USER 65532
ENTRYPOINT ["/vtorc"]

HEALTHCHECK NONE
```

but IMO we should include the final build (js/css etc) somehow.

I see two potential ways of fixing it:
- we copy raw `web/vtadmin` to `docker/lite` and it's up to the user of `docker/lite` to build the frontend. That adds 10M to the `docker/lite` size. This is what is done in that PR for now - https://github.com/vitessio/vitess/pull/9405/files#diff-bdcbcdaa92c26cad86d024c8d1061d14c36e20f67c5c8e577c17e3bbb196c607R54
- we run `npm run build` as a part of `docker/lite` build and copy `build/` dir into final `docker/lite`. That adds additional 13M to the size of `docker/lite` (just checked)
- we start building separate `vtadmin-web` image:
    - use `node` Docker image as a builder
    - copy all vtadmin web files
    - run `npm install`
    - copy over generated js/css files to the final image (and serve them by `nginx` or other webserver)

cc @doeg ^ 

I will be happy to send a PR if we decide on the final form.

doeg: @jakon89 wow, thanks so much for the thorough notes! I really really appreciate it. 

Between the three, I don't have a strong opinion. I'm happy to go with whatever is most ergonomic for Vitess operators! 🤔 The third option is the closest to what we do at Slack, though that doesn't mean it's the best for everyone. @jakon89 @derekperkins do either you have a preference? 

P.S. There is definitely work to be done to make the front-end build smaller; the TypeScript generated from the protobufs is [particularly huge](https://github.com/vitessio/vitess/projects/12#card-60828184).

derekperkins: #3 is my preference, that for now I would put in `/docker/k8s/vtadmin`. You mentioned `vtadmin-web`, are you building separate images for the server and the web ui?

It's probably also worth building into `lite` per #2, but I'll leave that decision to @deepthi.

deepthi: If we follow the same pattern as we do for Golang binaries, we should do #2. I'm fine with that.

deepthi (approved):

docker/k8s/Dockerfile
jakon89: @deepthi let's merge that, I will implement building vtadmin-ui in new PR. Thanks

deepthi:
Not sure why all the CI actions got canceled. I re-ran a unit test and a cluster test before merging.

docker/lite/Dockerfile.mysql57
```diff
@@ -51,6 +51,7 @@ ENV PATH $VTROOT/bin:$PATH
 # Copy artifacts from builder layer.
 COPY --from=builder --chown=vitess:vitess /vt/install /vt
 COPY --from=builder --chown=vitess:vitess /vt/src/vitess.io/vitess/web/orchestrator /vt/web/orchestrator
+COPY --from=builder --chown=vitess:vitess /vt/src/vitess.io/vitess/web/vtadmin /vt/web/vtadmin
```
L3o-pold, line 54: Why this part was not included in mysql8?

deepthi:
@jakon89 how are you using this? see discussion in #10543

jakon89:
@deepthi we are building separate images `vtadmin-ui` and `vtadmin-api` out of vitess-lite docker image. AFAIK https://github.com/vitessio/vitess/pull/10543 introduced a single image that can be used to run any of the app (frontend or backend) so most probably we are going to migrate to it once vitess starts publishing a new image to the official docker repo.

PR #9409, opened 2021-12-17
Use PRS in Vtorc for electing new primary

GuptaManan100:
## Description

This PR changes the implementation of `electNewPrimary` to use `PlannedReparentShard` instead of having its own codepath. It also changes some checks in `PlannedReparentShard` to allow calling it with no new primary specified even for initialisation. No new end to end test has been added for this since there is already extensive coverage in the vtorc's tests which rely on `electNewPrimary` (`PlannedReparentShard`) to initialise the cluster.

## Related Issue(s)

- #8975 

## Checklist


GuptaManan100:
The implementation of the `AtLeast` method on `Position` has also been fixed after I found the bug via a unit test that I was adding that compared a nil position with a non-nil one. In this case, we should say that the non-nil position is at least as advanced as the nil one. On further investigation it turned out that the underlying function call to `Contains` does not handle nil case correctly as changed in https://github.com/vitessio/vitess/commit/031e3a39fc3baf5b845223cdc15f8a3edf7c5d75. So nils have to be handled at a level above the function call.

go/mysql/replication_position.go
```diff
@@ -68,8 +68,11 @@ func (rp Position) Equal(other Position) bool {
 
 // AtLeast returns true if this position is equal to or after another.
 func (rp Position) AtLeast(other Position) bool {
+	if other.GTIDSet == nil {
+		return true
+	}
 	if rp.GTIDSet == nil {
-		return other.GTIDSet == nil
+		return false
```
deepthi, line 75: Can you add some comments here? This is non-intuitive.

go/vt/orchestrator/inst/analysis_dao.go
```diff
@@ -308,7 +308,7 @@ func GetReplicationAnalysis(clusterName string, hints *ReplicationAnalysisHints)
 		) AS count_distinct_logging_major_versions
 	FROM
 		vitess_tablet
-		LEFT JOIN database_instance primary_instance ON (
+		JOIN database_instance primary_instance ON (
```
deepthi, line 311: Why did this need to change?

go/vt/vtctl/reparentutil/util.go
```diff
@@ -52,9 +52,6 @@ func ChooseNewPrimary(
 	// (TODO:@ajm188) it's a little gross we need to pass this, maybe embed in the context?
 	logger logutil.Logger,
 ) (*topodatapb.TabletAlias, error) {
-	if avoidPrimaryAlias == nil {
-		return nil, vterrors.Errorf(vtrpc.Code_INVALID_ARGUMENT, "tablet to avoid for reparent is not provided, cannot choose new primary")
-	}
```
deepthi, line 57: What case was this check trying to handle?

go/vt/orchestrator/inst/analysis_dao.go
GuptaManan100: The `database_instance` table contains critical information to be able to run any failover scenario. I encountered a bug wherein a vttablet instance was discovered and the `vitess_tablet` table was populated but before it's cluster information could be added to the `database_instance` we tried to run a failover. This tablet had no record in the `database_instance` table. This led to the LEFT JOIN filling all the values from that table with nils even the cluster name. So we ended up concluding that this tablet was the only tablet in its cluster and run a `clusterHasNoPrimary` failover.
I realised that we should only be running failover on tablets for which we have information in both the tables hence changed the LEFT JOIN to a normal JOIN and this fixed the bug. While discovery we populate both the tables but there is short time where we have populated the `vitess_tablet` table after reading the topo but have not populated the `database_instance` table since we are reading variables from the underlying mysql instance like the replication_lag, etc. We should discount these instances from running failovers until their information has been properly stored.

go/mysql/replication_position.go
GuptaManan100: Added comments explaining the code

go/vt/vtctl/reparentutil/util.go
GuptaManan100: Earlier, even when the user did not provide an `avoidPrimaryAlias` we used to set it to the shard's current primary and so never called `ChooseNewPrimary` with a nil `avoidPrimaryAlias`. But this had to be changed for the case of initialisation wherein we can choose  a new primary and there is not tablet which we should be avoiding. So, this check for a nil avoidPrimaryAlias has been removed and test cases for calling `ChooseNewPrimary` with `avoidPrimaryAlias` nil have been added.

deepthi (approved):

go/vt/vtctl/reparentutil/util.go
GuptaManan100: This change was indeed not correct and is the cause of the issue https://github.com/vitessio/vitess/issues/12018. However the problem that this fix intended to solve was still a concern. All of this has been addressed in https://github.com/vitessio/vitess/pull/12019

PR #9418, opened 2021-12-17
evalengine: lazy!

vmg:
## Description

This is my last contribution for 2021 :tada:

After a lot of research, I'm starting to understand why our evaluation engine wasn't always giving the same results as MySQL's (particularly for queries that were OK in MySQL but were returning errors for us).

There are two main reasons for this behavior:

- **The engines have different set of inputs**: the evaluation engine in MySQL only takes column values (i.e. the actual rows being processed), while our evaluation engine takes values _and_ bind variables. This is because our evaluation engine is processing query plans, which we normalize via these bind variables -- so when actually compiling a query we do not have all the type information required to ensure it is semantically valid. MySQL on the other hand does have this information, so it can catch errors _earlier_ than we do.

    This results in different sets of errors between the two implementations, and in some particular cases, in wrong results altogether. Since we need to continue supporting these bind variables, I have implemented a separate typechecking step that runs after compilation but before evaluation (when we already have access to the bind variables). This manual step does a good job of simulating the same type checks that MySQL performs.

- **The MySQL engine is lazy**, and ours isn't. The current implementation of the engine is a depth-first evaluator for an AST, and this does not match MySQL's behavior, which delays the evaluation of all the expressions until they're actually being used. The most obvious example of how this results in different evaluation results are _boolean expressions_, which in MySQL always short-circuit (like in most programming languages). For e.g. a boolean AND that has an expression error on the right-hand side, MySQL would return `true` if the left hand side was true, but we would return the error because we evaluate both hands before returning the result.

    Fixing this was not trivial: I refactored the whole `evalengine` package to support lazy evaluation without changing the external API. The `EvalResult` object now wraps an `Expr`, and only resolves the result of the expression when an evaluation function requires accessing the result.

In order to support this kind of lazy evaluation without going insane with error handling, I'm using a trick that I borrowed from the JSON marshaller in the Go Standard Library. The idea is using specially-wrapped `panic` statements when an error needs to be thrown, and then `recover`ing those panics looking for the specific wrapper that marks the panic as an error that must be handled as opposed to just a normal panic during execution. The result is quite efficient and elegant.

There's some more cleanup work to be done in this PR, but it greatly improves the compatibility with MySQL already, and it is ready for review.

## Checklist


go/vt/vtgate/evalengine/arithmetic_expr.go
```diff
@@ -50,61 +50,60 @@ func (b *ArithmeticExpr) collation() collations.TypedCollation {
...
+			return 0, err
+		}
+		rtype, err := b.Right.typeof(env)
+		if err != nil {
+			return 0, err
+		}
+		typ := mergeNumericalTypes(ltype, rtype)
+		return b.Op.typeof(typ), nil
+	*/
+	// TODO: make this less aggressive
```
systay, line 83: should this still be commented out?

go/vt/vtgate/evalengine/arithmetic_expr.go
```diff
@@ -30,7 +30,7 @@ type (
 
 	// ArithmeticOp allows arithmetic expressions to not have to evaluate child expressions - this is done by the BinaryExpr
 	ArithmeticOp interface {
-		eval(left, right EvalResult) (EvalResult, error)
+		eval(left, right, out *EvalResult) error
```
systay, line 33: I don't understand this change. Why is the output of the function one of the input params?

go/vt/vtgate/evalengine/eval_result.go
```diff
@@ -45,46 +50,265 @@ type (
 	}
 )
 
-var (
-	resultNull = EvalResult{typ: sqltypes.Null, collation: collationNull}
-)
+func (er *EvalResult) init(env *ExpressionEnv, expr Expr) {
```
systay, line 53: a comment would be helpful here

go/vt/vtgate/evalengine/eval_result.go
```diff
@@ -45,46 +50,265 @@ type (
...
+	er.expr = expr
+	er.env = env
+	er.typ3 = expr.typeof(env)
+}
+
+const typecheckEval = false
+
+func (er *EvalResult) resolve() {
```
systay, line 61: a comment would be helpful

go/vt/vtgate/evalengine/eval_result.go
```diff
@@ -31,12 +33,15 @@ import (
...
-		typ       querypb.Type
-		collation collations.TypedCollation
-		numval    uint64
-		bytes     []byte
-		tuple     *[]EvalResult
-		decimal   *decimalResult
+		expr Expr
+		env  *ExpressionEnv
+
+		typ3       querypb.Type
```
systay, line 39: What does the `3` in these fields mean?

go/vt/vtgate/evalengine/arithmetic_expr.go
vmg: Yes, I've updated the comment to explain why this is still a TODO

go/vt/vtgate/evalengine/arithmetic_expr.go
vmg: All the evaluation has been changed to take an `EvalResult` as an output argument instead of returning it because it opens up more possibilities for optimization. For simple evaluation cases, it allows us to pass the output result directly from our stack (assuming the compiler doesn't make it escape, which it does sometimes). For more complex cases, it'll allow us to use a memory pool for reusing temporary/intermediate evaluation results, something that couldn't be done if the result is returned directly by the evaluator.

go/vt/vtgate/evalengine/eval_result.go
vmg: Commented!

go/vt/vtgate/evalengine/eval_result.go
vmg: Commented!

go/vt/vtgate/evalengine/eval_result.go
vmg: I forgot to clean this up. I've given the fields more clean names and commented them.

frouioui (approved):
This is great work!

go/vt/vtgate/evalengine/eval_result.go
```diff
@@ -114,21 +364,399 @@ func (e *EvalResult) ToBooleanStrict() (bool, error) {
...
+			er.setRaw(v.Type(), v.Raw(), collations.TypedCollation{})
+			return nil
+		default:
+			return vterrors.Errorf(vtrpcpb.Code_INTERNAL, "coercion should not try to coerce this value to a text: %v", v)
+		}
+	}
+	return vterrors.Errorf(vtrpcpb.Code_INTERNAL, "coercion should not try to coerce this value: %v", v)
+}
+
+func (er *EvalResult) setBindVar1(typ querypb.Type, value []byte, collation collations.TypedCollation) {
```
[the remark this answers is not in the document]
vmg: This is a common pattern for internal methods that should not be called directly; as you can see, the `setBindVar` method calls `setBindVar1`. We could rename it to `setBindVarInternal` if you'd like.

systay (approved):

PR #9427, opened 2021-12-21
replace master with primary

arvind-murty:
## Description

Replace master with primary

## Checklist


deepthi (approved):
welcome to Vitess! thank you for your contribution.

PR #9428, opened 2021-12-21
Subsharding vindex

### LINKED ISSUE #7332: Feature Request: VSchema for “Subsharding” VIndex
## Goal

Specify a Multicolumn VIndex that contains (at least) two parts: a “primary” sharding key and a “secondary” sharding key. This allows a sharding strategy that constrains scatter queries that include the primary or “main” sharding key to a subset of shards.

For example, given two columns `A` and `B`, with values `a`  and `b` we can construct a keyspace id like:

`hash(a)[0..N] | hash(b)[0..M]`

Which would take the first `N` bytes from the hash of `a` and concatenate that with the first `M` bytes from the hash of `b`.

## Use Cases

A common use case for Vitess is when a given entity’s data will fit on more than one shard. However, if there is not a suitable intermediary grouping construct, storing that entity’s data across all shards will lead to expensive scatter queries.

A solution is using a prefix/postfix solution, which is really a subset of a generic tree-like sharding mechanism. In this solution, a keyspace id would include information about the larger entity and then progressively include more specific sharding information.

This would allow operators to still take advantage of splitting keyspaces as designed/intended while maintaining a higher level of performance than scatter queries when both:
1. the fully qualified shard key is not known
2. maintaining a lookup vindex is also prohibitive

## Query Patterns

When a query includes just the value `a`, we would like Vitess to not scatter and instead use the computation of `hash(a)` to constrain scatter queries to only the shards on which that hash would exist.

### Fan-out (but not scatter)
For example:

` select * from customers where A = a;`

Would target N shards, but not fully scatter.

We would also expect this to work with aggregates, ie:

`select c, sum(d) from customers where A = a;` 

### Point lookup
When read or write query includes both `a` and `b`, Vitess will compute the complete Keyspace ID and deterministically route to a specific shard.

For example:

`select * from customers where A = a and B = b;`

would hit one shard

### Insertion
Insertion will use column A and B to route the request to a specific shard.

`insert into customers (A,B,C) values (a,b,c)`

## Suggested VSchemas

Names are hard. For these example VSchemas, I will call this a “subshard” index but naming rights are left to the implementor.

### Simplest
This is a suggested example of the simplest way to configure this type of VIndex:
```
    "subshard_idx": {
      "type": "subshard"
    }
```

We would use it thus:
```
    "customer": {
      "column_vindexes": [
        {
            "columns": ["A", "B"],
            "name": "subshard_vdx"
        },
```
For this use, we would:
1. be inferring the “main shard” and “sub shard” from the position of the columns within the array of `columns`.
2. allowing the implementor to determine the hash key
3. allowing the implementor to determine the number of bytes used from the hash

I believe for (1) we'll always need to rely on the position of the columns in the array, but the number of columns does not need to be fixed at 2. Defaults (2) and (3) could potentially be parameterized. Examples of each are below.

### Enhancements
Support an arbitrary number of columns:
```
    "customer": {
      "column_vindexes": [
        {
            "columns": ["A", "B", "C"],
            "name": "subshard_vdx"
        },
```

It would also be possible to configure the number of bytes used for each part of the key:
```
    "subshard_vdx": {
      "type": "subshard",
      "params": {
          "main_shard_bytes" : 10,
          "sub_shard_bytes": 10
      }
    }
```

Or in a world with more than two columns:
```
    "subshard_vdx": {
      "type": "subshard",
      "params": {
          "bytes" : 10
      }
    }
```

It would also be possible to configure the hash algorithm:
```
    "subshard_vdx": {
      "type": "subshard",
      "params": {
          "hash_func" : "xxxhash"
      }
    }
```

We can use the issue to discuss which enhancements are worth doing :)

harshit-gangal:
## Description

This PR adds a new generic MultiColumn Vindex that can be used only as a primary vindex.

This Vindex takes in 3 inputs
1. `column_count` - the number of columns that would be provided for using the vindex.
2. `column_vindex` - hashing function each column will use to provide hash value for that column
3. `column_bytes` - bytes to be used from each column's hash value after applying hashing function on it to produce keyspace id.

Usage in VSchema:

```json
"vindexes": {
    "multicol_vdx": {
	  "type": "multicol",
	  "params": {
		"column_count": "3",
		"column_bytes": "1,3,4",
		"column_vindex": "hash,binary,unicode_loose_xxhash"
	  }
    }
}
```
```json
"tables": {
   "multicol_tbl": {
	  "column_vindexes": [
	    {
                "columns": ["cola","colb","colc"],
		"name": "multicol_vdx"
	    }
      ]
   }
}
```
`column_count` is the mandatory parameter that needs to be provided.
A maximum of 8 columns can be used in this vindex i.e. `column_count <= 8`

`column_vindex` should contain the vindex name in a comma-separated list. It should be less than equal to column_count.
Default vindex is `hash` vindex, any column for which vindex is not provided, the default vindex will be used.
Vindex in `column_vindex` should implement the below interface otherwise the initialization will fail.

```go
// Hashing defined the interface for the vindexes that export the Hash function to be used by multi-column vindex.
type Hashing interface {
	Hash(id sqltypes.Value) ([]byte, error)
}
```

`column_bytes` should contain bytes in a comma-separated list. The total count should be equal to 8 bytes.
If for some columns bytes are not represented then it is calculated by assigning equal bytes to remaining unassigned columns.

Eg:
```
Given:
column_count = 5
column_bytes = 1, , 3

col 1 -> 1
col 2 -> not-provided
col 3 -> 3
col 4 -> not-provided
col 5 -> not-provided

Calculated:
remaining bytes = 8 - 1 - 3 -> 4
remaining columns = 5 - 2 -> 3
col 2 -> 2
col 4 -> 1
col 5 -> 1
```

## Related Issue(s)

- Closes #7332 
- Related to #3481 

## Checklist

### DIFF
--- a/go/vt/vtgate/vindexes/multicol.go
+++ b/go/vt/vtgate/vindexes/multicol.go
@@ -33,7 +33,7 @@ type MultiCol struct {
 	name        string
 	cost        int
 	noOfCols    int
-	columnVdx   map[int]Vindex
+	columnVdx   map[int]Hashing
 	columnBytes map[int]int
 }
 
@@ -68,7 +68,7 @@ func NewMultiCol(name string, m map[string]string) (Vindex, error) {
 	}, nil
 }
 
-func getColumnVindex(m map[string]string, colCount int) (map[int]Vindex, int, error) {
+func getColumnVindex(m map[string]string, colCount int) (map[int]Hashing, int, error) {
 	var colVdxs []string
 	colVdxsStr, ok := m[paramColumnVindex]
 	if ok {
@@ -77,7 +77,7 @@ func getColumnVindex(m map[string]string, colCount int) (map[int]Vindex, int, er
 	if len(colVdxs) > colCount {
 		return nil, 0, vterrors.Errorf(vtrpcpb.Code_INVALID_ARGUMENT, "number of vindex function provided are more than column count in the parameter '%s'", paramColumnVindex)
 	}
-	columnVdx := make(map[int]Vindex, colCount)
+	columnVdx := make(map[int]Hashing, colCount)
 	vindexCost := 0
 	for i := 0; i < colCount; i++ {
 		selVdx := defaultVindex
@@ -92,12 +92,12 @@ func getColumnVindex(m map[string]string, colCount int) (map[int]Vindex, int, er
 		if err != nil {
 			return nil, 0, err
 		}
-		if !vdx.IsUnique() || vdx.NeedsVCursor() {
-			return nil, 0, vterrors.Errorf(vtrpcpb.Code_INVALID_ARGUMENT, "multicol vindex supports only unique and non-vcursor vindex function, passed vindex '%s' is invalid", selVdx)
+		hashVdx, isHashVdx := vdx.(Hashing)
+		if !isHashVdx || !vdx.IsUnique() || vdx.NeedsVCursor() {
+			return nil, 0, vterrors.Errorf(vtrpcpb.Code_INVALID_ARGUMENT, "multicol vindex supports vindexes that exports hashing function, are unique and are non-lookup vindex, passed vindex '%s' is invalid", selVdx)
 		}
 		vindexCost = vindexCost + vdx.Cost()
-		columnVdx[i] = vdx
-
+		columnVdx[i] = hashVdx
 	}
 	return columnVdx, vindexCost, nil
 }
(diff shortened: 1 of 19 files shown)


harshit-gangal:
@csquared take a look at the PR description.

frouioui:
@harshit-gangal, thank you for such a good description, we should reuse it for the documentation.

I have a question regarding `column_bytes`. Let's assume `column_count` equals 2, and `column_bytes` equals `4,4`. The 8 bytes used for the key will be created using 4 bytes from both columns (as instructed by `column_bytes`). Now, which 4 bytes will be taken from each column? The left-most or right-most bytes, or else?

go/vt/vtgate/vindexes/multicol.go
```diff
@@ -0,0 +1,245 @@
...
+		if !isHashVdx || !vdx.IsUnique() || vdx.NeedsVCursor() {
+			return nil, 0, vterrors.Errorf(vtrpcpb.Code_INVALID_ARGUMENT, "multicol vindex supports vindexes that exports hashing function, are unique and are non-lookup vindex, passed vindex '%s' is invalid", selVdx)
+		}
+		vindexCost = vindexCost + vdx.Cost()
+		columnVdx[i] = hashVdx
+	}
+	return columnVdx, vindexCost, nil
+}
+
+func getColumnBytes(m map[string]string, colCount int) (map[int]int, error) {
```
frouioui, line 189: nit: we should add a test for this

go/vt/vtgate/vindexes/multicol.go
```diff
@@ -0,0 +1,245 @@
...
+	}
+	partial := m.noOfCols > len(colValues)
+	return partial, ksid, nil
+}
+
+func init() {
+	Register("multicol", NewMultiCol)
+}
+
+func getColumnVindex(m map[string]string, colCount int) (map[int]Hashing, int, error) {
```
frouioui, line 155: nit: we should add a test for this

frouioui (approved):
Looks great to me!

go/vt/vtgate/vindexes/multicol.go
harshit-gangal: Other tests covers it.

go/vt/vtgate/vindexes/multicol.go
harshit-gangal: Other tests cover it.

harshit-gangal:
> @harshit-gangal, thank you for such a good description, we should reuse it for the documentation.
> 
> I have a question regarding `column_bytes`. Let's assume `column_count` equals 2, and `column_bytes` equals `4,4`. The 8 bytes used for the key will be created using 4 bytes from both columns (as instructed by `column_bytes`). Now, which 4 bytes will be taken from each column? The left-most or right-most bytes, or else?

_c1__c2_c2_c2__c3__c3__c3_c3_
|_0_|_1_|_2_|_3_|_4_|_5_|_6_|_7_|

For the column Cn = [0 : n_bytes_allocated_for_column]

PR #9430, opened 2021-12-22
Replace master with primary

arvind-murty:
## Description

Replace master with primary

## Checklist

### DIFF
--- a/go/vt/wrangler/vdiff_env_test.go
+++ b/go/vt/wrangler/vdiff_env_test.go
@@ -335,11 +335,6 @@ func (tmc *testVDiffTMClient) VReplicationWaitForPos(ctx context.Context, tablet
 	return nil
 }
 
-// TODO(deepthi): remove this after v13.0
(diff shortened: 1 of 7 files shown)


go/vt/wrangler/vdiff_env_test.go
```diff
@@ -336,9 +336,9 @@ func (tmc *testVDiffTMClient) VReplicationWaitForPos(ctx context.Context, tablet
 }
 
 // TODO(deepthi): remove this after v13.0
-func (tmc *testVDiffTMClient) MasterPosition(ctx context.Context, tablet *topodatapb.Tablet) (string, error) {
-	return tmc.PrimaryPosition(ctx, tablet)
-}
+// func (tmc *testVDiffTMClient) MasterPosition(ctx context.Context, tablet *topodatapb.Tablet) (string, error) {
+// 	return tmc.PrimaryPosition(ctx, tablet)
+// }
```
deepthi, line 341: Looks like this can be deleted instead of commenting out.
If upgrade/downgrade testing requires it we can always bring it back.

deepthi (approved):

Issue #8619, opened 2021-08-10
Task: rename column master_alias in table _vt.reparent_journal


deepthi:
Rename master_alias -> primary_alias

This has to be done in 2 phases for +/- 1 version compatibility.
- [x] in v13: add an alter statement that renames `primary_alias` to `master_alias`. This will take effect if someone tries to downgrade from v14 to v13. #9439 
- [ ] in v14: add an alter statement that renames `master_alias` to `primary_alias`. This will take effect during the upgrade from v13 to v14. #10098

PR #9439, opened 2021-12-23
reparent_journal: add backwards compatible alter statement

deepthi:
## Description

Part 1 of renaming master_alias to primary_alias

## Related Issue(s)

#8619 

## Checklist
- [ ] Tests were added or are not required - we should verify this works as part of upgrade/downgrade testing

### DIFF
--- a/go/cmd/vtbackup/vtbackup.go
+++ b/go/cmd/vtbackup/vtbackup.go
@@ -274,6 +274,11 @@ func takeBackup(ctx context.Context, topoServer *topo.Server, backupStorage back
 		if err := mysqld.ExecuteSuperQueryList(ctx, cmds); err != nil {
 			return fmt.Errorf("can't initialize database: %v", err)
 		}
+
+		// Execute Alter commands on reparent_journal and ignore errors
+		cmds = mysqlctl.AlterReparentJournal()
+		_ = mysqld.ExecuteSuperQueryList(ctx, cmds)
+
 		backupParams.BackupTime = time.Now()
 		// Now we're ready to take the backup.
 		if err := mysqlctl.Backup(ctx, backupParams); err != nil {
--- a/go/vt/discovery/topology_watcher.go
+++ b/go/vt/discovery/topology_watcher.go
@@ -125,19 +125,19 @@ func NewCellTabletsWatcher(ctx context.Context, topoServer *topo.Server, tr Tabl
 // Start starts the topology watcher
 func (tw *TopologyWatcher) Start() {
 	tw.wg.Add(1)
-	go func() {
-		defer tw.wg.Done()
-		ticker := time.NewTicker(tw.refreshInterval)
+	go func(t *TopologyWatcher) {
+		defer t.wg.Done()
+		ticker := time.NewTicker(t.refreshInterval)
 		defer ticker.Stop()
 		for {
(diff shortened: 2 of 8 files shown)


deepthi:
After fixing the panic, there is still a test failure
```
$ cd go/vt/vtgate
$ go test 
...
2021-12-31 09:29:38     INFO select.go:82] retrying plan after cnf: select * from INFORMATION_SCHEMA.foo
E1231 09:29:38.402715    2928 route.go:522] failed to route information_schema query to keyspace [vt_ks]
E1231 09:29:38.404243    2928 route.go:522] failed to route information_schema query to keyspace [vt_ks]
2021-12-31 09:29:38     INFO select.go:82] retrying plan after cnf: select * from INFORMATION_SCHEMA.`TABLES` where TABLE_SCHEMA = database()
E1231 09:29:38.579140    2928 vstream_manager.go:579] vstream for TestUnsharded/0 error: context canceled
E1231 09:29:38.579166    2928 vstream_manager.go:313] Error in vstream for keyspace:"TestUnsharded" shard:"0" gtid:"gtid01": context canceled
--- FAIL: TestStreamSQLSharded (0.01s)
    executor_stream_test.go:185: result: <nil>, want &{Fields:[name:"id" type:INT32 name:"value" type:VARCHAR] RowsAffected:0 InsertID:0 Rows:[[INT32(1) VARCHAR("foo")] [INT32(1) VARCHAR("foo")] [INT32(1) VARCHAR("foo")] [INT32(1) VARCHAR("foo")] [INT32(1) VARCHAR("foo")] [INT32(1) VARCHAR("foo")] [INT32(1) VARCHAR("foo")] [INT32(1) VARCHAR("foo")]] SessionStateChanges: StatusFlags:0}
W1231 09:29:38.591591    2928 executor.go:182] "select * from main1" exceeds warning threshold of max memory rows: 3
```
However, if I run just that one test, it passes
```
$ go test -run=TestStreamSQLSharded
ERROR: logging before flag.Parse: I1231 09:30:45.272020    2997 flags.go:185] vtgate buffer not enabled.
ERROR: logging before flag.Parse: I1231 09:30:45.272252    2997 discoverygateway.go:123] loading tablets for cells:
ERROR: logging before flag.Parse: I1231 09:30:45.272305    2997 gateway.go:103] Gateway waiting for serving tablets of types [] ...
ERROR: logging before flag.Parse: I1231 09:30:45.272312    2997 gateway.go:111] Waiting for tablets completed
ERROR: logging before flag.Parse: I1231 09:30:45.272329    2997 vtgate.go:106] Transaction mode: 'MULTI'
ERROR: logging before flag.Parse: I1231 09:30:45.272824    2997 vschema_manager.go:102] Received vschema update
ERROR: logging before flag.Parse: I1231 09:30:45.273192    2997 streamlog.go:158] Streaming logs from VTGate at /debug/querylog.
PASS
ok  	vitess.io/vitess/go/vt/vtgate	0.290s
```
Possibly there is some interaction between different tests that is causing this to fail only when the full suite is run.

deepthi:
Looks like this test broke in #9392, but moving the new test to the end of the file fixes it. There's an underlying problem with test dependencies but we can work around it with this change.

rohit-nayak-ps (approved):
lgtm

PR #9440, opened 2021-12-23
Fixed case-sensitive TabletType in SwitchTraffic

arvind-murty:
## Description

Currently SwitchTraffic requires TabletTypes to be all lowercase. This PR makes it case-insensitive.

## Checklist
- [X] Documentation was added or is not required https://github.com/vitessio/website/pull/919


rohit-nayak-ps (approved):
lgtm

PR #9441, opened 2021-12-24
Add flag to keep routing rules in v2 vrepl workflows

mattlord:
## Description
Once you've [completed a MoveTables](https://vitess.io/docs/13.0/reference/vreplication/complete/) e.g., you may want to keep the routing rules around for a while as you find and fix any application code that is still using the old keyspace. 

This adds a `-keep_routing_rules` flag to the `vtctl MoveTables {Complete,Cancel}` commands for the v2 workflows to support this use case.

## Checklist
- [x] Tests were added
- [ ] TODO: Documentation was added


rohit-nayak-ps (approved):
lgtm

PR #9447, opened 2021-12-28
Use uint64 for sequence in-memory structure

mattlord:
## Description
This then allows us to support `BIGINT UNSIGNED` columns in the backing MySQL table. When using a `SIGNED` value we're effectively wasting half of the ID space since negative values are not used.

## Related Issue(s)
I can then update the example in the improved docs here to use `BIGINT UNSIGNED` for the `user_seq` table: https://github.com/vitessio/website/pull/928

## Checklist
- [x] New tests are not required
- [x] Documentation was added

### DIFF
--- a/go/vt/vttablet/endtoend/sequence_test.go
+++ b/go/vt/vttablet/endtoend/sequence_test.go
@@ -35,14 +35,14 @@ func TestSequence(t *testing.T) {
 	want := &sqltypes.Result{
 		Fields: []*querypb.Field{{
 			Name: "nextval",
-			Type: sqltypes.Int64,
+			Type: sqltypes.Uint64,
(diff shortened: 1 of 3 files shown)


sougou:
For 64-bit numbers, the signed vs unsigned bit is not materially important. The general idea is that one is not expected to come anywhere close to this limit. However, there is an advantage to making it signed. This allows for client apps to use a signed integer, which is a more natural way for programs to represent values.

This does make it slightly incompatible with what MySQL can theoretically allow, but I don't think it has any practical implications.

mattlord:
Thanks, @sougou ! I'll close this for now then and we can revisit later if it ever becomes a practical issue.

Issue #9400, opened 2021-12-16
Broken reverse vreplication after SwitchWrites in a middle of long transaction


DeathBorn:
#### Overview of the Issue

Performing SwitchWrites in a middle of long transaction after data manipulation breaks reverse vreplication.

#### Reproduction Steps

[This happens across many versions](https://gist.github.com/DeathBorn/cbef57a8c957a95886c570b08080c1ce) 

```bash
# Start
./101_initial_cluster.sh
./201_customer_tablets.sh
mysql < ../common/insert_commerce_data.sql
mysql commerce -e 'select * from customer;'
# +-------------+--------------------+
# | customer_id | email              |
# +-------------+--------------------+
# |           1 | alice@domain.com   |
# |           2 | bob@domain.com     |
# |           3 | charlie@domain.com |
# |           4 | dan@domain.com     |
# |           5 | eve@domain.com     |
# +-------------+--------------------+

# Movetables
vtctlclient MoveTables -source commerce -tables 'customer' Create customer.commerce2customer
vtctlclient ReloadSchemaKeyspace customer

# Dummy insert
mysql -e "insert into customer(email) values('BEFORESWITCH@GOOD.COM');"
# Check data
mysql commerce -e 'select * from customer;'
# +-------------+-----------------------+
# | customer_id | email                 |
# +-------------+-----------------------+
# |           1 | alice@domain.com      |
# |           2 | bob@domain.com        |
# |           3 | charlie@domain.com    |
# |           4 | dan@domain.com        |
# |           5 | eve@domain.com        |
# |           6 | BEFORESWITCH@GOOD.COM |
# +-------------+-----------------------+

# Check Diff and switch reads
vtctlclient VDiff customer.commerce2customer
# ...
# Summary for table customer:
#   ProcessedRows: 6
#   MatchingRows: 6
#   MismatchedRows: 0
#   ExtraRowsSource: 0
#   ExtraRowsTarget: 0
# ...
vtctlclient MoveTables -tablet_types=rdonly,replica SwitchTraffic customer.commerce2customer

# Begin reproduction
mysql> begin;
Query OK, 0 rows affected (0.00 sec)

mysql> insert into customer(email) values('PROBLEM@PROBLEM.BUG');
Query OK, 1 row affected (0.00 sec)

vtctlclient MoveTables -tablet_types=primary SwitchTraffic customer.commerce2customer
[...]

PR #9450, opened 2021-12-29
Prevent lost writes on sources during SwitchWrites

mattlord:
Shutting down the query service — which in turn closes all connection pools, which in turn closes all MySQL connections in order to effectively rollback any running statements/transactions — on the source primary tablets when we switch writes away from them is the only way to ensure that no lost writes occur due to long running transactions (including long running single statement auto-commit transactions) on the source tablets committing after writes have been switched. I say this because:
- Simply blocking new write statements against the `foo` table is not enough. We have no reliable mechanism to see the graph of changed objects inside InnoDB for a transaction — so we cannot simply kill transactions that have touched table `foo`
    - `foo` may have already been changed in the transaction and only a `commit` comes later — we do track all user statements executed within a transaction already, but it would be a fair amount of resulting overhead in the hot query path for every single transaction if we were to examine all queries executed in the transaction and compare the tables (idents) involved in each against the keyspace/shard's current denied_tables list on `commit`, and...
    - `foo` may have been indirectly changed via triggers, functions, updatable views, etc on the MySQL/InnoDB side (which the tablet is otherwise unaware of)

Without doing this (before this PR is merged) you can not only break reverse vreplication -- which would have to be fixed manually -- but **_if you don't notice the errant writes and then manually reconcile the tables across keyspaces (old and new) then those writes are lost_**.

With this patch, when you go through [the manual test here](https://github.com/vitessio/vitess/issues/9400) and get to the point where you attempt another `insert` or `commit` in the previously opened transaction started against the source before we switched writes, you get an error like this:
```
ERROR 1317 (70100): vttablet: rpc error: code = Aborted desc = transaction 1640897030631460257: ended at 2021-12-30 20:45:20.739 UTC (transaction rolled back) (CallerID: userData1)
```

> :question: When stopping the query service for switching the writes, we override `-shutdown_grace_period` and use 100ms. Should we instead use the tablet config value for `-queryserver-config-transaction-timeout` (default is 30 seconds)?

## Related Issue(s)
Fixes: https://github.com/vitessio/vitess/issues/9400

## Checklist
- [x] Should this PR be backported? At least to 12.0, I think...
- [x] Tests were added
- [ ] Documentation: TODO


go/vt/vttablet/tabletservermock/controller.go
```diff
@@ -211,6 +211,14 @@ func (tqsc *Controller) SchemaEngine() *schema.Engine {
 	return nil
 }
 
+func (tqsc *Controller) StopQueryService(ctx context.Context) error {
+	return nil
+}
+
+func (tqsc *Controller) StartQueryService(ctx context.Context) error {
```
GuptaManan100, line 218: nit: These are exported methods, could you add comments for them?

GuptaManan100 (approved):
Everything else LGTM. I don't think we should be using the `queryserver-config-transaction-timeout` to wait, since within that time a new transaction can be started and then those would need to be cancelled

mattlord:
Closing this for now in favor of the `LOCK TABLES` based approach being taken in: https://github.com/vitessio/vitess/pull/9481

PR #9451, opened 2021-12-29
vitessdriver: add support for DistributedTx

derekperkins:
## Description
This enables users of the Go driver to serialize a session and connect to an existing transaction while still using the `database/sql` driver, as presented by @dkhenry here
https://www.cncf.io/online-programs/transactional-microservices-with-vitess-coordination-without-state/

This is very rough for the purposes of discussion about the api design

## Checklist

### DIFF
--- a/go/vt/vitessdriver/driver.go
+++ b/go/vt/vitessdriver/driver.go
@@ -167,6 +167,9 @@ type Configuration struct {
 	//
 	// Default: "vitess"
 	DriverName string `json:"-"`
+
+	// allows for some magic
+	Session *vtgateconn.VTGateSession
 }
 
 // toJSON converts Configuration to the JSON string which is required by the
@@ -205,7 +208,11 @@ func (c *conn) dial() error {
 	if err != nil {
 		return err
 	}
-	c.session = c.conn.Session(c.Target, nil)
+	if c.Configuration.Session != nil {
+		c.session = c.Configuration.Session
+	} else {
+		c.session = c.conn.Session(c.Target, nil)
+	}
 	return nil
 }
 
@@ -229,6 +236,26 @@ func (c *conn) PrepareContext(ctx context.Context, query string) (driver.Stmt, e
 func (c *conn) Close() error {
 	c.conn.Close()
 	return nil
+}
+
+// GetDistributedTx allows users to send sessions over the wire and reconnect to an existing transaction
+func GetDistributedTx(ctx context.Context, session *vtgateconn.VTGateSession) (*sql.Tx, error) {
+	db, err := OpenWithConfiguration(Configuration{
+		// include session here - there will be a new *DB created each time
+		// that stores the session state in the &conn{} struct
+		Session: session,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	// this should return the only connection associated with the db
+	c, err := db.Conn(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	return c.BeginTx(ctx, nil)
 }
 
 func (c *conn) Begin() (driver.Tx, error) {


go/vt/vitessdriver/driver.go
```diff
@@ -231,7 +251,88 @@ func (c *conn) Close() error {
...
 }
 
+// DistributedTxFromSessionToken allows users to send serialized sessions over the wire and
+// reconnect to an existing transaction. Setting the sessionToken and address on the
+// supplied configuration is the minimum required
+func DistributedTxFromSessionToken(ctx context.Context, c Configuration) (*sql.Tx, error) {
+	if c.SessionToken == "" {
+		return nil, errors.New("c.SessionToken is required")
+	}
+	if c.Address == "" {
```
dkhenry, line 261: Is this to prevent clients from connecting to different vtgates ? In theory the transaction is help open at the vttablet layer and we should be able to connect through any vtgate to have the transaction continue to work

derekperkins: The address isn't serialized as a part of the session object and so is required. I could remove this check and let it error further down the line

dkhenry (approved):
LGTM I like it, I would wait until @deepthi or @sougou can get 👀  on it before you land to make sure we aren't violating some unknown constraint.

go/vt/vitessdriver/driver.go
derekperkins: check removed

go/vt/vitessdriver/driver.go
```diff
@@ -167,6 +172,10 @@ type Configuration struct {
 	//
 	// Default: "vitess"
 	DriverName string `json:"-"`
+
+	// SessionToken is a protobuf encoded vtgatepb.Session represented as base64, which
+	// can be used to distribute a transaction over the wire.
+	SessionToken string
```
frouioui, line 178: The test `TestConfigurationToJSON` transforms the `Configuration` struct to a JSON string and compares the string with a constant expected output. Because we are adding the field `SessionToken` to `Configuration`, we must update the assertion of `TestConfigurationToJSON`.

The test is failing in GitHub Actions as observed here: https://github.com/vitessio/vitess/runs/4670869522?check_suite_focus=true

frouioui (approved):
The pull request looks great to me besides the failing unit test

go/vt/vtgate/vtgateconn/vtgateconn.go
```diff
@@ -55,6 +54,19 @@ func (conn *VTGateConn) Session(targetString string, options *querypb.ExecuteOpt
...
+}
+
+// SessionFromPb returns a VTGateSession based on the provided proto session.
+func (conn *VTGateConn) SessionFromPb(sn *vtgatepb.Session) *VTGateSession {
+	return &VTGateSession{
+		session: sn,
+		impl:    conn.impl,
+	}
+}
+
```
deepthi, line 69: I'd like to get @sougou / @systay to comment on whether we are comfortable adding these.

deepthi (commented):
Looks fine to me.
Let's see what @sougou and @systay have to say.

go/vt/vtgate/vtgateconn/vtgateconn.go
sougou: > I'd like to get @sougou / @systay to comment on whether we are comfortable adding these.

Feels like it's a necessary hack. I don't have an objection.

go/vt/vitessdriver/driver.go
derekperkins: Thanks for pointing that out. Fixed now.

go/vt/vitessdriver/driver.go
```diff
@@ -231,7 +249,129 @@ func (c *conn) Close() error {
...
+		}
+
+		if len(session.ShardSessions) > originalShardSessionCount {
+			return fmt.Errorf("mismatched ShardSession count: originally %d, now %d",
+				originalShardSessionCount, len(session.ShardSessions),
+			)
+		}
+
+		return nil
+	}
```
derekperkins, line 308: I tested this in our setup and found this pathological edge case where there were no errors, but transactional data just disappeared. After tracking it down, I hadn't done any transactional work before sending over the session token, so it had no shard sessions, leaving the new ones orphaned. 

The idea is that this function will help to prevent that from happening. I return this func rather than making a separate call, so that users are forced to be aware of it. The pattern is similar to a context cancel function that is returned on any new context operations, so it should feel familiar to people.

I thought about doing a deep comparison on the shard sessions, but that seemed like more work than just comparing the counts without adding any extra benefits.

derekperkins: @dkhenry @frouioui @deepthi @sougou this is a decent enough change that it's worth another quick look. The core functionality hasn't changed, this is just some safeguards, but I want to make sure that I'm fully understanding it.

frouioui (approved):

go/vt/vitessdriver/driver.go
deepthi: The new code lgtm.

derekperkins:
This is working well in our infrastructure, not at particularly high scale, but is working well. I'm very glad that the first case I tried it on exposed the shard session bug - I believe this should be fairly safe for general consumption with the safeguards in place. I'll go ahead and merge with the approvals given.

PR #9453, opened 2021-12-30
Do Not Merge: MoveTables: open transactions causes data loss after a SwitchWrites

rohit-nayak-ps:
## Description

There is a bug today where open DMLs can cause data consistency after SwitchWrites in a MoveTables workflow.  This PR confirms this by extending the TestMultiCellVreplicationWorkflow e2e test, which adds a transaction that stays open across SwitchWrites to mimic a long-running transaction. This test is currently failing because an extra row gets created in the source (corresponding to the row inserted in the long-running transaction).

This is not an issue for Reshard workflows because open transactions are closed (either by waiting for a grace period or finally cancelled) as part of the primary tablet shutting down since the source shards stop serving in SwitchWrites.

For MoveTables the source shard continues to serve: the moved tables are first added to the denied tables list in the source shards' tablet controls and the vschema of the source and target keyspaces updated to reflect the move. This is not enough for transactions that insert into or update the moved tables which stay open across the SwitchWrites and commit after SwitchWrites is done. At this time the forward replication streams have been frozen and the source table updated after that. So now we have data inconsistency: the source now has changes that have not been replicated to the target.

If the open transaction inserted a new row, reverse replication was turned on and an autoincrement PK is used then reverse replication is broken as reported by https://github.com/vitessio/vitess/issues/9400. Otherwise the data loss will be silent.
 
Also see https://github.com/vitessio/vitess/pull/9450 for a WIP fix.

## Checklist

### DIFF
--- a/go/test/endtoend/vreplication/cluster.go
+++ b/go/test/endtoend/vreplication/cluster.go
@@ -238,6 +238,7 @@ func (vc *VitessCluster) AddTablet(t testing.TB, cell *Cell, keyspace *Keyspace,
 		"-enable-lag-throttler",
 		"-heartbeat_enable",
 		"-heartbeat_interval", "250ms",
+		"-shutdown_grace_period", "5",
 	} //FIXME: for multi-cell initial schema doesn't seem to load without "-queryserver-config-schema-reload-time"
 
 	if mainClusterConfig.vreplicationCompressGTID {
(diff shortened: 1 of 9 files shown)


mattlord:
I'm going to close this for now as I've cherry-picked the test commit here: https://github.com/vitessio/vitess/pull/9450

Thanks, @rohit-nayak-ps ! I also copy some of your text from here regarding Reshard etc.

rohit-nayak-ps:
> I'm going to close this for now as I've cherry-picked the test commit here: #9450
> 

Reopened to try a  POC of a different approach of looking for a change in the denied table list inside updateLocked() so that we restart the query service automatically if it changes, without needing to explicitly do that within SwitchWrites.

rohit-nayak-ps:
Functionality has been completed in https://github.com/vitessio/vitess/pull/9481

PR #9456, opened 2022-01-04
Fix parsing table name starting with a number.

### LINKED ISSUE #4069: Table or column start with digit will cause parse error.
Hi，I found that when the column or table start with digit, sqlparser would send an error for the syntax.
But it's supported in MySQL.

when the tokenizer scan the query, if there is a digit, lexer will returen `tkn.scanNumber(false)`.

sqlparser/token.go
```
......
case isLetter(ch):
	tkn.next()
	if ch == 'X' || ch == 'x' {
		if tkn.lastChar == '\'' {
			tkn.next()
			return tkn.scanHex()
		}
	}
	if ch == 'B' || ch == 'b' {
		if tkn.lastChar == '\'' {
			tkn.next()
			return tkn.scanBitLiteral()
		}
	}
	isDbSystemVariable := false
	if ch == '@' && tkn.lastChar == '@' {
		isDbSystemVariable = true
	}
	return tkn.scanIdentifier(byte(ch), isDbSystemVariable)
case isDigit(ch):
	return tkn.scanNumber(false)
case ch == ':':
	return tkn.scanBindVar()
......
```

GuptaManan100:
## Description

This PR fixes the tokeniser to return an `ID` token when we see letters after numbers instead of returning errors, since this is closer to MySQL. `3t2` is a valid `ID` and should be parsed correctly. It can be a table name too. This means that the following query is valid and should work correctly in Vitess - 

```sql
create table 3t2 (c1 bigint not null, c2 text, primary key(c1))
```

Even after the changes the following query does not work with Vitess - 
```sql
create table 2t.3t2 (c1 bigint not null, c2 text, primary key(c1))
```
This is a valid MySQL query but does not yet work with Vitess. The problem is that the tokenizer takes `.3` as a single token which causes parsing error. We should instead be using `.` as a separate token and then `3t2` as an identifier. This highlights another problem, the tokenization has to be aware of the context of parsing! Since in an alternate query like `select .3e3t`, we should use `.3e3` as a single token FLOAT and then `t` as ID.
		  

## Related Issue(s)

Fixes #4069 

## Checklist

### DIFF
--- a/go/vt/sqlparser/parse_test.go
+++ b/go/vt/sqlparser/parse_test.go
@@ -48,6 +48,9 @@ var (
 		input:  "create table x (e enum('red','yellow') null collate 'utf8_bin')",
 		output: "create table x (\n\te enum('red', 'yellow') collate 'utf8_bin' null\n)",
 	}, {
+		input:  "create table 3t2 (c1 bigint not null, c2 text, primary key(c1))",
+		output: "create table 3t2 (\n\tc1 bigint not null,\n\tc2 text,\n\tprimary key (c1)\n)",
+	}, {
 		input:  "select 1 from t1 where exists (select 1) = TRUE",
 		output: "select 1 from t1 where exists (select 1 from dual) = true",
 	}, {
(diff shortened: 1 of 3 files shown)


harshit-gangal (commented):
Looks good.

Add a test for case like `create table 2k.1a (....)`. 

Also, why do backticks need to be added here for these identifiers?

frouioui (approved):
LGTM

GuptaManan100:
@harshit-gangal backtics are not necessarily needed in this case but we already had code for adding backtics when an identifier starts with a number.

PR #9457, opened 2022-01-04
Remove SSL cipher customization

dbussink:
This was done long ago when upstream Go had no mitigations in place for CBC oracle attacks. For the legacy ciphers they do now, so having these different definitions isn't really suited anymore.

In general, Go has a good upstream TLS configuration so we should modify / configure as little as possible so I think it's better to remove this customization.

It has cause some issues like https://github.com/vitessio/vitess/pull/9399 before removing it also prevents similar issues in the future if Go improves cipher support.

### DIFF
--- a/go/vt/orchestrator/ssl/ssl.go
+++ b/go/vt/orchestrator/ssl/ssl.go
@@ -17,19 +17,6 @@ import (
 	"vitess.io/vitess/go/vt/orchestrator/external/golib/log"
 )
 
-var cipherSuites = []uint16{
-	tls.TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,
-	tls.TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,
-	tls.TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,
-	tls.TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,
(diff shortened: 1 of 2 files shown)


go/vt/orchestrator/ssl/ssl.go
```diff
@@ -166,9 +150,9 @@ func ReadPEMData(pemFile string, pemPass []byte) ([]byte, error) {
 		pemData, err = x509.DecryptPEMBlock(pemBlock, pemPass) //nolint SA1019
 		if err != nil {
 			return pemData, err
-		} else {
-			log.Infof("Decrypted %v successfully", pemFile)
 		}
+		log.Infof("Decrypted %v successfully", pemFile)
```
dbussink, line 154: Linter complained about this one so fixed it up.

go/vt/orchestrator/ssl/ssl.go
```diff
@@ -47,9 +34,6 @@ func NewTLSConfig(caFile string, verifyCert bool) (*tls.Config, error) {
 
 	// Set to TLS 1.2 as a minimum.  This is overridden for mysql communication
 	c.MinVersion = tls.VersionTLS12
-	// Remove insecure ciphers from the list
-	c.CipherSuites = cipherSuites
-	c.PreferServerCipherSuites = true
```
dbussink, line 52: This option doesn't do anything anymore these days.

deepthi (approved):

PR #9303, opened 2021-11-30
vtctl ApplySchema accepts 'ALTER VITESS_MIGRATION...' statements

shlomi-noach:
## Description

The following is now supported:
```shell
$ vtctlclient ApplySchema -sql "alter vitess_migration 'dce5f77e_51de_11ec_a049_0a43f95f28a3' cancel" commerce
```
`ApplySchema` now accepts all variants of `ALTER VITESS_MIGRATION...`.

#### Why are we doing this?

We long since wanted to deprecate VExec for OnlineDDL. Right now `vtctl OnlineDDL <command>` heavily relies on wrangler/VExec to get the command to the underlying tablets. However, `ApplySchema` already has the required logic to get in to the `QueryService`. We will reimplement `vtctl OnlineDDL` on top of `ApplySchema` instead of `VExec`.

This PR is a first step towards that re-implementation.

## Related Issue(s)

- Tracking:  https://github.com/vitessio/vitess/issues/6926#issuecomment-817619302
- https://github.com/vitessio/vitess/pull/8971 as a previous attempt, reverted since.
- 
## Checklist
- [ ] Should this PR be backported?
- [ ] Tests were added or are not required
- [ ] Documentation was added or is not required

PR #9458, opened 2022-01-04
vtctl: Allow ALTER VITESS_MIGRATION commands in sharded environment

shlomi-noach:
## Description

Followup to #9303, which added support for `ALTER VITESS_MIGRATION` in `vtctl ApplySchema`. 

Unfortunately, on sharded keyspaces an internal validation fails the statement. This PR fixes the validation to allow `ALTER VITESS_MIGRATION` even in sharded (multi-shard) environments.

## Related Issue(s)

- Tracking: #6926 

## Checklist


dbussink (approved):
Nice, good to allow these as well then.

deepthi (approved):

Issue #9469, opened 2022-01-05
Sanitize Statements in Aborted Transaction Log Record


mattlord:
#### Feature Description

Currently the statements executed within a transaction are logged by `vttablet` when the transaction is aborted due to hitting the configured timeout set via the `-queryserver-config-transaction-timeout` flag (defaults to 30s). For example:
```
W0105 17:18:53.558389    3783 tx_pool.go:126] killing transaction (exceeded timeout: 30s): 1641402138560038177  'principal:"vitess" component:"127.0.0.1:56294" subcomponent:"VTGate MySQL Connector"'  'username:"userData1"'  Jan  5 17:18:22.025067  Jan  1 00:00:00.000000  -9223372036.854776              insert into customer(email) values ('PROBLEM@LOGGING.WHY');insert into customer(email) values ('PROBLEM2@LOGGING.WHY')
```

These statements should be sanitized to prevent leaking PII and other potentially sensitive info in the logs.

#### Use Case(s)

When you need to meet various compliance objectives it's critical to omit potentially sensitive information from logs which are typically considered non-sensitive with their own distinct copy and retention policies.

PR #9470, opened 2022-01-05
Support redacting statements in timed out transaction log records

mattlord:
## Description
This redacts potentially sensitive query parameters — if you have the `-queryserver-config-terse-errors` tablet flag enabled — from the transaction related log messages while still printing the statement shapes (normalized queries) which can still be helpful in debugging.

Before:
```
W0105 17:18:53.558389    3783 tx_pool.go:126] killing transaction (exceeded timeout: 30s): 1641402138560038177  'principal:"vitess" component:"127.0.0.1:56294" subcomponent:"VTGate MySQL Connector"'  'username:"userData1"'  Jan  5 17:18:22.025067  Jan  1 00:00:00.000000  -9223372036.854776              insert into customer(email) values ('PROBLEM@LOGGING.WHY');insert into customer(email) values ('PROBLEM2@LOGGING.WHY')
```

After:
```
W0105 17:36:01.783113     693 tx_pool.go:126] killing transaction (exceeded timeout: 30s): 1641404066211055126	'principal:"vitess" component:"127.0.0.1:56520" subcomponent:"VTGate MySQL Connector"'	'username:"userData1"'	Jan  5 17:35:30.360499	Jan  1 00:00:00.000000	-9223372036.854776		insert into customer(email) values (:redacted1);insert into customer(email) values (:redacted1)
```

## Related Issue(s)
Fixes: https://github.com/vitessio/vitess/issues/9469

## Checklist

### DIFF
--- a/go/vt/vttablet/tabletserver/tx/api.go
+++ b/go/vt/vttablet/tabletserver/tx/api.go
@@ -57,6 +57,9 @@ type (
 		Autocommit      bool
 		Conclusion      string
 		LogToFile       bool
+		// Redact query parameters in generic struct output such as .String()
+		// that may be logged or returned to the client
+		RedactQueries bool
 
 		Stats *servenv.TimingsWrapper
 	}
@@ -126,7 +129,7 @@ func (p *Properties) RecordQuery(query string) {
 func (p *Properties) InTransaction() bool { return p != nil }
 
 // String returns a printable version of the transaction
-func (p *Properties) String(redactQueries bool) string {
+func (p *Properties) String() string {
 	if p == nil {
 		return ""
 	}
@@ -134,7 +137,7 @@ func (p *Properties) String(redactQueries bool) string {
 	printQueries := func() string {
 		sb := strings.Builder{}
 		for _, query := range p.Queries {
-			if redactQueries {
+			if p.RedactQueries {
 				query, _ = sqlparser.RedactSQLQuery(query)
 			}
 			sb.WriteString(query)
--- a/go/vt/vttablet/tabletserver/tx/api.go
+++ b/go/vt/vttablet/tabletserver/tx/api.go
@@ -24,6 +24,7 @@ import (
 	querypb "vitess.io/vitess/go/vt/proto/query"
 	vtrpcpb "vitess.io/vitess/go/vt/proto/vtrpc"
 	"vitess.io/vitess/go/vt/servenv"
+	"vitess.io/vitess/go/vt/sqlparser"
 )
 
 type (
@@ -130,6 +131,12 @@ func (p *Properties) String() string {
 		return ""
 	}
 
+	redactedQueries := make([]string, len(p.Queries))
+	for i, query := range p.Queries {
+		rq, _ := sqlparser.RedactSQLQuery(query)
+		redactedQueries[i] = rq
+	}
+
 	return fmt.Sprintf(
 		"'%v'\t'%v'\t%v\t%v\t%.6f\t%v\t%v\t\n",
 		p.EffectiveCaller,
@@ -138,6 +145,6 @@ func (p *Properties) String() string {
 		p.EndTime.Format(time.StampMicro),
 		p.EndTime.Sub(p.StartTime).Seconds(),
 		p.Conclusion,
-		strings.Join(p.Queries, ";"),
+		strings.Join(redactedQueries, ";"),
 	)
 }
--- a/go/vt/vttablet/tabletserver/tx/api.go
+++ b/go/vt/vttablet/tabletserver/tx/api.go
@@ -21,7 +21,6 @@ import (
 	"strings"
 	"time"
 
-	"vitess.io/vitess/go/streamlog"
 	querypb "vitess.io/vitess/go/vt/proto/query"
 	vtrpcpb "vitess.io/vitess/go/vt/proto/vtrpc"
 	"vitess.io/vitess/go/vt/servenv"
@@ -127,7 +126,7 @@ func (p *Properties) RecordQuery(query string) {
 func (p *Properties) InTransaction() bool { return p != nil }
 
 // String returns a printable version of the transaction
-func (p *Properties) String() string {
+func (p *Properties) String(redactQueries bool) string {
 	if p == nil {
 		return ""
 	}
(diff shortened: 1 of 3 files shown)


harshit-gangal:
Isn't this make it expensive to log the query. We can log the query received by the tablet instead of the one sent to MySQL, that should also not contain the values and should just have arguments there.

mattlord:
> Isn't this make it expensive to log the query. We can log the query received by the tablet instead of the one sent to MySQL, which should also not contain the values and should just have arguments there.

I'm not sure what you mean. The stateful_session struct has a slice of strings for the SQL statements executed within it: https://github.com/vitessio/vitess/blob/main/go/vt/vttablet/tabletserver/tx/api.go#L116-L122

That is what gets printed in the `String()` function. At that point we have no concept of "what was received by the tablet or sent to mysql". Or are you asking that we change how the statements are recorded in the stateful session?

harshit-gangal:
Maybe it is better to record the planner query `qre.plan.FullQuery` than we do not have to parse the query again for redaction.

https://github.com/vitessio/vitess/blob/main/go/vt/vttablet/tabletserver/query_executor.go#L738-L753
```go
// txFetch fetches from a TxConnection.
func (qre *QueryExecutor) txFetch(conn *StatefulConnection, record bool) (*sqltypes.Result, error) {
	sql, _, err := qre.generateFinalSQL(qre.plan.FullQuery, qre.bindVars)
	if err != nil {
		return nil, err
	}
	qr, err := qre.execStatefulConn(conn, sql, true)
	if err != nil {
		return nil, err
	}
	// Only record successful queries.
	if record {
		conn.TxProperties().RecordQuery(sql)
	}
	return qr, nil
}
```

mattlord:
> Maybe it is better to record the planner query `qre.plan.FullQuery` so that we do not have to parse the query again for redaction.

@harshit-gangal Unfortunately we cannot do that. The recorded queries are redo logged when performing the first phase (prepare) for 2pc transactions: https://github.com/vitessio/vitess/blob/main/go/vt/vttablet/tabletserver/tx_executor.go#L44-L76

And then replayed from that redo log when executing the second phase of the transaction (finalize: commit or rollback): https://github.com/vitessio/vitess/blob/27367ea3eafc4d07daec11db5e1a669773e14928/go/vt/vttablet/tabletserver/tx_engine.go#L357-L414

It's also worth nothing that the `.String()` call is only used today for a debug log messages when a transaction is killed due to hitting the configured transaction timeout here: https://github.com/vitessio/vitess/blob/4c33b2b5a90dc200e6cc649681cbace2520725bd/go/vt/vttablet/tabletserver/tx_pool.go#L126

So it should not be too often and should not be performance sensitive (by default the trx has already been running for 30 seconds).

harshit-gangal:
The current 2PC code path in vttablet will most likely be deleted when cross-shard transactions are reimplemented. 
I also agree that the transaction killer should come less in the system. I was just looking to avoid overhead.

cc: @sougou what are your thoughts?

go/vt/vttablet/tabletserver/tx/api.go
```diff
@@ -138,6 +145,6 @@ func (p *Properties) String() string {
 		p.EndTime.Format(time.StampMicro),
 		p.EndTime.Sub(p.StartTime).Seconds(),
 		p.Conclusion,
-		strings.Join(p.Queries, ";"),
+		strings.Join(redactedQueries, ";"),
```
mattlord, line 148: We could also simply Omit the queries from the `.String()` output here.

sougou:
Ideally, it should follow the rule provided by the flag, unless `redactedQueries` honors the flag. For those who are willing to tolerate unredacted queries in the logs, the values are actually useful for identifying hot rows, which won't be possible in the redacted case.

mattlord:
> Ideally, it should follow the rule provided by the flag, unless `redactedQueries` honors the flag. For those who are willing to tolerate unredacted queries in the logs, the values are actually useful for identifying hot rows, which won't be possible in the redacted case.

OK, I'll explore this avenue. The current flag (`-redact-debug-ui-queries`) is about the `/debug/` tablet web UI views but we can re-use that or add a new flag.

mattlord:
We're now re-using `-redact-debug-ui-queries` here. That feels somewhat wrong to me, but it works. :-) The other option would be to add a new flag ~ `-redact-queries-in-logs` and apply that in any location where we may log the full query (not counting the query log itself).

deepthi:
> We're now re-using `-redact-debug-ui-queries` here. That feels somewhat wrong to me, but it works. 🙂 The other option would be to add a new flag ~ `-redact-queries-in-logs` and apply that in any location within the tablet where we may log the full query (not counting the query log itself). If we go that route, this PR would have to be closed or repurposed for that more general new feature.

I agree that we are overloading the existing flag. Maybe we should be renaming it to `redact-queries` to reflect how it actually works.

mattlord:
> I agree that we are overloading the existing flag. Maybe we should be renaming it to `redact-queries` to reflect how it actually works.

Yeah, IMO that makes sense. At that point though I think we'd probably need to do a review of all tablet code to ensure that we are redacting them everywhere (the original case here may not be the only one). I could deprecate the old flag name, add the new one, and make the old one an alias effectively in the PR if everything agrees. Then look for any other places where we should redact queries.

mattlord:
Another potential alternative was discussed and that is using another existing flag here:
```
-queryserver-config-terse-errors  (prevent bind vars from escaping in returned errors)
``` 

That at least feels *less* bad...

mattlord:
I pushed another commit that uses `-queryserver-config-terse-errors` to set a new parameter on the transaction pool properties called `RedactQueries`. I actually don't hate this as the messages/output can be logged, sent to clients, or other consumers/listeners. So we're saying that we want this output, which in practice here seem to be warnings/errors, to be terse (which already seems to mean redacted). 

If we don't like this though, then the next option is to create a new flag and use that in the same manner, but then also look for any other locations where we may need to redact queries in output.

deepthi (approved):
Let's merge this PR and revisit a couple of things later:
1. We already know that the `terse-errors` flag is also hiding the "original" mysql error from clients. I believe @harshit-gangal is looking at this.
2. Review the `redact-debug-ui-queries` flag and decide whether to rename/extend it.

sougou:
I looked at the code. There are many places where we print the query without redacting the values. This means that the problem needs to be solved more holistically. Fortunately, all those places seem to be invoking `TruncateForLog`. So, it may not be too hard.

Given this, it makes sense to introduce a separate flag for logs. If we overload existing flags, it may become too confusing in the future.

mattlord:
Thank you, @deepthi, @sougou, and @harshit-gangal for the helpful feedback! I'll create a new flag and then apply that in all the places I can find within tablet server where we might log queries.

PR #9471, opened 2022-01-06
Move reparent and vtorc tests to self-hosted runners

GuptaManan100:
## Description

`reparent` and `vtorc` tests have been flaky recently. This PR moves them to the self-hosted runners so that the flakiness can be traced and reproduced.

## Checklist


frouioui (commented):
Can we keep launchable in self-hosted workflows?

harshit-gangal (approved):

harshit-gangal:
> Can we keep launchable in self-hosted workflows?

Not yet.

PR #9481, opened 2022-01-06
Use LOCK TABLES during SwitchWrites to prevent lost writes

mattlord:
Issuing a [`LOCK TABLES`](https://dev.mysql.com/doc/refman/en/lock-tables.html) statement on all of the tables that we are doing a [SwitchWrites](https://vitess.io/docs/reference/vreplication/v1/switchwrites/) operation for as part of a [MoveTables](https://vitess.io/docs/reference/vreplication/movetables/) workflow ensures that no lost writes occur due to long running transactions (including long running single statement auto-commit transactions) on the source tablets committing after writes have been switched. I say this because:
1.  During [SwitchWrites](https://vitess.io/docs/reference/vreplication/v1/switchwrites/) we first we update the `DenyList` on the source tablets to prevent any new writes against the tables [at the tablet level in the query planner/executor](https://github.com/vitessio/vitess/blob/05afb39dcb70b1dd51ae6b65a5815d572fb39eb9/go/vt/vttablet/tabletserver/query_executor.go#L373-L389).
2. We then wait on a [`LOCK TABLES`](https://dev.mysql.com/doc/refman/en/lock-tables.html) call to return on the source mysqld's which will in turn wait for any open/running RW statements/transactions on those tables to complete or time out.
3. Af this point, after steps 1 and 2, we know that any existing writes against the source tables have completed and no new ones can come in. 
4. We then collect the vreplication metadata snapshot, put the reverse vreplication streams in place, and update the routing rules.
  
Without doing this (before this PR is merged) you can not only break reverse vreplication -- which would have to be fixed manually -- but **_if you don't notice the errant writes and then manually reconcile the tables across keyspaces (old and new) then those writes are lost_**.

With this patch, when you go through [the manual test here](https://github.com/vitessio/vitess/issues/9400) and get to the point where you attempt another `insert` or `commit` in the previously opened transaction started against the source before we switched writes, you get an error like this:
```
ERROR 1317 (70100): vttablet: rpc error: code = Aborted desc = transaction 1641661850058067114: ended at 2022-01-08 17:12:17.090 UTC (exceeded timeout: 30s) (CallerID: userData1)
```

## Related Issue(s)
Fixes: https://github.com/vitessio/vitess/issues/9400

## Checklist
- [x] Should this PR be backported? At least to 12.0, I think...
- [x] Tests were added
- [x] Documentation: I don't think we need any


go/vt/wrangler/traffic_switcher.go
```diff
@@ -58,9 +59,14 @@ const (
 // accessType specifies the type of access for a shard (allow/disallow writes).
 type accessType int
 
+// tableLockAction specifies if you want to lock or unlock tables
+type tableLockAction int
```
shlomi-noach, line 63: Let's make this a strongly typed `iota` to have better clarity, then have something like `tableLockActionAquire` and `tableLockActionRelease`. That way we don't need to define `tableLockAcquire = tableLockAction(0)` etc as integers.

go/vt/wrangler/traffic_switcher.go
```diff
@@ -963,7 +1000,9 @@ func (ts *trafficSwitcher) checkJournals(ctx context.Context) (journalsExist boo
 func (ts *trafficSwitcher) stopSourceWrites(ctx context.Context) error {
 	var err error
 	if ts.MigrationType() == binlogdatapb.MigrationType_TABLES {
-		err = ts.changeTableSourceWrites(ctx, disallowWrites)
+		if err = ts.changeTableSourceWrites(ctx, disallowWrites); err == nil {
+			err = ts.manageTableLocksSourceWrites(ctx, tableLockAcquire)
```
shlomi-noach, line 1004: I feel like there has to be a `defer ts.manageTableLocksSourceWrites(ctx, tableLockRelease)`; since here is probably not the right function to have this `defer`, maybe the calling function should do it. Decoupling the two altogether makes me uneasy, because the code could break and we would not release the lock.

mattlord: Good call. I just added/improved error handling and added a deferred func for this there. Thank you!

DeathBorn:
Before this implementation we found this manual workaround
` LOCK TABLES <all_tables_that_move_to_other_shard> READ;`

go/vt/wrangler/traffic_switcher.go
```diff
@@ -995,6 +1018,56 @@ func (ts *trafficSwitcher) changeTableSourceWrites(ctx context.Context, access a
...
+			return err
+		})
+	case tableLockCheck:
+		ts.Logger().Infof(logStrf, "CHECK")
+		// This will return an error if we still hold the table locks:
+		// ERROR 1100 (HY000): Table 'vreplication' was not locked with LOCK TABLES
+		//query = "select 1 from _vt.vreplication"
+		// If it doesn't return an error then we've somehow lost our session and the
+		// locks along with it
+		//expectError = true
```
mattlord, line 1063: Should I just remove this for now? I could also add a new RPC for this on the tablet side.

go/vt/wrangler/traffic_switcher.go
```diff
@@ -493,13 +496,31 @@ func (wr *Wrangler) SwitchWrites(ctx context.Context, targetKeyspace, workflowNa
...
+
 		ts.Logger().Infof("Stopping source writes")
 		if err := sw.stopSourceWrites(ctx); err != nil {
 			ts.Logger().Errorf("stopSourceWrites failed: %v", err)
 			sw.cancelMigration(ctx, sm)
 			return 0, nil, err
 		}
 
+		if ts.MigrationType() == binlogdatapb.MigrationType_TABLES {
+			lockTablesCount := 2
```
harshit-gangal, line 508: nit: can move this as constant outside.

go/vt/wrangler/traffic_switcher.go
```diff
@@ -493,13 +496,31 @@ func (wr *Wrangler) SwitchWrites(ctx context.Context, targetKeyspace, workflowNa
...
+			// the tablet's deny list check and the first mysqld side table lock.
+			for cnt := 1; cnt <= lockTablesCount; cnt++ {
+				if err := ts.executeLockTablesOnSource(ctx); err != nil {
+					ts.Logger().Errorf("Failed to execute LOCK TABLES (attempt %d of %d) on sources: %v", cnt, lockTablesCount, err)
+					sw.cancelMigration(ctx, sm)
+					return 0, nil, err
+				}
+				// No need to UNLOCK the tables as the connection was closed once the locks were acquired
+				// and thus the locks released.
+				time.Sleep(1 * time.Second)
```
harshit-gangal, line 520: nit: move the timeout value as constant outside.

go/vt/wrangler/traffic_switcher.go
```diff
@@ -995,6 +1016,38 @@ func (ts *trafficSwitcher) changeTableSourceWrites(ctx context.Context, access a
...
 }
 
+// executeLockTablesOnSource executes a LOCK TABLES tb1 READ, tbl2 READ,... statement on each
+// source shard's primary tablet using a non-pooled connection as the DBA user. The connection
+// is closed when the LOCK TABLES statement returns, so we immediately release the LOCKs.
+func (ts *trafficSwitcher) executeLockTablesOnSource(ctx context.Context) error {
+	ts.Logger().Infof("Locking (and then immediately unlocking) the following tables on source keyspace %v: %v", ts.SourceKeyspaceName(), ts.Tables())
+	if ts.Tables() == nil || len(ts.Tables()) == 0 {
+		return vterrors.Errorf(vtrpc.Code_INTERNAL, "no tables found in the source keyspace %v associated with the %s workflow", ts.SourceKeyspaceName(), ts.WorkflowName())
+	}
```
harshit-gangal, line 1026: nit: length check is enough, it also handle nil

harshit-gangal (commented):
looks good overall.
some nits.

go/vt/wrangler/traffic_switcher.go
```diff
@@ -995,6 +1016,38 @@ func (ts *trafficSwitcher) changeTableSourceWrites(ctx context.Context, access a
...
+	// trim extra trailing comma
+	lockStmt := sb.String()[:sb.Len()-1]
+
+	return ts.ForAllSources(func(source *workflow.MigrationSource) error {
+		primary := source.GetPrimary()
+		if primary == nil {
+			return vterrors.Errorf(vtrpc.Code_INTERNAL, "no primary found for source shard %s", source.GetShard())
+		}
+		tablet := primary.Tablet
+		_, err := ts.wr.ExecuteFetchAsDba(ctx, tablet.Alias, lockStmt, 1, false, true)
```
rohit-nayak-ps, line 1042: So `ExecuteFetchAsDba` will release the locks automatically since it doesn't use a pool and the connection will be closed after the locking succeeds, right?

go/test/endtoend/vreplication/cluster.go
```diff
@@ -27,6 +27,7 @@ var (
 	mainClusterConfig     *ClusterConfig
 	externalClusterConfig *ClusterConfig
 	extraVTGateArgs       = []string{"-tablet_refresh_interval", "10ms"}
+	extraVtctldArgs       = []string{"-remote_operation_timeout", "600s", "-topo_etcd_lease_ttl", "120"}
```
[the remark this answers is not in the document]
mattlord: Not RPC related, the operation timeout gives us more time for VDiffs etc as we add new tests, tables get larger, CI is slow, etc. The etcd lease extension ensures that the keyspace lock does not timeout during the two TABLE LOCKs if a transaction is timed out in each one (30s trx timeout). I made them large simply to avoid flappy tests. 

I can remove them though if you want? They're not strictly needed here.

go/vt/vttablet/tabletmanager/rpc_lock_tables.go
```diff
@@ -115,6 +116,14 @@ func (tm *TabletManager) lockTablesUsingLockTables(conn *dbconnpool.DBConnection
...
 	}
 
+	// Doing this in an implicit multi-statement transaction is needed for the desired
+	// behavior with InnoDB tables. It ensures that InnoDB maintains an internal table
+	// lock on each table until we explicitly release them. See:
+	//   https://dev.mysql.com/doc/refman/8.0/en/lock-tables.html#lock-tables-and-transactions
+	if _, err = conn.ExecuteFetch("SET autocommit=0", 0, false); err != nil {
+		return err
+	}
+
```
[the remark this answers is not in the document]
mattlord: You want to do this but I'm going to restore this file from main just to touch as little as needed.

go/vt/wrangler/traffic_switcher.go
```diff
@@ -47,6 +48,7 @@ import (
 	querypb "vitess.io/vitess/go/vt/proto/query"
 	topodatapb "vitess.io/vitess/go/vt/proto/topodata"
 	vschemapb "vitess.io/vitess/go/vt/proto/vschema"
+	"vitess.io/vitess/go/vt/proto/vtrpc"
```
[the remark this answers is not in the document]
mattlord: Good catch. Thanks!

go/vt/wrangler/traffic_switcher.go
mattlord: Right. I know this for sure because initially I _*wanted*_ to hold the locks and saw that they were getting released immediately and the LOCK/UNLOCK statements were done with different mysql connection IDs.

rohit-nayak-ps: Cool, maybe we can add a comment to that effect here.

rohit-nayak-ps (approved):
ltgm

mattlord:
Thank you for the review and comments, @rohit-nayak-ps and @harshit-gangal ! I just pushed another commit that addresses them. ❤️

go/test/endtoend/vreplication/cluster.go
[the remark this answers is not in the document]
harshit-gangal: If it helps keep the tests less flaky, should be fine to keep it.

harshit-gangal (approved):

Issue #9218, opened 2021-11-12
Task: delete deprecated discoverygateway and legacy healthcheck code


deepthi:
As part of #5750, we implemented a new gateway called `tabletgateway` and deprecated the old one called `discoverygateway`.
We have had several releases since then, and `tabletgateway` is now stable enough. We can now start deleting some of the old code.

Blocker: vtctld ui still uses legacy healthcheck. We need to either move that to the new healthcheck, or deprecate/delete that first.

Tasks
- [x] In release 13.0, deprecate and ignore `-gateway_implementation` flag #9482 
- [x] delete discoverygateway.go #9500 
- [x] Optionally also delete `Gateway` interface #9852
- [x] fix vtctld realtime_status to not use legacy healthcheck #9358
- [x] delete legacy_healthcheck.go and any related structs that are no longer needed https://github.com/vitessio/vitess/issues/10454
- [x] In release 14.0 delete `-gateway_implementation` flag #9852

EDIT Mar 8, 2022:
Split deletion into two separate tasks because of the dependency on #9358

deepthi:
All tasks are now complete. Thank you for all your work on this @frouioui!

PR #9482, opened 2022-01-07
Deprecate and ignore `gateway_implementation` flag to vtgate.

deepthi:
## Description

VTGate's `gateway_implementation` flag is now deprecated, and any value provided is ignored. `tabletgateway` is now the only supported implementation.

## Related Issue(s)

#9218 

## Checklist
- [ ] Documentation was added or is not required - release note has been added. Website PR will be created

## Deployment Notes

Scripts used to run `vtgate` will need to be updated to drop this flag. It will be removed in a future release (14.0)

### DIFF
--- a/go/vt/vtgate/gateway.go
+++ b/go/vt/vtgate/gateway.go
@@ -37,8 +37,8 @@ import (
 var (
 	_ = flag.String("gateway_implementation", "", "Deprecated. Only tabletgateway is now supported, discoverygateway is no longer available")
 	// We cannot reference tabletGatewayImplementation directly because it is const
-	DefaultGatewayImplementation = tabletGatewayImplementation
-	GatewayImplementation        = &DefaultGatewayImplementation
+	defaultGatewayImplementation = tabletGatewayImplementation
+	GatewayImplementation        = &defaultGatewayImplementation
 	bufferImplementation         = flag.String("buffer_implementation", "keyspace_events", "Allowed values: healthcheck (legacy implementation), keyspace_events (default)")
 	initialTabletTimeout         = flag.Duration("gateway_initial_tablet_timeout", 30*time.Second, "At startup, the gateway will wait up to that duration to get one tablet per keyspace/shard/tablettype")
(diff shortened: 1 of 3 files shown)


go/vt/vtgate/gateway.go
```diff
@@ -35,8 +35,8 @@ import (
 // a query targeted to a keyspace/shard/tablet_type and send it off.
 
 var (
-	// GatewayImplementation allows you to choose which gateway to use for vtgate routing. Defaults to tabletgateway, other option is discoverygateway
-	GatewayImplementation = flag.String("gateway_implementation", "tabletgateway", "Allowed values: discoverygateway (deprecated), tabletgateway (default)")
+	_                     = flag.String("gateway_implementation", "", "Deprecated. Only tabletgateway is now supported, discoverygateway is no longer available")
+	GatewayImplementation = new(string)
```
vmg, line 39: just make this `GatewayImplementation = &tabletGatewayImplementation` instead of assigning to it from an `init` function.

vmg (changes requested):

go/vt/vtgate/gateway.go
deepthi: That doesn't work because `tabletGatewayImplementation` is declared as a const.
```
go/vt/vtgate/gateway.go:39:26: cannot take the address of tabletGatewayImplementation
```

deepthi: I could change the type of `GatewayImplementation` from `string*` to `string` and change all references, but that will touch lots of files. I'm looking for the minimum necessary change.

vmg: @deepthi I see, in that case the minimal change would be making a temp assignment right next to it.

```
var defaultGatewayImplementation = tabletGatewayImplementation
var GatewayImplementation = &defaultGatewayImplementation
```

That's the minimal amount of changes and doesn't allocate on heap or use `init`,

vmg (approved):

PR #9487, opened 2022-01-10
Gen4Fallback bug fix

systay:
## Description
A minor planner fix

When using the `Gen4fallback` planner, we were not protecting the AST from changes from the Gen4 planner, which lead to the V3 planner getting an already updated AST and failing

### DIFF
--- a/go/vt/vtgate/planbuilder/abstract/join.go
+++ b/go/vt/vtgate/planbuilder/abstract/join.go
@@ -58,8 +58,8 @@ func (j *Join) PushPredicate(expr sqlparser.Expr, semTable *semantics.SemTable) 
 			j.LeftJoin = false
 			return j.RHS.PushPredicate(expr, semTable)
 		}
-		// TODO - we should do this on the vtgate level once we have a Filter primitive
-		return vterrors.Errorf(vtrpcpb.Code_UNIMPLEMENTED, "unsupported: cross-shard left join and where clause")
+		semTable.ProjectionErr = vterrors.Errorf(vtrpcpb.Code_UNIMPLEMENTED, "unsupported: cross-shard left join and where clause")
+		return nil
 	case deps.IsSolvedBy(j.LHS.TableID().Merge(j.RHS.TableID())):
 		j.Predicate = sqlparser.AndExpressions(j.Predicate, expr)
 		return nil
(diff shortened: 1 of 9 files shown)


GuptaManan100 (approved):

go/vt/vtgate/planbuilder/abstract/join.go
```diff
@@ -58,8 +58,8 @@ func (j *Join) PushPredicate(expr sqlparser.Expr, semTable *semantics.SemTable)
 			j.LeftJoin = false
 			return j.RHS.PushPredicate(expr, semTable)
 		}
-		// TODO - we should do this on the vtgate level once we have a Filter primitive
-		return vterrors.Errorf(vtrpcpb.Code_UNIMPLEMENTED, "unsupported: cross-shard left join and where clause")
+		semTable.ProjectionErr = vterrors.Errorf(vtrpcpb.Code_UNIMPLEMENTED, "unsupported: cross-shard left join and where clause")
```
GuptaManan100, line 61: We should rename this from `ProjectionErr` to something else, since it now contains errors that are more general than just projection errors

harshit-gangal (commented):
looks good, need a test for it :)

go/vt/vtgate/planbuilder/abstract/join.go
systay: :+1:

go/vt/vtgate/planbuilder/abstract/join.go
```diff
@@ -58,8 +58,7 @@ func (j *Join) PushPredicate(expr sqlparser.Expr, semTable *semantics.SemTable)
 			j.LeftJoin = false
 			return j.RHS.PushPredicate(expr, semTable)
 		}
-		semTable.ProjectionErr = vterrors.Errorf(vtrpcpb.Code_UNIMPLEMENTED, "unsupported: cross-shard left join and where clause")
-		return nil
```
GuptaManan100: Why revert this change? Isn't it okay to keep it as a `ShardedErr`?

systay: Well, we have to either return an error, or build something that adds the predicate to the operator tree. At the moment we don't have a filter operator, so this will be easy to fix post-remove-query-trees

PR #9494, opened 2022-01-10
Add stats to track when a pool's waiter queue is full

dasl-:
## Description

Stats should be available at the tablet's `/debug/vars` page

## Related Issue(s)

Follow-up to: https://github.com/vitessio/vitess/pull/9484

## Checklist
- [x] Tests are not required

### DIFF
--- a/go/vt/vttablet/tabletserver/connpool/pool.go
+++ b/go/vt/vttablet/tabletserver/connpool/pool.go
@@ -59,6 +59,7 @@ type Pool struct {
 	idleTimeout        time.Duration
 	waiterCap          int64
 	waiterCount        sync2.AtomicInt64
+	waiterQueueFull    sync2.AtomicInt64
 	dbaPool            *dbconnpool.ConnectionPool
 	appDebugParams     dbconfigs.Connector
 }
@@ -90,6 +91,7 @@ func NewPool(env tabletenv.Env, name string, cfg tabletenv.ConnPoolConfig) *Pool
 	env.Exporter().NewGaugeDurationFunc(name+"IdleTimeout", "Tablet server idle timeout", cp.IdleTimeout)
 	env.Exporter().NewCounterFunc(name+"IdleClosed", "Tablet server conn pool idle closed", cp.IdleClosed)
 	env.Exporter().NewCounterFunc(name+"Exhausted", "Number of times pool had zero available slots", cp.Exhausted)
+	env.Exporter().NewCounterFunc(name+"WaiterQueueFull", "Number of times the waiter queue was full", cp.waiterQueueFull.Get)
 	return cp
 }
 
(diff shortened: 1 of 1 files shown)


go/vt/vttablet/tabletserver/connpool/pool.go
```diff
@@ -160,6 +162,7 @@ func (cp *Pool) Get(ctx context.Context) (*DBConn, error) {
 		waiterCount := cp.waiterCount.Add(1)
 		defer cp.waiterCount.Add(-1)
 		if waiterCount > cp.waiterCap {
+			cp.waiterQueueFull.Add(1)
 			return nil, vterrors.Errorf(vtrpcpb.Code_RESOURCE_EXHAUSTED, "pool %s waiter count exceeded", cp.name)
 		}
 	}
```
dasl-, line 168: I noticed that this new metric won't be added to the `StatsJSON` output -- does that matter?

mattlord: Can you add it here, please?

dasl-: I'm not sure that's possible the way things are architected currently. In the code you linked, we only have access to the ResourcePool struct. But the metric we are adding here is only present on the Pool struct.

mattlord: Here's one potential idea: https://gist.github.com/mattlord/b305eabac4ab14c913fc9796538319d5

mattlord: Here's another rough idea: https://gist.github.com/mattlord/01dbb1c52c8c2fc84a103de1869e82df

mattlord: Here's another rough idea: https://gist.github.com/mattlord/684656143768d692b71402ff796e98c0

mattlord: Perhaps someone else has an even better idea 🙂

GuptaManan100: I agree with @mattlord that we should include them in the `StatsJSON` function. The easiest and IMO the cleanest solution is to just append to the output of the underlying call here before returning https://github.com/vitessio/vitess/blob/904c9a956028d58e5c4f666dfeaa23042ef482cf/go/vt/vttablet/tabletserver/connpool/pool.go#L233-L239

So it reads like -
```go
func (cp *Pool) StatsJSON() string { 
 	p := cp.pool() 
 	if p == nil { 
 		return "{}" 
 	} 
 	res:= p.StatsJSON()
 	res += fmt.Sprintf(`"WaiterQueueFull": %v`, p.waiterQueueFull.Get())
 	return res
}
```

GuptaManan100 (commented):
Once that comment is fixed, everything else LGTM!

dasl-:
@mattlord @GuptaManan100 thanks for the feedback. I've updated the PR based on your suggestions.

mattlord (approved):

GuptaManan100 (approved):

PR #9496, opened 2022-01-11
mysql: Add ConnectionReady callback to Handler interface

mattrobenolt:
This is is useful to determine the intermediate state after
NewConnection and before ConnectionClosed, basically for a handler to
know if authentication, etc, were successful. NewConnection is way too
soon in the process to determine if the connection got fully
established.

Signed-off-by: Matt Robenolt <matt@ydekproductions.com>

## Checklist

### DIFF
--- a/go/mysql/conn_flaky_test.go
+++ b/go/mysql/conn_flaky_test.go
@@ -820,6 +820,10 @@ func (t testRun) NewConnection(c *Conn) {
 	panic("implement me")
 }
 
+func (t testRun) ConnectionReady(c *Conn) {
+	panic("implement me")
+}
+
 func (t testRun) ConnectionClosed(c *Conn) {
 	panic("implement me")
 }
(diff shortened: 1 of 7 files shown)


go/mysql/conn_flaky_test.go
```diff
@@ -820,6 +820,10 @@ func (t testRun) NewConnection(c *Conn) {
 	panic("implement me")
 }
 
+func (t testRun) ConnectionReady(c *Conn) {
```
mattrobenolt, line 823: This is just filling in all of the other internal implementations of the `Handler` interface.

When doing this, I was thinking it might be nice to introduce a stub `UnimplementedHandler` struct that all implementation can embed that provides default no-op handlers. Similar to how gRPC gives you the stub to satisfy the interface. That way an implementation isn't _require_ to implement every callback if they're not needed.

I'd be happy to follow up with an `UnimplementedHandler` struct if you also think this is valuable. It'd help wiht the boilerplate in the tests and all that get touched in this PR.

mattrobenolt:
@frouioui correct, within vitess core, there's no use for it. It's a no-op hook so other implementations of `mysql.Handler` can use it. In our case, we need it for PlanetScale.

systay (approved):
Sorry you had to wait so long for this to get reviewed

PR #9501, opened 2022-01-11
Upgrade Downgrade testing for backups

frouioui:
## Checklist


examples/local/backups/take_backups.sh
```diff
@@ -0,0 +1,24 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# this script brings up new tablets for the two new shards that we will
+# be creating in the customer keyspace and copies the schema
+
+source ./env.sh
+
+vtctlclient BackupShard -allow_primary=true customer/-80
```
deepthi, line 22: The clusters we start have 3 tablets each. Is there a reason we need to pass `-allow_primary`?

deepthi (commented):
This is very good work.
I'm just wondering if the cluster we are creating needs to be this complicated. We are running MoveTables and Reshard on that cluster before we even start the backup flow.
Can we not simply create the sharded cluster we want without going through all these workflows? A problem in any of them can fail the upgrade/downgrade test (false positive).
I'm willing to consider having this merged as-is if you can simplify it later.

examples/local/backups/take_backups.sh
GuptaManan100: I don't think it should be required. The command first tries to find a tablet which is of type REPLICA or RDONLY. If it doesn't find anything then it checks that flag and selects a PRIMARY tablet. If we have a running REPLICA tablet, then it should not matter wether we provide this flag or not.

frouioui: Right, thanks for the explanation on how the flag works internally! I will try removing it

GuptaManan100 (approved):
Apart from the requested changes, everything else looks good to me!

frouioui:
@deepthi @GuptaManan100 - Thank you both for the review. Actually @deepthi you are right, the setup of the cluster was inefficient, via commit da541cd I simplified the whole process, going from ~2-5 mins to <1 min to set up the cluster and with a process that is less prone to errors

Issue #7883, opened 2021-04-16
Move Tables workflow errors out for source tables that have a column of datetime type


bharathramaprasad-okta:
#### Overview of the Issue

Move tables workflow throws the following error : Move Tables Error: rpc error: code = Unknown desc = Table Manager.Apply Schema on zone1- error: /usr/bin/mysql: exit status 1, output: ERROR 1067 (42000) at line 45: Invalid default value for 'eventTimestamp'
Each source table that has a eventTImestamp of datetime type throws this error. 

desc table of one of the tables for which movetable fails is show in the image below.

![image (2)](https://user-images.githubusercontent.com/77690797/115060935-5222d280-9e9d-11eb-9d57-d00e211399f5.png)

#### Reproduction Steps

Steps to reproduce this issue :

1. Use the latest vitess 8 version
2. Follow example scripts for docker local setup.
2. Deploy the below `schema` :

CREATE TABLE `Events` (
  `id` bigint(20) NOT NULL AUTO_INCREMENT,
  `eventId` char(25) NOT NULL,
  `eventTimestamp` datetime NOT NULL DEFAULT '0000-00-00 00:00:00',
   `message` varchar(2048) NOT NULL,
   PRIMARY KEY (`id`,`eventTimestamp`),
  UNIQUE KEY `idx_eventId_timestamp` (`eventId`,`eventTimestamp`),
  KEY `eventtimestamp` (`eventTimestamp`)
)  ENGINE=InnoDB DEFAULT CHARSET=utf8
/*!50100 PARTITION BY LIST (EXTRACT(YEAR_MONTH FROM eventTimestamp))
(PARTITION m202101 VALUES IN (202101) ENGINE = InnoDB,
 PARTITION m202102 VALUES IN (202102) ENGINE = InnoDB,
 PARTITION m202103 VALUES IN (202103) ENGINE = InnoDB,
 PARTITION m202104 VALUES IN (202104) ENGINE = InnoDB,
 PARTITION m202105 VALUES IN (202105) ENGINE = InnoDB,
 PARTITION m202106 VALUES IN (202106) ENGINE = InnoDB,
 PARTITION m202107 VALUES IN (202107) ENGINE = InnoDB,
 PARTITION m202108 VALUES IN (202108) ENGINE = InnoDB,
 PARTITION m202109 VALUES IN (202109) ENGINE = InnoDB,
 PARTITION m202110 VALUES IN (202110) ENGINE = InnoDB,
[...]

PR #6561, opened 2020-08-13
VReplication: allow invalid (zero) values in date columns

rohit-nayak-ps:
While migrating with MoveTables, it is likely that the source database allowed zero values in dates (like 0000-00-00). The default connection options treat these as invalid and vreplication fails on rows with zero date values at the moment. The user's app will work fine since they will (presumably) be setting the connection options in the app (or globally) to allow invalid dates. 

We need to relax the sql options to allow invalid dates during all phases of vreplication since we will find invalid dates both during copy as well as when we stream (the 'replicate' phase).

Signed-off-by: Rohit Nayak <rohit@planetscale.com>

PR #9505, opened 2022-01-12
VReplication Workflow: manage SQL_MODE as MySQL does in mysqldump

rohit-nayak-ps:
## Description

In VReplication we are tasked with taking a single logical table from a source and creating a logical copy of that on a target (typically partitioned/sharded in different ways on each side). In order to do this, we need to relax the `sql_mode` in order to ensure we can do so (the source could have any `sql_mode` configured). 

Given that VReplication logically implements the same functionality in Vitess that [`mysqldump`](https://dev.mysql.com/doc/refman/8.0/en/mysqldump.html) does in MySQL — generating SQL statements from a source that allow you to recreate database objects on a target — it makes sense for us to try and mimic MySQL (Vitess provides a single logical MySQL database as a `keyspace`). And what [`mysqldump`](https://dev.mysql.com/doc/refman/8.0/en/mysqldump.html) does is this:
```
$ mysqldump -u root mydb dtest 2>/dev/null | head -n 15 ; mysqldump -u root mydb dtest | tail -n 10
-- MySQL dump 10.13  Distrib 8.0.27, for macos11.6 (arm64)
--
-- Host: localhost    Database: mydb
-- ------------------------------------------------------
-- Server version	8.0.27

/*!40101 SET @OLD_CHARACTER_SET_CLIENT=@@CHARACTER_SET_CLIENT */;
/*!40101 SET @OLD_CHARACTER_SET_RESULTS=@@CHARACTER_SET_RESULTS */;
/*!40101 SET @OLD_COLLATION_CONNECTION=@@COLLATION_CONNECTION */;
/*!50503 SET NAMES utf8mb4 */;
/*!40103 SET @OLD_TIME_ZONE=@@TIME_ZONE */;
/*!40103 SET TIME_ZONE='+00:00' */;
/*!40014 SET @OLD_UNIQUE_CHECKS=@@UNIQUE_CHECKS, UNIQUE_CHECKS=0 */;
/*!40014 SET @OLD_FOREIGN_KEY_CHECKS=@@FOREIGN_KEY_CHECKS, FOREIGN_KEY_CHECKS=0 */;
/*!40101 SET @OLD_SQL_MODE=@@SQL_MODE, SQL_MODE='NO_AUTO_VALUE_ON_ZERO' */;

/*!40101 SET SQL_MODE=@OLD_SQL_MODE */;
/*!40014 SET FOREIGN_KEY_CHECKS=@OLD_FOREIGN_KEY_CHECKS */;
/*!40014 SET UNIQUE_CHECKS=@OLD_UNIQUE_CHECKS */;
/*!40101 SET CHARACTER_SET_CLIENT=@OLD_CHARACTER_SET_CLIENT */;
/*!40101 SET CHARACTER_SET_RESULTS=@OLD_CHARACTER_SET_RESULTS */;
/*!40101 SET COLLATION_CONNECTION=@OLD_COLLATION_CONNECTION */;
/*!40111 SET SQL_NOTES=@OLD_SQL_NOTES */;

-- Dump completed on 2022-01-13 13:29:28
```

So it saves the old `sql_mode`, sets it to `'NO_AUTO_VALUE_ON_ZERO'` in the session used to recreate the database objects (`CREATE`, `INSERT`), then at the end it sets the `sql_mode` back to what it was ([see source](https://github.com/mysql/mysql-server/blob/3290a66c89eb1625a7058e0ef732432b6952b435/client/mysqldump.cc#L795-L818)).

**_We can do the same thing in VReplication. That is what this PR sets out to do._**

This then not only deals with the zero date issue, but any other `sql_mode` related issue. This is a simple approach and, because we're doing the same thing MySQL does (remember that Vitess offers MySQL as a logical database spread across N nodes), we would not need a flag or anything user visible. 

## Related Issue(s)
https://github.com/vitessio/vitess/issues/7883
https://github.com/vitessio/vitess/pull/6561

## Checklist
- [X] Should this PR be backported? NO
- [X] Tests were added
- [X] Documentation not needed (it will now simply work as one would expect)


mattlord:
I wonder if we can instead use NULL if it's a nullable column or `1970-01-01 00:00:01` if it's not? I feel like these zero dates will just bite us later.

go/vt/mysqlctl/schema.go
```diff
@@ -439,6 +439,11 @@ func (mysqld *Mysqld) ApplySchemaChange(ctx context.Context, dbName string, chan
 	}
 
 	sql := change.SQL
+
+	if change.SQLMode != "" {
+		sql = fmt.Sprintf("SET @@session.sql_mode='%s';\n%s", change.SQLMode, sql)
+	}
+
```
shlomi-noach, line 446: :+1: pointing out to myself that this SQL is then written executed by `mysql` client, which terminates immediately after execution; hence, no need to worry about the session variables being changed.

go/vt/vttablet/tabletmanager/vreplication/vcopier_test.go
```diff
@@ -1361,3 +1362,81 @@ func TestPlayerCopyTablesWithGeneratedColumn(t *testing.T) {
...
+		fmt.Sprintf("create table %s.dst1(id int, dt date, primary key(id))", vrepldb),
+		"insert into src1 values(1, '2020-01-12'), (2, '0000-00-00');",
+	})
+
+	// default mysql flavor allows invalid dates: so disallow explicitly for this test
+	if err := env.Mysqld.ExecuteSuperQuery(context.Background(), "SET @@global.sql_mode=CONCAT(@@global.sql_mode, ',NO_ZERO_DATE,NO_ZERO_IN_DATE')"); err != nil {
+		fmt.Fprintf(os.Stderr, "%v", err)
+	}
+	defer func() {
+		if err := env.Mysqld.ExecuteSuperQuery(context.Background(), "SET @@global.sql_mode=REPLACE(@@global.sql_mode, ',NO_ZERO_DATE,NO_ZERO_IN_DATE','')"); err != nil {
```
shlomi-noach, line 1380: MySQL will re-arrange the `sql_mode` tokens as it sees fit. Earlier we appended `',NO_ZERO_DATE,NO_ZERO_IN_DATE'` to the end of `sql_mode`, but there's no guarantee that reading `sql_mode` the two will still be last, or in that order. 

The correct and safe way to remove the two directives is:

```sql
set @@session.sql_mode=REPLACE(REPLACE(@@session.sql_mode, 'NO_ZERO_DATE', ''), 'NO_ZERO_IN_DATE', '');
```

Note that there is no need to remove the comma separator.

go/vt/vttablet/tabletmanager/vreplication/vreplicator.go
```diff
@@ -72,6 +72,19 @@ var (
 	vreplicationStoreCompressedGTID = flag.Bool("vreplication_store_compressed_gtid", false, "Store compressed gtids in the pos column of _vt.vreplication")
 )
 
+const (
+	getSQLModeQuery = `SELECT @@session.sql_mode`
```
shlomi-noach, line 76: suggestion: use aliases: `SELECT @@session.sql_mode AS sql_mode`

go/vt/vttablet/tabletmanager/vreplication/vreplicator.go
```diff
@@ -406,6 +429,38 @@ func (vr *vreplicator) resetFKCheckAfterCopy() error {
...
 }
 
+func (vr *vreplicator) setSQLMode() (func(), error) {
+	// First save the original SQL mode if we have not already done so
+	if vr.originalSQLMode == "" {
+		res, err := vr.dbClient.Execute(getSQLModeQuery)
+		if err != nil || len(res.Rows) != 1 {
+			return nil, fmt.Errorf("could not get the original sql_mode on target: %v", err)
+		}
+		vr.originalSQLMode = res.Rows[0][0].ToString()
```
shlomi-noach, line 439: Opinion: we really should move away from `[0][0]`. It's archaic. Instead, you may:
```suggestion
		vr.originalSQLMode = res.Named().Row().AsString("sql_mode", "")
```

go/vt/vttablet/tabletmanager/vreplication/vreplicator.go
```diff
@@ -406,6 +429,38 @@ func (vr *vreplicator) resetFKCheckAfterCopy() error {
...
+	}
+
+	// Now set it to a permissive mode that will allow us to recreate
+	// any database object that exists on the source in full on the
+	// target
+	query := fmt.Sprintf(setSQLModeQueryf, SQLMode)
+	_, err := vr.dbClient.Execute(query)
+	if err != nil {
```
shlomi-noach, line 447: getting an error here doesn't mean the query isn't applied. Maybe the query is applied, then there's an error while reading the network.

Therefore, `resetFunc` should be calculated before this point, and returned next:

go/vt/vttablet/tabletmanager/vreplication/vreplicator.go
```diff
@@ -406,6 +429,38 @@ func (vr *vreplicator) resetFKCheckAfterCopy() error {
...
+		return nil, fmt.Errorf("could not set the permissive sql_mode on target using %s: %v", query, err)
```
shlomi-noach, line 448: (continued from above)
```suggestion
		return resetFunc, fmt.Errorf("could not set the permissive sql_mode on target using %s: %v", query, err)
```

go/vt/vttablet/tabletmanager/vreplication/vreplicator.go
```diff
@@ -406,6 +429,38 @@ func (vr *vreplicator) resetFKCheckAfterCopy() error {
...
+	// Lastly create a callback function for resetting the original
+	// SQL mode back at the end of the vreplication operation.
+	// You should defer this callback wherever you call setSQLMode()
+	resetFunc := func() {
+		query := fmt.Sprintf(setSQLModeQueryf, vr.originalSQLMode)
+		_, err := vr.dbClient.Execute(query)
+		if err != nil {
+			log.Warningf("could not reset sql_mode on target using %s: %v", query, err)
+		}
+	}
```
shlomi-noach, line 460: Evaluate this before setting sql mode

shlomi-noach (changes requested):
Overall looks good! A few changes requested (bugfixes) and a couple suggestions (coding style, change at your discretion), see inline.

go/vt/mysqlctl/schema.go
mattlord: I'll add a comment on that too. Thanks!

go/vt/vttablet/tabletmanager/vreplication/vreplicator.go
mattlord: Good point! Thanks!

mattlord:
> Overall looks good! A few changes requested (bugfixes) and a couple suggestions (coding style, change at your discretion), see inline.

THANK YOU for the great ideas/suggestions! :heart: I've implemented all of them now.

mattlord (approved):

shlomi-noach (approved):

mattlord:
(rebased on `vitess/main` to pick up upgrade/downgrade reparent flakiness fixes)

PR #9507, opened 2022-01-13
Fix flakiness in reparent tests

GuptaManan100:
## Description

It was noticed that the cluster 14 tests were a little flaky and on further investigation, it turned out that the reason for this flakiness was MySQL not shutting down properly and using the full 5 minute wait time given to it. On further investigation and looking at logs with @mattlord, it turned out that the reason that MySQL was hanging was because it was waiting for a Semi-sync ack for an internal Vitess command (guess) even though the socket file had disappeared and all the connections to the replicas had been closed. We are still unsure whether this is a MySQL bug or some configuration issue.
We discussed 3 possible solutions for fixing the CI - 
1. reduce the wait_time for the semi-sync acks by reducing the value for `rpl_semi_sync_master_timeout`. But this parameter is only available in 8.0 and the CI are running 5.7. Also this parameter would have to be changed after MySQL startup since the my.cnf file is setup by mysqlctl on startup.
2. Verify that the command that we are issuing from the tests on the primary are necessarily ACKed. This isn't required since the command would not have successfully returned until it was ACKed. It would have been blocked. So my guess is that the command blocked for semi-sync ACKs isn't the one issued from the tests but is an internal Vitess command that we run on primaries when it is first setup like setting up the local_metadata table, etc.
3. Reduce the time to wait for mysql to shutdown and kill the instance after it. To me 30 seconds seems like a reasonable amount of time to provided mysql to shutdown before killing it.

This PR goes ahead with the third alternate and kills the MySQL instance by using the pid file after 30 seconds if it hasn't shutdown already.

## Checklist

### DIFF
--- a/go/test/endtoend/cluster/mysqlctl_process.go
+++ b/go/test/endtoend/cluster/mysqlctl_process.go
@@ -23,7 +23,10 @@ import (
 	"os"
 	"os/exec"
 	"path"
+	"strconv"
 	"strings"
+	"syscall"
+	"time"
 
 	"vitess.io/vitess/go/mysql"
 	"vitess.io/vitess/go/vt/log"
@@ -125,7 +128,7 @@ ssl_key={{.Dir}}/server-001-key.pem
 	return tmpProcess, tmpProcess.Start()
 }
 
-// Stop executes mysqlctl command to stop mysql instance
+// Stop executes mysqlctl command to stop mysql instance and kills the mysql instance if it doesn't shutdown in 30 seconds.
 func (mysqlctl *MysqlctlProcess) Stop() (err error) {
 	log.Infof("Shutting down MySQL: %d", mysqlctl.TabletUID)
 	defer log.Infof("MySQL shutdown complete: %d", mysqlctl.TabletUID)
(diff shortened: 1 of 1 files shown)


harshit-gangal (approved):

mattlord (approved):

GuptaManan100:
Yes you are right. I went back to look at the docs and I misunderstood. Thankyou for letting me know 🙇‍♂️ !

deepthi (commented):
Which is the specific test that was failing/flaky?
I'm worried that if it is really the case that the new primary is waiting for a semi-sync ACK from an internal database write (like local_metadata), we might have an actual bug.
Do we just check for success after a PRS, or do we attempt to write to the primary to verify that it is really accepting writes?

GuptaManan100:
We check that the replication is working by writing to the primary and also checking that the write reaches the other tablets. As far as the test, the most severely affected tests are the ones which have only 1 acking replica. For example, in `TestERSPromoteRdonly` we change 2 of the 3 available replicas to rdonly instances, so there is only 1 replica which is sending ACKs

deepthi:
Discussed this with Manan. In at least one of the tests where the shutdown is hanging we first `confirmReplication` which means that writes from the primary are acknowledged correctly. It's unclear why there the primary still looks like it is hanging waiting for a semi-sync ACK.
While we haven't got to the bottom of this yet, it seems like it's not new flakiness. Let's go ahead and merge this PR and defer further investigation.

Issue #7113, opened 2020-12-04
Task: Change TabletType MASTER to PRIMARY


deepthi:
There has been discussion in the past on exactly which term to replace `master` with. Since vitess is a consensus-based system, `primary` seems like the most appropriate choice (even though `source` is popular for replication use cases).

TabletType is everywhere in the code base, so we will need to first determine the scope of this change and how to do it in a backwards compatible way.

Website docs should be updated along with code.

ajm188:
I _think_ that if we just rename the field in the enum, then anyone dependent on that should not notice, since it'll still be `int(1)` on the wire. Then, the code change becomes atomic; we do a global find/replace to `topodatapb.TabletType_PRIMARY`, and code compiled on the old version still works with the old field name, and code compiled on the new version works with the new field name.

derekperkins:
There are a few edge cases where the string representation is used, specifically when json marshaling, so we'd have to check for those types of usage.

derekperkins:
While I personally prefer `primary`, I'd vote to match MySQL and call it `source`
https://mysqlhighavailability.com/mysql-terminology-updates/
https://dev.mysql.com/doc/refman/8.0/en/replication.html

ajm188:
> There are a few edge cases where the string representation is used, specifically when json marshaling, so we'd have to check for those types of usage.

Ah, nice catch! I think we can use an alias in that case https://developers.google.com/protocol-buffers/docs/proto3#enum (I can't link precisely enough)

deepthi:
Fixed by #8511

PR #9516, opened 2022-01-15
naming: delete old code that was needed for version compatibility

deepthi:
## Description

Delete code that was kept around until we passed v12.0
A few other miscellaneous fixes

## Related Issue(s)

#7113

## Checklist

### DIFF
--- a/go/vt/vttablet/grpctabletconn/conn.go
+++ b/go/vt/vttablet/grpctabletconn/conn.go
@@ -221,11 +221,7 @@ func (conn *gRPCQueryClient) Begin(ctx context.Context, target *querypb.Target, 
 	if err != nil {
 		return 0, nil, tabletconn.ErrorFromGRPC(err)
 	}
-	// For backwards compatibility, we don't require tablet alias to be present in the response
-	// TODO(deepthi): After 7.0 change this
-	//	return br.TransactionId, br.TabletAlias, nil
-	// also assert that br.TabletAlias == conn.tablet.Alias
-	return br.TransactionId, conn.tablet.Alias, nil
+	return br.TransactionId, br.TabletAlias, nil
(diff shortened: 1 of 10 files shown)


go/vt/vttablet/grpctabletconn/conn.go
```diff
@@ -221,11 +221,7 @@ func (conn *gRPCQueryClient) Begin(ctx context.Context, target *querypb.Target,
 	if err != nil {
 		return 0, nil, tabletconn.ErrorFromGRPC(err)
 	}
-	// For backwards compatibility, we don't require tablet alias to be present in the response
-	// TODO(deepthi): After 7.0 change this
-	//	return br.TransactionId, br.TabletAlias, nil
-	// also assert that br.TabletAlias == conn.tablet.Alias
```
rohit-nayak-ps, line 227: Was the scope of this TODO to also assert `br.TabletAlias == conn.tablet.Alias`.

rohit-nayak-ps (approved):
lgtm

go/vt/vttablet/grpctabletconn/conn.go
deepthi: I thought that was a good idea at that time, but we don't in general put assertions into production code.
We _could_ put an if condition that returns an error but it seems like overkill.

deepthi: @systay @harshit-gangal any opinions on this?

harshit-gangal: I think this check is not necessary it is not giving any value here. as the conn here is the actual connection to the server it is unlikely it will differ.

harshit-gangal (approved):

PR #9522, opened 2022-01-17
Routing Rules: Improved escaping of reserved and invalid names for tables and keyspaces 

rohit-nayak-ps:
## Description
TBD

## Checklist


go/vt/mysqlctl/reparent.go
```diff
@@ -57,7 +57,7 @@ ENGINE=InnoDB`, mysql.MaximumPositionSize)}
 // This is to support in-place upgrades from 13.0.x to 14.0.x
 func AlterReparentJournal() []string {
 	return []string{
-		"`ALTER TABLE _vt.reparent_journal CHANGE COLUMN primary_alias master_alias VARBINARY(32) NOT NULL`",
+		"ALTER TABLE _vt.reparent_journal CHANGE COLUMN primary_alias master_alias VARBINARY(32) NOT NULL",
```
rohit-nayak-ps, line 60: @deepthi, FYI: this sql query had extra backticks which was causing failures I noticed in the logs.

go/sqlescape/ids.go
```diff
@@ -44,3 +44,12 @@ func EscapeIDs(identifiers []string) []string {
 	}
 	return result
 }
+
+// UnescapeID reverses any backticking in the input string.
+func UnescapeID(in string) string {
```
mattlord, line 49: Do we care about leading or trailing whitespace for idents? I'm not sure that's possible, but came to mind.

go/test/endtoend/vreplication/cluster.go
```diff
@@ -117,7 +117,8 @@ func getClusterConfig(idx int, dataRootDir string) *ClusterConfig {
 	}
 
 	return &ClusterConfig{
-		hostname:            "localhost",
+		// the hostname is also used to invoke the mysql client using sockets, if localhost is used mysql uses tcp
```
mattlord, line 120: FWIW, localhost uses the socket and 127.0.0.1 or ::1 uses tcp.

go/vt/mysqlctl/tmutils/schema.go
```diff
@@ -229,7 +231,11 @@ func SchemaDefinitionToSQLStrings(sd *tabletmanagerdatapb.SchemaDefinition) []st
 	sqlStrings := make([]string, 0, len(sd.TableDefinitions)+1)
 	createViewSQL := make([]string, 0, len(sd.TableDefinitions))
 
-	sqlStrings = append(sqlStrings, sd.DatabaseSchema)
+	// backtick database name since keyspace names appear in the routing rules, and they might need to be escaped
+	// unescape() first in case an already escaped string was passed
+	createDatabaseSQL := strings.Replace(sd.DatabaseSchema, "`{{.DatabaseName}}`", "{{.DatabaseName}}", -1)
```
mattlord, line 236: We should use `sqlescape.UnescapeID` here so that we have one spot to alter the unescaping if needed.

mattlord:
Nice work, @rohit-nayak-ps ! I think you addressed the two unexpected issues I noticed before when I tried to fix this quickly.

go/sqlescape/ids.go
rohit-nayak-ps: I would think the parser would provide identifiers without whitespaces, but don't know for sure.

mattlord: Yeah, I would think so too... we can leave it as-is.

go/vt/mysqlctl/tmutils/schema.go
mattlord: Ah, right. Thanks!

mattlord (approved):
I had previously created a draft PR where I was adding the backticks directly to the routing rules and that got messy. I like how this was addressed here — simple and clean. Thanks, @rohit-nayak-ps !

deepthi (approved):
LGTM

Issue #9153, opened 2021-11-06
Tablet manager detects and publishes `Identified mysql port: 0` during restore 


streadway:
#### Overview of the Issue

Tablets that restore from backup identify and publish a `mysql_port: 0` before transitioning to REPLICA. After transitioning to REPLICA, the port remains published as 0.

This causes vtorc to discover instances with a 0 port, and detects a failure `ClusterHasNoPrimary: 10.0.0.1:0`.

Restarting the vttablet/mysqld pair publishes the correct port.

#### Reproduction Steps

1. Use planetscale/vitess-operator to create a keyspace with a backupLocationName and replicas: 1
1. Wait for vtbackup init pods to complete
1. Wait for the vttablet pods to restore from the initial backup
1. Observe the tablet is PRIMARY
1. Update replicas to 10 (something large enough to find a potential race)
1. Wait for tablets to transition from RESTORE to REPLICA
1. Observe at least one published 0 port with `vtctlclient ListAllTablets | grep ":0 "`

#### Binary version

* vitess/lite:v12.0.0
* planetscale/vitess-operator:v2.5.1
* kubernetes: v1.19

#### Operating system and Environment details

* Cross-region etcd2 global topology server with 100-500ms write times
* Datacenter local etcd2 cell topology server with 1-10ms write times
* Reproduced with non-virtualized hardware, local SSD data volumes
* Reproduced with AWS EBS data volumes

Relevant vttablet flags:

```text
--backup_engine_implementation=xtrabackup
--backup_storage_implementation=gcs
--binlog_use_v3_resharding_mode=true
--disable_active_reparents=true
--enable_replication_reporter=true
--enable_semi_sync=false
--health_check_interval=5s
--mycnf_socket_file=/vt/socket/mysql.sock
--mysqlctl_socket=/vt/socket/mysqlctl.sock
--restore_concurrency=10
--restore_from_backup=true
--wait_for_backup_interval=10s
```

#### Log Fragments

```text
I1104 22:13:40.295180       1 xtrabackupengine.go:655] move-back stderr: 211104 22:13:40 completed OK!
I1104 22:13:40.299880       1 xtrabackupengine.go:401] Restore: returning replication position f1ccf5a2-3d93-11ec-8a89-aa86234d344d:1-3,ff3a5091-3d93-11ec-89f8-fa2c3fe42f95:1-28
I1104 22:13:40.299900       1 backup.go:330] Restore: starting mysqld for mysql_upgrade
[...]

PR #9523, opened 2022-01-18
evalengine: some last fixes

vmg:
## Description

With this last set of fixes, I can run a random expression fuzzer for 1min+ locally and not find any mismatches between the evaluation results of MySQL and ours. #9153 

I am 100% sure there are still corner cases that we're not handling properly, but we should be wildly more compatible than we were before I started this refactoring project, and I believe the `evalengine` is stable enough now to ship in the next version of Vitess and being enabled by default.

cc @deepthi @systay 

## Checklist


systay (approved):
LGTM

PR #9524, opened 2022-01-18
ci: Fix builds by importing MySQL keys

vmg:
## Description

CI is currently broken because Oracle has rotated the PGP keys for their APT repository for MySQL 8.0+.

This PR imports the new keys. In order to make it work, we need to push new bootstrap docker images to DockerHub, because those are also failing when used as dependencies because of their outdated keys.

cc @askdba @deepthi 

## Checklist


systay (approved):
Thanks for working on this :heart:

PR #9528, opened 2022-01-19
Add support for health service in gRPC server

### LINKED ISSUE #9333: Support gRPC health check service in VTGate
#### Feature Description

A written overview of the feature.

Currently the VTGate gRPC server doesn't register the [health check](https://github.com/grpc/grpc/blob/master/doc/health-checking.md) service. Here's the output listing services of a VTGate server using grpcurl:
```
$ ./grpcurl -plaintext  localhost:15991 list
grpc.reflection.v1alpha.ServerReflection
vtgateservice.Vitess
```

The health check service can be registered like this: https://github.com/vitessio/vitess/blob/main/go/vt/vtadmin/grpcserver/server.go#L138
```
healthServer := health.NewServer()
healthpb.RegisterHealthServer(gserv, healthServer)
```

If the gRPC server has health check service registers the `grpcurl ls` output is like this (using [temporal](https://github.com/temporalio/temporal) as an example)
```
grpc.health.v1.Health
grpc.reflection.v1alpha.ServerReflection
temporal.api.workflowservice.v1.WorkflowService
temporal.server.api.adminservice.v1.AdminService
```

https://github.com/vitessio/vitess/blob/main/go/vt/vtadmin/grpcserver/server.go#L138

#### Use Case(s)

In our ecosystem we use `grpc.health.v1.Health` endpoint to check the health of a gRPC server. It optionally takes a service name, such as `vtgateservice.Vitess` to query for health status.

shichao-an:
## Description

* Register gRPC health service to support health checks. See [GRPC Health Checking Protocol](https://github.com/grpc/grpc/blob/master/doc/health-checking.md) and [grpc_health_v1](https://pkg.go.dev/google.golang.org/grpc/health/grpc_health_v1).
* This change affects all use cases of `servenv.GRPCServer`

## Related Issue(s)

Fixes #9333

## Checklist

### DIFF
--- a/go/vt/servenv/grpc_server.go
+++ b/go/vt/servenv/grpc_server.go
@@ -26,11 +26,13 @@ import (
 
 	grpc_middleware "github.com/grpc-ecosystem/go-grpc-middleware"
 	grpc_prometheus "github.com/grpc-ecosystem/go-grpc-prometheus"
+	healthpb "google.golang.org/grpc/health/grpc_health_v1"
 
 	"vitess.io/vitess/go/trace"
 
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/health"
 	"google.golang.org/grpc/keepalive"
 	"google.golang.org/grpc/reflection"
 
@@ -227,6 +229,10 @@ func serveGRPC() {
 
 	// register reflection to support list calls :)
 	reflection.Register(GRPCServer)
+
+	// register health service to support health checks
+	healthServer := health.NewServer()
+	healthpb.RegisterHealthServer(GRPCServer, healthServer)
 
 	// listen on the port
 	log.Infof("Listening for gRPC calls on port %v", *GRPCPort)
--- a/go/vt/servenv/grpc_server.go
+++ b/go/vt/servenv/grpc_server.go
@@ -234,6 +234,10 @@ func serveGRPC() {
 	healthServer := health.NewServer()
 	healthpb.RegisterHealthServer(GRPCServer, healthServer)
 
+	for service := range GRPCServer.GetServiceInfo() {
+		healthServer.SetServingStatus(service, healthpb.HealthCheckResponse_SERVING)
+	}
+
 	// listen on the port
 	log.Infof("Listening for gRPC calls on port %v", *GRPCPort)
 	listener, err := net.Listen("tcp", fmt.Sprintf(":%d", *GRPCPort))


vmg:
This doesn't break anything, but I'm not exactly convinced that the generic GRPC health service API makes sense for Vitess, as our health status is already exposed via other APIs and it is more complex than what this generic API provides (most notably, Vitess distinguishes between serving/not serving -- which applies to queries but not the GRPC interface itself, and cannot be modelled with this API).

Is there a particular use case you have in mind for wiring up this API? If you're trying to health check a specific service in a Vitess cluster, I would recommend you use the GRPC APIs which we already provide and that relay much more relevant information about the actual health status of the Vitess cluster and its services. Cheers.

shichao-an:
> This doesn't break anything, but I'm not exactly convinced that the generic GRPC health service API makes sense for Vitess, as our health status is already exposed via other APIs and it is more complex than what this generic API provides (most notably, Vitess distinguishes between serving/not serving -- which applies to queries but not the GRPC interface itself, and cannot be modelled with this API).
> 
> Is there a particular use case you have in mind for wiring up this API? If you're trying to health check a specific service in a Vitess cluster, I would recommend you use the GRPC APIs which we already provide and that relay much more relevant information about the actual health status of the Vitess cluster and its services. Cheers.

@vmg Our Service Networking infrastructure uses Envoy health check ([link](https://github.com/envoyproxy/envoy/blob/main/api/envoy/config/core/v3/health_check.proto#L172)), which requires "grpc.health.v1.health" service endpoint and an optional service name (for example, "vtgateservice.Vitess") to send to this endpoint. Almost all our gRPC services have "grpc.health.v1.health" and that's why SN team made that assumption without supporting custom health endpoint for gRPC (they do support custom HTTP endpoint, but only for HTTP services). Without "grpc.health.v1.health", the Envoy HC simply fails.

On the higher level, it is for Envoy to determine load balancing semantics such as [panic mode](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/panic_threshold), so it doesn't really matter whether it is serving/not_serving from Vitess's perspective.

vmg:
Sounds good to me. Have you tested that this change makes your Envoy work in your infrastructure? From looking at the Godoc for the HealthCheck API, it seems like you may need to call [`SetServingStatus`](https://pkg.go.dev/google.golang.org/grpc/health#Server.SetServingStatus) at least once to register a healthy service by name?

shichao-an:
> Sounds good to me. Have you tested that this change makes your Envoy work in your infrastructure? From looking at the Godoc for the HealthCheck API, it seems like you may need to call [`SetServingStatus`](https://pkg.go.dev/google.golang.org/grpc/health#Server.SetServingStatus) at least once to register a healthy service by name?

@vmg Need your input on this as well, since by default our gRPC applications use a custom health check that implements the standard Health service interface (something like this, `health.RegisterHealthServer(grpcServer, customServer)`, where `customServer` implementing Check() and Watch() that returns serving/not_serving for HealthCheckRequest).

So if I only registered the default Server (as in this PR) like this `healthpb.RegisterHealthServer(gserv, health.NewServer())` without using `SetServingStatus` or use a custom health check (like what we do internally), it will work for HealthCheckRequest whose input service is not set, and return SERVING (see this [link](https://github.com/grpc/grpc-go/blob/master/health/server.go#L48).

I think for Vitess gRPC server (which is used by vtgate, vttablet, etc.), if we want to explicitly set serving services, we can simply use `SetServingStatus`. I wonder what services should I set as serving? Or should I simply set "grpc.health.v1.Health" as serving just as in https://github.com/vitessio/vitess/blob/f88d121a4b54fdbbbf5e24c47176e538ae031b9d/go/vt/vtadmin/grpcserver/server.go#L228

vmg:
The code seen in that commit looks good to me. Iterating the services and calling `SetServingStatus` will work fine. Can you push that commit to this PR?

shichao-an:
> The code seen in that commit looks good to me. Iterating the services and calling `SetServingStatus` will work fine. Can you push that commit to this PR?

Done. Here are the local tests I've done.

Make and start local docker:
```
make docker_local
make docker_run_local
```
Test gRPC calls using grpcurl
```
st-shichao1:~ shichao$ grpcurl  -plaintext localhost:15991 list
grpc.health.v1.Health
grpc.reflection.v1alpha.ServerReflection
vtgateservice.Vitess
st-shichao1:~ shichao$ grpcurl -plaintext localhost:15991 grpc.health.v1.Health.Check
{
  "status": "SERVING"
}
st-shichao1:~ shichao$ grpcurl -d '{ "service": "" }' -plaintext localhost:15991 grpc.health.v1.Health.Check
{
  "status": "SERVING"
}
st-shichao1:~ shichao$ grpcurl -d '{ "service": "grpc.health.v1.Health" }' -plaintext localhost:15991 grpc.health.v1.Health.Check
{
  "status": "SERVING"
}
st-shichao1:~ shichao$ grpcurl -d '{ "service": "vtgateservice.Vitess" }' -plaintext localhost:15991 grpc.health.v1.Health.Check
{
  "status": "SERVING"
}
st-shichao1:~ shichao$ grpcurl -d '{ "service": "grpc.reflection.v1alpha.ServerReflection" }' -plaintext localhost:15991 grpc.health.v1.Health.Check
{
  "status": "SERVING"
}
st-shichao1:~ shichao$ grpcurl -d '{ "service": "foobar" }' -plaintext localhost:15991 grpc.health.v1.Health.Check
ERROR:
  Code: NotFound
  Message: unknown service

```

vmg (approved):

shichao-an:
@vmg Thank you!

PR #9531, opened 2022-01-19
Filter after left join

systay:
## Description
Adds planner support for handling filtering after doing an outer join, such as in the following query:

```sql
SELECT user.id 
FROM user 
  LEFT JOIN user_extra ON user.col = user_extra.col
WHERE user_extra.id IS NULL
```

## Checklist


go/vt/vtgate/evalengine/is_expr.go
```diff
@@ -0,0 +1,113 @@
...
+)
+
+// IsExpr represents the IS expression in MySQL.
+// boolean_primary IS [NOT] {TRUE | FALSE | NULL}
+type IsExpr struct {
+	Expr Expr
+
+	Negate bool
+
+	// only one of the following should be set to true
```
vmg, line 31: Then we should make this as an `enum`, which always holds the property of only one being true :)

go/vt/vtgate/evalengine/is_expr.go
```diff
@@ -0,0 +1,113 @@
...
+
+import (
+	"vitess.io/vitess/go/mysql/collations"
+	querypb "vitess.io/vitess/go/vt/proto/query"
+)
+
+// IsExpr represents the IS expression in MySQL.
+// boolean_primary IS [NOT] {TRUE | FALSE | NULL}
+type IsExpr struct {
+	Expr Expr
```
vmg, line 27: If you embed `Unary` in this struct instead of `Expr`, you'll get a lot of the methods you've implemented for free.

go/vt/vtgate/planbuilder/operator_transformers.go
```diff
@@ -55,6 +55,29 @@ func transformToLogicalPlan(ctx *plancontext.PlanningContext, op abstract.Physic
...
+			return nil, err
+		}
+		scl := &simpleConverterLookup{
+			canPushProjection: true,
+			ctx:               ctx,
+			plan:              plan,
+		}
+		ast := sqlparser.AndExpressions(op.Predicates...)
+		predicate, err := evalengine.Convert(ast, scl)
+		if err != nil {
```
GuptaManan100, line 70: Should we throw a different error here after we see an error here. One similar to we used to have earlier - "unsupported: cross-shard left join and where clause"

GuptaManan100 (approved):

vmg (changes requested):
Some suggestions for the `IS` implementation in the `evalengine`.

go/vt/vtgate/planbuilder/operator_transformers.go
systay: Yeah, I was thinking the same

PR #9532, opened 2022-01-19
BinlogEvent: provide Bytes() function

shlomi-noach:
The `BinlogEvent` interface supports many introspection functions over the event, but has no way of getting the actual raw event data. The only user for `BinlogEvent` right now is VReplication, which doesn't need any raw data.

This PR adds the `Bytes()` function, which returns the very raw binary data composing the event. The use case for this function is, for now, some research into potential ideas. If those mature then we can discuss further, or refactor, or revise what kind of information a `BinlogEvent` should expose.

## Checklist


shlomi-noach:
converted to draft. I think I'm gonna need a little bit more. I may just as well open a branch with the full experiment.

shlomi-noach:
This is no longer required.

PR #9533, opened 2022-01-19
Pass Semi sync information from durability policy and use it to log discrepancies

GuptaManan100:
## Description

In release 13, we will introduce a new flag `durability` in vtctld and vtctl. That durability policy should be the one that reflects the semi-sync settings and we will eventually deprecate `enable_semi_sync` flag on the vttablets. But all of this cannot be accomplished in a single release because of upgrade downgrade considerations. So the plan is as follows - 
- Introduce the new flag `durability`
- Pass the information all through the components and just document the discrepancies in  configuration.
- Keep using `enable_semi_sync` flag in v13, so the new flag is only used for logging but it is expected that vtctl, vtctld binaries of v13 will have the flag set.
- In release 14, we would have the semi-sync information being passed everywhere, so we can deprecate `enable_semi_sync` and use the information coming from the durability policies.

This PR accomplishes this change of passing the information down and logging the differences for release 13. 

## Related Issue(s)

- #8975 

## Checklist
- [ ] Documentation was added or is not required - Pending


GuptaManan100:
Apart from the checks in this PR, I also made a change wherein we start using the durability policy information in `fixSemiSync` in #9552. That PR is 2 commits ahead of this PR - 
1. https://github.com/vitessio/vitess/pull/9552/commits/1380e76004e394a264f637b647ac8c6d45228ba4 wherein we change fixSemiSyncReplication
2. https://github.com/vitessio/vitess/pull/9552/commits/39e6f09c973d943cfcb97d0a9f2d5db3a71f34e5 which fixes some of the unit tests to set the correct durability policies and skips the backup tests.

On that PR, it can be observed that the only failing tests are from `cluster 11, 16, 19, 20, 21, 26, docker_test_cluster_10` which all run `Backup` RPC and `Upgrade Downgrade test for reparenting with old vtctl`. Their are 2 failing tests, one `TestSemiSyncSetupCorrectly` and `TestChangeTypeSemiSync` which both test semi-sync settings. We expect these tests to fail since old vtclt does not have durability policy information and the new vttablets in that branch only fixSemiSync using the parameter passed down. The failure in tests concerned with `Backup` RPC is also expected, since we will no longer be fixing semiSync after a `Backup` operation succeeds. We will rely on vtorc to fix it. The main reason for this -
  1. After the durability policy changes take effect, a tablet might be sending semi sync acks to one primary but may be configured to not send semi sync acks to some other primary, and since `Backup` is a long operation during which the primary might have changed, we will have no idea whether we should be setting our semi-sync or not. So we will rely on vtorc to fix it when it becomes a compulsory component of Vitess. An alternate is to run StartReplication RPC after running Backup RPC, and that will figure out the semi sync settings.

go/vt/orchestrator/inst/tablet_dao.go
```diff
@@ -80,15 +80,15 @@ func SwitchPrimary(newPrimaryKey, oldPrimaryKey InstanceKey) error {
 		log.Errore(err)
 		return nil
 	}
-	if _, err := ChangeTabletType(oldPrimaryKey, topodatapb.TabletType_REPLICA); err != nil {
+	if _, err := ChangeTabletType(oldPrimaryKey, topodatapb.TabletType_REPLICA, ReplicaSemiSync(newPrimaryKey, oldPrimaryKey)); err != nil {
```
shlomi-noach, line 83: The function name `ReplicaSemiSync` is confusing. I'd be more comfortable if it were called `IsReplicaSemiSync`

go/vt/vttablet/tabletmanager/rpc_actions.go
```diff
@@ -43,6 +43,17 @@ const (
...
 )
 
+// SemiSyncAction is used to tell fixSemiSync whether to change the semi-sync
+// settings or not.
+type SemiSyncAction int
+
+// Allowed values for SemiSyncAction
+const (
+	SemiSyncActionNone = SemiSyncAction(iota)
+	SemiSyncActionTrue
```
shlomi-noach, line 53: Perhaps name as `SemiSyncActionSet`?

go/vt/vttablet/tabletmanager/rpc_actions.go
```diff
@@ -43,6 +43,17 @@ const (
...
+	SemiSyncActionFalse
```
shlomi-noach, line 54: Perhaps name as `SemiSyncActionUnset`?

shlomi-noach (commented):
I don't have much quality review content for now, just a few naming suggestions.

go/vt/vttablet/tabletmanager/rpc_replication.go
```diff
@@ -475,18 +475,20 @@ func (tm *TabletManager) demotePrimary(ctx context.Context, revertPartialFailure
...
-	if err := tm.fixSemiSync(topodatapb.TabletType_REPLICA); err != nil {
-		return nil, err
-	}
-	defer func() {
-		if finalErr != nil && revertPartialFailure && wasPrimary {
-			// enable primary-side semi-sync again
-			if err := tm.fixSemiSync(topodatapb.TabletType_PRIMARY); err != nil {
-				log.Warningf("fixSemiSync(PRIMARY) failed during revert: %v", err)
-			}
+	if tm.isPrimarySideSemiSyncEnabled() {
```
deepthi, line 478: Can you add a comment here as to why we should check this first?

go/vt/worker/multi_split_diff.go
```diff
@@ -407,7 +407,7 @@ func (msdw *MultiSplitDiffWorker) stopReplicationOnSourceTabletAt(ctx context.Co
 		msdw.wr.Logger().Infof("stopping Replication %v at a minimum of %v", msdw.sourceAlias, vreplicationPos)
 
 		shortCtx, cancel = context.WithTimeout(ctx, *remoteActionsTimeout)
-		msdw.wr.TabletManagerClient().StartReplication(shortCtx, sourceTablet.Tablet)
+		msdw.wr.StartReplication(shortCtx, sourceTablet.Tablet)
```
deepthi, line 410: This is brilliant. MultiSplitDiffWorker is created using a vtctld command which means we have the durability policy and wrangler can enforce it. 💯

web/vtadmin/src/proto/vtadmin.d.ts
```diff
@@ -43087,6 +43135,12 @@ export namespace binlogdata {
 
         /** VEvent last_p_k_event */
         last_p_k_event?: (binlogdata.ILastPKEvent|null);
+
+        /** VEvent keyspace */
+        keyspace?: (string|null);
+
+        /** VEvent shard */
+        shard?: (string|null);
```
deepthi, line 43143: @doeg @ajm188 these fields were added to the proto definition some time ago, but the vtadmin files don't seem to have been regenerated at that time.

web/vtadmin/src/proto/vtadmin.js
```diff
@@ -30842,6 +30866,14 @@ $root.tabletmanagerdata = (function() {
...
 
+        /**
+         * ApplySchemaRequest sql_mode.
+         * @member {string} sql_mode
+         * @memberof tabletmanagerdata.ApplySchemaRequest
+         * @instance
+         */
+        ApplySchemaRequest.prototype.sql_mode = "";
+
         /**
```
deepthi, line 30877: This seems to be an artifact of a different PR as well.

web/vtadmin/src/proto/vtadmin.d.ts
deepthi: Can one of you post reminders in #maintainers and #developers so that people don't miss the step of generating vtadmin files?

web/vtadmin/src/proto/vtadmin.js
```diff
@@ -35887,6 +35932,7 @@ $root.tabletmanagerdata = (function() {
          * Properties of a StartReplicationRequest.
          * @memberof tabletmanagerdata
          * @interface IStartReplicationRequest
+         * @property {boolean|null} [semiSync] StartReplicationRequest semiSync
```
deepthi, line 35935: @notfelineit do these changes affect the "destructive" tablet actions UI?

deepthi (approved):
Nice work.
Please address all feedback and then this can be merged.

go/vt/orchestrator/inst/tablet_dao.go
GuptaManan100: Changed

go/vt/vttablet/tabletmanager/rpc_actions.go
GuptaManan100: Renamed

go/vt/vttablet/tabletmanager/rpc_actions.go
GuptaManan100: Renamed

go/cmd/vtworker/vtworker.go
```diff
@@ -37,13 +37,15 @@ import (
...
 	"vitess.io/vitess/go/vt/topo"
+	"vitess.io/vitess/go/vt/vtctl/reparentutil"
 	"vitess.io/vitess/go/vt/worker"
 )
 
 var (
 	cell                   = flag.String("cell", "", "cell to pick servers from")
 	commandDisplayInterval = flag.Duration("command_display_interval", time.Second, "Interval between each status update when vtworker is executing a single command from the command line")
 	username               = flag.String("username", "", "If set, value is set as immediate caller id in the request and used by vttablet for TableACL check")
+	durability             = flag.String("durability", "none", "type of durability to enforce. Default is none. Other values are dictated by registered plugins")
```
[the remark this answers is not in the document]
GuptaManan100: Fixed, it is good we did this now, otherwise we would have had to go through 2 more releases to fix it later 😆

go/vt/worker/multi_split_diff.go
GuptaManan100: It is actually created from vtworker, but we have added durability policy to that too! So everything works now!

go/vt/vttablet/tabletmanager/rpc_replication.go
GuptaManan100: Added

web/vtadmin/src/proto/vtadmin.d.ts
doeg: I've posted another reminder to both channels. (This was also mentioned back at the beginning of December.) I expect this will (understandably) be an issue until we can enforce this in CI. The ticket for that is https://github.com/vitessio/vitess/issues/9571.

web/vtadmin/src/proto/vtadmin.js
notfelineit: I don't think so, not as it exists - should we add a semiSync/durability option in the UI?

notfelineit: I don't think so, not as it exists - should we add a semiSync/durability option in the UI?

notfelineit: Screen Shot 2022-01-27 at 9 45 23 AM
Yes! It follows vtctld2's ui in that there is a singular button

deepthi: I suspect that this directly calls the TabletManager grpc API instead of going through vtctld API.
My suggestion is to add a new vtctld API. Then semi_sync won't need to be provided by the user, it can be computed by VtctldServer.

notfelineit: I'll check today and see which it is using - and switch to vtctld if necesary.

notfelineit: @deepthi vtadmin uses this codepath: https://cs.github.com/vitessio/vitess/blob/f906693c6441f4b41d8f842ffb9dba85258b5158/go/vt/vtctl/grpcvtctldserver/server.go?q=StartReplication#L2439 it does call the tablet manager API, but I think this is the vtctld API mentioned?

PR #9537, opened 2022-01-19
RFC: Add ApplySchema to vtctlservice

Phanatic:
## Description
This PR introduces a new grpc API in the `vtctlservice` service that will provide the same functionality as the [`ApplySchema`](https://vitess.io/docs/12.0/reference/programs/vtctl/schema-version-permissions/#applyschema) CLI invocation in `vtctlclient`, with a marked difference of being able to pass in an `callerID` parameter that allows databases with strict ACLs enabled to work with calls to `ApplySchema` with `gh-ost` as the DDL strategy.

As part of this PR, I've updated the existing `ApplySchema` CLI command to call this new API instead of calling the `schemamanager` directly

## Checklist
- [x] Should this PR be backported? : NO
- [x]  Tests were added.


go/vt/schemamanager/tablet_executor.go
```diff
@@ -49,9 +52,11 @@ type TabletExecutor struct {
 }
 
 // NewTabletExecutor creates a new TabletExecutor instance
-func NewTabletExecutor(requestContext string, wr *wrangler.Wrangler, waitReplicasTimeout time.Duration) *TabletExecutor {
+func NewTabletExecutor(requestContext string, ts *topo.Server, tmc tmclient.TabletManagerClient, logger logutil.Logger, waitReplicasTimeout time.Duration) *TabletExecutor {
```
Phanatic, line 55: I removed the reference to `wrangler` here and replaced it with `topo.Server` and `tabletManagerClient` to break an import cycle that will be introduced in this PR. 

``` 
    grpcvtctldserver  --> schema manager --> tablet_executor --> wrangler --> grpcvtctldserver
```

mattlord (commented):
Looks good! I had a few minor comments/nits. I can help with some manual user testing as well whenever you feel it may be helpful. Just let me know.

Phanatic:
> What does this change mean for current users of ApplySchema with ddl_strategy=direct? Will it continue to work as it did (even with ACLs configured on the tablets)?

@deepthi , existing callers of `ApplySchema` with a `ddl_strategy=direct` will be unaffected by this change. 

`ddl_strategy=direct` 

https://github.com/vitessio/vitess/blob/610fc78a3c1e09d72c6df4eea68366778e420218/go/vt/schemamanager/tablet_executor.go#L270

passes in `false` for `viaQueryService`

https://github.com/vitessio/vitess/blob/610fc78a3c1e09d72c6df4eea68366778e420218/go/vt/schemamanager/tablet_executor.go#L307


which calls `ExecuteFetchAsDBA`

https://github.com/vitessio/vitess/blob/610fc78a3c1e09d72c6df4eea68366778e420218/go/vt/schemamanager/tablet_executor.go#L454-L461

Which skips all ACL checks on the database.

go/vt/schemamanager/schemamanager_test.go
```diff
@@ -136,12 +131,10 @@ func TestSchemaManagerRun(t *testing.T) {
...
 
 	fakeTmc.AddSchemaDefinition("vt_test_keyspace", &tabletmanagerdatapb.SchemaDefinition{})
-
-	wr := wrangler.New(logutil.NewConsoleLogger(), newFakeTopo(t), fakeTmc)
-	executor := NewTabletExecutor("TestSchemaManagerRun", wr, testWaitReplicasTimeout)
+	executor := NewTabletExecutor("TestSchemaManagerRun", newFakeTopo(t), fakeTmc, logutil.NewConsoleLogger(), testWaitReplicasTimeout)
 
 	ctx := context.Background()
-	err := Run(ctx, controller, executor)
+	_, err := Run(ctx, controller, executor)
```
deepthi, line 137: We should validate the result from this call.

proto/vtctldata.proto
```diff
@@ -194,6 +195,34 @@ message ApplyRoutingRulesRequest {
 message ApplyRoutingRulesResponse {
 }
 
+message ApplySchemaRequest {
```
deepthi, line 198: protobuf changes need an extra build step.
From slack:

> For folks that update the .proto files with `make proto`, it would be very appreciated if, for now, you could also run the `make vtadmin_web_proto_types` command. This ensures that VTAdmin's generated TypeScript files stay in sync with the proto files.
I know that remembering things is the worst :) so this is only necessary until we can do this automagically in CI. (https://github.com/vitessio/vitess/issues/9571 is the ticket for that work.)
Note that running `make vtadmin_web_proto_types` requires an install of node LTS (>= 16.13). If your node version is too old, you may run into an error like this one. Node v16.13.x (LTS) can be installed via https://nodejs.org, or alternatively can be managed with something like nvm.

deepthi (commented):
Most of this looks good. A couple of things I noted inline + the commented out test needs to be fixed.

go/vt/vttablet/grpctmserver/server.go
```diff
@@ -200,6 +203,11 @@ func (s *server) UnlockTables(ctx context.Context, req *tabletmanagerdatapb.Unlo
 func (s *server) ExecuteQuery(ctx context.Context, request *tabletmanagerdatapb.ExecuteQueryRequest) (response *tabletmanagerdatapb.ExecuteQueryResponse, err error) {
 	defer s.tm.HandleRPCPanic(ctx, "ExecuteQuery", request, response, false /*verbose*/, &err)
 	ctx = callinfo.GRPCCallInfo(ctx)
+
+	// Attach the callerID as the EffectiveCallerID.
+	if request.CallerId != nil {
+		ctx = callerid.NewContext(ctx, request.CallerId, &querypb.VTGateCallerID{Username: request.CallerId.Principal})
+	}
```
deepthi, line 210: Note for future readers. We have not yet figured out whether ALL methods in this interface should process calledID in this way. In the meantime this PR fixes the known issue with `ApplySchema`.

proto/vtctldata.proto
Phanatic: Done!

doeg: Thank you both, it is appreciated. :)

shlomi-noach (commented):
I appreciate the endtoend tests to verify this behavior! I see `go/test/endtoend/onlineddl/ghost/onlineddl_ghost_test.go` was used. I'd like to suggest one of the other tests instead. Let's use `go/test/endtoend/onlineddl/vrepl/onlineddl_vrepl_test.go`.

We don't have quite the "general purpose" online DDL endtoend suite. I should really create one. But, moving forward we want to rely on vreplication based migrations; and who knows, maybe some day we will deprecate `gh-ost` migrations? In which case, we don't want to lose the ACL tests.

go/vt/schemamanager/tablet_executor_test.go
```diff
@@ -54,30 +50,32 @@ func TestTabletExecutorOpen(t *testing.T) {
...
+//	// Since this is a replica tablet, the Shard will have no primary.
+//	if err := wr.InitTablet(ctx, tablet, false /*allowPrimaryOverride*/, true /*createShardAndKeyspace*/, false /*allowUpdate*/); err != nil {
+//		t.Fatalf("InitTablet failed: %v", err)
+//	}
+//	executor := NewTabletExecutor("TestTabletExecutorOpenWithEmptyPrimaryAlias", ts, newFakeTabletManagerClient(), logutil.NewConsoleLogger(), testWaitReplicasTimeout)
+//	if err := executor.Open(ctx, "test_keyspace"); err == nil || !strings.Contains(err.Error(), "does not have a primary") {
+//		t.Fatalf("executor.Open() = '%v', want error", err)
+//	}
+//	executor.Close()
+//}
```
shlomi-noach, line 78: Any reason this commented code should not be removed?

go/vt/vtctl/grpcvtctldserver/server.go
```diff
@@ -154,6 +160,80 @@ func (s *VtctldServer) ApplyRoutingRules(ctx context.Context, req *vtctldatapb.A
...
+	if req.SkipPreflight {
+		executor.SkipPreflight()
+	}
+
+	if err := executor.SetDDLStrategy(req.DdlStrategy); err != nil {
+		return resp, vterrors.Wrapf(err, "invalid DdlStrategy: %s", req.DdlStrategy)
+	}
+
+	if len(req.UuidList) > 0 {
+		if err := executor.SetUUIDList(strings.Join(req.UuidList, ",")); err != nil {
```
shlomi-noach, line 221: In this new code, `vtctl/vtctl.go` splits `-uuid_list` variable to array of strings; it is now wasteful to concatenate them again, only to be parsed again inside `SetUUIDList`. Let's change function signature to accept `[]string`

go/vt/vtctl/vtctl.go
```diff
@@ -3240,37 +3241,41 @@ func commandApplySchema(ctx context.Context, wr *wrangler.Wrangler, subFlags *fl
...
-	if err := executor.SetDDLStrategy(*ddlStrategy); err != nil {
+
+	log.Info("Calling ApplySchema on VtctldServer")
+	resp, err := wr.VtctldServer().ApplySchema(ctx, &vtctldatapb.ApplySchemaRequest{
+		Keyspace:                keyspace,
+		AllowLongUnavailability: *allowLongUnavailability,
+		DdlStrategy:             *ddlStrategy,
+		Sql:                     strings.Split(change, ","),
+		SkipPreflight:           *skipPreflight,
+		UuidList:                strings.Split(*uuidList, ","),
```
shlomi-noach, line 3264:
```suggestion
		UuidList:                textutil.SplitDelimitedList(*uuidList),
```

`textutil.SplitDelimitedList()` also safely ignores spaces or empty entries

shlomi-noach (changes requested):
Minor request for change. Overall Online-DDL-wise this looks good! (And, @Phanatic , as I believe this is your first Vitess PR -- this is really well done and quite a daunting first change!)

go/vt/schemamanager/tablet_executor_test.go
Phanatic: This is fixed in https://github.com/vitessio/vitess/pull/9537/commits/4b37d644f3d3cb98b7d5c3df65ea8f6708652fc3

deepthi (approved):
Nice work.
Echoing Shlomi, this is quite complex for a first PR. well done!
We do need to update the command reference on the website to document the new flag.
https://vitess.io/docs/13.0/reference/programs/vtctl/schema-version-permissions/#applyschema

Issue #9525, opened 2022-01-18
Support checking workflow lag before initiating SwitchTraffic (SwitchReads/SwitchWrites)


aquarapid:
Scenario:
  * On a busy keyspace, a MoveTables (or Reshard) SwitchTraffic operation will today blindly switch reads, and then writes, as long as the copy phase is complete
  * If, due to heavy traffic, the vreplication lag is significant, the SwitchWrites part of SwitchTraffic will eventually fail (due to the 30s default operation timeout), but at that point write downtime will already have been incurred.
  * A workaround for this currently is to manually check the workflow vreplication lag (via a `Workflow... show` or the newer `MoveTables ... Progress` or `Reshard ... Progress`) before executing the `SwitchTraffic`
  * However, the user-friendly way of doing this might to do this check first, with a configurable "acceptable" lag, before executing either of the switching operations and incurring any read/write downtime.

kristee-planetscale:
@mvh-stripe

mattlord:
Closing this for now as I believe the issue was addressed in https://github.com/vitessio/vitess/pull/9538. Please let me know if that's not right and we can re-open this. Thanks!

PR #9538, opened 2022-01-19
SwitchTraffic: check vreplication lag before switching

rohit-nayak-ps:
## Description
This PR adds logic to measure the estimated lag between the last transaction at the source and the last transaction seen by the target. The target persists the transaction timestamp of each seen event in `_vt.vreplication`. The target may not receive events from the source in two cases: 
* where there is no activity at the source relevant to the target
* when the source is disconnected from the target (due to a vstreamer error, network issue or source tablets being unavailable). 

To differentiate between this the source starts sending "heartbeats" (special internal VEvents) every second if there are no actual events to be sent. As part of this PR we also start storing the last heartbeat time in `_vt.vreplication`.  

We calculate the transaction lag based on the transaction timestamp and the last heartbeat for each stream and the maximum across streams is used as the workflow lag: https://github.com/vitessio/vitess/blob/dc74ebcc2d1c70f36030ca07ec857c009afd6954/go/vt/wrangler/vexec.go#L536 

We now add a check before switching traffic (both read and write) to ensure:
1. that the lag is within acceptable bounds (defined by the duration flag `-max_transaction_lag_allowed`, default 30s) 
2. that there is no stream that has an error
3. that we are not still in the copy phase (this already exists)

Note that the `Workflow Show` command does already show a MaxVReplicationLag. This was implemented more to determine if there was an issue with the source not being available than measuring transaction lag. Since the name is already in-use, for backward compatibility, a new variable MaxVReplicationTransactionLag has been added for the purposes of this PR.

Other changes:
* Minor refactoring of the test framework for ease of maintenance
* Intial doc to describe the test framework and changes needed for this PR. We will update this in upcoming PRs.

New flag documented in website PR: https://github.com/vitessio/website/pull/957

## Related Issue(s)
https://github.com/vitessio/vitess/issues/9525

## Checklist

### DIFF
--- a/go/sqltypes/named_result.go
+++ b/go/sqltypes/named_result.go
@@ -110,6 +110,22 @@ func (r RowNamedValues) AsBool(fieldName string, def bool) bool {
 	return def
 }
 
+// ToBytes returns the named field as a byte array
+func (r RowNamedValues) ToBytes(fieldName string) ([]byte, error) {
+	if v, ok := r[fieldName]; ok {
+		return v.ToBytes()
+	}
+	return nil, ErrNoSuchField
+}
+
+// AsBytes returns the named field as a byte array, or default value if nonexistent/error
+func (r RowNamedValues) AsBytes(fieldName string, def []byte) []byte {
+	if v, err := r.ToBytes(fieldName); err == nil {
+		return v
+	}
+	return def
+}
+
 // NamedResult represents a query result with named values as opposed to ordinal values.
 type NamedResult struct {
 	Fields       []*querypb.Field `json:"fields"`
--- a/go/test/endtoend/tabletgateway/buffer/reshard/sharded_buffer_test.go
+++ b/go/test/endtoend/tabletgateway/buffer/reshard/sharded_buffer_test.go
@@ -19,6 +19,11 @@ package reshard
 import (
 	"fmt"
 	"testing"
+	"time"
+
+	"github.com/buger/jsonparser"
+
+	"vitess.io/vitess/go/vt/log"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
@@ -28,12 +33,42 @@ import (
 	"vitess.io/vitess/go/test/endtoend/cluster"
 )
 
+const (
+	maxWait              = 10 * time.Second
+	acceptableLagSeconds = 5
+)
+
+func waitForLowLag(t *testing.T, clusterInstance *cluster.LocalProcessCluster, keyspace, workflow string) {
+	var lagSeconds int64
+	waitDuration := 500 * time.Millisecond
+	duration := maxWait
+	for duration > 0 {
+		output, err := clusterInstance.VtctlclientProcess.ExecuteCommandWithOutput("Workflow", fmt.Sprintf("%s.%s", keyspace, workflow), "Show")
+		require.NoError(t, err)
+		lagSeconds, err = jsonparser.GetInt([]byte(output), "MaxVReplicationTransactionLag")
+
+		require.NoError(t, err)
+		if lagSeconds < acceptableLagSeconds {
+			log.Infof("waitForLowLag acceptable for workflow %s, keyspace %s, current lag is %d", workflow, keyspace, lagSeconds)
+			break
+		} else {
+			log.Infof("waitForLowLag too high for workflow %s, keyspace %s, current lag is %d", workflow, keyspace, lagSeconds)
+		}
+		time.Sleep(waitDuration)
+		duration -= waitDuration
+	}
+
+	if duration <= 0 {
+		t.Fatalf("waitForLowLag timed out for workflow %s, keyspace %s, current lag is %d", workflow, keyspace, lagSeconds)
+	}
+}
+
 func reshard02(t *testing.T, clusterInstance *cluster.LocalProcessCluster, keyspaceName string, reads, writes buffer.QueryEngine) {
 	keyspace := &cluster.Keyspace{Name: keyspaceName}
 	err := clusterInstance.StartKeyspace(*keyspace, []string{"-80", "80-"}, 1, false)
 	require.NoError(t, err)
-
-	workflow := keyspaceName + ".buf2buf"
+	workflowName := "buf2buf"
+	workflow := fmt.Sprintf("%s.%s", keyspaceName, "buf2buf")
 
 	err = clusterInstance.VtctlclientProcess.ExecuteCommand("Reshard", "-source_shards", "0", "-target_shards", "-80,80-", "Create", workflow)
 	require.NoError(t, err)
@@ -42,6 +77,7 @@ func reshard02(t *testing.T, clusterInstance *cluster.LocalProcessCluster, keysp
 	reads.ExpectQueries(25)
 	writes.ExpectQueries(25)
 
+	waitForLowLag(t, clusterInstance, keyspaceName, workflowName)
 	err = clusterInstance.VtctlclientProcess.ExecuteCommand("Reshard", "-tablet_types=rdonly,replica", "SwitchTraffic", workflow)
 	require.NoError(t, err)
 
--- a/go/test/endtoend/tabletgateway/buffer/reshard/sharded_buffer_test.go
+++ b/go/test/endtoend/tabletgateway/buffer/reshard/sharded_buffer_test.go
@@ -48,7 +48,7 @@ func waitForLowLag(t *testing.T, clusterInstance *cluster.LocalProcessCluster, k
 		lagSeconds, err = jsonparser.GetInt([]byte(output), "MaxVReplicationTransactionLag")
 
 		require.NoError(t, err)
-		if lagSeconds < acceptableLagSeconds {
+		if lagSeconds <= acceptableLagSeconds {
 			log.Infof("waitForLowLag acceptable for workflow %s, keyspace %s, current lag is %d", workflow, keyspace, lagSeconds)
 			break
 		} else {
--- a/go/test/endtoend/vreplication/vreplication_test.go
+++ b/go/test/endtoend/vreplication/vreplication_test.go
@@ -948,7 +948,7 @@ func waitForLowLag(t *testing.T, keyspace, workflow string) {
 		lagSeconds, err = jsonparser.GetInt([]byte(output), "MaxVReplicationTransactionLag")
 
 		require.NoError(t, err)
-		if lagSeconds < acceptableLagSeconds {
+		if lagSeconds <= acceptableLagSeconds {
 			log.Infof("waitForLowLag acceptable for workflow %s, keyspace %s, current lag is %d", workflow, keyspace, lagSeconds)
 			break
 		} else {
--- a/go/test/endtoend/vreplication/vreplication_test.go
+++ b/go/test/endtoend/vreplication/vreplication_test.go
@@ -27,6 +27,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/buger/jsonparser"
+
 	"vitess.io/vitess/go/sqltypes"
 	"vitess.io/vitess/go/test/utils"
 	querypb "vitess.io/vitess/go/vt/proto/query"
@@ -62,6 +64,7 @@ const (
 	deleteOpenTxQuery = "delete from customer where name = 'openTxQuery'"
 
 	merchantKeyspace = "merchant-type"
+	maxWait          = 10 * time.Second
 )
 
 func init() {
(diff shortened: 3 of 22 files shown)


aquarapid:
One question/comment I would have is what the default for this timeout check should be.  In the interests of not changing current behavior, an argument can be made for it to be infinite?  Or at least equal to the switch wait timeout?

mattlord:
> One question/comment I would have is what the default for this timeout check should be. In the interests of not changing current behavior, an argument can be made for it to be infinite? Or at least equal to the switch wait timeout?

Why would we NOT re-use the existing flag? That being: [`-timeout`](https://vitess.io/docs/13.0/reference/vreplication/switchtraffic/#-timeout)

If we check replication lag ahead of time and see that it's beyond that window, it's reasonable to assume it's not likely to catch up within that window.

See problems/issues with that?

rohit-nayak-ps:
> Why would we NOT re-use the existing flag? That being: [`-timeout`]

I thought of using a different flag here because:
* `-timeout` essentially specifies the maximum downtime that is acceptable while switching writes. This includes the overhead of coordinating between all the shards and across cells. 
* a new flag `-max_lag_allowed` would allow specifying a small lag, say 2 or 5 seconds, for both reads (minimize reading stale data) and writes

If we use a single flag for both we could end up not being able to specify a small lag because it will always timeout due to the overhead.

As Jacques suggested we could use the current wait timeout as default for the new flag, if we do decide to go with two separate flags.

go/vt/wrangler/vexec.go
```diff
@@ -403,18 +406,18 @@ type ReplicationStatus struct {
 	CopyState []copyState
 }
 
-func (wr *Wrangler) getReplicationStatusFromRow(ctx context.Context, row []sqltypes.Value, primary *topo.TabletInfo) (*ReplicationStatus, string, error) {
+func (wr *Wrangler) getReplicationStatusFromRow(ctx context.Context, row sqltypes.RowNamedValues, primary *topo.TabletInfo) (*ReplicationStatus, string, error) {
```
rohit-nayak-ps, line 409: The changes in this function are mainly refactoring to use name references instead of index-based (and accessing the new `time_heartbeat` column)

rohit-nayak-ps:
One functionality that might "break" due to this PR is that some SwitchTraffics that succeeded in the past will temporarily error out since the lag is too high. This can happen, for example, if the user created a workflow and immediately tried to SwitchTraffic without letting the workflows catchup. Earlier it would have just taken longer for the call to return since it would mostly like manage to catchup before the `wait timeout` (def: 30s) (or just timeout if there was too much data). 

Now it will return an error and user will need to monitor the lag and/or try again later. This is the correct way to use SwitchTraffic and one we want to encourage (and indeed the reason for this PR). So I don't believe we should keep a larger default but noting it here since it might give rise to a few extra support questions.

go/vt/vtctl/vtctl.go
```diff
@@ -2295,15 +2295,21 @@ func getSourceKeyspace(clusterKeyspace string) (clusterName string, sourceKeyspa
...
 	reverseReplication := subFlags.Bool("reverse_replication", true, "Also reverse the replication (default true). -reverse_replication is only supported for SwitchTraffic.")
 	keepData := subFlags.Bool("keep_data", false, "Do not drop tables or shards (if true, only vreplication artifacts are cleaned up).  -keep_data is only supported for Complete and Cancel.")
 	keepRoutingRules := subFlags.Bool("keep_routing_rules", false, "Do not remove the routing rules for the source keyspace.  -keep_routing_rules is only supported for Complete and Cancel.")
 	autoStart := subFlags.Bool("auto_start", true, "If false, streams will start in the Stopped state and will need to be explicitly started")
 	stopAfterCopy := subFlags.Bool("stop_after_copy", false, "Streams will be stopped once the copy phase is completed")
+	maxTransactionLagAllowed := subFlags.Duration("max_transaction_lag_allowed", defaultMaxTransactionLagAllowed, "Allow traffic to be switched only if vreplication lag is below this (in seconds)")
```
deepthi, line 2312: Why are we calling this `transaction_lag` instead of `replication_lag`?

mattlord: I would also expect/prefer this be called `replication_lag` or `vreplication_lag`.

go/sqltypes/named_result.go
```diff
@@ -110,6 +110,22 @@ func (r RowNamedValues) AsBool(fieldName string, def bool) bool {
...
+// ToBytes returns the named field as a byte array
+func (r RowNamedValues) ToBytes(fieldName string) ([]byte, error) {
+	if v, ok := r[fieldName]; ok {
+		return v.ToBytes()
+	}
+	return nil, ErrNoSuchField
+}
+
+// AsBytes returns the named field as a byte array, or default value if nonexistent/error
+func (r RowNamedValues) AsBytes(fieldName string, def []byte) []byte {
```
mattlord, line 122: Nit, but I would call the parameter `default` or `defaultValue`. At first I missed the comment and thought we were passing that in as a pointer to be written to.

go/test/endtoend/vreplication/vreplication_test.go
```diff
@@ -933,9 +936,35 @@ func verifyClusterHealth(t *testing.T, cluster *VitessCluster) {
 	iterateTablets(t, cluster, checkTabletHealth)
 }
 
+const acceptableLagSeconds = 5
+
+func waitForLowLag(t *testing.T, keyspace, workflow string) {
```
mattlord, line 941: Possible to share this code between the two end2end tests? If not, see my earlier comments about this code in `sharded_buffer_test.go`

go/vt/binlog/binlogplayer/binlog_player.go
```diff
@@ -564,6 +564,9 @@ var AlterVReplicationTable = []string{
 	"ALTER TABLE _vt.vreplication ADD KEY workflow_idx (workflow(64))",
 	"ALTER TABLE _vt.vreplication ADD COLUMN rows_copied BIGINT(20) NOT NULL DEFAULT 0",
 	"ALTER TABLE _vt.vreplication ADD COLUMN tags VARBINARY(1024) NOT NULL DEFAULT ''",
+
+	// records the time of the last heartbeat. Heartbeats are only received if the source has no recent events
+	"ALTER TABLE _vt.vreplication ADD COLUMN time_heartbeat BIGINT(20) NOT NULL DEFAULT 0",
```
mattlord, line 569: We should remove the default display widths here at some point (could come later), that will cause a warning in 8.0 and fail later on:
```
"ALTER TABLE _vt.vreplication ADD COLUMN time_heartbeat BIGINT NOT NULL DEFAULT 0",
```

```
mysql> select @@version;
+-----------+
| @@version |
+-----------+
| 8.0.27    |
+-----------+
1 row in set (0.00 sec)

mysql> create table uttest (time_heartbeat BIGINT(20) NOT NULL DEFAULT 0);
Query OK, 0 rows affected, 1 warning (0.01 sec)

mysql> show warnings;
+---------+------+------------------------------------------------------------------------------+
| Level   | Code | Message                                                                      |
+---------+------+------------------------------------------------------------------------------+
| Warning | 1681 | Integer display width is deprecated and will be removed in a future release. |
+---------+------+------------------------------------------------------------------------------+
1 row in set (0.00 sec)

mysql> show create table uttest\G
       Table: uttest
Create Table: CREATE TABLE `uttest` (
  `time_heartbeat` bigint NOT NULL DEFAULT '0'
1 row in set (0.00 sec)
```

Meanwhile, 5.7 will add the default display width (20 for this type) automatically.

go/vt/binlog/binlogplayer/binlog_player.go
```diff
@@ -659,12 +661,12 @@ func GenerateUpdateRowsCopied(uid uint32, rowsCopied int64) string {
...
 
-// GenerateUpdateTime returns a statement to update time_updated in the _vt.vreplication table.
-func GenerateUpdateTime(uid uint32, timeUpdated int64) (string, error) {
+// GenerateUpdateHeartbeat returns a statement to record the latest heartbeat in the _vt.vreplication table.
+func GenerateUpdateHeartbeat(uid uint32, timeUpdated int64) (string, error) {
 	if timeUpdated == 0 {
 		return "", fmt.Errorf("timeUpdated cannot be zero")
 	}
-	return fmt.Sprintf("update _vt.vreplication set time_updated=%v where id=%v", timeUpdated, uid), nil
+	return fmt.Sprintf("update _vt.vreplication set time_updated=%v, time_heartbeat=%v where id=%v", timeUpdated, timeUpdated, uid), nil
```
mattlord, line 669: Is the `time_heartbeat` ever different than `time_updated` in the current code? I didn't immediately see where they differ. Not sure if I'm missing something or perhaps we are allowing for future behavior.

go/vt/wrangler/doc_test.md
```diff
@@ -0,0 +1,67 @@
+## Wrangler test framework
```
mattlord, line 1: 😍

go/vt/wrangler/doc_test.md
```diff
@@ -0,0 +1,67 @@
...
+`go/vt/mysqlctl/fakemysqldaemon/fakemysqldaemon.go`
+
+Used to set primary positions to provide/validate gtids.
+
+#### The fake db client
+
+`go/vt/wrangler/fake_dbclient_test.go`
+
+This defines a mock db which is limited in scope to the vreplication engine. All queries that it mocks are related to
+the `_vt` database only.
```
mattlord, line 55: IMO, it's worth pointing out that these also serve to validate that the expected set of queries were generated. When updating/adding tests, you will need to tell it what queries are valid (and what it should ignore). Right?

go/vt/wrangler/vexec.go
```diff
@@ -346,7 +347,10 @@ type ReplicationStatusResult struct {
 	TargetLocation ReplicationLocation
 	// MaxVReplicationLag represents the maximum vreplication lag seen across all shards.
 	MaxVReplicationLag int64
-
+	// MaxVReplicationTransactionLag represents the lag between the current time and the timestamp of the last transaction seen across all shards.
```
mattlord, line 350: last replicated transaction timestamp OR heartbeat timestamp (if there have been no writes to replicate from the source), right?

go/vt/wrangler/vexec.go
```diff
@@ -499,11 +528,41 @@ func (wr *Wrangler) getStreams(ctx context.Context, workflow, keyspace string) (
...
+			// MaxVReplicationTransactionLag estimates the actual lag between the source and the target
+			// If we are still processing source events it is the difference b/w current time and the timestamp of the last event
+			// If heartbeats are more recent than the last event, then the lag is the time since the last heartbeat as
+			// there can be an actual event immediately after the heartbeat, but which has not yet
+			// been processed on the target
+			// We don't allow switching during the copy phase, so in that case we just return a large lag.
+			// All timestamps are in seconds since epoch
+			lastTransactionTimestamp := status.TransactionTimestamp
+			lastHeartbeatTime := status.TimeHeartbeat
+			if status.State != "Copying" {
```
mattlord, line 552: IMO it's easier to read if the predicate is close to the action. In this case what to do when State == Copying is much later. i.e. I think it's easier to follow if we did:
```
if status.State == "Copying" {
...
} else {
}
```

I'm guilty of doing it the current way though too.

mattlord (approved):
I had a few nits and questions, but otherwise LGTM! I really appreciate the time you put into the tests, docs, and refactoring! 

I'll approve so that you're free to merge after addressing any valid issues.

go/vt/vtctl/vtctl.go
rohit-nayak-ps: Changed to use `replication_lag`. 

The reason for the initially using transaction_lag was that there were already a metric `MaxVReplicationLag` measuring just the vreplication lag for liveness of source streams: the time that the target last saw the source. The new metric is called `MaxVReplicationTransactionLag`: I thought of renaming the previous one to a better name but it is already used by some users and would have been a breaking change.  I improved the comments around these metrics.

go/test/endtoend/tabletgateway/buffer/reshard/sharded_buffer_test.go
```diff
@@ -28,12 +33,42 @@ import (
...
+	var lagSeconds int64
+	waitDuration := 500 * time.Millisecond
+	duration := maxWait
+	for duration > 0 {
+		output, err := clusterInstance.VtctlclientProcess.ExecuteCommandWithOutput("Workflow", fmt.Sprintf("%s.%s", keyspace, workflow), "Show")
+		require.NoError(t, err)
+		lagSeconds, err = jsonparser.GetInt([]byte(output), "MaxVReplicationTransactionLag")
+
+		require.NoError(t, err)
+		if lagSeconds < acceptableLagSeconds {
```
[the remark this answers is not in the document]
rohit-nayak-ps: fixed

go/test/endtoend/tabletgateway/buffer/reshard/sharded_buffer_test.go
```diff
@@ -28,12 +33,42 @@ import (
...
+			log.Infof("waitForLowLag acceptable for workflow %s, keyspace %s, current lag is %d", workflow, keyspace, lagSeconds)
+			break
+		} else {
+			log.Infof("waitForLowLag too high for workflow %s, keyspace %s, current lag is %d", workflow, keyspace, lagSeconds)
+		}
+		time.Sleep(waitDuration)
+		duration -= waitDuration
+	}
+
+	if duration <= 0 {
```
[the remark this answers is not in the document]
rohit-nayak-ps: The `if` has a break, so the duration will not decrement if we were successful.

go/vt/binlog/binlogplayer/binlog_player.go
rohit-nayak-ps: When a transaction is seen and the `pos` column is updated with the recent gtid we also update `time_updated` (and not `time_heartbeat`)

go/vt/wrangler/vexec.go
rohit-nayak-ps: agree, fixed

go/vt/wrangler/vexec.go
rohit-nayak-ps: updated as per your suggestion

go/test/endtoend/vreplication/vreplication_test.go
rohit-nayak-ps: Yeah, classic anti-pattern :( Because they are in different packages and we don't yet have a `utils` package to reuse such code, I took the short-cut. Merging as is now to catch the release deadline, will address "soon".

go/sqltypes/named_result.go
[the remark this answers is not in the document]
rohit-nayak-ps: Agree, `def` is confusing. Since used in all other functions, so I went with this rather than just name it differently here or fix it everywhere. But I should have ...

PR #9540, opened 2022-01-19
Allow cross-build to work for cross-architecture too (e.g. arm64 on amd64)

### LINKED ISSUE #9539: cross-build fails when trying to build arm64 on amd64 linux due to rice exec failure
E.g.:

```
$ GOOS=linux GOARCH=arm64 make cross-build
Wed Jan 19 14:28:04 PST 2022: Building source tree
fork/exec /tmp/go-build3207728525/b001/exe/rice: exec format error
make: *** [Makefile:83: cross-build] Error 1
```

aquarapid:
`GOOS=linux GOARCH=arm64 make cross-build` currently fails on amd64 (linux), because we invoke the rice binary, which will be built using arm64, and the exec will fail.

Fixes #9539 

## Checklist


mattlord (approved):

deepthi:
This make target isn't relevant to the failing tests. Assuming they are flaky, we'll investigate them separately.

deepthi (approved):

PR #9543, opened 2022-01-20
Refactor routing primitive

harshit-gangal:
## Description

This PR refactors the route, update and delete engine primitive to use common routing logic. This makes it simpler to change the routing logic in one place now.

## Checklist

### DIFF
--- a/go/vt/vtgate/engine/routing.go
+++ b/go/vt/vtgate/engine/routing.go
@@ -71,6 +71,9 @@ type RoutingParameters struct {
 	// Required for - DBA
 	sysTableTableSchema []evalengine.Expr
 	sysTableTableName   map[string]evalengine.Expr
+
+	// Required for - ByDestination
+	targetDestination key.Destination
 }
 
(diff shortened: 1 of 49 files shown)


go/vt/vtgate/engine/routing.go
```diff
@@ -0,0 +1,604 @@
...
+	vtrpcpb "vitess.io/vitess/go/vt/proto/vtrpc"
+	"vitess.io/vitess/go/vt/sqlparser"
+	"vitess.io/vitess/go/vt/srvtopo"
+	"vitess.io/vitess/go/vt/vterrors"
+	"vitess.io/vitess/go/vt/vtgate/evalengine"
+	"vitess.io/vitess/go/vt/vtgate/vindexes"
+)
+
+// Opcode is a number representing the opcode
+// for any engine primitve.
```
systay, line 36: nit: typo `primitve`, and it doesn't represent the opcode for _any_ engine primitive, only the routing ones

GuptaManan100 (approved):
Apart from Andres's comment, I think everything else looks great!

go/vt/vtgate/engine/routing.go
harshit-gangal: will fix the comment in follow up PR

PR #9545, opened 2022-01-20
Planner improvements

systay:
## Description
Two minor planner improvements - first a query hint that allow users to select the planner to use. Sometimes we want to be able to specify a different planner than the one the vtgate is configured with, and this query hint enables this.

```sql
select /*vt+ PLANNER=gen4 */ * from commerce;
```

Also re-introduces the left-to-right planner which mimics `straight join` on mysql, which was lost during the refactor of gen4 this week.

### DIFF
--- a/go/test/endtoend/vtgate/misc_test.go
+++ b/go/test/endtoend/vtgate/misc_test.go
@@ -571,10 +571,10 @@ func TestSQLSelectLimit(t *testing.T) {
 		*/
 
 		//	without order by the results are not deterministic for testing purpose. Checking row count only.
-		qr := utils.Exec(t, conn, "select /* GEN4_COMPARE_ONLY_GEN4 */ uid, msg from t7_xxhash union all select uid, msg from t7_xxhash")
+		qr := utils.Exec(t, conn, "select /*vt+ PLANNER=gen4 */ uid, msg from t7_xxhash union all select uid, msg from t7_xxhash")
 		assert.Equal(t, 2, len(qr.Rows))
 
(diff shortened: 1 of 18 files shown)


systay:
@frouioui WDYT - should we use this to replace the `GEN4_COMPARE_ONLY_GEN4` annotations?

GuptaManan100 (approved):
Do we need a test for the left to right solver?

harshit-gangal:
> @frouioui WDYT - should we use this to replace the `GEN4_COMPARE_ONLY_GEN4` annotations?

this would basically mean that set PLANNER=Gen4 in the comment. I think we can do that.

harshit-gangal (approved):

GuptaManan100 (approved):

harshit-gangal:
looks like we have something already `GEN4_COMPARE_ONLY_V3`, `GEN4_COMPARE_ONLY_GEN4`but those names were not intuitive to use, this looks good.

frouioui:
> @frouioui WDYT - should we use this to replace the `GEN4_COMPARE_ONLY_GEN4` annotations?

Yes, that is a good idea!

GuptaManan100 (approved):

aquarapid:
Noting the valid values for the comment (`/*vt+ PLANNER=xxx */`):
 * `v3`
 * `gen4`
 * `gen4greedy` or `greedy`
 * `left2right`
 * `gen4fallback`
 * `gen4comparev3`

systay:
Docs added: https://vitess.io/docs/13.0/reference/compatibility/query_planner/

PR #9549, opened 2022-01-21
Fix text handling in Vitess native evaluation

vmg:
## Description

This PR started as an straightforward implementation of a couple functions for the `evalengine` (`COALESCE`, `LEAST` and `GREATEST`), but while implementing this I discovered that there was an underlying issue in the way Vitess was handling `Value` types from MySQL and that was preventing me from implementing these APIs cleanly: both when normalizing, and when interacting with bind variables, Vitess was coercing all byte-sequence data types to `VARBINARY` through the codebase, even those that MySQL tagged as representing "text" and originally sent as `VARCHAR`.

Hence, this PR overhauls this handling throughout the codebase so we can more accurately match MySQLs behavior. If you review the commits, you'll see the there are actually very few changes to the code: most of the fixes are trivial replacements in tests that were expecting to receive text values with a `VARBINARY` type and are now (correctly) receiving them as `VARCHAR`.

The only notable change is https://github.com/vitessio/vitess/pull/9549/commits/817fe5485c43d8535307a9bd3f1d3a8ee8d11181, in the `vstreamer` code. During our collation work this last year, we left some TODOs about wiring up collations throughout the codebase -- I've fixed one of them here because previously VStreamer filtering operators never handled "text" (they always compared using binary comparison because they saw as columns are being binary), but now textual columns are properly compared using collations and hence need to know what collation to use for comparison. cc @rohit-nayak-ps for review

The other major change that this PR introduces are planning changes for some query plans. These are well expected, most notably we can now plan a type of query with `COALESCE` which we couldn't handle before. @frouioui has reviewed and approved the new plans already.

**This PR introduces a minor breaking change to the `vtgate`s**, which is why I'd like to ship it before Tuesday's RC1 tagging: certain queries (most notably, selecting system variables) that previously returned from VTGate as `VARBINARY` blobs will now be `VARCHAR` instead. **This matches MySQL's behavior** (while previously, it didn't), and it shouldn't have an visible effect on the vast majority of MySQL drivers/connectors/clients because these perform coercion on the client to whatever type the user has requested. Furthermore, a lot of dynamic programming languages that would see a difference don't really have different types to differentiate between "bytes" and "string", and were already converting all `VARBINARY` rows to strings anyway, so they will see no difference here either.

## Checklist


go/vt/vtgate/evalengine/eval_result.go
```diff
@@ -126,6 +127,17 @@ func (er *EvalResult) int64() int64 {
...
+func (er *EvalResult) signedInteger() bool {
+	switch er.typeof() {
+	case querypb.Type_INT8, querypb.Type_INT16, querypb.Type_INT32, querypb.Type_INT64:
+		return true
+	case querypb.Type_UINT8, querypb.Type_UINT16, querypb.Type_UINT32, querypb.Type_UINT64:
+		return false
+	default:
+		panic("signedInteger called on non-integer")
+	}
+}
```
frouioui, line 139: It's not super important but in my opinion, calling this method something like `isSignedInteger` would make more sense considering that other methods with a similar name (`int64()`, `uint64()`, `decimal()`) have a different logic than `signedInteger()`

go/vt/vtgate/planbuilder/testdata/unsupported_cases.txt
```diff
@@ -71,7 +71,57 @@ Gen4 plan same as above
 # left join where clauses #2
 "select user.id from user left join user_extra on user.col = user_extra.col where coalesce(user_extra.col, 4) = 5"
 "unsupported: cross-shard left join and where clause"
-Gen4 error: expr cannot be converted, not supported: coalesce(user_extra.col, 4) = 5
+{
```
frouioui, line 74: We should move this plan test to `from_cases.txt`

frouioui (approved):
The new plans and your code changes look good to me, I am not seeing any specific problem

go/vt/vtgate/evalengine/eval_result.go
vmg: I've got a better idea: let's remove it since it's no longer used!

rohit-nayak-ps (approved):
Very nice. The vstreamer changes look good to me. 

I will separately check if we need to enhance the existing vreplication tests (esp e2e) for improved charset/data type coverage.

PR #3737, opened 2018-03-08
Prevent unescaped query from being returned to clients

acharis:
The context: https://vitess.slack.com/archives/C0PQY0PTK/p1520444718000317

I'm stripping out the query from the SQLError, because we're going to add in a properly escaped version right after. I've tried to leave the terseErrors codepath alone.

Signed-off-by: Alex Charis <acharis@hubspot.com>

PR #9550, opened 2022-01-21
Support sanitizing potentially sensitive information in log messages

mattlord:
## Description
Add a new `vttablet` flag:
   - Name: `sanitize_log_messages`
   - Default: false (no change in behavior, opt-in)
   - Description: Do not log potentially sensitive information in INFO, WARNING, and ERROR log messages such as query parameters

Use this where we make calls to `log.Info[f]`, `log.Warning[f]`, and `log.Error[f]` (glog) to ensure that we are sanitizing any potentially sensitive information. For example, here as this includes queries in the log message:
```
log.Warningf("killing transaction (exceeded timeout: %v): %s", tp.Timeout(), conn.String(tp.env.Config().SanitizeLogMessages))
```

Initially this will be used primarily to redact query/bind parameters from logged queries but this can be used going forward to sanitize any potentially sensitive information in the log messages.

## Related Issue(s)
Follow-up to: https://github.com/vitessio/vitess/pull/9470
Fixes: https://github.com/vitessio/vitess/issues/9469

## Checklist
- [x] Should this PR be backported? Yes, to release-13.0
- [x] Tests were added/updated
- [x] Documentation PR: https://github.com/vitessio/website/pull/950

### DIFF
--- a/go/vt/vttablet/tabletserver/tabletserver.go
+++ b/go/vt/vttablet/tabletserver/tabletserver.go
@@ -1364,8 +1364,10 @@ func (tsv *TabletServer) convertAndLogError(ctx context.Context, sql string, bin
 		errnum := sqlErr.Number()
 		if tsv.TerseErrors && len(bindVariables) != 0 && errCode != vtrpcpb.Code_FAILED_PRECONDITION {
 			err = vterrors.Errorf(errCode, "(errno %d) (sqlstate %s)%s: %s", errnum, sqlState, callerID, queryAsString(sql, bindVariables, tsv.TerseErrors))
+			// TerseErrors should generally NOT impact log messages -- rather client errors -- but keeping
+			// this existing exception in place so as not to change the behavior for users of the flag.
 			if logMethod != nil {
-				message = fmt.Sprintf("(errno %d) (sqlstate %s)%s: %s", errnum, sqlState, callerID, truncateSQLAndBindVars(sql, bindVariables, tsv.Config().SanitizeLogMessages))
+				message = fmt.Sprintf("(errno %d) (sqlstate %s)%s: %s", errnum, sqlState, callerID, truncateSQLAndBindVars(sql, bindVariables, tsv.TerseErrors))
 			}
 		} else {
 			err = vterrors.Errorf(errCode, "%s (errno %d) (sqlstate %s)%s: %s", sqlErr.Message, errnum, sqlState, callerID, queryAsString(sql, bindVariables, tsv.TerseErrors))
@@ -1392,23 +1394,25 @@ func (tsv *TabletServer) convertAndLogError(ctx context.Context, sql string, bin
 }
 
 // truncateSQLAndBindVars calls TruncateForLog which:
-//  splits off trailing comments, truncates the query, and re-adds the trailing comments
-// and if sanitize is not false appends quoted bindvar:value pairs in sorted order
-// truncates the resulting string
+//  splits off trailing comments, truncates the query, re-adds the trailing comments,
+//  if sanitize is false appends quoted bindvar:value pairs in sorted order, and
+//  lastly it truncates the resulting string
 func truncateSQLAndBindVars(sql string, bindVariables map[string]*querypb.BindVariable, sanitize bool) string {
 	truncatedQuery := sqlparser.TruncateForLog(sql)
 	buf := &bytes.Buffer{}
 	fmt.Fprintf(buf, "BindVars: {")
-	if sanitize {
-		fmt.Fprintf(buf, "[REDACTED]")
-	} else {
-		var keys []string
-		for key := range bindVariables {
-			keys = append(keys, key)
-		}
-		sort.Strings(keys)
-		for _, key := range keys {
-			fmt.Fprintf(buf, "%s: %q", key, fmt.Sprintf("%v", bindVariables[key]))
+	if len(bindVariables) > 0 {
+		if sanitize {
+			fmt.Fprintf(buf, "[REDACTED]")
+		} else {
+			var keys []string
+			for key := range bindVariables {
+				keys = append(keys, key)
+			}
+			sort.Strings(keys)
+			for _, key := range keys {
+				fmt.Fprintf(buf, "%s: %q", key, fmt.Sprintf("%v", bindVariables[key]))
+			}
 		}
 	}
 	fmt.Fprintf(buf, "}")
--- a/go/vt/vttablet/tabletserver/tabletserver.go
+++ b/go/vt/vttablet/tabletserver/tabletserver.go
@@ -1298,20 +1298,11 @@ func (tsv *TabletServer) handlePanicAndSendLogStats(
 	logStats *tabletenv.LogStats,
 ) {
 	if x := recover(); x != nil {
-		var errorMessage string
-		if tsv.TerseErrors && len(bindVariables) != 0 {
-			errorMessage = fmt.Sprintf(
-				"Uncaught panic for %v:\n%v\n%s",
-				queryAsString(sql, nil),
-				x,
-				tb.Stack(4) /* Skip the last 4 boiler-plate frames. */)
-		} else {
-			errorMessage = fmt.Sprintf(
-				"Uncaught panic for %v:\n%v\n%s",
-				queryAsString(sql, bindVariables),
-				x,
-				tb.Stack(4) /* Skip the last 4 boiler-plate frames. */)
-		}
+		errorMessage := fmt.Sprintf(
+			"Uncaught panic for %v:\n%v\n%s",
+			queryAsString(sql, bindVariables, (tsv.Config().SanitizeLogMessages || tsv.TerseErrors)),
+			x,
+			tb.Stack(4) /* Skip the last 4 boiler-plate frames. */)
 		log.Errorf(errorMessage)
 		terr := vterrors.Errorf(vtrpcpb.Code_UNKNOWN, "%s", errorMessage)
 		tsv.stats.InternalErrors.Add("Panic", 1)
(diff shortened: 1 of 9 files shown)


go/vt/vttablet/tabletserver/tabletserver.go
```diff
@@ -1476,26 +1467,20 @@ func (tsv *TabletServer) convertAndLogError(ctx context.Context, sql string, bin
 		sqlState := sqlErr.SQLState()
 		errnum := sqlErr.Number()
 		if tsv.TerseErrors && len(bindVariables) != 0 && errCode != vtrpcpb.Code_FAILED_PRECONDITION {
-			err = vterrors.Errorf(errCode, "(errno %d) (sqlstate %s)%s: %s", errnum, sqlState, callerID, queryAsString(sql, nil))
+			err = vterrors.Errorf(errCode, "(errno %d) (sqlstate %s)%s: %s", errnum, sqlState, callerID, queryAsString(sql, bindVariables, tsv.TerseErrors))
 			if logMethod != nil {
-				message = fmt.Sprintf("(errno %d) (sqlstate %s)%s: %s", errnum, sqlState, callerID, truncateSQLAndBindVars(sql, nil))
+				message = fmt.Sprintf("(errno %d) (sqlstate %s)%s: %s", errnum, sqlState, callerID, truncateSQLAndBindVars(sql, bindVariables, tsv.Config().SanitizeLogMessages))
 			}
 		} else {
```
GuptaManan100, line 1474: This part is a little confusing to me. As far as the documentation for the TerseErrors flag goes, it only talks about removing the bind variables, but it is also removing the error message, both from logs and errors!

go/vt/vttablet/tabletserver/tabletserver.go
```diff
@@ -1512,21 +1497,23 @@ func (tsv *TabletServer) convertAndLogError(ctx context.Context, sql string, bin
 
 // truncateSQLAndBindVars calls TruncateForLog which:
 //  splits off trailing comments, truncates the query, and re-adds the trailing comments
-// appends quoted bindvar: value pairs in sorted order
+// and if sanitize is not false appends quoted bindvar:value pairs in sorted order
```
GuptaManan100, line 1500: nit: `not false` -> `true`

GuptaManan100 (approved):
Everything else looks good to me, just want to clarify what all the 3 flags are intended to be doing - 
1. `redact-debug-ui-queries` - Remove bind variables to be shown on the UI debug pages.
2. `queryserver-config-terse-errors` - Removed bindVars from error messages. Also removes the SQL query from errors and logs?
3. `sanitize_log_messages` - Remove sensitive information including bind variables from log messages.

Please correct my understanding if I am wrong.

go/vt/vttablet/tabletserver/tabletserver.go
GuptaManan100, line 1474: This part is also changing the behaviour of the TerseErrors flag. So if the user has `queryserver-config-terse-errors` set, then they will start seeing the bind variables in the log messages too unless they switch on the new flag as well.

GuptaManan100 (changes requested):
The new flag isn't changing the default and as opt-in, but it is changing the behaviour of the `queryserver-config-terse-errors` flag. I think we shouldn't do that and hold off on that change until we decide to do so. Looking at the code I think there are 2 easy ways to fix it,
1. Set `SanitizeLogMessages` to true when `TerseErrors` is true.
2. Use `tsv.Config().SanitizeLogMessages || tsv.terseErrors` for logging changes.

deepthi:
> The new flag isn't changing the default and as opt-in, but it is changing the behaviour of the `queryserver-config-terse-errors` flag. I think we shouldn't do that and hold off on that change until we decide to do so. Looking at the code I think there are 2 easy ways to fix it,
> 
> 1. Set `SanitizeLogMessages` to true when `TerseErrors` is true.
> 2. Use `tsv.Config().SanitizeLogMessages || tsv.terseErrors` for logging changes.

I agree that as part of this PR we should NOT change the behavior of the `terse-errors` flag in any way.

mattlord:
> Everything else looks good to me, just want to clarify what all the 3 flags are intended to be doing -
> 
> 1. `redact-debug-ui-queries` - Remove bind variables to be shown on the UI debug pages.

Yep.

> 2. `queryserver-config-terse-errors` - Removed bindVars from error messages. Also removes the SQL query from errors and logs?

Well, the docs only said error messages. And the name and behavior of the flag also suggests the same. However, the -help output message also mentioned logs. I removed the mention of logs in the help output in this PR as I don't think that's right (was probably tacked on at some point to redact some specific log message). 

> 3. `sanitize_log_messages` - Remove sensitive information including bind variables from log messages.

Yep! I wanted a generic flag that we can re-use for sanitizing any sensitive info over time and not end up with N related flags.

go/vt/vttablet/tabletserver/tabletserver.go
mattlord: You're not the only one confused 😉

mattlord:
Thank you for the review, @deepthi and @GuptaManan100 ! I tried to address the comments here: https://github.com/vitessio/vitess/pull/9550/commits/eae4afab263577ce0ec1391f6175d6d9603eb411

Can you please let me know if you're good with the PR at this point? If so, then I will fix the unit tests. I want to put that off until the end and they will break again as other people add/modify tests and I have to rebase on main.

mattlord:
@deepthi and @GuptaManan100, I went ahead and fixed the unit tests. I also left a dedicated TerseErrors test, added a new dedicated SanitizeLogMessages test and combined the behavior testing where I think it made sense given that the test was for general behavior of sanitization/redaction. 

Please let me know if you have any additional comments/requests and I'll be happy to address them. Thanks again!

mattlord:
There is *at least* 1 place in `vtgate` where we also need to sanitize the log messages: https://github.com/vitessio/vitess/blob/main/go/vt/vtgate/vtgate.go#L550-L567

I will make this PR more general to cover `vtgate` and `vttablet` — in a uniform way.

deepthi (approved):

go/vt/vttablet/tabletserver/tabletserver.go
```diff
@@ -1372,26 +1365,24 @@ func (tsv *TabletServer) convertAndLogError(ctx context.Context, sql string, bin
...
+			err = vterrors.Errorf(errCode, "(errno %d) (sqlstate %s)%s: %s", errnum, sqlState, callerID, queryAsString(sql, bindVariables, tsv.TerseErrors))
 			if logMethod != nil {
-				message = fmt.Sprintf("(errno %d) (sqlstate %s)%s: %s", errnum, sqlState, callerID, truncateSQLAndBindVars(sql, nil))
+				// TerseErrors should generally NOT impact log messages -- rather client errors -- but keeping
+				// this existing exception in place so as not to change the behavior for users of the flag.
+				message = fmt.Sprintf("(errno %d) (sqlstate %s)%s: %s", errnum, sqlState, callerID, truncateSQLAndBindVars(sql, bindVariables, tsv.TerseErrors))
 			}
 		} else {
-			err = vterrors.Errorf(errCode, "%s (errno %d) (sqlstate %s)%s: %s", sqlErr.Message, errnum, sqlState, callerID, queryAsString(sql, bindVariables))
+			err = vterrors.Errorf(errCode, "%s (errno %d) (sqlstate %s)%s: %s", sqlErr.Message, errnum, sqlState, callerID, queryAsString(sql, bindVariables, tsv.TerseErrors))
```
GuptaManan100, line 1375: I think there is a behavioural change here for TerseErrors. If there is a statement with bindVariables and errCode is `FAILED_PRECONDITION`, the old codepath would not have redacted the bindVariables, but it will now. We should pass `false` to the `queryAsString` function call here.

go/vt/vttablet/tabletserver/tabletserver.go
```diff
@@ -1372,26 +1365,24 @@ func (tsv *TabletServer) convertAndLogError(ctx context.Context, sql string, bin
...
-				message = fmt.Sprintf("%v: %v", err, truncateSQLAndBindVars(sql, nil))
-			}
-		} else {
-			if logMethod != nil {
-				message = fmt.Sprintf("%v: %v", err, truncateSQLAndBindVars(sql, bindVariables))
-			}
+		if logMethod != nil {
+			// TerseErrors should generally NOT impact log messages -- rather client errors -- but keeping
+			// this existing exception in place so as not to change the behavior for users of the flag.
+			message = fmt.Sprintf("%v: %v", err, truncateSQLAndBindVars(sql, bindVariables, (tsv.TerseErrors || tsv.Config().SanitizeLogMessages)))
```
GuptaManan100, line 1385: Here too, `TerseErrors` will redact bindVariables for queries that it didn't previously.

go/vt/vttablet/tabletserver/tabletserver.go
mattlord: Correct, it will now redact the bind variables if TerseErrors is true. But I would say that this old behavior was a bug, no? The whole point of the flag was to redact bind variables in client error messages, wasn't it?

mattlord: It seems like the assumption made in the previous code as that if we're here, either TerseErrors is OFF *or* we have 0 bind variables so redaction does not matter.

go/vt/vttablet/tabletserver/tabletserver.go
mattlord: This is logging related, which it really shouldn't be touching anyway. And it was a place where the two flags overlapped directly so I don't think we can avoid it here. TerseErrors should be about client error messages -- it wasn't redacting logs anyway.

mattlord:
> I am not sure whether these differences are of any importance, since the primary reason for not touching errors with `FAILED_PRECONDITION` was so that vtgate could stop buffering when it saw these errors. Since, buffering tests passed, it might be okay. Also, I don't know whether it is even possible for us to get a `FAILED_PRECONDITION` error for a query with bindVariables. So, it might also be a case that we don't care about. @deepthi @mattlord WDYT?

Thank you for the additional look, @GuptaManan100 ! I replied to your comments. Hopefully they make sense. IMO, we should leave things as they are. I'm happy to make changes if you both disagree though.

go/vt/vttablet/tabletserver/tabletserver.go
mattlord: After having read that comment... about why the `FAILED_PRECONDITION` handling was in place I agree with you. We need to change this behavior back.

mattlord: This is addressed here: https://github.com/vitessio/vitess/pull/9634

GuptaManan100:
I agree with @mattlord, Terse Errors should not be changing logs. https://github.com/vitessio/vitess/pull/3737/ is the one which started (once again) readacting bind variables from logs. I think we can treat that as a bug, especially because the description says that it doesn’t intend to change the behaviour of TerseErrors. https://github.com/vitessio/vitess/pull/9094 has never been in a GA release (Thankyou @mattlord for noticing that) so it should be okay to revert that. We will be in a cleaner state where TerseErrors only changes errors and SanitizeLogMessages changes logs. We should document this change in release notes of release 13 however.

go/vt/vttablet/tabletserver/tabletserver.go
GuptaManan100: I agree 💯 %

PR #9551, opened 2022-01-22
Move vtorc back to normal runners

GuptaManan100:
## Description

The flakiness in the vtorc test suite is resolved now with no failing runs in the past 2 days. So this PR moves the vtorc test back to normal runners. Also, we document the bug fix that was run on the self-hosted runners to increase the Linux max concurrent I/O limit.

## Checklist


rohit-nayak-ps (approved):
lgtm

PR #9552, opened 2022-01-22
DO NOT MERGE - Semi sync durability policy final check

GuptaManan100:
## Checklist


GuptaManan100:
This PR was only created to check whether the end to end tests keep working even when we start using the durability policy information in `fixSemiSync`. We only expect the tests that run `Backup` RPC and the upgrade downgrade reparent test with old vtctl to fail.

PR #9554, opened 2022-01-24
non-unique vindex routing in update and delete query

harshit-gangal:
## Description

This PR adds supported for extracting routing information from update/delete query and using non-unique lookup vindex if available and having lowest cost among other options.

## Checklist


GuptaManan100 (approved):
Everything looks good to me! Is there a way to add an end to end test for this? I know it is hard since we won't know which vindex was used for routing, but we could do something like try deleting and check which all shards received that query

frouioui (approved):
Looks good to me

harshit-gangal:
> Everything looks good to me! Is there a way to add an end to end test for this? I know it is hard since we won't know which vindex was used for routing, but we could do something like try deleting and check which all shards received that query

already on it.

PR #9555, opened 2022-01-24
sqlparser: add a proper type for decimal literals

vmg:
## Description

For a long time, the tokenizer in our SQL Parser was playing fast and loose with decimal types in SQL queries: any number that had decimal numbers after a point was considered a "FLOAT". This does not match MySQL's behavior, which actually distinguishes between "DECIMAL" and "FLOAT" literals (a literal must contain an `e` followed by an exponent to be considered a "FLOAT", otherwise it is interpreted as an arbitrary precision decimal).

This PR fixes this behavior so it more closely matches MySQL's behavior and cleans up the `evalengine` so it no longer requires a workaround to tell between these two primitive literal types while evaluating expressions.

Most importantly, **this PR also fixes a serious bug in `handleUnaryMinus`** in the `sqlparser` that has existed for many Vitess releases: the SQL parser was coercing negative floating point numbers into integers _at parse time_. This was spotted by the `evalengine` improvements I've been implementing. It's not clear what would be the actual impact of this bug.

## Checklist


go/vt/sqlparser/ast_funcs.go
```diff
@@ -1447,13 +1452,14 @@ const (
...
-	if num, ok := expr.(*Literal); ok && (num.Type == IntVal || num.Type == FloatVal) {
+	if num, ok := expr.(*Literal); ok && (num.Type == IntVal || num.Type == FloatVal || num.Type == DecimalVal) {
 		// Handle double negative
 		if num.Val[0] == '-' {
 			num.Val = num.Val[1:]
 			return num
 		}
-		return NewIntLiteral("-" + num.Val)
+		num.Val = "-" + num.Val
+		return num
```
vmg, line 1462: cc @systay  -- this bug has been present for ages. Whenever we saw a literal float preceded by unary minus, this hardcoded call to `NewIntLiteral` would emit it with an integer type, even though the actual value of the bytes for the literal would still be a floating point number.

systay: :facepalm:

systay (approved):

frouioui (approved):
LGTM!

PR #9558, opened 2022-01-25
mysql: use UTF8mb4 as consistent connection charset

vmg:
## Description

OK, after a lot of testing I have what I believe is a working solution for the weird collation behavior we've been seeing in Vitess clusters. What I figured out is that the workaround that we implemented in our MySQL connection code to support collations >255 was not really working. As a reminder, when connecting to a MySQL server, a handshake is performed between client and server to agree on a "connection charset" (as defined in the MySQL documentation). Now, two critical facts about this: the wire protocol only has 1 byte for this charset, so it's not possible to use charsets higher than 255; and what the documentation calls "charset" is in fact the ID of a full collation, that can very well go above 255.

This PR hence does the following things:

1. It removes the `db_collation` flag that @frouioui and I introduced a couple months ago. Since this flag never made it to any Vitess releases, there's no deprecation notice needed. As a result of this investigation, we now understand that setting a "charset" and a "collation" for a connection is redundant, as the charset is actually a collation ID. The workaround we had to enforce a custom collation, which was doing it via SQL once the connection was open, is not giving reliable results: we were performing a `SET collation_connection` call, but the semantics of this change are _weird_ and it appears to only apply to some string comparisons in SQL. It's not a wholistic change like what we'd get by passing a charset via the handshake when connecting.

2. It adds support for passing collation names to the `db_charset` flag **as long as those collations have a collation ID <=255**, so that it fits in the handshake packet. This is a non-breaking change that matches the behavior of the commandline flags to `mysql` (the MySQL CLI client). You can pass a charset name like we were doing before, such as `utf8mb4`, but you can also pass a full collation like `utf8m4_general_ci`. Vitess will error out if you attempt to pass a collation that resolves to an ID > 255. Fortunately, all the collation IDs for the default collations for all charsets in all MySQL versions are <=255.

3. It reliably changes the default charset used for Vitess clusters **based on the underlying MySQL server version**. Vitess clusters based on MySQL 5.x will use `utf8mb4_general_ci` by default (unless configured otherwise with the `db_charset` flag). Vitess clusters based on MySQL 8+ will use `utf8mb4_0900_ai_ci`. To note: `utf8mb4_0900_ai_ci` _is_ the default charset for a MySQL 8.0 deployment, and Vitess will not match that behavior. However, `utf8mb4_general_ci` is _not_ the default charset for a MySQL 5.7 deployment: that'd be `utf8_general_ci`. We believe this is a terrible default, particularly since `utf8` is a deprecated charset in MySQL, so we want to enforce all users to use `utf8mb4` charsets even in 5.7 to prevent them from shooting themselves in the foot.

## Checklist

### DIFF
--- a/go/mysql/collations/env.go
+++ b/go/mysql/collations/env.go
@@ -221,12 +221,6 @@ func Local() *Environment {
 	return defaultEnv
 }
 
-// DefaultConnectionCharset is the default 1-byte charset flag that will be sent in all
-// MySQL clients created by Vitess. This is hardcoded to `utf8mb4_0900_ai_ci` because
-// it's the default in MySQL 8.0+. Older versions of MySQL that do not support this
-// charset will automatically fall back to their built-in default.
-const DefaultConnectionCharset = CollationUtf8mb4ID
-
 // A few interesting character set values.
 // See http://dev.mysql.com/doc/internals/en/character-set.html#packet-Protocol::CharacterSet
 const (
@@ -235,25 +229,29 @@ const (
 	CollationBinaryID  = 63
 )
 
-func ParseConnectionCharset(csname string) (uint8, error) {
-	csname = strings.ToLower(csname)
-	switch csname {
-	case "utf8mb4":
-		return CollationUtf8mb4ID, nil
-	case "utf8":
-		return CollationUtf8ID, nil
-	case "latin1":
-		return 8, nil
-	case "":
-		return DefaultConnectionCharset, nil
+func (env *Environment) DefaultConnectionCharset() uint8 {
+	switch env.version {
+	case collverMySQL80:
+		return CollationUtf8mb4ID
 	default:
-		for id, meta := range globalVersionInfo {
-			for _, alias := range meta.alias {
-				if alias == csname && id < 256 {
-					return uint8(id), nil
-				}
-			}
-		}
+		return 45
+	}
+}
+
+func (env *Environment) ParseConnectionCharset(csname string) (uint8, error) {
+	if csname == "" {
+		return env.DefaultConnectionCharset(), nil
+	}
+
+	var collid ID = 0
+	csname = strings.ToLower(csname)
+	if defaults, ok := env.byCharset[csname]; ok {
+		collid = defaults.Default.ID()
+	} else if coll, ok := env.byName[csname]; ok {
+		collid = coll.ID()
+	}
+	if collid == 0 || collid > 255 {
 		return 0, fmt.Errorf("unsupported connection charset: %q", csname)
 	}
+	return uint8(collid), nil
 }
--- a/go/mysql/collations/env.go
+++ b/go/mysql/collations/env.go
@@ -236,7 +236,8 @@ const (
 )
 
 func ParseConnectionCharset(csname string) (uint8, error) {
-	switch strings.ToLower(csname) {
+	csname = strings.ToLower(csname)
+	switch csname {
 	case "utf8mb4":
 		return CollationUtf8mb4ID, nil
 	case "utf8":
@@ -246,6 +247,13 @@ func ParseConnectionCharset(csname string) (uint8, error) {
 	case "":
 		return DefaultConnectionCharset, nil
 	default:
+		for id, meta := range globalVersionInfo {
+			for _, alias := range meta.alias {
+				if alias == csname && id < 256 {
+					return uint8(id), nil
+				}
+			}
+		}
 		return 0, fmt.Errorf("unsupported connection charset: %q", csname)
 	}
 }
--- a/go/mysql/collations/env.go
+++ b/go/mysql/collations/env.go
@@ -229,6 +229,11 @@ const (
 	CollationBinaryID  = 63
 )
 
+// DefaultConnectionCharset is the default charset that Vitess will use when negotiating a
+// charset in a MySQL connection handshake. Note that in this context, a 'charset' is equivalent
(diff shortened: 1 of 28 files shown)


go/mysql/server.go
```diff
@@ -591,7 +592,7 @@ func (c *Conn) writeHandshakeV10(serverVersion string, authServer AuthServer, en
 	pos = writeUint16(data, pos, uint16(capabilities))
 
 	// Character set.
-	pos = writeByte(data, pos, CharacterSetUtf8)
+	pos = writeByte(data, pos, CharacterSetUtf8mb4)
```
frouioui: `CharacterSetUtf8mb4` is set to `45` which is `uf8mb4_general_ci`, shouldn't we change that constant to use `utf8mb4_0900_ai_ci` (use `collations.DefaultConnectionCharset`)?

frouioui: It got resolved via https://github.com/vitessio/vitess/commit/8b26ffe088966a8e8cfe0c1fd1b1b22ed5f21940 :)

go/mysql/collations/env.go
```diff
@@ -246,3 +220,40 @@ func Local() *Environment {
...
+	default:
+		for id, meta := range globalVersionInfo {
+			for _, alias := range meta.alias {
+				if alias == csname && id < 256 {
+					return uint8(id), nil
+				}
+			}
+		}
+		return 0, fmt.Errorf("unsupported connection charset: %q", csname)
+	}
```
frouioui, line 258: If I understand correctly, Vitess will not support charsets with an ID > 256, for instance, `utf8mb4_es_0900_ai_ci` will not be usable by someone who explicitly sets the `collation` flags on both VTGate and VTTablet.
The fact that we are removing the `set collation_connection` post-handshake kind of bugs me. 
I understand we cannot use such ID for the handshake since we only have 1 byte available. However, to support collations with an ID greater than 256 could we reestablish the `set collation_connection`, or are we dropping support for them?

vmg:
OK, after a lot of testing I have what I believe is a working solution. What I figured out is that the workaround that we implemented in our MySQL connection code to support collations >255 was not really working. As a reminder, when connecting to a MySQL server, a handshake is performed between client and server to agree on a "connection charset" (as defined in the MySQL documentation). Now, two critical facts about this: the wire protocol only has 1 byte for this charset, so it's not possible to use charsets higher than 255; and what the documentation calls "charset" is in fact the ID of a full collation, that can very well go above 255.

This PR hence does the following things:

1. It removes the `db_collation` flag that @frouioui and I introduced a couple months ago. Since this flag never made it to any Vitess releases, there's no deprecation notice needed. As a result of this investigation, we now understand that setting a "charset" and a "collation" for a connection is redundant, as the charset is actually a collation ID. The workaround we had to enforce a custom collation, which was doing it via SQL once the connection was open, is not giving reliable results: we were performing a `SET collation_connection` call, but the semantics of this change are _weird_ and it appears to only apply to some string comparisons in SQL. It's not a wholistic change like what we'd get by passing a charset via the handshake when connecting.

2. It adds support for passing collation names to the `db_charset` flag **as long as those collations have a collation ID <=255**, so that it fits in the handshake packet. This is a non-breaking change that matches the behavior of the commandline flags to `mysql` (the MySQL CLI client). You can pass a charset name like we were doing before, such as `utf8mb4`, but you can also pass a full collation like `utf8m4_general_ci`. Vitess will error out if you attempt to pass a collation that resolves to an ID > 255. Fortunately, all the collation IDs for the default collations for all charsets in all MySQL versions are <=255.

3. It reliably changes the default charset used for Vitess clusters **based on the underlying MySQL server version**. Vitess clusters based on MySQL 5.x will use `utf8mb4_general_ci` by default (unless configured otherwise with the `db_charset` flag). Vitess clusters based on MySQL 8+ will use `utf8mb4_0900_ai_ci`. To note: `utf8mb4_0900_ai_ci` _is_ the default charset for a MySQL 8.0 deployment, and Vitess will not match that behavior. However, `utf8mb4_general_ci` is _not_ the default charset for a MySQL 5.7 deployment: that'd be `utf8_general_ci`. We believe this is a terrible default, particularly since `utf8` is a deprecated charset in MySQL, so we want to enforce all users to use `utf8mb4` charsets even in 5.7 to prevent them from shooting themselves in the foot.

vmg:
@frouioui green and ready for re-review

go/mysql/collations/env.go
```diff
@@ -246,3 +220,38 @@ func Local() *Environment {
...
+func (env *Environment) DefaultConnectionCharset() uint8 {
+	switch env.version {
+	case collverMySQL80:
+		return CollationUtf8mb4ID
+	default:
+		return 45
+	}
+}
+
+func (env *Environment) ParseConnectionCharset(csname string) (uint8, error) {
```
frouioui, line 241: nit: add comment for exported method

go/mysql/collations/env.go
```diff
@@ -246,3 +220,38 @@ func Local() *Environment {
...
+
+// A few interesting character set values.
+// See http://dev.mysql.com/doc/internals/en/character-set.html#packet-Protocol::CharacterSet
+const (
+	CollationUtf8ID    = 33
+	CollationUtf8mb4ID = 255
+	CollationBinaryID  = 63
+)
+
+func (env *Environment) DefaultConnectionCharset() uint8 {
```
frouioui, line 232: nit: add comment for exported method

frouioui (approved):
This is great! Looking good to me 🚀

PR #9561, opened 2022-01-25
docker: k8s release process updates

derekperkins:
This is an update to the release process for the `/docker/k8s/...` images for individual binaries. This should all be pushed into Docker Hub configuration, and then `release.sh` can be removed I just haven't had the chance to do it. I still manually build and publish images from my machine for every release.

### DIFF
--- a/docker/k8s/Dockerfile
+++ b/docker/k8s/Dockerfile
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 ARG VT_BASE_VER=latest
-ARG DEBIAN_VER=buster-slim
+ARG DEBIAN_VER=stable-slim
 
 FROM vitess/base:${VT_BASE_VER} AS base
 
--- a/docker/k8s/Dockerfile
+++ b/docker/k8s/Dockerfile
@@ -13,10 +13,11 @@
 # limitations under the License.
 
 ARG VT_BASE_VER=latest
+ARG DEBIAN_VER=buster-slim
 
 FROM vitess/base:${VT_BASE_VER} AS base
 
-FROM debian:buster-slim
+FROM debian:${DEBIAN_VER}
 
 # TODO: remove when https://github.com/vitessio/vitess/issues/3553 is fixed
 RUN apt-get update && \
--- a/docker/release.sh
+++ b/docker/release.sh
@@ -14,55 +14,55 @@ do
   docker build --build-arg VT_BASE_VER=$vt_base_version --build-arg DEBIAN_VER=$debian_version-slim -t vitess/k8s:$vt_base_version-$debian_version .
   docker tag vitess/k8s:$vt_base_version-$debian_version vitess/k8s:$vt_base_version
   docker push vitess/k8s:$vt_base_version-$debian_version
-  if [ $debian_version = $default_debian_version ]; then docker push vitess/k8s:$vt_base_version; fi
+  if [[ $debian_version == $default_debian_version ]]; then docker push vitess/k8s:$vt_base_version; fi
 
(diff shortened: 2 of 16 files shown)


derekperkins:
None of the failing tests are related to these changes

docker/k8s/Dockerfile
```diff
@@ -13,10 +13,11 @@
 # limitations under the License.
 
 ARG VT_BASE_VER=latest
+ARG DEBIAN_VER=buster-slim
```
mattlord, line 16: We're doing vitess latest, I wonder if it's worth using Debian stable (buster is currently oldstable)?

docker/release.sh
```diff
@@ -1,73 +1,68 @@
...
+  echo "####### Building vitess/vt:$debian_version"
 
-docker build --build-arg VT_BASE_VER=$vt_base_version -t vitess/vtgate:$vt_base_version-buster vtgate
-docker tag vitess/vtgate:$vt_base_version-buster vitess/vtgate:$vt_base_version
-docker push vitess/vtgate:$vt_base_version-buster
-docker push vitess/vtgate:$vt_base_version
+  docker build --build-arg VT_BASE_VER=$vt_base_version --build-arg DEBIAN_VER=$debian_version-slim -t vitess/k8s:$vt_base_version-$debian_version .
+  docker tag vitess/k8s:$vt_base_version-$debian_version vitess/k8s:$vt_base_version
+  docker push vitess/k8s:$vt_base_version-$debian_version
+  if [ $debian_version = $default_debian_version ]; then docker push vitess/k8s:$vt_base_version; fi
```
mattlord, line 17: I think that we should do this in one of two ways:
```
 if [ "$debian_version" = "$default_debian_version" ];
```
```
 if [[ $debian_version == $default_debian_version ]];
```

Single bracket is posix compliant, but requires quoting values for string comparisons. Double bracket allows for more natural operations. Double == is more explicit IMO as it's clearly full equality comparison rather than assignment, but either works with double brackets.

mattlord (changes requested):
We'll need to tweak the equality conditions but otherwise it seems fine. I'm not familiar with the context here though like you are. Is there anything related to this that we should do as part of the upcoming 13.0-RC and/or 13.0 (GA) releases?

Thank you!

docker/release.sh
derekperkins: I went with #2 - I trust your opinion :)

docker/k8s/Dockerfile
derekperkins: Done

derekperkins:
Thanks for the review. The only thought I have about v13 is about the default debian version. I'm publishing both `buster` and `bullseye` specific images, but the unspecified tag `v12.0.3` currently matches `buster`. I was thinking about switching that to `bullseye` for v13. 

IIRC, nothing in official Vitess documentation uses these anymore, but anyone running from the original helm charts would use these. Docker Hub shows a few hundred thousand downloads on average for most of these, but I'm not sure how active they are recently. FWIW, I didn't publish the log4j images until I submitted this PR, so nobody was beating down the door for the updates.

image

mattlord:
> Thanks for the review. The only thought I have about v13 is about the default debian version. I'm publishing both `buster` and `bullseye` specific images, but the unspecified tag `v12.0.3` currently matches `buster`. I was thinking about switching that to `bullseye` for v13.

I would default to pairing latest GA with latest GA here. So Vitess 13.0 on bullseye.

> IIRC, nothing in official Vitess documentation uses these anymore, but anyone running from the original helm charts would use these. Docker Hub shows a few hundred thousand downloads on average for most of these, but I'm not sure how active they are recently. FWIW, I didn't publish the log4j images until I submitted this PR, so nobody was beating down the door for the updates.

Those have been deprecated in previous releases and entirely removed in v13 AFAIK. So I think we're OK there. 

THANK YOU!

mattlord (approved):

PR #9563, opened 2022-01-26
Escape the DROP TABLE statements in TrafficSwitcher

mattlord:
## Description
This is a follow-up to https://github.com/vitessio/vitess/pull/9522

We missed escaping the final `DROP TABLE` statements and thus `MoveTables {Cancel,Complete}` can still fail when tables that were moved, which have special characters in them or use reserved words, are cleaned up.

Adding a table with special characters `Lead-1` to the vreplication e2e tests exposed another bug: while generating reverse streams (while switching writes) the table name in the BinlogSource rule filter was not being escaped causing failures in reverse replication and vdiffs of reverse workflows. Fixed in this PR.

## Related Issue(s)
Follow-up to: https://github.com/vitessio/vitess/pull/9522

## Checklist
- [x] Tests have been updated


rohit-nayak-ps (approved):
lgtm

go/vt/wrangler/traffic_switcher.go
```diff
@@ -1168,7 +1168,7 @@ func (ts *trafficSwitcher) createReverseVReplication(ctx context.Context) error
 					// We currently assume the primary vindex is the best way to filter, which may not be true.
 					inKeyrange = fmt.Sprintf(" where in_keyrange(%s, '%s.%s', '%s')", sqlparser.String(vtable.ColumnVindexes[0].Columns[0]), ts.SourceKeyspaceName(), vtable.ColumnVindexes[0].Name, key.KeyRangeString(source.GetShard().KeyRange))
```
mattlord, line 1169: We don't need to escape these too?

mattlord, line 1169: We don't need to escape these too?

go/vt/wrangler/traffic_switcher_test.go
```diff
@@ -822,17 +822,17 @@ func testTableMigrateOneToMany(t *testing.T, keepData, keepRoutingRules bool) {
...
 		tme.dbSourceClients[0].addQuery("delete from _vt.copy_state where vrepl_id in (3, 4)", &sqltypes.Result{}, nil)
 	}
 
 	createReverseVReplication := func() {
-		deleteReverseReplicaion()
+		deleteReverseReplication()
 
-		tme.dbSourceClients[0].addQueryRE(`insert into _vt.vreplication.*test_reverse.*ks2.*-80.*t1.*from t1\\".*t2.*from t2\\"`, &sqltypes.Result{InsertID: 1}, nil)
-		tme.dbSourceClients[0].addQueryRE(`insert into _vt.vreplication.*test_reverse.*ks2.*80-.*t1.*from t1\\".*t2.*from t2\\"`, &sqltypes.Result{InsertID: 2}, nil)
+		tme.dbSourceClients[0].addQueryRE(`insert into _vt.vreplication.*test_reverse.*ks2.*-80.*t1.*from `+"`"+"t1`"+`\\".*t2.*from `+"`"+"t2`"+`\\"`, &sqltypes.Result{InsertID: 1}, nil)
```
mattlord, line 834: Not an issue functionally, but curious why we did +"`"+"t1`"+ instead of +"`t1`"+. Probably just an oversight but wanted to be sure.

mattlord (commented):
I had one question about whether or not we also need to escape the identifiers in the `in_keyrange()` SQL function. Otherwise, it LGTM. Thanks, @rohit-nayak-ps !

go/vt/wrangler/traffic_switcher.go
rohit-nayak-ps: `in_keyrange` is analyzed in golang and not seen by mysql, so I believe we are good here.

go/vt/wrangler/traffic_switcher_test.go
rohit-nayak-ps: Yes, oversight it should be "\`t1\`"

PR #9564, opened 2022-01-26
Changed to assert and require for tests

arvind-murty:
# Description

Changed `t.Errorf` and `t.Fatalf` statements to `assert`, `require`, and so on.
Also fixed a linter error in `conn.go`.

## Checklist


vmg (approved):
Seems fine. Re-running the tests...

PR #9569, opened 2022-01-27
Add 30 second wait timeout in mysql stop for cluster Teardown

GuptaManan100:
## Description

The 5 minute timeout seen in #9507 is also affecting the Teardown method of the cluster. The same fix has been applied to this function to reduce the CI flakiness.

## Checklist

### DIFF
--- a/go/test/endtoend/cluster/cluster_process.go
+++ b/go/test/endtoend/cluster/cluster_process.go
@@ -28,6 +28,7 @@ import (
 	"path"
 	"regexp"
 	"strconv"
+	"strings"
 	"sync"
 	"syscall"
 	"time"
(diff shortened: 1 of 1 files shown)


go/test/endtoend/cluster/cluster_process.go
```diff
@@ -774,6 +777,49 @@ func (cluster *LocalProcessCluster) Teardown() {
...
+				log.Errorf("Error in mysqlctl teardown wait: %v", err)
+				break
+			}
+			pidFile := path.Join(os.Getenv("VTDATAROOT"), fmt.Sprintf("/vt_%010d/mysql.pid", tabletUID))
+			pidBytes, err := os.ReadFile(pidFile)
+			if err != nil {
+				// We can't read the file which means the PID file does not exist
+				// The server must have stopped
+				return
+			}
```
frouioui, line 808: It's maybe not super important, but it would be nice to log something here if we are unable to read the file. At least log the file path and the error

frouioui (approved):
Looks good to me!

go/test/endtoend/cluster/cluster_process.go
GuptaManan100: We don't need that error, since it only errors out when the file doesn't exist and that is a valid scenario. Let's say that mysql shuts down after we have run out of 30 seconds and the pid file disappears. In this case we don't want to log any errors.

PR #9572, opened 2022-01-27
vtctl ApplySchema: introduce -migration_context flag, synonym to -request_context

shlomi-noach:
## Description

Today, `vtctl ApplySchema` accepts a `-request_context` flag. We really call it the migration context. In `SHOW VITESS_MIGRATIONS ...` output it shows as `migration_context`.

So in this PR, `vtctl ApplySchema` now supports a `-migration_context` flag. The two are synonymous so this change is backwards compatible.

Also, we've renamed all variables/fields internally from `requestContext` to `migrationContext` and from `RequestContext` to `MigrationContext`. Thankfully none of these were in any way exported to JSON or to database by implicit name; we always aliased them properly, so this code change has no effect on existing data, and an upgrade shouldn't break anything.

## Related Issue(s)

- Tracking: #6926 

## Checklist


deepthi (approved):
Is there a plan to phase out `-request_context`?

shlomi-noach:
> Is there a plan to phase out -request_context?

Good question. I didn't make a plan. We can phase it out at leisure, the variable isn't really any burden, just a couple extra lines of code.

PR #9575, opened 2022-01-28
examples: how to use mysql80 images with operator

deepthi:
## Description

More detailed example of how to use mysql80 images

## Related Issue(s)

None

## Checklist


frouioui (approved):
That's a good adopter experience improvement! Thanks

Issue #9577, opened 2022-01-28
Nil pointer dereference in tabletserver VStreamRows code path


mvh-stripe:
#### Overview of the Issue

We hit a nil pointer dereference when trying to do a shard-split (see below for logs + stack trace). 

Here's what I think is happening:
 * In the source shard, there's a `_vt_PURGE` temp table which seems to get copied over to the target shards during schema copy.
 * When the target shards issue a `VStreamRows` request to the source shard for this table, it ends up calling into `localVSchema.FindColVindex(tablename)`. But since this temp table has no vschema associated with it, `localVSchema.findTable()` returns `(nil, nil)` and this results in a panic. See https://github.com/vitessio/vitess/blob/main/go/vt/vttablet/tabletserver/vstreamer/local_vschema.go#L36

```
[2022-01-27 20:54:10.563457] I0127 20:54:10.563318 2997134 engine.go:237] Streaming rows for query select * from _vt_PURGE_7decd7b1794811eca1f506784fed54f7_20220120165616 where in_keyrange('1780-1788'), lastpk: []
[2022-01-27 20:54:10.563655] I0127 20:54:10.563485 2997134 local_vschema.go:76] found internal table _vt_PURGE_7decd7b1794811eca1f506784fed54f7_20220120165616, ignoring in local vschema search

stream ID: 1, tabletAlias: northwest-0000107249. vttablet: rpc error: code = Unknown desc = uncaught panic: runtime error: invalid memory address or nil pointer dereference
. Stack-trace:
runtime/panic.go:212 (0x43619a)
runtime/signal_unix.go:734 (0x450092)
vitess.io/vitess/go/vt/vtgate/vindexes/vschema.go:634 (0xde1c89)
vitess.io/vitess/go/vt/vttablet/tabletserver/vstreamer/local_vschema.go:41 (0x135f684)
vitess.io/vitess/go/vt/vttablet/tabletserver/vstreamer/planbuilder.go:662 (0x1365fe4)
vitess.io/vitess/go/vt/vttablet/tabletserver/vstreamer/planbuilder.go:522 (0x1363a5a)
vitess.io/vitess/go/vt/vttablet/tabletserver/vstreamer/planbuilder.go:417 (0x136273e)
vitess.io/vitess/go/vt/vttablet/tabletserver/vstreamer/rowstreamer.go:135 (0x1367ba4)
[...]

PR #9578, opened 2022-01-28
Ignore internal tables in MoveTables and Reshard Workflows

mattlord:
## Description
We need to ignore internal operation tables (e.g. for OnlineDDL processing) in `MoveTables` and `Reshard` workflows. We need to do this both when processing the `-tables` command-line flags and when processing `SchemaDefinition` structures that we get from the database.

I also got rid of the logging messages for when these internal tables are seen and ignored/skipped as I don't think that they are helpful to the end user.

<details>
<summary>You can see a simple Reshard demonstration and test here</summary>

These are using the [docker_local container](https://vitess.io/docs/13.0/get-started/local-docker/).

### Before (main):
```
$ make docker_local && ./docker/local/run.sh
./201_customer_tablets.sh ; ./202_move_tables.sh ; sleep 10; ./203_switch_reads.sh ; ./204_switch_writes.sh ; ./205_clean_commerce.sh ; ./301_customer_sharded.sh ; ./302_new_shards.sh; sleep 5; mysql customer -e "create table _vt_PURGE_1f9194b43b2011eb8a0104ed332e05c2_20201210194431(id int, val varbinary(128), primary key(id))"; ./303_reshard.sh ; sleep 5; mysql customer/-80 -e "show tables"
...
+-----------------------------------------------------------+
| Tables_in_vt_customer                                     |
+-----------------------------------------------------------+
| _vt_PURGE_1f9194b43b2011eb8a0104ed332e05c2_20201210194431 |
| corder                                                    |
| customer                                                  |
+-----------------------------------------------------------+
```

### After (this branch):
```
$ make docker_local && ./docker/local/run.sh
./201_customer_tablets.sh ; ./202_move_tables.sh ; sleep 10; ./203_switch_reads.sh ; ./204_switch_writes.sh ; ./205_clean_commerce.sh ; ./301_customer_sharded.sh ; ./302_new_shards.sh; sleep 5; mysql customer -e "create table _vt_PURGE_1f9194b43b2011eb8a0104ed332e05c2_20201210194431(id int, val varbinary(128), primary key(id))"; ./303_reshard.sh ; sleep 5; mysql customer/-80 -e "show tables"
...
+-----------------------+
| Tables_in_vt_customer |
+-----------------------+
| corder                |
| customer              |
+-----------------------+
```
</details>

## Related Issue(s)
Follow-up to: https://github.com/vitessio/vitess/pull/8992
Fixes: https://github.com/vitessio/vitess/issues/9577

## Checklist
- [x] Tests were added
- [x] Documentation is not required (users should not even be aware of these tables)

### DIFF
--- a/go/test/endtoend/vreplication/helper.go
+++ b/go/test/endtoend/vreplication/helper.go
@@ -100,11 +100,9 @@ func verifyNoInternalTables(t *testing.T, conn *mysql.Conn, database string) {
 	require.NotNil(t, qr.Rows)
 	found := false
 	for _, row := range qr.Rows {
-		t.Logf("Found table: %s", row[0].ToString())
 		if schema.IsInternalOperationTableName(row[0].ToString()) {
 			t.Logf("Found internal table in keyspace: %s! Table name: %s", database, row[0].ToString())
 			found = true
-			break
 		}
 	}
 	require.Equal(t, false, found)
--- a/go/test/endtoend/vreplication/helper.go
+++ b/go/test/endtoend/vreplication/helper.go
@@ -13,6 +13,7 @@ import (
 	"time"
 
 	"github.com/buger/jsonparser"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
 	"vitess.io/vitess/go/test/endtoend/cluster"
@@ -93,20 +94,16 @@ func waitForQueryToExecute(t *testing.T, conn *mysql.Conn, database string, quer
 	}
 }
 
-func verifyNoInternalTables(t *testing.T, conn *mysql.Conn, database string) {
-	t.Logf("Checking that no internal tables exist in keyspace: %s", database)
-	qr := execVtgateQuery(t, conn, database, "show tables")
+// verifyNoInternalTables can e.g. be used to confirm that no internal tables were
+// copied from a source to a target during a MoveTables or Reshard operation.
+func verifyNoInternalTables(t *testing.T, conn *mysql.Conn, keyspaceShard string) {
+	qr := execVtgateQuery(t, conn, keyspaceShard, "show tables")
 	require.NotNil(t, qr)
 	require.NotNil(t, qr.Rows)
-	found := false
 	for _, row := range qr.Rows {
-		if schema.IsInternalOperationTableName(row[0].ToString()) {
-			t.Logf("Found internal table in keyspace: %s! Table name: %s", database, row[0].ToString())
-			found = true
-		}
+		tableName := row[0].ToString()
+		assert.False(t, schema.IsInternalOperationTableName(tableName), "found internal table %s in shard %s", tableName, keyspaceShard)
 	}
-	require.Equal(t, false, found)
-	t.Logf("Found no internal tables in keyspace: %s!", database)
 }
 
 func validateCount(t *testing.T, conn *mysql.Conn, database string, table string, want int) {
--- a/go/test/endtoend/vreplication/helper.go
+++ b/go/test/endtoend/vreplication/helper.go
@@ -16,6 +16,7 @@ import (
 	"github.com/stretchr/testify/require"
 
 	"vitess.io/vitess/go/test/endtoend/cluster"
+	"vitess.io/vitess/go/vt/schema"
 
 	"github.com/PuerkitoBio/goquery"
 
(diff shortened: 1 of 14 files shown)


mattlord:
@shlomi-noach could you please offer some insights/input here if you have time? You may be aware of potential problems in eliding the internal tables at the `mysqlctl.GetSchema` layer. If so, I'll have to find more locations to explicitly ignore these internal tables (Reshard still fails w/o that and will have to find an alternative fix).

mattlord:
> @shlomi-noach could you please offer some insights/input here if you have time? You may be aware of potential problems in eliding the internal tables at the `mysqlctl.GetSchema` layer. If so, I'll have to find more locations to explicitly ignore these internal tables (Reshard still fails w/o that and will have to find an alternative fix).

The failed tests seem to be telling me it is a problem. :-) I'll abandon that approach, at least for now.

go/vt/mysqlctl/tmutils/schema.go
```diff
@@ -273,9 +275,7 @@ func DiffSchema(leftName string, left *tabletmanagerdatapb.SchemaDefinition, rig
 	for leftIndex < len(left.TableDefinitions) && rightIndex < len(right.TableDefinitions) {
 		// extra table on the left side
 		if left.TableDefinitions[leftIndex].Name < right.TableDefinitions[rightIndex].Name {
-			if schema.IsInternalOperationTableName(left.TableDefinitions[leftIndex].Name) {
-				log.Infof("found internal table %v, skipping in schema diff", left.TableDefinitions[leftIndex].Name)
```
rohit-nayak-ps, line 277: The reason we had added these logs were for visibility in case some table mysteriously failed to get resharded and the user could identify why from the logs. But removing it should be fine.

go/vt/wrangler/materializer.go
```diff
@@ -93,6 +93,26 @@ func (wr *Wrangler) addTablesToVSchema(ctx context.Context, sourceKeyspace strin
 	return nil
 }
 
+func shouldInclude(table string, excludes []string) bool {
+	// We filter out internal tables elsewhere when processing SchemaDefinition
+	// structures built from the GetSchema database related API calls. In this
+	// case, however, the table list comes from the user via the -tables flag
+	// so we need to filter out internal table names here in case a user has
+	// explcility specified some.
```
rohit-nayak-ps, line 101: typo explcility=>explicitly. Maybe fix later, no need for CI to run again for this!

rohit-nayak-ps (approved):
lgtm

go/test/endtoend/vreplication/helper.go
```diff
@@ -92,6 +93,22 @@ func waitForQueryToExecute(t *testing.T, conn *mysql.Conn, database string, quer
 	}
 }
 
+func verifyNoInternalTables(t *testing.T, conn *mysql.Conn, database string) {
+	t.Logf("Checking that no internal tables exist in keyspace: %s", database)
+	qr := execVtgateQuery(t, conn, database, "show tables")
+	require.NotNil(t, qr)
+	require.NotNil(t, qr.Rows)
+	found := false
+	for _, row := range qr.Rows {
```
shlomi-noach, line 102: To gain more visibility into failure, I'd suggest:
```go
   tableName := row[0].ToString()
   assert.False(t, schema.IsInternalOperationTableName(tableName), "found internal table %s", tableName)
```

then, remove the block with `t.Logf` and the `require...` that follows.

mattlord: Good call. I was being noisy during testing to see what was happening, but don't need that in ever CI run going forward.

mattlord:
Thank you, @shlomi-noach and @rohit-nayak-ps! I've addressed those comments here: https://github.com/vitessio/vitess/pull/9578/commits/1101339d4267788583ec9222204867eccfc453cc

shlomi-noach (approved):

go/vt/mysqlctl/tmutils/schema.go
mattlord: Yeah, that was my thinking as well too (I added these specific log messages here). But it ends up as a lot of log messages in total and makes it harder to find the signal in the noise. I feel comfortable removing them as I double checked the REGEXPs and they are very precise:
  - https://github.com/vitessio/vitess/blob/a033d76e25a1ce5332b3956f8db9801f07bccb81/go/vt/schema/tablegc.go#L47
  - https://github.com/vitessio/vitess/blob/a033d76e25a1ce5332b3956f8db9801f07bccb81/go/vt/schema/online_ddl.go#L39

Issue #6407, opened 2020-07-03
RFC: Deprecate ExecuteBatch calls on autocommit mode in VTGate


harshit-gangal:
VTGate currently calls ExecuteBatch when autocommit is true this was due to the reason that earlier VTTablet Execute always create a transaction even on autocommit true.
Now, the underlying VTTablet Execute can handle autocommit mode and therefore it is safe to switch the call from ExecuteBatch to Execute. 
This also simplifies the VTGate execution code path.

PR #9586, opened 2022-01-28
Refactor: Remove unused ExecuteBatch and BeginExecuteBatch from vttablet

harshit-gangal:
## Description

This PR removes unused ExecuteBatch and BeginExecuteBatch. It is not used in queryservice.

## Related Issue(s)

- https://github.com/vitessio/vitess/issues/6407

## Checklist


deepthi (commented):
Even though we no longer call these from vtgate, they are a public API.
I'm not certain that it is safe / ok to delete these without going through a deprecation cycle.

harshit-gangal:
The deprecation was announced in 7.0 release. This is more of code cleanup now.

deepthi:
> The deprecation was announced in 7.0 release. This is more of code cleanup now.

The linked RFC says "deprecate calls in vtgate". Did we find out from the community that no one uses the vttablet query service API directly so it's ok to remove things from there?

GuptaManan100 (approved):
Everything else LGTM, but do we need to go through a deprecation cycle where we log a deprecation warning before removing it from a public API?

GuptaManan100:
Oh I see that was already brought up by @deepthi

harshit-gangal:
> > The deprecation was announced in 7.0 release. This is more of code cleanup now.
> 
> The linked RFC says "deprecate calls in vtgate". Did we find out from the community that no one uses the vttablet query service API directly so it's ok to remove things from there?

That is the general expectation that no one will use queryservice API directly as this would require managing connections and session by the caller. I will post on slack to see if there is any feedback.

deepthi (approved):

Issue #9566, opened 2022-01-26
Release of `v13.0.0 RC1`


frouioui:
## Description

The release of `v13.0.0 RC1` is planned for Tuesday, February 2. This issue contains a checklist of all the things to do for the release:

### Pre-Release

- [x] Make sure the `vitess-operator` repository builds Docker Images on `main`.
- [x] Change `main` to use `latest` Docker tags for Vitess operator example. (`./examples/operator/(101|201|302|306)_*.yaml` and `./examples/operator/operator.yaml` files) (#9582)
- [x] Locally make sure that changing the `vitess-operator` dependencies to use Vitess' `main` is not breaking.
- [x] Make sure outstanding pull requests are reviewed and merged into `main`.
_Only bug fixes can be added to `v13.0.0 GA` once `RC1` is out._
- [x] Notify the community on Slack for the upcoming release.

### Release

- [x] Follow the release instructions for release candidates: [here](https://github.com/vitessio/vitess/blob/main/doc/internal/ReleaseInstructions.md#pre-requisites-for-release-candidates-rc)

- [x] Continue with the Java release following [these](https://github.com/vitessio/vitess/blob/main/doc/internal/ReleaseInstructions.md#java-packages-deploy--release) instructions.
  > I will fill in more details about this step once I have met with @harshit-gangal about the Java releases.

- [x] `v13.0.0 RC1` is released on GitHub and DockerHub 🎉 
- [x] On the Vitess website repository, create a `v14.0` directory and update the sidebar. We should follow what was done in vitessio/website#907.
  Open a Pull Request and merge it.
  > Maintainers have until the release of `GA` to finish the documentation of `v13.0.0` features. New documentation for `v13.0.0` features will need to be added to both the `v13.0` and `v14.0` directories.
  > Note that since we are changing our release cycle to 4 months, we keep `v11.0` which will be deprecated in July 2022.

- [x] Merge the `13.0.0` release notes on `main`
- [x] Update `main` to use `v14.0.0-SNAPSHOT` instead of `v13.0.0-SNAPSHOT`

### Post-Release

- [x] No blog post for release candidates.
- [x] Notify the community on Slack that `RC1` is released.
- [ ] Make sure unfinished documentation gets merged before `GA`.

frouioui:
Closing this issue as all the steps are completed. We will keep monitoring 13's documentation until GA.

PR #9587, opened 2022-01-28
Automatically update vitess-operator example during a release

frouioui:
## Description

This pull request adds some logic to `do_release` to automatically update the vitess-operator example during a release.

In the `./examples/operator`, both the Kubernetes YAML files and the README are updated during the release. The `vitess/lite` Docker image will be tagged to the new release (`$RELEASE_VERSION`), and the `planetscale/vitess-operator` image will use the value of `$VTOP_VERSION` as a tag. Note that `$VTOP_VERSION` is not required to run `do_release`, if no value is passed, the `planetscale/vitess-operator` image will remain the same.

## Related Issue(s)
- Contributes to https://github.com/vitessio/vitess/issues/9566

## Checklist

### DIFF
--- a/tools/do_release.sh
+++ b/tools/do_release.sh
@@ -53,8 +53,9 @@ function updateJava () {
 }
 
 function updateVitessOperatorExample () {
-  vtop_example_files=$(find -E $ROOT/examples/operator -regex ".*.(md|yaml)")
-  sed -i.bak -E "s/vitess\/lite:(.*)(-mysql80)?/vitess\/lite:$RELEASE_VERSION/g" $vtop_example_files
+  vtop_example_files=$(find -E $ROOT/examples/operator -name "*.yaml")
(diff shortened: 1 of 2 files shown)


tools/do_release.sh
```diff
@@ -0,0 +1,108 @@
...
+git_status_output=$(git status --porcelain)
+if [ "$git_status_output" == "" ]; then
+  	echo so much clean
+else
+    echo cannot do release with dirty git state
+    exit 1
+fi
+
+# Preparing the release commit
+updateVitessOperatorExample
```
deepthi, line 74: nit: it seems inconsistent that you don't pass in RELEASE_VERSION here as an argument.

deepthi (approved):

tools/do_release.sh
frouioui: You're right, I changed it via 5a89899

PR #9593, opened 2022-01-28
tests: change the way we call net.Listen to avoid annoying popups on macOS

deepthi:
## Description

Running the unit tests on macOS produces a bunch of popups asking whether .test executables should be allowed to accept incoming network connections.
It is annoying because they steal the focus, and it is not desirable to turn off the firewall to avoid them.
This PR attempts to fix _most_ of these warnings/errors. There are still a few that I couldn't trace to the source and some that I couldn't change without causing tests to fail.

## Checklist

### DIFF
--- a/go/test/fuzzing/tablet_manager_fuzzer.go
+++ b/go/test/fuzzing/tablet_manager_fuzzer.go
@@ -33,7 +33,7 @@ func initTesting() {
 	testing.Init()
 }
 
-func FuzzTabletManager_ExecuteFetchAsDba(data []byte) int {
+func FuzzTabletManagerExecuteFetchAsDba(data []byte) int {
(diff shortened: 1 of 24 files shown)


go/test/fuzzing/tablet_manager_fuzzer.go
```diff
@@ -33,7 +33,7 @@ func initTesting() {
 	testing.Init()
 }
 
-func FuzzTabletManager_ExecuteFetchAsDba(data []byte) int {
+func FuzzTabletManagerExecuteFetchAsDba(data []byte) int {
```
GuptaManan100, line 36: Thankyou for fixing this, I saw this error in fuzzers too, but I had ignored it then!

GuptaManan100 (approved):
LGTM!

PR #9596, opened 2022-01-30
Binlog event parsing: better analysis ; support for semi-sync

shlomi-noach:
## Description

This is a general-purpose enhancement to `Conn`, `binlogEvent` and other low level constructs to:

- Better parse binlog events (more info now analyzed, such as next-file, next-pos, is-emi-sync-ack-requested, ...)
- Support semi-sync events: those are padded with two bytes (1st is magic number, 2nd indicates whether ACK is required or not). See https://dev.mysql.com/doc/internals/en/semi-sync-binlog-event.html
- Support for sending semi-sync ACK. See https://dev.mysql.com/doc/internals/en/semi-sync-ack-packet.html
- Initial support for handling `BinlogDumpGTID` requests

This does not necessarily need to be in release 13. It would be nice to have, but no urgency.

## Related Issue(s)

No specific issue. Just general enhancement.

## Checklist


shlomi-noach:
`onlineddl_vrepl` and `onlineddl_vrepl_stress` consistently fail on this PR (but not on other PRs), so I'm investigating what might have introduced a bug.

shlomi-noach:
Found the issue, and nice that the tests caught it! Turns out you cannot implicitly analyze the packet and decide that it is wrapped with semi-sync info. You have to have prior knowledge that semi-sync info is expected. That because a standard packet can randomly and incidentally have a timestamp that looks exactly like the semi-sync indicator (`0xef`, `0x00|0x01`).

So [Conn](go/mysql/conn.go) now has a stateful `ExpectSemiSyncIndicator bool`. It is the conn's user's responsibility to say whether semi sync is to be expected or not. Basically it's down to this flow the user may run with conn: https://dev.mysql.com/doc/internals/en/semi-sync-replication.html (check if source supports semi sync, `SET @rpl_semi_sync_slave = 1;`).

shlomi-noach:
It's worth noting that this PR has no effect on existing code. The only place where friction is possible is when parsing binlog events, but that requires an explicit instruction by the user, and of course current code has no such instruction. So this PR is mostly additions on top of existing functionality, and these additions are not actually in use by this PR.

go/mysql/binlog_dump.go
```diff
@@ -0,0 +1,43 @@
+package mysql
+
```
frouioui, line 2: nit: we are missing the header here

go/mysql/binlog_event_common.go
```diff
@@ -101,6 +108,11 @@ func (ev binlogEvent) Length() uint32 {
 	return binary.LittleEndian.Uint32(ev.Bytes()[9 : 9+4])
 }
 
+// nextPosition the nextPosition field from the header
```
frouioui, line 111: nit: `nextPosition` -> `NextPosition`

frouioui: actually, you probably meant `NextPosition returns the nextPosition ...`

go/mysql/conn.go
```diff
@@ -918,6 +924,23 @@ func (c *Conn) handleNextCommand(handler Handler) bool {
...
+	c.startWriterBuffering()
+	defer func() {
+		if err := c.endWriterBuffering(); err != nil {
+			log.Errorf("conn %v: flush() failed: %v", c.ID(), err)
+			kontinue = false
+		}
+	}()
+
+	_, _, position, _ := c.parseComBinlogDumpGTID(data)
+	handler.ComBinlogDumpGTID(c, position.GTIDSet)
```
frouioui, line 939: if `c.parseComBinlogDumpGTID` fails shouldn't we not continue? or is it already handled by `endWriterBuffering`?

frouioui (approved):

go/mysql/conn.go
shlomi-noach: good catch!

go/mysql/binlog_event_common.go
shlomi-noach: True!

shlomi-noach:
Review comments addressed

go/mysql/binlog_dump.go
shlomi-noach: fixed

PR #9597, opened 2022-01-31
evalengine: more collation fixes

vmg:
## Description

Some last minute fixes before next release. The goal of this PR is fixing the behavior of the `evalengine` when users launch a Vitess clusters with `-db_charset` to something that is _not_ its default value (i.e. something other than `utf8mb4`). This was not previously supported: the VTGates would evaluate expressions always assuming the default collation for literals that were _not_ explicitly collated to be `utf8mb4_0900_ai_ci`.

To fix this, we need to expand the `ExpressionEnv` so it contains the default collation for the connection we're working with. Fortunately, this is readily available in `VCursor` because @frouioui and I did a great job wiring this up.

## Checklist


vmg:
I've also implemented the builtin SQL function `COLLATION` so we can do integration testing for this feature. :)

frouioui (approved):
👌🏻

Issue #9383, opened 2021-12-14
Deprecate `IgnoreHealthError` at vtctl 


notfelineit:
#### Feature Description
We want to deprecate `IgnoreHealthError` rpc endpoint because it's deprecated at the tablet level:
![image](https://user-images.githubusercontent.com/31225471/146040677-fa77165d-ecd7-4e9e-92c0-2a51a2888fff.png)

We should also remove it from vtctl2 UI if we think that's relevant. 

#### Use Case(s)g)
In the vtctld2 UI and vtctl rpc endpoint:
![image](https://user-images.githubusercontent.com/31225471/146040818-c38b5c59-89e2-4adb-bb60-7575487c548b.png)

deepthi:
Fixed in #9594

Issue #8716, opened 2021-08-27
[vtadmin] Set tablet to ReadWrite


doeg:
# vtctld2 UI

Found at http://localhost:15000/app/shard?keyspace=commerce&shard=0 > tablet menu > "Set ReadWrite"

<img width="1792" alt="Screen Shot 2021-08-27 at 4 07 42 PM" src="https://user-images.githubusercontent.com/855595/131193502-82707a5e-28a9-4286-8eeb-b481d30f17ee.png">

<img width="1792" alt="Screen Shot 2021-08-27 at 4 07 44 PM" src="https://user-images.githubusercontent.com/855595/131193508-422e1437-67e8-46d0-9792-8833dd6e11ae.png">

# VTAdmin Implementation Notes

- Only applies to primary tablets.

Issue #8715, opened 2021-08-27
[vtadmin] Set tablet to ReadOnly


doeg:
# vtctld2 UI

Found at http://localhost:15000/app/shard?keyspace=commerce&shard=0 > tablet menu > "Set ReadOnly"

<img width="1792" alt="Screen Shot 2021-08-27 at 4 07 31 PM" src="https://user-images.githubusercontent.com/855595/131193383-2807f961-0bf2-4075-8ebe-95e65a3541ba.png">

<img width="1792" alt="Screen Shot 2021-08-27 at 4 07 33 PM" src="https://user-images.githubusercontent.com/855595/131193375-195fb043-b71e-473b-9c5c-3743dc580432.png">

# VTAdmin Implementation Notes

- Only applies to primary tablets.

Issue #8710, opened 2021-08-27
[vtadmin] Ignore Health Check for tablets


doeg:
# vtctld2 UI 

<img width="1792" alt="Screen Shot 2021-08-27 at 3 37 37 PM" src="https://user-images.githubusercontent.com/855595/131180101-96d7b68e-3e92-4492-aadb-09acd844524b.png">

<img width="1792" alt="Screen Shot 2021-08-27 at 3 37 33 PM" src="https://user-images.githubusercontent.com/855595/131180097-1d281349-83bc-441f-8d71-f38ec9116754.png">

# VTAdmin Implementation Notes

- The current UI uses both "Ignore Health Check" and "Ignore Health Error" phrasing

notfelineit:
I think we'll be leaving this out of the VTAdmin UI due to https://github.com/vitessio/vitess/issues/9383

deepthi:
we can go ahead and close this.

PR #9601, opened 2022-01-31
[VTAdmin] Tablet Actions (vtctld2 Parity)

notfelineit:
## Description
This PR adds `SetReadOnly` and `SetReadWrite` functionality to **only the primary tablet**, matching the behavior currently in vtctld2 UI.

This PR **does not** add `IgnoreHealthCheckError` since it is deprecated. See https://github.com/vitessio/vitess/issues/9383 for more information.

It also:
- Refactors dangerous actions into a reusable component
- Deletes and removes the 'change dialog' component from the tablets list, since these actions are moved in the 'Advanced' tab

## Manual Test
ezgif com-gif-maker (26)

## Related Issue(s)
- IgnoreHealthCheckError: https://github.com/vitessio/vitess/issues/8710
- SetReadOnly: https://github.com/vitessio/vitess/issues/8715
- SetReadWrite: https://github.com/vitessio/vitess/issues/8716

## Checklist

## Deployment Notes
No deployment notes.


examples/local/101_initial_cluster.sh
```diff
@@ -38,7 +39,7 @@ for i in 100 101 102; do
 done
 
 # set one of the replicas to primary
-vtctldclient PlannedReparentShard commerce/0 --new-primary zone1-100
+vtctldclient PlannedReparentShard commerce/0 --new-primary zone1-100 --server ${hostname}:${vtctld_grpc_port}
```
notfelineit, line 42: As @doeg pointed out, I think this is needed because of https://github.com/vitessio/vitess/pull/9579/files

GuptaManan100: No this change isn't required, since in the beginning of the file we call `source ./env.sh` and it aliases `vtctldclient` -
```
alias vtctldclient="command vtctldclient --server localhost:15999"
```

notfelineit: Ah, I'll remove this.

examples/local/101_initial_cluster.sh
```diff
@@ -18,6 +18,7 @@
 # required for a single shard deployment.
 
 source ./env.sh
+vtctld_grpc_port=15999
```
GuptaManan100, line 21: This change can also be removed, since we aren't using this variable anywhere now

go/vt/vtadmin/api.go
```diff
@@ -1119,6 +1121,72 @@ func (api *API) PingTablet(ctx context.Context, req *vtadminpb.PingTabletRequest
...
+	})
+
+	if err != nil {
+		return nil, fmt.Errorf("Error setting tablet to read-only: %w", err)
+	}
+
+	return &vtadminpb.SetReadOnlyResponse{}, nil
+}
+
+// SetReadOnly sets the tablet to read only mode
```
doeg, line 1157: This comment is from the `SetReadOnly` function :)

doeg (commented):
This looks great! Just one lil comment in addition to @GuptaManan100's and I'll merge this in. 🤠 

It'd be neat (someday) to expose the RO/RW status of a tablet in the UI somehow. 🤔 (I've never actually used these commands prior to now so it was a fun learning experience in using `SELECT @@global.read_only;` lol.)

GuptaManan100:
@doeg, Me and @deepthi  talked about this today morning (yesterday evening for her), that we would want to expose all the metrics that are visible in orchestrator UI like `replication lag`, `replication status`, `executed gtid set`. To do this, we would need to introduce an RPC in TabletManagerCient which collects this information

doeg:
Oh thank you both for the extra context, @notfelineit and @GuptaManan100!! ❤️ That is interesting + good to know. Wow yes an endpoint for that stuff would be soooo nice 😻

notfelineit:
I'm glad we'll have them anyway! Looks like adding vtorc will be a good thing all around 😄  @doeg I'm glad you brought it up in this context

doeg (approved):

PR #9609, opened 2022-02-01
Store innodbRowsRead in a Counter instead of a Gauge

GuptaManan100:
## Description

It makes more sense to store innodbRowsRead in a counter instead of a gauge since Mysql exporter (prometheus) also does so https://github.com/prometheus/mysqld_exporter/blob/c9d7a01501d047ddc3d0c03c33c262e12d28dd59/collector/global_status.go#L163-L166. This PR accomplishes that change.

## Checklist

### DIFF
--- a/go/stats/counter.go
+++ b/go/stats/counter.go
@@ -52,6 +52,11 @@ func (v *Counter) Add(delta int64) {
 	v.i.Add(delta)
 }
 
+// Set overwrites the current value.
+func (v *Counter) Set(value int64) {
+	v.i.Set(value)
+}
+
 // Reset resets the counter value to 0.
 func (v *Counter) Reset() {
 	v.i.Set(int64(0))
--- a/go/vt/vttablet/tabletserver/schema/engine.go
+++ b/go/vt/vttablet/tabletserver/schema/engine.go
@@ -81,7 +81,7 @@ type Engine struct {
 
 	tableFileSizeGauge      *stats.GaugesWithSingleLabel
 	tableAllocatedSizeGauge *stats.GaugesWithSingleLabel
-	innoDbReadRowsGauge     *stats.Gauge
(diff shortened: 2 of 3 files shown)


go/stats/counter.go
```diff
@@ -52,6 +52,11 @@ func (v *Counter) Add(delta int64) {
 	v.i.Add(delta)
 }
 
+// Set overwrites the current value.
+func (v *Counter) Set(value int64) {
+	v.i.Set(value)
+}
+
```
harshit-gangal, line 59: for counter we should call add function, set is used for guage.

deepthi: @GuptaManan100 and I were originally of this opinion as well. Because we are not incrementing this stat, but reading it from MySQL, it only made sense to make it a Gauge. However, Prometheus handles well-known metrics in specific ways. It is doing a substring match for `innodb_rows` and interpreting all of them as counters.
So yeah this is non-intuitive but needed for the metrics from vttablet to work correctly with prom.

deepthi: I suppose we could rename it to something that doesn't match `innodb_rows`.

GuptaManan100:
The major problem is arising from the fact that we are reading this metric from MySQL and setting it in VTTablet.
Since we are setting this metric directly, we don't have the incremental changes, so we aren't *adding* anything, but *setting* the value directly, so it made sense to keep it as a Gauge.
But the value we are reading from MySQL is by definition, incremental and is treated like a counter in MySQL and Prometheus as well.

Since, we want Vitess to be drop in replacement for MySQL, we have to keep it consistent with what it does and make it a counter. I am however going to add a comment there justifying this.

harshit-gangal (approved):

GuptaManan100:
Forgot the comment 🤦, adding it in #9664

PR #9614, opened 2022-02-01
[vtctld] Add 'enable_vtctld_ui' flag to vtctld to explicitly enable (or disable) the vtctld2 UI

### LINKED ISSUE #9194: Add a flag to optionally disable the vtctld2 UI
As the title says. 

This would permit operators to prevent the vtctld2 UI from running in their environment, if they want. Bonus points for preventing it from installing/building at all if the flag is defined. 

The http server is initialized here: https://github.com/vitessio/vitess/blob/3393897/go/vt/vtctld/vtctld.go#L138-L175

doeg:
## Description

This PR closes #9194 by adding a flag to enable/disable the vtctld2 UI. This default value is true, which means that the vtctld2 UI will continue to be enabled by default. In other words, **nothing will change unless operators explicitly opt out**  of the vtctld2 UI. 

With `-enable_vtctld_ui=false`, any requests to any of the following vtctld URLs will 404:
- /app/
- /app/**/*
- Any other unrecognized path, as those all get redirected to /app/ by [this handler](https://github.com/vitessio/vitess/blob/66109e7f240b3fe45e4da2dc0c9db488b14682f9/go/vt/vtctld/vtctld.go#L139-L142)

This flag does not affect the HTTP API that the vtctld serves on /api/ (used by the web app).

### On nomenclature.
I went back and forth for far too long on whether the flag should be `enable_` (and default to true) or `disable_` (and default to false). 

Ultimately, I went with `enable_` given the next step will be to _disable_ the UI by default, meaning this flag will default to `false` and Vitess operators will have to explicitly opt-in to enable the UI. (We can discuss the timing and communication around this next step in the forthcoming RFC for removing the vtctld2 codebase.) 

That said, I'm happy to switch this up if another approach would make more sense. :) 

### What it looks like

An easy way to test this out is to update [examples/local/scripts/vtctld-up.sh](https://github.com/vitessio/vitess/blob/988a839/examples/local/scripts/vtctld-up.sh) with the flag as follows:

```bash
vtctld \
 $TOPOLOGY_FLAGS \
 -enable_vtctld_ui=false \
# ... other vtctld flags go here
```

(Do note there is no space before the `false` value as described in [the flag package](https://pkg.go.dev/flag).)

Then, all requests to http://localhost:15000/app/ (and any subroutes) will 404. 

Screen Shot 2022-02-02 at 12 18 36 PM

One can also verify that /api/ requests still work:

```bash
$ curl http://localhost:15000/api/keyspaces/
[
  "commerce"
]
```

## Related Issue(s)

Closes #9194

## Checklist
- [x] Should this PR be backported? **No,** I think it's fine for this to wait for 14.0 unless folks feel strongly that it should be backported. 

## Deployment Notes

Noted above.

### DIFF
--- a/go/vt/vtctld/vtctld.go
+++ b/go/vt/vtctld/vtctld.go
@@ -141,38 +141,7 @@ func InitVtctld(ts *topo.Server) error {
 	})
 
 	// Serve the static files for the vtctld2 web app
-	http.HandleFunc(appPrefix, func(w http.ResponseWriter, r *http.Request) {
-		// Strip the prefix.
-		parts := strings.SplitN(r.URL.Path, "/", 3)
-		if len(parts) != 3 {
-			http.NotFound(w, r)
-			return
-		}
-		rest := parts[2]
-		if rest == "" {
-			rest = "index.html"
-		}
-
-		riceBox, err := rice.FindBox("../../../web/vtctld2/app")
-		if err != nil {
-			log.Errorf("Unable to open rice box %s", err)
-			http.NotFound(w, r)
-		}
-		fileToServe, err := riceBox.Open(rest)
-		if err != nil {
-			if !strings.ContainsAny(rest, "/.") {
-				//This is a virtual route so pass index.html
-				fileToServe, err = riceBox.Open("index.html")
-			}
-			if err != nil {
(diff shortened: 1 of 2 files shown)


deepthi (approved):
I approve of this change 😄

PR #9615, opened 2022-02-01
demo.go: fix call to ToBytes

pjambet:
The return value was changed from `[]byte` to `([]byte, error)` a little
while ago, making it impossible to run the demo.

This handles the error and returns nil after logging the error in case
the conversion fails.

## Related Issue(s)

- Fixes: https://github.com/vitessio/vitess/issues/9763#issuecomment-1048419739

## Checklist

### DIFF
--- a/examples/demo/demo.go
+++ b/examples/demo/demo.go
@@ -203,7 +203,12 @@ func resultToMap(title string, qr *sqltypes.Result) map[string]interface{} {
 		srow := make([]string, 0, len(row))
 		for _, value := range row {
 			if value.Type() == sqltypes.VarBinary {
-				srow = append(srow, hex.EncodeToString(value.ToBytes()))
+				bytes, err := value.ToBytes()
+				if err != nil {
+					log.Errorf("Error converting value to bytes: %v", err)
+					return nil
+				}
+				srow = append(srow, hex.EncodeToString(bytes))
 			} else {
 				srow = append(srow, value.ToString())
 			}


mattlord:
@pjambet sorry about that! Honestly, I wasn't even aware of this demo app. We must not have any tests for that... 

Please let me know if I can be of any help. We could add a test to catch these breakages going forward.

pjambet:
@mattlord Agreed, having some tests, at least to make sure the code compiles would be great. I found the demo really useful so far so I am biased in wanting to keep it in a clean state!

mattlord:
@pjambet I think simply doing this in our CI may be enough (would currently fail on main):
```
diff --git a/test/config.json b/test/config.json
index 24bcf08c18..b16a554f9c 100644
--- a/test/config.json
+++ b/test/config.json
@@ -1138,6 +1138,19 @@
                        "Shard": "vreplication_migrate",
                        "RetryMax": 1,
                        "Tags": []
-               }
+               },
+                "demo": {
+                        "File": "",
+                        "Args": [],
+                        "Command": [
+                                "go",
+                                "build"
+                                "examples/demo/demo.go"
+                        ],
+                        "Manual": false,
+                        "Shard": "",
+                        "RetryMax": 1,
+                        "Tags": []
+                }
        }
 }
```

```
$ go build ./examples/demo/demo.go
# command-line-arguments
examples/demo/demo.go:206:43: too many arguments in call to hex.EncodeToString
	have ([]byte, error)
	want ([]byte)
```

Do you want to try and add that?

ajm188:
hey @pjambet friendly bump! any update on adding Matt's suggestion? 😊

mattlord:
After thinking about this again, rather than adding a new CI job (which has a fair amount of overhead), we could simply add the demo build to `make unit_test` which seems reasonable to me as it compiling *is* effectively the unit test for the demo 🙂 :
```
diff --git a/Makefile b/Makefile
index 9a85f0e5ee..16a9d5585a 100644
--- a/Makefile
+++ b/Makefile
@@ -132,6 +132,9 @@ vtctldclient: go/vt/proto/vtctlservice/vtctlservice.pb.go
 parser:
        make -C go/vt/sqlparser

+demo:
+       go build ./examples/demo/demo.go
+
 codegen: asthelpergen sizegen parser astfmtgen

 visitor: asthelpergen
@@ -171,7 +174,7 @@ cleanall: clean
        # Remind people to run bootstrap.sh again
        echo "Please run 'make tools' again to setup your environment"

-unit_test: build dependency_check
+unit_test: build dependency_check demo
        echo $$(date): Running unit tests
        tools/unit_test_runner.sh

```

mattlord:
I went ahead and created a separate PR for the test: https://github.com/vitessio/vitess/pull/9766 I'll now get the tests green and approve this PR.

mattlord (approved):
Thank you for the contribution, @pjambet ! ❤️

PR #9618, opened 2022-02-02
[VTAdmin] Simplify JSON Discovery

notfelineit:
## Description
This PR greatly simplifies JSON discovery by:
- Renaming JSONDiscovery to DynamicDiscovery
- Refactoring out key logic in StaticFileDiscovery to JSONDiscovery
- Embedding JSONDiscovery in both StaticFileDiscovery and DynamicDiscovery, since they are both JSON-based

## Related Issue(s)
N/A

## Checklist
- [x] Tests were added or are not required **Tests remain the same for discovery_json_test.go**

## Deployment Notes
N/A

### DIFF
--- a/go/vt/vtadmin/api.go
+++ b/go/vt/vtadmin/api.go
@@ -245,7 +245,7 @@ func (api *API) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 				c, err := cluster.Config{
 					ID:            clusterID,
 					Name:          clusterID,
-					DiscoveryImpl: "json",
+					DiscoveryImpl: "dynamic",
 					DiscoveryFlagsByImpl: cluster.FlagsByImpl{
 						"json": map[string]string{
 							"discovery": string(decoded),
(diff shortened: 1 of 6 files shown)


doeg (approved):
Ohh this is such a satisfying diff! :D 

The only thing I'm not sure about is whether it'd be tidier to extract the common pieces into a separate file + avoid coupling the discovery implementations directly. 🤔 (For example, the `StaticFileDiscovery` that's embedded includes its `parseConfig` file reader function. 

So we could have something like `type DiscoveryConfig struct`, which gets embedded by both `JSONDiscovery` and `StaticFileDiscovery`, which then only need to define (and test) their constructor and/or `parseConfig` functions. 

This is more of a philosophical thing than a functional requirement and I don't feel strongly, so not a blocking comment! Thanks again for doing this. :)

notfelineit:
@doeg I like philosophy! That makes more sense and also puts Json and static file on "equal" footing. I'll go ahead and make that change today 😄

doeg (approved):
🎉

PR #9622, opened 2022-02-03
Update error description for deprecated command

dbussink:
The list command was renamed to show and list was deprecated. This wasn't updated in these error messages yet.

## Checklist


deepthi (approved):

deepthi:
We should get this into release 13.

PR #9627, opened 2022-02-03
Backport: Improve performance of hex type decoding

mattlord:
## Description
We replace the use of RegExps with Byte comparisons which should be much faster.

It's also simpler to read IMO since we're already using Byte slice indexes in related code.

Finally, it's actually more precise as it:
 1. Allows for X'1' and x'1' (MySQL accepts both)
 2. Requires x0\<byte\> as MySQL does not accept simply `0x` (it does
    accept `x''` though)
```
mysql> select @@version;
+-----------+
| @@version |
+-----------+
| 8.0.27    |
+-----------+
1 row in set (0.00 sec)

mysql> select x'';
+----------+
| x''      |
+----------+
| 0x       |
+----------+
1 row in set (0.00 sec)

mysql> select X'01';
+--------------+
| X'01'        |
+--------------+
| 0x01         |
+--------------+
1 row in set (0.00 sec)

mysql> select 0X01;
ERROR 1054 (42S22): Unknown column '0X01' in 'field list'
mysql> select 0x;
ERROR 1054 (42S22): Unknown column '0x' in 'field list'
```

## Related Issue(s)
Backport of: https://github.com/vitessio/vitess/pull/9625

## Checklist
- [x] Tests were are not required (existing tests for this)
- [x] Documentation is not required (internal cleanup)

### DIFF
--- a/go/sqltypes/value.go
+++ b/go/sqltypes/value.go
@@ -23,7 +23,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"regexp"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


frouioui (approved):
LGTM. Let’s merge this once the original PR is green and merged :)

PR #9631, opened 2022-02-03
Bypass normalization of already prepared statements.

aquarapid:
* The normalization we do in vtgate during planning before checking the plan cache is expensive for large statements (think multi-value inserts).
 * For statements that are already prepared, normalizing them adds little value, so we should bypass that.

The only thing to consider here is whether we should make this configurable behavior.

For a bit of performance context, on a workload where prepared statements are used to insert 100 rows into a 40-way sharded table at a time, this chops the plan time for each query in half in my test environment (from ~ 4 ms per query to < 2 ms per query).


aquarapid:
As context;  the use-case here are large inserts into MySQL backend(s) that are pretty speedy (running on bare metal), and the actual overall time spent per insert is 8-10 ms;  so the plan time ends up being almost 50% of that...

vmg (commented):
This seems like a sensible change to me, but maybe it does belong behind a config flag (defaulting to true).

aquarapid:
For release notes:

```
In previous Vitess releases, prepared statement queries were normalized by vtgate if
-normalize_queries is true (which it is by default).  We have introduced the new
vtgate flag -normalize_prepared_stmts which defaults to false.  This changes the
default behavior of vtgate to *not* normalize prepared statements.  This can increase
performance if your workload is prepared-statement heavy, by reducing the CPU overhead
in vtgate, and by reducing the planning latency, especially for large queries. You
can revert to the previous vtgate behavior by setting the -normalize_prepared_stmts
flag to true.
```

go/vt/vtgate/executor.go
```diff
@@ -1420,7 +1420,8 @@ func (e *Executor) getPlan(vcursor *vcursorImpl, sql string, comments sqlparser.
 	vcursor.SetIgnoreMaxMemoryRows(ignoreMaxMemoryRows)
 
 	// Normalize if possible and retry.
-	if (e.normalize && sqlparser.CanNormalize(stmt)) || sqlparser.MustRewriteAST(stmt, qo.getSelectLimit() > 0) {
+	// We do not normalize already prepared statements, by default
+	if (e.normalize && (*normalizePrepStmts || len(bindVars) == 0) && sqlparser.CanNormalize(stmt)) || sqlparser.MustRewriteAST(stmt, qo.getSelectLimit() > 0) {
```
vmg, line 1424: Can you triple check that `len(bindVars) == 0` is the right logic here? I see you went back and forth on it in your Git history. :p

...We're saying that a prepared statement cannot have bind variables. Is that right?

aquarapid: We are saying that prepared statements *will* have > 0 bind variables, and thus are eligible for normalization (or, if normalization is being forced).

vmg: OK I'm a bit lost here...  The title of the PR is "Bypass normalization of already prepared statements", which means that prepared statements should _not_ be eligible for normalization.

aquarapid: missed a *not* in my comment above

go/vt/vtgate/executor.go
```diff
@@ -1494,8 +1493,9 @@ func (e *Executor) getPlan(vcursor *vcursorImpl, sql string, comments sqlparser.
 	return e.checkThatPlanIsValid(stmt, plan)
 }
 
-func (e *Executor) canNormalizeStatement(stmt sqlparser.Statement, qo iQueryOption, setVarComment string) bool {
-	return (e.normalize && sqlparser.CanNormalize(stmt)) ||
+func (e *Executor) canNormalizeStatement(stmt sqlparser.Statement, bindVars map[string]*querypb.BindVariable, qo iQueryOption, setVarComment string) bool {
+	// We do not normalize already prepared statements, by default
+	return (e.normalize && (*normalizePrepStmts || len(bindVars) == 0) && sqlparser.CanNormalize(stmt)) ||
```
vmg, line 1498: The point of extracting this into a function was breaking this logic into several `if`s instead of a single logical expression. I think that would be a worthy refactoring that would make this easier to follow.

vmg (approved):

go/vt/vtgate/executor.go
aquarapid: wildo

maxenglander: Hey @vmg hopefully I understood your intention -- I broke up the components of this `if` into multiple `if` statements. Can break the first one down further if you like LMK if you had something else in mind here.

maxenglander:
@vmg I think I addressed the last outstanding comment. Let me know if there is anything else I should address. I saw @aquarapid mention above about release notes. Should I add those [here](https://github.com/vitessio/vitess/blob/main/doc/releasenotes/15_0_0_summary.md) as part of this PR?

go/vt/vtgate/vtgate.go
```diff
@@ -104,6 +105,7 @@ var (
 func registerFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&transactionMode, "transaction_mode", transactionMode, "SINGLE: disallow multi-db transactions, MULTI: allow multi-db transactions with best effort commit, TWOPC: allow multi-db transactions with 2pc commit")
 	fs.BoolVar(&normalizeQueries, "normalize_queries", normalizeQueries, "Rewrite queries with bind vars. Turn this off if the app itself sends normalized queries with bind vars.")
+	fs.BoolVar(&normalizePrepStmts, "normalize_prepared_stmts", normalizePrepStmts, "when set to true, normalize prepared statement queries as well, if -normalize_queries is already true")
```
harshit-gangal, line 108: new flags are following the convention of dashes over underscore

harshit-gangal (commented):
I think this makes sense to avoid normalization.
But I would suggest we handle it from the caller 
like in plugin_mysql_server on the call
`vh.vtg.Execute(ctx, session, prepare.PrepareStmt, prepare.BindVars)`
we could add a parameter that says it is prepared statement so do not normalize.

This will help the other users of the Vitess who uses prepared statement and does not rely on this flag.

PR #9633, opened 2022-02-04
Add flag to vtcombo to load a json-encoded test topology file

simon-engledew:
## Description

This PR introduces a new flag to vtcombo (`-json_topo`) which allows a user to load a topology file from a JSON-encoded document.

In our testing setup we create topology files dynamically. JSON is easier to generate from a range of sources.

## Checklist

## Deployment Notes

I don't think this applies to vtcombo?

### DIFF
--- a/go/cmd/vtcombo/main.go
+++ b/go/cmd/vtcombo/main.go
@@ -23,12 +23,12 @@ package main
 
 import (
 	"context"
-	"encoding/json"
 	"flag"
 	"os"
 	"strings"
 	"time"
 
+	"google.golang.org/protobuf/encoding/protojson"
 	"google.golang.org/protobuf/encoding/prototext"
 	"google.golang.org/protobuf/proto"
 
@@ -123,9 +123,7 @@ func main() {
 	tpb := &vttestpb.VTTestTopology{}
 	switch {
 	case *jsonTopo != "":
-		decoder := json.NewDecoder(strings.NewReader(*jsonTopo))
-		decoder.DisallowUnknownFields()
-		if err := decoder.Decode(tpb); err != nil {
+		if err := protojson.Unmarshal([]byte(*jsonTopo), tpb); err != nil {
 			log.Errorf("cannot parse topology: %v", err)
 			exit.Return(1)
 		}
--- a/go/cmd/vtcombo/main.go
+++ b/go/cmd/vtcombo/main.go
@@ -23,6 +23,7 @@ package main
 
 import (
 	"context"
+	"encoding/json"
 	"flag"
 	"os"
 	"strings"
@@ -53,6 +54,7 @@ import (
 
 var (
 	protoTopo = flag.String("proto_topo", "", "vttest proto definition of the topology, encoded in compact text format. See vttest.proto for more information.")
+	jsonTopo  = flag.String("json_topo", "", "vttest proto definition of the topology, encoded in json format. See vttest.proto for more information.")
 
 	schemaDir = flag.String("schema_dir", "", "Schema base directory. Should contain one directory per keyspace, with a vschema.json file if necessary.")
 
@@ -119,9 +121,19 @@ func main() {
 
 	// parse the input topology
 	tpb := &vttestpb.VTTestTopology{}
-	if err := prototext.Unmarshal([]byte(*protoTopo), tpb); err != nil {
-		log.Errorf("cannot parse topology: %v", err)
-		exit.Return(1)
+	switch {
+	case *jsonTopo != "":
+		decoder := json.NewDecoder(strings.NewReader(*jsonTopo))
+		decoder.DisallowUnknownFields()
+		if err := decoder.Decode(tpb); err != nil {
+			log.Errorf("cannot parse topology: %v", err)
+			exit.Return(1)
+		}
+	default:
+		if err := prototext.Unmarshal([]byte(*protoTopo), tpb); err != nil {
+			log.Errorf("cannot parse topology: %v", err)
+			exit.Return(1)
+		}
 	}
 
 	// Stash away a copy of the topology that vtcombo was started with.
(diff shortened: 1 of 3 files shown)


mattlord:
I like it, thanks @simon-engledew! I think that we should add a test for this here (can modify/reuse/mirror `vttest_sample_test`). Are you comfortable with that? Let us know if you'd like any help.

go/cmd/vtcombo/main.go
```diff
@@ -119,9 +121,19 @@ func main() {
 
 	// parse the input topology
 	tpb := &vttestpb.VTTestTopology{}
-	if err := prototext.Unmarshal([]byte(*protoTopo), tpb); err != nil {
-		log.Errorf("cannot parse topology: %v", err)
-		exit.Return(1)
+	switch {
+	case *jsonTopo != "":
+		decoder := json.NewDecoder(strings.NewReader(*jsonTopo))
```
vmg, line 126: The `encoding/json` package is **not** going to decode properly a JSON payload into a protobuf struct like `VTTestTopology`. You need to use https://pkg.go.dev/google.golang.org/protobuf/encoding/protojson instead or we're going to see weird behaviors.

vmg (changes requested):

go/cmd/vtcombo/main.go
simon-engledew: TIL! Thanks 🙇 ❤️

simon-engledew:
> Are you comfortable with that? Let us know if you'd like any help.

@mattlord - 🤔 that test doesn't seem to use vtcombo at all at the moment, or am I missing something? I could move the unmarshal into a method in `vttest` and use it in the setup of that test but it seems a bit like I would just be testing the test and making things needlessly indirect.

Happy to do the work but I think you'll have to describe what you had in mind in a bit more detail. 🙇

mattlord:
> > Are you comfortable with that? Let us know if you'd like any help.
> 
> @mattlord - 🤔 that test doesn't seem to use vtcombo at all at the moment, or am I missing something? I could move the unmarshal into a method in `vttest` and use it in the setup of that test but it seems a bit like I would just be testing the test and making things needlessly indirect.

AFAIUI, it exercises the underlying package that `vtcombo` uses, `vttest`: https://pkg.go.dev/vitess.io/vitess/go/vt/proto/vttest

But honestly, you know more about vtcombo than I do. 🙂 I had never used it until helping your team with a related issue recently.

> Happy to do the work but I think you'll have to describe what you had in mind in a bit more detail. 🙇

It seems like we test the protobuf portion here: https://github.com/vitessio/vitess/blob/main/go/test/endtoend/vtcombo/vttest_sample_test.go#L57-L71

That's why I was thinking that we can mimic or modify the test to also verify the JSON based definitions.

simon-engledew:
> That's why I was thinking that we can mimic or modify the test to also verify the JSON based definitions.

Yeah, the test needs to reuse values from that protobuf (e.g: `"redirected"`), and because it never actually starts vtcombo it would not be testing the block of code that loads the data from the flags. 🤔 

I ran vtcombo locally and checked that it works for both modes. I tried a few sketches where I moved things around but the added indirection/complexity is more than the change in the first place 😅.

vmg (approved):
Thanks for the `protojson` change!

mattlord:
@simon-engledew 

> > That's why I was thinking that we can mimic or modify the test to also verify the JSON based definitions.
> 
> Yeah, the test needs to reuse values from that protobuf (e.g: `"redirected"`) so it seems weird to hard code them into a json file, and because it never actually starts vtcombo it would not be testing the block of code that loads the data from the flags. 🤔
> 
> I ran vtcombo locally and checked that it works for both modes. I tried a few sketches where I moved things around but the added indirection/complexity is more than the change in the first place 😅.

Yeah, I think creating a similar test to the existing one makes more sense. Here we're building the protobuf definition for the topology: https://github.com/vitessio/vitess/blob/main/go/test/endtoend/vtcombo/vttest_sample_test.go#L56-L71

Don't we want to test the same basic thing, but going from JSON to Proto as an intermediate step? 

We could continue to add things to `vtcombo` w/o tests, but that will come back to bite us later as you and others make more use of this as of yet somewhat unknown `vtcombo` binary, as the way you use it in CI operations makes a lot of sense IMO.

simon-engledew:
@mattlord Ok, I've changed the code to work using the standard `flags.Value` interface and then used it from the test. 🤷

This is the best way I can think of doing it which doesn't end up having an unecessary wrapper function sitting inside vttest. I think the original code is simpler, but at least this way you could use the flags in future test commands without having copy and paste code everywhere.

go/test/endtoend/vtcombo/vttest_sample_test.go
```diff
@@ -46,32 +46,37 @@ var (
...
-					{Name: "80-"},
-				},
-				RdonlyCount:  1,
-				ReplicaCount: 2,
-			},
-			{
-				Name:       redirected,
-				ServedFrom: ks1,
-			},
+		flag := vttest.JsonTopoFlag(&topology)
```
mattlord, line 72: Related to my comment about the structure names, IMO this would be more clear:
```
topoData := vttest.JsonTopoData(&topology)
```

go/vt/vttest/local_cluster.go
```diff
@@ -194,6 +200,33 @@ func (cfg *Config) DbName() string {
 	return ""
 }
 
+type TopoFlag struct {
+	ptr       *vttestpb.VTTestTopology
```
mattlord, line 204: IMO, it would be clearer to call this variable `vtTestTopology`.

mattlord (commented):
I think that makes sense. Thanks, @simon-engledew ! I had a few nits about names but don't feel super strongly about it. Let me know what you think.

simon-engledew:
@mattlord I can't tell if these tests are flakes and I can't restart CI 😅

mattlord:
> @mattlord I can't tell if these tests are flakes and I can't restart CI 😅

Yeah, they were just flaky. All good now!

PR #9634, opened 2022-02-04
Remove queryserver-config-terse-errors impact on log messages

mattlord:
## Description
This effectively reverts https://github.com/vitessio/vitess/pull/9094.

That PR made undocumented behavior changes in a development release: `13.0.0-SNAPSHOT` (never part of a GA/stable release). It added some log message redaction behavior to the `-queryserver-config-terse-errors` flag and altered the meaning/intention of that flag. Additionally, we did not audit the `vttablet` codebase at that time to ensure that sensitive log messages were actually sanitized in all known potential places (it focused on one specific error message location).

In https://github.com/vitessio/vitess/pull/9550 we intended to properly sanitize all potentially sensitive `vttablet` log messages and we added a new flag for this purpose: `-sanitize_log_messages`. That work was complicated, however, by the fact that `-queryserver-config-terse-errors` had an inconsistent and non-uniform impact on log messages.

Given that the changes in https://github.com/vitessio/vitess/pull/9094 were:
1. Never in a GA/stable release
2. Were not documented 
3. Did not actually work as it should have (there were other log messages where the bind variables were not redacted)

We are making the following changes here:
1. `-queryserver-config-terse-errors` goes back to doing what it was intended and documented to do, redacting client error messages. The behavior of this flag then remains the same between 12.0 and 13.0.
2. Log message sanitization is handled exclusively via a dedicated flag now: `-sanitize_log_messages`.

## Related Issue(s)
This effectively reverts: https://github.com/vitessio/vitess/pull/9094
Follow-up to: https://github.com/vitessio/vitess/pull/9550

## Checklist
- [x] Should this PR be backported? YES, to release-13.0 https://github.com/vitessio/vitess/pull/9636
- [x] Tests were updated

### DIFF
--- a/go/vt/vttablet/tabletserver/tabletserver.go
+++ b/go/vt/vttablet/tabletserver/tabletserver.go
@@ -1298,15 +1298,20 @@ func (tsv *TabletServer) handlePanicAndSendLogStats(
 	logStats *tabletenv.LogStats,
 ) {
 	if x := recover(); x != nil {
-		// TerseErrors should generally NOT impact log messages -- rather client errors -- but keeping
-		// this existing exception in place so as not to change the behavior for users of the flag.
-		errorMessage := fmt.Sprintf(
-			"Uncaught panic for %v:\n%v\n%s",
-			queryAsString(sql, bindVariables, (tsv.TerseErrors || tsv.Config().SanitizeLogMessages)),
-			x,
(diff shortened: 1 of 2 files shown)


GuptaManan100 (approved):

GuptaManan100 (approved):
We should wait for @deepthi 's review before merging

deepthi (approved):
I agree with the analysis in the description.

PR #9639, opened 2022-02-07
Extract collation data to enable distinct aggregation

systay:
Signed-off-by: Andres Taylor <andres@planetscale.com>

## Checklist


harshit-gangal:
I think the example query should contain `count(distinct ...)`

go/vt/vtgate/engine/ordered_aggregate.go
```diff
@@ -58,6 +58,10 @@ type OrderedAggregate struct {
 	// from the result received. If 0, no truncation happens.
 	TruncateColumnCount int `json:",omitempty"`
 
+	// Collations stores the collation ID per column offset.
+	// It is used for grouping keys and distinct aggregate functions
+	Collations map[int]collations.ID
+
```
harshit-gangal, line 64: should we not remove from GroupByParams?
This can be used instead of keeping in GroupByParams or AggregateParams

systay: it's kind of nice to keep it in those data structures, since it makes the description printing much easier. do you feel strongly about this?

harshit-gangal: it is just duplicated information. future changes might miss updating in one of them.

go/vt/vtgate/planbuilder/ordered_aggregate.go
```diff
@@ -333,6 +335,18 @@ func (oa *orderedAggregate) Wireup(plan logicalPlan, jt *jointab) error {
 }
 
 func (oa *orderedAggregate) WireupGen4(semTable *semantics.SemTable) error {
+	colls := map[int]collations.ID{}
+	oa.eaggr.Collations = colls
```
frouioui, line 339: Why not doing directly:
```go
oa.eaggr.Collations = map[int]collations.ID{}
```

frouioui (approved):
It looks good to me. I felt like `GroupByKeys` and `Collations` were a bit redundant, but I can see the difference in their usages, and couldn't think of a nice way to remove that redundancy.

go/vt/vtgate/planbuilder/ordered_aggregate.go
systay: I keep using this variable :point_down:, and I didn't want to spell out `oa.eaggr.Collations` every time

go/vt/vtgate/engine/ordered_aggregate.go
systay: well, it is duplicated, but since the map is built from the `GroupByParams` and `AggregateParams` objects, and they are helpful as they are when printing the plan description, I thought it was nice to keep the collation info there

PR #9641, opened 2022-02-07
Reduce the number of reserved connections when setting system variables

frouioui:
## Description

This pull request reduces the number of reserved connections that will be used when some system variables are set in the session. This enhancement is only enabled with MySQL 80+.

When we issue a DML query and have system variable(s) set for the session, instead of using a reserved connection, the new code will add use the `SET_VAR(var_name = value)` query hint. This query hint is only available in MySQL80+. Queries ran against MySQL < 80 will keep using reserved connections. Additionally, the use of `SET_VAR` can be toggled with the VTGate flag `-enable_set_var` (by default set to `true`).

## Performance

This feature's goal is to reduce the overall number of reserved connections that Vitess maintains. We have added two types of benchmarks to ensure this feature would not create a performance regression. The first is to use `SET_VAR` against a MySQL server, and the second is to use `SET_VAR` against VTGate (more end-to-end benchmark).

The former is available in a separate pull request: https://github.com/vitessio/vitess/pull/9681 and shows positive results with `SET_VAR` being one of the fastest options.

The latter is added through this pull request and also shows no regression. `BenchmarkReservedConnWhenSettingSysVar` benchmarks the execution of four queries (insert, delete, select, update) first using `SET_VAR`, and then using reserved connections.

```
goos: darwin
goarch: amd64
pkg: vitess.io/vitess/go/test/endtoend/vtgate/mysql80
cpu: Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
BenchmarkReservedConnWhenSettingSysVar
BenchmarkReservedConnWhenSettingSysVar/Use_SET_VAR
BenchmarkReservedConnWhenSettingSysVar/Use_SET_VAR-16         	     381	   2753711 ns/op
BenchmarkReservedConnWhenSettingSysVar/Use_reserved_connections
BenchmarkReservedConnWhenSettingSysVar/Use_reserved_connections-16         	     536	   3459926 ns/op
```

## Related issue(s)

This change is the second proposal made in the [Handle reserved/tainted connections better RFC](https://github.com/vitessio/vitess/issues/9706).

## Checklist
- [x] Documentation was added or is not required (https://github.com/vitessio/website/pull/979)


go/vt/sqlparser/ast_funcs.go
```diff
@@ -282,6 +283,46 @@ func (ct *ColumnType) SQLType() querypb.Type {
 	return sqltypes.Null
 }
 
+// AddQueryHint adds the given string to list of comment.
+// If the list is empty, one will be created containing the query hint.
+// If the list already contains a query hint, the given string will be merged with the existing one.
+// This is done because only one query hint is allowed per query.
+func (node Comments) AddQueryHint(queryHint string) (Comments, error) {
```
frouioui, line 290: Here I have assumed that query hints only start by `/*+` (based on MySQL documentation), but I found that we also have `/*vt+` at this day, I am not sure whether I should consider those comments as query hints too.
cc @harshit-gangal @systay

go/vt/sqlparser/ast_funcs_test.go
```diff
@@ -0,0 +1,82 @@
...
+			queryHint: "SET_VAR(aa)",
+			expected:  Comments{"/*+ SET_VAR(aa) */", "/* toto */"},
+		},
+		{
+			node:      Comments{"/* toto */", "/*+ SET_VAR(bb) */"},
+			queryHint: "SET_VAR(aa)",
+			expected:  Comments{"/*+ SET_VAR(bb) SET_VAR(aa) */", "/* toto */"},
+		},
+		{
+			node:      Comments{"/* toto */", "/*+ SET_VAR(bb) "},
```
systay, line 55: nit: the parser would never produce a comment like this

go/vt/vtgate/executor.go
```diff
@@ -1461,6 +1467,30 @@ func (e *Executor) getPlan(vcursor *vcursorImpl, sql string, comments sqlparser.
...
+		vcursor.NeedsReservedConn()
+		return "", nil
+	}
+
+	res := &bytes.Buffer{}
+	for k, val := range sysVars {
+		_, err := res.WriteString(fmt.Sprintf("SET_VAR(%s = %s) ", k, val))
+		if err != nil {
+			return "", vterrors.Errorf(vtrpcpb.Code_INTERNAL, "Buffer exceeded memory limit")
+		}
```
vmg, line 1488:
```suggestion
    _ = fmt.Fprintf(&res, "SET_VAR(%s = %s) ", k, val)
```

There's no need to check for the error return here because `strings.Builder` cannot return an error. Same for your old code with `WriteString` on `bytes.Buffer`. Those functions cannot error. If they run out of heap they'll panic.

go/vt/sqlparser/ast_funcs_test.go
GuptaManan100: I maybe wrong, but I personally think that it is better to test the behaviour of a function even for reasonable inputs that it might never receive due to the functioning of a different function

go/vt/vtgate/executor.go
```diff
@@ -1461,6 +1467,30 @@ func (e *Executor) getPlan(vcursor *vcursorImpl, sql string, comments sqlparser.
...
+	}
+	res.Truncate(res.Len() - 1)
+	return res.String(), nil
```
vmg, line 1491:
```suggestion
	return res.String()[:res.Len() - 1], nil
```

go/vt/sqlparser/ast_funcs.go
harshit-gangal: we should not pass down /*vt+ comments down. It should be removed after using those hints at vtgate.

go/vt/vtgate/safe_session.go
```diff
@@ -387,6 +387,13 @@ func (session *SafeSession) SetSystemVariable(name string, expr string) {
 	session.SystemVariables[name] = expr
 }
 
+// GetSystemVariables gets the system variables in the session.
+func (session *SafeSession) GetSystemVariables() map[string]string {
+	session.mu.Lock()
+	defer session.mu.Unlock()
+	return session.SystemVariables
```
vmg, line 394: This is not safe. Note that there's a sibling method, `SetSystemVariable`, that can update the `session.SystemsVariables` map while holding `session.mu`. Here, you're only holding `session.mu` to read the value of `session.SystemVariables`, but you return the whole map to the caller, so the caller can then either _read a key from the map_ or _write a key to the map_ without holding the mutex. Neither of those things are safe to do because they can conflict with concurrent writes.

In order to have non-racy semantics, you'll need to create a hard copy of the map while you're holding `session.mu`, and then return that copy.

vmg (changes requested):

go/vt/vtgate/safe_session.go
systay: We should probably just change this method to take the key in and return the value, and we can do the read while protected by the mutex - the mirror of the `SetSystemVariable` method :point_up:

go/vt/sqlparser/ast_funcs.go
frouioui: Got it

go/vt/vtgate/engine/set.go
```diff
@@ -396,6 +401,10 @@ func sqlModeChangedValue(qr *sqltypes.Result) (bool, sqltypes.Value) {
 		changed = true
 	}
 
+	if qr.Rows[0][1].Len() == 0 {
+		return changed, sqltypes.NewVarChar(" ")
```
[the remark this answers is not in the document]
frouioui: MySQL does not handle empty strings (`""`) correctly in `SET_VAR`:
```sql
mysql> select /*+ SET_VAR(sql_mode = "") */ @@sql_mode;
+-----------------------------------------------------------------------------------------------------------------------+
| @@sql_mode                                                                                                            |
+-----------------------------------------------------------------------------------------------------------------------+
| ONLY_FULL_GROUP_BY,STRICT_TRANS_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO,NO_ENGINE_SUBSTITUTION |
+-----------------------------------------------------------------------------------------------------------------------+

mysql> select /*+ SET_VAR(sql_mode = " ") */ @@sql_mode;
+------------+
| @@sql_mode |
+------------+
|            |
+------------+
1 row in set (0.00 sec)
```

When running the first query MySQL produces a warning:

```sql
mysql> show warnings;
+---------+------+--------------------------------------------------------------------------+
| Level   | Code | Message                                                                  |
+---------+------+--------------------------------------------------------------------------+
| Warning | 1064 | Optimizer hint syntax error near '"") */ @@sql_mode from dual' at line 1 |
+---------+------+--------------------------------------------------------------------------+
1 row in set (0.00 sec)
```

As we can see `SET_VAR(sql_mode = )` is not happy with an empty string and ignore the optimizer hint. This is strange behavior knowing that running a `set @@sql_mode = ""` works just fine.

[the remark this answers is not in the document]
frouioui: Answered the same question [here](https://github.com/vitessio/vitess/pull/9641#discussion_r803694443). :)

go/vt/sqlparser/ast.go
```diff
@@ -37,6 +37,13 @@ type (
 		SQLNode
 	}
 
+	// Commentable represents statements that are worth being commented.
+	Commentable interface {
+		iCommentable()
```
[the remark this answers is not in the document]
systay: If that is what the interface means, I suggest renaming it to something like that. `SupportsSetVar` or something like that

go/vt/vtgate/safe_session.go
systay: sounds wasteful to me, but maybe I'm overthinking it

another alternative would be a visitor pattern that executes under the protection of the mutex :shrug:

frouioui: Fixed via f0037678c9b7bf4156f7537f8be2a195554ca315 :)

go/vt/vtgate/engine/set.go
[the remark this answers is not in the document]
frouioui: Done :)

vmg (approved):

go/vt/vtgate/executor.go
```diff
@@ -1461,6 +1465,31 @@ func (e *Executor) getPlan(vcursor *vcursorImpl, sql string, comments sqlparser.
...
+	if _, supportsOptimizerHint := stmt.(sqlparser.SupportOptimizerHint); !supportsOptimizerHint {
+		vcursor.NeedsReservedConn()
+		return "", nil
+	}
+
+	var res strings.Builder
+	vcursor.Session().GetSystemVariables(func(k, v string) {
+		res.WriteString(fmt.Sprintf("SET_VAR(%s = %s) ", k, v))
+	})
+	return strings.TrimSpace(res.String()), nil
```
harshit-gangal, line 1490: trying to understand if there is a special reason to send as a function to GetSystemVariables

harshit-gangal (approved):
looks like there are accidental changes to `go/vt/sqlparser/analyzer.go`, `go/vt/vtgate/engine/revert_migration.go` and `go/vt/vtgate/executor_test.go`

frouioui:
> looks like there are accidental changes to `go/vt/sqlparser/analyzer.go`, `go/vt/vtgate/engine/revert_migration.go` and `go/vt/vtgate/executor_test.go`

@harshit-gangal  - Those are voluntary changes: `go/vt/sqlparser/analyzer.go` and `go/vt/vtgate/engine/revert_migration.go` to add missing spaces in comments, and `go/vt/vtgate/executor_test.go` because we changed the query produced when running `set @@sql_mode = ''` the expected test output changed too

go/vt/vtgate/executor.go
frouioui: This is done to avoid data race while using the system variables map. More details on [this comment thread](https://github.com/vitessio/vitess/pull/9641#discussion_r803656132) :)

go/vt/vtgate/executor_select_test.go
```diff
@@ -113,6 +114,96 @@ func TestSelectDBA(t *testing.T) {
 	utils.MustMatch(t, wantQueries, sbc1.Queries)
 }
 
+func TestSystemVariablesUsingSetVarComment(t *testing.T) {
+	executor, sbc1, _, _ := createExecutorEnv()
+	executor.normalize = true
+
+	session := NewSafeSession(&vtgatepb.Session{TargetString: "TestExecutor", SystemVariables: map[string]string{"sql_mode": "only_full_group_by"}})
```
harshit-gangal, line 121: we should not set the session system variables directly then it would not test if there is any bug in calling set <sys_var> query.

instead, we should call executor.Execute with the set statement and then check the session for the InReservedConn value.

go/vt/vtgate/executor_select_test.go
```diff
@@ -113,6 +114,96 @@ func TestSelectDBA(t *testing.T) {
...
+	require.False(t, session.InReservedConn())
+	for _, shardSession := range session.ShardSessions {
+		require.EqualValues(t, 0, shardSession.ReservedId)
+	}
+
+	wantQueries := []*querypb.BoundQuery{
+		{Sql: "select /*+ SET_VAR(sql_mode = only_full_group_by) */ :vtg1 from information_schema.`table`", BindVariables: map[string]*querypb.BindVariable{"vtg1": {Type: sqltypes.Int64, Value: []byte("1")}}},
+	}
+
+	utils.MustMatch(t, wantQueries, sbc1.Queries)
```
harshit-gangal, line 135: after this check.
we should execute another set sys_var with sysvar that does not support SET_VAR. and check the session again.
then execute a select query and see that both the setting, previous one and the new one are applied and reserved connection is created.

go/vt/vtgate/safe_session.go
harshit-gangal: these settings are stored at the connection level, and MySQL protocol says that they should not be multiplexed. 
The purpose of the safeSession was actually in the execution phase where you are accessing same information for executing in parallel for multi-shards. 
The above information and SetSystemVariables are not in those paths. Ideally, we should just clean up these locks and only keep which are actually required.

go/vt/vtgate/executor_select_test.go
frouioui: done via 23c2f6a

go/vt/sqlparser/ast_rewriting.go
```diff
@@ -255,12 +263,20 @@ type expressionRewriter struct {
...
-	selectLimit int
+	keyspace      string
+	selectLimit   int
+	setVarComment string
+	sysVars       map[string]string
 }
 
-func newExpressionRewriter(keyspace string, selectLimit int) *expressionRewriter {
-	return &expressionRewriter{bindVars: &BindVarNeeds{}, keyspace: keyspace, selectLimit: selectLimit}
+func newExpressionRewriter(keyspace string, selectLimit int, setVarComment string, sysVars map[string]string) *expressionRewriter {
```
systay, line 272: nit: I know this isn't the first thing that `expressionRewriter` is responsible for that is not an expression. Still - maybe now the time has finally come to rename this struct to something broader than expression?

frouioui: that's a good point, I renamed it to `astRewriter` since we're basically rewriting most of the AST

go/test/endtoend/vtgate/mysql80/misc_test.go
```diff
@@ -114,6 +116,117 @@ func TestVersionCommentWorks(t *testing.T) {
...
+		})
+	}
+}
+
+func TestUseSystemAndUserVariables(t *testing.T) {
+	conn, err := mysql.Connect(context.Background(), &vtParams)
+	require.NoError(t, err)
+	defer conn.Close()
+
+	utils.Exec(t, conn, "set @@sql_mode = ''")
```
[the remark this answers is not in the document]
frouioui: I removed the assertion as it's not really meaningful anymore. We already have some other tests that test this behavior

PR #9642, opened 2022-02-07
OnlineDDL: 'vitess' strategy, synonym for 'online'

shlomi-noach:
## Description

Naming is hard, and the `online` ddl-strategy was created in the hope/intention that it becomes the one and only (or the default) strategy. This takes longer than hoped for, and the term "online" is confusing.

This PR introduces a `vitess` strategy. It is identical in all functionality to `online`: `vitess` is an online (see the confusion?) strategy that uses VReplication as the underlying backend for `ALTER TABLE` operations; just like `online` strategy.

At this time, the two strategies `vitess` and `online` co-exist side by side. For backwards compatibility, we do not modify one into another at this time. At some point in the future we will implicitly change any `online` strategy into `vitess`. But we're not there yet.

In the meantime, we will start using `vitess` as the preferred strategy in documentation.

Tests are updated to use `vitess` strategy in most cases, but we leave `online` in some cases to validate that both are supported.

## Related Issue(s)

- tracking: #6926 

## Checklist

cc @deepthi @rbranson


deepthi (commented):
Should we be making this change in 13.0?

shlomi-noach:
> Should we be making this change in 13.0?

However you like

shlomi-noach:
backport in https://github.com/vitessio/vitess/pull/9650

deepthi (approved):

PR #9643, opened 2022-02-07
Gen4: Rework how aggregation is planned

systay:
## Description
Rework of how group by and aggregations are planned in Gen4.

We push down aggregations using the logic from the paper [Orthogonal Optimization of Subqueries and Aggregation](https://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.563.8492&rep=rep1&type=pdf), by
Cesar A. Galindo-Legaria and Milind M. Joshi from Microsoft Corp.

It explains how one can split an aggregation into local aggregates that depend on only one side of the join.
The local aggregates can then be gathered together to produce the global
group by/aggregate query that the user asked for.

In Vitess, this is particularly useful because it allows us to push aggregation down to the routes, even when
we have to join the results at the vtgate level. Instead of doing all the grouping and aggregation at the
vtgate level, we can offload most of the work to MySQL, and at the vtgate just summarize the results.

It will support a lot more queries aggregating cross shards than the current planner does.

An example might be easier to follow. Take a simple aggregation query:

```sql
select user_extra.city, sum(user.salary) 
from user join user_extra on user.id = user_extra.id
group by user_extra.city
```

Because sharding and vschema, we are not able to merge the two sides of the join, and so we are forced to collect data from each table and do the join and aggregation on the vtgate.

So, we have to do joining on the vtgate, but we still need the inputs from the vttablets, The queries we'll send to the tablets will look like:

```sql
-- LEFT
select user.id, sum(user.salary)
from user 
group by user.id

-- RIGHT
select user_extra.city, count(*) 
from user_extra where  user_extra.id = :user_id
group by user_extra.city
```

The join we are executing is a nested loop join. So, for every row we see coming from the left input, we will execute the query on the right, with the bind variable `:user_id` set.

On the left hand side, we have changed the grouping to be on `user.id`, instead of the original, which was a column that is not available to us on the left side. Since the join predicate is on `user.id`, we can group all the rows with the same `user.id` together - we know they will match the same row on the right.

We add the `count(*)` to the right hand side query to delegate as much work as possible to MySQL - if there are multiple rows with the same `user_extra.city`, we want them grouped before they are sent over the wire. 

Now, after evaluating the join, we can multiply the `sum(user.salary)` with the `count(*)` from the right, and then do the grouping on the vtgate.


go/vt/vtgate/planbuilder/testdata/aggr_cases.txt
```diff
@@ -3108,11 +3089,342 @@ Gen4 plan same as above
...
+                        "Inputs": [
+                          {
+                            "OperatorType": "Route",
+                            "Variant": "Scatter",
+                            "Keyspace": {
+                              "Name": "user",
+                              "Sharded": true
+                            },
+                            "FieldQuery": "select ue.bar, count(*), count(ue.foo) from user_extra as ue where 1 != 1",
+                            "Query": "select ue.bar, count(*), count(ue.foo) from user_extra as ue where ue.bar = :u_foo",
```
systay, line 3404: missing group by `ue.bar`

frouioui: fixed via [9ce1dd9](https://github.com/vitessio/vitess/pull/9643/commits/9ce1dd9f892fb42ba36bc94acd01d3ad8de7958e)

systay:
I want to add more end to end tests, but there is enough to start reviewing already

go/vt/vtgate/evalengine/expressions.go
```diff
@@ -520,6 +520,9 @@ func (t TupleExpr) typeof(*ExpressionEnv) (sqltypes.Type, flag) {
 }
 
 func (c *Column) typeof(env *ExpressionEnv) (sqltypes.Type, flag) {
+	if len(env.Row) <= c.Offset {
+		return sqltypes.Null, 0
```
vmg, line 524: This seems fine, but you need to return the `flagNull|flagNullable` together with the `Null` value, otherwise other expressions will fail to typecheck.

systay:
This PR now depends on #9807 and #9810

systay:
Should not be merged yet! Still chasing a bug that only manifests before the schema tracking has had a chance to catch up

go/vt/vtgate/planbuilder/abstract/queryprojection.go
```diff
@@ -57,12 +60,42 @@ type (
 		Inner         sqlparser.Expr
 		WeightStrExpr sqlparser.Expr
 
-		// This is to add the distinct function expression in grouping column for pushing down but not be to used as grouping key at VTGate level.
-		// Starts with 1 so that default (0) means unassigned.
-		DistinctAggrIndex int
+		// The index at which the user expects to see this column. Set to nil, if the user does not ask for it
+		InnerIndex  *int
+		aliasedExpr *sqlparser.AliasedExpr
```
GuptaManan100, line 65: Could we also add a comment here for `aliasedExpr` which is set only if the group by field is present in the select list.

go/vt/vtgate/planbuilder/abstract/queryprojection.go
```diff
@@ -357,12 +412,143 @@ func (qp *QueryProjection) NeedsDistinct() bool {
...
+		}
+	}
+	return nil, nil
+}
+
+// AlignGroupByAndOrderBy aligns the group by and order by columns, so they are in the same order
+// The GROUP BY clause is a set - the order between the elements does not make any difference,
+// so we can simply re-arrange the column order
+// We are also free to add more ORDER BY columns than the user asked for which we leverage,
+// so the input is already ordered according to the GROUP BY columns used
```
GuptaManan100, line 507: ❤️

GuptaManan100 (approved):

frouioui (approved):
That's a huge change nicely done, hooray to reading research papers and implementing them in Vitess :) LGTM!

Issue #9588, opened 2022-01-28
Add CI check for drift between src/proto and generated pb.go files


doeg:
Similar to https://github.com/vitessio/vitess/issues/9571, this issue proposes we add a CI check to ensure that the `proto/*.proto` files are in sync with the generated `go/vt/proto/*.pb.go` files. This would catch any instance where a .proto file is updated but `make proto` is not run.

frouioui:
Closing this issue, #9644 is merged :)

Issue #9001, opened 2021-10-14
Add CI check for vtadmin-web proto/typegen drift


doeg:
## tl;dr

Right now, it's up to humans to remember to run `make vtadmin_web_proto_types` after modifying [vtadmin.proto](https://github.com/vitessio/vitess/blob/69e3766/proto/vtadmin.proto) or any of its dependencies. For branches that modify at least one [proto/*.proto](https://github.com/vitessio/vitess/tree/main/proto) file, I propose adding a CI check for drift between the generated types on the PR branch vs. the main branch. 

## Longer version

The [`vtadmin_web_proto_types` makefile target](https://github.com/vitessio/vitess/blob/main/Makefile#L460-L464) generates TypeScript declarations + .js wrappers from [vtadmin.proto](https://github.com/vitessio/vitess/blob/69e3766/proto/vtadmin.proto), into [vtadmin-web/src/proto/](https://github.com/vitessio/vitess/tree/14558dd/web/vtadmin/src/proto).

Anytime vtadmin.proto or any of its .proto dependencies are modified, it's necessary to run `make proto` (as usual) and _then_ `make vtadmin_web_proto_types`. 

Forgetting this step can result in vtadmin-web runtime errors, since vtadmin-web will be using different generated types, compared to vtadmin-api and the rest of Vitess.

This, combined with a CI check to confirm the front-end build is green (https://github.com/vitessio/vitess/pull/8993) should avoid situations like https://github.com/vitessio/vitess/pull/8623 and https://github.com/vitessio/vitess/pull/8566. In both cases, I wasn't around to review and it (understandably!!!) wasn't clear to the PR authors whether or not the changes broke/otherwise affected vtadmin-web.

@ajm188 mentioned that a proto diff/drift check could be more widely useful to other makefile targets. I wanted to keep this ticket scoped fairly narrowly to vtadmin-web since it's less work for me 😈 but if it's applicable elsewhere, then that's great + I'd love to know about it! 

## A solution + an alternative

Two potential solutions, both of which will address the problem of drift between vtadmin.proto and the front-end codegen. ...Well, more like: one good solution and then a less-good alternative that I will mention for completeness. :) 

[...]

PR #9644, opened 2022-02-07
Check for proto drift in the CI

frouioui:
## Description

This pull request adds the `check_make_proto` workflow to the CI. The workflow checks that the `go/vt/proto/*.pb.go` files are in accordance with the `proto/*.proto` files. This is done by running the `make proto` command and diffing `git status`'s output before and after executing the `make proto`.

If there is a diff (a drift), the workflow lists the files that are not in accordance. A commit that exhibits this behavior can be found [here](https://github.com/vitessio/vitess/pull/9644/commits/a96bb3350591db945bdea182e5cb2f1e7159de47), with its failing `check_make_proto` workflow [here](https://github.com/vitessio/vitess/runs/5095360428?check_suite_focus=true).

## Related Issue(s)

- Solves #9588 

## Checklist

### DIFF
--- a/tools/check_make_proto.sh
+++ b/tools/check_make_proto.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+
+source build.env
+
+first_output=$(git status --porcelain)
+
+make proto
+
+second_output=$(git status --porcelain)
+
+diff=$(diff <( echo "$first_output") <( echo "$second_output"))
+
+if [[ "$diff" != "" ]]; then
+  echo "ERROR: Regenerated proto files do not match the current version."
+  echo -e "List of files containing differences:\n$diff"
+  exit 1
+fi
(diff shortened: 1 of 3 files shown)


tools/check_make_proto.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+source build.env
+
+first_output=$(git status --porcelain)
```
doeg, line 5: Wouldn't an initial `git status` on a fresh checkout always be empty?

doeg (approved):
I love this! 😻 Thank you so much. 

.......... I will also be ~~stealing~~ borrowing your excellent idea for #9001 >:)

deepthi (approved):
You may merge after addressing the question from @doeg.
Can you squash-merge, please?

tools/check_make_proto.sh
frouioui: Yes, technically the first git status should be empty, however, I wanted to make sure that the only possible reason for a drift comes from make proto, no matter how or when this script is being run. So I assumed the script could be run in any situation/environment, and not necessarily on a clean git state :)

PR #9646, opened 2022-02-07
evalengine: use sqltypes

vmg:
## Description

This is a cleanup PR: the new implementation of the `evalengine` was using an ugly mixture of SQL types from `querypb` and from `sqltypes`, even though all these constants are aliases between each other.

This PR cleans this mess by switching everything to the constants defined in `sqltypes`, which are more readable.

While I was here, I also added a randomization element to the integration tests: now every test run decides at random whether to simplify the tested expressions before evaluation. The expressions should (obviously) give identical results with and without simplification, so this randomization is a good way to ensure this property over time without doubling the duration of the integration tests, which are already quite long.

## Checklist

### DIFF
--- a/go/sqltypes/type.go
+++ b/go/sqltypes/type.go
@@ -22,6 +22,8 @@ import (
 	querypb "vitess.io/vitess/go/vt/proto/query"
 )
 
+type Type = querypb.Type
+
 // This file provides wrappers and support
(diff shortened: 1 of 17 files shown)


frouioui (approved):
this is a great idea, thanks!

PR #9612, opened 2022-02-01
[vtadmin-web] Update postcss to v8.4.6

doeg:
Signed-off-by: Sara Bee <855595+doeg@users.noreply.github.com>

## Description

This updates to postcss v8.4.6 in order to fix [CVE-2021-23566](https://www.whitesourcesoftware.com/vulnerability-database/CVE-2021-23566) in the nested [nanoid](https://www.npmjs.com/package/nanoid) dependency. 

For reference, here is the [postcss CHANGELOG](https://github.com/postcss/postcss/blob/main/CHANGELOG.md).

## Related Issue(s)

- postcss was last updated in #9493 

## Checklist
- [x] Should this PR be backported? **Yes,** I can create a PR to backport this to 13.0 RC1.
- [x] Tests were added or are not required **N/A**
- [x] Documentation was added or is not required  **N/A**

## Deployment Notes

VTAdmin operators are recommended to redeploy VTAdmin to pick up the fix.

PR #9493, opened 2022-01-10
[VTAdmin] Update react-scripts, postcss

notfelineit:
## Description
This PR updates some vulnerable npm packages, namely:
- svgo to ^2.8.0
- react-scripts to ^5.0.0
    - Changelog: https://github.com/facebook/create-react-app/releases/tag/v5.0.0

**React Scripts**
react-script 5.0.0 has added support for tailwindcss. Followed the steps outlined here: https://github.com/facebook/create-react-app/pull/11717#issuecomment-987231653 to make current version compatible. That includes manually installing necessary postcss plugins.

There are also **postcss** config updates as suggested here: https://github.com/facebook/create-react-app/issues/11777

## Testing
Manual testing was done here, made sure the vtadmin portal rendered "as normal" - primarily the debug page with all the tailwind styles 😃 

## Related Issue(s)
N/A, run `npm audit fix`

## Checklist
- [x] Should this PR be backported? **No**
- [x] Tests were added or are not required
- [x] Documentation was added or is not required

## Deployment Notes
None

PR #9648, opened 2022-02-07
[vtadmin-web] Upgrade to tailwindcss v3

doeg:
## Description

This updates vtadmin-web to use tailwindcss v3, following the [Upgrade Guide](https://tailwindcss.com/docs/upgrade-guide).

Here's the announcement, for the curious: https://tailwindcss.com/blog/tailwindcss-v3

I verified that everything still looks + works ok by poking around in the UI. The most likely oversight (since it's difficult to grep for) is any instance of "dynamic classnames", which are [not supported](https://tailwindcss.com/docs/content-configuration); that is, class names that use string templates like this:

```jsx 
// Not supported
<div class="text-{{ error ? 'red' : 'green' }}-600"></div>

// OK!
<div class="{{ error ? 'text-red-600' : 'text-green-600' }}"></div>
```

Happy to fix forward anything I may have missed. I've long dreamed of something like [Percy](https://www.browserstack.com/percy) for visual regression testing... but that's a whole thing that comes with its own trade-offs. So! Manual verification for now. 

Beyond having the latest + greatest, the other motivation for doing this now is that the `tailwindcss@npm:@tailwindcss/postcss7-compat` dependency, [a requirement from tailwindcss v2](https://v2.tailwindcss.com/docs/guides/create-react-app), breaks `npm outdated`: 

```
🍕~/workspace/vitess/web/vtadmin $ npm outdated
npm ERR! Only tag, version, and range are supported

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/sarabee/.npm/_logs/2022-02-07T19_25_43_567Z-debug.log
```

## Related Issue(s)
 
This goes along with the react-scripts + postcss upgrade in #9493, and the subsequent postcss upgrade in #9612. 

## Checklist

## Deployment Notes

N/A


web/vtadmin/src/index.css
```diff
@@ -232,3 +230,121 @@ table tbody td[rowSpan] {
 .Toastify__toast-theme--dark {
     background: none !important;
 }
+
+/* See https://tailwindcss.com/docs/extracting-components */
+@layer components {
```
doeg, line 235: I pulled this CSS from src/style/components.css, and then deleted that file.

This is because tailwindcss v3 tweaked its base button styles in a way that conflicted with our `.btn` component, as described in https://github.com/tailwindlabs/tailwindcss/issues/6602. Trying to fix it resulted in a real nasty snarl of [import ordering issues](https://tailwindcss.com/docs/using-with-preprocessors#build-time-imports), so I've opted to simply pull everything into this one file for now.

web/vtadmin/src/components/routes/Debug.module.scss
```diff
@@ -56,3 +56,201 @@
 .tabContainer {
     margin: 2.4rem 0;
 }
+
+.danger {
```
doeg, line 60: This unpleasant part of the diff adds a bunch of CSS classes for rendering the colour chips on the debug page. The new JIT compiler in tailwindcss v3 really simply does _not_ work with "runtime classes", e.g., a JavaScript string template like `background-${foo}`.

There are more "elegant" ways to fix this with postcss loops but it looked like it would require adding a new postcss dependency to support loops and... 🤷

notfelineit (approved):
Thank you for handling this!

PR #9638, opened 2022-02-06
OnlineDDL: delete _vt.vreplication entry as part of CLEANUP operation

shlomi-noach:
## Description

Migration cleanup happens either on:

- explicit `ALTER VITESS_MIGRATION ... CLEANUP` query
- automated migration garbage collection

In either case, migration cleanup now also makes sure to delete any associated `vreplication` workflow used by the migration. There can only be one such workflow, and one exists when the migration is `ALTER` or an `ALTER`'s `REVERT` with 'online' strategy.

## Related Issue(s)

- Tracking: #6926 

## Checklist
- [ ] Should this PR be backported?
- [ ] Tests were added or are not required
- [ ] Documentation was added or is not required

cc @aquarapid

PR #9649, opened 2022-02-08
BACKPORT: OnlineDDL: delete _vt.vreplication entry as part of CLEANUP operation

shlomi-noach:
Backport of https://github.com/vitessio/vitess/pull/9638/

## Description

Migration cleanup happens either on:

- explicit `ALTER VITESS_MIGRATION ... CLEANUP` query
- automated migration garbage collection

In either case, migration cleanup now also makes sure to delete any associated `vreplication` workflow used by the migration. There can only be one such workflow, and one exists when the migration is `ALTER` or an `ALTER`'s `REVERT` with 'online' strategy.

## Related Issue(s)

- Tracking: #6926 

## Checklist

cc @aquarapid

### DIFF
--- a/go/vt/vttablet/onlineddl/executor.go
+++ b/go/vt/vttablet/onlineddl/executor.go
@@ -616,29 +616,18 @@ func (e *Executor) terminateVReplMigration(ctx context.Context, uuid string) err
 	if err != nil {
 		return err
 	}
-	{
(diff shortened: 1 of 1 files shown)


shlomi-noach:
:wave:  call for review, if we want to back port this functionality

mattlord (approved):
Thank you! ❤️

PR #9651, opened 2022-02-08
mysql: allow parsing the 'info' field of MySQL packets

vmg:
## Description

MySQL query response packets (i.e. `Query_OK`) contain an `info` field with arbitrary, human-readable data from the MySQL server. Right now our client implementation ignores this data -- it is read of the wire but it's never returned to the caller in our `query.Results` struct, and nothing is done with it.

This PR adds a commandline flag to Vitess, `mysql_client_result_info`, that changes the behavior of the client to actually parse and return the `info` field. Once this field is enabled, `vttablet` instances will read this field during any requests done to the upstream MySQL server, and the field will be relayed all the way to the `vtgate`s if it's there.

This will allow us, further down the road, to add logic to the `vtgate`s that parses and aggregates these fields.

cc @piki @rbranson 

## Checklist


go/mysql/query.go
```diff
@@ -448,6 +451,7 @@ func (c *Conn) ReadQueryResult(maxrows int, wantfields bool) (*sqltypes.Result,
 				more = (packetOk.statusFlags & ServerMoreResultsExists) != 0
 				result.SessionStateChanges = packetOk.sessionStateData
 				result.StatusFlags = packetOk.statusFlags
+				result.Info = packetOk.info
```
deepthi, line 454: I would have expected the `if` condition here rather than in `parseOKPacket`.

go/mysql/conn.go
```diff
@@ -1451,7 +1451,9 @@ func (c *Conn) parseOKPacket(in []byte) (*PacketOK, error) {
 	if c.Capabilities&uint32(CapabilityClientSessionTrack) == CapabilityClientSessionTrack {
 		// info
 		info, _ := data.readLenEncInfo()
-		packetOK.info = info
```
deepthi, line 1454: Unrelated to this PR, this same code is duplicated in the `if` and `else`. It can be moved prior to the `if`.

go/mysql/query.go
vmg: I went back and forth with it. It felt wasteful to move the `info` into the `packetOK` struct if it's not going to be used anywhere.

deepthi: Is packetOk.info completely unused right now? In which case 👍

go/mysql/conn.go
```diff
@@ -1451,7 +1451,9 @@ func (c *Conn) parseOKPacket(in []byte) (*PacketOK, error) {
...
+		if *mysqlClientResultInfo {
+			packetOK.info = info
+		}
```
deepthi, line 1456: If we _are_ keeping this, might as well move it out of the if/else.

deepthi (commented):
I'm done with my review, but we need one from @vitessio/query-serving

systay:
it would be great with a couple of unit tests and/or e2e test

vmg:
PR is green and includes e2e tests (which were hard to write! not a lot of MySQL queries return an `info` packet).

systay (approved):
:+1:

PR #9654, opened 2022-02-08
[vtadmin-web] Upgrade stylelint + related dependencies to latest

doeg:
## Description

This updates stylelint from version 13.x to version 14, the latest version. [Here is the changelog](https://github.com/stylelint/stylelint/blob/main/CHANGELOG.md).

The only meaningful change for us is that must now explicitly include support for SCSS. I had to tweak a few of the rules when switching from `stylelint-config-standard` to `stylelint-config-standard-scss`. In particular, overriding the default rules for class- and property-patterns is required since SCSS seems to prefer kebab-case, whereas we use lowerCamelCase in order to import class names in .tsx files using SCSS modules.

## Related Issue(s)

N/A

## Checklist
- [x] Should this PR be backported? **Yes,** I'll make a backport PR tomorrow. 

## Deployment Notes

N/A

### DIFF
--- a/web/vtadmin/.stylelintrc
+++ b/web/vtadmin/.stylelintrc
@@ -4,9 +4,14 @@
 		"stylelint-config-prettier"
 	],
 	"rules": {
-		"at-rule-no-unknown": null,
+		"alpha-value-notation": "number",
+		"color-function-notation": "legacy",
 		"custom-property-pattern": null,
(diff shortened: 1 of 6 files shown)


notfelineit (approved):

doeg:
I'll create a backport for this one once https://github.com/vitessio/vitess/pull/9660 is in, since they both modify the package.json files.

PR #9655, opened 2022-02-09
Strip database qualifier from VIEW definitions in GetSchema

mattlord:
## Description
The [TabletManager `GetSchema` RPC call](https://github.com/vitessio/vitess/blob/2a4550de4652bb2f8e4a1b81b4a9a7b15976e19f/go/vt/vttablet/tabletmanager/rpc_schema.go#L32-L35) results for SQL Views have the database name replaced by the template notation `{{.DatabaseName}}` when what we should do instead is remove the database name qualifier so that `CREATE VIEW` clauses match `CREATE TABLE` clauses in the results — meaning that they do not have the database name qualifier.
> The difference in the  base output coming from MySQL is due to the difference between what's shown in the `SHOW CREATE VIEW` output and what's in `information_schema.views`.

<details>
<summary>You can see a demonstration here...</summary>

### Here's the basic structures created and viewed from a `vtgate`:
```
mysql> create table t1 (id int)

mysql> create view v1 as select id from t1;

mysql> show create table t1\G
       Table: t1
Create Table: CREATE TABLE `t1` (
  `id` int DEFAULT NULL
1 row in set (0.00 sec)

mysql> show create view v1\G
                View: v1
         Create View: CREATE ALGORITHM=UNDEFINED DEFINER=`root`@`localhost` SQL SECURITY DEFINER VIEW `v1` AS select `t1`.`id` AS `id` from `t1`
character_set_client: utf8mb4
collation_connection: utf8mb4_0900_ai_ci
1 row in set (0.00 sec)

mysql> select * from information_schema.views where table_name="v1"\G
       TABLE_CATALOG: def
        TABLE_SCHEMA: vt_customer
          TABLE_NAME: v1
     VIEW_DEFINITION: select `vt_customer`.`t1`.`id` AS `id` from `vt_customer`.`t1`
        CHECK_OPTION: NONE
        IS_UPDATABLE: YES
             DEFINER: vt_app@localhost
       SECURITY_TYPE: DEFINER
CHARACTER_SET_CLIENT: utf8
COLLATION_CONNECTION: utf8_general_ci
1 row in set (0.00 sec)
```

### Here's what we see in the Vitess schema:
**Before:**
```
$ vtctlclient GetSchema -include-views=true -tables "t1,v1" zone1-200 | jq -r '.table_definitions[].schema'
CREATE TABLE `t1` (
  `id` int(11) DEFAULT NULL
) ENGINE=InnoDB DEFAULT CHARSET=utf8
CREATE ALGORITHM=UNDEFINED DEFINER=`vt_app`@`localhost` SQL SECURITY DEFINER VIEW {{.DatabaseName}}.`v1` AS select {{.DatabaseName}}.`t1`.`id` AS `id` from {{.DatabaseName}}.`t1`
```

**After:**
```
$ vtctlclient GetSchema -include-views=true -tables "t1,v1" zone1-200 | jq -r '.table_definitions[].schema'
CREATE TABLE `t1` (
  `id` int(11) DEFAULT NULL
) ENGINE=InnoDB DEFAULT CHARSET=utf8
CREATE ALGORITHM=UNDEFINED DEFINER=`vt_app`@`localhost` SQL SECURITY DEFINER VIEW `v1` AS select `t1`.`id` AS `id` from `t1`
```
</details>

## Checklist
- [x] Test was added

### DIFF
--- a/go/vt/mysqlctl/schema.go
+++ b/go/vt/mysqlctl/schema.go
@@ -244,9 +244,11 @@ func (mysqld *Mysqld) normalizedSchema(ctx context.Context, dbName, tableName, t
 	norm := qr.Rows[0][1].ToString()
 	norm = autoIncr.ReplaceAllLiteralString(norm, "")
 	if tableType == tmutils.TableView {
-		// Views will have the dbname in there, replace it
-		// with {{.DatabaseName}}
-		norm = strings.Replace(norm, backtickDBName, "{{.DatabaseName}}", -1)
+		// Views will have the dbname qualifier in there, we remove it
+		// so that the results for Views match:
+		//   1. The SHOW CREATE VIEW output from MySQL
+		//   2. The CREATE TABLE output in these results, which does not have the DB name qualifier
+		norm = strings.Replace(norm, backtickDBName+".", "", -1)
 	}
 
 	return norm, nil
--- a/go/test/endtoend/tabletmanager/commands_test.go
+++ b/go/test/endtoend/tabletmanager/commands_test.go
@@ -25,11 +25,17 @@ import (
 	"time"
 
 	"github.com/stretchr/testify/require"
+	"github.com/tidwall/gjson"
 
 	"github.com/stretchr/testify/assert"
 
 	"vitess.io/vitess/go/mysql"
 	"vitess.io/vitess/go/test/endtoend/cluster"
+)
+
+var (
+	getSchemaT1Results = "CREATE TABLE `t1` (\n  `id` bigint(20) NOT NULL,\n  `value` varchar(16) DEFAULT NULL,\n  PRIMARY KEY (`id`)\n) ENGINE=InnoDB DEFAULT CHARSET=utf8"
+	getSchemaV1Results = fmt.Sprintf("CREATE ALGORITHM=UNDEFINED DEFINER=`%s`@`%s` SQL SECURITY DEFINER VIEW `v1` AS select `t1`.`id` AS `id`,`t1`.`value` AS `value` from `t1`", username, hostname)
 )
 
 // TabletCommands tests the basic tablet commands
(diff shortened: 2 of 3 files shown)


go/vt/mysqlctl/schema.go
```diff
@@ -244,9 +244,11 @@ func (mysqld *Mysqld) normalizedSchema(ctx context.Context, dbName, tableName, t
...
 	norm = autoIncr.ReplaceAllLiteralString(norm, "")
 	if tableType == tmutils.TableView {
-		// Views will have the dbname in there, replace it
-		// with {{.DatabaseName}}
-		norm = strings.Replace(norm, backtickDBName, "{{.DatabaseName}}", -1)
+		// Views will have the dbname qualifier in there, we remove it
+		// so that the results for Views match:
+		//   1. The SHOW CREATE VIEW output from MySQL
+		//   2. The CREATE TABLE output in these results, which does not have the DB name qualifier
+		norm = strings.Replace(norm, backtickDBName+".", "", -1)
```
shlomi-noach, line 251: I'm wary of `strings.Replace()` as means to manipulate queries. What happens if there's a column that has same name as the database? At the very least I'd say lets replace `-1` with `1`:

```suggestion
		norm = strings.Replace(norm, backtickDBName+".", "", 1)
```

but then, when I experiment on my local machine:
```sql
master [localhost:21324] {msandbox} (test) > show create table `test`.`v` \G
                View: v
         Create View: CREATE ALGORITHM=UNDEFINED DEFINER=`msandbox`@`localhost` SQL SECURITY DEFINER VIEW `v` AS select `t`.`id` AS `id`,`t`.`i` AS `i`,`t`.`tx` AS `tx` from `t`
character_set_client: utf8mb4
collation_connection: utf8mb4_0900_ai_ci
```
I don't see a database qualifier in the resulting `CREATE VIEW` statement. therefore, should there be a column named `test`, it will be wrongly replaced.

I realize the bug predates this PR. But then, I also see no testing to this logic. Should we seize the opportunity to fix this? We can e.g. `sqlparser.ParseStatement` into a `CreateView`,  then remove the schema qualifier, then export back via `sqlparser.String()`.

WDYT?

shlomi-noach (commented):
please see inline notes

go/vt/mysqlctl/schema.go
mattlord: This is about the `vtctl GetSchema` results and the RPC it uses. You can see an example if you exapand `Here's a demonstration:` in the description. 

I had similar concerns about the potential for problems, but since we already had that issue in this code I set it aside. I can explore alternatives though, I'm going to move this back to draft anyway as I'm adding a new e2e test for `GetSchema` in `vitess.io/vitess/go/test/endtoend/tabletmanager`.

shlomi-noach: got it. I actually missed the `Here's a demonstration:` section.

I agree that the solution in this PR is better than already existing code.

shlomi-noach (approved):

mattlord:
Closing this as it's not 100% safe to do within Vitess and it will instead be done at higher layers on the caller side when it's known to be safe. 

I moved the new test to its own PR: https://github.com/vitessio/vitess/pull/9658

PR #7380, opened 2021-01-26
Add vtgate stats for RowsAffected similar to RowsReturned

deepthi:
## Description
Add a Gauge called VtgateApiRowsAffected similar to VtgateApiRowsReturned. This will record the number of rows that were inserted/updated/deleted.
While adding this, I discovered that the computation of RowsAffected is incorrect. This PR also attempts to fix that.

## Related Issue(s)
#7372 
Fixes #7415 

## Checklist
- [ ] Should this PR be backported?
- [x] Tests were added or are not required
- [x] Documentation was added or is not required

## Impacted Areas in Vitess
Components that this PR will affect:

- [x]  Query Serving
- [ ]  VReplication
- [ ]  Cluster Management
- [ ]  Build 
- [ ]  VTAdmin

PR #9656, opened 2022-02-09
Add QueryRowsAffected and QueryRowsReturned to vttablet metrics and deprecate QueryRowCounts

GuptaManan100:
## Description

It was observed that the metric `QueryRowCounts` was returning zeroes for all the `SELECT` and `SHOW` plans. This was tracked down to change in behaviour due to #7380. In that PR, we changed how we read the results from MySQL. Earlier we used to set the RowsAffected field to the number of rows returned. However, this was not correct according to MySQL, which returned zeroes for all `SELECT` statements except for `SELECT .... INTO OUTFILE`. The `QueryRowCounts` metric was exposing this RowsAffected field.

So, after we changed the RowsAffected field to have zeroes for `SELECT` statements like MySQL, we inadvertently changed the behaviour of this metric. Instead of showing the number of rows returned, it now started showing the RowsAffected.

We have opted to deprecate `QueryRowCounts` metric because of its confusing name. 
Instead, we introduce 2 additional metrics `QueryRowsAffected` and `QueryRowsReturned` to show the number of rows affected and number of rows returned respectively.

Also, we will only track these metrics for the plans where they will produce non-zero values.
For example, an `INSERT` query type will not return any rows, so the metric QueryRowsReturned would always be zero for them. So we opt to not display them at all.

After these changes, running the local example with the sql commands in the common directory, the following metrics are produced - 

```json
"QueryRowCounts": {"corder.Insert": 5, "Join.Show": 0, "customer.Select": 0, "product.Select": 0, "corder.Select": 0, "dual.Select": 0, "customer.Insert": 5, "product.Insert": 2},
"QueryRowsAffected": {"customer.Insert": 5, "product.Insert": 2, "corder.Insert": 5},
"QueryRowsReturned": {"dual.Select": 1, "Join.Show": 3, "customer.Select": 5, "product.Select": 2, "corder.Select": 5},
```

## Checklist

## Deployment Notes

Mark `QueryRowCounts` for deprecation


harshit-gangal:
I think we should expose two different metric names, one for DML and the other for reads.

deepthi:
> I think we should expose two different metric names, one for DML and the other for reads.

Can you explain why you think this is needed?
The metric already has a plan dimension (and a table dimension)
```
# TYPE vttablet_query_row_counts counter
vttablet_query_row_counts{plan="Insert",table="corder"} 5
vttablet_query_row_counts{plan="Insert",table="customer"} 5
vttablet_query_row_counts{plan="Insert",table="product"} 2
vttablet_query_row_counts{plan="Select",table="corder"} 0
vttablet_query_row_counts{plan="Select",table="customer"} 0
vttablet_query_row_counts{plan="Select",table="dual"} 0
vttablet_query_row_counts{plan="Select",table="product"} 0
```

GuptaManan100:
@deepthi We have exposed two metrics now -
`QueryRowCounts` returns the number of rows returned for the plan - This is what it used to do pre - #7380
`QueryRowsAffected` shows the RowsAffected for the plan - This is a new metric, so that we preserve the DML statements rows affected post #7380

This is also now consistent with the VTGate metrics and per plan metrics that we publish which have the RowsAffected and RowsReturned as separate fields

GuptaManan100:
After this change the metrics look like - 
```
"QueryCounts": {"customer.Insert": 5, "product.Insert": 2, "corder.Insert": 1, "Join.Show": 1, "customer.Select": 1, "product.Select": 1, "corder.Select": 1, "dual.Select": 1},
"QueryErrorCounts": {"product.Insert": 0, "corder.Insert": 0, "Join.Show": 0, "customer.Select": 0, "product.Select": 0, "corder.Select": 0, "dual.Select": 0, "customer.Insert": 0},
"QueryRowCounts": {"corder.Select": 5, "dual.Select": 1, "customer.Insert": 0, "product.Insert": 0, "corder.Insert": 0, "Join.Show": 3, "customer.Select": 5, "product.Select": 2},
"QueryRowsAffected": {"corder.Insert": 5, "Join.Show": 0, "customer.Select": 0, "product.Select": 0, "corder.Select": 0, "dual.Select": 0, "customer.Insert": 5, "product.Insert": 2},
```

The per-query stats look like -
```
{
  "Query": "select * from customer",
  "Table": "customer",
  "Plan": "Select",
  "QueryCount": 1,
  "Time": 479847,
  "MysqlTime": 344808,
  "RowsAffected": 0,
  "RowsReturned": 5,
  "ErrorCount": 0
},
{
  "Query": "insert into product(sku, description, price) values (:vtg1, :vtg2, :vtg3)",
  "Table": "product",
  "Plan": "Insert",
  "QueryCount": 2,
  "Time": 1876259,
  "MysqlTime": 1263828,
  "RowsAffected": 2,
  "RowsReturned": 0,
  "ErrorCount": 0
}
```

VTGate metric - 
```
"VtgateApiRowsAffected": {"Execute.commerce.primary": 12},
"VtgateApiRowsReturned": {"Execute.commerce.primary": 22},
```

harshit-gangal (approved):

deepthi (dismissed):
This PR is not backwards-compatible. Blocking the merge until we can resolve this.

harshit-gangal:
> > I think we should expose two different metric names, one for DML and the other for reads.
> 
> Can you explain why you think this is needed? The metric already has a plan dimension (and a table dimension)
> 
> ```
> # TYPE vttablet_query_row_counts counter
> vttablet_query_row_counts{plan="Insert",table="corder"} 5
> vttablet_query_row_counts{plan="Insert",table="customer"} 5
> vttablet_query_row_counts{plan="Insert",table="product"} 2
> vttablet_query_row_counts{plan="Select",table="corder"} 0
> vttablet_query_row_counts{plan="Select",table="customer"} 0
> vttablet_query_row_counts{plan="Select",table="dual"} 0
> vttablet_query_row_counts{plan="Select",table="product"} 0
> ```

The metric name does not give the right impression as what it is supposed to mean for DML. This also aligns us with the metric names in vtgate.

harshit-gangal:
we can create 2 new metrics RowsReturned and RowsAffected and deprecate this metric. In this PR we can fix the numbers for all three.

deepthi (commented):
Can you review the unit test coverage for these metrics and add whatever is needed?
There are no changes to unit tests which tells me that the original metric never had any. We should fix that.

go/vt/vttablet/tabletserver/query_executor.go
```diff
@@ -112,11 +112,12 @@ func (qre *QueryExecutor) Execute() (reply *sqltypes.Result, err error) {
...
 
 		if reply == nil {
-			qre.tsv.qe.AddStats(planName, tableName, 1, duration, mysqlTime, 0, 1)
+			qre.tsv.qe.AddStats(planName, tableName, 1, duration, mysqlTime, 0, 0, 1)
 			qre.plan.AddStats(1, duration, mysqlTime, 0, 0, 1)
 			return
 		}
-		qre.tsv.qe.AddStats(planName, tableName, 1, duration, mysqlTime, int64(reply.RowsAffected), 0)
+
+		qre.tsv.qe.AddStats(planName, tableName, 1, duration, mysqlTime, int64(len(reply.Rows)), int64(reply.RowsAffected), 0)
```
JTCunning, line 120: Since we're changing the existing metric from `reply.RowsAffected` to `reply.Rows`, will this mean that all `DELETE` plans will now produce zeroes instead of previously producing the amount of rows that were deleted?

deepthi: We'll rename QueryRowCounts to QueryRowsReturned (or something like that) and **exclude** irrelevant plan types.
So there will be no entries for Delete in that metric at all. And we'll stop emitting Select type metrics in the RowsAffected metric.

deepthi: when we discussed this earlier today, @harshit-gangal mentioned that metrics have been changed in the past without deprecation. I believe though that is was only changes to plan type names - a dimension of the metric, not the metric name itself.
To make it easier for users to upgrade let us add a new metric called `QueryRowsReturned` and deprecate the old one. We can schedule it for deletion in 15.0

GuptaManan100:
After discussing this, we reached the consensus that we want to deprecate `QueryRowCounts` since the name of the metric isn't too great. Instead we have opted to add 2 new metrics `QueryRowsAffected` and `QueryRowsReturned`. 

Like the name suggests `QueryRowsAffected` will track the rows affected for the given plan type and `QueryRowsReturned` will track the number of rows returned. Also, we will only track these metrics for the plans where they will produce non-zero values. 

For example, an `INSERT` query type will not return any rows, so the metric `QueryRowsReturned` would always be zero for them. So we opt to not display them at all.

GuptaManan100:
I have updated the description with the lates information and metric outputs

harshit-gangal (approved):

go/vt/vttablet/endtoend/misc_test.go
```diff
@@ -467,14 +467,41 @@ func TestQueryStats(t *testing.T) {
...
 		ErrorCount:   1,
 	}
-	if stat != want {
-		t.Errorf("stat: %+v, want %+v", stat, want)
-	}
+	utils.MustMatch(t, want, stat)
 	vend := framework.DebugVars()
 	compareIntDiff(t, vend, "QueryCounts/vitess_a.Select", vstart, 2)
 	compareIntDiff(t, vend, "QueryRowCounts/vitess_a.Select", vstart, 0)
+	compareIntDiff(t, vend, "QueryRowsReturned/vitess_a.Select", vstart, 2)
```
deepthi, line 474: For completeness we should assert that the `QueryRowsAffected` stat does not change for a select.

go/vt/vttablet/endtoend/misc_test.go
```diff
@@ -467,14 +467,41 @@ func TestQueryStats(t *testing.T) {
...
+		RowsAffected: 2,
+		RowsReturned: 0,
+		ErrorCount:   0,
+	}
+	utils.MustMatch(t, want, stat)
+	vend = framework.DebugVars()
+	compareIntDiff(t, vend, "QueryCounts/vitess_a.UpdateLimit", vstart, 1)
+	compareIntDiff(t, vend, "QueryRowCounts/vitess_a.UpdateLimit", vstart, 2)
+	compareIntDiff(t, vend, "QueryRowsAffected/vitess_a.UpdateLimit", vstart, 2)
+	compareIntDiff(t, vend, "QueryErrorCounts/vitess_a.UpdateLimit", vstart, 0)
```
deepthi, line 503: Similarly, we should assert here that the `QueryRowCounts` stat doesn't change.
Ideally we should also be testing `Delete` and `Insert` plans.

deepthi:
Can you also update the title to reflect the latest implementation?

go/vt/vttablet/endtoend/misc_test.go
GuptaManan100: Well, that stat will not be produced at all for `Select` query type

go/vt/vttablet/endtoend/misc_test.go
GuptaManan100: I'll add tests for `Delete` and `Insert`. Its the same thing, `QueryRowCounts` will not be available for these types. I'll assert that it doesn't exist

GuptaManan100: done

PR #9657, opened 2022-02-09
Use the gen4 planner for queries with outer joins

### LINKED ISSUE #7700: left join query does not produce correct results when filtering on sharding key with `is null`
#### Overview of the Issue

Filtering on `Left join` seems broken when using `where sharding_key is null`

#### Reproduction Steps

```
-- create two tables 
-- Note: company.id is the sharding key

create table company (id int, name varchar(255));
create table user (id int, companyID int, email varchar(255));

-- insert some data
insert into company(id, name) values (1, 'company_1'), (2, 'company_2');
insert into user(id, companyID, email) values (1, 1, 'user@valid.com'), (2, 100, 'user@invalid.com');
select * from company;
+------+-----------+
| id   | name      |
+------+-----------+
|    1 | company_1 |
|    2 | company_2 |
+------+-----------+

select * from user;
+------+-----------+------------------+
| id   | companyID | email            |
+------+-----------+------------------+
|    1 |         1 | user@valid.com   |
|    2 |       100 | user@invalid.com |
+------+-----------+------------------+


-- this query is supposed to return one row but it does not on Vitess
SELECT u.id, u.email, u.companyID, c.id AS cpID
FROM 
  user u   
  LEFT JOIN company c ON u.companyID = c.id
WHERE 
c.id is null;
+------+------------------+-----------+------+
| id   | email            | companyID | cpID |
+------+------------------+-----------+------+
|    2 | user@invalid.com |       100 | NULL |
+------+------------------+-----------+------+

```

#### Operating system and Environment details

OS, Architecture, and any other information you can provide
about the environment.

- vitess running on GKE

#### Log Fragments

N/A

systay:
## Description
Routing for queries that contain outer join was not being handled correctly.
It would require a lot of changes to support these queries correctly in V3, so instead this fix makes it so the gen4 planner is used in the presence of outer joins.

If you know that you want to use the v3 planner, that can still be accomplished by using the query hint `/*vt+ PLANNER=v3 */`

## Related Issue(s)
Fixes #7700

### DIFF
--- a/go/vt/vtexplain/testdata/multi-output/selectsharded-output.txt
+++ b/go/vt/vtexplain/testdata/multi-output/selectsharded-output.txt
@@ -191,13 +191,13 @@ select id from user where not id in (select col from music where music.user_id =
 ----------------------------------------------------------------------
 SELECT user.id, user.name, name_info.info FROM user INNER JOIN music ON (user.id = music.user_id) LEFT OUTER JOIN name_info ON (user.name = name_info.name)
 
-1 ks_sharded/-40: select `user`.id, `user`.`name` from `user` join music on `user`.id = music.user_id limit 10001
-1 ks_sharded/40-80: select `user`.id, `user`.`name` from `user` join music on `user`.id = music.user_id limit 10001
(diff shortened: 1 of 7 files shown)


systay:
Ping @arthurschreiber

PR #9659, opened 2022-02-09
Revocation phase using durability policies

GuptaManan100:
## Description

This PR adds the ability to check for revocation using the durability policies in `EmergencyReparentShard`. What that means is that instead of exiting incase of more than 1 failure, `ERS` will first try and find whether it is safe to proceed even when we have multiple failures! 
So, if the durability_policies allow for more than 1 failure, ERS will succeed!

Consider an example for the same. Let us say we are using `semi_sync` as our durability policy and we have four servers, A, B, C and D. A is our current primary, B and C are replicas and D is a rdonly tablet.
Lets say that A and D, both go down.

In the older version, ERS would have failed since there are more than 1 failures
But now, ERS will see that even with A and D down, we can proceed safely. Safety, here means that we can guarantee that the unreachable (equivalent to down) tablets cannot make forward progress. In this case, only A is eligible to be a primary and D cannot provide Semi_Sync ACKs, which means that A and D cannot make progress, so it is safe to continue with the ERS.

## Related Issue(s)

- #8975 

## Checklist


go/test/endtoend/reparent/utils/utils.go
```diff
@@ -172,6 +172,7 @@ func setupCluster(ctx context.Context, t *testing.T, shardName string, cells []s
 	// Wait for mysql processes to start
 	for _, proc := range mysqlCtlProcessList {
 		if err := proc.Wait(); err != nil {
+			clusterInstance.PrintMysqlctlLogFiles()
 			t.Fatalf("Error starting mysql: %s", err.Error())
```
deepthi, line 176: Can we replace this with `require.Fail` (or whatever the right equivalent is)?

go/vt/vtctl/reparentutil/durability.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2021 The Vitess Authors.
+Copyright 2022 The Vitess Authors.
```
deepthi, line 2: Until we decide what to do about the copyright, we should not be making any changes.
Technically, it is incorrect to change the copyright year.

go/vt/vtctl/reparentutil/durability_policies.go
```diff
@@ -0,0 +1,238 @@
...
+	return promotionrule.MustNot
+}
+
+func (d *durabilitySemiSync) semiSyncAckers(tablet *topodatapb.Tablet) int {
+	return 1
+}
+
+func (d *durabilitySemiSync) isReplicaSemiSync(primary, replica *topodatapb.Tablet) bool {
+	switch replica.Type {
+	case topodatapb.TabletType_PRIMARY, topodatapb.TabletType_REPLICA:
```
deepthi, line 153: Why would you ever pass a PRIMARY as `replica` to this function?

go/vt/vtctl/reparentutil/durability_policies.go
```diff
@@ -0,0 +1,238 @@
...
+	return promotionrule.MustNot
+}
+
+func (d *durabilityCrossCell) semiSyncAckers(tablet *topodatapb.Tablet) int {
+	return 1
+}
+
+func (d *durabilityCrossCell) isReplicaSemiSync(primary, replica *topodatapb.Tablet) bool {
+	// Prevent panics.
+	if primary.Alias == nil || replica.Alias == nil {
```
deepthi, line 180: Don't we need this check in `durabilitySemiSync` as well?

go/vt/vtctl/reparentutil/durability_policies.go
```diff
@@ -0,0 +1,238 @@
...
+		return false
+	}
+	switch replica.Type {
+	case topodatapb.TabletType_PRIMARY, topodatapb.TabletType_REPLICA:
```
deepthi, line 184: Same question as above about `PRIMARY`.

go/vt/vtctl/reparentutil/durability.go
```diff
@@ -17,222 +17,55 @@ limitations under the License.
...
-
-func (d *durabilitySemiSync) semiSyncAckers(tablet *topodatapb.Tablet) int {
-	return 1
+	return
 }
 
-func (d *durabilitySemiSync) isReplicaSemiSync(primary, replica *topodatapb.Tablet) bool {
-	switch replica.Type {
-	case topodatapb.TabletType_PRIMARY, topodatapb.TabletType_REPLICA:
+// RevokeForTablet checks whether we have reached enough tablets such that the given primary capable tablet cannot accept any new write
```
deepthi, line 38:
```suggestion
// RevokeForTablet checks whether we have reached enough tablets such that the given primary eligible tablet cannot accept any new writes
```

go/vt/vtctl/reparentutil/replication.go
```diff
@@ -239,7 +244,12 @@ func StopReplicationAndBuildStatusMaps(
...
 
 	errRecorder := errgroup.Wait(groupCancel, errChan)
-	if len(errRecorder.Errors) > 1 {
+	if len(errRecorder.Errors) <= 1 {
+		return statusMap, primaryStatusMap, nil
+	}
+	// check that the tablets we were able to reach are sufficient for us to guarantee that no new write will be accepted by any tablet
+	revokeSuccessful := Revoked(tabletsReachable, allTablets)
+	if !revokeSuccessful {
 		return nil, nil, vterrors.Wrapf(errRecorder.Error(), "encountered more than one error when trying to stop replication and get positions: %v", errRecorder.Error())
```
deepthi, line 253: You probably need to change this error message.

go/vt/vtctl/reparentutil/replication.go
```diff
@@ -171,6 +171,8 @@ func StopReplicationAndBuildStatusMaps(
 		primaryStatusMap = map[string]*replicationdatapb.PrimaryStatus{}
 		m                sync.Mutex
 		errChan          = make(chan concurrency.Error)
+		tabletsReachable []*topodatapb.Tablet
+		allTablets       []*topodatapb.Tablet
 	)
 
 	groupCtx, groupCancel := context.WithTimeout(ctx, waitReplicasTimeout)
```
deepthi, line 178: What is the full timeout for ERS? what do we expect the typical value for `waitReplicasTimeout` to be?

deepthi (commented):
Comments from a first pass. I still need to understand the new Revoke funcs.

go/vt/vtctl/reparentutil/durability.go
GuptaManan100: This is an unfortunate incorrect diff from GitHub...
I actually renamed the `durability.go` file to `durability_policies.go` since that file had the implementations of the durability policies.
Then later I add `durability.go` for having functions with respect to durability like `Revoked`.

So the file `durability_policies.go` has the year 2021, since it is unchanged except for rename. And `durability.go` now holds new code all introduced in this year.

go/vt/vtctl/reparentutil/durability_policies.go
GuptaManan100: There are two reasons why that would happen - 
1. As a fail safe for, if we have multiple tablets thinking themselves to be primary
2. We want to know whether the current primary will be a sending semi sync acks for a replica (which is a potential primary at this point). So the `replica` argument can be a tablet of type Primary.

go/vt/vtctl/reparentutil/durability_policies.go
GuptaManan100: I'll move that check to the Exported function which calls these methods, so that they can always assume that the provided values are non-negative

go/vt/vtctl/reparentutil/replication.go
GuptaManan100: Both are configurable, if you call ERS from the vtctlclient, then the default action_timeout (total ERS time) is 1 hour and `waitReplicasTimeout` is also configurable which defaults to 30 seconds.
However, if we want ERS to be able to proceed with more than 1 failures, we need the action_timeout to be more than wait_replicas_timeout.

go/test/endtoend/reparent/utils/utils.go
GuptaManan100: Done

go/vt/vtctl/reparentutil/durability.go
GuptaManan100: Done

go/vt/vtctl/reparentutil/replication.go
GuptaManan100: Done

GuptaManan100:
@deepthi I have renamed the files back so it is easier to review and added the requested comments in the `durabler` interface.

go/vt/vtctl/reparentutil/durability.go
```diff
@@ -61,8 +61,12 @@ func init() {
 
 // durabler is the interface which is used to get the promotion rules for candidates and the semi sync setup
 type durabler interface {
+	// promotionRule represents the precedence in which we want to tablets to be promoted.
+	// The lower the promotion rule of a tablet, the more we want it to be promoted in case of a failover
```
deepthi, line 65: Isn't this supposed to be higher, not lower?

go/vt/vtctl/reparentutil/durability.go
```diff
@@ -61,8 +61,12 @@ func init() {
...
 	promotionRule(*topodatapb.Tablet) promotionrule.CandidatePromotionRule
+	// semiSyncAckers represents the number of semi-sync ackers for a given tablet are required if it were to become the PRIMARY instance
```
deepthi, line 67:
```suggestion
	// semiSyncAckers represents the number of semi-sync ackers required for a given tablet if it were to become the PRIMARY instance
```

go/vt/vtctl/reparentutil/durability.go
```diff
@@ -61,8 +61,12 @@ func init() {
...
 	semiSyncAckers(*topodatapb.Tablet) int
+	// isReplicaSemiSync returns whether the replica should send semi-sync acks if primary were to become the PRIMARY instance
```
deepthi, line 69:
```suggestion
	// isReplicaSemiSync returns whether the "replica" should send semi-sync acks if "primary" were to become the PRIMARY instance
```

go/vt/vtctl/reparentutil/durability_funcs.go
```diff
@@ -0,0 +1,71 @@
...
+			semiSyncAckers = append(semiSyncAckers, tablet)
+		}
+	}
+	return
+}
+
+// RevokeForTablet checks whether we have reached enough tablets such that the given primary eligible tablet cannot accept any new writes
+func RevokeForTablet(primaryEligible *topodatapb.Tablet, tabletsReached []*topodatapb.Tablet, allTablets []*topodatapb.Tablet) bool {
+	// if we have reached the primaryEligible tablet and stopped its replication, then it will not
+	// accept any new writes
```
deepthi, line 41: Is this correct? Just because replication is stopped does not mean it can't accept writes, that is only true when semi-sync is enabled.

go/vt/vtctl/reparentutil/replication.go
```diff
@@ -222,6 +226,7 @@ func StopReplicationAndBuildStatusMaps(
 		tabletAliasToWaitFor = topoproto.TabletAliasString(tabletToWaitFor)
 	}
 	for alias, tabletInfo := range tabletMap {
+		allTablets = append(allTablets, tabletInfo.Tablet)
 		if !ignoredTablets.Has(alias) {
```
deepthi, line 230: How does this work with `ignoredTablets`? If we allow 1 failure but `ignoredTablets` is not empty, what happens?

go/vt/vtctl/reparentutil/durability.go
GuptaManan100: yes you are right, what was I thinking when I typed it

go/vt/vtctl/reparentutil/durability_funcs.go
GuptaManan100: For all tablets in this list that think they are Primary type, in `StopReplicationAndBuildStatusMaps`, we call `demotePrimary` on them, which succeeds only after setting the tablet to rdonly. So all the tablets in this list would not have just stopped replication but would also be read only, so just checking that the primary tablet is in the list is sufficient

go/vt/vtctl/reparentutil/durability.go
GuptaManan100: fixed

go/vt/vtctl/reparentutil/durability.go
GuptaManan100: fixed

go/vt/vtctl/reparentutil/durability.go
GuptaManan100: fixed

go/vt/vtctl/reparentutil/replication.go
GuptaManan100: IgnoredTablets functionality remains the same. We used to proceed with 1 failure in non-ignored tablets earlier which we will continue to do.
For the revocation check, we use the entire list of tablets to check for revocation to be sure that we only proceed with multiple failures if we can guarantee safety.

go/vt/topo/topoproto/tablet.go
```diff
@@ -63,6 +63,16 @@ func TabletAliasEqual(left, right *topodatapb.TabletAlias) bool {
 	return proto.Equal(left, right)
 }
 
+// TabletInList returns true if the tablet is in the list of tablets given
```
ajm188, line 66: Suggestion: rename to `IsTabletInList` to match `func IsTypeInList(tabletType topodatapb.TabletType, types []topodatapb.TabletType) bool {`

go/vt/topo/topoproto/tablet_test.go
```diff
@@ -70,3 +72,54 @@ func TestParseTabletAlias(t *testing.T) {
...
+			isInList:   true,
+		}, {
+			name:       "tablet not in list",
+			tablet:     tablet1,
+			allTablets: []*topodatapb.Tablet{tablet2, tablet3},
+			isInList:   false,
+		},
+	}
+
+	for _, testcase := range testcases {
```
ajm188, line 119: suggestion -- run these test cases in parallel with:

```go
func TestBlahBlah(t *testing.T) {
    t.Parallel()

    for _, testcase := range testcases {
        testcase := testcase // prevent loop shadowing
        t.Run(testcase.name, func(t *testing.T) {
            t.Parallel()
            // rest of test
        })
    }
}
```

go/vt/topo/topoproto/tablet_test.go
```diff
@@ -70,3 +72,54 @@ func TestParseTabletAlias(t *testing.T) {
...
+		t.Run(testcase.name, func(t *testing.T) {
+			out := TabletInList(testcase.tablet, testcase.allTablets)
+			require.Equal(t, testcase.isInList, out)
```
ajm188, line 122: I think you want `assert.Equal` here

go/vt/vtctl/reparentutil/durability_funcs.go
```diff
@@ -0,0 +1,71 @@
...
+		}
+		if IsReplicaSemiSync(primary, tablet) {
+			semiSyncAckers = append(semiSyncAckers, tablet)
+		}
+	}
+	return
+}
+
+// RevokeForTablet checks whether we have reached enough tablets such that the given primary eligible tablet cannot accept any new writes
+func RevokeForTablet(primaryEligible *topodatapb.Tablet, tabletsReached []*topodatapb.Tablet, allTablets []*topodatapb.Tablet) bool {
```
ajm188, line 39: for these functions that check and return bools, i'd suggest renaming to from `<Thing>` to `Is<Thing>` to make them read a little nicer where they are used

go/vt/vtctl/reparentutil/durability_funcs.go
```diff
@@ -0,0 +1,71 @@
...
+	// numOfSemiSyncAcksRequired is the number of semi sync Acks that the primaryEligible tablet requires
+	numOfSemiSyncAcksRequired := SemiSyncAckers(primaryEligible)
+
+	// if we have reached enough semi-sync Acking tablets such that the primaryEligible cannot accept a write
+	// we have revoked from the tablet
+	return len(allSemiSyncAckers)-len(semiSyncAckersReached) < numOfSemiSyncAcksRequired
+}
+
+// Revoked checks whether we have reached enough tablets to guarantee that no tablet eligible to become a primary can accept any write
+func Revoked(tabletsReached []*topodatapb.Tablet, allTablets []*topodatapb.Tablet) bool {
```
ajm188, line 61: same naming suggestion here

ajm188 (commented):
just some naming/testing suggestions! i think the rest is looking good :)

go/vt/vtctl/reparentutil/durability_funcs.go
deepthi: Hmm.. so usage of this function is dependent on some pre-condition being satisfied.
Is there a reason it has to be a separate function in a separate file? If it is being used only once, it can as well be just a code block, and then you don't have to worry about whether callers are satisfying the pre-condition.
If it _MUST_ be in a separate file for some good reason, you should make these two `Revoke` functions unexpected, so that they cannot be used outside the package. `s/Revoke/revoke`.

GuptaManan100: Yes that's a good point. The only reason that I have kept them as separate function is for testing and readability. It makes it much easier to have unit tests for these individual functions along with unit tests for the larger StopReplicationAndBuildStatusMap since that requires a larger setup with a tablet manager etc. 
Just to be extra careful, I'll add a comment as well there stating the assumption that the function has and unexport it

go/vt/vtctl/reparentutil/durability_funcs.go
GuptaManan100: How do you feel about, `haveRevokedForTablet`? instead of `isRevokedForTablet`?

go/vt/vtctl/reparentutil/durability_funcs.go
GuptaManan100: Fixed

go/vt/topo/topoproto/tablet.go
GuptaManan100: Fixed

go/vt/topo/topoproto/tablet_test.go
GuptaManan100: Fixed

go/vt/topo/topoproto/tablet_test.go
GuptaManan100: Fixed

go/vt/vtctl/reparentutil/durability_funcs.go
GuptaManan100: I have made the change for `haveRevokedForTablet`. If you feel the other name is better, I'll rename it

go/vt/vtctl/reparentutil/durability_funcs.go
GuptaManan100: Changed it to `haveRevoked`

GuptaManan100:
Thankyou @deepthi  and @ajm188  for the reviews! I have addressed all the review comments, could you take a look once, again?

go/vt/vtctl/reparentutil/durability_funcs.go
ajm188: fine by me! as long as it implies the boolean/question nature, and we're consistent with what we use for both, i'm happy :)

ajm188:
lgtm, but let's get deepthi's approval (and clean up the commit history length) before merging

deepthi (approved):

deepthi:
> lgtm, but let's get deepthi's approval (and clean up the commit history length) before merging

Do you mean squash some of the commits together?
If so, I recommend squashing only the last 5 (since we started review) either into one, or into whatever is the most relevant older commit. Typically we don't squash at all (though I'm well aware that this makes certain things like reverts and bisects that much harder).

Issue #6306, opened 2020-06-12
Help adding parser support for stream


derekperkins:
Currently the only way to subscribe to messages via SQL is to run `stream * from msg_tbl`. It seems like it should be trivial to add support to the parser for column selection, `stream col1, col2 from msg_tbl` for someone with experience with yacc. Once there is parser support, I can make sure that it plumbs through messages properly. cc @systay 
https://github.com/vitessio/vitess/blob/master/go/vt/sqlparser/sql.y#L413-L417

#5882

nstandif:
`select_expression` was used instead of `select_expression_list` in both the [AST ](https://github.com/vitessio/vitess/blob/master/go/vt/sqlparser/ast.go#L85)and the yacc. You'd need to change both in order to plumb it out correctly.

PR #9666, opened 2022-02-10
sqlparser: support stream specific columns

derekperkins:
## Description
Enable selecting specific columns instead of only `stream * from tbl`

## Related Issue(s)
https://github.com/vitessio/vitess/issues/6306

## Checklist
- [x] Should this PR be backported? - hoping to get it into v13


GuptaManan100:
Yes, there is a circular dependency, unfortunately right now there is no good way to fix this other than delete those lines from the files (yes the ones which will be generated again and overwritten).

GuptaManan100 (approved):
LGTM!

derekperkins:
Digging deeper, this doesn't provide much of a win because the whole message row is still loaded into memory, as the message manager can't guarantee that nobody will call `stream *`. I'll revisit this later with another proposal, probably something around setting visible/invisible columns in the table comments.

derekperkins:
Thanks to @GuptaManan100 and @systay for helping me debug the ast setup

PR #9669, opened 2022-02-10
Fixes #8277: vtctld logs leak pii http headers

### LINKED ISSUE #8277: vtctld should sanitize HTTP requests before logging to avoid logging sensitive headers
#### Overview of the Issue

It looks like the vtctld's long polling implementation logs HTTP requests without sanitizing headers. As a result, sensitive information like cookies can turn up in the logs.

Here's a specific example of this: https://github.com/vitessio/vitess/blob/c9b6d60/go/vt/workflow/long_polling.go#L160-L164

```go
func httpErrorf(w http.ResponseWriter, r *http.Request, format string, args ...interface{}) {
	errMsg := fmt.Sprintf(format, args...)
	log.Errorf("HTTP error on %v: %v, request: %#v", r.URL.Path, errMsg, r)
	http.Error(w, errMsg, http.StatusInternalServerError)
}
```

This is the only instance we've seen, but it'd be worth it to do a quick audit to see if there are any other places this should be patched.

A straightforward approach is to log all header keys, but omit the values for the "known sensitive" headers (`Authorization`, `Cookie`). For flexibility, we could consider adding a build flag to disable this behaviour (but I'm not sure if this is necessary...?)

vmogilev:
## Description

This PR fixes vtctld logging which is currently leaking http header values that can contain PII.

## Related Issue(s)

Fixes #8277

## Checklist

### DIFF
--- a/go/vt/workflow/long_polling.go
+++ b/go/vt/workflow/long_polling.go
@@ -28,8 +28,6 @@ import (
 	"sync"
 	"time"
 
-	"github.com/golang/glog"
-
 	"vitess.io/vitess/go/acl"
 	"vitess.io/vitess/go/timer"
 	"vitess.io/vitess/go/vt/log"
@@ -158,12 +156,9 @@ func getID(url, base string) (int, error) {
 	return i, nil
 }
 
-const (
-	LogLevelDumpAllHTTPHeaders glog.Level = 50
-	HTTPHeaderRedactedMessage  string     = "*** redacted by sanitizeRequestHeader() ***"
-)
+const HTTPHeaderRedactedMessage string = "*** redacted by sanitizeRequestHeader() ***"
 
-var whiteListedHTTPHeaders = map[string]interface{}{
+var allowedHTTPHeadersList = map[string]interface{}{
 	"Accept":                    nil,
 	"Accept-Encoding":           nil,
 	"Accept-Language":           nil,
@@ -184,7 +179,7 @@ func sanitizeRequestHeader(r *http.Request, debugOn bool) *http.Request {
 	s := r.Clone(r.Context())
 
 	for k := range s.Header {
-		if _, ok := whiteListedHTTPHeaders[k]; !ok {
+		if _, ok := allowedHTTPHeadersList[k]; !ok {
 			s.Header.Set(k, HTTPHeaderRedactedMessage)
 		}
 	}
@@ -192,32 +187,31 @@ func sanitizeRequestHeader(r *http.Request, debugOn bool) *http.Request {
 	return s
 }
 
-func httpErrorf(w http.ResponseWriter, r *http.Request, format string, args ...interface{}) {
+func (m *Manager) httpErrorf(w http.ResponseWriter, r *http.Request, format string, args ...interface{}) {
 	errMsg := fmt.Sprintf(format, args...)
-	debugOn := bool(log.V(LogLevelDumpAllHTTPHeaders))
 	log.Errorf("HTTP error on %v: %v, request: %#v",
 		r.URL.Path,
 		errMsg,
-		sanitizeRequestHeader(r, debugOn),
+		sanitizeRequestHeader(r, m.debugHTTPHeaders),
 	)
 	http.Error(w, errMsg, http.StatusInternalServerError)
 }
 
-func handleAPI(pattern string, handlerFunc func(w http.ResponseWriter, r *http.Request) error) {
+func (m *Manager) handleAPI(pattern string, handlerFunc func(w http.ResponseWriter, r *http.Request) error) {
 	http.HandleFunc(pattern, func(w http.ResponseWriter, r *http.Request) {
 		defer func() {
 			if x := recover(); x != nil {
-				httpErrorf(w, r, "uncaught panic: %v", x)
+				m.httpErrorf(w, r, "uncaught panic: %v", x)
 			}
 		}()
 
 		if err := acl.CheckAccessHTTP(r, acl.ADMIN); err != nil {
-			httpErrorf(w, r, "WorkflowManager acl.CheckAccessHTTP failed: %v", err)
+			m.httpErrorf(w, r, "WorkflowManager acl.CheckAccessHTTP failed: %v", err)
 			return
 		}
 
 		if err := handlerFunc(w, r); err != nil {
-			httpErrorf(w, r, "%v", err)
+			m.httpErrorf(w, r, "%v", err)
 		}
 	})
 }
@@ -227,7 +221,7 @@ func (m *Manager) HandleHTTPLongPolling(pattern string) {
 	log.Infof("workflow Manager listening to web traffic at %v/{create,poll,delete}", pattern)
 	lpm := newLongPollingManager(m)
 
-	handleAPI(pattern+"/create", func(w http.ResponseWriter, r *http.Request) error {
+	m.handleAPI(pattern+"/create", func(w http.ResponseWriter, r *http.Request) error {
 		result, i, err := lpm.create(r.URL)
 		if err != nil {
 			return fmt.Errorf("longPollingManager.create failed: %v", err)
@@ -241,7 +235,7 @@ func (m *Manager) HandleHTTPLongPolling(pattern string) {
 		return nil
 	})
 
-	handleAPI(pattern+"/poll/", func(w http.ResponseWriter, r *http.Request) error {
+	m.handleAPI(pattern+"/poll/", func(w http.ResponseWriter, r *http.Request) error {
 		i, err := getID(r.URL.Path, pattern+"/poll/")
 		if err != nil {
 			return err
@@ -260,7 +254,7 @@ func (m *Manager) HandleHTTPLongPolling(pattern string) {
 		return nil
 	})
 
-	handleAPI(pattern+"/remove/", func(w http.ResponseWriter, r *http.Request) error {
+	m.handleAPI(pattern+"/remove/", func(w http.ResponseWriter, r *http.Request) error {
 		i, err := getID(r.URL.Path, pattern+"/remove/")
 		if err != nil {
 			return err
@@ -271,7 +265,7 @@ func (m *Manager) HandleHTTPLongPolling(pattern string) {
 		return nil
 	})
 
-	handleAPI(pattern+"/action/", func(w http.ResponseWriter, r *http.Request) error {
+	m.handleAPI(pattern+"/action/", func(w http.ResponseWriter, r *http.Request) error {
 		_, err := getID(r.URL.Path, pattern+"/action/")
 		if err != nil {
 			return err
--- a/go/vt/workflow/long_polling.go
+++ b/go/vt/workflow/long_polling.go
@@ -17,6 +17,7 @@ limitations under the License.
 package workflow
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -27,7 +28,7 @@ import (
 	"sync"
 	"time"
 
-	"context"
+	"github.com/golang/glog"
 
 	"vitess.io/vitess/go/acl"
 	"vitess.io/vitess/go/timer"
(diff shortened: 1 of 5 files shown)


go/vt/workflow/long_polling.go
```diff
@@ -157,9 +158,48 @@ func getID(url, base string) (int, error) {
 	return i, nil
 }
 
+const (
+	LogLevelDumpAllHTTPHeaders glog.Level = 50
```
doeg, line 162: Ohh, using the loglevel to determine behavior is very elegant! 

Another way to do this that you may have already considered is to add a flag. Recently, I came across https://github.com/vitessio/vitess/pull/9550 by @mattlord, which is similar in spirit in that it redacts sensitive information from the logs (bind variables, in this particular case). That PR introduces the [`sanitize_log_messages` flag](https://github.com/vitessio/vitess/pull/9550/files#diff-ee5ffc675b719e69de7147068ab4ecf03f6d5db91a000d212a30f8ce61544691R167) on the `vttablet` for this purpose. 

I'd bias towards using a flag, since it's explicitly opt-in, and I can see an (admittedly imaginary) scenario where one would want debug logging on but to still redact the headers. However! I acknowledge there are definitely trade-offs in adding a flag. (vtctlds already have a billion flags, for example.) 

I don't feel very strongly about this, but I'm curious what you think. :) 

(@mattlord maybe you have an opinion too, since you've been thinking about this for https://github.com/vitessio/vitess/issues/9469 🙇 )

go/vt/workflow/long_polling_test.go
```diff
@@ -132,3 +134,87 @@ func TestLongPolling(t *testing.T) {
 	cancel()
 	wg.Wait()
 }
+
+func TestSanitizeRequestHeader(t *testing.T) {
```
doeg, line 138: These tests are :chef-kiss: ...thank you!

go/vt/workflow/long_polling.go
mattlord: Yeah, I think that adding a flag with the same name and general behavior makes sense for `vtctld` as well (I think we'll eventually add this to `vtgate` too). Thank you for working on this! ❤️

vmogilev: sounds good, it's good to be consistent with the rest of the project in how these toggles are implemented.  Let me see how to wire that up with a flag 🚧

vmogilev: any objections on making this new flag ( `dump_all_http_headers` ) default's value = `false`?

vmogilev: most likely it'll end-up being `workflow_manager_sanitize_http_headers` with the default value = `true`

how does that sound?

vmogilev: ... with the goal to make default behavior safe out of the box

doeg: Hmmm I do like making things safer by default, but because this is arguably a breaking change, it might be easier to start with a default of `workflow_manager_sanitize_http_headers=false`. That way the behaviour is entirely opt-in. This is what https://github.com/vitessio/vitess/pull/9550 does.

Either way, I would also suggest adding this to the release notes, which will be for Vitess 14.0, unless you feel strongly about backporting this to the current Vitess 13.0 release candidate. (FWIW it'll make no difference to Slack as we'll have to cherry pick this change to our fork anyhow, given that we're not yet ready for 13.0.) I don't think the release notes for 14.0 are being drafted yet, but I'll tag this PR with "release notes" so we don't forget! :)

doeg: Also, with respect to the flag name of `workflow_manager_sanitize_http_headers`, it might be fine to make this flag a bit more general, which is again what https://github.com/vitessio/vitess/pull/9550 does with its `sanitize_log_messages` flag. We could even use the same name, as @mattlord suggests.

vmogilev: ok this is done

doeg:
Hmm it looks like at least some of the CI failures are related to this change. For example, https://github.com/vitessio/vitess/runs/5160694796?check_suite_focus=true

```
ok  	vitess.io/vitess/go/vt/vtctl/workflow/vexec	0.018s
/tmp/go-build783617084/b1289/vtctld.test flag redefined: sanitize_log_messages
panic: /tmp/go-build783617084/b1289/vtctld.test flag redefined: sanitize_log_messages

goroutine 1 [running]:
flag.(*FlagSet).Var(0xc0000c8120, {0x2dba4e8, 0xc0000e1016}, {0x214d70d, 0x15}, {0x217307e, 0x24})
	/opt/hostedtoolcache/go/1.17.6/x64/src/flag/flag.go:879 +0x2f4
flag.(*FlagSet).BoolVar(...)
	/opt/hostedtoolcache/go/1.17.6/x64/src/flag/flag.go:632
flag.(*FlagSet).Bool(0x10, {0x214d70d, 0x15}, 0x0, {0x217307e, 0x24})
	/opt/hostedtoolcache/go/1.17.6/x64/src/flag/flag.go:645 +0x7d
flag.Bool(...)
	/opt/hostedtoolcache/go/1.17.6/x64/src/flag/flag.go:652
vitess.io/vitess/go/vt/vtctld.init()
	/home/runner/work/vitess/vitess/go/vt/vtctld/vtctld.go:45 +0x2db	
```

I wonder if this is somehow conflicting with [`sanitize_log_messages` in vt/vttablet/tabletserver/tabletenv/config.go](https://github.com/vitessio/vitess/blob/ec8d09b1ba00b435c27a2deebef75fcfcdf5fb78/go/vt/vttablet/tabletserver/tabletenv/config.go#L167)? That would be really surprising to me given that `vttablet` and `vtctld` are different things but I do know that scoping with the `flag` package can get messy. 😬 Let me double check this hypothesis in the Vitess #developers channel.

doeg:
@vmogilev so @mattlord kindly [brought me up to speed on flags](https://vitess.slack.com/archives/CMDJ2KFEZ/p1644607886669289) and offered some suggestions for this PR:

1. Re-use the existing `sanitize_log_messages` defined in `vttablet` like this:

```diff
diff --git a/go/vt/vtctld/vtctld.go b/go/vt/vtctld/vtctld.go
index 8ddfd67f3e..948eb348a2 100644
--- a/go/vt/vtctld/vtctld.go
+++ b/go/vt/vtctld/vtctld.go
@@ -43,6 +43,7 @@ var (
 	enableRealtimeStats = flag.Bool("enable_realtime_stats", false, "Required for the Realtime Stats view. If set, vtctld will maintain a streaming RPC to each tablet (in all cells) to gather the realtime health stats.")
 	enableUI            = flag.Bool("enable_vtctld_ui", true, "If true, the vtctld web interface will be enabled. Default is true.")
 	durabilityPolicy    = flag.String("durability_policy", "none", "type of durability to enforce. Default is none. Other values are dictated by registered plugins")
+	sanitizeLogMessages bool

 	_ = flag.String("web_dir", "", "NOT USED, here for backward compatibility")
 	_ = flag.String("web_dir2", "", "NOT USED, here for backward compatibility")
@@ -60,6 +61,9 @@ func InitVtctld(ts *topo.Server) error {
 		return err
 	}

+	sanitizeLogMessages = flag.Lookup("sanitize_log_messages").Value.(flag.Getter).Get().(bool)
+	log.Infof("Starting vtctld. Log sanitization: %v", sanitizeLogMessages)
+
 	actionRepo := NewActionRepository(ts)

 	// keyspace actions
```

2. Generalize the flag's help text defined [here](https://github.com/vitessio/vitess/blob/3aebec8/go/vt/vttablet/tabletserver/tabletenv/config.go#L167) as it's currently written to be specific for tablets. (I confirmed that it's ok to leave the actual `flag.BoolVar` definition there, even as we're re-using it elsewhere. Go flags are kinda like the wild west, but I think there are plans to use a more structured flag library at some point.)

  So perhaps we can update `flag.BoolVar(&currentConfig.SanitizeLogMessages, "sanitize_log_messages", false, "Remove potentially sensitive information in tablet INFO, WARNING, and ERROR log messages such as query parameters.")` to something like `Remove potentially sensitive information in INFO, WARNING, and ERROR log messages, such as query parameters and sensitive request headers.`
  
I don't _think_ you'll need to modify your tests at all, since this PR is calling `sanitizeRequestHeader` with a parameter rather than relying on flags. I think this is fine! If, for some reason, you do want to test this behaviour by toggling the flag, then https://github.com/vitessio/vitess/pull/9614/files#diff-522719ab2adeb062f5c7ccfbe820d27b8fdbd9e048946ce0858b1665108f7d50R46 might be a useful example on using `flag.Set` in a test. :)

doeg:
Ha, ok I'm back with an even more detailed example from @mattlord 💖 here's a way to more safely look up a flag (in our case, to use the existing flag):

```diff
diff --git a/go/vt/vtctld/vtctld.go b/go/vt/vtctld/vtctld.go
index 8ddfd67f3e..26bdb8cf8c 100644
--- a/go/vt/vtctld/vtctld.go
+++ b/go/vt/vtctld/vtctld.go
@@ -43,6 +43,7 @@ var (
 	enableRealtimeStats = flag.Bool("enable_realtime_stats", false, "Required for the Realtime Stats view. If set, vtctld will maintain a streaming RPC to each tablet (in all cells) to gather the realtime health stats.")
 	enableUI            = flag.Bool("enable_vtctld_ui", true, "If true, the vtctld web interface will be enabled. Default is true.")
 	durabilityPolicy    = flag.String("durability_policy", "none", "type of durability to enforce. Default is none. Other values are dictated by registered plugins")
+	sanitizeLogMessages, sanitizeErrorMessages bool

 	_ = flag.String("web_dir", "", "NOT USED, here for backward compatibility")
 	_ = flag.String("web_dir2", "", "NOT USED, here for backward compatibility")
@@ -60,6 +61,23 @@ func InitVtctld(ts *topo.Server) error {
 		return err
 	}

+	slmFlag := flag.Lookup("sanitize_log_messages")
+	if slmFlag != nil {
+		getter, found := slmFlag.Value.(flag.Getter)
+		if found {
+			sanitizeLogMessages = getter.Get().(bool)
+		}
+	}
+	semFlag := flag.Lookup("sanitize_error_messages")
+	if semFlag != nil {
+		getter, found := semFlag.Value.(flag.Getter)
+		if found {
+			sanitizeErrorMessages = getter.Get().(bool)
+		}
+	}
+
+	log.Infof("Starting vtctld. Log sanitization: %v Error sanitization: %v", sanitizeLogMessages, sanitizeErrorMessages)
+
 	actionRepo := NewActionRepository(ts)

 	// keyspace actions
```

Also, per suggestion (2) above, if we update the helptext for the existing `sanitize_log_messages` flag, we'll also want to (separately) update the references in the vitess.io website: https://github.com/vitessio/website/search?q=sanitize_log_messages

vmogilev:
Hi @doeg , thank you for helping with CI failures - good catch on the duplicate flag name.  

Small concern I'd like to run by you -- if we go with this shared flag solution, do you know if we'll be able to pull-in this change into our version of Vitess?

My concern is around the bit where we now depend on something we don't yet have [in our branch](https://github.com/tinyspeck/vitess/blob/main/go/vt/vttablet/tabletserver/tabletenv/config.go) (flag defined in another commit/pkg).

I also feel a bit weird about this indirect dependency, but if this is the pattern we already use - then perhaps it's warranted to repeat it here.

deepthi:
> I also feel a bit weird about this indirect dependency, but if this is the pattern we already use - then perhaps it's warranted to repeat it here.

It's unfortunate that vtctld depends on some vttablet packages. Here's the list of dependencies
`go list` tells me that one of them is `vitess.io/vitess/go/vt/vttablet/tabletserver/tabletenv` so unfortunately that means that vtctld includes all flags from that package whether or not they make any sense for vtctld.

We don't use the `flag.Lookup` pattern very much. It's only in a couple of places in the `logutil` package right now. But it does seem like that is the only option if we want to use the same flag name.

Someday we will implement [VEP-4](https://github.com/vitessio/enhancements/pull/8) and clean all this up.

doeg:
> between the two options:
> 
> 1. reuse vttablet flag ( sanitize_log_messages )
> 2. implement a local vtctld flag ( vtctld_sanitize_log_messages )
> I tend to favor 2)

The second option sounds fine to me, but I defer to @mattlord and @deepthi as I work in the vtctld codebase too rarely to have formed any strong opinions. :) 

Both ways have their merits. Selfishly, I do like that a separate flag will make cherry picking this onto our old release more straightforward.

(If Vitess's flag usage will change with https://github.com/vitessio/enhancements/pull/8, perhaps more granular flags will make a migration to Cobra a _tiny_ bit easier?)

doeg (approved):

mattlord (approved):
Nice work! Really appreciate the time and effort put into the tests. ❤️

PR #9670, opened 2022-02-11
messaging: support vt_message_cols to limit cols

derekperkins:
## Description
this enables users to set an allow-list of columns that will be loaded into memory and streamed to subscribers when they call `stream * from tbl`

## Related Issue(s)
#9666 

## Checklist
- [x] Tests were added

### DIFF
--- a/go/test/endtoend/messaging/message_test.go
+++ b/go/test/endtoend/messaging/message_test.go
@@ -25,22 +25,19 @@ import (
 	"testing"
 	"time"
 
-	"vitess.io/vitess/go/test/endtoend/utils"
-
-	cmp "vitess.io/vitess/go/test/utils"
-
-	"vitess.io/vitess/go/vt/vtgate/evalengine"
-
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
 	"vitess.io/vitess/go/mysql"
 	"vitess.io/vitess/go/sqltypes"
 	"vitess.io/vitess/go/test/endtoend/cluster"
+	"vitess.io/vitess/go/test/endtoend/utils"
+	cmp "vitess.io/vitess/go/test/utils"
 	"vitess.io/vitess/go/vt/log"
 	"vitess.io/vitess/go/vt/proto/query"
 	querypb "vitess.io/vitess/go/vt/proto/query"
 	"vitess.io/vitess/go/vt/proto/topodata"
+	"vitess.io/vitess/go/vt/vtgate/evalengine"
 	"vitess.io/vitess/go/vt/vtgate/vtgateconn"
 )
 
@@ -63,6 +60,7 @@ create table vitess_message(
 
 	# required indexes
 	primary key(id),
+	index next_idx(time_next),
 	index poller_idx(time_acked, priority, time_next desc)
 
 	# add any secondary indexes or foreign keys - no restrictions
@@ -200,6 +198,7 @@ create table vitess_message3(
 
 	# required indexes
 	primary key(id),
+	index next_idx(time_next),
 	index poller_idx(time_acked, priority, time_next desc)
 
 	# add any secondary indexes or foreign keys - no restrictions
@@ -271,6 +270,87 @@ func TestThreeColMessage(t *testing.T) {
 
 	// Verify Ack.
 	qr := utils.Exec(t, conn, "update vitess_message3 set time_acked = 123, time_next = null where id = 1 and time_acked is null")
+	assert.Equal(t, uint64(1), qr.RowsAffected)
+}
+
+var createSpecificStreamingColsMessage = `create table vitess_message4(
+	# required columns
+	id bigint NOT NULL COMMENT 'often an event id, can also be auto-increment or a sequence',
+	priority tinyint NOT NULL DEFAULT '50' COMMENT 'lower number priorities process first',
+	epoch bigint NOT NULL DEFAULT '0' COMMENT 'Vitess increments this each time it sends a message, and is used for incremental backoff doubling',
+	time_next bigint DEFAULT 0 COMMENT 'the earliest time the message will be sent in epoch nanoseconds. Must be null if time_acked is set',
+	time_acked bigint DEFAULT NULL COMMENT 'the time the message was acked in epoch nanoseconds. Must be null if time_next is set',
+
+	# add as many custom fields here as required
+	# optional - these are suggestions
+	tenant_id bigint,
+	message json,
+
+	# custom to this test
+	msg1 varchar(128),
+	msg2 bigint,
+
+	# required indexes
+	primary key(id),
+	index next_idx(time_next),
+	index poller_idx(time_acked, priority, time_next desc)
+
+	# add any secondary indexes or foreign keys - no restrictions
+) comment 'vitess_message,vt_message_cols=id|msg1,vt_ack_wait=1,vt_purge_after=3,vt_batch_size=2,vt_cache_size=10,vt_poller_interval=1'`
+
+func TestSpecificStreamingColsMessage(t *testing.T) {
+	ctx := context.Background()
+
+	vtParams := mysql.ConnParams{
+		Host: "localhost",
+		Port: clusterInstance.VtgateMySQLPort,
+	}
+	conn, err := mysql.Connect(ctx, &vtParams)
+	require.NoError(t, err)
+	defer conn.Close()
+
+	streamConn, err := mysql.Connect(ctx, &vtParams)
+	require.NoError(t, err)
+	defer streamConn.Close()
+
+	utils.Exec(t, conn, fmt.Sprintf("use %s", lookupKeyspace))
+	utils.Exec(t, conn, createSpecificStreamingColsMessage)
+	defer utils.Exec(t, conn, "drop table vitess_message4")
+
+	utils.Exec(t, streamConn, "set workload = 'olap'")
+	err = streamConn.ExecuteStreamFetch("stream * from vitess_message4")
+	require.NoError(t, err)
+
+	wantFields := []*querypb.Field{{
+		Name: "id",
+		Type: sqltypes.Int64,
+	}, {
+		Name: "msg1",
+		Type: sqltypes.VarChar,
+	}}
+	gotFields, err := streamConn.Fields()
+	for i, field := range gotFields {
+		// Remove other artifacts.
+		gotFields[i] = &querypb.Field{
+			Name: field.Name,
+			Type: field.Type,
+		}
+	}
+	require.NoError(t, err)
+	cmp.MustMatch(t, wantFields, gotFields)
+
+	utils.Exec(t, conn, "insert into vitess_message4(id, msg1, msg2) values(1, 'hello world', 3)")
+
+	got, err := streamConn.FetchNext(nil)
+	require.NoError(t, err)
+	want := []sqltypes.Value{
+		sqltypes.NewInt64(1),
+		sqltypes.NewVarChar("hello world"),
+	}
+	cmp.MustMatch(t, want, got)
+
+	// Verify Ack.
+	qr := utils.Exec(t, conn, "update vitess_message4 set time_acked = 123, time_next = null where id = 1 and time_acked is null")
 	assert.Equal(t, uint64(1), qr.RowsAffected)
 }
 
--- a/go/vt/vttablet/tabletserver/schema/load_table.go
+++ b/go/vt/vttablet/tabletserver/schema/load_table.go
@@ -23,10 +23,12 @@ import (
 	"strings"
 	"time"
 
+	"vitess.io/vitess/go/mysql/collations"
 	"vitess.io/vitess/go/sqltypes"
 	"vitess.io/vitess/go/vt/mysqlctl"
-
+	querypb "vitess.io/vitess/go/vt/proto/query"
 	"vitess.io/vitess/go/vt/sqlparser"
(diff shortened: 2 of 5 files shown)


mattlord:
@derekperkins I apologize for the delay, I will be reviewing this today.

derekperkins:
Great, thanks. I'll clean it up and get it passing tests again. I've been using this as our custom build branch trying to solve #9936. We've had this running in prod for 6 weeks or so and it has been working great.

go/test/endtoend/messaging/message_test.go
```diff
@@ -236,6 +233,75 @@ func TestThreeColMessage(t *testing.T) {
...
+	priority bigint default 0,
+	time_next bigint default 0,
+	epoch bigint,
+	time_acked bigint,
+	msg1 varchar(128),
+	msg2 bigint,
+	primary key(id),
+	index next_idx(priority, time_next desc),
+	index ack_idx(time_acked))
+comment 'vitess_message,vt_stream_cols=id;msg1,vt_ack_wait=1,vt_purge_after=3,vt_batch_size=2,vt_cache_size=10,vt_poller_interval=1'`
```
mattlord, line 247: Don't feel strongly, but I think `vt_message_cols` would be better since this applies to the messages at a higher level, and at the lower level applies to both the poller and the results stream.

And beyond that ~ `createLimitedStreamingColsMessage` or ~ `TestSpecificStreamingColsMessage` would be slightly better IMO. Also minor things that I don't feel strongly about.

go/vt/vttablet/tabletserver/schema/load_table.go
```diff
@@ -117,20 +102,49 @@ func loadMessageInfo(ta *Table, comment string) error {
...
 		}
 	}
 
-	// Load user-defined columns. Any "unrecognized" column is user-defined.
-	for _, field := range ta.Fields {
-		if _, ok := hiddenCols[strings.ToLower(field.Name)]; ok {
-			continue
+	allowedStreamingCols, _ := getCols(keyvals, "vt_stream_cols")
+
+	// make sure that all the specified columns are present
```
mattlord, line 131: This is where `allowed` feels misleading as you're saying that they are required. If they were allowed then I would expect that them not being present would be fine.

go/vt/vttablet/tabletserver/schema/load_table.go
```diff
@@ -117,20 +102,49 @@ func loadMessageInfo(ta *Table, comment string) error {
...
+	for _, col := range allowedStreamingCols {
+		num := ta.FindColumn(sqlparser.NewColIdent(col))
+		if num == -1 {
+			return fmt.Errorf("%s missing from message table: %s", col, ta.Name.String())
 		}
-		ta.MessageInfo.Fields = append(ta.MessageInfo.Fields, field)
 	}
+
+	if len(allowedStreamingCols) > 0 {
+		if allowedStreamingCols[0] != "id" {
```
mattlord, line 140: But I guess we don't need the other requiredCols here? 
```
		"priority",
 		"time_next",
 		"epoch",
 		"time_acked",
```

I guess those are required in the table structure but not the messages... that would make sense. This could be a reason that `vt_message_fields` is the best name...

go/vt/vttablet/tabletserver/schema/load_table.go
```diff
@@ -157,3 +171,39 @@ func getNum(in map[string]string, key string) (int, error) {
 	}
 	return v, nil
 }
+
+func getCols(in map[string]string, key string) ([]string, error) {
+	sv := in[key]
+	if sv == "" {
+		return nil, fmt.Errorf("attribute %s not specified for message table", key)
+	}
+	cols := strings.Split(sv, ";")
```
mattlord, line 180: personal preference but after so many years my brain sees the semi-colon in SQL as a hard stmt/context delimiter. I can see that comma has its own problems in this context, but I think I would prefer pipe as it's not uncommon to use that as a column separator in output.

Another thing I don't feel super strongly about and leave up to you to make the final call on.

go/vt/vttablet/tabletserver/schema/load_table.go
```diff
@@ -157,3 +171,39 @@ func getNum(in map[string]string, key string) (int, error) {
 	}
 	return v, nil
 }
+
+func getCols(in map[string]string, key string) ([]string, error) {
```
mattlord, line 175: I feel like this should be called `parseMessageFields`?  Minor thing, I just didn't realize what it was for at first. We could also address that with a function comment.

go/vt/vttablet/tabletserver/schema/load_table.go
```diff
@@ -157,3 +171,39 @@ func getNum(in map[string]string, key string) (int, error) {
...
+	}
+	return fields
+}
+
+func getAllowedMessageFields(tableFields []*querypb.Field, streamCols []string) []*querypb.Field {
+	fields := make([]*querypb.Field, 0, len(streamCols))
+	// Load user-defined columns. Any "unrecognized" column is user-defined.
+	for _, col := range streamCols {
+		for _, field := range tableFields {
+			if field.Name == col {
```
mattlord, line 202: This is case and white space sensitive, which is fine, probably just worth documenting as most identifier comparisons in SQL are not (or dependent on the collation used). We could potentially use the new evalengine here to compare the identifiers in the same way MySQL would.

mattlord (approved):
Looks good. Thank you for working on this! This will make it feasible to have much larger message cache sizes.

I had some questions/comments/nits mostly related to naming but you can make the final call on the naming aspects. Naming things is hard, and as long as we update the docs and explain the field well that's what matters most. Speaking of that, I assume you'll have a docs PR ready to review as well? Let me know. That has gotten much easier now that we have versioned docs and I can offer help if needed.

rohit-nayak-ps (commented):
aside from minor nits, functionality lgtm

go/vt/vttablet/tabletserver/schema/load_table.go
mattlord: To explain what I mean here, it would be something like this:
```
func getAllowedMessageFields(tableFields []*querypb.Field, streamCols []string) []*querypb.Field {
	fields := make([]*querypb.Field, 0, len(streamCols))
	// Load user-defined columns. Any "unrecognized" column is user-defined.
	for _, col := range streamCols {
		for _, field := range tableFields {
			if res, _ := evalengine.NullsafeCompare(sqltypes.NewVarChar(field.Name), sqltypes.NewVarChar(col), collations.Default()); res == 0 {
				fields = append(fields, field)
				break
			}
		}
	}
	return fields
}
```

mattlord: And if we wanted to deal with the user potentially having whitespace in the comment value:
```
if res, _ := evalengine.NullsafeCompare(sqltypes.NewVarChar(strings.TrimSpace(field.Name)), sqltypes.NewVarChar(strings.TrimSpace(col)), collations.Default()); res == 0 {
				fields = append(fields, field)
				break
			}
```

derekperkins: Perfect, I'll go ahead and use that. I may clean up the other functions here in a secondary commit too

mattlord: Cool! To be clear, this is what I was thinking for the final function (ignoring the function name which is a separate issue):
```
func getAllowedMessageFields(tableFields []*querypb.Field, streamCols []string) []*querypb.Field {
	fields := make([]*querypb.Field, 0, len(streamCols))
	// Load user-defined columns. Any "unrecognized" column is user-defined.
	for _, col := range streamCols {
		for _, field := range tableFields {
			if res, _ := evalengine.NullsafeCompare(sqltypes.NewVarChar(field.Name),
				sqltypes.NewVarChar(strings.TrimSpace(col)),
				collations.Default()); res == 0 {
				fields = append(fields, field)
				break
			}
		}
	}
	return fields
}
```

This way the comparison is case insensitive as the default collation with MySQL 5.7 is `utf8mb4_general_ci` and with 8.0 is `utf8mb4_0900_ai_ci` and we're stripping any leading or trailing whitespace from the identifiers in the comment directive as I can see that happening (in a list you may expect whitespace not to matter).

go/test/endtoend/messaging/message_test.go
derekperkins: done

go/vt/vttablet/tabletserver/schema/load_table.go
derekperkins: renamed to `specifiedCols`

go/vt/vttablet/tabletserver/schema/load_table.go
derekperkins: done

go/vt/vttablet/tabletserver/schema/load_table.go
```diff
@@ -157,3 +171,39 @@ func getNum(in map[string]string, key string) (int, error) {
 	}
 	return v, nil
 }
+
+func getCols(in map[string]string, key string) ([]string, error) {
+	sv := in[key]
+	if sv == "" {
+		return nil, fmt.Errorf("attribute %s not specified for message table", key)
```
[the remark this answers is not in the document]
derekperkins: Good call. I was just mirroring the other functions, but that was overkill. Fixed

go/vt/vttablet/tabletserver/schema/load_table.go
derekperkins: done

go/vt/vttablet/tabletserver/schema/load_table.go
derekperkins: done

go/vt/vttablet/tabletserver/schema/load_table_test.go
```diff
@@ -134,6 +134,59 @@ func TestLoadTableMessage(t *testing.T) {
...
+		Type: sqltypes.VarBinary,
+	}}
+	assert.Equal(t, want, table)
+
+	// Test loading id & message columns in reverse order from vt_stream_cols
+	_, err = newTestLoadTable("USER_TABLE", "vitess_message,vt_stream_cols=message;id,vt_ack_wait=30,vt_purge_after=120,vt_batch_size=1,vt_cache_size=10,vt_poller_interval=30,vt_min_backoff=10,vt_max_backoff=100", db)
+	require.Equal(t, errors.New("vt_stream_cols must begin with id: test_table"), err)
+
+	// Test setting zero columns on vt_stream_cols, which is ignored and loads the default columns
+	table, err = newTestLoadTable("USER_TABLE", "vitess_message,vt_stream_cols,vt_ack_wait=30,vt_purge_after=120,vt_batch_size=1,vt_cache_size=10,vt_poller_interval=30,vt_min_backoff=10,vt_max_backoff=100", db)
```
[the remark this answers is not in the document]
derekperkins: there were already tests above without ~vt_stream_cols~`vt_message_cols` specified that I left unchanged to handle that case

derekperkins:
Review comments are all addressed now in the last commit in the last 2 commits. This is rebased on top of #10132. Once that is fixed and merged, I'll rebase again to get the tests passing

mattlord:
> Review comments are all addressed now in the last commit in the last 2 commits. This is rebased on top of #10132. Once that is fixed and merged, I'll rebase again to get the tests passing

@derekperkins Should be good now, test wise. I am still working on the locking related stuff in the other PR but the tests should be done.

derekperkins:
@mattlord I think this should be good to go. I ran into some rebasing issues when switching from being rebased on #10132 to main, not realizing you had squashed. In the end, it was easiest for me to just squash mine too, vs leaving the separate review commit. I addressed all the review items, via name changes and comments. May be worth a quick once over again, but no major functionality changes.

go/test/endtoend/messaging/message_test.go
```diff
@@ -274,6 +271,75 @@ func TestThreeColMessage(t *testing.T) {
 	assert.Equal(t, uint64(1), qr.RowsAffected)
 }
 
+var createSpecificStreamingColsMessage = `create table vitess_message4(
```
mattlord, line 274: Any reasons not to use the new structure we have in the docs? Not a big deal, but I'd like to avoid drift between what we test and what we tell users to use if possible from here on out.

derekperkins: Just didn't think to change it. I'll update the tests

mattlord (approved):
Looks great! So much easier to read for me. Thank you for taking the time to make those changes, add comments, etc. It's not only easier to read, but more resilient and easier to use as well. ❤️ 


I had one minor question about the table structure used — hoping we can base it on what we now have in the docs — but that's it.

Thanks again!

mattlord:
Oh, one last thing... will you be opening a website PR as well so that we add details on this new configuration parameter to the v14 docs here: https://github.com/vitessio/website/blob/prod/content/en/docs/14.0/reference/features/messaging.md and here: https://github.com/vitessio/website/blob/prod/content/zh/docs/14.0/reference/messaging.md ?

I can help with that (creation or review), just let me know.

mattlord:
Friendly ping on this 🙂

go/test/endtoend/messaging/message_test.go
derekperkins: done

derekperkins:
Working on the docs PR now

Issue #9647, opened 2022-02-07
evalengine: Support all built-in MySQL functions


vmg:
The last major shortcoming of the evaluation engine are the many built-in SQL functions that ship with MySQL and which we currently do not support.  Since adding support for these functions is an arduous, iterative project, this tracking issue intends to act as an authoritative list of the progress we've made bringing these functions over.

> ## Notes for Potential GSoC Students:
> The evaluation engine in Vitess is one of the most critical parts of our query serving infrastructure. This engine is capable of evaluating arbitrary SQL expressions directly inside Vitess' process, without reaching out to a live MySQL instance, and this allows us to plan and execute complex user queries (e.g. queries that contain `WHERE` and similar filter clauses) between Vitess shards much more efficiently.
>
> The upcoming Vitess 13.0 release ships with a largely rewritten evaluation engine, which is capable of evaluating a larger amount of expressions to its predecessor while being more accurate to MySQL's behavior. 
>
> Before applying for this project, see the following PRs for a development history of the new evaluation engine:
>
> - https://github.com/vitessio/vitess/pull/9286
> - https://github.com/vitessio/vitess/pull/9310
> - https://github.com/vitessio/vitess/pull/9391
> - https://github.com/vitessio/vitess/pull/9418
> - https://github.com/vitessio/vitess/pull/9523
> - https://github.com/vitessio/vitess/pull/9549
> - https://github.com/vitessio/vitess/pull/9597
> - https://github.com/vitessio/vitess/pull/9623
> - https://github.com/vitessio/vitess/pull/9646
>
> Also, please review the implementation of the engine as it exists right now in our main branch: https://github.com/vitessio/vitess/tree/main/go/vt/vtgate/evalengine
>
> If you're interested in this GSoC project, your task for the summer will involve continuing the work on this evaluation engine by implementing more builtin SQL functions. This is not a complex task, because the hard parts of the engine, the actual symbolic evaluation, have already been implemented by us: you'll have at your disposal a wide array of helper functions to work on the SQL data types and implement the behaviors of each function with --hopefully-- not much pain.
>
> Do note that adding support for all the functions in this tracking issue is not a feasible task for a summer project. As part of your application, choose a sensible subset of these functions which you'd like to work on. Explain why you think this is an important set of functions to prioritize, and how you've estimated that you'll have enough time to port all the relevant functions during your project.
>
[...]

PR #9673, opened 2022-02-11
evalengine: More functions!

vmg:
## Description

Hot off the press! New functionality for the `evalengine`. This PR implements and tests:

- `WEIGHT_STRING`: useful to ensure our collations implementation is actually working (spoiler: it is). This includes some changes to the sqlparser because `WEIGHT_STRING` has special syntax.
- the `!` operator (which is an alias for `NOT`)
- the `&`, `|`, `^`, `<<`, `>>`, `~`, and `BIT_COUNT` bit-wise operators
- some cleanups to the way we're translating code

## Related Issue(s)

- https://github.com/vitessio/vitess/issues/9647

## Checklist


GuptaManan100 (approved):
All the changes look good to me. The shift-reduce conflict in the parser is unavoidable because we have a catch-all statement for functions in the end. Until, we remove it, any function which is a non-reserved keyword will have shift-reduce conflicts.

However, the `check_make_parser.sh` has to be fixed to expect a conflict

go/vt/vtgate/evalengine/convert_test.go
```diff
@@ -86,6 +86,8 @@ func TestConvertSimplification(t *testing.T) {
 		{"coalesce(NULL, 2, NULL, 4)", ok("COALESCE(NULL, INT64(2), NULL, INT64(4))"), ok("INT64(2)")},
 		{"coalesce(NULL, NULL)", ok("COALESCE(NULL, NULL)"), ok("NULL")},
 		{"coalesce(NULL)", ok("COALESCE(NULL)"), ok("NULL")},
+		{"weight_string('foobar')", ok(`WEIGHT_STRING(VARCHAR("foobar"))`), ok(`VARBINARY("\x00F\x00O\x00O\x00B\x00A\x00R")`)},
+		{"weight_string('foobar' as char(12))", ok(`WEIGHT_STRING(VARCHAR("foobar") AS CHAR(12))`), ok(`VARBINARY("\x00F\x00O\x00O\x00B\x00A\x00R\x00 \x00 \x00 \x00 \x00 \x00 ")`)},
```
systay, line 90: shouldn't we also add a couple of the bit operations to this test?

systay (approved):

go/vt/vtgate/evalengine/convert_test.go
vmg: The bit operations are integration tested in the major suite.

PR #9675, opened 2022-02-11
Skip repl config in restore when active reparents disabled

mattlord:
## Description

Do not attempt to setup replication when [active reparents are disabled](https://github.com/vitessio/vitess/blob/3adcf439960fbbde3d7f82998717a89f6ebcfb60/go/vt/mysqlctl/mysqld.go#L64).

It makes no sense to do any replication configuration related work as part of a [`RestoreFromBackup`](https://github.com/vitessio/vitess/blob/3adcf439960fbbde3d7f82998717a89f6ebcfb60/go/vt/vttablet/tabletmanager/rpc_backup.go#L155) RPC call because[ we do not ever start replication automatically when active reparents are disabled](https://github.com/vitessio/vitess/search?q=disableActiveReparents). We only need to restore the data and save the associated replication state metadata -- then it's up to the operator to (optionally) complete the replication setup if and as needed at a later time.

### An Example Use Case
As a Vitess operator, I need to have a business continuity plan in the event of a regional disaster. To meet this, I want to have an active/passive setup where there's an active Vitess cluster in one infrastructure region and a passive cluster in another sufficiently distant region. In order to do that, we may have a cell local topo service in each region and they are generally isolated from one another. I seed the initial passive region with backups from the active region's keyspace/shards and manage async MySQL replication myself — so active reparents are disabled — in order to support the ability of making this passive Vitess cluster the active one (with minimal potential data loss) in the event of a regional outage where the previous active cluster was running.

In this case, if you're doing a backup restore for a shard that *has* a primary — as seen in the global topo's shard record — but the primary cannot be reached (networking issue, in a different cell local topo server, etc) then the restore will fail.

This PR allows the restore to complete so that the user can finish the replication setup piece at a later time as appropriate. 

## Checklist
- [x] Tests are not required

### DIFF
--- a/go/vt/vttablet/tabletmanager/restore.go
+++ b/go/vt/vttablet/tabletmanager/restore.go
@@ -499,6 +499,15 @@ func (tm *TabletManager) startReplication(ctx context.Context, pos mysql.Positio
 		return vterrors.Wrap(err, "failed to set replication position")
 	}
 
+	// At this point we've restored the data from the backup _and_ we've saved the
+	// replication state metadata that is consistent with the snapshot of the data
+	// contained in the backup. This means that we have everything needed to start
+	// replication at a later time, and since with active parents disabled we do
+	// not ever start replication automatically, we can now safely return.
+	if *mysqlctl.DisableActiveReparents {
+		return nil
+	}
+
 	// Read the shard to find the current primary, and its location.
 	tablet := tm.Tablet()
 	si, err := tm.TopoServer.GetShard(ctx, tablet.Keyspace, tablet.Shard)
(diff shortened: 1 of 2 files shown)


mattlord:
I could limit this behavior change further and return an error in this scenario *unless* active reparents are disabled? I can see a reasonable case to be made that in the noted use case those should also be explicitly disabled.

deepthi:
> I could limit this behavior change further and continue returning an error in this scenario _unless_ active reparents are disabled? I can see a reasonable case to be made that in the noted use case those should also be explicitly disabled.

I like the idea of checking for `-disable_active_reparents` and skipping `startReplication` completely if that is set.

deepthi:
In fact I can now see that we have this block of code right **after** we attempt to start replication 😭 
```
	// If active reparents are disabled, we don't restart replication. So it makes no sense to wait for an update on the replica.
	// Return immediately.
	if *mysqlctl.DisableActiveReparents {
		return nil
	}
```
We should just move that check up to where we call `startReplication` and not even call it.

mattlord:
I went through the code to better understand what it actually means to disable active reparents and [we seem to just avoid touching MySQL Replication at all in that case and (virtually) immediately return in related calls](https://github.com/vitessio/vitess/search?q=disableActiveReparents). 

So I moved the check for that flag earlier in `startReplication` and improved the comment to reflect that when the flag is set all we need to do with regards to replication is save the replication state metadata associated with the backup. Then we can safely return when active reparents are disabled.

GuptaManan100 (approved):

deepthi (approved):

PR #9677, opened 2022-02-11
Add .PHONY to all Makefile targets

miniyarov:
## Description

This PR intends to add `.PHONY` to all Makefile targets to avoid any confusion when there's a filename created in the same directory.

## Checklist


deepthi:
Hi @miniyarov, thank you for the contribution. Can you signoff the commit? Instructions are at https://github.com/vitessio/vitess/pull/9677/checks?check_run_id=5163087051

PR #9679, opened 2022-02-14
Support VDiff across DB versions

mattlord:
## Description
VDiff did not support comparing tables across database versions. This was because it relied on the result of the passed down [`weight_string(column)`](https://dev.mysql.com/doc/refman/8.0/en/string-functions.html#function_weight-string) values in order to do the comparisons and that is not stable across MySQL versions and was not meant for critical external use.

This PR adds support for VDiff across MySQL versions by removing the usage of [`weight_string(column)`](https://dev.mysql.com/doc/refman/8.0/en/string-functions.html#function_weight-string) and instead leverages the new MySQL compatible collation support in Vitess to perform the comparisons within Vitess using the collation that the column is using in the schema definition or if none is available for the column within VDiff then we use our compatible implementation of MySQL's `binary` collation to compare as bytes.

When comparing the results on the source and target, we first compare the PK columns on each side — expecting them to be in the same order within the resultsets as we pass down an `ORDER BY pkcol1, pkcol2, ...` clause with the `SELECT` statement to MySQL for the table — and if we see the PKs don't match at a given resultset index then it is registered as an _**extra**_ row on either side. For PK's that do align we then go on to compare the rest of the row (non PK columns) on both sides and if they don't match it is registered as a _**mismatched**_ row.

Due to collation differences between versions that cannot be dealt with w/o making VDiffs up to ~ 60% slower — e.g. by adding binary or collation clauses to the ORDER BY passed down to MySQL, e.g.`ORDER BY binary pkcol` or `ORDER BY pkcol COLLATE utf8mb4_general_ci` [here](https://github.com/vitessio/vitess/blob/eb3168bd51275f71d127e1718e6187cdaa5c27e0/go/vt/wrangler/vdiff.go#L385-L421), which causes us to do a filesort with the full scan — we do a second pass on up to by default 1,000 registered _**extra**_ rows (capped to limit resource usage) seen to compare them before the final report. This way we're not reporting table _data_ differences when the only difference was the _order_ the records were returned from MySQL in the results. After the other changes made in this PR, the extra rows seen on both sides from resultset ordering differences should also be much, much smaller.

## Related Issue(s)
 - Fixes: https://github.com/vitessio/vitess/issues/9715

## Checklist
- [x] Should this PR be backported? No, I think this should probably bake in 14.0-SNAPSHOT and be backported by those that require it
- [x] Tests were added
- [x] Documentation: https://github.com/vitessio/website/pull/980


go/vt/wrangler/vdiff.go
```diff
@@ -1051,29 +1045,16 @@ func (td *tableDiffer) compare(sourceRow, targetRow []sqltypes.Value, cols []com
...
-				return 0, err
-			}
-			trowBytes, err := targetRow[compareIndex].ToBytes()
-			if err != nil {
-				return 0, err
-			}
-			c = bytes.Compare(srowBytes, trowBytes)
+		var collationID collations.ID
+		// if the collation is nil or unknown, use binary collation to compare as bytes
+		if col.collation == nil || col.collation.ID() == collations.Unknown {
```
vmg, line 1052: `col.collation.ID() == collations.Unknown` can never be true

vmg (approved):
The comparison part of this looks good! I would like to optimize `NullSafeCompare` since it's doing more work than it needs to, but that shouldn't be a blocker at this point.

go/test/endtoend/vreplication/cluster.go
```diff
@@ -106,6 +109,128 @@ func setTempVtDataRoot() string {
...
+		dbTypeMajorVersion = fmt.Sprintf("mysql-%s", majorVersion)
+	} else {
+		dbTypeMajorVersion = fmt.Sprintf("%s-%s", strings.ToLower(string(flavor)), majorVersion)
+	}
+	return dbTypeMajorVersion, nil
+}
+
+// downloadDBTypeVersion downloads a recent major version release build for the specified
+// DB type.
+// If the file already exists, it will not download it again.
```
rohit-nayak-ps, line 183: The tarball will already exist if the test is running on a runner that it ran earlier on, right? I guess the probability over time will be high this will be true. But is there a way to bake these into the images itself? I am wondering if this will add flakiness because of unpredictable download times and/or availability issues.

mattlord: Yes, if a runner is re-used it would already be there.

I don't believe we have any control over the images used in GH actions (AFAIK, they are the same images used for CodeSpaces).

We already rely on being able to download MySQL specific things, we download and install the MySQL repo. Then install the mysql packages using the repo. The tar download is *only* done in this cross version test and IMO does not add significant risk beyond what we already have. We have plenty of other tests that are legit flaky today, but I have seen no issues with this test. We can always address any seen flakiness later though.

go/vt/wrangler/vdiff.go
```diff
@@ -1051,29 +1045,16 @@ func (td *tableDiffer) compare(sourceRow, targetRow []sqltypes.Value, cols []com
...
+		var collationID collations.ID
+		// if the collation is nil or unknown, use binary collation to compare as bytes
+		if col.collation == nil {
+			collationID = collations.CollationBinaryID
 		} else {
-			// TODO(king-11) make collation aware
-			c, err = evalengine.NullsafeCompare(sourceRow[compareIndex], targetRow[compareIndex], collations.Unknown)
+			collationID = col.collation.ID()
 		}
+		c, err = evalengine.NullsafeCompare(sourceRow[compareIndex], targetRow[compareIndex], collationID)
```
rohit-nayak-ps, line 1057: Nice to be able to just use `NullsafeCompare` here and get rid of the `weight_string()` related code!

rohit-nayak-ps (approved):
LGTM

go/test/endtoend/vreplication/cluster.go
```diff
@@ -106,6 +109,128 @@ func setTempVtDataRoot() string {
...
+// binary is used by mysqlctl along with the correct flavor specific config file.
+func downloadDBTypeVersion(dbType string, majorVersion string, path string) error {
+	client := http.Client{
+		Timeout: 10 * time.Minute,
+	}
+	var url, file, versionFile string
+	dbType = strings.ToLower(dbType)
+
+	if dbType == "mysql" && majorVersion == "5.7" {
+		versionFile = "mysql-5.7.37-linux-glibc2.12-x86_64.tar.gz"
```
vmg, line 197: This is gonna crash in OSX. You should skip the test if not being ran on Linux to avoid weird behaviors.

rohit-nayak-ps (changes requested):
This looks good!
Only issue is about limiting the extra rows while reporting a json output.

#### Aside:

If we find cases where there are a huge number of extra rows we may want to optimize using one or both of these techniques:
1. Do the extra row compares in a "streaming" fashion as the rows appear rather than after collecting all. If the extra rows are in proximity then the buffer size of extra rows will be kept small, letting us match more rows. However if there is not much locality the cpu time will be high.
1. `sort.Slice()` the extra rows before comparing and fail on the first mismatch to minimize cpu due to lower number of overall iterations and also much smaller number of `reflect.DeepEqual()`s in case of actual failures. 

I am just noting this here, hopefully the known use-cases will work well with the current implemenation.

go/test/endtoend/vreplication/cluster.go
mattlord: Yeah, on x86_64 linux specifically. I'll add a check and log.Fatal for this. Thanks for the reminder! I was so focused on the CI ...

mattlord:
> This looks good! Only issue is about limiting the extra rows while reporting a json output.
> 
> #### Aside:
> If we find cases where there are a huge number of extra rows we may want to optimize using one or both of these techniques:
> 
> 1. Do the extra row compares in a "streaming" fashion as the rows appear rather than after collecting all. If the extra rows are in proximity then the buffer size of extra rows will be kept small, letting us match more rows. However if there is not much locality the cpu time will be high.

Yeah, the example case I was looking at — where the streaming would have not been more efficient — may not be too common.

> 2. `sort.Slice()` the extra rows before comparing and fail on the first mismatch to minimize cpu due to lower number of overall iterations and also much smaller number of `reflect.DeepEqual()`s in case of actual failures.
> 
> I am just noting this here, hopefully the known use-cases will work well with the current implemenation.

I think that this is probably worth doing now. 1,000 * 1,000 + deep equal each time is heavy (and these rows could be large with blobs etc). I'll work on sorting the slices by the PK cols, then we can compare the slices by index.

Thank you!

rohit-nayak-ps (approved):
lgtm!

PR #9680, opened 2022-02-14
Replace InitShardPrimary usage in tests by PlannedReparentShard

GuptaManan100:
## Description

This PR replaces usages of `InitShardPrimary` with `PlannedReparentShard` in more end to end tests since this is the new default to use.
 
## Checklist


go/test/endtoend/reparent/newfeaturetest/reparent_test.go
```diff
@@ -30,79 +30,6 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-// PRS TESTS
```
deepthi, line 33: Ok to delete this test because a bunch of tests now do the same thing.

deepthi (approved):

Issue #9682, opened 2022-02-14
Compatibility: support ALTER TABLE... PARTITION BY


shlomi-noach:
Our parser today does not support `ALTER TABLE ... PARTITION BY`. It does support statements like `ALTER TABLE ... ADD PARTITION` or `ALTER TABLE.... REORGANIZE PARTITION` etc. But is missing on the "completely repartition like so" option.

Example query:

```sql
ALTER TABLE t PARTITION BY HASH(id) PARTITIONS 4;
```

MySQL's `ALTER TABLE` lets you repartition everything according to a completely new scheme. Basically, it uses the same partitioning syntax from `CREATE TABLE`. We already have this supported in `CREATE TABLE` in the form of [PartitionOption](https://github.com/vitessio/vitess/blob/531d1274466636051bd078a3c408d89a8b9beb5e/go/vt/sqlparser/ast.go#L1521-L1534)

All we need to do to close this issue is to support `PartitionOption` inside [AlterTable](https://github.com/vitessio/vitess/blob/531d1274466636051bd078a3c408d89a8b9beb5e/go/vt/sqlparser/ast.go#L458-L465)

Self assigning

shlomi-noach:
Related: https://github.com/vitessio/vitess/issues/8604

There is at least another outstanding partitioning limitation to the parser; in https://github.com/vitessio/vitess/issues/8604 partitioning is marked as complete but I believe that's a mistake.

PR #9683, opened 2022-02-14
Support for ALTER TABLE ... PARTITION BY ...

shlomi-noach:
## Description

Fixes https://github.com/vitessio/vitess/issues/9682

The parser now supports _partition option_ in `ALTER TABLE` statements, such as in this query:
```sql
ALTER TABLE t PARTITION BY HASH(id) PARTITIONS 4
```

See `parse_test.go` where a query that was previously `partialDDL` now parses completely.

## Related Issue(s)

https://github.com/vitessio/vitess/issues/9682

## Checklist


GuptaManan100 (approved):
Thankyou for catching this! I was under the impression we had added the support for Partitions, and only sub-partitions was incomplete

Issue #9629, opened 2022-02-03
Feature Request: Add a CI check to validate copyright/license on new files


doeg:
### Feature Description

```markdown
Many, many times I have either forgotten to add the copyright/license text when adding a new file, or I get the year wrong. (Sometimes I manage to do both!)

It would be nice to have a CI check that validates the presence of the license text with the correct copyright year. 

This check should support either an allow list or an ignore list so that we can skip files that don't require a license; config files, documentation, and so on. 

I'd propose the check works as follows:
- Fail if a new file has an invalid license
- Fail if an updated file that _should_ have a license (based on the allow/ignore lists) does not (i.e., a past developer perhaps forgot to add it (looking at myself here!)) 
- Warn if an _updated_ file has an invalid copyright year
```

### Use Case(s)

```markdown
Noted above. :)
```

mattlord:
We might be able to use this existing action from the marketplace: https://github.com/marketplace/actions/reuse-compliance-check

deepthi:
> Warn if an updated file has an invalid copyright year

What is an invalid copyright year? Note that the copyright year should NOT be updated every time a file changes.

systay:
I don't like that we have to have a license header in every file, and reading up on it, it's pretty clear that this is not needed with Apache 2.
From https://infra.apache.org/apply-license.html#copy-per-file

```
Do I have to have a copy of the license in each source file?
You only need to add one full copy of the license per distribution. See the policy.
```

can we pretty please remove the licenses and just stick to the LICENSE file in the root?

mattlord:
@systay there's a disconnect here, at least in my mind. :-) You keep sharing that link, but it says "do I need a [full] _copy of the license in each source file_". That's not what we have, we have a license *header* that points to the license in each file. And in their FAQs they do seem to say that *this* is required (at least recommended).

The full copy of the license can be seen here: http://www.apache.org/licenses/LICENSE-2.0

We only link to that in our files. I don't think anybody disagrees with the assertion that 1we don't need the *full license text* in each file — we do not today AFAICT. 

In the same doc that you link to:
[...]

PR #9684, opened 2022-02-14
Remove license headers from files

systay:
## Description
My understanding of the ALv2 is that it is not necessary to add a license header to every single file - it's enough to clearly state the license in the repository once.

This according to: https://infra.apache.org/apply-license.html#copy-per-file

Since we have a license file already in [LICENSE](https://github.com/vitessio/vitess/blob/main/LICENSE), I don't think we need to list the license in all the files.


vmg (approved):
This PR crashed my browser.

deepthi:
Do we retain external authors' copyrights? For example, some code was copied from `gh-ost` and modified for Vitess.

deepthi:
Let's hold off on merging any license changes until after 13.0 is out the door (Feb 22).

mattlord:
I am not a lawyer, but as far as I understand things:
License and copyright are not the same thing. I don't have a strong opinion on the license per file part, although I would ask why (?), but copyright is separate. License is about acceptable use while copyright is about lineage and ownership (and it's sharable, transferrable, etc.). This is my work and these are the restrictions I place on its use.

deepthi:
We had actually checked with CNCF a while ago and this is what we were told:

> The general response is here: https://github.com/cncf/foundation/blob/master/copyright-notices.md#copyright-notices
>
> We do recommend a simple notice for every file, it could be just as simple as a one line "Copyright Vitess Authors, SPDX-License-Identifier: Apache-2.0"

If you follow the link, they recommend exactly what they told us in this response. No year required.
The exception is that if we copy code from some other open source repo we have to preserve whatever copyright they come with. Hence my comment/question about `gh-ost`.

mattlord:
Sharing my related comment on another ticket: https://github.com/vitessio/vitess/issues/9629#issuecomment-1049122484

mattlord:
Can we close this? 

You can see more details here: https://github.com/vitessio/vitess/issues/9629#issuecomment-1049122484

But in short I think you're confusing a license *header* with the actual *license text*. I think that merging this would be a serious mistake, violating FOSS norms and practices in general, along with what both the Apache Foundation and CNCF recommend. Then there's the copyright issue noted, in addition to the license header.

deepthi:
I agree with @mattlord that this cannot be merged in its current form.
We need _at least_ a "short" copyright header in each file AND we need to preserve the copyright headers on files copied from other OSS project.
Closing for now, we can open another PR if we want to replace the existing notice with a shorter one.

PR #9687, opened 2022-02-14
Fix missing time_heartbeat column error

utk9:
## Description
Seeing errors like the following in VReplication workflows:
```
error Unknown column 'time_heartbeat' in 'field list' (errno 1054) (sqlstate 42S22) during query: update _vt.vreplication set time_updated=1644542766, time_heartbeat=1644542766 where id=14 updating time
```
This column was added in this recent PR: https://github.com/vitessio/vitess/pull/9538/files. The DDL to add the time_heartbeat column is added [here](https://github.com/planetscale/vitess/blob/0bb4b29c7dfccdd8c17e27ad697b662d7faa776d/go/vt/binlog/binlogplayer/binlog_player.go#L569) which is eventually used in the withDDL [here](https://github.com/planetscale/vitess/blob/0bb4b29c7dfccdd8c17e27ad697b662d7faa776d/go/vt/vttablet/tabletmanager/vreplication/engine.go#L66). But the actual heartbeat query done [here](https://github.com/planetscale/vitess/blob/0bb4b29c7dfccdd8c17e27ad697b662d7faa776d/go/vt/vttablet/tabletmanager/vreplication/vplayer.go#L259) doesn't actaully use this withDDL. Instead this withDDL is used when the vreplication controller starts the binlogplayer [here](https://github.com/planetscale/vitess/blob/0bb4b29c7dfccdd8c17e27ad697b662d7faa776d/go/vt/vttablet/tabletmanager/vreplication/controller.go#L201). But the queries that are made here ([WithDDLInitialQueries](https://github.com/planetscale/vitess/blob/0bb4b29c7dfccdd8c17e27ad697b662d7faa776d/go/vt/binlog/binlogplayer/binlog_player.go#L572)) don't actually include one that uses this newly added time_heartbeat column, so the column never gets created. This PR fixes that issue.

## Checklist
- [x] Should this PR be backported? Yes, to release-13.0

### DIFF
--- a/go/vt/binlog/binlogplayer/binlog_player.go
+++ b/go/vt/binlog/binlogplayer/binlog_player.go
@@ -569,10 +569,14 @@ var AlterVReplicationTable = []string{
 	"ALTER TABLE _vt.vreplication ADD COLUMN time_heartbeat BIGINT(20) NOT NULL DEFAULT 0",
 }
 
-// WithDDLInitialQueries contains the queries to be expected by the mock db client during tests
+// WithDDLInitialQueries contains the queries that:
+// - are to be expected by the mock db client during tests, or
+// - trigger some of the above _vt.vreplication schema changes to take effect
+//   when the binlogplayer starts up
 var WithDDLInitialQueries = []string{
 	"SELECT db_name FROM _vt.vreplication LIMIT 0",
 	"SELECT rows_copied FROM _vt.vreplication LIMIT 0",
+	"SELECT time_heartbeat FROM _vt.vreplication LIMIT 0",
 }
 
 // VRSettings contains the settings of a vreplication table.


mattlord:
@utk9 thank you for the contribution! We were just looking into this issue on our side. Can you please correct the DCO issue as described here? https://github.com/vitessio/vitess/pull/9687/checks?check_run_id=5190784543

Then once we get all the other tests passing (I'll handle that), we can merge it.

go/vt/binlog/binlogplayer/binlog_player.go
```diff
@@ -569,10 +569,14 @@ var AlterVReplicationTable = []string{
 	"ALTER TABLE _vt.vreplication ADD COLUMN time_heartbeat BIGINT(20) NOT NULL DEFAULT 0",
 }
 
-// WithDDLInitialQueries contains the queries to be expected by the mock db client during tests
+// WithDDLInitialQueries contains the queries that:
+// - are to be expected by the mock db client during tests, or
+// - trigger some of the above _vt.vreplication schema changes to take effect
+//   when the binlogplayer starts up
```
deepthi, line 575: Nice comment 💯

go/vt/binlog/binlogplayer/binlog_player.go
```diff
@@ -569,10 +569,14 @@ var AlterVReplicationTable = []string{
...
 var WithDDLInitialQueries = []string{
 	"SELECT db_name FROM _vt.vreplication LIMIT 0",
 	"SELECT rows_copied FROM _vt.vreplication LIMIT 0",
+	"SELECT time_heartbeat FROM _vt.vreplication LIMIT 0",
```
rohit-nayak-ps, line 579: This change only affects the queries required by the mock db in unit tests and will not fix the issue of the missing column. https://github.com/vitessio/vitess/pull/9700 fixes the update query to create the column, if it is missing.

utk9: This does actually fix the problem outside of unit tests as well. Please see my PR description above as to why.

mattlord:
Closing this for now as this is superseded now by #9700. We still appreciate the issue and PR though! Thank you again, @utk9 ❤️

go/vt/binlog/binlogplayer/binlog_player.go
mattlord: What unit test failures are you seeing? I didn't see in any the original PR https://github.com/vitessio/vitess/pull/9538 or in https://github.com/vitessio/vitess/pull/9700.

utk9: There are no unit test failures. The error is in actual VReplication workflows in live environment.

mattlord: Sorry, I misunderstood your comment. The place that you added the statement would trigger the column to be added here (only): https://github.com/vitessio/vitess/blob/main/go/vt/vttablet/tabletmanager/vreplication/controller.go#L201-L205

It's possible that's enough, but the ALTER to add that column is already included in the withDDL object here: https://github.com/planetscale/vitess/blob/46ebbf98bfda218b50b2a016d7166f6428c0ac52/go/vt/vttablet/tabletmanager/vreplication/engine.go#L79

That's used to wrap (virtually) all of the other queries in the vreplication workflow. In Rohit's PR we added the wrapper in one place it was not used (it should have been but was missed), which may happen to be the only path hit depending on the sequence of events. 

That being said, your change can't hurt. One more way to ensure it gets created. 🙂 I'll cherry-pick your commit onto his branch. That way you also get permanent credit for it. 

Please let me know if you think I'm missing or misunderstanding something.

rohit-nayak-ps: > This does actually fix the problem outside of unit tests as well. Please see my PR description above as to why.

Many apologies, I hadn't read the PR description and the related code pointers well enough. Your change would have indeed fixed the issue. The `WithDDLInitialQueries` was initially created just for the mock db for testing purposes and I hadn't realized it was now used to harness WithDDL's side-effect of creating the desired schema.

To avoid such issues in the future, we will plan to move to a predictable mechanism of creating the desired schema on tablet startup rather than on a `WithDDL.Exec` failure.

Issue #9689, opened 2022-02-14
Run internal table GC operations as dba


malpani:
### Overview of the Issue

Internal GC transitions are currently executed as `vt_app`. This causes GC to fail in setups where `vt_app` user does not have alter privileges resulting in failures like

```[2022-02-14 18:39:39.535409] E0214 18:39:39.535275 3558744 tablegc.go:272] TableGC: error transitioning table _vt_PURGE_206a04b98b8f11ec9a090205111c4bd9_20220212230615 to EVAC: DROP, ALTER command denied to user 'vt_app'@'localhost' for table '_vt_PURGE_206a04b98b8f11ec9a090205111c4bd9_20220212230615' (errno 1142) (sqlstate 42000) during query: RENAME TABLE `_vt_PURGE_206a04b98b8f11ec9a090205111c4bd9_20220212230615` TO _vt_EVAC_206a04b98b8f11ec9a090205111c4bd9_20220217183939```

My proposal is to switch these to run as dba

### Reproduction Steps

```markdown
Trigger any schema changes in a setup where `vt_app` user does not have alter/drop privileges
```

### Binary Version

```shell
Version: 12.0.0-SNAPSHOT (Git revision 013134f367 branch 'master') (also present on head)
```

### Operating System and Environment details

```shell
n/a
```

### Log Fragments

```shell
[2022-02-14 18:39:39.535409] E0214 18:39:39.535275 3558744 tablegc.go:272] TableGC: error transitioning table _vt_PURGE_206a04b98b8f11ec9a090205111c4bd9_20220212230615 to EVAC: DROP, ALTER command denied to user 'vt_app'@'localhost' for table '_vt_PURGE_206a04b98b8f11ec9a090205111c4bd9_20220212230615' (errno 1142) (sqlstate 42000) during query: RENAME TABLE `_vt_PURGE_206a04b98b8f11ec9a090205111c4bd9_20220212230615` TO _vt_EVAC_206a04b98b8f11ec9a090205111c4bd9_20220217183939```
```

GuptaManan100:
@deepthi I think this will fall under our larger cleanup of which user to use where

PR #9690, opened 2022-02-15
Use AllPrivs user for table GC transitions

malpani:
Signed-off-by: Ankit Malpani <malpani@stripe.com>

## Description
Use admin user for tablet GC operations.

Context: Currently `vt_app` user is used for internal table GC operations. This however does not have enough privileges (like `alter/drop`) in all setups which results in GC failures. This change moves internal GC operations to be invoked as dba user

## Related Issue(s)
Fixes https://github.com/vitessio/vitess/issues/9689

## Checklist
- Should this PR be backported? - no
- Tests were added or are not required - tablegc unit tests pass but they dont exercise permission changes
- Documentation was added or is not required - not required

## Deployment Notes
n/a

### DIFF
--- a/go/vt/vttablet/tabletserver/gc/tablegc.go
+++ b/go/vt/vttablet/tabletserver/gc/tablegc.go
@@ -171,7 +171,7 @@ func (collector *TableGC) Open() (err error) {
 	}
 
 	log.Info("TableGC: opening")
-	collector.pool.Open(collector.env.Config().DB.DbaWithDB(), collector.env.Config().DB.DbaWithDB(), collector.env.Config().DB.AppDebugWithDB())
+	collector.pool.Open(collector.env.Config().DB.AllPrivsWithDB(), collector.env.Config().DB.DbaWithDB(), collector.env.Config().DB.AppDebugWithDB())
 	atomic.StoreInt64(&collector.isOpen, 1)
 
 	for _, t := range collector.tickers {
--- a/go/vt/vttablet/tabletserver/gc/tablegc.go
+++ b/go/vt/vttablet/tabletserver/gc/tablegc.go
@@ -171,7 +171,7 @@ func (collector *TableGC) Open() (err error) {
 	}
 
 	log.Info("TableGC: opening")
-	collector.pool.Open(collector.env.Config().DB.AppWithDB(), collector.env.Config().DB.DbaWithDB(), collector.env.Config().DB.AppDebugWithDB())
+	collector.pool.Open(collector.env.Config().DB.DbaWithDB(), collector.env.Config().DB.DbaWithDB(), collector.env.Config().DB.AppDebugWithDB())
 	atomic.StoreInt64(&collector.isOpen, 1)
 
 	for _, t := range collector.tickers {


deepthi:
@shlomi-noach this seems to be the same type of issue as #8835, fixed in #8836
Do we need to review the code to see if there are other places? How comfortable are we with doing more and more things using `dba` instead of `app`?

mattlord:
Any reason(s) to use DBA over AllPrivs? DBA risks problems when run on replicas not using super-read-only.

malpani:
> Any reason(s) to use DBA over AllPrivs? DBA risks problems when run on replicas not using super-read-only.

@mattlord `dba` is being used for internal transitions during online schema changes like in https://github.com/vitessio/vitess/pull/8836 and for [purge phase of GC](https://github.com/vitessio/vitess/blob/main/go/vt/vttablet/tabletserver/gc/tablegc.go#L423). So to stay consistent with that theme, i used it for other internal GC transitions

shlomi-noach:
> How comfortable are we with doing more and more things using dba instead of app?

I seem to remember we actually went the opposite way, moving from `dba` to `app` user. I generally feel comfortable with giving this kind of work to `dba` account. The GC code does check whether we're on a `read_only` server or not -- but I guess there can be race conditions, so if `super_read_only=0` and with a race condition, there could be writes to a replica.

However, the change in this PR means all connections used by GC Collector are DBA. Seems excessive.

I guess I'm also missing some context. Is this a specific environment where the operators have defined lesser privileges to the app user? (Just asking to clarify).

Should we possibly add a configuration flag to `vttable`/GC, like `-tabletgc-use-dba-account` so that users can choose desired behavior?

deepthi:
> Any reason(s) to use DBA over AllPrivs? DBA risks problems when run on replicas not using super-read-only.

Should #8836 be re-done to use AllPrivs as well?

malpani:
> Should https://github.com/vitessio/vitess/pull/8836 be re-done to use AllPrivs as well?

Could that break an existing setup as we would be moving from a user with higher privileges to a lower one unlike in this scenario? Either way, I would prefer making changes to schema change permissions on a different PR and restrict this to GC permissions if that sounds good to all?

mattlord:
> > Should #8836 be re-done to use AllPrivs as well?
> 
> Could that break an existing setup as we would be moving from a user with higher privileges to a lower one unlike in this scenario?

`vt_allprivs` is supposed to be used for anything that does NOT require the `SUPER` privilege (having everything but `SUPER` and `GRANT`) and DDL falls under that: https://github.com/vitessio/vitess/blob/main/config/init_db.sql#L65-L72

This is because the `SUPER` privilege should be used very carefully and sparingly, at least when not using super-read-only in the mysql and tablet configs (otherwise it's easy to perform errant writes on replicas in various scenarios).

> Either way, I would prefer making changes to schema change permissions on a different PR and restrict this to GC permissions if that sounds good to all?

That sounds good to me. We'll need a separate PR later to change this everything anyway. Thank you for the contribution and discussion! ❤️

shlomi-noach:
Sorry for dropping the ball! `AllPrivs` is good on my side for table GC. It should also be fine for online DDL (https://github.com/vitessio/vitess/pull/8836/files) because all we do is `RENAME TABLES...` and `AllPrivs` user has the privileges for that.

shlomi-noach (approved):
This LGTM. Last thoughts about merging this? If none, I'll merge tomorrow.

PR #9473, opened 2022-01-06
Add upgrade downgrade tests for reparent

GuptaManan100:
## Description

Continuation of #9300. Follows the same methodology as described there. This PR adds a workflow for the upgrade-downgrade tests for reparent operations.

## Related Issue(s)

- #7344 

## Checklist
- [ ] Should this PR be backported?
- [x] Tests were added or are not required
- [x] Documentation was added or is not required

PR #9692, opened 2022-02-15
Remove deprecated upgrade-downgrade CI build

frouioui:
## Description

This pull request removes the old upgrade-downgrade CI build. It is no longer meaningful to keep this test since the addition of #9300, #9473 and #9501.

## Checklist


systay (approved):

deepthi:
Merging this now. Don't we also need to remove this test from config.json and delete the associated source code?

PR #7060, opened 2020-11-20
Adds Planning and Parsing Support for Create View of MySQL 5.7

GuptaManan100:
Adds Planning and Parsing Support for Create View of MySQL 5.7

PR #9693, opened 2022-02-15
Parser fix for CREATE VIEW statement

### LINKED ISSUE #9691: sqlparser: fail parsing valid CREATE VIEW statement
I have the following simple view:
```sql
show create view v3 \G
                View: v3
         Create View: CREATE ALGORITHM=UNDEFINED DEFINER=`msandbox`@`localhost` SQL SECURITY DEFINER VIEW `v3` AS select `t`.`id` AS `id` from `t`
character_set_client: utf8mb4
collation_connection: utf8mb4_0900_ai_ci
```

Unfortunately, `sqlparser.Parse()` yields with:
```
Parse("CREATE ALGORITHM=UNDEFINED DEFINER=`msandbox`@`localhost` SQL SECURITY DEFINER VIEW `v3` AS select `t`.`id` AS `id` from `t`") err: Code: INVALID_ARGUMENT
        syntax error at position 58 near 'localhost'
```

GuptaManan100:
## Description

This PR fixes how we parser the user for `DEFINER` in `CREATE VIEW` and `ALTER VIEW` statements.

The statement we set out to fix was - 
```mysql
CREATE ALGORITHM=UNDEFINED DEFINER=`msandbox`@`localhost` SQL SECURITY DEFINER VIEW `v3` AS select `t`.`id` AS `id` from `t`
```
This works as intended now.

Along the way we also fixed how we parse Identifiers which should not be accepting `@` symbols in-between. i.e. - `localhost@x` is not a valid identifier, `localhost` is a valid identifier and the remaining part `@x` will be an AT_ID. This is part of the tech debt incurred when we first started supporting CREATE VIEW statements #7060

## Related Issue(s)

 - Fixes #9691
 - #8604 

## Checklist


shlomi-noach (approved):
Thank you @GuptaManan100 and @systay for the quick turnaround!

PR #9695, opened 2022-02-15
vtcombo: Add the ability to specify routing rules in test topology

simon-engledew:
We would like to be able to specify routing rules in our test setup and check them in to version control with the other vitess related config.

This extends the test topology used by vtcombo/vttest to also support specifying routing rules. It allows you to stand up a test setup with routing rules already loaded and avoids having to run vtctlclient commands when setting up testing.

Another possible option would be to allow a routing_rules.json file in the -schema-dir folder. That would not require changing a protobuf file. 🤔 

Happy to do whatever fits best! 🙇 

/cc @mattlord 😅 

## Checklist


mattlord:
Hi @simon-engledew !

I'm sorry, I just realized that you may have been waiting for initial feedback before marking this ready for review. I think the current direction is fine, the changes are localized to vtcombo/vttestserver. I like the general idea a lot -- again, I think others will find this kind of thing useful in their CI and dev tests too. Thank you!

mattlord:
Side note, I do think that being able to specify the routing rules as a JSON file would *also* be nice but we can do that later on. 

The only thing we're missing here is at least 1 test. We can do something similar to what we did here: https://github.com/vitessio/vitess/pull/9633

We should confirm that we are able to load the routing rules and they are in effect. Please let me know how much help/direction you want for the tests. Thanks again for continuing to make vtcombo better!

mattlord (changes requested):
The only thing we're missing here is at least 1 test. We can do something similar to what we did here: https://github.com/vitessio/vitess/pull/9633

We should confirm that we are able to load the routing rules and they are in effect. Please let me know how much help/direction you want for the tests. Thanks again for continuing to make vtcombo better!

mattlord:
Side note: given that I think this usage of `vtcombo` in dev tests (and/or CI) makes a lot of sense, and we're adding new features... we should have a dedicated page in the docs on `vtcombo` and `vttestserver`. What they are, what the common use cases are, how they work, what the flags are, what directory structures they expect, how to pass topology info and routing rules, etc. 

I just did a thorough read through of the `vtcombo` source and it was enlightening. I think that the usefulness of this is only improving — thanks in large part to your work and contributions! — and we should really surface this to users so that it's easy to discover and leverage.

simon-engledew:
@mattlord ok, I've added tests that prove the routing rules are being loaded.

The easiest way to do that was to set up a simple routed table and query it via the mysql protocol - a nice bonus because it means the tests now check that too.

mattlord (approved):
Thanks, @simon-engledew ! Nice work on this ❤️

Issue #9652, opened 2022-02-08
Vtexplain does not work for subsharding vindexes


aquarapid:
### Overview of the Issue

vtexplain does not work for subsharding vindexes

### Reproduction Steps

```markdown
Vschema (in vtexplain format):


{
  "sharded": {
    "sharded": true,
    "vindexes": {
      "t1_multicol": {
        "type": "multicol",
        "params": {
          "column_count": "3",
          "column_bytes": "4,2,2",
          "column_vindex": "xxhash,xxhash,xxhash"
        }
      },
      "xxhash": {
        "type": "xxhash"
      },
      "hash": {
        "type": "hash"
      }
    },
    "tables": {
      "t1": {
        "columnVindexes": [
          {
            "columns": [
              "c1",
              "c2",
              "c3"
            ],
            "name": "t1_multicol"
          }
        ]
      }
    }
  }
}
```

Schema:

```
CREATE TABLE t1 (
    c1 BIGINT NOT NULL,
    c2 BIGINT NOT NULL,
    c3 BIGINT NOT NULL,
    c4 BIGINT NOT NULL,
    PRIMARY KEY (c1)
);
```

Run vtexplain against this:

```
$ vtexplain -schema-file vtexplain.sql -vschema-file vtexplain.json -shards 16 -sql 'select * from t1 where c1=1 and c2=1 and c3=1;' 
----------------------------------------------------------------------
select * from t1 where c1=1 and c2=1 and c3=1

1 sharded/-10: select * from t1 where c1 = 1 and c2 = 1 and c3 = 1 limit 10001
1 sharded/10-20: select * from t1 where c1 = 1 and c2 = 1 and c3 = 1 limit 10001
1 sharded/20-30: select * from t1 where c1 = 1 and c2 = 1 and c3 = 1 limit 10001
1 sharded/30-40: select * from t1 where c1 = 1 and c2 = 1 and c3 = 1 limit 10001
1 sharded/40-50: select * from t1 where c1 = 1 and c2 = 1 and c3 = 1 limit 10001
1 sharded/50-60: select * from t1 where c1 = 1 and c2 = 1 and c3 = 1 limit 10001
1 sharded/60-70: select * from t1 where c1 = 1 and c2 = 1 and c3 = 1 limit 10001
1 sharded/70-80: select * from t1 where c1 = 1 and c2 = 1 and c3 = 1 limit 10001
[...]

PR #9697, opened 2022-02-15
vtexplain: added multicol test to show subshard routing

harshit-gangal:
## Description

This PR adds a test in vtexplain to show subshard vindex routing.

The test works locally, but on CI it fails with the error `limit on 8128 simultaneously alive goroutines is exceeded, dying`.
To show subsharding works, I have to create 512 shards.

Below is the test result.

```
----------------------------------------------------------------------
select * from user_region where regionId = 2305843009213693952 /* subshard */

1 ks_sharded/1f80-2000: select * from user_region where regionId = 2305843009213693952 limit 10001 /* subshard */
1 ks_sharded/2000-2080: select * from user_region where regionId = 2305843009213693952 limit 10001 /* subshard */
1 ks_sharded/2080-2100: select * from user_region where regionId = 2305843009213693952 limit 10001 /* subshard */

----------------------------------------------------------------------
```
I will change the test on the PR to have 4 shards only so that it works on CI but it will not reflect the actual value of subshard routing.

## Related Issue(s)

- Closes https://github.com/vitessio/vitess/issues/9652

## Checklist

### DIFF
--- a/go/vt/vtexplain/testdata/gen4-queries.sql
+++ b/go/vt/vtexplain/testdata/gen4-queries.sql
@@ -1,0 +1,3 @@
+
+select * from user_region where regionId = 2305843009213693952 and userId = 100 /* exact shard */;
+select * from user_region where regionId = 2305843009213693952 /* subshard */;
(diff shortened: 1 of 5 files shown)


deepthi (approved):
Nice that you used `region_id` and `user_id` in the tests 💯

PR #9698, opened 2022-02-15
evalengine: Conversions

vmg:
## Description

Here's another batch of functionality for the `evalengine` which I wanted to implement before the GSoC, since it's a bit tricky to get right.

This PR implements the following syntax:

- `CONVERT(expr, type)`
- `CONVERT(expr USING charset)`
- `CAST(expr AS type)`
- `BINARY expr`
- `HEX(expr)`

Some notes that I wanted to point out:

- The `evalengine` package already had a `Convert` API, which was the main entrypoint to convert a `sqlparser` AST into an `evalengine` expression. Since "convert" is actually a core operator of SQL, I decided to rename our entrypoint API to `Translate`, because the alternative made the code very confusing to follow.
- I had to do some changes to `sqlparser` to get this fully working (thanks to @GuptaManan100 and @systay who helped with these).
    - I removed `BINARY` as an specific unary operator in the AST, since this is a keyword that has been deprecated in MySQL 8.0. In order to keep backwards compatibility, I did what the official documents suggested: rewrite `BINARY expr` expressions as `CAST(expr AS BINARY)` -- the rewrite step happens in the parser itself, so in the future it'll be easier to deprecate.
    - I fixed the charset parsing for `CONVERT(expr, type CHAR CHARACTER SET charset)` and the equivalent `CAST(expr AS CHAR CHARACTER SET charset)`. More specifically, the aliased forms `CHAR ASCII` and `CHAR UNICODE` are now supported in the same way they are in MySQL
    - I added `DOUBLE`, `FLOAT` and `REAL` as supported convert types for these operators (these are new conversion types introduced in 8.0.17).
- I implemented the equivalent `CAST` and `CONVERT` syntaxes in the `sqlparser`. This was uneventuful, although I didn't get around to adding support for `DATE`/`DATETIME`/`YEAR`/etc types, since the evaluation engine doesn't have the helper functions required to deal with those yet (a good task for GSoC). Attempting to convert into those types will result in an "unsupported operation" error.
- I implemented the `HEX` built-in SQL function, since I needed it to accurately test the result of multi-byte conversions in our integration tests.

That's pretty much it. We now support all relevant type conversions!

## Checklist


vmg:
I have committed a significant refactoring of the internals of the `evalengine` to this PR; I wish I could do it on a separate PR, but it is required because of an issue that I found earlier this morning while debugging with @systay and @frouioui.

Basically, we found that the `env.TypeOf` method that the VTGate uses for fetching the field types of the results of a query was not behaving properly. This didn't come as a surprise to me: the `TypeOf` method was using the underlying `Expr.typeof` method of every `Expr` node in the `evalengine` to acquire the type of the expression without evaluating it. This is the same helper that the new evaluation engine uses to know the type of an expression without evaluating it, enabling actual "lazy evaluation".

Here's the problem: for many expression types, the `Expr.typeof` method was returning `-1` because it didn't know how to calculate the type of an expression without evaluating it. This was marked as a `TODO` even though it didn't have a meaningful impact on the behavior of the evaluation engine: whenever a node returned `-1` as its type during evaluation, we would simply eagerly evaluate it to get its actual type. This was _not_ ideal because it caused some expressions to be evaluated earlier than I ought to, but it hasn't proven to be a problem in practice.

The reason why some of these expressions were returning `-1` is because the actual output of many expressions in a SQL query can be `NULL`, depending on whether one of its arguments (or one of the nested arguments) is `NULL`. This made figuring out whether the type of the result is `NULL` without an actual evaluation essentially impossible. But here's the insight I was missing: `NULL` is not a meaningful result for the _type_ of an expression for this kind of symbolic evaluation. The only thing that matters is knowing whether an expression is nullable or not. As an example, the output of `SELECT HEX(:arg1)` is, in theory, unknowable because it depends on whether `:arg1` is `NULL` or not, but the actual answer we want for our result fields is that the output is a _nullable_ `VARCHAR`, and this is something that we can now calculate reliably.

go/vt/vtgate/evalengine/eval_result.go
```diff
@@ -253,24 +294,66 @@ func (er *EvalResult) setTuple(t []EvalResult) {
...
+	return true
+}
+
+func (er *EvalResult) truncate(size int) {
+	switch er.typeof() {
+	case sqltypes.VarBinary:
+		if size > len(er.bytes_) {
+			pad := make([]byte, size)
+			_ = append(pad[:0], er.bytes_...)
+			er.bytes_ = pad
```
frouioui, line 347: what does `_ = append(pad[:0], er.bytes_...)` do?

go/vt/vtgate/planbuilder/testdata/memory_sort_cases.txt
```diff
@@ -792,6 +792,7 @@ Gen4 plan same as above
 
 # unary expression
 "select a from user order by binary a desc"
+"unsupported: in scatter query: complex order by expression: convert(a, binary)"
 {
```
frouioui, line 796: meaning v3 does not support those queries anymore?

frouioui (approved):
this is amazing work @vmg 🏆 I don't have much to say besides the few comments I left!

go/vt/sqlparser/precedence.go
```diff
@@ -85,8 +85,6 @@ func precedenceFor(in Expr) Precendence {
 			return P4
 		case BangOp:
 			return P3
```
systay, line 87: Hmm... should we move down the precedences coming before? We're leaving a weird gap in the list of precedences. Probably not important though, all tests are happy

go/vt/vtgate/evalengine/eval_result.go
```diff
@@ -32,9 +32,29 @@ import (
 	"vitess.io/vitess/go/vt/vtgate/evalengine/decimal"
 )
 
+type flag uint16
+
 const (
-	flagHex = 1 << iota
-	flagBit
+	// flagNull marks that this value is null; implies flagNullable
```
systay, line 38: This information is awesome. We could extend it, and hook it up with the schema tracker so we know if columns are nullable.

I know at least one place where I would like to use this information in the planner, so we should consider making it a concept that is readable from other packages as well

go/vt/vtgate/evalengine/func.go
```diff
@@ -256,18 +325,29 @@ func builtinFuncCollation(env *ExpressionEnv, args []EvalResult, result *EvalRes
...
-func builtinFuncIsNull(_ *ExpressionEnv, args []EvalResult, result *EvalResult) {
+func (builtinCollation) typeof(_ *ExpressionEnv, args []Expr) (sqltypes.Type, flag) {
 	if len(args) != 1 {
-		throwArgError("ISNULL")
+		throwArgError("COLLATION")
 	}
+	return sqltypes.VarChar, 0
+}
+
+type builtinIsNull struct{}
```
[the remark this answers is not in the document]
vmg: Nice. I'll do that at the `sqlparser` level :+1:

go/vt/vtgate/evalengine/eval_result.go
vmg: Noted, I'll add that to my TODO list.

go/vt/vtgate/evalengine/eval_result.go
vmg: It means I brokened this! It was supposed to be a `copy` not an `append`. Gonna fix.

go/vt/vtgate/evalengine/decimal/scan.go
```diff
@@ -374,3 +374,21 @@ func (r byteReader) ReadByte() (byte, error) {
 func (r byteReader) UnreadByte() error {
 	return r.UnreadRune()
 }
+
+const nines = "99999999999999999999999999999999999999999999999999" +
```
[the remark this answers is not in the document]
vmg: I'm removing this code on my followup PR because I do agree it's a bit nastay :joy:

go/vt/vtgate/evalengine/bit.go
```diff
@@ -205,18 +213,33 @@ func (bit *BitwiseExpr) eval(env *ExpressionEnv, result *EvalResult) {
...
+	t1, f1 := bit.Left.typeof(env)
+	t2, f2 := bit.Right.typeof(env)
+
+	switch bit.Op.(type) {
+	case BitwiseBinaryOp:
+		if t1 == sqltypes.VarBinary && t2 == sqltypes.VarBinary &&
+			(f1&(flagHex|flagBit) == 0 || f2&(flagHex|flagBit) == 0) {
+			return sqltypes.VarBinary, f1 | f2
+		}
+	case BitWiseShiftOp:
```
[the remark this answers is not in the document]
vmg: Yeah normalized case :+1:

go/vt/vtgate/evalengine/func.go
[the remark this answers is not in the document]
vmg: Done, I did the rewrite in the evalengine because the `isnull` is not a keyword nor reserved word in MySQL, which means that they're also doing the rewrite in their eval engine. :smile:

PR #9700, opened 2022-02-15
VReplication Workflows: Use WithDDL while updating time_heartbeat to be backwardly compatible with upgrades to existing cluster

rohit-nayak-ps:
## Description

#9538 introduced a regression with existing clusters because a new column `time_heartbeat` was added to `_vt.vreplication` and the `WithDDL` package was not used while updating this new column.  The workflow would stall with these errors 
```
E0215 20:03:45.678794       1 dbclient.go:107] error in stream 14, retrying after 5s: error Unknown column 'time_heartbeat' in 'field list' (errno 1054) (sqlstate 42S22) during query: update _vt.vreplication set time_updated=1644955425, time_heartbeat=1644955425 where id=1 updating time
```

This PR has the following changes:
* the heartbeat update query is modified to use `WithDDL`.
* a test around this change to confirm that the column gets recreated
* a new query added to the `WithDDLInitialQueries` list: it accesses the new  `time_heartbeat` column to force the related ALTER TABLE to run when a vreplication controller starts up

## Checklist
- [X] Should this PR be backported? Yes, to release-13.0
- [X] Tests were updated

### DIFF
--- a/go/vt/vttablet/tabletmanager/vreplication/vplayer.go
+++ b/go/vt/vttablet/tabletmanager/vreplication/vplayer.go
@@ -256,7 +256,7 @@ func (vp *vplayer) updateHeartbeat(tm int64) error {
 	if err != nil {
 		return err
 	}
-	if _, err := vp.vr.dbClient.Execute(update); err != nil {
+	if _, err := withDDL.Exec(vp.vr.vre.ctx, update, vp.vr.dbClient.ExecuteFetch, vp.vr.dbClient.ExecuteFetch); err != nil {
 		return fmt.Errorf("error %v updating time", err)
 	}
 	return nil
--- a/go/vt/vttablet/tabletmanager/vreplication/vplayer_flaky_test.go
+++ b/go/vt/vttablet/tabletmanager/vreplication/vplayer_flaky_test.go
@@ -132,8 +132,7 @@ func TestVReplicationTimeUpdated(t *testing.T) {
 	// drop time_heartbeat column to test that heartbeat is updated using WithDDL and can self-heal by creating the column again
 	env.Mysqld.ExecuteSuperQuery(ctx, "alter table _vt.vreplication drop column time_heartbeat")
 	time.Sleep(2 * time.Second)
-	timeUpdated3, transactionTimestamp3, timeHeartbeat3 := getTimestamps()
-	_, _, _ = timeUpdated3, transactionTimestamp3, timeHeartbeat3
+	_, _, timeHeartbeat3 := getTimestamps()
 	require.Greater(t, timeHeartbeat3, timeHeartbeat2, "time_heartbeat not updated")
 }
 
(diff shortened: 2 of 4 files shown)


go/vt/vttablet/tabletmanager/vreplication/vplayer.go
```diff
@@ -256,7 +256,7 @@ func (vp *vplayer) updateHeartbeat(tm int64) error {
 	if err != nil {
 		return err
 	}
-	if _, err := vp.vr.dbClient.Execute(update); err != nil {
+	if _, err := withDDL.Exec(vp.vr.vre.ctx, update, vp.vr.dbClient.ExecuteFetch, vp.vr.dbClient.ExecuteFetch); err != nil {
```
mattlord, line 259: OK, I think that makes sense. 🙂 This was the only thing getting called (thus the seen log messages) before the final SwitchTraffic step on the workflow that had existed prior to the upgrade.

go/vt/vttablet/tabletmanager/vreplication/vplayer_flaky_test.go
```diff
@@ -104,26 +104,37 @@ func TestVReplicationTimeUpdated(t *testing.T) {
...
-	timeUpdated1, transactionTimestamp1 := getTimestamps()
+	timeUpdated1, transactionTimestamp1, timeHeartbeat1 := getTimestamps()
 	time.Sleep(2 * time.Second)
-	timeUpdated2, _ := getTimestamps()
+	timeUpdated2, _, timeHeartbeat2 := getTimestamps()
 	require.Greater(t, timeUpdated2, timeUpdated1, "time_updated not updated")
 	require.Greater(t, timeUpdated2, transactionTimestamp1, "transaction_timestamp should not be < time_updated")
+	require.Greater(t, timeHeartbeat2, timeHeartbeat1, "time_heartbeat not updated")
+
+	// drop time_heartbeat column to test that heartbeat is updated using WithDDL and can self-heal by creating the column again
```
mattlord, line 132: Thanks!

mattlord (approved):
LGTM

deepthi:
Is the change to `WithDDLInitialQueries` not required any more? I agree with the analysis by @utk9 in #9687, this is actually being used outside of test code.

mattlord:
@rohit-nayak-ps I hope you don't mind — I cherry-picked the commit from https://github.com/vitessio/vitess/pull/9687. You can [see my reasoning here](https://github.com/vitessio/vitess/pull/9687#discussion_r807507442).

rohit-nayak-ps:
> Is the change to `WithDDLInitialQueries` not required any more? I agree with the analysis by @utk9 in #9687, this is actually being used outside of test code. It seems to me that there are two possible paths of upgrading the schema.

Yes, as @utk9 pointed out, we do have the two different paths now, which is not ideal. I am working on a PR to to run the initial DDLs on tablet startup itself rather than depend on side-effects. We will refactor the associated WithDDL code in the near future. Having the expected schema upfront should also simplify the expected mock db queries in unit tests.

shlomi-noach (approved):
LGTM!

deepthi (approved):

PR #9637, opened 2022-02-06
OnlineDDL: reject duplicate UUID with different migration_context

shlomi-noach:
## Description

Followup to https://github.com/vitessio/vitess/pull/9325. In this PR we reject a migration if submitted with same UUID and different `migration_context`.

As reminder, the user is allowed to supply their own migration UUIDs via `vtctl ApplySchema -uuid_list='...'`. This functionality is used to allow the user control and idempotency over their migrations.

UUID is a unique migration identifier. It is impossible to have any two migrations on same shard with same UUID. Hence, when one submits `-uuid_list=...` with an already existing UUID, their submission does not create a new migration. However, it is then also reasonable to expect that the user provides `-migration_context`. If the same migration is submitted twice or more, with same UUID and same `migration_context`, that is fine. The newer submissions are silently ignored, and this is a desired behavior that allows idempotency.

However, if the new submission uses a different `migration_context` (either because the user has specified one, that is different from the original one; or the user has specified none at all, in which case `vtctl` generates a random, unique value), then we consider this as an unintentional duplication, and as of this PR we return an error.

Here is a valid resubmission:

```shell
$ vtctlclient ApplySchema -skip_preflight -ddl_strategy='online' -migration_context="1222-2342342-423423-42" -uuid_list="4b25f49c_8709_11ec_9f54_0a43f95f28a5" -sql "alter table customer engine=innodb" commerce
4b25f49c_8709_11ec_9f54_0a43f95f28a5
$ vtctlclient ApplySchema -skip_preflight -ddl_strategy='online' -migration_context="1222-2342342-423423-42" -uuid_list="4b25f49c_8709_11ec_9f54_0a43f95f28a5" -sql "alter table customer engine=innodb" commerce
4b25f49c_8709_11ec_9f54_0a43f95f28a5
```

Here is an invalid one:
```shell
$ vtctlclient ApplySchema -skip_preflight -ddl_strategy='online' -uuid_list="4b25f49c_8709_11ec_9f54_0a43f95f28a6" -sql "alter table customer engine=innodb" commerce
4b25f49c_8709_11ec_9f54_0a43f95f28a6
$ vtctlclient ApplySchema -skip_preflight -ddl_strategy='online' -uuid_list="4b25f49c_8709_11ec_9f54_0a43f95f28a6" -sql "alter table customer engine=innodb" commerce
E0206 04:59:15.849989  800011 main.go:67] E0206 04:59:15.849210 vtctl.go:3282] schema change failed, ExecuteResult: {
[...]

PR #9325, opened 2021-12-05
vtctl ApplySchema -uuid_list

shlomi-noach:
## Description

`vtctlient ApplySchema` now support a new optional `-uuid_list` flag. It is possible for the user to explicitly specify the UUID(s) for given migration(s). UUIDs must be in a specific format. If given, number of UUIDs must match the number of DDL statements. Example:

```shell
vtctlclient OnlineDDL ApplySchema -sql "drop table t1, drop table t2" -uuid_list "d08f0000_51c9_11ec_9cf2_0a43f95f28a3,d08f0001_51c9_11ec_9cf2_0a43f95f28a3" commerce
```

Format is comma delimited, zero or more UUIDs.

Vitess will assign each migration with given UUID in order of appearance.
It is the user's responsibility to ensure given UUIDs are globally unique. If the user submits a migration with an already existing UUID, that migration never gets scheduled nor executed.

The purpose of this change is to give users better control over submitted migrations. Moreover, it is a means for users to achiever idempotency: the user can _choose_ to resubmit a migration with same UUID, intentionally ensuring it does not execute again.

`-uuid_list` is optional. If omitted or if empty then Vitess auto-generates UUIDs for the migrations.

## Related Issue(s)

- Tracking: #6926 

## Checklist
- [ ] Should this PR be backported?
- [x] Tests were added or are not required
- [ ] Documentation was added or is not required

PR #9704, opened 2022-02-16
Online DDL: resubmitting a migration with same UUID retries it in case it was cancelled or failed.

shlomi-noach:
## Description

With the introduction of `-uuid_list`, https://github.com/vitessio/vitess/pull/9325, we let the users choose the migration's UUID. We do this because we want to let the users run idempotent migrations; resubmitting the same migration ([with same context](https://github.com/vitessio/vitess/pull/9637)) should not run the migration again.

Except... if it is in `failed` or `cancelled` state. In that scenario, we do really want the migration to run again. So this PR implements this behavior: to retry a migration if it's in those states.

The PR implements the desired behavior, in a scenario which I overlooked previously. So I consider this a bugfix rather than an enhancement.

Added an endtoend test to validate the new behavior.

## Related Issue(s)

- #9325 
- #9676 
- #6926 

## Checklist


go/vt/vttablet/onlineddl/executor.go
```diff
@@ -3524,46 +3524,68 @@ func (e *Executor) SubmitMigration(
 	}
 
 	if onlineDDL.StrategySetting().IsSingleton() || onlineDDL.StrategySetting().IsSingletonContext() {
-		e.migrationMutex.Lock()
-		defer e.migrationMutex.Unlock()
+		// we need two wrap some logic within a mutex, so as to make sure we can't submit the migration whilst
```
shlomi-noach, line 3527: Most of the change here is indentation :) try reviewing the change in https://github.com/vitessio/vitess/pull/9704/files?w=1

rohit-nayak-ps (approved):
lgtm

PR #9710, opened 2022-02-16
Make gen4 the default planner for release 14

systay:
## Description
Release 14 is still a long way away, but we might as well do this change now to flush out any bugs.

## Checklist
- [x] Should this PR be backported? ** NO **

### DIFF
--- a/go/vt/vtgate/planbuilder/builder.go
+++ b/go/vt/vtgate/planbuilder/builder.go
@@ -111,9 +111,11 @@ func getConfiguredPlanner(vschema plancontext.VSchema, v3planner func(string) se
 			fallback: v3planner(query),
 		}
 		return fp.plan, nil
-	default:
-		// default is v3 plan
+	case V3:
 		return v3planner(query), nil
+	default:
+		// default is gen4 plan
+		return gen4Planner(query, Gen4), nil
 	}
 }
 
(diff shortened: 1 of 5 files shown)


GuptaManan100 (approved):

GuptaManan100:
🤞

go/vt/vtgate/planbuilder/builder.go
```diff
@@ -111,9 +111,11 @@ func getConfiguredPlanner(vschema plancontext.VSchema, v3planner func(string) se
...
 		}
 		return fp.plan, nil
-	default:
-		// default is v3 plan
+	case V3:
 		return v3planner(query), nil
+	default:
+		// default is gen4 plan
+		return gen4Planner(query, Gen4), nil
 	}
```
harshit-gangal, line 119: nit: remove case Gen4 from above.

systay: the gen4 case is still needed. There are variations of gen4 (left2right and greedy only) that are not handled by the default case

frouioui:
With Gen4 being the new default planner, `TestExplain` started failing consistently on the `selectsharded` queries. I have detailed the issue in #9854 and proposed a fix for it in #9855.

harshit-gangal:
We should mark the flag for deprecation as well.

PR #9712, opened 2022-02-16
Cleanup of parsing of Partitions and additional parsing support

### LINKED ISSUE #9696: sqlparser: support `PARTITION BY LIST`
Our `sqlparser` does not parse the partition list in:
```sql
create table t1 (id int primary key) partition by list (id) (partition p1 values in(11,21), partition p2 values in (12,22))
```
It only produces `create table t1`

GuptaManan100:
## Description

This PR cleans up some of the constructs we use to parse Partitions. The AST struct for it has also be changed to be much leaner than before. Additional parsing support is also added. The following query is now parsed correctly - 
```sql
create table t1 (id int primary key) partition by list (id) (partition p1 values in(11,21), partition p2 values in (12,22))
```

## Related Issue(s)

- Fixes #9696 
- #8604 

## Checklist


GuptaManan100:
@shlomi-noach The PR is in a state where we now support the syntax that you requested in #9696. I want to add a little additional parsing support in PARTITIONS and SUBPARTITIONS. I can do that in a second following PR too if this needs to be merged before.

GuptaManan100:
Marking this ready for review, additional parsing support will be added later

shlomi-noach (approved):
Wow, amazing work! Let's push subpartitioning to later on in a different PR

PR #9718, opened 2022-02-17
Remove the DurabilityParams argument to the durability interface

GuptaManan100:
## Description

The DurabilityParams argument was added to the Durability interface for testing purposes. Currently we are in a state wherein we have can specify that map from vtorc but not from any other place. Therefore we are removing this parameter.

## Checklist
- [x] Should this PR be backported? I think so

### DIFF
--- a/go/cmd/vtctl/vtctl.go
+++ b/go/cmd/vtctl/vtctl.go
@@ -95,7 +95,7 @@ func main() {
 		log.Warningf("cannot connect to syslog: %v", err)
 	}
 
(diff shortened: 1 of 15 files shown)


GuptaManan100:
@deepthi I am not sure whether we want to back port this to release 13. That will be the first release where we introduce the durability interface in vtctl, vtctld etc. Since its pluggable, I think we should back port this change so that the interface is consistent for these binaries across releases. WDYT?

PR #9724, opened 2022-02-17
Fix `make embed_config` target

matthiasr:
## Description
Make executes each line in its own shell, so the `cd` had no effect on
the subsequent `go` invocations. Fix this by making each line `cd`
individually.

## Related Issue(s)

none

## Checklist
- [x] Should this PR be backported? - NO, only relevant for development

## Deployment Notes

None – development workflow fix only

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,9 +49,8 @@ export CGO_CFLAGS := -O1
 
 # regenerate rice-box.go when any of the .cnf files change
 embed_config:
-	cd go/vt/mysqlctl
-	go run github.com/GeertJohan/go.rice/rice embed-go
-	go build .
+	cd go/vt/mysqlctl && go run github.com/GeertJohan/go.rice/rice embed-go
+	cd go/vt/mysqlctl && go build .
 
 # build the vitess binaries with dynamic dependency on libc
 build-dyn:


Makefile
```diff
@@ -49,9 +49,8 @@ export CGO_CFLAGS := -O1
 
 # regenerate rice-box.go when any of the .cnf files change
 embed_config:
-	cd go/vt/mysqlctl
-	go run github.com/GeertJohan/go.rice/rice embed-go
-	go build .
+	cd go/vt/mysqlctl && go run github.com/GeertJohan/go.rice/rice embed-go
+	cd go/vt/mysqlctl && go build .
```
ajm188, line 53: any reason we don't one-line this?
```suggestion
	cd go/vt/mysqlctl && go run github.com/GeertJohan/go.rice/rice embed-go && go build .
```

matthiasr: I find it easier to read what is happening top-to-bottom, since the prefix is always the same and easy to ignore. If the chain gets longer, you would start breaking the one-liner up with `\` escaped newlines, but I've seen some hard to debug failures because you forgot one of those.

In this case, it doesn't really matter, so whichever way you prefer!

ajm188: Doesn't matter much to me either! Let's keep it the way you have it 😊

ajm188 (approved):
Thanks for the fix! I'll merge once CI passes.

matthiasr:
Many of the checks got canceled, and I don't seem to be able to retrigger them?

ajm188:
Hmm, I'm not sure why that happened, and the only way I can re-trigger is to click for each one individually ... can you no-op amend your previous commit and force push? That should do it.

matthiasr:
uuh I think I accidentally based that on the wrong branch, please hold …

matthiasr:
@ajm188 it seems like the rebase makes it so your workflow approval is required again … it's possible that I changed the commit to my work email. Could you approve CI again?

frouioui (approved):
LGTM

frouioui:
@matthiasr, thank you for your contribution 🙇🏻

PR #9725, opened 2022-02-17
Change semi-sync information to use the parameter passed and deprecate enable_semi_sync

GuptaManan100:
## Description

This PR deprecates `enable_semi_sync` flag and changes the implementation for `fixSemiSync` to use the flag passed down from the RPC. `Backup` and `RestoreFromBackup` RPC's are also fixed to call `SetReplicationSource` when they finish. This is needed since the primary could have changed during the backup or restore process and the semi-sync settings need to be updated.

## Related Issue(s)

#8975 

## Checklist
- [x] Should this PR be backported? *Some Part* -> The change to the RPCs needs to be backported to maintain backward compatibility during upgrade downgrade process. Essentially we need vtctl on version 13 to work with vttablets from version 14. #9730 

## Deployment Notes

`enable_semi_sync` is deprecated and the user should be setting the correct `durability_policy` instead.


go/vt/wrangler/testlib/planned_reparent_shard_test.go
```diff
@@ -686,6 +689,10 @@ func TestPlannedReparentShardRelayLogErrorStartReplication(t *testing.T) {
 	// simulate error that will trigger a call to RestartReplication
 	goodReplica1.FakeMysqlDaemon.StartReplicationError = errors.New("Slave failed to initialize relay log info structure from the repository")
 	goodReplica1.FakeMysqlDaemon.ExpectedExecuteSuperQueryList = []string{
+		// In SetReplicationSource, we find that the source host and port was already set correctly,
+		// So we try to stop and start replication. The first STOP SLAVE comes from there
+		"STOP SLAVE",
+		// During the START SLAVE call, we find a relay log error, so we try to restart replication.
```
deepthi, line 695: Is the relay log error from `START` or `STOP`?

deepthi (approved):
Mostly LGTM. I had only one question.
Nice work!

go/vt/wrangler/testlib/planned_reparent_shard_test.go
GuptaManan100: We get it from `StartReplicaton` command to mysql. In our test environment we are using the `FakeMySQLDaemon` 
```go
// StartReplication is part of the MysqlDaemon interface.
func (fmd *FakeMysqlDaemon) StartReplication(hookExtraEnv map[string]string) error {
	if fmd.StartReplicationError != nil {
		return fmd.StartReplicationError
	}
	return fmd.ExecuteSuperQueryList(context.Background(), []string{
		"START SLAVE",
	})
}
```
and we set the `StartReplicationError` as -
```go
goodReplica1.FakeMysqlDaemon.StartReplicationError = errors.New("Slave failed to initialize relay log info structure from the repository")
```

So we get the error in StartReplication, but it isn't seen in the Execute Super Query List

PR #9727, opened 2022-02-17
VReplication V2 test: bypass newly added lag test which is causing flakiness in CI

rohit-nayak-ps:
## Description
Since adding the lag check during cutover the v2 test is being very flaky in CI. Locally it never fails. 

It seems that there is some side-effect of the test code which waits for flakiness. While that is researched/fixed, this PR bypasses the check for now. 

It also changes the number of retries to 1 for the v2 test since retries always fail

## Checklist

### DIFF
--- a/go/test/endtoend/vreplication/vreplication_test.go
+++ b/go/test/endtoend/vreplication/vreplication_test.go
@@ -65,6 +65,7 @@ const (
 
 	merchantKeyspace = "merchant-type"
 	maxWait          = 10 * time.Second
+	BypassLagCheck   = true // temporary fix for flakiness seen only in CI when lag check is introduced
 )
 
 func init() {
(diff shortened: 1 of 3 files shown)


go/test/endtoend/vreplication/vreplication_test.go
```diff
@@ -65,6 +65,7 @@ const (
 
 	merchantKeyspace = "merchant-type"
 	maxWait          = 10 * time.Second
```
mattlord, line 67: We may just need to increase the max wait time. 10 seconds is pretty low given how flaky GitHub actions can be. 🙂

mattlord (approved):

go/test/endtoend/vreplication/vreplication_test.go
```diff
@@ -65,6 +65,7 @@ const (
...
+	BypassLagCheck   = true // temporary fix for flakiness seen only in CI when lag check is introduced
```
deepthi, line 68: nit: this doesn't need to be exported, but if this is temporary, not a big deal.

deepthi (approved):

PR #9729, opened 2022-02-18
sqlparser: Adding IsText field to ~ColumnDefinition~ ColumnType

shlomi-noach:
## Description

When parsing ~~`ColumnDefinition`~~ `ColumnType`, the parser parses and assigns the `Charset` value, if present. However, there is some loss of information. Some columns types are not at all textual (like numbers: they don't have charsets), while others are textual but happen to have no `charset` in their definition (ie `c char(10) not null`).

~~`ColumnDefinition`~~ `ColumnType` has a `Charset string` field, which is populated when charset is defined. But when the field is empty, it is impossible to determine whether this was because there isn't any charset clause in the definition, or whether there _can't be_ any charset clause.

This PR adds a `IsText bool` (maybe better called `IsTextual`? whichever), that indicates "yes, this column could potentially have a charset" (chars, varchars, json, enum, set) or "no, there cannot be a charset" (numeric, temporal, etc.).

The parser is responsible to populate `IsText`.

This information is later useful when investigating a table schema (I have a use case for `schemadiff`).

Is this approach acceptable?

## Checklist

### DIFF
--- a/go/vt/sqlparser/ast.go
+++ b/go/vt/sqlparser/ast.go
@@ -1593,6 +1593,8 @@ type ColumnType struct {
 	Scale    *Literal
 
 	// Text field options
+	// IsText indicates whether the column has textual properties, thus if it is capable fo having a charset (irrespective of having or not having a declared charset)
+	IsText  bool
 	Charset string
 
 	// Enum values
--- a/go/vt/sqlparser/ast_equals.go
+++ b/go/vt/sqlparser/ast_equals.go
@@ -1303,6 +1303,7 @@ func EqualsRefOfColumnType(a, b *ColumnType) bool {
 	return a.Type == b.Type &&
 		a.Unsigned == b.Unsigned &&
 		a.Zerofill == b.Zerofill &&
+		a.IsText == b.IsText &&
 		a.Charset == b.Charset &&
 		EqualsRefOfColumnTypeOptions(a.Options, b.Options) &&
 		EqualsRefOfLiteral(a.Length, b.Length) &&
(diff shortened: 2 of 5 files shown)


GuptaManan100 (commented):
We have a `Type` field in the ColumnType struct too. Can't we use that to infer whether it is a textual column or not? so `IsText` can be a method instead of a field in the struct.
Is there a reason we need it to be a field?

shlomi-noach:
>  Can't we use that to infer whether it is a textual column or not?

We can do that. My thinking was to have a single place that knows best. Since the parser already determines when a `CHARSET` clause can exist, it is able to determine that the colunm is textual. If we create a `IsText()` function, then it need to list down all the types where the parser has already determined are textual. It's no bid deal at all and easily achievable, if you think it's more correct!

shlomi-noach:
@GuptaManan100 thinking about this some more, you're right and I can see it's not the parser's job. I also wonder if it's the AST's job, or perhaps I should make this completely external.

GuptaManan100 (approved):
@shlomi-noach I don't have strong opinions, this also makes sense, the parser is the one who authoritatively knows whether a given data structure can have charsets assigned or not (these are the ones that have `charset_opt` rule), so it should be the one assigning this field. Also, it isn't new for us to have redundant fields (in the sense that the same information can be gleaned by combining the other fields information).

shlomi-noach:
if implemented outside the parser, all it takes is something as simple as this:

```golang
var charsetTypes = map[string]bool{
	"CHAR":       true,
	"VARCHAR":    true,
	"TEXT":       true,
	"TINYTEXT":   true,
	"MEDIUMTEXT": true,
	"LONGTEXT":   true,
	"ENUM":       true,
	"SET":        true,
	"JSON":       true,
}

// IsTextual
func (...) IsTextual(c *ColumnDefinition) bool {
	return charsetTypes[strings.ToUpper(c.Type.Type)]
}
```

shlomi-noach:
@GuptaManan100 ok, I'm going to start with something external to `go/vt/sqlparser`. Let's both continue to think about this and we can always decide to push it into either the parser or the ast. I'll meanwhile close this PR.

GuptaManan100:
Oh sorry I misunderstood

PR #9733, opened 2022-02-18
[VEP-4, phase 1] Flag Deprecation Warnings

ajm188:
## Description

This PR introduces a helper function, `_flag.Parse` to check for the flag usages that will break when we switch to `flag`. For more details, see [VEP-4](https://github.com/vitessio/enhancements/blob/main/veps/vep-4.md).

The general plan here is:
1. Merge this. If I'm fast enough, get a second branch into 13.0. If not, adjust all versions in the next steps by 1.
2. Adjust scripts in examples/ directory so they stop issuing warnings.
3. Drop-in replace all calls to `_flag.Parse` with `pflag.Parse`. Verify all binaries start correctly.
4. Delete this package.

### Testing:

Here's the local example:

```
➜  local git:(andrew/flag-deprecation) ✗ ./101_initial_cluster.sh 
add /vitess/global
add /vitess/zone1
add zone1 CellInfo
W0217 09:40:27.230091   62398 flag.go:30] Use of single-dash long flags is deprecated and will be removed in the next version of Vitess. Please use --topo_global_root instead
W0217 09:40:27.230545   62398 flag.go:30] Use of single-dash long flags is deprecated and will be removed in the next version of Vitess. Please use --topo_global_server_address instead
W0217 09:40:27.230549   62398 flag.go:30] Use of single-dash long flags is deprecated and will be removed in the next version of Vitess. Please use --topo_implementation instead
W0217 09:40:27.230552   62398 flag.go:43] Detected a positional argument beginning with a dash; This will be treated as a flag argument in the next version of Vitess. Please update your invocation to include a "--" before to continue treating --root as a positional argument.
W0217 09:40:27.230567   62398 flag.go:43] Detected a positional argument beginning with a dash; This will be treated as a flag argument in the next version of Vitess. Please update your invocation to include a "--" before to continue treating --server-address as a positional argument.
W0217 09:40:27.230578   62398 vtctl.go:88] WARNING: vtctl should only be used for VDiff workflows. Consider using vtctldclient for all other commands.
Created cell: zone1
```

## Related Issue(s)

https://github.com/vitessio/enhancements/issues/7

## Checklist
- [x] Should this PR be backported? yes, opening a branch now
- [x] Tests were added or are not required **n/a** -- did manual testing

### DIFF
--- a/go/internal/flag/flag.go
+++ b/go/internal/flag/flag.go
@@ -25,7 +25,6 @@ package flag
 
 import (
 	goflag "flag"
-	"fmt"
 	"os"
 	"reflect"
 	"strings"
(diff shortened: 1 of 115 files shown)


go/internal/flag/flag.go
```diff
@@ -0,0 +1,156 @@
...
+			buf.WriteString("--")
+			buf.WriteString(f.Name)
+			if name != "" {
+				fmt.Fprintf(&buf, " %s", name)
+			}
+			buf.WriteString("\n\t")
+
+			// Second line: usage and optional default, if not the zero value
+			// for the type.
+			buf.WriteString(usage)
```
ajm188, line 94: If you, like me, were wondering, _but what if the usage is empty? does the standard library really still print a tab character?_ oh boy does it!!

Screen Shot 2022-02-20 at 9 34 51 AM

([playground](https://go.dev/play/p/rbjzkOukVbI))

go/internal/flag/flag.go
```diff
@@ -0,0 +1,156 @@
...
+
+	// Check and warn for any mixed posarg / dashed-arg on the CLI.
+	posargs := goflag.Args()
+	for _, arg := range posargs {
+		if arg == "--" {
+			break
+		}
+
+		if strings.HasPrefix(arg, "-") {
+			log.Warningf("Detected a positional argument beginning with a dash; this will be treated as a flag argument in the next version of Vitess. Please update your invocation to include a \"--\" before to continue treating %s as a positional argument.", arg)
```
ajm188, line 128: @deepthi let me know if you have suggestions on how to improve this warning text. i don't think it's clear, but i can't think of a way to improve it, either :/

go/internal/flag/flag.go
```diff
@@ -0,0 +1,156 @@
...
+func Parse() {
+	// First, override the Usage func to make flags show in their double-dash
+	// forms to the user.
+	goflag.CommandLine.Usage = func() {
+		fmt.Fprintf(goflag.CommandLine.Output(), "Usage of %s:\n", os.Args[0])
+
+		var buf strings.Builder
+		goflag.CommandLine.VisitAll(func(f *goflag.Flag) {
+			defer buf.Reset()
+			defer func() { fmt.Fprintf(goflag.CommandLine.Output(), "%s\n", buf.String()) }()
```
ajm188, line 53: we have to wrap this defer inside an anonymous function to delay the execution of `buf.String()`

the other thing i briefly considered was:

```go
defer buf.Reset()

...

if bf, ok := f.Value.(maybeBoolFlag); ok && ... {
    fmt.Fprintf(...)
    goto print
}

// print first and second lines for non-bool flags to `buf`

print:
    fmt.Fprintf(goflag.CommandLine.Output(), "%s\n", buf.String())
```

but ended up deciding the goto was more gross than the `defer func() { ... }()`

go/internal/flag/flag.go
frouioui: I might be overthinking, but to make it clearer I would add:
`include a \"--\" before to ...` -> `include a \"--\" before %s to ...` with `%s` being `arg`

go/internal/flag/flag.go
frouioui: haha good catch

go/internal/flag/flag.go
frouioui: IMO, the `defer` does not seem bad. Another alternative would be to create an anonymous function where we have the current `defer`, and call that function where we `return`.

frouioui (approved):
I am excited for this VEP, phase 1 is looking good to me @ajm188  🚀  

We should update the `v14`/`v13` docs to reflect this deprecation and how the new flags should be written.

ajm188:
Ugh, okay this is breaking a lot of the cluster tests that check the "first line" output of vtctl because those first few lines are now all deprecation warnings ... I'm going to fix up all their invocations and see if that helps without breaking upgrade/downgrade testing.

ajm188:
🤞 🤞

go/internal/flag/flag.go
deepthi: I like this suggestion 👍

deepthi (commented):
We should change all the examples in the same PR. Otherwise the output will be full of warnings and difficult to parse by humans (unless you plan to fast-follow with an examples PR and want to keep PRs small and manageable).

ajm188:
I'm gonna DoNotMerge this, since I think there's a few more things to do, in particular:
- [ ] update all the tests to not break
- [ ] update the example scripts
- [ ] write the release notes

when i finish those 3 things i'll un-DoNotMerge and ping you again. sound good?

ajm188:
Alright, [hold on to your butts](https://www.youtube.com/watch?v=HKK4KmDlj8U), I'm force-pushing with what will, pending some back-and-forth on the release notes, be the final version of this change. I did all my debugging over in https://github.com/vitessio/vitess/pull/9831, which I will close.

deepthi (approved):
Nice work!

PR #9734, opened 2022-02-18
[backport] Flag Deprecation (13.0)

ajm188:
This back ports #9733 to release-13.0.

### DIFF
--- a/go/internal/flag/flag.go
+++ b/go/internal/flag/flag.go
@@ -1,0 +1,46 @@
+// Package flag is an internal package to allow us to gracefully transition
+// from the standard library's flag package to pflag. See VEP-4 for details.
+//
+// In general, this package should not be imported or depended on, except in the
+// cases of package servenv, and entrypoints in go/cmd. This package WILL be
+// deleted after the migration to pflag is completed, without any support for
+// compatibility.
+package flag
+
+import (
+	goflag "flag"
+	"os"
+	"strings"
+
+	"vitess.io/vitess/go/vt/log"
+)
+
+// Parse wraps the standard library's flag.Parse to perform some sanity checking
+// and issue deprecation warnings in advance of our move to pflag.
+//
+// See VEP-4, phase 1 for details: https://github.com/vitessio/enhancements/blob/c766ea905e55409cddeb666d6073cd2ac4c9783e/veps/vep-4.md#phase-1-preparation
+func Parse() {
+	goflag.Parse()
+
+	// Check and warn on any single-dash flags.
+	argv := os.Args
+	goflag.Visit(func(f *goflag.Flag) {
+		for _, arg := range argv {
+			if strings.HasPrefix(arg, "-"+f.Name) {
+				log.Warningf("Use of single-dash long flags is deprecated and will be removed in the next version of Vitess. Please use --%s instead", f.Name)
+			}
+		}
+	})
+
+	// Check and warn for any mixed posarg / dashed-arg on the CLI.
+	posargs := goflag.Args()
+	for _, arg := range posargs {
+		if arg == "--" {
+			break
+		}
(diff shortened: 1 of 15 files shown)


go/internal/flag/flag.go
```diff
@@ -0,0 +1,46 @@
+// Package flag is an internal package to allow us to gracefully transition
```
deepthi, line 1: In theory we need a copyright section here (though we are in a bit of flux around what that will look like).

deepthi (approved):
Approving provisionally (assuming no real test failures).

deepthi:
Second thoughts:
* What happens to the help output? Right now we print it out with a single `-` in front of each flag
* I see this in the main PR
```
W0217 09:40:27.230552   62398 flag.go:43] Detected a positional argument beginning with a dash; This will be treated as a flag argument in the next version of Vitess. Please update your invocation to include a "--" before to continue treating --root as a positional argument.
W0217 09:40:27.230567   62398 flag.go:43] Detected a positional argument beginning with a dash; This will be treated as a flag argument in the next version of Vitess. Please update your invocation to include a "--" before to continue treating --server-address as a positional argument.
```
That doesn't seem right. The command invocation is:
```
vtctl $TOPOLOGY_FLAGS VtctldCommand AddCellInfo \
  --root /vitess/$cell \
  --server-address "${ETCD_SERVER}" \
  $cell
```
which is using the flags correctly.

ajm188:
> which is using the flags correctly.

it's not. the correct usage there is to defer those double-dashed arguments to the subcommand, so the correct way to do this with `pflag` is _any of_:
* `vtctl $TOPOLOGY_FLAGS VtctldCommand AddCellInfo -- --server-address ...` OR
* `vtctl $TOPOLOGY_FLAGS VtctldCommand -- AddCellInfo --server-address ...` OR
* `vtctl $TOPOLOGY_FLAGS -- VtctldCommand AddCellInfo --server-address ...`

deepthi:
> > which is using the flags correctly.
> 
> it's not. the correct usage there is to defer those double-dashed arguments to the subcommand, so the correct way to do this with `pflag` is _any of_:
> 
> * `vtctl $TOPOLOGY_FLAGS VtctldCommand AddCellInfo -- --server-address ...` OR
> * `vtctl $TOPOLOGY_FLAGS VtctldCommand -- AddCellInfo --server-address ...` OR
> * `vtctl $TOPOLOGY_FLAGS -- VtctldCommand AddCellInfo --server-address ...`

Ah I missed that these go to the `AddCellInfo`. So that part is good.

deepthi:
OK, I went back and looked at the sample output. This also looks correct.
```
W0217 09:40:27.230091   62398 flag.go:30] Use of single-dash long flags is deprecated and will be removed in the next version of Vitess. Please use --topo_global_root instead
W0217 09:40:27.230545   62398 flag.go:30] Use of single-dash long flags is deprecated and will be removed in the next version of Vitess. Please use --topo_global_server_address instead
W0217 09:40:27.230549   62398 flag.go:30] Use of single-dash long flags is deprecated and will be removed in the next version of Vitess. Please use --topo_implementation instead
```

That leaves only the question about the help. Do we print out help text with `-` or `--` with this change?

ajm188:
labeling this with DoNotMerge as i've pushed up more changes to the original branch, and i'm skeptical i will get this done in time for the 13.0 release 😞

ajm188:
> That leaves only the question about the help. Do we print out help text with `-` or `--` with this change?

I figured out how to work with the flag internals, and we are now printing out `--` with this change (in the upstream version, I'm working on updating the backport branch now)

ajm188:
`v13` has already been created, so i'm closing this and will get it out for `v14`

PR #9739, opened 2022-02-18
Add CI workflow for vtadmin-web unit tests

doeg:

## Description

Well this is an embarrassing oversight! 🤡 Better late than never!

Here's what it looks like when it fails on purpose: https://github.com/tinyspeck/vitess/runs/5254002448?check_suite_focus=true

**Edited to add:** I also switched our CI tests from using `npm install` to using `npm ci`, since... this is the exact use case for `npm ci`. :) 

## Related Issue(s)

N/A

## Checklist
- [x] Tests were added or are not required **Many tests were added!**

## Deployment Notes
N/A

### DIFF
--- a/.github/workflows/vtadmin_web_unit_tests.yml
+++ b/.github/workflows/vtadmin_web_unit_tests.yml
@@ -32,3 +32,8 @@ jobs:
 
       - name: Run unit tests
         run: CI=true npm run test
+        
+    # Cancel pending and in-progress runs of this workflow if a newer ref is pushed to CI.
+    concurrency: 
+      group: ${{ github.workflow }}-${{ github.ref }}
+      cancel-in-progress: true
--- a/.github/workflows/vtadmin_web_unit_tests.yml
+++ b/.github/workflows/vtadmin_web_unit_tests.yml
@@ -28,7 +28,7 @@ jobs:
           node-version: '16.13.0'
 
       - name: Install dependencies
-        run: npm install
+        run: npm ci
 
       - name: Run unit tests
         run: CI=true npm run test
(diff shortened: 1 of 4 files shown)


.github/workflows/vtadmin_web_unit_tests.yml
```diff
@@ -0,0 +1,34 @@
...
+on:
+  push:
+    paths:
+      - '.github/workflows/vtadmin_web_unit_tests.yml'
+      - 'web/vtadmin/**'
+  pull_request:
+    paths:
+      - '.github/workflows/vtadmin_web_unit_tests.yml'
+      - 'web/vtadmin/**'
+
```
frouioui, line 14: We should use GitHub Actions' [`concurrency` feature](https://docs.github.com/en/actions/using-jobs/using-concurrency) here. It allows us to reduce the number of concurrent jobs for the same pull request. Let's say you have workflow `A` for commit `n`, if you push a commit `n+1` before `A` had time to finish, the concurrency feature will cancel `A` for commit `n`, and only run `A` for commit `n+1`.

I suggest the following chunk: (which is similar to other workflows like [here](https://github.com/vitessio/vitess/blob/dbbf84b9eaec42bd0b56c5de9a47b3f6d35cf09b/.github/workflows/cluster_endtoend_12.yml#L5) for example)
```yaml
concurrency:
  group: format('{0}-{1}', ${{ github.ref }}, 'vtadmin-web-unit-tests')
  cancel-in-progress: true
```

The `group` key is made unique by concatenating the branch (`github.ref`) and workflow names.

frouioui (approved):

.github/workflows/vtadmin_web_unit_tests.yml
ajm188: sweeeeeeet

PR #9743, opened 2022-02-21
Improve explanation of MySQLTopologyUser and MySQLReplicaUser so that their difference is clearer

GuptaManan100:
## Description

This PR just adds 2 comments explaining MySQLTopologyUser and MySQLReplicaUser parameters for VTOrc, so that the difference between the two is clearer.

## Checklist


deepthi (approved):

deepthi:
Let's get @shlomi-noach's review on this too.

shlomi-noach (approved):
This is very useful!

PR #9744, opened 2022-02-21
Increase async IO threads count in CI

GuptaManan100:
## Description

This PR adds a code to increase the permissible async IO thread count in the linux machine running the CI. Without this some of the tests fail because they are unable to start the large number of MySQL instances the test requires.

## Checklist

### DIFF
--- a/.github/workflows/cluster_endtoend_11.yml
+++ b/.github/workflows/cluster_endtoend_11.yml
@@ -28,6 +28,9 @@ jobs:
     - name: Tune the OS
       run: |
         echo '1024 65535' | sudo tee -a /proc/sys/net/ipv4/ip_local_port_range
+        # Increase the asynchronous non-blocking I/O. More information at https://dev.mysql.com/doc/refman/5.7/en/innodb-parameters.html#sysvar_innodb_use_native_aio
+        echo "fs.aio-max-nr = 1048576" | sudo tee -a /etc/sysctl.conf
+        sudo sysctl -p /etc/sysctl.conf
 
     - name: Check out code
       uses: actions/checkout@v2
--- a/test/templates/unit_test.tpl
+++ b/test/templates/unit_test.tpl
@@ -17,6 +17,9 @@ jobs:
     - name: Tune the OS
       run: |
         echo '1024 65535' | sudo tee -a /proc/sys/net/ipv4/ip_local_port_range
+        # Increase the asynchronous non-blocking I/O. More information at https://dev.mysql.com/doc/refman/5.7/en/innodb-parameters.html#sysvar_innodb_use_native_aio
+        echo "fs.aio-max-nr = 1048576" | sudo tee -a /etc/sysctl.conf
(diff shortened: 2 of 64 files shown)


.github/workflows/cluster_endtoend_11.yml
```diff
@@ -28,6 +28,9 @@ jobs:
     - name: Tune the OS
       run: |
         echo '1024 65535' | sudo tee -a /proc/sys/net/ipv4/ip_local_port_range
+        # Increase the asynchronous non-blocking I/O. More information at https://dev.mysql.com/doc/refman/5.7/en/innodb-parameters.html#sysvar_innodb_use_native_aio
```
mattlord, line 31: You need to use `echo -e` if you don’t simply want the \n characters printed.

test/templates/unit_test.tpl
```diff
@@ -17,6 +17,9 @@ jobs:
     - name: Tune the OS
       run: |
         echo '1024 65535' | sudo tee -a /proc/sys/net/ipv4/ip_local_port_range
+        # Increase the asynchronous non-blocking I/O. More information at https://dev.mysql.com/doc/refman/5.7/en/innodb-parameters.html#sysvar_innodb_use_native_aio
+        echo "\nfs.aio-max-nr = 1048576" >> /etc/sysctl.conf
```
mattlord, line 21: You need to use `echo -e` if you don’t simply want the \n characters printed but instead the new line escape character.

test/templates/unit_test.tpl
```diff
@@ -17,6 +17,9 @@ jobs:
...
+        sysctl -p /etc/sysctl.conf
```
mattlord, line 22: IMO we should simply set the value in the running kernel. IF the same machine is used after rebooting then we’d have this line in the file N times. Any reason not to simply do this:
```
sysctl -w nfs.aio-max-nr=1048576
```

This sets it in the kernel and is idempotent. The file is nice for permanent changes in machines you manage (in our case we have to setup the execution environment each time right?).

mattlord (commented):
Thank you for working on this! I think that this is likely at least one cause of problems seen by @rohit-nayak-ps with the vreplication_v2 tests when not using docker.

test/templates/unit_test.tpl
GuptaManan100: Yes, I was looking for something like this! All the CI's anyways failed because we can't touch the `etc/sysctl.conf` file without root privilege

GuptaManan100: No this doesn't work, `nfs.aio-max-nr` isn't a configuration that can be changed directly via sysctl

test/templates/unit_test.tpl
GuptaManan100: fixed this by using the same way as in the line before it, using `tee`

mattlord (approved):
🙇‍♂️

PR #9745, opened 2022-02-21
Improve upgrade downgrade workflows' strategy

frouioui:
## Description

@GuptaManan100 and I paired up on this pull request to improve the upgrade downgrade testing workflows by choosing the latest release in a better way.

Before, every commit would be tested against the last major release, even if this commit is made on, for instance, the `release-12.0` branch. ~~Now, commits made against a release branch are tested against the previous release branch (commits to `release-13.0` will be tested against `release-12.0`) ; and commits made to `main` and other branches are tested against the latest release branch.~~
Now, commits made against a release branch are tested against the previous releases' latest patch release and commits made to `main` and other branches are tested against the last major release tag unless a new release brach has been created without a tag, in which case that is used.

For example - 
| PR head | Tested against |
| ------- | ------- |
| `release-13.0` | `v12.0.3` |
| `release-12.0` | `v11.0.4` |
| `main` | `v13.0.0` | 
| `main` (after we create `release-14.0` but before creating `v14.0.0`) | `release-14.0` |

Needs to be backported to all three latest releases, but does not need to make it to `v13.0.0 GA`.

## Checklist

### DIFF
--- a/.github/workflows/upgrade_downgrade_test_backups_e2e.yml
+++ b/.github/workflows/upgrade_downgrade_test_backups_e2e.yml
@@ -25,14 +25,13 @@ jobs:
           label: Skip Upgrade Downgrade
 
   get_latest_release:
-    timeout-minutes: 10
     if: always() && (github.event_name != 'pull_request' || needs.get_upgrade_downgrade_label.outputs.hasLabel != 'true')
     name: Get latest release
     runs-on: ubuntu-latest
     needs:
       - get_upgrade_downgrade_label
     outputs:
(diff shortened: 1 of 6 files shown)


GuptaManan100 (approved):

frouioui:
Can't approve my own PR. But I approve @GuptaManan100's latest changes.

deepthi (commented):
My only comment is that we seem to be duplicating config in each upgrade/downgrade test.
Maybe we should write our own custom action and use that for all of these tests.

PR #9747, opened 2022-02-21
Query plan/rules: apply for multi-table statement

shlomi-noach:
## Description

Issue found: when query rules are applied to a table (e.g. `UpdateSourceDeniedTables`), query executor still lets some queries run and operate on the applied table. Specifically, it allows `SELECT`/`UPDATE`/`DELETE` queries that reference multiple tables to go unchecked.

Say table `corder` is denied. the following query gets rejected:
```sql
select * from corder;
ERROR 1105 (HY000) at line 1: target: commerce.0.primary: vttablet: rpc error: code = FailedPrecondition desc = disallowed due to rule: enforce denied tables (CallerID: userData1)
```
while the following query is allowed:
```sql
select * from corder, customer
```

As it turns out, the plan completely ignores the rules if it does not have exactly one referenced table. This is a bug and the behavior is undesired.

In this PR the query executor checks all plan's tables against all rules. It's enough that one rule conflicts with one table, and the query gets rejected.

The code affected by this PR is used in many places. To make the changes minimal, I kept backward-compatibility. Specifically, you will find that:

- `Plan` keeps the old `Table` field, and we add `AllTables` on top.

```golang
type Plan struct {
	// When the query indicates a single table
	Table *schema.Table
	// SELECT, UPDATE, DELETE statements may list multiple tables
	AllTables []*schema.Table
...
```

- I used variadic variables (`tableNames ...string`) as opposed to a slice, so that most existing code can keep calling functions with a single table name

```golang
func (qri *Map) FilterByPlan(query string, planid planbuilder.PlanType, tableNames ...string) (newqrs *Rules) {
```

This bug affects `vitess` migrations: the cut-over step marks the migrated table as denied; there must not be any `insert`s, `delete`s or `update`s to the table during the cut-over time. But the bug allows multi-table `update`s and `delete`s to pass through.

## Related Issue(s)

- #6926 

## Checklist

### DIFF
--- a/go/vt/vttablet/tabletserver/rules/rules_test.go
+++ b/go/vt/vttablet/tabletserver/rules/rules_test.go
@@ -181,6 +181,43 @@ func TestFilterByPlan(t *testing.T) {
 	if got != want {
 		t.Errorf("qrs1:\n%s, want\n%s", got, want)
 	}
+	{
+		// test multiple tables:
+		qrs1 := qrs.FilterByPlan("insert", planbuilder.PlanSelect, "a", "other_table")
+		want := compacted(`[{
+			"Description":"rule 2",
+			"Name":"r2",
+			"BindVarConds":[{
+				"Name":"a",
+				"OnAbsent":true,
+				"Operator":""
+			}],
+			"Action":"FAIL"
+		}]`)
+		got = marshalled(qrs1)
+		if got != want {
+			t.Errorf("qrs1:\n%s, want\n%s", got, want)
+		}
+
+	}
+	{
+		// test multiple tables:
+		qrs1 := qrs.FilterByPlan("insert", planbuilder.PlanSelect, "other_table", "a")
+		want := compacted(`[{
+			"Description":"rule 2",
+			"Name":"r2",
+			"BindVarConds":[{
+				"Name":"a",
+				"OnAbsent":true,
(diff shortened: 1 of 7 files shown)


go/vt/vttablet/tabletserver/rules/rules_test.go
```diff
@@ -181,6 +181,43 @@ func TestFilterByPlan(t *testing.T) {
...
+				"Name":"a",
+				"OnAbsent":true,
+				"Operator":""
+			}],
+			"Action":"FAIL"
+		}]`)
+		got = marshalled(qrs1)
+		if got != want {
+			t.Errorf("qrs1:\n%s, want\n%s", got, want)
+		}
```
GuptaManan100, line 219: For all of these we want to use `require` or `assert`

GuptaManan100 (approved):

go/vt/vttablet/tabletserver/rules/rules_test.go
shlomi-noach: right, we need to convert this entire test suite to use `testify`.

shlomi-noach:
Test failures are legit and I'm looking into.

shlomi-noach:
Fixed failing tests issue

PR #9749, opened 2022-02-21
evalengine: Decimals v2 Electric Boogaloo

vmg:
## Description

Alright we're back with some technical debt cleanup. For those following at home, you'll remember we introduced `DECIMAL` support for the `evalengine` a couple months ago. It was a hard requirement to be able to perform pretty much any kind of arithmetic divisions in the engine, as these are always performed as `DECIMAL` types.

In order to implement support for this data type, I had to come up with an arbitrary precision decimal number arithmetic library. I tried various existing OSS alternatives written in Go, and none of them had behaviors that really matched MySQL's. So, I ended up settling with https://github.com/ericlagergren/decimal (because it had the best benchmarks results I could find), which I vendored in our repository and started hacking into its guts until its behavior passed all the integration tests I cooked up for MySQL's behavior.

In retrospect, that was not the right choice. Eric Langergren's decimal library implements the [GDA](http://speleotrove.com/decimal/) spec to a tee -- including support for all kinds of whacky stuff like payloads in NaNs, trigonometrics, decomposition, series... MySQL, on the other hand, it don't do none of that shit. So I had to hack _a lot_ out of the original implementation and do a lot of very questionable things to make it match MySQL's behavior. It took weeks of effort, and after all that, we still had more than 10000 lines of Decimal code (yes, that's ten thousand) that I still don't fully understand to this day.

But, this last Friday, I got thinking: yes, I don't understand most of our `decimal` implementation at all. But I do understand (now) the changes I had to do to make it match MySQL's behavior. So what would happen if I pick the simplest `decimal` implementation I can find and apply the changes there?

This is the result of this effort, a very exciting `-10931 Deletions` PR, based on Spring's https://github.com/shopspring/decimal. It now occupies a single file, and matches MySQL's behavior with _fewer changes_ than the ones that I had to implement for the old library.

To top it off, it's actually _faster_ where it actually matters: converting decimals from and to strings, which are the most frequent part of decimal evaluation in MySQL. The new code was easy enough to understand that I could optimize all these cases quite trivially. :sparkles:

## Checklist


shlomi-noach:
> -10931 Deletions

Love it! :heart:

frouioui (approved):
I don't have anything to say, thanks for leaving helpful comments 👍🏻 
looks good to me!

PR #9751, opened 2022-02-21
Use latest vitess-operator example CRDs

frouioui:
## Description

This pull request uses the latest CRDs of the vitess-operator following the recent changes made in vitess-operator. The `operator.yaml` is now usable with the most recent version of Kubernetes that vitess-operator supports `1.22.*` as of today.

Additionally, a minor fix in the `do_release` script is pushed in this pull request. When updating the Docker tag in the vitess-operator examples, the `v` of `vX.X.X` was missing.

The entire vitess-operator example was tested on both `main` and `release-13.0`.

## Checklist


deepthi (approved):
Nice work!

deepthi (approved):

PR #9755, opened 2022-02-22
Online DDL vitess migration's cut-over: query buffering

shlomi-noach:
## Description

This PR implements query buffering for `vitess` (vreplication) migration cut-over phase, and replace the existing mechanism.

The main problem with the current cut-over mechanism is that during cut-over time we _reject_ queries based on a Deny list (implemented by deny rules). 

What we want to achieve is to _buffer_ queries rather than reject them, such that a query will stall until once cut-over is done, then proceed to run as if everything is normal. To the user this will seen as a latency spike, but no errors should be visible. Of course, we can only buffer _so many_ queries/connections; if the app hammers the database too much, then new incoming queries may be rejected.

In this PR I had to work around two major impediments of the current design:

1. Current way of blocking queries is to tell `topo` that a table is in the deny list. This then propagates the information to all shards.
  - This is wasteful! We don't need to inform all tables in all shards about this. We only need to block queries on the current shard. In fact, we only need to block queries on the Primary tablet of the current shard.
  - Every shard that cut-over puts the table under deny list for all shards, which means that the table will be unavailable on each shard `n` times, assuming number of shards is `n`
2. Currently, rules (ACLs) are evaluated once when evaluating query plan. The query executor then checks the plan, thus the rules, before running the query. The rules remain static. We need a way of saying "buffer the query, still buffer it, still buffer it, unbuffer it". So we need dynamic rules.

#### The following changes in this PR are of particular interest

- We introduce a new query rule: `QRBuffer`
- Dynamic rules: a `Rule` now has a `isCancelledFunc` member, a function that determines whether the rule still applies or if it's cancelled ("expired", if you will)
  `Rule.Copy()` makes sure to copy this callback function.
- `tabletserver`-scope rules: fortunately, both `onlineddl.Executor` and `QueryEngine` run within the same `TabletServer` space. There is no need to delegate any information to `topo`, not to any other shards, not to any other tablet, and not even to `TabletManager`. 
  - `TabletServer` initializes `onlineddl.Executor` with a "buffer/unbuffer queries" callback function. `onlineddl.Executor` is responsible to call it in the correct time (buffer before cut-over, release after cut-over)
  - `TabletServer` implements the callback function, and assigns/deassigns a `QRBuffer` rule to the given table
  - `TabletServer` initializes the rule with the `cancelFunc`, which checks whether buffering rules still exist for the table
  - `QueryExecutor` respects the new `QRBuffer` rule; if a query is ruled to be buffered, then it enters a loop with `1sec` interval, expiring at `10sec`. If buffering is cancelled before `10sec` have passed, the query proceeds to execute as normal. If `10sec` have passed and the query is still buffered, the query fails with error.

  With this:

  - Queries will be buffered during cut-over for up to `10sec`. A normal cut-over should only take around `3s-5s`, and `10s` is a lot of time. So no query should error during a migration.

#### Things to further improve:

- There is still a race condition because `QueryExecutor` first checks permissions, then proceeds to run a query. If rules change in between, then the query violates the rules. The solution I have in mind is on the MySQL side, in a later PR.
- We should actually only buffer _write_ queries, and let `SELECT` queries pass through. In a future PR.

## Related Issue(s)

- tracking: #6926 

## Checklist


shlomi-noach:
This PR extends https://github.com/vitessio/vitess/pull/9747, which I will merge shortly.

go/test/endtoend/onlineddl/revert/onlineddl_revert_test.go
```diff
@@ -889,11 +889,6 @@ func runSingleConnection(ctx context.Context, t *testing.T, done *int64) {
 		case 2:
 			err = generateDelete(t, conn)
 		}
-		if err != nil {
```
shlomi-noach, line 892: The tests now expect queries to be buffered, so the do not allow this error!

go/test/endtoend/onlineddl/revert/onlineddl_revert_test.go
```diff
@@ -889,11 +889,6 @@ func runSingleConnection(ctx context.Context, t *testing.T, done *int64) {
...
-			if strings.Contains(err.Error(), "disallowed due to rule: enforce denied tables") {
```
shlomi-noach, line 893: The tests now expect queries to be buffered, so the do not allow this error!

go/test/endtoend/onlineddl/vrepl_stress/onlineddl_vrepl_mini_stress_test.go
```diff
@@ -525,9 +525,7 @@ func runSingleConnection(ctx context.Context, t *testing.T, done *int64) {
 			err = generateDelete(t, conn)
 		}
 		if err != nil {
-			if strings.Contains(err.Error(), "disallowed due to rule: enforce denied tables") {
```
shlomi-noach, line 528: The tests now expect queries to be buffered, so the do not allow this error!

go/test/endtoend/onlineddl/vrepl_stress_suite/onlineddl_vrepl_stress_suite_test.go
```diff
@@ -664,9 +664,7 @@ func runSingleConnection(ctx context.Context, t *testing.T, done *int64) {
 			err = generateDelete(t, conn)
 		}
 		if err != nil {
-			if strings.Contains(err.Error(), "disallowed due to rule: enforce denied tables") {
```
shlomi-noach, line 667: The tests now expect queries to be buffered, so the do not allow this error!

go/vt/vttablet/onlineddl/executor.go
```diff
@@ -669,37 +670,24 @@ func (e *Executor) cutOverVReplMigration(ctx context.Context, s *VReplStream) er
 	}
 
 	// Preparation is complete. We proceed to cut-over.
-
-	// lock keyspace:
-	{
-		lctx, unlockKeyspace, err := e.ts.LockKeyspace(ctx, e.keyspace, "OnlineDDLCutOver")
```
shlomi-noach, line 675: No more talking to `topo`, no locking of the shard, we do everything internally t othe `tabletserver`

go/vt/vttablet/tabletserver/query_executor.go
```diff
@@ -387,14 +390,31 @@ func (qre *QueryExecutor) checkPermissions() error {
...
-	switch action {
-	case rules.QRFail:
-		return vterrors.Errorf(vtrpcpb.Code_INVALID_ARGUMENT, "disallowed due to rule: %s", desc)
-	case rules.QRFailRetry:
-		return vterrors.Errorf(vtrpcpb.Code_FAILED_PRECONDITION, "disallowed due to rule: %s", desc)
-	}
 
+	ctx, cancel := context.WithTimeout(qre.ctx, maxQueryBufferDuration)
+	defer cancel()
+	for goodToProceed := false; !goodToProceed; {
```
shlomi-noach, line 396: loop while buffering is active. If there's no buffering rule, this loop only executes once.

go/vt/vttablet/tabletserver/rules/rules.go
```diff
@@ -462,6 +476,9 @@ func (qr *Rule) GetAction(
 	bindVars map[string]*querypb.BindVariable,
 	marginComments sqlparser.MarginComments,
 ) Action {
+	if qr.isCancelledFunc != nil && qr.isCancelledFunc() {
```
shlomi-noach, line 479: `isCancelledFunc` makes the `Rue` dynamic; it can change behavior: the rule may expire.

shlomi-noach:
Ready for (careful) review! Please see original comment about how the solution in this PR uses a slightly (not terribly) different tabletserver-internal query rules paradigm. Please also note there's a known race condition which I leave for a different PR (the race condition already exists from before this PR and is not fixed here, though this PR contributes to the future fix's flow).

GuptaManan100 (commented):
Looks good to me! But I want others to take a look too

harshit-gangal (commented):
the query plans are cached. So, once the rule does not apply is it removed from the query plan.

go/vt/vttablet/tabletserver/query_executor.go
```diff
@@ -387,14 +390,31 @@ func (qre *QueryExecutor) checkPermissions() error {
...
+			return vterrors.Errorf(vtrpcpb.Code_INVALID_ARGUMENT, "disallowed due to rule: %s", desc)
+		case rules.QRFailRetry:
+			return vterrors.Errorf(vtrpcpb.Code_FAILED_PRECONDITION, "disallowed due to rule: %s", desc)
+		case rules.QRBuffer:
+			// We buffer up to some timeout. The timeout is determined by ctx.Done().
+			// If we're not at timeout yet, then we sleep for a short amount of time before
+			// resuming the loop.
+			select {
+			case <-time.After(time.Second):
+				// keep iteration
```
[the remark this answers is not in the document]
shlomi-noach: There will be potentially dozens, or hundreds, or thousands of waiting queries (up to max concurrent connections we allow in vtgate). So they will all be running this busy wait; this is why I picked a relatively relaxed interval.
we can expect cut-over time to be ~2s-3s.
WDYT?

shlomi-noach:
> the query plans are cached. So, once the rule does not apply is it removed from the query plan.

@harshit-gangal can you please look again at this code:

https://github.com/vitessio/vitess/pull/9755/files#diff-28fc9df9d28d69ca625b85e18a6fa652d61bfaf6de03012ce2d8d0338363d70eR240-R250

I believe it forces the rule into applying. What do you think?

The tests show that the behavior is correct: if I run some repeated consistent query against a table, then `ALTER` the table, the query gets blocked for the duration of cut-over, then resumes. If I run the same `ALTER` again, the query gets blocked again for the duration of cut-over, then resumes. This is true even when this applies:

https://github.com/vitessio/vitess/blob/29be10b93cfe556310397632049e2dbc81172242/go/vt/vttablet/tabletserver/query_engine.go#L294-L297

So I feel like this is a non-problem. However, can you point me to where in the code you think this might fail? :pray:

go/vt/vttablet/tabletserver/query_executor.go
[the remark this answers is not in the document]
frouioui: Personally, I think 1s is good. There will be at most one extra second to wait for the query to be executed. I think it is a good tradeoff between the query response time and the stress we put on VTGate.

go/vt/vttablet/tabletserver/rules/rules.go
```diff
@@ -237,6 +244,12 @@ func NewQueryRule(description, name string, act Action) (qr *Rule) {
 	return &Rule{Description: description, Name: name, act: act}
 }
 
+// NewQueryRule creates a new Rule.
+func NewBufferedTableQueryRule(tableName string, description string, isCancelledFunc func() bool) (qr *Rule) {
```
frouioui, line 248: nit:
```suggestion
// NewBufferedTableQueryRule creates a new Buffered Table Rule.
func NewBufferedTableQueryRule(tableName string, description string, isCancelledFunc func() bool) (qr *Rule) {
```

frouioui (approved):
The change looks good to me! 🙇🏻 I've left a few comments.

I am approving this Pull Request, however, I have limited knowledge on `Rule`s.

go/vt/vttablet/tabletserver/rules/rules.go
shlomi-noach: fixed, thank you!

frouioui: lgtm

go/vt/vttablet/tabletserver/query_executor.go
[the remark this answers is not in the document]
harshit-gangal: I think you are referring to VTTablet. which means the query is already sent down to the tablet. Just thinking in terms of reducing response time on the cutover is complete. This will also allow VTGate to send more queries down to Tablet which were blocked as the earlier query has not responded.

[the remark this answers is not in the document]
shlomi-noach: Right, but it comes with a cost: an increased busy wait on `tabletserver`, with potentially hundreds or thousands connections asking it the same question: "are we good to go?". So there has to be some tradeoff. I don't know what the exact number is; but if cut-over is expected to take some 3 seconds, then a 1sec wait interval isn't that excessive IMO. I'm good to lower this, say 500ms? We can iterate on that later.

harshit-gangal:
> > the query plans are cached. So, once the rule does not apply is it removed from the query plan.
> 
> @harshit-gangal can you please look again at this code:
> 
> https://github.com/vitessio/vitess/pull/9755/files#diff-28fc9df9d28d69ca625b85e18a6fa652d61bfaf6de03012ce2d8d0338363d70eR240-R250
> 
> I believe it forces the rule into applying. What do you think?
> 
> The tests show that the behavior is correct: if I run some repeated consistent query against a table, then `ALTER` the table, the query gets blocked for the duration of cut-over, then resumes. If I run the same `ALTER` again, the query gets blocked again for the duration of cut-over, then resumes. This is true even when this applies:
> 
> https://github.com/vitessio/vitess/blob/29be10b93cfe556310397632049e2dbc81172242/go/vt/vttablet/tabletserver/query_engine.go#L294-L297
> 
> So I feel like this is a non-problem. However, can you point me to where in the code you think this might fail? pray

This looks good.

go/vt/vttablet/tabletserver/query_executor.go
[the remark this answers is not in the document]
harshit-gangal: For this, I usually see how shard_buffer does it without using timeout. It would be nice we can have something similar. Otherwise the code looks good to me.

shlomi-noach:
After discussion with @harshit-gangal , the logic now switches from a cancelFunc to a `context.WithCancel` approach.

- Online DDL scheduler creates a `context.WithCancel` that is sent to the QRBuffer rule
- It is responsible to call `cancel()` when the cut-over is complete
- QueryExecutor now checks if the rule has a cancel context. If so, and rule is QRBuffer, then is blocks:
  - until the rule is cancelled (via `ctx.Done()`) - good, proceed with query
  - or until timeout is reached - fail and reject the query

harshit-gangal (approved):

go/vt/vttablet/tabletserver/query_executor.go
```diff
@@ -387,14 +390,31 @@ func (qre *QueryExecutor) checkPermissions() error {
 		remoteAddr = ci.RemoteAddr()
 		username = ci.Username()
 	}
-	action, desc := qre.plan.Rules.GetAction(remoteAddr, username, qre.bindVars, qre.marginComments)
+
+	bufferingTimeoutCtx, cancel := context.WithTimeout(qre.ctx, maxQueryBufferDuration)
```
malpani, line 394: this is apretty cool PR, i had a question - the fresh 10s timer applies on a per query basis. So if for some reason cutover is taking long - we will see 10s latency before the precondition failure. Is this intentional or should this instead be a shared context across all queries which will cause fail fast on preconditions for any cutover taking longer than 10s?

shlomi-noach: I'm not sure I understand the question. If the cutover is taking long we will see up to 10sec latency -- that's the expected behavior. A query must wait for up to 10sec. However, it does not wait the entire 10s if the cutover is either complete or failed. When that happens the wait is terminated prematurely.

malpani: Thanks i got the answer, but to add more details of my question

Scenario:
* Cutover took 60 seconds: Say it started at T0 and finished at T0+60s
* queries come in at the rate of 1 per second - ie. 1 query at T0, 1 at T0 + 1s ... 1 at T0+59s and so on

After this change:
* every query between T0 and T0+50s is rejected after 10 seconds latency
* queries between T0 + 51s and T0 + 59s will suceed due to the upto 10s buffer

My question was, would it instead be better to have a static buffer of 10s since cutover start ie. 
* any query between T0 and T0+10s gets buffered and rejected after 10s in this scenario
* queries between T0+11s and T0+59s get rejected immediately as buffer period of 10s since start of schema migration has expired
* queries coming in after T0+60s succeed as cutover has completed and write-block is cleared

I now understand that it is the former (just wanted to understand the behavior better)

shlomi-noach: Got it, thanks for clarifying! I'm open to alternatives if there's a clear advantage to one over the other, such as preserving resources. If you think a specific approach is better than another in all/most cases, then let's try and look at the numbers/advantage over a new issue!

Issue #9756, opened 2022-02-22
Bug Report: Lookup index with `uint32` fails when deleting entries with `lookup.Delete: Code: INTERNAL Type is not supported: "488" UINT32


akenneth:
### Overview of the Issue

When trying to delete records from a sharded table with a owned lookup table on an `unsigned int` field an error is thrown:
``` ruby
Trilogy::DatabaseError: trilogy_query_recv: 1105 lookup.Delete: Code: INTERNAL
Type is not supported: "488" UINT32
```

### Reproduction Steps

Using this vschema:

Customer_sharded.json
``` json
{
    "sharded": true,
    "vindexes": {
        "hash": {
            "type": "hash"
        },
        "things_on_order_id_ks_idx": {
            "type": "lookup",
            "params": {
              "table": "things_on_order_id_ks_idx",
              "from": "order_id,id",
              "to": "keyspace_id",
              "ignore_nulls": "true"
            },
            "owner": "things"
          }
    },
    "tables": {
        "customer": {
            "column_vindexes": [
                {
                    "column": "customer_id",
                    "name": "hash"
                }
            ],
            "auto_increment": {
                "column": "customer_id",
                "sequence": "customer_seq"
            }
        },
        "corder": {
            "column_vindexes": [
                {
                    "column": "customer_id",
                    "name": "hash"
                }
            ],
            "auto_increment": {
                "column": "order_id",
                "sequence": "order_seq"
            }
        },
        "things": {            
            "auto_increment": {
                "column": "id",
                "sequence": "things_seq"
            },
            "column_vindexes": [
                {
                "column": "order_id",
                "name": "hash"
                },                
                {
                "columns": [
                    "order_id",
                    "id"
                ],
                "name": "things_on_order_id_ks_idx"
                }
            ]            
        },
        "things_on_order_id_ks_idx": {
            "column_vindexes": [
              {
                "column": "order_id",
                "name": "hash"
              }
            ]
        }
        
    }
}
```

Commerce_seq.json
```
{
    "tables": {
        "customer_seq": {
            "type": "sequence"
        },
        "order_seq": {
            "type": "sequence"
        },
        "things_seq": {
            "type": "sequence"
        },
        "product": {}
    }
}
```

With this schema:
``` sql
create table product(
  sku varbinary(128),
  description varbinary(128),
  price bigint,
  primary key(sku)
) ENGINE=InnoDB;
create table customer(
  customer_id int unsigned not null auto_increment,
  email varbinary(128),
[...]

PR #9757, opened 2022-02-22
Add support for uint32 bind var + test

akenneth:
Signed-off-by: Alon Kenneth <11458012+akenneth@users.noreply.github.com>

## Description

Add missing support for `uint32` bindvars mapping.

## Related Issue(s)

* Fixes https://github.com/vitessio/vitess/issues/9756

## Checklist
- [ ] Should this PR be backported? **- akenneth: I don't know!**
- [x] Documentation was added or is not required - not required i think?

### DIFF
--- a/go/vt/vtgate/evalengine/eval_result.go
+++ b/go/vt/vtgate/evalengine/eval_result.go
@@ -733,7 +733,7 @@ func (er *EvalResult) setBindVar1(typ sqltypes.Type, value []byte, collation col
 	case sqltypes.Uint32:
 		ival, err := strconv.ParseUint(string(value), 10, 32)
 		if err != nil {
-			ival = 0
+			throwEvalError(err)
 		}
 		er.setUint64(ival)
 	case sqltypes.Uint64:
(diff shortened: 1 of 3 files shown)


go/vt/vtgate/evalengine/eval_result.go
```diff
@@ -730,6 +730,12 @@ func (er *EvalResult) setBindVar1(typ sqltypes.Type, value []byte, collation col
 		}
 		// TODO: type32
 		er.setInt64(ival)
+	case sqltypes.Uint32:
+		ival, err := strconv.ParseUint(string(value), 10, 32)
+		if err != nil {
+			ival = 0
```
vmg, line 736: You must return an error here just like the other cases.

vmg (changes requested):
Thanks for the contribution! Can you fix then error return?

go/vt/vtgate/evalengine/eval_result.go
akenneth: Thanks @vmg! I updated to `throwEvalError(err)` 

Please note that the other int cases are not throwing an error in the current state. For example:
https://github.com/vitessio/vitess/pull/9757/files#diff-8bd3f97165c3b882919fff81a641216b78e2c23ba08ca3d0e2d8dab3574bc72aL735-L737

Maybe it makes sense to change all of them to throw?

vmg: Huh, I see now that we have inconsistent behavior there. I'll fix it in a separate PR. Thanks for the heads up!

vmg (approved):

PR #9765, opened 2022-02-23
Filter candidates which cannot make forward progress after ERS

GuptaManan100:
## Description

After the changes in #9659, we started allowing multiple failures in ERS as long as we are able to guarantee safety.
This leads to an additional concern of a tablet being promoted which cannot make forward progress after being elected.

Consider the following case - 
We have 5 tablets (A, B, C, D and E) of which A, B and C are eligible to be primary tablets with C being the current primary. Let the durability policies be configured such that A requires 3 semi-sync acks to accept transactions  and both B and C require only 2. 

Consider the case when C and D both become unreachable. In this case we have more than 1 failures and pre #9659 we would have failed ERS. Post #9659, we will try to recover since having A, B and E reachable is enough to know that C won't be able to accept any writes, so it is safe to proceed.

In this case, if we promote tablet A, we will have promoted a tablet which cannot make forward progress (there are only 2 semi-sync acking tablets B and E available) and will hang on all the writes. Instead we must promote B.

This PR fixes this issue by creating a function analogous to Revocation, called Establish which checks whether a tablet can make forward progress given the list of reachable tablets.

This function is then used in ERS to filter out all the tablets which should not be promoted before we select a new tablet.
As part of this PR, other constraints checks have also been moved into a common function to filter primary candidates.  

## Related Issue(s)

- #8975 

## Checklist


go/vt/vtctl/reparentutil/replication.go
```diff
@@ -163,14 +164,16 @@ func StopReplicationAndBuildStatusMaps(
 	ignoredTablets sets.String,
 	tabletToWaitFor *topodatapb.TabletAlias,
 	logger logutil.Logger,
-) (map[string]*replicationdatapb.StopReplicationStatus, map[string]*replicationdatapb.PrimaryStatus, error) {
+) (map[string]*replicationdatapb.StopReplicationStatus, map[string]*replicationdatapb.PrimaryStatus, []*topodatapb.Tablet, error) {
```
deepthi, line 167: This return list is getting unwieldy. May be worth defining a struct for it.

go/vt/vtctl/reparentutil/durability_funcs.go
```diff
@@ -0,0 +1,90 @@
...
+		}
+		if !RevokeForTablet(tablet, tabletsReached, allTablets) {
+			return false
+		}
+	}
+	return true
+}
+
+// EstablishForTablet checks whether we have reached enough tablets to say that the given primary eligible tablet will be able to accept new writes
+func EstablishForTablet(primaryEligible *topodatapb.Tablet, tabletsReached []*topodatapb.Tablet) bool {
```
deepthi, line 74: Some of the feedback from the base PR is also valid here..
Whether this should be exported (may be not), prefix the name with `Is` etc.

go/vt/vtctl/reparentutil/durability_funcs.go
```diff
@@ -0,0 +1,90 @@
...
+
+	// semiSyncAckersReached is the list of reachable tablets capable of sending semi sync Acks for the given primaryEligible tablet
+	semiSyncAckersReached := SemiSyncAckersForPrimary(primaryEligible, tabletsReached)
+
+	// numOfSemiSyncAcksRequired is the number of semi sync Acks that the primaryEligible tablet requires
+	numOfSemiSyncAcksRequired := SemiSyncAckers(primaryEligible)
+
+	// if we have reached enough semi-sync Acking tablets such that the primaryEligible can accept a write
+	// we can safely promote this tablet
+	return len(semiSyncAckersReached) >= numOfSemiSyncAcksRequired
```
deepthi, line 89: It's nice how simple this logic is 👍

go/vt/vtctl/reparentutil/emergency_reparenter.go
```diff
@@ -211,13 +210,25 @@ func (erp *EmergencyReparenter) reparentShardLocked(ctx context.Context, ev *eve
 	// In case the user has specified a tablet specifically, then it is selected, as long as it is the most advanced.
 	// Here we also check for split brain scenarios and check that the selected replica must be more advanced than all the other valid candidates.
 	// We fail in case there is a split brain detected.
+	// The validCandidateTablets list is sorted by the replication positions with ties broken by promotion rules
```
ajm188, line 213: nit: add a period here please

go/vt/vtctl/reparentutil/emergency_reparenter.go
```diff
@@ -211,13 +210,25 @@ func (erp *EmergencyReparenter) reparentShardLocked(ctx context.Context, ev *eve
...
 	intermediateSource, validCandidateTablets, err = erp.findMostAdvanced(validCandidates, tabletMap, opts)
 	if err != nil {
 		return err
 	}
 	erp.logger.Infof("intermediate source selected - %v", intermediateSource.Alias)
 
+	// After finding the intermediate source, we want to filter the valid candidate list by the following criterion -
```
ajm188, line 220: nit: you have multiple conditions, therefore "criterion" should be "criteria"

go/vt/vtctl/reparentutil/emergency_reparenter.go
```diff
@@ -211,13 +210,25 @@ func (erp *EmergencyReparenter) reparentShardLocked(ctx context.Context, ev *eve
...
+	// 2. Remove the tablets with the Must_not promote rule
+	// 3. Remove cross-cell tablets if PreventCrossCellPromotion is specified
+	// Our final primary candidate MUST belong to this list of valid candidates
+	validCandidateTablets, err = erp.filterValidCandidates(validCandidateTablets, statusMaps.tabletsReachable, prevPrimary, opts)
+	if err != nil {
+		return err
+	}
+
 	// check whether the intermediate source candidate selected is ideal or if it can be improved later
+	// If the intermediateSource is ideal, then we can be certain that it is part of the valid candidates list
```
ajm188, line 231: nit: punctuation/capitalization here as well please

go/vt/vtctl/reparentutil/emergency_reparenter.go
```diff
@@ -417,17 +418,28 @@ func (erp *EmergencyReparenter) promoteIntermediateSource(
...
 		return nil, err
 	}
 
 	// also include the current tablet for being considered as part of valid candidates for ERS promotion
-	validCandidatesForImprovement = append(validCandidatesForImprovement, source)
+	reachableTablets = append(reachableTablets, source)
+
+	// The only valid candidates for improvement are the ones which are reachable and part of the valid candidate list
+	// Here we need to be careful not to mess up the ordering of tablets in validCandidateTablets since the list is sorted by the
+	// replication positions
```
ajm188, line 436: also here (basically anywhere you have large block comments, it makes it easier to parse to have full, punctuated sentences vs trying to figure out if something is one long run-on sentence)

go/vt/vtctl/reparentutil/emergency_reparenter.go
```diff
@@ -733,3 +738,37 @@ func (erp *EmergencyReparenter) promoteNewPrimary(
 	}
 	return nil
 }
+
+// filterValidCandidates filters valid tablets, keeping only the ones which can successfully be promoted without any constraint failures and can make forward progress on being promoted
+func (erp *EmergencyReparenter) filterValidCandidates(validTablets []*topodatapb.Tablet, tabletsReachable []*topodatapb.Tablet, prevPrimary *topodatapb.Tablet, opts EmergencyReparentOptions) ([]*topodatapb.Tablet, error) {
```
ajm188, line 743: 👍

go/vt/vtctl/reparentutil/durability_funcs_test.go
```diff
@@ -313,3 +313,70 @@ func Test_haveRevoked(t *testing.T) {
 		})
 	}
 }
+
+func Test_canEstablishForTablet(t *testing.T) {
```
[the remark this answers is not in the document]
GuptaManan100: Unfortunately no, all the tests that set the durability policy cannot run in parallel since the durabilityPoilicy is stored as a global variable, so they conflict with each other

go/vt/vtctl/reparentutil/replication.go
```diff
@@ -151,10 +151,19 @@ func ReplicaWasRunning(stopStatus *replicationdatapb.StopReplicationStatus) (boo
 	return stopStatus.Before.IoThreadRunning || stopStatus.Before.SqlThreadRunning, nil
 }
 
-// StopReplicationAndBuildStatusMaps stops replication on all replicas, then
+// statusMapsAndReachableTablets stores the status maps and the tablets that were reachable
+// when trying to stopReplicationAndBuildStatusMaps.
+type statusMapsAndReachableTablets struct {
```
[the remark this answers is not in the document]
GuptaManan100: This is great! Much better than what I came up with. I wasn't very convinced of the name myself and named this so for the lack of better alternates

go/vt/vtctl/reparentutil/durability_funcs_test.go
```diff
@@ -313,3 +313,70 @@ func Test_haveRevoked(t *testing.T) {
...
+	tests := []struct {
+		name             string
+		durabilityPolicy string
+		primaryEligible  *topodatapb.Tablet
+		tabletsReached   []*topodatapb.Tablet
+		canEstablish     bool
+	}{
+		{
+			name:             "'none' durability policy - primary not reached",
+			durabilityPolicy: "none",
```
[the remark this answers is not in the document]
GuptaManan100: Fixed

go/vt/vtctl/reparentutil/replication.go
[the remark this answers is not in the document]
GuptaManan100: Fixed

go/vt/vtctl/reparentutil/emergency_reparenter.go
```diff
@@ -139,11 +139,10 @@ func (erp *EmergencyReparenter) reparentShardLocked(ctx context.Context, ev *eve
 	ersCounter.Add(1)
 
 	var (
+		statusMaps                 *statusMapsAndReachableTablets
```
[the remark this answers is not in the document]
GuptaManan100: Fixed

go/vt/vtctl/reparentutil/emergency_reparenter.go
GuptaManan100: no hablo inglés 😆

GuptaManan100: Fixed 👍

go/vt/vtctl/reparentutil/emergency_reparenter.go
GuptaManan100: fixed

go/vt/vtctl/reparentutil/emergency_reparenter.go
GuptaManan100: Fixed

PR #9767, opened 2022-02-23
Rewriting `JoinCondition` produces ambiguous error in MySQL

frouioui:
## Description

For some reasons, MySQL accepts the following query:
```sql
select name from properties join supervisions using (propertyId) join managers using (managerId) where propertyId = 1 AND managerId = 1;
```

But fails with an ambiguous column error with the same query as above, but with the first `using` rewritten to an `on` condition:
```sql
select name from properties join supervisions on properties.propertyId = supervisions.propertyId join managers using (managerId) where propertyId = 1 AND managerId = 1;
```

Our planner rewrites `JoinCondition` that uses `USING` (as long as there is no `*` in the select list), into an `ON` `JoinCondition`.

https://github.com/vitessio/vitess/blob/8f1079efa84d6e7a3f06bf2594af29a7296e7fda/go/vt/sqlparser/ast_rewriting.go#L367

https://github.com/vitessio/vitess/blob/8f1079efa84d6e7a3f06bf2594af29a7296e7fda/go/vt/sqlparser/ast_rewriting.go#L446-L447

We do this rewriting because we don't have support (in both v3 and Gen4) for `USING` condition in sharded queries. When we execute the following:

```sql
select name from properties join supervisions using (propertyId) join managers using (managerId) where propertyId = 1 AND managerId = 1;
```

We rewrite the first `USING` condition to an `ON`, send it to MySQL and get an ambiguous column error.

## Solution

This pull request removes the rewriting of `JoinCondition` because it leads to inconsistent behavior with MySQL. Join with `USING` can be planned when sent to an unsharded keyspace, otherwise the planner returns an unsupported error.

## Checklist


systay (approved):

systay:
We should write something about this in the release log. Queries that used to work will stop working.

frouioui:
Good idea, I'll open a follow up PR to update the 14_0_0_summary

PR #9771, opened 2022-02-24
Externalize Lookup VIndexes properly when not stopping after copy

mattlord:
## Description

## Related Issue(s)
- Follow-up to: https://github.com/vitessio/vitess/pull/8572

## Checklist
- [x] Should this PR be backported? YES, at least to release-13.0
- [x] Tests were updated

### DIFF
--- a/go/vt/wrangler/materializer.go
+++ b/go/vt/wrangler/materializer.go
@@ -26,6 +26,7 @@ import (
 	"sync"
 	"text/template"
 
+	"google.golang.org/protobuf/encoding/prototext"
 	"google.golang.org/protobuf/proto"
 
 	"vitess.io/vitess/go/json2"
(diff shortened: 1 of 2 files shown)


rohit-nayak-ps (approved):
lgtm

mattlord:
If you do encounter this in v12.0 or v13.0 (this was only fixed in main/v14.0), the workaround is:
1. Use `VExec` to stop the workflow and set the message to `Stopped after copy`:
    ```
    vtctl VExec <keyspace>.<workflow> "update _vt.vreplication set state='Stopped', message='Stopped after copy' where workflow='<workflow>'"
    ```
2. Run `vtctl ExternalizeVindex <keyspace>.<vindex>` again

go/vt/wrangler/materializer.go
```diff
@@ -767,8 +768,17 @@ func (wr *Wrangler) ExternalizeVindex(ctx context.Context, qualifiedVindexName s
 			}
 			state := row[1].ToString()
 			message := row[2].ToString()
-			if sourceVindex.Owner == "" {
-				// If there's no owner, all streams need to be running.
+			var bls binlogdatapb.BinlogSource
+			sourceBytes, err := row[3].ToBytes()
```
sougou, line 772: sourceBytes looks unused.

sougou (commented):
I think this makes sense. Let's have @rohit-nayak-ps also eyeball this.

go/vt/wrangler/materializer.go
mattlord: See line 776 🙂 We merged it already, but I appreciate you looking! ❤️

PR #9772, opened 2022-02-24
Online DDL declarative migrations now use schemadiff, removing tengo

shlomi-noach:
## Description

Now that #9719 is merged, we switch `-declarative` migrations from using `tengo` into using `schemadiff`. In this change:

- `executor.go` now uses `schemadiff`'s declarative diff based on `SHOW CREATE TABLE` output from original and ghost tables
- There is one less table to create in the interim of the process
- We utilize existing Vitess DB connections, and remove use of `sqlx`
- No need to create a temporary account
- No longer using `tengo` library
- Removed some boilerplate parsing code
- Removed many package dependencies (see `go.mod`)

Tests-wise, the PR remains covered by `onlineddl_declarative`.

## Related Issue(s)

- #9719 
- #6926 

## Checklist


vmg (commented):
This is, again, great just from the sheer amount of dependencies it removes. I don't have a good enough understanding of the subsystem to approve the changes so I'm basically giving my thumbs up to the `go.mod` diff. :)

shlomi-noach:
call for review :pray:

harshit-gangal (approved):

shlomi-noach:

PR #9773, opened 2022-02-24
evalengine: Value cleanups

vmg:
## Description

Just some refactoring/cleanup around `sqltypes.Value` and the evaluation engine. We had a lot of duplicated code between the two packages which I've now merged.

## Checklist


vmg:
The refactoring for `Value` parsing has fixed some end2end tests that were previously returning `VARBINARY` for values that clearly were `VARCHAR` texts. Good!

systay (approved):

PR #9774, opened 2022-02-24
Abs function add

tianshishuo:
## Description

evalengine: Support built-in MySQL functions--abs()

## Related Issue(s)

Contributes to [evalengine: Support all built-in MySQL functions](https://github.com/vitessio/vitess/issues/9647)

## Checklist

### DIFF
--- a/go/vt/vtgate/evalengine/abs.go
+++ b/go/vt/vtgate/evalengine/abs.go
@@ -1,0 +1,52 @@
+package evalengine
+
+import (
+	"math"
+	"vitess.io/vitess/go/sqltypes"
+	vtrpcpb "vitess.io/vitess/go/vt/proto/vtrpc"
+	"vitess.io/vitess/go/vt/vterrors"
+)
+
+type buildinABS struct{}
+
+func (buildinABS) call(env *ExpressionEnv, args []EvalResult, result *EvalResult) {
+	toabs := &args[0]
+	if toabs.null() {
+		result.setNull()
+		return
+	}
+
+	switch toabs.typeof() {
+	case sqltypes.Int64:
+		res := toabs.int64()
+		if res >= 0 {
+			result.setInt64(res)
+		} else {
+			if res == math.MinInt64 {
+				//Same to mysql, return out of range error
+				throwEvalError(vterrors.Errorf(vtrpcpb.Code_OUT_OF_RANGE, "BIGINT value is out of range in: %s", toabs.String()))
+			} else {
+				result.setInt64(-res)
+			}
+		}
+	case sqltypes.Decimal:
+		res := toabs.decimal()
+		res.num.Abs(&res.num)
+		result.setDecimal(res)
+	case sqltypes.Float64:
+		res := toabs.float64()
+		res = math.Abs(res)
+		result.setFloat(res)
+	default:
+		throwEvalError(vterrors.Errorf(vtrpcpb.Code_UNIMPLEMENTED, "Unsupported ABS argument: %s", toabs.String()))
+	}
+}
+
+func (buildinABS) typeof(env *ExpressionEnv, args []Expr) (sqltypes.Type, flag) {
+	if len(args) < 1 {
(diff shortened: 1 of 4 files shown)


go/vt/vtgate/evalengine/abs.go
```diff
@@ -0,0 +1,52 @@
+package evalengine
+
+import (
+	"math"
+	"vitess.io/vitess/go/sqltypes"
+	vtrpcpb "vitess.io/vitess/go/vt/proto/vtrpc"
+	"vitess.io/vitess/go/vt/vterrors"
+)
+
+type buildinABS struct{}
```
vmg, line 10: Built-in is spelled with a `t` :sweat_smile: -- also, the name of the function doesn't need to be in all caps. `builtinAbs` would be a better name.

go/vt/vtgate/evalengine/abs_test.go
```diff
@@ -0,0 +1,76 @@
+package evalengine
+
+import (
+	"math"
+	"testing"
+)
+
+func Test_buildinABS_call(t *testing.T) {
```
vmg, line 8: Thanks for writing tests for this feature! However, we're trying to test the `evalengine` more reliably against MySQL itself: we write integration tests that verify the behavior of the function against a live MySQL instance. You can see an example here: https://github.com/vitessio/vitess/blob/main/go/vt/vtgate/evalengine/integration/comparison_test.go#L524-L551

go/vt/vtgate/evalengine/abs.go
```diff
@@ -0,0 +1,52 @@
+package evalengine
```
vmg, line 1: I don't think `abs` needs to have its own file. You can put this implementation if `func.go`.

vmg (changes requested):
Thanks for your first contribution to Vitess! I've left some feedback on things that need to be fixed -- particularly related to testing.

go/vt/vtgate/evalengine/abs.go
tianshishuo: Okay! I will revise it.

go/vt/vtgate/evalengine/abs.go
tianshishuo: I've solved it

go/vt/vtgate/evalengine/abs.go
tianshishuo: This question has been revised and submitted. Do you have any questions？

go/vt/vtgate/evalengine/abs.go
```diff
@@ -0,0 +1,52 @@
...
+			if res == math.MinInt64 {
+				//Same to mysql, return out of range error
+				throwEvalError(vterrors.Errorf(vtrpcpb.Code_OUT_OF_RANGE, "BIGINT value is out of range in: %s", toabs.String()))
+			} else {
+				result.setInt64(-res)
+			}
+		}
+	case sqltypes.Decimal:
+		res := toabs.decimal()
+		res.num.Abs(&res.num)
```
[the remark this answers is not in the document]
tianshishuo: This question has been revised and submitted. Do you have any questions

go/vt/vtgate/evalengine/abs_test.go
tianshishuo: Do I need to submit more test cases？

go/vt/vtgate/evalengine/internal/decimal/decimal.go
```diff
@@ -277,6 +277,10 @@ func (d Decimal) abs() Decimal {
 	}
 }
 
+func (d Decimal) Abs() Decimal {
```
vmg, line 280: You should make the `abs` function public instead of wrapping it.

vmg:
There are, for some reason, 98 commits in this PR now. It looks like you screwed up a merge or a rebase. You'll need to fix this into (ideally) a single commit so we can review. Thanks!

PR #9606, opened 2022-02-01
Support views in Online DDL

shlomi-noach:
## Description

This PR Adds support for `CREATE/ALTER/DROP VIEW` statements in Online DDL, thus:
- support in `vtctl ApplySchema`
- support via VTGate
- supported with all online strategies
- support `-declarative`
- support `OR REPLACE` nuance
- support `REVERT`

Things that need/needed to change:

- parser: support comments in `CREATE/ALTER/DROP VIEW` statements (comments are passed on to query service, to indicate the online ddl params)
- ast/format: analyze and rebuilt query with comments
- OnlineDDL: agree to accept View DDLs
- online ddl/parser: able to replace view name in-place
- tengo: 
  - https://github.com/planetscale/tengo/pull/15
  - https://github.com/planetscale/tengo/pull/18
  - https://github.com/planetscale/tengo/pull/19
- executor: obviously this is where the main logic changes are

endtoend tests to be added
documentation to be added

## Related Issue(s)

- Closes https://github.com/vitessio/vitess/issues/9573
- tracking: #6926 

## Checklist
- [ ] Should this PR be backported?
- [ ] Tests were added or are not required
- [ ] Documentation was added or is not required

cc @iheanyi

PR #9776, opened 2022-02-24
Table lifecycle: support views

shlomi-noach:
## Description

This PR now supports view in table-lifecycle flows. Specifically:

- We should not purge data from views. ie if a view is renamed to `_vt_PURGE_*****` then we must immediately consider it as purged. Purging data from views is either impossible (if the view is not-updatable) or wrong/damaging (because rows are actually purged from an underlying table).
- Views never need to spend time in `EVAC` state. There is no data in views, so there's no need to wait for data to be evacuated from the buffer pool.

## Related Issue(s)

- https://github.com/vitessio/vitess/pull/9606

## Checklist


rohit-nayak-ps (approved):
lgtm

PR #9777, opened 2022-02-24
[vtctldserver] Add locking checks to Delete{Keyspace,Shard}

ajm188:
## Description

This PR adds `Lock{Keyspace,Shard}` calls to our `Delete{Keyspace,Shard}` rpcs, respectively. Not having this can result in shards that appear orphaned to other clients that are walking the topo listings.

The diff is a little bigger, because I also tidied up the the other CLI flags/help texts in the keyspaces and shards files while I was in there.

## Checklist
- [x] Should this PR be backported? **NO**


go/vt/vtctl/grpcvtctldserver/topo.go
```diff
@@ -32,14 +32,43 @@ import (
...
 	span.Annotate("even_if_serving", evenIfServing)
+	span.Annotate("force", force)
+
+	lctx, unlock, lerr := ts.LockShard(ctx, keyspace, shard, "DeleteShard")
+	switch {
+	case lerr == nil:
+		// We locked the shard, all good
+		ctx = lctx
+	case !force:
+		return fmt.Errorf("failed to lock %s/%s: %w", keyspace, shard, lerr)
```
deepthi, line 51: we should probably add more text to the returned error. Something along the lines of
```
If you really want to delete this shard, re-run with --force
```

go/vt/vtctl/grpcvtctldserver/server_test.go
```diff
@@ -1985,6 +2058,78 @@ func TestDeleteShards(t *testing.T) {
 			},
 			shouldErr: false,
 		},
+		{
+			name: "shard is locked",
```
deepthi, line 2062: Nice tests!

Issue #6689, opened 2020-09-10
Tracking issue: safe, lazy, managed DROP TABLE


shlomi-noach:
`DROP TABLE` is a risky MySQL operation in production. There seem to be multiple components involved, the major being that if the table has pages in InnoDB's buffer pool (one or many), then the buffer pool is locked for the duration of the `DROP`. The duration of the `DROP` is also related with the time it takes the operating system to delete the `.ibd` file associated with the table (assuming `innodb_file_per_table`).

Noteworthy that the problem is in particular on the `primary` (`master`) node; Replicas are not affected as much.

Different companies solve `DROP TABLE` in different ways. An interesting discussion is found on `gh-ost`'s repo and on mysql bugs:

- https://github.com/github/gh-ost/issues/307
- https://bugs.mysql.com/bug.php?id=91977

Solutions differ in implementation, but all suggest _waiting_ for some time before actually dropping the table. That alone requires management around `DROP TABLE` operations. As explained below, _waiting_ enables reverting the operation.

Vitess should automate table drops and make the problem transparent to the user as much as possible. Breakdown of the suggested solution follows.

## Illustrating the DROP steps

We can make the `DROP` management stateful or stateless. I opt for stateless: no meta tables to describe the progress of the `DROP`. The state should be inferred from the tables themselves. Specifically, we will encode hints in the table names.

We wish to manage `DROP` requests. Most managed `DROP` requests will _wait_ before destroying data. If the user issued a `DROP TABLE` only to realize the app still expects the table to exist, then we make it possible to revert the operation.

This is done by first issuing a `RENAME TABLE my_table TO something_else`. To the app, it seems like the table is gone; but the user may easily restore it by running the revert query: `RENAME TABLE something_else TO my_table`.

That `something_else` name can be e.g. `_vt_HOLD_2201058f_f266_11ea_bab4_0242c0a8b007_20200910113042`.

At some point we decide that we can destroy the data. The "hold" period can either be determined by vitess or explicitly by the user. e.g. On a successful schema migration completion, Vitess can choose to purge the "old" table right away.
At that stage we rename the table to e.g. `_vt_PURGE_63b5db0c_f25c_11ea_bab4_0242c0a8b007_20200911070228`.
A table by that name is eligible to have its data purged.

In my experience (see `gh-ost` issue above), a safe method to purge data is to slowly remove rows, until the table is empty. Note:

[...]

PR #9778, opened 2022-02-24
Table lifecycle: support fast DROP TABLE introduced in MySQL 8.0.23

shlomi-noach:
## Description

Up till MySQL `8.0.23`, a `DROP TABLE` operation was a locking and blocking operation. It would involve locking and scanning the buffer pool while table was dropped and file (assuming innodb-file-per-table) removed.

As of `8.0.23`, this is no longer the case. a `DROP TABLE` will not place locks on the innodb buffer pool, and will not block queries that operate on other tables.

In this PR, if we identify the server to be MySQL8 >= `8.0.23`, we automatically skip `PURGE` and `EVAC` states. Tables can be in `HOLD` state if desired, but then the next state is `DROP`. This means tables are dropped, and disk space reclaimed, sooner, potentially days sooner.

## Related Issue(s)

- Tracking: #6689

## Checklist


rohit-nayak-ps (approved):
lgtm

PR #9780, opened 2022-02-24
[RFC] Add a max-in-use gauge for connection pools

djudd:
## Description
Proposal to add a gauge which reports the maximum value that InUse has taken for a connection pool over the gauge interval.

The current InUse gauge can hide spiky load, which is something we've seen in practice at Stripe.

It might make sense to also add an AvailableMin gauge, or even just add that since it requires less context to interpret? I'm also pretty unsure if this is the right implementation; it feels like something simpler ought to be possible, but maybe that's just not true with atomics. Looking for feedback!

## Related Issue(s)
n/a

## Checklist

## Deployment Notes

n/a

### DIFF
--- a/go/pools/resource_pool.go
+++ b/go/pools/resource_pool.go
@@ -73,6 +73,7 @@ type ResourcePool struct {
 	waitTime   sync2.AtomicDuration
 	idleClosed sync2.AtomicInt64
 	exhausted  sync2.AtomicInt64
+	maxInUse   sync2.AtomicInt64
 
 	capacity    sync2.AtomicInt64
 	idleTimeout sync2.AtomicDuration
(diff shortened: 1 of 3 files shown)


aquarapid:
Does this really add more signal than the `*_pool_exhausted`, `*_pool_wait_count` and `*_pool_wait_time` metrics?  Pools getting full (and there being waits) is an expected occurrence in busy systems.

If the point is to determine the highwatermark for the number of connections used during load peaks, while purposefully keeping the pool "too large" (so it never fills up) for the purposes of estimating what a "good" pool sizing would be, I guess that has some validity.  But, in general, the recommended way is to size the pools in accordance to the capability of the underlying MySQLd instance, so that connections queue on the Vitess side, not in MySQLd. So you might be coming at this from the wrong angle.

@sougou for opinion(s).

djudd:
Yeah, the idea is to be able to tell how close we are to capacity without actually running into exhaustion.

While we're still tuning our connection pools, I believe our current objective is to queue neither in Vitess nor in MySQL, but to have a pretty small cap on waiters and instead reject queries and backpressure the client (which is expected to back off and retry). If that's against your standard recommendation, maybe it's worth discussing in our shared Slack? I don't have full context on why we picked that objective (which does have the implication that we don't want to see routine pool exhaustion in latency-sensitive services in production, since the backoff will be expensive, and that's part of the context for this PR).

sougou:
I remember trying to track this at YouTube, and was not able to draw a good interpretation for this value. We eventually landed on the more palpable tail latency metric. Essentially, a spiky `InUse` will show up in your 99 and 95 percentiles. This is also a more meaningful number to tune on because it gets tied to other guarantees of the system.

PR #9781, opened 2022-02-24
Backport: Externalize Lookup VIndexes properly when not stopping after copy

mattlord:
## Description

## Related Issue(s)
- Backports: https://github.com/vitessio/vitess/pull/9771

## Checklist
- [ ] Should this PR be backported? Not directly, will backport the parent to 12.0 too
- [x] Tests were updated

### DIFF
--- a/go/vt/wrangler/materializer.go
+++ b/go/vt/wrangler/materializer.go
@@ -26,6 +26,7 @@ import (
 	"sync"
 	"text/template"
 
+	"google.golang.org/protobuf/encoding/prototext"
 	"google.golang.org/protobuf/proto"
 
 	"vitess.io/vitess/go/json2"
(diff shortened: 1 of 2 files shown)


deepthi:
What is the justification for back porting to 13 and 12?

mattlord:
> What is the justification for back porting to 13 and 12?

The [-continue_after_copy_with_owner=true](https://vitess.io/docs/13.0/user-guides/configuration-advanced/createlookupvindex/) flag was [added in v12.0](http://github.com/vitessio/vitess/pull/8572/commits/7e7722ac7b51aa62c42b81a7e679e8acb1c15384)) with a specific use case in mind:
```
-continue_after_copy_with_owner: By default, when an owner is provided, the VReplication streams will stop after the backfill completes. Set this flag if you don’t want this to happen. This is useful if, for example, the owner table is being migrated from an unsharded keyspace to a sharded keyspace using MoveTables.
```

A PlanetScale Enterprise Vitess support customer used the flag for this precise use case. But they were unable to externalize this VIndex because of the noted behavior. They got this error:
> vtctl ExternalizeVindex keyspace.workflowid
E0223 23:31:16.878922   70321 vtctl.go:116] action failed: ExternalizeVindex Code: UNKNOWN
stream 73 for keyspace.-20 is not in Stopped after copy state: Stopped,
stream 73 for keyspace.20-40 is not in Stopped after copy state: Stopped,
stream 73 for ...

In order to get past this you have to:
1. Use `vtctl Workflow keyspace.workflowid stop`
2. Use `vtctl VExec` to change the message for the workflow from `Stopped` `Stopped after copy state`
3. Run `vtctl ExternalizeVindex` again

That's not ideal, and not something we generally want users to do (directly modify the `_vt.vreplication` workflow metadata).

If you don't feel that this warrants a backport though, please please feel free to close it.

mattlord:
Closing this as it does not meet [the defined criteria for backports](https://github.com/vitessio/enhancements/blob/main/veps/vep-5.md#support-lifecycle). I had a different (incorrect) understanding of that.

If you do encounter this in v13.0, the workaround is:
1. Use `vtctl Workflow keyspace.workflowid stop`
2. Use `vtctl VExec keyspace.workflowid  -sql ` to change the message for the workflow from `Stopped` `Stopped after copy state`
3. Run `vtctl ExternalizeVindex` again

PR #9782, opened 2022-02-24
Backport: Externalize Lookup VIndexes properly when not stopping after copy

mattlord:
## Description

## Related Issue(s)
- Backports: https://github.com/vitessio/vitess/pull/9771

## Checklist
- [x] Should this PR be backported? NO further
- [x] Tests were updated

### DIFF
--- a/go/vt/wrangler/materializer.go
+++ b/go/vt/wrangler/materializer.go
@@ -25,6 +25,7 @@ import (
 	"sync"
 	"text/template"
 
+	"google.golang.org/protobuf/encoding/prototext"
 	"google.golang.org/protobuf/proto"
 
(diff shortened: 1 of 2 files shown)


mattlord:
Closing this as it does not meet [the defined criteria for backports](https://github.com/vitessio/enhancements/blob/main/veps/vep-5.md#support-lifecycle). I had a different (incorrect) understanding of that.

If you do encounter this in v12.0, the workaround is:
1. Use `vtctl Workflow keyspace.workflowid stop`
2. Use `vtctl VExec` to change the message for the workflow from `Stopped` `Stopped after copy state`
3. Run `vtctl ExternalizeVindex` again

Issue #9754, opened 2022-02-22
Bug Report: Errors on CREATE TABLE statements via nodejs/knex 


jseparovic:
### Overview of the Issue

While trying to create a number of tables via nodejs/knex, vitess has issues after some number of tables.

I see `vttablet: rpc error: code = Canceled`
followed by `vttablet: rpc error: code = Code(17)`

Here's a full example output trying to create 30 tables.
https://github.com/jseparovic/knex-test/blob/master/OUTPUT.md 

### Reproduction Steps

I've created a test repo at https://github.com/jseparovic/knex-test

Steps to reproduce are here:
https://github.com/jseparovic/knex-test/blob/master/README.md

I'm using the Kubernetes operator and the following vitess definition:
https://github.com/jseparovic/knex-test/blob/master/vitess_initial_cluster.yaml

I've created the following service to create tables via knex:
https://github.com/jseparovic/knex-test/blob/master/knex-test.yaml

It simply runs in a loop and creates tables from DDL via knex one at a time:
https://github.com/jseparovic/knex-test/blob/master/src/endpoint/KnexEndpoint.ts#L101-L103

Here's a sample of the DDL:
https://github.com/jseparovic/knex-test/blob/master/src/model/Table1.ts#L12-L22

Here's a full example output trying to create 30 tables.
https://github.com/jseparovic/knex-test/blob/master/OUTPUT.md

### Binary Version

```shell
[root@master1 knex-test]# vtgate --version
Version: 12.0.3 (Git revision cbc6ab6 branch 'HEAD') built on Mon Jan 10 16:26:09 UTC 2022 by runner@fv-az219-507 using go1.17.5 linux/amd64
```

### Operating System and Environment details

```shell
[root@master1 knex-test]# kubeadm version
kubeadm version: &version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.9", GitCommit:"b631974d68ac5045e076c86a5c66fba6f128dc72", GitTreeState:"clean", BuildDate:"2022-01-19T17:50:04Z", GoVersion:"go1.16.12", Compiler:"gc", Platform:"linux/amd64"}

[root@master1 knex-test]# cat /etc/os-release
NAME="AlmaLinux"
VERSION="8.5 (Arctic Sphynx)"
ID="almalinux"
ID_LIKE="rhel centos fedora"
VERSION_ID="8.5"
PLATFORM_ID="platform:el8"
[...]

PR #9784, opened 2022-02-25
Fix upgrade-downgrade CI workflows running even when Skip label specified

GuptaManan100:
## Description

Before #9745 we used to build a matrix from the latest releases. So, if the step to get the latest releases was skipped, our matrix used to be empty and no workflow for the upgrade downgrade steps used to run.

In #9745, we moved from a matrix having a single element to a string storing the releaes/tag to run the test against. So, when the step to get latest release was skipped, the output was still a string albeit an empty one. This triggered the next workflow to run with an empty previous release/tag like in https://github.com/vitessio/vitess/runs/5310337523?check_suite_focus=true.

This PR fixes this issue, so that the job to run the upgrade-downgrade test only works when the previous job of getting the release to test against succeeds.

## Checklist
- [x] Should this PR be backported? *Yes* along with #9754 

### DIFF
--- a/.github/workflows/upgrade_downgrade_test_backups_e2e.yml
+++ b/.github/workflows/upgrade_downgrade_test_backups_e2e.yml
@@ -48,7 +48,7 @@ jobs:
 
   upgrade_downgrade_test_e2e:
     timeout-minutes: 60
-    if: always() && (github.event_name != 'pull_request' || needs.get_upgrade_downgrade_label.outputs.hasLabel != 'false')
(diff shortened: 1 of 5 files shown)


frouioui (approved):
This is correct, otherwise we could have just fixed the `if` condition to be `needs.get_upgrade_downgrade_label.outputs.hasLabel != 'true'` instead of `needs.get_upgrade_downgrade_label.outputs.hasLabel != 'false'`

Issue #9788, opened 2022-02-25
Bug Report: vttablet replica is considered healthy (serving) even when not connect to primary


vkovacik:
### Overview of the Issue

**Short description:**

Replica tablet is serving reads even when replication IO thread is disconnected and replica data is stale for more than `unhealthy_threshold`.

**Long description:**

We were seeing inconsistency between data returned from replica and data returned from primary tablet. After investigation I found that the replication on the replica is not connected to the primary but vtgate shows that the replica is actively participating in the reads (serving).

Affected shard:
```
las-0832024891 marketplace -80 replica 10.77.24.84:15000 10.77.24.84:3306 [] <null>
las-4121544931 marketplace -80 primary 10.77.18.26:15000 10.77.18.26:3306 [] 2022-02-25T20:27:16Z
```

Replication status on replica (las-0832024891):
Note that IO thread is not connected, SQL thread is running and reported replication lag is 0.
```
vitess@vitess-scratch-dev-db-vttablet-las-0832024891-50c8c7e3:/$ mysql -S /vt/socket/mysql.sock -uroot -e 'show slave status \G' | grep -e "Slave_IO_Running\|Slave_SQL_Running\|SQL_Delay"
             Slave_IO_Running: Connecting
            Slave_SQL_Running: Yes
                    SQL_Delay: 0
      Slave_SQL_Running_State: Slave has read all relay log; waiting for more updates
```

Serving state:
```
mysql> show vitess_tablets;
+------+-------------+-------+------------+-------------+----------------+--------------+----------------------+
| Cell | Keyspace    | Shard | TabletType | State       | Alias          | Hostname     | PrimaryTermStartTime |
+------+-------------+-------+------------+-------------+----------------+--------------+----------------------+
| las  | marketplace | -80   | PRIMARY    | SERVING     | las-4121544931 | 10.77.18.26  | 2022-02-25T20:27:16Z |
| las  | marketplace | -80   | REPLICA    | SERVING     | las-0832024891 | 10.77.24.84  |                      |
...
```

Screenshot from vtgate status:
![Screenshot from 2022-02-25 12-47-12](https://user-images.githubusercontent.com/66448697/155800679-236739de-3898-4dac-a032-275e1cdaca59.png)

Full replication status output:
```
vitess@vitess-scratch-dev-db-vttablet-las-0832024891-50c8c7e3:/$ mysql -S /vt/socket/mysql.sock -uroot -e 'show slave status \G'              
[...]

PR #9789, opened 2022-02-25
Consider repl NOT healthy with IO_Thread connecting and last attempt failed

mattlord:
## Description
[We considered](https://github.com/vitessio/vitess/blob/6c20177eaff5e739d6eeee5b69b6112891e8bba6/go/mysql/flavor.go#L316-L321) the `Slave_IO_Running` state of `Connecting` as equivalent to `Running` in the [replication status](https://dev.mysql.com/doc/refman/en/replication-administration-status.html) results that we get from MySQL. I'm assuming this was done to avoid flapping on low traffic systems due to the [`-slave_net_timeout`](https://dev.mysql.com/doc/refman/en/replication-options-replica.html#sysvar_slave_net_timeout) reconnects and avoid doing a tablet repair when replication was healthy (with this PR we can distinguish running from healthy states and take differing actions depending on what we want).

After https://github.com/vitessio/vitess/pull/9308 we properly estimate the replica lag when MySQL is telling us that it does not know, meaning it returns a `NULL` value for the `Seconds_Behind_Master` field. But in some cases — e.g. it appears to happen when attempting the *first* connection to the source — MySQL will report a `Seconds_Behind_Master` value of `0` (meaning fully caught up and no lag) even when it is not connected to its replication source and has _failed to reconnect_ — meaning that this is not a simple reconnect for any reason (e.g. [`-slave_net_timeout`](https://dev.mysql.com/doc/refman/en/replication-options-replica.html#sysvar_slave_net_timeout)), but a reconnect with one or more _failures/errors_. This PR handles this latter case within Vitess by continuing to treat `Connecting` as equivalent to `Running` via `ReplicationStatus.Healthy()` — to prevent the noted flapping and errant/unnecessary tablet repairs — **_unless_**  we had an IO error the last time we tried to reconnect to the replication source. 

> Note: I think that [this](https://github.com/vitessio/vitess/pull/9789/files#r820941187) was also what in effect caused the bug seen in https://github.com/vitessio/vitess/issues/9788 as the new replica tablet was considered healthy when in fact it had not been able to connect to its source (for the very first time)

## Related Issue(s)
- Fixes: https://github.com/vitessio/vitess/issues/9788
- Related to: https://github.com/vitessio/vitess/pull/9308

## Checklist
- [x] Tests are not required (I don't think)


shlomi-noach:
I have a vague memory of having this discussion elsewhere in the code. Also, I had a similar in `orchestrator`. IMO the correct way to handle this is to avoid using a `bool` for `Slave_IO_Running`. The problem with keeping the behavior as before this PR is clear. But this PR also creates a confusion: it will say the IO thread is not running when it could be in the interim of completing a connection. So that's also confusing. Example for a risk to this approach: you check replication status, you get `false` for IO thread running; you assume replication is stopped; you run some logic that relies on this assumption, but unknown to you, the IO thread finally succeeds completing a connection and all of a sudden replication is running.

So IMO the state of IO thread should be an enum like `stopped`, `connecting`, `running`.

mattlord:
> I have a vague memory of having this discussion elsewhere in the code. Also, I had a similar in `orchestrator`. IMO the correct way to handle this is to avoid using a `bool` for `Slave_IO_Running`. The problem with keeping the behavior as before this PR is clear. But this PR also creates a confusion: it will say the IO thread is not running when it could be in the interim of completing a connection. So that's also confusing. Example for a risk to this approach: you check replication status, you get `false` for IO thread running; you assume replication is stopped; you run some logic that relies on this assumption, but unknown to you, the IO thread finally succeeds completing a connection and all of a sudden replication is running.
> 
> So IMO the state of IO thread should be an enum like `stopped`, `connecting`, `running`.

I like this idea. It's more accurate (the SQL_Thread really does only have a binary state whereas the IO_Thread has 3 states) and it allows the caller to decide whether or not to consider connecting and running equivalent or perform different actions. I'll work on this now.

Thank you!

go/mysql/constants.go
```diff
@@ -677,3 +678,33 @@ func IsSchemaApplyError(err error) bool {
...
+)
+
+// ReplicationStatusToState converts a value you have for the IO thread(s) or SQL
+// thread(s) or Group Replication applier thread(s) from MySQL or intermediate
+// layers to a mysql.ReplicationState.
+// on,yes,true == ReplicationStateRunning
+// off,no,false == ReplicationStateStopped
+// connecting == ReplicationStateConnecting
+// anything else == ReplicationStateUnknown
+func ReplicationStatusToState(s interface{}) ReplicationState {
```
shlomi-noach, line 698: IMHO the function should accept a `string`; it doesn't make sense for it to accept any type. It should be the caller's duty to provide a `string`.

go/mysql/constants.go
```diff
@@ -677,3 +678,33 @@ func IsSchemaApplyError(err error) bool {
...
+	// Group Replication uses ON instead of Yes
+	if status == "yes" || status == "on" || status == "true" {
+		return ReplicationStateRunning
+	} else if status == "no" || status == "off" || status == "false" {
+		return ReplicationStateStopped
+	} else if status == "connecting" {
+		return ReplicationStateConnecting
+	} else {
+		return ReplicationStateUnknown
+	}
```
shlomi-noach, line 709: purely a styling matter, but I think `switch` is cleaner here:
```go
	switch strings.ToLower(s) {
	case "yes", "on", "true":
		return ReplicationStateRunning
	case "no", "off", "false":
		return ReplicationStateStopped
	case "connecting":
		return ReplicationStateConnecting
	default:
		return ReplicationStateUnknown
	}
```
feel free to ignore any styling comment of mine.

go/mysql/flavor_mysqlgr.go
```diff
@@ -141,32 +141,32 @@ func (mysqlGRFlavor) status(c *Conn) (ReplicationStatus, error) {
 		return res, nil
 	}
 
-	// Populate IOThreadRunning from replication_connection_status
+	// Populate IOState from replication_connection_status
 	query = fmt.Sprintf(`SELECT SERVICE_STATE
 		FROM performance_schema.replication_connection_status
 		WHERE CHANNEL_NAME='%s'`, chanel)
-	var ioThreadRunning bool
+	var connectionState ReplicationState
```
shlomi-noach, line 148: the default will be `ReplicationStateRunning`

go/mysql/constants.go
```diff
@@ -677,3 +678,33 @@ func IsSchemaApplyError(err error) bool {
...
 	return false
 }
+
+type ReplicationState int
+
+const (
+	ReplicationStateRunning ReplicationState = iota
+	ReplicationStateStopped
+	ReplicationStateConnecting
+	ReplicationStateUnknown
```
shlomi-noach, line 688: let's make `ReplicationStateUnknown` the first - hence default - value. See below

go/mysql/flavor_mysqlgr.go
mattlord: Good catch! 😄 Will change the ordering so that 0 is unknown as you noted. Thanks!

go/mysql/constants.go
mattlord: Sure, I don't have a strong preference either way. I started on this patch and then switched and don't remember why. 🙂 Was making too many changes while tired, I think. I'll move to this. Thanks!

go/mysql/constants.go
mattlord: I don't feel strongly about it. I thought that sending a boolean would be nicer for some existing tests and uses. I can remove it though if there's good reason to.

mattlord: I'll do this now just to simplify things. Simplicity is itself a good reason on second thought. Thanks!

mattlord:
@shlomi-noach I applied all of your great suggestions and the tests are now passing. If you have time for another quick review pass (not urgent), I'd most appreciate it. 

Thank you again!

shlomi-noach (approved):
:rocket:

go/vt/mysqlctl/replication.go
```diff
@@ -47,7 +47,7 @@ func WaitForReplicationStart(mysqld MysqlDaemon, replicaStartDeadline int) error
 			return err
 		}
 
-		if status.ReplicationRunning() {
+		if status.Running() {
```
mattlord, line 50: Note: this was a bug. It should be using `Running()` and NOT `Healthy()` because it's supposed to wait until we know that we can connect to the source.

GuptaManan100 (commented):
Everything looks good to me! I just have 1 question. Is there any way to cleanup the old fields in Proto. I know they cannot be removed immediately for upgrade-downgrade considerations, but can they be marked for deprecation or something like that?

shlomi-noach (approved):
LGTM. I don';t have full context into the problem, and am assuming that the existing tests cover the behavior exhibited by this change.

mattlord:
> Everything looks good to me! I just have 1 question. Is there any way to cleanup the old fields in Proto. I know they cannot be removed immediately for upgrade-downgrade considerations, but can they be marked for deprecation or something like that?

Good question! I was wondering that too, but I don't know the answer. @deepthi do we have a defined process or an example to follow regarding making backwards compatibility breaking protobuf changes?

It would be nice to e.g. remove the `io_thread_running` and `io_thread_connecting` booleans and replace them with an `io_state` int field.

deepthi:
> > Everything looks good to me! I just have 1 question. Is there any way to cleanup the old fields in Proto. I know they cannot be removed immediately for upgrade-downgrade considerations, but can they be marked for deprecation or something like that?
> 
> Good question! I was wondering that too, but I don't know the answer. @deepthi do we have a defined process or an example to follow regarding making backwards compatibility breaking protobuf changes?
> 
> It would be nice to e.g. remove the `io_thread_running` and `io_thread_connecting` booleans and replace them with an `io_state` int field.

The way to make such changes to proto definitions is to use `reserved`. you rename the old fields to `reserved` and add the new field. I realize that we have a couple of fields with `deprecated=true`, those slipped through reviews, we should never have done that.

mattlord:
> > > Everything looks good to me! I just have 1 question. Is there any way to cleanup the old fields in Proto. I know they cannot be removed immediately for upgrade-downgrade considerations, but can they be marked for deprecation or something like that?
> > 
> > 
> > Good question! I was wondering that too, but I don't know the answer. @deepthi do we have a defined process or an example to follow regarding making backwards compatibility breaking protobuf changes?
> > It would be nice to e.g. remove the `io_thread_running` and `io_thread_connecting` booleans and replace them with an `io_state` int field.
> 
> The way to make such changes to proto definitions is to use `reserved`. you replace the old field with something like `reserved 2` and add the new field. I realize that we have a couple of fields with `deprecated=true`, those slipped through reviews, we should never have done that.

OK, thanks! I'll investigate to see if there's a way I can safely effectively deprecate the `io_thread_running` and `sql_thread_connecting` booleans and effectively replace them with `io_state` and `sql_state` ints (and get rid of the `io_thread_connecting` boolean added in the PR). I was assuming that would be forbidden but perhaps I was wrong. If we can do that it would keep the proto to struct translation clean and simple.

mattlord:
I apologize, but I'm going to rebase via FF merge and force-push as GitHub Actions is somehow checking out an invalid snapshot and I'm getting CI errors for code that *doesn't exist* in my branch's `HEAD`. I'll share another comment with the diff covering the changes since the last reviews once the tests are passing again.

mattlord:
Ugh, still happening. You can see an example here with errors about code that doesn't exist in my brach: https://github.com/vitessio/vitess/runs/5482587430?check_suite_focus=true

mattlord:
Force pushing from a fixed local branch to this PRs remote didn't seem to do it either. 😢 Moving this to a new PR: https://github.com/vitessio/vitess/pull/9853

PR #9790, opened 2022-02-26
Add TEXT field comparison support to evalengine (for VDiff)

mattlord:
## Description
We switched to using evalengine for `VDiff` comparisons in https://github.com/vitessio/vitess/pull/9679.

What we didn't realize at the time is that evalengine did not support comparing `TEXT` columns. When doing a `VDiff` against a table that has them, you get the following error (you can see a failed test after the test changes [here](https://github.com/vitessio/vitess/runs/5342687212?check_suite_focus=true)):
```
VDiff Error: rpc error: code = Unknown desc = diff: types are not comparable: TEXT vs TEXT
```

There's no logical reason for this as a `TEXT` field for in memory processing is no different from a `VARCHAR`. The main differences are the on disk format which stores the values off page due to the large maximum size for the type. The same is true for `BLOB` and `VARBINARY` fields, but we already treat them the same way in the evalengine. This PR treats `TEXT` fields the same as `VARCHAR` fields for comparisons. 

## Related Issue(s)
- Fixes: ToDo
- Related to: https://github.com/vitessio/vitess/pull/9679

## Checklist
- [x] Tests were updated


vmg (approved):
This looks correct. I forgot to handle TEXT explicitly.

PR #9793, opened 2022-02-27
Built-in string functions implementation: ASCII, BIN, BIT_LENGTH

wanghaoyang1995:
## Description

Add MySQL built-in functions `ascii`, `bin`, `bit_length` to evalengine.

## Related Issue(s)

https://github.com/vitessio/vitess/issues/9647

## Checklist

## Deployment Notes

N/A


go/vt/vtgate/evalengine/integration/string_func_test.go
```diff
@@ -0,0 +1,102 @@
...
+
+	t.Run("BIN", func(t *testing.T) {
+		for i := 0; i < len(elems); i++ {
+			query := fmt.Sprintf("SELECT BIN(%s)", elems[i])
+			compareRemoteQuery(t, conn, query)
+		}
+	})
+}
+
+func TestBuiltinBitLength(t *testing.T) {
```
vmg, line 77: Good job with these integration tests. I think the thing you're missing are strings with Unicode characters: it's important to test the behavior of these functions with Unicode strings. What happens if you call `ASCII()` on an emoji? What's the `BIT_COUNT` of a string that contains emoji or uses a multi-byte encoding?

wanghaoyang1995:
@vmg Hello,vmg. I have fixed some problems as suggested. For some case not defined in MySQL docs, there are different outputs between MySQL 5 and MySQL 8. Which one should I refer to?

go/vt/vtgate/evalengine/string.go
```diff
@@ -0,0 +1,125 @@
...
+
+func (builtinAscii) call(_ *ExpressionEnv, args []EvalResult, result *EvalResult) {
+	toascii := &args[0]
+	if toascii.isNull() {
+		result.setNull()
+		return
+	}
+
+	toascii.makeBinary()
+	if len(toascii.bytes()) > 0 {
```
vmg, line 43: Let's not call `bytes()` twice, we can store the result.

go/vt/vtgate/evalengine/string.go
```diff
@@ -0,0 +1,125 @@
...
+		}
+		for i, ch := range input {
+			if !unicode.IsDigit(ch) {
+				index = i
+				break
+			}
+		}
+		intstr := input[:index]
+		toint64, _ := strconv.ParseInt(intstr, 10, 64)
+		raw = strconv.AppendUint(raw, uint64(toint64), 2)
```
vmg, line 84: This doesn't look right. We already have a function that tries to parse numbers into `float` and handles this logic properly. I would try using that one, and then converting the `float64` to `uint64`. I have a hunch that's what MySQL is doing behind the scenes.

go/vt/vtgate/evalengine/string.go
```diff
@@ -0,0 +1,125 @@
...
+	if arg1.isNull() {
+		result.setNull()
+		return
+	}
+
+	switch t := arg1.typeof(); {
+	case sqltypes.IsQuoted(t):
+		result.setInt64(int64(8 * len(arg1.bytes())))
+	default:
+		result.setInt64(int64(8 * len(arg1.toRawBytes())))
```
vmg, line 115: Since `toRawBytes()` already handles the case where `IsQuoted(t)`, you can call it directly. You don't this switch at all.

vmg:
> For some case not defined in MySQL docs, there are different outputs between MySQL 5 and MySQL 8. Which one should I refer to?

That's great! We always target MySQL 8's behavior. It would be even better if you added a test for this specific mismatch with a comment that says it's not backwards compatible with MySQL 5.

go/vt/vtgate/evalengine/string.go
wanghaoyang1995: Thanks for your review!

In my consideration, the above code is mainly to deal with strings containing both letters and numbers, not only for float. Such as "123abc".

I tried to write a piece of code, firstly convert to float and then convert to uint64. Like this:

```go
func (builtinBin) call(env *ExpressionEnv, args []EvalResult, result *EvalResult) {
	tobin := &args[0]
	if tobin.isNull() {
		result.setNull()
		return
	}

	var raw []byte
	switch t := tobin.typeof(); {
	case sqltypes.IsNumber(t):
		tobin.makeUnsignedIntegral()
		raw = strconv.AppendUint(raw, tobin.uint64(), 2)
	case sqltypes.IsText(t):
		if len(tobin.bytes()) == 0 {
			return
		}
		tobin.makeFloat()
		float, _ := tobin.coerceToFloat()
		raw = strconv.AppendUint(raw, uint64(float), 2)
	default:
		throwEvalError(vterrors.Errorf(vtrpcpb.Code_UNIMPLEMENTED, "Unsupported BIN argument: %s", t.String()))
	}

	result.setRaw(sqltypes.VarChar, raw, collations.TypedCollation{
		Collation:    env.DefaultCollation,
		Coercibility: collations.CoerceCoercible,
		Repertoire:   collations.RepertoireASCII,
	})
}
```

It was well behaved when inputted ”123abc“ or "123.456".
But for strings like "2e2", it has a difference with MySQL's output.
```shell
mysql> select bin("2e2");
+------------+
| bin(:vtg1) |
+------------+
| 11001000   |
+------------+
1 row in set (1.42 sec)
```

While the MySQL5.7 returned `10`.
```shell
mysql> select bin("2e2");
+------------+
| bin("2e2") |
+------------+
| 10         |
+------------+
1 row in set (0.00 sec)
```

Is it possible to bypass string processing?
I'm not very familiar with go code and would like some advice from you, thanks!

Issue #9783, opened 2022-02-24
vstream: upgrade to v13 breaks vstream with invisible columns


derekperkins:
### Overview of the Issue

Support for invisible columns was added in 8.0.23. We use them in a few places to prevent unnecessary or larger blobs from being loaded on `select * from...` queries. That worked fine through v12, but after rolling a canary upgrade to v13, it now throws an error.
https://dev.mysql.com/doc/refman/8.0/en/invisible-columns.html

I think there are a few places where `SELECT *` has been used to return column lists, but I'm guessing that something is querying the information schema or something that causes a mismatch. This is an upgrade blocker for us, but upgrading isn't something urgent for us at the moment.

### Reproduction Steps

```markdown
CREATE TABLE `workspaces_rankings__scheduler_keyword_sources__msgs` (
  `id` bigint NOT NULL,
  `keyspace_id` bigint NOT NULL,
  `priority` tinyint NOT NULL DEFAULT '50',
  `epoch` smallint NOT NULL DEFAULT '0',
  `time_next` bigint DEFAULT NULL,
  `time_acked` bigint DEFAULT NULL,
  `time_scheduled` bigint NOT NULL /*!80023 INVISIBLE */,
  `time_created` bigint NOT NULL /*!80023 INVISIBLE */,
  `attributes` json DEFAULT NULL,
  `date` date NOT NULL /*!80023 INVISIBLE */,
  `team_id` bigint NOT NULL /*!80023 INVISIBLE */,
  `keyword_source_id` bigint NOT NULL /*!80023 INVISIBLE */,
  `keyword_source_version_id` bigint NOT NULL /*!80023 INVISIBLE */,
  PRIMARY KEY (`id`),
  KEY `ack_idx` (`time_acked`),
  KEY `next_idx` (`time_next`,`priority`),
  KEY `wr__scheduler_keyword_sources__msgs__kw_sources` (`keyspace_id`,`team_id`,`keyword_source_id`),
  CONSTRAINT `fk__workspaces_rankings__scheduler_keywords_sources_msgs__wksps` FOREIGN KEY (`keyspace_id`) REFERENCES `workspaces` (`workspace_id`),
  CONSTRAINT `wr__scheduler_keyword_sources__msgs__kw_sources` FOREIGN KEY (`keyspace_id`, `team_id`, `keyword_source_id`) REFERENCES `keyword_sources` (`workspace_id`, `team_id`, `keyword_source_id`)
[...]

PR #9794, opened 2022-02-27
Support invisible columns in vreplication

rohit-nayak-ps:
## Description

VReplication used queries like `select * from t1 where 1 = 0` for obtaining the list of columns and the `Fields` structure. This no longer works after invisible columns were introduced in `8.0.23` because `select *` does not return invisible columns. VReplication needs to do faithful copies / stream all data from a table. So it needs to also send the data from invisible columns.

This PR gets the list of all columns, including invisible ones, from `information_schema.columns` and passing on this list explicitly to the `select` clause to get the Fields . It also changes the queries to find primary keys to use `information_schema.columns` instead of `information_schema.key_column_usage`. The latter does not contain info about invisible columns.

The core changes for this PR are in `go/vt/vttablet/tabletserver/schema/load_table.go` and `go/vt/mysqlctl/schema.go`. 

Most of the remaining modifications are related to changes in the existing unit tests for mocking the updated queries, in test frameworks and new tests.

## Related Issue(s)
#9783 

## Checklist

### DIFF
--- a/go/mysql/fakesqldb/server.go
+++ b/go/mysql/fakesqldb/server.go
@@ -98,7 +98,7 @@ type DB struct {
 	// rejectedData maps tolower(query) to an error.
 	rejectedData map[string]error
 	// patternData is a list of regexp to results.
-	patternData []exprResult
+	patternData map[string]exprResult
 	// queryCalled keeps track of how many times a query was called.
 	queryCalled map[string]int
 	// querylog keeps track of all called queries
@@ -120,6 +120,8 @@ type DB struct {
 
 	// queryPatternUserCallback stores optional callbacks when a query with a pattern is called
 	queryPatternUserCallback map[*regexp.Regexp]func(string)
+
+	neverFail bool
 }
 
 // QueryHandler is the interface used by the DB to simulate executed queries
@@ -170,6 +172,7 @@ func New(t testing.TB) *DB {
 		queryCalled:              make(map[string]int),
 		connections:              make(map[uint32]*mysql.Conn),
 		queryPatternUserCallback: make(map[*regexp.Regexp]func(string)),
+		patternData:              make(map[string]exprResult),
 	}
 
 	db.Handler = db
@@ -341,7 +344,6 @@ func (db *DB) HandleQuery(c *mysql.Conn, query string, callback func(*sqltypes.R
 		}
 		return callback(result)
 	}
-
 	key := strings.ToLower(query)
 	db.mu.Lock()
 	defer db.mu.Unlock()
@@ -397,8 +399,13 @@ func (db *DB) HandleQuery(c *mysql.Conn, query string, callback func(*sqltypes.R
 		}
 	}
 
+	if db.neverFail {
+		return callback(&sqltypes.Result{})
+	}
 	// Nothing matched.
-	return fmt.Errorf("query: '%s' is not supported on %v", query, db.name)
+	err := fmt.Errorf("query: '%s' is not supported on %v", query, db.name)
+	log.Errorf("Query not found: %s", query)
+	return err
 }
 
 func (db *DB) comQueryOrdered(query string) (*sqltypes.Result, error) {
@@ -513,7 +520,7 @@ func (db *DB) AddQueryPattern(queryPattern string, expectedResult *sqltypes.Resu
 	result := *expectedResult
 	db.mu.Lock()
 	defer db.mu.Unlock()
-	db.patternData = append(db.patternData, exprResult{expr: expr, result: &result})
+	db.patternData[queryPattern] = exprResult{expr: expr, result: &result}
 }
 
 // RejectQueryPattern allows a query pattern to be rejected with an error
@@ -521,19 +528,19 @@ func (db *DB) RejectQueryPattern(queryPattern, error string) {
 	expr := regexp.MustCompile("(?is)^" + queryPattern + "$")
 	db.mu.Lock()
 	defer db.mu.Unlock()
-	db.patternData = append(db.patternData, exprResult{expr: expr, err: error})
+	db.patternData[queryPattern] = exprResult{expr: expr, err: error}
 }
 
 // ClearQueryPattern removes all query patterns set up
 func (db *DB) ClearQueryPattern() {
-	db.patternData = nil
+	db.patternData = make(map[string]exprResult)
 }
 
 // AddQueryPatternWithCallback is similar to AddQueryPattern: in addition it calls the provided callback function
 // The callback can be used to set user counters/variables for testing specific usecases
 func (db *DB) AddQueryPatternWithCallback(queryPattern string, expectedResult *sqltypes.Result, callback func(string)) {
 	db.AddQueryPattern(queryPattern, expectedResult)
-	db.queryPatternUserCallback[db.patternData[len(db.patternData)-1].expr] = callback
+	db.queryPatternUserCallback[db.patternData[queryPattern].expr] = callback
 }
 
 // DeleteQuery deletes query from the fake DB.
@@ -716,4 +723,8 @@ func (db *DB) VerifyAllExecutedOrFail() {
 	if db.expectedExecuteFetchIndex != len(db.expectedExecuteFetch) {
 		db.t.Errorf("%v: not all expected queries were executed. leftovers: %v", db.name, db.expectedExecuteFetch[db.expectedExecuteFetchIndex:])
 	}
+}
+
+func (db *DB) SetNeverFail(neverFail bool) {
+	db.neverFail = neverFail
 }
--- a/go/mysql/schema.go
+++ b/go/mysql/schema.go
@@ -32,9 +32,11 @@ import (
 
 const (
 	// BaseShowPrimary is the base query for fetching primary key info.
-	BaseShowPrimary = "SELECT table_name, column_name FROM information_schema.key_column_usage WHERE table_schema=database() AND constraint_name='PRIMARY' ORDER BY table_name, ordinal_position"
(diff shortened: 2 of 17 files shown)


go/mysql/schema.go
```diff
@@ -32,9 +32,11 @@ import (
 
 const (
 	// BaseShowPrimary is the base query for fetching primary key info.
-	BaseShowPrimary = "SELECT table_name, column_name FROM information_schema.key_column_usage WHERE table_schema=database() AND constraint_name='PRIMARY' ORDER BY table_name, ordinal_position"
-	// BaseShowTableUniqueKey returns names of colunms covered by a given unique constraint on a given table, in key order
-	BaseShowTableUniqueKey = "SELECT column_name as column_name FROM information_schema.key_column_usage WHERE table_schema=database() AND table_name=%a AND constraint_name=%a ORDER BY ordinal_position"
+	BaseShowPrimary = `SELECT table_name as table_name, COLUMN_NAME as column_name
```
mattlord, line 35: Nit: I would remove the column alias when it matches the column name to shorten and simplify the query (is it just that we want it in lower case (for the map) whereas the schema uses upper case?). I'd also make the case usage consistent. It can matter depending on the collation used and case sensitivity of the system.

go/test/endtoend/vreplication/config.go
```diff
@@ -14,7 +14,8 @@ var (
 	initialProductSchema = `
 create table product(pid int, description varbinary(128), date1 datetime not null default '0000-00-00 00:00:00', date2 datetime not null default '2021-00-01 00:00:00', primary key(pid)) CHARSET=utf8mb4;
 create table customer(cid int, name varchar(128) collate utf8mb4_bin, meta json default null, typ enum('individual','soho','enterprise'), sport set('football','cricket','baseball'),
-	ts timestamp not null default current_timestamp, bits bit(2) default b'11', date1 datetime not null default '0000-00-00 00:00:00', date2 datetime not null default '2021-00-01 00:00:00', primary key(cid)) CHARSET=utf8mb4;
+	ts timestamp not null default current_timestamp, bits bit(2) default b'11', date1 datetime not null default '0000-00-00 00:00:00', 
+	date2 datetime not null default '2021-00-01 00:00:00', primary key(cid)) CHARSET=utf8mb4;
```
mattlord, line 18: Just a note that this will cause a different collation to be used, depending on the DB version (5.7 will use utf8mb4_general_ci and 8.0 will use utf8mb4_0900_ai_ci).

go/vt/mysqlctl/schema.go
```diff
@@ -266,15 +265,60 @@ func ResolveTables(ctx context.Context, mysqld MysqlDaemon, dbName string, table
...
-// GetColumns returns the columns of table.
-func (mysqld *Mysqld) GetColumns(ctx context.Context, dbName, table string) ([]*querypb.Field, []string, error) {
-	conn, err := getPoolReconnect(ctx, mysqld.dbaPool)
+const (
+	GetColumnNamesQuery = `SELECT COLUMN_NAME as column_name
+		FROM INFORMATION_SCHEMA.COLUMNS
+		WHERE TABLE_SCHEMA = %s AND TABLE_NAME = '%s'
+		ORDER BY ORDINAL_POSITION`
+	GetFieldsQuery = "SELECT %s FROM %s WHERE 1 != 1"
+)
```
mattlord, line 274: I wonder if this is going to behave as we'd like when `GENERATED ALWAYS` (virtual) columns are used as you cannot specify a value on INSERT: https://dev.mysql.com/doc/refman/5.7/en/create-table-generated-columns.html
```
mysql> create table gt (col1 int, col2 int, col3 bigint generated always as (col1+col2));

mysql> select * from gt where 1!=1;
Field   1:  `col1`
Catalog:    `def`
Database:   `mydb`
Table:      `gt`
Org_table:  `gt`
Type:       LONG
Collation:  binary (63)
Length:     11
Max_length: 0
Decimals:   0
Flags:      NUM

Field   2:  `col2`
Catalog:    `def`
Database:   `mydb`
Table:      `gt`
Org_table:  `gt`
Type:       LONG
Collation:  binary (63)
Length:     11
Max_length: 0
Decimals:   0
Flags:      NUM

Field   3:  `col3`
Catalog:    `def`
Database:   `mydb`
Table:      `gt`
Org_table:  `gt`
Type:       LONGLONG
Collation:  binary (63)
Length:     20
Max_length: 0
Decimals:   0
Flags:      NUM

0 rows in set (0.00 sec)

mysql> insert into gt values (1, 2, 3);
ERROR 3105 (HY000): The value specified for generated column 'col3' in table 'gt' is not allowed.
```

I feel like we should leverage information_schema for everything or we're going to run into various edge cases over time (e.g. [server side query rewrites](https://dev.mysql.com/doc/refman/8.0/en/rewriter-query-rewrite-plugin.html) is another potential issue):
```
mysql> select * from information_schema.columns where table_name="gt"\G
           TABLE_CATALOG: def
            TABLE_SCHEMA: mydb
              TABLE_NAME: gt
             COLUMN_NAME: col1
        ORDINAL_POSITION: 1
          COLUMN_DEFAULT: NULL
             IS_NULLABLE: YES
               DATA_TYPE: int
CHARACTER_MAXIMUM_LENGTH: NULL
  CHARACTER_OCTET_LENGTH: NULL
       NUMERIC_PRECISION: 10
           NUMERIC_SCALE: 0
      DATETIME_PRECISION: NULL
      CHARACTER_SET_NAME: NULL
          COLLATION_NAME: NULL
             COLUMN_TYPE: int
              COLUMN_KEY:
                   EXTRA:
              PRIVILEGES: select,insert,update,references
          COLUMN_COMMENT:
   GENERATION_EXPRESSION:
                  SRS_ID: NULL
*************************** 2. row ***************************
           TABLE_CATALOG: def
            TABLE_SCHEMA: mydb
              TABLE_NAME: gt
             COLUMN_NAME: col2
        ORDINAL_POSITION: 2
          COLUMN_DEFAULT: NULL
             IS_NULLABLE: YES
               DATA_TYPE: int
CHARACTER_MAXIMUM_LENGTH: NULL
  CHARACTER_OCTET_LENGTH: NULL
       NUMERIC_PRECISION: 10
           NUMERIC_SCALE: 0
      DATETIME_PRECISION: NULL
      CHARACTER_SET_NAME: NULL
          COLLATION_NAME: NULL
             COLUMN_TYPE: int
              COLUMN_KEY:
                   EXTRA:
              PRIVILEGES: select,insert,update,references
          COLUMN_COMMENT:
   GENERATION_EXPRESSION:
                  SRS_ID: NULL
*************************** 3. row ***************************
           TABLE_CATALOG: def
            TABLE_SCHEMA: mydb
              TABLE_NAME: gt
             COLUMN_NAME: col3
        ORDINAL_POSITION: 3
          COLUMN_DEFAULT: NULL
             IS_NULLABLE: YES
               DATA_TYPE: bigint
CHARACTER_MAXIMUM_LENGTH: NULL
  CHARACTER_OCTET_LENGTH: NULL
       NUMERIC_PRECISION: 19
           NUMERIC_SCALE: 0
      DATETIME_PRECISION: NULL
      CHARACTER_SET_NAME: NULL
          COLLATION_NAME: NULL
             COLUMN_TYPE: bigint
              COLUMN_KEY:
                   EXTRA: VIRTUAL GENERATED
              PRIVILEGES: select,insert,update,references
          COLUMN_COMMENT:
   GENERATION_EXPRESSION: (`col1` + `col2`)
                  SRS_ID: NULL
3 rows in set (0.00 sec)
```

It shouldn't be an issue in replicating binary log events (I don't think), but it might cause problems in the copy rows phase?

mattlord (commented):
I have some vague concerns as noted, we can discuss directly (my concerns may be unfounded or can be deferred).

mattlord (approved):
@rohit-nayak-ps and I sat down together and addressed my concerns so LGTM!

go/mysql/fakesqldb/server.go
```diff
@@ -120,6 +120,10 @@ type DB struct {
 
 	// queryPatternUserCallback stores optional callbacks when a query with a pattern is called
 	queryPatternUserCallback map[*regexp.Regexp]func(string)
+
+	// if fakesqldb is asked to serve queries or query patterns that it has not been explicitly told about it will
+	// error out by default. However if you set this flag then any unmatched query results in an empty result
+	neverFail bool
```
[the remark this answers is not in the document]
rohit-nayak-ps: I am not using that functionality in any of my tests. There was an existing query executor test at https://github.com/vitessio/vitess/pull/9794/files#diff-4742074fc13ea6034afd3b30f3ed3946442b195719e731e98f037f2bb502476cR493 which was doing the same thing using `db.AddQueryPattern(".*", &sqltypes.Result{})`. I spent a lot of time debugging a failure around this and decided to make it more explicit.

go/mysql/schema.go
rohit-nayak-ps: The column alias to use lower case has been used in a  lot of places. We use the column name as an alias when we convert the query result to a map with the column name as the key using the `.Named()` mechanism. MySQL can return lower or upper case depending on how it is configured. This caused a difference in case between the code and what MySQL returned. Hence the explicit aliasing. 

I agree we should change this. One idea was to add both upper and lower cases to the map so that both are found, but that would also be a hack which wouldn't work if someone used camel case, for example!

go/vt/mysqlctl/schema.go
rohit-nayak-ps: Generated columns are handled by vreplication, so we drop them from the generated table plans.

derekperkins:
Thanks for the quick work on this @rohit-nayak-ps!

PR #9800, opened 2022-02-28
Add CHAR field comparison support to evalengine (for VDiff)

mattlord:
## Description
We switched to using evalengine for `VDiff` comparisons in https://github.com/vitessio/vitess/pull/9679.

What we didn't realize at the time is that evalengine did not support comparing `CHAR` columns. When doing a `VDiff` against a table that has them, you get the following error (you can see a failed test after the test changes [here](https://github.com/vitessio/vitess/runs/5364912331?check_suite_focus=true#step:9:238)):
```
VDiff Error: rpc error: code = Unknown desc = diff: types are not comparable: CHAR vs CHAR
```

There's no logical reason for this as a `CHAR` field for in memory processing is no different from a `VARCHAR`. The main differences are the on disk format which stores trailing padding, but that padding is stripped in the value returned in the resultset (and [even if not](https://dev.mysql.com/doc/refman/8.0/en/sql-mode.html#sqlmode_pad_char_to_full_length), we don't care for the purposes of comparing the in memory values). This PR treats `CHAR` fields the same as `VARCHAR` fields for comparisons. 

## Related Issue(s)
- Related to: https://github.com/vitessio/vitess/pull/9679
- Related to: https://github.com/vitessio/vitess/pull/9790

## Checklist
- [x] Tests were updated


vmg (approved):
Hmpf, the field types coming from VDiff are very different from what we are testing against MySQL by using the query API. I need to improve these steps so we stop crashing because of oversights like this.

mattlord:
> Hmpf, the field types coming from VDiff are very different from what we are testing against MySQL by using the query API. I need to improve these steps so we stop crashing because of oversights like this.

Thank you! Please let me know if I can be of any help.

Issue #9801, opened 2022-02-28
Bug Report: Hot Row Protection logs are not sanitized with -sanitize_log_messages flag


Anders-PlanetScale:
### Overview of the Issue

When utilizing -sanitize_log_messages vttablet flag logs written by HotRowProtection are not properly sanitized.

### Reproduction Steps

```markdown
Use -enable_hot_row_protection	and -sanitize_log_messages flags on vttablet.

Attempt simultaneous transactions against the same row.

HotRowProtection logging will print row information unsanitized to the vttablet log.
```

### Binary Version

```shell
Version: 14.0.0-SNAPSHOT (Git revision ccfe290 branch 'main') built on Fri Feb 25 05:06:14 UTC 2022 by vitess@buildkitsandbox using go1.17.6 linux/amd64
```

### Operating System and Environment details

```shell
Debian GNU/Linux
Linux 5.11.0-1027-aws
x86_64
```

### Log Fragments

```shell
Sensitive information so not going to log here.
```

PR #9802, opened 2022-03-01
Sanitize tx serializer log & error messages

mattlord:
## Description
The transaction serializer — which offers "hot row protection" that can be enabled in vttablets using the [`-enable_hot_row_protection`](https://vitess.io/docs/14.0/reference/programs/vttablet/) flag — logs the `key` for its map at various times. The key consists of the table name and the `WHERE` clause and the `WHERE` clause can contain sensitive info so we should be sanitizing it for log messages when the vttablet [`-sanitize_log_messages`](https://vitess.io/docs/14.0/reference/programs/vttablet/) flag is enabled. 

We should also be sanitizing the key in the returned error message when the vttablet [`-queryserver-config-terse-errors`](https://vitess.io/docs/14.0/reference/programs/vttablet/) flag is enabled.

## Related Issue(s)
 - Fixes: https://github.com/vitessio/vitess/issues/9801

## Checklist
- [x] Tests were added

### DIFF
--- a/go/vt/vttablet/tabletserver/txserializer/tx_serializer.go
+++ b/go/vt/vttablet/tabletserver/txserializer/tx_serializer.go
@@ -21,6 +21,7 @@ package txserializer
 import (
 	"fmt"
 	"net/http"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


deepthi (approved):
LGTM

PR #9804, opened 2022-03-01
evalengine: more comprehensive type coverage

vmg:
## Description

Since we've started using the evaluation engine for VDiff, @mattlord has found several issues where our type coverage was not good enough and we were failing to operate on types that were (in theory) supported but that we failed to handle because our type checks where too constricted (e.g. `TEXT` vs `CHAR` vs `VARCHAR`).

This PR attempts to fix any further cases that may come up in the future by switching the `evalengine` to using the `sqltypes.IsXXX` category checks instead of explicitly comparing against specific types. It also cleans up some technical debt by renaming some helper methods which were IMO a bit confusing.

## Checklist

### DIFF
--- a/go/vt/vtgate/evalengine/eval_result.go
+++ b/go/vt/vtgate/evalengine/eval_result.go
@@ -17,7 +17,6 @@ limitations under the License.
 package evalengine
 
 import (
-	"bytes"
 	"encoding/binary"
 	"fmt"
 	"math"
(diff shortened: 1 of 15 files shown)


go/vt/vtgate/evalengine/eval_result.go
```diff
@@ -339,16 +371,16 @@ func (er *EvalResult) makeTextualAndConvert(collation collations.ID) bool {
 }
 
 func (er *EvalResult) truncate(size int) {
```
mattlord, line 373: nit but IMO the function should be called ~ `enforceSize` since we truncate or pad as needed.

mattlord (approved):
Thank you! ❤️

go/vt/vtgate/evalengine/eval_result.go
vmg: :+1: will refactor

PR #9809, opened 2022-03-02
replmanager: Let TabletManger decide if replication needs to be repaired

dctrwatson:

## Description
If a tablet's Primary's hostname or port changes before the tablet has replicated any data, the tablet will not attempt to repair replication. So this adds a specific check to make sure the tablet's configured source still corresponds to the shard's primary tablet's info.

## Checklist

### DIFF
--- a/go/vt/vttablet/tabletmanager/replmanager.go
+++ b/go/vt/vttablet/tabletmanager/replmanager.go
@@ -17,6 +17,7 @@ limitations under the License.
 package tabletmanager
 
 import (
+	"fmt"
 	"os"
 	"path"
 	"sync"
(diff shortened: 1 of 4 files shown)


deepthi:
@mattlord how does this interact with #9789 (if at all)?

deepthi (commented):
@dctrwatson can you fix the import errors? That's what is causing a bunch of test failures as well.

mattlord:
> @mattlord how does this interact with #9789 (if at all)?

No impact. The only functional change in #9789 is that we no longer consider replication to be running and healthy if the replica has 0 for the seconds behind value and its io status is `connecting` IF there was an IO error the last time we tried to connect (this seems to happen the first time a replica tries to connect to its source, if it cannot reach it).

go/vt/vttablet/tabletmanager/replmanager.go
```diff
@@ -110,7 +132,10 @@ func (rm *replManager) checkActionLocked() {
 		// If only one of the threads is stopped, it's probably
 		// intentional. So, we don't repair replication.
 		if status.SQLThreadRunning || status.IOThreadRunning {
-			return
+			// Check if the Primary is still the source though
+			if primaryIsSource, err := rm.checkPrimaryIsSource(status); err == nil && primaryIsSource {
+				return
+			}
```
GuptaManan100, line 138: In case we get an error here, the error will be from a topo call either for `GetShard` or for `GetTablet`. In either case, I don't think we should try to fix the replication since the repairReplication will try to access the topo too. We should return in that case and wait for the next tick like we do in case of errors in above lines.

GuptaManan100 (commented):
I have 2 comments -  
1. As per our plan, we will be making vtorc GA and a compulsory component of Vitess in release-14. When VTOrc starts fixing the replication, the tablet will not need to do it. We actually run the end to end tests with `disable_active_reparents` turned on so that the replicationManager does not work. With this background, I am unsure whether we should be fixing this codepath at all, since we might just remove it entrirely.
2. Should we add unit testing for these changes?

mattlord:
@dctrwatson from your PR description... I wonder if this PR is necessary anymore. I think it may be the same symptom seen here: https://github.com/vitessio/vitess/pull/9789 That was fixed in: https://github.com/vitessio/vitess/pull/9853

The the underlying problem, I think, being here for new tablets starting up: https://github.com/vitessio/vitess/pull/9789/files#r820941187

PR #9810, opened 2022-03-02
Split aggregation in scalar and ordered aggregation

systay:
## Description
For aggregations without grouping columns (`select count(*) from tbl`), we could use a simple engine primitive.
This PR implements the scalar aggregation engine primitive, that handles these types of situations.

## Checklist
- [x] Should this PR be backported? *no*

### DIFF
--- a/go/vt/vtgate/engine/scalar_aggregation.go
+++ b/go/vt/vtgate/engine/scalar_aggregation.go
@@ -97,7 +97,6 @@ func (sa *ScalarAggregate) TryExecute(vcursor VCursor, bindVars map[string]*quer
 		if err != nil {
 			return nil, err
 		}
-		continue
 	}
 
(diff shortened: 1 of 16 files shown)


frouioui (approved):
Looks good to me!

go/vt/vtgate/engine/scalar_aggregation.go
```diff
@@ -0,0 +1,193 @@
...
+				return err
+			}
+		}
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+
+	return cb(&sqltypes.Result{Rows: [][]sqltypes.Value{current}})
```
harshit-gangal, line 157: the original oa code has a check for 
`current != nil `
does this make a difference if you are not performing the nil check on current?

PR #9813, opened 2022-03-02
Online DDL: ready_to_complete hint column

shlomi-noach:
## Description

This PR adds `ready_to_complete` column in `_vt.schema_migrations` which is then visible in `SHOW VITESS_MIGRATIONS ...` output.

The use case is intended for postponed migrations (although it provides visibility for non-postponed migrations as well): if a migration is postponed, the user wants to know "is now a good time to cut-over" ie run a `ALTER VITESS_MIGRATION ... COMPLETE` and expect a near-immediate cut-over.

`ready_to_complete` is a boolean (values `0` or `1`) that indicates that the migration is generally capable of cutting over. Specifically:

- `CREATE` and `DROP` migrations will always exhibit the value of `1` (immediately or shortly after submission). Note that this is true even if the migrations are still `queued`. Once they exhibit `1`, they stay with `1`.
- `vitess` `ALTER` migrations will indicate `1` when row-copy is complete and vplayer's lag is generally good. this value can drop back to `0` when the migration is _postponed_ and vplayer happens to accumulate lag. Once vplayer is caught up again, it will flip back to `1`.
- A `REVERT` of any of the above behaves similarly: a `CREATE` `REVERT` immediately exhibits the value of `1`; an `ALTER`'s `REVERT` exhibits `1` when vplayer's lag is good and `0` when vplayer accumulates lag
- All the above is also true for postponed migrations
- `gh-ost` `ALTER` migrations indicate `1` when entering postponed state, and does not have lag indication. It is only applicable to `postponed-completion` migrations.

Note that all changes to `onSchemaMigrationStatus` are for supporting `gh-ost` by adding a hint column to `gh-ost` callback API; it's not the important part of this PR.

## Related Issue(s)

Tracking: #6926 

## Checklist

### DIFF
--- a/go/test/endtoend/onlineddl/ghost/onlineddl_ghost_test.go
+++ b/go/test/endtoend/onlineddl/ghost/onlineddl_ghost_test.go
@@ -159,8 +159,8 @@ func TestMain(m *testing.M) {
 		clusterInstance.VtctldExtraArgs = []string{
 			"-schema_change_dir", schemaChangeDirectory,
 			"-schema_change_controller", "local",
-			"-schema_change_check_interval", "1",
 			"-online_ddl_check_interval", "3s",
+			"-schema_change_check_interval", "1",
 		}
 
 		clusterInstance.VtTabletExtraArgs = []string{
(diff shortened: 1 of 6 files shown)


go/test/endtoend/onlineddl/ghost/onlineddl_ghost_test.go
```diff
@@ -159,8 +159,8 @@ func TestMain(m *testing.M) {
 		clusterInstance.VtctldExtraArgs = []string{
 			"-schema_change_dir", schemaChangeDirectory,
 			"-schema_change_controller", "local",
-			"-schema_change_check_interval", "1",
 			"-online_ddl_check_interval", "3s",
+			"-schema_change_check_interval", "1",
```
dbussink, line 163: Doesn't matter, but any reason this order of arguments was changed?

go/vt/vttablet/onlineddl/executor.go
```diff
@@ -1635,9 +1641,40 @@ func (e *Executor) scheduleNextMigration(ctx context.Context) error {
...
+			case sqlparser.CreateStr, sqlparser.DropStr:
+				if err := e.updateMigrationReadyToComplete(ctx, uuid, true); err != nil {
+					return err
+				}
+			}
+		}
+		if ddlAction == sqlparser.AlterStr || !postponeCompletion {
+			// Any non-postponed migration can be scheduled
+			// postponed ALTER can be scheduled
+			// We only schedule a single migration i nthe execution of this function
```
dbussink, line 1669: Typo here, `i nthe` instead of `in the`.

dbussink (approved):
Looks like a pretty straightforward way to solve this issue. One small question and found a typo :smile:.

go/test/endtoend/onlineddl/ghost/onlineddl_ghost_test.go
shlomi-noach: Bedtime story: it turns out the `onlineddl_ghost` endtoend tests have not been running in a while. There was an exception thrown in this old `TestMain` function that was not caught (in itself something to investigate). What I did now was to copy+paste a brand new `TestMain` function from `onlineddl_vrepl` tests, which works.

Having said that, I'll reorder the arguments just to keep this PR cleaner.

go/vt/vttablet/onlineddl/executor.go

PR #9817, opened 2022-03-02
Run WithDDL queries on tablet startup

rohit-nayak-ps:
## Description

This PR runs a "fake" query using WithDDL on tablet startup so that we reach the desired schema right at the start rather than auto-magically during a WithDDL.Exec() invocation.

Since we don't enforce the usage of WithDDL.Exec() statically, it is possible for code to execute queries directly and cause regressions. Example: new code on an existing cluster expects a new column but does not use WithDDL.Exec() and hence that column never gets created.

## Checklist

### DIFF
--- a/go/vt/binlog/binlogplayer/mock_dbclient.go
+++ b/go/vt/binlog/binlogplayer/mock_dbclient.go
@@ -22,6 +22,8 @@ import (
 	"testing"
 	"time"
 
+	"vitess.io/vitess/go/vt/withddl"
+
 	"vitess.io/vitess/go/sqltypes"
 )
 
@@ -49,7 +51,10 @@ type mockExpect struct {
 
 func getQueriesToIgnore() []*mockExpect {
 	var queriesToIgnore []*mockExpect
-	for _, query := range WithDDLInitialQueries {
+	var queries []string
+	queries = append(queries, WithDDLInitialQueries...)
+	queries = append(queries, withddl.QueryToTriggerWithDDL)
+	for _, query := range queries {
 		exp := &mockExpect{
 			query:  query,
 			re:     nil,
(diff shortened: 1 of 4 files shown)


shlomi-noach (approved):
Looks good!

shlomi-noach (approved):
Approved again after additional changes have been made.

deepthi (approved):

deepthi:
@rohit-nayak-ps can you link the original issue+PR that motivated this change?

mattlord (commented):
LGTM

shlomi-noach:
@rohit-nayak-ps this turns out to not work correctly. The table/schema does not actually get created upon `Open()`. The reason is that `ensureValidSchema` calls `withDDL.ExecIgnore`. 
`withDDL.ExecIgnore` does _not_ actually run the `withDDL` sequence of schema changes on MySQL error.

https://github.com/vitessio/vitess/pull/9817/files#diff-87d780c52f2f803c01e5ab450a6df1d38c055acd9e0cf1d5f60022a08671f313R224

rohit-nayak-ps:
> @rohit-nayak-ps this turns out to not work correctly. The table/schema does not actually get created upon `Open()`. The reason is that `ensureValidSchema` calls `withDDL.ExecIgnore`. `withDDL.ExecIgnore` does _not_ actually run the `withDDL` sequence of schema changes on MySQL error.

My Duh :-( 

Changed to use `WithDDL.Exec()` in https://github.com/vitessio/vitess/pull/9860

mattlord:
Sorry that I didn't catch that in review! Thanks for noting it @shlomi-noach.

shlomi-noach:
I didn't catch it either 😛

PR #9820, opened 2022-03-03
mysql: add a flag to handle EnableQueryInfo

vmg:
## Description

We added the functionality to handle processing these fields a while ago, but it was disabled by default. This PR adds a flag, that will make itself available in all VTTablet instances, that lets users control this functionality at launch.

cc @piki @deepthi 

## Checklist


piki (approved):
LGTM

deepthi (approved):
Seems reasonable to make this behavior opt-in.

Issue #9826, opened 2022-03-03
Bug Report: segfault in StreamMigrator.identifyRuleType


mvh-stripe:
### Overview of the Issue

This issue is related to a previous fix: https://github.com/vitessio/vitess/issues/9576
In the previous fix, the segfault was fixed in the code in `Resharder`. But the same bug also exists in the `StreamMigrator`.

```
2022-03-03T00:34:28.022066078Z stderr F I0303 00:34:28.021980       1 locks.go:214] Locking keyspace devmvhks26 for action SwitchWrites
2022-03-03T00:34:28.234463345Z stderr F I0303 00:34:28.234351       1 traffic_switcher.go:481] No previous journals were found. Proceeding normally.
2022-03-03T00:34:28.24433431Z stderr F I0303 00:34:28.244247       1 stream_migrator.go:763] found internal table _0d5669fe_9a86_11ec_8418_021fd2308fa7_20220303000829_vrepl, ignoring in rule identification
2022-03-03T00:34:28.244453719Z stderr F I0303 00:34:28.244401       1 traffic_switcher.go:492] Stopping streams
2022-03-03T00:34:28.246447989Z stderr F panic: runtime error: invalid memory address or nil pointer dereference
2022-03-03T00:34:28.246457813Z stderr F [signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xf1a812]
2022-03-03T00:34:28.246463176Z stderr F
2022-03-03T00:34:28.246471758Z stderr F goroutine 422 [running]:
2022-03-03T00:34:28.246492528Z stderr F vitess.io/vitess/go/vt/vtctl/workflow.(*StreamMigrator).identifyRuleType(0xc0000ae210, 0xc000e822d0)
2022-03-03T00:34:28.246500321Z stderr F         vitess.io/vitess/go/vt/vtctl/workflow/stream_migrator.go:769 +0xf2
2022-03-03T00:34:28.246505908Z stderr F vitess.io/vitess/go/vt/vtctl/workflow.(*StreamMigrator).blsIsReference(0xc000002814, 0xc000e8c2c0)
2022-03-03T00:34:28.246531523Z stderr F         vitess.io/vitess/go/vt/vtctl/workflow/stream_migrator.go:735 +0x7c
[...]

PR #9828, opened 2022-03-03
Check for nil vschema.Table in StreamMigrator

mattlord:
## Description
We were not checking for this when it can be null due to processing an internal table such as those used for OnlineDDL.

## Related Issue(s)
 - Follow-up to: https://github.com/vitessio/vitess/pull/9578
 - Fixes: https://github.com/vitessio/vitess/issues/9826

## Checklist
- [x] Tests are not required
- [x] Documentation not required


rohit-nayak-ps (approved):
lgtm

PR #9830, opened 2022-03-04
Add vitess/lite ubi8 images for mysql80, including for arm64

aquarapid:
Today, Vitess has no facility for building non-x86_64 vitess/lite images.  This PR fixes that.  To simplify matters, I used the Red Hat ubi8 images, which has good aarch64 support.  I then added a vitess/lite image for both the ubi8/amd64(x86_64) and ubi8/arm64(aarch64) combos.

To build this on an Intel machine, you will need your local docker installation to support cross compiling for linux/arm64.  The easiest way to achieve this is to use https://github.com/tonistiigi/binfmt ,  specifically by running something like:

```
$ docker run --privileged --rm tonistiigi/binfmt --install arm64
```

before doing your build.

You can confirm that your installation is capable of cross-compilation by doing something like:

```
$ docker buildx ls
NAME/NODE DRIVER/ENDPOINT STATUS  PLATFORMS
default * docker                  
  default default         running linux/amd64, linux/386, linux/arm64
```

Note the `linux/arm64` option.

If you just want to build the 2 new images, you can do that:

for amd64:
```
$ make docker_lite_ubi8.mysql80
```

for arm64:
```
$ make docker_lite_ubi8.arm64.mysql80
```

Future cross-build docker containers just need to include the `arm64` string in the build target name (i.e. ubi8.arm64.mysql80 in this case) to trigger the cross-compilation.

Finally, the lite container for arm64/aarch64 excludes a few things that are not available on aarch64:
  * xtrabackup
  * Percona Toolkit
xtrabackup is the big one here;  so if you use these images to run your Vitess MySQL instances, you will have to either rely on the "builtin" (offline) backup;  or you will have to supply your own xtrabackup binaries (in theory it is possible to build it on aarch64, but Percona does not test this or provide binaries, yet).


Makefile
```diff
@@ -87,7 +87,12 @@ endif
 	# For the specified GOOS + GOARCH, build all the binaries by default with CGO disabled
 	CGO_ENABLED=0 GOOS=${GOOS} GOARCH=${GOARCH} go install -trimpath $(EXTRA_BUILD_FLAGS) $(VT_GO_PARALLEL) -ldflags "$(shell tools/build_version_flags.sh)" ./go/...
 	# unset GOOS and embed local resources in the vttablet executable
-	(cd go/cmd/vttablet && unset GOOS && unset GOARCH && go run github.com/GeertJohan/go.rice/rice --verbose append --exec=$${HOME}/go/bin/${GOOS}_${GOARCH}/vttablet)
+	if [ -d /go/bin ]; then
```
vmg, line 90: This can be made more reliable with `[ -f /run/.dockerenv ] || [ -f /run/.containerenv ]` (from some cursory googling

vmg (approved):
This looks good! How many Vitess contributors have M1 macs that could test this? It would be nice to have some confidence in the implementation besides the cross-compilation way.

deepthi:
> This looks good! How many Vitess contributors have M1 macs that could test this? It would be nice to have some confidence in the implementation besides the cross-compilation way.

On a M1 Mac running zsh
```
$ make docker_lite_ubi8.arm64.mysql80
Building vitess/lite:ubi8.arm64.mysql80
/bin/bash: -c: line 1: syntax error: unexpected end of file
make: *** [docker_lite_ubi8.arm64.mysql80] Error 2
```

aquarapid:
I don't have a Mac, so I cannot test any of this there.

aquarapid:
I cannot reproduce, even using zsh on Linux.

Makefile
aquarapid: Yeah, the comment here was to imply that we're inside the bootstrap image;  we don't really need to know if we're inside a container.  Will update the comment.

Makefile
```diff
@@ -87,7 +87,12 @@ endif
...
+		# Probably in the bootstrap container
+		(cd go/cmd/vttablet && go run github.com/GeertJohan/go.rice/rice --verbose append --exec=/go/bin/${GOOS}_${GOARCH}/vttablet)
+	else
+		(cd go/cmd/vttablet && unset GOOS && unset GOARCH && go run github.com/GeertJohan/go.rice/rice --verbose append --exec=$${HOME}/go/bin/${GOOS}_${GOARCH}/vttablet)
+	fi
```
mattlord, line 95: Curious why we don't rely on `${GOPATH}/bin` here? It is set inside the container, and should be set on any other system:
```
$ echo "GOBIN: ${GOPATH}/bin"                                                                                                                                                    
GOBIN: /Users/matt/go/bin
$ docker run -it vitess/bootstrap:mysql80 /bin/bash -c 'echo "GOBIN: ${GOPATH}/bin"'
WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
GOBIN: /go/bin
```

Would there be any other reason to use a different command?

Makefile
```diff
@@ -105,6 +110,13 @@ install: build
 	mkdir -p "$${PREFIX}/bin"
 	cp "$${VTROOT}/bin/"{mysqlctl,mysqlctld,vtorc,vtadmin,vtctld,vtctlclient,vtctldclient,vtgate,vttablet,vtworker,vtbackup} "$${PREFIX}/bin/"
 
+# Will only work inside the docker bootstrap for now
```
mattlord, line 113: Worth printing this out to the console then too, IMO.

docker/lite/Dockerfile.ubi8.arm64.mysql80
```diff
@@ -0,0 +1,108 @@
...
+
+# Create mount point for actual data (e.g. MySQL data dir)
+VOLUME /vt/vtdataroot
+USER vitess
+
+LABEL name="Vitess Lite image - MySQL Community Server 8.0" \
+      io.k8s.display-name="Vitess Lite image - MySQL Community Server 8.0" \
+      maintainer="cncf-vitess-maintainers@lists.cncf.io" \
+      vendor="CNCF" \
+      version="13.0.0" \
```
mattlord, line 102: This is the wrong version for main. I'd just remove it as I don't see where we have that in the other images. Or if we really need it, something like:
```
$ grep versionName go/vt/servenv/version.go | awk '{print $4}' | tr -d '"'
14.0.0-SNAPSHOT
```

mattlord (approved):
Thanks! I had a few questions but nothing that should block us. We can iterate from here.

PR #9831, opened 2022-03-05
[do not merge] testing e2e tests with flag deprecation shims

ajm188:


ajm188:
yayayayay, everything failing now is a flake. will clean up and push to the original PR branch

ajm188:
okay, i lied. the problem with the upgrade/downgrade tests is that the old vtctl parsing does not skip over `--` in `flag.Args()` (which i added in this branch), so we fail some tests with:

```
E0306 16:53:15.052277   26373 vtctl.go:170] action failed: AddCellInfo the <cell> argument is required for the AddCellInfo command
[485](https://github.com/vitessio/vitess/runs/5439625758?check_suite_focus=true#step:13:485)
```

options:
1. skip upgrade/downgrade tests for this branch
2. ship _just_ the `_flag.Args()` and `_flag.Arg(i)` "--" skip-shims in this PR, then do a second PR with the rest of the code here in a second one

PR #9832, opened 2022-03-06
Online DDL: vitess migrations cut-over to have zero race conditions

shlomi-noach:
## Description

#9755 introduced an improved query buffering logic for `vitess` migrations cut-over. This PR follows up by removing the last outstanding potential for race conditions where queries might "leak" during cut-over and where data loss can be exhibited.

The new `vitess` cut-over flow is:

1. Begin buffering queries
  - Any new query will be blocked
  - Existing queries can race condition this operation and proceed to execute
2. Short sleep
  - Pay some `100ms` price by letting the vast majority of queries who have raced-conditioned the above, to proceed to execute on the migrated table. In most scenarios, this wait time is enough to allow all queries to run
3. Audit/journal intention for next step (update `stowaway_table` migration's column)
4. `RENAME` the table away, under an unknown-to-the-app "stowaway" name
  - `RENAME` will wait for any remaining ongoing queries operating on the table
  - Once free to execute, it will block any further queries attempting to execute on the table
  - Once applied, the table is gone. Any remaining queries that attempt to operate on the table receive an error. No query operates on the renamed table (at the stowaway location) because no query knows the name of the table.
5. Capture MySQL position (`gtid_executed`)
6. VReplication to consume remaining binlogs up to captured position
7. Rename `_vrepl` _shadow_ table in place of original table
  - Table now exists "again". But there is still no traffic because all queries are buffered.
8. Unbuffer queries
  - queries no operate on new table
9. `RENAME` the table from stowaway location to `_vrepl` location. This is to allow a `REVERT` if need be
10. Remove the journal

The point of risk is where the table is renamed away, creating a "puncture' in the database.

- The cut-over logic will restore the stowaway table into original name on any failure
- In case the tablet itself fails during cut-over, the next tablet to run as Primary will identify the stowaway table journal entry and will restore the table into place (this takes place seconds after the new tablet process starts).

## Related Issue(s)

- https://github.com/vitessio/vitess/pull/9755
- #6926 

## Checklist


shlomi-noach:
Will squash merge this PR, seeing that there was a lot of back and forth `commit`s that make no sense independently.

go/vt/vttablet/onlineddl/executor.go
```diff
@@ -2826,6 +2889,23 @@ func (e *Executor) reviewRunningMigrations(ctx context.Context) (countRunnning i
...
+			// stowawayTable is an original table stowed away while cutting over a vrepl migration, see call to cutOverVReplMigration() down below in this function.
+			// In a normal operation, the table should not exist outside the scope of cutOverVReplMigration
+			// If it exists, that means a tablet crashed while running a cut-over, and left the database in a bad state, where the migrated table does not exist.
+			// thankfully, we have tracked this situation and just realized what happened. Now, first thing to do is to restore the original table.
+			log.Infof("found stowaway table %s in migration %s. Restoring it as %s", stowawayTable, uuid, onlineDDL.Table)
+			if _, err := e.renameTableIfApplicable(ctx, stowawayTable, onlineDDL.Table); err != nil {
+				// unable to restore table; we bail out, and we will try again next round.
+				return countRunnning, cancellable, err
+			}
+			// OK good, table restored. We can remove the record.
```
rohit-nayak-ps, line 2903: By design `renameTableIfApplicable()` doesn't return an err if the original or stowaway tables don't exist. So is it ok to expect that we have actually restored the table if it reaches here? If neither or both of the tables exist, we will not get an error even though the function didn't actually rename the tables. Possibly this is an impossible situation, but wanted to check.

go/vt/vttablet/onlineddl/executor.go
```diff
@@ -2275,8 +2313,33 @@ func (e *Executor) generateSwapTablesStatement(ctx context.Context, tableName1,
 	return parsed.Query, swapTableName, nil
 }
 
-func (e *Executor) executeAlterViewOnline(ctx context.Context, onlineDDL *schema.OnlineDDL) (err error) {
+// renameTableIfApplicable renames a table, assuming it exists and that the target does not exist.
+func (e *Executor) renameTableIfApplicable(ctx context.Context, fromTableName, toTableName string) (attemptMade bool, err error) {
```
rohit-nayak-ps, line 2317: I see that the `attemptMade` is discarded by the callers. Was it just added for debugging?

rohit-nayak-ps (commented):
I don't see any changes in the tests: I assume the existing tests and changes made in tests for the previous PR will cover these too?

go/vt/vttablet/onlineddl/executor.go
shlomi-noach: This is indeed intentional, which is why I named the function "ifApplicable". There is indeed a situation where this can happen:

Two lines below, after we've renamed the tables, we remove the `stowaway_table` value from the migration:

https://github.com/vitessio/vitess/pull/9832/files#diff-059c9f46e8d270d9c5514ef2b08679035eb0daaa8d95074e34ef43a81d50dc37R2904-R2906

However, if the tablet crashes just after renaming the table (back) and updating the fact that the table has been renamed -- then next time it runs -- or next tablet that runs -- sees a `stowaway_table` and thinks it still needs to do the rename. This is why I made the function idempotent. It OK to try it again and it will return with no error if there is nothing to rename.

rohit-nayak-ps (approved):
lgtm

shlomi-noach:
> lgtm

Hold on to your B*tts!

Issue #9827, opened 2022-03-03
Bug Report: Deadlock while shutting down a vreplication controller


mvh-stripe:
### Overview of the Issue

`vplayer.applyEvents` has a loop where it checks if it needs to throttle. If it needs to throttle, it sleeps for a short duration and keeps retrying.

```
    for {
		// check throttler.
		if !vp.vr.vre.throttlerClient.ThrottleCheckOKOrWait(ctx) {
			continue
		}
```
See https://github.com/vitessio/vitess/blob/main/go/vt/vttablet/tabletmanager/vreplication/vplayer.go#L339

The problem is it does not check if the context has been canceled. So when the vreplication controller is stopped, it deadlocks waiting for the vplayer to shut down.

### Reproduction Steps

```markdown
N/A
```

### Binary Version

```shell
N/A
```

### Operating System and Environment details

```shell
N/A
```

mattlord:
I'm closing this as we do now check for the context being cancelled there: https://github.com/vitessio/vitess/blob/main/go/vt/vttablet/tabletmanager/vreplication/vplayer.go#L335-L339

Please let me know if I'm missing or misunderstanding something and we can re-open it. Thank you, @mvh-stripe !

PR #9833, opened 2022-03-06
Check if context is cancelled in applyEvents() loop to avoid deadlocks

rohit-nayak-ps:
## Description

There is a code-path in `applyEvents()` in `vplayer.go` where it waits on a throttler check and sleeps if it is determined that a wait is needed. If the context is cancelled here then it can loop infinitely. The vreplication controller tries to shutdown due to the context failure, but it waits for all streams to end. This results in a deadlock.

This PR checks for a context cancellation in between throttle checks.

## Related Issue(s)
#9827 

## Checklist


shlomi-noach (approved):
LGTM

mattlord (approved):
LGTM!

PR #9834, opened 2022-03-06
Add parsing support for Trim grammar function

### LINKED ISSUE #4347: Add support for extended TRIM function grammar
#### Feature Description
We would like to utilize the TRIM grammar as defined in MySQL
https://dev.mysql.com/doc/refman/5.7/en/string-functions.html#function_trim

Currently, I can only use the basic version of TRIM and I couldn't find anything in the yacc file
https://github.com/vitessio/vitess/blob/a78587406b25f795429740f5426695681fc252ad/go/vt/sqlparser/sql.y

#### Use Case(s)

In relevance to implementing the Trim function in https://github.com/src-d/go-mysql-server we are using Vitess to parse the SQL rules.

K-Kumar-01:
## Description

Add parsing support for TRIM grammar function

#### Changes Made
- Support LTRIM(). Details [here](https://dev.mysql.com/doc/refman/5.7/en/string-functions.html#function_ltrim)
- Support RTRIM(). Details [here](https://dev.mysql.com/doc/refman/5.7/en/string-functions.html#function_rtrim)
- Support TRIM() partially. Details [here](https://dev.mysql.com/doc/refman/5.7/en/string-functions.html#function_trim)

For `TRIM`, trim types are left to be parsed.

## Related Issue(s)

Fixes #4347
#8604 

## Checklist


go/vt/sqlparser/sql.y
```diff
@@ -4509,6 +4513,19 @@ function_call_keyword
 	$$ = &BinaryExpr{Left: $1, Operator: JSONUnquoteExtractOp, Right: NewStrLiteral($3)}
   }
 
+trim_type_opt:
```
GuptaManan100, line 4516: No longer optional, you should drop `_opt`

GuptaManan100 (approved):

systay (approved):
👍 
nice stuff

PR #9840, opened 2022-03-08
sqlescape: Improve performance of EscapeID

mattrobenolt:
Using a strings.Builder over a bytes.Buffer yields one less allocation
per call due to the final .String() call requring an allocation with
bytes.Buffer, while strings.Builder is simply a pointer cast to a
string.

Also minor nits:

* Iterating over the string bytes, rather than runes is quite a bit
  faster, and we don't care about runes.

In general there are lots of places just glancing over the codebase that
are using bytes.Buffer in places that I'm pretty confident could/should
be converted over to strings.Builder instead. Constructing SQL strings,
etc. These would all likely benefit from a decent performance boost, but
I'm not entirely qualified to make these sweeping core changes in ways
that I can't easily test. But they should be reasonably easy to audit
and swap. They maintain the same API.

Benchmark against main:

```
goos: linux
goarch: amd64
pkg: vitess.io/vitess/go/sqlescape
cpu: 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz
BenchmarkEscapeID
BenchmarkEscapeID/aa
BenchmarkEscapeID/aa-8          13189527               122.3 ns/op            68 B/op        2 allocs/op
BenchmarkEscapeID/a`a
BenchmarkEscapeID/a`a-8          8941188               130.0 ns/op            72 B/op        2 allocs/op
BenchmarkEscapeID/long
BenchmarkEscapeID/long-8          659671              1661 ns/op             640 B/op        4 allocs/op
PASS
```

Benchmark against my change:

```
goos: linux
goarch: amd64
pkg: vitess.io/vitess/go/sqlescape
cpu: 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz
BenchmarkEscapeID
BenchmarkEscapeID/aa
BenchmarkEscapeID/aa-8          33748676                44.12 ns/op            8 B/op        1 allocs/op
BenchmarkEscapeID/a`a
BenchmarkEscapeID/a`a-8         26893474                48.27 ns/op            8 B/op        1 allocs/op
BenchmarkEscapeID/long
BenchmarkEscapeID/long-8         2717251               482.7 ns/op           160 B/op        1 allocs/op
PASS
```

As we can see, the difference is actually about 3-4x, and on long
strings, significantly better due to the pre-allocation.

### DIFF
--- a/go/sqlescape/ids.go
+++ b/go/sqlescape/ids.go
@@ -14,22 +14,29 @@ limitations under the License.
 package sqlescape
 
 import (
-	"bytes"
+	"strings"
 )
 
 // EscapeID returns a backticked identifier given an input string.
 func EscapeID(in string) string {
-	var buf bytes.Buffer
+	var buf strings.Builder
 	WriteEscapeID(&buf, in)
 	return buf.String()
 }
 
 // WriteEscapeID writes a backticked identifier from an input string into buf.
-func WriteEscapeID(buf *bytes.Buffer, in string) {
+func WriteEscapeID(buf *strings.Builder, in string) {
+	// growing by 4 more than the length, gives us room
+	// for guaranteed escaping with backticks on each end,
+	// plus a small amount of room just in case there are
+	// backticks within the symbol that needs to be double
+	// escaped. This is an unlikely edge case.
+	buf.Grow(4 + len(in))
+
 	buf.WriteByte('`')
-	for _, c := range in {
-		buf.WriteRune(c)
-		if c == '`' {
+	for i := 0; i < len(in); i++ {
+		buf.WriteByte(in[i])
+		if in[i] == '`' {
 			buf.WriteByte('`')
 		}
 	}
--- a/go/vt/worker/row_aggregator.go
+++ b/go/vt/worker/row_aggregator.go
@@ -17,7 +17,6 @@ limitations under the License.
 package worker
 
 import (
-	"bytes"
 	"fmt"
 	"strings"
 
@@ -52,7 +51,7 @@ type RowAggregator struct {
 	builder       QueryBuilder
 	statsCounters *stats.CountersWithSingleLabel
 
-	buffer       bytes.Buffer
+	buffer       strings.Builder
 	bufferedRows int
 }
 
@@ -141,13 +140,13 @@ func (ra *RowAggregator) Flush() error {
 // build the SQL query for one or more rows.
 type QueryBuilder interface {
 	// WriteHead writes the beginning of the query into the buffer.
-	WriteHead(*bytes.Buffer)
+	WriteHead(*strings.Builder)
 	// WriteTail writes any required tailing string into the buffer.
-	WriteTail(*bytes.Buffer)
+	WriteTail(*strings.Builder)
 	// Write the separator between two rows.
-	WriteSeparator(*bytes.Buffer)
+	WriteSeparator(*strings.Builder)
 	// Write the row itself.
-	WriteRow(*bytes.Buffer, []sqltypes.Value)
+	WriteRow(*strings.Builder, []sqltypes.Value)
 }
 
 // BaseQueryBuilder partially implements the QueryBuilder interface.
@@ -160,17 +159,17 @@ type BaseQueryBuilder struct {
 }
 
 // WriteHead implements the QueryBuilder interface.
-func (b *BaseQueryBuilder) WriteHead(buffer *bytes.Buffer) {
+func (b *BaseQueryBuilder) WriteHead(buffer *strings.Builder) {
 	buffer.WriteString(b.head)
 }
 
 // WriteTail implements the QueryBuilder interface.
-func (b *BaseQueryBuilder) WriteTail(buffer *bytes.Buffer) {
+func (b *BaseQueryBuilder) WriteTail(buffer *strings.Builder) {
 	buffer.WriteString(b.tail)
 }
 
 // WriteSeparator implements the QueryBuilder interface.
-func (b *BaseQueryBuilder) WriteSeparator(buffer *bytes.Buffer) {
+func (b *BaseQueryBuilder) WriteSeparator(buffer *strings.Builder) {
 	if b.separator == "" {
 		panic("BaseQueryBuilder.WriteSeparator(): separator not defined")
 	}
@@ -194,7 +193,7 @@ func NewInsertsQueryBuilder(dbName string, td *tabletmanagerdatapb.TableDefiniti
 }
 
 // WriteRow implements the QueryBuilder interface.
-func (*InsertsQueryBuilder) WriteRow(buffer *bytes.Buffer, row []sqltypes.Value) {
+func (*InsertsQueryBuilder) WriteRow(buffer *strings.Builder, row []sqltypes.Value) {
 	// Example: (0, 10, 'a'), (1, 11, 'b')
 	buffer.WriteByte('(')
 	for i, value := range row {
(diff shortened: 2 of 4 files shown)


go/vt/worker/row_aggregator.go
```diff
@@ -249,7 +249,7 @@ func (b *UpdatesQueryBuilder) WriteRow(buffer *bytes.Buffer, row []sqltypes.Valu
 		if i > 0 {
 			buffer.WriteByte(',')
 		}
-		sqlescape.WriteEscapeID(buffer, column)
+		sqlescape.WriteEscapeIDBytes(buffer, column)
```
mattrobenolt, line 252: These were a bit too deep in the callstack for me to safely and confidently convert all things to strings.Builder, but I'm pretty sure it can be fine, just wanted to not touch it for right now since it's much more invasive, so opted for a version that works with `*bytes.Buffer` still.

mattlord:
Nice! I’m still a golang n00b so I’m always learning interesting new things — in this case the testing lib’s Benchmark support. I love it.

I was operating under the assumption that when not dictated by the interface you would use a byte buffer when you want a byte slice in the end and a string builder when you want a string. But I saw byte buffers being used that violated this rule and I wasn’t sure if it was just old code or there were other considerations (highly concurrent envs can sometimes turn assumptions on their head). @vmg could probably comment on that.

mattrobenolt:
> I was operating under the assumption that when not dictated by the interface you would use a byte buffer when you want a byte slice in the end and a string builder when you want a string. But I saw byte buffers being used that violated this rule and I wasn’t sure if it was just old code or there were other considerations

Yup, this is generally true, but `strings.Builder` is much newer than `bytes.Buffer`. They are effectively the same thing, except as you mentioned, when you expect a `string` in the end, `strings.Builder` saves an allocation, whereas `bytes.Buffer` is better when you expect `[]byte` at the end. `strings.Builder` contains a nice trick to convert the underlying bytes to a string: https://github.com/golang/go/blob/master/src%2Fstrings%2Fbuilder.go#L47 :sunglasses: which avoids the extra allocation.

> highly concurrent envs can sometimes turn assumptions on their head

Neither `bytes.Buffer` or `strings.Builder` are safe for concurrent use, so we didn't change the assumption here.

go/sqlescape/ids.go
```diff
@@ -14,22 +14,33 @@ limitations under the License.
...
 
 import (
-	"bytes"
+	"strings"
 )
 
 // EscapeID returns a backticked identifier given an input string.
 func EscapeID(in string) string {
-	var buf bytes.Buffer
+	var buf strings.Builder
```
mattlord, line 22: I wonder if it wouldn’t be more efficient to just set the initial capacity to len(in)+4 here. That would cover the outer backticks and escaping of one inner. That should cover ~99.999999999999% of cases and we could skip the byte iteration and compare to count the backticks in the ident (I’ve never once seen a backtick used in a table name).

mattrobenolt: I can try that and benchmark it. I don't think the first loop to count adds anything really measurable to the result relative to the memory allocation. Typically in Go for high performance things, it's still favored to do extra loops to allocate once correctly. So the iteration _should_ be cheap, but I'll try it out for sure. You're likely right though. It's likely cheaper to allocate slightly more and cover most cases just fine. I assume as long as it's within the same slab class (or whatever Go's allocater pages are called), it'd be the same efficiency.

mattrobenolt: Ok, so interesting results.

before:

```
$ go test -v vitess.io/vitess/go/sqlescape -bench=. -benchmem
=== RUN   TestEscapeID
--- PASS: TestEscapeID (0.00s)
goos: linux
goarch: amd64
pkg: vitess.io/vitess/go/sqlescape
cpu: 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz
BenchmarkEscapeID
BenchmarkEscapeID/aa
BenchmarkEscapeID/aa-8          46028002                35.27 ns/op            4 B/op        1 allocs/op
BenchmarkEscapeID/a`a
BenchmarkEscapeID/a`a-8         20073259                53.98 ns/op            8 B/op        1 allocs/op
BenchmarkEscapeID/long
BenchmarkEscapeID/long-8         1790821               715.6 ns/op           144 B/op        1 allocs/op
PASS
```

after:

```
=== RUN   TestEscapeID
--- PASS: TestEscapeID (0.00s)
goos: linux
goarch: amd64
pkg: vitess.io/vitess/go/sqlescape
cpu: 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz
BenchmarkEscapeID
BenchmarkEscapeID/aa
BenchmarkEscapeID/aa-8          45377708                45.72 ns/op            8 B/op        1 allocs/op
BenchmarkEscapeID/a`a
BenchmarkEscapeID/a`a-8         24866187                48.87 ns/op            8 B/op        1 allocs/op
BenchmarkEscapeID/long
BenchmarkEscapeID/long-8         2398213               516.4 ns/op           160 B/op        1 allocs/op
PASS
```

With the diff: 

```diff
diff --git a/go/sqlescape/ids.go b/go/sqlescape/ids.go
index 9887ed2ccd..96ada5d2f2 100644
--- a/go/sqlescape/ids.go
+++ b/go/sqlescape/ids.go
@@ -26,19 +26,10 @@ func EscapeID(in string) string {
 
 // WriteEscapeID writes a backticked identifier from an input string into buf.
 func WriteEscapeID(buf *strings.Builder, in string) {
-	i := 0
-
-	// calculate the total length required after escaping
-	ll := 2 + len(in)
-	for i = 0; i < len(in); i++ {
-		if in[i] == '`' {
-			ll++
-		}
-	}
-	buf.Grow(ll)
+	buf.Grow(4 + len(in))
 
 	buf.WriteByte('`')
-	for i = 0; i < len(in); i++ {
+	for i := 0; i < len(in); i++ {
 		buf.WriteByte(in[i])
 		if in[i] == '`' {
 			buf.WriteByte('`')
```

So it's an objective improvement over _long_ identifiers because of obviously the extra O(N) loop.

We get still 1 allocation for a small identifier, but twice as large of an allocation (4B vs 8B) which makes sense since we're rounding up, which makes it decently slower, guessing because of the larger allocation.

So I guess we're at an improvement for long identifiers, but a loss for small identifiers, and likely a wash for some definition of average length.

Either way, I don't have a strong preference here, and I don't think it's worth adding in different logic to say `if len(in) > 20` or something and handle it differently.

But for posterity:

```
=== RUN   TestEscapeID
--- PASS: TestEscapeID (0.00s)
goos: linux
goarch: amd64
pkg: vitess.io/vitess/go/sqlescape
cpu: 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz
BenchmarkEscapeID
BenchmarkEscapeID/aa
BenchmarkEscapeID/aa-8          48326787                36.16 ns/op            4 B/op        1 allocs/op
BenchmarkEscapeID/a`a
BenchmarkEscapeID/a`a-8         36968911                53.68 ns/op            8 B/op        1 allocs/op
BenchmarkEscapeID/long
BenchmarkEscapeID/long-8         2654198               452.6 ns/op           160 B/op        1 allocs/op
PASS
```

Gets us kinda the best of both worlds with something like:

```diff
diff --git a/go/sqlescape/ids.go b/go/sqlescape/ids.go
index 9887ed2ccd..508d9daad2 100644
--- a/go/sqlescape/ids.go
+++ b/go/sqlescape/ids.go
@@ -26,19 +26,22 @@ func EscapeID(in string) string {
 
 // WriteEscapeID writes a backticked identifier from an input string into buf.
 func WriteEscapeID(buf *strings.Builder, in string) {
-	i := 0
-
-	// calculate the total length required after escaping
 	ll := 2 + len(in)
-	for i = 0; i < len(in); i++ {
-		if in[i] == '`' {
-			ll++
+
+	if len(in) < 20 {
+		for i := 0; i < len(in); i++ {
+			if in[i] == '`' {
+				ll++
+			}
 		}
+	} else {
+		ll += 2
 	}
+
 	buf.Grow(ll)
 
 	buf.WriteByte('`')
-	for i = 0; i < len(in); i++ {
+	for i := 0; i < len(in); i++ {
 		buf.WriteByte(in[i])
 		if in[i] == '`' {
 			buf.WriteByte('`')
```

I'm not sure the extra complexity for such a small gain is worth it though.

vmg (changes requested):
All these changes look fine. Thanks @mattrobenolt. That code was written when `strings.Builder` wasn't a thing, so porting small chunks of the codebase to it is a free win. Let's remove the extra loop in `Escape` before we merge: I think we should lean towards simpler heuristics.

go/sqlescape/ids.go
```diff
@@ -14,22 +14,33 @@ limitations under the License.
...
+func WriteEscapeID(buf *strings.Builder, in string) {
+	i := 0
+
+	// calculate the total length required after escaping
+	ll := 2 + len(in)
+	for i = 0; i < len(in); i++ {
+		if in[i] == '`' {
+			ll++
+		}
+	}
```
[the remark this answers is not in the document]
mattrobenolt: Done. That was my assumption as well, seemed a tad overkill relative to the other gains.

vmg (approved):
Thanks @mattrobenolt!

PR #9841, opened 2022-03-08
sqlescape: Fix UnescapeID

mattrobenolt:
UnescapeID previously was not unescaping already escaped double
backticks.

Added new tests and cases that would have previously failed.

### DIFF
--- a/go/sqlescape/ids.go
+++ b/go/sqlescape/ids.go
@@ -15,6 +15,7 @@ package sqlescape
 
 import (
 	"bytes"
+	"strings"
 )
 
 // EscapeID returns a backticked identifier given an input string.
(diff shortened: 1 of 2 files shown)


go/sqlescape/ids.go
```diff
@@ -48,8 +49,48 @@ func EscapeIDs(identifiers []string) []string {
...
-	if l >= 2 && in[0] == '`' && in[l-1] == '`' {
-		return in[1 : l-1]
+
+	// can't be escaped if only 2 chars
+	if l < 2 {
+		return in
+	}
+
+	// not escaped if the first and last chars aren't backticks
+	if in[0] != '`' || in[l-1] != '`' {
```
mattlord, line 59: This should be &&.

mattrobenolt: Changed, and added an additional test that covers this.

Issue #9838, opened 2022-03-07
Bug Report: infinite loop in tmState.ChangeTabletType


mvh-stripe:
### Overview of the Issue

```
		// Update the tablet record first.
		_, err := topotools.ChangeType(ctx, ts.tm.TopoServer, ts.tm.tabletAlias, tabletType, PrimaryTermStartTime)
		if err != nil {
			log.Errorf("error in changing type in tablet record in topo for tablet %s :- %v", topoproto.TabletAliasString(ts.tm.tabletAlias), err)
			log.Errorf("will keep trying to read from the topo server")
			// In case of a topo error, we aren't sure if the data has been written or not.
			// We must read the data again and verify whether the previous write succeeded or not.
			// The only way to guarantee safety is to keep retrying read until we succeed
			for {
				ti, errInReading := ts.tm.TopoServer.GetTablet(ctx, ts.tm.tabletAlias)
				if errInReading != nil {
					time.Sleep(100 * time.Millisecond)
					continue
				}
                                 ....
                         }
```

https://github.com/vitessio/vitess/commit/8b3f61fffce0de7114dd53b2396fa64d79755106
^ A recent change introduced the possibility of an infinite loop when the context gets canceled. In the for loop, we never check if the context has an error, and so it keeps trying to call `topoServer.GetTablet` every 100ms and never exit. This is bad because this goroutine also holds the mutex for the entire duration. So retries of the `ChangeTabletType` rpc also get stuck on the mutex.

### Reproduction Steps

```markdown
N/A
```

### Binary Version

```shell
N/A
```

### Operating System and Environment details

```shell
N/A
```

GuptaManan100:
Fixed by #9842

PR #9842, opened 2022-03-08
TabletManager: Check for context cancellation in loop within ChangeTabletType()

rohit-nayak-ps:
## Description
There is currently the possibility of an infinite loop inside `ChangeTabletType()` because we don't check for context cancellation inside an active polling loop. This PR adds the check.

## Related Issue(s)
#9838

## Checklist

### DIFF
--- a/go/vt/vttablet/tabletmanager/tm_state.go
+++ b/go/vt/vttablet/tabletmanager/tm_state.go
@@ -186,6 +186,9 @@ func (ts *tmState) ChangeTabletType(ctx context.Context, tabletType topodatapb.T
 			// We must read the data again and verify whether the previous write succeeded or not.
 			// The only way to guarantee safety is to keep retrying read until we succeed
 			for {
+				if ctx.Err() != nil {
+					return ctx.Err()
+				}
 				ti, errInReading := ts.tm.TopoServer.GetTablet(ctx, ts.tm.tabletAlias)
 				if errInReading != nil {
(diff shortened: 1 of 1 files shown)


go/vt/vttablet/tabletmanager/tm_state.go
```diff
@@ -186,6 +186,9 @@ func (ts *tmState) ChangeTabletType(ctx context.Context, tabletType topodatapb.T
 			// We must read the data again and verify whether the previous write succeeded or not.
 			// The only way to guarantee safety is to keep retrying read until we succeed
 			for {
+				if ctx.Err() != nil {
+					return ctx.Err()
```
GuptaManan100, line 190: We had added the infinite loop intentionally, because we enter that code only if a topo write has failed. Once a topo write fails, the only way forward is to keep trying to read the topo server, because the information may or may not have been written.
However, I do understand the use case where they want to retry doing the same operation again, so we need to exit sometime and not wait indefinitely.

I just think we should wrap the returned error and explicitly say that the topo information might be different from the tablet information and that the user must call ChangeTabletType again to fix it.
Other than that, this looks good to me

GuptaManan100 (approved):

go/vt/vttablet/tabletmanager/tm_state.go
GuptaManan100: Yes I think that will work!

PR #9848, opened 2022-03-09
add subquery support in update query

harshit-gangal:
## Checklist


harshit-gangal:
Use Gen4 framework to support this. https://github.com/vitessio/vitess/pull/9871

PR #9849, opened 2022-03-09
[topo/helpers] Make CopyShardReplications idempotent

### LINKED ISSUE #9844: topo2topo copy of shard replication is not idempotent
### Overview of the Issue

While using topo2topo to migrate topology data, we noticed that while the copying of all other topology data are idempotent, the copying of shard replication is not. Instead, it appends the tablets to the ShardReplication struct every time copy is run.

The code is doing a protobuf merge, which is probably simply appending:
https://github.com/vitessio/vitess/blob/main/go/vt/topo/helpers/copy.go#L172

### Reproduction Steps

1. Follow [this guide](https://github.com/aquarapid/vitess_examples/blob/master/topo2topo/README.md) to copy topology data , but run `topo2topo -do-shard-replications` twice
2. Run `topo2topo -do-shard-replications -compare` and the validation will fail.
3. Use topology specified CLI tool (e.g. etcdcli) to view the shard replication data

### Binary Version

```shell
n/a
```

```shell
n/a
```

### Log Fragments

```shell
n/a
```

ajm188:
## Description

This updates `helpers.CopyShardReplications` to build a set of tablet aliases in the fromTS topo so that repeated copies are idempotent.

## Related Issue(s)

Fixes #9844 

## Checklist

### DIFF
--- a/go/vt/topo/helpers/copy.go
+++ b/go/vt/topo/helpers/copy.go
@@ -22,7 +22,6 @@ import (
 	"context"
 
 	"google.golang.org/protobuf/proto"
-	"k8s.io/apimachinery/pkg/util/sets"
 
 	"vitess.io/vitess/go/vt/log"
 	"vitess.io/vitess/go/vt/topo"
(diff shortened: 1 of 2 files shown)


go/vt/topo/helpers/copy.go
```diff
@@ -168,8 +170,27 @@ func CopyShardReplications(ctx context.Context, fromTS, toTS *topo.Server) {
 					continue
 				}
 
+				sriNodes := sets.NewString()
```
vmg, line 173: This `k8s.io/apimachinery` dependency seems a bit out of place. We don't usually depend on these helpers in Vitess. Since all you need is a simple String Set, you can make one with a `map[string]struct{}` that would be just as readable/ergonomic.

vmg (changes requested):

vmg (approved):
Lookin gucci. Thanks!

PR #9352, opened 2021-12-10
Support new 8.0 functions

ritwizsinha:
Signed-off-by: ritwizsinha <ritwizsinha0@gmail.com>

### JSON
- [X] `JSON_TABLE(expr, path COLUMNS column_list) [AS] alias)`
- [X] Added the `->>` (inline path) operator, which is equivalent to calling `JSON_UNQUOTE()` on the result of `JSON_EXTRACT()`
- [X] `JSON_ARRAYAGG(col_or_expr)`
- [X] `JSON_OBJECTAGG(key, value)`
- [X] `JSON_PRETTY(json_val)`
- [X] `JSON_STORAGE_SIZE(json_val)`
- [X] `JSON_STORAGE_FREE(json_val)`
- [x] `JSON_MERGE_PATCH(json_doc, json_doc[, json_doc] ...)`
- [x] `JSON_MERGE()` function has been renamed `JSON_MERGE_PRESERVE()`. `JSON_MERGE()` continues to be recognized as an alias for `JSON_MERGE_PRESERVE()`
- [x] Added support in MySQL 8.0.2 for ranges such as $[1 to 5] in XPath expressions. Also added support in this version for the last keyword and relative addressing, such that $[last] always selects the last (highest-numbered) element in the array and $[last-1] the next to last element. last and expressions using it can also be included in range definitions; for example, $[last-2 to last-1] returns the last two elements but one from an array.

### Window
- [ ] `CUME_DIST() over_clause`
- [ ] `DENSE_RANK() over_clause`
- [ ] `FIRST_VALUE(expr) [null_treatment] over_clause`
- [ ] `LAG(expr [, N[, default]]) [null_treatment] over_clause`
- [ ] `LAST_VALUE(expr) [null_treatment] over_clause`
- [ ] `LEAD(expr [, N[, default]]) [null_treatment] over_clause`
- [ ] `NTH_VALUE(expr, N) [from_first_last] [null_treatment] over_clause`
- [ ] `NTILE(N) over_clause`
- [ ] `PERCENT_RANK() over_clause`
- [ ] `RANK() over_clause`
- [ ] `ROW_NUMBER() over_clause`

### Regular Expressions
- [ ] `REGEXP_INSTR(expr, pat[, pos[, occurrence[, return_option[, match_type]]]])`
- [ ] `REGEXP_LIKE(expr, pat[, match_type])`
- [ ] `REGEXP_REPLACE(expr, pat, repl[, pos[, occurrence[, match_type]]])`
[...]

Issue #4099, opened 2018-07-24
Support new 8.0 functions


derekperkins:
### JSON
- `JSON_TABLE(expr, path COLUMNS column_list) [AS] alias)`
- Added the `->>` (inline path) operator, which is equivalent to calling `JSON_UNQUOTE()` on the result of `JSON_EXTRACT()`
- `JSON_ARRAYAGG(col_or_expr)`
- `JSON_OBJECTAGG(key, value)`
- `JSON_PRETTY(json_val)`
- `JSON_STORAGE_SIZE(json_val)`
- `JSON_STORAGE_FREE(json_val)`
- `JSON_MERGE_PATCH(json_doc, json_doc[, json_doc] ...)`
- `JSON_MERGE()` function has been renamed `JSON_MERGE_PRESERVE()`. `JSON_MERGE()` continues to be recognized as an alias for `JSON_MERGE_PRESERVE()`
- Added support in MySQL 8.0.2 for ranges such as $[1 to 5] in XPath expressions. Also added support in this version for the last keyword and relative addressing, such that $[last] always selects the last (highest-numbered) element in the array and $[last-1] the next to last element. last and expressions using it can also be included in range definitions; for example, $[last-2 to last-1] returns the last two elements but one from an array.

### Window
- `CUME_DIST() over_clause`
- `DENSE_RANK() over_clause`
- `FIRST_VALUE(expr) [null_treatment] over_clause`
- `LAG(expr [, N[, default]]) [null_treatment] over_clause`
- `LAST_VALUE(expr) [null_treatment] over_clause`
- `LEAD(expr [, N[, default]]) [null_treatment] over_clause`
- `NTH_VALUE(expr, N) [from_first_last] [null_treatment] over_clause`
- `NTILE(N) over_clause`
- `PERCENT_RANK() over_clause`
- `RANK() over_clause`
- `ROW_NUMBER() over_clause`

### Regular Expressions
- `REGEXP_INSTR(expr, pat[, pos[, occurrence[, return_option[, match_type]]]])`
- `REGEXP_LIKE(expr, pat[, match_type])`
- `REGEXP_REPLACE(expr, pat, repl[, pos[, occurrence[, match_type]]])`
- `REGEXP_SUBSTR(expr, pat[, pos[, occurrence[, match_type]]])`

### WITH Syntax (Common Table Expressions)
- This feels like a bigger change rather than just adding function parsing support https://dev.mysql.com/doc/refman/8.0/en/with.html

[...]

PR #9850, opened 2022-03-09
Add parsing support for JSON Utility Functions

K-Kumar-01:
## Description

Adds the parsing support for JSON Utility functions. More information [here](https://dev.mysql.com/doc/refman/8.0/en/json-utility-functions.html)

#### Changes made
- Parsing of utility functions.
- Addition of test cases.

## Related Issue(s)

#8604 
#4099 

Related PR:  #9352 

## Checklist


go/vt/sqlparser/ast.go
```diff
@@ -2207,6 +2207,14 @@ type (
 		alternative  Expr // this is what will be used to Format this struct
 	}
 
+	// JSONUtilityExpr represents set of JSON utility functions
+	// More information on https://dev.mysql.com/doc/refman/8.0/en/json-utility-functions.html
+	JSONUtilityExpr struct {
+		Name      ColIdent
+		StringArg Expr
+		Column    *ColName
+	}
```
GuptaManan100, line 2216: The name should have been an enum. I would rather prefer having different structs for all the three functions, since they have very little in common other than being utility functions. Thoughts @systay @frouioui @harshit-gangal

go/vt/sqlparser/ast.go
```diff
@@ -2207,6 +2207,14 @@ type (
 		alternative  Expr // this is what will be used to Format this struct
 	}
 
+	// JSONUtilityExpr represents set of JSON utility functions
+	// More information on https://dev.mysql.com/doc/refman/8.0/en/json-utility-functions.html
+	JSONUtilityExpr struct {
+		Name      ColIdent
+		StringArg Expr
+		Column    *ColName
```
GuptaManan100, line 2215: I do not see the Column field being used in the `sql.y` file. Where is it being used?

go/vt/sqlparser/parse_test.go
```diff
@@ -2296,6 +2296,42 @@ var (
 	}, {
 		input:  "create table unused_reserved_keywords (dense_rank bigint, lead VARCHAR(255), percent_rank decimal(3, 0), row TINYINT, rows CHAR(10), constraint PK_project PRIMARY KEY (dense_rank))",
 		output: "create table unused_reserved_keywords (\n\t`dense_rank` bigint,\n\t`lead` VARCHAR(255),\n\t`percent_rank` decimal(3,0),\n\t`row` TINYINT,\n\t`rows` CHAR(10),\n\tconstraint PK_project PRIMARY KEY (`dense_rank`)\n)",
+	}, {
+		input:  "SELECT JSON_PRETTY('{\"a\":\"10\",\"b\":\"15\",\"x\":\"25\"}')",
+		output: "select json_pretty('{\\\"a\\\":\\\"10\\\",\\\"b\\\":\\\"15\\\",\\\"x\\\":\\\"25\\\"}') from dual",
+	}, {
+		input:  "SELECT JSON_PRETTY(N'{\"a\":\"10\",\"b\":\"15\",\"x\":\"25\"}')",
+		output: "select json_pretty(N'{\\\"a\\\":\\\"10\\\",\\\"b\\\":\\\"15\\\",\\\"x\\\":\\\"25\\\"}') from dual",
```
GuptaManan100, line 2304: We should instead use strings with starting with back tics (`) so that they are easier to read.

GuptaManan100 (changes requested):

go/vt/sqlparser/ast.go
K-Kumar-01: @GuptaManan100 
The structure for them is quite similar as all of them accept json_val as argument so it was created as one struct.
Info [here](https://dev.mysql.com/doc/refman/8.0/en/json-utility-functions.html)

go/vt/sqlparser/parse_test.go
```diff
@@ -2296,6 +2296,42 @@ var (
...
+		output: "select json_storage_size(`[100, \"sakila\", [1, 3, 5], 425.05]`) as A, json_storage_size(`{\"a\": 1000, \"b\": \"a\", \"c\": \"[1, 3, 5, 7]\"}`) as B, json_storage_size(`{\"a\": 1000, \"b\": \"wxyz\", \"c\": \"[1, 3, 5, 7]\"}`) as C, json_storage_size(`[100, \"json\", [[10, 20, 30], 3, 5], 425.05]`) as D from dual",
+	}, {
+		input:  "SELECT JSON_STORAGE_SIZE(@j)",
+		output: "select json_storage_size(@j) from dual",
+	}, {
+		input:  "SELECT JSON_STORAGE_FREE(jcol) FROM jtable",
+		output: "select json_storage_free(jcol) from jtable",
+	}, {
+		input:  "SELECT JSON_STORAGE_FREE(`{\"a\":\"10\",\"b\":\"15\",\"x\":\"25\"}`)",
+		output: "select json_storage_free(`{\"a\":\"10\",\"b\":\"15\",\"x\":\"25\"}`) from dual",
```
GuptaManan100, line 2328: No, not like this, this is changing the input of the test. We want to change the string in golang - 

```go
`SELECT JSON_STORAGE_FREE("{"a":"10","b":"15","x":"25"}")`
```

go/vt/sqlparser/ast.go
K-Kumar-01: Removed this.

go/vt/sqlparser/ast.go
```diff
@@ -2223,6 +2223,12 @@ type (
 		alternative  Expr // this is what will be used to Format this struct
 	}
 
+	// JSONUtilityExpr represents set of JSON utility functions
+	// More information on https://dev.mysql.com/doc/refman/8.0/en/json-utility-functions.html
+	JSONUtilityExpr struct {
+		Name      ColIdent
+		StringArg Expr
+	}
```
frouioui, line 2231: I think @GuptaManan100 is right. Even though the three different utility functions (`JSON_PRETTY`, `JSON_STORAGE_SIZE`, `JSON_STORAGE_FREE`) have the same signature, it would be more readable/debuggable to have them as separate AST nodes or have an identifiable name (such as an enum).

GuptaManan100 (approved):

PR #9853, opened 2022-03-09
Manage MySQL Replication Status States Properly

mattlord:
## Description
[We considered](https://github.com/vitessio/vitess/blob/6c20177eaff5e739d6eeee5b69b6112891e8bba6/go/mysql/flavor.go#L316-L321) the `Slave_IO_Running` state of `Connecting` as equivalent to `Running` in the [replication status](https://dev.mysql.com/doc/refman/en/replication-administration-status.html) results that we get from MySQL. I'm assuming this was done to avoid flapping on low traffic systems due to the [`-slave_net_timeout`](https://dev.mysql.com/doc/refman/en/replication-options-replica.html#sysvar_slave_net_timeout) reconnects and avoid doing a tablet repair when replication was healthy (with this PR we can distinguish running from healthy states and take differing actions depending on what we want).

After https://github.com/vitessio/vitess/pull/9308 we properly estimate the replica lag when MySQL is telling us that it does not know, meaning it returns a `NULL` value for the `Seconds_Behind_Master` field. But in some cases — e.g. it appears to happen when attempting the *first* connection to the source — MySQL will report a `Seconds_Behind_Master` value of `0` (meaning fully caught up and no lag) even when it is not connected to its replication source and has _failed to reconnect_ — meaning that this is not a simple reconnect for any reason (e.g. [`-slave_net_timeout`](https://dev.mysql.com/doc/refman/en/replication-options-replica.html#sysvar_slave_net_timeout)), but a reconnect with one or more _failures/errors_. This PR handles this latter case within Vitess by continuing to treat `Connecting` as equivalent to `Running` via `ReplicationStatus.Healthy()` — to prevent the noted flapping and errant/unnecessary tablet repairs — **_unless_**  we had an IO error the last time we tried to reconnect to the replication source. 

> Note: I think that [this](https://github.com/vitessio/vitess/pull/9789/files#r820941187) was also what in effect caused the bug seen in https://github.com/vitessio/vitess/issues/9788 as the new replica tablet was considered up with replication running when in fact it had not been able to connect to its source (for the very first time)

## Related Issue(s)
- Fixes: https://github.com/vitessio/vitess/issues/9788
- Replaces: https://github.com/vitessio/vitess/pull/9789
- Related to: https://github.com/vitessio/vitess/pull/9308

## Checklist
- [x] Tests are not required (I don't think)


mattlord:
@deepthi, @GuptaManan100, and @shlomi-noach: I apologize for the review hassle but unfortunately creating a new PR was the only way forward ([see here](https://github.com/vitessio/vitess/pull/9789#issuecomment-1063080893) for details). In the process, I consolidated the changes since the last round of reviews into a single commit: https://github.com/vitessio/vitess/pull/9853/commits/65226ad1d77629fe29e852393a19250a9a910024

Thanks in advance! 🙇‍♂️

GuptaManan100 (approved):
I feel sad that you had to jump through so many hoops for getting this in 😞

PR #9860, opened 2022-03-10
Change logic to reach desired _vt schema on vreplication startup

rohit-nayak-ps:
## Description

https://github.com/vitessio/vitess/pull/9817 was calling `WithDDL.ExecIgnore()` which _does not_ run the initial WithDDL queries to create the `_vt` schema. We need  to use `WithDDL.Exec()` instead for that.  However after trying that we found that some tests were failing. This happens when a primary is coming up, semi-sync is enabled, but there are no other replicas. This causes the tablet to deadlock and never come up.

So we just add the "impossible" query from the previous PR to the `WithDDLInitialQueries` list so that we will always run all the `_vt` schema queries, when a workflow starts.

We will also need the change in #9861, to handle the code paths using `VReplicationExec`  and `VExec`. This PR was originally supposed to make that redundant (since we planned to create the `_vt` schema on tablet startup.

## Related Issue(s)
#9817
#9861

## Checklist

### DIFF
--- a/go/vt/binlog/binlogplayer/binlog_player.go
+++ b/go/vt/binlog/binlogplayer/binlog_player.go
@@ -30,6 +30,8 @@ import (
 	"sync"
 	"time"
 
+	"vitess.io/vitess/go/vt/withddl"
+
 	"google.golang.org/protobuf/proto"
 
 	"vitess.io/vitess/go/vt/vtgate/evalengine"
@@ -573,10 +575,13 @@ var AlterVReplicationTable = []string{
 // - are to be expected by the mock db client during tests, or
 // - trigger some of the above _vt.vreplication schema changes to take effect
 //   when the binlogplayer starts up
+// todo: cleanup here. QueryToTriggerWithDDL will be enough to ensure vreplication schema gets created/altered correctly.
+//   So do that explicitly and move queries required into the mock code.
 var WithDDLInitialQueries = []string{
 	"SELECT db_name FROM _vt.vreplication LIMIT 0",
 	"SELECT rows_copied FROM _vt.vreplication LIMIT 0",
 	"SELECT time_heartbeat FROM _vt.vreplication LIMIT 0",
+	withddl.QueryToTriggerWithDDL,
 }
 
 // VRSettings contains the settings of a vreplication table.
--- a/go/test/endtoend/vreplication/vreplication_test.go
+++ b/go/test/endtoend/vreplication/vreplication_test.go
@@ -234,20 +234,6 @@ func TestCellAliasVreplicationWorkflow(t *testing.T) {
 	shardCustomer(t, true, []*Cell{cell1, cell2}, "alias", false)
 }
 
-/// confirm that WithDDL runs on tablet start and creates the expected _vt.vreplication schema
-func ensureVTSchemaIsUpToDate(t *testing.T) {
-	tablets := vc.getVttabletsInKeyspace(t, defaultCell, "product", "primary")
-	require.Equal(t, 1, len(tablets))
-	for _, tablet := range tablets {
-		qr, err := tablet.QueryTabletWithDB("show create table vreplication", "_vt")
-		require.NoError(t, err)
-		require.Equal(t, int64(1), int64(len(qr.Rows)))
-		// verify that a recently added column exists. ideally this should be the most recent one, but
-		// there is no way for an e2e test to know which column was added last.
-		require.Contains(t, qr.Rows[0][1].ToString(), "time_heartbeat")
-	}
-}
-
 // testVStreamFrom confirms that the "vstream * from" endpoint is serving data
 func testVStreamFrom(t *testing.T, table string, expectedRowCount int) {
 	ctx := context.Background()
--- a/go/test/endtoend/vreplication/vreplication_test.go
+++ b/go/test/endtoend/vreplication/vreplication_test.go
@@ -226,11 +226,29 @@ func TestCellAliasVreplicationWorkflow(t *testing.T) {
 	vtgateConn = getConnection(t, vc.ClusterConfig.hostname, vc.ClusterConfig.vtgateMySQLPort)
 	defer vtgateConn.Close()
 	verifyClusterHealth(t, vc)
+
+	t.Run("ensure _vt.vreplication table is created", func(t *testing.T) {
+		ensureVTSchemaIsUpToDate(t)
+	})
 	insertInitialData(t)
 	t.Run("VStreamFrom", func(t *testing.T) {
 		testVStreamFrom(t, "product", 2)
 	})
 	shardCustomer(t, true, []*Cell{cell1, cell2}, "alias", false)
+}
+
+/// confirm that WithDDL runs on tablet start and creates the expected _vt.vreplication schema
+func ensureVTSchemaIsUpToDate(t *testing.T) {
+	tablets := vc.getVttabletsInKeyspace(t, defaultCell, "product", "primary")
+	require.Equal(t, 1, len(tablets))
+	for _, tablet := range tablets {
+		qr, err := tablet.QueryTabletWithDB("show create table vreplication", "_vt")
+		require.NoError(t, err)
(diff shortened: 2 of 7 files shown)


rohit-nayak-ps:
While unit tests and vreplication tests pass, there are some tests around PRS like backups and vtworker end-to-end tests that are failing locally and in CI. The problem seems to be related to the fact that we are creating the `_vt` database as part of the tablet startup. During PRS the tablet designated to be primary does not come up claiming that replication is unhealthy. Not sure if this is because we are doing DMLs during  startup or because the WithDDL mechanism results in a failing query (which is ignored).

One test that fails is TestBackupMain in go/test/endtoend/backup/vtctlbackup. @GuptaManan100,  any ideas ?

GuptaManan100:
@rohit-nayak-ps No I haven't seen this type of a failure before in PRS. I will take a look and see if I find anything.

GuptaManan100:
@rohit-nayak-ps I looked at the test failures today and it seemed like ERS was always failing. Even the `TestTrivialERS` was failing.

I did some debugging and have found the issue. The changes in the PR are to move from using `ExecIgnore` to `Exec` with withDDL. Earlier `ExecIgnore` would just run a `SELECT` command and then exit out with an empty results in case of schema errors, but in `Exec` it tries to run the DDL commands to fix the issue.

The problem is that the Primary gets stuck on the DDL commands. We open the VReplication engine on transitioning to a primary instance, by then the `PromoteReplica` rpc has already fixed the semi-sync settings. Here is the stack-trace
```
vitess.io/vitess/go/vt/withddl.(*WithDDL).Exec(0xc0003162a0, {0x364e8d8, 0xc000054148}, {0x30339d2, 0x44}, {0x2c16620?, 0xc0018204c8?}, {0x2c16620, 0xc0018204e0})
	vitess.io/vitess/go/vt/withddl/withddl.go:86 +0x385
vitess.io/vitess/go/vt/vttablet/tabletmanager/vreplication.(*Engine).ensureValidSchema(0x2fa1e59?, {0x364e8d8, 0xc000054148})
	vitess.io/vitess/go/vt/vttablet/tabletmanager/vreplication/engine.go:224 +0x185
vitess.io/vitess/go/vt/vttablet/tabletmanager/vreplication.(*Engine).openLocked(0xc00047cdc0, {0x364e8d8, 0xc000054148})
	vitess.io/vitess/go/vt/vttablet/tabletmanager/vreplication/engine.go:230 +0x2c
vitess.io/vitess/go/vt/vttablet/tabletmanager/vreplication.(*Engine).Open(0xc00047cdc0, {0x364e8d8, 0xc000054148})
	vitess.io/vitess/go/vt/vttablet/tabletmanager/vreplication/engine.go:206 +0x145
vitess.io/vitess/go/vt/vttablet/tabletmanager.(*tmState).updateLocked(0xc000136c80, {0x364e948?, 0xc001a88750?})
	vitess.io/vitess/go/vt/vttablet/tabletmanager/tm_state.go:283 +0x38e
vitess.io/vitess/go/vt/vttablet/tabletmanager.(*tmState).ChangeTabletType(0xc000136c80, {0x364e948, 0xc001a88750}, 0x1, 0x1)
	vitess.io/vitess/go/vt/vttablet/tabletmanager/tm_state.go:224 +0x46c
vitess.io/vitess/go/vt/vttablet/tabletmanager.(*TabletManager).changeTypeLocked(0xc000436000, {0x364e948?, 0xc001a88750?}, 0x6aa0ee8?, 0x30?, 0x52b5f18?)
	vitess.io/vitess/go/vt/vttablet/tabletmanager/rpc_actions.go:97 +0x9d
vitess.io/vitess/go/vt/vttablet/tabletmanager.(*TabletManager).PromoteReplica(0xc000436000, {0x364e948, 0xc001a88750}, 0x1)
	vitess.io/vitess/go/vt/vttablet/tabletmanager/rpc_replication.go:861 +0x1c5
vitess.io/vitess/go/vt/vttablet/grpctmserver.(*server).PromoteReplica(0xc0005c0540, {0x364e948, 0xc001a886f0}, 0xc001a88720)
```
Because none of the other tablets are reparented yet, the call never succeeds until it times out and makes ERS fail

go/vt/binlog/binlogplayer/binlog_player.go
```diff
@@ -573,10 +575,13 @@ var AlterVReplicationTable = []string{
 // - are to be expected by the mock db client during tests, or
 // - trigger some of the above _vt.vreplication schema changes to take effect
 //   when the binlogplayer starts up
+// todo: cleanup here. QueryToTriggerWithDDL will be enough to ensure vreplication schema gets created/altered correctly.
+//   So do that explicitly and move queries required into the mock code.
 var WithDDLInitialQueries = []string{
 	"SELECT db_name FROM _vt.vreplication LIMIT 0",
 	"SELECT rows_copied FROM _vt.vreplication LIMIT 0",
 	"SELECT time_heartbeat FROM _vt.vreplication LIMIT 0",
+	withddl.QueryToTriggerWithDDL,
```
mattlord, line 584: Nit, but this naming feels really circular :-D

mattlord (commented):
Do we still want to move forward with this? At this point I feel like it might be worth stepping back and thinking about how we want to do this "the right way" in main.

mattlord (approved):

shlomi-noach (approved):

mattlord:
I *think* this may be leading to some [test flakiness](https://github.com/vitessio/vitess/runs/5857439070?check_suite_focus=true):
```
2022-04-06T20:06:14.5354485Z === RUN   TestV2WorkflowsAcrossDBVersions/reshardCustomer3to2SplitMerge/reshard/vdiff
2022-04-06T20:06:14.5355378Z I0406 20:05:43.072311   19509 vtctlclient_process.go:202] Executing vtctlclient with command: vtctlclient --server 127.0.0.1:15999 VDiff -- --tablet_types=primary --source_cell= --format json customer.c4c3
2022-04-06T20:06:14.5357412Z I0406 20:05:44.314820   19509 vreplication_test.go:787] vdiff err: exit status 1, output: VDiff Error: rpc error: code = Unknown desc = syncTargets: VReplicationWaitForPos for tablet zone1-0000001000: rpc error: code = Unknown desc = TabletManager.VReplicationWaitForPos on zone1-0000001000 error: Deadlock found when trying to get lock; try restarting transaction (errno 1213) (sqlstate 40001) during query: select pos, state, message from _vt.vreplication where id=1: Deadlock found when trying to get lock; try restarting transaction (errno 1213) (sqlstate 40001) during query: select pos, state, message from _vt.vreplication where id=1
2022-04-06T20:06:14.5359909Z E0406 20:05:44.311760   30457 main.go:103] remote error: rpc error: code = Unknown desc = syncTargets: VReplicationWaitForPos for tablet zone1-0000001000: rpc error: code = Unknown desc = TabletManager.VReplicationWaitForPos on zone1-0000001000 error: Deadlock found when trying to get lock; try restarting transaction (errno 1213) (sqlstate 40001) during query: select pos, state, message from _vt.vreplication where id=1: Deadlock found when trying to get lock; try restarting transaction (errno 1213) (sqlstate 40001) during query: select pos, state, message from _vt.vreplication where id=1
2022-04-06T20:06:14.5360907Z     vreplication_test.go:788: 
2022-04-06T20:06:14.5361360Z         	Error Trace:	vreplication_test.go:788
2022-04-06T20:06:14.5362225Z         	Error:      	Expected nil, but got: &exec.ExitError{ProcessState:(*os.ProcessState)(0xc00075a6c0), Stderr:[]uint8(nil)}
2022-04-06T20:06:14.5362982Z         	Test:       	TestV2WorkflowsAcrossDBVersions/reshardCustomer3to2SplitMerge/reshard/vdiff
```

I never noticed this before, but it seems to be causing frequent failures in the V2 workflow end to end tests. I'll investigate. It may also be triggered by GitHub actions being slower than usual and replication changes still flowing:
```
2022-04-06T20:06:14.5353975Z I0406 20:05:43.071412   19509 vreplication_test.go:694] Not waiting for vrepl to catch up on -40 since it is NOT a target shard
```

PR #9861, opened 2022-03-10
OnlineDDL executor: route all VReplicationExec through single function

shlomi-noach:
## Description
This PR just unifies all places where `onlineddl.Executor` calls `VReplicationExec()`. From now on, all these calls go through `e.vreplicationExec()` where:

- A `Once.Do()` ensures to init `_vt.vreplication`
- In the future, consolidate creation of `tmclient` and `tablet`.

This works around remaining issue described in #9817 and which is generically solved in #9860; but this PR has its own merits. Once #9860 is merged and good, then we can consider removing the `Once.Do()` aforementioned code.

## Related Issue(s)

- #6926 

## Checklist


rohit-nayak-ps (approved):
lgtm

PR #9863, opened 2022-03-10
feat: make the join columns output easier to read

systay:
## Description
Small PR that makes the plan descriptions containing join columns slightly easier to read.

Up until now, the list of columns that the join primitive produces has been shown using a list of integers, such as:
```
{
  "QueryType": "SELECT",
  "Original": "select user.name, music.genre from user join music on user.musicId = music.id",
  "Instructions": {
    "OperatorType": "Join",
    "Variant": "Join",
    "JoinColumnIndexes": "-2,1",
    "Inputs": [
      {
        "OperatorType": "Route",
        "Query": "select `user`.musicId, `user`.`name` from `user`",
      },
      {
        "OperatorType": "Route",
        "Query": "select music.genre from music where music.id = :user_musicId",
      }
    ]
  }
}
```

The interesting part here is **"JoinColumnIndexes": "-2,1"**.
Negative numbers come from the left, so -2 here means the second column coming from the left (`user.name`), and 1 means the first column from the right (`music.genre`).

My suggestion is that this field instead should show: **"JoinColumnIndexes": "L:1, R:0"**

## Checklist

### DIFF
--- a/go/vt/vtgate/engine/join.go
+++ b/go/vt/vtgate/engine/join.go
@@ -260,7 +260,7 @@ func combineVars(bv1, bv2 map[string]*querypb.BindVariable) map[string]*querypb.
 func (jn *Join) description() PrimitiveDescription {
 	other := map[string]interface{}{
 		"TableName":         jn.GetTableName(),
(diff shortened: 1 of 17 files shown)


frouioui (approved):
🚀

GuptaManan100 (approved):
Me likey

PR #9866, opened 2022-03-10
[VTAdmin] URL decode cookies

notfelineit:
## Description
This is a small bug fix - many times cookies will be URL encoded. We want to make sure the `cluster` cookie used for dynamic discovery is URL decoded before processing it as JSON.

It also makes a change to the Vitess logo - instead of importing the logo as an svg via relative path, it creates a VitessLogo component that inlines the svg:
Screen Shot 2022-03-10 at 11 47 43 AM

This makes vtadmin easier to import as an NPM package.

## Related Issue(s)
N/A - fixes a dynamic discovery bug: https://github.com/vitessio/vitess/pull/9544

## Checklist
- [x] Should this PR be backported? **No need.**
- [X] Documentation was added or is not required **Will be added with VTAdmin web page.**

## Deployment Notes
N/A


go/vt/vtadmin/api.go
```diff
@@ -236,33 +237,36 @@ func (api *API) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	clusterCookie, err := r.Cookie("cluster")
 
 	if err == nil {
-		decoded, err := base64.StdEncoding.DecodeString(clusterCookie.Value)
+		urlDecoded, err := url.QueryUnescape(clusterCookie.Value)
```
notfelineit, line 240: This is the only line added in this PR.

notfelineit:
@doeg absolutely! I'll add one!

doeg (approved):

doeg:
Merging this since the only failing tests are the upgrade downgrade tests, and this is marked with the "Skip Upgrade Downgrade" label.

PR #9867, opened 2022-03-11
docker: cleanup of release.sh

derekperkins:
This cleans up the release process for all the docker images inside /docker/k8s

- upgrades to v13 + defaults the base image to bullseye (discussed with @mattlord [here](https://github.com/vitessio/vitess/pull/9561#issuecomment-1024837539))
- explicitly build amd64 images (I accidentally pushed arm64 images from a new Mac)
- change paths to run in place - prior to this, it required to be executed in the /docker/k8s directory as ../release.sh. That was just an artifact after it moved from the since deleted /helm directory

### DIFF
--- a/docker/release.sh
+++ b/docker/release.sh
@@ -11,57 +11,57 @@ for debian_version in $debian_versions
 do
   echo "####### Building vitess/vt:$debian_version"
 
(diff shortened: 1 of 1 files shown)


mattlord (approved):
Thanks! I like the explicit platform flags, especially with coming ARM support in https://github.com/vitessio/vitess/pull/9830 🙂 (will review that one tomorrow)

derekperkins:
That force push was adding the explicit platform flag to the initial `docker pull` command. I'll merge once tests pass

Issue #9546, opened 2022-01-20
VReplication skipping binlog events when ENUM column used in PK


jakon89:
#### Overview of the Issue

First of all - this is a LONG issue - I tried to put here all details I found during debugging.

There is a `MoveTable` workflow `B` -> `A` (in fact this is MoveTables workflow for which I SwitchedTraffic a few days ago)
`B` -> shared keyspace, 2 shards
`A` -> unsharded keyspace, vttablet connects to MySQL over TCP

It looks that a single binlog event (`DELETE` in that case) was skipped when applying events from `B` to `A` thus causing workflow to hang.

step by step:
1. vplayer on `A` applies events from `B` with GTIDs `5435504` to `5435630` and updates status:
```
..lots of events applied in the current transaction, redacted... 

### UPDATE `_vt`.`vreplication`
### WHERE
###   @2='copy_reverse' /* VARSTRING(1000) meta=1000 nullable=1 is_null=0 */
###   @4='MySQL56/2e878e5f-e58d-11eb-a22a-42010a1d9536:1-9,ff18daa5-e58c-11eb-844d-42010a1d946d:1-5435504' /* VARSTRING(10000) meta=10000 nullable=0 is_null=0 */
###   @12='Running' /* VARSTRING(100) meta=100 nullable=0 is_null=0 */
### SET
###   @2='copy_reverse' /* VARSTRING(1000) meta=1000 nullable=1 is_null=0 */
###   @4='MySQL56/2e878e5f-e58d-11eb-a22a-42010a1d9536:1-9,ff18daa5-e58c-11eb-844d-42010a1d946d:1-5435630' /* VARSTRING(10000) meta=10000 nullable=0 is_null=0 */
# at 62295112
#220120 14:55:08 server id 1343706279  end_log_pos 62295143 CRC32 0x659a87d8 	Xid = 267258843
COMMIT/*!*/;
```

-----issue stars here-----

2. `B` primary generates new binlog events with GTIDs:
* * DELETE `ff18daa5-e58c-11eb-844d-42010a1d946d:5435631`
* * UPDATE `ff18daa5-e58c-11eb-844d-42010a1d946d:5435632` - not really important here, lets focus on events `5435631` and `5435633`
* * INSERT `ff18daa5-e58c-11eb-844d-42010a1d946d:5435633`

3. In the meantime`A` fetches above events and tries to apply **but fails**. `A`'s binlog: 
```
BEGIN
/*!*/;
# at 62295288
[...]

PR #9868, opened 2022-03-11
VPlayer: use stored/binlogged ENUM index value in WHERE clauses

mattlord:
## Description
With MySQL [ENUM types](https://dev.mysql.com/doc/refman/en/enum.html) you define a list of valid values for the column as a list of strings. What MySQL stores in the table[space] and writes to row based binary log events, however, is [the index](https://dev.mysql.com/doc/refman/en/enum.html#enum-indexes) value.

In VReplication's `VPlayer` component — which runs on the target tablet and applies row based binlog events received from the `VStreamer` component on the source tablet — we generate `DELETE` and `UPDATE` statements with a `WHERE` clause that contains the `PRIMARY KEY` columns (or other unique equivalents) so that statement applies to a unique row. When the `PRIMARY KEY` contains an `ENUM` field then that field will always be used in the `WHERE` clause. But because the _**index value**_ is binary logged for the field, e.g. (field 3):
```
## Table structure
CREATE TABLE `jobs_subscriptions` (
  `job_id` varchar(36) NOT NULL,
  `subscription_id` varchar(36) NOT NULL,
  `type` enum('RFS_WEBHOOKS','ESS') NOT NULL,
  `creation_time` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
  PRIMARY KEY (`job_id`,`type`),
  KEY `subscription_id_idx` (`subscription_id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8;

## Decoded RBR event resulting from SQL statement deleting a row from the table
#220119 15:07:07 server id 2749805  end_log_pos 788514835 CRC32 0xbd937ca8 	Delete_rows: table id 451 flags: STMT_END_F
### DELETE FROM `vt_main`.`jobs_subscriptions`
### WHERE
###   @1='4ef79a1e-29bf-48a1-8033-4ec8e406e929' /* VARSTRING(108) meta=108 nullable=0 is_null=0 */
###   @2='a0c8d8db-2ea0-495a-98c7-68b5e7926aeb' /* VARSTRING(108) meta=108 nullable=0 is_null=0 */
###   @3=2 /* ENUM(1 byte) meta=63233 nullable=0 is_null=0 */
###   @4=1638266996 /* TIMESTAMP(0) meta=0 nullable=0 is_null=0 */
```

We then end up with a `WHERE` clause that does not match the row because the string '2' is not in the ENUM values list and will thus gets converted to **0**, which is the index for any string specified that's not in the `ENUM` definition:
```
delete from jobs_subscriptions where job_id='4ef79a1e-29bf-48a1-8033-4ec8e406e929' and type='2';
```

While the column type is an `ENUM` which is normally a quoted string value in the parser as we use the string mappings available in the defined schema, we need to treat it as an integer type here or do the index to string value mapping again. But given that doing the mapping here would be completely unnecessary overhead as we already have the index value which is what gets stored, we convert the internal value type to a uint64 and use it directly (2 in the example). 

## Related Issue(s)
- Fixes: https://github.com/vitessio/vitess/issues/9546

## Checklist
- [x] Should this PR be backported? ???
- [x] Tests were added

### DIFF
--- a/go/vt/vttablet/tabletmanager/vreplication/replicator_plan.go
+++ b/go/vt/vttablet/tabletmanager/vreplication/replicator_plan.go
@@ -318,7 +318,7 @@ func (tp *TablePlan) bindFieldVal(field *querypb.Field, val *sqltypes.Value) (*q
 		return sqltypes.StringBindVariable(valString), nil
 	}
 	if enumValues, ok := tp.EnumValuesMap[field.Name]; ok && !val.IsNull() {
-		// The fact that this fielkd has a EnumValuesMap entry, means we must
+		// The fact that this field has a EnumValuesMap entry, means we must
 		// use the enum's text value as opposed to the enum's numerical value.
 		// Once known use case is with Online DDL, when a column is converted from
 		// ENUM to a VARCHAR/TEXT.
@@ -326,8 +326,19 @@ func (tp *TablePlan) bindFieldVal(field *querypb.Field, val *sqltypes.Value) (*q
 		if !enumValueOK {
 			return nil, vterrors.Errorf(vtrpcpb.Code_INTERNAL, "Invalid enum value: %v for field %s", val, field.Name)
 		}
-		// get the enum text fir this val
+		// get the enum text for this val
 		return sqltypes.StringBindVariable(enumValue), nil
+	}
+	if field.Type == querypb.Type_ENUM {
+		// This is an ENUM w/o a values map, which means that we are most likely using
+		// the index value -- what is stored and binlogged vs. the list of strings
+		// defined in the table schema -- and we must use an int bindvar or we'll have
+		// invalid/incorrect predicates like WHERE enumcol='2'.
+		// This will be the case when applying binlog events.
+		enumIndexVal := sqltypes.MakeTrusted(querypb.Type_UINT64, val.Raw())
+		if enumIndex, err := enumIndexVal.ToUint64(); err == nil {
+			return sqltypes.Uint64BindVariable(enumIndex), nil
+		}
 	}
 	return sqltypes.ValueBindVariable(*val), nil
 }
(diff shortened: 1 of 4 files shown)


go/vt/vttablet/tabletmanager/vreplication/replicator_plan.go
```diff
@@ -318,17 +318,28 @@ func (tp *TablePlan) bindFieldVal(field *querypb.Field, val *sqltypes.Value) (*q
...
 		enumValue, enumValueOK := enumValues[val.ToString()]
 		if !enumValueOK {
 			return nil, vterrors.Errorf(vtrpcpb.Code_INTERNAL, "Invalid enum value: %v for field %s", val, field.Name)
 		}
-		// get the enum text fir this val
+		// get the enum text for this val
 		return sqltypes.StringBindVariable(enumValue), nil
 	}
+	if field.Type == querypb.Type_ENUM {
+		// This is an ENUM w/o a values map, which means that we are most likely using
```
rohit-nayak-ps, line 333: you say "most likely": is it possible that we don't get the index here but the value? In that case this would fail, because we would be dealing with strings and not ints, right?

mattlord: Yeah, I think it's possible. That's why if we cannot convert the value to a uint64 then we don't return an error and we let it do what it was doing before the PR, the fallback of `sqltypes.ValueBindVariable(*val)`:
```
                enumIndexVal := sqltypes.MakeTrusted(querypb.Type_UINT64, val.Raw())
 		if enumIndex, err := enumIndexVal.ToUint64(); err == nil {
 			return sqltypes.Uint64BindVariable(enumIndex), nil
 		}
 	}
 	return sqltypes.ValueBindVariable(*val), nil
```

rohit-nayak-ps (approved):
LGTM
Another edge case fixed :-)

jakon89:
Hi guys, any chances it can be backported to v13?

rohit-nayak-ps:
> Hi guys, any chances it can be backported to v13?

Good point, this is a good candidate for a backport. Will do so asap.

mattlord:
Hi @jakon89,

I'm backporting the work to the `release-13.0` branch in https://github.com/vitessio/vitess/pull/9938. Please note that this won't trigger another release (e.g. `13.0.2`), but will allow you to easily build new 13.0 binaries. 

Thanks again!

Matt

PR #9869, opened 2022-03-11
fix: planner panic on derived tables sorting in query builder 

harshit-gangal:
## Description

This PR fixes a panic on table sorting on the final plan query output. It was caused when a query contains a join inside a derived table.

## Checklist
- [X] Should this PR be backported?

### DIFF
--- a/go/vt/vtgate/planbuilder/operator_to_query.go
+++ b/go/vt/vtgate/planbuilder/operator_to_query.go
@@ -21,6 +21,8 @@ import (
 	"sort"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


go/vt/vtgate/planbuilder/testdata/schema_test.json
```diff
@@ -431,7 +431,10 @@
               "name": "predef3"
             }
           ]
-        }
+        },
```
systay, line 434: oops. forgot to remove

GuptaManan100 (approved):

PR #9874, opened 2022-03-11
feat: add weightstring for distinct

systay:
## Description
When the planner is missing schema information for tables, we might be missing collation information for columns. To be able to correctly hash these, we fall back on using `weightstring()` for the columns that are missing collation information.
This avoids the error "text type with an unknown/unsupported collation cannot be hashed" that would sometimes show up.

## Checklist


frouioui (dismissed):

go/vt/vtgate/planbuilder/testdata/union_cases.txt
```diff
@@ -1808,11 +1855,15 @@ Gen4 plan same as above
...
-      1
+      2
     ],
     "Inputs": [
       {
         "OperatorType": "Distinct",
+        "Collations": [
+          "1: binary"
+        ],
+        "ResultColumns": 1,
```
harshit-gangal, line 1866: I think only 1 column will be returned from the input. The Columns on the SimpleProjection contains 2nd index. 
The plan may be incorrect.

go/vt/vtgate/planbuilder/testdata/union_cases.txt
```diff
@@ -1658,6 +1701,10 @@ Gen4 plan same as above
   "Original": "select col from unsharded union select id from user union select col2 from unsharded union select col from user_extra",
   "Instructions": {
     "OperatorType": "Distinct",
+    "Collations": [
+      "1: binary"
+    ],
+    "ResultColumns": 1,
     "Inputs": [
```
harshit-gangal, line 1708: The looks like the planner knows about collation, then the query should not contain weight_string column

go/vt/vtgate/planbuilder/testdata/union_cases.txt
```diff
@@ -1520,6 +1559,10 @@ Gen4 error: nesting of unions at the right-hand side is not yet supported
...
       {
         "OperatorType": "Distinct",
+        "Collations": [
+          "1: binary"
+        ],
+        "ResultColumns": 1,
```
harshit-gangal, line 1565: The looks like the planner knows about collation, then the query should not contain weight_string column

go/vt/vtgate/planbuilder/testdata/union_cases.txt
```diff
@@ -1435,6 +1470,10 @@ Gen4 error: nesting of unions at the right-hand side is not yet supported
   "Original": "(select id from user union select id from music order by id) union select 1 from unsharded",
   "Instructions": {
     "OperatorType": "Distinct",
+    "Collations": [
+      "1: binary"
+    ],
+    "ResultColumns": 1,
```
harshit-gangal, line 1476: The looks like the planner knows about collation, then the query should not contain weight_string column

go/vt/vtgate/planbuilder/testdata/union_cases.txt
```diff
@@ -1409,6 +1440,10 @@ Gen4 error: nesting of unions at the right-hand side is not yet supported
     "Inputs": [
       {
         "OperatorType": "Distinct",
+        "Collations": [
+          "1: binary"
+        ],
+        "ResultColumns": 1,
```
harshit-gangal, line 1446: The looks like the planner knows about collation, then the query should not contain weight_string column

go/vt/vtgate/engine/distinct.go
```diff
@@ -28,37 +32,40 @@ var _ Primitive = (*Distinct)(nil)
...
+	Source    Primitive
+	CheckCols []CheckCol
+	Truncate  bool
 }
 
-type row = []sqltypes.Value
+type CheckCol struct {
+	Idx       int
+	Collation collations.ID
+}
```
harshit-gangal, line 43: should this not keep track of weight_string column ?

go/vt/vtgate/engine/distinct.go
```diff
@@ -105,32 +113,42 @@ func (pt *probeTable) hashCodeForRow(inputRow row) (evalengine.HashCode, error)
...
 		}
-		hashcode, err := evalengine.NullsafeHashcode(value, collation, value.Type())
+		col := inputRow[checkCol.Col]
+		hashcode, err := evalengine.NullsafeHashcode(col, checkCol.Collation, col.Type())
 		if err != nil {
-			return 0, err
+			if err != evalengine.UnsupportedCollationHashError || checkCol.WsCol == nil {
+				return 0, err
+			}
+			checkCol = checkCol.SwitchToWeighString()
```
systay, line 126: nit: typo - `SwitchToWeighString` -> `SwitchToWeightString`

go/vt/vtgate/engine/distinct.go
```diff
@@ -105,32 +113,42 @@ func (pt *probeTable) hashCodeForRow(inputRow row) (evalengine.HashCode, error)
...
+	for i, checkCol := range pt.checkCols {
+		cmp, err := evalengine.NullsafeCompare(a[i], b[i], checkCol.Collation)
 		if err != nil {
-			return false, err
+			_, isComparisonErr := err.(evalengine.UnsupportedComparisonError)
+			if !isComparisonErr || checkCol.WsCol == nil {
+				return false, err
+			}
+			checkCol = checkCol.SwitchToWeighString()
+			pt.checkCols[i] = checkCol
```
systay, line 147: because CheckCols are by-value and not references, this does not update the original plan. maybe we should change it so that the plan remembers that we need to use the weight_string

go/vt/vtgate/engine/distinct.go
```diff
@@ -228,26 +248,47 @@ func (d *Distinct) Inputs() []Primitive {
...
+	if d.Truncate {
+		other["ResultColumns"] = len(d.CheckCols)
 	}
 	return PrimitiveDescription{
 		Other:        other,
 		OperatorType: "Distinct",
 	}
 }
+
+func (cc CheckCol) SwitchToWeighString() CheckCol {
```
systay, line 270: needs comment

go/vt/vtgate/engine/distinct.go
```diff
@@ -228,26 +248,47 @@ func (d *Distinct) Inputs() []Primitive {
...
+		WsCol:     nil,
+		Collation: collations.CollationBinaryID,
+	}
+}
+
+func (cc CheckCol) String() string {
+	coll := collations.Local().LookupByID(cc.Collation)
+	var collation string
+	if coll == nil {
+		collation = ""
```
systay, line 282: this block is not needed

go/vt/vtgate/planbuilder/operator_transformers.go
```diff
@@ -272,12 +274,96 @@ func transformUnionPlan(ctx *plancontext.PlanningContext, op *physical.Union) (l
...
+
+func getCheckColsForUnion(ctx *plancontext.PlanningContext, result logicalPlan, colls []collations.ID) ([]engine.CheckCol, error) {
+	checkCols := make([]engine.CheckCol, 0, len(colls))
+	for i, coll := range colls {
+		checkCol := engine.CheckCol{Col: i, Collation: coll}
+		if coll != collations.Unknown {
+			checkCols = append(checkCols, checkCol)
+			continue
+		}
+		// We might need a weight string - let's push one
```
systay, line 304: explain "might" better

go/vt/vtgate/planbuilder/testdata/union_cases.txt
```diff
@@ -1808,11 +1855,15 @@ Gen4 plan same as above
...
-      1
+      2
     ],
     "Inputs": [
       {
         "OperatorType": "Distinct",
+        "Collations": [
+          "(0:1)"
+        ],
+        "ResultColumns": 1,
```
systay, line 1866: This is wrong. SimpleProjection on top of this needs to use the literal 1 that has been pushed down here

frouioui: this is resolved via 9fd0332

go/vt/vtgate/engine/distinct.go
harshit-gangal: not necessary, as someone can change the type from varchar to int and changing the plan will make the query to not work. so, we should avoid modifying the plan

harshit-gangal: only local execution should remember it.

go/vt/vtgate/planbuilder/horizon_planning.go
```diff
@@ -97,9 +97,11 @@ func (hp *horizonPlanning) planHorizon(ctx *plancontext.PlanningContext, plan lo
...
-	if err != nil {
-		return nil, err
+	if truncateColumns {
+		plan, err = hp.truncateColumnsIfNeeded(ctx, plan)
+		if err != nil {
+			return nil, err
+		}
 	}
 
 	return plan, nil
```
harshit-gangal, line 107: nit: guard condition: if !truncateColumns { return plan, nil }

go/vt/vtgate/engine/concatenate.go
```diff
@@ -31,6 +31,24 @@ var _ Primitive = (*Concatenate)(nil)
...
 	Sources []Primitive
+
+	// These column offsets do not need to be typed checked - they usually contain weight_string()
+	// columns that are not going to be returned to the user
+	NoNeedToTypeCheck map[int]any
+}
+
+// NewConcatenate creates a Concatenate primitive. The ignoreCols slice contains the offsets that
+// don't need to have the same type between sources -
+// weight_string() sometimes returns VARBINARY and sometimes VARCHAR
```
GuptaManan100, line 42: TIL

go/vt/vtgate/engine/concatenate.go
```diff
@@ -101,22 +119,19 @@ func (c *Concatenate) TryExecute(vcursor VCursor, bindVars map[string]*querypb.B
...
-			continue
-		}
-		err := compareFields(fields, resFields)
-		if err != nil {
-			return nil, err
+		if r.Fields != nil {
+			err := c.compareFields(fields, r.Fields)
+			if err != nil {
+				return nil, err
+			}
```
GuptaManan100, line 131: The previous version of the code did not assume that the first result would always have a field set! The new code however assigns fields to `res[0].Fields` which might be nil. The next res could have a non empty field and then we will end up comparing nil and non-nil fields which will throw an error.
Previously this would have worked. Is this change in behaviour intentional or a side effect of the refactor which needs fixing?

go/vt/vtgate/engine/distinct.go
```diff
@@ -105,32 +113,42 @@ func (pt *probeTable) hashCodeForRow(inputRow row) (evalengine.HashCode, error)
...
-			return 0, err
+			if err != evalengine.UnsupportedCollationHashError || checkCol.WsCol == nil {
+				return 0, err
+			}
+			checkCol = checkCol.SwitchToWeightString()
+			pt.checkCols[i] = checkCol
+			hashcode, err = evalengine.NullsafeHashcode(inputRow[checkCol.Col], checkCol.Collation, col.Type())
+			if err != nil {
+				return 0, err
+			}
```
GuptaManan100, line 131: Should we not do this the other way around? Like check that `checkCol.WsCol` is not nil and use it. If it is nil, then we do the normal hash. In this case we won't have to check for the Unsupported collation error

go/vt/vtgate/evalengine/comparisons.go
```diff
@@ -146,7 +146,7 @@ func evalResultsAreDateAndNumeric(l, r *EvalResult) bool {
 func compareAsTuples(lVal, rVal *EvalResult) bool {
 	left := lVal.typeof() == sqltypes.Tuple
 	right := rVal.typeof() == sqltypes.Tuple
-	if left && right {
```
GuptaManan100, line 149: Is there a higher chance that the right side will be nil?

go/vt/vtgate/executor_select_test.go
```diff
@@ -842,12 +842,33 @@ func testRowCount(t *testing.T, executor *Executor, wantRowCount int64) {
...
+	require.NoError(t, err)
+	wantResult := &sqltypes.Result{
+		Fields: []*querypb.Field{
+			{Name: "id", Type: sqltypes.Int32},
+		},
+		Rows: [][]sqltypes.Value{{
+			sqltypes.NewInt32(52),
+		}},
+	}
+	utils.MustMatch(t, wantResult, got, "mismatch")
```
GuptaManan100, line 871: 😘

go/vt/vtgate/engine/distinct.go
systay: You are correct. I wrote a short unit test, and the plan seems to remember the new checkcol.

GuptaManan100 (approved):

go/vt/vtgate/engine/distinct.go
systay: Nah, if the column is numerical, we can't use the `weight_string` value and hashing works just fine. We could have checked the type here, but that is kind of what the evalengine is already doing, so I thought this would consolidate the type checking to fewer places

go/vt/vtgate/evalengine/comparisons.go
systay: 🤷 

don't know. why do you ask?

go/vt/vtgate/engine/concatenate.go
systay: Good point

go/vt/vtgate/evalengine/comparisons.go
systay: that was probably just my OCD

harshit-gangal (approved):

PR #9875, opened 2022-03-11
Examples: fix warnings

deepthi:
## Description

Fix warnings from examples using old flag syntax.
Along the way, found and fixed a panic from `vtctldclient ApplyVSchema`

## Related Issue(s)

#9733 

## Checklist

### DIFF
--- a/examples/region_sharding/scripts/vtctld-up.sh
+++ b/examples/region_sharding/scripts/vtctld-up.sh
@@ -6,7 +6,7 @@
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
 # 
-#     http://www.apache.org/licenses/LICENSE-2.0
+#     http://www.apache.org/licenses/LICENSE--2.0
 # 
 # Unless required by applicable law or agreed to in writing, software
 # distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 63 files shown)


examples/region_sharding/scripts/vtctld-up.sh
```diff
@@ -6,7 +6,7 @@
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
 # 
-#     http://www.apache.org/licenses/LICENSE-2.0
+#     http://www.apache.org/licenses/LICENSE--2.0
```
ajm188, line 9: Mistake?

go/cmd/vtctldclient/command/schema.go
```diff
@@ -109,7 +109,7 @@ func commandApplySchema(cmd *cobra.Command, args []string) error {
 	var allSQL string
 	if applySchemaOptions.SQLFile != "" {
 		if len(applySchemaOptions.SQL) != 0 {
-			return errors.New("Exactly one of --sql and --sql-file must be specified, not both.")
+			return errors.New("exactly one of --sql and --sql-file must be specified, not both")
```
ajm188, line 112: So I disagree with the linter here. This error text gets output directly to the user, so to me it's better to be a full sentence with capitalization and punctuation.

ajm188 (changes requested):
I think we should revert the error text change you made, otherwise all good on my end.

examples/region_sharding/scripts/vtctld-up.sh
deepthi: yep

go/cmd/vtctldclient/command/schema.go
deepthi: added `// nolint`

ajm188 (approved):

mattlord:
Thank you, @deepthi ! ❤️

PR #9876, opened 2022-03-12
[topo] ShardReplicationFix typed responses

ajm188:
## Description

This adds a protobuf data type to describe the problems being fixed by calls to `FixShardReplication` so that vtctldserver can surface this back to callers of the `ShardReplicationFix` RPC / vtctl command (which I am migrating in a PR to merge after this one).

## Checklist


proto/topodata.proto
```diff
@@ -333,6 +333,27 @@ message ShardReplication {
   repeated Node nodes = 1;
 }
 
+// ShardReplicationProblem describes the problem being fixed by
+// ShardReplicationFix.
+message ShardReplicationProblem {
```
rafael, line 338: nit with the name - It seems more intuitive to me to call this `ShardReplicationError`.  Problem is correct as well, but in my view it is an uncommon word in this context. We normally talk about error codes and errors. 

Was there a particular reason to name it `Problem`?

go/vt/topo/topoproto/replication.go
```diff
@@ -0,0 +1,39 @@
...
+	for k, v := range topodatapb.ShardReplicationProblem_Type_name {
+		shardReplicationProblemTypeName[k] = v
+	}
+}
+
+// ShardReplicationProblemTypeString returns a string representation of the
+// ShardReplicationProblem type, or "UNKNOWN" if not known.
+func ShardReplicationProblemTypeString(t topodatapb.ShardReplicationProblem_Type) string {
+	v, ok := shardReplicationProblemTypeName[int32(t)]
+	if !ok {
```
rafael, line 34: Under which scenario this could be `!ok`? It seems that would mean the caller is not passing a proper `ShardReplicationProblem_Type`. In that case, it seems like it should panic.

rafael (approved):
Added two tiny comments. Non blockers. This looks good to me

ajm188:
Tests are flakes that we've been seeing a lot lately

```
SIGQUIT: quit
[67](https://github.com/vitessio/vitess/runs/5524319577?check_suite_focus=true#step:9:67)
PC=0x47ffc8 m=0 sigcode=0
[68](https://github.com/vitessio/vitess/runs/5524319577?check_suite_focus=true#step:9:68)

[69](https://github.com/vitessio/vitess/runs/5524319577?check_suite_focus=true#step:9:69)
goroutine 13 [syscall, 10 minutes]:
[70](https://github.com/vitessio/vitess/runs/5524319577?check_suite_focus=true#step:9:70)
syscall.Syscall6(0xf7, 0x1, 0x48b1, 0xc000061e30, 0x1000004, 0x0, 0x0)
[71](https://github.com/vitessio/vitess/runs/5524319577?check_suite_focus=true#step:9:71)
	/opt/hostedtoolcache/go/1.17.7/x64/src/syscall/asm_linux_amd64.s:43 +0x5 fp=0xc000061de0 sp=0xc000061dd8 pc=0x47ffa5
[72](https://github.com/vitessio/vitess/runs/5524319577?check_suite_focus=true#step:9:72)
os.(*Process).blockUntilWaitable(0xc000046660)
[73](https://github.com/vitessio/vitess/runs/5524319577?check_suite_focus=true#step:9:73)
	/opt/hostedtoolcache/go/1.17.7/x64/src/os/wait_waitid.go:33 +0x9c fp=0xc000061ed0 sp=0xc000061de0 pc=0x4aed5c
[74](https://github.com/vitessio/vitess/runs/5524319577?check_suite_focus=true#step:9:74)
os.(*Process).wait(0xc000046660)
[75](https://github.com/vitessio/vitess/runs/5524319577?check_suite_focus=true#step:9:75)
	/opt/hostedtoolcache/go/1.17.7/x64/src/os/exec_unix.go:23 +0x28 fp=0xc000061f30 sp=0xc000061ed0 pc=0x4a80c8
[76](https://github.com/vitessio/vitess/runs/5524319577?check_suite_focus=true#step:9:76)
os.(*Process).Wait(...)
[77](https://github.com/vitessio/vitess/runs/5524319577?check_suite_focus=true#step:9:77)
	/opt/hostedtoolcache/go/1.17.7/x64/src/os/exec.go:132
[78](https://github.com/vitessio/vitess/runs/5524319577?check_suite_focus=true#step:9:78)
os/exec.(*Cmd).Wait(0xc0001ff600)
[79](https://github.com/vitessio/vitess/runs/5524319577?check_suite_focus=true#step:9:79)
	/opt/hostedtoolcache/go/1.17.7/x64/src/os/exec/exec.go:507 +0x54 fp=0xc000061fa8 sp=0xc000061f30 pc=0xbdff54
[80](https://github.com/vitessio/vitess/runs/5524319577?check_suite_focus=true#step:9:80)
vitess.io/vitess/go/test/endtoend/cluster.(*TopoProcess).SetupEtcd.func1()
[81](https://github.com/vitessio/vitess/runs/5524319577?check_suite_focus=true#step:9:81)
	/home/runner/work/vitess/vitess/go/test/endtoend/cluster/topo_process.go:101 +0x29 fp=0xc000061fe0 sp=0xc000061fa8 pc=0xdbddc9
[82](https://github.com/vitessio/vitess/runs/5524319577?check_suite_focus=true#step:9:82)
runtime.goexit()
```

I've kicked off a retry

go/vt/topo/topoproto/replication.go
ajm188: I just copied the implementations of `TabletTypeLString` and friends, which also went with the "never panic" approach.

I'm not sure its semantically different, either. If you don't pass a "proper" `ShardReplicationProblem_Type`, then the name is unknown to us.

proto/topodata.proto
ajm188: Good question! I don't have a particularly strong preference either way. The reason I went with "Problem" is that it's not _currently_ an error; it _used_ to be an error, but we just fixed it.

Separately (embarrassingly), I was going to say that it was because the `-help` for `vtctl ShardReplicationFix` read "and fixes the first _problem_ [ed. not 'error'] it encounters." I could have _sworn_ it did, but in fact it says "and fixes the first _error_ it encounters," so maybe I should go with "Error".

But, in either case, that was my thought process. Let me know what you think!

rafael: I see. I think I have taste preference for error and consistent with "and fixes the first error it encounters". I'm fine either way. You can take the final call on this

PR #9883, opened 2022-03-14
vstreamer: split packets between savepoint events

arthurschreiber:
## Description

On one of our clusters, `CreateLookupVindex` workflows have run into the following error:
```
vttablet: rpc error: code = ResourceExhausted desc = trying to send message larger than max (111926209 vs. 67108864)
```

We believe the root cause of this error is the same as https://github.com/vitessio/vitess/pull/9881 - vreplication is executing transactions containing many savepoints but no or only a small number of actual data changes.

`vstreamer` tries to send data to vplayer in packets. A packet either contains all binlog events that happened for a single transaction, or for large transactions, the event stream is split along row or statement events, once packet size grows bigger than `vstream_packet_size`.

It looks like savepoint events are ignored in the packet size calculation logic. If a single transaction contains many savepoints, that could lead to packets growing unusually large.

This pull request proposes to fix this by including the statement length of savepoints in the packet size calculation.

## Checklist

## Deployment Notes

None.


vmg (approved):
This function sure is overdue for some refactoring, as most of the blocks look the same. The change looks good to me though :)

PR #9892, opened 2022-03-15
vstreamer: flatten savepoint events

arthurschreiber:
## Description

This Pull Request implements flattening logic for savepoints, as described in https://github.com/vitessio/vitess/pull/9881. It also implements the logic to skip unused savepoint events as in https://github.com/vitessio/vitess/pull/9881, but on the vstreamer side instead of the vplayer side (thus causing less unused data being passed around).

I do think this flattening behaviour is sane, but would really appreciate someone else to also think this through carefully. 😬 

This essentially changes vstreamer to only emit the first `SAVEPOINT ...` event in a row of savepoint events, and re-writes `ROLLBACK TO ...` events to point to this "first" event. It now also skips any savepoint statements that are not followed by a row or dml event.

For example

```
begin;
insert into stream1 values (1, 'aaa');
savepoint a;
insert into stream1 values (2, 'aaa');
savepoint b;
savepoint c;
insert into stream1 values (3, 'aaa');
rollback to c;
savepoint d;
commit;
```

would be transformed into

```
begin;
insert into stream1 values (1, 'aaa');
savepoint a;
insert into stream1 values (2, 'aaa');
savepoint b;
insert into stream1 values (3, 'aaa');
rollback to b;
commit;
```

The improvement in this example might not seem like much, but as described in https://github.com/vitessio/vitess/pull/9881, this change greatly reduces the number of "useless" statements being executed for replication streams that filter for just a single table - in many cases this even allows completely skipping the transaction!

Testing the `ROLLBACK TO ...` behaviour is hard, because rollback statements can only land in the binlog of a MySQL instance that does not enforce GTID consistency and with a transaction that modifies a non-transactional table. If you have any suggestions on how to test this, I'm all ears! 🙇‍♂️ 

## Related Issue(s)

https://github.com/vitessio/vitess/pull/9881

## Checklist


arthurschreiber:
This also contains a variation of the fix proposed in https://github.com/vitessio/vitess/pull/9883 - savepoint statements are considered as part of the packet vstreamer packet length.

rohit-nayak-ps (approved):
lgtm

PR #9905, opened 2022-03-17
Improve network partitions handling

y5w:
## Description

This is an improvement on how we handle network partition. VTGR used to rely to "UNREACHABLE" state to determine network partition, it is not always reliable since when a node is really "partitioned", it will also possible to be unreachable. It introduced a race condition when the network is flapping: a partitioned VTGR will see everyone is OFFLINE and try to rebootstrap the group, if network recovered during the rebootstrap, VTGR will stop the functioning group incorrectly.

To address the issue, this PR mainly does two things:
1. Do not allow bootstrap unless we can reach expected number of nodes.
2. Do not allow rebootstrap unless we see all the nodes unless there is an override.

This helps VTGR to avoid some "false positive" diagnoses and hence avoid further incorrect repair operations.

## Related Issue(s)

^

## Checklist

### DIFF
--- a/go/vt/vtgr/controller/diagnose.go
+++ b/go/vt/vtgr/controller/diagnose.go
@@ -173,6 +173,9 @@ func (shard *GRShard) diagnoseLocked(ctx context.Context) (DiagnoseType, error) 
 	// if no, we should bootstrap one
 	mysqlGroup := shard.shardAgreedGroupName()
 	if mysqlGroup == "" {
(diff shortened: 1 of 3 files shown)


deepthi (approved):

deepthi:
Upgrade/downgrade query serving test was flaky when this PR was created. Merging in spite of those failures.

deepthi:
Re CI: please use a fork to create your PRs as documented in the contributing guide https://vitess.io/docs/14.0/contributing/github-workflow/

PR #9907, opened 2022-03-17
vstreamer: do not forward savepoint events

arthurschreiber:
## Description

In https://github.com/vitessio/vitess/pull/9881, I proposed a set of changes to flatten savepoint events send from vstreamer to vplayer to reduce the amount of data passed around (and written to the binlog) for filtered workflows (like `CreateLookupVindex`). While trying to write tests for this functionality, I noticed that my initial assumptions about savepoints in the MySQL binlog and how they're handled by vstreamer was not correct.

#### Regarding savepoints in the MySQL binlog

`SAVEPOINT ...` statements are written to the binlog by MySQL in all 3 replication modes, row based, statement based and mixed replication mode.

`ROLLBACK TO ...` is only written to the binlog by MySQL in a very specifc case, based on all the testing that I did. [`enforce_gtid_consistency`](https://dev.mysql.com/doc/refman/5.7/en/replication-options-gtids.html#sysvar_enforce_gtid_consistency) must be disabled (which implies that [`gtid_mode`](https://dev.mysql.com/doc/refman/5.7/en/replication-options-gtids.html#sysvar_gtid_mode) is also disabled), a savepoint must be created, a non-transactional table must be changed, and then a rollback must be made to the savepoint created earlier.

In all other cases, MySQL truncates the binlog so that all the statements executed (or row entries written in mixed or row based replication mode) since the savepoint creation actually never end up in the binlog.

#### Savepoints and VReplication

Now, in https://github.com/vitessio/vitess/pull/9881 I incorrectly assumed that both `SAVEPOINT ...` and `ROLLBACK TO ...` statements are parsed as `sqlparser.StmtSavepoint` when read from the binlog, and later end up as `VEventType_SAVEPOINT` events.

Instead `ROLLBACK TO ...` statements are parsed as `sqlparser.StmtSRollback`, which is not handled in vstreamer at all.

Putting this all together, we can say:
* The changes in https://github.com/vitessio/vitess/pull/9881 are incorrect but harmless and *do* improve vreplication performance with filtered replication workflows. 
* `SAVEPOINT ...` statements can be dropped in row based replication mode safely, because MySQL will never write `ROLLBACK TO ...` statements to the binlog in that mode.
* `ROLLBACK TO ...` statements are currently not encountered by Vitess, and if they are would cause vreplication to break. 

Based on this, I want to propose that we simply ignore all `sqlparser.StmtSavepoint`s.

If we want vitess to fully support unfiltered VReplication from a MySQL instance that uses mixed or statement based replication, we should pass through both `SAVEPOINT ...` as well as `ROLLBACK TO ...` statements. But this should only be done when we know that VReplication is executed without a filter (which implies it's used to simply copy a whole MySQL database/server). But this should probably be handled in a separate changeset with some refactoring of the vstreamer code.

## Related Issue(s)

https://github.com/vitessio/vitess/pull/9881

## Checklist

## Deployment Notes

None.


arthurschreiber:
@rohit-nayak-ps I'm sorry for the churn I might have caused with the changes https://github.com/vitessio/vitess/pull/9881. Would you mind taking a look over these changes here as well?

I'm also curious to understand to which extent `vstreamer` (and Vitess in general) are supposed to work with statement based replication, or if that support is not a priority at all.

Thanks! 🙇‍♂️

rohit-nayak-ps:
> @rohit-nayak-ps I'm sorry for the churn I might have caused with the changes in #9881. Would you mind taking a look over these changes here as well?

Deleting code is always good :-)   So I understand you correct, you are saying that TestPlayerSavepoint at https://github.com/vitessio/vitess/blob/a5bab573218c1d57c7052cb14e6c1707a236fdc9/go/vt/vttablet/tabletmanager/vreplication/vplayer_flaky_test.go#L393 is working because the rolled back statements are not streamed to the target at all.

Since, for RBR, binlogs are written only on commits we will never see intermediate rollbacks. 

> I'm also curious to understand to which extent `vstreamer` (and Vitess in general) are supposed to work with statement based replication, or if that support is not a priority at all.

Officially we only support RBR in VReplication.

mattlord (approved):
Nice work on this @arthurschreiber ! The effort put into this and of course the eventual contribution are very much appreciated. 

Since we only support transactional storage engines and row based replication in Vitess, I see no reason to do anything with SAVEPOINTs in vreplication as they are only binary logged for committed transactions (ROLLBACK is irrelevant and not binlogged) and are only binlogged at all because MySQL replication still has to generally support mixed storage engines (including non-transactional ones like MyISAM).

In our case it has no real usage and does not impact the final data or even intermediate states and what is externalized. For example (I rolled back to the mysv101 savepoint before committing):
```
# at 1660
#220323 13:06:06 server id 1  end_log_pos 1735 CRC32 0xe8804b84 	Query	thread_id=10	exec_time=0	error_code=0
SET TIMESTAMP=1648055166/*!*/;
BEGIN
/*!*/;
# at 1735
#220323 13:06:06 server id 1  end_log_pos 1783 CRC32 0x337060d7 	Table_map: `mydb`.`tt` mapped to number 164
# at 1783
#220323 13:06:06 server id 1  end_log_pos 1828 CRC32 0x939fc6c3 	Write_rows: table id 164 flags: STMT_END_F
### INSERT INTO `mydb`.`tt`
### SET
###   @1=201 /* INT meta=0 nullable=1 is_null=0 */
### INSERT INTO `mydb`.`tt`
### SET
###   @1=202 /* INT meta=0 nullable=1 is_null=0 */
# at 1828
#220323 13:06:10 server id 1  end_log_pos 1917 CRC32 0x7c235611 	Query	thread_id=10	exec_time=0	error_code=0
SET TIMESTAMP=1648055170/*!*/;
SAVEPOINT `mysv100`
/*!*/;
# at 1917
#220323 13:06:19 server id 1  end_log_pos 1965 CRC32 0xc8dfacfa 	Table_map: `mydb`.`tt` mapped to number 164
# at 1965
#220323 13:06:19 server id 1  end_log_pos 2010 CRC32 0xe9554d59 	Write_rows: table id 164 flags: STMT_END_F
### INSERT INTO `mydb`.`tt`
### SET
###   @1=301 /* INT meta=0 nullable=1 is_null=0 */
### INSERT INTO `mydb`.`tt`
### SET
###   @1=302 /* INT meta=0 nullable=1 is_null=0 */
# at 2010
#220323 13:06:22 server id 1  end_log_pos 2099 CRC32 0x0f2e1625 	Query	thread_id=10	exec_time=0	error_code=0
SET TIMESTAMP=1648055182/*!*/;
SAVEPOINT `mysv101`
/*!*/;
# at 2099
#220323 13:06:56 server id 1  end_log_pos 2130 CRC32 0xe1daef11 	Xid = 126
COMMIT/*!*/;
```

I'm not sure why we originally cared about SAVEPOINTs in the vstream but I see no clear reason to today so I think removing this overhead makes sense (especially moving forward as MySQL 8.0 does not use MyISAM at all for system tables etc).

mattlord:
If you rebase on main the region example test will be fixed too, just FYI.

deepthi:
I'm going to go ahead and merge this. The failing tests have (mostly) been fixed on main.

arthurschreiber:
@deepthi Thanks for merging this! 🎉

PR #9913, opened 2022-03-18
[VEP4, phase 1] Update flags and imports for `endtoend/{sharded,sharding}`

ajm188:
## Description

More more more!!

## Related Issue(s)

#9895 

## Checklist

### DIFF
--- a/go/test/endtoend/sharded/sharded_keyspace_test.go
+++ b/go/test/endtoend/sharded/sharded_keyspace_test.go
@@ -26,9 +26,8 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
-	"vitess.io/vitess/go/vt/log"
-
 	"vitess.io/vitess/go/test/endtoend/cluster"
+	"vitess.io/vitess/go/vt/log"
 )
 
 var (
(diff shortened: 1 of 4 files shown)


go/test/endtoend/sharded/sharded_keyspace_test.go
```diff
@@ -139,7 +138,7 @@ func TestShardedKeyspace(t *testing.T) {
 	_, _ = shard1Primary.VttabletProcess.QueryTablet("insert into vt_select_test (id, msg) values (1, 'test 1')", keyspaceName, true)
 	_, _ = shard2Primary.VttabletProcess.QueryTablet("insert into vt_select_test (id, msg) values (10, 'test 10')", keyspaceName, true)
 
-	err = clusterInstance.VtctlclientProcess.ExecuteCommand("Validate", "-ping-tablets")
+	err = clusterInstance.VtctlclientProcess.ExecuteCommand("Validate", "--", "--ping-tablets")
```
GuptaManan100, line 141: Why do we need the extra `--` before `--ping-tablets`

GuptaManan100 (approved):

go/test/endtoend/sharded/sharded_keyspace_test.go
ajm188: See [VEP-4](https://github.com/vitessio/enhancements/blob/main/veps/vep-4.md#phase-1-preparation):

> The second issue with dropping in `pflag` is that it supports mixing of positional and flag arguments.
> **This change will break subcommands, such as those in `vtctl` and `mysqlctl`.**
>
> To illustrate the problem, consider the command `./mycmd -a 1 -b 2 X -c 3 -d 4`.
> With `flag`, parsing this will result in two flags, `a` and `b`, `flag.NArg()` will return 5 and `flag.Args()` will return the slice `["X", "-c", "3", "-d", "4"]` ([c.f.](https://play.golang.org/p/sO4UlQPGg4l)).
With `pflag`, all 4 flags will be parsed, `pflag.NArg()` will return 1, and `pflag.Args()` will return the slice `["X"]`.
>
> In order to match this behavior, users should update their commands to insert a double-dash (`--`) before the first positional argument.
> This causes `pflag` to stop looking for flags and treat everything after as a positional argument (matching what `flag` does even without the double-dash), and has no effect on the parsing behavior of `flag` ([c.f.](https://play.golang.org/p/76FrFrLFOMT)).

PR #9914, opened 2022-03-18
1.18: rename interface{} to any

vmg:
## Description

Quoting the timeless words of Kant: "haha, oh fuck". This PR finishes the 1.18 migration by performing the same massive refactoring that was performed in `golang/go` before the 1.18 release:

https://github.com/golang/go/commit/2580d0e08d5e9f979b943758d3c49877fb2324cb

The reasons why we should perform this refactoring can be found in https://github.com/golang/go/issues/49884 and related issues, so I won't repeat them. Personally, I quite dig the new style and I think it makes a lot of code much easier to read. I also think this renaming will make @sougou upset and I find that funny, so that's a plus.

We will definitely need some consensus before merging this. The good news is that, even though this PR is quickly diverging, it's very easy to rebase because the renaming process is safe and fully automated behind a `go fmt` command.

Let me know your thoughts, team!

## Checklist


shlomi-noach (approved):
I do not have `any` strong opinions.

GuptaManan100 (approved):
I for one, think it is more readable

frouioui (approved):
LGTM

rohit-nayak-ps (approved):
lgtm

Issue #7362, opened 2021-01-24
Tracking issue: VReplication throttling


shlomi-noach:
We wish to introduce throttling in VReplication to avoid overwhelming the databases with reads/writes.

VReplication's current behavior is greedy: if it can read table data, it will read it and push downstream. If it can read binary logs, it will, and push downstream. If it can pull wither table data or binary logs from upstream, it will, and write them onto the database.

In the current design, the source engine and the target engine throttle one another, at the rate of their respective database capabilities. As example, assume the source is a replica tablet, and the target is a primary tablet. The target tablet requests to pull data from the source. The source reads from the replica, pushes downstream; the target intercepts and writes to the backend primary MySQL. If the backend primary MySQL is too busy, that will push back the target tablet, which will in turn stop processing events from upstream, which will in turn throttle reading from the MySQL replica. Conversely, if the source replica is slow to respond, that dictates the pace of writes to the target primary.

However, on both source and target sides, operations are aggressive on the MySQL servers. The engines will read as fast as the replica allows, or write as fast as the target primary allows. This, unfortunately, does not take into consideration the overall health of the source and target shards. We want to avoid overwhelming source/target shards so as to keep them healthy.

Specifically, we want to throttle writes on target when those writes generate replication lags on the target shard. We want to throttle reads from the source replica when that replica itself is lagging. This will not only keep the shards in a health ystate, it also makes the cut-over safer & quicker, as at any point in time the lag between source and target is known to be small.

To that effect we first need to be able to throttle on a lagging replica (source side). Our [existing table throttling](https://vitess.io/docs/reference/features/tablet-throttler/) only throttles on writes to a cluster's primary.
We will introduce lag-based "check-self" throttle check on all tablets. 

Some work has begun, illustrated in next comments.

cc @rohit-nayak-ps

shlomi-noach:
#7319 introduces a `/throttler/check-self` test, where each tablet (primary, replica, any) runs its own throttler mechanism to self check its own lag. This is true also for primaries as illustrated in the PR comment.

shlomi-noach:
#7324 extends #7319 and adds source-side throttling based on `/throttler/check-self`.

shlomi-noach:
Code-wise this is complete now that #7324 and #7364 are merged.

Documentaiton pending.

- https://github.com/vitessio/website/pull/686 is related since #7324 relies on `check-self` throttling.

shlomi-noach:
[...]

PR #9923, opened 2022-03-18
Support throttling vstreamer copy table work on source tablets

mattlord:
## Description
This provides mechanisms to limit the impact of large vreplication copy table (row streamer) workflow phases on production source tablets.

It supports throttling based on the [InnoDB history list length](https://orangematter.solarwinds.com/2015/07/20/what-is-innodb-history-list-length/) and the MySQL replication lag seen (if the source is a replica). This gives vreplication source tablets an opportunity to catch up in between copy table phase cycles (copy, catchup, fastforward). You can control this using two new vttablet flags:
- [vreplication_copy_phase_max_innodb_history_list_length](https://github.com/vitessio/vitess/pull/9923/files#diff-ee5ffc675b719e69de7147068ab4ecf03f6d5db91a000d212a30f8ce61544691R168)
- [vreplication_copy_phase_max_mysql_replication_lag](https://github.com/vitessio/vitess/pull/9923/files#diff-ee5ffc675b719e69de7147068ab4ecf03f6d5db91a000d212a30f8ce61544691R169)

## Related Issue(s)
 - Related to: https://github.com/vitessio/vitess/issues/7362

## Checklist
- [x] Tests were added
- [x] Documentation was added in https://github.com/vitessio/website/pull/1002


go/vt/vttablet/tabletserver/vstreamer/uvstreamer.go
```diff
@@ -432,6 +432,62 @@ func (uvs *uvstreamer) sendTestEvent(msg string) {
 	}
 }
 
+// waitForSource ensures that the source is able to stay within defined bounds for
+// its MVCC history list (trx rollback segment linked list for old versions of rows
+// that should be purged ASAP) and its replica lag (which will be -1 for non-replicas)
+// to help ensure that the vstream does not have an outsized harmful impact on the
+// source's ability to function normally.
```
rohit-nayak-ps, line 439: I think it is worth adding per-vstreamer metrics (`/debug/vars`), for diagnostics. Otherwise it will be difficult to understand delays/stalling in workflows. These metrics come to mind:

*  parameter settings for this tablet 
* current value of both history length and repl lag
* current wait time, if waiting in a rowstreamer 

wdyt?

go/vt/vttablet/tabletserver/tabletenv/config.go
```diff
@@ -164,7 +164,9 @@ func init() {
 
 	flag.BoolVar(&enableReplicationReporter, "enable_replication_reporter", false, "Use polling to track replication lag.")
 	flag.BoolVar(&currentConfig.EnableOnlineDDL, "queryserver_enable_online_ddl", true, "Enable online DDL.")
-	flag.BoolVar(&currentConfig.SanitizeLogMessages, "sanitize_log_messages", false, "Remove potentially sensitive information in tablet INFO, WARNING, and ERROR log messages such as query parameters.")
+
+	flag.Int64Var(&currentConfig.RowStreamer.MaxTrxHistLen, "vreplication_copy_phase_max_trx_history", 1000000, "The maximum transaction history that can exist on a vstreamer (source) before starting another round of copying rows. This helps to limit the impact on the source tablet.")
```
rohit-nayak-ps, line 168: Should we call it `vreplication_copy_phase_max_history_length` or `vreplication_copy_phase_max_history_list_length` to mirror the use in mysql docs?

go/vt/vttablet/tabletserver/tabletenv/config.go
```diff
@@ -164,7 +164,9 @@ func init() {
...
+	flag.Int64Var(&currentConfig.RowStreamer.MaxReplLagSecs, "vreplication_copy_phase_max_repl_lag", 43200, "The maximum replication lag (in seconds) that can exist on a vstreamer (source) before starting another round of copying rows. This helps to limit the impact on the source tablet.")
```
rohit-nayak-ps, line 169: Should note here that this is for the mysql replication lag on replicas, and not related to the lag reported by vreplication.

go/vt/vttablet/tabletserver/vstreamer/uvstreamer.go
mattlord: I like the idea of adding some metrics, for sure. I haven't done that yet so I'll look into it. Thanks!

go/vt/vttablet/tabletserver/tabletenv/config.go
mattlord: They're getting pretty long, but I think I like:
 - vreplication_copy_phase_max_innodb_history_list_length
 - vreplication_copy_phase_max_mysql_replication_lag

What do you think?

go/vt/vttablet/tabletserver/tabletenv/config.go
mattlord: Good point. That's also why I like adding mysql in the flag name. I'm going to go with those proposed names for now and specify MySQL in the description too.

go/vt/vttablet/tabletserver/vstreamer/uvstreamer.go
rohit-nayak-ps: You can add new metrics similar to those here:

https://github.com/vitessio/vitess/blob/441d4bd051245ecdae61746240b6fdab60fc9c19/go/vt/vttablet/tabletserver/vstreamer/engine.go#L81-L92

go/vt/vttablet/tabletserver/tabletenv/config.go
rohit-nayak-ps: nice, long is fine: they are self-documenting!

go/vt/vttablet/tabletserver/vstreamer/uvstreamer.go
mattlord: Resolving this after our discussion — to de-noise the PR a bit. Thanks again for the great suggestion and help!

rohit-nayak-ps (approved):
lgtm

rohit-nayak-ps:
Starting this discussion here to follow up on @deepthi's DM observation that these flag names are too long. @mattlord and I did note that, but we thought more descriptive the better since they would anyway reside in scripts or config files. However as @deepthi points out this may contribute to hitting OS command line length limits and also shorter the better, for recall purposes.

@mattlord, thoughts?

To start off suggestions!:
`vreplication_copy_phase_max_innodb_history_list_length` => `vrep_max_hist_len`
`vreplication_copy_phase_max_mysql_replication_lag` => `vrep_max_mysql_lag`

deepthi:
I'm not opposed to keeping the full name of the innodb variable in the flag for clarity. At a minimum, we can drop `phase` from the flag names, and `mysql` from the second one. Beyond that, I leave it up to you two.
I went and looked at other flags and we don't seem to be following any sort of standard for naming flags which are durations. Sometimes we suffix them with seconds if they are `int` but expected to be seconds, sometimes we don't. So even though other `replication_lag` flags are durations, I don't feel like we can impose that as a standard at this point.

mattlord:
IMO related flags should all have the same prefix. I would have used `vrepl` but we already had a number of flags with the `vreplication` prefix and beyond that `vreplication_copy_phase` so I used that:
```
$ vttablet --help 2>&1 | grep -E "vrepl|vstream" 
  --vreplication_copy_phase_duration duration
  --vreplication_copy_phase_max_innodb_history_list_length int
	The maximum InnoDB transaction history that can exist on a vstreamer (source) before starting another round of copying rows. This helps to limit the impact on the source tablet. (default 1000000)
  --vreplication_copy_phase_max_mysql_replication_lag int
	The maximum MySQL replication lag (in seconds) that can exist on a vstreamer (source) before starting another round of copying rows. This helps to limit the impact on the source tablet. (default 43200)
  --vreplication_experimental_flags int
	(Bitmask) of experimental features in vreplication to enable (default 1)
  --vreplication_healthcheck_retry_delay duration
  --vreplication_healthcheck_timeout duration
  --vreplication_healthcheck_topology_refresh duration
  --vreplication_heartbeat_update_interval int
	Frequency (in seconds, default 1, max 60) at which the time_updated column of a vreplication stream when idling (default 1)
  --vreplication_replica_lag_tolerance duration
  --vreplication_retry_delay duration
  --vreplication_store_compressed_gtid
	Store compressed gtids in the pos column of _vt.vreplication
  --vreplication_tablet_type string
  --vstream_dynamic_packet_size
  --vstream_packet_size int
```

The `mysql` is important in the second var name IMO as we also have vreplication lag in this context and that's not what this flag is about.

I don't really agree that shorter is easier to remember, but I don't feel too strongly here. 

I'm not sure that OS limits are a major concern, but it's possible:
```
$ sw_vers && echo -n "Max cmd-line value in bytes: " && getconf ARG_MAX
ProductName:	macOS
ProductVersion:	12.3.1
BuildVersion:	21E258
Max cmd-line value in bytes: 1048576


$ cat /etc/system-release && echo -n "Max cmd-line value in bytes: " && getconf ARG_MAX
Amazon Linux release 2 (Karoo)
Max cmd-line value in bytes: 2621440
```

IMO what we really need is a project/ticket to define some Vitess policies for command-line options/flags and then apply them to all flags across all of vitess. Want me to create an issue for that? I don't mind being the one to work on it either — the lack of a policy is the real issue IMO (don't feel too strongly on the specifics, but the lack of uniformity and predictability is bad for users).

deepthi:
Let's create an issue/task saying "we need to define a policy", and once there is a concrete proposal it can become an RFC. The current state is not good. We also have some flags with dashes and some with underscores.

PR #9926, opened 2022-03-21
deflake TestCanSwitch by switching from exact string match to loose regex

ajm188:
## Description

I had this test fail in [CI](https://github.com/vitessio/vitess/runs/5598763583?check_suite_focus=true) with the following:

```
--- FAIL: TestCanSwitch (0.15s)
[715](https://github.com/vitessio/vitess/runs/5598763583?check_suite_focus=true#step:7:715)
    --- FAIL: TestCanSwitch/High_Lag (0.03s)
[716](https://github.com/vitessio/vitess/runs/5598763583?check_suite_focus=true#step:7:716)
        workflow_test.go:158: 
[717](https://github.com/vitessio/vitess/runs/5598763583?check_suite_focus=true#step:7:717)
            	Error Trace:	workflow_test.go:158
[718](https://github.com/vitessio/vitess/runs/5598763583?check_suite_focus=true#step:7:718)
            	Error:      	"replication lag 6s is higher than allowed lag 5s" does not contain "replication lag 7s is higher than allowed lag 5s"
```

We switch to a regexp, to check that the lag is *greater* than the current lag, not that it's exactly the lag we specify, which is what we actually care about asserting. 

## Checklist

### DIFF
--- a/go/vt/wrangler/workflow_test.go
+++ b/go/vt/wrangler/workflow_test.go
@@ -19,6 +19,9 @@ package wrangler
 import (
 	"context"
 	"fmt"
+	"regexp"
+	"strconv"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


mattlord (approved):
Thanks!

PR #9929, opened 2022-03-21
Improve escape handling for vindex materializer

dbussink:
In case we handle table names here with for example a dash in them, we should escape it consistently like how we escape values in other parts of the vschema as well.

Additionally, use the proper escape functionality instead of manually putting the string between backticks to ensure consistent behavior.

## Checklist

## Deployment Notes

This would break an existing specific case in a vschema, namely using a table reference to a table that requires escaping. However, I think there's other areas already where this breaks from what I can gather so not sure if that's a practical concern then or not.

### DIFF
--- a/go/vt/wrangler/materializer.go
+++ b/go/vt/wrangler/materializer.go
@@ -30,6 +30,7 @@ import (
 	"google.golang.org/protobuf/proto"
 
 	"vitess.io/vitess/go/json2"
+	"vitess.io/vitess/go/sqlescape"
 	"vitess.io/vitess/go/sqltypes"
 	"vitess.io/vitess/go/vt/binlog/binlogplayer"
 	"vitess.io/vitess/go/vt/concurrency"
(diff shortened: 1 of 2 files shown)


go/vt/wrangler/materializer.go
```diff
@@ -454,11 +455,11 @@ func (wr *Wrangler) prepareCreateLookup(ctx context.Context, keyspace string, sp
 	if !strings.Contains(vindex.Type, "lookup") {
 		return nil, nil, nil, fmt.Errorf("vindex %s is not a lookup type", vindex.Type)
 	}
-	strs := strings.Split(vindex.Params["table"], ".")
-	if len(strs) != 2 {
-		return nil, nil, nil, fmt.Errorf("vindex 'table' must be <keyspace>.<table>: %v", vindex)
+
+	targetKeyspace, targetTableName, err = sqlparser.ParseTable(vindex.Params["table"])
+	if err != nil || targetKeyspace == "" {
+		return nil, nil, nil, fmt.Errorf("vindex 'table' must be <keyspace>.<table>: %v", vindex.Params["table"])
```
mattlord, line 461: Nit, and not something you added, but since we're here I'd say:
```
return nil, nil, nil, fmt.Errorf("vindex table name must be in the form <keyspace>.<table>. Got: %v", vindex.Params["table"])
```

go/vt/wrangler/materializer.go
```diff
@@ -727,12 +729,11 @@ func (wr *Wrangler) ExternalizeVindex(ctx context.Context, qualifiedVindexName s
...
 		return fmt.Errorf("vindex %s not found in vschema", qualifiedVindexName)
 	}
-	qualifiedTableName := sourceVindex.Params["table"]
-	splits = strings.Split(qualifiedTableName, ".")
-	if len(splits) != 2 {
-		return fmt.Errorf("table name in vindex should be of the form keyspace.table: %s", qualifiedTableName)
+
+	targetKeyspace, targetTableName, err := sqlparser.ParseTable(sourceVindex.Params["table"])
+	if err != nil || targetKeyspace == "" {
+		return fmt.Errorf("table name in vindex should be of the form keyspace.table: %s", sourceVindex.Params["table"])
```
mattlord, line 735: Again, nit and not something you added but we should make this consistent with the error message from above:
```
fmt.Errorf("vindex table name must be in the form <keyspace>.<table>. Got: %v", sourceVindex.Params["table"])
```

We could alternatively add a format string and use that instead to ensure uniformity...
```
tableNameErrStrf = `vindex table name must be in the form <keyspace>.<table>. Got: %v`
...
fmt.Errorf(tableNameErrStrf, sourceVindex.Params["table"])
```

I don't feel strongly either way, unless this is used in even more places within the file. :-)

mattlord (commented):
LGTM! Had a small nit about the error messages but those already existed before your PR too. :-) Let me know when you're ready for final review.

Thanks!

go/vt/wrangler/materializer.go
dbussink: Also fixed :smile:.

Issue #9925, opened 2022-03-20
Flaky TestV2WorkflowsAcrossDBVersions


rohit-nayak-ps:
### Overview of the Issue

`TestV2WorkflowsAcrossDBVersions` is very flaky. Since it is Not Required, we don't stall PRs for this, but I wonder if there is a real bug causing this or just a CI quirk.

There are two types of failures that I have seen. None of these are consistent, but they are always on tablet zone1-2000. It may just be that we have created a lot of tablets by that point and CI resource limitations are kicking in, but it is strange that it is on the same tablet.

1. A deadlock that happens in a vtctld command: either VDiff or a TrafficSwitch. See https://pipelines.actions.githubusercontent.com/TBP6MbgxaNRkq1k6DamNZYdWLId6AliZAF5kl9JXndalqVBili/_apis/pipelines/1/runs/782062/signedlogcontent/2?urlExpires=2022-03-20T20%3A36%3A53.5448773Z&urlSigningMethod=HMACV1&urlSignature=76BVUIR0uibNl8tYRYh8KU5JJVJuEZDcUc3pLrwIFAg%3D
`2022-03-20T19:24:43.8292730Z I0320 19:23:01.904571   19472 vttablet_process.go:484] WaitForVReplicationToCatchup succeeded at 2022-03-20 19:23:01.903977406 +0000 UTC m=+152.762029564
2022-03-20T19:24:43.8293318Z === RUN   TestV2WorkflowsAcrossDBVersions/reshardMerchant3to1Merge/reshard/vdiff
2022-03-20T19:24:43.8294291Z I0320 19:23:01.904714   19472 vtctlclient_process.go:202] Executing vtctlclient with command: vtctlclient --server 127.0.0.1:15999 VDiff -- --tablet_types=primary --source_cell= --format json merchant-type.m3m1
[...]

PR #9935, opened 2022-03-21
Address flakiness of vreplication tests

mattlord:
## Description
The `vreplication*` tests were flaky, but out of those the [`vreplication_across_db_versions`](https://github.com/vitessio/vitess/blob/main/.github/workflows/cluster_endtoend_vreplication_across_db_versions.yml) test was the most flaky. If the machine they were running on seemed to be under stress and have resource contention, then that seemed to in turn cause at least one of the `vreplication*` tests to fail (typically failing to start one of the mysqld instances later on in the test). This work:
1. Increases the OS AIO request limit for the machine where it runs (this can be necessary with a lot of mysqld's doing a lot of I/O)
2. Explicitly ensures we have 0 replica and rdonly tablets for the tests
3. Modifies the MySQL cnf file used so that we use minimal resources and reduce operational overhead
4. Increases the available local ephemeral ports in the OS (you can exhaust these when opening and closing a lot of connections due to the `TIMED_WAIT` state before reuse)
5. Increases the max file descriptors per process in our OS shell (when running a lot of processes that have a lot of open FDs you can hit this)

## Related Issue(s)
 - Fixes ❓: https://github.com/vitessio/vitess/issues/9925 ❓
 - Fixes ❓: https://github.com/vitessio/vitess/issues/9948 ❓

## Checklist
- [x] Tests are not required

### DIFF
--- a/.github/workflows/cluster_endtoend_schemadiff_vrepl.yml
+++ b/.github/workflows/cluster_endtoend_schemadiff_vrepl.yml
@@ -28,6 +28,9 @@ jobs:
     - name: Tune the OS
       run: |
         echo '1024 65535' | sudo tee -a /proc/sys/net/ipv4/ip_local_port_range
+        # Increase the asynchronous non-blocking I/O. More information at https://dev.mysql.com/doc/refman/5.7/en/innodb-parameters.html#sysvar_innodb_use_native_aio
+        echo "fs.aio-max-nr = 1048576" | sudo tee -a /etc/sysctl.conf
+        sudo sysctl -p /etc/sysctl.conf
 
     - name: Check out code
       uses: actions/checkout@v2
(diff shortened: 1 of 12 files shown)


mattlord:
It's anecdotal at this point but all of the `vreplication*` tests passed on the first try 3 times in a row (and finished relatively quickly)! That's a good sign (albeit premature to celebrate) given that 2-5 other tests failed each time — which in my experience seemed to be when at least one of the `vreplication*` ones would also fail.

I'd like to move forward with this and measure how the `vreplication*` workflows go over the next week or two and then revisit again if they're still showing signs of flakiness.

GuptaManan100 (approved):

rohit-nayak-ps (approved):
lgtm

PR #9937, opened 2022-03-21
region_example: sleep 5s before switching reads to allow copy to finish

deepthi:
## Description

After I changed the example to use v2 vreplication commands instead of v1, it started failing on several PRs (though it passed on the original PR #9875)
Temporarily adding a sleep to allow it to finish.
We should investigate whether v2 is slower than v1, keeping in mind that this is a tiny dataset.

## Related Issue(s)

#9875 

## Checklist


ajm188 (approved):
<3

PR #9939, opened 2022-03-22
ORDER BY scoping rules should not be used in group_concat

systay:
## Description
We do some special handling of scopes in ORDER BY, to allow the use of both aliases introduced in the SELECT expressions, but also columns coming from a table not listed in the SELECT list. These scoping rules should not be used on the ORDER BY that can be used inside the `group_concat` aggregation function.

## Checklist
- [x] Should this PR be backported? ``Maybe. It's simple and safe enough``


frouioui (approved):
We should also add a plan with `group_concat` in a join / sharded plan.

systay:
> We should also add a plan with `group_concat` in a join / sharded plan.

I don't think we support that today

PR #9941, opened 2022-03-22
cache: update Cache implementation to use Generics

vmg:
## Description

It's a brave new world out there! After a lot of research, this is the main refactoring I believe we can make in the codebase to introduce generics without negatively affecting performance.

Porting the `cache` package to use Generics makes the code easier to read and reduces the amount of memory we're using for caching without any measured slowdowns.

The main thing bothering me is that the `structcheck` linter doesn't work with generic code: it marks all fields in a generic struct as unused, so I've had to disable it for the `go/cache/` path.

## Checklist

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -45,6 +45,10 @@ issues:
       - errcheck
       - goimports
 
+    - path: '^go/cache/'
+      linters:
+        - structcheck
+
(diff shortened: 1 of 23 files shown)


deepthi:
https://github.com/vitessio/vitess/pull/9992#issuecomment-1080707290

Issue #9936, opened 2022-03-21
Bug Report: vttablet hanging when running PlannedReparentShard


mattlord:
### Overview of the Issue

During a PRS we sometimes see the tablet stuck waiting for something at this point in the logs:
```
I0317 04:56:50.637023    9434 rpc_replication.go:388] DemotePrimary
I0317 04:56:50.638689    9434 rpc_replication.go:438] DemotePrimary disabling query service
I0317 04:56:50.638703    9434 state_manager.go:214] Starting transition to PRIMARY Not Serving, timestamp: 2022-03-17 04:25:38.205679188 +0000 UTC
I0317 04:56:50.638773    9434 tablegc.go:212] TableGC: closing
...
```

Here's a sample debug blocking profile covering this: https://gist.github.com/derekperkins/dd6d54809a98b582c03909061e639766

### Reproduction Steps

We suspect that it involves:
1. Setting up many active vstreams: MoveTables,Reshard,OnlineDDL,Messaging
2. Doing PRS while those are active

### Binary Version

```sh
v13.0.0
```

### Operating System and Environment details

```sh
N/A
```

rohit-nayak-ps:
@derekperkins, @GuptaManan100  is going to create a branch tomorrow with added logs, around tablet shutdown, to see if we can narrow down where the tablet hangs.

mattlord:
@derekperkins after reading [your comments, attachments](https://github.com/vitessio/vitess/issues/8909), and thinking about it a bit more... I have a suspicion similar to yours in that I suspect PRS blocking is a symptom of another problem. I (also) suspect that for some reason the message table related queries are getting backlogged in the tablet's MySQL instance (perhaps the table size, structure, or queries used against it have changed over the recent months and vitess versions) — which would explain why you see no new messages generated even before the PRS (maybe that's WHY you do the PRS?) — which in turn causes the PRS to block during primary demotion waiting for those open statements/transactions to finish on the current/old primary during the demotion step. PRS will, by default, wait indefinitely for any running statements/transactions to finish on the current primary during demotion.

Note that with vreplication (branching, online DDL, etc) you can have an open transaction on the source tablet/mysqld that lasts minutes or even hours to maintain the consistent snapshot for a copy rows phase/cycle. 

This behavior was described and investigated in the past here:
 - https://github.com/vitessio/vitess/pull/6645
 - https://github.com/vitessio/vitess/issues/6982
[...]

PR #9942, opened 2022-03-22
Adds comments to UnserveCommon to find potential Deadlock

GuptaManan100:
## Description

This PR adds a lot of comments in all the Close functions in `unserveCommon` called before killing all OLAP queries. The comments are added recursively inside the Close functions as well to be able to debug a potential Deadlock in `DemotePrimary`.

## Related Issue(s)

#9936 

## Checklist


rohit-nayak-ps (approved):
lgtm
fingers crossed!

GuptaManan100:
Yep that would certainly be better! Do we want to add these many comments? WDYT @mattlord @rohit-nayak-ps

rohit-nayak-ps:
> Yep that would certainly be better! Do we want to add these many comments? WDYT @mattlord @rohit-nayak-ps

* Not sure if all users facing this issue can easily set verbosity levels on and off
* Since the issue is not consistently reproducible, having logs all the time is at least needed until we find the root cause
* Since logging is only during shutdown it should have negligible affect on log sizes

In any case we can  revisit once we have this in production for a month or two, or until we find the root cause. Maybe we can consolidate the consecutive logs in `unserveCommon()`.

GuptaManan100:
Alright! I'll make the changes

GuptaManan100:
@mattlord @rohit-nayak-ps I have made the changes for consecutive logs in `unserveCommon`. However, I haven't made the change for the verbosity level that @mattlord suggested after talking to @deepthi. We are currently not using the verbosity levels anywhere else, and when we do, it should be a much larger change.

I agree with both of you that we should merge this on main and revisit once we have found the bug and fixed it.

mattlord (approved):
Thanks! The TableGC CI test is also one that's flaky IIRC... we could potentially modify that test to print out the tail end of the tablet INFO logs on failure. I wonder if we're hitting the same issue w/o knowing it.

PR #9957, opened 2022-03-23
Flakiness fix for Upgrade Downgrade Backup Manual test

GuptaManan100:
## Description

The upgrade downgrade backups manual test has been flaky on the ci for a while. On investigation of the same with the help of https://github.com/vitessio/vitess/pull/9944 led to the following discovery - 

While restarting the tablets with a backup, the tablet transitions to `RESTORE` type to restore from the local backup. Without waiting for the backup to finish, we call `InitShardPrimary` which sees the tablet type as `RESTORE` and turns off semi-sync on that tablet.

Later restoration completes and the tablet transitions to `REPLICA`, but its semi-sync settings aren't fixed, which leads to the primary stalling indefinitely on writes.

This PR fixes this issue by first waiting for the restoration phase to finish before calling InitShardPrimary. Earlier we had a 5 second hard wait to ensure that this phase had finished but on the CI it generally takes longer. Locally 5 seconds is sufficient, which made reproducibility a challenge. 

I think it is better to have an explicit check on the tablet type by `curl`ing the tablet server API endpoint instead of any fixed hard time sleep.

## Checklist
- [x] Tests were added or are not required Yes, the upgrade downgrade tests for backup

### DIFF
--- a/examples/local/backups/restart_tablets.sh
+++ b/examples/local/backups/restart_tablets.sh
@@ -23,19 +23,31 @@ for i in 100 101 102; do
   CELL=zone1 TABLET_UID=$i ./scripts/mysqlctl-up.sh
   CELL=zone1 KEYSPACE=commerce TABLET_UID=$i ./scripts/vttablet-up.sh
 done
-sleep 5
-vtctldclient InitShardPrimary --force commerce/0 zone1-100
 
 for i in 200 201 202; do
   CELL=zone1 TABLET_UID=$i ./scripts/mysqlctl-up.sh
   SHARD=-80 CELL=zone1 KEYSPACE=customer TABLET_UID=$i ./scripts/vttablet-up.sh
 done
-sleep 5
-vtctldclient InitShardPrimary --force customer/-80 zone1-200
 
 for i in 300 301 302; do
   CELL=zone1 TABLET_UID=$i ./scripts/mysqlctl-up.sh
   SHARD=80- CELL=zone1 KEYSPACE=customer TABLET_UID=$i ./scripts/vttablet-up.sh
 done
 sleep 5
+
+# Wait for all the replica tablets to be in the serving state before initiating
+# InitShardPrimary. This is essential, since we want the RESTORE phase to be
+# complete before we start InitShardPrimary, otherwise we end up reading the
+# tablet type to RESTORE and do not set semi-sync, which leads to the primary
+# hanging on writes.
+for i in 101 201 301; do
+  for _ in $(seq 0 300); do
+   status=$(curl "http://$hostname:15$i/debug/status_details")
+   echo "$status" | grep "REPLICA: Serving" && break
+   sleep 0.1
+  done
+done
+
+vtctldclient InitShardPrimary --force commerce/0 zone1-100
+vtctldclient InitShardPrimary --force customer/-80 zone1-200
 vtctldclient InitShardPrimary --force customer/80- zone1-300
--- a/examples/local/backups/restart_tablets.sh
+++ b/examples/local/backups/restart_tablets.sh
@@ -40,12 +40,22 @@ sleep 5
 # complete before we start InitShardPrimary, otherwise we end up reading the
 # tablet type to RESTORE and do not set semi-sync, which leads to the primary
 # hanging on writes.
+totalTime=300
 for i in 101 201 301; do
-  for _ in $(seq 0 300); do
-   status=$(curl "http://$hostname:15$i/debug/status_details")
-   echo "$status" | grep "REPLICA: Serving" && break
-   sleep 0.1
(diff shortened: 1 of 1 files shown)


examples/local/backups/restart_tablets.sh
```diff
@@ -23,19 +23,31 @@ for i in 100 101 102; do
...
+  for _ in $(seq 0 300); do
+   status=$(curl "http://$hostname:15$i/debug/status_details")
+   echo "$status" | grep "REPLICA: Serving" && break
+   sleep 0.1
+  done
+done
+
+vtctldclient InitShardPrimary --force commerce/0 zone1-100
+vtctldclient InitShardPrimary --force customer/-80 zone1-200
 vtctldclient InitShardPrimary --force customer/80- zone1-300
```
ajm188, line 53: nit: can you add the trailing newline here?

examples/local/backups/restart_tablets.sh
```diff
@@ -23,19 +23,31 @@ for i in 100 101 102; do
...
   SHARD=80- CELL=zone1 KEYSPACE=customer TABLET_UID=$i ./scripts/vttablet-up.sh
 done
 sleep 5
+
+# Wait for all the replica tablets to be in the serving state before initiating
+# InitShardPrimary. This is essential, since we want the RESTORE phase to be
+# complete before we start InitShardPrimary, otherwise we end up reading the
+# tablet type to RESTORE and do not set semi-sync, which leads to the primary
+# hanging on writes.
+for i in 101 201 301; do
```
ajm188, line 43: the way this is structured, your wait time scales linearly with the number of shards. we could speed this up by waiting ~~ sort of ~~ concurrently:

```bash
declare -a healthy

for _ in $(seq 0 300); do
  for i in 1 2 3; do
    if [[ "${healthy[$i] -eq 1 ]]; then
      continue;
    fi

    status=$(curl "http://$hostname:15${i}01/debug/status_details" | grep "REPLICA: serving")
    if [[ -n "$status" ]]; then
        healthy[$i]=1;
    fi
  done;
  # iterate and check if all 3 are healthy, then break
done;

# fail if not all 3 healthy
```

GuptaManan100: Why would the wait time scale linearly with my original code? The tablets are already started and we are just waiting for them to reach a specific state. The loop for all the tablets after the slowest tablet will just run immediately, right? So we will only be waiting for the time of the slowest tablet.

deepthi (commented):
lgtm

examples/local/backups/restart_tablets.sh
mattlord: If we want to do this in parallel, we could use gnu parallel. For example, something like this:
```
for _ in $(seq 0 300); do
  cmd=""
  for i in 1 2 3; do
    if [[ "${healthy[$i]}" -eq 1 ]]; then
      continue;
    fi

    cmd+='if curl --max-time 1 "http://$hostname:15${i}01/debug/status_details" | grep -v "REPLICA: serving"; then healthy[${i}]=1; fi; '
  done;
  parallel -nocite ::: eval "${cmd}"
  # iterate and check if all 3 are healthy, then break
done;
```

mattlord: If we want to do this in parallel, we could use gnu parallel. For example, something like this:
```
for _ in $(seq 0 300); do
  cmd=""
  for i in 1 2 3; do
    if [[ "${healthy[$i]}" -eq 1 ]]; then
      continue;
    fi

    cmd+="if curl --max-time 1 \"http://$hostname:15${i}01/debug/status_details\" | grep -q \"REPLICA: serving\"; then healthy[${i}]=1; fi; "
  done;
  echo -n "CMD: ${cmd}"
  parallel -nocite ::: eval "${cmd}"
  # iterate and check if all 3 are healthy, then break
done;
```

mattlord: Note that it's been a while since I used parallel and it's a little esoteric so please don't consider that viable code ^^ :D but just a pointer if we care enough to parallelize it.

GuptaManan100: Thankyou @mattlord for this information. TIL gnu parallel exists!

examples/local/backups/restart_tablets.sh
GuptaManan100: fixed

examples/local/backups/restart_tablets.sh
GuptaManan100: @ajm188 @mattlord I have done something different achieving the same goal, which I find more intuitive to understand. Could you take a look at this again?

ajm188 (approved):
works for me!! very nice

PR #9958, opened 2022-03-23
Online DDL: identify VReplication retrying failure, terminate migration

shlomi-noach:
## Description
There's a set of errors that VReplication encounters, that it doesn't fail on, it just keeps retrying forever. Recently, https://github.com/vitessio/vitess/pull/9538 introduced `time_heartbeat`. By looking at both `time_heartbeat` and `time_updated`
we are able to determine whether a vreplication stream is making no progress. We look at the greater value of the two. If it is not making progress for X minutes, then vreplication is effectively not doing its work.

Online DDL executor now looks at that (greater of the two) value, persists it in `schema_migrations`, and only updates `liveness_timestamp` of the (`vitess` strategy) migration if it sees an increase in the value.

What this means is in effect, if VReplication is stuck/retrying, then `liveness_timestamp` does not increase. In turn, the executor will see that the migration has not indicated liveness in quite a while, and will eventually (configured at 10min) terminate the migration and mark it as `failed`.

So this gives us a `10min` reaction time to a hanging VReplication stream.

## Related Issue(s)

Tracking: #6926 

## Checklist


shlomi-noach:
The only failing test is `Upgrade Downgrade Testing Query Serving`, which we are aware of and is unrelated to this PR.

shlomi-noach:
On top of the fix in this PR, we also have https://github.com/vitessio/vitess/pull/9973 which identifies specific error codes and completely fails the vreplication workflow on those errors. This PR then serves as a sort of "plan B" for when vreplication still insists on retrying and never makes progress.

shlomi-noach:
implicitly merged by https://github.com/vitessio/vitess/pull/9973

PR #9963, opened 2022-03-23
VReplication: use strict sql_mode for Online DDL

shlomi-noach:
## Description

OnlineDDL requires `STRICT_ALL_TABLES` to ensure it does not accidentally modify data while copying to the shadow table. This can happen in cases when column scope is reduced (e.g. `BIGINT` to `INT`, or `VARCHAR(20)` to `VARCHAR(10)`). VReplication's default behavior is to just use `NO_AUTO_VALUE_ON_ZERO`.

Stealing some minimal code from #9742, `vreplication` becomes aware of the _type_ of workflow its using: materialization, move tables, online ddl, ...

In this PR, when vreplication identifies that it is being run by Online DDL, it adds `STRICT_ALL_TABLES` to its `sql_mode`.

## Related Issue(s)

- #9742
- #6926 

## Checklist


go/vt/binlog/binlogplayer/binlog_player.go
```diff
@@ -586,12 +585,14 @@ type VRSettings struct {
 	MaxTPS            int64
 	MaxReplicationLag int64
 	State             string
+	WorkflowType      int64
```
mattlord, line 588: Nice! I was going to need to do something similar if we ever decide to move forward with https://github.com/vitessio/vitess/pull/9845 as we only want to force the PK for MoveTables.

go/vt/binlog/binlogplayer/binlog_player.go
```diff
@@ -600,31 +601,36 @@ func ReadVRSettings(dbClient DBClient, uid uint32) (VRSettings, error) {
 	if len(qr.Rows) != 1 {
 		return VRSettings{}, fmt.Errorf("checkpoint information not available in db for %v", uid)
 	}
-	vrRow := qr.Rows[0]
+	vrRow := qr.Named().Row()
```
mattlord, line 604: Thank you for refactoring this usage as we go!

go/vt/binlog/binlogplayer/binlog_player_test.go
```diff
@@ -133,25 +145,37 @@ func TestApplyEventsFail(t *testing.T) {
...
 }
 
+var settingsFields []*querypb.Field = []*querypb.Field{
+	{Name: "pos", Type: sqltypes.VarBinary},
+	{Name: "stop_pos", Type: sqltypes.VarBinary},
+	{Name: "max_tps", Type: sqltypes.Int64},
+	{Name: "max_replication_lag", Type: sqltypes.Int64},
+	{Name: "state", Type: sqltypes.VarBinary},
+	{Name: "workflow_type", Type: sqltypes.Int64},
+	{Name: "workflow", Type: sqltypes.VarChar},
```
mattlord, line 155: This is workflow name, I'm assuming? If so, might be worth being explicit about that (even though it's a reasonable assumption).

go/vt/vttablet/onlineddl/executor.go
```diff
@@ -1042,6 +1042,15 @@ func (e *Executor) ExecuteWithVReplication(ctx context.Context, onlineDDL *schem
...
 		}
+
+		{
+			// temporary hack. todo: this should be done when inserting any _vt.vreplication record across all workflow types
+			query := fmt.Sprintf("update _vt.vreplication set workflow_type = %d where workflow = '%s'",
+				binlogdatapb.VReplicationWorkflowType_ONLINEDDL, v.workflow)
+			if _, err := e.vreplicationExec(ctx, tmClient, tablet.Tablet, query); err != nil {
+				return vterrors.Wrapf(err, "VReplicationExec(%v, %s)", tablet.Tablet, query)
+			}
+		}
```
mattlord, line 1053: This is to "upgrade" any existing workflow records, right?

go/vt/vttablet/tabletmanager/vreplication/vreplicator.go
```diff
@@ -82,6 +82,7 @@ const (
 	// in effect on the source when it was created:
 	//   https://github.com/mysql/mysql-server/blob/3290a66c89eb1625a7058e0ef732432b6952b435/client/mysqldump.cc#L795-L818
 	SQLMode          = "NO_AUTO_VALUE_ON_ZERO"
```
mattlord, line 84: Now that we have `StrictSQLMode`, probably worth renaming this to ~ `PermissiveSQLMode`

proto/binlogdata.proto
```diff
@@ -200,6 +203,16 @@ enum OnDDLAction {
...
 }
 
+// VReplicationWorkflowType define types of vreplication workflows.
+enum VReplicationWorkflowType {
+  MATERIALIZE = 0;
+  MOVETABLES = 1;
+  CREATELOOKUPINDEX = 2;
+  MIGRATE = 3;
+  RESHARD = 4;
+  ONLINEDDL = 5;
```
mattlord, line 213: Worth adding MESSAGING here? I feel like we probably should.

mattlord (commented):
I had a few nits and comments but LGTM! The workflow type stuff will make this https://github.com/vitessio/vitess/pull/9845 easier if we ever decide to move forward with it as it should only apply to MoveTables or Reshard (not materialize or messaging). I was planning on doing something similar to what you did here to take different actions depending on the workflow type and I would no longer need to do that. ❤️

rohit-nayak-ps (approved):
lgtm

mattlord (approved):

go/vt/binlog/binlogplayer/binlog_player.go
shlomi-noach: This was stolen from #9742 authored by @rohit-nayak-ps

go/vt/binlog/binlogplayer/binlog_player.go
shlomi-noach: This again was stolen from #9742 authored by @rohit-nayak-ps

go/vt/binlog/binlogplayer/binlog_player_test.go
shlomi-noach: This again was stolen from #9742 authored by @rohit-nayak-ps. Anyway, `workflow` here is the UUID, not the name. I think it's too late to change this name, it's all over the place.

go/vt/vttablet/onlineddl/executor.go
shlomi-noach: No, actually. This immediately follows an INSERT. Because I did not want to import the entirety of #9742 , the INSERT does not yet populate the new `workflow_type` column. So right after we `INSERT`, we then update the `workflow_type` column.
In the future, the above call (see `insertVReplicationQuery`) will include the columns.

So this only affects a newly submitted online ddl workflow.

go/vt/vttablet/tabletmanager/vreplication/vreplicator.go
shlomi-noach: That's a good idea! I'll leave this for another PR as I'm keen to keep this PR as small as possible.

proto/binlogdata.proto
shlomi-noach: Delegating to #9742

shlomi-noach:
> I had a few nits and comments but LGTM! The workflow type stuff will make this https://github.com/vitessio/vitess/pull/9845 easier if we ever decide to move forward with it as it should only apply to MoveTables or Reshard (not materialize or messaging). I was planning on doing something similar to what you did here to take different actions depending on the workflow type and I would no longer need to do that. heart

@mattlord - it's all @rohit-nayak-ps 's work in #9742 actually... All credit goes to him. I merely stole some of that code.

shlomi-noach:
Only failing test is `Upgrade Downgrade Testing Query Serving` which we are aware of, and is unrelated. Given all other tests pass and two approving reviews, I'm now merging this.

PR #9967, opened 2022-03-23
Address flakiness of some cluster e2e tests

mattlord:
## Description
In https://github.com/vitessio/vitess/pull/9935 we added support for marking a test as requiring additional configuration in order to limit resource usage. This was done to make our heaviest (vreplication) tests more reliable as they would often fail when there was any resource contention on the runners. In this PR we're leveraging that work to modify some of the generic Cluster e2e tests that also often failed when resource contention seemed high.

In the process we also rename those tests to be something more meaningful and append `_heavy` to the name as a generic marker that they should be configured to limit resource usage. So going forward we can append that string to any other test name if we feel that resource contention is the common cause of failures.

## Related Issue(s)
 - Related to: https://github.com/vitessio/vitess/pull/9935

## Checklist
- [x] Tests are not required

### DIFF
--- a/.github/workflows/cluster_endtoend_ers_prs_newfeatures_heavy.yml
+++ b/.github/workflows/cluster_endtoend_ers_prs_newfeatures_heavy.yml
@@ -1,9 +1,9 @@
 # DO NOT MODIFY: THIS FILE IS GENERATED USING "make generate_ci_workflows"
 
-name: Cluster (16)
+name: Cluster (ers_prs_newfeatures_heavy)
 on: [push, pull_request]
 concurrency:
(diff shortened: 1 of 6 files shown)


mattlord:
@deepthi I’ll need you or someone else with admin level access to the GitHub repo to swap the test/action names as required

I believe that’s why we still have e.g.
```
Run endtoend tests on Cluster (23) Expected — Waiting for status to be reported
```
And the one it was renamed to is not marked as Required. We will need to coordinate the merge with the repo config changes

deepthi (approved):
👍

Issue #9653, opened 2022-02-08
vtgate panic - select 1 as one from dual


jakon89:
### Overview of the Issue

vtgate panics when trying to execute below query:

```sql
select 1 as one
from dual
where exists (
  select 1 as one
  from information_schema.TABLES
  where (
    information_schema.TABLES.TABLE_NAME = 'proc'
    and information_schema.TABLES.TABLE_SCHEMA = 'mysql'
  )
)
```

The above query is sent by JOOQ library when reading DB schema (that schema is used then to generate java classes)

This happens for both shared and unsharded keyspaces.

### Reproduction Steps

```markdown
* login to mysql
* mysql> select 1 as one from dual where exists (   select 1 as one   from information_schema.TABLES   where (     information_schema.TABLES.TABLE_NAME = 'proc'     and information_schema.TABLES.TABLE_SCHEMA = 'mysql'   ) );
ERROR 2013 (HY000): Lost connection to MySQL server during query
No connection. Trying to reconnect...
Connection id:    57
Current database: cloudsync_db

ERROR 2013 (HY000): Lost connection to MySQL server during query
No connection. Trying to reconnect...
Connection id:    58
Current database: cloudsync_db

ERROR 2013 (HY000): Lost connection to MySQL server during query
```

### Binary Version

```shell
vitess@vitess:/vt/vtdataroot/tmp$ vtgate -version
ERROR: logging before flag.Parse: E0208 19:40:39.319003    1368 syslogger.go:149] can't connect to syslog
Version: 13.0.0-SNAPSHOT (Git revision 727d8fe893 branch 'main') built on Tue Feb  1 15:49:03 UTC 2022 by vitess@buildkitsandbox using go1.17.6 linux/amd64
```
```


### Operating System and Environment details

```shell
docker/linux
```

### Log Fragments

```shell
vtgate:

E0208 19:41:12.424888     401 server.go:319] mysql_server caught panic:
assignment to entry in nil map
runtime/map_faststr.go:204 (0x41306e)
vitess.io/vitess/go/vt/vtgate/planbuilder/route.go:383 (0xe5c56e)
vitess.io/vitess/go/vt/vtgate/planbuilder/expr.go:134 (0xe34a24)
vitess.io/vitess/go/vt/vtgate/planbuilder/select.go:407 (0xe612c9)
vitess.io/vitess/go/vt/vtgate/planbuilder/select.go:208 (0xe5fc44)
[...]

PR #9969, opened 2022-03-23
fix: dual query with exists clause having system table query in it

harshit-gangal:
## Description

This PR adds an empty map if the system table map is uninitialized in route.

## Related Issue(s)

- Fixes https://github.com/vitessio/vitess/issues/9653

## Checklist
- [ ] Should this PR be backported? Not sure


frouioui (approved):
LGTM, let's rebase this PR on `main` once https://github.com/vitessio/vitess/pull/9975 is merged.

Issue #9952, opened 2022-03-23
Track full git SHA1 instead of short version


malpani:
This is a minor feature request to return the full git SHA1 instead of the short version. 
https://github.com/vitessio/vitess/blob/4c33b2b5a90dc200e6cc649681cbace2520725bd/tools/build_version_flags.sh#L24 

This is to enable the deployment + monitoring setup in my team to easily detect and appropriately update any stale vttablets.

PR #9970, opened 2022-03-23
Track full git sha1 string instead of the short version

malpani:
Signed-off-by: Ankit Malpani <malpani@stripe.com>

## Description
Track full git sha1 string instead of the short version
This is needed for our deployment monitoring systems to identify and remediate any stale running vttablets

```
Before Change:
"BuildGitRev": "c7481c8249"
"BuildInformation": {"<redacted>.1644873069.c7481c8249.<branch>.4": 1}

After Change:
"BuildGitRev": "1d878d8a0be09b678bb4490060e8f23f890ddb92"
"BuildInformation": {"<redacted>.1648062731.1d878d8a0be09b678bb4490060e8f23f890ddb92.<branch>.48": 1},
```
An alternate is to not change the existing `BuildGitRev` but introduce a new variable `BuildGitFullRev`. I thought that would be redundant and hence updating the existing value to be complete

## Related Issue(s)
Fixes https://github.com/vitessio/vitess/issues/9952

## Checklist
- [no] Should this PR be backported?
- [updated] Tests were added or are not required
- [ no] Documentation was added or is not required

## Deployment Notes
na

### DIFF
--- a/go/vt/servenv/buildinfo_test.go
+++ b/go/vt/servenv/buildinfo_test.go
@@ -31,7 +31,7 @@ func TestVersionString(t *testing.T) {
 		buildUser:       "user",
 		buildTime:       now.Unix(),
 		buildTimePretty: "time is now",
-		buildGitRev:     "d54b87c",
+		buildGitRev:     "d54b87ca0be09b678bb4490060e8f23f890ddb92",
 		buildGitBranch:  "gitBranch",
 		goVersion:       "1.18",
 		goOS:            "amiga",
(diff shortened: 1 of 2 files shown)


malpani:
@deepthi @ajm188 could you please review this minor PR

deepthi:
We want to be sure not to break anyone depending on this, that may take a bit more time.
https://vitess.slack.com/archives/CMDJ2KFEZ/p1648139160472689

malpani:
> We want to be sure not to break anyone depending on this, that may take a bit more time. 

@deepthi Makes sense. In that case, would it be safer and faster to approve if i introduced a new variable called `BuildGitFullRev` instead, and leaving the current `BuildGitRev` as is so that in case anyone out there might be relying on the short string is not affected?

> https://vitess.slack.com/archives/CMDJ2KFEZ/p1648139160472689

I dont have access to that slack org. Is there a external facing channel where i can chime in if it helps

deepthi:
> @deepthi Makes sense. In that case, would it be safer and faster to approve if i introduced a new variable called `BuildGitFullRev` instead, and leaving the current `BuildGitRev` as is so that in case anyone out there might be relying on the short string is not affected?
> 

That is an option, but what would we print on `--version`? Would we continue to print `BuildGitRev` or change it to `BuildGitFullRev`? If the latter, then user-facing behavior changes anyway. If at all possible, we might as well keep one variable instead of two.

> I dont have access to that slack org. Is there a external facing channel where i can chime in if it helps.

This is the public vitess slack workspace. You can join by clicking the invite link (green slack icon) at the top of the website (vitess.io)

malpani:
thanks @deepthi ! I was able to join the slack community and based on the discussion there, as long as there no concerns raised by eod, the plan is to merge this in.

deepthi (approved):
We did not get any responses from the community, so assuming this is safe to change.

Issue #9419, opened 2021-12-17
[vtadmin-api] Allow RBAC config file to be optional


doeg:
The `--rbac-config` flag is required when running vtadmin-api. If omitted:

```
F1217 12:14:01.251523   31578 main.go:72] open rbac.yaml: no such file or directory
```

The local example provides a default rbac.yaml file here: https://github.com/vitessio/vitess/blob/main/examples/local/vtadmin/rbac.yaml 

However, it might be more ergonomic + explicit if Vitess operators could _either_:
- Provide an `--rbac` flag and an `--rbac-config` file, or
- Provide a `--no-rbac` (or `--rbac=false`) flag to explicitly opt out of using RBAC.

setassociative:
I was reading the related backscroll with [jakon](https://vitess.slack.com/archives/C01H307F68J/p1639611736036300) and had the thought that defaulting allow-all when no rbac config flag is convenient but allows failure in an unsafe way: I can imagine fucking up a chef flag -> not passing an rbac flag -> opening write actions to the world.

Still extremely into easy allow-all option but we should make it opt-in under `--no-rbac` or similar.

notfelineit:
A clarifying q, does that mean we're either saying 1) must pass in rbac config or 2) must explicitly pass --no-rbac to default to all?

jakon89:
@setassociative +1 for `--no-rbac` 
Anyway, unknown flags aren't allowed so its unlikely that scenario is possible:
```
adam@local ~ % docker run vtadmin:latest --rbac-confi /non-exisiting-file
ERROR: logging before flag.Parse: E1217 18:05:40.620713       1 syslogger.go:149] can't connect to syslog
Error: unknown flag: --rbac-confi
Usage:
  vtadmin [flags]
```

setassociative:
@notfelineit yea, that's my suggestion

@jakon89 so when constructing args to set up vtadmin there is the error where you typo the arg and it's safe as you point out. I'm more worried about a scenario where it gets omitted entirely for whatever reason and I'd rather fail hard and fast in that situation

doeg:
This sounds great. I updated the issue description to specify an explicit `--no-rbac` flag (instead of "failing open" if an RBAC config file is not provided).

doeg:
Also what about `--rbac` and `--rbac=false`? 🤔 Not sure if Vitess has a convention for boolean flags.

notfelineit:
[...]

PR #9972, opened 2022-03-24
Allow --no-rbac flag that allows users to not pass rbac config

notfelineit:

## Description
**Breaking Change**
This PR adds a flag `--no-rbac` that allows users to bypass passing the flag `--rbac-config`. In doing so, a default rbac configuration is used:
```
rules:
  - resource: "*"
    actions:
    - "get"
    - "create"
    - "delete"
    - "put"
    - "ping"
    subjects: ["*"]
    clusters: ["*"]
```

It also makes it so that a user must **explicitly** pass in either `--no-rbac` or `--rbac`, otherwise vtadmin will fail to come up:
Screen Shot 2022-03-23 at 4 59 44 PM

This is a **breaking change** since users who were using `--rbac-config` must now explicitly pass in `--rbac` flag as well.

## Cases
`--no-rbac`:
Screen Shot 2022-03-23 at 5 01 08 PM

`--rbac`:
Screen Shot 2022-03-23 at 4 59 24 PM

Failing to pass in either flag:
Screen Shot 2022-03-23 at 4 59 44 PM

## Related Issue(s)
This addresses this issue: https://github.com/vitessio/vitess/issues/9419

## Checklist
- [x] Documentation was added or is not required **Documentation should be added to release notes**

## Deployment Notes

N/A


go/cmd/vtadmin/main.go
```diff
@@ -163,6 +171,8 @@ func main() {
 
 	// rbac flags
 	rootCmd.Flags().StringVar(&rbacConfigPath, "rbac-config", "rbac.yaml", "")
+	rootCmd.Flags().BoolVar(&enableRbac, "rbac", false, "whether to enable rbac")
+	rootCmd.Flags().BoolVar(&disableRbac, "no-rbac", false, "whether to disable rbac")
```
ajm188, line 175: i'm not sure which is better; i'm curious what you think about this!

```go
rootCmd.Flags().StringVar(&rbacConfigPath, "rbac-config", "", "path to an RBAC config file. must be set if not pasing --no-rbac") // now defaulting to empty
rootCmd.Flags().BoolVar(&disableRBAC, "no-rbac", false, "whether to disable RBAC. must be set if not passing --rbac-config")
```

and then in the run func:

```go
if rbacConfigPath == "" {
	if !disableRBAC {
		// fail, must pass exactly one
	}

	rbacConfig = rbac.DefaultConfig()
} else if disableRBAC {
	// fail, must pass exactly one
} else {
	cfg, err := rbac.LoadConfig(rbacConfigPath)
	if err != nil {
		fatal(err)
	}

	rbacConfig = cfg
}
```

go/vt/vtadmin/rbac/config.go
```diff
@@ -163,3 +163,41 @@ func (c *Config) GetAuthenticator() Authenticator {
 func (c *Config) GetAuthorizer() *Authorizer {
 	return c.authorizer
 }
+
+func DefaultConfig() *Config {
+	log.Info("[rbac]: using default rbac configuration")
+	actions := []string{"get", "create", "delete", "put", "ping"}
```
ajm188, line 169: we should use the exported constants here

go/vt/vtadmin/rbac/config.go
```diff
@@ -163,3 +163,41 @@ func (c *Config) GetAuthenticator() Authenticator {
 func (c *Config) GetAuthorizer() *Authorizer {
 	return c.authorizer
 }
+
+func DefaultConfig() *Config {
```
ajm188, line 167: this needs a doc comment

go/cmd/vtadmin/main.go
```diff
@@ -43,6 +43,8 @@ var (
 	defaultClusterConfig cluster.Config
 
 	rbacConfigPath string
+	enableRbac     bool
+	disableRbac    bool
```
ajm188, line 47: per [Go guidelines](https://zchee.github.io/golang-wiki/CodeReviewComments/#initialisms), these should be written as "{enable,disable}RBAC" (fwiw, the reason `rbacConfigPath` is written that way is to not export it, and "rBAC" would be also incorrect)

ajm188 (commented):
all i've got are style commentary!

go/cmd/vtadmin/main.go
notfelineit: I like removing the default for the rbac config. I guess the other difference between this and the other is that the current implemention requires explicitly passing an "enable rbac" flag, specified by the issue: https://github.com/vitessio/vitess/issues/9419. I personally prefer explicitly enabling rbac per the issue 😄

ajm188 (commented):
one more round 😅

go/cmd/vtadmin/main.go
```diff
@@ -170,9 +170,9 @@ func main() {
 	rootCmd.Flags().BoolVar(&httpOpts.EnableDynamicClusters, "http-enable-dynamic-clusters", false, "whether to enable dynamic clusters that are set by request header cookies")
 
 	// rbac flags
-	rootCmd.Flags().StringVar(&rbacConfigPath, "rbac-config", "rbac.yaml", "")
-	rootCmd.Flags().BoolVar(&enableRbac, "rbac", false, "whether to enable rbac")
-	rootCmd.Flags().BoolVar(&disableRbac, "no-rbac", false, "whether to disable rbac")
+	rootCmd.Flags().StringVar(&rbacConfigPath, "rbac-config", "", "path to an RBAC config file. must be set if passing --rbac")
+	rootCmd.Flags().BoolVar(&enableRBAC, "rbac", false, "whether to enable rbac. must be set if not passing --rbac")
+	rootCmd.Flags().BoolVar(&disableRBAC, "no-rbac", false, "whether to disable rbac. must be set if not passing --no-rbac")
```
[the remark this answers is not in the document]
notfelineit: Whoo! Good idea

go/cmd/vtadmin/main.go
```diff
@@ -103,19 +103,19 @@ func run(cmd *cobra.Command, args []string) {
...
 		if err != nil {
 			fatal(err)
 		}
 
 		rbacConfig = cfg
-	}
-	if disableRbac && !enableRbac {
-		rbacConfig = rbac.DefaultConfig()
+	} else if enableRBAC && rbacConfigPath == "" {
+		fatal("[rbac] must pass --rbac-config path when enabling rbac")
```
notfelineit, line 116:
```suggestion
		fatal("must pass --rbac-config path when enabling rbac")
```

go/cmd/vtadmin/main.go
```diff
@@ -103,19 +103,19 @@ func run(cmd *cobra.Command, args []string) {
...
+	} else {
+		fatal("[rbac] must explicitly enable or disable RBAC by passing --no-rbac or --rbac")
```
notfelineit, line 118:
```suggestion
		fatal("must explicitly enable or disable RBAC by passing --no-rbac or --rbac")
```

PR #9974, opened 2022-03-24
Move VTOrc tests to run against MySQL 8.0

GuptaManan100:
## Description

This PR moves the VTOrc tests to run against MySQL 8.0 instead of MySQL 5.7

## Checklist

### DIFF
--- a/test/config.json
+++ b/test/config.json
@@ -1130,6 +1130,33 @@
 			"RetryMax": 3,
 			"Tags": []
 		},
+		"vtorc_primary_failure_8.0": {
+			"File": "unused.go",
+			"Args": ["vitess.io/vitess/go/test/endtoend/vtorc/primaryfailure"],
+			"Command": [],
+			"Manual": false,
+			"Shard": "vtorc_8.0",
+			"RetryMax": 3,
+			"Tags": []
+		},
+		"vtorc_graceful_takeover_8.0": {
+			"File": "unused.go",
+			"Args": ["vitess.io/vitess/go/test/endtoend/vtorc/gracefultakeover"],
+			"Command": [],
+			"Manual": false,
+			"Shard": "vtorc_8.0",
+			"RetryMax": 3,
+			"Tags": []
+		},
+		"vtorc_general_8.0": {
+			"File": "unused.go",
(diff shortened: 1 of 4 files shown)


vmg (approved):
Strong :+1:. We need more stuff running in 8.0 because the 5.7 EOL is fast approaching.

deepthi:
I'm conflicted about whether we can skip running this on 5.7 yet. How about running it on both?

GuptaManan100:
I am conflicted too, at one hand we want to test both versions, on the other hand having tests for both versions is quickly going to go out of hand from the CI perspective. We only have 60 concurrent runners and expanding our test suite will reduce the velocity with which we can work.
I don't know what the best way forward is. I think the ideal situation would be have unit tests for these scenarios which we can run on both the versions of MySQL and then have the end to end tests run on either one version. WDYT?

deepthi:
For now let us run this on both. This is the _only_ vtorc endtoend test, unlike other components where we have tons of endtoend tests and we could possibly choose to split them between 5.7 and 8.0.

GuptaManan100:
@deepthi you got it. I'll make the changes

test/config.json
```diff
@@ -1130,6 +1130,33 @@
...
+		},
+		"vtorc_general_8.0": {
+			"File": "unused.go",
+			"Args": ["vitess.io/vitess/go/test/endtoend/vtorc/general"],
+			"Command": [],
+			"Manual": false,
+			"Shard": "vtorc_8.0",
+			"RetryMax": 3,
+			"Tags": []
+		},
```
GuptaManan100, line 1159: I can avoid this duplication of same values of _8.0 by adding code to workflow generator where we remove `_8.0` suffix from the name of the test to find the shard to follow. This would mean that `vtorc_8.0` and `vtorc` will both have the command to follow the shard `vtorc` in the workflow and then we can remove this duplicate code.

I don't have a specific preference between the two, if you feel the other alternate is more readable/better, I'll make the change.

deepthi: This is fine. Naming the test clearly helps in debugging when there are failures in CI.

PR #9975, opened 2022-03-24
Skip end-to-end test if binary is below a certain major version

frouioui:
## Description

This pull request adds a new helper function to the `endtoend/utils` package that skips a test if the major version of a binary is below a certain increment.

Some new functionalities can only be tested from a certain version of `vtgate`, `vttablet`, etc. However, the upgrade-downgrade tests execute all the end-to-end tests of a package regardless of which binary version supports them. This new helper function alleviates this issue.

Recently, the `TestSubqueriesExists` test started failing in the upgrade-downgrade test suite. This test uses a vtgate feature that only exists in `>= 14`, thus the new helper function is called at the beginning of the test to skip it if vtgate is below `v14`.

## Checklist

### DIFF
--- a/go/test/endtoend/cluster/cluster_process.go
+++ b/go/test/endtoend/cluster/cluster_process.go
@@ -364,7 +364,7 @@ func (cluster *LocalProcessCluster) StartKeyspace(keyspace Keyspace, shardNames 
 		}
 	}
 
-	//Apply VSchema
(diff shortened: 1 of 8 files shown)


GuptaManan100 (approved):

shlomi-noach (approved):
:heart:

Issue #9976, opened 2022-03-24
Bug Report: change streaming crash when streaming from position older then mysql restart


aliulis:
### Overview of the Issue

Change streaming crash when streaming from position older then Mysql restart. This happen, because Mysql restart gives new id to tables in binlog. Vitess cache table schema based on id (https://github.com/vitessio/vitess/blob/main/go/vt/vttablet/tabletserver/vstreamer/vstreamer.go#L604). If same id is assigned to different table, then Vitess tries to process data using old cached and most probably crash or send incorrect data.

Example error:
```
rpc error: code = Unknown desc = vttablet: rpc error: code = Unknown desc = stream (at source tablet) error @ 2cea1ce3-ab5a-11ec-87b8-0242ac110002:1-48: index out of range, colExpr.ColNum: 2, len(values): 2
```

### Reproduction Steps

For reproduction used https://vitess.io/docs/14.0/get-started/local-docker/:

Save current GTID:
```
mysql> show global variables like "gtid_executed";
+---------------+-------------------------------------------+
| Variable_name | Value                                     |
+---------------+-------------------------------------------+
| gtid_executed | 2cea1ce3-ab5a-11ec-87b8-0242ac110002:1-40 |
+---------------+-------------------------------------------+
1 row in set (0.00 sec)
```
Insert data in all tables:
```
mysql> insert into customer (email) values ("test@email.com");
Query OK, 1 row affected (0.02 sec)

mysql> insert into corder (customer_id,price) values (4571,458777);
Query OK, 1 row affected (0.01 sec)

mysql> insert into product (sku, description,price) values ("skuskuskus","something intresting",878787);
Query OK, 1 row affected (0.01 sec)
```
Check table id in binlogs in RDONLY mysql:
```
vitess@bd98ffb5e0c0:/vt/vtdataroot/vt_0000000102/bin-logs$ mysqlbinlog --verbose --base64-output=decode-rows vt-0000000102-bin.000001 | grep Table_map | awk '{print $11 " " $15}' | sort | uniq | sort -n -k 2
`_vt`.`reparent_journal` 114
`vt_commerce`.`product` 115
`vt_commerce`.`customer` 116
`vt_commerce`.`corder` 117
```
Restart RDONLY mysql.

Insert something preferably in different order:
```
mysql> insert into product (sku, description,price) values ("skuskuskus11111","something intresting111111",878787878777);
Query OK, 1 row affected (0.01 sec)

[...]

PR #9978, opened 2022-03-24
VStreamer: recompute table plan if a new table is encountered for the same id

rohit-nayak-ps:
## Description

VStreamer caches a "table plan" for each new table it encounters via the binlog TABLE_MAP_EVENT.  

MySQL internally generates a unique (incrementing) id for a table as it inserts them into the binlog. The first time we see an id, we compute the table plan (essentially the list of columns with name and type information). If we get the same id again, we currently assume it is the same table and use the cached plan. 

However when the mysql server restarts it starts a new numbering for table ids and depending on the DMLs encountered the table id map can be different from the earlier one. 

Thus the binlog will contain the same id for different tables. While streaming from an older GTID, it can happen that we have cached a plan for table1 for id 100, say. And later we get the same id 100 for table2. Usually this will result in an error because the fields for the table will not match the row image. If a vstreamer encounters an error it just restarts after 5 seconds, when it recreates the plans and heals itself.

But in rare situations where the schemas are almost identical we can generate unrelated FieldEvents and consequently a DML for the wrong table.

This PR adds a check before using a cached plan. If the table name for the cached plan doesn't match that for the table map event, it nullifies that cache entry and recomputes the plan for this id.

## Related Issue(s)

#9976 

## Checklist

### DIFF
--- a/go/vt/vttablet/tabletserver/vstreamer/vstreamer.go
+++ b/go/vt/vttablet/tabletserver/vstreamer/vstreamer.go
@@ -601,14 +601,23 @@ func (vs *vstreamer) parseEvent(ev mysql.BinlogEvent) ([]*binlogdatapb.VEvent, e
 		// will generate a new plan and FIELD event.
 		id := ev.TableID(vs.format)
 
-		if _, ok := vs.plans[id]; ok {
-			return nil, nil
-		}
-
 		tm, err := ev.TableMap(vs.format)
 		if err != nil {
 			return nil, err
 		}
+
+		if plan, ok := vs.plans[id]; ok {
+			// When the underlying mysql server restarts the table map can change.
+			// Usually the vstreamer will also error out when this happens, and vstreamer re-initializes its table map.
+			// But if the vstreamer is not aware of the restart, we could get an id that matches one in the cache, but
+			// is for a different table. We then invalidate and recompute the plan for this id.
+			if plan == nil || plan.Table.Name == tm.Name {
+				return nil, nil
+			}
+			vs.plans[id] = nil
+			log.Infof("table map changed: id %d for %s has changed to %s", id, plan.Table.Name, tm.Name)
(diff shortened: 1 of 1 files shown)


go/vt/vttablet/tabletserver/vstreamer/vstreamer.go
```diff
@@ -601,14 +601,23 @@ func (vs *vstreamer) parseEvent(ev mysql.BinlogEvent) ([]*binlogdatapb.VEvent, e
...
+		if plan, ok := vs.plans[id]; ok {
+			// When the underlying mysql server restarts the table map can change.
+			// Usually the vstreamer will also error out when this happens, and vstreamer re-initializes its table map.
+			// But if the vstreamer is not aware of the restart, we could get an id that matches one in the cache, but
+			// is for a different table. We then invalidate and recompute the plan for this id.
+			if plan == nil || plan.Table.Name == tm.Name {
+				return nil, nil
+			}
+			vs.plans[id] = nil
+			log.Infof("table map changed: id %d for %s has changed to %s", id, plan.Table.Name, tm.Name)
```
mattlord, line 618: Maybe also note that we will reload and cache a new plan for the table to avoid concern and indicate corrective action being taken ~:
```
log.Infof("Table plan map changed in VStream: the id %d was associated with %s and now represents %s. We will reload and cache a new plan for the table.", id, plan.Table.Name, tm.Name)
```

mattlord (approved):
Just a small nit about the log message (feel free to ignore). 

Thank you!

go/vt/vttablet/tabletserver/vstreamer/vstreamer.go
rohit-nayak-ps: Good suggestion, I will update in my next PR though, rather than start another CI cycle for this ...

PR #9979, opened 2022-03-25
fix: make concatenate and limit concurrent safe

### LINKED ISSUE #9945: Flaky union tests (vtgate_queries -> vtgate_queries_aggregation)
### Overview of the Issue

Union tests are among the most flaky in the test suite.
union

These are found here: https://github.com/vitessio/vitess/blob/main/go/test/endtoend/vtgate/queries/union/union_test.go

And are included in the `vtgate_queries_aggregation` test which is part of the `vtgate_queries` shard/suite CI job: https://github.com/vitessio/vitess/blob/8d4e8c3a301f6517fcc7b27e1bcd797c2389ef34/test/config.json#L654-L662

### Reproduction Steps

As seen in CI / Launchable.
https://app.launchableinc.com/organizations/vitess/workspaces/vitess-app/flaky-tests

### Binary Version

```sh
main
```

```sh
GH CI Actions
```

### Log Fragments

_No response_

systay:
## Description
Some of our end-to-end tests for UNION were flaky and failing intermittently. These changes make the concatenate and the limit operator more resilient to running concurrent queries.

## Related Issue(s)
Fixes #9945

### DIFF
--- a/go/vt/vtgate/engine/concatenate.go
+++ b/go/vt/vtgate/engine/concatenate.go
@@ -19,8 +19,6 @@ package engine
 import (
 	"sync"
 
-	"vitess.io/vitess/go/sync2"
-
 	"vitess.io/vitess/go/sqltypes"
 	querypb "vitess.io/vitess/go/vt/proto/query"
 	vtrpcpb "vitess.io/vitess/go/vt/proto/vtrpc"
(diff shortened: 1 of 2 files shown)


frouioui (approved):
I thought `AtomicBool` was able to handle this kind of concurrency, can you explain why using a boolean+mutex is better?
LGTM!

GuptaManan100:
atomic boolean only guarantees that the boolean will be read or written to by a processor, but we want to ensure that no two processes ever write to the same boolean twice, i.e. we want to prevent two process from running some part of the code altogether. This needs a mutex

GuptaManan100 (approved):

frouioui:
got it

harshit-gangal (approved):

PR #9983, opened 2022-03-25
Add gitpod support

DAlperin:
Signed-off-by: DAlperin <dzalperin@gmail.com>

## Description
I have begun to rely on vitess but as a student I do not have consistent access to network and compute at a level that makes it convenient to build vitess locally. In these situations I often turn to gitpod.io to work on a remote machine. This PR just easily allows for others to spin up a working gitpod dev environment.

## Checklist


vmg:
Hey @DAlperin! Thanks for your contribution! I've been reviewing the container definitions you've submitted and I have concerns about the long term viability of maintaining this long-term. GitPod is a commercial product that nobody on the Vitess maintainers team really uses, so we don't have any insights on whether these definitions are correct or not, and no incentive or ability to maintain these scripts long term.

My advice, if you're interested on developing Vitess remotely, would be to focus on a `.devcontainer/devcontainer.json` file, which as far as I can see is a much more well established OSS standard, which should work on the open-source VSCode editor and several commercial products (from cursory reading, even GitPod, with some small hassle). If you'd be interested on contributing that to the Vitess project, we'd happily merge it and maintain it long term, because I believe that'd be much more generally interesting to the community.

But for now we won't accept configurations that only work on a single proprietary platform. Sorry!

PR #9987, opened 2022-03-26
Partial Movetables: allow moving a table one shard at a time

rohit-nayak-ps:
> **Warning!&nbsp;**
> This feature is expected to be used only for migrating large Vitess clusters across data centers and should not be enabled on an ongoing basis since it adds performance overhead during query serving. It also changes the paradigm hitherto expected in Vitess that a table is served, at any given time, from a single keyspace. The latter might lead to unstable behavior, with global routing, for example.

### TL;DR;

* MoveTables can now take a subset of shards
* ShardRoutingRules have been added, to route shards already moved from source to target
* ShardRoutingRules apply both for shard targeted and globally routed queries

## Description

This feature introduces the concept of partial keyspaces where some shards are served from a different keyspace. This is
useful for a specific but critical use-case where a large production Vitess setup (100s of shards) is being migrated to
a new data center or provider. Migrating the entire cluster in one go using `MoveTables` could cause an unacceptable
downtime due to the large number of primaries that need to be synced when writes are switched.

#### _partial_ MoveTables signalled by `--source_shards`

`vtctlclient MoveTables -- -source customer --tables 'customer,corder' --source_shards '-80' Create customer2.partial1`

#### VDiff works as-is

`vtctlclient VDiff customer2.partial1`

#### SwitchTraffic generates this shard routing rule
`vtctlclient MoveTables -- SwitchTraffic customer2.partial1`

`{"rules":[{"from_keyspace":"customer", "to_keyspace":"customer2", "shards":"-80"}]}`

#### Demo that shard routing now works

```
# stop workflow to stop reverse replication from running 
vtctlclient Workflow customer2.partial stop
# update customer2 database directly.
mysql -S $VTDATAROOT/vt_0000000500/mysql.sock -u vt_dba -e "update vt_customer2.customer set email = concat('routed.', email)"

# use shard targeting using customer (not customer2). 
mysql -e "use customer:-80; select customer_id, email from customer order by customer_id"

+-------------+---------------------------+
| customer_id | email                     |
+-------------+---------------------------+
|           1 | routed.alice@domain.com   |
|           2 | routed.bob@domain.com     |
|           3 | routed.charlie@domain.com |
|           5 | routed.eve@domain.com     |
+-------------+---------------------------+

```

# Summary of code changes

### Workflow Show changes
While we had the possibility of partial reads being switched earlier, now writes can also be partially switched in a workflow.

#### Topo 
Shard Routing Rules are a new concept introduced for this feature. It maps a (keyspace, shard) tuple to another keyspace. This is used to create a new cluster level map which maps a (keyspace, shard) to another keyspace. These are set by SwitchTraffic in a partial MoveTables and  used by vtgate while routing queries. vtctlclient commands `ApplyShardRoutingRules` and `GetShardRoutingRules`allow setting/getting of these rules.

#### vtgate Shard Targeting
The shard targeted query routing in vtgate's bypass mechanism `go/vt/vtgate/planbuilder/bypass.go`. We create a map from the SrvVSchema's shard routing rules object and check if a specified shard destination needs to be rerouted.

#### vtgate Global Routing
The global query routing using vtgate's ResolveDestinations() methods `go/vt/vtgate/vcursor_impl.go`. We go through all selected shard destinations and modify those that are mapped in the shard routing rules.

### 

### New column workflow_sub_type in _vt.vreplication
There is a new bool column `workflow_sub_type` added to `_vt.vreplication`, set for partial movetables. It is used for
visibility and for bypassing certain validations that expect a full keyspace.

#### vtgate --enable_shard_routing
Default: false. It is used when a cluster is setup with shard routing rules and tells vtgate to use these rules while routing queries.

#### MoveTables --source_shards
The only flag needed to tell MoveTables to perform a partial movetables is to pass it this flag, example `--source_shards -80`. This flag already exists and is used by Reshard.

## Notes

Both read and write traffic is switched at the same time when a shard routing is deemed complete (using SwitchTraffic). This is because we add a shard routing rule when this happens. Switching of read and write separately is done by updating the regular routing rules by targeting @replica and @rdonly.

#### e2e test TestPartialMoveTables
It creates a workflow which moves tables from one shard only to the target shard, ensures it completed correctly and switches traffic. It ensures that the shard routing rules and regular routing rules are setup correctly and that the vtgate queries both shard targeted and global queries are routed correctly.

#### Modified vtgate tests
A new set of tests have been added where the test cluster is setup as a partial cluster: one shard is in a different keyspace with the ShardRoutingRules setup. The main change is that the vtgate params need to specify the default keyspace as the `DbName` to avoid ambiguity.

Some tests have to be skipped for partial keyspaces because they intrinsically affect tablets in the base cluster which are now not all serving. 

TODOs:
[ ] Add explicit reasons for Skipped vtgate unit tests during partial keyspace

## Checklist


test.go
```diff
@@ -174,6 +175,9 @@ func (t *Test) run(dir, dataDir string) ([]byte, error) {
 			testCmd = []string{"test/" + t.File, "-v", "--skip-build", "--keep-logs"}
 			testCmd = append(testCmd, t.Args...)
 		}
+		if *partialKeyspace {
+			testCmd = append(testCmd, "-partial-keyspace")
```
mattlord, line 179: Need to use double dashes (single dash will be unsupported soon).

mattlord (commented):
Looks great! Nice work! 

I only had some minor requests, comments, questions.

go/test/endtoend/vtgate/queries/insert/insert_test.go
```diff
@@ -58,6 +58,9 @@ func start(t *testing.T) (utils.MySQLCompare, func()) {
 }
 
 func TestSimpleInsertSelect(t *testing.T) {
+	if clusterInstance.HasPartialKeyspaces {
+		t.Skip("don't run on partial keyspaces")
```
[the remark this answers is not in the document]
harshit-gangal: I think the skip message can be more clear as to why this particular test is skipped.
So, I would say that each skip message should be explain the reason for skipping that test.

go/test/endtoend/vtgate/errors_as_warnings/main_test.go
```diff
@@ -93,16 +93,14 @@ func TestMain(m *testing.M) {
...
-		}
+		vtParams = clusterInstance.GetVTParams(KeyspaceName)
 		return m.Run()
 	}()
 	os.Exit(exitCode)
 }
 
 func TestScatterErrsAsWarns(t *testing.T) {
+	t.Skip("partial keyspace: this test kills primary on source shard, but query will be on target shard")
 	oltp, err := mysql.Connect(context.Background(), &vtParams)
```
harshit-gangal, line 104: we should skip only in case of partial_keyspace testing.

go/test/endtoend/vtgate/errors_as_warnings/main_test.go
```diff
@@ -136,7 +134,7 @@ func TestScatterErrsAsWarns(t *testing.T) {
 	for _, mode := range modes {
 		t.Run(mode.m, func(t *testing.T) {
 			// connection setup
-			utils.Exec(t, mode.conn, "use @replica")
+			utils.Exec(t, mode.conn, "use ks@replica")
```
harshit-gangal, line 137: as we will be skipping the test for partial_keyspace, this change would not be needed.

go/test/endtoend/vtgate/prefixfanout/main_test.go
```diff
@@ -156,12 +156,10 @@ func TestMain(m *testing.M) {
 }
 
 func TestCFCPrefixQueryNoHash(t *testing.T) {
+	//t.Skip("partial keyspace")
```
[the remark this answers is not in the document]
harshit-gangal: all these should be removed in this file

go/test/endtoend/vtgate/queries/informationschema/informationschema_test.go
```diff
@@ -57,6 +57,9 @@ func start(t *testing.T) (utils.MySQLCompare, func()) {
 }
 
 func TestDbNameOverride(t *testing.T) {
+	if clusterInstance.HasPartialKeyspaces {
+		t.Skip("partial keyspace")
```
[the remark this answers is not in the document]
harshit-gangal: adding reason for skipping will help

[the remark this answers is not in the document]
harshit-gangal: also there are other places as well in this file

go/vt/vtgate/vtgate.go
```diff
@@ -303,7 +306,7 @@ func Init(
 	}
 
 	initAPI(gw.hc)
-
+	planbuilder.EnableShardRoutingFlag(*EnableShardRouting)
```
harshit-gangal, line 309: I think we should not set this like this.
instead in planbuilder we can call into vcusor to check if the flag is enabled or disabled.

go/vt/vtgate/vcursor_impl.go
```diff
@@ -573,12 +573,51 @@ func (vc *vcursorImpl) setRollbackOnPartialExecIfRequired(atleastOneSuccess bool
 	}
 }
 
+// fixupPartiallyMovedShards checks if any of the shards in the route has a ShardRoutingRule (true when a keyspace
+// is in the middle of being moved to another keyspace using MoveTables moving one shard at a time
```
harshit-gangal, line 577:
```suggestion
// is in the middle of being moved to another keyspace using MoveTables moving subset of shards at a time
```

go/vt/topotools/routing_rules_test.go
```diff
@@ -66,3 +66,21 @@ func TestRoutingRulesErrors(t *testing.T) {
...
+	ctx := context.Background()
+	ts := memorytopo.NewServer("zone1")
+
+	srr := map[string]string{
+		"ks1.shard1": "ks2",
+		"ks3.shard2": "ks4",
+	}
+
+	err := SaveShardRoutingRules(ctx, ts, srr)
+	require.NoError(t, err, "could not save shard routing rules to topo %v", srr)
```
mattlord, line 80: Should be using err and not srr:
```
require.NoError(t, err, "could not save shard routing rules to topo: %v", err)
```

go/vt/topotools/routing_rules_test.go
```diff
@@ -66,3 +66,21 @@ func TestRoutingRulesErrors(t *testing.T) {
...
+
+	roundtripRules, err := GetShardRoutingRules(ctx, ts)
+	require.NoError(t, err, "could not fetch shard routing rules from topo")
```
mattlord, line 83:
```
require.NoError(t, err, "could not fetch shard routing rules from topo: %v", err)
```

go/vt/vtctl/workflow/stream_migrator.go
```diff
@@ -564,7 +573,8 @@ func (sm *StreamMigrator) createTargetStreams(ctx context.Context, tmpl []*VRepl
 				rule.Filter = buf.String()
 			}
 
-			ig.AddRow(vrs.Workflow, vrs.BinlogSource, mysql.EncodePosition(vrs.Position), "", "")
+			ig.AddRow(vrs.Workflow, vrs.BinlogSource, mysql.EncodePosition(vrs.Position), "", "",
+				int64(int(vrs.WorkflowType)), int64(int(vrs.WorkflowSubType)))
```
mattlord, line 577:
```
int64(vrs.WorkflowType), int64(vrs.WorkflowSubType))
```

go/vt/wrangler/materializer.go
```diff
@@ -941,22 +941,50 @@ func (wr *Wrangler) buildMaterializer(ctx context.Context, ms *vtctldatapb.Mater
...
+		var sourceShards2 []*topo.ShardInfo
+		for _, shard := range sourceShards {
+			for _, shard2 := range ms.SourceShards {
+				if shard.ShardName() == shard2 {
+					sourceShards2 = append(sourceShards2, shard)
+					break
+				}
+			}
+		}
+		sourceShards = sourceShards2
```
mattlord, line 960: If sourceShards2 is empty here... is that an error or does it just mean that it's not in fact partial?

go/vt/wrangler/materializer.go
```diff
@@ -941,22 +941,50 @@ func (wr *Wrangler) buildMaterializer(ctx context.Context, ms *vtctldatapb.Mater
...
+		var targetShards2 []*topo.ShardInfo
+		for _, shard := range targetShards {
+			for _, shard2 := range ms.SourceShards {
+				if shard.ShardName() == shard2 {
+					targetShards2 = append(targetShards2, shard)
+					break
+				}
+			}
+		}
+		targetShards = targetShards2
```
mattlord, line 976: Same question here for targetShards2

go/vt/wrangler/materializer.go
```diff
@@ -941,22 +941,50 @@ func (wr *Wrangler) buildMaterializer(ctx context.Context, ms *vtctldatapb.Mater
...
+	}
+	if len(sourceShards) == 0 || len(targetShards) == 0 {
+		return nil, fmt.Errorf("no source and/or target shards specified for workflow %s ", ms.Workflow)
+	}
```
mattlord, line 980: Ah, this answers my previous two questions :). IMO it would be better to error for each one as we return sooner (do less work) and we can provide a more specific error.

go/vt/wrangler/traffic_switcher.go
```diff
@@ -1328,20 +1446,39 @@ func (ts *trafficSwitcher) changeRouting(ctx context.Context) error {
...
-	if err := topotools.SaveRoutingRules(ctx, ts.TopoServer(), rules); err != nil {
-		return err
+	if ts.isPartialMigration {
+		srr, err := topotools.GetShardRoutingRules(ctx, ts.TopoServer())
+		if err != nil {
+			return err
+		}
+		for _, si := range ts.SourceShards() {
+			delete(srr, fmt.Sprintf("%s.%s", ts.TargetKeyspaceName(), si.ShardName()))
+			ts.Logger().Infof("Delete shard routing: %v:%v", ts.TargetKeyspaceName(), si.ShardName())
```
mattlord, line 1456: The log messages should be past tense, `Deleted`. Same for the one below: `Added`

go/vt/wrangler/traffic_switcher.go
```diff
@@ -1328,20 +1446,39 @@ func (ts *trafficSwitcher) changeRouting(ctx context.Context) error {
...
+			return err
+		}
+		for _, table := range ts.Tables() {
+			targetKsTable := fmt.Sprintf("%s.%s", ts.TargetKeyspaceName(), table)
+			sourceKsTable := fmt.Sprintf("%s.%s", ts.SourceKeyspaceName(), table)
+			delete(rules, targetKsTable)
+			ts.Logger().Infof("Delete routing: %s", targetKsTable)
+			rules[table] = []string{targetKsTable}
+			rules[sourceKsTable] = []string{targetKsTable}
+			ts.Logger().Infof("Add routing: %v %v", table, sourceKsTable)
```
mattlord, line 1475: Another one that should be past tense

mattlord (commented):
Looks good! I had some minor questions/comments/suggestions. We also have some unresolved comments from @harshit-gangal to resolve in one way or another. I can approve quickly once we go through those things and resolve them. Thanks! ❤️

go/test/endtoend/vtgate/misc_test.go
```diff
@@ -395,6 +395,9 @@ func TestDeleteAlias(t *testing.T) {
 }
 
 func TestFunctionInDefault(t *testing.T) {
+	if clusterInstance.HasPartialKeyspaces {
+		t.Skip("TODO: why is this failing?")
```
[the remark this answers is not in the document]
rohit-nayak-ps: done

go/vt/topotools/routing_rules.go
```diff
@@ -56,3 +58,39 @@ func SaveRoutingRules(ctx context.Context, ts *topo.Server, rules map[string][]s
 
 	return ts.SaveRoutingRules(ctx, rrs)
 }
+
+// GetShardRoutingRules fetches shard routing rules from the topology server and returns a
+// mapping of fromKeyspace.Shard=>toKeyspace.
+func GetShardRoutingRules(ctx context.Context, ts *topo.Server) (map[string]string, error) {
+	rrs, err := ts.GetShardRoutingRules(ctx)
```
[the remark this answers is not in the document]
rohit-nayak-ps: The topo server is passed down by the traffic switcher which has already referred to it, so if it is nil it should have been already reported (or panicked!) before it comes here.

go/vt/vtgate/planbuilder/bypass.go
```diff
@@ -23,20 +23,43 @@ import (
 	"vitess.io/vitess/go/vt/vterrors"
 	"vitess.io/vitess/go/vt/vtgate/engine"
 	"vitess.io/vitess/go/vt/vtgate/planbuilder/plancontext"
+	"vitess.io/vitess/go/vt/vtgate/vindexes"
 )
 
+var enableShardRouting bool
```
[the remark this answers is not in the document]
rohit-nayak-ps: The set  is done as part of vtgate.Init() and get is done once vtgate comes up. I had originally added a mutex but realized later that they can't be called concurrently..

go/vt/vtgate/vindexes/vschema.go
```diff
@@ -59,10 +59,11 @@ const (
...
 type VSchema struct {
-	RoutingRules   map[string]*RoutingRule `json:"routing_rules"`
-	uniqueTables   map[string]*Table
-	uniqueVindexes map[string]Vindex
-	Keyspaces      map[string]*KeyspaceSchema `json:"keyspaces"`
+	RoutingRules      map[string]*RoutingRule `json:"routing_rules"`
+	uniqueTables      map[string]*Table
+	uniqueVindexes    map[string]Vindex
+	Keyspaces         map[string]*KeyspaceSchema `json:"keyspaces"`
+	ShardRoutingRules map[string]string          `json:"shard_routing_rules"`
```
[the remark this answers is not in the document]
rohit-nayak-ps: I couldn't understand where the looping that is referred to: maybe the code has changed since this comment.

go/vt/vtgate/vindexes/vschema.go
```diff
@@ -526,6 +528,16 @@ outer:
 	}
 }
 
+func buildShardRoutingRule(source *vschemapb.SrvVSchema, vschema *VSchema) {
+	if source.ShardRoutingRules == nil || len(source.ShardRoutingRules.Rules) == 0 {
+		return
```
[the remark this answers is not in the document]
rohit-nayak-ps: The len() is on a separate (parent) object that the nil check.

go/vt/wrangler/materializer.go
rohit-nayak-ps: fixed

go/vt/wrangler/materializer.go
rohit-nayak-ps: fixed

go/vt/wrangler/resharder.go
```diff
@@ -328,11 +328,16 @@ func (rs *resharder) createStreams(ctx context.Context) error {
...
+				int64(binlogdatapb.VReplicationWorkflowType_Reshard),
+				int64(binlogdatapb.VReplicationWorkflowSubType_None))
 		}
 
 		for _, rstream := range rs.refStreams {
-			ig.AddRow(rstream.workflow, rstream.bls, "", rstream.cell, rstream.tabletTypes)
+			ig.AddRow(rstream.workflow, rstream.bls, "", rstream.cell, rstream.tabletTypes,
+				//todo: fix based on original stream
+				int64(binlogdatapb.VReplicationWorkflowType_Reshard),
+				int64(binlogdatapb.VReplicationWorkflowSubType_None))
```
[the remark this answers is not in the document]
rohit-nayak-ps: Yes, we need to query the vrep table. A lot of tests started failing when I did that initially, We don't really use that field other than for reporting in Workflow Show, so I will do this in a separate PR once this one is merged.

go/vt/wrangler/traffic_switcher.go
```diff
@@ -881,15 +887,110 @@ func (wr *Wrangler) buildTrafficSwitcher(ctx context.Context, targetKeyspace, wo
...
+	for _, si := range ts.TargetShards() {
+		targetShards = append(targetShards, si.ShardName())
+	}
+	return sourceShards, targetShards
+}
+
+// has to be a move tables, same number of shards, not covering entire shard range, one-to-one shards in source and target
+func (ts *trafficSwitcher) isPartialMoveTables(sourceShards, targetShards []string) (bool, error) {
+
+	if ts.MigrationType() != binlogdatapb.MigrationType_TABLES {
```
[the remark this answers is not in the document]
rohit-nayak-ps: They are two different Enums. I only changed the case in the WorkflowType in this PR.

.github/workflows/cluster_endtoend_vreplication_partialmovetables.yml
```diff
@@ -0,0 +1,125 @@
+# DO NOT MODIFY: THIS FILE IS GENERATED USING "make generate_ci_workflows"
+
+name: Cluster (vreplication_partialmovetables)
+on: [push, pull_request]
+concurrency:
```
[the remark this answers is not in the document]
rohit-nayak-ps: Changed it to use mysql80

.github/workflows/cluster_endtoend_vtgate_transaction_partial_keyspace.yml
```diff
@@ -0,0 +1,106 @@
+# DO NOT MODIFY: THIS FILE IS GENERATED USING "make generate_ci_workflows"
+
+name: Cluster (vtgate_transaction) (partial keyspace)
+on: [push, pull_request]
+concurrency:
```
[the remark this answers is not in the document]
rohit-nayak-ps: Done. I also refactored test.go to allow specifying multiple config.*json. The config.json file was becoming pretty long and this will allow specifying tests in a modular way.

go/cmd/vtctldclient/command/shard_routing_rules.go
```diff
@@ -0,0 +1,156 @@
...
+		return err
+	}
+	// Round-trip so when we display the result it's readable.
+	data, err := cli.MarshalJSON(srr)
+	if err != nil {
+		return err
+	}
+
+	if applyShardRoutingRulesOptions.DryRun {
+		fmt.Printf("[DRY RUN] Would have saved new ShardRoutingRules object:\n%s\n", data)
```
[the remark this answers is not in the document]
rohit-nayak-ps: yes

test/config_partial_keyspace.json
```diff
@@ -0,0 +1,157 @@
+{
```
mattlord, line 1: A lot of these have retries, do we expect them to be flaky?

mattlord (changes requested):
Nice work on this! There were a couple of things we should change before merging (example related), but otherwise it's a few minor questions and suggestions that you can make the final call on. I'll review your feedback tomorrow and quickly approve. Thanks!

test/config_partial_keyspace.json
rohit-nayak-ps: I used the same settings as the original vtgate tests.

mattlord (approved):
❤️

Issue #9951, opened 2022-03-22
vttablet: query regression for json extract operator ->


derekperkins:
v12.0 correctly handled queries using `->` and `->>`, however v13 throws a syntax error. Note that it is a vttablet error, not a vtgate syntax error

https://dev.mysql.com/doc/refman/8.0/en/json-search-functions.html#operator_json-column-path

```
select data ->> '$.kindId'
     from iam__hubspot_workspace_user_syncer__msgs
     where attributes ->> '$.latestError' = 'user not found'
[2022-03-22 16:07:40] [HY000][1105] target: iam.0.primary: vttablet: rpc error: code = InvalidArgument desc = syntax error at position 24 near ':vtg1' (CallerID: nozzleadmin)
```

GuptaManan100:
Thankyou for reporting this issue @derekperkins. The fix for it is ready at https://github.com/vitessio/vitess/pull/10015. I will also backport this to release-13 when it is merged and we will do a minor release as well, since this issue causes regression in queries which worked previously

GuptaManan100:
Backport PR is ready at https://github.com/vitessio/vitess/pull/10022

PR #9990, opened 2022-03-27
Add parsing support for JSON Search Functions

K-Kumar-01:
## Description

Add the parsing support for JSON search functions. More details [here](https://dev.mysql.com/doc/refman/8.0/en/json-search-functions.html)

Need discussion on `MEMBER OF` .

## Related Issue(s)

#8604 
#4099 

## Checklist


go/vt/sqlparser/sql.y
```diff
@@ -6190,6 +6256,12 @@ non_reserved_keyword:
 | JSON_ARRAY
 | JSON_OBJECT
 | JSON_QUOTE
+| JSON_CONTAINS
+| JSON_CONTAINS_PATH
+| JSON_EXTRACT
+| JSON_KEYS
+| JSON_SEARCH
+| JSON_VALUE
```
GuptaManan100, line 6264: You missed JSON_OVERLAPS

go/vt/sqlparser/sql.y
```diff
@@ -6190,6 +6256,12 @@ non_reserved_keyword:
 | JSON_ARRAY
 | JSON_OBJECT
 | JSON_QUOTE
+| JSON_CONTAINS
```
GuptaManan100, line 6259: This should be alphabetically sorted.

go/vt/sqlparser/keywords.go
```diff
@@ -318,6 +318,13 @@ var keywords = []keyword{
...
+	{"json_contains", JSON_CONTAINS},
+	{"json_contains_path", JSON_CONTAINS_PATH},
+	{"json_extract", JSON_EXTRACT},
+	{"json_keys", JSON_KEYS},
+	{"json_overlaps", JSON_OVERLAPS},
+	{"json_search", JSON_SEARCH},
+	{"json_value", JSON_VALUE},
 	{"json_pretty", JSON_PRETTY},
 	{"json_storage_size", JSON_STORAGE_SIZE},
 	{"json_storage_free", JSON_STORAGE_FREE},
```
GuptaManan100, line 330: This should also be alphabetically sorted

GuptaManan100 (commented):
We should also add tests for the shorthands `->` and `->>`

go/vt/sqlparser/keywords.go
K-Kumar-01: > We should also add tests for the shorthands `->` and `->>`

There were some tests already present so I didn't add them.

derekperkins: This might not be the place for the tests, but there was a regression in v13 #9951

GuptaManan100: @derekperkins I plan to look at the issue before merging this PR today. I already tried running the query through the parser mentioned in #9951 and it passed. I think the issue might be in the planner, but I'll know more today and update on the issue

GuptaManan100 (approved):

PR #9992, opened 2022-03-28
Compatibility Fixes for 1.18

vmg:
## Description

Here's a couple things that were broken after the 1.18 upgrade:

- the `tlstest` package was signing our certs with SHA1, which has been deprecated in 1.18. Change it to use sha256 to fix the tests.
- `golangci-lint` was properly installed in CI, but not locally, so some people were seeing local errors.

cc @systay @ajm188

## Checklist

### DIFF
--- a/.github/workflows/golangci-linter.yml
+++ b/.github/workflows/golangci-linter.yml
@@ -25,7 +25,7 @@ jobs:
       uses: actions/checkout@v2
 
     - name: Install golangci-lint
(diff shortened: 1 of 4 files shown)


dbussink (approved):

systay (approved):

derekperkins:
golangci-lint still has some significant issues with 1.18 - enough where we rolled back internally, but we also use a lot of linters, so that may not be as applicable here. https://github.com/golangci/golangci-lint/issues/2649

PR #9997, opened 2022-03-29
VTTablet Init: fix race which can result in  schemacopy table not getting created

rohit-nayak-ps:
## Description
During the first tick of the health streamer the `schemacopy` table is attempted to be created. However we first try to create the `_vt` database. If it fails the following ddls are not run and hence the table doesn't get created. There are other code paths (in online ddl, vexec and vreplication) that can also create `_vt`. This race can result in the `schemacopy` table not getting created at all.
```
func (hs *healthStreamer) InitSchemaLocked(conn *connpool.DBConn) (bool, error) {
	for _, query := range mysql.VTDatabaseInit {
		_, err := conn.Exec(hs.ctx, query, 1, false)
		if err != nil {
			return false, err
		}
	}

	return true, nil
}
```
This PR ignores the errors from previous ddls to ensure that this race is handled.

### Note
The current method of ensuring that the `_vt` schema is up-to-date is not ideal: it is being done in multiple places in the code and we have several bugs surface where the correct code path was not followed resulting in the desired schema not being available.

There is a plan to consolidate the schema initialization and there should be a PR to achieve this soon. 

## Checklist

### DIFF
--- a/go/vt/vttablet/tabletserver/health_streamer.go
+++ b/go/vt/vttablet/tabletserver/health_streamer.go
@@ -396,7 +396,9 @@ func (hs *healthStreamer) InitSchemaLocked(conn *connpool.DBConn) (bool, error) 
 	for _, query := range mysql.VTDatabaseInit {
 		_, err := conn.Exec(hs.ctx, query, 1, false)
 		if err != nil {
-			return false, err
+			// it is possible that the _vt schema was already created by online ddl or vreplication as part of its init
+			// so ignore any errors and continue running future ddls
+			log.Infof("InitSchemaLocked: unable to run query %s: %s, ddl was probably already executed", query, err)
 		}
 	}
 


GuptaManan100 (commented):
I think this is okay 👍

harshit-gangal:
The first query that it executes is `create database if not exists _vt` does it means the user does not have permissions to create otherwise why will this query fail?

deepthi:
I have the same question as @harshit-gangal. Given that we are using `if not exists`, any errors from the first ddl are likely not because `_vt` already exists, but because the backing mysql is not yet ready. In which case ignoring the first error and executing the second DDL _may_ work sometimes (if the mysql becomes ready after the first DDL and before the second one), but it not guaranteed to solve the issue you are trying to solve.

frouioui:
Agreed with @deepthi and @harshit-gangal, the first query failure might not mean that the second query will be successful. I am still unsure how to fix this.

rohit-nayak-ps:
Had a brief discussion with @harshit-gangal last week and we couldn't think of an obvious code path that causes this failure. We do use the dba user, as in the other ddl creations, so unlikely it is a permission issue.

If this repeats we may want to add logs in `InitSchemaLocked()`  to log both successful table creation and failures.

I am closing this issue for now.

Issue #9996, opened 2022-03-29
ApplySchema -strategy "direct -allow-zero-in-date" silently drops DDL errors


shlomi-noach:
When running `vtctlclient ApplySchema -skip_preflight -strategy "direct -allow-zero-in-date"`, and when -`sql` has an error, then the command does not report the error.

Specifically, we were looking at `create view problem_view as select * from no_such_table` where `no_such_table` does not exist. The view creation fails, but `vtctlclient ApplySchema` does not return an error.

The reason is with this code:

https://github.com/vitessio/vitess/blob/68cbda31c046866aadad094b3b47077d76fe01e6/go/vt/schemamanager/tablet_executor.go#L432-L435

Because we prepend a valid SQL (`SET sql_mode...`) before the DDL, we get the error code for _that_ instead of for the DDL, and the error code is empty.

I'm not sure what's the best way to go about this. I'm thinking adding a `bool allow_zero_in_date = 6;` to `message ExecuteFetchAsDbaRequest` in `proto`, and then add a `allowZeroInDate bool` argument in all `ExecuteFetchAsDba` function signatures. But it seems excessive measures for just this particular scenario.

any other ideas are welcome.

rohit-nayak-ps:
What about adding "vitess" comments to the sql and parsing it in `ExecuteFetchAsDba`? This will make it more extensible, in case there are other flags that need similar handling in the future.

shlomi-noach:
@rohit-nayak-ps that makes sense, good idea! While not all query types support comments, all DDLs do, so that should work.

shlomi-noach:
OK this change was simple, PR to follow.

shlomi-noach:
Addressed by #9998

PR #9273, opened 2021-11-21
Support -allow-zero-in-date in ApplySchema 'direct' strategy

shlomi-noach:
## Description

`-allow-zero-in-date` is now respected in `vtctl ApplySchema` with `direct` strategy. For example:

```
vtctlclient ApplySchema -skip_preflight  -ddl_strategy='direct -allow-zero-in-date' -sql "create table if not exists t2(id int primary key, dt datetime default '0000-00-00 00:00:00')" commerce
```

Without `-allow-zero-in-date` this command would fail on `Invalid default value for 'dt' (errno 1067) (sqlstate 42000)`.

Note that `-skip_preflight` is required for this command to pass.

Tests added in `endtoend/onlineddl/declarative`

## Related Issue(s)

Followup to #9089 and #9142

## Checklist
- [ ] Should this PR be backported?
- [x] Tests were added or are not required
- [ ] Documentation was added or is not required

PR #9998, opened 2022-03-29
ApplySchema: allow-zero-in-date embedded as query comment in call to ExecuteFetchAsDba

shlomi-noach:
## Description

Fixes https://github.com/vitessio/vitess/issues/9996

Previously, `vtctlclient ApplySchema -skip_preflight -strategy "direct -allow-zero-in-date"` would prepend a given SQL with a `set @@session.sql_mode=` before sending to `ExecuteFetchAsDba`. This caused errors in the SQL to be silently dropped.

Now, instead, we embed a `/*vt+ allowZeroInDate=true */` hint in the SQL (only applicable to DDLs, which is fine, this is all we need).

Then, on `tabletmaneger`'s side, we parse the query, and if we're able to find such `/*vt+ allowZeroInDate=true */`, we run `set @@session.sql_mode=` before actually executing the SQL.

## Related Issue(s)

- https://github.com/vitessio/vitess/issues/9996
- caused by #9273 

Added validation test

## Checklist

### DIFF
--- a/go/vt/vttablet/tabletmanager/rpc_query.go
+++ b/go/vt/vttablet/tabletmanager/rpc_query.go
@@ -22,6 +22,7 @@ import (
 	"vitess.io/vitess/go/sqlescape"
 	"vitess.io/vitess/go/sqltypes"
 	"vitess.io/vitess/go/vt/log"
+	"vitess.io/vitess/go/vt/sqlparser"
 
 	querypb "vitess.io/vitess/go/vt/proto/query"
 )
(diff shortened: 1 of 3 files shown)


go/vt/vttablet/tabletmanager/rpc_query.go
```diff
@@ -49,6 +50,23 @@ func (tm *TabletManager) ExecuteFetchAsDba(ctx context.Context, query []byte, db
...
+	if stmt, err := sqlparser.Parse(string(query)); err == nil {
+		if ddlStmt, ok := stmt.(sqlparser.DDLStatement); ok {
+			if comments := ddlStmt.GetComments(); comments != nil {
+				directives := sqlparser.ExtractCommentDirectives(comments)
+				if _, ok := directives["allowZeroInDate"]; ok {
+					allowZeroInDate = true
+				}
+			}
+		}
+	}
```
harshit-gangal, line 64: nit: kindly reduce the cyclomatic complexity

go/vt/vttablet/tabletmanager/rpc_query.go
```diff
@@ -49,6 +50,23 @@ func (tm *TabletManager) ExecuteFetchAsDba(ctx context.Context, query []byte, db
...
+	if allowZeroInDate {
+		if _, err := conn.ExecuteFetch("set @@session.sql_mode=REPLACE(REPLACE(@@session.sql_mode, 'NO_ZERO_DATE', ''), 'NO_ZERO_IN_DATE', '')", 1, false); err != nil {
+			return nil, err
+		}
+	}
```
harshit-gangal, line 69: I think this can be done inline where the boolean is set.

harshit-gangal (approved):
overall looks good. some nits.

go/vt/vttablet/tabletmanager/rpc_query.go
shlomi-noach: @harshit-gangal I've turned it into something else; I don't think the rewrite is any better; the original made more sense to me. Please let me know what you think.

go/vt/vttablet/tabletmanager/rpc_query.go
shlomi-noach: done

harshit-gangal (approved):

go/vt/vttablet/tabletmanager/rpc_query.go
harshit-gangal: I felt it is easier to follow.

PR #9999, opened 2022-03-29
Replace OpenSSL usage for test certificates with Go crypto

dbussink:
This replaces the shelling out to OpenSSL with native Go crypto to generate CAs, intermediates & leaf certificates.

There's some improvements over the original behavior for correctness and performance reasons.

- First, this now uses ECDSA certificates with are a lot faster and generating a private key for them is much faster as well.
- There were no real leaf certificates. Each certificate was also marked as a CA which isn't correct usage of certificates. The changes here ensure that a signed cert is now a proper leaf certificate. It also adds an explicit separate method to generate an intermediate CA.
- Lifetime of the certificates generated is significantly reduces. Before it was almost 10 years, now it is one day. These temporary test certificates should not be used for anything long term, so a day is long enough for testing and using them in CI setups.
- It was setting 127.0.0.1 as a DNS SAN which doesn't work. This now adds 127.0.0.1 (and also ::1) as a proper IP SAN.
- The original OpenSSL certificates didn't setup bits like x509.KeyUsageCRLSign even though basic usage was marked as critical. This was technically in violation of the spec but a bunch of tools accept it (although not Go's CRL generation logic).

The API otherwise is unchanged for the tlstest package so any users shouldn't see any significant change, just a package that's a lot faster.

Unscientific benchmark for the test runtime shows a huge improvement:

#### Before

```
go test ./go/vt/tlstest -count=1
ok  	vitess.io/vitess/go/vt/tlstest	8.024s
```

#### After

```
go test ./go/vt/tlstest -count=1
ok  	vitess.io/vitess/go/vt/tlstest	0.156s
```

## Checklist


go/vt/tlstest/tlstest.go
```diff
@@ -19,207 +19,360 @@ limitations under the License.
...
- basicConstraints       = CA:TRUE
- subjectAltName         = @alternate_names
- extendedKeyUsage       =serverAuth,clientAuth
-[ alternate_names ]
- DNS.1                  = localhost
- DNS.2                  = 127.0.0.1
- DNS.3                  = %s
-`
+	CA          = "ca"
+	permissions = 0700
```
mdlayher, line 45: Nit: Go also supports `0o700` now for octal literals now which I usually encourage for clarity. But no big deal either way.

go/vt/tlstest/tlstest.go
```diff
@@ -19,207 +19,362 @@ limitations under the License.
...
+	var extKeyUsage []x509.ExtKeyUsage
+	var dnsNames []string
+	var ipAddresses []net.IP
+
+	if ca {
+		keyUsage = keyUsage | x509.KeyUsageCRLSign | x509.KeyUsageCertSign
+	} else {
+		extKeyUsage = []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth, x509.ExtKeyUsageClientAuth}
+		dnsNames = []string{"localhost", commonName}
+		ipAddresses = []net.IP{net.ParseIP("127.0.0.1"), net.ParseIP("::1")}
```
mdlayher, line 133: Can also just use net.IPv6loopback.

go/vt/tlstest/tlstest.go
```diff
@@ -19,207 +19,362 @@ limitations under the License.
...
-		"-new",
-		"-x509",
-		"-days", "3600",
-		"-nodes",
-		"-batch",
-		"-sha256",
-		"-config", config,
-		"-key", key,
-		"-out", cert)
+	keyPath := path.Join(root, "ca-key.pem")
```
mdlayher, line 168: For files, you specifically want `filepath.Join` as that will also work on Windows. See the package doc https://pkg.go.dev/path/filepath.

mdlayher (approved):
Nice work!

go/vt/tlstest/tlstest.go
dbussink: @mdlayher I haven't updated any of this and followed the pattern that's already there for now. I doubt Vitess on Windows is even supported at all.

vmg (approved):
You're a blessing @dbussink :pray:

PR #10005, opened 2022-03-29
Fix: reserved connection retry logic when vttablet or mysql is down

harshit-gangal:
## Description

This PR fixes issues with reserved connection when the vttablet or the underlying MySQL process is down.
The VTGate needs to understand the error codes/messages to redirect it to the new available tablet.

```
Issue 1: 
MySQL error: code 2002/2003 when the underlying MySQL is not available for connection.

Issue 2: 
Vttablet is down: ERROR 1105 (HY000): tablet: cell:"zone1" uid:100 is either down or nonexistent

Issue 3:
Vttablet tx engine is closed: tx engine can't accept new connections in state NotServing
```

In all the above scenarios the current reserved connection held should be dropped and a new one should be created by calling the ReserveExecute API to the tablet.

## Checklist
- [X] Should this PR be backported?

### DIFF
--- a/go/vt/vtgate/scatter_conn.go
+++ b/go/vt/vtgate/scatter_conn.go
@@ -701,9 +701,14 @@ func wasConnectionClosed(err error) bool {
 	sqlErr := mysql.NewSQLErrorFromError(err).(*mysql.SQLError)
 	message := sqlErr.Error()
 
-	return sqlErr.Number() == mysql.CRServerGone ||
-		sqlErr.Number() == mysql.CRServerLost ||
-		(sqlErr.Number() == mysql.ERQueryInterrupted && txClosed.MatchString(message))
+	switch sqlErr.Number() {
+	case mysql.CRServerGone, mysql.CRServerLost, mysql.CRConnectionError, mysql.CRConnHostError:
+		return true
+	case mysql.ERQueryInterrupted:
+		return txClosed.MatchString(message)
+	default:
+		return false
+	}
 }
 
 func requireNewQS(err error, target *querypb.Target) bool {
--- a/go/vt/vtgate/scatter_conn.go
+++ b/go/vt/vtgate/scatter_conn.go
@@ -183,7 +183,14 @@ func (stc *ScatterConn) ExecuteMultiShard(
 
 			qs, err = getQueryService(rs, info)
 			if err != nil {
-				return nil, err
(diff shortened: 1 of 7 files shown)


go/vt/vtgate/scatter_conn.go
```diff
@@ -183,7 +183,14 @@ func (stc *ScatterConn) ExecuteMultiShard(
...
 			if err != nil {
-				return nil, err
+				if info.reservedID == 0 || info.transactionID != 0 {
+					return nil, err
+				}
+				err = session.ResetShard(info.alias)
+				if err != nil {
+					return nil, err
+				}
+				qs = rs.Gateway
```
deepthi, line 193: Is the idea here that we route it through the gateway again and it will pick a tablet for us?
That's clever 💯

go/vt/vtgate/scatter_conn.go
```diff
@@ -701,16 +715,33 @@ func wasConnectionClosed(err error) bool {
 	sqlErr := mysql.NewSQLErrorFromError(err).(*mysql.SQLError)
 	message := sqlErr.Error()
 
-	return sqlErr.Number() == mysql.CRServerGone ||
-		sqlErr.Number() == mysql.CRServerLost ||
-		(sqlErr.Number() == mysql.ERQueryInterrupted && txClosed.MatchString(message))
+	switch sqlErr.Number() {
```
systay, line 718: much nicer to read like this

go/vt/vtgate/scatter_conn.go
```diff
@@ -701,16 +715,33 @@ func wasConnectionClosed(err error) bool {
...
+		return false
+	}
 }
 
 func requireNewQS(err error, target *querypb.Target) bool {
 	code := vterrors.Code(err)
 	msg := err.Error()
-	return (code == vtrpcpb.Code_FAILED_PRECONDITION && vterrors.RxWrongTablet.MatchString(msg)) ||
-		(code == vtrpcpb.Code_CLUSTER_EVENT && ((target != nil && target.TabletType == topodatapb.TabletType_PRIMARY) || vterrors.RxOp.MatchString(msg)))
+	if (code == vtrpcpb.Code_FAILED_PRECONDITION && vterrors.RxWrongTablet.MatchString(msg)) ||
```
systay, line 731: nit: maybe this could benefit from the same treatment as ☝️

systay (approved):

PR #10007, opened 2022-03-30
Fix 9900

### LINKED ISSUE #9900: vtgate: query regression sends EXISTS() queries to the wrong keyspace/shard
We use `EXISTS()` in a few places, which has worked I believe on all prior Vitess versions before v13. After upgrading, `from dual` was appended to these queries, resulting in them being sent to the wrong keyspace, and thus failing for missing tables. There was no difference whether I used the default or the Gen4 planner (I didn't see anywhere in the logs to prove that my Gen4 planner flag was applied, but I did provide it).

### Original Query
```sql
select exists (select * from workspace_users where 1 and user_id = ? and workspace_id = ?)

# also fails when explicitly targeting the keyspace, but the error is that the schema can't be found
select exists (select * from iam.workspace_users where 1 and user_id = ? and workspace_id = ?)
```

### Rewritten Query
```sql
select exists (select * from workspace_users where :vtg1 and user_id = :v1 and workspace_id = :v2) from dual
```

### Full Error
```
target: companies.0.primary: vttablet: Table 'companies.workspace_users' doesn't exist (errno 1146) (sqlstate 42S02) (CallerID: unsecure_grpc_client): Sql: "select exists (select * from workspace_users where :vtg1 and user_id = :v1 and workspace_id = :v2) from dual", BindVars: {v1: "type:INT64 value:\"133566477358474\""v2: "type:INT64 value:\"807210318492062\""vtg1: "type:INT64 value:\"1\""}
```

### Caveat

This was tested on a recent commit (https://github.com/vitessio/vitess/tree/vt_streamed_cols), not v13 proper, so it's possible the error was introduced post v13, in which case a backport is unnecessary, but a fix and a test case should be introduced.

systay:
## Description
Fixes two different bugs -
1. Queries with a dual subquery should not be replaced with literals when inside an `EXISTS` call
2. When merging routes where one side is a reference route, we have to copy keyspace information as well

## Related Issue(s)
Fixes #9900 

## Checklist
- [x] Should this PR be backported? *Yes*


go/vt/vtgate/planbuilder/plan_test.go
```diff
@@ -372,6 +396,21 @@ func TestWithDefaultKeyspaceFromFile(t *testing.T) {
 	testFile(t, "call_cases.txt", testOutputTempDir, vschema)
 }
 
+func TestWithDefaultKeyspaceFromFileSharded(t *testing.T) {
+	// We are testing this separately so we can set a default keyspace
+	vschema := &vschemaWrapper{
+		v: loadSchema(t, "schema_test.json", true),
+		keyspace: &vindexes.Keyspace{
+			Name:    "second_user",
```
GuptaManan100, line 404: I like the name of the keyspace 🤣

GuptaManan100 (approved):
LGTM! <3

Issue #9946, opened 2022-03-22
Flaky tabletmanager tests


deepthi:
### Overview of the Issue

![tm](https://user-images.githubusercontent.com/388311/159539465-176b8a0c-e570-49e2-88f6-3bab44abf727.png)

### Reproduction Steps

Seen in CI
https://app.launchableinc.com/organizations/vitess/workspaces/vitess-app/flaky-tests

### Binary Version

```sh
main
```

### Operating System and Environment details

```sh
GH Actions
```

GuptaManan100:
I don't think we are tracking anything on this issue anymore. Should we close this?

PR #10008, opened 2022-03-30
Fix flakiness in TestHealthCheck

GuptaManan100:
## Description

`TestHealthCheck` tests that Vitess reports the health of replication correctly. Underneath, it is using the Polling method (instead of the replication tracker), which relies on MySQL's `SHOW REPLICA STATUS` output. After we start the replication, sometimes it takes a while for replication lag to go away and be reflected in MySQL too. This led to the test being flaky.

The proposed fix in this PR is straightforward. Instead of directly checking that the status is healthy, we should have a timeout and check periodically until it is fixed.

## Related Issue(s)

- #9946 

## Checklist


go/test/endtoend/tabletmanager/tablet_health_test.go
```diff
@@ -186,6 +207,19 @@ func TestHealthCheck(t *testing.T) {
 
 	time.Sleep(tabletHealthcheckRefreshInterval)
 
+	// After replication has started, we sometimes see that MySQL reports a high
+	// replication lag but this clears really quick.
+	// In order to prevent the test from failing, we should keep checking with a timeout -
+	timeout = time.Now().Add(10 * time.Second)
```
systay, line 213: nit: this a pattern we might want to move to our test utils so we can reuse it

systay (approved):

go/test/endtoend/tabletmanager/tablet_health_test.go
GuptaManan100: Where do you want me to move this?

systay: Hmmm... That is in the `vtgate` package. Maybe we don't move it now, unless you can think of a good place for it to live in

deepthi (commented):
Based on the PR description, is it feasible to change the test to use the `heartbeat` method of computing replication lag? Then we may not need the wait.

GuptaManan100:
I don't know whether this test explicitly _wants_ to test the polling method, or it can be changed. But the entire test suite is using polling (setup in TestMain)

deepthi:
> I don't know whether this test explicitly _wants_ to test the polling method, or it can be changed. But the entire test suite is using polling (setup in TestMain)

I think that is just historical. Most likely either the test predates the `heartbeat` method or it was copied from one that did.

GuptaManan100:
@deepthi @systay I have made the changes. Now we use the heartbeats instead of polling. I had to make one more change though to set mysql as read-write. Other than that, everything works perfectly!

deepthi:
Thanks @GuptaManan100 for the numerous flaky tests fixes you have done in recent days!

PR #10010, opened 2022-03-30
[vtadmin-web] Fix default redirect on Tablet view

doeg:
## Description

This fixes a bug where the ReadOnlyGate component was breaking redirects when used within a `<Switch>`:

Screen Shot 2022-03-30 at 11 25 08 AM

Turns out my comment was wholly incorrect! Writing these tests took way more time than it was worth! And, best of all, this will likely need to be [rewritten](https://reactrouter.com/docs/en/v6/upgrading/v5#remove-redirects-inside-switch) when we upgrade to the next version of react-router anyway! 🤡 Oh well.

## Related Issue(s)

N/A

## Checklist

## Deployment Notes

N/A

### DIFF
--- a/web/vtadmin/src/components/ReadOnlyGate.tsx
+++ b/web/vtadmin/src/components/ReadOnlyGate.tsx
@@ -23,12 +23,26 @@ import { isReadOnlyMode } from '../util/env';
  *      <ReadOnlyGate>
  *          <SomeSpicyWriteAction />
  *      </ReadOnlyGate>
+ *
+ * Do not use ReadOnly gate in a react-router Switch statement as it will break
+ * redirects. The following will NOT work:
+ *
+ *      <Switch>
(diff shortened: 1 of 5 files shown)


ajm188 (approved):

web/vtadmin/src/components/ReadOnlyGate.tsx
```diff
@@ -23,12 +23,26 @@ import { isReadOnlyMode } from '../util/env';
  *      <ReadOnlyGate>
  *          <SomeSpicyWriteAction />
  *      </ReadOnlyGate>
+ *
+ * Do not use ReadOnly gate in a react-router Switch statement as it will break
```
notfelineit, line 27: I love the example!

notfelineit (approved):

PR #10013, opened 2022-03-30
[vtctldserver] Migrate GetPermissions PR#9903

rsajwani:
## Description
This PR migrates GetPermissions rpc to VtctldServer

Run it locally: vtctldclient --server ":15999" GetPermissions zone1-100
unit test: go test vitess.io/vitess/go/vt/vtctl/grpcvtctldserver -run TestGetPermissions

Issue: https://github.com/vitessio/vitess/issues/9903

## Related Issue(s)
Fixes # 9903

## Checklist
- [ ] Should this PR be backported? No


go/cmd/vtctldclient/command/tablets.go
```diff
@@ -24,6 +24,8 @@ import (
 
 	"github.com/spf13/cobra"
 
+	"vitess.io/vitess/go/json2"
```
ajm188, line 27: there is a `cli.MarshalJSON` you should use

go/cmd/vtctldclient/command/tablets.go
```diff
@@ -190,6 +192,15 @@ Note that, in the SleepTablet implementation, the value should be positively-sig
 		Args:                  cobra.ExactArgs(1),
 		RunE:                  commandStopReplication,
 	}
+	// GetPermissions asks the remote tablet for its permissions list.
+	GetPermissions = &cobra.Command{
+		Use:                   "GetPermissions <tablet_alias>",
+		Aliases:               []string{"GetPermissions"},
```
ajm188, line 198: This is not necessary, since you already have `GetPermissions` in the Use ([c.f.](https://pkg.go.dev/github.com/spf13/cobra#Command)

go/cmd/vtctldclient/command/tablets.go
```diff
@@ -190,6 +192,15 @@ Note that, in the SleepTablet implementation, the value should be positively-sig
...
+		Short:                 "Asks the remote tablet for its permissions list.",
```
ajm188, line 199: A better help message to the user would be something more like: "Returns the permissions list from the remote tablet." (or whatever is in `go/vt/vtctl/` for this equivalent)

go/cmd/vtctldclient/command/tablets.go
```diff
@@ -190,6 +192,15 @@ Note that, in the SleepTablet implementation, the value should be positively-sig
 		Args:                  cobra.ExactArgs(1),
 		RunE:                  commandStopReplication,
 	}
+	// GetPermissions asks the remote tablet for its permissions list.
```
ajm188, line 195: follow the other command comments please :)

go/cmd/vtctldclient/command/tablets.go
```diff
@@ -190,6 +192,15 @@ Note that, in the SleepTablet implementation, the value should be positively-sig
...
+	GetPermissions = &cobra.Command{
```
ajm188, line 196: please keep these sorted, like the others are :)

go/cmd/vtctldclient/command/tablets.go
```diff
@@ -557,6 +568,26 @@ func commandStopReplication(cmd *cobra.Command, args []string) error {
...
+		return err
+	}
+
+	cli.FinishedParsing(cmd)
+
+	resp, err := client.GetPermissions(commandCtx, &vtctldatapb.GetPermissionsRequest{
+		TabletAlias: alias,
+	})
+	if err != nil {
+		fmt.Printf("%v\n", err)
```
ajm188, line 583: you can just `return err`, and cobra will do the rest

go/cmd/vtctldclient/command/tablets.go
```diff
@@ -557,6 +568,26 @@ func commandStopReplication(cmd *cobra.Command, args []string) error {
...
+		return err
+	}
+	p, err := json2.MarshalIndentPB(resp.Permissions, "	")
+	fmt.Printf("%s\n", p)
```
ajm188, line 587: `cli.MarshalJSON` and also, if you get an error marshaling, you should probably not try to print

go/vt/vtctl/grpcvtctldserver/server.go
```diff
@@ -2371,6 +2371,26 @@ func (s *VtctldServer) RunHealthCheck(ctx context.Context, req *vtctldatapb.RunH
...
 }
 
+func (s *VtctldServer) GetPermissions(ctx context.Context, req *vtctldatapb.GetPermissionsRequest) (*vtctldatapb.GetPermissionsResponse, error) {
+	span, ctx := trace.NewSpan(ctx, "VtctldServer.GetPermissions")
+	defer span.Finish()
+
+	span.Annotate("tablet_alias", topoproto.TabletAliasString(req.TabletAlias))
+	ti, err := s.ts.GetTablet(ctx, req.TabletAlias)
+	if err != nil {
+		return nil, err
```
ajm188, line 2381: i would recommend wrapping this particular case in `vterrors.Errorf(vtrpc.Code_NOT_FOUND, ...)` (there are other examples here you can refer to!)

go/vt/vtctl/grpcvtctldserver/server.go
```diff
@@ -2371,6 +2371,26 @@ func (s *VtctldServer) RunHealthCheck(ctx context.Context, req *vtctldatapb.RunH
 	return &vtctldatapb.RunHealthCheckResponse{}, nil
 }
 
+func (s *VtctldServer) GetPermissions(ctx context.Context, req *vtctldatapb.GetPermissionsRequest) (*vtctldatapb.GetPermissionsResponse, error) {
```
ajm188, line 2374: sorting please :) also this needs a doc comment

go/vt/vtctl/grpcvtctldserver/server_test.go
```diff
@@ -11638,3 +11640,155 @@ func TestValidateShard(t *testing.T) {
...
+			if tt.shouldErr {
+				assert.Error(t, err)
+				return
+			}
+			// we should expect same user and DB permissions as assigned
+			assert.Equal(t, resp.Permissions.DbPermissions[0].Host, "host2")
+			assert.Equal(t, resp.Permissions.UserPermissions[0].Host, "host1")
+
+			// result should be Marshalled-able
+			perms, err := json2.MarshalIndentPB(resp.Permissions, "	")
```
ajm188, line 11787: I don't think we need to test this, as it's an implementation detail of `protobuf`, and not any code that we've written. Another way to think about this is, if we can't marshal protobuf messages to json, then _everything_ is broken, not the code this test is (supposed to be) testing.

go/vt/vtctl/grpcvtctldserver/testutil/test_tmclient.go
```diff
@@ -183,6 +183,13 @@ type TabletManagerClient struct {
 		Error  error
 	}
 	// keyed by tablet alias.
+	GetPermissionsDelays map[string]time.Duration
```
ajm188, line 186: sorting :)

go/vt/vtctl/grpcvtctldserver/testutil/test_tmclient.go
```diff
@@ -549,6 +556,36 @@ func (fake *TabletManagerClient) GetSchema(ctx context.Context, tablet *topodata
 	return nil, fmt.Errorf("%w: no schemas for %s", assert.AnError, key)
 }
 
+// GetPermission is part of the tmclient.TabletManagerClient interface.
+func (fake *TabletManagerClient) GetPermissions(ctx context.Context, tablet *topodatapb.Tablet) (*tabletmanagerdatapb.Permissions, error) {
```
ajm188, line 560: sorting :)

go/vt/vtctl/vtctl.go
```diff
@@ -3538,11 +3538,19 @@ func commandGetPermissions(ctx context.Context, wr *wrangler.Wrangler, subFlags
...
+		return err
 	}
-	return err
+	p, err := json2.MarshalIndentPB(resp.Permissions, "	")
+	if err != nil {
+		wr.Logger().Printf("%v\n", err)
+		return err
+	}
+	wr.Logger().Printf("%s\n", p)
+	return nil
```
ajm188, line 3553: the first part is correct; the second part should just be:

```go
if err != nil {
     return err
}

printJSON(wr.Logger(), p.Permissions)
return nil
```

proto/vtctldata.proto
```diff
@@ -1172,3 +1172,11 @@ message ValidateVSchemaResponse {
...
   map<string, ValidateShardResponse> results_by_shard = 2;
 }
+
+message GetPermissionsRequest {
+  topodata.TabletAlias tablet_alias = 1;
+}
+
+message GetPermissionsResponse {
+  tabletmanagerdata.Permissions permissions = 1;
+}
```
ajm188, line 1182: missing trailing newline, also sorting :)

proto/vtctlservice.proto
```diff
@@ -284,4 +284,6 @@ service Vtctld {
   rpc ValidateVersionKeyspace(vtctldata.ValidateVersionKeyspaceRequest) returns (vtctldata.ValidateVersionKeyspaceResponse) {};
   // ValidateVSchema compares the schema of each primary tablet in "keyspace/shards..." to the vschema and errs if there are differences.
   rpc ValidateVSchema(vtctldata.ValidateVSchemaRequest) returns (vtctldata.ValidateVSchemaResponse) {};
+  // GetPermissions returns permission set on the remote tablet.
+  rpc GetPermissions(vtctldata.GetPermissionsRequest) returns (vtctldata.GetPermissionsResponse) {};
```
ajm188, line 288: keep this ordered please

ajm188 (changes requested):
nice stuff! code is mostly there, just a couple comments around structure, tests, and "which marshal function do i use" (_sorry_ there are so many 😂) and then a bunch of nitpicks

go/cmd/vtctldclient/command/tablets.go
rsajwani: are you talking about line 195 here ?

go/cmd/vtctldclient/command/tablets.go
rsajwani: I believe you meant to remove fmt.print command in case of error...

go/vt/vtctl/grpcvtctldserver/server.go
rsajwani: shouldn't we making it consistent across other apis as well .. what do you recommend if I can change other in same PR or create another PR for changing across other API as well.. e.g healthcheck etc.

go/cmd/vtctldclient/command/tablets.go
ajm188: You don't need Aliases, since your only alias is the Use directly above

go/vt/vtctl/grpcvtctldserver/server.go
ajm188: Different failures have different semantics. Let's just focus on this one in this PR

rsajwani: Let just do it in another PR ..

ajm188: Well, I think we should introduce the pattern we want _here_, now, rather than merge something we know we immediately want to change. I'm just saying we should do it everywhere _else_ right now.

ajm188 (approved):

PR #10014, opened 2022-03-31
sqlparser: cache comment parsing

vmg:
## Description

This is from an internal issue by @aquarapid. Our `sqlparser` code is not very efficient at parsing directives inside of SQL comments. Very often during planning, we parse comment sets over and over again. We can do better!

This branch refactors the `sqlparser` so that the AST structure that represents a group of comments only needs to parse directives in the comments once, because it now caches the parsing internally. While I was in the area, I tweaked the behavior of how we store parsed directives. The old `map[string]any` wasn't great: it was eagerly parsing the types of the directives when it needn't need to. This design is simpler and uses way less memory. :ok_hand: 

cc @systay 

## Checklist

### DIFF
--- a/go/vt/vtgate/planbuilder/route.go
+++ b/go/vt/vtgate/planbuilder/route.go
@@ -17,6 +17,8 @@ limitations under the License.
 package planbuilder
 
 import (
+	"strconv"
+
 	"vitess.io/vitess/go/mysql/collations"
 	vtrpcpb "vitess.io/vitess/go/vt/proto/vtrpc"
 	"vitess.io/vitess/go/vt/vterrors"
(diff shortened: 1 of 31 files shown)


go/vt/vtgate/planbuilder/route.go
```diff
@@ -842,18 +844,10 @@ func (rb *route) exprIsValue(expr sqlparser.Expr) bool {
...
-
-	val, ok := d[sqlparser.DirectiveQueryTimeout]
-	if !ok {
-		return 0
-	}
-
-	intVal, ok := val.(int)
-	if ok {
-		return intVal
+	if val, ok := d[sqlparser.DirectiveQueryTimeout]; ok {
```
systay, line 847: before we were checking if the incoming map was nil, we don't need to do that anymore?

systay (approved):
🌶️ Nice stuff! 🚀 
LGTM

shlomi-noach (commented):
Looks good on my side of things (OnlineDDL). @vmg can you please also look into https://github.com/vitessio/vitess/blob/2b7e2ef42e23fb9809d5032249e48a0cd4a75c7f/go/vt/vttablet/tabletmanager/rpc_query.go#L53-L81 ?

go/vt/vtgate/planbuilder/route.go
vmg: A read from a `nil` map in Go is the same as a read from an empty map! It returns the default value + `false`. So we're handling that just fine!

shlomi-noach (approved):
:two_hearts:

PR #10017, opened 2022-03-31
Rename drop_constraints to drop_fk_constraints on MoveTables

Phanatic:
Signed-off-by: Phani Raj <phani@planetscale.com>

## Description
As a follow-up to https://github.com/vitessio/vitess/pull/9904, we're renaming `drop_constraints` to `drop_foreign_keys` on all tables when a user chooses to do so, on running MoveTables.

## Related Issue(s)
N/A

## Checklist


go/vt/wrangler/materializer.go
```diff
@@ -1064,6 +1074,36 @@ func (mz *materializer) deploySchema(ctx context.Context) error {
...
+	ast, err := sqlparser.ParseStrictDDL(ddl)
+	if err != nil {
+		return "", err
+	}
+
+	stripFKConstraints := func(cursor *sqlparser.Cursor) bool {
+		switch node := cursor.Node().(type) {
+		case sqlparser.DDLStatement:
+			if node.GetTableSpec() != nil {
+				noFKConstraints := make([]*sqlparser.ConstraintDefinition, 0, len(node.GetTableSpec().Constraints))
```
mdlayher, line 1088: Since we are now filtering the slice, allocating extra capacity up front with the length of the input slice could be wasteful. I'd probably do `var noFKConstraints []*sqlparser.ConstraintDefinition` and let `append()` do its thing.

go/vt/vtctl/vtctl.go
```diff
@@ -2263,7 +2263,7 @@ func commandMoveTables(ctx context.Context, wr *wrangler.Wrangler, subFlags *fla
 
 	autoStart := subFlags.Bool("auto_start", true, "If false, streams will start in the Stopped state and will need to be explicitly started")
 	stopAfterCopy := subFlags.Bool("stop_after_copy", false, "Streams will be stopped once the copy phase is completed")
-	dropConstraints := subFlags.Bool("drop_constraints", false, "If true, tables in the target keyspace will be created without constraints.")
+	dropFKConstraints := subFlags.Bool("drop_fk_constraints", false, "If true, tables in the target keyspace will be created without foreign key constraints.")
```
mdlayher, line 2266: Is it okay to completely replace this flag, and rename it? Are there valid uses of the old flag and path? If not, should that path be cleaned up too?

mdlayher (approved):
LGTM overall.

go/vt/vtctl/vtctl.go
Phanatic: yeah, its not part of any vitess release right now and should be ok to replace, @deepthi or @rohit-nayak-ps to confirm.

mattlord (approved):
I had one question but otherwise LGTM. Thanks!

go/vt/vtctl/vtctl.go
deepthi: Since this hasn't crossed a release boundary, it is fine to just rename it.
Though I would suggest calling it `drop_foreign_keys` instead.. @rohit-nayak-ps WDYT?

rohit-nayak-ps: Agree with @deepthi, about calling it `drop_foreign_keys`

go/vt/wrangler/materializer.go
```diff
@@ -63,8 +63,9 @@ type materializer struct {
 }
 
 const (
-	createDDLAsCopy               = "copy"
-	createDDLAsCopyDropConstraint = "copy:drop_constraint"
+	createDDLAsCopy                 = "copy"
+	createDDLAsCopyDropConstraint   = "copy:drop_constraint"
```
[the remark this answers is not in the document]
rohit-nayak-ps: Can we say authoritatively that we will never need to drop all constraints? If so, we will need to run a deprecation cycle to remove `copy:drop_constraint`, since it is possible there are applications using this flag at the moment.

rohit-nayak-ps (approved):
lgtm

go/vt/wrangler/materializer.go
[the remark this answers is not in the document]
deepthi: @Phanatic Why wasn't it replaced endtoend instead of only at the user interface? I would have expected the property here to be renamed instead of a new one being added.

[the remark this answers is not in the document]
deepthi: Yeah, that makes sense.

PR #10018, opened 2022-03-31
Increase max vrepl workflow source definition size from 64KiB to 16MiB

mattlord:
## Description
If you have a large number of tables and or shards involved in a vreplication workflow then you can bump into the existing 64KiB limit for the source definition. Moving the source column to `MEDIUMBLOB` increases the maximum size to 16MiB which should easily cover any reasonable definition. 

## Checklist
- [x] Tests are not required


rohit-nayak-ps (approved):
lgtm

PR #10023, opened 2022-04-04
Backport: Skip end-to-end test if binary is below a certain major version 

frouioui:
## Description
This pull request is a backport of #9975, allowing us to skip certain end-to-end tests based on the major version of a binary. One of the subquery end-to-end test is skipped when ran with vtgate below v14.

## Related Issue(s)
- Backport of #9975

## Checklist

### DIFF
--- a/go/test/endtoend/vtgate/queries/subquery/subquery_test.go
+++ b/go/test/endtoend/vtgate/queries/subquery/subquery_test.go
@@ -41,8 +41,10 @@ func TestSubqueriesHasValues(t *testing.T) {
 	utils.AssertMatches(t, conn, `SELECT id2 FROM t1 WHERE id1 NOT IN (SELECT id1 FROM t1 WHERE id1 > 10) ORDER BY id2`, `[[INT64(1)] [INT64(2)] [INT64(3)] [INT64(4)] [INT64(5)] [INT64(6)]]`)
 }
 
+// Test only supported in >= v14.0.0
 func TestSubqueriesExists(t *testing.T) {
 	defer cluster.PanicHandler(t)
+	utils.SkipIfBinaryIsBelowVersion(t, 14, "vtgate")
 	ctx := context.Background()
 	conn, err := mysql.Connect(ctx, &vtParams)
 	require.NoError(t, err)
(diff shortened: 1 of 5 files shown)


go/test/endtoend/vtgate/queries/subquery/subquery_test.go
```diff
@@ -41,8 +41,10 @@ func TestSubqueriesHasValues(t *testing.T) {
 	utils.AssertMatches(t, conn, `SELECT id2 FROM t1 WHERE id1 NOT IN (SELECT id1 FROM t1 WHERE id1 > 10) ORDER BY id2`, `[[INT64(1)] [INT64(2)] [INT64(3)] [INT64(4)] [INT64(5)] [INT64(6)]]`)
 }
 
+// Test only supported in >= v14.0.0
 func TestSubqueriesExists(t *testing.T) {
 	defer cluster.PanicHandler(t)
+	utils.SkipIfBinaryIsBelowVersion(t, 14, "vtgate")
```
GuptaManan100, line 47: Yeah, this won't work. With the backport the issue is fixed in release-13, but not in 12. So, adding this line will actually skip the test for this branch too, so we might as well just delete this test, since it won't run on any CI.
We should actually have it set to 13 here, so that it works in the normal CI tests and also in upgrade tests when the binaries are from release-13

GuptaManan100 (changes requested):

go/test/endtoend/vtgate/schematracker/sharded/st_sharded_test.go
```diff
@@ -24,6 +24,8 @@ import (
 	"testing"
 	"time"
 
+	"vitess.io/vitess/go/sqltypes"
+
```
GuptaManan100, line 28: This change also doesn't make sense, the imports were formatted fine previously

go/test/endtoend/vtgate/queries/subquery/subquery_test.go
frouioui: Just saw that we added support for this query got added in v13, my bad :) I'll set it to 13, makes more sense if we support it.

go/test/endtoend/vtgate/queries/subquery/subquery_test.go
```diff
@@ -41,8 +41,10 @@ func TestSubqueriesHasValues(t *testing.T) {
 	utils.AssertMatches(t, conn, `SELECT id2 FROM t1 WHERE id1 NOT IN (SELECT id1 FROM t1 WHERE id1 > 10) ORDER BY id2`, `[[INT64(1)] [INT64(2)] [INT64(3)] [INT64(4)] [INT64(5)] [INT64(6)]]`)
 }
 
+// Test only supported in >= v14.0.0
```
GuptaManan100, line 44: Comment needs to be updated too

GuptaManan100 (approved):
Everything else LGTM

PR #10031, opened 2022-04-04
Skip CI workflows based on the list of modified files

frouioui:
## Description

During the last few months, we've observed congestion in the GitHub Actions workflow queue. When there are several concurrent commits, the number of queued workflows increases a lot. We have recently introduced the `concurrency` feature of GitHub Actions. It enables us to cancel duplicated workflows in the same Pull Request, however, it does not help reduce the size of the queue when new commits are pushed to different Pull Request. The hypothesis is that we can reduce the number of queued workflows when new commits are pushed, without impeding our testing strategy.

The changes proposed in this Pull Request aim to reduce the size of that queue. Based on the list of files that have changed (added, modified, or deleted), we want to run certain workflows. For instance, modifying a README does not require running the entire set of workflows. The solution proposed was initially designed by @doeg.

GitHub Actions has two instructions: `on.paths` and `on.paths-ignore`, they allow us to skip or run workflow depending on a list of changed files. This solution works well, but not when the underlying job is marked as `required` in the repository's settings. Which is the case for most of our tests. Some issues regarding the same problem have been raised [here](https://github.community/t/unable-to-merge-pull-request-due-to-branch-protection-rules-since-github-actions-check-doesnt-trigger/126250) and [here](https://stackoverflow.com/questions/66751567/return-passing-status-on-github-workflow-when-using-paths-ignore). GitHub provides a [workaround](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks#handling-skipped-but-required-checks) which isn't the most elegant solution.

We started using [paths-filter](https://github.com/dorny/paths-filter). Instead of skipping the entire job, this alternative allows us to skip individual steps. It means that the required job will still run from start to finish, and only some of the steps will be skipped, resulting in a successful required job. For each workflow, we've added a new step that uses `paths-filter` and determines if relevant files have changed. The list of relevant files varies with the type of workflow we're running. A boolean is then outputted from this step. We then check this boolean in the `if` condition of all subsequent steps.

## Checklist

### DIFF
--- a/.github/workflows/check_formatting.yml
+++ b/.github/workflows/check_formatting.yml
@@ -1,30 +1,43 @@
 name: check_formatting
 on: [pull_request,push]
 jobs:
-
   build:
     name: Check Formatting
     runs-on: ubuntu-latest
     steps:
+    - name: Check out code
+      if: needs.check_relevant_files.outputs.go_files == 'true'
+      uses: actions/checkout@v2
+
+    - name: Check for changes in Go files
+      uses: dorny/paths-filter@v2
+      id: changes
+      with:
+        filters: |
+          go_files:
+            - '**/*.go'
+            - '*.go'
+            - 'go.[sumod]'
 
     - name: Set up Go
+      if: steps.changes.outputs.go_files == 'true'
       uses: actions/setup-go@v2
       with:
         go-version: 1.18
 
     - name: Tune the OS
+      if: steps.changes.outputs.go_files == 'true'
       run: |
         echo '1024 65535' | sudo tee -a /proc/sys/net/ipv4/ip_local_port_range
 
     # TEMPORARY WHILE GITHUB FIXES THIS https://github.com/actions/virtual-environments/issues/3185
     - name: Add the current IP address, long hostname and short hostname record to /etc/hosts file
+      if: steps.changes.outputs.go_files == 'true'
       run: |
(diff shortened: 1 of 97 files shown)


frouioui:
FYI reviewers:

- The commit https://github.com/vitessio/vitess/pull/10031/commits/01276b562bfa254f8a8c7b24f7df86be24dbaeb9 contains no modified Go files, resulting in most of the workflows being skipped. (see workflows)
- In contrast, the commit https://github.com/vitessio/vitess/pull/10031/commits/433d57266f08ccc5d0b03b0768899eee266fa3b7 was made at a time when a Go file was changed in this Pull Request, resulting in almost all the workflows to run. (see workflows)

frouioui:
@doeg - VTAdmin uses the `on.paths` feature in three workflows:
- vtadmin_web_build.yml
- vtadmin_web_lint.yml
- vtadmin_web_unit_tests.yml

I did not change them to use `paths-filter`, let me know if you think it needs to be changed.

GuptaManan100 (approved):
This looks good to me!

doeg (approved):
This is so great!!! ✨ Thank you so much for doing this! 

It's been a bit since I looked at those VTAdmin tests so I'll revisit them + can upgrade them to `paths-filter` if needed. I think there was some reason I wanted it... but, of course, I can't remember what it is now. 😆

PR #10033, opened 2022-04-04
VReplication: maintain original column case for pkCols

shlomi-noach:
## Description

Currently, vreplication maps `pkCol` (primary key column list) based on lower case column names. This is incorrect, and went under the radar because it only shows in a specific scenario: a PK column that is non-lower case is updated during vreplication workflow.

The fix is an extract from @rohit-nayak-ps 's #9742, where he just recently identified the same.

Added a `onlineddl/vrepl-suite` test to validate new behavior (the test fails without the fix in this PR).

## Checklist

### DIFF
--- a/go/test/endtoend/onlineddl/vrepl_suite/testdata/update-pk-col-uppercase/alter
+++ b/go/test/endtoend/onlineddl/vrepl_suite/testdata/update-pk-col-uppercase/alter
@@ -1,0 +1,1 @@
+engine=innodb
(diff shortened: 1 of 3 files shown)


go/test/endtoend/onlineddl/vrepl_suite/testdata/update-pk-col-uppercase/alter
```diff
@@ -0,0 +1 @@
+engine=innodb
```
rohit-nayak-ps, line 1: So is this alter which is triggering that bug. I didn't see the case of the pk being changed explicitly in this test, as I had expected.

rohit-nayak-ps (approved):
lgtm

PR #10034, opened 2022-04-04
feat: allow more complex expressions to be used in outer queries

systay:
## Description
Complex expressions on top of outer joins were not allowed to pass through the planner. The runtime does not seem to have any issues with these queries, so the planner check was probably no longer valid. This PR adds a few tests and removes the limitation in the planner.

## Checklist
- [x] Should this PR be backported? *no!*


frouioui (approved):
I don't see any objection to this change, but let's wait for @harshit-gangal's word

harshit-gangal (commented):
looks good. just need one additional test where the expression belongs to both the tables of the join.

harshit-gangal (approved):

PR #10039, opened 2022-04-05
collations/hack: cleanup dependencies

vmg:
## Description

Here's some cleanups for the `collations` package. I've done a small refactoring so the package is freestanding and can be used externally without depending on all of `vitess`.

While I was at it, I took a stab at the `go/hack` package and added some compatibility code so it doesn't necessarily depend on the default runtime of the Go Compiler. It should now also work on `gccgo` (if you're, like, a masochist) and `tinygo`.

## Checklist


go/hack/hack.go
```diff
@@ -48,11 +44,3 @@ func StringBytes(s string) []byte {
...
 	return b
 }
-
-// StringClone returns a newly allocated copy of the string that doesn't share
-// its underlying memory storage.
-func StringClone(s string) string {
-	b := make([]byte, len(s))
-	copy(b, s)
-	return *(*string)(unsafe.Pointer(&b))
-}
```
vmg, line 58: This function is now part of the stdlib as of Go 1.18 :tada:

go/hack/hack.go
```diff
@@ -32,12 +34,6 @@ func String(b []byte) (s string) {
 	return *(*string)(unsafe.Pointer(&b))
 }
 
-// StringPointer returns &s[0], which is not allowed in go
-func StringPointer(s string) unsafe.Pointer {
-	pstring := (*reflect.StringHeader)(unsafe.Pointer(&s))
-	return unsafe.Pointer(pstring.Data)
-}
-
```
vmg, line 40: This function was unused :tada:

go/hack/compat.go
```diff
@@ -0,0 +1,36 @@
+//go:build !gc || wasm
```
systay, line 1: shouldn't we have our header here?

go/hack/compat.go
```diff
@@ -0,0 +1,36 @@
...
+
+package hack
+
+func String(b []byte) (s string) {
```
systay, line 5: Similarly - what about comments for these public methods?

go/hack/compat.go
```diff
@@ -0,0 +1,36 @@
...
+
+func StringBytes(s string) []byte {
+	return []byte(s)
+}
+
+func DisableProtoBufRandomness() {}
+
+func RuntimeMemhash(b []byte, hash uint64) uint64 {
+	for i := 0; i < len(b); i++ {
+		hash *= 1099511628211
```
frouioui, line 34: Out of curiosity, what's `1099511628211`?

go/hack/compat.go
vmg: Fixed!

frouioui (approved):

go/hack/compat.go
systay: Maybe it has to do with [this](https://en.wikipedia.org/wiki/Fowler%E2%80%93Noll%E2%80%93Vo_hash_function)?

vmg: The default prime constant in FNV1: https://cs.opensource.google/go/go/+/refs/tags/go1.18:src/hash/fnv/fnv.go;l=36

PR #10040, opened 2022-04-05
Prefer using REPLICA tablets when selecting vreplication sources

mattlord:
## Description
This changes the defaults for [tablet selection](https://vitess.io/docs/14.0/reference/vreplication/tablet_selection/) when choosing vreplication sources so that we use a `REPLICA` tablet when one is available and only fall back to using a `PRIMARY` tablet when necessary.

This is a good default as the vreplication workflow can have a very significant impact on the source tablet and should generally not be done on a `PRIMARY` tablet when possible since it's likely serving production traffic.

## Related Issue(s)
- ???

## Checklist
- [x] Tests are not required (existing ones are fine)
- [x] Documentation was added in: https://github.com/vitessio/website/pull/1003

### DIFF
--- a/go/vt/vtctl/vtctl.go
+++ b/go/vt/vtctl/vtctl.go
@@ -455,7 +455,7 @@ var commands = []commandGroup{
 				name:   "Reshard",
 				method: commandReshard,
 				params: "[-source_shards=<source_shards>] [-target_shards=<target_shards>] [-cells=<cells>] [-tablet_types=<source_tablet_types>]  [-skip_schema_copy] <action> 'action must be one of the following: Create, Complete, Cancel, SwitchTraffic, ReverseTrafffic, Show, or Progress' <keyspace.workflow>",
-				help:   "Start a Resharding process. Example: Reshard -cells='zone1,alias1' -tablet_types='primary,replica,rdonly'  ks.workflow001 '0' '-80,80-'",
+				help:   "Start a Resharding process. Example: Reshard -cells='zone1,alias1' -tablet_types='PRIMARY,REPLICA,RDONLY'  ks.workflow001 '0' '-80,80-'",
 			},
 			{
 				name:   "MoveTables",
@@ -510,7 +510,7 @@ var commands = []commandGroup{
 			{
 				name:   "VDiff",
 				method: commandVDiff,
-				params: "[-source_cell=<cell>] [-target_cell=<cell>] [-tablet_types=primary,replica,rdonly] [-filtered_replication_wait_time=30s] [-max_extra_rows_to_compare=1000] <keyspace.workflow>",
+				params: "[-source_cell=<cell>] [-target_cell=<cell>] [-tablet_types=PRIMARY,REPLICA,RDONLY] [-filtered_replication_wait_time=30s] [-max_extra_rows_to_compare=1000] <keyspace.workflow>",
 				help:   "Perform a diff of all tables in the workflow",
 			},
 			{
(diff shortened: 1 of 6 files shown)


go/vt/vtctl/vtctl.go
```diff
@@ -2257,7 +2257,7 @@ func commandMoveTables(ctx context.Context, wr *wrangler.Wrangler, subFlags *fla
 
 	workflow := subFlags.String("workflow", "", "Workflow name. Can be any descriptive string. Will be used to later migrate traffic via SwitchReads/SwitchWrites.")
 	cells := subFlags.String("cells", "", "Cell(s) or CellAlias(es) (comma-separated) to replicate from.")
-	tabletTypes := subFlags.String("tablet_types", "", "Source tablet types to replicate from (e.g. PRIMARY, REPLICA, RDONLY). Defaults to -vreplication_tablet_type parameter value for the tablet, which has the default value of PRIMARY,REPLICA.")
```
rohit-nayak-ps, line 2260: We also have a similar helptext in the v2 umbrella command `commandVRWorkflow` ...

rohit-nayak-ps (approved):
one additional heptext needs to be modified, but otherwise lgtm

go/vt/vtctl/vtctl.go
mattlord: Good catch! I think I covered that here: https://github.com/vitessio/vitess/pull/10040/commits/52a76e89edc846660ab0a4f7ca5666f7a7248919

go/vt/vtctl/vtctl.go
```diff
@@ -2330,7 +2330,7 @@ func commandVRWorkflow(ctx context.Context, wr *wrangler.Wrangler, subFlags *fla
 	const defaultMaxReplicationLagAllowed = defaultWaitTime
 
 	cells := subFlags.String("cells", "", "Cell(s) or CellAlias(es) (comma-separated) to replicate from.")
-	tabletTypes := subFlags.String("tablet_types", "primary,replica,rdonly", "Source tablet types to replicate from (e.g. primary, replica, rdonly). Defaults to -vreplication_tablet_type parameter value for the tablet, which has the default value of replica.")
+	tabletTypes := subFlags.String("tablet_types", "in_order:REPLICA,PRIMARY", "Source tablet types to replicate from (e.g. primary, replica, rdonly). Defaults to --vreplication_tablet_type parameter value for the tablet, which has the default value of in_order:REPLICA,PRIMARY.")
```
deepthi, line 2333: nit: we seem to be using uppercase in the default value, but lowercase in the help text.

mattlord: Where do you see it lower case?

go/vt/vtctl/vtctl.go
mattlord: That required a follow-up: https://github.com/vitessio/vitess/pull/10040/commits/1ecf139817991b78c994b892ec862b0111529b50

go/vt/vtctl/vtctl.go
mattlord: ah, you mean more generally than the default value itself e.g. `(e.g. primary, replica, rdonly)`. I prefer capitalized but don't feel too strongly about it.

mattlord: At least started doing that here https://github.com/vitessio/vitess/pull/10040/commits/ebc046911df1391544292fed78a06b569eb603b2 and here https://github.com/vitessio/website/pull/1003/commits/679678ea7cd6cf2ca3fad988c9aac6e03251f7fd

go/vt/vtctl/vtctl.go
```diff
@@ -2486,7 +2486,7 @@ func commandVRWorkflow(ctx context.Context, wr *wrangler.Wrangler, subFlags *fla
 		vrwp.Cells = *cells
 		vrwp.TabletTypes = *tabletTypes
 		if vrwp.TabletTypes == "" {
-			vrwp.TabletTypes = "primary,replica,rdonly"
+			vrwp.TabletTypes = "in_order:REPLICA,PRIMARY"
```
rohit-nayak-ps, line 2489: Was `rdonly` removed  on purpose?

mattlord: Yeah, according to our docs were use the default of the tablet flag if you don't specify and that's what we're actually now doing here. After this PR, the only place RDONLY is potentially used by default is for VDiff.

Let me know if not using the default was intentional before and perhaps the docs were incorrect.

mattlord: Related to this, I'm not sure why are have this `if vrwp.TabletTypes == "" {` code block at all since we are specifying a default value here: https://github.com/vitessio/vitess/pull/10040/files#diff-ad37010accc5a29c6751d18124328bb882457e8b2527db688b854c81fe23861cR2333

Maybe that wasn't always working as I'd expect.

mattlord: @rohit-nayak-ps turns out that you were right! I shouldn't have removed that. Re-adding it fix this: https://github.com/vitessio/vitess/issues/10421

:facepalm:

mattlord: Well, after further testing... this didn't actually change any behavior. We were never switching `RDONLY` traffic by default -- as the `tablet_types` flag variable was never an empty string as it has a default value (which is what we changed) -- and we weren't testing for this either. Will correct both in https://github.com/vitessio/vitess/issues/10421.

PR #10041, opened 2022-04-05
[vtadmin|tests] Add mutex to prevent data races in tests using cluster/resolver

ajm188:
## Description

Because grpc is triggering calls to ResolveNow in a background goroutine,
which in turn make calls to a Discovery implementation (fakediscovery, in
these cases), we can have a data race when trying to wipe out a particular
fakediscovery's host set.

Before:

```
$ go test -race ./go/vt/vtadmin/... >/dev/null ; echo $?
1
```

After:

```
$ go test -race ./go/vt/vtadmin/... >/dev/null; echo $?
0
```

## Checklist

### DIFF
--- a/go/vt/vtadmin/cluster/discovery/fakediscovery/discovery.go
+++ b/go/vt/vtadmin/cluster/discovery/fakediscovery/discovery.go
@@ -20,6 +20,7 @@ package fakediscovery
 import (
 	"context"
 	"math/rand"
+	"sync"
 
 	"github.com/stretchr/testify/assert"
 
@@ -44,6 +45,8 @@ type gates struct {
 type Fake struct {
 	gates   *gates
 	vtctlds *vtctlds
+
+	m sync.Mutex
 }
 
 // New returns a new fake.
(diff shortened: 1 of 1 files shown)


go/vt/vtadmin/cluster/discovery/fakediscovery/discovery.go
```diff
@@ -103,19 +115,28 @@ func (d *Fake) AddTaggedVtctlds(tags []string, vtctlds ...*vtadminpb.Vtctld) {
...
+	defer d.m.Unlock()
+
 	d.vtctlds.shouldErr = shouldErr
 }
 
 var _ discovery.Discovery = (*Fake)(nil)
 
 // DiscoverVTGates is part of the discovery.Discovery interface.
 func (d *Fake) DiscoverVTGates(ctx context.Context, tags []string) ([]*vtadminpb.VTGate, error) {
+	d.m.Lock()
```
ajm188, line 137: note that all the other DiscoverVTGate* functions call this, so they get the lock for free (ditto on vtctld functions)

mattlord (approved):
❤️

PR #10043, opened 2022-04-05
[Dynamic Clusters] No Duplicates

notfelineit:

## Description

This PR fixes dynamic cluster behavior by only creating adding a cluster to `clusterMap`, `clusters`, and `clusterCache` if it does not already exist in `clusterMap`. 

In this case, `clusterMap` is the source of truth for what exists across the 3 stores.

## Related Issue(s)

N/A

## Checklist
- [ ] Tests were added or are not required **Tests are added**
- [ ] Documentation was added or is not required **No**

## Deployment Notes

N/A

### DIFF
--- a/go/vt/vtadmin/api.go
+++ b/go/vt/vtadmin/api.go
@@ -246,22 +246,24 @@ func (api *API) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 				err = json.Unmarshal(decoded, &clusterJSON)
 				if err == nil {
 					clusterID := clusterJSON.ClusterName
-					c, err := cluster.Config{
-						ID:            clusterID,
-						Name:          clusterID,
-						DiscoveryImpl: "dynamic",
-						DiscoveryFlagsByImpl: cluster.FlagsByImpl{
-							"dynamic": map[string]string{
-								"discovery": string(decoded),
+					if _, exists := api.clusterMap[clusterID]; !exists {
+						c, err := cluster.Config{
+							ID:            clusterID,
+							Name:          clusterID,
+							DiscoveryImpl: "dynamic",
+							DiscoveryFlagsByImpl: cluster.FlagsByImpl{
+								"dynamic": map[string]string{
+									"discovery": string(decoded),
+								},
 							},
-						},
-					}.Cluster()
-					if err == nil {
-						api.clusterMap[clusterID] = c
-						api.clusters = append(api.clusters, c)
-						err = api.clusterCache.Add(clusterID, c, 24*time.Hour)
-						if err != nil {
-							log.Infof("could not add dynamic cluster %s to cluster cache: %+v", clusterID, err)
+						}.Cluster()
+						if err == nil {
+							api.clusterMap[clusterID] = c
+							api.clusters = append(api.clusters, c)
+							err = api.clusterCache.Add(clusterID, c, 24*time.Hour)
+							if err != nil {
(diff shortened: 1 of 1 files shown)


go/vt/vtadmin/api.go
```diff
@@ -246,22 +246,24 @@ func (api *API) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
-						api.clusterMap[clusterID] = c
-						api.clusters = append(api.clusters, c)
-						err = api.clusterCache.Add(clusterID, c, 24*time.Hour)
-						if err != nil {
-							log.Infof("could not add dynamic cluster %s to cluster cache: %+v", clusterID, err)
+						}.Cluster()
+						if err == nil {
+							api.clusterMap[clusterID] = c
+							api.clusters = append(api.clusters, c)
+							err = api.clusterCache.Add(clusterID, c, 24*time.Hour)
```
ajm188, line 263: i was looking at the docs, and i think we should use `cache.DefaultExpiration` which will cover us in case we update the one constant without the other

ajm188: also i filed https://github.com/vitessio/vitess/projects/12#card-80225244

go/vt/vtadmin/api.go
```diff
@@ -338,6 +340,15 @@ func (api *API) EjectDynamicCluster(key string, value any) {
 	if ok {
 		delete(api.clusterMap, key)
 	}
+
+	// Maintain order of clusters when removing dynamic cluster
+	clusters := []*cluster.Cluster{}
```
ajm188, line 345: `NewAPI` sorts these by ID, so you can use `sort.Search` (from the standard lib, not from `vtadmin/sort`) to do this in log(n) time.

go/vt/vtadmin/api_test.go
```diff
@@ -4942,6 +4967,10 @@ func TestServeHTTP(t *testing.T) {
 
 			api.ServeHTTP(w, req)
 
+			if tt.repeat {
```
ajm188, line 4970: [two hops this time!!](https://www.youtube.com/watch?v=wZv62ShoStY)

notfelineit: easiest way to do it 😅  😅

go/vt/vtadmin/api.go
notfelineit: Oh sweet! Good to know

go/vt/vtadmin/api.go
notfelineit: That's going to be cool!

ajm188 (approved):

go/vt/vtadmin/api.go
```diff
@@ -246,22 +246,24 @@ func (api *API) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
-					if err == nil {
-						api.clusterMap[clusterID] = c
-						api.clusters = append(api.clusters, c)
-						err = api.clusterCache.Add(clusterID, c, 24*time.Hour)
-						if err != nil {
-							log.Infof("could not add dynamic cluster %s to cluster cache: %+v", clusterID, err)
+						}.Cluster()
+						if err == nil {
+							api.clusterMap[clusterID] = c
+							api.clusters = append(api.clusters, c)
```
ajm188, line 262: oh, i missed that we need to re-run this through `sort.ClustersBy` to preserve the ordering. i'll fix that in my branch, sorry about that!

PR #10044, opened 2022-04-05
[vtadmin] threadsafe dynamic clusters

ajm188:
## Description

Dynamic clusters currently make any api endpoint accessing `.clusters` or `.clusterMap` not thread safe, as the dynamic handler _could_ be mutating those structures as another endpoint attempts to access them. This fixes that by adding a mutex (abstracted away under the `getCluster(s)ForRequest`) and all other callsites updated to use that.

There's a second piece around duplicate clusters which @notfelineit is working on.

## Related Issue(s)

- #9544
- #10043 

## Checklist
- [x] Tests were added or are not required — tested manually, still able to add the dynamic cluster, and serve static clusters

### DIFF
--- a/go/vt/vtadmin/api.go
+++ b/go/vt/vtadmin/api.go
@@ -64,9 +64,10 @@ import (
 type API struct {
 	vtadminpb.UnimplementedVTAdminServer
 
+	clusterMu    sync.Mutex // guards `clusters` and `clusterMap`
 	clusters     []*cluster.Cluster
-	clusterCache *cache.Cache
 	clusterMap   map[string]*cluster.Cluster
+	clusterCache *cache.Cache
 	serv         *grpcserver.Server
 	router       *mux.Router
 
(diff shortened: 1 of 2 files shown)


go/vt/vtadmin/api.go
```diff
@@ -222,15 +223,7 @@ func (api *API) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
-	dynamicAPI := &API{
-		clusters:   api.clusters,
-		clusterMap: api.clusterMap,
-		router:     api.router,
-		serv:       api.serv,
-		authz:      api.authz,
-		options:    api.options,
-	}
-
+	var dynamicAPI *API
```
ajm188, line 226: if we're able to unpack the cookie, this will become a second API with its own `clusters` and `clusterMap` structures. if we get to the bottom and we're still `nil`, we use the original api, with the original structures and mutex.

go/vt/vtadmin/api.go
```diff
@@ -253,23 +249,45 @@ func (api *API) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 								},
 							},
 						}.Cluster()
+
 						if err == nil {
 							api.clusterMap[clusterID] = c
 							api.clusters = append(api.clusters, c)
+							sort.ClustersBy(func(c1, c2 *cluster.Cluster) bool {
+								return c1.ID < c2.ID
+							}).Sort(api.clusters)
```
ajm188, line 258: https://github.com/vitessio/vitess/pull/10043/files#r843924919

PR #10048, opened 2022-04-06
schemadiff: load, normalize, validate and compare schemas

shlomi-noach:
## Description

Up till now [schemadiff](https://github.com/vitessio/vitess/pull/9719) allowed us to compare a table with another table, or a view with another view. This is [now in use](https://github.com/vitessio/vitess/pull/9772) in declarative schema changes.

This PR introduces schema comparisons.

A `Schema` is a collection of database objects -- right now this means tables and views. You may load a schema by:

- a list of well formed entities, or
- a list of well formed `Statement`s, or
- a list of `CREATE ...` queries, or
- a SQL blob that contains multiple `CREATE` queries

A schema is validated upon loading:

- can't have two objects with same name
- all view dependencies are resolved

As for the latter: `schemadiff` parses each view and analyze its dependencies. It validates that all dependencies are met. e.g. you can't create a view that reads from a nonexistent table. Also, you can't create two views that have circular dependency.

Lastly, the schema is normalized: we reorder the objects by order of valid creation: first, the tables; then, views that only depend on tables; then, 2nd level views, etc.

### Diff

`schemadiff` can now do, well, a schema-diff. Compare two `Schema` objects. This returns a (possibly empty) list of diffs, which create/alter/drop tables and views so as to mutate one schema to another. This is all done declaratively.

See the added tests for more insights.

## Checklist

### DIFF
--- a/go/vt/schemadiff/diff.go
+++ b/go/vt/schemadiff/diff.go
@@ -59,9 +59,9 @@ func DiffTables(create1 *sqlparser.CreateTable, create2 *sqlparser.CreateTable, 
 	case create1 == nil && create2 == nil:
 		return nil, nil
 	case create1 == nil:
-		return NewCreateTableEntity(create2).Create()
+		return NewCreateTableEntity(create2).Create(), nil
 	case create2 == nil:
-		return NewCreateTableEntity(create1).Drop()
+		return NewCreateTableEntity(create1).Drop(), nil
 	default:
 		c1 := NewCreateTableEntity(create1)
 		c2 := NewCreateTableEntity(create2)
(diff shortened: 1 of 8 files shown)


go/vt/schemadiff/diff.go
```diff
@@ -116,15 +113,46 @@ func DiffViews(create1 *sqlparser.CreateView, create2 *sqlparser.CreateView, hin
...
+	}
+	return schema1.Diff(schema2, hints)
+}
+
+// DiffSchemasSQL compares two schemas and returns the list of diffs that turn
+// 1st schema into 2nd. Any of the schemas may be nil.
+func DiffSchemas(schema1 *Schema, schema2 *Schema, hints *DiffHints) ([]EntityDiff, error) {
+	var err error
+	if schema1 == nil {
+		schema1, err = NewSchemaFromStatements(nil)
```
vmg, line 146: Is there a reason why this isn't `newEmptySchema` directly?

vmg (approved):
This looks great. I only have a small nit. :sparkles:

go/vt/schemadiff/diff.go
shlomi-noach: Good catch! Now simplified.

shlomi-noach:

PR #10049, opened 2022-04-06
Re-add sanitize_log_messages flag and add tests for flags

mattlord:
## Description
The `--sanitize_log_messages` vttablet flag was errantly/accidentally removed in: https://github.com/vitessio/vitess/pull/9923

We re-add it here and ensure it's there and working in the `TestFlags` unit test. This was ultimately caused by me not adding (didn't know better at the time) a unit test for the new config option when adding it in https://github.com/vitessio/vitess/pull/9550. 

⚠️ NOTE: after realizing that the unit test wouldn't have caught this either... I added a new e2e test to check the command flags to be sure that we're intentional about changes and can prevent future accidental changes. This e2e test will be run as part of [the generic endtoend workflow](https://github.com/vitessio/vitess/blob/b979359f861d838ed3e0d7adc3cbba57a2975c7e/test/config.json#L26-L36). ⚠️ 

## Related Issue(s)
 - https://github.com/vitessio/vitess/pull/9550
 - https://github.com/vitessio/vitess/pull/9923

## Checklist
- [x] Tests were added

### DIFF
--- a/go/flags/endtoend/flags_test.go
+++ b/go/flags/endtoend/flags_test.go
@@ -12,17 +12,12 @@ distributed under the License is distributed on an "AS IS" BASIS,
 WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
-
-This test cell aliases feature
-
-We start with no aliases and assert that vtgates can't route to replicas/rondly tablets.
-Then we add an alias, and these tablets should be routable
 */
 
 package flags
 
 // These tests ensure that we are changing flags intentionally and do not accidentally make
-// changes such as removing a flag by accident. Since there's no way to test the command-line
+// changes such as removing a flag. Since there's no way to test the command-line
 // flag handling portion explicitly in the unit tests we do so here.
 
 import (
--- a/go/flags/endtoend/flags_test.go
+++ b/go/flags/endtoend/flags_test.go
@@ -1,0 +1,1469 @@
+/*
+Copyright 2022 The Vitess Authors.
+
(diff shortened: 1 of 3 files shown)


go/flags/endtoend/flags_test.go
```diff
@@ -0,0 +1,1469 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+This test cell aliases feature
+
+We start with no aliases and assert that vtgates can't route to replicas/rondly tablets.
+Then we add an alias, and these tablets should be routable
```
deepthi, line 19: Copy-paste? should be removed

go/flags/endtoend/flags_test.go
```diff
@@ -0,0 +1,1469 @@
...
+*/
+
+package flags
+
+// These tests ensure that we are changing flags intentionally and do not accidentally make
+// changes such as removing a flag by accident. Since there's no way to test the command-line
```
deepthi, line 25:
```suggestion
// These tests ensure that we are changing flags intentionally and do not accidentally make
// changes such as removing a flag. Since there's no way to test the command-line
```

go/flags/endtoend/flags_test.go
mattlord: Good catch! I didn't expect anything else in the copyright/license header :-)

deepthi (approved):

PR #10050, opened 2022-04-06
[vtadmin] grpc dynamic clusters

ajm188:
## Description

This refactor enables the specifying of a dynamic cluster via [grpc metadata](https://github.com/grpc/grpc-go/blob/master/Documentation/grpc-metadata.md) in addition to an http cookie.

**NOTE** this deprecates the `--http-enable-dynamic-clusters` flag in favor of the `--enable-dynamic-clusters`. After we've deployed this, I'll circle back around to clean up the flag.

### Testing

- `curl`-ing with a cookie still works, as before
- grpc demo: https://gist.github.com/ajm188/5b5c8ba0cc5660298697e0f762081d45

### Preparatory refactor

The first thing we do is extract as much of the non-"related to serving an http request" code out of `ServeHTTP(w, r)` as possible, moving it to `package dynamic` (in `vitess.io/vitess/go/vt/vtadmin/cluster/dynamic`). Now, the ServeHTTP handler "simply":
1. says "hey, here's a cookie, can you get me a cluster out of it?" (this is `dynamic.ClusterFromString`)
2. then, depending on the answer to that question:
    a. if we can extract a cluster, says "hey, give me a dynamic API with this cluster or with this cluster id" (this is the `WithCluster` method of the `dynamic.API` interface, which `vtadmin.API` implements), and we invoke that API's `Handler()` to serve the request
    b. if we can't extract a cluster, then the original API is used

### gRPC metadata

Again, the [docs on metadata](https://github.com/grpc/grpc-go/blob/master/Documentation/grpc-metadata.md) are actually quite thorough and good, so if you want to dig deeper I would recommend reading those, but the tl;dr is:

* when you're sending metadata to someone, you put it into an "outgoing" context
* when you're checking if you were sent metadata from someone else, you extract it from an "incoming" context
* metadata is just a map of `map[string][]string` (that is, specifying the same key multiple times will append to a list instead of overwriting existing values)

### The Interceptors

For this write-up, I'm only going to discuss the unary interceptor because:
1. vtadmin has no streaming rpcs (............... yet?)
2. the streaming interceptor is basically the same "shape" of code as the unary, just with slightly different types

So! The first thing we do, on package `init` is cache a mapping of `fullMethodName => methodHandler`, so we can lookup later (the Service Descriptor provides a list of method handlers, so you would have to scan the entire list if we didn't do this).

A `methodHandler` is a function with the type:

```go
type methodHandler func(srv interface{}, ctx context.Context, dec func(in interface{}) error) interceptor grpc.UnaryServerInterceptor) (resp interface{}, err error)
```

If we look at how one of these method handlers is used in the generated vtadmin_grpc.pb.go code, we'll get a better sense of what these arguments actually are in practice:

```go

func _VTAdmin_GetGates_Handler(srv interface{}, ctx context.Context, dec func(interface{}) error, interceptor grpc.UnaryServerInterceptor) (interface{}, error) {
	in := new(GetGatesRequest)
	if err := dec(in); err != nil {
		return nil, err
	}
	if interceptor == nil {
		return srv.(VTAdminServer).GetGates(ctx, in)
	}
	info := &grpc.UnaryServerInfo{
		Server:     srv,
		FullMethod: "/vtadmin.VTAdmin/GetGates",
	}
	handler := func(ctx context.Context, req interface{}) (interface{}, error) {
		return srv.(VTAdminServer).GetGates(ctx, req.(*GetGatesRequest))
	}
	return interceptor(ctx, in, info, handler)
}
```

Here we see that:
- `srv` is _some_ implementation of `VTAdminServer` (in practice, the thing you pass to `vtadminpb.RegisterVTAdminServer(api.serv.GRPCServer(), api)`)
- `ctx` is the context that gets passed as the first argument on the concrete `GetGates` method that we wrote in `api.go`.
- `dec` is a function that takes an input which comes from ....... somewhere (off the wire, decoded by grpc), and decoded into an instance of `GetGatesRequest` (this will become the `req` argument to our concrete `GetGates` method in `api.go`). if we can't decode the input ("if `dec` returns an error") we fail immediately.
- `interceptor` is the next function in the interceptor chain. if we're at the end of the chain (`interceptor == nil`), then we invoke the concrete `GetGates` method on `srv` (i.e. we go directly to our method in `api.go`). if we have more interceptors to call, we keep walking down the chain until either an interceptor fails, or we invoke the `handler`.

Soooooo, there's nothing stopping us from accessing this function and providing it a _different_ `srv` argument than what we registered via `vtadminpb.RegisterVTAdminServer(api.serv.GRPCServer(), api)`, which _also_ implements the `VTAdminServer` interface. Let's take another look at the type of `dynamic.API`:

```go
// vtadmin/cluster/dynamic/api.go

type API interface {
    vtadminpb.VTAdminServer
    WithCluster(*cluster.Cluster, string) API
    Handler() http.Handler
}
```

Oh, how convenient! The thing we get back from `WithCluster` is also of type `dynamic.API`, which implements `VTAdminServer`, which means we can safely use it as the `srv` argument to the `methodHandler`! Neat!

So, taking a look at our interceptor:

```go
func UnaryServerInterceptor(api API) grpc.UnaryServerInterceptor {
	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (resp interface{}, err error) {
		c, id, ok, err := clusterFromIncomingContextMetadata(ctx)
		switch {
		case !ok:
			// No dynamic cluster metadata, proceed directly to handler.
			return handler(ctx, req)
		case id == "":
			// There was a cluster spec in the metadata, but we couldn't even
			// get an id out of it. Warn and fallback to static API.
			log.Warningf("failed to unmarshal dynamic cluster spec from incoming context metadata; falling back to static API; error: %v", err)
			return handler(ctx, req)
		}

		if err != nil {
			log.Warningf("failed to extract valid cluster from incoming metadata; attempting to use existing cluster with id=%s; error: %v", id, err)
		}

		dynamicAPI := api.WithCluster(c, id)
		method := methodHandlersByName[nameFromFullMethod(info.FullMethod)]
		return method(dynamicAPI, ctx, dec(req), nil)
	}
}
```

This mirrors the structure of the `ServeHTTP` version, with some grpc/interceptor specifics sprinkled on top. Namely:
1. try to get a cluster out of the metadata (`ok == false` if there was no metadata, or no "cluster" key)
2. if we failed to get a cluster, invoke `handler`, which continues us down the interceptor to chain to eventually invoke the method on the original, static API (this is the existing behavior)
3. if we did get a cluster, call `WithCluster` on the static API to get a dynamic API back, and lookup `methodHandler` for this method that we cached at `init`-time; pass in the dynamic API as `srv`, and a dummy `dec` function that just proto-clones the request.
    a. we know it's safe to do this request cloning/casting, because grpc has actually already decoded the original request into a object of the correct request type
    b. we currently don't support further interceptors after this one (we pass `nil` as the interceptor argument). we can change that by adding a method to `dynamic.API` to access an interceptor chain if we ever need to do that. for now, we just have to be sure these interceptors are the final ones we pass in `grpcserver.NewServer` (which we do)

## Related Issue(s)

- #9544 
- #10044 

## Checklist

### DIFF
--- a/go/vt/vtadmin/cluster/dynamic/interceptors.go
+++ b/go/vt/vtadmin/cluster/dynamic/interceptors.go
@@ -1,0 +1,134 @@
+package dynamic
+
+import (
+	"context"
+	"strings"
+
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
+	"google.golang.org/protobuf/proto"
+
+	"vitess.io/vitess/go/vt/log"
+	"vitess.io/vitess/go/vt/vtadmin/cluster"
+
+	vtadminpb "vitess.io/vitess/go/vt/proto/vtadmin"
+)
+
+// StreamServerInterceptor returns a StreamServerInterceptor that redirects a
+// streaming RPC to a dynamic API if the incoming context has a cluster spec in
+// its metadata. Otherwise, the interceptor is a no-op, and the original stream
+// handler is invoked.
+func StreamServerInterceptor(api API) grpc.StreamServerInterceptor {
+	return func(srv interface{}, ss grpc.ServerStream, info *grpc.StreamServerInfo, handler grpc.StreamHandler) error {
+		c, id, ok, err := clusterFromIncomingContextMetadata(ss.Context())
+		switch { // TODO: see if it's possible to de-duplicate this somehow. Unfortunately the callbacks have different signatures which might make it impossible.
+		case !ok:
+			// No dynamic cluster metadata, proceed directly to handler.
+			return handler(srv, ss)
+		case id == "":
+			// There was a cluster spec in the metadata, but we couldn't even
+			// get an id out of it. Warn and fallback to static API.
+			log.Warningf("failed to unmarshal dynamic cluster spec from incoming context metadata; falling back to static API; error: %v", err)
+			return handler(srv, ss)
+		}
+
+		if err != nil {
+			log.Warningf("failed to extract valid cluster from incoming metadata; attempting to use existing cluster with id=%s; error: %v", id, err)
+		}
+
+		dynamicAPI := api.WithCluster(c, id)
+		streamHandler := streamHandlersByName[nameFromFullMethod(info.FullMethod)]
+		return streamHandler(dynamicAPI, ss)
+	}
+}
+
+// UnaryServerInterceptor returns a UnaryServerInterceptor that redirects a
+// unary RPC to a dynamic API if the incoming context has a cluster spec in its
+// metadata. Otherwise, the interceptor is a no-op, and the original method
+// handler is invoked.
+func UnaryServerInterceptor(api API) grpc.UnaryServerInterceptor {
+	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (resp interface{}, err error) {
+		c, id, ok, err := clusterFromIncomingContextMetadata(ctx)
+		switch {
+		case !ok:
+			// No dynamic cluster metadata, proceed directly to handler.
+			return handler(ctx, req)
+		case id == "":
+			// There was a cluster spec in the metadata, but we couldn't even
+			// get an id out of it. Warn and fallback to static API.
+			log.Warningf("failed to unmarshal dynamic cluster spec from incoming context metadata; falling back to static API; error: %v", err)
+			return handler(ctx, req)
+		}
+
+		if err != nil {
+			log.Warningf("failed to extract valid cluster from incoming metadata; attempting to use existing cluster with id=%s; error: %v", id, err)
+		}
+
+		dynamicAPI := api.WithCluster(c, id)
+		method := methodHandlersByName[nameFromFullMethod(info.FullMethod)]
+
+		// NOTE: because we don't have access to the interceptor
+		// chain (but we _could_ if we wanted to add a method to the
+		// DynamicAPI interface), this MUST be the last interceptor
+		// in the chain.
+		return method(dynamicAPI, ctx, dec(req), nil)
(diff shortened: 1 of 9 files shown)


go/vt/vtadmin/cluster/dynamic/interceptors.go
```diff
@@ -0,0 +1,134 @@
+package dynamic
```
ajm188, line 1: happy to discuss anything in this file in person if it's helpful; it's .... a lot 😅

ajm188: (actually let me take a stab at _actually_ writing docs first, so it exists for posterity; _then_ i will make my offer)

doeg (approved):
This is neat. I don't have any constructive golang feedback, but the commentary was A+. I've been wanting to learn about interceptors for a bit (since it's a common pattern in client-side stuff) and your documentation was super helpful in getting a handle (heh heh ..) on everything. Thanks for that!

Are you planning to use (or already using) VTAdmin's gRPC API at PlanetScale? Cool if so.

ajm188:
> Are you planning to use (or already using) VTAdmin's gRPC API at PlanetScale? Cool if so.

dunno, but we can (with this change)! this is more part of my philosophical stance that http/grpc should have complete parity (and, whenever node-grpc stops requiring server-side nonsense, grpc becomes (transparently to the react codebase) the de facto way) (to be clear, i'm fully-committed to the parity stance, the "grpc-only" is more a far-off dream that may never happen or even be possible)

notfelineit (approved):
This is so nice and clean!! Love it!

PR #10052, opened 2022-04-07
[13.0] Fix: reserved connection retry logic when vttablet or mysql is down

harshit-gangal:
## Description

This is backport PR of #10005 

This PR fixes issues with reserved connection when the vttablet or the underlying MySQL process is down.
The VTGate needs to understand the error codes/messages to redirect it to the new available tablet.

```
Issue 1: 
MySQL error: code 2002/2003 when the underlying MySQL is not available for connection.

Issue 2: 
Vttablet is down: ERROR 1105 (HY000): tablet: cell:"zone1" uid:100 is either down or nonexistent

Issue 3:
Vttablet tx engine is closed: tx engine can't accept new connections in state NotServing
```

In all the above scenarios the current reserved connection held should be dropped and a new one should be created by calling the ReserveExecute API to the tablet.

## Checklist

### DIFF
--- a/go/test/endtoend/vtgate/reservedconn/reconnect3/main_test.go
+++ b/go/test/endtoend/vtgate/reservedconn/reconnect3/main_test.go
@@ -1,0 +1,96 @@
+/*
+Copyright 2022 The Vitess Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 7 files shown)


GuptaManan100 (approved):

frouioui (approved):
Looks good to me

PR #9971, opened 2022-03-23
Add parsing support for Json schema validation functions

K-Kumar-01:
## Description

Adds the parsing support for JSON schema validation functions.
More information [here](https://dev.mysql.com/doc/refman/8.0/en/json-validation-functions.html).

### Changes
- Parser file grammar support.
- Valid and Invalid test cases.

## Related Issue(s)

#8604 
#4099 

## Checklist
- [ ] Should this PR be backported?
- [x] Tests were added or are not required
- [x] Documentation was added or is not required

PR #10053, opened 2022-04-07
Reduce shift-reduce conflicts in the grammar by using Precedence

GuptaManan100:
## Description

With ongoing work to support more and more functions in parsing, the number of shift-reduce conflicts is increasing by a lot.

This PR removes the shift-reduce conflicts occurring due to addition of special grammar rules for functions whose names are non-reserved keywords using precedence rules.

I am, in general, not a fan of precedence rules, and think that they should only be used if it is absolutely required in the grammar, since they make it much more complex and harder to read and debug. But, with each additional function's parsing being added, the shift-reduce conflicts are increasing by one. We have already reached 20 now. Having these many shift-reduce conflicts is not ideal either, since it becomes much harder to track what conflicts we had in the first place. This PR, resolves these conflicts using precedence rules by using the docs - https://www.ibm.com/docs/en/zos/2.4.0?topic=ambiguities-resolving-conflicts-by-precedence and https://www.ibm.com/docs/en/zos/2.4.0?topic=section-precedence-associativity

Eventually, when we have parsing for all the functions in MySQL and we can remove the function_call_generic grammar rules, we can also remove this precedence symbol.

## Related Issue(s)

- #8604 
- Should be merged after #9971 

## Checklist


GuptaManan100:
This PR is based on top of #9971 and should be merged after it.

frouioui (approved):
awesome!

systay (approved):
Nice stuff.

However:
> Eventually, when we have parsing for all the functions in MySQL and we can remove the function_call_generic grammar rules, we can also remove this precedence symbol.

Since you can in MySQL declare your own functions, won't we forever be stuck with a `function_call_generic` for functions that are not native?

harshit-gangal:
> Nice stuff.
> 
> However:
> 
> > Eventually, when we have parsing for all the functions in MySQL and we can remove the function_call_generic grammar rules, we can also remove this precedence symbol.
> 
> Since you can in MySQL declare your own functions, won't we forever be stuck with a `function_call_generic` for functions that are not native?

how can you do that? just curious

harshit-gangal (approved):

GuptaManan100:
This is the doc for it - https://dev.mysql.com/doc/refman/8.0/en/create-procedure.html. That's a really good point @systay. So I guess, this change is here to stay then!

PR #10055, opened 2022-04-07
[13.0] fix: dual query with exists clause having system table query in it

harshit-gangal:
## Description

This is backport of PR #9969 

This PR adds an empty map if the system table map is uninitialized in route.

## Related Issue(s)

- Fixes https://github.com/vitessio/vitess/issues/9653

## Checklist


frouioui (approved):
lgtm!

PR #7505, opened 2021-02-17
MoveTables: Refresh SrvVSchema (for Routing Rules) and source tablets (for Blacklisted Tables) on completion

rohit-nayak-ps:
## Description
This PR fixes two refresh problems during vreplication workflows:

* SrvVSchema was not being rebuilt after DropSources resulting in existing routing rules still being visible to vtgate since vttablets still saw the old rules. This PR updates the SrvVSchema when routing rules are modify.
* When blacklisted tables are updated only the current primary tablet was being refreshed. However this causes a problem if there is a reparenting operation. This PR now refreshes all tablets in the shard.

## Related Issue(s)
https://github.com/vitessio/vitess/issues/7477

## Checklist
- [ ] Should this PR be backported?
- [X] Tests were added or are not required
- [X] Documentation was added or is not required

## Impacted Areas in Vitess
Components that this PR will affect:

- [X]  Query Serving
- [X]  VReplication
- [ ]  Cluster Management
- [ ]  Build/CI
- [ ]  VTAdmin

PR #10058, opened 2022-04-07
Improve Tablet Refresh Behavior in VReplication Traffic Switch Handling

mattlord:
## Description
Because we refresh the tablets in source and target shards involved in a vreplication workflow during the operation (e.g. `SwitchWrites`) — in order to update the serving vschema and query execution rules — let's also do that for the dry run so that we can provide a warning for users that the operation could fail. Example output:
```
$ time vtctlclient --server=127.0.0.1:15999 MoveTables -- --dry_run --tablet_types=rdonly,replica SwitchTraffic customer.commerce2customer

Following vreplication streams are running for workflow customer.commerce2customer:

id=1 on 0/zone1-0000000200: Status: Running. VStream Lag: 0s.

MoveTables Error: rpc error: code = Unknown desc = cannot switch traffic for workflow commerce2customer at this time: could not refresh all of the tablets involved in the operation:
failed to successfully refresh all tablets in the customer/0 target shard (<nil>):
  failed to refresh tablet zone1-0000000202: rpc error: code = DeadlineExceeded desc = latest balancer error: connection error: desc = "transport: Error while dialing dial tcp 172.31.41.160:16202: connect: connection refused"


real	0m30.081s
user	0m0.013s
sys	0m0.005s
✘-1 ~/git/vitess/examples/local [improve_vrepl_dry_run|…1]
```

The [`topotools.RefreshTabletsByShard` call](https://github.com/vitessio/vitess/blob/5374002ec7f627d6e14647d78eb74788e089ac74/go/vt/topotools/keyspace.go#L31-L39) is best-effort and lets the caller know if the results were not complete (partial). We use a lower timeout within the traffic switcher when refreshing the state for each tablet (done in a goroutine) as:
  1. The default etcd lock/lease TTL is 60 seconds and if any of the tablets are unresponsive then the topo lock is likely to be lost by the time we go to release it
  2. The caller will know if the results were partial and can take appropriate action

We also add tablet refresh to the traffic switcher's pre-check call to ensure that we have healthy tablets on the source and target shards before we begin the operation as otherwise the operation is not safe and can fail. Example output:
```
$ time vtctlclient --server=127.0.0.1:15999 MoveTables -- --tablet_types=primary,rdonly,replica SwitchTraffic customer.commerce2customer 2>/dev/null

Following vreplication streams are running for workflow customer.commerce2customer:

id=1 on 0/zone1-0000000200: Status: Running. VStream Lag: 0s.

MoveTables Error: rpc error: code = Unknown desc = cannot switch traffic for workflow commerce2customer at this time: could not refresh all of the tablets involved in the operation:
failed to successfully refresh all tablets in the commerce/0 source shard:
  failed to refresh tablet zone1-0000000102: rpc error: code = DeadlineExceeded desc = latest balancer error: connection error: desc = "transport: Error while dialing dial tcp 172.31.41.160:16102: connect: connection refused"


real	0m30.064s
user	0m0.008s
sys	0m0.009s
✘-1 ~/git/vitess/examples/local [improve_vrepl_dry_run|…1]
```

**_All of these things help to make vreplication workflows more predictable, more reliable, and incur less downtime._**

## Related Issue(s)
  - https://github.com/vitessio/vitess/pull/7505

## Checklist
- [x] Tests were not required (I don't think... for the dry run output)
- [x] Documentation: https://github.com/vitessio/website/pull/1013


go/vt/topotools/keyspace.go
```diff
@@ -54,6 +54,9 @@ func RefreshTabletsByShard(ctx context.Context, ts *topo.Server, tmc tmclient.Ta
 	var (
 		m  sync.Mutex
 		wg sync.WaitGroup
+		// Set a relatively low timeout and the callers will know if we got a partial refresh and decide
+		// what to do if so.
+		refreshTimeout = 2 * time.Second
```
rohit-nayak-ps, line 59: Who are the other consumers of this RPC? Is it ok to reduce it to 2s for everybody or does it need to be parameterized?

mattlord:
I ended up changing the function signature for `topotools.RefreshTabletsByShard()` so we can easily see where it's being called in the PR changes. The only thing left is going through each of those to see if we should START checking for partial results (they were ignored before). @rohit-nayak-ps or @deepthi happy to walk through that if either of you have opinions. The time were we could check and cancel is between:
- The pre-check: https://github.com/vitessio/vitess/blob/18f9573a405c3e50058038759dcec493f64191bb/go/vt/wrangler/workflow.go#L276-L282
- The point of no return, where we create the journal records:
  - For SwitchReads: ... not clear we have one
  - For SwitchWrites: https://github.com/vitessio/vitess/blob/18f9573a405c3e50058038759dcec493f64191bb/go/vt/wrangler/traffic_switcher.go#L572-L577

rohit-nayak-ps (approved):
lgtm

go/vt/topotools/keyspace.go
```diff
@@ -121,7 +137,7 @@ func UpdateShardRecords(
 		// For 'to' shards, refresh to make them serve. The 'from' shards will
 		// be refreshed after traffic has migrated.
 		if !isFrom {
-			if _, err := RefreshTabletsByShard(ctx, ts, tmc, si, cells, logger); err != nil {
+			if _, _, err := RefreshTabletsByShard(ctx, ts, tmc, si, cells, logger); err != nil {
```
mattlord, line 140: This is called from a number of places via `Wrangler.updateShardRecords()`. I don't think that we can change the partial handling here w/o other changes.

go/vt/vtctl/grpcvtctldserver/server.go
```diff
@@ -2064,7 +2064,7 @@ func (s *VtctldServer) RefreshStateByShard(ctx context.Context, req *vtctldatapb
 		return nil, fmt.Errorf("Failed to get shard %s/%s/: %w", req.Keyspace, req.Shard, err)
 	}
 
-	isPartial, err := topotools.RefreshTabletsByShard(ctx, s.ts, s.tmc, si, req.Cells, logutil.NewCallbackLogger(func(e *logutilpb.Event) {
+	isPartial, _, err := topotools.RefreshTabletsByShard(ctx, s.ts, s.tmc, si, req.Cells, logutil.NewCallbackLogger(func(e *logutilpb.Event) {
```
mattlord, line 2067: I don't think it makes sense to change the partial handling here, other than we could add the reasons for partial to the response message. I'm going to try and do this. At least then callers can take action and know the causes for it.

mattlord: Done: https://github.com/vitessio/vitess/pull/10058/commits/96b96bbf44126d57e5efeab10c1955a83762f04e

go/vt/wrangler/keyspace.go
```diff
@@ -354,7 +354,7 @@ func (wr *Wrangler) cancelHorizontalResharding(ctx context.Context, keyspace, sh
 
 		destinationShards[i] = updatedShard
 
-		if _, err := topotools.RefreshTabletsByShard(ctx, wr.ts, wr.tmc, si, nil, wr.Logger()); err != nil {
+		if _, _, err := topotools.RefreshTabletsByShard(ctx, wr.ts, wr.tmc, si, nil, wr.Logger()); err != nil {
```
mattlord, line 357: Doesn't make sense to return an error when canceling.

go/vt/wrangler/keyspace.go
```diff
@@ -442,7 +442,7 @@ func (wr *Wrangler) MigrateServedTypes(ctx context.Context, keyspace, shard stri
 		refreshShards = destinationShards
 	}
 	for _, si := range refreshShards {
-		_, err := topotools.RefreshTabletsByShard(ctx, wr.ts, wr.tmc, si, cells, wr.Logger())
+		_, _, err := topotools.RefreshTabletsByShard(ctx, wr.ts, wr.tmc, si, cells, wr.Logger())
```
mattlord, line 445: I believe this is past the point of no return.

go/vt/wrangler/keyspace.go
```diff
@@ -792,7 +792,7 @@ func (wr *Wrangler) masterMigrateServedType(ctx context.Context, keyspace string
 	}
 
 	for _, si := range destinationShards {
-		if _, err := topotools.RefreshTabletsByShard(ctx, wr.ts, wr.tmc, si, nil, wr.Logger()); err != nil {
+		if _, _, err := topotools.RefreshTabletsByShard(ctx, wr.ts, wr.tmc, si, nil, wr.Logger()); err != nil {
```
mattlord, line 795: I believe this is past the point of no return.

go/vt/wrangler/keyspace.go
```diff
@@ -1226,7 +1226,7 @@ func (wr *Wrangler) replicaMigrateServedFrom(ctx context.Context, ki *topo.Keysp
 
 	// Now refresh the source servers so they reload the denylist
 	event.DispatchUpdate(ev, "refreshing sources tablets state so they update their denied tables")
-	_, err := topotools.RefreshTabletsByShard(ctx, wr.ts, wr.tmc, sourceShard, cells, wr.Logger())
+	_, _, err := topotools.RefreshTabletsByShard(ctx, wr.ts, wr.tmc, sourceShard, cells, wr.Logger())
```
mattlord, line 1229: I believe this is past the point of no return.

go/vt/wrangler/traffic_switcher.go
```diff
@@ -1389,7 +1403,9 @@ func (ts *trafficSwitcher) dropSourceDeniedTables(ctx context.Context) error {
 		}); err != nil {
 			return err
 		}
-		_, err := topotools.RefreshTabletsByShard(ctx, ts.TopoServer(), ts.TabletManagerClient(), source.GetShard(), nil, ts.Logger())
+		rtbsCtx, cancel := context.WithTimeout(ctx, shardTabletRefreshTimeout)
+		defer cancel()
+		_, _, err := topotools.RefreshTabletsByShard(rtbsCtx, ts.TopoServer(), ts.TabletManagerClient(), source.GetShard(), nil, ts.Logger())
```
mattlord, line 1408: This is called when canceling.

go/vt/wrangler/traffic_switcher.go
```diff
@@ -1015,7 +1025,9 @@ func (ts *trafficSwitcher) changeTableSourceWrites(ctx context.Context, access a
...
```
mattlord, line 1030: This is a safe/good place to return an error if we could not refresh all tablets as its early on in the operation.

go/vt/wrangler/traffic_switcher.go
```diff
@@ -1283,7 +1295,9 @@ func (ts *trafficSwitcher) allowTableTargetWrites(ctx context.Context) error {
 		}); err != nil {
 			return err
 		}
-		_, err := topotools.RefreshTabletsByShard(ctx, ts.TopoServer(), ts.TabletManagerClient(), target.GetShard(), nil, ts.Logger())
+		rtbsCtx, cancel := context.WithTimeout(ctx, shardTabletRefreshTimeout)
+		defer cancel()
+		_, _, err := topotools.RefreshTabletsByShard(rtbsCtx, ts.TopoServer(), ts.TabletManagerClient(), target.GetShard(), nil, ts.Logger())
```
mattlord, line 1300: This is after the point of no return.

go/vt/vtctl/grpcvtctldserver/server.go
deepthi: 👍

go/vt/wrangler/traffic_switcher.go
deepthi: sgtm

mattlord: Done: https://github.com/vitessio/vitess/pull/10058/commits/90f656fdb3c0ef4f3f60afa464f6b536ded5712b

deepthi (approved):
Nice work! lgtm

PR #10060, opened 2022-04-08
Use KILL QUERY instead of KILL in DBConn.Kill

### LINKED ISSUE #10059: Potential connection churn caused by DBConn.Kill()
We currently use `KILL` to kill the running query per https://github.com/vitessio/vitess/blob/main/go/vt/vttablet/tabletserver/connpool/dbconn.go#L363

Using KILL would cause the underlying Mysql connection to be closed, which means they need to recreated later on when the resource object is checked out. If there are many kills due to request cancellation, e.g. client just gives up after sending many requests, it could end up with a connection churn in the pool, i.e. Mysql connection is killed as a result of the cancellation but later on created to serve new request which in turn is cancelled. 

KILL QUERY just terminates the query while keeping the connection around, which could reduce the churn. The downside is that KILL QUERY was introduced in 5.7 so it will just break on older versions of Mysql.

wangmeng99:
KILL QUERY leaves the connection intact https://dev.mysql.com/doc/refman/5.7/en/kill.html so we can avoid connection open/close churn if we have to kill many queries.

Note that it only works for 5.7 and above.

Signed-off-by: Meng Wang <wangmeng@gmail.com>

## Description

KILL in DBConn.Kill() could cause connection churn see more in https://github.com/vitessio/vitess/issues/10059. This PR introduces `KILL QUERY` which makes the Mysql connection intact after terminating the running query. The catch is that KILL QUERY won't work on Mysql versions earlier than 5.7

## Related Issue(s)

Fixes #10059 10059

## Checklist


systay (approved):
LGTM 👍

mattlord (changes requested):
Thank you for working on this! I had a few nits, but beyond that my biggest concern is the usage for stateful connections here: https://github.com/vitessio/vitess/blob/8a88844e29b74adbba3c7f82525d2b5096aaab66/go/vt/vttablet/tabletserver/stateful_connection.go#L196-L199

When killing the connection you clear out any session state such as session variables, transactional state, temporary tables, etc. So for stateful sessions this is a significant change that could potentially break some assumptions and cause issues.

When shutting down we also seem to have some assumptions that the connections are killed here: https://github.com/vitessio/vitess/blob/8a88844e29b74adbba3c7f82525d2b5096aaab66/go/vt/vttablet/tabletserver/query_list.go#L81-L100

For these reasons, I feel like we should be intentional here and probably add a new function called `KillQuery` and use that when we explicitly want that behavior. But then I'm not sure where we call `Kill` *except* for those two locations:
```
$ git grep 'Kill(' | grep -viE "syscall|Mobile|test.go|libsonnet|Counters|tx|Set|killing|Process" 
go/vt/vttablet/tabletserver/connpool/dbconn.go:func (dbc *DBConn) Kill(reason string, elapsed time.Duration) error {
go/vt/vttablet/tabletserver/query_list.go:      Kill(message string, elapsed time.Duration) error
go/vt/vttablet/tabletserver/query_list.go:      qd.conn.Kill("QueryList.Terminate()", time.Since(qd.start))
go/vt/vttablet/tabletserver/query_list.go:              qd.conn.Kill("QueryList.TerminateAll()", time.Since(qd.start))
go/vt/vttablet/tabletserver/stateful_connection.go:func (sc *StatefulConnection) Kill(reason string, elapsed time.Duration) error {
go/vt/vttablet/tabletserver/stateful_connection.go:     return sc.dbConn.Kill(reason, elapsed)
```

In that case I think we should probably explore those two uses further to be sure we are not breaking any assumptions that may cause problems in certain cases. 

What do you think? I'm happy to help with that investigation and discussing it. We can do it here or via Slack, whichever you prefer.

systay:
@mattlord do you think it's worth saving these commits and finishing the PR on our own?

mattlord:
@systay I think we should hold off for now. I'm not saying it isn't something we may want to do, but it's not a priority. We can revisit in the future.

PR #10066, opened 2022-04-10
ComBinlogDumpGTID and downstream replication protocol

shlomi-noach:
## Description

**Update**: see https://github.com/vitessio/vitess/pull/10066#issuecomment-1099131530 ; this is now a general purpose PR to have better support for downstream replication.

A [COM_BINLOG_DUMP_GTID](https://dev.mysql.com/doc/internals/en/com-binlog-dump-gtid.html) request may specify a GTID, or it may specify a log file, or it may just leave both empty. These are all valid options.

As such, `Handler`'s `ComBinlogDumpGTID()` function should receive `logFile` and `logPos` parsed from `COM_BINLOG_DUMP_GTID` request, in addition to GTIDSet.

## Relates issues

https://github.com/vitessio/vitess/issues/11939

## Checklist


shlomi-noach:
Had a couple fixes/additions. Good to review.

shlomi-noach:
converted to draft: I'm experimenting offline with replication protocol and I notice there's a few changes to be made. Once I'm done experimenting I'll push back all changes.

shlomi-noach:
- Added `parseComBinlogDump`
- Added `BinlogEvent.IsHeartbeat()`
- Fixed `NewRotateEvent()` bug
- Added `NewFakeRotateEvent()`
- Added `NewHeartbeatEvent()`
- `mysql/conn.go` handles `ComBinlogDump`
- `mysql/conn.go` handles `ComRegisterReplica`
- Added `Conn.WriteBinlogEvent()`
- `SendBinlogDumpCommand()` accepts a binlog file name, per https://dev.mysql.com/doc/internals/en/com-binlog-dump-gtid.html

shlomi-noach:
All right, I'm submitting this while still experimenting offline with some replication stuff. Vitess' support for downstream replication protocol was incomplete and with a couple bugs -- which makes sense because at this time Vitess doesn't actually implement any of it.

So anyway, this PR fixes and adds many things around `COM_BINLOG_DUMP` and `COM_BINLOG_DUMP_GTID` including responding to said COMs, sending binlog events downstream, registering replicas; constructing rotate events, fake rotate events, heartbeat events.

deepthi:
> I should point out: there's no added tests here, as current behavior is untested in the first place. So code changes here don't really affect production because they're never used (though I changed a couple function signatures).
> 
> While I'm working offline on using the replication protocol, I'll be adding some tests - mainly `endtoend` for now.

Even so, it will be good to add unit tests if possible.

shlomi-noach:
Added just a couple unit tests. For the majority of the changes I don't see an existing framework for unit testing, as this is so close to MySQL communication bare bones. I can't see unit testing for parallel functionality to the ones I added. Otherwise, there's ample `endtoend` testing downstream (to be contributed upstream eventually).

shlomi-noach:
Disagree. This PR should stand.

shlomi-noach:
Added more tests. This PR is ready to review.

shlomi-noach:
Right now the use case is outside the scope of Vitess, as we are experimenting with code that hooks into the replication stream, and uses Vitess connectivity/functionality. The code is still in preliminary state, and we are not sure whether it will mature and be open sourced. In the meantime, we add replication protocol functionality inside Vitess, for completeness of the MySQL protocol.

The hook's objective is to read the replication stream. So this happens per-shard. The hook itself can be agnostic to Vitess sharding and only needs to know about MySQL events.

shlomi-noach:
Related issue: https://github.com/vitessio/vitess/issues/11939

harshit-gangal (approved):

deepthi (approved):

PR #10071, opened 2022-04-11
[vtadmin] full dynamic cluster config

ajm188:
## Description

This PR changes the unmarshalling technique used for dynamic cluster configs, to support a fully-configured dynamic cluster with parity to what you can do on the CLI.

The most notable changes for end users are:
- you must specify an `id` key rather than a `name` key. you _should_ also specify a `name` key, but it's not strictly required.
- you must specify your discovery implementation name. this means you can use dynamic clusters with _other_ discovery implementations (like consul!) "for free" 
- when using dynamic discovery with a dynamic cluster (boy that's a lotta "dynamic"), you are technically going through a double layer (mmmm, burritosssss) of json unmarshalling, which means your `discovery-dynamic-discovery` option must be properly quoted and escaped (see the tests).

Below are some screenshots where I uploaded first the original dynamic cluster, and then a dynamic cluster with some configuration options. Here are the configs used:

### minimal dynamic cluster
```json
{
    "id": "dynamic",
    "name": "my-dynamic-cluster",
    "discovery": "dynamic",
    "discovery-dynamic-discovery": "{\"vtctlds\": [ { \"host\": { \"fqdn\": \"localhost:15000\", \"hostname\": \"localhost:15999\" } } ], \"vtgates\": [ { \"host\": {\"hostname\": \"localhost:15991\" } } ] }"
}
```

### configured

```json
{
    "id": "dynamic2",
    "name": "my-configured-dynamic-cluster",
    "discovery": "dynamic",
    "discovery-dynamic-discovery": "{\"vtctlds\": [ { \"host\": { \"fqdn\": \"localhost:15000\", \"hostname\": \"localhost:15999\" } } ], \"vtgates\": [ { \"host\": {\"hostname\": \"localhost:15991\" } } ] }",
    "vtsql-discovery-timeout": "1h"
}
```

Pulling up the debug page for the configured cluster, we can see that the `vtsql` discovery timeout did indeed get set to 1h, rather than the default 100ms:

```
 curl -s localhost:14200/debug/cluster/dynamic2 | jq '.' | grep timeout      
      "discovery-timeout": "1h"
      "discovery_timeout": 100000000,
      "discovery_timeout": 3600000000000,
```

Screen Shot 2022-04-11 at 1 54 26 PM
Screen Shot 2022-04-11 at 1 56 46 PM

## Checklist

### DIFF
--- a/go/vt/vtadmin/api_test.go
+++ b/go/vt/vtadmin/api_test.go
@@ -4871,7 +4871,7 @@ func TestServeHTTP(t *testing.T) {
 		{
 			name:                  "multiple clusters with dynamic clusters",
 			enableDynamicClusters: true,
-			cookie:                `{"name": "dynamiccluster1", "vtctlds": [{"host":{"fqdn": "localhost:15000", "hostname": "localhost:15999"}}], "vtgates": [{"host": {"hostname": "localhost:15991"}}]}`,
+			cookie:                `{"id": "dynamiccluster1", "name": "dynamiccluster1", "discovery": "dynamic", "discovery-dynamic-discovery": "{\"vtctlds\": [{\"host\":{\"fqdn\": \"localhost:15000\", \"hostname\": \"localhost:15999\"}}], \"vtgates\": [{\"host\": {\"hostname\": \"localhost:15991\"}}]}"}`,
 			clusters: []*cluster.Cluster{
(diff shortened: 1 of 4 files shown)


notfelineit (approved):
🎉

PR #10074, opened 2022-04-12
Fix Gen4 group/order by with derived table

### LINKED ISSUE #10011: vtgate: Gen4 only_full_group_by regression
I think the Gen4 planner is erroring with using aliases in group by columns.

`(1055, "Expression of SELECT list is not in GROUP BY clause and contains nonaggregated column 'Count' which is not functionally dependent on columns in GROUP BY clause; this is incompatible with sql_mode=only_full_group_by")`

This query has worked previously on vtgate and still works on v13 with the V3 planner

```sql
SELECT
  latest_error as `Latest Error`,
  count(*) AS `Count`,
  sum(if(latest_tried_at BETWEEN FROM_UNIXTIME(1648578304) AND FROM_UNIXTIME(1648664704), 1, 0)) AS `Last Errored in Period`
FROM (
  select
    JSON_UNQUOTE(JSON_EXTRACT(attributes, '$.latestError')) as latest_error,
    TIMESTAMP(JSON_EXTRACT(attributes, '$.latestTriedAt')) as latest_tried_at
  from searches__extractor__msgs
  where time_next is not null
) as msgs_errors
WHERE latest_error IS NOT NULL
GROUP BY latest_error
ORDER BY `Count` desc
LIMIT 30;
```

I'm rolling out #10007 today, maybe the subquery is fixed there, otherwise that'll get me back to the Gen4 planner where I can do an `explain format=vitess`.

frouioui:
## Description

This PR fixes an issue (observed in #10011) that arises when the planner was rewriting expressions to their derived form, even though they did not need to be rewritten.

## Related Issue(s)

- fixes #10011

## Checklist


derekperkins:
Does this PR have to land to fix https://github.com/vitessio/vitess/issues/10011?

frouioui:
> Does this PR have to land to fix #10011?

@derekperkins Yes.

systay (approved):

PR #10080, opened 2022-04-12
VReplication: use db_filtered user for vstreams

mattlord:
## Description
When performing [VReplication workflows](https://vitess.io/docs/13.0/reference/vreplication/) — such as [`MoveTables`](https://vitess.io/docs/13.0/reference/vreplication/movetables/) — we need to stream table rows, query results, and binary log events from the source in order to copy data and changes to the target.

The [`com_binlog_dump_gtid`](https://dev.mysql.com/doc/internals/en/com-binlog-dump-gtid.html) command is used on the source side in order to stream the binary log events (vstreamer->binlog streamer) from the source tablet's mysqld to the vttablet for filtering before sending to the target vttablets (vcopier->vplayer). In order to execute this command the user needs the [`REPLICATION SLAVE`](https://dev.mysql.com/doc/refman/5.7/en/privileges-provided.html#priv_replication-slave) privilege.

A big issue this PR addresses is that the `db_app` user was used for this. Not only is this unexpected (non-intuitive and not documented), but it requires that the application user has replication privileges which is not the norm. In this PR we:
1. Use the `db_filtered` user — which already has the [`REPLICATION SLAVE`](https://dev.mysql.com/doc/refman/5.7/en/privileges-provided.html#priv_replication-slave) privilege — for streaming the binlog events ([binlog streamer](https://github.com/vitessio/vitess/blob/main/go/vt/binlog/binlog_streamer.go))
   - ℹ️  There are also SQL commands like `SHOW BINARY LOGS` executed in [related connection/code](https://github.com/vitessio/vitess/blob/main/go/vt/binlog/binlog_connection.go) that require the [`REPLICATION CLIENT`](https://dev.mysql.com/doc/refman/5.7/en/privileges-provided.html#priv_replication-client) privilege, which `db_filtered` already has
2. Use the `db_filtered` user [for all other base vstream types](https://github.com/vitessio/vitess/pull/10080/commits/bbdb16a5def4bdf3119b2dc60f40c2b586963c78): row streamer and results streamer
   - ℹ️ This also applies to more specialized vstreams such as used for [messaging](https://github.com/vitessio/vitess/blob/911ddc960067adab1f67d737af82e89902fcbd1d/go/vt/vttablet/tabletserver/messager/message_manager.go)
   - 💡 This then aligns the product with the existing documentation: https://vitess.io/docs/11.0/reference/vreplication/faq/
3. We remove the  [`REPLICATION SLAVE`](https://dev.mysql.com/doc/refman/5.7/en/privileges-provided.html#priv_replication-slave) privilege from the default `db_app` user
   - ⚠️ A follow-up ticket is needed to update [the embedded content in mysqlctl](https://github.com/vitessio/vitess/blob/911ddc960067adab1f67d737af82e89902fcbd1d/go/vt/mysqlctl/rice-box.go#L72-L77) (done via [`make embed_config`](https://github.com/vitessio/vitess/blob/911ddc960067adab1f67d737af82e89902fcbd1d/Makefile#L50-L53) which itself was broken), including this change and the recent [super_read_only change](https://github.com/vitessio/vitess/pull/9312) (see related issue below)

## Related Issue(s)
- Follow-up work: https://github.com/vitessio/vitess/pull/10094

## Checklist
- [x] Documentation updates: https://github.com/vitessio/website/pull/1011


rohit-nayak-ps (approved):
LGTM
nice catch

mattlord:
> LGTM
> nice catch

I increased the scope a bit to use the `vt_filtered` user for all base vstreamers: row, results, binlog. I believe that:
1. This simplifies future work for VReplication as we lock down the MySQL accounts used (for vstreams at least, it's all `db_filtered`)
2. This simplifies operations as VReplication is using a single MySQL user account for vstreams
3. The product then aligns with the docs: https://vitess.io/docs/14.0/reference/vreplication/faq/

That's why I requested a re-review. Sorry. It was a long round about adventure to find out why the PITR recovery tests (which use rippled) where failing (dbconfig parts of https://github.com/vitessio/vitess/pull/10080/commits/bbdb16a5def4bdf3119b2dc60f40c2b586963c78).

mattlord:
@derekperkins please let me know if you have any messaging related questions or concerns. Thanks!

rohit-nayak-ps (approved):
lgtm

PR #10085, opened 2022-04-13
Adds parsing for NOW in DEFAULT clause

### LINKED ISSUE #10064: sqlparser: error parsing/constructing temporal column with `DEFAULT (NOW())`
Consider the following valid SQL statement:
```sql
create table t (id int primary key, dt datetime DEFAULT (now()))
```
running this in MySQL:
```sql
mysql [localhost:8020] {msandbox} (an_issue) > create table t (id int primary key, dt datetime DEFAULT (now()));
Query OK, 0 rows affected (0.04 sec)

mysql [localhost:8020] {msandbox} (an_issue) > show create table t \G
       Table: t
Create Table: CREATE TABLE `t` (
  `id` int NOT NULL,
  `dt` datetime DEFAULT (now()),
  PRIMARY KEY (`id`)
```

notice that in `SHOW CREATE TABLE` the column has `DEFAULT (now())`, exactly like we've created.

`sqlparser.Parse()` parses the query correctly. However, `sqlparser.String()` outputs the following:
```sql
create table t (
  id int primary key,
  dt datetime default now()
)
```
notice `default now()` and not `default (now())`.

`sqlparser` is now _unable_ to parse the resulting query that it just generated. Consider this test:
```go

func TestParseNow(t *testing.T) {
	query := "create table t (id int primary key, dt datetime DEFAULT now())"
	stmt, err := Parse(query)
	assert.NoError(t, err, "Oh no! Got an error: %v", err)
	ddl, ok := stmt.(DDLStatement)
	assert.True(t, ok)
	assert.True(t, ddl.IsFullyParsed())
}
```

```shell
$ go test ./go/vt/sqlparser/... -run TestParseNow
W0410 11:08:12.939015 1613553 parser.go:89] ignoring error parsing DDL 'create table t (id int primary key, dt datetime DEFAULT now())': syntax error at position 60 near 'now'
--- FAIL: TestParseNow (0.00s)
    parse_test.go:4716: 
        	Error Trace:	parse_test.go:4716
        	Error:      	Should be true
        	Test:       	TestParseNow
FAIL
FAIL	vitess.io/vitess/go/vt/sqlparser	0.006s
?   	vitess.io/vitess/go/vt/sqlparser/goyacc	[no test files]
FAIL
```
Line `4716` refers to `assert.True(t, ddl.IsFullyParsed())`

The statement is valid and should be parsable. 

cc @vitessio/query-serving

GuptaManan100:
## Description

As seen in #10064, we were not able to parse the query - 
```sql
create table t (
  id int primary key,
  dt datetime default now()
)
```
even though this is a valid query and should work. This PR fixes this issue and adds parsing support for it.

## Related Issue(s)

- Fixes #10064  

## Checklist


shlomi-noach (approved):
:heart:

PR #10090, opened 2022-04-13
Gen4 feature: optimize OR queries to IN

systay:
## Description
When using a vindex is difficult because the necessary predicates are hidden behind `OR`, we can try and see if the `OR`s can be merged into an `IN` instead, which is easier to plan efficiently. 

## Checklist
- [x] Should this PR be backported? `nope`

### DIFF
--- a/go/vt/vtgate/planbuilder/physical/route_planning.go
+++ b/go/vt/vtgate/planbuilder/physical/route_planning.go
@@ -238,7 +238,85 @@ func createRoute(ctx *plancontext.PlanningContext, table *abstract.QueryTable, s
 		}
 	}
 
+	if plan.RouteOpCode == engine.Scatter && len(table.Predicates) > 0 {
+		// If we have a scatter query, it's worth spending a little extra time seeing if we can't improve it
+		for _, pred := range table.Predicates {
+			rewritten := tryRewriteOrToIn(pred)
+			if rewritten != nil {
+				err = plan.UpdateRoutingLogic(ctx, rewritten)
+				if err != nil {
+					return nil, err
+				}
+			}
+		}
+	}
+
 	return plan, nil
+}
+
+func tryRewriteOrToIn(expr sqlparser.Expr) sqlparser.Expr {
+	rewrote := false
+	newPred := sqlparser.Rewrite(sqlparser.CloneExpr(expr), func(cursor *sqlparser.Cursor) bool {
+		_, ok := cursor.Node().(*sqlparser.OrExpr)
+		return ok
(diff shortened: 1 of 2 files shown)


go/vt/vtgate/planbuilder/physical/route_planning.go
```diff
@@ -238,9 +238,87 @@ func createRoute(ctx *plancontext.PlanningContext, table *abstract.QueryTable, s
...
+			}
+			rgtCmp, ok := or.Right.(*sqlparser.ComparisonExpr)
+			if !ok {
+				return true
+			}
+
+			col, ok := lftCmp.Left.(*sqlparser.ColName)
+			if !ok || !sqlparser.EqualsExpr(lftCmp.Left, rgtCmp.Left) {
+				return true
+			}
```
frouioui, line 278: Can we do this for any type of node? For instance for `c+1`. If they are equal on the left and right comparisons, it should be safe to convert them to an `IN`

go/vt/vtgate/planbuilder/testdata/filter_cases.txt
```diff
@@ -3927,3 +3927,114 @@ Gen4 plan same as above
...
+  "Original": "select col from user where (id = 1 or id = 2) or (id = 3 or id = 4)",
+  "Instructions": {
+    "OperatorType": "Route",
+    "Variant": "IN",
+    "Keyspace": {
+      "Name": "user",
+      "Sharded": true
+    },
+    "FieldQuery": "select col from `user` where 1 != 1",
+    "Query": "select col from `user` where id = 1 or id = 2 or (id = 3 or id = 4)",
```
frouioui, line 4033: I don't understand why the query is not rewritten with the `IN` comparison. Are we printing the original query, but execute the query `IN` instead of `=`s ?

frouioui (commented):
That looks great to me! This is a nice change.
I am having trouble understanding why the plans are outputted this way, thus I didn't approve yet

go/vt/vtgate/planbuilder/testdata/filter_cases.txt
systay: we're not changing the query, only the routing

frouioui (approved):

go/vt/vtgate/planbuilder/physical/route_planning.go
systay: technically and semantically we could, but since these predicates are only used to update the planning logic, and if the LHS is not a bare column, we can't use it for a vindex (it's not [sargable](https://en.wikipedia.org/wiki/Sargable))

PR #10099, opened 2022-04-15
delete RPCs that were deprecated in 13.0

### LINKED ISSUE #8839: Inclusive Naming: delete deprecated RPCs
WARNING: This should only be done after release 13.0

All RPCs that are deprecated as part of #8837 can be deleted.

deepthi:
## Description
Delete all references to the RPCs listed in the linked issue

## Related Issue(s)
Fixes #8839 

## Checklist


mattlord:
Reading through this again: https://github.com/vitessio/enhancements/blob/main/veps/vep-1.md#example-2-deprecating-and-removing-an-obsolete-utility

Did we return an error e.g. if someone hit this code via a cmd-line command or API call? Does that matter in this case?

deepthi:
> Reading through this again: https://github.com/vitessio/enhancements/blob/main/veps/vep-1.md#example-2-deprecating-and-removing-an-obsolete-utility
> 
> Did we return an error e.g. if someone hit this code via a cmd-line command or API call? Does that matter in this case?

The older versions of the RPCs are only called from the command-line of older version vtctlds. We have no way of going back and printing a warning from there.
If they upgrade their vtctld to a newer version (13+), it uses the new RPCs.
The 2-version process we follow for RPCs is meant to make this completely transparent to users, so we don't even try to print a warning. IOW, as long as you stay +/-1 version among the various vitess components, there will be no errors. If you have version drift beyond that, then there are no guarantees. You will in fact get an error.

I don't think we officially support direct RPC calls to vttablets from clients right now, so we don't worry about that.

go/test/fuzzing/vttablet_fuzzer.go
```diff
@@ -291,17 +291,6 @@ func (fs *fuzzStore) callExecuteFetchAsApp() error {
...
-// callInitMaster implements a wrapper
-// for fuzzing InitMaster
-func (fs *fuzzStore) callInitMaster() error {
-	tablet, err := fs.getTablet()
-	if err != nil {
-		return err
-	}
-	_, _ = fs.client.InitMaster(context.Background(), tablet, false)
-	return nil
-}
```
mattlord, line 303: We shouldn't rename this? I don't see an existing `callInitPrimary()` function. In the vttablet_fuzzer we seem to be removing functions rather than renaming/replacing. Is that intentional and correct? Otherwise the changes LGTM.

mattlord (commented):
I had a question about the vttablet fuzzer as there we seem to be removing rather than renaming/replacing, but otherwise this LGTM.

go/test/fuzzing/vttablet_fuzzer.go
deepthi: yeah we should rename if there isn't already one. I'll fix it.

deepthi: Renamed the ones where the new RPCs weren't already included.

mattlord (approved):
Thanks again!

PR #9497, opened 2022-01-11
Fixed legacy vreplication data that may be using master tablet type

arvind-murty:
## Description

When people update to a new version, any instances of "master" in the vreplication table should be changed to "primary".

## Checklist
- [ ] Should this PR be backported?
- [x] Tests were added or are not required
- [x] Documentation was added or is not required

PR #10100, opened 2022-04-15
remove changeMasterToPrimary, we no longer need it

### LINKED ISSUE #9514: Task: migrate data in vreplication table
We have made changes in the code to expect tablet_type to be primary instead of master.
There may be existing VReplication flows using master tablet_type that break when the Vitess version is upgraded.
To facilitate such migrations, we need to do the following:

- [x] In 13.0 add a repair query for vreplication table
- [ ] In 14.0 remove the repair query

deepthi:
## Description
In 13.0 we added a migration sql to change existing vreplication metadata.
We should not need that any longer.

## Related Issue(s)
Fixes #9514
Related to #9497 

## Checklist


rohit-nayak-ps (approved):
lgtm

PR #10102, opened 2022-04-15
MoveTables: adjust datetimes when importing from non-UTC sources into UTC targets

rohit-nayak-ps:
## Description

`MoveTables` is a common technique used to import data into Vitess. Many users run their MySQL servers using the `SYSTEM` timezone and/or storing `datetime` fields in local time. Vitess recommends running clusters in UTC, expecting apps to do the conversion to local time using, for example, `convert_tz()` in queries, `:application_timezone` in the mysql2 driver, or setting `@@session.time_zone` if using only `timestamp` data types.

Datetime values are usually stored using `datetime` or `timestamp` columns. MySQL stores `timestamp` s internally using UTC, but `datetime` is effectively a string value which is timezone-unaware. 

If you import data from a server running in a different time zone into a server using UTC. `timestamp`s will be correctly reproduced but datetimes can result in bugs which are difficult to track and fix. One such scenario is:

* User is running a cluster in Berlin, using SYSTEM time zone with the server time zone being local.
* The table contains a datetime column with `current_timestamp` as the default.
* On the source any inserts into the table will contain the current Berlin local time.
* The table is moved into a Vitess cluster, also using SYSTEM but with the UTC time zone.
*  Today the datetimes will be copied as-is.
* The next insert will store the current time in UTC. 
* So some values will be in the Berlin time zone and others in UTC. Since the datetime column has no time zone information the app has no way to differentiate between these.

This PR adds an option to MoveTables to specify a `-source_time_zone`. If specified, VReplication will convert `datetime` values into UTC before storing it. 

#### User Facing
* a new option `-source_time_zone` added to MoveTables.
* `Workflow Show` now displays SourceTimeZone

#### Internal
* a new attribute `SourceTimeZone` has been added to `BinlogSource` (whose proto is stored in the `source` column of `_vt.vreplication`).
* VDiff has been modified to use `BinlogSource.SourceTimeZone` to correctly compare datetimes.
* an e2e test `TestMoveTablesTZ` reflects the supported use case

### Not supported
* non-datetime fields, such as textual fields which store datetimes: conversion is only for `datetime` data types
* tables where datetime columns are stored in varying time zones: you can either convert all datetimes or none
* automatically doing  the conversion to local time in Vitess queries: the application has to be locale-aware and has to itself handle the conversion from UTC timestamps
* automatically detecting the source timezone: while this is possible we don't want to always convert datetime fields because we don't know what time zone (local or utc) the application has stored the values in.

## Checklist

-   [ ] Documentation was added or is not required - TBD


go/vt/vttablet/tabletmanager/vreplication/table_plan_builder.go
```diff
@@ -669,9 +675,17 @@ func (tpb *tablePlanBuilder) generateValuesPart(buf *sqlparser.TrackedBuffer, bv
...
 				buf.Myprintf("convert(%v using utf8mb4)", cexpr.expr)
-			} else {
+			case querypb.Type_DATETIME:
+				sourceTZ := tpb.source.GetSourceTimeZone()
+				if sourceTZ != "" {
+					buf.Myprintf("convert_tz(%v, '%s', '%s')", cexpr.expr, sourceTZ, "UTC")
+				} else {
+					buf.Myprintf("%v", cexpr.expr)
+				}
+			default:
```
shlomi-noach, line 688: :+1:  Just a question: is `UTC` preferable to `+00:00`? No opinion myself.

shlomi-noach (approved):
LGTM. The actual code change is some 10 lines, with a gazillion changes on top just to support the new information and to validate `vdiff`.

rohit-nayak-ps:
@shlomi-noach, I just realized that we are only supporting the forward migration of data at the moment: reverse replication streams are not setup correctly. Many users do like to run reverse streams into existing setups for a while before switching over entirely to Vitess. So we will need to add a `TargetTimeZone` in addition to the current `SourceTimeZone` in the BinlogSource proto for the reverse vreplication streams to work correctly.

It is a simple change and I had instinctively started with that design even for forward workflows, but thought I was over-designing since we wanted to fix the target to UTC.

Once this is implemented it means we can intrinsically support any time zone on source and target sides, but for now MoveTables will only allow the source time zone to be specified and default target time zone to UTC.

go/vt/vttablet/tabletmanager/vreplication/table_plan_builder.go
rohit-nayak-ps: > +1 Just a question: is `UTC` preferable to `+00:00`? No opinion myself.

Based on my comment below about reverse streams, the UTC timezone will make it to the proto in the vreplication table. So it is probably better to leave the `UTC` timezone code in there rather than the `+00:00` offset as a nudge to avoid setting up the source time zones in terms of offsets (to avoid DST-related issues)

shlomi-noach:
>  I just realized that we are only supporting the forward migration of data at the moment: reverse replication streams are not setup correctly.

Oh that's right! Sounds good.

shlomi-noach (approved):
:rocket:

go/test/endtoend/vreplication/time_zone_test.go
```diff
@@ -0,0 +1,194 @@
...
+	require.NoError(t, err)
+	require.NotNil(t, qrSourceUSPacific)
+
+	qrTargetUTC, err := customerTab.QueryTablet(query, targetKs, true)
+	require.NoError(t, err)
+	require.NotNil(t, qrTargetUTC)
+
+	require.Equal(t, len(qrSourceUSPacific.Rows), len(qrTargetUTC.Rows))
+
+	pst, err := time.LoadLocation("US/Pacific")
```
mattlord, line 127: Nit, but isn't PST Pacific Standard Time vs. PDT or Pacific Daylight Time? It makes it confusing for me when we later check to see if `pst` is PST or PDT. :-) 

https://github.com/vitessio/vitess/pull/10102/files#diff-b44124544b9ac2d695bf061044407b2f57699f9a63e7ecc2769f6d5a5aac1260R150-R153

go/vt/vttablet/tabletmanager/vreplication/table_plan_builder.go
```diff
@@ -669,9 +675,18 @@ func (tpb *tablePlanBuilder) generateValuesPart(buf *sqlparser.TrackedBuffer, bv
...
-			if cexpr.colType == querypb.Type_JSON {
+			switch cexpr.colType {
+			case querypb.Type_JSON:
 				buf.Myprintf("convert(%v using utf8mb4)", cexpr.expr)
-			} else {
+			case querypb.Type_DATETIME:
+				sourceTZ := tpb.source.SourceTimeZone
+				targetTZ := tpb.source.TargetTimeZone
+				if sourceTZ != "" && targetTZ != "" {
+					buf.Myprintf("convert_tz(%v, '%s', '%s')", cexpr.expr, sourceTZ, targetTZ)
```
mattlord, line 685: I don't see where we ever validate that the user specified a valid named time zone for the flag? And if the specified something invalid:
```
mysql> select convert_tz("2022-04-26 00:02:33", "US/Eastern", "UTC");
+--------------------------------------------------------+
| convert_tz("2022-04-26 00:02:33", "US/Eastern", "UTC") |
+--------------------------------------------------------+
| 2022-04-26 04:02:33                                    |
+--------------------------------------------------------+
1 row in set (0.00 sec)

mysql> select convert_tz("2022-04-26 00:02:33", "US/Nosir", "UTC");
+------------------------------------------------------+
| convert_tz("2022-04-26 00:02:33", "US/Nosir", "UTC") |
+------------------------------------------------------+
| NULL                                                 |
+------------------------------------------------------+
1 row in set (0.00 sec)
```

I'm assuming that VDiff will catch this? Still, IMO we should validate the flag before we get here. We can be sure that the target MySQL *both* has the time zone info tables loaded, *and* that the user specified a valid one this way: 
```
mysql> SET session time_zone = 'US/Nosir';
ERROR 1298 (HY000): Unknown or incorrect time zone: 'US/Nosir'

mysql> SET session time_zone = 'US/Pacific';
```

go/vt/wrangler/vdiff.go
```diff
@@ -456,6 +460,66 @@ func findPKs(table *tabletmanagerdatapb.TableDefinition, targetSelect *sqlparser
...
 }
 
+// If SourceTimeZone is defined in the BinlogSource, the VReplication workflow would have converted the datetime
+// columns expecting the source to have been in the SourceTimeZone and target in TargetTimeZone. We need to do the reverse
+// conversion in VDiff before comparing to the source
+func (df *vdiff) adjustForSourceTimeZone(targetSelectExprs sqlparser.SelectExprs, fields map[string]querypb.Type) sqlparser.SelectExprs {
+	if df.sourceTimeZone == "" {
+		return targetSelectExprs
+	}
+	log.Infof("found source time zone %s", df.sourceTimeZone)
```
mattlord, line 470: Nit, but I think we're using what was specified right? Rather than what we found e.g. to be in use in MySQL.

go/vt/wrangler/vdiff.go
```diff
@@ -456,6 +460,66 @@ func findPKs(table *tabletmanagerdatapb.TableDefinition, targetSelect *sqlparser
...
+	var modified bool
+	for _, expr := range targetSelectExprs {
+		converted := false
+		switch selExpr := expr.(type) {
+		case *sqlparser.AliasedExpr:
+			if colAs, ok := selExpr.Expr.(*sqlparser.ColName); ok {
+				var convertTZFuncExpr *sqlparser.FuncExpr
+				colName := colAs.Name.Lowered()
+				fieldType := fields[colName]
+				if strings.EqualFold(fieldType.String(), "DATETIME") {
```
mattlord, line 481: Curious why we do this instead of looking at the actual type? The UCA string comparison will be much more expensive.

go/vt/wrangler/vdiff.go
```diff
@@ -456,6 +460,66 @@ func findPKs(table *tabletmanagerdatapb.TableDefinition, targetSelect *sqlparser
...
+					convertTZFuncExpr = &sqlparser.FuncExpr{
+						Name: sqlparser.NewColIdent("convert_tz"),
+						Exprs: sqlparser.SelectExprs{
+							expr,
+							&sqlparser.AliasedExpr{Expr: sqlparser.NewStrLiteral(df.targetTimeZone)},
+							&sqlparser.AliasedExpr{Expr: sqlparser.NewStrLiteral(df.sourceTimeZone)},
```
mattlord, line 487: We wouldn't catch the case where the user specified an invalid source time zone then, right? Because we'd be comparing NULL to NULL, I think.

go/vt/wrangler/vdiff.go
```diff
@@ -456,6 +460,66 @@ func findPKs(table *tabletmanagerdatapb.TableDefinition, targetSelect *sqlparser
...
+
+func getColumnNameForSelectExpr(selectExpression sqlparser.SelectExpr) (string, error) {
+	aliasedExpr := selectExpression.(*sqlparser.AliasedExpr)
+	expr := aliasedExpr.Expr
+	var colname string
+	switch t := expr.(type) {
+	case *sqlparser.ColName:
+		colname = t.Name.Lowered()
+	case *sqlparser.FuncExpr: // only in case datetime was converted using convert_tz()
+		colname = aliasedExpr.As.String()
```
mattlord, line 516: Curious why we force lower case in the former case but not here?

mattlord (commented):
Overall it LGTM. Nice work on this! 

I do think that we should add a test for when an invalid named time zone is specified for the flag, and I think that we should check for that and handle it in the command too. 

Let's chat about it on Slack if I missed or misunderstood anything. 

Thanks!

go/vt/vtctl/vtctl.go
```diff
@@ -2346,6 +2346,10 @@ func commandVRWorkflow(ctx context.Context, wr *wrangler.Wrangler, subFlags *fla
 	excludes := subFlags.String("exclude", "", "MoveTables only. Tables to exclude (comma-separated) if -all is specified")
 	sourceKeyspace := subFlags.String("source", "", "MoveTables only. Source keyspace")
 
+	// if sourceTimeZone is specified, the target needs to have time zones loaded
+	// note we make an opinionated decision to not allow specifying a different target time zone than UTC.
+	sourceTimeZone := subFlags.String("source_time_zone", "", "MoveTables only. Specifying this causes any DATETIME fields to be converted from given time zone into UTC")
```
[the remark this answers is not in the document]
rohit-nayak-ps: I think we can leave this to the user: they may be populating datetimes in the app in a different time zone than system.

go/test/endtoend/vreplication/time_zone_test.go
rohit-nayak-ps: Changed variable name to `pacificLocation`

go/vt/vttablet/tabletmanager/vreplication/table_plan_builder.go
rohit-nayak-ps: Added a validation, so that MoveTables will return an error to the user if time zone is not found on the _current_ target primaries. Note that it is always possible for a new tablet to become the primary: we can, of course, check every current replica, but user can always add a new replica without loading the time zone ...

go/vt/wrangler/vdiff.go
rohit-nayak-ps: changed

go/vt/wrangler/vdiff.go
rohit-nayak-ps: fixed

go/vt/wrangler/materializer.go
```diff
@@ -1305,3 +1312,28 @@ func (mz *materializer) forAllTargets(f func(*topo.ShardInfo) error) error {
 	wg.Wait()
 	return allErrors.AggrError(vterrors.Aggregate)
 }
+
+// checkIfTimeZoneIsLoaded is a light-weight consistency check to validate that, if a source time zone is specified to MoveTables,
+// that the current primary has the time zone loaded in order to run the convert_tz() function used by VReplication to do the
+// datetime conversions. We only check the current primaries on each shard and note here that it is possible a new primary
+// gets elected: in this case user will either see errors during vreplication or vdiff will report mismatches.
+func (mz *materializer) checkIfTimeZoneIsLoaded(ctx context.Context, tz string, keyspace string) error {
```
mattlord, line 1320: nit but I think a better name is `checkTZConversion` or `validateTZConversion` since we're ensuring that the named time zone info is available and the specified source timezone is a valid name.

go/vt/wrangler/materializer.go
```diff
@@ -1305,3 +1312,28 @@ func (mz *materializer) forAllTargets(f func(*topo.ShardInfo) error) error {
...
+			return vterrors.Wrapf(err, "GetTablet(%v) failed", target.PrimaryAlias)
+		}
+		query := fmt.Sprintf("select convert_tz(now(), %s, 'UTC')", encodeString(tz))
+		qrproto, err := mz.wr.tmc.ExecuteFetchAsApp(ctx, targetPrimary.Tablet, false, []byte(query), 1)
+		if err != nil {
+			return vterrors.Wrapf(err, "ExecuteFetchAsApp(%v, %s)", targetPrimary.Tablet, query)
+		}
+		qr := sqltypes.Proto3ToResult(qrproto)
+		gotDate := qr.Rows[0][0].ToString()
+		if gotDate == "" {
```
mattlord, line 1333: I'm pretty sure (from looking at the `sqltypes.Value` code) that `gotDate` will be the string literal "NULL" here when we are not able to perform the conversion because either of the specified named time zones could not be found (either because one was invalid or the time zone tables have not been loaded). 

So I think that we should check for `gotDate == "NULL"` or **_even better_** to confirm we got a valid date and provide a more informative error message when we did not:
```
		dateFormat := "2006-01-02 15:04:05"
 		if gotDate, err := time.Parse(dateFormat, qr.Rows[0][0].ToString()); err != nil {
                       return fmt.Errorf("unable to perform time_zone conversions from %s to UTC — result of the attempt was: %s. Either the specified source time zone is invalid or the time zone tables have not been loaded on the %s tablet", tz, gotDate, targetPrimary.Alias)
 		}
```

mattlord (changes requested):
Thank you, @rohit-nayak-ps! I had one specific request, we can clear that one up quickly and then it's good!

go/vt/wrangler/materializer.go
mattlord: ❤️

mattlord (approved):
Thank you for hanging in there with me. ❤️

PR #10103, opened 2022-04-15
Only start SQL thread temporarily to WaitForPosition if needed

mattlord:
## Description

It https://github.com/vitessio/vitess/pull/9512 we blindly attempted to start the replication `SQL_Thread`(s) when waiting for a given replication position. The problem with this, however, is that [if the `SQL_Thread` is running but the `IO_Thread` is not then the tablet repair does not try and start replication on a replica tablet](https://github.com/planetscale/vitess/blob/ef7363e918e5c7093d72f15471f4c0d408ac0d10/go/vt/vttablet/tabletmanager/replmanager.go#L106-L117). So in certain states such as when initializing a shard, replication may end up in a non-healthy state and never be repaired.

1. We only attempt to start the `SQL_Thread` if it's not already healthy
2. We use [`START SLAVE SQL_THREAD UNTIL SQL_AFTER_GTIDS`](https://dev.mysql.com/doc/refman/en/start-slave.html) instead of `START SLAVE SQL_THREAD` so that the SQL_Thread will be stopped once we reach the desired position. This will allow the tablet repair to repair the replica when needed.

## Related Issue(s)
  - Caused by: https://github.com/vitessio/vitess/pull/9512

## Checklist

### DIFF
--- a/go/mysql/flavor.go
+++ b/go/mysql/flavor.go
@@ -70,6 +70,10 @@ type flavor interface {
 	// to run until `pos` is reached. After reaching pos, replication will be stopped again
 	startReplicationUntilAfter(pos Position) string
 
+	// startReplicationSQLUntilAfter will restart replication's SQL_thread, but only allow it
+	// to run until `pos` is reached. After reaching pos, it will be stopped again
+	startReplicationSQLUntilAfter(pos Position) string
+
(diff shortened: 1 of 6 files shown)


mattlord:
Moved here: https://github.com/vitessio/vitess/pull/10104

PR #10104, opened 2022-04-15
Only start SQL thread temporarily to WaitForPosition if needed

mattlord:
## Description

After https://github.com/vitessio/vitess/pull/9512 we always attempted to start the replication `SQL_Thread`(s) when waiting for a given position. The problem with this, however, is that [if the `SQL_Thread` is running but the `IO_Thread` is not then the tablet repair does not try and start replication on a replica tablet](https://github.com/planetscale/vitess/blob/ef7363e918e5c7093d72f15471f4c0d408ac0d10/go/vt/vttablet/tabletmanager/replmanager.go#L106-L117). So in certain states such as when initializing a shard, replication may end up in a non-healthy state and never be repaired.

1. We only attempt to start the `SQL_Thread`(s) if it's not already running
2. If we explicitly start the `SQL_Thread(s)` then we also explicitly reset it to what it was (stopped) as we exit the call

Because the caller should be/have a TabletManager which has a mutex, this should ensure that the replication manager calls are serialized and because we are resetting the replication state after mutating it, everything should work as it did before https://github.com/vitessio/vitess/pull/9512 with the exception being that when waiting we ensure that the replica at least has the possibility of catching up.

## Related Issue(s)
  - Follow-up to: https://github.com/vitessio/vitess/pull/9512

## Checklist

### DIFF
--- a/go/mysql/flavor.go
+++ b/go/mysql/flavor.go
@@ -255,7 +255,8 @@ func (c *Conn) StartReplicationUntilAfterCommand(pos Position) string {
 	return c.flavor.startReplicationUntilAfter(pos)
 }
 
-// StartReplicationSQLUntilAfterCommand returns the command to start the replication's SQL thread.
+// StartReplicationSQLUntilAfterCommand returns the command to start the replication's SQL
+// thread and have it run until it has reached the given position, at which point it will stop.
 func (c *Conn) StartReplicationSQLUntilAfterCommand(pos Position) string {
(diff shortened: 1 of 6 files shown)


GuptaManan100 (approved):
This is an amazing catch and fix! I messed up the first time around. Then we had thought that we would only be wanting to wait for position _only_ if there was some way to reach it. But, I see now that we mutated the state of replication, which this PR fixes! Thankyou for this. 💯💕

PR #10107, opened 2022-04-19
OnlineDDL: skip GetSchema where possible

shlomi-noach:
## Description

In `ApplySchema`, there is a check for big schema changes. This is a legacy check from a time when we did not support online DDL. Big-schema-changes is _mostly_ skipped when there are online DDLs. However, one step is not skipped, which is `GetSchema`. 

Currently, we first `GetSchema`, then iterate all DDLs, and for each DDL check if it is online or not, and if not  - whether it is a "big change".

When all DDLs are online, the `GetSchema` step is wasteful. It reads all tables with some metadata.

In this PR, we first do a preliminary check to see whether there is at all any non-online DDL. If there is at least one non-online DDL, we proceed with `GetSchema`. But if all DDLs are online - we skip `GetSchema` altogether.

## Related Issue(s)

- #6926 

## Checklist


rohit-nayak-ps (approved):
lgtm

PR #10108, opened 2022-04-19
[vtadmin] Refactor api.getSchemas => cluster.GetSchemas

ajm188:
## Description

This is part 1 of a preparatory refactor to get caching in front of /schema related endpoints (which I'm writing up an issue for and will post here when I have it). The main effort is to get all of the fetching logic happening at the cluster-level instead of the api-level, so that we can have per-cluster caches.

## Related Issue(s)

- TK

## Checklist

    - (my first pass at this refactor actually broke several schema-related tests, which makes me pretty confident i've got it right this time -- also I tested locally)

### DIFF
--- a/go/vt/vtadmin/cluster/cluster.go
+++ b/go/vt/vtadmin/cluster/cluster.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"database/sql"
 	"encoding/json"
+	stderrors "errors"
 	"fmt"
 	"math/rand"
 	"sort"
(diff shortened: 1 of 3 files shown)


go/vt/vtadmin/cluster/cluster.go
```diff
@@ -1026,6 +1030,148 @@ func (c *Cluster) GetSchema(ctx context.Context, keyspace string, opts GetSchema
...
+		}
+	}()
+
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+
+		// TODO: (ajm188) we can't use c.GetKeyspaces because it also makes a
+		// FindAllShardsInKeyspace call for each keyspace, which we may or may
+		// not need. Refactor that method so we can get better code reuse.
```
ajm188, line 1088: This comes from [here](https://github.com/vitessio/vitess/pull/10108/files#diff-3e212bf9fa780da22b2bbce441ebb9ed37a8c590bfa017c13ecb6e8b245fd5e3L847-L856) (which, fun fact, was _also_ buggy because it bypassed the topo read pool 😭 ... i have put so many bugs in here and i have .... Plans to clean this up, but I'll do that separately.

PR #10116, opened 2022-04-20
Vitess online ddl: modify a column from textual to non-textual, ignore charset

shlomi-noach:
## Description

In Online DDL, in `vitess` strategy, modifying a textual column to an `INT` (and other types) results with:
```
Character set  not supported for column <column name>
```
This is due to a bug that attempts to validate the character set on the target column. There is none, because the column is converted away from being textual in the first place.

In this fix we verify whether the target column is textual.

Added tests to the suite to support the change.

## Checklist


dbussink (approved):

shlomi-noach:
DCO seems to be stalling. I've verified manually that commits are signed off. Merging this PR without waiting for DCO.

Issue #10093, opened 2022-04-13
Bug Report: Zookeeper unit tests are failing on Mac M1


rsajwani:
### Overview of the Issue

Two Issues

Issue #1.  'Make Tools' is failing on Mac M1. The error is related to zookeeper

build/classes
    [javac] warning: [options] bootstrap class path not set in conjunction with -source 6
    [javac] error: Source option 6 is no longer supported. Use 7 or later.
    [javac] error: Target option 6 is no longer supported. Use 7 or later.

BUILD FAILED
/Users/rameezsajwani/Code/fork/vitess/dist/vt-zookeeper-3.4.14/zookeeper-3.4.14/build.xml:321: Compile failed; see the compiler error output for details.

Issue #2 Zookeeper unit test cases are failing on local box. 

### Reproduction Steps

Run
make tools

Run
go test vitess.io/vitess/go/vt/topo/zk2topo

### Binary Version

```sh
Main
```

### Operating System and Environment details

```sh
Mac M1
```

deepthi:
@rsajwani can this be closed? #10117 should have fixed this, correct?

PR #10117, opened 2022-04-20
Changes for make tools

rsajwani:

## Description

1. 'Make Tools' on Mac M1 is failing
2. Following unit tests were failing on Mac M1
- go test vitess.io/vitess/go/vt/topo/k8stopo
- go test vitess.io/vitess/go/vt/zkctl
- go test vitess.io/vitess/go/vt/topo/zk2topo
3. Updated versions of zk, etcd, consul for distribution.
- check out my PR https://github.com/vitessio/vitess-resources/pull/8
- New vitess resource release https://github.com/vitessio/vitess-resources/releases/tag/v2.0

## Related Issue(s)
https://github.com/vitessio/vitess/issues/10093

## Checklist

-   [ ] "Backport me!" label has been added if this change should be backported. NO

## Testing

1. Ran 'Make Tools' on Mac M1 
2. Ran 'Make Tools' on Linux
3. Ran 'Make Tools' on Mac but non-M1 machine.


deepthi (commented):
I expected `ensure_bootstrap_updated` check to fail because we are changing bootstrap.sh.

Looks like the path filtering is incorrect and we need to fix it by adding `bootstrap.sh` to the list here:
https://github.com/vitessio/vitess/blob/main/.github/workflows/ensure_bootstrap_updated.yml#L18

We also need to build and push new bootstrap images (and update the bootstrap image version in the Makefile).

GuptaManan100 (approved):
This looks good to me, but the proto check is failing.

vmg:
Changes look fine, but since you forgot to bump the bootstrap version in `test.go`, it won't be enough to simply re-build and push the docker images now. The caches in CI won't pick up the changes. You're going to need to bump the version _again_ everywhere, and then push those images. That will make the build pass.

.github/workflows/ensure_bootstrap_updated.yml
```diff
@@ -19,6 +19,7 @@ jobs:
             - 'docker/**'
             - 'test.go'
             - 'Makefile'
+            - 'bootstrap.sh'
```
frouioui, line 22: ❤️

frouioui (approved):
looks good to me!

test.go
```diff
@@ -74,7 +74,7 @@ For example:
 // Flags
 var (
 	flavor           = flag.String("flavor", "mysql57", "comma-separated bootstrap flavor(s) to run against (when using Docker mode). Available flavors: all,"+flavors)
-	bootstrapVersion = flag.String("bootstrap-version", "6", "the version identifier to use for the docker images")
+	bootstrapVersion = flag.String("bootstrap-version", "7", "the version identifier to use for the docker images")
```
frouioui, line 77: @vmg, the bootstrap version is bumped in `test.go`

vmg: Yes, but that happened after @deepthi pushed the docker images, so they need to be pushed _again_ (this is my understanding from reading the Slack logs).

deepthi: Makes sense. we'll do this today.

rsajwani: I have push a commit with all the files updated to 8 .. @deepthi

deepthi: Pushed new bootstrap images.

PR #10120, opened 2022-04-20
[vtadmin] schema cache

### LINKED ISSUE #10109: RFC: VTAdmin schema caching
# Motivation and Rationale

In larger vtadmin / vitess deployments, schemas represent some of the largest payloads that have to be shuttled from VtctldServer in cluster(s) back to vtadmin. When there are a bunch of tables, or we have to aggregate across a bunch of shards, or when clusters are geographically far from vtadmin-api, this can make loading schema endpoints take a much longer time than other endpoints (especially because we wait for every cluster to respond, so there's no real way to "eager load" the faster clusters without us knowing in advance that they will be faster).

Schemas _also_ don't change that often (the sizes do on heavy-write tables, but if we're off by a few rows/megs for a bit, who cares _that_ much (assuming there's a way to force a refresh)), so this is a good candidate to cache.

# Design

## Where to cache

We want to cache schemas at the cluster-level (so each cluster will get a `schemaCache` field), _not_ at the API-level (where we would have a single `schemaCache` field on the `vtadmin.API` struct). We want to do this because:

- We will expose (or _could_ expose) configuration options for caching behavior (how long to cache, whether to background refresh/fill, and so on), and caching per-cluster makes finer-grained control much easier to provide to end users.
- The semantics just _feel_ better; the schema lives at the cluster level, so that's where it should be cached. I suspect that trying to have a single global cache will make things feel clunkier (from the vtadmin–developer perspective) and possibly back ourselves into entangled bits of code that are harder to refactor. It's just a less clean api (not `vtadmin.API`, but "api" api) boundary definition.

## How to cache

We want to cache the _full_ schema payload (so with full size aggregation, all the tables, etc). If a request comes in for a subset of that info, we'll fetch it (or pull from the cache and then extract the subset out of the cached data), and in a background goroutine, spin up a second request to fetch the full payload to fill in the cache.

### Avoiding spamming requests

In order to avoid spamming requests (if 10 requests for /schemas/<mycluster>/<mytable> come in, we could accidentally make 10 background threads in quick succession to "fill" the cache, when 1 would have done), we'll use a single background goroutine that listens for "fill the cache please" requests on a channel, roughly like:

```go
func New() *Cluster {
    c := &Cluster{...}
    c.schemaCacheCh := make(chan *fillSchemaCacheRequest, someBufferSize)
    c.schemaCacheFillCtx, c.schemaCacheFillCancel = context.WithCancel(context.Background())
    go c.schemaCacheFiller()

    return c
}

func (c *Cluster) schemaCacheFiller() {
    for {
        select {
        case <-c.schemaCacheFillCtx.Done():
                return
        case in := <-c.schemaCacheCh:
                // do some de-duping of similar requests (i.e. don't fill for the same table within 30s)
                c.GetSchema(in)
        }
    }
}
```

### Double caching

Full disclosure: haven't come up with a solution for this one yet, but I need to document it as a potential issue.

Because we want to cache the full size aggregation (which includes the rollup as well as the per-shard data), we have a potential issue with the rollup specifically: a request can specify whether to include non-serving shards or not, which would result in different rollup values.

I _think_ the simplest solution here is to simply fill twice, and incorporate the value for `req.TableSizeOptions.IncludeNonServingShards` in the cache key.

### FindSchema cache

`FindSchema` requests are different from GetSchema(s) by definition, because they don't necessarily know which cluster(s) are involved.

The way I'm currently thinking about how to address this is to have a _second_ cache that maps table names to lists cluster IDs (plural, if multiple clusters have a table with that name). Then the cached version of `FindSchema` looks like (roughly):

```go
func (api *API) FindSchema(ctx context.Context, req *vtadminpb.FindSchemaRequest) (*vtadminpb.FindSchemaResponse, error) {
    clusterIDs := req.ClusterIds

    if api.tableToClustersCache.Has(req.TableName) {
        clustersWithTable := api.tableToClustersCache.Get(req.TableName).([]string)
        // do a set difference with req.ClusterIds and clustersWithTable, assign to clusterIDs
    }

    clusters, _ := api.getClustersForRequest(clusterIDs)
    for _, c := range clusters {
        // do a schema lookup (concurrently), and fill in the lookup cache with the final results
    }
}
```

ajm188:
## Description

This PR introduces a caching mechanism for vtadmin, with our first use-case being schemas.

It does not use the `go/cache` package, primarily because I wanted a mechanism to enqueue backfills as a first-class concern, which I'll talk about a bit later.

Heeeeeeeeeeere we go:

### Cache Design

The cache design is pretty much what you would expect in generics: we store a mapping of keys (of type `K`) to values (of type `V`, which can be `any`-thing).

You might expect the cache to be defined as `Cache[string, V any]`, but that requires every callsite to know how to turn _something_ into a key.
By pushing this into an interface (`Keyer`) that keys have to implement, there can only be one place (the `Key()` method) where cache keys are created, which reduces duplication of code and therefore the chance two places accidentally use different key schemes for the same cache.
Further, I chose not to leverage the existing `Stringer` interface, in case we ever need to turn a struct into a string for reasons other than "unique cache key".

_Anyway._

Our generic cache is built on top of [go-cache](https://pkg.go.dev/github.com/patrickmn/go-cache@v2.1.0+incompatible) and provides only a subset of the functionality.
We can expand this set if we end up needing more features, but for now `Add()` and `Get()` are sufficient.

### The Backfill Queue

_Or: "The Reason We Aren't Using `vitess.io/vitess/go/cache`"_.

The schema endpoints in vtadmin have a bunch of different options which control the size and shape of the data returned to the caller.
We do cross-shard size aggregations, can request subsets of tables in a keyspace (including fetching just a single table's `TableDefinition`), include or exclude views, and so on.

We could build a cache that incorporates all of these options into the key, and cache each permutation of possible request/response pairs for every keyspace in a cluster, but:

1. We would be storing a ton of duplicate data.
2. We would get really poor cache reuse.

    a. Imagine requesting `/schemas/ks1/` and then `/schemas/ks1?tables=t1,t2` &mdash; we would have to make two full round-trips from `vtadmin-api` to the cluster to first fetch all the table definitions and then again to get `t1` and `t2` _even though_ we literally just fetched them.

So, what we do instead is cache only the largest response payload (in other words, the "give me all of everything" request equivalent) in the cache.
Then, when pulling data out of the cache, we look at the particular request's filtering options to whittle down the full payload into the subset that the caller actually wants.

However, this means that if the first request (i.e. "nothing cached") is not for the full payload, we won't be able to fill the cache before returning.
So instead, we make one blocking round-trip to the cluster to get the payload for that request, and then in the background instruct the cache to go get the full payload, via `EnqueueBackfill()`.

Then, future requests will have the fully-cached payload, which they can extract subsets of depending on their specific options.

### Cache Busting

Finally, we need a way to force a cache eviction/bypass/refresh/_whatever_, so we aren't stuck serving stale data waiting for responses to get evicted organically.

I wanted to solve this globally so that we would not need to update every single request proto message when we got around to adding caching for it.

So!
`vtadmin-api` now takes a global flag called `--cache-refresh-key` (despite using it as the title of this section, I dislike the term "cache busting" and this was the best alternative I could come up with &mdash; suggestions welcome!).

This gets incorporated as an HTTP header called `X-<the-flag-value>` and a gRPC metadata key, each of which is used to propagate "please bypass and refresh the cache for this request" from the caller down to a particular cluster method.

This is covered by the functions `ShouldRefreshFromIncomingContext` (gRPC metadata case) and `ShouldRefreshFromRequest` (HTTP case), as well as in our HTTP-to-gRPC adapter layer, which propagates from an HTTP header to gRPC metadata.

### Outstanding TODOs

The main thing missing from schema caching in particular is `FindSchemas`.
Because we need not only to know the schema definition(s) but which cluster(s) those definitions live in, we cannot cache solely at the cluster level.

We'll need to add a second cache at the API level to track which clusters have which schemas, and then look to just those clusters for cached schemas.
This PR is already a pretty significant chunk of work, so I am punting on that for now.

## Related Issue(s)

- Closes #10109 
- https://github.com/vitessio/vitess/projects/12#card-81082788

## Checklist


ajm188:
I _think_ I've fixed the race-y tests (which kinda blew up the diff a bit more 😅), so this is ready for review! cc @vmg @doeg @notfelineit

vmg:
I'll be :eye: this today. Lots of code!

go/vt/vtadmin/cache/cache.go
```diff
@@ -0,0 +1,262 @@
...
+		var req *backfillRequest[Key]
+		select {
+		case <-c.ctx.Done():
+			return
+		case req = <-c.backfills:
+			select {
+			case <-c.ctx.Done():
+				return
+			default:
+			}
```
vmg, line 213: This is not doing anything really -- the `ctx.Done()` could be closed throughout any of the lines of this function, so the whole function should be able to operate with a done context. Checking it just one extra time at the beginning is no different than the check in the parent `select` loop.

go/vt/vtadmin/cache/cache.go
```diff
@@ -0,0 +1,262 @@
...
+//		var c *cache.Cache[BackfillSchemaRequest, *vtadminpb.Schema]
+//		c := cache.New(func(ctx context.Context, req BackfillSchemaRequest) (*vtadminpb.Schema, error) {
+//			// Fetch schema based on fields in `req`.
+//			// If err is nil, the backfilled schema will be added to the cache.
+//			return cluster.fetchSchema(ctx, req)
+//		})
+//
+type Cache[Key Keyer, Value any] struct {
+	cache     *cache.Cache
+	fillcache *cache.Cache
```
vmg, line 102: I don't fully understand the behavior of `fillcache`. It seems to be keeping track of the last time we enqueued a fill, so it doesn't happen too often, but if that's the only purpose, it doesn't need to be a cache. It can simply be a `map[Key]time.Time`.

go/vt/vtadmin/cache/cache.go
```diff
@@ -0,0 +1,262 @@
...
+
+// EnqueueBackfill submits a request to the backfill queue.
+func (c *Cache[Key, Value]) EnqueueBackfill(k Key) bool {
+	req := &backfillRequest[Key]{
+		k:           k,
+		requestedAt: time.Now().UTC(),
+	}
+
+	ctx, cancel := context.WithTimeout(c.ctx, c.cfg.BackfillEnqueueWaitTime)
+	defer cancel()
```
vmg, line 177: Since you're not passing the context anywhere, this is better modeled as a `time.After`.

go/vt/vtadmin/cache/cache.go
vmg: (just make sure to `delete` from the map after each read)

go/vt/vtadmin/cache/cache.go
```diff
@@ -0,0 +1,262 @@
...
+		"config":                c.cfg,
+		"backfill_queue_length": len(c.backfills),
+		"closed":                false,
+	}
+
+	select {
+	case <-c.ctx.Done():
+		m["closed"] = true
+	default:
+	}
```
vmg, line 259: This is equivalent to `c.ctx.Err() != nil`.

vmg (changes requested):
Sorry for the delay on this! I took a look this morning. I don't fully understand the double-cache usage, so I left a suggestion. Also some changes to racy stuff with contexts. Cheers!

go/vt/vtadmin/cache/cache.go
ajm188: yep, you have the intended purpose exactly right! i was using the cache because it's threadsafe, if i take your suggestion i'll need to manually manage a mutex for it (which is fine, i think in hindsight i was just being a little lazy and "well, i've already got caches on the brain")

ajm188: > (just make sure to `delete` from the map after each read)

I actually don't think so, but double check my thinking:

Imagine your backfill queue was `[k1, k1, k1]` and your wait interval was 10 minutes.

On the first loop, you fill `k1`, and record `time.Now()`.
On the second loop, you try to fill `k1` again, but it's been less than 10 minutes, so you drop the request.
If you delete the key from `fillcache`, the third loop would then re-fill `k1` even though it hasn't been 10 minutes.

So I think the map has to grow forever. The only reason to delete would be if, when you read, the fill time is outside the wait period, but then you're going to re-fill the cache which will just put the key right back in the map. The only "space" you would save is if your re-fills are failing (i.e. `fillFunc` returns `err != nil`, which _should_ happen (ideally) rarely.

go/vt/vtadmin/cache/cache.go
ajm188: Ah true, good catch!! Ideally, the given `fillFunc` should fail on a done context as well, so we should get an error back from there and then the outer `select` will pick up the done-ness on the next iteration.

go/vt/vtadmin/cache/cache.go
vmg: Huh, wouldn't that also apply to the cache though?

ajm188: yeah in most practical cases, probably (but we also don't explicitly `delete` from the cache currently, we just expire things)

ajm188: well, okay, _we_ never explicitly delete things, but if you specify a non-zero `CleanupInterval`, then `go-cache` runs a background thread that deletes things that are expired, so infrequently-fetched schemas will get purged over time

go/vt/vtadmin/cache/cache.go
```diff
@@ -0,0 +1,255 @@
...
+}
+
+func (c *Cache[Key, Value]) add(key string, val Value, d time.Duration) error {
+	c.m.Lock()
+	defer c.m.Unlock()
+
+	// Record the time we last cached this key, to check against
+	c.lastFill[key] = time.Now().UTC()
+	// Then cache the actual value.
+	return c.cache.Add(key, val, d)
```
vmg, line 158: I would keep `c.cache.Add` outside of the locking for `c.m`. It adds unnecessary contention.

go/vt/vtadmin/cache/cache.go
```diff
@@ -0,0 +1,255 @@
...
+		if req.requestedAt.Add(c.cfg.BackfillRequestTTL).Before(time.Now()) {
+			// We took too long to get to this request, per config options.
+			log.Warningf("backfill for %s requested at %s; discarding due to exceeding TTL (%s)", req.k.Key(), req.requestedAt, c.cfg.BackfillRequestTTL)
+			continue
+		}
+
+		key := req.k.Key()
+
+		c.m.Lock()
+		if t, ok := c.lastFill[key]; ok {
```
vmg, line 222: I see you've decided not to clean-up `lastFill`, which probably makes sense for most growth patterns. I would comment this behavior explicitly.

vmg (approved):
Looks good! You can reduce the locking in `Get` though.

PR #10122, opened 2022-04-21
Fix StreamExecute in Gen4CompareV3

frouioui:
## Description

This PR fixes the `TryStreamExecute` of the `Gen4CompareV3` planner. @systay found that we need to use a mutex when appending new rows from the children's engine Primitive.

## Checklist


systay (approved):

frouioui:
Adding `Do Not Merge` to this PR. It seems that `TestUnionDistinct` is also very flaky. Trying to fix this as part of this PR.

PR #10123, opened 2022-04-21
Backport: Only start SQL thread temporarily to WaitForPosition if needed

mattlord:
**:warning: This does not require/warrant another 13.0 patch release but rather should be included in the next patch release whenever it's made :warning:** 

## Description

After https://github.com/vitessio/vitess/pull/9512 we always attempted to start the replication `SQL_Thread`(s) when waiting for a given position. The problem with this, however, is that [if the `SQL_Thread` is running but the `IO_Thread` is not then the tablet repair does not try and start replication on a replica tablet](https://github.com/planetscale/vitess/blob/ef7363e918e5c7093d72f15471f4c0d408ac0d10/go/vt/vttablet/tabletmanager/replmanager.go#L106-L117). So in certain states such as when initializing a shard, replication may end up in a non-healthy state and never be repaired.

1. We only attempt to start the `SQL_Thread`(s) if it's not already running
2. If we explicitly start the `SQL_Thread(s)` then we also explicitly reset it to what it was (stopped) as we exit the call

Because the caller should be/have a TabletManager which has a mutex, this should ensure that the replication manager calls are serialized and because we are resetting the replication state after mutating it, everything should work as it did before https://github.com/vitessio/vitess/pull/9512 with the exception being that when waiting we ensure that the replica at least has the possibility of catching up.

## Related Issue(s)
  - Follow-up to: https://github.com/vitessio/vitess/pull/9512
  - Backport-of: https://github.com/vitessio/vitess/pull/10104

## Checklist

### DIFF
--- a/go/vt/mysqlctl/replication.go
+++ b/go/vt/mysqlctl/replication.go
@@ -27,6 +27,7 @@ import (
 	"strings"
 	"time"
 
+	"vitess.io/vitess/go/vt/vterrors"
 	"vitess.io/vitess/go/vt/vtgate/evalengine"
 
 	"context"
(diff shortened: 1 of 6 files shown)


go/vt/mysqlctl/replication.go
```diff
@@ -244,14 +269,32 @@ func (mysqld *Mysqld) WaitSourcePos(ctx context.Context, targetPos mysql.Positio
...
 
+		// If the SQL thread(s) is not already running -- e.g. in the case of EmergencyReparentShard where the
+		// instance is transitioning to PRIMARY (elect) and we can no longer talk to the old PRIMARY, we need
+		// it to catch up as much as possible by executing all of the locally queued binary log events (in
+		// the existing relay logs) before it starts serving traffic for the shard to minimize potential data
+		// loss -- then we try to start the SQL Thread(s) before waiting for the position to be reached at
+		// which point the SQL thread(s) will be stopped again, since the replicas can only make forward
+		// progress if the SQL thread is started and we have already verified that the replica is not already
+		// as advanced as we want it to be
+		if !replicationStatus.SQLHealthy() {
```
GuptaManan100, line 285: We'll have to unroll the `SQLHealthy` because it isn't defined in release-13

GuptaManan100 (approved):

PR #10127, opened 2022-04-22
Add parsing support for Partition Definitions

K-Kumar-01:
## Description
Complete the parsing support for partition definitions.
Now supports the following in partition definition:
- `[[STORAGE] ENGINE [=] engine_name ]`
- `[COMMENT [=] 'string' ]`
- `[[STORAGE] ENGINE [=] engine_name]`
- `[DATA DIRECTORY [=] 'data_dir']`
- `[INDEX DIRECTORY [=] 'index_dir']`
- `[MAX_ROWS [=] max_number_of_rows]`
- `[MIN_ROWS [=] min_number_of_rows]`
- `[TABLESPACE [=] tablespace_name]`

Complete information available under `partition_definition` code section (present in the create table code section) in [this](https://dev.mysql.com/doc/refman/8.0/en/create-table.html) page.

## Related Issue(s)
#8604 

## Checklist


go/vt/sqlparser/ast.go
```diff
@@ -1626,6 +1637,42 @@ type PartitionValueRange struct {
...
 }
 
+type PartitionEngine struct {
+	Storage bool
+	Equal   string
+	Name    string
+}
+
+type PartitionComment struct {
+	Equal   string
```
GuptaManan100, line 1647: You can drop all the `Equal` fields. Having an `=` in the query where it is optional doesn't change the meaning of the query and is therefore of no use to store in the struct. You can look at the other places where we have used `equal_opt` in sql.y

go/vt/sqlparser/ast.go
```diff
@@ -1626,6 +1637,42 @@ type PartitionValueRange struct {
...
+
+type PartitionMinRows struct {
+	Equal string
+	Rows  int
+}
+
+type PartitionTableSpace struct {
+	Equal string
+	Name  string
+}
```
GuptaManan100, line 1674: After you remove the Equal field, the structs can be unrolled into the main struct directly

GuptaManan100 (commented):
Apart from these changes, everything else looks good to me.

go/vt/sqlparser/ast.go
GuptaManan100: Fixed

go/vt/sqlparser/ast.go
```diff
@@ -1613,8 +1613,19 @@ type PartitionSpecAction int8
...
+	Options *PartitionDefinitionOptions
+}
+
+type PartitionDefinitionOptions struct {
+	ValueRange     *PartitionValueRange
+	Comment        string
+	Engine         *PartitionEngine
+	DataDirectory  string
+	IndexDirectory string
+	MaxRows        int
```
GuptaManan100, line 1626: @systay @frouioui @harshit-gangal `MaxRows`, `MinRows`, `DataDirectory` etc are fields that are optional in the Partition definition, so just storing the int/string might be insufficient. i.e., not specifying MaxRows is different from MaxRows being 0.
@K-Kumar-01 opted to set the default value to -1 in the parser, but I think that's not a great idea, in the sense that, someone using this struct later, would need to know that -1 means that no value is specified, otherwise the user might think (rightly so) that it is a positive value, leading to errors down the line.

We have two alternatives and wanted to know what is the best of the two that you guys think-
1. Store them as *string/ *int instead, so the default value is nil
2. Introduce a struct for them (one for each, just having a single value) and store a pointer to that struct.

WDYT?

GuptaManan100 (changes requested):

go/vt/sqlparser/ast.go
GuptaManan100: That decides it then, @K-Kumar-01, could you make the necessary changes, and then we can merge it

GuptaManan100 (approved):

shlomi-noach (approved):
:man_cook:

PR #10129, opened 2022-04-22
[VTAdmin] Replace changed clusters when using Dynamic Discovery

notfelineit:
## Description
This PR updates dynamic cluster behavior so that an existing cluster can be updated if its values change. Prior behavior would skip adding the new cluster, making it impossible to update a cluster once it's in the cluster map.

## Related Issue(s)
N/A

## Checklist

-   [X] "Backport me!" label has been added if this change should be backported **No**
-   [X] Tests were added or are not required **Test added**

## Deployment Notes
N/A


notfelineit:
@ajm188 no, hence the prior (unsuccessful) remove ghost file commit

go/vt/vtadmin/api.go
```diff
@@ -271,16 +271,14 @@ func (api *API) WithCluster(c *cluster.Cluster, id string) dynamic.API {
 	}
 
 	if c != nil {
-		if _, exists := api.clusterMap[id]; !exists {
+		if existingCluster, exists := api.clusterMap[id]; !exists || existingCluster != c {
```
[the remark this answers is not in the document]
notfelineit: Good point, probably Equal just in case!

go/vt/vtadmin/api.go
```diff
@@ -271,7 +272,7 @@ func (api *API) WithCluster(c *cluster.Cluster, id string) dynamic.API {
 	}
 
 	if c != nil {
-		if existingCluster, exists := api.clusterMap[id]; !exists || existingCluster != c {
+		if existingCluster, exists := api.clusterMap[id]; !exists || !reflect.DeepEqual(existingCluster, c) {
```
ajm188, line 275: ah, so to clarify: i don't think we should use reflection, i think we should either be okay with pointer equality OR define our own Equal method and decide on what _we_ mean for two clusters to be equivalent

go/vt/vtadmin/cluster/cluster.go
```diff
@@ -1682,3 +1683,42 @@ func (c *Cluster) Debug() map[string]any {
...
+	}
+
+	for _, addr := range list1 {
+		if !slices.Contains(list2, addr) {
+			return false
+		}
+	}
+
+	return true
+}
```
ajm188, line 1724: this (guessing) is going to do a full scan for each `slices.Contains` call, but we have `[]string` slices, so:

```go
sort.Strings(list1)
sort.Strings(list2)
for i, e1 := range list1 {
    e2 := list2[i]
    if e1 != e2 {
        return false
    }
}
```

go.mod
```diff
@@ -120,6 +120,8 @@ require (
 	sigs.k8s.io/yaml v1.2.0
 )
 
+require k8s.io/utils v0.0.0-20220210201930-3a6ce19ff2f9
```
ajm188, line 123: I think you need a `go mod tidy`, otherwise I'm confused why this change caused this 🤔

notfelineit: I thought I ran it 🤔 going to do it again

ajm188 (approved):

PR #10131, opened 2022-04-24
schemadiff: support for RangeRotationStrategy diff hint

shlomi-noach:
## Description

Diffing partitions is a complex topic. There can be scenarios where a change of partitioning scheme required an entire rebuild of the table (e.g. change of number of `HASH` partitions), and cases where optimally one could get by with only modifying some of the partitions (e.g. `COALESCE` two partitions into one).

Normally, when we see a diff of partitions, we just invoke an `ALTER TABLE... PARTITION BY...` with the entire new partitioning scheme. The idea? Even a `COALESCE` for two partitions out of `200` is a blocking schema change. As we push forward Online DDL, we are opinionated that it's better to rebuild the entire table, online, than just `2` partitions out of `200`, in a blocking operation.

However, there are exceptions to this.

A common partitioning scheme that does not require blocking: rotation of [range partitioning](https://dev.mysql.com/doc/refman/8.0/en/partitioning-range.html). The classic example is, say, partitioning by day, such that every day we add a new partition; and possibly every day we also purge an old partition.
We could be adding a bunch of partitions at a time, or drop a bunch of partitions at a time.
Adding new partitions is super cheap, and dropping old partitions can be cheap or expensive, depending on the size of dropped partition and version of MySQL.

In this PR `schemadiff` identifies a rotating range partitioning scheme: two tables that are partitioned by `RANGE`, share exactly one consecutive non-empty sequence of partitions (i.e., the two tables have _some partitions in common_), and possibly:
- 1st table (aka "old" table) does not have some newer partitions (these have just been rotated in), and
- 2nd table (aka "new" table) does not have some older partitions (they have been rotated away)

As of this PR `schemadiff` supports a `RangeRotationStrategy`, of the following values:

- `RangeRotationFullSpec`: generate a full partitioning spec, meaning do not optimize for partition rotation; this behavior is the same as before this PR
- `RangeRotationIgnore`: consider this as a non-diff. This is similar in concept to ignoring an `AUTO_INCREMENT` increase.
- `RangeRotationStatements` **NOT YET IMPLEMENTED**. Will throw an error. Once implemented, will return the proper `DROP PARTITION ...` and `ADD PARTITION ...` statements that perform the rotation.

## Checklist

### DIFF
--- a/go/vt/schemadiff/table.go
+++ b/go/vt/schemadiff/table.go
@@ -370,6 +370,59 @@ func (c *CreateTableEntity) diffOptions(alterTable *sqlparser.AlterTable,
 	return nil
 }
 
+// rangePartitionsAddedRemoved returns true when:
+// - both table partitions are RANGE type
+// - there is exactly one consequitive non-empty shared sequence of partitions (same names, same range values, in same order)
+// - table1 may have non-empty list of partitions _preceding_ this sequence, and table2 may not
+// - table2 may have non-empty list of partitions _following_ this sequence, and table1 may not
+func (c *CreateTableEntity) isRangePartitionsRotation(
+	t1Partitions *sqlparser.PartitionOption,
+	t2Partitions *sqlparser.PartitionOption,
+) bool {
+	// Validate that both tables have range partitioning
+	if t1Partitions.Type != t2Partitions.Type {
+		return false
+	}
+	if t1Partitions.Type != sqlparser.RangeType {
+		return false
+	}
+	definitions1 := t1Partitions.Definitions
+	definitions2 := t2Partitions.Definitions
+	// there has to be a non-empty shared list, therefore both definitions must be non-empty:
+	if len(definitions1) == 0 {
+		return false
+	}
+	if len(definitions2) == 0 {
+		return false
+	}
+	// It's OK for prefix of t1 partitions to be nonexistent in t2 (as they may have been rotated away in t2)
+	for len(definitions1) > 0 && sqlparser.String(definitions1[0]) != sqlparser.String(definitions2[0]) {
+		definitions1 = definitions1[1:]
+	}
+	if len(definitions1) == 0 {
+		// We've exhaused definition1 trying to find a shared partition with definitions2. Nothing found.
+		// so there is no shared sequence between the two tables.
+		return false
+	}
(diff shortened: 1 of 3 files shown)


shlomi-noach:
To support `RangeRotationStatements` we'd first need an `AddPartition` and `DropPartition` statements in AST, that implement `AlterOption` via `iAlterOption()`.

GuptaManan100 (approved):
It looks good to me! What might be worth noting is that, we are expanding the parsing support for partition definitions (#10127 ) along with sub partitions (soon to follow).

shlomi-noach:
> What might be worth noting is that, we are expanding the parsing support for partition definitions (https://github.com/vitessio/vitess/pull/10127 ) along with sub partitions

Oh wonderful! Do you think we might be able to then have:

> `AddPartition` and `DropPartition` statements in AST, that implement `AlterOption` via `iAlterOption()`.

?

GuptaManan100:
> Oh wonderful! Do you think we might be able to then have:
> 
> > `AddPartition` and `DropPartition` statements in AST, that implement `AlterOption` via `iAlterOption()`.
> 
> ?

Yes! Most definitely. We can work on that before subpartitions, since I see you need it for completing this work!

shlomi-noach:
@GuptaManan100 thank you! There is no urgency for this work.

GuptaManan100:
@shlomi-noach I checked, and apparently we already have parsing support for `ADD PARTITION`, `DROP PARTITION`, etc already. It was added way back in https://github.com/vitessio/vitess/pull/7199. These don't implement the AlterOption interface and instead are specified in `PartitionSpec` in the `AlterTable` field. I don't remember exactly why I implemented it this way but my guess would be, we did this because there can be multiple alter options, but only 1 partition alter in a statement. This is written at the end of the doc here - https://dev.mysql.com/doc/refman/8.0/en/alter-table.html

GuptaManan100:
With [this ](https://github.com/vitessio/vitess/pull/10127) we augmented the partition definition, which also added that additional parsing to `ADD PARTITION` so we should be good there too

shlomi-noach:
@GuptaManan100 ahhh, that's right! Now I see it. Thank you.

shlomi-noach:

PR #5948, opened 2020-03-21
messaging: simplified -> Breaking Change

sougou:
Closes #5947 

Details are also covered in the issue.

Issue #5947, opened 2020-03-21
RFC: Messager simplification -> Breaking change


sougou:
After the move to RBR for messaging, and discussions with @derekperkins, we're simplifying the messaging API. On the upside, this will support all insert constructs. On the downside, the application has more responsibilities on correctness and features. What is changing:
* time_created and time_scheduled are removed from the list of mandatory columns. They will neither be checked nor be populated. If necessary, the application can add these columns and keep them populated.
* The primary key that used to be (time_scheduled,id) is now just id.
* The unique key on id is not required any more, since the PK covers that.
* time_next remains a nullable column, but it's recommended to default it to 0. Setting it to 0 (or any value < now) will schedule the message for immediate delivery. Previously a 0-value was treated as an acked message. This is not the case any more. This means that one could insert a row without specifying a time_next. The code checks for time_acked to contain a non-null (and non-zero) value to determine that a message was acked.
* The default value for epoch can also be 0, but the code treats null values as 0 also.
* Purging is by time_acked. Consequently, a new index is needed on time_acked.

Here's an idiomatic message table definition:
```
var createMessage = `create table vitess_message(
	id bigint,
	time_next bigint default 0,
	epoch bigint,
	time_acked bigint,
	message varchar(128),
	primary key(id),
	index next_idx(time_next, epoch),
	index ack_idx(time_acked))
comment 'vitess_message,vt_ack_wait=30,vt_purge_after=120,vt_batch_size=1,vt_cache_size=1000,vt_poller_interval=30'
```

Here's an idiomatic insert:
```
insert into vitess_message(id, message) values(1, 'hello world')
```

derekperkins:
For reference when we come back to document this, `time_acked`, and probably `time_next`, still aren't fully treating 0 as null. Setting `time_acked` to 0 results in immediate purging - starting with v6 and still true today.

https://github.com/vitessio/vitess/blob/main/go/vt/vttablet/tabletserver/messager/message_manager.go#L251-L258

derekperkins:
Verified that `time_next` has to be null, otherwise it gets picked up and continually resent. There must be some 0 handling because epoch doesn't get incremented.

PR #10132, opened 2022-04-24
Avoid deadlocks related to 0 receiver behavior

mattlord:
## Description

There's conditional behavior and code paths in the message manager when the number of receivers is 0. The main mutex and the stream mutex locks are used to ensure overall consistency. But:
1. The order that those locks are taken in was not enforced or consistent
2. There was a gap in the poller between when the 0 receiver code path was checked — specifically that it was NOT 0 — and the main mutex lock was taken to ensure that this invariant holds during the polling run here: https://github.com/vitessio/vitess/blob/c9853797bda8752202583695133c41a25d390930/go/vt/vttablet/tabletserver/messager/message_manager.go#L752-L787

### These two factors can lead to deadlocks when the receiver count goes to/from 0 while the the poller runs:

In [`unsubscribe()` we have the main mutex](https://github.com/vitessio/vitess/blob/c9853797bda8752202583695133c41a25d390930/go/vt/vttablet/tabletserver/messager/message_manager.go#L421-L423), and while holding it we then [call `stopVStream()` if the subscriber/receiver count is at 0](https://github.com/vitessio/vitess/blob/c9853797bda8752202583695133c41a25d390930/go/vt/vttablet/tabletserver/messager/message_manager.go#L437-L439). Then `stopVStream()` [tries to get the stream mutex](https://github.com/vitessio/vitess/blob/c9853797bda8752202583695133c41a25d390930/go/vt/vttablet/tabletserver/messager/message_manager.go#L629-L631).

Concurrently, `runPoller()` — which by default, or more accurately when using [the documented message table structure here](https://vitess.io/docs/13.0/reference/features/messaging/#creating-a-message-table) via the `vt_poller_interval` comment directive, runs on a 30 second ticker — [gets the stream mutex first](https://github.com/vitessio/vitess/blob/c9853797bda8752202583695133c41a25d390930/go/vt/vttablet/tabletserver/messager/message_manager.go#L751-L758), then it [later tries to get the main mutex](https://github.com/vitessio/vitess/blob/c9853797bda8752202583695133c41a25d390930/go/vt/vttablet/tabletserver/messager/message_manager.go#L776-L778).

Now we are deadlocked. They're both waiting on each other.

This then in turn causes things like `PlannedReparentShard` to hang because [during the transition from primary serving we stop/close the message service](https://github.com/vitessio/vitess/blob/c9853797bda8752202583695133c41a25d390930/go/vt/vttablet/tabletserver/state_manager.go#L450), which in turn [will wait trying to get the main mutex](https://github.com/vitessio/vitess/blob/c9853797bda8752202583695133c41a25d390930/go/vt/vttablet/tabletserver/messager/message_manager.go#L346-L354).

### Changes

1. **We clean up the mutex usage**.  In order to ensure that the cache stays in sync with the database and we are moving forward linearly though the logical GTID stream — updated by the poller and its results streamer, and the binlog streamer — we now use the `cacheManagementMu` lock. The explicit usage of this new mutex allows us to avoid the deadlocks seen before because it's taken immediately when the poller or binlog streamer process data coming from the database (which is used to update the cache), and we no longer need to take this lock when starting or stopping the binlog streamer itself.
    - After the initial changes made in the PR I could see that we were still prone to deadlocks in various places due to checking the receiver count all the time (I could see this in the race test stack traces, see Notes below), which required the main mutex, while holding the stream mutex. Anywhere we were holding the stream mutex, and checking receiver count — which we did A LOT, for example when processing every row event from the binlog stream via [`processRowEvent()`](https://github.com/vitessio/vitess/blob/341d80a5f96b65ce9a16211ebd4d2034f859ab78/go/vt/vttablet/tabletserver/messager/message_manager.go#L707)->[`Add()`](https://github.com/vitessio/vitess/blob/341d80a5f96b65ce9a16211ebd4d2034f859ab78/go/vt/vttablet/tabletserver/messager/message_manager.go#L469-L474) — we were at risk of a deadlock. The main mutex became VERY hot because we were checking the receiver count all the time to ensure the fast path (do nothing when there's no receivers).
2. **We add an important predicate to the poller query** — `time_acked is null` — to try and limit the size of scans done on the underlying message table to try and ensure that the poller runs quickly. This aspect is important because the binlog event processing is blocked during this time. It also prevents us from doing unnecessary work as InnoDB does most of the work — all query (WHERE) predicates become index condition pushdowns and the new `poller_idx` index can be used for the ordering to apply the `LIMIT` as well — and passes the minimum amount of potentially useful data up the stack InnoDB->MySQL->MessageManager->Cache.
3. **We update the tests so that we're using the new recommended/documented (see docs ticket below) message table structure and thus we're testing behavior with the structure we tell users to use.**

## Notes

This is able to complete on `vitessio/main`:
```
$ make build; failed=0; for i in {1..100}; do if ! go test -race -v -failfast -timeout 1m vitess.io/vitess/go/vt/vttablet/tabletserver/messager; then echo -e '\nUh oh! We have a problem...\n'; failed=1; break; fi; go clean -testcache; done; if [[ ${failed} -eq 0 ]]; then echo -e '\nWe made it! All good...\n'; fi
```

Whatever changes are made here we should continue to pass that test (it does). It's even more reliable now — you can run it 1,000 times and no flakiness.

## Related Issue(s)
  - Fixes: https://github.com/vitessio/vitess/issues/9936

## Checklist

-   [x] Tests are not required (the existing unit and e2e tests cover the behavior)
-   [x] Documentation: https://github.com/vitessio/website/pull/1015

### DIFF
--- a/go/vt/vttablet/tabletserver/messager/message_manager.go
+++ b/go/vt/vttablet/tabletserver/messager/message_manager.go
@@ -469,9 +469,7 @@ func (mm *messageManager) rescanReceivers(start int) {
 // if successful. If the message is already present,
 // it still returns true.
 func (mm *messageManager) Add(mr *MessageRow) bool {
-	mm.mu.Lock()
-	defer mm.mu.Unlock()
-	if len(mm.receivers) == 0 {
+	if mm.receiverCount() == 0 {
 		return false
 	}
 	// If cache is empty, we have to broadcast that we're not empty
@@ -879,7 +877,7 @@ func (mm *messageManager) GeneratePurgeQuery(timeCutoff int64) (string, map[stri
 	}
 }
 
-// BuildMessageRow builds a MessageRow for a db row.
+// BuildMessageRow builds a MessageRow from a db row.
 func BuildMessageRow(row []sqltypes.Value) (*MessageRow, error) {
 	mr := &MessageRow{Row: row[4:]}
 	if !row[0].IsNull() {
--- a/go/vt/vttablet/tabletserver/messager/message_manager.go
+++ b/go/vt/vttablet/tabletserver/messager/message_manager.go
@@ -189,7 +189,7 @@ type messageManager struct {
 	curReceiver     int
 	messagesPending bool
 
-	// lastPostionMu protects the lastPollPosition variable which is the main
+	// lastPollPostionMu protects the lastPollPosition variable which is the main
 	// point of coordination between the poller and the binlog streamer to
 	// ensure that we are not re-processing older events.
 	lastPollPositionMu sync.Mutex
@@ -754,6 +754,8 @@ func (mm *messageManager) processRowEvent(fields []*querypb.Field, rowEvent *bin
 }
 
 func (mm *messageManager) runPoller() {
+	mm.streamProcessingMu.Lock()
+	defer mm.streamProcessingMu.Unlock()
 	mm.mu.Lock()
 	defer mm.mu.Unlock()
 
@@ -773,9 +775,6 @@ func (mm *messageManager) runPoller() {
 		"time_next": sqltypes.Int64BindVariable(time.Now().UnixNano()),
 		"max":       sqltypes.Int64BindVariable(int64(size)),
 	}
-
-	mm.streamProcessingMu.Lock()
-	defer mm.streamProcessingMu.Unlock()
 
 	qr, err := mm.readPending(ctx, bindVars)
 	if err != nil {
--- a/go/vt/vttablet/tabletserver/messager/message_manager.go
+++ b/go/vt/vttablet/tabletserver/messager/message_manager.go
@@ -754,13 +754,13 @@ func (mm *messageManager) processRowEvent(fields []*querypb.Field, rowEvent *bin
 }
 
 func (mm *messageManager) runPoller() {
-	mm.getExclusiveLock()
-	defer mm.releaseExclusiveLock()
-
 	// Fast-path. Skip all the work.
 	if mm.receiverCount.Get() == 0 {
 		return
 	}
+
+	mm.getExclusiveLock()
+	defer mm.releaseExclusiveLock()
 
 	ctx, cancel := context.WithTimeout(tabletenv.LocalContext(), mm.pollerTicks.Interval())
 	defer func() {
@@ -773,6 +773,7 @@ func (mm *messageManager) runPoller() {
 		"time_next": sqltypes.Int64BindVariable(time.Now().UnixNano()),
 		"max":       sqltypes.Int64BindVariable(int64(size)),
 	}
+
 	qr, err := mm.readPending(ctx, bindVars)
 	if err != nil {
 		return
@@ -947,9 +948,8 @@ func (mm *messageManager) readPending(ctx context.Context, bindVars map[string]*
 }
 
 // This grants the caller exclusive access to the message service.
-// When this is needed for a function, you must get the mutexes in
-// main mutex, stream mutex order to avoid deadlocks with other places
-// that conditionally get the mutexes in this same order.
+// When this is needed for a function, you can use this to
+// enforce consistent locking order.
 func (mm *messageManager) getExclusiveLock() {
 	mm.mu.Lock()
 	mm.streamMu.Lock()
--- a/go/vt/vttablet/tabletserver/messager/message_manager.go
+++ b/go/vt/vttablet/tabletserver/messager/message_manager.go
@@ -479,7 +479,9 @@ func (mm *messageManager) rescanReceivers(start int) {
 // if successful. If the message is already present,
 // it still returns true.
 func (mm *messageManager) Add(mr *MessageRow) bool {
-	if mm.getReceiverCount() == 0 {
+	mm.mu.Lock()
+	defer mm.mu.Unlock()
+	if len(mm.receivers) == 0 {
 		return false
 	}
 	// If cache is empty, we have to broadcast that we're not empty
--- a/go/vt/vttablet/tabletserver/messager/message_manager.go
+++ b/go/vt/vttablet/tabletserver/messager/message_manager.go
@@ -183,16 +183,24 @@ type messageManager struct {
 	isOpen bool
 	// cond waits on curReceiver == -1 || cache.IsEmpty():
 	// No current receivers available or cache is empty.
-	cond      sync.Cond
-	cache     *cache
-	receivers []*receiverWithStatus
-	// Way to track the receiver count in a consistent way w/o locks
-	receiverCount   sync2.AtomicInt64
-	curReceiver     int64
-	messagesPending sync2.AtomicBool
+	cond            sync.Cond
+	cache           *cache
+	receivers       []*receiverWithStatus
+	curReceiver     int
+	messagesPending bool
 
-	// streamMu keeps the cache and database consistent with each other.
-	// Specifically:
+	// lastPostionMu protects the lastPollPosition variable which is the main
+	// point of coordination between the poller and the binlog streamer to
+	// ensure that we are not re-processing older events.
+	lastPollPositionMu sync.Mutex
+	lastPollPosition   *mysql.Position
+
+	// streamProcessingMu keeps the cache and database consistent with
+	// each other by ensuring that only one of the streams is processing
+	// changes at a time. The Poller uses a results streamer to pull directly
+	// from the message table and the message manager uses a binlog streamer
+	// to process change events. This mutex ensures that only one of them are
+	// updating the cache at any one time.
 	// It prevents items from being removed from cache while the poller
 	// reads from the db and adds items to it. Otherwise, the poller
 	// might add an older snapshot of a row that was just postponed.
@@ -201,12 +209,11 @@ type messageManager struct {
 	// lastPollPosition must be ignored by the vstream. It consequently
 	// also blocks vstream from updating the cache while the poller is
 	// active.
-	streamMu sync.Mutex
+	streamProcessingMu sync.Mutex
 	// streamCancel is set when a vstream is running, and is reset
-	// to nil after a cancel. This allows for startVStream and stopVstream
+	// to nil after a cancel. This allows for startVStream and stopVStream
 	// to be idempotent.
-	streamCancel     func()
-	lastPollPosition *mysql.Position
+	streamCancel func()
 
 	// wg is for ensuring all running goroutines have returned
 	// before we can close the manager. You need to Add before
@@ -241,7 +248,7 @@ func newMessageManager(tsv TabletService, vs VStreamer, table *schema.Table, pos
 		pollerTicks:     timer.NewTimer(table.MessageInfo.PollInterval),
 		purgeTicks:      timer.NewTimer(table.MessageInfo.PollInterval),
 		postponeSema:    postponeSema,
-		messagesPending: sync2.NewAtomicBool(true),
+		messagesPending: true,
 	}
 	mm.cond.L = &mm.mu
 
@@ -366,7 +373,6 @@ func (mm *messageManager) Close() {
 	for _, rcvr := range mm.receivers {
 		rcvr.receiver.cancel()
 	}
-	mm.receiverCount.Set(0)
 	mm.receivers = nil
 	MessageStats.Set([]string{mm.name.String(), "ClientCount"}, 0)
 	log.Infof("messageManager - clearing cache")
@@ -406,12 +412,11 @@ func (mm *messageManager) Subscribe(ctx context.Context, send func(*sqltypes.Res
 	withStatus := &receiverWithStatus{
 		receiver: receiver,
 	}
-	if mm.receiverCount.Get() == 0 {
+	if len(mm.receivers) == 0 {
 		mm.startVStream()
 	}
 	mm.receivers = append(mm.receivers, withStatus)
-	mm.receiverCount.Add(1)
-	MessageStats.Set([]string{mm.name.String(), "ClientCount"}, mm.receiverCount.Get())
+	MessageStats.Set([]string{mm.name.String(), "ClientCount"}, int64(len(mm.receivers)))
 	if mm.curReceiver == -1 {
 		mm.rescanReceivers(-1)
 	}
@@ -432,17 +437,16 @@ func (mm *messageManager) unsubscribe(receiver *messageReceiver) {
 			continue
 		}
 		// Delete the item at current position.
-		n := mm.receiverCount.Get()
+		n := len(mm.receivers)
 		copy(mm.receivers[i:n-1], mm.receivers[i+1:n])
 		mm.receivers = mm.receivers[0 : n-1]
-		mm.receiverCount.Add(-1)
-		MessageStats.Set([]string{mm.name.String(), "ClientCount"}, mm.receiverCount.Get())
+		MessageStats.Set([]string{mm.name.String(), "ClientCount"}, int64(len(mm.receivers)))
 		break
 	}
 	// curReceiver is obsolete. Recompute.
 	mm.rescanReceivers(-1)
 	// If there are no receivers. Shut down the cache.
-	if mm.receiverCount.Get() == 0 {
+	if len(mm.receivers) == 0 {
 		mm.stopVStream()
 		mm.cache.Clear()
 	}
@@ -454,10 +458,10 @@ func (mm *messageManager) unsubscribe(receiver *messageReceiver) {
 // was previously -1, it broadcasts. If none was found,
 // curReceiver is set to -1. If there's no starting point,
 // it must be specified as -1.
-func (mm *messageManager) rescanReceivers(start int64) {
+func (mm *messageManager) rescanReceivers(start int) {
 	cur := start
 	for range mm.receivers {
-		cur = (cur + 1) % mm.receiverCount.Get()
+		cur = (cur + 1) % len(mm.receivers)
 		if !mm.receivers[cur].busy {
 			if mm.curReceiver == -1 {
 				mm.cond.Broadcast()
@@ -474,7 +478,7 @@ func (mm *messageManager) rescanReceivers(start int64) {
 // if successful. If the message is already present,
 // it still returns true.
 func (mm *messageManager) Add(mr *MessageRow) bool {
-	if mm.receiverCount.Get() == 0 {
+	if mm.getReceiverCount() == 0 {
 		return false
 	}
 	// If cache is empty, we have to broadcast that we're not empty
@@ -484,7 +488,7 @@ func (mm *messageManager) Add(mr *MessageRow) bool {
 	}
 	if !mm.cache.Add(mr) {
 		// Cache is full. Enter "messagesPending" mode.
-		mm.messagesPending.Set(true)
+		mm.messagesPending = true
 		return false
 	}
 	return true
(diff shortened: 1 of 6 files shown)


go/vt/vttablet/tabletserver/messager/message_manager.go
```diff
@@ -949,3 +946,17 @@ func (mm *messageManager) readPending(ctx context.Context, bindVars map[string]*
...
 	return qr, err
 }
+
+// This grants the caller exclusive access to the message service.
+// When this is needed for a function, you must get the mutexes in
+// stream mutex, main mutex order to avoid deadlocks with other places
+// that conditionally get the mutexes in this same order.
+func (mm *messageManager) getExclusiveLock() {
+	mm.streamMu.Lock()
+	mm.mu.Lock()
```
rohit-nayak-ps, line 956: Don't we need to change the order here? `unsubscribe()` will acquire `mu` first and then `streamMu` if it needs to stop the vstream.

go/vt/vttablet/tabletserver/messager/message_manager.go
```diff
@@ -252,7 +252,7 @@ func newMessageManager(tsv TabletService, vs VStreamer, table *schema.Table, pos
 		}},
 	}
 	mm.readByPriorityAndTimeNext = sqlparser.BuildParsedQuery(
-		"select priority, time_next, epoch, time_acked, %s from %v where time_next < %a order by priority, time_next desc limit %a",
+		"select priority, time_next, epoch, time_acked, %s from %v where time_acked is null and time_next < %a order by priority, time_next desc limit %a",
```
derekperkins, line 255: I'm surprised that this showed a query plan difference, assuming that you have an index on `time_next`

### original

```json
{
  "query_block": {
    "select_id": 1,
    "cost_info": {
      "query_cost": "1.71"
    },
    "ordering_operation": {
      "using_filesort": true,
      "cost_info": {
        "sort_cost": "1.00"
      },
      "table": {
        "table_name": "searches__extractor__msgs",
        "access_type": "range",
        "possible_keys": [
          "time_next"
        ],
        "key": "time_next",
        "used_key_parts": [
          "time_next"
        ],
        "key_length": "9",
        "rows_examined_per_scan": 1,
        "rows_produced_per_join": 1,
        "filtered": "100.00",
        "index_condition": "(`keywords`.`searches__extractor__msgs`.`time_next` < <cache>((unix_timestamp(now()) * 1e9)))",
        "cost_info": {
          "read_cost": "0.61",
          "eval_cost": "0.10",
          "prefix_cost": "0.71",
          "data_read_per_join": "728"
        },
        "used_columns": [
          "id",
          "keyspace_id",
          "priority",
          "epoch",
          "time_next",
          "time_acked"
        ]
      }
    }
  }
}
```

### with time_acked is null
```json
{
  "query_block": {
    "select_id": 1,
    "cost_info": {
      "query_cost": "1.21"
    },
    "ordering_operation": {
      "using_filesort": true,
      "cost_info": {
        "sort_cost": "0.50"
      },
      "table": {
        "table_name": "searches__extractor__msgs",
        "access_type": "range",
        "possible_keys": [
          "ack_idx",
          "time_next"
        ],
        "key": "time_next",
        "used_key_parts": [
          "time_next"
        ],
        "key_length": "9",
        "rows_examined_per_scan": 1,
        "rows_produced_per_join": 0,
        "filtered": "50.00",
        "index_condition": "(`keywords`.`searches__extractor__msgs`.`time_next` < <cache>((unix_timestamp(now()) * 1e9)))",
        "cost_info": {
          "read_cost": "0.66",
          "eval_cost": "0.05",
          "prefix_cost": "0.71",
          "data_read_per_join": "363"
        },
        "used_columns": [
          "id",
          "keyspace_id",
          "priority",
          "epoch",
          "time_next",
          "time_acked"
        ],
        "attached_condition": "(`keywords`.`searches__extractor__msgs`.`time_acked` is null)"
      }
    }
  }
}
```

go/vt/vttablet/tabletserver/messager/message_manager.go
mattlord: Yes! Good catch :-) Corrected

go/vt/vttablet/tabletserver/messager/message_manager.go
mattlord: @derekperkins Yes, this would only serve to avoid sending rows up the stack that will only be ignored when we go to add it to the message cache w/o some table side changes. I added a comment about what index I think should be there. The key factors are:
1. I can't imaging why the poller would ever need to read already acked messages. I would assume that this would often filter out a large portion of the rows and at a minimum it prevents us from returning rows to the message manager that we know won't end up added to the cache in the end
2. An *new* index on `(time_acked, priority, time_next desc)` should allow us to:
    - Push the index conditions down to the storage engine ([index condition pushdown](https://dev.mysql.com/doc/refman/8.0/en/index-condition-pushdown-optimization.html)) — which you'll see identified in the standard `EXPLAIN` output with `Using index condition` and JSON output with `"index_condition:" ...` — so that the filtering can be done there w/o having to pass information back through the storage engine API to the mysql query execution layer
    -  Avoid having to do a filesort — which we will see in the standard `EXPLAIN` output when `Using filesort` is absent and in the JSON output when we see `"using_filesort": false` — as we can get the order  from the index records that match `time_acked` predicate **_and_** apply the limit in the storage engine as well

So the new suggested/documented message table structure would be (also changing the `time_next` default to -1 so that messages are not skipped/lost when the binlog streamer connection is restarted):
```
create table my_message(
  time_scheduled bigint,
  id bigint,
  time_next bigint DEFAULT -1,
  epoch bigint,
  time_created bigint,
  time_acked bigint,
  message varchar(128),
  priority tinyint NOT NULL DEFAULT '0',
  primary key(time_scheduled, id),
  unique index id_idx(id),
  index poller_idx (time_acked, priority asc, time_next desc)
) comment 'vitess_message,vt_ack_wait=30,vt_purge_after=86400,vt_batch_size=10,vt_cache_size=10000,vt_poller_interval=30'
```
    
This should allow us to do the minimum amount of work necessary up the stack: InnoDB->MySQL->MessageManager->heap/cache. For example:
```
mysql> select count(*) from mt where time_acked is null and time_next < UNIX_TIMESTAMP(NOW())*1e9;
+----------+
| count(*) |
+----------+
|      463 |
+----------+
1 row in set (0.00 sec)

mysql> explain format=JSON select priority, time_next, epoch, time_acked from mt where time_acked is null and time_next < UNIX_TIMESTAMP(NOW())*1e9 order by priority, time_next desc limit 10000\G
EXPLAIN: {
  "query_block": {
    "select_id": 1,
    "cost_info": {
      "query_cost": "57.87"
    },
    "ordering_operation": {
      "using_filesort": false,
      "table": {
        "table_name": "mt",
        "access_type": "ref",
        "possible_keys": [
          "poller_idx",
          "poller_idx2"
        ],
        "key": "poller_idx",
        "used_key_parts": [
          "time_acked"
        ],
        "key_length": "9",
        "ref": [
          "const"
        ],
        "rows_examined_per_scan": 463,
        "rows_produced_per_join": 154,
        "filtered": "33.33",
        "index_condition": "((`vt_commerce`.`mt`.`time_acked` is null) and (`vt_commerce`.`mt`.`time_next` < <cache>((unix_timestamp(now()) * 1e9))))",
        "cost_info": {
          "read_cost": "11.57",
          "eval_cost": "15.43",
          "prefix_cost": "57.88",
          "data_read_per_join": "85K"
        },
        "used_columns": [
          "time_scheduled",
          "id",
          "time_next",
          "epoch",
          "time_acked",
          "priority"
        ]
      }
    }
  }
}
```

Would you be interested in testing?

derekperkins: The new index doesn't seem to have changed anything

```json
{
  "query_block": {
    "select_id": 1,
    "cost_info": {
      "query_cost": "1.21"
    },
    "ordering_operation": {
      "using_filesort": true,
      "cost_info": {
        "sort_cost": "0.50"
      },
      "table": {
        "table_name": "searches__extractor__msgs",
        "access_type": "range",
        "possible_keys": [
          "ack_idx",
          "time_next",
          "time_acked__priority__time_next"
        ],
        "key": "time_next",
        "used_key_parts": [
          "time_next"
        ],
        "key_length": "9",
        "rows_examined_per_scan": 1,
        "rows_produced_per_join": 0,
        "filtered": "50.00",
        "index_condition": "(`keywords`.`searches__extractor__msgs`.`time_next` < <cache>((unix_timestamp(now()) * 1e9)))",
        "cost_info": {
          "read_cost": "0.66",
          "eval_cost": "0.05",
          "prefix_cost": "0.71",
          "data_read_per_join": "364"
        },
        "used_columns": [
          "id",
          "keyspace_id",
          "priority",
          "epoch",
          "time_next",
          "time_acked"
        ],
        "attached_condition": "(`keywords`.`searches__extractor__msgs`.`time_acked` is null)"
      }
    }
  }
}
```

mattlord: Hi @derekperkins 

Can you please include the query along with the explain plan? Can you also please share the index definitions? From your explain output here, I *think* you're using the new updated/proposed query but the new `time_acked__priority__time_next` index is not in DESC order for the time_next field.

derekperkins: Some other changes that haven't made it to the docs yet from v6, some of which change the table structure. https://github.com/vitessio/vitess/releases/tag/v6.0.20-20200429. I suggested some changes in your docs PR

> Messaging: First Class Citizen + API Improvements
> The messaging feature has had several breaking changes and feature additions
> 
> - Direct inserts are now supported: INSERT INTO msg_table SELECT col1, col2...
> - time_scheduled and time_created are no longer required nor automatically populated.
> - priority is a new required field. Messages with a lower priority will be processed first.
> - Message retention is now measured after time_acked instead of time_scheduled
> - New table options vt_min_backoff and vt_max_backoff set bounds on exponential backoff for message retries
> - 33% backoff jitter has been added for messages that are getting postponed to prevent thundering herds.
> 
> Users of the messages feature are encouraged to read through the full details in issue [#5947](https://github.com/vitessio/vitess/issues/5947) [#5948](https://github.com/vitessio/vitess/pull/5948) [#6114](https://github.com/vitessio/vitess/pull/6114)

derekperkins: MySQL always chooses the `time_next` index. In a very unscientific test, I ssh'd into the MySQL instance in question to remove all network and Vitess latency, and repeatedly ran queries to get a decent average. The table in question has:
- 9.5M rows
- 6.5M with `time_next=NULL`, `time_acked=12345`
- 3M "failed" with very high  but < 10 matched the query on any given run.

#### Test results:
- Non force index: < 5 ms
- Force index: ~50 ms

```sql
select priority, time_next, epoch, time_acked, id, keyspace_id 
from searches__extractor__msgs # optional force index (time_acked__priority__time_next)
where time_acked is null and time_next < UNIX_TIMESTAMP(NOW())*1e9 
order by priority, time_next desc limit 10000
```

#### Force Index query plan

```json
{
  "query_block": {
    "select_id": 1,
    "cost_info": {
      "query_cost": "268.10"
    },
    "ordering_operation": {
      "using_filesort": false,
      "table": {
        "table_name": "searches__request_dispatcher__msgs",
        "access_type": "ref",
        "possible_keys": [
          "time_acked__priority__time_next"
        ],
        "key": "time_acked__priority__time_next",
        "used_key_parts": [
          "time_acked"
        ],
        "key_length": "9",
        "ref": [
          "const"
        ],
        "rows_examined_per_scan": 766,
        "rows_produced_per_join": 255,
        "filtered": "33.33",
        "index_condition": "((`keywords`.`searches__request_dispatcher__msgs`.`time_acked` is null) and (`keywords`.`searches__request_dispatcher__msgs`.`time_next` < <cache>((unix_timestamp(now()) * 1e9))))",
        "cost_info": {
          "read_cost": "191.50",
          "eval_cost": "25.53",
          "prefix_cost": "268.10",
          "data_read_per_join": "181K"
        },
        "used_columns": [
          "id",
          "keyspace_id",
          "priority",
          "epoch",
          "time_next",
          "time_acked"
        ]
      }
    }
  }
}
```

derekperkins: Continuing my non-scientific benchmark, I changed the where clause to max bigint, so it would now match all 3M rows, to give a better idea for the polling query on a backed up queue. This time, it did choose the new index by itself, and I had to force the `time_next` index. The new index is a clear winner here. I'm happy to test other permutations as well.

#### Test results:
- `time_next`: 350-400 ms
- `time_acked__priority__time_next`: 20-30 ms

```sql
select priority, time_next, epoch, time_acked, id, keyspace_id 
from searches__extractor__msgs # optional force index (time_next)
where time_acked is null and time_next < 9223372036854775807 
order by priority, time_next desc limit 10000
```


#### time_next query plan

```json
{
  "query_block": {
    "select_id": 1,
    "cost_info": {
      "query_cost": "121413.86"
    },
    "ordering_operation": {
      "using_filesort": true,
      "table": {
        "table_name": "searches__extractor__msgs",
        "access_type": "range",
        "possible_keys": [
          "time_next"
        ],
        "key": "time_next",
        "used_key_parts": [
          "time_next"
        ],
        "key_length": "9",
        "rows_examined_per_scan": 269808,
        "rows_produced_per_join": 26980,
        "filtered": "10.00",
        "index_condition": "(`keywords`.`searches__extractor__msgs`.`time_next` < 9223372036854775807)",
        "cost_info": {
          "read_cost": "118715.78",
          "eval_cost": "2698.08",
          "prefix_cost": "121413.86",
          "data_read_per_join": "18M"
        },
        "used_columns": [
          "id",
          "keyspace_id",
          "priority",
          "epoch",
          "time_next",
          "time_acked"
        ],
        "attached_condition": "(`keywords`.`searches__extractor__msgs`.`time_acked` is null)"
      }
    }
  }
}
```

#### time_acked__priority__time_next query plan

```json
{
  "query_block": {
    "select_id": 1,
    "cost_info": {
      "query_cost": "34230.00"
    },
    "ordering_operation": {
      "using_filesort": false,
      "table": {
        "table_name": "searches__extractor__msgs",
        "access_type": "ref",
        "possible_keys": [
          "time_acked__priority__time_next"
        ],
        "key": "time_acked__priority__time_next",
        "used_key_parts": [
          "time_acked"
        ],
        "key_length": "9",
        "ref": [
          "const"
        ],
        "rows_examined_per_scan": 267105,
        "rows_produced_per_join": 89026,
        "filtered": "33.33",
        "index_condition": "((`keywords`.`searches__extractor__msgs`.`time_acked` is null) and (`keywords`.`searches__extractor__msgs`.`time_next` < 9223372036854775807))",
        "cost_info": {
          "read_cost": "7519.50",
          "eval_cost": "8902.61",
          "prefix_cost": "34230.00",
          "data_read_per_join": "61M"
        },
        "used_columns": [
          "id",
          "keyspace_id",
          "priority",
          "epoch",
          "time_next",
          "time_acked"
        ]
      }
    }
  }
}
```

derekperkins: I tried adding fields to your suggestion to make it a covering index, to see if that would change performance. It did speed it up 2-4x, but probably not worth it.

Test results:
- time_next: 350-400 ms
- time_acked__priority__time_next: 20-30 ms
- time_acked__priority__time_next__epoch__id__keyspace_id: 5-15 ms

```sql
alter table searches__extractor__msgs 
add index `time_acked__priority__time_next__epoch__id__keyspace_id` 
          (time_acked, priority, time_next desc, epoch, id, keyspace_id)
```

```json
{
  "query_block": {
    "select_id": 1,
    "cost_info": {
      "query_cost": "32684.49"
    },
    "ordering_operation": {
      "using_filesort": false,
      "table": {
        "table_name": "searches__extractor__msgs",
        "access_type": "ref",
        "possible_keys": [
          "ack_idx",
          "time_next",
          "time_acked__priority__time_next",
          "time_acked__priority__time_next__epoch__id__keyspace_id"
        ],
        "key": "time_acked__priority__time_next__epoch__id__keyspace_id",
        "used_key_parts": [
          "time_acked"
        ],
        "key_length": "9",
        "ref": [
          "const"
        ],
        "rows_examined_per_scan": 313005,
        "rows_produced_per_join": 156502,
        "filtered": "50.00",
        "using_index": true,
        "cost_info": {
          "read_cost": "1383.99",
          "eval_cost": "15650.25",
          "prefix_cost": "32684.49",
          "data_read_per_join": "108M"
        },
        "used_columns": [
          "id",
          "keyspace_id",
          "priority",
          "epoch",
          "time_next",
          "time_acked"
        ],
        "attached_condition": "((`keywords`.`searches__extractor__msgs`.`time_acked` is null) and (`keywords`.`searches__extractor__msgs`.`time_next` < 9223372036854775807))"
      }
    }
  }
}
```

go/vt/vttablet/tabletserver/messager/message_manager.go
derekperkins: Should we add a comment in the code explaining why the ordering? If you got it backwards after diving in deep, it seems likely that we'll have no idea in a few months. :)

go/vt/vttablet/tabletserver/messager/message_manager.go
derekperkins: > Can you please include the query along with the explain plan? Can you also please share the index definitions? From your explain output here, I think you're using the new updated/proposed query but the new time_acked__priority__time_next index is not in DESC order for the time_next field.

Just saw this. I am using the updated query referencing `time_acked IS NULL`, and my multi-column indexes have always had `time_next desc`. There is a separate `time_next ASC` index that is being selected sometimes, since it matches the `time_next < ?` clause better. Does that answer your question?

```sql
  KEY `time_next` (`time_next`),
  KEY `time_acked__priority__time_next` (`time_acked`,`priority`,`time_next` DESC),
  KEY `time_acked__priority__time_next__epoch__id__keyspace_id` (`time_acked`,`priority`,`time_next` DESC,`epoch`,`id`,`keyspace_id`),
```

mattlord: I believe that I've now addressed the deadlocks. I am now leaning toward removing the query perf aspects and we can investigate and play with that more separately. What do you think? I would remove that index change from the docs PR too.

derekperkins: Let's get it merged. I'm happy leaving the query perf in, as I did see some clear wins on the larger queries, but also happy to kick it to a separate PR

derekperkins (approved):

go/vt/vttablet/tabletserver/messager/message_manager.go
mattlord: OK, thanks! Sorry, I hadn't yet caught up on all of your more recent tests and comments. THANK YOU for that. I'll keep those changes in then.

derekperkins:
Looks like the majority of tests are failing on 

```
--- FAIL: TestMessageStreamingPlan (0.00s)
    plan_test.go:239: 
        	Error Trace:	plan_test.go:239
        	Error:      	Received unexpected error:
        	            	Code: FAILED_PRECONDITION
        	            	'msg' is not a message table
        	Test:       	TestMessageStreamingPlan
```

derekperkins:
@mattlord you really rolled through quite a few iterations, thanks again for digging in so deep. I don't think I have enough context around vstream to provide much feedback on the latest design. Once @rohit-nayak-ps approves it, I'm ready to roll it out in prod to test it out.

mattlord:
Note that the Cluster 20 -> backup_xtrabackup test failure is due to: https://github.com/vitessio/vitess/issues/10146

derekperkins:
@mattlord is this still waiting on @rohit-nayak-ps, or was pairing with @sougou sufficient review?

sougou (approved):

derekperkins:
🎉  Thanks so much for tracking this down!! I'll be rolling it out later today

mattlord:
> 🎉 Thanks so much for tracking this down!! I'll be rolling it out later today

Excellent! Please let me know how things go.

PR #10135, opened 2022-04-25
fix: remove set_var session field from proto

harshit-gangal:
## Checklist


go/vt/vtgate/executor_select_test.go
```diff
@@ -160,8 +160,9 @@ func TestSystemVariablesMySQLBelow80(t *testing.T) {
 	executor.normalize = true
 
 	sqlparser.MySQLVersion = "57000"
+	*setVarEnabled = true
```
frouioui, line 163: shouldn't we add a defer to set `setVarEnabled` to false?

frouioui (approved):

go/vt/vtgate/executor_select_test.go
harshit-gangal: next PR I can do it as it is not important.

proto/vtgate.proto
```diff
@@ -147,9 +147,8 @@ message Session {
   // enable_system_settings defines if we can use reserved connections.
   bool enable_system_settings = 23;
 
-  // enable_set_var enables the use of SET_VAR query hint to reduce the number of reserved connection
-  // this feature is part of this RFC (https://github.com/vitessio/vitess/issues/9706)'s second proposal
-  bool enable_set_var = 24;
+  // enable_set_var is not required in session
+  reserved 24;
 }
```
deepthi, line 152: Since the original change in #9641 did not go into v13, why do we need to reserve this number? We can delete it.

harshit-gangal: I reserved it for the reason if someone picked up a build from main and so it does not break for them.

harshit-gangal: I am ok to delete this, kindly confirm

deepthi: Should be fine. We don't have to maintain compatibility between day to day builds off main.

PR #10137, opened 2022-04-25
CODEOWNERS: specify messaging related folders

derekperkins:
As more people have gotten up to speed with messaging, I think it makes sense to be specific in CODEOWNERS. This incidentally also removes since deleted helm permissions


mattlord (approved):
Agreed. Thanks! Just need to fix the DCO.

PR #10138, opened 2022-04-25
Avoid enabling reserved connection on show query

harshit-gangal:
## Description

This PR avoids using the reserved connection for Show query and handles `show variables` by doing variable replace at vtgate level.

## Checklist

### DIFF
--- a/go/vt/vtgate/executor_set_test.go
+++ b/go/vt/vtgate/executor_set_test.go
@@ -21,6 +21,8 @@ import (
 	"fmt"
 	"testing"
 
+	"vitess.io/vitess/go/vt/sqlparser"
+
 	querypb "vitess.io/vitess/go/vt/proto/query"
 
 	"vitess.io/vitess/go/test/utils"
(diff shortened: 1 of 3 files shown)


go/vt/vtgate/executor_set_test.go
```diff
@@ -485,3 +487,76 @@ func createMap(keys []string, values []any) map[string]*querypb.BindVariable {
...
+		rc  bool
+	}{
+		{sql: "select 1 from user"},
+		{sql: "update user set col = 2"},
+		{sql: "delete from user"},
+		{sql: "insert into user (id) values (1)"},
+		{sql: "replace into user(id, col) values (1, 'new')"},
+		{sql: "set autocommit = 0"},
+		{sql: "show create table user"}, // reserved connection should not be set.
+		{sql: "create table foo(bar bigint)", rc: true},
```
frouioui, line 520: Can you remind me why the reserved connection is used for the create table?

frouioui (approved):
lgtm

go/vt/vtgate/executor_set_test.go
frouioui: got it

Issue #9889, opened 2022-03-14
Bug Report: false negative sql_mode=only_full_group_by error message from vtgate


Santiclause:
### Overview of the Issue

Issuing a query with aggregation which selects a non-aggregated column which is also not in the group by clause will fail with the error `Expression of SELECT list is not in GROUP BY clause and contains nonaggregated column [...] which is not functionally dependent on columns in GROUP BY clause; this is incompatible with sql_mode=only_full_group_by`. This error message indicates that the problem is with `sql_mode`, but there is no `sql_mode` setting involved here; this error happens at the vtgate before it reaches any tablets, and occurs irrespective of your actual `sql_mode` setting on your tablets (we do not have this `only_full_group_by` setting set).

It also appears that the query has to have table aliasing in order to run into this problem.

### Reproduction Steps

Example vschema:
```json
{
  "tables": {
    "foo": {
      "column_vindexes": [
        {
          "column": "id",
          "name": "foo_id_hash"
        }
      ]
    },
    "bar": {
      "column_vindexes": [
        {
          "column": "foo_id",
          "name": "foo_id_hash"
        }
      ]
    }
  },
  "vindexes": {
    "foo_id_hash": {
      "type": "hash"
    }
  },
  "sharded": true
}
```

Example schema:
```sql
CREATE TABLE `foo` (
  `id` bigint(20) unsigned NOT NULL AUTO_INCREMENT,
  `name` varchar(255) DEFAULT NULL,
  PRIMARY KEY (`id`)
) ENGINE=InnoDB DEFAULT CHARSET=latin1;
CREATE TABLE `bar` (
  `id` bigint(20) unsigned NOT NULL AUTO_INCREMENT,
  `foo_id` bigint(20) unsigned NOT NULL,
  PRIMARY KEY (`id`)
) ENGINE=InnoDB DEFAULT CHARSET=latin1;
```

Observe that the following query _succeeds_:
```sql
SELECT foo.id, foo.name, COUNT(bar.id)
FROM foo
LEFT JOIN bar on bar.foo_id = foo.id
GROUP BY foo.id;
```

Yet _this_ query, identical in all respects other than table alias, fails:
```sql
SELECT f.id, f.name, COUNT(bar.id)
FROM foo f
LEFT JOIN bar on bar.foo_id = f.id
GROUP BY f.id;
```

### Binary Version

```sh
[...]

PR #10139, opened 2022-04-25
Add support for non aggregated columns in `OrderedAggregate`

frouioui:
## Description

We do not keep track of the [ONLY_FULL_GROUP_BY](https://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sqlmode_only_full_group_by) setting in VTGate, thus we were failing certain query like the one below even though they should be executed if the setting is set to false.

```
select city, age from user group by city
# age is a non aggregated column here, the query should fail if sql_mode=only_full_group_by
```

This PR adds support for non aggregated columns in the `OrderedAggregate`. We should be able to plan a query that has non aggregated columns and send it to MySQL. Bases on the MySQL mode, MySQL can decide to fail this query or not.

## Related Issue(s)

- fixes https://github.com/vitessio/vitess/issues/9889

## Checklist


frouioui:
This Pull Request depends on https://github.com/vitessio/vitess/pull/10163 for a fix in how we handle reserved connections for field queries. Do not merge until https://github.com/vitessio/vitess/pull/10163 is merged.

derekperkins:
To clarify on the description, this will also allow for similar queries to succeed even if [ONLY_FULL_GROUP_BY](https://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sqlmode_only_full_group_by) is enabled, when MySQL can determine that those columns are functionally dependent on a grouping column

go/vt/vtgate/engine/ordered_aggregate.go
```diff
@@ -102,8 +102,9 @@ type AggregateParams struct {
 	WAssigned   bool
 	CollationID collations.ID
 
-	Alias string `json:",omitempty"`
-	Expr  sqlparser.Expr
+	Alias    string `json:",omitempty"`
+	Expr     sqlparser.Expr
+	Original *sqlparser.AliasedExpr
```
systay, line 107: Why do we need `original`? The only uses I see should be possible to handle using the Expr field, right?

frouioui: `Expr` field is the AST node that contains the aggregate function (i.e. `count(id)`), whereas `Original` is simply `id`.

frouioui: `Expr` field is the AST node that contains the aggregate function (i.e. `count(id)`), whereas `Original` is simply `id`.

systay: Hmm... Now am I'm even more confused. Why is the expression wrapped in AliasedExpr?

go/vt/vtgate/planbuilder/aggregation_pushing.go
```diff
@@ -28,17 +28,11 @@ import (
...
 
 // pushAggregation pushes grouping and aggregation as far down in the tree as possible
-func (hp *horizonPlanning) pushAggregation(
-	ctx *plancontext.PlanningContext,
-	plan logicalPlan,
-	grouping []abstract.GroupBy,
-	aggregations []abstract.Aggr,
-	ignoreOutputOrder bool,
-) (output logicalPlan, groupingOffsets []offsets, outputAggrsOffset [][]offsets, err error) {
+func (hp *horizonPlanning) pushAggregation(ctx *plancontext.PlanningContext, plan logicalPlan, grouping []abstract.GroupBy, aggregations []abstract.Aggr, ignoreOutputOrder bool) (output logicalPlan, groupingOffsets []offsets, outputAggrsOffset [][]offsets, err error) {
```
systay, line 31: nit: these lines are difficult to read when in a single line. mind splitting it again?

go/vt/vtgate/planbuilder/aggregation_pushing.go
```diff
@@ -373,7 +371,7 @@ func splitAggregationsToLeftAndRight(
 	var lhsAggrs, rhsAggrs []*abstract.Aggr
 	for _, aggr := range aggregations {
 		newAggr := aggr
-		if isCountStar(aggr.Func) {
+		if aggr.Func != nil && isCountStar(aggr.Func) {
```
systay, line 374: nit: you could move in the nil checking into `isCountStar`

go/vt/vtgate/engine/ordered_aggregate.go
```diff
@@ -476,6 +478,7 @@ func merge(
 			data, _ := proto.Marshal(vgtid)
 			val, _ := sqltypes.NewValue(sqltypes.VarBinary, data)
 			result[aggr.Col] = val
+		case AggregateRandom:
```
systay, line 481: this is correct, but also a little confusing. maybe add a comment explaining what's going on here?

go/vt/vtgate/planbuilder/abstract/queryprojection.go
```diff
@@ -46,7 +46,6 @@ type (
 		OrderExprs         []OrderBy
 		CanPushDownSorting bool
 		HasStar            bool
-		ProjectionError    error
```
systay, line 49: 🎉

systay (approved):
LGTM

You probably need to merge main into this - the plantest outputs have changed.

systay:
we should probably write something about this in our docs, so that people know that Vitess allows it and that the functionality depends on mysql to also allow it

go/vt/vtgate/engine/ordered_aggregate.go
systay: Never mind. I get it now

frouioui:
> we should probably write something about this in our docs, so that people know that Vitess allows it and that the functionality depends on mysql to also allow it

I feel like this PR is enhancing the compatibility with MySQL by letting MySQL handle more work. We are just mimicking the behavior that users expect when using MySQL.

PR #10140, opened 2022-04-26
refactor: removed last reserved field from vtgate session

harshit-gangal:
## Description

This PR updates the proto by removing the field introduced in https://github.com/vitessio/vitess/pull/9641

## Checklist


harshit-gangal:
https://github.com/vitessio/vitess/pull/10135

systay (approved):

deepthi:
For context: https://github.com/vitessio/vitess/pull/10135#discussion_r857949480
This is pretty low-risk, since for someone to be affected, they need to be running a daily build + write a gRPC client that depends on the new field introduced after v13 was released.

PR #10141, opened 2022-04-26
feat: add max_repl_lag func

systay:
## Description
Adds a new function, `max_repl_lag()` that returns the largest current replication lag to any tablet. 

## Checklist


go/test/endtoend/tabletgateway/vtgate_test.go
```diff
@@ -66,7 +66,15 @@ func TestVtgateReplicationStatusCheck(t *testing.T) {
...
-	assert.Equal(t, expectNumRows, numRows, fmt.Sprintf("wrong number of results from show vitess_replication_status. Expected %d, got %d", expectNumRows, numRows))
+	assert.Equal(t, expectNumRows, numRows, "wrong number of results from show vitess_replication_status")
+
+	qr = utils.Exec(t, conn, "select max_repl_lag()")
+	expectNumRows = 1
+	numRows = len(qr.Rows)
+	assert.Equal(t, expectNumRows, numRows, "expected single row from `select max_repl_lag()`")
+
+	_ = utils.Exec(t, conn, "use @spare")
+	utils.AssertMatches(t, conn, `select max_repl_lag()`, `[[NULL]]`)
```
deepthi, line 77: I would make the function name `max_replication_lag` (repl fully spelled out instead of contracted)

go/vt/vtgate/executor.go
```diff
@@ -884,6 +886,45 @@ func (e *Executor) showVitessReplicationStatus(ctx context.Context, filter *sqlp
 	}, nil
 }
 
+func (e *Executor) getMaxReplicationLag(session *SafeSession) sqltypes.Value {
+	status := e.scatterConn.GetHealthCheckCacheStatus()
+	replLag := -1
+	targetString := session.GetTargetString()
```
deepthi, line 892: If user doesn't specify a tablet_type, does this default to `PRIMARY`?
For a default connection aka primary connection, we want lag to be reported as 0.
Let us add a test to verify this use case.

harshit-gangal: if no tablet_type is provided then it will go through all the tablets and report maximum replication delay.
Only when a specific tablet_type is provided via targetString then only for that tablet the replication delay will be reported.

go/test/endtoend/tabletgateway/vtgate_test.go
harshit-gangal: I think we can name it `vt_max_replication_lag` so it does not interfere with any MySQL functions.

go/vt/vtgate/executor.go
deepthi: If I connect via mysql client to vtgate and don't issue a `use`, is tablet_type set?

PR #10147, opened 2022-04-26
bootstrap changelog

deepthi:
## Description
When we started versioning the bootstrap images, a Changelog file was added to provide a human-readable history.
This PR adds all known versions to that file.

## Checklist

-   [ ] "Backport me!" label has been added if this change should be backported NO


vmg (approved):
Ah this makes a lot of sense. Will keep in mind the next time I do a bootstrap bump.

PR #10150, opened 2022-04-27
VTOrc: Refactor and reload of ephemeral information for remaining recovery functions

GuptaManan100:
## Description

This is the follow up PR for https://github.com/vitessio/vitess/pull/10115.

In that PR, we had added the functionality to VTOrc to refresh its ephemeral information before running any cluster operation. But according to our discussions, we should be doing this for all failure scenarios. For example, we see that a tablet has its replication stopped, but we might be using old information and actually it is a BACKUP tablet_type. So, even in these cases, we should not be going ahead with the recovery to start the replication, since it might break other parts of Vitess. This PR addresses that concern and adds this reload of information for the other recovery functions as well. 

In this PR, we have also adds some unit tests to VTOrc. We have also refactored the replicationAnalysis codepath to do the locking of the shard and refreshing of the information before we call the recoveryFunction, to prevent repeated code. This PR also deletes any unused failure scenarios that were coming from `orchestrator` legacy code. Vitess doesn't support Co-primary or intermediate-primary configuration, so it is safe to remove these failure scenarios. We anyways didn't have any associated recoveries for them.

Moreover, the codepath for `PrimaryHasPrimary` failure has been refactored, to only reset the replication on the primary instance, instead of running a force `DeadPrimary` recovery. 

## Related Issue(s)

- #8975 

## Checklist


go/vt/orchestrator/logic/topology_recovery.go
```diff
@@ -1305,68 +1210,61 @@ func checkAndExecuteFailureDetectionProcesses(analysisEntry inst.ReplicationAnal
...
 ) {
 	switch analysisCode {
 	// primary
 	case inst.DeadPrimary, inst.DeadPrimaryAndSomeReplicas:
 		if isInEmergencyOperationGracefulPeriod(analyzedInstanceKey) {
-			return checkAndRecoverGenericProblem, false
+			return checkAndRecoverGenericProblem, 1, false
 		}
-		return checkAndRecoverDeadPrimary, true
+		return recoverDeadPrimary, 2, true
```
GuptaManan100, line 1227: I am a little unsure of this way. We need some way to compare that the function we get back are the same, but golang doesn't support equality comparison on functions. So, I have returned a integer for each of the functions, which is unique for it and used it for checkin equality. If we prefer, I can make this an enum and give the integers more meaningful variable names. Maybe @vmg has a better solution?

go/vt/orchestrator/inst/instance_topology.go
```diff
@@ -929,7 +929,7 @@ func MakeCoPrimary(instanceKey *InstanceKey) (*Instance, error) {
 	}
 	log.Infof("Will make %+v co-primary of %+v", instanceKey, primary.Key)
 
-	var gitHint OperationGTIDHint = GTIDHintNeutral
```
GuptaManan100, line 932: This is just a linter fix

go/vt/orchestrator/logic/topology_recovery.go
harshit-gangal: You can register the function and then check if they correspond to the same.

GuptaManan100: Right, so a map from a string to the function and use the string for checking equality?

deepthi (commented):
Mostly looks fine to me.
Can you link the catch-all vtorc issue into this PR?
I went and looked at how we decide on `PrimaryHasPrimary` and now I'm having second thoughts about simply removing the replication. Even though the tablet type is PRIMARY, we should make sure that the underlying mysqld is writable as well. That doesn't seem to be something that is considered when setting `IsClusterPrimary` to true.

GuptaManan100:
The PRIMARY tablet not being writable is a failure scenario in itself. If the designated primary is read only, it will spawn the `PrimaryIsReadOnly` failure which will set the primary to read-write.
I am unsure of what to do about `PrimaryHasPrimary` too, the failure could be as simple as someone messed up when entering the tablet host port when they wanted to the set replication source (maybe for experimental tablets, etc?). The earlier code assumed that something was seriously wrong for the Primary to have a replication source set, and ran an ERS forcefully.

go/vt/orchestrator/logic/topology_recovery.go
vmg: You could simply return the same of the function as a string, e.g. `"recoverDeadPrimary"` and compare that. You don't need a map. I think pretty much anything works here, the only difficulty is making sure that you're not reusing IDs, which will lead to bugs, which is why returning a number seems questionable.

GuptaManan100: Thankyou @vmg That makes sense. I was feeling iffy with the integers too. I'll change them to an enum instead, which would give them more meaningful names and we would also not have the problems of typing errors in the name of the returned function!

GuptaManan100: Fixed it.

deepthi (approved):

PR #10151, opened 2022-04-27
feat: make gen4 plan subsharding queries better

systay:
## Description
The gen4 planner was treating partial vindex lookups as equivalent to vindex lookups with all columns accounted for.
This lead to bad plans when some of the column comparisons were `IN` comparisons.

## Checklist


go/vt/vtgate/planbuilder/testdata/dml_cases.txt
```diff
@@ -3654,7 +3679,26 @@ Gen4 plan same as above
...
+      "Sharded": true
+    },
+    "TargetTabletType": "PRIMARY",
+    "MultiShardAutocommit": false,
+    "Query": "update multicol_tbl set x = 1 where `name` = 'foo' and cola = 2",
+    "Table": "multicol_tbl",
+    "Values": [
+      "VARCHAR(\"foo\")"
+    ],
+    "Vindex": "name_muticoltbl_map"
```
systay, line 3699: Not sure that a lookup vindex is a better plan than doing a SubShard query against the primary vindex, but this should be addressed separately, imho.

frouioui (approved):

go/vt/vtgate/engine/routing.go
```diff
@@ -68,8 +68,8 @@ const (
 	// Is used when the query explicitly sets a target destination:
 	// in the clause e.g: UPDATE `keyspace[-]`.x1 SET foo=1
 	ByDestination
-	// NumOpcodes is the number of opcodes
-	NumOpcodes
+	// SubShard is for when we are missing one or more columns from a composite vindex
```
harshit-gangal, line 71: adding it here does not mean that Scatter is better than SubShard?

go/vt/vtgate/planbuilder/route_test.go
```diff
@@ -28,22 +28,26 @@ import (
...
+	Equal       	 2
+	IN          	 3
+	MultiEqual  	 4
+	Scatter     	 5
+	Next        	 6
+	DBA         	 7
+	Reference   	 8
+	None        	 9
+	Subshard         10 <- not covered
+	NumRouteOpcodes  11
```
harshit-gangal, line 46: NumRouteOpcodes is removed from opcode list

harshit-gangal (commented):
left some comments

PR #10152, opened 2022-04-27
Fix for empty results when no shards can be found to route to

### LINKED ISSUE #10144: Bug Report: exists may returns random result on a sharded table
### Overview of the Issue

We found an issue with exists : I can't provide the exact query we use, but for the query like 
```
select exists(select * from foo where some_field=some_value)
```
here foo is in a sharded keyspace, it seems if the inner query has result in one shard but have no result in another shard, then the outer query will return a random 0 or 1.  Not sure if it is related, but for our particular query, the table is sharded on the field 'some_field', so if Vitess is smart enough, it doesn't need to send the query to all the shard. 
We tried using gen4 planner, got the same result.

### Reproduction Steps

See description.

### Binary Version

```sh
v13.0.0
```

```sh
# cat /etc/os-release
PRETTY_NAME="Debian GNU/Linux 10 (buster)"
NAME="Debian GNU/Linux"
VERSION_ID="10"
VERSION="10 (buster)"
VERSION_CODENAME=buster
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
```

### Log Fragments

_No response_

GuptaManan100:
## Description
Queries that follow the following pattern - 
```mysql
select exists(select * from foo where some_field=some_value)
```

The table `foo` would has a lookup vindex on `some_field` and there are no matching rows for `some_value`. What Vitess would do when no shards could be found to route to is to return an empty result. This is correct for most cases, but not always. The query above would return an empty result instead of a `0`.

## Related Issue(s)
- Fixes #10144 

## Checklist
-   [x] "Backport me!" label has been added if this change should be backported YES

### DIFF
--- a/go/vt/vtgate/planbuilder/gen4_planner.go
+++ b/go/vt/vtgate/planbuilder/gen4_planner.go
@@ -96,6 +96,10 @@ func gen4SelectStmtPlanner(
 
 	primitive := plan.Primitive()
 	if rb, ok := primitive.(*engine.Route); ok {
+		// this is done because engine.Route doesn't handle the empty result well
+		// if it doesn't find a shard to send the query to.
+		// All other engine primitives can handle this, so we only need it when
+		// Route is the last (and only) instruction before the user sees a result
 		rb.NoRoutesSpecialHandling = true
 	}
 
--- a/go/vt/vtgate/planbuilder/gen4_planner.go
+++ b/go/vt/vtgate/planbuilder/gen4_planner.go
@@ -93,7 +93,13 @@ func gen4SelectStmtPlanner(
 			return primitive, nil
 		}
 	}
-	return plan.Primitive(), nil
+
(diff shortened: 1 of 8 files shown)


go/vt/vtgate/planbuilder/gen4_planner.go
```diff
@@ -93,7 +93,13 @@ func gen4SelectStmtPlanner(
 			return primitive, nil
 		}
 	}
-	return plan.Primitive(), nil
+
+	primitive := plan.Primitive()
+	if rb, ok := primitive.(*engine.Route); ok {
```
GuptaManan100, line 98: I think a comment here as to why the previous behaviour is desirable for JOINs and others would be good.

GuptaManan100 (commented):
Everything else LGTM!

PR #10161, opened 2022-04-28
schemadiff: EntityDiff has a 'Entities() (from Entity, to Entity)' function

shlomi-noach:
## Description

`Entities() (from Entity, to Entity)` return from "from" and "to" entities for the diff.
- When the diff is e.g. a `AlterTableEntityDiff`, this returns both the "from"'s CreateTable and the "to"s CreateTable.
- if the diff is a `CreateTableEntityDiff` then there's only a "to" ; "from" is `nil`.
- if the diff is a `DropTableEntityDiff` then there's only a "from" ; "to" is `nil`.

this is symmetrical in View diffs.

## Checklist

cc @dbussink

### DIFF
--- a/go/vt/schemadiff/diff_test.go
+++ b/go/vt/schemadiff/diff_test.go
@@ -27,13 +27,15 @@ import (
 
 func TestDiffTables(t *testing.T) {
 	tt := []struct {
-		name    string
-		from    string
-		to      string
-		diff    string
-		cdiff   string
-		action  string
-		isError bool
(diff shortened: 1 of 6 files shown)


dbussink (approved):

shlomi-noach:
`Cluster (20)` is the remaining failing test. At this time we know it ia flaky and not `rewuired`. Merging.

shlomi-noach:

PR #10148, opened 2022-04-27
Do not mutate replication state in WaitSourcePos

mattlord:
## Description

Mutating the state there results in race conditions with tablet repairs that can leave
replication broken and requiring manual repair using using `START SLAVE`.

This reverts the state mutating parts from https://github.com/vitessio/vitess/pull/9512 and the follow-up https://github.com/vitessio/vitess/pull/10104.

It leaves the other helper functions in place as they could be useful for other purposes at some point. They are effectively unused though so we could remove them as well since it's currently dead code.

## Related Issue(s)
 - https://github.com/vitessio/vitess/pull/9512 — this issue needs to be re-opened and addressed differently
 - https://github.com/vitessio/vitess/pull/10104 — this is effectively reverted (the state mutating parts anyway, the other functions could be useful at some point)

## Checklist

-   [x] "Backport me!" label has been added if this change should be backported (this should be backported to release-13.0)
-   [ ] Tests were added or are not required
-   [x] Documentation is not required

PR #10167, opened 2022-04-28
Backwards compatible replication status to state transition

mattlord:
## Description

It turns out that this was NOT fully backwards compatible: https://github.com/vitessio/vitess/pull/9853/files

Later ERS related upgrade/downgrade tests added in https://github.com/vitessio/vitess/pull/10148 showed this. The problem was that a v14 `vtctl` (client: `vtctlclient`->`vtctld` or `vtctl`) would send an RPC to a v13 `vttablet` (server) and the `vttablet` would respond with the v13 `ReplicationStatusResponse` message that does not have the `IoStatus` or `SqlStatus` fields in it so the v14 `vtctl` would process that response and consider the IO and SQL statuses as Unknown and be unable to proceed with some actions (such as reparenting). In order to support a v14 `vtctl` with v13 `vttablet`s, when processing the RPC response we need to check for an IO and SQL state of Unknown and revert to using the binary IO and SQL thread running values (the older tablets cannot tell us if the IO state was connecting). 

In order to support correct RPC communication between v13 and v14 we must also leave the old fields in place so that it's a clean upgrade from v13 to v14 — and the v14 `vtctl` will process the older `io_thread_running` and `sql_thread_running` `ReplicationStatusResponse` fields — while adding the new fields for v14+ `vttablets` and clients to use. This way in v15+ we can remove the old fields and complete the transition for smooth upgrades. 

## Related Issue(s)
  - Follow-up to: https://github.com/vitessio/vitess/pull/9853
  - Blocking: https://github.com/vitessio/vitess/pull/10148

## Checklist

-   [x] Tests were added. Also:
    - Existing tests should continue to pass
    - The backwards compatibility is tested in new vttablet upgrade/downgrade tests added in https://github.com/vitessio/vitess/pull/10148
-   [x] Documentation is not required (internal)


go/mysql/replication_status.go
```diff
@@ -129,11 +136,20 @@ func ProtoToReplicationStatus(s *replicationdatapb.Status) ReplicationStatus {
...
+		replstatus.IOState = ReplicationStateConnecting
+	} else {
+		replstatus.IOState = ReplicationStateStopped
+	}
+	if s.SqlThreadRunning {
+		replstatus.SQLState = ReplicationStateRunning
+	} else {
+		replstatus.SQLState = ReplicationStateStopped
+	}
+	return replstatus
```
GuptaManan100, line 152: I think we should do it the other way around. Like so - 
```go
if replstatus.IOState = ReplicationStateUnknown {
  if s.IOThreadRunning {
    replSTatus.IOState = ReplicationStateRunning
  } else {
    replStatus.IOState = replicationstatestopped
  }
}
```

go/vt/vtctl/reparentutil/replication.go
```diff
@@ -150,7 +150,12 @@ func ReplicaWasRunning(stopStatus *replicationdatapb.StopReplicationStatus) (boo
 		return false, vterrors.Errorf(vtrpc.Code_INVALID_ARGUMENT, "could not determine Before state of StopReplicationStatus %v", stopStatus)
 	}
 
-	return stopStatus.Before.IoState == int32(mysql.ReplicationStateRunning) || stopStatus.Before.SqlState == int32(mysql.ReplicationStateRunning), nil
+	// v13 and older tablets will only provide the binary IoThreadRunning and
+	// SqlThreadRunning values while v14 and newer tablets will provide the
+	// non-binary replication states.
+	// This backwards compatible check can be removed in v15+.
+	return ((stopStatus.Before.IoState == int32(mysql.ReplicationStateRunning) || stopStatus.Before.IoThreadRunning) ||
+		(stopStatus.Before.SqlState == int32(mysql.ReplicationStateRunning) || stopStatus.Before.SqlThreadRunning)), nil
```
GuptaManan100, line 158: We can get rid of this here, if we first converted the Before value using `ProtoToReplicationStatus` and just used the IoState. That will ensure that we have the upgrade downgrade logic, all in one file. WDYT? I am going ahead with the change, we can revert if you don't agree

go/mysql/replication_status.go
```diff
@@ -91,6 +91,21 @@ func ReplicationStatusToProto(s ReplicationStatus) *replicationdatapb.Status {
 		SqlState:              int32(s.SQLState),
 		LastSqlError:          s.LastSQLError,
 	}
+
+	// We need to be able to send gRPC response messages from v14 and newer tablets to
```
deepthi, line 95: 👍

deepthi (approved):
lgtm.
Apologies for missing the compatibility issue on the earlier PR.

PR #10174, opened 2022-04-29
feat: optimize EXISTS queries through AST rewriting

systay:
## Description
For EXISTS queries, we just need to find out if the subquery returns at least one row. This PR adds an AST rewriter that rewrites EXISTS expression to be simpler and faster.

Normally this is an optimization that MySQL would do internally, but since we are splitting up the query, MySQL might not find out that the result of a query will be used for an existential test.

## Checklist


frouioui (approved):

GuptaManan100 (approved):
These changes look good to me! Gen4 is getting 😎  and I ❤️ it!

PR #10177, opened 2022-04-29
Delay sysvars init;  it pulls in collations, which depends on

aquarapid:
mysql_server_version flag in servenv already being initialized

This showed up as `vtcombo` ignoring the `-mysql_server_version` flag and always thinking that it was running against 8.0.  This would manifest itself in vtcombo always using `utf8` against MySQL 5.7, even if it was running with `-db_charset utf8mb4`


aquarapid:
Closing, since this won't work, see #10193

Issue #10146, opened 2022-04-26
Xtrabackup tests are failing


deepthi:
### Overview of the Issue

Cluster(20) has been failing for the last 12 hours or so, with most of the xtrabackup tests failing. The ones using xbstream are still passing.
Since we download the latest binary in CI, this may have nothing to do with our code, but with some change to the binary. Needs to be investigated.

### Reproduction Steps

See https://github.com/vitessio/vitess/actions/workflows/cluster_endtoend_20.yml

### Binary Version

```sh
main
```

### Operating System and Environment details

```sh
CI
```

### Log Fragments

```sh
I0426 21:08:30.081012   18259 vtctlclient_process.go:209] Executing vtctlclient with command: vtctlclient --server localhost:16006 Backup zone1-0000002642 (attempt 1 of 10)
E0426 21:08:33.658627   18259 vtctlclient_process.go:186] Output:
E0426 21:08:33.657708   21427 main.go:94] E0426 21:08:33.657563 backup.go:107] E0426 21:08:33.657289 backup.go:127] backup is not usable, aborting it: [exit status 2
xtrabackup failed with error. Output=]
Backup Error: rpc error: code = Unknown desc = TabletManager.Backup on zone1-0000002642 error: xtrabackup failed with error. Output=: exit status 2: xtrabackup failed with error. Output=: exit status 2
E0426 21:08:33.657991   21427 main.go:103] remote error: rpc error: code = Unknown desc = TabletManager.Backup on zone1-0000002642 error: xtrabackup failed with error. Output=: exit status 2: xtrabackup failed with error. Output=: exit status 2
    backup_utils.go:603: 
        	Error Trace:	backup_utils.go:603
        	            				backup_utils.go:222
        	Error:      	Expected nil, but got: &exec.ExitError{ProcessState:(*os.ProcessState)(0xc0005660a8), Stderr:[]uint8(nil)}
        	Test:       	TestXtrabackup/TestReplicaBackup
```

deepthi:
@mattlord pointed out that a new release came out today
https://docs.percona.com/percona-xtrabackup/2.4/release-notes/2.4/2.4.25.html

mattlord:
Here's the problem. `xtrabackup` is crashing:
```
xtrabackupengine.go:147] Detected MySQL flavor: MySQL56
xtrabackupengine.go:157] Starting backup with 0 stripe(s)
[...]

PR #10180, opened 2022-04-29
[vtadmin-web] Remove placeholder links from NavRail

doeg:
## Description

- Removes the placeholder links in the NavRail, since they just 404. :) 

- Uses text instead of SVG icons for the single-character NavRail "hotkeys"; scare quotes since they aren't bound to any keyboard shortcuts (yet). I opted for this change as I moved "Workflows" under the other infrastructure links (i.e., those with counts) but the previous icon, a wrench, looked out of place... and exporting icons from Figma is sort of a pain. (Even if we switch to something like fontawesome for our icons, I think rendering these "icons" with CSS is still more lightweight + consistent than messing around with SVG!) 

- Deletes the `letter*.svg` files that we're no longer using. 

Before and after, for comparison. The new "icons" are noticeably lighter, so I'm definitely open to opinions. :) 

**Before:**
Screen Shot 2022-04-29 at 3 21 01 PM

**After:**
Screen Shot 2022-04-29 at 3 20 54 PM

## Related Issue(s)

N/A

## Checklist


web/vtadmin/src/components/NavRail.module.scss
```diff
@@ -117,3 +117,24 @@ a.navLink:focus .icon,
 a.navLink:hover .icon {
     fill: var(--colorPrimary);
 }
+
+.hotkey::before {
+    $hotkeySize: 16px;
+
+    @apply rounded-sm;
```
doeg, line 124: I never really know where to draw the line between `@apply` and just writing out CSS rules with variables, so open to suggestions here!

notfelineit: Me neither! I'd only say that since we have tailwind, I'd personally use it to its ✨ fullest ✨

notfelineit (approved):
I like the hotkey abstraction!!

doeg:
The `Cluster (20)` test is failing (see https://github.com/vitessio/vitess/issues/10146) but since it's not actually required, I'm going to go ahead and merge this in. :)

Issue #8969, opened 2021-10-10
Flaky tests: onlineddl_vrepl_stress_suite


shlomi-noach:
Example for flaky run: https://github.com/vitessio/vitess/runs/3800279012
```
--- FAIL: TestSchemaChange (721.18s)
...
    --- PASS: TestSchemaChange/different_PRIMARY_KEY_vs_UNIQUE_KEY (52.39s)
        --- PASS: TestSchemaChange/different_PRIMARY_KEY_vs_UNIQUE_KEY/cancel_pending_migrations (0.00s)
        --- PASS: TestSchemaChange/different_PRIMARY_KEY_vs_UNIQUE_KEY/create_schema (0.26s)
        --- PASS: TestSchemaChange/different_PRIMARY_KEY_vs_UNIQUE_KEY/prepare_table (0.00s)
        --- PASS: TestSchemaChange/different_PRIMARY_KEY_vs_UNIQUE_KEY/init_table_data (4.66s)
        --- PASS: TestSchemaChange/different_PRIMARY_KEY_vs_UNIQUE_KEY/migrate (47.47s)
    --- FAIL: TestSchemaChange/no_shared_UK,_multiple_options (65.56s)
        --- PASS: TestSchemaChange/no_shared_UK,_multiple_options/cancel_pending_migrations (0.00s)
        --- PASS: TestSchemaChange/no_shared_UK,_multiple_options/create_schema (0.26s)
        --- PASS: TestSchemaChange/no_shared_UK,_multiple_options/prepare_table (0.01s)
        --- PASS: TestSchemaChange/no_shared_UK,_multiple_options/init_table_data (4.75s)
        --- FAIL: TestSchemaChange/no_shared_UK,_multiple_options/migrate (60.55s)
    --- PASS: TestSchemaChange/fail;_no_uk_on_target (5.93s)
        --- PASS: TestSchemaChange/fail;_no_uk_on_target/cancel_pending_migrations (0.03s)
        --- PASS: TestSchemaChange/fail;_no_uk_on_target/create_schema (0.25s)
        --- PASS: TestSchemaChange/fail;_no_uk_on_target/prepare_table (0.00s)
        --- PASS: TestSchemaChange/fail;_no_uk_on_target/init_table_data (4.58s)
        --- PASS: TestSchemaChange/fail;_no_uk_on_target/migrate (1.07s)
...
FAIL
[...]

PR #10181, opened 2022-04-29
Propagate error from 'updateLocked' to client

### LINKED ISSUE #10145: Bug Report: ChangeTabletType is considered success without updating all the internal states
#### Background
We did some manual operation on a shard to mitigate an unrelated issue, which unexpectedly broke the shard in such a way that any newly restored tablet will run into the following problem
```
Tablespace is missing for table `vt_<redacted>`.`_bfd62e12_a055_11ec_a8d6_022a23b59a1c_20220310093754_vrepl`. (errno 1812) (sqlstate HY000) during query: select * from _bfd62e12_a055_11ec_a8d6_022a23b59a1c_20220310093754_vrepl where 1 != 1
```
#### What happened?
We see the above error from the following ChangeTabletType code paths
https://github.com/vitessio/vitess/blob/40df731e82daa28cfa5b89024dca12c60958f9e1/go/vt/vttablet/tabletmanager/tm_state.go#L262
https://github.com/vitessio/vitess/blob/40df731e82daa28cfa5b89024dca12c60958f9e1/go/vt/vttablet/tabletmanager/tm_state.go#L300

Example log below (see more in the log fragments [here](https://gist.github.com/zhongr3n/d233711203846d42efe48b0c94f4faf0
))
```
[2022-03-28 16:57:25.392780] E0328 16:57:25.392079 3564410 tm_state.go:265] SetServingType(serving=false) failed: Tablespace is missing for table `vt_<redacted>`.`_bfd62e12_a055_11ec_a8d6_022a23b59a1c_20220310093754_vrepl`. (errno 1812) (sqlstate HY000) during query: select * from _bfd62e12_a055_11ec_a8d6_022a23b59a1c_20220310093754_vrepl where 1 != 1
```

However, the problem is that the error doesn't cause ChangeTabletType to fail. As a result, the newly restored tablet was never able to serve traffic, but successfully transitioned from SPARE to REPLICA and eventually PRIMARY. When it actually became PRIMARY, all queries against this shard failed with the following error
```
Failed FullyQualifiedShard: cluster:"fourxlarge" keyspace:"<redacted>" shard:"-80" rpc error: code = FailedPrecondition desc = target: <redacted>.-80.primary: vttablet: rpc error: code = FailedPrecondition desc = wrong tablet type: PRIMARY, want: REPLICA or []
```
Error code path https://github.com/vitessio/vitess/blob/40df731e82daa28cfa5b89024dca12c60958f9e1/go/vt/vttablet/tabletserver/state_manager.go#L396

#### Possible solution
Looking at the code, it seems like we swallow any error in `updateLocked` method, which is called at multiple places including ChangeTabletType
https://github.com/vitessio/vitess/blob/40df731e82daa28cfa5b89024dca12c60958f9e1/go/vt/vttablet/tabletmanager/tm_state.go#L246

I think ChangeTabletType should keep the invariant that all the internal states about the tablet type should be updated (ideally synchronously within the call) before returning a success response. Do you think that would make sense?

### Reproduction Steps

N/A

### Binary Version

```sh
N/A
```

```sh
N/A
```

### Log Fragments

https://gist.github.com/zhongr3n/d233711203846d42efe48b0c94f4faf0

rsajwani:

## Description

SetServingType can result in error. Although we have forever retries to fix these transient errors but, under certain conditions these errors are non-transient (see https://github.com/vitessio/vitess/issues/10145).
There is no way to distinguish between retry (transient) and non-retryable errors, therefore we will always return error from 'SetServingType' and 'applyDenyList' to our client. It is up to them to handle it accordingly.
UpdateLock is called from 'ChangeTableType', 'Open' and 'RefreshFromTopoInfo'. For 'Open' and 'RefreshTopoInfo' we don't need to propagate error to client hence no changes there but we will propagate error from 'ChangeTableType' to client.
	
This is a bug. I will discuss with team if it needs a back porting or not. I believe No.

## Related Issue(s)

Fixes #10145

## Checklist

### DIFF
--- a/go/vt/vttablet/tabletmanager/tm_state.go
+++ b/go/vt/vttablet/tabletmanager/tm_state.go
@@ -223,7 +223,7 @@ func (ts *tmState) ChangeTabletType(ctx context.Context, tabletType topodatapb.T
 	statsTabletTypeCount.Add(s, 1)
 
 	err := ts.updateLocked(ctx)
-	// no need to short circuit. apply all steps and return error in the end.
+	// No need to short circuit. Apply all steps and return error in the end.
 	ts.publishStateLocked(ctx)
 	ts.tm.notifyShardSync()
 	return err
@@ -270,7 +270,7 @@ func (ts *tmState) updateLocked(ctx context.Context) error {
 		if err := ts.tm.QueryServiceControl.SetServingType(ts.tablet.Type, terTime, false, reason); err != nil {
 			errStr := fmt.Sprintf("SetServingType(serving=false) failed: %v", err)
 			log.Errorf(errStr)
-			// no need to short circuit. apply all steps and return error in the end.
+			// No need to short circuit. Apply all steps and return error in the end.
 			returnErr = vterrors.Wrapf(err, errStr)
 		}
 	}
@@ -278,7 +278,7 @@ func (ts *tmState) updateLocked(ctx context.Context) error {
 	if err := ts.applyDenyList(ctx); err != nil {
 		errStr := fmt.Sprintf("Cannot update denied tables rule: %v", err)
 		log.Errorf(errStr)
-		// no need to short circuit. apply all steps and return error in the end.
+		// No need to short circuit. Apply all steps and return error in the end.
 		returnErr = vterrors.Wrapf(err, errStr)
 	}
 
--- a/go/vt/vttablet/tabletmanager/tm_state.go
+++ b/go/vt/vttablet/tabletmanager/tm_state.go
@@ -249,15 +249,9 @@ func (ts *tmState) updateLocked(ctx context.Context) error {
 	span, ctx := trace.NewSpan(ctx, "tmState.update")
 	defer span.Finish()
 	ts.publishForDisplay()
-	// SetServingType can result in error. Although we have forever retries to fix these transient errors
-	// but, under certain condition these errors are non-transient (see https://github.com/vitessio/vitess/issues/10145).
-	// There is no way to distinguish between retry (transient) and non-retryable errors, therefore we will
-	// always return error from 'SetServingType' and 'applyDenyList' to our client. It is up to them to handle it accordingly.
-	// UpdateLock is called from 'ChangeTableType', 'Open' and 'RefreshFromTopoInfo'. For 'Open' and 'RefreshTopoInfo' we don't need
-	// to propagate error to client hence no changes there but we will propagate error from 'ChangeTableType' to client.
 	var returnErr error
 	if !ts.isOpen {
-		return fmt.Errorf("TabletManager state is not open")
+		return nil
 	}
 
 	terTime := logutil.ProtoToTime(ts.tablet.PrimaryTermStartTime)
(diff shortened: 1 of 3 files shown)


sougou (approved):

go/vt/vttablet/tabletmanager/tm_state.go
```diff
@@ -243,13 +245,19 @@ func (ts *tmState) UpdateTablet(update func(tablet *topodatapb.Tablet)) {
...
 }
 
-func (ts *tmState) updateLocked(ctx context.Context) {
+func (ts *tmState) updateLocked(ctx context.Context) error {
 	span, ctx := trace.NewSpan(ctx, "tmState.update")
 	defer span.Finish()
 	ts.publishForDisplay()
-
+	// SetServingType can result in error. Although we have forever retries to fix these transient errors
+	// but, under certain condition these errors are non-transient (see https://github.com/vitessio/vitess/issues/10145).
```
deepthi, line 253:
```suggestion
	// but, under certain conditions these errors are non-transient (see https://github.com/vitessio/vitess/issues/10145).
```

go/vt/vttablet/tabletmanager/tm_state.go
```diff
@@ -243,13 +245,19 @@ func (ts *tmState) UpdateTablet(update func(tablet *topodatapb.Tablet)) {
...
+	// There is no way to distinguish between retry (transient) and non-retryable errors, therefore we will
+	// always return error from 'SetServingType' and 'applyDenyList' to our client. It is up to them to handle it accordingly.
+	// UpdateLock is called from 'ChangeTableType', 'Open' and 'RefreshFromTopoInfo'. For 'Open' and 'RefreshTopoInfo' we don't need
```
deepthi, line 256:
```suggestion
	// updateLocked is called from 'ChangeTabletType', 'Open' and 'RefreshFromTopoInfo'. For 'Open' and 'RefreshFromTopoInfo' we don't need
```

go/vt/vttablet/tabletmanager/tm_state.go
```diff
@@ -243,13 +245,19 @@ func (ts *tmState) UpdateTablet(update func(tablet *topodatapb.Tablet)) {
...
+	// to propagate error to client hence no changes there but we will propagate error from 'ChangeTableType' to client.
```
deepthi, line 257:
```suggestion
	// to propagate error to client hence no changes there but we will propagate error from 'ChangeTabletType' to client.
```

go/vt/vttablet/tabletmanager/tm_state.go
```diff
@@ -243,13 +245,19 @@ func (ts *tmState) UpdateTablet(update func(tablet *topodatapb.Tablet)) {
...
+	var returnErr error
 	if !ts.isOpen {
-		return
+		return fmt.Errorf("TabletManager state is not open")
```
deepthi, line 260: This change should be reverted. In the code path where `RefreshFromTopoInfo` is called, `isOpen` is false.
This is most likely what is causing the CI test failures.

go/vt/vttablet/tabletmanager/tm_state.go
```diff
@@ -243,13 +245,19 @@ func (ts *tmState) UpdateTablet(update func(tablet *topodatapb.Tablet)) {
...
 	defer span.Finish()
 	ts.publishForDisplay()
-
+	// SetServingType can result in error. Although we have forever retries to fix these transient errors
+	// but, under certain condition these errors are non-transient (see https://github.com/vitessio/vitess/issues/10145).
+	// There is no way to distinguish between retry (transient) and non-retryable errors, therefore we will
+	// always return error from 'SetServingType' and 'applyDenyList' to our client. It is up to them to handle it accordingly.
+	// UpdateLock is called from 'ChangeTableType', 'Open' and 'RefreshFromTopoInfo'. For 'Open' and 'RefreshTopoInfo' we don't need
+	// to propagate error to client hence no changes there but we will propagate error from 'ChangeTableType' to client.
+	var returnErr error
```
deepthi, line 258: Move this line closer to where we first use this var.

deepthi:
Test failures:
20 - will be fixed by #10194 
onlineddl_vrepl_stress_suite - seems to be a known flaky test #8969

PR #10094, opened 2022-04-14
Update the embedded config data in mysqlctl

mattlord:
## Description

I will need to rebase this on main after https://github.com/vitessio/vitess/pull/10080 is resolved and merged as that also makes some changes to the config files that get embedded. But the bulk of the work in this PR is about getting everything to work once `super_read_only` is actually applied.

## Related Issue(s)
- Follow-up to: https://github.com/vitessio/vitess/pull/9312
- Related to: https://github.com/vitessio/vitess/pull/10080
- Blocker for: https://github.com/vitessio/vitess/pull/10182
- Blocker for: https://github.com/vitessio/vitess/pull/10295

## Checklist

-   [ ] "Backport me!" label has been added if this change should be backported
-   [ ] Tests were added or are not required
-   [ ] Documentation was added or is not required

PR #10182, opened 2022-04-30
Unit test to confirm that the mysqlctl rice box is current

rohit-nayak-ps:
## Description

We recently had an issue where the `init_db.sql` has been updated but the corresponding rice box was not updated.  (rice box reference: https://github.com/GeertJohan/go.rice)

All files in the config directory are part of this rice box. This PR adds a unit test to compare these files on disk with that embedded in the rice box.

The other rice-box usage is around the old vtctld UI and embedded vitess binaries, those are not tested for.

### DIFF
--- a/go/vt/mysqlctl/rice_box_test.go
+++ b/go/vt/mysqlctl/rice_box_test.go
@@ -19,6 +19,9 @@ package mysqlctl
 import (
 	"fmt"
 	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
 	"testing"
 
 	rice "github.com/GeertJohan/go.rice"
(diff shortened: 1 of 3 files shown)


mattlord:
Related to this: https://github.com/vitessio/vitess/pull/10094

mattlord (approved):
Nice! I like it ❤️

go/vt/mysqlctl/rice_box_test.go
```diff
@@ -0,0 +1,64 @@
...
+	"path/filepath"
+	"strings"
+	"testing"
+
+	rice "github.com/GeertJohan/go.rice"
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	mysqlCtlRiceBoxLocation string = "../../../config"
```
deepthi, line 32: nit: should be `mySqlCtl` or `mysqlctl`. I prefer the second option.
EDIT: Or even `mySQLCtl`?? 🤷

deepthi (approved):

PR #10164, opened 2022-04-28
schemadiff: Schema.Entity(), ToStatements(), ToQueries(), ToSQL()

shlomi-noach:
## Description

Usability functions for `Schema`: export to SQL, to statements ; get entity by name

## Checklist

-   [ ] "Backport me!" label has been added if this change should be backported
-   [x] Tests were added or are not required
-   [ ] Documentation was added or is not required

cc @dbussink

PR #10183, opened 2022-05-01
schemadiff: logically Apply() a diff to a CreateTable, CreateView or to a Schema

shlomi-noach:
## Description

In this PR `schemadiff` can `Apply` a change onto a declarative table, view or schema, via these new public functions:
```golang
func (c *CreateTableEntity) Apply(diff EntityDiff) (Entity, error)
func (c *CreateViewEntity) Apply(diff EntityDiff) (Entity, error)
func (s *Schema) Apply(diffs []EntityDiff) (*Schema, error)
```

as can be seen in the above, the things we can apply are:

- An `ALTER TABLE` onto a `CREATE TABLE` entity
- An `ALTER VIEW` onto a `CREATE VIEW` entity
- A list of diffs on a `Schema` object. These diffs are `CREATE/DROP/ALTER TABLE/SCHEMA` statements

In each of the above functions:

- we return an `error` if the operation is invalid (e.g. dropping a column when the column does not exist; adding a column when it already exists; etc)
- we do not modify the receiver; instead, is successful, a new object is returns
- we only support diffs that are generated by `Diff()`

So the idea is to be able to run purely-declarative schema changes, including ALTER TABLE statements.

The tests are trivially integrated into pre-existing tests: where we already validate a diff between `object1` and `object2`, now attempt to apply the diff to `object1` and expect the result to be identical to `object2`.

## Related Issue(s)

This PR extends https://github.com/vitessio/vitess/pull/10164 as it relies on some of its functionality

## Checklist

cc @dbussink


shlomi-noach:
Working on final test failures. Moving to draft in the meantime.

shlomi-noach:
OK all good. I accidentally made a regression and pushed it before testing. Now fixed.

go/vt/schemadiff/schema.go
```diff
@@ -134,6 +135,8 @@ func getViewDependentTableNames(createView *sqlparser.CreateView) (names []strin
 // normalize is called as part of Schema creation process. The user may only get a hold of normalized schema.
 // It validates some cross-entity constraints, and orders entity based on dependencies (e.g. tables, views that read from tables, 2nd level views, etc.)
 func (s *Schema) normalize() error {
+	s.named = map[string]Entity{}
+	s.sorted = []Entity{}
```
dbussink, line 139: Are these added for a specific reason?

go/vt/schemadiff/schema.go
```diff
@@ -282,3 +285,150 @@ func (s *Schema) Diff(other *Schema, hints *DiffHints) (diffs []EntityDiff, err
...
+				if t.Table.Name.String() == diff.from.Table.Name.String() {
+					to, err := t.Apply(diff)
+					if err != nil {
+						return err
+					}
+					toCreateTableEntity, ok := to.(*CreateTableEntity)
+					if !ok {
+						return ErrEntityTypeMismatch
+					}
+					s.tables[i] = toCreateTableEntity
```
dbussink, line 380: Does `s.named` need to be updated here?

go/vt/schemadiff/schema.go
```diff
@@ -282,3 +285,150 @@ func (s *Schema) Diff(other *Schema, hints *DiffHints) (diffs []EntityDiff, err
...
+				if v.ViewName.Name.String() == diff.from.ViewName.Name.String() {
+					to, err := v.Apply(diff)
+					if err != nil {
+						return err
+					}
+					toCreateViewEntity, ok := to.(*CreateViewEntity)
+					if !ok {
+						return ErrEntityTypeMismatch
+					}
+					s.views[i] = toCreateViewEntity
```
dbussink, line 401: Same here, does `s.named` need to be updated?

go/vt/schemadiff/table.go
```diff
@@ -275,6 +277,46 @@ func (c *CreateTableEntity) diffTableCharset(
 	return ""
 }
 
+// isDefaultTableOptionValue sees if the value for a TableOption is also its default value
+func isDefaultTableOptionValue(option *sqlparser.TableOption) bool {
```
dbussink, line 281: Not for this PR, but should there be something similar for what the defaults are for column level options?

go/vt/schemadiff/table.go
```diff
@@ -799,3 +841,228 @@ func (c *CreateTableEntity) Drop() EntityDiff {
...
+						newCols = append(newCols, col)
+						newCols = append(newCols, c.TableSpec.Columns[a+1:colIndex]...)
+						newCols = append(newCols, c.TableSpec.Columns[colIndex+1:]...)
+					}
+					afterColFound = true
+					break
+				}
+			}
+			if !afterColFound {
+				return errors.Wrapf(ErrApplyColumnNotFound, "%v", after.Name.String())
```
dbussink, line 894: Can this be simplified to `errors.Wrap(ErrApplyColumnNotFound, after.Name.String())`? And for all the other further cases down in this file?

go/vt/schemadiff/table.go
```diff
@@ -799,3 +841,228 @@ func (c *CreateTableEntity) Drop() EntityDiff {
...
+			}
+		default:
+			// no change in positiion
```
dbussink, line 897: Double `ii`

go/vt/schemadiff/table.go
```diff
@@ -799,3 +841,228 @@ func (c *CreateTableEntity) Drop() EntityDiff {
...
+// apply attempts to apply an ALTER TABLE diff onto this entity's table definition.
+// supported modifications are only those created by schemadiff's Diff() function.
+func (c *CreateTableEntity) apply(diff *AlterTableEntityDiff) error {
+	if spec := diff.alterTable.PartitionSpec; spec != nil {
+		switch {
+		case spec.Action == sqlparser.RemoveAction && spec.IsAll:
+			// Remove partitioning
+			c.TableSpec.PartitionOption = nil
+		default:
+			return errors.Wrapf(ErrUnsupportedApplyOperation, "%v", sqlparser.String(spec))
```
dbussink, line 854: Can this be `errors.Wrap(ErrUnsupportedApplyOperation, sqlparser.String(spec))`?

go/vt/schemadiff/table.go
shlomi-noach: fixed

go/vt/schemadiff/table.go
shlomi-noach: done

go/vt/schemadiff/table.go
shlomi-noach: done

go/vt/schemadiff/schema.go
shlomi-noach: They are. At the end of `apply()` we call normalize, and we need to make sure that existing `named` values are erased. `sorted` happens to be unpopulated; but `normalize` really expects both to be empty. 
We could clear both in `apply`, as right now this is the only path where the values get populated before `normalize`. But I feel `normalize` is the safer place to do that.

go/vt/schemadiff/schema.go
shlomi-noach: It does not, because it is implied that the view name already exists in `named`, and an update would just re-apply same name.

go/vt/schemadiff/schema.go
shlomi-noach: same as above, there is no need.

go/vt/schemadiff/table.go
shlomi-noach: for our discussion later on.

go/vt/schemadiff/schema.go
dbussink: @shlomi-noach I'm probably missing something, but doesn't then calling for example `Entity(name)` return the old entity without the change applied? Or is this updated then somewhere else?

The reason I was asking is because this does get updated on create / drop so that's what caught my eye.

shlomi-noach: @dbussink gotcha. `named` is only used in _this_ function to determine existence/nonexistence of an entity. At the end of this function, we call `normalize()` which completely rebuilds `named`. We could update the entity value here, but it then gets deleted and recreated in `normalize` anyhow.

dbussink: @shlomi-noach Ah ok, if that's the case I think there's an inconsistency though in the add / drop logic. 

When a create is done, we don't add it to `named` so if the `diffs` slice has multiple creates for the same table, it would not error but have a last one wins.

On the other hand, if `diffs` contains multiple drops for the same table, it would error on the second one since it does delete from `named`. 

I have no super strong opinion on what the behavior should be, but I think it should be consistent between create and drop?

shlomi-noach: You're right, there is an inconsistency. Fixing.

shlomi-noach: Please see updates to `Schema.apply()` and consistent use of `s.named`

dbussink (approved):

shlomi-noach:

PR #10184, opened 2022-05-01
Using LCS to achieve shortest column reordering list

shlomi-noach:
## Description

As requested by @dbussink , switching column reordering algorithm to use LCS (longest common subsequence), as used in various diffs including `git diff`. By using LCS we end up with a minimal number of column reordering steps.

The tests are updated to reflect the impact of the change.

## Checklist


go/vt/schemadiff/table.go
```diff
@@ -630,44 +632,37 @@ func (c *CreateTableEntity) diffKeys(alterTable *sqlparser.AlterTable,
...
 	for i, t2Col := range t2SharedColumns {
-		// for each position in the target list of columns, we check whether we need to
-		// reorder a column.
-		// t2ColName is the desired column name at this position:
 		t2ColName := t2Col.Name.String()
-		if t2ColName == buf[0].Name.String() {
-			// no need to reorder this column. continue
-			buf = buf[1:]
-			continue
+		// see if this column is in longest common subsequence. If so, need to reorder it. If not, it must be reordered.
```
dbussink, line 660: I think the comment is wrong here, "if so, no need to reorder it"?

go/vt/schemadiff/table.go
```diff
@@ -630,44 +632,37 @@ func (c *CreateTableEntity) diffKeys(alterTable *sqlparser.AlterTable,
...
+		if _, ok := lcsNames[t2ColName]; !ok {
+			minimalColumnReordering[t2ColName] = i
```
dbussink, line 662: Does this index computation work properly if there are also columns added or deleted? And if so, would it be good to ensure there's test cases for that situation as well?

shlomi-noach: This function is only fed with the list of shared columns. Handling of removed or added columns takes place outside this function.

Also, we have unit tests to cover the scenarios of reordering with adding or with removal of columns:

https://github.com/vitessio/vitess/pull/10184/files#diff-2de8e8fe0a6b562dc9d4566b3de63c1718f3d3df4a938f9ce428ca22a7ebfc66R142-R155

go/vt/schemadiff/table.go
shlomi-noach: good catch. Comment fixed.

dbussink (approved):

PR #10185, opened 2022-05-02
Add failing diff case for index reordering

dbussink:
Add a new test case for index reordering that is currently failing.

This shouldn't generate a diff, but it does but the `ALTER` statement itself is also invalid:

```
alter table pets drop key, add KEY (login)
```

## Checklist


shlomi-noach:
The failure is caused because the indexes do not have names. Technically, an output of `SHOW CREATE TABLE` will never look like this. However, this is a good case for normalization, and we can handle it using similar logic to MySQL (it attempts to create an index named after the covered columns, or start enumerating if the name is occupied).

shlomi-noach:
Test incorporated into, and solved, in https://github.com/vitessio/vitess/pull/10188

dbussink:
Closing this as our other work has resolved this.

PR #10188, opened 2022-05-02
schemadiff: normalize() table. Set names to all keys

shlomi-noach:
`NewCreateTableEntity` now normalizes a table upon creation, modifying its `CreateStatement` into a canonical form.

This PR is a first in several iterations. This PR ensures all keys are named. For example, in the following statement:

```sql
create table t2 (id int primary key, i int, key (i))
```
the key over `i` is not named.
`normalize()` will set a name to the key. The name is chosen in similar way to MySQL's algorithm:

- Try the name of first column covered by the key, `i` in our example
- If that name is taken, try `i_2`
- If that name is taken, try `i_3`
- etc.

unit tests added to both validate and explain the logic.

Note that normalization edits the `CreateTable` statement in-place.

### DIFF
--- a/go/vt/schemadiff/table.go
+++ b/go/vt/schemadiff/table.go
@@ -17,6 +17,7 @@ limitations under the License.
 package schemadiff
 
 import (
+	"fmt"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


dbussink (approved):

shlomi-noach:

PR #10189, opened 2022-05-02
schemadiff: validate() table structure at the end of apply()

shlomi-noach:
## Description

First PR of several. This PR kicks schema validation for a `CreateTableEntity`. Some stuff we can validate delaratively; this PR validates that keys-to-columns relationship are valid. That is, that all columns referenced by keys do actually exist.


go/vt/schemadiff/table_test.go
```diff
@@ -705,3 +706,104 @@ func TestCreateTableDiff(t *testing.T) {
...
+			name:  "drop column, ok",
+			from:  "create table t (id int primary key, i int, i2 int, key i_idx(i))",
+			alter: "alter table t drop column i2",
+		},
+		{
+			name:      "drop column, break key",
+			from:      "create table t (id int primary key, i int, key i_idx(i))",
+			alter:     "alter table t drop column i",
+			expectErr: ErrInvalidColumnInKey,
+		},
```
dbussink, line 754: Right now this makes sense as a safeguard until we improve the drop column logic to also drop references from the index to mimic MySQL behavior.

dbussink (approved):

go/vt/schemadiff/table_test.go
shlomi-noach: If we choose to go that path then I'll update this PR; no need to postpone that to a future PR.

shlomi-noach:

PR #10190, opened 2022-05-02
schemadiff: schema is immutable through Apply()

shlomi-noach:
## Description

`Schema.Apply()` returns a new, likely modified, schema. But it must not affect the receiver schema.
This PR:

- Uses query export/import to ensure the two schemas do not share any pointers/maps/slices
- Adds tests to validate schema is unaffected by `Apply()`

## Checklist


dbussink (approved):

shlomi-noach:

PR #10191, opened 2022-05-02
sqlparser: Handle case sensitive AST option for table options

dbussink:
In case of a `CREATE TABLE` statement, the options are rendered directly and the TableOptions logic is not used. This means it doesn't handle character set or collation options like an `ALTER TABLE` does.

This updates the logic to be consistent with column level settings and with `ALTER TABLE`.

## Related Issue(s)

Part of extension work after https://github.com/vitessio/vitess/pull/9719

## Checklist


vmg (approved):
Easy! Thanks for fixing this!

Issue #7120, opened 2020-12-04
The queries generated by MoveTables are incorrect (slow) when the target Tables don't have PRIMARY KEYS 


hmenghouzz:
#### Overview of the Issue

For example, there is a table with NO Primary Key (but has one UNIQUE key):

```sql
CREATE TABLE `demo_review` (
  `vendor_id` int(10) unsigned NOT NULL,
  `old` decimal(19,4) DEFAULT NULL,
  `new` decimal(19,4) DEFAULT NULL,
  `status` int(11) DEFAULT '0',
  `reason` varchar(500) DEFAULT NULL,
  `details` varchar(2000) DEFAULT NULL,
  `changed_by` varchar(80) NOT NULL,
  `created` datetime NOT NULL,
  `modified` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
  `strategy` int(5) unsigned DEFAULT NULL,
  `data_source` varchar(20) DEFAULT NULL,
  UNIQUE KEY `vendor_id` (`vendor_id`),
  KEY `modified` (`modified`),
  KEY `vendor_id_status` (`vendor_id`,`status`),
  KEY `status` (`status`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8;
```
Then vitess MoveTable will generate this query to fetch the data:
```SQL
select vendor_id, old, new, `status`, reason, details, changed_by, created, modified, strategy, data_source 
from demo_review order by vendor_id, old, new, `status`, reason 
```
From mysql doc 
**If a table contains no primary key, InnoDB automatically promotes the first UNIQUE NOT NULL index to the primary key. This is not supported for UNIQUE NOT NULL indexes that have functional key parts**

So I guess the correct MoveTable query should be
```sql
select vendor_id, old, new, `status`, reason, details, changed_by, created, modified, strategy, data_source 
from demo_review order by vendor_id
```
As you know, order by several columns (not the keys) will take a long long time

#### Operating system and Environment details
vitess running in k8s.
vtctld: vitess/lite:v8.0.0
vtgate: vitess/lite:v8.0.0
vttablet: vitess/lite:v8.0.0

shlomi-noach:
I lean towards requiring existence of `PRIMARY KEY` in the first place. Vitess is opinionated by design, and I think it's against good practices to have tables without `PRIMARY KEY`, so Vitess may just as well enforce a good practice.

mattlord:
[...]

Issue #10134, opened 2022-04-24
RFC: VDiff2: Reimplementing VDiff on tablets


rohit-nayak-ps:
# VDiff2: VDiff on tablets

## TL;DR;

VDiff currently uses vtctld. For large tables, VDiff is very slow; cannot be restarted on an error; and can impact app read/write performance. In addition,  vtctld might need to be differently provisioned depending on the size of tables being VDiffed. 

To address these, VDiff 2.0 will run on tablets (one per shard), to address performance and provisioning, and be resumable. 

### Current VDiff Flow

VDiff compares exactly the source and target data at a point in time for a given VReplication workflow. This is a high-level view of the flow.

* Stop target streams
* Start snapshots on the source primaries, record gtids of snapshots
* Start target streams, stopping at the recorded gtids
* Start snapshots on the target primaries
* Stream data from source and target; 
    * Merge-sort the data from different shards as they arrive
    * Make sure every shard has responded
    * Compare the top row. 
    * Rinse and repeat

Note that, since we continue to have writes on the source after the snapshot is taken, the source and target databases *can* diverge. VDiff only guarantees that the data is the same at the time of the snapshot.

#### Single Consistent Snapshot

VDiff uses consistent snapshots and compares the entire table using that snapshot. So it needs to hold that snapshot for the entire duration of the comparison. If there is a network issue then we cannot resume the operation. A fresh snapshot has to be taken and we start VDiff from scratch.

In addition holding the snapshot for so long can impact MVCC (increased history list length) on primaries and replica lags on replicas.

#### Performance

Since a single process is doing the entire comparison, it takes very long. VDiffs have been known to run for several hours or days in some cases. Of course, every time there is a network issue or if a server is restarted/reparented we need to restart it, increasing the perceived latency.

#### Uses vtctld

It runs, not on tablets, but on vtctld. vtctld is generally used as a lightweight wrapper to send commands to vttablets or vtgates. VDiff breaks that abstraction. For large tables it can take a lot of time (hours and days) and require a large amount of memory. This means that, if you want to use vtctld for VDiff the VM needs to be specially provisioned. 

#### Increased network traffic

Also the entire source and target database is pulled by VDiff causing a potentially massive increase in network traffic between tablets and vtctld.

## Proposed Changes

VDiff2 will run on tablets, similar to how VReplication workflows run today. There will be one stream on the primary of each target shard for each VDiff. Each stream will compare the data copied to that target shard from all the source shards that map to it.

[...]

PR #10192, opened 2022-05-02
VReplication: use PK equivalent (PKE) if no PK found

mattlord:
## Description

This allows VReplication to be more MySQL compatible as MySQL will automatically try to use a PRIMARY KEY equivalent (PKE) when the ability to uniquely identify rows for internal operations is needed, if no defined PRIMARY KEY exists. It also ensures that the query used on the source and target tablets is more efficient when there is no explicitly defined PRIMARY KEY on the table.

ℹ️ **NOTE:** This modifies/improves the query used on tablets. We can make the same basic improvement for the VDiff query used on source and target, but I'm punting on that for now as [we are in the process of moving VDiff work to tablets](https://github.com/vitessio/vitess/issues/10134) (currently it's done in `vtctl`) which would then allow us to improve the query in the same way this PR does (w/o having to add RPCs for getting the PKEs and use those from `vtctl` -> `vttablet` -> `mysqld`).

## Manual Test

This cut and paste test shows that the VReplication query (`ORDER BY`) matches [what we see in the new unit test:](https://github.com/vitessio/vitess/pull/10192/files#diff-f61d8ccd19060e84eaa9ab08e07c174de6bd8ad667ad1f0b0f0a2397db9d8292R102-R153)

```
$ make build && pushd ./examples/local; ./101_initial_cluster.sh; ./201_customer_tablets.sh; popd; \
mysql -u root --socket=/vt/vtdataroot/vt_0000000100/mysql.sock -e 'set global general_log=ON; use vt_commerce; CREATE TABLE onepke_t (id INT NOT NULL, col1 VARCHAR(25), UNIQUE KEY (id)); CREATE TABLE onemcpke_t (col1 VARCHAR(25) NOT NULL, col2 VARCHAR(25) NOT NULL, col3 VARCHAR(25) NOT NULL, col4 VARCHAR(25), UNIQUE KEY c4_c2_c1 (col4, col2, col1), UNIQUE KEY c1_c2 (col1, col2), UNIQUE KEY c1_c2_c4 (col1, col2, col4), KEY nc1_nc2 (col1, col2)); CREATE TABLE twomcpke_t (col1 VARCHAR(25) NOT NULL, col2 VARCHAR(25) NOT NULL, col3 VARCHAR(25) NOT NULL, col4 VARCHAR(25), UNIQUE KEY (col4), UNIQUE KEY c4_c2_c1 (col4, col2, col1), UNIQUE KEY c1_c2_c3 (col1, col2, col3), UNIQUE KEY c1_c2 (col1, col2)); CREATE TABLE oneintpke1charpke_t (col1 VARCHAR(25) NOT NULL, col2 VARCHAR(25) NOT NULL, col3 VARCHAR(25) NOT NULL, id1 INT NOT NULL, id2 INT NOT NULL, UNIQUE KEY c1_c2 (col1, col2), UNIQUE KEY id1_id2 (id1, id2)); CREATE TABLE twointvsvcharpke_t (col1 VARCHAR(25) NOT NULL, id1 INT NOT NULL, id2 INT NOT NULL, UNIQUE KEY c1 (col1), UNIQUE KEY id1_id2 (id1, id2)); CREATE TABLE tinyintvsbigint_t (tid1 TINYINT NOT NULL, id1 INT NOT NULL, UNIQUE KEY tid1 (tid1), UNIQUE KEY id1 (id1)); CREATE TABLE vcharintvsinttwovchar_t (id1 INT NOT NULL, col1 VARCHAR(25) NOT NULL, col2 VARCHAR(25) NOT NULL, UNIQUE KEY col1_id1 (col1, id1), UNIQUE KEY id1_col1_col2 (id1, col1, col2)); CREATE TABLE vcharvsintthree_t (id1 INT NOT NULL, id2 INT NOT NULL, id3 INT NOT NULL, col1 VARCHAR(50) NOT NULL, UNIQUE KEY col1 (col1), UNIQUE KEY id1_id2_id3 (id1, id2, id3))' ; \
vtctlclient --server=localhost:15999 MoveTables -- --source commerce --all --tablet_types=primary Create customer.commerce2customer ; \
sleep 10 && grep "_t order by" /vt/vtdataroot/vt_0000000100/data/ip-172-31-41-160.log
```

Results being:
```
2022-05-09T14:56:24.473476Z	   87 Query	select col1, col2, col3, id1, id2 from oneintpke1charpke_t order by id1, id2
2022-05-09T14:56:25.553185Z	   95 Query	select col1, col2, col3, col4 from onemcpke_t order by col1, col2
2022-05-09T14:56:26.590831Z	  103 Query	select id, col1 from onepke_t order by id
2022-05-09T14:56:28.652515Z	  118 Query	select tid1, id1 from tinyintvsbigint_t order by tid1
2022-05-09T14:56:29.685208Z	  126 Query	select col1, id1, id2 from twointvsvcharpke_t order by id1, id2
2022-05-09T14:56:30.734427Z	  134 Query	select col1, col2, col3, col4 from twomcpke_t order by col1, col2
2022-05-09T14:56:31.766876Z	  142 Query	select id1, col1, col2 from vcharintvsinttwovchar_t order by col1, id1
2022-05-09T14:56:32.803098Z	  150 Query	select id1, id2, id3, col1 from vcharvsintthree_t order by id1, id2, id3
```

## Related Issue(s)
  - Fixes: https://github.com/vitessio/vitess/issues/7120
  
## Checklist

-   [x] Documentation: https://github.com/vitessio/website/pull/1021


rohit-nayak-ps:
I am on `8.0.21`. I don't see `UNI` as the column key for a composite unique key, but `MUL`

```
CREATE TABLE `t1` (
  `id1` int DEFAULT NULL,
  `id2` int DEFAULT NULL,
  `id3` int DEFAULT NULL,
  `val1` varchar(10) DEFAULT NULL,
  `val2` varchar(20) DEFAULT NULL,
  UNIQUE KEY `x1` (`id1`),
  UNIQUE KEY `x23` (`id2`,`id3`)
)

  SELECT column_name, column_key FROM INFORMATION_SCHEMA.COLUMNS WHERE TABLE_SCHEMA = 'test' AND TABLE_NAME IN ('t1')  ORDER BY t
able_name, ordinal_position;
+-------------+------------+
| COLUMN_NAME | COLUMN_KEY |
+-------------+------------+
| id1         | UNI        |
| id2         | MUL        |
| id3         |            |
| val1        |            |
| val2        |            |
+-------------+------------+
5 rows in set (0.00 sec)


SELECT table_name as table_name, ordinal_position as ordinal_position, COLUMN_NAME as column_name FROM INFORMATION_SCHEMA.COLUMNS WHERE TABLE_SCHEMA = 'test' AND TABLE_NAME IN ('t1') AND COLUMN_KEY = 'PRI' ORDER BY table_name, ordinal_position;
Empty set (0.01 sec)
```

rohit-nayak-ps:
There might be easier ways to find unique key columns, but found this one that works for me.
```
select stat.table_schema as database_name,        stat.table_name,        stat.index_name,        group_concat(stat.column_name
          order by stat.seq_in_index separator ', ') as columns,        tco.constraint_type from information_schema.statistics stat join information_schema.table_constraints tco      on stat.table_schema = tco.table_schema      and stat.table_name = tco.table_name      and stat.index_name = tco.constraint_name where stat.non_unique = 0       and stat.table_schema not in ('information_schema', 'sys',                                     'performance_schema', 'mysql') and stat.table_name = 't1a' group by stat.table_schema,          stat.table_name,          stat.index_name,          tco.constraint_type order by stat.table_schema,          stat.table_name;


+---------------+------------+------------+----------+-----------------+
| database_name | TABLE_NAME | INDEX_NAME | columns  | CONSTRAINT_TYPE |
+---------------+------------+------------+----------+-----------------+
| test          | t1a        | v1         | val1     | UNIQUE          |
| test          | t1a        | x1         | id1      | UNIQUE          |
| test          | t1a        | x23        | id2, id3 | UNIQUE          |
+---------------+------------+------------+----------+-----------------+
3 rows in set (0,01 sec)

```

mattlord:
@rohit-nayak-ps yeah, my initial quick tests were done on a table that already had a primary key. That `columns` table is not useful for anything but the PK columns. 

Good minds think alike! 😄 I am just now seeing your latest comment. I also settled on the `statistics` table as that uses the same underlying data that `SHOW INDEXES` uses (which was my initial idea, but filtering is harder). I've moved to that query here: https://github.com/vitessio/vitess/pull/10192/commits/415a5a2d8e6f0b2b1e2f4558b43cf54c4e542987

Thanks!

rohit-nayak-ps:
While tweaking an existing e2e test to test the new functionality. I modified the customer table to use a unique key instead of the existing primary key, but found that GetPrimaryKeyEquivalentColumns() never got called. This is because, as noted previously by Matt, the unique key shows up as a PRI in `information_schema.columns.column_key`. 

For mysql that behaves as above, returning the unique key as a PK, it looks like the original bug report https://github.com/vitessio/vitess/issues/7120 (which was for v8) was already fixed by https://github.com/vitessio/vitess/pull/9794 in v13. 

I guess the change in this PR should still help since the behaviour of mysql flagging the unique key as a primary key seems inconsistent.

go/vt/mysqlctl/schema.go
```diff
@@ -533,6 +533,52 @@ func (mysqld *Mysqld) ApplySchemaChange(ctx context.Context, dbName string, chan
 	return &tabletmanagerdatapb.SchemaChangeResult{BeforeSchema: beforeSchema, AfterSchema: afterSchema}, nil
 }
 
+// getPrimaryKeyEquivalentColumns can be used if the table has
```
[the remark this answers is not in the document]
mattlord: Good catch! 🙂

go/vt/mysqlctl/schema.go
```diff
@@ -533,6 +533,52 @@ func (mysqld *Mysqld) ApplySchemaChange(ctx context.Context, dbName string, chan
...
+// no defined PRIMARY KEY. It will return the columns in a
+// viable PRIMARY KEY equivalent (PKE) -- a NON-NULL UNIQUE
+// KEY -- in the specified table. When multiple PKE indexes
+// are available it will use the one with the fewest columns
+// in it.
+// If this function is used on a table that DOES have a
```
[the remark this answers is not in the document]
mattlord: Correct. Was just noting that for anyone that may use it in the future.

mattlord:
> The description says that the source query used by rowstreamer will be affected by this PR. However this PR is only changing the logic on the target and not source. The rowstreamer query is generated on the source where pk columns are determined using the table's schema in the tabletserver cache (using `buildPKColumns()`). What am I missing?

I was looking here: https://github.com/planetscale/vitess/blob/46cb4679c198c96fbe7b51f40219d8196f4284a7/go/vt/vttablet/tabletmanager/vreplication/vreplicator.go#L165-L185

But the target tablet does coordinate... and that function does go on to start a vcopier: https://github.com/planetscale/vitess/blob/46cb4679c198c96fbe7b51f40219d8196f4284a7/go/vt/vttablet/tabletmanager/vreplication/vreplicator.go#L220

Thanks! ❤️ I was so focused on the functionality to get/provide the PKEs that I didn't revisit this part. I will update the `vstreamer.RowStreamer` in the same way.

shlomi-noach:
If you're looking for a query that finds unique key columns, and moreover picks the "best" unique key available, then we already employ such a query, borrowed from `gh-ost`: 

https://github.com/vitessio/vitess/blob/4dd5a11b275acbab1e9ee06638cb5b663834eb23/go/vt/vttablet/onlineddl/schema.go#L438-L494

go/vt/mysqlctl/schema.go
```diff
@@ -533,6 +533,52 @@ func (mysqld *Mysqld) ApplySchemaChange(ctx context.Context, dbName string, chan
...
+// GetPrimaryKeyEquivalentColumns can be used if the table has
+// no defined PRIMARY KEY. It will return the columns in a
+// viable PRIMARY KEY equivalent (PKE) -- a NON-NULL UNIQUE
+// KEY -- in the specified table. When multiple PKE indexes
+// are available it will use the one with the fewest columns
+// in it.
+// If this function is used on a table that DOES have a
+// defined PRIMARY KEY then it may return the columns for
+// that index if it has the fewest columns in it among the
+// available PKE indexes on the table.
```
shlomi-noach, line 545: Can we also optimize for `INT` columns over textual columns?

mattlord: Nice! Thank you for pointing out that query. I had not seen it but I did just read about that behavior here a few days ago: https://vitess.io/docs/14.0/design-docs/vreplication/keys/

Any reason _NOT_ to re-use the same query that's being used for OnlineDDL? I like that added optimization of favoring INTEGER types in ASC capacity/size over CHAR/BYTE based types. That would also make the documentation relatively simple and more unified: *generalize the noted docs page above a bit more as it would apply to VReplication more generally*.

I'm curious if we would still require these fields and the code around them for OnlineDDL:
```
    source_unique_key_columns:"order_id,customer_id" 
    target_unique_key_columns:"order_id,cust_id" 
    source_unique_key_target_columns:"order_id,cust_id"
```

I'm guessing there's still some cases for it but wanted to ask. 

Thank you for looking at this and sharing the info! ❤️

shlomi-noach: The query is still heuristic. It only inspects the data type of the 1st column, and does not take into consideration the "average" data type for all columns.
Obviously we want to only use non-nullable keys (the query includes nullable keys, but clearly indicates them so it's easy to filter those.

@mattlord I don't know if you've seen [this](https://github.com/vitessio/vitess/pull/8364) PR to support non-PK keys in `vreplication`. This was a year ago and I need a refresher, but basically we can write the names of an "overriding" unqiue key columns into the vreplication's `Rule`:

https://github.com/vitessio/vitess/pull/8364/files#diff-5e69c397e15dcfda72bb6fe8dba93b2d11a8d2d7b594ebfbbdcfdc07bfc3791d

mattlord:
@rohit-nayak-ps and @shlomi-noach based on your helpful input, I did two things here: https://github.com/vitessio/vitess/pull/10192/commits/99f3c6e0ff638cfa9cb793ed79002e5f15edf027

1. Modified the `Mysqld.getPrimaryKeyColumns()` function so that it ONLY returns the _actual PK_ columns. MySQL will (at least in all my tests) choose any PKE and add that PRI flag for it in i_s.columns. The problems with that are 1) I haven't seen that behavior clearly documented, so we should not relay on it 2) It doesn't select the best PKE available 3) It's non-deterministic. Now this functions only returns the _actual PK_, and we will then use the new `Mysqld.GetPrimaryKeyEquivalentColumns()` function when needed (before this, I could never actually get it to be used because MySQL was choosing a PKE on its own to expose via the old query), to use the best available PKE for each table.
2. Take the column types into account in `Mysqld.GetPrimaryKeyEquivalentColumns()` so that we consider not just the column type but the total index size and ordering cost
    - This then matches the docs and what OnlineDDL does: https://vitess.io/docs/14.0/design-docs/vreplication/keys/

I will fix any tests this breaks, then update the `Mysqld.GetPrimaryKeyEquivalentColumns()` comments to align with the new behavior. 

Lastly I will add a slew of additional test cases to the `Mysqld.GetPrimaryKeyEquivalentColumns()` unit test to confirm that we always choose the best PKE available. e.g. an (int, int) index over an (int, varchar) index.

THANK YOU both for the great feedback!

go/vt/mysqlctl/schema.go
mattlord: Yeah, I think it's worth leaving that in as having an override can certainly be useful. 

I've taken the index column types into account now in this query. 🙇

go/mysql/schema.go
```diff
@@ -32,11 +32,11 @@ import (
 
 const (
 	// BaseShowPrimary is the base query for fetching primary key info.
```
rohit-nayak-ps, line 34: We are breaking compatibility with MySQL here, right? It might be a breaking change for consumers of the schema's PKColumns (we use the same query to compute that), since some tables which advertised a PK will no longer to so. 

 At least VTExplain and schema tracker uses this, not sure if any other non-vreplication modules do.

go/vt/mysqlctl/schema.go
```diff
@@ -533,6 +532,79 @@ func (mysqld *Mysqld) ApplySchemaChange(ctx context.Context, dbName string, chan
...
+                                                WHEN 'datetime' THEN 14
+                                                WHEN 'binary' THEN 15
+                                                WHEN 'char' THEN 16
+                                                WHEN 'varbinary' THEN 17
+                                                WHEN 'varchar' THEN 18
+                                                WHEN 'tinyblob' THEN 19
+                                                WHEN 'tinytext' THEN 20
+                                                ELSE 1000
+                                              END
+                                            ) AS type_cost, COUNT(stats.COLUMN_NAME) AS col_count FROM information_schema.STATISTICS AS stats INNER JOIN
```
rohit-nayak-ps, line 583: Is the method of computing type cost something we/others use, or is this our heuristic? Using this logic, 2 int columns are better than one varchar column, but not 3 int columns, for example.

mattlord: It's our heuristic. It's an improved version of the one already used for OnlineDDL — where we only take the first column type into account and favor integer types over char types.

go/mysql/schema.go
mattlord: And keep in mind https://github.com/vitessio/vitess/pull/9794 is only in v14.0-SNAPSHOT, and THAT is what actually changed the behavior here (that PR is what _unintentionally_ addressed https://github.com/vitessio/vitess/issues/7120). My PR takes it back to only providing the actual defined PKs.

go/vt/mysqlctl/schema.go
mattlord: I just tweaked it a bit. Thanks for pointing this case out as it got me thinking... 🙂

go/mysql/schema.go
rohit-nayak-ps: Found some context on the selection of unique key as the primary key: https://bugs.mysql.com/bug.php?id=11228

As to which unique key is selected is still ambiguous:

A comment in this issue suggests it is the first unique key in the create ddl:
"The intention with the original code was to show that MySQL treats the first given unique key as a primary key. Clients can use the marked primary key as a real primary key to validate row changes in case of conflicting updates.  The ODBC driver (and other drivers) may also use this fact to optimize/check updates."

BUT, in the mysql docs it just says "A UNIQUE index may be displayed as PRI if it cannot contain NULL values and there is no PRIMARY KEY in the table".

My guess is that they needed to select a key for a partial binlog_row_image, since the partial row, in addition to the changed columns, needs to contain the columns to uniquely identify a row.

Anyway, we don't need to worry about this for Vitess, just noting it here for completeness.

rohit-nayak-ps (approved):
lgtm

go/mysql/schema.go
shlomi-noach: IMHO, and for predictability, we should always rely on our function rather than the PKE MySQL suggests (which is news to me, I wasn't aware it can do that).

- Firstly, because it looks like MySQL isn't deterministic about the key it chooses (consider we can have two near identical tables with a changed ordering of keys - they may produce different PKEs).
- Second, because our query can at least heuristiclaly optimize for the better (ie smaller) UK

go/vt/mysqlctl/schema.go
shlomi-noach: I think the heuristic here is "good enough" and predicts a good result most of the time. This is just due to that fact that unique key columns are not randomly assorted by some fuzzer, they are created by human beings, and if the first column is an integer as opposed to a text, it's _likely_ that overall the index is gonna be better. I'm just speculating of course, I don't have numbers.
At any case, like @mattlord says, this is an improvement over the already existing query used in `vitess` migrations, which is derived from `gh-ost`, where it worked/works very well. In fact, I don't remember a `gh-ost` issue complaining about the chosen index. Let's tick with this heuristic!

go/vt/vttablet/tabletserver/vstreamer/engine.go
```diff
@@ -511,3 +513,28 @@ func (vse *Engine) getMySQLEndpoint(ctx context.Context, db dbconfigs.Connector)
...
+				matches++
+				break
+			}
+		}
+		if matches == len(pkeColNames) {
+			break
+		}
+	}
+	return pkeCols, nil
+}
```
shlomi-noach, line 540: So I'm a bit unsure (as in "I didn't dive deep into this and so don't have a formed opinion") whether this code is needed, or should we reuse the logic in https://github.com/vitessio/vitess/pull/8364/files#diff-5e69c397e15dcfda72bb6fe8dba93b2d11a8d2d7b594ebfbbdcfdc07bfc3791d, which `vreplication` then picks up to populate the pk columns.

I'm mostly concerned that we might end up supporting two different methods of populating non-PK columns in `vreplication`.

mattlord: I had the same concern/thoughts. I have no confidence today in saying that the OnlineDDL `ukColumns` related behavior — getting the PKE or PKE with its own query, passing that info along as table comments, then storing the values in the workflow definition using the `source_unique_key_columns`, `target_unique_key_columns`, and `source_unique_key_target_columns` fields — is no longer needed. I do like having some override mechanism as an outlet for the rare times it may be needed, but otherwise it does feel redundant to me.

I think at least some of that is carried over from Gh-ost?

If you think that much of this could likely be trimmed out to unify OnlineDDL more with other VReplication workflows then I'll create an issue to investigate this and come up with a plan. That will also give me an excuse to finally dive into the OnlineDDL code. 🙂

mattlord:
> really nice work! Added my review as comments inline. Highlights:
> 
> * Let's keep the heuristic function, and moreover **skip/ignore** the mysql PKE check
> * Should we reuse `source_unique_key_columns` and `target_unique_key_columns` from [VReplication support for non-PRIMARY KEY (Online DDL context) #8364](https://github.com/vitessio/vitess/pull/8364)?

I'm unsure about this as noted in my other comment: https://github.com/vitessio/vitess/pull/10192#discussion_r869342045

I'm not sure if we should re-use those or get rid of them. Do you think it makes sense for me to create that investigation ticket so that I can get a better feel for these questions?

Thanks again for all of the help! ❤️

mattlord:
I created this issue to follow-up on how to align the OnlineDDL workflow with other VReplication workflows with regard to how PK and PKE indexes are chosen and managed: https://github.com/vitessio/vitess/issues/10259

And the follow-up documentation changes for this PR are in progress now here:https://github.com/vitessio/website/pull/1021

shlomi-noach:
> I'm not sure if we should re-use those or get rid of them.

Good question! No opinion, whichever way you choose!

PR #10194, opened 2022-05-02
Temp: Pin XtraBackup version used at 2.4.24 for 5.7 tests

mattlord:
## Description

⚠️ **Note:** pinning the version is *_temporary_* until XtraBackup 2.5.26 is released, at which time we will switch back to using the latest ⚠️ 

This pins the XtraBackup version used in our non-8.0 CI jobs to 2.4.24 as [2.4.25 crashes with the `--xtrabackup_stream_mode=tar` flag used with 5.7](https://jira.percona.com/browse/PXB-2756) (with 8.0 we use `--xtrabackup_stream_mode=xbstream`).

## Open Questions
1. Should we also pin the version for the 8.0 CI jobs for consistency? 
   * Decision: NO
2. Should we pin this permanently, even after 2.4.26 comes out with the fix in the coming weeks?
    * Decision: NO

## Related Issue(s)
  - Fixes: https://github.com/vitessio/vitess/issues/10146

## Checklist

### DIFF
--- a/.github/workflows/cluster_endtoend_12.yml
+++ b/.github/workflows/cluster_endtoend_12.yml
@@ -32,6 +32,7 @@ jobs:
               - 'config/**'
               - '.github/docker/**'
               - 'bootstrap.sh'
+              - '.github/workflows/**'
 
       - name: Build Docker Image
         if: steps.changes.outputs.end_to_end == 'true'
--- a/test/templates/cluster_endtoend_test.tpl
+++ b/test/templates/cluster_endtoend_test.tpl
@@ -39,6 +39,7 @@ jobs:
             - 'tools/**'
             - 'config/**'
             - 'bootstrap.sh'
+            - '.github/workflows/**'
 
     - name: Set up Go
       if: steps.changes.outputs.end_to_end == 'true'
--- a/test/templates/cluster_endtoend_test.tpl
+++ b/test/templates/cluster_endtoend_test.tpl
@@ -80,10 +80,10 @@ jobs:
         sudo apt-get install -y gnupg2
         sudo dpkg -i "percona-release_latest.$(lsb_release -sc)_all.deb"
         sudo apt-get update
-        if [[ -n ${XTRABACKUP_VERSION} ]]; then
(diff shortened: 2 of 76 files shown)


deepthi:
> Should we also pin the version for the 8.0 CI jobs for consistency?

Let's not do that. If they are running fine against the latest, that's good.

> Should we pin this permanently, even after 2.4.26 comes out with the fix in the coming weeks?

No, we should not.

test/templates/cluster_endtoend_test.tpl
```diff
@@ -6,6 +6,12 @@ concurrency:
 
 env:
   LAUNCHABLE_ORGANIZATION: "vitess"
+  # This is used if we need to pin the xtrabackup version used in tests.
+  # Doing so here because 2.4.25 crashes in our 5.7 tests. See:
+  #   https://jira.percona.com/browse/PXB-2756
+  # If this is NOT set then the latest version available will be used IF
+  # needed (see the InstallXtraBackup templating variable).
+  XTRABACKUP_VERSION: "2.4.24-1"
```
deepthi, line 14: Can we make this block also conditional on `InstallXtraBackup` instead of generating it into all the files that don't need it?

deepthi (commented):
Thank you for improving the workflow files. This should give us some slight improvements in CI run times too.

.github/workflows/cluster_endtoend_12.yml
```diff
@@ -32,6 +32,7 @@ jobs:
               - 'config/**'
               - '.github/docker/**'
               - 'bootstrap.sh'
+              - '.github/workflows/**'
```
deepthi, line 35: 👍

test/templates/cluster_endtoend_test.tpl
mattlord: Good idea! Done.

deepthi:
Things I noticed:
- We should probably start running xtrabackup tests against MySQL 8.0 as well
- Backup upgrade/downgrade tests don't seem to test xtrabackup
- vtgate_gen4 is flaky

PR #10195, opened 2022-05-03
refactor: move the Subshard opcode

systay:
## Description
This changes the engine opcode list to be according to most preferable first.

## Related Issue(s)
Based on feedback from #10151

## Checklist


systay:
👁️ 👁️ @harshit-gangal

harshit-gangal (approved):

PR #10197, opened 2022-05-03
Emit the ENGINE field for table options as case sensitive

dbussink:
This value has different casing in the canonical format from MySQL, so the parser here should not always uppercase the value assigned for the engine.

## Related Issue(s)

Part of extension work after https://github.com/vitessio/vitess/pull/9719

## Checklist


shlomi-noach:
@dbussink I believe you need to run `make codegen` to update `cached_size.go`

shlomi-noach (approved):
LGTM but delegating to someone from the query serving team

vmg (approved):

PR #10199, opened 2022-05-03
Add parsing support for Window Functions

K-Kumar-01:
Signed-off-by: Kushal Kumar <kushalkumargupta4@gmail.com>

## Description

- Parsing support for window functions.
- Add new test cases
- Update old test cases

More information on window functions [here](https://dev.mysql.com/doc/refman/8.0/en/window-functions.html)

## Related Issue(s)

#8604  
#4099
## Checklist


go/vt/sqlparser/ast.go
```diff
@@ -1991,6 +1991,67 @@ type TrimFuncType int8
 // TrimType is an enum to get types of Trim
 type TrimType int8
 
+// WindowSpecification represents window_spec
```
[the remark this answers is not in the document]
K-Kumar-01: Updated.

go/vt/sqlparser/ast.go
```diff
@@ -2077,6 +2078,84 @@ type TrimFuncType int8
...
+	// FrameUnitType is an enum to get types of Unit used in FrameClause.
+	FrameUnitType int8
+
+	// FrameUnitType is an enum to get types of FramePoint.
+	FramePointType int8
+
+	// NullTreatmentClause refers to null_treatment
+	// According to SQL Docs:  Some window functions permit a null_treatment clause that specifies how to handle NULL values when calculating results.
+	// This clause is optional. It is part of the SQL standard, but the MySQL implementation permits only RESPECT NULLS (which is also the default).
+	// This means that NULL values are considered when calculating results. IGNORE NULLS is parsed, but produces an error.
```
GuptaManan100, line 2138: ♥️

go/vt/sqlparser/sql.y
```diff
@@ -245,6 +262,7 @@ func bindVariable(yylex yyLexer, bvar string) {
 %left <str> OR '|'
 %left <str> XOR
 %left <str> AND
+%left <str> WINDOW_EXPR
```
GuptaManan100, line 265: Did you look at the precedence for Window_expr anywhere? Why did you add this in between the operators? The order in which symbols are specified has to do with the precedence of these operators.

go/vt/sqlparser/sql.y
```diff
@@ -5171,6 +5215,205 @@ trim_type:
...
+    $$ = CurrentRowType
+  }
+| UNBOUNDED PRECEDING
+  {
+    $$ = UnboundedPrecedingType
+  }
+| UNBOUNDED FOLLOWING
+  {
+    $$ = UnboundedFollowingType
+  }
```
GuptaManan100, line 5240: This rule is used only in 1 place. We should unroll this definition, since this rule does not actually list all the frame point types (`ExprPrecedingType` is not here) and it might someone else later to use this rule incorrectly

go/vt/sqlparser/ast.go
```diff
@@ -2560,6 +2639,60 @@ type (
 	JSONUnquoteExpr struct {
 		JSONValue Expr
 	}
+
+	// ArgumentLessWindowExpr stands for the following window_functions: CUME_DIST, DENSE_RANK, PERCENT_RANK, RANK, ROW_NUMBER
+	ArgumentLessWindowExpr struct {
+		Type       ArgumentLessWindowExprType
+		OverClause *OverClause
+	}
```
GuptaManan100, line 2647: Is this struct for window expression which don't take an argument?

go/vt/sqlparser/sql.y
```diff
@@ -6693,6 +7028,8 @@ reserved_keyword:
 | REPLACE
 | RIGHT
 | ROW_NUMBER
+| ROW
```
GuptaManan100, line 7031: nit: ROW should be above ROW_NUMBER

go/vt/sqlparser/sql.y
```diff
@@ -6386,6 +6717,10 @@ row_tuple:
   {
     $$ = ValTuple($2)
   }
+| ROW openb expression_list closeb
+  {
+    $$ = ValTuple($3)
+  }
```
GuptaManan100, line 6723: We were missing this for sooo long. 💕

go/vt/sqlparser/sql.y
```diff
@@ -5171,6 +5215,205 @@ trim_type:
...
+window_spec:
+sql_id window_partition_clause_opt order_by_opt frame_clause_opt
+  {
+    $$ = &WindowSpecification{ Name: $1.String(), PartitionClause: $2, OrderClause: $3, FrameClause: $4}
+  }
+|
+window_partition_clause_opt order_by_opt frame_clause_opt
+  {
+    $$ = &WindowSpecification{ PartitionClause: $1, OrderClause: $2, FrameClause: $3}
+  }
```
GuptaManan100, line 5325: Why not club them with `sql_id_opt`

go/vt/sqlparser/ast.go
```diff
@@ -2077,6 +2078,84 @@ type TrimFuncType int8
...
+	// More information available here: https://dev.mysql.com/doc/refman/8.0/en/window-functions-frames.html
+	FramePoint struct {
+		Type FramePointType
+		Expr Expr
+	}
+
+	// OverClause refers to over_clause
+	// More information available here: https://dev.mysql.com/doc/refman/8.0/en/window-functions-usage.html
+	OverClause struct {
+		WindowName string
```
GuptaManan100, line 2125: We should store this too as a ColIdent

GuptaManan100 (commented):
I have gone through the PR once. This is a big change! Well done @K-Kumar-01. I'll do a second pass once we address my comments and questions

go/vt/sqlparser/sql.y
K-Kumar-01: No, I didn't look for the precedence anywhere.
So, where should I reorder this value?

go/vt/sqlparser/ast.go
K-Kumar-01: Yes

go/vt/sqlparser/sql.y
K-Kumar-01: As far as I remember, there was a s/r conflict occurring when I tried to do so. I will recheck once and let you know.

go/vt/sqlparser/ast.go
K-Kumar-01: Will update this.

go/vt/sqlparser/sql.y
GuptaManan100: Put it lower in the list where other tokens are defined

GuptaManan100 (approved):

go/vt/sqlparser/ast.go
```diff
@@ -2561,6 +2640,61 @@ type (
...
+		OverClause          *OverClause
+	}
+
+	// FirstOrLastValueExprType is an enum to get types of FirstOrLastValueExpr.
+	FirstOrLastValueExprType int8
+
+	// NtileExpr stands for the NTILE()
+	NtileExpr struct {
+		IntValue   *int
+		VarValue   ColIdent
```
systay, line 2668: Maybe it's just me, but I think I'd prefer `VarValue` to be a ref, and if both `VarValue` and `IntValue` are nil, we know we are dealing with a null.

The problem with `IsNull` here (and admittedly it's a minor one) is that you can construct a struct that has "inconsistent state". if `IsNull` is true _and_ `VarValue` has a non-empty value, how do you know what you have?

go/vt/sqlparser/ast.go
```diff
@@ -2077,6 +2078,84 @@ type TrimFuncType int8
...
+		Windows WindowDefinitions
+	}
+
+	NamedWindows []*NamedWindow
+
+	// FrameClause represents frame_clause
+	// More information available here: https://dev.mysql.com/doc/refman/8.0/en/window-functions-frames.html
+	FrameClause struct {
+		Unit      FrameUnitType
+		IsBetween bool
```
systay, line 2110: This bool is not needed. If `End` is not nil, we are dealing with a Between

go/vt/sqlparser/parse_test.go
```diff
@@ -2943,6 +2940,72 @@ var (
...
 		output: "create table ts (\n\tid INT,\n\tpurchased DATE\n)\npartition by range (YEAR(purchased)) subpartition by hash (TO_DAYS(purchased))\n(partition p0 values less than (1990) (subpartition s0, subpartition s1),\n partition p1 values less than (2000),\n partition p2 values less than maxvalue (subpartition s2, subpartition `s3`))",
+	}, {
+		input:  "SELECT val, CUME_DIST() OVER w, ROW_NUMBER() OVER w, DENSE_RANK() OVER w, PERCENT_RANK() OVER w, RANK() OVER w AS 'cd' FROM numbers",
+		output: "select val, cume_dist() over w, row_number() over w, dense_rank() over w, percent_rank() over w, rank() over w as cd from numbers",
+	}, {
+		input:  "SELECT year, country, product, profit, CUME_DIST() OVER() AS total_profit FROM sales",
+		output: "select `year`, country, product, profit, cume_dist() over () as total_profit from sales",
+	}, {
+		input:  "SELECT val, CUME_DIST() OVER (ORDER BY val) AS 'cd' FROM numbers",
+		output: "select val, cume_dist() over ( order by val asc) as cd from numbers",
```
systay, line 2951: is the space before the `order by val...` intentional?

go/vt/sqlparser/ast_format.go
```diff
@@ -1446,6 +1453,140 @@ func (node *JSONStorageSizeExpr) Format(buf *TrackedBuffer) {
...
+	}
+	buf.WriteString(")")
+	if node.OverClause != nil {
+		buf.astPrintf(node, " %v", node.OverClause)
+	}
+}
+
+// Format formats the node
+func (node *NTHValueExpr) Format(buf *TrackedBuffer) {
+	buf.astPrintf(node, "nth_value(%v,", node.Expr)
```
systay, line 1548: we should have a space before the comma here, right?

```
"nth_value(%v,"
             👆
```
```

systay (approved):
LGTM

Some nitpicking, that's all I had to contribute :)

go/vt/sqlparser/ast_format.go
GuptaManan100: You are right. I am surprised that go-lint didn't complain!

go/vt/sqlparser/parse_test.go
K-Kumar-01: I used the formatting for this which was already implemented. So, If in overClause, there are no values prior to order by we will get this. 
I decided not to change the original formatting as that would change a lot of things.

go/vt/sqlparser/ast.go
K-Kumar-01: How the values are passed to the struct in the parser is decided by us. So will it affect us much?

go/vt/sqlparser/ast_format.go
K-Kumar-01: Updated this.

go/vt/sqlparser/ast.go
K-Kumar-01: Remove the field.

go/vt/sqlparser/ast.go
systay: well... now I would have to trust the people that wrote that parser, and you know they can't be trusted... ;)

like I wrote - it's not super-important. I just prefer to design structs so that they can't contain invalid data, unless there's a strong reason to do something different.

GuptaManan100: We don't even need to make VarValue a ref, there is a IsEmpty() function on ColIdent

GuptaManan100: I have opted for a different route. I stored a single value in the struct of type Expr instead of having 3 there of which only 1 should be specified. I think this is cleaner and easier to use down the line.

GuptaManan100: One refactor which has been deferred for now, and probably out of scope of this PR is that, we are storing user-defined variables, in columnName struct which isn't super correct, but we can fix it later.
For example, `select @@x` will have @@x stored inside a colident inside a column name

PR #10201, opened 2022-05-03
Revert semi_sync change to reintroduce enable_semi_sync and remove durability_policy

GuptaManan100:
## Description

We found a problem with the current implementation of the durability_policy flag. This flag only takes one value for the vtctld binary. This is incorrect, because people can be running multiple keyspaces each having its own different durability_policy. Since vtctld is a global component, we cannot require the user to create a vtctld for each keyspace having different durability_policies.

The proposed solution is to store the durability_policy in the topo server instead of having it as a flag. This change however will also require a two release upgrade process. In release-13 we are using the flag `enable_semi_sync` flag, and we want to continue doing that in release-14. We also want to deprecate the durability_policy flag. 

This PR reintroduces the enable_semi_sync flag and its usage. It removes the usages of durability_policy flag in the tests and binaries. Two tests in the ERS package which rely on the durability_policy information have been skipped for now.

This PR will be followed up by another PR which will add the RPCs to store and update the durability_policies in the keyspace information.

## Related Issue(s)

- #8975 

## Checklist

-   [x] Documentation was added or is not required - Release notes documentation and online docs will be updated after the next PR is also ready.

### DIFF
--- a/go/cmd/vtctl/vtctl.go
+++ b/go/cmd/vtctl/vtctl.go
@@ -49,9 +49,9 @@ import (
 )
 
 var (
-	waitTime         = flag.Duration("wait-time", 24*time.Hour, "time to wait on an action")
-	detachedMode     = flag.Bool("detach", false, "detached mode - run vtcl detached from the terminal")
-	durabilityPolicy = flag.String("durability_policy", "none", "type of durability to enforce. Default is none. Other values are dictated by registered plugins")
+	waitTime     = flag.Duration("wait-time", 24*time.Hour, "time to wait on an action")
+	detachedMode = flag.Bool("detach", false, "detached mode - run vtcl detached from the terminal")
+	_            = flag.String("durability_policy", "none", "type of durability to enforce. Default is none. Other values are dictated by registered plugins")
 )
 
 func init() {
@@ -100,7 +100,7 @@ func main() {
 		log.Warningf("cannot connect to syslog: %v", err)
 	}
 
-	if err := reparentutil.SetDurabilityPolicy(*durabilityPolicy); err != nil {
+	if err := reparentutil.SetDurabilityPolicy("none"); err != nil {
 		log.Errorf("error in setting durability policy: %v", err)
 		exit.Return(1)
 	}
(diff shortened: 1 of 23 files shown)


go/cmd/vtctl/vtctl.go
```diff
@@ -100,7 +100,7 @@ func main() {
 		log.Warningf("cannot connect to syslog: %v", err)
 	}
 
-	if err := reparentutil.SetDurabilityPolicy(*durabilityPolicy); err != nil {
+	if err := reparentutil.SetDurabilityPolicy("none"); err != nil {
```
deepthi, line 103: Why do we even need to set this?

deepthi (commented):
I didn't understand why we are setting durability policy to none in vtctl/vtctld/vtworker.
Beyond that, can you add the necessary release note to the summary doc for v14?

go/cmd/vtctl/vtctl.go
GuptaManan100: We need to set some durability policy so that the code does not panic. Since we want to remove the flag, we need to set something. I am going to refactor this in an upcoming PR, where we first check the keyspace information for the durability_policy and we set it there.

GuptaManan100:
@deepthi Would it be okay if I did all the documentation in website and the release notes in a PR after we have made all the code changes, because we are going to do a little more refactor.

deepthi:
> @deepthi Would it be okay if I did all the documentation in website and the release notes in a PR after we have made all the code changes, because we are going to do a little more refactor.

Additional refactor will not change the UX, will it? Requiring `-enable_semi_sync` again, not requiring `durability_policy` etc. are all happening in this PR, right?

go/cmd/vtctl/vtctl.go
deepthi: Ok, that's fine.

deepthi (approved):

GuptaManan100:
I will do the release notes along with the PR which adds the changes to use the durability_policy

GuptaManan100:
Website docs - https://github.com/vitessio/website/pull/1040

PR #10202, opened 2022-05-03
Fix for empty results when no shards can be found to route to [R13, v3]

### LINKED ISSUE #10144: Bug Report: exists may returns random result on a sharded table

systay:
## Description
Queries that follow the following pattern - 
```mysql
select exists(select * from foo where some_field=some_value)
```

The table `foo` would has a lookup vindex on `some_field` and there are no matching rows for `some_value`. What Vitess would do when no shards could be found to route to is to return an empty result. This is correct for most cases, but not always. The query above would return an empty result instead of a `0`.

This fix is for the V3 planner.

## Related Issue(s)
Fixes #10144

## Checklist


go/vt/vtgate/planbuilder/plan_test.go
```diff
@@ -411,6 +422,21 @@ func TestWithDefaultKeyspaceFromFileSharded(t *testing.T) {
...
+		keyspace: &vindexes.Keyspace{
+			Name:    "user",
+			Sharded: true,
+		},
+		tabletType: topodatapb.TabletType_PRIMARY,
+	}
+
+	testOutputTempDir := makeTestOutput(t)
+	testFile(t, "select_cases_with_user_as_default.txt", testOutputTempDir, vschema)
+}
```
frouioui, line 438: I might be missing, but I don't understand the clear difference between `TestOneWithUserAsDefault` and `TestWithUserDefaultKeyspaceFromFileSharded`. Is `select_cases_with_user_as_default.txt` a file where we will keep adding plan tests, or a file that serves for one-case test (like `onecase.txt`)?

go/vt/vtgate/planbuilder/route.go
```diff
@@ -466,7 +472,7 @@ func (rb *route) SubqueryCanMerge(pb *primitiveBuilder, inner *route) bool {
 
 	// if either side is a reference table, we can just merge it and use the opcode of the other side
 	if rb.eroute.Opcode == engine.Reference || inner.eroute.Opcode == engine.Reference {
-		return true
+		return rb.isSingleShard() && inner.isSingleShard()
```
frouioui, line 475: We should update the comment

frouioui (approved):
There are failing tests, the expected output should be changed. Otherwise, looks good to me, I left a few nits and questions.

go/vt/vtgate/planbuilder/plan_test.go
systay: yes. `TestOne...` is for one-offs, and `TestWithUserDefaultKeyspaceFromFileSharded` is forever

PR #10205, opened 2022-05-03
[vtadmin] non-blocking resolver

ajm188:
## Description

This PR updates the cluster discovery resolver to make `ResolveNow` a non-blocking call, by signaling on a channel to watcher goroutine which is continually updating the address list for a cluster component with backoff on failure. It's largely a port of grpc-go's dns-resolver with the "do a dns lookup" replaced with "make a discovery call". (See [dns_resolver](https://github.com/grpc/grpc-go/blob/3bf6719fc8ab5dac43b8494fcdc7e892efde6ea1/internal/resolver/dns/dns_resolver.go#L209) for reference).

The only notable changes are:
* I added extra backoff strategies, namely "linear" (which does `cur + multiplier` instead of exponential's `cur * multiplier`) and "none" (which never backs-off).
* I fixed the backoff for exponential and linear, to prevent jittering from exceeding the configured MaxDelay. This is an oversight in grpc-go's implementation as far as I can tell.
* I added a configuration option to govern the minimum wait time between successful iterations of the watcher loop. In grpc-go this is hard-coded to 30s.
* I corrected some usages of the timer. The DNS resolver leaks open timer channels in certain conditions.

## Related Issue(s)

#9977 

## Checklist

### DIFF
--- a/go/vt/vtadmin/internal/backoff/backoff.go
+++ b/go/vt/vtadmin/internal/backoff/backoff.go
@@ -1,0 +1,127 @@
+/*
+Copyright 2022 The Vitess Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package backoff implements different backoff strategies for retrying failed
+// operations in VTAdmin.
+//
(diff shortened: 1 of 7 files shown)


go/vt/vtadmin/internal/backoff/backoff.go
```diff
@@ -0,0 +1,127 @@
...
+	// Randomize backoff delays so that if a cluster of requests start at
+	// the same time, they won't operate in lockstep.
+	backoff *= 1 + cfg.Jitter*(vtrand.Float64()*2-1)
+	if backoff < 0 {
+		return 0
+	}
+
+	// NOTE: We differ from grpc's exponential backoff here, which actually can
+	// jitter to a backoff that exceeds the config's MaxDelay, which in my (ajm188)
+	// opinion is a bug.
```
doeg, line 93: A jitterbug?

doeg (approved):
This looks good to me, though I admit I don't really use channels enough to have a good intuition for them. 

I deployed this internally and it works great. ✅

PR #10207, opened 2022-05-03
Allow updating unowned lookup vindex columns.

arthurschreiber:
Signed-off-by: Arthur Schreiber <arthurschreiber@github.com>

## Checklist

### DIFF
--- a/go/vt/vtgate/endtoend/update/lookup_unique_test.go
+++ b/go/vt/vtgate/endtoend/update/lookup_unique_test.go
@@ -1,0 +1,256 @@
+/*
+Copyright 2019 The Vitess Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package endtoend
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"vitess.io/vitess/go/vt/log"
+
+	"vitess.io/vitess/go/mysql"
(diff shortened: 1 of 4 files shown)


go/vt/vtgate/engine/update.go
```diff
@@ -147,40 +147,70 @@ func (upd *Update) updateVindexEntries(vcursor VCursor, bindVars map[string]*que
...
 				}
 
-				if err := colVindex.Vindex.(vindexes.Lookup).Update(vcursor, fromIds, ksid, vindexColumnKeys); err != nil {
+				// All columns for this Vindex are set to null, so we can skip verification
+				if allNulls {
+					continue
+				}
+
+				// If values were supplied, we validate against keyspace id.
+				verified, err := vindexes.Verify(colVindex.Vindex, vcursor, [][]sqltypes.Value{vindexColumnKeys}, [][]byte{ksid})
```
arthurschreiber, line 206: This _could_ be improved by first building a list of all keyspace ids and all column values, but I feel like that can happen in a separate pull request - the existing call to update owned vindexes does not batch operations either.

go/vt/vtgate/endtoend/update/lookup_unique_test.go
```diff
@@ -0,0 +1,256 @@
...
+}
+
+func exec(t *testing.T, conn *mysql.Conn, query string) *sqltypes.Result {
+	t.Helper()
+	qr, err := conn.ExecuteFetch(query, 1000, true)
+	if err != nil {
+		t.Fatal(err)
+	}
+	return qr
+}
```
frouioui, line 256: nit: you can use `utils.Exec` instead of this function

arthurschreiber: Thanks! I changed this as suggested. 🙇‍♂️

frouioui (approved):
Looks good to me! Thank you for this change. I'd probably wait for @harshit-gangal's approval as well as he is more knowledgeable about vindexes.

frouioui:
I have the feeling that [this documentation](https://vitess.io/docs/14.0/reference/features/vindexes/#consistent-lookup-usage) needs to be updated according to the new changes proposed by this pull request. 

> Due to this, a current limitation with consistent lookup is that it cannot support an insert followed by an update or delete in the same transaction for the same consistent lookup column value.

frouioui:
Hello @arthurschreiber, everything looks good to me.

PR #10209, opened 2022-05-03
Implement fallback behavior for cross-build

FancyFane:
Signed-off-by: FancyFane <fane@planetscale.com>

## Description

Changes were recently made to the cross-compile section of the Makefile. These changes use an if statement, which never falls back to the old behavior. This causes our weekly Vitess releases to fail when compiling Apple ARM binaries. This change allows the fallback behavior for this scenario. 

## Checklist

-   [ n/a ] Tests were added or are not required
-   [ n/a ] Documentation was added or is not required

## Deployment Notes
None.


FancyFane:
@mattlord this was marked stale again. Making a comment to remove the stale flag.

mattlord (changes requested):
We'll need to revisit this after https://github.com/vitessio/vitess/pull/10956

Do we still need to make any changes? If so, the changes made will be different as we're not using the rice boxes anymore.

FancyFane:
@mattlord just tested this on a new build and the changes have resolved this issue. No further changes will be needed, and I am closing this out.

PR #10212, opened 2022-05-04
check for connectionID before adding to querylist

rsajwani:

## Description

This is a bug where long running streaming queries with transaction fails to shutdown gracefully. Previously during execStreamSQL we were adding query to OLAP queryList regardless of whether it is transactional or not. With this change we put streaming transaction queries to stateful list, which result it in graceful shutdown of these queries during handleShutdownGracePeriod  if it exceeds the grace period.

I decided to pass isTransaction as parameter to execStreamSQL because I don't want execStreamSQL to have a logic of isTransaction. Let me know if you feel otherwise. 

## Checklist

### DIFF
--- a/go/vt/vttablet/endtoend/transaction_test.go
+++ b/go/vt/vttablet/endtoend/transaction_test.go
@@ -366,6 +366,110 @@ func TestShutdownGracePeriod(t *testing.T) {
 	client.Rollback()
 }
 
+func TestShutdownGracePeriodWithStreamExecute(t *testing.T) {
+	client := framework.NewClient()
+
+	err := client.Begin(false)
+	require.NoError(t, err)
+	go func() {
+		_, err = client.StreamExecute("select sleep(10) from dual", nil)
+		assert.Error(t, err)
+	}()
+
+	started := false
+	for i := 0; i < 10; i++ {
+		queries := framework.LiveQueryz()
+		if len(queries) == 1 {
+			started = true
+			break
+		}
+		time.Sleep(10 * time.Millisecond)
+	}
+	assert.True(t, started)
+
+	start := time.Now()
+	err = client.SetServingType(topodatapb.TabletType_REPLICA)
+	require.NoError(t, err)
+	assert.True(t, time.Since(start) < 5*time.Second, time.Since(start))
+	client.Rollback()
+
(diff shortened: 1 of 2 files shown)


go/vt/vttablet/endtoend/transaction_test.go
```diff
@@ -366,6 +366,110 @@ func TestShutdownGracePeriod(t *testing.T) {
...
+
+	start := time.Now()
+	err = client.SetServingType(topodatapb.TabletType_REPLICA)
+	require.NoError(t, err)
+	assert.True(t, time.Since(start) < 5*time.Second, time.Since(start))
+	client.Rollback()
+
+	client = framework.NewClientWithTabletType(topodatapb.TabletType_REPLICA)
+	err = client.Begin(false)
+	require.NoError(t, err)
```
harshit-gangal, line 398: nit: may be you can continue using the old client.

go/vt/vttablet/endtoend/transaction_test.go
```diff
@@ -366,6 +366,110 @@ func TestShutdownGracePeriod(t *testing.T) {
...
```
harshit-gangal, line 450: nit: may be you can continue using the old client.

go/vt/vttablet/tabletserver/query_executor.go
```diff
@@ -301,7 +301,7 @@ func (qre *QueryExecutor) Stream(callback StreamCallback) error {
 						return err
 					}
 					defer dbConn.Recycle()
-					return qre.execStreamSQL(dbConn, sql, func(result *sqltypes.Result) error {
+					return qre.execStreamSQL(dbConn, qre.connID != 0, sql, func(result *sqltypes.Result) error {
```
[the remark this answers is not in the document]
rsajwani: If you don't feel strongly about this .. i would rather keep it like this so in the future if anyone do any kind of refactoring/change in this code block, it will be less likely to result in regression.

go/vt/vttablet/endtoend/transaction_test.go
rsajwani: spoke offline and we decided to keep it as it is

PR #10218, opened 2022-05-04
[vtadmin] Tidy up /settings (née /debug) page

doeg:

## Description

Some tidying up before GA, but this is also (sort of) a prerequisite for single-component VTAdmin (https://github.com/vitessio/vitess/issue/10118 and https://github.com/vitessio/vitess/pull/10214). 

Specifically, with single-component VTAdmin, the front- and back-end will (by definition) share the same "route space". (This is a downside, but not a big one.) vtadmin-api already [defines several `/debug` routes
](https://github.com/vitessio/vitess/blob/99d5299/go/vt/vtadmin/api.go#L168-L183), and while it doesn't have a handler for the root `/debug` route, it just seemed weird and unnecessary to have vtadmin-web overlap when it doesn't really need to.

Here's what this PR does instead: 

- Avoids route overlap by moving the `/debug` view to the `/settings` route, since that's likely a route we will use ✨ someday ✨ for user preferences (dark mode, time formatting, etc.). 

- Omits the """style guide""" (I use that term loosely, lol) for production builds, since it's only really useful in dev mode. It also makes the production build a lil smaller and simpler.

- Removed the "Debug" link from the NavRail, since this "settings page" in current form (just a blob of `NODE_ENV` json) isn't relevant to anyone except those that already know about it. 🔮 

Here's what it looks like now for a prod build (on my local machine):

Screen Shot 2022-05-04 at 3 18 40 PM

## Checklist

## Deployment Notes

N/A

### DIFF
--- a/web/vtadmin/src/components/App.tsx
+++ b/web/vtadmin/src/components/App.tsx
@@ -107,7 +107,7 @@ export const App = () => {
                             <Workflow />
                         </Route>
 
-                        <Route path="/debug">
+                        <Route path="/settings">
                             <Debug />
                         </Route>
 
--- a/web/vtadmin/src/components/NavRail.module.scss
+++ b/web/vtadmin/src/components/NavRail.module.scss
@@ -21,7 +21,6 @@ $navRailHoverTransition: background-color 0.1s ease-in-out;
     border-right: solid 1px var(--backgroundSecondaryHighlight);
     display: flex;
     flex-direction: column;
-    justify-content: space-between;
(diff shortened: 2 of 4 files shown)


ajm188 (approved):
lgtm! my only question is if we should rename `Debug.tsx` to `Settings.tsx` to match the route, but i defer to you!

doeg:
> my only question is if we should rename Debug.tsx to Settings.tsx to match the route

Oh thanks for catching that!

PR #10222, opened 2022-05-05
Flaky test fix: TestFoundRows

GuptaManan100:
## Description

The test `TestFoundRows` runs queries that are dependent on schema tracking running successfully before the queries are executed. This however, makes the test flaky since some times, schema tracking doesn't run before we execute the queries, which leads to errors.
This PR fixes this issue, by waiting for the schema-tracking to run and mark the table used as authoritative before we run any queries against it.

## Checklist

-   [x] "Backport me!" label has been added if this change should be backported -> Forward port required

### DIFF
--- a/go/test/endtoend/vtgate/queries/foundrows/found_rows_test.go
+++ b/go/test/endtoend/vtgate/queries/foundrows/found_rows_test.go
@@ -18,14 +18,15 @@ package foundrows
 
 import (
 	"context"
+	"fmt"
 	"testing"
-
-	"vitess.io/vitess/go/test/endtoend/vtgate/utils"
+	"time"
 
 	"github.com/stretchr/testify/require"
 
 	"vitess.io/vitess/go/mysql"
 	"vitess.io/vitess/go/test/endtoend/cluster"
+	"vitess.io/vitess/go/test/endtoend/vtgate/utils"
 )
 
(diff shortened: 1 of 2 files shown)


go/test/endtoend/vtgate/queries/foundrows/found_rows_test.go
```diff
@@ -58,3 +64,38 @@ func TestFoundRows(t *testing.T) {
...
+	}
+}
+
+func getTableT2Map(res *interface{}) map[string]interface{} {
+	t2Map := convertToMap(convertToMap(convertToMap(convertToMap(convertToMap(*res)["keyspaces"])[KeyspaceName])["tables"])["t2"])
+	return t2Map
+}
+
+func convertToMap(input interface{}) map[string]interface{} {
+	output, _ := input.(map[string]interface{})
```
systay, line 99: nit: if you don't care about the boolean, you can just omit it altogether:

```
output := input.(map[string]interface{})
```

systay (approved):

go/test/endtoend/vtgate/queries/foundrows/found_rows_test.go
GuptaManan100: Okay 👍

Issue #10203, opened 2022-05-03
schemadiff: tracking issue


shlomi-noach:
Fashionably late, setting up a tracking issue for `schemadiff`, introduced in #9719.

## General overview

`schemadiff` is a `vitess` library that manages schema definitions, and is able to present a logical diff between two tables, two views, or two schemas. It does so declaratively, without accessing a MySQL server, and is based on `sqlparser`, a core library in `vitess`.

## Entities

`schemadiff` recognizes these entities:

- A table, mapped to a database `TABLE`
- A view, mapped to a database `VIEW`
- A schema, a collection of tables and views

Tables and views are also entities known to `sqlparser`'s AST. The parser can parse `CREATE TABLE` or `CREATE VIEW` statements into well formed constructs. `schemadiff`'s main logic is to utilize these well formed constructs, comparing them, manipulating them or validating them.

## Diffs

At the heart of `schemadiff` is to be able to diff:

- Two tables, resulting in a potentially empty `ALTER TABLE` diff
- Two views, resulting in a potentially empty `ALTER VIEW` diff
- Two schemas, resulting in a potentially empty list of `DROP/ALTER/CREATE TABLE/VIEW` statements

A resulting diff is a well formed construct in itself. e.g. the diff of two tables, if not empty, is a `ALTER TABLE`, represented by `sqlparser`'s matching AST construct. The diff also attains context to its "from" and "to" entities.

Taking a tables diff as the classic example, the diff is an `ALTER TABLE` query that takes a table from one format to another (ignoring table name). The alteration may include:

- Adding, removal, modification of columns
- Adding, removal of keys
- Adding, removal of foreign keys
- Adding, removal, modification of table options
- Changing of partitioning scheme

## Input and output

`schemadiff` reads queries, or statements. Because it does not access a MySQL server, by design, it is fed with `CREATE` statements that depict a table, view or schema.
SQL can have many formatting options, and a table can be presented in multiple ways (see [normalization](#normalization) below). It is advisable to feed `schemadiff` with formal output of a `SHOW CREATE TABLE` or `SHOW CREATE VIEW` statements.

Entity constructs can export their well formed constructs, or they can export their `CREATE` SQL statements. For example, you may export a schema into a list of SQL queries via `ToQueries()` or into one big blob of definitions with `ToSQL()`.

Diffs can export their formal constructs, or they can export their SQL statements.

## Validation

[...]

PR #10224, opened 2022-05-05
Fix parsing of the foreign key constraint actions in different order

dbussink:
On a `FOREIGN KEY` one can defined an action `ON UPDATE` and `ON DELETE`. The parser would only allow one specific order where first the `ON DELETE` is specified and then the `ON UPDATE`.

This meant that a query which had it in the reverse order would fail to parse. This also allows bypassing the check on DDL statements that add a foreign key constraint too.

Additionally, we move the `schemadiff` logic here to be strict for parsing. It was blowing up with a panic on a schema with such a differently ordered foreign key constraint but it should have error on any parser error.

In case `schemadiff` can't parse something, we can't trust it to do any further operations on the schema.

## Related Issue(s)

- Found as part of the `schemadiff` work from https://github.com/vitessio/vitess/issues/10203
- https://github.com/vitessio/vitess/issues/8604#

## Checklist


go/vt/sqlparser/parse_test.go
```diff
@@ -3912,6 +3915,39 @@ func TestCreateTable(t *testing.T) {
 	constraint second_ibfk_1 foreign key (k, j) references simple (a, b) on delete cascade on update set default,
 	constraint second_ibfk_1 foreign key (k, j) references simple (a, b) on delete set default on update set null,
 	constraint second_ibfk_1 foreign key (k, j) references simple (a, b) on delete set null on update restrict,
+	constraint second_ibfk_1 foreign key (k, j) references simple (a, b) on update set default on delete cascade ,
+	constraint second_ibfk_1 foreign key (k, j) references simple (a, b) on update set null on delete set default,
+	constraint second_ibfk_1 foreign key (k, j) references simple (a, b) on update restrict on delete set null,
+	constraint second_ibfk_1 foreign key (k, j) references simple (a, b) on update no action,
+	constraint second_ibfk_1 foreign key (k, j) references simple (a, b) on update cascade
+)`,
+			output: `create table t (
```
dbussink, line 3924: This specific output was needed since the parser parses the actions into a struct so the ordering specifically of this is lost. 

We have verified though that MySQL doesn't accidentally allows a list of these items, just only both orders so there's no potential semantic difference due to the different order.

PR #10226, opened 2022-05-06
gen4: JOIN USING planning

### LINKED ISSUE #9201: Support join with `USING` construct
Join queries with a `USING` construct should be supported by our planner, as of now, we return an unsupported error.

Query:
```sql
select * from user join user_extra using(id)
```

Current error:
```
unsupported: join with USING(column_list) clause for complex queries
```

systay:
## Description
Adds logic to handle `JOIN USING` queries.

This involves three rewrites of the query:

  1. _ExpandStar_ needs to handle these columns. There is special logic for which columns should be visible when the tables are joined with `USING`
  2. We need to rewrite uses of the columns in the `WHERE` clause and other clauses where it could appear.
  3. Remove the `USING` clause and instead add the appropiate predicates to the `WHERE` clause

Example:

```sql
select * 
from t1 
   join t2 using (c1) 
where c1 = 13
```

The table `t1` has two columns - `c1` and `c2`. The table `t2` has two columns - `c1` and `c3`.

The same query  where all these rewrites have been performed:

```sql
select t1.c1 as c1, t1.c2 as c2, t2.c3 as c3 
from t1 
   join t2 
where t1.c1 = t2.c1 
  and t1.c1 = 13
```

All of these rewrites happen during semantic analysis, so we didn't have to change the planner to handle these queries.

## Related Issue(s)
Fixes #9201


go/test/endtoend/vtgate/gen4/gen4_test.go
```diff
@@ -31,15 +32,31 @@ import (
 	"vitess.io/vitess/go/mysql"
 )
 
-func TestOrderBy(t *testing.T) {
+func start(t *testing.T) (*mysql.Conn, func()) {
```
frouioui, line 35: the `start` function could use the "mysql compare vitess" tool

go/vt/vtgate/semantics/binder.go
```diff
@@ -98,6 +136,44 @@ func (b *binder) up(cursor *sqlparser.Cursor) error {
 	return nil
 }
 
+func (b *binder) rewriteJoinUsingColName(deps dependency, node *sqlparser.ColName, currentScope *scope) (dependency, error) {
+	newTbl := deps.recursive.Constituents()[0]
```
frouioui, line 140: nit: it's probably safer to first check if the constituents slice is not empty or nil

go/vt/vtgate/semantics/early_rewriter.go
```diff
@@ -149,7 +150,74 @@ func realCloneOfColNames(expr sqlparser.Expr, union bool) sqlparser.Expr {
...
+				sel.Where = &sqlparser.Where{
+					Type: sqlparser.WhereClause,
+					Expr: sqlparser.AndExpressions(predicates...),
+				}
+			} else {
+				sel.Where.Expr = sqlparser.AndExpressions(append(predicates, sel.Where.Expr)...)
+			}
+			return nil
+		}
+	}
```
frouioui, line 211: Shouldn't we keep going up the scope here by changing the value of `current`?

frouioui (approved):

go/vt/vtgate/semantics/early_rewriter.go
systay: 🤦

go/test/endtoend/vtgate/gen4/gen4_test.go
```diff
@@ -32,51 +32,35 @@ import (
 )
 
 func TestOrderBy(t *testing.T) {
```
frouioui, line 34: I see in this file most of the test use `utils.Exec` with `mcmp.VtConn`, could we execute the query against MySQL too?

systay: I tried, but it failed in a bunch of places and I didn't want go hunting it down in this PR

go/vt/vtgate/semantics/vtable.go
```diff
@@ -37,18 +37,14 @@ var _ TableInfo = (*vTableInfo)(nil)
 // dependencies implements the TableInfo interface
 func (v *vTableInfo) dependencies(colName string, org originable) (dependencies, error) {
```
harshit-gangal, line 38: nit: the function does not return error from any of the implementations, hence the signature can be changed.

go/test/endtoend/vtgate/gen4/gen4_test.go
frouioui: got it

harshit-gangal (commented):
As per MySQL, columns in the using join list should be the first fields when expanding star expression.

failure case
```
sql:    "select * from t1 join t5 using (b)",
expSQL: "select t1.b as b, t1.a as a, t1.c as c, t5.a as a from t1 join t5 where t1.b = t5.b",
```

more details: https://dev.mysql.com/doc/refman/8.0/en/join.html

harshit-gangal:
failure case on having clause
```
sql:    "select * from t2 join t4 using (c1) having c1 = 2",
expSQL: "select t2.c1 as c1, t2.c2 as c2, t4.c4 as c4 from t2 join t4 where t2.c1 = t4.c1 having t2.c1 = 2",
```

harshit-gangal (approved):
awesome work here.

go/test/endtoend/utils/cmp.go
```diff
@@ -89,6 +89,16 @@ func (mcmp *MySQLCompare) AssertMatchesNoOrder(query, expected string) {
...
+// AssertMatchesNoOrderInclColumnNames executes the given query against both Vitess and MySQL.
+// The test will be marked as failed if there is a mismatch between the two result sets.
+// This method also checks that the column names are the same and in the same order
+func (mcmp *MySQLCompare) AssertMatchesNoOrderInclColumnNames(query, expected string) {
+	mcmp.t.Helper()
+	qr := mcmp.ExecWithColumnCompare(query)
+	actual := fmt.Sprintf("%v", qr.Rows)
+	assert.Equal(mcmp.t, utils.SortString(expected), utils.SortString(actual), "for query: [%s] expected \n%s \nbut actual \n%s", query, expected, actual)
+}
+
```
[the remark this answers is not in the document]
systay: https://github.com/vitessio/vitess/blob/be791d71bbb0aca251f9dee9cb7a42ab167730e4/go/test/endtoend/vtgate/gen4/gen4_test.go#L446-L455

Issue #8705, opened 2021-08-27
[vtadmin] View cells and cell info 


doeg:
# vtctld2 UI 

<img width="1792" alt="Screen Shot 2021-08-27 at 1 23 50 PM" src="https://user-images.githubusercontent.com/855595/131165742-beff8a7e-7934-4a2f-8930-47b306e22adf.png">
<img width="1792" alt="Screen Shot 2021-08-27 at 1 24 56 PM" src="https://user-images.githubusercontent.com/855595/131165774-b875edb5-a011-43d0-b97e-c042a3c21abd.png">
<img width="1792" alt="Screen Shot 2021-08-27 at 1 24 57 PM" src="https://user-images.githubusercontent.com/855595/131165783-b10ff19d-08ef-4abf-a0bf-6b51cd99dbcd.png">

doeg:
I thiiiink this probably rolls into the topology browser (#8704)... not sure why I ticketed it as a separate thing. 🤔 Maybe because of this (on that ticket):

> This ticket is specifically for the topology browser interface. Sub-functionality is ticketed separately (e.g., viewing vtctld elections) even when it's part of the topology browser in the vtctld2 UI.

@notfelineit does putting (well, keeping) this in the topo browser sound right to you? If so, I can close it out. Otherwise, I'm more than happy to take it, just wanted to check! :)

PR #10227, opened 2022-05-06
[vtadmin] cell apis

ajm188:
## Description

vtadmin variants of:
- `GetCellInfoNames` (via `/cells?names_only=true`)
- `GetCellInfo`
- `GetCellsAliases`

### Demo

```
➜  local git:(andrew/vtadmin/cell-apis) ✗ curl -s localhost:14200/api/cells\?cluster\=local | jq '.'
{
  "result": {
    "cell_infos": [
      {
        "cluster": {
          "id": "local",
          "name": "local"
        },
        "name": "zone1",
        "cell_info": {
          "server_address": "localhost:2379",
          "root": "/vitess/zone1"
        }
      }
    ]
  },
  "ok": true
}
```

## Related Issue(s)

#8705

## Checklist

-   [x] Tests were added or are not required -- TODO!


go/vt/vtadmin/cluster/cluster.go
```diff
@@ -768,6 +768,116 @@ func (c *Cluster) getShardSets(ctx context.Context, keyspaces []string, keyspace
...
+// GetCellInfos returns a list of ClusterCellInfo objects for cells in the
+// given cluster.
+//
+// If req.Cells is set, cells are restricted only to cells with those names.
+// Note: specifying a cell name that does not exist in the cluster fails the
+// overall request.
+//
+// If req.NamesOnly is set, each ClusterCellInfo will only contain the Cluster
+// and Name fields. req.Cells takes precedence over this option.
+func (c *Cluster) GetCellInfos(ctx context.Context, req *vtadminpb.GetCellInfosRequest) ([]*vtadminpb.ClusterCellInfo, error) {
```
doeg, line 780: Obligatory question of "do we need tests for either of these"

go/vt/vtadmin/cluster/cluster.go
```diff
@@ -768,6 +768,116 @@ func (c *Cluster) getShardSets(ctx context.Context, keyspaces []string, keyspace
...
+		resp, err := c.Vtctld.GetCellInfoNames(ctx, &vtctldatapb.GetCellInfoNamesRequest{})
+		if err != nil {
+			return nil, fmt.Errorf("failed to GetCellInfoNames: %w", err)
+		}
+
+		names = resp.Names
+	}
+
+	namesOnly := req.NamesOnly
+	if namesOnly {
```
doeg, line 799: Just a question about """idiomatic go""": isn't this the same as `if namesOnly && len(req.Cells) != 0`?

doeg (commented):
Thank you for doing this!!! Obligatory blocking question about tests but otherwise this is :chef-kiss: and I super appreciate it.

go/vt/vtadmin/cluster/cluster.go
ajm188: yuppppp (i originally had An Thought about doing the early return namesOnly loop in this section too, but never came back here), thanks!

go/vt/vtadmin/cluster/cluster.go
ajm188: yes and yes

ajm188: annnd done

doeg (approved):

ajm188:
i'm gonna wait for #10233 and rebase, because i added a `Dial` call here that'll need to get cleaned up, if that's okay with you!

PR #10228, opened 2022-05-07
Flakes: Add retries to ZooKeeper unit test around startup

mattlord:
## Description

The unit test CI workflows fail periodically and the most common cause I've seen over time is ZooKeeper unit tests ([for example](https://github.com/vitessio/vitess/runs/6335473985?check_suite_focus=true)).

This improves the retry and wait functionality to deal with ephemeral network issues or longer startup and init times.

We also address some linter warnings as this code has not been touched in some time.

## Checklist


ajm188 (approved):
LGTM!

PR #10233, opened 2022-05-08
[vtadmin] Remove explicit Dial-ing for vtctldclient proxy

ajm188:
## Description

(sorry this is lorge, i can break it up in a few stages if that's helpful!)

TODO add more detailed description here

### change breakdown
- refactor Dial/Redial tests to operate under the new model
- thread contexts from startup all the way through to `vtctldclient.New`
- remove all explicit Dial calls
- remove `Dial` from the interface

### testing

- all unit tests pass (with both `-count=50` and `-race` so i'm feeling good)
- can run the local example

## Checklist

-   [ ] Documentation was added or is not required (leaving this unchecked until i write an actual PR description)

### DIFF
--- a/go/cmd/vtadmin/main.go
+++ b/go/cmd/vtadmin/main.go
@@ -92,7 +92,7 @@ func startTracing(cmd *cobra.Command) {
 }
 
 func run(cmd *cobra.Command, args []string) {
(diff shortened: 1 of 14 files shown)


doeg (approved):
🌟

PR #10234, opened 2022-05-08
schemadiff: analyze and apply ADD PARTITION and DROP PARTITION statements

shlomi-noach:
## Description

In this PR, `schemadiff` is able to:

- Utilize `RangeRotationStatements` in `DiffHints`
- Diff two range-partition tables and analyze a `ADD PARTITION` or a `DROP PARTITION` diff.
  - the change must be a single dropped partition or a single added partition
  - the change must be a rotation; ie if a partition is dropped, then it must be the first partition. If a partition is added, it must be the last.
  - In all other cases, we error
  - When `RangeRotationStrategy` is `RangeRotationFullSpec` (the default) then there is no error if more than one partition is added/dropped and behavior remains a full partitioning spec
  - When `RangeRotationStrategy` is `RangeRotationIgnore` then there is no error if more than one partition is added/dropped, and if this is a rotation, the partitioning change is silently ignored. If this is not a rotation, a full spec is generated.
- `Apply` a `alter table ... add partition` or a `alter table drop partition`. Some validations apply (the table is partitiond, no name conflict etc.) but at this time there is no range value validation.

## Related Issue(s)

#10203 

## Checklist


shlomi-noach:
I've converted this PR to Draft. I want to explore an idea by @dbussink where a table diff might generate a sequence of diffs, not just a single one; such that we might diff two tables with multiple partition rotation changes, and generate a one-change-at-a-time sequence of alters.

go/vt/schemadiff/table.go
```diff
@@ -1359,8 +1473,18 @@ func (c *CreateTableEntity) Apply(diff EntityDiff) (Entity, error) {
 		dupCreateTable.Comments = &d
 	}
 	dup := &CreateTableEntity{CreateTable: *dupCreateTable}
-	if err := dup.apply(alterDiff); err != nil {
-		return nil, err
+	for diff != nil && !diff.IsEmpty() {
```
dbussink, line 1476: I guess it's explicitly not allowed / supported to supply a list of diffs where there might be an empty diff in that sequence? Because those logic would halt at the first empty diff even if that diff references a next non empty diff.

dbussink (approved):
One small question here, but nothing blocking.

PR #10237, opened 2022-05-09
[vtadmin] Removing explicit Dial-ing for vtsql proxy

ajm188:
## Description

This builds on #10233, but for vtsql this time (it's based on the same branch, so it should be merged _after_ that PR).

It's the same changes, applied in the same order (config options, context plumbing through the call stack, etc etc) but applied to the `DB` interface and `VTGateProxy` type instead. In addition:

- make `vtgateconn`'s private `drivers` map thread safe. this was tripping the race detector during tests (see 69cf2f2975585025aea1cc7ab58d35aa1d5b4dfa)

## Checklist

### DIFF
--- a/go/vt/vtgate/vtgateconn/vtgateconn.go
+++ b/go/vt/vtgate/vtgateconn/vtgateconn.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"sync"
 
 	"vitess.io/vitess/go/sqltypes"
 	"vitess.io/vitess/go/vt/log"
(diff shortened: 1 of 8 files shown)


go/vt/vtgate/vtgateconn/vtgateconn.go
```diff
@@ -183,7 +190,10 @@ func RegisterDialer(name string, dialer DialerFunc) {
 
 // DialProtocol dials a specific protocol, and returns the *VTGateConn
 func DialProtocol(ctx context.Context, protocol string, address string) (*VTGateConn, error) {
+	dialersM.Lock()
 	dialer, ok := dialers[protocol]
+	dialersM.Unlock()
```
doeg, line 195: Is there a reason for not using `defer` beyond "it's like, one line"? (Purely curious/not a leading question lol.)

doeg (approved):
🌟 🌟

go/vt/vtgate/vtgateconn/vtgateconn.go
ajm188: because we don't need to hold the mutex to invoke the dialer, just to get it out of the map

PR #10238, opened 2022-05-09
Add convenience functions for working with schemadiff objects

dbussink:
This adds a number of convenience methods to schemadiff that are useful when working with schema related objects and inspecting what a schema looks like.

It avoids having to do some casts often when for example you only care about getting a table from a schema (and not views).

The diff methods are useful when getting the underlying object so you don't have to grab it with `Statement()` and then cast it again to the real underlying `sqlparser` object.

## Related Issue(s)

#10203 

## Checklist

### DIFF
--- a/go/vt/schemadiff/schema.go
+++ b/go/vt/schemadiff/schema.go
@@ -241,6 +241,46 @@ func (s *Schema) EntityNames() []string {
 	return names
 }
 
+// Tables returns this schema's tables in good order (may be applied without error)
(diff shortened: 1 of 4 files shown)


shlomi-noach (approved):

shlomi-noach:
@dbussink add a couple tests e.g. for `TableNames()` and `ViewNames()`?

yields (approved):
LGTM

shlomi-noach:
Added tests look legit

PR #10243, opened 2022-05-09
Fix handling of VISIBLE or INVISIBLE keyword on indexes

dbussink:
MySQL allows for having a VISIBLE or INVISIBLE keyword on an index definition. The default is VISIBLE, but INVISIBLE allows for adding an index that is then not used by the planner.

This wasn't supported in the sqlparser and this change adds support for it to make sure it can be properly parsed.

## Related Issue(s)

Found when working on #10203 

## Checklist


go/vt/sqlparser/keywords.go
```diff
@@ -301,6 +301,7 @@ var keywords = []keyword{
 	{"insensitive", UNUSED},
 	{"insert", INSERT},
 	{"insert_method", INSERT_METHOD},
+	{"invisible", INVISIBLE},
```
dbussink, line 304: It looks like this keyword and `visible` as well are already defined in the parser so they don't need to be updated there.

GuptaManan100 (approved):

PR #10245, opened 2022-05-10
Fix formatting of character set annotation on string literals

dbussink:
Like other character sets, these are lowercase in canonical MySQL output, so that should be matched here.

## Related Issue(s)

#10203 

## Checklist


shlomi-noach (approved):
:pear:

GuptaManan100 (approved):

PR #10246, opened 2022-05-10
schemadiff: use ParseStrictDDL whenever expecting a DDL statement

shlomi-noach:
## Description

`sqlparser.ParseStrictDDL()` is safer to use than `sqlparser.Parse()` when the input is known to be a `DDLStatement`. `ParseStrictDDL()` rejects an incomplete parse. This is the behavior we want in `schemadiff` because `schemadiff` needs to have a well formed AST, manipulate it, validate it, etc.

## Related Issue(s)

- tracking: #10203 

## Checklist


dbussink (approved):

shlomi-noach:
The test `Cluster (shardedrecovery_stress_verticalsplit_heavy)` never ends. I tried cancelling and restarting it, same. It is not required, and so I'm merging.

PR #10236, opened 2022-05-09
fix endtoend onlineddl flakyness caused by --heartbeat_on_demand_duration

shlomi-noach:
## Description

Looking into flakiness introduced by https://github.com/vitessio/vitess/pull/10198
I first want to unblock pending PRs; will then investigate to see why this is caused.

## Related Issue(s)

https://github.com/vitessio/vitess/pull/10198

## Checklist

-   [ ] "Backport me!" label has been added if this change should be backported
-   [x] Tests were added or are not required
-   [ ] Documentation was added or is not required

PR #10198, opened 2022-05-03
On demand heartbeats via `--heartbeat_on_demand_duration`, used by the tablet throttler

shlomi-noach:
## Description

Fixes https://github.com/vitessio/vitess/issues/10196, and see the general description of the problem and suggested fix in https://github.com/vitessio/vitess/issues/10196.

This PR introduces a new command line flag, `--heartbeat_on_demand_duration`. By default it's zero, which means it's disabled and has no effect (no change in behavior).

When positive, e.g. `--heartbeat_on_demand_duration=5s`, and assuming `--heartbeat_enable` is set, then the tablet does not in fact start generating heartbeats on startup. Instead, it only generates heartbeats on a lease per demand.

In this PR the tablet throttler is currently the only entity which asks the heartbeats writer to generate on demand heartbeats. Right now, the tablet throttler makes such requests _any time_ anyone runs a throttler `check`. Normally, no one does, and so the throttler does not make any request for heartbeats, and so no heartbeats are generated.

However, whenever a `vreplication` workflow begins, or whenever an Online DDL operation begins, the throttler is called (assuming `--enable_lag_throttler` is enabled, of course), and in turn the throttler asks the heartbeat writer to generate heartbeats.

Once the workflow or migration complete, and assuming no other workflows/migrations running, the throttler stops asking for heartbeats, the leas expires, the heartbeat writer ceases to write heartbeats.

The effect is that we keep the binary logs small, as we don't write heartbeats when those are not strictly needed.

As tests go, I've added `--heartbeat_on_demand_duration=5s` to all `endtoend/onlineddl` tests where `--enable_lag_throttler` is set; a couple `endtoend` where `--heartbeat_on_demand_duration=5s` is set are left without  `--heartbeat_on_demand_duration`.

## Related Issue(s)

https://github.com/vitessio/vitess/issues/10196

## Checklist

-   [ ] "Backport me!" label has been added if this change should be backported
-   [x] Tests were added or are not required
-   [ ] Documentation was added or is not required

PR #10247, opened 2022-05-10
Fix to --heartbeat_on_demand_duration race condition

shlomi-noach:
## Description

#10198 introduced `--heartbeat_on_demand_duration` flag. The flag was added to multiple `onlineddl` `endtoend` testing workflows. A couple, but specifically `onlineddl_vrepl` started exhibiting recurring failures.

The flag is opt-in so no hard done to anyone. So no pressure other than CI jobs failing. https://github.com/vitessio/vitess/pull/10236 was first merged to revert the CI changes.

I was happy to then be able to reproduce the issue on a local dev env. As always with GitHub CI, it's about thresholds and execution speed, and I was able to figure out a configuration setup that makes the problem reproducible.

I'm also happy to have pinpointed the problem. There was a race condition.

- on one hand, `RequestHeartbeats()` calls were selectively accepted/dropped - as expected
- on the other hand, an unrelated call to `Close()` stopped heartbeats
- leading to the situation where the heartbeat writer thought it should keep rejecting calls to `RequestHeartbeats()` (based on thinking it just recently let a request through, so why bother letting another one), while at the same time heartbeats were disabled. Thus, the heartbeat writer was not producing any heartbeats when it should have.

The solution: any turning off of heartbeats now explicitly ensure the next call to `RequestHeartbeats()` is allowed to pass through. Another change is that any time a request does pass through, it just goes ahead and enables the heartbeats regardless of how many such requests are pending.

Anyway. Re-added `--heartbeat_on_demand_duration` in `onlineddl_vrepl` and in `onlineddl_ghost`. This is tested not only with normal migrations, but also with tablet reparenting. 

## Related Issue(s)

- https://github.com/vitessio/vitess/pull/10236
- #10198 

## Checklist

### DIFF
--- a/go/test/endtoend/onlineddl/vrepl/onlineddl_vrepl_test.go
+++ b/go/test/endtoend/onlineddl/vrepl/onlineddl_vrepl_test.go
@@ -171,6 +171,7 @@ func TestMain(m *testing.M) {
 			"--throttle_threshold", "1s",
 			"--heartbeat_enable",
 			"--heartbeat_interval", "250ms",
+			"--heartbeat_on_demand_duration", "5s",
 			"--migration_check_interval", "5s",
 		}
 		clusterInstance.VtGateExtraArgs = []string{
--- a/go/test/endtoend/onlineddl/vrepl/onlineddl_vrepl_test.go
+++ b/go/test/endtoend/onlineddl/vrepl/onlineddl_vrepl_test.go
@@ -45,6 +45,9 @@ var (
 	httpClient       = throttlebase.SetupHTTPClient(time.Second)
 	throttlerAppName = "vreplication"
 
+	normalMigrationWait   = 20 * time.Second
+	extendedMigrationWait = 20 * time.Second
+
 	hostname              = "localhost"
 	keyspaceName          = "ks"
 	cell                  = "zone1"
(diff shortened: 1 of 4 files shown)


shlomi-noach:
Good to review! Solves a race condition in on-demand heartbeats. Updated original comment.

go/test/endtoend/onlineddl/vrepl/onlineddl_vrepl_test.go
```diff
@@ -171,6 +174,7 @@ func TestMain(m *testing.M) {
 			"--throttle_threshold", "1s",
 			"--heartbeat_enable",
 			"--heartbeat_interval", "250ms",
+			"--heartbeat_on_demand_duration", "5s",
```
shlomi-noach, line 177: re-adding `--heartbeat_on_demand_duration`; the most interesting scenario in this `endtoend` suite is a PRS that runs throughout a throttled migration.

go/test/endtoend/onlineddl/vrepl/onlineddl_vrepl_test.go
```diff
@@ -45,6 +45,9 @@ var (
 	httpClient       = throttlebase.SetupHTTPClient(time.Second)
 	throttlerAppName = "vreplication"
 
+	normalMigrationWait   = 20 * time.Second
+	extendedMigrationWait = 20 * time.Second
```
shlomi-noach, line 49: The increased timeouts are not strictly relevant to fixing the race condition, but rahter just to allow for some GitHub CI slowness.

rohit-nayak-ps (approved):
lgtm

Issue #10193, opened 2022-05-02
Bug Report: `vtcombo` does not respect `-mysql_server_version`


aquarapid:
### Overview of the Issue

`vtcombo` (and possibly other components) relies on `-mysql_server_version` to setup collation code appropriately (MySQL 5.7 vs 8.0).  See this code:

https://github.com/vitessio/vitess/blob/main/go/mysql/collations/local.go#L32-L41

Unfortunately, this does not work correctly if you do not have MySQL 8.0.  This is because the `collation` module is init-ed via the module init from `planbuilder`;  and thus runs long before the CLI flags (including `-mysql_server_version`) is parsed via `flag.Parse()`

### Reproduction Steps

 * Instrument https://github.com/vitessio/vitess/blob/main/go/mysql/collations/local.go#L32-L41 with a print statement to print the value of `*servenv.MySQLServerVersion` when it is used.
 * Run vtcombo, providing `-mysql_server_version` of (for example) `5.7.30-vitess`.
 * Observe from the output of the print statement that the value of `*servenv.MySQLServerVersion` is the default value (empty string) at the time the collation code in `local.go` is run.

### Binary Version

```sh
Recent main:  31c9780fd396612c7427c57e240f8b83c8b6f3ca
```

### Operating System and Environment details

```sh
Fedora 35, x86_64.
```

### Log Fragments

```sh
Instrumenting the code in `local.go` to print a backtrace, shows the init code path that results in the `collation` module being initialized (early), before `flag.Parse()` runs:

 * https://github.com/vitessio/vitess/blob/main/go/vt/vtgate/planbuilder/system_variables.go#L30 to
 * https://github.com/vitessio/vitess/blob/main/go/vt/vtgate/planbuilder/system_variables.go#L52 to
 * https://github.com/vitessio/vitess/blob/main/go/vt/vtgate/planbuilder/system_variables.go#L65 to
 * https://github.com/vitessio/vitess/blob/main/go/vt/vtgate/evalengine/translate.go#L571 to
 * https://github.com/vitessio/vitess/blob/main/go/vt/vtgate/evalengine/translate.go#L562 to
 * https://github.com/vitessio/vitess/blob/main/go/mysql/collations/local.go#L47 to
 * https://github.com/vitessio/vitess/blob/main/go/mysql/collations/local.go#L34
```

aquarapid:
[...]

PR #10248, opened 2022-05-10
plancache: Lazy sysvar planner functions

vmg:
## Description

Fixes https://github.com/vitessio/vitess/issues/10193

The collations `Local()` code was never designed to be called early; it needs to be called after the current Vitess process has been fully parsed. As seen on #10193, the `planbuilder` code was accessing the local collation environment to pre-cache some sysvar plan functions during `init()`, which was causing the wrong version of the collations environment to be loaded for the duration of the process.

Fix this by refactoring the `planbuilder` code to only pre-cache the sysvar plan helpers _once_, when they're actually used for the first time. Also, update the `collations` package so it crashes if somebody calls it too early.

## Checklist


systay (approved):

go/vt/vtgate/planbuilder/system_variables.go
```diff
@@ -18,26 +18,23 @@ package planbuilder
...
-func init() {
-	forSettings(sysvars.ReadOnly, buildSetOpReadOnly)
-	forSettings(sysvars.IgnoreThese, buildSetOpIgnore)
-	forSettings(sysvars.UseReservedConn, buildSetOpReservedConn)
-	forSettings(sysvars.CheckAndIgnore, buildSetOpCheckAndIgnore)
-	forSettings(sysvars.NotSupported, buildNotSupported)
-	forSettings(sysvars.VitessAware, buildSetOpVitessAware)
+type sysvarPlanCache struct {
+	funcs map[string]planFunc
+	once  sync.Once
```
mattlord, line 32: NICE! TiL: https://medium.com/easyread/just-call-your-code-only-once-256f69ed39a8

PR #10249, opened 2022-05-10
Fix parsing of encryption attribute for create database

dbussink:
The create database statement also allows for an encryption value to be set, so this fixes parsing support for that.

It also renames the `CharsetAndCollation` to `DatabaseOption` since `CharsetAndCollationAndEncryption` seems overdoing things and `DatabaseOption` more matches how we call this in other places, like a `TableOption`.

Updates the test to parse unconditional with no magic comments so we don't depend on the version and the comment logic is already tested separately somewhere else, so I think it's better to not mix those here.

## Related Issue(s)

Found as part of #10203 

## Checklist


GuptaManan100 (approved):
Changes look good to me!

GuptaManan100 (commented):
I have looked at the shift-reduce conflict that is being generated because of these changes. Essentially they are coming from `ALTER DATABASE` command. In this command the database name is optional and if it is not provided, then we use the current database in use. 

Since Encryption is a non-reserved keyword, on `ALTER DATABASE ENCRYPTION` there is a shift-reduce conflict. Reduction would mean that we are assuming we have no database name provided, shifting would mean that we are using ENCRYPTION as the database name.

The full parsing info for it - 
```
ALTER {DATABASE | SCHEMA} [db_name]
    alter_option ...

alter_option: {
    [DEFAULT] CHARACTER SET [=] charset_name
  | [DEFAULT] COLLATE [=] collation_name
  | [DEFAULT] ENCRYPTION [=] {'Y' | 'N'}
  | READ ONLY [=] {DEFAULT | 0 | 1}
}
```

GuptaManan100:
We can remove the shift-reduce conflict using precedence rules, but not doing so is also fine because this is a one-off conflict and we already have 1 in our parser. We should document this however, so that later if we look at the conflicts, we know where it is coming from.

To fix the CI, the expectedOutput in `check_make_parser.sh` file needs to be fixed

GuptaManan100:
Shift-reduce conflicts default to shifting and I can verify MySQL does the same - 
```mysql
mysql [localhost:8026] {msandbox} (rails_app) > alter database encryption = 'y';
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '= 'y'' at line 1
mysql [localhost:8026] {msandbox} (rails_app) > alter database encryption encryption = 'y';
ERROR 3503 (42Y07): Database 'encryption' doesn't exist
```

dbussink:
> To fix the CI, the expectedOutput in `check_make_parser.sh` file needs to be fixed

Applied in https://github.com/vitessio/vitess/pull/10249/commits/6a49045fe20467a8d8a3ba14fd87b914d1967c90 and added a comment there as well for this case.

PR #10250, opened 2022-05-10
Add support for parsing MATCH in a foreign key definition

dbussink:
The parser right now lacks support for parsing the optional MATCH clause in a foreign key reference definition.

This adds support for it, see also https://dev.mysql.com/doc/refman/8.0/en/create-table.html for where this is specified.

## Related Issue(s)

Found as part of #10203

## Checklist


go/vt/sqlparser/parse_test.go
```diff
@@ -3916,60 +3916,72 @@ func TestCreateTable(t *testing.T) {
...
-	newCol int references simple (a) on delete no action,
-	newCol int references simple (a) on delete cascade on update set default,
-	newCol int references simple (a) on delete set default on update set null,
-	newCol int references simple (a) on delete set null on update restrict,
-	newCol int references simple (a) on update set default on delete cascade,
-	newCol int references simple (a) on update set null on delete set default,
-	newCol int references simple (a) on update restrict on delete set null,
-	newCol int references simple (a) on update no action,
-	newCol int references simple (a) on update cascade,
+	newCol int references t2 (a),
```
dbussink, line 3919: I've renamed there referenced table here since `simple` is now treated as a keyword and always escaped (since there's a `MATCH SIMPLE` and Vitess pre-emptively escapes any keyword). 

Adding the backtick here to escape was annoying since it's a string defined with backticks so it seemed easier to rename the referenced table for these tests.

GuptaManan100 (approved):
😍

go/vt/sqlparser/parse_test.go
GuptaManan100: Yes, this is okay 👍. We were earlier using these as table names in tests to check that they can indeed be used for table names and are marked non_reserved. But we now have an alternate, much more comprehensive test in `TestCompatibility` in keywords_test.go

PR #10252, opened 2022-05-10
evalengine: proper float to integer conversions

vmg:
## Description

Did you know that `float64` to `uint64` conversion is implementation defined behavior in Go? That's right. `float64(-2.0)` returns different results between x64 and an M1 Mac using ARM64. @systay noticed this.

Fortunately for us it is also undefined behavior in C and C++ so I've reverse engineered the trick they're using there by trial and error: it's the most obvious solution. Explained below:

```
		// We want to convert a float64 to its uint64 representation.
		// However, the cast `uint64(f)`, when f < 0, is actually implementation-defined
		// behavior in Go, so we cannot use it here.
		// The most noticeable example of this are M1 Macs with their ARM64 chipsets, where
		// the underflow is clamped at 0:
		//
		//		GOARCH=amd64 | uint64(-2.0) == 18446744073709551614
		// 		GOARCH=arm64 | uint64(-2.0) == 0
		//
		// The most obvious way to keep this well-defined is to do a two-step conversion:
		//		float64 -> int64 -> uint64
		// where every step of the conversion is well-defined. However, this conversion overflows
		// a range of floats, those larger than MaxInt64 but that would still fit in a 64-bit unsigned
		// integer. What's the right way to handle this overflow?
		//
		// Fortunately for us, the `uint64(f)` conversion for negative numbers is also undefined
		// behavior in C and C++, so MySQL is already handling this case! From running this
		// integration test, we can verify that MySQL is using a two-step conversion and it's clamping
		// the value to MaxInt64 on overflow:
		//
		//		mysql> SELECT CAST(18446744073709540000e0 AS UNSIGNED);
		//		+------------------------------------------+
		//		| CAST(18446744073709540000e0 AS UNSIGNED) |
		//		+------------------------------------------+
		//		|                      9223372036854775807 |
		//		+------------------------------------------+
		//
```

## Checklist

### DIFF
--- a/go/vt/vtgate/evalengine/eval_result.go
+++ b/go/vt/vtgate/evalengine/eval_result.go
@@ -760,8 +760,39 @@ func (er *EvalResult) makeUnsignedIntegral() {
 	case sqltypes.IsIntegral(tt):
 		er.type_ = int16(sqltypes.Uint64)
 	case sqltypes.IsFloat(tt):
+		// We want to convert a float64 to its uint64 representation.
+		// However, the cast `uint64(f)`, when f < 0, is actually implementation-defined
+		// behavior in Go, so we cannot use it here.
+		// The most noticeable example of this are M1 Macs with their ARM64 chipsets, where
+		// the underflow is clamped at 0:
+		//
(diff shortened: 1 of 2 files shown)


systay (approved):
❤️

PR #10253, opened 2022-05-10
Handle additional DDL in column definitions

dbussink:
This adds support for handling additional options in DDL for column definitions.

Newer MySQL versions add a few things here like the `VISIBLE` / `INVISIBLE` attributes for columns.

Only adds support to the parser, so there's no behavior change otherwise except for making sure we can parse more from what MySQL defines.

## Related Issue(s)

Part of completing `CREATE TABLE` support for #10203

## Checklist


go/vt/sqlparser/ast_format.go
```diff
@@ -665,6 +665,27 @@ func (ct *ColumnType) Format(buf *TrackedBuffer) {
...
+				buf.astPrintf(ct, " %s", keywordStrings[VISIBLE])
+			}
+		}
+		switch ct.Options.Format {
+		case FixedFormat:
+			buf.astPrintf(ct, " %s %s", keywordStrings[COLUMN_FORMAT], keywordStrings[FIXED])
+		case DynamicFormat:
+			buf.astPrintf(ct, " %s %s", keywordStrings[COLUMN_FORMAT], keywordStrings[DYNAMIC])
+		case DefaultFormat:
+			buf.astPrintf(ct, " %s %s", keywordStrings[COLUMN_FORMAT], keywordStrings[DEFAULT])
```
GuptaManan100, line 681: We could create a ToString function in ast_funcs file and use that here. The benefit is that if we use this column_format field in some other place, we won't be repeating this logic

GuptaManan100 (approved):
Everything else looks good to me! 🚀

go/vt/sqlparser/ast_format.go
dbussink: @GuptaManan100 Done!

PR #10256, opened 2022-05-10
Check for vtadmin JS types drift in CI

rohit-nayak-ps:
## Description

This pull request adds the `check_make_vtadmin_web_ proto` workflow to the CI. The workflow checks that the generated vtadmin javascript types files are in sync with the proto files. This is done by running the `make vtadmin_web_proto_types` command and diffing `git status`'s output before and after.

If there is a diff (a drift), the workflow lists the files that are not in accordance: in this case the generated vtadmin proto files.

I manually tested this action by adding a dummy field to a proto and testing it both locally and in CI.

## Related Issues

#9001

## Checklist

### DIFF
--- a/.github/workflows/check_make_vtadmin_web_proto.yml
+++ b/.github/workflows/check_make_vtadmin_web_proto.yml
@@ -23,6 +23,7 @@ jobs:
             - 'Makefile'
             - 'go/vt/proto/**'
             - 'proto/*.proto'
+            - 'web/vtadmin/src/proto/**'
 
     - name: Set up Go
       uses: actions/setup-go@v2
--- a/.github/workflows/check_make_vtadmin_web_proto.yml
+++ b/.github/workflows/check_make_vtadmin_web_proto.yml
@@ -1,0 +1,48 @@
+name: check_make_vtadmin_web_proto
+on: [push, pull_request]
+jobs:
+
+  build:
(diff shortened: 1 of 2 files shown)


.github/workflows/check_make_vtadmin_web_proto.yml
```diff
@@ -0,0 +1,48 @@
...
+    - name: Check out code
+      uses: actions/checkout@v2
+
+    - name: Check for changes in relevant files
+      uses: frouioui/paths-filter@main
+      id: changes
+      with:
+        token: ''
+        filters: |
+          proto_changes:
```
doeg, line 18: It might be worth adding `web/vtadmin/src/proto` to this list, to catch potentially weird scenarios where the .js is updated but the .proto is not.

doeg (commented):
Thanks so much for taking this one! Looks great to me! Just one small comment since I ran into a similar thing on https://github.com/vitessio/vitess/pull/10214.

.github/workflows/check_make_vtadmin_web_proto.yml
rohit-nayak-ps: excellent point, added, thanks!

doeg (approved):
Thank you again! I really appreciate it. :)

PR #10257, opened 2022-05-10
fix: concatenate engine primitive

harshit-gangal:
## Description
Engine primitives communicate with each other all the time. When they do, the context passed between them is not passed explicitly as a method call argument as is normally done in golang code. Instead it's baked into one of the fields of the `vcursor` instance that engine primitives receive when executing.

The problem with this setup is when we concurrently call multiple children, like UNION does. The way that the context was shared between primitives means that sometimes, by mistake, one primitive would cancel another primitives context.

This manifested as intermittent errors when doing queries containing lots of UNION code.

## Checklist


harshit-gangal (commented):
-

systay (approved):

frouioui (approved):

PR #10262, opened 2022-05-11
evalengine: CASE expressions

systay:
## Description
This PR adds support in the evalengine for `CASE` expressions.


vmg:
Added support for the alternate form of `CASE`. Our version does not match MySQL's behavior fully because I suspect MySQL has a bug here. I've reported upstream: https://bugs.mysql.com/bug.php?id=107287

harshit-gangal (approved):

PR #10264, opened 2022-05-11
fix: vtexplain commit order flaky

harshit-gangal:
## Description

In vtExplain output, commit order depends upon the shard order that gets added to shard session first. This is non-deterministic. So added logic to sort the shard session lexicographically.

## Checklist


dbussink (approved):
Yes please! I've hit this one a lot already today :smile:.

Issue #8751, opened 2021-08-31
[vtadmin] UID in tablet alias is missing leading zeroes


doeg:
When filing a bug, please include the following headings if
possible. Any example text in this template can be deleted.

#### Overview of the Issue

In VTAdmin (both the API and the UI), tablet aliases like `some_cell-0300012345` are displayed as `some_cell-300012345`. 

VTAdmin generally looks at the `Alias` property of a tablet, which uses a uint32 for the UID, from the `Tablet` proto definition from [topodata.proto](https://github.com/tinyspeck/vitess/blob/1acb6e0/proto/topodata.proto#L62-L70): 

```
// TabletAlias is a globally unique tablet identifier.
message TabletAlias {
  // cell is the cell (or datacenter) the tablet is in
  string cell = 1;

  // uid is a unique id for this tablet within the shard
  // (this is the MySQL server id as well).
  uint32 uid = 2;
}
```

For the example above, this would look like:

```
  "tablet": {
    "alias": {
      "cell": "some_cell",
      "uid": 300012345
    },
```

The tablet debug page, however, shows the correct string (with the leading zeroes) [here](https://github.com/tinyspeck/vitess/blob/main/go/cmd/vttablet/status.go#L55):

```
      Alias: {{github_com_vitessio_vitess_vtctld_tablet .Tablet.AliasString}}<br>
```

... which is defined [here](https://github.com/tinyspeck/vitess/blob/47611bc/go/vt/topo/tablet.go#L178), and likely something we can use to add an `AliasString` property to vtadmin.proto's Tablet definition [here](https://github.com/vitessio/vitess/blob/f3c4dbc/proto/vtadmin.proto#L142-L154).

We'll also need to update the front-end to use the new `AliasString` property most (all?) places.

PR #10266, opened 2022-05-11
[vtadmin] Add "Replication Status" view to Tablet page

### LINKED ISSUE #8719: [vtadmin] View shard replication positions
# vtctld2 UI

Found at http://localhost:15000/app/shard?keyspace=commerce&shard=0 > menu > shard replication positions

Screen Shot 2021-08-27 at 6 18 02 PM
Screen Shot 2021-08-27 at 6 18 03 PM
Screen Shot 2021-08-27 at 6 19 04 PM

doeg:
## Description

T'isn't the prettiest, by any means. I have plans to enhance this with a nice tabular display, but this is on par with the vtctld2 UI so we can call it ✨ an MVP. ✨ 

The biggest divergence, in the parity sense, is that the vtctld2 UI displays replication positions for an entire shard, on its equivalent of the shard detail page. It also requires several clicks since it's hidden behind a modal and a confirmation. 

image

One thing the vtctld2 UI does that is nice is show the command it uses. Will have to think about a nice way to show that in VTAdmin overall, since we could apply that pattern in many places.

Anyway. Here's what this looks like for RDONLY/REPLICA tablets:

Screen Shot 2022-05-11 at 7 09 49 AM

And a primary:

Screen Shot 2022-05-11 at 7 09 44 AM

And finally, when a tablet doesn't have any replication status. In this case, the tablet is the primary tablet in a shard that is messed up. (Specifically, the RDONLY and REPLICA tablets came up as non-serving when I ran the local example, which was not on purpose, but useful in this instance!) 

Screen Shot 2022-05-11 at 7 09 38 AM

## Related Issue(s)

Closes #8719

`formatPaddedAlias` was added as a stopgap until we get around to https://github.com/vitessio/vitess/issues/8751. It was required for this PR since `shard_replication_positions` keys on the _padded_ alias, like this:

```js
{
  "result": {
    "replication_positions": [
      {
        "cluster": {
          "id": "local",
          "name": "local"
        },
        "keyspace": "customer",
        "shard": "0",
        "position_info": {
          "replication_statuses": {
            "zone1-0000000200": { // <--- like this
              "position": "MySQL56/97b40be6-d08a-11ec-ba75-6e8242c51720:1-82"
            },
        // etc. 
```

## Checklist

## Deployment Notes

N/A

### DIFF
--- a/web/vtadmin/src/api/http.ts
+++ b/web/vtadmin/src/api/http.ts
@@ -481,8 +481,6 @@ export const fetchShardReplicationPositions = async ({
     keyspaces = [],
     keyspaceShards = [],
 }: FetchShardReplicationPositionsParams) => {
-    // As an easy enhancement for later, we can also validate the request parameters on the front-end
-    // instead of defaulting to '', to save a round trip.
     const req = new URLSearchParams();
     clusterIDs.forEach((c) => c && req.append('cluster', c));
     keyspaces.forEach((k) => k && req.append('keyspace', k));
--- a/web/vtadmin/src/api/http.ts
+++ b/web/vtadmin/src/api/http.ts
@@ -469,3 +469,28 @@ export const validateVersionKeyspace = async ({ clusterID, keyspace }: ValidateV
 
     return vtctldata.ValidateVersionKeyspaceResponse.create(result);
 };
+
+export interface FetchShardReplicationPositionsParams {
+    clusterIDs?: (string | null | undefined)[];
+    keyspaces?: (string | null | undefined)[];
+    keyspaceShards?: (string | null | undefined)[];
+}
+
+export const fetchShardReplicationPositions = async ({
+    clusterIDs = [],
+    keyspaces = [],
+    keyspaceShards = [],
+}: FetchShardReplicationPositionsParams) => {
+    // As an easy enhancement for later, we can also validate the request parameters on the front-end
+    // instead of defaulting to '', to save a round trip.
+    const req = new URLSearchParams();
+    clusterIDs.forEach((c) => c && req.append('cluster', c));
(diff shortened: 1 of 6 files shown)


ajm188 (approved):

derekperkins:
On the "no replication status" screenshot, is there enough information to be explicit about it being broken? My first thought seeing that screen would be that the api call to get replication status failed, not that it succeeded and was showing the actual lack of data.

ajm188:
you won't see anything at all (not even that message) if the RPC failed, from my read [here](https://github.com/vitessio/vitess/pull/10266/files#diff-b38fc60209d0506f26223f131556b50c536bec3c05d95c185f6dfb10b9415295R57-R63)

derekperkins:
Awesome, as long as consistently in the UI we can trust that empty is different from error

doeg:
Yep, Andrew's right! `QueryErrorPlaceholder` will render the error whenever the query itself fails (i.e., 4xx or 5xx error). Here's what that looks like; usually there's actual error messaging of course:

Screen Shot 2022-05-12 at 10 23 05 AM

The "no replication status" placeholder is shown when the query succeeds but there are no entries for that tablet, which I expect is a rather rare occurrence in practice:

```js
 "position_info": {
          "replication_statuses": {
            "zone1-0000000100": {
              "position": "MySQL56/59867adc-cd44-11ec-a410-7a3bbaf28ff6:1-59,6119432e-cd44-11ec-90a3-f8341ef67b83:1-6",
              // ...
            },
            "zone1-0000000101": {}, // <--- empty
            // etc.
```

Happy to take any suggestions to update the wording, if it's ambiguous! I agree that it'd be nice to offer more information in this case 🤔 but I think that might require some back-end plumbing.

PR #10267, opened 2022-05-11
vt/sqlparser: support SRID

yields:
## Description

This PR adds support for parsing `SRID` attributes on columns, we've noticed that the parser errors when it encounters a column definition like the one below.

```sql
`location` point /*!80003 SRID 4326 */ DEFAULT NULL,
```

The changes only parse the `SRID` attribute, we don't do anything with it.

## Checklist


dbussink (approved):
Looks good to me!

go/vt/sqlparser/parse_test.go
```diff
@@ -3862,6 +3862,13 @@ func TestCreateTable(t *testing.T) {
 	username varchar,
 	unique key by_username (username) engine_attribute '{}' secondary_engine_attribute '{}',
 	unique index by_username3 (username)
+)`,
+		},
+		// test defining SRID
+		{
+			input: `create table t (
+	p point srid 0,
+	g geometry not null srid 4326
```
shlomi-noach, line 3871: :+1:

shlomi-noach (approved):
LGTM - nice work! I think the change is simple enough that I can approve it; I'll defer to @GuptaManan100 or others just to be on the safe side, unless they're unavailable to review.

PR #10268, opened 2022-05-11
Improve dealing with check constraints

dbussink:
This fixes a number of issues for check constaints. First it allows parsing both `DROP CONSTRAINT` and `DROP CHECK` in the parser so we can correctly drop an existing constraint. This was missing so far and only a foreign key constraint could be dropped.

Secondly, we improve dealing with constraints in schemadiff as well. Normalization there needs to generate names for constraints much like how we generate names for keys as well.

We also need to teach the diffing logic then how to create the proper `DROP` statements instead of a `DROP FOREIGN KEY` for all constraints that get dropped.

## Related Issue(s)

Part of #10203

## Checklist


dbussink:
There's one specific diffing optimization for check constraints that isn't implemented yet that I wanted to look at separately, which is for enabling / disabling the `ENFORCED` flag on an existing constraint which can be changed with an `ALTER TABLE ALTER CONSTRAINT <name> [ENFORCED | NOT ENFORCED]` DDL style statement.

go/vt/schemadiff/table.go
```diff
@@ -441,6 +442,35 @@ func (c *CreateTableEntity) normalizeUnnamedKeys() {
...
+		if name := constraint.Name.String(); name == "" {
+			nameFormat := "%s_chk_%d"
+			if _, fk := constraint.Details.(*sqlparser.ForeignKeyDefinition); fk {
+				nameFormat = "%s_ibfk_%d"
+			}
+			suggestedCheckName := fmt.Sprintf(nameFormat, c.CreateTable.Table.Name.String(), 1)
+			// now let's see if that name is taken; if it is, enumerate new news until we find a free name
+			for enumerate := 2; constraintNameExists[suggestedCheckName]; enumerate++ {
+				suggestedCheckName = fmt.Sprintf(nameFormat, c.CreateTable.Table.Name.String(), enumerate)
+			}
```
shlomi-noach, line 466: is there a maxlength handling? We've seen in normal keys, that the name for key#2 and onwards is a 30 characters prefix, then padded with `_2`... etc.

shlomi-noach: @dbussink sounds good.

shlomi-noach (approved):

PR #10269, opened 2022-05-11
Add parsing and schemadiff support for ALTER TABLE ALTER CHECK

dbussink:
With the checks support in newer MySQL, there was also support added for being able to alter a check.

The support is specifically limited to flip the enforced toggle on a check, so this adds the minimal support for that across the board.

First the parser needs to be tought about this change so we can parse the ALTER TABLE query properly when seen.

We also update schemadiff to generate the more optimal ALTER TABLE ALTER CHECK if it sees that the check expression is the same and only enforced is changed.

## Related Issue(s)

Part of #10203 to improve schemadiff

## Checklist


shlomi-noach:
Wasn't even aware `ALTER TABLE ALTER CHECK` existed

shlomi-noach (approved):
LGTM

PR #10271, opened 2022-05-11
Fix `tabletmanager_throttler` flakyness: increase wait time

shlomi-noach:
## Description

Quick fix to `tabletmanager_throttler` test that increases a timeout to reduce flakiness. The change to timeout is legit an immaterial to the overall test, and can combat some race condition or slowness in some test envs.

## Checklist

### DIFF
--- a/go/test/endtoend/tabletmanager/throttler/throttler_test.go
+++ b/go/test/endtoend/tabletmanager/throttler/throttler_test.go
@@ -73,10 +73,10 @@ var (
 )
 
 const (
-	throttlerInitWait            = 15 * time.Second
+	throttlerInitWait            = 30 * time.Second
 	accumulateLagWait            = 2 * time.Second
-	throttlerRefreshIntervalWait = 12 * time.Second
-	replicationCatchUpWait       = 5 * time.Second
+	throttlerRefreshIntervalWait = 20 * time.Second
+	replicationCatchUpWait       = 10 * time.Second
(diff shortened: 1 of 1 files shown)


go/test/endtoend/tabletmanager/throttler/throttler_test.go
```diff
@@ -73,7 +73,7 @@ var (
 )
 
 const (
-	throttlerInitWait            = 10 * time.Second
+	throttlerInitWait            = 15 * time.Second
```
mattlord, line 76: Any reason not to make it longer? If the init fails, a test still runs for up to 10-20 minutes (1 try, with potentially 1 retry, and the second retry doesn't seem to cause the CI workflow to pass); only then do you know it's failed and all we can do is re-run the workflow and wait again. Another 15 seconds at the beginning of the test feels much better. 🙂 I'd use 30 seconds here and increase the others below as well to 5, 30, 15, 15.

I'll approve so that you're unblocked as this is subjective — and what's here now may be all that's needed. 

Thank you!

shlomi-noach: @mattlord you're right! I've increased this and a couple more timeouts even more.

shlomi-noach:
Fix one thing, break another... Turning into Draft

shlomi-noach:
OK, that last moment increase of timeouts did not consider an expected side-effect, which actually validates the tests :wink: 
Now fixed.

PR #10051, opened 2022-04-06
Address flakiness of vreplication v2 workflow test

mattlord:
## Description
For whatever reasons, we've been seeing more InnoDB Deadlock errors when getting the workflow status for VDiff lately (e.g. [here](https://github.com/vitessio/vitess/runs/5996959498?check_suite_focus=true) and virtually all failures for [the workflow](https://github.com/vitessio/vitess/actions/workflows/cluster_endtoend_vreplication_across_db_versions.yml)). This work adds retry support (up to 10 times) when this occurs, with a sleep of 1 second in between tries to give the database state time to clear up.

> I suspect that some recent changes have made these deadlocks more likely — at least in our VReplication e2e tests where we only have `PRIMARY` tablets and we do a lot of VReplication workflows in quick succession — especially with MySQL 8.0 as the usage of MySQL 8 on the target tablets is the only significant difference between the [across_db_versions test](https://github.com/vitessio/vitess/blob/main/go/test/endtoend/vreplication/vreplication_test.go#L203-L207) and the [basic test](https://github.com/vitessio/vitess/blob/67f656651786035ceb87d5a19caf41e3ab739c24/go/test/endtoend/vreplication/vreplication_test.go#L145-L149).

For example, this command where the MySQL `deadlock detected` was most often hit before and we now see it succeed after retrying (from [this run](https://github.com/vitessio/vitess/runs/5999037974?check_suite_focus=true)):
```
2022-04-12T23:33:52.9832227Z === RUN   TestV2WorkflowsAcrossDBVersions/reshardCustomer3to2SplitMerge/reshard/vdiff
2022-04-12T23:33:52.9832944Z I0412 23:33:11.553832   19813 vtctlclient_process.go:209] Executing vtctlclient with command: vtctlclient --server 127.0.0.1:15999 VDiff -- --tablet_types=primary --source_cell= --format json customer.c4c3 (attempt 1 of 10)
2022-04-12T23:33:52.9833651Z I0412 23:33:13.755188   19813 vtctlclient_process.go:209] Executing vtctlclient with command: vtctlclient --server 127.0.0.1:15999 VDiff -- --tablet_types=primary --source_cell= --format json customer.c4c3 (attempt 2 of 10)
[...]

PR #10275, opened 2022-05-11
CI: Add query retries to vttablet process

mattlord:
## Description

This should reduce some CI flakes due to ephemeral issues. For example, I've periodically seen failures in `TestV2WorkflowsAcrossDBVersions/reshardMerchant3to1Merge` here ([recent failure](https://github.com/vitessio/vitess/runs/6393037092?check_suite_focus=true#step:10:1008)):
```
I0511 17:42:11.027619   17525 vttablet_process.go:466] Executing query select count(*) from _vt.vreplication where workflow = "m3m1" and db_name = "vt_merchant-type" and pos = '' on zone1-2000
1029
    vttablet_process.go:470: Deadlock found when trying to get lock; try restarting transaction (errno 1213) (sqlstate 40001) during query: select count(*) from _vt.vreplication where workflow = "m3m1" and db_name = "vt_merchant-type" and pos = ''
```

That can happen because when we get any error we fail today: https://github.com/vitessio/vitess/blob/c7579d0ab44b2296f6de9f1766898cc24733ff8b/go/test/endtoend/cluster/vttablet_process.go#L466-L471

This PR adds automatic retries to deal with any ephemeral issues like this — similar to what was previously done for the vtctlclient_process commands in the e2e tests in https://github.com/vitessio/vitess/pull/10051. 

## Related Issue(s)
 - https://github.com/vitessio/vitess/pull/10051

## Checklist

### DIFF
--- a/go/test/endtoend/cluster/vttablet_process.go
+++ b/go/test/endtoend/cluster/vttablet_process.go
@@ -414,8 +414,25 @@ func (vttablet *VttabletProcess) QueryTabletWithDB(query string, dbname string) 
 	return executeQuery(conn, query)
 }
 
+// executeQuery will retry the query up to 10 times with a small sleep in between each try.
+// This allows the tests to be more robust in the face of transient failures.
 func executeQuery(dbConn *mysql.Conn, query string) (*sqltypes.Result, error) {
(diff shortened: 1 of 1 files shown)


rohit-nayak-ps (approved):
lgtm

PR #10286, opened 2022-05-12
BugFix: Resolve Recoveries at the end

GuptaManan100:
## Description

Whenever we `AttemptRecoveryRegistration`, we should also be eventually calling `resolveRecovery` when our recovery completes, irrespective of it failing or not.

We weren't doing this in 2 places, `fixReplica` and `fixPrimary`. This caused the following failure - 

- We ran a `fixReplica` failure on tablet A.
- We didn't resolve the recovery on tablet A.
- We called `GracefulPrimaryTakeover` API on tablet A. This API call should succeed as it does not wait for any tablets in active recovery.
- However, this call blocks because we haven't resolved a previous recovery, so it is still assumed to be in progress.
- We wait until `ClearActiveRecoveries` removes all the old recoveries.

This behaviour is incorrect, as we are making an API call wait for the duration of `RecoveryPeriodBlockSeconds` since the last recovery, even though during registration we don't want to fail for them. We should only fail if a recovery is actually in progress.

The fix for this problem is simple, we should be calling `resolveRecovery` at the end of our recovery function. This PR fixes this and refactors the places that call `AttemptRecoveryRegistration` to defer a function call to `resolveRecovery` to ensure that it is always run. (Other than panics ofcourse :3)

## Checklist

### DIFF
--- a/go/vt/orchestrator/logic/topology_recovery.go
+++ b/go/vt/orchestrator/logic/topology_recovery.go
@@ -635,6 +635,11 @@ func recoverPrimaryHasPrimary(ctx context.Context, analysisEntry inst.Replicatio
 		return false, nil, err
 	}
 	log.Infof("Analysis: %v, will fix incorrect primaryship %+v", analysisEntry.Analysis, analysisEntry.AnalyzedInstanceKey)
+	// This has to be done in the end; whether successful or not, we should mark that the recovery is done.
+	// So that after the active period passes, we are able to run other recoveries.
+	defer func() {
+		resolveRecovery(topologyRecovery, nil)
+	}()
 
 	// Reset replication on current primary.
 	_, err = inst.ResetReplicationOperation(&analysisEntry.AnalyzedInstanceKey)
@@ -668,6 +673,12 @@ func recoverDeadPrimary(ctx context.Context, analysisEntry inst.ReplicationAnaly
 		return false, nil, err
 	}
 	log.Infof("Analysis: %v, deadprimary %+v", analysisEntry.Analysis, analysisEntry.AnalyzedInstanceKey)
+	var promotedReplica *inst.Instance
+	// This has to be done in the end; whether successful or not, we should mark that the recovery is done.
+	// So that after the active period passes, we are able to run other recoveries.
+	defer func() {
+		resolveRecovery(topologyRecovery, promotedReplica)
+	}()
 
 	ev, err := reparentutil.NewEmergencyReparenter(ts, tmclient.NewTabletManagerClient(), logutil.NewCallbackLogger(func(event *logutilpb.Event) {
 		level := event.GetLevel()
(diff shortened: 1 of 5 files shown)


go/vt/orchestrator/logic/topology_recovery.go
```diff
@@ -668,6 +673,12 @@ func recoverDeadPrimary(ctx context.Context, analysisEntry inst.ReplicationAnaly
 		return false, nil, err
 	}
 	log.Infof("Analysis: %v, deadprimary %+v", analysisEntry.Analysis, analysisEntry.AnalyzedInstanceKey)
+	var promotedReplica *inst.Instance
+	// This has to be done in the end; whether successful or not, we should mark that the recovery is done.
+	// So that after the active period passes, we are able to run other recoveries.
+	defer func() {
+		resolveRecovery(topologyRecovery, promotedReplica)
```
shlomi-noach, line 680: @GuptaManan100 the way `golang`'s `defer` works is that the value of `promotedReplica` is evaluated _right now_, and so it is `nil`, even if you later change it to something else.
You can use `*(&promotedReplica)` to get the actual value you want.

go/vt/orchestrator/logic/topology_recovery.go
```diff
@@ -1616,6 +1617,12 @@ func GracefulPrimaryTakeover(clusterName string, designatedKey *inst.InstanceKey
 	// a recovery for some time.
 	// Let's audit anything that happens from this point on, including any early return
 	AuditTopologyRecovery(topologyRecovery, fmt.Sprintf("registered recovery on %+v. Recovery: %+v", analysisEntry.AnalyzedInstanceKey, topologyRecovery))
+	var promotedReplica *inst.Instance
+	// This has to be done in the end; whether successful or not, we should mark that the recovery is done.
+	// So that after the active period passes, we are able to run other recoveries.
+	defer func() {
+		resolveRecovery(topologyRecovery, promotedReplica)
+	}()
```
shlomi-noach, line 1625: same comment about `defer`

go/vt/orchestrator/logic/topology_recovery.go
```diff
@@ -1731,6 +1735,13 @@ func electNewPrimary(ctx context.Context, analysisEntry inst.ReplicationAnalysis
 	}
 	log.Infof("Analysis: %v, will elect a new primary: %v", analysisEntry.Analysis, analysisEntry.SuggestedClusterAlias)
 
+	var promotedReplica *inst.Instance
+	// This has to be done in the end; whether successful or not, we should mark that the recovery is done.
+	// So that after the active period passes, we are able to run other recoveries.
+	defer func() {
+		resolveRecovery(topologyRecovery, promotedReplica)
+	}()
```
shlomi-noach, line 1743: same comment about `defer`

shlomi-noach (changes requested):
See inline comment for use of `defer`.

Otherwise: I'm not sure if we need to `resolveRecovery()` at the end of every single operation. If we do that, then what do we even need recovery block period for?
I think that either way, we should at least "pad" the resolve with a few seconds, to allow `vtorc` to get a refreshed view of the topology. It's true that the recovery process will ensure to re-read instance status for affected servers (e.g. the promoted primary), but it will takes `vtorc` a second or two _after that_ to re-map the entire topology and to make a new analysis and get rid of the previous analysis.

go/vt/orchestrator/logic/topology_recovery.go
shlomi-noach: @GuptaManan100 is correct, and the problem is only when running a `defer` statement. The wrapping inside a function is fine :+1:

shlomi-noach (approved):

shlomi-noach:
Converted to draft

shlomi-noach:
Changes look good to me

PR #10288, opened 2022-05-12
Fix parsing the special case convert charset logic

dbussink:
In a `CAST`, there's a special option to use a charset short name, as described in:

https://dev.mysql.com/doc/refman/8.0/en/cast-functions.html

Under "CHAR[(N)] [charset_info]"

This is not however logic that should be applied to all charset options used everywhere and seems specific to just `CAST`.

For example, ASCII only is an alias in that context at least according to the document, also compare to:

https://dev.mysql.com/doc/refman/8.0/en/charset-repertoire.html

There ASCII is described as an explicit charset only allowing the lower 127 bits range and no higher characters, unlike latin1 so it's not alias. It's even explicitly mentioned there it should error.

This also lead to another parsing error, when trying to specify ASCII as the encoding for a table. That error out while it was perfectly valid to do so.

This cleans it up and keeps a separate expression just for `CAST` and uses a simpler generic form for all other places that deal with characters set specifications.

## Related Issue(s)

Found as part of #10203 for a schema which specified `ASCII` as the desired encoding.

## Checklist


go/vt/sqlparser/sql.y
```diff
@@ -5694,7 +5711,16 @@ charset:
...
+| ASCII
+  {
+    // ASCII: Shorthand for CHARACTER SET latin1.
+    $$ = "ascii"
+  }
+| UNICODE
+  {
+    // UNICODE: Shorthand for CHARACTER SET ucs2.
+    $$ = "ucs2"
+  }
```
dbussink, line 5723: These need to be added since a regular charset can also specify these but they'd otherwise be treated as tokens and not strings.

go/vt/sqlparser/sql.y
```diff
@@ -5682,7 +5699,7 @@ match_option:
  }
 
 charset:
-  id_or_var
+  sql_id
```
dbussink, line 5702: This is more consistent with how we parse this in other places, directly allowing non reserved keywords like `ASCII` in this case.

go/vt/sqlparser/sql.y
```diff
@@ -5714,7 +5731,7 @@ convert_type:
   {
     $$ = &ConvertType{Type: string($1), Length: $2}
   }
-| CHAR length_opt charset_opt
+| CHAR length_opt charset_convert_opt
```
dbussink, line 5734: This now uses the separate explicit option that has the conversions documented at https://dev.mysql.com/doc/refman/8.0/en/cast-functions.html

GuptaManan100 (approved):

PR #10290, opened 2022-05-12
[vtadmin] proper alias types for tablet RPCs

ajm188:
## Description

Way back when I added `{Get,Delete}Tablet`, I made the RPC take a tablet alias as a `string`, rather than a `topodatapb.TabletAlias`. I have no idea why. This then became precedent and spread through the rest of the tablet api methods.

The issue is (in addition to being reminiscent of the "oops all strings" vtctl grpc "interface") this makes the gRPC interface to vtadmin clunky to use, for ex if you have a `TabletAlias`, you have to convert it to a `string` in order to invoke an RPC, only for it to get parsed back into a `TabletAlias` on the server side.

This rectifies that.

### Testing

Unit tests still pass (after adjusting the types), and I was able to run `Ping`, `RunHealthCheck`, and `RefreshState` from the vtadmin-web UI.

## Checklist


go/vt/vtadmin/api.go
```diff
@@ -1867,16 +1861,18 @@ func (api *API) getTabletForAction(ctx context.Context, span trace.Span, action
 		go func(c *cluster.Cluster) {
 			defer wg.Done()
 
-			ts, err := c.GetTablets(ctx)
+			ts, err := c.FindTablets(ctx, func(t *vtadminpb.Tablet) bool {
```
doeg, line 1864: Would using `c.FindTablet` instead apply here? If so we might be able to 🔪 some lines.

go/vt/vtadmin/http/request.go
```diff
@@ -73,3 +76,32 @@ func (r Request) ParseQueryParamAsUint32(name string, defaultVal uint32) (uint32
...
+//
+// See (gorilla/mux).Vars for details. We define a type here to add some
+// additional behavior for extracting non-string values.
+type Vars map[string]string
+
+// GetTabletAlias returns the route named `key` as a TabletAlias.
+//
+// It returns an error if the route has no variable with that name, or if it
+// cannot be parsed as a TabletAlias.
+func (v Vars) GetTabletAlias(key string) (*topodatapb.TabletAlias, error) {
```
doeg, line 90: :noice:

doeg (approved):
One non-blocking comment. Appreciate that this didn't require any front-end changes!

go/vt/vtadmin/api.go
doeg: Ohh. Yeah the `found` var is what prompted the question. so that is gr8.

doeg (approved):
Another +1, if you want it!

PR #10294, opened 2022-05-12
VStream API: allow cells to be specified for picking source tablets to stream from

rohit-nayak-ps:
## Description

The VTGate VStreamAPI, internally needs to source data from tablets using the `RowStreamer` and `VStreamer` APIs. It uses a `TabletPicker` to choose a tablet per shard. If a `REPLICA` tablet type is specified, currently the picker looks at tablets in the same cell as the vtgate to choose a replica to source from. However it is possible that there is no replica in the cell of the vtgate one connects to.

This PR adds the ability to specify a list of (comma separated) cell names/cell aliases. Instead of changing the API signature we add a new `Cells` field to the VStreamFlags struct for this purpose. 

Sample usage:
```
options := &vtgatepb.VStreamFlags{
	Cells: "zone1, zone4"
}
reader, err := conn.VStream(ctx, topodatapb.TabletType_REPLICA, vgtid, filter, options)
```

The functionality is validated by a e2e test.

## Checklist


deepthi:
I assume this will also work with cell-aliases?

rohit-nayak-ps:
> I assume this will also work with cell-aliases?

Yes, because it is just calls the tablet picker with the cell names and we already resolve aliases there.

mattlord:
Sorry, @rohit-nayak-ps ! Somehow I completely missed this one. I will review in the next few hours.

go/test/endtoend/vreplication/vreplication_test.go
```diff
@@ -231,6 +237,88 @@ func TestMultiCellVreplicationWorkflow(t *testing.T) {
...
+	vstreamTestCases := []vstreamTestCase{
+		{"zone1,zone2", false},
+		{"zone7", true},
+		{"", false},
+	}
+
+	for _, tc := range vstreamTestCases {
+		t.Run("VStreamCellsFlag/"+tc.cells, func(t *testing.T) {
+			conn, err := vtgateconn.Dial(ctx, fmt.Sprintf("localhost:%d", vc.ClusterConfig.vtgateGrpcPort))
+			if err != nil {
```
mattlord, line 274: Nit, but why not use `require.NoError(t, err)` ?

go/test/endtoend/vreplication/vreplication_test.go
```diff
@@ -231,6 +237,88 @@ func TestMultiCellVreplicationWorkflow(t *testing.T) {
...
+				flags.Cells = tc.cells
+			}
+
+			ctx2, cancel := context.WithTimeout(ctx, 30*time.Second)
+			reader, err := conn.VStream(ctx2, topodatapb.TabletType_REPLICA, vgtid, filter, flags)
+			require.NoError(t, err)
+
+			rowsReceived := false
+			var wg sync.WaitGroup
+			wg.Add(1)
```
mattlord, line 290: Curious what advantages this has when there's only one thread/coroutine and we wait for it?

go/test/endtoend/vreplication/vreplication_test.go
```diff
@@ -231,6 +237,88 @@ func TestMultiCellVreplicationWorkflow(t *testing.T) {
...
+			}()
+			wg.Wait()
+
+			if tc.expectError {
+				require.False(t, rowsReceived)
+
+				// if no tablet was found the tablet picker adds a key which includes the cell name to the vtgate TabletPickerNoTabletFoundErrorCount stat
+				pickerErrorStat, err := getDebugVar(t, vc.ClusterConfig.vtgatePort, []string{"TabletPickerNoTabletFoundErrorCount"})
+				require.NoError(t, err)
+				require.Contains(t, pickerErrorStat, "zone7")
```
mattlord, line 316: We shouldn't we replace the hardcoded `"zone7"` with `tc.cells`?

go/vt/vtgate/vstream_manager.go
```diff
@@ -408,6 +411,21 @@ func (vs *vstream) alignStreams(ctx context.Context, event *binlogdatapb.VEvent,
 	}
 }
 
+func (vs *vstream) getCells() []string {
+	var cells []string
+	if vs.optCells != "" {
+		for _, cell := range strings.Split(strings.TrimSpace(vs.optCells), ",") {
+			cells = append(cells, strings.TrimSpace(cell))
+		}
+	}
```
mattlord, line 420: Wouldn't this one liner be the same?
```
cells := strings.Split(strings.TrimSpace(vs.optCells), ",")
```

Seems that way: https://go.dev/play/p/WwiNpEnhwsJ

IMO that's easier to read and it should be more efficient too.

go/vt/vtgate/vstream_manager.go
```diff
@@ -408,6 +411,21 @@ func (vs *vstream) alignStreams(ctx context.Context, event *binlogdatapb.VEvent,
...
+
+	vtgateCell := vs.vsm.cell
```
mattlord, line 422: Not sure if this is a pointer or not, but any reason not to move this into the conditional and even remove it entirely?
```
 	if len(cells) == 0 {
                # use the vtgate's cell by default
 		return []string{vs.vsm.cell}
 	}
```

So when combined with the previous comment, the function could be shortened to:
```
if cells := strings.Split(strings.TrimSpace(vs.optCells), ","); len(cells) != 0 {
    return cells
} else {
    # use the vtgate's cell by default
    return []string{vs.vsm.cell}
}
```

Nit, but IMO that's simpler for humans and the runtime.

mattlord (approved):
I had some questions and nits, but nothing major. Will approve so that you're not blocked if I misunderstood something or you disagree.

go/test/endtoend/vreplication/vreplication_test.go
rohit-nayak-ps: > Nit, but why not use `require.NoError(t, err)` ?
fixed

go/test/endtoend/vreplication/vreplication_test.go
rohit-nayak-ps: Actually it started off as two subroutines but I slmplified the test and this waitgroup paradigm looked like an elegant-enough solution for the purpose of waiting for a result or a context timeout. Do you see any major issue with it?

go/test/endtoend/vreplication/vreplication_test.go
rohit-nayak-ps: "zone7" is actually an invalid (non-existent cell). Made it more explicit using a variable instead of the literal.

go/vt/vtgate/vstream_manager.go
rohit-nayak-ps: removed the intermediate variable and replaced with suggested comment.

go/test/endtoend/vreplication/vreplication_test.go
mattlord: No, no problem! I was just curious if there was some advantage to doing it that way in this case. No reason not to leave it, as e.g. we may want to use more than 1 goroutine later.

go/vt/vtgate/vstream_manager.go
mattlord: Ah, good point 🙂 you’re right then

PR #10296, opened 2022-05-13
Add additional normalization rules for schemadiff

dbussink:
This adds additional normalization rules for schemadiff to reduce churn around default options and to remove things that are not supported.

- `USING BTREE` on an index is the default, so we drop that if specified.
- `VISIBLE` is the default for columns & indexes, so this is dropped when specified.
- `ALGORITHM = undefined` is the default for a `CREATE VIEW` so this is dropped when specified.
- `SQL SECURITY DEFINER` is the default for a `CREATE VIEW` so this is dropped when specified.

## Related Issue(s)

Part of #10203

## Checklist


GuptaManan100 (approved):
This looks good to me! 🌀

shlomi-noach:
:+1:

PR #10298, opened 2022-05-13
Add a failing test case showing Gen4 planner failure.

arthurschreiber:
## Description

This test case shows a (recent) regression in the query planner for queries that use `IS NULL` condition on an unowned lookup vindex column.

```
=== RUN   TestUnownedLookupSelectNull
    lookup_test.go:88: 
                Error Trace:    utils.go:105
                                                        lookup_test.go:88
                Error:          Received unexpected error:
                                v3 and Gen4 failed with different errors: v3: [lookup.Map: Code: INTERNAL
                                Gen4 failed while v3 did not: target: ks.80-.primary: vttablet: rpc error: code = InvalidArgument desc = missing bind var __vals (CallerID: userData1)
                                ], Gen4: [lookup.Map: Code: INTERNAL
                                Gen4 failed while v3 did not: target: ks.-80.primary: vttablet: rpc error: code = InvalidArgument desc = missing bind var __vals (CallerID: userData1)
                                ] (errno 1815) (sqlstate HY000) during query: select * from t8 WHERE t9_id IS NULL
                Test:           TestUnownedLookupSelectNull
                Messages:       for query: select * from t8 WHERE t9_id IS NULL
--- FAIL: TestUnownedLookupSelectNull (0.12s)
```

## Checklist

### DIFF
--- a/go/test/endtoend/vtgate/lookup_test.go
+++ b/go/test/endtoend/vtgate/lookup_test.go
@@ -77,6 +77,20 @@ func TestUnownedLookupInsertChecksKeyspaceIdsAreMatching(t *testing.T) {
 	utils.Exec(t, conn, "delete from t9 WHERE parent_id = 1")
 }
 
+func TestUnownedLookupSelectNull(t *testing.T) {
+	defer cluster.PanicHandler(t)
+
+	ctx := context.Background()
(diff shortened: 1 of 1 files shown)


arthurschreiber:
This has been superseded by https://github.com/vitessio/vitess/pull/10346. 🙇‍♂️

PR #10299, opened 2022-05-13
Support for Group by all

GuptaManan100:
## Description

This PR adds support for `Group by all` which is a sugar-coating that Vitess now offers. If the user wants to group by all the non-aggregated columns, they can choose to say `GROUP BY ALL` instead of listing out the columns in the group by.

For example, the two following queries are equivalent for Vitess - 
```sql
select id6, id7, count(*) k from t3 group by id6, id7 order by k
```

```sql
select id6, id7, count(*) k from t3 group by all order by k
```

## Checklist

-   [x] "Backport me!" label has been added if this change should be backported No
-   [x] Tests were added or are not required Yes
-   [ ] Documentation was added or is not required Needed


go/vt/vtgate/planbuilder/grouping.go
```diff
@@ -23,8 +23,8 @@ import (
 	"vitess.io/vitess/go/vt/vtgate/engine"
 )
 
-func planGroupBy(pb *primitiveBuilder, input logicalPlan, groupBy sqlparser.GroupBy) (logicalPlan, error) {
-	if len(groupBy) == 0 {
+func planGroupBy(pb *primitiveBuilder, input logicalPlan, groupBy *sqlparser.GroupBy) (logicalPlan, error) {
```
harshit-gangal, line 26: I think we can just send in []Expr here instead of groupBy

go/vt/vtgate/semantics/early_rewriter.go
```diff
@@ -68,8 +68,31 @@ func (r *earlyRewriter) down(cursor *sqlparser.Cursor) error {
...
 	case *sqlparser.Order:
 		r.clause = "order clause"
-	case sqlparser.GroupBy:
+	case *sqlparser.GroupBy:
 		r.clause = "group statement"
+		if node.All {
+			// Get the current scope and find the select statement from it
+			selStmt, isSelect := r.scoper.currentScope().stmt.(*sqlparser.Select)
+			if !isSelect {
+				return vterrors.Errorf(vtrpcpb.Code_INTERNAL, "found a group but the current scope is not a select statement")
```
harshit-gangal, line 77: how about we do this in the normalizer? And from there own we will only have []Expr in the groupBy.

systay: seems reasonable

GuptaManan100: I am going ahead with the changes to move this logic to the normaliser, but just so that all of us are on the same page, it would mean that this syntactic sugar will only work if the vtgate is configured to have normalisation turned on.

GuptaManan100: We have the `MustRewriteAST` function that we can change too, to make the normalizer run always if the `group by all` syntax is used, but I don't think we should do that

GuptaManan100: Doing this change in the normaliser has an added advantage of also working with v3, 😆

GuptaManan100: Done

go/vt/vtgate/planbuilder/grouping.go
GuptaManan100: But then the group by being nil will have to checked outside where the function is called from. Is there a reason why we want to send []Expr instead of a GroupBy

PR #10303, opened 2022-05-14
Mark web/vtadmin/src/proto/** as generated files in .gitattributes

doeg:
Signed-off-by: Sara Bee <855595+doeg@users.noreply.github.com>

## Description

This adds a .gitattributes file and flags the files in [web/vtadmin/src/proto/](https://github.com/vitessio/vitess/tree/main/web/vtadmin/src/proto) as "generated files" to hide them by default in GitHub's diff view. These are the JS/TS bindings produced by `make vtadmin_web_proto_types`, so these files don't require manual review. (Doubly so now that we have CI in place to validate them thanks to #10256!) 

This feature is documented here: https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github

Here's a before and after of what it looks for an unrelated branch of mine:

Screen Shot 2022-05-14 at 9 29 09 AM

Screen Shot 2022-05-14 at 9 22 13 AM

## Related Issue(s)

N/A

## Checklist

-   [x] "Backport me!" label has been added if this change should be backported
-   [x] Tests were added or are not required
-   [x] Documentation was added or is not required

## Deployment Notes

N/A

PR #10306, opened 2022-05-14
Hints to git cli and github diff to hide diffs in proto generated files by default

rohit-nayak-ps:
## Description

Adds a couple of hints to hide the generated protobuf files:

* Adds .gitattributes to the generated proto directories. This tells `git diff` to only inform that the files differ but not show the actual diffs which can be hundreds of lines for some `pb.go` files. The files continue to show in the `git status` output.
* Follow-up of #10303. Adds golang protobuf generated files using `make proto` to the root .gitattributes file to optimize GitHub's diff view.

## Related Issue(s)
#10303 

## Checklist

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -3,3 +3,4 @@ web/vtadmin/src/proto/*.js -diff
 web/vtadmin/src/proto/*.ts -diff
 go/vt/proto/** linguist-generated=true
 go/vt/proto/*/*.pb.go -diff
+go/vt/sqlparser/sql.go linguist-generated=true -diff
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,5 @@
 web/vtadmin/src/proto/** linguist-generated=true
+web/vtadmin/src/proto/*/*.pb.go -diff
 go/vt/proto/** linguist-generated=true
+go/vt/proto/*.js -diff
+go/vt/proto/*.ts -diff
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,5 +1,5 @@
 web/vtadmin/src/proto/** linguist-generated=true
-web/vtadmin/src/proto/*/*.pb.go -diff
+web/vtadmin/src/proto/*.js -diff
+web/vtadmin/src/proto/*.ts -diff
 go/vt/proto/** linguist-generated=true
-go/vt/proto/*.js -diff
-go/vt/proto/*.ts -diff
+go/vt/proto/*/*.pb.go -diff
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,2 @@
 web/vtadmin/src/proto/** linguist-generated=true
+go/vt/proto/** linguist-generated=true
--- a/go/vt/proto/.gitattributes
+++ b/go/vt/proto/.gitattributes
@@ -1,2 +1,0 @@
-.gitattributes
-*/*.pb.go -diff
--- a/go/vt/proto/.gitattributes
+++ b/go/vt/proto/.gitattributes
@@ -1,0 +1,2 @@
+.gitattributes
+*/*.pb.go -diff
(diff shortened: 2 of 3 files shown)


rohit-nayak-ps:
Closing this since this approach doesn't seem to be working for go lang proto files. Even without it the files were already hidden by github. I had expected the statistics that show changed lines to be different, but that is not happening: https://github.com/vitessio/vitess/pull/10306/files

go/vt/proto/.gitattributes
```diff
@@ -0,0 +1,2 @@
+.gitattributes
+*/*.pb.go -diff
```
ajm188, line 2: is there a reason we can't put these in the top `.gitattributes` file? (same question for the web/vtadmin one)

rohit-nayak-ps: I had tried to put the directives in the root `.gitattributes` initially and it had not worked, but with these separate ones it did. 

I was trying to reproduce it now, but looks like git is caching the old files: even after deleting all the `.gitattributes` files in the repo `git diff` is still hiding the generated files! 

Tried using `git rm --cached` on the `.gitattributes` files after starting fresh to see if the cache gets deleted, but it still seems to be using the cached ones.  Will try again tomorrow after rebooting the machine ...

frouioui (commented):
Same question as @ajm188, otherwise it looks good to me

go/vt/proto/.gitattributes
rohit-nayak-ps: Thanks @ajm188 @frouioui . Moved to root level. _Seems_ to work, but I have not been able to consistently work around the cached .gitattributes file. Rebooting didn't help, but if put in incorrect file extensions the generated files do show in `git diff`, so presumably having the hints in the root file should work for everyone.

dbussink:
Can we also add the generated parser code to this? `go/vt/sqlparser/sql.go` is the path.

rohit-nayak-ps:
> Can we also add the generated parser code to this? `go/vt/sqlparser/sql.go` is the path.

Good idea, done.

ajm188 (approved):

.gitattributes
```diff
@@ -1 +1,6 @@
 web/vtadmin/src/proto/** linguist-generated=true
+web/vtadmin/src/proto/*.js -diff
+web/vtadmin/src/proto/*.ts -diff
+go/vt/proto/** linguist-generated=true
+go/vt/proto/*/*.pb.go -diff
```
mattlord, line 5: I like it! Any reason not to use `-merge` for the generated files too? Then the merge will be treated as if they were binary files too and it will just use the entire latest file. I don't think I've seen any merge conflicts with the generated files though... so not likely needed. ❤️

rohit-nayak-ps: Don't get the use case you are referring to: we regenerate these files in its entirety and any merge conflicts would need to be fixed in the .proto file.

Will merge this PR as-is and we can extend this for your use-case separately.

doeg:
Late to the party, but this is really cool!! I love the way this shows up as bytes changed with `git pull`, etc.

```
~/workspace/vitess  🍕 $ git pull
remote: Enumerating objects: 36, done.
remote: Counting objects: 100% (36/36), done.
remote: Compressing objects: 100% (13/13), done.
remote: Total 36 (delta 22), reused 33 (delta 22), pack-reused 0
Unpacking objects: 100% (36/36), 48.42 KiB | 286.00 KiB/s, done.
From github.com:vitessio/vitess
   4fe8c36951..cdd4acdb84  main       -> origin/main
Updating 4fe8c36951..cdd4acdb84
Fast-forward
 examples/local/vtadmin/rbac.yaml                            |   7 ++
 go/vt/proto/vtadmin/vtadmin.pb.go                           | Bin 260809 -> 286029 bytes
 go/vt/proto/vtadmin/vtadmin_grpc.pb.go                      | Bin 64673 -> 71553 bytes
 go/vt/proto/vtadmin/vtadmin_vtproto.pb.go                   | Bin 362903 -> 396785 bytes
```

PR #10307, opened 2022-05-14
Remove usage of additional uuid package

dbussink:
This removes using an additional UUID package. This was the only place this package was used, everywhere else uses github.com/google/uuid. Even this library uses that package as a dependency.

Fewer dependencies is better and also updating github.com/google/uuid while I was at it.

## Checklist


rohit-nayak-ps (approved):
lgtm

PR #10309, opened 2022-05-15
schemadiff: RangeRotationDistinctStatements partitioning hint

shlomi-noach:
## Description

Follow up to https://github.com/vitessio/vitess/pull/10234. We change behavior as follows:

- The flag `RangeRotationStatements` is renamed to `RangeRotationDistinctStatements` to better describe what it does (as of this PR)
- When two tables are diffed, and partition rotation hint is `RangeRotationDistinctStatements`, and there are indeed partition rotation changes, then the result is a sequence of diffs (one diff with linked subsequent diffs). If the table definition it self was also diffed (e.g. a modified column), that is allowed, and that will always be the primary diff, to which the partition diffs are appended. If the table is otherwise unmodified, then the primary diff is the first in partition changes.
- As reminder, when `RangeRotationStrategy` is `RangeRotationFullSpec`, the above is irrelevant, and the diff results with a single `ALTER TABLE`, that may include a full partitioning spec.
- 

## Related Issue(s)

- https://github.com/vitessio/vitess/pull/10234
- #10203 

## Checklist


dbussink (approved):
I like how allowing this functionality is mostly removing code :smile:.

PR #10311, opened 2022-05-15
Remove internal savepoint post query execution

harshit-gangal:
## Description

This PR removed the internal savepoint created during the execution of a DML when the query execution is complete.

When a query is executed and it is detected that it will go to multiple shard then vitess created an internal savepoint for that transaction.
Once the query execution is complete that savepoint has no purpose.

Other things about keeping that savepoint around is that the next query if it goes to different shard then the old internal savepoint will also be sent along and created which has no use for the current query.

## Checklist

### DIFF
--- a/go/vt/vtgate/vtgate_test.go
+++ b/go/vt/vtgate/vtgate_test.go
@@ -519,7 +519,7 @@ var shardedVSchema = `
 		}
 	},
 	"tables": {
-		"user_extra": {
+		"sp_tbl": {
 			"column_vindexes": [
 				{
 					"column": "user_id",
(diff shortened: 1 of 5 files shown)


go/vt/vtgate/vtgate_test.go
```diff
@@ -509,3 +509,117 @@ func TestErrorIssuesRollback(t *testing.T) {
...
+		}
+	}
+}
+`
+
+func TestMultiInternalSavepointVtGate(t *testing.T) {
+	s := createSandbox(KsTestSharded)
+	s.ShardSpec = "-40-80-"
+	s.VSchema = shardedVSchema
+	srvSchema := getSandboxSrvVSchema()
```
frouioui, line 538: I think we need to change this so we can use `shardedVSchema` instead of the default sandbox's vschema.

harshit-gangal: The vschema always needs to be assigned (otherwise it will remain empty), there is a default shardSpec that needs to be overridden.

frouioui (approved):

PR #10315, opened 2022-05-16
Online DDL: fast (and correct) RANGE PARTITION operations, 1st iteration

shlomi-noach:
## Description

This PR introduces support for fast `ADD PARTITION` and `DROP PARTITION` in `RANGE` partitioned tables. At this time these fast migrations are not revertible, but a future iteration will make them revertible.

### Why, what's the problem with existing implementation?

For both speed and correctness.

- Speed: adding a new range partition is a simple operation, which does not require a full blown online schema change that copies the entire table over. The MySQL implementation for adding a partition is creating an empty table.
- Correctness: in a range partitioned table, dropping the oldest partition _should_ also delete all data in that partition. However, the way all online-schema-change tools are built, including vitess's `vitess` migrations, handles this incorrectly. When you `DROP PARTITION` in an online schema change tool, the tool creates a shadow table with that partition removed. It then copies over all table rows from the original table. What used to be the 2nd partition (now the first after dropping the oldest partition) turns out to be a good host for all the old partition's rows. The end result is that no rows are deleted in the process. Which not what the user wanted.

### How

Confronted with an `ALTER TABLE` statement, and before sending it over to the appropriate handler (`vitess`, `gh-ost`, `pt-osc`), the executor checks if it can execute the migration using a "special plan". By parsing and analyzing the exact semantics of the query, `vitess` is able to know a statement is a `ADD|DROP` of a `RANGE PARTITION`; and if that is the case, it executes the query directly on the backend MySQL server, rather then invoke an OSC.

### Limitations

- At this time, these migrations are not revertible. There is a path and plan to make them revertible in a next iteration.
- In anticipation of making these operations revertible, a `ALTER TABLE ... DROP PARTITION` may only drop one partition at a time (the MySQL syntax allows dropping multiple partitions at once).
- Right now and while this is a new and **in progress** behavior, it is protected by a new and undocumented ~~`--fast-over-revertible`~~ `--fast-range-rotation` flag in `ddl_strategy`. This flag will be useful to additional migration types, but we expect it to eventually be irrelevant to `ADD|DROP PARTITION` queries, as we do intend to make them revertible.

There are no release notes in the meantime as I want to see how next iterations make progress. No documentation for this new behavior yet until we consider it stable.

## Related Issue(s)

- https://github.com/vitessio/vitess/issues/10317
- #6926 

## Checklist


go/vt/vttablet/onlineddl/analysis.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+	createTable, ok := stmt.(*sqlparser.CreateTable)
+	if !ok {
+		return nil, vterrors.Errorf(vtrpcpb.Code_INTERNAL, "expected CREATE TABLE. Got %v", sqlparser.CanonicalString(stmt))
+	}
+	return createTable, nil
+}
+
+// analyzeDropFirstOrLastRangePartition sees if the online DDL drops the first or last partition
+//  in a range partitioned table
```
dbussink, line 86: Isn't this only checking if it's the first partition? 

And isn't adding checking if it's adding the last? So is this comment here correct?

go/vt/vttablet/onlineddl/analysis.go
```diff
@@ -0,0 +1,209 @@
...
+		return nil
+	}
+	if len(spec.Definitions) != 1 {
+		return nil
+	}
+	partitionDefinition := spec.Definitions[0]
+	partitionName := partitionDefinition.Name.String()
+	// OK then!
+
+	// Now, is this query adding a parititon in a RANGE partitioned table?
```
dbussink, line 162: `parititon` -> `partition`

go/vt/vttablet/onlineddl/analysis.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+	op := NewSpecialAlterOperation(addRangePartitionSpecialOperation, alterTable, createTable)
+	op.SetDetail("partition_name", partitionName)
+	op.SetDetail("partition_definition", sqlparser.CanonicalString(partitionDefinition))
+	return op
+}
+
+// analyzeSpecialAlterScenarios checks if the given ALTER onlineDDL, and for the current state of affected table,
+// can be executed in a special way
+func (e *Executor) analyzeSpecialAlterPlan(ctx context.Context, onlineDDL *schema.OnlineDDL) (*SpecialAlterPlan, error) {
```
dbussink, line 178: Comment is wrong for the function name here.

shlomi-noach: fixed

go/vt/vttablet/onlineddl/analysis.go
shlomi-noach: fixed

go/vt/vttablet/onlineddl/analysis.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+	op := NewSpecialAlterOperation(dropRangePartitionSpecialOperation, alterTable, createTable)
+	op.SetDetail("partition_name", partitionName)
+	op.SetDetail("partition_definition", sqlparser.CanonicalString(partitionDefinition))
+	op.SetDetail("next_partition_name", nextPartitionName)
+	return op, nil
+}
+
+// analyzeDropFirstOrLastRangePartition sees if the online DDL drops the first or last partition
+//  in a range partitioned table
```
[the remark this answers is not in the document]
shlomi-noach: fixed

go/vt/vttablet/onlineddl/analysis.go
shlomi-noach: fixed. The code was refactored into a more general form, and the original comment was left behind.

shlomi-noach:
Ugh edited online and that lost the signature. Gona force-push shortly.

shlomi-noach:
Related: https://github.com/vitessio/vitess/issues/10317

This PR implements the forward migration aspect.

dbussink:
> * Right now and while this is a new and **in progress** behavior, it is protected by a new and undocumented `--fast-over-revertible` flag in `ddl_strategy`. This flag will be useful to additional migration types, but we expect it to eventually be irrelevant to `ADD|DROP PARTITION` queries, as we do intend to make them revertible.

This was a thing I had most questions over reading the code, but I think this makes sense. My concern was mostly on whether it's needed to control revertible or not strategies at a finer level or not, say, wanting to use it for partitions but not for an `INSTANT` `ADD COLUMN` style DDL operation.

But if this flag becomes eventually irrelevant for partitions, I think most of that concern disappears. That's also because partitions seem to be quite a different beast in general but with a separate strategy for dealing with them as discussed in #10317 I think we're ok.

So tl;dr, I think that if we would keep partitions as is without #10317, we would want to have a separate flag to control partitions vs. other fast over revertible cases, but given #10317 I don't think we need that (so one flag is better then I think).

shlomi-noach:
> if we would keep partitions as is without https://github.com/vitessio/vitess/issues/10317, we would want to have a separate flag to control partitions vs. other fast over revertible cases,

My thought as well, these probably deserve different flags. Let's see how we make progress.

shlomi-noach:
Out of caution, added a new `ddl_strategy` flag, `--fast-range-rotation`, which is now what the new fast `ADD/DROP PARTITION` functionality requires.

`--fast-over-revertible` remains for now, unused. I see a future for `--fast-over-revertible` for other types of migrations.

Both flags remain undocumented.

shlomi-noach:
Ping for review from the Vitess team

deepthi (approved):
LGTM

PR #10318, opened 2022-05-16
Update the ALTER TABLE formatting

dbussink:
This updates the `ALTER TABLE` formatting to more close match what `CREATE TABLE` looks like as well.

In `CREATE TABLE` we have a newline indented with tab for each column, this change makes `ALTER TABLE` match that by having a newline for each separate change within an `ALTER TABLE`.

This makes these queries more readable for a human operator and avoids super long SQL statements.

## Related Issue(s)

#10203 since this helps improve schemadiff canonical output.

## Checklist


shlomi-noach (approved):
Small change, good luck with the gazillion tests...

dbussink:
Not going to further pursue this at this time.

Issue #8732, opened 2021-08-30
[vtadmin] Delete shard


doeg:
# vtctld2 UI

Found at http://localhost:15000/app/shard?keyspace=commerce&shard=0 > menu > Delete Shard

<img width="1792" alt="Screen Shot 2021-08-30 at 11 52 26 AM" src="https://user-images.githubusercontent.com/855595/131367810-6f25a9e7-3c1a-4c30-9d9f-53b102d60dc2.png">
<img width="1792" alt="Screen Shot 2021-08-30 at 11 52 29 AM" src="https://user-images.githubusercontent.com/855595/131367820-256a4caa-4a93-4b3f-b8fa-1da6fec5d65a.png">
<img width="1792" alt="Screen Shot 2021-08-30 at 11 52 31 AM" src="https://user-images.githubusercontent.com/855595/131367829-84e98ac5-9939-445a-acde-06b92f90869b.png">
<img width="1787" alt="Screen Shot 2021-08-30 at 11 53 22 AM" src="https://user-images.githubusercontent.com/855595/131367834-eb157cab-bb1e-4e8d-b5ff-a155525e5047.png">

PR #10319, opened 2022-05-16
[vtadmin] Further generalize DangerAction component

doeg:
## Description

The same DangerAction we know and love, but (further) generalized for use in contexts other than the Tablet page. This is another prefactor for deleting a shard (#8732), which is up next. 

delete-tablet

(The tablet in the above gif doesn't disappear right away; this is expected + noted in the snackbar notification.)

Obligatory Top Gun reference:

image

## Related Issue(s)

- Prefactor for #8732 
- Follow-up to #10293 

## Checklist

## Deployment Notes

N/A


notfelineit (approved):
🔝 🔫 🎉

PR #10321, opened 2022-05-17
Add support for additional encoding configurations

dbussink:
This adds support for a number of already deprecated things in the parser, but that people are using these days.

Without full parser support for these, it can be used to bypass checks that block things like foreign key constraints at the vtgate level.

## Related Issue(s)

Found during work for #10203

## Checklist


go/vt/sqlparser/ast.go
```diff
@@ -1746,12 +1746,17 @@ type ColumnType struct {
...
 
 	// Enum values
 	EnumValues []string
 }
 
+type ColumnCharset struct {
+	Name   string
+	Binary bool
+}
+
```
dbussink, line 1759: This type was added because in the type definition it's possible to add the binary marker for a character set, so we need to track this. We can't at the point of where we parse things backfill this into the existing collation. Firstly because we don't have access to that node, but more importantly because it would generate syntax that is invalid.

So we must track this as a separate property.

See also https://dev.mysql.com/doc/refman/8.0/en/cast-functions.html and specifically:

```
A COLLATE clause is not permitted within a [CONVERT()](https://dev.mysql.com/doc/refman/8.0/en/cast-functions.html#function_convert) or [CAST()](https://dev.mysql.com/doc/refman/8.0/en/cast-functions.html#function_cast) call, but you can apply it to the function result. For example, these are legal:
```

But syntax like the following is valid:

```
SELECT CONVERT('test', CHAR CHARACTER SET utf8mb4 binary)
```

Which will end up using the `utf8mb4_bin` collation (see also the evalengine changes in this PR).

go/vt/sqlparser/sql.y
```diff
@@ -1931,6 +1934,12 @@ char_type:
   {
     $$ = ColumnType{Type: string($1), Length: $2, Charset: $3}
   }
+| CHAR length_opt BYTE
+  {
+    // CHAR BYTE is an alias for binary. See also:
+    // https://dev.mysql.com/doc/refman/8.0/en/string-type-syntax.html
```
dbussink, line 1940: This is a fun one, this is actually an alias for `binary`. See also the linked page, under:

```
The [CHAR BYTE](https://dev.mysql.com/doc/refman/8.0/en/binary-varbinary.html) data type is an alias for the [BINARY](https://dev.mysql.com/doc/refman/8.0/en/binary-varbinary.html) data type. This is a compatibility feature.
```

go/vt/sqlparser/sql.y
```diff
@@ -2097,29 +2106,53 @@ zero_fill_opt:
...
-    $$ = encodeSQLString($2)
+    $$ = ColumnCharset{Name: encodeSQLString($2), Binary: $3}
   }
 | charset_or_character_set BINARY
   {
-    $$ = string($2)
+    $$ = ColumnCharset{Name: string($2)}
   }
-| ASCII
+| ASCII binary_opt
```
dbussink, line 2123: The ASCII and UNICODE shorthands accept optionally a `binary` added as well.

go/vt/sqlparser/sql.y
```diff
@@ -2097,29 +2106,53 @@ zero_fill_opt:
...
+    $$ = ColumnCharset{Name: "latin1", Binary: $2}
   }
-| UNICODE
+| UNICODE binary_opt
   {
     // UNICODE: Shorthand for CHARACTER SET ucs2.
-    $$ = "ucs2"
+    $$ = ColumnCharset{Name: "ucs2", Binary: $2}
+  }
+| BINARY
```
dbussink, line 2133: `binary` on its own is also allowed, which means the default character set but with the binary collation.

go/vt/sqlparser/sql.y
```diff
@@ -2097,29 +2106,53 @@ zero_fill_opt:
...
+  {
+    // BINARY: Shorthand for default CHARACTER SET but with binary collation
+    $$ = ColumnCharset{Name: "", Binary: true}
+  }
+| BINARY ASCII
```
dbussink, line 2138: Both orderings of `ascii binary` and `binary ascii` are accepted and mean the same thing.

go/vt/sqlparser/testdata/select_cases.txt
```diff
@@ -12913,8 +12913,8 @@ END
 INPUT
 select collation(cast('a' as char(2))), collation(cast('a' as char(2) binary));
 END
-ERROR
-syntax error at position 77 near 'binary'
+OUTPUT
+select collation(convert('a', char(2))), collation(convert('a', char(2) binary)) from dual
```
dbussink, line 12917: Yay, we now can handle these cases that were before marked as a syntax error :smile:. They also work on MySQL itself.

go/vt/vtgate/evalengine/format.go
```diff
@@ -189,7 +189,6 @@ func (b *BitwiseNotExpr) format(buf *formatter, depth int) {
 func (c *ConvertExpr) format(buf *formatter, depth int) {
 	buf.WriteString("CONVERT(")
 	c.Inner.format(buf, depth)
-	buf.WriteString(", ")
```
dbussink, line 192: This was causing a duplicate `, ` in the converted output. I'm not sure how this is even working today, but this was generated invalid SQL since the switch below which adds the type also prefixes it with `, `. 

There didn't seem to be any test coverage for this, so I added tests also to verify the newly added binary behavior.

go/vt/vtgate/evalengine/translate.go
```diff
@@ -402,20 +402,32 @@ func translateUnaryExpr(unary *sqlparser.UnaryExpr, lookup TranslationLookup) (E
 	}
 }
 
-func translateConvertCharset(charset string, lookup TranslationLookup) (collations.ID, error) {
+func binaryCollationForCollation(collation collations.ID) collations.ID {
```
dbussink, line 405: Dunno if there's a more efficient way to do this, or if we should care to be more efficient here for this fairly edge case behavior?

go/vt/vtgate/evalengine/translate.go
```diff
@@ -481,7 +493,7 @@ func translateConvertUsingExpr(expr *sqlparser.ConvertUsingExpr, lookup Translat
 		return nil, err
 	}
 
-	using.Collation, err = translateConvertCharset(expr.Type, lookup)
+	using.Collation, err = translateConvertCharset(expr.Type, false, lookup)
```
dbussink, line 496: No `binary` or collation here can be specified, so we can always pass in `false` correctly here.

go/vt/vtgate/evalengine/translate_test.go
```diff
@@ -90,6 +90,8 @@ func TestTranslateSimplification(t *testing.T) {
 		{"weight_string('foobar' as char(12))", ok(`WEIGHT_STRING(VARCHAR("foobar") AS CHAR(12))`), ok(`VARBINARY("\x00F\x00O\x00O\x00B\x00A\x00R\x00 \x00 \x00 \x00 \x00 \x00 ")`)},
 		{"case when 1 = 1 then 2 else 3 end", ok("CASE WHEN INT64(1) = INT64(1) THEN INT64(2) ELSE INT64(3)"), ok("INT64(2)")},
 		{"case when null then 2 when 12 = 4 then 'ohnoes' else 42 end", ok(`CASE WHEN NULL THEN INT64(2) WHEN INT64(12) = INT64(4) THEN VARCHAR("ohnoes") ELSE INT64(42)`), ok(`VARCHAR("42")`)},
+		{"convert('a', char(2) character set utf8mb4)", ok(`CONVERT(VARCHAR("a"), CHAR(2) CHARACTER SET utf8mb4_0900_ai_ci)`), ok(`VARCHAR("a")`)},
+		{"convert('a', char(2) character set latin1 binary)", ok(`CONVERT(VARCHAR("a"), CHAR(2) CHARACTER SET latin1_bin)`), ok(`VARCHAR("a")`)},
```
dbussink, line 94: These now properly work in the evalengine also after fixing that stray `, `.

go/vt/sqlparser/ast.go
frouioui: nit: let's merge this block with the block above

frouioui: > This type was added because in the type definition it's possible to add the binary marker for a character set, so we need to track this. We can't at the point of where we parse things backfill this into the existing collation. Firstly because we don't have access to that node, but more importantly because it would generate syntax that is invalid.

Makes sense, thank you!

go/vt/vtgate/evalengine/translate.go
```diff
@@ -402,20 +402,32 @@ func translateUnaryExpr(unary *sqlparser.UnaryExpr, lookup TranslationLookup) (E
 	}
 }
 
-func translateConvertCharset(charset string, lookup TranslationLookup) (collations.ID, error) {
+func binaryCollationForCollation(collation collations.ID) collations.ID {
+	binary := collations.Local().LookupByID(collation)
+	return collations.Local().BinaryCollationForCharset(binary.Charset().Name()).ID()
```
frouioui, line 407: I feel like we may want to ensure that:
```go
collations.Local().BinaryCollationForCharset(binary.Charset().Name())
```
is not returning `nil` before calling `ID()`.

frouioui: if the given collation ID does not have a binary equivalent it would return `nil`

frouioui (approved):
I left a few nits/comments, otherwise everything looks good to me.

go/vt/sqlparser/ast.go
frouioui: I think we should write a code comment here that contains what you wrote in your GitHub comment:
> This type was added because in the type definition it's possible to add the binary marker for a character set, so we need to track this. We can't at the point of where we parse things backfill this into the existing collation. Firstly because we don't have access to that node, but more importantly because it would generate syntax that is invalid.

go/vt/vtgate/evalengine/translate.go
dbussink: @frouioui Yeah, can add some safety checks. I think in practice though, every charset and thus also every charset through a collation has a binary collation available. It's always `<charset>_bin`. But a check won't hurt here and good to have just in case.

go/vt/sqlparser/ast.go
dbussink: > nit: let's merge this block with the block above

Not entirely sure what you mean by this?

dbussink: Rest is addressed!

go/vt/vtgate/evalengine/translate.go
dbussink: Done!

go/vt/sqlparser/sql.y
shlomi-noach: gah

shlomi-noach (approved):

PR #10322, opened 2022-05-17
Add parsing support for Regular Expressions

K-Kumar-01:
## Description

Add parsing support for regular expressions. More information [here](https://dev.mysql.com/doc/refman/8.0/en/regexp.html)

## Related Issue(s)

#8604 
#4099

## Checklist


go/vt/sqlparser/parse_test.go
```diff
@@ -2943,6 +2943,57 @@ var (
...
+		output: "select regexp_substr('abc def ghi', '[a-z]+', 1, 3) from dual",
+	}, {
+		input:  "SELECT REGEXP_SUBSTR('abc def ghi', '[a-z]+', 1, 3, TRIM(' n '));",
+		output: "select regexp_substr('abc def ghi', '[a-z]+', 1, 3, trim(' n ')) from dual",
+	}, {
+		input:  "SELECT 'Michael!' RLIKE '.*';",
+		output: "select 'Michael!' rlike '.*' from dual",
+	}, {
+		input:  "SELECT TRIM('Michael!') RLIKE @j",
+		output: "select trim('Michael!') rlike @j from dual",
```
dbussink, line 2996: Can you also add some tests to `go/vt/sqlparser/tracked_buffer_test.go` to verify what the canonical output format of regular expressions looks like?

go/vt/sqlparser/ast.go
```diff
@@ -2541,6 +2541,53 @@ type (
...
+		Expr       Expr
+		Pattern    Expr
+		Occurrence Expr
+		Position   Expr
+		MatchType  Expr
+	}
+
+	// RLikeExpr represents RLIKE
+	// For more information, visit https://dev.mysql.com/doc/refman/8.0/en/regexp.html#operator_regexp
+	RLikeExpr struct {
```
dbussink, line 2587: Do we need / want separate structures for this or do we want to put everything into `REGEXP_LIKE` and always output that format when serializing again?

We also always change `CAST` into `CONVERT` as well for example.

go/vt/sqlparser/parse_test.go
K-Kumar-01: Done @dbussink

go/vt/sqlparser/ast.go
dbussink: > Also, I don't quite understand regarding the CAST into CONVERT thing.

That's an example of where slightly different syntax is also stored in the AST in the same structure.

go/vt/sqlparser/parse_test.go
```diff
@@ -563,11 +563,11 @@ var (
 	}, {
 		input: "select /* not regexp */ 1 from t where a not regexp b",
 	}, {
-		input:  "select /* rlike */ 1 from t where a rlike b",
```
systay, line 566: I think we should keep the input as `rlike`

go/vt/sqlparser/ast.go
GuptaManan100: I agree with @dbussink, we should normalise into the same AST struct, which I see we have already done now

GuptaManan100: @K-Kumar-01 Oh no! This won't work. I see in the docs https://dev.mysql.com/doc/refman/5.7/en/regexp.html, that mysql 5.7 didn't have the `regexp_like` function at all. So we can't normalise all the inputs to that function call blindly, otherwise things that used to work with mysql 5.7 will start failing 😢

GuptaManan100: There are 2 options here - 
1. Go back to having two different structs, one for `regexp_like` and one for `rlike` and `regexp`
2. Have one struct, but check whether Match_type is empty in `regexp_like` and if it is, print it out as `regexp` and not a function call.

I think both solutions have there own issues, the first one will increase the number of structs we have and maybe a little harder to use later on. The second solution will mean that we print out `regexp` whenever we can, and use `regexp_like` only when we need to. I don't know if this conversion between the two will always be safe or not, in nested function calls, etc. We tried to do something similar with printing of brackets in default clause, and it didn't work out all that well with us finding multiple bugs in it later on.

WDYT @vitessio/query-serving?

go/vt/sqlparser/sql.y
```diff
@@ -4988,19 +4988,27 @@ bit_expr IN col_tuple
   {
 	$$ = &ComparisonExpr{Left: $1, Operator: NotLikeOp, Right: $4, Escape: $6}
   }
-| bit_expr REGEXP bit_expr
+| bit_expr regexp_symbol bit_expr
   {
-	$$ = &ComparisonExpr{Left: $1, Operator: RegexpOp, Right: $3}
+	$$ = &RegexpLikeExpr{Expr: $1, Pattern: $3}
```
GuptaManan100, line 4993: This conversion might not be safe 😞

GuptaManan100 (changes requested):

go/vt/sqlparser/sql.y
GuptaManan100: More discussion in the thread above

go/vt/sqlparser/ast.go
dbussink: > Go back to having two different structs, one for regexp_like and one for rlike and regexp

Seems like this is more robust and when we get to the future point of deprecating 5.7 support, we can clean it up then if we want to / if it's a significant burden?

go/vt/sqlparser/sql.y
K-Kumar-01: I will update accordingly.

go/vt/sqlparser/ast.go
K-Kumar-01: @dbussink @GuptaManan100 
The earlier implementation for `regex` didn't have any struct. It was implemented as:
```
| bit_expr REGEXP bit_expr
  {
	$$ = &ComparisonExpr{Left: $1, Operator: RegexpOp, Right: $3}
  }
| bit_expr NOT REGEXP bit_expr
  {
	 $$ = &ComparisonExpr{Left: $1, Operator: NotRegexpOp, Right: $4}
  }
```
So, I was thinking of not creating the struct for Rlike. Instead extend this parsing to support Rlike expressions.
Should I go ahead with this implementation?

go/vt/sqlparser/keywords.go
```diff
@@ -482,6 +482,11 @@ var keywords = []keyword{
 	{"redundant", REDUNDANT},
 	{"references", REFERENCES},
 	{"regexp", REGEXP},
+	{"regexp_instr", REGEXP_INSTR},
+	{"regexp_like", REGEXP_LIKE},
+	{"regexp_replace", REGEXP_REPLACE},
+	{"regexp_substr", REGEXP_SUBSTR},
+	{"rlike", RLIKE},
```
GuptaManan100, line 489: RLIKE was in the correct alphabetic position before

go/vt/sqlparser/testdata/select_cases.txt
```diff
@@ -14609,7 +14609,7 @@ OUTPUT
 select std(e1 / e2) from bug22555
 END
 INPUT
-select 'вася' rlike '[[:<:]]вася[[:>:]]';
+select 'вася' regexp '[[:<:]]вася[[:>:]]';
```
GuptaManan100, line 14612: Why change the input lines?

go/vt/sqlparser/testdata/select_cases.txt
```diff
@@ -15983,7 +15983,7 @@ OUTPUT
 select Fld1, max(Fld2) as q from t1 group by Fld1 having q is not null
 END
 INPUT
-select 'zвасяz' rlike '[[:<:]]вася[[:>:]]';
+select 'zвасяz' regexp '[[:<:]]вася[[:>:]]';
```
GuptaManan100, line 15986: This is an input too

go/vt/sqlparser/testdata/select_cases.txt
```diff
@@ -17465,7 +17465,7 @@ OUTPUT
 select table_type from information_schema.`tables` where table_name = 'v1'
 END
 INPUT
-select _latin1'a' regexp _latin1'A' collate latin1_bin;
+select _latin1'a' rlike _latin1'A' collate latin1_bin;
```
GuptaManan100, line 17468: So is this

go/vt/sqlparser/testdata/select_cases.txt
```diff
@@ -21413,7 +21413,7 @@ OUTPUT
 select t1.a, t1.b, t2.c from t1 left join t2 on t1.a = t2.a and t1.b = t2.b and t2.c = 3 where t1.a = 1 and t2.c is null
 END
 INPUT
-select 'вася ' rlike '[[:<:]]вася[[:>:]]';
+select 'вася ' regexp '[[:<:]]вася[[:>:]]';
```
GuptaManan100, line 21416: and this

GuptaManan100 (changes requested):

go/vt/sqlparser/keywords.go
K-Kumar-01: Updated.

go/vt/sqlparser/ast_format.go
```diff
@@ -1353,6 +1353,66 @@ func (node *ExtractFuncExpr) Format(buf *TrackedBuffer) {
...
+	if node.MatchType != nil {
+		buf.astPrintf(node, ", %v", node.MatchType)
+	}
+	buf.WriteByte(')')
+}
+
+// Format formats the node
+func (node *RegexpLikeExpr) Format(buf *TrackedBuffer) {
+	if node.IsNot {
+		buf.WriteString("not ")
```
dbussink, line 1377: I think we also want `buf.astPrintf` here to have consistent casing of `NOT` also for the canonical format.

go/vt/sqlparser/parse_test.go
```diff
@@ -559,9 +559,11 @@ var (
 	}, {
 		input: "select /* not like escape */ 1 from t where a not like b escape '$'",
 	}, {
-		input: "select /* regexp */ 1 from t where a regexp b",
+		input:  "select /* regexp */ 1 from t where a regexp b",
+		output: "select /* regexp */ 1 from t where a regexp b",
```
dbussink, line 563: Input and output are the same here, so I think we can drop the output here to reduce the diff?

go/vt/sqlparser/parse_test.go
```diff
@@ -559,9 +559,11 @@ var (
...
 	}, {
-		input: "select /* not regexp */ 1 from t where a not regexp b",
+		input:  "select /* not regexp */ 1 from t where a not regexp b",
+		output: "select /* not regexp */ 1 from t where a not regexp b",
```
dbussink, line 566: Same here, input and output are the same so we can drop the output and reduce the diff.

go/vt/sqlparser/tracked_buffer_test.go
```diff
@@ -188,6 +188,10 @@ func TestCanonicalOutput(t *testing.T) {
 			"create table t (p point srid 0, g geometry not null srid 4326)",
 			"CREATE TABLE `t` (\n\t`p` point SRID 0,\n\t`g` geometry NOT NULL SRID 4326\n)",
 		},
+		{
+			input:     "select regexp_replace('abc def ghi', '[a-z]+', 'X', 1, 3, 'c'), REGEXP_LIKE('dog cat dog', 'dog'), REGEXP_SUBSTR('abc def ghi', '[a-z]+', 1), REGEXP_INSTR('aa aaa aaaa aaaa aaaa aaaa', 'a{4}',1), 'Michael!' RLIKE '.*' from dual",
+			canonical: "SELECT REGEXP_REPLACE('abc def ghi', '[a-z]+', 'X', 1, 3, 'c'), REGEXP_LIKE('dog cat dog', 'dog'), REGEXP_SUBSTR('abc def ghi', '[a-z]+', 1), REGEXP_INSTR('aa aaa aaaa aaaa aaaa aaaa', 'a{4}', 1), 'Michael!' REGEXP '.*' FROM `dual`",
```
dbussink, line 193: All other cases here don't use the prefix but depend on ordering. I'm more of a fan of including the names, but even more a fan of staying consistent with the rest :smile:. So I think we should stay consistent here.

dbussink: Can you also add a case here for the `NOT` (also in reference to my earlier comment of using `astPrintf` for that.

go/vt/sqlparser/ast_format.go
K-Kumar-01: The field `IsNot` was added to represent `Regexp` and `Rlike` in terms of `REGEXP_LIKE` when the syntax looked something similar to : `expr NOT REGEXP expr`. 
As we parse this differently now itself, this is redundant and won't be used anywhere.
I will remove this field.

go/vt/sqlparser/tracked_buffer_test.go
K-Kumar-01: > All other cases here don't use the prefix but depend on ordering. I'm more of a fan of including the names, but even more a fan of staying consistent with the rest smile. So I think we should stay consistent here.

I did not understand what needs to be done here. :sweat_smile:

dbussink: @K-Kumar-01 You can drop the `input:` and `canonical:` prefixes here, so it looks like entries above.

go/vt/sqlparser/ast_format.go
dbussink: @K-Kumar-01 Ah, if it's redundant, even better to remove it!

go/vt/sqlparser/tracked_buffer_test.go
```diff
@@ -188,6 +188,14 @@ func TestCanonicalOutput(t *testing.T) {
 			"create table t (p point srid 0, g geometry not null srid 4326)",
 			"CREATE TABLE `t` (\n\t`p` point SRID 0,\n\t`g` geometry NOT NULL SRID 4326\n)",
 		},
+		{
+			"select regexp_replace('abc def ghi', '[a-z]+', 'X', 1, 3, 'c'), REGEXP_LIKE('dog cat dog', 'dog'), REGEXP_SUBSTR('abc def ghi', '[a-z]+', 1), REGEXP_INSTR('aa aaa aaaa aaaa aaaa aaaa', 'a{4}',1), 'Michael!' RLIKE '.*' from dual",
+			"SELECT REGEXP_REPLACE('abc def ghi', '[a-z]+', 'X', 1, 3, 'c'), REGEXP_LIKE('dog cat dog', 'dog'), REGEXP_SUBSTR('abc def ghi', '[a-z]+', 1), REGEXP_INSTR('aa aaa aaaa aaaa aaaa aaaa', 'a{4}', 1), 'Michael!' REGEXP '.*' FROM `dual`",
+		},
+		{
+			"select regexp_replace('abc def ghi', '[a-z]+', 'X', 1, 3, 'c'), REGEXP_LIKE('dog cat dog', 'dog'), REGEXP_SUBSTR('abc def ghi', '[a-z]+', 1), REGEXP_INSTR('aa aaa aaaa aaaa aaaa aaaa', 'a{4}',1), 'Michael!' RLIKE '.*' from dual",
+			"SELECT REGEXP_REPLACE('abc def ghi', '[a-z]+', 'X', 1, 3, 'c'), REGEXP_LIKE('dog cat dog', 'dog'), REGEXP_SUBSTR('abc def ghi', '[a-z]+', 1), REGEXP_INSTR('aa aaa aaaa aaaa aaaa aaaa', 'a{4}', 1), 'Michael!' REGEXP '.*' FROM `dual`",
```
dbussink, line 197: @K-Kumar-01 I think you wanted to add a case for `NOT` here, but it looks like there's now two equivalent cases here? Maybe a copy paste and then forgetting to update the second case?

K-Kumar-01: @dbussink Yeah, my bad. I updated the test case in parse_test.go but forgot to update the one here. Update the code.

dbussink: @K-Kumar-01 Thnx!

GuptaManan100 (approved):

dbussink (approved):

PR #10330, opened 2022-05-18
schemadiff: case insensitive col/index/partition name comparison

shlomi-noach:
## Description

Column, key, partition etc. names are case insensitive. This PR iterates towards making the following case insensitive aware:

- Diff()
- Apply()
- normalize()
- postApplyNormalize()
- validate()

## Related Issue(s)

#10203 

## Checklist


shlomi-noach:
This is now ready to review. What's done and what's not done in this PR:

- When a column's case is modified, `schemadiff` understands this is the same column, and there is still a `ALTER TABLE... MODIFY COLUMN` statement to rename it with the new case
- All validation (key covers existing columns, functional index references column, partition referencing columns) are case-insensitive. i.e. validation is satisfied if there's a column called `id` and an index `KEY id_idx(ID)`
- if column name case changes inside an index definition or partition definition, that's an `ALTER TABLE` to rebuild key/partition.
-

dbussink (approved):

PR #10335, opened 2022-05-18
Fix schema tracking issue when `PRIMARY` tablet changes

frouioui:
## Description

Hola, hola! 👋 This Pull Request fixes a nasty bug in our schema tracking feature (🙃). What's the bug? When creating a Vitess cluster and enabling schema tracking, the new VTTablets are added to the Topology. The VTGate then picks up the new changes from the Topology and sees that a bunch of new VTTablets was added. It saves the state of those VTTablets and starts reading the stream of healthcheck from them. Any updates to those VTTablets are sent to VTGate through that stream using a `StreamHealthResponse` structure. In VTGate, the data structure that keeps track of a VTTablet's health is `tabletHealthCheck`, that structure keeps track of both the `current` and `target` tablet type (`primary`, `replica`, ...). The `target` tablet type is updated every time we receive a healthcheck from the VTTablet, however, the `current` is **never** updated (😱) ... So what's the link with schema tracking? In the schema tracking code, whenever we receive an update, the first thing we check is _"is the healthcheck coming from a `primary` VTTablet?"_, if the answer to that question is _"no"_ then we dismiss the update and ignore it. And... how do we check the tablet type? By looking at the `current` tablet type, which is never updated. Meaning that, in a scenario where we have 2 VTTablets (`A` and `B`) in each shard, where `A` is the `primary` by default, if at one point we do a PRS ([Planned Reparent Shard](https://vitess.io/docs/user-guides/configuration-advanced/reparenting/)), the `primary` VTTablet will change to be `B`, however, the VTGate will not update the current tablet type of both VTTablets, thus, it will keep thinking that the primary VTTablet is `A` even though it is `B`. Meaning that when `B` sends a valid schema tracking update, VTGate ignores it.

This Pull Request proposes an end-to-end test that reproduces the issue, and two fixes:
- The first one (a6b5ca28a0d745ec0b59d1bca21f4583dcca5955) keeps track of the `current` tablet type in VTGate.
- The second one (3dd05e0b70d6cccb98390497159162ba2e28f293) uses the `target` tablet type in the schema tracker. In VTTablet, we also use the `target` tablet type to check whether the VTTablet has to send the schema tracking update or not.

## Related Issue(s)

- Couldn't find any, don't hesitate to link them though

## Checklist

### DIFF
--- a/go/test/endtoend/vtgate/schematracker/sharded_prs/st_sharded_test.go
+++ b/go/test/endtoend/vtgate/schematracker/sharded_prs/st_sharded_test.go
@@ -1,0 +1,217 @@
+/*
+Copyright 2021 The Vitess Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package shardedprs
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"os"
+	"testing"
+	"time"
+
+	vtrpcpb "vitess.io/vitess/go/vt/proto/vtrpc"
+	"vitess.io/vitess/go/vt/vterrors"
+
+	"vitess.io/vitess/go/test/endtoend/utils"
+
+	"github.com/stretchr/testify/require"
+
+	"vitess.io/vitess/go/mysql"
+	"vitess.io/vitess/go/test/endtoend/cluster"
+)
+
+var (
+	clusterInstance *cluster.LocalProcessCluster
+	vtParams        mysql.ConnParams
+	KeyspaceName    = "ks"
+	Cell            = "test"
+	SchemaSQL       = `
+create table t2(
+	id3 bigint,
+	id4 bigint,
+	primary key(id3)
+) Engine=InnoDB;
+
+create table t2_id4_idx(
+	id bigint not null auto_increment,
+	id4 bigint,
+	id3 bigint,
+	primary key(id),
+	key idx_id4(id4)
+) Engine=InnoDB;
+
+create table t8(
+	id8 bigint,
+	testId bigint,
+	primary key(id8)
+) Engine=InnoDB;
+`
+
+	VSchema = `
+{
+  "sharded": true,
+  "vindexes": {
+    "unicode_loose_xxhash" : {
+	  "type": "unicode_loose_xxhash"
+    },
+    "unicode_loose_md5" : {
+	  "type": "unicode_loose_md5"
+    },
+    "hash": {
+      "type": "hash"
+    },
+    "xxhash": {
+      "type": "xxhash"
+    },
+    "t2_id4_idx": {
+      "type": "lookup_hash",
+      "params": {
+        "table": "t2_id4_idx",
+        "from": "id4",
+        "to": "id3",
+        "autocommit": "true"
+      },
+      "owner": "t2"
+    }
+  },
+  "tables": {
+    "t2": {
+      "column_vindexes": [
+        {
+          "column": "id3",
+          "name": "hash"
+        },
+        {
+          "column": "id4",
+          "name": "t2_id4_idx"
+        }
+      ]
+    },
+    "t2_id4_idx": {
+      "column_vindexes": [
+        {
+          "column": "id4",
+          "name": "hash"
+        }
+      ]
+    },
+    "t8": {
+      "column_vindexes": [
+        {
+          "column": "id8",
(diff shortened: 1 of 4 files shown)


go/vt/vtgate/schema/update_controller.go
```diff
@@ -119,7 +119,7 @@ func (u *updateController) getItemFromQueueLocked() *discovery.TabletHealth {
 
 func (u *updateController) add(th *discovery.TabletHealth) {
 	// For non-primary tablet health, there is no schema tracking.
-	if th.Tablet.Type != topodatapb.TabletType_PRIMARY {
+	if th.Target.TabletType != topodatapb.TabletType_PRIMARY {
```
frouioui, line 122: This is the "second" fix where we look at the target instead of the current tablet type, the same approach is used in VTTablet

go/vt/discovery/tablet_health_check.go
```diff
@@ -177,6 +177,12 @@ func (thc *tabletHealthCheck) processResponse(hc *HealthCheckImpl, shr *query.St
 		serving = false
 	}
 
+	ti, err := hc.ts.GetTablet(thc.ctx, thc.Tablet.Alias)
+	if err != nil {
+		return fmt.Errorf("could not get tablet info from topo: %v", err)
+	}
+	thc.Tablet = ti.Tablet
```
frouioui, line 184: This might lead to a high impact on performance since we have to question the topo for every health update. However, we can remove the commit that introduces this, the bug is fixable with only one or two of the two fixes proposed by this PR.

go/test/endtoend/vtgate/schematracker/sharded_prs/st_sharded_test.go
```diff
@@ -0,0 +1,217 @@
...
+			return 1
+		}
+
+		// Start keyspace
+		keyspace := &cluster.Keyspace{
+			Name:      KeyspaceName,
+			SchemaSQL: SchemaSQL,
+			VSchema:   VSchema,
+		}
+		err = clusterInstance.StartKeyspace(*keyspace, []string{"-80", "80-"}, 3, true)
```
harshit-gangal, line 148: this menas 1+3 replica and 1 rdonly.
we can reduce the resource requirement by (2, false)

go/vt/discovery/tablet_health_check.go
harshit-gangal: We should not do topo calls here as you rightly said it will have a performance impact.
Also, it will increase load on topo server. 
If we plan to fix this, we should do it in a separate PR and also understand how to do it better.

harshit-gangal (commented):
this is an awesome found.
Just we can do the healthcheck fix separately.

harshit-gangal (approved):

go/vt/discovery/tablet_health_check.go
dbussink: @frouioui @harshit-gangal Wondering about this, if the specific state of a tablet here is not regularly updated, should that state be removed also from the object? 

This is more for a future more deeper change, but I wonder if it's better to not make the information available at all if it can't be relied on. That way it won't be used in other places and no new usage would be added that might be unknowingly broken? 

Not sure how hard this would be to change though but avoiding making an easy mistake might be worth it?

dbussink (approved):

go/vt/discovery/tablet_health_check.go
frouioui: I think you're right @dbussink. I feel like both variables serve two different purposes, it is worth evaluating whether this can be removed or not.

go/test/endtoend/vtgate/schematracker/sharded_prs/st_sharded_test.go
```diff
@@ -0,0 +1,218 @@
...
+	timeout := time.After(5 * time.Minute)
+	for {
+		select {
+		case <-timeout:
+			return vterrors.New(vtrpcpb.Code_INTERNAL, "timeout")
+		default:
+			err := clusterInstance.WaitForTabletsToHealthyInVtgate()
+			if err != nil {
+				return err
+			}
```
frouioui, line 201: Just realized that this is wrong. We should not `return nil`, instead we should retry

frouioui:
We should backport this fix to `v13`.

harshit-gangal:
> We should backport this fix to `v13`.

schema tracking is experimental we do not need to backport it.

dbussink:
> schema tracking is experimental we do not need to backport it.

Ah ok, even if it's fairly fundamentally broken as it is now? I get it for more edge cases, but dunno if there's other criteria for more broader bugs?

go/vt/discovery/tablet_health_check.go
deepthi: This is complex. @harshit-gangal and I looked at the code to see whether we can completely remove the reference to the Tablet struct. We are using other information from the Tablet struct in various places, so we can't do that easily.

dbussink: @deepthi Do you think it would work to create two separate types for this? One with only the static information that can still be depended on in places where we do today and where we don't need access to information that changes (like the tablet replica / primary state).

And then have a separate type for the information that does change and gets updated by health checks?

Issue #10317, opened 2022-05-16
Design: REVERT for fast ADD|DROP PARTITION for range partitioned tables


shlomi-noach:
This issue is to explain how I wish to implement fast revert for `ADD|DROP PARTITION` operations. Some initial context: https://github.com/vitessio/vitess/pull/10315

## The problem

Consider the following range partitioned table:

```sql
CREATE TABLE tp (
    id INT NOT NULL,
    ts TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
    primary key (id)
)
PARTITION BY RANGE (id) (
    PARTITION p1 VALUES LESS THAN (10),
    PARTITION p2 VALUES LESS THAN (20),
    PARTITION p3 VALUES LESS THAN (30),
    PARTITION p4 VALUES LESS THAN (40),
    PARTITION p5 VALUES LESS THAN (50),
    PARTITION p6 VALUES LESS THAN (60)
);
```

Assuming the table is very large, and assuming the user wants to routinely rotate it, we can expect periodic `ALTER TABLE ... DROP PARTITION` and `ALTER TABLE ... ADD PARTITION` DDLs, which the user expect to operate very quickly. Moreover, in the spirit of Online DDL, we want to be able to revert such statements.

Nothing is trivial here and the problems begin with almost the simplest query.

## We cannot use good-old online schema change to add/drop partitions because it's slow

The canonical way for OSC is to create a new empty shadow table, alter that table, then copy over the data from the original table. If the table is large, then the operation is slow. But range partitioning's greatest incentive is that we can operate on smaller tables (partitions), and where operations are proportional to the size of affected partitions, not of the entire table. In particular, `ALTER TABLE ... ADD PARTITION` is a near instant operation. But OSC can take hours to complete it.

## We cannot use good-old online schema change to drop partitions because it's incorrect

This is true for all OSC tools. Consider that we issue a `ALTER TABLE tp DROP PARTITION p1` on our sample table above. Alkso, consider the table is populated, and has an id value of `2`, which belongs in `p1`.

The user _assumes_ that dropping `p1` also drops the row where `id = 2`. But in an OSC we build a new table:

```sql
CREATE TABLE tp (
    id INT NOT NULL,
    ts TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
    primary key (id)
)
PARTITION BY RANGE (id) (
    PARTITION p2 VALUES LESS THAN (20),
[...]

PR #10337, opened 2022-05-19
ON HOLD: Revertible fast partition migrations

shlomi-noach:
## Description

Implementing the solution described in https://github.com/vitessio/vitess/issues/10317. This PR makes partition rotation migrations:

- [x] fast
- [x] revertible
- [ ] consistent (data not swapped out while table accessible to app)

The revert logic is already implemented and tested.

Remaining work is to make the operation atomic/safe to use, and consider the correct course of action in the following:

- Tablet crashed midway
- Migration fails or is cancelled midway

One approach is to make these migrations idempotent and so we can recover from both the above by reexecuting them. However, this is not consistent with standard migration flow. e.g. in a `add partition p20 values less than (20)`, and if we find `p20` to exist, how will we make the distinction between a first-time obviously-conflicting migration, and a retry of a halfway broken migration?

One answer is to journal the entire process; maybe that's the most correct solution, but I want to explore a bit and see if we can come up with a simple set of rules. If it turn to be too complex, then journal it is.

## Related Issue(s)

- Design: https://github.com/vitessio/vitess/issues/10317
- Previous iteration: https://github.com/vitessio/vitess/pull/10315

## Checklist


shlomi-noach:
Putting this PR on hold. While we have the logic for fast partition rotation reverts, I'm dissatisfied with the introduced complexity. I want to let this sit still; see if we come up with more ideas, or encounter new use cases that may suggest the best path forward, it at all. In the meantime, fast partition rotation migrations are non-revertible.

shlomi-noach:
I don't think we should auto close stale Draft PRs

PR #10341, opened 2022-05-19
[vtadmin] Refactor getTabletForAction

ajm188:
## Description

I extracted this out of another PR I'm working on because it's self-contained and useful! (the ShardResource thing is coming from that, by way of explanation). Anyway, onward!

<hr/>

This refactors `getTabletForAction` => `getTabletForResourceAndAction` to support more resource types (`rbac.ShardResource` being my immediate use case), as well as return the `cluster.Cluster` object the tablet was found in.

Returning the cluster directly minimizes the number of times we need
to hold the `clusterMap` mutex, since every request needs to

(1) get the tablet then
(2) get the cluster for that tablet

we already have the cluster, so we iterate twice and return it directly

## Checklist


notfelineit (approved):
ooh nice refactor

doeg (approved):
so nice!!!

PR #10354, opened 2022-05-22
[vtadmin] Exclude keyspaces from schema cache

ajm188:
## Description

- TODO

## Checklist

### DIFF
--- a/go/vt/vtadmin/cluster/cluster.go
+++ b/go/vt/vtadmin/cluster/cluster.go
@@ -1362,14 +1362,11 @@ func (c *Cluster) GetSchemas(ctx context.Context, opts GetSchemaOptions) ([]*vta
 	}
 
 	var (
-		m   sync.Mutex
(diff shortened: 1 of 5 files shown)


falun:
Going to push back this -- it adds a bit of complexity for an unclear use case. Even original requestor was intending to exclude all keyspaces.

My suggestion if we want to go forward with something like this would be to add a noop cache that either never stores or (iirc it may be difficult to plumb that and keep things still working) treats all cache hit as if at is expired.

That said once original requestor discovered `DefaultExpiration` flag it was decided that that was sufficient and switched over to using that. We did get confirmation that the proactive refresh [in thread](https://vitess.slack.com/archives/C01H307F68J/p1662919616941229?thread_ts=1662676636.119309&cid=C01H307F68J) works as well. 

All told I think the complexity trade off + workable/better existing solutions means we should drop this as a potential path forward.

PR #10355, opened 2022-05-22
[vtadmin] cluster Closer implementation + dynamic Cluster bugfix

ajm188:
## Description

This PR enhances the `io.Closer` implementation of Cluster to also cleanly close any open gRPC connections (to vtgate/vtctld) and then use this when evicted old dynamic clusters, as well as address an issue in `api.clusters` management which I'll discuss in a bit!

The change is in 4 parts:
1. vtsql.Close was mutating `vtgate.conn`, which query-related methods would check and return `ErrConnClosed` if it was `nil`. This causes data races (ty `go test -race`). No one was explicitly relying on that named error anyway, so I removed it in favor of just letting queries fail if the gRPC connection gets closed out from under them (which produces a similar error string).

    a. Then, to continue supporting an `is_connected` field in the debug info, I added a `closed` flag which _is_ guarded by the mutex, and only touched in `dial` / `Close` / `Debug`, and most importantly _never_ in the hot path for queries.

2. Update `Cluster.Close` to close Vtctld + DB connections. Borrrrrrring
3. Update `EvictDynamicCluster` to `Close` a cluster before losing the reference to it. Not doing this was fine before, but after I added background resolvers and the schema cache, it would mean leaking both connections and goroutines as clusters are evicted over the lifetime of a vtadmin-api process.
4. While doing (3) I realized the way we were managing the `clusters` slice only worked if the cluster being evicted _happened_ to already be in slot 0; otherwise, we'd lose whatever cluster was in that slot for good and end up in a weird state. So I changed the logic to skip over `clusterIndex` rather than doing the `clusters[0]; clusters[1:]`.

    a. N.B. The following would also have worked:

    ```go
    api.clusters[0], api.clusters[clusterIndex] = api.clusters[clusterIndex], api.clusters[0]
    api.clusters = api.clusters[1:]
    // sort.ClustersBy(...)
    ```

## Checklist

### DIFF
--- a/go/vt/vtadmin/api.go
+++ b/go/vt/vtadmin/api.go
@@ -298,6 +298,19 @@ func (api *API) WithCluster(c *cluster.Cluster, id string) dynamic.API {
 			shouldAddCluster = shouldAddCluster || !isEqual
 		}
 		if shouldAddCluster {
+			if existingCluster != nil {
+				if err := existingCluster.Close(); err != nil {
+					log.Errorf("%s; some connections and goroutines may linger", err.Error())
+				}
+
+				idx := stdsort.Search(len(api.clusters), func(i int) bool {
+					return api.clusters[i].ID == existingCluster.ID
+				})
+				if idx >= 0 && idx < len(api.clusters) {
+					api.clusters = append(api.clusters[:idx], api.clusters[idx+1:]...)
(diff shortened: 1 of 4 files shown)


notfelineit (approved):
LGTM!

PR #10357, opened 2022-05-23
Add validation of check constraints

dbussink:
This adds validation of check constraints to schemadiff. It verifies that all referenced columns are actually present in the table.

Additionally, this implements the specific logic that a check constraint which references only a single column will automatically be dropped by MySQL when that column is dropped.

Any check that references multiple columns won't be dropped automatically and a drop column for such a column will fail. This is also matched in the behavior here.

## Related Issue(s)

Part of #10203

## Checklist

### DIFF
--- a/go/vt/schemadiff/table_test.go
+++ b/go/vt/schemadiff/table_test.go
@@ -17,7 +17,6 @@ limitations under the License.
 package schemadiff
 
 import (
-	"errors"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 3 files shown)


go/vt/schemadiff/table_test.go
```diff
@@ -1188,11 +1217,7 @@ func TestValidate(t *testing.T) {
 			applied, err := from.Apply(a)
 			if ts.expectErr != nil {
 				assert.Error(t, err)
-				if errors.Is(err, ts.expectErr) {
-					assert.ErrorIsf(t, err, ts.expectErr, "error mismatch. expected: %v, got: %v", ts.expectErr, err)
-				} else {
-					assert.EqualError(t, err, ts.expectErr.Error())
-				}
```
dbussink, line 1195: This was a left over that accidentally stayed from when I was refactoring this to the structured error types. Now that everything converted, this is no longer needed.

PR #10358, opened 2022-05-23
test refactor: extract const literal to file

systay:
## Description
Extracts two string consts into files, which should make them easier to work with from moden IDEs.

Is there any reason we should not move all our tests to use embedded files like this?

## Checklist


arthurschreiber:
Yes, this is exactly what I had in mind! 🤗

shlomi-noach:
I'm a fan of the idea, especially since this embedding happens in testing.

systay:
it even looks better here in github. syntax highlighting just works since gh can see through the file ending what the content should be

frouioui (approved):
this is really powerful, I love it! some hot takes:
- I don't think it is required, but maybe the `*.sql` and `*vschema.json` files need a header
- For future work, we could consider reusing certain schemas between different test packages
- We should transform every global variable to be private variables instead of a mix of public and private

Looks good to me!

PR #10367, opened 2022-05-25
Sanitize log messages

rsajwani:

## Description

This is an effort to sanitize and redact PII from error logging.

Important Note: I am not able to use *terseErrors (vtgate-config-terse-errors) , given its defined in vtgate package and using it in shared_buffer.go results in circular dependency. It will require some refactoring which I believe we can do in separate task.

## Checklist

### DIFF
--- a/go/vt/vtgate/errorsanitizer/errorSanitizer.go
+++ b/go/vt/vtgate/errorsanitizer/errorSanitizer.go
@@ -33,6 +33,19 @@ var (
 	reTruncateErrorAfter = regexp.MustCompile(`code = \S+|Duplicate entry|syntax error at position \d+`)
 )
 
+/* Error messages often have PII in them. That can come from vttablet or from mysql, so by the time it gets to vtgate,
+it's an opaque string. To avoid leaking PII into, we normalize error messages before sending them out.
+We remove five different sorts of strings:
+*/
+
+// * Anything after /code = \S+/
+// * Anything after /Duplicate entry/
+// * Anything after /syntax error at position \d+/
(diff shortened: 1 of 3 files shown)


vmg (approved):
Looks good! Thanks for the backport!

deepthi (commented):
Can you change the file names to follow the convention? underscores (aka snake case), not camelCase.
For the rest, @systay can you review as well before we merge?

go/vt/vtgate/errorsanitizer/errorSanitizer.go
```diff
@@ -0,0 +1,60 @@
...
+	// Truncate errors longer than this
+	maxErrorLength = 256
+)
+
+var (
+	// Remove any BindVars or Sql, and anything after it
+	reTruncateError = regexp.MustCompile(`[,:] BindVars:|[,:] Sql:`)
+
+	// Keep code = foo, Duplicate entry, or syntax error, but remove anything after it
+	reTruncateErrorAfter = regexp.MustCompile(`code = \S+|Duplicate entry|syntax error at position \d+`)
```
deepthi, line 33: If there are other error messages that don't match these patterns, we won't be truncating, right?

systay: that is my understanding

deepthi:
@rsajwani since the PR was merged without addressing my feedback, can you please create a follow up PR?

rsajwani:
> 
for renaming 

https://github.com/vitessio/vitess/pull/10406

PR #9740, opened 2022-02-20
make generate_ci_workflows: mysql80 workflows for selected tests

shlomi-noach:
## Description

`test/ci_workflow_gen.go` will now generate dual workflow files for selected tests. Specifically, all `onlineddl_*` tests will from now on have both `mysql57` and `mysql80` workflows. They require so because `mysql57` and `mysql80` have significant differences that affect online DDL, such as different types, handling of `utf`, as well as potential locking mechanisms (e.g. `autoinc` mode).

In `test/ci_workflow_gen.go`, we can now list, per test or per group of tests, whether the test runs on `mysql57` (the default, to be backwards compatible with the current implementation), or only on `mysql80`, or perhaps on both. In the latter case, `make generate_ci_workflows` generates two different workflow files per test cluster. For example:

- `.github/workflows/cluster_endtoend_onlineddl_revert.yml`
- `.github/workflows/cluster_endtoend_onlineddl_revert_mysql80.yml`

Again, for backwards compatibility, `mysql57` is considered the default, thus the file `.github/workflows/cluster_endtoend_onlineddl_revert.yml` uses `mysql57`. Otherwise, the name of the version is listed in the file name, as is the case for `.github/workflows/cluster_endtoend_onlineddl_revert_mysql80.yml`

## Related Issue(s)

- #6926 

## Checklist
- [ ] Should this PR be backported?
- [x] Tests were added or are not required
- [ ] Documentation was added or is not required

PR #10369, opened 2022-05-26
Online DDL: support for CHECK CONSTRAINTS

shlomi-noach:
## Description

MySQL `8.0.16` introduced [CHECK constraints](https://dev.mysql.com/doc/refman/8.0/en/create-table-check-constraints.html).

While syntactically a `CHECK` constraint is part of a table, the _name_ of the constraint is _unique to schema_:

> The SQL standard specifies that all types of constraints (primary key, unique index, foreign key, check) belong to the same namespace. In MySQL, each constraint type has its own namespace per schema (database). Consequently, CHECK constraint names must be unique per schema; no two tables in the same schema can share a CHECK constraint name. (Exception: A TEMPORARY table hides a non-TEMPORARY table of the same name, so it can have the same CHECK constraint names as well.)

This has an important impact to Online DDL: as we create the shadow table, that table cannot have constraints of same names as the original table.

Up until this PR, this was auto-managed by MySQL's `CREATE TABLE ... LIKE ...` statement. For example:
```sql
drop table if exists t;
CREATE TABLE `t` (
  `id` int NOT NULL,
  `i` int NOT NULL DEFAULT '0',
  PRIMARY KEY (`id`),
  CONSTRAINT `check_even` CHECK (`i` % 2 = 0)
);

create table t2 like t;

show create table t2\G
CREATE TABLE `t2` (
  `id` int NOT NULL,
  `i` int NOT NULL DEFAULT '0',
  PRIMARY KEY (`id`),
  CONSTRAINT `t2_chk_1` CHECK (((`i` % 2) = 0))
```

In the above we can see MySQL chose arbitrary new names for the constraints. See the docs for more about naming.

This means some migrations will not work well. For example: `ALTER TABLE t DROP CHECK check_even` will fail. That is because we don't actually modify the original table, we modify the shadow table. But on the shadow table there is no constraint named `check_even`.

As of this PR, we take more control over the creation of the shadow table. Instead of letting MySQL generate the table via `CREATE TABLE ... LIKE ...` we utilize the parser to read the `CreateTable` AST of the original table, and modify it to generate a `CREATE TABLE` statement for the shadow table. We replace constraint names such that:

- New names are unique
- New names are deterministic (so that the same name is generated on all shards in a sharded environment)
- New names include the original constraint name, up to the max 64 character limit of the `CHECK` constraint's name

Then, we also parse the `AlterTable` AST and modify any references to `CHECK` constraints (or, any constraints, really), and swap old names for new names.

Along the way, this PR gets rid of some very old parsing code, which was needed at the time the Vitess parser did not have sufficient support for all DDL statements. We remove a ton of `regexp` based parsing, and switch over to using the formal AST.

**Notice** that either before or after this PR, check constraint names change on every single migration, even if the migration had nothing to do with the constraints. This is an unfortunate side effect of constraints being unique across the schema.

Tests added. They are only relevant to `8.0` setups. Related work: https://github.com/vitessio/vitess/pull/9740

## Related Issue(s)

- #6926 

## Checklist


shlomi-noach:
Illustrating constraint naming behavior:

Assume this original table:
```sql
CREATE TABLE `t` (
  `id` int NOT NULL,
  `test` int NOT NULL DEFAULT '0',
  PRIMARY KEY (`id`),
  CONSTRAINT `check1` CHECK ((`test` >= 0)),
  CONSTRAINT `check2` CHECK ((`test` >= 0)),
  CONSTRAINT `check3` CHECK ((`test` >= 0)),
  CONSTRAINT `check4` CHECK ((`test` <> 4)),
  CONSTRAINT `check5` CHECK ((`test` < 5))
```

Say we do a trivial `ALTER TABLE t ENGINE=InnoDB` migration. The table may turn to:
```sql
CREATE TABLE `t` (
  `id` int NOT NULL,
  `test` int NOT NULL DEFAULT '0',
  PRIMARY KEY (`id`),
  CONSTRAINT `chk_04a5b01db9d15e218979970c23bfc13a_check1` CHECK ((`test` >= 0)),
  CONSTRAINT `chk_2dbf8418af7250d783beb208344ae16b_check5` CHECK ((`test` < 5)),
  CONSTRAINT `chk_353c799af3d858eb99b87d6419133d5c_check2` CHECK ((`test` >= 0)),
  CONSTRAINT `chk_5b7f19d3521f59bd85045742d6d95913_check3` CHECK ((`test` >= 0)),
  CONSTRAINT `chk_d3ea4658fb83587f9beae3064c354041_check4` CHECK ((`test` <> 4))
```

- Notice all names have changed, even though the migration had nothing to do specifically with `CHECK` constraints.
- Also notice the original check name is preserved as a suffix in the `CHECK`'s name

Say we _again_ run a trivial  `ALTER TABLE t ENGINE=InnoDB` migration. We will get something like:

```sql
CREATE TABLE `t` (
  `id` int NOT NULL,
  `test` int NOT NULL DEFAULT '0',
  PRIMARY KEY (`id`),
  CONSTRAINT `chk_326258d255a154229107a02402b2b6f1_check1` CHECK ((`test` >= 0)),
  CONSTRAINT `chk_74954dcba13f5deda55006e52b1d5004_check5` CHECK ((`test` < 5)),
  CONSTRAINT `chk_87ef2bbaac825426a5f9ff8ad88d8b4a_check3` CHECK ((`test` >= 0)),
  CONSTRAINT `chk_9148a507c23d5275845b208a7eb52a57_check4` CHECK ((`test` <> 4)),
  CONSTRAINT `chk_d77103119cf35487aa954c488f66f736_check2` CHECK ((`test` >= 0))
```

- again, all names have changed
- and still, original names are preserved.

shlomi-noach:
Switched to `Draft` while investigating failing tests.

shlomi-noach:
Tests are all good. Added release notes. This is good to review.

shlomi-noach:
`Upgrade Downgrade Testing Query Serving` is extremely flaky today; I'm going to ignore it.

go/test/endtoend/cluster/vttablet_process.go
```diff
@@ -423,8 +423,11 @@ func executeQuery(dbConn *mysql.Conn, query string) (*sqltypes.Result, error) {
 	)
 	retries := 10
 	retryDelay := 1 * time.Second
-	for i := 1; i <= retries; i++ {
-		log.Infof("Executing query %s (attempt %d of %d)", query, i, retries)
+	for i := 0; i < retries; i++ {
```
mattlord, line 426: 👍

go/test/endtoend/cluster/vttablet_process.go
```diff
@@ -423,8 +423,11 @@ func executeQuery(dbConn *mysql.Conn, query string) (*sqltypes.Result, error) {
...
+		if i > 0 {
+			// We only audit from 2nd attempt and onwards, otherwise this is just too verbose.
+			log.Infof("Executing query %s (attempt %d of %d)", query, (i + 1), retries)
+		}
```
mattlord, line 430: 👍

mattlord, line 430: 👍

go/vt/vttablet/onlineddl/executor.go
```diff
@@ -888,6 +891,94 @@ func (e *Executor) initMigrationSQLMode(ctx context.Context, onlineDDL *schema.O
...
+	}
+	return constraintMap, nil
+}
+
+// validateAndEditAlterTableStatement inspects the AlterTable statement and:
+// - modifies any CONSTRAINT name according to given name mapping
+func (e *Executor) validateAndEditAlterTableStatement(ctx context.Context, alterTable *sqlparser.AlterTable, constraintMap map[string]string) (err error) {
+	validateWalk := func(node sqlparser.SQLNode) (kontinue bool, err error) {
+		switch node := node.(type) {
+		case *sqlparser.DropKey:
```
mattlord, line 956: Side note: I would have expected this to be DropConstraint — for FK and CHECK constraints — as they are rules rather than indexes, but I see we don't have one. Perhaps it's not needed...

mattlord, line 956: Side note: I would have expected this to be DropConstraint (for FK and CHECK constraints), but I see we don't have one. Perhaps it's not needed...

go/vt/vttablet/onlineddl/executor_test.go
```diff
@@ -19,3 +19,107 @@ Functionality of this Executor is tested in go/test/endtoend/onlineddl/...
...
+			err = sqlparser.Walk(func(node sqlparser.SQLNode) (kontinue bool, err error) {
+				switch node := node.(type) {
+				case *sqlparser.ConstraintDefinition:
+					uniqueConstraintNames[node.Name.String()] = true
+				}
+				return true, nil
+			}, createTable)
+			assert.NoError(t, err)
+			assert.Equal(t, tc.countConstraints, len(uniqueConstraintNames))
+			assert.Equal(t, tc.countConstraints, len(constraintMap))
```
mattlord, line 122: Side note: curious why you use assert here rather than require? I haven't figured out clear reasons when to use one over the other and maybe you have. :-)

mattlord, line 122: Side note: curious why you use assert here rather than require? I haven't figured out clear reasons when to use one over the other and maybe you have. :-)

go/vt/vttablet/onlineddl/vrepl/parser.go
```diff
@@ -21,19 +21,11 @@ limitations under the License.
...
 
-	"vitess.io/vitess/go/vt/schema"
-)
-
-var (
-	sanitizeQuotesRegexp = regexp.MustCompile("('[^']*')")
-	renameColumnRegexp   = regexp.MustCompile(`(?i)\bchange\s+(column\s+|)([\S]+)\s+([\S]+)\s+`)
-	dropColumnRegexp     = regexp.MustCompile(`(?i)\bdrop\s+(column\s+|)([\S]+)$`)
-	renameTableRegexp    = regexp.MustCompile(`(?i)\brename\s+(to|as)\s+`)
-	autoIncrementRegexp  = regexp.MustCompile(`(?i)\bauto_increment[\s]*[=]?[\s]*([0-9]+)`)
```
mattlord, line 36: 🥳

mattlord, line 36: 🥳

mattlord (approved):
LGTM!

go/textutil/hash.go
```diff
@@ -20,15 +20,30 @@ import (
...
-	rb := make([]byte, size)
+	rb := make([]byte, randomHashSize)
 	_, _ = rand.Read(rb)
 
 	hasher := sha256.New()
 	hasher.Write(rb)
 	return hex.EncodeToString(hasher.Sum(nil))
 }
+
+// UUID5 creeates a UUID v5 string based on the given inputs. We use a SHA256 algorithm.
```
deepthi, line 42:
```suggestion
// UUIDv5 creeates a UUID v5 string based on the given inputs. We use a SHA256 algorithm.
```

go/vt/vttablet/onlineddl/executor.go
```diff
@@ -888,6 +891,94 @@ func (e *Executor) initMigrationSQLMode(ctx context.Context, onlineDDL *schema.O
...
+		return err
+	}
+	return nil
+}
+
+// initVreplicationOriginalMigration performs the first steps towards running a VRepl ALTER migration:
+// - analyze the original table
+// - formalize a new CreateTable statement
+// - inspect the ALTER TABLE query
+// - formalize a AlterTable statement
```
deepthi, line 978:
```suggestion
// - formalize an AlterTable statement
```

deepthi (approved):
Your PRs are always a pleasure to read. Found a couple of typos, otherwise LGTM

go/vt/vttablet/onlineddl/executor.go
shlomi-noach: I agree, `DropKey` is confusing.

go/vt/vttablet/onlineddl/executor_test.go
shlomi-noach: Great question! `require` completely bails out the test when the value does not match the expected value. I only `require` when there's no programmatic point in pursuing the next step. E.g. if the value is `nil` then further attempting to access some fields will yield in nil reference exception / segmentation fault. But I mostly prefer `assert`: often times, the followup tests can shed some more light on _why_ exactly the assertion failed. It's nice to know that some error is in fact nil when we expected it to be not-nil - but why? The first tests are usually simplistic (`NotNil`, `NotEmpty`, `True`, etc.) where followup tests drill down into details.

go/vt/vttablet/onlineddl/executor.go

go/textutil/hash.go
shlomi-noach: fixed

shlomi-noach:
Change of format, see https://github.com/vitessio/vitess/pull/10380#issuecomment-1140898760

# Change of constraint name format

To allow more character space for original names, and to maintain original name order, the new constraint name format is:

`<original name>_[0-9a-z]{25}`

The checksum is a `128` bit (`UUIDv5` based on `SHA256`) encoded in `base36`, which is `25` characters (down from `32`). We remove `chk_` or `fk_` because those are superfluous to the constraint definition.

### Example:

original table:
```sql
CREATE TABLE `t` (
  `id` int NOT NULL,
  `test` int NOT NULL DEFAULT '0',
  PRIMARY KEY (`id`),
  CONSTRAINT `check1` CHECK ((`test` >= 0)),
  CONSTRAINT `check2` CHECK ((`test` >= 0)),
  CONSTRAINT `check3` CHECK ((`test` >= 0)),
  CONSTRAINT `check4` CHECK ((`test` <> 4)),
  CONSTRAINT `check5` CHECK ((`test` < 5))
)
```

modified once:

```sql
CREATE TABLE `t` (
  `id` int NOT NULL,
  `test` int NOT NULL DEFAULT '0',
  PRIMARY KEY (`id`),
  CONSTRAINT `check1_19l09s37kbhj4axnzmi10e18k` CHECK ((`test` >= 0)),
  CONSTRAINT `check2_cukwabxd742sgycn96xj7n87g` CHECK ((`test` >= 0)),
  CONSTRAINT `check3_d9ict50651cj9qzlzmqxjrerq` CHECK ((`test` >= 0)),
  CONSTRAINT `check4_0j5zhr7pheqtftoc6jyib997o` CHECK ((`test` <> 4)),
  CONSTRAINT `check5_ccqceb4vzonxgy9plzynofyr3` CHECK ((`test` < 5))
)
```

modified again:

```sql
CREATE TABLE `t` (
  `id` int NOT NULL,
  `test` int NOT NULL DEFAULT '0',
  PRIMARY KEY (`id`),
  CONSTRAINT `check1_cps1okb4uafunfqusi2lp22u3` CHECK ((`test` >= 0)),
  CONSTRAINT `check2_atleezf4dlbjvon55mflylhjb` CHECK ((`test` >= 0)),
  CONSTRAINT `check3_8nj6a51pszqbtbycrm18qri0n` CHECK ((`test` >= 0)),
  CONSTRAINT `check4_dvc5bs03r4twlyo2a5i76ll8k` CHECK ((`test` <> 4)),
  CONSTRAINT `check5_b7c76z82dbqpp90bv3ywokzqo` CHECK ((`test` < 5))
)
```

You may see how original names are clearer, and constraint names consistently sorted by original names.

PR #10371, opened 2022-05-26
Revert the keep-data flag default value to false

GuptaManan100:
## Description

We have recently noticed CI failures, complaining with `no space left on device`. In order to fix this issue, we are reverting the change we made to the default value of `-keep-data` flag. For the self-hosted runners, this flag is being provided as part of the command, so no change is required there.
After this change, GitHub-runners will not keep the data of the runs (it wasn't helping before either, since we couldn't access them), but the self-hosted runners will continue to do so.

## Related Issue(s)
- Failures like https://github.com/vitessio/vitess/runs/6580642446?check_suite_focus=true

## Checklist

### DIFF
--- a/go/test/endtoend/cluster/cluster_process.go
+++ b/go/test/endtoend/cluster/cluster_process.go
@@ -46,7 +46,7 @@ const (
 )
 
 var (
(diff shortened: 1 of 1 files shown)


harshit-gangal (approved):
LGTM

PR #10372, opened 2022-05-26
Add local devcontainer

ajm188:
## Description

**Disclaimer!** I have never done this before, and I'm not sure about to do devcontainers in a project as large and multifaceted as vitess, but this is both useful and working (plus, if I may humbly add, pretty cool), so I wanted to at least try to contribute it back up.

This adds a devcontainer capable of running the local examples (including vtadmin!), edit Go files and run tests, and commit changes (N.B. the source tree is volume-mounted so you can also just commit outside the devcontainer, if we have Feelings about that `git` installation).

You'll have to kinda take my word for it, but here's safari on my laptop pointed at vtadmin-web running with the full local example inside a devcontainer spawned from this branch.

Screen Shot 2022-05-25 at 3 08 17 PM

My understanding is that we can also use this container for code-spaces, to do in-browser/cloud Vitess development, which is potentially very cool.

## Checklist

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,62 @@
+/*
+A devcontainer capable of running the local examples (in ../examples/local),
+including vtadmin-api and vtadmin-web.
+
+Note that all paths are relative to the .devcontainer directory, so ".." is the
+repository root.
+
+To run:
+1. In vscode's command palette, select "Remote-Containers: Open Folder in Container"
+2. Select the root vitess checkout from the file finder.
+3. Wait for the container to build. This can take a while.
+   ....... keep waiting
+   At this point vscode will open a new window which is the devcontainer.
+4. Accept the prompt to install go-tools (gopls, staticcheck, and so on).
+   You can also use the command palette from this window to install/update them
+   just like you would in any old vscode instance.
+5. To run the local example:
+   a. cd examples/local
+   b. ./101_initial_cluster.sh
+   c. HOST=127.0.0.1 ./scripts/vtadmin-up.sh
+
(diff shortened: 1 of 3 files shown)


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,62 @@
...
+3. Wait for the container to build. This can take a while.
+   ....... keep waiting
+   At this point vscode will open a new window which is the devcontainer.
+4. Accept the prompt to install go-tools (gopls, staticcheck, and so on).
+   You can also use the command palette from this window to install/update them
+   just like you would in any old vscode instance.
+5. To run the local example:
+   a. cd examples/local
+   b. ./101_initial_cluster.sh
+   c. HOST=127.0.0.1 ./scripts/vtadmin-up.sh
```
deepthi, line 20: vtadmin-up.sh is now included in the 101 script, right?

deepthi (commented):
Any new tool/example we add might require ongoing maintenance. Is this something we think has enough of a benefit for us to take that on?
If we agree that yes, we want to, then you'll want to update this PR and we can ask someone who uses vscode to review.

.devcontainer/devcontainer.json
frouioui: it is on main

frouioui: maybe we can add the `HOST=127.0.0.1` variable when we execute `101_initial_cluster.sh` so we can still specify the host for vtadmin and remove the call to `vtadmin-up.sh` (after we merge main onto this branch)

frouioui (commented):
I was not able to run it successfully on my VSCode on the first try. I will try again in a later time.

PR #10373, opened 2022-05-27
vt/wrangler: fix deleting primary tablet if `Shard` does not exist

fatih:
This PR makes sure that `wrangler.DeleteTablet()` no longer returns an error when deleting a Primary if a `Shard` no longer exists. A `Shard` might not exist, when for example someone explicitly deletes it before calling `wrangler.DeleteTablet()`. By making this change, the `wrangler.DeleteTablet()` function is also idempotent, and safe to use in reconciling loops (such as in Kubernetes Operators).

### DIFF
--- a/go/vt/wrangler/tablet.go
+++ b/go/vt/wrangler/tablet.go
@@ -65,7 +65,7 @@ func (wr *Wrangler) DeleteTablet(ctx context.Context, tabletAlias *topodatapb.Ta
 		defer unlock(&err)
 
 		// update the shard record's primary
-		if _, err := wr.ts.UpdateShardFields(ctx, ti.Keyspace, ti.Shard, func(si *topo.ShardInfo) error {
+		_, err := wr.ts.UpdateShardFields(ctx, ti.Keyspace, ti.Shard, func(si *topo.ShardInfo) error {
 			if !topoproto.TabletAliasEqual(si.PrimaryAlias, tabletAlias) {
 				wr.Logger().Warningf("Deleting primary %v from shard %v/%v but primary in Shard object was %v", topoproto.TabletAliasString(tabletAlias), ti.Keyspace, ti.Shard, topoproto.TabletAliasString(si.PrimaryAlias))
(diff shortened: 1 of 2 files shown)


rohit-nayak-ps (dismissed):
@fatih, I tried the new unit test locally and it failed. `DeleteTablet()` fails before it reaches your new error handling at `go/vt/wrangler/tablet.go:52` because the shard doesn't exist anymore.

```
	wasPrimary, err := wr.isPrimaryTablet(ctx, ti)
	if err != nil {
		return err
	}
```

fatih:
@rohit-nayak-ps I see. It looks like `isPrimaryTablet`is also calling `GetShard`, so it's worth adding the same modification there too. I'm going to push another fix now.

fatih:
It passes locally:

```
$ go test
...
PASS
ok      vitess.io/vitess/go/vt/wrangler 5.003s
```

But I'm not sure why it's failing in the CI.

go/vt/wrangler/tablet.go
```diff
@@ -46,7 +46,7 @@ func (wr *Wrangler) DeleteTablet(ctx context.Context, tabletAlias *topodatapb.Ta
 	}
 
 	wasPrimary, err := wr.isPrimaryTablet(ctx, ti)
-	if err != nil {
+	if err != nil && !topo.IsErrType(err, topo.NoNode) {
```
rohit-nayak-ps, line 49: If we find that the shard is no longer present, should we just exit the function at this point and not proceed or do we want it to try to cleanup the topo tablet record?  The `topotools.DeleteTablet()`, which will get called if we ignore the error, updates some shard related topo entries and it is not clear if it will fail in case the shard has already been deleted.

rohit-nayak-ps:
> But I'm not sure why it's failing in the CI.

Probably flakiness, I retried them all and the required tests passed.

rohit-nayak-ps (approved):
nice, lgtm

PR #10374, opened 2022-05-27
Vtexplain no globals

systay:
## Description
Minor refactoring of vtexplain, removing global variables that did not need to be global.

## Checklist


go/cmd/vtexplain/vtexplain.go
```diff
@@ -113,7 +113,6 @@ func getFileParam(flag, flagFile, name string, required bool) (string, error) {
 }
 
 func main() {
-	defer vtexplain.Stop()
```
GuptaManan100, line 116: Shouldn't we call vte.Stop() or some other equivalent of this in the end?

GuptaManan100 (approved):

go/cmd/vtexplain/vtexplain.go
systay: We are. I just moved the defer to after we have done the Init

ajm188:
as a followup, maybe `NewExecutor` is now a better name than `Init` for that function?

systay:
@ajm188 sounds good to me

PR #10375, opened 2022-05-27
Refactor Durability Policy implementation and usage to read the durability policy from the keyspace

GuptaManan100:
## Description

This PR exports the Durabler interface and moves away from the architecture of having a single global durability policy for a binary.
All the RPCs that need to use the durability policy, now first consult the topo server to read the keyspace information to find the durability_policy. Once they have the durability policy, they get a new Durabler and use that for all the other operations needing durability policies.

As part of the PR, changes to the local and legacy local examples has also been made to set the correct durability policies for the keyspaces that they create.

## Related Issue(s)

- #8975 

## Checklist

-   [x] Documentation was added or is not required - https://github.com/vitessio/vitess/pull/10392

### DIFF
--- a/go/vt/orchestrator/inst/durability.go
+++ b/go/vt/orchestrator/inst/durability.go
@@ -17,28 +17,65 @@ limitations under the License.
 package inst
 
 import (
+	topodatapb "vitess.io/vitess/go/vt/proto/topodata"
 	"vitess.io/vitess/go/vt/vtctl/reparentutil"
+	"vitess.io/vitess/go/vt/vtctl/reparentutil/promotionrule"
 )
 
+var (
+	// durabilityPolicy is the durability policy in use for VTOrc
+	durabilityPolicy reparentutil.Durabler
+)
+
+// SetDurabilityPolicy is used to set the durability policy for VTOrc once during startup
+func SetDurabilityPolicy(name string) (err error) {
+	durabilityPolicy, err = reparentutil.GetDurabilityPolicy(name)
+	return err
+}
+
 // IsReplicaSemiSync returns the replica semi-sync setting for the instance.
-func IsReplicaSemiSync(primaryKey, replicaKey InstanceKey) bool {
-	primary, err := ReadTablet(primaryKey)
+func IsReplicaSemiSync[V InstanceKey | *topodatapb.Tablet](primaryInstance V, replicaInstance V) bool {
+	primary, err := getTablet(primaryInstance)
 	if err != nil {
 		return false
 	}
-	replica, err := ReadTablet(replicaKey)
+	replica, err := getTablet(replicaInstance)
 	if err != nil {
 		return false
 	}
-	return reparentutil.IsReplicaSemiSync(nil, primary, replica)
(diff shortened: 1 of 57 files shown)


GuptaManan100:
https://github.com/vitessio/vitess/pull/10393 PR is used for checking that the durability policy changes will continue to work after we start using them for the semi-sync settings instead of `enable_semi_sync` flag. As can be seen on that PR, all the tests are passing on it.

go/vt/vtctl/reparentutil/planned_reparenter_flaky_test.go
```diff
@@ -1641,6 +1641,10 @@ func TestPlannedReparenter_performGracefulPromotion(t *testing.T) {
 				ctx = _ctx
 			}
 
+			durability, err := GetDurabilityPolicy("none")
+			require.NoError(t, err)
+			tt.opts.durability = durability
+
```
deepthi, line 1647: Do we have any unit tests for the other policies? All the tests in this file seem to be using "none".
If not, it will be good to add some (a follow up PR is fine for that).

deepthi (approved):
Nice work! It's all coming together well now :)

go/vt/orchestrator/inst/durability.go
```diff
@@ -17,28 +17,65 @@ limitations under the License.
...
+// SetDurabilityPolicy is used to set the durability policy for VTOrc once during startup
+func SetDurabilityPolicy(name string) (err error) {
+	durabilityPolicy, err = reparentutil.GetDurabilityPolicy(name)
+	return err
+}
+
 // IsReplicaSemiSync returns the replica semi-sync setting for the instance.
-func IsReplicaSemiSync(primaryKey, replicaKey InstanceKey) bool {
-	primary, err := ReadTablet(primaryKey)
+func IsReplicaSemiSync[V InstanceKey | *topodatapb.Tablet](primaryInstance V, replicaInstance V) bool {
```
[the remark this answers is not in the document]
GuptaManan100: Yes! Earlier we were using two different functions for it, it made sense to use generics now that we have them!

go/vt/vtctl/reparentutil/planned_reparenter_flaky_test.go
GuptaManan100: Planned reparent shard isn't using the durability policy for any revocation or establishment logic. It is only using the policies for knowing who sends semi-sync acks, but that is something none of the tests have, so any durability policy is fine for these tests. But, I'll add one more test which tests semi-sync based on the durability-policy being used in a follow up PR!

GuptaManan100:
Website docs - https://github.com/vitessio/website/pull/1040

go/vt/vtctl/reparentutil/planned_reparenter_flaky_test.go
GuptaManan100: As promised, this test for checking that the correct value of semi-sync is sent from PRS has been added - https://github.com/vitessio/vitess/pull/10425

PR #10377, opened 2022-05-27
Ensure that semi-sync plugin does not block primary demotion

mattlord:

## Checklist

### DIFF
--- a/go/vt/vttablet/tabletmanager/rpc_replication.go
+++ b/go/vt/vttablet/tabletmanager/rpc_replication.go
@@ -392,6 +392,7 @@ func (tm *TabletManager) demotePrimary(ctx context.Context, revertPartialFailure
 	if err != nil {
 		return nil, err
 	}
+	wasSemiSyncSourceEnabled, wasSemiSyncReplicaEnabled := tm.MysqlDaemon.SemiSyncEnabled()
 
 	// If we are a primary tablet and not yet read-only, stop accepting new
 	// queries and wait for in-flight queries to complete. If we are not primary,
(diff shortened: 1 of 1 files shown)


go/vt/vttablet/tabletmanager/rpc_replication.go
```diff
@@ -419,8 +420,30 @@ func (tm *TabletManager) demotePrimary(ctx context.Context, revertPartialFailure
...
+		//   https://bugs.mysql.com/bug.php?id=104015
+		//   https://bugs.mysql.com/bug.php?id=101056
+		// This in turns causes the statement to disable [Super]ReadOnly mode to block on a mutex that
+		// the semi-sync plugin is holding while waiting for the ACK. This then can cause the shard to
+		// be left in an unhealthy state without a serving PRIMARY. So here we need to disable SemiSync
+		// which will ensure that we are unblocked by this scenario (tablet repair should also fix this
+		// as needed via fixSemiSync()).
+		if wasSemiSyncSourceEnabled {
+			log.Infof("DemotePrimary disabling source semi-sync")
+			if err := tm.MysqlDaemon.SetSemiSyncEnabled(false, wasSemiSyncReplicaEnabled); err != nil {
```
deepthi, line 435: It is safe to do this because we have already shut down the tablet server by calling `SetServingType(..., serving=false, ...)`. Tablet Server shutdown gives a grace period to open transactions and then force-kills them. Even if a grace period is not set, a transaction that is hanging because of the semi-sync bug will get killed after the transaction timeout.
Nice work @mattlord

deepthi (commented):
It's probably non-trivial to write an actual test case for this. Can you look into whether it is feasible? We know existing unit tests / endtoend tests will cover this code path, yet we have no easy way to test that semi-sync was indeed toggled the way we want it to be.

go/vt/vttablet/tabletmanager/rpc_replication.go
```diff
@@ -419,8 +420,30 @@ func (tm *TabletManager) demotePrimary(ctx context.Context, revertPartialFailure
...
+				return nil, vterrors.Wrapf(err, "SetSemiSyncEnabled(source=false, replica=%v) failed", wasSemiSyncReplicaEnabled)
+			}
```
GuptaManan100, line 437: I don't think we need these changes, we have code to stop the semi-sync primary side just below the set read only part. We should just reorder them, because this code makes it so, that that code will not run

GuptaManan100: I have created a PR for these changes - https://github.com/vitessio/vitess/pull/10383. WDYT @deepthi and @mattlord about reordering the statements instead of adding these which essentially accomplish the same things?

mattlord: Sounds good! I’ll close this one

mattlord:
Closing this in favor of https://github.com/vitessio/vitess/pull/10383

go/vt/vttablet/tabletmanager/rpc_replication.go
deepthi: The perils of looking at GH diffs without looking at the rest of the file 😭

GuptaManan100: I know right. I knew that we had code for semi-sync changes down below because I worked with it for the semi-sync changes for vtorc

PR #10382, opened 2022-05-29
Vdiff2: initial release

rohit-nayak-ps:
VDiff2: VDiff on tablets

This is the initial PR that moves vdiff to tablets. RFC is at https://github.com/vitessio/vitess/issues/10134 and some implementation details regarding implementation at `https://github.com/vitessio/vitess/issues/10134`

## Implemented
* VDiff runs on the primary of each shard because it needs to store state (for progress, resumability, status etc). Note that the data on the targets can get pulled from a replica in that shard, but the actual snapshotting and comparison is done only on the primary.
* It runs as a new Engine in the scope of the TabletManager.
* Major change is that we only compare data that is present on each shard: so only that data that is vreplicated to a particular shard is pulled by that shard's primary.
* A new VDiff tablet GRPC call implements all the VDiff actions and is invoked by vtctld for each of the primaries in a workflow.
* The snapshoting logic to get a logically equal source and target database, as well as the data comparison code is identical with VDiff2
* The original code was in a single directory, but has been now organized logically across files for better readability.

## Path to Production
This is a "Preview" version of vdiff2 intended to make it into Vitess v14. We hope to get feedback on this from users who can also run vdiff2 in parallel with vdiff to test both functionality and observe performance improvements, find bugs etc ...

In v15 we will release an alpha version and then a beta version which will include:

## Alpha version

* Support Mount/Migrate
* More manual testing
* Move current test cases over: this will be a non-trivial port of the mock framework from vtctld to vttablet
* Check coverage and add test cases for filling the gaps
* Enhance e2e tests
* Add Create action argument to wait for the VDiff to complete before returning, showing a progress meter of some kind
* Add Progress field to Show output when the Status == started to display a % of work done and ETA for completion
* Add Cancel action
* Record and display LastError and LastErrorNo values for a VDiff
    * Use this error number to drive behavior, e.g. move to the `error` state if we encountered a non-ephemeral/non-recoverable error 

### Resumable VDiffs

* Support stopping/restarting vdiffs
* Automatic recovery from network errors, PRS, tablet restarts
* Supporting vtctld commands for above
* VTAdmin to control and monitor VDiff2 workflows


mattlord (approved):
LGTM! 🥳 🥳

PR #10383, opened 2022-05-30
Reorder Operations in DemotePrimary to ensure that semi-sync plugin does not block primary demotion

GuptaManan100:
## Description

If for whatever reason a primary is network partitioned or does not receive the semi-sync acks from its replicas, it can even block the command used to set [super]ReadOnly. This blocks the call to DemotePrimary as well until it times-out.
The fix for this issue is simple, we reorder the commands to first stop primary side semi-sync after stopping query service but before we issue the commands for setting [super] read only. 

## Checklist


GuptaManan100:
This PR supersedes #10377

mattlord:
If we can easily piggy back on an existing test, then I think it's worth adding one. You can manually repeat the problem on main this way from the base of the git repo:
```
cd example/local

sed -i '58i \  --enable_semi_sync \\' scripts/vttablet-up.sh

./101_initial_cluster.sh

for i in {1,2}; do mysql -u root --socket=/vt/vtdataroot/vt_000000010${i}/mysql.sock -e "STOP REPLICA IO_THREAD; CHANGE REPLICATION SOURCE TO SOURCE_HEARTBEAT_PERIOD=0; START REPLICA IO_THREAD"; done

kill -9 $(cat /vt/vtdataroot/vt_0000000101/mysql.pid) $(cat /vt/vtdataroot/vt_0000000102/mysql.pid)

mysql -u root --socket=/vt/vtdataroot/vt_0000000100/mysql.sock vt_commerce -e "insert into customer values (2, 'ml')" &

mysql -u root --socket=/vt/vtdataroot/vt_0000000100/mysql.sock -e "set global super_read_only=ON" &

mysql -u root --socket=/vt/vtdataroot/vt_0000000100/mysql.sock -e "select @@version; show processlist;"
```

Final results being:
```
+-----------+
| @@version |
+-----------+
| 8.0.29    |
+-----------+
+----+-----------------+---------------------+-------------+------------------+------+-----------------------------------------------------------------+-------------------------------------------------------------------------------+
| Id | User            | Host                | db          | Command          | Time | State                                                           | Info                                                                          |
+----+-----------------+---------------------+-------------+------------------+------+-----------------------------------------------------------------+-------------------------------------------------------------------------------+
|  5 | event_scheduler | localhost           | NULL        | Daemon           |   19 | Waiting on empty queue                                          | NULL                                                                          |
| 10 | vt_dba          | localhost           | NULL        | Sleep            |    5 |                                                                 | NULL                                                                          |
| 12 | vt_dba          | localhost           | NULL        | Sleep            |    5 |                                                                 | NULL                                                                          |
| 14 | vt_dba          | localhost           | NULL        | Sleep            |    5 |                                                                 | NULL                                                                          |
| 16 | vt_dba          | localhost           | NULL        | Sleep            |    5 |                                                                 | NULL                                                                          |
| 22 | vt_dba          | localhost           | NULL        | Sleep            |    5 |                                                                 | NULL                                                                          |
| 23 | vt_dba          | localhost           | NULL        | Sleep            |    5 |                                                                 | NULL                                                                          |
| 29 | vt_dba          | localhost           | NULL        | Sleep            |    5 |                                                                 | NULL                                                                          |
| 30 | vt_dba          | localhost           | NULL        | Sleep            |    5 |                                                                 | NULL                                                                          |
| 34 | vt_dba          | localhost           | NULL        | Sleep            |    5 |                                                                 | NULL                                                                          |
| 35 | vt_dba          | localhost           | NULL        | Sleep            |    5 |                                                                 | NULL                                                                          |
| 36 | vt_dba          | localhost           | NULL        | Sleep            |    5 |                                                                 | NULL                                                                          |
| 37 | vt_dba          | localhost           | NULL        | Sleep            |    5 |                                                                 | NULL                                                                          |
| 38 | vt_dba          | localhost           | NULL        | Sleep            |    5 |                                                                 | NULL                                                                          |
| 42 | vt_dba          | localhost           | vt_commerce | Sleep            |    4 |                                                                 | NULL                                                                          |
| 43 | vt_app          | localhost           | vt_commerce | Sleep            |    6 |                                                                 | NULL                                                                          |
| 45 | vt_dba          | localhost           | NULL        | Sleep            |    5 |                                                                 | NULL                                                                          |
| 46 | vt_dba          | localhost           | NULL        | Sleep            |    5 |                                                                 | NULL                                                                          |
| 49 | vt_dba          | localhost           | NULL        | Sleep            |    5 |                                                                 | NULL                                                                          |
| 50 | vt_dba          | localhost           | NULL        | Sleep            |    4 |                                                                 | NULL                                                                          |
| 52 | vt_dba          | localhost           | NULL        | Sleep            |    5 |                                                                 | NULL                                                                          |
| 53 | vt_dba          | localhost           | NULL        | Sleep            |    5 |                                                                 | NULL                                                                          |
| 54 | vt_dba          | localhost           | NULL        | Sleep            |    5 |                                                                 | NULL                                                                          |
| 61 | vt_dba          | localhost           | vt_commerce | Sleep            |    5 |                                                                 | NULL                                                                          |
| 63 | vt_dba          | localhost           | vt_commerce | Sleep            |    5 |                                                                 | NULL                                                                          |
| 64 | vt_dba          | localhost           | NULL        | Query            |    2 | Waiting for global read lock                                    | ALTER TABLE _vt.schema_migrations add column stowaway_table tinytext NOT NULL |
| 66 | vt_repl         | 172.31.41.160:38586 | NULL        | Binlog Dump GTID |    3 | Source has sent all binlog to replica; waiting for more updates | NULL                                                                          |
| 67 | vt_repl         | 172.31.41.160:38588 | NULL        | Binlog Dump GTID |    3 | Source has sent all binlog to replica; waiting for more updates | NULL                                                                          |
| 68 | root            | localhost           | vt_commerce | Query            |    3 | Waiting for semi-sync ACK from slave                            | insert into customer values (2, 'ml')                                         |
| 69 | root            | localhost           | NULL        | Query            |    3 | Waiting for global read lock                                    | set global super_read_only=ON                                                 |
| 70 | root            | localhost           | NULL        | Query            |    0 | init                                                            | show processlist                                                              |
+----+-----------------+---------------------+-------------+------------------+------+-----------------------------------------------------------------+-------------------------------------------------------------------------------+
```

At least we then have a manual test that we can use to confirm the fix.

Thanks!

GuptaManan100:
@mattlord I don't know if there is a test that we can piggy back on, but I can add a new end to end test

shlomi-noach:
Trying to understand the problem better.

A semisync primary that has lost network to its replicas will not be able to apply DMLs and DDLs - right! But, why would it not be able to apply `set @@global.super_read_only`?

Is `DemotePrimary` part of PRS, ERS, or both?

- In PRS, we must block
- In ERS, we must not block (we can give the primary a few graceful seconds, but eventually timeout and continue the ERS flow)
- 
> we reorder the commands to first stop primary side semi-sync after stopping query service but before we issue the commands for setting [super] read only.

I'd like to rephrase it, please let me know if this rephrase is correct:

- Stop query service, ie primary will not be getting any new queries
- Disable semi-sync on primary (ie `set @@global.rpl_semi_sync_master_enabled=0`)
- Set `super_read_only`

Is that correct? Followup questions:

1. How long do we need to wait after stopping query service until all queries terminate?
2. Should we also `KILL` all running queries?
3. What happens as we disable semi-sync on primary? Won't a few queries suddenly execute? (and never make it to the replicas?)

I feel like the problem is not very clear, maybe I'm just not understanding this right.

GuptaManan100:
@shlomi-noach I'll answer all the questions in the order that they have been posed.

> A semisync primary that has lost network to its replicas will not be able to apply DMLs and DDLs - right! But, why would it not be able to apply `set @@global.super_read_only`?

It is the MySQL bugs https://bugs.mysql.com/bug.php?id=101056 and https://bugs.mysql.com/bug.php?id=104015 which @mattlord found out. 
The first bug makes it so that a transaction that didn't receive semi-sync acks stays blocked even if the replica's receive it and have applied it! The only way to unblock is to have another write which gets semi-sync acked, implicitly acking the blocked one too.
The second aggravates the problem by making commands like `set @@global.super_read_only` and `flush binary logs` to also block if a transaction is blocked.


Now consider this case in PRS, - 
1. Everything was going fine and we are stopping the query service on the primary to reparent.
2. The last transaction/write got blocked because of some transient issue, but the replicas are working fine and have the transaction.
3. The two MySQL bugs make it so that we block on `set @@global.super_read_only` and DemotePrimary never succeeds. Because we have already stopped query service, we don't even have an out in the sense that no new write is going to be sent!
This is the problem that we have set out to fix.

> Is `DemotePrimary` part of PRS, ERS, or both?

Yes, `DemotePrimary` is part of both PRS and ERS, but in slightly different ways. PRS will not proceed if `DemotePrimary` fails. It is essentially blocking, if the call fails.
On ERS, however, this happens in a separate go routine and we tolerate atmost 1 failure (sometimes more based on durability policies), but it is not blocking. This is exactly how you suggest it should be. 


> > we reorder the commands to first stop primary side semi-sync after stopping query service but before we issue the commands for setting [super] read only.
> 
> I'd like to rephrase it, please let me know if this rephrase is correct:
> 
> * Stop query service, ie primary will not be getting any new queries
> * Disable semi-sync on primary (ie `set @@global.rpl_semi_sync_master_enabled=0`)
> * Set `super_read_only`
> 
> Is that correct?

Yes, that is absolutely correct.
 
> Followup questions:
> 
> 1. How long do we need to wait after stopping query service until all queries terminate?
> 2. Should we also `KILL` all running queries?
The wait period is configurable by the flag `shutdown_grace_period`. As part of the `unserveCommon` function (which is called from the 1st step in DemotePrimary), we kill all the OLTP queries after waiting for `shutdown_grace_period` seconds. For the OLAP queries, we do not wait at all before killing them.
Between `sm.handleShutdownGracePeriod()` and `sm.te.Close()` we guarantee to have all the transactions either committed, rolled back or killed when we return from `unserveCommon`. 
Also, since we set our tabletserver state to StateNotServing, no new writes are accepted, this happens in `tsv.sm.StartRequest` part of the `tsv.sm.execRequest` function.
@harshit-gangal knows more about the code-level details of these operations if something is unclear. 


> 3. What happens as we disable semi-sync on primary? Won't a few queries suddenly execute? (and never make it to the replicas?)
By the time we are at the step to disable semi-sync on the primary, we should have no open transactions, pending writes or anything of that sort. At this point, if we have not already failed and exited, we want to set the primary to read only (or super read only based on configuration) and turn off primary side semi-sync.
If we do the read-only switch first, there is a chance it might block and this fails the entire DemotePrimary. Now, we have chosen to turn off primary side semi-sync first and it is safe because we know there are no writes currently in the system. After this, we set the tablet to read-only and proceed like we did before.

GuptaManan100:
@mattlord Have you been able to reproduce this issue locally. I have been trying for a while, but I can't seem to get it to fail on main even with the commands that you provided. The mysql just restarts if we kill it like we are doing in the second step 😟.

shlomi-noach:
> The second aggravates the problem by making commands like set @@global.super_read_only and flush binary logs to also block if a transaction is blocked.

Oh dear! that's bad. I wonder if there's a hack around this. Just weird stuff, like sometimes a `FLUSH BINARY LOGS` can reset a few things; so sometimes I prepend some queries with "block releaser" queries, just in case.

shlomi-noach:
> By the time we are at the step to disable semi-sync on the primary, we should have no open transactions, pending writes or anything of that sort.

So, I believe @dctrwatson points out that Online DDL may surprise with some writes. There's just too many things going on there, and it can't check for `read_only` or `super_read_only` at all times; it just assumes it can either do its work, or it's prevented from doing its work.

Online DDL operates as `SUPER`, and as I understand removing `super_read_only` can cause the demoted primary to then have errant GTIDs.

GuptaManan100:
> So, I believe @dctrwatson points out that Online DDL may surprise with some writes. There's just too many things going on there, and it can't check for `read_only` or `super_read_only` at all times; it just assumes it can either do its work, or it's prevented from doing its work.

@shlomi-noach As part of the `unserveCommon` function, we close the online ddl executor too. Is it still possible that some writes come from there?

> Online DDL operates as `SUPER`, and as I understand removing `super_read_only` can cause the demoted primary to then have errant GTIDs.
 
In this PR we aren't dealing with whether to `super_read_only` or not. It is just reordering the commands that set read only (or super read only) and the commands that turn off primary side semi-sync

GuptaManan100:
> Understood! I'm pointing out that if you first disable semi-sync and then disable `super_read_only`, then there may be a situation where Online DDL's Executor might sneak in a write.

Yes, that will be bad. The only reason that I have been thinking of this change being safe, because I have been believing that we will in no cases get a write at that point. If there is any way we can receive a write, then this fix is not safe and we should think of an alternate.
We can't bypass the durability that the user has set.

harshit-gangal:
> The wait period is configurable by the flag `shutdown_grace_period`. As part of the `unserveCommon` function (which is called from the 1st step in DemotePrimary), we kill all the OLTP queries after waiting for `shutdown_grace_period` seconds. For the OLAP queries, we do not wait at all before killing them. Between `sm.handleShutdownGracePeriod()` and `sm.te.Close()` we guarantee to have all the transactions either committed, rolled back or killed when we return from `unserveCommon`. Also, since we set our tabletserver state to StateNotServing, no new writes are accepted, this happens in `tsv.sm.StartRequest` part of the `tsv.sm.execRequest` function. @harshit-gangal knows more about the code-level details of these operations if something is unclear.
> 

Not related to thread going here. But, first we close the transaction engine and then we close any open non-transaction olap queries.

```
	sm.te.Close()
	log.Infof("Finished txEngine close. Killing all OLAP queries")
	sm.olapql.TerminateAll()
	log.Info("Finished Killing all OLAP queries. Started tracker close")

```

mattlord:
> @mattlord Have you been able to reproduce this issue locally. I have been trying for a while, but I can't seem to get it to fail on main even with the commands that you provided. The mysql just restarts if we kill it like we are doing in the second step 😟.

Yeah, we can repeat the hanging behavior in the first half of the steps before the PRS part where we restart those mysqlds. I also could not repeat the case where it does not self-heal once the replica mysqld's start up and re-connect to the primary. I wonder if PRS timing is involved there... will a replica tablet reconnect to the primary and attempt to replicate from it when the primary is in the middle of getting demoted?

GuptaManan100:
@mattlord Me and @deepthi were looking at this issue and we have had an epiphany. The MySQL bugs hang the [super] read only command only if a write is blocked on a semi-sync ack. Our entire argument for the safety of this change is that there are no writes in progress, so we can set semi-syck to false before we set read only. So intuitively, our fix either won't work or isn't safe.

Thinking about it further, if we don't have any writes in progress at the time we are trying to set read only, then how did it block at all? @deepthi then realised that we had had a bug which allowed a query to not be killed as part of the `unserveCommon` function. The PR that fixed it was https://github.com/vitessio/vitess/pull/10212. So we might not even require this change at all. The inherent issue would then be the query which wasn't killed, than the part of super read-only command blocking and the linked PR already fixes it

mattlord:
> Thinking about it further, if we don't have any writes in progress at the time we are trying to set read only, then how did it block at all? @deepthi then realised that we had had a bug which allowed a query to not be killed as part of the `unserveCommon` function. The PR that fixed it was #10212. So we might not even require this change at all. The inherent issue would then be the query which wasn't killed, than the part of super read-only command blocking and the linked PR already fixes it

We do issue the `KILL` for the query and session behind the write that we're waiting for the semi-sync ACK on. When doing that, however, it results in the killed flag setting set on the session and does NOT mean it goes away -- think of it like SIGTERM vs SIGKILL. While the thread is marked as killed, it continues to wait and in doing so it holds a mutex (effectively indicating that there's a currently in-progress write) that blocks the `super_read_only` change.

> This also explains why we can't repeat the problem. Even if we call PRS with a blocked write, it kills it and then super read-only command does not block irrespective of the semi-sync settings

I can repeat the killed session waiting for semi-sync asks, but the problem is that it's hard to repeat the case where when the replica's come back they don't unblock things by properly registering and then receiving and responding to the open ACK request -- that part seems to require a more complex timing+packet flow scenario as shown in the "how to repeat" section here: https://bugs.mysql.com/bug.php?id=104015

GuptaManan100:
Okay @mattlord that makes sense. Let me try repeating the failure with PRS. If we can't, then how do we proceed @deepthi ? We know there is a bug, and we know that this PR will fix it, but it is very hard to reproduce. 
Anyways, I would like to give it one more shot reproducing the bug.
Thankyou @mattlord for the information! ❤️

GuptaManan100:
@mattlord @deepthi Something anti-climactic happened 😿!
So I set out today to add a failing test so that we can verify that our fix works and instead I found another issue!
Let me describe the steps that the newly added test `TestDemotePrimarySemiSyncHang` takes and what the issue is.

The test starts with following the exact steps that @mattlord describes by creating a cluster and then removing all the replica tablets so that we only have the primary in the topology.

Then, we try and insert a value into a table in a separate go-routine.
After it is stuck, we try and set the primary to super read-only in a separate go-routine.
We expect both these statements to block because of the MySQL bug, and we can indeed see that is the case - 
```MySQL
mysql> show processlist
    -> ;
+----+-----------------+-----------+-------+------------------+------+-----------------------------------------------------------------+---------------------------------------------+
| Id | User            | Host      | db    | Command          | Time | State                                                           | Info                                        |
+----+-----------------+-----------+-------+------------------+------+-----------------------------------------------------------------+---------------------------------------------+
|  5 | event_scheduler | localhost | NULL  | Daemon           |  129 | Waiting on empty queue                                          | NULL                                        |
| 26 | vt_dba          | localhost | vt_ks | Sleep            |  125 |                                                                 | NULL                                        |
| 27 | vt_dba          | localhost | vt_ks | Sleep            |  125 |                                                                 | NULL                                        |
| 28 | vt_app          | localhost | vt_ks | Sleep            |  125 |                                                                 | NULL                                        |
| 29 | vt_dba          | localhost | vt_ks | Sleep            |  125 |                                                                 | NULL                                        |
| 41 | vt_filtered     | localhost | vt_ks | Binlog Dump GTID |  125 | Source has sent all binlog to replica; waiting for more updates | NULL                                        |
| 46 | vt_dba          | localhost | vt_ks | Sleep            |    5 |                                                                 | NULL                                        |
| 48 | root            | localhost | vt_ks | Query            |   37 | Waiting for semi-sync ACK from slave                            | insert into vt_insert_test values (2022,'') |
| 63 | root            | localhost | NULL  | Query            |   12 | Waiting for global read lock                                    | set global super_read_only=ON               |
| 64 | root            | localhost | NULL  | Query            |    0 | init                                                            | show processlist                            |
| 65 | vt_dba          | localhost | NULL  | Sleep            |    1 |                                                                 | NULL                                        |
| 66 | vt_dba          | localhost | NULL  | Sleep            |    1 |                                                                 | NULL                                        |
| 67 | vt_dba          | localhost | NULL  | Sleep            |    1 |                                                                 | NULL                                        |
| 68 | vt_dba          | localhost | NULL  | Sleep            |    1 |                                                                 | NULL                                        |
| 69 | vt_dba          | localhost | NULL  | Sleep            |    1 |                                                                 | NULL                                        |
| 70 | vt_dba          | localhost | NULL  | Sleep            |    1 |                                                                 | NULL                                        |
| 71 | vt_dba          | localhost | NULL  | Sleep            |    1 |                                                                 | NULL                                        |
| 72 | vt_dba          | localhost | NULL  | Sleep            |    1 |                                                                 | NULL                                        |
| 73 | vt_dba          | localhost | NULL  | Sleep            |    1 |                                                                 | NULL                                        |
| 74 | vt_dba          | localhost | NULL  | Sleep            |    1 |                                                                 | NULL                                        |
| 75 | vt_dba          | localhost | NULL  | Sleep            |    1 |                                                                 | NULL                                        |
| 76 | vt_dba          | localhost | NULL  | Sleep            |    1 |                                                                 | NULL                                        |
| 77 | vt_dba          | localhost | NULL  | Sleep            |    1 |                                                                 | NULL                                        |
+----+-----------------+-----------+-------+------------------+------+-----------------------------------------------------------------+---------------------------------------------+
23 rows in set (0.00 sec)
```
We can see the server waiting for a semi-sync ACK and also stuck on the super read-only command for the lock.

Next, I try to kill the insert query, as we do in `demotePrimary`. This command succeeds as expected.
```MySQL
mysql> kill 48;
```

At this point, I recheck the processlist on the primary tablet, and just as @mattlord pointed out, killing the insert query does not unblock anything. 
```MySQL
mysql> show processlist
    -> ;
+----+-----------------+-----------+-------+------------------+------+-----------------------------------------------------------------+---------------------------------------------+
| Id | User            | Host      | db    | Command          | Time | State                                                           | Info                                        |
+----+-----------------+-----------+-------+------------------+------+-----------------------------------------------------------------+---------------------------------------------+
|  5 | event_scheduler | localhost | NULL  | Daemon           |  143 | Waiting on empty queue                                          | NULL                                        |
| 26 | vt_dba          | localhost | vt_ks | Sleep            |  139 |                                                                 | NULL                                        |
| 27 | vt_dba          | localhost | vt_ks | Sleep            |  139 |                                                                 | NULL                                        |
| 28 | vt_app          | localhost | vt_ks | Sleep            |  139 |                                                                 | NULL                                        |
| 29 | vt_dba          | localhost | vt_ks | Sleep            |  139 |                                                                 | NULL                                        |
| 41 | vt_filtered     | localhost | vt_ks | Binlog Dump GTID |  139 | Source has sent all binlog to replica; waiting for more updates | NULL                                        |
| 46 | vt_dba          | localhost | vt_ks | Sleep            |    4 |                                                                 | NULL                                        |
| 48 | root            | localhost | vt_ks | Killed           |   51 | Waiting for semi-sync ACK from slave                            | insert into vt_insert_test values (2022,'') |
| 63 | root            | localhost | NULL  | Query            |   26 | Waiting for global read lock                                    | set global super_read_only=ON               |
| 64 | root            | localhost | NULL  | Query            |    0 | init                                                            | show processlist                            |
| 65 | vt_dba          | localhost | NULL  | Sleep            |   15 |                                                                 | NULL                                        |
| 66 | vt_dba          | localhost | NULL  | Sleep            |   15 |                                                                 | NULL                                        |
| 67 | vt_dba          | localhost | NULL  | Sleep            |   15 |                                                                 | NULL                                        |
| 68 | vt_dba          | localhost | NULL  | Sleep            |   15 |                                                                 | NULL                                        |
| 69 | vt_dba          | localhost | NULL  | Sleep            |   15 |                                                                 | NULL                                        |
| 70 | vt_dba          | localhost | NULL  | Sleep            |   15 |                                                                 | NULL                                        |
| 71 | vt_dba          | localhost | NULL  | Sleep            |   15 |                                                                 | NULL                                        |
| 72 | vt_dba          | localhost | NULL  | Sleep            |   15 |                                                                 | NULL                                        |
| 73 | vt_dba          | localhost | NULL  | Sleep            |   15 |                                                                 | NULL                                        |
| 74 | vt_dba          | localhost | NULL  | Sleep            |   15 |                                                                 | NULL                                        |
| 75 | vt_dba          | localhost | NULL  | Sleep            |   15 |                                                                 | NULL                                        |
| 76 | vt_dba          | localhost | NULL  | Sleep            |   15 |                                                                 | NULL                                        |
| 77 | vt_dba          | localhost | NULL  | Sleep            |   15 |                                                                 | NULL                                        |
| 78 | root            | localhost | vt_ks | Query            |    3 | Waiting for global read lock                                    | insert into vt_insert_test values (2022,'') |
+----+-----------------+-----------+-------+------------------+------+-----------------------------------------------------------------+---------------------------------------------+
24 rows in set (0.01 sec)
```

Now, I turn off semi-sync on the primary tablet and verify that everything is unstuck.
```MySQL
mysql> set global rpl_semi_sync_master_enabled = off;

mysql> show processlist;
+----+-----------------+-----------+-------+------------------+------+-----------------------------------------------------------------+------------------+
| Id | User            | Host      | db    | Command          | Time | State                                                           | Info             |
+----+-----------------+-----------+-------+------------------+------+-----------------------------------------------------------------+------------------+
|  5 | event_scheduler | localhost | NULL  | Daemon           |  184 | Waiting on empty queue                                          | NULL             |
| 26 | vt_dba          | localhost | vt_ks | Sleep            |  180 |                                                                 | NULL             |
| 27 | vt_dba          | localhost | vt_ks | Sleep            |  180 |                                                                 | NULL             |
| 28 | vt_app          | localhost | vt_ks | Sleep            |  180 |                                                                 | NULL             |
| 29 | vt_dba          | localhost | vt_ks | Sleep            |  180 |                                                                 | NULL             |
| 41 | vt_filtered     | localhost | vt_ks | Binlog Dump GTID |  180 | Source has sent all binlog to replica; waiting for more updates | NULL             |
| 46 | vt_dba          | localhost | vt_ks | Sleep            |    0 |                                                                 | NULL             |
| 63 | root            | localhost | NULL  | Sleep            |   67 |                                                                 | NULL             |
| 64 | root            | localhost | NULL  | Query            |    0 | init                                                            | show processlist |
| 65 | vt_dba          | localhost | NULL  | Sleep            |   56 |                                                                 | NULL             |
| 66 | vt_dba          | localhost | NULL  | Sleep            |   56 |                                                                 | NULL             |
| 67 | vt_dba          | localhost | NULL  | Sleep            |   56 |                                                                 | NULL             |
| 68 | vt_dba          | localhost | NULL  | Sleep            |   56 |                                                                 | NULL             |
| 69 | vt_dba          | localhost | NULL  | Sleep            |   56 |                                                                 | NULL             |
| 70 | vt_dba          | localhost | NULL  | Sleep            |   56 |                                                                 | NULL             |
| 71 | vt_dba          | localhost | NULL  | Sleep            |   56 |                                                                 | NULL             |
| 72 | vt_dba          | localhost | NULL  | Sleep            |   56 |                                                                 | NULL             |
| 73 | vt_dba          | localhost | NULL  | Sleep            |   56 |                                                                 | NULL             |
| 74 | vt_dba          | localhost | NULL  | Sleep            |   56 |                                                                 | NULL             |
| 75 | vt_dba          | localhost | NULL  | Sleep            |   56 |                                                                 | NULL             |
| 76 | vt_dba          | localhost | NULL  | Sleep            |   56 |                                                                 | NULL             |
| 77 | vt_dba          | localhost | NULL  | Sleep            |   56 |                                                                 | NULL             |
| 78 | root            | localhost | vt_ks | Sleep            |   44 |                                                                 | NULL             |
+----+-----------------+-----------+-------+------------------+------+-----------------------------------------------------------------+------------------+
23 rows in set (0.00 sec)
```

So far so good, here comes the problem. In the end, I verify that we don't see any data in the table that we were trying to insert the data into.
But to my surprise, the data shows up - 
```mysql
mysql> select * from vt_insert_test;
+------+------+
| id   | msg  |
+------+------+
| 2022 |      |
+------+------+
1 row in set (0.00 sec)
```

This seems wrong to me. What is the point of us killing a query, if it can still succeed after it has been killed? Do we do something different in `demotePrimary` that I am missing? 
It follows from ☝️ that our fix is not going to be safe, since we could have killed queries that get rolled forward if we turn off semi-sync before setting read-only. And since we turn off semi-sync, we can't even guarantee that these queries will be replicated. There is a chance that they get lost entirely.
So I guess, we can't turn off semi-sync before setting read-only 🤷‍♂️

GuptaManan100:
Discussion on fix for this is going on offline and will update when we have a way forward.

PR #10384, opened 2022-05-30
Streaming implementation of Projection engine primitive

harshit-gangal:
## Description

Implement streaming in projection engine primitive.

## Checklist


go/vt/vtgate/engine/projection.go
```diff
@@ -81,11 +83,43 @@ func (p *Projection) TryExecute(vcursor VCursor, bindVars map[string]*querypb.Bi
...
 func (p *Projection) TryStreamExecute(vcursor VCursor, bindVars map[string]*querypb.BindVariable, wantfields bool, callback func(*sqltypes.Result) error) error {
-	result, err := p.TryExecute(vcursor, bindVars, wantfields)
-	if err != nil {
-		return err
-	}
-	return callback(result)
+	env := evalengine.EnvWithBindVars(bindVars, vcursor.ConnCollation())
+	var once sync.Once
+	return p.Input.TryStreamExecute(vcursor, bindVars, wantfields, func(qr *sqltypes.Result) error {
+		result := &sqltypes.Result{}
```
systay, line 89: I think we should re-use the incoming `qr` and just replace the `Rows`. Or at least copy most of the variables. I'm thinking of `RowsAffected`, `StatusFlags` and `Info`

systay (approved):
lgtm

go/vt/vtgate/engine/projection.go
harshit-gangal: RowsAffected, StatusFlags, and Info are not affected by the select queries. I can see if I can reuse the incoming result, but with projection, we most likely change everything i.e. fields and row values.

PR #10386, opened 2022-05-30
Fix schema engine close and ticks race

GuptaManan100:
## Description

We had found a deadlock which resulted in schema Engine close being blocked. 
The steps that led to the deadlock are - 
 * We start the schema engine shutdown (engine.Close())
 * we take the schema engine mutex lock, and start the rest of the schema engine shutdown
 * but before we can get to stopping the timer for the schema reload
 * the timer starts up and executes engine.ReloadAt, which tries to take the schema engine mutex lock (waits).
 * At this point engine.Close() tries to stop the timer, but since the timer is executing, that waits for the execution to complete.  And we have a deadlock.

This PR adds a unit test for this failure and also fixes it. 

The proposed fix is to try and stop the timer in a separate go routine and wait for it to finish **after** releasing the mutex lock. What this accomplishes is that, if there is a `engine.ReloadAt` function blocked on the lock, it will start running and consequently unblock the Stop function as well. 
As far the outward guarantees of the Close functions go, they remain the same with the function only returning after the timer has stopped.
The part that has changed is that the timer is being closed in a go routine and we could have already released the lock by the time we actually stop the timer (This is what prevents the deadlock). Which means that there could be a call to Open on the schema engine which could acquire the lock and try and access the timer in parallel. But this is also safe, because the timer has its own locking and will block.

## Checklist

-   [x] "Backport me!" label has been added if this change should be backported No

### DIFF
--- a/go/vt/vttablet/tabletserver/schema/engine.go
+++ b/go/vt/vttablet/tabletserver/schema/engine.go
@@ -241,10 +241,11 @@ func (se *Engine) closeLocked() {
 	// but before closing the timer, in which case Stop function will wait for the
 	// configured function to complete running and that function (ReloadAt) will block
 	// on the lock we have already acquired
-	ticksWaitChan := make(chan bool, 0)
+	wg := sync.WaitGroup{}
+	wg.Add(1)
 	go func() {
 		se.ticks.Stop()
-		ticksWaitChan <- true
+		wg.Done()
 	}()
 	se.historian.Close()
 	se.conns.Close()
@@ -257,7 +258,7 @@ func (se *Engine) closeLocked() {
 	// Unlock the mutex. If there is a tick blocked on this lock,
 	// then it will run and we wait for the Stop function to finish its execution
 	se.mu.Unlock()
-	<-ticksWaitChan
+	wg.Wait()
 }
 
 // MakeNonPrimary clears the sequence caches to make sure that
--- a/go/vt/vttablet/tabletserver/schema/engine.go
+++ b/go/vt/vttablet/tabletserver/schema/engine.go
@@ -248,7 +248,7 @@ func (se *Engine) Close() {
 	se.isOpen = false
 
 	// Unlock the mutex. If there is a tick blocked on this lock,
-	// then it will run and we wait for the Stop function to finish it's execution
+	// then it will run and we wait for the Stop function to finish its execution
 	se.mu.Unlock()
 	<-ticksWaitChan
 	log.Info("Schema Engine: closed")
--- a/go/vt/vttablet/tabletserver/schema/engine.go
+++ b/go/vt/vttablet/tabletserver/schema/engine.go
@@ -229,6 +229,13 @@ func (se *Engine) Close() {
 		return
 	}
 
+	se.closeLocked()
+	log.Info("Schema Engine: closed")
+}
+
+// closeLocked closes the schema engine. It is meant to be called after locking the mutex of the schema engine.
+// It also unlocks the engine when it returns.
+func (se *Engine) closeLocked() {
 	// Close the Timer in a separate go routine because
 	// there might be a tick after we have acquired the lock above
 	// but before closing the timer, in which case Stop function will wait for the
@@ -251,7 +258,6 @@ func (se *Engine) Close() {
 	// then it will run and we wait for the Stop function to finish its execution
 	se.mu.Unlock()
 	<-ticksWaitChan
-	log.Info("Schema Engine: closed")
 }
 
 // MakeNonPrimary clears the sequence caches to make sure that
--- a/go/vt/vttablet/tabletserver/schema/engine.go
+++ b/go/vt/vttablet/tabletserver/schema/engine.go
@@ -224,12 +224,21 @@ func (se *Engine) IsOpen() bool {
 // It can be re-opened after Close.
 func (se *Engine) Close() {
 	se.mu.Lock()
-	defer se.mu.Unlock()
 	if !se.isOpen {
+		se.mu.Unlock()
 		return
 	}
 
-	se.ticks.Stop()
+	// Close the Timer in a separate go routine because
+	// there might be a tick after we have acquired the lock above
+	// but before closing the timer, in which case Stop function will wait for the
+	// configured function to complete running and that function (ReloadAt) will block
+	// on the lock we have already acquired
+	ticksWaitChan := make(chan bool, 0)
+	go func() {
+		se.ticks.Stop()
+		ticksWaitChan <- true
+	}()
 	se.historian.Close()
 	se.conns.Close()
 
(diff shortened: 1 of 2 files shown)


go/timer/timer.go
```diff
@@ -69,7 +69,7 @@ type Timer struct {
 // NewTimer creates a new Timer object
 func NewTimer(interval time.Duration) *Timer {
 	tm := &Timer{
-		msg: make(chan typeAction),
+		msg: make(chan typeAction, 1),
```
mattlord, line 72: Golang n00b question here... 

OK, so making it a buffered channel, even with a buffer size of 1, prevents the write to the channel from blocking in this case. I'm curious what rules we can apply here... 

A buffer size of one allows us to buffer 1 unread message in the channel: https://go.dev/play/p/CGi0hp-kbEs

And a buffer size of two allows us to buffer two: https://go.dev/play/p/MC4lKPxDrCk

It's not clear to me why that is, or is not, the right thing to do in this case. The Timer is generally intended to support exactly 1-2 multiple goroutines write to it concurrently w/o requiring the reads/writes be synchronized (using the mutex Timer has)?

GuptaManan100: @mattlord Your concern is absolutely legitimate! I was thinking the same thing when I had thought of this solution. I had thought of 2 things - 1. Is it okay to increase the buffer size of the channel from 0 which will unblock some writes to it? and 2. How much should we increase it to? 1, 2 or 1000?

Let me answer both these questions and my thoughts as to how I arrived there.
To effectively answer these question lets look at the usages of `msg`, the channel in question.
There are 3 writers to the channel - `Stop`, `Trigger` and `SetInterval`. There is only 1 reader which is the `run` function. Let's look at each in detail -
- The purpose of the `Trigger` function to trigger a new run. Earlier it would return only after the run had started (but not finished). If the writes are buffered, then it will return after scheduling a run, but not waiting for it to start. From the outside the two situations are indistinguishable since the run function can take while to finish execution so in both cases, running a `Trigger` only guarantees you that the function will run eventually.
- The purpose of the `SetInterval` function is to change the timer intervals between each execution. Earlier, this function would return after the interval had been changed. Now the function returns after the interval change has been scheduled and after the current run (if any) finishes, that will take effect. The next run would still be with the new interval. Again there is no distinguishable change from the caller's perspective.
- The final function is `Stop`, which is used to stop the timer's execution. As per the comment of that function `It guarantees that the timer will not execute any more calls to keephouse once it has returned.` But there is no promise for the function which is executing now! If there is keep house function running, then in the run function we are blocked on the `msg` channel read, and we would immediately pick up the message stop pushed and not run any more keephouse functions. If we are executing the keephouse function, then it gets a little more interesting. According to the guarantee, it is okay to finish execution of this keephouse function, but we shouldn't start another one. So, if we buffer the message that Stop function is sending, we have to be sure that it gets processed after the current function's execution and but before a new execution starts. This is what constraints us to the buffer size 1! If we had a buffer size 2, then there could already be a message in the msg channel. Stop adds one more and returns, but the first messaged is processed first (which could be a trigger message) and this starts a new keephouse run even though Stop has returned!

So, keeping in mind all the observable changes and the promises of the timer module, we now have the answer to our two initial questions - 
1. It is safe to increase the buffer size.
2. But the buffer size can only be atmost 1, because we need to guarantee that the Stop functions message is at the top of the msg channel when it returns! (There are other ways to guarantee this, by increasing the buffer size more and then waiting in the stop function until it is empty, but we don't need to go there since that is not required)

Because of this analysis, I think it is safe to make this simple, albeit a large change as you point out.
I don't expect there to be any issues if the usages of the timer are written considering it's guarantees, which are still held. The only function that could potentially create issues is the `Stop` function which (now) returns after scheduling a stop but a keephouse function might still be in progress. Earlier we would have waited until everything had stopped.
However, in this case, I would argue that the usage of the timer `Stop` is incorrect and not this change, because we still uphold the guarantees and we would fix that place (if any).

I have run this by @harshit-gangal too. Does my analysis make sense?
cc - @deepthi if she has any insights if this would break something

GuptaManan100: I'll do that 👍

deepthi (dismissed):
Looks good. May be merged once a comment is added to the LOC that changed.

GuptaManan100:
@mattlord My analysis goes out the window, because I just saw the comment https://github.com/vitessio/vitess/blob/f20905aec55b58a77ab7f85fd4416bc9680ae935/go/timer/timer.go#L48-L49
So we can't increase the buffer size of the channel.
The comment on top does not match the comment on the Stop function which only says https://github.com/vitessio/vitess/blob/f20905aec55b58a77ab7f85fd4416bc9680ae935/go/timer/timer.go#L145-L146

We'll have to find an alternate fix.

GuptaManan100:
@mattlord @deepthi I have found an alternate fix to the problem. 
The new proposed fix is to try and stop the timer in a separate go routine and wait for it to finish **after** releasing the mutex lock. What this accomplishes is that, if there is a `engine.ReloadAt` function blocked on the lock, it will start running and consequently unblock the Stop function as well. 
As far the outward guarantees of the Close functions go, they remain the same with the function only returning after the timer has stopped.
The part that has changed is that the timer is being closed in a go routine and we could have already released the lock by the time we actually stop the timer (This is what prevents the deadlock). Which means that there could be a call to Open on the schema engine which could acquire the lock and try and access the timer in parallel. But this is also safe, because the timer has its own locking and will block the newer call.

Do these changes make sense? @mattlord is there any scenario that you can think which could cause this fix to fail? I really want to be sure, we don't create another issue from this fix.

deepthi (commented):
LGTM
Let's wait for @mattlord also to review.

mattlord:
That seems OK, although the behavior still feels wrong.

Do we ever want more than one Schema Engine per tablet? I see that we create a new one in `externalConnector`'s and `replicaConnector`'s in VReplication. Should SchemaEngine be a singleton?

I also noticed that the schema is reloaded at various places, e.g. here: https://github.com/vitessio/vitess/blob/main/go/vt/vttablet/tabletmanager/rpc_query.go#L31 I wonder if we're shutting down the query service before the schema engine?

Anyway... I don't want to block this as we are in the shutdown process and I'm assuming the new test fails on main? If so, I'm fine approving as-is.

Thank you!

GuptaManan100:
@mattlord I do not quite understand your concern. 

How does it make a difference if the Schema engine is per tablet or not? Even if we have multiple of them, their timers and locks would be all separate, so the won't interfere with each other.

The code changes in this PR are trying to address the problem of deadlock when there is a schema-engine `Close` and its timer tick happens.

> I also noticed that the schema is reloaded at various places

Yes, exactly. I had an alternate solution in mind to change ReloadAt to only try and acquire the lock, fail if it can't instead of waiting. But I didn't go ahead with that solution because there are places using it that rely on its behaviour to guarantee schema reload when it returns.

Do you have any other concerns with the proposed fix?

mattlord:
Thanks, @GuptaManan100! I don't want to block this more as I'm unaware of all the semantic implications involved around this code. I'm good with it if you and the Query Serving team are.

systay (approved):
DRY is good. 👍

PR #10388, opened 2022-05-30
feat: do not use lookup vindexes for IS NULL predicates

systay:
## Description
When a query is using a `IS NULL` predicate against a lookup vindex column, we should have a very hight cost for these, since they will always need to do a scatter query. The lookup table is keyed by the column and so does not support `null` values.

## Related Issue(s)
#10346

## Checklist

### DIFF
--- a/go/vt/vtgate/planbuilder/testdata/filter_cases.txt
+++ b/go/vt/vtgate/planbuilder/testdata/filter_cases.txt
@@ -2138,11 +2138,7 @@ Gen4 plan same as above
     },
     "FieldQuery": "select id from music where 1 != 1",
     "Query": "select id from music where id is null",
-    "Table": "music",
-    "Values": [
-      "NULL"
-    ],
-    "Vindex": "music_user_map"
+    "Table": "music"
   }
 }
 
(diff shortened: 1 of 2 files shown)


go/vt/vtgate/planbuilder/testdata/filter_cases.txt
```diff
@@ -2126,7 +2126,25 @@ Gen4 plan same as above
...
+      "Name": "user",
+      "Sharded": true
+    },
+    "FieldQuery": "select id from music where 1 != 1",
+    "Query": "select id from music where id is null",
+    "Table": "music",
+    "Values": [
+      "NULL"
+    ],
+    "Vindex": "music_user_map"
```
GuptaManan100, line 2145: Why are we using the vindex at all, if it is a scatter?

systay: we could clean it up and not use these at all. at the moment, we are just changing the opcode to scatter so they are valued much less and only used when no other alternative exists. You're right though, this is confusing. I'll clean it up.

/cc @arthurschreiber

GuptaManan100 (approved):
Other than that 👍

PR #10390, opened 2022-05-31
OnlineDDL/vitess fix: convert data type to JSON 

shlomi-noach:
## Description

This PR fixes `vitess` migrations where a column is modified from a non-`JSON` type into `JSON`. When converting to `JSON`, `vreplication` adds a `convert(%v using utf8mb4)` wrapper around the column. However, because in our scenario the original column is non-`JSON`, the wrapper is not added.

We now analyze the scenario where source column is non-`JSON` and target column is `JSON`, and ensure to add `convert()` around the column name.

**Also**, this PR now identifies a series of errors as non-recoverable, such that a `vreplication` stream will fail the moment it encounters them. The added errors are all `JSON` conversion -related (e.g. empty document, invalid JSON format, data too long, etc.)

`vrepl_suite` tests are added to cover multiple scenarios.

## Related Issue(s)

#6926 

## Checklist


rohit-nayak-ps (approved):
lgtm

The gh-ost test failed once and succeeded the second time, but as we have found the action still failed. Restarted it: we should keep an eye out since we have found recently that many of the flaky tests did have underlying bugs.

PR #10221, opened 2022-05-05
Adds DurabilityPolicy to the KeyspaceInfo in the topo and the associated RPCs

GuptaManan100:
## Description

Since we decided to move away from the `durability_policy` flag approach which we had previously taken (refer https://github.com/vitessio/vitess/pull/10201), we are now moving towards storing the durability_policy in the topo server.
This PR adds the durability_policy to the keyspaceInfo stored in the keyspace record and provides RPC function to update these. The existing RPC functions for creating and getting a keyspace recorded have also been augmented and tested as part of the PR.
Essentially, this PR provides the complete tool-kit to read and update the durability_policies in the keyspace record.  

The next steps after these changes, is to read the durability_policy in all the operations that need it and set it then. As part of this change, we will also move away from the architecture of a single durability_policy stored as a global variable towards a parameter based approach.

## Checklist

-   [x] "Backport me!" label has been added if this change should be backported
-   [x] Tests were added or are not required
-   [x] Documentation was added or is not required

PR #10395, opened 2022-05-31
Fail if durability policy provided for snapshot keyspaces

GuptaManan100:
## Description
In the PR https://github.com/vitessio/vitess/pull/10221, we missed the check that durability policy should not be provided for snapshot keyspaces. If it is, then we should fail. This small PR fixes that miss.

## Checklist

### DIFF
--- a/go/vt/vtctld/api_test.go
+++ b/go/vt/vtctld/api_test.go
@@ -187,7 +187,7 @@ func TestAPI(t *testing.T) {
 	}{
 		// Create snapshot keyspace with durability policy specified
 		{"POST", "vtctl/", `["CreateKeyspace", "-keyspace_type=SNAPSHOT", "-base_keyspace=ks1", "-snapshot_time=2006-01-02T15:04:05+00:00", "-durability-policy=semi_sync", "ks3"]`, `{
-  "Error": "durability-policy must not be specified while creating a snapshot keyspace"`, http.StatusOK},
+  "Error": "durability-policy cannot be specified while creating a snapshot keyspace"`, http.StatusOK},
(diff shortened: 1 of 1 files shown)


go/cmd/vtctldclient/command/keyspaces.go
```diff
@@ -163,6 +163,10 @@ func commandCreateKeyspace(cmd *cobra.Command, args []string) error {
 
 	var snapshotTime *vttime.Time
 	if topodatapb.KeyspaceType(createKeyspaceOptions.KeyspaceType) == topodatapb.KeyspaceType_SNAPSHOT {
+		if createKeyspaceOptions.DurabilityPolicy != "none" {
```
deepthi, line 166: I wonder whether we should just log a warning and ignore the parameter instead of erroring out.

GuptaManan100: I think that ignoring the durability policy could be a little non-intuitive from the user perspective. I think it is better to explicitly ask them to rerun the command without specifying the durability at all.

ajm188: I'm inclined toward the "error out" approach; the error message is clear and the user can just re-run the command with the correct incantation (vs maybe getting an unexpected "surprise" behavior).

Only thing I would change is `--durability-policy must not be specified ...` => `--durability-policy cannot be specified`

ajm188 (approved):

go/cmd/vtctldclient/command/keyspaces.go
deepthi: sgtm

PR #10397, opened 2022-05-31
[vtadmin] Add infrastructure for generating authz tests for vtadmin

ajm188:
## Description

The lack of verifying authz checks are where they should be is one of the
most glaring issues in vtadmin (in my opinion; it's also my "fault" things
are this way). At the same time, writing all the code by hand to verify
every single endpoint would be a giant pain (which is the main reason
things are this way). So, let's codegen all the bits we don't care about!
The bonus here is that the config.json now can serve as authoritative on
what permissions are required for what endpoints.

The goal here is to have the config primarily specify the rules needed for
each endpoint, with as minimal "overhead" (currently specifying test cases
and mock data) as possible.

I want to separate the introduction of this setup from its complete
adoption, so I will submit a follow-up change that adds the rest of the
endpoint tests.

### Outstanding questions

- What's the right spot for me to inject a make target to run `go generate ./go/vt/vtadmin/...` and `go fmt ./go/vt/vtadmin` ?

## Checklist


deepthi (commented):
This is a great idea!
> What's the right spot for me to inject a make target to run go generate ./go/vt/vtadmin/... and go fmt ./go/vt/vtadmin ?

The pattern we have followed is to create a separate (new) make target for code generation AND add a CI check to make sure generated files are up to date.

ajm188:
> The pattern we have followed is to create a separate (new) make target for code generation AND add a CI check to make sure generated files are up to date.

Done!!

@frouioui I will fully cop to admitting I just copied and modified the setup for `make proto`, can you give the CI bits a once-over just to double-check my work? It seems to work but I just want to be sure (especially) that I'm using your GitHub action correctly:

```
./tools/check_make_vtadmin_authz_testgen.sh
go/vt/vtadmin/api_authz_test.go
ERROR: Regenerated vtadmin_test files do not match the current version.
List of files containing differences:
0a1
>  M go/vt/vtadmin/api_authz_test.go
```

frouioui (approved):
It looks good to me, nothing to report on the CI part! :)

frouioui:
> This is a great idea!

Plus one on that 😃

frouioui:
> ```
> ./tools/check_make_vtadmin_authz_testgen.sh
> go/vt/vtadmin/api_authz_test.go
> ERROR: Regenerated vtadmin_test files do not match the current version.
> List of files containing differences:
> 0a1
> >  M go/vt/vtadmin/api_authz_test.go
> ```

@ajm188, the above is fine in the case where we see a difference in `api_authz_test.go `

ajm188:
> > ```
> > ./tools/check_make_vtadmin_authz_testgen.sh
> > go/vt/vtadmin/api_authz_test.go
> > ERROR: Regenerated vtadmin_test files do not match the current version.
> > List of files containing differences:
> > 0a1
> > >  M go/vt/vtadmin/api_authz_test.go
> > ```
> 
> 
> @ajm188, the above is fine in the case where we see a difference in `api_authz_test.go`

oh, yeah 😅 i was just demonstrating (without, ya know, actually making that clear _at all_) that if i change the config but don't re-gen, the check fails (as desired)

PR #10404, opened 2022-06-01
[vtadmin] Rename reparent-related RPCs, endpoints, methods, RBAC actions

ajm188:
## Description

Rename reparent-related RPCs, endpoints, methods, RBAC actions

- `EmergencyReparentShard` => `EmergencyFailoverShard`
- `PlannedReparentShard` => `PlannedFailoverShard`
- `ReparentTablet` => `RefreshTabletReplicationSource`
- `TabletExternallyReparented` => `TabletExternallyPromoted`

## Checklist


go/vt/vtadmin/cluster/cluster.go
```diff
@@ -1968,6 +1968,35 @@ func (c *Cluster) RefreshState(ctx context.Context, tablet *vtadminpb.Tablet) er
 	return err
 }
 
+// RefreshTabletReplicationSource performs a `CHANGE REPLICATION SOURCE TO` on
+// a tablet to replicate from the current primary in the shard. This only works
+// if the current replica position matches the last known
```
deepthi, line 1973: How about we delete the sentence we don't understand? I mean this part.
```
This only works
 // if the current replica position matches the last known
 // `CHANGE REPLICATION SOURCE TO` or failover action.
 ```

deepthi (approved):
Looks fine to me, except for one small comment.

go/vt/vtadmin/cluster/cluster.go
deepthi: The real deal is that this function will fail if the replica you are trying to change the source on has transactions that the new "source" does not have. I don't think it is necessary to get into that detail here at all.

PR #10408, opened 2022-06-01
topo: cache CellInfo along with conn

rsajwani:
Co-authored-by: Rameez Sajwani <rameezwazirali@hotmail.com>

## Description

Issue: vtctlclient ListAllTablets does not show those tablets in cells that were deleted and re-added with same name.
The root cause is that we cache connections to cells by name, and the name is being re-used.

We are fixing this by caching the CellInfo along with conn so that we can verify that it is the exact "same" cell before using the cached connection.

## Repro

We create a cell "abc" with some rootPath like path1/abc. Then we delete it, and create it again with the same name but a different path like path2/abc.
The vtctld cache will continue to contains a connection with path path1/abc which no longer exists.

## Checklist

### DIFF
--- a/.github/workflows/cluster_endtoend_topo_connection_cache.yml
+++ b/.github/workflows/cluster_endtoend_topo_connection_cache.yml
@@ -1,0 +1,74 @@
+# DO NOT MODIFY: THIS FILE IS GENERATED USING "make generate_ci_workflows"
+
+name: Cluster (topo_connection_cache)
+on: [push]
+concurrency:
+  group: format('{0}-{1}', ${{ github.ref }}, 'Cluster (topo_connection_cache)')
+  cancel-in-progress: true
+
+jobs:
+  build:
+    name: Run endtoend tests on Cluster (topo_connection_cache)
+    runs-on: ubuntu-18.04
+
+    steps:
+    - name: Configure git private repo access
+      env:
+        GITHUB_TOKEN: ${{ secrets.PLANETSCALE_ACTIONS_BOT_TOKEN }}
+      run: |
+        git config --global --add url."https://${GITHUB_TOKEN}@github.com/".insteadOf "https://github.com/"
+
+    - name: Set up Go
+      uses: actions/setup-go@v2
+      with:
+        go-version: 1.17.9
+
+    - name: Set up python
+      uses: actions/setup-python@v2
+
+    - name: Tune the OS
+      run: |
+        echo '1024 65535' | sudo tee -a /proc/sys/net/ipv4/ip_local_port_range
+        # Increase the asynchronous non-blocking I/O. More information at https://dev.mysql.com/doc/refman/5.7/en/innodb-parameters.html#sysvar_innodb_use_native_aio
+        echo "fs.aio-max-nr = 1048576" | sudo tee -a /etc/sysctl.conf
+        sudo sysctl -p /etc/sysctl.conf
+
+    - name: Check out code
+      uses: actions/checkout@v2
+
+    - name: Get dependencies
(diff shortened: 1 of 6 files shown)


test/config.json
```diff
@@ -1228,6 +1228,24 @@
...
+		},
+		"vreplication_source_time_zone": {
+			"File": "unused.go",
+			"Args": ["vitess.io/vitess/go/test/endtoend/vreplication", "-run", "TestMoveTablesTZ"],
+			"Command": [],
+			"Manual": false,
+			"Shard": "vreplication_migrate",
+			"RetryMax": 1,
+			"Tags": []
+		},
```
deepthi, line 1240: This change seems unrelated, maybe it should not be in this PR?

rsajwani: yeah looks like it added as part of auto conflict resolution .. let me remove it.

.github/workflows/cluster_endtoend_topo_connection_cache.yml
```diff
@@ -0,0 +1,74 @@
...
+    - name: Configure git private repo access
+      env:
+        GITHUB_TOKEN: ${{ secrets.PLANETSCALE_ACTIONS_BOT_TOKEN }}
+      run: |
+        git config --global --add url."https://${GITHUB_TOKEN}@github.com/".insteadOf "https://github.com/"
+
+    - name: Set up Go
+      uses: actions/setup-go@v2
+      with:
+        go-version: 1.17.9
```
frouioui, line 24: This should be at least `go.1.18.1`

.github/workflows/cluster_endtoend_topo_connection_cache.yml
```diff
@@ -0,0 +1,74 @@
+# DO NOT MODIFY: THIS FILE IS GENERATED USING "make generate_ci_workflows"
+
+name: Cluster (topo_connection_cache)
+on: [push]
```
frouioui, line 4: You should add `pull_request` to the list of targets as well. Otherwise, the build will never start on a "regular" commit that lives in a Pull Request. The target `push` only works when a commit is pushed on any branch of `vitessio/vitess` (no its forks).

go/vt/topo/server.go
```diff
@@ -56,7 +58,7 @@ import (
 
 const (
 	// GlobalCell is the name of the global cell.  It is special
-	// as it contains the global topology, and references the other cells.
+	// as it contains the global topology, and references the other cellConns.
```
deepthi, line 61: This change should be reverted.

go/vt/topo/server.go
```diff
@@ -98,7 +100,7 @@ type Factory interface {
 	// HasGlobalReadOnlyCell returns true if the global cell
 	// has read-only replicas of the topology data. The global topology
 	// is usually more expensive to read from / write to, as it is
-	// replicated over many cells. Some topology services provide
+	// replicated over many cellConns. Some topology services provide
```
deepthi, line 103: This should also be reverted.

deepthi, line 103: This should also be reverted.

go/vt/topo/server.go
```diff
@@ -234,46 +241,46 @@ func Open() *Server {
 }
 
 // ConnForCell returns a Conn object for the given cell.
-// It caches Conn objects from previously requested cells.
+// It caches Conn objects from previously requested cellConns.
```
deepthi, line 244: This should also be reverted.

deepthi, line 244: This should also be reverted.

deepthi (approved):

PR #10411, opened 2022-06-02
Online DDL fix: instant is possible for VIRTUAL, not STORED

shlomi-noach:
## Description

Followup to https://github.com/vitessio/vitess/pull/10402

In https://github.com/vitessio/vitess/pull/10402 we add (undocumented) support for Instant DDL. We identify ADD/DROP columns scenarios. By mistake, the logic allowed any `GENERATED` column ADD/DROP as instant DDL, where it should only allow `VIRTUAL` columns, which are a subset of `GENERATED` columns. See https://dev.mysql.com/doc/refman/8.0/en/innodb-online-ddl-operations.html

## Related Issue(s)

#6926 

## Checklist


deepthi (approved):

dbussink (approved):

shlomi-noach:
The `Cluster (onlineddl_vrepl_stress_suite) mysql80` CI is flaky and its failure is accounted for; it is related to CI/mysql8 issues, which I will investigate separately.

PR #10413, opened 2022-06-02
schemadiff: column name check in CHECK constraint is case insensitive

shlomi-noach:
## Description

Column names are case insensitive. So should be our checks in `postApplyNormalize` and `Validate()` as we look into columns referenced by check constraints.

## Related Issue(s)

#10203 

## Checklist

### DIFF
--- a/go/vt/schemadiff/table.go
+++ b/go/vt/schemadiff/table.go
@@ -29,7 +29,6 @@ import (
 	"vitess.io/vitess/go/vt/sqlparser"
 )
 
-//
 type AlterTableEntityDiff struct {
 	from       *CreateTableEntity
 	to         *CreateTableEntity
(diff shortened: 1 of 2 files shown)


go/vt/schemadiff/table.go
```diff
@@ -1858,7 +1858,7 @@ func (c *CreateTableEntity) validate() error {
 		err := sqlparser.Walk(func(node sqlparser.SQLNode) (kontinue bool, err error) {
 			switch node := node.(type) {
 			case *sqlparser.ColName:
-				referencedColumns = append(referencedColumns, node.Name.String())
+				referencedColumns = append(referencedColumns, node.Name.Lowered())
```
dbussink, line 1861: The list here is later used in error messages, so this means that in the error message we lose the proper casing. I think we need to move this to the map lookup (and take that approach in general then).

dbussink (approved):
Fixed my comment and added more additional test cases as well and further improvements. See the commit message for more details.

PR #9632, opened 2022-02-04
Improve performance of information schema query.

arthurschreiber:
## Description

Querying the `information_schema` tables can be quite expensive in MySQL 5.7.

On the test database I'm using (169 tables), each execution of the `TablesWithSize57` query takes multiple seconds, making this query unbearably slow.

But the runtime of this query can be improved with a few tricks.

For reference, here's the original query:

```sql
	SELECT t.table_name, t.table_type, unix_timestamp(t.create_time), t.table_comment, sum(i.file_size), sum(i.allocated_size) 
	FROM information_schema.tables t, information_schema.innodb_sys_tablespaces i 
	WHERE t.table_schema = database() and 
	(i.name = concat(t.table_schema,'/',t.table_name) or i.name like concat(t.table_schema,'/',t.table_name, '#p#%')) 
	group by t.table_name, t.table_type, unix_timestamp(t.create_time), t.table_comment, i.file_size
UNION ALL
	SELECT table_name, table_type, unix_timestamp(create_time), table_comment, SUM( data_length + index_length), SUM( data_length + index_length)
	FROM information_schema.tables t
	WHERE table_schema = database() AND 
	NOT EXISTS(SELECT * FROM information_schema.innodb_sys_tablespaces i WHERE i.name = concat(t.table_schema,'/',t.table_name) or i.name like concat(t.table_schema,'/',t.table_name, '#p#%')) 
	group by table_name, table_type, unix_timestamp(create_time), table_comment
```

```
169 rows in set (5.36 sec)
```

The first improvement is to reduce the number of accesses to the `information_schema.innodb_sys_tablespaces` table by replacing the `NOT EXISTS` part of the query with a `LEFT OUTER JOIN`:

```sql
	SELECT t.table_name, t.table_type, unix_timestamp(t.create_time), t.table_comment, sum(i.file_size), sum(i.allocated_size) 
	FROM information_schema.tables t, information_schema.innodb_sys_tablespaces i 
	WHERE t.table_schema = database() and 
	(i.name = concat(t.table_schema,'/',t.table_name) or i.name like concat(t.table_schema,'/',t.table_name, '#p#%')) 
	group by t.table_name, t.table_type, unix_timestamp(t.create_time), t.table_comment, i.file_size
UNION ALL
	SELECT table_name, table_type, unix_timestamp(create_time), table_comment, SUM( data_length + index_length), SUM( data_length + index_length)
	FROM information_schema.tables t
[...]

PR #10416, opened 2022-06-02
Improve performance of the `DetectSchemaChange` query

arthurschreiber:
## Description

Accessing the `information_schema.columns` table is expensive in MySQL 5.7, especially when it's accessed in a nested loop (e.g. via a `JOIN`). The query to detect schema changes was accessing `information_schema.columns` multiple times, which could easily add up to execution times of 0.5 seconds per query on my test system.

This pull request proposes a replacement query that is more careful about accessing `information_schema.columns` (and `_vt.schemacopy`) only once to determine all tables that have been modified and need to be re-cached.

## Related Issue(s)

See https://github.com/vitessio/vitess/pull/9632 for a similar optimization made previously.

## Checklist

### DIFF
--- a/go/mysql/schema.go
+++ b/go/mysql/schema.go
@@ -56,48 +56,31 @@ CREATE TABLE if not exists _vt.schemacopy (
 	column_key varchar(3) NOT NULL,
 	PRIMARY KEY (table_schema, table_name, ordinal_position))`
 
-	detectNewColumns = `
-select ISC.table_name
-from information_schema.columns as ISC
-	 left join _vt.schemacopy as c on 
-		ISC.table_name = c.table_name and 
-		ISC.table_schema=c.table_schema and 
-		ISC.ordinal_position = c.ordinal_position
-where ISC.table_schema = database() AND c.table_schema is null`
-
-	detectChangeColumns = `
-select ISC.table_name
-from information_schema.columns as ISC
-	  join _vt.schemacopy as c on 
-		ISC.table_name = c.table_name and 
-		ISC.table_schema=c.table_schema and 
-		ISC.ordinal_position = c.ordinal_position
-where ISC.table_schema = database() 
-	AND (not(c.column_name <=> ISC.column_name) 
-	OR not(ISC.character_set_name <=> c.character_set_name) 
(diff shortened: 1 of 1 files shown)


go/mysql/schema.go
```diff
@@ -56,64 +56,47 @@ CREATE TABLE if not exists _vt.schemacopy (
...
+	WHERE table_schema = database()
+
+	UNION ALL
+
+	SELECT table_name, column_name, character_set_name, collation_name, data_type, column_key
+	FROM _vt.schemacopy c
+	WHERE table_schema = database()
+) _inner
+GROUP BY table_name, column_name, character_set_name, collation_name, data_type, column_key
+HAVING COUNT(*) = 1
```
arthurschreiber, line 74: This new query fetches all the data from the schema copy and the `information_schema.columns` table, and uses a simple counting "trick" to determine whether a table was modified or not.

When rows that exist both in `information_schema.columns` and `_vt.schemacopy` will end up twice in the result of the `UNION ALL`, while rows that have been added / removed / modified will exist only once. Through the `HAVING COUNT(*) = 1`, we can very quickly determine the tables that have been modified.

"Look ma, no joins!"

arthurschreiber:
Parts of the old query also took `ordinal_position` into account. It would be easy to add it to the new query as well, but I wasn't quite sure whether the column position is relevant for the schema copy. I _don't_ think it is, but if someone who has more knowledge of this code could confirm this, that'd be great! 🙇‍♂️

harshit-gangal:
> Parts of the old query also took `ordinal_position` into account. It would be easy to add it to the new query as well, but I wasn't quite sure whether the column position is relevant for the schema copy. I _don't_ think it is, but if someone who has more knowledge of this code could confirm this, that'd be great! bowing_man

ordinal_position is used to know if the column position is changed, the above modified query will miss those changes.

e.g.
t_a (a, b, c) -> t_a (a, c, b)

this should be considered as schema change otherwise expansion of select * queries will result in the wrong output position.

arthurschreiber:
Thanks @harshit-gangal for the explanation! 🙇‍♂️ I updated the query to take `ordinal_position` into account as well.

harshit-gangal (approved):

PR #10417, opened 2022-06-02
making log sanitization more precise

rsajwani:

## Description
Removing code = \S+ can be removed completely as a cut point. There is useful content after it that would be nice to preserve. 
We can replace important PII stuff with '<val>' duplicate entry

## Checklist


deepthi (approved):
Looks fine to me. This was in fact the substance of my comment on the previous PR - that we might be dropping text too aggressively.
I didn't understand this statement though
```
We can replace important PII stuff with '' duplicate entry
```
Can you update the description to make that clearer?

deepthi:
vtctld_web_test seems like a flaky failure. Completely unrelated to this small code change.

PR #10420, opened 2022-06-03
replace planner-version with planner_version in vtexplain

systay:
## Description
vtexplain mistakenly used `planner-version` instead of `planner_version`, note the underscore vs the hyphen.

We want to keep the same flag that the vtgate is using, so this PR changes it to use the new flag.

## Checklist


GuptaManan100 (approved):

frouioui (approved):

frouioui:
On the doc, [this](https://vitess.io/docs/13.0/reference/programs/vtexplain/) page has to change as well

systay:
This PR pairs very well together with https://github.com/vitessio/website/pull/1039

ajm188:
++ please use `-` when adding new flags; it'll make my life easier in v15 ❤️

PR #10392, opened 2022-05-31
Add durability policy changes to the summary

GuptaManan100:
## Description
This PR adds the durability policy changes to the summary of release-14 for the following PRs-
- https://github.com/vitessio/vitess/pull/10201
- https://github.com/vitessio/vitess/pull/10221
- https://github.com/vitessio/vitess/pull/10375
- https://github.com/vitessio/vitess/pull/10423

## Checklist

-   [x] "Backport me!" label has been added if this change should be backported
-   [x] Tests were added or are not required
-   [x] Documentation was added or is not required

PR #10423, opened 2022-06-03
Use Durability Policy in the topo server in VTOrc and deprecate Durability config

GuptaManan100:
## Description

This PR continues the work of using the durability policy stored in the topo server started in https://github.com/vitessio/vitess/pull/10392. 
We make changes to VTOrc in this PR to deprecate and remove the `Durability` configuration. Instead VTOrc will now also read the durability policy from the topo server. Since, VTOrc needs this information for running the error analysis, it will also need to store the same in its data store to avoid a call to the topo server on each error analysis tick.
This PR accomplishes all of these changes and makes it so that VTOrc too can work with multiple keyspaces having different durability policies.
This PR also adds the check to VTOrc to ignore snapshot keyspaces when running error analysis

VTOrc will ignore keyspaces which don't have their durability_policy set in the topo server. The upgrade path is to upgrade vtctld first, run `SetKeyspaceDurabilityPolicy` and then upgrade VTOrc. 

## Related Issue(s)

- #8975 

## Checklist

-   [x] Documentation was added or is not required https://github.com/vitessio/vitess/pull/10392 and website docs - https://github.com/vitessio/website/pull/1040

## Deployment Notes

When upgrading to release-14, VTOrc will not work until the durability policy is set in the topo server.

### DIFF
--- a/go/test/endtoend/vtorc/general/vtorc_test.go
+++ b/go/test/endtoend/vtorc/general/vtorc_test.go
@@ -406,3 +406,23 @@ func TestVtorcWithPrs(t *testing.T) {
 	utils.CheckPrimaryTablet(t, clusterInfo, replica, true)
 	utils.VerifyWritesSucceed(t, clusterInfo, replica, shard0.Vttablets, 10*time.Second)
 }
+
+// TestMultipleDurabilities tests that VTOrc works with 2 keyspaces having 2 different durability policies
+func TestMultipleDurabilities(t *testing.T) {
+	defer cluster.PanicHandler(t)
+	// Setup a normal cluster and start vtorc
+	utils.SetupVttabletsAndVtorc(t, clusterInfo, 1, 1, nil, cluster.VtorcConfiguration{}, 1, "")
+	// Setup a semi-sync cluster
+	utils.AddSemiSyncKeyspace(t, clusterInfo)
+
+	keyspaceNone := &clusterInfo.ClusterInstance.Keyspaces[0]
+	shardNone := &keyspaceNone.Shards[0]
+	utils.CheckPrimaryTablet(t, clusterInfo, shardNone.Vttablets[0], true)
+	utils.CheckReplication(t, clusterInfo, shardNone.Vttablets[0], shardNone.Vttablets[1:], 10*time.Second)
+
+	keyspaceSemiSync := &clusterInfo.ClusterInstance.Keyspaces[1]
+	shardSemiSync := &keyspaceSemiSync.Shards[0]
+	// find primary from topo
+	primary := utils.ShardPrimaryTablet(t, clusterInfo, keyspaceSemiSync, shardSemiSync)
+	assert.NotNil(t, primary, "should have elected a primary")
+}
--- a/go/test/endtoend/vtorc/general/vtorc_test.go
+++ b/go/test/endtoend/vtorc/general/vtorc_test.go
@@ -40,7 +40,7 @@ func TestPrimaryElection(t *testing.T) {
 	defer cluster.PanicHandler(t)
 	utils.SetupVttabletsAndVtorc(t, clusterInfo, 2, 1, nil, cluster.VtorcConfiguration{
 		PreventCrossDataCenterPrimaryFailover: true,
-	}, 2)
+	}, 2, "")
 	keyspace := &clusterInfo.ClusterInstance.Keyspaces[0]
 	shard0 := &keyspace.Shards[0]
 
@@ -67,7 +67,7 @@ func TestSingleKeyspace(t *testing.T) {
 	defer cluster.PanicHandler(t)
 	utils.SetupVttabletsAndVtorc(t, clusterInfo, 1, 1, []string{"--clusters_to_watch", "ks"}, cluster.VtorcConfiguration{
 		PreventCrossDataCenterPrimaryFailover: true,
-	}, 1)
+	}, 1, "")
 	keyspace := &clusterInfo.ClusterInstance.Keyspaces[0]
 	shard0 := &keyspace.Shards[0]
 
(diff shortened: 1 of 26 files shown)


go/vt/orchestrator/config/config.go
```diff
@@ -68,7 +68,6 @@ type Configuration struct {
 	ListenSocket                                string // Where orchestrator HTTP should listen for unix socket (default: empty; when given, TCP is disabled)
 	HTTPAdvertise                               string // optional, for raft setups, what is the HTTP address this node will advertise to its peers (potentially use where behind NAT or when rerouting ports; example: "http://11.22.33.44:3030")
 	AgentsServerPort                            string // port orchestrator agents talk back to
-	Durability                                  string // The type of durability to enforce. Default is "none". Other values are dictated by registered plugins
```
GuptaManan100, line 71: Durability policy config has been cleaned up! 😍

go/vt/orchestrator/inst/analysis_dao.go
```diff
@@ -58,6 +59,7 @@ func initializeAnalysisDaoPostConfiguration() {
 type clusterAnalysis struct {
 	hasClusterwideAction bool
 	primaryKey           *InstanceKey
+	durability           reparentutil.Durabler
```
GuptaManan100, line 62: Durability is stored at the cluster level to avoid creating a new one for all the tablets of a shard. This could move to a keyspace level, but we don't have keyspace level common things yet.

go/vt/orchestrator/inst/analysis_dao_test.go
```diff
@@ -424,22 +440,52 @@ func TestGetReplicationAnalysis(t *testing.T) {
...
+					Keyspace:      "ks",
+					Shard:         "0",
+					Type:          topodatapb.TabletType_REPLICA,
+					MysqlHostname: "localhost",
+					MysqlPort:     6709,
+				},
+				LastCheckValid: 1,
+			}},
+			// We will ignore these keyspaces too until the durability policy is set in the topo server
+			codeWanted: NoProblem,
```
GuptaManan100, line 484: If the durability policy is not specified in the topo server, VTOrc will ignore the keyspace.

go/vt/orchestrator/inst/keyspace_dao_test.go
```diff
@@ -0,0 +1,117 @@
...
+
+	"github.com/stretchr/testify/require"
+
+	"vitess.io/vitess/go/vt/orchestrator/db"
+	topodatapb "vitess.io/vitess/go/vt/proto/topodata"
+	"vitess.io/vitess/go/vt/topo"
+	"vitess.io/vitess/go/vt/topotools"
+)
+
+func TestSaveAndReadKeyspace(t *testing.T) {
```
GuptaManan100, line 30: I am very happy with this entire test file, since it spawns a sqlite database in memory and uses it to verify that the queries in the helper functions are correct. I would like to add similar tests to other vtorc queries too

go/vt/orchestrator/logic/topology_recovery.go
```diff
@@ -1370,16 +1370,10 @@ func executeCheckAndRecoverFunction(analysisEntry inst.ReplicationAnalysis, cand
...
+		err := RefreshKeyspace(analysisEntry.AnalyzedKeyspace)
 		if err != nil {
 			return false, nil, err
 		}
-		if ksInfo.GetDurabilityPolicy() != "" && ksInfo.GetDurabilityPolicy() != config.Config.Durability {
-			err = log.Errorf("executeCheckAndRecoverFunction: Analysis: %+v, InstanceKey: %+v: Keyspace has a different durability policy configured than this vtorc instance: %v",
-				analysisEntry.Analysis, analysisEntry.AnalyzedInstanceKey, analysisEntry.AnalyzedKeyspace)
-			return false, nil, err
-		}
-
```
GuptaManan100, line 1382: We no longer need this check, since we are now refreshing the keyspace which will reload the durability policy if it has changed, followed by a check for if failure is actually required.

go/test/endtoend/vtorc/general/vtorc_test.go
```diff
@@ -407,3 +406,23 @@ func TestVtorcWithPrs(t *testing.T) {
...
+	shardNone := &keyspaceNone.Shards[0]
+	utils.CheckPrimaryTablet(t, clusterInfo, shardNone.Vttablets[0], true)
+	utils.CheckReplication(t, clusterInfo, shardNone.Vttablets[0], shardNone.Vttablets[1:], 10*time.Second)
+
+	keyspaceSemiSync := &clusterInfo.ClusterInstance.Keyspaces[1]
+	shardSemiSync := &keyspaceSemiSync.Shards[0]
+	// find primary from topo
+	primary := utils.ShardPrimaryTablet(t, clusterInfo, keyspaceSemiSync, shardSemiSync)
+	assert.NotNil(t, primary, "should have elected a primary")
+}
```
GuptaManan100, line 428: This test is very simple. All it does is have two unintialized keyspaces with 2 different durability policies and waits for VTOrc to fix them both. Earlier this would have failed with VTOrc refusing to fix the keyspace with a different durability than its own config. But now this works 😻

go/vt/orchestrator/inst/analysis_dao_test.go
deepthi: hmm.. shouldn't vtorc still watch/repair the keyspace and default the durability policy to "none"?

deepthi (commented):
Nice work!
I had only one question. We can merge this once that is resolved.

go/vt/orchestrator/inst/analysis_dao_test.go
deepthi: Based on offline discussion, this is fine since
(a) VTOrc is not GA yet
(b) For anyone already using it, there will be a documented upgrade path in the release notes to avoid getting into this situation in the first place.

PR #10425, opened 2022-06-03
Addition of more logging to ERS and tests to PRS

GuptaManan100:
## Description

This PR makes two minor changes - 
1. Adds 2 more logging statements to ERS for better debugging
2. Adds a test to planned reparent shard tests to verify that it sets the correct value of semi-sync when performing promotion.

## Checklist

### DIFF
--- a/go/vt/vtctl/reparentutil/replication.go
+++ b/go/vt/vtctl/reparentutil/replication.go
@@ -255,10 +255,10 @@ func stopReplicationAndBuildStatusMaps(
 
 				primaryStatus, err = tmc.DemotePrimary(groupCtx, tabletInfo.Tablet)
 				if err != nil {
-					msg := "replica %v thinks it's primary but we failed to demote it"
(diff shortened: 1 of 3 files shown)


go/vt/vtctl/reparentutil/replication.go
```diff
@@ -256,7 +258,7 @@ func stopReplicationAndBuildStatusMaps(
 					msg := "replica %v thinks it's primary but we failed to demote it"
 					err = vterrors.Wrapf(err, msg+": %v", alias, err)
 
-					logger.Warningf(msg, alias)
+					logger.Warningf(msg+": %v", alias, err)
```
ajm188, line 261: suggestion (GitHub won't let me on deleted lines):

```go
msg := "replica %v thinks it's primary but we failed to demote it: %v"
err = vterrors.Wrapf(err, msg, alias, err)
logger.Warningf(msg, alias, err)
```

ajm188 (approved):

PR #10426, opened 2022-06-03
Make vtgate streamlog buffer configurable

DeathBorn:
If vtgate streamlog channel becomes full, query logs will be dropped. 
This makes its size configurable.

PR #10427, opened 2022-06-03
Add SessionUUID and transaction mark to vtgate query logs

DeathBorn:
We are doing query log analysis and it is very hard to reconstruct transactions. Adding session id ~~and transaction mark ~~can really help to identify client transactions and query path.

Before logging any query with `logStats.Send()`, augment it with 
* client `SessionUUID`
* ~~and `InTransaction` which tells us if we are in transaction after executing SQL statement.~~
  *  ~~`BEGIN` and everything between `BEGIN` and `COMMIT/ROLLBACK` will be logged with `InTransaction: true`~~
  * ~~`COMMIT/ROLLBACK` will be logged with `InTransaction: false`.~~
  * ~~If `autocommit=0` then uncommited statements will be logged with `InTransaction: true`~~
  
These changes do not look into vtgate <-> vttablet communication.

* **UPDATE** - `InTransaction` will be tested in our fork.


deepthi:
In general it should be fine to add things to logs at the end, as long as it is not PII (which is usually in the bind vars and nowhere else). We should wait for someone from @vitessio/query-serving to actually approve the PR.
In the meantime @DeathBorn can you fix the DCO? Since that requires a force-push we'll run CI after that.

DeathBorn:
> In general it should be fine to add things to logs at the end, as long as it is not PII (which is usually in the bind vars and nowhere else). We should wait for someone from @vitessio/query-serving to actually approve the PR.
> In the meantime @DeathBorn can you fix the DCO? Since that requires a force-push we'll run CI after that.

Will fix this

shlomi-noach:
@DeathBorn thank you for your contribution! A question, if you will: is there a particular reason you use `force-push` instead of `push`? Generally speaking, the Vitess project prefers `push` over `force-push`. With `force-push`, any maintainer review becomes meaningless because it loses context, which means any new `force-push` requires a complete re-review. With normal `push`, we can suffice with the specific commit changes.

DeathBorn:
> @DeathBorn thank you for your contribution! A question, if you will: is there a particular reason you use `force-push` instead of `push`? Generally speaking, the Vitess project prefers `push` over `force-push`. With `force-push`, any maintainer review becomes meaningless because it loses context, which means any new `force-push` requires a complete re-review. With normal `push`, we can suffice with the specific commit changes.

Had to edit single (and there is only 1) commit for proper signing, normally I don't use force-push.

harshit-gangal:
I think it is good idea to send sessionUUID on newStats as that do not change.

I am currently not sure for the usage/significance of logging inTransaction. If Vitess creates an internal transaction that would anyway will not be visible to the caller as in the end the inTransaction will be `false`

DeathBorn:
@harshit-gangal 
> I think it is good idea to send sessionUUID on newStats as that do not change.
> 
ok. will move it there

> I am currently not sure for the usage/significance of logging inTransaction. If Vitess creates an internal transaction that would anyway will not be visible to the caller as in the end the inTransaction will be `false`
> If we just log the sessionUID then it will be similar to how we see at mysql in general logs and can use that information to club the queries per session.

We noticed that sometimes logs can be dropped if buffer is full https://github.com/vitessio/vitess/pull/10426 and it was  somewhat frustrating when begin or commit/rollback statements were not logged - one could could see strange cases with consecutive begin or commit statements. `inTransaction` could help to group statements even if something is missing. However, reading session properties introduce additional mutex locks.. So, I think this really needs some thoughts since query logging is best effort anyway.

DeathBorn:
@harshit-gangal Ok. Lets add sessionUUID only. We will test `InTransaction` presence in our fork.

DeathBorn:
@harshit-gangal can't see if this fail is related to my change...

harshit-gangal:
> @harshit-gangal can't see if this fail is related to my change...

could you please merge from main and push again.

frouioui (commented):
Reviewed and everything looks good to me. I will let someone with more context on the subject to approve the pull request.

harshit-gangal (approved):

DeathBorn:
This needs one more approval

frouioui (approved):

frouioui:
@DeathBorn, hello, lgtm! This branch has a conflict with `main`:
> This branch has conflicts that must be resolved

DeathBorn:
@frouioui merged main once again in order to resolve conflicts.

PR #10429, opened 2022-06-03
Fail VReplication workflows on errors that persist and unrecoverable errors

rohit-nayak-ps:
## Description

As part of an initial design decision, VReplication workflows always retry in case it encounters an error after sleeping for 5 seconds. The reasoning was that, for large reshards/migrations and perpetual materialize workflows, we could often encounter recoverable errors like PRS, restarting of vttablets/mysql servers, network partitions etc. So rather than error out waiting for an operator to manually restart workflows we decided to keep retrying. 

Since we only retried every five seconds any resource wastage due to continuously retrying unrecoverable workflows would be small and in most cases we would transparently recover and make forward progress with minimum downtime. This is especially important for Materialize workflows were the user is expecting near realtime performance.

Usually the vreplication workflows would be setup manually and the possibility of errors due to schema issues was minimal and so this approach worked well. However with the introduction of vreplication-based online DDL workflows we see a lot of automated use where user-specified DDLs are directly used to configure vreplication workflows.  Incorrect DDLs can thus result in errors that result in prolonged retries that are not recoverable.  

Error reporting in VReplication is also not great: we update the `message` column in the `_vt.vreplication` table, but that can get overwritten when we retry. We do also log errors in the `_vt.vreplication_log` table

A change was introduced recently in Online DDL workflows to mitigate this: we look up the error against a set of MySQL errors that we knew were not recoverable and in that case we put the workflow in an error state. Then there are no more automated retries and a manual restart after fixing the error is expected.

However there are still unrecoverable schema-related errors that are not yet mapped or do not map cleanly to MySQL errors.  There could also be misconfigured workflows (example: no replicas in a keyspace when the tablet type is set to only replicas, incorrect cell settings etc). Continuously retrying workflows in such cases can delay detecting them.

This PR:
* extends the check for unrecoverable errors to all workflow types, not just Online DDLs
* for all workflows, detects errors that persist for more than the &#x1F6A9; **new vttablet flag** 
`--vreplication_max_time_to_retry_errors`  (default: 15 minutes). 

For above cases it directly moves the workflow to `Error` state, which is then reported in `Workflow Show`.

## Checklist

    - We should backport this to 14.0.0-rc, but no further

### DIFF
--- a/go/vt/vttablet/tabletmanager/vreplication/controller.go
+++ b/go/vt/vttablet/tabletmanager/vreplication/controller.go
@@ -46,7 +46,9 @@ var (
 	_          = flag.Duration("vreplication_healthcheck_topology_refresh", 30*time.Second, "refresh interval for re-reading the topology")
 	_          = flag.Duration("vreplication_healthcheck_retry_delay", 5*time.Second, "healthcheck retry delay")
 	_          = flag.Duration("vreplication_healthcheck_timeout", 1*time.Minute, "healthcheck retry delay")
-	retryDelay = flag.Duration("vreplication_retry_delay", 5*time.Second, "delay before retrying a failed binlog connection")
+	retryDelay = flag.Duration("vreplication_retry_delay", 5*time.Second, "delay before retrying a failed workflow event in the replication phase")
+
+	maxTimeToRetryErrors = flag.Duration("vreplication_max_time_to_retry_errors", 1*time.Minute, "stop trying to retry after this time")
 )
 
 // controller is created by Engine. Members are initialized upfront.
@@ -69,6 +71,8 @@ type controller struct {
 
 	// The following fields are updated after start. So, they need synchronization.
 	sourceTablet sync2.AtomicString
+
+	lastWorkflowError *lastError
 }
 
 // newController creates a new controller. Unless a stream is explicitly 'Stopped',
@@ -78,12 +82,13 @@ func newController(ctx context.Context, params map[string]string, dbClientFactor
 		blpStats = binlogplayer.NewStats()
 	}
 	ct := &controller{
-		vre:             vre,
-		dbClientFactory: dbClientFactory,
-		mysqld:          mysqld,
-		blpStats:        blpStats,
-		done:            make(chan struct{}),
-		source:          &binlogdatapb.BinlogSource{},
+		vre:               vre,
+		dbClientFactory:   dbClientFactory,
+		mysqld:            mysqld,
+		blpStats:          blpStats,
+		done:              make(chan struct{}),
+		source:            &binlogdatapb.BinlogSource{},
+		lastWorkflowError: newLastError("VReplication Controller", *maxTimeToRetryErrors),
 	}
 	log.Infof("creating controller with cell: %v, tabletTypes: %v, and params: %v", cell, tabletTypesStr, params)
 
@@ -95,9 +100,10 @@ func newController(ctx context.Context, params map[string]string, dbClientFactor
 	ct.id = uint32(id)
 	ct.workflow = params["workflow"]
 
-	blpStats.State.Set(params["state"])
-	// Nothing to do if replication is stopped.
-	if params["state"] == binlogplayer.BlpStopped {
+	state := params["state"]
+	blpStats.State.Set(state)
+	// Nothing to do if replication is stopped or is known to have an unrecoverable error.
+	if state == binlogplayer.BlpStopped || state == binlogplayer.BlpError {
 		ct.cancel = func() {}
 		close(ct.done)
 		return ct, nil
@@ -161,8 +167,10 @@ func (ct *controller) run(ctx context.Context) {
 			return
 		default:
 		}
-		binlogplayer.LogError(fmt.Sprintf("error in stream %v, retrying after %v", ct.id, *retryDelay), err)
+
 		ct.blpStats.ErrorCounts.Add([]string{"Stream Error"}, 1)
+		binlogplayer.LogError(fmt.Sprintf("error in stream %v, retrying after %v", ct.id, *retryDelay), err)
+		log.Flush()
 		timer := time.NewTimer(*retryDelay)
 		select {
 		case <-ctx.Done():
@@ -270,18 +278,16 @@ func (ct *controller) runBlp(ctx context.Context) (err error) {
 
 		vr := newVReplicator(ct.id, ct.source, vsClient, ct.blpStats, dbClient, ct.mysqld, ct.vre)
 		err = vr.Replicate(ctx)
-		if isUnrecoverableError(err) {
-			settings, _, errSetting := vr.readSettings(ctx)
-			if errSetting != nil {
-				return err // yes, err and not errSetting.
+
+		ct.lastWorkflowError.record(err)
+		if isUnrecoverableError(err) /* mysql error that we know needs manual intervention */ ||
+			!ct.lastWorkflowError.canRetry() /* cannot detect if this is recoverable, but it is persisting too long */ {
+
+			log.Errorf("vreplication stream %d going into error state due to %+v", ct.id, err)
+			if errSetState := vr.setState(binlogplayer.BlpError, err.Error()); errSetState != nil {
+				return err // yes, err and not errSetState.
 			}
-			if settings.WorkflowType == int64(binlogdatapb.VReplicationWorkflowType_ONLINEDDL) {
-				// Specific to OnlineDDL, if we encounter an "unrecoverable error", we change the migration state into Error and then we quit the workflow
-				if errSetState := vr.setState(binlogplayer.BlpError, err.Error()); errSetState != nil {
-					return err // yes, err and not errSetState.
-				}
-				return nil // this will cause vreplicate to quit the workflow
-			}
+			return nil // this will cause vreplicate to quit the workflow
 		}
 		return err
 	}
--- a/go/vt/vttablet/tabletmanager/vreplication/controller.go
+++ b/go/vt/vttablet/tabletmanager/vreplication/controller.go
@@ -46,9 +46,9 @@ var (
 	_          = flag.Duration("vreplication_healthcheck_topology_refresh", 30*time.Second, "refresh interval for re-reading the topology")
 	_          = flag.Duration("vreplication_healthcheck_retry_delay", 5*time.Second, "healthcheck retry delay")
 	_          = flag.Duration("vreplication_healthcheck_timeout", 1*time.Minute, "healthcheck retry delay")
-	retryDelay = flag.Duration("vreplication_retry_delay", 5*time.Second, "delay before retrying a failed workflow event in the replication phase")
+	retryDelay = flag.Duration("vreplication_retry_delay", 5*time.Second, "delay before retrying a failed workflow event in the replication phase (min: 5 seconds)")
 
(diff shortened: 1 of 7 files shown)


go/vt/vttablet/tabletmanager/vreplication/controller.go
```diff
@@ -46,7 +46,9 @@ var (
 	_          = flag.Duration("vreplication_healthcheck_topology_refresh", 30*time.Second, "refresh interval for re-reading the topology")
 	_          = flag.Duration("vreplication_healthcheck_retry_delay", 5*time.Second, "healthcheck retry delay")
 	_          = flag.Duration("vreplication_healthcheck_timeout", 1*time.Minute, "healthcheck retry delay")
-	retryDelay = flag.Duration("vreplication_retry_delay", 5*time.Second, "delay before retrying a failed binlog connection")
+	retryDelay = flag.Duration("vreplication_retry_delay", 5*time.Second, "delay before retrying a failed workflow event in the replication phase")
+
+	maxTimeToRetryErrors = flag.Duration("vreplication_max_time_to_retry_errors", 15*time.Minute, "stop trying to retry after this time")
```
deepthi, line 51:
```suggestion
	maxTimeToRetryErrors = flag.Duration("vreplication_max_time_to_retry_errors", 15*time.Minute, "stop retrying after this time")
```

deepthi (commented):
@ajm188 what is the guidance on new command-line flags added to binaries?
dash or underscore?

go/vt/vttablet/tabletmanager/vreplication/controller.go
mattlord: Should be dashes but I also think that consistency is important and that we should migrate all Vitess flag names from underscores to dashes, either early in 15.0 or 16.0 SNAPSHOT.

deepthi: that makes sense. we can do them all at once (and allow both versions for at least one release).

go/vt/vttablet/tabletmanager/vreplication/last_error.go
```diff
@@ -0,0 +1,69 @@
...
+		le.lastErrorStartTime = time.Time{}
+		return
+	}
+	if le.lastError == nil || !strings.EqualFold(err.Error(), le.lastError.Error()) {
+		le.lastErrorStartTime = time.Now()
+	}
+	le.lastError = err
+}
+
+func (le *lastError) canRetry() bool {
```
mattlord, line 60: Again, IMO we should call this `shouldRetry()` as we always can. It's an (annoying) nit though, I know.

go/vt/vttablet/tabletmanager/vreplication/last_error.go
```diff
@@ -0,0 +1,69 @@
...
+	le.lastErrorMu.Lock()
+	defer le.lastErrorMu.Unlock()
+	if !time.Time.IsZero(le.lastErrorStartTime) && time.Since(le.lastErrorStartTime) > le.maxTimeInError {
+		log.Errorf("Got same error since %s, will not retry anymore: you will need to manually restart workflow once error '%s' is fixed",
```
mattlord, line 64:
```
log.Errorf("The same error has been seen continuously since %s, we will assume this is a non-recoverable error and will not retry anymore: workflow will need to be manually restarted once error '%s' has been addressed",
```

mattlord (commented):
I had a few questions/nits, but I LOVE this! We can then store this information e.g. in the `_vt.vdiff` table too.

The only thing that kept me from approving -- as most of my comments are not blockers and subjective preference -- is that I think using `vterror` would be better. What do you think?

Thanks!

go/vt/vttablet/tabletmanager/vreplication/controller.go
```diff
@@ -161,8 +167,10 @@ func (ct *controller) run(ctx context.Context) {
 			return
 		default:
 		}
-		binlogplayer.LogError(fmt.Sprintf("error in stream %v, retrying after %v", ct.id, *retryDelay), err)
+
 		ct.blpStats.ErrorCounts.Add([]string{"Stream Error"}, 1)
+		binlogplayer.LogError(fmt.Sprintf("error in stream %v, retrying after %v", ct.id, *retryDelay), err)
+		log.Flush()
```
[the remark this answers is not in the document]
shlomi-noach: Though I'm not sure why `flush` is used here in particular, I think this point of code is infrequent enough that it can justify ` `flush`.

go/vt/vttablet/tabletmanager/vreplication/controller.go
```diff
@@ -270,18 +278,16 @@ func (ct *controller) runBlp(ctx context.Context) (err error) {
...
-			}
-			if settings.WorkflowType == int64(binlogdatapb.VReplicationWorkflowType_ONLINEDDL) {
-				// Specific to OnlineDDL, if we encounter an "unrecoverable error", we change the migration state into Error and then we quit the workflow
-				if errSetState := vr.setState(binlogplayer.BlpError, err.Error()); errSetState != nil {
-					return err // yes, err and not errSetState.
-				}
-				return nil // this will cause vreplicate to quit the workflow
+
+		ct.lastWorkflowError.record(err)
+		if isUnrecoverableError(err) /* mysql error that we know needs manual intervention */ ||
```
shlomi-noach, line 283: The `||` at the end of the line is hidden because of the comment, and due to `gofmt` indentation, it _looks like_ `!ct.lastWorkflowError.canRetry()` is somehow inside the `if` block. Let's move `||` before the code comment for better clarity
```suggestion
		if isUnrecoverableError(err) || // mysql error that we know needs manual intervention 
```

go/vt/vttablet/tabletmanager/vreplication/controller.go
```diff
@@ -270,18 +278,16 @@ func (ct *controller) runBlp(ctx context.Context) (err error) {
...
+			!ct.lastWorkflowError.canRetry() /* cannot detect if this is recoverable, but it is persisting too long */ {
```
shlomi-noach, line 284: Let's move `{` before the code comment for better clarity
```suggestion
			!ct.lastWorkflowError.canRetry() { // cannot detect if this is recoverable, but it is persisting too long
```

shlomi-noach (changes requested):
This looks great! The code change is small and simple and solves a big problem. My only request for change is dealing with unexpected input value for `--vreplication_max_time_to_retry_errors`

mattlord:
@rohit-nayak-ps I tried to address my own nits and comments here: https://github.com/vitessio/vitess/pull/10429/commits/6e2630840aaa287c3ea6a21ac7d0b61947dd24a6

I can revert some or all of it if you disagree, but I thought you'd be fine with them so attempted to save you some time. 🙂 

@shlomi-noach for the input validation, were you thinking something like this?
```
diff --git a/go/vt/vttablet/tabletmanager/vreplication/controller.go b/go/vt/vttablet/tabletmanager/vreplication/controller.go
index cf906cebcf..945fcc695c 100644
--- a/go/vt/vttablet/tabletmanager/vreplication/controller.go
+++ b/go/vt/vttablet/tabletmanager/vreplication/controller.go
@@ -48,7 +48,7 @@ var (
        _          = flag.Duration("vreplication_healthcheck_timeout", 1*time.Minute, "healthcheck retry delay")
        retryDelay = flag.Duration("vreplication_retry_delay", 5*time.Second, "delay before retrying a failed workflow event in the replication phase")

-       maxTimeToRetryError = flag.Duration("vreplication_max_time_to_retry_on_error", 15*time.Minute, "stop automatically retrying when we've had consecutive failures with the same error for this long after the first occurrence")
+       maxTimeToRetryError = flag.Duration("vreplication_max_time_to_retry_on_error", 15*time.Minute, "stop automatically retrying when we've had consecutive failures with the same error for this long after the first occurrence (min: vreplication_retry_delay * 5, max: 24 hours)")
 )

 // controller is created by Engine. Members are initialized upfront.
@@ -81,6 +81,11 @@ func newController(ctx context.Context, params map[string]string, dbClientFactor
        if blpStats == nil {
                blpStats = binlogplayer.NewStats()
        }
+       if *maxTimeToRetryError < *retryDelay*5 {
+               *maxTimeToRetryError = *retryDelay * 5
+       } else if *maxTimeToRetryError > 24*time.Hour {
+               *maxTimeToRetryError = 24 * time.Hour
+       }
        ct := &controller{
                vre:               vre,
                dbClientFactory:   dbClientFactory,
```

Or did I misunderstand? Thanks!

shlomi-noach:
> @shlomi-noach for the input validation, were you thinking something like this?

Sure! Although, I notice this adds a new dependency on `--vreplication_retry_delay`, which in itself isn't constrained (what happens if `--vreplication_retry_delay` is zero or negative?)

shlomi-noach:
All, let's pursue this PR? As far as I'm concerned, there is just the remaining issue of validating the input; but I'm also OK to merge this as-is to get this fix sooner into `main`, in the assumption that we will follow up with validation on a different PR. Opinions?

rohit-nayak-ps:
> All, let's pursue this PR? As far as I'm concerned, there is just the remaining issue of validating the input; but I'm also OK to merge this as-is to get this fix sooner into `main`, in the assumption that we will follow up with validation on a different PR. Opinions?

Agreed that we should merge this soon. I will take a fine look at this later today to review the newer commits and aim to push this today itself.

shlomi-noach (approved):

go/vt/vttablet/tabletmanager/vreplication/controller.go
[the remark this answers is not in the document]
rohit-nayak-ps: Was leftover from a debug session, removed

go/vt/vttablet/tabletmanager/vreplication/controller.go
```diff
@@ -270,18 +278,16 @@ func (ct *controller) runBlp(ctx context.Context) (err error) {
...
+
+			log.Errorf("vreplication stream %d going into error state due to %+v", ct.id, err)
+			if errSetState := vr.setState(binlogplayer.BlpError, err.Error()); errSetState != nil {
+				return err // yes, err and not errSetState.
```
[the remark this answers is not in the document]
rohit-nayak-ps: Wasn't earlier, so leaving it as such now.

mattlord (approved):
Had a very small nit, but LGTM!

go/vt/vttablet/tabletmanager/vreplication/utils.go
```diff
@@ -161,6 +163,7 @@ func isUnrecoverableError(err error) bool {
 		mysql.ERInvalidCastToJSON,
 		mysql.ERJSONValueTooBig,
 		mysql.ERJSONDocumentTooDeep:
+		log.Errorf("got unrecoverable error: %v", sqlErr)
```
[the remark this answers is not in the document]
shlomi-noach: I thought it was the other way around? That we had to uncapitalize everything, because those messages are preceded by prefix like `Error: `?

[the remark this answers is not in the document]
shlomi-noach: Capitalized per request.

shlomi-noach:
`Cluster (shardedrecovery_stress_verticalsplit_heavy)` is super flaky.

shlomi-noach:
Backport: https://github.com/vitessio/vitess/pull/10573

PR #10431, opened 2022-06-03
Aggregate static and other quick CI check actions to reduce #workflows

rohit-nayak-ps:
## Description

This PR combines several static checks and light-weight CI workflows. The aim is to reduce the total number of Github CI workflows hoping to improve CI stability and performance. If all sub-tests run it takes under 7 minutes. Normally it should be faster because of the file filters for each sub-task.

The following actions are now part of the Static Check, Etc. workflow:
* fossa
* check_formatting
* check_imports
* check_make_parser
* check_make_proto
* check_make_sizegen* check_make_visitor
* ensure_bootstrap_updated
* golangci-linter
* gomod-tidy

Some possible candidates like pr-labels and vtadmin related workflows have not been included: pr-labels, because it is often overlooked and seemed like a good idea to leave it as a separate workflow that fails early so that it can be fixed; vtadmin ones because they are relatively new and moreover it might be better to have a separate workflow for all vtadmin-related ones.

I tested the new workflow by temporarily adding some dummy changes in files that are part of all the filters.

## Checklist


mattlord:
💯

deepthi (approved):

PR #10437, opened 2022-06-05
test: refactor and enabled vschema ddl test

harshit-gangal:

## Description

This PR enables and refactor's a vschema ddl unit test which was skipped previously at it was not planned using plan builder.

## Checklist

### DIFF
--- a/go/vt/vtgate/executor_vschema_ddl_test.go
+++ b/go/vt/vtgate/executor_vschema_ddl_test.go
@@ -22,6 +22,8 @@ import (
 	"testing"
 	"time"
 
+	"vitess.io/vitess/go/test/utils"
+
 	"vitess.io/vitess/go/vt/callerid"
 	querypb "vitess.io/vitess/go/vt/proto/query"
 	vtrpcpb "vitess.io/vitess/go/vt/proto/vtrpc"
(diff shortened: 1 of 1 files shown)


go/vt/vtgate/executor_vschema_ddl_test.go
```diff
@@ -401,95 +403,65 @@ func TestExecutorAddDropVindexDDL(t *testing.T) {
...
 	wantqr = &sqltypes.Result{
 		Fields: buildVarCharFields("Columns", "Name", "Type", "Params", "Owner"),
 		Rows: [][]sqltypes.Value{
 			buildVarCharRow("id", "test_hash", "hash", "", ""),
 		},
-		RowsAffected: 1,
-	}
-	if !reflect.DeepEqual(qr, wantqr) {
-		t.Errorf("show vschema vindexes on TestExecutor.test:\n%+v, want\n%+v", qr, wantqr)
+		RowsAffected: 0,
```
GuptaManan100, line 454: Is the change in this expectation of `RowsAffected` alright?

go/vt/vtgate/executor_vschema_ddl_test.go
```diff
@@ -541,171 +504,112 @@ func TestExecutorAddDropVindexDDL(t *testing.T) {
...
-		t.Fatalf("error in show vschema vindexes on TestExecutor.test2: %v", err)
-	}
+	require.NoError(t, err)
 	wantqr = &sqltypes.Result{
 		Fields: buildVarCharFields("Columns", "Name", "Type", "Params", "Owner"),
 		Rows: [][]sqltypes.Value{
 			buildVarCharRow("id", "test_hash", "hash", "", ""),
 			buildVarCharRow("c1, c2", "test_lookup", "lookup", "from=c1,c2; table=test_lookup; to=keyspace_id", "test"),
 		},
-		RowsAffected: 2,
```
GuptaManan100, line 648: Same here

harshit-gangal: select / show queries does not have any value set for RowsAffected

harshit-gangal: they only make sense for DMLs

GuptaManan100 (approved):

go/vt/vtgate/executor_vschema_ddl_test.go
GuptaManan100: How was the test working previously then?

harshit-gangal: there was change where we stopped setting rowsAffected for non-dml queries.
But, this test was always in skip from along time, hence it was never changed (fixed).

PR #10438, opened 2022-06-06
Parsing Support for XML functions

K-Kumar-01:
## Description
Add parsing support for [XML functions](https://dev.mysql.com/doc/refman/8.0/en/xml-functions.html).

## Related Issue(s)

#8604 

## Checklist


GuptaManan100 (approved):
Looks good to me!

Issue #10400, opened 2022-05-31
Bug Report: Changing ThrottleMetricThreshold via debug/env not working properly  


gedgar:
### Overview of the Issue

Dynamically changing the `ThrottleMetricThreshold` via `/debug/env` Internal Variables appears to be broken.

I receive a confirmation message saying `Setting ThrottleMetricThreshold to: 23`, and the value appears to be persisted when I re-visit the page, however curl'ing the status endpoint still reflects the initial value:

```
curl -s http://localhost:15100/throttler/check | jq .
{
  "StatusCode": 200,
  "Value": 4,
  "Threshold": 50,
  "Message": ""
}
```

### Reproduction Steps

1) Enable vttablet throttler
2) Change ThrottleMetricThreshold via /debug/env 
3) Curl throttler status endpoint 
Expected: Threshold value reflects new `ThrottleMetricThreshold`
Actual: Threshold value reflects original/old `ThrottleMetricThreshold`

### Binary Version

```sh
Version: 14.0.0-SNAPSHOT (Git revision 49d92c53bfc8d7a692e27d3c970a510cf69e7a8b branch 'main') built on Tue May 31 09:39:46 PDT 2022 by gary@thinkpad using go1.18.1 linux/amd64
```
```


### Operating System and Environment details

```sh
cat /etc/os-release 
NAME="Ubuntu"
VERSION="20.04.4 LTS (Focal Fossa)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 20.04.4 LTS"
VERSION_ID="20.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=focal
UBUNTU_CODENAME=focal
```

```
Linux thinkpad 5.4.0-110-generic #124-Ubuntu SMP Thu Apr 14 19:46:19 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
```
```


### Log Fragments

_No response_

shlomi-noach:
> Dynamically changing the ThrottleMetricThreshold via /debug/env Internal Variables appears to be broken.

[...]

PR #7189, opened 2020-12-16
vttablet: debug/env page to change variables in real-time

sougou:
## Backport
NO

## Status
READY

## Description
Feature implemented as specified in RFC.
Also added links to the new page from `/debug/status`

## Related Issue(s)
Fixes #7171 
Fixes #2983

## Todos
- [X] Tests
- [ ] Documentation

## Deployment Notes
None

## Impacted Areas in Vitess
List general components of the application that this PR will affect:

- [X]  Query Serving
- [ ]  VReplication
- [ ]  Cluster Management
- [ ]  Build

PR #10439, opened 2022-06-06
Tablet throttler: dynamic ThrottleMetricThreshold

shlomi-noach:
## Description

Fixes https://github.com/vitessio/vitess/issues/10400

The tablet throttler code was contributed at about the same time as https://github.com/vitessio/vitess/pull/7189, and while superficially `/debug/env` supported a dynamic change of `ThrottleMetricThreshold`, the tablet throttler itself did not re-read that value dynamically; it only applied it once upon initialization.

As of this PR, the value is reflected dynamically. A change to `ThrottleMetricThreshold` will be reflected in the throttler's behavior, or can be seen in its API, e.g.:

```shell
% curl -s http://localhost:15100/throttler/check | jq .
{
  "StatusCode": 200,
  "Value": 4,
  "Threshold": 50,
  "Message": ""
}
```

Do note that change may take up to `10sec` to take effect, per [mysqlRefreshInterval](https://github.com/vitessio/vitess/blob/080c0079229aaed4e8d37a7cda359f13d336a393/go/vt/vttablet/tabletserver/throttle/throttler.go#L45).

We also take the opportunity to clean up some unused code.

Best reviewed [ignoring whitespace](https://github.com/vitessio/vitess/pull/10439/files?w=1).

## Related Issue(s)

#10400 

## Checklist

### DIFF
--- a/go/vt/vttablet/tabletserver/throttle/config/config.go
+++ b/go/vt/vttablet/tabletserver/throttle/config/config.go
@@ -27,11 +27,3 @@ type ConfigurationSettings struct {
 	EnableProfiling bool // enable pprof profiling http api
 	Stores          StoresSettings
 }
-
-// PostReadAdjustments validates and fixes config
-func (settings *ConfigurationSettings) PostReadAdjustments() error {
-	if err := settings.Stores.postReadAdjustments(); err != nil {
-		return err
-	}
-	return nil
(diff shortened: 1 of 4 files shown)


rohit-nayak-ps (approved):
lgtm

PR #10442, opened 2022-06-06
feat: read after write - wait for gtid executed set

harshit-gangal:
## Checklist


proto/query.proto
```diff
@@ -322,6 +322,8 @@ message ExecuteOptions {
   // if the user has created temp tables, Vitess will not reuse plans created for this session in other sessions.
   // The current session can still use other sessions cached plans.
   bool has_created_temp_tables = 12;
+
+  string gtid_set = 13;
```
deepthi, line 326: This field name seems too non-specific. Which gtid_set is this supposed to be?

deepthi (commented):
I know this is still Draft, but it's exciting!
A general comment:
Until now we have hidden `gtid` behind a generic name like `Position`. And we encapsulated all `gtid` logic behind a `Flavor` interface. We seem to be abandoning that with this change.
I know we are deprecating MariaDB support, but this is still an inconsistency that might trip up future readers/writers of code.

proto/query.proto
harshit-gangal: After offline discussion and thinking more about it. I think it would be better to call it position and use the flavour interface to interactive with this position.

harshit-gangal:
Closing for now, will open when starts working on it again.

PR #10370, opened 2022-05-26
Tablet throttler: serve metrics on /throttler/metrics

shlomi-noach:
## Description

This PR adds a `/throttler/metrics` API endpoint to `vttablet`, which exposes some Throttler metrics. For example:

```shell
$ curl -s 'http://127.0.0.1:15100/throttler/metrics' | jq . | grep throttler
  "throttler.aggregated.mysql.self": 0.123279,
  "throttler.aggregated.mysql.shard": 0.92436,
  "throttler.check.8a3b5a56_dce5_11ec_a840_0a43f95f28a3:vreplication:online-ddl.mysql.shard.total": 4,
  "throttler.check.8a3b5a56_dce5_11ec_a840_0a43f95f28a3:vreplication:online-ddl.total": 4,
  "throttler.check.any.error": 47,
  "throttler.check.any.mysql.self.error": 23,
  "throttler.check.any.mysql.self.total": 307,
  "throttler.check.any.mysql.shard.error": 24,
  "throttler.check.any.mysql.shard.total": 311,
  "throttler.check.any.total": 618,
  "throttler.check.freno.error": 47,
  "throttler.check.freno.mysql.self.error": 23,
  "throttler.check.freno.mysql.self.total": 307,
  "throttler.check.freno.mysql.shard.error": 24,
  "throttler.check.freno.mysql.shard.total": 307,
  "throttler.check.freno.total": 614,
  "throttler.check.mysql.self.seconds_since_healthy": 0,
  "throttler.check.mysql.shard.seconds_since_healthy": 0
```

The above shows us the aggregated metrics for existing metrics (first two lines), then check results for each app:

- `total` means how many checks were made by the app
- `error` are how many times the throttler returned with non-success, out of `total`
- `any` is a combination of all apps
- `mysql.shard` are checks for shard lag (the standard `/throttler/check` API call)
- `mysql.self` are checks for the state of the specific tablet's MySQL (`/throttler/check-self` API call)

## Related Issue(s)

#10045 

## Checklist

-   [ ] "Backport me!" label has been added if this change should be backported
-   [ ] Tests were added or are not required
-   [ ] Documentation was added or is not required

PR #10045, opened 2022-04-06
tablet lag throttler: small API improvements

shlomi-noach:
## Description
- `/throttler/throttle` now accepts `ratio` query param, `float64` in the range `[0..1]`

Example usage:
```sh
curl -s 'http://127.0.0.1:15100/throttler/throttle-app?app=vreplication&duration=1h&ratio=0.2' | jq .
```
```json
{
  "AppName": "vreplication",
  "ExpireAt": "2022-04-06T10:19:52.897780736Z",
  "Ratio": 0.2
}
```

- support `/throttler/throttled-apps` to get information about apps currently throttled.

Example output:
```sh
curl -s http://127.0.0.1:15100/throttler/throttled-apps | jq .
```
```json
{
  "abusing-app": {
    "Object": {
      "AppName": "abusing-app",
      "ExpireAt": "2032-04-03T09:18:30.331164282Z",
      "Ratio": 1
    },
    "Expiration": 0
  },
  "vreplication": {
    "Object": {
      "AppName": "vreplication",
      "ExpireAt": "2022-04-06T10:19:52.897780736Z",
      "Ratio": 0.2
    },
    "Expiration": 0
  }
}
```

- Tests to follow
- Release notes to follow

## Checklist
- [ ] Should this PR be backported?
- [x] Tests were added or are not required
- [x] Documentation was added or is not required

PR #10443, opened 2022-06-06
Throttler: stats in /debug/vars

shlomi-noach:
## Description

This PR exposes throttler metrics on `/debug/vars`. For example:

```shell
$ curl -s http://127.0.0.1:15100/debug/vars | jq . | grep 'throttler[.]'
  "throttler.aggregated.mysql.self": 133.19334,
  "throttler.aggregated.mysql.shard": 132.997847,
  "throttler.check.any.error": 1086,
  "throttler.check.any.mysql.self.error": 542,
  "throttler.check.any.mysql.self.total": 570,
  "throttler.check.any.mysql.shard.error": 544,
  "throttler.check.any.mysql.shard.total": 570,
  "throttler.check.any.total": 1140,
  "throttler.check.mysql.self.seconds_since_healthy": 132,
  "throttler.check.mysql.shard.seconds_since_healthy": 132,
  "throttler.check.vitess.error": 1086,
  "throttler.check.vitess.mysql.self.error": 542,
  "throttler.check.vitess.mysql.self.total": 570,
  "throttler.check.vitess.mysql.shard.error": 544,
  "throttler.check.vitess.mysql.shard.total": 570,
  "throttler.check.vitess.total": 1140,
  "throttler.probes.latency": 292982,
  "throttler.probes.total": 1138
```

The above shows us the aggregated metrics for existing metrics (first two lines), then check results for each app:

- `total` means how many checks were made by the app
- `error` are how many times the throttler returned with non-success, out of `total`
- `any` is a combination of all apps
- `mysql.shard` are checks for shard lag (the standard `/throttler/check` API call)
- `mysql.self` are checks for the state of the specific tablet's MySQL (`/throttler/check-self` API call)
- `vitess` is an internal check, that periodically hits the throttler for sanity and basic exposure of metrics.

### Implementation notes

The metrics exported here require a `float64` gauge. see for example `throttler.aggregated.mysql.shard`, which tells us the replicatoin lag on a shard. It is imperative that we have subsecond resolution, and a fraction number makes sense (it would be possible to achieve the same with `uint64` as nanoseconds, but we inherit the fraction behavior from `freno`, and it's been in `vitess` for multiple versions now).

To that effect, I created `Gauge64`, which in turn means changes in `prometheus`, `opentsdb`, `statsd`, `exporter`, .... 
Notable, `exporter` assumes all counters/gauges are `int64` based; notice I haven't found a good solution and implemented like so:
```go
func (e *Exporter) NewGaugeFloat64(name string, help string) *stats.GaugeFloat64 {
	return nil
}
```

Please review carefully those parts and let me know if there is a risk in there. 

## Related Issue(s)

- Replaces https://github.com/vitessio/vitess/pull/10370
- #10045 

## Checklist


deepthi:
Porting over my question from the previous PR:
1. Why are the metric names formatted the way they are? Vitess convention would be `ThrottlerAggregatedMysqlSelf` versus `throttler.aggregated.mysql.self`
2. What do they look like if you access the same metrics from the `/metrics` endpoint?

deepthi:
> I also question the value of the current vitess naming scheme? What is the advantage of single word CamelCase?

That is a valid question :) 
When you get metrics from `/metrics`, CamelCase gets converted to snake_case. E.g. `AppConnPoolActive` becomes `vttablet_app_conn_pool_active` (also prefixed with the binary from which the metrics are being emitted).
Hence my second question:
> What do they look like if you access the same metrics from the /metrics endpoint?

What I haven't looked at is whether this conversion is something we have implemented in our own Prometheus backend code.

deepthi:
Looks like we do convert `.` to `_` for Prometheus metrics, but it will be nice to verify on a running instance with these new metrics that is works as expected.
https://github.com/vitessio/vitess/blob/main/go/stats/snake_case_converter.go#L58
The only rationale for CamelCase at this point is consistency. `/debug/vars` is not meant to be piped to production metric systems, for that we have `/metrics`. And an integration with something like DataDog would require someone to create a new stats backend for it similar to what we have for prometheus, and that backend would do the conversion to the desired naming convention.

shlomi-noach:
This is what `/metrics` looks like:
```shell
$ curl -s http://127.0.0.1:15100/metrics | grep throttler
# HELP vttablet_throttler_aggregated_mysql_self aggregated value for mysql.self
# TYPE vttablet_throttler_aggregated_mysql_self gauge
vttablet_throttler_aggregated_mysql_self 60620.693474
# HELP vttablet_throttler_aggregated_mysql_shard aggregated value for mysql.shard
# TYPE vttablet_throttler_aggregated_mysql_shard gauge
vttablet_throttler_aggregated_mysql_shard 60620.497982
# HELP vttablet_throttler_check_any_error total number of failed checks
# TYPE vttablet_throttler_check_any_error counter
vttablet_throttler_check_any_error 484982
# HELP vttablet_throttler_check_any_mysql_self_error 
# TYPE vttablet_throttler_check_any_mysql_self_error counter
vttablet_throttler_check_any_mysql_self_error 242490
# HELP vttablet_throttler_check_any_mysql_self_total 
# TYPE vttablet_throttler_check_any_mysql_self_total counter
vttablet_throttler_check_any_mysql_self_total 242518
# HELP vttablet_throttler_check_any_mysql_shard_error 
# TYPE vttablet_throttler_check_any_mysql_shard_error counter
vttablet_throttler_check_any_mysql_shard_error 242492
# HELP vttablet_throttler_check_any_mysql_shard_total 
# TYPE vttablet_throttler_check_any_mysql_shard_total counter
vttablet_throttler_check_any_mysql_shard_total 242518
# HELP vttablet_throttler_check_any_total total number of checks
# TYPE vttablet_throttler_check_any_total counter
vttablet_throttler_check_any_total 485036
# HELP vttablet_throttler_check_mysql_self_seconds_since_healthy seconds since last healthy cehck for mysql.self
# TYPE vttablet_throttler_check_mysql_self_seconds_since_healthy gauge
vttablet_throttler_check_mysql_self_seconds_since_healthy 60619
# HELP vttablet_throttler_check_mysql_shard_seconds_since_healthy seconds since last healthy cehck for mysql.shard
# TYPE vttablet_throttler_check_mysql_shard_seconds_since_healthy gauge
vttablet_throttler_check_mysql_shard_seconds_since_healthy 60619
# HELP vttablet_throttler_check_vitess_error total number of failed checks for vitess
# TYPE vttablet_throttler_check_vitess_error counter
vttablet_throttler_check_vitess_error 484982
# HELP vttablet_throttler_check_vitess_mysql_self_error 
# TYPE vttablet_throttler_check_vitess_mysql_self_error counter
...
```

shlomi-noach:
> /debug/vars is not meant to be piped to production metric systems, for that we have /metrics

TIL and I really wasn't aware of `/metrics` until this issue.

shlomi-noach:
Let me look into camel casing metric names.

shlomi-noach:
Converting to Draft while looking into a few things

shlomi-noach:
@deepthi how about the following:

```shell
$ curl -s http://127.0.0.1:15100/debug/vars | jq . | grep Throttler | grep -v Pool
  "ThrottlerAggregatedMysqlSelf": 0.191718,
  "ThrottlerAggregatedMysqlShard": 0.960054,
  "ThrottlerCheckAnyError": 27,
  "ThrottlerCheckAnyMysqlSelfError": 13,
  "ThrottlerCheckAnyMysqlSelfTotal": 38,
  "ThrottlerCheckAnyMysqlShardError": 14,
  "ThrottlerCheckAnyMysqlShardTotal": 42,
  "ThrottlerCheckAnyTotal": 80,
  "ThrottlerCheckMysqlSelfSecondsSinceHealthy": 0,
  "ThrottlerCheckMysqlShardSecondsSinceHealthy": 0,
  "ThrottlerProbesLatency": 355523,
  "ThrottlerProbesTotal": 74,

$ curl -s http://127.0.0.1:15100/metrics | grep -i throttler | grep -v pool
# HELP vttablet_throttler_aggregated_mysql_self aggregated value for mysql.self
# TYPE vttablet_throttler_aggregated_mysql_self gauge
vttablet_throttler_aggregated_mysql_self 0.827354
# HELP vttablet_throttler_aggregated_mysql_shard aggregated value for mysql.shard
# TYPE vttablet_throttler_aggregated_mysql_shard gauge
vttablet_throttler_aggregated_mysql_shard 0.591876
# HELP vttablet_throttler_check_any_error total number of failed checks
# TYPE vttablet_throttler_check_any_error counter
vttablet_throttler_check_any_error 27
# HELP vttablet_throttler_check_any_mysql_self_error 
# TYPE vttablet_throttler_check_any_mysql_self_error counter
vttablet_throttler_check_any_mysql_self_error 13
# HELP vttablet_throttler_check_any_mysql_self_total 
# TYPE vttablet_throttler_check_any_mysql_self_total counter
vttablet_throttler_check_any_mysql_self_total 57
# HELP vttablet_throttler_check_any_mysql_shard_error 
# TYPE vttablet_throttler_check_any_mysql_shard_error counter
vttablet_throttler_check_any_mysql_shard_error 14
# HELP vttablet_throttler_check_any_mysql_shard_total 
# TYPE vttablet_throttler_check_any_mysql_shard_total counter
vttablet_throttler_check_any_mysql_shard_total 64
# HELP vttablet_throttler_check_any_total total number of checks
# TYPE vttablet_throttler_check_any_total counter
vttablet_throttler_check_any_total 121
# HELP vttablet_throttler_check_mysql_self_seconds_since_healthy seconds since last healthy cehck for mysql.self
# TYPE vttablet_throttler_check_mysql_self_seconds_since_healthy gauge
vttablet_throttler_check_mysql_self_seconds_since_healthy 0
# HELP vttablet_throttler_check_mysql_shard_seconds_since_healthy seconds since last healthy cehck for mysql.shard
# TYPE vttablet_throttler_check_mysql_shard_seconds_since_healthy gauge
vttablet_throttler_check_mysql_shard_seconds_since_healthy 0
# HELP vttablet_throttler_probes_latency probes latency
# TYPE vttablet_throttler_probes_latency gauge
vttablet_throttler_probes_latency 347382
# HELP vttablet_throttler_probes_total total probes
# TYPE vttablet_throttler_probes_total counter
vttablet_throttler_probes_total 114
```

In the above:

- everything is camel cased
- I chose to (meanwhile?) not provide stats for specific apps/workflows. It can get ugly and spammy with UUIDs, an dcan cause bloating of `/debug/vars` and bloating of memory. 

I confess to dislike the CamelCase approach, because in a scenario where I need more than one word to describe something, such as in the above `SecondsSinceHealthy`, I want that description to be atomic, rather then be split into three parts.

deepthi:
In terms of naming, this now looks good.
I previously forgot to address another part of your comments, apologies for that, and I'll do it now.

> But, also, the metrics names are parameterized; mysql, self, shard are parameterized parts. Let's say we can call them MySQL, Self, Shard because we control them -- the app names are also parameterized. gh-ost, vreplication, ...

> The existing vitess names like RowStreamerMaxInnoDBTrxHistLen are constants.

> I do see examples of parameterized connection pool params.

> But I don't have clarity on how we might deal with parameterized app names.

The stats package allows you to attach labels to the same metric. For example,
```
# TYPE vttablet_query_row_counts counter
vttablet_query_row_counts{plan="Insert",table="corder"} 5
vttablet_query_row_counts{plan="Insert",table="customer"} 5
vttablet_query_row_counts{plan="Insert",table="product"} 2
vttablet_query_row_counts{plan="Select",table="corder"} 0
vttablet_query_row_counts{plan="Select",table="customer"} 0
vttablet_query_row_counts{plan="Select",table="dual"} 0
vttablet_query_row_counts{plan="Select",table="product"} 0
```
labels don't have to be any particular case as you can see from this example.
If self/shard are values of a particular property, then the property can be defined as a label, and these become the values that we emit for that label. Similarly `app` can be a label with certain values that we produce.
Popular monitoring tools all support querying / filtering by labels, so this approach works well for not needing to add new metrics every time we decide to support a new value for a certain property of the metric.

Does this help?

shlomi-noach:
> Does this help?

Yes, thank you!

go/stats/counter_map.go
```diff
@@ -0,0 +1,95 @@
+/*
+Copyright 2019 The Vitess Authors.
```
systay, line 2: 🙄

go/stats/counter_test.go
```diff
@@ -71,3 +71,31 @@ func TestGaugeFunc(t *testing.T) {
...
+func TestGaugeFloat64(t *testing.T) {
+	var gotname string
+	var gotv *GaugeFloat64
+	clear()
+	Register(func(name string, v expvar.Var) {
+		gotname = name
+		gotv = v.(*GaugeFloat64)
+	})
+	v := NewGaugeFloat64("f", "help")
+	if gotname != "f" {
```
systay, line 84: nit: I personally find stretchr assertions like `assert.Equal(t, 3.14, v.Get())` much easier to read

go/stats/counter_test.go
```diff
@@ -71,3 +71,31 @@ func TestGaugeFunc(t *testing.T) {
...
+		t.Errorf("want Int, got %s", gotname)
+	}
+	if gotv != v {
+		t.Errorf("want %#v, got %#v", v, gotv)
+	}
+	v.Set(3.14)
+	if v.Get() != 3.14 {
```
systay, line 91: nit: I personally find stretchr assertions like `assert.Equal(t, 3.14, v.Get())` much easier to read

go/textutil/strings_test.go
```diff
@@ -66,3 +66,45 @@ func TestSplitUnescape(t *testing.T) {
...
+		},
+		{
+			word:   "mySQL",
+			expect: "Mysql",
+		},
+	}
+	for _, tc := range tt {
+		t.Run(tc.word, func(t *testing.T) {
+			camel := SingleWordCamel(tc.word)
+			assert.Equal(t, tc.expect, camel)
```
systay, line 107: 🤩

go/vt/servenv/exporter.go
```diff
@@ -307,6 +307,10 @@ func (e *Exporter) NewGauge(name string, help string) *stats.Gauge {
 	return lvar
 }
 
+func (e *Exporter) NewGaugeFloat64(name string, help string) *stats.GaugeFloat64 {
+	return nil
```
systay, line 311: I don't follow why this is returning `nil`. Maybe add a comment?

systay (approved):

go/vt/servenv/exporter.go
shlomi-noach: See original comment, where this is explained. I'll also add as code comment.

go/stats/counter_test.go
shlomi-noach: You're absolutely right; I copied+pasted existing tests and kept the original code conventions (I assume this was written way before testify was in use). I'll update.

shlomi-noach: fixed

PR #10445, opened 2022-06-06
Formalize MySQL capabilities by flavor/version

shlomi-noach:
## Description

This PR formalizes _which_ MySQL flavor/version is capable of doing _what_.

Up till now, we have approached the question in two different ways, both of which are deprecated by this PR.

1. Implement specific functions in `flavor` interface, such as [`supportsFastDropTable()`](https://github.com/vitessio/vitess/blob/457541aa26700e032a66435ec501f1119a53d7db/go/mysql/flavor.go#L145)
2. identify [flavor families](https://github.com/vitessio/vitess/blob/457541aa26700e032a66435ec501f1119a53d7db/go/mysql/flavor.go#L43-L50), then let anyone who wants to [make decisions](https://github.com/vitessio/vitess/blob/457541aa26700e032a66435ec501f1119a53d7db/go/vt/vttablet/onlineddl/analysis.go#L245-L248) based on that. See https://github.com/vitessio/vitess/pull/10402

The problems with both approaches:

1. we are looking into growing our observation of capabilities: is the server able to run instant DDL that adds a column in the middle of the table? Does it support transactional `gtid_executed`? Does it support `backup` command? etc.
  If we declare a function for each of these, we'll never see the end of it. We'll keep changing the interface when mostly a single implementing flavor (namely, `mysql80`) has the potential to answer "yes" to these.
2. This scatters the logic outside the `mysql` package and outside the flavor, and it's a mess.

With this PR we change as follows:

- We define a `FlavorCapability` type
- We define a `iota` list of supported capabilities. If we wish to add a new capability, it's a new `iota` constant. e.g. `FastDropTableFlavorCapability` or `InstantAddLastColumnFlavorCapability`, etc.
- `flavor` has a single `supportsCapability(serverVersion string, capability FlavorCapability) (bool, error)` function.
- implementing flavors all need to implement this function: by default everyone just says `fase`
- Those flavors that _do_ have a capability, opt in to say so. Right now this is only `mysql80`. Most new capabilities we will add will only be in `mysql80`, and so only one flavor will need to be updated.
- `GetFlavor()` returns a `CapableOf(capability FlavorCapability)` function
- logic that wants to know whether the current flavor supports some capability, calls said `CapableOf()` function. See implementation in `onlineddl/analysis.go`.
- We get rid of `FlavorFamily`. Not needed anymore.

I believe this is much cleaner and more formal. The only place we compare e.g. to version `8.0.17` or `8.0.29` is within the flavor itself.
## Related Issue(s)

Refactors some of https://github.com/vitessio/vitess/pull/10402

## Checklist


go/mysql/flavor_mysqlgr.go
```diff
@@ -238,9 +238,12 @@ func (mysqlGRFlavor) baseShowTablesWithSizes() string {
 	return TablesWithSize80
 }
 
-// supportsFastDropTable is part of the Flavor interface.
-func (mysqlGRFlavor) supportsFastDropTable(c *Conn) (bool, error) {
-	return false, nil
+// supportsCapability is part of the Flavor interface.
+func (mysqlGRFlavor) supportsCapability(serverVersion string, capability FlavorCapability) (bool, error) {
```
deepthi, line 242: technically GR is orthogonal to the MySQL version. However, the fact that GR is a flavor will mean that none of the new features can be used with GR.
@y5w why is this flavor required? where is it useful?

deepthi (approved):
I'm going to approve this now. The GR question can be fixed forward once we determine the right solution.

go/mysql/flavor_mysqlgr.go
y5w: > the fact that GR is a flavor will mean that none of the new features can be used with GR.

Could you elaborate more what does this mean and what kind of "feature" are you having in mind?

It is a flavor so that we can override status (for vtgate's health check) based on different mysql query

shlomi-noach: > Could you elaborate more what does this mean and what kind of "feature" are you having in mind?

@y5w per "flavor" we might want to determine:

- does the database support INSTANT DDL?
- does it support transactional `gtid_executed`?
- does it support JSON columns?
- does it support the `BACKUP` command?
- etc.

I'm unaware where `mysqlGRFlavor` is used, but please see this followup PR which implements capabilities for `mysqlGRFlavor`: https://github.com/vitessio/vitess/pull/10451

shlomi-noach: @y5w thanks for explaining. I'l look into the refactor later on, thankyou

PR #10101, opened 2022-04-15
Upgrade main to `go1.18.1`

frouioui:
## Description
This Pull Request upgrades the golang version used on `main` to `go1.18.1`. The Go team has recently released a new patch release for `go1.18` that contains vulnerability fixes, that impact Vitess.

Snippet from the [Go release history website](https://go.dev/doc/devel/release#go1.18.minor):
> go1.18.1 (released 2022-04-12) includes security fixes to the crypto/elliptic, crypto/x509, and encoding/pem packages, as well as bug fixes to the compiler, linker, runtime, the go command, vet, and the bytes, crypto/x509, and go/types packages. See the [Go 1.18.1 milestone](https://github.com/golang/go/issues?q=milestone%3AGo1.18.1+label%3ACherryPickApproved) on our issue tracker for details.

Note that the minimum required Go version is `1.18.1`, if this requirement is not met, the following error will appear when building Vitess:
```
Version 1.18.1+ required. See https://vitess.io/contributing/build-from-source for install instructions.
```

## Checklist

-   [ ] "Backport me!" label has been added if this change should be backported
-   [x] Tests were added or are not required
-   [x] Documentation was added or is not required

PR #10447, opened 2022-06-06
Upgrade to `go1.18.3`

frouioui:
## Description

This Pull Request bumps the version of `go` from `go1.18.1` to `go1.18.3`. There are several security patches in the last two minor releases of `go` which affect packages used by Vitess. More information [here](https://go.dev/doc/devel/release#go1.18).

```
go1.18.2 (released 2022-05-10) includes security fixes to the syscall package, as well as bug fixes to the compiler, runtime, the go command, and the crypto/x509, go/types, net/http/httptest, reflect, and sync/atomic packages.

go1.18.3 (released 2022-06-01) includes security fixes to the crypto/rand, crypto/tls, os/exec, and path/filepath packages, as well as bug fixes to the compiler, and the crypto/tls and text/template/parse packages.
```

For reference, here is the link to the last go upgrade: https://github.com/vitessio/vitess/pull/10101.

## Checklist


deepthi (approved):
I see that
- bootstrap_version has been updated
- new bootstrap images are already available on dockerhub

Issue #9141, opened 2021-11-04
Error Running the 101_init_cluster.yaml


tonycoder0525:
Issues on GitHub are intended to be related to bugs or feature requests, so we recommend using our other community resources instead of asking here.

- [Vitess User Guide](https://vitess.io/user-guide/introduction/)
- Any other questions can be asked in the community [Slack workspace](https://vitess.io/slack)

I was trying to follow the guide on [Vitess Operator for Kubernetes](https://vitess.io/docs/get-started/operator/) and ran it on k8s v1.19.8, I found that it requires PersistantVolumnClaim and I added, modified as follows:

```yaml
# The following example is minimalist. The security policies
# and resource specifications are not meant to be used in production.
# Please refer to the operator documentation for recommendations on
# production settings.
apiVersion: planetscale.com/v2
kind: VitessCluster
metadata:
  name: example
spec:
  images:
    vtctld: vitess/lite:v12.0.0
    vtgate: vitess/lite:v12.0.0
    vttablet: vitess/lite:v12.0.0
    vtbackup: vitess/lite:v12.0.0
    mysqld:
      mysql56Compatible: vitess/lite:v12.0.0
    mysqldExporter: prom/mysqld-exporter:v0.11.0
  cells:
  - name: zone1
    gateway:
      authentication:
        static:
          secret:
            name: example-cluster-config
            key: users.json
      replicas: 1
      resources:
        requests:
          cpu: 100m
          memory: 256Mi
        limits:
          memory: 256Mi
  globalLockserver:
    etcd:
      dataVolumeClaimTemplate:
        accessModes: ["ReadWriteOnce"]
        storageClassName: vitess-local
        resources:
          requests:
            storage: 10Gi
  vitessDashboard:
    cells:
    - zone1
    extraFlags:
      security_policy: read-only
    replicas: 1
    resources:
      limits:
        memory: 128Mi
      requests:
        cpu: 100m
        memory: 128Mi

  keyspaces:
  - name: commerce
    turndownPolicy: Immediate
    partitionings:
    - equal:
        parts: 1
        shardTemplate:
          databaseInitScriptSecret:
            name: example-cluster-config
            key: init_db.sql
          replication:
            enforceSemiSync: false
          tabletPools:
          - cell: zone1
            type: replica
            replicas: 2
            vttablet:
              extraFlags:
                db_charset: utf8mb4
              resources:
                limits:
                  memory: 256Mi
                requests:
                  cpu: 100m
[...]

PR #10450, opened 2022-06-06
Halve request and double limit for RAM in the example cluster

mattlord:
## Description

Here we halve the request and double the limit for RAM in [the example cluster.](https://vitess.io/docs/14.0/get-started/operator/) We do this because:
1. We don't need/want everything to have the [Guaranteed QoS](https://kubernetes.io/docs/tasks/configure-pod-container/quality-service-pod/#create-a-pod-that-gets-assigned-a-qos-class-of-guaranteed) (request == limit)
2. It's common for mysqld (especially 8.0) to go beyond the old limit and end up in a crash loop

## Related Issue(s)

Fixes: https://github.com/vitessio/vitess/issues/9141

## Checklist

### DIFF
--- a/examples/operator/101_initial_cluster.yaml
+++ b/examples/operator/101_initial_cluster.yaml
@@ -37,11 +37,11 @@ spec:
       security_policy: read-only
     replicas: 1
     resources:
-      limits:
-        memory: 256Mi
       requests:
         cpu: 100m
         memory: 64Mi
+      limits:
+        memory: 256Mi
 
   keyspaces:
   - name: commerce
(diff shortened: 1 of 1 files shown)


examples/operator/101_initial_cluster.yaml
```diff
@@ -27,9 +27,9 @@ spec:
       resources:
         requests:
           cpu: 100m
-          memory: 256Mi
+          memory: 128Mi
         limits:
-          memory: 256Mi
+          memory: 512Mi
```
rsajwani, line 32: very minor: can you move limits above request , in order to be consistent with rest of the yaml file. will be easy to read. provided it does not break any thing.

mattlord: Sure! That lack of uniformity caught me at first too. I'll do that now, although I'll use requests,limits ordering as I think that's a more logical order for readers (does not matter in yaml parsing).

deepthi (approved):
We should have done this a long time ago for MySQL 8.0
Thank you for fixing it.

PR #10453, opened 2022-06-07
fix: change planner_version to planner-version everywhere

systay:
## Description
The new default for our flags will use hyphens instead of underscore, so this PR changes `planner_version` to `planner-version` everywhere.

Should be backported to r14

## Related Issue(s)
Docs change: https://github.com/vitessio/website/pull/1047
#10420

## Checklist

### DIFF
--- a/go/test/endtoend/tabletmanager/qps_test.go
+++ b/go/test/endtoend/tabletmanager/qps_test.go
@@ -60,9 +60,7 @@ func TestQPS(t *testing.T) {
 	//       after that we'll see 0.0 QPS rates again. If this becomes actually
 	//       flaky, we need to read continuously in a separate thread.
 
-	n := 0
-	for n < 15 {
-		n++
+	for n := 0; n < 15; n++ {
 		// Run queries via vtGate so that they are counted.
 		utils.Exec(t, vtGateConn, "select * from t1")
 	}
--- a/go/vt/vtgate/vtgate.go
+++ b/go/vt/vtgate/vtgate.go
@@ -80,9 +80,10 @@ var (
 	warnPayloadSize    = flag.Int("warn_payload_size", 0, "The warning threshold for query payloads in bytes. A payload greater than this threshold will cause the VtGateWarnings.WarnPayloadSizeExceeded counter to be incremented.")
 
 	// Put set-passthrough under a flag.
-	sysVarSetEnabled = flag.Bool("enable_system_settings", true, "This will enable the system settings to be changed per session at the database connection level")
-	setVarEnabled    = flag.Bool("enable_set_var", true, "This will enable the use of MySQL's SET_VAR query hint for certain system variables instead of using reserved connections")
-	plannerVersion   = flag.String("planner_version", "gen4", "Sets the default planner to use when the session has not changed it. Valid values are: V3, Gen4, Gen4Greedy and Gen4Fallback. Gen4Fallback tries the gen4 planner and falls back to the V3 planner if the gen4 fails.")
+	sysVarSetEnabled         = flag.Bool("enable_system_settings", true, "This will enable the system settings to be changed per session at the database connection level")
+	setVarEnabled            = flag.Bool("enable_set_var", true, "This will enable the use of MySQL's SET_VAR query hint for certain system variables instead of using reserved connections")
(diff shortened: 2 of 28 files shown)


frouioui (approved):

go/vt/vtgate/vtgate.go
```diff
@@ -80,9 +80,10 @@ var (
...
-	plannerVersion   = flag.String("planner_version", "gen4", "Sets the default planner to use when the session has not changed it. Valid values are: V3, Gen4, Gen4Greedy and Gen4Fallback. Gen4Fallback tries the gen4 planner and falls back to the V3 planner if the gen4 fails.")
+	sysVarSetEnabled         = flag.Bool("enable_system_settings", true, "This will enable the system settings to be changed per session at the database connection level")
+	setVarEnabled            = flag.Bool("enable_set_var", true, "This will enable the use of MySQL's SET_VAR query hint for certain system variables instead of using reserved connections")
+	plannerVersion           = flag.String("planner-version", "gen4", "Sets the default planner to use when the session has not changed it. Valid values are: V3, Gen4, Gen4Greedy and Gen4Fallback. Gen4Fallback tries the gen4 planner and falls back to the V3 planner if the gen4 fails.")
+	plannerVersionDeprecated = flag.String("planner_version", "", "Deprecated flag. Use planner_version instead")
```
frouioui, line 86: we need a default here too

systay: why?

frouioui: from what i understand `plannerVersionDeprecated` will be set to `""` and thus be different than `plannerVersion`

frouioui: i think we should actually check for `len(plannerVersionDeprecated)` instead of `plannerVersionDeprecated != nil`

rohit-nayak-ps (dismissed):
TestHelpOutput  is failing and needs to be fixed for the new flags. @systay

rohit-nayak-ps:
I presume this will also fix the current panic in `main`?
```
rohit@rohit-ubuntu:~/vitess-ps$ vtexplain
vtexplain flag redefined: planner_version
panic: vtexplain flag redefined: planner_version

goroutine 1 [running]:
flag.(*FlagSet).Var(0xc000116120, {0x1872fe8, 0xc000486310}, {0x13e0f63, 0xf}, {0x1437603, 0x66})
	flag/flag.go:879 +0x2f4
flag.(*FlagSet).StringVar(...)
	flag/flag.go:762
flag.(*FlagSet).String(0x58?, {0x13e0f63, 0xf}, {0x13d207f, 0x4}, {0x1437603, 0x66})
	flag/flag.go:775 +0xac
flag.String(...)
	flag/flag.go:782
main.init()
	vitess.io/vitess/go/cmd/vtexplain/vtexplain.go:54 +0x4aa
```

ajm188:
No, because it's defined twice (still) [here](https://github.com/vitessio/vitess/pull/10453/files#diff-386dafefbaf1d87f35fb8c081ef12aed4af6aeb89db721806b21dacc8c49a438R85-R86). We should remove the one in `cmd/vtexplain` since it imports that package and will still show up in the help (post v14, we'll be able to tidy this up in a more sane way)

go/test/endtoend/tabletmanager/qps_test.go
```diff
@@ -60,9 +60,7 @@ func TestQPS(t *testing.T) {
 	//       after that we'll see 0.0 QPS rates again. If this becomes actually
 	//       flaky, we need to read continuously in a separate thread.
 
-	n := 0
-	for n < 15 {
-		n++
+	for n := 0; n < 15; n++ {
```
ajm188, line 63: so strange

ajm188 (approved):
hero

deepthi:
@ajm188 we have support for both forms for some flags like this:
```
	flagutil.DualFormatBoolVar(&enableConsolidator, "enable_consolidator", true, "This option enables the query consolidator.")
```
https://github.com/vitessio/vitess/blob/main/go/vt/vttablet/tabletserver/tabletenv/config.go#L158
Is that an option we should consider as part of flag migration? Then we won't care whether users use `-` or `_` and avoid a whole class of issues.

ajm188:
So, I'm actually against that convention, because it increases the matrix of possible ways you can "configure" vitess (e.g. "oh you said --some-flag but i'm using --some_flag, does that matter?" or "is --query_server-some-flag allowed? what about --query-server_some-flag? --query_server_some_flag?") better for there to be exactly one way to do things; avoids unnecessary complexity / opportunities for confusion/miscommunications

PR #10455, opened 2022-06-07
enable schema tracking by default

harshit-gangal:
## Description
This PR enables schema tracking by default. It's mostly about changing default settings and then updating integration tests that start failing.

## Checklist


go/test/endtoend/vreplication/resharding_workflows_v2_test.go
```diff
@@ -39,7 +39,7 @@ const (
 	reverseKsWorkflow = sourceKs + "." + workflowName + "_reverse"
 	tablesToMove      = "customer"
 	defaultCellName   = "zone1"
-	readQuery         = "select * from customer"
+	readQuery         = "select cid from customer"
```
deepthi, line 42: Why is this change required?

harshit-gangal: the test expectation was not matching as the actual query getting executed is the expanded one with cid instead of `star`

This is what I can understand from Rohit's commit message.

systay (approved):

PR #10458, opened 2022-06-07
Post release `v14.0.0-RC1` steps

frouioui:
## Description

This PR is a follow-up to the `v14.0.0-RC1` release. The different version constants have been incremented to `v15.0.0-SNAPSHOT` and the `v14.0.0` releases notes are added to the codebase.

### DIFF
--- a/doc/releasenotes/14_0_0_release_notes.md
+++ b/doc/releasenotes/14_0_0_release_notes.md
@@ -1,0 +1,650 @@
+# Release of Vitess v14.0.0-RC1
+
+## Major Changes
+
+### Command-line syntax deprecations
+
+Vitess has begun a transition to a new library for CLI flag parsing.
+In order to facilitate a smooth transition, certain syntaxes that will not be supported in the future now issue deprecation warnings when used.
+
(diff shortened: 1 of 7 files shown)


doc/releasenotes/14_0_0_release_notes.md
```diff
@@ -0,0 +1,650 @@
+# Release of Vitess v14.0.0-RC1
```
mattlord, line 1: I think we only did this part for GA in the past? https://github.com/vitessio/vitess/commits/main/doc/releasenotes/13_0_0_release_notes.md

I'm not sure though. And also don't feel too strongly about whether or not the past should guide the present/future. :-) 

If we leave this in place then we'll just of course have to be sure to update/replace it for the GA release.

mattlord (approved):
Not sure about the release notes part, but will leave that up to you and us collectively as I don't feel strongly about it either way.

doc/releasenotes/14_0_0_release_notes.md
frouioui: Right, this will be replaced during GA

ajm188 (approved):

PR #10460, opened 2022-06-07
[vtadmin] Rename ERS/PRS pools+flags properly

ajm188:
## Description

I missed this when doing #10404 🙈; ideally this lands in 14.0

## Checklist


deepthi (approved):

deepthi:
> ideally this lands in 14.0

You just need to create another PR 😄 
General rule for rc1 -> GA timeframe is that any and all bug fixes are allowed. Plus minor changes / cleanups / bug fixes on top of features that are new in the release are certainly allowed.

Issue #6612, opened 2020-08-23
RFC: Orchestrator as a native Vitess component


sougou:
## Introduction
With #6582, we are starting on a Vitess-specific fork of the Orchestrator that was originally created by @shlomi-noach.

Previously, Vitess provided an integration to loosely cooperate with an install of Orchestrator managing automated failover for the MySQL instances underlying Vitess. However, users often encountered corner cases where the two sides could make conflicting changes, potentially leaving the system in an inconsistent state.

We intend to resolve these issues and provide a smooth and unified experience to the end-user.

## Approach
Many approaches were considered:
1. Build a Vitess-native failover mechanism.
2. Copy portions of the Orchestrator code into Vitess.
3. Extend Orchestrator to allow tighter integration and cooperation with Vitess.
4. Modify Orchestrator to become a first-class Vitess component.

Of the above approaches, we have decided on option 4. We reached this conclusion after studying the Orchestrator code, and realizing that it is written with the intent to allow customization.

Architecturally, the Orchestrator is closest in function to the work performed by vtctld and has some overlapping functionality. For example, `stop-replica` vs `StopReplica`, `graceful-master-takeover` vs `PlannedReparentShard`.

On the other hand: For massive clusters with >10K nodes, it may be better to divide the Orchestrator's responsibility into specific keyspaces or shards. In this situation, it starts to deviate from the vtctld model. This requires further brainstorming.

### To fork or not to fork
We debated extending the Orchestrator to accommodate Vitess into the existing code base, vs. forking the code. The conclusion was to fork due to the following reasons:
* Model: Orchestrator’s failover code depends on the discovered replication graph that connected replicas to their master. This model is substantially different from Vitess where cluster membership is strongly dictated by the keyspace-shard an instance belongs to.
* Responsibility: Orchestrator is primarily responsible for performing failovers. However, this separation gave rise to many corner cases, especially when Vitess would wire-up replication using a view of the world that differed from what Orchestrator saw. For this integration to work smoothly, the Orchestrator needs to take responsibility for all functions of the cluster.
* Opinions: Orchestrator has used a hands-free approach with the intent to accommodate with the desires of how administrators would like to configure their MySQL instances. In Vitess, there is an expectation that a cluster should be mostly self-managed, with certain critical state transitions occurring manually. So, many of the flexibilities that Orchestrator accommodates are not allowed in Vitess.
* Velocity: If we forked, we can move faster because there is no fear of breaking use cases that are not supported by Vitess.

[...]

PR #10464, opened 2022-06-08
Adds RPCs to vttablet that vtorc requires

GuptaManan100:
## Description

This PR adds 2 new Tablet manager client RPCs that VTOrc requires -
- ResetReplicationParameters
  -  This RPC stops the replication (if running) and resets the replication parameters like source host, port, etc from the server. It does not however clear out the executed gtid set or retrieved gtid set. In MySQL equivalent this function does `STOP REPLICA; RESET REPLICA ALL`
  - This RPC is supposed to be used in fixPrimaryHasPrimary in VTOrc to reset the replication parameters on the primary tablet
- FullStatus
  - This RPC collects information from the server including replication information, semi-sync settings, server settings among others. The full list of outputs in the full status are - 
    - ServerID: Server ID configured :- `select @@global.server_id`
    - ServerUUID: Server UUID configured/assigned :- `select @@global.server_uuid`
    - ReplicationStatus: Replication status of the server :- `show replica status`
    - PrimaryStatus: Binlog position and executed gtid set :- `show master status`
    - GTIDPurged: purged gtid set :- `select @@gtid_purged`
    - Version: Server version in the format `majorVersion.minorVersion.patchRelease`  
    - VersionComment: Version comment in the server :- `select @@global.version_comment`
    - ReadOnly: Whether the server is read only :- `show variables like 'read_only'`
    - BinLogFormat: format of the binlog of the server :- `select @@global.binlog_format`
    - BinLogRowImage: the way to store changes to a row in the binlog :- `select @@global.binlog_row_image`
    - LogBinEnabled: Whether logging in the binlog is enabled :- `select @@global.log_bin`
    - LogReplicaUpdates: Whether the server will also add the updates acquired via replication to the binlog :- `select @@global.log_slave_updates`
    - GTIDMode: `select @@global.gtid_mode`
    - SemiSyncPrimaryEnabled: Whether the primary side semi-sync plugin is enabled :- `show global variables like 'rpl_semi_sync_master_enabled'`
    - SemiSyncReplicaEnabled: Whether the replica side semi-sync plugin is enabled :- `show global variables like 'rpl_semi_sync_slave_enabled'`
    - SemiSyncPrimaryStatus: Status of the primary side semi-sync :- `show status like 'Rpl_semi_sync_master_status'`
    - SemiSyncReplicaStatus: Status of the replica side semi-sync :- `show status like 'Rpl_semi_sync_slave_status'`
    - SemiSyncPrimaryClients: Number of semi-sync clients of the server :- `show status like 'semi_sync_primary_clients'`
    - SemiSyncPrimaryTimeout: Amount of time to wait for semi-sync ack before reverting to async replication :- `show status like 'rpl_semi_sync_master_timeout`
    - SemiSyncWaitForReplicaCount: Number of semi-sync acks required by the server to accept a write :- `show status like 'rpl_semi_sync_master_wait_for_slave_count`, 
  - This RPC is meant to be used in `ReadTopologyInstanceBufferable` to collect information of the MySQL server without querying it directly

Apart from these 2 RPCs, this PR also augments the ReplicationStatus struct and adds the following new fields to it -
 - RelayLogFilePosition - Position of the relay log of the server.
 - SourceUser - The source user for the replica.
 - SQLDelay - The configured delay that the sql thread is supposed to maintain.
 - AutoPosition - Whether GTID auto-positioning is in use
 - UsingGTID - If the MariaDB server is using GTIDs for replication
 - HasReplicationFilters - If there are any replication filters configured
 - SSLAllowed - If the server allows ssl
 - ReplicationLagUnknown - Set true if the replication lag is unknown and parsing Seconds_Behind_Primary failed
An existing field `FileRelayLogPosition` has been renamed to `RelayLogSourceBinLogEquivalentPosition` since it is a more apt variable name. The previous variable name would have been too confusing after introducing `RelayLogFilePosition`
A test that checks that the rpc changes are backward compatible has also been added.

## Related Issue(s)

- #6612 

## Checklist


GuptaManan100:
https://github.com/vitessio/vitess/pull/10464 uses these introduced RFCs and verifies that they serve their purpose

go/mysql/flavor.go
```diff
@@ -424,6 +469,8 @@ func parseReplicationStatus(fields map[string]string) ReplicationStatus {
 	}
 	parseUint, _ = strconv.ParseUint(fields["Master_Server_Id"], 10, 0)
 	status.SourceServerID = uint(parseUint)
+	parseUint, _ = strconv.ParseUint(fields["SQL_Delay"], 10, 0)
+	status.SQLDelay = uint(parseUint)
```
deepthi, line 473: How are we using SQL_Delay? Can we assume it is never set?

go/mysql/replication_status.go
```diff
@@ -25,17 +25,28 @@ import (
...
+	// it is same as FileRelayLogPosition
+	RelayLogPosition Position
+	// FilePosition stores the position of the source tablets binary log
+	// upto which the SQL thread of the replica has run.
+	FilePosition Position
+	// FileRelayLogPosition stores the position of the source tablets binary log
+	// upto which the IO thread has read and added to the relay log
+	FileRelayLogPosition Position
+	// RelayLogFilePosition stores the position in the relay log file
+	RelayLogFilePosition  Position
```
deepthi, line 49: These two field names are confusing. It is going to be too easy to make a mistake while making changes in future.

go/mysql/flavor.go
```diff
@@ -75,6 +75,15 @@ type flavor interface {
...
 
+	// purgedGTIDSet returns the purged GTIDSet of a server.
+	purgedGTIDSet(c *Conn) (GTIDSet, error)
+
+	// gtidMode returns the gtid mode of a server.
+	gtidMode(c *Conn) (string, error)
+
+	// serverUUID returns the UUID of a server.
+	serverUUID(c *Conn) (string, error)
+
```
deepthi, line 86: technically these are MySQL-specific terms which breaks the "abstraction" of the interface. At this point, it doesn't really matter, since for all practical purposes we can only support MySQL/Percona.

go/mysql/flavor.go
deepthi: Same thing for Replication filters. Do we expect those to be ever set on Vitess-managed MySQL databases?
And if Vitess is not managing them, then we don't care anyway.

go/vt/mysqlctl/replication.go
```diff
@@ -188,6 +189,33 @@ func (mysqld *Mysqld) GetMysqlPort() (int32, error) {
 	return int32(utemp), nil
 }
 
+// GetServerID returns mysql server id
+func (mysqld *Mysqld) GetServerID(ctx context.Context) (uint32, error) {
+	qr, err := mysqld.FetchSuperQuery(ctx, "select @@global.server_id")
+	if err != nil {
```
deepthi, line 195: Why is this query here instead of being accessed through the `flavor` interface?

go/mysql/flavor.go
GuptaManan100: We do not expect them to be set, but if they are then it could lead to unexpected failures in other components of Vitess. For example, if someone sets the SQL delay to 1 minute (for whatever imaginable reason) on a tablet, the ERS will keep failing over and over on trying to wait for the tablet to apply its relay log. It just wouldn't until the timeout (30 seconds) expires. Orchestrator used to track them and showed them on the UI.

It is easier to have the fields if we need them and remove them if we don't instead of not having them and needing them later on. So for now, I have elected to add all the fields that VTOrc tracks.

Also, we might want to add a failure scenario where we set SQL delay to 0 if not already. At the very least, ERS can benefit from it, ignoring the tablets that have some SQL delay setup, since they don't seem like viable primary candidates

go/mysql/replication_status.go
GuptaManan100: Yes, renaming the field `FileRelayLogPosition` to `RelayLogSourceBinLogEquivalentPosition`

go/vt/mysqlctl/replication.go
GuptaManan100: This query is the same for all the `flavor` interface implementations. Adding a function there would have meant having the same function for all the implementations. So I have chosen to run this without using the interface. I have followed a similar architecture to `GetMysqlPort` which too tries to run a command that is the same for all the different MySQL - equivalent variants

go/mysql/flavor.go
deepthi: OK, let's err on the side of having more fields than fewer fields.

go/mysql/replication_status.go
deepthi: 😆 
Guess that's the best we can do to avoid confusion.
I don't know whether binlog is one word - so Binlog vs BinLog
_goes to grep vitessio/vitess_
OK, we use `Binlog` almost everywhere except in a few test files. I know this is super-nitpicky, but could you please change it?

go/vt/mysqlctl/replication.go
deepthi: Sounds good.

deepthi (approved):
One nit, the rest LGTM.
You may merge once that is addressed.

go/mysql/replication_status.go
GuptaManan100: Okay 👍

PR #10465, opened 2022-06-08
fix: scalar aggregation engine primitive

harshit-gangal:
## Description

This PR fixes two bugs in scalar aggregation engine primitive
1. Scalar aggregation was returning wrong output for empty row on count aggregation
2. Scalar aggregation in streaming mode was not handling empty row and also final convert logic.

This was caused when we moved from using AggregateSum over AggregateCount as opcode for count(*). As the actual operation at the engine primitive level was sum of count, this caused a negative effect of returning `NULL` for empty row instead of `0`.

## Checklist

-   [X] "Backport me!" label has been added if this change should be backported -> 14.0


go/vt/vtgate/engine/ordered_aggregate.go
```diff
@@ -127,18 +131,23 @@ func (ap *AggregateParams) String() string {
...
+	dispOrigOp := ""
+	if ap.OrigOpcode != AggregateUnassigned && ap.OrigOpcode != ap.Opcode {
+		dispOrigOp = "_" + ap.OrigOpcode.String()
+	}
 	if ap.Alias != "" {
-		return fmt.Sprintf("%s(%s) AS %s", ap.Opcode.String(), keyCol, ap.Alias)
+		return fmt.Sprintf("%s%s(%s) AS %s", ap.Opcode.String(), dispOrigOp, keyCol, ap.Alias)
 	}
-	return fmt.Sprintf("%s(%s)", ap.Opcode.String(), keyCol)
+	return fmt.Sprintf("%s%s(%s)", ap.Opcode.String(), dispOrigOp, keyCol)
```
frouioui, line 141: I think the plans would be more readable if we had a field that contains the original op code to the description instead of suffixing the original op code to the current op code. I don't have strong opinions though, the current change works well too imo.

frouioui (approved):

PR #10472, opened 2022-06-08
schemadiff: `ColumnRenameStrategy` in DiffHints

shlomi-noach:
## Description

This PR is a first iteration for a heuristic detection of column renames.

A new diff hint `ColumnRenameStrategy` can take the values:

- `ColumnRenameAssumeDifferent` (default); for backwards compatibility, `schemadiff` does not try to resolve renames. It treats a column rename like a `DROP COLUMN` + `ADD COLUMN`
- `ColumnRenameHeuristicStatement`: `schemadiff` attempts to generate a `RENAME COLUMN` statement for what it _suspects_ to be a column rename.

Right now the heuristic is very basic; the problem is complex and in some scenarios it's impossible to tell which columns are renamed. Pasting the code comment from this PR:

```go
		// The check is heuristic, and looks like this:
		// We consider a column renamed iff:
		// - the DROP and ADD column definitions are identical other than the column name, and
		// - the DROPped and ADDded column are both FIRST, or they come AFTER the same column, and
		// - the DROPped and ADDded column are both last, or they come before the same column
		// This v1 chcek therefore cannot handle a case where two successive columns are renamed.
		// the problem is complex, and with successive renamed, or drops and adds, it can be
		// impossible to tell apart different scenarios.
		// At any case, once we heuristically decide that we found a RENAME, we cancel the DROP,
		// cancel the ADD, and inject a RENAME in place of both.
```

## Related Issue(s)

#10203 

## Checklist


go/vt/schemadiff/errors.go
```diff
@@ -239,3 +239,13 @@ func (e *InvalidColumnInForeignKeyConstraintError) Error() string {
...
+
+type ColumnRenamedError struct {
+	Table  string
+	Column string
+}
+
+func (e *ColumnRenamedError) Error() string {
+	return fmt.Sprintf("column %s renamed in table %s",
+		sqlescape.EscapeID(e.Column), sqlescape.EscapeID(e.Table))
+}
```
dbussink, line 251: @shlomi-noach Are these errors used anywhere?

shlomi-noach: I raced you and won, there are removed

go/vt/schemadiff/table.go
```diff
@@ -1284,9 +1297,69 @@ func (c *CreateTableEntity) diffColumns(alterTable *sqlparser.AlterTable,
 				}
 			}
 			expectAppendIndex++
-			alterTable.AlterOptions = append(alterTable.AlterOptions, addColumn)
+			addColumns = append(addColumns, addColumn)
+		}
+	}
+	renameColumns := []*sqlparser.RenameColumn{}
+	if hints.ColumnRenameStrategy == ColumnRenameHeuristicStatement {
```
dbussink, line 1304: Would it be nicer to extract the below logic into a separate method for easier reading etc?

shlomi-noach:
Ignoring flaky `Cluster (onlineddl_vrepl_stress_suite) mysql80`

PR #10478, opened 2022-06-09
Add parsing support for performance schema functions

K-Kumar-01:
## Description

Add parsing support for Performance Schema functions. More information [here](https://dev.mysql.com/doc/refman/8.0/en/performance-schema-functions.html).

## Related Issue(s)

#8604 
#4099 ( These are introduced in version 8.0 or later)

## Checklist


GuptaManan100 (approved):
LGTM!

Issue #10378, opened 2022-05-28
Task: delete unused sharding_column_name and sharding_column_type


deepthi:
### Question

Keyspace protobuf definition still contains sharding_column_name and sharding_column_type. These are from v2 resharding which predates vindex based resharding. These fields can now be removed.

PR #10488, opened 2022-06-09
deprecation of CreateKeyspace flags

deepthi:
## Description
Mark sharding-column-name and sharding-column-type for CreateKeyspace command as deprecated. They will be removed in v15.

## Related Issue(s)
#10378

## Checklist


deepthi:
Technically we should not be deprecating after RC, but this functionality is almost certainly used by nobody at this point.

ajm188 (approved):

Issue #6518, opened 2020-07-31
consistent_lookup_unique vindex reports unrelated error


ankitthakwani:
#### Overview of the Issue
In a multi shard keyspace with a vtgate flag -transaction_mode set to SINGLE, a transaction which violates the global unique constraint of consistent_lookup_unique vindex fails with unrelated error message.

#### Reproduction Steps

Steps to reproduce this issue, example:

1. Deploy the following `vschema`:

    ```javascript
    {
      "sharded": true,
      "vindexes": {
        "unicode_loose_md5_vdx": {
          "type": "unicode_loose_md5"
        },
        "uniqueConstraint_vdx": {
          "type": "consistent_lookup_unique",
          "params": {
            "table": "uniqueConstraint_vdx",
            "from": "unique_constraint",
            "to": "keyspace_id",
            "autocommit": "true"
          },
          "owner": "txn_unique_constraints"      
        }
      },
      "tables": {
        "txn_unique_constraints": {
          "column_vindexes": [
            {
              "column": "txn_id",
              "name": "unicode_loose_md5_vdx"
            },
            {
              "column": "unique_constraint",
              "name": "uniqueConstraint_vdx"
            }
          ]
        },    
        "uniqueConstraint_vdx": {
          "column_vindexes": [
            {
              "column": "unique_constraint",
              "name": "unicode_loose_md5_vdx"
            }
          ]      
        }
      }
    }
    ```

2. Deploy the following `schema`:

    ```sql
    DROP TABLE IF EXISTS txn_unique_constraints;
    CREATE TABLE `txn_unique_constraints` (
      `id`                varchar(100) NOT NULL ,
      `txn_id`            varchar(50) NOT NULL,
      `unique_constraint` varchar(100) NOT NULL,
      `created_on`        timestamp(3) NOT NULL DEFAULT CURRENT_TIMESTAMP(3),
      PRIMARY KEY (`id`),
      UNIQUE KEY `idx_txn_unique_constraint` (`unique_constraint`),
      KEY `idx_txn_unique_constraints_txn_id` (`txn_id`),
      KEY `idx_txn_unique_constraints_created_on` (`created_on`)
    ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
    
    DROP TABLE IF EXISTS uniqueConstraint_vdx;
    CREATE TABLE uniqueConstraint_vdx(
      `unique_constraint` VARCHAR(50) NOT NULL,
[...]

PR #10493, opened 2022-06-13
Fix: handle all cases for consistent lookup unique on single transaction mode

harshit-gangal:
## Description

This PR enhances the consistent_lookup implementation around dangling rows when the transaction mode is set to `SINGLE`.
Now, it can handle more cases and allow the transaction to complete without throwing multi-db error, until an actual dml happens 2 or more shards.

## Related Issue(s)

- Closes https://github.com/vitessio/vitess/issues/6518

## Checklist


go/test/endtoend/vtgate/transaction/single/main_test.go
```diff
@@ -0,0 +1,225 @@
+/*
+Copyright 2019 The Vitess Authors.
```
frouioui, line 2: nit: 2022

go/test/endtoend/vtgate/transaction/single/main_test.go
```diff
@@ -0,0 +1,225 @@
...
+			Host: clusterInstance.Hostname,
+			Port: clusterInstance.VtgateMySQLPort,
+		}
+		return m.Run()
+	}()
+	os.Exit(exitCode)
+}
+
+func TestSingleOneWay(t *testing.T) {
+	conn, err := mysql.Connect(context.Background(), &vtParams)
```
frouioui, line 89: I think this test and the one below should be tested against both Vitess and mysql, we can use the `utils` package for that.

go/vt/vtgate/safe_session.go
```diff
@@ -294,10 +299,21 @@ func (session *SafeSession) Find(keyspace, shard string, tabletType topodatapb.T
 	}
 	for _, shardSession := range sessions {
 		if keyspace == shardSession.Target.Keyspace && tabletType == shardSession.Target.TabletType && shard == shardSession.Target.Shard {
-			return shardSession.TransactionId, shardSession.ReservedId, shardSession.TabletAlias
+			if getTxMode() == vtgatepb.TransactionMode_SINGLE && shardSession.VindexOnly && !session.queryFromVindex {
+				count := actualNoOfShardSession(session.ShardSessions)
+				// If the count of shard session which are non vindex only is greater than 0, then it is a
```
frouioui, line 304: nit: `then it is an error`

go/vt/vtgate/vcursor_impl.go
```diff
@@ -556,9 +556,18 @@ func (vc *vcursorImpl) ExecuteKeyspaceID(keyspace string, ksid []byte, query str
 		Sql:           query,
 		BindVariables: bindVars,
 	}}
+	if !rollbackOnError {
+		vc.safeSession.queryFromVindex = true
+		defer func() {
+			vc.safeSession.queryFromVindex = false
+		}()
+	}
```
frouioui, line 564: I don't understand this change. Maybe we should add a comment that explains why we do this

go/vt/vtgate/vcursor_impl.go
```diff
@@ -556,9 +556,18 @@ func (vc *vcursorImpl) ExecuteKeyspaceID(keyspace string, ksid []byte, query str
...
 	qr, errs := vc.ExecuteMultiShard(rss, queries, rollbackOnError, autocommit)
-
-	return qr, vterrors.Aggregate(errs)
+	err = vterrors.Aggregate(errs)
+	if err != nil {
+		return nil, vterrors.Wrap(err, "I know I know")
```
frouioui, line 568: We should change the error message here

frouioui: Or directly return the error that `vterrors.Aggregate` returns

frouioui (approved):

go/vt/vtgate/vcursor_impl.go
harshit-gangal: this needs to go away

go/vt/vtgate/safe_session.go
```diff
@@ -282,7 +287,7 @@ func (session *SafeSession) InTransaction() bool {
 }
 
 // Find returns the transactionId and tabletAlias, if any, for a session
-func (session *SafeSession) Find(keyspace, shard string, tabletType topodatapb.TabletType) (transactionID int64, reservedID int64, alias *topodatapb.TabletAlias) {
+func (session *SafeSession) Find(keyspace, shard string, tabletType topodatapb.TabletType) (int64, int64, *topodatapb.TabletAlias, error) {
```
systay, line 290: I didn't find it intuitive that a method with the name `Find` actually changes the state of the shard sessions. Maybe we should rename it?

systay (approved):

PR #10499, opened 2022-06-14
User defined and sys variables

GuptaManan100:
## Checklist


go/vt/sqlparser/sql.y
```diff
@@ -5787,21 +5783,25 @@ UTC_DATE func_paren_opt
...
   {
     $$ = &JSONExtractExpr{JSONDoc: $3, PathList: $5}
   }
-| JSON_KEYS openb expression json_path_param_list_opt closeb
+| JSON_KEYS openb expression closeb
   {
-    $$ = &JSONKeysExpr{JSONDoc: $3, PathList: $4}
+    $$ = &JSONKeysExpr{JSONDoc: $3}
+  }
+| JSON_KEYS openb expression ',' expression closeb
```
systay, line 5802: is this a bad way of doing it, @GuptaManan100? I have two cases here with `JSON_KEYS`

GuptaManan100: No, this is fine as long as this happens only for one of these functions. In general I try to just reduce the number of total rules that are there in the grammar. So making `expression_comma_opt` will reduce one rule but add one more (its own definition), so it's not worth it.

go/vt/sqlparser/testdata/select_cases.txt
```diff
@@ -9260,7 +9260,7 @@ INPUT
 select @@local.profiling;
 END
 OUTPUT
-select @@local.profiling from dual
+select @@`local.profiling` from dual
```
GuptaManan100, line 9263: This is incorrect. Just checked on MySQL - 
```mysql
mysql [localhost:8026] {msandbox} (test) > select @@local.profiling;
+-------------------+
| @@local.profiling |
+-------------------+
|                 0 |
+-------------------+

mysql [localhost:8026] {msandbox} (test) > select @@`local.profiling`;
ERROR 1193 (HY000): Unknown system variable 'local.profiling'
```

systay: nice catch

go/vt/sqlparser/sql.y
```diff
@@ -6950,26 +6958,35 @@ set_list:
   }
 
 set_expression:
-  reserved_sql_id '=' ON
+  set_variable '=' ON
```
GuptaManan100, line 6961: set_exprsesion, apparently also used to take in non-reserved and reserved keywords. It does not now. Is this change intentional?

systay (changes requested):
1. Need to handle `@@local.xyz` as `@@session.xyz`
2. If we get a scope that is unknown, we should fail and not accept it as an identifier - `@@foo.var` is an illegal variable name.
3. Using illegal scopes should not be syntax errors

go/vt/sqlparser/sql.y
systay: no, and probably wrong.

go/vt/sqlparser/ast_rewriting.go
```diff
@@ -354,13 +350,18 @@ func (er *astRewriter) rewrite(cursor *Cursor) bool {
 		}
 	case *FuncExpr:
 		er.funcRewrite(cursor, node)
-	case *ColName:
-		switch node.Name.at {
-		case SingleAt:
+	case *Variable:
+		if v, isSet := cursor.Parent().(*SetExpr); isSet && v.Var == node {
+			break
+		}
```
GuptaManan100, line 356: A comment here would be nice

go/vt/sqlparser/ast_rewriting.go
```diff
@@ -354,13 +350,18 @@ func (er *astRewriter) rewrite(cursor *Cursor) bool {
...
+		switch node.Scope {
+		case VariableScope:
 			er.udvRewrite(cursor, node)
-		case DoubleAt:
+		case GlobalScope, SessionScope:
 			er.sysVarRewrite(cursor, node)
 		}
+	case *SetExpr:
+
```
GuptaManan100, line 364: If the case is empty, do we even need it?

GuptaManan100 (commented):

go/vt/sqlparser/constants.go
```diff
@@ -446,14 +444,26 @@ const (
...
-	VitessMetadataScope
-	VariableScope
-	LocalScope
+	NoScope             Scope = iota
+	SessionScope              // [SESSION | @@SESSION. | @@] This is the default if no scope is given
+	GlobalScope               // {GLOBAL | @@GLOBAL.} system_var_name
+	VitessMetadataScope       // @@vitess_metadata.system_var_name
+	VariableScope             // @var_name   This is used for user defined variables. Not really a scope, but it works.
+	PersistSysScope           // {PERSIST_ONLY | @@PERSIST_ONLY.} system_var_name
+	PersistOnlySysScope       // {PERSIST_ONLY | @@PERSIST_ONLY.} system_var_name
```
frouioui, line 466: maybe adding a link to the MySQL doc page would help

go/vt/sqlparser/ast_funcs.go
```diff
@@ -737,41 +776,41 @@ func (node *ColIdent) UnmarshalJSON(b []byte) error {
...
-func NewTableIdent(str string) TableIdent {
+// NewIdentifierCS creates a new IdentifierCS.
+func NewIdentifierCS(str string) IdentifierCS {
 	// Use StringClone on the table name to ensure it is not pinned to the
 	// underlying query string that has been generated by the parser. This
 	// could lead to a significant increase in memory usage when the table
 	// name comes from a large query.
-	return TableIdent{v: strings.Clone(str)}
+	return IdentifierCS{v: strings.Clone(str)}
 }
```
harshit-gangal, line 786: here as well

frouioui (commented):
Everything looks fine to me besides what has already been commented.

go/vt/sqlparser/parse_test.go
```diff
@@ -1066,65 +1069,85 @@ var (
...
 	}, {
-		input: "set transaction isolation level repeatable read",
+		input:  "set transaction isolation level repeatable read",
+		output: "set session transaction isolation level repeatable read",
 	}, {
 		input: "set global transaction isolation level repeatable read",
 	}, {
-		input: "set transaction isolation level repeatable read",
+		input:  "set transaction isolation level repeatable read",
+		output: "set session transaction isolation level repeatable read",
```
harshit-gangal, line 1094: this has different meaning.
the input statement means only for next transaction and not at session level.

harshit-gangal: but the parser output makes it session level

systay: nice catch!

systay:
> 1. Need to handle `@@local.xyz` as `@@session.xyz`
> 2. If we get a scope that is unknown, we should fail and not accept it as an identifier - `@@foo.var` is an illegal variable name.
> 3. Using illegal scopes should not be syntax errors

1. ✔️ 
2. We don't need to fail at the parser level. Just need to fail later on. ✔️ 
3. As a consequence of (2) ,bad scope names will not turn into syntax errors ✔️

go/vt/sqlparser/ast_funcs.go
```diff
@@ -601,27 +601,27 @@ func (node *FuncExpr) IsAggregate() bool {
 	return Aggregates[node.Name.Lowered()]
 }
 
-// NewColIdent makes a new ColIdent.
-func NewColIdent(str string) ColIdent {
-	return ColIdent{
+// NewIdentifierCI makes a new IdentifierCI.
+func NewIdentifierCI(str string) IdentifierCI {
```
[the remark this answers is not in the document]
systay: I disagree. I see these methods as constructors for the type, and constructors usually are names using `New<typeName>`

systay:
Re-opened in my name #10547

PR #10502, opened 2022-06-14
[vtctldclient] Update CLI docs for usages, flags, and aliases

ajm188:
## Description

This is a final pass/cleanup before 14.0, as part of https://github.com/vitessio/website/pull/1054. I'll regenerate the website docs after addressing any feedback on this PR and backporting to `release-14.0`.

Some things in particular to call out:
1. I added command aliases for PRS/ERS so users can also specify `EmergencyFailoverShard`, etc, if they wish.
2. The `Short` usage for `CreateKeyspace` still referenced the deprecated ShardingColumn* options, so I removed them from the example to further discourage users from setting them.

## Checklist


deepthi (approved):
> I added command aliases for PRS/ERS so users can also specify EmergencyFailoverShard, etc, if they wish.

This is too consequential a change to get in without further discussion. Let's leave it out of this PR.

> The Short usage for CreateKeyspace still referenced the deprecated ShardingColumn* options, so I removed them from the example to further discourage users from setting them.

Good catch 👍 

The rest LGTM

PR #10503, opened 2022-06-14
Fix upgrade-downgrade build on `release-13.0`

frouioui:
## Description

This Pull Request changes the Go version used in the upgrade downgrade tests. It was set to `go1.17.11`, with this change it will use `go1.18.3` which is compatible with any release of Vitess.

While fixing this problem, I have realized that upgrade downgrade tests on `v13` are testing the current commit and the latest version of vitess, which is in this case: `release-14.0` ; instead it should use the "previous" release: `release-12.0`. I will open a follow-up issue.

## Checklist

### DIFF
--- a/.github/workflows/upgrade_downgrade_test_backups_e2e.yml
+++ b/.github/workflows/upgrade_downgrade_test_backups_e2e.yml
@@ -59,7 +59,7 @@ jobs:
     - name: Set up Go
       uses: actions/setup-go@v2
       with:
-        go-version: 1.17.11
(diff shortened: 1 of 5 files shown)


deepthi (approved):

deepthi:
Unfortunately most of the upgrade/downgrade tests are failing on this PR.

deepthi:
`onlineddl_vrepl_stress_suite` has known flakiness. Merging now that everything else is passing.

PR #10506, opened 2022-06-14
feat: don't stop if compilation errors are happening on the generated files

systay:
## Description
Changes so `make codegen` doesn't stop because of compilation errors on the files with static methods (`ast_rewrite.go`, `ast_visit.go`, `ast_equals.go` and `ast_clone.go`). Since these files will be overwritten anyway, if all the compilation errors are on these, we'll just ignore the error.

Unfortunately, this does not work for the `ast_format_fast` generator. It produces methods that are interface implementations, and errors here are more difficult to ignore.

## Checklist

### DIFF
--- a/go/tools/asthelpergen/asthelpergen.go
+++ b/go/tools/asthelpergen/asthelpergen.go
@@ -27,8 +27,6 @@ import (
 
 	"vitess.io/vitess/go/tools/goimports"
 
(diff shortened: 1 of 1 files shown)


systay:
marking it with `release notes: none` since this pure vitess developer tooling

rsajwani (approved):
looks good to me

PR #10508, opened 2022-06-14
Rework how the `release notes` labels are handled by the CI

frouioui:
## Description

This Pull Request changes how the CI checks the `release notes` labels. The `release notes (none)` label is removed, and the reviewers are responsible for making sure the `release notes` label is set if needed. In fact, the review checklist now includes an item that invites the reviewers to check the proper setting of the `release notes` label.

The `release notes (needs details)` label is now used as a _temporary_ label, meaning that the Pull Request cannot be merged if this label is set. Setting this label on a Pull Request is a signal that the authors/reviewers need to add some content to the release notes summary file. While this label is set, the new workflow `Check Release Notes labels` will fail to prevent the merge, a sample result can be seen [here](https://github.com/vitessio/vitess/runs/6888085112?check_suite_focus=true). Once a content that describes the changes brought by the Pull Request is written, the authors/reviewers can unset the `release notes (needs details)` label, and set the `release notes` label instead. This is done to not forget the redaction of the release notes summary.

The workflow that checks whether one of the three release notes labels is set is removed in this Pull Request as we can now create a Pull Request without any release notes label.

## Checklist


frouioui:
@deepthi, we will need to remove the `required` check on the `analyze_pr_labels` workflow. We can remove it in all the different protection rules that we have since this Pull Request will be backported to active release branches.

frouioui:
Note: once this Pull Request and its backports are merged, we can remove the `release notes (none)` label from the list of labels.

.github/workflows/pr-labels.yml
```diff
@@ -1,50 +0,0 @@
...
-          "https://api.github.com/repos/${GITHUB_REPOSITORY}/issues/${PR_NUMBER}/labels" \
-          > "$LABELS_JSON"
-
-        if ! cat ${LABELS_JSON} | jq -r '.[].name ' | grep -q 'Component:' ; then
-          echo "Expecting PR to have label 'Component: ...'"
-          exit 1
-        fi
-        if ! cat ${LABELS_JSON} | jq -r '.[].name ' | grep -q 'Type:' ; then
-          echo "Expecting PR to have label 'Type: ...'"
-          exit 1
```
deepthi, line 44: Let us not delete this file. Instead, we can just remove the `release notes` label checking from here.
It is still important to set `Component` and `Type` so that PRs appear in the right section of the changelog.

deepthi (commented):
We also need to edit the Review checklist in this PR.
The section that needs to change is this
```
Ensure that the Pull Request has the correct `release notes` label. `release notes none` should only be used for PRs that are so trivial that they need not be included.
```
In fact we need only one label going forward and that is the `release notes (needs details)`, since we plan to change the release notes generation to include all PRs.
My suggestion for how this text should change:
```
If this is a change that users need to know about, please apply the `release notes (needs details)` label so that merging is blocked unless the summary release notes document is included
```

.github/workflows/pr-labels.yml
frouioui: My bad, I added it back in the new workflow file to avoid having two workflows that do (almost) the same job

frouioui:
> In fact we need only one label going forward and that is the release notes (needs details), since we plan to change the release notes generation to include all PRs.
My suggestion for how this text should change:
> ```
> If this is a change that users need to know about, please apply the `release notes (needs details)` label so that merging > is blocked unless the summary release notes document is included
> ```

The review checklist is updated accordingly.

.github/workflows/release_notes_label.yml
```diff
@@ -0,0 +1,51 @@
...
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          mode: exactly
+          count: 0
+          labels: "release notes (needs details)"
+
+      - name: Print helper
+        if: failure() && steps.required_label.outcome == 'failure'
+        run: |
+          echo The "release notes (needs details)" label is set. The changes made in this Pull Request need to be documented in the release notes summary ('./doc/releasenotes/x_x_x_summary.md'). Once documented, the "release notes (needs details)" can be set to "release notes".
```
deepthi, line 30:
```suggestion
          echo The "release notes (needs details)" label is set. The changes made in this Pull Request need to be documented in the release notes summary ('./doc/releasenotes/x_x_x_summary.md'). Once documented, the "release notes (needs details)" label can be removed.
```

deepthi (approved):
One more change and then this will be good.
Approving provisionally, you may merge once that is addressed.

deepthi:
I went ahead and removed `analyze-pr-labels` from `Required` checks on main and 14.0. It doesn't seem to exist on prior releases.
Once this PR is merged we can add `Check Pull Request labels`.

PR #10509, opened 2022-06-14
[backport] [vtctldclient] Update CLI docs for usages, flags, and aliases (#10502)

ajm188:
## Description

This backports #10502 to release-14.0

## Checklist


go/cmd/vtctldclient/command/keyspaces.go
```diff
@@ -37,14 +37,15 @@ import (
 var (
 	// CreateKeyspace makes a CreateKeyspace gRPC call to a vtctld.
 	CreateKeyspace = &cobra.Command{
-		Use:   "CreateKeyspace <keyspace> [--force|-f] [--type KEYSPACE_TYPE] [--sharding-column-name NAME --sharding-column-type TYPE] [--base-keyspace KEYSPACE --snapshot-timestamp TIME] [--served-from DB_TYPE:KEYSPACE ...]  [--durability-policy=policy_name]",
+		Use:   "CreateKeyspace <keyspace> [--force|-f] [--type KEYSPACE_TYPE] [--base-keyspace KEYSPACE --snapshot-timestamp TIME] [--served-from DB_TYPE:KEYSPACE ...]  [--durability-policy <policy_name>]",
```
deepthi, line 40: It made sense to remove these flags on main from `Use` because we have already deleted the underlying functionality.
On 14.0 however, we have only deprecated them, not deleted them. Should we leave them in `Use`? Or do something like `[--sharding-column-name NAME (deprecated) --sharding-column-type TYPE (deprecated)]`

deepthi: Ok, I'm convinced that `Use` should display "recommended" usage versus "deprecated" usage.

deepthi (approved):

PR #10513, opened 2022-06-15
[vtadmin] Add authzdocsgen to generate some website docs

ajm188:
## Description

This extends the `authztestgen` code to include a template to also generate the markdown table of RBAC rules on the website, used in https://github.com/vitessio/website/pull/1018.

### NOTES

Explicitly calling these out (happy to hear disagreement on either point!):
- I don't think we should bother backporting this PR.
- I think this is one of the very few cases we actually want release-notes-none, but happy to hear disagreement!

## Checklist


deepthi (approved):
👍

PR #10517, opened 2022-06-16
Backport to release-14: Fix parsing of CAST() statements #10512 and #10514

GuptaManan100:
## Description

This is a backport of the PRs #10512 and #10514 for release-14. These PRs fix the parsing of CAST function when used in a CREATE TABLE statement.

## Checklist


dbussink (approved):

skywhat:
I have tried to parse `cast(user_id as int)` or `cast(user_id as bigint)`, it's failed and looks like it's not supported. could you help take a look? Thanks

dbussink:
> I have tried to parse `cast(user_id as int)` or `cast(user_id as bigint)`, it's failed and looks like it's not supported. could you help take a look? Thanks

@skywhat A quick test shows that MySQL does not support this syntax either:

```
mysql> select cast(1 as bigint);
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'bigint)' at line 1
```

So Vitess would not support this either as it aims to be compatible with MySQL. If you do have a query that is parsed on MySQL but is not on Vitess, please open a new issue with the exact query so we can look at why it isn't working then.

PR #10519, opened 2022-06-16
Fix panic in VTOrc

GuptaManan100:
## Description

A panic was seen in VTOrc with the following log - 
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x58 pc=0xec5205]

goroutine 1444 [running]:
vitess.io/vitess/go/vt/orchestrator/logic.electNewPrimary({_, _}, {{{0xc00134cd70, 0xc}, 0xcea}, {{0xc00134ce80, 0xc}, 0xcea}, 0x2, {0x0, ...}, ...}, ...)
	vitess.io/vitess/go/vt/orchestrator/logic/topology_recovery.go:1778 +0x485
vitess.io/vitess/go/vt/orchestrator/logic.executeCheckAndRecoverFunction({{{0xc00134cd70, 0xc}, 0xcea}, {{0xc00134ce80, 0xc}, 0xcea}, 0x2, {0x0, 0x0, 0x0}, ...}, ...)
	vitess.io/vitess/go/vt/orchestrator/logic/topology_recovery.go:1395 +0xbbe
vitess.io/vitess/go/vt/orchestrator/logic.CheckAndRecover.func1()
	vitess.io/vitess/go/vt/orchestrator/logic/topology_recovery.go:1483 +0x5f
created by vitess.io/vitess/go/vt/orchestrator/logic.CheckAndRecover
	vitess.io/vitess/go/vt/orchestrator/logic/topology_recovery.go:1482 +0x470
```

On investigation it was found to occur because the `*events.Reparent` field returned was nil and the code does not check if the field is nil.

This PR fixes this panic by checking if `ev` is nil before using its fields.

This PR also increases the timeout of the default values of `LockShardTimeoutSeconds` and `WaitReplicasTimeoutSeconds` to 30 seconds each.
## Checklist

### DIFF
--- a/go/vt/orchestrator/config/config.go
+++ b/go/vt/orchestrator/config/config.go
@@ -388,8 +388,8 @@ func newConfiguration() *Configuration {
 		WebMessage:                                  "",
 		MaxConcurrentReplicaOperations:              5,
 		InstanceDBExecContextTimeoutSeconds:         30,
-		LockShardTimeoutSeconds:                     1,
-		WaitReplicasTimeoutSeconds:                  1,
+		LockShardTimeoutSeconds:                     30,
+		WaitReplicasTimeoutSeconds:                  30,
 	}
 }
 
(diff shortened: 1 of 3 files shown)


deepthi (commented):
The changes look good.
I'm just wondering if it is possible to write a test for this at all.
If that is too hard, can you confirm that this was verified via manual testing?

ajm188 (approved):

PR #10521, opened 2022-06-16
Make etcd based tests more deterministic and surface errors

mattlord:
## Description

We were relying on an ENV variable being set — `ETCDCTL_API=2` — when running tests locally with [etcd 3.4+ where the v2 api is disabled by default](https://etcd.io/docs/v3.4/upgrades/upgrade_3_4/#make-etcd---enable-v2false-default).

And we were swallowing errors when creating the initial topo data. This led to confusing test failures later on when e.g.
creating the keyspace:
```
E0615 13:58:10.394159   48632 vtctl_process.go:66] CreateKeyspace returned err: exit status 255, output: E0615 13:58:10.276250   48641 vtctl.go:170] action failed: CreateKeyspace GetKnownCells failed: node doesn't exist: /vitess/global/cells/
```

## Checklist

-   [x] Tests were updated

### DIFF
--- a/go/test/endtoend/cluster/topo_process.go
+++ b/go/test/endtoend/cluster/topo_process.go
@@ -57,6 +57,9 @@ func (topo *TopoProcess) Setup(topoFlavor string, cluster *LocalProcessCluster) 
 	case "consul":
 		return topo.SetupConsul(cluster)
 	default:
+		// We still rely on the etcd v2 API for things like mkdir
+		// If this ENV var is not set then tests will fail with etcd 3.x
+		os.Setenv("ETCDCTL_API", "2")
 		return topo.SetupEtcd()
 	}
 }
--- a/go/test/endtoend/cluster/topo_process.go
+++ b/go/test/endtoend/cluster/topo_process.go
@@ -57,8 +57,9 @@ func (topo *TopoProcess) Setup(topoFlavor string, cluster *LocalProcessCluster) 
 	case "consul":
 		return topo.SetupConsul(cluster)
 	default:
-		// We still rely on the etcd v2 API for things like mkdir
-		// If this ENV var is not set then tests will fail with etcd 3.x
+		// We still rely on the etcd v2 API for things like mkdir.
+		// If this ENV var is not set then some tests may fail with etcd 3.4+
+		// where the v2 API is disabled by default in both the client and server.
 		os.Setenv("ETCDCTL_API", "2")
 		return topo.SetupEtcd()
 	}
--- a/go/test/endtoend/vreplication/cluster_test.go
+++ b/go/test/endtoend/vreplication/cluster_test.go
@@ -310,10 +310,12 @@ func NewVitessCluster(t *testing.T, name string, cellNames []string, clusterConf
 
 	require.NotNil(t, topo)
 	require.Nil(t, topo.Setup("etcd2", nil))
-	topo.ManageTopoDir("mkdir", "/vitess/global")
+	err := topo.ManageTopoDir("mkdir", "/vitess/global")
+	require.NoError(t, err)
 	vc.Topo = topo
 	for _, cellName := range cellNames {
-		topo.ManageTopoDir("mkdir", "/vitess/"+cellName)
+		err := topo.ManageTopoDir("mkdir", "/vitess/"+cellName)
+		require.NoError(t, err)
 	}
 
 	vtctld := cluster.VtctldProcessInstance(vc.ClusterConfig.vtctldPort, vc.ClusterConfig.vtctldGrpcPort,


go/test/endtoend/cluster/topo_process.go
```diff
@@ -57,6 +57,9 @@ func (topo *TopoProcess) Setup(topoFlavor string, cluster *LocalProcessCluster)
 	case "consul":
 		return topo.SetupConsul(cluster)
 	default:
+		// We still rely on the etcd v2 API for things like mkdir
+		// If this ENV var is not set then tests will fail with etcd 3.x
+		os.Setenv("ETCDCTL_API", "2")
```
deepthi, line 62: How are tests working for people without setting this? I don't see it being set in `dev.env` or `build.env`.

go/test/endtoend/vreplication/cluster_test.go
```diff
@@ -310,10 +310,12 @@ func NewVitessCluster(t *testing.T, name string, cellNames []string, clusterConf
...
 	require.NotNil(t, topo)
 	require.Nil(t, topo.Setup("etcd2", nil))
-	topo.ManageTopoDir("mkdir", "/vitess/global")
+	err := topo.ManageTopoDir("mkdir", "/vitess/global")
+	require.NoError(t, err)
 	vc.Topo = topo
 	for _, cellName := range cellNames {
-		topo.ManageTopoDir("mkdir", "/vitess/"+cellName)
+		err := topo.ManageTopoDir("mkdir", "/vitess/"+cellName)
+		require.NoError(t, err)
```
deepthi, line 318: 👍

go/test/endtoend/cluster/topo_process.go
mattlord: v2 was disabled by default in 3.4: https://etcd.io/docs/v3.4/upgrades/upgrade_3_4/#make-etcd---enable-v2false-default

So if you have etcd 3.3 or older installed locally, or you have the ENV var set, things will work fine. I have 3.5.x installed locally and set the ENV var in my shell rc files. 

In the CI we explicitly use 3.3 in the unit tests: https://github.com/vitessio/vitess/blob/e74f1f1f821fa0be42effcd921e206f202926ad0/test/templates/unit_test.tpl#L125-L127

And we use Ubuntu 18.04 and 20.04 in the endtoend tests (where we install etcd from the distro repo): https://github.com/vitessio/vitess/blob/e74f1f1f821fa0be42effcd921e206f202926ad0/test/templates/cluster_endtoend_test.tpl#L20

And both ubuntu versions use 3.3: https://packages.ubuntu.com/search?keywords=etcd

So the lack of determinism in the test shows up ONLY when:
1.  You're running the tests locally
2. You have etcd 3.4+ installed
3. You do NOT have the ENV var set

mattlord: Alternatively, we could pass the `--enable-v2=true` command-line flag to etcd whenever/wherever we start it. AFAIUI, the impact is the same on the etcd server as it starts and initializes.

deepthi: Sounds good to me

deepthi: actually we are already doing that here
https://github.com/vitessio/vitess/blob/main/go/test/endtoend/cluster/topo_process.go#L76

deepthi: And I have etcd 3.5.1 locally and can run tests without setting the env variable.

dbussink: I had to set this. It seems like the env var is needed to set the client side version, not the server side of Vitess? `etcd` for me locally was started with `--enable-v2=true` already but the client still failed since it seemed to be using v3.

My local `etcd` version is 3.5.4.

dbussink (approved):

go/test/endtoend/cluster/topo_process.go
deepthi: The client in this case is vtctl, which uses a `v3` etcd client internally. The version I have has these built into it (from `go version -m vtctl`)
```
	dep	go.etcd.io/etcd/api/v3	v3.5.0	h1:GsV3S+OfZEOCNXdtNkBSR7kgLobAa/SO6tCxRa0GAYw=
	dep	go.etcd.io/etcd/client/pkg/v3	v3.5.0	h1:2aQv6F436YnN7I4VbI8PPYrBhu+SmrTaADcf8Mi/6PU=
	dep	go.etcd.io/etcd/client/v3	v3.5.0	h1:62Eh0XOro+rDwkrypAGDfgmNh5Joq+z+W9HZdlXMzek=
```
So that looks 3.4+
I don't know if there's any actual harm in setting this env variable in test code, so I'm inclined to say let's commit this and move on. It would just be nice to know for sure that this is the root cause. However, we can do that if anyone else runs into issues.
Before we wrap this up, @dbussink is it possible for you to take the other changes (the ones that don't ignore errors from `mkdir`) and get us the error messages from your test runs?

PR #10522, opened 2022-06-16
[14.0] Take into account `github.ref` when doing upgrade-downgrade tests

frouioui:
## Description

This is a backport of #10504 on `release-14.0`.

## Checklist

### DIFF
--- a/.github/workflows/static_checks_etc.yml
+++ b/.github/workflows/static_checks_etc.yml
@@ -59,49 +59,19 @@ jobs:
               - 'tools/**'
               - 'bootstrap.sh'
               - 'misc/git/hooks/asthelpers'
-            end_to_end:
+            docker:
               - 'docker/**'
               - 'test.go'
               - 'Makefile'
               - 'bootstrap.sh'
 
-
+      ## Setup and install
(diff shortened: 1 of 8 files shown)


.github/workflows/static_checks_etc.yml
```diff
@@ -59,85 +59,51 @@ jobs:
               - 'tools/**'
               - 'bootstrap.sh'
               - 'misc/git/hooks/asthelpers'
-            end_to_end:
+            docker:
```
deepthi, line 62: Were changes to this file intentional?

deepthi (approved):
Approving provisionally, see my comment

.github/workflows/static_checks_etc.yml
frouioui: Yes, it was, I thought it made more sense as we are not using any end-to-end tests in this workflow. This filtering block is used for the docker checks.

deepthi:
Let's create an issue to track the upgrade/downgrade test failures. Most of the failures are hangs during vttablet startup / vtgate shutdown, so I'm suspecting a problem with the test framework.

Issue #10363, opened 2022-05-24
Task: Add full super_read_only support in Vitess


mattlord:
The schema.engine, schemamanager, restore, … components perform DDL and assume they can do it anywhere, anytime — which they could before when using the `root` and/or `vt_dba` user which has the `SUPER` privilege and is thus not impacted by `read_only`. We really should be leveraging `super_read_only` as w/o it there will be times when we allow errant GTIDs that are then lying in wait to cause later failures, but as you can see in https://github.com/vitessio/vitess/pull/10094 this will require some work.

We'll need to think this through and IMO do some refactoring so that we put a vttablet+mysqld into a specific mode where `super_read_only` is intentionally disabled for it — locking the tablet in the topo — to perform some specific task(s) (e.g. tablet init) and then re-enabling `super_read_only` upon completion of the task(s), rather than attempting to turn that global variable ON and OFF per query or session.

> Note: `super_read_only` is a global variable so when disabling it so that we can execute specific queries in our session, we begin fighting/racing with tablet repair and other sessions

> Note: the `init_db.sql` file contents would not support MariaDB – the actual file is used in various places too (from shell scripts, yaml files, etc), not just the embedded content (this is fine as MariaDB support is deprecated in v14.0)

## Related Issues
  - Related to: https://github.com/vitessio/vitess/pull/10094

PR #10533, opened 2022-06-16
Draft: Tablet schema initialization refactoring

rsajwani:

## Description

This is an attempt to consolidated all schema initialization and DB creation at the tablet start-up. The idea is if we can do all schema changes in one place and then seal the db with super-read-only, it will help us avoid errant GTIDs.

## Details 

We have created Schema_Initialization class which contains list of initialization modules needed to be run during tablet setup. Every module needs to register themselves with schema_Initializer e.g metadata_tables.go, online DDL/executor.go, rpc_replication.go etc. 

During tablet initialization at various places in code-base whether we are bootstrapping a completely new tablet or restoring from backup we executes these modules registered as part of their init() routine. Once all the modules are executed we set 'super_read_only'  false. This means we won't allow any mutation outside primary tablet. This will help us avoid errand GTIDS.

During tablet promotion we set super-read-only to false and during tablet demotion we set super-read-only to true. This way apart from primary tablet no other replica will be able to produce schema changes from outside.

## For Reference 
Some related work items done in the past

https://github.com/vitessio/vitess/pull/10094
https://github.com/vitessio/vitess/pull/9312
https://github.com/vitessio/vitess/pull/10448

## Unresolved Issued during code changes:

- I have to create a separate init_db.sql for maria db since maria db does not have super_read_only. I could either have an if-then-else condition inside maria db but then I choose instead of creating a separate init_db file.
- Adding # add custom sql here  inside init-db.sql file. This is because in our test we modify the file and add custom slq like (change password of dba etc). Now that we expect super_read_only to execute at the end, we need these custom sql to be added before that last line. See example go/test/endtoend/backup/vtbackup/main_test.go
- In one of our end to end test we have to set super-read-only test to false explicitly else our upgrade-downgrad test failed since they run with N-1 vttablet binary, which does not have code to disable super-read-only during InitializeShard. e.g 
go/test/endtoend/cluster/cluster_process.go
- In tests like backup_only_test.go, I have to induce sleep else the tests are failing. I have a hunch why it is failing but I am not able to find a good solution for that.
- Was not able to initialize the schema for health_streamer. Because if we create schema-copy-table beforehand then a test case like vitess.io/vitess/go/test/endtoend/vtgate/schematracker/loadkeyspace -run TestBlockedLoadKeyspace fails. The reason is they expect schema-copy-table to present only if we have enabled schema tracking. 
- Should we return false / error if any of the schema initialization fails … Right now I just log them but I believe we should return fail tablet initialization if any of the schema initialization fails e.g createReparentJournal.

## Related Issue(s)

#10363

## Checklist


config/init_maria_db.sql
```diff
@@ -0,0 +1,104 @@
+# This file is executed immediately after mysql_install_db,
+# to initialize a fresh data directory.
```
GuptaManan100, line 2: Is this separate file created for the super_read_only changes which are only present in the MySQL file?

go/vt/vttest/environment.go
```diff
@@ -144,9 +147,14 @@ func (env *LocalTestEnv) BinaryPath(binary string) string {
 
 // MySQLManager implements MySQLManager for LocalTestEnv
 func (env *LocalTestEnv) MySQLManager(mycnf []string, snapshot string) (MySQLManager, error) {
+	// maria db doesn't recognize super-read-only, therefore we have separate sql for that.
+	var initFile = path.Join(os.Getenv("VTROOT"), "config/init_db.sql")
+	if !env.IsSQLFlavor() {
+		initFile = path.Join(os.Getenv("VTROOT"), "config/init_maria_db.sql")
+	}
```
GuptaManan100, line 154: Instead of asking the environment, we could query the MySQL instance directly to know if it is mariadb distribution or not. This would prevent users from running into configuration issues.

test/templates/unit_test.tpl
```diff
@@ -128,4 +128,8 @@ jobs:
       if: steps.skip-workflow.outputs.skip-workflow == 'false' && steps.changes.outputs.unit_tests == 'true'
       timeout-minutes: 30
       run: |
+        {{if (eq .Platform "mariadb103")}}
+        # mariadb103
+        export MYSQL_FLAVOR="MariaDB103"
+        {{end}}
```
GuptaManan100, line 134: If we decide to go with reading the MySQL version from the instance itself, this won't be required.

go/vt/srvtopo/query.go
```diff
@@ -60,6 +60,19 @@ func (q *resilientQuery) getCurrentValue(ctx context.Context, wkey fmt.Stringer,
...
+	if *srvTopoNoCacheForGet {
+		entry := &queryEntry{
+			key: wkey,
+		}
+		result, err := q.query(ctx, entry)
+		if err != nil {
+			return nil, err
+		}
+		return result, nil
+	}
```
GuptaManan100, line 74: Is this change related to schema initialization?

go/flags/endtoend/vtctld.txt
```diff
@@ -103,6 +103,7 @@ Usage of vtctld:
       --hot_row_protection_concurrent_transactions int                   Number of concurrent transactions let through to the txpool/MySQL for the same hot row. Should be > 1 to have enough 'ready' transactions in MySQL and benefit from a pipelining effect. (default 5)
       --hot_row_protection_max_global_queue_size int                     Global queue limit across all row (ranges). Useful to prevent that the queue can grow unbounded. (default 1000)
       --hot_row_protection_max_queue_size int                            Maximum number of BeginExecute RPCs which will be queued for the same row (range). (default 20)
+      --init_populate_metadata                                           (init parameter) populate metadata tables even if restore_from_backup is disabled. If restore_from_backup is enabled, metadata tables are always populated regardless of this flag.
```
GuptaManan100, line 106: This flag shouldn't be in vtctld

PR #10534, opened 2022-06-17
fix: handle planner_version and planner-version correctly

systay:
## Description
I recently tried fixing this situation, and I think I made it worse in that PR. 🤦 

So here we go again.

Some applications were using `planner-version`, and some were using `planner_version`. This change makes it possible to use either, but either they have to agree, or only one of them set.

If the user is using `planner_version`, which is the deprecated form, they will get a log message saying they shouldn't use that anymore.

## Related Issue(s)
Original attempt: #10453 

## Checklist


go/cmd/vtcombo/main.go
```diff
@@ -61,7 +61,7 @@ var (
 	mysqlPort          = flag.Int("mysql_port", 3306, "mysql port")
 	externalTopoServer = flag.Bool("external_topo_server", false, "Should vtcombo use an external topology server instead of starting its own in-memory topology server. "+
 		"If true, vtcombo will use the flags defined in topo/server.go to open topo server")
-	plannerVersion           = flag.String("planner-version", "gen4", "Sets the default planner to use when the session has not changed it. Valid values are: V3, Gen4, Gen4Greedy and Gen4Fallback. Gen4Fallback tries the gen4 planner and falls back to the V3 planner if the gen4 fails.")
```
GuptaManan100, line 64: This makes a lot of sense. Otherwise users would have provided the old flag, and it would have conflicted with the default in this flag!

GuptaManan100 (approved):

Issue #10520, opened 2022-06-16
Bug Report: Vtgate query log table field empty for DML queries if executed in unsharded keyspace


DeathBorn:
No TableName is returned for DML queries in **unsharded** keyspace when looking into vtgate query logs. This potentially affects multiple parts of code too.
v3 and gen4 both affected. The issue happens with old vitess versions too.

Here https://github.com/vitessio/vitess/blob/main/go/vt/vtgate/planbuilder/dml_planner.go#L288-L303 if keyspace is unsharded, then it won't set edml table field making it empty.
Additionally, multi-table updates will pick last table in randomly ordered list of tables if this https://github.com/vitessio/vitess/blob/main/go/vt/vtgate/planbuilder/dml_planner.go#L323-L326 would be moved before return block.
Furthermore, there are multiple places where this field is used in code.

There is already a case written in [`dml_cases.txt`](https://github.com/vitessio/vitess/blob/main/go/vt/vtgate/planbuilder/testdata/dml_cases.txt#L720-L734) but without table field.

PR #10536, opened 2022-06-17
Fix vtgate query log table name extraction for DML statements

DeathBorn:
## Description
Instead of single table for a dml primitive, now we have a list of tables.
For unsharded table queries, now we have multiple table names extracted.
For sharded table queries, there should be only one table in this list, so we will always pick first element in the list where it is needed.

Now both v3 and gen4 gives us a list of all used tables in DMLs.

## Related Issue(s)

Initial try to fix https://github.com/vitessio/vitess/issues/10520

## Checklist

Signed-off-by: Vilius Okockis <vilius.okockis@vinted.com>


go/vt/vtgate/engine/delete.go
```diff
@@ -128,7 +141,7 @@ func (del *Delete) deleteVindexEntries(vcursor VCursor, bindVars map[string]*que
 			return err
 		}
 		colnum := del.KsidLength
-		for _, colVindex := range del.Table.Owned {
+		for _, colVindex := range del.Table[0].Owned {
```
DeathBorn, line 144: This should be ok since multi table/table delete is not supported

DeathBorn:
https://github.com/vitessio/vitess/pull/10536/files#diff-3f911a9c256378c52bc89d073fe17a1867c5d4e8e6d197c7106105f20e93bf6eR288 @systay I seem to hit a complex code path here. Due to subquery alias, I get zero length table name. Not sure where to go next...

go/vt/vtgate/planbuilder/from.go
```diff
@@ -153,8 +154,14 @@ func (pb *primitiveBuilder) processAliasedTable(tableExpr *sqlparser.AliasedTabl
 		// a new set of column references will be generated against the new tables,
 		// and those vindex maps will be returned. They have to replace the old vindex
 		// maps of the inherited route options.
+		var tableNames []string
```
DeathBorn, line 157: this saves subquery original table names into "virtual" alias table. Without this, subquery tables are lost further on. The drawback of this approach  - duplicate table names appear (delete dml plan test added for this)

go/vt/vtgate/planbuilder/symtab.go
```diff
@@ -205,6 +205,20 @@ func (st *symtab) AllTables() []*table {
...
 }
 
+// AllVschemaTableNames returns an ordered list of all current vschema tables.
+func (st *symtab) AllVschemaTableNames() []*vindexes.Table {
+	if len(st.tableNames) == 0 {
+		return nil
+	}
+	tables := make([]*vindexes.Table, 0, len(st.tableNames))
+	for _, tname := range st.tableNames {
+		if vschemaTable := st.tables[tname].vschemaTable; vschemaTable != nil {
```
DeathBorn, line 215: vschemaTable  is nil for `information_schema` tables, so we ignore them here. these tables seem to appear correctly anyway.
Anyhow, I found this https://github.com/vitessio/vitess/blob/acb8c2f762faea1637e62790bcd934d96b47e538/go/vt/vtgate/planbuilder/testdata/from_cases.txt#L2243-L2278 - table names are bad. Maybe there is a reason for this, I don't know...

go/vt/vtgate/engine/update.go
```diff
@@ -148,7 +160,7 @@ func (upd *Update) updateVindexEntries(vcursor VCursor, bindVars map[string]*que
 			return err
 		}
 
-		for _, colVindex := range upd.Table.ColumnVindexes {
+		for _, colVindex := range upd.Table[0].ColumnVindexes {
```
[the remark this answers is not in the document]
systay: WDYT about doing something like this?

```golang
func (del *Delete) singleTable() (*vindexes.Table, error) {
	if len(del.Table) > 1 {
		return nil, vterrors.Errorf(vtrpc.Code_INTERNAL, "expected single table to exist")
	}
	return del.Table[0], nil
}
```

this way if our assumptions change, we'll get a clear error

go/vt/vtgate/engine/update.go
```diff
@@ -64,7 +65,18 @@ func (upd *Update) GetKeyspaceName() string {
 // GetTableName specifies the table that this primitive routes to.
 func (upd *Update) GetTableName() string {
 	if upd.Table != nil {
-		return upd.Table.Name.String()
+		tableNameMap := map[string]any{}
```
systay, line 68: this is the same code as for delete. we should extract it into a helper method on DML

go/vt/vtgate/planbuilder/expr.go
```diff
@@ -265,15 +268,16 @@ func (pb *primitiveBuilder) finalizeUnshardedDMLSubqueries(reservedVars *sqlpars
 					samePlan = false
 					return false, dummyErr
 				}
+				tables = append(tables, spb.st.AllVschemaTableNames()...)
```
systay, line 271: we don't need this here, it's enough to handle it on the per-Select cases.

go/vt/vtgate/semantics/analyzer_test.go
```diff
@@ -1380,7 +1380,8 @@ func TestSingleUnshardedKeyspace(t *testing.T) {
 	for _, test := range tests {
 		t.Run(test.query, func(t *testing.T) {
 			_, semTable := parseAndAnalyze(t, test.query, "d")
-			assert.Equal(t, test.unsharded, semTable.SingleUnshardedKeyspace())
+			queryIsUnsharded, _ := semTable.SingleUnshardedKeyspace()
```
systay, line 1383: might as well check the returned table names as well

DeathBorn:
@systay could you please add pull request labels ?

go/vt/vtgate/planbuilder/update_planner.go
```diff
@@ -46,7 +46,7 @@ func buildUpdatePlan(query string) stmtPlanner {
 			return eupd, nil
 		}
 
-		cvv, ovq, err := buildChangedVindexesValues(upd, eupd.Table, ksidVindex.Columns)
+		cvv, ovq, err := buildChangedVindexesValues(upd, eupd.Table[0], ksidVindex.Columns)
```
DeathBorn, line 49: missed this

go/vt/vtgate/planbuilder/symtab.go
DeathBorn: This is due to https://github.com/vitessio/vitess/blob/main/go/vt/vtgate/engine/join.go#L241
Could be fixed in another PR since it involves lots of cases.

DeathBorn:
One thing -> metric labels for vtgates will change to this, since, multiple table names are joined.

```
vtgate_queries_routed_by_table{keyspace="k1",plan="SelectUnsharded",table="foo, bar, gaa"}
```

DeathBorn:
@frouioui could you look into this ?

DeathBorn:
At this moment it is ready for final review

Issue #10454, opened 2022-06-07
Delete `legacy_healthcheck.go` and any related structs that are no longer needed


frouioui:
Following the removal of the legacy gateway, the legacy healthcheck should also be removed entirely. Its use in vtctld was stopped through https://github.com/vitessio/vitess/issues/9358.

This blocks https://github.com/vitessio/vitess/issues/9218.

PR #10278, opened 2022-05-11
Delete all legacy sharding related code

mattlord:
Signed-off-by: Matt Lord <mattalord@gmail.com>

## Checklist

-   [ ] "Backport me!" label has been added if this change should be backported
-   [ ] Tests were added or are not required
-   [ ] Documentation was added or is not required

PR #10542, opened 2022-06-17
Remove legacy healthcheck files and structures

frouioui:
## Description

This Pull Request removes the legacy healthcheck and the related code (legacy topo watcher, legacy replication lag, etc). The different usages have been replaced by the new implementation of the healthcheck. The different files hosting the legacy code have been removed.

## Related Issue(s)

- Part of https://github.com/vitessio/vitess/issues/9218
- Closes https://github.com/vitessio/vitess/issues/10454

## Checklist


deepthi:
Since this is cleanup, we don't need to back port it to v14. It's a teeny bit risky to refactor this close to release on the release branch.

go/vt/discovery/healthcheck.go
```diff
@@ -625,6 +631,25 @@ func (hc *HealthCheckImpl) GetHealthyTabletStats(target *query.Target) []*Tablet
 	return append(result, hc.healthy[KeyFromTarget(target)]...)
 }
 
+// GetHealthyTabletStats returns only the healthy tablets.
```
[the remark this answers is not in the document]
frouioui: good catch, thank you

go/vt/worker/executor.go
```diff
@@ -41,7 +41,7 @@ import (
 // executor is also used for executing vreplication and RefreshState commands.
 type executor struct {
 	wr        *wrangler.Wrangler
-	tsc       *discovery.LegacyTabletStatsCache
+	hc        *discovery.HealthCheckImpl
```
systay, line 44: Couldn't we use the interface instead of the impl?

frouioui: we definitely can! good catch

go/vt/vtgate/vtgate.go
```diff
@@ -141,7 +141,6 @@ type VTGate struct {
 	vsm      *vstreamManager
 	txConn   *TxConn
 	gw       *TabletGateway
-	pv       plancontext.PlannerVersion
```
deepthi, line 144: This doesn't seem to belong in this PR.

go/vt/discovery/healthcheck.go
```diff
@@ -625,14 +631,17 @@ func (hc *HealthCheckImpl) GetHealthyTabletStats(target *query.Target) []*Tablet
...
 // For TabletType_PRIMARY, this will only return at most one entry,
 // the most recent tablet of type primary.
-func (hc *HealthCheckImpl) getTabletStats(target *query.Target) []*TabletHealth {
+func (hc *HealthCheckImpl) GetTabletStats(target *query.Target) []*TabletHealth {
 	var result []*TabletHealth
 	hc.mu.Lock()
 	defer hc.mu.Unlock()
+	if target.Shard == "" {
+		target.Shard = "0"
+	}
```
deepthi, line 644: This looks like a bad thing to do. Why is the Shard not specified when we get the target?

go/vt/discovery/healthcheck.go
```diff
@@ -166,6 +170,8 @@ type tabletAliasString string
 
 // HealthCheck declares what the TabletGateway needs from the HealthCheck
 type HealthCheck interface {
+	TabletRecorder
+
```
deepthi, line 174: Why are we embedding an interface into another? Do we have implementations of TabletRecorder that don't implement HealthCheck?

go/vt/wrangler/keyspace.go
```diff
@@ -917,7 +919,7 @@ func (wr *Wrangler) updateFrozenFlag(ctx context.Context, shards []*topo.ShardIn
 // the tablet was actually drained. At later times, a QPS rate > 0.0 could still
 // be observed.
 func (wr *Wrangler) WaitForDrain(ctx context.Context, cells []string, keyspace, shard string, servedType topodatapb.TabletType,
-	retryDelay, healthCheckTopologyRefresh, healthcheckRetryDelay, healthCheckTimeout, initialWait time.Duration) error {
```
deepthi, line 920: hmm, why have we removed this parameter?

go/vt/discovery/healthcheck.go
```diff
@@ -81,12 +82,15 @@ var (
...
+	// How much to sleep between each check.
+	waitAvailableTabletInterval = 100 * time.Millisecond
 )
 
 // See the documentation for NewHealthCheck below for an explanation of these parameters.
 const (
-	defaultHealthCheckRetryDelay = 5 * time.Second
-	defaultHealthCheckTimeout    = 1 * time.Minute
+	DefaultHealthCheckRetryDelay = 5 * time.Second
+	DefaultHealthCheckTimeout    = 1 * time.Minute
```
deepthi, line 93: Why do we need to export these?

go/vt/wrangler/keyspace.go
```diff
@@ -974,15 +970,15 @@ func (wr *Wrangler) waitForDrainInCell(ctx context.Context, cell, keyspace, shar
 	startTime := time.Now()
 	for {
 		// map key: tablet uid
-		drainedHealthyTablets := make(map[uint32]*discovery.LegacyTabletStats)
-		notDrainedHealtyTablets := make(map[uint32]*discovery.LegacyTabletStats)
+		drainedHealthyTablets := make(map[uint32]*discovery.TabletHealth)
+		notDrainedHealtyTablets := make(map[uint32]*discovery.TabletHealth)
 
-		healthyTablets := tsc.GetHealthyTabletStats(keyspace, shard, servedType)
```
deepthi, line 980: Why `GetTabletStats` instead of `GetHealthy..`? The new healthcheck also has a `GetHealthy..` func.

go/vt/throttler/manager.go
```diff
@@ -207,7 +207,7 @@ func (m *managerImpl) throttlerNamesLocked() []string {
 
 // Log returns the most recent changes of the MaxReplicationLag module.
 // There will be one result for each processed replication lag record.
-func (m *managerImpl) Log(throttlerName string) ([]result, error) {
+func (m *managerImpl) Log(throttlerName string) ([]Result, error) {
```
deepthi, line 210: Why does `Result` need to be exported? I tried to figure it out but it is not obvious.

go/vt/discovery/healthcheck.go
frouioui: > Do we have implementations of TabletRecorder that don't implement HealthCheck?

No, we don't.

> Why are we embedding an interface into another?

This was done so we can use the healthcheck (`discovery.HealthCheck`) as a tablet recorder in `tx throttler`: https://github.com/vitessio/vitess/blob/4654a771c2868fd0a6eab8ac96af39f82e5c532c/go/vt/vttablet/tabletserver/txthrottler/tx_throttler.go#L199

Ultimately, the topo watcher in the `discovery` package uses a `TabletRecorder`.

Previously, the `LegacyHealthCheck` interface was embedding `LegacyTabletRecorder`:
https://github.com/vitessio/vitess/blob/d037f24c8c4efd52cf70fac2e11fcb2941ce0a32/go/vt/discovery/legacy_healthcheck.go#L285-L289

deepthi: we should not have to embed it for things to work.

go/vt/wrangler/keyspace.go
deepthi: #10278 deletes WaitForDrain, so it doesn't really matter what we do here any more.

go/vt/worker/executor.go
frouioui: this file has been removed on main

frouioui:
> Only one change is required - to remove TabletRecorder from HealthCheck interface.

We decided to remove the `TabletRecorder` interface and move its methods directly to the `HealthCheck` interface. They were both implemented by the same set of structures. This was done via https://github.com/vitessio/vitess/pull/10542/commits/b89882fdd19b35a1037614bbb083d10de5da288f.

PR #10543, opened 2022-06-18
add vtadmin docker image

L3o-pold:
## Description

Add vtadmin docker image. Add the possibility to serve the vtadmin frontend inside docker image.

For now it's using node serving binary to serve the content but it could be change to have a single-component vtadmin.

Frontend have been modified in order to be able to change env var at the run time (not only build time). 

Usage in k8s looks like this:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: vtadmin-web-config
data:
  config.js: |
    window.env = {
    'REACT_APP_VTADMIN_API_ADDRESS': "https://my-vtadmin-endpoint.local",
    'REACT_APP_FETCH_CREDENTIALS': "omit",
    'REACT_APP_ENABLE_EXPERIMENTAL_TABLET_DEBUG_VARS': true,
    'REACT_APP_BUGSNAG_API_KEY': "",
    'REACT_APP_DOCUMENT_TITLE': "",
    'REACT_APP_READONLY_MODE': false,
  };
```

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: vtadmin-api-config
data:
  discovery.json: |
    {
        "vtctlds": [
            {
                "host": {
                    "fqdn": "vtctld:15999",
                    "hostname": "vtctld:15999"
                }
            }
        ],
        "vtgates": [
            {
                "host": {
                    "hostname": "vtgate:15999"
                }
            }
        ]
    }
  rbac.yaml: |
    rules:
      - resource: "*"
        actions:
          - "get"
          - "create"
          - "delete"
          - "put"
          - "ping"
        subjects: ["*"]
        clusters: ["*"]
      - resource: "Shard"
        actions:
          - "emergency_reparent_shard"
          - "planned_reparent_shard"
        subjects: ["*"]
        clusters:
          - "cell1"

```

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: vtadmin
  labels:
    app.kubernetes.io/name: vtadmin
spec:
  replicas: 1
  selector:
    matchLabels:
      app.kubernetes.io/name: vtadmin
  template:
    metadata:
      labels:
        app.kubernetes.io/name: vtadmin
    spec:
      terminationGracePeriodSeconds: 30
      containers:
        - name: vtadmin
          image: vitess/vtadmin:v13.0.0-bullseye
          imagePullPolicy: IfNotPresent
          command:
            - /vt/bin/vtadmin
          args:
            - --addr=:14200
            - --http-origin=*
            - --cluster=id=cell1,name=cell1,discovery=staticfile,discovery-staticfile-path=/var/vtadmin/discovery.json,tablet-fqdn-tmpl={{ .Tablet.Hostname }}
            - --rbac
            - --rbac-config=/var/vtadmin/rbac.yaml
            - --tracer=opentracing-jaeger
            - --grpc-tracing
            - --http-tracing
            - --logtostderr
            - --alsologtostderr
          env:
            - name: GOMAXPROCS
              value: "4"
          ports:
            - containerPort: 14200
              name: tcp-api
              protocol: TCP
          resources:
            limits:
              cpu: 200m
              memory: 512Mi
            requests:
              cpu: 100m
              memory: 256Mi
          securityContext:
            runAsUser: 999
          volumeMounts:
            - mountPath: /var/vtadmin
              name: vtadmin-config
              readOnly: true
        - name: web
          image: vitess/vtadmin:v13.0.0-bullseye
          imagePullPolicy: IfNotPresent
          ports:
            - containerPort: 14201
              name: tcp-web
              protocol: TCP
          securityContext:
            runAsUser: 999
          volumeMounts:
            - mountPath: /vt/web/vtadmin/build/config
              name: vtadmin-web-config
              readOnly: true
          resources:
            limits:
              cpu: 200m
              memory: 512Mi
            requests:
              cpu: 100m
              memory: 256Mi
      volumes:
        - name: vtadmin-api-config
          configMap:
            name: vtadmin-api-config
        - name: vtadmin-web-config
          configMap:
            name: vtadmin-web-config
```

## Checklist


docker/k8s/vtadmin/Dockerfile
```diff
@@ -0,0 +1,43 @@
+# Copyright 2019 The Vitess Authors.
```
derekperkins, line 1: 2022

docker/base/Dockerfile
```diff
@@ -34,11 +48,12 @@ ARG BUILD_NUMBER
...
+COPY --from=node /vt/web/vtadmin /vt/src/vitess.io/vitess/web/vtadmin
 COPY . /vt/src/vitess.io/vitess
 
 # Build Vitess
 RUN make build
 
 # Fix permissions
 RUN chown -R vitess:vitess /vt
-USER vitess
+USER vitess
```
ajm188, line 59: nit: newline at EOF please!

L3o-pold:
yeah I saw where it was failing but no idea on how to fix it yet.

notfelineit:
Hey @L3o-pold, have you tried making this a function instead of a var like so:
```
export const env = () => NodeJS.ProcessEnv = { ...window.env, ...process.env };
```
I'll take a deeper look at your PR as well, that's just a javascript-y hunch 😃

L3o-pold:
image

My PR runs but I didn't try with the `REACT_APP_ENABLE_EXPERIMENTAL_TABLET_DEBUG_VARS` env var set to true.

ajm188:
the problem is that `window.env` is not of type `NodeJS.ProcessEnv`, do we _need_ that variable, or can we type coerce it somehow so everything plays nice?

in any case, once we get that working, you're going to want to update the before/after mock infra (see http.test.ts for an example, which is explicitly mocking `process.env`, so you'll want to sub in your util function there instead)

L3o-pold:
The goal of using `window.env` is to be able to edit environment variable as `REACT_APP_VTADMIN_API_ADDRESS` on the runtime. 

With this you can build a vtadmin dist version and allow people to edit those var when running the frontend. 

For example building a Vtadmin docker image and let users specify their own `REACT_APP_VTADMIN_API_ADDRESS` when running it.

L3o-pold:
@ajm188 did a trick for fixing the CI. IMO cookie is ok only if we plan of adding an UI to edit those settings live. 
For now (with this PR), users need to play with `sed` in order to edit this settings during the runtime, but if we plan to replace `serve` dependency with go it could also be done directly in the binary I think.

ajm188:
@notfelineit this[1] totally _would_ have worked, but there are tests for the error handler that mutate the `process.env` to make some assertions

[1] "this" meaning your suggestion here

> Are we able to access the variables we need simply by exporting an env var?
>
> `const env = {...window.env, ...process.env}`

notfelineit:
@ajm188 could you link to the tests mutating process.env? That sounds like funny test behavior 🤔

ajm188:
yep! https://github.com/vitessio/vitess/blob/main/web/vtadmin/src/api/http.test.ts#L58-L90 and also https://github.com/vitessio/vitess/blob/main/web/vtadmin/src/errors/errorHandler.test.ts#L21-L45

notfelineit:
@ajm188 so I guess the issue with having a static object like `const env = {...window.env, ...process.env}` for the tests is that `env` might be set before the test process.env is applied? I don't see any tests that explicity make assertions against the process.env object, only the before/afters that set them.

I think a function would handle the above case ^ (recalculating `env` each time so it's always fresh). For handling type safety, would this work?:
```
const env = () => { ...window.env, ...process.env } as NodeJs.ProcessEnv
```

ajm188:
that's ... exactly what i did. did you look at https://github.com/vitessio/vitess/pull/10543/commits/d8ca36a013978c5a34bb273d10f30ef8b0be7da3 ?

notfelineit:
@ajm188 No I missed it! If that works for the tests, that sounds great

GuptaManan100 (commented):
The changes to building the vtadmin web binaries are made in the `base` image which k8s is using so it works out there. But the `lite` images don't use `base`. They use `bootstrap` so this change may be needed in all of those images too.

docker/base/Dockerfile
```diff
@@ -24,6 +24,14 @@
...
 ARG image="vitess/bootstrap:${bootstrap_version}-mysql57"
 
+FROM node:16-bullseye-slim as node
+
+COPY web/vtadmin /vt/web/vtadmin
+
+# install/build/clean web dependencies
+RUN npm --prefix /vt/web/vtadmin ci && \
+    npm run --prefix /vt/web/vtadmin build
+
```
deepthi, line 34: If we are adding this to the bootstrap image, then we don't need to add it here.
Instead, we need to first build and push bootstrap images with a new version number (10), and increment the version number in all the Dockerfiles that use that as the base.

deepthi (commented):
Adding node to the `lite` image in order to run `vtadmin` is acceptable. We just need to measure the image size before and after and document the increase.

docker/k8s/vtadmin/Dockerfile
```diff
@@ -43,8 +43,13 @@ RUN deluser node && \
    groupadd -r --gid 2000 vitess && \
    useradd -r -g vitess --uid 1000 vitess && \
    chown -R vitess:vitess /vt && \
-   chown -R vitess:vitess /vtdataroot
+   chown -R vitess:vitess /vtdataroot && \
+   # the entrypoint need to edit vtadmin configuration, so it needs to be writable by every user
+   chmod 777 /vt/web/vtadmin/build/index.html
```
derekperkins, line 48: I don't understand why that needs to be writeable. Changes aren't persisted to the file, they are generally stored in the topo

derekperkins: Hmm... Shouldn't making it writable by the vitess user be sufficient like the other folders? IIUC, the entrypoint should be executed as that user

L3o-pold: no because we want to allow user to run the image with any user (like in the vitess operator with an non existing one)

docker/k8s/vtadmin/Dockerfile
```diff
@@ -0,0 +1,55 @@
...
+   chown -R vitess:vitess /vt && \
+   chown -R vitess:vitess /vtdataroot && \
+   # the entrypoint need to edit vtadmin configuration, so it needs to be writable by every user
+   chmod 777 /vt/web/vtadmin/build/index.html
+
+USER vitess
+
+ENTRYPOINT ["/vt/entrypoint.sh"]
+
+CMD ["/vt/web/vtadmin/node_modules/.bin/serve", "--no-clipboard", "-l", "$VTADMIN_WEB_PORT", "-s", "/vt/web/vtadmin/build"]
```
derekperkins, line 54: What's the reason for having both `ENTRYPOINT` and `CMD`?

L3o-pold: It's easier to overwrite the command in a k8s deployment for example.

docker/bootstrap/Dockerfile.common
```diff
@@ -1,3 +1,11 @@
+FROM node:16-bullseye-slim as node
```
dbussink, line 1: This is using `bullseye`, but then we're using `buster` for the Go image. I don't think we should mix different bistro versions to avoid risk of mismatches in binary symbols etc.

docker/base/Dockerfile
deepthi: based on offline discussion, I now see that we are actually _building_ vtadmin in this step. So it doesn't belong in the bootstrap image at all.
bootstrap image should contain static dependencies (like node and npm). Builds need to happen in base and lite.
I was confused because the changes were also made to the bootstrap image.
One way forward is to revert changes to bootstrap image and make them in all the lite images. Then there is no need for a new bootstrap version either.
I'll leave it up to you, @ajm188 and @GuptaManan100 to decide whether node/npm need to be installed in the bootstrap image at all.

GuptaManan100, line 34: We should do the build of the vtadmin binaries in the k8s vtadmin docker file itself. We only use this build for that image, so it doesn't make sense to add it to all the other images.

docker/bootstrap/Dockerfile.common
```diff
@@ -29,6 +29,7 @@ ENV USER vitess
 COPY bootstrap.sh dev.env build.env go.mod go.sum /vt/src/vitess.io/vitess/
 COPY config /vt/src/vitess.io/vitess/config
 COPY tools /vt/src/vitess.io/vitess/tools
+COPY --from=node /vt/web/vtadmin /vt/src/vitess.io/vitess/web/vtadmin
```
GuptaManan100, line 32: This line should be removed. node is longer defined here. It will fail

GuptaManan100 (commented):
Me and @deepthi had a discussion and I think for now it makes sense to only have the node installation, build, etc in the vtadmin image only. Let us not do this for the `lite` images until we have a clean way of doing this without any security concerns. 
This will allow us time after the release to evaluate our options and make changes without being rushed. Meanwhile we can introduce a new vtadmin image which can have node installed, do the build and the `sed` command changes, allowing us to deploy vtadmin in vtop so that people can start trying it out!

ajm188:
@L3o-pold regarding the `sed` business and the need to `chmod 777` ...

i think instead we should provide a volume mount point, where the user can run a container with some sort of .env file (that has all the `REACT_VTADMIN_*` variables. Separately, we include a shell script in the image that, if that file is present, does the `sed`. This separates the image-build / container-init phases, doesn't require the arbitrary write perms, and solves the problem. Let me know what you think!

L3o-pold:
@ajm188 not sure to understand. The sed requires the right on the `index.html` file. I don't see how providing a volume mount for `.env` file will solve this.

L3o-pold:
I'll re-work to move from a `sed` / `index.html` thing to a `config.js` script loaded into the `index.html`. It will avoid any permission and weird entrypoint magic and will let the user mount his config file like (or if) he wants.

web/vtadmin/public/config.js
```diff
@@ -0,0 +1,8 @@
+window.env = {
+    'REACT_APP_VTADMIN_API_ADDRESS': "",
+    'REACT_APP_FETCH_CREDENTIALS': "omit",
+    'REACT_APP_ENABLE_EXPERIMENTAL_TABLET_DEBUG_VARS': true,
+    'REACT_APP_BUGSNAG_API_KEY': "",
+    'REACT_APP_DOCUMENT_TITLE': "",
+    'REACT_APP_READONLY_MODE': false,
+};
```
ajm188, line 8: nit: add a trailing newline please

docker/lite/Dockerfile.mysql57
```diff
@@ -51,7 +51,6 @@ ENV PATH $VTROOT/bin:$PATH
 # Copy artifacts from builder layer.
 COPY --from=builder --chown=vitess:vitess /vt/install /vt
 COPY --from=builder --chown=vitess:vitess /vt/src/vitess.io/vitess/web/orchestrator /vt/web/orchestrator
-COPY --from=builder --chown=vitess:vitess /vt/src/vitess.io/vitess/web/vtadmin /vt/web/vtadmin
```
deepthi, line 54: This reverts #9405 and probably breaks anyone currently relying on it.
Let's revert this one line for now.

deepthi:
Since vtadmin is its own image, the example usage in the description needs to be changed to reflect that.
Have you verified that the example config actually works with a vtadmin image built with this PR?

deepthi (approved):
This is looking good to me. It looks like all feedback has been addressed
- no changes to bootstrap images
- no changes to lite images
- distro (bullseye) is not hard-coded, it is passed to the release script
- working example (confirmed with PR author via slack) in the description

If anyone finds problems with this after we merge it, we can fix it forward.

PR #10552, opened 2022-06-21
vtctl GetSchema --table_schema_only

shlomi-noach:
## Description

This PR adds the flag `--table_schema_only` to `GetSchema`. When specified, `GetSchema` skips columns introspection. It will not query column identities nor attempt to discover column types.

The output will include general schema details, and specifically it will include the `CREATE TABLE|VIEW` statement as part of the `schema` field.

This flag actually makes `GetSchema` skip introspection, and thus makes it faster. If column information is not needed, this flag is recommended.

Also, with this flag, it is possible to get information about views with invalid dependencies. For example, if a table that some view depends on is dropped, it's still possible to get the view's `CREATE` statement. Without the flag, `GetSchema` errors because it cannot determine columns/fields.

It's a bit unfortunate that so many files have to be affected by the addition of this command line flag, but c'est la vie.

## Checklist


go/cmd/vtctldclient/command/schema.go
```diff
@@ -301,6 +303,7 @@ func init() {
 	GetSchema.Flags().BoolVar(&getSchemaOptions.IncludeViews, "include-views", false, "Includes views in the output in addition to base tables.")
 	GetSchema.Flags().BoolVarP(&getSchemaOptions.TableNamesOnly, "table-names-only", "n", false, "Display only table names in the result.")
 	GetSchema.Flags().BoolVarP(&getSchemaOptions.TableSizesOnly, "table-sizes-only", "s", false, "Display only size information for matching tables. Ignored if --table-names-only is set.")
+	GetSchema.Flags().BoolVarP(&getSchemaOptions.TableSchemaOnly, "table-schema-only", "", false, "Display only table schemas in the result, skip columns and fields.")
```
dbussink, line 306: Maybe something like `skip including columns and fields metadata`? Now it reads as if there's maybe no column information at all, not even in the `CREATE` statements either.

go/vt/mysqlctl/schema.go
```diff
@@ -254,7 +255,7 @@ func (mysqld *Mysqld) normalizedSchema(ctx context.Context, dbName, tableName, t
 // ResolveTables returns a list of actual tables+views matching a list
 // of regexps
 func ResolveTables(ctx context.Context, mysqld MysqlDaemon, dbName string, tables []string) ([]string, error) {
-	sd, err := mysqld.GetSchema(ctx, dbName, tables, nil, true)
+	sd, err := mysqld.GetSchema(ctx, dbName, tables, nil, true, false)
```
dbussink, line 258: I think this can be `true` here as the columns information is not used in this function? I think this applies in more places than just here?

shlomi-noach: Not sure? This function is called by vreplication here:

https://github.com/vitessio/vitess/blob/bc2f459cd175e5f52d97f1b52244019b4ae03e3b/go/vt/vttablet/tabletmanager/vreplication/controller.go#L229-L235

And this goes deep, but normally vreplication is interested in column/field information; I'd need to dig to verify whether it does or does not, but it feels like a risk.

go/cmd/vtctldclient/command/schema.go
shlomi-noach: updated

shlomi-noach:
It seems like in unit tests the flag does not apply correctly:
```
--- FAIL: TestGetSchema (0.01s)
    --- FAIL: TestGetSchema/table_schema_only (0.00s)
        server_test.go:4718: []: (-want +got)
              (*vtctldata.GetSchemaResponse)(
            - 	s`schema:{database_schema:"CREATE DATABASE vt_testkeyspace" table_definitions:{name:"t1" schema:"CREATE TABLE t1 (\n\tid int(11) not null,\n\tPRIMARY KEY (id)\n);" type:"BASE" data_length:100 row_count:50}}`,
            + 	s`schema:{database_schema:"CREATE DATABASE vt_testkeyspace" table_definitions:{name:"t1" schema:"CREATE TABLE t1 (\n\tid int(11) not null,\n\tPRIMARY KEY (id)\n);" columns:"id" type:"BASE" data_length:100 row_count:50 fields:{name:"id" type:INT32}}}`,              )
```

I'm having hard time debugging this. As far as I can see, the execution/invocation path is correct.

go/vt/mysqlctl/schema.go
dbussink: @shlomi-noach `ResolveTables` returns a list of the tables, it gets the `Name` from each entry in `TableDefinitions` so that wouldn't include any column information? And that then is returned, so from what I see in the code all the rest of the data from the schema is thrown away?

shlomi-noach: @dbussink you're right. I see it now. Changed to `true` with comment.

dbussink (approved):

doc/releasenotes/15_0_0_release_notes.md
```diff
@@ -5,6 +5,10 @@
 
 ### New command line flags and behavior
 
+#### vtctl GetSchema --table_schema_only
```
ajm188, line 8: can you make this `--table-schema-only` please? ❤️ (it'll make flag standardizing less work for us, which we're coming to next)

ajm188:
> It seems like in unit tests the flag does not apply correctly:
> 
> ```
> --- FAIL: TestGetSchema (0.01s)
>     --- FAIL: TestGetSchema/table_schema_only (0.00s)
>         server_test.go:4718: []: (-want +got)
>               (*vtctldata.GetSchemaResponse)(
>             - 	s`schema:{database_schema:"CREATE DATABASE vt_testkeyspace" table_definitions:{name:"t1" schema:"CREATE TABLE t1 (\n\tid int(11) not null,\n\tPRIMARY KEY (id)\n);" type:"BASE" data_length:100 row_count:50}}`,
>             + 	s`schema:{database_schema:"CREATE DATABASE vt_testkeyspace" table_definitions:{name:"t1" schema:"CREATE TABLE t1 (\n\tid int(11) not null,\n\tPRIMARY KEY (id)\n);" columns:"id" type:"BASE" data_length:100 row_count:50 fields:{name:"id" type:INT32}}}`,              )
> ```
> 
> I'm having hard time debugging this. As far as I can see, the execution/invocation path is correct.

It's coming from [this mock response](https://github.com/vitessio/vitess/blob/main/go/vt/vtctl/grpcvtctldserver/server_test.go#L4533-L4563), but I'm not not 100% sure since I don't see the test you added in the PR diff 🤔

shlomi-noach:
> It's coming from [this mock response](https://github.com/vitessio/vitess/blob/main/go/vt/vtctl/grpcvtctldserver/server_test.go#L4533-L4563), but I'm not not 100% sure since I don't see the test you added in the PR diff thinking

Thank you @ajm188 ! I removed said test in https://github.com/vitessio/vitess/pull/10552/commits/306f0190c71db1c327e5f5c5fa7d5d13c87cd362

ajm188:
> I removed said test in https://github.com/vitessio/vitess/commit/306f0190c71db1c327e5f5c5fa7d5d13c87cd362

ahh! yeah if you wanted, i would make an entirely separate test and use a separate mock (the `tmc` is singleton-by-design which makes this unfortunately more annoying to work-around 😞 )

shlomi-noach:
> if you wanted, i would make an entirely separate test

I think a mock isn't what we need here. The flag I've introduced changes (removes) the query on the database side. Thus, you'd need to mock a database to test it within this context. The other flags behave differently: they have no impact on the queries, they wait for the full result to return, and then trim it down. They're more compatible with a mock test.

If anything, we can get this in a followup PR. Thank you!

PR #10554, opened 2022-06-21
[14.0] fix: aggregation empty row on join with grouping and aggregations

harshit-gangal:
## Description

Backport of https://github.com/vitessio/vitess/pull/10480

This PR fixes an aggregation-grouping bug in gen4 that does non-scalar aggregates on a top of a join. 
This happens when the left side of the join has data and no data is present on the right hand side.
If the right hand side of the join only contains an aggregation then even though the table has no data it will return count(*) as `0` and that will get mapped to the join output.
To avoid such scenario the change adds a grouping condition on the right hand side using literal 1 so that the rows from right hand side is only returned when the join condition matches.

## Checklist


frouioui (approved):

systay (approved):

deepthi:
Not sure why DCO is stalled. I've manually verified that the commit has been signed off.

Issue #7335, opened 2021-01-21
Feature Request: Incorporate more VTExplain-style information into EXPLAIN FORMAT=vitess


jmoldow:
#### Feature Description

VTExplain, particularly with `-output-mode json`, provides a rich set of information that isn't available via `EXPLAIN FORMAT=vitess`. Placing more of that information into the `EXPLAIN FORMAT=vitess` (or into a new `EXPLAIN` format; or optionally enabled with an additional parameter to `EXPLAIN FORMAT=vitess`) would increase the utility of `EXPLAIN FORMAT=vitess`, and decrease the amount of work needed to expose that information.

I'm open to other suggestions here. This also isn't an important request. Just seems like it'd be nice-to-have for the ecosystem.

This complements #7333 . Neither feature is an important request. And there is a lot of overlap between them, so if we do one, we don't have to do both. On the other hand, each one also supports use-cases that aren't supported by the other, so they can both be implemented without being a waste of effort.

This would also be complemented nicely by #4448 / #7273. In VTExplain, if you have a query that utilizes a lookup vindex, the expansion of the lookup table query will show the correct shard (since it is based on the static hash value), but the expansion of the final query will show an arbitrary shard, since VTExplain cannot actually perform the lookup. One advantage that `EXPLAIN FORMAT=vitess` has in this case over the proposal in #7333, is that with #4448 / #7273, `EXPLAIN` could actually perform the lookup in step 1, and then display the actual target shards for step 2 (or the absence of target shards, if the lookup doesn't find a match).

#### Use Case(s)

Seeing the query plans of live queries in the live vschema, made easily available to all service owners. For example, to see the vindex that gets utilized by a query, and to see any lookup vindex table queries that are auto-generated due to the selection of a lookup vindex, and to see which shards are targeted by the query, etc.

harshit-gangal:
PR: https://github.com/vitessio/vitess/pull/10556

Issue #4448, opened 2018-12-12
vexplain statement


acharis:
it would be great if developers could get the vitess execution plan for a statement.
this is different from vtexplain because it's running against a real live vitess installation.
we should probably also include the mysql explain results as well.

note that i'm thinking this is v-explain, not vex-plain

derekperkins:
@acharis I've been wanting the exact same thing and have been meaning to post an issue. The workflow of going to vtgate output with query plans / stats and then hunting for your query is very tedious. This would be fantastic.

Implementation wise, I don't think we necessarily need to put a ton of work into formatting the output. I'd be happy with a giant dump into a single cell like `SHOW ENGINE INNODB STATUS`.

demmer:
Doing this in a way that just dumps out some representation of the vtgate query plan would be pretty trivial.

But this wouldn't _actually_ tell you what mysql queries would execute, since the vttablet itself has its own query planner and rewriter. A slightly more advanced implementation would augment the planner / executor with an `explain` option so that the queries would be routed to each tablet as per usual, where they would be query planned but not executed, more like what vtexplain does. Now, for complicated queries (lookup vindex, cross-shard joins, sequences, etc), the actual vtgate behavior depends on the contents of subqueries so this becomes harder to model without fake rows coming back...

So probably just dumping the vtgate query plan makes the most sense.

dweitzman:
Might be useful to have a way to run a query and get back a log of what shards & tablets it routed to as it executed.

derekperkins:
Currently I just have a hacked together command that executes queries on all shards if I need to find out which shard a row lives on. This would be very helpful.

Similarly, I would also like to have some shard `GROUP BY` commands, so I could run queries like `select VITESS_SHARD, count(*) from tbl`, but that's for another issue.

demmer:
++ To that idea -- we talked about similar things at some point ages ago.

Exposing the shard and/or keyspace id as a queryable primitive would also enable cross-shard paginated queries in ways that are inefficient today.

harshit-gangal:
There is initial code to provide vitess plans with explain statement #6110

PR #10556, opened 2022-06-21
New explain format: VTEXPLAIN

systay:
## Description
Allows you to run a query from a SQL connection to vitess and then find out what was actually sent to the tablets for your query.

```sql
mysql> explain format=vtexplain insert into user (id,lookup,lookup_unique) values (1,'apa','apa'),(2,'apa','bandar'),(3,'monkey','monkey');
+------+----------+-------+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| #    | keyspace | shard | query                                                                                                                                                                                                  |
+------+----------+-------+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
|    0 | ks       | -40   | begin                                                                                                                                                                                                  |
|    0 | ks       | -40   | insert into lookup(lookup, id, keyspace_id) values ('apa', 1, 'k@�J�K�'), ('apa', 2, '��\"Βp�') on duplicate key update lookup = values(lookup), id = values(id), keyspace_id = values(keyspace_id)        |
|    1 | ks       | 40-80 | begin                                                                                                                                                                                                  |
|    1 | ks       | 40-80 | insert into lookup(lookup, id, keyspace_id) values ('monkey', 3, 'N��ɢ��') on duplicate key update lookup = values(lookup), id = values(id), keyspace_id = values(keyspace_id)                            |
|    2 | ks       | -40   | commit                                                                                                                                                                                                 |
|    3 | ks       | 40-80 | commit                                                                                                                                                                                                 |
|    4 | ks       | 40-80 | begin                                                                                                                                                                                                  |
|    4 | ks       | 40-80 | insert into lookup_unique(lookup_unique, keyspace_id) values ('monkey', 'N��ɢ��')                                                                                                                         |
|    5 | ks       | -40   | begin                                                                                                                                                                                                  |
|    5 | ks       | -40   | insert into lookup_unique(lookup_unique, keyspace_id) values ('apa', 'k@�J�K�'), ('bandar', '��\"Βp�')                                                                                                     |
|    6 | ks       | 40-80 | commit                                                                                                                                                                                                 |
|    7 | ks       | -40   | commit                                                                                                                                                                                                 |
|    8 | ks       | 40-80 | begin                                                                                                                                                                                                  |
|    8 | ks       | 40-80 | insert into `user`(id, lookup, lookup_unique) values (3, 'monkey', 'monkey')                                                                                                                           |
|    9 | ks       | -40   | begin                                                                                                                                                                                                  |
|    9 | ks       | -40   | insert into `user`(id, lookup, lookup_unique) values (1, 'apa', 'apa'), (2, 'apa', 'bandar')                                                                                                           |
|   10 | ks       | 40-80 | commit                                                                                                                                                                                                 |
|   11 | ks       | -40   | commit                                                                                                                                                                                                 |
+------+----------+-------+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
18 rows in set (0.03 sec)
```

Since `explain format=vtexplain` actually runs queries, and this is not the normal behavior for `EXPLAIN`, we protect users from accidentally running queries by forcing a comment directive if the query is a DML query.

Example:
```sql
mysql> explain format=vtexplain insert into tabl values (1,2,3);
ERROR 12345 (23000): explain format = vtexplain will actually run queries. `/*vt+ ACTUALLY_RUN_QUERIES */` must be set be added to run DML queries in vtexplain. Example: `explain /*vt+ ACTUALLY_RUN_QUERIES */ format = vtexplain delete from t1`

mysql> explain /*vt+ actually_run_queries */ format=vtexplain insert into tabl values (1,2,3);

+------+----------+-------+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| #    | keyspace | shard | query                                                                                                                                                                                                  |
+------+----------+-------+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
|    0 | ks       | -40   | begin                                                                                                                                                                                                  | 
|  ... |          |       |                                                                                                                                                                                                        | 
+------+----------+-------+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
```

Documentation PR can be found [here](https://github.com/vitessio/website/pull/1118).

## Related Issue(s)

<!-- List related issues and pull requests. If this PR fixes an issue, please add it using Fixes #????  -->

- https://github.com/vitessio/vitess/issues/4448
- https://github.com/vitessio/vitess/issues/7335

## Checklist

## Deployment Notes

<!-- Notes regarding deployment of the contained body of work. These should note any db migrations, etc. -->


systay:
One more feature needed before we merge this: We want to add GRPC counters to the calls, so the user can see that some calls go together - BEGIN is never sent on it's own, for example. So each grpc call between the vtgate and the vttablets needs to get logged and numbered

go/vt/sqlparser/comments.go
```diff
@@ -42,6 +42,8 @@ const (
 	DirectiveAllowHashJoin = "ALLOW_HASH_JOIN"
 	// DirectiveQueryPlanner lets the user specify per query which planner should be used
 	DirectiveQueryPlanner = "PLANNER"
+	// DirectiveActuallyRunQueries tells explain format = vtexplain that it is okay to also run the query.
+	DirectiveActuallyRunQueries = "ACTUALLY_RUN_QUERIES"
```
harshit-gangal, line 46: I think this should be applicable only for DML queries.
If that is the case we can call it `EXECUTE_DML_QUERIES`

go/vt/sqlparser/comments.go
```diff
@@ -258,8 +260,16 @@ func (d CommentDirectives) IsSet(key string) bool {
...
+	var val string
+	exists := false
+	for commentDirective, value := range d {
+		if strings.EqualFold(commentDirective, key) {
+			val = value
+			exists = true
+			break
+		}
+	}
+	if !exists {
```
harshit-gangal, line 272: I think we can avoid this check by setting the map key in one defined case like `lower_case` and also checking the key in `lower_case`.

go/vt/vtgate/safe_session.go
```diff
@@ -22,6 +22,10 @@ import (
 	"sync"
 	"time"
 
+	"vitess.io/vitess/go/vt/sqlparser"
+
+	"vitess.io/vitess/go/vt/vtgate/engine"
+
 	"google.golang.org/protobuf/proto"
 
 	"vitess.io/vitess/go/vt/vterrors"
```
GuptaManan100, line 31: imports for vitess should be all together

systay: I don't look at the imports formatting. IMHO - our formatting rules are to follow the linter. if it doesn't complain, we should fix the linter and not the code changes

It does not complain about these lines

go/vt/sqlparser/comments.go
systay: 👍

doc/releasenotes/15_0_0_summary.md
```diff
@@ -184,4 +185,10 @@ This is different from the existing `autocommit` parameter where the query is se
...
 A new durability policy `cross_cell` is now supported. `cross_cell` durability policy only allows replica tablets from a different cell than the current primary to
-send semi-sync ACKs. This ensures that any committed write exists in at least 2 tablets belonging to different cells.
\ No newline at end of file
+send semi-sync ACKs. This ensures that any committed write exists in at least 2 tablets belonging to different cells.
+
+### New EXPLAIN format
+
+#### FORMAT=vtexplain
+
+With this new `explain` format, you can get an output that is very similar to the command line `vtexplain` app, but from a running `vtgate`, through a MySQL query.
```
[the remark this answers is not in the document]
systay: Nah, this is just meant to be a summary. The error message and docs should be enough

go/vt/sqlparser/comments.go
systay: my thinking is that for the user it's pretty clear already that they are dealing with DML. They'll write 

```sql
EXPLAIN /*vt+ ACTUALLY_RUN_QUERIES */ INSERT INTO tbl...
```

That this is DML is already pretty clear. The `INSERT` says it. So the comment directive should be focus on the running of the query, not the type of query.

¯\_(ツ)_/¯

@deepthi or someone else with opinions?

go/vt/vtgate/engine/vtexplain.go
```diff
@@ -0,0 +1,89 @@
...
+	vcursor.Session().VtExplainLogging()
+	return vcursor.ExecutePrimitive(ctx, v.Input, bindVars, wantfields)
+}
+
+// TryStreamExecute implements the Primitive interface
+func (v *VTExplain) TryStreamExecute(ctx context.Context, vcursor VCursor, bindVars map[string]*querypb.BindVariable, wantfields bool, callback func(*sqltypes.Result) error) error {
+	//TODO implement me
+	panic("implement me")
+}
+
```
harshit-gangal, line 79: we should add support for olap as well.

go/vt/vtgate/executor_test.go
```diff
@@ -2503,6 +2503,18 @@ func TestExecutorDescHash(t *testing.T) {
...
+func TestExecutorVtExplain(t *testing.T) {
+	executor, _, _, sbclookup := createExecutorEnv()
+	session := NewAutocommitSession(&vtgatepb.Session{})
+
+	sbclookup.SetResults([]*sqltypes.Result{
+		sqltypes.MakeTestResult(sqltypes.MakeTestFields("name|user_id", "varchar|int64"), "apa|1", "apa|2"),
+	})
+	qr, err := executor.Execute(ctx, "TestExecutorVtExplain", session, "explain format=vtexplain select * from user where name = 'apa'", nil)
+	require.NoError(t, err)
+	fmt.Printf("%v\n", qr.Rows)
```
harshit-gangal, line 2515: looks like this fmt was for debuging, this may be changed into checking row values.

go/vt/vtgate/planbuilder/show.go
```diff
@@ -603,8 +603,12 @@ func buildWarnings() (engine.Primitive, error) {
 		rows := make([][]sqltypes.Value, 0, len(warns))
 
 		for _, warn := range warns {
+			txt := "Warning"
+			if warn.Code == 1003 {
+				txt = "Note"
+			}
 			rows = append(rows, []sqltypes.Value{
-				sqltypes.NewVarChar("Warning"),
+				sqltypes.NewVarChar(txt),
```
harshit-gangal, line 611: I have two things here.
1. I think we should keep the error code in go/mysql/constants.go
2. Do we ever log 1003? 
I know this will work when we start getting warnings from vttablet, just wanted to understand we are doing any logging with this code

go/vt/vtgate/vcursor_impl.go
```diff
@@ -423,10 +423,15 @@ func (vc *vcursorImpl) StreamExecutePrimitiveStandalone(ctx context.Context, pri
 
 // Execute is part of the engine.VCursor interface.
 func (vc *vcursorImpl) Execute(ctx context.Context, method string, query string, bindVars map[string]*querypb.BindVariable, rollbackOnError bool, co vtgatepb.CommitOrder) (*sqltypes.Result, error) {
+	vc.safeSession.vindexExec = true
+	defer func() {
+		vc.safeSession.vindexExec = false
+	}()
 	session := vc.safeSession
```
harshit-gangal, line 430: similar change should go in StreamExecute

go/vt/vtgate/executor_test.go
systay: 👍

go/vt/sqlparser/comments.go
deepthi: oh, so `explain format=vtexplain ...` will end up inserting the rows. 
My question is.. what happens when it is not a DML? Are we "actually running queries" without the directive?

deepthi (commented):
How does this differ from `explain format=vitess`?
Is that something we plan to cover in the website docs?

systay:
> How does this differ from `explain format=vitess`?

`format=vitess` shows the vtgate query plan in tabular mode. `format=vtexplain` shows the queries sent to the tablets. Docs should definitely show the difference between them.

systay:
Running queries to produce EXPLAIN output is in line with `EXPLAIN ANALYZE` that was introduced in MySQL: https://dev.mysql.com/doc/refman/8.0/en/explain.html#explain-analyze

`EXPLAIN ANALYZE` is only available for `SELECT`, but we also allow it for other query types.

deepthi:
If a user is in an autocommit session and they `explain` a DML, that will be committed, right?
Whereas if they do this within a transaction, they can choose to commit/rollback. In other words, usual vitess/mysql semantics around these modes, but we should probably include it in the docs.
Rest LGTM. This is an awesome improvement that lots of people have asked for.

go/vt/sqlparser/comments.go
systay: 👍

GuptaManan100 (approved):

PR #10394, opened 2022-05-31
OnlineDDL/vrepl: more error codes, spatial types

shlomi-noach:
## Description

This PR introduces several spatial types related errors, which are now deemed as "unrecoverable" `vreplication` errors.
Adding a single spatial casting test (validates one of those error codes)

## Related Issue(s)

- #6926 
- followup to #9973, #10390 

## Checklist

-   [ ] "Backport me!" label has been added if this change should be backported
-   [ ] Tests were added or are not required
-   [ ] Documentation was added or is not required

PR #10559, opened 2022-06-21
VReplication: more unrecoverable error codes

shlomi-noach:
## Description

followup to #9973, #10390, #10394

Adding error codes related to `DATETIME` value conversion/truncation. This will fail a `VReplication` workflow, and specifically an Online DDL `vitess` migration, immediately upon receipt of such error.

## Related Issue(s)

#6926 

## Checklist


dbussink:
@shlomi-noach as a bugfix, would this also be eligible to backport to the 14 release branch as well?

shlomi-noach:
> would this also be eligible to backport to the 14 release branch as well?

I think so!

shlomi-noach:
> but I see some onlineddl workflows are failing.

yes, there is increased flakiness lately, which I don't have good explanation for, yet.

shlomi-noach:
Ohhhh. The `vrepl` tests are failing because I accidentally committed a temp directory into the test suite. Removing it.

PR #10563, opened 2022-06-22
Proto file lint fix and vtadmin generated file

GuptaManan100:
## Description

This PR just fixes a linting issue in a proto file.

## Checklist


web/vtadmin/src/proto/vtadmin.d.ts
```diff
@@ -14536,6 +14536,9 @@ export namespace tabletmanagerdata {
 
         /** GetSchemaRequest exclude_tables */
         exclude_tables?: (string[]|null);
+
+        /** GetSchemaRequest table_schema_only */
+        table_schema_only?: (boolean|null);
```
GuptaManan100, line 14541: These changes have nothing to do with my new line change. I thought we had a workflow to test that vtadmin files are upto-date on the PR being merged. How did this happen? @ajm188

ajm188: my best guess: in #10552 @shlomi-noach ran the generators (and CI passed on his branch). separately, you ran the generators (and CI passed on _your_ branch), but because these two changes didn't step on each other, git happily merged the latter (yours) without making you rebase. contrast this with `protoc-gen-go` which includes that magic byte array that always forces a conflict

ajm188 (approved):

web/vtadmin/src/proto/vtadmin.d.ts
ajm188: well, even simpler explanation, it turns out: 
Screen Shot 2022-06-22 at 6 17 01 AM

PR #10571, opened 2022-06-22
VStream API: Fix vtgate memory leaks when context gets cancelled

rohit-nayak-ps:
## Description

This PR fixes a couple of memory leaks in the VStream API which was causing vtgates to OOM.

* Check for cancelled context when events received from the source were being sent into a event channel for eventual streaming to the vstream api client.
* When a context is cancelled, in the vstream API we need to return an error in the health streamer callback. Otherwise it keeps streaming and leaking (goroutines and ) grpc request buffers.
 
## Checklist


mattlord (approved):

deepthi:
We've already included this in v13.0.2. Based on the report in [slack](https://vitess.slack.com/archives/C0PQY0PTK/p1659068255242519) we should back port it to release-12.0 as well.

PR #10577, opened 2022-06-24
Insert with select using streaming call

harshit-gangal:
## Description

This PR does enhancement on the earlier support of insert with select by using the streaming calls for select and there by breaking the insert calls to the tablet into batches. This will also lead to less memory requirement at vttablet and vtgate.

## Checklist

-   [ ] "Backport me!" label has been added if this change should be backported - Not sure


go/vt/vtgate/engine/insert.go
```diff
@@ -243,11 +244,57 @@ func (ins *Insert) TryExecute(vcursor VCursor, bindVars map[string]*querypb.Bind
 
 // TryStreamExecute performs a streaming exec.
 func (ins *Insert) TryStreamExecute(vcursor VCursor, bindVars map[string]*querypb.BindVariable, wantfields bool, callback func(*sqltypes.Result) error) error {
-	res, err := ins.TryExecute(vcursor, bindVars, wantfields)
+	if ins.Opcode != InsertSelect {
+		res, err := ins.TryExecute(vcursor, bindVars, wantfields)
```
systay, line 248: Shouldn't primitive to primitive calls always go through the vcursor?

harshit-gangal: earlier was not needed as the call came through vcursor itself and was delegated, here may be the case.

go/test/endtoend/vtgate/queries/insert/main_test.go
```diff
@@ -97,6 +99,7 @@ func TestMain(m *testing.M) {
 		}
 
 		// Start vtgate
+		clusterInstance.VtGatePlannerVersion = planbuilder.Gen4
```
frouioui, line 102: With `gen4` being the new default planner, do we need this line?

go/vt/vtgate/planbuilder/insert.go
```diff
@@ -209,6 +210,10 @@ func buildInsertSelectPlan(ins *sqlparser.Insert, table *vindexes.Table, reserve
 	}
 	eins.Input = plan
 
+	if strings.Contains(plan.GetTableName(), table.Name.String()) {
+		eins.ForceNonStreaming = true
+	}
```
frouioui, line 215: We should add a comment here

harshit-gangal: I added the comment just above this condition.

harshit-gangal: just realised I did not push that.

go/test/endtoend/vtgate/queries/insert/main_test.go
frouioui: Right, makes sense!

frouioui (approved):
LGTM, left a nit to comment on a piece of code.

PR #10578, opened 2022-06-26
Cleanup: un-explode GetSchema and reuse GetSchemaRequest struct

shlomi-noach:
## Description

This is a code cleanup PR, following up on Vitess Office Hours from `2022-06-23`.

We "unexplode" the `GetSchema` function signature, across the board, and reuse `tabletmanagerdatapb.GetSchemaRequest` as a replacement container for the multiple arguments it receives. Example:

```go
GetSchema(ctx context.Context, request *tabletmanagerdatapb.GetSchemaRequest) (*tabletmanagerdatapb.SchemaDefinition, error)
```

In some places we now just need to pass the request around. In others (notably `vdiff` related code) we programmatically construct a `tabletmanagerdatapb.GetSchemaRequest` instance and send it to `GetSchema`.

I think this went well. The code is simpler; when constructing `tabletmanagerdatapb.GetSchemaRequest`, we only point out the non-default variables, which is very clear. In the future, adding more fields to `tabletmanagerdatapb.GetSchemaRequest` will be simple and painless.

## Related Issue(s)

Related to https://github.com/vitessio/vitess/pull/10552, where we recently added a new variable.

Tests were adapted to new function signatures, and should otherwise work.

## Checklist

### DIFF
--- a/go/vt/mysqlctl/fakemysqldaemon/fakemysqldaemon.go
+++ b/go/vt/mysqlctl/fakemysqldaemon/fakemysqldaemon.go
@@ -543,14 +543,14 @@ func (fmd *FakeMysqlDaemon) CheckSuperQueryList() error {
 }
 
 // GetSchema is part of the MysqlDaemon interface
-func (fmd *FakeMysqlDaemon) GetSchema(ctx context.Context, dbName string, tables, excludeTables []string, includeViews bool, tableSchemaOnly bool) (*tabletmanagerdatapb.SchemaDefinition, error) {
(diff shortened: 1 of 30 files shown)


deepthi (approved):
This also makes the code more readable. No more argument lists like `nil, nil, false, false` with no idea what those mean.
I don't remember who suggested this, maybe @ajm188 but I like the way it turned out.
Just out of curiosity, how many files can we avoid touching the next time around?

shlomi-noach:
> Just out of curiosity, how many files can we avoid touching the next time around?

I expect 3-4 files, excluding any specific testing.

PR #10581, opened 2022-06-26
add vtadmin web files to all lite images

deepthi:
## Description
In #9405 we added vtadmin web files (not binaries) to only the mysql57 version of `vitess/lite`. This PR adds them to all flavors of the `lite` image.

## Related Issue(s)
#9405 #10543

## Checklist


deepthi:
This needs to be forward-ported to `main` after merge.

frouioui (approved):

PR #10589, opened 2022-06-27
Backport: VDiff2: Support Resuming VDiffs

mattlord:
## Description

This is a backport of https://github.com/vitessio/vitess/pull/10497. While the primary intention of that work was to add a new feature — namely `Resume` — in the process of implementing that it became clear that [the existing VDiff2 work in 14.0](https://github.com/vitessio/vitess/pull/10382) could easily produce incorrect results and was not easy to validate or debug. There were also a number of other smaller/minor bugs that were fixed in that work. All of this made it very challenging for early adopters to try it out, get value, and provide meaningful feedback (which was the entire reason to get the experimental version into 14.0 in the first place). Because of this, combined with the fact that VDiff2 is experimental and has isolated code, we are backporting this work for 14.0 GA (normally new feature work would only go into main).

## Related Issue(s)

- Cherry-pick from: https://github.com/vitessio/vitess/pull/10497

## Checklist

-   [x] Tests were added
-   [x] Documentation: https://github.com/vitessio/website/pull/1060

### DIFF
--- a/doc/releasenotes/14_0_0_summary.md
+++ b/doc/releasenotes/14_0_0_summary.md
@@ -218,9 +218,12 @@ VDiff bf9dfc5f-e5e6-11ec-823d-0aa62e50dd24 scheduled on target shards, use show 
 
 $ vtctlclient --server=localhost:15999 VDiff -- --v2 customer.commerce2customer show last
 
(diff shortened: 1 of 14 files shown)


rohit-nayak-ps (approved):
lgtm

PR #10591, opened 2022-06-27
[14.0] Schema tracking acl error logging

harshit-gangal:
## Description

This PR log a warning when schema tracking is failing due to permission issues for user to see and correct the vtgate deployment for schema tracking to work.

## Checklist

### DIFF
--- a/go/test/endtoend/vtgate/schematracker/unauthorized/unauthorized_test.go
+++ b/go/test/endtoend/vtgate/schematracker/unauthorized/unauthorized_test.go
@@ -1,0 +1,108 @@
+/*
+Copyright 2022 The Vitess Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package unauthorized
+
+import (
+	"context"
+	_ "embed"
+	"flag"
+	"os"
+	"path"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+
+	"vitess.io/vitess/go/mysql"
+	"vitess.io/vitess/go/test/endtoend/cluster"
+)
+
+var (
+	clusterInstance *cluster.LocalProcessCluster
+	vtParams        mysql.ConnParams
+	KeyspaceName    = "ks"
+	Cell            = "test"
+
+	//go:embed schema.sql
+	SchemaSQL string
+
+	//go:embed vschema.json
+	VSchema string
+)
+
+func TestMain(m *testing.M) {
+	defer cluster.PanicHandler(nil)
+	flag.Parse()
+
+	exitCode := func() int {
+		clusterInstance = cluster.NewCluster(Cell, "localhost")
+		defer clusterInstance.Teardown()
+
+		// Start topo server
+		err := clusterInstance.StartTopo()
+		if err != nil {
+			return 1
+		}
+
+		// Start keyspace
+		keyspace := &cluster.Keyspace{
+			Name:      KeyspaceName,
+			SchemaSQL: SchemaSQL,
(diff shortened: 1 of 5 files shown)


go/vt/vtgate/schema/tracker.go
```diff
@@ -136,6 +139,10 @@ func (t *Tracker) initKeyspace(th *discovery.TabletHealth) error {
 	err := t.LoadKeyspace(th.Conn, th.Target)
 	if err != nil {
 		log.Warningf("Unable to add the %s keyspace to the schema tracker: %v", th.Target.Keyspace, err)
+		code := vterrors.Code(err)
+		if code == vtrpcpb.Code_UNAUTHENTICATED || code == vtrpcpb.Code_PERMISSION_DENIED {
+			log.Warning("Table ACL might be enabled, --schema_change_signal_user needs to be passed to VTGate for schema tracking to work. More details on vitess.io")
```
deepthi, line 144: Instead of `More details on vitess.io`, we should say `See <link>` and link to the docs page.
Of course we'll need to update that link in every release 😓

go/vt/vtgate/schema/tracker.go
```diff
@@ -181,6 +188,10 @@ func (t *Tracker) updateSchema(th *discovery.TabletHealth) bool {
 		t.tracked[th.Target.Keyspace].setLoaded(false)
 		// TODO: optimize for the tables that got errored out.
 		log.Warningf("error fetching new schema for %v, making them non-authoritative: %v", tablesUpdated, err)
+		code := vterrors.Code(err)
+		if code == vtrpcpb.Code_UNAUTHENTICATED || code == vtrpcpb.Code_PERMISSION_DENIED {
+			log.Warning("Table ACL might be enabled, --schema_change_signal_user needs to be passed to VTGate for schema tracking to work. More details on vitess.io")
```
deepthi, line 193: See previous comment.

deepthi (approved):
Can you change the error message according to the feedback prior to merge?

go/vt/vtgate/schema/tracker.go
harshit-gangal: for the same reason I did not give full link

go/test/endtoend/vtgate/schematracker/unauthorized/unauthorized_test.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+func TestVschemaInfo(t *testing.T) {
+	ctx := context.Background()
+	conn, err := mysql.Connect(ctx, &vtParams)
+	require.NoError(t, err)
+	defer conn.Close()
+
+	// wait for some time to try loading schema.
+	time.Sleep(1 * time.Minute)
```
systay, line 96: This does not make me happy . That is a very long timeout

systay:
Should this be forward-merged into main? Are we not supposed to do backports?

go/test/endtoend/vtgate/schematracker/unauthorized/unauthorized_test.go
harshit-gangal: was bring it down to 30s

```
=== RUN   TestSchemaTrackingError
--- PASS: TestSchemaTrackingError (30.02s)
PASS

```

systay (approved):

go/vt/vtgate/schema/tracker.go
systay: We should really start using error codes. Makes it very easy to search for the more info on specific errors, and is easier to manage on both ends.

This is how Oracle does it:
https://docs.oracle.com/cd/E11882_01/server.112/e17766/toc.htm

harshit-gangal: how about `Check 'schema tracking' docs on vitess.io`

harshit-gangal: I agree, we can take this up for 15.0 but I like to avoid putting link in the error message inside the code.
should I go ahead with this change?

systay: sure, this should work as a stop-gap measure, IMHO. Let's leave it to @deepthi to make a call on it

harshit-gangal:
> Should this be forward-merged into main? Are we not supposed to do backports?

the dicussion was to do forward port to avoid backport time.

go/vt/vtgate/schema/tracker.go
deepthi: that's fine for now.

PR #10593, opened 2022-06-28
Add support for INSERT() string function

dbussink:
This adds explicit parsing support for the `INSERT()` string function. This isn't parsed today by the generic function parsing logic since `INSERT` is a keyword.
Ref - https://dev.mysql.com/doc/refman/8.0/en/string-functions.html#function_insert
## Related Issue(s)

#10203
#8604

## Checklist


systay:
Does this mean that there are other functions hiding behind reserved keywords that we are missing? Should this be a more general solution for these? @GuptaManan100?

GuptaManan100 (commented):
I'll just add a planner test. This looks good to me overall!

GuptaManan100:
@systay I don't think we can have a more general solution since they are reserved keywords. If we add a generic rule for reserved_keyword, it is going to result in a bunch of shift-reduce conflicts which aren't going to be just noisy but result in incorrect behaviour too.

GuptaManan100 (approved):
This looks good to me!

PR #10596, opened 2022-06-28
fix: Allow gen4 to handle variables on authoritative tables

systay:
## Description
In V15, this has been fixes in a [different way](https://github.com/vitessio/vitess/pull/10547) but for the backported fix, we want to change as little as possible.

This change allows Gen4 to handle these queries, even if V3 still fails.

## Checklist


systay:
It's part of #10595

PR #10599, opened 2022-06-28
Parse INTERVAL() function

dbussink:
This adds support for parsing the INTERVAL() function. It adds a single shift / reduce conflict but the behavior there is what we want to match the MySQL behavior as wel too.
Ref - https://dev.mysql.com/doc/refman/8.0/en/comparison-operators.html#function_interval

## Related Issue(s)

#10203
#8604 

## Checklist


dbussink:
> would you mind adding a couple of tests to `go/vt/sqlparser/parse_test.go`?

@systay Can certainly do that, but since it was already covered a bunch in `go/vt/sqlparser/testdata/select_cases.txt` I didn't immediately add things there too.

systay (approved):

GuptaManan100 (approved):
This looks good to me too!

PR #10601, opened 2022-06-28
VReplication: throttling info for both source and target; Online DDL propagates said info

shlomi-noach:
## Description

This PR is primarily an enhancement to VReplication, adding visibility into throttling status of an active workflow.

In addition, Online DDL now utilizes said information and presents it as part of `SHOW VITESS_MIGRATIONS`.

Lastly, there's a deadlock solution to a particular throttling scenario.

### Throttling visibility in VReplication

The following components in VReplication consult the throttler (way before this PR):

- `vstreamer`, on source tablet
- `rowstreamer`, on source tablet
- `vplayer`, on target tablet
- `vcopier`, on target tablet

As of this PR, all four further advertise their throttling state. The table `_vt.vreplication` now has two new columns:

- `time_throttled` (unix time, seconds)
- `component_throttled` (name of component)

These present the last known incident where a VReplication component was throttled. Example values might be:
```
             time_throttled: 1656420364
        component_throttled: vcopier
```

For `vplayer` and `vcopier`, whenever they are throttled, they now ask `vreplicator` to update those values, e.g. via:
```go
vr.updateTimeThrottled("vplayer")
```

`vreplicator` rate-limits those writes, to at most one per second. It's fine if calls to `updateTimeThrottled()` are frequent. Some calls may be dropped. The database will only be affected at most once per second.

As for `vstreamer` and `rowstreamer`, they don't have immediate access to `_vt.vreplication` or to `vreplicator`, because they are on the source tablet. A new `bool throttled` proto field is added in both `VEvent` and `VStreamRowsResponse`. If `vstreamer` is throttled, it sends (rate limited) heartbeat events with `Throttled: true`. If `VStreamRowsResponse` is throttled, it sends (rate limited) responses where `Throttled: true`. `vplayer` and `vcopier` receive and identify those responses, respectively, and call `updateTimeThrottled()`.

This means we can see in `_vt.vreplication` the last known throttling incident: when it happened, and which component was affected, one of `vstreamer, rowstreamer, vplayer, vcopier`.

This information does _not_ include frequency of throttling, number of successful/rejected throttler checks, etc. We leave that to a future PR.

### Online DDL throttling info

When reviewing running migrations, and for `vitess` migration, the online DDL executor now reads the above two columns, and propagate them as two new columns in `_vt.schema_migrations` table:

- `last_throttled_timestamp` (`TIMESTAMP`)
- `component_throttled` (textual)

These are only sampled once per minute, sometimes more frequently.

### A deadlock scenario, now solved

With the introduction of [on demand heartbeats](https://github.com/vitessio/vitess/pull/10198), we've introduced a throttling deadlock scenario. Consider the following scenario (Copy of comment from this PR):

- a vitess migration
- with on demand heartbeats
- the streamer running on a replica
- the streamer was throttled for long enough
- then vplayer and vcopier are blocked, waiting for the streamer to do something
- since they are blocked, they're not running throttler checks
- since streamer runs on replica, it only checks that replica
- therefore no one asking for on-demand heartbeats
- then, if the conditions for the streamer's throttling are done, the streamer then thinks there's replication lag, with nothing to remediate it.
- it's a deadlock.

And so, once per `reviewRunningMigrations()`, and assuming there _are_ running migrations, we ensure to hit a throttler check. This will kick on-demand heartbeats, unlocking the deadlock.

### rowstreamer heartbeats

`rowstreamer` now sends heartbeats every `10sec`. This means even if it's blocked on reading the next batch of rows (blocked, slow, hangs, for whatever reason), it still reports heartbeats. these are intercepted by `vcopier` which then updates `_vt.vreplication.time_heartbeat`.

## Related Issue(s)

#6926 
#10198
 
## Checklist


shlomi-noach:
I forgot to mention. One of the side effects to this PR, which is actually the root trigger to working on it:

Without this PR, a VReplication workflow that was throttled, would be identified by the Online DDL executor as "stale" after 10 minutes of inactivity, and terminated.

Now, with this PR, this does not happen because `time_updated`  in `_vt.vreplication` is updated whenever a throttling is identified. Thus, Online DDL executor keeps seeing the migration as "fresh", and does not terminate it.

shlomi-noach:
Looking into unit test failures, which seem to be related to the `RateLimiter`.

shlomi-noach:
Obviously only failing in GitHub CI :stuck_out_tongue:  and not on local envs.

shlomi-noach:
Added the tow new columns in `ReplicationStatusResult`, which then shows up in `vtctl Workflow ... Show`

shlomi-noach:
Also added:

`rowstreamer` now sends heartbeats every `10sec`. This means even if it's blocked on reading the next batch of rows (blocked, slow, hangs, for whatever reason), it still reports heartbeats. these are intercepted by `vcopier` which then updates `_vt.vreplication.time_heartbeat`.

shlomi-noach:
Fixed all unit tests, new ones showing! Still working through them.

shlomi-noach:
Phew, the tests were brutal but legit! Now good to go!

go/vt/vttablet/onlineddl/vrepl.go
```diff
@@ -60,13 +62,12 @@ type VReplStream struct {
...
 
 // livenessTimeIndicator returns a time indicator for last known healthy state.
-// vreplication uses two indicators: time_updates and time_heartbeat. Either one making progress is good news. The greater of the two indicates the
-// latest progress. Note that both indicate timestamp of events in the binary log stream, rather than time "now".
-// A vreplication stream health is determined by "is there any progress in either of the two counters in the past X minutes"
+// vreplication uses three indicators:
+// - transaction_timestamp
+// - time_heartbeat
+// - time_throttled.
+// Update of any of them, also updates time_updated, indicating liveness.
```
mattlord, line 69: `Updating` any of them, right?

go/vt/vttablet/tabletmanager/vreplication/vcopier.go
```diff
@@ -238,9 +238,19 @@ func (vc *vcopier) copyTable(ctx context.Context, tableName string, copyState ma
 				return io.EOF
 			default:
 			}
+			if rows.Throttled {
+				_ = vc.vr.updateTimeThrottled("rowstreamer")
```
mattlord, line 242: I think this could be a resultstreamer too, but this is precise enough (I'm thinking of Materialization and VDiff1), but I could be wrong too). If that's true, and we care to distinguish them here, we could infer this from workflow type etc.

go/vt/vttablet/tabletmanager/vreplication/vreplicator.go
```diff
@@ -500,6 +505,32 @@ func (vr *vreplicator) throttlerAppName() string {
 	return strings.Join(names, ":")
 }
 
+func (vr *vreplicator) updateTimeThrottled(componentThrottled string) error {
```
mattlord, line 508: I feel we it may be worth having constants defined for the components, e.g. `vreplication.Compenent` type (alias for string) with a value of e.g. `vreplication.VPlayerComponent` being a const string for "vplayer"? Strings introduces the opportunity for typos, invalid components, etc.

go/vt/vttablet/tabletserver/vstreamer/vstreamer.go
```diff
@@ -268,24 +269,36 @@ func (vs *vstreamer) parseEvents(ctx context.Context, events <-chan mysql.Binlog
...
+	hbTimer := time.NewTimer(HeartbeatTime)
+	defer hbTimer.Stop()
+
+	injectHeartbeat := func(throttled bool) error {
+		now := time.Now().UnixNano()
+		err := bufferAndTransmit(&binlogdatapb.VEvent{
+			Type:        binlogdatapb.VEventType_HEARTBEAT,
+			Timestamp:   now / 1e9,
+			CurrentTime: now,
+			Throttled:   throttled,
```
mattlord, line 281: I'm curious why we get a nanosecond precision timestamp only to turn it back into a second precision timestamp for the protobuf message?

mattlord (commented):
I like it! This will greatly improve the observability/debugability of VReplication workflows (especially when we enable tablet throttling by default). Thank you for working on it!

I had a few questions and comments, but nothing major. I'll look out for your responses and can approve quickly if needed.

@rohit-nayak-ps if you have time, would you mind giving it a pass too? Thanks!

go/vt/vttablet/tabletmanager/vreplication/vcopier.go
rohit-nayak-ps: > I think this could be a resultstreamer too, but this is precise enough (I'm thinking of Materialization and VDiff1), but I could be wrong too). If that's true, and we care to distinguish them here, we could infer this from workflow type etc.

Only VDiff1 uses resultstreamer and it doesn't use the vcopier, so this is good

rohit-nayak-ps (approved):
lgtm

go/vt/vttablet/tabletserver/vstreamer/vstreamer.go
shlomi-noach: I'm not sure, this block of code is moved from elsewhere and otherwise untouched.

go/vt/binlog/binlogplayer/binlog_player.go
```diff
@@ -568,6 +568,8 @@ var AlterVReplicationTable = []string{
 	// records the time of the last heartbeat. Heartbeats are only received if the source has no recent events
 	"ALTER TABLE _vt.vreplication ADD COLUMN time_heartbeat BIGINT(20) NOT NULL DEFAULT 0",
 	"ALTER TABLE _vt.vreplication ADD COLUMN workflow_type int NOT NULL DEFAULT 0",
+	"ALTER TABLE _vt.vreplication ADD COLUMN time_throttled BIGINT(20) NOT NULL DEFAULT 0",
```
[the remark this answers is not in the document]
shlomi-noach: Absolutely right. Too many copy_pastes.

go/vt/vttablet/onlineddl/executor.go
```diff
@@ -3222,6 +3227,25 @@ func (e *Executor) reviewRunningMigrations(ctx context.Context) (countRunnning i
...
+					go throttlerOnce.Do(func() {
+						if e.lagThrottler.CheckIsReady() != nil {
+							return
+						}
+						// Self healing: in the following scenario:
+						// - a vitess migration
+						// - with on demand heartbeats
+						// - the streamer running on a replica
+						// - the streamer was throttled for long enough
+						// - then vplayer and vcopier are lbocked, waiting for the streamer to do something
```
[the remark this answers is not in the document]
shlomi-noach: typo for `locaked`. Fixed.

go/vt/binlog/binlogplayer/binlog_player.go
```diff
@@ -684,6 +686,14 @@ func GenerateUpdateHeartbeat(uid uint32, timeUpdated int64) (string, error) {
 	return fmt.Sprintf("update _vt.vreplication set time_updated=%v, time_heartbeat=%v where id=%v", timeUpdated, timeUpdated, uid), nil
 }
 
+// GenerateUpdateTimeThrottled returns a statement to record the latest throttle time in the _vt.vreplication table.
+func GenerateUpdateTimeThrottled(uid uint32, timeThrottled int64, componentThrottled string) (string, error) {
```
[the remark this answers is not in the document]
shlomi-noach: Renamed as `timeThrottledUnix`

go/vt/binlog/binlogplayer/binlog_player.go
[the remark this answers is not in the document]
shlomi-noach: fixed

go/vt/vttablet/onlineddl/schema.go
```diff
@@ -77,6 +77,8 @@ const (
 	alterSchemaMigrationsTableVreplLivenessIndicator   = "ALTER TABLE _vt.schema_migrations add column vitess_liveness_indicator bigint NOT NULL DEFAULT 0"
 	alterSchemaMigrationsTableUserThrottleRatio        = "ALTER TABLE _vt.schema_migrations add column user_throttle_ratio float NOT NULL DEFAULT 0"
 	alterSchemaMigrationsTableSpecialPlan              = "ALTER TABLE _vt.schema_migrations add column special_plan text NOT NULL"
+	alterSchemaMigrationsLastThrottled                 = "ALTER TABLE _vt.schema_migrations add column last_throttled_timestamp timestamp NULL DEFAULT NULL"
+	alterSchemaMigrationsComponentThrottled            = "ALTER TABLE _vt.schema_migrations add column component_throttled tinytext NOT NULL"
```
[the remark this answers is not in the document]
shlomi-noach: We're running out on `varchar` capacity in `_vt.schema_migrations`, which is why I now use `*text` types for everything. But, I'm curious, isn't `tinytext` an equivalent of `varchar(255)`? And it can be indexed just as with a `varchar`? At least I've always indexed `*text` types without a problem.

go/vt/vttablet/onlineddl/schema.go
```diff
@@ -98,7 +100,7 @@ const (
 		reverted_uuid,
 		is_view
 	) VALUES (
-		%a, %a, %a, %a, %a, %a, %a, %a, %a, FROM_UNIXTIME(NOW()), %a, %a, %a, %a, %a, %a, %a, %a
+		%a, %a, %a, %a, %a, %a, %a, %a, %a, NOW(), %a, %a, %a, %a, %a, %a, %a, %a
```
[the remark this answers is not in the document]
shlomi-noach: So just to clarify, the column type in this case is a `TIMESTAMP` and there was a long standing bug where we erroneously `INSERT`ed a  `FROM_UNIXTIME()` value, which of course did not make any sense.

We are now looking at a 1.5 year old decision, to use `TIMESTAMP` values in `_vt.schema_migrations` table, long before it was even associated with `_vt.vreplication`, and long before I knew of th eexistence of `_vt.vreplication`. This is how the code "grew" and now it's too late to change `_vt.schema_migrations` from `TIMESTAMP` and into `INT`.

The fact that we copy values from `_vt.vreplication`, that happen to be `INT` types, and into `_vt.schema_migrations` with `TIMESTAMP` types, is unfortunate, but also does not tell the entire story. For long, `schema_migrations` acted with `gh-ost` and `pt-osc`, oblivious of `vreplication`.

go/vt/vttablet/onlineddl/vrepl.go
shlomi-noach: I didn't know that, but you're the native English speaker :sweat_smile:

go/test/endtoend/onlineddl/vrepl/onlineddl_vrepl_test.go
```diff
@@ -358,17 +359,78 @@ func TestSchemaChange(t *testing.T) {
...
+			require.NotNil(t, rs)
+			for _, row := range rs.Named().Rows {
+				startedTimestamp := row.AsString("started_timestamp", "")
+				require.NotEmpty(t, startedTimestamp)
+				lastThrottledTimestamp := row.AsString("last_throttled_timestamp", "")
+				assert.NotEmpty(t, lastThrottledTimestamp)
+				// rowstreamer throttle timestamp only updates once in 10 seconds, so greater or equals" is good enough here.
+				assert.GreaterOrEqual(t, lastThrottledTimestamp, startedTimestamp)
+				component := row.AsString("component_throttled", "")
+				assert.Contains(t, []string{"vstreamer", "rowstreamer"}, component)
```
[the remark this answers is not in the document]
shlomi-noach: Refactored and now using a `WaitForThrottledTimestamp()` function.

go/test/endtoend/onlineddl/vrepl/onlineddl_vrepl_test.go
```diff
@@ -358,17 +359,78 @@ func TestSchemaChange(t *testing.T) {
...
 
 		uuid := testOnlineDDLStatement(t, alterTableTrivialStatement, "vitess", providedUUID, providedMigrationContext, "vtgate", "test_val", "", true)
 		_ = onlineddl.WaitForMigrationStatus(t, &vtParams, shards, uuid, normalMigrationWait, schema.OnlineDDLStatusRunning)
 		onlineddl.CheckMigrationStatus(t, &vtParams, shards, uuid, schema.OnlineDDLStatusRunning)
 		testRows(t)
 
+		// gotta give the migration a few seconds to read throttling info from _vt.vreplication and write
+		// to _vt.schema_migrations
+		time.Sleep(5 * time.Second)
+		rs := onlineddl.ReadMigrations(t, &vtParams, uuid)
```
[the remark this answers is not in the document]
shlomi-noach: Refactored and now using a `WaitForThrottledTimestamp()` function.

go/vt/vttablet/tabletmanager/vreplication/vreplicator.go
shlomi-noach: Fixed. Now formalized as `VReplicationComponent`.

shlomi-noach: Rather, `ComponentName`.

go/vt/vttablet/tabletserver/vstreamer/vstreamer.go
mattlord: OK, let's assume there was some good reason. 😄

mattlord (approved):
Looks great! Thank you ❤️

Issue #10565, opened 2022-06-22
Close stale Pull Requests


frouioui:
### Feature Description

When a Pull Request has been stale for a long time, we should warn the author by automatically commenting on it. If after a delay the Pull Request is still stale, we should automatically close it.

There are bots that can achieve this in the GitHub Marketplace.

### Use Case(s)

Let's say someone creates a Pull Request, it gets reviewed and some changes are required, if after `x` weeks or months the author has not come back to the Pull Request and made the required changes, we would automatically comment the Pull Request. This will allow the author and participants to receive a notification about the Pull Request. If no one takes action on the Pull Request after a delay, let's say 7 days, we would automatically close the Pull Request.

If the author or someone else needs, they can always re-open the Pull Request.

ajm188:
I think we should look at the official (? maybe, it's under their org) [github/stale](https://github.com/actions/stale) action. Based on my read we want the following config:

```yaml
# 1 month of no activity, or, for us, "one third of a release cycle"
days-before-stale: 30
# Never mark issues as stale.
days-before-issue-stale: -1
# Close 7 days after transitioning to Stale. This is the default, so technically we don't need to specify.
days-before-close: 7

stale-pr-message: TODO, work on wording, but we should set this
close-pr-message: TODO, work on wording, but we should set this

# This is the default value. We'll need to create this label in the vitess repo.
stale-pr-label: Stale

# This defaults to false, but i think we should talk about whether we want to close drafts (I'm not sure!).
# Note also the docs specify a big warning that we should take into account: https://github.com/actions/stale#exempt-draft-pr
exempt-draft-pr: true
```

PR #10603, opened 2022-06-28
Add stale PRs action

ajm188:
## Description

This adds a new workflow using https://github.com/actions/stale to close stale PRs after a period of time.

It currently is scheduled on a cron to run at 1am UTC, but obviously we can change that. I've also left it in debug-only (so it won't actually close anything) and with stats on, so we can see what it's doing before turning it on For Real. The other thing is I _believe_ we'll need to manually create the `Stale` label, so I'll do that once we're happy and ready to merge this.

I'm not 100% sure how to test this without merging and then manually triggering the workflow, but if you have sweet gh-actions debugging tips I'd love to hear them!

The other _other_ thing is there's currently no way to tag the PR author, but there is an [open issue for that feature](https://github.com/actions/stale/issues/714)

## Related Issue(s)

#10565 

## Checklist

## Deployment Notes

Before-merge:
- [ ] create Stale label

After-merge:
- [ ] trigger manually to verify
- [ ] wait a few days and check debug logs after nightly runs
- [ ] follow-up PR to turn off stats and `debug-only`

### DIFF
--- a/.github/workflows/close_stale_pull_requests.yml
+++ b/.github/workflows/close_stale_pull_requests.yml
@@ -1,0 +1,26 @@
+name: close_stale_pull_requests
+on:
+  schedule:
+    - cron: '0 1 * * *'
+
+permissions:
+  pull-requests: write
+
+jobs:
+  close_stale_pull_requests:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


deepthi (approved):
👍🏼

PR #10614, opened 2022-06-29
Online DDL: increase stale migration timeout

shlomi-noach:
## Description

A very simple change, increasing stale migration timeout from `10min` to `180min`  (`3h`).

Stale migration timeout is when Online DDL sees no liveness indication on a migration for a period of time. In such case, it cancels the migration, forcibly terminating it.

Normally, a 10minute timeout is a reasonable value. However, we are seeing some scenarios where this may be insufficient, relating to how VReplication reports (or does not report) liveness. Until VReplication's behavior is reviewed, this simple increase should do well to avoid terminating some migrations prematurely. Downside of this change, is that migrations that really go rogue, will take 3h to terminate clean up.
Thankfully, rogue migrations are a rare thing, and so aforementioned downside is also rare.

## Related Issue(s)

#6926 

## Checklist

### DIFF
--- a/go/vt/vttablet/onlineddl/executor.go
+++ b/go/vt/vttablet/onlineddl/executor.go
@@ -113,7 +113,7 @@ var maxConstraintNameLength = 64
 
 const (
 	maxPasswordLength                        = 32 // MySQL's *replication* password may not exceed 32 characters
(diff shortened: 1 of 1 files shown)


rohit-nayak-ps (approved):
lgtm

PR #10619, opened 2022-06-29
[cmd/*] Switch to pflag for all CLI flag parsing

ajm188:
## Description

This PR begins the next step in the implementation of [VEP-4](https://github.com/vitessio/enhancements/blob/main/veps/vep-4.md), officially moving us to using `pflag` in all of our CLI flag parsing. It does **not** migrate any flag definitions (except one small proof-of-concept/example), or move any flags (at all) out of the global flagset, which can be done incrementally in follow-up PRs.

### How to move a package's flags out of the global flagset — The Guide

(if you think of a good place to house this, let me know!)

For each flag in the package you are tackling, you are going to want to determine if it truly should be global and available to every binary (probably flags that control logging and tracing behavior are good candidates), or if they only make sense for a particular binary or particular subset of binaries.

To see which `cmd` entrypoints actually import your package _at all_ (which doesn't actually tell you if the _value_ that flag is set to has any impact on that binary), you can run:

```
go list -f '{{ .ImportPath }}| {{ join .Deps " " }} ' ./go/cmd/... | awk -F"|" '$2 ~ " <full import path of your package here> " {print $1}'
```

The spacing in the `-f` argument to `go list` might seem arbitrary, but it's specific and important here, as it lets us do an exact match in the `awk` (e.g. you want to know if `vitess.io/vitess/go/vt/vtgate` is imported, but _not_ get false positives if someone imports `vitess.io/vitess/go/vt/vtgate/buffer`, or anything else underneath the `go/vt/vtgate/*` tree).

As an example, here are the `cmd`s that import `grpcvtgateservice`:

```
$ go list -f '{{ .ImportPath }}| {{ join .Deps " " }} ' ./go/cmd/... | awk -F"|" '$2 ~ " vitess.io/vitess/go/vt/vtgate/grpcvtgateservice " {print $1}'
vitess.io/vitess/go/cmd/vtgate
vitess.io/vitess/go/cmd/vtgateclienttest
```

Armed with this list, you need to make a judgement call as to whether each of these commands actually _needs_ the flags defined in your package. I don't have much to say in the way of guidance here, except to ask the domain experts of the package and the `cmd` entrypoint. This is also why we have endtoend tests, and release candidates :) 

Once you've decided, you can move your flags to a function (`RegisterFlags` is a good name, as is `InstallFlags`, which takes a `*pflag.FlagSet` as input), for example:

```diff
var myCoolFlag int

- flag.IntVar(&myCoolFlag, "my-cool-flag", 0, "this is my cool flag, but it's in the global flagset")
+ func RegisterFlags(fs *pflag.FlagSet) {
+     fs.IntVar(&myCoolFlag, "my-cool-flag", 0, "this is my cool flag, and now it's only installed on this one flagset")
+ }
```

Now that you've done that, you'll hook into `servenv`'s new hook points to ensure your flag gets installed in the right places. If you decided your flag should truly be global, you will do:

```go
func init() {
    servenv.OnParse(RegisterFlags)
}
```

If you've decided your flag belongs only on one or two commands, you'll do:

```go
func init() {
    servenv.OnParseFor("vttablet", RegisterFlags)
    servenv.OnParseFor("vtgate", RegisterFlags) // this is also why I suggest writing a function rather than inlining
}
```

## Related Issue(s)

- https://github.com/vitessio/enhancements/issues/7
- Closes https://github.com/vitessio/vitess/issues/10696

## Checklist

### DIFF
--- a/go/cmd/mysqlctl/mysqlctl.go
+++ b/go/cmd/mysqlctl/mysqlctl.go
@@ -25,6 +25,8 @@ import (
 	"os"
 	"time"
 
+	"github.com/spf13/pflag"
+
 	"vitess.io/vitess/go/cmd"
 	"vitess.io/vitess/go/exit"
 	"vitess.io/vitess/go/flagutil"
@@ -248,7 +250,7 @@ func main() {
 		exit.Return(1)
 	}
 	dbconfigs.RegisterFlags(dbconfigs.Dba)
-	_flag.Parse()
+	_flag.Parse(pflag.NewFlagSet("mysqlctl", pflag.ExitOnError))
 
 	tabletAddr = netutil.JoinHostPort("localhost", int32(*port))
 
--- a/go/cmd/query_analyzer/query_analyzer.go
+++ b/go/cmd/query_analyzer/query_analyzer.go
@@ -24,6 +24,8 @@ import (
 	"os"
 	"sort"
 
+	"github.com/spf13/pflag"
+
 	"vitess.io/vitess/go/exit"
 	"vitess.io/vitess/go/vt/log"
 	"vitess.io/vitess/go/vt/sqlparser"
@@ -59,7 +61,7 @@ func (a stats) Less(i, j int) bool { return a[i].Count > a[j].Count }
 
 func main() {
 	defer exit.Recover()
-	_flag.Parse()
+	_flag.Parse(pflag.NewFlagSet("query_analyzer", pflag.ExitOnError))
 	for _, filename := range _flag.Args() {
 		fmt.Printf("processing: %s\n", filename)
 		if err := processFile(filename); err != nil {
--- a/go/cmd/topo2topo/topo2topo.go
+++ b/go/cmd/topo2topo/topo2topo.go
@@ -22,6 +22,8 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/spf13/pflag"
+
 	"vitess.io/vitess/go/exit"
 	"vitess.io/vitess/go/vt/log"
 	"vitess.io/vitess/go/vt/logutil"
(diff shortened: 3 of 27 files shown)


deepthi:
> (if you think of a good place to house this, let me know!)

How about converting the TODOs in the tracking project to Issues?
Then the issue that this PR fixes (partially, I assume), can include a checklist + the guide, and as we make progress we can check things off.

ajm188:
rebased to pick up the help text change [here](https://github.com/vitessio/vitess/compare/c38f6d7b480f443e135f28ae599ceb6e8cf41d06..1385debaabf68c264dcda6b6f53fa59467af7c82#diff-91994b1c6929e7c6d390002e2f006a76560d31954731c9d0d72143e85e7f888f)

deepthi (approved):

shlomi-noach:
:wave: this PR seems to break `vtctl/vtctlclient`, who no longer accept any command-specific flags. For example:
```shell
$ vtctlclient ApplySchema --skip_preflight --ddl_strategy "online -allow-zero-in-date --allow-concurrent --postpone-completion" --sql "alter table corder engine=innodb" commerce
unknown flag: --skip_preflight
```
or
```shell
$ vtctlclient ApplySchema --ddl_strategy "online -allow-zero-in-date --allow-concurrent --postpone-completion" --sql "alter table corder engine=innodb" commerce
unknown flag: --ddl_strategy
```
or
```shell
$ vtctlclient ApplySchema --sql "alter table corder engine=innodb" commerce
unknown flag: --sql
```

I'm not sure how tests passed, as they ought to have captured this malfunction. At any case, right now `main` seems to be broken in this regard; we're going to revert this PR and re-examine it later on.

ajm188:
(just closing the loop here) we discussed this in other fora, and the issue was a combination of being on different checkout and therefore needing the legacy `--` as in v14 deprecation warnings.

Issue #10590, opened 2022-06-27
Bug Report: Investigate MySQL 8.0 CI flakiness


mattlord:
### Overview of the Issue

I've noticed that with MySQL 8.0 we get more deadlock errors in vttablet during VReplication operations. For example, in the TestVDiff2 test when run locally using MySQL 8.0 `go test -timeout 5m -v -failfast -run TestVDiff2 vitess.io/vitess/go/test/endtoend/vreplication`:
```
E0626 20:23:49.325548   25364 controller.go:136] run() failed: WaitForPosition for stream id 1: Deadlock found when trying to get lock; try restarting transaction (errno 1213) (sqlstate 40001) during query: select pos, state, message from _vt.vreplication where id=1

E0626 20:23:56.585635   25839 table_differ.go:267] WaitForPosition error: 1: error waiting for pos: MySQL56/388e47d6-f5af-11ec-9ebe-1ae11deb438d:1-705, last pos: MySQL56/388e47d6-f5af-11ec-9ebe-1ae11deb438d:1-702: context canceled, wait time: 461.43875ms: possibly no tablets are available to stream in the source keyspace for your cell and tablet_types setting
```

Upon initial investigation I could see that InnoDB does _NOT_ in fact report a deadlock so I suspect that the tablet is treating an `innodb_lock_wait_timeout` as a deadlock or something else. At least the new VDiff2 test exposes it often enough that we can investigate further locally. 

One thing I already tried was adding an auto-retry here by only returning if `! isUnrecoverableError(err)` but then the test just timed out: https://github.com/vitessio/vitess/blob/13c7bccce7f792806f86de102b00df2fcdc91174/go/vt/vttablet/tabletmanager/vreplication/engine.go#L712-L756 

So something odd is happening with the state on the MySQL side.

### Reproduction Steps

If you have MySQL 8.0 installed locally and `VT_MYSQL_ROOT` is pointing to the install dir, then you'll see VDiff encounter an error if you run this enough times (not too many):
```
$ go test -timeout 5m -v -failfast -run TestVDiff2 vitess.io/vitess/go/test/endtoend/vreplication
```

### Binary Version

```sh
v15.0.0-SNAPSHOT
```

### Operating System and Environment details

```sh
N/A
```

Issue #10620, opened 2022-06-29
[14.0]: Tracking issue - flaky tests


deepthi:
### Overview of the Issue

While waiting for a PR to finish running on the release-14.0 branch, @frouioui and I noticed the following tests being flaky
- [x] TestSchemaChange/online_CREATE_TABLE:  [#10674](https://github.com/vitessio/vitess/issues/10674)
- [x] TestSchemaChange/revert_revert_CREATE_TABLE:  [#10674](https://github.com/vitessio/vitess/issues/10674)
- [x] Cluster (shardedrecovery_stress_verticalsplit_heavy) - TestVerticalSplit: https://github.com/vitessio/vitess/pull/10278
- [ ] Cluster (onlineddl_ghost) mysql80:
- [x] Cluster (vreplication_across_db_versions): https://github.com/vitessio/vitess/pull/10621

### Reproduction Steps

Links to failed runs:
https://github.com/vitessio/vitess/runs/7097331285?check_suite_focus=true
https://github.com/vitessio/vitess/runs/7097333414?check_suite_focus=true
https://github.com/vitessio/vitess/runs/7097328820?check_suite_focus=true
https://github.com/vitessio/vitess/runs/7097330419?check_suite_focus=true

### Binary Version

```sh
release-14.0 commit `3319a6bf39f4547b5be01744cf53f733462ee89d`
```

### Operating System and Environment details

```sh
CI
```

deepthi:
Re: `TestVerticalSplit` we should just start on deleting the old resharding code and this will go away.

mattlord:
The flakiness of these two are related to MySQL 8 usage:
  Cluster (onlineddl_ghost) mysql80
  Cluster (vreplication_across_db_versions)

They are otherwise identical to other workflows which use 5.7 and are not (so) flakey. I have an open issue on this: https://github.com/vitessio/vitess/issues/10590

GuptaManan100:
I don't think we need this issue anymore.

PR #10621, opened 2022-06-30
VReplication: retry in WaitForPos when read of pos is killed off by deadlock detector

mattlord:
## Description

For whatever reason(s), we've seen the deadlock detector fire more frequently when using MySQL 8.0 in our CI tests. This has, in particular, hit our VReplication related tests where there is very high contention on a small number of [`_vt.vreplication`](https://github.com/vitessio/vitess/blob/9199470f84406a62b8e70c7766df53f3e6e46c96/go/vt/binlog/binlogplayer/binlog_player.go#L540-L556) records. The record(s) in [the table](https://github.com/vitessio/vitess/blob/9199470f84406a62b8e70c7766df53f3e6e46c96/go/vt/binlog/binlogplayer/binlog_player.go#L540-L556) are getting read by different connections (by `id`, which is the Primary Key) while also getting constantly updated in other connections to update the progress: `pos`, `rows_copied`, `heartbeat`, etc.

This has led to the VReplication related tests that use MySQL 8.0 to be flaky, in particular:
- onlineddl_vrepl_stress_mysql80
- vreplication_across_db_versions

In this PR we retry the read of the `_vt.vreplication.pos` field if we get a [deadlock detected error](https://dev.mysql.com/doc/mysql-errors/en/server-error-reference.html#error_er_lock_deadlock) when waiting for a tablet to reach a replication position, until we hit the context timeout.

## Related Issue(s)

Fixes: https://github.com/vitessio/vitess/issues/10590
Related-to: https://github.com/vitessio/vitess/issues/10620

## Checklist


shlomi-noach (changes requested):
Looks good! Thank you for investigating this!

One thing, since calls to `dbClient.ExecuteFetch` can go into loop, is that we must now check `ctx.Done()` or `ctx.Err()` before calling `qr, err := dbClient.ExecuteFetch(binlogplayer.ReadVReplicationStatus(uint32(id)), 10)`. Which requires some delicate refactoring because we then cannot populate the error message with data from `qr`.

mattlord:
> Looks good! Thank you for investigating this!
> 
> One thing, since calls to `dbClient.ExecuteFetch` can go into loop, is that we must now check `ctx.Done()` or `ctx.Err()` before calling `qr, err := dbClient.ExecuteFetch(binlogplayer.ReadVReplicationStatus(uint32(id)), 10)`. Which requires some delicate refactoring because we then cannot populate the error message with data from `qr`.

Good point! If we continually hit the deadlock error then it's possible that we won't get to the `select` where we read from the context's done channel... I'll think about this.

shlomi-noach (approved):

mattlord:
Well, the DCO check flakiness has forced me to run the CI tests quite a few times. Good news is that it has demonstrated that the tests using MySQL 8.0 are now very stable! 🙂They would often need to be run 2 or 3 times to pass, now they're passing on the first try. 🥳

PR #10629, opened 2022-07-04
Parse LOCATE(), POSITION() and CHAR() functions

dbussink:
This adds parsing support for the `LOCATE()` and `POSITION()` functions. `POSITION()` specifically here was the goal as it's special syntax that doesn't look like a regular function.

It is an alias though for `LOCATE()` so this normalizes to that format if we see `POSITION()` in the SQL input.

## Related Issue(s)

#10203, #8604

## Checklist


dbussink:
I have extended this to also support the `CHAR()` function which is another string function that doesn't look like a regular function due to the `USING` syntax.

PR #10633, opened 2022-07-04
Improve handling of MATCH AGAINST

dbussink:
This improves the parsing of MATCH AGAINST. The column list is not a full expression in MySQL either, as queries like the following generate a parser error:

```
mysql> select * from t where match(1, a, 1) against ((select a from b));
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '1, a, 1) against ((select a from b))' at line 1
m
```

So we can be more strict here and only really accept column names. Additionally, the parenthesis are optional and it also works if those are not provided. This fixes that case as well and updates the select_cases.txt accordingly.

## Related Issue(s)

#10203, #8604

## Checklist


go/vt/sqlparser/ast_test.go
```diff
@@ -519,15 +519,6 @@ func TestReplaceExpr(t *testing.T) {
...
 		out: "convert(:a using utf8)",
-	}, {
-		in:  "select * from t where match((select a from b), 1) against (a)",
-		out: "match(:a, 1) against (a)",
-	}, {
-		in:  "select * from t where match(1, (select a from b), 1) against (a)",
-		out: "match(1, :a, 1) against (a)",
-	}, {
-		in:  "select * from t where match(1, a, 1) against ((select a from b))",
-		out: "match(1, a, 1) against (:a)",
```
dbussink, line 530: MySQL parses non of these queries so it doesn't really make sense we test these here.

GuptaManan100 (approved):
Looks good to me!

PR #10636, opened 2022-07-06
Refactor VTAdmin local example script to pass the cluster information as a file.

GuptaManan100:
## Description

The VTAdmin start-up script in the local example currently passes the cluster configuration as a comma-separated list of values, but VTAdmin has a flag to allow to pass the same information in a yaml file. Since the yaml file is the recommended way of providing this information, we should be using that in our local example too.
This PR addresses this issue.

## Checklist

### DIFF
--- a/examples/local/scripts/vtadmin-up.sh
+++ b/examples/local/scripts/vtadmin-up.sh
@@ -19,7 +19,7 @@ vtadmin \
   --alsologtostderr \
   --rbac \
   --rbac-config="./vtadmin/rbac.yaml" \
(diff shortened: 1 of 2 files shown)


ajm188:
So I agree that in general, using the `--cluster-config` flag is cleaner than passing a bunch of CSV `--cluster` flags on the command line. However

> Since the yaml file is the recommended way of providing this information

we actually don’t make any recommendation of which way you provide the information, and both are acceptable (and therefore, supported equally).

Second, I think this might actually make things more confusing, because the config file has to understand that the script reading it is being run from `./examples/local/` (and therefore the discovery config path is relatively `./vtadmin/discovery.json` instead of relative to the cluster config file itself, which would be just `./discovery.json`).

I don't think we want this for the local example, but continuing to use the `--cluster-config` method for VTOp is perfectly fine! In fact, probably good that we have examples that demonstrate both methods.

PR #10638, opened 2022-07-06
Online DDL: deal with autogenerated CHECK constraint names

shlomi-noach:
## Description

Workaround to https://bugs.mysql.com/bug.php?id=107772. Whe na table is created with an unnamed `CHECK` constraint name, MySQL generates one based on `<tablename>_chk_<number>`. Then, when the table is renamed, MySQL attempts to replace the table name. Vitess uses long table names while cutting-over tables at the end of a migration, e.g. `_vt_HOLD_6ace8bcef73211ea87e9f875a4d24e90_20200915120410`.

Next, as we already append a checksum to a `CHECK` constraint (see https://github.com/vitessio/vitess/pull/10369#issuecomment-1140900778), that name becomes too long, and the `RENAME TABLE` operation can fail.

In this PR we identify and dissipate  the `<tablename>_chk_<number>` naming format, so that it never escalates.

## Related Issue(s)

#10369 
#6926 

## Checklist


shlomi-noach:
Additional changes in this PR: identify error `1059` (name too long) and terminate migration if it is intercepted; update migrations' error message.

deepthi (commented):
Can we add an alter to the test case?
That is what exposed this issue. The initial create works fine, but an alter to a table with an autogenerated constraint name fails.

shlomi-noach:
@deepthi there is always an `alter` -- if none is defined, then it is assumed to be the trivial `alter table onlineddl_Test engine=innodb`.

dbussink (approved):

PR #10641, opened 2022-07-06
Accept geomcollection as alias for geometrycollection

dbussink:
https://dev.mysql.com/doc/refman/8.0/en/spatial-type-overview.html does not mention `geomcollection` as an alias, but
https://dev.mysql.com/doc/refman/8.0/en/gis-class-geometrycollection.html
does describe it and it even mentions it as preferred.

This preference shows in `SHOW CREATE TABLE` statements where `geomcollection` is always used as the type. We need to be able to parse that for `schemadiff` then.

In a follow up we can add normalization of this to `schemadiff` which seems like a better place to match the preference then that MySQL has than in the parser which accepts now both tokens.

## Related Issue(s)

#10203

## Checklist


GuptaManan100 (approved):
LGTM!

Issue #10626, opened 2022-06-30
Queries that don't get cached shouldn't be reported as cache misses


GuptaManan100:
## Description 

Currently, the queries which don't get cached like DML queries are still marked as query cache misses, but that doesn't seem correct. It masks the actual count of cache misses and makes the metric less useful.

GuptaManan100:
Fixed by #10643

PR #10643, opened 2022-07-07
Fix stats for cache miss and add CachePlan for Vtgate

rsajwani:

## Description
We have seen the cache miss metric numbers are typically on the higher side. I looks like we are trying to get plan from cache for SQL queries like "COMMIT", "ROLLBACK" etc. This result in unnecessary increment in cache miss counter. 
Also inorder to debug better we are adding PlanCache field in logstats object.

## Related Issue(s)

#10626 

## Checklist

### DIFF
--- a/go/vt/vtgate/executor.go
+++ b/go/vt/vtgate/executor.go
@@ -1010,7 +1010,7 @@ func (e *Executor) getPlan(vcursor *vcursorImpl, sql string, comments sqlparser.
 	_, _ = planHash.Write(hack.StringBytes(query))
 	planKey := hex.EncodeToString(planHash.Sum(nil))
 
-	if qo.cachePlan() && sqlparser.CachePlan(statement) {
+	if sqlparser.CachePlan(statement) && qo.cachePlan() {
 		if plan, ok := e.plans.Get(planKey); ok {
 			if logStats != nil {
 				logStats.CachedPlan = true
--- a/go/vt/vtgate/executor.go
+++ b/go/vt/vtgate/executor.go
@@ -999,10 +999,8 @@ func (e *Executor) getPlan(vcursor *vcursorImpl, sql string, comments sqlparser.
 		query = sqlparser.String(statement)
 	}
 
-	if logStats != nil {
-		logStats.SQL = comments.Leading + query + comments.Trailing
-		logStats.BindVariables = sqltypes.CopyBindVariables(bindVars)
-	}
+	logStats.SQL = comments.Leading + query + comments.Trailing
+	logStats.BindVariables = sqltypes.CopyBindVariables(bindVars)
 
 	planHash := sha256.New()
 	_, _ = planHash.Write([]byte(vcursor.planPrefixKey()))
@@ -1012,9 +1010,7 @@ func (e *Executor) getPlan(vcursor *vcursorImpl, sql string, comments sqlparser.
 
 	if sqlparser.CachePlan(statement) && qo.cachePlan() {
 		if plan, ok := e.plans.Get(planKey); ok {
-			if logStats != nil {
-				logStats.CachedPlan = true
-			}
+			logStats.CachedPlan = true
 			return plan.(*engine.Plan), nil
 		}
 	}
--- a/go/vt/vtgate/executor.go
+++ b/go/vt/vtgate/executor.go
@@ -1010,8 +1010,13 @@ func (e *Executor) getPlan(vcursor *vcursorImpl, sql string, comments sqlparser.
 	_, _ = planHash.Write(hack.StringBytes(query))
 	planKey := hex.EncodeToString(planHash.Sum(nil))
 
-	if plan, ok := e.plans.Get(planKey); ok {
-		return plan.(*engine.Plan), nil
+	if qo.cachePlan() && sqlparser.CachePlan(statement) {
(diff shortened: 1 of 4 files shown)


go/vt/vttablet/tabletserver/query_engine.go
```diff
@@ -300,6 +295,13 @@ func (qe *QueryEngine) GetPlan(ctx context.Context, logStats *tabletenv.LogStats
 	qe.mu.RLock()
 	defer qe.mu.RUnlock()
 	statement, err := sqlparser.Parse(sql)
+	if !skipQueryPlanCache && !sqlparser.SkipQueryPlanCacheDirective(statement) {
```
deepthi, line 298: Moving this here means that even when the plan is in the cache and can be re-used, we still parse the sql (because we want to check the Skip directive).
Is this worth it? 
Is it something that should be benchmarked?

go/vt/vtgate/logstats_test.go
```diff
@@ -193,7 +193,8 @@ func TestLogStatsContextHTML(t *testing.T) {
 	}
 	ctx := callinfo.NewContext(context.Background(), callInfo)
 	logStats := NewLogStats(ctx, "test", "sql1", "", map[string]*querypb.BindVariable{})
-	if string(logStats.ContextHTML()) != html {
+	var tmp = string(logStats.ContextHTML())
+	if tmp != html {
```
deepthi, line 197: This change seems unnecessary. We are not even using `tmp` to print out the failure message.

deepthi (commented):
Can you fix the PR title? We are not actually adding any additional stats for cache misses at all. We seem to instead be correcting a bug where the `CachedPlan` property on logStats was not being set even when the plan is being used from the cache.

go/vt/vtgate/executor.go
```diff
@@ -1010,8 +1010,13 @@ func (e *Executor) getPlan(vcursor *vcursorImpl, sql string, comments sqlparser.
 	_, _ = planHash.Write(hack.StringBytes(query))
 	planKey := hex.EncodeToString(planHash.Sum(nil))
 
-	if plan, ok := e.plans.Get(planKey); ok {
-		return plan.(*engine.Plan), nil
+	if qo.cachePlan() && sqlparser.CachePlan(statement) {
```
harshit-gangal, line 1013: nit: reverse the order of the check as `qo.cachePlan()` will most likely will always be true, so it will only depend on `sqlparser.CachePlan(statement)`

go/vt/vtgate/executor.go
```diff
@@ -1010,8 +1010,13 @@ func (e *Executor) getPlan(vcursor *vcursorImpl, sql string, comments sqlparser.
...
+		if plan, ok := e.plans.Get(planKey); ok {
+			if logStats != nil {
```
harshit-gangal, line 1015: I think we can skip doing nil check for `logStats`. It is expected to be not nil.

go/vt/vttablet/tabletserver/query_engine.go
```diff
@@ -300,6 +295,13 @@ func (qe *QueryEngine) GetPlan(ctx context.Context, logStats *tabletenv.LogStats
...
+		if plan := qe.getQuery(sql); plan != nil {
+			logStats.CachedPlan = true
+			return plan, nil
+		}
+	}
+
 	if err != nil {
 		return nil, err
 	}
```
harshit-gangal, line 307: I think we should not move getting from cached plan to here in vttablet
Reasons:
* vttablet receive a normalized query from vtgate so it can avoid query parsing here and look directly for cached plan and return - this avoids parsing overhead
* begin/commit.. other transactional things do not go through this code path so the chances of cache miss getting added are the least
* vttablet can use the skipQueryPlanCache directly at the top before parsing to check where it should look it in cached plan or not

go/vt/vttablet/tabletserver/query_engine.go
frouioui: I think it is a good idea to add a micro-benchmark on top of `GetPlan`, if the results are not conclusive we can always add the `Benchmark Me` label to this PR.

frouioui (commented):
agreed with the comments from @harshit-gangal  & @deepthi

go/vt/vtgate/executor.go
rsajwani: I did it because on line 1002 there is a nil check as well.. If that is the case then I will remove that one as well? thoughts ?

go/vt/vtgate/logstats_test.go
rsajwani: yeah I did it for debugging purpose .. forgot to revert it ..

harshit-gangal:
> Can you fix the PR title? We are not actually adding any additional stats for cache misses at all. We seem to instead be correcting a bug where the `CachedPlan` property on logStats was not being set even when the plan is being used from the cache, and another where we try to get a plan from the cache even when we should **know** that it is not in the cache.

logstats struct is not shared with vtgate and vttablet.
vttablet already has cacheplan, whereas vtgate did not have. @rsajwani added it in this PR.

harshit-gangal (approved):
everything looks good, just one comment about removing logstats nil check.

harshit-gangal (approved):

PR #10646, opened 2022-07-07
[vtctl] Delete query commands

### LINKED ISSUE #9933: Delete vtctl query commands
- [x] finish #9932 
- [ ] delete code after next release (v14)

ajm188:
## Description

This PR deletes the `vtctl` query commands, which were deprecated in https://github.com/vitessio/vitess/pull/9934.

I updated the `cluster.LocalProcess` used by endtoend tests to provide methods that duplicate the logic; ideally, we would rewrite those tests to validate their expectations in some other way, but I think this is fine for now.

The only outstanding thing is the `vttestserver` tests, which do not use `cluster.LocalProcess`, and have exactly two invocations of `VtGateExecute`. For now, I've commented them out, so that I can validate CI runs cleanly, but I'm not sure what to do there, doubly-so because I'm not super familiar with that component. cc @deepthi 

## Related Issue(s)

- Closes #9933
- #9934

## Checklist

-   [x] Documentation was added or is not required: https://github.com/vitessio/website/pull/1095


go/cmd/vttestserver/vttestserver_test.go
```diff
@@ -213,7 +211,7 @@ func TestExternalTopoServerConsul(t *testing.T) {
 	assert.NoError(t, err)
 	defer cluster.TearDown()
 
-	assertVtGateExecute(t, cluster)
+	// assertVtGateExecute(t, cluster)
```
deepthi, line 214: I think it's fine to delete this. However, we should replace it with some other operation / command that requires the topo server to be functional.

ajm188: lemme know what you think! https://github.com/vitessio/vitess/pull/10646/commits/53c8920c45044ddd73fecd0b67625cf3600b948e

deepthi (approved):

PR #10651, opened 2022-07-08
Add order, limit, skip options to onlineddl show command

Phanatic:
Signed-off-by: Phani Raj <phani@planetscale.com>

## Description

This PR adds the following options to the `vtctlclient OnlineDDL show` command

1. `order` to specify the order in which we return schema_migrations, based on the `added_timestamp` field.
2. `skip` to skip a set of rows from the result set. 
3. `limit` to limit how many rows we return from the results.

## Checklist

-   [x] Documentation was added or is not required https://github.com/vitessio/website/pull/1111


Phanatic:
Thanks for the reviews @notfelineit and @ajm188 I had the PR description wrong and all new parameters are single word and should match the current convention for flags.

ajm188:
> all new parameters are single word and should match the current convention for flags.

we're talking about the usage text, not the flag names themselves. So instead of "limit the number of rows returned", it should be "Limit the number of rows returned." (note the punctation as well as the capitalization)

Phanatic:
@ajm188 thanks for the clarification, I've update the help text to match the existing pattern. Let me know if this looks good.

deepthi:
@ajm188 does this need a manual update of website docs? Or will we autogen them?

deepthi:
Also, @shlomi-noach typically adds any user-facing online ddl changes to the summary release notes doc. See https://github.com/vitessio/vitess/blob/main/doc/releasenotes/15_0_0_summary.md

deepthi (changes requested):
A few more things are needed
- A test case (unit test is sufficient)
- Website docs PR
- Summary release notes update

go/vt/vtctl/vtctl.go
```diff
@@ -3168,9 +3174,19 @@ func commandOnlineDDL(ctx context.Context, wr *wrangler.Wrangler, subFlags *flag
 				condition, bindErr = sqlparser.ParseAndBind("migration_context=%a", sqltypes.StringBindVariable(arg))
 			}
 		}
+		order := " order by added_timestamp ascending"
+		if *orderBy == "descending" {
+			order = " order by added_timestamp descending"
```
shlomi-noach, line 3179: This column is not indexed, which means this will be a full table scan. Suggestion: `order by id desc`. `id` is the `PRIMARY KEY` for this table, and is an `AUTO_INCREMENT`. Thus, both `id` and `added_timestamp` are monotonously increasing.

shlomi-noach (changes requested):
Let's order by `id`, rather than `added_timestamp` -- see inline comment. Otherwise looks good! Feel free to dismiss the review once the change is made.

Phanatic:
@deepthi and @shlomi-noach thanks for the reviews . 
I've updated this PR to add unit tests in : https://github.com/vitessio/vitess/pull/10651/commits/9d3ba2f0d75206c7ea2ad2e94d8396a5ba72d175
and documentation update is in https://github.com/vitessio/website/pull/1111

shlomi-noach (approved):

deepthi (approved):

PR #10652, opened 2022-07-08
Lookup multi shard autocommit

harshit-gangal:
## Checklist

### DIFF
--- a/go/vt/vtgate/vindexes/lookup_internal.go
+++ b/go/vt/vtgate/vindexes/lookup_internal.go
@@ -24,25 +24,29 @@ import (
 	"strconv"
 	"strings"
 
+	"vitess.io/vitess/go/vt/vterrors"
+
 	"vitess.io/vitess/go/sqltypes"
 
 	querypb "vitess.io/vitess/go/vt/proto/query"
 	vtgatepb "vitess.io/vitess/go/vt/proto/vtgate"
+	vtrpcpb "vitess.io/vitess/go/vt/proto/vtrpc"
 )
 
 // lookupInternal implements the functions for the Lookup vindexes.
 type lookupInternal struct {
-	Table         string   `json:"table"`
-	FromColumns   []string `json:"from_columns"`
-	To            string   `json:"to"`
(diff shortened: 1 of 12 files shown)


go/vt/vtgate/vindexes/lookup_internal.go
```diff
@@ -24,25 +24,29 @@ import (
...
-	Autocommit    bool     `json:"autocommit,omitempty"`
-	Upsert        bool     `json:"upsert,omitempty"`
-	IgnoreNulls   bool     `json:"ignore_nulls,omitempty"`
-	BatchLookup   bool     `json:"batch_lookup,omitempty"`
-	sel, ver, del string
+	Table                string   `json:"table"`
+	FromColumns          []string `json:"from_columns"`
+	To                   string   `json:"to"`
+	Autocommit           bool     `json:"autocommit,omitempty"`
+	MultiShardAutocommit bool     `json:"multi_shard_autocommit,omitempty"`
```
frouioui, line 42: Can't we use commonConfig here?

frouioui (approved):

frouioui:
> If this is a change that users need to know about, please apply the release notes (needs details) label so that merging is blocked unless the summary release notes document is included.

@harshit-gangal, do you think we should mention this in the release note summary for v15?

frouioui:
Also @harshit-gangal, should we update [this](https://vitess.io/docs/user-guides/configuration-advanced/shard-isolation-atomicity/#method-1--the-naive-way) section of the documentation?

harshit-gangal:
> Also @harshit-gangal, should we update [this](https://vitess.io/docs/user-guides/configuration-advanced/shard-isolation-atomicity/#method-1--the-naive-way) section of the documentation?

that section is unrelated. will have to update lookup vindex page.

go/vt/vtgate/vindexes/lookup_internal.go
harshit-gangal: that will make these property of a json object. which we should not do for backward compatibility.

Issue #9961, opened 2022-03-23
Delete vtctl throttler commands


ajm188:
- [x] Finish #9960 
- [ ] delete code after next release (v14)

PR #9962, opened 2022-03-23
[vtctl] Deprecate Throttler RPCs

ajm188:
## Description

The new VReplication flows do not rely on these, which means we can delete them after vtworker is gone.

### Demo

```
➜  local git:(andrew/vtctldserver-deprecate-throttler) ✗ vtctlclient --server ":15999" help | grep Throttler                  
Resharding Throttler:
  (DEPRECATED) ThrottlerMaxRates -server <vtworker or vttablet>
  (DEPRECATED) ThrottlerSetMaxRate -server <vtworker or vttablet> <rate>
  (DEPRECATED) GetThrottlerConfiguration -server <vtworker or vttablet> [<throttler name>]
  (DEPRECATED) UpdateThrottlerConfiguration -server <vtworker or vttablet> [-copy_zero_values] "<configuration protobuf text>" [<throttler name>]
  (DEPRECATED) ResetThrottlerConfiguration -server <vtworker or vttablet> [<throttler name>]
➜  local git:(andrew/vtctldserver-deprecate-throttler) ✗ vtctlclient --server ":15999" ThrottlerMaxRates -- -h
WARNING: ThrottlerMaxRates is deprecated and will be removed in a future release. Use the new Reshard/MoveTables workflows instead.
Usage: ThrottlerMaxRates -server <vtworker or vttablet>

Returns the current max rate of all active resharding throttlers on the server.

  --server string
        vtworker or vttablet to connect to
➜  local git:(andrew/vtctldserver-deprecate-throttler) ✗ 
```

## Related Issue(s)

Closes #9960

## Checklist
- [x] Should this PR be backported? **no**
- [x] Tests were added or are not required **n/a**
- [x] Documentation was added or is not required

PR #10661, opened 2022-07-09
[vtctl] delete all throttler commands and associated cmd imports

ajm188:
## Description

This deletes the `Throttler*` vtctl commands, which were only needed for the legacy resharding code. These were deprecated in #9962.

Additionally, now that `vtctl` and `vtctld` have no need to call into `throttlerclient`, we delete the `plugin_grpcthrottlerclient` imports in both of their `cmd` entrypoints.

## Related Issue(s)

- Closes https://github.com/vitessio/vitess/issues/9961

## Checklist

-   [x] Documentation was added or is not required: https://github.com/vitessio/website/pull/1095


deepthi (approved):
The legacy resharding code has been deleted as of #10278.
Can you update the PR description to note that instead of the deprecation?

Rest LGTM

ajm188:
Oh I see my description was unclear. When I said "these were deprecated in ..." I was referring to the throttler RPCs, not the legacy resharding code. Let me clear that up!

PR #10662, opened 2022-07-11
Fix VTOrc Discovery to also retry discovering tablets which aren't present in database_instance table

### LINKED ISSUE #10650: Bug Report: VTOrc does not discover tablets if the durability policy is not set initially
### Overview of the Issue

If a user brings up a new cluster and starts VTOrc, then we see int the VTOrc logs that the tablets are ignored since the durability policy is not explicitly set for the keyspace - 
```
2022-07-08 08:37:52  WARNING orchestrator.go:250]  DiscoverInstance(10.244.0.23:3306) instance is nil in 0.005s (Backend: 0.002s, Instance: 0.002s), error=durability policy  not found
2022-07-08 08:37:52     INFO tablet_discovery.go:199] Discovered: alias:{cell:"zone1" uid:2548885007} hostname:"10.244.0.23" port_map:{key:"grpc" value:15999} port_map:{key:"vt" value:15000} keyspace:"commerce" shard:"-" key_range:{} type:PRIMARY db_name_override:"vt_commerce" mysql_hostname:"10.244.0.23" mysql_port:3306 primary_term_start_time:{seconds:1657269463 nanoseconds:158207300} db_server_version:"5.7.31" default_conn_collation:255
2022-07-08 08:37:52    DEBUG resolve.go:158] Hostname unresolved yet: 10.244.0.22
2022-07-08 08:37:52    DEBUG resolve.go:175] Cache hostname resolve 10.244.0.22 as 10.244.0.22
2022-07-08 08:37:52  WARNING orchestrator.go:250]  DiscoverInstance(10.244.0.22:3306) instance is nil in 0.003s (Backend: 0.002s, Instance: 0.002s), error=durability policy  not found
2022-07-08 08:37:52     INFO tablet_discovery.go:199] Discovered: alias:{cell:"zone1" uid:2469782763} hostname:"10.244.0.22" port_map:{key:"grpc" value:15999} port_map:{key:"vt" value:15000} keyspace:"commerce" shard:"-" key_range:{} type:REPLICA db_name_override:"vt_commerce" mysql_hostname:"10.244.0.22" mysql_port:3306 db_server_version:"5.7.31" default_conn_collation:255
```

If the user now runs `SetKeyspaceDurabilityPolicy`, then VTOrc should pick up these changes and discover the tablets, but it does not do that right now even after a lot of time elapses.

Restarting VTOrc fixes the problem though.

### Reproduction Steps

1. Start a new cluster, without specifiying the durability policies
2. Start VTOrc.
3. Specify the durability policy using `SetKeyspaceDurabilityPolicy`
4. Wait and see that VTOrc doesn't discover any tablet.

### Binary Version

```sh
Problem introduced in release-14.0 wherein the changes for the durability policy in the topo server were added.
```

```sh
All
```

### Log Fragments

_No response_

GuptaManan100:
## Description

This PR adds and end to end reproducing the linked issue wherein if the durability policy is set after VTOrc startup, VTOrc cannot pick up the changes and does not discover the vttablets at all, even after a lot of time elapses.

The problem was found in the loop where we refresh the MySQL information for vttablets. There we were only looking for tablets that were outdated i.e. their last valid check was older than the configured time.
In case the durability policy doesn't exist when the tablets are first discovered, we do not read the MySQL information and exit out early, leading to the tablet not being present in the `database_instance` table at all.
Therefore, the fix entails also checking for these tablets as well in the refresh code. The way to find these tablets is to find the ones that are present in the `vitess_tablet` table but not in the `database_instance` table. This operation is equivalent to an anti-join between the two tables, which can be written in MySQL as a left outer join with a predicate for only selecting rows that didn't have a matching counterpart in the `database_instance` table.

## Related Issue(s)

- Fixes #10650

## Checklist

### DIFF
--- a/go/test/endtoend/vtorc/general/vtorc_test.go
+++ b/go/test/endtoend/vtorc/general/vtorc_test.go
@@ -426,9 +426,9 @@ func TestMultipleDurabilities(t *testing.T) {
 	assert.NotNil(t, primary, "should have elected a primary")
 }
 
-// TestDurabilityPolicySetLate tests that VTOrc works even if the durability policy of the keyspace is
+// TestDurabilityPolicySetLater tests that VTOrc works even if the durability policy of the keyspace is
 // set after VTOrc has been started.
-func TestDurabilityPolicySetLate(t *testing.T) {
+func TestDurabilityPolicySetLater(t *testing.T) {
 	// stop any vtorc instance running due to a previous test.
 	utils.StopVtorcs(t, clusterInfo)
 	newCluster := utils.SetupNewClusterSemiSync(t)
@@ -436,7 +436,7 @@ func TestDurabilityPolicySetLate(t *testing.T) {
 	shard0 := &keyspace.Shards[0]
 	// Before starting VTOrc we explicity want to set the durability policy of the keyspace to an empty string
 	func() {
-		ctx, unlock, lockErr := newCluster.Ts.LockKeyspace(context.Background(), keyspace.Name, "TestDurabilityPolicySetLate")
+		ctx, unlock, lockErr := newCluster.Ts.LockKeyspace(context.Background(), keyspace.Name, "TestDurabilityPolicySetLater")
 		require.NoError(t, lockErr)
 		defer unlock(&lockErr)
 		ki, err := newCluster.Ts.GetKeyspace(ctx, keyspace.Name)
--- a/go/test/endtoend/vtorc/general/vtorc_test.go
+++ b/go/test/endtoend/vtorc/general/vtorc_test.go
@@ -17,18 +17,17 @@ limitations under the License.
 package general
 
 import (
+	"context"
 	"fmt"
 	"testing"
 	"time"
-
-	"vitess.io/vitess/go/vt/log"
-
-	"vitess.io/vitess/go/test/endtoend/vtorc/utils"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
 	"vitess.io/vitess/go/test/endtoend/cluster"
+	"vitess.io/vitess/go/test/endtoend/vtorc/utils"
+	"vitess.io/vitess/go/vt/log"
 )
 
 // Cases to test:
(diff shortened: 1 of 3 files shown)


frouioui (approved):
Paired with @GuptaManan100 on this. Looks good to me.

go/test/endtoend/vtorc/general/vtorc_test.go
```diff
@@ -426,3 +425,51 @@ func TestMultipleDurabilities(t *testing.T) {
 	primary := utils.ShardPrimaryTablet(t, clusterInfo, keyspaceSemiSync, shardSemiSync)
 	assert.NotNil(t, primary, "should have elected a primary")
 }
+
+// TestDurabilityPolicySetLate tests that VTOrc works even if the durability policy of the keyspace is
+// set after VTOrc has been started.
+func TestDurabilityPolicySetLate(t *testing.T) {
```
deepthi, line 431: nit: I would have called it `Later` :)

deepthi (approved):

go/test/endtoend/vtorc/general/vtorc_test.go
GuptaManan100: Fixed

vkovacik:
@GuptaManan100 @deepthi This fix has unfortunately not been merged into 14.0.2 released 11 days ago. 

We have been periodically affected by this issue in our large cluster. Every day there about 5-10 vttablets rescheduled by Kubernetes, as a part of normal hw maintenance operations. Almost in every case, restarted vttablet does not start responding immediately to vtorc discovery query, which makes vtorc to infinitely forget the instance.

As a result vtorc audit recovery log reports that the restarted instance is broken (e.g. ReplicationStopped) but vtorc never gets to clearing out the status even though the instance has been completely healthy. 

**User impact:**
- vtorc stops managing these restarted instances potentially resulting in shard breaking in future
- We need to restart all the affected vtorc pods every day to clear out alerting on the vtorc's audit log.

Can we please prioritize next 14.x release to include this fix?

GuptaManan100:
@vkovacik We have decided to backport this fix and will be available in the next patch release of v14.

PR #10664, opened 2022-07-11
Use TMC RPCs in VTOrc

GuptaManan100:
## Description

This PR changes the last remaining recovery function of VTOrc `fixReplica` to also use TabletManagerClient RPCs instead of running operations on the MySQL instance directly. 
In `fixReplica`, there are multiple things that need to be fixed - 
- The correct primary source should be set
- Replication should be started
- Correct semi-sync settings should be set
- The tablet should be read-only

The first 3 things are accomplished by `SetReplicationSource` RPC and the last by `SetReadOnly` RPC. Functionally, there is no difference from the user standpoint, but internally, now VTOrc talks to MySQL through the vttablet. 

As part of the PR, I also clean up the `fixPrimary` function. The code to set semi-sync directly on the MySQL instance was redundant since the following RPC already accomplished it. 

## Related Issue(s)

- #6612 

## Checklist

### DIFF
--- a/go/vt/orchestrator/logic/tablet_discovery.go
+++ b/go/vt/orchestrator/logic/tablet_discovery.go
@@ -320,28 +320,33 @@ func tabletDemotePrimary(instanceKey inst.InstanceKey, forward bool) error {
 	return err
 }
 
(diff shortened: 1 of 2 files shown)


deepthi (approved):
LGTM

PR #10665, opened 2022-07-11
Include more tests in upgrade tests

GuptaManan100:
## Description

The tests in `newfeaturetest` can now be moved out to the packages `emergencyreparent` and `plannedreparent` so that upgrade-downgrade tests run on them too.
As part of moving the test a couple of assertions had to be changed to support running with both MySQL versions. Also, the flakiness from the test was removed.

## Checklist


frouioui (approved):
Looks good to me. Should we remove the `newfeaturetest` package entirely now that it is empty?

GuptaManan100:
@frouioui I am inclined to keep it around since very soon I'll enable semi-sync and add a couple of tests that will only work in the current release. So, I'll be adding it back in a couple of days, if I remove it now.

go/test/endtoend/reparent/plannedreparent/reparent_test.go
```diff
@@ -439,3 +441,94 @@ func TestReplicationStatus(t *testing.T) {
 	assert.NotEmpty(t, replicationStatus.Position)
 	assert.NotEmpty(t, replicationStatus.RelayLogPosition)
 }
+
+// TestFullStatus tests that the RPC FullStatus works as intended.
+func TestFullStatus(t *testing.T) {
```
deepthi, line 446: Nice test 💯

PR #10667, opened 2022-07-11
MoveTables: use source timezone to adjust datetime columns on update statements

rohit-nayak-ps:
## Description

Add logic to translate timezones, when specified during MoveTables, in update statements also. It was currently only present in the path for insert statements. 

Tests for this path were missing and have been added in this fix.

## Related Issue(s)
https://github.com/vitessio/vitess/pull/10102

## Checklist

### DIFF
--- a/.github/workflows/cluster_endtoend_vreplication_movetables_tz.yml
+++ b/.github/workflows/cluster_endtoend_vreplication_movetables_tz.yml
@@ -1,0 +1,125 @@
+# DO NOT MODIFY: THIS FILE IS GENERATED USING "make generate_ci_workflows"
+
+name: Cluster (vreplication_movetables_tz)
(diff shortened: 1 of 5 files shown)


mattlord (approved):
LGTM!

I would only like to see us avoid adding another workflow whenever possible as we already have ~ 100 and we quickly develop a problematic queue when there are multiple PRs/merges happening as we can only have 50 workflows active at a time for the project/repo. I pushed a change to rename the `vreplication_migrate` workflow (where vdiff2 was recently added) to `vreplication_migrate_vdiff2_convert_tz` and added the new test to that workflow. 

If you disagree you can revert my commit and move forward. If you're OK with this approach, then we'll need to mark the old test name `vreplication_migrate` as not required and make the new name required.

PR #10670, opened 2022-07-12
Remove built-in decompression flag

rsajwani:

## Description
This PR includes bunch of optimization and some changes around backward compatibility scenario. 

1- Making code logic of deducing decompressor between Built-in and XtraBackup same.  
2- We don't need decompressor flag given we are always going to write what is the built-in compression used in MANIFEST
3- Adding compression engine field in Xtrabackup MANIFEST
4- Addressing the scenario when MANIFEST file doest not contain any value for compression engine. We should default to what we were using before which is "gzip". This can happen if the backup is done using vitess version <= v14 and we try to restore using main / v15
5- Introducing 'external' as one of possible value of compression engine. External specify if we are using custom flags or custom tool for compression and decompression.
6- Rename built-in compressor to more generic name 'compression-engine-name'
7- Change mysql version to 8.0.28 for upgrade-downgrade test

With this change here are the possible value of `--compression-engine-name`

- pgzip
- pargzip
- lz4
- zstd
- external

where 'external' is set only when we are using compressor other than the ones mentioned above. If you want to use any of the built-in compressors, simply set one of the above values for `--compression-engine-name`. Value specified in `--compression-engine-name` is saved in MANIFEST, which is later read by restore routine to decide what engine to use for decompression. Default value for engine is 'pgzip'

## Details
Xtrabackup engine was using file extension as a mean to determine the value of compressor during decompression, whereas built-in engine was using MANIFEST file to find out the value of compressor. This PR will make both engine use same logic which is to read it from MANIFEST file to determine the engine used during compression. As a result we found out that we don't need flags like Built-in decompressor. Furthermore we realized that Builtin-engine is used as a hint for decompressor to find out the compression engine name, so we changed its name to more generic 'compression-engine-name'.  Here is the list of different scenarios (possible values) which client can provide to these flags.

# Use Cases

- 1 (Use pgzip as compressor and decompressor)
compression-engine: pgzip
External compression: “”
External decompression: “”
  
 
- 2 (Use some external command as compressor and decompressor)
compression-engine: “external”
External compression: externalTool -c -y
External decompression: externalTool -d -y
external-compressor-extension: .xyz
 
- 3 (Use custom lz4 as compression and default built-in lz4 as decompressor)
compression-engine: “lz4”
External compression: lz4 -flag 
External decompression:
 
- 4 (Use custom lz4 as compression and decompressor)
compression-engine: “external”
External compression: lz4 -flag 
External decompression: lz4 -flag
 
 
# Migration Cases: 
- 5 (Changing from gzip to lz4 (built-in to built-in))

Original state: 
compression-engine: “gzip”
External compression: “” 
External decompression: “”
 
Final state:
compression-engine: “lz4”
External compression: “” 
External decompression: “”
 
- 6 (Changing from pgzip to pgzip custom (built-in to custom built-in))

Original state: 
compression-engine: “pgzip”
External compression: “” 
External decompression: “”
 
Final state: 
compression-engine: “pgzip”
External compression: “pgzip -flag”
External decompression: “”
 
- 7 (Changing from gzip to custom (built-in to custom))

Original state: 
compression-engine: “pgzip”
External compression: “” 
External decompression: “”
 
Final state: 
compression-engine: “external”
External compression: “external -c -flag”
External decompression: “external -d -flag”
 
- 8 (Changing from custom to lz4 (custom to built-in))

Original state: 
compression-engine: “external”
External compression: “custom -c flag” 
External decompression: “custom -d flag”
 
Intermediate  state: 
compression-engine: “lz4”
External compression: “”
External decompression: “external -d -flag”
 
finale  state: 
compression-engine: “lz4”
External compression: “”
External decompression: “”
 
- 9 (Changing from custom-builtin to lz4 (custom built-in to built-in))

Original state: 
compression-engine: “pgzip”
External compression: “pgzip -c” 
External decompression: “”
 
final  state: 
compression-engine: “lz4”
External compression: “lz4 -c”
External decompression: “”
 
- 10 (Changing from builtin to custom-built-in decompressor)

Original state: 
compression-engine: “pgzip”
External compression: “” 
External decompression: “”

Wont work: 
compression-engine: “pgzip”
External compression: “”
External decompression: “pgzip -c”
 
final state: 
compression-engine: “external”
External compression: “pgzip -c”
External decompression: “pgzip -c flag”
 
- 11 (Changing from external to external)

Original state: 
compression-engine: “external”
External compression: “tool -c” 
External decompression: “tool -d”
 
final  state: 
compression-engine: “external”
External compression: “tool2 -c”
External decompression: “tool1 -d”

P.S: In this case #11 we will run into issue where some of the older backups won't get restore after final state unless tool1 and tool2 are compatible.

## Related Issue(s)

#7802

## Checklist

-   [x] Documentation was added or is not required https://github.com/vitessio/website/pull/1091


go/vt/wrangler/testlib/backup_test.go
```diff
@@ -59,23 +58,20 @@ func TestBackupRestore(t *testing.T) {
 	require.NoError(t, err)
 }
 
-// TODO: @rameez. I was expecting this test to fail but it turns out
-// we infer decompressor through compression engine in builtinEngine.
+// Given we infer decompressor through Manifest file.
 // It is only in xtrabackup where we infer decompressor through extension & BuiltinDecompressor param.
```
deepthi, line 62: This comment is no longer valid, correct?

deepthi (commented):
Code changes look fine to me.
Do we have an update/downgrade test that verifies that we can restore a backup taken with an older version binary even if we provide new flags to the binary we are restoring with?

Also need to update release notes summary file, and related website PR https://github.com/vitessio/website/pull/1091 (cc @rvrangel)

go/vt/wrangler/testlib/backup_test.go
rsajwani: I just uncover couple of more cases for example if someone tries to move from external to buit-in and vice versa. I am sending an update ..

rsajwani:
> The idea behind a separate `--builtin-decompressor` flag is that you might want more control over your backup (for example, you want to do it at a minimum CPU priority or maybe a specific CPU affinity setting by using `nice`/`taskset`), but you don't care as much about the restore, since it when the host is restoring, it is not in service yet.
> 
> In that case we would benefit of being able to set an external compressor but use a more convenient builtin decompressor. This change introduces a new `external` engine that is set on the MANIFEST for any externally compressed backups, which prevents one from being able to use just the `--external-compressor` and `--external-compressor-extension` flags without having to set `--external-decompressor` as well. If that is the case, do we even care to have a `--external-compressor-extension` at all? If we always need to set both external compressor/decompressor, than the only reason it would exist is to make `xtrabackup` backups display that on the filename, as it is not really used when trying to detect which kind of decompression engine to use later on.
> 
> I think it is a good idea to also add the compression engine to the `xtrabackup` MANIFEST and to make the code logic of deducing the decompressor the same, but removing `--builtin-decompressor` takes away part of the flexibility of this change. Just my 2 cents.

The extension point you mentioned above does make sense and probably we can get rid of it. The problem with previous approach was consider a scenario where you decide to move from lz4 to pgzip. In that case the transition will be two phase first (step) you only change value of compressor to pgzip and keep the decompressor to still lz4 so latest backup can be decompressed and then as soon as you get your first backup with new compressor pgzip then you have to now change your decompressor to pgzip (second step). The whole idea of reading from manifest is to make logic simple else we will run into a lot of edge cases.

rvrangel:
If you want to move engines, you can (and should) keep `--builtin-decompressor auto`. so in your example, you change the compressor from `lz4` to `pgzip` and set the builtin-decompressor to `auto` (if it is not already like that). that means the tablet should be able to decompress both the old and new backups without a problem (the engine used on the backup is set in the manifest after all).

Perhaps if you want to simplify the logic, another option would be to replace `--external-compressor-extension` with `--external-compressor-engine` so a compatible format that is supported by the builtin decompressors is written to the MANIFEST as a hint that can be used  for decompressing and we skip dealing we extensions altogether. What do you think?

rsajwani:
> Perhaps if you want to simplify the logic, another option would be to replace `--external-compressor-extension` with `--external-compressor-engine` so a compatible format that is supported by the builtin decompressors is written to the MANIFEST as a hint that can be used for decompressing and we skip dealing we extensions altogether. What do you think?

Why do you need --external-compressor-engine separately, you can use *builtin-compressor to write to manifest just like you are doing it in Builtin-Engine.

doc/releasenotes/15_0_0_summary.md
```diff
@@ -94,32 +94,45 @@ Please see the VDiff2 [documentation](https://vitess.io/docs/15.0/reference/vrep
...
-so that compressed files are created with the correct extension. There is no need to override `--builtin-compressor` and `--builtin-decompressor`
-when using an external compressor/decompressor. Please note that if you want the current behavior then you don't need to change anything
-in these flags. You can read more about backup & restore [here] (https://vitess.io/docs/15.0/user-guides/operating-vitess/backup-and-restore/).
+so that compressed files are created with the correct extension. If the external command is not using any of the built-in compression engine
+(i-e pgzip, pargzip, lz4 or zstd) then you need to set `--compression-engine-name` to value 'external'.
+
+Please note that if you want the current production behavior then you don't need to change any of these flags.
+You can read more about backup & restore [here] (https://vitess.io/docs/15.0/user-guides/operating-vitess/backup-and-restore/).
+
+Important Note: If you decided to switch from external compressor to built-in supported compressors (i-e pgzi,, pargzip, lz4 or zstd) at any point
```
maxenglander, line 127:
```suggestion
Important Note: If you decided to switch from external compressor to built-in supported compressors (i-e pgzip, pargzip, lz4 or zstd) at any point
```

doc/releasenotes/15_0_0_summary.md
```diff
@@ -94,32 +94,45 @@ Please see the VDiff2 [documentation](https://vitess.io/docs/15.0/reference/vrep
...
+in the future. You might need to do it in two steps.
+
+- step #1, set `--external-compressor` and `--external-compressor-extension` flag values to empty and change `--compression-engine-name` to desired value.
+- Step #2, after few cycles of backups now set `--external-decompressor` flag value to empty.
```
maxenglander, line 131: > few cycles of backups

Why a few? Why not just 1?

doc/releasenotes/15_0_0_summary.md
```diff
@@ -94,32 +94,45 @@ Please see the VDiff2 [documentation](https://vitess.io/docs/15.0/reference/vrep
...
+
+The reason we recommend not to change all the values together is because restore routine have no way to find out which external decompressor
```
maxenglander, line 133: Would it make sense to have the binary reject (or at least issue a warning for) combinations of flags that can lead to this scenario?

maxenglander (commented):
I was asked to take a look at this and offer perspective on how this will affect PlanetScale and PS customers. Just a couple questions for now. Thanks!

go/vt/mysqlctl/xtrabackupengine.go
```diff
@@ -554,16 +557,28 @@ func (be *XtrabackupEngine) extractFiles(ctx context.Context, logger logutil.Log
 		// Create the decompressor if needed.
 		if compressed {
 			var decompressor io.ReadCloser
-
+			var deCompressionEngine = bm.CompressionEngine
+			if deCompressionEngine == "" {
+				// For backward compatibility. Incase if Manifest is from N-1 binary
+				// then we assign the default value of compressionEngine.
+				deCompressionEngine = PgzipCompressor
```
maxenglander, line 564: Is it possible that a user would have compressed with a non-default engine, and that this wouldn't have been recorded in the manifest file?

If yes, then in that case what error would the user get?

Wondering if it would be more helpful to require that the user specify compression engine if it's not present in the manifest, even nicer if we can hint what we think the right value is (e.g. if we can peek at magic numbers).

go/vt/mysqlctl/xtrabackupengine.go
```diff
@@ -80,7 +79,11 @@ const (
 type xtraBackupManifest struct {
 	// BackupManifest is an anonymous embedding of the base manifest struct.
 	BackupManifest
-
+	// CompressionEngine stores which compression engine was originally provided
+	// to compress the files. Please note that if user has provided externalCompressorCmd
+	// then it will contain value 'external'. This field is used during restore routine to
```
maxenglander, line 84: Would it also be helpful to store the external compressor command in the manifest? Even if it's not something Vitess can use, it might be helpful as an historical record for the user in case they end up in a scenario where they lose track of the command they used to compress.

go/vt/mysqlctl/xtrabackupengine.go
```diff
@@ -554,16 +557,28 @@ func (be *XtrabackupEngine) extractFiles(ctx context.Context, logger logutil.Log
...
+				// For backward compatibility. Incase if Manifest is from N-1 binary
+				// then we assign the default value of compressionEngine.
+				deCompressionEngine = PgzipCompressor
+			}
 			if *ExternalDecompressorCmd != "" {
-				decompressor, err = newExternalDecompressor(ctx, *ExternalDecompressorCmd, reader, logger)
+				if deCompressionEngine == ExternalCompressor {
+					deCompressionEngine = *ExternalDecompressorCmd
+					decompressor, err = newExternalDecompressor(ctx, deCompressionEngine, reader, logger)
+				} else {
```
maxenglander, line 570: If I'm reading this right it looks like if the user specifies `--external-decompressor`, but the manifest doesn't specify that the compression engine is `external`, then we'll ignore the user's request and instead decompress according to the manifest. Is that right?

go/vt/mysqlctl/xtrabackupengine.go
```diff
@@ -554,16 +557,28 @@ func (be *XtrabackupEngine) extractFiles(ctx context.Context, logger logutil.Log
...
+					decompressor, err = newBuiltinDecompressor(deCompressionEngine, reader, logger)
+				}
 			} else {
-				decompressor, err = newBuiltinDecompressorFromExtension(extension, *BuiltinDecompressor, reader, logger)
+				if deCompressionEngine == ExternalCompressor {
```
maxenglander, line 574: If the manifest specifies `external` compression engine, but the user didn't specify `--external-decompressor`, then we return an error. Is that right?

doc/releasenotes/15_0_0_summary.md
rsajwani: yes, it should be at least one.. let me fix it

doc/releasenotes/15_0_0_summary.md
rsajwani: it will make code unnecessarily complex given external compressor can be used for builtin compressor engine as well (e.g scenario 6 above in description). I need to get into business of parsing the commands

go/vt/mysqlctl/xtrabackupengine.go
rsajwani: Is it possible that a user would have compressed with a non-default engine, and that this wouldn't have been recorded in the manifest file?
>> No, thats not possible .. current binary in production uses pgzip as default. Only scenario when we don't see a value in manifest is during our first deployment to this feature. Once this feature is deployed then we expect manifest to alway have some value.

If we end up reverting the feature then what you specify can happen but even in this case "a flag user can use to specify compression engine" won't help as we will be reverting to older bits. The solution I can think of in that case is to move values to default (pgzip), perform a manual backup and then revert the binary.

go/vt/mysqlctl/xtrabackupengine.go
rsajwani: Yes I thought of doing it. I am assuming external compressor is not something people will use a lot. So i decided to keep it as a separate work item if needed in future. Let me know if you think otherwise. 
https://github.com/planetscale/vitess-private/issues/934

maxenglander: Sounds like a plan!

go/vt/mysqlctl/xtrabackupengine.go
maxenglander: > Only scenario when we don't see a value in manifest is during our first deployment to this feature.

In that scenario, I guess they will see an error like "gzip can't decompress file". I think it would be helpful if the user got a recommendation to remember to set `--external-decompressor`, or if we told the user that they are required to set `--external-decompressor` when the manifest doesn't specify compression.

Just my opinion on UX, so feel free to move on if you disagree.

go/vt/mysqlctl/xtrabackupengine.go
rsajwani: yes.

go/vt/mysqlctl/xtrabackupengine.go
rsajwani: yes

go/vt/mysqlctl/xtrabackupengine.go
maxenglander: If the user is coming from N-1 binary version, where they don't have the compression engine specified in their manifest, then how will they be able to decompress a backup that was compressed with non-default compression engine?

deepthi: This situation is not possible because in N-1 we didn't allow them to use anything other than the default (`pgzip`)

deepthi:
Removed the `needs details` label after verifying that changes to summary file are included in the PR.

maxenglander:
> P.S: In this case https://github.com/vitessio/vitess/pull/11 we will run into issue where some of the older backups won't get restore after final state unless tool1 and tool2 are compatible.

It would be nice if there was a way to deal with this with a set of flags, e.g. `--external-decompressor-transition` that could be used as a fallback decompressor if `--external-decompressor` didn't work. In absence of that, I think it would be helpful to add documentation to set users' expectations about this.

maxenglander (approved):
Aside from some notes I left on usability and documentation above, I think this PR will add the flexibility we need to address the first three bullet points in [this issue](https://github.com/vitessio/vitess/issues/7802).

rvrangel (approved):

deepthi:
> > P.S: In this case #11 we will run into issue where some of the older backups won't get restore after final state unless tool1 and tool2 are compatible.
> 
> It would be nice if there was a way to deal with this with a set of flags, e.g. `--external-decompressor-transition` that could be used as a fallback decompressor if `--external-decompressor` didn't work. In absence of that, I think it would be helpful to add documentation to set users' expectations about this.

+1 on the docs.

go/vt/mysqlctl/compression_test.go
```diff
@@ -98,6 +100,17 @@ func TestBuiltinCompressors(t *testing.T) {
 	}
 }
 
+func TestUnSupportedBuiltinCompressors(t *testing.T) {
+	logger := logutil.NewMemoryLogger()
+
+	for _, engine := range []string{"external", "foobar"} {
+		t.Run(engine, func(t *testing.T) {
+			_, err := newBuiltinCompressor(engine, nil, logger)
+			require.ErrorContains(t, err, "unsupported engine value for CompressionEngineName. supported values are `external`, `pgzip`, `pargzip`, `zstd`, `lz4` value:")
```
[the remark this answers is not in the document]
deepthi: an example of how to use `require.Regexp` here https://github.com/vitessio/vitess/blob/main/go/mysql/client_test.go#L134

go/vt/mysqlctl/compression_test.go
```diff
@@ -194,3 +207,29 @@ func TestValidateExternalCmd(t *testing.T) {
 		})
 	}
 }
+
+func TestValidateCompressionEngineName(t *testing.T) {
```
deepthi, line 211: How is this test different from `TestUnsupprtedBuiltinCompressors`?

go/vt/wrangler/testlib/backup_test.go
```diff
@@ -119,10 +115,7 @@ func testBackupRestore(t *testing.T, cDetails *compressionDetails) error {
 	backupstorage.BackupStorageImplementation = "file"
 	if cDetails != nil {
 		if cDetails.BuiltinCompressor != "" {
-			*mysqlctl.BuiltinCompressor = cDetails.BuiltinCompressor
-		}
-		if cDetails.BuiltinDecompressor != "" {
-			*mysqlctl.BuiltinDecompressor = cDetails.BuiltinDecompressor
+			*mysqlctl.CompressionEngineName = cDetails.BuiltinCompressor
```
deepthi, line 118: I thought `BuiltinCompressor` in `compressionDetails` has been renamed along with the flag. How is this working?

doc/releasenotes/15_0_0_summary.md
```diff
@@ -94,32 +94,45 @@ Please see the VDiff2 [documentation](https://vitess.io/docs/15.0/reference/vrep
...
 - lz4
 - zstd
+- external
 
-If you want to use any of the builtin compressors, simply set one of the above values for `--builtin-compressor`. You don't need to set
-the `--builtin-decompressor` flag in this case as we infer it automatically from the MANIFEST file. The default value for
-`--builtin-decompressor` is  `auto`.
+where 'external' is set only when we are using custom command or external tool other than the ones mentioned above. If you want to use any of the built-in compressors,
+simply set one of the above values for `--compression-engine-name`. Value specified in `--compression-engine-name` is saved in MANIFEST,
+which is later read by restore routine to decide what engine to use for decompression. Default value for engine is 'pgzip'.
```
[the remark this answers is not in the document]
rsajwani: I want to keep list as literal values as expected for flag

go/vt/mysqlctl/compression_test.go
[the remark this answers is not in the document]
rsajwani: ignoring it for now

go/vt/mysqlctl/compression_test.go
rsajwani: mainly builtin will fail 'external' but validcompressionEngineName will not fail 'external'

go/vt/wrangler/testlib/backup_test.go
rsajwani: Its a structure for unit test

deepthi (approved):
Nice work @rsajwani!
Thank you @rvrangel for all your help

PR #10671, opened 2022-07-12
refactor: unexplode VStreamRows() and reuse VStreamRowsRequest, unexplode VStream() and reuse VStreamRequest

shlomi-noach:
## Description

Similar to #10578 , we change function signatures:

- `VStream()` now uses `VStreamRequest` instead of 4 arguments
- `VStreamRows()` now uses `VStreamRowsRequest` instead of 3 argument

After this change, it is cleaner to pass the request, and easier to add new flags/parameters -- we just need to add them to `proto` as fields in `VStreamRequest/VStreamRowsRequest` instead of editing a dozen files and edit function signatures again.

## Related Issue(s)

none. similar to #10578

## Checklist


shlomi-noach:
Request for review.

dbussink (approved):
Looking good and making future changes easier :tada:.

PR #10680, opened 2022-07-13
vtgate e2e unsharded TestFloatValueDefault: expect different value for mysql80 

rohit-nayak-ps:

## Checklist

### DIFF
--- a/go/test/endtoend/utils/utils.go
+++ b/go/test/endtoend/utils/utils.go
@@ -44,6 +44,20 @@ func AssertMatches(t testing.TB, conn *mysql.Conn, query, expected string) {
 	}
 }
 
+// AssertMatchesOneOf ensures the given query produces one of the expected results.
+// useful when a test can give different results depending on mysql flavor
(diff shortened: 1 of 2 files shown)


harshit-gangal (approved):

go/test/endtoend/vtgate/misc_test.go
```diff
@@ -477,9 +478,12 @@ func TestFunctionInDefault(t *testing.T) {
 	// set the sql mode ALLOW_INVALID_DATES
 	utils.Exec(t, conn, `SET sql_mode = 'ALLOW_INVALID_DATES'`)
 
-	_, err = conn.ExecuteFetch(`create table function_default (x varchar(25) DEFAULT (TRIM(" check ")))`, 1000, true)
+	// commenting this out since we are planning to switch to 8.0 for unit tests, currently vtgate reports
```
systay, line 481: we should do this with t.Skip() instead, so we don't forget

Issue #10511, opened 2022-06-15
Bug Report: Orchestrator doesn't work with latest vitess and operator version


resworld:
### Overview of the Issue

I am following steps for running operator with orchestrator in latest (15) version of vitess and orchestrator can't discover new vitess cluster I've run. Scripts I've used are from https://github.com/vitessio/vitess/tree/main/examples/operator

### Reproduction Steps

1. Clone repo latest version: 
```sh
git clone https://github.com/vitessio/vitess.git
cd vitess && git log -1
```
Current commit hash I've tested it is:
```sh
commit 9f08615cb180aa3806d6a45eadeecf45455ab52b (HEAD -> main, origin/main, origin/HEAD)
Author: Manan Gupta <35839558+GuptaManan100@users.noreply.github.com>
Date:   Tue Jun 14 18:58:09 2022 +0530
```

2. Run minikube:
```sh
minikube start --cpus=8 --memory=11000 --disk-size=50g --kubernetes-version=v1.19.5
```

3. Run operator: 
```sh
kubectl apply -f operator.yaml
```

4. Start vitess cluster with vtorc example:
```sh
kubectl apply -f vtorc_example.yaml
```

5. Start forwarding script:
```sh
./orcpf.sh
```

6. Open http://localhost:3000 and check cluster status
```sh
No clusters found
```

### Binary Version

```sh
Image:         vitess/lite:latest
    Image ID:      docker-pullable://vitess/lite@sha256:89717535f987ceb87142be08a7cd9c029c3244fefd963aedee1979244cad8ec7
```

### Operating System and Environment details

```sh
running a minikube in macOS
```

### Log Fragments

```sh
I am attaching logs from vtorc container:


[martini] Started GET /web/clusters for 172.17.0.1:36540
[martini] Completed 200 OK in 1.496781ms
2022-06-15 06:45:08    DEBUG orchestrator.go:332] kv.SubmitPrimariesToKvStores, clusterName: , force: false: numPairs: 0
2022-06-15 06:45:08    DEBUG orchestrator.go:355] kv.SubmitPrimariesToKvStores: submitKvPairs: 0
2022-06-15 06:45:08     INFO audit_dao.go:105] auditType:review-unseen-instances instance::0 cluster: message:Operations: 0
2022-06-15 06:45:08     INFO audit_dao.go:105] auditType:resolve-unknown-primaries instance::0 cluster: message:Num resolved hostnames: 0
[...]

PR #10687, opened 2022-07-13
Fix VTop Example

GuptaManan100:
## Description

The VTop example is fixed and augmented to run VTAdmin and VTOrc too. As part of this PR, I attempted to also add a workflow to the CI to run the operator example as described on the website with all the steps, including move tables and resharding.
Unfortunately, this wasn't possible since the github-runners don't have enough resources to be able to support such a heavy workflow. I have removed the workflow now, but I'll reintroduce it later when we have stronger CI runners.

## Related Issue(s)

- Fixes https://github.com/vitessio/vitess/issues/10511

## Checklist

-   [x] Documentation was added or is not required - Website - https://github.com/vitessio/website/pull/1101


examples/operator/201_customer_tablets.yaml
```diff
@@ -67,16 +103,17 @@ spec:
...
-                  memory: 256Mi
+                  memory: 512Mi
             dataVolumeClaimTemplate:
               accessModes: ["ReadWriteOnce"]
               resources:
                 requests:
                   storage: 10Gi
   - name: customer
+    durabilityPolicy: none
     turndownPolicy: Immediate
```
GuptaManan100, line 117: I am not running VTOrc for the second keyspace we run. Maybe we should 🤷‍♂️ . I had not added it initially because I didn't want to make the test too bulky.

examples/operator/101_initial_cluster.yaml
```diff
@@ -27,25 +29,58 @@ spec:
       resources:
         requests:
           cpu: 100m
-          memory: 128Mi
+          memory: 256Mi
         limits:
-          memory: 512Mi
+          memory: 256Mi
```
deepthi, line 34: There are changes to requests/limits of the various components in this file. I understand needing to increase the request values. What is the reason for lowering the limits?

examples/operator/101_initial_cluster.yaml
```diff
@@ -146,8 +182,8 @@ stringData:
     CREATE USER 'vt_app'@'localhost';
     GRANT SELECT, INSERT, UPDATE, DELETE, CREATE, DROP, RELOAD, PROCESS, FILE,
       REFERENCES, INDEX, ALTER, SHOW DATABASES, CREATE TEMPORARY TABLES,
-      LOCK TABLES, EXECUTE, REPLICATION CLIENT, CREATE VIEW, SHOW VIEW,
-      CREATE ROUTINE, ALTER ROUTINE, CREATE USER, EVENT, TRIGGER
+      LOCK TABLES, EXECUTE, REPLICATION SLAVE, REPLICATION CLIENT, CREATE VIEW,
```
deepthi, line 185: Why do we need to add the additional grant? This section was originally copied from init_db.sql which has not changed in the same way.
The two should be kept in sync.

tools/get_kubectl_kind.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+
+set -euo pipefail
+source build.env
+
+# This script downloads kind binary that are
+# used as part of the vtop test environment,
+# and places them in vitess/bin/.
+#
+# The upgrade test framework expects kind binary to be found in the PATH.
```
deepthi, line 10: There isn't an upgrade test, is there? Maybe this comment just needs to be corrected..

examples/operator/101_initial_cluster.yaml
GuptaManan100: I didn't look too much into that, just copied over the configs from the operator example on VTop. I can revert changes to these if you like

examples/operator/101_initial_cluster.yaml
GuptaManan100: Same as above, I just copied what we have on VTop

tools/get_kubectl_kind.sh
GuptaManan100: Oh yes, that should be vtop example test, I'll fix it

examples/operator/101_initial_cluster.yaml
deepthi: Should be fine to copy over, but let's make sure to make them uniform across all the yaml files. It looks like the later steps request more (which can lead to unnecessary pod restarts). Just copy over the values from 302 to all the files.

examples/operator/101_initial_cluster.yaml
deepthi: Can you check what is embedded in vtop code?

deepthi: Is that in the example / test, or the version that is embedded in vtop code?

examples/operator/101_initial_cluster.yaml
```diff
@@ -27,25 +29,58 @@ spec:
...
+      - http://localhost:14001
+    replicas: 1
+    readOnly: false
+    apiResources:
+      limits:
+        memory: 128Mi
       requests:
         cpu: 100m
-        memory: 64Mi
+        memory: 128Mi
```
frouioui, line 62: Same thing here, maybe it would be good to ask @ajm188 or @notfelineit what they think the default resources allocation and the limit should be.

frouioui: I see that the old UI used to have a base request lower than the upper-bond limit

examples/operator/101_initial_cluster.yaml
frouioui: @deepthi should we keep the `REPLICATION SLAVE` term in our codebase? Is there an alternative?

examples/operator/README.md
```diff
@@ -12,15 +12,18 @@ kubectl apply -f operator.yaml
 # Example:
 #  images:
 #    vtctld: vitess/lite:mysql80
+#    vtadmin: vitess/vtadmin:latest
 #    vtgate: vitess/lite:mysql80
 #    vttablet: vitess/lite:mysql80
 #    vtbackup: vitess/lite:mysql80
+#    vtorc: vitess/lite:mysql80
 #    mysqld:
 #      mysql80Compatible: vitess/lite:mysql80
```
frouioui, line 21: Should we document the images to all be `:latest`?

Makefile
```diff
@@ -456,3 +456,6 @@ generate_ci_workflows:
 
 release-notes:
 	go run ./go/tools/release-notes -from "$(FROM)" -to "$(TO)" -version "$(VERSION)" -summary "$(SUMMARY)"
+
+install_kubectl_kind:
+	./tools/get_kubectl_kind.sh
```
frouioui, line 461: I don't see this being used anywhere in the codebase

tools/get_kubectl_kind.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+
+set -euo pipefail
+source build.env
+
+# This script downloads kind binary that are
+# used as part of the vtop test environment,
+# and places them in vitess/bin/.
```
frouioui, line 8: Is that file supposed to be used in the CI? If so, I recommend we use a GitHub Action to get Kind directly to ease maintenance. For instance:
- https://github.com/engineerd/setup-kind
- https://github.com/marketplace/actions/kind-cluster

examples/operator/101_initial_cluster.yaml
GuptaManan100: You got it.

examples/operator/README.md
GuptaManan100: That comment is following a line saying that if you want to run mysql80 then change the files like the following, that is why all of them are mysql80 instead of latest.

Makefile
GuptaManan100: This was used for the CI workflow which I have removed now. I have elected to keep all the other changes since it helps the next time we add the CI which should be soon enough

tools/get_kubectl_kind.sh
GuptaManan100: This is a better alternate, but we'll add the workflow back when we are in buildkite. I'll look for alternates to the file then

examples/operator/101_initial_cluster.yaml
GuptaManan100: Verified that they are already in sync. There are no unnecessary restarts in the example.

GuptaManan100: They aren't in sync on `main` 🤣

examples/operator/101_initial_cluster.yaml
GuptaManan100: There is no version of the `init_db.sql` embedded in VTop code. But this extra permission to `REPLICATION SLAVE` to `vt_app` is provided in all the examples and tests in VTop. But on `main`, we do not provide this permission to `vt_app`.

I have been able to track where this change occurred. In the PR https://github.com/vitessio/vitess/pull/10080 we removed this privilege from `vt_app`  but we didn't make the corresponding change in VTop. So I guess, the permissions on main are correct and I'll remove it here and create a PR on VTop too.

GuptaManan100: I have fixed it now.

tools/get_kubectl_kind.sh
GuptaManan100: Fixed

examples/operator/101_initial_cluster.yaml
GuptaManan100: PR on VTop for fixing there too - https://github.com/planetscale/vitess-operator/pull/293

examples/operator/101_initial_cluster.yaml
GuptaManan100: I don't really have any expertise in this matter, I'll defer to them to decide.

examples/operator/README.md
frouioui: Oh i missed that! my bad :)

examples/operator/101_initial_cluster.yaml
ajm188: @notfelineit would be able to answer better than me (I have no idea!)

notfelineit: The web resources look ok - I'm not sure about the api resources!

examples/operator/201_customer_tablets.yaml
deepthi: should be fine.

examples/operator/101_initial_cluster.yaml
deepthi: If the example is running fine for Manan locally, this should be fine. Anyone running into issues on their environment can easily edit the provided yaml.

deepthi (approved):
Looks like all feedback has been addressed.
We are choosing to leave in the infrastructure for CI without adding a CI workflow yet.

PR #10689, opened 2022-07-13
Parameterize BIT types and fixes in HEX types

harshit-gangal:
## Description

This PR makes few changes to Hex and Bit data types
1. Add parsing support for binary number representation i.e 0b1001
2. Normalize the query with binary number.
3. hex number and bit number to return varbinary types back to client for MySQL compatibility

Currently binary number representation is rewritten to hex number representation to maintain backward compatibility as currently binary number representation in normailized form is not understood by vttablet. 
This PR adds that support but it will be only used after v15 release.

The side-effect of that hex numbers on arithmetic return UINT64 and binary number returns INT64. In this PR it returns UINT64 for both.

## Checklist


go/sqltypes/bind_variables_test.go
```diff
@@ -22,6 +22,8 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
+
```
GuptaManan100, line 26: This new lines shouldn't be there

go/sqltypes/type.go
```diff
@@ -159,6 +159,7 @@ const (
 	HexNum     = querypb.Type_HEXNUM
 	HexVal     = querypb.Type_HEXVAL
 	Tuple      = querypb.Type_TUPLE
+	BitNum     = querypb.Type_BITNUM
```
GuptaManan100, line 162: Don't we need BitVal too?

go/sqltypes/value.go
```diff
@@ -250,10 +255,12 @@ func (v Value) ToBytes() ([]byte, error) {
 	switch v.typ {
 	case Expression:
 		return nil, vterrors.New(vtrpcpb.Code_INVALID_ARGUMENT, "expression cannot be converted to bytes")
-	case HexVal:
+	case HexVal: // TODO: all the decode below have problem when decoding odd number of bytes. This needs to be fixed.
```
GuptaManan100, line 258: What is the issue are you talking about here?

go/sqltypes/value_test.go
```diff
@@ -22,6 +22,8 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
+
```
GuptaManan100, line 26: Same new line not required here

go/vt/sqlparser/normalizer_test.go
```diff
@@ -25,6 +25,8 @@ import (
 	"strconv"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
+
```
GuptaManan100, line 29: new line here too

GuptaManan100 (commented):
Rest looks good to me. I think we should get more review for this PR

go/sqltypes/type.go
harshit-gangal: we can just convert it to BitNum, do not see a point in keeping two representation.

go/sqltypes/value.go
GuptaManan100: So, do odd numbered values don't work right now?

harshit-gangal: If we execute a `ToBytes()` function on them in the vtgate code.
which most likely happen in vindexes code.
So, if a vindex is presented as hex or bit it will not work today for odd string presentation

eg. `0x0f` will fail whereas `0x0f0f` will pass etc.

We should fix this as separate PR.

go/test/endtoend/vtgate/queries/misc/misc_test.go
```diff
@@ -0,0 +1,54 @@
+package misc
```
systay, line 1: license header missing

systay (approved):
👍

deepthi:
@harshit-gangal does this PR fix #10776?

harshit-gangal:
> @harshit-gangal does this PR fix #10776?

The issue mentioned here would need a support in evalengine. 
This PR fixes the parser and bind vars.

PR #10694, opened 2022-07-14
Online DDL: ALTER VITESS_MIGRATION COMPLETE ALL

shlomi-noach:
## Description

Similarly to https://github.com/vitessio/vitess/pull/7099, this PR adds a `ALTER VITESS_MIGRATION COMPLETE ALL` command, that internally runs a `ALTER VITESS_MIGRATION '<uuid>' COMPLETE` for all pending migrations, ie migration that are in these states:

- `queued`
- `ready`
- `running`

This query is supported in `vtgate` and in `vtctlclient ApplySchema`. Calling this query is faster than invoking `vtctlclient` multiple times. Of course, use with care as this will affect multiple migrations.

The main use case is to be able to quickly cut-over (`COMPLETE`) multiple concurrent vitess migrations that are running concurrently (per #10410).

## Related Issue(s)

- #6926 
- https://github.com/vitessio/vitess/pull/7099
- #10410 

## Checklist


GuptaManan100:
@shlomi-noach Do we need to add summary docs for this new command?

GuptaManan100 (commented):

shlomi-noach:
> Do we need to add summary docs for this new command?

@GuptaManan100 absolutely right, just added now.

GuptaManan100 (approved):
Great! I believe we also need website changes, but otherwise this looks good to me!

PR #10698, opened 2022-07-14
test: set parameter on vtgate than on vttablet

harshit-gangal:
## Description

This PR fixes the e2e test setup for reserved connection where the parameter to be set on vtgate was set on vttablet. This fixes that.

## Checklist


deepthi (approved):
This is an interesting issue. When we did collations we started reading this flag in vttablet as well, so it's an easy mistake to make.
I certify that I went through the review checklist 😄

PR #10699, opened 2022-07-14
[14.0] set parameter on vtgate than on vttablet

harshit-gangal:
## Description

Backport of https://github.com/vitessio/vitess/pull/10698

This PR fixes the e2e test setup for reserved connection where the parameter to be set on vtgate was set on vttablet. This fixes that.

## Checklist


rohit-nayak-ps (approved):
lgtm

PR #10701, opened 2022-07-15
Auto Detect MySQL Version and Use in vtgate mysql_server_version Flag

mattlord:
## Description

We were using the default value for the [vtgate](https://vitess.io/docs/15.0/reference/programs/vtgate/) `--mysql_server_version` flag of 5.7.9-vitess even when the tests were using MySQL 8.0. This work auto detects the MySQL version that will be used for the tests and uses that in the [vtgate](https://vitess.io/docs/15.0/reference/programs/vtgate/) `--mysql_server_version` flag **_unless_** a specific `mysql_server_version` has been specified for the test. This way we retain the ability to use custom values for this flag when needed, while otherwise defaulting to a reasonable version.

This will allow us to move more tests to MySQL 8.0 in a seamless way over the coming months  as 5.7 is slated for EOL in October, 2023.

⚠️ This PR exposed a bug in v14+ that could cause incorrect collation handling and vtgate crashes: https://github.com/vitessio/vitess/issues/10712. This PR addresses that as well, which is why I added the `Backport to: release-14.0` label.

## Related Issue(s)
  - Related to: https://github.com/vitessio/vitess/pull/10691
  - Fixes: https://github.com/vitessio/vitess/issues/10712

## Checklist

### DIFF
--- a/go/mysql/collations/local.go
+++ b/go/mysql/collations/local.go
@@ -20,6 +20,7 @@ package collations
 
 import (
 	"flag"
+	"strings"
 	"sync"
 
 	"vitess.io/vitess/go/vt/servenv"
@@ -36,9 +37,12 @@ func Local() *Environment {
 			panic("collations.Local() called too early")
 		}
 		if *servenv.MySQLServerVersion == "" {
-			defaultEnv = fetchCacheEnvironment(collverMySQL80)
+			// The default server version used by vtgate is 5.7.9
+			// NOTE: this should be changed along with the effective default
+			//       for the vtgate mysql_server_version flag.
+			defaultEnv = fetchCacheEnvironment(collverMySQL57)
 		} else {
-			defaultEnv = NewEnvironment(*servenv.MySQLServerVersion)
+			defaultEnv = NewEnvironment(strings.Clone(*servenv.MySQLServerVersion))
 		}
 	})
 	return defaultEnv
--- a/go/mysql/collations/local.go
+++ b/go/mysql/collations/local.go
@@ -20,7 +20,6 @@ package collations
 
 import (
 	"flag"
-	"strings"
 	"sync"
 
 	"vitess.io/vitess/go/vt/servenv"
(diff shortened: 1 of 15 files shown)


GuptaManan100 (approved):
Looks good to me! We might have to remove the `mysql_server_version` being set in some tests that are currently doing it because we now do it in this central place.

mattlord:
> Looks good to me! We might have to remove the `mysql_server_version` being set in some tests that are currently doing it because we now do it in this central place.

I pushed another commit removing any other place we used a static string in the endtoend tests: https://github.com/vitessio/vitess/pull/10701/commits/c5cf717478d2575a1e3d7f183aad56e95b85009b

If it causes tests failures then I'll either revert that or investigate and fix the failures.

frouioui (approved):
Looks good to me!

go/mysql/collations/local.go
```diff
@@ -36,9 +37,12 @@ func Local() *Environment {
...
 		}
 		if *servenv.MySQLServerVersion == "" {
-			defaultEnv = fetchCacheEnvironment(collverMySQL80)
+			// The default server version used by vtgate is 5.7.9
+			// NOTE: this should be changed along with the effective default
+			//       for the vtgate mysql_server_version flag.
+			defaultEnv = fetchCacheEnvironment(collverMySQL57)
 		} else {
-			defaultEnv = NewEnvironment(*servenv.MySQLServerVersion)
+			defaultEnv = NewEnvironment(strings.Clone(*servenv.MySQLServerVersion))
```
mattlord, line 45: This may have only impacted me locally, using macOS on arm64.

vmg (approved):
I like this approach, a lot of the engine/planner tests have always been written for 8.0 and porting them to 5.7 right before 5.7 is EOL seems wasteful.

systay:
> ⚠️ This PR exposed a bug in v14+ that could cause incorrect collation handling and vtgate crashes: https://github.com/vitessio/vitess/issues/10712. This PR addresses that as well, which is why I added the Backport to: release-14.0 label.

I think the prudent way to solve this would be to separate the autodetect and the bugfix into two different PRs. One should be backported, and not the other

systay:
> > ⚠️ This PR exposed a bug in v14+ that could cause incorrect collation handling and vtgate crashes: #10712. This PR addresses that as well, which is why I added the Backport to: release-14.0 label.
> 
> I think the prudent way to solve this would be to separate the autodetect and the bugfix into two different PRs. One should be backported, and not the other

Ignore me. Didn't realize that this was a test-only change. Me gusta

systay (approved):

frouioui (approved):

PR #10702, opened 2022-07-15
allow desired hashing function in CreateLookupVindex command

L3o-pold:
Modify CreateLookupVindex command to create the consistent_lookup_vindex with desired hashing function rather than it's automatic selection (optional).

```shell
vtctlclient --server localhost:15999 CreateLookupVindex -- --tablet_types=RDONLY customer "$(cat lookup_vindex.json)"
cat lookup_vindex.json
```

```json
{
    "sharded": true,
    "vindexes": {
        "corder_lookup": {
            "type": "consistent_lookup_unique",
            "params": {
                "table": "customer.corder_lookup",
                "from": "sku",
                "to": "keyspace_id"
            },
            "owner": "corder"
        }
    },
    "tables": {
        "corder": {
            "column_vindexes": [
                {
                    "column": "sku",
                    "name": "corder_lookup"
                }
            ]
        },
        "corder_lookup": {
          "column_vindexes": [
                {
                    "column": "sku",
                    "name": "binary_md5"
                }
            ]
        }
    }
}
```

## Checklist

## Deployment Notes

Documentation note can follow once approved.

### DIFF
--- a/go/vt/wrangler/materializer.go
+++ b/go/vt/wrangler/materializer.go
@@ -497,8 +497,8 @@ func (wr *Wrangler) prepareCreateLookup(ctx context.Context, keyspace string, sp
 	}
 
 	// Validate input table
-	if len(specs.Tables) != 1 {
-		return nil, nil, nil, fmt.Errorf("exactly one table must be specified in the specs: %v", specs.Tables)
+	if len(specs.Tables) < 1 || len(specs.Tables) > 2 {
+		return nil, nil, nil, fmt.Errorf("one or two table must be specified in the specs: %v", specs.Tables)
 	}
 	// Loop executes once.
 	for k, ti := range specs.Tables {
@@ -507,6 +507,7 @@ func (wr *Wrangler) prepareCreateLookup(ctx context.Context, keyspace string, sp
 		}
 		sourceTableName = k
 		sourceTable = ti
+		break
 	}
 
 	// Validate input table and vindex consistency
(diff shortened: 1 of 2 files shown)


go/vt/wrangler/materializer.go
```diff
@@ -497,8 +497,8 @@ func (wr *Wrangler) prepareCreateLookup(ctx context.Context, keyspace string, sp
 	}
 
 	// Validate input table
-	if len(specs.Tables) != 1 {
-		return nil, nil, nil, fmt.Errorf("exactly one table must be specified in the specs: %v", specs.Tables)
+	if len(specs.Tables) < 1 || len(specs.Tables) > 2 {
+		return nil, nil, nil, fmt.Errorf("one or two table must be specified in the specs: %v", specs.Tables)
```
mattlord, line 501: This is not a terribly helpful error message, IMO. What about something ~:
```
return nil, nil, nil, fmt.Errorf("you must specify the source table and optionally the target lookup table with the specific vindex of choice: %v", specs.Tables)
```

go/vt/wrangler/materializer.go
```diff
@@ -655,18 +656,36 @@ func (wr *Wrangler) prepareCreateLookup(ctx context.Context, keyspace string, sp
...
+					break
 				}
-				break
 			}
+
 		}
+
 		if targetVindex == nil {
 			// Unreachable. We validated column names when generating the DDL.
 			return nil, nil, nil, fmt.Errorf("column %s not found in schema %v", sourceVindexColumns[0], tableSchema.TableDefinitions[0])
```
mattlord, line 691: This should be changed too, as they may have specified another table but the key did not match the target table name in the `vindexes` definition.

go/vt/wrangler/materializer_test.go
```diff
@@ -1162,6 +1162,126 @@ func TestCreateLookupVindexSameKeyspace(t *testing.T) {
 		t.Errorf("same keyspace: got:\n%v, want\n%v", got, want)
 	}
 }
+
+func TestCreateLookupVindexCustomHashing(t *testing.T) {
```
mattlord, line 1166: Not all vindex types use hashing. IMO, we should use the terminology `CustomVindexType` here, and similar in the PR description and title, and in the new docs.

go/vt/wrangler/materializer_test.go
```diff
@@ -1162,6 +1162,126 @@ func TestCreateLookupVindexSameKeyspace(t *testing.T) {
...
+	ms := &vtctldatapb.MaterializeSettings{
+		SourceKeyspace: "ks",
+		TargetKeyspace: "ks",
+	}
+	env := newTestMaterializerEnv(t, ms, []string{"0"}, []string{"0"})
+	defer env.close()
+
+	specs := &vschemapb.Keyspace{
```
mattlord, line 1174: Nit, but this would be easier to read/follow if the lookup vindex spec followed the schema and vschema definitions.

go/vt/wrangler/materializer_test.go
```diff
@@ -1480,11 +1600,34 @@ func TestCreateLookupVindexFailures(t *testing.T) {
 		},
 		err: `vindexType "lookup_noexist" not found`,
 	}, {
-		description: "only one table",
+		description: "only one or two table",
```
mattlord, line 1603: I'd also prefer to add clarity here with ~ `only source and optionally target tables` and similar in the error message below.

mattlord (commented):
Overall, it LGTM. I had some minor comments and requests that I think will make this a bit more intuitive / easy to understand both for developers and users. 

We will also need to update the docs here accordingly to help explain this new optional syntax: https://vitess.io/docs/15.0/user-guides/configuration-advanced/createlookupvindex/

If there's already an open docs PR, can you please add it to the Documentation checkbox in the description? If there's not already one, would you mind creating one (if you need help, ping me in the Vitess Slack)? I want to make sure that we have some clear documentation because as I'm sure you're already well aware, it's not the most intuitive or self-describing syntax. 🙂 

Thank you for working on this! ❤️

go/vt/wrangler/materializer.go
harshit-gangal: Here we should validate that the if two tables are specified then one of them is the lookup table it self and throw the error accordingly.

PR #10708, opened 2022-07-15
[mysqlctl] Rewrite CLI to cobra

ajm188:
## Description

This expands on the work in #10619, rewriting the `mysqlctl` command with all its subcommands to use `spf13/cobra`.

### Testing

Starting the local example initial cluster, which invokes `mysqlctl` 

## Related Issue(s)

TK

## Checklist

### DIFF
--- a/go/cmd/mysqlctl/mysqlctl.go
+++ b/go/cmd/mysqlctl/mysqlctl.go
@@ -21,76 +21,146 @@ import (
 	"context"
 	"flag"
 	"fmt"
-	"io"
 	"os"
 	"time"
 
-	"github.com/spf13/pflag"
+	"github.com/spf13/cobra"
 
-	"vitess.io/vitess/go/cmd"
+	"vitess.io/vitess/go/cmd/vtctldclient/cli"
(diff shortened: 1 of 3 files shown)


go/cmd/mysqlctl/mysqlctl.go
```diff
@@ -21,76 +21,146 @@ import (
...
 	"fmt"
-	"io"
 	"os"
 	"time"
 
-	"github.com/spf13/pflag"
+	"github.com/spf13/cobra"
 
-	"vitess.io/vitess/go/cmd"
+	"vitess.io/vitess/go/cmd/vtctldclient/cli"
```
ajm188, line 29: we should find a better home for this `cli.FinishedParsing` function, now that places besides `vtctldclient` will be using it.

ajm188:
Looking at the test failures here, I think we need to wait one more release before switching to `cobra`. (`pflag` works with our `--` hack, but cobra is trying to be too smart).

So this will go in v16, and v15 will contain _just_ the transition from `flag.Parse` to `pflag.Parse`.

Issue #10713, opened 2022-07-16
Bug Report: Incorrect vtgate join plan


wcurrie:
### Overview of the Issue

Joining between 3 tables a join condition appears to be lost, causing a query to return incorrect data. The following query returns 4 rows instead of two:
```
mysql> select user.id, team.id
    -> from user
    ->          join team_member on user.id = team_member.user
    ->          join team on team_member.team = team.id
    -> where user.email = 'alice@gmail.com';
+----+----+
| id | id |
+----+----+
|  1 | 11 |
|  1 | 11 |
|  1 | 22 |
|  1 | 22 |
+----+----+
```
Explain shows 
```
mysql> explain format=vitess select user.id, team.id
    -> from user
    ->          join team_member on user.id = team_member.user
    ->          join team on team_member.team = team.id
    -> where user.email = 'alice@gmail.com';
+--------------+-------------+----------+-------------+------------+---------------------------------------------------------------------------+
| operator     | variant     | keyspace | destination | tabletType | query                                                                     |
+--------------+-------------+----------+-------------+------------+---------------------------------------------------------------------------+
| Join         | Join        |          |             | UNKNOWN    |                                                                           |
| ├─ Route     | EqualUnique | customer |             | UNKNOWN    | select `user`.id from `user` where `user`.email = 'alice@gmail.com'       |
| └─ Route     | Scatter     | customer |             | UNKNOWN    | select team.id from team_member, team where team_member.`user` = :user_id |
+--------------+-------------+----------+-------------+------------+---------------------------------------------------------------------------+
```
In the second query I'd expect something like ` and team_member.team = team.id` propagated from `join team on team_member.team = team.id` in the input sql.

Same thing happens with the simpler query too: 4 rows instead of two and same plan
```
mysql> explain format=vitess select user.id, team.id
    -> from user
    ->          join team_member on user.id = team_member.user
    ->          join team on team_member.team = team.id
    -> where user.id = 1;
+--------------+-------------+----------+-------------+------------+---------------------------------------------------------------------------+
| operator     | variant     | keyspace | destination | tabletType | query                                                                     |
+--------------+-------------+----------+-------------+------------+---------------------------------------------------------------------------+
| Join         | Join        |          |             | UNKNOWN    |                                                                           |
| ├─ Route     | EqualUnique | customer |             | UNKNOWN    | select `user`.id from `user` where `user`.id = 1                          |
[...]

PR #10715, opened 2022-07-17
BugFix: Keep predicates in join when pushing new ones

GuptaManan100:
## Description

This PR fixes the bug found in https://github.com/vitessio/vitess/issues/10713. 
The plan for the query found there with Gen4 using the vschema provided was - 
```json
{
  "QueryType": "SELECT",
  "Original": "select user.id, team.id from user join team_member on user.id = team_member.user join team on team_member.team = team.id where user.email = 'alice@gmail.com'",
  "Instructions": {
    "OperatorType": "Join",
    "Variant": "Join",
    "JoinColumnIndexes": "L:0,R:0",
    "JoinVars": {
      "user_id": 0
    },
    "TableName": "`user`_team, team_member",
    "Inputs": [
      {
        "OperatorType": "Route",
        "Variant": "EqualUnique",
        "Keyspace": {
          "Name": "customer",
          "Sharded": true
        },
        "FieldQuery": "select `user`.id from `user` where 1 != 1",
        "Query": "select `user`.id from `user` where `user`.email = 'alice@gmail.com'",
        "Table": "`user`",
        "Values": [
          "VARCHAR(\"alice@gmail.com\")"
        ],
        "Vindex": "email_user_map"
      },
      {
        "OperatorType": "Route",
        "Variant": "Scatter",
        "Keyspace": {
          "Name": "customer",
          "Sharded": true
        },
        "FieldQuery": "select team.id from team_member, team where 1 != 1",
        "Query": "select team.id from team_member, team where team_member.`user` = :user_id",
        "Table": "team, team_member"
      }
    ]
  }
}
```

In the plan, it is clearly visible that the predicate `team_member.team = team.id` is not applied anywhere. This results in extra rows being returned due to insufficient filtering as reported by the user.

After the changes in the PR, the plan looks like 
```json
{
  "QueryType": "SELECT",
  "Original": "select user.id, team.id from user join team_member on user.id = team_member.user join team on team_member.team = team.id where user.email = 'alice@gmail.com'",
  "Instructions": {
    "OperatorType": "Join",
    "Variant": "Join",
    "JoinColumnIndexes": "L:0,R:0",
    "JoinVars": {
      "user_id": 0
    },
    "TableName": "`user`_team, team_member",
    "Inputs": [
      {
        "OperatorType": "Route",
        "Variant": "EqualUnique",
        "Keyspace": {
          "Name": "customer",
          "Sharded": true
        },
        "FieldQuery": "select `user`.id from `user` where 1 != 1",
        "Query": "select `user`.id from `user` where `user`.email = 'alice@gmail.com'",
        "Table": "`user`",
        "Values": [
          "VARCHAR(\"alice@gmail.com\")"
        ],
        "Vindex": "email_user_map"
      },
      {
        "OperatorType": "Route",
        "Variant": "Scatter",
        "Keyspace": {
          "Name": "customer",
          "Sharded": true
        },
        "FieldQuery": "select team.id from team_member, team where 1 != 1",
        "Query": "select team.id from team_member, team where team_member.`user` = :user_id and team_member.team = team.id",
        "Table": "team, team_member"
      }
    ]
  }
}
```
Now, the predicate is found in the last route.

The earliest release affected by the bug is 13.0 and as such, the PR should be backported to both release-14.0 and release-13.0

## Related Issue(s)

- https://github.com/vitessio/vitess/issues/10713

## Checklist

-   [x] "Backport me!" label has been added if this change should be backported - Both to release-13.0 and release-14.0

### DIFF
--- a/go/test/endtoend/vtgate/misc_test.go
+++ b/go/test/endtoend/vtgate/misc_test.go
@@ -828,3 +828,25 @@ func TestEmptyQuery(t *testing.T) {
 	utils.AssertContainsError(t, conn, ";", "Query was empty")
 	utils.AssertIsEmpty(t, conn, "-- this is a comment")
 }
+
+// TestJoinWithMergedRouteWithPredicate checks the issue found in https://github.com/vitessio/vitess/issues/10713
+func TestJoinWithMergedRouteWithPredicate(t *testing.T) {
+	defer cluster.PanicHandler(t)
+	ctx := context.Background()
+	conn, err := mysql.Connect(ctx, &vtParams)
+	require.NoError(t, err)
+	defer conn.Close()
+
+	utils.Exec(t, conn, "delete from t1")
+	defer utils.Exec(t, conn, "delete from t1")
+	utils.Exec(t, conn, "delete from t2")
+	defer utils.Exec(t, conn, "delete from t2")
+	utils.Exec(t, conn, "delete from t3")
+	defer utils.Exec(t, conn, "delete from t3")
+
+	utils.Exec(t, conn, "insert into t1 (id1,id2) values (1, 13)")
+	utils.Exec(t, conn, "insert into t2 (id3,id4) values (5, 10), (15, 20)")
+	utils.Exec(t, conn, "insert into t3 (id5,id6,id7) values (13, 5, 8)")
+
+	utils.AssertMatches(t, conn, "select t3.id7, t2.id3, t3.id6 from t1 join t3 on t1.id2 = t3.id5 join t2 on t3.id6 = t2.id3 where t1.id2 = 13", `[[INT64(8) INT64(5) INT64(5)]]`)
+}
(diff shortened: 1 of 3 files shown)


harshit-gangal:
The changes looks good.
Can you add an e2e test as well?

I’m not sure about v13 backport as Gen4 was experimental on that.

frouioui (approved):
the fix looks good to me

GuptaManan100:
@harshit-gangal I and @frouioui think we should backport to 13.0 too since the bug is making Gen4 give incorrect results. We wouldn't have backported if the bug was an unsupported query or failing with an error, but incorrect results even in v13 is not a good user experience. They won't know that the query is giving incorrect results and it will be hard for us to document which kind of queries won't work. WDYT?

harshit-gangal:
> @harshit-gangal I and @frouioui think we should backport to 13.0 too since the bug is making Gen4 give incorrect results. We wouldn't have backported if the bug was an unsupported query or failing with an error, but incorrect results even in v13 is not a good user experience. They won't know that the query is giving incorrect results and it will be hard for us to document which kind of queries won't work. WDYT?

we can keep correctness over experimental feature :+1:

harshit-gangal (approved):

systay (approved):

systay:
if it's easy to backport to r13, why not just do it, right?

frouioui:
> if it's easy to backport to r13, why not just do it, right?

That's what @GuptaManan100 and I were thinking 👍🏻

PR #10716, opened 2022-07-17
ApplySchema: deprecate '--skip_preflight' flag

shlomi-noach:
## Description

This PR deprecates the flag `--skip_preflight` in `vtctlclient ApplySchema`. The new behavior is as if `--skip_preflight=true`. That is, to always skip preflight. In fact, the preflight code is completely removed in this PR.

The Vitess team discussed deprecating `--skip_preflight` a long time ago, with the advent of Online DDL. Moreover, we have noticed that in production environments we always set the flag. And furthermore, the logic used by `--skip_preflight` is not on par with new schema change logic: Online DDL and `schemadiff`, and we've encountered scenarios where the logic was flawed.

As `Vitess` now recommends running migrations via Online DDL, we should move away from `preflight` checks.

## Related Issue(s)

#6926 

## Checklist


mattlord:
I feel that deprecating the flag would be setting the default to true and adding a deprecation warning in the CLI and docs while removal would be removing the code done here along the the preflight tablet manager client code. The PR seems to be in the middle. Do we need to do a deprecation cycle? What's the value of keeping the flag if you can't set it?

Please keep in mind that I am glad we're getting rid of this and appreciate  you working on that. 🙂

shlomi-noach:
Added deprecation warning

go/cmd/vtctldclient/command/schema.go
```diff
@@ -141,7 +141,7 @@ func commandApplySchema(cmd *cobra.Command, args []string) error {
 		AllowLongUnavailability: applySchemaOptions.AllowLongUnavailability,
 		DdlStrategy:             applySchemaOptions.DDLStrategy,
 		Sql:                     parts,
-		SkipPreflight:           applySchemaOptions.SkipPreflight,
+		SkipPreflight:           true,
```
ajm188, line 144: same comments I made in https://github.com/vitessio/vitess/pull/10717#discussion_r928711419, but for this flag

shlomi-noach: :+1:

deepthi (dismissed):

go/cmd/vtctldclient/command/schema.go
shlomi-noach: Comments addressed

deepthi (approved):

go/vt/vtctl/vtctl.go
```diff
@@ -3110,7 +3112,7 @@ func commandApplySchema(ctx context.Context, wr *wrangler.Wrangler, subFlags *fl
 		AllowLongUnavailability: *allowLongUnavailability,
 		DdlStrategy:             *ddlStrategy,
 		Sql:                     parts,
-		SkipPreflight:           *skipPreflight,
+		SkipPreflight:           true,
```
deepthi, line 3115: Perhaps we should be deprecating the protobuf field for this as well? I'm not certain about that though - whether we do this now and deprecate the protobuf field in the next release.
@ajm188 what do you suggest?

ajm188: i know i just said "yes" on the other PR, but i think we need to wait one more cycle, since old tablets will be consuming this field, which will take the zero value (`false`) instead of what we unconditionally pass here (`true`).

so deprecate flag + always pass `true` (v15) => remove flag + deprecate protobuf field + tablets ignore value and always skip-preflight (v16)

shlomi-noach:
`Cluster (vtgate_schema)` seems to be consistently failing. Unsure yet how this PR affects it -- but it looks like it does.

shlomi-noach:
This is not stale. I'm just stuck. Can't solve the CI issue.

shlomi-noach:
Still stuck here.

shlomi-noach:
Fashionably late, this PR is ready for review!

dbussink (approved):
Yay code cleanup!

vmg (approved):

shlomi-noach:
Current error in `Upgrade Downgrade Testing Query Serving (Schema)`:
```
2023-02-27T15:36:44.2867788Z ERROR: Go version reported: go version go1.19.4 linux/amd64. Version 1.20.1+ required. See https://vitess.io/contributing/build-from-source for install instructions.
2023-02-27T15:36:44.2873075Z ##[error]Process completed with exit code 1.
```

shlomi-noach:
`Upgrade Downgrade Testing Query Serving (Schema)` failure is a known issue to be fixed imminently; meanwhile merging this PR.

PR #10719, opened 2022-07-17
Tablet Executor: consolidate ReloadSchema calls, and skip for Online DDL

shlomi-noach:
## Description

This PR optimizes calls to `ApplySchema`, as following:

### Consolidate ReloadSchema

Previously, and when multiple SQLs are received, `ApplySchema` would call a `ReloadSchema` after every single SQL execution, and wait on replica to catch up. This is wasteful, because the schema is known to change again and again for next SQLs.

As of this PR, TabletExecutor first applies all SQLs (still one by one), and then, once for each shard, waits on replication lag and calls `ReloadSchema`.

### Skip ReloadSchema for Online DDL

Online DDL calls are asynchronous. The schema is not going to change after submitting the Online DDL anyway. If all SQLs are Online DDLs, then we skip `ReloadSchema` altogether.

We make no change to existing tests. We expect this refactor to pass all tests.

## Related Issue(s)

#6926 

## Checklist

### DIFF
--- a/go/vt/schemamanager/tablet_executor.go
+++ b/go/vt/schemamanager/tablet_executor.go
@@ -276,10 +276,10 @@ func (exec *TabletExecutor) preflightSchemaChanges(ctx context.Context, sqls []s
 
 // executeSQL executes a single SQL statement either as online DDL or synchronously on all tablets.
 // In online DDL case, the query may be exploded into multiple queries during
-func (exec *TabletExecutor) executeSQL(ctx context.Context, sql string, providedUUID string, execResult *ExecuteResult) error {
+func (exec *TabletExecutor) executeSQL(ctx context.Context, sql string, providedUUID string, execResult *ExecuteResult) (executedAsynchronously bool, err error) {
 	stmt, err := sqlparser.Parse(sql)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


dbussink (approved):
> Online DDL calls are asynchronous. The schema is not going to change after submitting the Online DDL anyway. If all SQLs are Online DDLs, then we skip `ReloadSchema` altogether.

Is there any mechanism that calls `ReloadSchema` after an Online DDL completes? 

I think this is a separate question though from the issue here. We don't have that working today either yet it looks like because the reload runs right after the DDL is scheduled right and not when completed?

That makes this not a regression I think but a separate maybe existing issue, so I think this change makes sense.

shlomi-noach:
> Is there any mechanism that calls ReloadSchema after an Online DDL completes?

There is a specific call to `ReloadSchema` after a VReplication-based migration. But `CREATE`, `DROP`, and `gh-ost` migrations do not have an associated `ReloadSchema`.

The other thing is, something we've set our eyes on, long ago, that there is no need to reload the entire schema when a single table changes. But the fix for that is going to be elaborate.

Agreed that we should look into reloading the schema after any schema change.

PR #10639, opened 2022-07-06
VDiff2: Auto retry to continue on error

mattlord:
## Description

We now record and display errors for a VDiff on a per shard basis:
```
$ vtctlclient VDiff -- --v2 customer.commerce2customer show last

VDiff Summary for customer.p1c2 (c6e8698e-01b3-11ed-b76d-920702940ee0)
State:        error
              Error: (shard -80) vttablet: rpc error: code = Unknown desc = (errno 1213) (sqlstate 40001): Deadlock found when trying to get lock; try restarting transaction
              Error: (shard 80-) vttablet: rpc error: code = Unknown desc = (errno 1213) (sqlstate 40001): Deadlock found when trying to get lock; try restarting transaction
RowsCompared: 105
HasMismatch:  false
StartedAt:    2022-07-12 07:25:17

Use "--format=json" for more detailed output.


$ vtctlclient VDiff -- --v2 --format=json customer.commerce2customer show last
{
	"Workflow": "p1c2",
	"Keyspace": "customer",
	"State": "error",
	"UUID": "c6e8698e-01b3-11ed-b76d-920702940ee0",
	"RowsCompared": 106,
	"HasMismatch": false,
	"Shards": "-80,80-",
	"StartedAt": "2022-07-12 07:25:17",
	"Errors": {
		"-80": "vttablet: rpc error: code = Unknown desc = (errno 1213) (sqlstate 40001): Deadlock found when trying to get lock; try restarting transaction",
		"80-": "vttablet: rpc error: code = Unknown desc = (errno 1213) (sqlstate 40001): Deadlock found when trying to get lock; try restarting transaction"
	}
}
```

And we automatically retry those VDiffs (every 30 seconds) that ended with an error, when the auto-retry option has been set for it (now defaults to true) and we can see that it's an ephemeral error which warrants an immediate retry. For example (from a manual local test):
```
$ vtctlclient VDiff -- --v2 --format=json --auto-retry customer.commerce2customer show last
	"Workflow": "commerce2customer",
	"Keyspace": "customer",
	"State": "error",
	"UUID": "da4c5e14-fece-11ec-a217-920702940ee0",
	"RowsCompared": 2,
	"HasMismatch": false,
	"Shards": "0",
	"StartedAt": "2022-07-08 15:01:32",
	"Errors": {
		"0": "vttablet: rpc error: code = Unknown desc = (errno 1213) (sqlstate 40001): Deadlock found when trying to get lock; try restarting transaction"
	}
}

[...]

PR #10720, opened 2022-07-17
Reduce Flakiness of ERS/PRS e2e Tests Using Retries With a Timeout

mattlord:
## Description

The `ers_prs_heavy` workflow has been very flaky in this PR: https://github.com/vitessio/vitess/pull/10639. Often having to run 4+ times to succeed. I suspect that all of [the new DDL](https://github.com/vitessio/vitess/pull/10639/files#diff-0842ac0cabd40352e88bfcb98b9a9ae213d441b11eabc4bc51afbd47e211db56) executed at tablet init when the VDiff engine opens is causing replication to be slightly slower and thus we're getting this basic error from various individual tests (the ID and specific test differing on the failed runs):
```
2022-07-17T21:32:58.8852569Z     utils.go:348: 
2022-07-17T21:32:58.8852912Z         	Error Trace:	utils.go:354
2022-07-17T21:32:58.8853472Z         	            				utils.go:348
2022-07-17T21:32:58.8853942Z         	            				utils.go:526
2022-07-17T21:32:58.8854402Z         	            				utils.go:460
2022-07-17T21:32:58.8854875Z         	            				ers_test.go:453
2022-07-17T21:32:58.8855872Z         	Error:      	Expected nil, but got: &mysql.SQLError{Num:1146, State:"42S02", Message:"Table 'vt_ks.vt_insert_test' doesn't exist", Query:"select msg from vt_insert_test where id=20"}
```

This patch was [tested and verified here](https://github.com/vitessio/vitess/pull/10639/commits/bcef7958eb76fbc812600265fc837ca135f4299e) and it [passed on the first try](https://github.com/vitessio/vitess/runs/7380971321?check_suite_focus=true). 🥳 It was then re-tested and re-verified after [this minor follow-up](https://github.com/vitessio/vitess/pull/10639/commits/1225cb7e90e3bba0c79d285e14e13a09493f60c7) and it [passed on the first try again](https://github.com/vitessio/vitess/runs/7381224980?check_suite_focus=true). 🥳 🥳 

## Related Issue(s)
  - https://github.com/vitessio/vitess/pull/10639

## Checklist

### DIFF
--- a/go/test/endtoend/reparent/emergencyreparent/ers_test.go
+++ b/go/test/endtoend/reparent/emergencyreparent/ers_test.go
@@ -22,7 +22,6 @@ import (
 	"testing"
 	"time"
 
-	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
 	"vitess.io/vitess/go/test/endtoend/cluster"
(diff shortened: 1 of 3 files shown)


GuptaManan100 (approved):
This is great! I had seen this failure a couple of times before too and had this on my TODO list for a while now. Thank you for making this change.

PR #10725, opened 2022-07-18
VDiff2: ignore errors while attempting to purge vdiff tables 

### LINKED ISSUE #10726: Some vreplication workflow actions log vdiff2 errors regarding purging of vdiff tables
### Overview of the Issue

While running the local example we see errors that vdiff2 tables in`_vt` fail because the tables do not exist. 

These logs will be seen anytime one completes or cancels a workflow and the user has never started a vdiff2 workflow until then.

### Reproduction Steps

Run local example

### Binary Version

```sh
`58f12557465d6bbaa55533873cda0673ef23ae2a`. This will also be present in the current `release-14.0`
```

```sh
-
```

### Log Fragments

_No response_

rohit-nayak-ps:
## Description

It is possible that the _vt.vdiff tables do not exist when attempting to purge records when a vreplication workflow is deleted, so we no longer log the resulting errors

## Related Issue(s)

fixes #10726

## Checklist

### DIFF
--- a/go/vt/wrangler/traffic_switcher.go
+++ b/go/vt/wrangler/traffic_switcher.go
@@ -57,9 +57,6 @@ const (
 	renameTableTemplate = "_%.59s_old" // limit table name to 64 characters
 
 	sqlDeleteWorkflow = "delete from _vt.vreplication where db_name = %s and workflow = %s"
-	sqlDeleteVDiffs   = `delete from vd, vdt, vdl using _vt.vdiff as vd inner join _vt.vdiff_table as vdt on (vd.id = vdt.vdiff_id)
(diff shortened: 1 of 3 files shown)


mattlord (changes requested):
I would propose only ignoring the error completely if the table's don't exist:
```
			if sqlErr, ok := err.(*mysql.SQLError); ok && sqlErr.Num != mysql.ERNoSuchTable { // the tables may not exist if no vdiffs have been run
				ts.Logger().Infof("Error deleting vdiff data for %s.%s workflow: %v", source.GetPrimary().Keyspace, ts.reverseWorkflow, err)
			}
```

mattlord (approved):
LGTM! Thanks for doing this!

PR #10727, opened 2022-07-18
ApplySchema: renew keyspace lock while iterating SQLs

shlomi-noach:
## Description

Followup to #10719. Before applying DDLs, `TabletExecutor` acquires a _keyspace_ lock. It then iterates all SQLs, applies them, and finally releases the lock.

When there's a large number of SQLs, the iteration can take a lon gtime. #10719 already reduces that time. However, applying SQLs can still take longer than the lock timeout. What we've seen happening was that SQLs were applied, but when it came to unlocking the keyspace, the lock was expired and the unlock operation failed.

In this PR we renew the lock on each SQL iteration. This is a naive implementation and perhaps too aggressive. We could later refine this to renew once per X seconds, for example. **UPDATE:** now calls are rate limited.

With this change, the lock will not expire on reasonable DDLs. Specifically, Online DDLs are very quick and are never expected to run long. Large `direct` `ALTER`s may still exceed lock timeout. This PR does not attempt to solve that scenario.

We have not changes the tests here. We've changes internal implementation of `CheckKeyspaceLocked()` and a re expecting to see all tests passing. In case of a problem, we will restore original behavior of this function, and create a new function which we will then use in `TabletExecutor`. **U{DATE**: refactored into a new function.

## Related Issue(s)

#6926 
#10719 

## Checklist

cc @dbussink

### DIFF
--- a/go/vt/schemamanager/tablet_executor.go
+++ b/go/vt/schemamanager/tablet_executor.go
@@ -36,10 +36,6 @@ import (
 	"vitess.io/vitess/go/vt/vttablet/tmclient"
 )
 
-var (
-	lockRenewTimeout = 10 * time.Second // has to be a fraction of Topo's defaultLockTimeout (30sec at this time)
-)
-
 // TabletExecutor applies schema changes to all tablets.
 type TabletExecutor struct {
 	migrationContext     string
@@ -363,7 +359,7 @@ func (exec *TabletExecutor) Execute(ctx context.Context, sqls []string) *Execute
 	}
 	providedUUID := ""
 
-	rl := timer.NewRateLimiter(lockRenewTimeout)
+	rl := timer.NewRateLimiter(*topo.RemoteOperationTimeout / 4)
 	defer rl.Stop()
 
 	syncOperationExecuted := false
--- a/go/vt/schemamanager/tablet_executor.go
+++ b/go/vt/schemamanager/tablet_executor.go
@@ -368,8 +368,8 @@ func (exec *TabletExecutor) Execute(ctx context.Context, sqls []string) *Execute
 
 	syncOperationExecuted := false
 	for index, sql := range sqls {
+		// Attempt to renew lease:
 		if err := rl.Do(func() error { return topo.CheckKeyspaceLockedAndRenew(ctx, exec.keyspace) }); err != nil {
-			// This renews the lock lease
 			execResult.ExecutorErr = vterrors.Wrapf(err, "CheckKeyspaceLocked in ApplySchemaKeyspace %v", exec.keyspace).Error()
 			return &execResult
 		}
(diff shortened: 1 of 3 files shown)


shlomi-noach:
Refactored: restored original `CheckKeyspaceLocked()` function, created new `CheckKeyspaceLockedAndRenew()`. We now call it rate-limited so as not to overwhelm `ApplySchema` and `topo`.

go/vt/schemamanager/tablet_executor.go
```diff
@@ -357,8 +363,16 @@ func (exec *TabletExecutor) Execute(ctx context.Context, sqls []string) *Execute
 	}
 	providedUUID := ""
 
+	rl := timer.NewRateLimiter(lockRenewTimeout)
+	defer rl.Stop()
```
mattlord, line 367: Nice to see this getting more usage! I will eventually use it for something too... just have to remember. 🙂

mattlord (approved):
I like it! I think there's other places we'll want to use this new `CheckKeyspaceLockedAndRenew` function over time. I remember seeing other places where we had lost the lease during a long running operation and it was unexpected (in the code and in my mind). Might have been a `SwitchTraffic` call...

go/vt/schemamanager/tablet_executor.go
```diff
@@ -31,9 +32,14 @@ import (
 	"vitess.io/vitess/go/vt/sqlparser"
 	"vitess.io/vitess/go/vt/topo"
 	"vitess.io/vitess/go/vt/vtctl/schematools"
+	"vitess.io/vitess/go/vt/vterrors"
 	"vitess.io/vitess/go/vt/vttablet/tmclient"
 )
 
+var (
+	lockRenewTimeout = 10 * time.Second // has to be a fraction of Topo's defaultLockTimeout (30sec at this time)
+)
```
GuptaManan100, line 41: Should this be a flag, living in the same location as `remote_operation_timeout`. We allow users to tweek the remote-operation timeout which is what is used in lock shard. This should either be set to a fraction of it or be a flag too.

shlomi-noach: Removed the variable, and set the rate limiter to use `*topo.RemoteOperationTimeout / 4`

GuptaManan100: Thankyou! This change is awesome ❤️. I'll for sure use this refreshing of the lease in VTOrc.

GuptaManan100 (approved):

shlomi-noach:
I'm lost with the test failures. `vreplication_v2` keeps failing, on different points, but consistently.

shlomi-noach:
Plausible explanation to test failures found! Let's see if https://github.com/vitessio/vitess/pull/10727/commits/6327e58e25fdd0227b7f8614b7abc41981288b9f solves this

PR #10728, opened 2022-07-18
Cleanup: ERS and PRS tests by removing setupShardLegacy

GuptaManan100:
## Description

This PR cleans up the ERS and PRS tests by removing `setupShardLegacy`. We should use `setupShard` instead since we did the release. If we need version specific flags, then we should use the `version` information that we store now instead of duplicating the setup functions.

## Checklist


frouioui (approved):
Looks good to me

deepthi (approved):

PR #10735, opened 2022-07-18
Flakes: Increase timeouts for upgrade_downgrade workflows

mattlord:
## Description

The upgrade_downgrade workflows as a whole are somewhat flaky, with a common cause of failure being hitting the default 10m timeout (for example [here](https://github.com/vitessio/vitess/runs/7397201740?check_suite_focus=true) and [here](https://github.com/vitessio/vitess/runs/7397508682?check_suite_focus=true)).

This bumps the individual test timeouts used in the upgrade_downgrade workflows (by tag) from 10m to 20m and the timeout for the workflow as a whole to 60m. 

## Checklist


GuptaManan100 (approved):
I am going to refactor some of the tests in these packages to make them faster and reduce their count, but for now these changes are good!

Issue #10753, opened 2022-07-19
Bug Report: Gen4 planner handles `<=>` operator with `NULL` operand incorrectly


arthurschreiber:
### Overview of the Issue

Queries that uses that `<=>` operator with a `NULL` operand are not handled correctly in the Gen4 planner. They are treated as a `None` `variant` query.

### Reproduction Steps

This can be reproduced using the following query (using the `TestOne` test function from `go/vt/vtgate/planbuilder/plan_test.go`):

```
SELECT COUNT(*) FROM `music` WHERE NULL <=> music.song
```

Query plan for V3:

```json
        {
          "QueryType": "SELECT",
          "Original": "SELECT COUNT(*) FROM `music` WHERE music.song \u003c=\u003e NULL",
          "Instructions": {
            "OperatorType": "Aggregate",
            "Variant": "Scalar",
            "Aggregates": "sum_count(0) AS count",
            "Inputs": [
              {
                "OperatorType": "Route",
                "Variant": "Scatter",
                "Keyspace": {
                  "Name": "user",
                  "Sharded": true
                },
                "FieldQuery": "select count(*) from music where 1 != 1",
                "Query": "select count(*) from music where music.song \u003c=\u003e null",
                "Table": "music"
              }
            ]
          }
        }
```

Query plan for Gen4:

```json
        {
          "QueryType": "SELECT",
          "Original": "SELECT COUNT(*) FROM `music` WHERE music.song \u003c=\u003e NULL",
          "Instructions": {
            "OperatorType": "Aggregate",
            "Variant": "Scalar",
            "Aggregates": "sum_count_star(0) AS count(*)",
            "Inputs": [
              {
                "OperatorType": "Route",
                "Variant": "None",
                "Keyspace": {
                  "Name": "user",
                  "Sharded": true
                },
                "FieldQuery": "select count(*) from music where 1 != 1",
                "Query": "select count(*) from music where music.song \u003c=\u003e null",
                "Table": "music"
              }
            ]
          }
        }
```

### Binary Version

```sh
Version: 15.0.0-SNAPSHOT (Git revision 665cb4139ce2cc705b7d8d99b7c760f37a285ee1 branch 'main') built on Mon Jun 27 10:10:31 UTC 2022 by codespace@codespaces-7bc63a using go1.18.3 linux/amd64
```

### Operating System and Environment details

```sh
$ cat /etc/os-release
NAME="Ubuntu"
[...]

PR #10754, opened 2022-07-19
Fix gen4 planner handling of `<=>` operator with `NULL` operand

arthurschreiber:
## Description

Currently, all comparison operators that have a `NULL` operand bail out early and cause a query to not return any results. This behaviour is correct for all operators except for the null safe equality operator (`<=>`).

## Related Issue(s)

Fixes: https://github.com/vitessio/vitess/issues/10753

## Checklist


frouioui (approved):
Looks good to me! Thanks for finding the issue and fixing it ❤️

frouioui:
@arthurschreiber can you add an end-to-end test to test null comparison? Seems like we are lacking some tests around this type of predicate. You can add the test in `go/test/endtoend/vtgate/misc_test.go` or elsewhere in the same (sub)package.

arthurschreiber:
@frouioui Sure, I'll see if I can add one.

systay (approved):

arthurschreiber:
@frouioui I extended an existing end to end test case with use of the `<=>` operator. Is that what you had in mind? 🙇‍♂️

frouioui:
> @frouioui I extended an existing end to end test case with use of the `<=>` operator. Is that what you had in mind? 🙇‍♂️

Looks great to me @arthurschreiber, thank you 🙏🏻

PR #10755, opened 2022-07-19
RateLimiter: exit goroutine at Stop()

shlomi-noach:
## Description

`RateLimiter` was leaking `goroutine` that never got destroyed. The fix is here.

## Checklist


dbussink (approved):
Let's not have a busy loop here and a goroutine spinning at 100% CPU when we stop the timer...

PR #10756, opened 2022-07-19
Audit/Tests for Go Leaks

mattlord:
## Description

I did a quick audit of places where we were using a confined Ticker — created as a local variable in a function — but not ensuring it gets stopped with a defer statement. This will help to prevent resource leaks. This was prompted by the fact I realized I forgot to do this in a PR yesterday: https://github.com/vitessio/vitess/pull/10720

I'm also investigating usage of this: https://github.com/uber-go/goleak
Unfortunately this does not seem useful to use in e2e tests.

## Related Issue(s)

  - Follow-up to: https://github.com/vitessio/vitess/pull/10720

## Checklist


go/vt/vttablet/tabletserver/tabletserver.go
```diff
@@ -349,6 +349,7 @@ func (tsv *TabletServer) InitACL(tableACLConfigFile string, enforceTableACLConfi
 
 	if reloadACLConfigFileInterval != 0 {
 		ticker := time.NewTicker(reloadACLConfigFileInterval)
+		defer ticker.Stop()
```
rohit-nayak-ps, line 352: Don't we want this timer to keep running in the background to reload the acl config every `reloadACLConfigFileInterval` seconds?

mattlord: Hmm, you're right! Good catch! We have a never ending for loop below but that's in a goroutine so we don't want to stop the ticker when this outer function ends. Will change.

go/mysql/collations/tools/makecolldata/contractions.go
```diff
@@ -58,7 +58,7 @@ func printContraction1(g *codegen.Generator, wa *weightarray, incont []uca.Contr
 			trie[r] = append(trie[r], cont)
 		}
 		if depth == len(cont.Path) {
-			leaf = &cont
+			leaf = &cont // nolint:exportloopref
```
deepthi, line 61: Fixed differently by @dbussink in #10765

deepthi (commented):
This looks good in principle.
What did you mean by the `goleak` library not being useful in e2e tests?

mattlord:
> This looks good in principle.
> What did you mean by the `goleak` library not being useful in e2e tests?

Seems that it only tests the main command go test executes whereas the e2e tests start up a bunch of processes.

Issue #10627, opened 2022-06-30
Bug Report: `--queryserver-config-terse-errors` does more than just redact bind variables, obscuring crucial error details


aquarapid:
### Overview of the Issue

If you have `--queryserver-config-terse-errors` configured on your `vttablet` and you exceed the `vttablet` row limit (in this case the limit is 10000 rows), you will see something like:

```
mysql> select * from sbtest1 limit 10001;                     
ERROR 10001 (HY000): target: unsharded.0.primary: vttablet: rpc error: code = Aborted desc = (errno 10001) (sqlstate HY000) (CallerID: user): Sql: "select id, k, c, pad from sbtest1 limit :vtg1", BindVars: {[REDACTED]
```

This generic abort message can be confusing to application-side users, since it's not clear why the query was aborted.

Contrast with the error if you do not have `--queryserver-config-terse-errors`:

```
mysql> select * from sbtest1 limit 10001;
ERROR 10001 (HY000): target: unsharded.0.primary: vttablet: rpc error: code = Aborted desc = Row count exceeded 10000 (errno 10001) (sqlstate HY000) (CallerID: user): Sql: "select id, k, c, pad from sbtest1 limit :vtg1", BindVars: {#maxLimit: "type:INT64 value:\"10001\""vtg1: "type:INT64 value:\"10001\""
```

This message tells you what limit was exceeded, and what the value of the limit was.

### Reproduction Steps

See Overview

### Binary Version

```sh
Version: 15.0.0-SNAPSHOT (Git revision 9199470f84406a62b8e70c7766df53f3e6e46c96 branch 'main') built on Wed Jun 29 17:29:14 PDT 2022 by jacques@dhoomdeskw.localdomain using go1.18.3 linux/amd64

but older versions also affected
```

### Operating System and Environment details

```sh
Fedora 36 amd64
```

aquarapid:
To expand slightly;  note that the help for `--queryserver-config-terse-errors` says:

```
prevent bind vars from escaping in client error messages
```

and nothing to indicate to the user that it will also non-bind variable error details, such as in this case.

PR #10762, opened 2022-07-20
BugFix: `--queryserver-config-terse-errors` shouldn't redact `Row count exceeded` error and shouldn't affect ignoring `Duplicate entry` error in lookup vindexes

### LINKED ISSUE #10777: Bug Report: Creating Lookups doesn't ignore duplicate entry error with `--queryserver-config-terse-errors`
### Overview of the Issue

If you have `--queryserver-config-terse-errors` configured on your `vttablet` and you try to do an insert that would lead to insertion in a lookup vindex already having that row, will cause the insert to fail, even though if it is run without the flag, the `duplicate entry` errors are ignored in vindex insertions.

### Reproduction Steps

Create a sharded database with table t1 and t1_id2_idx with the following schema - 
```SQL
create table t1
(
    id1 bigint,
    id2 bigint,
    primary key (id1)
) Engine = InnoDB;

create table t1_id2_idx
(
    id2         bigint,
    keyspace_id varbinary(10),
    primary key (id2)
) Engine = InnoDB;

```
and vschema -
```JSON
{
  "sharded": true,
  "vindexes": {
    "hash": {
      "type": "hash"
    },
    "t1_id2_vdx": {
      "type": "consistent_lookup_unique",
      "params": {
        "table": "t1_id2_idx",
        "from": "id2",
        "to": "keyspace_id"
      },
      "owner": "t1"
    }
  },
  "tables": {
    "t1": {
      "column_vindexes": [
        {
          "column": "id1",
          "name": "hash"
        },
        {
          "column": "id2",
          "name": "t1_id2_vdx"
        }
      ]
    },
    "t1_id2_idx": {
      "column_vindexes": [
        {
          "column": "id2",
          "name": "hash"
        }
      ]
    }
  }
}
```

then insert the following row - 
```SQL
insert into t1(id1, id2) values(1, 4)
```

Next, target the shard and delete only the main row and not the lookup entry -
```SQL
use `ks:-80`;
delete from t1 where id1=1
```

Try inserting again
```SQL
insert into t1(id1, id2) values(1, 4)
```

If `--queryserver-config-terse-errors` flag is not specified, then the query succeeds but fails if the flag is present. However, this behavior is non-intuitive since the flag is only about redacting information from the error messages returned to users.

### Binary Version

```sh
main
```

```sh
main
```

### Log Fragments

```sh
-
```

GuptaManan100:
## Description

This PR fixes the issue described in https://github.com/vitessio/vitess/issues/10627. Specifying `--queryserver-config-terse-errors` shouldn't redact the error message `Row count exceeded`.
This PR also fixes the issue described in https://github.com/vitessio/vitess/issues/10777. Specifying `--queryserver-config-terse-errors` shouldn't affect ignoring the `Duplicate entry` error in insertions of vindex rows.

As part of the PR, some tests have also been cleaned up to move to the new architecture of a `start` function deleting all the old rows and artifacts from the previous tests, so that each individual test no longer has to worry about cleanup.

## Related Issue(s)

- Fixes https://github.com/vitessio/vitess/issues/10627
- Fixes #10777 

## Checklist


systay (approved):

go/test/endtoend/vtgate/lookup_test.go
```diff
@@ -122,7 +92,7 @@ func TestConsistentLookup(t *testing.T) {
 	mysqlErr := err.(*mysql.SQLError)
 	assert.Equal(t, 1062, mysqlErr.Num)
 	assert.Equal(t, "23000", mysqlErr.State)
-	assert.Contains(t, mysqlErr.Message, "Duplicate entry")
+	assert.Contains(t, mysqlErr.Message, "reverted partial DML execution")
```
harshit-gangal, line 95: this is masking duplicate error. we should return the original error as well.

GuptaManan100: It is expected to mask the duplicate error. We are using `queryserver-config-terse-errors` now and the `Duplicate entry` error message has PII information, so we don't return that.

harshit-gangal: It is hard to like them! **_terse-errors_**

harshit-gangal (approved):

PR #10766, opened 2022-07-20
Online DDL: issue a ReloadSchema at the completion of any migration

shlomi-noach:
## Description

As we look to reduce the footprint of `ApplySchema`, we removed excessive calls to `ReloadSchema` (see linked PRs follwing). We also noted that any `ReloadSchema` for Online DDL operations is irrelevant in `ApplySchema` because the actual schema change runs asynchronously, and almost certainly _after_ `ApplySchema` completes.

In this PR we make sure to call `ReloadSchema` at the end of any successful schema change. Up till now, this was only true for `vitess` `ALTER` migrations. Not it applies to `CREATE`, `DROP`, `gh-ost`, `pt-osc` migrations, including "fast" migrations and View migrations.

In the future we wish to further reduce the footprint by only reloading a single table, instead of the entire schema.

No tests are changed. We expect all existing tests to pass.

## Related Issue(s)

- https://github.com/vitessio/vitess/pull/10739
- https://github.com/vitessio/vitess/pull/10719
- https://github.com/vitessio/vitess/pull/10727 
- #6926 

## Checklist

### DIFF
--- a/go/vt/vttablet/onlineddl/executor.go
+++ b/go/vt/vttablet/onlineddl/executor.go
@@ -150,6 +150,7 @@ type Executor struct {
 	pool                  *connpool.Pool
 	tabletTypeFunc        func() topodatapb.TabletType
 	ts                    *topo.Server
+	tmClient              tmclient.TabletManagerClient
 	lagThrottler          *throttle.Throttler
 	toggleBufferTableFunc func(cancelCtx context.Context, tableName string, bufferQueries bool)
 	tabletAlias           *topodatapb.TabletAlias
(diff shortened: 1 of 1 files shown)


go/vt/vttablet/onlineddl/executor.go
```diff
@@ -1436,6 +1440,7 @@ exit $exit_code
 			return err
 		}
 		// Migration successful!
+		defer e.reloadSchema(ctx)
```
shlomi-noach, line 1443: reload schema at the end of a `gh-ost` migration

go/vt/vttablet/onlineddl/executor.go
```diff
@@ -1660,6 +1665,7 @@ export MYSQL_PWD
...
```
shlomi-noach, line 1668: reload schema at the end of a `gh-ost` migration

dbussink (approved):
Looking good. I think this then makes things consistent that we always reload after changes and never uselessly before 🎉

PR #10767, opened 2022-07-20
Wrong length creating the buffer needed to ask for full auth

ofiriluz:
## Description
Minor bug fix where in the full auth flow of caching sha2 the buffer created has the wrong length to write 1 byte instead of 2, causing a panic

## Related Issue(s)

https://github.com/vitessio/vitess/issues/10769

## Checklist


frouioui:
Hello @ofiriluz, thank you for your first contribution. Can you create an issue to describe the bug more in-depth and how to reproduce it?

ofiriluz:
> Hello @ofiriluz, thank you for your first contribution. Can you create an issue to describe the bug more in-depth and how to reproduce it?

Done, https://github.com/vitessio/vitess/issues/10769

dbussink:
@ofiriluz If you ensure you rebase onto latest `main`, the linter error in CI would also be fixed. 

Separately, would you be able to add a test for this?

dbussink:
I think we still want this fix, but would be good to have a test for this.

dbussink (approved):
I have added a test here to ensure we cover the fallback path now through the auth server logic.

harshit-gangal (approved):

PR #10768, opened 2022-07-20
Skip CI workflows on `push` for pull requests

frouioui:
## Description

This Pull Request changes the GitHub Actions workflows to skip all the steps when a workflow is triggered by a `push` event and is made against a feature branch.

This will allow us to run the vitess-bot's auto-backport feature without overcrowding the CI's queue.

This Pull Request was created on a `vitessio/vitess` branch and thus creates twice as many workflows: those triggered by `push` and those triggered by `pull_request` events. All the `push` workflows were completed in seconds since all their steps got skipped.

## Checklist


.github/workflows/check_make_vtadmin_authz_testgen.yml
```diff
@@ -6,10 +6,22 @@ jobs:
     name: Check Make vtadmin_authz_testgen
     runs-on: ubuntu-latest
     steps:
+    - name: Check if workflow needs to be skipped
+      id: skip-workflow
+      run: |
+        skip=0
+        if [[ "${{github.event.pull_request}}" ==  "" ]] && [[ "${{github.ref}}" != "refs/heads/main" ]] && [[ ! "${{github.ref}}" =~ "refs/heads/release-.*\.0" ]] && [[ ! "${{github.ref}}" =~ "refs/tags/.*" ]]; then
```
mattlord, line 13: Nit, but I think for production code we should try be more precise with regex's. For example:
```
"^refs/heads/release-[0-9]+\.[0-9]$"
```

The second part is so that we don't have to find places like this and fix them if we ever decide to start doing minor releases (e.g. 14.1.0).

.github/workflows/check_make_vtadmin_authz_testgen.yml
```diff
@@ -6,10 +6,22 @@ jobs:
...
+          skip=1
```
mattlord, line 14: Another nit, but I think we should be consistent and use `true` / `false` like we already are in the workflows rather than 1/0. It's also more explicit for humans reading the code, which I like.

mattlord (approved):
A couple of nits, but LGTM. I'll leave it to you to decide. Thanks!

.github/workflows/check_make_vtadmin_authz_testgen.yml
frouioui: Done!

.github/workflows/check_make_vtadmin_authz_testgen.yml
frouioui: Done too! :)

Issue #10676, opened 2022-07-12
Feature Request: Replace `serve` npm dependency for vtadmin-web


pbibra:
### Feature Description

We're currently using a library called `serve` that depends on a vulnerable version of `minimatch` (version 3.0.4). It looks like the `serve` package is not well maintained. Need to replace this with an alternative (maybe nginx). My understanding is that there is and effort to make VTAdmin a single component (https://github.com/vitessio/vitess/pull/10214). Not sure the priority of this though. 

### Use Case(s)

The current `serve` library depends on a vulnerable `minimatch` version 3.0.4

PR #10770, opened 2022-07-20
using nginx for vtadmin web

pbibra:

## Description

Need to move away from using the serve library due `PRISMA-2022-0039 (CVSS 7.5) - minimatch 3.0.4`
It's underlying dependency serve-handler uses a vulnerable version of minimatch and the dependency itself is not well maintained. Recently learned that minimatch is also a dependency of react-dev-utils which uses recursive-readdir; also not well mainained.

This PR uses a multistage docker build to first build the static content using node/npm and serve it using nginx in the final image. This way, our dependencies and any related vulnerabilities they may have are not leaked into the deployed image.

## Related Issue

https://github.com/vitessio/vitess/issues/10676

## Testing Notes

Image was built using `docker build -t vtadmin-test`, successfully runs on local host after `docker run -p 8080:14201 vtadmin-test:latest`


docker/k8s/vtadmin/entrypoint.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+set -x
+
+cat /etc/nginx/nginx.conf
```
L3o-pold, line 5: why are we displaying the config here?

docker/k8s/vtadmin/Dockerfile
```diff
@@ -47,9 +53,15 @@ RUN deluser node && \
    chown -R vitess:vitess /vt && \
    chown -R vitess:vitess /vtdataroot
 
-USER vitess
+FROM nginx:latest AS nginx
```
L3o-pold, line 56: use a static version here

docker/k8s/vtadmin/Dockerfile
```diff
@@ -17,11 +17,13 @@ ARG DEBIAN_VER=bullseye-slim
 
 FROM vitess/k8s:${VT_BASE_VER} AS k8s
 
+COPY nginx.conf nginx.conf
+COPY entrypoint.sh /entrypoint.sh
+
```
L3o-pold, line 22: we should not copy files here see comment bellow

docker/k8s/vtadmin/Dockerfile
```diff
@@ -36,6 +38,10 @@ COPY --from=k8s /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificate
 # copy web admin files
 COPY --from=k8s /vt/web/vtadmin /vt/web/vtadmin
 
+# copy nginx config
+COPY --from=k8s nginx.conf nginx.conf
+COPY --from=k8s entrypoint.sh /entrypoint.sh
```
L3o-pold, line 43: you should remove the from arg

docker/k8s/vtadmin/Dockerfile
```diff
@@ -47,9 +53,15 @@ RUN deluser node && \
    chown -R vitess:vitess /vt && \
    chown -R vitess:vitess /vtdataroot
 
-USER vitess
+FROM nginx:latest AS nginx
+
+COPY --from=node /vt/web/vtadmin/build /var/www/
+COPY --from=node nginx.conf /etc/nginx/nginx.conf
+COPY --from=node entrypoint.sh /entrypoint.sh
```
L3o-pold, line 60: you should remove the from arg

docker/k8s/vtadmin/nginx.conf
```diff
@@ -0,0 +1,11 @@
+events {}
+http {
+    include mime.types;
+    server {
+        listen 14201;
+        root /var/www;
```
L3o-pold, line 6: we should add headers + gzip for caching static files expect for the `config.js` file

docker/k8s/vtadmin/Dockerfile
```diff
@@ -47,9 +53,15 @@ RUN deluser node && \
...
+
```
L3o-pold, line 61: we should specify a user (vitess) to not run the container as root

docker/k8s/vtadmin/entrypoint.sh
pbibra: ahh I was just debugging, can remove

docker/k8s/vtadmin/nginx.conf
L3o-pold: something like this (not tested)

```
gzip on;
gzip_disable "msie6";

gzip_vary on;
gzip_proxied any;
gzip_comp_level 6;
gzip_buffers 16 8k;
gzip_http_version 1.1;
gzip_min_length 256;
gzip_types
  application/atom+xml
  application/geo+json
  application/javascript
  application/x-javascript
  application/json
  application/ld+json
  application/manifest+json
  application/rdf+xml
  application/rss+xml
  application/xhtml+xml
  application/xml
  font/eot
  font/otf
  font/ttf
  image/svg+xml
  text/css
  text/javascript
  text/plain
  text/xml;

location = /config/config.js {
        expires off;
    }

    location ~* \.(css|gif|jpg|js|png|ico|otf|sng|xls|doc|exe|jpeg|tgx)$ {
        expires max;
    }

    location / {
       client_max_body_size 50M;
       try_files $uri /index.html;
    }
 ```

File: docker/k8s/vtadmin/Dockerfile (hunk 7 above)
Comment (pbibra) [reply docker/k8s/vtadmin/Dockerfile:61]: Going to specify the user as nginx in this case since it's available. I'll give it the necessary permissions

File: docker/k8s/vtadmin/nginx.conf (hunk 6 above)
Comment (pbibra) [reply docker/k8s/vtadmin/nginx.conf:8]: The committed config works for caching: 
image

Comment (pbibra) [reply docker/k8s/vtadmin/nginx.conf:29]: added the gzip types and those have the correct gzip response headers: 
image

File: docker/k8s/vtadmin/Dockerfile [hunk 8]
```diff
@@ -21,7 +21,6 @@ FROM node:16-${DEBIAN_VER} as node
 
 # Set up Vitess environment (just enough to run pre-built Go binaries)
 ENV VTROOT /vt
-ENV VTADMIN_WEB_PORT=14201
```
L3o-pold, line 24: I think we still need this to configure the port used by vtadmin web (nginx in your PR)

docker/k8s/vtadmin/Dockerfile
```diff
@@ -40,16 +39,29 @@ COPY --from=k8s /vt/web/vtadmin /vt/web/vtadmin
...
+RUN touch /var/run/nginx.pid && \
+   chown -R nginx:nginx /var/run/nginx.pid
+
+USER nginx
+
+RUN ["chmod", "+x", "/entrypoint.sh"]
+
+ENTRYPOINT ["sh", "/entrypoint.sh"]
 
-VOLUME /vt/web/vtadmin/build/config
```
L3o-pold, line 66: Their is too many empty lines at the end of the file.

L3o-pold:
Since `serve` node dependencies is only used for development, you should move it from `dependencies` to `devDependencies`.

docker/k8s/vtadmin/nginx.conf
```diff
@@ -0,0 +1,43 @@
+events {}
+http {
+    include mime.types;
```
L3o-pold, line 3: add `server_tokens off; ` to remove sensitive information about the nginx version.

docker/k8s/vtadmin/Dockerfile
```diff
@@ -40,16 +39,29 @@ COPY --from=k8s /vt/web/vtadmin /vt/web/vtadmin
...
 
-# add vitess user/group and add permissions
-RUN deluser node && \
-   groupadd -r --gid 2000 vitess && \
-   useradd -r -g vitess --uid 1000 vitess && \
-   chown -R vitess:vitess /vt && \
-   chown -R vitess:vitess /vtdataroot
 
-USER vitess
+FROM nginx:1.23.1-alpine AS nginx
```
L3o-pold, line 43: I recommend you to use official nginx unprivileged image it will avoid to re-write some logic.

docker/k8s/vtadmin/Dockerfile
```diff
@@ -40,16 +39,29 @@ COPY --from=k8s /vt/web/vtadmin /vt/web/vtadmin
...
+   chown -R nginx:nginx /var/cache/nginx && \
+   chown -R nginx:nginx /var/log/nginx && \
+   chown -R nginx:nginx /usr/sbin/nginx && \
+   chown nginx:nginx /etc/nginx/nginx.conf && \
+   chown nginx:nginx /entrypoint.sh
+
+RUN touch /var/run/nginx.pid && \
+   chown -R nginx:nginx /var/run/nginx.pid
+
+USER nginx
```
L3o-pold, line 60: this can be removed if using official nginx unprivileged image.

docker/k8s/vtadmin/Dockerfile
```diff
@@ -40,16 +39,29 @@ COPY --from=k8s /vt/web/vtadmin /vt/web/vtadmin
...
-   useradd -r -g vitess --uid 1000 vitess && \
-   chown -R vitess:vitess /vt && \
-   chown -R vitess:vitess /vtdataroot
 
-USER vitess
+FROM nginx:1.23.1-alpine AS nginx
+
+COPY --from=node /vt/web/vtadmin/build /var/www/
+COPY nginx.conf /etc/nginx/nginx.conf
+COPY entrypoint.sh /entrypoint.sh
```
L3o-pold, line 47: I think having a dedicated entrypoint is not needed.

docker/k8s/vtadmin/Dockerfile
```diff
@@ -40,16 +39,29 @@ COPY --from=k8s /vt/web/vtadmin /vt/web/vtadmin
...
+   chown nginx:nginx /entrypoint.sh
+
+RUN touch /var/run/nginx.pid && \
+   chown -R nginx:nginx /var/run/nginx.pid
+
+USER nginx
+
+RUN ["chmod", "+x", "/entrypoint.sh"]
+
+ENTRYPOINT ["sh", "/entrypoint.sh"]
```
L3o-pold, line 64: I think having a dedicated entrypoint is not needed.

docker/k8s/vtadmin/nginx.conf
L3o-pold: it works well 👍

docker/k8s/vtadmin/Dockerfile
```diff
@@ -40,16 +39,29 @@ COPY --from=k8s /vt/web/vtadmin /vt/web/vtadmin
...
-   groupadd -r --gid 2000 vitess && \
-   useradd -r -g vitess --uid 1000 vitess && \
-   chown -R vitess:vitess /vt && \
-   chown -R vitess:vitess /vtdataroot
 
-USER vitess
+FROM nginx:1.23.1-alpine AS nginx
+
+COPY --from=node /vt/web/vtadmin/build /var/www/
+COPY nginx.conf /etc/nginx/nginx.conf
```
L3o-pold, line 46: you can add ([COPY](https://docs.docker.com/engine/reference/builder/#copy)) `--chown` attribute to set ownership/permission to the nginx user.

docker/k8s/vtadmin/Dockerfile
```diff
@@ -40,16 +39,12 @@ COPY --from=k8s /vt/web/vtadmin /vt/web/vtadmin
...
+FROM nginxinc/nginx-unprivileged:1.22-alpine AS nginx
 
-VOLUME /vt/web/vtadmin/build/config
+ENV VTADMIN_WEB_PORT=14201
 
-CMD ["/bin/bash", "-c", "/vt/web/vtadmin/node_modules/.bin/serve --no-clipboard -l $VTADMIN_WEB_PORT -s /vt/web/vtadmin/build"]
+COPY --from=node /vt/web/vtadmin/build /var/www/
+COPY default.conf /etc/nginx/templates/default.conf.template
 
+CMD ["nginx", "-g", "daemon off;"]
```
L3o-pold, line 50: we don't need to redefine the CMD, it's already in the base image: https://github.com/nginxinc/docker-nginx-unprivileged/blob/main/stable/alpine/Dockerfile#L150

docker/k8s/vtadmin/default.conf
```diff
@@ -0,0 +1,39 @@
...
+    }
+    location ~* \.(css|js)$ {
+        expires max;
+        add_header Cache-Control public;
+    }
+    location / {
+        client_max_body_size 50M;
+        try_files $uri /index.html;  
+    }
+}
```
ajm188, line 39: newline here please

L3o-pold:
LGTM

deepthi (commented):
Is the intent that we will continue to use `npm serve` for local development (and local example), but `nginx` for real deployments?
I'm trying to decide whether the example (which includes vtadmin) needs to change along with this.

pbibra:
> Is the intent that we will continue to use `npm serve` for local development (and local example), but `nginx` for real deployments? I'm trying to decide whether the example (which includes vtadmin) needs to change along with this.

Yeah that's the intent, since getting the local example to work with nginx is probably overkill. `serve` has now at least become a dev dependency. But happy to revisit this if you disagree!

deepthi:
> 
> Yeah that's the intent, since getting the local example to work with nginx is probably overkill. `serve` has now at least become a dev dependency. But happy to revisit this if you disagree!

I think that is fine for dev. Is it accurate to say that any future docker file changes will need to be tested by hand, similar to what you did?

deepthi:
Unit test failure has been fixed in #10839. No point trying to get it passing on UI-only PR. That would require a rebase/merge from main.

deepthi (approved):

pbibra:
> > Yeah that's the intent, since getting the local example to work with nginx is probably overkill. `serve` has now at least become a dev dependency. But happy to revisit this if you disagree!
> 
> I think that is fine for dev. Is it accurate to say that any future docker file changes will need to be tested by hand, similar to what you did?

I'm not sure what other ways we've tested in the past, but yes, these can be tested by hand using the method described in the PR description.

deepthi:
afterthought: @GuptaManan100 does this affect how we deploy vtadmin with vitess-operator?

GuptaManan100:
@deepthi I am not entirely sure what this PR changed for how we deploy vtadmin. The PR description says that we don't want to use `serve` package anymore, but I see that it is still present in the `web/vtadmin/package.json` file, just reordered.

Also, if we changed how we want to deploy vtadmin, shouldn't we have also changed the local example to reflect the same? It is still using `${web_dir}/node_modules/.bin/serve` to deploy vtadmin-web.

GuptaManan100:
@pbibra I don't quite understand how the docker image is running at all after this change. Could you look at the ☝️ message as well?

I see that you are using a nginx base image in the end, but don't copy the /vt folder, which means that `/vt/bin/vtadmin` isn't present in the docker image at all. If the vtadmin binary itself isn't present, I fail to understand how this is working.

I saw that the weekly VTOp tests fail https://buildkite.com/planetscale/vitess-operator/builds/140#01827c2d-00dc-4c34-9dce-e00837542ac0 and on investigation traced the error to this PR.

We would have caught this sooner in an earlier VTOp cron test, but we didn't probably because our dockerhub account had some issues leading to new images not being pushed for a while.

I see there is a testing note in the PR description, I tried to follow similar steps but it doesn't work for me. I would appreciate a little more detail on that front as well. 

cc - @deepthi

L3o-pold:
@GuptaManan100 files are copied here https://github.com/tinyspeck/vitess/blob/3ba396c507caca5e992cbbe148c4010db6fdafdf/docker/k8s/vtadmin/Dockerfile#L47

GuptaManan100:
@L3o-pold That isn't the vtadmin binary. It was copied here at https://github.com/tinyspeck/vitess/blob/3ba396c507caca5e992cbbe148c4010db6fdafdf/docker/k8s/vtadmin/Dockerfile#L30, but not in the second base image of Nginx.

We need to `vtadmin` binary to run the API, but I don't think that is copied.

L3o-pold:
you totally right!

L3o-pold:
Trying to fix it but vtadmin binary in latest `vitess/k8s` seems broken 😞 

```shell
docker pull vitess/k8s:latest
latest: Pulling from vitess/k8s
Digest: sha256:3b379fa3d06141f87e049d0ac53faddce39ad33bcd771c31021b7972d094549a
Status: Image is up to date for vitess/k8s:latest
docker.io/vitess/k8s:latest

docker run -it --rm vitess/k8s:latest bash
/vt/bin/vtadmin 
ERROR: logging before flag.Parse: E0822 08:26:33.752065       8 syslogger.go:149] can't connect to syslog
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x57027e]

goroutine 1 [running]:
github.com/spf13/pflag.(*FlagSet).AddGoFlag(0xc00034e500, 0x0)
	github.com/spf13/pflag@v1.0.5/golangflag.go:86 +0x1e
main.main()
	vitess.io/vitess/go/cmd/vtadmin/main.go:166 +0x206
```

GuptaManan100:
@ajm188 ☝️

ajm188:
sorrrrrrrryyyyyyyyyyyyyy https://github.com/vitessio/vitess/pull/11063

GuptaManan100:
I tried fixing the issues in the image today, but I couldn't, so I have created a reverting PR here at https://github.com/vitessio/vitess/pull/11075.

L3o-pold:
@GuptaManan100 could you try this PR https://github.com/vitessio/vitess/pull/11076 see if it works for you?

GuptaManan100:
@L3o-pold I tried making a similar change locally and testing, but the issue isn't just limited to the `vtadmin` binary not being present. Since we are using the `nginx` base image in the end, we don't have the `node` binary in the final image either. So the earlier way of deploying using `/vt/web/vtadmin/node_modules/.bin/serve` doesn't work and I couldn't figure out how to get nginx to work. Where to put the configuration file for VTAdmin, etc.

L3o-pold:
yeah removing node was one of the goal of the PR I guess. It should not be an issue and a deployment should look like this.

The config can be mounted here `/var/www/config`.

```yaml
      containers:
        - name: vtadmin
          image: vitess/vtadmin:v14.0.1
          imagePullPolicy: IfNotPresent
          command:
            - /vt/bin/vtadmin
          args:
            - --addr=:14200
            - --http-origin=*
            - --http-tablet-url-tmpl=http://{{ .Tablet.Hostname }}:15000
            - --cluster=id=d3,name=d3,discovery=staticfile,discovery-staticfile-path=/var/vtadmin/discovery.json,tablet-fqdn-tmpl={{ .Tablet.Hostname }}:15000
            - --rbac
            - --rbac-config=/var/vtadmin/rbac.yaml
            - --logtostderr
            - --alsologtostderr
          env:
            - name: GOMAXPROCS
              value: "4"
          ports:
            - containerPort: 14200
              name: tcp-api
              protocol: TCP
          securityContext:
            runAsUser: 999
          volumeMounts:
            - mountPath: /var/vtadmin
              name: vtadmin-config
              readOnly: true
        - name: web
          image: vitess/vtadmin:v14.0.1
          imagePullPolicy: IfNotPresent
          ports:
            - containerPort: 14201
              name: tcp-web
              protocol: TCP
```

GuptaManan100:
I tried it out but I get the following error - 
```
Containers:
  vtadmin-api:
    Container ID:  containerd://bd89b6be2cb72b12913daed5278aef9542fb70e20379dc25b63e3b23a6f8a127
    Image:         vitess/vtadmin:latest
    Image ID:      sha256:b3becfa3a751374f045f6ff7eba46d06106f7100e4dbb0a6479bd7b26e9d7b80
    Port:          15001/TCP
    Host Port:     0/TCP
    Command:
      /vt/bin/vtadmin
    Args:
      --addr=:15001
      --alsologtostderr=true
      --cluster-config=/vt/secrets/cluster-config/cluster-config.yaml
      --grpc-tracing=true
      --http-origin=*
      --http-tracing=true
      --logtostderr=true
      --rbac=true
      --rbac-config=/vt/secrets/rbac-config/rbac.yaml
      --tracer=opentracing-jaeger
    State:          Running
      Started:      Tue, 23 Aug 2022 15:00:15 +0530
    Ready:          True
    Restart Count:  0
    Liveness:       http-get http://:api/health delay=300s timeout=1s period=10s #success=1 #failure=30
    Readiness:      http-get http://:api/health delay=0s timeout=1s period=10s #success=1 #failure=3
    Environment:    <none>
    Mounts:
      /var/run/secrets/kubernetes.io/serviceaccount from kube-api-access-g8d9n (ro)
      /vt/secrets/cluster-config from cluster-config-secret (ro)
      /vt/secrets/discovery-config from discovery-config-secret (ro)
      /vt/secrets/rbac-config from rbac-config-secret (ro)
  vtadmin-web:
    Container ID:  containerd://3f8281fa7e434ac6485cdd28cc6eb03adc382e21e3fa9a502d9194221cf80b29
    Image:         vitess/vtadmin:latest
    Image ID:      sha256:b3becfa3a751374f045f6ff7eba46d06106f7100e4dbb0a6479bd7b26e9d7b80
    Port:          15000/TCP
    Host Port:     0/TCP
    Command:
      nginx
    Args:
      -g
      daemon off;
    State:          Running
      Started:      Tue, 23 Aug 2022 15:10:11 +0530
    Last State:     Terminated
      Reason:       Completed
      Exit Code:    0
      Started:      Tue, 23 Aug 2022 15:00:15 +0530
      Finished:     Tue, 23 Aug 2022 15:10:11 +0530
    Ready:          False
    Restart Count:  1
    Liveness:       http-get http://:web/ delay=300s timeout=1s period=10s #success=1 #failure=30
    Readiness:      http-get http://:web/ delay=0s timeout=1s period=10s #success=1 #failure=3
    Environment:
      VTADMIN_WEB_PORT:  15000
    Mounts:
      /var/run/secrets/kubernetes.io/serviceaccount from kube-api-access-g8d9n (ro)
      /var/www/config from config-js-secret (ro)
```

```
Events:
  Type     Reason       Age                   From                           Message
  ----     ------       ----                  ----                           -------
  Normal   Scheduled    14m                   default-scheduler              Successfully assigned default/example-zone1-vtadmin-c03d7eae-545c486cbd-pd2qf to kind-0-control-plane
  Warning  FailedMount  14m                   kubelet, kind-0-control-plane  MountVolume.SetUp failed for volume "rbac-config-secret" : failed to sync secret cache: timed out waiting for the condition
  Normal   Pulled       14m                   kubelet, kind-0-control-plane  Container image "vitess/vtadmin:latest" already present on machine
  Normal   Created      14m                   kubelet, kind-0-control-plane  Created container vtadmin-api
  Normal   Started      14m                   kubelet, kind-0-control-plane  Started container vtadmin-api
  Normal   Pulled       14m                   kubelet, kind-0-control-plane  Container image "vitess/vtadmin:latest" already present on machine
  Normal   Created      14m                   kubelet, kind-0-control-plane  Created container vtadmin-web
  Normal   Started      14m                   kubelet, kind-0-control-plane  Started container vtadmin-web
  Warning  Unhealthy    9m31s                 kubelet, kind-0-control-plane  Liveness probe failed: Get "http://10.244.0.11:15000/": dial tcp 10.244.0.11:15000: connect: connection refused
  Warning  Unhealthy    4m40s (x74 over 14m)  kubelet, kind-0-control-plane  Readiness probe failed: Get "http://10.244.0.11:15000/": dial tcp 10.244.0.11:15000: connect: connection refused
 ~/vitess-operator   main ±  kubectl logs example-zone1-vtadmin-c03d7eae-545c486cbd-pd2qf -c vtadmin-web 
2022/08/23 09:40:11 [notice] 1#1: using the "epoll" event method
2022/08/23 09:40:11 [notice] 1#1: nginx/1.22.0
2022/08/23 09:40:11 [notice] 1#1: built by gcc 11.2.1 20220219 (Alpine 11.2.1_git20220219) 
2022/08/23 09:40:11 [notice] 1#1: OS: Linux 5.10.76-linuxkit
2022/08/23 09:40:11 [notice] 1#1: getrlimit(RLIMIT_NOFILE): 1048576:1048576
2022/08/23 09:40:11 [notice] 1#1: start worker processes
2022/08/23 09:40:11 [notice] 1#1: start worker process 15
2022/08/23 09:40:11 [notice] 1#1: start worker process 16
2022/08/23 09:40:11 [notice] 1#1: start worker process 17
2022/08/23 09:40:11 [notice] 1#1: start worker process 18
2022/08/23 09:40:11 [notice] 1#1: start worker process 19
```

GuptaManan100:
@L3o-pold I get this output - 
```
 ✘  ~/vitess-operator   main ±  kubectl exec example-zone1-vtadmin-c03d7eae-545c486cbd-4bd44 -c vtadmin-web cat /etc/nginx/conf.d/default.conf
server {
    listen       8080;
    server_name  localhost;

    #access_log  /var/log/nginx/host.access.log  main;

    location / {
        root   /usr/share/nginx/html;
        index  index.html index.htm;
    }

    #error_page  404              /404.html;

    # redirect server error pages to the static page /50x.html
    #
    error_page   500 502 503 504  /50x.html;
    location = /50x.html {
        root   /usr/share/nginx/html;
    }

    # proxy the PHP scripts to Apache listening on 127.0.0.1:80
    #
    #location ~ \.php$ {
    #    proxy_pass   http://127.0.0.1;
    #}

    # pass the PHP scripts to FastCGI server listening on 127.0.0.1:9000
    #
    #location ~ \.php$ {
    #    root           html;
    #    fastcgi_pass   127.0.0.1:9000;
    #    fastcgi_index  index.php;
    #    fastcgi_param  SCRIPT_FILENAME  /scripts$fastcgi_script_name;
    #    include        fastcgi_params;
    #}

    # deny access to .htaccess files, if Apache's document root
    # concurs with nginx's one
    #
    #location ~ /\.ht {
    #    deny  all;
    #}
}
```

L3o-pold:
ok I think I understand. Did you specify yourself the command and args?

```yaml
Command:
      nginx
    Args:
      -g
      daemon off;
 ```

it should not be specified in the deployment or should be 


```yaml
command:
- /docker-entrypoint.sh
args:
- nginx
- g
- daemon off;
```

GuptaManan100:
@L3o-pold Let me try that, the docker image said
 ```
# command to run nginx is in the base image
 # https://github.com/nginxinc/docker-nginx-unprivileged/blob/main/stable/alpine/Dockerfile#L150
```

So it thought I could copy the command from there

**Comment by L3o-pold** [comment]
docker entrypoint == kubernetes container command
docker command == kubernetes container args

**Comment by GuptaManan100** [comment]
@L3o-pold First of all, that was the issue! It does work!
I'll close the revert PR in favor of your fix and I'll open a PR on the operator side with the changes I have to make it work.

Second, I am a little confused as to why this worked...
Reading https://codewithyury.com/docker-run-vs-cmd-vs-entrypoint/, I understand that the entrypoint part of the docker file is something that can never be overridden and cmd parts are the ones that can be specified in the docker image. 
So running the nginx docker image should still be something like `docker run ... nginx -g -daemon-off` instead of `docker run ... /docker-entrypoint.sh nginx -g ...`?

**Comment by L3o-pold** [comment]
you can overwrite an entrypoint like this `docker run --entrypoint=... ` but yeah you right you can run it like this:

```
docker run nginx:latest nginx -g -daemon-off
docker run --entrypoint=/docker-entrypoint.sh nginx:latest nginx -g -daemon-off
docker run nginx
```

GuptaManan100:
Oh i see, maybe that's what the operator does with the command and the args. Like puts the command in the entrypoint and the args in the cmd for docker, which would explain why it works now. Thank-you for the fix and the explanation! 💕

GuptaManan100:
Here is the PR for the VTop changes - https://github.com/planetscale/vitess-operator/pull/304

PR #10773, opened 2022-07-21
Add drop_foreign_keys to v2 MoveTables command

dbussink:
## Description

This change was missed in https://github.com/vitessio/vitess/pull/9904 so it doesn't work for v2 right now.

## Related Issue(s)

https://github.com/vitessio/vitess/pull/9904 which missed this change

## Checklist

This also will need to be backported to 14 then which is missing this bit as well.


rohit-nayak-ps (approved):
lgtm

deepthi:
Thank you @dbussink!

PR #10775, opened 2022-07-21
Refactor: unexplode query execution methods

dbussink:
This follows a similar path as refactors like https://github.com/vitessio/vitess/pull/10671 where we return the
protobuf objects directly instead of exploding them already in the interface.

Doing this makes it significantly easier to add things later on. One thing we're experimenting with is being able to send down GTID information at the start of transactions for things like https://github.com/vitessio/vitess/pull/10663 to get a consistent GTID.

Adding that today would mean having to do a similar change as this one to add an additional return argument, but we'd need that type of change on each such addition.

In general, the goal was to make these types of changes easier and this refactor aims to do that.

The one main thing here that's worth pointing out is that `Execute` now returns a protobuf response and not directly a `sqltypes.Result`. This means that there a few more callers that now have to make that conversion.

In total though, we don't add any roundtrips from `sqltypes.Result` to `querypb.QueryResult` except for in test cases. In non test code paths, the total number of conversions should stay the same and not have a performance impact.

## Related Issue(s)

See https://github.com/vitessio/vitess/pull/10671 for similar refactors. 

## Checklist


go/vt/vttablet/queryservice/queryservice.go
```diff
@@ -109,78 +103,16 @@ type QueryService interface {
...
-			case rs.ch <- qr:
-			}
-			return nil
-		})
-		if rs.err == nil {
-			rs.err = io.EOF
-		}
-	}()
-	return rs
-}
```
dbussink, line 186: All the code above seemed to be unused code, so I removed it.

shlomi-noach: I think this is used in `diff_utils.go`?
```
81:	stream := queryservice.ExecuteWithStreamer(ctx, conn, &querypb.Target{
```
```
115:	stream := queryservice.ExecuteWithTransactionalStreamer(ctx, conn, &querypb.Target{
```

dbussink: @shlomi-noach I don't see a `diff_utils.go` anywhere in the codebase and if I search for `ExecuteWithStreamer` it comes up with no results as well?

shlomi-noach: I was looking at an old code; `diff_utils.go` does not exist anymore.

PR #10779, opened 2022-07-21
looking into onlineddl_vrepl_stress_suite flakiness in CI

shlomi-noach:
~~work in progress~~

`[UPDATE]`

ready for review

This PR fixes `onlineddl_vrepl_stress_suite` flakiness by giving some breathing room for GitHub CI. The flakiness was seen in `mysql80` tests. The evidence was that migrations could not complete in the suite. It seems like in `mysql80` and specifically in GitHub CI the workload is too high.

Instead of reducing the workload at all times, we keep the load 90% of the time, and take 10% of the time to rest. This keeps contention high most of the time, and still lets the system recover.

We also reduce `innodb_lock_wait_timeout` and ignore deadlock errors, as these seem more frequent in `mysql80`.


shlomi-noach:
I'm happy with this solution. This keeps the load high `90%` of the time, with high contention, and then takes a rest `10%` of the time. `[(onlineddl_vrepl_stress_suite) mysql80](https://github.com/vitessio/vitess/runs/7450983849?check_suite_focus=true#logs)` passes in about `15min`.

shlomi-noach:
Two other things in this PR:

- we reduce `innodb_lock_wait_timeout` because there's no point waiting on a lock
- we ignore deadlock errors

go/test/endtoend/onlineddl/vrepl_stress_suite/onlineddl_vrepl_stress_suite_test.go
```diff
@@ -697,9 +703,27 @@ func runSingleConnection(ctx context.Context, t *testing.T, autoIncInsert bool,
...
+				case mysql.ERLockDeadlock:
+					// That's fine. We create a lot of contention; some transactions will deadlock and
+					// rollback. It happens, and we can ignore those and keep on going.
+					err = nil
+				}
+			}
 		}
 		assert.Nil(t, err)
 		time.Sleep(singleConnectionSleepInterval)
+		// Most o fthe time, we want the load to be high, so as to create real stress and potentially
```
deepthi, line 717:
```suggestion
		// Most of the time, we want the load to be high, so as to create real stress and potentially
```

deepthi (approved):

go/test/endtoend/onlineddl/vrepl_stress_suite/onlineddl_vrepl_stress_suite_test.go
shlomi-noach: Whoops. Merged without noticing this comment.

PR #10781, opened 2022-07-21
Prefix column names with tables names in Gen4

frouioui:
## Description

This Pull Request changes the gen4 planner to produce plans that prefix column names with the table name they belong to.

This makes reading the queries in the plan output easier, and supports other parts of the code base that does not have the semantic analysis available, such as the vttablet.

## Checklist


systay:
This is bad. Don't do it.

PR #10784, opened 2022-07-22
refactor: make resource pool as interface and pool refresh as common

harshit-gangal:
## Description

This PR refactors the resource_pool and creates an interface that will be used.
It will also moved refresh pool code out of the resource_pool and taken in the ResourcePool interface as input to work on the watcher.

This is a prep work for adding a new pool implementation.

## Checklist


go/pools/refresh_pool.go
```diff
@@ -0,0 +1,97 @@
...
+		refreshInterval time.Duration
+		refreshTicker   *time.Ticker
+		refreshStop     chan struct{}
+		refreshWg       sync.WaitGroup
+
+		pool IRefreshPool
+	}
+)
+
+type IRefreshPool interface {
```
systay, line 44: I personally don't mind it, but I think the `IName`  is mostly a Java/C# thing, not very idiomatic golang.

go/pools/rp_bench_test.go
```diff
@@ -0,0 +1,35 @@
+package pools
```
systay, line 1: license header missing

systay (approved):

PR #10786, opened 2022-07-22
BugFix: VTOrc should repair replication if either replication thread is stopped

### LINKED ISSUE #10785: Bug Report: VTOrc doesn't fix replication when only 1 thread is stopped
### Overview of the Issue

VTOrc doesn't fix replication on a tablet if one of the threads is stopped and only 1 is running.

We could end up in this state because of a failure to write configuration like `Error writing master configuration`.

### Reproduction Steps

1. Start a cluster with or without semi-sync. Doesn't matter much other than the logs produced.
2. Stop replication of only the IO thread on all the replicas by issuing `STOP SLAVE IO_THREAD` on them all.
3. Wait for replication to fix, but it doesn't happen.

### Binary Version

```sh
main
```

```sh
main
```

### Log Fragments

Logs in VTOrc if semi-sync is setup are like the following - 

```
2022-07-22 13:12:40     INFO topology_recovery.go:1381] executeCheckAndRecoverFunction: proceeding with LockedSemiSyncPrimary recovery on localhost:30109; isRecoverable?: true; skipProcesses: false
I0722 13:12:40.347405   21093 locks.go:415] Unlocking shard ks/0 for successful action Orc Recovery
2022-07-22 13:12:41     INFO topology_recovery.go:1316] executeCheckAndRecoverFunction: proceeding with LockedSemiSyncPrimary detection on localhost:30109; isActionable?: true; skipProcesses: false
I0722 13:12:41.163302   21093 locks.go:375] Locking shard ks/0 for action Orc Recovery
```

So we do detect that the primary is locked on semi-sync ACKs but we don't do anything about it.

GuptaManan100:
## Description

This PR fixes the issue described in https://github.com/vitessio/vitess/issues/10785. 
VTOrc should be repairing replication if either of the threads is stopped and not only when both the threads are stopped. 
In Vitess we only stop 1 thread in the following cases - 
1. `StopReplicationAndGetStatus` RPC of Tablet Manager Client. This RPC is called only from `EmergencyReparentShard` which is run after acquiring a shard lock. By the end, we fix replication on all the tablets, so VTOrc will not interfere in this case.
2. `catchupToGTID` function used in restore. This happens after we change the tablet type to the `RESTORE` type. VTOrc does not fix replication on tablets of this type, so we shouldn't conflict here either.

## Related Issue(s)

- Fixes #10785 

## Checklist

### DIFF
--- a/go/test/endtoend/vtorc/general/vtorc_test.go
+++ b/go/test/endtoend/vtorc/general/vtorc_test.go
@@ -155,10 +155,6 @@ func TestStopReplication(t *testing.T) {
 	curPrimary := utils.ShardPrimaryTablet(t, clusterInfo, keyspace, shard0)
 	assert.NotNil(t, curPrimary, "should have elected a primary")
 
-	// TODO(deepthi): we should not need to do this, the DB should be created automatically
-	_, err := curPrimary.VttabletProcess.QueryTablet(fmt.Sprintf("create database IF NOT EXISTS vt_%s", keyspace.Name), keyspace.Name, false)
-	require.NoError(t, err)
-
 	var replica *cluster.Vttablet
 	for _, tablet := range shard0.Vttablets {
 		// we know we have only two tablets, so the "other" one must be the new primary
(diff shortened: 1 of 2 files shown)


mattlord (approved):
LGTM! Nice work, @GuptaManan100 !

deepthi (approved):

Issue #10692, opened 2022-07-13
Feature Request: consistent MySQL-based API to discover failed/skipped shards in `MULTI` transactional mode


maxenglander:
### Feature Description

When using `MULTI` transactional mode, and when issuing `UPDATE` or `INSERT` queries that touch one or more shards, the overall query may partially succeed so that the query is successfully applied to some shards, but not others. That is good and working as expected.

However, it is not currently possible, but would be very useful, to determine which shards failed, so that the client can optionally make a decision to log or correct the partial failures. This ability appears to be what the comment directive [`SCATTER_ERRORS_AS_WARNINGS`](https://vitess.io/docs/13.0/user-guides/configuration-advanced/comment-directives/#scatter-errors-as-warnings-scatter_errors_as_warnings) was intended to enable.

Some current challenges to this goal:

 * The error messages returned by Vitess pre-commit and at-commit are not consistent. Pre-commit errors have the format `target: <shard>: <message>`, whereas at-commit errors have `<message>`. I've made a [PR attempting to make these consistent](https://github.com/vitessio/vitess/pull/10669).
 * `SCATTER_ERRORS_AS_WARNINGS` is inconsistently applied. It appears to currently work for `SELECT` ~and `UPDATE`~, but not for DML. I've put together a [PR attempting to enable `SCATTER_ERRORS_AS_WARNINGS` for `INSERT`](https://github.com/vitessio/vitess/pull/10681).
 * `SCATTER_ERRORS_AS_WARNINGS` returns an error if all shards involved in a query failed, but records warnings if some shards failed while other succeeded. 
    * This isn't a huge deal if the client knows to check both the error and the warnings area for failures, although it will be a bit of a nuisance for them to have to check both the MySQL error and warnings, and to implement separate parsing logic for those two paths.
    * The error message can quickly get long when there are multiple failed shards involved. If the client has a maximum error message length ([MySQL C library has a 512 char limit](https://github.com/mysql/mysql-server/blob/8d8c986e5716e38cb776b627a8eee9e92241b4ce/include/mysql_com.h#L877)), then it won't be able to find out all the shard names.
 * `SCATTER_ERRORS_AS_WARNINGS` works outside of transactions and in-transaction before `COMMIT`, but is not applied when `COMMIT` is issued. In `MULTI` transactional mode, `COMMIT` is issued to shards in-order, and the sequence is aborted at the first failure. The error does not currently report which shard failed, nor which shards were "skipped".

[...]

PR #10669, opened 2022-07-11
Inject shard name in commit-phase multi-shard errors

maxenglander:
## Description

In MULTI transactional mode, errors can occur at different phases:

 1. When the transaction is started (e.g. with `BEGIN`).
 2. Inside the transaction.
 3. When the transaction is committed (e.g. with `COMMIT`).

Error messages generated within-transaction include the shard name,
e.g.:

> ERROR 1317 (70100): target: keyspace2.c0-.primary: vttablet: ...

However, messages generated in the commit phase do not:

> ERROR 1317 (70100): vttablet: rpc error: code = Canceled desc...

This commit injects shard name into commit-phase errors.

## Related Issue(s)

#10692 

## Checklist

-   [x] "Backport me!" label has been added if this change should be backported
-   [x] Tests were added or are not required
-   [ ] Documentation was added or is not required

PR #10790, opened 2022-07-22
Backporting to 12: Inject shard name in commit-phase multi-shard errors

maxenglander:
## Description

This is a backport of https://github.com/vitessio/vitess/pull/10669 into the release-12.0 branch.

## Related Issue(s)

 * Original PR: #10669 
 * Issue: #10692 

## Checklist

### DIFF
--- a/go/vt/vtgate/tabletgateway.go
+++ b/go/vt/vtgate/tabletgateway.go
@@ -107,7 +107,7 @@ func NewTabletGateway(ctx context.Context, hc discovery.HealthCheck, serv srvtop
 		statusAggregators: make(map[string]*TabletStatusAggregator),
 	}
 	gw.setupBuffering(ctx)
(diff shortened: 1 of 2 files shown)


dbussink:
@maxenglander What's the reason for specifically only backporting this to 12? In general, this looks like a new feature so not sure we would backport it. But even in the case we do, it must be consistent across all released versions and it needs to go into 13 & 14 as well.

maxenglander:
Hey @dbussink there are PlanetScale customers on 12 that want to know which shards failed in a scatter or during a failure that occurs during commit in a MULTI-mode transaction.

I'll open backports to 13 and 14 as well.

harshit-gangal:
This is an enhancement, therefore it does qualify for a backport. 
Closing this PR.

Issue #10789, opened 2022-07-22
Bug Report: Incorrect results from Filter on >= N against int32 column


wcurrie:
### Overview of the Issue

For the following query plan, the Filter on `team_fact.team \u003e= :vtg2` produces incorrect results when `team_fact.team` is declared as `int` (int32) to mysql instead of `bigint` (int64)
```json
{
  "Key": "select team.id from team_fact join team on team.id = team_fact.team left join team_member on team_member.team = team.id where team_fact.fact = :vtg1 and team_member.`user` is null and team_fact.team \u003e= :vtg2",
  "Value": {
   "QueryType": "SELECT",
   "Original": "select team.id from team_fact join team on team.id = team_fact.team left join team_member on team_member.team = team.id where team_fact.fact = :vtg1 and team_member.`user` is null and team_fact.team \u003e= :vtg2",
   "Instructions": {
    "OperatorType": "SimpleProjection",
    "Columns": [
     2
    ],
    "Inputs": [
     {
      "OperatorType": "Filter",
      "Predicate": "team_member.`user` is null and team_fact.team \u003e= :vtg2",
      "Inputs": [
       {
        "OperatorType": "Join",
        "Variant": "LeftJoin",
        "JoinColumnIndexes": "R:0,L:1,L:0",
        "JoinVars": {
         "team_id": 0
        },
        "TableName": "team_fact_team_team_member",
        "Inputs": [
         {
          "OperatorType": "Join",
          "Variant": "Join",
          "JoinColumnIndexes": "R:0,L:0",
          "JoinVars": {
           "team_fact_team": 0
          },
          "TableName": "team_fact_team",
          "Inputs": [
           {
            "OperatorType": "Route",
            "Variant": "Scatter",
            "Keyspace": {
             "Name": "customer",
             "Sharded": true
            },
            "FieldQuery": "select team_fact.team from team_fact where 1 != 1",
            "Query": "select team_fact.team from team_fact where team_fact.fact = :vtg1",
            "Table": "team_fact"
           },
           {
            "OperatorType": "Route",
            "Variant": "EqualUnique",
            "Keyspace": {
             "Name": "customer",
             "Sharded": true
            },
            "FieldQuery": "select team.id from team where 1 != 1",
            "Query": "select team.id from team where team.id = :team_fact_team",
            "Table": "team",
            "Values": [
             ":team_fact_team"
            ],
            "Vindex": "hash"
           }
          ]
         },
         {
          "OperatorType": "Route",
[...]

PR #10793, opened 2022-07-23
fix: evalengine - check compare numeric of same type

harshit-gangal:
## Description

This PR fails the numeric compare if the type are not same after the type conversion.
This is to prevent from accidentally returning wrong result.

The follow up PR should handle the type conversion for all the missing numeric types.

## Related Issue(s)

- Fixes https://github.com/vitessio/vitess/issues/10789

## Checklist


frouioui (approved):
LGTM. Let's add an E2E test as well!

PR #10797, opened 2022-07-24
Online DDL vrepl tests: fix flakiness

shlomi-noach:
Investigating and fixing `onlineddl_vrepl (mysql80)` flakiness


shlomi-noach:
This is good to go as an immediate remediation for test flakiness. There possibly be some work done on better reloading schema on `vstreamer/rowstreamer` side. But until then this is a legit patch.

shlomi-noach:
Excerpt from an internal discussion:

---

I have a good understanding of the `onlineddl_vrepl (mysql80)` test failures now. The culprit is a missing `ReloadSchema` on replica tablets in the test cluster.
The thing is, in this test suite we create a table, modify it, modify again, drop it, recreate it, modify again, etc. etc. etc.
The single most recurring error in the flaky tests is during a "normal" `vitess` online ddl operation; the migration fails because some column cannot be found. But the migration is just `alter table ... engine=innodb`, so why would this error even happen?
It just so happens that previously the table was migrated via `vreplication` when that column did not exist. Then dropped, then recreated _with_ that column, then our migration came, and when attempting to vstream, it was using a cached schema, where the column did not exist.
Fixing this in the tests is simple enough, and the tests now reliably pass: https://github.com/vitessio/vitess/pull/10797/commits/0d163dec387620b927f81ba7c29c7229ef4648a0

But there is a fundamental issue here; `ReloadSchema` just doesn't run on replica tables after an Online DDL operation. `ApplySchema` may _attempt_ it, but since the ddl is async, it may not run in the correct time. The online DDL executor runs a `ReloadSchema` at the end of a migration -- but that happens on the primary tablet.

So the real solution is to find a way to `ReloadSchema` on a replica tablet, either immediately after a migration, or immediately before a new migration (and as much as possible, avoid doing this repeatedly in new gRPCs).

shlomi-noach:
Gonna make yet another experiment in a separate PR, using schema versioning. Do not merge yet.

shlomi-noach:
https://github.com/vitessio/vitess/pull/10804 fixes the flakiness more authoritatively.

Issue #10798, opened 2022-07-24
Add Create action argument to wait for the VDiff to complete before returning, showing a progress meter of some kind


mattlord:
As a human or when integrating vtctl client commands in your automation, you may want to run the VDiff and then wait until it completes before the client exits. This way you can see the progress as you wait and know immediately when it's completes or exits with an error.

PR #10799, opened 2022-07-24
VDiff2: Add --wait flag to Create/Resume actions

mattlord:
## Description

This adds the `--wait` flag for the [`Create`](https://vitess.io/docs/15.0/reference/vreplication/vdiff2/#start-a-new-vdiff) and [`Resume`](https://vitess.io/docs/15.0/reference/vreplication/vdiff2/#resume-a-previous-vdiff) actions so that you can see the progress as it runs — current status checked and displayed every `--wait-update-interval` which is 1 minute by default — and be informed  when it successfully completes or exits with an error.

## Related Issue(s)
  - Fixes: https://github.com/vitessio/vitess/issues/10798

## Checklist

-   [x] Documentation: https://github.com/vitessio/website/pull/1110

### DIFF
--- a/go/test/endtoend/vreplication/vdiff2_test.go
+++ b/go/test/endtoend/vreplication/vdiff2_test.go
@@ -319,10 +319,9 @@ func testAutoRetryError(t *testing.T, tc *testCase, cells string) {
 
 func testCLICreateWait(t *testing.T, ksWorkflow string, cells string) {
 	t.Run("vtctl create and wait", func(t *testing.T) {
-		finished := make(chan bool)
-
+		chCompleted := make(chan bool)
 		go func() {
-			_, output := performVDiff2Action(t, ksWorkflow, cells, "create", "", false, "--wait")
+			_, output := performVDiff2Action(t, ksWorkflow, cells, "create", "", false, "--wait", "--wait-update-interval=5s")
 			completed := false
 			// We don't try to parse the JSON output as it may contain a series of outputs
 			// that together do not form a valid JSON document. We can change this in the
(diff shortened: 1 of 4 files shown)


rohit-nayak-ps (commented):
Looks good.

I was wondering if we should use a single flag combining `--wait` and `--wait-update-interval`, say `wait_duration` which takes the duration to wait with default 0 (no wait)? Trying to keep the number of flags to a minimum since we have seen how that goes with `vttablet`!

mattlord:
> Looks good.
> 
> I was wondering if we should use a single flag combining `--wait` and `--wait-update-interval`, say `wait_duration` which takes the duration to wait with default 0 (no wait)? Trying to keep the number of flags to a minimum since we have seen how that goes with `vttablet`!

The issue with flags is usability and overall UX. You should have good defaults so that most users don't need to try and understand a bunch of flags, try to measure the impact of them, etc. Choices can be debilitating, especially when you don't understand them — when and why you'd use them, what you might set them to — and can't easily measure their impacts.
1. We don't have a lot of VDiff flags today: https://vitess.io/docs/15.0/reference/vreplication/vdiff2/#start-a-new-vdiff
2. The defaults are good, so most users will not even be aware of them
3. If they are looking for different behavior, the help output is pretty concise
4. With client CLI flags it's quick and easy to see the impact and determine if that's what you want or not -- this is empowering rather than debilitating 
5. Compare that with server flags like vttablet: https://vitess.io/docs/15.0/reference/programs/vttablet/
    - The list is a little overwhelming and for many options it's not clear what signals would indicate one should consider using the flag, what values they should use when, and how they'd measure the impact to determine if things are better or worse.

Again, the underlying issue is usability and UX. I think that trying to combine these two clear flags into one hurts UX as there are two aspects involved:
  1. I want to wait for the command to finish
  2. I want status/progress updates (with an ETA) every <this often> as things go along (the client is polling the status and displaying the results) — I as a developer cannot determine an objectively good value for this as you may want this to be very low for things that will finish in seconds, but longer for runs that finishes in hours or even days

rohit-nayak-ps (approved):
lgtm

PR #10802, opened 2022-07-25
Unexplode return values for queryservice

dbussink:
This adds a small wrapper struct instead of separate return arguments to make it easier to refactor and change this in the future.

This is a much simpler alternative to https://github.com/vitessio/vitess/pull/10775 which ended up being much more invasive with all the conversion changes etc.

The changes here are much more targeted and only specifically for transaction / reservation related return values so we can simplify those explicitly. 

## Related Issue(s)

See https://github.com/vitessio/vitess/pull/10671 for similar refactors.

## Checklist

### DIFF
--- a/go/vt/vttablet/queryservice/queryservice.go
+++ b/go/vt/vttablet/queryservice/queryservice.go
@@ -19,8 +19,6 @@ limitations under the License.
 package queryservice
 
 import (
-	"io"
-
 	topodatapb "vitess.io/vitess/go/vt/proto/topodata"
 
 	"context"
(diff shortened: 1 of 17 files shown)


frouioui (approved):
The refactor looks good to me

go/vt/vttablet/queryservice/queryservice.go
```diff
@@ -109,78 +107,32 @@ type QueryService interface {
...
-				return io.EOF
-			case rs.ch <- qr:
-			}
-			return nil
-		})
-		if rs.err == nil {
-			rs.err = io.EOF
-		}
-	}()
-	return rs
```
frouioui, line 185: Why are we removing these?

dbussink: @frouioui This code looked to be entirely unused, so I cleaned it up while I was in here :smile:.

frouioui: Got it!

PR #10826, opened 2022-07-25
Add semgrep CI workflow

mattlord:
## Description

This adds CI integration with [Semgrep](https://semgrep.dev). This is a trial run of the integration — to back out we would revert this PR and remove the `SEMGREP_APP_TOKEN` GitHub secret that it uses. The hope is that this helps us find and fix more issues before merging, such as goroutine leaks. It will utilize a set of rules in the workflow run which are configured in the semgrep.dev dashboard (I will invite more people if/as needed on that side — log in using GitHub OAuth). 

## Checklist

### DIFF
--- a/.github/workflows/semgrep.yml
+++ b/.github/workflows/semgrep.yml
@@ -1,0 +1,21 @@
+on:
+   pull_request: {}
+   push:
+     branches:
+     - main
+     paths:
+     - .github/workflows/semgrep.yml
+   schedule:
+   - cron: '0 0 * * 0'
+ name: Semgrep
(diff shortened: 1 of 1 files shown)


deepthi (approved):
The check is passing, does that mean it found no problems?

mattlord:
> The check is passing, does that mean it found no problems?

The CI check should only be dealing with changes made in the PR (or a least the files changed). You can see full audits of the repo in the semgrep.dev dashboard. I will figure out the best way to make that available to other maintainers next.

PR #10788, opened 2022-07-22
WIP vcopier parallel bulk insert proof-of-concept

maxenglander:
Don't look at me. Not ready for review. I'm aware of other efforts (#7728 #8934)  to parallelize, did not discover those until after I did this work..

PR #10828, opened 2022-07-25
Do Not Merge: VReplication Copy Phase: Parallelize bulk insert to check for performance improvements

rohit-nayak-ps:
## POC

Batches multiple bulk inserts in the copy phase to test if concurrent bulk inserts improve copy phase performance.

## Description
During the copy phase we stream rows and for every `PacketSize` of rows (default 250K bytes) we do a bulk insert. This PR batches multiple of these sets. 

We add a `VReplicationTableCopyTimings` stat to monitor when a table copy starts and when it ends for benchmarking purposes. This metric is per vreplication stream. 

This feature is behind an experimental vttablet flag. To enable it use `-vreplication_experimental_flags 2` on the target.

You can specify the number of bulk inserts using `-vreplication_parallel_bulk_inserts 16` Default is 4

## Approach
Instead of inserting one batch at a time we collect N batches and insert them in parallel. Commits are done in order.

## Checklist

## Impacted Areas in Vitess

- [ ]  Query Serving
- [X]  VReplication
- [ ]  Cluster Management
- [ ]  Build/CI
- [ ]  VTAdmin


maxenglander:
With @mattlord help I have been benchmarking the performance of this PR alongside #10788. Both PRs improve performance between 10-200%, depending on factors like CPU load on the data source.

Because this PR came first, already has undergone some code review, and is simpler than #10788, I'm going to add more work to this PR.

 [ ] Completely avoid async path when experimental flag not present
 [ ] Get E2E tests to pass
 [ ] Get unit tests to pass
 [ ] Add more unit tests

go/vt/vttablet/tabletmanager/vreplication/engine.go
```diff
@@ -333,8 +333,10 @@ func (vre *Engine) Exec(query string) (*sqltypes.Result, error) {
 // Exec executes the query and the related actions.
 // Example insert statement:
 // insert into _vt.vreplication
+//
```
maxenglander, line 336: gofmt changes

go/vt/vttablet/tabletmanager/vreplication/shard_sorter.go
```diff
@@ -18,25 +18,25 @@ package vreplication
```
maxenglander, line 18: gofmt changes

go/vt/vttablet/tabletmanager/vreplication/vplayer.go
```diff
@@ -72,13 +72,18 @@ type vplayer struct {
```
maxenglander, line 72: gofmt changes

go/vt/vttablet/tabletmanager/vreplication/vreplicator.go
```diff
@@ -127,20 +127,21 @@ type vreplicator struct {
 // The Filter can be empty: get all rows and columns.
 // The Filter can be a keyrange, like "-80": get all rows that are within the keyrange.
 // The Filter can be a select expression. Examples.
-//   "select * from t", same as an empty Filter,
```
maxenglander, line 130: gofmt changes

go/vt/vttablet/tabletmanager/vreplication/engine.go
maxenglander, line 336: gofmt

go/vt/vttablet/tabletmanager/vreplication/vreplicator.go
```diff
@@ -502,12 +503,12 @@ func (vr *vreplicator) setSQLMode(ctx context.Context) (func(), error) {
 
 // throttlerAppName returns the app name to be used by throttlerClient for this particular workflow
 // example results:
-// - "vreplication" for most flows
```
maxenglander, line 505: gofmt changes

go/vt/binlog/binlogplayer/binlog_player.go
```diff
@@ -471,12 +475,12 @@ func (blp *BinlogPlayer) exec(sql string) (*sqltypes.Result, error) {
 // writeRecoveryPosition writes the current GTID as the recovery position
 // for the next transaction.
 // It also tries to get the timestamp for the transaction. Two cases:
-// - we have statements, and they start with a SET TIMESTAMP that we
```
maxenglander, line 474: gofmt

go/vt/binlog/binlogplayer/binlog_player.go
```diff
@@ -573,11 +577,13 @@ var AlterVReplicationTable = []string{
 }
 
 // WithDDLInitialQueries contains the queries that:
-// - are to be expected by the mock db client during tests, or
```
maxenglander, line 576: gofmt

go/vt/vttablet/tabletmanager/vreplication/flags.go
```diff
@@ -0,0 +1,69 @@
+package vreplication
```
maxenglander, line 1: Most of this is moved out of `vreplicator.go`. See comments below identifying the parts that are new.

go/vt/vttablet/tabletmanager/vreplication/flags.go
```diff
@@ -0,0 +1,69 @@
...
+	// 		you have too many streams the extra write qps or cpu load due to these updates are unacceptable
+	//		you have too many streams and/or a large source field (lot of participating tables) which generates unacceptable increase in your binlog size
+	vreplicationHeartbeatUpdateInterval = flag.Int("vreplication_heartbeat_update_interval", 1, "Frequency (in seconds, default 1, max 60) at which the time_updated column of a vreplication stream when idling")
+	// vreplicationMinimumHeartbeatUpdateInterval overrides vreplicationHeartbeatUpdateInterval if the latter is higher than this
+	// to ensure that it satisfies liveness criteria implicitly expected by internal processes like Online DDL
+	vreplicationMinimumHeartbeatUpdateInterval = 60
+
+	vreplicationExperimentalFlags                        = flag.Int64("vreplication_experimental_flags", 0x01, "(Bitmask) of experimental features in vreplication to enable")
+	vreplicationParallelBulkInserts                      = flag.String("vreplication_parallel_bulk_inserts", "auto", "Number of parallel insertion workers to use during copy phase. Set to a number >= 1 or to \"auto\". \"auto\" uses GOMAXPROCS or 4, whichever is smaller.")
```
maxenglander, line 37: New

go/vt/vttablet/tabletmanager/vreplication/flags.go
```diff
@@ -0,0 +1,69 @@
...
+	vreplicationExperimentalFlagOptimizeInserts    int64 = 1
+	vreplicationExperimentalParallelizeBulkInserts int64 = 0x02
```
maxenglander, line 39: New

go/vt/vttablet/tabletmanager/vreplication/flags.go
```diff
@@ -0,0 +1,69 @@
...
+	vreplicationStoreCompressedGTID                      = flag.Bool("vreplication_store_compressed_gtid", false, "Store compressed gtids in the pos column of _vt.vreplication")
+)
+
+func getCopyInsertConcurrency() int {
```
maxenglander, line 43: All below is new

go/vt/vttablet/tabletmanager/vreplication/vreplicator.go
```diff
@@ -44,36 +43,6 @@ import (
 	tabletmanagerdatapb "vitess.io/vitess/go/vt/proto/tabletmanagerdata"
 )
 
-var (
```
maxenglander, line 47: Move to `flags.go`

go/vt/vttablet/tabletmanager/vreplication/vcopier.go
```diff
@@ -364,3 +437,268 @@ func (vc *vcopier) fastForward(ctx context.Context, copyState map[string]*sqltyp
...
+		return vbc.vdbClient.ExecuteWithRetry(ctx, sql)
+	})
+}
+
+func (vbc *vcopierBasicCopier) updateCopyState(ctx context.Context, lastpk *querypb.Row) error {
+	if !vbc.IsOpen() {
+		return fmt.Errorf("failed to update copy state with vcopioer basic copier: not open")
+	}
+
+	vbc.BeforeUpdateCopyState().Fire()
```
rohit-nayak-ps, line 577: We should wrap this in a select so as to handle context cancellations. Otherwise we will end up with orphan goroutines if the previous batches don't complete due to invalid contexts.

go/vt/vttablet/tabletmanager/vreplication/vcopier.go
```diff
@@ -364,3 +437,268 @@ func (vc *vcopier) fastForward(ctx context.Context, copyState map[string]*sqltyp
...
+	if err = vbc.vdbClient.Commit(); err != nil {
+		err = fmt.Errorf("error commiting copy-update-state transaction: %s", err.Error())
+		goto DONE
+	}
+
+	vbc.lastCommittedPk = lastpk
+
+DONE:
+	// Update stats.
+	vbc.stats.CopyRowCount.Add(int64(len(rows)))
```
maxenglander, line 510: Don't update stats when err != nil

go/vt/vttablet/tabletmanager/vreplication/framework_test.go
```diff
@@ -71,10 +72,10 @@ type LogExpectation struct {
 var heartbeatRe *regexp.Regexp
 
 // setFlag() sets a flag for a test in a non-racy way:
-//	* it registers the flag using a different flagset scope
-//	* clears other flags by passing a dummy os.Args() while parsing this flagset
-//	* sets the specific flag, if it has not already been defined
-//	* resets the os.Args() so that the remaining flagsets can be parsed correctly
+//   - it registers the flag using a different flagset scope
```
maxenglander, line 75: gofmt

go/vt/vttablet/tabletmanager/vreplication/vcopier.go
```diff
@@ -264,86 +427,167 @@ func (vc *vcopier) copyTable(ctx context.Context, tableName string, copyState ma
...
+		currT := newVCopierCopyTask(taskID, newVCopierCopyTaskArgs(rows.Rows, rows.Lastpk))
+		taskID++
+
+		// Sequence the previous task with the current task so that:
+		// * The previous task is completed before we begin updating
+		//   _vt.copy_state for the current task.
+		// * If the previous task fails or is canceled, the current task is
+		//   canceled.
+		if prevT != nil {
+			prevCh := make(chan *vcopierCopyTaskResult, 1)
```
maxenglander, line 463: I'm not explicitly closing this channel. I think golang will clean this one up automatically when these assignments to `prevT` and `currT` are garbage collected?

If not, I can explicitly cleanup this channel when `currT` is done.

go/vt/vttablet/tabletmanager/vreplication/vcopier.go
```diff
@@ -264,86 +427,167 @@ func (vc *vcopier) copyTable(ctx context.Context, tableName string, copyState ma
...
+			prevT.Lifecycle().OnResult().SendTo(prevCh)
+			currT.Lifecycle().Before(vcopierCopyTaskUpdateCopyState).AwaitCompletion(prevCh)
```
maxenglander, line 465: One outcome of this sequencing is that if any task T fails, every subsequent task T+1, ..., T+N will also fail. I think this behavior makes sense. 

As soon as T fails, we want to return the error to VStreamRows as soon as possible. In async execution, it's possible a failure in T won't be immediately available to us, and while we're waiting for the result of T we may enqueue T+1, ..., T+N for execution. The sequencing behavior will guarantee that T1, ..., T+N won't succeed if T1 fails.

go/vt/vttablet/tabletmanager/vreplication/vcopier.go
```diff
@@ -364,3 +631,562 @@ func (vc *vcopier) fastForward(ctx context.Context, copyState map[string]*sqltyp
...
+	}
+
+	poolCapacity := int(math.Max(float64(vcq.maxDepth), 1))
+	vcq.workerPool = pools.NewResourcePool(
+		/* factory */
+		func(ctx context.Context) (pools.Resource, error) {
+			worker, err := vcq.workerFactory(ctx)
+			if err != nil {
+				return nil, fmt.Errorf(
+					"failed to create copier for resource pool: %s",
```
mattlord, line 750: Nit, but I think we should say `failed to create vcopier worker`.

go/vt/vttablet/tabletmanager/vreplication/vcopier.go
```diff
@@ -364,3 +631,562 @@ func (vc *vcopier) fastForward(ctx context.Context, copyState map[string]*sqltyp
...
+		0,            /* idle timeout */
+		nil,          /* log wait */
+		nil,          /* refresh check */
+		0,            /* refresh interval */
+	)
+
+	vcq.isOpen = true
+}
+
+func (vct *vcopierCopyTask) Args() *vcopierCopyTaskArgs {
```
mattlord, line 768: Do you think we'll ever need/want these to be accessible outside of the package? It feels like implementation details to me, but maybe not? I actually see the exported functions used (e.g. `vct.IsOpen()`) within the package, are there any reasons for that? Maybe something I'm not aware of.

go/vt/vttablet/tabletmanager/vreplication/vcopier.go
```diff
@@ -364,3 +631,562 @@ func (vc *vcopier) fastForward(ctx context.Context, copyState map[string]*sqltyp
...
+}
+
+// Open the work queue. The provided arguments are used to generate INSERT and
+// UPDATE statements.
+func (vcq *vcopierCopyWorkQueue) Open(
+	copyStateUpdate *sqlparser.ParsedQuery,
+	pkfields []*querypb.Field,
+	tablePlan *TablePlan,
+) {
+	if vcq.IsOpen() {
```
mattlord, line 739: Any reason to use the public/exported functions within the package?

go/vt/vttablet/tabletmanager/vreplication/vcopier.go
```diff
@@ -364,3 +631,562 @@ func (vc *vcopier) fastForward(ctx context.Context, copyState map[string]*sqltyp
...
+//	}
+func (vtl *vcopierCopyTaskLifecycle) TryAdvance(
+	ctx context.Context,
+	args *vcopierCopyTaskArgs,
+	nextState vcopierCopyTaskState,
+	fn func(context.Context, *vcopierCopyTaskArgs) error,
+) (newState vcopierCopyTaskState, err error) {
+	// Canceling the task when there is a context error seems like a sensible
+	// thing to do. E.g. when copyPhaseDuration elapses it will cause the task
+	// to be canceled. Hower this is more aggressive cancelation behavior than
```
mattlord, line 845: Nitty nit, but however and cancellation

go/vt/vttablet/tabletmanager/vreplication/vcopier.go
```diff
@@ -364,3 +631,562 @@ func (vc *vcopier) fastForward(ctx context.Context, copyState map[string]*sqltyp
...
+	//defer func() {
+	//	if err != nil {
+	//		//if errors.Is(err, context.Canceled) ||
+	//		//	errors.Is(err, context.DeadlineExceeded) {
+	//		//	newState = vcopierCopyTaskCancel
+	//		//} else {
+	//		//	newState = vcopierCopyTaskFail
+	//		//}
+	//	}
+	//}()
```
mattlord, line 858: Is not enabling this going to cause potential PRS/ERS failures? We need to be able to cancel and close things fairly quickly in order to prevent PRS failures.

go/vt/vttablet/tabletmanager/vreplication/vcopier.go
```diff
@@ -364,3 +631,562 @@ func (vc *vcopier) fastForward(ctx context.Context, copyState map[string]*sqltyp
...
+//	task.Lifecycle().OnResult().SendTo(resultCh)
+//	defer func() {
+//	  result := <-resultCh
+//	}()
+func (vrh *vcopierCopyTaskResultHooks) SendTo(ch chan<- *vcopierCopyTaskResult) {
+	vrh.Do(func(ctx context.Context, result *vcopierCopyTaskResult) {
+		select {
+		case ch <- result:
+		case <-ctx.Done():
+			// Log a warning?
```
mattlord, line 911: What does this signify? We could block indefinitely on the channel receive, right? We failed to perform the requested task of sending the result to the channel, so shouldn't we return an error? We may even want to back all the way out, or no? What's the worst possible consequence of this?

go/vt/vttablet/tabletmanager/vreplication/vcopier.go
```diff
@@ -364,3 +631,562 @@ func (vc *vcopier) fastForward(ctx context.Context, copyState map[string]*sqltyp
...
+				return fmt.Errorf("received result is not done")
+			}
+			if result.State() != vcopierCopyTaskComplete {
+				return fmt.Errorf("received result is not complete")
+			}
+			return nil
+		case <-ctx.Done():
+			// Arguably we should return ctx.Err() here.
+			return nil
+		}
```
mattlord, line 941: Again, here this seems like a potentially serious failure, or no? I would expect this to be treated as an error but we don't seem to be doing any error handling here or on the caller side. Similar questions as above for `SendTo()`.

go/vt/vttablet/tabletmanager/vreplication/vcopier.go
```diff
@@ -364,3 +631,562 @@ func (vc *vcopier) fastForward(ctx context.Context, copyState map[string]*sqltyp
...
+		case vcopierCopyTaskComplete:
+			advanceFn = func(context.Context, *vcopierCopyTaskArgs) error { return nil }
+		}
+
+		// Maybe we should panic here?
+		if advanceFn == nil {
+			err = fmt.Errorf("don't know how to advance from %s to %s", state, nextState)
+			state = vcopierCopyTaskFail
+			break
+		}
```
mattlord, line 1093: I don't think we should panic as the vttablet can be serving unrelated production traffic. This should just be in the case's default though, no? And treated as an invalid state which causes the workflow to fail.

go/vt/vttablet/tabletmanager/vreplication/vcopier.go
```diff
@@ -364,3 +631,562 @@ func (vc *vcopier) fastForward(ctx context.Context, copyState map[string]*sqltyp
...
+		// If all lifecycle hooks and advanceFn are successfully executed,
+		// TryAdvance will return state, which should be == the nextState that
+		// we provided.
+		//
+		// If there was a failure executing lifecycle hooks or advanceFn,
+		// TryAdvance will probably return a failure state (i.e. canceled or
+		// failed), along with a diagnostic error.
+		if state, err = task.Lifecycle().TryAdvance(ctx, task.Args(), nextState, advanceFn); err != nil {
+			break
+		}
```
mattlord, line 1107: I'm a little uneasy/uncertain about the error/failure handling here. We can walk through it.

go/vt/vttablet/tabletmanager/vreplication/vreplicator.go
```diff
@@ -55,7 +55,8 @@ var (
 	// to ensure that it satisfies liveness criteria implicitly expected by internal processes like Online DDL
 	vreplicationMinimumHeartbeatUpdateInterval = 60
 
-	vreplicationExperimentalFlagOptimizeInserts int64 = 1
+	vreplicationExperimentalFlagOptimizeInserts    int64 = 1
+	vreplicationExperimentalParallelizeBulkInserts int64 = 0x02
```
mattlord, line 59: Any reason not to use the same base as the variable, i.e. 2?

mattlord (commented):
First off, great work on this! It's a little heavier on OO patterns than the rest of Vitess but I like it. 🙂 

Most of my comments/concerns fall into these categories:
1. Usage of the `--vreplication_experimental_flags` flag
2. Exporting package functions that I'm not sure we need to
3. Error/failure handling

Happy to sit down together and walk through things. Many of them may simply be misunderstandings on my part. 

Thanks! ❤️

.github/workflows/cluster_endtoend_vreplication_copy_parallel.yml
```diff
@@ -0,0 +1,147 @@
+# DO NOT MODIFY: THIS FILE IS GENERATED USING "make generate_ci_workflows"
```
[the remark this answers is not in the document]
maxenglander: Makes sense, I'll move this to run in the `vreplication_basic` workflow.

go/flags/endtoend/vttablet.txt
```diff
@@ -461,6 +461,7 @@ Usage of vttablet:
       --vreplication_healthcheck_topology_refresh duration               refresh interval for re-reading the topology (default 30s)
       --vreplication_heartbeat_update_interval int                       Frequency (in seconds, default 1, max 60) at which the time_updated column of a vreplication stream when idling (default 1)
       --vreplication_max_time_to_retry_on_error duration                 stop automatically retrying when we've had consecutive failures with the same error for this long after the first occurrence
+      --vreplication_parallel_bulk_inserts string                        Number of parallel insertion workers to use during copy phase. Set to a number >= 1 or to "auto". "auto" uses GOMAXPROCS or 4, whichever is smaller. (default "auto")
```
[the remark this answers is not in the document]
maxenglander: Will rename and make a note about how this flag is gated by `--vreplication_experimental_flags`.

go/vt/vttablet/tabletmanager/vreplication/flags.go
```diff
@@ -39,6 +42,7 @@ var (
 	vreplicationHeartbeatUpdateInterval = 1
 	vreplicationExperimentalFlags       = int64(0x01) // enable vreplicationExperimentalFlagOptimizeInserts by default
 	vreplicationStoreCompressedGTID     = false
+	vreplicationParallelBulkInserts     = "auto"
```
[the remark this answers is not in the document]
maxenglander: > I think that we should avoid using the experimental flag entirely and instead make the default be --vreplication_parallel_insert_workers=1 so that it's disabled by default.

Good suggestion, will do that

go/vt/vttablet/tabletmanager/vreplication/flags.go
```diff
@@ -68,9 +72,39 @@ func registerVReplicationFlags(fs *pflag.FlagSet) {
...
+	if !isExperimentalParallelBulkInsertsEnabled() {
+		return 1
+	}
+	copyInsertConcurrency := string(vreplicationParallelBulkInserts)
+	if copyInsertConcurrency == "auto" {
+		gomaxprocs := runtime.GOMAXPROCS(0)
+		if gomaxprocs <= 4 {
+			return gomaxprocs
+		}
+		return 4
```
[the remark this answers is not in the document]
maxenglander: Good suggestions, will do

go/vt/vttablet/tabletmanager/vreplication/vcopier.go
```diff
@@ -264,86 +441,176 @@ func (vc *vcopier) copyTable(ctx context.Context, tableName string, copyState ma
...
+	var empty bool
+	var terr error
+	for !empty {
+		select {
+		case result := <-resultCh:
+			switch result.State() {
+			case vcopierCopyTaskComplete:
+				lastpk = result.Args().Lastpk()
+			case vcopierCopyTaskFail:
+				if terr != nil {
```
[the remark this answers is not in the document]
maxenglander: Yes this is a bug, good catch.

go/vt/vttablet/tabletmanager/vreplication/vcopier.go
```diff
@@ -264,86 +441,176 @@ func (vc *vcopier) copyTable(ctx context.Context, tableName string, copyState ma
...
+			empty = true
+		}
+	}
+	if terr != nil {
+		log.Warningf("task error: %s", terr.Error())
+		return fmt.Errorf("task error: %s", terr.Error())
+	}
+
+	// Get the last committed pk into a loggable form.
+	var lastpkbuf []byte
```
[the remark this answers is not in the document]
maxenglander: Hm not sure, but agree it seems not needed.

go/vt/vttablet/tabletmanager/vreplication/vcopier.go
```diff
@@ -364,3 +631,562 @@ func (vc *vcopier) fastForward(ctx context.Context, copyState map[string]*sqltyp
...
+}
+
+func (vc *vcopier) newCopyWorkQueue(
+	workerFactory func(context.Context) (*vcopierCopyWorker, error),
+) *vcopierCopyWorkQueue {
+	concurrent := false
+	maxDepth := 1
+	if isExperimentalParallelBulkInsertsEnabled() {
+		concurrent = true
+		maxDepth = getCopyInsertConcurrency()
```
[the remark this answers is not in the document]
maxenglander: Yes, expect to be static

go/vt/vttablet/tabletmanager/vreplication/vcopier.go
maxenglander: Not sure I understand this q, can you rephrase?

go/vt/vttablet/tabletmanager/vreplication/vcopier.go
maxenglander: I think for all of these new types, we don't need to be exporting anything. Also based on your questions I think there's a Golang convention I wasn't aware of until now (a package shouldn't use functions that it exports).

Since none of the methods on these new types need to be exported, I'll make them private.

go/vt/vttablet/tabletmanager/vreplication/vcopier.go
maxenglander: Oops thanks

go/vt/vttablet/tabletmanager/vreplication/vcopier.go
maxenglander: My thinking is that this doesn't change the way things currently work in `main`, which is that if a copy task is inflight, it has to either complete or fail. A context expiration doesn't actually cancel an inflight task.

go/vt/vttablet/tabletmanager/vreplication/vcopier.go
maxenglander: Right now we only use `SendTo` to do two things:

 * Notify the parent `copyTable` scope of completed work, by sending to the `resultCh` declared in that scope.
 * Notify the "next" copy task that it can proceed with updating `_vt.copy_state`.

In terms of what it signifies that `SendTo` would fail to send a result to `ch`, it could mean two things.

One: the parent `copyTable` scope isn't consuming completed work fast enough so `resultCh` fills up. If we get a context expiration while in that state, then `copyTable` scope could in theory close out the work pool, drain `resultCh`, but end up missing this result. If this result completed successfully, that would only impact stats and logging. But if it had an error, that would be bad. I need to think about this.

Two: this task can't notify the "next" task to proceed with `_vt.copy_state`. If we get a context expiration in that state, then the next task will also get a context expiration. That should be fine.

go/vt/vttablet/tabletmanager/vreplication/vreplicator.go
maxenglander: I'm not sure, I think I haven't fully understood how flags are intended to work. In any case, will be following your suggestion above not to use experimental flags at all for this PR.

test/config.json
```diff
@@ -1017,6 +1017,15 @@
 			"RetryMax": 2,
 			"Tags": []
 		},
+		"vreplication_copy_parallel": {
+			"File": "unused.go",
+			"Args": ["vitess.io/vitess/go/test/endtoend/vreplication", "-run", "TestVreplicationCopyParallel"],
+			"Command": [],
+			"Manual": false,
+			"Shard": "vreplication_basic",
+			"RetryMax": 2,
```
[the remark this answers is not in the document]
maxenglander: `vreplication_basic` and `vreplication_copy_throttling` tests, which use the same underlying `vreplication_basic` shard, both have `RetryMax` set to `2`. Does that change your thinking at all, e.g. maybe these tests have some amount of flakiness?

go/vt/vttablet/tabletmanager/vreplication/vcopier.go
maxenglander: But the better answer is we totally can and should enable this. We will just need to re-worker the tests in `vcopier_test.go`. Will talk through that when we sync.

go/vt/vttablet/tabletmanager/vreplication/vcopier.go
maxenglander: We totally can and should enable this. We will just need to re-worker the tests in `vcopier_test.go`. Will talk through that when we sync.

maxenglander: I could have sworn last time I returned `ctx.Err()` from here it broke tests, but just added it back and now it's working. Yay shrug.

go/vt/vttablet/tabletmanager/vreplication/vcopier.go
maxenglander: I've added this back. Turns out it doesn't break tests, I must have misunderstood the test failures I was getting when I last had this code active.

test/config.json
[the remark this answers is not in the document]
maxenglander: Talked and the plan is to bump retries down to 1 in this PR and also open a separate PR to let that bake in `main`.

go/vt/vttablet/tabletmanager/vreplication/vcopier.go
```diff
@@ -364,3 +632,497 @@ func (vc *vcopier) fastForward(ctx context.Context, copyState map[string]*sqltyp
...
+				if err := vbc.vdbClient.Commit(); err != nil {
+					return fmt.Errorf("error commiting transaction: %s", err.Error())
+				}
+				return nil
+			}
+		case vcopierCopyTaskComplete:
+			advanceFn = func(context.Context, *vcopierCopyTaskArgs) error { return nil }
+		}
+
+		if advanceFn == nil {
```
maxenglander, line 1050: Move to `default` case

go/vt/vttablet/tabletmanager/vreplication/vcopier.go
```diff
@@ -364,3 +632,497 @@ func (vc *vcopier) fastForward(ctx context.Context, copyState map[string]*sqltyp
...
+// or failed.
+func (vbc *vcopierCopyWorker) execute(ctx context.Context, task *vcopierCopyTask) *vcopierCopyTaskResult {
+	startedAt := time.Now()
+	state := vcopierCopyTaskPending
+
+	var err error
+
+	// As long as the current state is not done, keep trying to advance to the
+	// next state.
+	for nextState := state.next(); !state.isDone(); nextState = state.next() {
```
maxenglander, line 1007: Consider renaming `next()` so that it's clear that it's not mutating

maxenglander: And/or consider using functions like `isDone(state)`

go/vt/vttablet/tabletmanager/vreplication/vcopier.go
```diff
@@ -364,3 +632,497 @@ func (vc *vcopier) fastForward(ctx context.Context, copyState map[string]*sqltyp
...
+			}
+			if !result.state.isDone() {
+				return fmt.Errorf("received result is not done")
+			}
+			if result.state != vcopierCopyTaskComplete {
+				return fmt.Errorf("received result is not complete")
+			}
+			return nil
+		case <-ctx.Done():
+			return ctx.Err()
```
maxenglander, line 896: Matt: "this cancelation could happen from a PRS, which closes engines when a tablet transitions. We'll typically call `cancel()` on various contexts. In this case, how does this cancelation get propagated back to the parent loop.`

go/vt/vttablet/tabletmanager/vreplication/vcopier.go
```diff
@@ -264,86 +443,175 @@ func (vc *vcopier) copyTable(ctx context.Context, tableName string, copyState ma
...
+
+	// When tasks are executed async, there may be tasks that complete (or fail)
+	// after the last VStreamRows callback exits. Get the lastpk from completed
+	// tasks, or errors from failed ones.
+	var empty bool
+	var terr error
+	for !empty {
+		select {
+		case result := <-resultCh:
+			switch result.state {
```
maxenglander, line 561: Need to handle the canceled case

go/vt/vttablet/tabletmanager/vreplication/vcopier.go
maxenglander: From Zoom discussion:
 * In the current code in `main`, a context deadline/cancelation is handled by logging `Copy stopped` and by returning `nil` in `copyTable`.
 * The current async code in this branch works the same way, so that's good. We can improve things by trying to make it more clear that it works that way by code comments and maybe a bit of refactoring.

go/vt/vttablet/tabletmanager/vreplication/vcopier.go
maxenglander: Max: "this `awaitCompletion` hook is triggered in `tryAdvance`. This `ctx.Err()` will be captured by `tryAdvance()`, which will return that `err` along with `vcopierCopyTaskCancel` to `vcopierCopyWorker.execute()`, which will publish a failure result to any channels that are subscribed to task results. The "global" `resultCh` is one such subscriber, which will see that the task was canceled and return `io.EOF` while inside the `VStreamRows` loop, or `nil` when outside the `VStreamRows` loop."

go/flags/endtoend/vttablet.txt
```diff
@@ -460,6 +460,7 @@ Usage of vttablet:
       --vreplication_healthcheck_topology_refresh duration               refresh interval for re-reading the topology (default 30s)
       --vreplication_heartbeat_update_interval int                       Frequency (in seconds, default 1, max 60) at which the time_updated column of a vreplication stream when idling (default 1)
       --vreplication_max_time_to_retry_on_error duration                 stop automatically retrying when we've had consecutive failures with the same error for this long after the first occurrence
+      --vreplication_parallel_insert_workers int                         Number of parallel insertion workers to use during copy phase. Set <= 1 to disable parallelism, or > 1 enable concurrent insertion during copy phase. (default 1)
```
mattlord, line 463: We should use dashes rather than underscores in new flag names too. Sorry that I didn't note this earlier. The current plan is to move all flags to dashes in v16 so this will avoid future work.

go/vt/vttablet/tabletmanager/vreplication/vcopier.go
```diff
@@ -364,3 +640,522 @@ func (vc *vcopier) fastForward(ctx context.Context, copyState map[string]*sqltyp
...
+			return nil
+		case <-ctx.Done():
+			// A context expiration probably indicates a PlannedReparentShard
+			// or an elapsed copy phase duration. Those aren't treated as error
+			// conditions, but we'll return the context error here anyway.
+			//
+			// Task execution will detect the presence of the error, mark this
+			// task canceled, and abort. Subsequent tasks won't execute because
+			// this task didn't complete.
+			return ctx.Err()
```
mattlord, line 926: We should follow [the established convention](https://github.com/vitessio/vitess/search?q=Code_Canceled) of returning a `vterror` with the Canceled code:
```
return vterrors.Errorf(vtrpcpb.Code_CANCELED, "context has expired")
```

go/vt/vttablet/tabletmanager/vreplication/vcopier.go
```diff
@@ -264,86 +443,183 @@ func (vc *vcopier) copyTable(ctx context.Context, tableName string, copyState ma
...
 		}
 
-		// The number of rows we receive depends on the packet size set
-		// for the row streamer. Since the packet size is roughly equivalent
-		// to data size, this should map to a uniform amount of pages affected
-		// per statement. A packet size of 30K will roughly translate to 8
-		// mysql pages of 4K each.
-		if err := vc.vr.dbClient.Begin(); err != nil {
-			return err
+		// Clone rows, since pointer values will change while async work is
```
rohit-nayak-ps, line 464: @maxenglander, should we optimize this by not cloning, in case we have only one worker there is no parallelism?

maxenglander: Yes will do

go/vt/vttablet/tabletmanager/vreplication/vcopier.go
```diff
@@ -391,3 +668,526 @@ func (vc *vcopier) fastForward(ctx context.Context, copyState map[string]*sqltyp
...
+
+		// tryAdvance tries to execute advanceFn, and also executes any
+		// lifecycle hooks surrounding the provided state.
+		//
+		// If all lifecycle hooks and advanceFn are successfully executed,
+		// tryAdvance will return state, which should be == the nextState that
+		// we provided.
+		//
+		// If there was a failure executing lifecycle hooks or advanceFn,
+		// tryAdvance will probably return a failure state (i.e. canceled or
```
mattlord, line 1097: `probably` makes me a little nervous here 🙂 We should have some code to handle the unexpected/edge cases

go/vt/vttablet/tabletmanager/vreplication/vreplicator.go
```diff
@@ -350,8 +350,18 @@ func (vr *vreplicator) buildColInfoMap(ctx context.Context) (map[string][]*Colum
 	return colInfoMap, nil
 }
 
-func (vr *vreplicator) readSettings(ctx context.Context) (settings binlogplayer.VRSettings, numTablesToCopy int64, err error) {
-	settings, err = binlogplayer.ReadVRSettings(vr.dbClient, vr.id)
+// Same as readSettings, but stores some of the results on this vr.
+func (vr *vreplicator) loadSettings(ctx context.Context, dbClient *vdbClient) (settings binlogplayer.VRSettings, numTablesToCopy int64, err error) {
```
mattlord, line 354: I don't see the need for these two function changes, what am I missing?

mattlord (commented):
Once again, great work on this! I only had some minor nits/questions/suggestions. We can huddle on those and get this merged. :heart:

go/vt/binlog/binlogplayer/binlog_player.go
```diff
@@ -160,6 +163,8 @@ func NewStats() *Stats {
 	bps.NoopQueryCount = stats.NewCountersWithSingleLabel("", "", "Statement", "")
 	bps.VReplicationLags = stats.NewTimings("", "", "")
 	bps.VReplicationLagRates = stats.NewRates("", bps.VReplicationLags, 15*60/5, 5*time.Second)
+	bps.TableCopyRowCounts = stats.NewCountersWithSingleLabel("", "", "Table", "")
+	bps.TableCopyTimings = stats.NewTimings("", "", "Table")
```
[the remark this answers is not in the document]
maxenglander: Here's a sample:
```
"VReplicationTableCopyRowCounts": {"src.-.copy.1.data": 8388608},
```

go/vt/vttablet/tabletmanager/vreplication/stats.go
```diff
@@ -355,6 +355,47 @@ func (st *vrStats) register() {
...
+			result := make(map[string]int64, len(st.controllers))
+			for _, ct := range st.controllers {
+				for table, t := range ct.blpStats.TableCopyTimings.Histograms() {
+					result[ct.source.Keyspace+"."+ct.source.Shard+"."+ct.workflow+"."+fmt.Sprintf("%v", ct.id)+"."+table] = t.Total()
+				}
+			}
+			return result
+		})
+
+	stats.NewGaugeFunc("VReplicationBulkInsertConcurrency", "Number of concurrent bulk inserts", st.bulkInsertConcurrency)
```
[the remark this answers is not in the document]
maxenglander: I think we probably don't need this anymore now that we're not using `GOMAXPROCS` to choose the parallelism. Removing.

go/vt/vttablet/tabletmanager/vreplication/vcopier.go
```diff
@@ -32,26 +40,185 @@ import (
...
+
+func newVCopierCopyTaskHooks() *vcopierCopyTaskHooks {
+	return &vcopierCopyTaskHooks{
+		fns: make([]func(context.Context, *vcopierCopyTaskArgs) error, 0),
+	}
+}
+
+func newVCopierCopyTaskLifecycle() *vcopierCopyTaskLifecycle {
+	return &vcopierCopyTaskLifecycle{
+		hooks:       make(map[string]*vcopierCopyTaskHooks),
```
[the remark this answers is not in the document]
maxenglander: `newVCopierCopyTaskHooks()` returns `*vcopierCopyTaskHooks`, which is a struct.

go/vt/vttablet/tabletmanager/vreplication/vcopier.go
```diff
@@ -291,86 +470,184 @@ func (vc *vcopier) copyTable(ctx context.Context, tableName string, copyState ma
...
-			vc.vr.stats.QueryCount.Add("copy", 1)
-			return qr, err
-		})
-		if err != nil {
-			return err
+		// Prepare a vcopierCopyTask for the current batch of work.
+		// TODO(maxeng) see if using a pre-allocated pool will speed things up.
+		currCh := make(chan *vcopierCopyTaskResult, 1)
+		currT := newVCopierCopyTask(taskID, newVCopierCopyTaskArgs(rows.Rows, rows.Lastpk))
+		taskID++
```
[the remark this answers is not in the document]
maxenglander: I'm not sure. Right now `taskID` isn't used for anything. I had been using it during development for debugging. Going to remove it in next commit to avoid any confusion.

go/vt/vttablet/tabletmanager/vreplication/vcopier.go
```diff
@@ -291,86 +470,184 @@ func (vc *vcopier) copyTable(ctx context.Context, tableName string, copyState ma
...
+	if terr != nil {
+		log.Warningf("task error: %s", terr.Error())
+		return fmt.Errorf("task error: %s", terr.Error())
+	}
+
+	// Get the last committed pk into a loggable form.
+	lastpkbuf, merr := prototext.Marshal(&querypb.QueryResult{
+		Fields: pkfields,
+		Rows:   []*querypb.Row{lastpk},
+	})
```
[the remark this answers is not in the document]
maxenglander: I'm not sure either, better in what sense?

go/vt/vttablet/tabletmanager/vreplication/vcopier.go
```diff
@@ -391,3 +668,526 @@ func (vc *vcopier) fastForward(ctx context.Context, copyState map[string]*sqltyp
...
+			if err != nil {
+				return nil, fmt.Errorf("failed to create new db client: %s", err.Error())
+			}
+			return newVCopierCopyWorker(
+				true, /* close db client */
+				dbClient,
+			), nil
+		}
+	}
+	return workerFactory
```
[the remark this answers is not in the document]
maxenglander: gofmt prefers I don't use if-else, but I think I can remove the inefficiency

go/vt/vttablet/tabletmanager/vreplication/vcopier.go
```diff
@@ -391,3 +668,526 @@ func (vc *vcopier) fastForward(ctx context.Context, copyState map[string]*sqltyp
...
+		nil,          /* log wait */
+		nil,          /* refresh check */
+		0,            /* refresh interval */
+	)
+
+	vcq.isOpen = true
+}
+
+// after returns a vcopierCopyTaskHooks that can be used to register callbacks
+// to be triggered after the task successfully advances past to the specified
```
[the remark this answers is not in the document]
maxenglander: I can make that change, but I think "finishes/completes" isn't totally accurate since hooks can be executed after any `vcopierCopyTaskState`, and not only when a task finishes/completes.

go/vt/vttablet/tabletmanager/vreplication/vcopier.go
```diff
@@ -391,3 +668,526 @@ func (vc *vcopier) fastForward(ctx context.Context, copyState map[string]*sqltyp
...
+			//     stopping the copy phase without completing it.
+		}
+	})
+}
+
+// awaitCompletion registers a callback that returns an error unless the
+// provided chan produces a vcopierTaskResult in a complete state.
+//
+// This is useful for sequencing vcopierCopyTasks, e.g.:
+//
```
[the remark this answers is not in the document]
maxenglander: Not sure, will try removing

go/vt/vttablet/tabletmanager/vreplication/vcopier.go
```diff
@@ -391,3 +668,526 @@ func (vc *vcopier) fastForward(ctx context.Context, copyState map[string]*sqltyp
...
+			// Task execution will detect the presence of the error, mark this
+			// task canceled, and abort. Subsequent tasks won't execute because
+			// this task didn't complete.
+			return vterrors.Errorf(vtrpcpb.Code_CANCELED, "context has expired")
+		}
+	})
+}
+
+// do registers a callback with the vcopierCopyTaskResultHooks, to be triggered
+// when as a vcopierCopyTask advances up to or beyond a user-specified state.
```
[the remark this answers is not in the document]
maxenglander: Yes, I'll switch to using `before or after` instead of `advances up to or beyond`

go/vt/vttablet/tabletmanager/vreplication/vreplicator.go
maxenglander: I want to be able to re-use logic in `vreplicator.go` for setting things like `sql_mode` and disabling FK checks. I want to use that logic in each new parallel MySQL conn that's created in `vcopier.go`. The current `vreplicator.go` code makes that a little difficult because it changes state in `vreplicator`. 

To address that I move the logic I wanted to re-use into a stateless fn (`readSettings`), and move the stateful stuff into `loadSettings`.

mattlord (approved):

PR #10835, opened 2022-07-26
Always close body for HTTP requests in tests

dbussink:
Not closing the body leads to leaking goroutines for the reader / writer of the body.

Found when validating that tests don't leak goroutines (to ensure other things get closed properly, but these cases where also found).

## Related Issue(s)

See also https://github.com/vitessio/vitess/pull/10756 for another set of leak fixes.

## Checklist


vmg (approved):

vmg (approved):

ajm188 (approved):
LGTM

Issue #10841, opened 2022-07-26
Bug Report: Builtin Backups Fail with default behaviour of MySQL 8.0.30


GuptaManan100:
### Overview of the Issue

We noticed that the CI was failing on #10839 https://github.com/vitessio/vitess/runs/7520246527?check_suite_focus=true.
On investigation, it turns out that the failure is because of the recently released MySQL 8.0.30. In this version the default way we store the redo-logs has been changed. 

In MySQL 8.0.30 docs it says - 
```
Redo log files reside in the #innodb_redo directory in the data directory unless a different directory is specified by the innodb_log_group_home_dir variable. There are two types of redo log files, ordinary and spare. Ordinary redo log files are those being used. Spare redo log files are those waiting to be used. InnoDB tries to maintain 32 redo log files in total, with each file equal in size to 1/32 * innodb_redo_log_capacity; however, file sizes may differ for a time after modifying the innodb_redo_log_capacity setting.
```
and prior to 8.0.30 - 
```
Prior to MySQL 8.0.30, InnoDB creates two redo log files in the data directory by default, named ib_logfile0 and ib_logfile1, and writes to these files in a circular fashion.
```

This change in default behavior causes the builtin backups to fail - 
```
E0726 12:49:25.895688   31639 main.go:96] E0726 12:49:25.895533 backup.go:107] E0726 12:49:25.895397 backup.go:127] backup is not usable, aborting it: [cannot copy data: read /home/runner/work/vitess/vitess/vtdataroot/vt_0000000102/innodb/logs/#innodb_redo: is a directory]
[60](https://github.com/vitessio/vitess/runs/7520246527?check_suite_focus=true#step:18:61)
BackupShard Error: rpc error: code = Unknown desc = TabletManager.Backup on zone1-0000000102 error: cannot copy data: read /home/runner/work/vitess/vitess/vtdataroot/vt_0000000102/innodb/logs/#innodb_redo: is a directory: cannot copy data: read /home/runner/work/vitess/vitess/vtdataroot/vt_0000000102/innodb/logs/#innodb_redo: is a directory
[61](https://github.com/vitessio/vitess/runs/7520246527?check_suite_focus=true#step:18:62)
[...]

PR #10847, opened 2022-07-26
Backups: Support InnoDB Redo Log Location With 8.0.30+

mattlord:
## Description

[The default location of the InnoDB redo logs changed to `<innodb_log_group_home_dir>/#innodb_redo/` in 8.0.30](https://dev.mysql.com/doc/refman/8.0/en/innodb-redo-log.html#innodb-modifying-redo-log-capacity). From the [8.0.30 release notes](https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-30.html):
> InnoDB: InnoDB now supports dynamic configuration of redo log capacity. The [innodb_redo_log_capacity](https://dev.mysql.com/doc/refman/8.0/en/innodb-parameters.html#sysvar_innodb_redo_log_capacity) system variable can be set at runtime to increase or decrease the total amount of disk space occupied by redo log files.
>
> With this change, the number of redo log files and their default location has also changed. From MySQL 8.0.30, InnoDB maintains 32 redo log files in an #innodb_redo directory in the data directory. Previously, InnoDB created two redo log files in the data directory by default, and the number and size of redo log files were controlled by the and [innodb_log_files_in_group](https://dev.mysql.com/doc/refman/8.0/en/innodb-parameters.html#sysvar_innodb_log_files_in_group) and [innodb_log_file_size](https://dev.mysql.com/doc/refman/8.0/en/innodb-parameters.html#sysvar_innodb_log_file_size) variables. These two variables are now deprecated.
>
> When an innodb_redo_log_capacity setting is defined, innodb_log_files_in_group and innodb_log_file_size settings are ignored; otherwise, those settings are used to compute the innodb_redo_log_capacity setting (innodb_log_files_in_group * innodb_log_file_size = innodb_redo_log_capacity). If none of those variables are set, redo log capacity is set to the innodb_redo_log_capacity default value, which is 104857600 bytes (100MB).
>
> Several status variables are provided for monitoring the redo log and redo log capacity r`esize operations.
>
> As is generally required for any upgrade, this change requires a clean shutdown before upgrading.
>
> For more information about this feature, see [Redo Log](https://dev.mysql.com/doc/refman/8.0/en/innodb-redo-log.html).

This PR alters the backups and tests to account for this change when 8.0.30+ is being used. 

ℹ️ This PR also pins the version of MySQL used in the `Upgrade Downgrade Testing - Backups - Manual` workflow to 8.0.29 because v14.0 binaries do not have these fixes, and ubuntu-20.04 since MySQL has no 8.0.29 packages for 22.04. This pinning can be removed once either of the following occurs:
- The fixes in this PR are backported to `vitessio/release-14`
- v15 is GA and `vitessio/main` has a dev version of v16.0.0-SNAPSHOT

My current tentative plan is to backport these fixes — _excluding the workflow changes_ — to at least v14. I would then do a follow-up PR on main to revert the changes made to the workflow (`.github/workflows/upgrade_downgrade_test_backups_manual.yml`).

## Related Issue(s)

 - Fixes: https://github.com/vitessio/vitess/issues/10841

## Checklist

-   [x] Tests were added/modified


go/mysql/flavor.go
```diff
@@ -51,6 +51,7 @@ const (
 	InstantChangeColumnDefaultFlavorCapability
 	MySQLJSONFlavorCapability
 	MySQLUpgradeInServerFlavorCapability
+	DynamicRedoLogCapacityFlavorCapability
```
deepthi, line 54: add a comment like `// supported in 8.0.30 and above ...`?

go/vt/mysqlctl/backupengine.go
```diff
@@ -381,40 +381,65 @@ func addMySQL8DataDictionary(fes []FileEntry, base string, baseDir string) ([]Fi
...
-	result, size, err := addDirectory(result, backupInnodbLogGroupHomeDir, cnf.InnodbLogGroupHomeDir, "")
-	if err != nil {
-		return nil, 0, err
+	var size, totalSize int64
+	var flavor MySQLFlavor
+	var version ServerVersion
+	var features capabilitySet
+
+	// get the flavor and version to deal with any behavioral differences
+	{
```
deepthi, line 390: nit: we have tried to avoid this style of using `{ ... }` to delineate blocks of code. The comments in between the various blocks should be sufficient.

deepthi (commented):
Nice work, very clean PR.
The test cases are good. I have a couple of minor suggestions, otherwise LGTM.
We'll let the others take a look as well before approve/merge.

mattlord:
> Nice work, very clean PR.
> The test cases are good. I have a couple of minor suggestions, otherwise LGTM.
> We'll let the others take a look as well before approve/merge.

Thanks, @deepthi ! I've implemented both of your suggestions here: https://github.com/vitessio/vitess/pull/10847/commits/40e9c298e201aa974c972a2ce25fa8ab6bdcbe7a

mattlord:
Moved this back to draft as I need to confirm something and cannot seem to ATM due to: https://bugs.mysql.com/bug.php?id=108016

If there are other files in the `innodb_log_group_home_dir` directory — aside from the `#innodb_redo` subdirectory, then we need to treat that log dir in the same [way that we treat the `datadir` for backups](https://github.com/vitessio/vitess/blob/22406441548e303b801da3ad277a06de0482f784/go/vt/mysqlctl/backupengine.go#L403-L418): backing up each file in that directory AND any immediate subdirectories (only 1 level deep).

mattlord:
Update: the issue I ran into was NOT 8.0.30 related in the end: https://bugs.mysql.com/bug.php?id=108017

Will do more testing and another patch if needed.

GuptaManan100 (commented):
The code LGTM especially the capability check using the server version 💯

GuptaManan100 (commented):
We need release notes changes too for v14 and v13. That however will be separate PRs right?

mattlord:
> We need release notes changes too for v14 and v13. That however will be separate PRs right?

Yeah, I'll do backport PRs. You think we should do 14 and 13? The bug fix/feature will be in the release notes either way, you're saying we should mention it in the release notes summary?

deepthi:
> We need release notes changes too for v14 and v13. That however will be separate PRs right?

You mean documenting the 8.0.30 incompatibility as a known issue? Yeah, those have to be separate PRs on the release branches + update to the actual release content in GH releases.

deepthi (approved):

rsajwani:
Changes LGTM. Wondering we don't have similar test cases for XtrabackupEngine which can verify these changes for Xtrabackup

mattlord:
> Changes LGTM. Wondering we don't have similar test cases for XtrabackupEngine which can verify these changes for Xtrabackup

Our XtraBackup tests use Percona's Apt Repo as the version of XtraBackup 8.0 often does not support the latest upstream MySQL 8.0 release for some time: https://github.com/vitessio/vitess/pull/10691

I did run the xbbackup and xbstream e2e tests locally though with 8.0.30 and they passed.

mattlord:
> Changes LGTM. Wondering we don't have similar test cases for XtrabackupEngine which can verify these changes for Xtrabackup

Our XtraBackup tests use Percona's Apt Repo as the latest version of XtraBackup 8.0 ([8.0.29 as of July 17](https://docs.percona.com/percona-xtrabackup/8.0/release-notes/8.0/8.0.29-22.0.html)) often does not support the latest upstream MySQL 8.0 release (8.0.30 as of today) for some time: https://github.com/vitessio/vitess/pull/10691

Issue #10846, opened 2022-07-26
Release of `v12.0.5`


frouioui:
## Description

This issue lists the actions that need to be taken before/during/after the `v12.0.5` release.

### Pre-Release

- [x] Make sure the release note summary is prepared and clean
  - [x] https://github.com/vitessio/vitess/pull/10864
- [x] Make sure the Pull Request titles that are included in the release notes are well-written and explicit
- [x] Make sure backports are reviewed and merged onto `release-12.0`
- [x] Upgrade go-version to `go1.17.12` #10865 

### Release

- [x] Follow the release instructions for release: [here](https://github.com/vitessio/vitess/blob/main/doc/internal/ReleaseInstructions.md#pre-requisites-for-releases)
- [x] Continue with the Java release following [these](https://github.com/vitessio/vitess/blob/main/doc/internal/ReleaseInstructions.md#java-packages-deploy--release) instructions.
- [x] `v12.0.5` is released on GitHub and DockerHub
  - [x] DockerHub
  - [x] GitHub
- [x] Merge the `12.0.5` release notes on `main`
  - [x] https://github.com/vitessio/vitess/pull/10873
- [x] Merge the `12.0.5` release notes on `release-12.0`
  - [x] https://github.com/vitessio/vitess/pull/10872
- [x] Make sure arewefastyet benchmarks the new release (https://benchmark.vitess.io/status)

### Post-Release

- [x] Benchmarking results: https://benchmark.vitess.io/macrobench?ltag=12.0.0&rtag=12.0.5
- [x] Notify the community on:
  - [x] Slack
  - [x] Twitter

PR #10864, opened 2022-07-27
Addition of the release summary for v12.0.5

frouioui:
## Description

This Pull Request adds the release summary for v12.0.5.

## Related Issue(s)
- https://github.com/vitessio/vitess/issues/10846

## Checklist

### DIFF
--- a/.github/workflows/cluster_endtoend_upgrade.yml
+++ b/.github/workflows/cluster_endtoend_upgrade.yml
@@ -94,9 +94,6 @@ jobs:
         mkdir -p /tmp/vitess-build-${{ matrix.project }}/
         cp -R bin /tmp/vitess-build-${{ matrix.project }}/
 
-    - name: Check out HEAD
-      uses: actions/checkout@v2
-
     - name: Run cluster endtoend test ${{ matrix.project }} (create cluster)
       timeout-minutes: 5
       run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/cluster_endtoend_upgrade.yml
```diff
@@ -94,9 +94,6 @@ jobs:
         mkdir -p /tmp/vitess-build-${{ matrix.project }}/
         cp -R bin /tmp/vitess-build-${{ matrix.project }}/
 
-    - name: Check out HEAD
-      uses: actions/checkout@v2
-
```
GuptaManan100, line 99: Why are we removing this?

GuptaManan100 (approved):

.github/workflows/cluster_endtoend_upgrade.yml
frouioui: We want to create a cluster using v13.0.0 data so we can later run the E2E test of HEAD against that cluster. This was creating a cluster using HEAD data instead of v13.0.0

go/test/endtoend/cluster/vtctlclient_process.go
```diff
@@ -47,8 +47,16 @@ func (vtctlclient *VtctlClientProcess) InitShardPrimary(Keyspace string, Shard s
...
+	if err != nil {
+		return err
+	}
+	cmd := "InitShardMaster"
+	if version > 12 {
+		cmd = "InitShardPrimary"
+	}
 	output, err := vtctlclient.ExecuteCommandWithOutput(
-		"InitShardMaster",
+		cmd,
```
frouioui, line 59: This change was required to unblock the upgrade/downgrade tests. We used to execute v13.0.0 tests using the v12 codebase, and `InitShardMaster` does not exist in v13.0.0

PR #10863, opened 2022-07-26
[cli][discovery]: migrate discovery flags to pflag

deepthi:
## Description
Scopes flags in discovery package to the appropriate binaries.
* healthcheck flags - vtgate and vtcombo
* replication lag flags - vtgate only

In the process, I have also renamed `TabletTypeListType` to `TabletTypeListFlag`, implemented `Type()` to conform to `pflag.Value` and deleted the no-longer needed helper function.

Additionally, this PR extends the `topoproto.TabletTypeFlag` type to also implement `pflag.Value`, and then switches the single occurrence (in `go/vt/vtgate/`) to `pflag` and scopes to the binaries that need it (N.B. `vtadmin` also imports it but does not care about the flag being registered).

Then, we no longer need the helper function, so I deleted it.

EDIT: Per @ajm188's suggestion offline, I've folded #10867 into this PR and copied over the description.

## Related Issue(s)
Fixes #10751 
Fixes #10861 
Closes #10867 

## Checklist

-   [x] "Backport me!" label has been added if this change should be backported
-   [x] Tests were added or are not required
-   [x] Documentation was added or is not required

PR #10867, opened 2022-07-27
[cli] [vtgate] Add `pflag.Value` implementation to `topoproto.TabletTypeFlag`, migrate only flag instance

### LINKED ISSUE #10861: Switch flag definitions to be on pflag instead of flag in `package go/vt/topo/topoproto`

Current flags:

```
$ git grep -E "\bflag\.[A-Z]" -- go/vt/topo/topoproto/*.go
go/vt/topo/topoproto/flag.go:   flag.Var((*TabletTypeListValue)(p), name, usage)
go/vt/topo/topoproto/flag.go:   flag.Var((*TabletTypeFlag)(p), name, usage)
go/vt/topo/topoproto/flag.go:// TabletTypeListValue implements the flag.Value interface, for parsing a command-line comma-separated
go/vt/topo/topoproto/flag.go:// String is part of the flag.Value interface.
go/vt/topo/topoproto/flag.go:// Set is part of the flag.Value interface.
go/vt/topo/topoproto/flag.go:// TabletTypeFlag implements the flag.Value interface, for parsing a command-line value into a TabletType.
go/vt/topo/topoproto/flag.go:// String is part of the flag.Value interface.
go/vt/topo/topoproto/flag.go:// Set is part of the flag.Value interface.
```

ajm188:
## Description

This extends the `topoproto.TabletTypeFlag` type to also implement `pflag.Value`, and then switches the single occurrence (in `go/vt/vtgate/`) to pflag and scopes to the binaries that need it (N.B. `vtadmin` also imports it but does not care about the flag being registered).

Then, we no longer need the helper function, so I deleted it.

## Related Issue(s)

Partially fixes #10861.

## Checklist

-   [x] "Backport me!" label has been added if this change should be backported - n/a

### DIFF
--- a/go/vt/topo/topoproto/flag.go
+++ b/go/vt/topo/topoproto/flag.go
@@ -29,13 +29,6 @@ func TabletTypeListVar(p *[]topodatapb.TabletType, name string, usage string) {
 	flag.Var((*TabletTypeListValue)(p), name, usage)
 }
 
-// TabletTypeVar defines a TabletType flag with the specified name, default value and usage
-// string. The argument 'p' points to a tabletType in which to store the value of the flag.
-func TabletTypeVar(p *topodatapb.TabletType, name string, defaultValue topodatapb.TabletType, usage string) {
-	*p = defaultValue
-	flag.Var((*TabletTypeFlag)(p), name, usage)
-}
-
 // TabletTypeListValue implements the flag.Value interface, for parsing a command-line comma-separated
 // list of value into a slice of TabletTypes.
 type TabletTypeListValue []topodatapb.TabletType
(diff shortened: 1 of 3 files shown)


go/vt/topo/topoproto/flag.go
```diff
@@ -56,21 +49,24 @@ func (ttlv *TabletTypeListValue) Get() any {
 	return *ttlv
 }
 
-// TabletTypeFlag implements the flag.Value interface, for parsing a command-line value into a TabletType.
+// TabletTypeFlag implements the flag.Value and pflag.Value interfaces, for parsing a command-line value into a TabletType.
```
deepthi, line 52: Do we really need to keep the `flag.Value` implementation?

deepthi: If we decide no, we can delete `Get()` and change the comments on the remaining funcs.

deepthi (approved):

go/vt/topo/topoproto/flag.go
ajm188: https://github.com/vitessio/vitess/pull/10863#discussion_r931565945

Issue #10832, opened 2022-07-26
Question: 


vczyh:
### Question

The `COM_QUIT ` command was not sent before closing TCP connection. It would cause mysql status `Aborted_clients`  to increase.

```
mysql> show global status like "aborted_clients";
+-----------------+-------+
| Variable_name   | Value |
+-----------------+-------+
| Aborted_clients | 578   |
+-----------------+-------+
1 row in set (0.00 sec)
```

The right way to close connection: https://dev.mysql.com/doc/internals/en/com-quit.html

PR #10869, opened 2022-07-27
Fix the MySQL connection is closed incorrectly

vczyh:
Signed-off-by: Zhang,Yuheng <zhangeek@qq.com>

## Related Issue(s)

#10832

## Checklist


harshit-gangal (commented):
Could add a test for this change.

deepthi:
actually `unit_race` is failing. @vczyh can you run that locally and fix that? It looks related to this code change.

PR #10875, opened 2022-07-28
Fix pprof toggling via SIGUSR1 and waitSig flag

saunderst:
## Description

**Bug fix:**
When using the `waitSig` option with the `pprof` flag (e.g. `-pprof=cpu,waitSig`, which should enable CPU profiling after sending `SIGUSR1` to the target process), I expected that the first signal sent would enable profiling, while the second signal would disable profiling and perform any cleanup required to complete the profile.  I consistently observed that the first signal had no effect, the second signal enabled profiling, and subsequent signals had no effect.  The only way to complete profiling was to kill the process, which was not ideal for my use case.  The cause proved to be that [these two goroutines ](https://github.com/vitessio/vitess/blob/main/go/vt/servenv/pprof.go#L313-L325) were sharing a notification channel, and the "stop" action was being executed before the "start".

With this change, the feature behaves as follows: with the `waitSig` option, profiling will begin upon receipt of the first signal; without `waitSig`, profiling will begin immediately upon startup.  In either case, subsequent signals will toggle profiling, with each "start" creating a new profile.  Profiling will also complete (if running) upon exit, as before.

## Related Issue(s)

none

## Checklist

-   [ ] "Backport me!" label has not been added
-   [ ] Tests were added
-   [ ] Additional documentation not required

## Deployment Notes

none


go/vt/servenv/pprof.go
```diff
@@ -299,32 +299,45 @@ func (prof *profile) init() (start func(), stop func()) {
 	}
 }
```
saunderst, line 301: This diff isn't very readable.  It may be easier to just look at the file starting at line 300.

go/vt/servenv/pprof.go
```diff
@@ -299,32 +299,45 @@ func (prof *profile) init() (start func(), stop func()) {
...
 
-			OnTerm(stop)
-		}
-	})
+		OnTerm(stop)
+	}
+}
+
+func init() {
+	OnInit(pprof_init)
```
saunderst, line 342: Extracted `pprof_init` for testability.

frouioui (approved):
Hello @saunderst, thank you for your contribution! Your change is looking good to me. I will approve it.

@vmg, we added this feature a while ago, do you know if we wanted to keep the use of signals around? (in the long term)

vmg (approved):
We should definitely keep signal profiling. Thanks for the contribution!

PR #10876, opened 2022-07-28
onlineddl_vrepl flakiness: further work

shlomi-noach:
This PR is based on @mattlord 's #10874. Some flaky failures we only ever see on "weak" machines, such as in GitHub CI, and they do not reproduce in a strong test environment.

This PR is gonna be kept open for a while, with many re-runs of `onlineddl_vrepl` tests (mainly over `mysql80`). I'll be incrementally adding changes from over #10874 .

### DIFF
--- a/go/test/endtoend/onlineddl/vrepl/onlineddl_vrepl_test.go
+++ b/go/test/endtoend/onlineddl/vrepl/onlineddl_vrepl_test.go
@@ -374,7 +374,7 @@ func TestSchemaChange(t *testing.T) {
 		require.NotNil(t, row)
 		// vplayer and vcopier update throttle timestamp every second, so we expect the value
 		// to be strictly higher than started_timestamp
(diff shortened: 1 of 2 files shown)


shlomi-noach:
@mattlord with this partial import from your #10874 I'm unable to reproduce flakiness in `onlineddl_vrepl` tests; I've run them about 20 times or so, and they consistently pass. There might be further more rare flakiness -- but would it make sense to you to merge this PR as is, and based on further flakiness iterate on the rest of your PR?

mattlord (approved):
Thank you! ❤️

shlomi-noach:
> Although I'm not sure what you don't like in the larger PR?

Haha nothing, I just realized with these initial iterations I don't see anymore flakiness so I felt like the minimal set of changes could be a good choice? Also, I haven't fully reviewed the remaining changes. I spot at least two out of the eight tests you mention above, that I know the PR fixes. I'll keep an eye over the rest.

PR #10879, opened 2022-07-28
feat: don't use the vtgatehandler unless it is known to have been initialized

systay:
## Description

When a vtgate is started and very quickly terminated, it can panic. This is because the shutdown process tried to use parts of the system that have not yet been fully initialized. This change protects the `vtgateHandler` from accidental use before init has happened.

### DIFF
--- a/go/vt/vtgate/plugin_mysql_server.go
+++ b/go/vt/vtgate/plugin_mysql_server.go
@@ -550,6 +550,11 @@ func rollbackAtShutdown() {
 		}
 	}()
 
+	if vtgateHandle == nil {
+		// we still haven't been able to initialise the vtgateHandler, so we don't need to rollback anything
+		return
+	}
+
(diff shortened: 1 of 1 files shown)


frouioui (approved):
LGTM

go/vt/vtgate/plugin_mysql_server.go
```diff
@@ -550,6 +550,11 @@ func rollbackAtShutdown() {
 		}
 	}()
 
+	if vtgateHandle == nil {
```
deepthi, line 553: nit: I'd have done this before starting the goroutine which again checks for nil.

systay: You and your logic... 😀

systay: https://github.com/vitessio/vitess/pull/10887

Issue #10880, opened 2022-07-28
Bug Report: Semi-sync not setup correctly on VTTablet startup


GuptaManan100:
### Overview of the Issue

This bug was found when investigating flakiness in an EmergencyReparentShard test `TestERSPromoteRdonly`. 

It was observed that the following sequence of steps were followed - 
1. We bring up a new cluster with new tablets
2. We try to run PRS
3. Meanwhile, tablets go to RESTORE mode as part of the startup
4. PRS sees the tablets are of type RESTORE and doesn’t set semi-sync on them
5. Tablet complete RESTORE and go to type REPLICA (but semi-sync is off)
6. Test fails assuming semi-sync is on (Blocks on a write)

This uncovered a deeper problem. A vttablet on startup won't set its semi-sync properly. 
The reason for this is quite clear. Earlier vttablets used the `enable_semi_sync` flag which meant that when the replication manager repaired replication (even on startup), semi-sync was automatically fixed.

This however changed, since we started sending the semi-sync information in the RPCs. This led us to not fix semi-sync at all on startup.

### Reproduction Steps

1. Setup a cluster with `semi_sync` durability
2. Bring up a new `REPLICA` tablet
3. Wait and see that semi-sync is off on the tablet.

### Binary Version

```sh
main. release-14.0 and before used the flag, so they are unaffected.
```

### Operating System and Environment details

Found on CI.

PR #10881, opened 2022-07-28
BugFix: Vttablet semi-sync settings on startup

GuptaManan100:
## Description
This PR addresses the issue raised in https://github.com/vitessio/vitess/issues/10880. 

The proposed fix for the problem is to initialize the replication once on startup. Since we have moved to the architecture of storing the durability policy in the topo server instead of being passed in vtctld flags, even vttablets can access that information.

This allows the vttablets to find the current primary when they start and consult the durability policies to know whether they should be setting semi-sync or not.

A new method called `initializeReplication` has been added to the tablet manager which (as the name suggests) is responsible for initializing replication. This method is called in 3 places - 
1. In restoreDataLocked in the case where no backup exists.
2. In restoreDataLocked when a backup exists (as part of `tm.startReplication` call)
3. At the end of `Start` of tablet manager which handles the case wherein we don't restore at all.4. 

Together these 3 calls ensure that we run `initializeReplication` once in all the cases of different restores.

An end to end test has also been added which exercises the `initializeReplication` logic and checks that semi-sync is setup correctly.

## Related Issue(s)

- Fixes https://github.com/vitessio/vitess/issues/10880

## Checklist

### DIFF
--- a/go/cmd/vtcombo/main.go
+++ b/go/cmd/vtcombo/main.go
@@ -307,3 +307,8 @@ func (mysqld *vtcomboMysqld) StartReplicationUntilAfter(ctx context.Context, pos
 func (mysqld *vtcomboMysqld) StopReplication(hookExtraEnv map[string]string) error {
 	return nil
 }
+
+// SetSemiSyncEnabled implements the MysqlDaemon interface
+func (mysqld *vtcomboMysqld) SetSemiSyncEnabled(source, replica bool) error {
+	return nil
+}
--- a/go/cmd/vtcombo/main.go
+++ b/go/cmd/vtcombo/main.go
@@ -28,15 +28,12 @@ import (
 	"strings"
 	"time"
 
-	"vitess.io/vitess/go/vt/env"
-	"vitess.io/vitess/go/vt/vtgate/planbuilder/plancontext"
-
-	"vitess.io/vitess/go/vt/vttest"
-
 	"google.golang.org/protobuf/proto"
 
 	"vitess.io/vitess/go/exit"
+	"vitess.io/vitess/go/mysql"
 	"vitess.io/vitess/go/vt/dbconfigs"
+	"vitess.io/vitess/go/vt/env"
 	"vitess.io/vitess/go/vt/log"
 	"vitess.io/vitess/go/vt/logutil"
 	"vitess.io/vitess/go/vt/mysqlctl"
@@ -48,7 +45,9 @@ import (
 	"vitess.io/vitess/go/vt/vtcombo"
 	"vitess.io/vitess/go/vt/vtctld"
 	"vitess.io/vitess/go/vt/vtgate"
+	"vitess.io/vitess/go/vt/vtgate/planbuilder/plancontext"
 	"vitess.io/vitess/go/vt/vttablet/tabletserver/tabletenv"
+	"vitess.io/vitess/go/vt/vttest"
 	"vitess.io/vitess/go/vt/wrangler"
 
 	topodatapb "vitess.io/vitess/go/vt/proto/topodata"
@@ -160,10 +159,10 @@ func main() {
 	servenv.Init()
 	tabletenv.Init()
 
-	var mysqld *mysqlctl.Mysqld
+	mysqld := &vtcomboMysqld{}
 	var cnf *mysqlctl.Mycnf
 	if *startMysql {
-		mysqld, cnf = startMysqld(1)
+		mysqld.Mysqld, cnf = startMysqld(1)
 		servenv.OnClose(func() {
 			mysqld.Shutdown(context.TODO(), cnf, true)
 		})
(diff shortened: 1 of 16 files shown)


go/vt/vttablet/tabletmanager/restore.go
```diff
@@ -499,52 +504,13 @@ func (tm *TabletManager) startReplication(ctx context.Context, pos mysql.Positio
...
-	ti, err := tm.TopoServer.GetTablet(ctx, si.PrimaryAlias)
-	if err != nil {
-		return vterrors.Wrapf(err, "Cannot read primary tablet %v", si.PrimaryAlias)
-	}
-
-	// If using semi-sync, we need to enable it before connecting to primary.
-	if err := tm.fixSemiSync(tabletType, SemiSyncActionNone); err != nil {
+	primary, err := tm.initializeReplication(ctx, tabletType)
+	// If we ran into an error while initializing replication, then there is no point in waiting for catch-up.
+	// Also, if there is primary tablet in the shard, we don't need to proceed further.
```
mattlord, line 509: ... there is *no* primary tablet ...

go/vt/vttablet/tabletmanager/tm_init.go
```diff
@@ -839,3 +841,59 @@ func (tm *TabletManager) hookExtraEnv() map[string]string {
...
+	if topoproto.TabletAliasEqual(si.PrimaryAlias, tablet.Alias) {
+		// We used to be the primary before we got restarted,
+		// and no other primary has been elected in the meantime.
+		// There isn't anything to do here either.
+		log.Warningf("Cannot start replication during initialization: primary in shard record still points to this tablet.")
+		return nil, nil
+	}
+	currentPrimary, err := tm.TopoServer.GetTablet(ctx, si.PrimaryAlias)
+	if err != nil {
+		return nil, vterrors.Wrapf(err, "Cannot read primary tablet %v", si.PrimaryAlias)
```
mattlord, line 874: Nit, but we shouldn't capitalize the first word in error messages (as they get embedded) -- applies to the subsequent two errors as well.

go/vt/vttablet/tabletmanager/tm_init.go
```diff
@@ -839,3 +841,59 @@ func (tm *TabletManager) hookExtraEnv() map[string]string {
...
+	}
+
+	durabilityName, err := tm.TopoServer.GetKeyspaceDurability(ctx, tablet.Keyspace)
+	if err != nil {
+		return nil, vterrors.Wrapf(err, "Cannot read keyspace durability %v", tablet.Keyspace)
```
mattlord, line 879: `durability policy` to be more clear and consistent.

mattlord (approved):
LGTM! I had a few very minor (non functional) comments/nits. Thank you!

deepthi (approved):
The changes look fine. This is more of a general question.
- When a tablet starts up, if there is no primary for the shard, we don't setup replication, and we also skip semi-sync handling.
- This is fine because whenever we do elect a primary, we will fix the replication (including semi-sync) on all the replicas. Correct? I want to get this written down for future reference :)

GuptaManan100:
@deepthi Yes exactly. Whenever a primary is elected, it will be by PlannedReparentShard or EmergencyReparentShard, both of those will fix replication on all the tablets too.

go/vt/vttablet/tabletmanager/restore.go
GuptaManan100: Fixed

go/vt/vttablet/tabletmanager/tm_init.go
GuptaManan100: I have fixed the error messages, but just asking for my knowledge. What does it mean for an error message to be embedded?

go/vt/vttablet/tabletmanager/tm_init.go
GuptaManan100: Fixed

go/cmd/vtcombo/main.go
```diff
@@ -275,3 +274,36 @@ func main() {
 	})
 	servenv.RunDefault()
 }
+
+// vtcomboMysqld is a wrapper on top of mysqlctl.Mysqld.
+// We need this wrapper because vtcombo runs with a single MySQL instance
+// which all the tablets connect to. (replica, primary, all). This means that we shouldn't
+// be trying to run any replication related commands on it, otherwise they fail.
```
GuptaManan100, line 281: While looking at the test failures, I found a couple using vtcombo and discovered that it was not starting at all.
On further investigation, I found that vtcombo runs with 1 MySQL instance so running replication commands makes it fail.
Vtcombo started failing in initialization because we are now trying to set replication source on startup in vttablets.

The problem has been resolved by adding a wrapper on mysqld to ignore replication-related rpcs.

This change in my opinion, is large enough to require a new review.

go/vt/vttablet/tabletmanager/tm_init.go
```diff
@@ -403,12 +404,16 @@ func (tm *TabletManager) Start(tablet *topodatapb.Tablet, healthCheckInterval ti
...
 		// If restore was triggered, it will take care
-		// of updating the tablet state.
+		// of updating the tablet state and initializing replication.
 		return nil
 	}
 
+	// We should be re-read the tablet from tabletManager and use the type specified there.
+	// We shouldn't use the base tablet type directly, since the type could have changed to PRIMARY
+	// earlier in tm.checkPrimaryShip code.
+	_, err = tm.initializeReplication(ctx, tm.Tablet().Type)
```
GuptaManan100, line 414: Also I found a bug with the help of another test. This change is also worth noting.

deepthi (approved):
LGTM

PR #10883, opened 2022-07-28
Separate function for creating bind variables

dbussink:
This doesn't depend on state in the normalizer, so it can be an independent function.

Also makes it public since it can be useful for library style usage of Vitess for example.

## Checklist


vmg (approved):
:+1: Using Vitess as a library is cool :+1:

systay (approved):
me gusta

PR #10887, opened 2022-07-29
check for nil earlier

systay:
## Description
Very small fix. We can check for nil even earlier

## Related Issue(s)
#10879


deepthi (approved):

deepthi:
Not sure what's up with the "Pull Request labels" workflow. It seems to have run twice at the exact same time, one of them was cancelled, and here it shows up as still `Expected`.

PR #10889, opened 2022-07-29
[vtcombo] Delete `flag.Set` call on non-existent flag

### LINKED ISSUE #10888: Remove outdated `flag.Set` from `package go/vt/vtcombo`

`package vtcombo` contains one instance of `flag.[A-Z]`, namely:

```
$ git grep -E "\bflag\.[A-Z]" -- go/vt/vtcombo/*.go
        flag.Set("debug-url-prefix", fmt.Sprintf("/debug-%d", uid))
```

This flag does not seem to exist anywhere in the codebase whatsoever, so this call is _probably_ returning a non-`nil` error (the function is of type `Set(flag, value string) error`, but we ignore the return value).

ajm188:
## Description

`git grep` doesn't turn up a definition for this flag anywhere. As I describe in the issue, this is almost certainly returning a non-nil error, which we also are ignoring. 

## Related Issue(s)

Fixes #10888.

## Checklist

### DIFF
--- a/go/vt/vtcombo/tablet_map.go
+++ b/go/vt/vtcombo/tablet_map.go
@@ -18,7 +18,6 @@ package vtcombo
 
 import (
 	"context"
-	"flag"
 	"fmt"
 	"os"
 	"path"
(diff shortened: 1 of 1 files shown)


deepthi (approved):
Looks like the flag was deleted a long time ago (sometime between 2.0 & 3.0).
👍

PR #10902, opened 2022-08-01
BugFix: vtctld panic with `enable_realtime_stats`

### LINKED ISSUE #10901: Bug Report: vtctld crashes with `enable_realtime_stats` flag
### Overview of the Issue

Starting vtctld with `enable_realtime_stats` flag causes it to crash with the following error message - 
```
panic: non-positive interval for NewTicker

goroutine 69 [running]:
time.NewTicker(0x323d070?)
        time/tick.go:24 +0x10f
vitess.io/vitess/go/vt/discovery.(*TopologyWatcher).Start.func1(0xc00020f2c0)
        vitess.io/vitess/go/vt/discovery/topology_watcher.go:130 +0x77
created by vitess.io/vitess/go/vt/discovery.(*TopologyWatcher).Start
        vitess.io/vitess/go/vt/discovery/topology_watcher.go:128 +0x75
```

### Reproduction Steps

Run vtctld with `enable_realtime_stats` flag

### Binary Version

```sh
main. Release-14.0 and before don't seem to be affected.
```

```sh
All
```

### Log Fragments

_No response_

GuptaManan100:
## Description

This PR fixes the panic as described in https://github.com/vitessio/vitess/issues/10901.

The problem was traced back to https://github.com/vitessio/vitess/pull/10863. In this refactor of flags, we accidentally removed discovery flags for being registered for vtctld, which caused the default value of the flag not being used while starting the timer, causing it to panic.

## Related Issue(s)

Fixes #10901 

## Checklist

### DIFF
--- a/go/vt/discovery/healthcheck.go
+++ b/go/vt/discovery/healthcheck.go
@@ -66,7 +66,7 @@ var (
 	healthcheckOnce           sync.Once
 
 	// TabletURLTemplateString is a flag to generate URLs for the tablets that vtgate discovers.
-	TabletURLTemplateString string
+	TabletURLTemplateString = "http://{{.GetTabletHostPort}}"
 	tabletURLTemplate       *template.Template
 
 	// AllowedTabletTypes is the list of allowed tablet types. e.g. {PRIMARY, REPLICA}.
@@ -80,13 +80,13 @@ var (
 	tabletFilters []string
 
 	// refreshInterval is the interval at which healthcheck refreshes its list of tablets from topo.
-	refreshInterval time.Duration
+	refreshInterval = 1 * time.Minute
 
 	// refreshKnownTablets tells us whether to process all tablets or only new tablets.
-	refreshKnownTablets bool
+	refreshKnownTablets = true
 
 	// topoReadConcurrency tells us how many topo reads are allowed in parallel.
-	topoReadConcurrency int
+	topoReadConcurrency = 32
 
 	// How much to sleep between each check.
 	waitAvailableTabletInterval = 100 * time.Millisecond
--- a/go/vt/discovery/healthcheck.go
+++ b/go/vt/discovery/healthcheck.go
@@ -149,6 +149,7 @@ func ParseTabletURLTemplateFromFlag() {
 func init() {
 	servenv.OnParseFor("vtgate", registerDiscoveryFlags)
 	servenv.OnParseFor("vtcombo", registerDiscoveryFlags)
+	servenv.OnParseFor("vtctld", registerDiscoveryFlags)
 }
 
 func registerDiscoveryFlags(fs *pflag.FlagSet) {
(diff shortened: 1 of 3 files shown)


go/vt/discovery/healthcheck.go
```diff
@@ -312,6 +313,7 @@ func NewHealthCheck(ctx context.Context, retryDelay, healthCheckTimeout time.Dur
 	if cellsToWatch == "" {
 		cells = append(cells, localCell)
 	}
+	log.Errorf("Refresh Interval - %v", refreshInterval)
```
ajm188, line 316: nit: we should delete this

go/vt/discovery/healthcheck.go
```diff
@@ -149,6 +149,7 @@ func ParseTabletURLTemplateFromFlag() {
 func init() {
 	servenv.OnParseFor("vtgate", registerDiscoveryFlags)
 	servenv.OnParseFor("vtcombo", registerDiscoveryFlags)
+	servenv.OnParseFor("vtctld", registerDiscoveryFlags)
```
ajm188, line 152: do we need `vtctl` as well, or _just_ `vtctld` (i assume just `vtctld`, but want to make sure we've thought about it)

ajm188 (changes requested):

go/vt/discovery/healthcheck.go
GuptaManan100: Oh yes, for sure. Sorry I was using that to debug what was going on!

go/vt/discovery/healthcheck.go
GuptaManan100: I don't think `vtctl` requires it, but there is something in `go/vt/throttler/demo/throttler_demo.go` which is also using the healthcheck. I don't if we need to add it 🤷‍♂️

GuptaManan100: `Vttablet` too

deepthi (commented):
Thank you for catching this.
Here's what I think we need to do
- Add the flags for `vtctld` (this PR already covers this) + a test for `vtctld`
- `throttler_demo` is a separate binary. Instead of adding flags to it, we should make it use the defaults
- `vttablet`'s tx_throttler should use default values (similar to how it already calls `NewHealthCheck`)
- `vtexplain` uses `FakeHealthCheck` so I don't think it needs to change
- I'm unsure about `vtctl`, we'll need to dig deeper to figure this out.

deepthi:
When I did #10863, @ajm188 and I went back and forth about setting default values along with the var declaration versus setting them in tests. I'm glad that the way we did it exposed this issue, but it is time to consider setting defaults for the `throttler_demo` and `vttablet` usages to work.

go/vt/discovery/healthcheck.go
GuptaManan100: Fixed

GuptaManan100 (commented):
I have added the test and added the defaults for the values.

deepthi (approved):

ajm188 (approved):

PR #10903, opened 2022-08-01
evalengine: expose Filter operations

vmg:
## Description

This PR includes a few small fixes to the `evalengine` and some new exports that will make working with the engine as a library less painful.

## Checklist

### DIFF
--- a/go/vt/vtgate/engine/filter_test.go
+++ b/go/vt/vtgate/engine/filter_test.go
@@ -25,17 +25,40 @@ import (
 
 	"vitess.io/vitess/go/mysql/collations"
 	"vitess.io/vitess/go/sqltypes"
+	"vitess.io/vitess/go/vt/sqlparser"
 	"vitess.io/vitess/go/vt/vtgate/evalengine"
 )
 
-func TestFilterPass(t *testing.T) {
-	predicate := &evalengine.ComparisonExpr{
-		BinaryExpr: evalengine.BinaryExpr{
-			Left:  evalengine.NewColumn(0, defaultCollation()),
-			Right: evalengine.NewColumn(1, defaultCollation()),
-		},
-		Op: evalengine.CompareGT{},
+type dummyTranslator struct{}
+
+func (d dummyTranslator) ColumnLookup(col *sqlparser.ColName) (int, error) {
+	switch col.Name.String() {
+	case "left":
+		return 0, nil
+	case "right":
+		return 1, nil
+	default:
+		panic("unexpected column name")
 	}
+}
+
+func (d dummyTranslator) CollationForExpr(_ sqlparser.Expr) collations.ID {
+	return collationEnv.LookupByName("utf8mb4_bin").ID()
+}
+
+func (d dummyTranslator) DefaultCollation() collations.ID {
+	return collationEnv.LookupByName("utf8mb4_bin").ID()
+}
+
+func TestFilterPass(t *testing.T) {
+	predicate := &sqlparser.ComparisonExpr{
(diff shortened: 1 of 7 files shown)


go/vt/vtgate/engine/filter_test.go
```diff
@@ -25,18 +25,41 @@ import (
...
-	predicate := &evalengine.ComparisonExpr{
-		BinaryExpr: evalengine.BinaryExpr{
-			Left:  evalengine.NewColumn(0, defaultCollation()),
-			Right: evalengine.NewColumn(1, defaultCollation()),
-		},
-		Op: evalengine.CompareGT{},
+	predicate := &sqlparser.ComparisonExpr{
+		Operator: sqlparser.GreaterThanOp,
+		Left:     sqlparser.NewColName("left"),
+		Right:    sqlparser.NewColName("right"),
```
vmg, line 57: @harshit-gangal: I refactored this for you. You initially exported `CompareGT` to be able to test this, which is not a good pattern. The reason none of the ops are exported is because their implementation is unstable. The safe way to test evalengine expressions is to simply have a `sqlparser` AST and convert it to an `evalengine.Expr` with `evalengine.Translate`. That way our tests remain stable even when the `evalengine` translator changes its behavior.

vmg:
cc @dbussink for review

go/vt/vtgate/engine/filter_test.go
```diff
@@ -25,18 +25,41 @@ import (
...
+		panic("unexpected column name")
+	}
+}
+
+func (d dummyTranslator) CollationForExpr(_ sqlparser.Expr) collations.ID {
+	return collationEnv.LookupByName("utf8mb4_bin").ID()
+}
+
+func (d dummyTranslator) DefaultCollation() collations.ID {
+	return collationEnv.LookupByName("utf8mb4_bin").ID()
```
dbussink, line 50: Any specific reason this is using the binary collation and not return whatever the default collation is? Of course it's a dummy, but was wondering if there's more behind it or not then.

go/vt/vtgate/engine/filter_test.go
```diff
@@ -96,19 +120,11 @@ func TestFilterMixedFail(t *testing.T) {
...
 			}
 			_, err := filter.TryExecute(context.Background(), &noopVCursor{}, nil, false)
 			require.EqualError(t, err, tc.expErr)
 		})
 	}
 }
-
-func defaultCollation() collations.TypedCollation {
-	return collations.TypedCollation{
-		Collation:    collationEnv.LookupByName("utf8mb4_bin").ID(),
```
dbussink, line 110: I guess it was moved from here.

vmg: That's right, the test was using the `utf8mb4_bin` collation so we've just ported that.

go/vt/vtgate/evalengine/eval_result.go
```diff
@@ -475,6 +475,84 @@ func (er *EvalResult) nullSafeHashcode() (HashCode, error) {
 	}
 }
 
+// NullsafeHashCodeInPlace behaves like NullsafeHashCode but the type coercion is performed
+// in-place for performance reasons. Eventually this method will replace the old implementation.
+func NullsafeHashCodeInPlace(v sqltypes.Value, collation collations.ID, typ sqltypes.Type) (HashCode, error) {
```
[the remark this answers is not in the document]
vmg: Yeah! It's a very useful API! Anybody could use it anytime!

dbussink (approved):

PR #10906, opened 2022-08-01
Improve topo handling and add additional functionality

dbussink:
The change here adds specific additional functionality to the topo. It adds `WatchRecursive` which allows for watching a specific prefix or directory which means we can more efficiently monitor changes in the topo without having to busy loop, or without having to setup multiple watchers.

It also adds a more efficient way to track leadership changes which under the hood work in a similar way. 

Lastly, this series of changes starts with a refactor on `Watch` to make it more idiomatic Go and better match the added `WatchRecursive`. 

The changes are best reviewed commit by commit here as they build it up incrementally. 

Nothing immediately starts using these new methods here, that is something to be added separately. 

## Checklist

### DIFF
--- a/go/vt/topo/conn.go
+++ b/go/vt/topo/conn.go
@@ -386,4 +386,14 @@ type LeaderParticipation interface {
 	// GetCurrentLeaderID returns the current primary id.
 	// This may not work after Stop has been called.
 	GetCurrentLeaderID(ctx context.Context) (string, error)
+
+	// WaitForNewLeader allows for nodes to wait until a leadership
+	// election cycle completes and to get subsequent updates of
+	// leadership changes. This way logic that needs to know if leadership
+	// changes also if we're not the leader ourselves doesn't need to
+	// poll for leadership status.
+	//
+	// For topo implementation that have this, it can be used more
+	// efficiently than needing a busy wait loop.
+	WaitForNewLeader(ctx context.Context) (<-chan string, error)
 }
--- a/go/vt/topo/conn.go
+++ b/go/vt/topo/conn.go
@@ -17,9 +17,8 @@ limitations under the License.
 package topo
 
 import (
-	"sort"
-
 	"context"
+	"sort"
 )
 
 // Conn defines the interface that must be implemented by topology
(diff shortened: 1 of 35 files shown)


vmg (approved):
To add to @dbussink's explanation, these are changes that we've already tested internally with more stringent tests than the ones we're performing in Vitess right now, including running topology servers in end2end and integration tests and doing a Goroutine check at the end of the test to ensure that the whole server has exited cleanly.

So far so good :ok_hand:

go/vt/topo/k8stopo/file.go
```diff
@@ -243,11 +242,16 @@ func (s *Server) List(ctx context.Context, filePathPrefix string) ([]topo.KVInfo
...
+		if strings.HasPrefix(node.Data.Key, rootPrefix) {
+			out, err := unpackValue([]byte(node.Data.Value))
+			if err != nil {
+				return results, convertError(err, node.Data.Key)
+			}
 			results = append(results, topo.KVInfo{
 				Key:     []byte(node.Data.Key),
-				Value:   []byte(node.Data.Value),
+				Value:   out,
 				Version: KubernetesVersion(node.GetResourceVersion()),
```
dbussink, line 255: @mattlord FYI since it looks like your name is on the `git blame`, but there were 3 separate bugs here in the K8s topo `List` implementation. It wasn't checking against the correct path prefix (the `root` was ignored), it was checking against the `.Value` instead of the `.Key` and the `.Value` was not decompressed to match that compression is transparent.

This PR added some basic tests for `List` that uncovered all those issues. It does make me wonder if anyone is really using the K8s topo if there's bugs like this in there?

go/vt/topo/conn.go
```diff
@@ -158,7 +153,45 @@ type Conn interface {
 	// being correct quickly, as long as it eventually gets there.
 	//
 	// filePath is a path relative to the root directory of the cell.
-	Watch(ctx context.Context, filePath string) (current *WatchData, changes <-chan *WatchData, cancel CancelFunc)
+	Watch(ctx context.Context, filePath string) (current *WatchData, changes <-chan *WatchData, err error)
+
```
GuptaManan100, line 157: This seems like a breaking change requiring release-notes changes. My understanding is that the users can have their own implementation for the topo server and therefore any change to the interface is breaking.

GuptaManan100 (commented):
This is the only comment from my side.

go/vt/topo/k8stopo/file.go
mattlord: Good catch, thank you! I'm also not sure if anyone is using it.

deepthi: not sure either. It was added by @carsonoid who may be able to tell us. or @derekperkins

go/vt/topo/conn.go
deepthi: I don't think we need to worry about this. Unlike vindexes, we don't expect custom topo server implementations.
The interface is exported because it has to be, but that doesn't mean that every exported interface is part of the "public" interface for the vitess codebase.

deepthi (approved):
Very nice PR with lots of good fixes.
- Changes to Watch should be fine.
- The quibble with `WatchRecursive` is that it is currently only implemented for etcd. It's good that you documented how it can be implemented for the other flavors. Since this is not yet being used in Vitess, we'll simply need to get the other flavors implemented whenever we want to start using `WatchRecursive` (for example in vtgate's healthcheck)
- The one remaining question is regarding `List`. Currently it is a strict prefix check, so given `/toplevel/nested/myfile`, List("/top") matches and returns it versus assuming that `top` is a "directory". Looking at the comment on the interface, that seems intentional, so we should leave it as-is unless @mattlord thinks we should change it.

deepthi:
I believe the backup test failure has been fixed via #10895. Merging..

Issue #10660, opened 2022-07-08
Bug Report: Stream Consolidations do not work by default, unclear why to the user


aquarapid:
### Overview of the Issue

`vttablet` has the query consolidator on by default (for historical reasons, the default can be debated, but that is for another issue).  When we added the stream consolidator (for streaming or `workload=olap` queries), we added the sizing configuration variables to `tabletenv/config.go`;  but did not add flags for `vttablet`.  In addition the defaults for the streaming config variables are `0`, which leads to the (streaming) consolidator not getting created at tablet startup.

The result is that you will not get any streaming consolidations unless:
  * You use the YAML tablet configuration (not that common)
  * and you set the relevant stream consolidation sizing vars (`consolidatorStreamQuerySize` and `consolidatorStreamTotalSize`) both to something appropriate, non-0

Additionally, we don't log anything at startup to indicate the state of the stream consolidator (or the consolidator in general).  This made this problem challenging to debug.

### Reproduction Steps

 * Deploy an unsharded keyspace with default-ish `vttablet` parameters
 * Create a test table (I used `sysbench ... prepare` to create a table called `sbtest1` and insert 1000 rows in it)
 * Use a threaded client to set `workload=olap` and run `select * from sbtest1` from multiple threads.
 * Capture the querylog for the tablet using curl from the `/debug/querylog` endpoint on the tablet HTTP interface
 * Observe that from the query log there are no lines with `consolidator` as the query source, indicating that no consolidations are occurring.

Now:
  * either restart `vttablet` with YAML config that includes the missing streaming variables;  or patch `tabletenv/config.go:Init()` to set them by default
  * Re-run the test above
  * You will observe from the querylog that consolidations are now occurring

Test client I used:

```
import mysql.connector
from threading import Thread

def run():
    cnx = mysql.connector.Connect(host='localhost', port=15306,
        database='unsharded', user='user', password="password",
        charset='utf8', get_warnings=True, autocommit=True)
    curprep = cnx.cursor(prepared=False)
    stmt = 'set workload=olap;'
    curprep.execute(stmt, ())
    stmt = 'select * from sbtest1;'
    curprep.execute(stmt, ())

threads = []
for j in range(0,100):
    print(j)
    for i in range(0,50):
        t = Thread(target=run)
        threads.append(t)
        t.start()

    for t in threads:
        t.join()
```

### Binary Version

```sh
release-14.0 ; but I believe this problem exists since we merged the stream consolidator.
```

[...]

PR #10907, opened 2022-08-01
add vttablet cli flags for stream consolidator

maxenglander:
## Description

When stream consolidation was added to VTTablet, fields were included in
the YAML configuration interface to set stream consolidator query/total
size. The fields are were set to non-zero by default.

However, equivalent flags were not added to the CLI configuration
interface, which resulted in the default values for stream consolidator
query/total size set to zero, effectively disabling the stream
consolidator, with no way for users to enable it via CLI flags.

This change adds `--consolidator-stream-query-size` and
`--consolidator-stream-query-total` flags to the VTTablet CLI
configuration interface. The default values for these flags are set to
the same values as the default values for the YAML configuration fields.

## Related Issue(s)

Addresses #10660

## Checklist

### DIFF
--- a/go/vt/vttablet/tabletserver/query_engine.go
+++ b/go/vt/vttablet/tabletserver/query_engine.go
@@ -178,6 +178,8 @@ func NewQueryEngine(env tabletenv.Env, se *schema.Engine) *QueryEngine {
 	qe.consolidator = sync2.NewConsolidator()
 	if config.ConsolidatorStreamTotalSize > 0 && config.ConsolidatorStreamQuerySize > 0 {
 		qe.streamConsolidator = NewStreamConsolidator(config.ConsolidatorStreamTotalSize, config.ConsolidatorStreamQuerySize, returnStreamResult)
+	} else {
+		log.Info("Disabling stream consolidator.")
 	}
 	qe.txSerializer = txserializer.New(env)
 
--- a/go/vt/vttablet/tabletserver/query_engine.go
+++ b/go/vt/vttablet/tabletserver/query_engine.go
@@ -179,7 +179,7 @@ func NewQueryEngine(env tabletenv.Env, se *schema.Engine) *QueryEngine {
 	if config.ConsolidatorStreamTotalSize > 0 && config.ConsolidatorStreamQuerySize > 0 {
 		qe.streamConsolidator = NewStreamConsolidator(config.ConsolidatorStreamTotalSize, config.ConsolidatorStreamQuerySize, returnStreamResult)
 	} else {
-		log.Info("Disabling stream consolidator.")
+		log.Info("Stream consolidator is not enabled.")
 	}
 	qe.txSerializer = txserializer.New(env)
 
--- a/go/vt/vttablet/tabletserver/query_engine.go
+++ b/go/vt/vttablet/tabletserver/query_engine.go
@@ -177,6 +177,8 @@ func NewQueryEngine(env tabletenv.Env, se *schema.Engine) *QueryEngine {
 	qe.consolidatorMode.Set(config.Consolidator)
 	qe.consolidator = sync2.NewConsolidator()
 	if config.ConsolidatorStreamTotalSize > 0 && config.ConsolidatorStreamQuerySize > 0 {
+		log.Info("Stream consolidator is enabled with query size set to %d and total size set to %d.",
+			config.ConsolidatorStreamQuerySize, config.ConsolidatorStreamTotalSize)
 		qe.streamConsolidator = NewStreamConsolidator(config.ConsolidatorStreamTotalSize, config.ConsolidatorStreamQuerySize, returnStreamResult)
 	} else {
 		log.Info("Stream consolidator is not enabled.")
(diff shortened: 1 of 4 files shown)


maxenglander:
Here's the output from `/debug/querylog` with these changes and the following flags set via CLI:

 * `--consolidator_stream_query_size=2097152`
 * `--consolidator_stream_total_size: "134217728`

```
StreamExecute   10.1.2.113:41436:/queryservice.Query/StreamExecute(gRPC)  gRPC    'maxenglander'  'maxenglander'  2022-08-01 18:14:39.736985        2022-08-01 18:14:39.8291270.092142        SelectStream    "select id, text1 from sbtest1"   map[]   0       ""      consolidator    0.000000 0.000000 0       0       0       "(errno 2013) due to context canceled, elapsed time: 50.713625ms, killing query ID 88 (CallerID: maxenglander)"
```

go/vt/vttablet/tabletserver/tabletenv/config.go
```diff
@@ -139,6 +139,8 @@ func init() {
 	flag.BoolVar(&currentConfig.EnforceStrictTransTables, "enforce_strict_trans_tables", defaultConfig.EnforceStrictTransTables, "If true, vttablet requires MySQL to run with STRICT_TRANS_TABLES or STRICT_ALL_TABLES on. It is recommended to not turn this flag off. Otherwise MySQL may alter your supplied values before saving them to the database.")
 	flagutil.DualFormatBoolVar(&enableConsolidator, "enable_consolidator", true, "This option enables the query consolidator.")
 	flagutil.DualFormatBoolVar(&enableConsolidatorReplicas, "enable_consolidator_replicas", false, "This option enables the query consolidator only on replicas.")
+	flagutil.DualFormatInt64Var(&currentConfig.ConsolidatorStreamQuerySize, "consolidator_stream_query_size", defaultConfig.ConsolidatorStreamQuerySize, "Configure the stream consolidator query size. Setting to 0 disables the stream consolidator.")
+	flagutil.DualFormatInt64Var(&currentConfig.ConsolidatorStreamTotalSize, "consolidator_stream_total_size", defaultConfig.ConsolidatorStreamTotalSize, "Configure the stream consolidator total size. Setting to 0 disables the stream consolidator.")
```
deepthi, line 143: What is the unit for these values? We should document it in the flag help.
Also since we are introducing new flags, let us use just hyphens and not do DualFormat at all.
cc @ajm188 - do we have this guidance on dashes vs underscores written down somewhere?

deepthi (commented):
Thank you :)

go/vt/vttablet/tabletserver/tabletenv/config.go
maxenglander: Hyphenated and documented!

ajm188: we don't, but we should. do you have a suggestion on where that should live? i'm happy to do the writing, i just don't know if in vitessio/vitess, or in enhancements, or on the website is the right place for it to live

deepthi: How about adding a section to the [contribute page](https://vitess.io/docs/15.0/contributing/code-reviews/) on the website?
And optionally, adding it to the [PR template](https://github.com/vitessio/vitess/blob/main/.github/pull_request_template.md) and/or [review template](https://github.com/vitessio/vitess-bot/blob/main/index.js#L9).
It might even be worth adding a new `CLI` section to the review template.

go/vt/vttablet/tabletserver/query_engine.go
```diff
@@ -178,6 +178,8 @@ func NewQueryEngine(env tabletenv.Env, se *schema.Engine) *QueryEngine {
 	qe.consolidator = sync2.NewConsolidator()
 	if config.ConsolidatorStreamTotalSize > 0 && config.ConsolidatorStreamQuerySize > 0 {
 		qe.streamConsolidator = NewStreamConsolidator(config.ConsolidatorStreamTotalSize, config.ConsolidatorStreamQuerySize, returnStreamResult)
```
deepthi, line 180: How about adding a log line here as well? Something like "Enabling stream consolidation with ...."

go/vt/vttablet/tabletserver/query_engine.go
```diff
@@ -178,6 +178,8 @@ func NewQueryEngine(env tabletenv.Env, se *schema.Engine) *QueryEngine {
...
+	} else {
+		log.Info("Disabling stream consolidator.")
```
deepthi, line 182:
```suggestion
		log.Info("Stream consolidator is not enabled.")
```

deepthi (commented):
You will need to fix the endtoend test failure. You can run it locally from `go/flags/endtoend` with just `go test`. You can copy over the new output from running vttablet to the expected output file. I usually just do `vttablet --help 2>&1 > vttablet.txt`. The new flags are present in the help text now, which is what is failing the test.

deepthi:
> Here's the output from `/debug/querylog` with these changes and the following flags set via CLI:
> 
> * `--consolidator_stream_query_size=2097152`
> * `--consolidator_stream_total_size: "134217728`
> 
> ```
> StreamExecute   10.1.2.113:41436:/queryservice.Query/StreamExecute(gRPC)  gRPC    'maxenglander'  'maxenglander'  2022-08-01 18:14:39.736985        2022-08-01 18:14:39.8291270.092142        SelectStream    "select id, text1 from sbtest1"   map[]   0       ""      consolidator    0.000000 0.000000 0       0       0       "(errno 2013) due to context canceled, elapsed time: 50.713625ms, killing query ID 88 (CallerID: maxenglander)"
> ```

Max, can you edit this comment and fix the formatting of the flags?

maxenglander:
@deepthi thanks for all the comments, I think I've addressed them all.

deepthi (approved):

dbussink:
Looks like this PR together with https://github.com/vitessio/vitess/pull/10938#issuecomment-1207260344 broken `main`.

maxenglander:
@dbussink thanks for catching and fixing!

PR #10908, opened 2022-08-01
Latest 14.0

rsajwani:
## Description

We have seen the cache miss metric numbers are typically on the higher side. I looks like we are trying to get plan from cache for SQL queries like "COMMIT", "ROLLBACK" etc. This result in unnecessary increment in cache miss counter.
Also inorder to debug better we are adding PlanCache field in logstats object.

## Checklist

-   [x ] Tests were added or are not required
-   [ x] Documentation was added or is not required


rsajwani:
Created PR with wrong push!!!!

PR #10911, opened 2022-08-01
Port vtorc CLI to servenv and pflag

ajm188:
## Description
Port vtorc CLI to servenv and pflag
    
This is a mea culpa on my part; we missed this one before the v14 ship date (because I was unsure the status of vtorc GA and forgot to ask). Consequently, we have to do some extra work to avoid breaking compatibility.
    
Therefore, I introduce a function to inline transform the args passed by the user on the CLI from those that worked with stdlib flag into those that work with pflag. If there is a diff, we print a warning to the user so they know how to update their invocation, so they don't experience issues when we drop this shim code in v16.

## Related Issue(s)

Closes https://github.com/vitessio/vitess/issues/10913

## Checklist


GuptaManan100 (approved):
There are flags here that we might remove before the next release, but we would still need this change for the flags that we want to keep around like `config`. So overall, LGTM 🥇

PR #10896, opened 2022-07-31
Revert temporary workflow changes made in #10847

mattlord:
## Description

This should not be merged until after the backport is merged: https://github.com/vitessio/vitess/pull/10895

Once that PR is merged then `release-14.0` will also have the backup fixes and the upgrade/downgrade backup tests will be working again with the latest MySQL releases (8.0.30+).

## Related Issue(s)

  - Follow-up to: https://github.com/vitessio/vitess/pull/10847
  - Dependent-on: https://github.com/vitessio/vitess/pull/10895

## Checklist

-   [x] "Backport me!" label has been added if this change should be backported
-   [x] Tests were added or are not required
-   [x] Documentation was added or is not required

PR #10914, opened 2022-08-03
Revert: Revert temporary workflow changes made in #10847

mattlord:
## Description

In https://github.com/vitessio/vitess/pull/10896 we reverted the temporary [upgrade/downgrade backup manual workflow](https://github.com/vitessio/vitess/blob/main/.github/workflows/upgrade_downgrade_test_backups_manual.yml) changes made in https://github.com/vitessio/vitess/pull/10847. 

This was done because the MySQL 8.0.30 backup fixes were merged into the `release-14.0` branch via https://github.com/vitessio/vitess/pull/10895 and given that "v14 had the fix" (though not in an official release) we should no longer need to pin the MySQL version.

Well... I was mistaken about what we downgraded *to* in [this workflow](https://github.com/vitessio/vitess/blob/main/.github/workflows/upgrade_downgrade_test_backups_manual.yml). I thought we built binaries on the previous release branch, in this case `release-14.0`, but instead we checkout the latest release tag which is `v14.0.1` today.

You can see this here: https://github.com/vitessio/vitess/blob/853d88d4eb3928d132d2cb8f717c062f4e9bb73a/.github/workflows/upgrade_downgrade_test_backups_manual.yml#L24-L29

And here: https://github.com/vitessio/vitess/blob/main/tools/get_previous_release.sh

And you can see an example workflow which fails to perform the backup on the downgraded v14.0.1 binaries here: https://github.com/vitessio/vitess/runs/7642110846?check_suite_focus=true

So in this PR we put the temporary workflow changes _*back*_ in place and update the comment in the workflow to reflect that these MySQL version pinning changes cannot be removed until either of the following occur:
1. `v14.0.2` is released 
2. The dev version on main becomes `16.0.0-SNAPSHOT` (meaning the `v15.0.0` release candidate has been cut)

## Related Issue(s)

  - Follow-up to: https://github.com/vitessio/vitess/pull/10847
  - Reverts: https://github.com/vitessio/vitess/pull/10896

## Checklist

### DIFF
--- a/.github/workflows/upgrade_downgrade_test_backups_manual.yml
+++ b/.github/workflows/upgrade_downgrade_test_backups_manual.yml
@@ -10,8 +10,8 @@ concurrency:
 jobs:
   get_previous_release:
     if: always() && github.event_name == 'pull_request'
-    name: Get latest release
-    runs-on: ubuntu-latest
+    name: Get a recent LTS release
+    runs-on: ubuntu-20.04
     outputs:
       previous_release: ${{ steps.output-previous-release-ref.outputs.previous_release_ref }}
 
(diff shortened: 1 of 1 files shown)


deepthi (approved):
Thank you

Issue #10899, opened 2022-08-01
Idea: Online DDL syntax to postpone execution of migrations, and per-shard execution


shlomi-noach:
We hit a use case where we wanted to run a migration on a multisharded keyspace, but first only wanted to run the migration on a single shard, to validate the results, by way of reducing the blast radius. Right now there is no support for this in Vitess other than manually injecting entries into `_vt.schema_migrations`. At the time, we had this idea: https://github.com/vitessio/vitess/pull/7825, but I think it's a wrong approach, because it can lead to discrepancies between shards.

I'd like to suggest this new idea: consider the following command:

```sql
set  @@ddl_strategy='online --postpone-execution`;
alter table t add column i int;
```

We can introduce a `--postpone-execution` flag. With this flag, the migration does not get auto-executed. It is submitted, queued, but never started. Contrast this with the existing `--postpone-completion` flag, where the migration starts but runs forever or until told to complete.

Then, we add the following syntax and variations:

```sql
ALTER VITESS_MIGRATION '<uuid>' EXECUTE;
ALTER VITESS_MIGRATION '<uuid>' EXECUTE SHARD '<shard-name>';
ALTER VITESS_MIGRATION EXECUTE ALL;
```

This is symmetrical to `ALTER VITESS_MIGRATION COMPLETE|CANCEL` but with the addition of a per-shard instruction (which we can later also add to `COMPLETE|CANCEL`).

With this new syntax, we will be able to submit a migration, which is queued on _all shards_, making the queue consistent, but we can then choose to only _start it_ on a single, specific shard. When satisfied, we can then proceed to execute it on all shards.

We take upon ourselves the risk of having an inconsistent schema between the different shards -- this is really most useful to adding/modifying indexing.

This will be a pretty straightforward development; it will work well with all existing scheduling options; it will be possible to mix `--postpone-execution` with `--postpone-completion` for example (though it does sound a little bit over the top to do this mixture).

Thoughts welcome.

deepthi:
This sounds great, and much safer than #7825.
We'll still have put in BOLD ALL-CAPS that people are assuming the risk of inconsistent schema.
@vitessio/query-serving what sort of spurious bug reports might we get from this? And are there things we can do in code to head those off?

shlomi-noach:
> And are there things we can do in code to head those off?

[...]

PR #10915, opened 2022-08-03
Online DDL: --pospone-launch, ALTER VITESS_MIGRATION ... LAUNCH

shlomi-noach:
## Description

Closes https://github.com/vitessio/vitess/issues/10899

This PR introduces a new DDL strategy, and new sntax, that make it possible to:

- submit a migration such that it doesn't _launch_, ie does not begin to execute
- instruct such a migration to _launch_, potentially on a per-shard basis.

In a sense, it is the counterpart for postponed-completion migrations. But where a `--postpone-completion` migration starts running and defers final cutover, the newly introduced `--postpone-launch` migration remains in `queued` state and never starts running.

A migration can be submitted with `--ddl_strategy="online --postpone-launch"`. Such a migration will enter the `queued` state and will remain in that state until instructed otherwise. The way to take it out of `queued` state is by running either:

1. `ALTER VITESS_MIGRATION '<uuid>' LAUNCH`
  This unblocks the specific migration on all shards. At scheduler's discretion, the migration will then become `ready` and then at some stage get executed.
2. `ALTER VITESS_MIGRATION '<uuid>' LAUNCH VITESS_SHARDS 'comma,delimited,names,of,shards'`
  This unblocks the specific migration _on specific shards_. Only shards listed are affected; the rest of the shards ignore the commands. This can be useful in testing a migration on a single shard before applying it everywhere.
> **Warning**: this can cause shards to get out of sync. A migration that adds or removes columns can cause queries to fail as shards become inconsistent, and this is an unsafe operation. Acceptable scenarios are for example adding/removing indexes.
3. `ALTER VITESS_MIGRATION '<uuid>' LAUNCH ALL`
  Unblock all migrations on all shards, that have been submitted with `--postpone-launch`

`--postpone-launch` can be combined with other flags, including `--postpone-completion`, though this combination is likely an overkill.

Tests added.

## Related Issue(s)

- #10899 
- #6926

## Checklist


shlomi-noach:
Ping for review.

go/vt/sqlparser/sql.y
```diff
@@ -313,7 +312,7 @@ func bindVariable(yylex yyLexer, bvar string) {
 %token <str> SEQUENCE MERGE TEMPORARY TEMPTABLE INVOKER SECURITY FIRST AFTER LAST
 
 // Migration tokens
-%token <str> VITESS_MIGRATION CANCEL RETRY COMPLETE CLEANUP THROTTLE UNTHROTTLE EXPIRE RATIO
+%token <str> VITESS_MIGRATION CANCEL RETRY LAUNCH COMPLETE CLEANUP THROTTLE UNTHROTTLE EXPIRE RATIO
```
GuptaManan100, line 315: Could you add LAUNCH to the non_reserved_keywords list as well? Otherwise statements like `create tablet launch (id int)` will stop working.

GuptaManan100 (commented):

go/vt/sqlparser/sql.y
GuptaManan100: Tested it out too. It returns an error like `Code: INVALID_ARGUMENT syntax error at position 20 near 'launch'`

shlomi-noach: Done!

GuptaManan100 (approved):
LGTM!

shlomi-noach:
Keeping open for a bit more, in case anyone has any design comment

harshit-gangal:
looking at it now.

go/vt/sqlparser/sql.y
```diff
@@ -2,8 +2,7 @@
 Copyright 2019 The Vitess Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
-you may not use this file except in compliance with the License.
-You may obtain a copy of the License at
+xYou may obtain a copy of the License at
```
harshit-gangal, line 5: looks like unintentional change, kindly revert.

harshit-gangal:
I might have not followed other PRs but I am not able to understand the use of such command, when would someone want to queue a migration and not launch it? I understand postpone completion but not able to figure out a reason for postpone-launch.

Ultimately the user still have to execute additional command of launch, so why not queue the migration at that point in time only.

harshit-gangal:
I understand the reason behind subset shard DDL execution, can that be handled in another way than using postpone?

shlomi-noach:
The original comment did not tell the full story. The main use case is to be able to test a migration on a single shard, before applying it on all shards. Again the classic example is to e.g. modify an index. So you'd be able to see how long the migration takes, on a single shard, and what the output is like.
The thing is, applying a migration on a single shard risks inconsistency. So the idea is for the migration to at least -exist_ on all shards. If you're happy with the one shard, you can then apply it on all shards, and you don't need to submit new migrations on the other shards. It's the same migration everywhere, just not running on most shards.

Hope this clarifies why we might want this feature. The use case came from an actual user.

go/vt/sqlparser/sql.y
shlomi-noach: fixed

harshit-gangal:
> The original comment did not tell the full story. The main use case is to be able to test a migration on a single shard, before applying it on all shards. Again the classic example is to e.g. modify an index. So you'd be able to see how long the migration takes, on a single shard, and what the output is like. The thing is, applying a migration on a single shard risks inconsistency. So the idea is for the migration to at least -exist_ on all shards. If you're happy with the one shard, you can then apply it on all shards, and you don't need to submit new migrations on the other shards. It's the same migration everywhere, just not running on most shards.
> 
> Hope this clarifies why we might want this feature. The use case came from an actual user.

I think out of different options you would have thought (and I can think), this looks simple to use and understand.

harshit-gangal (approved):

Issue #10916, opened 2022-08-03
Bug Report: Using backups from v13 doesn't work with v14 and vice versa.


GuptaManan100:
### Overview of the Issue

Using the backups created when Vitess is on v14 don't work when Vitess is upgraded to v13.
Using those backups and bringing up the tablets and then running InitShardPrimary times out - 
```
E0801 17:17:28.359690   77435 main.go:42] rpc error: code = Unknown desc = tablet zone1-0000000101 InitReplica failed: rpc error: code = DeadlineExceeded desc = context deadline exceeded;tablet zone1-0000000102 InitReplica failed: rpc error: code = DeadlineExceeded desc = context deadline exceeded
```

On further investigation it was found that the same issue exists with using backups from v13 with v14 binaries.

The cause of the issue is the different structure of the reparent journal on the primary and replica tablets - 
On the primary the structure is - 
```MySQL
mysql> describe _vt.reparent_journal;
+----------------------+------------------+------+-----+---------+-------+
| Field                | Type             | Null | Key | Default | Extra |
+----------------------+------------------+------+-----+---------+-------+
| time_created_ns      | bigint unsigned  | NO   | PRI | NULL    |       |
| action_name          | varbinary(250)   | NO   |     | NULL    |       |
| master_alias         | varbinary(32)    | NO   |     | NULL    |       |
| replication_position | varbinary(64000) | YES  |     | NULL    |       |
+----------------------+------------------+------+-----+---------+-------+
4 rows in set (0.00 sec)
```

And on the replica it is 
```MySQL
mysql> describe _vt.reparent_journal
    -> ;
+----------------------+------------------+------+-----+---------+-------+
| Field                | Type             | Null | Key | Default | Extra |
+----------------------+------------------+------+-----+---------+-------+
| time_created_ns      | bigint unsigned  | NO   | PRI | NULL    |       |
| action_name          | varbinary(250)   | NO   |     | NULL    |       |
| primary_alias        | varbinary(32)    | NO   |     | NULL    |       |
| replication_position | varbinary(64000) | YES  |     | NULL    |       |
+----------------------+------------------+------+-----+---------+-------+
4 rows in set (0.01 sec)
```

This difference between `primary_alias` and `master_alias` causes the select query using in the waiting for the reparent journal to fail 
```MySQL
mysql> SELECT action_name, master_alias, replication_position FROM _vt.reparent_journal WHERE time_created_ns=1659422530611711000;
[...]

PR #10917, opened 2022-08-03
BugFix: Using backups from v14 should work with v13

GuptaManan100:
## Description

This PR fixes the downgrade part of the problem as described in #10916.
After this change using backups created from v14 work fine with binaries of v13.

The proposed fix is to run the ALTER statement after taking the primary position. This allows for the statement to be replicated to the replicas if it succeeds.

## Related Issue(s)

- #10916 

## Checklist


mattlord (approved):

mattlord (approved):
I was able to confirm that this fixes the issue! 🙇 :heart:

GuptaManan100:
The upgrade test is failing since we are running with MySQL 8.0.30. However, the fix is tested manually by both me and @mattlord

PR #10918, opened 2022-08-03
Backport: Online DDL: deal with autogenerated CHECK constraint names (#10638)

shlomi-noach:
Backport of https://github.com/vitessio/vitess/pull/10638

---

## Description

Workaround to https://bugs.mysql.com/bug.php?id=107772. Whe na table is created with an unnamed `CHECK` constraint name, MySQL generates one based on `<tablename>_chk_<number>`. Then, when the table is renamed, MySQL attempts to replace the table name. Vitess uses long table names while cutting-over tables at the end of a migration, e.g. `_vt_HOLD_6ace8bcef73211ea87e9f875a4d24e90_20200915120410`.

Next, as we already append a checksum to a `CHECK` constraint (see https://github.com/vitessio/vitess/pull/10369#issuecomment-1140900778), that name becomes too long, and the `RENAME TABLE` operation can fail.

In this PR we identify and dissipate  the `<tablename>_chk_<number>` naming format, so that it never escalates.

## Related Issue(s)

#10369 
#6926 

## Checklist


deepthi (approved):
Backport itself looks fine. Are the test failures flaky?
I was under the impression that fixes for flakiness for these tests have been ported from main to the release branch. If there are some that haven't, we should port them.

shlomi-noach:
Yeah, we did not port the flaky tests fixes back to 14! Doing so now.

shlomi-noach:
`onlineddl_vrepl` flakiness fix backport is in https://github.com/vitessio/vitess/pull/10928

dbussink (approved):

deepthi:
@GuptaManan100 backups manual is failing on this.
Is the PR just based on a commit prior to when that was fixed?

PR #10921, opened 2022-08-03
Implement date, time and timestamp literals

dbussink:
This implements the `date`, `time` and `timestamp` literals in the parser and ensures that bind variables also can be generated using these literals. 

## Related Issue(s)

Part of https://github.com/vitessio/vitess/issues/8604

## Checklist


dbussink:
@GuptaManan100 This adds a number of shift / reduce conflicts and I was wondering if you think it's possible to fix those or not here.

dbussink:
The other main question I have here is how to deal with values we can't parse. What is the general approach we should take for those?

GuptaManan100 (approved):
The shift-reduce conflicts have been fixed using precedence rules. 
STRING_TYPE_PREFIX_NON_KEYWORD is used to resolve shift-reduce conflicts occuring due to column_name symbol and being able to use keywords like DATE and TIME as prefixes to strings to denote their type. The shift-reduce conflict occurs because after seeing one of these non-reserved keywords, if we see a STRING, then we can either shift to use the STRING typed rule in literal or reduce the non-reserved keyword into column_name and eventually use a rule from simple_expr. The way to fix this conflict is to give shifting higher precedence than reducing.

dbussink:
@GuptaManan100 @systay I've also added logic here to validate the actual literals and tried to match MySQL behavior within reason.

dbussink:
The CI failure here in `endtoend` is unrelated to the changes here, see also https://github.com/vitessio/vitess/pull/10953

go/vt/sqlparser/normalizer.go
```diff
@@ -146,6 +163,11 @@ func (nz *normalizer) convertLiteralDedup(node *Literal, cursor *Cursor) {
 
 // convertLiteral converts an Literal without the dedup.
 func (nz *normalizer) convertLiteral(node *Literal, cursor *Cursor) {
+	err := validateLiteral(node)
```
dbussink, line 166: Tried various approaches, but injecting this in the normalization step seemed the easiest but it needs to be checked then from here and also in `convertLiteralDedup`.

go/vt/sqlparser/parse_date.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	"time"
+
+	vtrpcpb "vitess.io/vitess/go/vt/proto/vtrpc"
+	"vitess.io/vitess/go/vt/vterrors"
+)
+
+var dateFormats = []string{"2006-01-02", "06-01-02", "20060102", "060102"}
+var datetimeFormats = []string{"2006-01-02 15:04:05", "06-01-02 15:04:05", "20060102150405", "060102150405"}
```
dbussink, line 11: These formats are based on https://dev.mysql.com/doc/refman/8.0/en/date-and-time-literals.html. I haven't expanded `TIME` further because that format isn't really compatible with Go at all, especially when multiple days are involved.

The `TIME` type is more a duration in a way than a timestamp, although you can also do comparisons in MySQL with a `DATE` or `DATETIME`. But expanding that is something separate based on need.

go/vt/vtgate/evalengine/evalengine.go
```diff
@@ -235,39 +235,27 @@ func compareNumeric(v1, v2 *EvalResult) (int, error) {
 func parseDate(expr *EvalResult) (t time.Time, err error) {
```
dbussink, line 235: Unified what this function depends on also into the single parsing logic in the `sqlparser` package.

go/vt/vtgate/evalengine/evalengine.go
```diff
@@ -235,39 +235,27 @@ func compareNumeric(v1, v2 *EvalResult) (int, error) {
...
+	t, err = sqlparser.ParseDate(expr.string())
+	if err == nil {
+		return
+	}
+	t, err = sqlparser.ParseDateTime(expr.string())
+	if err == nil {
+		return
 	}
+	t, err = sqlparser.ParseTime(expr.string())
 	return
```
dbussink, line 259: The new logic above should be equivalent to trying things in the loop. The error message depends on the last type checked, but that already was the case anyway.

go/vt/sqlparser/parse_date.go
dbussink: I've added more complete `TIME` parsing in https://github.com/vitessio/vitess/pull/10921/commits/57046c394208f67ffd1ad0b65fea4441929a90b1

GuptaManan100 (approved):
I like this approach 🥇

PR #10924, opened 2022-08-03
add the selected keyspace to LogStats

systay:
## Description
Adds the selected database to `LogStats` logs.

Makes it easier to read the queries if the context for the query is available.

## Checklist

### DIFF
--- a/go/vt/vtgate/logstats/logstats.go
+++ b/go/vt/vtgate/logstats/logstats.go
@@ -38,24 +38,25 @@ import (
 
 // LogStats records the stats for a single vtgate query
 type LogStats struct {
-	Ctx           context.Context
-	Method        string
-	TabletType    string
-	StmtType      string
-	SQL           string
-	BindVariables map[string]*querypb.BindVariable
-	StartTime     time.Time
-	EndTime       time.Time
-	ShardQueries  uint64
-	RowsAffected  uint64
-	RowsReturned  uint64
-	PlanTime      time.Duration
-	ExecuteTime   time.Duration
-	CommitTime    time.Duration
-	Error         error
-	TablesUsed    []string
-	SessionUUID   string
-	CachedPlan    bool
+	Ctx            context.Context
+	Method         string
+	TabletType     string
+	StmtType       string
+	SQL            string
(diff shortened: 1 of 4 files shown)


rafer (approved):

go/vt/vtgate/logstats/logstats.go
```diff
@@ -155,9 +156,9 @@ func (stats *LogStats) Logf(w io.Writer, params url.Values) error {
 	var fmtString string
 	switch *streamlog.QueryLogFormat {
 	case streamlog.QueryLogFormatText:
-		fmtString = "%v\t%v\t%v\t'%v'\t'%v'\t%v\t%v\t%.6f\t%.6f\t%.6f\t%.6f\t%v\t%q\t%v\t%v\t%v\t%q\t%q\t%q\t%q\t%q\t%v\t%v\n"
+		fmtString = "%v\t%v\t%v\t'%v'\t'%v'\t%v\t%v\t%.6f\t%.6f\t%.6f\t%.6f\t%v\t%q\t%v\t%v\t%v\t%q\t%q\t%q\t%q\t%q\t%v\t%v\t\"%v\"\n"
```
GuptaManan100, line 159: Do we need %v in quotes, The other places seem to have them in single quotes or not at all

GuptaManan100 (approved):

GuptaManan100:
@systay You should also rename the PR header to not include `feat`

go/vt/vtgate/logstats/logstats.go
systay: good point

deepthi:
I've addressed the feedback. Note that the output format does not change, so `\"%v\"` is in fact equivalent to `%q` but `%q` is more compact.

go/vt/vtgate/logstats/logstats.go
deepthi: I've addressed the feedback. Note that the output format does not change, so \"%v\" is in fact equivalent to %q but %q is more compact.

frouioui (approved):
Everything looks good to me.

PR #10940, opened 2022-08-04
schemadiff: rich error for unmet view dependencies

shlomi-noach:
## Description

This PR turns the flat `ErrViewDependencyUnresolved` error into a richer `ViewDependencyUnresolvedError`, indicating the identity of the view in question.

Existing tests adapted to new error type.

## Related Issue(s)

- Tracking: https://github.com/vitessio/vitess/issues/10203

## Checklist

### DIFF
--- a/go/vt/schemadiff/errors.go
+++ b/go/vt/schemadiff/errors.go
@@ -14,7 +14,6 @@ var (
 	ErrUnexpectedTableSpec            = errors.New("unexpected table spec")
 	ErrExpectedCreateTable            = errors.New("expected a CREATE TABLE statement")
 	ErrExpectedCreateView             = errors.New("expected a CREATE VIEW statement")
-	ErrViewDependencyUnresolved       = errors.New("views have unresolved/loop dependencies")
 )
 
 type UnsupportedEntityError struct {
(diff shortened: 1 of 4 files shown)


go/vt/schemadiff/errors.go
```diff
@@ -239,3 +238,11 @@ func (e *InvalidColumnInForeignKeyConstraintError) Error() string {
 	return fmt.Sprintf("invalid column %s referenced by foreign key constraint %s in table %s",
 		sqlescape.EscapeID(e.Column), sqlescape.EscapeID(e.Constraint), sqlescape.EscapeID(e.Table))
 }
+
+type ViewDependencyUnresolvedError struct {
+	View string
```
dbussink, line 243: Can we also here have a reference to the entity it tries to refer to? Like if a another table / view is missing, we'd also know the name here as well? 

We could then further improve the error to "view %s references unknown entity %s" or something along those lines? 

And maybe different from having a loop and have a separate error type for that?

shlomi-noach: So the way we detect these dependencies is by creating a connected graph, and then looking at "what's outside the graph". We don't explicitly then check which tables/views are referenced by the left-out views. We could definitely do that.

go/vt/schemadiff/schema.go
```diff
@@ -230,7 +230,13 @@ func (s *Schema) normalize() error {
 		// We have leftover views. This can happen if the schema definition is invalid:
 		// - a view depends on a nonexistent table
 		// - two views have a circular dependency
-		return ErrViewDependencyUnresolved
+		for _, v := range s.views {
+			if _, ok := dependencyLevels[v.Name()]; !ok {
+				// We _know_ that in this iteration, at least one view is found unassugned a dependency level.
```
iheanyi, line 235:
```suggestion
				// We _know_ that in this iteration, at least one view is found unassigned a dependency level.
```

iheanyi (approved):

go/vt/schemadiff/errors.go
shlomi-noach: @dbussink let's iterate on that in a followup PR though, if that's OK

go/vt/schemadiff/schema.go

go/vt/schemadiff/errors.go
dbussink: Yeah, we can improve on a further iteration.

dbussink (approved):

Issue #10819, opened 2022-07-25
Bug Report: Gen4 planner generates inefficient (and potentially incorrect) query when joining multiple tables 


evaccaro:
### Overview of the Issue

In our use case, we have three tables, one of them acting as a join table between the other two tables. The join is performed via unique lookup vindex columns. The V3 planner handles this type of query correctly (sends the full query to the target shard), but the Gen4 planner seems to get confused and breaks up the query into two parts. 

### `vschema.json`

```json
{
  "sharded": true,
  "vindexes": {
    "hash": {
      "type": "hash"
    },
    "issues_id_ks_idx": {
      "type": "lookup_test",
      "owner": "issues"
    },
    "issues_labels_id_ks_idx": {
      "type": "lookup_test",
      "owner": "issues_labels"
    },
    "labels_id_ks_idx": {
      "type": "lookup_test",
      "owner": "labels"
    }
  },
  "tables": {
    "issues": {
      "column_vindexes": [
        {
          "column": "repository_id",
          "name": "hash"
        },
        {
          "column": "id",
          "name": "issues_id_ks_idx"
        }
      ]
    },
    "issues_labels": {
      "column_vindexes": [
        {
          "column": "repository_id",
          "name": "hash"
        },
        {
          "column": "id",
          "name": "issues_labels_id_ks_idx"
        },
        {
          "column": "issue_id",
          "name": "issues_id_ks_idx"
        },
        {
          "column": "label_id",
          "name": "labels_id_ks_idx"
        }
      ]
    },
    "labels": {
      "column_vindexes": [
        {
          "column": "repository_id",
          "name": "hash"
        },
        {
          "column": "id",
          "name": "labels_id_ks_idx"
        }
      ]
    }
  }
}
```

### Query

```sql
SELECT COUNT(*) 
FROM `issues` 
INNER JOIN `issues_labels` ON `issues_labels`.`issue_id` = `issues`.`id` 
INNER JOIN `labels` ON `labels`.`id` = `issues_labels`.`label_id` 
WHERE `issues`.`repository_id` = 143 AND (issues_labels.label_id = '152')
```

### V3 Query Plan

```json
{
    "QueryType": "SELECT",
[...]

PR #10942, opened 2022-08-04
Improve route merging for queries that have conditions on different vindexes, but can be merged via join predicates.

arthurschreiber:
## Description

This pull request imrpoves query route merging for queries that consist of multiple joins (via vindex columns), where some of the joined tables also have additional vindex conditions defined.

An example can be found in https://github.com/vitessio/vitess/issues/10819, and in the test case that was added to `go/vt/vtgate/planbuilder/testdata/select_cases.txt`.

Previously, the Gen4 planner would break the query up into multiple pieces, one for each "direct" vindex condition defined, even though the query could be send to a single shard based on the join conditions between all tables.

For the query from the added test case, the plan generated by Gen4 would look like this:

```json
{
  "QueryType": "SELECT",
  "Original": "SELECT user.id FROM user INNER JOIN music_extra ON user.id = music_extra.user_id INNER JOIN music ON music_extra.user_id = music.user_id WHERE user.id = 123 and music.id = 456",
  "Instructions": {
    "OperatorType": "Join",
    "Variant": "Join",
    "JoinColumnIndexes": "R:0",
    "JoinVars": {
      "music_user_id": 0
    },
    "TableName": "music_`user`, music_extra",
    "Inputs": [
      {
        "OperatorType": "Route",
        "Variant": "EqualUnique",
        "Keyspace": {
          "Name": "user",
          "Sharded": true
        },
        "FieldQuery": "select music.user_id from music where 1 != 1",
        "Query": "select music.user_id from music where music.id = 456",
        "Table": "music",
        "Values": [
          "INT64(456)"
        ],
        "Vindex": "music_user_map"
      },
      {
        "OperatorType": "Route",
        "Variant": "EqualUnique",
        "Keyspace": {
          "Name": "user",
          "Sharded": true
        },
        "FieldQuery": "select `user`.id from `user`, music_extra where 1 != 1",
        "Query": "select `user`.id from `user`, music_extra where `user`.id = 123 and music_extra.user_id = :music_user_id and `user`.id = music_extra.user_id",
        "Table": "`user`, music_extra",
        "Values": [
          ":music_user_id"
        ],
        "Vindex": "user_index"
      }
    ]
  },
  "TablesUsed": [
    "user.music",
    "user.music_extra",
    "user.user"
  ]
}
```

The Gen4 planner would only consider merging `EqualUnique` routes if the routes matched completely, and would immediately give up merging the routes if they didn't. Now, we try to also consider join predicates and merge routes based on them instead.

## Related Issue(s)

Fixes the incorrect routing identified in https://github.com/vitessio/vitess/issues/10819, but does not fix the incorrect query generation (which is a separate issue).

## Checklist

### DIFF
--- a/go/vt/vtgate/planbuilder/physical/route_planning.go
+++ b/go/vt/vtgate/planbuilder/physical/route_planning.go
@@ -644,7 +644,7 @@ func tryMerge(
 			return merger(aRoute, bRoute)
 		}
 	case engine.EqualUnique:
-		// if they are already both being sent to the same shard, we can merge
+		// If the two routes fully match, they can be merged together.
 		if bRoute.RouteOpCode == engine.EqualUnique {
 			aVdx := aRoute.SelectedVindex()
 			bVdx := bRoute.SelectedVindex()
@@ -653,9 +653,12 @@ func tryMerge(
 			if aVdx == bVdx && gen4ValuesEqual(ctx, aExpr, bExpr) {
 				return merger(aRoute, bRoute)
 			}
-			return nil, nil
 		}
+
+		// If the two routes don't match, fall through to the next case and see if we
+		// can merge via join predicates instead.
 		fallthrough
+
 	case engine.Scatter, engine.IN:
 		if len(joinPredicates) == 0 {
 			// If we are doing two Scatters, we have to make sure that the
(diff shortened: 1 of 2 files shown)


systay (approved):

deepthi:
`vtgate_tablet_healthcheck_cache` was fixed in #10961. Totally unrelated, so we can merge this.

PR #10943, opened 2022-08-04
Replicas should be able to heal if replication is not initialised properly

GuptaManan100:
## Description

This PR addresses the issue raised in https://github.com/vitessio/vitess/issues/10955. The way to repair replication once it is stuck as described in https://github.com/vitessio/vitess/issues/10955, is to call `RESET REPLICA` and then trying `START REPLICA` again.

The proposed solution is to always `RESET REPLICA` in `setReplicationSourceLocked` when we change the primary source/port or when `forceStartReplication` is specified. 

This allows both VTOrc and the replication manager to be able to address this issue since they both call `setReplicationSourceLocked` with forceStartReplication. 

## Related Issue(s)

- Fixes https://github.com/vitessio/vitess/issues/10955

## Checklist


deepthi:
Can you create an issue that describes the problem? I think it is important for this particular issue for us to document it.

mattlord (approved):
LGTM! ❤️

deepthi (approved):
Nice work. LGTM

Issue #10945, opened 2022-08-05
OLAP mode still enforces TX timeouts (-queryserver-config-query-timeout)


maxenglander:
(Clone of internal PlanetScale issue.)

Background:

  * For a long time, you could not use `workload=olap` with transactions
  * This changed for Vitess 13.0, see https://github.com/vitessio/vitess/pull/9115

Issue:
  * Since `workload=olap` bypasses the query timeouts (`-queryserver-config-query-timeout`) and also row limits, the natural assumption is that it also bypasses the transaction timeout.
  * This is not the case, e.g. for a tablet where the `-queryserver-config-transaction-timeout` is `10`:
```
mysql> set workload=olap;                   
Query OK, 0 rows affected (0.00 sec)

mysql> begin ; select * from t1; select sleep(35); commit;
Query OK, 0 rows affected (0.00 sec)

+----+------+
| c1 | c2   |
+----+------+
|  3 |    4 |
|  4 |    5 |
|  5 |    6 |
|  6 |    7 |
+----+------+
4 rows in set (0.00 sec)

ERROR 1317 (70100): vttablet: rpc error: code = Canceled desc = (errno 2013) due to context deadline exceeded, elapsed time: 10.000470709s, killing query ID 5463 (CallerID: user)
ERROR 1317 (70100): vttablet: rpc error: code = Aborted desc = transaction 1657804966728853781: ended at 2022-07-14 12:42:04.175 PDT (unlocked closed connection) (CallerID: user)
```
  * Workaround is to have OLAP mode clients not use transactions at all, this could be reasonable in some cases.
  * However, many of the use-cases for OLAP transactions are "streaming-transactional", e.g. Spark or Sqoop both use transactions when streaming its rows out of tables.  It is more challenging for average users to modify Spark or Sqoop to not use transactions.

Note that PSDB is also affected, since we set the `-queryserver-config-transaction-timeout` relatively low (20 secs), vs a high query timeout.

TL;DR:
  * There should be a way of having longer (or infinite?) transaction timeouts in OLAP mode without having to raise `-queryserver-config-transaction-timeout`, which would affect OLTP mode clients adversely.
  * We should also update the documentation accordingly.  Right now it just says:
```
By default, Vitess sets some intentional restrictions on the execution time and number of rows that a query can return. This default workload mode is called OLTP. This can be disabled by setting the workload to OLAP
```

deepthi:
Fixed in #10946

PR #10946, opened 2022-08-05
decouple olap tx timeout from oltp tx timeout

maxenglander:
## Description

Since `workload=olap` bypasses the query timeouts
(`--queryserver-config-query-timeout`) and also row limits, the natural
assumption is that it also bypasses the transaction timeout.

This is not the case, e.g. for a tablet where the
`--queryserver-config-transaction-timeout` is 10.

This PR

 * Adds new CLI flag and YAML field to independently configure TX
   timeouts for OLAP workloads `--queryserver-config-olap-transaction-timeout` with a default value of `0` seconds, disabling OLAP TX timeouts.
 * Decouples TX kill interval from OLTP TX timeout via new CLI flag and
   YAML field `--queryserver-config-transaction-killer-interval` defaulting to `3` seconds.

One subtlety is that the timeout that is applied to the transaction is based on the value of the workload setting at the beginning of the transaction. If the workload is changed mid-transaction, that may change the timeout applied to queries within the transaction, but it won't change the transaction timeout.

## Demo

Using (new) default values), connected to VTGate.

```
mysql> set workload=oltp;

mysql> begin ; select 1 from data limit 1; select sleep(35); commit;

+---+
| 1 |
+---+
| 1 |
+---+
1 row in set (0.01 sec)

ERROR 1317 (70100): target: dst.-.primary: vttablet: (errno 2013) due to context deadline exceeded, elapsed time: 30.00055243s, killing query ID 106 (CallerID: maxenglander)
ERROR 1317 (70100): target: dst.-.primary: vttablet: rpc error: code = Aborted desc = transaction 1659665741215176173: ended at 2022-08-05 02:18:29.360 UTC (unlocked closed connection) (CallerID: maxenglander)
mysql> set workload=olap;

mysql> begin ; select 1 from data limit 1; select sleep(35); commit;

+---+
| 1 |
+---+
| 1 |
+---+
1 row in set (0.01 sec)

+-----------+
| sleep(35) |
+-----------+
|         0 |
+-----------+
1 row in set (35.00 sec)

```

## Breaking changes

Currently OLAP transactions are killed after `--queryserver-config-transaction-timeout` seconds. With this PR, OLAP transactions are killed after `--queryserver-config-olap-transaction-timeout` seconds (default value `0` means transactions are not timed out).

Currently OLTP and OLAP transactions are evaluated for killing every `--queryserver-config-transaction-timeout` seconds divided by `10`. With this PR, OLAP and OLTP transactions are evaluated for killing every `--queryserver-config-transaction-killer-interval` seconds.

## Related Issue(s)

#10945

## Checklist


deepthi:
@maxenglander we'll get this reviewed. Can you add notes to the 15_0_0_summary.md file and resolve the conflicts in the meantime?

go/vt/vttablet/tabletserver/stateful_connection.go
```diff
@@ -51,6 +51,7 @@ type StatefulConnection struct {
 	reservedProps  *Properties
 	tainted        bool
 	enforceTimeout bool
+	timeout        time.Duration
```
harshit-gangal, line 54: not that timeout is stored with the connection, we can store the actual time when it times out than storing the timeout value.
This will simplify the GetOutdated Logic. No need to pass the timeout value there.

go/vt/vttablet/tabletserver/tabletenv/config.go
```diff
@@ -437,6 +470,9 @@ var defaultConfig = TabletConfig{
 		IdleTimeoutSeconds: 30 * 60,
 		MaxWaiters:         5000,
 	},
+	Olap: OlapConfig{
+		TxTimeoutSeconds: 0,
+	},
```
harshit-gangal, line 475: I think default we should keep to 30 seconds.
I would not recommend `0` for any transaction timeout. It would look the rows for infinite time. 
running a query without a timeout outside a transaction is ok as it does not look any row. But, this goes against the general rule of Vitess that it make things move forward by means of timeouts.

go/vt/vttablet/tabletserver/tx_pool.go
```diff
@@ -77,18 +75,18 @@ type (
...
+		ticks:   timer.NewTimer(config.TransactionKillerIntervalSeconds.Get()),
+		limiter: limiter,
+		txStats: env.Exporter().NewTimings("Transactions", "Transaction stats", "operation"),
 	}
 	// Careful: conns also exports name+"xxx" vars,
 	// but we know it doesn't export Timeout.
-	env.Exporter().NewGaugeDurationFunc("TransactionTimeout", "Transaction timeout", axp.transactionTimeout.Get)
+	env.Exporter().NewGaugeDurationFunc("TransactionTimeout", "Transaction timeout", func() time.Duration {
+		return config.TxTimeoutForWorkload(querypb.ExecuteOptions_OLTP)
+	})
```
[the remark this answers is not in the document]
maxenglander: 👍 do you think a separate metric makes sense or is single metric with labels better here? I guess the question comes down to are we comfortable with a breaking change for folks who are aggregating this value.

go/vt/vttablet/tabletserver/tx_pool.go
```diff
@@ -340,15 +346,8 @@ func (tp *TxPool) LogActive() {
...
-func (tp *TxPool) Timeout() time.Duration {
-	return tp.transactionTimeout.Get()
-}
-
-// SetTimeout sets the transaction timeout.
-func (tp *TxPool) SetTimeout(timeout time.Duration) {
-	tp.transactionTimeout.Set(timeout)
-	tp.ticks.SetInterval(timeout / 10)
+func (tp *TxPool) SetTransactionKillerInterval(d time.Duration) {
+	tp.ticks.SetInterval(d)
```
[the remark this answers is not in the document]
maxenglander: That works! I like that.

go/vt/vttablet/tabletserver/tabletenv/config.go
maxenglander: OK sounds good to me. As another upside it will mean fewer breaking changes.

go/vt/vttablet/tabletserver/tx_pool.go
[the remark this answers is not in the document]
harshit-gangal: separate metric for olap transaction timeout, I believe both of those metrics are totally uninteresting as they will emit the same value all the time.

go/vt/vttablet/tabletserver/stateful_connection.go
maxenglander: I tried making some improvements in b908164. I didn't end up storing a pre-calculated timeout time like you suggested, because:

 * Outdated is driven by `pools.Numbered.timeUsed` and `StatefulConnection.timeout` (duration)
 * These values can evolve independently: 
   * `timeUsed` changes when a connection is created or returned to pool
   * `timeout` changes based on `workload`

With each independent change, the number of times we would need to recalculate a `timeoutAt` value increases. Because of that, I felt like it would be easier to continue to store `timeUsed` and `timeout` as separate values, and calculate "is connection outdated?" in just one place. 

However, I did try to get at some of your comment. Instead of passing timeout values to the `pools.Numbered`, I moved `timeUsed` to `StatefulConnection`, so it is stored next to `timeout`. The end result is that `pools.Numbered` and the logic for getting elapsed connections is simpler.

maxenglander:
@deepthi conflicts resolved and release notes added

go/pools/numbered_test.go
```diff
@@ -55,61 +54,17 @@ func TestNumberedGeneral(t *testing.T) {
...
-	p.Unregister(vals[0].(int64), "test")
-
-	// p has 0, 1, and 2
-	if p.Size() != 3 {
-		t.Errorf("want 3, got %v", p.Size())
-	}
-	go func() {
-		p.Unregister(0, "test")
-		p.Unregister(1, "test")
-		p.Unregister(2, "test")
```
harshit-gangal, line 103: are these test cases moved to different place?

go/vt/vttablet/tabletserver/stateful_connection.go
```diff
@@ -51,6 +52,8 @@ type StatefulConnection struct {
...
+	timeUsed       *atomic.Value
```
harshit-gangal, line 56: any reason for adding atomic here? One connection will be accessed only by one thread at a time, this is protected by a higher order GetAndLock method.

go/vt/vttablet/tabletserver/tx/api.go
```diff
@@ -57,6 +57,7 @@ type (
 		Autocommit      bool
 		Conclusion      string
 		LogToFile       bool
+		Timeout         time.Duration
```
harshit-gangal, line 60: how is this timeout different from `StatefulConnection` timeout ?

go/vt/vttablet/tabletserver/stateful_connection.go
harshit-gangal: I agree that we need to keep `timeout`, but I still think we can keep the expiryTime rather than timeUsed, and update the expiryTime based on the `updateTime` bool in `put` call.

go/vt/vttablet/tabletserver/tabletenv/config.go
```diff
@@ -369,6 +377,29 @@ func (c *TabletConfig) Clone() *TabletConfig {
...
 }
 
+func (c *TabletConfig) SetTxTimeoutForWorkload(val time.Duration, workload querypb.ExecuteOptions_Workload) {
+	switch workload {
+	case querypb.ExecuteOptions_OLAP:
+		c.Olap.TxTimeoutSeconds.Set(val)
+	case querypb.ExecuteOptions_OLTP:
+		c.Oltp.TxTimeoutSeconds.Set(val)
+	default:
+	}
```
harshit-gangal, line 387: may be we should add panic here, incase something goes wrong it will be logged.

go/vt/vttablet/tabletserver/tx_pool.go
```diff
@@ -269,7 +277,12 @@ func (tp *TxPool) begin(ctx context.Context, options *querypb.ExecuteOptions, re
 		return "", err
 	}
 
-	conn.txProps = tp.NewTxProps(immediateCaller, effectiveCaller, autocommit)
+	workload := options.GetWorkload()
+	if workload == querypb.ExecuteOptions_UNSPECIFIED {
+		workload = querypb.ExecuteOptions_OLTP
+	}
+	timeout := tp.env.Config().TxTimeoutForWorkload(workload)
```
harshit-gangal, line 284: the default is oltp timeout in `tp.env.Config().TxTimeoutForWorkload` , the check can be removed.

go/vt/vttablet/tabletserver/tabletserver.go
```diff
@@ -1176,8 +1178,11 @@ func (tsv *TabletServer) ReserveExecute(ctx context.Context, target *querypb.Tar
 	timeout := tsv.QueryTimeout.Get()
 	if transactionID != 0 {
 		allowOnShutdown = true
+		// ReserveExecute is for OLAP only, so we can directly fetch the OLAP
+		// TX timeout.
+		txTimeout := tsv.config.TxTimeoutForWorkload(querypb.ExecuteOptions_OLAP)
```
harshit-gangal, line 1183:
```suggestion
		txTimeout := tsv.config.TxTimeoutForWorkload(querypb.ExecuteOptions_OLTP)
```

go/pools/numbered_test.go
maxenglander: They are just deleted. It looked to me like these cases were related to the idle/outdated code in `pools.Numbered`. I deleted that code, so I didn't think these cases were needed. I thought the other remaining tests for `pools.Numbered` seemed good enough to cover the remaining functionality.

go/vt/vttablet/tabletserver/stateful_connection.go
maxenglander: It's because of the transaction killer uses this value and which I think can access connections in use in other threads in order to force close ones that are idle or taking too long.

go/vt/vttablet/tabletserver/tx/api.go
maxenglander: To be honest I'm a bit unclear about this myself. Here's my understanding though:

 * When a stateful connection starts, its timeout may be based on either the `OLTP` or `OLAP` TX timeouts.
 * When a transaction starts, its timeout may be based on either the `OLTP` or `OLAP` TX timeouts, and that timeout may be different than the timeout set on the underlying stateful connection.
 * When the transaction ends, a new `updateTime` is set based on the timeout that was originally set on the stateful connection.

In other words, the timeout associated with the stateful connection persists for the lifetime of that connection, but the timeout associated with a given transaction can change based on the most recent value of `SET workload`.

I'm not really sure if this behavior makes sense. I think the alternative would be, instead of including a `Timeout` with `txProps`, any time a new `Begin()` is called, I could update the `timeout` value on the stateful connection.

maxenglander: The one exception to this seems to be reserved connections, which are re-used across TXs. In 6aba785d1a I added code to just update the timeout on existing reserved connections.

go/vt/vttablet/tabletserver/stateful_connection.go
maxenglander: OK makes sense. I tried this in 6aba785d1a.

go/vt/vttablet/tabletserver/stateful_connection.go
harshit-gangal: If a connection is inUse already we cannot foreclose and only a put can update this value. I believe we do not need to safe guard it. earlier as well, this was not protected by mutex.

maxenglander: OK thanks. I had some statements in the wrong order. Fixing that let me get rid of the atomic value.

go/vt/vttablet/tabletserver/tabletenv/config.go
```diff
@@ -369,6 +377,31 @@ func (c *TabletConfig) Clone() *TabletConfig {
 	return &tc
 }
 
+// Set transaction timeouts. Used in tests only.
```
harshit-gangal, line 380: nit: comment should start with method name `SetTxTimeoutForWorkload`

go/vt/vttablet/tabletserver/tabletenv/config.go
```diff
@@ -369,6 +377,31 @@ func (c *TabletConfig) Clone() *TabletConfig {
...
+	case querypb.ExecuteOptions_OLAP:
+		c.Olap.TxTimeoutSeconds.Set(val)
+	case querypb.ExecuteOptions_OLTP:
+		c.Oltp.TxTimeoutSeconds.Set(val)
+	default:
+		panic(fmt.Sprintf("unsupported workload type: %v", workload))
+	}
+}
+
+// TxTimeout returns the transaction timeout for the given workload type.
```
harshit-gangal, line 392: nit: comment should start with method name `TxTimeoutForWorkload`

go/vt/vttablet/tabletserver/tx_pool.go
```diff
@@ -76,31 +74,39 @@ type (
...
+		tp.ticks.Start(func() { tp.transactionKiller() })
+	}
 }
 
 // Close closes the TxPool. A closed pool can be reopened.
 func (tp *TxPool) Close() {
-	tp.ticks.Stop()
+	if tp.ticks.Running() {
+		tp.ticks.Stop()
+	}
```
harshit-gangal, line 109: stop is an idempotent method, we can avoid the running check on it.

go/vt/vttablet/tabletserver/tx_pool.go
```diff
@@ -76,31 +74,39 @@ type (
...
+	})
 	return axp
 }
 
 // Open makes the TxPool operational. This also starts the transaction killer
 // that will kill long-running transactions.
 func (tp *TxPool) Open(appParams, dbaParams, appDebugParams dbconfigs.Connector) {
 	tp.scp.Open(appParams, dbaParams, appDebugParams)
-	tp.ticks.Start(func() { tp.transactionKiller() })
+	tp.ticks.SetInterval(txKillerTimeoutInterval(tp.env.Config()))
```
harshit-gangal, line 99: this is initialized once in Open of txpool, we can remove this setting here.

go/vt/vttablet/tabletserver/tx_pool.go
```diff
@@ -141,7 +148,7 @@ func (tp *TxPool) transactionKiller() {
 		if conn.IsInTransaction() {
 			tp.txComplete(conn, tx.TxKill)
 		}
-		conn.Releasef("exceeded timeout: %v", tp.Timeout())
+		conn.Releasef("exceeded timeout: %v", timeout)
```
harshit-gangal, line 151: nit: can use `conn.timeout` directly here in the method.

harshit-gangal (commented):
Other things mostly looks good.

I think we should add more tests for Idle transaction getting killed by transaction killer and expiry time getting updated when we swtich workload. 
We might have those tests already, just check atleast.

go/vt/vttablet/tabletserver/tx_pool.go
maxenglander: Are you saying "we call `tp.ticks.SetInterval` once in `NewTxPool`", so we can remove it in `Open`?

The reason I have it in both places is because we have some unit tests that first call `NewTxPool`, and then change the TX timeout, and then do `Open`.

I will see if I can simplify things either in tests or in `NewTxPool`, so we only set the interval once.

go/cmd/vttestserver/main.go
```diff
@@ -206,6 +206,7 @@ func (t *topoFlags) buildTopology() (*vttestpb.VTTestTopology, error) {
 // Annoying, but in unit tests, parseFlags gets called multiple times per process
 // (anytime startCluster is called), so we need to guard against the second test
 // to run failing with:
+//
```
harshit-gangal, line 209: nit: looks like unrelated line change.

go/vt/servenv/grpc_auth.go
```diff
@@ -33,7 +33,7 @@ var grpcAuthServerFlagHooks []func(*pflag.FlagSet)
 // RegisterGRPCServerAuthFlags registers flags required to enable server-side
 // authentication in vitess gRPC services.
 //
-// `go/cmd/*`` entrypoints should call this function before
+// `go/cmd/*“ entrypoints should call this function before
```
harshit-gangal, line 36: nit: looks like unrelated change.

go/vt/vttablet/tabletmanager/vreplication/framework_test.go
```diff
@@ -72,10 +72,10 @@ type LogExpectation struct {
...
 
 // setFlag() sets a flag for a test in a non-racy way:
-//	* it registers the flag using a different flagset scope
-//	* clears other flags by passing a dummy os.Args() while parsing this flagset
-//	* sets the specific flag, if it has not already been defined
-//	* resets the os.Args() so that the remaining flagsets can be parsed correctly
+//   - it registers the flag using a different flagset scope
+//   - clears other flags by passing a dummy os.Args() while parsing this flagset
+//   - sets the specific flag, if it has not already been defined
+//   - resets the os.Args() so that the remaining flagsets can be parsed correctly
```
harshit-gangal, line 78: nit: looks like unrelated change.

harshit-gangal (commented):
I think few comments are still pending to be addressed.

harshit-gangal (approved):

go/vt/vttablet/tabletmanager/vreplication/framework_test.go
maxenglander: Was due to pre-commit hook forcing me to accept these formatting changes. Should I run with `git commit --no-verify`, or maybe my local tooling is out of date?

go/vt/servenv/grpc_auth.go
maxenglander: Was due to pre-commit hook forcing me to accept these formatting changes. Should I run with `git commit --no-verify`, or maybe my local tooling is out of date?

go/cmd/vttestserver/main.go
ajm188: > ... or maybe my local tooling is out of date?

actually the opposite; i noticed this after updating Go versions and `golangci-lint`. I actually added this newline in another PR, because golint yelled at me for it (so if you rebase/pull, this diff should actually go away)

go/vt/servenv/grpc_auth.go
ajm188: so, to be clear,

```suggestion
// `go/cmd/*` entrypoints should call this function before
```

go/vt/vttablet/tabletserver/tx_pool.go
maxenglander: Oops sorry I didn't implement this suggestion earlier. I saw it, but I misunderstood it 🤦

maxenglander:
@harshit-gangal I implemented the one outstanding suggestion I found (`timeout` => `conn.timeout`), sorry for missing that until now. Also reverted the unrelated fmt changes (used `git commit --no-verify`).

harshit-gangal (approved):

go/vt/vttablet/tabletmanager/vreplication/framework_test.go
harshit-gangal: to my understanding you did not change those packages so linter should not have failed for them.

ajm188: 🤷 maybe they did at one point, which triggered the hook to complain, and then they ended throwing away the change that caused this but keeping the formatting change

maxenglander: I think what happened is:

 1. I merged `main` into my branch
 2. There was a merge conflict, which left a bunch of upstream changes in the git staging area
 3. I resolved those conflicts and tried to commit
 4. Linting reformatted a bunch of the files from step 2.

PR #10938, opened 2022-08-04
[vtexplain] Switch vtexplain flags to use pflag hooks

ajm188:
## Description

What it says in the title :) 

```
➜  vitess git:(andrew/pflag-vtexplain) ✗ vtexplain --help 2>&1 | grep batch
      --batch-interval duration                                          Interval between logical time slots. (default 10ms)
➜  vitess git:(andrew/pflag-vtexplain) ✗ 
```

## Related Issue(s)

Closes #10930

## Checklist

-   [ ] "Backport me!" label has been added if this change should be backported
-   [x] Tests were added or are not required
-   [x] Documentation was added or is not required

Issue #10697, opened 2022-07-14
[VEP-4, phase 3] Switch flag definitions to be on `pflag` instead of `flag` and add make them package private.


ajm188:
This is the general, and tracking, issue for [VEP-4](https://github.com/vitessio/enhancements/blob/main/veps/vep-4.md), [phase 3](https://github.com/vitessio/enhancements/blob/main/veps/vep-4.md#phase-3-restructuring-and-encapsulation), in which flags currently defined on the global `flag.CommandLine` in packages throughout vitess are moved simultaneously to a `pflag.FlagSet` and out of the global namespace (and therefore, out of the usage text for commands that don't consume those flags at all).

## Table of Contents

- [Packages to Move](#packages-to-move)
- [Mover's Manual](#movers-guide)

## Packages to move <a href="#user-content-packages-to-move" id="packages-to-move">#</a>

This issue is considered done when each of these is complete

(I am slowly filing issues for each of these packages, please bear with me!)

N.B. this list was generated with the following command `git grep -E "\bflag\.[A-Z]" -- go/ | cut -d':' -f1 | xargs -I{} dirname {} | uniq` and then applying some light filtering (e.g. I know that `vtadmin` is fine, and I don't care about `wrangler/testlib`). If you disagree or have questions about a package, please don't hesitate to ask!!

<details><summary>Full command with filtering I used</summary>

```
git grep -E "\bflag\.[A-Z]" -- go/ | cut -d':' -f1 | xargs -I{} dirname {} | sort | uniq | grep -Ev '(testlib|vtadmin|endtoend)' | xargs -I{} echo "- [ ] \`{}\`"
```

</details>

- [x] `go/acl` (https://github.com/vitessio/vitess/issues/10731)
- [x] `go/cmd/mysqlctl` (#11266)
- [x] `go/cmd/mysqlctld` (#11267)
- [x] `go/cmd/query_analyzer` (nothing to do now)
- [x] `go/cmd/rulesctl/cmd` (https://github.com/vitessio/vitess/issues/11276)
- [x] `go/cmd/topo2topo` (https://github.com/vitessio/vitess/issues/11277)
- [x] `go/cmd/vtaclcheck` (https://github.com/vitessio/vitess/issues/11278)
- [x] `go/cmd/vtbackup` (https://github.com/vitessio/vitess/issues/11279)
- [x] `go/cmd/vtbench` (https://github.com/vitessio/vitess/issues/11287)
[...]

PR #10953, opened 2022-08-06
Fix main in CI

dbussink:
Right now tests are broken because of the interaction of https://github.com/vitessio/vitess/pull/10938 and https://github.com/vitessio/vitess/pull/10907

## Checklist


dbussink:
@ajm188 @deepthi dunno if this is the entirely correct change or that these flags should not be visible on `vtctld` and / or `vtexplain` but it makes sure that `main` passes CI again.

ajm188 (approved):
this is the right fix (for now, a later change for #10697) will clean this up so they don't appear (or any of the `tabletenv` flags, for that matter) at all

Issue #7572, opened 2021-03-01
Update Vitess to golang 1.16


systay:
Go 1.16 has been released. We should update Vitess to use this new version before too much time has passed.

Opening this issue to track remember and track the upgrade

dweitzman:
After upgrading to 1.16 it would also become possible to replace go.rice with the new embed directive

PR #10956, opened 2022-08-08
Cleanup usage of go.rice in favor of go:embed

dbussink:
The usage of go.rice predates the availability of go:embed, but we should switch to using go:embed instead to ship specific assets like config files that we need.

go.rice is also incompatible with Go 1.19 and while it might see a fix in the future, it seems better to go with the recommended Go approach that is available these days.

~~Right now is still a WIP since it doesn't remove the usage in `vtctld` yet where we package the UI assets. Want to see first what the fallout of this looks like on CI.~~

## Related Issue(s)

Was also mentioned before in the Go 1.16 upgrade but never tackled since: https://github.com/vitessio/vitess/issues/7572#issuecomment-787992585

## Checklist


mdlayher (approved):
LGTM, solid cleanup.

GuptaManan100 (approved):
Looks good to me

PR #10957, opened 2022-08-08
[cli] Migrate `vterrors` to `pflag`

### LINKED ISSUE #10894: Switch flag definitions to be on pflag instead of flag in `package go/vt/vterrors`

Current flags:

```
$ git grep -E "\bflag\.[A-Z]" -- go/vt/vterrors/*.go
go/vt/vterrors/vterrors.go:     flag.BoolVar(&LogErrStacks, "log_err_stacks", false, "log stack traces for errors")
```

ajm188:
## Description

Since every binary will involve errors in some way, shape, or form, I
decided to just register the vterrors flags to all binaries, rather than
naming each binary explicitly.

This notably does the _inverse_ of the code samples in the [Mover's Guide][1],
because `servenv` has `vterrors` in its dependency chain, so trying to
import `servenv` from `vterrors` would cause an import cycle.

This is the approach we should continue to use for any vitess package
imported by `servenv`.

[1]: see https://github.com/vitessio/vitess/issues/10697#user-content-movers-guide)

I also went ahead and made `logErrStacks` private; no one outside `vterrors` referenced it at all, so I didn't see a reason to keep it exported as a potential source of races.

## Related Issue(s)

Closes #10894.

## Checklist

### DIFF
--- a/go/vt/servenv/servenv.go
+++ b/go/vt/servenv/servenv.go
@@ -47,6 +47,7 @@ import (
 	"vitess.io/vitess/go/netutil"
 	"vitess.io/vitess/go/stats"
 	"vitess.io/vitess/go/vt/log"
+	"vitess.io/vitess/go/vt/vterrors"
 
 	// register the proper init and shutdown hooks for logging
 	_ "vitess.io/vitess/go/vt/logutil"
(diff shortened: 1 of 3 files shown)


deepthi (approved):
LGTM
Can you add a section to the `Mover's Guide` with this exception?
Writing it down only in this PR means that others might miss this nuance and have to discover it for themselves.

PR #10958, opened 2022-08-08
Fix parsing of `PARTITION BY KEY`

dbussink:
The `PARTITION BY KEY` syntax allows for passing in an empty column list which signifies that the primary key is used for partitioning.

This needs to be added to the parser to support this case.

See also https://dev.mysql.com/doc/refman/8.0/en/partitioning-key.html, specifically:

```
KEY takes only a list of zero or more column names. Any columns used as the partitioning key must
comprise part or all of the table's primary key, if the table has one. Where no column name is
specified as the partitioning key, the table's primary key is used, if there is one.
```

## Related Issue(s)

Found as part of #10203

Fixes https://github.com/vitessio/vitess/issues/10959

## Checklist


go/vt/sqlparser/sql.y
```diff
@@ -4778,6 +4778,19 @@ column_list_opt:
...
+    $$ = nil
+  }
+| sql_id
+  {
+    $$ = Columns{$1}
+  }
+| column_list_empty ',' sql_id
+  {
+    $$ = append($$, $3)
+  }
```
GuptaManan100, line 4792: Why not just use column_list here?

dbussink: Good point, that's nicer indeed. Let me fix that up.

dbussink: @GuptaManan100 Cleaned this up!

GuptaManan100:
@dbussink Do you mind creating an issue for the bug as well?

GuptaManan100 (approved):

shlomi-noach:
A new day, a new MySQL parser syntax surprise.

PR #10960, opened 2022-08-08
Move topo connection cache test to self hosted

rsajwani:
## Description

topo_connection_cache is flaky. Moving it to self hosted to be able to debug it.

## Checklist


deepthi:
Of course it passes on this PR...
You might want click re-run on Actions to try to get it to fail. I'm doing the first re-run now.

deepthi:
Fixed in #11007

PR #10961, opened 2022-08-09
Fixing flakiness in TestCrossCellDurability and TestHealthCheckCacheWithTabletChurn

GuptaManan100:
## Description

This PR attempts to fix the flakiness observed in `TestCrossCellDurability` and `TestHealthCheckCacheWithTabletChurn`.
The flakiness was tracked to the part where we bring up a new vttablet. When this tablet comes up, a shard primary already exists and the tablet goes to serving status very quickly. 
However, the test code waits for the tablet to reach the non-serving status. This led to a race between the code that checks the non-serving status and the tablet process which goes to serving status eventually.

This has been resolved by waiting for the tablet to reach the Serving status since that is the end state instead of waiting for the non-serving status which is the default.

## Checklist

### DIFF
--- a/go/test/endtoend/reparent/utils/utils.go
+++ b/go/test/endtoend/reparent/utils/utils.go
@@ -229,9 +229,9 @@ func StartNewVTTablet(t *testing.T, clusterInstance *cluster.LocalProcessCluster
 		require.FailNow(t, "Error starting mysql: %s", err.Error())
 	}
 
+	// The tablet should come up as serving since the primary for the shard already exists
+	tablet.VttabletProcess.ServingStatus = "SERVING"
 	err = tablet.VttabletProcess.Setup()
-	require.NoError(t, err)
-	err = tablet.VttabletProcess.WaitForTabletStatuses([]string{"SERVING", "NOT_SERVING"})
 	require.NoError(t, err)
 	return tablet
 }
--- a/go/test/endtoend/vtgate/tablet_healthcheck_cache/correctness_test.go
+++ b/go/test/endtoend/vtgate/tablet_healthcheck_cache/correctness_test.go
@@ -208,6 +208,7 @@ func addTablet(t *testing.T, tabletUID int, tabletType string) *cluster.Vttablet
 	err = proc.Wait()
 	require.Nil(t, err)
 
+	tablet.VttabletProcess.ServingStatus = ""
 	err = tablet.VttabletProcess.Setup()
 	require.Nil(t, err)
 


go/test/endtoend/reparent/utils/utils.go
```diff
@@ -229,10 +229,10 @@ func StartNewVTTablet(t *testing.T, clusterInstance *cluster.LocalProcessCluster
 		require.FailNow(t, "Error starting mysql: %s", err.Error())
 	}
 
+	// The tablet should come up as serving since the primary for the shard already exists
+	tablet.VttabletProcess.ServingStatus = "SERVING"
 	err = tablet.VttabletProcess.Setup()
 	require.NoError(t, err)
-	err = tablet.VttabletProcess.WaitForTabletStatuses([]string{"SERVING", "NOT_SERVING"})
```
mattlord, line 234: Why remove the wait, shouldn't it be SERVING?

GuptaManan100: There is a wait in `Setup` too, which is controlled by the `tablet.VttabletProcess.ServingStatus` property. This is why I set the property to `SERVING` a couple of lines earlier and removed the wait here entirely. It is redundant.

rsajwani (approved):

rsajwani:
given that you are fixing this code path can you fix the error message in vttablet_process.go where it says process '%s' timed out after 10s (while waiting for WaitForTabletStatus) , I believe this should be 30 secs.

mattlord (approved):
Thanks! Can you please note how you confirmed that it removed the flakiness? e.g. I would typically run the test locally 100 times in a loop that exits with loud warnings if there was a failure. Something like this (which I did for the new VDiff2 test):
```
$ for try in {1..100}; do go clean -testcache; go test -timeout 30m -v -failfast -run TestVDiff2 vitess.io/vitess/go/test/endtoend/vreplication; if [[ $? != 0 ]]; then echo "Failed on try # ${try}" && say "Danger, danger, yooooooo"; break; fi; done
```

GuptaManan100:
@mattlord I can confirm that this flakiness has been resolved. However, there is an alternate flakiness because of the race between the tablet type and PRS reading that tablet type which I haven't addressed in this PR yet.

GuptaManan100:
@rsajwani The timeout is 10 seconds. https://github.com/vitessio/vitess/blob/a56442af5cbb60498f8b026eebe72d8f8d82d080/go/test/endtoend/cluster/vttablet_process.go#L243

rsajwani:
> vttablet_process.go

I was talking about below.

	if vttablet.ServingStatus != "" {
		if err = vttablet.WaitForTabletStatus(vttablet.ServingStatus); err != nil {
			errFileContent, _ := os.ReadFile(fname)
			if errFileContent != nil {
				log.Infof("vttablet error:\n%s\n", string(errFileContent))
			}
			return fmt.Errorf("process '%s' timed out after 10s (err: %s)", vttablet.Name, err)
		}
	}
	
	and this 
	
	// WaitForTabletStatus waits for 10 second till expected status is reached
func (vttablet *VttabletProcess) WaitForTabletStatus(expectedStatus string) error {
	return vttablet.WaitForTabletStatusesForTimeout([]string{expectedStatus}, 30*time.Second)
}

PR #10966, opened 2022-08-09
Merge subqueries that "join" on lookup index columns.

arthurschreiber:
## Description

This pull request tries to improve the handling of queries that contain `IN` or `=` operations with one operant being a lookup vindex column, and the other operant being a subquery that selects a column from the same or a different table that uses the same lookup vindex.

Here's an example:

```sql
SELECT music.id FROM music WHERE music.id IN (SELECT music.id FROM music WHERE music.user_id IN (1, 2, 3))
```

Without this change, the planner breaks this query into two parts, one part executing the inner query and another for the outer query.

But because we're selecting data from `music.id` in the inner query, which has a unique lookup index defined, and comparing it to `music.id` in the outer query, we can merge the inner and outer queries together.

I'm fairly confident the code changes are bad, but I think they show that this change is viable. 😅 

## Checklist


go/vt/vtgate/planbuilder/physical/route_planning.go
```diff
@@ -784,10 +785,55 @@ func canMergeOnFilter(ctx *plancontext.PlanningContext, a, b *Route, predicate s
...
+				return nil
+			}
+
+			exp = aliasedExpr.Expr
+
+			_, isCol = exp.(*sqlparser.ColName)
+			if !isCol {
+				return nil
+			}
+		}
```
arthurschreiber, line 836: This should probably be extracted into a seperate function.

glortho:
Using this branch, we're seeing this error occasionally when trying to vtexplain these types of queries:

```
[BUG] unknown type encountered: *physical.Table (transformToLogicalPlan)
```

Eyeballing them, it seems all the problematic ones have a `LEFT OUTER JOIN` + at least one other join in common.

arthurschreiber:
> Using this branch, we're seeing this error occasionally when trying to vtexplain some of these types of queries:

I believe this was fixed with some of the recent changes I pushed. 👍

arthurschreiber:
This pull request also contains the change described in https://github.com/vitessio/vitess/pull/11072#issuecomment-1223981492

go/vt/vtgate/planbuilder/physical/route_planning.go
```diff
@@ -867,12 +868,168 @@ func canMergeOnFilter(ctx *plancontext.PlanningContext, a, b *Route, predicate s
...
 func findColumnVindex(ctx *plancontext.PlanningContext, a *Route, exp sqlparser.Expr) vindexes.SingleColumn {
-	_, isCol := exp.(*sqlparser.ColName)
+	colName, isCol := exp.(*sqlparser.ColName)
 	if !isCol {
 		return nil
 	}
 
+	derived, isDerived := a.Source.(*Derived)
+	if isDerived {
+		selectQuery := derived.Query.(*sqlparser.Select)
```
systay, line 1008: union will panic here

go/vt/vtgate/planbuilder/physical/route_planning.go
```diff
@@ -919,12 +920,71 @@ func canMergeOnFilter(ctx *plancontext.PlanningContext, a, b *Route, predicate s
...
+	rightSelection := extractSingleColumnSubquerySelection(predicate.Subquery)
+	if rightSelection == nil {
+		return false
+	}
+
+	rVindex := findColumnVindex(ctx, b, rightSelection)
+	if rVindex == nil {
+		return false
+	}
+	return rVindex == lVindex
```
GuptaManan100, line 944: Do we need to find the column vindex given that `extractSingleColumnSubquerySelection` will return a column and we have a lVindex. Isn't there a way to check that the column is covered by this vindex. What would happen if we have two or more vindexes on the same column?

go/vt/vtgate/planbuilder/physical/subquery_planning.go
```diff
@@ -127,9 +130,59 @@ func mergeSubQueryOp(ctx *plancontext.PlanningContext, outer *Route, inner *Rout
 	if err != nil {
 		return nil, err
 	}
+
+	if subQueryWasPredicate {
+		// Copy Vindex predicates from the inner route to the upper route (probably not right)
```
GuptaManan100, line 135: What do we mean by `probably not right`?

go/vt/vtgate/planbuilder/testdata/select_cases.txt
```diff
@@ -5891,40 +5841,19 @@ Gen4 plan same as above
...
   "Original": "SELECT music.id FROM music WHERE music.id IN (SELECT music.id FROM music WHERE music.foo = 'bar') AND music.user_id IN (3, 4, 5)",
   "Instructions": {
-    "OperatorType": "Subquery",
-    "Variant": "PulloutIn",
-    "PulloutVars": [
-      "__sq_has_values1",
-      "__sq1"
+    "OperatorType": "Route",
+    "Variant": "IN",
+    "Keyspace": {
```
GuptaManan100, line 5846: If I understand correctly, we should also be able to merge something like `SELECT music.id FROM music WHERE music.id IN (SELECT music.id FROM music WHERE music.foo = 'bar')` Could we add a test for this?

arthurschreiber: I think we already have this test case here:

https://github.com/vitessio/vitess/blob/3719ffb79d56881049bc4ba014b4699d6280f655/go/vt/vtgate/planbuilder/testdata/select_cases.txt#L6111-L6170

That seems to be exactly the test you suggested? 🙇‍♂️

go/vt/vtgate/planbuilder/physical/subquery_planning.go
arthurschreiber: I think that comment is outdated. I'll check back with @systay, but when I initially implemented this I was not sure if just copying the predicates over was the correct way of doing this.

go/vt/vtgate/planbuilder/testdata/select_cases.txt
GuptaManan100: Yep yep, that works 👍

go/vt/vtgate/planbuilder/physical/route_planning.go
systay: I don't understand why someone would create two different unique vindexes on the same column. if that would happen, and the vschema would give us different vindexes when querying the two parts of the query, the worst thing that would happen would be that we don't merge the subquery into the outer part of the query, and that is not that bad

go/vt/vtgate/planbuilder/physical/subquery_planning.go
```diff
@@ -141,12 +194,51 @@ func tryMergeSubQueryOp(
 	var err error
 	switch outerOp := outer.(type) {
 	case *Route:
-		if shouldTryMergingSubquery(outerOp, subq) {
-			merged, err = tryMerge(ctx, outerOp, subq, joinPredicates, merger)
+		subqueryRoute, isRoute := subq.(*Route)
+		if !isRoute {
+			return nil, nil
+		}
+
```
harshit-gangal, line 201: the switch case has become too long, may be better to have some small functions.

harshit-gangal (approved):
This is really great optimization.
See if some helpful comments can be added on top of the functions to help understand the merge rules.

go/vt/vtgate/planbuilder/physical/route_planning.go
GuptaManan100: Okay, makes sense.

systay (approved):

PR #10969, opened 2022-08-10
Ensure VDiff Engine is open in RPC entrypoint

mattlord:
## Description

`VDiffEngine.PerformVDiffAction` is the RPC entrypoint for `VDiff2` in `TabletManager`: https://github.com/vitessio/vitess/blob/f727bb192666e297fa16be2fac8b76ab9d998f8d/go/vt/vttablet/tabletmanager/rpc_vdiff.go#L9-L12

We should check to be sure that the engine is open — which should only be true on a `PRIMARY` tablet — to prevent operations from being attempted on non-PRIMARY / read-only tablets.

## Checklist


rsajwani:
Is there a unit test to verify the change??

mattlord:
> Is there a unit test to verify the change??

No. Are you're asking if I can add one? What do you think the unit test should verify (with mock tablets and vdiff engines, which we don't have today)? The RPC should not be called on non-primary tablets but this is an additional safety mechanism to prevent anything from potentially coming in e.g. during transitions like PRS/ERS (which should also be blocked by semi-sync and read-only mode). This is not a behavior change it's just an added safety check.

The vreplication engine has the same behavior at its entry point as a failsafe check. Even though it *does* have an extensive unit test framework it doesn't test that specific thing: https://github.com/vitessio/vitess/blob/main/go/vt/vttablet/tabletmanager/vreplication/engine_test.go

I went ahead and added a simple unit test for it: https://github.com/vitessio/vitess/pull/10969/commits/1fdded36dde0370fd984905f6750b4f3f4870be9

rsajwani (approved):

PR #10997, opened 2022-08-11
evalengine: support mismatched numerical types

vmg:
## Description

Found this while working on some internal stuff. The `evalengine` does not properly handle mismatched numeric types because we were not properly normalizing. E.g. an UINT8(0) would fail to compare equally to an UINT64(0), even though an INT8(0) would actually compare equally to a UINT64(0). This PR fixes this issue by upcasting all comparison types before performing a numeric comparison.

It probably requires a backport. 

cc @systay @deepthi 

## Checklist

### DIFF
--- a/go/test/endtoend/vtgate/gen4/gen4_test.go
+++ b/go/test/endtoend/vtgate/gen4/gen4_test.go
@@ -493,7 +493,5 @@ func TestFilterOnLeftOuterJoin(t *testing.T) {
 				  and team_fact.team >= 22
 				)`
 
-	// Ideally we should get `[[INT32(22)] [INT32(33)]]`
-	// Change the expectation when we support better conversion handling in numeric comparison.
-	utils.AssertContainsError(t, mcmp.VtConn, query, `unsupported: cannot compare INT32 and INT64`)
+	mcmp.AssertMatches(query, "[[INT32(22)] [INT32(33)]]")
 }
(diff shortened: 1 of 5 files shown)


dbussink (approved):
Kinda surprised this bug has apparently been lingering in here for a long time then, but good that we have found it and fixed it. 

Agreed that we should back port this since it could lead to weird and subtle bugs I suspect we don't really know about.

vmg:
We've just noticed @harshit-gangal's https://github.com/vitessio/vitess/pull/10793, which found this bug already and added a check to the `evalengine` to error out when the bug was triggered.

This PR follows up on that by removing the check and instead always performing the right comparison. The tests that were previously testing the error have been updated and are now green. :sparkles:

vmg:
@harshit-gangal: the end2end test no longer errors, but it doesn't appear to be giving the right results either:

```
=== RUN   TestFilterOnLeftOuterJoin
    mysql.go:188: Query (select team.id
        				from team_fact
        				  join team on team.id = team_fact.team
        				  left outer join team_member on team_member.team = team.id
        				where (
        				  team_fact.fact = 'A'
        				  and team_member.user is null
        				  and team_fact.team >= 22
        				)) results mismatched.
        Vitess Results:
        [INT32(22)]
        [INT32(33)]
        MySQL Results:
        
--- FAIL: TestFilterOnLeftOuterJoin (0.05s)
```

Did I not write the test properly?

dbussink:
> Did I not write the test properly?

There was a bug in the test that I fixed up, so it should be green now.

harshit-gangal (approved):

Issue #9706, opened 2022-02-16
RFC: Handle reserved/tainted connections better


harshit-gangal:
## Problem Statement

Vitess has a feature where it allows changing session-level system settings on a connection. In Vitess terminology, these connections are called tainted connections. As the connection properties are modified on those connections, they cannot be returned to the originating connection pool. After the client session is closed, these connections are closed.

As these connections cannot be returned to the pool, till the lifetime of the session they are reserved for those sessions. On the VTGate side, these connection ids are stored on the session. They are called reserved connections. Any further query will use these reserved connections to send the query to MySQL.

The usage of this feature was meant to be restricted for one-off cases. It was advised that the Vitess users should configure the MySQL properties in such a way that they do not need to use this feature (i.e. change the system properties upfront at MySQL level, to match the properties the application requires).

Over time we have seen that Vitess connection will not come from a single application but multiple different applications with different application frameworks. Due to this, every application framework will try to set different kinds/values various system properties. This causes many or all of the connections between vttablet and MySQL to be reserved connections for the session, and we will run into the maximum allowed number of MySQL connections.

Also, we have seen that sometimes Vitess users do not have control over changing the system settings at the MySQL level, and hence it causes MySQL to run out of connections because of the above reserved connections problem.

To avoid these scenarios, we have to come up with a design that can handle these cases and deal with application frameworks setting the system variables upfront on the connections.

#### Proposal 1:
Set the appropriate session properties after acquiring a connection from the pool and reset the settings on returning to the pool.

#### Proposal 2:
For MySQL 8.0+ we can use optimizer hint SET_VAR as a query comment. The optimizer hints are only applicable to the scope of a single query. This will not change the connection permanently and hence can be returned to the pool.

#### Proposal 3:
In the transaction pool, currently, we have an originating pool and a transactions registry where active transactions live (currently tainted connections are also present there without having an active transaction). 
The proposal is to write a new pool implementation which can handle the connection which have settings applied on it.
Any request comes that also updates/changes the system settings, gets the connection from this new pool which will know how to efficiently provide and manage such connections

## Benchmarks

| Proposal | Description | Operations | Time per Op |
| ------------ | --------------- | --------------- | ---------------- |
| 2 | System variables as query hints | 2233 | 549261 ns/op |
| 1 | System variables exec before each query exec | 1663 | 651643 ns/op |
| 1 | System variables exec along with query exec | 1946 | 553904 ns/op |
[...]

PR #11002, opened 2022-08-12
Remove prefill logic from resource pool

harshit-gangal:
## Description

This PR modifies the resource pool by removing prefill logic and modifying benchmark tests.

## Related Issue(s)

- #9706 

## Checklist


rsajwani:
I went through RFC as well but not able to find if this change is part of proposal #1, 2 or 3 .. Can you give me more detail why removing pre-fill logic

harshit-gangal:
> I went through RFC as well but not able to find if this change is part of proposal #1, 2 or 3 .. Can you give me more detail why removing pre-fill logic

It is refactor of the pool, and not directly related to RFC.

The prefill logic was not used in vitess deployment as the default value is `0` and the connection pool should be having lazy behaviour regarding opening connections to mysql.

go/pools/resource_pool.go
```diff
@@ -238,21 +206,13 @@ func (rp *ResourcePool) reopen() {
 // it will wait till the next resource becomes available or a timeout.
 // A timeout of 0 is an indefinite wait.
 func (rp *ResourcePool) Get(ctx context.Context) (resource Resource, err error) {
-	span, ctx := trace.NewSpan(ctx, "ResourcePool.Get")
```
systay, line 241: this seems unrelated. Why remove the tracing here?

harshit-gangal: The trace is already happening at one higher up level in connpool.Pool
Seems like a unnecessary trace.

```
	span.Annotate("capacity", p.Capacity())
	span.Annotate("in_use", p.InUse())
	span.Annotate("available", p.Available())
	span.Annotate("active", p.Active())

	if cp.timeout != 0 {
		var cancel context.CancelFunc
		ctx, cancel = context.WithTimeout(ctx, cp.timeout)
		defer cancel()
	}
	r, err := p.Get(ctx, nil)
```

systay (approved):

PR #11006, opened 2022-08-12
Tests: AddShard should use targeted MySQL version

mattlord:
## Description

When the host machine had a different major version of MySQL
installed in the standard location than we targeted in the
test, the shards added for the Reshard workflows were using
the default version installed which then differed from the
other shards -- which were using the version downloaded into
/tmp -- and caused schema differences that were surfaced in
CopySchemaShard.

## Checklist


deepthi (approved):

rsajwani (approved):
LGTM

PR #11008, opened 2022-08-13
Run latest gofmt on everything & address linter warnings

mattlord:
## Description

After upgrading to go 1.19 and the latest [`golangci-lint`](https://golangci-lint.run/usage/install/) I worked on some pretty large merges that touched 500+ files. I was getting `gofmt` warnings a lot and also seeing this pop up (had never seen it before):
```
SA1019: "io/ioutil" has been deprecated since Go 1.16: As of Go 1.16, the same functionality is now provided by package io or package os, and those implementations should be preferred in new code. See the specific function documentation for details. (staticcheck)
	"io/ioutil"
```

So I addressed those linter warnings by removing our usage of [that deprecated package](https://github.com/go-critic/go-critic/issues/1019). Ran the latest `gofmt` on everything with `gofmt -s -w .` and then addressed the new linter warnings (since now I was touching so many files). 

## Checklist

-   [x] Tests wereare not required


deepthi:
Let's say we merge this PR. What happens when someone who's running go1.18.4 locally tries to make changes? Will they now start getting the opposite changes from the 1.18.4 version of gofmt and have to undo many of the changes in this PR to satisfy it?
My take is that the only thing we should do right now is change `ioutil` to `os` if that. All other changes will need to happen as part of upgrading the official go version to 1.19, and we are not ready for that.

mattlord:
> Let's say we merge this PR. What happens when someone who's running go1.18.4 locally tries to make changes? Will they now start getting the opposite changes from the 1.18.4 version of gofmt and have to undo many of the changes in this PR to satisfy it?

No. The new fmt directives did not exist in the older version so it's not bothered by them:
```
$ go version
go version go1.18.4 darwin/arm64

$ gofmt -s -w .

$ git status
On branch linter_SA1019
Your branch is up to date with 'origin/linter_SA1019'.

nothing to commit, working tree clean
`

> My take is that the only thing we should do right now is change `ioutil` to `os` if that. All other changes will need to happen as part of upgrading the official go version to 1.19, and we are not ready for that.

Then IMO we should prevent people from using 1.19, which we do not today (we only enforce a minimum version).

File: examples/local/vstream_client.go [hunk 1]
```diff
@@ -33,8 +33,8 @@ import (
 )
 
 /*
-	This is a sample client for streaming using the vstream API. It is setup to work with the local example and you can
-    either stream from the unsharded commerce keyspace or the customer keyspace after the sharding step.
+		This is a sample client for streaming using the vstream API. It is setup to work with the local example and you can
+	    either stream from the unsharded commerce keyspace or the customer keyspace after the sharding step.
```
deepthi, line 37: Not sure why the formatter is being picky like this. It seems unnecessary, but 🤷

deepthi (approved):

PR #11011, opened 2022-08-15
Online DDL, CancelMigration: distinguish user-issued vs. internally-issued cancellation

shlomi-noach:
## Description

#10900 introduced `cancelled_timestamp` column, which is populated when a migration is cancelled. Then, based on the value of this column, and assuming a problem in the migration, a migration transitions to either `failed` (if the column is `NULL`) or to `cancelled` (if the column is `NOT NULL`).

As quick recap, `gh-ost`, `pt-osc` and even `vreplication` itself, to some extent, act as 3rd party tools for Online DDL. To cancel a migration is to fail the migration; #10900 was made so that we can make a distinction between a "legitimately failed" migration to a "user cancelled" migration.

However, we left out a couple use cases, where the Online DDL executor _itself_ may cancel a migration:

- if the migration is stale
- if there's an unrecoverable error in the migration flow

In these scenarios, the executor calls `CancelMigration`, but we expect the terminal state to be `failed`, not `cancelled`.

In this PR we make the distinction between a user-generated `CANCEL` (e.g. user issued a `ALTER VITESS_MIGRATION ... CANCEL` command), and an internal-generated cancellation. the latter now leads to `failed` state.

## Related Issue(s)

- #6926

## Checklist

### DIFF
--- a/go/vt/vttablet/onlineddl/executor.go
+++ b/go/vt/vttablet/onlineddl/executor.go
@@ -1769,7 +1769,7 @@ func (e *Executor) terminateMigration(ctx context.Context, onlineDDL *schema.Onl
 }
 
 // CancelMigration attempts to abort a scheduled or a running migration
-func (e *Executor) CancelMigration(ctx context.Context, uuid string, message string) (result *sqltypes.Result, err error) {
+func (e *Executor) CancelMigration(ctx context.Context, uuid string, message string, issuedByUser bool) (result *sqltypes.Result, err error) {
 	if !e.isOpen {
 		return nil, vterrors.New(vtrpcpb.Code_FAILED_PRECONDITION, "online ddl is disabled")
 	}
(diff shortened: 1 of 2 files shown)


deepthi (commented):
LGTM
What about tests?

shlomi-noach:
Mentioned in a private channel, tests are complex. I'm not sure right now. Reason is that the scenarios where a migration gets cancelled by the scheduler are a bit extreme and will be difficult to reproduce synthetically.

deepthi:
> Mentioned in a private channel, tests are complex. I'm not sure right now. Reason is that the scenarios where a migration gets cancelled by the scheduler are a bit extreme and will be difficult to reproduce synthetically.

You may need to build a unit test framework to simulate vreplication stream errors. We can let this through for now, because that will surely take time to do.

deepthi (approved):

PR #11013, opened 2022-08-15
Online DDL: support multiple 'ADD FULLTEXT KEY' in single ALTER

shlomi-noach:
## Description

MySQL does not support multiple `ADD FULLTEXT KEY` statements in a single `ALTER` (for InnoDB tables). But in this PR, Online DDL does. It identifies such statement and explodes it into multiple valid MySQL statements. To the user, and to the Online DDL scheduler / state machine, this appears to be atomic.

Tests added.

## Related Issue(s)

- https://github.com/vitessio/vitess/pull/11012
- #6926

## Checklist


go/vt/vttablet/onlineddl/executor.go
```diff
@@ -983,9 +984,38 @@ func (e *Executor) validateAndEditAlterTableStatement(ctx context.Context, alter
...
+		}
 	}
-	return nil
+	for i := 0; i < len(alterTable.AlterOptions); {
+		if alterTable.AlterOptions[i] == nil {
+			alterTable.AlterOptions = append(alterTable.AlterOptions[0:i], alterTable.AlterOptions[i+1:]...)
+		} else {
+			i++
+		}
+	}
```
dbussink, line 1017: Guess there's not really a much more simple way to get these out of the slice? Alternatively I guess the previous loop could build a new one and we replace that at the end, but dunno if that looks much clearer?

dbussink (approved):

go/vt/vttablet/onlineddl/executor.go
shlomi-noach: Honestly I spent an hour over this silly thing; there is no canonical way in `golang`, that I can find, for removing elemtns from a slice while iterating it. I don't know what's the least evil way of doing it.

dbussink: @shlomi-noach Yeah, have solved it in the past with allocating a same sized slice and then copy into that instead vs. trying to remove from an existing slice. But this also works.

PR #11018, opened 2022-08-16
Use the correct error type for dependencies

dbussink:
This was using accidentally the wrong error type that was for views not found during diff application. Nothing was using this new error type yet, which is incorrect.

## Related Issue(s)

Changed in https://github.com/vitessio/vitess/pull/10940 which added the type but didn't use it. 

## Checklist


shlomi-noach (approved):
Whoops. Thanks for spotting this!

PR #11021, opened 2022-08-16
refactor: group all system setting query into single set statement

harshit-gangal:
## Description

This PR sends the multiple set statements into a single set statement to vttablet.
This will avoid multiple round trips at vttablet-mysql with a single call.

## Related Issue(s)

- #9706 

## Checklist

### DIFF
--- a/go/vt/vtgate/safe_session.go
+++ b/go/vt/vtgate/safe_session.go
@@ -18,17 +18,16 @@ package vtgate
 
 import (
 	"fmt"
+	"sort"
 	"strings"
 	"sync"
 	"time"
 
-	"vitess.io/vitess/go/vt/sqlparser"
-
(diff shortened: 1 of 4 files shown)


go/vt/vtgate/safe_session.go
```diff
@@ -543,13 +542,31 @@ func (session *SafeSession) SetReservedConn(reservedConn bool) {
 func (session *SafeSession) SetPreQueries() []string {
```
systay, line 542: don't we want to change the return type to be a single string instead of a slice?

harshit-gangal: currently the vttablet api takes in []string so we cannot change that one release.

systay (approved):

PR #11022, opened 2022-08-16
Run upgrade/downgrade tests on main

frouioui:
## Description

This PR enables upgrade downgrade tests on the `main` branch of this repository. It was previously disabled to ensure we ran the tests only when the `Skip Upgrade/Downgrade tests` label was not set. But since we have removed this assertion, we can safely re-run the tests on `main`.

## Checklist


deepthi (approved):
Thank you!

deepthi:
vtbackup_transform has been flaky. we are going to move it to a self-hosted runner for debugging.

PR #11026, opened 2022-08-17
CI: Move all ubuntu-18.04 to ubuntu-20.04

deepthi:
## Description
Ideally we should run on `ubuntu-latest` but let's see if this works first.

## Checklist

### DIFF
--- a/.github/workflows/cluster_endtoend_13.yml
+++ b/.github/workflows/cluster_endtoend_13.yml
@@ -14,7 +14,7 @@ env:
 jobs:
   build:
     name: Run endtoend tests on Cluster (13)
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
 
     steps:
     - name: Check if workflow needs to be skipped
--- a/.github/workflows/cluster_endtoend_15.yml
+++ b/.github/workflows/cluster_endtoend_15.yml
@@ -14,7 +14,7 @@ env:
 jobs:
   build:
     name: Run endtoend tests on Cluster (15)
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
 
     steps:
     - name: Check if workflow needs to be skipped
--- a/.github/workflows/cluster_endtoend_21.yml
+++ b/.github/workflows/cluster_endtoend_21.yml
@@ -14,7 +14,7 @@ env:
 jobs:
   build:
     name: Run endtoend tests on Cluster (21)
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
 
     steps:
     - name: Check if workflow needs to be skipped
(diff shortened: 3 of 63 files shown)


dbussink:
> Ideally we should run on `ubuntu-latest` but let's see if this works first.

Probably good to define a specific version so we don't have to scramble if Actions changes `ubuntu-latest`. Fwiw, `ubuntu-latest` still points at `20.04` anyway and not at `22.04`. 

With `22.04` there's also other issues like it ships with OpenSSL 3.0 and only MySQL 8.0.30 has support for that. I guess binary packages might have those fixes backported by maintainers but it might be a friction point there. I also don't know if MySQL 5.7 would ever see OpenSSL 3.0 support so we might be stuck there with an older base for CI anyway.

dbussink:
Hmm, is the `xtrabackup` related issue that we've upgraded it also for MySQL 5.7 but we now get a version there that's doesn't work with 5.7? Since we already seemed to have MySQL 8.0 running against `20.04` if I read the diff right?

dbussink:
We also seem to have some other stuff that might now run against MySQL 8.0 and not 5.7?

```
--- FAIL: TestSchemaVersioning (0.25s)
[325](https://github.com/vitessio/vitess/runs/7952828721?check_suite_focus=true#step:11:326)
    vstreamer_test.go:169: insert into vitess_version values(1, 10) (0): event:
[326](https://github.com/vitessio/vitess/runs/7952828721?check_suite_focus=true#step:11:327)
        "type:FIELD field_event:{table_name:\"vitess_version\" fields:{name:\"id1\" type:INT32 table:\"vitess_version\" org_table:\"vitess_version\" database:\"vttest\" org_name:\"id1\" column_length:11 charset:63 column_type:\"int\"} fields:{name:\"id2\" type:INT32 table:\"vitess_version\" org_table:\"vitess_version\" database:\"vttest\" org_name:\"id2\" column_length:11 charset:63 column_type:\"int\"}}", want
[327](https://github.com/vitessio/vitess/runs/7952828721?check_suite_focus=true#step:11:328)
        "type:FIELD field_event:{table_name:\"vitess_version\" fields:{name:\"id1\" type:INT32 table:\"vitess_version\" org_table:\"vitess_version\" database:\"vttest\" org_name:\"id1\" column_length:11 charset:63 column_type:\"int(11)\"} fields:{name:\"id2\" type:INT32 table:\"vitess_version\" org_table:\"vitess_version\" database:\"vttest\" org_name:\"id2\" column_length:11 charset:63 column_type:\"int(11)\"}}"
[328](https://github.com/vitessio/vitess/runs/7952828721?check_suite_focus=true#step:11:329)
FAIL
```

That looks very much like an issue where on MySQL 5.7 integer lengths are still printed for a schema, but that doesn't happen anymore for MySQL 8.0.

deepthi:
Some parts of this have already been picked out and merged, and #11153 handles the rest. Hence, closing.

PR #11027, opened 2022-08-17
evalengine: Support built-in MySQL function CEIL()

Weijun-H:

## Description
Implement support CEIL() to understand how the evaluation engine works

## Related Issue(s)
- #9647 

## Checklist


go/vt/vtgate/evalengine/func.go
```diff
@@ -643,3 +645,48 @@ func (ta *typeAggregation) result() sqltypes.Type {
...
+	argtype := inarg.typeof()
+	if inarg.isNull() {
+		result.setNull()
+		return
+	}
+
+	if sqltypes.IsIntegral(argtype) {
+		inarg.makeFloat()
+		result.setInt64(int64(math.Ceil(inarg.float64())))
+	} else if sqltypes.Decimal == argtype {
```
vmg, line 662: This is a small nit, but we don't have any reverse comparisons in the Vitess codebase. As you may be aware of, reverse comparisons have their origin in the C family of languages because `if (argtype = sqltypes.Decimal)` is valid syntax and does something very different than `if (argtype == sqltypes.Decimal)`, so reversing the comparison may prevent a bad bug from being deployed.

However, Go does not allow assignments in `if` expressions, so we always write `if argtype == sqltypes.Decimal {` -- there's no point on doing otherwise except making the code harder to read!

go/vt/vtgate/evalengine/func.go
```diff
@@ -643,3 +645,48 @@ func (ta *typeAggregation) result() sqltypes.Type {
...
+	inarg := &args[0]
+	argtype := inarg.typeof()
+	if inarg.isNull() {
+		result.setNull()
+		return
+	}
+
+	if sqltypes.IsIntegral(argtype) {
+		inarg.makeFloat()
+		result.setInt64(int64(math.Ceil(inarg.float64())))
```
vmg, line 661: I don't see how this could possibly do anything! The result of converting an integral to a float64 will always be a round number, so calling `math.Ceil` on it will be a no-op. The whole roundtrip from integral -> float -> integral seems to be superfluous.

Have you found any corner cases that that roundtrip handles properly? From testing in the MySQL CLI, `CEIL` on an integer looks like a noop -- it always returns the same integer.

go/vt/vtgate/evalengine/func.go
```diff
@@ -643,3 +645,48 @@ func (ta *typeAggregation) result() sqltypes.Type {
...
+		if num.Cmp(decimal.NewFromInt(math.MaxInt32)) == 1 || num.Cmp(decimal.NewFromInt(math.MinInt32)) == -1 {
+			if num.Sign() == 1 {
+				result.setDecimal(num.Add(decimal.NewFromInt(1)), 0)
+			} else {
+				result.setDecimal(num.Add(decimal.NewFromInt(-1)), 0)
+			}
+		} else {
+			floatpart, _ := num.Float64()
+			result.setInt64(int64(math.Ceil(floatpart)))
+		}
```
vmg, line 673: Can you explain a little bit what are you trying to do here? It seems like you're using different rounding modes for decimals that are larger than INT32_MAX or smaller than INT32_MIN -- otherwise you're just rounding the decimal as if it were a float...

I don't think this is exactly MySQL's behavior. Are you attempting an optimization here? I.e. doing the float rounding if it would always be safe to do so? I'm afraid this won't behave like you're expecting it to... Converting a decimal to a `float64` is super expensive! The comparisons against `math.MaxInt32` and `math.MaxInt32` are also expensive, particularly because they allocate two temporary decimals, and the `num.Float64` would tell you anyway if the conversion to `float64` was safe to perform.

I think this is a case where the simplest code would would also be the safest: you can add (or subtract) one from the decimal and truncate the decimals, and that should give you a good ceil... but here's an even better tip: if you check the implementation of our `decimal` package, you'll see that we've adapted an existing arbitrary-precision decimal implementation to behave like MySQL's. Maybe the original implementation has a `Ceil` function you could port over? :)

go/vt/vtgate/evalengine/integration/comparison_test.go
```diff
@@ -681,3 +685,27 @@ func TestCaseExprWithValue(t *testing.T) {
...
+		"'1.5'",
+		"NULL",
+		"'ABC'",
+		"1.5e0",
+		"-1.5e0",
+		"9223372036854775810.4",
+		"7223372036854775010.1",
+		"2147483640.1",
+		"9223372036854775010.4",
+	}
```
vmg, line 706: Good start with the tests. Can you think of more corner cases that would be worth adding here? Larger negative floats? Larger negative decimals?

vmg (changes requested):
Hey Alex! Off to a good start! I left you some feedback on the things that need fixing.

go/vt/vtgate/evalengine/func.go
Weijun-H: I am confused with the mysql behaviors here. For example, when I test the value like `1.5`
``` bash
/home/hwj/cncf/vitess/go/vt/vtgate/evalengine/integration/comparison_test.go:710: 
different results: DECIMAL(2); mysql response: INT64(2) (local collation: binary; mysql collation: binary)
      query: SELECT CEIL(1.5) (SIMPLIFY=true)
```
Then I checked the type of the `1.5`, it is not the `FLOAT`, but `Demical`. Do I misunderstand the process here?

vmg: From the MySQL docs:

> For exact-value numeric arguments, the return value has an exact-value numeric type. For string or floating-point arguments, the return value has a floating-point type.

It seems like rounding a DECIMAL, which is an exact value numeric type, will return INT64 if the result is small enough to fit, otherwise a DECIMAL.

Weijun-H: It it tricky to detect when to return `INT`, because:

``` go 

"9223372036854775810.4", // bigger than MAXINT64, return DECIMAL
"7223372036854775010.1", // bigger than MAXINT32, return DECIMAL
"2147483648.1", // bigger than MAXINT32, return INT

```

vmg: All you have to check is whether the result of the rounding fits in `int64` or not -- there are no intermediate types. If it fits, it is returned as `INT64`, if not as `DECIMAL`.

Weijun-H: When the type of args is `DECIMAL`, but the result is `INT64`. How could we return the true type in the `type` function. My current solution:
``` go
func (builtinCeil) call(env *ExpressionEnv, args []EvalResult, result *EvalResult) {
	inarg := &args[0]
	argtype := inarg.typeof()
	if inarg.isNull() {
		result.setNull()
		return
	}

	if sqltypes.IsIntegral(argtype) {
		result.setInt64(inarg.int64())
	} else if sqltypes.Decimal == argtype {
		num := inarg.decimal()
		num = num.Ceil()
		intnum, isfit := num.Int64()
		if isfit {
			inarg.makeSignedIntegral()
			result.setInt64(intnum)	
		} else {
			result.setDecimal(num, 0)
		}
	} else {
		inarg.makeFloat()
		result.setFloat(math.Ceil(inarg.float64()))
	}
}

func (builtinCeil) typeof(env *ExpressionEnv, args []Expr) (sqltypes.Type, flag) {
	if len(args) != 1 {
		throwArgError("CEIL")
	}
	t, f := args[0].typeof(env)
	if sqltypes.IsIntegral(t) {
		return sqltypes.Int64, f
	} else if sqltypes.Decimal == t {
		return sqltypes.Decimal, f
	} else {
		return sqltypes.Float64, f
	}
}
```

How could we catch the new return type, like `1.5`?

vmg: Huh, that's a good question. MySQL's behavior here is really bad... Our `typeof` helpers cannot (should not, really) return variable types for the same input, so we're gonna have to hack it and do something different to MySQL... Let's assume for now that `typeof` returns `Decimal` for `Decimal`. That'll give some bad `typeof` results from the tests, and we'll have to ignore them, but I think that's the best we can do for now.

vmg: The `inarg.makeSignedIntegral()` line is superfluous by the way. You can remove it.

Weijun-H:
> I don't think your `Ceil` implementation is quite correct. Here's a big hint: https://github.com/shopspring/decimal/blob/f55dd564545cec84cf84f7a53fb3025cdbec1c4f/decimal.go#L1288-L1305

Hi @vmg, I have modied `Ceil` function, could you have a look?

vmg:
We'll need to ignore the bad `typeof` results from the tests. :+1:

vmg (approved):

vmg:
Great job on your first PR! It wasn't an easy one!

PR #10558, opened 2022-06-21
Backup/Restore: add support for external compressors and decompressors

rsajwani:
## Description
Replacement PR for #7978
Signed-off-by: Renan Rangel <renan@slack-corp.com>

This change will allow us to use an external decompressor when available as opposed to the internal pgzip currently used. We ran some tests using pigz -d -c and we get a nice speed bump on decompression, around 30% or more faster decompressing when restoring from a backup in S3. While decompressing is still single-threaded in pigz, it uses separate threads for reading, writing and checksum, and it results in faster performance.

I wanted to get an initial feedback on this change, as we would like to also to plug our own compressor to use other encryption algorithms (like lz4 or zstd) with an external binary as this is a more flexible setup for users, but also work on the change to add built-in support for one of these compression algorithms.

## Related Issue(s)
https://github.com/vitessio/vitess/issues/7802

## Checklist

-   [ ] "Backport me!" label has been added if this change should be backported
-   [x] Tests were added or are not required
-   [ ] Documentation was added or is not required - separate website PR is planned

PR #7037, opened 2020-11-14
backup: Use pargzip instead of pgzip for compression.

enisoc:
Use pargzip instead of pgzip for compression to fix #5613. We still use pgzip for decompression because we haven't seen problems in that phase.

PR #11029, opened 2022-08-17
revert default compression engine

rvrangel:
Signed-off-by: Renan Rangel <rrangel@slack-corp.com>

## Description

As part of https://github.com/vitessio/vitess/pull/10558, it seems we have changed the default compression engine from `pargzip` to `pgzip`. We can see [here](https://github.com/vitessio/vitess/pull/10558/files#diff-ac82970a17868e6ae810d9e2ea6908f3bea25da4d013231910a7d6e3acbedc88L501-R514) that this was the compressor used before. Currently we are defaulting to `pgzip` and that is not intended as it was moved from `pgzip` => `pargzip` in https://github.com/vitessio/vitess/pull/7037

This will make sure the default engine is `pargzip` again, as there is a considerable difference on resource usage when using the other engine.

## Checklist

### DIFF
--- a/doc/releasenotes/15_0_0_summary.md
+++ b/doc/releasenotes/15_0_0_summary.md
@@ -107,8 +107,8 @@ use-case. Here are the flags that control this feature
 
 `--compression-engine-name` specifies the engine used for compression. It can have one of the following values
 
-- pgzip (Default)
-- pargzip
+- pargzip (Default)
+- pgzip
 - lz4
 - zstd
 - external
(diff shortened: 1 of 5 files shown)


deepthi (approved):
LGTM. You will need to change the expected output for flags_test

rsajwani (approved):

rsajwani (dismissed):
You need to fix the test case failures

deepthi:
> Run endtoend tests on Cluster (21)
> isn't reporting any status .. not sure why .. it has backup/restore tests ... I think we should run them before sign-off..

Looks like there might have been a problem with GH Actions.
@rvrangel the only way to kick it is to make an empty commit and push again.

PR #11030, opened 2022-08-17
Move vtbackup_transform test to self hosted

rsajwani:

## Description

vtbackup_transform is flaky. Moving it to self hosted to be able to debug it.

## Checklist


deepthi:
Fixed in #11352 so we can close this.

PR #11032, opened 2022-08-18
Add workflow file to the filter rules

GuptaManan100:
## Description

This PR adds the workflow file to the filtering rules of the workflows so that the test runs even when there are changes to the workflow file and no other changes.

## Checklist


frouioui (approved):
Looks good to me, thank you!

PR #11033, opened 2022-08-18
Add upgrade-downgrade tests for next releases

GuptaManan100:
## Description

This PR adds the next-release counterparts of the current upgrade-downgrade tests. Currently, we have tests that test the upgrade process from the previous release to the current one. For example on `main` the tests run against release-14.0. On release-14.0 they run against 13.0.2 and so on.
But it is essential to also run the tests against the next release. For example, if we introduce a change in release-13.0 which would prevent it from upgrading to 14.0.1, we won't catch it at all until a new patch release is done, in which case it would fail in a PR based against release-14.0

To also catch these scenarios, we have chosen to add upgrade-downgrade tests to test against the next release apart from just the previous release. For `main`, the tests will just be skipped. For release-14.0, `main` will be used, for release-13.0, `14.0.1` will be used and so on.

## Checklist

### DIFF
--- a/tools/get_next_release.sh
+++ b/tools/get_next_release.sh
@@ -14,10 +14,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# This script is used to build and copy the Angular 2 based vtctld UI
-# into the release folder (app) for checkin. Prior to running this script,
-# bootstrap.sh and bootstrap_web.sh should already have been run.
-
 # github.base_ref $1
 
 target_release=""
--- a/tools/get_next_release.sh
+++ b/tools/get_next_release.sh
@@ -1,0 +1,44 @@
+#!/bin/bash
+
+# Copyright 2022 The Vitess Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This script is used to build and copy the Angular 2 based vtctld UI
+# into the release folder (app) for checkin. Prior to running this script,
+# bootstrap.sh and bootstrap_web.sh should already have been run.
+
+# github.base_ref $1
+
+target_release=""
+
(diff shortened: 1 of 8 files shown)


tools/get_next_release.sh
```diff
@@ -0,0 +1,44 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This script is used to build and copy the Angular 2 based vtctld UI
+# into the release folder (app) for checkin. Prior to running this script,
+# bootstrap.sh and bootstrap_web.sh should already have been run.
```
frouioui, line 19: nit: we can remove this

GuptaManan100: Oh yes, that seems wrong!

frouioui:
How about reusing the already existing workflows and changing them to run using a dynamic matrix that gets created in the `get_next_release`/`get_previous_release` job. The next job, `upgrade_downgrade_test`, can run using this matrix.

A similar use case is used in `release-12.0`: https://github.com/vitessio/vitess/blob/e00c46a6380d21c318ec19737fa16d438326e1fa/.github/workflows/cluster_endtoend_upgrade.yml#L43-L45

GuptaManan100:
@frouioui I thought of doing that, but then I thought that it would serialize the workflows and increase its total running time a lot. I think separate workflows is fine since each of them anyways runs for like 40-50 minutes.

tools/get_next_release.sh
GuptaManan100: Fixed

GuptaManan100:
It can be seen in the backport PRs to release-14.0 https://github.com/vitessio/vitess/pull/11034 and release-13.0 https://github.com/vitessio/vitess/pull/11040, the tests are working as expected

frouioui:
> @frouioui I thought of doing that, but then I thought that it would serialize the workflows and increase its total running time a lot. I think separate workflows is fine since each of them anyways runs for like 40-50 minutes.

@GuptaManan100, the jobs of a matrix can be parallelized: https://docs.github.com/en/actions/using-workflows/reusing-workflows. The PR is looking good to me anyway, feel free to merge or use the matrix thingy.

frouioui (approved):

GuptaManan100:
@frouioui I agree the matrix approach would be better because there will be fewer workflows to maintain, but since the workflows already work and backports are ready, I'm going to merge this PR. We can refactor the workflows later when we want to add more tests.

PR #11034, opened 2022-08-18
Add upgrade-downgrade tests for next releases (release-14.0)

GuptaManan100:
## Description

This is a backport of https://github.com/vitessio/vitess/pull/11033.

A couple of test fixes are also made in this PR to make the tests pass.
The first is the moving of `wait_for_drain_sleep_rdonly` deprecated flag to only tests that require it which don't run in the upgrade-downgrade suite.
The second fix is in RepllicationThreadState check to also take into account that the test will also run with v15 tablets.

## Checklist


frouioui (approved):

go/test/endtoend/reparent/utils/utils.go
```diff
@@ -762,8 +762,10 @@ func ReplicationThreadsStatus(t *testing.T, status *replicationdatapb.Status, vt
 		// and should agree with the new ones
 		require.NotEqual(t, mysql.ReplicationStateUnknown, mysql.ReplicationState(status.IoState))
 		require.NotEqual(t, mysql.ReplicationStateUnknown, mysql.ReplicationState(status.SqlState))
-		require.Equal(t, status.IoThreadRunning, mysql.ReplicationState(status.IoState) == mysql.ReplicationStateRunning || mysql.ReplicationState(status.IoState) == mysql.ReplicationStateConnecting)
-		require.Equal(t, status.SqlThreadRunning, mysql.ReplicationState(status.SqlState) == mysql.ReplicationStateRunning || mysql.ReplicationState(status.SqlState) == mysql.ReplicationStateConnecting)
+		if vttabletVersion == 14 {
```
rsajwani, line 765: are they specific to 14 only or should it be >= 14

rsajwani (approved):
LGTM

go/test/endtoend/reparent/utils/utils.go
GuptaManan100: That is specific to v14, because the fields IoThreadRunning are legacy fields removed in v15+. They are present in v14 only for compatilbilty.

PR #11035, opened 2022-08-18
schemadiff performance improvements

dbussink:
This change includes two commits that address two specific issues. First there is the way that we clone a schema for an `Apply` operation. This is switched to using the already available `sqlparser` clone helpers instead of serializing to string and then parsing again.

The second change is to use the `sqlparser` comparison methods instead of serializing to a string and then comparing.

Both these impact performance significantly when dealing with large schemas. The improvements here show a 4x performance improvement on internal PSDB usage of `schemadiff` when dealing with large schemas. 

## Related Issue(s)

#10203

## Checklist

### DIFF
--- a/go/vt/schemadiff/table.go
+++ b/go/vt/schemadiff/table.go
@@ -300,6 +300,7 @@ func (c *CreateTableEntity) normalize() *CreateTableEntity {
 	c.normalizeUnnamedConstraints()
 	c.normalizeTableOptions()
 	c.normalizeColumnOptions()
+	c.normalizeIndexOptions()
 	c.normalizePartitionOptions()
 	return c
 }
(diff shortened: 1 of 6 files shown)


go/vt/schemadiff/table.go
```diff
@@ -458,6 +464,17 @@ func (c *CreateTableEntity) normalizeColumnOptions() {
 	}
 }
 
+func (c *CreateTableEntity) normalizeIndexOptions() {
```
shlomi-noach, line 467: I supposed this function isn't strictly related to the performance improvements?

shlomi-noach (approved):
Looking good!

go/vt/schemadiff/table.go
dbussink: It is definitely relevant to the changes. This is because the `sqlparser` comparison functions compare the direct AST and there we have these values stored in a non normalized way before this change. This means that specific index options with different casing were seen as a change instead of the same.

Before the `CanonicalString` usage was hiding that, since that would normalize to upcase but if we use the `sqlparser` functions, we need this normalization to be consistent.

It's similar to how we normalize other things though like the type for a column.

PR #11028, opened 2022-08-17
[trace] Migrate `go/trace` to use `pflag` for flag definitions

ajm188:
## Description

Had to do things a bit backwards here to avoid an import cycle between `servenv` and `trace`.

## Related Issue(s)

Closes #10747.

## Checklist

-   [ ] "Backport me!" label has been added if this change should be backported
-   [x] Tests were added or are not required
-   [x] Documentation was added or is not required

PR #11036, opened 2022-08-18
[log] Migrate `go/vt/log` flags to `pflag`

### LINKED ISSUE #10808: Switch flag definitions to be on pflag instead of flag in `package go/vt/log`

Current flags:

```
$ git grep -E "\bflag\.[A-Z]" -- go/vt/log/*.go
        flag.Uint64Var(&glog.MaxSize, "log_rotate_max_size", glog.MaxSize, "size in bytes at which logs are rotated (glog.MaxSize)")
```

ajm188:
## Description

This migrates the flags defined in `go/vt/log` to pflag. Similar to #11028, the flag registration has to happen backwards, and then for double-fun, (a) every binary needs logging and (b) not every binary uses `servenv.ParseFlags(WithArgs)?`, so this one needed some additional special handling, on top of the aforementioned special handling.

## Related Issue(s)

Closes #10808.

## Checklist

### DIFF
--- a/go/cmd/vtorc/main.go
+++ b/go/cmd/vtorc/main.go
@@ -25,6 +25,7 @@ import (
 	_ "github.com/mattn/go-sqlite3"
 	"github.com/spf13/pflag"
 
+	vtlog "vitess.io/vitess/go/vt/log"
 	"vitess.io/vitess/go/vt/orchestrator/app"
 	"vitess.io/vitess/go/vt/orchestrator/config"
 	"vitess.io/vitess/go/vt/orchestrator/external/golib/log"
@@ -101,6 +102,7 @@ func main() {
 	// TODO(ajm188): after v15, remove this pflag hack and use servenv.ParseFlags
 	// directly.
 	fs := pflag.NewFlagSet("vtorc", pflag.ExitOnError)
+	vtlog.RegisterFlags(fs)
 
 	args := append([]string{}, os.Args...)
 	os.Args = os.Args[0:1]
--- a/go/vt/log/log.go
+++ b/go/vt/log/log.go
@@ -22,9 +22,8 @@ limitations under the License.
 package log
 
 import (
-	"flag"
-
 	"github.com/golang/glog"
+	"github.com/spf13/pflag"
 )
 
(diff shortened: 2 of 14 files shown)


go/vt/log/log.go
```diff
@@ -73,6 +72,11 @@ var (
 	FatalDepth = glog.FatalDepth
 )
 
-func init() {
-	flag.Uint64Var(&glog.MaxSize, "log_rotate_max_size", glog.MaxSize, "size in bytes at which logs are rotated (glog.MaxSize)")
+// RegisterFlags installs log flags on the given FlagSet.
+//
+// `go/cmd/*` entrypoints should either use servenv.ParseFlags(WithArgs)? which
```
deepthi, line 77: Why the `?`

go/cmd/vtorc/main.go
```diff
@@ -101,6 +102,7 @@ func main() {
 	// TODO(ajm188): after v15, remove this pflag hack and use servenv.ParseFlags
 	// directly.
 	fs := pflag.NewFlagSet("vtorc", pflag.ExitOnError)
+	vtlog.RegisterFlags(fs)
```
deepthi, line 105: @GuptaManan100 @rsajwani IIUC this should go away when we change vtorc to start using servenv.

deepthi (approved):
Wow, we have a lot of binaries that don't use servenv.
I had to review this one commit at a time, but LGTM.

go/cmd/vtorc/main.go
GuptaManan100: I would think yes, since serenv already has these flags registered

go/vt/log/log.go
ajm188: regex: ParseFlags + ParseFlagsWithArgs

go/cmd/vtorc/main.go
ajm188: yes, anyone that uses `servenv` to parse flags (WithArgs or not) doesn't need to do this manually

GuptaManan100 (approved):

rsajwani (commented):
LGTM

PR #11037, opened 2022-08-18
System Settings connections pool implementation

harshit-gangal:
## Description

This PR adds a separate channel to keep the connections which has different settings from the default connection settings and on a get it will return the connection with the correct settings. 
On a put the connection will go to the right channel.
Similarly other methods like closeIdleConnection, refresh pool, etc. are modified to accomodate the settings channel.

This is an initial implementation and there will be further modiffication to make the implementation better
1. Currently the resource is closed if the settings mismatch on the existing connection. This will be changed to reset old settings and apply new settings
2. Add metrics around the new channel usage

This PR will atleast for now enable to intergrate other parts of the query serving flow in vttablet.

## Related Issue(s)

- #9706 

## Checklist

### DIFF
--- a/go/pools/resource_pool.go
+++ b/go/pools/resource_pool.go
@@ -237,17 +237,27 @@ func (rp *ResourcePool) get(ctx context.Context) (resource Resource, err error) 
 	// Fetch
 	var wrapper resourceWrapper
 	var ok bool
+	// If we put both the channel together, then, go select can read from any channel
+	// this way we guarantee it will try to read from the channel we intended to read it from first
+	// and then try to read from next best available resource.
 	select {
+	// check normal resources first
 	case wrapper, ok = <-rp.resources:
 	default:
-		startTime := time.Now()
 		select {
-		case wrapper, ok = <-rp.resources:
+		// then checking setting resources
 		case wrapper, ok = <-rp.settingResources:
-		case <-ctx.Done():
-			return nil, ErrTimeout
+		default:
+			// now waiting
+			startTime := time.Now()
+			select {
+			case wrapper, ok = <-rp.resources:
+			case wrapper, ok = <-rp.settingResources:
+			case <-ctx.Done():
+				return nil, ErrTimeout
+			}
+			rp.recordWait(startTime)
 		}
-		rp.recordWait(startTime)
 	}
 	if !ok {
 		return nil, ErrClosed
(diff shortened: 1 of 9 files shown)


go/pools/rp_bench_test.go
```diff
@@ -102,3 +102,34 @@ func BenchmarkGetPutWithSettings(b *testing.B) {
...
+				if err := resource.ApplySettings(ctx, tcase.settings); err != nil {
+					b.Error(err)
+				}
+				if resource.SettingHash() == 0 {
+					b.Error("setting hash is 0")
+				}
+			}
+		})
+	}
+}
```
harshit-gangal, line 135: Result

| Run | Split | Combined | Improvement(%) |
| ----------- | ----------- | ---------- | --------- |
| 1 Query | 4066 306456 ns/op | - | - |
| 2 Query | 3450	    319346 ns/op | 4018	    287157 ns/op | 16 |
| 3 Query | 3303	    360246 ns/op | 4064	    296723 ns/op | 23 |
| 4 Query | 2995	    394188 ns/op | 4018	    302197 ns/op | 34 |

More the number system settings getting changed, higher the improvement.

go/vt/vttablet/endtoend/reserve_test.go
```diff
@@ -923,3 +923,56 @@ func killConnection(t *testing.T, connID string) {
 	require.NoError(t, err)
 	defer client.Release()
 }
+
+func BenchmarkPreQueries(b *testing.B) {
```
harshit-gangal, line 927: https://github.com/vitessio/vitess/pull/11037#discussion_r950039522

go/pools/resource_pool.go
```diff
@@ -205,34 +222,110 @@ func (rp *ResourcePool) reopen() {
...
+			rp.recordWait(startTime)
 		}
-		rp.recordWait(startTime)
 	}
 	if !ok {
 		return nil, ErrClosed
 	}
 
+	// Checking setting hash id, if it is different, we will close the resource and return a new one later in unwrap
+	if wrapper.resource != nil && wrapper.resource.SettingHash() > 0 && wrapper.resource.SettingHash() != settingHash {
```
systay, line 323: shouldn't we check if the settings are actually equal here? just checking the hash could lead to situations where to different settings yield the same hash code, right?

go/pools/resource_pool.go
```diff
@@ -311,7 +435,14 @@ func (rp *ResourcePool) SetCapacity(capacity int) error {
 
 	if capacity < oldcap {
 		for i := 0; i < oldcap-capacity; i++ {
-			wrapper := <-rp.resources
+			var wrapper resourceWrapper
```
systay, line 438: I would love a comment that explains the following code. It looks clever and interesting, but not obvious

systay (changes requested):
Squashing multiple strings into a single uint64 is bound to be lossy. Probably hard to find clashes, but still, we should compare the settings arrays and not just trust the hash

go/pools/resource_pool.go
harshit-gangal: in the next iteration will get rid hash itself and the settings will become []string to just string. so comparison will be simpler.

go/vt/vttablet/tabletserver/connpool/dbconn.go
```diff
@@ -329,15 +328,30 @@ func (dbc *DBConn) ApplySettings(ctx context.Context, settings []string) error {
...
 }
 
-func (dbc *DBConn) SettingHash() uint64 {
-	return dbc.settingsHash
+func (dbc *DBConn) IsSettingsApplied() bool {
+	return len(dbc.settings) > 0
+}
+
+func (dbc *DBConn) IsSameSetting(settings []string) bool {
+	return compareStringSlice(settings, dbc.settings)
```
systay, line 342: `slices.Equal(settings, dbc.settings)` should be enough

harshit-gangal: I could not find such library :(

systay (approved):

go/vt/vttablet/tabletserver/connpool/dbconn.go
systay: ah, it's an experimental package. Didn't realise it.
`golang.org/x/exp/slices`

PR #11041, opened 2022-08-19
schemadiff: Shallow copy of the schema

dbussink:
Since each entity `Alter()` operation already returns a copy itself of the entity that is modified, it is enough to only shallow copy the schema.

The shallow copy includes the references to the tables, views etc. and will return an already correct new schema when creating or dropping entities with no further issues.

When executing an `Apply()` on a specific entity, that already returns a clone of that entity itself so it also ensures that the new returned value is only changed and the original is unmodified.

This avoids a huge deal of copying when working with large schemas which greatly improves performance there. We're seeing another 6x improvement here in dealing with some large schemas.

## Related Issue(s)

#10203

## Checklist

### DIFF
--- a/go/vt/schemadiff/schema.go
+++ b/go/vt/schemadiff/schema.go
@@ -143,8 +143,8 @@ func getViewDependentTableNames(createView *sqlparser.CreateView) (names []strin
 // normalize is called as part of Schema creation process. The user may only get a hold of normalized schema.
 // It validates some cross-entity constraints, and orders entity based on dependencies (e.g. tables, views that read from tables, 2nd level views, etc.)
 func (s *Schema) normalize() error {
-	s.named = map[string]Entity{}
-	s.sorted = []Entity{}
+	s.named = make(map[string]Entity, len(s.tables)+len(s.views))
+	s.sorted = make([]Entity, 0, len(s.tables)+len(s.views))
 	// Verify no two entities share same name
 	for _, t := range s.tables {
 		name := t.Name()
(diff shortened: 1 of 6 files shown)


go/vt/schemadiff/schema.go
```diff
@@ -143,8 +143,8 @@ func getViewDependentTableNames(createView *sqlparser.CreateView) (names []strin
 // normalize is called as part of Schema creation process. The user may only get a hold of normalized schema.
 // It validates some cross-entity constraints, and orders entity based on dependencies (e.g. tables, views that read from tables, 2nd level views, etc.)
 func (s *Schema) normalize() error {
-	s.named = map[string]Entity{}
-	s.sorted = []Entity{}
+	s.named = make(map[string]Entity, len(s.tables)+len(s.views))
+	s.sorted = make([]Entity, 0, len(s.tables)+len(s.views))
```
dbussink, line 147: Pre-allocating these helps significantly with large schemas since they otherwise see a lot of resizes. And we know the size we expect here, so pre-allocating makes sense.

go/vt/schemadiff/schema.go
```diff
@@ -174,7 +174,7 @@ func (s *Schema) normalize() error {
 	// We actually prioritise all tables first, then views.
 	// If a view v1 depends on v2, then v2 must come before v1, even though v1
 	// precedes v2 alphabetically
-	dependencyLevels := map[string]int{}
+	dependencyLevels := make(map[string]int, len(s.tables)+len(s.views))
```
dbussink, line 177: Same here, we can pre-allocate this as well.

go/vt/schemadiff/schema.go
```diff
@@ -204,25 +204,25 @@ func (s *Schema) normalize() error {
 	// - etc.
 	// we stop when we have been unable to find a view in an iteration.
 	for iterationLevel := 1; ; iterationLevel++ {
-		handledAnyViewsInItration := false
+		handledAnyViewsInIteration := false
```
dbussink, line 207: Typo in variable name :smile:.

shlomi-noach (approved):
Awesome!

PR #11044, opened 2022-08-19
[cli] [logutil] Migrate flags defined in `go/vt/logutil` to `pflag`

### LINKED ISSUE #10809: Switch flag definitions to be on pflag instead of flag in `package go/vt/logutil`

Current flags:

```
$ git grep -E "\bflag\.[A-Z]" -- go/vt/logutil/*.go
go/vt/logutil/level.go: threshold := flag.Lookup("stderrthreshold")
go/vt/logutil/purge.go: keepLogsByCtime   = flag.Duration("keep_logs", 0, "keep logs for this long (using ctime) (zero to keep forever)")
go/vt/logutil/purge.go: keepLogsByMtime   = flag.Duration("keep_logs_by_mtime", 0, "keep logs for this long (using mtime) (zero to keep forever)")
go/vt/logutil/purge.go: purgeLogsInterval = flag.Duration("purge_logs_interval", 1*time.Hour, "how often try to remove old logs")
go/vt/logutil/purge.go: f := flag.Lookup("log_dir")
```

ajm188:
## Description

Basically the same general shape of #11036, but this time for `go/vt/logutil`. The other notable callouts:
- We have a `_flag.Lookup` helper which tries to find a pflag Flag globally defined; failing that, a go flag Flag (which we convert to a pflag Flag).
- The vttestserver Once wrapper is now generic to flag names, so it'll be less cumbersome to keep up to date in the future.

## Related Issue(s)

Closes #10809

## Checklist

### DIFF
--- a/go/cmd/mysqlctl/mysqlctl.go
+++ b/go/cmd/mysqlctl/mysqlctl.go
@@ -252,6 +252,7 @@ func main() {
 	dbconfigs.RegisterFlags(dbconfigs.Dba)
 	fs := pflag.NewFlagSet("mysqlctl", pflag.ExitOnError)
 	log.RegisterFlags(fs)
+	logutil.RegisterFlags(fs)
 	_flag.Parse(fs)
 
 	tabletAddr = netutil.JoinHostPort("localhost", int32(*port))
--- a/go/cmd/query_analyzer/query_analyzer.go
+++ b/go/cmd/query_analyzer/query_analyzer.go
@@ -28,6 +28,7 @@ import (
 
 	"vitess.io/vitess/go/exit"
 	"vitess.io/vitess/go/vt/log"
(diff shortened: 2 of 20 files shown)


rsajwani:
LGTM ... general suggestion and feel free to disregard it.. 
general flow appearing is 

fs := pflag.NewFlagSet("zkctl", pflag.ExitOnError)
log.RegisterFlags(fs)
logutil.RegisterFlags(fs)
_flag.Parse(fs) 

can we wrap it in a method and call that instead everywhere.

rsajwani (approved):

ajm188:
> 

this is a good suggestion! _however_ (😅), really we should have these use `servenv` (which is a whole separate thing) for parsing, which would obviate the point entirely, so i'm actually in favor of making this more "painful" to continue not using `servenv` since that's where we ultimately want to end up

PR #11046, opened 2022-08-19
Fix AST copying of basic types

dbussink:
The code generated for copying basic types in the AST was broken. It would generate code like the following:

```
res := make([]string, 0, len(n))
copy(res, n)
```

The issue is that `copy` uses the logical size of the slice, not the allocated size. This means that it doesn't copy anything at all and that `res` here stays empty.

The change here fixes this issue. It changes the `make` to allocate the logical size as the same value. It then changes the non basic type code to use indexing instead.

That seemed easier to change than having two separate code boths of how the slice is allocated for basic types vs. non basic types.

Do we have a better place to add a test for this except in `schemadiff`? 

## Related Issue(s)

The change https://github.com/vitessio/vitess/pull/11035 more easily exposed this issue because on any schema that uses a `enum` it empties the `enum` list when cloning a table. 

## Checklist

### DIFF
--- a/go/tools/asthelpergen/clone_gen.go
+++ b/go/tools/asthelpergen/clone_gen.go
@@ -93,7 +93,7 @@ func (c *cloneGen) sliceMethod(t types.Type, slice *types.Slice, spi generatorSP
 			// if n == nil { return nil }
 			ifNilReturnNil("n"),
 			//	res := make(Bytes, len(n))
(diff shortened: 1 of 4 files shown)


frouioui (approved):
This looks good to me. I don't see where else to test this, `schemadiff` does not sound like an issue to me :)

PR #11047, opened 2022-08-19
Address additional causes of OnlineDDL test flakiness

mattlord:
## Description

This work makes the following changes:
  - Replaces any `time.Sleep()` calls with a `WaitForMigrationStatus()` call with the appropriate status(es)
  - Use `WaitForMigrationStatus`:Failed,Cancelled for cancel tests
- ~~Use `CheckForStatus`:Failed,Cancelled for cancel tests (instead of just Cancelled)~~
    - As we see both in the CI but we were requiring cancelled
    - See: https://github.com/vitessio/vitess/pull/11011
    - ℹ️ **This was wrong and subsequently removed** -- the test uses user initiated cancel's and the migrations should NOT be marked as failed. We will investigate those scenarios separately.

## Related Issue(s)
  - https://github.com/vitessio/vitess/pull/11011

## Checklist


shlomi-noach (commented):
:+1: for all changes from `time.Sleep(whatever)` to `status := onlineddl.WaitForMigrationStatus()`

:+1: for adding `OnlineDDLStatusFailed` in `status := onlineddl.WaitForMigrationStatus(t, &vtParams, shards, throttledUUID, 20*time.Second, schema.OnlineDDLStatusFailed, schema.OnlineDDLStatusCancelled)`

However, the addition of `schema.OnlineDDLStatusFailed` as a valid terminal state in `onlineddl.CheckMigrationStatus(t, &vtParams, shards, revertUUID, schema.OnlineDDLStatusFailed, schema.OnlineDDLStatusCancelled)` is IMO incorrect; the tests where this state is added are where we actively cancel a migration. The behavior _should be_ that the migration transitions to `cancelled`, not to `failed`. I think we should investigate why that isn't the case in those flaky tests.

mattlord:
> 👍 for all changes from `time.Sleep(whatever)` to `status := onlineddl.WaitForMigrationStatus()`
> 
> 👍 for adding `OnlineDDLStatusFailed` in `status := onlineddl.WaitForMigrationStatus(t, &vtParams, shards, throttledUUID, 20*time.Second, schema.OnlineDDLStatusFailed, schema.OnlineDDLStatusCancelled)`
> 
> However, the addition of `schema.OnlineDDLStatusFailed` as a valid terminal state in `onlineddl.CheckMigrationStatus(t, &vtParams, shards, revertUUID, schema.OnlineDDLStatusFailed, schema.OnlineDDLStatusCancelled)` is IMO incorrect; the tests where this state is added are where we actively cancel a migration. The behavior _should be_ that the migration transitions to `cancelled`, not to `failed`. I think we should investigate why that isn't the case in those flaky tests.

OK, thank you! I wasn't sure about that part. I can share examples of test runs where that occurred if it's helpful. For now I'll remove those changes from the PR.

shlomi-noach:
> I can share examples of test runs where that occurred if it's helpful. 

That will be helpful! Let's see what the error message says.

mattlord:
@shlomi-noach OK, I've gotten rid of the checks for Failed on the user issued REVERTs. Noting that we'll be investigating the test failures where REVERT did end with a failure state separately. Thanks again!

go/vt/vttablet/onlineddl/executor.go
```diff
@@ -1826,17 +1826,16 @@ func (e *Executor) CancelMigration(ctx context.Context, uuid string, message str
...
+			return nil, err
+		}
 	}
 	defer e.failMigration(ctx, onlineDDL, errors.New(message))
 	defer e.triggerNextCheckInterval()
 
 	switch onlineDDL.Status {
 	case schema.OnlineDDLStatusQueued, schema.OnlineDDLStatusReady:
 		log.Infof("CancelMigration: cancelling %s with status: %v", uuid, onlineDDL.Status)
-		if err := e.updateMigrationStatus(ctx, onlineDDL.UUID, schema.OnlineDDLStatusCancelled); err != nil {
```
shlomi-noach, line 1837: these lines have been removed because call to `defer e.failMigration(ctx, onlineDDL, errors.New(message))` already takes care of updating the migration status.

PR #11048, opened 2022-08-19
Flakes: Use waits instead of checks in vrepl e2e tests

mattlord:
## Description

In this PR we replace all of the `time.Sleep()` calls that were waiting for "something" to happen before the next step with usage of a `waitFor...` function that waits for the specific state. We also replace the one time checks for a given state with a wait for that state. This should alleviate some flakiness, with `TestMigrate` having been the worst one. 

## Checklist


rohit-nayak-ps (approved):
lgtm
great work, should help considerably in reducing flakiness!

PR #11059, opened 2022-08-21
schemadiff: Fix handling of primary key

dbussink:
This fixes the handling of primary key options. When dropping a primary key, we should use `drop primary key` and not use the name.

## Related Issue(s)

#10203

## Checklist

### DIFF
--- a/go/vt/schemadiff/table.go
+++ b/go/vt/schemadiff/table.go
@@ -1140,13 +1140,13 @@ func (c *CreateTableEntity) diffKeys(alterTable *sqlparser.AlterTable,
 		t2KeysMap[key.Info.Name.String()] = key
 	}
 
-	dropKeyStatement := func(name sqlparser.IdentifierCI) *sqlparser.DropKey {
+	dropKeyStatement := func(info *sqlparser.IndexInfo) *sqlparser.DropKey {
 		dropKey := &sqlparser.DropKey{}
-		if strings.EqualFold(dropKey.Name.String(), "PRIMARY") {
+		if strings.EqualFold(info.Type, sqlparser.PrimaryKeyTypeStr) {
 			dropKey.Type = sqlparser.PrimaryKeyType
 		} else {
 			dropKey.Type = sqlparser.NormalKeyType
-			dropKey.Name = name
+			dropKey.Name = info.Name
 		}
 		return dropKey
 	}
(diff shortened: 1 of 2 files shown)


go/vt/schemadiff/table.go
```diff
@@ -1140,13 +1140,13 @@ func (c *CreateTableEntity) diffKeys(alterTable *sqlparser.AlterTable,
 		t2KeysMap[key.Info.Name.String()] = key
 	}
 
-	dropKeyStatement := func(name sqlparser.IdentifierCI) *sqlparser.DropKey {
+	dropKeyStatement := func(info *sqlparser.IndexInfo) *sqlparser.DropKey {
 		dropKey := &sqlparser.DropKey{}
-		if strings.EqualFold(dropKey.Name.String(), "PRIMARY") {
+		if strings.EqualFold(info.Type, sqlparser.PrimaryKeyTypeStr) {
```
dbussink, line 1145: This wasn't even checking anything at all. It checked the name of `dropKey` but that was just initialized to a new struct so this check would never be true at all. By passing in the more details from `info` we can check the actual type.

go/vt/schemadiff/table.go
```diff
@@ -1631,7 +1631,15 @@ func (c *CreateTableEntity) apply(diff *AlterTableEntityDiff) error {
 			// we expect the named key to be found
 			found := false
 			switch opt.Type {
-			case sqlparser.NormalKeyType, sqlparser.PrimaryKeyType:
+			case sqlparser.PrimaryKeyType:
```
dbussink, line 1634: This should be handled separately since we want to check the type in this case, not the name. There's only one primary key anyway so we need to find that.

shlomi-noach:
> When dropping a primary key, we should use drop primary key and not use the name.

The statement `alter table tpk1 drop key primary;` is valid in MySQL and is equivalent to `alter table tpk1 drop primary key`. Is there a particular reason why we have to use the 2nd form?

dbussink:
> The statement `alter table tpk1 drop key primary;` is valid in MySQL and is equivalent to `alter table tpk1 drop primary key`. Is there a particular reason why we have to use the 2nd form?

Ah ok, I had no idea tbh that that worked at all 🤦. Never seen the syntax, always see `drop primary key`, I guess that's why I looked at changing it to the second form. 

If we want to keep the existing though, we still have an issue I think since https://github.com/vitessio/vitess/pull/11059#discussion_r950885122 seems like it would be wrong still since `dropKey` is new and it would never match `PRIMARY` there anyway? That additionally confused me and made me think this was wrong.

shlomi-noach:
Yeah, so the name of the `PRIMARY KEY` is `"PRIMARY"` and the name can be used like any other key. Definitely agree that https://github.com/vitessio/vitess/pull/11059#discussion_r950885122 is trivially a bug. Also agree that `DROP PRIMARY KEY` is the more commonly known syntax.

shlomi-noach (approved):

PR #11064, opened 2022-08-22
FlakinessFix: Reparent tests by removing `restore_from_backup`

GuptaManan100:
## Description

Some tests of ERS and PRS have been flaky. The following steps are followed leading to failure -  

1. We bring up a new cluster with new tablets
2. We try to run PRS
3. Meanwhile, tablets go to RESTORE mode as part of the startup
4. PRS sees the tablets are of type RESTORE and doesn’t set semi-sync on them
5. Tablet complete RESTORE and go to type REPLICA (but semi-sync is off). Here initializing logic for replication exists, but that doesn't run since we run our tests with `disable_active_reparents` to prevent the replication manager from running.
6. Test fails assuming semi-sync is on (Blocks on a write)7. 

The proposed fix is to remove the flag `restore_from_backup` from the vttablets which are coming up for the first time. This will prevent the vttablets from going into the RESTORE state at all and everything will work fine from their on.

After this fix, the test was run 222 times without a failure locally.

## Checklist


ajm188 (approved):

rsajwani (commented):
LGTM

Issue #10823, opened 2022-07-25
Bug Report: Gen4 planner generates an incorrect and potentially inefficient query when using a correlated subquery


evaccaro:
### Overview of the Issue

Our use case is a query that is using a correlated subquery between two tables, with an aggregate function in the subquery. The V3 planner handles this type of query correctly (sends the full query to the target shard), but the Gen4 planner breaks the query into two parts, with one query referencing a table that is not part of the query, causing that query to fail completely. 

On the performance side, even if that query would be valid, it would be significantly worse than what is generated by the V3 planner, as for each row that is returned in the first query, we are running the second query that many times (creating an n+1 issue). 

### `vschema.json`

```json
{
  "sharded": true,
  "vindexes": {
    "hash": {
      "type": "hash"
    },
    "deployment_statuses_id_ks_idx": {
      "type": "lookup_test",
      "owner": "deployment_statuses"
    },
    "deployments_id_ks_idx": {
      "type": "lookup_test",
      "owner": "deployments"
    }
  },
  "tables": {
    "deployment_statuses": {
      "column_vindexes": [
        {
          "column": "repository_id",
          "name": "hash"
        },
        {
          "column": "id",
          "name": "deployment_statuses_id_ks_idx"
        },
        {
          "column": "deployment_id",
          "name": "deployments_id_ks_idx"
        }
      ]
    },
    "deployments": {
      "column_vindexes": [
        {
          "column": "repository_id",
          "name": "hash"
        },
        {
          "column": "id",
          "name": "deployments_id_ks_idx"
        }
      ]
    }
  }
}
```

### `Query`

```sql
SELECT `deployments`.`id` FROM `deployments` INNER JOIN `deployment_statuses`
ON `deployment_statuses`.`deployment_id` = `deployments`.`id` WHERE `deployments`.`repository_id` = 176
AND `deployment_statuses`.`id` = (
  SELECT MAX(`deployment_statuses`.`id`) FROM `deployment_statuses` 
  WHERE `deployment_statuses`.`deployment_id` = `deployments`.`id`
)
```

### `V3 Query Plan`

```json
{
  "QueryType": "SELECT",
[...]

PR #11072, opened 2022-08-23
[Gen4] Only remove extracted subquery from predicates on merge

arthurschreiber:
## Description

When merging a subquery into an outer route, all vindex predicates are currently reset and filled again from the outer route's `SeenPredicates`.

https://github.com/vitessio/vitess/blob/3d376eaf12b5fbc6705a5375b1868edf2bcfc2dc/go/vt/vtgate/planbuilder/physical/route.go#L603-L623

For routes of `ApplyJoin` nodes, the `SeenPredicates` list is empty and thus routing information could not be restored correctly.

Instead of resetting the vindex predicates, we can just remove all vindex predicate options for the merged subquery instead, and leave all other options untouched. This ensures that we still can pick any vindexes derived from the join operation, instead of falling back to a `Scatter` route.

## Related Issue(s)

This fixes https://github.com/vitessio/vitess/issues/10823.

## Checklist

### DIFF
--- a/go/vt/vtgate/planbuilder/physical/subquery_planning.go
+++ b/go/vt/vtgate/planbuilder/physical/subquery_planning.go
@@ -82,6 +82,37 @@ func optimizeSubQuery(ctx *plancontext.PlanningContext, op *abstract.SubQuery) (
 	return outerOp, nil
 }
 
+// Remove the given subquery from any predicates and vindex predicates.
+//
+// If a subquery is merged into it's outer query, it needs to be removed from
+// the route as it can't be used as a routing predicate anymore.
+func RemoveExtractedSubqueryPredicate(route *Route, extractedSubquery *sqlparser.ExtractedSubquery) {
+	for _, vp := range route.VindexPreds {
+		idx := -1
+
+		for j, option := range vp.Options {
+			if sqlparser.EqualsExpr(option.Predicates[0], extractedSubquery.Original) {
+				idx = j
+			}
+		}
+
+		if idx != -1 {
+			vp.Options = append(vp.Options[:idx], vp.Options[idx+1:]...)
+		}
+	}
+
+	idx := -1
+	for i, predicate := range route.SeenPredicates {
+		if sqlparser.EqualsExpr(predicate, extractedSubquery) {
+			idx = i
+		}
+	}
+
+	if idx != -1 {
+		route.SeenPredicates = append(route.SeenPredicates[:idx], route.SeenPredicates[idx+1:]...)
+	}
+}
+
 func mergeSubQueryOp(ctx *plancontext.PlanningContext, outer *Route, inner *Route, subq *abstract.SubQueryInner) (*Route, error) {
 	subq.ExtractedSubquery.NeedsRewrite = true
 
@@ -107,10 +138,17 @@ func mergeSubQueryOp(ctx *plancontext.PlanningContext, outer *Route, inner *Rout
 		outer.SysTableTableName[k] = v
 	}
 
-	err = outer.resetRoutingSelections(ctx)
-	if err != nil {
-		return nil, err
+	outer.Selected = nil
+	switch outer.RouteOpCode {
+	case engine.DBA, engine.Next, engine.Reference, engine.Unsharded:
+		// these we keep as is
+	default:
+		outer.RouteOpCode = engine.Scatter
 	}
+
+	RemoveExtractedSubqueryPredicate(outer, subq.ExtractedSubquery)
+	outer.PickBestAvailableVindex()
+
 	return outer, nil
 }
 
--- a/go/vt/vtgate/planbuilder/physical/route.go
+++ b/go/vt/vtgate/planbuilder/physical/route.go
@@ -600,28 +600,6 @@ func (r *Route) planCompositeInOpRecursive(
 	return foundVindex
 }
 
-func (r *Route) resetRoutingSelections(ctx *plancontext.PlanningContext) error {
-	switch r.RouteOpCode {
-	case engine.DBA, engine.Next, engine.Reference, engine.Unsharded:
-		// these we keep as is
-	default:
-		r.RouteOpCode = engine.Scatter
-	}
-
-	r.Selected = nil
-	for i, vp := range r.VindexPreds {
-		r.VindexPreds[i] = &VindexPlusPredicates{ColVindex: vp.ColVindex, TableID: vp.TableID}
-	}
-
-	for _, predicate := range r.SeenPredicates {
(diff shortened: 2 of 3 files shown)


go/vt/vtgate/planbuilder/physical/subquery_planning.go
```diff
@@ -82,6 +82,37 @@ func optimizeSubQuery(ctx *plancontext.PlanningContext, op *abstract.SubQuery) (
...
+//
+// If a subquery is merged into it's outer query, it needs to be removed from
+// the route as it can't be used as a routing predicate anymore.
+func RemoveExtractedSubqueryPredicate(route *Route, extractedSubquery *sqlparser.ExtractedSubquery) {
+	for _, vp := range route.VindexPreds {
+		idx := -1
+
+		for j, option := range vp.Options {
+			if sqlparser.EqualsExpr(option.Predicates[0], extractedSubquery.Original) {
+				idx = j
```
arthurschreiber, line 95: I guess we could `break` here after we found one matching option?

go/vt/vtgate/planbuilder/physical/subquery_planning.go
```diff
@@ -82,6 +82,37 @@ func optimizeSubQuery(ctx *plancontext.PlanningContext, op *abstract.SubQuery) (
...
+// Remove the given subquery from any predicates and vindex predicates.
+//
+// If a subquery is merged into it's outer query, it needs to be removed from
+// the route as it can't be used as a routing predicate anymore.
+func RemoveExtractedSubqueryPredicate(route *Route, extractedSubquery *sqlparser.ExtractedSubquery) {
+	for _, vp := range route.VindexPreds {
+		idx := -1
+
+		for j, option := range vp.Options {
+			if sqlparser.EqualsExpr(option.Predicates[0], extractedSubquery.Original) {
```
arthurschreiber, line 94: This only looks at the first predicate in the list. I think that's correct for all single column vindexes. Do we need to do anything special here for multi column vindexes? I don't think there's any other support for subqueries that match against tuples, so multi column vindex (and thus vindex predicates with more than one value) are not relevant yet.

go/vt/vtgate/planbuilder/physical/subquery_planning.go
```diff
@@ -82,6 +82,37 @@ func optimizeSubQuery(ctx *plancontext.PlanningContext, op *abstract.SubQuery) (
...
+
+		if idx != -1 {
+			vp.Options = append(vp.Options[:idx], vp.Options[idx+1:]...)
+		}
+	}
+
+	idx := -1
+	for i, predicate := range route.SeenPredicates {
+		if sqlparser.EqualsExpr(predicate, extractedSubquery) {
+			idx = i
```
arthurschreiber, line 107: Same here, we could `break` after we found one match.

go/vt/vtgate/planbuilder/physical/subquery_planning.go
```diff
@@ -107,10 +138,17 @@ func mergeSubQueryOp(ctx *plancontext.PlanningContext, outer *Route, inner *Rout
 		outer.SysTableTableName[k] = v
 	}
 
-	err = outer.resetRoutingSelections(ctx)
-	if err != nil {
-		return nil, err
+	outer.Selected = nil
+	switch outer.RouteOpCode {
+	case engine.DBA, engine.Next, engine.Reference, engine.Unsharded:
```
arthurschreiber, line 143: I've taken this list from `resetRoutingSelections`, but I'm not sure we can actually ever reach this code with one of these routing op codes? 🤔

arthurschreiber:
@systay After having talked this through with @evaccaro and @pudiva, there might be a different way of fixing this:

```diff
diff --git a/go/vt/vtgate/planbuilder/physical/route_planning.go b/go/vt/vtgate/planbuilder/physical/route_planning.go
index 8c78dc4667..5ac19a3e94 100644
--- a/go/vt/vtgate/planbuilder/physical/route_planning.go
+++ b/go/vt/vtgate/planbuilder/physical/route_planning.go
@@ -563,6 +563,7 @@ func createRouteOperatorForJoin(aRoute, bRoute *Route, joinPredicates []sqlparse
                VindexPreds:         append(aRoute.VindexPreds, bRoute.VindexPreds...),
                SysTableTableSchema: append(aRoute.SysTableTableSchema, bRoute.SysTableTableSchema...),
                SysTableTableName:   sysTableName,
+               SeenPredicates:      append(aRoute.SeenPredicates, bRoute.SeenPredicates...),
                Source: &ApplyJoin{
                        LHS:       aRoute.Source,
                        RHS:       bRoute.Source,
```

By merging the `SeenPredicates` from both routes of the join into the merged join route, `resetRoutingSelections` would be able to re-create the required Vindex predicate options and pick the correct vindex option for routing.

Which fix do you prefer? 🙇‍♂️

go/vt/vtgate/planbuilder/physical/subquery_planning.go
frouioui: I am unsure too. I am assuming if, for instance, we had a DBA query with a subquery? cc @systay

go/vt/vtgate/planbuilder/physical/subquery_planning.go
frouioui: I think the safest bet would be to check the length of `option.Predicates`, and, only if the length is equal to one, compare the two predicates.

go/vt/vtgate/planbuilder/physical/subquery_planning.go
frouioui: I am unsure whether the extracted subquery can match more than one vindex option or not. @harshit-gangal, do you know?

Otherwise, let's break yes

frouioui (commented):
Looks good to me. I will let one more person ack this and answer my questions.

arthurschreiber:
@frouioui Thank you for reviewing this! 🙇‍♂️ 

I opened https://github.com/vitessio/vitess/pull/11072 which contains an alternative (and I think architecturally more sound) approach to fixing the same issue.

I'm going to close this pull request.

go/vt/vtgate/planbuilder/physical/subquery_planning.go
arthurschreiber: I double checked this. Even if the same inner query exists multiple times in an outer query, each `ExtractedSubquery` object is unique. This is due to the unique name generated for each subquery's replacement bind variable.

I added a comment over at https://github.com/vitessio/vitess/pull/11104/files#diff-ecda8980a2e86d0f2a98ac67bb7b55d759fa7d1e8037dab2497f8fd167ada4c5R120-R121 that describes this.

Issue #7094, opened 2020-12-01
Enable query consolidation only by opt-in


aquarapid:
Since Vitess query consolidation can break normal MySQL read committed guarantees outside of transactions, it can be difficult to strongly verify if an application can handle this before switching query consolidation on.  We can already switch on query consolidation on the replica, but not the master (via `-enable-consolidator-replicas` and `-enable-consolidator=false`).

However, some applications need to send queries to the master that can also benefit from query consolidation.  So one idea is to allow queries to opt into the consolidator by using a "Vitess comment", as we do for custom query timeouts, etc.  In this mode only queries that contain the comment will be eligible for consolidation.

techwolf359:
+1, I can see this being beneficial for cases where you have slow changing data but high rates of reads. Typical caching works well until the cache fails, then you have a thundering herd of traffic to the database. Enabling consolidation on those specific queries would allow you one more layer of defense preventing the saturation on MySQL.

dweitzman:
I think this is a dupe of https://github.com/vitessio/vitess/issues/4058

Susurrus19:
+1. Going to raise this again as something that would be useful in cases where consistency is  paramount. With the flags being enabled in [10907](https://github.com/vitessio/vitess/pull/10907/files) we would like to implement this tech, but would like to do it selectively.

harshit-gangal:
I think we should change the flag default to `--enable-consolidator=false --enable-consolidator-replicas`. This will make consolidator to be enabled only on replica and we can use the new approach to enable query based with #11080

deepthi:
> I think we should change the flag default to `--enable-consolidator=false --enable-consolidator-replicas=true`. This will make consolidator to be enabled only on replica and we can use the new approach to enable query based with #11080 on primary

Changing the defaults will be a breaking change for anyone who is relying on the current behavior. I would not recommend it, especially since it is easy enough for those who want to opt-in to #11080 to change their configuration.

dweitzman:
It's not breaking in the sense that it would cause issues with query results. If it causes issues for anyone, the issue would be increased tablet load.

I don't think the tablet version upgrade agreement promises that people would be able to do a non-incremental rollout across all tablets and expect identical performance to the prior release, so from my perspective at least this doesn't sound like a breaking change

harshit-gangal:
The main concern here is consistency gurantees of a primary when running a query without a transaction is should always return the latest value. With running consolidator by default there can be instances where it can return stale result.

[...]

PR #11080, opened 2022-08-23
opt in/out of query consolidation

maxenglander:
## Description

Since Vitess query consolidation can break normal MySQL read committed
guarantees outside of transactions, it can be difficult to strongly verify if
an application can handle this before switching query consolidation on. We can
already switch on query consolidation on the replica, but not the primary (via
`-enable-consolidator-replicas` and `-enable-consolidator=false`).

However, some applications need to send queries to the primary that can also
benefit from query consolidation.

This PR adds the ability to opt into or out of query consolidation with

    mysql> SELECT /*vt+ CONSOLIDATOR=disabled */ 1 FROM customer LIMIT 1;
    mysql> SELECT /*vt+ CONSOLIDATOR=enabled */ 1 FROM customer LIMIT 1;
    mysql> SELECT /*vt+ CONSOLIDATOR=enabled_replicas */ 1 FROM customer LIMIT 1;

When this directive is set on a query, it overrides the VTTablet-level setting.
This allows disabling query consolidation on the primary by default, and opting
into query consolidation for specific queries. Or, conversely, it allows enabling
query consolidation by queries by default, and opting out for queries that are
known to not behave well when consolidated.

## Details

Some nuances to be aware of.

### Opt-in behavior

If the consolidator is disabled at the VTTablet-level, and the following queries
are issued consecutively...

```
select * from t limit 10001;
select /*vt+ CONSOLIDATOR=enabled */ * from t limit 10001;
select * from t limit 10001;
select /*vt+ CONSOLIDATR=enabled */ * from t limit 10001;
select * from t limit 10001;
```

Only the 4th query is consolidated. Queries without `CONSOLIDATOR=enabled`
are not consolidated, and don't "participate" in consolidation with those that do.

The second query (with `CONSOLIDATOR=enabled`) is served from MySQL because
there is no preceding query participating in consolidation. The fourth query is
consolidated into the second.

### Opt-out behavior

If the consolidator is enabled at the VTTablet-level, and the following queries
are issued consecutively...

```
select * from t limit 10001;
select /*vt+ CONSOLIDATOR=disabled */ * from t limit 10001;
select * from t limit 10001;
select /*vt+ CONSOLIDATOR=disabled */ * from t limit 10001;
select * from t limit 10001;
```

Queries 1, 2 and 4 are served from MySQL. Queries 3 and 5 are consolidated
into query 1.

### Comments about interior comments

It may surprise users to know that, given the following sequence of queries...

```
select * from t limit 10001;
select /* p2 */ * from t limit 10001;
select /* p3 */ * from t limit 10001;
```

All three queries are served from MySQL, none are consolidated. This is behavior
which predates this PR. The changes in this PR are subject to this behavior. This PR
doesn't change this behavior.

## Related Issue(s)

Addresses #7094 

## Checklist


go/mysql/fakesqldb/server.go
```diff
@@ -414,6 +411,17 @@ func (db *DB) HandleQuery(c *mysql.Conn, query string, callback func(*sqltypes.R
 }
 
 func (db *DB) comQueryOrdered(query string) (*sqltypes.Result, error) {
+	var afterFn func() = func() {}
```
maxenglander, line 414: There was only one other use of `OrderMatters()` in the code base. I refactored this to suit the purposes of this PR a bit better. The main change is that when `SetNeverFail(true)` is enabled, this path won't fail when there are non-matching queries, it will advance forward when there are matching queries.

go/mysql/fakesqldb/server.go
```diff
@@ -629,14 +639,6 @@ func (db *DB) AddExpectedExecuteFetch(entry ExpectedExecuteFetch) {
 	db.AddExpectedExecuteFetchAtIndex(appendEntry, entry)
 }
 
-// EnableInfinite turns on the infinite flag (the last ordered query is used).
-func (db *DB) EnableInfinite() {
```
maxenglander, line 633: There weren't any uses of this in the code base.

go/vt/vttablet/tabletserver/query_engine.go
```diff
@@ -176,10 +177,10 @@ func NewQueryEngine(env tabletenv.Env, se *schema.Engine) *QueryEngine {
 	qe.streamConns = connpool.NewPool(env, "StreamConnPool", config.OlapReadPool)
 	qe.consolidatorMode.Set(config.Consolidator)
 	qe.consolidator = sync2.NewConsolidator()
+	qe.streamConsolidator = NewStreamConsolidator(config.ConsolidatorStreamTotalSize, config.ConsolidatorStreamQuerySize, returnStreamResult)
 	if config.ConsolidatorStreamTotalSize > 0 && config.ConsolidatorStreamQuerySize > 0 {
 		log.Info("Stream consolidator is enabled with query size set to %d and total size set to %d.",
 			config.ConsolidatorStreamQuerySize, config.ConsolidatorStreamTotalSize)
-		qe.streamConsolidator = NewStreamConsolidator(config.ConsolidatorStreamTotalSize, config.ConsolidatorStreamQuerySize, returnStreamResult)
```
GuptaManan100, line 182: Is there a reason to move this statement out of the conditional check for ConsolidatorStreamTotalSize and ConsolidatorStreamQuerySize

go/vt/vttablet/tabletserver/query_engine.go
```diff
@@ -56,6 +56,7 @@ import (
 type TabletPlan struct {
 	*planbuilder.Plan
 	Original   string
+	Statement  sqlparser.Statement
```
GuptaManan100, line 59: I am a little unsure about adding the Statement here too since we have the query in *planbuilder.Plan

go/vt/vttablet/tabletserver/query_executor.go
```diff
@@ -99,6 +99,45 @@ func (qre *QueryExecutor) shouldConsolidate() bool {
...
+	}
+
+	d := p.Directives()
+
+	// If none of the comments are directives, default to vttablet-level
+	// settings.
+	if p.Directives() == nil {
+		return qre.shouldConsolidate()
+	}
+
```
harshit-gangal, line 129: this part will not work, from vtgate we do not pass down comments directlive, the only way to receive it is through ExecuteOptions and the problem with that is some default needs to be set on it (eg. false is default), which will defeat the purpose that whether client wants to disable the consolidation or the client have not set the value.

I think we cannot achieve all the cases, I would suggest that choose a path based on the requirement 
i.e. 
client can only enable query consilidation otherwise default of vttablet will be considered.

maxenglander: Hey @harshit-gangal understood about not being able to rely on comments directives being passed down from VTG to VTT. 

Regarding `ExecuteOptions`, could we do what appears to be done with `workload`, where the default value is `UNSPECIFIED`? In that case we would rely on VTT-level settings.

go/vt/vttablet/tabletserver/query_engine.go
maxenglander: I originally tried to use the `FullStmt` available in `*planbuilder.Plan`. Ran into two issues:

 * Many of the `Plan` produced by planbuilder don't include `FullStmt`, including `SELECT` plans.
 * I tried modifying the `planbuilder` code to always include `FullStmt`, but that broken some of the tests we have around plan caching, and I was hesitant about making any changes there.

go/vt/vttablet/tabletserver/query_engine.go
maxenglander: No I think this can be reverted, thanks for catching

go/vt/vttablet/tabletserver/query_executor.go
harshit-gangal: Using enum it can be done.

maxenglander: I updated the PR to parse the comment directive and update the VTGate session options, which are then passed down to VTTablet. Let me know if you think the approach looks right. I will add tests and docs once I know I'm on the right track.

go/vt/sqlparser/comments.go
```diff
@@ -369,3 +373,26 @@ func AllowScatterDirective(stmt Statement) bool {
...
+	directives := comments.Directives()
+	strv, isSet := directives.GetString(DirectiveConsolidator, "")
+	if !isSet {
+		return querypb.ExecuteOptions_CONSOLIDATOR_UNSPECIFIED
+	}
+	if i32v, ok := querypb.ExecuteOptions_Consolidator_value["CONSOLIDATOR_"+strings.ToUpper(strv)]; ok {
+		return querypb.ExecuteOptions_Consolidator(i32v)
+	}
+	return querypb.ExecuteOptions_CONSOLIDATOR_UNSPECIFIED
+}
```
harshit-gangal, line 398: kindly add tests for it in `comments_test.go`

go/vt/vtgate/executor.go
```diff
@@ -985,6 +985,10 @@ func (e *Executor) getPlan(ctx context.Context, vcursor *vcursorImpl, sql string
 	}
 	ignoreMaxMemoryRows := sqlparser.IgnoreMaxMaxMemoryRowsDirective(stmt)
 	vcursor.SetIgnoreMaxMemoryRows(ignoreMaxMemoryRows)
+	consolidator := sqlparser.Consolidator(stmt)
+	if consolidator != querypb.ExecuteOptions_CONSOLIDATOR_UNSPECIFIED {
+		vcursor.SetConsolidator(consolidator)
+	}
```
harshit-gangal, line 991: this needs to be always set irrespective of `ExecuteOptions_CONSOLIDATOR_UNSPECIFIED` otherwise the last value set on the consolidator will continue 
e.g.
Q1: select /*vt+ CONSOLIDATOR=ENABLE */ ...
Q2: select ...
Q3: select  ...

Q2 and Q3 will also continue to consolidate the result.

go/vt/sqlparser/comments.go
maxenglander: Done

go/vt/vtgate/executor.go
maxenglander: Good catch, fixed 😬

maxenglander:
Thanks for the last review @harshit-gangal I think I addressed all of your comments

go/vt/vtgate/vcursor_impl.go
```diff
@@ -794,6 +794,11 @@ func (vc *vcursorImpl) SetPlannerVersion(v plancontext.PlannerVersion) {
 	vc.safeSession.GetOrCreateOptions().PlannerVersion = v
 }
 
+// SetConsolidator implements the SessionActions interface
+func (vc *vcursorImpl) SetConsolidator(consolidator querypb.ExecuteOptions_Consolidator) {
+	vc.safeSession.GetOrCreateOptions().Consolidator = consolidator
```
harshit-gangal, line 799: I would suggest avoiding creating an Option if it does not already exist and the consolidator is unspecified.

harshit-gangal (approved):

harshit-gangal:
DCO is failing, that needs to be fixed.

PR #11082, opened 2022-08-23
Fix incorrect use of loop variable in parallel test

renatolabs:
This commit fixes an occurrence of a loop variable being captured in a
parallel test. With very high probability, only the last test case
will actually be exercised. To work around this problem, we create a
local copy of the range variable before the parallel test, as advised
by the Go documentation at:

https://pkg.go.dev/testing#hdr-Subtests_and_Sub_benchmarks

Issue was found automatically using the `loopvarcapture` linter.

### DIFF
--- a/go/vt/topo/topoproto/tablet_test.go
+++ b/go/vt/topo/topoproto/tablet_test.go
@@ -118,6 +118,10 @@ func TestIsTabletsInList(t *testing.T) {
 	}
 
 	for _, testcase := range testcases {
+		// We create an explicit copy of the range variable for each parallel runner
+		// to be sure that they each run as expected. You can see more information on
+		// this here: https://pkg.go.dev/testing#hdr-Subtests_and_Sub_benchmarks
+		testcase := testcase
 		t.Run(testcase.name, func(t *testing.T) {
 			t.Parallel()
 			out := IsTabletInList(testcase.tablet, testcase.allTablets)


go/vt/topo/topoproto/tablet_test.go
```diff
@@ -118,6 +118,7 @@ func TestIsTabletsInList(t *testing.T) {
 	}
 
 	for _, testcase := range testcases {
+		testcase := testcase // capture range variable
```
mattlord, line 121: Can you please add a comment here as well just to be sure nobody comes along later and removes this "unnecessary" line :) Something like this:
```
// We create an explicit copy of the range variable for each parallel runner
// to be sure that they each run as expected. You can see more information on
// this here: https://pkg.go.dev/testing#hdr-Subtests_and_Sub_benchmarks
// The problem is covered by govet's rangeloop check:
//   https://github.com/golangci/govet/blob/master/rangeloop.go
```

Otherwise it LGTM. Thank you for contributing to Vitess! ❤️

renatolabs: Thanks for reviewing!

I included the comment as you suggested, except for the last two lines, as `vet` does not capture these types of bugs right now (https://github.com/golang/go/issues/16520).

As a side note, the dynamic race detector that ships with Go (`go test -race`) is also not able to detect bugs of this type because of https://github.com/golang/go/issues/35670.

mattlord (approved):
Thanks again! ❤️

PR #11087, opened 2022-08-24
[cli] [tabletserver/vstreamer] Migrate vstreamer's packet size flags to pflags

### LINKED ISSUE #10986: Switch flag definitions to be on pflag instead of flag in `package go/vt/vttablet/tabletserver/vstreamer`

Current flags:

```
$ git grep -E "\bflag\.[A-Z]" -- go/vt/vttablet/tabletserver/vstreamer/*.go
go/vt/vttablet/tabletserver/vstreamer/main_flaky_test.go:       flag.Parse() // Do not remove this comment, import into google3 depends on it
go/vt/vttablet/tabletserver/vstreamer/packet_size.go:var defaultPacketSize = flag.Int("vstream_packet_size", 250000, "Suggested packet size for VReplication streamer. This is used only as a recommendation. The actual packet size may be more or less than this amount.")
go/vt/vttablet/tabletserver/vstreamer/packet_size.go:var useDynamicPacketSize = flag.Bool("vstream_dynamic_packet_size", true, "Enable dynamic packet sizing for VReplication. This will adjust the packet size during replication to improve performance.")
```

rohit-nayak-ps:
## Description

Moved `vstream_packet_size` and `vstream_dynamic_packet_size`  from tabletserver/vstreamer to pflags

## Related Issue(s)

Closes #10986

## Checklist


go/vt/vttablet/tabletserver/vstreamer/main_flaky_test.go
```diff
@@ -38,7 +38,7 @@ var (
 )
 
 func TestMain(m *testing.M) {
-	flag.Parse() // Do not remove this comment, import into google3 depends on it
+	pflag.Parse()
 	ignoreKeyspaceShardInFieldAndRowEvents = true
 	if testing.Short() {
```
rohit-nayak-ps, line 43: @ajm188,  should we just be calling `flag.Parse()` here rather than `pflag.Parse()` so that `testing.Short()` works or is there another way to handle this issue?

```
goroutine 1 [running]:
testing.Short(...)
	/opt/hostedtoolcache/go/1.18.4/x64/src/testing/testing.go:534
vitess.io/vitess/go/vt/vttablet/tabletserver/vstreamer.TestMain(0x442831?)
	/home/runner/work/vitess/vitess/go/vt/vttablet/tabletserver/vstreamer/main_flaky_test.go:43 +0xc7
main.main()
	_testmain.go:159 +0x1d3
FAIL	vitess.io/vitess/go/vt/vttablet/tabletserver/vstreamer	0.021s
FAIL
FAILED (try 2/3) in vitess.io/vitess/go/vt/vttablet/tabletserver/vstreamer (return code 1). See above for errors.
panic: testing: Short called before Parse
```

ajm188: You still need to parse on `flag` (because the `testing` package is super inflexible).

For extra safety, I would suggest doing something like:

```go
var tmp []string
os.Args, tmp = os.Args[0:1], os.Args[1:]
flag.Parse()
os.Args = append(os.Args, tmp...)

// rest of the code
```

ajm188: alternatively (my preference, but also more work) we could rewrite this to not depend on `testing.Short`. up to you!

go/vt/vttablet/tabletserver/vstreamer/packet_size.go
```diff
@@ -17,17 +17,31 @@ limitations under the License.
 package vstreamer
 
 import (
-	"flag"
 	"time"
 
+	"github.com/spf13/pflag"
+
+	"vitess.io/vitess/go/vt/servenv"
+
```
ajm188, line 25: nit: you don't need this newline here

ajm188 (commented):
everything looks good except the `testing.Short` issue

go/vt/vttablet/tabletserver/vstreamer/main_flaky_test.go
rohit-nayak-ps: > alternatively (my preference, but also more work) we could rewrite this to not depend on `testing.Short`. up to you!

What alternative way are you thinking of here?

rohit-nayak-ps: > For extra safety, I would suggest doing something like:
 
 ```go
 var tmp []string
 os.Args, tmp = os.Args[0:1], os.Args[1:]
 flag.Parse()
 os.Args = append(os.Args, tmp...)
 
> // rest of the code
 ```

What is this code protecting against?

ajm188: > What alternative way are you thinking of here?

not sure, it depends entirely on what the goal of the `Short` check was for here, so if you know that'd be useful to know! for example we could do something like `if os.Getenv("SOME_VARIABLE")` instead but that might not make sense for the use case, or maybe `testing.Short()` is _never_ true in our usage and we should just delete it. it's not really possible for me to say

ajm188: > What is this code protecting against?

`os.Args` is literally just the stuff passed to the program, which in this case is `go test ...`, so if there are other flags in there, this lets us ignore them, because we only care about getting `flag.Parsed()` to return true

rohit-nayak-ps: > > What alternative way are you thinking of here?
> 
> not sure, it depends entirely on what the goal of the `Short` check was for here, so if you know that'd be useful to know! for example we could do something like `if os.Getenv("SOME_VARIABLE")` instead but that might not make sense for the use case, or maybe `testing.Short()` is _never_ true in our usage and we should just delete it. it's not really possible for me to say

For now changed it to flag.Parse() to check if tests pass, which they do. I will bring up the testing.Short() usage in Open Hour. I don't see it getting used in the makefiles or ci workflows anymore, so maybe we can remove that usage entirely.

ajm188 (approved):

PR #11088, opened 2022-08-24
No reserved connection on modifying system settings

harshit-gangal:
## Description

This PR integrates the Reserve APIs of vttablet with using Settings Pool without reserving a connection for each session.
This is kept behind the flag `enable_settings_pool` on vttablet.

VTGate code is written in such a way that it does not warant any changes there and everything works. The end to end test on vtgate confirms that it works as expected.

## Related Issue(s)

 - #11037 
 - #9706 

## Checklist


go/flags/endtoend/vtctld.txt
```diff
@@ -164,6 +164,7 @@ Usage of vtctld:
       --queryserver-config-txpool-waiter-cap int                         query server transaction pool waiter limit, this is the maximum number of transactions that can be queued waiting to get a connection (default 5000)
       --queryserver-config-warn-result-size int                          query server result size warning threshold, warn if number of rows returned from vttablet for non-streaming queries exceeds this
       --queryserver_enable_online_ddl                                    Enable online DDL. (default true)
+      --queryserver_enable_settings_pool                                 Enable pooling of connections with modified system settings
       --remote_operation_timeout duration                                time to wait for a remote operation (default 30s)
```
frouioui, line 168: We should update the documentation to reflect this change, and the change to vtexplain and vttablet

go/vt/vttablet/tabletserver/planbuilder/plan.go
```diff
@@ -324,31 +308,20 @@ func BuildMessageStreaming(name string, tables map[string]*schema.Table) (*Plan,
...
+	_ = sqlparser.Walk(func(in sqlparser.SQLNode) (bool, error) {
 		lFunc, isLFunc := in.(*sqlparser.LockingFunc)
 		if !isLFunc {
 			return true, nil
 		}
 		if lFunc.Type == sqlparser.GetLock {
-			return false, genError(lFunc)
+			found = true
+			return false, nil
 		}
```
frouioui, line 323: Shouldn't we also include `ReleaseAllLocks`?

frouioui (approved):

go/vt/vttablet/tabletserver/planbuilder/plan.go
harshit-gangal: only get_lock should happen with reserved conn, otherwise it does not matter.

go/flags/endtoend/vtctld.txt
harshit-gangal: I will add the doc

go/vt/vttablet/tabletserver/planbuilder/plan.go
frouioui: got it

harshit-gangal:
Website doc PR: https://github.com/vitessio/website/pull/1145

go/vt/vttablet/tabletserver/planbuilder/plan_test.go
```diff
@@ -106,10 +110,11 @@ func TestPlanPoolUnsafe(t *testing.T) {
...
-			// In Pooled Connection, plan building will fail.
-			plan, err = Build(statement, testSchema, false /* isReservedConn */, "dbName")
-			require.Error(t, err)
-			out := err.Error()
+			// Plan building will not fail, but it will mark that reserved connection is needed.
+			plan, err = Build(statement, testSchema, "dbName")
+			require.NoError(t, err)
+			require.True(t, plan.NeedsReservedConn)
+			out := fmt.Sprintf("%s not allowed without reserved connection", plan.PlanID.String())
 			if out != tcase.output {
```
GuptaManan100, line 118: If there is no error here from the Build step, why are we asserting on a error output stored in the file? The error doesn't come from the codebase, we are checking on something that is present in the test files only

deepthi:
This PR needs a release note. 
We are also violating the new flag convention documented in the review template. New flags should use dashes, not underscores.
> If a new flag is being introduced, review whether it is really needed. The flag names should be clear and intuitive (as far as possible), and the flag's help should be descriptive. Additionally, flag names should use dashes (-) as word separators rather than underscores (_).

go/vt/vttablet/tabletserver/planbuilder/plan_test.go
harshit-gangal: The plan is allowed, but the execution will fail. I felt it makes sense to put the error message in file and recreate here but I think may be it is not important and move this test to place where it actually returns this error.

harshit-gangal:
Will take care of both the comments in a follow up PR

Issue #10825, opened 2022-07-25
Switch flag definitions to be on pflag instead of flag in `package go/vt/sqlparser/goyacc`


ajm188:
Part of https://github.com/vitessio/vitess/issues/10697.

Current flags:

```
$ git grep -E "\bflag\.[A-Z]" -- go/vt/sqlparser/goyacc/*.go
        flag.StringVar(&oflag, "o", "y.go", "parser output")
        flag.StringVar(&prefix, "p", "yy", "name prefix to use in generated code")
        flag.StringVar(&vflag, "v", "y.output", "create parsing tables")
        flag.BoolVar(&lflag, "l", false, "disable line directives")
        flag.BoolVar(&allowFastAppend, "fast-append", false, "enable fast-append optimization")
        flag.Parse()
        if flag.NArg() != 1 {
        infile = flag.Arg(0)
```

PR #10970, opened 2022-08-10
[cli] [go/mysql/collations/...] Migrate all flags to `pflag`

ajm188:
## Description

I decided to migrate the tests/tools as well, just for consistency's sake.

@vmg do you mind checking if you have any other scripts that use `makecolldata` (I did a `git grep` and updated everything I found, but I am not sure if you have a local workflow) still work on this branch? If not, I can back out just those changes and revisit that piece.

## Related Issue(s)

- Closes #10963 
- Closes #10964 
- Closes #10965 

## Checklist

-   [ ] "Backport me!" label has been added if this change should be backported
-   [x] Tests were added or are not required
-   [x] Documentation was added or is not required

PR #11092, opened 2022-08-25
Move goyacc to use pflags package

GuptaManan100:
## Description

This PR moves `package go/vt/sqlparser/goyacc` package to use pflags instead of flags.

## Related Issue(s)

- Fixes https://github.com/vitessio/vitess/issues/10825
- #10697

## Checklist


GuptaManan100 (commented):
@ajm188 Do i need to introduce a flagSet for goyacc too? It is a tool that no one imports, only used for code generation.

ajm188:
> @ajm188 Do i need to introduce a flagSet for goyacc too? It is a tool that no one imports, only used for code generation.

you _could_ but it in this case (because it's standalone, as you pointed out) it's totally optional. for `makecoldata` it looks like I opted to not add a `FlagSet` either: https://github.com/vitessio/vitess/pull/10970/files#diff-65355e8d129c5ec782065c996a75ef3bea46d50021b7dbb2beb9f133baa8423c

go/vt/sqlparser/goyacc/goyacc.go
```diff
@@ -166,11 +167,11 @@ var prefix string // name prefix for identifiers, default yy
...
 func init() {
-	flag.StringVar(&oflag, "o", "y.go", "parser output")
-	flag.StringVar(&prefix, "p", "yy", "name prefix to use in generated code")
-	flag.StringVar(&vflag, "v", "y.output", "create parsing tables")
-	flag.BoolVar(&lflag, "l", false, "disable line directives")
-	flag.BoolVar(&allowFastAppend, "fast-append", false, "enable fast-append optimization")
+	pflag.StringVar(&oflag, "o", "y.go", "parser output")
+	pflag.StringVar(&prefix, "p", "yy", "name prefix to use in generated code")
+	pflag.StringVar(&vflag, "v", "y.output", "create parsing tables")
+	pflag.BoolVar(&lflag, "l", false, "disable line directives")
```
ajm188, line 173: for all these, i would suggest something like:

```go
pflag.StringVarP(&oflag, "output", "o", "y.go", "parser output")
// similar long-opt names for the others (e.g. --name-prefix, --parsing-tables, --disable-line-directives)
```

this let's you do _either_ `goyacc --output "somepath.go"` or `goyacc -o somepath.go` (and for bools, you can jam them together, e.g. (assuming you have `-a` for `--allow-fast-append`) `goyacc -alo "somepath.go"`)

go/vt/sqlparser/Makefile
```diff
@@ -15,7 +15,7 @@
 MAKEFLAGS = -s
 
 sql.go: sql.y
-	go run ./goyacc -fast-append -o sql.go sql.y
+	go run ./goyacc --fast-append --o sql.go -- sql.y
```
ajm188, line 18: you should not need the `--` separator here

GuptaManan100: Yes, it works without it too. When exactly do we need it?

ajm188: `--` stops parsing regardless of what follows. we needed it to transition _onto_ `pflag`, because `pflag` allows positional and flag args to be mixed, whereas `flag` does not (first thing to not start with a `-` is considered the beginning of posargs, which broke anything with a subcommand tree). As an example:

```
./mycmd --foo "bar" subcommand --subflag "subval"
```

is parsed by `flag` into a single flag `--foo`, and `Args()` slice of: `[]string{"subcommand", "--subflag", "\"subval\""}`
but is parsed by `pflag` into flags of `--foo` and `--subflag` and an `Args()` slice of `[]string{"subcommand"}`

because our current subcommands are written on the assumption of `flag` parsing behavior, this would break because the top-level command's FlagSet did not define the `subflag` flag, so we inject the `--` to make it work in either case. since we're now 100% on `pflag`, we don't need it anymore

go/vt/sqlparser/goyacc/goyacc.go
GuptaManan100: Fixed!

go/vt/sqlparser/Makefile
GuptaManan100: That is a great explanation! Thank-you. Fixed.

ajm188 (approved):

Issue #10824, opened 2022-07-25
Switch flag definitions to be on pflag instead of flag in `package go/vt/sqlparser`


ajm188:
Part of https://github.com/vitessio/vitess/issues/10697.

Current flags:

```
$ git grep -E "\bflag\.[A-Z]" -- go/vt/sqlparser/*.go
go/vt/sqlparser/parser.go:      if flag.Parsed() {
go/vt/sqlparser/truncate_query.go:      TruncateUILen = flag.Int("sql-max-length-ui", 512, "truncate queries in debug UIs to the given length (default 512)")
go/vt/sqlparser/truncate_query.go:      TruncateErrLen = flag.Int("sql-max-length-errors", 0, "truncate queries in error logs to the given length (default unlimited)")
```

PR #11010, opened 2022-08-14
[cli] [vtcombo|tests] Migrate `vtcombo` to `pflag` and rewrite tabletconn tests to not need TabletProtocol exported

ajm188:
## Description

2-for-1 special! What it says in the title (I realized I needed to do the vtcombo bits to properly do the rewrite, so I just jammed these together (separate commits though))

## Related Issue(s)

- Closes #11009.
- Closes #10993.

## Checklist

-   [ ] "Backport me!" label has been added if this change should be backported
-   [x] Tests were added or are not required
-   [x] Documentation was added or is not required

PR #11094, opened 2022-08-25
Move sqlparser flags to use pflags

GuptaManan100:
## Description

This PR moves the flags in sqlparser to use pflags package. It is part of https://github.com/vitessio/vitess/issues/10697#movers-guide work. The only flags available here are the ones that control query truncation for logging and UI, so they have been deemed to be required by all the binaries that import the package.

## Related Issue(s)

- https://github.com/vitessio/vitess/issues/10697#movers-guide
- Fixes https://github.com/vitessio/vitess/issues/10824

## Checklist

### DIFF
--- a/go/flags/endtoend/vtgr.txt
+++ b/go/flags/endtoend/vtgr.txt
@@ -62,8 +62,6 @@ Usage of vtgr:
       --scan_repair_timeout duration                                     Time to wait for a Diagnose and repair operation. (default 3s)
       --security_policy string                                           the name of a registered security policy to use for controlling access to URLs - empty means allow all for anyone (built-in policies: deny-all, read-only)
       --service_map StringList                                           comma separated list of services to enable (or disable if prefixed with '-') Example: grpc-queryservice
-      --sql-max-length-errors int                                        truncate queries in error logs to the given length (default unlimited)
-      --sql-max-length-ui int                                            truncate queries in debug UIs to the given length (default 512) (default 512)
       --stats_backend string                                             The name of the registered push-based monitoring/stats backend to use
       --stats_combine_dimensions string                                  List of dimensions to be combined into a single "all" value in exported stats vars
       --stats_common_tags string                                         Comma-separated list of common tags for the stats backend. It provides both label and values. Example: label1:value1,label2:value2
--- a/go/vt/sqlparser/truncate_query.go
+++ b/go/vt/sqlparser/truncate_query.go
@@ -17,16 +17,27 @@ limitations under the License.
 package sqlparser
 
 import (
-	"flag"
+	"github.com/spf13/pflag"
+
+	"vitess.io/vitess/go/vt/servenv"
 )
 
 var (
 	// TruncateUILen truncate queries in debug UIs to the given length. 0 means unlimited.
-	TruncateUILen = flag.Int("sql-max-length-ui", 512, "truncate queries in debug UIs to the given length (default 512)")
+	TruncateUILen int
 
 	// TruncateErrLen truncate queries in error logs to the given length. 0 means unlimited.
-	TruncateErrLen = flag.Int("sql-max-length-errors", 0, "truncate queries in error logs to the given length (default unlimited)")
+	TruncateErrLen int
 )
+
+func registerQueryTruncationFlags(fs *pflag.FlagSet) {
+	fs.IntVar(&TruncateUILen, "sql-max-length-ui", 512, "truncate queries in debug UIs to the given length (default 512)")
+	fs.IntVar(&TruncateErrLen, "sql-max-length-errors", 0, "truncate queries in error logs to the given length (default unlimited)")
+}
+
+func init() {
+	servenv.OnParse(registerQueryTruncationFlags)
+}
 
 func truncateQuery(query string, max int) string {
 	sql, comments := SplitMarginComments(query)
@@ -41,12 +52,12 @@ func truncateQuery(query string, max int) string {
 // TruncateForUI is used when displaying queries on various Vitess status pages
 // to keep the pages small enough to load and render properly
 func TruncateForUI(query string) string {
-	return truncateQuery(query, *TruncateUILen)
+	return truncateQuery(query, TruncateUILen)
 }
 
 // TruncateForLog is used when displaying queries as part of error logs
 // to avoid overwhelming logging systems with potentially long queries and
 // bind value data.
 func TruncateForLog(query string) string {
-	return truncateQuery(query, *TruncateErrLen)
+	return truncateQuery(query, TruncateErrLen)
 }
--- a/go/vt/sqlparser/truncate_query.go
+++ b/go/vt/sqlparser/truncate_query.go
@@ -36,7 +36,15 @@ func registerQueryTruncationFlags(fs *pflag.FlagSet) {
 }
 
 func init() {
-	servenv.OnParse(registerQueryTruncationFlags)
+	servenv.OnParseFor("vtgate", registerQueryTruncationFlags)
+	servenv.OnParseFor("vttablet", registerQueryTruncationFlags)
+	servenv.OnParseFor("vtcombo", registerQueryTruncationFlags)
+	servenv.OnParseFor("vtctld", registerQueryTruncationFlags)
+	servenv.OnParseFor("vtctl", registerQueryTruncationFlags)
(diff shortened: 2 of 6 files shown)


go/vt/sqlparser/truncate_query.go
```diff
@@ -17,17 +17,28 @@ limitations under the License.
...
-	"flag"
+	"github.com/spf13/pflag"
+
+	"vitess.io/vitess/go/vt/servenv"
 )
 
 var (
 	// TruncateUILen truncate queries in debug UIs to the given length. 0 means unlimited.
-	TruncateUILen = flag.Int("sql-max-length-ui", 512, "truncate queries in debug UIs to the given length (default 512)")
+	TruncateUILen int
```
ajm188, line 27: if the default was not the zero-value, you should assign it here:

```go
TruncateUILen = 512
```

also do these need to be exported, or can we do a similar thing to https://github.com/vitessio/vitess/pull/11010

go/vt/sqlparser/truncate_query.go
```diff
@@ -17,17 +17,28 @@ limitations under the License.
...
 
 	// TruncateErrLen truncate queries in error logs to the given length. 0 means unlimited.
-	TruncateErrLen = flag.Int("sql-max-length-errors", 0, "truncate queries in error logs to the given length (default unlimited)")
+	TruncateErrLen int
 )
 
+func registerQueryTruncationFlags(fs *pflag.FlagSet) {
+	fs.IntVar(&TruncateUILen, "sql-max-length-ui", 512, "truncate queries in debug UIs to the given length (default 512)")
```
ajm188, line 34: if you define the variables with their default, you can do:

```go
fs.IntVar(&TruncateUILen, "...", TruncateUILen, "...")
// ditto ErrLen
```

go/vt/sqlparser/truncate_query.go
```diff
@@ -17,17 +17,28 @@ limitations under the License.
...
+	fs.IntVar(&TruncateErrLen, "sql-max-length-errors", 0, "truncate queries in error logs to the given length (default unlimited)")
+}
+
+func init() {
+	servenv.OnParse(registerQueryTruncationFlags)
```
ajm188, line 39: did you mean to install this on _every_ binary?

GuptaManan100: Yes, the function that uses these flags is actually in truncating the logs. So any binary which receives an SQLError and tries to log it will be calling these functions and will need these flags. I ran the command that the merger-guide lists, and these flags are currently present on all the binaries - 
```
 ~/vitess   pflags-sqlparser-fix  go list -f '{{ .ImportPath }}| {{ join .Deps " " }} ' ./go/cmd/... | awk -F"|" '$2 ~ " vitess.io/vitess/go/vt/sqlparser " {print $1}'
vitess.io/vitess/go/cmd/mysqlctl
vitess.io/vitess/go/cmd/mysqlctld
vitess.io/vitess/go/cmd/query_analyzer
vitess.io/vitess/go/cmd/rulesctl
vitess.io/vitess/go/cmd/rulesctl/cmd
vitess.io/vitess/go/cmd/rulesctl/common
vitess.io/vitess/go/cmd/topo2topo
vitess.io/vitess/go/cmd/vtaclcheck
vitess.io/vitess/go/cmd/vtadmin
vitess.io/vitess/go/cmd/vtbackup
vitess.io/vitess/go/cmd/vtbench
vitess.io/vitess/go/cmd/vtclient
vitess.io/vitess/go/cmd/vtctl
vitess.io/vitess/go/cmd/vtctld
vitess.io/vitess/go/cmd/vtctldclient
vitess.io/vitess/go/cmd/vtctldclient/cli
vitess.io/vitess/go/cmd/vtctldclient/command
vitess.io/vitess/go/cmd/vtctldclient/docgen
vitess.io/vitess/go/cmd/vtexplain
vitess.io/vitess/go/cmd/vtgate
vitess.io/vitess/go/cmd/vtgateclienttest
vitess.io/vitess/go/cmd/vtgr
vitess.io/vitess/go/cmd/vtorc
vitess.io/vitess/go/cmd/vttablet
vitess.io/vitess/go/cmd/vttestserver
vitess.io/vitess/go/cmd/zk
```

go/vt/sqlparser/truncate_query.go
GuptaManan100: Fixed

go/vt/sqlparser/truncate_query.go
GuptaManan100: I have done the default change and unexported the variables, albeit only using Getter and Setter functions. Is there some other benefit we get by doing it in the way of flag parsing function registeration?

go/vt/sqlparser/truncate_query.go
ajm188: sure, those _import_ the package, but for example (you can see several of my PRs that describe this), `cmd/zk` imports `go/vt/vtctl` but only to call `DecodeContent`, which doesn't use this flag's value, so it should not be registered or in the help/usage for that command.

unfortunately you need to actually trace the import dependencies to see if "imports the flag" means "needs the flag" or not; this command is just the launch-point for that analysis.

ajm188: (i think that anything with a `/debug/*` http endpoint is probably the one that needs this, with the exception of vtadmin which is doing it's own thing - but you would need to trace the code, in particular how it uses `go/vt/vtexplain`)

GuptaManan100: That is what I thought in the beginning too, but it's not the UI that is troublesome, its the logging part. `TruncateForLog` gets called in `Error` function of `SQLError`. So any binary which can receive a SQLError will need the flags, because it is used in printing the Error. So, `vtgate`, `vtcombo`, `vttablet`, `vtctld`, `vtctl` definitely need it. I am not sure how to check if the others need it or not, because there is not good way to track if they'll receive SQL error in any of there functions. How do you suggest we proceed?
The first way was to add the flag to everything importing it, but that seems too much. The other way is to only add it to the five binaries listed above. Incase any other binary is using it, it they'll just use the default values, but then we'll have to rely on users to tell us if some binary actually needs it and we didn't add it

go/vt/sqlparser/truncate_query.go
ajm188: > Is there some other benefit we get by doing it in the way of flag parsing function registeration?

I don't quite follow this question, are you talking about the default or the un-exporting?

GuptaManan100: un-exporting. I have done it using just a Getter and Sette function. Is that alright?

go/vt/sqlparser/truncate_query.go
GuptaManan100: Okay, @ajm188 following your advice, I looked at all the commands that were receiving a `SQLError` or trying to convert to it after receiving a string error message and I think I have a list of binaries that use the flags - 
```
"vtgate"
"vttablet"
"vtcombo"
"vtctld"
"vtctl"
"vtexplain"
"vtbackup"
"vttestserver"
"vtbench"
```

Pushed changes accordingly

ajm188 (approved):
one minor change, then feel free to merge

go/flags/endtoend/vtgr.txt
```diff
@@ -62,8 +62,6 @@ Usage of vtgr:
       --scan_repair_timeout duration                                     Time to wait for a Diagnose and repair operation. (default 3s)
       --security_policy string                                           the name of a registered security policy to use for controlling access to URLs - empty means allow all for anyone (built-in policies: deny-all, read-only)
       --service_map StringList                                           comma separated list of services to enable (or disable if prefixed with '-') Example: grpc-queryservice
-      --sql-max-length-errors int                                        truncate queries in error logs to the given length (default unlimited)
```
ajm188, line 65: fwiw, i always[^1] run this command anytime i do a flags PR:

```
$ find go/flags/endtoend -iname '*.txt' | xargs basename | cut -d. -f1 | xargs -I{} bash -c "{} --help 2>go/flags/endtoend/{}.txt"
```

[^1]: okay very occasionally i forget

ajm188 (approved):

go/flags/endtoend/vtgr.txt
GuptaManan100: I ran the command and there are no changes. `TestHelpOutput` also passes 👍. Merging the PR now.

PR #11095, opened 2022-08-25
[cli] [tabletserver/vreplication] Migrate vreplication flags to pflags

### LINKED ISSUE #10981: Switch flag definitions to be on pflag instead of flag in `package go/vt/vttablet/tabletmanager/vreplication`

Current flags:

```
$ git grep -E "\bflag\.[A-Z]" -- go/vt/vttablet/tabletmanager/vreplication/*.go
go/vt/vttablet/tabletmanager/vreplication/controller.go:        _          = flag.Duration("vreplication_healthcheck_topology_refresh", 30*time.Second, "refresh interval for re-reading the topology")
go/vt/vttablet/tabletmanager/vreplication/controller.go:        _          = flag.Duration("vreplication_healthcheck_retry_delay", 5*time.Second, "healthcheck retry delay")
go/vt/vttablet/tabletmanager/vreplication/controller.go:        _          = flag.Duration("vreplication_healthcheck_timeout", 1*time.Minute, "healthcheck retry delay")
go/vt/vttablet/tabletmanager/vreplication/controller.go:        retryDelay = flag.Duration("vreplication_retry_delay", 5*time.Second, "delay before retrying a failed workflow event in the replication phase")
go/vt/vttablet/tabletmanager/vreplication/controller.go:        maxTimeToRetryError = flag.Duration("vreplication_max_time_to_retry_on_error", 15*time.Minute, "stop automatically retrying when we've had consecutive failures with the same error for this long after the first occurrence")
go/vt/vttablet/tabletmanager/vreplication/engine.go:var tabletTypesStr = flag.String("vreplication_tablet_type", "in_order:REPLICA,PRIMARY", "comma separated list of tablet types used as a source")
go/vt/vttablet/tabletmanager/vreplication/framework_test.go:    flag.Set("tablet_protocol", "test")
go/vt/vttablet/tabletmanager/vreplication/framework_test.go:    flag.Set("binlog_player_protocol", "test")
go/vt/vttablet/tabletmanager/vreplication/framework_test.go:    flag.Parse() // Do not remove this comment, import into google3 depends on it
go/vt/vttablet/tabletmanager/vreplication/vplayer_flaky_test.go:        flag.Set("vstream_packet_size", "10")
go/vt/vttablet/tabletmanager/vreplication/vplayer_flaky_test.go:        defer flag.Set("vstream_packet_size", "10000")
go/vt/vttablet/tabletmanager/vreplication/vreplicator.go:       relayLogMaxSize  = flag.Int("relay_log_max_size", 250000, "Maximum buffer size (in bytes) for VReplication target buffering. If single rows are larger than this, a single row is buffered at a time.")
go/vt/vttablet/tabletmanager/vreplication/vreplicator.go:       relayLogMaxItems = flag.Int("relay_log_max_items", 5000, "Maximum number of rows for VReplication target buffering.")
go/vt/vttablet/tabletmanager/vreplication/vreplicator.go:       copyPhaseDuration   = flag.Duration("vreplication_copy_phase_duration", 1*time.Hour, "Duration for each copy phase loop (before running the next catchup: default 1h)")
go/vt/vttablet/tabletmanager/vreplication/vreplicator.go:       replicaLagTolerance = flag.Duration("vreplication_replica_lag_tolerance", 1*time.Minute, "Replica lag threshold duration: once lag is below this we switch from copy phase to the replication (streaming) phase")
go/vt/vttablet/tabletmanager/vreplication/vreplicator.go:       vreplicationHeartbeatUpdateInterval = flag.Int("vreplication_heartbeat_update_interval", 1, "Frequency (in seconds, default 1, max 60) at which the time_updated column of a vreplication stream when idling")
go/vt/vttablet/tabletmanager/vreplication/vreplicator.go:       vreplicationExperimentalFlags = flag.Int64("vreplication_experimental_flags", 0x01, "(Bitmask) of experimental features in vreplication to enable")
go/vt/vttablet/tabletmanager/vreplication/vreplicator.go:       vreplicationStoreCompressedGTID = flag.Bool("vreplication_store_compressed_gtid", false, "Store compressed gtids in the pos column of _vt.vreplication")
```

rohit-nayak-ps:
## Description

Updates flag references specified in #10981

## Related Issue(s)

Fixes #10981

## Checklist


go/vt/vttablet/tabletmanager/vreplication/controller.go
```diff
@@ -41,16 +40,6 @@ import (
 	topodatapb "vitess.io/vitess/go/vt/proto/topodata"
 )
 
-var (
```
rohit-nayak-ps, line 44: @deepthi, is it safe to remove the healthcheck flags? They were deprecated several versions earlier but do we want to play safe and leave them in for now?

rohit-nayak-ps: Actually there are several tests that depend on them, so added them back

go/vt/vttablet/tabletmanager/vreplication/flags.go
```diff
@@ -0,0 +1,76 @@
...
+	vreplicationHeartbeatUpdateInterval = 1
+	vreplicationExperimentalFlags       = int64(0x01) // enable vreplicationExperimentalFlagOptimizeInserts by default
+	vreplicationStoreCompressedGTID     = false
+)
+
+func registerVReplicationFlags(fs *pflag.FlagSet) {
+	fs.DurationVar(&retryDelay, "vreplication_retry_delay", retryDelay, "delay before retrying a failed workflow event in the replication phase")
+	fs.DurationVar(&maxTimeToRetryError, "vreplication_max_time_to_retry_on_error", maxTimeToRetryError, "stop automatically retrying when we've had consecutive failures with the same error for this long after the first occurrence")
+
+	// this are the default tablet_types that will be used by the tablet picker to find source tablets for a vreplication stream
```
ajm188, line 48:
```suggestion
	// these are the default tablet_types that will be used by the tablet picker to find source tablets for a vreplication stream
```

go/vt/vttablet/tabletmanager/vreplication/flags.go
```diff
@@ -0,0 +1,76 @@
...
+	// it can be overridden by passing a different list to the MoveTables or Reshard commands
+	fs.StringVar(&tabletTypesStr, "vreplication_tablet_type", tabletTypesStr, "comma separated list of tablet types used as a source")
```
ajm188, line 50: have you looked at `fs.StringSliceVar` ? it might let us delete some extra code for doing the splitting

go/vt/vttablet/tabletmanager/vreplication/flags.go
```diff
@@ -0,0 +1,76 @@
...
+	//		you have too many streams and/or a large source field (lot of participating tables) which generates unacceptable increase in your binlog size
+	fs.IntVar(&vreplicationHeartbeatUpdateInterval, "vreplication_heartbeat_update_interval", vreplicationHeartbeatUpdateInterval, "Frequency (in seconds, default 1, max 60) at which the time_updated column of a vreplication stream when idling")
+	fs.Int64Var(&vreplicationExperimentalFlags, "vreplication_experimental_flags", vreplicationExperimentalFlags, "(Bitmask) of experimental features in vreplication to enable")
+	fs.BoolVar(&vreplicationStoreCompressedGTID, "vreplication_store_compressed_gtid", vreplicationStoreCompressedGTID, "Store compressed gtids in the pos column of _vt.vreplication")
+
+	// deprecated flags (7.0), however there are several e2e tests that still depend on them
+	fs.Duration("vreplication_healthcheck_topology_refresh", 30*time.Second, "refresh interval for re-reading the topology")
+	fs.Duration("vreplication_healthcheck_retry_delay", 5*time.Second, "healthcheck retry delay")
+	fs.Duration("vreplication_healthcheck_timeout", 1*time.Minute, "healthcheck retry delay")
```
ajm188, line 70: we should delete these, they've been "deprecated" for 8 (!!!!) versions, i kinda don't feel bad about it 😂

go/vt/vttablet/tabletmanager/vreplication/framework_test.go
```diff
@@ -79,13 +80,13 @@ func init() {
...
 	binlogplayer.RegisterClientFactory("test", func() binlogplayer.Client { return globalFBC })
-	flag.Set("binlog_player_protocol", "test")
+	pflag.Set("binlog_player_protocol", "test")
 
 	heartbeatRe = regexp.MustCompile(`update _vt.vreplication set time_updated=\d+ where id=\d+`)
 }
 
 func TestMain(m *testing.M) {
-	flag.Parse() // Do not remove this comment, import into google3 depends on it
+	pflag.Parse() // Do not remove this comment, import into google3 depends on it
```
ajm188, line 89: i don't fully understand this; do you?

go/vt/vttablet/tabletmanager/vreplication/framework_test.go
```diff
@@ -79,13 +80,13 @@ func init() {
 	tabletconntest.SetProtocol("go.vt.vttablet.tabletmanager.vreplication.framework_test", "test")
 
 	binlogplayer.RegisterClientFactory("test", func() binlogplayer.Client { return globalFBC })
-	flag.Set("binlog_player_protocol", "test")
+	pflag.Set("binlog_player_protocol", "test")
```
ajm188, line 83: this is racy whether it works or not; see https://github.com/vitessio/vitess/pull/11010 for a deterministic approach to the problem

go/vt/vttablet/tabletmanager/vreplication/vplayer_flaky_test.go
```diff
@@ -1998,8 +1999,8 @@ func TestPlayerIdleUpdate(t *testing.T) {
 
 func TestPlayerSplitTransaction(t *testing.T) {
 	defer deleteTablet(addTablet(100))
-	flag.Set("vstream_packet_size", "10")
-	defer flag.Set("vstream_packet_size", "10000")
+	pflag.Set("vstream_packet_size", "10")
+	defer pflag.Set("vstream_packet_size", "10000")
```
ajm188, line 2003: same comment about raciness here

go/vt/vttablet/tabletmanager/vreplication/flags.go
rohit-nayak-ps: This is not a simple list. We support options like `in_order:REPLICA,PRIMARY`

go/vt/vttablet/tabletmanager/vreplication/flags.go
rohit-nayak-ps: I have pinged @deepthi elsewhere in the issue about this: there are several tests that fail if these are removed.

go/vt/vttablet/tabletmanager/vreplication/flags.go
rohit-nayak-ps: fixed

go/vt/vttablet/tabletmanager/vreplication/framework_test.go
rohit-nayak-ps: Removed the line and tests continue to work. 

googled(!) for google3 and found it is the internal project name for their code repo. I guess this was from the days google was using vitess internally. I had left it in because I didn't want to break any external dependencies.

go/vt/vttablet/tabletmanager/vreplication/framework_test.go
rohit-nayak-ps: I don't know if I have understood the cause of the raciness correctly. Can you take a look at https://github.com/vitessio/vitess/pull/11095/files#diff-56ead0241f68e27b40866e956a40b8335919c438e888db80cb1c8b2f287ef256R73 and see if the approach is correct? Else please help me understand the issue/solution approach better.

ajm188: `pflag` sets the global flag set, and depends on _someone_ having registered the flag there already (which this test was not doing before), so if that were true, this call would work correctly. if this test happened to be run in a different order (before someone called `registerFlags`), then the call would have no effect and the test would either fail (probably with a weird/confusing error) or, worse, pass even if the change introduced a bug

go/vt/vttablet/tabletmanager/vreplication/flags.go
ajm188: then we should probably delete or rewrite those tests, right?

ajm188 (approved):

PR #11096, opened 2022-08-25
return resource back to pool on apply settings failure

harshit-gangal:
## Description

This PR on ApplySettings failure return the resource back to the pool for accounting and avoiding leaking the connection.

## Related Issue(s)

- #9706 

## Checklist

### DIFF
--- a/go/pools/resource_pool.go
+++ b/go/pools/resource_pool.go
@@ -333,6 +333,9 @@ func (rp *ResourcePool) getWithSettings(ctx context.Context, settings []string) 
 
 	if !wrapper.resource.IsSettingsApplied() {
 		if err = wrapper.resource.ApplySettings(ctx, settings); err != nil {
+			// as we are not able to apply settings, we can return this connection to non-settings channel.
+			// TODO: may check the error code to see if it is recoverable or not.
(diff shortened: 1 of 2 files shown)


go/pools/resource_pool.go
```diff
@@ -333,6 +333,9 @@ func (rp *ResourcePool) getWithSettings(ctx context.Context, settings []string)
 
 	if !wrapper.resource.IsSettingsApplied() {
 		if err = wrapper.resource.ApplySettings(ctx, settings); err != nil {
+			// as we are not able to apply settings, we can return this connection to non-settings channel.
+			// TODO: may check the error code to see if it is recoverable or not.
+			rp.resources <- wrapper
```
frouioui, line 338: Is there anything to reset on the wrapper so it can be cleanly reused?

frouioui (approved):

go/pools/resource_pool.go
harshit-gangal: There is nothing here to clean as of now. It comes here only when no settings are applied earlier and now the current settings failed so it can be reused.

frouioui: Got it

PR #11097, opened 2022-08-25
Incremental logical backup and point in time recovery

shlomi-noach:
## Description

This PR introduces incremental backups, towards point in time recovery (which may or may not happen in this PR).

Incremental backups are only available for:

- `builtin` backup method
- Running MySQL GTID

As reminder, `builtin` backup is a full backup taken by shutting down the server and copying over files.
The incremental backup is done by creating a backup of binary log files. The MySQL server is not stopped nor interrupted by this operation.

An incremental backup is taken like so:

```
vtctlclient -- Backup --incremental_from_pos "MySQL56/16b1039f-22b6-11ed-b765-0a43f95f28a3:1-615" zone1-0000000102 
```

In the above we ask Vitess to create an incremental backup that covers `MySQL56/16b1039f-22b6-11ed-b765-0a43f95f28a3:1-615`. Since the backup method is to copy complete binlog files, it's possible that the backup actually starts at an _earlier_ position. Vitess does not impose that the position is at a binary log rotation.

The backup fails if:

- The first available binlog file does not cover required position (ie binlog files have been rotated and purged, and the incremental GTID do not exist anymore)
- No binary log file contains the requested position
- There is nothing to backup (there are no new GTID entries in the binary logs on top of requested position, ie the requested position is either at the end of existing binary logs or even more futuristic).

The backup's manifest has been updated with new fields. Here's a new manifest:
```json
{
  "BackupMethod": "builtin",
  "Position": "MySQL56/16b1039f-22b6-11ed-b765-0a43f95f28a3:1-883",
  "FromPosition": "MySQL56/16b1039f-22b6-11ed-b765-0a43f95f28a3:1-867",
  "Incremental": true,
  "BackupTime": "2022-08-25T12:55:05Z",
  "FinishedTime": "2022-08-25T12:55:05Z",
  "ServerUUID": "1ea0631b-22b6-11ed-933f-0a43f95f28a3",
  "TabletAlias": "zone1-0000000102",
  "CompressionEngine": "pargzip",
  "FileEntries": [
     ..
  ]
}
```

- The above is an incremental backup's manifest. Clearly indicated by `"Incremental": true,`
- "FileEntries" will only list binary logs
- "FromPosition" indicates the first position covered by the backup. It is smaller than or equal to the requested `--incremental_from_pos`.  This value is empty for full backup.
- `ServerUUID` is new and self explanatory, added for convenience
- `TabletAlias` is new and self explanatory, added for convenience

Added some unit tests. WIP to add endtoend tests.

## Related Issue(s)

https://github.com/vitessio/vitess/issues/11227

## Checklist


shlomi-noach:
Milestone progress, as we now have endtoend tests to validate PITR restore.
```
Sep 21 10:24:22 --- PASS: TestIncrementalBackupMysqlctld (306.89s)
Sep 21 10:24:22     --- PASS: TestIncrementalBackupMysqlctld/full_backup (31.30s)
Sep 21 10:24:22     --- PASS: TestIncrementalBackupMysqlctld/first_incremental_backup (1.22s)
Sep 21 10:24:22     --- PASS: TestIncrementalBackupMysqlctld/make_writes,_succeed (1.21s)
Sep 21 10:24:22     --- PASS: TestIncrementalBackupMysqlctld/fail,_no_binary_logs_to_backup (1.19s)
Sep 21 10:24:22     --- PASS: TestIncrementalBackupMysqlctld/make_writes_again,_succeed (1.21s)
Sep 21 10:24:22     --- PASS: TestIncrementalBackupMysqlctld/auto_position,_succeed (1.21s)
Sep 21 10:24:22     --- PASS: TestIncrementalBackupMysqlctld/fail_auto_position,_no_binary_logs_to_backup (1.18s)
Sep 21 10:24:22     --- PASS: TestIncrementalBackupMysqlctld/auto_position,_make_writes_again,_succeed (1.20s)
Sep 21 10:24:22     --- PASS: TestIncrementalBackupMysqlctld/from_full_backup_position (1.21s)
Sep 21 10:24:22     --- PASS: TestIncrementalBackupMysqlctld/MySQL56/da800267-3996-11ed-8c32-0a43f95f28a3:1-57,_5_records (18.53s)
Sep 21 10:24:22     --- PASS: TestIncrementalBackupMysqlctld/MySQL56/da800267-3996-11ed-8c32-0a43f95f28a3:1-54,_2_records (18.49s)
Sep 21 10:24:22     --- PASS: TestIncrementalBackupMysqlctld/MySQL56/da800267-3996-11ed-8c32-0a43f95f28a3:1-55,_3_records (18.50s)
Sep 21 10:24:22     --- PASS: TestIncrementalBackupMysqlctld/MySQL56/da800267-3996-11ed-8c32-0a43f95f28a3:1-56,_4_records (17.56s)
Sep 21 10:24:22     --- PASS: TestIncrementalBackupMysqlctld/MySQL56/da800267-3996-11ed-8c32-0a43f95f28a3:1-58,_6_records (18.54s)
Sep 21 10:24:22     --- PASS: TestIncrementalBackupMysqlctld/MySQL56/da800267-3996-11ed-8c32-0a43f95f28a3:1-53,_1_records (18.45s)
Sep 21 10:24:22     --- PASS: TestIncrementalBackupMysqlctld/PITR (110.06s)
Sep 21 10:24:22     --- PASS: TestIncrementalBackupMysqlctld/remove_full_position_backups (0.01s)
Sep 21 10:24:22     --- PASS: TestIncrementalBackupMysqlctld/MySQL56/da800267-3996-11ed-8c32-0a43f95f28a3:1-53,_1_records#01 (18.44s)
Sep 21 10:24:22     --- PASS: TestIncrementalBackupMysqlctld/MySQL56/da800267-3996-11ed-8c32-0a43f95f28a3:1-55,_3_records#01 (18.45s)
Sep 21 10:24:22     --- PASS: TestIncrementalBackupMysqlctld/MySQL56/da800267-3996-11ed-8c32-0a43f95f28a3:1-54,_2_records#01 (18.44s)
Sep 21 10:24:22     --- PASS: TestIncrementalBackupMysqlctld/MySQL56/da800267-3996-11ed-8c32-0a43f95f28a3:1-58,_6_records#01 (18.56s)
Sep 21 10:24:22     --- PASS: TestIncrementalBackupMysqlctld/MySQL56/da800267-3996-11ed-8c32-0a43f95f28a3:1-56,_4_records#01 (18.48s)
Sep 21 10:24:22     --- PASS: TestIncrementalBackupMysqlctld/MySQL56/da800267-3996-11ed-8c32-0a43f95f28a3:1-57,_5_records#01 (18.53s)
Sep 21 10:24:22     --- PASS: TestIncrementalBackupMysqlctld/PITR-2 (110.90s)
Sep 21 10:24:22 PASS
Sep 21 10:24:22 ok  	vitess.io/vitess/go/test/endtoend/backup/pitr	306.902s
```

shlomi-noach:
The manifest of a full backup now also includes a `PurgedPosition` field, which is the value of MySQL `@@gtid_purged` where applicable. This value is _essential_ to evaluating a recovery path from incremental backups. The reason is that backed up binary logs may have an incomplete view of `Previous-GTIDs`. For example, the primary may have a GTID set of `f3a47136-3993-11ed-9678-0a43f95f28a3:1-53`, while the first binary log on a replica can state `f3a47136-3993-11ed-9678-0a43f95f28a3:3-53` (notice `3-53`). This happens when the replica is reset, starts from scratch with purged GTIDs; it's a valid situation but then causes confusion: are the binary logs actually covering, say, a backup for `...:1-53` or are they missing anything?
Knowing that the full backup was made with purged GTIDs `...:1-2` sets our mind to rest, and validate that the binary log is legitimately following up on that full backup.

shlomi-noach:
# Overview of changes in this PR

This PR is now ready to review. Here's what it provides:

1. Ability to run an incremental backup: a backup of binary logs from either:
  - A given position (GTID), or
	- Position of last good backup ("auto" position)
2. Ability to restore to a specific position (GTID)
3. Additions to the backup manifest
4. Un-explode function calls by passing `RestoreFromBackupRequest` proto
5. `endtoend` tests for incremental backup / PITR

Let's drill down, in constructive order. Please refer to the [RFC](https://github.com/vitessio/vitess/issues/11227) for background on why we doing this.

## Additions to the backup manifest

The following fields are added:

- `Incremental`:    boolean, to identify whether this is a full or incremental backup
- `PurgedPosition`: specific for MySQL, the `gtid_purged` at time of backup. This is an essential information for a point in time restore
- `FromPosition`:   in an incremental backup, this is the position from which the backup applies. the increment is from `FromPosition` (exclusive) and up to `Position` (inclusive)
- `ServerUUID`:     useful information for debugging/analyzing, the UUID of the mysql server being backed up
- `TabletAlias`:    useful information for debugging/analyzing, the alias of the tablet issuing the backup
- `Keyspace`:       useful information for debugging/analyzing
- `Shard`:          useful information for debugging/analyzing

## Running an incremental backup

The `Backup` command now supports `--incremental_from_pos` flag, which can receive a valid position or the value `auto`. For example:

```shell
$ vtctlclient -- Backup --incremental_from_pos "MySQL56/16b1039f-22b6-11ed-b765-0a43f95f28a3:1-615" zone1-0000000102
$ vtctlclient -- Backup --incremental_from_pos "auto" zone1-0000000102
```

When the value is `auto`, the position is evaluated as the last successful backup's `Position`. The idea with incremental backups is to create a contiguous (overlaps allowed) sequence of backups that store all changes from last full backup.

The incremental backup copies binary log files. It does not take MySQL down nor places any locks. It does not interrupt traffic on the MySQL server. The incremental backup copies comlete binlog files. It initially rotates binary logs, then copies anything from the requested position and up to the last completed binary log.

The backup thus does not necessarily start _exactly_ at the requested position. It starts with the first binary log that has newer entries than requested position. It is OK if the binary logs include transactions prior to the equested position. The restore process will discard any duplicates.

Normally, you can expect the backups to be precisely contiguous. Consider an `auto` value: due to the nature of log rotation and the fact we copy complete binlog files, the next incremental backup will start with the first binay log not covered by the previous backup, which in itself copied the one previous binlog file in full. Again, it is completely valid to enter any good position.

The incremental backup fails if it is unable to attain binary logs from given position (ie binary logs have been purged).

The manifest of an incremental backup has a non-empty `FromPosition` value, and a `Incremental = true` value.

## Running a point in time recovery

We call this "point in time" but it is in fact "point in position". The `RestoreFromBackup` command now supports these new flags:

- `--restore_to_pos`: request to restore the server up to the given position (inclusive) and not one step further.
- `--dry_run`:        when `true`, calculate the restore process, if possible, evaluate a path, but exit without actually making any changes to the server.

Examples:

```
$ vtctlclient -- RestoreFromBackup  --restore_to_pos  "MySQL56/16b1039f-22b6-11ed-b765-0a43f95f28a3:1-220" zone1-0000000102
```

The restore process seeks a restore _path_: a sequence of backups (handles/manifests) consisting of one full backup followed by zero or more incremental backups, that can bring the server up to the requested position, inclusive.

The command fails if it cannot evaluate a restore path. Possible reasons:

- there's gaps in the incremental backups
- existing backups don't reach as far as requested position
- all full backups exceed requested position (so there's no way to get into an ealier position)

The command outputs the restore path.

There may be multiple restore paths, the command prefers a path with the least number of backups. This has nothing to say about the amount and size of binary logs involved.

The `RestoreFromBackup  --restore_to_pos` ends with:

- the restored server in intentionally broken replication setup
- tablet type is `DRAINED`

The reasoning is that we want to restore up to a specific point and not a single step further. We therefore do not let the restored server join the replication stream. We also do not want it to serve traffic, though we do want to be able to query it explicitly.

## Testing

Unit tests validate logic such as constructing a restore path from a list of manifests, or finding sequences of binary log files.

We introduce a new `endtoend` test shard: `backup_pitr`. This test validates incremental backups as well as point in time restores.

The test works by running a cluster and taking backups off one of the replicas. We injest data into the database such that we know what data to expect in each backup.

We validate pre-calculated or `auto` positions for incremental backups. We validate invalid positions and failures.

We then validate restores by randomly restoring into any one of our pre-recorded positions. We know what data to expect per restore and validate that. The reason for random restores is so that we validate we can restore into both older and newer positions, and ensure that all existing data is erased.

GuptaManan100 (commented):

go/vt/mysqlctl/binlogs_gtid.go
```diff
@@ -0,0 +1,198 @@
...
+		// if the next manifest is eligible to be part of the path, try it out
+		if IsValidIncrementalBakcup(baseGTIDSet, purgedGTIDSet, remainingManifests[0]) {
+			nextGTIDSet := baseGTIDSet.Union(remainingManifests[0].Position.GTIDSet)
+			findPaths(nextGTIDSet, append(pathManifests, remainingManifests[0]), remainingManifests[1:])
+		}
+		// also, try without the next manifest
+		findPaths(baseGTIDSet, pathManifests, remainingManifests[1:])
+	}
+	// find all paths, entry point
+	findPaths(fullBackup.Position.GTIDSet, sortedManifests[0:1], sortedManifests[1:])
```
GuptaManan100, line 182: Another thing that I noticed from my understanding which might not be a cause for concern now is that this algorithm has a worse-case exponential run-time. For example, if we have a base backup for position [1,50] and we have 10 incremental backups like - [10, 100], [11, 101], [12, 102], [13, 103].... [19, 109] and the users asks us to restore until the position `109`, then all the subsets of the incremental backups containing the start and the final backup are actually valid.
So, the algorithm will end up populating the validRestorePaths array to 2^9 possible paths, only one of which will have the shortest length.

In the future, if we decide to improve the performance specifically for these corner cases, we can use a dynamic programming-based approach. Please disregard this if my understanding of the function is flawed.

shlomi-noach: > then all the subsets of the incremental backups containing the start and the last incremental backup are actually valid.

You're right. In practice I don't see the situation where this happens because typically you will create an incremental backup from the end of the last known good backup. And so typically we expect a linear sequence of backups, or may be two such sequences if backups are taken from two different machines.

Of course reality can always surprise. A dynamic approach sounds good. We can definitely iterate.

go/vt/mysqlctl/binlogs_gtid.go
```diff
@@ -0,0 +1,198 @@
...
+func FindPITRPath(restoreToGTIDSet mysql.GTIDSet, manifests [](*BackupManifest)) (shortestPath [](*BackupManifest), err error) {
+	sortedManifests := make([](*BackupManifest), 0, len(manifests))
+	for _, m := range manifests {
+		if m != nil {
+			sortedManifests = append(sortedManifests, m)
+		}
+	}
+	sort.SliceStable(sortedManifests, func(i, j int) bool {
+		return sortedManifests[j].Position.GTIDSet.Union(sortedManifests[i].PurgedPosition.GTIDSet).Contains(sortedManifests[i].Position.GTIDSet)
+	})
```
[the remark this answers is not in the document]
shlomi-noach: In theory you're probably right that it's fine to compare just the end positions, e.g. if the backup is `3-200` o `5-300` it's probably fine to just compare `200` with `300`. But as a matter of consistency I compare GTIDs in full. Who knows, maybe some backup will have `3-100,117-200`, with a gaping hole, for whatever reason. I just don't want to run into this situation.

The thing with the purged GTIDs is very confusing, to me too. It's entirely possible for a backup to say `100-120` and yet be applicable to `1-99` as well, if these were purged. Two different backups from two different replicas may have different purged values. Two backups taken on two perfectly consistent and identical replicas can say `1-120` and `100-120`, and both will be fine, assuming the 2nd will have a purged GTID value of `1-99`. So, when comparing two backups, you have to support the idea that they were taken on different hosts, which may have different purged GTID values. This is why we do that union.

go/vt/mysqlctl/binlogs_gtid.go
GuptaManan100: Gotcha! 👍

go/vt/mysqlctl/binlogs_gtid.go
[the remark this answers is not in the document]
GuptaManan100: Another thing that I am struggling with is that is it always necessary for two manifests to always be comparable. Like i < j or i > j?

go/vt/mysqlctl/binlogs_gtid.go
shlomi-noach: @GuptaManan100 let me look into it. If things turn to be too complex, then we can always choose the first valid solution, even if suboptimal.

GuptaManan100: Thank-you! 💕

shlomi-noach: @GuptaManan100 in https://github.com/vitessio/vitess/pull/11097/commits/2ea863cd8def8857a06da66f87bdf50dead09e16 we stop the search immediately after finding the first valid restore path, even if it's not the optimal. I've left the rest of the algorithm intact, including the followup `sort` of the restore paths (now there's just one) in case we even want to re-examine.

Tests were adapted (previously validated that shortest path is chosen, now they validate that first path is chosen)

GuptaManan100: Awesome! Thank-you! 👍

go/vt/mysqlctl/binlogs_gtid.go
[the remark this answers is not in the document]
shlomi-noach: @GuptaManan100 missed this comment. Can you please clarify the question? Not sure I follow. Are you asking whether any two manifests should always be comparable, or are you stating that any two manifests should always be comparable? And, in what context?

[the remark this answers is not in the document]
GuptaManan100: I am asking if it is necessary if two manifests are always comparable or not.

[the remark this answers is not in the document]
shlomi-noach: @GuptaManan100 got it. Two manifests can in fact be uncompare-able, and the referenced code does not mind if two are not comparable. But -- there's no explicit test to validate that, so let me add one.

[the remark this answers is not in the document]
shlomi-noach: OK, so there's now a _partial_ validation for this. The thing I'm struggling with is to understand whether I understand the possible situations where GTID sets are non-comparable. The part that confuses me is where purged GTIDs are concerned. I _think_ we're good on that front, but need to consider this more. I suggest that we iterate on that specific issue after we've merged this PR.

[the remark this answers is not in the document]
GuptaManan100: Okay! That sounds good to me!

shlomi-noach:
`Cluster (vttablet_prscomplex)` fails repeatedly, but I see it has nothing to do with this PR.

deepthi:
> `Cluster (vttablet_prscomplex)` fails repeatedly, but I see it has nothing to do with this PR.

Yes, it was broken by #11592. @harshit-gangal is planning to revert that and fix it properly.

go/vt/mysqlctl/backup.go
```diff
@@ -396,10 +404,26 @@ func Restore(ctx context.Context, params RestoreParams) (*BackupManifest, error)
...
+	if handles := restorePath.IncrementalBackupHandles(); len(handles) > 0 {
+		params.Logger.Infof("Restore: applying %v incremental backups", len(handles))
+		for _, bh := range handles {
+			manifest, err := re.ExecuteRestore(ctx, params, bh)
+			if err != nil {
+				return nil, err
+			}
+			// Add backupTime and restorePosition to LocalMetadata
+			params.LocalMetadata["RestoredBackupTime"] = manifest.BackupTime
+			params.LocalMetadata["RestorePosition"] = mysql.EncodePosition(manifest.Position)
```
deepthi, line 416: Does this go into `_vt.local_metadata`? We are about to delete that table because it was being used primary by orchestrator integration (which has now been deleted in main).

go/vt/mysqlctl/backupengine.go
```diff
@@ -193,51 +205,198 @@ type BackupManifest struct {
...
 	FinishedTime string
+
+	// ServerUUID identifies the server from which backup was taken
+	ServerUUID string
+
+	TabletAlias string
+
+	Keyspace string
+
+	Shard string
```
deepthi, line 232: Are the changes to the manifest backwards compatible? I.e. can I take a full backup on v16, downgrade to v15 and restore from that backup?

go/vt/mysqlctl/xtrabackupengine.go
```diff
@@ -156,6 +156,9 @@ func closeFile(wc io.WriteCloser, fileName string, logger logutil.Logger, finalE
 // and an overall error.
 func (be *XtrabackupEngine) ExecuteBackup(ctx context.Context, params BackupParams, bh backupstorage.BackupHandle) (complete bool, finalErr error) {
 
+	if params.IncrementalFromPos != "" {
```
deepthi, line 159: I understand that this doesn't make sense for xtrabackup. Is there a way for us (in a separate PR) to support xtrabackup's own incremental backups?

deepthi (commented):
Mostly looks good, but there are some concerns.

go/vt/mysqlctl/backup.go
deepthi: The tablet record in topo maybe the right place for this. The bigger question is: why do we want to record this information?

shlomi-noach: @deepthi I'm happy to remove these lines -- I only added them for consistency with existing code, and am not using this data anywhere. If the existing code is to be removed, then there's no need for these two lines of code.

go/vt/mysqlctl/backupengine.go
shlomi-noach: @deepthi yes, you can. The added fields are not actually read/used for now (with the exception of PITR-specific data). For now, they're added as means of visibility and simpler means for debugging and tracing "what happened where and by whom".

go/vt/mysqlctl/xtrabackupengine.go
shlomi-noach: Possibly, yes. I haven't looked into it at all.

go/vt/mysqlctl/backup.go
shlomi-noach: Removed.

deepthi (approved):
I think this is almost ready to merge. Just needs a release note, and then we can remove that label and merge.
Nice work!

deepthi:
Not sure which all files require other reviews. We should probably add more code owners for vtadmin proto files - basically duplicate the owners of the proto files.
cc @harshit-gangal

go/vt/mysqlctl/backup.go
```diff
@@ -323,16 +324,23 @@ func Restore(ctx context.Context, params RestoreParams) (*BackupManifest, error)
...
+	bh := restorePath.FullBackupHandle()
 	re, err := GetRestoreEngine(ctx, bh)
 	if err != nil {
 		return nil, vterrors.Wrap(err, "Failed to find restore engine")
 	}
-
+	params.Logger.Infof("Restore: %v", restorePath.String())
+	if params.DryRun {
+		return nil, vterrors.Errorf(vtrpc.Code_CANCELED, "Restore: dry run, aborting operation")
+	}
```
harshit-gangal, line 343: in the dry run case, do we need to return an error?

go/vt/mysqlctl/backupengine.go
```diff
@@ -246,8 +405,25 @@ func FindBackupToRestore(ctx context.Context, params RestoreParams, bhs []backup
 		// up empty.
 		return nil, ErrNoCompleteBackup
 	}
-
-	return bh, nil
+	// Anything taken before the full backup that we picked, is not of interest:
+	manifests = manifests[fullBackupIndex:]
+	path := &RestorePath{
```
harshit-gangal, line 410: nit: `path` collides with import package `path`

go/vt/mysqlctl/builtinbackupengine.go
```diff
@@ -157,13 +177,138 @@ func (fe *FileEntry) open(cnf *Mycnf, readOnly bool) (*os.File, error) {
...
+		return be.executeIncrementalBackup(ctx, params, bh)
+	}
+	return be.executeFullBackup(ctx, params, bh)
+}
+
+// executeIncrementalBackup runs an incremental backup, based on given 'incremental_from_pos', which can be:
+// - A valid position
+// - "auto", indicating the incremental backup should begin with last successful backup end position.
+func (be *BuiltinBackupEngine) executeIncrementalBackup(ctx context.Context, params BackupParams, bh backupstorage.BackupHandle) (bool, error) {
+	if params.IncrementalFromPos == "auto" {
```
harshit-gangal, line 195: nit: can move magic string to a const

harshit-gangal (approved):

go/vt/mysqlctl/backupengine.go
shlomi-noach: renamed as `restorePath`

go/vt/mysqlctl/builtinbackupengine.go
shlomi-noach: done

go/vt/mysqlctl/backup.go
shlomi-noach: It's a way to break through all layers up, without anyone attempting to actually use the backup result. Do you think we should return a `nil` for error?

harshit-gangal: my understanding is dry run is used by the user to see what steps will be taken and does all the pre-condition checks pass or not, before actually running it.

shlomi-noach: The change was made. No longer returning an error, and now exiting gracefully. The log looks something like this:
```
restore.go:202] Restore: original tablet type=RDONLY
backup.go:289] Restore: looking for a suitable backup to restore
backup.go:340] Restore: RestorePath: [full:2022-11-28.063331.zone1-0000000102, incremental:2022-11-28.063340.zone1-0000000102, incremental:2022-11-28.063343.zone1-0000000102, incremental:2022-11-28.063348.zone1-0000000102]
restore.go:230] Restore: got a restore manifest: <nil>, err=<nil>, waitForBackupInterval=0s
restore.go:289] Dry run. No changes made
restore.go:311] Restore: changing tablet type to RDONLY
```

shlomi-noach:
I still need to write release notes, as well as website docs.

harshit-gangal (approved):

shlomi-noach:
Release notes added.

shlomi-noach:
bwahahaha! It is merged!

PR #11102, opened 2022-08-25
Use vtparams instead of clusterInstance in TestNormalizeAllFields

frouioui:
## Description

Use `vtParams.Host` instead of `clusterInstance.Hostname` in `TestNormalizeAllFields` to keep the same format throughout the Gen4 e2e tests.

## Checklist


rsajwani (approved):
LGTM

Issue #6277, opened 2020-06-06
RFC VStream Copy: streaming events from the beginning


rohit-nayak-ps:
## Motivation

Currently the vstream API streams events starting either from the current position of the binlog or from a position specified by the client. The VStream Copy feature adds support to send all events starting from the first position of the binlog.

A naive extension of the current mechanism is to stream from the starting position. However this is 
impractical for any database/table of a reasonable size. We will extend VStream to make use of the bulk copy based mechanism similar to vreplication streams, used in MoveTables or Reshard sharding workflows.

Note that with vstream copy the client vstream will not faithfully reproduce the events from the binlog. The aim is to be eventually (and rapidly) consistent with the current database snapshot. This improves performance since we will be merging multiple row updates into a single transaction.
Once we have caught up (i.e. the replication lag is small) binlog events will again be directly streamed similar to the current implementation.

### Current API

Clients create vstreams by grpc-ing to VTGate using the Vstream API call. In golang:

```
conn, _ := 	VTGate.Dial(ctx, "localhost:15991")
// tabletType is one of replica/master/rdonly, filter,vgtid: see below
reader, _ := 	VStream(ctx, tabletType, vgtid, filter)
e, _  := 	reader.Recv() //receive VEvents in a loop until io.EOF
```

It is possible for network errors to occur or for client process to fail. In addition, the vstreamer itself might fail at VTGate or at VTTablet. Thus, VTGate needs to send state frequently allowing VTGate to be stateless and clients to recover properly from failures. 

Also, while creating the stream the client can specify multiple shards and/or keyspaces from which to stream events.

The vgtid structure facilitates both: determining the stream sources and maintaining state. vgtid is a list of tuples: (keyspace, shard, gtid). When a stream is created, gtid can either be “current” or a valid binlog position at which the vstream starts streaming events.

Some examples:
```
// stream from current position from two shards
vgtid := &binlogdatapb.VGtid{
         		ShardGtids: []*binlogdatapb.ShardGtid{{
         			Keyspace: "ks",
         			Shard:    "-40",
         			Gtid:     "current",
         		},{
         			Keyspace: "ks",
         			Shard:    "80-c0",
         			Gtid:     "current",
         		}}
         }

// stream from specific position from all shards in keyspace ks
vgtid := &binlogdatapb.VGtid{
         		ShardGtids: []*binlogdatapb.ShardGtid{{
         			Keyspace: "ks",
         			Gtid:     "MariaDB/0-41983-20",
[...]

PR #11103, opened 2022-08-26
VSCopy: Resume the copy phase consistently from given GTID and lastpk

yoheimuta:
## Description

The current vstream API doesn't resume a copy phase given the below `VGTID`.
Instead, it starts replicating the table from the given `GTID` while ignoring the `lastpk` information.

```
vgtid {
  shard_gtids {
    keyspace: "test_sharded_keyspace"
    shard: "-80"
    gtid: "MySQL56/ff540f7a-1d4d-11ed-93b5-0242ac110002:1-28"
    table_p_ks {
      table_name: "t1"
      lastpk {
        rows {
          lengths: 4
          values: "1000"
 ...
```

This PR allows vstream consumers to deal with the `VGTID` event as [an opaque offset](https://github.com/vitessio/vitess/issues/6277#issue-632492593).
Given the non-empty GTID and the lastpk, the API is supposed to do the following:

1. Catch up the t1 table from "MySQL56/ff540f7a-1d4d-11ed-93b5-0242ac110002:1-28" to the nearly latest
2. Copy the t1 table with the condition "pk > 1000"

## Related Issue(s)

Discussion: https://vitess.slack.com/archives/C0PQY0PTK/p1661155731380559

### Modify uvstreamer Init
Allow a non-empty position when tablePKs are specified. In this case an initlal catchup+fastforward phase will run before restarting copy phase on incompletely copied tables. (reference: https://vitess.io/docs/design-docs/vreplication/life-of-a-stream/)

### Change how we filter events
Since we now run a catchup phase when we restart table copies, it is possible that duplicate events can be sent for rows that are not already copied, since we don't yet have logic to compare primary keys. To minimise this we filter out events for tables whose copy phase has not yet started.

### Test changes

1. New test TestVStreamCopyResume: passes both tablePKs and a specific position to validate new logic
https://github.com/vitessio/vitess/pull/11103/files?diff=unified&w=1#diff-c98b8f318aac471272bf650c7824713c9664b46895380acf322ec3498a2fe578R236

2. Updated test TestVStreamCopyCompleteFlow to account for change in logic for selecting events to send

## Checklist

-   [x] Documentation PR: https://github.com/vitessio/website/pull/1216

## Pending issues
1.  Filter out events for rows not copied, for tables whose primary keys can be compared in code (essentially those that our eval engine can compare). This is not critical, since clients have to handle duplicate events in any case.
2. Batch the copy phase (like we do for vreplication) so that we don't end up with a large vreplication lag (and possibly rotated binlogs). We didn't expect this to be critical since we are just streaming events, rather than inserting rows on the target, like we do for vreplication. But it is possible that some clients do more cpu/IO-intensive operations than just sending them to event sinks.


go/vt/vtgate/endtoend/vstream_test.go
```diff
@@ -232,6 +232,95 @@ func TestVStreamCopyBasic(t *testing.T) {
...
+	require.NotNil(t, reader)
+	var evs []*binlogdatapb.VEvent
+	for {
+		e, err := reader.Recv()
+		switch err {
+		case nil:
+			for _, ev := range e {
+				if ev.Type == binlogdatapb.VEventType_ROW {
+					evs = append(evs, ev)
+					printEvents(evs) // for debugging ci failures
```
yoheimuta, line 307: [memo] The output should be like the below:

```
===START===
Event 0; type:ROW timestamp:1661421104 row_event:{table_name:"ks.t1" row_changes:{before:{lengths:1 lengths:1 values:"11"} after:{lengths:1 lengths:2 values:"110"}} keyspace:"ks" shard:"-80"} current_time:1661421104807890000 keyspace:"ks" shard:"-80"
Event 1; type:ROW timestamp:1661421104 row_event:{table_name:"ks.t1" row_changes:{after:{lengths:1 lengths:1 values:"99"}} keyspace:"ks" shard:"-80"} current_time:1661421104810565000 keyspace:"ks" shard:"-80"
Event 2; type:ROW row_event:{table_name:"ks.t1" row_changes:{after:{lengths:1 lengths:1 values:"55"}} keyspace:"ks" shard:"-80"} keyspace:"ks" shard:"-80"
Event 3; type:ROW row_event:{table_name:"ks.t1" row_changes:{after:{lengths:1 lengths:1 values:"99"}} keyspace:"ks" shard:"-80"} keyspace:"ks" shard:"-80"
Event 4; type:ROW row_event:{table_name:"ks.t1" row_changes:{after:{lengths:1 lengths:1 values:"66"}} keyspace:"ks" shard:"80-"} keyspace:"ks" shard:"80-"
Event 5; type:ROW row_event:{table_name:"ks.t1" row_changes:{after:{lengths:1 lengths:1 values:"77"}} keyspace:"ks" shard:"80-"} keyspace:"ks" shard:"80-"
Event 6; type:ROW row_event:{table_name:"ks.t1" row_changes:{after:{lengths:1 lengths:1 values:"88"}} keyspace:"ks" shard:"80-"} keyspace:"ks" shard:"80-"
===END===
```

go/vt/vttablet/tabletserver/vstreamer/uvstreamer.go
```diff
@@ -253,9 +253,15 @@ func (uvs *uvstreamer) filterEvents(evs []*binlogdatapb.VEvent) []*binlogdatapb.
 		}
 		if !shouldSend && tableName != "" {
 			shouldSend = true
-			_, ok := uvs.plans[tableName]
+			plan, ok := uvs.plans[tableName]
 			if ok {
-				shouldSend = false
+				// Ideally we should compare the PKs and never send the rows whose PK has already been copied.
+				// For now, we send all the changes by accepting the duplicate events.
```
yoheimuta, line 259: [Question] I'm not sure this compromise is acceptable in general.
I guess there is already a discussion about the balance between the accuracy and the performance.

For example: https://github.com/vitessio/vitess/blob/3f25ae8f9e26493c92f03b984231b7f31ed1529e/go/vt/vttablet/tabletmanager/vreplication/replicator_plan.go#L253-L269

yoheimuta: Sifting through like "pk <= lastpk" without any negative performance impact is ideal.

mattlord (changes requested):
I think that we need to find a way to implement this high level feature w/o changing the basic copy phase event filtering for all VStreams in uvstreamer (the unified vstreamer abstraction/primitive that everything else uses) as this changes key behavior for VReplication workflows and other things.

I also don't think that change should be necessary as we should be able to mirror the VReplication MoveTables/Reshard behavior here. @rohit-nayak-ps and I can discuss further.

go/vt/vtgate/endtoend/vstream_test.go
```diff
@@ -232,6 +233,129 @@ func TestVStreamCopyBasic(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	_, err = conn.ExecuteFetch("update t1 set id2 = 219 where id1 = 7", 1, false)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	expectedRowCopyEvents = 5 // id1 and id2 IN(5,6,7,8,9)
+	expectedCatchupEvents = 2
```
yoheimuta, line 346: Actually, the two update events will be caught in the replication phase, not the initial copy/catchup phase.
The expected catchup is [skipped](https://github.com/vitessio/vitess/blob/567fe98ce6e54db2324733e58ce5fc0bfc18fba3/go/vt/vttablet/tabletserver/vstreamer/copy.go#L51) because of missing `uvs.pos`.

The test is passed because the updates happened after the vstream started the copy.

yoheimuta: I want to pass the test changed like the below.
It means that the next connection can fetch all the changes in between.

```diff
diff --git a/go/vt/vtgate/endtoend/vstream_test.go b/go/vt/vtgate/endtoend/vstream_test.go
index a523d5e1eb..f7ce55c8b0 100644
--- a/go/vt/vtgate/endtoend/vstream_test.go
+++ b/go/vt/vtgate/endtoend/vstream_test.go
@@ -323,12 +323,6 @@ func TestVStreamCopyResume(t *testing.T) {
        }
        defer gconn.Close()

-       reader, err = gconn.VStream(ctx, topodatapb.TabletType_PRIMARY, vgtid, filter, flags)
-       if err != nil {
-               t.Fatal(err)
-       }
-       require.NotNil(t, reader)
-
        _, err = conn.ExecuteFetch("insert into t1(id1,id2) values (5,5), (6,6), (7,7), (8,8), (9,9)", 1, false)
        if err != nil {
                t.Fatal(err)
@@ -342,6 +336,12 @@ func TestVStreamCopyResume(t *testing.T) {
                t.Fatal(err)
        }

+       reader, err = gconn.VStream(ctx, topodatapb.TabletType_PRIMARY, vgtid, filter, flags)
+       if err != nil {
+               t.Fatal(err)
+       }
+       require.NotNil(t, reader)
+
        expectedRowCopyEvents = 5 // id1 and id2 IN(5,6,7,8,9)
        expectedCatchupEvents = 2
        expectedEvents = []string{
```

yoheimuta: From this point of view, I thought [the previous conditional](https://github.com/vitessio/vitess/pull/11103/commits/9971669e28b63e6c8fecdea106dd07c935cf1844#diff-32dbe2effb6b8b7e1151736ee190783bb236feb90ad7c04fe8ec6eec3ebea30fL360) was necessary.

go/vt/vttablet/tabletserver/vstreamer/uvstreamer.go
```diff
@@ -351,12 +366,18 @@ func (uvs *uvstreamer) init() error {
...
+	}
+
+	if uvs.startPos == "" || len(uvs.inTablePKs) > 0 {
 		if err := uvs.buildTablePlan(); err != nil {
 			return err
 		}
 	}
 
+	if uvs.startPos != "" && len(uvs.inTablePKs) > 0 {
+		uvs.pickingUpInputOffset = true
```
yoheimuta, line 378: `pickingUpInputOffset` acts like a new custom behavior type.
It depends on only the input parameters.

We aim that the existing logic is clearly intact if this type is false.

yoheimuta: In this sense, I can re-write https://github.com/vitessio/vitess/pull/11103/files#diff-32dbe2effb6b8b7e1151736ee190783bb236feb90ad7c04fe8ec6eec3ebea30fR371 like:

```diff
 func (uvs *uvstreamer) init() error {
-       if uvs.startPos != "" {
+       if uvs.startPos != "" && len(uvs.inTablePKs) > 0 {
+               uvs.pickingUpInputOffset = true
                if err := uvs.setStreamStartPosition(); err != nil {
                        return err
                }
-       } else if uvs.startPos == "" || len(uvs.inTablePKs) > 0 {
                if err := uvs.buildTablePlan(); err != nil {
                        return err
                }
+       } else {
+               if uvs.startPos != "" {
+                       if err := uvs.setStreamStartPosition(); err != nil {
+                               return err
+                       }
+               } else if uvs.startPos == "" || len(uvs.inTablePKs) > 0 {
+                       if err := uvs.buildTablePlan(); err != nil {
+                               return err
+                       }
+               }
        }
```

Although the existing condition in `init()` can remain intact for the safest bet, this code is more redundant.

yoheimuta:
@rohit-nayak-ps I reviewed my PR as you gave me the following suggestion.

> One possible option for enhancing the way it currently works is as follows:
add a new flag that will support the behaviour we use in vreplication. Default will be off (current behaviour)
then it will be possible to pass an intermediate VGTID with the last seen PK and GTID to restart a copy
we will then send the possibly duplicate events, mentioned above, and expect that the client knows how to handle it
>> https://vitess.slack.com/archives/C0PQY0PTK/p1665576567329369?thread_ts=1661155731.380559&cid=C0PQY0PTK

And I found that this PR was basically in line with your suggestion.

The slight difference is how to turn on a new flag.
I thought it was enough to change the current behaviour based on the existence of VGTID with both lastPK and GTID because this usage doesn't look specifically documented. At least [the current end-to-end test](https://github.com/vitessio/vitess/blob/567fe98ce6e54db2324733e58ce5fc0bfc18fba3/go/vt/vtgate/endtoend/vstream_test.go#L190) covers the case with lastPK and an empty GTID.

Given [the current counterintuitive behaviour](https://github.com/vitessio/vitess/pull/11103#issue-1351694162), I think it would be acceptable to regard the setting of lastpk and GTID together as a new flag for a resumable one. If you want to avoid duplicate events, you can simply set GTID to an empty string as before.
But, it also makes sense that adding a new flag to `vtgatepb.VStreamFlags` is the safest bet and clearer. If you think so, I'm willing to change that part.

Moreover, I realized through [the discussion](https://vitess.slack.com/archives/C0PQY0PTK/p1667039125635289?thread_ts=1661155731.380559&cid=C0PQY0PTK) that this PR is worth it when copying a single large table, at least.

Does this sound right, or am I missing something?

yoheimuta:
> I think it would be acceptable to regard the setting of lastpk and GTID together as a new flag for a resumable one.

The current PR changes the behaviour as the following table.

| GTID  | LastPK | Current behaviour                                                  | New behaviour                                                       |
|-------|--------|--------------------------------------------------------------------|---------------------------------------------------------------------|
| Empty | Empty  | Start a full copy                                                  | No change                                                           |
| Empty | Set    | Restart the copy from the lastpk without possible duplicate events | No change                                                           |
| Set   | Empty  | Start a replicate                                                  | No change                                                           |
| Set   | Set    | Start a replicate                                                  | **Restart the copy from the lastpk with possible duplicate events** |

rohit-nayak-ps:
@yoheimuta, sorry for the delay in responding to this PR. We were busy with the Vitess release and then I was at kubecon. 

We should definitely go ahead with this PR. They enhance the VStream API both from a functionality and a scalability point of view. Thanks for the work so far on this.

Your observations make sense. I realized that we are already requiring the VStream API client to require handling duplicates if a copy phase fails midway and has to be restarted. So we don't really need any new flags to be supported.

Give me a day or two: I am trying out a slightly different logic in `uvstreamer` to achieve this functionality. If it works I will add a commit to this PR. We can then discuss and plan to merge this PR by next week.

yoheimuta:
@rohit-nayak-ps Good to hear that! Congrats on the v15.0.0 and the KubeCon 🎉 
Thank you for updating me. Got it.

rohit-nayak-ps:
@yoheimuta, I made the changes I referred to above. Please review and make any eventual changes, when you get a chance. We can then request @mattlord for a PR review.

yoheimuta:
@rohit-nayak-ps Thank you for your updates!
I reviewed it, and looks good to me 👍

go/vt/vttablet/tabletserver/vstreamer/uvstreamer.go
```diff
@@ -346,17 +375,22 @@ func (uvs *uvstreamer) currentPosition() (mysql.Position, error) {
...
 
+// Possible states:
+// 1. TablePKs nil, startPos set to gtid or "current" => start replicating from pos
+// 2. TablePKs nil, startPos empty => full table copy of tables matching filter
+// 3. TablePKs not nil, startPos empty => table copy (for pks > lastPK)
+// 4. TablePKs not nil, startPos set => run catchup from startPos, then table copy  (for pks > lastPK)
 func (uvs *uvstreamer) init() error {
-	if uvs.startPos != "" {
-		if err := uvs.setStreamStartPosition(); err != nil {
+	if uvs.startPos == "" /* full copy */ || len(uvs.inTablePKs) > 0 /* copy specified tables */ {
```
mattlord, line 384: `full copy` and `resume copy` are better comments, no?

mattlord (commented):
Looks good and makes sense to me! I only had some very minor question/comments/suggestions.

I'll push a commit now to mention this new feature in the 16.0 release notes.

We should also add a new `VStream` page in the [16.0 `VReplication` docs](https://vitess.io/docs/16.0/reference/vreplication/) that can reference the existing [copy table design doc](https://vitess.io/docs/design-docs/vreplication/vstream/vscopy/). There we can touch on what it is, what use cases it serves, how it works and how to use it. We don't have any docs on the `VTGate` `VStream` API feature today (only [the somewhat dated design doc)](https://vitess.io/docs/design-docs/vreplication/vstream/vscopy/) — so many people don't even know that it exists. I can get a PR started for that and add a reference in the description.

mattlord (approved):
My review comments were extremely minor -- really only comment related -- so I'm going to approve now that I pushed an update to the release notes and started on the docs work (which is now linked): https://github.com/vitessio/website/pull/1216

Thank you so much for the time you spent on this @yoheimuta !

rohit-nayak-ps (approved):
lgtm

go/vt/vtgate/endtoend/vstream_test.go
```diff
@@ -232,6 +233,123 @@ func TestVStreamCopyBasic(t *testing.T) {
...
 }
 
+func TestVStreamCopyResume(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	gconn, conn, mconn, closeConnections := initialize(ctx, t)
+	defer closeConnections()
+
+	_, err := conn.ExecuteFetch("insert into t1_copy_resume(id1,id2) values(1,1), (2,2), (3,3), (4,4), (5,5), (6,6), (7,7), (8,8)", 1, false)
+	if err != nil {
```
systay, line 243: nit-picking: we usually use `stretchr/testify` and `require.NoError(t, err)` for these assertiong

go/vt/vtgate/endtoend/vstream_test.go
```diff
@@ -232,6 +233,123 @@ func TestVStreamCopyBasic(t *testing.T) {
...
+	catchupQueries := []string{
+		"insert into t1_copy_resume(id1,id2) values(9,9)", // this row will show up twice: once in catchup and copy
+		"update t1_copy_resume set id2 = 10 where id1 = 1",
+		"insert into t1(id1, id2) values(100,100)",
+		"delete from t1_copy_resume where id1 = 1",
+		"update t1_copy_resume set id2 = 90 where id1 = 9",
+	}
+	for _, query := range catchupQueries {
+		_, err = conn.ExecuteFetch(query, 1, false)
+		if err != nil {
```
systay, line 272: no need to use `if` here - `require.NoError(t, err)` does the nil check alrady

systay (approved):
some nit-picks, but they are not stopping this PR

GuptaManan100 (approved):

PR #11110, opened 2022-08-29
Add explicit close state to memory topo connection

dbussink:
The model of where we clear out the factory instance to nil is not really tenable since it always by definition races with `Watch` shutdown and there's no way to resolve that without explicit communication.

This introduces an explicit closed state flag instead and we error out if we're already closed to avoid issues around usage after close. This resolves all the issues.

The tests here are also updated to essentially always trigger the previous problem and it's gone with the changes applied here.

## Related Issue(s)

Previous attempts like https://github.com/vitessio/vitess/pull/11065, https://github.com/vitessio/vitess/pull/10954 & https://github.com/vitessio/vitess/pull/10906 tried to improve this but don't solve the more fundamental issue. Tried to avoid this change before but I think we have to make it.

## Checklist

### DIFF
--- a/go/vt/topo/memorytopo/memorytopo.go
+++ b/go/vt/topo/memorytopo/memorytopo.go
@@ -21,6 +21,7 @@ package memorytopo
 
 import (
 	"context"
+	"errors"
 	"math/rand"
 	"strings"
 	"sync"
@@ -36,6 +37,8 @@ const (
 	// Path components
 	electionsPath = "elections"
 )
+
+var ErrConnectionClosed = errors.New("connection closed")
 
 const (
 	// UnreachableServerAddr is a sentinel value for CellInfo.ServerAddr.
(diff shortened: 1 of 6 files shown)


go/vt/topo/memorytopo/memorytopo.go
```diff
@@ -140,12 +147,8 @@ func (c *Conn) dial(ctx context.Context) error {
 }
 
 // Close is part of the topo.Conn interface.
-// It nils out factory, so any subsequent call will panic.
 func (c *Conn) Close() {
-	c.factory.Lock()
-	f := c.factory
-	defer f.Unlock()
-	c.factory = nil
+	c.closed = true
```
mdlayher, line 151: Can topo.Conns be used concurrently? If so, you could use an atomic increment and then load to check if closed. Otherwise this could also race reads/writes.

dbussink: @mdlayher They're not intended to be used concurrently here. The main race issue is between these and the watchers which spawn a separate goroutine.

harshit-gangal (approved):

Issue #11112, opened 2022-08-29
Bug Report: Planner should treat single value tuples in `IN` conditions like `=`


arthurschreiber:
### Overview of the Issue

We noticed that Vitess treats vindex conditions that use the `IN` operator with a single value tuple differently than vindex conditions that use the `=` operator.

This affects cost calculations in the planner and leads it to believe that `=` conditions on lookup vindexes are cheaper than `IN` conditions on hash vindexes with single value tuples (which is not the case).

It also leads to suboptimal behaviour when checking for route mergeability.

### Reproduction Steps

Using the vschema defined in `go/vt/vtgate/planbuilder/plan_test.go`

```sql
SELECT music.id FROM music WHERE music.user_id IN (5) AND music.id = 3
```

Generates this plan:

```json
{
  "QueryType": "SELECT",
  "Original": "SELECT music.id FROM music WHERE music.user_id IN (5) AND music.id = 3",
  "Instructions": {
    "OperatorType": "Route",
    "Variant": "EqualUnique",
    "Keyspace": {
      "Name": "user",
      "Sharded": true
    },
    "FieldQuery": "select music.id from music where 1 != 1",
    "Query": "select music.id from music where music.user_id in (5) and music.id = 3",
    "Table": "music",
    "Values": [
      "INT64(3)"
    ],
    "Vindex": "music_user_map"
  },
  "TablesUsed": [
    "user.music"
  ]
}
```

Whereas

```sql
SELECT music.id FROM music WHERE music.user_id IN (5) AND music.id = 3
```

Generates this (better) plan, as it does not involve the additional lookup on the `music_user_map` vindex:

```json
{
  "QueryType": "SELECT",
  "Original": "SELECT music.id FROM music WHERE music.user_id = 5 AND music.id = 3",
  "Instructions": {
    "OperatorType": "Route",
    "Variant": "EqualUnique",
    "Keyspace": {
      "Name": "user",
      "Sharded": true
    },
    "FieldQuery": "select music.id from music where 1 != 1",
    "Query": "select music.id from music where music.user_id = 5 and music.id = 3",
    "Table": "music",
    "Values": [
      "INT64(5)"
    ],
    "Vindex": "user_index"
  },
  "TablesUsed": [
    "user.music"
  ]
}
```

---

```sql
SELECT music.id FROM music WHERE music.user_id IN (5) AND music.id IN (SELECT music.id FROM music WHERE music.user_id = 5)
```

generates this plan:

```json
{
[...]

PR #11123, opened 2022-08-30
Treat `IN` operations on single value tuples as `Equal` operations.

arthurschreiber:
## Description

This improves routing mergeability and vindex selection by treating `IN` operations with single value tuples like `Equal` operations.

A query like this:

```sql
SELECT music.id FROM music WHERE music.user_id IN (5)
```

Will now have a plan like this:

```json
{
  "QueryType": "SELECT",
  "Original": "SELECT music.id FROM music WHERE music.user_id IN (5)",
  "Instructions": {
    "OperatorType": "Route",
    "Variant": "EqualUnique",
    "Keyspace": {
      "Name": "user",
      "Sharded": true
    },
    "FieldQuery": "select music.id from music where 1 != 1",
    "Query": "select music.id from music where music.user_id in (5)",
    "Table": "music",
    "Values": [
      "INT64(5)"
    ],
    "Vindex": "user_index"
  },
  "TablesUsed": [
    "user.music"
  ]
}
```

---

This is a very simple fix for this issue, and it does not handle the following edge cases:

* Multiple repeated values: `music.user_id IN (5, 5, 5)` will not be treated as a single value tuple, although it essentially is equivalent to `music.user_id IN (5)`. Can we use the evaluation engine to perform a "distinct" operation on the tuple values?
* Values using different types. `music.user_id IN (5,  '5')` will not be treated as a single value tuple, although MySQLs type conversion rules will essentially perform comparisons with types that don't match after conversion to floating point values. Can we again use the evaluation engine to take the column type into account when performing the "distinct" operation?

## Related Issue(s)

Fixes: https://github.com/vitessio/vitess/issues/11112

## Checklist

### DIFF
--- a/go/vt/vtgate/planbuilder/physical/route.go
+++ b/go/vt/vtgate/planbuilder/physical/route.go
@@ -498,6 +498,12 @@ func (r *Route) planInOp(ctx *plancontext.PlanningContext, cmp *sqlparser.Compar
 	switch left := cmp.Left.(type) {
 	case *sqlparser.ColName:
 		vdValue := cmp.Right
+
+		valTuple, isTuple := vdValue.(sqlparser.ValTuple)
+		if isTuple && len(valTuple) == 1 {
+			return r.planEqualOp(ctx, &sqlparser.ComparisonExpr{Left: left, Right: valTuple[0], Operator: sqlparser.EqualOp})
+		}
+
 		value := r.makeEvalEngineExpr(ctx, vdValue)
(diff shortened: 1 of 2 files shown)


systay (approved):

systay:
For the `music.user_id IN (5, 5, 5)` and `music.user_id IN (5, '5')` examples - we should do this as early as possible. If we could push this logic into the AST rewriter, that would be great.

PR #11124, opened 2022-08-30
Upgrade vtorc CI

dbussink:
Extracted from https://github.com/vitessio/vitess/pull/11026

## Related Issue(s)

https://github.com/vitessio/vitess/pull/11026

## Checklist


dbussink:
Closing this as it's not correct, since we now would run things twice against MySQL 8.0. Instead, we should do https://github.com/vitessio/vitess/pull/11153

PR #11125, opened 2022-08-30
[cli] [tabletserver/streamlog] Migrate streamlog flags to pflags

### LINKED ISSUE #10746: Switch flag definitions to be on pflag instead of flag in `package go/streamlog`

Current flags:

```
$ git grep -E "\bflag\.[A-Z]" -- go/streamlog/*.go
go/streamlog/streamlog.go:      RedactDebugUIQueries = flag.Bool("redact-debug-ui-queries", false, "redact full queries and bind variables from debug UI")
go/streamlog/streamlog.go:      QueryLogFormat = flag.String("querylog-format", "text", "format for query logs (\"text\" or \"json\")")
go/streamlog/streamlog.go:      QueryLogFilterTag = flag.String("querylog-filter-tag", "", "string that must be present in the query for it to be logged; if using a value as the tag, you need to disable query normalization")
go/streamlog/streamlog.go:      QueryLogRowThreshold = flag.Uint64("querylog-row-threshold", 0, "Number of rows a query has to return or affect before being logged; not useful for streaming queries. 0 means all queries will be logged.")
```

rohit-nayak-ps:
## Description

Updates tabletenv related flag references specified in https://github.com/vitessio/vitess/issues/10746

## Related Issue(s)

Fixes #10746

## Checklist

### DIFF
--- a/go/streamlog/streamlog.go
+++ b/go/streamlog/streamlog.go
@@ -18,7 +18,6 @@ limitations under the License.
 package streamlog
 
 import (
-	"flag"
 	"fmt"
 	"io"
 	"net/http"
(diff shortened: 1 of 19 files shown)


go/streamlog/streamlog.go
```diff
@@ -29,24 +28,16 @@ import (
 	"sync"
 	"syscall"
 
+	"github.com/spf13/pflag"
+
+	"vitess.io/vitess/go/vt/servenv"
+
```
ajm188, line 34: nit :)

go/streamlog/streamlog.go
```diff
@@ -58,6 +49,34 @@ var (
 		[]string{"Log", "Subscriber"})
 )
 
+var (
+	RedactDebugUIQueries = false
+	QueryLogFormat       = "text"
+	QueryLogFilterTag    = ""
+	QueryLogRowThreshold = uint64(0)
```
ajm188, line 56: for everything but `QueryLogFormat`, these are the zero values, so you can just do:

```go
var (
    RedactDebugUIQueries bool
    QueryLogFormat = "text"
    QueryLogFilterTag string
    QueryLogRowThreshold uint64
)
```

ajm188: can we un-export these as well? if it's helpful, what i have on [another PR](https://github.com/vitessio/vitess/pull/11122/files#diff-2f4a6e06176fc55981e641da8319d0722a1ae12ee347153a866fed8508ac5c7fR69) is:

1. `MyVariable` => `myVariable`
2. `func MyVariable() T { return myVariable }`

ajm188 (approved):

PR #11127, opened 2022-08-30
CLI Pflag migration work: Fix regression caused by pflag-vreplication PR

rohit-nayak-ps:
## Description

Previous PR https://github.com/vitessio/vitess/pull/11095 had introduced  a bug which related to running the tests manually on the command line. It would ignore all passed test flags due to two bugs: one a typo in the parsing of test flags and another, a missing call to perform the parsing in `TestMain` in `tabletmanager/vreplication`

## Related Issue(s)

#11095 

## Checklist


ajm188 (approved):
Lgtm

PR #11134, opened 2022-08-31
do_release: fix updateVitessExamples function call

frouioui:
## Description

This PR fixes how we call `updateVitessExamples` in do_release.

## Checklist

### DIFF
--- a/tools/do_release.sh
+++ b/tools/do_release.sh
@@ -101,7 +101,7 @@ else
 fi
 
 # Preparing the release commit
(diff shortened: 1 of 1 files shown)


GuptaManan100 (approved):
Hah, good catch

frouioui:
> Hah, good catch

Yeah I kind of forgot in the original PR :/

GuptaManan100:
That makes 2 of us.

PR #11118, opened 2022-08-30
Upgrade vtgate CI jobs to 20.04

dbussink:
This is extracted from https://github.com/vitessio/vitess/pull/11026

## Related Issue(s)

https://github.com/vitessio/vitess/pull/11026

## Checklist

-   [x] "Backport me!" label has been added if this change should be backported
-   [x] Tests were added or are not required
-   [x] Documentation was added or is not required

PR #11153, opened 2022-09-01
Move towards MySQL 8.0 as the default template generation

dbussink:
This upgrades the remaining things to Ubuntu 20.04 and makes MySQL 8.0 the default we run tests against. We still have tests for MySQL 5.7 but those are now explicitly opted into.

This should finish up the Ubuntu 20.04 upgrade and also makes things easier for the future when we need to upgrade again.

It is also marked as back port since we want to keep this updated on back ports as well or otherwise we're be blocked in the somewhat near future when 18.04 is deprecated. We probably want to look at changing the default maybe back there to 5.7 (to minimize other test changes that happened in previous PRs like https://github.com/vitessio/vitess/pull/11118) but that's a separate step to look at during back porting. 

## Related Issue(s)

Part of the upgrade already started with https://github.com/vitessio/vitess/pull/11026 but keeping things sensible in the CI generation scripts is much easier with this upgrade we think.

## Checklist


test/ci_workflow_gen.go
```diff
@@ -121,15 +120,14 @@ var (
 		"vreplication_cellalias",
```
deepthi, line 120: Line 117: do we still need this?

test/ci_workflow_gen.go
```diff
@@ -175,18 +169,12 @@ func clusterMySQLVersions(clusterName string) mysqlVersions {
...
 	case clusterName == "tabletmanager_tablegc":
 		return allMySQLVersions
-	case clusterName == "mysql80":
-		return []mysqlVersion{mysql80}
-	case clusterName == "vtorc_8.0":
-		return []mysqlVersion{mysql80}
-	case clusterName == "vreplication_across_db_versions":
-		return []mysqlVersion{mysql80}
+	case clusterName == "vtorc":
+		return allMySQLVersions
```
deepthi, line 173: This means that **two** workflow files are generated, one each for 5.7 and 8.0. This is why `vtorc_8.0` has been deleted from `clusterList`.

.github/workflows/cluster_endtoend_vtorc_mysql57.yml
```diff
@@ -115,7 +128,7 @@ jobs:
         set -x
 
         # run the tests however you normally do, then produce a JUnit XML file
-        eatmydata -- go run test.go -docker=false -follow -shard vtorc_8.0 | tee -a output.txt | go-junit-report -set-exit-code > report.xml
+        eatmydata -- go run test.go -docker=false -follow -shard vtorc | tee -a output.txt | go-junit-report -set-exit-code > report.xml
```
deepthi, line 131: This needs to be `vtorc_5.7` because we don't want to run the topology tests on 5.7

go/test/endtoend/vtorc/readtopologyinstance/main_test.go
```diff
@@ -93,7 +93,7 @@ func TestReadTopologyInstanceBufferable(t *testing.T) {
 	assert.Contains(t, primaryInstance.InstanceAlias, "zone1")
 	assert.NotEqual(t, 0, primaryInstance.ServerID)
 	assert.Greater(t, len(primaryInstance.ServerUUID), 10)
-	assert.Contains(t, primaryInstance.Version, "5.7")
+	assert.Contains(t, primaryInstance.Version, "8.0")
```
GuptaManan100, line 96: We can use `assert.Regexp(t, `[58]\.[07].*`, replicaInstance.Version)` instead. This will make this test work with both 5.7 and 8.0

GuptaManan100 (commented):

.github/workflows/cluster_endtoend_vtorc_mysql57.yml
GuptaManan100: Yes, we would need to change this, or we can do what I suggest https://github.com/vitessio/vitess/pull/11153#discussion_r976281053 and run the topology tests for both versions and remove the vtorc_5.7 entirely from config.json

deepthi:
I've addressed the feedback, and except for the known issue with vtbackup_transform, tests are passing.
When we merge this, we also need to remove a bunch of required checks and add new ones.
I'm inclined to wait until all of the WIP flag PRs are merged and then do one more merge from main into this PR before we merge it.

.github/workflows/cluster_endtoend_vtorc_mysql57.yml
deepthi: Done

mattlord (approved):
❤️

PR #11085, opened 2022-08-23
Getting rid of external logging

rsajwani:
Signed-off-by: Rameez Sajwani <rameezwazirali@hotmail.com>

## Description

This PR will remove dependency from orchestrator/external/golib/log and use vitess/go/vt/log in VTORC.

Log entries before the change
=======================

I0824 10:42:08.275460   26474 log.go:226] starting orchestrator
2022-08-24 10:42:08    DEBUG db.go:148] Connected to orchestrator backend: sqlite on file::memory:?mode=memory&cache=shared
2022-08-24 10:42:08    DEBUG db.go:181] Initializing orchestrator
2022-08-24 10:42:08    DEBUG db.go:316] Migrating database schema
2022-08-24 10:42:08    DEBUG db.go:242] Migrated database schema to version []
2022-08-24 10:42:08     INFO db.go:196] Connecting to backend :3306: maxConnections: 128, maxIdleConns: 32
I0824 10:42:08.340530   26474 log.go:226] Starting Discovery
I0824 10:42:08.340548   26474 log.go:226] Registering endpoints
2022-08-24 10:42:08     INFO orchestrator.go:343] continuous discovery: setting up
2022-08-24 10:42:08    DEBUG queue.go:96] Queue.startMonitoring(DEFAULT)
2022-08-24 10:42:08     INFO http.go:165] Starting HTTP listener on :6931
2022-08-24 10:42:08     INFO orchestrator.go:375] continuous discovery: starting

Log entries after the change 
======================
I0824 10:55:34.771867   32574 main.go:183] starting orchestrator
I0824 10:55:34.773021   32574 config.go:512] Read config: /Users/rameezsajwani/Code/fork/vitess/vtdataroot/vtroot_7101/tmp_7103/orc-config-1661363733717036000.json
I0824 10:55:34.773237   32574 db.go:148] Connected to orchestrator backend: sqlite on file::memory:?mode=memory&cache=shared
I0824 10:55:34.773256   32574 db.go:311] Initializing orchestrator
I0824 10:55:34.774282   32574 db.go:321] Migrating database schema
I0824 10:55:34.830636   32574 db.go:244] Migrated database schema to version []
I0824 10:55:34.830699   32574 db.go:198] Connecting to backend :3306: maxConnections: 128, maxIdleConns: 32
I0824 10:55:34.835418   32574 http.go:130] Starting Discovery
I0824 10:55:34.835438   32574 http.go:134] Registering endpoints
[...]

PR #11156, opened 2022-09-01
Nil-check errors before printing them in VTOrc

GuptaManan100:
## Description

After moving away from external logging in https://github.com/vitessio/vitess/pull/11085, we started seeing error messages like 
```
E0901 16:38:28.412132   63266 topology_recovery_dao.go:112] <nil>
E0901 16:38:28.412186   63266 topology_recovery_dao.go:356] <nil>
E0901 16:38:28.412247   63266 topology_recovery_dao.go:260] <nil>
E0901 16:38:28.412352   63266 cluster_alias_dao.go:142] <nil>
```

The problem was traced to not doing a nil-check before printing the error message. The previous implementation of `log.Errore` used to do the nil-check before printing the error, so it wasn't required in the codebase.
## Checklist

### DIFF
--- a/go/vt/orchestrator/app/cli.go
+++ b/go/vt/orchestrator/app/cli.go
@@ -187,7 +187,9 @@ func Cli(command string, strict bool, instance string, destination string, owner
 				log.Fatal(err)
 			} else {
 				for _, e := range errs {
-					log.Error(e)
+					if e != nil {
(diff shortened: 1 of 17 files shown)


dbussink (approved):

rsajwani (commented):
LGTM

PR #11168, opened 2022-09-01
adding tablet check before calling InitPrimary

rsajwani:

## Description

Vtbackup_transform is flaky due to tablet sometime not healthy before we call InitPrimary

Initial approach of tablet.VttabletProcess.GetStatus(); did not work as registerDebugEnvHandler() is called before tm.Start().

## Checklist

### DIFF
--- a/go/test/endtoend/backup/vtbackup/backup_only_test.go
+++ b/go/test/endtoend/backup/vtbackup/backup_only_test.go
@@ -234,20 +234,28 @@ func initTablets(t *testing.T, startTablet bool, initShardPrimary bool) {
 		}
 	}
 
-	if startTablet {
-		var healthyTab = 0
-		for _, tablet := range []cluster.Vttablet{*primary, *replica1} {
-			// wait for both tablet to get into healthy state
-			waitUntil := time.Now().Add(30 * time.Second)
-			for time.Now().Before(waitUntil) {
-				if result := tablet.VttabletProcess.GetStatus(); result != "" {
-					healthyTab++
-					break
+	// wait for both tablet to get into replica state in topo
+	waitUntil := time.Now().Add(10 * time.Second)
+	for time.Now().Before(waitUntil) {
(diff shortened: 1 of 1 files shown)


deepthi:
I'm going to defer this review to @mattlord since he has most recently been working on flaky tests and can advise as to the best strategy.

go/test/endtoend/backup/vtbackup/backup_only_test.go
```diff
@@ -234,6 +234,30 @@ func initTablets(t *testing.T, startTablet bool, initShardPrimary bool) {
 		}
 	}
 
+	// wait for both tablet to get into replica state in topo
+	waitUntil := time.Now().Add(10 * time.Second)
```
mattlord, line 238: I think 30 seconds is a better timeout for something like this. GitHub actions can at times be super slow, and failing the workflow because it took 15 seconds doesn't really help us (we will have to re-run the entire workflow). 

BUT, `tablet.VttabletProcess.Setup()` waits for the process to start and for the tablet to reach the NOT_SERVING status here (as we didn't specify). So I'm not sure what this additional check is doing here?

Do we think that the tablet is not registered in the topo yet for some reason? Do you have examples of test failures that we think this will resolve? The failures that I had seem seemed to be due to [the transform hook](https://github.com/vitessio/vitess/blob/main/test/vthook-test_backup_transform) never returning. Is that somehow related to that? Did you perhaps re-run that workflow on this PR 10+ times to confirm it was no longer flaky? If it helps, then that's all we really care about. :)

mattlord: I re-tried it and it failed on the 3rd try: https://github.com/vitessio/vitess/runs/8254729459?check_suite_focus=true

It seems like once it fails once, it seems to fail several times in a row. At least that was my experience. So I wonder if it's some state on the Actions host or some other context issue.

PR #11170, opened 2022-09-02
Fix problematic watch cancellation due to context cancellation

dbussink:
Right now we pass in the context when starting a Watch that is also used for the request context. This means that the Watch ends up being cancelled when the original request that started it as a side effects ends up completing and cancels the context to clean up.

This is of course not as intended. Before the refactor in https://github.com/vitessio/vitess/pull/10906 this wasn't causing a
practical issue yet. We'd still have the expired context internally in the watcher and it would be passed through with updating entries, but there were no calls that ended up validating the context expiry, avoiding any immediate issue.

This is bound to fail though at some point if something would be added that does care about the context. What is needed is that the watcher we start sets up it's own context based on the background context since it is detached from the original request that might trigger starting the watcher as a side effect.

Additionally, it means that the tracked context for an error isn't really useful. It would often be an already cancelled context from a mostly unrelated request which doesn't provide useful information. Even more so, it would keep a reference to that context so it would never be garbage collection potentially and would keep more request data alive than necessary.

With the fix, the context is always from the background context with a cancel on top for that watcher. This isn't very useful either. Also we don't use this context tracking for any error messaging or reporting anywhere, so I believe it's better to clean up this tracking.

By cleaning up that tracking, we also avoid the need to pass down the context in entry updates and that is all cleaned up here as well.

Lastly, a failing test is introduced that verifies the original issue. It retrieves serving keyspace information, cancels the original request that triggered that and then validates the watcher is still running by updating the value again within the timeout window. This failed before this fix as the watcher would be cancelled and the cached old value was returned before the TTL expired.

The main problem of this bug is not an issue of correctness, but of a serious performance degration in the vtgate. Each second we'd restart context setup if we ever had a failure on the path triggered by regular queries and the system would not recover from this situation and heavily query the topo server and make things very expensive.

## Related Issue(s)

Follow up for https://github.com/vitessio/vitess/pull/10906

## Checklist

### DIFF
--- a/go/vt/srvtopo/resilient_server_test.go
+++ b/go/vt/srvtopo/resilient_server_test.go
@@ -45,8 +45,8 @@ import (
 // TestGetSrvKeyspace will test we properly return updated SrvKeyspace.
 func TestGetSrvKeyspace(t *testing.T) {
 	ts, factory := memorytopo.NewServerAndFactory("test_cell")
-	*srvTopoCacheTTL = time.Duration(100 * time.Millisecond)
-	*srvTopoCacheRefresh = time.Duration(40 * time.Millisecond)
+	*srvTopoCacheTTL = time.Duration(200 * time.Millisecond)
+	*srvTopoCacheRefresh = time.Duration(80 * time.Millisecond)
 	defer func() {
 		*srvTopoCacheTTL = 1 * time.Second
 		*srvTopoCacheRefresh = 1 * time.Second
@@ -70,9 +70,11 @@ func TestGetSrvKeyspace(t *testing.T) {
 
 	// wait until we get the right value
 	var got *topodatapb.SrvKeyspace
-	expiry := time.Now().Add(5 * time.Second)
+	expiry := time.Now().Add(*srvTopoCacheRefresh - 20*time.Millisecond)
 	for {
-		got, err = rs.GetSrvKeyspace(context.Background(), "test_cell", "test_ks")
+		ctx, cancel := context.WithCancel(context.Background())
+		got, err = rs.GetSrvKeyspace(ctx, "test_cell", "test_ks")
+		cancel()
 
 		if err != nil {
 			t.Fatalf("GetSrvKeyspace got unexpected error: %v", err)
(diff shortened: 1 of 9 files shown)


go/vt/srvtopo/resilient_server_test.go
```diff
@@ -45,8 +45,8 @@ import (
 // TestGetSrvKeyspace will test we properly return updated SrvKeyspace.
 func TestGetSrvKeyspace(t *testing.T) {
 	ts, factory := memorytopo.NewServerAndFactory("test_cell")
-	*srvTopoCacheTTL = time.Duration(100 * time.Millisecond)
-	*srvTopoCacheRefresh = time.Duration(40 * time.Millisecond)
+	*srvTopoCacheTTL = time.Duration(200 * time.Millisecond)
+	*srvTopoCacheRefresh = time.Duration(80 * time.Millisecond)
```
dbussink, line 49: Changed the timing to have some extra room here so it's possible to verify more reliably below that things work as expected also within this timeout window.

deepthi (commented):
LGTM

deepthi:
> LGTM

Let's wait for @mattlord to review before we merge this.

go/vt/topo/etcd2topo/watch.go
```diff
@@ -54,23 +54,27 @@ func (s *Server) Watch(ctx context.Context, filePath string) (*topo.WatchData, <
...
 		watchCancel()
+		outerCancel()
 		return nil, nil, vterrors.Errorf(vtrpc.Code_INVALID_ARGUMENT, "Watch failed")
 	}
 
 	// Create the notifications channel, send updates to it.
 	notifications := make(chan *topo.WatchData, 10)
 	go func() {
 		defer close(notifications)
-		defer watchCancel()
```
mattlord, line 73: Curious why we removed this line but call watchCancel() before outerCancel() in other places like like 192?

mattlord (commented):
The changes make sense to me, I only had one question about the inner watchCancel() usage.

I also had one minor request about the test as flakiness is a persistent problem in the CI. 

Thank you!

mattlord (approved):
We discussed using a `waitForSrvKeyspaceResult()` function but this case is not as straightforward as the typical waitForX case. The new test modifications also line up pretty well with the rest of the test so we can put off any potential test de-flaking until the test actually turns out to be flaky. 🙂  

For any future reference, I think that something like this may be what we'd want to use if we do need to start adding waits with a timeout:
```
func waitForSrvKeyspaceResult(t *testing.T, rs *ResilientServer, cell string, keyspace string, want *topodatapb.SrvKeyspace) {
	// Allow for several attempts to deal with any ephemeral issues
	// and prevent flaky tests.
	timeout := *srvTopoCacheRefresh * 5
	tick := *srvTopoCacheRefresh
	timer := time.NewTimer(timeout)
	defer timer.Stop()
	var got *topodatapb.SrvKeyspace
	ctx := context.Background()

	for {
		select {
		case <-timer.C:
			t.Fatalf("Did not get expected GetSrvKeyspace() result of %+v before the timeout of %s. Last seen value: %+v",
				want, timeout, got)
		default:
			status := rs.CacheStatus()
			for _, sk := range status.SrvKeyspaces {
				if sk.Cell == cell && sk.Keyspace == keyspace && sk.LastError != nil {
					t.Fatalf("SrvKeyspace cache entry for cell %s and keyspace %s recorded the following error, which should not happen: %v",
						cell, keyspace, sk.LastError)
				}
			}
			got, err := rs.GetSrvKeyspace(ctx, cell, keyspace)
			if err != nil {
				t.Fatalf("GetSrvKeyspace() call had unexpected error: %v", err)
			}
			if proto.Equal(want, got) {
				return
			}
		}
		time.Sleep(tick)
	}
}
```

rsajwani (commented):
LGTM

PR #11171, opened 2022-09-04
evalengine: Support built-in MySQL function LOWER() 

Weijun-H:

## Description
Implement support LOWER()

## Related Issue(s)
- #9647 

## Checklist

### DIFF
--- a/go/vt/vtgate/evalengine/string.go
+++ b/go/vt/vtgate/evalengine/string.go
@@ -1,0 +1,54 @@
+/*
+Copyright 2022 The Vitess Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


go/vt/vtgate/evalengine/string.go
```diff
@@ -0,0 +1,81 @@
...
+		result.setNull()
+		return
+	}
+
+	if sqltypes.IsText(t) {
+		tolower := strings.ToLower(inarg.value().RawStr())
+		result.setString(tolower, inarg.collation())
+	} else {
+		tolower := inarg.value().RawStr()
+		inarg.makeTextualAndConvert(collations.CollationUtf8mb4ID)
```
frouioui, line 44: I think we should use the default collation stored in `ExpressionEnv` here and in `builtinLcase.call` too.

PR #11173, opened 2022-09-05
Prevent martini from logging in VTOrc

GuptaManan100:
## Description

It has been observed that `martini` code does excessive logging on each HTTP request served in VTOrc. This causes logs to fill up quickly and also makes it harder to find logs that are actually worth looking at.

This PR removes the logging from `martini` by setting its logger to a discard endpoint.

## Checklist

### DIFF
--- a/go/vt/orchestrator/app/http.go
+++ b/go/vt/orchestrator/app/http.go
@@ -18,6 +18,8 @@ package app
 
 import (
 	"flag"
+	"io/ioutil"
+	golog "log"
 	"net"
 	nethttp "net/http"
 	"path"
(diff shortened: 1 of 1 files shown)


dbussink (approved):
Is there an easy way to test that it doesn't log anything from the framework to the regular logger?

GuptaManan100:
I have tested it manually. Should I look into adding a test for checking no logs are printed from martini?

dbussink:
> I have tested it manually. Should I look into adding a test for checking no logs are printed from martini?

@GuptaManan100 I was wondering if there was a not too problematic way to do that, if so, might be nice? But if this is slated to be cleaned up / removed on the somewhat near term it might not be needed?

GuptaManan100:
@dbussink Found a way to test the changes 👍

dbussink (approved):

rsajwani (commented):
LGTM

PR #11175, opened 2022-09-06
adding setting pool metrics

harshit-gangal:
## Description

This PR adds metrics around the usage of settings pool.

## Related Issue(s)

- #9706 

## Checklist


GuptaManan100 (approved):
Did you check that these metrics are also visible on the vttablet UI?

harshit-gangal:
> Did you check that these metrics are also visible on the vttablet UI?

It is visible.

image

PR #11061, opened 2022-08-22
Add tracking session state changes for transaction start

dbussink:
This logic depends on the custom option that PlanetScale as in it's MySQL fork to get the current committed GTID at the start of a transaction snapshot.

With this GTID, we can optimize vreplication to avoid having to lock tables around vreplication operations.

See https://github.com/planetscale/mysql-server/commit/9fa67ec3ebdeda923a67963260a9e744aa635a3a for the change in MySQL that is published in our fork.

The change here allows getting back this value through any query operation that might start a transaction.

With this change, we can essentially do what was attempted in https://github.com/vitessio/vitess/pull/9991 but it will guarantee now a consistent GTID for the snapshot. 

The first commit here wires up passing through the session state changes on transaction start, the second commit starts using it for VReplication.

## Related Issue(s)

See https://github.com/vitessio/vitess/pull/9991 for a previous failed attempt.

## Checklist

-   [x] "Backport me!" label has been added if this change should be backported
-   [x] Tests were added or are not required
-   [x] Documentation was added or is not required

PR #11176, opened 2022-09-06
Fix OK packet parsing logic

dbussink:
Today the OK packet parsing logic has a significant bug. It assumes that GTID lengths always fit within a single byte length and it assumes that it can read a string then directly with a single byte.

Instead, we need to properly read an encoded length for the packet and then process further with that.

Before this fix, any time a GTID set was returned larger than what can be encoded with a 1 byte length bit, the parser would fail and return broken data.

## Related Issue(s)

This was discovered due to the logic added in https://github.com/vitessio/vitess/pull/11061

## Checklist


shlomi-noach (approved):

go/mysql/conn.go
```diff
@@ -1474,30 +1474,40 @@ func (c *Conn) parseOKPacket(in []byte) (*PacketOK, error) {
...
-			// read (and ignore for now) the GTIDS encoding specification code: 1 byte
-			_, ok = data.readByte()
-			if !ok {
-				return fail("invalid OK packet gtids type: %v", data)
-			}
-			gtids, ok := data.readLenEncString()
-			if !ok {
-				return fail("invalid OK packet gtids: %v", data)
+			// Alright, now we need to read each sub packet from the session state change.
+			for data.pos < len(data.data) {
```
harshit-gangal, line 1479: I believe we can check for length > 0 here.

dbussink: It's what I tried first, but that fails since we're missing the nested inner bytes that encode the length of the inner strings. So that's why I moved away from that, but it looks like I didn't clean up the `length` stuff so I'll do that to simplify this.

dbussink: I have cleaned this up a bit and simplified how the loop terminates.

harshit-gangal (approved):

PR #11177, opened 2022-09-07
Adds delete planning to Gen4

GuptaManan100:
## Description

This PR adds delete planning to Gen4 planner. This was required specifically for the following types of queries leads to incorrect tables used populated by v3 - 
```json
# delete with join from multi table join subquery.
"delete foo from unsharded as foo join (select id from unsharded a join unsharded_b b on a.user_id = b.user_id) as keepers on foo.id = keepers.id where keepers.id is null and foo.col is not null and foo.col \u003c 1000"
{
  "QueryType": "DELETE",
  "Original": "delete foo from unsharded as foo join (select id from unsharded a join unsharded_b b on a.user_id = b.user_id) as keepers on foo.id = keepers.id where keepers.id is null and foo.col is not null and foo.col \u003c 1000",
  "Instructions": {
    "OperatorType": "Delete",
    "Variant": "Unsharded",
    "Keyspace": {
      "Name": "main",
      "Sharded": false
    },
    "TargetTabletType": "PRIMARY",
    "MultiShardAutocommit": false,
    "Query": "delete foo from unsharded as foo join (select id from unsharded as a join unsharded_b as b on a.user_id = b.user_id) as keepers on foo.id = keepers.id where keepers.id is null and foo.col is not null and foo.col \u003c 1000",
    "Table": "unsharded, unsharded, unsharded_b"
  },
  "TablesUsed": [
    "main.unsharded",
    "main.unsharded, unsharded_b"
  ]
}
{
  "QueryType": "DELETE",
  "Original": "delete foo from unsharded as foo join (select id from unsharded a join unsharded_b b on a.user_id = b.user_id) as keepers on foo.id = keepers.id where keepers.id is null and foo.col is not null and foo.col \u003c 1000",
  "Instructions": {
    "OperatorType": "Delete",
    "Variant": "Unsharded",
    "Keyspace": {
      "Name": "main",
      "Sharded": false
    },
    "TargetTabletType": "PRIMARY",
    "MultiShardAutocommit": false,
    "Query": "delete foo from unsharded as foo join (select id from unsharded as a join unsharded_b as b on a.user_id = b.user_id) as keepers on foo.id = keepers.id where keepers.id is null and foo.col is not null and foo.col \u003c 1000",
    "Table": "unsharded, unsharded_b"
  },
  "TablesUsed": [
    "main.unsharded",
    "main.unsharded_b"
  ]
}

```

## Checklist

### DIFF
--- a/go/vt/vtgate/planbuilder/abstract/operator.go
+++ b/go/vt/vtgate/planbuilder/abstract/operator.go
@@ -236,39 +236,9 @@ func createOperatorFromSelect(sel *sqlparser.Select, semTable *semantics.SemTabl
 }
 
 func createOperatorFromUpdate(updStmt *sqlparser.Update, semTable *semantics.SemTable) (LogicalOperator, error) {
-	subq, err := createSubqueryFromStatement(updStmt, semTable)
+	tableInfo, qt, err := createQueryTableForDML(updStmt.TableExprs[0], semTable, updStmt.Where)
 	if err != nil {
 		return nil, err
-	}
-	alTbl, ok := updStmt.TableExprs[0].(*sqlparser.AliasedTableExpr)
-	if !ok {
-		return nil, vterrors.Errorf(vtrpcpb.Code_INTERNAL, "expected AliasedTableExpr")
-	}
-	tblName, ok := alTbl.Expr.(sqlparser.TableName)
-	if !ok {
-		return nil, vterrors.Errorf(vtrpcpb.Code_INTERNAL, "expected TableName")
-	}
-
-	tableID := semTable.TableSetFor(alTbl)
-	tableInfo, err := semTable.TableInfoFor(tableID)
-	if err != nil {
-		return nil, err
-	}
-
-	if tableInfo.IsInfSchema() {
-		return nil, vterrors.Errorf(vtrpcpb.Code_UNIMPLEMENTED, "can't update information schema tables")
-	}
-
-	var predicates []sqlparser.Expr
-	if updStmt.Where != nil {
-		predicates = sqlparser.SplitAndExpression(nil, updStmt.Where.Expr)
-	}
-	qt := &QueryTable{
-		ID:          tableID,
-		Alias:       alTbl,
-		Table:       tblName,
-		Predicates:  predicates,
-		IsInfSchema: false,
 	}
 
 	assignments := make(map[string]sqlparser.Expr)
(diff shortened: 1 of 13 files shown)


harshit-gangal (commented):
do you think common update and delete planner code can be extracted, and create dml operator and later transform to update/delete engine.

go/vt/vtgate/planbuilder/abstract/operator.go
```diff
@@ -287,6 +289,56 @@ func createOperatorFromUpdate(updStmt *sqlparser.Update, semTable *semantics.Sem
...
 
+func createOperatorFromDelete(deleteStmt *sqlparser.Delete, semTable *semantics.SemTable) (LogicalOperator, error) {
+	subq, err := createSubqueryFromStatement(deleteStmt, semTable)
+	if err != nil {
+		return nil, err
+	}
+
+	alTbl, ok := deleteStmt.TableExprs[0].(*sqlparser.AliasedTableExpr)
+	if !ok {
+		return nil, vterrors.Errorf(vtrpcpb.Code_INTERNAL, "expected AliasedTableExpr")
```
systay, line 300: are these really internal errors? should it not be 'Code_UNIMPLEMENTED'?

go/vt/vtgate/planbuilder/abstract/operator.go
```diff
@@ -287,6 +289,56 @@ func createOperatorFromUpdate(updStmt *sqlparser.Update, semTable *semantics.Sem
 	return subq, nil
 }
 
+func createOperatorFromDelete(deleteStmt *sqlparser.Delete, semTable *semantics.SemTable) (LogicalOperator, error) {
+	subq, err := createSubqueryFromStatement(deleteStmt, semTable)
+	if err != nil {
+		return nil, err
+	}
```
frouioui, line 296: nit: we could move this to line 335 where we do a nil check (`subq == nil`), so we don't have to prepare the subquery if the delete statement has to fail

go/vt/vtgate/planbuilder/physical/route_planning.go
```diff
@@ -161,12 +159,91 @@ func CreatePhysicalOperator(ctx *plancontext.PlanningContext, opTree abstract.Lo
 			return nil, vterrors.Errorf(vtrpcpb.Code_UNIMPLEMENTED, "multi shard update with limit is not supported")
 		}
 
+		return r, nil
+	case *abstract.Delete:
+		var typ topodatapb.TabletType
+		var dest key.Destination
+		var ovq string
+
```
frouioui, line 167: nit: it would be nice to extract this case into a method. Preferably, the rest of the cases should be extracted too

go/vt/vtgate/planbuilder/abstract/operator.go
GuptaManan100: Fixed

go/vt/vtgate/planbuilder/physical/route_planning.go
GuptaManan100: Fixed

harshit-gangal (approved):

PR #11181, opened 2022-09-07
generate settings plan in tablet with query and reset setting query

harshit-gangal:
## Description

This PR improves the implementation when a resource is taken from setting pool but does not have the required setting or does not needs to have any setting applied. So, instead of closing and opening a new connection we can use the reset setting query and make the same connection available to use. This will avoid the close/open overhead with an additional query to mysql.

## Related Issue(s)

- #9706 

## Checklist


go/pools/rp_bench_test.go
```diff
@@ -137,35 +137,3 @@ func BenchmarkGetPutMixedMulti(b *testing.B) {
 		}
 	}
 }
-
-func BenchmarkApplySettings(b *testing.B) {
-	b.ReportAllocs()
-	ctx := context.Background()
```
GuptaManan100, line 143: Any reason why the Benchmark is removed? Do we not need it any more?

go/vt/vttablet/tabletserver/connpool/dbconn.go
```diff
@@ -324,34 +325,35 @@ func (dbc *DBConn) Close() {
...
+// IsSettingApplied implements the pools.Resource interface.
+func (dbc *DBConn) IsSettingApplied() bool {
+	return dbc.setting != ""
 }
 
-func (dbc *DBConn) IsSameSetting(settings []string) bool {
-	return compareStringSlice(settings, dbc.settings)
+// IsSameSetting implements the pools.Resource interface.
+func (dbc *DBConn) IsSameSetting(setting string) bool {
+	return strings.EqualFold(setting, dbc.setting)
```
GuptaManan100, line 346: What would happen if the settings are the same but reordered? Like `set x=a, y=b` and `set y=b, x=a`? Do we consider them to be different settings?

go/vt/vttablet/tabletserver/query_executor.go
```diff
@@ -334,9 +334,9 @@ func (qre *QueryExecutor) Stream(callback StreamCallback) error {
 			return err
 		}
 		defer txConn.Unlock()
-		if len(qre.settings) > 0 {
-			if err = txConn.ApplySettings(qre.ctx, qre.settings); err != nil {
-				return vterrors.Wrap(err, "failed to execute system settings on the connection")
+		if qre.setting != nil {
```
GuptaManan100, line 337: Do we need a `IsNil()` test too here?

GuptaManan100 (commented):

go/pools/rp_bench_test.go
harshit-gangal: earlier this benchmark shows how executing a single query over multiple query have better effect, but with the current changes there will always be only single query that will be received. making this benchmark unnecessary.

go/vt/vttablet/tabletserver/connpool/dbconn.go
harshit-gangal: that is on vtgate to send in same order, it will not be handled here. 
VTGate already does that today in sending in alphanumeric sorted.

go/vt/vttablet/tabletserver/query_executor.go
harshit-gangal: it is only needed when you convert a struct reference to interface. So, it is not applicable here.

systay (approved):
LGTM

PR #11185, opened 2022-09-07
evalengine: Support built-in MySQL function for string functions and operations

Weijun-H:

## Description
The following functions will be supported
- [x]  LOWER(), LCASE()
- [x] UPPER(), UCASE()
- [x] CHAR_LENGTH(), CHARACTER_LENGTH()
- [x] LENGTH(), OCTET_LENGTH()
- [x] BIT_LENGTH() 
- [x] REPEAT()

## Related Issue(s)
- #9647 

## Checklist


Weijun-H:
I am working on `LOCATE`, but the following error
```bash
/vitess/go/vt/vtgate/evalengine/integration/string_fun_test.go:96: Code: UNIMPLEMENTED
    expr cannot be translated, not supported: locate('xxx', 'ef')
    ; mysql response: INT64(0)
    query: SELECT LOCATE ("xxx", "ef") (SIMPLIFY=true)
```
Does it mean Vitess does not support `LOCATE`? I do not know this error.

Weijun-H:
> I am working on `LOCATE`, but the following error
> 
> ```shell
> /vitess/go/vt/vtgate/evalengine/integration/string_fun_test.go:96: Code: UNIMPLEMENTED
>     expr cannot be translated, not supported: locate('xxx', 'ef')
>     ; mysql response: INT64(0)
>     query: SELECT LOCATE ("xxx", "ef") (SIMPLIFY=true)
> ```
> 
> Does it mean Vitess does not support `LOCATE`? I do not know this error.

I will fix this issue in another pr.

go/mysql/collations/8bit.go
```diff
@@ -113,6 +115,17 @@ func (c *Collation_8bit_bin) Wildcard(pat []byte, matchOne rune, matchMany rune,
...
+	for _, c := range c.simpletables.tolower {
+		fmt.Println(c)
+	}
+	return c.simpletables.tolower
+}
+
+func (c *Collation_8bit_bin) GetUpperTable() *[256]byte {
+	return c.simpletables.toupper
+}
+
```
vmg, line 128: You need to implement `ToUpper` and `ToLower` methods in the `Collation` interface. Exposing the lower/upper tables directly won't work because other collations (such as UTF8) do not use tables for upcasing.

go/mysql/collations/collation.go
```diff
@@ -144,10 +144,10 @@ type Collation interface {
 	IsBinary() bool
 
 	// GetUpperTable returns the upper table about related encoding
-	ToUpper([]byte) []byte
+	ToUpper([]byte, []byte) []byte
```
vmg, line 147: Like in the other APIs, the `dst` buffer should be the first argument for the function, not the last. You should also name the arguments and update the comments!

go/mysql/collations/collation.go
```diff
@@ -176,14 +176,14 @@ func register(c Collation) {
 	globalAllCollations[c.ID()] = c
 }
 
-func ToUpper(id ID, raw []byte) []byte {
+func ToUpper(id ID, raw []byte, dst []byte) []byte {
```
vmg, line 179: This API is not necessary. You're not supposed to use `globalAllCollations` because some MySQL versions don't support _some_ of the collations. That's why we only expose these methods from the `Collation` interface.

vmg (changes requested):
Good progress! Careful with all the cases that are not supported yet. It's OK to `panic` while they're unsupported but don't return wrong results.

go/vt/vtgate/evalengine/string.go
```diff
@@ -35,14 +35,17 @@ func (builtinLower) call(env *ExpressionEnv, args []EvalResult, result *EvalResu
 		return
 	}
 
+	coll := collations.Local().LookupByID(inarg.Collation())
```
vmg, line 38: This is the correct way to lookup a collation! Good job!

vmg (changes requested):
We're now looking up the collations properly. Good! But we still need to remove all the unimplemented methods which are now panicking.

vmg:
@Weijun-H can you rebase this branch on top of `main`? We've merged a bugfix that was causing all CI builds to fail.

vmg (approved):
@Weijun-H I've pushed a cleanup commit that mostly fixes the logic in `strings.go` to use the right EvalResult APIs. Looks good otherwise. Thanks!

PR #11193, opened 2022-09-09
Cluster-Alias cleanup for VTOrc

GuptaManan100:
## Description

This PR addresses one of the long-standing cleanups required for VTOrc. 
Orchestrator used to identify a cluster by the `host:port` name of the primary, which was used as the cluster_name. In order to have a more human-readable variant, it also had `cluster_alias` which could be set by regex matching from configurations, etc. Manual overrides could also be performed for a `cluster_alias`.

In the first iteration of VTOrc, an additional parameter called `suggested_cluster_alias` was added which was set to `keyspace/shard`.

Since VTOrc can just use this `keyspace/shard` to identify a cluster, it is much simpler to just use this for cluster_name and get rid of the other two fields. This leads to other cleanups too, where we don't have to change the cluster_name in case of a reparent, etc.

This PR makes this change, by cleaning up these two fields and setting cluster_name to `keyspace/shard`

## Related Issue(s)

- #6612 

## Checklist


go/test/endtoend/vtorc/readtopologyinstance/main_test.go
```diff
@@ -117,6 +117,7 @@ func TestReadTopologyInstanceBufferable(t *testing.T) {
 	assert.False(t, primaryInstance.HasReplicationCredentials)
 	assert.Equal(t, primaryInstance.ReplicationIOThreadState, inst.ReplicationThreadStateNoThread)
 	assert.Equal(t, primaryInstance.ReplicationSQLThreadState, inst.ReplicationThreadStateNoThread)
+	assert.Equal(t, fmt.Sprintf("%v:%v", keyspace.Name, shard0.Name), primaryInstance.ClusterName)
```
GuptaManan100, line 120: This augments the existing `ReadTopologyInstanceBufferable` end to end test to also verify that we set the cluster_name correctly.

go/vt/orchestrator/app/cli.go
```diff
@@ -504,15 +504,7 @@ func Cli(command string, strict bool, instance string, destination string, owner
...
-			instanceKey, _ = inst.FigureInstanceKey(instanceKey, thisInstanceKey)
-			if instanceKey == nil {
-				log.Fatal("Cannot deduce instance:", instance)
-			}
-			_, err := inst.ReattachReplicaPrimaryHost(instanceKey)
-			if err != nil {
-				log.Fatal(err)
-			}
-			fmt.Println(instanceKey.DisplayString())
+			fmt.Println("Command deprecated")
```
GuptaManan100, line 507: I have removed some API and CLI commands that would have required some changes and that we don't intend to support anyways. I could have removed them as a separate PR, but I elected to do it here as I went along.

go/vt/orchestrator/inst/instance_dao.go
```diff
@@ -1120,8 +1095,9 @@ func ReadClusterPrimary(clusterName string) ([](*Instance), error) {
 	condition := `
 		cluster_name = ?
 		and (replication_depth = 0 or is_co_primary)
+		and tablet_type = ?
```
GuptaManan100, line 1098: This additional step is required because earlier when we started a new tablet, it used to come up and its cluster_name would be set to the `hostname:port` of itself. But now, it will be set the keyspace/shard and then that new tablet used to show up in this `ReadClusterPrimary` function.

A better way is to also employ filtering on the tablet_type to be Primary.

Currently this function isn't unit tested, but it is end-to-end tested in `GracefulPrimaryTakeover` tests. I intend to add unit tests for all these functions that are reading data from the backend, but it will require some testing code setup, so I have deferred it to another PR.

GuptaManan100:
Personally, I would have liked to go a step further and cleanup a few more things like make `tablet_alias` as the primary key everywhere instead of `hostname,port` and explicitly make `cluster_name` as two fields `keyspace` and `shard` to formalize it even further, but I will do that in separate PRs to keep them easy to review and test.

deepthi (approved):
LGTM.
Thank you for the inline comments. They made this much easier to review.

shlomi-noach:
looks good to me!

PR #11194, opened 2022-09-09
Replication Manager Improvements

GuptaManan100:
## Description

This PR improves the replication manager. The proposed improvements are two-fold - 
1. The first is that the replication manager should be able to catch the case where the replication is broken because the primary's hostname and port changed due to kubernetes pod reschedule or some other reason. 
2. The second improvement is the addition of a flag that allows the replication manager to be turned off, if so desired.

Tests of both these changes have been added. An end to end test for the first one and a unit test for the other.

## Related Issue(s)
#9809 

## Checklist

### DIFF
--- a/go/flags/endtoend/vttablet.txt
+++ b/go/flags/endtoend/vttablet.txt
@@ -181,8 +181,8 @@ Usage of vttablet:
       --dba_idle_timeout duration                                        Idle timeout for dba connections (default 1m0s)
       --dba_pool_size int                                                Size of the connection pool for dba connections (default 20)
       --degraded_threshold duration                                      replication lag after which a replica is considered degraded (default 30s)
+      --disable-replication-manager                                      Disable replication manager to prevent replication repairs.
       --disable_active_reparents                                         if set, do not allow active reparents. Use this to protect a cluster using external reparents.
-      --disable_replication_manager                                      Disable replication manager to prevent replication repairs.
       --emit_stats                                                       If set, emit stats to push-based monitoring and stats backends
       --enable-consolidator                                              Synonym to -enable_consolidator (default true)
       --enable-consolidator-replicas                                     Synonym to -enable_consolidator_replicas
--- a/go/flags/endtoend/vttablet.txt
+++ b/go/flags/endtoend/vttablet.txt
@@ -182,6 +182,7 @@ Usage of vttablet:
       --dba_pool_size int                                                Size of the connection pool for dba connections (default 20)
       --degraded_threshold duration                                      replication lag after which a replica is considered degraded (default 30s)
       --disable_active_reparents                                         if set, do not allow active reparents. Use this to protect a cluster using external reparents.
+      --disable_replication_manager                                      Disable replication manager to prevent replication repairs.
       --emit_stats                                                       If set, emit stats to push-based monitoring and stats backends
       --enable-consolidator                                              Synonym to -enable_consolidator (default true)
       --enable-consolidator-replicas                                     Synonym to -enable_consolidator_replicas
--- a/go/test/endtoend/tabletmanager/replication_manager/tablet_test.go
+++ b/go/test/endtoend/tabletmanager/replication_manager/tablet_test.go
@@ -1,0 +1,191 @@
+/*
+Copyright 2022 The Vitess Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package tabletmanager
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"os"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+
+	"vitess.io/vitess/go/test/endtoend/cluster"
+	replicationdatapb "vitess.io/vitess/go/vt/proto/replicationdata"
+	tabletpb "vitess.io/vitess/go/vt/proto/topodata"
+	tmc "vitess.io/vitess/go/vt/vttablet/grpctmclient"
+)
+
+var (
+	clusterInstance *cluster.LocalProcessCluster
+	tmClient        *tmc.Client
+	primaryTablet   cluster.Vttablet
+	replicaTablet   cluster.Vttablet
+	hostname        = "localhost"
+	keyspaceName    = "ks"
+	cell            = "zone1"
+	sqlSchema       = `
+	create table t1(
+		id bigint,
+		value varchar(16),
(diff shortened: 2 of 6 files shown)


go/test/endtoend/tabletmanager/replication_manager/tablet_test.go
```diff
@@ -0,0 +1,191 @@
...
+	err := waitForSourcePort(ctx, t, replicaTablet, int32(primaryTablet.MySQLPort))
+	require.NoError(t, err)
+
+	// Stop the primary tablet
+	stopTablet(t, primaryTablet)
+
+	// Change the MySQL port of the primary tablet
+	newMysqlPort := clusterInstance.GetAndReservePort()
+	primaryTablet.MySQLPort = newMysqlPort
+	primaryTablet.MysqlctlProcess.MySQLPort = newMysqlPort
```
mattlord, line 183: 👍

mattlord (approved):
Thanks! ❤️

go/flags/endtoend/vttablet.txt
```diff
@@ -182,6 +182,7 @@ Usage of vttablet:
       --dba_pool_size int                                                Size of the connection pool for dba connections (default 20)
       --degraded_threshold duration                                      replication lag after which a replica is considered degraded (default 30s)
       --disable_active_reparents                                         if set, do not allow active reparents. Use this to protect a cluster using external reparents.
+      --disable_replication_manager                                      Disable replication manager to prevent replication repairs.
```
mattlord, line 185: We should use dashes for new flags. That will avoid future work as we move all of them to dashes.

go/vt/vttablet/tabletmanager/replmanager.go
```diff
@@ -109,9 +109,14 @@ func (rm *replManager) checkActionLocked() {
 			return
 		}
 	} else {
-		// If only one of the threads is stopped, it's probably
-		// intentional. So, we don't repair replication.
-		if status.SQLHealthy() || status.IOHealthy() {
+		if status.SQLHealthy() && status.IOState == mysql.ReplicationStateConnecting && status.LastIOError != "" {
```
rsajwani, line 112: my proposed solution for better readability. Feel free to ignore it :). My purpose it to avoid empty if block.

onlySQLHealthy = status.SQLHealthy() && !status.IOHealthy()
onlyIOHealth = !status.SQLHealthy() && status.IOHealthy()

if onlySQLHealthy {
return
} else if onlyIOHealthy{
   if !mysql.ReplicationStateConnection || status.LastIOError == "" {
       return
   }
}

go/vt/vttablet/tabletmanager/replmanager.go
```diff
@@ -109,9 +109,14 @@ func (rm *replManager) checkActionLocked() {
...
+			// SQL thread is healthy, but IO thread is in Connecting state with an IO Error,
+			// then we know that no Vitess operation stopped the IO thread and it is having trouble connecting to the primary
+			// Maybe the primary host-port changed, or something else happened. This could be an ephemeral error, so we should
+			// try and fix it.
+		} else if status.SQLHealthy() || status.IOHealthy() {
+			// If only one of the threads is stopped, it's probably
```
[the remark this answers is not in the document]
deepthi: Good point. We should fix the comment. If both threads are healthy we don't need to do anything either.

deepthi (commented):
I've updated the PR description with to link #9809, since this covers that case as well.
Once the flag change is done, this should be good to go.

go/vt/vttablet/tabletmanager/replmanager.go
GuptaManan100: This would actually be incorrect. If only SQL is healthy, then we want to check the IO error not the other way around.

GuptaManan100: Instead I have these two alternates - 
```golang
// Cases to consider for replication repair - 
// 1. If both the threads are healthy, then there is nothing to repair.
// 2. If the IO thread is healthy but SQL thread isn't, then we assume it is intentional, and we don't repair replication.
// 3. If SQL thread is healthy but IO thread is in Connecting state with an IO Error, then we know that no Vitess operation stopped the IO thread 
//	  and it is having trouble connecting to the primary Maybe the primary host-port changed, or something else happened. 
//	  This could be an ephemeral error, so we should try and fix it.
// 4. If SQL thread is healthy but IO thread is stopped, then we assume it is intentional, and we don't repair replication.
// 5. Both the threads are unhealthy, then we need to repair replication
if status.IOHealthy() {
	// This covers cases 1 and 2.
	return
}
if status.SQLHealthy() && (status.IOState != mysql.ReplicationStateConnecting || status.LastIOError == "") {
	// This covers case 4.
	return
}
// We now attempt to repair replication.
// This covers cases 3 and 5.
```

and what I have now - 
```golang
if status.SQLHealthy() && status.IOState == mysql.ReplicationStateConnecting && status.LastIOError != "" {
	// SQL thread is healthy, but IO thread is in Connecting state with an IO Error,
	// then we know that no Vitess operation stopped the IO thread and it is having trouble connecting to the primary
	// Maybe the primary host-port changed, or something else happened. This could be an ephemeral error, so we should
	// try and fix it.
} else if status.SQLHealthy() || status.IOHealthy() {
	// If both threads are fine then there is no need to repair.
	// If only one of the threads is stopped, it's probably
	// intentional. So, we don't repair replication if either thread is healthy.
	return
}
```

I have gone through the exercise of proving that these two are functionally the same from first order logic perspective, so whichever you feel is more readable we'll keep that one.

go/flags/endtoend/vttablet.txt
GuptaManan100: Fixed

go/vt/vttablet/tabletmanager/rpc_replication.go
```diff
@@ -39,12 +39,14 @@ import (
...
 
 var setSuperReadOnly bool
+var disableReplicationManager bool
 
 func registerReplicationFlags(fs *pflag.FlagSet) {
 	fs.Bool("enable_semi_sync", false, "")
 	fs.MarkDeprecated("enable_semi_sync", "--enable_semi_sync is deprecated; please set the correct durability policy on the keyspace instead.")
 
 	fs.BoolVar(&setSuperReadOnly, "use_super_read_only", setSuperReadOnly, "Set super_read_only flag when performing planned failover.")
+	fs.BoolVar(&disableReplicationManager, "disable_replication_manager", disableReplicationManager, "Disable replication manager to prevent replication repairs.")
```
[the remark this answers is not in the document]
GuptaManan100: I think we want to keep them upper-cased going forward. They show up as statements in the help, so upper-case seems the way to go.

go/vt/vttablet/tabletmanager/replmanager.go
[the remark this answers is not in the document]
GuptaManan100: Fixed the comment

rsajwani (approved):

go/vt/vttablet/tabletmanager/replmanager.go
GuptaManan100: Updated code with first approach.

deepthi: Nice comments 💯

PR #11195, opened 2022-09-09
Add support for views in vtgate

systay:
## Description
This PR is a subtask for the RFC.

Adds support for views that are expanded in the vtgate. In the vschema, if the views definition is available, like so:

```json
  "keyspaces": {
    "user": {
      "sharded": true,
      "tables": {
        "user": {},
        "user_extra": {}
      },
      "views": {
        "user_details_view": "select user.id, user_extra.col from user join user_extra on user.id = user_extra.user_id"
      }
    }
```

they will be used to expand into a derived table. Example:

The query:

```sql
select * from user_details_view
```

would get expanded so it looks like:

```sql
select * from (
  select user.id, user_extra.col from user join user_extra on user.id = user_extra.user_id
) as user_details_view
```

## Related Issue(s)

- https://github.com/vitessio/vitess/issues/11559

## Checklist

-   [X] Documentation was added or is not required - It will be added as part of the RFC


go/vt/vtgate/vindexes/vschema.go
```diff
@@ -623,6 +653,48 @@ func (vschema *VSchema) FindTableOrVindex(keyspace, name string, tabletType topo
...
+
+	// We do this to make sure there is no shared state between uses of this AST
+	statement = sqlparser.CloneSelectStatement(statement)
+	sqlparser.Rewrite(statement, func(cursor *sqlparser.Cursor) bool {
+		col, ok := cursor.Node().(*sqlparser.ColName)
+		if ok {
+			cursor.Replace(sqlparser.NewColNameWithQualifier(col.Name.String(), col.Qualifier))
+		}
+		return true
+	}, nil)
```
harshit-gangal, line 694: could you explain why this rewriting is needed here

systay: you mean add a comment to it, right? the why is because of the stupid `Metadata` field in this stuct already being filled in.

go/vt/vtgate/vindexes/vschema_test.go
```diff
@@ -247,57 +247,30 @@ func TestUnshardedVSchemaValid(t *testing.T) {
...
-	dual := &Table{
-		Name:     sqlparser.NewIdentifierCS("dual"),
-		Keyspace: ks,
-		Type:     TypeReference,
-	}
-	want := &VSchema{
-		RoutingRules: map[string]*RoutingRule{},
-		uniqueTables: map[string]*Table{
-			"t1":   t1,
-			"dual": dual,
```
harshit-gangal, line 284: it is unclear why so many assertions are removed from this test file

go/vt/sqlparser/ast_rewriting.go
```diff
@@ -386,21 +400,36 @@ func (er *astRewriter) rewrite(cursor *Cursor) bool {
...
+			break
+		}
+		if SystemSchema(er.keyspace) {
+			if er.keyspace != "" && aliasTableName.Qualifier.IsEmpty() {
+				aliasTableName.Qualifier = NewIdentifierCS(er.keyspace)
+				node.Expr = aliasTableName
+				cursor.Replace(node)
+			}
+			break
+		}
```
harshit-gangal, line 419: `er.keyspace != ""` this should be true always as the earlier check `SystemSchema(er.keyspace)` is true.
other question is why this rewriting is needed here?

systay: 👍

go/vt/vtgate/vindexes/vschema_test.go
systay: we were using `!reflect.DeepEqual(got, want)` in a lot of places where I felt it was better to have test that exercise methods instead of testing the inner state of structs. testing the state makes the tests extremely fragile - a change in how we store the data, like this PR does, forces us to have to change ALL assertions. Instead, we want to keep the _behaviour_ of the code, not which internal state we use to achieve this behaviour.

My intention was not to just remove assertions, I wanted to replace them with behaviour tests.

go/vt/vtgate/vindexes/vschema_test.go
```diff
@@ -2033,11 +1765,11 @@ func TestSequence(t *testing.T) {
 	}
 	want := &VSchema{
 		RoutingRules: map[string]*RoutingRule{},
-		uniqueTables: map[string]*Table{
-			"seq":  seq,
-			"t1":   t1,
-			"t2":   t2,
-			"dual": dualb,
+		uniqueTables: map[string]*string{
```
systay, line 1768: @harshit-gangal this is the type of change I don't like having to do. I got tired of fixing tests in the middle of the file, so I kept some of them using golden tests

go/vt/vtgate/vindexes/vschema.go
arthurschreiber: What do you think about changing this rewriting to happen once when the vschema is loaded, instead of on each call to `FindView`? The rewriting does not seem to be dependent on `keyspace` or `name`.

deepthi:
As a new feature, this is worth writing up in an issue/RFC.
The UX is important, it means a change to how people currently define their views.
Some other things that come to mind:
- Are we adding `ALTER VSCHEMA CREATE VIEW` type functionality?
- How does this interact with existing support in online DDL for views (which works for unsharded, but not sharded)
- Do we expect to hit topo server object size limits?

deepthi (commented):
Is the description still accurate? We don't intend to store views in vschema, do we?
There was some offline discussion about storing them in `_vt`.

deepthi (dismissed):
Please create an issue / RFC with the proposed (I guess now implemented) design so that people can comment on that.

harshit-gangal:
> Is the description still accurate? We don't intend to store views in vschema, do we? There was some offline discussion about storing them in `_vt`.

The idea behind storing views in topo is that it is related to Vschema and today we store VSchema in topo. Therefore it becomes the obvious choice.
If we plan to move VSchema out of topo then, views will also move out naturally.

go/vt/topotools/vschema_ddl.go
```diff
@@ -237,6 +237,39 @@ func ApplyVSchemaDDL(ksName string, ks *vschemapb.Keyspace, alterVschema *sqlpar
 			Sequence: sequenceFqn,
 		}
 
+		return ks, nil
+	case sqlparser.CreateViewDDLAction:
+		createView := alterVschema.ViewSpec.CreateView
+		name := createView.ViewName.Name.String()
+		if _, ok := ks.Tables[name]; ok {
+			return nil, vterrors.NewErrorf(vtrpcpb.Code_ALREADY_EXISTS, vterrors.TableExists, "vschema: Table '%s' already exists in keyspace %s", name, ksName)
```
frouioui, line 245: Shouldn't we have an error that is more specific to views, such as:
```
vschema: View '%s' already exists in keyspace %s
```

go/test/endtoend/vtgate/queries/misc/misc_test.go
```diff
@@ -112,3 +112,15 @@ func TestQueryTimeout(t *testing.T) {
 	_, err = utils.ExecAllowError(t, mcmp.VtConn, "select /*vt+ PLANNER=gen4 QUERY_TIMEOUT_MS=10 */ sleep(0.001) from dual")
 	assert.NoError(t, err)
 }
+
+func TestViews(t *testing.T) {
+	mcmp, closer := start(t)
+	defer closer()
+
```
frouioui, line 119: We should also add an end-to-end test that will run VTctld commands to create a view (`CreateViewDDLAction`), run queries using that view, expect the queries to work, and then drop (`DropViewDDLAction`) the view and expect queries to fail.

We can also add another test to test the different error paths defined in `go/vt/topotools/vschema_ddl.go`, where we try to create a view with the same name as an already existing table, where we try to drop a view that does not exist, etc.

I reckon we have unit tests for this in `go/vt/vtgate/executor_vschema_ddl_test.go`, but we don't have E2E tests.

doc/releasenotes/16_0_0_summary.md
```diff
@@ -9,6 +9,31 @@
...
+  "tables": {},
+  "views": {
+    "view1": "select * from t1",
+    "view2": "select * from t2",
+  }
+}
+```
+
+This can also be modified using VSchema DDLs.
+`alter vschema <create view definition> | <drop view definition>`
```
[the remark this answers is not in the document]
harshit-gangal: We have not natively supported modifying VSchema with MySQL DDLs, if were are taking that approach then it is a paradigm shift in how we do DDLs. Currently, this leaves the option to users of unsharded keyspace that they want to create the views in underlying MySQL or only at the VTGate. 

We can still support this syntax and later can even change the create/drop view planner/exec to modify VSchema directly.

go/test/endtoend/vtgate/queries/misc/misc_test.go
harshit-gangal: There is no vtctld commands for create or drop view.

e2e tests are costly to add, unit tests suffice the purpose here.

frouioui: Right, I meant the DDL queries you've added :)

GuptaManan100 (approved):
lgtm

GuptaManan100:
Are we storign the views in `_vt` now?

go/vt/proto/vschema/vschema.pb.go
```diff
@@ -156,7 +156,8 @@ type Keyspace struct {
 	Vindexes map[string]*Vindex `protobuf:"bytes,2,rep,name=vindexes,proto3" json:"vindexes,omitempty" protobuf_key:"bytes,1,opt,name=key,proto3" protobuf_val:"bytes,2,opt,name=value,proto3"`
 	Tables   map[string]*Table  `protobuf:"bytes,3,rep,name=tables,proto3" json:"tables,omitempty" protobuf_key:"bytes,1,opt,name=key,proto3" protobuf_val:"bytes,2,opt,name=value,proto3"`
 	// If require_explicit_routing is true, vindexes and tables are not added to global routing
-	RequireExplicitRouting bool `protobuf:"varint,4,opt,name=require_explicit_routing,json=requireExplicitRouting,proto3" json:"require_explicit_routing,omitempty"`
+	RequireExplicitRouting bool              `protobuf:"varint,4,opt,name=require_explicit_routing,json=requireExplicitRouting,proto3" json:"require_explicit_routing,omitempty"`
+	Views                  map[string]string `protobuf:"bytes,5,rep,name=views,proto3" json:"views,omitempty" protobuf_key:"bytes,1,opt,name=key,proto3" protobuf_val:"bytes,2,opt,name=value,proto3"`
```
GuptaManan100, line 160: We won't need this then, right?

frouioui:
> Are we storign the views in _vt now?

@GuptaManan100, yes this will be done in https://github.com/vitessio/vitess/pull/11896

go/vt/proto/vschema/vschema.pb.go
harshit-gangal: this would still be needed for internal storage, in the next PR https://github.com/vitessio/vitess/pull/11896 
I will see if this can be avoided.

doc/releasenotes/16_0_0_summary.md
```diff
@@ -9,6 +9,31 @@
...
+It can be overridden by setting the `query_timeout` session variable.
 Setting it as command line directive with `QUERY_TIMEOUT_MS` will override other values.
+
+### Mysql Compatibility
+
+#### Support for views
+
+Vitess now supports views in sharded keyspace. Views are not created on the underlying database but are logically stored
+in vschema.
+Any query using view will get re-rewritten as derived table during query planning.
```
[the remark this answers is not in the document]
harshit-gangal: This will be updated at a later stage, it is just a part of the feature support.

go/test/endtoend/vtgate/queries/misc/misc_test.go
harshit-gangal: does not apply anymore

go/vt/proto/vschema/vschema.pb.go
harshit-gangal: The reason I kept it is to show the e2e test and plan test works. But, may be not important.

GuptaManan100 (approved):

PR #11208, opened 2022-09-12
Flakes: Eliminate TestVreplicationCopyThrottling Flakes

mattlord:
## Description

`TestVreplicationCopyThrottling` has been somewhat flaky over the last week. I suspect it was due to the CI slowness experienced during GitHub's recent partial outages (due to a DDOS attack, I believe). Whatever the causes, this PR makes two changes which should eliminate the failures that I could find:
1. Extends the transaction timeout, as we rely on holding open transactions for some time with many changes (I saw failures were the transaction timed out during the final `rollback` statement). 
2. Adds a waitFor function that waits for the expected innodb history list length before moving forward (I didn't see any failures in the CI for this but I did see it locally)

I ran this locally and did not see any more failures:
```
$ for try in {1..100}; do go clean -testcache; go test -timeout 30m -v -failfast -run TestVreplicationCopyThrottling vitess.io/vitess/go/test/endtoend/vreplication; if [[ $? != 0 ]]; then echo "Failed on try # ${try}" && say "Danger, danger, yooooooo"; break; fi; done
```

## Checklist

-   [x] Tests were modified

### DIFF
--- a/go/test/endtoend/vreplication/vreplication_test.go
+++ b/go/test/endtoend/vreplication/vreplication_test.go
@@ -68,6 +68,8 @@ const (
 	openTxQuery       = "insert into customer(cid, name, typ, sport, meta) values(4, 'openTxQuery',1,'football,baseball','{}');"
 	deleteOpenTxQuery = "delete from customer where name = 'openTxQuery'"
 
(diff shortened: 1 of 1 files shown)


rohit-nayak-ps (approved):
lgtm

PR #11209, opened 2022-09-12
Fix build errors

frouioui:
## Description

This PR fixes two compilation errors observed when running `govulncheck ./...`.

## Checklist


tools/rowlog/rowlog.go
```diff
@@ -301,10 +301,12 @@ func processRowEvent(plan *TablePlan, gtid string, ev *binlogdatapb.VEvent) []*R
 			before = sqltypes.MakeRowTrusted(plan.fields, change.Before)
 		}
 		for _, val := range after {
-			afterVals = append(afterVals, string(val.ToBytes()))
+			bytes, _ := val.ToBytes()
```
deepthi, line 304: not good to ignore errors.
@rohit-nayak-ps should we log an error and move on, or fail?

rohit-nayak-ps: We could just panic , to exit the rowlog tool, in case of an errorhere , since we are just parsing binlogs and it should not normally occur. Of if it did we would have already seen failing vreplication workflows.

frouioui: Fixed via d15ec9eec8

rohit-nayak-ps (approved):
lgtm

Issue #11210, opened 2022-09-12
Bug Report: VReplication does not properly replicate decimal(N,0) fields with a value of 0


mattlord:
### Overview of the Issue

If you execute a `MoveTables` workflow and one of the tables involved has a decimal field with a scale of 0 (e.g. `decimal(8,0)`) and a value of 0 is specified in the binlog stream (replicated changes after the copy phase completes), the row change for that field is not properly replicated. This is because we're not properly translating the RBR after image in this case. The after image has this:
```
### SET
###   @1='bbbbb' /* VARSTRING(128) meta=128 nullable=0 is_null=0 */
###   @2='test' /* VARSTRING(128) meta=128 nullable=1 is_null=0 */
###   @3=0 /* DECIMAL(8,0) meta=2048 nullable=1 is_null=0 */
```

But once we've processed it we have a 0 length value for the field:
```
replicator_plan.go:371] vrepl_decimal.applyChange.after: fields: [name:"sku" type:VARBINARY table:"product" org_table:"product" database:"vt_commerce" org_name:"sku" column_length:128 charset:63 flags:20611 column_type:"varbinary(128)" name:"description" type:VARBINARY table:"product" org_table:"product" database:"vt_commerce" org_name:"description" column_length:128 charset:63 flags:128 column_type:"varbinary(128)" name:"price" type:DECIMAL table:"product" org_table:"product" database:"vt_commerce" org_name:"price" column_length:9 charset:63 flags:32768 column_type:"decimal(8,0)"], rowChange.After: lengths:5 lengths:4 lengths:0 values:"bbbbbtest"
```

Which results in an empty `DECIMAL()` bind value:
```
replicator_plan.go:373] vrepl_decimal.applyChange.after: fields: [name:"sku" type:VARBINARY table:"product" org_table:"product" database:"vt_commerce" org_name:"sku" column_length:128 charset:63 flags:20611 column_type:"varbinary(128)" name:"description" type:VARBINARY table:"product" org_table:"product" database:"vt_commerce" org_name:"description" column_length:128 charset:63 flags:128 column_type:"varbinary(128)" name:"price" type:DECIMAL table:"product" org_table:"product" database:"vt_commerce" org_name:"price" column_length:9 charset:63 flags:32768 column_type:"decimal(8,0)"], vals: [VARBINARY("bbbbb") VARBINARY("test") DECIMAL()]
```

The final user visible result is an error for the VReplication workflow like this:
```
[...]

PR #11212, opened 2022-09-13
VReplication: Handle DECIMAL 0 Value Edge Case

mattlord:
## Description

With a 0 scale decimal field, with a value of 0, the parsed RBR event has an empty 0 byte length value. This then resulted in a `DECIMAL()` value getting used as the bind variable that in turn resulted in an invalid SQL query. More details can be found in [the issue](https://github.com/vitessio/vitess/issues/11210).

## Related Issue(s)

  - Fixes: https://github.com/vitessio/vitess/issues/11210

## Checklist

-   [x] Tests were added


rohit-nayak-ps (approved):
nice catch. lgtm

deepthi (commented):
This is a model PR. There is a bug report that documents the issue, there are code comments explaining the fix and there are targeted unit tests.
Nice work 💯

PR #11213, opened 2022-09-13
Add Metric For Time Elapsed In Getting Connection In Pools

GuptaManan100:
## Description

This PR added a metric called `GetConnTime` with labels `with` and `without` for `settings`, that publishes the amount of time it took to get a connection with and without having settings respectively.

## Related Issue(s)

- #9706 

## Checklist

### DIFF
--- a/go/vt/vttablet/tabletserver/connpool/pool.go
+++ b/go/vt/vttablet/tabletserver/connpool/pool.go
@@ -26,7 +26,6 @@ import (
 
 	"vitess.io/vitess/go/netutil"
 	"vitess.io/vitess/go/pools"
-	"vitess.io/vitess/go/stats"
 	"vitess.io/vitess/go/sync2"
 	"vitess.io/vitess/go/trace"
 	"vitess.io/vitess/go/vt/callerid"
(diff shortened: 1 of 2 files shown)


go/vt/vttablet/tabletserver/connpool/pool.go
```diff
@@ -200,16 +205,18 @@ func (cp *Pool) Get(ctx context.Context, setting *pools.Setting) (*DBConn, error
...
-	timeElapsed := time.Since(start)
-	if setting == nil {
-		cp.getConnTime.Set("without", timeElapsed.Nanoseconds())
-	} else {
-		cp.getConnTime.Set("with", timeElapsed.Nanoseconds())
+	if cp.getConnTime != nil {
+		if setting == nil {
+			cp.getConnTime.Record(getWithoutS, start)
+		} else {
+			cp.getConnTime.Record(getWithS, start)
```
GuptaManan100, line 218: This is just recording the time when the Get function was called. It is not recording the time that it took to finish execution. Is that what we want? If that is the case, then the metric's help should change, because currently, it says `Tracks the amount of time it takes to get a connection`

harshit-gangal: the record function does the diff for us. it is a helper function to the Add in timing stats.

harshit-gangal:
```go
// Record is a convenience function that records completion
// timing data based on the provided start time of an event.
func (t *Timings) Record(name string, startTime time.Time) {
	if t.labelCombined {
		name = StatsAllStr
	}
	t.Add(name, time.Since(startTime))
}

```

GuptaManan100: Oh okay 👍

PR #11215, opened 2022-09-13
Add getNumOfRows() in Java client to get result size for select queries

roderickyao:
## Description
Add a getNumOfRows() method in Java client to return the number of rows query got. Currently there is only one method to return the number of affected rows. But it is not useful for SELECT queries. 

## Checklist

## Deployment Notes
Only affecting the java client. Not impact to vitess.


harshit-gangal (approved):

harshit-gangal:
There is one CI test stuck, could you rebase and push for it to re-run

PR #11217, opened 2022-09-14
Online DDL: more info in a conflicting migration message

shlomi-noach:
## Description

Today, when the Online DDL executor attempts to run a migration that has a conflict, vitess returns a generic error message: 

> cannot run migration since a migration is already running

As of this PR, the message indicates the UUID and table name that cause the conflict. UUID is enough, but table name is added for more clarity.

## Related Issue(s)

#6926

## Checklist


rohit-nayak-ps (approved):
lgtm

PR #11220, opened 2022-09-14
Only refresh required tablet's information in VTOrc

GuptaManan100:
## Description

In #10115 and #10150 we added the capability of refreshing VTOrc's ephemeral information before it ran any fix. This was required to help us guarantee safety however, the first iteration of this change was inefficient.

We used to refresh all the tablets that are in the VTOrc instance's purview for each and every recovery.  As part of the PRs a TODO for fixing this was also added - 
```golang
// TODO (@GuptaManan100): Refresh only the shard tablet information instead of all the tablets
```

This change couldn't be immediately accomplished because we first required the cleanup of `cluster_alias`, `cluster_name`, and `suggested_cluster_alias`. This cleanup was addressed in #11193.

This PR addresses this `TODO` that was introduced then in order to make the recoveries more efficient and faster. Instead of the proposed refreshing all tablets in a shard in the TODO, this PR takes it a step further and only refreshes the tablets that are required.

To this end, all the recoveries have been categorized into two types, the ones that are cluster-wide recoveries and the ones that aren't.

If we are about to run a cluster-wide recovery like `electNewPrimary` or `recoverDeadPrimary`, then it is imperative to first refresh all the tablets of a shard because a new tablet could have been promoted, and we need to have this visibility before we run a cluster operation of our own.

Non-cluster-wide recoveries are only concerned with the specific tablet on which the failure occurred and the primary instance of the shard. For example, ConnectedToWrongPrimary analysis only cares for whom the current primary tablet is and the host-port set on the tablet in question. So, we only need to refresh the tablet info records (to know if the primary tablet has changed), and the replication data of the new primary and this tablet.

These changes make VTOrc recoveries much faster, while still guaranteeing correctness like they used to 🧞 

## Related Issue(s)

 - #8975

## Checklist

### DIFF
--- a/go/vt/orchestrator/inst/analysis.go
+++ b/go/vt/orchestrator/inst/analysis.go
@@ -122,6 +122,7 @@ type ReplicationAnalysis struct {
 	AnalyzedInstanceDataCenter                string
 	AnalyzedInstanceRegion                    string
 	AnalyzedKeyspace                          string
+	AnalyzedShard                             string
 	AnalyzedInstancePhysicalEnvironment       string
 	AnalyzedInstanceBinlogCoordinates         BinlogCoordinates
 	IsPrimary                                 bool
--- a/go/vt/orchestrator/logic/orchestrator.go
+++ b/go/vt/orchestrator/logic/orchestrator.go
@@ -452,7 +452,7 @@ func ContinuousDiscovery() {
 			}()
 		case <-tabletTopoTick:
 			go RefreshAllKeyspaces()
-			go RefreshTablets(false /* forceRefresh */)
+			go refreshAllTablets()
 		}
 	}
 }
--- a/go/vt/orchestrator/logic/orchestrator.go
+++ b/go/vt/orchestrator/logic/orchestrator.go
@@ -256,6 +256,10 @@ func DiscoverInstance(instanceKey inst.InstanceKey, forceDiscovery bool) {
 		return
 	}
 
+	if forceDiscovery {
+		log.Infof("Force discovered - %+v", instance)
+	}
+
 	discoveryMetrics.Append(&discovery.Metric{
 		Timestamp:       time.Now(),
 		InstanceKey:     instanceKey,
--- a/go/vt/orchestrator/logic/tablet_discovery.go
+++ b/go/vt/orchestrator/logic/tablet_discovery.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"errors"
 	"flag"
-	"fmt"
 	"strings"
 	"sync"
 	"sync/atomic"
@@ -69,11 +68,11 @@ func OpenTabletDiscovery() <-chan time.Time {
 	return time.Tick(15 * time.Second) //nolint SA1015: using time.Tick leaks the underlying ticker
 }
 
-// RefreshTablets reloads the tablets from topo.
-func RefreshTablets(forceRefresh bool) {
+// refreshAllTablets reloads the tablets from topo and discovers the ones which haven't been refreshed in a while
+func refreshAllTablets() {
 	refreshTabletsUsing(func(instanceKey *inst.InstanceKey) {
-		DiscoverInstance(*instanceKey, forceRefresh)
-	}, forceRefresh)
+		DiscoverInstance(*instanceKey, false)
+	}, false)
 }
 
 func refreshTabletsUsing(loader func(instanceKey *inst.InstanceKey), forceRefresh bool) {
@@ -155,6 +154,31 @@ func refreshTabletsInCell(ctx context.Context, cell string, loader func(instance
 	query := "select hostname, port, info from vitess_tablet where cell = ?"
 	args := sqlutils.Args(cell)
 	refreshTablets(tablets, query, args, loader, forceRefresh)
+}
+
+// forceRefreshAllTabletsInShard is used to refresh all the tablet's information (both MySQL information and topo records)
+// for a given shard. This function is meant to be called before or after a cluster-wide operation that we know will
+// change the replication information for the entire cluster drastically enough to warrant a full forceful refresh
+func forceRefreshAllTabletsInShard(ctx context.Context, keyspace, shard string) {
+	log.Infof("force refresh of all tablets in shard - %v/%v", keyspace, shard)
+	if ctx == nil {
+		refreshCtx, refreshCancel := context.WithTimeout(context.Background(), *topo.RemoteOperationTimeout)
+		defer refreshCancel()
+		ctx = refreshCtx
+	}
+	refreshTabletsInKeyspaceShard(ctx, keyspace, shard, func(instanceKey *inst.InstanceKey) {
+		DiscoverInstance(*instanceKey, true)
+	}, true)
+}
+
+// refreshTabletInfoOfShard only refreshes the tablet records from the topo-server for all the tablets
+// of the given keyspace-shard.
+func refreshTabletInfoOfShard(ctx context.Context, keyspace, shard string) {
+	log.Infof("refresh of tablet records of shard - %v/%v", keyspace, shard)
+	refreshTabletsInKeyspaceShard(ctx, keyspace, shard, func(instanceKey *inst.InstanceKey) {
+		// No-op
+		// We only want to refresh the tablet information for the given shard
+	}, false)
 }
 
 func refreshTabletsInKeyspaceShard(ctx context.Context, keyspace, shard string, loader func(instanceKey *inst.InstanceKey), forceRefresh bool) {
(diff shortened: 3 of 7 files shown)


go/vt/orchestrator/inst/analysis.go
```diff
@@ -122,6 +122,7 @@ type ReplicationAnalysis struct {
 	AnalyzedInstanceDataCenter                string
 	AnalyzedInstanceRegion                    string
 	AnalyzedKeyspace                          string
+	AnalyzedShard                             string
```
GuptaManan100, line 125: We need the name of the shard that was analyzed too now that we want to restrict the number of tablets we want to refresh

go/vt/orchestrator/logic/orchestrator.go
```diff
@@ -256,6 +256,10 @@ func DiscoverInstance(instanceKey inst.InstanceKey, forceDiscovery bool) {
 		return
 	}
 
+	if forceDiscovery {
+		log.Infof("Force discovered - %+v", instance)
+	}
```
GuptaManan100, line 261: This addition of logging is intentional. Until we have a metrics page where we export the internal database information of VTOrc, this is going to be very useful in debugging. I had it in my mind to add this log and I am just piggy-backing on this PR.

go/vt/orchestrator/logic/tablet_discovery.go
```diff
@@ -285,21 +309,29 @@ func setReplicationSource(ctx context.Context, replica *topodatapb.Tablet, prima
...
+	LIMIT 1
+`
+	err = db.Db.QueryOrchestrator(query, sqlutils.Args(keyspace, shard), func(m sqlutils.RowMap) error {
+		if primary == nil {
+			primary = &topodatapb.Tablet{}
+			return prototext.Unmarshal([]byte(m.GetString("info")), primary)
+		}
+		return nil
+	})
+	return primary, err
```
GuptaManan100, line 336: This is another small enhancement that was made in this PR, wherein we can use the tablet information we have collected to find the shard primary tablet. We use something similar in `GetReplicationAnalysis` to find who the primary is. 
Previously we used to do a topo-server call to read the shard and tablet record, but it isn't required.

go/vt/orchestrator/logic/topology_recovery.go
```diff
@@ -1010,10 +1107,7 @@ func executeCheckAndRecoverFunction(analysisEntry inst.ReplicationAnalysis, cand
 // checkIfAlreadyFixed checks whether the problem that the analysis entry represents has already been fixed by another agent or not
 func checkIfAlreadyFixed(analysisEntry inst.ReplicationAnalysis) (bool, error) {
 	// Run a replication analysis again. We will check if the problem persisted
-	// TODO (@GuptaManan100): Use specific cluster name to filter the scope of replication analysis.
-	// Can't do this now since SuggestedClusterAlias, ClusterName, ClusterAlias aren't consistent
-	// and passing any one causes issues in some failures
-	analysisEntries, err := inst.GetReplicationAnalysis("", &inst.ReplicationAnalysisHints{})
+	analysisEntries, err := inst.GetReplicationAnalysis(analysisEntry.ClusterDetails.ClusterName, &inst.ReplicationAnalysisHints{})
```
GuptaManan100, line 1110: This is a change that could have happened in #11193 but I am piggy-backing on this PR to not create a separate one just for this change.

go/vt/orchestrator/logic/tablet_discovery.go
```diff
@@ -157,6 +158,31 @@ func refreshTabletsInCell(ctx context.Context, cell string, loader func(instance
...
+	refreshTabletsInKeyspaceShard(ctx, keyspace, shard, func(instanceKey *inst.InstanceKey) {
+		DiscoverInstance(*instanceKey, true)
+	}, true)
+}
+
+// refreshTabletInfoOfShard only refreshes the tablet records from the topo-server for all the tablets
+// of the given keyspace-shard.
+func refreshTabletInfoOfShard(ctx context.Context, keyspace, shard string) {
+	log.Infof("refresh of tablet records of shard - %v/%v", keyspace, shard)
+	refreshTabletsInKeyspaceShard(ctx, keyspace, shard, func(instanceKey *inst.InstanceKey) {
```
rsajwani, line 180: we don't need context.background here?

go/vt/orchestrator/logic/tablet_discovery.go
```diff
@@ -285,21 +311,32 @@ func setReplicationSource(ctx context.Context, replica *topodatapb.Tablet, prima
...
-	}
-	// TODO(GuptaManan100): Instead of another topo call, use the local information by calling
-	// ReadTablet. Currently this isn't possible since we only have the primary alias and not the source host and port
-	// This should be fixed once the tablet alias is changed to be the primary key of the table
-	primaryInfo, err := ts.GetTablet(ctx, si.PrimaryAlias)
-	if err != nil {
-		return nil, err
+// shardPrimary finds the primary of the given keyspace-shard by reading the orchestrator backend
+func shardPrimary(keyspace string, shard string) (primary *topodatapb.Tablet, err error) {
+	query := `SELECT
```
rsajwani, line 316: general comment: should we not put all query execution in some retryable template function?

rsajwani (approved):
Overall LGTM.

go/vt/orchestrator/logic/tablet_discovery.go
```diff
@@ -69,11 +70,11 @@ func OpenTabletDiscovery() <-chan time.Time {
...
 
-// RefreshTablets reloads the tablets from topo.
-func RefreshTablets(forceRefresh bool) {
+// refreshAllTablets reloads the tablets from topo and discovers the ones which haven't been refreshed in a while
+func refreshAllTablets() {
 	refreshTabletsUsing(func(instanceKey *inst.InstanceKey) {
-		DiscoverInstance(*instanceKey, forceRefresh)
-	}, forceRefresh)
+		DiscoverInstance(*instanceKey, false)
+	}, false)
```
deepthi, line 77:
```suggestion
		DiscoverInstance(*instanceKey, false /* forceDiscovery */)
	}, false /* forceRefresh */)
```

go/vt/orchestrator/logic/tablet_discovery.go
```diff
@@ -157,6 +158,31 @@ func refreshTabletsInCell(ctx context.Context, cell string, loader func(instance
...
 
+// forceRefreshAllTabletsInShard is used to refresh all the tablet's information (both MySQL information and topo records)
+// for a given shard. This function is meant to be called before or after a cluster-wide operation that we know will
+// change the replication information for the entire cluster drastically enough to warrant a full forceful refresh
+func forceRefreshAllTabletsInShard(ctx context.Context, keyspace, shard string) {
+	log.Infof("force refresh of all tablets in shard - %v/%v", keyspace, shard)
+	if ctx == context.Background() {
+		refreshCtx, refreshCancel := context.WithTimeout(context.Background(), *topo.RemoteOperationTimeout)
+		defer refreshCancel()
+		ctx = refreshCtx
```
deepthi, line 169: I don't know if it is good practice to override the passed-in context like this..

go/vt/orchestrator/logic/tablet_discovery.go
```diff
@@ -285,21 +311,32 @@ func setReplicationSource(ctx context.Context, replica *topodatapb.Tablet, prima
...
+	FROM 
+		vitess_tablet
+	WHERE
+		keyspace = ? AND shard = ?
+		AND tablet_type = ?
+	ORDER BY
+		primary_timestamp DESC
+	LIMIT 1
+`
+	err = db.Db.QueryOrchestrator(query, sqlutils.Args(keyspace, shard, topodatapb.TabletType_PRIMARY), func(m sqlutils.RowMap) error {
```
deepthi, line 331: As part of cleaning up, we should rename some of these functions. `QueryOrchestrator` is not the right name for this function. Similarly we have `OpenOrchestrator` too.

go/vt/orchestrator/logic/tablet_discovery.go
deepthi: why do we need to check for the background context at all? we can apply `RemoteOperationTimeout` regardless. If the passed-in context has a deadline that is closer than current time + `RemoteOperationTimeout` that will anyway take precedence.

go/vt/topotools/tablet.go
```diff
@@ -246,3 +246,32 @@ func TabletIdent(tablet *topodatapb.Tablet) string {
 func TargetIdent(target *querypb.Target) string {
 	return fmt.Sprintf("%s/%s (%s)", target.Keyspace, target.Shard, target.TabletType)
 }
+
+// TabletEquality returns true iff two Tablets are identical for testing purposes
+func TabletEquality(left, right *topodatapb.Tablet) bool {
```
deepthi, line 251: Why do we need a custom function? Why can't we use `cmp.Comparer`?
https://github.com/golang/protobuf/issues/1238#issuecomment-723547616

go/vt/orchestrator/logic/tablet_discovery.go
GuptaManan100: No, this happens before the recovery starts. We should be using the ctx passed in.

go/vt/orchestrator/logic/tablet_discovery.go
GuptaManan100: Yes, that would be a good addition, but so far not really needed, because even if the read fails, we just fail the recovery and then retry later.

go/vt/orchestrator/logic/tablet_discovery.go
GuptaManan100: I like this suggestion. I'll change the code.

go/vt/orchestrator/logic/topology_recovery.go
```diff
@@ -954,12 +1008,43 @@ func executeCheckAndRecoverFunction(analysisEntry inst.ReplicationAnalysis, cand
...
+			if err != nil {
+				log.Errorf("executeCheckAndRecoverFunction: Analysis: %+v, InstanceKey: %+v, candidateInstanceKey: %+v, "+"skipProcesses: %v: error while finding the shard primary: %v",
+					analysisEntry.Analysis, analysisEntry.AnalyzedInstanceKey, candidateInstanceKey, skipProcesses, err)
+				return false, nil, err
+			}
+			primaryInstanceKey := inst.InstanceKey{
+				Hostname: primaryTablet.MysqlHostname,
+				Port:     int(primaryTablet.MysqlPort),
+			}
+			// We can skip the refresh if we the tablet we are looking at is the primary tablet.
```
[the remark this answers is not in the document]
GuptaManan100: Fixed

go/vt/orchestrator/logic/tablet_discovery.go
GuptaManan100: Fixed

go/vt/orchestrator/logic/tablet_discovery.go
GuptaManan100: Fixed

go/vt/topotools/tablet.go
GuptaManan100: We don't need the custom function. `proto.Equal` in `cmp.Diff` is working just fine. I have changed the code and removed that function.

go/vt/orchestrator/logic/tablet_discovery.go
GuptaManan100: Yes, I had it in my mind to get rid of the `Orchestrator` references everywhere, from Parameters, flags, package names, function names, etc. I'll do in a follow-up PR so that it is easier to review

Issue #11222, opened 2022-09-14
Bug Report: vtcombo mutates options


olyazavr:
### Overview of the Issue

When using vtcombo to run unit tests, we noticed that for transactional tests, transactions didn't seem to be obeyed. Upon further inspection, statements inside of a transaction would always get autocommitted. We could not reproduce this with actual vtgates/vttablets, so it had to be isolated to vtcombo. Turns out, the `options` [here](https://github.com/vitessio/vitess/blob/b277bcb871e847c6fb280431d6cf8e62aa455ff7/go/vt/vtcombo/tablet_map.go#L441-L454) would get mutated [here](https://github.com/vitessio/vitess/blob/1487772f3d59c5fc6fb221402a5100f90c69857f/go/vt/vttablet/tabletserver/query_executor.go#L199), so any previous query that was indeed autocommit=true and not in a transaction would cause future queries to have `qre.options.TransactionIsolation = querypb.ExecuteOptions_AUTOCOMMIT` set

Slack conversation: https://vitess.slack.com/archives/C0PQY0PTK/p1662725270016499

### Reproduction Steps

Using vtcombo, run a unit test with the following:
1. Create table
2. Insert 3 rows with ids 1,2,3..100
3. In one thread, in a transaction update rows with ids 1,2,3..100
4. In another thread, in a transaction update rows with ids 100, 99, 98...1 (in reverse of step 3)
5. Expect a deadlock, but since in step 2 we were not in transaction and it did autocommit, the other statements inherit this and do not deadlock as each one autocommits

### Binary Version

```sh
v13.0.0, commit bc4a960
```

### Operating System and Environment details

```sh
Can repro on both macos and linux, running unit tests in Java
```

Issue #10398, opened 2022-05-31
Bug Report: transaction not rolled back after insert


tdryer:
### Overview of the Issue

When executing the following queries on the same connection, I'm observing that the `UPDATE` is not rolled back by the transaction as expected, leaving the inserted row modified (`test=1`) instead of in its original state (`test=0`):

```sql
INSERT INTO test_table (test) VALUES (0);
BEGIN;
UPDATE test_table SET test=1;
ROLLBACK;
```

### Reproduction Steps

1. Start a `vttestserver` container:
```
docker run --detach --rm --name vitess --env KEYSPACES=test --env NUM_SHARDS=1 \
    --env PORT=30000 vitess/vttestserver:mysql57
```

2. Wait for the container to be ready:
```
until docker exec vitess mysqladmin ping --host 127.0.0.1 --port 30003 \
    --silent; do sleep 1 ; done
```

3. Execute the query sequence:
```
docker exec vitess mysql --host 127.0.0.1 --port 30003 test -e '
CREATE TABLE `test_table` (`test` bigint unsigned NOT NULL, PRIMARY KEY (`test`));
INSERT INTO test_table (test) VALUES (0);
BEGIN;
UPDATE test_table SET test=1;
ROLLBACK;
SELECT * FROM test_table;
'
```
4. Observe from the result of the final query that the transaction was not rolled back as expected:
```
test
1
```
5. Stop the container:
```
docker stop --time 1 vitess
```

### Binary Version

```sh
$ docker exec vitess vtcombo --version
Version: 14.0.0-SNAPSHOT (Git revision f5222ab0aaae0cba35fc42b0b60790beae0247dd branch 'main') built on Mon May 30 15:40:28 UTC 2022 by vitess@buildkitsandbox using go1.18.1 linux/amd64
```

### Operating System and Environment details

```sh
$ cat /etc/os-release 
NAME="Ubuntu"
VERSION="20.04.4 LTS (Focal Fossa)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 20.04.4 LTS"
VERSION_ID="20.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=focal
[...]

PR #11223, opened 2022-09-14
vtcombo mutates options, make a copy to avoid this

olyazavr:
## Description

When using vtcombo to run unit tests, we noticed that for transactional tests, transactions didn't seem to be obeyed. Upon further inspection, statements inside of a transaction would always get autocommitted. We could not reproduce this with actual vtgates/vttablets, so it had to be isolated to vtcombo. Turns out, the `options` [here](https://github.com/vitessio/vitess/blob/b277bcb871e847c6fb280431d6cf8e62aa455ff7/go/vt/vtcombo/tablet_map.go#L441-L454) would get mutated [here](https://github.com/vitessio/vitess/blob/1487772f3d59c5fc6fb221402a5100f90c69857f/go/vt/vttablet/tabletserver/query_executor.go#L199), so any previous query that was indeed autocommit=true and not in a transaction would cause future queries to have `qre.options.TransactionIsolation = querypb.ExecuteOptions_AUTOCOMMIT` set

Slack conversation: https://vitess.slack.com/archives/C0PQY0PTK/p1662725270016499

Example:

1. Create table
2. Insert 3 rows with ids 1,2,3..100
3. In one thread, in a transaction update rows with ids 1,2,3..100
4. In another thread, in a transaction update rows with ids 100, 99, 98...1 (in reverse of step 3)
5. Expect a deadlock, but since in step 2 we were not in transaction and it did autocommit, the other statements inherit this and do not deadlock as each one autocommits

## Related Issue(s)

- Fixes https://github.com/vitessio/vitess/issues/11222
- Fixes https://github.com/vitessio/vitess/issues/10398

## Checklist

## Deployment Notes

None, only concerns vtcombo

### DIFF
--- a/go/test/endtoend/vtcombo/vttest_sample_test.go
+++ b/go/test/endtoend/vtcombo/vttest_sample_test.go
@@ -161,6 +161,7 @@ func TestStandalone(t *testing.T) {
 
 	assertInsertedRowsExist(ctx, t, conn, idStart, rowCount)
 	assertTabletsPresent(t)
+	assertTransactionalityAndRollbackObeyed(ctx, t, conn, idStart)
 }
 
 func assertInsertedRowsExist(ctx context.Context, t *testing.T, conn *vtgateconn.VTGateConn, idStart, rowCount int) {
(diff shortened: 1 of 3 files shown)


harshit-gangal (commented):
I think an e2e test would help see the fix that is done.

harshit-gangal (commented):
I think the correct way to fix would be to clone it here 
```go
func (qre *QueryExecutor) execAutocommit(f func(conn *StatefulConnection) (*sqltypes.Result, error)) (reply *sqltypes.Result, err error) {
	if qre.options == nil {
		qre.options = &querypb.ExecuteOptions{}
	}
	qre.options.TransactionIsolation = querypb.ExecuteOptions_AUTOCOMMIT
```

Also use `proto.Clone(qre.options)`, proto inbuild deep copy helper function for cloning.

harshit-gangal (approved):

Issue #11233, opened 2022-09-15
VReplication Workflows: Running vdiff on workflows with writes switched result in new incorrect running streams


rohit-nayak-ps:
### Overview of the Issue

For a workflow running a `VDiff` once `SwitchWrites` has been run results in new streams getting created incorrectly on the targets. These streams are configured to stream from the current target shards. Any writes to the keyspace then result in these workflows erroring out since they are trying to replay the same writes back to the targets.

These don't affect data integrity since the workflows never proceed and any reverse replication streams are untouched. However they keep erroring out in the logs resulting in ops alerts and uncertainty whether the reshard has been run correctly.

Here are some incidents that this error has created: https://vitess.slack.com/archives/C0PQY0PTK/p1663211793826149

### Reproduction Steps

* Run local example upto and including `305_switch_writes.sh`
* Now run `vtctlclient VDiff -- customer.cust2cust`
* Insert a row into the `customer` table
* `vtctlclient Workflow customer.cust2cust show` will show only two new vreplication streams and the shard into which the inserted row went will show a Duplicate Key error

### Binary Version

```sh
`main.b5d8281bb552585f817cbb74cce6ae6d943c9cde`
```

PR #11234, opened 2022-09-15
Don't run VDiff on frozen workflows

rohit-nayak-ps:
## Description

https://github.com/vitessio/vitess/issues/11233 has details on the issue this PR is trying to fix. 

The root cause is that `SwitchWrites` will `Stop` the workflows on the target since writes have now been switched to them. However `vdiff` restarts workflows once it is done irrespective of the previous status of the workflow. 

As part of `SwitchWrites` all streams in a workflow are marked as `frozen`.  This PR adds a check in `VDiff` (both v1 and v2) and does not start it in case the workflow is found to be frozen.

This is the most common issue that has been reported related to VDiff starting stopped workflows. Another use-case would be if the user had stopped a workflow for some reason.  That requires some more discussion: on the one hand we can look for Stopped workflow and not restart them. However, VDiff will temporarily restart all workflows while syncing up the source and target, which will mean that workflows will be running for a while and writing to the target. If the user had stopped the workflow to prevent that, for some reason, we will be overriding that. At this time we don't have a reasonable justification for this use-case and also it will take a bit longer to implement it. So we only fix the specific bug referred to in #11233  

## Related Issue(s)

Fixes https://github.com/vitessio/vitess/issues/11233

## Checklist

### DIFF
--- a/go/test/endtoend/vreplication/vreplication_test.go
+++ b/go/test/endtoend/vreplication/vreplication_test.go
@@ -550,6 +550,8 @@ func shardCustomer(t *testing.T, testReverse bool, cells []*Cell, sourceCellOrAl
 		}
 		switchWritesDryRun(t, ksWorkflow, dryRunResultsSwitchWritesCustomerShard)
 		switchWrites(t, ksWorkflow, false)
+		checkThatVDiffFails(t, targetKs, workflow)
+
 		if withOpenTx && commit != nil {
 			commit(t)
 		}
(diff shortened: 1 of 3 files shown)


mattlord:
I think that running VDiff on a stopped workflow is fine and could potentially be desired, but in that case we should skip restarting the workflow at the end. We don't HAVE to do that here though as this PR resolves the bigger problem noted in the issue.

rohit-nayak-ps:
> I think that running VDiff on a stopped workflow is fine and could potentially be desired, but in that case we should skip restarting the workflow at the end. We don't have to do that work here though as this PR resolves the bigger problem noted in the issue.

Thanks, let this PR then only fix #11233.  I don't think the solution for not restarting, initially stopped (but not frozen) workflows is not an obvious one, as noted in the description.

mattlord (approved):
❤️

Issue #10133, opened 2022-04-24
VTTablet _vt: mechanism to reach desired _vt schema needs refactoring


rohit-nayak-ps:
## TL;DR
Reaching desired `_vt` schema is currently fragile and we need a systemic fix. Either by using one of the existing methods better or coming with a new technique. 

## Motivation

There are two main reasons for this issue:
1. Rationalization of the process of creating/updating the `_vt` schema since it is currently difficult to reason and causes intermittent bugs/outages during version upgrades
2. It is very inefficient. Sidecar schema DDLs  are executed in several code paths. Usually they are noops since we already have reached the desired schema. However running these queries do take up non-trivial amount of time. Since they are not in critical paths today it is not affecting the user. However if we want to run them all at tablet init it takes several seconds.

### History

Several Vitess modules use the `_vt` sidecar database. The schema of this database can keep changing as we add more modules or enhance the functionality of a module: new tables can be created or columns created/altered. Existing installations which already contain an older version of the `_vt` database need to be upgraded automatically to reach the latest schema.

The first module to use `_vt` was VReplication. We defined an initial schema for the required tables as a list of `create table` statements. `alter`s or `create` ddls were added to it, as the schema was upgraded. There was a single entry point, where the vreplication engine was initialized, where we could run all the statements in order to reach the latest schema.

With time more entry points into vreplication were created. As a result, it was possible that, on existing installations, to end up with queries being executed that expected the new schema before the codepath which updated the schema was traversed. On some paths just the error itself would trigger another code path and "self heal", but more usually we would have irrecoverable errors requiring manual schema updation.

The `WithDDL` module was added by Online DDL to help fix this. WithDDL is initialized the same list of DDLs described above. Any query that targeted the `_vt` database was required to use the `WithDDL.Exec()` function to avail of the automatic schema updation: if there was an error while executing the query, the entire list of DDLs were first run, and the query retried. So if the error was due to the schema not being up to date, we would automagically apply the required DDLs.

Instead of refactoring all of VReplication to use WithDDL, a handful of queries in the entry points into VReplication were modified to use WithDDL. As more Vitess functionality started using `_vt`, more entry points were created. The use of `WithDDL` was not always enforced either due to oversight or due to the contributors not being aware of it. 

[...]

PR #11235, opened 2022-09-15
POC: _vt schema initialization on tablet init using declarative schemas

rohit-nayak-ps:
## Description

This PR is an attempt to use our `schemadiff` library to declare the desired `_vt` schema as of the current Vitess `main`. On tablet init then we use `schemadiff` to calculate the required alters based on the current _vt schema and reach the desired schema.

If this works well we can avoid using `withddl` altogether. 

There is an experimental vttablet flag `--init-vt-schema-on-tablet-init` that has been introduced which can be used to trigger this functionality. For now this POC defaults the flag to `true` so that all tests will run in CI with this feature.

The unit_tests now run. Some tests that specifically depend on the old functionality (either withddl or other schema creation logic) needed to be skipped/modified based on the flag.

There are a few points now which invoke the `sidecardb.Init()`:
* In `mysqlctl.PopulateMetadataTables()` which is a common entry point when a tablet comes up with or without backup/restore.
* In `rpc_replication.InitPrimary()` because there were calls directly to this function which seemed to bypass the previous step: need to investigate this more
* In `healthstreamer.InitSchemaLocked`: this was added only to pass `vttestserver.TestForeignKeysAndDDLModes`. Need to investigate this more

## Related Issue(s)

#10133 

## Checklist


rohit-nayak-ps:
Closing in favor of https://github.com/vitessio/vitess/pull/11520

PR #11252, opened 2022-09-17
unit test: use require and assert

arvind-murty:
## Description

This PR replaces `t.Fatalf` and `t.Errorf` with `require` and `assert`, respectively, in the `mysql` package.

## Checklist

### DIFF
--- a/go/mysql/endtoend/client_test.go
+++ b/go/mysql/endtoend/client_test.go
@@ -243,10 +243,10 @@ func expectNoError(t *testing.T, err error) {
 
 func expectFlag(t *testing.T, msg string, flag, want bool) {
 	t.Helper()
-	if flag != want {
+	require.Equal(t, want, flag, "%s: %v, want: %v",
 		// We cannot continue the test if flag is incorrect.
-		t.Fatalf("%s: %v, want: %v", msg, flag, want)
-	}
+		msg, flag, want)
+
 }
 
 // TestTLS tests our client can connect via SSL.
@@ -256,12 +256,8 @@ func TestTLS(t *testing.T) {
 
 	// First make sure the official 'mysql' client can connect.
 	output, ok := runMysql(t, &params, "status")
-	if !ok {
-		t.Fatalf("'mysql -e status' failed: %v", output)
-	}
-	if !strings.Contains(output, "Cipher in use is") {
-		t.Fatalf("cannot connect via SSL: %v", output)
-	}
+	require.True(t, ok, "'mysql -e status' failed: %v", output)
+	require.True(t, strings.Contains(output, "Cipher in use is"), "cannot connect via SSL: %v", output)
 
 	// Now connect with our client.
 	ctx := context.Background()
@@ -272,9 +268,8 @@ func TestTLS(t *testing.T) {
 	defer conn.Close()
 
 	result, err := conn.ExecuteFetch("SHOW STATUS LIKE 'Ssl_cipher'", 10, true)
-	if err != nil {
-		t.Fatalf("SHOW STATUS LIKE 'Ssl_cipher' failed: %v", err)
-	}
+	require.NoError(t, err, "SHOW STATUS LIKE 'Ssl_cipher' failed: %v", err)
+
 	if len(result.Rows) != 1 || result.Rows[0][0].ToString() != "Ssl_cipher" ||
 		result.Rows[0][1].ToString() == "" {
 		t.Fatalf("SHOW STATUS LIKE 'Ssl_cipher' returned unexpected result: %v", result)
@@ -291,9 +286,8 @@ func TestReplicationStatus(t *testing.T) {
 	defer conn.Close()
 
 	status, err := conn.ShowReplicationStatus()
-	if err != mysql.ErrNotReplica {
-		t.Errorf("Got unexpected result for ShowReplicationStatus: %v %v", status, err)
-	}
+	assert.Equal(t, mysql.ErrNotReplica, err, "Got unexpected result for ShowReplicationStatus: %v %v", status, err)
+
 }
 
 func TestSessionTrackGTIDs(t *testing.T) {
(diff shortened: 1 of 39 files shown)


go/mysql/endtoend/replication_test.go
```diff
@@ -92,28 +90,23 @@ func connectForReplication(t *testing.T, rbr bool) (*mysql.Conn, mysql.BinlogFor
...
 		// Skip rotate packets. These are normal as first packets.
 		if be.IsRotate() {
 			t.Logf("Got a rotate packet: %v", be)
 			continue
 		}
+		require.
+
+			// And we want a FORMAT_DESCRIPTION_EVENT.
+			// Print a few things about the event for sanity checks.
+			True(t, be.IsFormatDescription(), "Unexpected packet: %v", be)
```
mattlord, line 105: another spot

go/mysql/endtoend/query_test.go
```diff
@@ -264,35 +244,25 @@ func doTestWarnings(t *testing.T, disableClientDeprecateEOF bool) {
...
 	// Try a simple insert with a null value
 	result, warnings, err := conn.ExecuteFetchWithWarningCount("insert into a(id) values(10)", 1000, true)
-	if err != nil {
-		t.Fatalf("insert failed: %v", err)
-	}
+	require.NoError(t, err, "insert failed: %v", err)
+
 	if result.RowsAffected != 1 || len(result.Rows) != 0 {
 		t.Errorf("unexpected result for insert: %v", result)
 	}
```
mattlord, line 260: Can't we use this?
```
assert.True(t, (result.RowsAffected == 1 && len(result.Rows) == 0), "unexpected result for insert: %v", result)
```

go/mysql/endtoend/main_test.go
```diff
@@ -41,38 +43,28 @@ var (
...
-	if serr.State != sqlState {
-		t.Fatalf("was expecting SQLError %v / %v / %v but got state %v", code, sqlState, subtext, serr.State)
-	}
+	require.True(t, ok, "was expecting SQLError %v / %v / %v but got: %v", code, sqlState, subtext, err)
+	require.Equal(t, code, serr.Num, "was expecting SQLError %v / %v / %v but got code %v", code, sqlState, subtext, serr.Num)
+	require.Equal(t, sqlState, serr.State, "was expecting SQLError %v / %v / %v but got state %v", code, sqlState, subtext, serr.State)
+
 	if subtext != "" && !strings.Contains(serr.Message, subtext) {
 		t.Fatalf("was expecting SQLError %v / %v / %v but got message %v", code, sqlState, subtext, serr.Message)
 	}
```
mattlord, line 55: Here we can use:
```
require.True(t, (subtext == "" && strings.Contains(serr.Message, subtext)), "was expecting SQLError %v / %v / %v but got message %v", code, sqlState, subtext, serr.Message)
```

go/mysql/endtoend/client_test.go
```diff
@@ -243,10 +243,10 @@ func expectNoError(t *testing.T, err error) {
 
 func expectFlag(t *testing.T, msg string, flag, want bool) {
 	t.Helper()
-	if flag != want {
+	require.Equal(t, want, flag, "%s: %v, want: %v",
 		// We cannot continue the test if flag is incorrect.
-		t.Fatalf("%s: %v, want: %v", msg, flag, want)
-	}
+		msg, flag, want)
+
```
mattlord, line 249: We should move the comment up to its own line IMO.

mattlord (commented):
Thank you for working on this, @arvind-murty ! I only had some minor comments/nits/requests. Once those are resolved in some way I'll do another quick review, approve, and merge. ❤️

go/mysql/endtoend/query_test.go
ajm188: I would vote for separate assertions here, to make it easier to see what actually failed:

```go
assert.Equal(t, 1, result.RowsAffected, "unexpected rows affected by query; result: %v", result)
assert.GreaterThan(t, 0, len(result.Rows), "unexpected row count in query result: %v", result)
```

(also i never remember the ordering for the `GreaterThan` / `LessThan` assertions, so you should double check me on that 😆 )

go/mysql/endtoend/main_test.go
ajm188: same comment about multiple assertions (`require`s, in this case)

go/mysql/endtoend/client_test.go
ajm188: (i'd argue this comment is superfluous; it's what `require` communicates IMO)

go/mysql/endtoend/query_test.go
deepthi: I agree with @ajm188 😄

go/mysql/endtoend/main_test.go
deepthi: @mattlord when you take the negation `&&` will become `||`.

arvind-murty:
I'm not sure why this says there are 208 commits in this PR; I think I messed up the merging somehow. Sorry about all the automatic review requests.

systay:
@arvind-murty Thanks for this!

Yeah, you did something weird when merging. Not sure what happened :)

I cherry-picked your commit on top of main, you can see it here: https://github.com/planetscale/vitess/tree/arvind-unit-tests

LGTM!

go/mysql/endtoend/main_test.go
arvind-murty: @ajm188 I don't think we can use multiple assertions here because it's `||`.

go/mysql/endtoend/query_test.go
arvind-murty: @mattlord I used a tool (https://github.com/arvind-murty/test-rewriter) to edit these files and it currently ignores the more complicated binary expressions (using `&&` or `||`), which is why this wasn't changed. This is also why some of the comments are in weird places.

I manually changed this instance, but I'm currently focused on getting through more packages to get more of the code base cleaned up.

systay (approved):

arvind-murty:
@systay Thanks for approving my changes. Can you merge it?

PR #11254, opened 2022-09-19
Parameterize VTOrc constants

GuptaManan100:
## Description

This PR parameterizes some of the VTOrc constants that make sense to be configurable. Also as part of this PR, the configurations that have `Orchestrator` in their name have been changed to `VTOrc` while maintaining backward compatibility. 

## Checklist

### DIFF
--- a/go/vt/vtorc/config/config.go
+++ b/go/vt/vtorc/config/config.go
@@ -86,20 +86,34 @@ type Configuration struct {
 	DefaultRaftPort                             int      // if a RaftNodes entry does not specify port, use this one
 	RaftNodes                                   []string // Raft nodes to make initial connection with
 	ExpectFailureAnalysisConcensus              bool
-	MySQLOrchestratorHost                       string
-	MySQLOrchestratorMaxPoolConnections         int // The maximum size of the connection pool to the VTOrc backend.
-	MySQLOrchestratorPort                       uint
-	MySQLOrchestratorDatabase                   string
-	MySQLOrchestratorUser                       string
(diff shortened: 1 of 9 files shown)


doc/releasenotes/15_0_0_summary.md
```diff
@@ -248,3 +248,34 @@ send semi-sync ACKs. This ensures that any committed write exists in at least 2
 #### FORMAT=vtexplain
 
 With this new `explain` format, you can get an output that is very similar to the command line `vtexplain` app, but from a running `vtgate`, through a MySQL query.
+
+### VTOrc
+
+#### Configuration
+
+VTOrc configurations that had `Orchestrator` as a substring have been renamed to have `VTOrc` instead. The old configuration won't 
+work in this release. So if backward compatibility is desired, then before upgrading it is suggested to copy the old configurations
```
rsajwani, line 257: I am little confuse here.. for backward compatibility should we do as suggested 

<quotes>So if backward compatibility is desired, then before upgrading it is suggested to copy the old configurations</quotes>

1- first we copy old config to new configurations
2- We then roll new bits
3- remove the old config from configurations

OR

<quotes>VTOrc ignores the configurations that it doesn't understand. So new configurations can be added while running the previous release.</unquotes>

1- first we copy new config to existing configurations
2- We then roll out new bits
3- remove the old config from configurations

GuptaManan100: I have improved the summary doc and added an example too to make it clear.

rsajwani (approved):

PR #11256, opened 2022-09-19
Reapply system settings on connection reconnect

harshit-gangal:
## Description

This PR handles the case when a reconnect happens from vttablet and the old connection has system setting so that the new connection also applies the same settings.

## Related Issue(s)

- #9706 

## Checklist


GuptaManan100 (approved):
Looks good to me! 👍

Issue #11259, opened 2022-09-19
Allow non-SSL callers to VTGate RPC APIs to specify group information for the CallerID


Phanatic:
The current mechanism for passing along a security principal to VTGate's Grpc service, only allows us to send a `Principal` in the context. 

https://github.com/vitessio/vitess/blob/e35a118344e79e17833450c7833aab27c3d4ee52/proto/vtrpc.proto#L34-L51

Which, when used by the GRPC service to authorize queries, will only read the `Principal` from the CallerID structure.
 
https://github.com/vitessio/vitess/blob/e35a118344e79e17833450c7833aab27c3d4ee52/go/vt/vtgate/grpcvtgateservice/server.go#L96-L99

We would like to pass along the `groups` that this caller is part of, in a non-SSL  request and use the groups information to authorize queries in vtgate.

PR #11260, opened 2022-09-19
Allow non-SSL callers of VTGate RPC APIs to specify group information for the CallerID

Phanatic:
## Description

This PR adds a `groups []string` property to the vtrpc.CallerID structure so that non-SSL callers can pass in security group information along with the caller-id. 

It also adds a `grpc-use-effective-groups` CLI flag to vtgate that we can use to control access to this feature.

[Current state of the art for overriding caller information in vtgate](https://vitess.io/docs/13.0/reference/features/transport-security-model/#caller-id-override)
## Related Issue(s)
https://github.com/vitessio/vitess/issues/11259

## Checklist

-    "Backport me!" label has been added if this change should be backported


go/vt/vtgate/grpcvtgateservice/server.go
```diff
@@ -45,10 +45,14 @@ const (
...
 
-var useEffective bool
+var (
+	useEffective       bool
+	useEffectiveGroups bool
+)
 
 func registerFlags(fs *pflag.FlagSet) {
 	fs.BoolVar(&useEffective, "grpc_use_effective_callerid", false, "If set, and SSL is not used, will set the immediate caller id from the effective caller id's principal.")
+	fs.BoolVar(&useEffectiveGroups, "grpc_use_effective_groups", false, "If set, and SSL is not used, will set the immediate caller's security groups from the effective caller id's groups.")
```
harshit-gangal, line 55: the new flag should follow `dashes` over `underscore` format.

harshit-gangal (commented):
looks good, just one comment.

harshit-gangal (approved):

harshit-gangal:
looks like your protobuf version is old.
try doing `make tools` and then `make proto`

PR #11263, opened 2022-09-20
Introduce `servenv` status pages in VTOrc

GuptaManan100:
## Description
This PR introduces `servenv` in VTOrc. This allows us to add `/debug/status` and `/debug/vars` page to VTOrc. The port that they will be available on is added as a flag called `--port`. The tests of VTOrc, don't use this flag and test that VTOrc continues to function the way it used to without these debug pages if `--port` flag is not specified.

The following image shows how the `/debug/status` page looks after these changes. It just displays the recent recoveries that VTOrc has completed.

Screenshot 2022-09-20 at 5 07 42 PM

The metrics that were added previously are also now visible in the `/debug/vars` page - 
Screenshot 2022-09-20 at 4 55 17 PM

## Related Issue(s)

- #6612

## Checklist


go/cmd/vtorc/index.go
```diff
@@ -0,0 +1,31 @@
...
+*/
+
+package main
+
+import (
+	"net/http"
+)
+
+// This is a separate file, so it can be selectively included/excluded from
+// builds to opt in/out of the redirect.
```
ajm188, line 24: does anyone _actually_ leverage this "feature"? deleting files and rebuilding vitess from scratch seems like a kind of heavy-handed way to avoid a redirect, so i'm not actually sure how much we've benefitted from this (i know this is done in other places but i've always been skeptical)

GuptaManan100: Me too. I didn't understand it, so copied it over. Maybe @deepthi knows why we do it this way 🤷‍♂️

go/cmd/vtorc/main.go
```diff
@@ -151,9 +152,9 @@ Please update your scripts before the next version, when this will begin to brea
...
 	}
 
-	_flag.Parse(fs)
+	servenv.ParseFlags("vtorc")
 	// N.B. Also duplicated from go/internal/flag, for the same reason as above.
-	if help || fs.Arg(0) == "help" {
+	if help {
 		pflag.Usage()
 		os.Exit(0)
 	}
```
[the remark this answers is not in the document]
GuptaManan100: I checked removing this shim code. Then `--help` and `-h` work, but `-help` does not. 
In `transformArgsForPflag`, we don't change flags that haven't been registered yet. Like 
```go
if fs.Lookup(name) != nil {
				// Case 1: We have a long opt with this name, so we need to
				// prepend an additional hyphen.
				result = append(result, "-"+arg)
			} else {
				// Case 2: No transformation needed.
				result = append(result, arg)
			}
```
What this means is that we don't change `-help` to `--help` and then pflag complains with unknown shorthand `-e`.
So we need to define the `help` flag before we transform, but if we define the flag ourselves beforehand, then the defer also doesn't work, so we need to catch this too here.

Does it make sense?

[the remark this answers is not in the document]
ajm188: i like your alternate (explicitly checking for the literal `"help"`); since we're not doing this generally but for `vtorc` specifically that seems totally fine to me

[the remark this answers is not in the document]
GuptaManan100: Changed it 👍. Thank-you for the help with the flag changes! ❤️

go/cmd/vtorc/index.go
deepthi: I have no idea :(

rsajwani (approved):
Overall LGTM.
Does this export vtorc metrics to Grafana as well ??

go/cmd/vtorc/index.go
GuptaManan100: I removed that comment since it doesn't make any sense, at least not that any of us can make any.

examples/compose/docker-compose.beginners.yml
```diff
@@ -285,10 +285,12 @@ services:
       - vtctld
       - set_keyspace_durability_policy
     ports:
+      - "13200:8080"
       - "13000:3000"
     volumes:
       - ".:/script"
     environment:
+      - WEB_PORT=8080
```
GuptaManan100, line 293: I have fixed the docker-compose examples to also run the vtorc debug UI. This fail on `main` until the new vitess-lite image is out. I have manually tested that the example works with the new image that I built locally.

go/cmd/vtorc/plugin_prometheusbackend.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	"vitess.io/vitess/go/stats/prometheusbackend"
+	"vitess.io/vitess/go/vt/servenv"
+)
+
+func init() {
+	servenv.OnRun(func() {
+		prometheusbackend.Init("vtorc")
+	})
+}
```
GuptaManan100, line 31: @rsajwani I had to add this file like we do for `vttablet` and other binaries to register the metrics we export on the `debug/vars` page to be available on prometheus too. Thank-you for catching this.

GuptaManan100: I don't think we have to do anything else for `Grafana`. I couldn't find anything that the other binaries were doing, so I think this is it.

GuptaManan100:
I have added release notes for the changes too.

GuptaManan100:
Website changes - https://github.com/vitessio/website/pull/1150

ajm188 (approved):
flag pieces lgtm; i'm not familiar enough with the docker-compose setup

rsajwani (approved):

PR #11269, opened 2022-09-20
Create new api for topo lock shard exists

rsajwani:

## Description

Topo.ShardLock is a blocking call. When concurrent calls to lock same shard happen the client gets block until the lock releases itself or context times out. But this create an issue where a lot of clients try to acquire lock all together and result in server getting overwhelmed. 

This PR introduce TryLock, which is non-blocking at best effort. It tries to find if there is already lock for given shard and returns back immediately instead of blocking incase the lock is already acquired. It is best effort because there is a possibility that we find there is no lock but by the time we try to acquire lock someone else get it before, in this case client will block until the former one releases the lock or current client call times out. In short lock checking and acquiring is not under the same mutex in TryLock.

Currently we have following implementation of Topo Servers

1. Zookeeper
2. Consul
3. Etcd
4. Memory
5. Tee
6. K8s
7. Fake

Out of them Zookeeper, Consul & Etcd have implemented TryLock with non-block implementation (in this PR). Rest of the topo servers have implemented TryLock as blocking call (i-e same as Lock operation).   

Since this is a huge change, in order to reduce the surface area, in this PR we are only changing tablet_discovery.go to use TryLockShard instead of LockShard during recovery. Once we build some confidence that the change is working positively then we can go on and replace LockShard to TryLockShard in other places. 

## Related Issue(s)

#11271 

## Checklist


deepthi:
@rsajwani can you create an issue that describes the symptoms caused by this issue?

go/vt/topo/locks.go
```diff
@@ -392,6 +467,29 @@ func (l *Lock) lockShard(ctx context.Context, ts *Server, keyspace, shard string
...
+	span.Annotate("shard", shard)
+	defer span.Finish()
+
+	shardPath := path.Join(KeyspacesPath, keyspace, ShardsPath, shard)
+	j, err := l.ToJSON()
+	if err != nil {
+		return nil, err
+	}
+	return ts.globalCell.TryLock(ctx, shardPath, j)
+}
```
GuptaManan100, line 491: Same comment about extracting common code with `(l *Lock) lockShard`

go/vt/topo/test/testing.go
```diff
@@ -81,6 +81,11 @@ func TopoServerTestSuite(t *testing.T, factory func() *topo.Server) {
 	checkLock(t, ts)
 	ts.Close()
 
+	t.Log("=== checkLock")
```
GuptaManan100, line 84: `checkTryLock` instead of `checkLock`?

go/vt/topo/test/trylock.go
```diff
@@ -0,0 +1,178 @@
+/*
+Copyright 2019 The Vitess Authors.
```
GuptaManan100, line 2: 2022 in the license.

go/vt/topo/test/trylock.go
```diff
@@ -0,0 +1,178 @@
...
+	t.Log("===      checkLockUnblocks")
+	checkTryLockUnblocks(ctx, t, conn)
+}
+
+func checkTryLockTimeout(ctx context.Context, t *testing.T, conn topo.Conn) {
+	keyspacePath := path.Join(topo.KeyspacesPath, "test_keyspace")
+	lockDescriptor, err := conn.TryLock(ctx, keyspacePath, "")
+	if err != nil {
+		t.Fatalf("TryLock: %v", err)
+	}
```
GuptaManan100, line 58: can we instead use the `require` and `assert` packages.

go/vt/topo/test/trylock.go
```diff
@@ -0,0 +1,178 @@
...
+func checkTryLockMissing(ctx context.Context, t *testing.T, conn topo.Conn) {
+	keyspacePath := path.Join(topo.KeyspacesPath, "test_keyspace_666")
+	if _, err := conn.TryLock(ctx, keyspacePath, "missing"); err == nil {
+		t.Fatalf("TryLock(test_keyspace_666) worked for non-existing keyspace")
+	}
+}
+
+// checkLockUnblocks makes sure that a routine waiting on a lock
+// is unblocked when another routine frees the lock
+func checkTryLockUnblocks(ctx context.Context, t *testing.T, conn topo.Conn) {
```
GuptaManan100, line 130: Shouldn't tryLock not wait at all?

go/vt/topo/test/trylock.go
```diff
@@ -0,0 +1,178 @@
...
+
+	// As soon as we're unblocked, we try to lock the keyspace.
+	go func() {
+		<-unblock
+		waitUntil := time.Now().Add(10 * time.Second)
+		for time.Now().Before(waitUntil) {
+			lockDescriptor, err := conn.TryLock(ctx, keyspacePath, "unblocks")
+			if err != nil {
+				if !topo.IsErrType(err, topo.NodeExists) {
+					t.Errorf("expected nonode exists during trylock")
```
GuptaManan100, line 143: expected node exitsts error

go/vt/topo/zk2topo/lock.go
```diff
@@ -39,7 +43,45 @@ type zkLockDescriptor struct {
...
+	// if there is a folder '/locks' with some entries in it then we can assume that keyspace already have a lock
+	// throw error in this case
+	for _, e := range entries {
+		path := string(e.Key[:])
+		if nodeUnderLockPath.MatchString(path) {
+			return nil, topo.NewError(topo.NodeExists, fmt.Sprintf("lock already exists at path %s", dirPath))
+		}
+
+		// TODO: instead of list should I call listDir and assume /locks directory only exists if there is a lock
+		// TODO: Should we check if all the children under lock is ephemeral
```
GuptaManan100, line 75: Same locks being a file/directory

go/vt/topo/test/trylock.go
rsajwani: fixed

go/vt/topo/consultopo/lock.go
```diff
@@ -49,6 +53,44 @@ func (s *Server) Lock(ctx context.Context, dirPath, contents string) (topo.LockD
...
+	// throw error in this case
+	for _, e := range entries {
+		path := string(e.Key[:])
+		if nodeUnderLockPath.MatchString(path) {
+			return nil, topo.NewError(topo.NodeExists, fmt.Sprintf("lock already exists at path %s", dirPath))
+		}
+
+		// TODO: instead of list should I call listDir and assume /locks directory only exists if there is a lock
+		// TODO: Should we check if all the children under lock is ephemeral
+	}
```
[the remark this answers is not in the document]
rsajwani: fixed it ..

go/vt/topo/etcd2topo/lock.go
```diff
@@ -121,6 +123,39 @@ type etcdLockDescriptor struct {
...
+	// if there is a folder '/locks' with some entries in it then we can assume that keyspace already have a lock
+	// throw error in this case
+	for _, e := range entries {
+		path := string(e.Key[:])
+		if nodeUnderLockPath.MatchString(path) {
+			return nil, topo.NewError(topo.NodeExists, fmt.Sprintf("lock already exists at path %s", dirPath))
+		}
+
+		// TODO: instead of list should I call listDir and assume /locks directory only exists if there is a lock
+		// TODO: Should we check if all the children under lock is ephemeral
```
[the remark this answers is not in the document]
rsajwani: apart from consul every other implementation looks for folder. So we should be good here

go/vt/topo/zk2topo/lock.go
rsajwani: apart from consul every other implementation looks for folder. So we should be good here

GuptaManan100 (commented):
Other than this comment, I think the PR looks good!

go/vt/topo/consultopo/lock.go
```diff
@@ -49,6 +53,40 @@ func (s *Server) Lock(ctx context.Context, dirPath, contents string) (topo.LockD
...
+		// maybe logging the error would work here.
+		return nil, convertError(err, dirPath)
+	}
+
+	// if there is a file 'lock' in it then we can assume that keyspace already have a lock
+	// throw error in this case
+	for _, e := range entries {
+		if nodeUnderLockPath.MatchString(e.Name) && e.Type == topo.TypeFile && e.Ephemeral {
+			return nil, topo.NewError(topo.NodeExists, fmt.Sprintf("lock already exists at path %s", dirPath))
+		}
```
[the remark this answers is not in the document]
GuptaManan100: If we decided to make this `paths.Split` change, we can do so for the other topo-services too.

GuptaManan100 (approved):

go/vt/topo/conn.go
```diff
@@ -113,6 +113,13 @@ type Conn interface {
 	// Returns ErrInterrupted if ctx is canceled.
 	Lock(ctx context.Context, dirPath, contents string) (LockDescriptor, error)
 
+	// TryLock provides exactly same functionality as 'Lock', the only difference is
+	// it tires its best to be non-blocking call. Non-blocking is the best effort though. Depends
+	// on implementation it might result in blocking client (for some scenario) just like any other
```
deepthi, line 118:
```suggestion
	// on the implementation it may result in the client being blocked in some scenarios just like any other
```

go/vt/topo/conn.go
```diff
@@ -113,6 +113,13 @@ type Conn interface {
...
+	// Lock call for concurrent operations. If there is already lock exists for dirPath then TryLock
```
deepthi, line 119:
```suggestion
	// Lock call for concurrent operations. If a lock already exists for dirPath then TryLock
```

go/vt/topo/conn.go
deepthi: We are saying "some scenarios". Do we know what scenarios those are?

go/vt/topo/consultopo/lock.go
```diff
@@ -49,6 +53,40 @@ func (s *Server) Lock(ctx context.Context, dirPath, contents string) (topo.LockD
...
 	}
 
+	return s.lock(ctx, dirPath, contents)
+}
+
+// TryLock is part of the topo.Conn interface.
+// TryLock provides exactly same functionality as 'Lock', the only difference is
+// it tires its best to be non-blocking call. Non-blocking is the best effort though.
+// If there is already lock exists for dirPath then TryLock
+// unlike Lock will return immediately with error 'lock already exists'.
```
deepthi, line 63: Since we have the same comments in the interface, we can delete them here and leave just the first line which says `part of topo.Conn interface`.

go/vt/topo/consultopo/lock.go
```diff
@@ -49,6 +53,40 @@ func (s *Server) Lock(ctx context.Context, dirPath, contents string) (topo.LockD
...
+
+	// if there is a file 'lock' in it then we can assume that keyspace already have a lock
+	// throw error in this case
+	for _, e := range entries {
+		if nodeUnderLockPath.MatchString(e.Name) && e.Type == topo.TypeFile && e.Ephemeral {
+			return nil, topo.NewError(topo.NodeExists, fmt.Sprintf("lock already exists at path %s", dirPath))
+		}
+	}
+
+	// everything is good lets acquire lock.
```
deepthi, line 84:
```suggestion
	// everything is good, let's acquire the lock.
```

go/vt/topo/etcd2topo/lock.go
```diff
@@ -131,6 +133,35 @@ type etcdLockDescriptor struct {
 	leaseID clientv3.LeaseID
 }
 
+// TryLock is part of the topo.Conn interface.
+// TryLock provides exactly same functionality as 'Lock', the only difference is
+// it tires its best to be non-blocking call. Non-blocking is the best effort though.
+// If there is already lock exists for dirPath then TryLock
+// unlike Lock will return immediately with error 'lock already exists'.
```
deepthi, line 140: Delete these 4 lines

go/vt/topo/etcd2topo/lock.go
```diff
@@ -131,6 +133,35 @@ type etcdLockDescriptor struct {
...
+
+	// if there is a folder '/locks' with some entries in it then we can assume that keyspace already have a lock
+	// throw error in this case
+	for _, e := range entries {
+		if nodeUnderLockPath.MatchString(e.Name) && e.Type == topo.TypeDirectory {
+			return nil, topo.NewError(topo.NodeExists, fmt.Sprintf("lock already exists at path %s", dirPath))
+		}
+	}
+
+	// everything is good lets acquire lock.
```
deepthi, line 161:
```suggestion
	// everything is good, let's acquire the lock.
```

go/vt/topo/consultopo/lock.go
```diff
@@ -49,6 +53,40 @@ func (s *Server) Lock(ctx context.Context, dirPath, contents string) (topo.LockD
...
+	if err != nil {
+		// We need to return the right error codes, like
+		// topo.ErrNoNode and topo.ErrInterrupted, and the
+		// easiest way to do this is to return convertError(err).
+		// It may lose some of the context, if this is an issue,
+		// maybe logging the error would work here.
+		return nil, convertError(err, dirPath)
+	}
+
+	// if there is a file 'lock' in it then we can assume that keyspace already have a lock
```
deepthi, line 76:
```suggestion
	// if there is a file 'lock' in it then we can assume that someone else already has a lock
```

go/vt/topo/etcd2topo/lock.go
```diff
@@ -131,6 +133,35 @@ type etcdLockDescriptor struct {
...
+	if err != nil {
+		// We need to return the right error codes, like
+		// topo.ErrNoNode and topo.ErrInterrupted, and the
+		// easiest way to do this is to return convertError(err).
+		// It may lose some of the context, if this is an issue,
+		// maybe logging the error would work here.
+		return nil, convertError(err, dirPath)
+	}
+
+	// if there is a folder '/locks' with some entries in it then we can assume that keyspace already have a lock
```
deepthi, line 153:
```suggestion
	// if there is a folder '/locks' with some entries in it then we can assume that someone else already has a lock
```

go/vt/topo/locks.go
```diff
@@ -301,9 +301,54 @@ func (l *Lock) unlockKeyspace(ctx context.Context, ts *Server, keyspace string,
...
 func (ts *Server) LockShard(ctx context.Context, keyspace, shard, action string) (context.Context, func(*error), error) {
+	return ts.internalLockShard(ctx, keyspace, shard, action, true)
+}
+
+// TryLockShard will lock the shard, and return:
+// - a context with a locksInfo structure for future reference.
+// - an unlock method
+// - an error if anything failed.
+
+// TryLockShard is different than LockShard. It is non-blocking (best effort) which means
```
deepthi, line 322:
```suggestion
// TryLockShard is different from LockShard. It is non-blocking (best effort) which means
```

go/vt/topo/locks.go
```diff
@@ -301,9 +301,54 @@ func (l *Lock) unlockKeyspace(ctx context.Context, ts *Server, keyspace string,
...
+// if there is already a lock at Global-cell level for a given shard then instead
+// of waiting (block) on that shard it returns immediately with error Lock already acquired.
```
deepthi, line 324:
```suggestion
// of waiting (blocking) on that lock it returns immediately with an error.
```

go/vt/topo/locks.go
```diff
@@ -301,9 +301,54 @@ func (l *Lock) unlockKeyspace(ctx context.Context, ts *Server, keyspace string,
...
+// It is the best effort because there is a possibility that we find there is no lock but by the
```
deepthi, line 325:
```suggestion
// It is best-effort because there is a possibility that we find there is no lock but by the
```

go/vt/topo/locks.go
```diff
@@ -301,9 +301,54 @@ func (l *Lock) unlockKeyspace(ctx context.Context, ts *Server, keyspace string,
...
+// time we try to acquire lock someone else get it before, in this case client will block until
```
deepthi, line 326:
```suggestion
// time we try to acquire lock someone else has already acquired it. In this case the client will block until
```

go/vt/topo/locks.go
```diff
@@ -301,9 +301,54 @@ func (l *Lock) unlockKeyspace(ctx context.Context, ts *Server, keyspace string,
...
+// the former one releases the lock or current client call times out.
```
deepthi, line 327:
```suggestion
// the other caller releases the lock or the client call times out.
```

go/vt/topo/test/testing.go
```diff
@@ -40,94 +40,108 @@ func newKeyRange(value string) *topodatapb.KeyRange {
...
+		}
+	}
+	f(t, ts)
+}
+
 // TopoServerTestSuite runs the full topo.Server/Conn test suite.
 // The factory method should return a topo.Server that has a single cell
 // called LocalCellName.
-func TopoServerTestSuite(t *testing.T, factory func() *topo.Server) {
+func TopoServerTestSuite(t *testing.T, factory func() *topo.Server, ignoreList []string) {
```
deepthi, line 55: What's the significance of `ignoreList`? I couldn't see where it is being used.

go/vt/topo/test/trylock.go
```diff
@@ -0,0 +1,182 @@
...
+	"github.com/stretchr/testify/require"
+
+	"vitess.io/vitess/go/vt/topo"
+
+	topodatapb "vitess.io/vitess/go/vt/proto/topodata"
+)
+
+// checkLock checks we can lock / unlock as expected. It's using a keyspace
+// as the lock target.
+func checkTryLock(t *testing.T, ts *topo.Server) {
```
deepthi, line 34: Good tests. Just to be on the safe side (though it shouldn't matter), it might make sense to run some of these at keyspace level and some at shard level.

go/vt/topo/conn.go
rsajwani: change the wordings a bit.

go/vt/topo/test/testing.go
rsajwani: Currently we have following implementation of Topo Servers

1. Zookeeper
2. Consul
3. Etcd
4. Memory
5. Tee
6. K8s
7. Fake

Out of them Zookeeper, Consul & Etcd have implemented TryLock with non-block implementation (in this PR). Rest of the topo servers have implemented TryLock as blocking call (i-e same as Lock operation).  
IgnoreList is used to ignore TryLock test case for topo servers which does not have non-blocking implementation of 'TryLock' yet. e.g go/vt/topo/memorytopo/server_test.go

go/vt/topo/test/trylock.go
rsajwani: The way I am reading the code , we would need to create semantics of TryCreateKeyspace etc. I think we can do that as we move towards these semantics in future.

go/vt/topo/consultopo/lock.go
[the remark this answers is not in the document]
rsajwani: No need to split. ListDir only give element Name. I originally wrote code using List call therefore it uses regular expression. I have changed code to use == instead.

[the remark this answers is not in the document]
GuptaManan100: Awesome!

GuptaManan100 (approved):
LGTM now!

GuptaManan100:
`vttablet_prscomplex` failure is unrelated to this PR.

go/vt/topo/test/shard.go
```diff
@@ -94,3 +96,72 @@ func checkShard(t *testing.T, ts *topo.Server) {
...
+				if isUnLocked1 {
+					close(finished)
+				}
+				break
+			}
+		}
+	}()
+
+	// sleep for a while so we're sure the go routine is blocking
+	time.Sleep(2 * time.Second)
```
shlomi-noach, line 156: A comment I often get in my own PRs, the magic number used here can cause tests to be flaky. Is it possible to wait _until_ the goroutine is certain to be blocking? Wait for some channel/event?

go/vt/topo/conn.go
```diff
@@ -113,6 +113,17 @@ type Conn interface {
 	// Returns ErrInterrupted if ctx is canceled.
 	Lock(ctx context.Context, dirPath, contents string) (LockDescriptor, error)
 
+	// TryLock provides exactly same functionality as 'Lock', the only difference is
+	// it tries its best to be non-blocking. With current implementation for concurrent
```
mattlord, line 117: IMO ~ `non-blocking (best-effort)` language makes no sense as something is non-blocking or it isn't. I think we should instead refer to it as only attempting to acquire the lock if it is likely to succeed. In other places we call it non-blocking w/o any caveats and that is definitely wrong. Can we just remove that wording, since IMO it does not add clarity?

If we need some short-hand way to refer to the difference between Lock and TryLock can we could instead note that the difference is that it attempts to fail-fast?

go/vt/topo/helpers/tee.go
```diff
@@ -187,6 +187,17 @@ type teeTopoLockDescriptor struct {
 
 // Lock is part of the topo.Conn interface.
 func (c *TeeConn) Lock(ctx context.Context, dirPath, contents string) (topo.LockDescriptor, error) {
+	return c.lock(ctx, dirPath, contents)
+}
+
+// TryLock is part of the topo.Conn interface.
+// As of today it provides same functionality as Lock
```
mattlord, line 194: Nit, but we are not consistently using sentence punctuation throughout such as here.

go/vt/topo/stats_conn.go
```diff
@@ -160,6 +160,23 @@ func (st *StatsConn) Lock(ctx context.Context, dirPath, contents string) (LockDe
...
+		return nil, vterrors.Errorf(vtrpc.Code_READ_ONLY, readOnlyErrorStrFormat, statsKey[0], dirPath)
+	}
+	startTime := time.Now()
+	defer topoStatsConnTimings.Record(statsKey, startTime)
+	res, err := st.conn.TryLock(ctx, dirPath, contents)
+	if err != nil {
+		topoStatsConnErrors.Add(statsKey, int64(1))
+		return res, err
+	}
+	return res, err
```
mattlord, line 177: Why not just call Lock() here? This duplication could easily be missed in the future so that the TryLock actually does NOT match Lock.

go/vt/topo/stats_conn_test.go
```diff
@@ -104,6 +104,19 @@ func (st *fakeConn) Lock(ctx context.Context, dirPath, contents string) (lock Lo
...
+// As of today it provides same functionality as Lock
+func (st *fakeConn) TryLock(ctx context.Context, dirPath, contents string) (lock LockDescriptor, err error) {
+	if st.readOnly {
+		return nil, vterrors.Errorf(vtrpc.Code_READ_ONLY, "topo server connection is read-only")
+	}
+	if dirPath == "error" {
+		return lock, fmt.Errorf("dummy error")
+
+	}
+	return lock, err
```
mattlord, line 117: Same comment here about code duplication that can drift.

go/vt/topo/test/shard.go
```diff
@@ -94,3 +96,72 @@ func checkShard(t *testing.T, ts *topo.Server) {
...
+func checkShardWithLock(t *testing.T, ts *topo.Server) {
+	ctx := context.Background()
+	if err := ts.CreateKeyspace(ctx, "test_keyspace", &topodatapb.Keyspace{}); err != nil {
+		t.Fatalf("CreateKeyspace: %v", err)
+	}
+
+	unblock := make(chan struct{})
+	finished := make(chan struct{})
+
+	// Check GetShardNames returns [], nil for existing keyspace with no shards.
```
mattlord, line 110: Nit, but it's a string slice.

go/vt/topo/test/shard.go
```diff
@@ -94,3 +96,72 @@ func checkShard(t *testing.T, ts *topo.Server) {
...
+					require.Fail(t, "expected node exists during trylock", err.Error())
+				}
+				time.Sleep(1 * time.Second)
+				var finalErr error
+				unlock1(&finalErr)
+				isUnLocked1 = true
+				if finalErr != nil {
+					require.Fail(t, "Unlock(test_keyspace) failed", finalErr.Error())
+				}
+			} else {
```
mattlord, line 142: Shouldn't this be considered a test failure as the Try should have failed, right? That's what we're testing.

go/vt/topo/test/shard.go
mattlord: +1. I don't see the point of this sleep. The goroutine will block on writing a message to the non-buffered channel so it will block until we close that channel below. Why sleep? What are we waiting for? Can we check for that?

go/vt/topo/test/shard.go
```diff
@@ -94,3 +96,72 @@ func checkShard(t *testing.T, ts *topo.Server) {
...
+	// sleep for a while so we're sure the go routine is blocking
+	time.Sleep(2 * time.Second)
+
+	// unblock the go routine so it starts waiting
+	close(unblock)
+
+	timeout := time.After(15 * time.Second)
+	select {
+	case <-finished:
+	case <-timeout:
```
mattlord, line 164: The time.After can be declared here.

go/vt/topo/test/shard.go
```diff
@@ -94,3 +96,72 @@ func checkShard(t *testing.T, ts *topo.Server) {
...
+		t.Fatalf("unlocking timed out")
```
mattlord, line 165: Nit, but I don't think this accurately reflects what timed out.

go/vt/topo/test/testing.go
```diff
@@ -40,94 +40,113 @@ func newKeyRange(value string) *topodatapb.KeyRange {
...
```
mattlord, line 55: Comments about what these changes are about would be helpful.

go/vt/topo/test/testing.go
```diff
@@ -40,94 +40,113 @@ func newKeyRange(value string) *topodatapb.KeyRange {
 	return result
 }
 
+func executeTestSuite(f func(*testing.T, *topo.Server), t *testing.T, ts *topo.Server, ignoreList []string, name string) {
+	for _, n := range ignoreList {
+		if n == name {
+			return
```
mattlord, line 46: IMO, it's worth logging that it's being skipped to prevent confusion at higher levels.

go/vt/topo/conn.go
mattlord: Alternatively we could add a short timeout to the lock attempt in try lock so that it is non-blocking.

go/vt/topo/test/trylock.go
```diff
@@ -0,0 +1,182 @@
...
+
+	// test we can interrupt taking the lock
+	interruptCtx, cancel := context.WithCancel(ctx)
+	go func() {
+		time.Sleep(1 * time.Second)
+		cancel()
+	}()
+	if _, err := conn.TryLock(interruptCtx, keyspacePath, "interrupted"); !topo.IsErrType(err, topo.NodeExists) {
+		require.Fail(t, "TryLock failed", err.Error())
+	}
```
mattlord, line 107: How does this test the cancellation? Shouldn't we be checking that the context being done is what caused the failure to lock?

go/vt/topo/test/trylock.go
```diff
@@ -0,0 +1,182 @@
...
+
+	// test we can't unlock again
+	if err := lockDescriptor.Unlock(ctx); err == nil {
+		require.Fail(t, "Unlock failed", err.Error())
+	}
+}
+
+// checkTryLockMissing makes sure we can't lock a non-existing directory.
+func checkTryLockMissing(ctx context.Context, t *testing.T, conn topo.Conn) {
+	keyspacePath := path.Join(topo.KeyspacesPath, "test_keyspace_666")
```
mattlord, line 125: 😈

go/vt/topo/test/trylock.go
```diff
@@ -0,0 +1,182 @@
...
+	lockDescriptor2, err := conn.TryLock(ctx, keyspacePath, "")
+	if err != nil {
+		require.Fail(t, "Lock(test_keyspace) failed", err.Error())
+	}
+
+	// unblock the go routine so it starts waiting
+	close(unblock)
+
+	// sleep for a while so we're sure the go routine is blocking
+	time.Sleep(5 * time.Second)
```
mattlord, line 170: Same comment here. I don't see the point of this. What are we actually waiting for?

go/vt/topo/stats_conn.go
rsajwani: good point. I have changed it .

go/vt/topo/conn.go
rsajwani: I have changed the description as per your comments. Please review it and let me know if it is more readable now.

go/vt/topo/test/shard.go
rsajwani: thats what we are trying to do here. We are doing TryLockShard and expect NodeExists error until it prior lock gets unlock and then we expect to acquire lock in else case.

go/vt/topo/test/shard.go
rsajwani: I have restructured the test

go/vt/topo/test/shard.go
mattlord: Unless I'm misreading things, the test would currently pass if:
1.  TryLock fails with a NodeExists error
2. TryLock succeeds 

And it only fails if TryLock fails with a different error. So what are we trying to test and confirm here? I thought the test should only pass if 1 is true, no?

go/vt/topo/locks.go
```diff
@@ -383,6 +419,17 @@ func CheckShardLocked(ctx context.Context, keyspace, shard string) error {
 // lockShard will lock the shard in the topology server.
 // UnlockShard should be called if this returns no error.
 func (l *Lock) lockShard(ctx context.Context, ts *Server, keyspace, shard string) (LockDescriptor, error) {
+	return l.internalLockShard(ctx, ts, keyspace, shard, true)
+}
+
+// tryLockShard will lock the shard in the topology server.
+// It is non-blocking call for blocking call lockshard instead.
```
mattlord, line 426: Another spot where I think the description is misleading.

rsajwani: If everything else looks good then I can go ahead and fix this last change. Just wanted to save CI cycles :) ..

go/vt/topo/test/shard.go
mattlord: We walked through this together and it now makes sense after the refactoring. We'll be adding comments to make it even more obvious to the reader.

mattlord (approved):
Nice work, @rsajwani !

Issue #10740, opened 2022-07-19
Switch flag definitions to be on pflag instead of flag in `package go/mysql`


ajm188:
Part of https://github.com/vitessio/vitess/issues/10697.

Current flags:

```
$ git grep -E "\bflag\.[A-Z]" -- go/mysql/*.go
go/mysql/auth_server_clientcert.go:var clientcertAuthMethod = flag.String("mysql_clientcert_auth_method", string(MysqlClearPassword), "client-side authentication method to use. Supported values: mysql_clear_password, dialog.")
go/mysql/auth_server_clientcert.go:     if flag.CommandLine.Lookup("mysql_server_ssl_ca").Value.String() == "" {
go/mysql/auth_server_static.go: mysqlAuthServerStaticFile           = flag.String("mysql_auth_server_static_file", "", "JSON File to read the users/passwords from.")
go/mysql/auth_server_static.go: mysqlAuthServerStaticString         = flag.String("mysql_auth_server_static_string", "", "JSON representation of the users/passwords config.")
go/mysql/auth_server_static.go: mysqlAuthServerStaticReloadInterval = flag.Duration("mysql_auth_static_reload_interval", 0, "Ticker to reload credentials")
go/mysql/conn.go:var mysqlServerFlushDelay = flag.Duration("mysql_server_flush_delay", 100*time.Millisecond, "Delay after which buffered response will be flushed to the client.")
go/mysql/query_benchmark_test.go:       flag.IntVar(&testReadConnBufferSize, "test.read_conn_buffer_size", connBufferSize, "buffer size for reads from connections in tests")
```

PR #10752, opened 2022-07-19
[go/mysql/*] Move all authserver–related flags off of global flagset

ajm188:
## Description

(This builds on top of #10619, and should be rebased after that is merged.)

Anywayyyyy, this properly scopes the `mysql_auth_*` flags to just the `vtgate` binary, which is the only place where the functions that register these flags actually get called. Consequently, the `vttablet.txt` test data was also updated to reflect these changes.

## Related Issue(s)

- Partial fix for #10740 (I didn't address the `--mysql_server_flush_delay` flag in `conn.go`, which I _think_ is query serving but not confident enough to include in this change).
- Closes #10742 
- Closes #10743 
- Partial fix for #4932 

## Checklist

-   [x] "Backport me!" label has been added if this change should be backported
-   [x] Tests were added or are not required
-   [x] Documentation was added or is not required

PR #4196, opened 2018-09-14
[go/mysql] add ability to create server connection with unbuffered reads

LK4D4:
Old functions NewFromListener and NewListener preserve their behavior
(use buffered reader with connBufferSize buffer).
New function NewListenerWithConfig isn't used in vitess itself for now,
but it can be used to control buffer size (or to disable buffering).
/cc @sougou @danieltahara

PR #11272, opened 2022-09-21
Move go/mysql flags to pflags

mattlord:
## Description

This moves the `go/mysql` package flags to pflags and limits their usage to vtgate. 

## Related Issue(s)

  - Fixes: https://github.com/vitessio/vitess/issues/10740
  - Follow-up to: https://github.com/vitessio/vitess/pull/10752

## Checklist


go/mysql/query_benchmark_test.go
```diff
@@ -29,6 +29,7 @@ import (
 var testReadConnBufferSize = connBufferSize
 
 func init() {
+	// Don't see a clear way to use pflag for this go test -benchmem flag
 	flag.IntVar(&testReadConnBufferSize, "test.read_conn_buffer_size", connBufferSize, "buffer size for reads from connections in tests")
```
GuptaManan100, line 33: I don't see this flag being used on `arewefastyet` or anywhere else in Vitess. It was introduced in https://github.com/vitessio/vitess/pull/4196.
I think we can just switch to the `pflag` package and not do anything else i.e. just change `flag.IntVar` to `pflag.IntVar`.
This flag doesn't show up anywhere in any of the binaries. Only used when running the benchmarks in this file.

GuptaManan100: WDYT @mattlord?

GuptaManan100 (approved):
Once we address ☝️, everything else LGTM!

go/mysql/auth_server_static.go
```diff
@@ -41,13 +41,15 @@ var (
 	mysqlAuthServerStaticFile           string
 	mysqlAuthServerStaticString         string
 	mysqlAuthServerStaticReloadInterval time.Duration
+	mysqlServerFlushDelay               time.Duration
```
ajm188, line 44: this needs the default value here:

```go
mysqlServerFlushDelay = 100 * time.Millisecond


....

OnParseFor("...", func(...) {
    fs.DurationVar(&mysqlServerFlushDelay, "mysql_server_flush_delay", mysqlServerFlushDelay, "<usage text goes here>")
}
```

ajm188 (commented):
one other change that needs to go in (in addition to @GuptaManan100's comment), then we should be good

thanks for doing this!

go/mysql/auth_server_static.go
mattlord: I've changed this. But it's not really required is it? It's not clear to me why we would handle this particular flag value differently — knowing this will help me with the others.

ajm188: it's required for any package that imports (uses the default value) without exposing the flag, which does happen from time to time

GuptaManan100 (approved):
Looks good to me!

PR #11306, opened 2022-09-21
feat: rewrite column names in HAVING

systay:
## Description
When trying to figure out the offsets for some HAVING and ORDER BY expression, it's much easier for the planner if we have the real aggregation function and not the column alias introduced. This PR rewrites these queries, so later phases of the planner don't have to take this possibility into account.

Example:

```sql
select id, sum(foo) as sumOfFoo from tbl group by id having sumOfFoo > 10
```

will get rewritten to:

```sql
select id, sum(foo) as sumOfFoo from tbl group by id having sum(foo) > 10
```


vmg (approved):

frouioui (approved):
LGTM, we have several tests in `aggregation_test.go` for this.

PR #11309, opened 2022-09-21
Fix query list override issue on mysql restart

harshit-gangal:
## Description

This PR fixes the issue with the query list getting overridden when a MySQL restarts and the same connection ID is assigned to a different connection.

## Checklist


deepthi (approved):
💯 
A few more comments and then this should be good to merge.

deepthi:
We are in the in-between state right now where release-15.0 branch has been created but is not yet open for back ports (until rc1 is released). When that opens up, we'll need to manually port this fix for release 15.

Issue #11275, opened 2022-09-21
Bug Report: `LEFT JOIN` rewritten incorrectly, leading to incorrect behaviour


johanstenberg92:
### Overview of the Issue

Hi, under certain conditions Vitess rewrites a left join to a query with vastly different meanings than the original query.  Essentially Vitess lifts conditions from the `where` part of the query into the join condition, however this can't be done on left joins - since they return both matched and un-matched elements.

### Reproduction Steps

1. Deploy the following `vschema`:
```
{
  "mainkeyspace": {
    "sharded": true,
    "vindexes": {
      "hash": {
        "type": "hash"
      },
      "md5": {
        "type": "unicode_loose_md5",
        "params": {},
        "owner": ""
      }
    },
    "tables": {
      "users": {
        "column_vindexes": [
          {
            "column": "id",
            "name": "hash"
          }
        ],
        "columns": [
          {
            "name": "id",
            "type": "UINT64"
          },
          {
            "name": "name",
            "type": "VARCHAR"
          },
          {
            "name": "title",
            "type": "VARCHAR"
          }
        ],
        "auto_increment": null
      },
      "purchases": {
        "column_vindexes": [
          {
            "column": "user_id",
            "name": "hash"
          }
        ],
        "columns": [
          {
            "name": "id",
            "type": "UINT64"
          },
          {
            "name": "user_id",
            "type": "UINT64"
          },
          {
            "name": "product_name",
            "type": "VARCHAR"
          }
        ],
        "auto_increment": null
      }
    }
  }
}
```

2. Deploy the following `schema`:
```
CREATE TABLE users(
  id bigint,
  name VARCHAR(250) not null,
  title VARCHAR(250) not null,
  primary key(id)
);

CREATE TABLE purchases(
  id bigint,
  user_id bigint not null,
  product_name VARCHAR(250),
  primary key(id)
);
```

3. Input query:
```
SELECT users.name FROM users LEFT JOIN purchases on users.id = purchases.user_id WHERE users.title = "mr" and ((purchases.id > 12 AND users.name = "marc") or purchases.product_name = 'ABC') order by users.id;
```

4. Run:
```
vtexplain --vschema-file=vschema-testing.json --schema-file vschema-sql-file.sql --planner-version gen4 --output-mode json --sql-file vitess-bug-in.sql
```

5. Output:
```
[
    {
[...]

PR #11310, opened 2022-09-21
Fix conditions after `<=>` operator on left joined table columns being ignored for routing purposes.

arthurschreiber:
## Description

This fixes an edge case where a comparison using the `<=>` operator would prevent the use of any following predicates after for routing purposes (only on left joins).

Here's an example query:

```sql
SELECT music.id FROM music LEFT OUTER JOIN user ON music.user_id = user.id WHERE user.id <=> NULL AND music.user_id = 10
```

This would generate the following plan:

```json
{
  "QueryType": "SELECT",
  "Original": "SELECT music.id FROM music LEFT OUTER JOIN user ON music.user_id = user.id WHERE user.id \u003c=\u003e NULL AND music.user_id = 10",
  "Instructions": {
    "OperatorType": "Route",
    "Variant": "Scatter",
    "Keyspace": {
      "Name": "user",
      "Sharded": true
    },
    "FieldQuery": "select music.id from music left join `user` on music.user_id = `user`.id where 1 != 1",
    "Query": "select music.id from music left join `user` on music.user_id = `user`.id where `user`.id \u003c=\u003e null and music.user_id = 10",
    "Table": "`user`, music"
  },
  "TablesUsed": [
    "user.music",
    "user.user"
  ]
}
```

With this fix in place, this plan is generated instead:

```json
{
  "QueryType": "SELECT",
  "Original": "SELECT music.id FROM music LEFT OUTER JOIN user ON music.user_id = user.id WHERE user.id \u003c=\u003e NULL AND music.user_id = 10",
  "Instructions": {
    "OperatorType": "Route",
    "Variant": "EqualUnique",
    "Keyspace": {
      "Name": "user",
      "Sharded": true
    },
    "FieldQuery": "select music.id from music left join `user` on music.user_id = `user`.id where 1 != 1",
    "Query": "select music.id from music left join `user` on music.user_id = `user`.id where music.user_id = 10 and `user`.id \u003c=\u003e null",
    "Table": "`user`, music",
    "Values": [
      "INT64(10)"
    ],
    "Vindex": "user_index"
  },
  "TablesUsed": [
    "user.music",
    "user.user"
  ]
}
```

## Related Issue(s)

I stumbled upon this issue while working on a fix for https://github.com/vitessio/vitess/issues/11275.

## Checklist

### DIFF
--- a/go/vt/vtgate/planbuilder/abstract/filter.go
+++ b/go/vt/vtgate/planbuilder/abstract/filter.go
@@ -48,8 +48,20 @@ func (f *Filter) CheckValid() error {
 
 // PushPredicate implements the LogicalOperator interface
 func (f *Filter) PushPredicate(expr sqlparser.Expr, semTable *semantics.SemTable) (LogicalOperator, error) {
-	f.Predicates = append(f.Predicates, expr)
-	return f, nil
+	op, err := f.Source.PushPredicate(expr, semTable)
+	if err != nil {
+		return nil, err
+	}
+
+	if filter, isFilter := op.(*Filter); isFilter {
+		filter.Predicates = append(f.Predicates, filter.Predicates...)
+		return filter, err
+	}
+
+	return &Filter{
+		Source:     op,
+		Predicates: f.Predicates,
+	}, nil
 }
 
 // Compact implements the LogicalOperator interface
--- a/go/vt/vtgate/planbuilder/testdata/filter_cases.txt
+++ b/go/vt/vtgate/planbuilder/testdata/filter_cases.txt
@@ -5938,6 +5938,33 @@ Gen4 plan same as above
 }
 Gen4 plan same as above
 
+# conditions following a null safe comparison operator can be used for routing
+"SELECT music.id FROM music LEFT OUTER JOIN user ON music.user_id = user.id WHERE user.id <=> NULL AND music.user_id = 10"
+{
+  "QueryType": "SELECT",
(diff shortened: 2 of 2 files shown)


arthurschreiber:
The root of this issue is that when pushing predicates to a `Join` operator, we handle nullsafe equality comparisons (and other more complex predicates) by putting them into a `Filter` operator that wraps the `Join`. From that point on, all further predicates get pushed to the `Filter`, thus never reaching the `Join` operator and never being used for routing purposes.

I fixed this by changing the logic for when predicates are pushed into a `Filter` - we first attempt to push the predicate down the operator that is wrapped by the filter. If that returns a new non-`Filter` predicate, we wrap it again into a `Filter`. But if pushing the predicate down returns a `Filter` as well, we merge the filters together.

go/vt/vtgate/planbuilder/abstract/filter.go
```diff
@@ -48,8 +48,20 @@ func (f *Filter) CheckValid() error {
...
+
+	if filter, isFilter := op.(*Filter); isFilter {
+		filter.Predicates = append(f.Predicates, filter.Predicates...)
+		return filter, err
+	}
+
+	return &Filter{
+		Source:     op,
+		Predicates: f.Predicates,
+	}, nil
```
arthurschreiber, line 64: I guess this could also re-use `f` here? But I wasn't sure about the supposed semantics around `PushPredicate` and whether structs returned here are supposed to be modified / reused.

```suggestion
        f.Source = op
        return f, nil
```

frouioui: I think returning `f` does not make a strong difference.

frouioui (approved):
That looks good to me @arthurschreiber! :)

PR #11313, opened 2022-09-21
optionally disable verify-after-insert behavior of lookup vindexes

maxenglander:
## Description

Add a VIndex option to skip verify-after-insert behavior when using `INSERT...IGNORE` or `INSERT...ON DUPLICATE KEY UPDATE`.

### Context

The default behavior of a lookup VIndex when inserting with `IGNORE` or `ON DUPLICATE KEY` into a table with a Lookup VIndex is to:

 1. Generate an `INSERT` statement for the table that has the Lookup VIndex.
    * In the process, obtain the keyspace IDs that the Primary VIndex key IDs map to.
 1. Use the obtained keyspace IDs to update the Lookup Vindex.
    * Insert Lookup VIndex `from` keys into the lookup table, mapping them to those keyspace IDs.
    * Ignore any duplicate key errors.
 1. Verify (with a `SELECT` statement) that the Lookup VIndex `from` values map to the keyspace IDs used in step 2.
    * If not, remove unmapped keyspace IDs from the original list of keyspace IDs.
 1. Produce a set of destinations from the final list of keyspace IDs produced in step 3.
 1. Rewrite `INSERT` statements for each destination.

The purpose of verify-after-insert step (step 3) is find out which keyspace IDs are not mapped to by any Lookup VIndex `from` value and, for any unmapped keyspace ID, exclude that keyspace ID from steps 4 and 5.

### Use Case(s)

This verify-after-insert step can have a substantial impact on performance. There are Vitess users using lookup indexes in which:

 1. The `from` value is never updated by the application, and therefore...
 1. The `Verify` step will always succeed, even when there are duplicate keys, and therefore...
 1. We can safely skip it.

## Checklist

-   [x] Documentation was added or is not required: https://github.com/vitessio/website/pull/1185


harshit-gangal:
I think this would need website documentation.

maxenglander:
@harshit-gangal
 *  https://github.com/vitessio/website/pull/1185

harshit-gangal:
can you merge `main` and push once, so that pending CI checks can run?

harshit-gangal (approved):

maxenglander:
@harshit-gangal you bet, done

PR #11317, opened 2022-09-22
Fix complex predicates being pulled into `ON` conditions for `LEFT JOIN` statements.

arthurschreiber:
## Description

Complex `WHERE` predicates for joined tables (stuff along the lines of `(lhs.a = ... OR rhs.b = ...)` or even just `lhs.a = rhs.b`) are pulled into the join conditions. This makes sense for `INNER JOIN` statements, but for `LEFT JOIN` statements this will cause the result to be totally wrong.

This PR fixes this issue by performing this movement of predicates only for non-`LEFT JOIN`s. On top of that, we now also detect more cases where `LEFT JOIN`s can be transformed to `INNER JOIN`s due to "null intolerant" comparisons (e.g. `lhs.a = rhs.b` in the `WHERE` condition allows rewriting a left join between `lhs` and `rhs` to an inner join).

```sql
SELECT music.id FROM music LEFT OUTER JOIN user ON music.user_id = user.id WHERE (user.name = 'Trent Reznor' OR music.genre = 'pop') AND music.user_id = 5
```

had the following plan generated:

```json
{
  "QueryType": "SELECT",
  "Original": "SELECT music.id FROM music LEFT OUTER JOIN user ON music.user_id = user.id WHERE (user.name = 'Trent Reznor' OR music.genre = 'pop') AND music.user_id = 5",
  "Instructions": {
    "OperatorType": "Route",
    "Variant": "EqualUnique",
    "Keyspace": {
      "Name": "user",
      "Sharded": true
    },
    "FieldQuery": "select music.id from music left join `user` on music.user_id = `user`.id and (`user`.`name` = 'Trent Reznor' or music.genre = 'pop') where 1 != 1",
    "Query": "select music.id from music left join `user` on music.user_id = `user`.id and (`user`.`name` = 'Trent Reznor' or music.genre = 'pop') where music.user_id = 5",
    "Table": "`user`, music",
    "Values": [
      "INT64(5)"
    ],
    "Vindex": "user_index"
  },
  "TablesUsed": [
    "user.music",
    "user.user"
  ]
}
```

the executed query became:

```sql
select music.id from music left join `user` on music.user_id = `user`.id and (`user`.`name` = 'Trent Reznor' or music.genre = 'pop') where music.user_id = 5
```

Which has a very different meaning from the original query.

The new plan looks like this:

```json
{
  "QueryType": "SELECT",
  "Original": "SELECT music.id FROM music LEFT OUTER JOIN user ON music.user_id = user.id WHERE (user.name = 'Trent Reznor' OR music.genre = 'pop') AND music.user_id = 5",
  "Instructions": {
    "OperatorType": "Route",
    "Variant": "EqualUnique",
    "Keyspace": {
      "Name": "user",
      "Sharded": true
    },
    "FieldQuery": "select music.id from music left join `user` on music.user_id = `user`.id where 1 != 1",
    "Query": "select music.id from music left join `user` on music.user_id = `user`.id where music.user_id = 5 and (`user`.`name` = 'Trent Reznor' or music.genre = 'pop')",
    "Table": "`user`, music",
    "Values": [
      "INT64(5)"
    ],
    "Vindex": "user_index"
  },
  "TablesUsed": [
    "user.music",
    "user.user"
  ]
}
```

And the predicates are kept in the `WHERE` part of the query.

## Related Issue(s)

Fixes https://github.com/vitessio/vitess/issues/11275

## Checklist

### DIFF
--- a/go/vt/vtgate/planbuilder/abstract/join.go
+++ b/go/vt/vtgate/planbuilder/abstract/join.go
@@ -46,22 +46,10 @@ func (j *Join) PushPredicate(expr sqlparser.Expr, semTable *semantics.SemTable) 
 		}
 		j.LHS = lhs
 		return j, nil
-	case deps.IsSolvedBy(j.RHS.TableID()):
-		// we are looking for predicates like `tbl.col = <>` or `<> = tbl.col`,
-		// where tbl is on the rhs of the left outer join
-		if cmp, isCmp := expr.(*sqlparser.ComparisonExpr); isCmp && cmp.Operator != sqlparser.NullSafeEqualOp &&
-			(sqlparser.IsColName(cmp.Left) && semTable.RecursiveDeps(cmp.Left).IsSolvedBy(j.RHS.TableID()) ||
-				sqlparser.IsColName(cmp.Right) && semTable.RecursiveDeps(cmp.Right).IsSolvedBy(j.RHS.TableID())) {
-			// When the predicate we are pushing is using information from an outer table, we can
-			// check whether the predicate is "null-intolerant" or not. Null-intolerant in this context means that
-			// the predicate will not return true if the table columns are null.
-			// Since an outer join is an inner join with the addition of all the rows from the left-hand side that
-			// matched no rows on the right-hand, if we are later going to remove all the rows where the right-hand
-			// side did not match, we might as well turn the join into an inner join.
 
-			// This is based on the paper "Canonical Abstraction for Outerjoin Optimization" by J Rao et al
-			j.LeftJoin = false
-		}
+	case deps.IsSolvedBy(j.RHS.TableID()):
+		j.tryConvertToInnerJoin(expr, semTable)
+
 		if !j.LeftJoin {
 			rhs, err := j.RHS.PushPredicate(expr, semTable)
 			if err != nil {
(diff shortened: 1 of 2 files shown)


go/vt/vtgate/planbuilder/abstract/join.go
```diff
@@ -46,22 +46,26 @@ func (j *Join) PushPredicate(expr sqlparser.Expr, semTable *semantics.SemTable)
...
-			// When the predicate we are pushing is using information from an outer table, we can
-			// check whether the predicate is "null-intolerant" or not. Null-intolerant in this context means that
-			// the predicate will not return true if the table columns are null.
-			// Since an outer join is an inner join with the addition of all the rows from the left-hand side that
-			// matched no rows on the right-hand, if we are later going to remove all the rows where the right-hand
-			// side did not match, we might as well turn the join into an inner join.
-
-			// This is based on the paper "Canonical Abstraction for Outerjoin Optimization" by J Rao et al
-			j.LeftJoin = false
+		if j.LeftJoin {
```
arthurschreiber, line 51: I wrapped this in a condition as the optimization check that is happening here only makes sense if the join is a left join.

go/vt/vtgate/planbuilder/abstract/join.go
```diff
@@ -70,19 +58,61 @@ func (j *Join) PushPredicate(expr sqlparser.Expr, semTable *semantics.SemTable)
...
+// When a predicate uses information from an outer table, we can convert from an outer join to an inner join
+// if the predicate is "null-intolerant".
+//
+// Null-intolerant in this context means that the predicate will not be true if the table columns are null.
+//
+// Since an outer join is an inner join with the addition of all the rows from the left-hand side that
+// matched no rows on the right-hand, if we are later going to remove all the rows where the right-hand
+// side did not match, we might as well turn the join into an inner join.
+//
+// This is based on the paper "Canonical Abstraction for Outerjoin Optimization" by J Rao et al
```
arthurschreiber, line 95: MySQL actually also documents this optimization in much greater detail here: https://dev.mysql.com/doc/refman/5.7/en/outer-join-simplification.html

On that page, there's additional cases being described that we're not implementing yet, with detailed examples. I think a "simple" case we're currently missing is also performing the optimization for `rhs.a IS NOT NULL`. I'll push a change for that in a moment.

But it might make sense to check out the other examples on that page and see if we can implement them as well? 🤔 

/cc @systay

systay: Nice find! Yeah, we should check which of these situations we could find and be clever around

systay (approved):
🌹 Beautiful 🌷

deepthi:
The labels look correct to me. Not worth asking for a force-push for the stuck "Check Pull Request labels" check. Merging based on @systay's approval.

PR #11322, opened 2022-09-22
Fix adding flags to vtctlclient and vtctldclient

GuptaManan100:
## Description

It was noticed that the flag parsing in `vtctlclient` and `vtctldclient` does not go through `servenv.ParseFlagsFor`. So all the places where we use `servenv.OnParseFor`, for these binaries don't end up actually adding the flags for them. Instead we have to add them directly on their `main.go` files.

This PR fixes this issue and adds test for their flags.

## Related Issue(s)

- #10697 

## Checklist


go/cmd/vtctlclient/main.go
```diff
@@ -73,6 +76,9 @@ func main() {
 	defer exit.Recover()
 
 	fs := pflag.NewFlagSet("vtctlclient", pflag.ExitOnError)
+	grpcclient.RegisterFlags(fs)
+	grpccommon.RegisterFlags(fs)
+	trace.RegisterFlags(fs)
```
ajm188, line 81: This should be removed from both

GuptaManan100: Fixed.

go/cmd/vtctlclient/main.go
```diff
@@ -27,6 +27,9 @@ import (
 
 	"github.com/spf13/pflag"
 
+	"vitess.io/vitess/go/vt/grpcclient"
+	"vitess.io/vitess/go/vt/grpccommon"
+
```
ajm188, line 32: Why are these in a separate group?

go/flags/endtoend/flags_test.go
```diff
@@ -67,6 +75,7 @@ func TestHelpOutput(t *testing.T) {
 			cmd := exec.Command(binary, args...)
 			output := bytes.Buffer{}
 			cmd.Stderr = &output
+			cmd.Stdout = &output
```
ajm188, line 78: Why did we need to add this?

GuptaManan100: `vtctldclient` prints the help on Stdout and not on Stderr. Weird.

ajm188: Then I think we should fix that?

go/cmd/vtctlclient/main.go
ajm188: Can you fix it anyway?

GuptaManan100: Fixed it

go/flags/endtoend/flags_test.go
GuptaManan100: I don't know how to. I checked and the help is coming from `cobra`. Is there some way to redirect it from Stdout to Stderr?

ajm188 (approved):

ajm188 (approved):

PR #11323, opened 2022-09-22
make MySQL56-flavor schema queries forward-compatible

### LINKED ISSUE #11312: -db_flavor FilePos doesn't work well out-of-the-box with external MySQL 8.0 tablets
@aquarapid reported this to me.

# Description

Starting a tablet on top of:

 * An external MySQL 8.0 data source
 * With `-db_flavor=FilePos`

The tablet does not reach ready state, and logs messages similar to these:

```
E0921 05:57:46.351677       1 tm_state.go:331] Cannot start query service: Expression #2 of SELECT list is not in GROUP BY clause and contains nonaggregated column 'information_schema.tables.TABLE_TYPE' which is not functionally dependent on columns in GROUP BY clause; this is incompatible with sql_mode=only_full_group_by (errno 1055) (sqlstate 42000) during query: SELECT table_name, table_type, unix_timestamp(create_time), table_comment, SUM( data_length + index_length), SUM( data_length + index_length)
```

Note that the tablet is trying to execute MySQL queries that the external MySQL doesn't support.

# Details

The issue seems to be that the `filePosFlavor` is [hard-coded](https://github.com/vitessio/vitess/blob/0b6c64442/go/mysql/flavor_filepos.go#L329-L332) to use a MySQL 5.6 flavor information schema query:

```
// baseShowTablesWithSizes is part of the Flavor interface.
func (*filePosFlavor) baseShowTablesWithSizes() string {
	return TablesWithSize56
}
```

Removing `only_full_group_by` from the SQL mode fixes the issue and allows the tablet to come online, but we think it would be better if we didn't have to do this.

maxenglander:
## Description

The various SQL flavors (56, 57, 80, MariaDB, FilePos) use different queries for schema inspection. The FilePos flavor is using a 56 flavor query which isn't forward compatible with MySQL 80.
## Related Issue(s)

This commit updates the 56 flavor to be forward-compatible with 57, 80. s/o @mattlord for the suggested fix.

Fixes #11312.

## Checklist

### DIFF
--- a/go/mysql/endtoend/client_test.go
+++ b/go/mysql/endtoend/client_test.go
@@ -377,3 +377,32 @@ func TestClientInfo(t *testing.T) {
 		t.Fatalf("expected result.Info=%q, got=%q", infoPrepared, result.Info)
 	}
 }
+
+func TestBaseShowTables(t *testing.T) {
+	params := connParams
+	ctx := context.Background()
+	conn, err := mysql.Connect(ctx, &params)
+	require.NoError(t, err)
+	defer conn.Close()
+
+	sql := conn.BaseShowTables()
+	// An improved test would make assertions about the results. This test just
+	// makes sure there aren't any errors.
+	_, err = conn.ExecuteFetch(sql, -1, true)
+	require.NoError(t, err)
+}
+
+func TestBaseShowTablesFilePos(t *testing.T) {
+	params := connParams
+	params.Flavor = "FilePos"
+	ctx := context.Background()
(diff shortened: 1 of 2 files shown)


mattlord (approved):
Thanks! ❤️ I'll let Query Serving review too and they can merge.

dbussink (approved):
Thnx for the fix!

deepthi (approved):
Technically query-serving but we don't need to wait for their review.
The approvals we have so far are sufficient.

PR #11330, opened 2022-09-22
vtbackup: disable redo log before starting replication

maxenglander:
## Description

According to MySQL docs:

> As of MySQL 8.0.21, you can disable redo logging using the ALTER
> INSTANCE DISABLE INNODB REDO_LOG statement. This functionality is
> intended for loading data into a new MySQL instance. Disabling redo
> logging speeds up data loading by avoiding redo log writes and
> doublewrite buffering.

See: https://dev.mysql.com/doc/refman/8.0/en/innodb-redo-log.html#innodb-disable-redo-logging

We can take advantage of this in vtbackup. This change disables the redo log on MySQL >= 8.0.21 before starting replication, and re-enables the redo log after stopping replication.

## Details

Backups happen in phases:

 1. `vtbackup` downloads the most recent backup.
 2. Then restores the backup on disk.
 3. Then retrieves and applies the binlog from the MySQL primary.
 4. Then takes and uploads a new backup.

The goal of this PR is to make step 3 faster. The redo log is meant to aid in crash-recovery. Writing to the redo log slows things down a bit. By disabling the redo log, we can get a performance boost. We can afford to lose the ability to recover from crashes, since the currently `vtbackup` is not designed to be crash-recoverable (except by creating a new `vtbackup` process).

## Checklist


maxenglander:
vtbackup_transform is failing consistently which is concerning, but I also see this test failing a bunch on `main` recently, so maybe it's a flake?

go/vt/mysqlctl/redo_log.go
```diff
@@ -0,0 +1,57 @@
...
+	return mysqld.ExecuteSuperQuery(ctx, "ALTER INSTANCE ENABLE INNODB REDO_LOG")
+}
+
+func (mysqld *Mysqld) IsRedoLogEnabled(ctx context.Context) (bool, error) {
+	if !mysqld.CanDisableRedoLog() {
+		// redo_log is most likely enabled, but return false anyway. Possible
+		// that mysqld is using non-InnoDB engine.
+		return false, fmt.Errorf("mysqld >= 8.0.21 required to inspect redo_log status")
+	}
+	qr, err := mysqld.FetchSuperQuery(ctx, "SHOW GLOBAL STATUS LIKE 'Innodb_redo_log_enabled'")
```
mattlord, line 49: Nit, but IMO it's nicer to do:
```
mysql> select variable_value from performance_schema.global_status where variable_name='innodb_redo_log_enabled';
+----------------+
| variable_value |
+----------------+
| ON             |
+----------------+
1 row in set (0.00 sec)
```

go/vt/mysqlctl/redo_log.go
```diff
@@ -0,0 +1,57 @@
...
+	if err != nil {
+		return false, err
+	}
+	if len(qr.Rows) != 1 {
+		return false, errors.New("no Innodb_redo_log_enabled variable in mysql")
+	}
+	return qr.Rows[0][1].ToString() == "ON", nil
```
mattlord, line 56: To be a little more robust:
```
val := strings.ToLower(qr.Rows[0][1].ToString())
return (val == "on" || val == "1"), nil 
```
To make it case insensitive and to manage the general equivalence MySQL has for variables/status where 1/ON and 0/OFF are equivalent.

go/vt/mysqlctl/redo_log.go
```diff
@@ -0,0 +1,57 @@
...
+		// redo_log is most likely enabled, but return false anyway. Possible
+		// that mysqld is using non-InnoDB engine.
+		return false, fmt.Errorf("mysqld >= 8.0.21 required to inspect redo_log status")
+	}
+	qr, err := mysqld.FetchSuperQuery(ctx, "SHOW GLOBAL STATUS LIKE 'Innodb_redo_log_enabled'")
+	if err != nil {
+		return false, err
+	}
+	if len(qr.Rows) != 1 {
+		return false, errors.New("no Innodb_redo_log_enabled variable in mysql")
```
mattlord, line 54:
```
return false, errors.New("no Innodb_redo_log_enabled status value found")
```

mattlord (changes requested):
We should add this check to the unit test: [go/mysql/flavor_test.go](https://github.com/vitessio/vitess/blob/main/go/mysql/flavor_test.go)

It would also be good to update the e2e test so that we confirm the expected behavior:
   - If the MySQL version is < 8.0.21 we didn't attempt anything
   - If the MySQL version is > 8.0.21 we did disable and re-enable redo logging

Unfortunately I don't see any metrics around disabling/enabling it... so this would be the best way I can find:
```
mysql> select * from performance_schema.error_log where logged < now()-interval 5 minute and data like "InnoDB redo logging%";
+----------------------------+-----------+---------+------------+-----------+--------------------------------------------------------------------------------------------------+
| LOGGED                     | THREAD_ID | PRIO    | ERROR_CODE | SUBSYSTEM | DATA                                                                                             |
+----------------------------+-----------+---------+------------+-----------+--------------------------------------------------------------------------------------------------+
| 2022-09-23 14:12:46.984801 |        24 | Warning | MY-013600  | InnoDB    | InnoDB redo logging is disabled. All data could be lost in case of a server crash.               |
| 2022-09-23 14:13:00.358888 |        24 | Warning | MY-013601  | InnoDB    | InnoDB redo logging is enabled. Data is now safe and can be recovered in case of a server crash. |
+----------------------------+-----------+---------+------------+-----------+--------------------------------------------------------------------------------------------------+
2 rows in set (0.00 sec)
```

mattlord:
> vtbackup_transform is failing consistently which is concerning, but I also see this test failing a bunch on `main` recently, so maybe it's a flake?

`vtbackup_transform` has been VERY flaky so not likely related. @rsajwani has been looking into it but we haven't found the cause yet (only seems to happen in the CI).

maxenglander:
Hey @mattlord I was able to address most of your suggestions except the E2E test. I tried adding some checks to the E2E test, but the problem seems to be that the E2E test is executing the `vtbackup` binary, which starts and destroys a fresh mysqld instance. From the test code, I don't see a way to access that mysqld.

go/cmd/vtbackup/vtbackup.go
```diff
@@ -349,6 +347,16 @@ func takeBackup(ctx context.Context, topoServer *topo.Server, backupStorage back
 		return fmt.Errorf("can't restore from backup: %v", err)
 	}
 
+	// Disable redo logging (if we can) before we start replication.
+	disabledRedoLog := false
+	if mysqld.CanDisableRedoLog() {
+		if err := mysqld.DisableRedoLog(ctx); err != nil {
+			log.Warningf("error disabling redo logging: %v", err)
```
mattlord, line 354: Nit, but I think we should use sentence capitalization for log messages

go/cmd/vtbackup/vtbackup.go
```diff
@@ -349,6 +347,16 @@ func takeBackup(ctx context.Context, topoServer *topo.Server, backupStorage back
 		return fmt.Errorf("can't restore from backup: %v", err)
 	}
 
+	// Disable redo logging (if we can) before we start replication.
+	disabledRedoLog := false
+	if mysqld.CanDisableRedoLog() {
+		if err := mysqld.DisableRedoLog(ctx); err != nil {
```
mattlord, line 353: What is the goal of this PR? Seems that it is to allow a just restored MySQL instance (by vtbackup) to catch back up with the shard replication position more quickly so that the larger shard backup operation finishes more quickly? I believe so, but had to learn more about vtbackup for this to be clear. :-)

go/cmd/vtbackup/vtbackup.go
```diff
@@ -431,6 +439,13 @@ func takeBackup(ctx context.Context, topoServer *topo.Server, backupStorage back
 		return fmt.Errorf("not taking backup: replication did not make any progress from restore point: %v", restorePos)
 	}
 
+	// Re-enable redo logging.
+	if disabledRedoLog {
+		if err := mysqld.EnableRedoLog(ctx); err != nil {
+			return fmt.Errorf("failed to enable redo log: %v", err)
```
mattlord, line 445: What happens now? I think vtbackup will throw things away and retry, right? This could be problematic with the new flag to keep the temporary data though, right? I would think we might run out of disk space.

go/vt/mysqlctl/cmd.go
```diff
@@ -30,6 +30,35 @@ import (
...
+}
+
+// CreateMysqldAndMycnf returns a Mysqld and a Mycnf object to use for working with a MySQL
+// installation that hasn't been set up yet.
+func CreateMysqldAndMycnfFromSources(tabletUID uint32, mysqlSocket string, mysqlPort int32) (*Mysqld, *Mycnf, error) {
+	mycnf, err := NewMycnfFromSources(tabletUID, mysqlPort)
+	if err != nil {
+		return nil, nil, err
+	}
+
```
mattlord, line 61: How are the changes in this file related to the PR? I don't see it mentioned anywhere as an issue that the PR is meant to address. Is this needed because of other work added in this PR? Or perhaps it's related to the new vitess pflag work?

go/vt/mysqlctl/mycnf.go
```diff
@@ -223,3 +223,70 @@ func ReadMycnf(mycnf *Mycnf) (*Mycnf, error) {
 
 	return mycnf, nil
 }
+
+// merge will create a new Mycnf produced from merging provided Mycnfs from
+// right-to-left, skipping non-zero values.
+func merge(cnfs ...*Mycnf) *Mycnf {
```
mattlord, line 229: Same question as above here.

go/vt/mysqlctl/mycnf_file.go
```diff
@@ -0,0 +1,15 @@
...
+
+func NewMycnfFromFile(tabletUID uint32, path string) (*Mycnf, error) {
+	cnf := NewMycnf(tabletUID, 0)
+	cnf.Path = path
+	cnf, err := ReadMycnf(cnf)
+	if err != nil {
+		return nil, err
+	}
+	return cnf, nil
+}
```
mattlord, line 15: And here :-)

go/vt/mysqlctl/mycnf_sources.go
```diff
@@ -0,0 +1,69 @@
+/*
+Copyright 2022 The Vitess Authors.
```
mattlord, line 2: Same question here for this file.

go/test/endtoend/cluster/vtbackup_process.go
```diff
@@ -52,9 +52,20 @@ type VtbackupProcess struct {
...
 }
 
+func (vtbackup *VtbackupProcess) Help() ([]byte, error) {
+	vtbackup.proc = exec.Command(
+		vtbackup.Binary,
+		"--help",
+	)
+
+	vtbackup.proc.Env = append(vtbackup.proc.Env, os.Environ()...)
+	log.Infof("Running vtbackup with args: %v", strings.Join(vtbackup.proc.Args, " "))
```
mattlord, line 62: Nit, but %s would be good in this case.

mattlord (commented):
@maxenglander Let's chat about this whenever you get back. No major concerns but I want to be sure that I understand everything fully/correctly. 

Thanks!

go/cmd/vtbackup/vtbackup.go
```diff
@@ -128,10 +128,12 @@ func registerFlags(fs *pflag.FlagSet) {
 	fs.BoolVar(&allowFirstBackup, "allow_first_backup", allowFirstBackup, "Allow this job to take the first backup of an existing shard.")
 	fs.BoolVar(&restartBeforeBackup, "restart_before_backup", restartBeforeBackup, "Perform a mysqld clean/full restart after applying binlogs, but before taking the backup. Only makes sense to work around xtrabackup bugs.")
 	// vttablet-like flags
+	fs.BoolVar(&keepTemporaryFiles, "keep-temporary-files", keepTemporaryFiles, "Retain temporary files and directories after exit.")
 	fs.StringVar(&initDbNameOverride, "init_db_name_override", initDbNameOverride, "(init parameter) override the name of the db used by vttablet")
 	fs.StringVar(&initKeyspace, "init_keyspace", initKeyspace, "(init parameter) keyspace to use for this tablet")
 	fs.StringVar(&initShard, "init_shard", initShard, "(init parameter) shard to use for this tablet")
 	fs.IntVar(&concurrency, "concurrency", concurrency, "(init restore parameter) how many concurrent files to restore at once")
+	fs.StringVar(&tabletPath, "tablet-path", tabletPath, "tablet alias")
```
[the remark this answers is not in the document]
maxenglander: Hey @mattlord I should have documented the intention in the PR description, but we can hash out the details here first, and once we decide whether these flags are needed I can improve the PR description and (if we keep them) flag descriptions.

> Were these flags added for e2e tests?

Yes, the only reason I added these flags was so that I could make the e2e tests work. 

Prior to the PR, the `vtbackup` binary [deletes all temporary files](https://github.com/vitessio/vitess/blob/55695d1/go/cmd/vtbackup/vtbackup.go#L244-L250) associated with the backup, including MySQL data and logs. I wasn't sure how else to answer the question "did MySQL disable and then redo the redo log during its run" without looking at the logs. Thus the `--keep-temporary-files` flag.

Prior to this PR, `vtbackup` internally [invents a random tablet alias](https://github.com/vitessio/vitess/blob/55695d1/go/cmd/vtbackup/vtbackup.go#L232-L239) for each run, and that [alias is then included in the file path](https://github.com/vitessio/vitess/blob/55695d1/go/cmd/vtbackup/vtbackup.go#L244) where log files are stored. I couldn't see how to figure out the log file location from the E2E tests, so I came up with passing `--tablet-path` in as a solution.

Now to answer some of your other questions...

> Is it a full tablet-alias, which determines the sub-directory to use within a backup dir? Or is it just a tablet alias? It doesn't seem like it's a full path, is it?

It is the path to the tablet data, and includes the tablet alias. It could be either a full path or a relative path. Internally, during a normal run `vtbackup` calculates a random relative path, but I believe a full path would work as well.

> So maybe we should just call this flag --tablet-uid or --tablet-alias

That would be fine with me, and would still let me solve some of the problems I ran into during the E2E tests. The alias is the only random part of the path. As long as I can be confident that `vtbackup` will store logs in a deterministic location, I can answer the question "did MySQL disable and then redo the redo log during its run".

FWIW the reason I chose `--tablet-path` was for consistency with other VT binaries. E.g. `vttablet` binary has [a `--tablet-path` flag](https://github.com/vitessio/vitess/blob/55695d1/go/cmd/vttablet/vttablet.go#L87).

> Why not make the flag default be vtbackup instead of hiding that in the code?

The behavior of `vtbackup` prior to this PR is to choose a random tablet alias each run. [There is a documented reason for that](https://github.com/vitessio/vitess/blob/55695d1/go/cmd/vtbackup/vtbackup.go#L227-L231), and I wasn't confident that I should change the logic.

> What's the use case for making this a user variable? I'm not sure how this part relates to the PR

As you sensed towards the end of your comment, these flags are purely to make it easier to write the E2E test.

> I feel like adding this flag only makes things more confusing for users w/o a clear use case it meets

Totally agree. Would prefer not to have these flags if we can think of a better way to answer the "did MySQL disable and then redo the redo log during its run" question. Alternately, happy to adjust the flags or give them better names. Or maybe pflags has a way to "hide" flags, so they don't show up when users run `--help`?

> The tablet-path also seems to impact the cell used. Is there any impact there? Is that part of why we allow the user to specify?

There's no impact to choosing a fake or made-up cell. `vtbackup` [already chooses a made-up cell](https://github.com/vitessio/vitess/blob/55695d1/go/cmd/vtbackup/vtbackup.go#L237). However, I'm not sure what would happen if a user chose a real cell and an already-existing UID. That might be bad.

Being able to choose the cell wasn't the motivation. Only motivation is being able to determine the location of MySQL log files.

go/cmd/vtbackup/vtbackup.go
maxenglander: You sensed the goal correctly. Sorry for not calling this out in the PR description. Have added a `Details` section in the PR description to make this more clear.

go/cmd/vtbackup/vtbackup.go
maxenglander: By default `vtbackup` will delete the temporary files, but you're right that if we `--keep-temporary-files` then we could run out of disk space.

go/vt/mysqlctl/cmd.go
maxenglander: All of the CNF stuff in this PR is a side-effect of trying to establish a deterministic log file location, in order to answer the E2E test question "did MySQL disable and then redo the redo log during its run".

maxenglander: To say a bit more about the CNF challenges:
 * Throughout the code base there are different ways to create a MySQL CNF:
   * Create an empty MyCNF, don't use `--flags` and on-disk path.
   * Using only mysqlctl `--flags`.
   * Using a combination of `--mycnf-path` and the contents of that path.
   * Using a tablet-default MyCNF path and the contents of that path.
 * Prior to the PR `vtbackup` [uses the "create empty MyCNF" variant](https://github.com/vitessio/vitess/blob/55695d1/go/cmd/vtbackup/vtbackup.go#L253).
 * With this PR, I wanted a way to be able to tell `vtbackup` where to store the MySQL error log. In order to do that, I reworked some of the mysqlctl CNF-loading code.
   * You might be asking yourself, "if you can tell vtbackup to store the error log at `--mycnf_error_log_path`, why do you even need `--tablet-path`?" For the life of me I cannot remember if I just didn't think of this until now, or if I ever had a good reason for it, but at the moment, it seems to me like I could just rely purely on setting `--mycnf_error_log_path` from the E2E test, and do without `--tablet-path`.

maxenglander:
@mattlord was able to do what you suggested: 

 * Connect to mysqld directly from the E2E test.
 * Verify redo log changes with performance schema error sink.
 * Remove those flags and the CNF cruft.

A bit worried about flakiness since mysqld could shut down before test is able to verify, but the test is passing consistently locally 👍

go/mysql/flavor.go
```diff
@@ -52,6 +52,7 @@ const (
 	MySQLJSONFlavorCapability
 	MySQLUpgradeInServerFlavorCapability
 	DynamicRedoLogCapacityFlavorCapability // supported in MySQL 8.0.30 and above: https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-30.html
+	DisableRedoLogCapacityFlavorCapability // supported in MySQL 8.0.21 and above: https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-21.html
```
mattlord: Sorry, I totally missed the name earlier. This should be: `DisableRedoLogFlavorCapability`. It doesn't have anything to do with the capacity.

go/test/endtoend/backup/vtbackup/backup_only_test.go
```diff
@@ -17,22 +17,23 @@ limitations under the License.
...
 import (
+	"context"
 	"fmt"
 	"os"
 	"path"
 	"strings"
 	"testing"
 	"time"
 
+	mysql "vitess.io/vitess/go/mysql"
```
mattlord: Nit, but we should remove the alias as it's no different than the package name.

mattlord (approved):
Nice work on this! ❤️

deepthi:
@maxenglander your last commit needed to include changes to unit tests.
It's good to run `make unit_test` locally before pushing changes. CI is a very expensive way to detect unit test failures.

maxenglander:
Understood @deepthi. In the case the most recent commits were a result of misunderstanding some PR comments that I got corrections for in Slack.

mattlord (approved):
Just officially noting that it LGTM after the subsequent changes.

PR #11331, opened 2022-09-22
[cli] [zkctl] Migrate zkctl flags to pflags

rsajwani:

## Description

Updates throttler related flag references specified in https://github.com/vitessio/vitess/issues/11294

## Related Issue(s)

#11294

## Checklist

### DIFF
--- a/go/cmd/zkctl/zkctl.go
+++ b/go/cmd/zkctl/zkctl.go
@@ -24,6 +24,8 @@ import (
 	"io"
 	"os"
 
+	"vitess.io/vitess/go/vt/servenv"
+
 	"github.com/spf13/pflag"
 
 	"vitess.io/vitess/go/exit"
@@ -42,11 +44,6 @@ Commands:
 `
 
 var (
-	zkCfg = flag.String("zk.cfg", "6@<hostname>:3801:3802:3803",
-		"zkid@server1:leaderPort1:electionPort1:clientPort1,...)")
-	myID = flag.Uint("zk.myid", 0,
-		"which server do you want to be? only needed when running multiple instance on one box, otherwise myid is implied by hostname")
-
 	// Reason for nolint : Used in line 54 (stdin = bufio.NewReader(os.Stdin)) in the init function
 	stdin *bufio.Reader //nolint
 )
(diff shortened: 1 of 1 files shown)


go/cmd/zkctl/zkctl.go
```diff
@@ -63,20 +61,22 @@ func main() {
...
+
 	log.RegisterFlags(fs)
 	logutil.RegisterFlags(fs)
-	_flag.Parse(fs)
-	args := _flag.Args()
+	args := servenv.ParseFlagsWithArgs("zkctl")
 
 	if len(args) == 0 {
-		flag.Usage()
+		fs.Usage()
```
ajm188, line 73: this doesn't include the global flags (defined on `servenv`'s global hooks), you would need to also grab a reference to the flag set that `servenv` is using (in an `OnParse` callback), store it in a variable, and then add those flags onto the `fs` variable you have here (via `fs.AddFlagSet`)

ajm188 (commented):
almost there

go/cmd/zkctl/zkctl.go
rsajwani: I just realized that line#72 is unreachable because if len(args) is zero then  ParseFlagWithArgs will exit after printing message. Thus simplifying the logic.

https://github.com/vitessio/vitess/blob/main/go/vt/servenv/servenv.go#L344

go/cmd/zkctl/zkctl.go
```diff
@@ -61,18 +60,10 @@ func main() {
...
 
 	fs := pflag.NewFlagSet("zkctl", pflag.ExitOnError)
-	fs.Usage = func() {
-		fmt.Fprintf(os.Stderr, usage)
-	}
-
 	log.RegisterFlags(fs)
 	logutil.RegisterFlags(fs)
-	args := servenv.ParseFlagsWithArgs("zkctl")
+	servenv.ParseFlagsWithArgs("zkctl")
```
ajm188, line 65: why are we parsing flags with args but then not doing anything with them?

ajm188: ah, i see, [this](https://github.com/vitessio/vitess/pull/11331/files/8a6f06a044fc7de1d36042bd991b0ace8bb92548..f20d7a532e15e3ba0b21663f7e4a14db1b0320cf#diff-832edc3db69da386548b153cd7268a79e0655592b1061dfb3cbadc1d03977a35R70) should be `args[0]`

rsajwani: args[0] and pflag.Arg(0) is producing same result so I used later one.. If you feel strongly then I can use args[0] .. I would advice if this is the only outstanding issue then please approve it and i will do that change before I update and merge ...

ajm188: you should use `args` because it's _technically_ different (there's some extra stuff happening under the hood to facilitate the migration in v14)

ajm188 (approved):
please make the `args` change and then merge

Issue #11292, opened 2022-09-21
Switch flag definitions to be on pflag instead of flag in `package go/cmd/zk`


ajm188:
Part of https://github.com/vitessio/vitess/issues/10697.

Current flags:

```
$ git grep -E "\bflag\.[A-Z]" -- go/cmd/zk/*.go
type cmdFunc func(ctx context.Context, subFlags *flag.FlagSet, args []string) error
        server = flag.String("server", "", "server(s) to connect to")
        _flag.SetUsage(flag.CommandLine, _flag.UsageOptions{ // TODO: hmmm
                flag.Usage()
        subFlags := flag.NewFlagSet(cmdName, flag.ExitOnError)
func cmdWait(ctx context.Context, subFlags *flag.FlagSet, args []string) error {
func cmdWatch(ctx context.Context, subFlags *flag.FlagSet, args []string) error {
func cmdLs(ctx context.Context, subFlags *flag.FlagSet, args []string) error {
func cmdTouch(ctx context.Context, subFlags *flag.FlagSet, args []string) error {
func cmdRm(ctx context.Context, subFlags *flag.FlagSet, args []string) error {
func cmdAddAuth(ctx context.Context, subFlags *flag.FlagSet, args []string) error {
func cmdCat(ctx context.Context, subFlags *flag.FlagSet, args []string) error {
func cmdEdit(ctx context.Context, subFlags *flag.FlagSet, args []string) error {
func cmdStat(ctx context.Context, subFlags *flag.FlagSet, args []string) error {
func cmdChmod(ctx context.Context, subFlags *flag.FlagSet, args []string) error {
func cmdCp(ctx context.Context, subFlags *flag.FlagSet, args []string) error {
func cmdZip(ctx context.Context, subFlags *flag.FlagSet, args []string) error {
func cmdUnzip(ctx context.Context, subFlags *flag.FlagSet, args []string) error {
```

PR #11332, opened 2022-09-23
[cli] [zk] Migrate zk flags to pflags

rsajwani:

## Description

Updates zk related flag references specified in https://github.com/vitessio/vitess/issues/11292 
Instead of using servenv I used pflag directly. Also the binary is not vitess specific so I kept it clean and did not add any go flags to it. I ended up doing some custom logic for --help/-help/h

## Checklist


go/cmd/zk/zkcmd.go
```diff
@@ -133,35 +130,41 @@ func init() {
...
+	server = pflag.String("server", "", "server(s) to connect to")
+
+	// handling case of --help, -help & -h
+	args := os.Args[1:]
+	var help bool
+	if args[0] == "-h" {
+		pflag.BoolVarP(&help, "help", "h", false, "display usage and exit")
+	} else if args[0] == "--help" || args[0] == "-help" {
+		pflag.BoolVar(&help, "help", false, "display usage and exit")
+	}
```
ajm188, line 145: why not just define the `BoolVarP` regardless? (also `--help` should work even if it's not the first arg ...)

go/cmd/zk/zkcmd.go
```diff
@@ -133,35 +130,41 @@ func init() {
...
 
-	fs := pflag.NewFlagSet("zkcmd", pflag.ExitOnError)
-	log.RegisterFlags(fs)
-	logutil.RegisterFlags(fs)
```
ajm188, line 146: you've lost the log/logutil flags, i think

go/cmd/zk/zkcmd.go
```diff
@@ -133,35 +130,41 @@ func init() {
...
+	if help || pflag.Arg(0) == "help" {
+		pflag.Usage()
+		os.Exit(0)
+	}
+
+	// if no zk command is provided after --server then we need to print doc & usage both
+	args = pflag.Args()
 	if len(args) == 0 {
-		flag.Usage()
+		fmt.Fprintf(os.Stderr, doc)
```
ajm188, line 157: you should override `pflag.CommandLine.Usage` instead of doing this

go/cmd/zk/zkcmd.go
```diff
@@ -193,12 +196,14 @@ func isZkFile(path string) bool {
...
-func cmdWait(ctx context.Context, subFlags *flag.FlagSet, args []string) error {
+func cmdWait(ctx context.Context, subFlags *pflag.FlagSet, args []string) error {
 	var (
 		exitIfExists = subFlags.Bool("e", false, "exit if the path already exists")
 	)
 
-	subFlags.Parse(args)
+	if err := subFlags.Parse(args); err != nil {
+		return err
+	}
```
ajm188, line 206: because you specified `ExitOnError` for the sub flags, this will never return with a non-nil error (it will, well, exit, instead)

ajm188: (to be clear, i think you want to keep this bit, but change the sub flags to `ContinueOnError`)

go/cmd/zk/zkcmd.go
```diff
@@ -456,13 +467,15 @@ func cmdTouch(ctx context.Context, subFlags *flag.FlagSet, args []string) error
 	return nil
 }
 
-func cmdRm(ctx context.Context, subFlags *flag.FlagSet, args []string) error {
+func cmdRm(ctx context.Context, subFlags *pflag.FlagSet, args []string) error {
 	var (
 		force             = subFlags.Bool("f", false, "no warning on nonexistent node")
 		recursiveDelete   = subFlags.Bool("r", false, "recursive delete")
 		forceAndRecursive = subFlags.Bool("rf", false, "shorthand for -r -f")
```
ajm188, line 474: here, if you do this:

```go
force = subFlags.BoolVarP("force", "f", ...)
recursive = subFlags.BoolVarP("recursive", "r", ...)
```

then pflag will let you combine short opts, so the following all work:

```
zk rm -f --recursive
zk rm --force -r
zk rm -f -r
zk rm -rf
```

go/cmd/zk/zkcmd.go
```diff
@@ -510,22 +523,26 @@ func cmdRm(ctx context.Context, subFlags *flag.FlagSet, args []string) error {
...
 	scheme, auth := subFlags.Arg(0), subFlags.Arg(1)
 	return zconn.AddAuth(ctx, scheme, []byte(auth))
 }
 
-func cmdCat(ctx context.Context, subFlags *flag.FlagSet, args []string) error {
+func cmdCat(ctx context.Context, subFlags *pflag.FlagSet, args []string) error {
 	var (
 		longListing = subFlags.Bool("l", false, "long listing")
 		force       = subFlags.Bool("f", false, "no warning on nonexistent node")
 		decodeProto = subFlags.Bool("p", false, "decode proto files and display them as text")
```
ajm188, line 541: you should move these to `BoolVarP` (and you should also update any other single-character flags to use full names with a `*VarP` function

go/cmd/zk/zkcmd.go
rsajwani: zkcmd is non-vitess binary. So I don't think it needs vitess flags... Therefore I kept it simple.. I have added it in description as well.

go/cmd/zk/zkcmd.go
rsajwani: BooVarP is only added when client is passing 'h'. If we add BoolVarP everytime then we will run into error when client pass -help. 
unknown shorthand flag: 'e' in -elp

I saw @GuptaManan100  also ran into same problem and it was fixed similar way. 
https://github.com/vitessio/vitess/pull/10911/files#diff-9a94072feabe23b4b635a733c84aa97dcf1a023e9ebc0d3a44daaf869038c8e0R109

go/cmd/zk/zkcmd.go
ajm188: it never calls into `log` functions? i find that surprising

go/cmd/zk/zkcmd.go
ajm188: no, that was done because `vtorc` hadn't been migrated with the deprecation notice before the last version.

you don't need to handle that case

ajm188: (for the record, i was the one that introduced that code, not manan 😂 )

go/cmd/zk/zkcmd.go
rsajwani: done

go/cmd/zk/zkcmd.go
rsajwani: But I am getting error if I just do BoolVarP for all cases. Haven't got time to debug through it though.

go/cmd/zk/zkcmd.go
rsajwani: How do I add log.RegisterFlags here given i am not using flagset and is using pure pflag fucntions..

go/cmd/zk/zkcmd.go
rsajwani: done

go/cmd/zk/zkcmd.go
rsajwani: done

go/cmd/zk/zkcmd.go
ajm188: with `pflag` you cannot pass `-help`, only `--help` and `-h` are allowed

go/cmd/zk/zkcmd.go
ajm188: `log.RegisterFlags(pflag.CommandLine)`

go/cmd/zk/zkcmd.go
```diff
@@ -134,14 +134,16 @@ func main() {
...
-		pflag.BoolVarP(&help, "help", "h", false, "display usage and exit")
-	} else if args[0] == "--help" || args[0] == "-help" {
-		pflag.BoolVar(&help, "help", false, "display usage and exit")
+	pflag.BoolVarP(&help, "help", "h", false, "display usage and exit")
+	log.RegisterFlags(pflag.CommandLine)
+	logutil.RegisterFlags(pflag.CommandLine)
+
+	pflag.CommandLine.Usage = func() {
+		fmt.Fprintf(os.Stderr, doc)
+		pflag.Usage()
```
ajm188, line 146: pretty sure this creates a stack overflow, have you tested this?

go/cmd/zk/zkcmd.go
```diff
@@ -289,11 +290,16 @@ func cmdWatch(ctx context.Context, subFlags *pflag.FlagSet, args []string) error
...
 func cmdLs(ctx context.Context, subFlags *pflag.FlagSet, args []string) error {
 	var (
-		longListing      = subFlags.Bool("l", false, "long listing")
-		directoryListing = subFlags.Bool("d", false, "list directory instead of contents")
-		force            = subFlags.Bool("f", false, "no warning on nonexistent node")
-		recursiveListing = subFlags.Bool("R", false, "recursive listing")
+		longListing      *bool
+		directoryListing *bool
+		force            *bool
+		recursiveListing *bool
```
ajm188, line 296: this doesn't work, these need to be `bool`, not `*bool`

ajm188: (and similar for all your unitizialized pointer variables)

go/cmd/zk/zkcmd.go
rsajwani: No pflg.usage and pflag.commandline.Usage are different functions.. I initially thought same but it isn't the case

go/cmd/zk/zkcmd.go
```diff
@@ -111,10 +107,11 @@ const (
 	timeFmtMicro = "2006-01-02 15:04:05.000000"
 )
 
-type cmdFunc func(ctx context.Context, subFlags *flag.FlagSet, args []string) error
+type cmdFunc func(ctx context.Context, subFlags *pflag.FlagSet, args []string) error
 
 var cmdMap map[string]cmdFunc
 var zconn *zk2topo.ZkConn
+var server *string
```
ajm188, line 114: if we're going to declare this up here, let's go with the standard form:

```go
var server string

// below
pflag.StringVar(&server)
```

go/cmd/zk/zkcmd.go
```diff
@@ -195,12 +197,14 @@ func isZkFile(path string) bool {
 	return strings.HasPrefix(path, "/zk")
 }
 
-func cmdWait(ctx context.Context, subFlags *flag.FlagSet, args []string) error {
-	var (
-		exitIfExists = subFlags.Bool("e", false, "exit if the path already exists")
-	)
+func cmdWait(ctx context.Context, subFlags *pflag.FlagSet, args []string) error {
+
```
ajm188, line 201: nit: remove this blank

go/vt/proto/vtrpc/vtrpc.pb.go
```diff
@@ -39,7 +39,9 @@ const (
 
 // Code represents canonical error codes. The names, numbers and comments
 // must match the ones defined by grpc (0-16):
-//   https://godoc.org/google.golang.org/grpc/codes.
+//
```
ajm188, line 42: you should revert this change (it's unrelated to your PR)

ajm188 (approved):
make the few changes (especially the protobuf diff and the one about `server`), then this is ready to merge

go/vt/proto/vtrpc/vtrpc.pb.go
rsajwani: this is done by gofmt.

ajm188: i understand that, but the commit hooks _should_ only be running on packages you actually touched

ajm188: which is why [this is failing](https://github.com/vitessio/vitess/actions/runs/3153516810/jobs/5131699790#step:23:1493)

Issue #10750, opened 2022-07-19
Switch flag definitions to be on pflag instead of flag in `package go/vt/dbconfigs`


ajm188:
Part of https://github.com/vitessio/vitess/issues/10697.

Current flags:

```
$ git grep -E "\bflag\.[A-Z]" -- go/vt/dbconfigs/*.go
go/vt/dbconfigs/credentials.go: dbCredentialsServer = flag.String("db-credentials-server", "file", "db credentials server type ('file' - file implementation; 'vault' - HashiCorp Vault implementation)")
go/vt/dbconfigs/credentials.go: dbCredentialsFile = flag.String("db-credentials-file", "", "db credentials file; send SIGHUP to reload this file")
go/vt/dbconfigs/credentials.go: vaultAddr             = flag.String("db-credentials-vault-addr", "", "URL to Vault server")
go/vt/dbconfigs/credentials.go: vaultTimeout          = flag.Duration("db-credentials-vault-timeout", 10*time.Second, "Timeout for vault API operations")
go/vt/dbconfigs/credentials.go: vaultCACert           = flag.String("db-credentials-vault-tls-ca", "", "Path to CA PEM for validating Vault server certificate")
go/vt/dbconfigs/credentials.go: vaultPath             = flag.String("db-credentials-vault-path", "", "Vault path to credentials JSON blob, e.g.: secret/data/prod/dbcreds")
go/vt/dbconfigs/credentials.go: vaultCacheTTL         = flag.Duration("db-credentials-vault-ttl", 30*time.Minute, "How long to cache DB credentials from the Vault server")
go/vt/dbconfigs/credentials.go: vaultTokenFile        = flag.String("db-credentials-vault-tokenfile", "", "Path to file containing Vault auth token; token can also be passed using VAULT_TOKEN environment variable")
go/vt/dbconfigs/credentials.go: vaultRoleID           = flag.String("db-credentials-vault-roleid", "", "Vault AppRole id; can also be passed using VAULT_ROLEID environment variable")
go/vt/dbconfigs/credentials.go: vaultRoleSecretIDFile = flag.String("db-credentials-vault-role-secretidfile", "", "Path to file containing Vault AppRole secret_id; can also be passed using VAULT_SECRETID environment variable")
[...]

PR #11336, opened 2022-09-23
Move dbconfigs to pflag and remove deprecated flags

mattlord:
## Description

This migrates the `dbconfig` package flags from go flag to pflag. It also takes the opportunity to finally remove all of the long deprecated flags (first [deprecated in 3.0](https://github.com/vitessio/vitess/commit/3f904811afe67b98999751473e8f23e27a860cbe)).

## Related Issue(s)

  - Fixes: https://github.com/vitessio/vitess/issues/10750

## Checklist


go/vt/dbconfigs/dbconfigs.go
```diff
@@ -49,7 +51,7 @@ const (
 )
 
 var (
-	// GlobalDBConfigs contains the initial values of dbconfgis from flags.
+	// GlobalDBConfigs contains the initial values of dbconfgs from flags.
```
ajm188, line 54: you, uh, swapped a typo for a different typo 😂

go/vt/dbconfigs/dbconfigs.go
```diff
@@ -113,64 +115,47 @@ type UserConfig struct {
...
-	flag.StringVar(&GlobalDBConfigs.SslCert, "db_ssl_cert", "", "connection ssl certificate")
-	flag.StringVar(&GlobalDBConfigs.SslKey, "db_ssl_key", "", "connection ssl key")
-	flag.StringVar(&GlobalDBConfigs.TLSMinVersion, "db_tls_min_version", "", "Configures the minimal TLS version negotiated when SSL is enabled. Defaults to TLSv1.2. Options: TLSv1.0, TLSv1.1, TLSv1.2, TLSv1.3.")
-	flag.StringVar(&GlobalDBConfigs.ServerName, "db_server_name", "", "server name of the DB we are connecting to.")
-	flag.IntVar(&GlobalDBConfigs.ConnectTimeoutMilliseconds, "db_connect_timeout_ms", 0, "connection timeout to mysqld in milliseconds (0 for no timeout)")
-	flag.BoolVar(&GlobalDBConfigs.EnableQueryInfo, "db_conn_query_info", false, "enable parsing and processing of QUERY_OK info fields")
+	servenv.OnParse(func(fs *pflag.FlagSet) {
+		registerBaseFlags(fs)
+		for _, userKey := range userKeys {
+			uc, cp := GlobalDBConfigs.getParams(userKey, &GlobalDBConfigs)
```
ajm188, line 126: i don't have a particular answer in mind, but the question occurred to me: for both the base flags (on `GlobalDBConfigs`) and the per-user flags, does it make sense to use the variable as the default value?

by which i mean, instead of `fs.StringVar(&GlobalDBConfigs.Socket, "db_socket", "", "the unix socket ...")` we do `fs.StringVar(&GlobalDBConfigs.Socket, "db_socket", GlobalDBConfigs.Socket, "the unix socket ...")` and so on?

go/vt/vttls/vttls.go
```diff
@@ -61,6 +61,12 @@ func (mode *SslMode) String() string {
 	return string(*mode)
 }
 
+// Type returns the value type, part of the pflag Value interface
+// for allowing this to be used as a generic flag.
+func (mode *SslMode) Type() string {
+	return "SslMode"
```
ajm188, line 67: imo this should return "string" but idk what the best-practice for pflag is here see for example how `StringSliceVar` shows up as "strings" in the help output

ajm188 (approved):
this looks great!!! awesome stuff ❤️

go/vt/vttls/vttls.go
mattlord: That's what I thought too, that's what I started with but then had failures.

go/vt/dbconfigs/dbconfigs.go
mattlord: Good catch. That's what I've been doing, but overlooked it here. Will update it.

go/vt/vttls/vttls.go
ajm188: no no, i mean the _value_ returned by this function should be the literal string `"string"`, not the string `"SslMode"`

ajm188 (approved):

go/vt/vttls/vttls.go
mattlord: That's what I thought you meant. It's also what I meant. 😄

ajm188: oh weird! TIL

PR #11338, opened 2022-09-23
Addition of Metrics to VTOrc to track the number of recoveries ran and their success count.

GuptaManan100:
## Description

This PR adds metrics to VTOrc to track the different number of recoveries that it runs and the count of them that fail and succeed.

The debug/vars page looks like the following after these changes - 
Screenshot 2022-09-23 at 3 19 30 PM

## Related Issue(s)

- #6612 

## Checklist


rsajwani:
LGTM.. clean code.

rsajwani (approved):

PR #11344, opened 2022-09-23
VStreams: Rotate Binary Log For Snapshot Connections

mattlord:
## Description

Rotate the binary logs as needed in order to limit the GTID auto positioning overhead when starting a results/row stream from a given GTID set/position in VStreams. This is needed as the currently open binary log (which can be up to 1G in size by default) will need to be scanned and empty GTID events will be streamed for those GTIDs in the log that we are skipping. In total, this can add a lot of overhead on both the mysqld instance and the tablet. It can also cause a significant delay — particularly on connections with higher latency such as when the source and target are communicating over the internet with TLS — for otherwise fast actions such as VDiff on relatively small tables. Rotating the log ensures that we are processing a fresh binary log that will be minimal in size and GTID events.

> **Note**
>  We only rotate the binary log if it's > 64MiB in size by default. This avoids unnecessary flushes when there are many vstreams running  concurrently and/or there's very little write activity happening.

We had previously discovered this behavior during other events which require syncing with a GTID position/set, specifically during reparents, and used the same solution. See related issue and its links below.

## Related Issue(s)

  - Fixes: https://github.com/vitessio/vitess/issues/11345
  - Related to: https://github.com/vitessio/vitess/issues/4161

## Checklist

-   [x] Tests were added
-   [x] Documentation: https://github.com/vitessio/website/pull/1153

### DIFF
--- a/go/flags/endtoend/vttablet.txt
+++ b/go/flags/endtoend/vttablet.txt
@@ -465,6 +465,7 @@ Usage of vttablet:
       --vreplication_retry_delay duration                                delay before retrying a failed workflow event in the replication phase (default 5s)
       --vreplication_store_compressed_gtid                               Store compressed gtids in the pos column of _vt.vreplication
       --vreplication_tablet_type string                                  comma separated list of tablet types used as a source (default "in_order:REPLICA,PRIMARY")
+      --vstream_binlog_rotation_threshold int                            Byte size at which a VStreamer will attempt to rotate the source's open binary log before starting a GTID snapshot based stream (e.g. a ResultStreamer or RowStreamer) (default 67108864)
       --vstream_dynamic_packet_size                                      Enable dynamic packet sizing for VReplication. This will adjust the packet size during replication to improve performance. (default true)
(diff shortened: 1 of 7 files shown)


aquarapid (commented):
Shouldn't this just ignore/log if it fails?  E.g. if we don't have the RELOAD privilege, or are doing the snapshot from a replica?

rohit-nayak-ps (approved):
lgtm, this is nice!

rohit-nayak-ps (approved):
lgtm

aquarapid (commented):
LGTM!  Only other thing I could suggest is to test the error path (e.g. if the show binary logs fails because of lack of RELOAD privs)

mattlord:
> LGTM! Only other thing I could suggest is to test the error path (e.g. if the show binary logs fails because of lack of RELOAD privs)

I assume you mean for the log message here: https://github.com/vitessio/vitess/pull/11344/files#diff-efb6bf2b0113d05b4466e1c54f4abf7ccad01637cc8ad1f01ceb2bb916dd67fdR81-R82

If so, we log the error info we get back from MySQL which is pretty explanatory IMO:
```
ERROR 1227 (42000): Access denied; you need (at least one of) the RELOAD privilege(s) for this operation
```

Please let me know if you had something else in mind though. Thanks!

PR #11352, opened 2022-09-26
Flakes: Fix Backup Transform Test Flakiness

mattlord:
## Description

The `TestBackupTransform` test was not fully adjusted for MySQL Semi-Sync replication when we enabled it for all tests. I say that because the test only initially setup a single PRIMARY and a single REPLICA tablet. With [lossless semi-sync](http://my-replication-life.blogspot.com/2013/09/loss-less-semi-synchronous-replication.html) — which is [the default ](https://dev.mysql.com/doc/refman/en/replication-options-source.html#sysvar_rpl_semi_sync_master_wait_point)— this means that we only have a single ACKer and if that ACKer / REPLICA tablet is shutdown after receiving the GTID but before ACKing it, the GTID will exist on the REPLICA (will be applied from the local relay log by the time it starts up again) but will NOT exist on the PRIMARY. At this point the PRIMARY is also left indefinitely waiting for an ACK and is blocked from making further changes.

In the built-in backup engine, we enter this for loop because the REPLICA's GTID position will not equal the PRIMARY's, and we do not exit the for loop until we hit the 1 hour timeout for the Backup RPC call because the REPLICA's position will never change/advance because we don't execute any more statements on the PRIMARY (and if we tried, the PRIMARY will block it anyway as it's still waiting for an ACK on the previous GTID): https://github.com/vitessio/vitess/blob/d281e580a15b777208adf44ed2d584f4647ba917/go/vt/mysqlctl/builtinbackupengine.go#L262-L283

This PR addresses the problems by:
1. Initially setting up 2 REPLICA tablets so that we have 2 ACKers and the `replica2` will also ACK the GTID even if `replica1` does not due to being shutdown for the backup
2. Confirming that semi-sync replication is setup properly on both REPLICAs
3. Confirming that the test data exists on the PRIMARY and both REPLICAs before proceeding with the Backup

For whatever reasons, the test when run in the CI would trigger this race condition regularly. Before this PR, I would guess that the `TestBackupTransform` test was anecdotally failing over 50% of the time based on my personal experience. With these fixes, it successfully ran 20 times in a row:
  - https://github.com/vitessio/vitess/actions/runs/3131716643 (failed workflow runs were NOT due to the transform test)
  - https://github.com/vitessio/vitess/actions/runs/3132595831 (after the fix for reliably clearing out existing tablet data noted below, BOTH tests in the workflow passed every time — 12 times in a row)

> **Note**
> The workflow was still pretty flaky because of the `TestBackup` test. That needed to be investigated and fixed separately. In those failures I saw this sequence:
>   - `Error:      	Expected nil, but got: &fs.PathError{Op:"unlinkat", Path:"/tmp/vt_685235216/vtroot_16001/vt_0000002723", Err:0x27}` (0x27 is decimal 39 which is OS error `ENOTEMPTY`) followed by:
>   - `vttablet.go:170] mycnf read failed: open /tmp/vt_784732455/vtroot_16001/vt_0000005331/my.cnf: no such file or directory` followed by:
>   - `Expected nil, but got: &errors.errorString{s:"process 'vttablet' timed out after 10s (err: process 'vttablet' exited prematurely (err: exit status 1))"}`

So in order to fix that remaining flakiness due to the sequence of errors (successful runs did not have the `unlinkat` errors while all of the failed ones did) this PR also [uses a more reliable method for clearing out existing vttablet/mysql data](https://github.com/vitessio/vitess/pull/11352/commits/351e1b698909cbac493783f30bc3cf998e6cc986).

## Checklist

-   [x] Tests were updated

### DIFF
--- a/go/test/endtoend/backup/transform/backup_transform_utils.go
+++ b/go/test/endtoend/backup/transform/backup_transform_utils.go
@@ -217,18 +217,15 @@ func TestBackupTransformImpl(t *testing.T) {
 	err = localCluster.VtctlclientProcess.ExecuteCommand("Backup", replica1.Alias)
 	require.Nil(t, err)
 
-	log.Errorf("Inserting data...")
 	// insert data in primary
 	_, err = primary.VttabletProcess.QueryTablet("insert into vt_insert_test (msg) values ('test2')", keyspaceName, true)
 	require.Nil(t, err)
 
-	log.Errorf("Verifying backup...")
 	// validate backup_list, expecting 1 backup available
 	backups := localCluster.VerifyBackupCount(t, shardKsName, 1)
 
 	backupLocation := localCluster.CurrentVTDATAROOT + "/backups/" + shardKsName + "/" + backups[0]
 
-	log.Errorf("Validating manifest...")
 	// validate that MANIFEST has TransformHook
 	// every file should start with 'header'
 	validateManifestFile(t, backupLocation)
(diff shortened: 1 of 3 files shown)


GuptaManan100 (approved):
The changes look good to me!

PR #11357, opened 2022-09-26
refactor: minor refactor in partial shard routing  and change in flag to dashes

harshit-gangal:
## Description

This PR addresses some review comments from https://github.com/vitessio/vitess/pull/9987

## Checklist


rohit-nayak-ps (approved):
lgtm ❤️

PR #11364, opened 2022-09-27
[vtexplain] Migrate to pflags

### LINKED ISSUE #11285: Switch flag definitions to be on pflag instead of flag in `package go/cmd/vtexplain`

Current flags:

```
$ git grep -E "\bflag\.[A-Z]" -- go/cmd/vtexplain/*.go
        sqlFlag            = flag.String("sql", "", "A list of semicolon-delimited SQL commands to analyze")
        sqlFileFlag        = flag.String("sql-file", "", "Identifies the file that contains the SQL commands to analyze")
        schemaFlag         = flag.String("schema", "", "The SQL table schema")
        schemaFileFlag     = flag.String("schema-file", "", "Identifies the file that contains the SQL table schema")
        vschemaFlag        = flag.String("vschema", "", "Identifies the VTGate routing schema")
        vschemaFileFlag    = flag.String("vschema-file", "", "Identifies the VTGate routing schema file")
        ksShardMapFlag     = flag.String("ks-shard-map", "", "JSON map of keyspace name -> shard name -> ShardReference object. The inner map is the same as the output of FindAllShardsInKeyspace")
        ksShardMapFileFlag = flag.String("ks-shard-map-file", "", "File containing json blob of keyspace name -> shard name -> ShardReference object")
        numShards          = flag.Int("shards", 2, "Number of shards per keyspace. Passing --ks-shard-map/--ks-shard-map-file causes this flag to be ignored.")
        executionMode      = flag.String("execution-mode", "multi", "The execution mode to simulate -- must be set to multi, legacy-autocommit, or twopc")
        replicationMode    = flag.String("replication-mode", "ROW", "The replication mode to simulate -- must be set to either ROW or STATEMENT")
        normalize          = flag.Bool("normalize", false, "Whether to enable vtgate normalization")
        outputMode         = flag.String("output-mode", "text", "Output in human-friendly text or json")
        dbName             = flag.String("dbname", "", "Optional database target to override normal routing")
        plannerVersionStr  = flag.String("planner-version", "", "Sets the query planner version to use when generating the explain output. Valid values are V3 and Gen4")
        badPlannerVersion  = flag.String("planner_version", "", "Deprecated flag. Use planner-version instead")
        flag.CommandLine.SetOutput(logutil.NewLoggerWriter(logger))
        _flag.SetUsage(flag.CommandLine, _flag.UsageOptions{
                FlagFilter: func(f *flag.Flag) bool {
```

systay:
## Description
Switches `vtexplain` to use pflags according to https://github.com/vitessio/vitess/issues/10697

## Related Issue(s)
Fixes #11285

### DIFF
--- a/go/cmd/vtexplain/vtexplain.go
+++ b/go/cmd/vtexplain/vtexplain.go
@@ -41,33 +41,37 @@ var (
 	vschemaFileFlag    string
 	ksShardMapFlag     string
 	ksShardMapFileFlag string
-	numShards          int
-	executionMode      string
 	replicationMode    string
 	normalize          bool
-	outputMode         string
 	dbName             string
 	plannerVersionStr  string
+
+	numShards     = 2
+	executionMode = "multi"
+	outputMode    = "text"
 )
 
+func registerFlags(fs *pflag.FlagSet) {
+	fs.StringVar(&sqlFlag, "sql", "", "A list of semicolon-delimited SQL commands to analyze")
+	fs.StringVar(&sqlFileFlag, "sql-file", "", "Identifies the file that contains the SQL commands to analyze")
+	fs.StringVar(&schemaFlag, "schema", "", "The SQL table schema")
+	fs.StringVar(&schemaFileFlag, "schema-file", "", "Identifies the file that contains the SQL table schema")
+	fs.StringVar(&vschemaFlag, "vschema", "", "Identifies the VTGate routing schema")
+	fs.StringVar(&vschemaFileFlag, "vschema-file", "", "Identifies the VTGate routing schema file")
+	fs.StringVar(&ksShardMapFlag, "ks-shard-map", "", "JSON map of keyspace name -> shard name -> ShardReference object. The inner map is the same as the output of FindAllShardsInKeyspace")
+	fs.StringVar(&ksShardMapFileFlag, "ks-shard-map-file", "", "File containing json blob of keyspace name -> shard name -> ShardReference object")
+	fs.StringVar(&replicationMode, "replication-mode", "ROW", "The replication mode to simulate -- must be set to either ROW or STATEMENT")
+	fs.BoolVar(&normalize, "normalize", false, "Whether to enable vtgate normalization")
+	fs.StringVar(&dbName, "dbname", "", "Optional database target to override normal routing")
+	fs.StringVar(&plannerVersionStr, "planner-version", "", "Sets the query planner version to use when generating the explain output. Valid values are V3 and Gen4")
+
+	fs.IntVar(&numShards, "shards", numShards, "Number of shards per keyspace. Passing --ks-shard-map/--ks-shard-map-file causes this flag to be ignored.")
+	fs.StringVar(&executionMode, "execution-mode", executionMode, "The execution mode to simulate -- must be set to multi, legacy-autocommit, or twopc")
+	fs.StringVar(&outputMode, "output-mode", outputMode, "Output in human-friendly text or json")
+}
+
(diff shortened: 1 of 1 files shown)


go/cmd/vtexplain/vtexplain.go
```diff
@@ -29,62 +28,46 @@ import (
...
-		FlagFilter: func(f *flag.Flag) bool {
-			for _, name := range vtexplainFlags {
-				if f.Name == name {
-					return true
-				}
-			}
-
-			return false
-		},
+	servenv.OnParse(func(fs *pflag.FlagSet) {
```
GuptaManan100, line 55: We could have used `OrParseFor("vtexplain", func...)` but I don't think it matters too much since it is the cmd file

GuptaManan100 (approved):

go/cmd/vtexplain/vtexplain.go
systay: @ajm188 wdyt?

ajm188: i have been pushing toward using simply `OnParse` when you're in a cmd; the argument i have used is since it's a `main` package, no one can import you, so there's no material difference between a flag registered to "a binary" (since there is just the one), and "all binaries".

go/cmd/vtexplain/vtexplain.go
```diff
@@ -29,62 +28,46 @@ import (
...
-	}
+	sqlFlag            string
+	sqlFileFlag        string
+	schemaFlag         string
+	schemaFileFlag     string
+	vschemaFlag        string
+	vschemaFileFlag    string
+	ksShardMapFlag     string
+	ksShardMapFileFlag string
+	numShards          int
```
ajm188, line 45: this used to default to 2; the pattern we have been doing when the default is not the zero-value for the given type is:

```go
var myFlag = 2


    fs.IntVar(&myFlag, "my-flag", myFlag, "my flag's usage text")
}
```

ajm188: (and then ditto for the `string` variables that don't default to the empty string, and so on)

frouioui: til

go/cmd/vtexplain/vtexplain.go
```diff
@@ -29,63 +28,51 @@ import (
...
+
+	numShards     = 2
+	executionMode = "multi"
+	outputMode    = "text"
 )
 
+func registerFlags(fs *pflag.FlagSet) {
+	fs.StringVar(&sqlFlag, "sql", "", "A list of semicolon-delimited SQL commands to analyze")
+	fs.StringVar(&sqlFileFlag, "sql-file", "", "Identifies the file that contains the SQL commands to analyze")
+	fs.StringVar(&schemaFlag, "schema", "", "The SQL table schema")
```
ajm188, line 58: these should still use the variable as the "default value" argument, for example:

```diff
- fs.StringVar(&sqlFlag, "sql", "", "A list of semicolon-delimited SQL commands to analyze")
+ fs.StringVar(&sqlFlag, "sql", sqlFlag, "A list of semicolon-delimited SQL commands to analyze")
```

the reason being if we ever change the variable default, then we don't need to remember to also update the flag default, since they are inextricably linked

ajm188 (approved):
one last thing, then merge merge merge!

go/cmd/vtexplain/vtexplain.go
systay: 👍

Issue #11287, opened 2022-09-21
Switch flag definitions to be on pflag instead of flag in `package go/cmd/vtbench`


ajm188:
Part of https://github.com/vitessio/vitess/issues/10697.

Current flags:

```
$ git grep -E "\bflag\.[A-Z]" -- go/cmd/vtbench/*.go
        host       = flag.String("host", "", "vtgate host(s) in the form 'host1,host2,...'")
        port       = flag.Int("port", 0, "vtgate port")
        unixSocket = flag.String("unix_socket", "", "vtgate unix socket")
        protocol   = flag.String("protocol", "mysql", "client protocol, either mysql (default), grpc-vtgate, or grpc-vttablet")
        user       = flag.String("user", "", "username to connect using mysql (password comes from the db-credentials-file)")
        db         = flag.String("db", "", "db name to use when connecting / running the queries (e.g. @replica, keyspace, keyspace/shard etc)")
        deadline = flag.Duration("deadline", 5*time.Minute, "maximum duration for the test run (default 5 minutes)")
        sql      = flag.String("sql", "", "sql statement to execute")
        threads  = flag.Int("threads", 2, "number of parallel threads to run")
        count    = flag.Int("count", 1000, "number of queries per thread")
        flag.CommandLine.SetOutput(logutil.NewLoggerWriter(logger))
        flag.Lookup("logtostderr").Value.Set("true")
```

PR #11366, opened 2022-09-27
Migrates `cmd/vtbench` to pflags

frouioui:
## Description

This PR changes `cmd/vtbench` to use pflags as instructed in https://github.com/vitessio/vitess/issues/10697.

## Related Issue(s)

- Fixes https://github.com/vitessio/vitess/issues/11287

## Checklist


go/cmd/vtbench/vtbench.go
```diff
@@ -86,22 +86,32 @@ import (
...
+	pflag.IntVar(&threads, "threads", threads, "number of parallel threads to run")
+	pflag.IntVar(&count, "count", count, "number of queries per thread")
+}
+
 func main() {
+	servenv.OnParse(func(fs *pflag.FlagSet) {
+		initFlags()
+	})
 	logger := logutil.NewConsoleLogger()
 	flag.CommandLine.SetOutput(logutil.NewLoggerWriter(logger))
```
frouioui, line 116: @ajm188, I was not sure whether this needs to be removed or not. I can see that https://github.com/vitessio/vitess/pull/11364 is getting rid of it. Let me know what I should follow :)

ajm188: well, it depends on what you want 😅 

if you want to keep the behavior, you want to something like:

```go
servenv.OnParse(func(fs *pflag.FlagSet) {
    fs.SetOutput(logutil.NewLoggerWriter(logger))
}
```

go/cmd/vtbench/vtbench.go
```diff
@@ -18,16 +18,16 @@ package main
...
 	"context"
-	"flag"
 	"fmt"
 	"strings"
 	"time"
 
 	"github.com/spf13/pflag"
 
+	"vitess.io/vitess/go/vt/grpccommon"
+
```
GuptaManan100, line 28: We don't really need a space here, but I don't care as long as `goimports` passes

go/cmd/vtbench/vtbench.go
```diff
@@ -86,89 +83,97 @@ import (
...
+	fs.StringVar(&unixSocket, "unix_socket", "", "vtgate unix socket")
+	fs.StringVar(&protocol, "protocol", protocol, "client protocol, either mysql (default), grpc-vtgate, or grpc-vttablet")
+	fs.StringVar(&user, "user", "", "username to connect using mysql (password comes from the db-credentials-file)")
+	fs.StringVar(&db, "db", "", "db name to use when connecting / running the queries (e.g. @replica, keyspace, keyspace/shard etc)")
 
-	defer exit.Recover()
+	fs.DurationVar(&deadline, "deadline", deadline, "maximum duration for the test run (default 5 minutes)")
+	fs.StringVar(&sql, "sql", "", "sql statement to execute")
+	fs.IntVar(&threads, "threads", threads, "number of parallel threads to run")
+	fs.IntVar(&count, "count", count, "number of queries per thread")
```
GuptaManan100, line 105: We should use the variable itself for the default value. For example `fs.IntVar(&port, "port", port, "vtgate port")`

go/cmd/vtbench/vtbench.go
```diff
@@ -86,89 +83,97 @@ import (
...
+	threads                         = 2
+	count                           = 1000
 )
 
-func main() {
-	logger := logutil.NewConsoleLogger()
-	flag.CommandLine.SetOutput(logutil.NewLoggerWriter(logger))
+func initFlags(fs *pflag.FlagSet) {
+	fs.StringVar(&host, "host", "", "vtgate host(s) in the form 'host1,host2,...'")
+	fs.IntVar(&port, "port", 0, "vtgate port")
```
GuptaManan100, line 96: It is isn't strictly part of the flags move, but we want to start capitalizing the help messages

go/cmd/vtbench/vtbench.go
```diff
@@ -86,89 +83,97 @@ import (
...
+	port                            int
+	protocol                        = "mysql"
+	deadline                        = 5 * time.Minute
+	threads                         = 2
+	count                           = 1000
 )
 
-func main() {
-	logger := logutil.NewConsoleLogger()
-	flag.CommandLine.SetOutput(logutil.NewLoggerWriter(logger))
```
GuptaManan100, line 106: You still need an equivalent of this line, which is pretty much the same but using pflags.

go/cmd/vtbench/vtbench.go
frouioui: I was thinking that since the default value is 0 for both the variable and the flag I might as well use a literal, just like we do for string flags

go/cmd/vtbench/vtbench.go
frouioui: I'll change that

go/cmd/vtbench/vtbench.go
frouioui: Right, i'll add this too

go/cmd/vtbench/vtbench.go
ajm188: see my comment [here](https://github.com/vitessio/vitess/pull/11364#discussion_r982233227) (we should _also_ not use literals for the string flags)

GuptaManan100 (approved):
LGTM!

Issue #11267, opened 2022-09-20
Switch flag definitions to be on pflag instead of flag in `package go/cmd/mysqlctld`


mattlord:
Part of https://github.com/vitessio/vitess/issues/10697.

Current flags:

```
$ git grep -E "\bflag\.[A-Z]" -- go/cmd/mysqlctld/*.go
go/cmd/mysqlctld/mysqlctld.go:  mysqlPort   = flag.Int("mysql_port", 3306, "mysql port")
go/cmd/mysqlctld/mysqlctld.go:  tabletUID   = flag.Uint("tablet_uid", 41983, "tablet uid")
go/cmd/mysqlctld/mysqlctld.go:  mysqlSocket = flag.String("mysql_socket", "", "path to the mysql socket")
go/cmd/mysqlctld/mysqlctld.go:  waitTime      = flag.Duration("wait_time", 5*time.Minute, "how long to wait for mysqld startup or shutdown")
go/cmd/mysqlctld/mysqlctld.go:  initDBSQLFile = flag.String("init_db_sql_file", "", "path to .sql file to run after mysql_install_db")
```

PR #11376, opened 2022-09-28
Migrate mysqlctld from flag to pflag

mattlord:
## Description

This moves the `mysqlctld` command/binary from using flag to pflag.

## Related Issue(s)

  - Fixes: https://github.com/vitessio/vitess/issues/11267

## Checklist


GuptaManan100 (approved):
LGTM!

go/cmd/mysqlctld/mysqlctld.go
```diff
@@ -54,14 +55,21 @@ func init() {
...
 	servenv.RegisterGRPCServerAuthFlags()
 	servenv.RegisterServiceMapFlag()
+	// mysqlctld only starts and stops mysql, only needs dba.
+	dbconfigs.RegisterFlags(dbconfigs.Dba)
+	servenv.OnParse(func(fs *pflag.FlagSet) {
+		fs.IntVar(&mysqlPort, "mysql_port", mysqlPort, "mysql port")
+		fs.UintVar(&tabletUID, "tablet_uid", tabletUID, "tablet uid")
+		fs.StringVar(&mysqlSocket, "mysql_socket", mysqlSocket, "path to the mysql socket")
+		fs.DurationVar(&waitTime, "wait_time", waitTime, "how long to wait for mysqld startup or shutdown")
+		fs.StringVar(&initDBSQLFile, "init_db_sql_file", initDBSQLFile, "path to .sql file to run after mysql_install_db")
```
GuptaManan100, line 65: Just one comment in hindsight. We should make the help for these commands start with capital case. This is not really related to flags change, but something we want to fix going forward

mattlord: Will do! I thought about it doing it at the time, but then left it as I wasn't sure we had all agreed on that (I have been doing this). 🙂

mattlord: Just FYI, since I was there I also cleaned up the text a bit (for example, `mysql_install_db` is no longer used for init): https://github.com/vitessio/vitess/pull/11376/commits/21f8c2306c06fb65759b6335f67c77899424ff4d

PR #11379, opened 2022-09-28
Push down derived tables under route when possible

### LINKED ISSUE #11093: Bug Report: Predicates incorrectly pushed into derived tables by planner
### Overview of the Issue

When building the logical operators for a query that has derived tables, the Gen4 planner incorrectly tries to push conditions on derived tables as deep down as possible. There are a few query types where that's a valid optimization to do, but there's many more types where it's not, and I believe overall it might be better to skip this optimization and instead make better routing decisions.

### Reproduction Steps

Here's an example query that uses the vschema defined in `go/vt/vtgate/planbuilder/testdata/schema_test.json`

```sql
SELECT music.id FROM (
  SELECT id FROM (
    SELECT id, user_id, genre FROM music WHERE music.user_id IN (5, 6) LIMIT 10
  ) _inner WHERE _inner.genre = 'pop'
) music
```

Here's the plan that is generated for this query:

```json
{
  "QueryType": "SELECT",
  "Original": "SELECT music.id FROM (SELECT id FROM (SELECT id, user_id, genre FROM music WHERE music.user_id IN (5, 6) LIMIT 10) _inner WHERE _inner.genre = 'pop') music ",
  "Instructions": {
    "OperatorType": "SimpleProjection",
    "Columns": [
      0
    ],
    "Inputs": [
      {
        "OperatorType": "SimpleProjection",
        "Columns": [
          0
        ],
        "Inputs": [
          {
            "OperatorType": "Limit",
            "Count": "INT64(10)",
            "Inputs": [
              {
                "OperatorType": "Route",
                "Variant": "IN",
                "Keyspace": {
                  "Name": "user",
                  "Sharded": true
                },
                "FieldQuery": "select id, user_id, genre from music where 1 != 1",
                "Query": "select id, user_id, genre from music where music.user_id in ::__vals and genre = 'pop' limit :__upper_limit",
                "Table": "music",
                "Values": [
                  "(INT64(5), INT64(6))"
                ],
                "Vindex": "user_index"
              }
            ]
          }
        ]
      }
    ]
  },
  "TablesUsed": [
    "user.music"
  ]
}
```

You can see here that the `_inner.genre = 'pop'` condition was pushed down into the `SELECT` statement that creates the innermost derived table. Unfortunately, that modification is _not_ correct - instead of selecting the first 10 rows that match `music.user_id IN (5, 6)` and then filtering down these 10 rows to the ones that also match `_inner.genre = 'pop'`, the result will be the first 10 rows that match `music.user_id IN (5, 6) AND genre = 'pop'`. 😞 

---

I'd propose fixing this by not doing this optimization in the logical operator creation phase, but do it at the physical planning stage / via route merging instead.

If the innermost `SELECT` is a `EqualUnique` route, the route can be pulled up to include the wrapping `SELECT` statements, and the push down of predicates on the wrapping `SELECT` statements can be handled by MySQL itself.

If the innermost `SELECT` is any other route type (e.g. `IN` or `Scatter`), then the route can not be pulled up, and the wrapping `SELECT` statements need to be converted into operations happening on the vtgate layer instead.

### Binary Version

```sh
@arthurschreiber ➜ /workspaces/vitess (main ✗) $ bin/vtgate --version
Version: 15.0.0-SNAPSHOT (Git revision 98582736d6fcc9f3b08fc7263539927d0a83698a branch 'main') built on Thu Aug 25 09:19:15 UTC 2022 by codespace@codespaces-7bc63a using go1.18.5 linux/amd64
@
```

```sh
Not relevant
```

### Log Fragments

_No response_

systay:
## Description
When we are checking if two parts of the query can be merged, either through a join or through a subquery, if we are also dealing with derived tables, things become tricky.

To simplify this logic, we want to separate the decision if we can push a derived table underneath a route or not, from the decision if we can merge two routes into a single one.

This PR pushes down the derived table under the route when possible. 
It is possible, when either the derived table is a single-shard query, or all of these conditions must be met:
* The derived table can't do a LIMIT
* If the derived table contains grouping, it has to be grouping on a unique vindex column
* If no grouping is done, there can't be any aggregations.

We also use this logic to figure out when we can have a derived table on the RHS of a join that is performed on the vtgate side. Since these joins are nested loop joins, to guarantee that the RHS of a join returns the correct values, we check the above rules. If the derived table doesn't live up the conditions listed above, we can't have this input on the RHS. The planner will try to switch sides, and if the other side also can't be on the RHS, it will give up.

A lot of this logic has been extracted from https://github.com/vitessio/vitess/pull/10966 together with @arthurschreiber 

More PRs are coming that will build on this logic.

## Checklist

## Related issues
Fixes #11093
Followed by #10966


go/vt/vtgate/planbuilder/physical/derived.go
```diff
@@ -110,3 +111,46 @@ func (d *Derived) findOutputColumn(name *sqlparser.ColName) (int, error) {
...
+		return false
+	}
+
+	// if we have grouping, we have already checked that it's safe, and don't need to check for aggregations
+	// but if we don't have groupings, we need to check if there are aggregations that will mess with us
+	for _, expr := range sel.SelectExprs {
+		if sqlparser.ContainsAggregation(expr) {
+			return false
+		}
+	}
```
arthurschreiber, line 153: I think we need to extend this to also check the `HAVING` expressions to look for aggregations as well.

```suggestion
	for _, expr := range sel.SelectExprs {
		if sqlparser.ContainsAggregation(expr) {
			return false
		}
	}

	if sel.Having != nil {
		if sqlparser.ContainsAggregation(sel.Having.Expr) {
			return false
		}
	}
```

go/test/endtoend/vtgate/queries/derived/derived_test.go
```diff
@@ -0,0 +1,91 @@
...
+	defer closer()
+
+	mcmp.Exec("insert into music(id, user_id) values(1,1), (2,5), (3,1), (4,2), (5,3), (6,4), (7,5)")
+	mcmp.Exec("insert into user(id, name) values(1,'toto'), (2,'tata'), (3,'titi'), (4,'tete'), (5,'foo')")
+
+	mcmp.Exec("set sql_mode = ''")
+
+	// this is probably flaky? the id returned from the derived table could be any of the ids from user.
+	// works on my machine (TM)
+	mcmp.Exec("select  /*vt+ PLANNER=Gen4 */ * from (select id from user having count(*) >= 1) s")
```
frouioui, line 90: We could insert only one row in the `user` table, the row returned by the derived table will always be the same.

go/test/endtoend/vtgate/queries/derived/main_test.go
```diff
@@ -0,0 +1,94 @@
...
+	"vitess.io/vitess/go/mysql"
+	"vitess.io/vitess/go/test/endtoend/cluster"
+)
+
+var (
+	clusterInstance *cluster.LocalProcessCluster
+	vtParams        mysql.ConnParams
+	mysqlParams     mysql.ConnParams
+	keyspaceName    = "ks_music"
+	cell            = "test_music"
```
frouioui, line 37: nit:
`ks_music` -> `ks_derived`
`test_music` -> `test_derived`

go/test/endtoend/vtgate/queries/derived/main_test.go
```diff
@@ -0,0 +1,94 @@
+/*
+Copyright 2022 The Vitess Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
```
frouioui, line 8: I suggest we move the E2E test located in `go/test/endtoend/vtgate/mysql80/derived` to this package.

go/test/endtoend/vtgate/queries/derived/derived_test.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+func TestDerivedTableWithOrderByLimit(t *testing.T) {
+	mcmp, closer := start(t)
+	defer closer()
+
+	mcmp.Exec("insert into music(id, user_id) values(1,1), (2,5), (3,1), (4,2), (5,3), (6,4), (7,5)")
+	mcmp.Exec("insert into user(id, name) values(1,'toto'), (2,'tata'), (3,'titi'), (4,'tete'), (5,'foo')")
+
+	mcmp.Exec("select /*vt+ PLANNER=Gen4 */ music.id from music join (select id,name from user order by id limit 2) as d on music.user_id = d.id")
```
harshit-gangal, line 55: I think we should assert the results as well in all the tests in this file.

systay: Sure, can do that. The bugs all showed up when comparing w mysql, but it might be interesting to assert on the values as well

Issue #10862, opened 2022-07-26
Switch flag definitions to be on pflag instead of flag in `package go/vt/topo/zk2topo`


ajm188:
Part of https://github.com/vitessio/vitess/issues/10697.

Current flags:

```
$ git grep -E "\bflag\.[A-Z]" -- go/vt/topo/zk2topo/*.go
go/vt/topo/zk2topo/zk_conn.go:  maxConcurrency = flag.Int("topo_zk_max_concurrency", 64, "maximum number of pending requests to send to a Zookeeper server.")
go/vt/topo/zk2topo/zk_conn.go:  baseTimeout = flag.Duration("topo_zk_base_timeout", 30*time.Second, "zk base timeout (see zk.Connect)")
go/vt/topo/zk2topo/zk_conn.go:  certPath = flag.String("topo_zk_tls_cert", "", "the cert to use to connect to the zk topo server, requires topo_zk_tls_key, enables TLS")
go/vt/topo/zk2topo/zk_conn.go:  keyPath  = flag.String("topo_zk_tls_key", "", "the key to use to connect to the zk topo server, enables TLS")
go/vt/topo/zk2topo/zk_conn.go:  caPath   = flag.String("topo_zk_tls_ca", "", "the server ca to use to validate servers when connecting to the zk topo server")
go/vt/topo/zk2topo/zk_conn.go:  authFile = flag.String("topo_zk_auth_file", "", "auth to use when connecting to the zk topo server, file contents should be <scheme>:<auth>, e.g., digest:user:pass")
```

Issue #10860, opened 2022-07-26
Switch flag definitions to be on pflag instead of flag in `package go/vt/topo/k8stopo`


ajm188:
Part of https://github.com/vitessio/vitess/issues/10697.

Current flags:

```
$ git grep -E "\bflag\.[A-Z]" -- go/vt/topo/k8stopo/*.go
go/vt/topo/k8stopo/server.go:   kubeconfigPath = flag.String("topo_k8s_kubeconfig", "", "Path to a valid kubeconfig file. When running as a k8s pod inside the same cluster you wish to use as the topo, you may omit this and the below arguments, and Vitess is capable of auto-discovering the correct values. https://kubernetes.io/docs/tasks/access-application-cluster/access-cluster/#accessing-the-api-from-a-pod")
go/vt/topo/k8stopo/server.go:   configContext = flag.String("topo_k8s_context", "", "The kubeconfig context to use, overrides the 'current-context' from the config")
go/vt/topo/k8stopo/server.go:   configNamespace = flag.String("topo_k8s_namespace", "", "The kubernetes namespace to use for all objects. Default comes from the context or in-cluster config")
go/vt/topo/k8stopo/server_flaky_test.go:        flag.Set("topo_k8s_kubeconfig", testConfigPath)
```

PR #11106, opened 2022-08-26
[cli] [topo/consultopo] Migrate consul flags to pflags

rsajwani:
Signed-off-by: Rameez Sajwani <rameezwazirali@hotmail.com>

## Description
Updates throttler related flag references specified in https://github.com/vitessio/vitess/issues/10858

Here are the commands registered
go list -f '{{ .ImportPath }}| {{ join .Deps " " }} ' ./go/cmd/... | awk -F"|" '$2 ~ "go/vt/topo/consultopo" {print $1}'
vitess.io/vitess/go/cmd/topo2topo
vitess.io/vitess/go/cmd/vtbackup
vitess.io/vitess/go/cmd/vtcombo
vitess.io/vitess/go/cmd/vtctl
vitess.io/vitess/go/cmd/vtctld
vitess.io/vitess/go/cmd/vtgate
vitess.io/vitess/go/cmd/vtgr
vitess.io/vitess/go/cmd/vtorc
vitess.io/vitess/go/cmd/vttablet
vitess.io/vitess/go/cmd/vttestserver
vitess.io/vitess/go/cmd/zk

## Related Issue(s)

Fixes #10858 

## Checklist

-   [ ] "Backport me!" label has been added if this change should be backported
-   [x] Tests were added or are not required
-   [x] Documentation was added or is not required

PR #11401, opened 2022-09-29
[cli] Topo: Migrate zk2topo and k8stopo to pflag

rohit-nayak-ps:
## Description
Migrate zk2topo and k8stopo flags to pflag

## Related Issue(s)
Fixes https://github.com/vitessio/vitess/issues/10860
Fixes https://github.com/vitessio/vitess/issues/10862

## Checklist


go/vt/servenv/servenv.go
```diff
@@ -385,3 +385,9 @@ func init() {
 	// Flags in package logutil are installed for all binaries.
 	OnParse(logutil.RegisterFlags)
 }
+
+func RegisterFlagsForTopoBinaries(registerFlags func(fs *pflag.FlagSet)) {
```
ajm188, line 389: ugh, that's annoying, sorry about that!

ajm188 (commented):
code changes looking good, whenever you're ready to move out of draft

ajm188:
i don't think this has anything to do with flags, it's just environment variables: https://github.com/kubernetes/client-go/blob/e6538dd42b4fe55b6c754e41c66b43133ba41a59/rest/config.go#L516-L519

ajm188:
ahhh, https://github.com/kubernetes/client-go/blob/e6538dd42b4fe55b6c754e41c66b43133ba41a59/tools/clientcmd/client_config.go#L614-L617 let's see where those come from

ajm188:
https://github.com/vitessio/vitess/blob/main/go/vt/topo/k8stopo/server_flaky_test.go#L86-L92 cc @rohit-nayak-ps

rohit-nayak-ps:
> i don't think this has anything to do with flags, it's just environment variables: https://github.com/kubernetes/client-go/blob/e6538dd42b4fe55b6c754e41c66b43133ba41a59/rest/config.go#L516-L519

Those environment variables are accessed from within a pod, which I believe, are passed via a kube config file and that seems to get messed up if we replace `flag` with `pflag` here: https://github.com/vitessio/vitess/blob/main/go/vt/topo/k8stopo/server.go#L50-L59

ajm188:
> Those environment variables are accessed from within a pod, which I believe, are passed via a kube config file and that seems to get messed up if we replace flag with pflag here:

do you know where that config file lives? maybe they're using single dashes?

i don't understand how this couldn't work, unless that pod is starting up via a binary that doesn't use `servenv` for parsing

deepthi:
A few more help text files need to be fixed
```
$ find go/flags/endtoend -iname '*.txt' | xargs basename | cut -d. -f1 | xargs -I{} bash -c "{} --help 2>go/flags/endtoend/{}.txt"
```
and revert any changes to vtctldclient.txt

go/flags/endtoend/vtorc.txt
```diff
@@ -59,15 +59,6 @@ Usage of vtorc:
...
-      --topo_k8s_namespace string                    The kubernetes namespace to use for all objects. Default comes from the context or in-cluster config
-      --topo_zk_auth_file string                     auth to use when connecting to the zk topo server, file contents should be <scheme>:<auth>, e.g., digest:user:pass
-      --topo_zk_base_timeout duration                zk base timeout (see zk.Connect) (default 30s)
-      --topo_zk_max_concurrency int                  maximum number of pending requests to send to a Zookeeper server. (default 64)
-      --topo_zk_tls_ca string                        the server ca to use to validate servers when connecting to the zk topo server
-      --topo_zk_tls_cert string                      the cert to use to connect to the zk topo server, requires topo_zk_tls_key, enables TLS
-      --topo_zk_tls_key string                       the key to use to connect to the zk topo server, enables TLS
```
deepthi, line 70: not sure about the other flags, but I'm pretty sure we want to keep these for vtorc.
cc @GuptaManan100

GuptaManan100: Yes, we do, whatever flags we need to connect to a topo-server. We'll need them.

rohit-nayak-ps: I updated the list of  topo binaries in servenv to add `vtorc`. I had just refactored the list from main into this, so I guess previous PRs too had this issue. 

```
func RegisterFlagsForTopoBinaries(registerFlags func(fs *pflag.FlagSet)) {
	topoBinaries := []string{
		"vtbackup",
		"vtcombo",
		"vtctl",
		"vtctld",
		"vtgate",
		"vtgr",
		"vttablet",
		"vttestserver",
		"zk",
		"vtorc",
	}
	for _, cmd := range topoBinaries {
		OnParseFor(cmd, registerFlags)
	}
}
```

go/vt/topo/k8stopo/server.go
```diff
@@ -47,16 +48,24 @@ import (
...
 )
 
-var (
+var kubeconfigPath, configContext, configNamespace string
+var k8stopoFlags *pflag.FlagSet
+
+func init() {
+	k8stopoFlags = pflag.NewFlagSet("k8stopo", pflag.ExitOnError)
+	registerFlags(k8stopoFlags)
+}
```
GuptaManan100, line 57: I don't understand this part 😕. Shouldn't we register the k8stopo like the other topo services?

rohit-nayak-ps: The other mechanisms didn't work. Maybe there is a better way, but I needed to do this in `go/vt/topo/k8stopo/server_flaky_test.go`  using the above technique  to fix `TestKubernetesTopo`.

```
k8stopoFlags.Set("topo_k8s_kubeconfig", testConfigPath)
```

GuptaManan100: Can't you directly set `kubeconfigPath = testConfigPath`? I think we do that in other test code too.
We use something like 
```go
oldConfigPath := kubeconfigPath
kubeconfigPath = testPath
defer func() {
    kubeconfigPath = oldConfigPath
}()
```

GuptaManan100: Also, don't we need these flags in the binaries that also have the zookeeper flags, like VTOrc etc?

GuptaManan100: I think they will be used if someone uses `k8s` flavour of topo-server, but I am not sure.

ajm188: imo, this code should happen only in the flaky test file (ie. you should do what things like `tabletconntest.SetProtocol` do)

go/flags/endtoend/vtorc.txt
deepthi: `zk` doesn't need to be in this list, does it?

rohit-nayak-ps: That list comes from https://github.com/vitessio/vitess/pull/11106/files

go/vt/topo/k8stopo/server.go
rohit-nayak-ps: @GuptaManan100's suggestion above worked.

go/vt/topo/k8stopo/server.go
```diff
@@ -28,12 +28,15 @@ We follow these conventions within this package:
 package k8stopo
 
 import (
-	"flag"
 	"fmt"
 	"os"
 	"path/filepath"
 	"strings"
 
+	"vitess.io/vitess/go/vt/servenv"
```
rsajwani, line 36: can you move this to down at line 46 with other vitess.io packages

rsajwani (approved):

PR #11403, opened 2022-09-30
gen4 planner: small cleanup

systay:
## Description
When we have built a simple projection on top of a simpleProjection, we can clean these out quite simply.

## Checklist


frouioui (approved):
LGTM

PR #11410, opened 2022-09-30
Enable VTOrc in backup tests

GuptaManan100:
## Description

This PR enables VTOrc for the backup tests that take backup of a `replica`. Unfortunately, we can't enable VTOrc for the tests that take backup from `primary` tablets, since VTOrc will see there is no primary and end up promoting one which interferes with the tests invariants.

## Related Issue(s)

- #6612 

## Checklist


rsajwani (approved):

GuptaManan100:
The reduction in the topo-refresh-time for VTOrc is a nice optimization in making the tests faster. 

Before - 
Screenshot 2022-10-01 at 2 02 49 PM

After - 
Screenshot 2022-10-01 at 1 52 52 PM

This just helps VTOrc setup faster, so it shaves off about 10 seconds from all the tests. So overall it has a lower impact for long-running tests, but is much more useful for shorter tests.

go/test/endtoend/cluster/cluster_util.go
```diff
@@ -94,7 +94,7 @@ func GetPrimaryPosition(t *testing.T, vttablet Vttablet, hostname string) (strin
 // VerifyRowsInTabletForTable verifies the total number of rows in a table.
 // This is used to check that replication has caught up with the changes on primary.
 func VerifyRowsInTabletForTable(t *testing.T, vttablet *Vttablet, ksName string, expectedRows int, tableName string) {
-	timeout := time.Now().Add(10 * time.Second)
+	timeout := time.Now().Add(1 * time.Minute)
```
GuptaManan100, line 97: This timeout has been increased for VTOrc. With the defaults, it takes VTOrc 15 seconds to discover the tablets and then it can start repairing them. This wait was for 10 seconds which turned out to be too small. I have improved VTOrc performance ☝️, but it doesn't hurt to increase the timeout.

PR #11413, opened 2022-09-30
VDiff: Make restarting VReplication workflow more robust

mattlord:
## Description

First we use a new background context for the restart operation. Otherwise the restart can fail due to the parent operation's expired/canceled context:
```
E0929 14:06:08.618848   85788 vdiff.go:247] Could not restart workflow commerce2customer: rpc error: code = Canceled desc = context canceled, please restart it manually
E0929 14:06:08.619983   85788 vtctl.go:2630] vdiff returning with error: rpc error: code = Canceled desc = context canceled
```

We also add retries if we encounter an ephemeral error — such as [deadlock detected](https://github.com/vitessio/vitess/pull/10621) — when updating the workflow's record in the `_vt` sidecar database.

I don't think we need a new test here as the context timeout portion (which led to the initial PR) was found with a static code analysis and manually tested with an artificial change (added a 90s sleep in VDiff and then CTRL-C'd the client during that sleep) and the retries are a generic known issue?

## Related Issue(s)

  - Related to: https://github.com/vitessio/vitess/pull/10621
  - Related to: https://github.com/vitessio/vitess/pull/10275
  - Related to: https://github.com/vitessio/vitess/pull/10051

## Checklist

-   [x] Tests were are not required


rohit-nayak-ps (approved):
lgtm

PR #11415, opened 2022-10-01
For partial MoveTables, setup reverse shard routing rules on workflow creation

mattlord:
## Description

For normal `MoveTables` operations — where you're moving tables from one keyspace to another, i.e. the move is at the keyspace level — global routing rules are put in place when the workflow is created so that any attempted access of those tables on the target keyspace are routed to the source keyspace. (In this scenario shard targeting is how you would explicitly get around this routing.)

For partial `MoveTables` operations, you're moving the tables from one keyspace to another **one shard at a time** — so the move is **per shard** rather than per keyspace. In this scenario you want all traffic to go to one side or the other for a given shard, *including* shard targeted queries. This then offers the same kind of "global" (scoped to a shard) routing at the shard level that you would expect when doing keyspace level `MoveTables`. This PR implements that functionality.

> **Note**
> We have NO documentation on partial MoveTables today (not even the relevant `vttablet` and `vtgate` flags). Should I create a new page in the docs here, or continue to defer that part? 
> Answer: YES. See: https://github.com/vitessio/website/pull/1160

## Checklist

-   [x] Tests were updated
-   [x] Documentation: https://github.com/vitessio/website/pull/1160


rohit-nayak-ps:
> I think we should document it. @rohit-nayak-ps thoughts?

Yes, that is the plan. We will have a link from `MoveTables` to a separate page highlighting the use cases, shard routing rules and their effect, limitations (need to disable schema tracking, for example) and the fact that it is experimental. Should be ready this week.

rohit-nayak-ps (approved):
lgtm

mattlord:
Please note that this will be merged just after the v15 RC1 release so that:
1. We don't interfere with that timely release
2. We have a feature complete shard level migration feature (see docs here: https://github.com/vitessio/website/pull/1160)

PR #11417, opened 2022-10-01
Fix vtbackup binary by adding the flags it needs that we missed before

GuptaManan100:
## Description

This PR adds the flags that `vtbackup` needs but we missed to add them during the flag work. These were found because of VTop failures while running the backup end to end tests.

## Related Issue(s)

- #10697 

## Checklist

### DIFF
--- a/go/vt/mysqlctl/backupengine.go
+++ b/go/vt/mysqlctl/backupengine.go
@@ -112,7 +112,7 @@ type BackupRestoreEngine interface {
 var BackupRestoreEngineMap = make(map[string]BackupRestoreEngine)
 
 func init() {
-	for _, cmd := range []string{"mysqlctl", "mysqlctld", "vtcombo", "vttablet", "vttestserver", "vtctld", "vtctldclient", "vtexplain"} {
+	for _, cmd := range []string{"mysqlctl", "mysqlctld", "vtcombo", "vttablet", "vttestserver", "vtctld", "vtctldclient", "vtexplain", "vtbackup"} {
 		servenv.OnParseFor(cmd, registerBackupEngineFlags)
 	}
 }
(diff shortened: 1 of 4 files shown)


go/vt/mysqlctl/backupengine.go
```diff
@@ -112,7 +112,7 @@ type BackupRestoreEngine interface {
 var BackupRestoreEngineMap = make(map[string]BackupRestoreEngine)
 
 func init() {
-	for _, cmd := range []string{"mysqlctl", "mysqlctld", "vtcombo", "vttablet", "vttestserver", "vtctld", "vtctldclient", "vtexplain"} {
+	for _, cmd := range []string{"mysqlctl", "mysqlctld", "vtcombo", "vttablet", "vttestserver", "vtctld", "vtctldclient", "vtexplain", "vtbackup"} {
```
mattlord, line 115: I would think vtbackup, vttablet, and maybe vtctld need backup flags. We can pare this down later.

mattlord (approved):

go/vt/mysqlctl/backupengine.go
GuptaManan100: Honestly, I have no idea. Maybe @deepthi or @rsajwani know 🤷‍♂️

GuptaManan100: I am going ahead and merging this PR, because I want the latest lite image to be built so that I can continue testing https://github.com/planetscale/vitess-operator/pull/324. We can add these flags to vttablet and vtctld in a follow up PR if needed.

deepthi: We have to do the opposite, i.e, remove the other binaries: mysqlctl, vtexplain etc.
Not sure whether vtcombo and vttestserver are supposed to support backups.

rsajwani (commented):
LGTM

go/vt/mysqlctl/backupengine.go
rsajwani: > vtbackup, vttablet, and m

Yes vttablet , vtctld will also need backup flags.

Issue #11226, opened 2022-09-15
Bug Report: query error Table doesn't exist during DDL


L3o-pold:
### Overview of the Issue

During online migration, while adding an index to a table, vttablet reported that the table was missing. 

### Reproduction Steps

1. deploy a shared keyspace (commerce)
2. deploy the following schema

```sql
create table users
(
    id char(36),
    team_id char(36),
    primary key (id)
) ENGINE=InnoDB;

create table users_keyspace_id
(
    `id`          char(36),
    `keyspace_id` varbinary(128) DEFAULT NULL,
    PRIMARY KEY (`id`)
) ENGINE=InnoDB;
```

3. deploy the following vschema

```json
{
    "sharded": true,
    "vindexes": {
        "unicode_loose_md5": {
            "type": "unicode_loose_md5"
        },
        "users_keyspace_id": {
            "type": "consistent_lookup_unique",
            "params": {
                "batch_lookup": "true",
                "from": "id",
                "table": "commerce.users_keyspace_id",
                "to": "keyspace_id"
            },
            "owner": "users"
        }
    },
    "tables": {
        "users": {
            "column_vindexes": [
                {
                    "column": "team_id",
                    "name": "unicode_loose_md5"
                },
                {
                    "column": "id",
                    "name": "users_keyspace_id"
                }
            ]
        },
        "users_keyspace_id": {
            "column_vindexes": [
                {
                    "column": "id",
                    "name": "unicode_loose_md5"
                }
            ]
        }
    }
}
```

4. Populate some records

```shell
while [ true ];
do
    mysql -h localhost -u user --port=15306 commerce -e 'INSERT INTO `users` (`id`, `team_id`) VALUES ("'$(uuidgen)'", "'$(uuidgen)'");'
done;
```

6. Run select queries during an online DDL

```shell
set -e

while [ true ];
do
    mysql -h localhost -u user --port=15306 commerce -e "SELECT id, team_id from users LIMIT 1;"
done;
```

7. while select queries are running run an online DDL on the table

```sql
SET @@ddl_strategy='vitess';
ALTER TABLE `users` ADD INDEX `team_id` (`team_id`);
```

8. Check the terminal where you are running select queries and see the error

```shell
+--------------------------------------+--------------------------------------+
| id                                   | team_id                              |
+--------------------------------------+--------------------------------------+
[...]

PR #11418, opened 2022-10-02
WIP: vitess Online DDL atomic cut-over

shlomi-noach:
## Description

This is work in progress for an atomic `RENAME` (not two-step gaping hole) in the cut-over process of a vitess/vreplication Online DDL migration.

Curently, we run a two-step rename, when the original table is renamed away, and then a 2nd rename moves the vreplicatoin table in its place. This is protected under a buffering rule on the primary, but replicas see two distinct renames and there is a point in time on the replica, where the table does not exist. Of course this means queries are failing.
There is also a scenario depicted in https://github.com/vitessio/vitess/issues/11226, where even queries against the primary may fail.

This PR attempts to introduce an algorithm that is similar to the `gh-ost` cut-over. It is simplified, because the buffering rule gives us some safety net.

Still work in progress as I/m seeing some flunctuations.

## Related Issue(s)

#6926

## Checklist


shlomi-noach:
So far no good results. Hammering the branch against `onlineddl_vrepl_stress`, with increased number of tests, there's always a failing test. I've tried many variations thus far.
I'm going to actually step back and validate that `main` itself passed the increased volume of tests.

shlomi-noach:
Superseded by https://github.com/vitessio/vitess/pull/11460, which takes a stricter and safer approach.

PR #11424, opened 2022-10-03
VTop: Adds a function to get the flag set for a given command

GuptaManan100:
## Description

With the recent flag changes, we discovered that VTop binary broke. The Vitess-operator requires the backup flags like `bakcup_storage_implementation`, `ceph_backup_storage_config` among others. Previously, when we used the `flag` package these got implicitly imported when we imported the packages that defined these flags. Now that we have moved to the `pflag` package, we register the flags only for the binaries that require it. This leads to the flags not being registered in the vitess-operator binary. 

One possible solution would have been to add the `vitess-operator` binary to the list of the binaries where we register certain flags, but that doesn't seem the best approach because it adds a cross-repository code-base dependency. Instead, an alternate fix is proposed in this PR.

This PR refactors some part of the `servenv.ParseFlags` function to return the flagSet of the given command without parsing it. On the VTop side we can ask for the flag set of the `vtbackup` binary and add the flags that we require from it to the VTop binary.

## Related Issue(s)

- https://github.com/vitessio/vitess/issues/10697

## Checklist

### DIFF
--- a/go/vt/servenv/servenv.go
+++ b/go/vt/servenv/servenv.go
@@ -305,10 +305,7 @@ func getFlagHooksFor(cmd string) (hooks []func(fs *pflag.FlagSet)) {
 // ParseFlags initializes flags and handles the common case when no positional
 // arguments are expected.
 func ParseFlags(cmd string) {
-	fs := pflag.NewFlagSet(cmd, pflag.ExitOnError)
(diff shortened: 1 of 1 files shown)


ajm188 (approved):
Lgtm. On the off chance you need to push more commits, want to also punctuate those function comments? :heart:

PR #11429, opened 2022-10-04
feat: added query timeout to vtgate default and per session

harshit-gangal:
## Description

This PR adds query timeout to vtgate by using `query-timeout` flag or setting it via system variable on per session bases using `set query_timeout = <>` . It takes the value in miliseconds.

## Checklist


GuptaManan100 (approved):
LGTM!

PR #11433, opened 2022-10-04
Handle aliasing of collation names

dbussink:
With MySQL 8.0.30 and later, utf8mb3 is always reported as the charset in output for `SHOW CREATE TABLE` which is what `schemadiff` uses. We already today normalize all `charset` output to use the `utf8mb3` name to avoid any ambiguity in what is intended.

We didn't do this though for collations. Today if `schemadiff` is fed schemas generated both with MySQL 8.0.30 and older versions to compare, it would indicate there's a difference when there is none.

The change here always normalizes to use the more explicit `utf8mb3_` names for the collation if it can be found, based on the charset aliases configured. This ensures that comparisons between such schemas don't see accidental or stray diffs that are not really changes.

## Related Issue(s)

This is part of the #10203 schemadiff work.

## Checklist

### DIFF
--- a/go/mysql/collations/env.go
+++ b/go/mysql/collations/env.go
@@ -232,6 +232,33 @@ func (env *Environment) CharsetAlias(charset string) (alias string, ok bool) {
 	return
 }
 
+// CollationAlias returns the internal collaction name for the given charset.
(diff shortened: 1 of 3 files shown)


GuptaManan100 (approved):
Looks good to me!

PR #11434, opened 2022-10-04
Audit of flags between 15.0 and 14.0

rohit-nayak-ps:
There has been a major refactor of flags in 15.0. Since flags were (mostly) global earlier we ended up with a lot of unrelated flags being part of binaries. The refactor should result in binaries contain only (and all) the flags they use.

However it is possible that:
* **we have deleted required flags by mistake**. This is possible because we now have variables defined for flags, (using `pflags.StringVar` instead of `pflags.String`, for example) and we explicitly register the flags for specified binaries. In this case there won't be a compilation error if a binary uses this variable: it will use the default value and the actual issue will be noticed only if a user tries to override that flag and gets "flag not defined for this binary" error. 
* **there are still extraneous flags**, either due to them being registered unnecessarily for a binary, or, if we need to make the flag grouping more granular.
* **flags not defined for dependencies**:  for example, logs, tests etc.

Since tests all run these issues should be minimal, but it is possible that rarely used flags have not all been tested for in our test suite.

This PR is intended to ensure that every developer (or at least one per module) has eyeballed these changes to see if anything obvious stands out. 

### Binaries
- mysqlctl
- mysqlctld
- vtaclcheck
- vtbackup
- vtctlclient
- vtctldclient
- vtctld
- vtexplain
- vtgate
- vtgr
- vtorc
- vttablet
- vttestserver
- vttlstest
- zkctld
- zkctl
- zk

### Auditors

Use this diff for auditing: https://github.com/rohit-nayak-ps/vreplication-docs/pull/2/files?diff=split&w=1. 
The 15.0 list is as of the RC1 cutoff and will be (manually) updated if there are relevant PRs that affect flags.  Vitess Commit sha of diff: 0c257dd138186b311e8f9716dfed73b587b61c5b

#### Query Serving 
- [x] @systay 
- [x] @harshit-gangal 
- [x] @frouioui 
- [x] @GuptaManan100 

#### Cluster Management/Backup/VTOrc
- [x] @deepthi 
- [x] @rsajwani 
- [x] @GuptaManan100 

#### VReplication
- [x] @mattlord 
- [x] @rohit-nayak-ps 

#### Online DDL (also throttler)
- [x] @shlomi-noach 

#### VTAdmin
- [x] @ajm188  

#### Misc (logs/test/stats)
- [x] @ajm188 
- [x] @deepthi 
- [x] @rohit-nayak-ps


ajm188:
@rohit-nayak-ps I checked off the vtadmin box for completeness, but there's no `vtadmin.txt` in that diff (does that mean no change, or that it didn't get included for some reason?)

rohit-nayak-ps:
> @rohit-nayak-ps I checked off the vtadmin box for completeness, but there's no `vtadmin.txt` in that diff (does that mean no change, or that it didn't get included for some reason?)

Hmm, I missed it since it wasn't in the help output of 14.0 and didn't add one for vtadmin it when I generated the help for additional files. 

I just added vtadmin.txt for 15.0 to that pr for completeness, but it doesn't have a 14.0 for comparison, so it will show up as a new file :(

deepthi:
I believe this is now complete.

PR #11436, opened 2022-10-04
Update flag parsing for tests to work with "go tool test2json", used by GoLand to run unit tests

rohit-nayak-ps:
## Description

The flag parsing logic in `go/internal/flag/flag.go` to handle `go test` flags using `ParseFlagsForTest()`, worked on the CLI with `go test`, but not from within GoLand. The latter uses `go tool test2json`.

I am not entirely sure why this change works, but including the parameter passed to `-run` in the test arguments works both with the CLI and GoLand. 

## Checklist

### DIFF
--- a/go/internal/flag/flag.go
+++ b/go/internal/flag/flag.go
@@ -105,14 +105,20 @@ func Usage() {
 // filterTestFlags returns two slices: the second one has just the flags for `go test` and the first one contains
 // the rest of the flags.
 const goTestFlagSuffix = "-test"
+const goTestRunFlag = "-test.run"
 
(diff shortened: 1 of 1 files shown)


GuptaManan100 (approved):
@rohit-nayak-ps I don't think this is a PR we need to rush. Is it okay to merge this after the code-freeze has been disabled on this branch?

rohit-nayak-ps:
> @rohit-nayak-ps I don't think this is a PR we need to rush. Is it okay to merge this after the code-freeze has been disabled on this branch?

Not a priority

Issue #11441, opened 2022-10-05
Bug Report: VTOrc does not let go of the shards it is holding during shutdown


GuptaManan100:
### Overview of the Issue

This bug was found while investigating flakiness in the VTOrc test suite - 
```
2022-10-03T08:01:40.0637090Z I1003 08:01:40.062775    1659 locks.go:386] Locking shard ks/0 for action resetShardPrimary-vtorc-endtoend-test
2022-10-03T08:02:10.0657474Z     utils.go:271: 
2022-10-03T08:02:10.0660578Z         	Error Trace:	utils.go:271
2022-10-03T08:02:10.0661616Z         	            				primary_failure_test.go:482
2022-10-03T08:02:10.0662619Z         	Error:      	Received unexpected error:
2022-10-03T08:02:10.0663780Z         	            	deadline exceeded: /vitess/global/keyspaces/ks/shards/0/locks
2022-10-03T08:02:10.0665704Z         	Test:       	TestDownPrimaryPromotionRuleWithLagCrossCenter
```

These are the VTOrc logs -
```
I1003 08:01:39.943636   14094 locks.go:386] Locking shard ks/0 for action Orc Recovery
I1003 08:01:39.945379   14094 tablet_discovery.go:183] refresh of tablet records of shard - ks/0
W1003 08:01:39.948356   14094 component.go:41] [core] grpc: addrConn.createTransport failed to connect to {localhost:17217 localhost:17217 <nil> <nil> 0 <nil>}. Err: connection error: desc = "transport: Error while dialing dial tcp 127.0.0.1:17217: connect: connection refused"
I1003 08:01:40.001062   14094 orchestrator.go:123] Received SIGTERM. Starting shutdown
I1003 08:01:40.001099   14094 audit_dao.go:108] auditType:shutdown instance::0 cluster: message:Triggered via SIGTERM
I1003 08:01:40.001114   14094 run.go:61] Entering lameduck mode for at least 50ms
I1003 08:01:40.001132   14094 run.go:62] Firing asynchronous OnTerm hooks
I1003 08:01:40.001140   14094 servenv.go:206] Firing OnTermSync hooks and waiting up to 10s for them
I1003 08:01:40.001153   14094 servenv.go:219] OnTermSync hooks finished
I1003 08:01:40.002738   14094 run.go:67] Sleeping an extra 48.377284ms after OnTermSync to finish lameduck period
I1003 08:01:40.052061   14094 run.go:71] Shutting down gracefully
[...]

PR #11442, opened 2022-10-05
Fix VTOrc holding locks after shutdown

GuptaManan100:
## Description

This PR fixes the issue described in #11441. On further investigation it was found that this problem was introduced when we started using the `servenv` package in VTOrc.

The proposed fix is to run the code that wait for the shards to be let go as part of the synchronous on termination hooks.
Unit tests validating the wait function have also been added.

## Related Issue(s)

- Fixes https://github.com/vitessio/vitess/issues/11441

## Checklist

### DIFF
--- a/go/vt/vtorc/logic/orchestrator.go
+++ b/go/vt/vtorc/logic/orchestrator.go
@@ -105,7 +105,7 @@ func instancePollSecondsDuration() time.Duration {
 	return time.Duration(config.Config.InstancePollSeconds) * time.Second
 }
 
-// acceptSighupSignal registers for OS signals
+// acceptSighupSignal registers for SIGHUP signal from the OS to reload the configuration files.
 func acceptSighupSignal() {
 	c := make(chan os.Signal, 1)
 


deepthi:
I think this is good practice anyway, but I do wonder if we have a bug in the lock code where we are not checking for a canceled context.

go/vt/vtorc/logic/orchestrator.go
```diff
@@ -105,49 +105,52 @@ func instancePollSecondsDuration() time.Duration {
 	return time.Duration(config.Config.InstancePollSeconds) * time.Second
 }
 
-// acceptSignals registers for OS signals
-func acceptSignals() {
+// acceptSighupSignal registers for OS signals
```
deepthi, line 108: nit: comment is not accurate

deepthi (approved):
IIUC, we already had the code for this, we just needed to register it properly with servenv, correct?

go/vt/vtorc/logic/orchestrator.go
GuptaManan100: Fixed.

GuptaManan100:
@deepthi Yes, we had code for this already. The thing is that when we started using `servenv`, it also came with its own handling of SIGTERM. So this code and servenv code would both see the SIGTERM, this code would block until all the locks were let go, but in the meantime servenv would finish, which would cause the main go routine to finish and the program would stop killing all the other go-routines including this one.

Now, we register this function as a on-termination hook to be run, so SIGTERM is only caught by servenv, but as part of the closing of the servenv, this function is called and this blocks until all the locks are released.

PR #11447, opened 2022-10-05
Backport: For partial MoveTables, setup reverse shard routing rules on workflow creation

mattlord:
## Description

For normal `MoveTables` operations — where you're moving tables from one keyspace to another, i.e. the move is at the keyspace level — global routing rules are put in place when the workflow is created so that any attempted access of those tables on the target keyspace are routed to the source keyspace. (In this scenario shard targeting is how you would explicitly get around this routing.)

For partial `MoveTables` operations, you're moving the tables from one keyspace to another **one shard at a time** — so the move is **per shard** rather than per keyspace. In this scenario you want all traffic to go to one side or the other for a given shard, *including* shard targeted queries. This then offers the same kind of "global" (scoped to a shard) routing at the shard level that you would expect when doing keyspace level `MoveTables`. This PR implements that functionality.

> **Note**
> We are backporting this to v15 *after* RC because it makes the new shard level migration feature complete. [See the new docs](https://vitess.io/docs/15.0/reference/vreplication/shardlevelmigrations/) for more info.

## Related Issue(s)
  - Backports: https://github.com/vitessio/vitess/pull/11415
  
## Checklist

-   [x] Tests were updated
-   [x] Documentation: https://github.com/vitessio/website/pull/1160

### DIFF
--- a/go/test/endtoend/vreplication/helper_test.go
+++ b/go/test/endtoend/vreplication/helper_test.go
@@ -23,6 +23,7 @@ import (
 	"net/http"
 	"os/exec"
 	"regexp"
+	"sort"
 	"strconv"
 	"strings"
 	"testing"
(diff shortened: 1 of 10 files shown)


rohit-nayak-ps (approved):
lgtm

Issue #11385, opened 2022-09-28
updates to _vt.copy_state contribute to replica lag


maxenglander:
At PlanetScale we use MoveTables to import customer data from external databases into the PlanetScale environment. Using the `main` code, during a MoveTables operation a replica might lag behind the primary, typically by a few hundred seconds, and by much more when we use [newer experimental work](https://github.com/vitessio/vitess/pull/10828). In extreme cases, we find ourselves having to rebuild replicas from a fresh backup.

We've tried increasing replica CPU/memory/disk  resources, relaxing IO constraints (e.g. `sync_binlog=0`) and also setting `innodb_io_write_threads` > the default, without success.

We think that one bottleneck which may be contributing to replica lag is that a single `_vt.copy_state` row is used by all transactions during the Copy phase for a given table, which prevents replicas from applying binlog events in parallel. To elaborate on this a bit more:

  1. During the Copy phase, for each table rows are read from the source and inserted in batches on the target primary.
  1. For each batch the Copy phase:
      1. Starts a transaction.
      1. Inserts the batch.
      1. Updates the `_vt.copy_state` rows for the current table.
      1. Commits the transaction. 
  1. Target replicas apply these these changes as they appear in the binlog.
  1. Parallelizing IO threads on replicas doesn't help, because of transaction dependencies. Because each transaction updates the same `_vt.copy_state` row, InnoDB ends up partially or completely serializing transactions.

PR #11451, opened 2022-10-06
Use unique rows in copy_state to support parallel replication

mattlord:
## Description

During the VReplication copy table phase, the `lastpk` value copied is stored for the table as part of the larger MySQL transaction used to apply the vstream packet so that we make progress atomically. Because we used a single row in the `_vt.copy_state` table for the table in the workflow, however, this meant that each transaction in MySQL replication on the target keyspace had overlapping write sets and could not be executed in parallel. This caused MySQL replication on the target shards to be single threaded and could lead to a very large MySQL replication lag when copying very large tables — even to the point where the replica had to be re-provisioned from backup because the primary had purged needed binary logs.

This PR causes us to use a _unique_ row in the `_vt.copy_state` table when applying a vstream packet for a table in the workflow, which should then allow for optimum parallelization on the target shard replicas. It also adds garbage collection for the older `copy_state` rows that are no longer needed, and it adds work to ensure that the `copy_state` table stays optimized over time and this new behavior does not cause any long-term issues like auto-inc exhaustion, very large size on disk (all delete marked rows), etc.

## Manual Testing

I confirmed via local testing with 10GiB+ tables that the `copy_state` garbage collection is working as expected on the target primaries, and that we're able to utilize MySQL parallel replication on the target replicas (requires `slave_parallel_type=LOGICAL_CLOCK` and `binlog_transaction_dependency_tracking=WRITESET` on the target replicas):
```
mysql> select WORKER_ID, LAST_APPLIED_TRANSACTION_END_APPLY_TIMESTAMP from performance_schema.replication_applier_status_by_worker;
+-----------+----------------------------------------------+
| WORKER_ID | LAST_APPLIED_TRANSACTION_END_APPLY_TIMESTAMP |
+-----------+----------------------------------------------+
|         1 | 2022-10-08 17:04:06.761868                   |
|         2 | 2022-10-08 17:04:06.469547                   |
|         3 | 2022-10-08 17:04:06.475058                   |
|         4 | 2022-10-08 17:04:06.475054                   |
|         5 | 2022-10-08 17:04:06.475054                   |
|         6 | 2022-10-08 17:04:06.476267                   |
|         7 | 2022-10-08 17:03:31.860522                   |
|         8 | 0000-00-00 00:00:00.000000                   |
+-----------+----------------------------------------------+
8 rows in set (0.00 sec)
```

Also confirmed everything was good with VDiff2:
```
$ vtctlclient VDiff -- --v2 customer.commerce2customer --format=json show last
{
	"Keyspace": "customer",
	"State": "completed",
	"UUID": "eb4e263e-474e-11ed-96c4-920702940ee0",
	"RowsCompared": 20568999,
	"HasMismatch": false,
	"Shards": "0",
	"StartedAt": "2022-10-08 21:19:39",
	"CompletedAt": "2022-10-08 21:20:31"
}
```

Lastly, I confirmed that after `Cancel`ing the `MoveTables` workflow that the `_vt.copy_state` table was in an optimal state — meaning no `data_free` (delete marked rows were cleaned up on disk) and the `auto_increment` value is back at 1:
```
mysql> show table status like "copy_state"\G
           Name: copy_state
         Engine: InnoDB
        Version: 10
     Row_format: Dynamic
           Rows: 0
 Avg_row_length: 0
    Data_length: 16384
Max_data_length: 0
   Index_length: 16384
      Data_free: 0
 Auto_increment: 1
    Create_time: 2022-10-08 17:31:23
    Update_time: NULL
     Check_time: NULL
      Collation: utf8mb4_0900_ai_ci
       Checksum: NULL
 Create_options:
        Comment:
1 row in set (0.00 sec)
```

## Related Issue(s)

  - Fixes: https://github.com/vitessio/vitess/issues/11385

## Checklist

-   [x] Tests were added and updated


arthurschreiber:
We've encountered this issue before as well (/cc @drogart). This is a great change! ❤️

go/vt/vttablet/tabletmanager/vdiff/workflow_differ.go
```diff
@@ -293,11 +293,11 @@ func (wd *workflowDiffer) initVDiffTables(dbClient binlogplayer.DBClient) error
 		}
 	}
 	query := fmt.Sprintf(sqlGetAllTableRows, encodeString(wd.ct.vde.dbName), tableIn.String())
-	qr, err := dbClient.ExecuteFetch(query, -1)
+	isqr, err := dbClient.ExecuteFetch(query, -1)
```
rohit-nayak-ps, line 296: did you intend to rename this to `isqr`?

go/vt/wrangler/workflow.go
```diff
@@ -715,4 +715,31 @@ func (wr *Wrangler) deleteWorkflowVDiffData(ctx context.Context, tablet *topodat
 	}
 }
 
+// optimizeCopyStateTable rebuilds the copy_state table to ensure the on-disk
+// structures are minimal and optimized and resets the auto-inc value for
+// subsequent inserts.
+// This helps to ensure that the size, storage, and performance related factors
+// for the table remain optimal over time and that we don't ever exhaust the
+// available auto-inc values for the table.
+func (wr *Wrangler) optimizeCopyStateTable(ctx context.Context, tablet *topodatapb.Tablet) {
```
rohit-nayak-ps, line 724: Is it worth running this in a goroutine, since `optimize table` might not get prioritized during heavy load?

rohit-nayak-ps (commented):
Looks great! 

Left a couple of comments. Either this or the parallel vcopier one will need to fix conflicts depending on which is merged earlier! Any preference?

Also you need to rebase this since one required test seems to have been added recently.

go/vt/vttablet/tabletmanager/vdiff/workflow_differ.go
mattlord: Yeah, I realized that I was using the same variable name there and inside the for loop — which is fine, due to locality rules — and renamed this outer one for improved readability.

go/vt/wrangler/workflow.go
mattlord: Yeah, that's a good idea. I've done that here: https://github.com/vitessio/vitess/pull/11451/commits/cf259c37b0580bbd122725e0ce706a67647be4f1

rohit-nayak-ps (approved):
lgtm

PR #11461, opened 2022-10-11
Fix #11455 - skip vindex operations for `DELETE` statements against unsharded tables

### LINKED ISSUE #11455: Bug Report: `DELETE FROM <sequence table>` fails with `does not have a primary vindex`
### Overview of the Issue

Executing a `DELETE FROM` query that targets a sequence table stored in an unsharded keyspace, the query fails with:

```
table '<sequence table>' does not have a primary vindex (errno 1173) (sqlstate 42000) during query: DELETE FROM <sequence table>
```

This is unexpected, as tables inside unsharded keyspaces don't require vindexes. Other DML operations (e.g. `INSERT INTO`) are not affected.

### Reproduction Steps

Using the SQL and VSchemas defined in: `go/test/endtoend/vtgate/sequence/seq_test.go`

Execute:

```sql
DELETE FROM sequence_test_seq
```

### Binary Version

```sh
Version: 15.0.0-SNAPSHOT (Git revision b3eab569a4da969eac1e7f19ae7feac087672fb9 branch 'main') built on Fri Oct  7 14:34:49 UTC 2022 by codespace@codespaces-7bc63a using go1.18.5 linux/amd64
```

```sh
not relevant
```

### Log Fragments

_No response_

arthurschreiber:
## Description

Executing a `DELETE FROM` query against a sequence table would fail with:

```
table '<sequence table>' does not have a primary vindex (errno 1173) (sqlstate 42000) during query: DELETE FROM <sequence table>
```

This PR fixes that error by skipping all vindex routing logic for unsharded queries.

## Related Issue(s)

Fixes https://github.com/vitessio/vitess/issues/11455.

## Checklist

### DIFF
--- a/go/vt/vtgate/planbuilder/physical/route_planning.go
+++ b/go/vt/vtgate/planbuilder/physical/route_planning.go
@@ -226,7 +226,7 @@ func createPhysicalOperatorFromDelete(ctx *plancontext.PlanningContext, op *abst
 		return nil, err
 	}
 
-	if opCode == engine.Unsharded {
+	if !vindexTable.Keyspace.Sharded {
 		return &Route{
 			Source: &Delete{
 				QTable: op.Table,
--- a/go/vt/vtgate/planbuilder/physical/route_planning.go
+++ b/go/vt/vtgate/planbuilder/physical/route_planning.go
@@ -226,6 +226,19 @@ func createPhysicalOperatorFromDelete(ctx *plancontext.PlanningContext, op *abst
 		return nil, err
 	}
 
+	if opCode == engine.Unsharded {
+		return &Route{
+			Source: &Delete{
+				QTable: op.Table,
+				VTable: vindexTable,
+				AST:    op.AST,
+			},
+			RouteOpCode:       opCode,
+			Keyspace:          vindexTable.Keyspace,
+			TargetDestination: dest,
+		}, nil
+	}
+
 	primaryVindex, vindexAndPredicates, err := getVindexInformation(op.TableID(), op.Table.Predicates, vindexTable)
 	if err != nil {
 		return nil, err
--- a/go/test/endtoend/vtgate/sequence/seq_test.go
+++ b/go/test/endtoend/vtgate/sequence/seq_test.go
@@ -47,9 +47,9 @@ var (
 	)Engine=InnoDB;
 
 	create table sequence_test_seq (
-		id int default 0, 
-		next_id bigint default null, 
-		cache bigint default null, 
+		id int default 0,
+		next_id bigint default null,
+		cache bigint default null,
 		primary key(id)
 	) comment 'vitess_sequence' Engine=InnoDB;
 
@@ -60,13 +60,13 @@ INSERT INTO id_seq (id, next_id, cache) values (0, 1, 1000);
 	`
 
 	unshardedVSchema = `
-		{	
+		{
 			"sharded":false,
 			"vindexes": {
 				"hash_index": {
 					"type": "hash"
 				}
-			},	
+			},
 			"tables": {
 				"sequence_test":{
 					"auto_increment":{
@@ -147,7 +147,7 @@ CREATE TABLE allDefaults (
 				"column": "id",
 				"sequence": "id_seq"
 			  }
-			},			
+			},
 			"allDefaults": {
 			  "columnVindexes": [
 				{
@@ -263,6 +263,12 @@ func TestSeq(t *testing.T) {
 	want := "Duplicate entry"
 	if err == nil || !strings.Contains(err.Error(), want) {
 		t.Errorf("wrong insert: %v, must contain %s", err, want)
+	}
+
+	utils.Exec(t, conn, "DELETE FROM sequence_test_seq")
+	qr = utils.Exec(t, conn, "select * from sequence_test_seq")
+	if got, want := fmt.Sprintf("%v", qr.Rows), `[]`; got != want {
+		t.Errorf("select:\n%v want\n%v", got, want)
 	}
 }
 
--- a/go/vt/vtgate/planbuilder/testdata/dml_cases.json
+++ b/go/vt/vtgate/planbuilder/testdata/dml_cases.json
@@ -264,6 +264,29 @@
     }
   },
   {
+    "comment": "delete from sequence",
+    "query": "DELETE FROM seq",
+    "plan": {
+      "QueryType": "DELETE",
+      "Original": "DELETE FROM seq",
+      "Instructions": {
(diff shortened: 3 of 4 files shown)


arthurschreiber:
The reason why we even end up checking for vindexes on sequence tables is because `SingleUnshardedKeyspace` returns `nil` if any of the tables involved in the query is either a `reference` or `sequence` table:

https://github.com/vitessio/vitess/blob/893cc9d7b6f80eab3b103ec327330ad501e16e4f/go/vt/vtgate/semantics/semantic_state.go#L375-L383

I tried removing the `vindexTable.Type != ""` condition, but that causes a lot of tests to break as queries like `select next 2 values from seq` do no longer get treated as `Next` variant routes.

harshit-gangal:
Could you provide the use case for such a statement on the sequence table?

arthurschreiber:
@harshit-gangal No specific use case in particular, just something I noticed that is broken. From what I gather, sequence tables are supposed to work just like "regular" tables (`INSERT`, `SELECT`, `DELETE`), with the exception that they support the `SELECT NEXT VALUE` syntax, right?

frouioui:
> @harshit-gangal No specific use case in particular, just something I noticed that is broken. From what I gather, sequence tables are supposed to work just like "regular" tables (`INSERT`, `SELECT`, `DELETE`), with the exception that they support the `SELECT NEXT VALUE` syntax, right?

That's also my understanding. Although, from what I understand, it is pretty uncommon to send such queries to sequence tables.

frouioui:
@arthurschreiber you should rebase your PR on top of latest `main`, without this the `vtgate_queries` and `vtgate_partial_keyspace` are going to fail.

go/vt/vtgate/planbuilder/physical/route_planning.go
```diff
@@ -226,33 +226,47 @@ func createPhysicalOperatorFromDelete(ctx *plancontext.PlanningContext, op *abst
...
+		r = &Route{
+			Source: &Delete{
+				QTable: op.Table,
+				VTable: vindexTable,
+				AST:    op.AST,
+			},
+			RouteOpCode:       opCode,
+			Keyspace:          vindexTable.Keyspace,
+			TargetDestination: dest,
+		}
```
frouioui, line 240: I think inside this `if` block, we should simply return the `*Route`. That way we can remove the entire `else`.

frouioui (commented):
The rest of the PR looks good to me. I'll wait for a second approval. cc @harshit-gangal

go/vt/vtgate/planbuilder/physical/route_planning.go
arthurschreiber: Ah, yeah, good idea! 👍

arthurschreiber: Done!

harshit-gangal:
> @harshit-gangal No specific use case in particular, just something I noticed that is broken. From what I gather, sequence tables are supposed to work just like "regular" tables (`INSERT`, `SELECT`, `DELETE`), with the exception that they support the `SELECT NEXT VALUE` syntax, right?

sequence table are specialized tables. It should not be used like a regular table.
It should be used only for auto_increment columns. DMLs queries should fail on them. 
We might add some special syntax for them like we have `select next` 

We should treat these as vitess internal tables.

arthurschreiber:
> sequence table are specialized tables. It should not be used like a regular table. It should be used only for auto_increment columns. DMLs queries should fail on them. We might add some special syntax for them like we have `select next`
> 
> We should treat these as vitess internal tables.

IMHO, it's awkward if those tables "look" like regular tables, but then don't behave like regular tables.

For example, we have tooling that "cleans" development databases by listing all tables tables and deleting their contents. Of course, this doesn't make a ton of sense for sequences, because they become unusable until they're initialized again, but unless Vitess somehow "hides" these tables (and provides actual SQL statements for things like sequence creation / initialization and reset), regular DMLs should just work.

arthurschreiber:
Also, this worked fine in V3, and no longer works in Gen4. 🤷

harshit-gangal:
> Also, this worked fine in V3, and no longer works in Gen4. shrug

ok. that's a regression. 
In future, we want to keep this hidden and should be auto-created.

harshit-gangal:
The issue with delete or update support on sequence is, that the values are cached in vttablet. So, the effect of DML will only works when the vttablet runs out of the cached value.
It will not give the desired outcome a user would expect.

arthurschreiber:
> The issue with delete or update support on sequence is, that the values are cached in vttablet. So, the effect of DML will only works when the vttablet runs out of the cached value.
> It will not give the desired outcome a user would expect.

Yeah, that doesn't sound great. That seems like a separate thing that needs to be fixed? The `vttablet` could understand that modifications on the sequence require the cache to be cleared (that would also need to be done with custom SQL statements for resetting a sequence).

go/vt/vtgate/planbuilder/physical/route_planning.go
```diff
@@ -226,6 +226,19 @@ func createPhysicalOperatorFromDelete(ctx *plancontext.PlanningContext, op *abst
...
+				AST:    op.AST,
+			},
+			RouteOpCode:       opCode,
+			Keyspace:          vindexTable.Keyspace,
+			TargetDestination: dest,
+		}, nil
+	}
+
 	primaryVindex, vindexAndPredicates, err := getVindexInformation(op.TableID(), op.Table.Predicates, vindexTable)
 	if err != nil {
```
harshit-gangal, line 243: instead of this let's do the same things we do in `Update` in `getVindexInformation`  check for the table keyspace if unsharded, return immediately. 

This is clearly a miss(BUG) and will impact any delete query for an unsharded table if that table is declared in vschema.

harshit-gangal:
This PR is not just related to sequence table anymore.
@arthurschreiber thank you for finding this bug and fixing it.

arthurschreiber:
> This is clearly a miss(BUG) and will impact any delete query for an unsharded table if that table is declared in vschema.

@harshit-gangal I think this only affects `reference` and `sequence` tables in unsharded keyspaces. Regular tables are unaffected. The logic in `SingleUnshardedKeyspace` takes care of regular tables and ensures we never call in that case `createPhysicalOperatorFromDelete`.

arthurschreiber:
Let me add another test for reference tables as well, before you merge this.

arthurschreiber:
@harshit-gangal I added a `reference` table test case, and modified the check as you suggested. 🙇‍♂️

harshit-gangal (approved):

PR #11463, opened 2022-10-11
Give all permissions in rbac in examples

GuptaManan100:
## Description

Today while running the local example I noticed that we don't give all the permissions to the users when we run the local-examples. This leads to the user not being able to manage the tablets' replication from the vtadmin page. This is a little counter-intuitive and not easy to debug. For first time users of Vitess, all the vtadmin operations should work without any issues. Rbac is a more of advanced thing that is of more importance when going into production.

This PR fixes the examples to give all the users all the accesses available so that everything works without any issues.

## Checklist

### DIFF
--- a/examples/local/vtadmin/rbac.yaml
+++ b/examples/local/vtadmin/rbac.yaml
@@ -1,17 +1,5 @@
 rules:
   - resource: "*"
-    actions:
-    - "get"
-    - "create"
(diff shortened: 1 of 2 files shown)


ajm188 (approved):

rsajwani (commented):
LGTM

PR #11465, opened 2022-10-11
Fix the script `check_make_sizegen`

frouioui:
## Description

This PR fixes and simplifies the `check_make_sizegen` script. It was previously failing as we were moving all the `cache_file` to the `/tmp` folder. With this change, the script has a similar structure as the `check_make_proto` script and runs correctly.

The failures could be observed in the CI, although the build was still successful.

## Checklist


tools/check_make_sizegen.sh
```diff
@@ -1,49 +1,22 @@
...
-  fi
-  mv "$TMP" "$SRC"
-done
+second_output=$(git status --porcelain)
 
-exit $STATUS
+diff=$(diff <( echo "$first_output") <( echo "$second_output"))
 
+if [[ "$diff" != "" ]]; then
+  echo "ERROR: Regenerated cache_size files do not match the current version."
```
deepthi, line 19:
```suggestion
  echo "ERROR: Regenerated cached_size files do not match the current version."
```

deepthi (commented):
👍 
Can we collect status from each check and fail at the end if one or more have failed?
cc @rohit-nayak-ps

frouioui:
> Can we collect status from each check and fail at the end if one or more have failed?

@deepthi, we can and we should. I'll update this PR.

tools/check_make_sizegen.sh
```diff
@@ -1,49 +1,22 @@
 #!/bin/bash
+
 #
 # Validate that the current version of the generated cache_size files match the output
 # generated by sizegen.
 #
-# This is used in Travis to verify that the currently committed version was
```
ajm188, line 6: > in Travis

lol

ajm188 (approved):
LGTM

PR #11438, opened 2022-10-04
[VTAdmin] `Validate`, `ValidateShard`, `ValidateVersionShard`, `GetFullStatus`

notfelineit:
## Description
This PR completes "VTAdmin Parity" by implementing:
- `Validate`
- `ValidateShard`
- `ValidateVersions`
- `GetFullStatus`

**Validate**
Validate is added to the clusters list, since validation occurs at the cluster level:
validate

**ValidateShard**
validateShard

**GetFullStatus**
`GetFullStatus` replaces `Replication Status` since it also includes that in its payload:
Screen Shot 2022-10-04 at 1 47 28 PM

**ValidateVersionShard**
validateVersionShard

## Related Issue(s)
- [x] https://github.com/vitessio/vitess/issues/8741
- [x] https://github.com/vitessio/vitess/issues/11258
- [x] https://github.com/vitessio/vitess/issues/8718
- [x] https://github.com/vitessio/vitess/issues/8718

## Checklist

-   [x] "Backport me!" label has been added if this change should be backported **Should be backported to 15**
-   [x] Tests were added or are not required **Not necessary ATM**
-   [x] Documentation was added or is not required **Not required**

PR #11470, opened 2022-10-12
Full Status tab improvements for VTAdmin

GuptaManan100:
## Description

The PR https://github.com/vitessio/vitess/pull/11438 introduced the `Full Status` tab in the vttablets in VTAdmin which shows the full replication status of the tablets. 
In this output it was noticed that the IO state and SQL state are printed as integers (because they are stored as an enum), instead of any meaningful output, so it becomes hard for the users to read this data unless they are familiar with Vitess internals on how the enum works.

This PR attempts to improve this situation by decoding these values before printing them using a replacer function in the `json.Stringify`.

Another thing that was noticed was that the semi-sync properties weren't printed if they are false. This is not desirable. We instead always want to print these settings. This PR addresses this too. It checks for these fields being explicitly set to true. If they aren't, then we set them to false.

Here are the images of the Full Status tab after the changes - 

Screenshot 2022-10-12 at 1 46 11 PM

Screenshot 2022-10-12 at 1 45 59 PM

## Checklist

### DIFF
--- a/web/vtadmin/src/components/routes/tablet/FullStatus.tsx
+++ b/web/vtadmin/src/components/routes/tablet/FullStatus.tsx
@@ -1,24 +1,24 @@
 import React from 'react';
 import { Link } from 'react-router-dom';
 import { useGetFullStatus } from '../../../hooks/api';
-import {topodata, vtadmin} from '../../../proto/vtadmin';
+import { topodata, vtadmin } from '../../../proto/vtadmin';
 import { formatAlias } from '../../../util/tablets';
 import { Code } from '../../Code';
 import style from './Tablet.module.scss';
-import {isNil} from "lodash-es";
+import { isNil } from 'lodash-es';
 
 interface Props {
     tablet: vtadmin.Tablet;
 }
 
 function stateReplacer(key: any, val: any) {
-    if ( key === 'io_state' || key == 'sql_state' ) {
+    if (key === 'io_state' || key === 'sql_state') {
         if (val === 3) {
-            return "Replication Running"
+            return 'Replication Running';
         } else if (val === 2) {
-            return "Replication Connected"
+            return 'Replication Connected';
         } else if (val === 1) {
-            return "Replication Stopped"
+            return 'Replication Stopped';
         }
     }
     return val;
@@ -46,16 +46,16 @@ const FullStatus: React.FC<Props> = ({ tablet }) => {
     }
     if (!isNil(data) && !isNil(data.status)) {
         if (data.status.semi_sync_primary_enabled !== true) {
-            data.status.semi_sync_primary_enabled = false
+            data.status.semi_sync_primary_enabled = false;
         }
         if (data.status.semi_sync_replica_enabled !== true) {
-            data.status.semi_sync_replica_enabled = false
+            data.status.semi_sync_replica_enabled = false;
         }
         if (data.status.semi_sync_primary_status !== true) {
-            data.status.semi_sync_primary_status = false
+            data.status.semi_sync_primary_status = false;
         }
         if (data.status.semi_sync_replica_status !== true) {
-            data.status.semi_sync_replica_status = false
+            data.status.semi_sync_replica_status = false;
         }
     }
 
--- a/web/vtadmin/src/components/routes/tablet/FullStatus.tsx
+++ b/web/vtadmin/src/components/routes/tablet/FullStatus.tsx
@@ -1,24 +1,23 @@
 import React from 'react';
 import { Link } from 'react-router-dom';
 import { useGetFullStatus } from '../../../hooks/api';
-import { topodata, vtadmin } from '../../../proto/vtadmin';
+import { vtadmin } from '../../../proto/vtadmin';
 import { formatAlias } from '../../../util/tablets';
 import { Code } from '../../Code';
 import style from './Tablet.module.scss';
-import { isNil } from 'lodash-es';
 
 interface Props {
     tablet: vtadmin.Tablet;
 }
 
-function stateReplacer(key: any, val: any) {
+function stateReplacer(key: string, val: number) {
     if (key === 'io_state' || key === 'sql_state') {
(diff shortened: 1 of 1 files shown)


web/vtadmin/src/components/routes/tablet/FullStatus.tsx
```diff
@@ -1,15 +1,29 @@
...
 
+function stateReplacer(key: any, val: any) {
+    if (key === 'io_state' || key === 'sql_state') {
+        if (val === 3) {
+            return 'Replication Running';
+        } else if (val === 2) {
+            return 'Replication Connected';
+        } else if (val === 1) {
+            return 'Replication Stopped';
+        }
```
ajm188, line 22: since this is already in the context of "IO Thread" or "SQL Thread" (in the ui, i mean), is "replication" maybe redundant here? what i am thinking is we could maybe just do:

```json
{
  "io_thread": "{running|connected|stopped}",
  "sql_thread": "{running|connected|stopped}"
}
```

ajm188 (approved):
LGTM, but a question about the presentation :)

web/vtadmin/src/components/routes/tablet/FullStatus.tsx
```diff
@@ -1,15 +1,29 @@
...
 import { formatAlias } from '../../../util/tablets';
 import { Code } from '../../Code';
 import style from './Tablet.module.scss';
+import { isNil } from 'lodash-es';
 
 interface Props {
     tablet: vtadmin.Tablet;
 }
 
+function stateReplacer(key: any, val: any) {
```
notfelineit, line 14:
```suggestion
function stateReplacer(key: string, val: number) {
```

notfelineit: You can add these types here! 😄

web/vtadmin/src/components/routes/tablet/FullStatus.tsx
```diff
@@ -30,8 +44,22 @@ const FullStatus: React.FC<Props> = ({ tablet }) => {
             </div>
         );
     }
+    if (!isNil(data) && !isNil(data.status)) {
```
notfelineit, line 47: Here you can simplify and do:
```suggestion
    if (data && data.status) {
```

web/vtadmin/src/components/routes/tablet/FullStatus.tsx
```diff
@@ -30,8 +44,22 @@ const FullStatus: React.FC<Props> = ({ tablet }) => {
...
+        if (data.status.semi_sync_primary_enabled !== true) {
+            data.status.semi_sync_primary_enabled = false;
+        }
```
notfelineit, line 50: Here if you are trying to set `data.status.semi_sync_primary_enabled` to a boolean, you can simplify by doing:
```suggestion
        data.status.semi_sync_primary_enabled = !!data.status.semi_sync_primary_enabled
```

notfelineit:
I left a few suggestions for cleaning up the code 😄 Potentially the swaps from integer --> human readable state could be done at the API level in the future - since there could be multiple clients for this API.

web/vtadmin/src/components/routes/tablet/FullStatus.tsx
deepthi: Agree with @ajm188. Further, assuming that the source for this enum is https://github.com/vitessio/vitess/blob/dcd1d3b2e4f6d4a4f71979a6403ca889df3bb8c7/go/mysql/constants.go#L739-L744 
2 corresponds to `Connecting` not `Connected`. And we might want to complete the set and represent `0` as `Unknown`.

deepthi:
> I left a few suggestions for cleaning up the code 😄 Potentially the swaps from integer --> human readable state could be done at the API level in the future - since there could be multiple clients for this API.

This would be preferable so that we don't have to keep two places in sync. However, I suspect that because we are returning a protobuf from the RPC, we don't have much of a choice.
Unless we change the protobuf itself to contain strings instead of enum values.

web/vtadmin/src/components/routes/tablet/FullStatus.tsx
GuptaManan100: Fixed the strings. I didn't add the Unknown field for 0 because for the primary tablet, the IO_State and SQL_state aren't populated, which then defaults to being 0. We don't want to show Unknown state there.

web/vtadmin/src/components/routes/tablet/FullStatus.tsx
GuptaManan100: Fixed

web/vtadmin/src/components/routes/tablet/FullStatus.tsx
GuptaManan100: Fixed

web/vtadmin/src/components/routes/tablet/FullStatus.tsx
GuptaManan100: Fixed

GuptaManan100:
@notfelineit Thank-you for all the suggestions! I have fixed them all up. @deepthi is right about the reason why I implemented the changes in the typescript side instead of the API. In the typescript code, we also verify that we indeed get the result matching the fields of the proto message and their types. 
The alternate way to do to in the API would entail creating a different type called `FullStatusReadable` which has the enum changed to strings and `omitempty` removed from some of the fields that we always want printed. Then the API would return this type instead of `FullStatus`. However, we would still have the same drawback that we have now, which is to keep the two in sync. Right now it is the enum and string values we print in the UI, the alternate would be the two proto message types.

GuptaManan100:
This is how the UI looks after the review comments - 
Screenshot 2022-10-13 at 11 48 36 AM

notfelineit:
@deepthi @GuptaManan100 that makes sense regarding the API proto! This looks good!

PR #11482, opened 2022-10-13
Move vtorc runners back to normal github runners

GuptaManan100:
## Description

There have been no failures in the past 24 hours on the self-hosted runners for VTOrc. This PR moves them back given that the flakiness has been fixed.

## Checklist


frouioui (approved):
That looks good to me. Correct me if I am wrong but the original fix is this PR: https://github.com/vitessio/vitess/pull/11442, right? I see it has been backported and merged onto `release-15.0`, we should also backport this current PR to `release-15.0`.

I removed the old required check for `Cluster (vtorc)(mysql57)`, I have tried adding the new one for `Cluster (vtorc) mysql57` but I cannot as it has not run on the branches yet. I'll add the check once this PR is merged.

PR #11489, opened 2022-10-13
[15.0] Fix VTOrc to handle multiple failures

### LINKED ISSUE #11488: Bug Report: VTOrc not being able to handle multiple failures
### Overview of the Issue

VTOrc is unable to handle multiple failures even though the durability policies should allow it. This was found when trying out a demo of VTOrc. 

The situations is such that we have a keyspace with semi_sync durability policy and a rdonly tablet and primary tablet's MySQL instances are dead. According to the durability policies, this is a recoverable failure if we have 2 more tablets to make forward progress. But VTOrc isn't able to fix it.

This is the screenshot from VTAdmin showing the state of the cluster
Screenshot 2022-10-12 at 4 41 43 PM

Here is something from the VTOrc replication analysis endpoint which seemed weird to me- 
Screenshot 2022-10-12 at 4 42 04 PM

The count of valid replicating replicas should be 0 but it isn't. For some reason the rdonly tablet still shows up as though it is a valid replicating replica.

### Reproduction Steps

1. Have a keyspace with 3 replicas, 1 rdonly tablet and semi-sync durability.
2. Shutdown the rdonly and primary's MySQL instances.
3. Wait and see VTOrc not being able to fix the issue

### Binary Version

```sh
main, release-15.0 and even before
```

```sh
main, release-15.0 and even before
```

### Log Fragments

_No response_

GuptaManan100:
## Description

This PR fixes the issue described in #11488. There were actually 2 underlying issues - 
1. We were using the incorrect value when we were trying to update the last_checked field in VTOrc. This lead to us not updating that field at all and VTOrc thinking that the last information it had from the rdonly (which was already dead) was the latest information. This led to the rdonly being marked as a valid replicating replica of the primary, even though it had long since died.
2. We failed fast in ERS when we see more than 1 errors. This led to the context being cancelled on the fillStatus call to the other replica's and then ERS errors out. Surprisingly we already had a test that was testing ERS with multiple failures, `TestRecoverWithMultipleFailures`. The key difference between the two situations is that, in the test we terminate the tablets too along with the MySQL instances, so the replica's return with the fillStatus output, before the grpc call to the tablets that are down times out. So the fail-fast code doesn't really matter, because the replicas have already finished.
When the tablets are up, but MySQL down, then if the failed tablets return earlier than the running ones, then we end up cancelling the context.

This PR removes this context cancellation when we receive more than 1 failures. This allows ERS to work even in the cases where the multiple failures are safe to fix based on the durability policies.

## Related Issue(s)

- Fixes #11488 
- #8975

## Checklist


deepthi (commented):
vtorc tests are failing :(

go/test/endtoend/vtorc/utils/utils.go
```diff
@@ -647,7 +644,7 @@ func PermanentlyRemoveVttablet(clusterInfo *VTOrcClusterInfo, tablet *cluster.Vt
 		for i, vttablet := range cellInfo.RdonlyTablets {
 			if vttablet == tablet {
 				// remove this tablet since its mysql has stopped
-				cellInfo.ReplicaTablets = append(cellInfo.ReplicaTablets[:i], cellInfo.ReplicaTablets[i+1:]...)
+				cellInfo.RdonlyTablets = append(cellInfo.RdonlyTablets[:i], cellInfo.RdonlyTablets[i+1:]...)
```
GuptaManan100, line 647: I can't believe this was the issue 😵‍💫 ...
We were deleting the tablet from the wrong list. After finding the index of the tablet to remove from the rdonly list, we removed a tablet from the replica list 🙄

GuptaManan100: This rdonly tablet which we don't remove, causes issues later as its nor really in a usable state.

deepthi (approved):

rsajwani (commented):
LGTM

PR #11492, opened 2022-10-13
Add GenerateShardRanges to vtctldclient

ajm188:
## Description

I also refactored the `ensureServerArg` to actually create the client (for commands that need it, now that we have one that doesn't), and renamed it accordingly:

```
vtctldclient  GenerateShardRanges 2                 
[
  "-80",
  "80-"
]
```

```
# showing that other commands still need `--server`
vtctldclient  GetCellInfo foo 2>&1 | tail -n 1      
E1013 15:48:33.918217   20963 main.go:56] please specify --server <vtctld_host:vtctld_port> to specify the vtctld server to connect to
```

## Checklist

### DIFF
--- a/go/cmd/vtctldclient/command/shards.go
+++ b/go/cmd/vtctldclient/command/shards.go
@@ -23,6 +23,7 @@ import (
 	"github.com/spf13/cobra"
 
 	"vitess.io/vitess/go/cmd/vtctldclient/cli"
+	"vitess.io/vitess/go/vt/key"
 	"vitess.io/vitess/go/vt/topo"
 	"vitess.io/vitess/go/vt/topo/topoproto"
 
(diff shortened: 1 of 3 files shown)


go/cmd/vtctldclient/command/shards.go
```diff
@@ -52,6 +53,38 @@ that shard.`,
...
+			if err != nil {
+				return err
+			}
+
+			fmt.Printf("%s\n", data)
+			return nil
+		},
+		Annotations: map[string]string{
+			skipClientCreationKey: "true",
+		},
```
mattlord, line 86: It doesn't make sense to use `VtctldClientProtocol` for this and set it to `nil` or `"none"` for this? At first glance it feels more intuitive to me given the existing `grpc` and `local` values.

mattlord (approved):
I had a few questions, but will let you make the final call on them. Thanks!

mattlord:
One other thing... could you please add a link in the description to the website PR where we'll add this to the docs?

go/cmd/vtctldclient/command/shards.go
ajm188: are you saying to do something like:

```go
var GenerateShardRanges = &cobra.Command{
	Annotations: map[string]string{
		"protocol": "local",
	}
}
```

?

If you're saying to avoid using `Annotations` altogether in favor of "just" checking the `VtctldClientProtocol` variable ... there's no way to force that to a specific value for _just_ this command, as the root pre-run (where that `ensureServerArg/getClient` func is called) runs before any funcs on a child command such as this one.

go/cmd/vtctldclient/command/shards.go
```diff
@@ -52,6 +53,38 @@ that shard.`,
...
+		Args:                  cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			n, err := strconv.ParseInt(cmd.Flags().Arg(0), 10, 64)
+			if err != nil {
+				return err
+			}
+
+			cli.FinishedParsing(cmd)
+
+			shards, err := key.GenerateShardRanges(int(n))
```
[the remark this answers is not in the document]
ajm188: i'm gonna leave this alone per your final sentence 😂 if someone actually ends up scaling to a keyspace with more than 2 billion shards, we can fix it then :)

go/cmd/vtctldclient/command/shards.go
mattlord: I thought we might be able to set `cmd.VtctldClientProtocol="none"` here. If not (and I think you're saying NOT here), then never mind. Thanks!

ajm188:
> One other thing... could you please add a link in the description to the website PR where we'll add this to the docs? The help output test file also needs to be updated. 🙂

https://github.com/vitessio/website/pull/1177

PR #11493, opened 2022-10-13
Move CI workflow to use latest community version of mysql 8.0

rsajwani:

## Description
This PR is build on top of https://github.com/vitessio/vitess/pull/11153. 
We are moving end to end and unit test to use latest version of MySql 8.0 instead of using percona. The reason for this is that Percona is still at 8.0.29, this doesn't allow our code base to test against the latest version of mysql.

Unit test are already on mysql community version but we are upgrading to higher apt-config version

## Related Issue(s)

#11511 

## Checklist


test/templates/cluster_endtoend_test.tpl
```diff
@@ -79,6 +81,21 @@ jobs:
 
         # Install everything else we need, and configure
         sudo apt-get install -y percona-server-server percona-server-client make unzip g++ etcd git wget eatmydata xz-utils
+
+        {{else}}
+q
```
frouioui, line 86: The workflows are in the `expected` state, aka not running, because of this line

test/templates/unit_test.tpl
```diff
@@ -92,7 +92,7 @@ jobs:
         sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 467B942D3A79BD29
 
         # mysql80
-        wget -c https://dev.mysql.com/get/mysql-apt-config_0.8.14-1_all.deb
+        wget -c https://dev.mysql.com/get/mysql-apt-config_0.8.3-1_all.deb
```
frouioui, line 95: Correct me if i am wrong, but shouldn't this be `mysql-apt-config_0.8.30-1_all.deb` or `mysql-apt-config_0.8.31-1_all.deb` ?

test/templates/cluster_endtoend_test.tpl
```diff
@@ -79,6 +81,21 @@ jobs:
...
+        # Get key to latest MySQL repo
+        sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 467B942D3A79BD29
+        # Setup MySQL 8.0
+        wget -c https://dev.mysql.com/get/mysql-apt-config_0.8.3-1_all.deb
+        echo mysql-apt-config mysql-apt-config/select-server select mysql-8.0 | sudo debconf-set-selections
+        sudo DEBIAN_FRONTEND="noninteractive" dpkg -i mysql-apt-config*
+        sudo apt-get update
+        # Install everything else we need, and configure
+        sudo apt-get install -y mysql-server mysql-client make unzip g++ etcd curl git wget eatmydata xz-utils
```
frouioui, line 95: Do we need `xz-utils` in this case? I thought it was required only when doing backups.

frouioui (commented):
Left a few comments. We should also update the PR's description to describe this change and ideally link to an issue if there is one.

frouioui:
By the way, do we need to forwardport this to `main` considering the work that has been merged in this PR: https://github.com/vitessio/vitess/pull/11153?

test/templates/unit_test.tpl
deepthi: This is the apt-config package, not the actual package. The latest version available for it is `0.8.20-1`, and that is what we should use

test/templates/cluster_endtoend_test.tpl
deepthi: some of the backup tests will continue to use the basic template. only xtrabackup tests are different. If we want to install this only for backup tests, we'll need to add another property to the template and write some conditional code.

deepthi:
> By the way, do we need to forwardport this to `main` considering the work that has been merged in this PR: #11153?

This is why we need a better PR description.
#11153 is already present in release-15.0. On top of that this PR is changing the MySQL version we use from Percona to community version.
The reason for this is that Percona is still at 8.0.29, so the new backup code we wrote for changes in 8.0.30 is not being exercised.
Unit tests are already using community version. This PR changes endtoend tests to do the same (except xtrabackup tests).

test/templates/cluster_endtoend_test.tpl
rsajwani: I think i tried to sneak in a quick push and leave the letter here :)

rsajwani:
FYI, [Cluster (26)](https://github.com/vitessio/vitess/actions/workflows/cluster_endtoend_26.yml)
was failing ..I tried to debug but didn't see it is because of my change. Running it manually then passed... So i am attributing it to flakiness...

test/templates/cluster_endtoend_test.tpl
frouioui: Okay

test/templates/unit_test.tpl
frouioui: Okay, makes sense!

frouioui (approved):
Looks good to me now

deepthi:
@frouioui @GuptaManan100 can you review the upgrade/downgrade tests to see what versions we use there, and whether any changes are needed?

frouioui:
@deepthi, this issue should be addressed as soon as we merge https://github.com/vitessio/vitess/pull/11509

Issue #8704, opened 2021-08-27
[vtadmin] Topology browser


doeg:
# vtctld2 UI

Found under http://localhost:15000/app/topo

<img width="1792" alt="Screen Shot 2021-08-27 at 12 46 30 PM" src="https://user-images.githubusercontent.com/855595/131161446-7c5f212d-3bc5-4b0e-9c59-16e194b7403e.png">

# VTAdmin Implementation Notes

- This ticket is _specifically_ for the topology browser interface. Sub-functionality is ticketed separately (e.g., viewing vtctld elections) even when it's part of the topology browser in the vtctld2 UI. 

- The topology browser offers a way to browse the Vitess topology in a more nest-y, treelike way. VTAdmin already has something akin to the topology browser in its infrastructure list pages.  Before re-implementing the vtctld2 topology browser as-is, let's consider whether the combination of the existing VTAdmin UI + global search would fulfill the same use cases (with less surface area to write + maintain). 

- We'll need to add a couple more vtadmin-api endpoints; specific endpoints TBD.

notfelineit:
The old topology feature provides the following info:
<img width="1850" alt="topology@2x" src="https://user-images.githubusercontent.com/31225471/163279401-d0a366fc-8244-4c24-9a34-247619abd401.png">

Auditing vtadmin to check if it shows the same info!

notfelineit:
Corresponding VTAdmin routes for the global topo:
<img width="1156" alt="topology-vtadmin-global@2x" src="https://user-images.githubusercontent.com/31225471/163287206-338b262d-620f-4844-ae9d-0738b7fbb1ef.png">

notfelineit:
Corresponding VTAdmin routes for the local topo:
<img width="991" alt="topology-vtadmin-local@2x" src="https://user-images.githubusercontent.com/31225471/163490561-a75dd40b-6300-4c44-84b5-0e056b853d23.png">

notfelineit:
I believe this is the main file that powers topology in the current vtctld2 UI: https://github.com/vitessio/vitess/blob/main/go/vt/vtctld/explorer.go

It should be useful for replicating topology in VTAdmin!

PR #11496, opened 2022-10-14
[VTAdmin] Topology Browser

notfelineit:
## Description
This PR implements the topology browser in VTAdmin. In doing so, it implements `VtctldServer.GetTopologyPath`. Given a `path`, `GetTopologyPath` returns the file located there in the topology server.

### Web Topology Browser
Only the root topology directories are loaded at first. This has the same behavior as the topology browser in `vtctld2`. "Nodes" can be expanded if they are not a "terminal" node (a file node). 

`GetTopologyPath(path)` is **only** called when the `Expand` button is clicked, and only for that specific path.

The graph is then built recursively with only the paths that have been expanded so far.

**VTAdmin example:**
Oct-13-2022 16-23-29

## Related Issue(s)
- https://github.com/vitessio/vitess/issues/8704
- https://github.com/vitessio/vitess/issues/8705

## Checklist

-   [x] Documentation was added or is not required **Will be added in vtadmin web documentation efforts later**

## Deployment Notes
N/A


go/vt/vtctl/grpcvtctldserver/topo.go
```diff
@@ -29,9 +34,65 @@ import (
...
 
 	topodatapb "vitess.io/vitess/go/vt/proto/topodata"
+	vschemapb "vitess.io/vitess/go/vt/proto/vschema"
 	"vitess.io/vitess/go/vt/proto/vtrpc"
 )
 
+// decodeContent uses the filename to imply a type, and proto-decodes
+// the right object, then echoes it as a string.
+//
+// This is copied from vtctl/topo. We cannot import vtctl/topo because of an import loop.
```
notfelineit, line 44: Is there a place to refactor this method so both packages can use it?

go/vt/vtctl/grpcvtctldserver/server.go
```diff
@@ -4183,6 +4216,54 @@ func StartServer(s *grpc.Server, ts *topo.Server) {
...
+	relativePath := cellPath[len(cell)+1:]
+	topoCell := vtctldatapb.TopologyCell{Name: parts[len(parts)-1], Path: cellPath}
+
+	conn, err := s.ts.ConnForCell(ctx, cell)
+	if err != nil {
+		err := vterrors.Errorf(vtrpc.Code_UNAVAILABLE, "error fetching connection to cell %s: %v", cell, err)
+		return nil, err
+	}
+
+	data, _, dataErr := conn.Get(ctx, relativePath)
```
notfelineit, line 4237: We cannot early-return if this error is non-nil, because when the cell is a directory cell, `conn.Get` will return an error when "getting" data.

Instead, we return an error if both `conn.ListDir` and `conn.Get` both return errors.

.github/workflows/code_freeze.yml
```diff
@@ -9,4 +9,4 @@ jobs:
     steps:
     - name: Fail if Code Freeze is enabled
       run: |
-        exit 0
+        exit 1
```
frouioui, line 12: This should be reverted back to `exit 0`

ajm188 (commented):
We're also missing the cli command to correspond with the VtctldServer method

notfelineit:
@ajm188 I wasn't sure if we needed it because there's the topocat command. I'll go ahead and add one for this anyway.

.github/workflows/code_freeze.yml
notfelineit: Sounds good!

go/vt/vtctl/grpcvtctldserver/topo.go
notfelineit: Added it to `topo` because adding it to `topoproto` would introduce an import cycle.

ajm188: wait that doesn't make sense; this package imports `topoproto`, where is the import cycle?

go/vt/vtctl/grpcvtctldserver/server.go
ajm188: > @ajm188 I wasn't sure if we wanted another cmd like the topocat one instead of this one for the command line client. I'll go ahead and add this one though.

sounds good! i would have then argued if people can "just" use topocat for this functionality then so should we 😂

go/vt/vtadmin/api.go
```diff
@@ -1170,6 +1171,19 @@ func (api *API) GetTablets(ctx context.Context, req *vtadminpb.GetTabletsRequest
 	}, nil
 }
 
+// GetTopologyPath is part of the vtadminpb.VTAdminServer interface.
+func (api *API) GetTopologyPath(ctx context.Context, req *vtadminpb.GetTopologyPathRequest) (*vtctldatapb.GetTopologyPathResponse, error) {
+	span, ctx := trace.NewSpan(ctx, "API.GetTopologyPath")
+	defer span.Finish()
+
+	c, err := api.getClusterForRequest(req.ClusterId)
```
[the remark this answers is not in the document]
notfelineit: Added a `TopologyResource` resource instead of using the `ClusterResource`.

go/vt/vtctl/grpcvtctldserver/server.go
ajm188: uhhhhh, i have no idea why this got threaded on this particular comment, sorry about that!

go/vt/vtctl/grpcvtctldserver/topo.go
ajm188: ohhh it's because of the `topo.GetExternalVitessClusterDir` got it

go/vt/vtctl/grpcvtctldserver/server.go
notfelineit: Oh I didn't see this because it got put in a weird place... I went ahead and added the command anyway!

go/vt/vtctl/grpcvtctldserver/topo.go
notfelineit: Yeah! It uses a lot of the types from `topo`!

dgraham (approved):

deepthi:
Upgrade/downgrade test failures are because we upgraded golang on release-15 to 1.18.7.
Since this PR is only adding a new vtctld API and not changing existing ones, overriding.
If they fail on the release-15.0 after back port, we'll need to investigate.

deepthi (approved):

PR #11498, opened 2022-10-14
[gen4 planner] Operator refactoring

systay:
## Description
I think it's about time to clean up the operator code a bit.

Here, I have unified the logical and physical operator trees, so an operator can be both if that is sensible. Some of the old operators had a logical and physical version with barely any difference between them.

How the tree structure is handled also changes much with this PR. A lot of the old code was concerned with how to send data from inputs to parents. With this new way of doing things, there is much less "tree traversal" code spread out.

I have tried to be careful and keep all the tests the same, so we can think of this as a pure refactoring. I have not been 100% successful - some of the operator tests unfortunately had to change. I was however successful in not changing the plan test expectations.


go/vt/vtgate/planbuilder/operators/derived.go
```diff
@@ -37,40 +35,18 @@ type Derived struct {
...
-}
-
-// Clone implements the PhysicalOperator interface
-func (d *Derived) Clone() abstract.PhysicalOperator {
+// Clone implements the Operator interface
+func (d *Derived) Clone(inputs []Operator) Operator {
+	checkSize(inputs, 1)
 	clone := *d
-	clone.Source = d.Source.Clone()
+	clone.Source = inputs[0]
```
frouioui, line 47: Shouldn't we call `Clone` on the input? Or is the caller already calling `Clone`?

go/vt/vtgate/planbuilder/operators/derived.go
```diff
@@ -37,40 +35,18 @@ type Derived struct {
...
 	clone.ColumnAliases = sqlparser.CloneColumns(d.ColumnAliases)
-	clone.Columns = make([]*sqlparser.ColName, 0, len(d.Columns))
-	for _, x := range d.Columns {
-		clone.Columns = append(clone.Columns, sqlparser.CloneRefOfColName(x))
-	}
+	clone.Columns = append([]*sqlparser.ColName{}, d.Columns...)
```
frouioui, line 49: Why not continuing to clone the `ColName`s of `d` here?

go/vt/vtgate/planbuilder/operators/node.go
```diff
@@ -0,0 +1,76 @@
+package operators
+
```
frouioui, line 2: the header is missing here

go/vt/vtgate/planbuilder/operators/derived.go
systay: if you look at the implementation of `CloneRefOfColName`, all it does is:

```go
func CloneRefOfColName(n *ColName) *ColName {
	return n
}
```

go/vt/vtgate/planbuilder/operators/delete.go
```diff
@@ -36,3 +36,8 @@ func (d *Delete) Introduces() semantics.TableSet {
 
 // ThisIsAnOperator implements the Operator interface
 func (d *Delete) ThisIsAnOperator() {}
+
+func (d *Delete) Clone(inputs []Operator) Operator {
+	checkSize(inputs, 0)
+	return d
+}
```
GuptaManan100, line 53: We don't clone delete at all?

go/vt/vtgate/planbuilder/operators/derived.go
systay: yeah, the idea is that the new `Clone` method gets the inputs given to it. that way, the operators don't have to think about what to do with inputs

go/vt/vtgate/planbuilder/operators/delete.go
systay: `Clone` is used for trying different alternatives of plans. these three ops are not really optimised, so there was no need to do an actual Clone. I could implement something to have it there, so it doesn't look so weird, but it's not really needed

go/vt/vtgate/planbuilder/operators/route_planning.go
```diff
@@ -63,7 +63,9 @@ func CreatePhysicalOperator(ctx *plancontext.PlanningContext, opTree Operator) (
 	case *Update:
 		return createPhysicalOperatorFromUpdate(ctx, op)
 	case *Delete:
-		return createPhysicalOperatorFromDelete(ctx, op)
+		panic(4)
```
GuptaManan100: Maybe remove this case entirely? We'll get a error instead of a panic then

go/vt/vtgate/planbuilder/operators/delete.go
systay: actually, it looks like all three structs have OK clone methods. it probably got fixed in later commits

go/vt/vtgate/planbuilder/operators/operator_test_data.txt
```diff
@@ -444,35 +441,3 @@ OuterJoin: {
...
-		Tables:
-			TableSet{1}:user_extra where id = 42
-		}
-	}]
-	Outer: 	Update {
-		table: 	TableSet{0}:`user` where id = (select id from user_extra where id = 42)
-	assignments:
-		u = 1
-	}
-}
```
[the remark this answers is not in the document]
systay: they need route planning to happen, and these tests do not have correct vschema to do that. I felt like we could lean on the plan tests instead of making this test more similar to the plan_tests

go/vt/vtgate/planbuilder/operators/route_planning.go
```diff
@@ -42,35 +41,44 @@ type (
...
+func mapToPhys(ctx *plancontext.PlanningContext, ops []Operator) (sources []Operator, err error) {
+	for _, in := range ops {
+		physOp, err := CreatePhysicalOperator(ctx, in)
+		if err != nil {
+			return nil, err
+		}
+		sources = append(sources, physOp)
+	}
+	return
+}
```
frouioui, line 56: In my opinion, this name is not very intuitive. It does `CreatePhysicalOperator` for a given slice of `Operators`, what about, for instance: `createMultiplePhysicalOperators` or simply `createPhysicalOperators`

systay: sure I can change this. using `map` for a function that maps values from one type to another is a pretty well established name, but we don't have much functional code, so maybe that is why you reacted to it

go/vt/vtgate/planbuilder/operators/subquery_planning.go
```diff
@@ -1,17 +1,16 @@
-package physical
+package operators
```
frouioui, line 2: The header is missing here too

frouioui (approved):
Looks good to me, I left a few comments. Arewefastyet is currently benchmarking the PR, let's wait and see the results.

go/vt/vtgate/planbuilder/operators/derived.go
frouioui: that's clever, okay!

go/vt/vtgate/planbuilder/operators/derived.go
frouioui: Makes sense now

go/vt/vtgate/planbuilder/operators/operator_test_data.txt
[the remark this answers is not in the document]
frouioui: What's missing in the `FakeSI` to do route planning?

frouioui:
The upgrade/downgrade tests are failing but this is expected as long as https://github.com/vitessio/vitess/pull/11509 is not merged. However, the tests are not required :) cc @systay

harshit-gangal (approved):
LGTM.

go/vt/vtgate/planbuilder/operators/operator_test_data.txt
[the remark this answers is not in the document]
systay: I tried adding it, but it requires a full vschema, and that seems like a distraction here. it would be nice to have more unit tests, but I don't think I feel they are worth the cost in this particular instance

PR #11499, opened 2022-10-14
bug fix: using self-referencing columns in HAVING should not overflow

systay:
## Description
The planner rewrites the HAVING clause in two steps, and in the second part, we were doing the the wrong rewrite, which lead to a cycle in the expression tree. When we then tried to turn the expression back to SQL, we ran into a stack overflow.

Example query that was failing:

```sql
select a+2 as a from user having a = 42
```

## Checklist


frouioui (approved):
That looks good to me. It's a good fix!

harshit-gangal (approved):

PR #11500, opened 2022-10-14
feat: deconstruct tuple comparisons so we can use them for routing decisions

systay:
## Description
Rewrites tuple comparisons early in the planner processing, so that later parts, such as deciding how to route the query, can use the predicates.

```sql
WHERE (a,b,c) = (1,2,3)
# gets rewritten to
WHERE a = 1 AND b = 2 AND c = 3
```

## Checklist


frouioui (approved):
LGTM

PR #11501, opened 2022-10-14
Properly deprecate flags and fix default for `--cell`

GuptaManan100:
## Description

This PR fixes two things -
1. It properly deprecates the flags by marking them as deprecated instead of just including `DEPRECATED` in their help output.
2. It removes the default for the `--cell` flag.

## Related Issue(s)

- #10697

## Checklist


go/cmd/vtgate/vtgate.go
```diff
@@ -36,12 +37,10 @@ import (
 	"vitess.io/vitess/go/vt/vterrors"
 	"vitess.io/vitess/go/vt/vtgate"
 	"vitess.io/vitess/go/vt/vtgate/planbuilder/plancontext"
-
-	topodatapb "vitess.io/vitess/go/vt/proto/topodata"
 )
 
 var (
-	cell                           = "test_nj"
+	cell                           = ""
```
GuptaManan100, line 43: I couldn't find how to mark a flag as required in pflags. We already enforce the flag's requirement manually by failing if the value is empty.

ajm188: `fs.MarkRequired`

ajm188 (approved):
Mark required and then lgtm!

go/cmd/vtgate/vtgate.go
GuptaManan100: I checked, there is nothing like `fs.MarkRequired` in `pflags`.

ajm188: ohhh shoot i forgot; it's a [cobra thing](https://pkg.go.dev/github.com/spf13/cobra#MarkFlagRequired)

GuptaManan100: I tried adding it, but it doesn't really work. I can still call vtgate without the cell flag specified and flag parsing doesn't complain.

deepthi (commented):
may be better to do the release notes changes on #11502 instead.
rest LGTM

doc/releasenotes/15_0_0_summary.md
```diff
@@ -71,7 +72,7 @@ The following VTTablet flags were deprecated in 7.0. They have now been deleted
 - --enable-autocommit Autocommit is always allowed
 
 #### vttablet startup flag deprecations
-- --enable-query-plan-field-caching has been deprecated. It will be removed in v16.
+- --enable-query-plan-field-caching and --enable_query_plan_field_caching have been deprecated. They will be removed in v16.
```
deepthi, line 75: I don't think we need to show both forms. Users will get it.

doc/releasenotes/15_0_0_summary.md
```diff
@@ -25,6 +25,7 @@
 - The deprecated vtgate/vtexplain/vtcombo flag `--planner_version` has been removed. Please use `--planner-version` instead.
 - The deprecated flag `--master_connect_retry` has been removed. Please use `--replication_connect_retry` instead.
 - `vtctl` commands that take shard names and ranges as positional arguments (e.g. `vtctl Reshard ks.workflow -80 -40,40-80`) need to have their positional arguments separated from their flag arguments by a double-dash separator to avoid the new parsing library from mistaking them as flags (e.g. `vtctl Reshard ks.workflow -- -80 -40,40-80`).
+- The `--cell` flag in the `vtgate` binary no longer has a default value. It is a required argument that has to be specified for the binary to run. Please explicitly specify the flag, if dependent on the flag's default value.
```
deepthi, line 28: Let's move this to #11502

GuptaManan100:
I'll resolve the conflicts in #11502 and make the release-notes changes.

PR #11514, opened 2022-10-17
Skip CI with the `Skip CI` label

frouioui:
## Description

This Pull Request uses the newly introduced `Skip CI` label to skip all the workflows. When this label is applied to a Pull Request, all future commits will not be tested through our CI, until someone removes the label and pushes another commit. This is useful when we, for instance, have a draft on which we keep pushing commits but we don't want to overload the CI's queue.

Here, all the workflows will be failed. The workflows complete in ~5 seconds.

## Checklist


rohit-nayak-ps (approved):
lgtm! 

a very welcome PR!

PR #11515, opened 2022-10-17
Fix `HAVING` rewriting made in #11306

frouioui:
## Description

This Pull Request fixes a recent change introduced by https://github.com/vitessio/vitess/pull/11306. An issue appeared several days ago as described in https://github.com/planetscale/discussion/discussions/308. We were rewriting the `having` expression to whatever was underneath the alias in the select list which is incorrect.

We can only safely do this rewrite when the expression we want to filter by is not using any columns from the tables, so we should check that the predicate doesn't reference any columns that are out of scope.

## Related Issue(s)

- https://github.com/planetscale/discussion/discussions/308

## Checklist

### DIFF
--- a/go/vt/vtgate/planbuilder/rewrite.go
+++ b/go/vt/vtgate/planbuilder/rewrite.go
@@ -174,29 +174,40 @@ func rewriteHavingClause(node *sqlparser.Select) {
 	exprs := sqlparser.SplitAndExpression(nil, node.Having.Expr)
 	node.Having = nil
 	for _, expr := range exprs {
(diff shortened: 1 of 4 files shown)


dbussink (approved):

frouioui:
Not backporting this as the original Pull Request (#11306) was not backported.

frouioui:
> Not backporting this as the original Pull Request (#11306) was not backported.

You can dismiss my comment. The issue lives in `release-15.0` too. Backport PR: https://github.com/vitessio/vitess/pull/11526

PR #11524, opened 2022-10-18
Add Gauge For CheckMySQL Running

GuptaManan100:
## Description

We have been seeing a failure of `DemotePrimary` wherein it just gets stuck indefinitely. The failure was traced to a call to acquiring a lock which was held by the `CheckMySQL` function which too is stuck indefinitely. This PR aims to add a gauge which tells us if `CheckMySQL` is running or not. If the gauge is constantly 1 for more than 10 minutes, then it is an indication that something is wrong and we can take a look.

## Checklist

### DIFF
--- a/go/vt/vttablet/tabletserver/state_manager.go
+++ b/go/vt/vttablet/tabletserver/state_manager.go
@@ -122,6 +122,7 @@ type stateManager struct {
 	// checkMySQLThrottler ensures that CheckMysql
 	// doesn't get spammed.
 	checkMySQLThrottler *sync2.Semaphore
+	checkMySQLRunning   bool
 
 	timebombDuration      time.Duration
 	unhealthyThreshold    sync2.AtomicDuration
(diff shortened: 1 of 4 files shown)


go/vt/vttablet/tabletserver/state_manager.go
```diff
@@ -330,6 +335,14 @@ func (sm *stateManager) CheckMySQL() {
 	}()
 }
 
+// isCheckMySQLRunning returns 1 if CheckMySQL function is in progress
+func (sm *stateManager) isCheckMySQLRunning() int64 {
```
systay, line 339: why int64 instead of bool?

systay (approved):

go/vt/vttablet/tabletserver/state_manager.go
GuptaManan100: We return int64 because the gauge is of that type. So return 1 when the value is true, 0 otherwise.

PR #11525, opened 2022-10-18
Update GitHub Actions worklflows to latest versions

dbussink:
To keep things nice and tidy, let's update to the latest versions which improves things like caching across builds etc.

## Checklist


frouioui (approved):
Looks good to me. I don't think a backport is necessary in this situation. We can always backport this PR in the future.

PR #11529, opened 2022-10-18
Remove example script that caused some confusion

mattlord:
## Description

I had added this script some time ago in https://github.com/vitessio/vitess/pull/8685 as I found it useful with the docker_local container. Since then, however, I've completely stopped using it. Meanwhile, some have mentioned that they found its existence confusing. So let's just remove it.

## Related Issue(s)

  - Follow-up to: https://github.com/vitessio/vitess/pull/8685

## Checklist

-   [x] Tests were are not required


frouioui (approved):
LGTM

PR #9426, opened 2021-12-20
Add list support to vindex functions

mattlord:
## Description
This adds support for vindex function batch lookups using the `IN` SQL operator:
```
mysql> select * from customer.hash where id = 100;
+------+-------------+-------------+-----------+------------------+-------+
| id   | keyspace_id | range_start | range_end | hex_keyspace_id  | shard |
+------+-------------+-------------+-----------+------------------+-------+
| 100  | ���V��    | �           |           | 83aab1569cbe1b08 | 80-   |
+------+-------------+-------------+-----------+------------------+-------+
1 row in set (0.00 sec)

mysql> select * from customer.hash where id in (100,1,999);
+------+-------------+-------------+-----------+------------------+-------+
| id   | keyspace_id | range_start | range_end | hex_keyspace_id  | shard |
+------+-------------+-------------+-----------+------------------+-------+
| 100  | ���V��    | �           |           | 83aab1569cbe1b08 | 80-   |
| 1    | k@�J�K�    |             | �         | 166b40b44aba4bd6 | -80   |
| 999  | X�Ο��IH    |             | �         | 58afce9fbfe54948 | -80   |
+------+-------------+-------------+-----------+------------------+-------+
3 rows in set (0.00 sec)

mysql> select * from customer.hash where id like "5";
ERROR 1105 (HY000): unsupported: where clause for vindex function must be of the form id = <val> or id in(<val>,...) (not equality)
```

For more info about the use case, please see: https://github.com/vitessio/vitess/issues/9042

## Related Issue(s)
Fixes: https://github.com/vitessio/vitess/issues/9042

## Checklist
- [x] Should this PR be backported? _**I don't think so**_
- [x] Tests were added
- [x] Documentation is not required -- **_I could not find any documentation to modify for using the vindex functions???_**

PR #11531, opened 2022-10-18
Optimize List Support In Vindex Functions

mattlord:
## Description

In #9426 we added list support for [Vindex Functions](https://vitess.io/docs/reference/features/vindexes/#query-vindex-functions) using the`IN` operator so that you can get the mappings for N values. When doing so, [we loop over the values in the list and call `Vindex.Map()` for each one](https://github.com/vitessio/vitess/blob/e803ac7255f7a5a6bce74c232a92867e28b4c465/go/vt/vtgate/engine/vindex_func.go#L113-L137). This was fine for the built-in Vindex types like [`hash`](https://github.com/vitessio/vitess/blob/e803ac7255f7a5a6bce74c232a92867e28b4c465/go/vt/vtgate/vindexes/hash.go#L76-L87) as the same amount of work was done either way, but for [Unique (Consistent) Lookup Vindexes](https://vitess.io/docs/user-guides/vschema-guide/unique-lookup/) this is very inefficient because its `Vindex.Map()` implementation [makes a single call to the backing table for all passed values](https://github.com/vitessio/vitess/blob/e803ac7255f7a5a6bce74c232a92867e28b4c465/go/vt/vtgate/vindexes/consistent_lookup.go#L90-L112) and that lookup is also much more expensive than the built-in ones like `hash` as it requires querying the remote backing table.

This PR adds an optimization *for unique vindexes* by moving to a single `Vindex.Map()` call for the list which will cut down dramatically on the work done for Unique Lookup Vindexes and any implementation that directly supports mapping N values at a time.

## Related Issue(s)

  - Follow-up to: https://github.com/vitessio/vitess/pull/9426

## Checklist

-   [x] Tests were are not required (internal optimization, relying on existing tests)
-   [x] Documentation is not required (internal optimization)

### DIFF
--- a/go/vt/vtgate/engine/vindex_func.go
+++ b/go/vt/vtgate/engine/vindex_func.go
@@ -139,8 +139,8 @@ func (vf *VindexFunc) mapVindex(ctx context.Context, vcursor VCursor, bindVars m
 	result := &sqltypes.Result{
 		Fields: vf.Fields,
 	}
+	var destination key.Destination
 	for i, value := range values {
-		var destination key.Destination
 		if vf.Vindex.IsUnique() {
 			destination = destinations[i]
 		} else {
--- a/go/vt/vtgate/engine/vindex_func.go
+++ b/go/vt/vtgate/engine/vindex_func.go
@@ -126,6 +126,11 @@ func (vf *VindexFunc) mapVindex(ctx context.Context, vcursor VCursor, bindVars m
 	if err != nil {
 		return nil, err
 	}
+	if len(destinations) != len(values) {
+		// should never happen
+		return nil, vterrors.Errorf(vtrpcpb.Code_INTERNAL, "Vindex.Map() length mismatch: input values count is %d, output destinations count is %d",
+			len(values), len(destinations))
+	}
 	result := &sqltypes.Result{
 		Fields: vf.Fields,
 	}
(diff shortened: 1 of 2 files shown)


go/vt/vtgate/engine/vindex_func.go
```diff
@@ -125,16 +125,21 @@ func (vf *VindexFunc) mapVindex(ctx context.Context, vcursor VCursor, bindVars m
...
 		Fields: vf.Fields,
 	}
-	for _, value := range values {
+	destinations, err := vf.Vindex.Map(ctx, vcursor, values)
+	if err != nil {
+		return nil, err
+	}
+	if len(destinations) != len(values) {
+		// should never happen
+		return nil, vterrors.Errorf(vtrpcpb.Code_INTERNAL, "unique Vindex.Map() length mismatch: input values count is %d, output destinations count is %d",
```
harshit-gangal, line 134: nit: remove `unique` from the error message.

harshit-gangal (approved):

go/vt/vtgate/engine/vindex_func.go
mattlord: Yeah, I noticed that too. Thanks! Fixed in: https://github.com/vitessio/vitess/pull/11531/commits/0ce233cf2d80ed836f22f15ee66f8490111f82f6

Issue #11454, opened 2022-10-07
Bug Report: Transaction errors when querying using `union all` after querying `information_schema`


johanstenberg92:
### Overview of the Issue

When starting a transaction, querying the `information_schema` and then using `union all` on the regular selected **sharded** database, this error message shows:

```
ERROR 1317 (70100): target: customer.0.primary: vttablet: rpc error: code = Aborted desc = transaction 1665138448496974090: in use: for query (CallerID: userData1)
```

### Reproduction Steps

This one is somewhat difficult to reproduce in a public setting, but I used the docker local image to repro it. Set up your local env using docker here: https://vitess.io/docs/14.0/get-started/local-docker/. I used the version shown below 👇.

Run the following commands from the `README.md` when running the docker image:

```
# Bring up initial cluster and commerce keyspace
./101_initial_cluster.sh

# Insert and verify data
mysql < ../common/insert_commerce_data.sql
mysql --table < ../common/select_commerce_data.sql

# Bring up customer keyspace
./201_customer_tablets.sh

# Initiate move tables
vtctlclient MoveTables -- --source commerce --tables 'customer,corder' Create customer.commerce2customer

# Validate
vtctlclient VDiff customer.commerce2customer

# Cut-over
vtctlclient MoveTables -- --tablet_types=rdonly,replica SwitchTraffic customer.commerce2customer
vtctlclient MoveTables -- --tablet_types=primary SwitchTraffic customer.commerce2customer

# Clean-up
vtctlclient MoveTables Complete customer.commerce2customer

# Prepare for resharding
./301_customer_sharded.sh
./302_new_shards.sh

# Reshard
vtctlclient Reshard -- --source_shards '0' --target_shards '-80,80-' Create customer.cust2cust

# Validate
vtctlclient VDiff customer.cust2cust

# Cut-over
vtctlclient Reshard -- --tablet_types=rdonly,replica SwitchTraffic customer.cust2cust
vtctlclient Reshard -- --tablet_types=primary SwitchTraffic customer.cust2cust
```

Now, enter `mysql` and use the `customer` database. Then run:
```
start transaction; select table_name from information_schema.tables where table_schema = database() and table_comment <> "vitess_sequence"; (select 'corder' from corder limit 1) union all (select 'customer' from customer limit 1);
```

You'll see an error message such as:
```
[...]

PR #11534, opened 2022-10-19
[15.0] Fix: concatenate engine in transaction

harshit-gangal:
## Description

This PR forces concatenate engine to do sequential execution of queries when the session is in a transaction.

## Related Issue(s)

- Fixes https://github.com/vitessio/vitess/issues/11454

## Checklist

### DIFF
--- a/go/test/endtoend/vtgate/queries/informationschema/informationschema_test.go
+++ b/go/test/endtoend/vtgate/queries/informationschema/informationschema_test.go
@@ -214,8 +214,13 @@ func TestInfrSchemaAndUnionAll(t *testing.T) {
 	conn, err := mysql.Connect(context.Background(), &vtConnParams)
 	require.NoError(t, err)
 
-	utils.Exec(t, conn, "start transaction")
-	utils.Exec(t, conn, `select connection_id()`)
-	utils.Exec(t, conn, `(select 'corder' from t1 limit 1) union all (select 'customer' from t7_xxhash limit 1)`)
-	utils.Exec(t, conn, "rollback")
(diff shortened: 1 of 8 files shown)


systay (approved):
Nice code. 👍

GuptaManan100 (approved):
Looks good to me!

go/test/endtoend/vtgate/queries/informationschema/informationschema_test.go
```diff
@@ -206,3 +203,24 @@ func TestMultipleSchemaPredicates(t *testing.T) {
 	require.Error(t, err)
 	require.Contains(t, err.Error(), "specifying two different database in the query is not supported")
 }
+
+func TestInfrSchemaAndUnionAll(t *testing.T) {
+	clusterInstance.VtGateExtraArgs = append(clusterInstance.VtGateExtraArgs, "--planner-version=gen4")
+	require.NoError(t,
+		clusterInstance.RestartVtgate())
```
frouioui, line 210: We should change the planner version of VTGate back to its original value once this test is finished. Otherwise the other tests will only be executed using the `Gen4` planner. Otherwise, we could simply use the planner version comment directive.

frouioui (approved):
Left an important comment regarding the tests. Otherwise everything looks good to me.

go/test/endtoend/vtgate/queries/informationschema/informationschema_test.go
harshit-gangal: I thought about that. but does not see that as an important change. Any new test added to the file can change if seems required.

PR #11552, opened 2022-10-21
[refactor] Predicate push down

systay:
## Description
This is mostly a refactoring of how predicates are pushed down the operator tree. The exception is how we handle predicate pushing into `UNION` clauses. This can happen if we have a predicate on a derived table that consists of an `UNION`. The old behaviour was not really correct.

## Checklist


go/vt/vtgate/planbuilder/operators/apply_join.go
```diff
@@ -73,3 +75,56 @@ func (a *ApplyJoin) Clone(inputs []Operator) Operator {
...
+	a.LHS = lhs
+	vars := map[string]int{}
+	for i, idx := range idxs {
+		vars[bvName[i]] = idx
+	}
+	rhs, err := PushPredicate(ctx, predicate, a.RHS)
+	if err != nil {
+		return err
+	}
+	a.RHS = rhs
```
frouioui, line 127: you could use `a.setRHS(rhs)` here

go/vt/vtgate/planbuilder/operators/apply_join.go
```diff
@@ -73,3 +75,56 @@ func (a *ApplyJoin) Clone(inputs []Operator) Operator {
...
+func (a *ApplyJoin) addJoinPredicate(ctx *plancontext.PlanningContext, expr sqlparser.Expr) error {
+	bvName, cols, predicate, err := BreakExpressionInLHSandRHS(ctx, expr, TableID(a))
+	if err != nil {
+		return err
+	}
+	lhs, idxs, err := PushOutputColumns(ctx, a.LHS, cols...)
+	if err != nil {
+		return err
+	}
+	a.LHS = lhs
```
frouioui, line 118: Same here, you could use `setLHS`

go/vt/vtgate/planbuilder/operators/apply_join.go
systay: sure. what would be the benefit? the interface methods are useful when you can't touch the fields directly, but here I can

go/vt/vtgate/planbuilder/operators/operator_funcs.go
```diff
@@ -368,3 +299,104 @@ func BreakExpressionInLHSandRHS(
...
+type joinOperator interface {
+	Operator
+	getLHS() Operator
+	getRHS() Operator
+	setLHS(Operator)
+	setRHS(Operator)
+	makeInner()
+	isInner() bool
+	addJoinPredicate(ctx *plancontext.PlanningContext, expr sqlparser.Expr) error
+}
```
frouioui, line 312: nit: i think we should move this to a better place, like in one of the join files (`apply_join.go` or `join.go`)

go/vt/vtgate/planbuilder/operators/operator_funcs.go
```diff
@@ -368,3 +299,104 @@ func BreakExpressionInLHSandRHS(
...
+		return join, err
+
+	case deps.IsSolvedBy(TableID(join)):
+		// if we are dealing with an outer join, always start by checking if this predicate can turn
+		// the join into an inner join
+		if !join.isInner() && canConvertToInner(ctx, expr, TableID(join.getRHS())) {
+			join.makeInner()
+		}
+
+		if !join.isInner() {
```
frouioui, line 353: it seems that `makeInner` always change the join into an inner join for both `applyJoin` and `join`.

go/vt/vtgate/planbuilder/operators/vindex.go
```diff
@@ -82,11 +83,54 @@ outer:
 	return idxs, nil
 }
 
-// CheckValid implements the Operator interface
+// checkValid implements the Operator interface
 func (v *Vindex) CheckValid() error {
```
frouioui, line 87: i think you can revert this change

frouioui (approved):

Issue #11554, opened 2022-10-21
Bug Report: container based deployments don't allow for separate MySQL upgrades


dbussink:
### Overview of the Issue

When using the Vitess based containers, it's not possible today upgrade MySQL separately. The specific immediate issue that breaks is backups when using MySQL 8.0.30. 

The problem shows up because the Vitess containers would still version check against the bundled `mysqld` in the Vitess container and not against the actual MySQL container running. This results in backups being broken because it doesn't handle the new redo log location. 

### Reproduction Steps

Start a Vitess cluster with the Vitess docker containers, but run with MySQL 8.0.30 containers for MySQL separately. 

This is something possible with the Vitess operator, see https://github.com/planetscale/vitess-operator/blob/3065813f791d89598194506bece6ada0189d4696/pkg/operator/vttablet/pod.go#L171 where it's possible to specify a separate MySQL image. 

### Binary Version

```sh
Any recent Vitess version since at least Vitess 13.
```

### Operating System and Environment details

```sh
Any container based setup
```

### Log Fragments

```sh
I1019 09:10:22.574372       1 main.go:94] I1019 09:10:22.573739 backup.go:107] I1019 09:10:22.581679 builtinbackupengine.go:450] Backup "#innodb_redo": 0.00% (0.00/4.00kb)
I1019 09:10:22.860363       1 main.go:94] I1019 09:10:22.860112 backup.go:107] I1019 09:10:22.867733 builtinbackupengine.go:223] resetting mysqld read-only to true
I1019 09:10:22.861468       1 main.go:94] I1019 09:10:22.861277 backup.go:107] I1019 09:10:22.869277 builtinbackupengine.go:240] restarting mysql replication
E1019 09:10:22.988394       1 main.go:94] E1019 09:10:22.988161 backup.go:107] E1019 09:10:22.995924 backup.go:127] backup is not usable, aborting it: [cannot copy data: read /vt/vtdataroot/vt_1479756646/innodb/logs/#innodb_redo: is a directory]
E1019 09:10:23.049860       1 main.go:103] remote error: rpc error: code = Unknown desc = TabletManager.Backup on aws_useast1a_1-1479756646 error: cannot copy data: read /vt/vtdataroot/vt_1479756646/innodb/logs/#innodb_redo: is a directory: cannot copy data: read /vt/vtdataroot/vt_1479756646/innodb/logs/#innodb_redo: is a directory
[...]

PR #11555, opened 2022-10-21
Detect redo log location dynamically based on presence

dbussink:
## Description

Instead of relying on a specific version check, detect if the new redo log location is present or not.

This makes the backup logic independent from the version check against MySQL. With this change, 8.0.30 can be backed up just as well as any other version.

## Related Issue(s)

Solves the backup part of
https://github.com/vitessio/vitess/issues/11554 by removing the dependency on the version check. There might be still other places where that issue can crop up, but it solves the immediate breaking issue with backups.

## Checklist

Marked this for back porting to 14 and 15, also since we back ported the other 8.0.30 fixes to 14 so we need this one as well.

### DIFF
--- a/go/vt/mysqlctl/backup_test.go
+++ b/go/vt/mysqlctl/backup_test.go
@@ -23,20 +23,11 @@ import (
 	"sort"
 	"testing"
 
-	"github.com/stretchr/testify/require"
-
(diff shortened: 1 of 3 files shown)


deepthi (approved):
💯

PR #11566, opened 2022-10-24
evalengine: Support built-in MySQL function for CONV function

Weijun-H:

## Description
The following functions will be supported
- [x]  CONV()

## Related Issue(s)
- #9647 

## Checklist


go/vt/vtgate/evalengine/string.go
```diff
@@ -265,3 +268,37 @@ func (builtinRepeat) typeof(env *ExpressionEnv, args []Expr) (sqltypes.Type, fla
...
+
+type builtinConv struct {
+}
+
+func (builtinConv) call(env *ExpressionEnv, args []EvalResult, result *EvalResult) {
+	inarg := &args[0]
+	fromBase := args[1]
+	toBase := args[2]
+
+	if inarg.isNull() || fromBase.isNull() || toBase.isNull() || fromBase.int64() < 2 || fromBase.int64() > 36 || toBase.int64() < 2 || toBase.int64() > 36 {
```
vmg, line 280: You cannot call `int64` over and over; `toBase` and `fromBase` need to be converted to their integer forms before they can be checked.

go/vt/vtgate/evalengine/string.go
```diff
@@ -265,3 +268,37 @@ func (builtinRepeat) typeof(env *ExpressionEnv, args []Expr) (sqltypes.Type, fla
...
+		result.setNull()
+		return
+	}
+
+	inarg.makeSignedIntegral()
```
vmg, line 285: You don't want a signed integral for `inarg`. You want the raw textual bytes so you can process them.

go/vt/vtgate/evalengine/string.go
```diff
@@ -265,3 +268,37 @@ func (builtinRepeat) typeof(env *ExpressionEnv, args []Expr) (sqltypes.Type, fla
...
+
+	num, _ := strconv.ParseInt(fmt.Sprint(inarg.int64()), int(fromBase.int64()), 64)
```
vmg, line 287: MySQL handles overflow with saturation, not 0-truncation.

Try:

```
+-----------------------------------------------------------------------------------------------------------------------+
| conv(99999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999, 10, 2) |
+-----------------------------------------------------------------------------------------------------------------------+
| 1111111111111111111111111111111111111111111111111111111111111111                                                      |
+-----------------------------------------------------------------------------------------------------------------------+
```

go/vt/vtgate/evalengine/string.go
```diff
@@ -265,3 +268,37 @@ func (builtinRepeat) typeof(env *ExpressionEnv, args []Expr) (sqltypes.Type, fla
...
+
+	convNum := strconv.FormatUint(uint64(num), int(toBase.int64()))
+	convNum = strings.ToUpper(convNum)
+	result.setRaw(sqltypes.VarChar, []byte(convNum), inarg.collation())
+	result.makeTextual(env.DefaultCollation)
```
vmg, line 292: Absolutely no point on calling `makeTextual` after `setRaw`. Particularly if you're going to change the collation. You can set the desired collation directly on the previous call.

go/vt/vtgate/evalengine/string.go
```diff
@@ -265,3 +268,37 @@ func (builtinRepeat) typeof(env *ExpressionEnv, args []Expr) (sqltypes.Type, fla
...
+}
+
+func (builtinConv) typeof(env *ExpressionEnv, args []Expr) (sqltypes.Type, flag) {
+	if len(args) != 3 {
+		throwArgError("CONV")
+	}
+	_, f1 := args[0].typeof(env)
+	// typecheck the right-hand argument but ignore its flags
+	args[1].typeof(env)
+
```
vmg, line 302: Missing typecheck for `args[2]`.

go/vt/vtgate/evalengine/integration/string_fun_test.go
```diff
@@ -204,3 +204,35 @@ func TestBuiltinRepeat(t *testing.T) {
...
+	bases := []string{
+		"-1",
+		"1",
+		"2",
+		"4",
+		"8",
+		"10",
+		"16",
+		"32",
+	}
```
vmg, line 226: This is a very limited set of test cases. You need to widen this. Particularly important: hex literals.

vmg (changes requested):
This is not a great implementation @Weijun-H. You need to put more thought into this. If you widen your set of integration test cases you'll see that it breaks at the seams very easily.

go/vt/vtgate/evalengine/integration/string_fun_test.go
```diff
@@ -204,3 +204,51 @@ func TestBuiltinRepeat(t *testing.T) {
...
+		"10 + '10' + 10",
+		"10 + '10' - 10",
+		"-10",
+		"'10'",
+		"10+'10'+'10a'+X'0a'",
+		"10 / 10",
+		"X'0FFFFFFFFFFFFFF'",
+		"99999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999",
+		"-99999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999",
+	}
```
GuptaManan100, line 226: These still aren't strong enough cases. What happens if we have a `+` in the beginning? What if we have 2 of those in the beginning? What if it is a string with a + and not a number?

go/vt/vtgate/evalengine/string.go
```diff
@@ -265,3 +270,109 @@ func (builtinRepeat) typeof(env *ExpressionEnv, args []Expr) (sqltypes.Type, fla
...
+	inarg3 := &args[2]
+	fromBase := inarg2.int64()
+	toBase := inarg3.int64()
+	fromNum = 0
+	t := inarg.typeof()
+
+	if inarg.isNull() || inarg2.isNull() || inarg3.isNull() || fromBase < 2 || fromBase > 36 || toBase < 2 || toBase > 36 {
+		result.setNull()
+		return
+	}
```
GuptaManan100, line 294: I don't think the checks `inarg2.isNull()` or `inarge3.isNull()` is required. The values of fromBase and toBase would be 0 if they were null respectively, so the condition would still be true.

go/vt/vtgate/evalengine/string.go
```diff
@@ -265,3 +270,109 @@ func (builtinRepeat) typeof(env *ExpressionEnv, args []Expr) (sqltypes.Type, fla
...
+				continue
+			}
+			if (fromBase <= 9 && c >= '0' && c <= rune('0'+fromBase)) || (fromBase > 9 && ((c >= '0' && c <= '9') || (c >= 'a' && c <= rune('a'+fromBase-9)))) {
+				continue
+			} else {
+				rawString = rawString[:i]
+				break
+			}
+		}
+	}
```
GuptaManan100, line 311: Is there any specific reason why you need the float type as a separate case?

go/vt/vtgate/evalengine/string.go
```diff
@@ -265,3 +270,109 @@ func (builtinRepeat) typeof(env *ExpressionEnv, args []Expr) (sqltypes.Type, fla
...
+
+	re, _ := regexp.Compile(`[+-]?[0-9.x]+[a-vA-Vx]*`)
+	for _, num := range re.FindAllString(rawString, -1) {
```
GuptaManan100, line 314: Why do you need all substring matches? Don't you only need the first one? In what case would have more than 1 matches?

GuptaManan100 (changes requested):
I actually checked `'+10'` and it fails.
I also checked and `'10-9+10'` also fails. You aren't supposed to evaluate the string in the conv function. Just use the raw bytes and interpret them in the given base. Your output for this string is `9` (which I don't understand how you got to anyways, `11` I could understand) but MySQL just says 10, because it ignores everything after the - sign.

GuptaManan100:
In my opinion, the best way to do the conversion is to do it in two parts. You should convert using from_base and handle the behaviour of a negative from_base value. Once that conversion is complete, then do a second conversion into the to_base value and also handle the negative values there.
From what I have observed from MySQL behaviour is that, if the from_base is negative, then your input is treated as a signed integer, so the limits of int64 apply. Otherwise, the limits of uint64 apply!
```
mysql [localhost:8026] {msandbox} (test) > SELECT CONV(1000000000000000000000000000000,10,10);
+---------------------------------------------+
| CONV(1000000000000000000000000000000,10,10) |
+---------------------------------------------+
| 18446744073709551615                        |
+---------------------------------------------+

mysql [localhost:8026] {msandbox} (test) > SELECT CONV(1000000000000000000000000000000,-10,10);
+----------------------------------------------+
| CONV(1000000000000000000000000000000,-10,10) |
+----------------------------------------------+
| 9223372036854775807                          |
+----------------------------------------------+
```

and then if your to_base is negative, then your output can be signed - 
```

mysql [localhost:8026] {msandbox} (test) > SELECT CONV(-1,-10,10);
+----------------------+
| CONV(-1,-10,10)      |
+----------------------+
| 18446744073709551615 |
+----------------------+
1 row in set (0.00 sec)

mysql [localhost:8026] {msandbox} (test) > SELECT CONV(-1,-10,-10);
+------------------+
| CONV(-1,-10,-10) |
+------------------+
| -1               |
+------------------+
1 row in set (0.01 sec)
```

GuptaManan100:
These are all worthwhile test cases to have -
```sql
mysql [localhost:8026] {msandbox} (test) > SELECT CONV(10000000000000000000000000,-10,-10);
+------------------------------------------+
| CONV(10000000000000000000000000,-10,-10) |
+------------------------------------------+
| 9223372036854775807                      |
+------------------------------------------+

mysql [localhost:8026] {msandbox} (test) > SELECT CONV(10000000000000000000000000,-10,10);
+-----------------------------------------+
| CONV(10000000000000000000000000,-10,10) |
+-----------------------------------------+
| 9223372036854775807                     |
+-----------------------------------------+

mysql [localhost:8026] {msandbox} (test) > SELECT CONV(10000000000000000000000000,10,10);
+----------------------------------------+
| CONV(10000000000000000000000000,10,10) |
+----------------------------------------+
| 18446744073709551615                   |
+----------------------------------------+

mysql [localhost:8026] {msandbox} (test) > SELECT CONV(10000000000000000000000000,10,-10);
+-----------------------------------------+
| CONV(10000000000000000000000000,10,-10) |
+-----------------------------------------+
| -1                                      |
+-----------------------------------------+
```

go/vt/vtgate/evalengine/string.go
```diff
@@ -265,3 +267,110 @@ func (builtinRepeat) typeof(env *ExpressionEnv, args []Expr) (sqltypes.Type, fla
...
+	var toNum string
+	if isNeg {
+		temp := strconv.FormatUint(uint64(-fromNum), int(toBase))
+		toNum = strings.ToUpper(temp)
+	} else {
+		temp := strconv.FormatUint(fromNum, int(toBase))
+		toNum = strings.ToUpper(temp)
+	}
+
+	inarg.makeTextualAndConvert(env.DefaultCollation)
```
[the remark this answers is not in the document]
Weijun-H: > You can just extract the collation from inarg and change the collation and then type cast the result directly using that.

I have no idea how to do in this way.

PR #11570, opened 2022-10-25
mysql: Improve MySQL 5.6 GTID parsing performance

vmg:
## Description

This is a small optimization pass for MySQL 5.6 GTID Set parsing performance. This is something that we do very often during VStreaming and which we were performing quite inefficiently.

Before/after:

```
name                   old time/op    new time/op    delta
MySQL56GTIDParsing-16    13.0µs ± 1%     6.8µs ± 2%  -47.60%  (p=0.000 n=9+10)

name                   old alloc/op   new alloc/op   delta
MySQL56GTIDParsing-16    4.38kB ± 0%    3.10kB ± 0%  -29.07%  (p=0.000 n=10+10)

name                   old allocs/op  new allocs/op  delta
MySQL56GTIDParsing-16      68.0 ± 0%      25.0 ± 0%  -63.24%  (p=0.000 n=10+10)
```

cc @dbussink 

## Checklist

### DIFF
--- a/go/mysql/mysql56_gtid_set.go
+++ b/go/mysql/mysql56_gtid_set.go
@@ -61,8 +61,9 @@ func parseInterval(s string) (interval, error) {
 // ParseMysql56GTIDSet is registered as a GTIDSet parser.
 //
 // https://dev.mysql.com/doc/refman/5.6/en/replication-gtids-concepts.html
-func ParseMysql56GTIDSet(input string) (Mysql56GTIDSet, error) {
+func ParseMysql56GTIDSet(s string) (Mysql56GTIDSet, error) {
 	set := make(Mysql56GTIDSet)
+	input := s
 
 	// gtid_set: uuid_set [, uuid_set] ...
 	for len(input) > 0 {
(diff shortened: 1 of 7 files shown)


systay (approved):

go/mysql/mysql56_gtid_set.go
```diff
@@ -35,70 +36,74 @@ func (iv interval) contains(other interval) bool {
...
 		if uuidSet == "" {
 			continue
 		}
 
 		// uuid_set: uuid:interval[:interval]...
-		parts := strings.Split(uuidSet, ":")
-		if len(parts) < 2 {
+		head, tail, ok := strings.Cut(uuidSet, ":")
+		if !ok {
 			return nil, vterrors.Errorf(vtrpc.Code_INTERNAL, "invalid MySQL 5.6 GTID set (%q): expected uuid:interval", s)
```
mattlord, line 87: I'd like to take this opportunity to clarify user facing errors if you don't mind as 5.6 makes no sense -- it's really just MySQL with the 5.6 part being when they were added.

mattlord (approved):
Thanks! ❤️ I'll leave the error message part up to you.

PR #11571, opened 2022-10-25
planner enhancement: use column names for bind variables when normali…

systay:
…zing

Signed-off-by: Andres Taylor <andres@planetscale.com>

## Checklist

### DIFF
--- a/go/vt/sqlparser/normalizer.go
+++ b/go/vt/sqlparser/normalizer.go
@@ -140,15 +140,7 @@ func (nz *normalizer) convertLiteralDedup(node *Literal, cursor *Cursor) {
 	}
 
 	// Check if there's a bindvar for that value already.
-	var key string
-	if bval.Type == sqltypes.VarBinary || bval.Type == sqltypes.VarChar {
-		// Prefixing strings with "'" ensures that a string
-		// and number that have the same representation don't
-		// collide.
-		key = "'" + node.Val
-	} else {
-		key = node.Val
-	}
+	key := keyFor(bval, node)
 	bvname, ok := nz.vals[key]
 	if !ok {
 		// If there's no such bindvar, make a new one.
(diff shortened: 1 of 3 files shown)


go/vt/sqlparser/normalizer.go
```diff
@@ -185,13 +207,54 @@ func (nz *normalizer) convertLiteral(node *Literal, cursor *Cursor) {
...
+	key := keyFor(bval, lit)
+	bvname, ok := nz.vals[key]
+	if !ok || len(lit.Val) > 256 {
+		// If there's no such bindvar, or we have a big value, make a new one.
+		// Big values are most likely not for vindexes.
+		// We save a lot of CPU because we avoid building
+		bvname = nz.reserved.ReserveColName(col)
+		nz.vals[key] = bvname
+		nz.bindVars[bvname] = bval
+	}
```
frouioui, line 246: Shouldn't we return here?

systay: Nah, we do want to change the literal to a bindvar. I rewrote this method to be easier to read. let me know if I succeeded or not

dbussink:
@systay Not for this PR, but one of the tests still failing shows something I was wondering about as well:

```
"/* leading */ update `user` set a = :vtg1 where id = :id /* trailing */
```

Would it be possible in a follow up to also change the naming for the `set` in this case? And call that `:a` for example?

systay:
> Would it be possible in a follow up to also change the naming for the `set` in this case? And call that `:a` for example?

Yeah, it should be straight forward to do. I'll do a follow-up with that

go/vt/vtgate/executor_framework_test.go
```diff
@@ -576,6 +576,7 @@ func executorStream(executor *Executor, sql string) (qr *sqltypes.Result, err er
 }
 
 func assertQueries(t *testing.T, sbc *sandboxconn.SandboxConn, wantQueries []*querypb.BoundQuery) {
+	t.Helper()
```
frouioui, line 579: I think we can remove this duplication

go/vt/vtgate/executor_test.go
```diff
@@ -496,21 +497,34 @@ func TestExecutorShowColumns(t *testing.T) {
...
+	sort.Strings(s)
+	return strings.Join(s, "")
+}
+
+func assertMatchesNoOrder(t *testing.T, expected, got string) {
+	t.Helper()
+	if sortString(expected) != sortString(got) {
+		t.Errorf("for query: expected \n%s \nbut actual \n%s", expected, got)
+	}
+}
```
frouioui, line 511: You can use the `AssertMatchesNoOrder` method in the `vitess.io/vitess/go/test/endtoend/utils` package

frouioui (approved):

go/vt/vtgate/executor_test.go
systay: not really. that method expects a `*mysql.Conn`, and I don't have one

go/vt/vtgate/executor_framework_test.go
systay: oops

systay:
> Yeah, it should be straight forward to do. I'll do a follow-up with that
@dbussink it was pretty easy to add. https://github.com/vitessio/vitess/pull/11581

PR #11572, opened 2022-10-25
fixing urls

rsajwani:

## Description

This PR fix the urls of help diff in release notes and summary.

## Checklist


frouioui:
I think either way is fine, since we are merging the diff files on main too. But this is cleaner and depends only on `release-15.0`.

PR #11581, opened 2022-10-26
planner enhancement: nice bindvar names for update

systay:
## Description
Changes the auto-parameterized update queries to user nicer bindvar names when possible.

```sql
# original query
update user set a=2 where id = 1

# old behaviour
update user set a=:bv1 where id = :id

# new behaviour
update user set a=:a where id = :id
```

## Related Issue(s)
Continuation of the work in #11571

### DIFF
--- a/go/vt/sqlparser/normalizer_test.go
+++ b/go/vt/sqlparser/normalizer_test.go
@@ -120,19 +120,17 @@ func TestNormalize(t *testing.T) {
 	}, {
 		// val should be reused only in subqueries of DMLs
 		in:      "update a set v1=(select 5 from t), v2=5, v3=(select 5 from t), v4=5",
-		outstmt: "update a set v1 = (select :bv1 from t), v2 = :bv2, v3 = (select :bv1 from t), v4 = :bv3",
+		outstmt: "update a set v1 = (select :bv1 from t), v2 = :bv1, v3 = (select :bv1 from t), v4 = :bv1",
 		outbv: map[string]*querypb.BindVariable{
 			"bv1": sqltypes.Int64BindVariable(5),
-			"bv2": sqltypes.Int64BindVariable(5),
-			"bv3": sqltypes.Int64BindVariable(5),
 		},
 	}, {
 		// list vars should work for DMLs also
 		in:      "update a set v1=5 where v2 in (1, 4, 5)",
-		outstmt: "update a set v1 = :bv1 where v2 in ::bv2",
+		outstmt: "update a set v1 = :v1 where v2 in ::bv1",
 		outbv: map[string]*querypb.BindVariable{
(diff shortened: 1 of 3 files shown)


go/vt/sqlparser/normalizer_test.go
```diff
@@ -120,19 +120,17 @@ func TestNormalize(t *testing.T) {
 	}, {
 		// val should be reused only in subqueries of DMLs
 		in:      "update a set v1=(select 5 from t), v2=5, v3=(select 5 from t), v4=5",
-		outstmt: "update a set v1 = (select :bv1 from t), v2 = :bv2, v3 = (select :bv1 from t), v4 = :bv3",
+		outstmt: "update a set v1 = (select :bv1 from t), v2 = :bv1, v3 = (select :bv1 from t), v4 = :bv1",
```
dbussink, line 123: Is this correct that the names here now get reused? Like it's the same constant value `5` I guess but it means that the bind var naming does show information about what the original value was (like if they were the same)?

Should all literals be treated as always different here?

frouioui (approved):
looks good to me!

go/vt/sqlparser/normalizer_test.go
frouioui: ah that's a good point.

systay: I don't have opinions in what the behaviour should be. Don't know how well thought-out this was, and how much of it was accidental. WDYT?

dbussink: I think not aliasing is the safer / less confusing option here?

systay: So, the balancing act we have to perform here is between readable bindvars, and how well the planner can plan the queries.

If we have a query written like this:

```sql
select * 
from user, user_extra 
where user.id = 5 
  and user_extra.user_id = 5
```

Say that both `user.id` and `user_extra.user_id` are the sharding keys for the respective tables.
When parameterizing this query, if we extract these two literal `5`s to different bindvars, the planner will not know that the two columns have to have the same value, and won't be able merge the queries into a single tablet call, and will need to plan two calls.

So, the normalizer tried to use as few bindvars as possible, so the planner can see what values are the same and which are not.

My comment `Don't know how well thought-out this was, and how much of it was accidental. ` was more around the logic for when to re-use literal values across subqueries, not in general.

dbussink: @systay Hmm, that's tricky yeah. It's I think kinda impossible to know up front if there's a meaningful relation between the values like in your example, or if they are just accidentally the same value but are otherwise entirely unrelated. I was more thinking about the latter, but that doesn't work for the former case.

PR #11585, opened 2022-10-26
[don't review]Self hosted cluster 26

rsajwani:

## Checklist


deepthi:
@rsajwani this can be closed now, right? We've deleted these tests on main..

PR #11594, opened 2022-10-27
vtbackup: add --disable-redo-log flag (default false)

maxenglander:
Recently [vtbackup was changed to disable innodb redo log](https://github.com/vitessio/vitess/pull/11330) during the replicate-from-primary stage of vtbackup. The goal was to squeeze out a bit more performance. 

My mental model at the time was that "if mysqld of vtbackup fails, we won't try to restart from where we left off, so if any data gets corrupted the next attempt won't re-use any corrupted data".

It turns out that isn't true. If mysqld OOMs, mysqld_safe will restart it, and the next mysqld process may re-use InnoDB data corrupted in a previous run.

I think it's probably better to put this behind a flag and let users opt in to it, at least until we have a better handle on things like mysqld OOMs.

### DIFF
--- a/go/cmd/vtbackup/vtbackup.go
+++ b/go/cmd/vtbackup/vtbackup.go
@@ -118,6 +118,7 @@ var (
 	initDBSQLFile    string
 	detachedMode     bool
 	keepAliveTimeout = 0 * time.Second
+	disableRedoLog   = false
 )
 
 func registerFlags(fs *pflag.FlagSet) {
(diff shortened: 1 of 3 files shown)


mattlord (approved):

mattlord:
Please don't forget the docs PR 🙂

maxenglander:
@mattlord I don't see any reference to `vtbackup` in the docs. Should I create a reference page similar to the other programs?

mattlord:
> @mattlord I don't see any reference to `vtbackup` in the docs. Should I create a reference page similar to the other programs?

Good point! You're off the hook then. 🙂 We can add general docs for it separately.

PR #11596, opened 2022-10-27
Add commit, plan, and execute timing information to `QueryResult` and `Result`

iheanyi:
## Description

This pull request adds various timing information for plan, execute, and commit to `QueryResult`. This is valuable information that callers communicating with `vtgate` can use for observability information and also having an accurate idea of how long various statements took, without the network overhead.

## Related Issue(s)
N/A

## Checklist


proto/query.proto
```diff
@@ -398,6 +399,9 @@ message QueryResult {
   repeated Row rows = 4;
   string info = 6;
   string session_state_changes = 7;
+  vttime.Duration execute_duration = 8;
+  vttime.Duration plan_duration = 9;
+  vttime.Duration commit_duration = 10;
```
mattrobenolt, line 404: Would it make sense to wrap this in a `QueryStats` message instead?

So something like: 

```proto
message QueryStats {
  vttime.Duration execute = 1;
  vttime.Duration plan = 2;
  vttime.Duration commit = 3;
```

Then just add:

```
message QueryResult {
 ...
  QueryStats stats = x;
}```

Seems a bit better to wrap this in it's own Message imo, but I don't have strong opinions about this.

frouioui (approved):
This change looks good to me. I will wait for a second approval to merge this.

proto/query.proto
iheanyi: Okay, this has been addressed.

vmg (changes requested):
I don't think this is the right design for what we're trying to do here. The `QueryResult` is a core proto message for all query serving, i.e. it's using as an intermediate result for every stream event during stream executes, in the vstream, etc, and in many places where having `QueryStats` either doesn't make sense or is redundant.

I would suggest rethinking this to accomplish the goal we have without cluttering this core message. If we're looking for the runtime stats of queries when they're `Executed`, why not return a `QueryStats` result via GRPC metadata instead of embedding a new field?

PR #11601, opened 2022-10-28
Speedup DDLs by not reloading table size stats

pudiva:
## Description

Currently, obtaining table sizes from mysql involves joining `information_schema.tables`, which can be very costly on systems with a large number of tables.

My tests on a system with 13k tables took around 20s without this patch, and only 4s with it.

Instead of synchronously recalculating table size stats after every DDL, let them be outdated until the periodic schema reload fixes it.

## Related Issue(s)

This is an simpler and less risky alternative to:
* #11386 

## Checklist

cc @arthurschreiber

### DIFF
--- a/go/mysql/flavor.go
+++ b/go/mysql/flavor.go
@@ -157,6 +157,7 @@ type flavor interface {
 	enableBinlogPlaybackCommand() string
 	disableBinlogPlaybackCommand() string
 
+	baseShowTables() string
 	baseShowTablesWithSizes() string
 
 	supportsCapability(serverVersion string, capability FlavorCapability) (bool, error)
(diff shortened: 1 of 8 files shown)


pudiva:
Hiii @systay! 👋

If you have a little time, would you mind taking a look at this one, please? It's quite important for us here!

Thanks!!1 ✨

pudiva:
btw most of the PR is boilerplate and substitution, the real magic happens in: **go/vt/vttablet/tabletserver/schema/engine.go**

harshit-gangal (approved):
I like this idea of a delayed load of table sizes to speed up base table updates.

pudiva:
Sorry for the force push, some tests were failing because I forgot to substitute some instances of `Reload(ctx)` with `Reload(ctx, true)`...

Then while I was looking at it, I realised that I could keep the `Reload()` API and all the calls to it unchanged by creating another lower level function `ReloadAtEx()` that accepted all the options, and use that one just in the DDLs, where I needed the extra flexibility.

So now the patch is much less noisy 🙂

harshit-gangal:
> Oh yes, one more thing: would it be possible to backport this to v15?

Backport will not be possible as it does not qualify as bug or regression.

pudiva:
Rebased from main ✨

PR #11606, opened 2022-10-31
[refactor planner] Columns and predicates on operators

systay:
## Description
A few more touch ups around operators and the expressions they pass through. Here I have refactored how we push down predicates and columns between the operators.


go/vt/vtgate/semantics/semantic_state.go
```diff
@@ -407,3 +407,22 @@ func (st *SemTable) SingleUnshardedKeyspace() (*vindexes.Keyspace, []*vindexes.T
 	}
 	return ks, tables
 }
+
+// EqualsExpr compares two expressions using the semantic analysis information.
+// This means that we use the binding info to recognize that two ColName's can point to the same
+// table column even though they are written differently. Example would be the `foobar` column in the following query:
+// `SELECT foobar FROM tbl ORDER BY tbl.foobar`
+// The expression in the select list is not equal to the one in the ORDER BY,
+// but they point to the same column and would be considered equal to this method
```
GuptaManan100, line 416: 'by' this method.

go/vt/vtgate/planbuilder/operators/vindex.go
```diff
@@ -67,70 +67,70 @@ func (v *Vindex) Clone(inputs []Operator) Operator {
...
-outer:
-	for i, newCol := range columns {
-		for j, existingCol := range v.Columns {
-			if sqlparser.EqualsExpr(newCol, existingCol) {
-				idxs[i] = j
-				continue outer
-			}
+func (v *Vindex) AddColumn(_ *plancontext.PlanningContext, expr sqlparser.Expr) (int, error) {
+	// TODO: unify with table
+	col, ok := expr.(*sqlparser.ColName)
```
GuptaManan100, line 72: Do you want to handle this TODO in this PR or a follow-up for later? I am fine either way

go/vt/vtgate/planbuilder/operators/operator.go
```diff
@@ -35,6 +35,14 @@ type (
 	Operator interface {
 		Clone(inputs []Operator) Operator
 		Inputs() []Operator
+
+		// AddPredicate is used to push predicates. It pushed it as far down as is possible in the tree.
+		// If we encounter a join and the predicate depends on both sides of the join, the predicate will be split into two parts,
+		// where data is fetched from the LHS of the join to be used in the evaluation on the RHS
+		AddPredicate(ctx *plancontext.PlanningContext, expr sqlparser.Expr) (Operator, error)
+
+		// AddColumn tells an operator which columns it should output.
```
GuptaManan100, line 44: `AddColumn tells an operator to also output an additional column specified. If the column already exists, its offset is returned.`
I think this comment makes more sense because AddColumn doesn't provide an exhaustive list. Like it doesn't say these are all the columns an operator needs to output.

go/vt/vtgate/planbuilder/operators/filter.go
```diff
@@ -60,8 +67,15 @@ func (f *Filter) UnsolvedPredicates(st *semantics.SemTable) []sqlparser.Expr {
...
-func newFilter(op Operator, expr ...sqlparser.Expr) Operator {
-	return &Filter{
-		Source: op, Predicates: expr,
+func (f *Filter) AddPredicate(ctx *plancontext.PlanningContext, expr sqlparser.Expr) (Operator, error) {
+	newSrc, err := f.Source.AddPredicate(ctx, expr)
+	if err != nil {
+		return nil, err
 	}
+	f.Source = newSrc
+	return f, nil
```
GuptaManan100, line 76: I think here we can check if the source ends up creating a filter for this predicate, we can coalesce the two filters into one.

GuptaManan100 (approved):
Everything else LGTM!

go/vt/vtgate/planbuilder/operators/filter.go
GuptaManan100: This happens in filter.compact, so all good.

PR #11611, opened 2022-10-31
Removing SharedPitr_tls and Backup_transfrom test from CI

rsajwani:
## Description

This PR is doing bunch of cleanup in CI 

1. Remove pitr_tls test
    -  Since the test is not doing what it is suppose to do. It was trying to use SSL which is not a Vitess feature instead it is MySQL feature and we don't need to test it. So test was not adding any value.
    -  During SSL certificate generation, we were creating the certificate but we weren't using them correctly in our cnf file. Instead of using hardcoded server-001 for all the sql servers we are using their respective certificates.
     This is an example of excerpt of my.cnf after this change:
    ```
    ## /Users/rameezsajwani/Code/fork/vitess/vtdataroot/vtroot_8101/ssl_0000004121/ssl.cnf

    ssl_ca=/Users/rameezsajwani/Code/fork/vitess/vtdataroot/vtroot_8101/ssl_0000004121/ca-cert.pem
    ssl_cert=/Users/rameezsajwani/Code/fork/vitess/vtdataroot/vtroot_8101/ssl_0000004121/server-instance-009-cert.pem
    ssl_key=/Users/rameezsajwani/Code/fork/vitess/vtdataroot/vtroot_8101/ssl_0000004121/server-instance-009-key.pem
    ```
log before this change
```
2022-10-31T20:38:04.004973Z 0 [Warning] [MY-013595] [Server] Failed to initialize TLS for channel: mysql_main. See below for the description of exact issue.
2022-10-31T20:38:04.004993Z 0 [Warning] [MY-010069] [Server] Failed to set up SSL because of the following SSL library error: Unable to get certificate
```

logs after this change.
```
2022-10-31T21:56:46.060895Z 0 [Warning] [MY-010068] [Server] CA certificate /Users/rameezsajwani/Code/fork/vitess/vtdataroot/vtroot_8301/ssl_0000001869/ca-cert.pem is self signed.
2022-10-31T21:56:46.060914Z 0 [System] [MY-013602] [Server] Channel mysql_main configured to support TLS. Encrypted connections are now supported for this channel.
```

2. Standard pitr test was using semi_sync but with only one replica and primary, which was resulting in indeterministic behavior during backup. So we are changing the test topology to use two replica with a primary.
3. Since backup_transform is deprecated, we are removing it from CI.
4. Delete vtbackup_transofrm test.
5. Rename 'vtbackup_transform' cluster name to 'vtbackup' because it no longer have any transform test.

## Related Issue(s)

  - Fixes: #11642

## Checklist


rsajwani:
spoke to Matt offline. I am going to delete the vtbackup_transform test as well.

deepthi:
> Since backup_transform is deprecated, we are removing it from CI.

And this is fine because the code is slated for deletion before v16

test/config.json
```diff
@@ -118,15 +118,6 @@
...
-		"backup_transform": {
-			"File": "unused.go",
-			"Args": ["vitess.io/vitess/go/test/endtoend/backup/transform", "-timeout", "15m"],
-			"Command": [],
-			"Manual": false,
-			"Shard": "vtbackup_transform",
-			"RetryMax": 1,
-			"Tags": ["upgrade_downgrade_backups"]
-		},
 		"backup_transform_mysqlctld": {
```
deepthi, line 121: We can delete this test also. In fact we should be able to delete the whole `test/endtoend/backup/transform` directory.

deepthi (commented):
Nice work, and well-written PR description.
We need to remove one more test and delete the test code for the transform tests.

go/test/endtoend/cluster/mysqlctl_process.go
```diff
@@ -104,13 +104,22 @@ func (mysqlctl *MysqlctlProcess) startProcess(init bool) (*exec.Cmd, error) {
...
 			sslPathData := struct {
-				Dir string
+				Dir        string
+				ServerCert string
+				ServerKey  string
 			}{
-				Dir: sslPath,
+				Dir:        sslPath,
+				ServerCert: clientServerKeyPair.ServerCert,
+				ServerKey:  clientServerKeyPair.ServerKey,
```
mattlord, line 120: Nice! ❤️

deepthi (approved):
💯

deepthi:
Overriding code owner approval requirement from @vitessio/query-serving - the only change in vtgate code fixes a lint error.

PR #11617, opened 2022-11-01
VReplication: Avoid gap lock contention for parallel insert workers

mattlord:
## Checklist


mattlord:
Closing this for now as InnoDB should only be using [insert intention locks](https://dev.mysql.com/doc/refman/8.0/en/innodb-locking.html#innodb-insert-intention-locks) for the INSERTs we do in the workers. I don't think we'll have locking issues with the copy_state queries either but we can re-open this later if it seems worthwhile.

PR #11622, opened 2022-11-02
Move horizon planning to operators

frouioui:
## Description
A refactoring that introduces a Horizon as a logical operator.

During query planning, we start by figuring out where we are going to send the queries. This is done by inspecting the `FROM` & `WHERE` clauses. 

A horizon is what we call in the planner the stuff that is outside the `FROM ... WHERE` clause. If we are able to send the whole query to a single shard, we don't do much more planning - we simply add the rest of the clauses (projections, ordering, limit, aggregation and more) to the query and we are done.

If we can't delegate everything to MySQL, we do more planning, and this is what we call horizon planning.

This change moves the horizon into the operator tree, instead of having special handling of derived tables, unions and subqueries.

## Related Issue(s)
#11626

## Checklist


go/vt/vtgate/planbuilder/operators/apply_join.go
```diff
@@ -68,7 +68,7 @@ func NewApplyJoin(lhs, rhs Operator, predicate sqlparser.Expr, leftOuterJoin boo
 func (a *ApplyJoin) IPhysical() {}
 
 // Clone implements the Operator interface
-func (a *ApplyJoin) Clone(inputs []Operator) Operator {
+func (a *ApplyJoin) clone(inputs []Operator) Operator {
```
frouioui, line 71: nit: we can also change the comment `Clone` -> `clone`

go/vt/vtgate/planbuilder/operators/horizon_planning.go
```diff
@@ -0,0 +1,55 @@
...
+
+func planHorizon(ctx *plancontext.PlanningContext, in *Horizon) (Operator, error) {
+	rb, isRoute := in.Source.(*Route)
+	if isRoute && rb.IsSingleShard() {
+		return planSingleShardRoute(in.Select, rb, in)
+	}
+
+	return nil, errNotHorizonPlanned
+}
+func planSingleShardRoute(statement sqlparser.SelectStatement, rb *Route, horizon *Horizon) (Operator, error) {
```
frouioui, line 53: nit: new line :)

go/vt/vtgate/planbuilder/operators/horizon.go
```diff
@@ -0,0 +1,57 @@
...
+*/
+
+package operators
+
+import (
+	"vitess.io/vitess/go/vt/sqlparser"
+	"vitess.io/vitess/go/vt/vtgate/planbuilder/plancontext"
+)
+
+// Horizon is an operator we use until we decide how to handle the source to the horizon.
```
frouioui, line 24: I am not sure I fully understand this. We add the `Horizon` operator into the tree until we have decided how the rest of the tree will do horizon planning?

frouioui: Looking at the operator test data, it feels like it encapsulates each node in the tree.

go/vt/vtgate/planbuilder/operators/operator.go
```diff
@@ -14,13 +14,31 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+// Package operators contains the operators used to plan queries.
+/*
+The operators go through a few phases while planning:
+1. Logical
```
frouioui, line 20: That's a nice write up

go/vt/vtgate/planbuilder/operators/horizon.go
systay: Yes. We need to do the physical route planning before we do the horizon planning

go/vt/vtgate/planbuilder/operators/operator.go
systay: I'll add info about horizons in the op tree as well

frouioui (commented):
LGTM, can't approve as this is my PR.

frouioui:
We should update the Pull Request description.

GuptaManan100 (approved):
I like the changes!

PR #11631, opened 2022-11-03
normalize more expressions

systay:
## Description
We were not auto-parameterizing literal values on the `ORDER BY` and `GROUP BY`.

This is because literals can have special meaning when used in these clauses.

For example:

```sql
select name from user order by 1
```

In the query above, `1` does not mean the integer value 1, instead it refers to the first column returned.
On the other hand, in the following query `1` really means integer 1.

```sql
select name from user order by age + 1
```

This PR changes the normalizer to change literals in ORDER BY/GROUP BY as long as they are not directly following the clause.

## Checklist

### DIFF
--- a/go/vt/sqlparser/normalizer_test.go
+++ b/go/vt/sqlparser/normalizer_test.go
@@ -185,7 +185,7 @@ func TestNormalize(t *testing.T) {
 		outstmt: "select a, b from t group by 1",
 		outbv:   map[string]*querypb.BindVariable{},
 	}, {
-		// ORDER BY column_position
+		// ORDER BY with literal inside complex expression
 		in:      "select a, b from t order by field(a,1,2,3) asc",
 		outstmt: "select a, b from t order by field(a, :bv1, :bv2, :bv3) asc",
 		outbv: map[string]*querypb.BindVariable{
--- a/go/vt/sqlparser/normalizer_test.go
+++ b/go/vt/sqlparser/normalizer_test.go
@@ -180,6 +180,15 @@ func TestNormalize(t *testing.T) {
 		outstmt: "select a, b from t order by 1 asc",
 		outbv:   map[string]*querypb.BindVariable{},
 	}, {
+		// ORDER BY column_position
+		in:      "select a, b from t order by field(a,1,2,3) asc",
+		outstmt: "select a, b from t order by field(a, :bv1, :bv2, :bv3) asc",
+		outbv: map[string]*querypb.BindVariable{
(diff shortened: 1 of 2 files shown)


go/vt/sqlparser/normalizer_test.go
```diff
@@ -179,6 +179,20 @@ func TestNormalize(t *testing.T) {
 		in:      "select a, b from t order by 1 asc",
 		outstmt: "select a, b from t order by 1 asc",
 		outbv:   map[string]*querypb.BindVariable{},
+	}, {
+		// GROUP BY column_position
+		in:      "select a, b from t group by 1",
+		outstmt: "select a, b from t group by 1",
+		outbv:   map[string]*querypb.BindVariable{},
+	}, {
+		// ORDER BY column_position
```
dbussink, line 188: I think this comment is not correct?

go/vt/sqlparser/normalizer_test.go
```diff
@@ -228,6 +242,13 @@ func TestNormalize(t *testing.T) {
 		outbv: map[string]*querypb.BindVariable{
 			"bv1": sqltypes.TestBindVariable([]any{1, "2"}),
 		},
+	}, {
+		// EXPLAIN queries
```
dbussink, line 246: Did this case already work correctly and is this just to add test coverage?

go/vt/sqlparser/normalizer_test.go
```diff
@@ -282,6 +303,14 @@ func TestNormalize(t *testing.T) {
 		outbv: map[string]*querypb.BindVariable{
 			"bv1": sqltypes.ValueBindVariable(sqltypes.MakeTrusted(sqltypes.Datetime, []byte("2022-08-06 17:05:12"))),
 		},
+	}, {
+		// TimestampVal should also be normalized
```
dbussink, line 307: Similar here, to ensure additional test coverage?

dbussink (approved):
A small comment nit and some questions, but nothing blocking.

go/vt/sqlparser/normalizer_test.go
systay: Yep

go/vt/sqlparser/normalizer_test.go
systay: Correct

go/vt/sqlparser/normalizer_test.go
systay: you are correct that the comment is incorrect

Issue #11632, opened 2022-11-03
Bug Report: during reshards, vtexplain gives inconsistent (and sometimes wrong) results


ejortegau:
### Overview of the Issue

Say you are in the middle of a reshard for the keyrange `80-` into `80-a0`, `a0-c0`, `c0-e0` and `e0-`. Both source and target shards already exist and the vreplication workflow is working between them. When you run `vtexplain` for a query that falls in, for example, keyrange `c0-e0`, the output will sometimes indicate that the query will be sent to target shard `80-` and other times that it will be sent to source shard `80-`. Example:

![image](https://user-images.githubusercontent.com/5791035/199701262-72750e00-209c-4fb8-aff5-c275b7dabf6a.png)

### Reproduction Steps

1. Deploy this schema
```sql
CREATE TABLE `sessions_cookies` (
  `global_session_id` bigint(20) unsigned NOT NULL,
  `team_id` bigint(20) unsigned NOT NULL,
  `user_id` bigint(20) unsigned NOT NULL,
  `team_session_id` bigint(20) unsigned NOT NULL,
  PRIMARY KEY (`global_session_id`,`team_id`,`user_id`),
  KEY `team_id_idx` (`team_id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 ROW_FORMAT=COMPRESSED;
```
2. Deploy this vschema:
```json
{
  "auth1": {
    "sharded": true,
    "vindexes": {
      "hash": {
        "type": "hash"
      },
      "md5": {
        "type": "unicode_loose_md5"
      }
    },
    "tables": {
      "sessions_cookies": {
        "columnVindexes": [
          {
            "column": "global_session_id",
            "name": "hash"
          }
        ]
      }
    }
  }
}
```
3. Use this topology shard map by keyspace for `vtexplain`'s `--ks-shard-map-file` argument - which includes both source and target shards of a split:
```json
{
  "auth1": {
    "-20": {
      "primary_alias": {
        "cell": "us_east_1e",
        "uid": 300233160
      },
      "primary_term_start_time": {
        "seconds": 1666657906,
        "nanoseconds": 728234238
      },
      "key_range": {
        "end": "IA=="
      },
      "is_primary_serving": true
    },
    "20-40": {
      "primary_alias": {
        "cell": "us_east_1e",
        "uid": 300215096
      },
      "primary_term_start_time": {
        "seconds": 1666657724,
        "nanoseconds": 58961392
      },
[...]

PR #11636, opened 2022-11-03
VTexplain topology only uses serving shards

ejortegau:
## Description
This addresses issue https://github.com/vitessio/vitess/issues/11632 , which
causes vtexplain to sometimes give bad results if the keyspace is being
resharded, because sometimes it picks source shards and other times target
shards, for routing the query.

The issue is that the `VTExplain.buildTopolog()` adds both source and
destination shards to the map that holds shards per keyspace, when only one of
them is actually serving traffic at any point in time. Later on, vtexplain
loops over this map. Because looping over the map gives a non-deterministic
order, sometimes the results are correct, and sometimes incorrect - that is,
sometimes it gives the result of the shard that is serving, and other times,
the shard that is not serving.

This change ensures that only the serving shards are added to the shards per
keyspace map, thus avoiding the incorrect vtexplain.

Signed-off-by: Eduardo J. Ortega U. <eduardo.ortega@slack-corp.com>

## Related Issue(s)
https://github.com/vitessio/vitess/issues/11632

## Checklist

## Deployment Notes

N/A

### DIFF
--- a/go/vt/vtexplain/vtexplain_vtgate.go
+++ b/go/vt/vtexplain/vtexplain_vtgate.go
@@ -131,6 +131,14 @@ func (vte *VTExplain) buildTopology(opts *Options, vschemaStr string, ksShardMap
 		vte.explainTopo.KeyspaceShards[ks] = make(map[string]*topodatapb.ShardReference)
 
 		for _, shard := range shards {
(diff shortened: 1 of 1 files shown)


ejortegau:
Closing because I realized I am using the wrong workflow. Will open another one

PR #11653, opened 2022-11-07
Fix `OR 1=0` causing queries to scatter

### LINKED ISSUE #11658: Feature Request: Fix `OR 1=0` causing queries to scatter
### Feature Description

Currently, adding the `OR 1=0` conditional to a _simple_ query always causes it to _scatter_, even though the expression is completely redundant.

The feature request is for an optimisation that detects and removes such expressions, so that they don't cause the query to _scatter_ unnecessarily.

### Use Case(s)

ActiveRecord, Ruby on Rails' ORM, translates the following ruby code
```ruby
Things.where(id: [])
```

into the following SQL:
```sql
SELECT * FROM things WHERE 1=0
```

So if the expression is _or_'d in ruby:
```ruby
Things.where(id: []).or(Things.where(id: 1))
```

we end up with
```SQL
SELECT * FROM things WHERE (1=0 OR id=1)
```

pudiva:
## Description

Prevents some meaningless conditionals such as `OR 1=0` from causing queries to scatter.

## Related Issue(s)

* closes #11658

## Checklist

### DIFF
--- a/go/vt/vtgate/planbuilder/operators/route_planning.go
+++ b/go/vt/vtgate/planbuilder/operators/route_planning.go
@@ -337,57 +337,6 @@ func tryRewriteOrToIn(expr sqlparser.Expr) sqlparser.Expr {
 	return nil
 }
 
-func tryRewriteOrFalse(expr sqlparser.Expr) sqlparser.Expr {
-	rewrote := false
-	newPred := sqlparser.Rewrite(sqlparser.CloneExpr(expr), func(cursor *sqlparser.Cursor) bool {
-		_, ok := cursor.Node().(*sqlparser.OrExpr)
-		return ok
-	}, func(cursor *sqlparser.Cursor) bool {
-		// we are looking for the pattern WHERE c = 1 OR 1 = 0
-		switch or := cursor.Node().(type) {
-		case *sqlparser.OrExpr:
-			isFalse := func(subExpr sqlparser.Expr) bool {
-				cmp, ok := subExpr.(*sqlparser.ComparisonExpr)
-				if !ok {
-					return false
-				}
-
-				left, ok := cmp.Left.(*sqlparser.Literal)
-				if !ok {
-					return false
-				}
-
-				right, ok := cmp.Right.(*sqlparser.Literal)
-				if !ok {
-					return false
-				}
-
-				switch cmp.Operator {
-				case sqlparser.EqualOp:
-					if left.Val != right.Val {
-						return true
-					}
-				}
-
-				return false
-			}
-
-			if isFalse(or.Left) {
-				rewrote = true
-				cursor.Replace(or.Right)
-			} else if isFalse(or.Right) {
-				rewrote = true
-				cursor.Replace(or.Left)
-			}
-		}
-		return true
-	})
-	if rewrote {
-		return newPred.(sqlparser.Expr)
-	}
-	return nil
-}
-
 func createInfSchemaRoute(ctx *plancontext.PlanningContext, table *QueryTable) (*Route, error) {
 	ks, err := ctx.VSchema.AnyKeyspace()
 	if err != nil {
--- a/go/vt/vtgate/planbuilder/operators/route_planning.go
+++ b/go/vt/vtgate/planbuilder/operators/route_planning.go
@@ -337,6 +337,57 @@ func tryRewriteOrToIn(expr sqlparser.Expr) sqlparser.Expr {
 	return nil
 }
 
+func tryRewriteOrFalse(expr sqlparser.Expr) sqlparser.Expr {
+	rewrote := false
+	newPred := sqlparser.Rewrite(sqlparser.CloneExpr(expr), func(cursor *sqlparser.Cursor) bool {
+		_, ok := cursor.Node().(*sqlparser.OrExpr)
+		return ok
+	}, func(cursor *sqlparser.Cursor) bool {
+		// we are looking for the pattern WHERE c = 1 OR 1 = 0
+		switch or := cursor.Node().(type) {
+		case *sqlparser.OrExpr:
+			isFalse := func(subExpr sqlparser.Expr) bool {
+				cmp, ok := subExpr.(*sqlparser.ComparisonExpr)
+				if !ok {
+					return false
+				}
+
+				left, ok := cmp.Left.(*sqlparser.Literal)
+				if !ok {
+					return false
+				}
+
+				right, ok := cmp.Right.(*sqlparser.Literal)
+				if !ok {
+					return false
+				}
+
+				switch cmp.Operator {
+				case sqlparser.EqualOp:
+					if left.Val != right.Val {
+						return true
+					}
+				}
+
+				return false
+			}
+
+			if isFalse(or.Left) {
+				rewrote = true
+				cursor.Replace(or.Right)
+			} else if isFalse(or.Right) {
+				rewrote = true
+				cursor.Replace(or.Left)
+			}
+		}
+		return true
+	})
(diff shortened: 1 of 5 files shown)


pudiva:
☝️ @arthurschreiber

go/vt/vtgate/planbuilder/operators/route_planning.go
```diff
@@ -333,6 +333,57 @@ func tryRewriteOrToIn(expr sqlparser.Expr) sqlparser.Expr {
...
+				if !ok {
+					return false
+				}
+
+				switch cmp.Operator {
+				case sqlparser.EqualOp:
+					if left.Val != right.Val {
+						return true
+					}
+				}
```
frouioui, line 366: I think perhaps we should use the evalengine's Evaluate method to evaluate more comparisons. That way we would be able to support more than just equalities.

WDYT @systay

frouioui (approved):
Hello @pudiva, thank you for your contribution! Your Pull Request looks good to me. We do require PR authors to open an issue that documents enhancements:
```
New features should have a link to a feature request issue or an RFC that documents the use cases, corner cases and test cases.
```
Can you create issue and link it to your Pull Request please?

go/vt/vtgate/planbuilder/operators/route_planning.go
pudiva: @frouioui yes I was hoping someone would say that! Can you point me to where this method is defined?

frouioui: For sure. First you'll want to use `evalengine.Translate` to translate the `sqlparser.Expr` into an evalengine expression. Then, you'll need to create and envalengine environment using `evalengine.EmptyExpressionEnv()`, and finally call the `Evaluate` method on your environment with the comparison expression.

It gives us the following snippet:

```go
func TestEvalEngineCmp(t *testing.T) {
	predicate := &sqlparser.ComparisonExpr{
		Operator: sqlparser.EqualOp,
		Left:     sqlparser.NewIntLiteral("1"),
		Right:    sqlparser.NewIntLiteral("1"),
	}

	pred, err := evalengine.Translate(predicate, &dummyTranslator{})
	require.NoError(t, err)

	env := evalengine.EmptyExpressionEnv()
	res, err := env.Evaluate(pred)
	require.NoError(t, err)

	t.Log(res.String()) // will print "INT64(1)"
}
```

Note that you will need to `Translate` the predicate using the semantic table that is available in the planning context.

pudiva:
Created feature request issue
* #11658

go/vt/vtgate/planbuilder/operators/route_planning.go
pudiva: Okay I found `Evaluate()` but I'm not sure how I can use it with a `sqlparser.Expr`... trying to find out how ~ help is welcome. 🙂

frouioui: @pudiva, can you see this comment: https://github.com/vitessio/vitess/pull/11653#discussion_r1015720711? I am happy to explain again otherwise 😃

pudiva: sorry your comments didn't show up here! 👀

frouioui: No worries! Let me know if I am missing something in the original response for your problem

systay:
Hi @pudiva!

I think this is a great idea, but I'd like to do it slightly differently. 
We should do this as early as possible - why not do this in the `earlyRewriter.go` instead of doing it only after we ended up in a scatter?

Also, there is one important issue to overcome - most queries are normalized, which means that literals are replaced with bindvar arguments. This works fine in our planner tests, because we have turned off normalization for these tests.

I have a few ideas on how we could solve it, but I'd love to hear your take on it.

frouioui (commented):
The new code looks good to me @pudiva

pudiva:
> The new code looks good to me @pudiva

Thank you @frouioui! What do you think about that normalisation that @systay mentioned?

pudiva:
Ohhh btw I'm wrapping up the `evalengine.Evaluate()` thing. Should push soon. 🙂

frouioui:
Since normalisation happens after early rewriting, I don't think we need to worry about this at the moment. The query we normalise will be the product of the early rewriting and other early planning steps. So if we change `OR`s in the query the normaliser will not care.

@systay and I had a discussion earlier today about [constant folding](https://en.wikipedia.org/wiki/Constant_folding), we would like, in a future PR, to take what you've done to a higher and more global level. Meaning that everything that can be simplified should be simplified as early as possible during planning. For instance `1+1+1` should become `3`, `x OR 1 = 0` should become `x`, etc. We are still in the process or figuring out how to achieve this properly in the early rewriter.

systay (approved):

systay:
> Oh yes, one more thing: would it be possible to backport this to v15?

No, I wouldn't think so. This is not a bug fix per se, and we only backport bugfixes. It's so easy to accidentally introduce new bugs when doing seemingly innocent improvements, that we have that as policy.

Thank you for the time spent on this, and welcome, brave new Vitess contributor! :)

GuptaManan100 (approved):

harshit-gangal (approved):

harshit-gangal:
can you rebase from mail, some CI checks are not running

pudiva:
Rebased from main ✨

pudiva:
One of the checks is failing with this:

```
2022-11-11T15:28:41.6677115Z     onlineddl_revert_test.go:890: 
2022-11-11T15:28:41.6677513Z         	Error Trace:	onlineddl_revert_test.go:890
2022-11-11T15:28:41.6678105Z         	            				onlineddl_revert_test.go:363
2022-11-11T15:28:41.6679192Z         	Error:      	"CREATE ALGORITHM=UNDEFINED DEFINER=`vt_dba`@`localhost` SQL SECURITY DEFINER VIEW `view_test` AS select 'success_create' AS `msg` from `view_base_table_test`" does not contain "success_replace"
2022-11-11T15:28:41.6679823Z         	Test:       	TestSchemaChange/revert_ALTER_VIEW_where_view_exists
```

Do you think it has anything to do with my patch? I doesn't seem to me...

harshit-gangal:
CI is sometimes a pain, can you push an empty commit to run the queued CI checks?

frouioui:
@pudiva, the `Static Code Checks, Etc. / Static Code Checks Etc (pull_request)` workflow is failing during the `go mod tidy` sub-test. Can you run `go mod tidy` and push the changes? This will make the test go green.

https://github.com/vitessio/vitess/actions/runs/3463419368/jobs/5783625186#step:23:47

pudiva:
> Can you run `go mod tidy` and push the changes?

@frouioui just run `go mod tidy`? I ran it but it produced no changes.

frouioui:
> > Can you run `go mod tidy` and push the changes?
> 
> @frouioui just run `go mod tidy`? I ran it but it produced no changes.

I just saw that @pudiva. After I merged `main` into your branch locally and ran `go mod tidy` it produced a change. Let's try to update your branch with the latest changes on `main` and then update `go.sum` by running `go mod tidy`

frouioui:
I believe the commit that is used by the CI is a merge commit of your branch + `vitessio:main`

pudiva:
Okay I'll rebase and push then

frouioui:
> Okay I'll rebase and push then

I think you might still need to run `go mod tidy` before pushing 😃

pudiva:
finally!!1 ✨

PR #11656, opened 2022-11-07
Send errors in stream instead of a grpc error from streaming rpcs when transaction or reserved connection is acquired

harshit-gangal:
## Description

The old streaming execution that started a transaction or a reserved connection does not propagate the execution error message back. The change done in https://github.com/vitessio/vitess/pull/11592 was to fix the error propagation. The change resulted in missing propagating the open transaction or reserved connection information back in case of failure.

This PR handles those cases and adds additional tests/asserts to ensure it fixes those issues.

## Related Issue(s)

- Fixes https://github.com/vitessio/vitess/issues/11674

## Checklist

-   [X] "Backport me!" label has been added if this change should be backported 15.0, 14.0, 13.0

### DIFF
--- a/go/test/endtoend/reparent/prscomplex/main_test.go
+++ b/go/test/endtoend/reparent/prscomplex/main_test.go
@@ -24,6 +24,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
 	"vitess.io/vitess/go/mysql"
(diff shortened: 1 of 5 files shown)


dbussink (approved):
This was a regression introduced in the bug fix in #11592 where a different bug was fixed. The changes here make sure but the original bug from #11592 doesn't happen anymore, but also the regression is fixed and we have test coverage now for it as well.

GuptaManan100 (approved):

shlomi-noach (approved):

deepthi:
@harshit-gangal do we have an issue? If not can you please create one that describes the bug and link it in?

harshit-gangal:
> @harshit-gangal do we have an issue? If not can you please create one that describes the bug and link it in?

I have added the issue. Would still need ACK to merge this PR.

deepthi (approved):

deepthi:
Still needs code owner approval from @systay 
Should we add @GuptaManan100 as another code owner for the grpc packages?

GuptaManan100:
I don't mind, if that will help.

systay (approved):

PR #11663, opened 2022-11-08
Online DDL: normalize/idempotentize CHECK CONSTRAINTs in ALTER TABLE statement

shlomi-noach:
## Description

Complementing https://github.com/vitessio/vitess/pull/10638, the `vitess` Online DDL algorithm identifies `ADD CONSTRAINT` clauses in `ALTER TABLE`  statements, and manipulates the constraint name. 

This manipulation is consistent with the similar manipulation done in #10638
But more importantly it addresses the specific case where the constraint is created unnamed, via e.g. `alter table t add constraint check (id != 5)`. In this type of query, MySQL autogenerates a name (e.g. `t_chk_1` in this example). This leads to risks because the name is susceptible to `RENAME` statement behavior, which attempts to rename the constraint according to the new table's name.

With this PR, the `ALTER` is manipulated such that the constraint gets an almost-guaranteed unique name, which is also idempotent. Thus, a multishard migration evaluates the same check constraint name for all shards.

Added unit tests, which are sufficient.

## Related Issue(s)

Tracking: https://github.com/vitessio/vitess/issues/6926

## Checklist


go/vt/vttablet/onlineddl/executor_test.go
```diff
@@ -120,6 +120,18 @@ func TestValidateAndEditAlterTableStatement(t *testing.T) {
 			alter:  "alter table t add fulltext key name0_ft (name0), add column i int, add fulltext key name1_ft (name1), add fulltext key name2_ft (name2)",
 			expect: []string{"alter table t add fulltext key name0_ft (name0), add column i int", "alter table t add fulltext key name1_ft (name1)", "alter table t add fulltext key name2_ft (name2)"},
 		},
+		{
+			alter:  "alter table t add constraint check (id != 1)",
+			expect: []string{"alter table t add constraint _aulpn7bjeortljhguy86phdn9 check (id != 1)"},
```
dbussink, line 125: We don't want to keep the default mysql name prefix here if it's short enough to fit? Or the naming as we do in the next test to stay consistent? Not a huge deal, but it starting with `_` seems a bit "off".

shlomi-noach: done

dbussink (approved):
CI should go green for `Cluster (vttablet_prscomplex)` if you rebase on latest `main`.

Issue #11667, opened 2022-11-08
Dead links in `test/config.json` leading to CI potential failures


frouioui:
### Overview of the Issue

Found this issue while working on vitessio/vitess#11244, we have "dead links" in our `test/config.json` file, meaning that we are trying to test a package that no longer exists in our codebase, which lead to a CI failure in some cases. The following output comes from CI running on `release-14.0` where we `tabletgateway/cellalias` was called in `test/config.json` but the package did not exist in the codebase:

```
running tests for  vitess.io/vitess/go/test/endtoend/tabletgateway/cellalias
no required module provides package vitess.io/vitess/go/test/endtoend/tabletgateway/cellalias; to add it:
	go get vitess.io/vitess/go/test/endtoend/tabletgateway/cellalias
2022/11/08 17:22:52 local.tabletgateway_cellalias: saving test output to _test/20221108-171119.3654/local.tabletgateway_cellalias-1.1.log
2022/11/08 17:22:52 local.tabletgateway_cellalias: FAILED (try 1/1) in 300ms: exit status 1
```

The cluster started failing only after I attempted to remove launchable from the workflows. I am still unsure why launchable would prevent such failures from happening, but in order to keep our CI light and avoid dead code we should remove those dead links.

### Reproduction Steps

n/a

### Binary Version

```sh
Must be done on all branches.
```

### Operating System and Environment details

```sh
n/a
```

PR #11668, opened 2022-11-08
Addition of a CI tool to detect dead links in test/config.json

frouioui:
## Description

This Pull Request fixes the issue detailed in https://github.com/vitessio/vitess/issues/11667. It removes the tests in `test/config.json` that no longer exists and adds a go tool to automatically check that there is no dead links. This tool has been added to the CI workflow `Static Code Checks Etc` to automatically prevent the merge of dead links in `test/config.json`.

We most likely need to do a manual backport on all branches.

## Related Issue(s)

- Fixes https://github.com/vitessio/vitess/issues/11667

## Checklist


GuptaManan100 (approved):
I like the change.

systay (approved):

test/config.json
```diff
@@ -720,15 +720,6 @@
 			"RetryMax": 1,
 			"Tags": []
 		},
-		"vtgate_mysql80_derived": {
```
deepthi, line 723: The reason this became a dead link is that the test was moved from `derived` to `queries` and then the directory was deleted.
However, the test has not been added to config.json with the new path, which means that this test has actually been lost from CI. It could be broken and we won't know.
PR where original change happened: #11379 
I would have expected that this test failed on that PR before it was merged, but I haven't checked.
In any case, instead of deleting this, we should be renaming it to vtgate_queries_derived and fixing the path.

deepthi (dismissed):

test/config.json
frouioui: That's a good catch! I'll fix that.

However, I will leave the tool as it is, it simply reports what are the dead links, then it is up to the user to delete or rename the cluster path.

deepthi: Agreed, No changes to the tool, just to config.json.
Once that is done, you may dismiss my review and merge this PR.

rsajwani:
Still some test failures are there which seems related to this PR..

deepthi:
> Still some test failures are there which seems related to this PR..

It seems to me that the `derived` test has not been running for some time in CI and is possibly broken for real.
cc @systay

frouioui:
> It seems to me that the derived test has not been running for some time in CI and is possibly broken for real.

Should be fixed now. The clean up code was missing a table, a test was flaky, and a test is now skipped and will be fixed later (https://github.com/vitessio/vitess/issues/11703).

mattlord (approved):
❤️

deepthi (approved):

PR #11670, opened 2022-11-08
VReplication:  escape identifiers when renaming source tables

WilliamLu99:
## Description

When completing a MoveTables operation with `--rename-tables`, the db and tables are not escaped. This causes the operation to fail in a variety of scenarios, including when the db name has dashes in its name.

I'm not sure if this should be backported. It's not a breaking bug, just an inconvenience. In the end, we just ran the complete command with `--keep-data` and manually renamed the tables. 

Credit goes to @hubb for the fix!
## Checklist


WilliamLu99:
Took a few tries but was able to get CI passing in the forked repo.

https://github.com/Shopify/vitess/pull/56

mattlord (approved):
LGTM. Thank you @WilliamLu99 and @hubb ! ❤️

mattlord:
@WilliamLu99 would you mind merging in latest upstream main and pushing again? The `Run endtoend tests on Cluster (vtgate_general_heavy)` CI workflow seems to be stuck and in this case I would typically merge in latest upstream and push that merge commit to make sure that I have the latest workflow/test changes and to kick the CI, but I cannot in this case as the branch upstream is your fork. Thanks!

Issue #4127, opened 2018-08-09
SET TRANSACTION ISOLATION LEVEL is not session scope


danieltahara:
https://github.com/vitessio/vitess/blob/master/go/vt/sqlparser/analyzer.go#L278 specifies the defaults scope as session.

For SET TRANSACTION statements, this is incorrect. Writing a statement without global or session means it applies only to NEXT transaction, and not all txns in a session.

Functionally, you can't specify @@session.isolation level, but you can for rw. We need to delineate the two.

https://dev.mysql.com/doc/refman/5.7/en/set-transaction.html
```Without any SESSION or GLOBAL keyword, the statement applies to the next (not started) transaction performed within the current session. Subsequent transactions revert to using the SESSION isolation level.```

**Comment by danieltahara** [comment]
Also @sougou points out the wiring is probably not right from vtgate.

Additionally, vtqueryservice wasn't updated in original diff.

**Comment by jakon89** [comment]
Hi,

I just hit similar issue that can be reproduced by running below SQL:
```sql
CREATE TABLE test (id INT(6) UNSIGNED AUTO_INCREMENT PRIMARY KEY,firstname VARCHAR(30) NOT NULL);

insert into test (firstname) values ('xx');

SET autocommit=1;
select @@session.tx_read_only;
set session transaction read only;
SET autocommit=0;
select * from test;
commit;
insert into pod_central_p1.test (firstname) values ('adam');
commit;

select @@session.tx_read_only;
```

The issue is that it doesn't work when running against MySQL 5.7 (because `set session transaction read only;` + `insert`) however it works fine when run on Vites.

This wouldn't be an issue under normal circumstances but I am using JdbcTemplate from Spring and it doesn't like the fact that `select @@session.tx_read_only;` returns `0` instead of `1` after executing `set session transaction read only;`.


Any thoughts on that?

PR #11673, opened 2022-11-09
add support for transaction isolation level and make it vitess aware setting

harshit-gangal:
## Description

This PR moves `transaction_isolation` (`tx_isolation` removed in 8.0) from reserved connection system settings to vitess aware system setting.
This means that Vitess handles this internally rather than passing it down to MySQL straight forwards via reserving a connection for the session.
This PR also adds support for additional transaction isolation level syntax: `set transaction isolation level <isolation_level>`

Vitess stores this information on its internal connection session cache and sets it correctly when a transaction starts on MySQL.

MySQL support 3 kinds of scope for such transaction system variable.
1. Global - if set then any new connection will have those new settings.
2. Session - only the current connection settings are changed.
3. Next Transaction - only the next transaction settings are changed. Resets for the next one.

Vitess, when supported the above system setting did not differentiate between the Session and Next Transaction scope.
To keep the backward compatibility, this behavior remains as-is, but a warning will be reported when Vitess does this scope change from `next transaction` to `session`.

## Related Issue(s)

- Closes https://github.com/vitessio/vitess/issues/4127

## Checklist


go/vt/sqlparser/sql.y
```diff
@@ -1009,56 +1003,98 @@ set_statement:
...
 | READ UNCOMMITTED
   {
-    $$ = ReadUncommitted
+    $$ = "read-uncommitted"
   }
 | SERIALIZABLE
   {
-    $$ = Serializable
+    $$ = "serializable"
   }
```
GuptaManan100, line 1098: Instead we should use constants like `RepeatableReadStr` etc.

go/vt/sqlparser/ast_format.go
```diff
@@ -2760,9 +2718,15 @@ func (node *Variable) Format(buf *TrackedBuffer) {
 	case VariableScope:
 		buf.literal("@")
 	case SessionScope:
+		if node.Name.EqualString("transaction_isolation") || node.Name.EqualString("transaction_read_only") {
```
GuptaManan100, line 2721: We should make constants for these too like `TransactionIsolationStr` etc.

go/vt/sqlparser/ast_format.go
```diff
@@ -2760,9 +2718,15 @@ func (node *Variable) Format(buf *TrackedBuffer) {
...
+			buf.astPrintf(node, "@@%s.", node.Scope.ToString())
```
GuptaManan100, line 2722: Why do we need to print the session scope specifically for the transaction case?

go/vt/sqlparser/ast_format.go
```diff
@@ -2760,9 +2718,15 @@ func (node *Variable) Format(buf *TrackedBuffer) {
...
+			break
+		}
 		buf.literal("@@")
 	case GlobalScope, PersistSysScope, PersistOnlySysScope:
 		buf.astPrintf(node, "@@%s.", node.Scope.ToString())
+	case NoScope:
+		buf.literal("@@")
```
GuptaManan100, line 2729: This doesn't seem correct to me. From MySQL docs - 
```
You can set transaction characteristics globally, for the current session, or for the next transaction only:

With the GLOBAL keyword:

The statement applies globally for all subsequent sessions.

Existing sessions are unaffected.

With the SESSION keyword:

The statement applies to all subsequent transactions performed within the current session.

The statement is permitted within transactions, but does not affect the current ongoing transaction.

If executed between transactions, the statement overrides any preceding statement that sets the next-transaction value of the named characteristics.

Without any SESSION or GLOBAL keyword:

The statement applies only to the next single transaction performed within the session.

Subsequent transactions revert to using the session value of the named characteristics.
```

So if no scope is provided, we shouldn't change it to session scope

GuptaManan100: I see that we disallow no scope in vtgate later, but even then, we `sqlparser` output should be correct in the sense that the scope of the query isn't changed implicitly. This is because `sqlparser` as a package is used outside Vitess too

go/vt/sqlparser/ast_format.go
harshit-gangal: without `session` they have meaning for next transaction scope only. I will add a new scope to explicitly call that out.

go/vt/sqlparser/ast_format.go
harshit-gangal: I have to add a nextTxScope to explicitly call that out. I just wanted to you have a quick look that it is not insane to change setTransaction to setExpr type

go/vt/sqlparser/ast_format.go
harshit-gangal: because `@@` is different than `@@session.` in transaction related variables.

systay (approved):

systay:
this probably needs docs + release notes

PR #11675, opened 2022-11-09
schemadiff: normalize index option value (string)

shlomi-noach:
## Description

With this PR `schemadiff` normalizes (to lower case) the value of an index option. For example: 

`fulltext key name_ft(name) WITH PARSER NGRAM` is normalized into
`fulltext key name_ft(name) WITH PARSER ngram`

## Related Issue(s)

Tracking: https://github.com/vitessio/vitess/issues/10203

## Checklist

### DIFF
--- a/go/vt/schemadiff/table.go
+++ b/go/vt/schemadiff/table.go
@@ -482,6 +482,7 @@ func (c *CreateTableEntity) normalizeIndexOptions() {
 		idx.Info.Type = strings.ToLower(idx.Info.Type)
 		for _, opt := range idx.Options {
 			opt.Name = strings.ToLower(opt.Name)
+			opt.String = strings.ToLower(opt.String)
 		}
 	}
 }
--- a/go/vt/schemadiff/table_test.go
+++ b/go/vt/schemadiff/table_test.go
@@ -443,7 +443,7 @@ func TestCreateTableDiff(t *testing.T) {
 			from:  "create table t1 (id int primary key, name tinytext not null)",
 			to:    "create table t1 (id int primary key, name tinytext not null, fulltext key name_ft(name) with parser ngram)",
 			diff:  "alter table t1 add fulltext key name_ft (`name`) with parser ngram",
-			cdiff: "ALTER TABLE `t1` ADD FULLTEXT KEY `name_ft` (`name`) WITH PARSER NGRAM",
+			cdiff: "ALTER TABLE `t1` ADD FULLTEXT KEY `name_ft` (`name`) WITH PARSER ngram",
 		},
 		{
 			name:  "add one fulltext key and one normal key",
@@ -1712,7 +1712,7 @@ func TestNormalize(t *testing.T) {
 		{
 			name: "does not drop non-default index type",
 			from: "create table t (id int primary key, i1 int, key i1_idx(i1) using hash)",
-			to:   "CREATE TABLE `t` (\n\t`id` int PRIMARY KEY,\n\t`i1` int,\n\tKEY `i1_idx` (`i1`) USING HASH\n)",
+			to:   "CREATE TABLE `t` (\n\t`id` int PRIMARY KEY,\n\t`i1` int,\n\tKEY `i1_idx` (`i1`) USING hash\n)",
 		},
 		{
 			name: "drops default index visibility",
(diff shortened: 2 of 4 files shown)


go/vt/schemadiff/table.go
```diff
@@ -482,6 +482,7 @@ func (c *CreateTableEntity) normalizeIndexOptions() {
 		idx.Info.Type = strings.ToLower(idx.Info.Type)
 		for _, opt := range idx.Options {
 			opt.Name = strings.ToLower(opt.Name)
+			opt.String = strings.ToLower(opt.String)
```
dbussink, line 485: Hmm, would this always lowercase things if you have for example a comment set? It seems like we should not normalize that then. I wonder if we should make this specific only to `WITH PARSER`?

go/vt/schemadiff/table_test.go
```diff
@@ -438,6 +438,20 @@ func TestCreateTableDiff(t *testing.T) {
 			diff:  "alter table t1 add fulltext key name_ft (`name`)",
 			cdiff: "ALTER TABLE `t1` ADD FULLTEXT KEY `name_ft` (`name`)",
 		},
+		{
+			name:  "add one fulltext key with explicit parser",
+			from:  "create table t1 (id int primary key, name tinytext not null)",
+			to:    "create table t1 (id int primary key, name tinytext not null, fulltext key name_ft(name) with parser ngram)",
+			diff:  "alter table t1 add fulltext key name_ft (`name`) with parser ngram",
+			cdiff: "ALTER TABLE `t1` ADD FULLTEXT KEY `name_ft` (`name`) WITH PARSER NGRAM",
```
dbussink, line 446: I think we also want to / need to change the canonical output format of this. It looks like the parser formatter always upcases this. I think we should add something like the following as well:

```patch
diff --git i/go/vt/sqlparser/ast_format.go w/go/vt/sqlparser/ast_format.go
index e94a1b24ab..65756a46a9 100644
--- i/go/vt/sqlparser/ast_format.go
+++ w/go/vt/sqlparser/ast_format.go
@@ -826,7 +826,7 @@ func (idx *IndexDefinition) Format(buf *TrackedBuffer) {
        for _, opt := range idx.Options {
                buf.astPrintf(idx, " %s", opt.Name)
                if opt.String != "" {
-                       buf.astPrintf(idx, " %s", opt.String)
+                       buf.astPrintf(idx, " %#s", opt.String)
                } else if opt.Value != nil {
                        buf.astPrintf(idx, " %v", opt.Value)
                }
diff --git i/go/vt/sqlparser/tracked_buffer_test.go w/go/vt/sqlparser/tracked_buffer_test.go
index 02ea192a5d..614a215c0e 100644
--- i/go/vt/sqlparser/tracked_buffer_test.go
+++ w/go/vt/sqlparser/tracked_buffer_test.go
@@ -220,6 +220,10 @@ func TestCanonicalOutput(t *testing.T) {
                        "select char(77, 121, 83, 81, '76' using utf8mb4) from dual",
                        "SELECT CHAR(77, 121, 83, 81, '76' USING utf8mb4) FROM `dual`",
                },
+               {
+                       "create table t1 (id int primary key, name tinytext not null, fulltext key name_ft(name) with parser ngram)",
+                       "CREATE TABLE `t1` (\n\t`id` int PRIMARY KEY,\n\t`name` tinytext NOT NULL,\n\tFULLTEXT KEY `name_ft` (`name`) WITH PARSER ngram\n)",
+               },
        }
 
        for _, tc := range testcases {
```

go/vt/schemadiff/table.go
shlomi-noach: Per our parser:

https://github.com/vitessio/vitess/blob/1ace3b47046a40e4c43b1ee0d2088db01f35ef0d/go/vt/sqlparser/sql.y#L2299-L2332

It looks like the `String` value is only ever used (at this time?) in a `WITH PARSER` clause. No other index clause/option utilizes it.

shlomi-noach: hmmm also here:

https://github.com/vitessio/vitess/blob/1ace3b47046a40e4c43b1ee0d2088db01f35ef0d/go/vt/sqlparser/sql.y#L7240-L7244

`USING ...`, so this affects e.g. `USING HASH` -> `USING hash`

go/vt/schemadiff/table_test.go
shlomi-noach: added

shlomi-noach:
Normalization now takes place in both `ast_format` and in `schemadiff`. Only two syntax are affected by this chane:

- `USING ...` for "normal" index types (`USING HASH`, `USING BTREE`)
- `WITH PARSER ...` for fulltext index types (`USING PARSER NGRAM`)

`schemadiff` tests adapted. `sqlparser` tests unchanged and all good.

go/vt/schemadiff/table.go
GuptaManan100: I don't see any issues with lower-casing `USING` or `WITH PARSER` values

GuptaManan100 (approved):

PR #11677, opened 2022-11-09
Leverage pFlag's Changed function to detect user specified flag

mattlord:
## Description

Now that we've moved to [pFlag](https://github.com/spf13/pflag) we can leverage the [`Changed`](https://pkg.go.dev/github.com/spf13/pflag#FlagSet.Changed) function to see if the user specified the flag on the command-line.

### Example
```
tabletTypes := subFlags.String("tablet_types", "in_order:RDONLY,REPLICA,PRIMARY", "Tablet types for source and target")
```

If the user  does not specify the `--tablet_types` flag on the command-line, then [`AddFlag`](https://github.com/spf13/pflag/blob/v1.0.5/flag.go#L841) is not called on the flagSet and thus the `tabletTypes` variable has the address of the string variable created for the string literal "in_order:RDONLY,REPLICA,PRIMARY" there rather than the address of the normalized flag map value. [`Changed`](https://github.com/spf13/pflag/blob/v1.0.5/flag.go#L508) then just checks whether or not the key exists in the normalized flags map.

## Checklist

-   [x] Tests are not required (existing test verifies this behavior)

### DIFF
--- a/go/vt/vtctl/vtctl.go
+++ b/go/vt/vtctl/vtctl.go
@@ -2337,7 +2337,7 @@ func commandVRWorkflow(ctx context.Context, wr *wrangler.Wrangler, subFlags *pfl
 		vrwp.TabletTypes = *tabletTypes
 	case vReplicationWorkflowActionSwitchTraffic, vReplicationWorkflowActionReverseTraffic:
 		vrwp.Cells = *cells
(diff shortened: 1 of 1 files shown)


notfelineit (approved):
LGTM!

rohit-nayak-ps (approved):
lgtm

Issue #11626, opened 2022-11-02
Planner to move to Operator model


systay:
### Internal Refactoring

Since forever, we have had the `logicalPlan` interface that is used during planning. This served us well on the V3 planner.
In the gen4 work, we introduced the `Operator` interface that we use when doing the join ordering and route merging.

Once that is done, we transform the operator tree to a logical plan tree to do the grouping/order by/limit planning.

Unfortunately, this means that in some situations we cannot decide early on if we can merge a route with another or not. For example when using derived tables and UNION, if the inner queries are using aggregation, we can't make a good decision during join ordering, and later when we do the aggregation planning, it's too late to merge the routes, because we are by then on logical plans instead.

What I want to do is:

- [x] Make it possible to mix logical and physical operators in the same tree. This will allow us to transform the tree piece by piece, instead of being forced to do everything at once.
- [x] Move a lot of the operator tree traversal to outside the operators. Today a lot of the code on the operators is concerned with sending on calls to the inputs.
- [x] Add column and offset handling to the operators
- [x] Planning projections https://github.com/vitessio/vitess/pull/12790
- [x] - ordering https://github.com/vitessio/vitess/pull/12954
- [x] - distinct https://github.com/vitessio/vitess/pull/13201
- [x] - limit https://github.com/vitessio/vitess/pull/12927
- [x] - aggregation https://github.com/vitessio/vitess/pull/12994
- [x] - HAVING https://github.com/vitessio/vitess/pull/13289
- [x] - Distinct aggregations (`sum(distinct col))`)
- [x] Derived tables https://github.com/vitessio/vitess/pull/13361
- [x] Subqueries https://github.com/vitessio/vitess/pull/13750
- [x] UNION #13450
- [ ] Make it possible to produce 'engine.Primitive' structs straight from the operator tree without first having to transform them to logicalPlan
Hopefully the refactoring should make the code easier to work with and understand as well.

harshit-gangal:
Closing this as new operator model is used to plan all the queries.
transform to engine primitive is very lean now.

PR #11680, opened 2022-11-10
planner operators refactoring

systay:
## Description
This is part of my ongoing refactoring of the planner operators in Vitess.

In this PR we handle horizon planning for the simplest cases (single sharded route plans) on the operators instead of using the logical plan version of horizon planning.

I'm also changing the rewriter to be easier to read - instead of true and false everywhere, I use named constants - `rewrite.NewTree` and `rewrite.SkipChildren` are easier to read and understand than literal booleans.

## Related Issue(s)
https://github.com/vitessio/vitess/issues/11626


go/vt/vtgate/planbuilder/operators/rewrite/rewriters.go
```diff
@@ -0,0 +1,128 @@
...
+		return op, SameTree, VisitChildren, nil
+	})
+	return err
+}
+
+// BottomUp rewrites an operator tree from the bottom up. BottomUp applies a transformation function to
+// the given operator tree from the bottom up. Each callback [f] returns a TreeIdentity that is aggregated
+// into a final output indicating whether the operator tree was changed.
+func BottomUp(root ops.Operator, f Func) (ops.Operator, error) {
+	op, _, err := bottomUp(root, f)
```
harshit-gangal, line 60: looks like the comment should actually move to `bottomUp` function where the logic resides.

go/vt/vtgate/planbuilder/operators/rewrite/rewriters.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+	return op, nil
+}
+
+// TopDown applies a transformation function to the given operator tree from the bottom up. =
+// Each callback [f] returns a TreeIdentity that is aggregated into a final output indicating whether the
+// operator tree was changed.
+// The callback also returns a VisitRule that signals whether the children of this operator should be visited or not
+func TopDown(in ops.Operator, rewriter BreakableFunc) (ops.Operator, error) {
+	op, _, err := breakableTopDown(in, rewriter)
```
harshit-gangal, line 72: similar comment here

harshit-gangal (approved):

go/vt/vtgate/planbuilder/operators/rewrite/rewriters.go
systay: nah, the comment belongs here. the internal `bottomUp` does the actual work, but the description belongs here, I think.

PR #11685, opened 2022-11-10
VTAdmin: display workflow type in workflows list

rohit-nayak-ps:
## Description

Adds the workflow type (and if applicable the workflow sub type) to the current `/workflows` endpoint. 

Workflows   VTAdmin (1)
Workflows   VTAdmin

## Related Issue(s)

https://github.com/vitessio/vitess/issues/11690

## Checklist


go/vt/vtctl/workflow/server.go
```diff
@@ -332,42 +334,44 @@ func (s *Server) GetWorkflows(ctx context.Context, req *vtctldatapb.GetWorkflows
 		span.Annotate("workflow", workflow.Name)
 		span.Annotate("tablet_alias", tablet.AliasString())
 
-		id, err := evalengine.ToInt64(row[0])
+		id, err := evalengine.ToInt64(row["id"])
```
mattlord, line 337: Just a note that with named values the field names are case sensitive and we're not guaranteeing case with `AS` alias clauses in the `SELECT` right now. Should be fine since these are vitess tables and the columns do seem to be all lower case in our defined schema.

mattlord (approved):
LGTM. I'm not sure how I feel about cramming so much into that one UI column w/o labels or anything. We could instead have a workflow type/details column or add `name`,`type` labels to the data in the column. I'll let the VTAdmin leads comment on that stuff though. I'll approve for now, FWIW.

deepthi:
> LGTM. I'm not sure how I feel about cramming so much into that one UI column w/o labels or anything. We could instead have a workflow type/details column OR add `name`,`type` labels to the data in the column. I'll let the VTAdmin leads comment on that stuff though. I'll approve for now, FWIW.

We debated this yesterday while walking through the PR with @notfelineit. @notfelineit and I both prefer adding a column, but @rohit-nayak-ps is concerned about horizontal scrolling.

deepthi (approved):
I'd _prefer_ to see this as a new column, but I'm OK with the current UX.

web/vtadmin/src/components/routes/Workflows.tsx
```diff
@@ -70,6 +72,12 @@ export const Workflows = () => {
                 <tr key={idx}>
                     <DataCell>
                         <div className="font-bold">{href ? <Link to={href}>{row.name}</Link> : row.name}</div>
+                        <div className="text-secondary text-success-200">
+                            {row.workflowType}
+                            <span className="text-sm">
+                                {row.workflowSubType !== 'None' ? ' (' + row.workflowSubType + ')' : ''}
+                            </span>
```
notfelineit, line 79: Here you could conditionally render the entire `span` if the workflow subtype is 'none;
```suggestion
                            {row.workflowSubtype && row.workflowSubtype != 'None' && (
                                <span className="text-sm">
                                    {' (' + row.workflowSubType + ')'}
                                </span>
                            )}
```

notfelineit: It also takes care of the nil check for `row.workflowSubType`!

web/vtadmin/src/components/routes/Workflows.tsx
```diff
@@ -70,6 +72,12 @@ export const Workflows = () => {
                 <tr key={idx}>
                     <DataCell>
                         <div className="font-bold">{href ? <Link to={href}>{row.name}</Link> : row.name}</div>
+                        <div className="text-secondary text-success-200">
+                            {row.workflowType}
```
notfelineit, line 76:
```suggestion
                            {row.workflowType || 'N/A'}
```

notfelineit: ^ Just adding a nil check for `row.workflowType` - feel free to change from 'N/A' if something else is appropriate.

web/vtadmin/src/components/routes/Workflows.tsx
```diff
@@ -70,6 +72,16 @@ export const Workflows = () => {
                 <tr key={idx}>
                     <DataCell>
                         <div className="font-bold">{href ? <Link to={href}>{row.name}</Link> : row.name}</div>
+                        {row.workflowType && (
+                            <div className="text-secondary text-success-200">
+                                {row.workflowType}
```
notfelineit, line 77: For this one we'll still want to handle the null case from [line 57](https://github.com/vitessio/vitess/pull/11685/files#diff-654f842b149e2f895f03d04f85c065655325acfc3d045c19db272c5e69871ebaR57) 😄 
```suggestion
                                {row.workflowType || 'N/A'}
```

rohit-nayak-ps: > For this one we'll still want to handle the null case from [line 57](https://github.com/vitessio/vitess/pull/11685/files#diff-654f842b149e2f895f03d04f85c065655325acfc3d045c19db272c5e69871ebaR57) smile

I thought it would never reach this point (line 77) if `row.workflowType` is null, because of the check on line 75?

notfelineit: Ah I missed that! You're correct. This is great!

notfelineit (approved):

PR #11691, opened 2022-11-11
improve handling of ORDER BY/HAVING rewriting

### LINKED ISSUE #11641: v15 Regression: complex HAVING query - Invalid use of group function
### Overview of the Issue

This was working fine in v14, and works fine against MySQL directly, but started failing after upgrading to v15. This is against an unsharded keyspace.

```sql
select
  loc_id,
  min(created_at) as created_at,
  from_unixtime(min(deadline_at)) as deadline_at,
  count(distinct phrase) total_unique,
  count(*) total_pipelines
from keywordsu.semetrics_grepwords_pipeline
where retrieved_at = 0
  and batch_id = 0
  and api_error = ''
  and provider != 'invalid'
  and location_type not in ('Postal Code', 'Borough', 'Neighborhood', 'National Park', 'City Region')
group by loc_id
having count(*) >= 999
    or min(created_at) + interval 2 hour <= now()
order by total_pipelines desc
limit 100
```

The error is related to the complex `HAVING min(created_at) + interval 2 hour <= now()` clause, as this query runs fine without it.

### Reproduction Steps

Run a query against any table using `HAVING min(created_at) + interval 2 hour <= now()`

### Binary Version

```sh
Docker v15.0.0 for all components
```

```sh
GKE 1.21
```

systay:
## Description
MySQLs scoping rules around HAVING and ORDER BY are not easy to mimic.
The issue here was that we rewrite column names used in `ORDER BY`/`HAVING` if they match an alias introduced in `SELECT`.

For the query:

```sql
select min(id) as id from tbl having id > 10
```

the `HAVING` clause is comparing the aggregate value of `min(id)`, not the column id on the table. This can easily be shown by not introducing the alias:

```sql
> select min(id) from apa having id = 10;

Error: (1054, "Unknown column 'id' in 'having clause'")
```

But if you then wrap the `id` column in an aggregation, somehow that changes the scoping, and the following query is valid in MySQL:

```sql
select min(id) as id from tbl having min(id) > 10
```

Why we want to rewrite the use of `id` outside aggregations but not inside them is beyond me, but this is MySQLs behaviours, from my black box testing of it.

## Related Issue(s)
Fixes #11641

### DIFF
--- a/go/vt/vtgate/semantics/early_rewriter.go
+++ b/go/vt/vtgate/semantics/early_rewriter.go
@@ -53,7 +53,7 @@ func (r *earlyRewriter) down(cursor *sqlparser.Cursor) error {
 			node.Join = sqlparser.NormalJoinType
 			r.warning = "straight join is converted to normal join"
 		}
-	case *sqlparser.Order:
+	case sqlparser.OrderBy:
 		r.clause = "order clause"
 		rewriteHavingAndOrderBy(cursor, node)
 	case sqlparser.GroupBy:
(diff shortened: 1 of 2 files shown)


vmg (approved):

dbussink (approved):

harshit-gangal (approved):

go/vt/vtgate/semantics/early_rewriter.go
```diff
@@ -140,12 +149,18 @@ func rewriteHavingAndOrderBy(cursor *sqlparser.Cursor, node sqlparser.SQLNode) {
 			if !col.Qualifier.IsEmpty() {
 				return false
 			}
+			_, parentIsAggr := inner.Parent().(sqlparser.AggrFunc)
```
nussjustin, line 152: Just skimmed this PR and noticed something:

While this fixes the case of simple aggregations like `sum(foo)` the new code still does not handle more _complex_ aggregations correctly.

For example

```sql
select id, sum(foo) as foo from t1 having sum(foo + 1) > 1
```

This is still rewritten to

```sql
select id, sum(foo) as foo from t1 having sum(sum(foo) + 1) > 1
```

PR #11697, opened 2022-11-12
feat: add support for query timeout from comment directive in shard target mode

harshit-gangal:
## Description

This PR adds support for query timeout in comment directive when using shard targeting.

## Checklist


systay (approved):

deepthi (commented):
Needs an issue + website docs.

PR #11698, opened 2022-11-12
refactoring: Horizon planning on the operators

systay:
## Checklist


go/vt/vtgate/planbuilder/operators/logical.go
```diff
@@ -124,20 +118,22 @@ func createOperatorFromUpdate(ctx *plancontext.PlanningContext, updStmt *sqlpars
...
-		VindexPreds:       vp,
-		TargetDestination: dest,
-	}
+		opCode,
+		vindexTable.Keyspace,
+		dest,
+		vp,
+		nil,
+		nil,
+	)
```
arthurschreiber, line 136: Not a blocker at all, but just my personal opinion: It's quite unfortunate go does not support named arguments. The previous version that build the struct directly was much more readable, as you could see what the individual struct members were set to, where as now `newRoute` takes a bunch of values but it's not obvious what they mean (e.g. those two `nil` arguments at the end).

Issue #11699, opened 2022-11-12
Feature Request: VReplication Defer Secondary Index Creation


mattlord:
### Feature Description

For [`MoveTables`](https://vitess.io/docs/reference/vreplication/movetables/) and  [`Reshard`](https://vitess.io/docs/reference/vreplication/reshard/) , VReplication should defer secondary index creation for a table until after the table has been copied.

As a quick example, you can see ~ 20% improvement for a small dataset this way:
```
cd /tmp
mkdir /tmp/myidxtestdata
git clone https://github.com/mattlord/test_db.git
cd test_db

# Install MySQL 8.0.31: https://dev.mysql.com/downloads/mysql/
mysqld --initialize-insecure --datadir=/tmp/myidxtestdata
mysqld --no-defaults --datadir=/tmp/myidxtestdata &

# Load data with secondary indexes
time mysql -u root < ./employees.sql
# ~ 25 seconds on avg
mysql -u root -e "drop database employees"

# Load data without secondary indexes and add at the end
time mysql -u root < ./employees_deferred_keys.sql
# ~ 20 seconds on avg
mysql -u root -e "drop database employees"

mysqladmin -u root shutdown
rm -rf /tmp/myidxtestdata
```

In practice we've seen upwards of 10x improvement when importing tables that are hundreds of GiBs when creating the secondary indexes after the tables have been copied during a [`MoveTables`](https://vitess.io/docs/reference/vreplication/movetables/).

The one-shot secondary index creation after table copy should be even faster with [MySQL 8.0.31+ due to InnoDB support for parallel index builds](https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-31.html):
> InnoDB: InnoDB now supports parallel index builds, which improves index build performance. In particular, loading sorted index entries into a B-tree is now multithreaded. Previously, this action was performed by a single thread.

### Use Case(s)

* Moving/Importing very large tables
* Resharding very large tables

Note: The value of doing this becomes exponential as the table size and number of secondary indexes grow. We've seen some cases where the time to copy a very large table goes from days to hours.

PR #11700, opened 2022-11-12
VReplication: Defer Secondary Index Creation

mattlord:
## Description

In this PR we add a new _experimental_ `--defer-secondary-keys` flag for the `MoveTables`, `Reshard`, and `Migrate` `vtctl` client commands. When this flag is specified the target table has all secondary indexes removed from its schema prior to copying rows from the source and the indexes are re-added to the schema at the end of the copy phase.

> **Note**: All secondary keys are re-added in a single `ALTER TABLE` statement and [this an an *online*/*in-place* DDL operation](https://dev.mysql.com/doc/refman/8.0/en/innodb-online-ddl-operations.html#online-ddl-index-operations). __BUT__, we execute this `ALTER TABLE` synchronously at this time to ensure the table is in its final expected state before we complete the copy phase for the table and move forward with the workflow. This is important as it means the source will need to retain binary logs for a period beyond how long this statement takes to complete ([see docs](https://dev.mysql.com/doc/refman/en/replication-options-binary-log.html#sysvar_binlog_expire_logs_seconds)) as there will not be another [fast forward and catch-up cycle](https://vitess.io/docs/design-docs/vreplication/life-of-a-stream/) while it runs. Future TODO: try and figure out a safe and robust way to proceed with the workflow (copying next table or entering the running phase) while the `ALTER TABLE` is running while ensuring that all ALTERs are successfully applied before the workflow can proceed beyond the running phase (`SwitchTraffic`).

> **Note**: Foreign keys are not supported generally with VReplication today. Fulltext keys are NOT deferred as they require a full table rebuild to add an internal `FTS_DOC_ID` field to each record and you cannot add/drop more than 1 of them per ALTER statement.

For the WHY, please see [the feature request](https://github.com/vitessio/vitess/issues/11699). 

###  Manual Test and Demonstration:

<details>
  <summary>Click here for details</summary>

```
pushd examples/local
./101_initial_cluster.sh; ./201_customer_tablets.sh;
echo

vtctldclient InitShardPrimary --force commerce/0 zone1-100
echo

command mysql -u root --socket=/opt/vtdataroot/vt_0000000100/mysql.sock  vt_commerce -e "create table sit (id int not null primary key, c1 varchar(100), c2 varchar(200), c3 varchar(200), unique key (c1,c2,c3), key (c1), key (c2))"
command mysql -u root --socket=/opt/vtdataroot/vt_0000000100/mysql.sock  vt_commerce -e "create table sit2 (id int not null, c1 varchar(100), c2 varchar(200), c3 varchar(200), unique key (c1,c2,c3), key (c1), key (c2), primary key (id, c1))"
echo

vtctlclient MoveTables -- --source commerce --all --defer-secondary-keys Create customer.commerce2customer
echo

command mysql -u root --socket=/opt/vtdataroot/vt_0000000200/mysql.sock  vt_customer -e "show create table sit\G show create table sit2\G"
echo
command mysql -v -u root --socket=/opt/vtdataroot/vt_0000000200/mysql.sock  vt_customer --binary-as-hex=false -e "select * from _vt.post_copy_action"
echo

sleep 5
command mysql -u root --socket=/opt/vtdataroot/vt_0000000200/mysql.sock  vt_customer -e "show create table sit\G show create table sit2\G"
echo
command mysql -v -u root --socket=/opt/vtdataroot/vt_0000000200/mysql.sock  vt_customer --binary-as-hex=false -e "select * from _vt.post_copy_action"
echo

vtctlclient workflow customer.commerce2customer show
echo

./401_teardown.sh
popd
```

Final results:
```
       Table: sit
Create Table: CREATE TABLE `sit` (
  `id` int NOT NULL,
  `c1` varchar(100) DEFAULT NULL,
  `c2` varchar(200) DEFAULT NULL,
  `c3` varchar(200) DEFAULT NULL,
  PRIMARY KEY (`id`)
       Table: sit2
Create Table: CREATE TABLE `sit2` (
  `id` int NOT NULL,
  `c1` varchar(100) NOT NULL,
  `c2` varchar(200) DEFAULT NULL,
  `c3` varchar(200) DEFAULT NULL,
  PRIMARY KEY (`id`,`c1`)

--------------
select * from _vt.post_copy_action
--------------

+----+----------+------------+------------------------------------------------------------------------------------------------------------------------+
| id | vrepl_id | table_name | action                                                                                                                 |
+----+----------+------------+------------------------------------------------------------------------------------------------------------------------+
|  1 |        1 | sit2       | {"task": "alter table vt_customer.sit2 add UNIQUE KEY c1 (c1, c2, c3), add KEY c1_2 (c1), add KEY c2 (c2)", "type": 1} |
|  2 |        1 | sit        | {"task": "alter table vt_customer.sit add UNIQUE KEY c1 (c1, c2, c3), add KEY c1_2 (c1), add KEY c2 (c2)", "type": 1}  |
+----+----------+------------+------------------------------------------------------------------------------------------------------------------------+

       Table: sit
Create Table: CREATE TABLE `sit` (
  `id` int NOT NULL,
  `c1` varchar(100) DEFAULT NULL,
  `c2` varchar(200) DEFAULT NULL,
  `c3` varchar(200) DEFAULT NULL,
  PRIMARY KEY (`id`),
  UNIQUE KEY `c1` (`c1`,`c2`,`c3`),
  KEY `c1_2` (`c1`),
  KEY `c2` (`c2`)
       Table: sit2
Create Table: CREATE TABLE `sit2` (
  `id` int NOT NULL,
  `c1` varchar(100) NOT NULL,
  `c2` varchar(200) DEFAULT NULL,
  `c3` varchar(200) DEFAULT NULL,
  PRIMARY KEY (`id`,`c1`),
  UNIQUE KEY `c1` (`c1`,`c2`,`c3`),
  KEY `c1_2` (`c1`),
  KEY `c2` (`c2`)

--------------
select * from _vt.post_copy_action
--------------


{
	"SourceLocation": {
		"Keyspace": "commerce",
		"Shards": [
			"0"
		]
	},
	"TargetLocation": {
		"Keyspace": "customer",
		"Shards": [
			"0"
		]
	},
	"MaxVReplicationLag": 0,
	"MaxVReplicationTransactionLag": 0,
	"Frozen": false,
	"ShardStatuses": {
		"0/zone1-0000000201": {
			"PrimaryReplicationStatuses": [
				{
					"Shard": "0",
					"Tablet": "zone1-0000000201",
					"ID": 1,
					"Bls": {
						"keyspace": "commerce",
						"shard": "0",
						"filter": {
							"rules": [
								{
									"match": "corder",
									"filter": "select * from corder"
								},
								{
									"match": "customer",
									"filter": "select * from customer"
								},
								{
									"match": "product",
									"filter": "select * from product"
								},
								{
									"match": "sit",
									"filter": "select * from sit"
								},
								{
									"match": "sit2",
									"filter": "select * from sit2"
								}
							]
						}
					},
					"Pos": "1d1f22e8-8364-11ed-8489-8259370a8da6:1-27",
					"StopPos": "",
					"State": "Running",
					"DBName": "vt_customer",
					"TransactionTimestamp": 1671870227,
					"TimeUpdated": 1671870231,
					"TimeHeartbeat": 1671870231,
					"TimeThrottled": 0,
					"ComponentThrottled": "",
					"Message": "",
					"Tags": "",
					"WorkflowType": "MoveTables",
					"WorkflowSubType": "None",
					"CopyState": null
				}
			],
			"TabletControls": null,
			"PrimaryIsServing": true
		}
	},
	"SourceTimeZone": "",
	"TargetTimeZone": "",
	"DeferSecondaryKeys": true
}
```
</details>

## Related Issue(s)

  - Fixes: https://github.com/vitessio/vitess/issues/11699

## Checklist

-   [x] Tests were added
-   [x] Documentation: https://github.com/vitessio/website/pull/1349


go/vt/vttablet/tabletmanager/vreplication/vreplicator.go
```diff
@@ -558,3 +570,298 @@ func recalculatePKColsInfoByColumnNames(uniqueKeyColumnNames []string, colInfos
...
+		}
+		if _, err := dbClient.ExecuteFetch(sqlparser.String(alterDrop), 1); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (vr *vreplicator) getTableSecondaryKeys(ctx context.Context, tableName string) ([]*sqlparser.IndexDefinition, error) {
```
shlomi-noach, line 650: I'm just noting that we drop and re-add all types of secondary keys, including `UNIQUE`s. This is fine, because `vreplication` is unique-compliant and does not need a `UNIQUE` constraint on the target table in order to validate anything.

shlomi-noach: However, if you drop multiple `FULLTEXT` indexes, that's a problem, because you cannot add more than one `FULLTEXT` key in a single `ALTER TABLE` statement.

shlomi-noach: It's implied that the source table must not have any foreign keys. That's already the case today, there is no added restriction, so that's fine.

go/vt/vttablet/tabletmanager/vreplication/vreplicator.go
```diff
@@ -558,3 +570,298 @@ func recalculatePKColsInfoByColumnNames(uniqueKeyColumnNames []string, colInfos
...
+	query, err := sqlparser.ParseAndBind(sqlGetPostCopyActions, sqltypes.Uint32BindVariable(vr.id),
+		sqltypes.StringBindVariable(tableName))
+	if err != nil {
+		return err
+	}
+	qr, err := dbClient.ExecuteFetch(query, -1)
+	if err != nil {
+		return err
+	}
+	if qr == nil || len(qr.Rows) == 0 {
```
shlomi-noach, line 711: In what situation would `qr == nil`?

go/vt/vttablet/tabletmanager/vreplication/vreplicator.go
```diff
@@ -558,3 +570,298 @@ func recalculatePKColsInfoByColumnNames(uniqueKeyColumnNames []string, colInfos
...
+	}, parsedDDL)
+
+	if !foundCreateTable {
+		return nil, fmt.Errorf("could not find expected CREATE TABLE statement in DDL: %s", ddl)
+	}
+
+	return secondaryKeys, err
+}
+
+func (vr *vreplicator) execPostCopyActions(ctx context.Context, tableName string) error {
```
shlomi-noach, line 690: I'm trying to understand the behavior in case of error. There could be multiple actions involved here, any one of which could fail with error, if I understand correctly.

1. Will `copyTable` at all ensure to re-call `execPostCopyActions()`?
2. How does `execPostCopyActions()` maintain idempotency of post-copy operations? If it failed mid-way, how does it skip the successful operations and only apply the unapplied operations?
3. Supposedly we could use `ERDupIndex`? I see it in `constants.go` but I don't see where it's being used.

go/vt/vttablet/tabletmanager/vreplication/vreplicator.go
```diff
@@ -558,3 +570,298 @@ func recalculatePKColsInfoByColumnNames(uniqueKeyColumnNames []string, colInfos
...
+	// listen for context cancellation, which would indicate that
+	// the controller is stopping due to engine shutdown (tablet
+	// shutdown or transition). If that happens we attempt to kill
+	// the running ALTER statement using a DBA connection.
+	// If we don't do this then we could e.g. cause a PRS to fail as
+	// the running ALTER will block setting [super_]read_only.
+	// A failed/killed ALTER will be tried again when the copy
+	// phase starts up again on the (new) PRIMARY.
+	var action PostCopyAction
+	done := make(chan struct{})
```
shlomi-noach, line 725: Personal preference, create a `chan bool`. Just because it's much more limited than `struct{}` and because it's clearer.

shlomi-noach: I can't see where `done` is being populated. Looking for `done <- something`.

go/vt/vttablet/tabletserver/vstreamer/testenv/testenv.go
```diff
@@ -182,11 +182,11 @@ func (te *Env) RemoveAnyDeprecatedDisplayWidths(orig string) string {
 	}
 	var adjusted string
 	baseIntType := "int"
-	intRE := regexp.MustCompile(`(i)?int\(([0-9]*)?\)`)
+	intRE := regexp.MustCompile(`(?i)int\(([0-9]*)?\)`)
```
shlomi-noach, line 185: if this definition comes from `information_schema` then it's guaranteed to be lower case.

go/vt/wrangler/materializer.go
```diff
@@ -1310,7 +1311,8 @@ func (mz *materializer) generateInserts(ctx context.Context, targetShard *topo.S
 		}
 		ig.AddRow(mz.ms.Workflow, bls, "", mz.ms.Cell, mz.ms.TabletTypes,
 			int64(mz.ms.MaterializationIntent),
-			int64(workflowSubType))
+			int64(workflowSubType),
+			mz.ms.DeferSecondaryKeys)
```
shlomi-noach, line 1315:
```suggestion
			mz.ms.DeferSecondaryKeys,
		)
```

go/vt/vttablet/tabletmanager/vreplication/vreplicator.go
```diff
@@ -558,3 +570,298 @@ func recalculatePKColsInfoByColumnNames(uniqueKeyColumnNames []string, colInfos
...
+	parsedDDL, err := sqlparser.ParseStrictDDL(ddl)
+	var foundCreateTable bool
+	if err != nil {
+		return secondaryKeys, err
+	}
+
+	err = sqlparser.Walk(func(node sqlparser.SQLNode) (kontinue bool, err error) {
+		switch node := node.(type) {
+		case sqlparser.DDLStatement:
+			foundCreateTable = true
```
shlomi-noach, line 671: not sure about the terminology here. We would any `DDLStatement` imply a `foundCreateTable`?

go/vt/vttablet/tabletmanager/vreplication/vreplicator.go
```diff
@@ -558,3 +570,298 @@ func recalculatePKColsInfoByColumnNames(uniqueKeyColumnNames []string, colInfos
...
+	// shutdown or transition). If that happens we attempt to kill
+	// the running ALTER statement using a DBA connection.
+	// If we don't do this then we could e.g. cause a PRS to fail as
+	// the running ALTER will block setting [super_]read_only.
+	// A failed/killed ALTER will be tried again when the copy
+	// phase starts up again on the (new) PRIMARY.
+	var action PostCopyAction
+	done := make(chan struct{})
+	defer close(done)
+	killAlterStmt := func(alterStmt string) error {
```
shlomi-noach, line 727: Suggesting an alternative: grab the `CONNECTION_ID()` in the connection issuing the `ALTER` statement, before running `ALTER`; then, you may just `KILL QUERY %d` with that connection id. Searching by user & query could be tricky sometimes, e.g. with escaping issues or stripped comments.

go/vt/vttablet/tabletmanager/vreplication/vreplicator.go
```diff
@@ -558,3 +570,298 @@ func recalculatePKColsInfoByColumnNames(uniqueKeyColumnNames []string, colInfos
...
+				// the subsequent DELETE of the post_copy_action record failed.
+				// For example, the context could be cancelled in-between.
+				// It's also possible that the user has modified the schema on
+				// the target side.
+				// If we get a duplicate key/index error then let's see if the
+				// index definitions that we would have added already exist in
+				// the table schema and if so move forward and delete the
+				// post_copy_action record.
+				if sqlErr, ok := err.(*mysql.SQLError); ok && sqlErr.Number() == mysql.ERDupKeyName {
+					stmt, err := sqlparser.ParseStrictDDL(action.Task)
```
shlomi-noach, line 803: ah, OK, I think this answers my earlier question about idempotency. Please confirm.

go/vt/vttablet/tabletmanager/vreplication/vreplicator.go
```diff
@@ -558,3 +570,298 @@ func recalculatePKColsInfoByColumnNames(uniqueKeyColumnNames []string, colInfos
...
+						return failedAlterErr
+					}
+					for _, alterOption := range alterStmt.AlterOptions {
+						addKey, ok := alterOption.(*sqlparser.AddIndexDefinition)
+						if !ok {
+							return failedAlterErr
+						}
+						found := false
+						for _, currentSK := range currentSKs {
+							if reflect.DeepEqual(addKey.IndexDefinition, currentSK) {
```
shlomi-noach, line 825: Is it possible to use `sqlparser.Equals.RefOfIndexDefinition()` instead of `DeepQual`?

shlomi-noach (commented):
Overall this looks good to me. Please see inline comments. `FULLTEXT` is an issue that needs to be resolved. 
I'd change the way the `ALTER` query gets killed, see comments.

go/vt/vttablet/tabletmanager/vreplication/vreplicator.go
mattlord: Great points! I wasn't aware of the full text limitations. It would be quite rare, I think, to have more than one in a table but another limitation I realized is that adding one requires a full table rebuild. So I will explicitly leave them in place:
```
		for _, secondaryKey := range secondaryKeys {
			// Primary should never happen. Fulltext keys are
			// not supported for deferral and retained during
			// the copy phase as it has some unique behaviors
			// and constraints:
			// - Adding a fulltext key requires a full table
			//   rebuild to add the internal FTS_DOC_ID field
			//   to each record.
			// - You can not add/remove multiple fulltext keys
			//   in a single ALTER statement.
			if secondaryKey.Info.Primary || secondaryKey.Info.Fulltext {
				continue
			}
```

go/vt/vttablet/tabletmanager/vreplication/vreplicator.go
mattlord: It should never happen, just an extra check to prevent some unexpected situation from causing a crash. I'll add a comment to note that it should never happen.

go/vt/vttablet/tabletmanager/vreplication/vreplicator.go
mattlord: There can only be a single action per table today — the stash will only store one ALTER per table and there's a unique index on (vrepl_id, table) — but we may remove that limitation in the future. The action(s) happen after we've finished what would be the final copy phase cycle for the table and before we finalize it and move forward with the workflow (copying next table or moving to running phase). If we encounter an error/failure performing the actions then we will do another copy phase cycle for the table and retry any failed actions.

go/vt/vttablet/tabletmanager/vreplication/vreplicator.go
mattlord: The done channel is closed in a defer so the goroutine would end when either the context is cancelled (in which case the context's Done channel is closed) or when the function completes.

go/vt/vttablet/tabletserver/vstreamer/testenv/testenv.go
mattlord: Not sure that it's guaranteed, but AFAIK it would be today. Have it in there just to be sure.

go/vt/vttablet/tabletmanager/vreplication/vreplicator.go
mattlord: Great idea! I like it. Will switch to that.

go/vt/vttablet/tabletmanager/vreplication/vreplicator.go
mattlord: Yep!

go/vt/vttablet/tabletmanager/vreplication/vreplicator.go
mattlord: Oooh, definitely. Didn't realize that existed. ❤️

mattlord:
> Overall this looks good to me. Please see inline comments. `FULLTEXT` is an issue that needs to be resolved.
> I'd change the way the `ALTER` query gets killed, see comments.

Thanks, @shlomi-noach ! I've attempted to address your comments/suggestions here: https://github.com/vitessio/vitess/pull/11700/commits/922a35002022d83f6c4b9564f85e24da29acd73b

Thanks again!

go/vt/vttablet/tabletmanager/vreplication/vreplicator.go
mattlord: Yeah, I also didn't like that verbiage but forgot to change it. Will do so now.

go/vt/vttablet/tabletmanager/vreplication/vreplicator.go
shlomi-noach: :+1: makes sense to just exclude `FULLTEXT`

go/vt/vttablet/tabletserver/vstreamer/testenv/testenv.go
shlomi-noach: :+1:

go/vt/vttablet/tabletmanager/vreplication/vreplicator.go
```diff
@@ -558,3 +568,304 @@ func recalculatePKColsInfoByColumnNames(uniqueKeyColumnNames []string, colInfos
...
+	// mode.
+	dbClient := vr.vre.dbClientFactoryFiltered()
+	if err := dbClient.Connect(); err != nil {
+		return fmt.Errorf("unable to connect to the database when executing post copy actions: %v", err)
+	}
+	defer dbClient.Close()
+
+	// Save our connection ID so we can use it to easily KILL any
+	// running SQL action we may perform later if needed.
+	qr, err := dbClient.ExecuteFetch("select connection_id()", 1)
```
shlomi-noach, line 702: so as I believe, `dbClient` is a pool. A firther change is needed: you must grab a specific `Conn`, with which you will issue both `select connection_id()` as well as the `alter`. Otherwise, the two could potentially run from two different connections.

mattlord: I don't see where there's any pooling involved (which could cause various problems for vreplication as state is critical). A `binlogplayer.DBClient` is used: https://github.com/vitessio/vitess/blob/main/go/vt/binlog/binlogplayer/dbclient.go#L40-L44

That dbclient wraps a `mysql.Conn`, and `mysql.Connect` ends up getting called, which has no pool and does a Dial: https://github.com/vitessio/vitess/blob/9cb518e489c7fce5653ac52449688f4e38f94890/go/vt/dbconfigs/dbconfigs.go#L173-L184

https://github.com/vitessio/vitess/blob/main/go/mysql/client.go#L49

Am I missing something?

All that being said.... I'll think about the best way to add some assurance that it's still the right connection. I could do something similar to the I_S query I was doing before...

mattlord: Double and triple checked (also talked with @rohit-nayak-ps), no connection pooling is used. So IMO the current code is good. Any reliable verification would be back to what I started with, which is also fine IMO as the `information_schema.INFO` field has the query *received* by the server (same as the `general_log` entry) and we were using the same query (`ALTER TABLE`) sent to the server in the query to get the connection ID running the `ALTER`.

So which do you prefer — what I originally had or what we have now? 🙂

shlomi-noach: My mistake then! I personally prefer the current solution, as it does not involve a search in `processlist`, and just used a known connection ID. At your discretion!

shlomi-noach (approved):

go/vt/vttablet/tabletmanager/vreplication/vreplicator.go
mattlord: I agree. Will stick with the current `connection_id()` approach. ❤️

go/vt/vttablet/tabletmanager/vreplication/vreplicator_test.go
```diff
@@ -173,3 +182,426 @@ func TestPrimaryKeyEquivalentColumns(t *testing.T) {
 		})
 	}
 }
+
+func TestDeferSecondaryKeys(t *testing.T) {
```
rohit-nayak-ps, line 186: Per our new checklist new Test funcs need a comment about what it is doing ...

rohit-nayak-ps (commented):
Functionality looks good. Nice work!

* I guess you have manually tested the full flow for this feature. But maybe we should add a couple of secondary indexes to one of the e2e tables so that the complete flow is tested as part of the e2e tests
* Can we also log the start/end of the post-copy alters to the vreplication_log table for keeping track of how long it takes?
* The PhaseTimings stat for copy phase is now going to include time for the alter. Should we also record PhaseTimings for the post-copy alter, so that we can differentiate this from the copy insert bandwidth?

mattlord:
> Functionality looks good. Nice work!
> 
> * I guess you have manually tested the full flow for this feature. But maybe we should add a couple of secondary indexes to one of the e2e tables so that the complete flow is tested as part of the e2e tests
> * Can we also log the start/end of the post-copy alters to the vreplication_log table for keeping track of how long it takes?
> * The PhaseTimings stat for copy phase is now going to include time for the alter. Should we also record PhaseTimings for the post-copy alter, so that we can differentiate this from the copy insert bandwidth?

Thanks for the great suggestions, @rohit-nayak-ps! I believe that I addressed all of them here: https://github.com/vitessio/vitess/pull/11700/commits/5879ccc48cbcad55512167a9136e5e5a8b1be55a

Adding secondary indexes to the e2e tables and enabling `--defer-secondary-keys` (for v2) actually highlighted a couple of edge cases with `SQL_MODE` and `Reshard` merges/consolidations that I addressed in the commit. 🙂

go/vt/vtctl/vtctl.go
```diff
@@ -2273,6 +2273,15 @@ func commandVRWorkflow(ctx context.Context, wr *wrangler.Wrangler, subFlags *pfl
 			vrwp.TargetShards = strings.Split(*targetShards, ",")
 			vrwp.SkipSchemaCopy = *skipSchemaCopy
 			vrwp.SourceKeyspace = target
+			// We can't properly handle secondary key deferral when
+			// doing Reshard merges/consolidations (e.g. going from
+			// 2 shards to 1) as you have N vreplication records
+			// and streams running on the same target shard for the
+			// same table.
+			if *deferNonPKeys && len(vrwp.TargetShards) < len(vrwp.SourceShards) {
```
rohit-nayak-ps, line 2281: This can happen even when there are more source shards than target shards: say, going from `-80, 80-` to `-40, 40-c0, c0-`. If you enable this feature for `TestBasicVreplicationWorkflow` which runs `reshardMerchant2to3SplitMerge` we should see this.

mattlord: Good point! I realized that it's (of course) not Reshard specific either because you can move a table from a sharded keyspace to an unsharded one. That gave me the impetus to add support for these cases in: https://github.com/vitessio/vitess/pull/11700/commits/f9cb3d2092a19e22705041736e980ad036c2fac5

go/vt/vttablet/tabletmanager/vreplication/vreplicator.go
```diff
@@ -686,14 +689,18 @@ func (vr *vreplicator) getTableSecondaryKeys(ctx context.Context, tableName stri
...
+
 	// Use a new DB client to avoid interfering with open transactions
 	// in the shared client as DDL includes an implied commit.
 	// We're also NOT using a DBA connection here because we want to be
 	// sure that the work fails if the instance is somehow in READ-ONLY
 	// mode.
-	dbClient := vr.vre.dbClientFactoryFiltered()
-	if err := dbClient.Connect(); err != nil {
-		return fmt.Errorf("unable to connect to the database when executing post copy actions: %v", err)
+	dbClient, err := newVCopier(vr).newClientConnection(ctx)
```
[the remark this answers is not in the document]
mattlord: I'll try that. I don't like it either (feels wrong) but it was added for the multiple insert workers. If it doesn't cause any issues around that or others then I'll do move it.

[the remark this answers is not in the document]
mattlord: Done: https://github.com/vitessio/vitess/pull/11700/commits/5537455b32293bb2f205a1e6a7505272932f617d

rohit-nayak-ps (approved):
Nice work! Should significantly speed up a lot of production workflows :-)

PR #11701, opened 2022-11-14
Online DDL: more scheduler triggering following successful operations

shlomi-noach:
## Description

Long title for a minor change. Normally Online DDL scheduler wakes up once per minute to see if there's work to be done. But we increase the wakeup density after any change that is likely to actually cause more work. For example, when a migration is made `ready`, there is likelihood that the scheduler will also be able to execute it right away. Or is a migration is `compelte`, there is likelihood that there's another on in the queue that could execute, etc.

This PR adds a bit more scheduling density in a bit more scenarios.

## Related Issue(s)

#6926

## Checklist


mattlord (approved):
LGTM

PR #11708, opened 2022-11-15
Readable plan tests

systay:
## Description
Using a nicer diff when comparing JSONs. Makes the tests so much easier to read.

Here I changed `sharded` from true to false in a test

With the new diff tool:
```json
        {
            "Instructions": {
                "FieldQuery": "select `user`.col + 2 as a from `user` where 1 != 1",
                "Keyspace": {
                    "Name": "user",
                    "Sharded": false => true
                },
                "OperatorType": "Route",
                "Query": "select `user`.col + 2 as a from `user` having a = 42",
                "Table": "`user`",
                "Variant": "Scatter"
            },
            "Original": "select user.col + 2 as a from user having a = 42",
            "QueryType": "SELECT"
        }
```

Using the old tool:
```
        Diff:
          any(
        - 	json.RawMessage("{\n      \"QueryType\": \"SELECT\",\n      \"Original\": \"select user.col + 2 as a from user having a = 42\",\n      \"Instructions\": {\n        \"OperatorType\": \"Route\",\n        \"Variant\": \"Scatter\",\n        \"Keyspace\": {\n          \"Name\": \"user\",\n          \"Sharded\":"...),
        + 	string("{\n  \"QueryType\": \"SELECT\",\n  \"Original\": \"select user.col + 2 as a from user having a = 42\",\n  \"Instructions\": {\n    \"OperatorType\": \"Route\",\n    \"Variant\": \"Scatter\",\n    \"Keyspace\": {\n      \"Name\": \"user\",\n      \"Sharded\": true\n    },\n    \"FieldQuery\": \""...),
          )
```

## Checklist

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -157,6 +157,7 @@ require (
 	github.com/mitchellh/mapstructure v1.4.1 // indirect
 	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
 	github.com/modern-go/reflect2 v1.0.1 // indirect
+	github.com/nsf/jsondiff v0.0.0-20210926074059-1e845ec5d249 // indirect
 	github.com/onsi/ginkgo v1.12.1 // indirect
 	github.com/onsi/gomega v1.10.3 // indirect
 	github.com/pelletier/go-toml v1.9.3 // indirect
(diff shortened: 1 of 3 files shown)


harshit-gangal (approved):

frouioui (approved):
This is a nice improvement!

deepthi (approved):

PR #11726, opened 2022-11-16
deps: update prom client_golang dependency

deepthi:
## Description
Our previous dependency has a known CVE. Though according to `govulncheck`, our code doesn't use the vulnerable paths, upgrading this to be safe.

## Checklist

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -63,7 +63,7 @@ require (
 	github.com/pkg/errors v0.9.1
 	github.com/planetscale/pargzip v0.0.0-20201116224723-90c7fc03ea8a
 	github.com/planetscale/vtprotobuf v0.3.0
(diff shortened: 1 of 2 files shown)


frouioui (approved):
This change looks good to me. Version `1.11.1` fixes a known CVE, however, `client_golang` had 4 new releases since version `1.11.1`. It would be even better to upgrade to their latest version: `1.14.0`.

deepthi:
> This change looks good to me. Version `1.11.1` fixes a known CVE, however, `client_golang` had 4 new releases since version `1.11.1`. It would be even better to upgrade to their latest version: `1.14.0`.

Good point. #11741 actually does that. Let's close this PR and move that forward.

PR #11694, opened 2022-11-11
update jsonparser dependency

deepthi:
## Description
PR fixes 

1. Use a released version instead of random SHA
2. Fixes typo in the pull request template

## Checklist

-   [x] "Backport me!" label has been added if this change should be backported
-   [x] Tests were added or are not required
-   [x] Documentation was added or is not required

PR #11741, opened 2022-11-16
Update all the Go dependencies

dbussink:
## Description

This seems to work for the basics just fine, so let's have CI take a run at this as well to update these.

Only one small update to the Azure blob storage handling seems needed so far.

## Related Issue(s)

Some more critical changes have been separate updated already in for example #11694 

## Checklist


deepthi:
I think this is a good idea. We can review after the tests are fixed (since they look like they might need actual code changes to pass).

dbussink:
> I think this is a good idea. We can review after the tests are fixed (since they look like they might need actual code changes to pass).

@deepthi Yeah, the main issue is that we have many places that end up using `reflect.DeepEqual` with protobufs which is not supported and only kinda sometimes work. I've updated a lot of cases here but we still have many more of those cases that could break at any change in the protobuf package.

dbussink:
This is now ready for review. The issue in the `vtgate_queries` is unrelated to the changes here, but @harshit-gangal is looking at that separately.

systay (approved):
Thanks for doing this! ❤️

go/sqltypes/result.go
```diff
@@ -90,10 +90,11 @@ func (result *Result) ReplaceKeyspace(keyspace string) {
...
 func (result *Result) Copy() *Result {
 	out := &Result{
-		InsertID:            result.InsertID,
 		RowsAffected:        result.RowsAffected,
-		Info:                result.Info,
+		InsertID:            result.InsertID,
 		SessionStateChanges: result.SessionStateChanges,
+		StatusFlags:         result.StatusFlags,
+		Info:                result.Info,
 	}
```
harshit-gangal, line 98: looks like unrelated changes for the PR `StatusFlags: result.StatusFlags`

harshit-gangal (approved):

notfelineit (approved):

ajm188 (approved):

frouioui (approved):

PR #11743, opened 2022-11-16
Fix two additional flaky test sources in endtoend tests

dbussink:
## Description

This change fixes two more source of flaky tests in endtoend tests. The first once is that base port allocation for tests is racy and it that it can allocate the same base port for two parallel endtoend tests.

The second one is that we can also collide with ephemeral client ports since we allow those to overlap with the listening ports. This can cause flaky tests as well if a component can't actually be started in endtoend tests.

See also the commit messages for more details.

## Related Issue(s)

This is a follow up to #11707 which helped some CI hangs and timeouts but it now shows real underlying issues which could be caused by the two issues this PR fixes. 

## Checklist

### DIFF
--- a/dev.env
+++ b/dev.env
@@ -22,7 +22,7 @@
 
 source ./build.env
 
-export VTPORTSTART=15000
+export VTPORTSTART=6700
 
 
 # Add chromedriver to path for Selenium tests.
--- a/go/test/endtoend/filelock/filelock.go
+++ b/go/test/endtoend/filelock/filelock.go
@@ -1,0 +1,99 @@
+// Copyright 2018 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package filelock provides a platform-independent API for advisory file
+// locking. Calls to functions in this package on platforms that do not support
+// advisory locks will return errors for which IsNotSupported returns true.
+package filelock
+
+import (
+	"errors"
+	"io/fs"
+	"os"
+)
+
+// A File provides the minimal set of methods required to lock an open file.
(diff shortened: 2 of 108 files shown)


vmg (approved):

rohit-nayak-ps (approved):
lgtm

frouioui (approved):

go/test/endtoend/filelock/filelock.go
```diff
@@ -0,0 +1,99 @@
+// Copyright 2018 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
```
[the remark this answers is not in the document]
dbussink: @mattlord Yeah, it's internal in Go and not available. But it's the canonical implementation essentially. See also the first commit message, where I reference to the upstream discussion:

---

Go has a proposal open to make filelock available, but that is still pending in https://github.com/golang/go/issues/33974. Once that is resolved, we can remove our copy of the implementation.

dev.env
```diff
@@ -22,7 +22,7 @@
 
 source ./build.env
 
-export VTPORTSTART=15000
+export VTPORTSTART=6700
```
deepthi, line 25: I'm wondering if this will break the local example, but it seems to be passing 🤷 
@dbussink can you verify that you can `source dev.env` and then go to examples/local and run the example without problems?

deepthi (approved):
Nice work!
We can merge once my one comment is addressed.

dev.env
dbussink: This is the same value as what we use if this env var is not set.

PR #11760, opened 2022-11-16
Fix `codeql` workflow timeout issue

frouioui:
## Description

The CodeQL workflow has been failing due to timeouts. This Pull Request increases the timeout for the `make build`.

There is also a new integration that sends a slack message whenever the workflow fails so we can easily know if the build starts failing. It is usually tough the know if the build fails, we have to manually look at it since the PR is executed in a CRON. The message goes to the `#codeql` channel of our OSS Slack workplace.

## Related Issue(s)
Fixes https://github.com/vitessio/vitess/issues/11761

## Checklist


frouioui:
In commit 6aba78c you can see the workflow fail (https://github.com/vitessio/vitess/actions/runs/3484154076/jobs/5828414742) and send a slack notification: https://vitess.slack.com/archives/C04BMRXQUL9/p1668644474844189

rohit-nayak-ps (approved):

mattlord (approved):
Nice!

PR #11765, opened 2022-11-17
rewrite predicates to expose routing opportunities

### LINKED ISSUE #11766: Feature Request: simplify predicates
### Feature Description

When the sharding key is hidden behind OR logic, like here:

```sql
where (id = 5 and B) or (id = 5 and C)
```

Vtgate is not able to route it well and falls back on a `Scatter`. It would be great if this could be rewritten to:

```sql
where id = 5 and (B or C)
```

systay:
## Description
We had some logic for turning `OR` predicates into `IN`, and we had some code to do conjuctive normal form, but the two were not connected never used together.

In this PR, I connected them, and also added more logic to simplify `OR` expressions. Specifically, I set out to solve a pattern that looks like this:

```sql
WHERE (A and B) or (A and C)
```

If the predicate A in this case could help us with routing, it would be nice to extract it out to an equivalent form outside the `OR`, like:

```sql
WHERE A and (B or C)
```

## Related Issue(s)
Fixes #11766

## Checklist


go/vt/sqlparser/predicate_rewriting_test.go
```diff
@@ -0,0 +1,134 @@
...
+package sqlparser
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestRewriteToCNF(in *testing.T) {
```
dbussink, line 26: @systay Is the test name here still correct? I think it does more than `CNF` now?  Since we run `simplifyExpression` so it's a generic test for that function?

go/vt/sqlparser/predicate_rewriting_test.go
```diff
@@ -0,0 +1,134 @@
...
+		expected: "A and B",
+	}, {
+		in:       "(A or B) and A",
+		expected: "A",
+	}, {
+		in:       "A and (A or B)",
+		expected: "A",
+	}, {
+		in:       "(A and B) OR (A and C)",
+		expected: "A and (B or C)",
```
dbussink, line 77: Should we also add some tests that it works if you have a deeper tree like `(A and B) OR (A and C) OR (A and D)` and that we turn that into `A and (B or C or D)`? More as a regression test as well?

go/vt/vtgate/planbuilder/gen4_planner.go
```diff
@@ -30,6 +30,7 @@ import (
 	"vitess.io/vitess/go/vt/vtgate/vindexes"
 )
 
+//goland:noinspection GoVarAndConstTypeMayBeOmitted
```
dbussink, line 33: Do we use editor specific annotations in other places as well or should this be dropped?

arthurschreiber (approved):
This is such a nice (and straightforward) improvement. ❤️

go/vt/sqlparser/predicate_rewriting_test.go
systay: we have that in the `TestRewritePredicate` test

dbussink: Ah yes, I see it now.

go/vt/sqlparser/predicate_rewriting.go
```diff
@@ -0,0 +1,320 @@
...
+type RewriteState bool
+
+// RewritePredicate walks the input AST and rewrites any boolean logic into a simpler form
+// This simpler form is CNF plus logic for extracting predicates from OR, plus logic for turning ORs into IN
+// Note: In order to re-plan, we need to empty the accumulated metadata in the AST,
+// so ColName.Metadata will be nil:ed out as part of this rewrite
+func RewritePredicate(ast SQLNode) SQLNode {
+	for {
+		finishedRewrite := true
+		log.Errorf(String(ast))
```
dbussink, line 35: @systay Some stray debugging here?

dbussink (approved):
Very nice!

frouioui (approved):

harshit-gangal (approved):
This is really nice simplification.

Issue #11742, opened 2022-11-16
Bug Report: VDiff2 becoming stuck


arthurschreiber:
### Overview of the Issue

We are running VDiff2 to verify a `MoveTables` workflow (from an unsharded cluster to a 16 shard cluster).

We have encountered that VDiff2 becomes "stuck". Here's an excerpt from the vttablet log files:

```
E1114 21:40:32.413910   40919 table_differ.go:476] Failed to update vdiff progress on <table> table: conn 5080565: Write(packet) failed: write tcp 10.48.125.40:55164->10.48.125.40:3306: write: broken pipe (errno 2006) (sqlstate HY000) during query: update _vt.vdiff_table set rows_compared = 22836878, lastpk = 'fields:{name:\"id\" type:INT32 table:\"<table>\" org_table:\"<table>\" database:\"<database>\" org_name:\"id\" column_length:11 charset:63 flags:49667} rows:{lengths:9 values:\"425986171\"}', report = '{\"TableName\":\"<table>\",\"ProcessedRows\":22836878,\"MatchingRows\":22836878,\"MismatchedRows\":0,\"ExtraRowsSource\":0,\"ExtraRowsTarget\":0}' where vdiff_id = 1 and table_name = '<table>'
E1114 21:40:32.414278   40919 workflow_differ.go:113] Encountered an error diffing table <table> for vdiff 93cbbb05-6464-11ed-bb35-b49691ae0fc4: Code: UNKNOWN
vttablet: rpc error: code = Unknown desc = read tcp 10.52.116.15:41952->10.52.116.15:3306: read: connection reset by peer
io.ReadFull(packet body of length 396) failed (errno 2013) (sqlstate HY000)
E1114 21:40:32.414401   40919 controller.go:230] Encountered an error performing workflow diff for vdiff 93cbbb05-6464-11ed-bb35-b49691ae0fc4: conn 5080528: Write(packet) failed: write tcp 10.48.125.40:55052->10.48.125.40:3306: write: broken pipe (errno 2006) (sqlstate HY000) during query: update _vt.vdiff_table set state = 'error' where vdiff_id = 1 and table_name = '<table>'
[...]

PR #11768, opened 2022-11-17
VReplication: Prevent Orphaned VDiff2 Jobs

mattlord:
## Description

This addresses the two ways a VDiff could become "orphaned" and require a manual `Stop` and `Resume` step:
1. A tablet and/or mysqld are not shutdown normally or otherwise cannot save the final state
    - Fix: when opening the engine, restart any vdiffs that are in the started state as this indicates it did not complete and was unable to save the final state and must be restarted.
3. The vdiff run fails with an error (e.g. the MySQL connection is closed/lost)
    - Fix: when a vdiff run fails, retry saving the error state with an exponential backoff until the engine shuts down. This way the normal retry mechanism will  eventually kick in OR the behavior in item 1 will kick in when the engine is next opened on the primary tablet.

## Related Issue(s)

  - Fixes: https://github.com/vitessio/vitess/issues/11742

## Checklist


go/vt/vttablet/tabletmanager/vdiff/engine.go
```diff
@@ -175,7 +189,13 @@ func (vde *Engine) retry(ctx context.Context, err error) {
 // addController creates a new controller using the given vdiff record and adds it to the engine.
 // You must already have the main engine mutex (mu) locked before calling this.
 func (vde *Engine) addController(row sqltypes.RowNamedValues, options *tabletmanagerdata.VDiffOptions) error {
-	ct, err := newController(vde.ctx, row, vde.dbClientFactoryDba, vde.ts, vde, options)
+	var tmc tmclient.TabletManagerClient
+	if vde.testingTMC != nil {
```
rohit-nayak-ps, line 193: Why not pass `tmc` while creating the engine itself, rather than the `testingTMC` attribute?

go/vt/vttablet/tabletmanager/vdiff/framework_test.go
```diff
@@ -0,0 +1,413 @@
...
+		vstreamerEngine.InitDBConfig(env.KeyspaceName, env.ShardName)
+		vstreamerEngine.Open()
+		defer vstreamerEngine.Close()
+
+		if err := env.Mysqld.ExecuteSuperQuery(context.Background(), fmt.Sprintf("create database %s", vdiffdb)); err != nil {
+			fmt.Fprintf(os.Stderr, "%v", err)
+			return 1
+		}
+
+		if err := env.Mysqld.ExecuteSuperQueryList(context.Background(), binlogplayer.CreateVReplicationTable()); err != nil {
```
rohit-nayak-ps, line 96: can specify a list of expected queries and iterate over them?

go/vt/vttablet/tabletmanager/vdiff/engine.go
mattlord: I removed these changes in the engine and controller and am now doing this properly in the test framework: https://github.com/vitessio/vitess/pull/11768/files#diff-eb9190e65920b52acd2700209d868f655e0308eecb2e4893ad2ff20adf74da1bR76-R77

go/vt/vttablet/tabletmanager/vdiff/framework_test.go
mattlord: Done!

harshit-gangal (approved):

rohit-nayak-ps (approved):
Nice. Thanks for initiating the unit test framework for vdifff2!

Issue #11770, opened 2022-11-18
Feature Request: Add fallback to VTGate cell alias for VStreams with local cell preference


pbibra:
### Feature Description

More context in this thread: https://vitess.slack.com/archives/C0PQY0PTK/p1668557424770139

We'd like to make a server side update to default to using a VTGate's cell alias for tablet selection when cells are not specified in the VStreamsRequest object in the gRPC call. In addition, to address potential performance concerns, we'd like to add local cell preference to the TabletPicker when specified.

### Use Case(s)

VStreams use case. Similar behavior exists within the regular VTGate query execution flow.

Issue #11579, opened 2022-10-25
Feature Request: Specifying multiple TabletTypes (e.g. REPLICA,PRIMARY) during VStream client subscription.


HenryCaiHaiying:
### Feature Description

Currently, when the client initiate a VStream subscription, we can only specify one TabletType:
vtgate.proto
  // VStreamRequest is the payload for VStream.
  message VStreamRequest {
   vtrpc.CallerID caller_id = 1;
   topodata.TabletType tablet_type = 2;

But we want to specify multiple tablet types to have the choice to fallback to a different type of tablet.  And looks like VReplication path already supports that:
https://vitess.io/docs/14.0/reference/vreplication/flags/#vreplication_tablet_type
  vreplication_tablet_type
  Default in_order:REPLICA,PRIMARY

The server side implementation is also ready: https://github.com/vitessio/vitess/blob/main/go/vt/discovery/tablet_picker.go#L170

### Use Case(s)

for VStream client (debezium) subscription, we usually specify TabletType=Replica, however in some of AZs there is only Primary tablet (no Replica tablet), we would prefer to specify multiple tablet types to enlarge the tablet selection.  Otherwise the client needs to do a relatively complex fallback/retry logic and we don't really own that client project (debezium) to modify that logic easily.

HenryCaiHaiying:
We have semi-sync configured for our Vitess layout.  And to make sure there is at least 2 AZs gets the data, the Primary and Replica tablet are laid out in two different AZs.  And VtGate is laid out per AZ and it only knows the tablet in its own AZ.  When the debezium client connects to the VtGate in its own AZ with subscription request tabletType=Replica, that VtGate might not have a Replica tablet to match (it might only have a Primary tablet in its AZ).  In this case, we would prefer the VtGate to fall back to that Primary tablet.

HenryCaiHaiying:
We are mostly interested in the fallback option during initial connection, not necessarily during the middle of the streaming.

falun:
Moving some thoughts from community slack thread: With the caveat of not having looked at the VStream code in a long time I _thought_ that being able to pick tablets across cells was resolved with https://github.com/vitessio/vitess/pull/6442. It's feasible that that code isn't in the vstream tablet selection path though?

If not it may be a fast & "easy" fix to solve the underlying problem which (seems to be?) "can't select tablet type that isn't in vtgate's cell" in addition to adding the ability to have fallback tablet type selectable.

HenryCaiHaiying:
[...]

PR #11771, opened 2022-11-18
tablet picker cell alias fallback with local cell preference

pbibra:

## Description

Allow for cell alias fallback during tablet selection for VStreams when client does not specify list of cells. In addition, add the option for local cell preference during tablet selection.

Overall, this PR is trying to solve 2 problems: 

Problem 1: We're trying to avoid passing in the list of cells/cell alias that `TabletPicker` can choose from in the gRPC request itself. Instead, we will now select a candidate from a list of tablets in the VTGate's local cell + any cell within a defined cell alias for the local cell, if one exists. 

e.g. We have a cell alias called `region-1` which includes the following cells `{region-1a, region-1b, region-1c}`
A VTGate in `region-1b` sets up a VStream. The gRPC `VStreamRequest` object does not have any `cells` sent from the client, but a cell alias is defined.

As a result `TabletPicker.PickForStreaming()` will choose from candidates not just in the VTGate's local cell, but in all cells within the `region-1` alias without the client having to explicitly specify this.

Problem 2: Of all the candidates selected by `TabletPicker`, we need a way to allow it to give priority to the local cell. This is where the `localPreferenceHint` comes in. Two paths in which this can be specified: 

a. If no cells specified by the client, send in the list of cells to the TabletPicker as `local:region-1b, region-1`. The first item in the list is the local cell we want to prioritize and the second is the cell alias that `region-1b` belongs to. 

b. The second way is to allow the client side to also specify this if they choose to pass in the cells within the gRPC request itself rather than use the new flag. They'd send in `local:,region-1` or if no cell alias, then something like `local:,region-1a, region-1c` within the VStreamRequest object.

In both cases, `TabletPicker` can then order candidates like: 

`tablet-1-region-1b` - PRIMARY
`tablet-2-region-1b` - REPLICA
`tablet-1-region-1c` - REPLICA
`tablet-1-region-1a` - REPLICA

where `tablet-1-region-1b` would be prioritized since it is in the local cell. The `in_order` hint is then applied on top of this, where we order by tablet type within each "group" (local cells first, then all others). For example, if the tablet type ordering is `inOrder:REPLICA,PRIMARY`, then we get the following ordering of candidates:

`tablet-2-region-1b` - REPLICA
`tablet-1-region-1b` - PRIMARY
`tablet-1-region-1c` - REPLICA
`tablet-1-region-1a` - REPLICA

## Testing

We did some testing on the Slack side and looks like cell alias selection and local cell preference is working as expected:

```
I1129 13:40:14.716001    4703 vstream_manager.go:445] [VSTREAM MANAGER] no cells specified by client, falling back to alias...
I1129 13:40:14.716020    4703 vstream_manager.go:461] [VSTREAM MANAGER] cells to pick from [local:us_east_1e us_east_1]
I1129 13:40:14.717516    4703 vstream_manager.go:445] [VSTREAM MANAGER] no cells specified by client, falling back to alias...
I1129 13:40:14.717536    4703 vstream_manager.go:461] [VSTREAM MANAGER] cells to pick from [local:us_east_1e us_east_1]
```
For a vtgate in `us-east-1b` got the following tablet selection logs:
```
I1128 16:48:05.447026    4703 vstream_manager.go:534] Starting to vstream from cell:"us_east_1b" ...
I1128 16:48:05.762611    4703 vstream_manager.go:534] Starting to vstream from cell:"us_east_1d" ...
I1128 16:48:05.762815    4703 vstream_manager.go:534] Starting to vstream from cell:"us_east_1e" ...
I1128 16:48:05.763515    4703 vstream_manager.go:534] Starting to vstream from cell:"us_east_1b" ...
I1128 16:48:05.765803    4703 vstream_manager.go:534] Starting to vstream from cell:"us_east_1e" ...
```

## Related Issue(s)

  - Fixes: https://github.com/vitessio/vitess/issues/11770

More discussion here: https://vitess.slack.com/archives/C0PQY0PTK/p1668557424770139
## Checklist


mattlord (commented):
Thank you for working on this! ❤️ The core of it seems good but the user interface is unclear to me.

When you say “a vtgate’s cell alias” what do you mean by that? Do you just mean the vtgate process's cell?

I was imagining, user would specify a vtgate flag of ~ `--vstream-prefer-local-cell` and then if the vtgate’s local cell was one of those specified in the vstream gRPC call then we’d prefer any of the tablets in that cell.

So then if you start a vtgate this way: `vtgate --tablet_types="in_order:replica,primary" --vstream-prefer-local-cell` we should be able to stream from tablets in the same cell as the vtgate. It seems that instead you were thinking that in the vstream gRPC the user would add a prefix to the cells of `localPreference:<local cell, which can be an alias>` but that feels unintuitive to me. Maybe I'm missing some things though.

I’m not sure where Cell alias comes into play, which e.g. could be zone1,zone2,zone3 (a common alias in the past was all cells). Maybe you have N cells per data center, and an alias is for all of those local cells? Or maybe alias is errantly thrown in there?

go/vt/vtgate/vtgate.go
```diff
@@ -132,6 +134,7 @@ func registerFlags(fs *pflag.FlagSet) {
 	fs.BoolVar(&enableSchemaChangeSignal, "schema_change_signal", enableSchemaChangeSignal, "Enable the schema tracker; requires queryserver-config-schema-change-signal to be enabled on the underlying vttablets for this to work")
 	fs.StringVar(&schemaChangeUser, "schema_change_signal_user", schemaChangeUser, "User to be used to send down query to vttablet to retrieve schema changes")
 	fs.IntVar(&queryTimeout, "query-timeout", queryTimeout, "Sets the default query timeout (in ms). Can be overridden by session variable (query_timeout) or comment directive (QUERY_TIMEOUT_MS)")
+	fs.BoolVar(&vstreamCellAliasFallback, "vstream_cell_alias_fallback", vstreamCellAliasFallback, "Determines whether to default to cell aliases for VStream tablet selection instead of having the gRPC request specify this in the VStreamrRequest object.")
```
mattlord, line 137: We're moving from underscores to dashes so we should use dashes for all new flags. I think the description would be a little more clear if it was ~:
```
fs.BoolVar(&vstreamUseCellAlias, "vstream-use-cell-alias", vstreamUseCellAlias, "By default, use a cell alias which contains the vtgate's local cell when selecting tablets for streaming")
```

Since I think the default, when no specific behavior was selected in the gRPC call, goes from `search for tablets in the local cell` to `search for tables in a cell alias that contains the local cell or if there isn't one, then search for tablets in the local cell`. No?

mattlord, line 137: I think this would be correct and more clear:
```
"By default, when no tablets can be found in the local cell for streaming, also look for tablets in a cell alias which includes the local cell")
```

As the gRPC cells I'm assuming will be used when specified and this flag then has no impact. So it only alters the default behavior when no cells are provided and searches the alias as a fallback if needed.

go/vt/vtgate/vstream_manager.go
```diff
@@ -411,14 +411,27 @@ func (vs *vstream) alignStreams(ctx context.Context, event *binlogdatapb.VEvent,
...
+	// perform cell alias fallback
+	if vstreamCellAliasFallback && len(cells) == 0 {
+		// append the alias this cell belongs to, otherwise appends the vtgate's cell
+		alias := topo.GetAliasByCell(ctx, vs.ts, vs.vsm.cell)
+		// an alias was actually found
+		if alias != vs.vsm.cell {
+			// send in the vtgate's cell for local cell preference
+			alias = strings.Join([]string{"local:", vs.vsm.cell, alias}, ",")
+		}
+		cells = append(cells, alias)
```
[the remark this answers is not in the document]
mattlord: And the behavior here should be the same if we had specified this for the VStreamRequest message of the VStream RPC call:
`cells: "local:zone1-a,zone1"`

go/vt/discovery/tablet_picker.go
```diff
@@ -92,19 +94,56 @@ func NewTabletPicker(ts *topo.Server, cells []string, keyspace, shard, tabletTyp
 		return nil, vterrors.Errorf(vtrpcpb.Code_FAILED_PRECONDITION,
 			fmt.Sprintf("Missing required field(s) for tablet picker: %s", strings.Join(missingFields, ", ")))
 	}
+
+	localPreference := ""
+	if strings.HasPrefix(cells[0], "local:") {
+		localPreference = cells[0][len(localPreferenceHint):]
+		cells = cells[1:]
```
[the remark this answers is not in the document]
pbibra: I'm not sure what you mean, does this not compare the tablet's cell to the local cell value? The TabletPicker.GetMatchingTablets function resolves the cell alias passed in into the cells that are included within it.

go/vt/vtgate/vstream_manager.go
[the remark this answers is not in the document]
pbibra: ahh yeah, we'd want alias to be a list and not a string join, I can fix that!

rohit-nayak-ps:
Thanks for the detailed use case and the PR. I wanted to bring up another aspect: the tablet picker is also used by the standard VReplication flows. 

Prefering the local cell when multiple cells are provided will be useful for these flows as well. The current implementation forces the user to choose between:
* specifying the local cell only, to avoid cross-AZ traffic. This can lead to starvation if there is no tablet in the local cell at some point
* specifying multiple cells to always find a candidate, in which case we could incur significant cross-AZ traffic  

The main issue is that the picker doesn't _know_ the default cell of the caller today. One option is to update the tablet picker api, so that we can pass that in and add picker strategy options. The options can be used instead of the "local:" hint (and we will also deprecate the "in_order:" hint for tablet types replacing it with an option). 

For cell selection the options would be:
* **Default** // prefer local cell, then specified cells
* **ExtendedDefault** // prefer local cell, then cell alias of local cell, then specified cells
* **Specified** // prefer specified cells, no fallback  

I am ok with combining the two defaults and providing the functionality of the ExtendedDefault as the Default. But it will be a breaking change and I will need to validate internally if that choice is acceptable from a backward compatibility viewpoint.

Since I am joining this discussion late, I hope I have understood your requirement correctly. I know you have already had a call with @mattlord, @pbibra , but if it will help we can have another one. I was thinking of including the requirements of https://github.com/vitessio/vitess/issues/11579 as well in the refactor @HenryCaiHaiying.

Regarding implementation, since it impacts other vreplication workflows as well, if you prefer, I can work on the tablet picker refactor this week. Your PR can then use that version of the tablet picker: by specifying vtgate's cell as the local cell and any strategy override. The VStream API's VStreamFlags can be used to specify any strategy override. Users with older Debezium will end up using the default behavior in their Vitess cluster.

Let us know what you think.

pbibra:
Hi @rohit-nayak-ps! Thanks for the additional context. Looks like modifying the tablet picker is the better long term change, but I think @HenryCaiHaiying might have some concerns because we'd like this update to be in our v13 release. Modifying tablet picker might make that backport tougher. Let's do a huddle tomorrow and I will start a chat to discuss this in Slack.

As for the options, I think having Default and ExtendedDefault be separate is completely fine as long as we can specify these server side in the VStream use case and not have to modify the client side gRPC request.

Thank you!

HenryCaiHaiying:
I think @pbibra 's fix already satisfies my original request in https://github.com/vitessio/vitess/issues/11579.  It's better to close in on the current PR rather than doing another complex refactoring (which might also be harder to back port to V13).

pbibra:
Started a discussion here: https://vitess.slack.com/archives/C04DCRJS6QZ/p1669761282103889

go/vt/discovery/tablet_picker_test.go
```diff
@@ -364,6 +364,108 @@ func TestPickUsingCellAlias(t *testing.T) {
 	assert.True(t, picked2)
 }
 
+func TestPickLocalPreference(t *testing.T) {
```
rohit-nayak-ps, line 367: Since the tests are very similar, you can combine all the three tests into a single table-based one. For example, see `TestVStreamSkew` or `TestGTIDCompress` . Essentially you can define a struct type with test options, inputs and expected outputs and iterate over it with conditional logic.

go/vt/vtgate/vstream_manager.go
```diff
@@ -411,18 +411,55 @@ func (vs *vstream) alignStreams(ctx context.Context, event *binlogdatapb.VEvent,
...
 			cells = append(cells, strings.TrimSpace(cell))
 		}
 	}
 
+	// if fallback requested in VTGate startup and no override provided in gRPC request,
+	// perform cell alias fallback
+	if vstreamCellAliasFallback && len(cells) == 0 {
+		log.Info("No cells provided by client, falling back to local cell and alias...\n")
+		// append the alias this cell belongs to, otherwise appends the vtgate's cell
+		alias := []string{topo.GetAliasByCell(ctx, vs.ts, vs.vsm.cell)}
```
rohit-nayak-ps, line 447: `alias` can be just a string, right? We will never get a slice from `GetAliasByCell()`.

go/vt/discovery/tablet_picker_test.go
pbibra: I will do this for the tests introduced in `vstream_manager_test.go`. The ones here are a bit more complex since they have different number of tablet, types, cells in each. It's becoming more complex to merge these.

go/vt/vtgate/vstream_manager.go
```diff
@@ -411,18 +411,54 @@ func (vs *vstream) alignStreams(ctx context.Context, event *binlogdatapb.VEvent,
...
+// If the "local" tag is passed in as an option in the list of optCells,
+// the local cell of the VTGate will take precedence over any other cell specified.
+func (vs *vstream) getCells(ctx context.Context) []string {
 	var cells []string
 	if vs.optCells != "" {
-		for _, cell := range strings.Split(strings.TrimSpace(vs.optCells), ",") {
+		for i, cell := range strings.Split(strings.TrimSpace(vs.optCells), ",") {
+			// if the local tag is passed in, we must give local cell priority
+			// during tablet selection. Append the VTGate's local cell to the list of cells
+			if i == 0 && cell == "local" {
```
rohit-nayak-ps, line 434: Should we use `local:` instead of `local` here? This way it is uniform with the way we request cell locality. 

Also, if a user names a cell as `local` then it doesn't cause any ambiguity.

pbibra: It seems a bit strange to have a semicolon without anything to the right of it, but yeah we do need some differentiator if someone can also name a cell "local"

rohit-nayak-ps:
@pbibra, code changes look good. Sorry, in my previous review, I missed a doc change that needs to be part of this PR too. Similar to this update in an other PR: https://github.com/vitessio/vitess/pull/11874/files#diff-ccb9bf989b7450df22db2bf4bc687668cb4046e1974dd7d5e5e12b2b54a29de3.

We need to update `summary.md` to add the enhancement to the picker. I would have done it except that I can't push commits to your branch. Just add a section to it summarizing your changes and I will approve.

go/vt/vtgate/vstream_manager.go
```diff
@@ -415,18 +415,52 @@ func (vs *vstream) alignStreams(ctx context.Context, event *binlogdatapb.VEvent,
...
+// If the "local:" tag is passed in as an option in the list of optCells,
+// the local cell of the VTGate will take precedence over any other cell specified.
+func (vs *vstream) getCells(ctx context.Context) []string {
 	var cells []string
 	if vs.optCells != "" {
-		for _, cell := range strings.Split(strings.TrimSpace(vs.optCells), ",") {
+		for i, cell := range strings.Split(strings.TrimSpace(vs.optCells), ",") {
+			// if the local tag is passed in, we must give local cell priority
+			// during tablet selection. Append the VTGate's local cell to the list of cells
+			if i == 0 && cell == "local:" {
```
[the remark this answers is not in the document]
mattlord: The local cell related changes in the TabletPicker also affect vttablet<->vttablet vstreams, but there we do not have this local cell correction. How are we going to document that in the [Tablet Picker docs](https://vitess.io/docs/15.0/reference/vreplication/tablet_selection/)?

I'd rather us do the same for vreplication: https://github.com/vitessio/vitess/blob/main/go/vt/vttablet/tabletmanager/vreplication/controller.go#L126 and for vdiff: https://github.com/vitessio/vitess/blob/main/go/vt/vttablet/tabletmanager/vdiff/table_differ.go#LL252

As you can see though, this starts to get tricker as it's being done in multiple places — which could unintentionally drift over time — and we may add new locations in the future. So what we should do instead, IMO, is move this local cell correction code to the TabletPicker itself. That discovery component will be running in the vtgate and vttablet, and it knows what its local cell is. Then this correction code would also be covered by the new unit tests and we can confirm the potential "lost cell" issue noted above.

go/vt/vtgate/vstream_manager.go
```diff
@@ -415,18 +415,52 @@ func (vs *vstream) alignStreams(ctx context.Context, event *binlogdatapb.VEvent,
...
+// and only tablets belonging to the specified cells will be selected.
+// If the "local:" tag is passed in as an option in the list of optCells,
+// the local cell of the VTGate will take precedence over any other cell specified.
+func (vs *vstream) getCells(ctx context.Context) []string {
 	var cells []string
 	if vs.optCells != "" {
-		for _, cell := range strings.Split(strings.TrimSpace(vs.optCells), ",") {
+		for i, cell := range strings.Split(strings.TrimSpace(vs.optCells), ",") {
+			// if the local tag is passed in, we must give local cell priority
+			// during tablet selection. Append the VTGate's local cell to the list of cells
```
mattlord, line 435: More accurately, I think we're prepending the local cell.

go/vt/vtgate/vstream_manager.go
```diff
@@ -415,18 +415,52 @@ func (vs *vstream) alignStreams(ctx context.Context, event *binlogdatapb.VEvent,
...
+	if len(cells) == 0 {
+		log.Info("No cells provided by client, falling back to local cell and alias...\n")
+		// append the alias this cell belongs to, otherwise appends the vtgate's cell
+		alias := topo.GetAliasByCell(ctx, vs.ts, vs.vsm.cell)
+		// an alias was actually found
+		if alias != vs.vsm.cell {
+			// send in the vtgate's cell for local cell preference
+			cells = append(cells, fmt.Sprintf("local:%s", vs.vsm.cell))
+		}
+		cells = append(cells, alias)
```
mattlord, line 454: Hmm, so this new behavior is also specific to vtgate vstreams at this point. I thought we wanted to change the default behavior of the tablet picker itself so that when no cells are specified then we use the cell alias as a fallback when we find no candidate tablets in the local cell. That's what the docs seemed to suggest and I think that we should do that.

Keep in mind that we need to document this new behavior here: https://vitess.io/docs/16.0/reference/vreplication/tablet_selection/
We should also modify it to note that it applies to vtgate vstreams as well.

go/vt/vtgate/vstream_manager_test.go
```diff
@@ -274,6 +274,74 @@ func TestVStreamChunks(t *testing.T) {
...
+		name        string
+		optCells    string
+		cellAlias   bool
+		resultCells []string
+	}
+
+	tcases := []*testArgs{
+		{"default-local", "", false, []string{"aa"}},
+		{"default-local-cell-alias", "", true, []string{"local:aa", "region1"}},
+		{"with-opt-cells", "local:,bb,cc", true, []string{"local:aa", "bb", "cc"}},
```
mattlord, line 288: Same comment here about using the struct field names to make it more readable.

go/vt/vtgate/vstream_manager_test.go
```diff
@@ -274,6 +274,74 @@ func TestVStreamChunks(t *testing.T) {
...
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	cell := "aa"
```
mattlord, line 294: We should make the local cell a test case variable too. That would have likely highlighted the "lost cell" issue too.

go/vt/vtgate/vstream_manager_test.go
```diff
@@ -274,6 +274,74 @@ func TestVStreamChunks(t *testing.T) {
...
+				vsm:                vsm,
+				eventCh:            nil,
+				heartbeatInterval:  0,
+				ts:                 ts,
+			}
+
+			if tcase.cellAlias {
+				cellsAlias := &topodatapb.CellsAlias{
+					Cells: []string{"aa", "bb"},
+				}
```
mattlord, line 325: I think there's value in using different cell alias definitions too. i.e. making this a test case string slice variable rather than a bool.

mattlord (commented):
I mostly had some minor comments/nits/suggestions, but I also have some larger concerns about tablet selection and the tablet picker behavior. I'm happy to chat about them via Slack or Zoom this week. Thanks! ❤️

go/vt/vtgate/vstream_manager.go
```diff
@@ -415,18 +415,52 @@ func (vs *vstream) alignStreams(ctx context.Context, event *binlogdatapb.VEvent,
...
+				cells = append(cells, fmt.Sprintf("local:%s", vs.vsm.cell))
+				continue
+			}
 			cells = append(cells, strings.TrimSpace(cell))
 		}
 	}
 
+	// if no override provided in gRPC request, perform cell alias fallback
+	if len(cells) == 0 {
+		log.Info("No cells provided by client, falling back to local cell and alias...\n")
```
mattlord, line 446: No need for an extra newline here as log messages always end with one.

go/vt/vtgate/vstream_manager.go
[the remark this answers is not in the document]
pbibra: > Wouldn't optCells e.g. be: local:zone1a,zone1b,zone1c? In that case, wouldn't the first element be local:zone1a rather than local:? 

No, since the client has no way of telling us that what the local cell actually is, we decided on tagging just "local" to indicate local cell preference. @rohit-nayak-ps suggested making this "local:" to keep consistent with the hint in the tablet picker. 

I get your point about moving this into the tabletpicker though. Earlier we had the flag, but now we don't so it makes more sense for this to live there. I can update.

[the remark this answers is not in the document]
mattlord: Ahh, OK. I missed that. Sorry. 

That syntax feels awkward to me and is not consistent with `in_order:<type>,...`. It's similar but different — `local:,<cell>,...` — which I don't like. That feels like one small part of a bad UX to me (non-intuitive). We could align both though and make it more clear (IMO) that it's a general preference and not related to the cell immediately following it, if we used this syntax (a variant of what I shared above could parse this correctly):
`local_preference:<cell1>,<cell2>,<cell3>,....`  

I think that keyword conveys the idea better and maintains the same kind of syntax as `in_order:<type1>,<type2>,...`. 

If you both prefer the current syntax, then I think we could make it more clear that this is a dynamic placeholder by using something like:
`{{local_cell}},cell1,cell2,...`.

What do you both think?

Then the next issue is that the TabletPicker itself doesn't have direct access to anything it can use to determine the local cell it's in. So that's something that the callers (vtgate and vttablet) — who do have that info easily accessible — would have to pass in.

doc/releasenotes/16_0_0_summary.md
```diff
@@ -15,6 +15,13 @@
...
 
+### New TabletPicker Options and Default Cell Behavior
+
+In [PR 11771](https://github.com/vitessio/vitess/pull/11771) we allow for default cell alias fallback during tablet selection for VStreams when client 
+does not specify list of cells. In addition, we add the option for local cell preference during tablet selection. 
+The local cell preference takes precedence over tablet type.See PR description for examples. If a client wants to specify local cell preference in the gRPC request, 
+they can pass in a new "local:" tag with the rest of the cells under VStreamFlags. e.g. "local:,cella,cellb".
```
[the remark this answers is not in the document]
mattlord: Please note that we have in-progress VStream gRPC docs in: https://github.com/vitessio/website/pull/1216

Preview: https://deploy-preview-1216--vitess.netlify.app/docs/16.0/reference/vreplication/vstream/

pbibra:
closing in favor of solution here: https://github.com/vitessio/vitess/issues/11999

PR #11790, opened 2022-11-22
Switch ApplySchema `--sql` argument to be `StringArray` instead of `StringSlice`

### LINKED ISSUE #11785: Operator Examples: Add `commerce` sequences with vtctldclient fails on parsing
### Overview of the Issue

When running the operator example, at the third step (resharding), we are asked to add the vitess sequences to the `commerce` keyspace using `vtctldclient` using the following commands:
```
vtctldclient ApplySchema --sql="$(cat create_commerce_seq.sql)" commerce
vtctldclient ApplyVSchema --vschema="$(cat vschema_commerce_seq.json)" commerce
```

However, the first command fails with a parsing error on the first line of the `create_commerce_seq.sql` file. The file is well written:

```sql
create table customer_seq(id int, next_id bigint, cache bigint, primary key(id)) comment 'vitess_sequence';
insert into customer_seq(id, next_id, cache) values(0, 1000, 100);
create table order_seq(id int, next_id bigint, cache bigint, primary key(id)) comment 'vitess_sequence';
insert into order_seq(id, next_id, cache) values(0, 1000, 100);
```

It seems that the first line of the file is always the issue. If I commend the first two lines and executes, it passes without an issue. If I add a comment before the 4 existing line and execute, it passes well.

My current workaround is to simply add one line of comment at the beginning of the file.

### Reproduction Steps

Run the local examples until the resharding phase.

### Binary Version

```sh
main
```

```sh
OSX
```

### Log Fragments

_No response_

ajm188:
## Description

See [this comment](https://github.com/vitessio/vitess/issues/11785#issuecomment-1323641942) for an explanation.

## Related Issue(s)

Fixes #11785.

## Checklist

### DIFF
--- a/go/cmd/vtctldclient/command/schema.go
+++ b/go/cmd/vtctldclient/command/schema.go
@@ -293,7 +293,7 @@ func init() {
 	ApplySchema.Flags().DurationVar(&applySchemaOptions.WaitReplicasTimeout, "wait-replicas-timeout", wrangler.DefaultWaitReplicasTimeout, "Amount of time to wait for replicas to receive the schema change via replication.")
 	ApplySchema.Flags().BoolVar(&applySchemaOptions.SkipPreflight, "skip-preflight", false, "Skip pre-apply schema checks, and directly forward schema change query to shards.")
 	ApplySchema.Flags().StringVar(&applySchemaOptions.CallerID, "caller-id", "", "Effective caller ID used for the operation and should map to an ACL name which grants this identity the necessary permissions to perform the operation (this is only necessary when strict table ACLs are used).")
-	ApplySchema.Flags().StringSliceVar(&applySchemaOptions.SQL, "sql", nil, "Semicolon-delimited, repeatable SQL commands to apply. Exactly one of --sql|--sql-file is required.")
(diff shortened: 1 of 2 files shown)


dbussink (approved):
Is there a way we can easily add a test for this?

ajm188:
> Is there a way we can easily add a test for this?

@dbussink done. this test fails on `vitessio:main`, with:

```
        	Error:      	Received unexpected error:
        	            	invalid argument "\"CREATE TABLE foo(id int not null primary key, name varchar(255)); CREATE TABLE bar (id int not null primary key, foo_id int not null);" for "--sql" flag: parse error on line 1, column 136: extraneous or missing " in quoted-field
        	Test:       	TestApplySchema
```

and passes with my changes

deepthi (commented):
What about the other commands where we are using `StringSliceVar`? I see at least `GetSchema` and `ApplyVSchema`.

frouioui (approved):
This is great. Thank you for fixing this!

ajm188:
> What about the other commands where we are using `StringSliceVar`? I see at least `GetSchema` and `ApplyVSchema`.

in most cases, `StringSliceVar` is what you want (ie. you're taking input of multiple names of "things", and it saves you the hassle of taking a raw string and doing `strings.Split(input, ",")` on it). this is one of the few exceptions, since the input itself can (and _especially_ in the case of sql statements, _almost always_) correctly contain commas, so they can't be considered separators.

PR #11791, opened 2022-11-22
Update to latest etcd release

dbussink:
This release also has MacOS M1 arm64 binaries available, so we can use the upstream artifacts directly and also don't need to fall back to a locally installed etcd.

This way we always install etcd and use a consistent version as well independent of what folks already have installed to make it more reliable.

## Checklist

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -161,9 +161,8 @@ install_etcd() {
   file="etcd-${version}-${platform}-${target}.${ext}"
 
   # This is how we'd download directly from source:
-  # download_url=https://github.com/etcd-io/etcd/releases/download
-  # wget "$download_url/$version/$file"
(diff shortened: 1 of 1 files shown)


vmg (approved):

frouioui (approved):
This looks good to me.

frouioui:
@dbussink You'll need to push an empty commit or amend+force push your current HEAD to re-run the `Check Pull Request labels` workflow which seems to be stuck.

dbussink:
> @dbussink You'll need to push an empty commit or amend+force push your current HEAD to re-run the `Check Pull Request labels` workflow which seems to be stuck.

@frouioui Adding and then removing a label also triggered it and is a lot cheaper :smile:.

PR #11794, opened 2022-11-22
Prevent etcd topology locks from silently expiring.

arthurschreiber:
## Description

If the context passed in to the `Lock` function was canceled / timed out / had exceeded its deadline _after_ the lock was acquired, Vitess silently stopped sending keepAlive requests to etcd, and the lock would be lost after reaching the lease ttl.

This change prevents that from happening by using a fresh context struct for the `KeepAlive` call, which is canceled once the lock is released.

## Related Issue(s)

https://github.com/vitessio/vitess/issues/11811

## Checklist

### DIFF
--- a/go/vt/topo/etcd2topo/lock.go
+++ b/go/vt/topo/etcd2topo/lock.go
@@ -127,8 +127,9 @@ func (s *Server) waitOnLastRev(ctx context.Context, cli *clientv3.Client, nodePa
 
 // etcdLockDescriptor implements topo.LockDescriptor.
 type etcdLockDescriptor struct {
-	s       *Server
-	leaseID clientv3.LeaseID
+	s               *Server
+	leaseID         clientv3.LeaseID
+	keepAliveCancel context.CancelFunc
 }
 
 // TryLock is part of the topo.Conn interface.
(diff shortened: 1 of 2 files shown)


arthurschreiber:
I think this _should_ be backported, as without this change locking with an etcd2 is potentially unsafe. 🤔

arthurschreiber:
Some more context - we encountered this while using VDiff2. When a vdiff workflow was stopped / resumed, we'd sometimes end with the source and target streams using _different_ snapshots, or vreplication streams ending up in weird states. The VDiff code was using a keyspace lock to prevent concurrent modification of `vreplication` tables, but _somehow_ we saw that concurrent modification was happening anyway.

Digging deeper, we noticed that stopping a vdiff workflow is implemented using a cancelable context, and that cancellation would eventually propagate down to the context passed to `Lock`. If the context was canceled _after_ acquiring the lock, the lock was released _before_ the call to `Unlock`, and the call to `Unlock` failed with:

```
E1121 15:00:14.185298   33058 table_differ.go:105] UnlockKeyspace issues_pull_requests_ks failed: <nil>
```

Due to a separate bug in the VDiff code, we can't see the unlock failure message here (only `<nil>`), but I'm pretty sure the unlocking failed because the lock was no longer held.

GuptaManan100:
@arthurschreiber Could you create an issue and link it to this PR? All bug fixes in Vitess now require having an associated issue describing the bug.

arthurschreiber:
@GuptaManan100 I added https://github.com/vitessio/vitess/issues/11811.

GuptaManan100 (changes requested):
I don’t think that the proposed fix for the bug described in the PR is correct. From what I understand, the implicit unlock seems to be the proper behavior. What I think is incorrect is the assumption in the vdiff code to assume that the lock is held even after the context expires. In ERS and PRS we have explicit checks to see if we still hold the lock, for this reason, otherwise, those checks won’t be required since our context could expire and we would continue to hold the lock.

There would be no safety mechanism either if we allowed this change. Like a process gets a lock. For some reason it gets stuck (deadlocked or something else). The context expires but we won’t let go of the lock leading to the entire system being stuck. 
I believe that the lock being released on context expiry isn't just etcd topology specific but a decision decision to guarantee forward progress in face of deadlocks (if any)

PR #11797, opened 2022-11-23
Planner Bug: `bug: not supported pushWeightStringForDistinct on *planbuilder.limit`

arthurschreiber:
## Description

If a query contains more than 64 union statements, it fails with `bug: not supported pushWeightStringForDistinct on *planbuilder.limit` when using the Gen4 planner.

## Related Issue(s)

https://github.com/vitessio/vitess/issues/11805

## Checklist


arthurschreiber:
I think the tableset switches from small to large once we hit 64 tables, and then the planning goes sideways. The 65th query then is seen as a scatter query instead of having a route, and the `LIMIT 11` is represented by an intermediate `Limit` node - and the distinct operator for the `UNION` can't be pushed on a `Limit`

frouioui:
Hello @arthurschreiber! Can you open an issue for this issue and link it to this PR please?

arthurschreiber:
This was fixed by https://github.com/vitessio/vitess/pull/11819

Issue #11793, opened 2022-11-22
Bug Report: `transaction_isolation` level reports as `DEFAULT`


aquarapid:
### Overview of the Issue

```
mysql> select @@transaction_isolation;
+-------------------------+
| @@transaction_isolation |
+-------------------------+
| DEFAULT                 |
+-------------------------+
1 row in set (0.00 sec)
```

### Reproduction Steps

 * Deploy recent/current `main`
 * Connect to `vtgate`
 * Run `select @@transaction_isolation;`
 * Observe result;  `DEFAULT` is not a known isolation level to MySQL clients (e.g. MySQL connector-j), and will result in errors (e.g. failure to connect, since the Java mysql connector introspects this variable upon connection).
 * Doing the same test with `release-15.0` branch produces:
```
mysql> select @@transaction_isolation;
+-------------------------+
| @@transaction_isolation |
+-------------------------+
| REPEATABLE-READ         |
+-------------------------+
1 row in set (0.00 sec)
```
as expected

### Binary Version

```sh
$ vttablet --version
Version: 16.0.0-SNAPSHOT (Git revision 9c0ecb56505397579dd093e452aa6dd15c0e5f27 branch 'main') built on Tue Nov 22 09:00:15 PST 2022 by jacques@dhoomdeskw.localdomain using go1.19.3 linux/amd64
```

### Operating System and Environment details

```sh
Fedora 37, amd64, Kernel 6.0.9-300
```

### Log Fragments

The error you might see from a Spring Boot / Hibernate application might look something like this:

```
Caused by: java.sql.SQLException: Could not map transaction isolation 'DEFAULT' to a valid JDBC level.
        at com.mysql.cj.jdbc.exceptions.SQLError.createSQLException(SQLError.java:129) ~[mysql-connector-java-8.0.29.jar:8.0.29]
        at com.mysql.cj.jdbc.exceptions.SQLError.createSQLException(SQLError.java:97) ~[mysql-connector-java-8.0.29.jar:8.0.29]
        at com.mysql.cj.jdbc.exceptions.SQLError.createSQLException(SQLError.java:89) ~[mysql-connector-java-8.0.29.jar:8.0.29]
        at com.mysql.cj.jdbc.exceptions.SQLError.createSQLException(SQLError.java:63) ~[mysql-connector-java-8.0.29.jar:8.0.29]
[...]

PR #11804, opened 2022-11-23
Fix: return allowed transaction isolation level value on select query

harshit-gangal:
## Description

This PR fixes the regression caused by https://github.com/vitessio/vitess/pull/11673 where when transaction isolation is not set returns `DEFAULT` which is an unknown value from the transaction isolation system variable.
Now, this PR will return the default value from the connection which will be one of these
`READ-UNCOMMITTED`, `READ-COMMITTED`, `REPEATABLE-READ` and `SERIALIZABLE`

## Related Issue(s)

- Fixes https://github.com/vitessio/vitess/issues/11793

## Checklist

### DIFF
--- a/go/vt/sqlparser/ast_rewriting_test.go
+++ b/go/vt/sqlparser/ast_rewriting_test.go
@@ -166,18 +166,6 @@ func TestRewrites(in *testing.T) {
 		expected: "SELECT :__vtworkload as `@@workload`",
 		workload: true,
 	}, {
-		in:          "SELECT @@tx_isolation",
-		expected:    "select :__vttx_isolation as `@@tx_isolation` from dual",
-		txIsolation: true,
-	}, {
-		in:          "SELECT @@transaction_isolation",
-		expected:    "select :__vttransaction_isolation as `@@transaction_isolation` from dual",
-		txIsolation: true,
-	}, {
-		in:          "SELECT @@session.transaction_isolation",
-		expected:    "select :__vttransaction_isolation as `@@session.transaction_isolation` from dual",
-		txIsolation: true,
-	}, {
 		in:       "SELECT @@socket",
 		expected: "SELECT :__vtsocket as `@@socket`",
 		socket:   true,
@@ -317,7 +305,6 @@ func TestRewrites(in *testing.T) {
 		sessTrackGTID:               true,
 		socket:                      true,
 		queryTimeout:                true,
-		txIsolation:                 true,
 	}, {
 		in:                          "SHOW GLOBAL VARIABLES",
 		expected:                    "SHOW GLOBAL VARIABLES",
@@ -337,7 +324,6 @@ func TestRewrites(in *testing.T) {
 		sessTrackGTID:               true,
 		socket:                      true,
 		queryTimeout:                true,
-		txIsolation:                 true,
 	}}
 
 	for _, tc := range tests {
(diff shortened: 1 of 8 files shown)


go/vt/sqlparser/ast_rewriting_test.go
```diff
@@ -442,6 +426,27 @@ func TestRewritesSysVar(in *testing.T) {
...
+	}, {
+		in:       "SELECT @@transaction_isolation",
+		expected: "select @@transaction_isolation from dual",
+	}, {
+		in:       "SELECT @@session.transaction_isolation",
+		expected: "select @@session.transaction_isolation from dual",
+	}, {
+		in:       "SELECT @@tx_isolation",
+		expected: "select :__vttx_isolation as `@@tx_isolation` from dual",
+		sysVar:   map[string]string{"tx_isolation": "'READ-COMMITTED'"},
```
systay, line 441: nit: I found it difficult to read these cases, until I realised that sysVar is an input and not an output. `expected` should come last

go/vt/vtgate/engine/set.go
```diff
@@ -472,11 +472,14 @@ func (svss *SysVarSetAware) Execute(ctx context.Context, vcursor VCursor, env *e
 		if err != nil {
 			return err
 		}
-		out, ok := isolationLevelToExecuteOptionIsolationLevel[strings.ToLower(str)]
+		upperStr := strings.ToUpper(str)
```
systay, line 475: curious to why you switched from lower to upper. not important, just curious :)

go/vt/vtgate/engine/set.go
```diff
@@ -472,11 +472,14 @@ func (svss *SysVarSetAware) Execute(ctx context.Context, vcursor VCursor, env *e
...
+		out, ok := isolationLevelToExecuteOptionIsolationLevel[upperStr]
 		if !ok {
 			return vterrors.NewErrorf(vtrpcpb.Code_INVALID_ARGUMENT, vterrors.WrongValueForVar, "Variable '%s' can't be set to the value of '%s'", svss.Name, str)
 		}
-		vcursor.Session().SetTransactionIsolation(querypb.ExecuteOptions_TransactionIsolation(out))
+		vcursor.Session().SetTransactionIsolation(out)
+		vcursor.Session().SetSysVar(sysvars.TxIsolation.Name, fmt.Sprintf("'%s'", upperStr))
```
systay, line 481: I'm not really understanding why you write it twice to the session under different names. is it really needed?

go/vt/vtgate/engine/set.go
harshit-gangal: mysql returns in upper case, so how we store, we return that after rewrite.

go/vt/vtgate/engine/set.go
harshit-gangal: tx_isolation is the old system variable, transaction_isolation was introduced in 5.7.20. 
In 8.0, tx_isolation is removed.
just to support the case where the client set tx_isolation and selects transaction_isolation, this is done.

systay: ah, good to know. maybe we should add a comment saying this?

systay (approved):

frouioui (approved):
I am not very knowledgable on transaction isolation, but code bits that I understand look good to me.

Issue #11807, opened 2022-11-23
Migrate to GitHub OIDC based auth for Launchable


draftcode:
(The reporter, @draftcode, works for Launchable.)

Launchable is used in the Vitess CI scripts (For example, https://github.com/vitessio/vitess/blob/fb629d3368e6aec9b4be8520167b350554eb324e/.github/workflows/cluster_endtoend_12.yml#L104-L137). Launchable recently implemented a new authentication flow for GitHub Actions, and it's currently doing the migration. For details, see https://docs.launchableinc.com/sending-data-to-launchable/migration-to-github-oidc-auth.

We'd like to migrate Vitess to the new auth method. @draftcode is driving the entire effort on the Launchable side and can open a PR for Vitess. It'll be just one PR/one commit, and it should be easy to revert when there's an issue. When there's an issue for the migration, you can directly ping him (his timezone is Pacific Time, so it's better to make the migration happening during that time). Once migration is done, we can confirm it's working from the Launchable side.

PR #11808, opened 2022-11-23
Migrate to GitHub OIDC based auth for Launchable

draftcode:
## Description

This change updates the CI workflows to use the new GitHub OpenID connect based authentication flow.

GitHub started to provide a public-key signed token that contain pull-request data. This is commonly used as a short-lived token in the authentication flow (Open ID Connect). Launchable recently started supporting this. Migrate to this new method.

See
https://docs.launchableinc.com/sending-data-to-launchable/migration-to-github-oidc-auth for the overview and the process.

## Related Issue(s)

Fixes https://github.com/vitessio/vitess/issues/11807

## Checklist

    - N/A
    - Not required
    - Not required

## Deployment Notes

CI script change only. Even if this goes wrong, the blast radius is contained to the testing process.

### DIFF
--- a/.github/workflows/cluster_endtoend_12.yml
+++ b/.github/workflows/cluster_endtoend_12.yml
@@ -9,12 +9,15 @@ concurrency:
 env:
   LAUNCHABLE_ORGANIZATION: "vitess"
   LAUNCHABLE_WORKSPACE: "vitess-app"
-  GITHUB_PR_HEAD_SHA: "${{ github.event.pull_request.head.sha }}"
+  EXPERIMENTAL_GITHUB_OIDC_TOKEN_AUTH: 1
 
 jobs:
   build:
     name: Run endtoend tests on Cluster (12)
     runs-on: ubuntu-20.04
+    permissions:
+      id-token: write
+      contents: read
 
     steps:
     - name: Skip CI
--- a/.github/workflows/cluster_endtoend_13.yml
+++ b/.github/workflows/cluster_endtoend_13.yml
@@ -9,12 +9,15 @@ concurrency:
 env:
   LAUNCHABLE_ORGANIZATION: "vitess"
   LAUNCHABLE_WORKSPACE: "vitess-app"
-  GITHUB_PR_HEAD_SHA: "${{ github.event.pull_request.head.sha }}"
+  EXPERIMENTAL_GITHUB_OIDC_TOKEN_AUTH: 1
 
 jobs:
   build:
     name: Run endtoend tests on Cluster (13)
     runs-on: ubuntu-20.04
+    permissions:
+      id-token: write
+      contents: read
 
     steps:
     - name: Skip CI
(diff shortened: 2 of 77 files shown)


harshit-gangal (approved):

GuptaManan100 (approved):

harshit-gangal:
@draftcode none of the e2e test ran which send test report to launchable.

frouioui (dismissed):
Hello @draftcode! Thanks a lot for your contribution.

The workflows cannot run as their syntax is incorrect. Referring to GitHub Actions documentation: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions, there is no such `permissions` field for a job's step. I think what you'll want to do is to set the permissions at the job level instead of the step level. This is explained in: https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

```yaml
jobs:
  myjob:
    runs-on: ubuntu-latest

    permissions:
      id-token: write
```

Reading through the issue, I think the Launchable documentation (https://docs.launchableinc.com/sending-data-to-launchable/migration-to-github-oidc-auth) is misleading as it instructs us to add the `permissions` field to the `step` instead of the `job`.

draftcode:
Oh. Thanks! I appreciate you debugging this for me. I updated the PR and will update the doc as well. Let's see if e2e runs.

draftcode:
(Hmm. Removal is not my intention. I clicked the re-review button, and it removed the rest of the people automatically.)

.github/workflows/cluster_endtoend_12.yml
```diff
@@ -9,12 +9,15 @@ concurrency:
...
-  GITHUB_PR_HEAD_SHA: "${{ github.event.pull_request.head.sha }}"
+  EXPERIMENTAL_GITHUB_OIDC_TOKEN_AUTH: 1
 
 jobs:
   build:
     name: Run endtoend tests on Cluster (12)
     runs-on: ubuntu-20.04
+    permissions:
+      id-token: write
+      contents: read
```
frouioui, line 20: Thank you for fixing the `permissions` key, it seems to run fine now 🥳

However, can we limit the scope to `id-token: write` only? Your documentation only asks for `id-token`, not `contents`.

draftcode: Please see the commit message. tldr is that contents: read is given by default if unspecified, and this is added to partially recreate the default. The existing steps implicitly have been using it. Launchable's CLI itself won't need it.

draftcode: https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token is the default config when "permissions" part doesn't exist. If we drop "contents: read", the e2e will fail at `actions/checkout@v3`.

frouioui: Awesome! Looks good to me :)

frouioui (approved):

frouioui:
@draftcode thank you for your contribution! 🥳

PR #11813, opened 2022-11-24
Optimize Mysql56GTIDSet.String(), add benchmark

timvaillancourt:
## Description

This PR optimizes the `.String()` method of `mysql.Mysql56GTIDSet` by appending the byte-representation of the GTID-SID to the `bytes.Buffer` vs converting and writing it to the buffer as a `string`. Also a benchmark test for the `.String()` method was added

I also attempted to optimize further by writing the `start`/`end` intervals _(`int64`s)_ to the `bytes.Buffer` using `binary.Write()`, but confusingly, this performed worse than this change and I can't explain why 😕

In benchmarks this PR improves the efficiency of `.String()` by about 27%: `200.9 ns/op` vs `275.5 ns/op`

#### Before benchmark
```bash
$ go test -v -run BenchmarkMysql56GTIDSetString -bench=BenchmarkMysql56GTIDSetString ./go/mysql/...
goos: darwin
goarch: amd64
pkg: [vitess.io/vitess/go/mysql](http://vitess.io/vitess/go/mysql)
cpu: Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
BenchmarkMysql56GTIDSetString
BenchmarkMysql56GTIDSetString-16    	 4378540	       275.5 ns/op
PASS
```

#### After benchmark
```bash
$ go test -v -run BenchmarkMysql56GTIDSetString -bench=BenchmarkMysql56GTIDSetString ./go/mysql/...
goos: darwin
goarch: amd64
pkg: [vitess.io/vitess/go/mysql](http://vitess.io/vitess/go/mysql)
cpu: Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
BenchmarkMysql56GTIDSetString
BenchmarkMysql56GTIDSetString-16    	 6013492	       200.9 ns/op
PASS
```

If possible, please backport this to v13+ 🙇 

## Checklist

### DIFF
--- a/go/mysql/mysql56_gtid_set.go
+++ b/go/mysql/mysql56_gtid_set.go
@@ -155,7 +155,7 @@ func (set Mysql56GTIDSet) String() string {
 		if i != 0 {
 			buf.WriteByte(',')
 		}
-		buf.WriteString(sid.String())
+		buf.Write(sid[:])
 
 		for _, interval := range set[sid] {
 			buf.WriteByte(':')
--- a/go/mysql/mysql56_gtid_set_test.go
+++ b/go/mysql/mysql56_gtid_set_test.go
@@ -151,6 +151,14 @@ func TestMysql56GTIDSetString(t *testing.T) {
 	}
 }
 
+func BenchmarkMysql56GTIDSetString(b *testing.B) {
+	sid := SID{0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15}
+	set := Mysql56GTIDSet{sid: []interval{{1, 5}}}
+	for n := 0; n < b.N; n++ {
+		set.String()
+	}
+}
+
 func TestMysql56GTIDSetFlavor(t *testing.T) {
 	input := Mysql56GTIDSet{}
 	if got, want := input.Flavor(), "MySQL56"; got != want {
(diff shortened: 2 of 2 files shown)


go/mysql/mysql56_gtid_set.go
```diff
@@ -155,7 +155,7 @@ func (set Mysql56GTIDSet) String() string {
 		if i != 0 {
 			buf.WriteByte(',')
 		}
-		buf.WriteString(sid.String())
+		buf.Write(sid[:])
```
dbussink, line 158: I don't think this is equivalent? It also shows in the `endtoend` test suite:

```
vttablet: rpc error: code = Internal desc = could not decode position: invalid MySQL 5.6 GTID set ("\xd1\xf8\b\x8dl.\x11\xed\x8ao\x00\r:;\xcep:1-144"): invalid MySQL 5.6 SID "\xd1\xf8\b\x8dl.\x11\xed\x8ao\x00\r"
[1090](https://github.com/vitessio/vitess/actions/runs/3543219398/jobs/5949479892#step:12:1091)
```

This writes the bytes directly, but `sid.String()` converts it to a hex representation for the string which other places depend on.

timvaillancourt: @dbussink 👋 yeah, I'm seeing the same test failures now. I'm not sure how I missed that 🤦 

I'll close this PR for now and look further into it

PR #11817, opened 2022-11-25
[planner bugfix] make sure queries with more than 64 tables work

### LINKED ISSUE #11805: Bug Report: `bug: not supported pushWeightStringForDistinct on *planbuilder.limit`
### Overview of the Issue

If a query contains more than 64 tables joined via union statements, it fails with `bug: not supported pushWeightStringForDistinct on *planbuilder.limit` when using the Gen4 planner.

### Reproduction Steps

See https://github.com/vitessio/vitess/pull/11797 for an reproducible example.

### Binary Version

```sh
v15.0.0
```

```sh
N/A
```

### Log Fragments

_No response_

systay:
## Description
If a query contains more than 64 tables, the bitset we use to keep track of tables goes from being just a uint64 to be a slice of uint64 values. Once that happens, we can't compare bitsets using `=` and instead must use the `Equals` method on the TableSet.

## Related Issue(s)
Fixes #11797
Fixes #11805

## Checklist


go/vt/vtgate/planbuilder/operators/querygraph.go
```diff
@@ -143,7 +143,7 @@ func (qg *QueryGraph) collectPredicate(ctx *plancontext.PlanningContext, predica
 
 func (qg *QueryGraph) addToSingleTable(table semantics.TableSet, predicate sqlparser.Expr) bool {
 	for _, t := range qg.Tables {
-		if table == t.ID {
+		if table.Equals(t.ID) {
```
arthurschreiber, line 146: ✨

arthurschreiber (approved):

systay:
found more issues around TableSet comparisons. will mark as ready for review when we have a fix for those

PR #11819, opened 2022-11-25
semantics: Use a BitSet

### LINKED ISSUE #11805: Bug Report: `bug: not supported pushWeightStringForDistinct on *planbuilder.limit`

vmg:
Following up on https://github.com/vitessio/vitess/pull/11817 -- the bug was caused because our current implementation of `TableSet` is not strictly comparable with equality (i.e. `==`).

This PR fixes the issue in depth by bringing in a new implementation of an immutable bitset that is safe to compare with equality _and_ to be used as a key in a hashmap (something we also do quite often).

Fixes #11805 

cc @systay


vmg:
There's a new commit that fixes some of the superluous renaming but it's not showing up in the web UI. I think GitHub is having issues again.

systay (approved):

arthurschreiber:
@vmg Mind renaming `tabletset.go` to `table_set.go` while you're touching the file? 😅

harshit-gangal:
@vmg could you add more documentation to the bitset implementation about how it works? Also, any place where I can read more about it?

arthurschreiber:
Sorry for the nitpicking, but I think the filename is still incorrect - it's currently `tablet_set.go` instead of `table_set.go`. 🙈

go/vt/vtgate/semantics/tablet_set.go
```diff
@@ -0,0 +1,129 @@
...
+	for _, t := range tss {
+		result = result.Or(bitset.Bitset(t))
+	}
+	return TableSet(result)
+}
+
+// TableSetFromIds returns TableSet for all the id passed in argument.
+func TableSetFromIds(tids ...int) (ts TableSet) {
+	return TableSet(bitset.Build(tids...))
+}
```
harshit-gangal, line 129: this method is not used in the codebase which will also imply that `Build` method is not required in `bitset`

harshit-gangal:
> @vmg could you add more documentation to the bitset implementation about how it works? Also, any place where I can read more about it?

it was nice reading some math-intensive code.

go/vt/vtgate/semantics/bitset/bitset.go
```diff
@@ -0,0 +1,241 @@
...
+		bitset := bs[i]
+		for bitset != 0 {
+			t := bitset & -bitset
+			r := bits.TrailingZeros8(bitset)
+			yield(i*bitsetWidth + r)
+			bitset ^= t
+		}
+	}
+}
+
```
harshit-gangal, line 197: an example of how this works will help. A general example for overall implementation will also work.

go/vt/vtgate/semantics/tablet_set.go
vmg: We intend to use this method in some PlanetScale-internal projects!

go/vt/vtgate/semantics/bitset/bitset.go
```diff
@@ -0,0 +1,241 @@
...
+		words[b/bitsetWidth] |= 1 << (b % bitsetWidth)
+	}
+	return toBitset(words)
+}
+
+const singleton = "\x00\x00\x00\x01\x00\x00\x00\x02\x00\x00\x00\x04\x00\x00\x00\x08\x00\x00\x00\x10\x00\x00\x00\x20\x00\x00\x00\x40\x00\x00\x00\x80"
+
+// Single returns a new Bitset where only the given bit is set.
+// If the given bit is less than 32, Single does not allocate to create a new Bitset.
+func Single(bit int) Bitset {
```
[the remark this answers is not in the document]
vmg: There is no limit! Any bitset larger than 32 will allocate, but you can make it as large as you want.

frouioui (approved):

go/vt/vtgate/semantics/tablet_set.go
harshit-gangal: I would suggest having at least a test for it so that if anyone uses it here, does not break the usage.

harshit-gangal (approved):

GuptaManan100 (approved):

PR #11823, opened 2022-11-27
BugFix: Escaping Percentage and Underscore require special handling

### LINKED ISSUE #11599: Escape symbols in LIKE clause are removed by vtgate
### Overview of the Issue

I have build a testcase:
Prepared a SQL and ran it through vtgate:

`select 'yan' from jobs where partner_id = 123 and partner_job_code like '\_yanzT4C4j\_7yLWngWP93ZtXgnukm6NuPchqtS2u0mHbSnI\%';
`
In MySQL processlist I see Escape slash characters are removed.`
| 16793605 | vt_app          | localhost                    | vt_test      | Query            |       4 | executing                                                       | select 'yan' from jobs where partner_id = 3790221 and partner_job_code like '_yanzT4C4j_7yLWngWP93ZtXgnukm6NuPchqtS2u0mHbSnI%' limit 10000000001 |`

Any way to avoid it Vitess? It breaks functionality of SQL.

### Reproduction Steps

Run a SQL with LIKE and escape character with backslash. It will be removed.

### Binary Version

```sh
Version: 11.0.4 (Git revision cb142ee branch 'HEAD') built on Mon Jan 10 15:55:07 UTC 2022 by runner@fv-az201-838 using go1.16.13 linux/amd64
```

```sh
Linux 3.10.0-1160.53.1.el7.x86_64 #1 SMP Thu Dec 16 10:19:28 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
```

### Log Fragments

_No response_

GuptaManan100:
## Description

This PR addresses the issue described in #11599. The escaping rules used by MySQL are described in the manual https://dev.mysql.com/doc/refman/5.7/en/string-literals.html. The escaping of `%` and `_` is special. The `\%` and `\_` sequences are used to search for literal instances of `%` and `_` in pattern-matching contexts where they would otherwise be interpreted as wildcard characters. If you use `\%` or `\_` outside of pattern-matching contexts, they evaluate to the strings `\%` and `\_`, not to `%` and `_`.

In Vitess, the way we are choosing to handle this behavior is to always preserve the escaping of `%` and `_` as is in all the places and handle it like MySQL in our evaluation engine for `Like`.

The encoding and decoding of strings have been fixed to this end. The evalengine doesn't require any changes and it already handles the escaping correctly.

Unit tests and end-to-end tests for testing this escaping have been added for pass-through queries and also for the ones that require evaluation on the vtgate level.

## Related Issue(s)

Fixes #11599

## Checklist


go/sqltypes/value.go
```diff
@@ -548,7 +548,12 @@ func encodeBytesSQL(val []byte, b BinWriter) {
 
 func encodeBytesSQLBytes2(val []byte, buf *bytes2.Buffer) {
 	buf.WriteByte('\'')
-	for _, ch := range val {
+	for idx, ch := range val {
+		// If \% or \_ is present, we want to keep them as is, and don't want to escape \ again
+		if ch == '\\' && idx+1 < len(val) && (val[idx+1] == '%' || val[idx+1] == '_') {
+			buf.WriteByte(ch)
+			continue
+		}
```
GuptaManan100, line 556: I had to duplicate this code in all three encoding functions we have. I couldn't come up with a better way 😢. We can merge two of them using an interface which requires a `WriteByte` function, but the last remains separate.

GuptaManan100: In this light, I have not merged any of these functions. Open to suggestions for improvement.

frouioui: we could extract `ch == '\\' && idx+1 < len(val) && (val[idx+1] == '%' || val[idx+1] == '_')` to a method to avoid duplicate, and send in `ch`, `idx`, `val`

frouioui: otherwise it looks good

frouioui (approved):

go/sqltypes/value.go
```diff
@@ -548,7 +548,12 @@ func encodeBytesSQL(val []byte, b BinWriter) {
 
 func encodeBytesSQLBytes2(val []byte, buf *bytes2.Buffer) {
 	buf.WriteByte('\'')
-	for _, ch := range val {
+	for idx, ch := range val {
+		// If \% or \_ is present, we want to keep them as is, and don't want to escape \ again
+		if ch == '\\' && idx+1 < len(val) && (val[idx+1] == '%' || val[idx+1] == '_') {
```
harshit-gangal, line 553: is '%' and '_' the only ones that need this change or there are others as well?

harshit-gangal (approved):

go/sqltypes/value.go
GuptaManan100: These are the only ones. I'll caveat by adding that these are the only ones according to the MySQL documentation! 😆

go/sqltypes/value.go
GuptaManan100: I tried it out, but that too doesn't seem to work all that well, since val is a byte array in two of the funcs and a string in the other, and ch is a byte in two and rune in the other. I am leaving this as is and we can iterate on this later.

PR #11838, opened 2022-11-29
Remove Tests from Self-hosted runners

GuptaManan100:
## Description

We don't need to run the clusters 12 and 18 on self-hosted runners in release-14.0 since no one is actively using the results of these failures (if any) to debug anything.

## Related Issue(s)

- https://github.com/vitessio/vitess/issues/11840

## Checklist


.github/workflows/cluster_endtoend_onlineddl_revert_mysql80.yml
```diff
@@ -38,7 +38,8 @@ jobs:
             - 'test.go'
             - 'Makefile'
             - 'build.env'
-            - 'go.[sumod]'
+            - 'go.sum'
+            - 'go.mod'
```
GuptaManan100, line 42: Running the ci workflow generator changed a bunch of other tests too, but that is unrelated to this PR.

frouioui: sounds good

frouioui (approved):

rohit-nayak-ps (approved):

PR #11842, opened 2022-11-29
Fix closing the body for HTTP requests

### LINKED ISSUE #11841: Bug Report: leaking HTTP response bodies
### Overview of the Issue

There's several places where we leak HTTP response bodies and don't close them. It means we have goroutines for those leaking.

It's almost all test cases, but there's one specific production case where we have this for non test logic:

https://github.com/vitessio/vitess/blob/e84bfdb217238370139c64fdae81d4d77b896cb8/go/vt/vttablet/tabletserver/throttle/throttler.go#L664-L673

Here we do a `GET` request, read the body but then don't close the body. 

### Reproduction Steps

Running the throttler means we leak resources

### Binary Version

```sh
-
```

```sh
-
```

### Log Fragments

_No response_

dbussink:
One of the easy things to miss that can cause resource leaks is not closing a response body for an HTTP request.

There's a linter available to make it easier to catch this, so this enables that linter and fixes all the cases reported.

This is almost all test cases, except for one production code path case in the throttler.

## Related Issue(s)

Fixes #11841

## Checklist

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -29,6 +29,7 @@ linters:
     - gofmt
     - goimports
     - exportloopref
+    - bodyclose
 
     # revive is a replacement for golint, but we do not run it in CI for now.
     # This is only enabled as a post-commit hook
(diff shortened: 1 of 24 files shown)


shlomi-noach (approved):
thorough as always

frouioui (approved):

harshit-gangal (approved):

PR #11843, opened 2022-11-29
Simplify recursive data structure in CFC Vindex

harshit-gangal:
## Description

This PR simplifies the self-reference of the CFC struct by prefix Vindex in the CFC Vindex.
Prefix Vindex is created when it is required rather than upfront. This resolved the cyclic dependency which was causing cache_size to crash with stack full error.

## Related Issue(s)

- Fixes CFCInsert Flakyness reported via launchable
- Fixes https://github.com/vitessio/vitess/issues/11858

## Checklist

### DIFF
--- a/go/vt/vtgate/executor_framework_test.go
+++ b/go/vt/vtgate/executor_framework_test.go
@@ -177,7 +177,10 @@ var executorVSchema = `
 			"params": {
 				"region_bytes": "1"
 			}
-    	}
+    	},
+		"cfc": {
+          "type": "cfc"
+		}
 	},
 	"tables": {
 		"user": {
@@ -371,6 +374,20 @@ var executorVSchema = `
 				{
 					"columns": ["cola","colb"],
 					"name": "regional_vdx"
+				}
+			]
+    	},
+		"tbl_cfc": {
+			"column_vindexes": [
+                {
+                    "column": "c1",
+                    "name": "cfc"
+                }
+			],
+			"columns": [
+				{
+					"name": "c2",
+					"type": "VARCHAR"
 				}
 			]
     	}
--- a/go/vt/vtgate/executor_framework_test.go
+++ b/go/vt/vtgate/executor_framework_test.go
@@ -179,7 +179,7 @@ var executorVSchema = `
 			}
     	},
 		"cfc": {
-          "type": "cfc"
+			"type": "cfc"
 		}
 	},
 	"tables": {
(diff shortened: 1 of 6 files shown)


GuptaManan100:
The unit tests are failing. Could you take a look at those?

harshit-gangal:
> The unit tests are failing. Could you take a look at those?

updated test expectation.

harshit-gangal:
cc: @wangmeng99

go/vt/vtgate/executor_framework_test.go
```diff
@@ -177,7 +177,10 @@ var executorVSchema = `
 			"params": {
 				"region_bytes": "1"
 			}
-    	}
+    	},
+		"cfc": {
+          "type": "cfc"
```
deepthi, line 182: The formatting looks off. I'm surprised that `gofmt` is passing though.

GuptaManan100: This is in a string, so I think that's why `gofmt` is not complaining

go/vt/vtgate/vindexes/cfc.go
```diff
@@ -244,7 +241,8 @@ func (vind *CFC) Map(ctx context.Context, vcursor VCursor, ids []sqltypes.Value)
 
 // PrefixVindex switches the vindex to prefix mode
 func (vind *CFC) PrefixVindex() SingleColumn {
-	return vind.prefixVindex
+	// prefixCFC is only used in 'LIKE' compare expressions.
+	return &prefixCFC{CFC: vind}
 }
```
GuptaManan100, line 246: Everytime we call this `PrefixVindex()` function, it would return a new struct. Is that what we want?

deepthi:
Does this failure occur only in CI, or is it possible in production? If it can fail in production we should document the expected failure as a bug (with an issue).

go/vt/vtgate/vindexes/cfc.go
harshit-gangal: that is not ideal, let me re-write this part.

harshit-gangal:
> Does this failure occur only in CI, or is it possible in production? If it can fail in production we should document the expected failure as a bug (with an issue).

It is a production bug for users using CFC vindex.

GuptaManan100 (approved):
Looks good to me now other than the indentation Deepthi pointed out

go/vt/vtgate/executor_framework_test.go
harshit-gangal: fixed.

harshit-gangal:
> Looks good to me now other than the indentation Deepthi pointed out

that is resolved.

systay (commented):
LGTM. Only thing missing would an issue for the bugfix, right?

PR #11844, opened 2022-11-29
Fix deprecated usage of set-output

dbussink:
The `set-output` command is deprecated and will break in the future. This moves to the new style for all our GitHub Actions.

See also https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

Also marking this for back port so we don't use it on supported releases either once this is really removed. 

## Checklist

### DIFF
--- a/.github/workflows/check_make_vtadmin_authz_testgen.yml
+++ b/.github/workflows/check_make_vtadmin_authz_testgen.yml
@@ -21,7 +21,7 @@ jobs:
           skip='true'
         fi
         echo Skip ${skip}
-        echo "::set-output name=skip-workflow::${skip}"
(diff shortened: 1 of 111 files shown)


frouioui (approved):
LGTM

frouioui:
CI is failing with the following error: `next_release_ref=${$next_release_ref}: bad substitution`, I think we can get rid of the inner `$` right?

dbussink:
> CI is failing with the following error: `next_release_ref=${$next_release_ref}: bad substitution`, I think we can get rid of the inner `$` right?

Yeah, already saw that and fixed it up. Bad copy paste 😞

deepthi (approved):

PR #11845, opened 2022-11-29
Back to dev mode after v14.0.4

frouioui:
Includes the changes required to go back into dev mode (v14.0.5-SNAPSHOT) after the release of v14.0.4.

### DIFF
--- a/docker/release.sh
+++ b/docker/release.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -ex
 
-vt_base_version='v13.0.0'
+vt_base_version='14.0.5-SNAPSHOT'
 debian_versions='buster  bullseye'
 default_debian_version='bullseye'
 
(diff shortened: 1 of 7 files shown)


docker/release.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -ex
 
-vt_base_version='v13.0.0'
+vt_base_version='14.0.5-SNAPSHOT'
```
deepthi, line 4: This doesn't seem right, it should be `v14.0.5`.

deepthi:
Moving this to draft because it should be merged after #11846

docker/release.sh
GuptaManan100: Like the other files that have the version it is kept at `-SNAPSHOT`. The release script changes it to the release version and then back to a snapshot version

deepthi (approved):

PR #11848, opened 2022-11-29
Back to dev mode after v15.0.1

rsajwani:
Includes the changes required to go back into dev mode (v15.0.2-SNAPSHOT) after the release of v15.0.1.

### DIFF
--- a/docker/release.sh
+++ b/docker/release.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -ex
 
-vt_base_version='v15.0.1'
+vt_base_version='15.0.2-SNAPSHOT'
 debian_versions='buster  bullseye'
 default_debian_version='bullseye'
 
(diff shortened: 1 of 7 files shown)


frouioui (approved):

frouioui (approved):

mattlord (approved):

docker/release.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -ex
 
-vt_base_version='v15.0.1'
+vt_base_version='15.0.2-SNAPSHOT'
```
deepthi, line 4: @frouioui I believe this needs to be `v15.0.2`. Can you either confirm, or tell me why the format has been changed?

frouioui: This is correct.

PR #11769, opened 2022-11-17
Move to MySQL 8.0 as the default template generation (#11153)

rsajwani:
* Move towards MySQL 8.0 as the default template generation

This upgrades the remaining things to Ubuntu 20.04 and makes MySQL 8.0 the default we run tests against. We still have tests for MySQL 5.7 but those are now explicitly opted into.

This should finish up the Ubuntu 20.04 upgrade and also makes things easier for the future when we need to upgrade again.

Signed-off-by: Dirkjan Bussink <d.bussink@gmail.com>

* CI: rename shard vtorc_8.0 to vtorc_5.7, change expected test output for 8.0

Signed-off-by: deepthi <deepthi@planetscale.com>

* CI: increase timeout for 8.0 tests on the actual test step from 30 to 45 mins

Signed-off-by: deepthi <deepthi@planetscale.com>

* CI: increase timeout to 45 minutes for mysql57 tests too. We really only need this for vtorc, but I've made the change to the template so all tests get it.

Signed-off-by: deepthi <deepthi@planetscale.com>

* CI: fix vtorc test to work with both 5.7 and 8.0

Signed-off-by: deepthi <deepthi@planetscale.com>

* CI: missed docker flag in mysql57 template, one more fix to vtorc test

Signed-off-by: deepthi <deepthi@planetscale.com>

* removing spaces from pb file

Signed-off-by: Rameez Sajwani <rameezwazirali@hotmail.com>

* removing spaces in pb file part 2

Signed-off-by: Rameez Sajwani <rameezwazirali@hotmail.com>

Signed-off-by: Dirkjan Bussink <d.bussink@gmail.com>
Signed-off-by: deepthi <deepthi@planetscale.com>
Signed-off-by: Rameez Sajwani <rameezwazirali@hotmail.com>
Co-authored-by: deepthi <deepthi@planetscale.com>
Co-authored-by: Rameez Sajwani <rameezwazirali@hotmail.com>

## Description

There are two main changes in CI workflow been done as part of this PR.
- Move to ubuntu 20.04
- Upgrade Mysql 5.7 to Mysql 8.0.31 (latest)

This PR cherry-pick (#11153) the change from main where we have made Mysql8.0.x as default database (from Mysql 5.7). We used the same approach as used in 'main' which is using mysql-apt-config to install mysql server. Currently it is 8.0.31.

A lot of vreplication tests were failing. The reason for failure was the machine was running out of resources. We change innodb_buffer_pool_size = 16M in order to get around that.

[...]

PR #11866, opened 2022-12-01
Fix deprecated usage of set-output (#11844) (#11855)

dbussink:
The `set-output` command is deprecated and will break in the future. This moves to the new style for all our GitHub Actions.

See also https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

Also upgrades some more runners to Ubuntu 20.04 that were missed in previous backports.

Signed-off-by: Dirkjan Bussink <d.bussink@gmail.com>

## Checklist

### DIFF
--- a/.github/workflows/cluster_endtoend_26.yml
+++ b/.github/workflows/cluster_endtoend_26.yml
@@ -9,7 +9,7 @@ concurrency:
 jobs:
   build:
     name: Run endtoend tests on Cluster (26)
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
 
     steps:
     - name: Check if workflow needs to be skipped
@@ -20,7 +20,7 @@ jobs:
           skip='true'
         fi
         echo Skip ${skip}
-        echo "::set-output name=skip-workflow::${skip}"
+        echo "skip-workflow=${skip}" >> $GITHUB_OUTPUT
 
     - name: Check out code
       if: steps.skip-workflow.outputs.skip-workflow == 'false'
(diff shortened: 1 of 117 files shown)


.github/workflows/cluster_endtoend_26.yml
```diff
@@ -9,7 +9,7 @@ concurrency:
 jobs:
   build:
     name: Run endtoend tests on Cluster (26)
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
```
dbussink, line 12: @rsajwani Looks like you missed some cases in #11769, so I added them here.

.github/workflows/upgrade_downgrade_test_backups_e2e.yml
```diff
@@ -44,13 +44,13 @@ jobs:
...
           echo $previous_release_ref
-          echo "::set-output name=previous_release_ref::${previous_release_ref}"
+          echo "previous_release_ref=${previous_release_ref}" >> $GITHUB_OUTPUT
 
   upgrade_downgrade_test_e2e:
     timeout-minutes: 60
     if: always() && (needs.get_previous_release.result == 'success')
     name: Run Upgrade Downgrade Test
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
```
rsajwani, line 53: This might be delibrate.. downgraded version doesn't handled backing-up mysql8.0 correctly .. We will see if the test fails..

deepthi (approved):

.github/workflows/upgrade_downgrade_test_backups_e2e.yml
dbussink: @rsajwani The goal of upgrading was that we can still run CI when GitHub removes 18.04 support. So we only have 2 choices, either we upgrade or we remove the CI job. We can't leave it on 18.04 since that will stop working then. 

We can also look at forcing a MySQL version earlier than 8.0.30 to fix this?

rsajwani: @dbussink , did you check all the CI workflows were passing before pushing this PR ?? 
ideally we shouldn't have pushed this PR until all the test are passing. I can see Run Upgrade Downgrade Test was failing...

PR #11871, opened 2022-12-02
Parse binlog variable length encoded columns

### LINKED ISSUE #11870: Bug Report: Binlog parser fails on tables with 251 columns or more
### Overview of the Issue

Whenever a table has 251 or more columns, it will fail when we use bin logs for that table. This happens in vreplication, online DDL etc. This means that any table with that many columns can't be changed with online DDL.

It results in an error like:

```
 "error": "vttablet: rpc error: code = Unknown desc = stream (at source tablet) error @ x-x-x-x-x:1-x: Code: INTERNAL\nunexpected metadata end: got 536 was expecting 353 (data=[x])"
```

There is already a `FIXME` annotation in the code that this is a known issue, see:

https://github.com/vitessio/vitess/blob/22067c7a4e334a7e90850a2119135e6b1ae23279/go/mysql/binlog_event_rbr.go#L74-L76

MySQL variable length encoded integers are defined in https://dev.mysql.com/doc/dev/mysql-server/latest/page_protocol_basic_dt_integers.html and we already have the code to deal with these, such as reading them in:

https://github.com/vitessio/vitess/blob/22067c7a4e334a7e90850a2119135e6b1ae23279/go/mysql/encoding.go#L212

### Reproduction Steps

Create a table, try to run online DDL so that the number of columns is 251 or more and see it fail.

### Binary Version

```sh
Any supported version (and going back to at least Vitess 3 and likely earlier).
```

```sh
n/a
```

### Log Fragments

_No response_

dbussink:
The MySQL binlog format uses variable length encoded integers for the number of columns and the metadata length. Before the Vitess parsing would assume only one byte lengths, breaking anything with 251 or more columns.

See also https://dev.mysql.com/doc/dev/mysql-server/latest/classbinary__log_1_1Table__map__event.html and the referenced https://dev.mysql.com/doc/dev/mysql-server/latest/classbinary__log_1_1Binary__log__event.html#packed_integer for this format.

This is the MySQL length encoded integer type, also described at https://dev.mysql.com/doc/dev/mysql-server/latest/page_protocol_basic_dt_integers.html.

We already have `lenEncIntSize` to compute the size of the encoded integer, `writeLenEncInt` to write a length encoded integer and `readLenEncInt` to read them. This uses all those utility functions.

I've also marked this for back porting since this seems like a significant bug, although having 251 or more columns is I guess not that common or we'd have hit this before more often.

@shlomi-noach I think it would be nice to also have a full end-to-end test for online DDL with a table with many columns, but you probably know how to add that much better than I do.

## Related Issue(s)

Fixes #11870 

## Checklist


go/mysql/binlog_event_make.go
```diff
@@ -296,9 +296,9 @@ func NewTableMapEvent(f BinlogFormat, s *FakeBinlogStream, tableID uint64, tm *T
 		1 + // table name length
 		len(tm.Name) +
 		1 + // [00]
-		1 + // column-count FIXME(alainjobart) len enc
+		lenEncIntSize(uint64(len(tm.Types))) + // column-count len enc
```
deepthi, line 299: Looks like `lenEncIntSize` expects a uint64 and hence the type coercion.
Do you think it's worth putting in a comment line for future readers/editors?

go/mysql/binlog_event_rbr.go
```diff
@@ -71,21 +71,25 @@ func (ev binlogEvent) TableMap(f BinlogFormat) (*TableMap, error) {
...
-	pos++
+	columnCount, read, ok := readLenEncInt(data, pos)
+	if !ok {
+		return nil, vterrors.Errorf(vtrpc.Code_INTERNAL, "expected column count at position %v (data=%v)", pos, data)
+	}
+	pos = read
 
-	result.Types = data[pos : pos+columnCount]
-	pos += columnCount
+	result.Types = data[pos : pos+int(columnCount)]
```
deepthi, line 80: Why do we need the type coercion here? because `readLenEncInt` returns `uint64`.
I assume this is safe only because we don't expect columnCount to ever be large enough for this coercion to turn it into a negative number.
These coercions do smell though. Do you think it's possible to clean them up?

deepthi (commented):
Nice work!
The only concern is that there isn't a new unit test in binlog_event_rbr_test.go for the other side (read versus write).
In fact at first glance, test coverage for binlong_event_rbr.go seems thin.
Would you mind running `go test -cover` on this package and see what we get for this file?

go/mysql/binlog_event_rbr.go
dbussink: Go needs to always have an `int` to address within a slice. 

> I assume this is safe only because we don't expect columnCount to ever be large enough for this coercion to turn it into a negative number.

There's no way this could cause problems unless we want to also defend against actively malicious binlogs? But then we have a much broader problem here that needs an extensive effort. 

MySQL has a hard limit of 4096 columns, so that's nowhere near any level where this is a problem. 

> These coercions do smell though. Do you think it's possible to clean them up?

No, not unless we also change the signature of functions like `readLenEncInt`. But I don't think it's a practical problem given existing MySQL limits.

go/mysql/binlog_event_make.go
dbussink: Not really, if the coercion is not there, the code does not compile so it seems like a comment is about then something the compiler already enforces as well?

dbussink:
> The only concern is that there isn't a new unit test in binlog_event_rbr_test.go for the other side (read versus write).
In fact at first glance, test coverage for binlong_event_rbr.go seems thin.

All existing tests are in `binlog_event_make_test.go` but they cover both reading and writing, since it first creates binlog events and then parses it. So coverage wise, we do handle both reads and writes.

shlomi-noach (approved):
LGTM based on @dbussink 's comments and discussion with @deepthi . The one thing about testing, which I also had trouble in this package, is that we create our own event, then validate parsing the veent which we created ourselves. This is not as strong as parsing an actual binary sequence from a real binary log.

dbussink:
> LGTM based on @dbussink 's comments and discussion with @deepthi . The one thing about testing, which I also had trouble in this package, is that we create our own event, then validate parsing the veent which we created ourselves. This is not as strong as parsing an actual binary sequence from a real binary log.

@shlomi-noach Yeah, that's why I asked if you'd be able to help with a real end-to-end test for online DDL on a table with a large number of columns. That then would function as the way to validate against a real binary log as well.

shlomi-noach:
> that's why I asked if you'd be able to help with a real end-to-end test for online DDL on a table with a large number of columns. That then would function as the way to validate against a real binary log as well.

Ah, great, yes! Let me push a test into this PR

shlomi-noach:
Added an `endtoend` test. Affected CI: `onlineddl_vrepl_suite`

dbussink:
> Added an `endtoend` test. Affected CI: `onlineddl_vrepl_suite`

Is a create here enough? Does that trigger the issue already or do we need an `alter table` that adds more columns with a table with data in it as well to force data in the binlog?

shlomi-noach:
The test should be enough, but for sake of completeness I added another test that adds a column on top of 257 existing columns.

shlomi-noach:
just to elaborate, the first test runs a mysql trigger that inserts data to a 258 (`id`+257 columns) table, then updates the last row, and specifically updates col `i257`. This is enough to check vreplication's parsing of `> 256` columns and specifically reading the value of a last column.

shlomi-noach:
Verified that the new tests fail on `main`.

GuptaManan100 (approved):
LGTM! i was not aware that we had this limitation! 💯

rohit-nayak-ps (approved):
Nice. 

Crazy we haven't hit this yet!

PR #11874, opened 2022-12-04
Migrate Materialize Command to Vtctldclient

rohit-nayak-ps:
## Description

When completed, this PR will migrate all current vreplication vtctl-based commands to the new vtctldserver-based approach.

The following commands will be migrated as part of this PR, in order: 
Materialize, MoveTables, Reshard, CreateLookupVindex, VDiff,  Mount, Migrate

We will move commands one by one to vtctld and copy over all required wrangler related functionality to get each command workfing into the `vtctl\workflow` package. The vtctl command is then stubbed to call the vtctldserver instead. e2e tests should confirm that this works.

Note that since all the logic is being moved over, we will need to migrate the unit test framework and tests as well. This will be done once all commands have been ported over and the e2e tests pass.

We will also take this opportunity to refactor the code while the new unit tests are being written.

## Related Issue(s)
  - Related to: https://github.com/vitessio/vitess/issues/12152

## Checklist


mattlord:
@rohit-nayak-ps this will be a project, so IMO we should create a tracking issue for it. We can then repurpose this issue to be about Materialize. If you are good with that, I'll create a tracking issue — where each ticket will include everything for that command, including unit tests and any refactoring — and then start working on another command.

rohit-nayak-ps:
> @rohit-nayak-ps this will be a project, so IMO we should create a tracking issue for it. We can then repurpose this issue to be about Materialize. If you are good with that, I'll create a tracking issue — where each ticket will include everything for that command, including unit tests and any refactoring — and then start working on another command.

Yeah, we can either create a project or a parent task. We should probably only merge the sub-tasks into v17, so that we don't have some commands moved and other not, in v16.

Issue #11864, opened 2022-11-30
RFC: routing for unqualified reference tables


maxenglander:
# Introduction

The goal of this RFC is to make reference tables more useful, and to find a solution for one of its current limitations. This RFC is mainly motivated by #11863. However, depending on if/how we solve the problem, it may have a broader impact.

[Reference tables](https://vitess.io/docs/15.0/user-guides/vschema-guide/advanced-vschema/#reference-tables) are intended to be helpful in situations like:

 * You have an unsharded table, like `products`.
 * You have a sharded table, like `customers`.
 * You sometimes want to be able to join `customers` to `products`.
 * This query can be very inefficient because of the cross-keyspace join.

Reference tables are designed to solve this by letting you:

 * First materialize the unsharded table `products` table into each shard of the keyspace where `customers` lives.
 * Flag the materialized table as being of `type` `reference` in the `customers` VSchema.
 * From there, VTGate will route to a shard-local copy of `products` whenever it is joined to `customers`.

# Problem statement

There are limitations on how helpful VTGate will be with routing. 

Let's say you have want to have `products` table have the same name in both the unsharded and sharded keyspace. And then you want to issue keyspace-unqualified queries like this:

```
INSERT INTO `products` ...
SELECT * FROM `customers` JOIN `products` ON ...
```

These queries won't work. They will product this error: https://github.com/vitessio/vitess/blob/4ff02c0f7af35a520bfc90d32e77fc13810dfdd9/go/vt/vtgate/vindexes/vschema.go#L566

## Use cases

Just to show that this isn't a contrived problem, here is a use case that describes the situation of a customer of PlanetScale:

 1. They have a Java application using Spring and Hibernate.
 2. Their development and staging environments use an unsharded single keyspace.
 3. In production they have mix of unsharded keyspace and sharded keyspace.
 4. They want performance to be fast, so they materialize tables from the unsharded keyspace to reference tables in the sharded keyspace.
 5. They need writes to be sent to the unsharded keyspace, and joins in the sharded keyspace to use the reference tables.
 6. They have a limited ability to make their code aware of this setup; having separate Hibernate/JPA entities for unsharded source table and the sharded reference table, and re-writing their code to use these two entities, is prohibitively difficult for them.
 7. Even if it were an option for them, it would not be an attractive one because of the added complexity and because it could create separate code paths for production vs. other environments.

## Historical Support

It used to be possible to route queries to either the source table or reference table.

[...]

PR #11875, opened 2022-12-04
optimize joins, redirect dml for reference tables

maxenglander:
Addresses #11864.

# Description

Make some improvements to routing for [reference tables](https://vitess.io/docs/15.0/user-guides/vschema-guide/advanced-vschema/#reference-tables):
 * Support for [globally routing](https://vitess.io/docs/15.0/reference/features/global-routing) reference tables that are identically named with their source.
 * Optimize joining so that reference tables are joined with the optimal keyspace.
 * Redirect DML for reference tables back to their source.

Given:
    
* An unsharded keyspace `k1` and an sharded keyspace `k2`.
* Source table `k1.r` and reference table `k2.r`.
* Unsharded table `k1.a`
* Sharded table `k2.b`.

And these queries:

1. `SELECT k2.b JOIN k1.r`
2. `SELECT a JOIN r`
3. `SELECT b JOIN r`
4. `INSERT INTO r`
       
The first query will succeed but will issue queries to both `k1` and `k2`, even though the query _could_ be satisifed just by routing to `k2`.

The next three queries will fail with https://github.com/vitessio/vitess/blob/4ff02c0f7af35a520bfc90d32e77fc13810dfdd9/go/vt/vtgate/vindexes/vschema.go#L566

This PR adds limited support for optimally joining and globally routing these three queries:

1. The first query is routed exclusively to `k2`.
2 The second query will join `k1.a` to `k1.r`.
* The third query will join `k2.b` to `k2.r`.
* The fourth query will be routed to the unsharded source table `r`.

# Details

Some notes about the implementation:

 * Have intentionally put a lot of constraints (more on this in [docs](https://github.com/vitessio/website/pull/1270)) on this feature. For example requiring that a `source` is unique in a given keyspace. Future PRs could relax these constraints, but figured it's better to start with a smaller initially.
 * This feature will override qualified queries. E.g. `SELECT k2.a JOIN k1.r` will route `r` to `k2`, even though the user has explicitly requested `k1.r`.
 * Routing rules take precedence over this feature.

# Checklist

 * [x] Add support for `source` to table VSchema definition.
 * [x] Treat reference tables and their source as globally unambiguous.
 * [x] Route unqualified `INSERT` to reference tables back to the source.
 * [x] Join unqualified table to the keyspace-local reference.
 * [x] Write tests
 * [x] Add docs


go/vt/vtgate/planbuilder/operators/route_planning.go
```diff
@@ -497,6 +497,14 @@ func tryMerge(
 		return nil, nil
 	}
 
+	if altARoute := aRoute.AlternateInKeyspace(bRoute.Keyspace); altARoute != nil {
```
systay, line 500: I don't understand this. do we always want to do this on both routes? shouldn't we just do this if they are not on the same keyspace? Maybe add a comment explaining this logic?

go/vt/vtgate/vindexes/vschema.go
```diff
@@ -245,6 +248,116 @@ func buildKeyspaces(source *vschemapb.SrvVSchema, vschema *VSchema) {
...
+		ksvschema := vschema.Keyspaces[ksname]
+		buildGlobalTables(ks, vschema, ksvschema)
+	}
+}
+
+func buildGlobalTables(ks *vschemapb.Keyspace, vschema *VSchema, ksvschema *KeyspaceSchema) {
+	for tname, t := range ksvschema.Tables {
+		// If the keyspace requires explicit routing, don't include any of
+		// its tables in global tables.
+		if !ks.RequireExplicitRouting {
```
systay, line 271: I think go wants you to inverse this and continue:

```golang
if ks.RequireExplicitRouting {
  continue
}
```

systay:
Would you mind adding your three queries as plan_tests? Somewhere here: 

https://github.com/vitessio/vitess/tree/main/go/vt/vtgate/planbuilder/testdata

go/vt/vtgate/vindexes/vschema.go
```diff
@@ -232,7 +234,8 @@ func ValidateKeyspace(input *vschemapb.Keyspace) error {
 	return err
 }
 
-func buildKeyspaces(source *vschemapb.SrvVSchema, vschema *VSchema) {
+func buildKeyspaces(source *vschemapb.SrvVSchema, vschema *VSchema, allowUnresolvedReferences bool) {
+	// In the first loop, build tables.
```
systay, line 238: this method is looking nice ✨

go/test/endtoend/vtgate/reference/reference_test.go
```diff
@@ -0,0 +1,130 @@
...
+
+	"github.com/stretchr/testify/require"
+
+	"vitess.io/vitess/go/test/endtoend/utils"
+
+	"vitess.io/vitess/go/mysql"
+	"vitess.io/vitess/go/test/endtoend/cluster"
+)
+
+// TestGlobalReferenceRouting tests that unqualified queries for reference
```
systay, line 32: Nice with these comments

go/test/endtoend/vtgate/reference/reference_test.go
```diff
@@ -0,0 +1,130 @@
...
+// When: we execute `SELECT ... FROM zip JOIN zip_detail ...`,
+// Then: `zip_detail` should be routed to `uks`.
+//
+// When: we execute `SELECT ... FROM delivery_failure JOIN zip_detail ...`,
+// Then: `zip_detail` should be routed to `sks`.
+//
+// When: we execute `DELETE FROM zip_detail ...`,
+// Then: `zip_detail` should be routed to `uks`.
+func TestReferenceRouting(t *testing.T) {
+	defer cluster.PanicHandler(t)
```
systay, line 56: we have started using a new pattern for our end2end tests. You can see it in use [here](https://github.com/vitessio/vitess/blob/main/go/test/endtoend/vtgate/queries/aggregation/aggregation_test.go#L52).

The idea is to put the repetitive code we have to write out in any test method used. We set stuff up and return a destructor that knows how to clean out state, so tests don't accidentally step on each other.

go/test/endtoend/vtgate/reference/reference_test.go
```diff
@@ -0,0 +1,130 @@
...
+	conn, err := mysql.Connect(ctx, &vtParams)
+	require.Nil(t, err)
+	defer conn.Close()
+
+	// INSERT should route an unqualified zip_detail to unsharded keyspace.
+	utils.Exec(t, conn,
+		"INSERT INTO zip_detail(id, zip_id, discontinued_at) VALUES(3, 1, DATE('2022-12-03'))")
+	// Verify with qualified zip_detail queries to each keyspace. The unsharded
+	// keyspace should have an extra row.
+	uqr := utils.Exec(t, conn,
```
systay, line 71: instead of `utils.Exec` and then `fmt.Sprintf`, you should use `utils.AssertMatches`

systay (commented):
Really nice

go/vt/vtgate/vindexes/vschema.go
harshit-gangal: This should also move out of the for loop at the top of the method

go/vt/vtgate/vindexes/vschema.go
```diff
@@ -245,6 +248,116 @@ func buildKeyspaces(source *vschemapb.SrvVSchema, vschema *VSchema) {
...
+				// There is already an entry table stored in global tables
+				// with this name.
+				if gt == nil {
+					// Table name is already marked ambiguous, nothing to do.
+					continue
+				} else if _, ok := t.ReferencedBy[gt.Keyspace.Name]; ok {
+					// If the stored table refers to this table, store this
+					// table instead.
+					vschema.globalTables[tname] = t
+				} else if _, ok := gt.ReferencedBy[t.Keyspace.Name]; ok {
```
harshit-gangal, line 282: these can throw `map nil` error as they are not always initialized.
nil check would be needed here.

harshit-gangal:
As discussed offline. Few more things to be addressed

- Always tell VTGate to select optimal route, unless routing rules dictate otherwise. Later on can opt out with a query hint. Probably will do this with a second FindRoutedTable call in route planning code.
- Add txt-file based route planning tests, in a separate file from other tests.
- The reference routing does not have to rely on global routing.

go/vt/vtgate/planbuilder/operators/route.go
```diff
@@ -717,35 +722,72 @@ func (r *Route) planIsExpr(ctx *plancontext.PlanningContext, node *sqlparser.IsE
...
+	vschemaTable *vindexes.Table,
+	solves semantics.TableSet,
+	planAlternates bool,
+) (*Route, error) {
+	if vschemaTable.Name.String() != queryTable.Table.Name.String() {
 		// we are dealing with a routed table
-		name := table.Table.Name
-		table.Table.Name = vschemaTable.Name
-		astTable, ok := table.Alias.Expr.(sqlparser.TableName)
+		queryTable = queryTable.Clone()
```
maxenglander, line 774: Writing txt-based plan tests helped me catch this bug 💪 

The bug:

 1. `createRouteFromVSchemaTable` passes `queryTable to `createAlternateRoutesFromVSchemaTable`,
 2. `createAlternateRoutesFromVSchemaTable` passes `queryTable` to `findVSchemaTableAndCreateRoute`,
 3. `findVSchemaTableAndCreateRoute` passes `queryTable` to `createRouteFromVSchemaTable`,
 4. `createRouteFromVSchemaTable` modifies `queryTable`, which impacts step 1!

go/vt/vtgate/planbuilder/testdata/dml_cases.json
```diff
@@ -6199,5 +6199,28 @@
         "user.user"
       ]
     }
+  },
+  {
+    "comment": "insert into ref; TODO(maxeng) is this a bug?",
```
maxenglander, line 6204: Tested this on `main` branch and get the same results. Is this a bug? When we do similar `UPDATE` or `DELETE` we get a `primary vindex not found` error, but this `INSERT` succeeds.

go/vt/vtgate/planbuilder/testdata/reference_cases.json
```diff
@@ -0,0 +1,644 @@
...
+        },
+        "OperatorType": "Route",
+        "Query": "select ambiguous_ref_with_source.col from ambiguous_ref_with_source, `user`",
+        "Table": "`user`, ambiguous_ref_with_source",
+        "Variant": "Scatter"
+      },
+      "Original": "select ambiguous_ref_with_source.col from ambiguous_ref_with_source join user",
+      "QueryType": "SELECT",
+      "TablesUsed": [
+        "main.ambiguous_ref_with_source",
```
maxenglander, line 185: The instruction seems right, but not sure this is?

go/vt/vtgate/planbuilder/testdata/reference_cases.json
```diff
@@ -0,0 +1,644 @@
...
+        },
+        "OperatorType": "Route",
+        "Query": "select `user`.col from `user`, ambiguous_ref_with_source",
+        "Table": "`user`, ambiguous_ref_with_source",
+        "Variant": "Scatter"
+      },
+      "Original": "select user.col from user join ambiguous_ref_with_source",
+      "QueryType": "SELECT",
+      "TablesUsed": [
+        "main.ambiguous_ref_with_source",
```
maxenglander, line 91: The instruction seems right, but not sure this is?

go/vt/vtgate/planbuilder/testdata/reference_cases.json
```diff
@@ -0,0 +1,644 @@
...
+        "Values": [
+          "INT64(1)"
+        ],
+        "Variant": "EqualUnique",
+        "Vindex": "user_index"
+      },
+      "Original": "select ambiguous_ref_with_source.col from ambiguous_ref_with_source join (select aa from user where user.id=1) user",
+      "QueryType": "SELECT",
+      "TablesUsed": [
+        "main.ambiguous_ref_with_source",
```
maxenglander, line 250: Instruction seems right but not sure this is?

go/vt/vtgate/planbuilder/testdata/reference_cases.json
```diff
@@ -0,0 +1,644 @@
...
+        },
+        "OperatorType": "Route",
+        "Query": "select `user`.col from `user`, ambiguous_ref_with_source",
+        "Table": "`user`, ambiguous_ref_with_source",
+        "Variant": "Scatter"
+      },
+      "Original": "select user.col from user join main.ambiguous_ref_with_source",
+      "QueryType": "SELECT",
+      "TablesUsed": [
+        "main.ambiguous_ref_with_source",
```
maxenglander, line 307: The instruction seems right, but not sure this is?

go/vt/vtgate/planbuilder/testdata/reference_cases.json
```diff
@@ -0,0 +1,644 @@
...
+        "TargetTabletType": "PRIMARY",
+        "Variant": "Sharded"
+      },
+      "Original": "insert into user.ambiguous_ref_with_source(col) values(1)",
+      "QueryType": "INSERT",
+      "TablesUsed": [
+        "user.ambiguous_ref_with_source"
+      ]
+    },
+    "gen4-plan": {
```
maxenglander, line 376: I was expecting an error here because `UPDATE` and `DELETE` give errors. Added a similar (passing) test to `dml_cases.json` and tested it on `main`.

go/vt/vtgate/planbuilder/testdata/reference_cases.json
```diff
@@ -0,0 +1,644 @@
...
+        },
+        "OperatorType": "Route",
+        "Query": "select `user`.col from `user`, ref as rerouted_ref",
+        "Table": "`user`, ref",
+        "Variant": "Scatter"
+      },
+      "Original": "select user.col from user join rerouted_ref",
+      "QueryType": "SELECT",
+      "TablesUsed": [
+        "main.rerouted_ref",
```
maxenglander, line 639: The instruction seems right, but not sure this is?

maxenglander:
Hey @systay and @harshit-gangal I think I addressed all of your feedback. One thing I'm not sure if we've explicitly discussed is what the behavior should be when the source and reference tables have different names:

_vschema_
```
{
  "keyspaces": [
    "k1": {
      "sharded": false,
      "tables": {
         "user": {}
         "ref_source": {
            "type": "reference",
            "source": "k2.ref_target"
          }
        }
    },
    "k2": {
      "sharded": false,
      "tables": {
         "ref_source": {
            "type": "reference"
          }
       }
    }
  ]
}
```

What should happen with this query:

```sql
SELECT * FROM user JOIN ref_source;
```

The behavior prior to and with this PR is that it joins `ref_source`, and, even though it has the information needed to join `ref_target`, it doesn't.

In writing this PR I wanted to keep the scope of changes small, so I limited the new behavior to tables that have the same name. Let me know if you want me to expand it to include differently-named tables.

go/vt/vtgate/vindexes/vschema.go
```diff
@@ -245,6 +248,116 @@ func buildKeyspaces(source *vschemapb.SrvVSchema, vschema *VSchema) {
...
+
+		_, err := escapeQualifiedTable(table.Source)
+		if err != nil {
+			return fmt.Errorf("invalid source %q for reference table: %s; %v", table.Source, tname, err)
+		}
+
+		sourceParts := strings.Split(table.Source, ".")
+		sourceKsname := sourceParts[0]
+		if sourceKsname == keyspace.Name {
+			return fmt.Errorf("source %q may not reference a table in the same keyspace as table: %s", table.Source, tname)
```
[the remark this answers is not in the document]
systay: I noticed this, but the rest of the file is still using `fmt.Errorf`
🤷

go/vt/vtgate/planbuilder/operators/route.go
systay: Mutable state is the source of sooo many bugs. I miss Scala

go/vt/vtgate/planbuilder/testdata/reference_cases.json
systay: looks correct to me - these are the two tables you are querying, right?

systay (approved):
LGTM

go/test/endtoend/vtgate/reference/main_test.go
```diff
@@ -0,0 +1,271 @@
...
+			return 1
+		}
+
+		sKeyspace := &cluster.Keyspace{
+			Name:      shardedKeyspaceName,
+			SchemaSQL: shardedSQLSchema,
+			VSchema:   shardedVSchema,
+		}
+		if err := clusterInstance.StartKeyspace(*sKeyspace, []string{"-80", "80-"}, 1, false); err != nil {
+			return 1
```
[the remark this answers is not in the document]
harshit-gangal: same for `StartUnshardedKeyspace` if materialize happens from Primary

go/vt/vtgate/vindexes/vschema.go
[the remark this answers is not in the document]
maxenglander: In the latest commit I changed everything in the to use `vterrors`. In some cases I wasn't confident about the error code to use, might be good to double check that.

go/test/endtoend/vtgate/reference/main_test.go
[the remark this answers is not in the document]
maxenglander: Sorry, fixed

go/vt/vtgate/planbuilder/testdata/dml_cases.json
harshit-gangal: issue for a different PR

maxenglander: 👍 want me to revert this test case or leave in place in this PR?

go/vt/vtgate/vindexes/vschema.go
```diff
@@ -216,11 +237,12 @@ func BuildKeyspaceSchema(input *vschemapb.Keyspace, keyspace string) (*KeyspaceS
 		},
 	}
 	vschema := &VSchema{
-		uniqueTables:   make(map[string]*Table),
+		globalTables:   make(map[string]*Table),
 		uniqueVindexes: make(map[string]Vindex),
 		Keyspaces:      make(map[string]*KeyspaceSchema),
 	}
 	buildKeyspaces(formal, vschema)
+	buildReferences(formal, vschema, true /*allowUnresolvedReferences*/)
```
harshit-gangal, line 245: I think this is not needed on `BuildKeyspaceSchema` as it is used for validation for a schema provided for a keyspace and here we are anyways ignoring unresolved references.

go/vt/vtgate/planbuilder/testdata/reference_cases.json
harshit-gangal: The Route looks good, but the TablesUsed should just have `ambiguous_ref_with_source`.
`main.ambiguous_ref_with_source` does not look correct.

go/vt/vtgate/planbuilder/testdata/reference_cases.json
harshit-gangal: same. should not refer to `main` keyspace.

go/vt/vtgate/planbuilder/testdata/reference_cases.json
harshit-gangal: true, I checked the execution will fail for that insert plan, ideally should fail at planning time.

go/vt/vtgate/planbuilder/testdata/reference_cases.json
harshit-gangal: this needs to be looked into

harshit-gangal:
join queries are one part, probably we should also utilize this logic in `subquery` merge as well.

harshit-gangal:
reference e2e test is failing, Looks like some issue with the Materialize command

```
2022-12-06T15:57:19.1001622Z I1206 15:44:25.940548   37785 vtctl_process.go:103] Executing vtctlclient with arguments vtctl --topo_implementation etcd2 --topo_global_server_address localhost:17602 --topo_global_root /vitess/global Materialize {
2022-12-06T15:57:19.1001798Z 				"workflow": "materialize_zip_detail",
2022-12-06T15:57:19.1001937Z 				"source_keyspace": "uks",
2022-12-06T15:57:19.1002071Z 				"target_keyspace": "sks",
2022-12-06T15:57:19.1002193Z 				"table_settings": [
2022-12-06T15:57:19.1002265Z 					{
2022-12-06T15:57:19.1002427Z 						"target_table": "zip_detail",
2022-12-06T15:57:19.1002648Z 						"source_expression": "select * from zip_detail"
2022-12-06T15:57:19.1002731Z 					}
2022-12-06T15:57:19.1002809Z 				],
2022-12-06T15:57:19.1002940Z 				"stop_after_copy": true
2022-12-06T15:57:19.1003010Z 			}
2022-12-06T15:57:19.1003086Z SIGQUIT: quit
2022-12-06T15:57:19.1003187Z PC=0x40472c m=0 sigcode=0
```

go/vt/vtgate/planbuilder/testdata/reference_cases.json
```diff
@@ -0,0 +1,644 @@
...
+      "TablesUsed": [
+        "main.ambiguous_ref_with_source"
+      ]
+    }
+  },
+  {
+    "comment": "delete from qualified ambiguous reference table error no primary vindex",
+    "query": "delete from user.ambiguous_ref_with_source where col = 1",
+    "v3-plan": "table 'ambiguous_ref_with_source' does not have a primary vindex",
+    "gen4-plan": "table 'ambiguous_ref_with_source' does not have a primary vindex"
```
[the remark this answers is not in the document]
maxenglander: If I'm understanding right, you want the reference logic to redirect `INSERT [qualified].[reference]` back to the source?

go/test/endtoend/vtgate/queries/reference/main_test.go
```diff
@@ -0,0 +1,292 @@
...
+			INSERT INTO delivery_failure(id, zip_detail_id, reason)
+			VALUES (1, 1, 'Failed delivery due to discontinued zipcode.'),
+			       (2, 2, 'Failed delivery due to discontinued zipcode.'),
+			       (3, 3, 'Failed delivery due to unknown reason.');
+		`, map[string]*querypb.BindVariable{}); err != nil {
+			return 1
+		}
+
+		// TODO(maxeng) remove when we have a proper way to check
+		// materialization lag and cutover.
```
mattlord, line 211: Just FYI, if it's helpful:
  - https://github.com/vitessio/vitess/blob/main/go/test/endtoend/vreplication/helper_test.go#L154
  - https://github.com/vitessio/vitess/blob/main/go/test/endtoend/vreplication/helper_test.go#L245
  - https://github.com/vitessio/vitess/blob/main/go/test/endtoend/vreplication/helper_test.go#L210

go/vt/vtgate/planbuilder/testdata/reference_cases.json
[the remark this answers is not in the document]
harshit-gangal: yes, all the DML queries should be directed to the source.

go/vt/vtgate/planbuilder/testdata/reference_cases.json
systay: I think you are wrong, @harshit-gangal. We always add the keyspace prefix to table names in the TablesUsed field. We should never have a bare table name there

go/vt/vtgate/planbuilder/testdata/reference_cases.json
systay: same as above. it most definitely should say which ks that table lives on

go/vt/vtgate/planbuilder/testdata/reference_cases.json
harshit-gangal: I mean it should say it is going to `user`.`ambiguous_ref_with_source` and not `main`.`ambiguous_ref_with_source`

maxenglander:
Hey @harshit-gangal I didn't address all of your recent feedback, but I addressed some of it. I wanted to share some thoughts on the latest commit.

In your last feedback you noted that some of the txt plan tests were not right. For example: 

image

The cause of this issue is that:
 1. `semantics.Analyzer` and `semantics.TableCollector` are responsible for collecting table names in the current planner. They use the VSchema to lookup a VTable with a query table name.
 2. With the new logic, VSchema is no longer authoritative on on what VTable we end up routing to. With the new logic, routing can sometimes select an alternate VTable. `semantics` package doesn't know about that logic.
 3. So we end up in a situation where the `Instruction` is correct, but `semantics` is not in sync with this view.

I can think of a couple ways to resolve this:
 * Mirror the merge-routing logic from `planbuilder` over to `semantics`, so they stay in sync. Or find a way to extract the logic to a common package, in a way that is generic enough to work with both packages.
 * Make VSchema smart enough to be able to select the alternate VTable, e.g. with `FindTableOrVindexWithHints(&Hint{Keyspace, IsDML})`.
 * Instead of having the `semantics` package collect table names, let the `engine` package do this.

I ended up going with the 3rd option in 80f9c9439d. It required touching a lot of files, but (to me) the logic is straightforward. Here's how it works:
 1. When creating primitives, `planbuilder` collects table names from QTables and assigns them to the primitive.
 2. Primitive returns those tables via `GetTablesUsed`, qualifying the table names with the Primitive's keyspace.
 3. For complex Primitives (e.g. joins), `GetTablesUsed` queries its own `Inputs().GetTablesUsed()`.

go/vt/vtgate/engine/helpers.go
```diff
@@ -0,0 +1,61 @@
...
+	"fmt"
+	"sort"
+
+	"vitess.io/vitess/go/vt/sqlparser"
+	"vitess.io/vitess/go/vt/vtgate/vindexes"
+)
+
+var noUsedTables = []string{}
+
+func collectSortedUniqueStrings() (add func(string), collect func() []string) {
```
maxenglander, line 13: TODO these might not be efficient

go/vt/vtgate/engine/ordered_aggregate.go
```diff
@@ -221,6 +221,15 @@ func (oa *OrderedAggregate) GetTableName() string {
 	return oa.Input.GetTableName()
 }
 
+// GetTablesUsed specifies the table that this primitive routes to.
```
maxenglander, line 224: TODO fix comment

go/vt/vtgate/engine/pullout_subquery.go
```diff
@@ -62,6 +62,15 @@ func (ps *PulloutSubquery) GetTableName() string {
 	return ps.Underlying.GetTableName()
 }
 
+// GetTablesUsed specifies the table that this primitive routes to.
```
maxenglander, line 65: TODO fix comment

go/vt/vtgate/engine/simple_projection.go
```diff
@@ -53,6 +53,15 @@ func (sc *SimpleProjection) GetTableName() string {
 	return sc.Input.GetTableName()
 }
 
+// GetTablesUsed specifies the table that this primitive routes to.
```
maxenglander, line 56: TODO fix comment

go/vt/vtgate/engine/singlerow.go
```diff
@@ -46,6 +46,11 @@ func (s *SingleRow) GetTableName() string {
 	return ""
 }
 
+// GetTablesUsed specifies the table that this primitive routes to.
```
maxenglander, line 49: TODO fix comment

go/vt/vtgate/engine/vindex_func.go
```diff
@@ -88,6 +90,11 @@ func (vf *VindexFunc) GetTableName() string {
 	return ""
 }
 
+// GetTablesUsed specifies the table that this primitive routes to.
```
maxenglander, line 93: TODO fix comment

go/vt/vtgate/engine/dml.go
```diff
@@ -120,6 +122,22 @@ func (dml *DML) GetTableName() string {
 	return ""
 }
 
+// GetTablesUsed specifies the table that this primitive routes to.
```
maxenglander, line 125: TODO fix comment

go/vt/vtgate/engine/vindex_lookup.go
```diff
@@ -69,6 +69,16 @@ func (vr *VindexLookup) GetTableName() string {
 	return vr.SendTo.GetTableName()
 }
 
+// GetTablesUsed implements the Primitive interface
+func (vr *VindexLookup) GetTablesUsed() []string {
+	//add, collect := concatSortedUniqueStringSlices()
```
maxenglander, line 74: TODO delete commented out code

go/vt/vtgate/planbuilder/operators/route.go
```diff
@@ -615,6 +622,7 @@ func (r *Route) resetRoutingSelections(ctx *plancontext.PlanningContext) error {
 			return err
 		}
 	}
+
```
maxenglander, line 625: TODO undo

go/vt/vtgate/planbuilder/testdata/select_cases.json
```diff
@@ -3542,7 +3542,7 @@
         "Table": "`user`, dual"
       },
       "TablesUsed": [
-        "main.dual",
+        "user.dual",
```
maxenglander, line 3545: This change seems OK to me, potentially an improvement. If you don't agree I can make code changes to preserve the existing plan.

go/vt/vtgate/planbuilder/testdata/select_cases.json
```diff
@@ -5170,7 +5170,7 @@
         "Vindex": "user_index"
       },
       "TablesUsed": [
-        "main.dual",
+        "user.dual",
```
maxenglander, line 5173: Ditto

go/vt/vtgate/planbuilder/testdata/reference_cases.json
maxenglander: Note to sell, still TODO: route gen4 ref inserts back to source.

go/vt/vtgate/planbuilder/testdata/reference_cases.json
```diff
@@ -0,0 +1,680 @@
...
+      "Original": "select user.col from user join ref_with_source",
+      "QueryType": "SELECT",
+      "TablesUsed": [
+        "user.ref_with_source",
+        "user.user"
+      ]
+    }
+  },
+  {
+    "comment": "join with unqualified reference does not optimize routes if source & reference different names; TODO(maxeng) not working",
```
maxenglander, line 567: My intention with the PR was to only allow optimal reference routing to happen when both the reference and source have the same table name. It looks like that constraint isn't in place.

TODO: put constraint in place, unless a PR reviewer says they prefer the current behavior.

go/vt/vtgate/planbuilder/testdata/reference_cases.json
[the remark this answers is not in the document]
maxenglander: OK this is working right for UPDATE and DELETE now. INSERT still TODO.

go/vt/vtgate/planbuilder/testdata/reference_cases.json
```diff
@@ -0,0 +1,644 @@
...
+      "TablesUsed": [
+        "main.ambiguous_ref_with_source"
+      ]
+    }
+  },
+  {
+    "comment": "update qualified ambiguous reference table error no primary vindex",
+    "query": "update user.ambiguous_ref_with_source set col = 1",
+    "v3-plan": "table 'ambiguous_ref_with_source' does not have a primary vindex",
+    "gen4-plan": "table 'ambiguous_ref_with_source' does not have a primary vindex"
```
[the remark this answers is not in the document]
maxenglander: Working now. TODO update test case comment to reflect.

go/vt/vtgate/planbuilder/testdata/reference_cases.json
harshit-gangal: The name should have nothing to do with optimal routing, it should just be determined if any reference table is present in the current keyspace for the source table or not.

I like this behaviour, I would have pushed for this if any constraint would have placed :)

harshit-gangal: I would say you can change the test comment.

go/vt/vtgate/planbuilder/testdata/select_cases_with_default.json
```diff
@@ -61,9 +61,9 @@
         "Vindex": "user_index"
       },
       "TablesUsed": [
-        "second_user.dual",
+        "user.dual",
```
maxenglander, line 64: This seems OK to me, maybe even an improvement. LMK if we prefer the current plan over this.

go/vt/vtgate/engine/helpers.go
```diff
@@ -0,0 +1,61 @@
+package engine
```
systay, line 1: license missing

maxenglander: Moved contents to `go/vt/vtgate/planbuilder/operators/helpers.go` for now, which already has a license.

go/vt/vtgate/planbuilder/operator_transformers.go
```diff
@@ -153,13 +153,7 @@ func transformApplyJoinPlan(ctx *plancontext.PlanningContext, n *operators.Apply
...
 }
 
-func transformRoutePlan(ctx *plancontext.PlanningContext, op *operators.Route) (logicalPlan, error) {
-	switch src := op.Source.(type) {
-	case *operators.Update:
-		return transformUpdatePlan(ctx, op, src)
-	case *operators.Delete:
-		return transformDeletePlan(ctx, op, src)
-	}
+func routeToEngineRoute(ctx *plancontext.PlanningContext, op *operators.Route) (*engine.Route, error) {
```
maxenglander, line 156: Don't need this separate fn anymore after moving `TablesUsed()` code from `engine` to `planbuilder`, but I don't think it hurts. Can revert just LMK.

go/vt/vtgate/planbuilder/testdata/reference_cases.json
maxenglander: Done in 8c8da64fe1

go/vt/vtgate/planbuilder/testdata/reference_cases.json
```diff
@@ -0,0 +1,680 @@
...
+      },
+      "Original": "delete from ambiguous_ref_with_source where col = 1",
+      "QueryType": "DELETE",
+      "TablesUsed": [
+        "main.ambiguous_ref_with_source"
+      ]
+    }
+  },
+  {
+    "comment": "delete from qualified ambiguous reference table error no primary vindex",
```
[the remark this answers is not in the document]
maxenglander: Done in 8c8da64fe1

go/vt/vtgate/planbuilder/operators/route.go
maxenglander: Done 8c8da64fe1

maxenglander:
> join queries are one part, probably we should also utilize this logic in `subquery` merge as well.

@harshit-gangal is it OK if I make that a separate PR task? This PR is getting a bit big.

go/vt/vtgate/planbuilder/testdata/reference_cases.json
maxenglander: Done

go/vt/vtgate/planbuilder/testdata/reference_cases.json
maxenglander: Done

go/vt/vtgate/planbuilder/testdata/reference_cases.json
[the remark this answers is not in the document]
maxenglander: Done

go/vt/vtgate/planbuilder/testdata/reference_cases.json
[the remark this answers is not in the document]
maxenglander: Done

go/vt/vtgate/planbuilder/operators/table.go
```diff
@@ -79,6 +79,13 @@ func (to *Table) AddCol(col *sqlparser.ColName) {
 	to.Columns = append(to.Columns, col)
 }
 
+func (to *Table) TablesUsed() []string {
+	if sqlparser.SystemSchema(to.QTable.Table.Qualifier.String()) {
```
maxenglander, line 83: I'm just preserving logic that was already in the code, but am not sure why we do this?

go/vt/vtgate/planbuilder/single_sharded_shortcut.go
```diff
@@ -44,47 +45,60 @@ func unshardedShortcut(ctx *plancontext.PlanningContext, stmt sqlparser.SelectSt
...
 	for _, tableInfo := range semTable.Tables {
 		tblObj := tableInfo.GetVindexTable()
 		if tblObj == nil {
 			// probably a derived table
 			continue
 		}
-		var name string
 		if tableInfo.IsInfSchema() {
-			name = "tableName"
+			tableNameMap["tableName"] = sqlparser.TableName{
```
maxenglander, line 85: Preserving logic that's already in the codebase, but don't understand what this is about.

go/vt/vtgate/planbuilder/testdata/select_cases.json
```diff
@@ -5170,7 +5170,7 @@
         "Vindex": "user_index"
       },
       "TablesUsed": [
-        "user.dual",
+        "main.dual",
         "user.user"
       ]
```
harshit-gangal, line 5175: this change looks incorrect to me

harshit-gangal: this is giving wrong information, though the plan is correct.

harshit-gangal (commented):
LGTM overall.

go/vt/vtgate/planbuilder/testdata/dml_cases.json
harshit-gangal: Leaving as-is is good

go/vt/vtgate/planbuilder/from.go
```diff
@@ -310,6 +319,26 @@ func (pb *primitiveBuilder) processJoin(ajoin *sqlparser.JoinTableExpr, reserved
...
+// If the primitiveBuilder context is a Gen4 planner, the statement is an
+// INSERT, and the vschema table is a reference with a valid source reference,
+// then redirect the INSERT back to the source.
+func (pb *primitiveBuilder) tryRedirectGen4InsertToSource(vschemaTable *vindexes.Table) (*vindexes.Table, error) {
+	if pb.stmt == nil {
+		return nil, nil
+	}
+	if _, ok := pb.stmt.(*sqlparser.Insert); !ok {
+		return nil, nil
+	}
```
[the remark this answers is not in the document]
maxenglander: This function is called from a general primitive builder context, which in V3 is used by SELECT/UPDATE/DELETE in addition to INSERT. I want to filter out those cases (handled separately in Gen4 planning), and only handle INSERT here.

go/vt/vtgate/planbuilder/testdata/select_cases.json
maxenglander: Fixed, no longer any diff.

go/vt/vtgate/engine/vindex_func.go
```diff
@@ -43,8 +44,9 @@ type VindexFunc struct {
 	// Cols contains source column numbers: 0 for id, 1 for keyspace_id.
 	Cols []int
 	// TODO(sougou): add support for MultiColumn.
-	Vindex vindexes.SingleColumn
-	Value  evalengine.Expr
+	Vindex    vindexes.SingleColumn
+	Value     evalengine.Expr
+	TableName sqlparser.TableName
```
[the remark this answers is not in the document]
maxenglander: Fixed

systay (approved):

harshit-gangal (approved):

maxenglander:
@harshit-gangal @systay thanks so much for the great reviews and for the approvals. Please hold off on merging until I do a last round of validation, thanks!

maxenglander:
OK tested  ready to merge from my POV :)

systay:
🎉

PR #11876, opened 2022-12-04
support virtual global keyspace names

maxenglander:
# Description
    
Vitess currently supports globally routing queries to unique tables when the connection does not have a default keyspace and the tables in the `FROM` list are not keyspace-qualified.
    
This PR adds support for explicitly requesting global routing:
    
* Adds a VTGate flag `--global-keyspace` that can be specified multiple times.
* When a connection uses a `--global-keyspace`, it behaves the same was as if no default keyspace is set on the connection.
* When a query for a table is qualified with a `--global-keyspace`, it is routed to a globally unique table, even if it is in a different keyspace than the connection default keyspace.

# Details

Some details about the implementation:

 * Adds a `global` object to `*vindexes.VSchema`, and moves `uniqueTables` and `uniqueVindexes` into that object.
 * If an existing keyspace shares a name with a `--global-keyspace`, the existing keyspace will be used with non-global routing. This conflict resolution strategy prefers preserving existing behavior.
 * When a connection is set to use a `--global-keyspace`, `SELECT database()` will return that keyspace name. Maybe this isn't what we want?
 * Queries like `SHOW TABLES IN [global_ks]` do not currently work.

# Use cases
    
This feature is useful for frameworks like Hibernate which are required by to set a database name on the connection. Those frameworks are unable to take advantage of global routing.

# Checklist

 * [x] Add support for `--global-keyspace`.
 * [x] Write tests.
 * [ ] Write docs.


maxenglander:
Closed in favor of https://github.com/vitessio/vitess/pull/11938 because Vitess team prefers non-Vitess-team PRs come from forks.

PR #11881, opened 2022-12-05
Timeout Fixes and VTOrc Improvement

### LINKED ISSUE #11662: Bug Report: VTOrc repairs fail when vttablet is unreachable
### Overview of the Issue

It has been noticed that if the vttablet is unreachable then, VTOrc fails to repair the issue with the default values set for `LockShardTimeoutSeconds` and `TopoRemoteOperationTimeout`.

Both of these values have a default for 30 seconds. The `RemoteOperationTimeout` is used to wait for a tablet's response when we want to refresh the information for a shard in VTOrc. Since the `LockShardTimeout` is also 30 seconds, we end up spending all the time in the refresh phase itself before we even reach the repair. 

This leads to VTOrc timing out again and again and failing to run the repairs.

This problem is generally less likely to happen in an environment deployed by an operator because the vttablet pods that are unreachable are readily restarted by the operator. Most of the times there we run into issues with the vttablet being not serving or MySQL not being in the correct state. These situations VTOrc handles already.

This same problem is also present in the ERS codepath, wherein we use the TopoRemoteOperationTimeout for waiting for DemotePrimary to return. In case of multiple failures due to vttablets not being reachable, we won't be able to handle them.

### Reproduction Steps

Have a cluster running with VTOrc. Shutdown a vttablet instance and wait for VTOrc to repair. Check logs to see the failures in repair that VTOrc runs into.

### Binary Version

```sh
All
```

```sh
All
```

### Log Fragments

_No response_

GuptaManan100:
## Description

This PR implements the ideas discussed in https://github.com/vitessio/vitess/issues/11662#issuecomment-1328948688  and https://github.com/vitessio/vitess/issues/11662#issuecomment-1337087050.

As discussed, this PR makes the following changes - 
1. Remove the `LockShardTimeout` from VTOrc as it doesn't provide any value.
2. Add parallelism to VTOrc refresh phase so that we wait at most `RemoteOperationTimeout` seconds before proceeding.
3. Introduce a new flag `lock-timeout` which defaults to 45 seconds instead of using `RemoteOperationTimeout` for the shard lock duration. The default for `RemoteOperationTimeout` has also been changed to 15 seconds. 
4. Fix the usage of `WaitReplicasTimeout` in ERS code-path. This makes it so that we are able to handle multiple failures with the set default values.

## Related Issue(s)

- Fixes #11662 

## Checklist


GuptaManan100:
Even with these changes, VTOrc is unable to fix the failure of a single `vttablet`. i.e. if the primary's `vttablet` crashes, VTOrc doesn't run an ERS.
The issues still pending are - 
1. vttablets in their `tm.Close()` function, delete the MySQL hostname and some more information from the vttablet record. This causes VTOrc to think that the vttablet isn't part of the topology anymore. This assumption is however correct. Until the tablet's record is removed from the topo-server, it should be considered part of the topology.
2. The above only happens sometimes on the vttablets. The reason is that the `onclose_timeout` defaults to 1ns, so the `tm.Close()` function runs successfully only sometimes. This makes the VTOrc behavior unpredictable.
3. When a vttablet fails, VTOrc thinks it is unreachable, but doesn't think it's dead. The reason is that for MySQL failures, VTOrc tries to take in information from the replicas as well to see if they are able to reach the primary MySQL server. If the replication is working fine for the replicas, then VTOrc assumes that it is network partitioned from the MySQL server and doesn't run an ERS.
For vttablet failures, this however becomes a problem! Since vttablets don't contact each other, if the MySQL servers are working fine, VTOrc doesn't run an ERS. We would need some way to know if the vttablet is actually down, or just network partitioned from VTOrc. We could potentially use the vtgate for it 🤷‍♂️, but this is just an idea. It needs more thought on how we want to handle this.


Irrespective of the problems listed above, this PR is still an improvement for both VTOrc and ERS, since it does allow ERS to handle multiple failures with the default values and it allows VTOrc to handle the situations where both vttablet and MySQL server is down. A test for this situation has been added to verify that this indeed does work after the changes.

GuptaManan100:
I have now fixed the default timeout of `onclose_timeout` flag. Do we need to include this in the release notes summary too?
I have also patched VTOrc not to forget vttablet instances with an empty hostname as long as they are present in the topo server. A test has been added for the same.

deepthi:
> I have now fixed the default timeout of `onclose_timeout` flag. Do we need to include this in the release notes summary too?

I doubt anyone is actually using it, so this should be fine without a release note.

doc/releasenotes/16_0_0_summary.md
```diff
@@ -40,6 +40,14 @@ In [PR #11097](https://github.com/vitessio/vitess/pull/11097) we introduced nati
 Orchestrator integration in `vttablet` was deprecated in the previous release and is deleted in this release.
 Consider using `VTOrc` instead of `Orchestrator`.
 
+#### `lock-timeout` and `remote_operation_timeout` Changes
+Earlier, the shard and keyspace locks used to be capped by the `remote_operation_timeout`. This is no longer the case and instead a new flag called `lock-timeout` is introduced. 
+For backward compatibility, if `lock-timeout` is unspecified and `remote_operation_timeout` flag is provided, then its value will also be used for `lock-timeout` as well.
+The default value for `remote_operation_timeout` has also changed from 30 seconds to 15 seconds. The default for the new flag `lock-timeout` is 45 seconds.
+
+During upgrades, if the users want to preserve the same behaviour as previous releases, then they should provide the `remote_operation_timeout` flag explicitly before upgrading.
+After the upgrade, they should then alter their configuration to also specify `lock-timeout` explicitly.
```
deepthi, line 49: Specifying both remote_operation_timeout and lock-timeout to 30s will expose users to the bug that is being fixed in this PR.
Instead we should recommend that if (and only if) they get timeouts during PRS or ERS, they may need to increase remote_operation_timeout and lock-timeout at the same time while making sure that lock-timeout is at least 15 seconds more than remote_operation_timeout.

go/flags/endtoend/vtctld.txt
```diff
@@ -1,5 +1,5 @@
 Usage of vtctld:
-      --action_timeout duration                                          time to wait for an action before resorting to force (default 2m0s)
+      --action_timeout duration                                          time to wait for an action before resorting to force (default 1m0s)
```
deepthi, line 2: Why is this changing? Can you also check the blame so that we can see whether it changed recently?
There's an action_timeout that defaults to 1h on the vtctlclient and vtctldclient binaries. How does this relate to those?

go/flags/endtoend/vtorc.txt
```diff
@@ -23,14 +23,14 @@ Usage of vtorc:
       --keep_logs duration                           keep logs for this long (using ctime) (zero to keep forever)
       --keep_logs_by_mtime duration                  keep logs for this long (using mtime) (zero to keep forever)
       --lameduck-period duration                     keep running at least this long after SIGTERM before stopping (default 50ms)
-      --lock-shard-timeout duration                  Duration for which a shard lock is held when running a recovery (default 30s)
```
deepthi, line 26: do we need to deprecate this flag?

go/internal/flag/flag.go
```diff
@@ -70,6 +70,17 @@ func Parse(fs *flag.FlagSet) {
 	flag.Parse()
 }
 
+// IsFlagPassed returns if the given flag has been provided by the user explicitly or not
```
deepthi, line 73: suggestion: IsFlagProvided
WDYT? also @ajm188

go/test/endtoend/reparent/plannedreparent/reparent_test.go
```diff
@@ -376,3 +381,142 @@ func TestReparentDoesntHangIfPrimaryFails(t *testing.T) {
 	require.Error(t, err)
 	assert.Contains(t, out, "primary failed to PopulateReparentJournal")
 }
+
+// TestCrossCellDurability tests 2 things -
+// 1. When PRS is run with the cross_cell durability policy setup, then the semi-sync settings on all the tablets are as expected
+// 2. Bringing up a new vttablet should have its replication and semi-sync setup correctly without any external interference
```
deepthi, line 387:
```suggestion
// 2. Bringing up a new vttablet should have its replication and semi-sync setup correctly without manual intervention
```

go/vt/topo/locks.go
```diff
@@ -265,10 +266,8 @@ func (l *Lock) unlockKeyspace(ctx context.Context, ts *Server, keyspace string,
 	// Detach from the parent timeout, but copy the trace span.
 	// We need to still release the lock even if the parent
 	// context timed out.
-	// Note that we are not using the user provided RemoteOperationTimeout
-	// here because it is possible that that timeout is too short.
 	ctx = trace.CopySpan(context.TODO(), ctx)
-	ctx, cancel := context.WithTimeout(ctx, defaultLockTimeout)
+	ctx, cancel := context.WithTimeout(ctx, RemoteOperationTimeout)
```
deepthi, line 270: 👍

go/vt/topo/locks.go
```diff
@@ -456,10 +455,8 @@ func (l *Lock) internalLockShard(ctx context.Context, ts *Server, keyspace, shar
 func (l *Lock) unlockShard(ctx context.Context, ts *Server, keyspace, shard string, lockDescriptor LockDescriptor, actionError error) error {
 	// Detach from the parent timeout, but copy the trace span.
 	// We need to still release the lock even if the parent context timed out.
-	// Note that we are not using the user provided RemoteOperationTimeout
-	// here because it is possible that that timeout is too short.
 	ctx = trace.CopySpan(context.TODO(), ctx)
-	ctx, cancel := context.WithTimeout(ctx, defaultLockTimeout)
+	ctx, cancel := context.WithTimeout(ctx, RemoteOperationTimeout)
```
deepthi, line 459: 👍

go/vt/topo/locks_test.go
```diff
@@ -0,0 +1,81 @@
...
+
+	"github.com/spf13/pflag"
+	"github.com/stretchr/testify/require"
+
+	"vitess.io/vitess/go/internal/flag"
+)
+
+func TestGetLockTimeout(t *testing.T) {
+	tests := []struct {
+		desription                  string
```
deepthi, line 32:
```suggestion
		description                  string
```

go/vt/vttablet/tabletmanager/tm_init.go
```diff
@@ -901,6 +901,10 @@ func (tm *TabletManager) initializeReplication(ctx context.Context, tabletType t
 	}
 
 	// Set primary and start replication.
+	if currentPrimary.Tablet.MysqlHostname == "" {
+		log.Warningf("primary tablet in the shard record doesn't have mysql hostname specified. probably because that tablet shutdown.")
```
deepthi, line 905:
```suggestion
		log.Warningf("primary tablet in the shard record does not have mysql hostname specified, possibly because that tablet has been shut down.")
```

deepthi (approved):
Nice work cleaning up a whole bunch of incorrect timeout handling in this code.
Please address all feedback before merging.

doc/releasenotes/16_0_0_summary.md
GuptaManan100: Actually, the defaults are 15 and 45 seconds, and we don't want lock-timeout to be  at least 15 seconds more than remote_operation_timeout, we want it to be atleaset 2 times it and a littlle more.

Well in the previous release this bug exists, and only after upgrading and specifying both the flags (or specifying neither) can they really get rid of the bug. So the recommended process is to specify the remote-operation_timeout first so that while upgrading their behaviour remains the same. Once ugpraded they can choose whatever values they want.

go/flags/endtoend/vtctld.txt
GuptaManan100: This is a effect of https://github.com/vitessio/vitess/blob/8b8b135187bb8bf88409711e7d3948a5e953361c/go/vt/wrangler/wrangler.go#L39 and https://github.com/vitessio/vitess/blob/8b8b135187bb8bf88409711e7d3948a5e953361c/go/vt/vtctld/action_repository.go#L38

Apparently there are two action_timeouts 🤣. That action_timeout is a parameter for vtctldclient and vtctlclient. This one is taken by vtctld and applies only to `ApplyKeyspaceAction`, `ApplyShardAction` and `ApplyTabletAction`. This is the commit that introduced it - https://github.com/GuptaManan100/vitess/commit/facdb181da5c5adc680eca9cafc543086ad9e714

go/flags/endtoend/vtorc.txt
GuptaManan100: I have deprecated it - 
```
fs.Duration("lock-shard-timeout", 30*time.Second, "Duration for which a shard lock is held when running a recovery")
_ = fs.MarkDeprecated("lock-shard-timeout", "Please use lock-timeout instead.")
```

Deprecated flags don't show up in the flags output either.

go/internal/flag/flag.go
GuptaManan100: I don't have strong opinions. Whatever we feel is best. I'll change it to IsFlagProvided for now.

go/test/endtoend/reparent/newfeaturetest/reparent_test.go
```diff
@@ -17,155 +17,36 @@ limitations under the License.
...
-		clusterInstance.Keyspaces[0].Shards[0].Vttablets = append(clusterInstance.Keyspaces[0].Shards[0].Vttablets, newReplica)
-		// Check that we can replicate to it and semi-sync is setup correctly on it
-		utils.ConfirmReplication(t, tablets[3], []*cluster.Vttablet{tablets[0], tablets[1], tablets[2], newReplica})
-		utils.CheckSemiSyncSetupCorrectly(t, newReplica, "ON")
-	}
-}
-
-// TestFullStatus tests that the RPC FullStatus works as intended.
-func TestFullStatus(t *testing.T) {
+func TestRecoverWithMultipleVttabletFailures(t *testing.T) {
```
rsajwani, line 28: Can we write some header (comments) around what this test basically is testing..

go/vt/servenv/servenv_test.go
```diff
@@ -65,9 +65,7 @@ func TestFireOnCloseHooksTimeout(t *testing.T) {
 		time.Sleep(1 * time.Second)
 	})
 
-	// we deliberatly test the flag to make sure it's not accidently set to a
-	// high value.
-	if finished, want := fireOnCloseHooks(onCloseTimeout), false; finished != want {
+	if finished, want := fireOnCloseHooks(1*time.Nanosecond), false; finished != want {
```
rsajwani, line 68: 1ns is here to check backward compact. given the default value is change now ?

go/vt/topo/locks_test.go
```diff
@@ -0,0 +1,81 @@
...
+	"os"
+	"testing"
+	"time"
+
+	"github.com/spf13/pflag"
+	"github.com/stretchr/testify/require"
+
+	"vitess.io/vitess/go/internal/flag"
+)
+
```
rsajwani, line 29: Header missing.. Can you write few lines about what this test is doing.

rsajwani (approved):
overall LGTM

go/vt/topo/locks.go
GuptaManan100: Unlock is just a single remote operation. It doesn't need more time than a RemoteOperationTimeout. getLockTimeout is the time for which we want to lock the shard for.

go/test/endtoend/reparent/utils/utils.go
```diff
@@ -129,7 +129,7 @@ func setupCluster(ctx context.Context, t *testing.T, shardName string, cells []s
 		// the replication manager to silently fix the replication in case ERS or PRS mess up. All the
 		// tests in this test suite should work irrespective of this flag. Each run of ERS, PRS should be
 		// setting up the replication correctly.
-		"--disable_active_reparents")
+		"--disable-replication-manager")
```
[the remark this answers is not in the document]
GuptaManan100: This change is required because in the tests we test that a tablet that is restarted starts replicating from the primary again. So we want the replication initialization code to run. `disable_active_reparents` blocks both the replication manager and the replication initialization, but `disable-replication-manager` only blocks the first.

We required this change after increasing the onclose timeout. Earlier, because onclose timeout was 1ns, the vttablets when shutting down wouldn't have the time to remove the hostname, port from the vttablet records. This caused a call to ERS to have a hanging SetReplicationSource. This call would succeed when the tablet was up again. But this is incorrect and not want the test really wanted to test.

After increasing the onclose timeout, the vttablets now remove their hostname, port information from the tablet records. This causes the SetReplicationSource call to be directed at a tablet with address `:0`. So that call doesn't setup the replication. Instead, we require the replication initialization to do it properly. This is the reason for this change.

These changes resolve flakiness in the ERS tests which showed itself as `vttablet timed out after 10s` failing to reach the Serving state, in the few cases that onclose function did complete before timing out.

go/vt/servenv/servenv_test.go
GuptaManan100: I made this change, because otherwise the test would need to run for more than 10 seconds. The intent of the test is to verify that `fireOnCloseHooks` respects the timeout provided to it and that it returns after that much time. If we give the default here, we would need a sleep for like 11 seconds or something in the Onclose function and that would just slow down the test.

go/vt/topo/locks_test.go
GuptaManan100: Done.

go/test/endtoend/reparent/newfeaturetest/reparent_test.go
GuptaManan100: Done

go/flags/endtoend/vtorc.txt
GuptaManan100: Added this to the summary too.

PR #11888, opened 2022-12-06
Fix MySQL56 GTID parsing when SID/UUID repeats

shlomi-noach:
## Description

Current codebase has a bug parsing MySQL56 GTID values, when the same SID (or UUID) repeats itself. For example, the following GTID set is valid:
```
00021324-1111-1111-1111-111111111111:100-200,00021324-1111-1111-1111-111111111111:300-400
```
canonically, it is equal to:
```
00021324-1111-1111-1111-111111111111:100-200:300-400
```

but the current code erroneously overwrites existing SID entries and so we end up parsing as:
```
00021324-1111-1111-1111-111111111111:300-400
```

This PR fixes the parsing logic and adds a few tests. The new behavior is compatible with MySQL.

## Checklist

### DIFF
--- a/go/mysql/mysql56_gtid_set.go
+++ b/go/mysql/mysql56_gtid_set.go
@@ -119,6 +119,12 @@ func ParseMysql56GTIDSet(s string) (Mysql56GTIDSet, error) {
 			continue
 		}
 
+		if sidIntervals, ok := set[sid]; ok {
+			// SID already exists, we append
+			// Example:  "00010203-0405-0607-0809-0a0b0c0d0e0f:1-5,00010203-0405-0607-0809-0a0b0c0d0e0f:10-20"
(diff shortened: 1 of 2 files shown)


dbussink (approved):
Can you also open an issue for this and link to it?

mattlord (approved):

mattlord:
At a quick glance, it looks like we probably have the same issue with MariaDB GTIDs: https://github.com/vitessio/vitess/blob/main/go/mysql/mariadb_gtid.go#L79

We no longer support MariaDB outside of MoveTables for imports... but we may still want to fix it there too.

I approved w/o it as the MySQL fix is a serious one and obviously the priority.

shlomi-noach:
Cool - merging this and we can open a new PR for MariaDB

PR #11889, opened 2022-12-06
Online DDL: fix 'vtctlclient OnlineDDL' template queries

shlomi-noach:
## Description

At some point the behavior of `sqlparser.QueryMatchesTemplates(...)` changed such that if the template has multiple recurring `'val'` values, the new behavior expects them to be equal.

This breaks the `vtctlclient OnlineDDL` behavior, because OnlineDDL's query templates used recurring `'val'`. The templates are now updated, and a new unit test validates their behavior.

## Related Issue(s)

#6926

## Checklist

### DIFF
--- a/go/vt/vttablet/onlineddl/executor.go
+++ b/go/vt/vttablet/onlineddl/executor.go
@@ -78,9 +78,9 @@ var (
 )
 
 var vexecUpdateTemplates = []string{
-	`update _vt.schema_migrations set migration_status='val' where mysql_schema='val'`,
-	`update _vt.schema_migrations set migration_status='val' where migration_uuid='val' and mysql_schema='val'`,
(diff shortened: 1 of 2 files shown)


dbussink (approved):

deepthi:
@shlomi-noach can you (when you get a chance), create a smaller issue for this specific bug and link it to this PR?
I assume that the unit tests you have created mean that we don't need to actually trace this back to when it changed.

shlomi-noach:
@deepthi I will create an issue tomorrow, sorry for not doing this upfront. I haven't investigated yet where the behavior has changed. The new unit test protects Online DDL 's VExec against further behavior changes.

shlomi-noach:
Fixes https://github.com/vitessio/vitess/issues/11893

shlomi-noach:
Backport to v15: https://github.com/vitessio/vitess/pull/11951

PR #11892, opened 2022-12-07
VExplain statement

### LINKED ISSUE #4448: vexplain statement

### LINKED ISSUE #11621: RFC: New vexplain command having a format to glean information about vtgate and MySQL plans
### Description

We propose adding a new Vitess-specific command called `vexplain`, which extends the `explain` query that MySQL offers. The command would have multiple FORMATs that it supports -

| **Type Name** | **Description**                                                                           | **Actually Runs Queries**                 |
|:---------------:|---------------------------------------|------------------------|
|       PLAN (DEFAULT)      | VTGate plan in JSON format                                                                | No. (Only gets the VTGate Plan)           |
|     QUERIES    | All the MySQL queries run along with information about the keyspace shard they are run on | Yes.                                      |
|    ALL   | Same as JSON format, annotated with the MySQL plans for all the queries in ROUTEs         | Yes. (Also runs EXPLAIN queries on MySQL) |

The command `VEXPLAIN PLAN` is going to have more information than `EXPLAIN FORMAT = VITESS` and is intended to replace its usage. The output format is going to be different, though. The output of `VEXPLAIN QUERIES` is going to be equivalent to `EXPLAIN FORMAT = VTEXPLAIN`. We would like to deprecate the latter commands entirely, so as the clean up the `EXPLAIN` command.

The end goal is that the `EXPLAIN` command only routes the query to MySQL instances and for any Vitess level planning information, `VEXPLAIN` should be used.

We also propose the addition of the ALL type, which will also gather the MySQL plans for the executed queries and show them along with the VTGate plans. An example for this format would look something like this - 

```sql
mysql [localhost:8026] (ks) > VEXPLAIN ALL SELECT name FROM user where id = 1
+------------------------------------------------------------------------------------------------+
| VEXPLAIN                                                                                      |
+------------------------------------------------------------------------------------------------+
{
  "OperatorType": "Route",
  "Variant": "EqualUnique",
  "Keyspace": {
    "Name": "user",
    "Sharded": true
  },
  "FieldQuery": "select `name` from `user` where 1 != 1",
  "Query": "select `name` from `user` where id = 1",
  "Table": "`user`",
  "Values": [
    "INT64(1)"
  ],
  "Vindex": "user_index",
  "MySQLPlan": {
    "query_block": {
      "select_id": 1,
      "cost_info": {
        "query_cost": "1.00"
      },
      "table": {
        "table_name": "user",
        "access_type": "const",
        "possible_keys": [
          "PRIMARY"
        ],
        "key": "PRIMARY",
        "used_key_parts": [
          "id"
        ],
        "key_length": "8",
        "ref": [
          "const"
        ],
        "rows_examined_per_scan": 1,
        "rows_produced_per_join": 1,
        "filtered": "100.00",
        "cost_info": {
          "read_cost": "0.00",
          "eval_cost": "0.10",
          "prefix_cost": "0.00",
          "data_read_per_join": "24"
        },
        "used_columns": [
          "id",
          "name"
        ]
      }
    }
  }
}
+------------------------------------------------------------------------------------------------+
```
In the above output, we have the vtgate plan which is annotated with the MySQL plan for the actual query executed, which can be seen in the field `MySQLPlan` on all `Route` operators.

An example output for the ALL mode used for a join query `select id, user_id, name from user join user_extra where user.id = user_extra.col` will look like this -
```json
{
  "OperatorType": "Join",
  "Variant": "Join",
  "JoinColumnIndexes": "R:0,L:1,R:1",
  "JoinVars": {
    "user_extra_col": 0
  },
  "TableName": "user_extra_`user`",
  "Inputs": [
    {
      "OperatorType": "Route",
      "Variant": "Scatter",
      "Keyspace": {
        "Name": "user",
        "Sharded": true
      },
      "FieldQuery": "select user_extra.col, user_id from user_extra where 1 != 1",
      "Query": "select user_extra.col, user_id from user_extra",
      "Table": "user_extra",
      "MySQLPlan": {
        "query_block": {
          "select_id": 1,
          "cost_info": {
            "query_cost": "0.45"
          },
          "table": {
            "table_name": "user",
            "access_type": "ALL",
            "rows_examined_per_scan": 2,
            "rows_produced_per_join": 2,
            "filtered": "100.00",
            "cost_info": {
              "read_cost": "0.25",
              "eval_cost": "0.20",
              "prefix_cost": "0.45",
              "data_read_per_join": "48"
            },
            "used_columns": [
              "id",
              "name"
            ]
          }
        }
      }
    },
    {
      "OperatorType": "Route",
      "Variant": "EqualUnique",
      "Keyspace": {
        "Name": "user",
        "Sharded": true
      },
      "FieldQuery": "select id, `name` from `user` where 1 != 1",
      "Query": "select id, `name` from `user` where `user`.id = :user_extra_col",
      "Table": "`user`",
      "Values": [
        ":user_extra_col"
      ],
      "Vindex": "user_index",
      "MySQLPlan": {
        "query_block": {
          "select_id": 1,
          "cost_info": {
            "query_cost": "1.00"
          },
          "table": {
            "table_name": "user_extra",
            "access_type": "const",
            "possible_keys": [
              "PRIMARY"
            ],
            "key": "PRIMARY",
            "used_key_parts": [
              "col"
            ],
            "key_length": "8",
            "ref": [
              "const"
            ],
            "rows_examined_per_scan": 1,
            "rows_produced_per_join": 1,
            "filtered": "100.00",
            "cost_info": {
              "read_cost": "0.00",
              "eval_cost": "0.10",
              "prefix_cost": "0.00",
              "data_read_per_join": "24"
            },
            "used_columns": [
              "user_id",
              "col"
            ]
          }
        }
      }
    }
  ]
}
```

GuptaManan100:
## Description

In https://github.com/vitessio/vitess/issues/11621 we proposed adding a new Vitess-specific command called `vexplain`, which extends the `explain` query that MySQL offers. This PR does just that. :)

We are also deprecating `explain format=vitess` and `explain format=vtexplain`.

## Related Issue(s)
Fixes #11621
Fixes #4448

## Checklist

-   [x] Documentation was added or is not required - https://github.com/vitessio/website/pull/1306 and https://github.com/vitessio/website/pull/1303


go/vt/vtgate/engine/vexplain.go
```diff
@@ -0,0 +1,213 @@
...
+}
+
+func (v *VExplain) convertToVExplainAllResult(ctx context.Context, vcursor VCursor) (*sqltypes.Result, error) {
+	logEntries := vcursor.Session().GetVExplainLogs()
+	explainResults := make(map[Primitive]string)
+	for _, entry := range logEntries {
+		if entry.Target == nil || entry.Gateway == nil || entry.FiredFrom == nil {
+			continue
+		}
+		if explainResults[entry.FiredFrom] != "" || strings.Contains(entry.Query, "where 1 != 1") {
```
systay, line 120: we should not filter out impossible queries here

go/vt/vtgate/engine/vexplain.go
```diff
@@ -0,0 +1,213 @@
...
+		explainResults[entry.FiredFrom] = res.Rows[0][0].ToString()
+	}
+
+	planDescription := primitiveToPlanDescriptionWithSQLResults(v.Input, explainResults)
+	resultBytes, err := json.MarshalIndent(planDescription, "", "\t")
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO: interleave mysql plan with the vtgate plan instead of appending them
```
systay, line 140: remove this TODO

go/vt/vtgate/engine/delete.go
```diff
@@ -91,7 +91,7 @@ func (del *Delete) deleteVindexEntries(ctx context.Context, vcursor VCursor, bin
 	for i := range rss {
 		queries[i] = &querypb.BoundQuery{Sql: del.OwnedVindexQuery, BindVariables: bindVars}
 	}
-	subQueryResults, errors := vcursor.ExecuteMultiShard(ctx, rss, queries, false /* rollbackOnError */, false /* canAutocommit */)
+	subQueryResults, errors := vcursor.ExecuteMultiShard(ctx, del, rss, queries, false, false)
```
harshit-gangal, line 94: nit: can we keep the comments for the bool passed, easier to know what those `false, false` mean.

Long term we could have const that can be used instead of `false` directly.

harshit-gangal: If you agree, then there are other instances as well.

go/vt/vtgate/engine/delete.go
```diff
@@ -57,9 +57,9 @@ func (del *Delete) TryExecute(ctx context.Context, vcursor VCursor, bindVars map
 
 	switch del.Opcode {
 	case Unsharded:
-		return del.execUnsharded(ctx, vcursor, bindVars, rss)
+		return del.execUnsharded(ctx, del, vcursor, bindVars, rss)
 	case Equal, IN, Scatter, ByDestination, SubShard, EqualUnique, MultiEqual:
-		return del.execMultiDestination(ctx, vcursor, bindVars, rss, del.deleteVindexEntries)
+		return del.execMultiDestination(ctx, del, vcursor, bindVars, rss, del.deleteVindexEntries)
```
harshit-gangal, line 62: I think we do not need to pass primitive here, as the caller is the same primitive.

harshit-gangal (commented):
Can you add a test that shows the VExplain output will all the three types.
Not clear how it looks.

go/vt/vtgate/engine/vexplain.go
GuptaManan100: Done

go/vt/vtgate/engine/vexplain.go
GuptaManan100: Fixed

go/vt/vtgate/engine/delete.go
GuptaManan100: Fxied

systay:
> Can you add a test that shows the VExplain output will all the three types. Not clear how it looks.

Even more better. We added docs that show the output: https://deploy-preview-1306--vitess.netlify.app/docs/16.0/user-guides/sql/vexplain/

go/vt/sqlparser/keywords.go
```diff
@@ -253,6 +253,7 @@ var keywords = []keyword{
 	{"escape", ESCAPE},
 	{"escaped", ESCAPED},
 	{"event", EVENT},
+	{"exhaustive", EXHAUSTIVE},
```
systay, line 256: can we remove this?

go/vt/sqlparser/sql.y
```diff
@@ -302,7 +303,7 @@ func bindVariable(yylex yyLexer, bvar string) {
 
 // DDL Tokens
 %token <str> CREATE ALTER DROP RENAME ANALYZE ADD FLUSH CHANGE MODIFY DEALLOCATE
-%token <str> REVERT
+%token <str> REVERT QUERIES EXHAUSTIVE
```
systay, line 306: `EXHAUSTIVE` is no longer needed, I think

go/vt/sqlparser/keywords.go
GuptaManan100: Done

go/vt/sqlparser/sql.y
GuptaManan100: Done

systay (approved):

harshit-gangal (approved):

PR #11895, opened 2022-12-07
Fix CheckMySQL by setting the correct wanted state

### LINKED ISSUE #11885: Tracking CheckMySQL and Query Engine history
There have been multiple failures in `CheckMySQL` code that cause it to get stuck. This issue documents the history of all the changes that have been made to this codepath to better understand the intent and the evolution of the code over time.

GuptaManan100:
## Description

This PR fixes the bug uncovered in https://github.com/vitessio/vitess/issues/11885#issuecomment-1339141876. 
The main issue is that when we call `CheckMySQL` we do not set the correct wanted state. This leads to new queries not being blocked while we are waiting for the requests waitgroup to be empty. 
After looking at the history of how the code evolved in https://github.com/vitessio/vitess/issues/11885, the intent of the code and the bug was clear.

This PR fixes the bug and augments an existing test to verify that we indeed do not allow new queries to come in when checkMySQL has started running.

## Related Issue(s)

- Fixes #11885 

## Checklist

### DIFF
--- a/go/vt/vttablet/tabletserver/state_manager.go
+++ b/go/vt/vttablet/tabletserver/state_manager.go
@@ -38,10 +38,14 @@ type servingState int64
 
 const (
 	// StateNotConnected is the state where tabletserver is not
-	// connected to an underlying mysql instance.
+	// connected to an underlying mysql instance. In this state we close
+	// query engine since MySQL is probably unavailable
 	StateNotConnected = servingState(iota)
 	// StateNotServing is the state where tabletserver is connected
 	// to an underlying mysql instance, but is not serving queries.
+	// We do not close the query engine to not close the pool. We keep
+	// the query engine open but prevent queries from running by blocking them
+	// in StartRequest.
 	StateNotServing
 	// StateServing is where queries are allowed.
 	StateServing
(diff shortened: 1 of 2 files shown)


go/vt/vttablet/tabletserver/state_manager.go
```diff
@@ -330,11 +334,25 @@ func (sm *stateManager) checkMySQL() {
...
+		// unless they are part of a running transaction.
+		sm.setWantState(StateNotConnected)
 		sm.closeAll()
+
+		// Now that we reached the NotConnected state, we want to go back to the
+		// Serving state. The retry will only succeed once MySQL is reachable again
+		// Until then EnsureConnectionAndDB will error out.
+		sm.setWantState(StateServing)
 		sm.retryTransition(fmt.Sprintf("Cannot connect to MySQL, shutting down query service: %v", err))
 	}()
```
harshit-gangal, line 347: changes are good, 

Just wanted to know why we cannot just call `SetServingType` methods which take the same actions. Also, we would have a single code path for state changes.

GuptaManan100: We can't use that because in `mustTransition` we use `Acquire` whereas here we want to use `TryAcquire`. Also that call requires a tablet type to call to, which is something we don't want to tweek at all. So there is a trade-off between the two implementations, one. will have a single code-path but with many more if-else conditions, and this wherein there are 2 code-paths but both succinct in their implementations

harshit-gangal (approved):

frouioui (approved):

Issue #11559, opened 2022-10-21
RFC: View Support


harshit-gangal:
# Introduction

MySQL supports read views and updatable views. A View acts as a virtual table that is stored as a query and produces a result set when invoked. There are multiple use cases of why an application wants to use views but in this RFC we will focus on how we will add support for views in Vitess.

# Problem Statement

## Current Support
View support exists for unsharded keyspace via creating it on the underlying database.
For sharded keyspace, the support is limited to a single table as the user would be able to define sharding key for that view in the VSchema. This is useful to users who want to restrict the viewable columns for a table.

## New Support
There are limitations to the current support, like for sharded keyspace the current version is barely usable as it is very restrictive. Also, the users who are currently having views in unsharded keyspace cannot move to sharded keyspace.

This enhancement over current support will add complete support for **_readable views_** in the **_sharded keyspace_**.

# Proposal

This proposed document has its scope limited to only readable views. In the next iteration, we might extend it for updatable views, which will be a separate RFC.

## Query Serving
In the query serving path, the view definition will be cached at VTGates. The view table will be rewritten as a derived table containing the view definition. The normal query planning will follow thereafter.

An example is always helpful. For a view like:

```sql
create view user_view as select id, name from user
```

The query:

```sql
select * from user_view
```

would be rewritten to:

```sql
select * from (select id, name from user) as user_view
```

For the information_schema queries, describe queries, show queries, etc. that return the metadata, the view information would need to be injected to return the expected resultset back to the client.

## Storage / DDL Support
There are two ways we can store the view definition. Based on the location of the storage of views, there will be differences in how DDLs will be supported.

### 1: In the Topology Server (TopoServer)
The views will be stored as part of Vschema like how vindexes and tables information is stored.
Example for a keyspace:
```json
{
  "sharded": true,
  "vindexes": {...}
  "tables": {
    "user": {...},
    "user_extra": {...}
  },
  "views": {
    "user_view": "select id, name from user",
    "user_details_view": "select user.id, user_extra.col from user join user_extra on user.id = user_extra.user_id"
  }
}
```

#### DDL Support
The view definition can be added/removed using the vtctld `ApplyVSchema` command by providing the complete VSchema containing the view definition.
It can also be modified using VSchema DDL
```sql
alter vschema create view <view_name> as <select_statement>
alter vschema drop view <view_name>
```
Create View/Alter View/Drop View statements will be re-rewritten by vitess into VSchema DDLs.

Pros:
[...]

PR #11896, opened 2022-12-07
Add support for Views DDL

frouioui:
## Description

This PR adds DDL support for Views. The views will be stored by vttablet's underlying storage layer (MySQL) inside internal `_vt` schema in the `views` table.

## Related Issue(s)

- Part of https://github.com/vitessio/vitess/issues/11559

## Checklist


systay (changes requested):
Instead of sending down a query, I think we want to use an RPC to the tablet

harshit-gangal:
All CI checks are failing as I had a `skip CI` label on the PR. It will run on the next commit push.

go/vt/vttablet/tabletserver/query_executor.go
```diff
@@ -289,6 +293,130 @@ func (qre *QueryExecutor) txConnExec(conn *StatefulConnection) (*sqltypes.Result
...
+		Select:      stmt.Select,
+		CheckOption: stmt.CheckOption,
+		Comments:    stmt.Comments,
+	}
+	bindVars := qre.generateBindVarsForViewDDLInsert(createViewDDL)
+	sql, _, err := qre.generateFinalSQL(qre.plan.FullQuery, bindVars)
+	if err != nil {
+		return nil, err
+	}
+	qr, err := withddl.New([]string{mysql.CreateViewsTable}).Exec(qre.ctx, sql, conn.Exec, conn.Exec)
```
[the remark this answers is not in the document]
harshit-gangal: **ack**

harshit-gangal:
> Looks pretty good. Had one small comment, and one question.
> 
> Should we not add the enable-views flag to the `vtexplain` binary as well?

I do not think we can support views on vtexplain. As there is no storage layer and/or communication layer (health checks).

go/vt/vttablet/tabletserver/query_executor.go
```diff
@@ -289,6 +293,130 @@ func (qre *QueryExecutor) txConnExec(conn *StatefulConnection) (*sqltypes.Result
...
+	}
+	sql, _, err := qre.generateFinalSQL(p.GenerateFullQuery(sel), bindVars)
+	if err != nil {
+		return err
+	}
+
+	qr, err := withddl.New([]string{mysql.CreateViewsTable}).Exec(qre.ctx, sql, conn.Exec, conn.Exec)
+	if err != nil {
+		return err
+	}
```
GuptaManan100, line 405: Do we really need it to be withDDL, with the create view query given that we already add it to the inializaiton?

GuptaManan100 (approved):
Same comment for other usages withDDL, otherwise LGTM

GuptaManan100:
@systay with the work we are doing to add `vexplain` command, won't it make sense to just deprecate the `vtexplain` binary entirely? We can support the views very easily in `vexplain` from vtgate, with almost no effort.

systay:
> @systay with the work we are doing to add `vexplain` command, won't it make sense to just deprecate the `vtexplain` binary entirely? We can support the views very easily in `vexplain` from vtgate, with almost no effort.

No-one wants to get rid of vtexplain more than me, but `vexplain` is not enough. One very common usage pattern for vtexplain is to check how queries would behave in sharded mode given different vschemas, and that is difficult to do using `vexplain`

GuptaManan100:
I see. You make a very good point.

go/vt/vttablet/tabletserver/query_executor.go
harshit-gangal: View table gets created when the first time view DDL is received. 
If we do not use with, then have to check on the error message received whether it is due to non-existing _vt.views or something else.

harshit-gangal: the initialization is in the health_streamer and we have seen in the past that it sometimes is behind. In the meantime a query can be received which will return with _vt.Views not existing.

systay (approved):

PR #11897, opened 2022-12-07
Upgrade to `go1.18.9`

frouioui:
## Description

This Pull Request upgrades the go version used on `release-15.0` to `go1.18.9`.

### DIFF
--- a/.github/workflows/check_make_vtadmin_authz_testgen.yml
+++ b/.github/workflows/check_make_vtadmin_authz_testgen.yml
@@ -41,7 +41,7 @@ jobs:
       uses: actions/setup-go@v2
       if: steps.skip-workflow.outputs.skip-workflow == 'false' && steps.changes.outputs.vtadmin_changes == 'true'
       with:
-        go-version: 1.18.7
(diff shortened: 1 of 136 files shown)


deepthi (changes requested):
We should not do major golang version upgrades on release branches. We should only be doing security patch upgrades.

deepthi (approved):

rsajwani:
does go upgrade always result in upgrade in bootstrap version or is it just you did both as part of same PR to save CI cycle.?
Rest looks good to me.

GuptaManan100 (approved):
Approval for after answering Rameez's comment.

frouioui:
> does go upgrade always result in upgrade in bootstrap version or is it just you did both as part of same PR to save CI cycle.?
Rest looks good to me.
Also its would be good if you can mention bootstrap version upgrade in description as well.

Hello @rsajwani, we always bump the bootstrap version when upgrading the Go version. When we upgrade the Go version on a release branch, we bump the bootstrap version by `0.1`. On `main`, we bump the version by `1.0`.

PR #11905, opened 2022-12-07
Upgrade to `go1.19.4`

frouioui:
## Description

This Pull Request upgrades the Go version of the project to `go1.19.4`.

## Checklist

### DIFF
--- a/docker/base/Dockerfile
+++ b/docker/base/Dockerfile
@@ -21,7 +21,7 @@
 # TODO(mberlin): Remove the symlink and this note once
 # https://github.com/docker/hub-feedback/issues/292 is fixed.
 
-ARG bootstrap_version=13.2
+ARG bootstrap_version=14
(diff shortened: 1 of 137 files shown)


docker/base/Dockerfile
```diff
@@ -21,7 +21,7 @@
 # TODO(mberlin): Remove the symlink and this note once
 # https://github.com/docker/hub-feedback/issues/292 is fixed.
 
-ARG bootstrap_version=13
+ARG bootstrap_version=13.2
```
deepthi, line 24: Why are we using a `.2` version on main? We should be using 14.

deepthi (changes requested):

docker/base/Dockerfile
frouioui: This is solved by e816bdcda3

deepthi (approved):

rsajwani (approved):

GuptaManan100 (approved):

PR #11906, opened 2022-12-07
sqlparser: new Equality API

vmg:
## Description

This PR is based on @systay's changes for semantic-aware comparison. I had an idea this morning on how to improve the comparison API so it doesn't take `nil` parameters everywhere. This is the result! I'm quite happy of how it looks!

The `sqlparser` package now exports a global `Equals` method that performs identical comparison through a `Comparator` struct, so the callers everywhere can use `sqlparser.Equals.Expr(a, b)`, or whatever the type is, to perform naive comparisons.

Likewise, advanced users like the semantics table can compare ASTs with semantically aware checks by returning their own `Comparator`!

While I was around, I cleaned up the way we invoke the `asthelpergen` generators.

cc @systay 

## Checklist


go/vt/sqlparser/ast_format_fast.go
```diff
@@ -19,7 +19,6 @@ package sqlparser
 
 import (
 	"fmt"
-
```
GuptaManan100, line 22: This change fails the static check. Goimports ins't happy

GuptaManan100 (approved):
I like this change.

systay (approved):

PR #11908, opened 2022-12-07
Pbibra fix rdonly patch for tabletgateway

pbibra:
## Description

Fixes a nuance with the `tabletgateway` code and our RDONLY patch where if the tablet selected is a RDONLY, tabletgateway passes in the original target request as REPLICA instead of RDONLY. 
This PR updates the target to whatever the selected tablet's target is similar to what `discoverygatweay` does.

More context: https://slack-pde.slack.com/archives/C01CVUUUREF/p1670437058252249

## Checklist


pbibra:
wrong base branch

PR #11911, opened 2022-12-08
bugfix: allow predicates without dependencies with derived tables to be handled correctly

### LINKED ISSUE #11912: bug: predicates on top of derived table
For a query such as:

```sql
SELECT 
  * 
FROM 
  (
    SELECT 
      'foobar', 
      NOW() as 'lock_date', 
      NOW() as 'created_at'
  ) AS tmp 
WHERE 
  NOT EXISTS (
    SELECT 
      lock_type 
    FROM 
      stoppable_lock 
    WHERE 
      lock_type = 'foobar' 
      AND deleted_flag = 'N'
);
```

Vitess will return:

```
SQL Error [1105] [HY000]: unsupported: unable to split predicates to derived table: not :__sq_has_values1
```

This is on a sharded keyspace

systay:
## Description
This PR is fixing two separate bugs that affected the same query.

First off: the gen4 planner was having a hard time dealing with predicates on top of derived tables that did not depend on the derived table.

Example:

```sql
SELECT 1 
FROM (SELECT `hello world`) AS tmp 
WHERE EXISTS(SELECT 1 FROM user WHERE id = 123)
```

This type of query failed with `unsupported: unable to split predicates to derived table: not :__sq_has_values1`.

Once that problem was solved, a different problem surfaced.

Vitess has a normalizer that auto-parameterizes queries so the plan cache can be shared between more queries. This means finding literals and replacing them with bindvars. It also does something called "expand star" - basically it finds all instances of `SELECT *` and replaces the star with the actual column being returned.

Say the query was something like:

```sql
select * from (select 1) as tmp
```

First the normalizer has a go at the query, and it replaces the literals in the query with bindvars. So the query after this step is:

```sql
select * from (select :vtg1) as tmp
```

Later, expandStar works on the query, and it will end up looking like this:

```sql
select tmp.`:vtg1` from (select :vtg1) as tmp
```

Finally, before sending it down to MySQL, we replace the bindvars with the actual value, so the query that is sent down is:

```sql
select tmp.`:vtg1` from (select 1) as tmp
```

And this query makes mysql sad.

## Related Issue(s)
Fixes #11912

## Checklist

### DIFF
--- a/go/test/endtoend/vtgate/queries/subquery/subquery_test.go
+++ b/go/test/endtoend/vtgate/queries/subquery/subquery_test.go
@@ -66,6 +66,7 @@ func TestSubqueriesExists(t *testing.T) {
 
 	mcmp.Exec("insert into t1(id1, id2) values (0,1),(1,2),(2,3),(3,4),(4,5),(5,6)")
 	mcmp.AssertMatches(`SELECT id2 FROM t1 WHERE EXISTS (SELECT id1 FROM t1 WHERE id1 > 0) ORDER BY id2`, `[[INT64(1)] [INT64(2)] [INT64(3)] [INT64(4)] [INT64(5)] [INT64(6)]]`)
+	mcmp.AssertMatches(`select * from (select 1) as tmp where exists(select 1 from t1 where id1 = 1)`, `[[INT32(1)]]`)
 }
 
 func TestQueryAndSubQWithLimit(t *testing.T) {
--- a/go/vt/sqlparser/normalizer.go
+++ b/go/vt/sqlparser/normalizer.go
@@ -42,10 +42,11 @@ func Normalize(stmt Statement, reserved *ReservedVars, bindVars map[string]*quer
 }
 
 type normalizer struct {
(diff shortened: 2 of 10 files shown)


harshit-gangal (approved):
LGTM

dbussink (approved):

lokeshpkumar:
Any idea when will this be available in a stable release?

frouioui:
> Any idea when will this be available in a stable release?

Hello @lokeshpkumar, we will release a new patch release this week. There is an issue that motivates this patch release: https://github.com/vitessio/vitess/issues/11945.

lokeshpkumar:
Hi @frouioui will this fix be released in 15.0.2 or some other patch ?

frouioui:
> Hi @frouioui will this fix be released in 15.0.2 or some other patch ?

Hello @lokeshpkumar, yes it will be included in `v15.0.2`. It was backported onto our release branch here: https://github.com/vitessio/vitess/pull/11922

lokeshpkumar:
Thank you @frouioui ....

PR #11920, opened 2022-12-09
Fix sending a ServerLost error when reading a packet fails

### LINKED ISSUE #11921: Bug Report: Panic in Rollback of a transaction after MySQL crash
### Overview of the Issue

Vttablet runs into a panic while running `Rollback` on a transaction after MySQL crashes. Because the go-routine panics, we don't call `endRequest` and this causes the request to wait in CheckMySQL to block.

The panic looks like this - 
```
E1128 15:11:07.099516       1 tabletserver.go:1361] Uncaught panic for Sql: "rollback", BindVars: {}:
startEphemeralPacketWithHeader cannot be used while a packet is already started.
vitess.io/vitess/go/mysql/conn.go:637 (0xde7fbc)
vitess.io/vitess/go/mysql/query.go:45 (0xdfea99)
vitess.io/vitess/go/mysql/query.go:320 (0xe01373)
vitess.io/vitess/go/mysql/query.go:303 (0x1202d08)
vitess.io/vitess/go/vt/dbconnpool/connection.go:47 (0x1202cfe)
vitess.io/vitess/go/vt/vttablet/tabletserver/connpool/dbconn.go:160 (0x134a804)
vitess.io/vitess/go/vt/vttablet/tabletserver/connpool/dbconn.go:188 (0x1483824)
vitess.io/vitess/go/vt/vttablet/tabletserver/stateful_connection.go:98 (0x1483816)
vitess.io/vitess/go/vt/vttablet/tabletserver/tx_pool.go:222 (0x14a5844)
...
```

### Reproduction Steps

1. Setup a cluster
2. Run transactions via vtgate
3. Stop MySQL
4. Repeat until the panic is observed. This is inherently a timing issue and doesn't repeat easily

### Binary Version

```sh
main and all previous versions
```

```sh
all
```

### Log Fragments

_No response_

GuptaManan100:
## Description

Once a connection has received a connection error from MySQL, we close the connection in `dbconn` and other callers. In order for this close to work properly we should be sending a MySQL.`CRServerLost` error. 
All the callers of `readEphemeralPacket` wrap the error and send this error so that the connection can be closed. However, there is one code-path that doesn't do this.

We believe that is an issue. Not sending a mysql.`CRServerLost` error causes the connection to remain open and be reused for the rollback call, even though the connection isn't usable any more. This causes a panic in the code, since we didn't cleanup the state when we error out the first time.

This PR fixes the issue after adding a failing test which ensures that we receive a `CRServerLost` error on MySQL crashing after only sending a field packet.

## Related Issue(s)

- Fixes #11921

## Checklist

### DIFF
--- a/go/mysql/query.go
+++ b/go/mysql/query.go
@@ -416,7 +416,7 @@ func (c *Conn) ReadQueryResult(maxrows int, wantfields bool) (*sqltypes.Result, 
 	for {
 		data, err := c.readEphemeralPacket()
 		if err != nil {
-			return nil, false, 0, err
+			return nil, false, 0, NewSQLError(CRServerLost, SSUnknownSQLState, "%v", err)
 		}
 
 		if c.isEOFPacket(data) {
--- a/go/mysql/conn_flaky_test.go
+++ b/go/mysql/conn_flaky_test.go
@@ -1172,6 +1172,15 @@ func (t testRun) ComQuery(c *Conn, query string, callback func(*sqltypes.Result)
 	if strings.Contains(query, "panic") {
 		panic("test panic attack!")
 	}
+	if strings.Contains(query, "close before rows read") {
+		c.writeFields(selectRowsResult)
+		// We want to close the connection after the fields are written
+		// and read on the client. So we sleep for 100 milliseconds
+		time.Sleep(100 * time.Millisecond)
+		c.Close()
+		return nil
+	}
+
 	if strings.Contains(query, "twice") {
 		callback(selectRowsResult)
 	}
--- a/go/mysql/query_test.go
+++ b/go/mysql/query_test.go
@@ -351,6 +351,32 @@ func TestComStmtClose(t *testing.T) {
 
 }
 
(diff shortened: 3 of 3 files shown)


go/mysql/query.go
```diff
@@ -416,7 +416,7 @@ func (c *Conn) ReadQueryResult(maxrows int, wantfields bool) (*sqltypes.Result,
 	for {
 		data, err := c.readEphemeralPacket()
 		if err != nil {
-			return nil, false, 0, err
+			return nil, false, 0, NewSQLError(CRServerLost, SSUnknownSQLState, "%v", err)
```
frouioui, line 419: Just wondering, why are we returning a `SSUnknownSQLState`?

frouioui (approved):

go/mysql/query.go
harshit-gangal: This is what MySQL set for SSState when CRServerLost is the error code.

GuptaManan100: Yep exactly

PR #11923, opened 2022-12-09
[ci issue] Tests are running on older versions that do not support the query

systay:
## Description
In a recent PR, I accidentally did not mark on of the end to end tests to only run on the latest version of Vitess.
This change set fixes that issue.

## Related Issue(s)
The problem was introduced in #11875


go/test/endtoend/vtgate/queries/subquery/subquery_test.go
```diff
@@ -60,7 +60,7 @@ func TestSubqueriesHasValues(t *testing.T) {
 
 // Test only supported in >= v14.0.0
 func TestSubqueriesExists(t *testing.T) {
-	utils.SkipIfBinaryIsBelowVersion(t, 14, "vtgate")
+	utils.SkipIfBinaryIsBelowVersion(t, 16, "vtgate")
```
frouioui, line 63: We should also change the comment above the method

frouioui (approved):

harshit-gangal (approved):

PR #11926, opened 2022-12-09
Better clone of the VCursor

frouioui:
## Description

This Pull Request enhances the clone of the VCursor by performing a safer clone of the whole struct.

## Checklist


systay (changes requested):
so, I think you should put the new methods on `vcursorImpl`, `noopVCursor` and `loggingVCursor` together with the rest of the definition of these structs.

there is not clear reason why these should live in separate files

frouioui:
> so, I think you should put the new methods on `vcursorImpl`, `noopVCursor` and `loggingVCursor` together with the rest of the definition of these structs.
> 
> there is not clear reason why these should live in separate files

That's a good point. I moved the new methods in their corresponding files via [056d7c7](https://github.com/vitessio/vitess/pull/11926/commits/056d7c7d32bef4fb11202cc2edb8e53a928f7b91)

go/vt/vtgate/vcursor_impl.go
```diff
@@ -1085,3 +1093,10 @@ func (vc *vcursorImpl) GetVTExplainLogs() []engine.ExecuteEntry {
 func (vc *vcursorImpl) FindRoutedShard(keyspace, shard string) (keyspaceName string, err error) {
 	return vc.vschema.FindRoutedShard(keyspace, shard)
 }
+
+func (vc *vcursorImpl) BackgroundVCursor() engine.VCursor {
+	cloneCtx := vc.clone()
+	sess := proto.Clone(cloneCtx.safeSession.Session).(*vtgatepb.Session)
```
vmg, line 1099: Why are you not setting `context.Background()` here?

go/vt/vtgate/engine/primitive.go
```diff
@@ -194,6 +194,9 @@ type (
 
 		// SetTransactionIsolation sets the transaction isolation level for any new transaction on the session.
 		SetTransactionIsolation(isolation querypb.ExecuteOptions_TransactionIsolation)
+
+		// BackgroundVCursor returns the underlying clonned VCursor
```
vmg, line 198:
```suggestion
		// BackgroundVCursor returns a copy of this VCursor that can be used to perform Vitess queries after the RPC that instantiated this cursor has terminated.
```

vmg (changes requested):
You forgot to backport the `context.Background()` reset?

vmg (approved):

systay (approved):

harshit-gangal:
The PR description does not state why this was done. Can you state the reason for doing this?

PR #11938, opened 2022-12-11
support named global keyspaces

maxenglander:
# Description
    
Vitess currently supports globally routing queries to unique tables when the connection does not have a default keyspace and the tables in the `FROM` list are not keyspace-qualified.
    
This PR adds support for explicitly requesting global routing:
    
* Adds a VTGate flag `--global-keyspace` that can be specified multiple times.
* When a connection uses a `--global-keyspace`, it behaves the same was as if no default keyspace is set on the connection.
* When a query for a table is qualified with a `--global-keyspace`, it is routed to a globally unique table, even if it is in a different keyspace than the connection default keyspace.

# Details

Some details about the implementation:

 * Adds a `global` object to `*vindexes.VSchema`, and moves `uniqueTables` and `uniqueVindexes` into that object.
 * If an existing keyspace shares a name with a `--global-keyspace`, the existing keyspace will be used with non-global routing. This conflict resolution strategy prefers preserving existing behavior.
 * When a connection is set to use a `--global-keyspace`, `SELECT database()` will return that keyspace name. Maybe this isn't what we want?
 * Queries like `SHOW TABLES IN [global_ks]` do not currently work.

# Use cases
    
This feature is useful for frameworks like Hibernate which are required by to set a database name on the connection. Those frameworks are unable to take advantage of global routing.

# Checklist

 * [x] Add support for `--global-keyspace`.
 * [x] Write tests.
 * [ ] Write docs.


maxenglander:
Hi @deepthi I know better by now that I'm supposed to always link a PR to an issue, so my bad! Create an issue: https://github.com/vitessio/vitess/issues/11940.

For context, I originally developed this PR for an enterprise PlanetScale customer, and then mentioned it in that offline discussion because it seems like it could help there too.

maxenglander:
> We have enough flags as-is, anything that adds a new flag needs to be really necessary.

That's fair. I wonder if there's a reasonable way to do this without a new flag. Choosing a database like `global` is obviously not going to work for people that already have a VSchema keyspace by that name. 

But we could introduce this feature in such a way that that any existing VSchema keyspaces named `global` override a virtual global keyspace by that name. If in the future this were to become popular enough that people want to be able to use it in spite of having an existing `global` keyspace, then maybe that would be a good time to consider a `--global-keyspace` flag.

harshit-gangal:
I understand the intent here but I could not understand the use case for this change. Adding a use-case will help if this is useful to the users.

maxenglander:
@harshit-gangal added! There are also a lot more details in https://github.com/vitessio/vitess/issues/11940.

harshit-gangal:
> @harshit-gangal added! There are also a lot more details in #11940.

I am not sure how the problem with `show full tables` is solved with the change.
As only partial tables will be returned which will be present on that keyspace.

maxenglander:
@harshit-gangal yes good observation. I had noted in the issue Details section:

> Queries like SHOW TABLES IN [global_ks] do not currently work.

In order to address that we will need a PR like this one https://github.com/vitessio/vitess/pull/11882. I can revive that PR if we decide to move forward with this one.

maxenglander:
Returned to draft until I fix the merge conflicts.

test/config.json
```diff
@@ -621,6 +621,15 @@
...
+		"vtgate_queries_global": {
+			"File": "unused.go",
+			"Args": ["vitess.io/vitess/go/test/endtoend/vtgate/queries/global"],
+			"Command": [],
+			"Manual": false,
+			"Shard": "vtgate_queries",
+			"RetryMax": 1,
+			"Tags": []
+		},
 		"vtgate_concurrentdml": {
```
harshit-gangal, line 633: let's not crease another CI workflow for it and try to reuse any existing ones.

go/vt/vtgate/vtgate.go
```diff
@@ -142,6 +146,7 @@ func registerFlags(fs *pflag.FlagSet) {
 	fs.BoolVar(&enableSchemaChangeSignal, "schema_change_signal", enableSchemaChangeSignal, "Enable the schema tracker; requires queryserver-config-schema-change-signal to be enabled on the underlying vttablets for this to work")
 	fs.StringVar(&schemaChangeUser, "schema_change_signal_user", schemaChangeUser, "User to be used to send down query to vttablet to retrieve schema changes")
 	fs.IntVar(&queryTimeout, "query-timeout", queryTimeout, "Sets the default query timeout (in ms). Can be overridden by session variable (query_timeout) or comment directive (QUERY_TIMEOUT_MS)")
+	fs.StringSliceVar(&globalKeyspaceNames, "global-keyspace", globalKeyspaceNames, "Global routing behavior applies when no default keyspace is set on a connection, or when a global keyspace is used. Multiple global keyspaces may defined by specifying this flag multiple times.")
 	fs.StringVar(&queryLogToFile, "log_queries_to_file", queryLogToFile, "Enable query logging to the specified file")
```
harshit-gangal, line 150: why do we need multiple names, I think 1 should be enough to switch to global routing.

go/vt/vtgate/vindexes/vschema.go
```diff
@@ -166,6 +178,53 @@ func (col *Column) MarshalJSON() ([]byte, error) {
 	})
 }
 
+type globalSchema struct {
+	keyspaceNames map[string]any
+	tables        map[string]*Table
+	vindexes      map[string]Vindex
+}
+
```
harshit-gangal, line 186: it is not clear how they will receive input for global schema to keyspaces mapping.
Should we not just have a single global schema.

go/vt/vtgate/vtgate.go
maxenglander: We don't, not sure why I did this. I think we could also hardcode a global ks name if we can be sure that it won't conflict with existing user db names. Do you think `@global` could work?

go/vt/vtgate/vindexes/vschema.go
maxenglander: I'm not sure I'm following the question or the suggestion. Could you elaborate or rephrase?

maxenglander:
Spoke with members of Vitess eng. team on a call and while everyone appreciates the motivation behind this PR, there are concerns with layering additional functionality on top of global routing in this way. They would like to take a step back and think about the problem more holistically. Will leave the parent issue open for now.

Issue #11755, opened 2022-11-16
Add a VTGate API to do expand star on information schema queries


frouioui:
We would like to add an API that allows our planner to get the column information of information schema views in order to do expand star on information schema queries.

Part of https://github.com/vitessio/vitess/issues/11754.

PR #11941, opened 2022-12-12
[planner] Schema information on the information_schema views

systay:
## Description
This makes it so that Vitess knows the columns and types of all `information_schema` views, which can then be used during semantic analysis.

While working on this I also found and fixed an [issue](https://github.com/vitessio/vitess/issues/11942) around resolving of derived table columns with different cases.

## Related Issue(s)
Fixes https://github.com/vitessio/vitess/issues/11755
Fixes https://github.com/vitessio/vitess/issues/11942

## Checklist


go/vt/vtgate/planbuilder/testdata/unsupported_cases.json
```diff
@@ -392,11 +392,6 @@
     "query": "select distinct a, b as a from user",
     "plan": "generating order by clause: ambiguous symbol reference: a"
   },
-  {
```
systay, line 395: this query didn't make much sense to me, so I removed it. The problem is with `in (select *` - `IN` works against a single column. I think we acceptable coverage of star expansion on system tables even without this query

go/vt/vtgate/planbuilder/testdata/systemtables_cases.json
```diff
@@ -208,6 +190,30 @@
       "Original": "select * from information_schema.tables where table_schema = 'user' union select * from information_schema.tables where table_schema = 'main'",
```
arthurschreiber, line 190: This reminds me that we could look into optimizing `UNION` queries to not _always_ split them up (e.g. in cases we will end up sending them to the same shard anyway).

go/vt/vtgate/planbuilder/testdata/systemtables_cases.json
```diff
@@ -1,164 +1,146 @@
...
         "Keyspace": {
           "Name": "main",
           "Sharded": false
         },
-        "FieldQuery": "select column_name from information_schema.`columns` where 1 != 1",
-        "Query": "select column_name from information_schema.`columns` where table_schema = schema()",
-        "Table": "information_schema.`columns`"
+        "FieldQuery": "select a.`ENGINE`, b.DATA_TYPE from information_schema.`TABLES` as a, information_schema.`COLUMNS` as b where 1 != 1",
+        "Query": "select a.`ENGINE`, b.DATA_TYPE from information_schema.`TABLES` as a, information_schema.`COLUMNS` as b",
+        "Table": "information_schema.`COLUMNS`, information_schema.`TABLES`"
```
frouioui, line 51: any idea why the two tables are in inverse orders between the Gen4 and V3 plans? Is that flakky?

go/vt/vtgate/semantics/info_schema.go
```diff
@@ -0,0 +1,973 @@
...
+	"vitess.io/vitess/go/vt/proto/query"
+	topodatapb "vitess.io/vitess/go/vt/proto/topodata"
+	"vitess.io/vitess/go/vt/sqlparser"
+	"vitess.io/vitess/go/vt/vtgate/vindexes"
+)
+
+// getInfoSchema returns a map of all information_schema tables and their columns with types
+// To recreate this information from MySQL, you can run the test in info_schema_gen_test.go0
+func getInfoSchema() map[string]map[string]query.Type {
+	infSchema := map[string]map[string]query.Type{}
```
frouioui, line 34: nit: maybe adding a link to the doc could help to know what each table/column is used for: https://dev.mysql.com/doc/refman/8.0/en/information-schema-table-reference.html

go/vt/vtgate/semantics/info_schema.go
```diff
@@ -0,0 +1,973 @@
...
+	cols["USER"] = query.Type(6165)
+	cols["HOST"] = query.Type(6165)
+	cols["GRANTEE"] = query.Type(6165)
+	cols["GRANTEE_HOST"] = query.Type(6165)
+	cols["ROLE_NAME"] = query.Type(6165)
+	cols["ROLE_HOST"] = query.Type(6165)
+	cols["IS_GRANTABLE"] = query.Type(6165)
+	cols["IS_DEFAULT"] = query.Type(6165)
+	cols["IS_MANDATORY"] = query.Type(6165)
+	infSchema["ADMINISTRABLE_ROLE_AUTHORIZATIONS"] = cols
```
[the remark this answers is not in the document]
systay: it's generated and not meant to be readable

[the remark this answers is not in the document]
frouioui: Ah that's what `// To recreate this information from MySQL, you can run the test in info_schema_gen_test.go0
` is, okay!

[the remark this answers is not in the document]
frouioui: makes sense then

go/vt/vtgate/semantics/info_schema_gen_test.go
```diff
@@ -0,0 +1,163 @@
...
+	b := new(strings.Builder)
+
+	db, err := sql.Open("mysql", "root@tcp(127.0.0.1:3306)/test")
+	require.NoError(t, err)
+	defer db.Close()
+
+	for _, tbl := range informationSchemaTables {
+		b.WriteString("cols = map[string]query.Type{}\n")
+		result, err := db.Query(fmt.Sprintf("show columns from information_schema.`%s`", tbl))
+		require.NoError(t, err)
```
frouioui, line 43: ah this is smart, yes

go/vt/vtgate/planbuilder/testdata/systemtables_cases.json
systay: don't know why. I don't think it's flaky, it's never failed locally even after many runs

go/vt/vtgate/planbuilder/testdata/union_cases.json
```diff
@@ -2390,5 +2390,27 @@
         "user.user_extra"
       ]
     }
+  },
+  {
+    "comment": "systable union query in derived table with constraint on outside (star projection)",
```
systay, line 2395: this query was moved here from `unsuported_cases.json`

go/vt/vtgate/semantics/info_schema.go
```diff
@@ -0,0 +1,973 @@
...
+	cols["CHARACTER_SET_CLIENT"] = query.Type(6165)
+	cols["COLLATION_CONNECTION"] = query.Type(6165)
+	infSchema["VIEWS"] = cols
+	return infSchema
+}
+
+type infoSchemaWithColumns struct {
+	inner          SchemaInformation
+	infoSchemaData map[string]map[string]query.Type
+}
```
frouioui, line 931: Why adding a wrapper of our `SchemaInformation` and not adding this directly to the `VSchema` interface?

go/vt/vtgate/semantics/info_schema_gen_test.go
```diff
@@ -0,0 +1,163 @@
...
+	"INNODB_FT_CONFIG",
+	"INNODB_FT_DEFAULT_STOPWORD",
+	"INNODB_FT_DELETED",
+	"INNODB_FT_INDEX_CACHE",
+	"INNODB_FT_INDEX_TABLE",
+	"INNODB_INDEXES",
+	"INNODB_METRICS",
+	"INNODB_SESSION_TEMP_TABLESPACES",
+	"INNODB_TABLES",
+	"INNODB_TABLESPACES",
```
arthurschreiber, line 112: Does this mean we only support the 8.0 table names? E.g. `innodb_tablespaces` was called `innodb_sys_tablespaces` in MySQL 5.7, and I think there's others that got renamed as well.

arthurschreiber: Should we support the "old" table names when the vtgate runs in mysql 5.7 mode?

go/vt/vtgate/semantics/info_schema.go
```diff
@@ -0,0 +1,973 @@
...
+	for name, typ := range i.infoSchemaData[strings.ToUpper(tbl.Name.String())] {
+		cols = append(cols, vindexes.Column{
+			Name: sqlparser.NewIdentifierCI(name),
+			Type: typ,
+		})
+	}
+
+	sort.Slice(cols, func(i, j int) bool {
+		return cols[i].Name.String() < cols[j].Name.String()
+	})
```
frouioui, line 958: should the columns be sorted? according to MySQL docs, they do not seem to follow a specific alphabetical order: https://dev.mysql.com/doc/refman/8.0/en/information-schema-administrable-role-authorizations-table.html

go/vt/vtgate/semantics/info_schema_gen_test.go
systay: excellent point. yeah, we should support both

go/vt/vtgate/semantics/info_schema.go
systay: I sort them because otherwise all the expandStar tests become flaky

go/vt/vtgate/semantics/info_schema_gen_test.go
systay: added a commit for this now

go/vt/vtgate/semantics/info_schema.go
frouioui: makes sense okay, should they be sorted the same way MySQL sort them though? I am unsure if this is relevant. But imagine someone doing a `select * from information_schema.COLUMNS` and they expect to have `TABLE_CATALOG ` as the first column, but instead our first column is `COLUMN_COMMENT `.

systay: fixed in my last commit

go/vt/vtgate/planbuilder/testdata/systemtables_cases57.json
```diff
@@ -0,0 +1,1548 @@
...
+        "Query": "select fk.referenced_table_name as to_table, fk.referenced_column_name as primary_key, fk.column_name as `column`, fk.constraint_name as `name`, rc.update_rule as on_update, rc.delete_rule as on_delete from information_schema.referential_constraints as rc, information_schema.key_column_usage as fk where rc.constraint_schema = :__vtschemaname and rc.table_name = :rc_table_name and fk.referenced_column_name is not null and fk.table_schema = :__vtschemaname and fk.table_name = :fk_table_name",
+        "SysTableTableName": "[fk_table_name:VARCHAR(\"table_name\"), rc_table_name:VARCHAR(\"table_name\")]",
+        "SysTableTableSchema": "[VARCHAR(\"table_schema\"), VARCHAR(\"table_schema\")]",
+        "Table": "information_schema.key_column_usage, information_schema.referential_constraints"
+      }
+    }
+  },
+  {
+    "comment": "rails_query 6",
```
systay, line 556: here we should have rails query 5, but it uses an info_schema view that did not exist in 5.7 `information_schema.check_constraints`. There were two more queries in this file that I had to delete because they used this view

go/vt/vtgate/semantics/derived_table.go
```diff
@@ -66,7 +68,7 @@ func createDerivedTableForExpressions(expressions sqlparser.SelectExprs, cols sq
 func (dt *DerivedTable) dependencies(colName string, org originable) (dependencies, error) {
 	directDeps := org.tableSetFor(dt.ASTNode)
 	for i, name := range dt.columnNames {
-		if name != colName {
+		if !strings.EqualFold(name, colName) {
```
arthurschreiber, line 71: Is this fixing a potential bug / does this need to be backported? 🤔

arthurschreiber (approved):

go/vt/vtgate/semantics/derived_table.go
systay: It fixes https://github.com/vitessio/vitess/issues/11942

Yeah, we could backport just this change. We already have a test that exposes the problem in the plan tests. That's how I found it

go/vt/vtgate/semantics/table_collector.go
```diff
@@ -84,19 +84,16 @@ func (tc *tableCollector) up(cursor *sqlparser.Cursor) error {
...
+		isInfSchema := sqlparser.SystemSchema(t.Qualifier.String())
+		var err error
+		tbl, vindex, _, _, _, err = tc.si.FindTableOrVindex(t)
+		if err != nil {
+			return err
 		}
+		if tbl == nil && vindex != nil {
+			tbl = newVindexTable(t.Name)
+		}
+
```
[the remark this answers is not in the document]
systay: yeah, this is by design. `tc.si` now knows about information_schema views, so we want to ask it about these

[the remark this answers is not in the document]
harshit-gangal: I see now, Thank you

PR #11944, opened 2022-12-13
schemadiff: foreign key validation (tables and columns)

shlomi-noach:
## Description

We improve `schemadiff`'s support/validation for foreign keys as follows:

- `schemadiff` analyzes foreign key dependency across the schema
- it sorts tables by fk dependency (ie parents first, then children)
- it validates that there are no circular dependencies (only circle allowed is a self-referencing table)
- #10359 already validates that the local columns covered by a foreign key do in fact exist
- we now also check that the referenced (parent) table exists
- and that referenced columns in the parent table - exist
- and that the foreign key lists same count of columns local-vs-referenced
- and that the column types local-vs-referenced match: InnoDB only allows matching an `int` to an `int`, for example, and does not allow `int`-to-`bigint`. It does allow different `VARCHAR` length. 

Most of this takes place at the schema level, not at the table level.

Many tests added.

Still todo (in a different PR):

- validate indexes: each foreign key must have an appropriate index that covered columns in local table, and an appropriate index that covers referenced columns in parent table
  - an appropriate index covers the foreign key columns in same order as they appear in the foreign key. It may further extend to cover other columns, but the foreign key columns must be the prefix of the index column list.
- apply(): when adding a foreign key and no appropriate index exists, implicitly add it
- apply(): when creating a new table with foreign key and no appropriate index exists, implicitly add it
- apply(): do not allow `drop key` if it leaves a foreign key without an appropriate index
- validate uniqueness of constraints name across the schema (foreign keys and check constraints alike)

## Related Issue(s)

Foreign keys tracking issue: https://github.com/vitessio/vitess/issues/11975
`schemadiff` tracking issue: https://github.com/vitessio/vitess/issues/10203
Followup to https://github.com/vitessio/vitess/pull/10359

## Checklist


shlomi-noach:
Foreign keys tracking issue: https://github.com/vitessio/vitess/issues/11975

go/vt/schemadiff/errors.go
```diff
@@ -228,17 +228,70 @@ func (e *InvalidColumnInCheckConstraintError) Error() string {
...
+	ReferencedTable  string
+	ReferencedColumn string
+}
+
+func (e *InvalidReferencedColumnInForeignKeyConstraintError) Error() string {
+	return fmt.Sprintf("invalid column %s.%s referenced by foreign key constraint %s in table %s",
+		sqlescape.EscapeID(e.ReferencedTable), sqlescape.EscapeID(e.ReferencedColumn), sqlescape.EscapeID(e.Constraint), sqlescape.EscapeID(e.Table))
+}
+
+type MismatchingForeignKeyColumnCountError struct {
```
deepthi, line 263: Name suggestion: `ForeignKeyColumnCountMismatchError`. I don't know if users ever see this though so it may not matter too much.

shlomi-noach: Renamed

go/vt/schemadiff/errors.go
```diff
@@ -228,17 +228,70 @@ func (e *InvalidColumnInCheckConstraintError) Error() string {
 		sqlescape.EscapeID(e.Column), sqlescape.EscapeID(e.Constraint), sqlescape.EscapeID(e.Table))
 }
 
+type ForeignKeyDependencyUnresolvedError struct {
+	Table string
+}
+
+func (e *ForeignKeyDependencyUnresolvedError) Error() string {
```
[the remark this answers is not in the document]
shlomi-noach: You must have missed them, the tests are as follows:

`ForeignKeyDependencyUnresolvedError` has these unit tests:
- https://github.com/vitessio/vitess/pull/11944/files#diff-22a7f102b05d7044eda2deaa4242957aaffa51bfe99fb9a3e70dda855c6dc544R358-R378
- https://github.com/vitessio/vitess/pull/11944/files#diff-22a7f102b05d7044eda2deaa4242957aaffa51bfe99fb9a3e70dda855c6dc544R307-R310

`InvalidColumnInForeignKeyConstraintError` has https://github.com/vitessio/vitess/pull/11944/files#diff-22a7f102b05d7044eda2deaa4242957aaffa51bfe99fb9a3e70dda855c6dc544R295-R298

`ForeignKeyColumnCountMismatchError` has:
- https://github.com/vitessio/vitess/pull/11944/files#diff-22a7f102b05d7044eda2deaa4242957aaffa51bfe99fb9a3e70dda855c6dc544R299-R306
- https://github.com/vitessio/vitess/pull/11944/files#diff-2de8e8fe0a6b562dc9d4566b3de63c1718f3d3df4a938f9ce428ca22a7ebfc66R1481-R1486

`MismatchingForeignKeyColumnTypeError` has https://github.com/vitessio/vitess/pull/11944/files#diff-22a7f102b05d7044eda2deaa4242957aaffa51bfe99fb9a3e70dda855c6dc544R319-R342

shlomi-noach:
@deepthi for another look please :pray:

go/vt/schemadiff/errors.go
```diff
@@ -228,17 +228,70 @@ func (e *InvalidColumnInCheckConstraintError) Error() string {
...
+	ReferencedTable       string
+	ReferencedColumnCount int
+}
+
+func (e *ForeignKeyColumnCountMismatchError) Error() string {
+	return fmt.Sprintf("mismatching column count %d referenced by foreign key constraint %s in table %s. Expected %d",
+		e.ReferencedColumnCount, sqlescape.EscapeID(e.Constraint), sqlescape.EscapeID(e.Table), e.ColumnCount)
+}
+
+type MismatchingForeignKeyColumnTypeError struct {
```
deepthi, line 276: Sorry I didn't catch this in the previous review. It will be good to rename this as well - to `ForeignKeyColumnTypeMismatchError`.

deepthi (approved):

go/vt/schemadiff/errors.go
shlomi-noach: Renamed!

go/vt/schemadiff/schema.go
```diff
@@ -198,12 +212,57 @@ func (s *Schema) normalize() error {
...
 	}
 
+	// We now iterate all tables. We iterate "dependency levels":
+	// - first we want all tables that don't have foreign keys or which only reference themselves
+	// - then we only want tables that reference 1st level tables. these are 2nd level tables
+	// - etc.
+	// we stop when we have been unable to find a table in an iteration.
+	fkParents := map[string]bool{}
+	iterationLevel := 0
+	for ; ; iterationLevel++ {
```
dbussink, line 222: I'd probably write this slight different like this:

```go
for {
     iterationLevel++ 
     ...
}
```

Reads simpler to me, but no strong opinion really on this. 

The other question though is, would it be possible to craft a schema manually that triggers an infinite loop here? Should we have a safeguard for a max level we support?

go/vt/schemadiff/schema.go
```diff
@@ -238,6 +306,69 @@ func (s *Schema) normalize() error {
...
+			a.Unsigned == b.Unsigned &&
+			a.Zerofill == b.Zerofill &&
+			sqlparser.Equals.ColumnCharset(a.Charset, b.Charset) &&
+			sqlparser.Equals.SliceOfString(a.EnumValues, b.EnumValues)
+	}
+
+	// Now validate foreign key columns:
+	// - referenced table columns must exist
+	// - foreign key columns must match in count and type to referenced table columns
+	// - referenced table as an appropriate index over referenced columns
```
dbussink, line 327: `as an` => `has an`?

go/vt/schemadiff/schema.go
```diff
@@ -238,6 +306,69 @@ func (s *Schema) normalize() error {
...
+				referencedColumn, ok := referencedColumns[referencedColumnName]
+				if !ok {
+					return &InvalidReferencedColumnInForeignKeyConstraintError{Table: t.Name(), Constraint: cs.Name.String(), ReferencedTable: referencedTableName, ReferencedColumn: referencedColumnName}
+				}
+				if !colTypeEqualForForeignKey(coveredColumn.Type, referencedColumn.Type) {
+					return &ForeignKeyColumnTypeMismatchError{Table: t.Name(), Constraint: cs.Name.String(), Column: coveredColumn.Name.String(), ReferencedTable: referencedTableName, ReferencedColumn: referencedColumnName}
+				}
+			}
+
+			// TODO(shlomi): find a valid index
```
dbussink, line 369: Is having an index a logical strict requirement or more an implementation detail of how MySQL uses it to enforce constraints?

dbussink (approved):
Mostly some questions but nothing blocking.

go/vt/schemadiff/schema.go
shlomi-noach: > would it be possible to craft a schema manually that triggers an infinite loop

The algorithm addresses infinite loops. It leaves out any infinitely dependent views. There's a unit test for that: https://github.com/vitessio/vitess/blob/ec7e1a875224507ffd026b180f793205126cb336/go/vt/schemadiff/schema_test.go#L126-L135

shlomi-noach: just clarifying, that the rewrite would be:

```go
for {
     ...
     iterationLevel++ 
}
```

go/vt/schemadiff/schema.go
shlomi-noach: fixed

go/vt/schemadiff/schema.go
shlomi-noach: MySQL requires you to have an index over foreign key columns; a future PR enforces that. If you create a foreign key without creating an index, MySQL implicitly creates one on your behalf. The key must cover all the FK columns as a prefix. See future PR.

go/vt/schemadiff/schema.go

go/vt/schemadiff/schema.go
shlomi-noach: The future PR being https://github.com/vitessio/vitess/issues/12026

PR #11955, opened 2022-12-14
Deprecating VExec part1: removing client-side references

shlomi-noach:
## Description

We intended to deprecate `VExec` in `v12` and didn't. A previous iteration, https://github.com/vitessio/vitess/pull/8414, was incorrect because it broke `vtctl OnlineDDL`.

This PR resumes work to be included in `v16`. Additional and final work will only be included in `v17`. Changes are:

- `vtctl OnlineDDL` does not use `VExec` now. Instead, it uses:
  - `ApplySchema` for `cancel`, `retry`, `complete` commands
    the output is changed: previously the output indicated the number of affected migrations. The output now is empty on success.
  - `ExecuteFetchAsDBA` for `select` commands
    the output remains identical to existing codebase
- `vtctl VExec` command is removed.

For backwards compatibility, the server handling of `VExec` is still untouched. We will remove that code in `v17`.

## Related Issue(s)

- https://github.com/vitessio/vitess/pull/8414
- 
## Checklist

### DIFF
--- a/go/vt/vtctl/vtctl.go
+++ b/go/vt/vtctl/vtctl.go
@@ -3021,11 +3021,13 @@ func commandOnlineDDL(ctx context.Context, wr *wrangler.Wrangler, subFlags *pfla
 
 		tabletResults := map[string]*sqltypes.Result{}
 		for _, tablet := range resp.Tablets {
+			tabletAlias := topoproto.TabletAliasString(tablet.Alias)
+
 			qrproto, err := wr.ExecuteFetchAsDba(ctx, tablet.Alias, executeFetchQuery, 10000, false, false)
 			if err != nil {
 				return err
 			}
-			tabletResults[tablet.Alias.String()] = sqltypes.Proto3ToResult(qrproto)
+			tabletResults[tabletAlias] = sqltypes.Proto3ToResult(qrproto)
 		}
 		// combine results. This loses sorting if there's more then 1 tablet
 		combinedResults := queryResultForTabletResults(tabletResults)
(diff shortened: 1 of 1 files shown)


frouioui:
@shlomi-noach, I have added the `release notes (need details)` label as I believe this change should be written down in the `16_0_0_summary.md` file to inform users of Vitess. Feel free to remove the label if you think this is not applicable or if the update to the summary file is completed

shlomi-noach:
@frouioui I've added release notes and have removed the release-notes label

deepthi (approved):
Can you please create a follow up issue for removing the server side implementation? That should be added to the v17.0 project so that we don't forget.

shlomi-noach:
Followup issue: https://github.com/vitessio/vitess/issues/12053

shlomi-noach:
@frouioui for review/approval :pray:

frouioui (approved):
LGTM!

mattlord:
Note to myself and @rohit-nayak-ps that this means we need to provide an alternative to `VExec` in v17 for running a query against all tablets in a workflow e.g. to manually update the tablet_types or cells value if the topology changes in such a way that the workflow can no longer find a source tablet:
```
❯ vtctlclient VExec -- --help
Usage: VExec <ks.workflow> <query> --dry-run

Runs query on all tablets in workflow. Example: VExec merchant.morders "update _vt.vreplication set Status='Running'"

      --dry_run   Does a dry run of VExec and only reports the final query and list of tablets on which it will be applied
      --json      Output JSON instead of human-readable table
```

Technically you can do something like this, but it's obviously not the most user friendly thing:
```
vtctlclient VReplicationExec -- $(vtctlclient ListAllTablets -- --keyspace=customer --tablet_type=primary | awk '{print $1}') 'update _vt.vreplication set tablet_types="replica,primary" where workflow="commerce2customer"'
```

mattlord:
Note that we'll need to remove this docs page: https://vitess.io/docs/16.0/reference/vreplication/vexec/

shlomi-noach:
Followup in https://github.com/vitessio/vitess/pull/12074

PR #11959, opened 2022-12-14
[WIP] Consistent Hashing for tablet selection

arthurschreiber:
## Description

This pull request proposes changing the tablet selection logic in `vtgate` from a random selection (with preference to tablets in local cells), to using a consistent hashing scheme based on the session UUID (with preference to tablets in local cells).

When tablet selection is performed, we want to re-use the same tablet for the same session. We do this by making use of the session UUID. Each MySQL connection coming into `vtgate` is assigned a random UUID. The session UUID is fixed for the lifetime of the connection, with the exception of performing a session reset (which also resets the session UUID).

Tablets are put into a consistent hash, and a tablet is selected based on the session UUID.

This leads to the same tablet being selected over the lifetime of an incoming MySQL connection, until the point where the tablet topology changes (i.e. a new tablet is added, or an existing tablet becomes unhealthy). When this happens, the consistent hashing scheme will automatically redistribute some of the incoming connections to other available replicas, but most connections should stay unaffected.

Implementation details:
* `xxhash` is used for hashing values.
* "replicas"/"virtual node count" is set at 10. I'm not sure what a good value for this is.

---

Note: I haven't performed any benchmarking for this. My intention behind this PR is to first gather input on what other people think about this change, and then polishing the implementation.

## Related Issue(s)

https://github.com/vitessio/vitess/issues/11971

## Checklist


go/vt/proto/query/query.pb.go
```diff
@@ -1283,6 +1283,7 @@ type ExecuteOptions struct {
 	// TransactionAccessMode specifies the access modes to be used while starting the transaction i.e. READ WRITE/READ ONLY/WITH CONSISTENT SNAPSHOT
 	// If not specified, the transaction will be started with the default access mode on the connection.
 	TransactionAccessMode []ExecuteOptions_TransactionAccessMode `protobuf:"varint,14,rep,packed,name=transaction_access_mode,json=transactionAccessMode,proto3,enum=query.ExecuteOptions_TransactionAccessMode" json:"transaction_access_mode,omitempty"`
+	SessionUUID           string                                 `protobuf:"bytes,15,opt,name=SessionUUID,proto3" json:"SessionUUID,omitempty"`
```
arthurschreiber, line 1286: I'm not too happy about adding the `SessionUUID` here, but I couldn't find a better way to pass it through to the tablet selection logic. If anyone has a better idea how that could be solved, I'm all ears. 🙇‍♂️

go/vt/vtgate/tabletgateway_test.go
```diff
@@ -84,7 +85,7 @@ func TestTabletGatewayBeginExecute(t *testing.T) {
 
 func TestTabletGatewayShuffleTablets(t *testing.T) {
 	hc := discovery.NewFakeHealthCheck(nil)
-	tg := NewTabletGateway(context.Background(), hc, nil, "local")
+	tg := NewTabletGateway(context.Background(), hc, nil, "cell1")
```
arthurschreiber, line 88: I don't really understand how the existing `shuffleTablets` test cases where even passing? 🤔

go/vt/vtgate/tabletgateway.go
```diff
@@ -311,16 +313,16 @@ func (gw *TabletGateway) withRetry(ctx context.Context, target *querypb.Target,
...
-				th = t
-				break
-			}
+
+		// Old tablet selection implementation
+		if sessionUUID == "" {
+			th = gw.selectRandomTablet(tablets, invalidTablets)
+		} else {
+			th = gw.selectConsistentTablet(sessionUUID, tablets, invalidTablets)
 		}
```
arthurschreiber, line 324: We could put the tablet selection logic behind a CLI flag, if there's any concern in changing this logic.

go/vt/vtgate/tabletgateway.go
```diff
@@ -381,6 +383,109 @@ func (gw *TabletGateway) getStatsAggregator(target *querypb.Target) *TabletStatu
...
+
+		if tablet.Tablet.Alias.GetCell() == gw.localCell {
+			localTablets = append(localTablets, tablet)
+		} else {
+			externalTablets = append(externalTablets, tablet)
+		}
+	}
+
+	if len(localTablets) > 0 {
+		hash := NewConsistentHash()
```
arthurschreiber, line 465: Instead of re-generating the hashes here on each call, we could generate them whenever we get notified of a topology event (via a topology watcher) and then clone the hash struct and it's data when we need to make modifications during tablet selection.

That'd reduce the hash calls down to just the session UUID. And even that could be changed to just happen once per query.

go/vt/vtgate/tabletgateway.go
harshit-gangal: I believe that should be the way to move forward by putting behind a flag like `tablet-selection-algorithm`

go/vt/proto/query/query.pb.go
harshit-gangal: Another way is to pass SessionUUID explicity to all the methods.

go/vt/vtgate/tabletgateway.go
```diff
@@ -381,6 +383,109 @@ func (gw *TabletGateway) getStatsAggregator(target *querypb.Target) *TabletStatu
...
+		ch.hashMap[hash] = tabletHealth
+	}
+}
+
+func (ch *ConsistentHash) Sort() {
+	sort.SliceStable(ch.nodes, func(i, j int) bool { return ch.nodes[i] < ch.nodes[j] })
+}
+
+func (ch *ConsistentHash) Get(sessionUUID string) *discovery.TabletHealth {
+	hash := xxhash.Sum64String(sessionUUID)
```
harshit-gangal, line 428: this can be done once on the connection, it is not required on every query.

go/vt/vtgate/tabletgateway.go
harshit-gangal: all should be done as part of health check subscription calls.

PR #11318, opened 2022-09-22
vtgate pflags migration

harshit-gangal:
## Description

This migrates the following tools to pflag:
* `go/cmd/vtgate`
* `go/vt/vtgate/vtgateconn`
* `go/vt/vtgate`

## Related Issue(s)

- Closes https://github.com/vitessio/vitess/issues/11286
- Closes https://github.com/vitessio/vitess/issues/10937
- Closes https://github.com/vitessio/vitess/issues/10931

## Checklist

-   [X] Tests were added or are not required
-   [X] Documentation was added or is not required

Issue #10931, opened 2022-08-04
Switch flag definitions to be on pflag instead of flag in `package go/vt/vtgate`


ajm188:
Part of https://github.com/vitessio/vitess/issues/10697.

Current flags:

```
$ git grep -E "\bflag\.[A-Z]" -- go/vt/vtgate/*.go
go/vt/vtgate/plugin_mysql_server.go:    mysqlServerPort               = flag.Int("mysql_server_port", -1, "If set, also listen for MySQL binary protocol connections on this port.")
go/vt/vtgate/plugin_mysql_server.go:    mysqlServerBindAddress        = flag.String("mysql_server_bind_address", "", "Binds on this address when listening to MySQL binary protocol. Useful to restrict listening to 'localhost' only for instance.")
go/vt/vtgate/plugin_mysql_server.go:    mysqlServerSocketPath         = flag.String("mysql_server_socket_path", "", "This option specifies the Unix socket file to use when listening for local connections. By default it will be empty and it won't listen to a unix socket")
go/vt/vtgate/plugin_mysql_server.go:    mysqlTCPVersion               = flag.String("mysql_tcp_version", "tcp", "Select tcp, tcp4, or tcp6 to control the socket type.")
go/vt/vtgate/plugin_mysql_server.go:    mysqlAuthServerImpl           = flag.String("mysql_auth_server_impl", "static", "Which auth server implementation to use. Options: none, ldap, clientcert, static, vault.")
go/vt/vtgate/plugin_mysql_server.go:    mysqlAllowClearTextWithoutTLS = flag.Bool("mysql_allow_clear_text_without_tls", false, "If set, the server will allow the use of a clear text password over non-SSL connections.")
go/vt/vtgate/plugin_mysql_server.go:    mysqlProxyProtocol            = flag.Bool("proxy_protocol", false, "Enable HAProxy PROXY protocol on MySQL listener socket")
go/vt/vtgate/plugin_mysql_server.go:    mysqlServerRequireSecureTransport = flag.Bool("mysql_server_require_secure_transport", false, "Reject insecure connections but only if mysql_server_ssl_cert and mysql_server_ssl_key are provided")
go/vt/vtgate/plugin_mysql_server.go:    mysqlSslCert = flag.String("mysql_server_ssl_cert", "", "Path to the ssl cert for mysql server plugin SSL")
[...]

PR #11966, opened 2022-12-14
moved missed flags to pflags in vtgate

harshit-gangal:
## Description

This PR moves some remaining flags to pflags in the VTGate.

This would need to backport to 15.0 to be inline with the work carried out in 15.0 regarding the migration.

## Related Issue(s)

- https://github.com/vitessio/vitess/pull/11318
- https://github.com/vitessio/vitess/issues/10931

## Checklist

-   [X] "Backport to:" labels have been added if this change should be back-ported - v15

### DIFF
--- a/go/vt/vtgate/plugin_mysql_server.go
+++ b/go/vt/vtgate/plugin_mysql_server.go
@@ -18,7 +18,6 @@ package vtgate
 
 import (
 	"context"
-	"flag"
 	"fmt"
 	"net"
 	"os"
@@ -29,6 +28,8 @@ import (
 	"sync/atomic"
 	"syscall"
 	"time"
+
+	"github.com/spf13/pflag"
 
 	"vitess.io/vitess/go/vt/sqlparser"
 	"vitess.io/vitess/go/vt/vterrors"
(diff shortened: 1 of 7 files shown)


go/vt/vtgate/endtoend/update/lookup_unique_test.go
```diff
@@ -18,13 +18,14 @@ package endtoend
...
 import (
 	"context"
-	"flag"
 	"fmt"
 	"os"
 	"testing"
 
 	"github.com/stretchr/testify/require"
 
+	_flag "vitess.io/vitess/go/internal/flag"
```
ajm188, line 27: fwiw, you don't need this rename

go/vt/vtgate/plugin_mysql_server.go
```diff
@@ -50,39 +51,56 @@ import (
...
-	mysqlConnReadTimeout  = flag.Duration("mysql_server_read_timeout", 0, "connection read timeout")
-	mysqlConnWriteTimeout = flag.Duration("mysql_server_write_timeout", 0, "connection write timeout")
-	mysqlQueryTimeout     = flag.Duration("mysql_server_query_timeout", 0, "mysql query timeout")
-
-	mysqlConnBufferPooling = flag.Bool("mysql-server-pool-conn-read-buffers", false, "If set, the server will pool incoming connection read buffers")
-
-	mysqlDefaultWorkloadName = flag.String("mysql_default_workload", "OLTP", "Default session workload (OLTP, OLAP, DBA)")
+	mysqlServerPort                   = -1
+	mysqlServerBindAddress            = ""
+	mysqlServerSocketPath             = ""
```
ajm188, line 56: these are the zero value and thus can just be `<var name> string`

ajm188 (approved):
overall LGTM, couple small points

deepthi:
> This would need to backport to 15.0 to be inline with the work carried out in 15.0 regarding the migration.

I don't understand why this needs to be back ported. I see no user-visible change. Can you elaborate?

harshit-gangal:
> > This would need to backport to 15.0 to be inline with the work carried out in 15.0 regarding the migration.
> 
> I don't understand why this needs to be back ported. I see no user-visible change. Can you elaborate?

My understanding is that for Viper implementation, pflags work should be completed in 15.0.
If that is not the case, no need to backport.

systay (approved):

PR #11918, opened 2022-12-08
vttestserver: make tablet_refresh_interval configurable and reduce default value

rsajwani:
Signed-off-by: Rameez Sajwani <rameezwazirali@hotmail.com>

## Description
This is an attempt to make vttestserver faster. Since its used for testing purpose we have seen people finding it slow especially while creating new database. It typically was taking more than 30second (on avg) to create empty database.  This PR fixes the issue by reducing the value of. 

[refreshInterval](https://github.com/vitessio/vitess/blob/main/go/vt/discovery/healthcheck.go#L83)

Even though we could have reduced the value of other flag 

[leaderCheckInterval](https://github.com/vitessio/vitess/blob/main/go/vt/vttablet/tabletserver/gc/tablegc.go#L46)

but we choose not to do that as it is going to be removed in subsequent refactoring PR by @shlomi-noach [details here](https://github.com/vitessio/vitess/pull/11918#discussion_r1049436148)

Reducing these values might put more stress on Topo Server but for testing purpose we should be good. All the remaining details have been captured in the issue #11868  

## Testing

local testing using a script
```
bin/vttestserver \
	--port "33804" \
	--keyspaces "unsharded" \
	--num_shards "1" \
	--mysql_bind_host "127.0.0.1" \
	--charset "utf8mb4" \
	--foreign_key_mode "allow" \
	--enable_online_ddl="true" \
	--enable_direct_ddl="true" \
	--planner-version="v3" \
	--vschema_ddl_authorized_users=% \
	--tablet_refresh_interval "11s"
	--schema_dir="/vtdataroot/vt/schema/"
```
I was able to verify that tablet_refresh_interval is pass all the way to vtcombo

```
[...]

PR #11972, opened 2022-12-15
Table GC: rely on tm state to determine operation mode

shlomi-noach:
## Description

This is a simplification and an improvement to the tablet's table garbage collector operation.

Currently, the collector periodically checks to see whether the tablet is a `PRIMARY`. If yes, it enables the general operation, enables tickers, probes tables etc. If not, it turns off the tickers and stops doing any GC work. But it keeps polling. There is a `leaderCheckInterval = 5 * time.Second` variable, and this is where and when the collector checks for the tablet type.

As of this PR, this logic is simplified, and the variable is removed. The garbage collector only runs on Primary tablets. As such, we just use the fact that the tablet's state manager `Open()`s and `Close()`s the collector when going into and out of Primary type, respectively. There is no more polling/interval. It happens when it needs to happen.

The function `Operate()` now gets called upon `Open()` and terminates upon `Close()` using a `context`. Some variables are removed from the class and made function-local.

Existing `endtoend` tests are left unchanged and validate that the new behavior is good.

## Related Issue(s)

- https://github.com/vitessio/vitess/issues/11986
- Somewhat similar changes to https://github.com/vitessio/vitess/pull/11604
- Some discussion in https://github.com/vitessio/vitess/pull/11918#discussion_r1048832961

## Checklist


deepthi (dismissed):
LGTM!

go/vt/vttablet/tabletserver/gc/tablegc.go
```diff
@@ -513,6 +501,10 @@ func (collector *TableGC) purge(ctx context.Context) (tableName string, err erro
 
 	log.Infof("TableGC: purge begin for %s", tableName)
 	for {
+		if ctx.Err() != nil {
+			// cancelled
+			return
```
deepthi, line 506: Hmm, should the error be returned?

deepthi:
I was debating whether this requires an issue. You have labeled it as an Enhancement, to me it looks like an Internal Cleanup without any user-facing change. As such, we can do without an issue, but we might want to change the label.

deepthi:
The unit test failures are very weird. This PR doesn't do anything that can affect `vtexplain` at all.
If at all I would have expected these failures on #11941 but they seemed to have passed there.

go/vt/vttablet/tabletserver/gc/tablegc.go
shlomi-noach: good catch, fixed!

shlomi-noach:
Fixes https://github.com/vitessio/vitess/issues/11986

shlomi-noach:
@deepthi I both changes the label and created and issue, https://github.com/vitessio/vitess/issues/11986

shlomi-noach:
Unit test failures still taking place, even after merging latest `main`

shlomi-noach:
Unit tests continue to fail for me on a clean checkout on a different box.

deepthi:
> Unit tests continue to fail for me on a clean checkout on a different box.

can you create an issue with details and tag @vitessio/query-serving?
It's quite mysterious because they also seem to be passing on most merges into main.

shlomi-noach:
Created issue at https://github.com/vitessio/vitess/issues/11995

go/vt/vttablet/tabletserver/gc/tablegc.go
```diff
@@ -238,73 +222,81 @@ func (collector *TableGC) Close() {
...
-		defer collector.initMutex.Unlock()
+	dropTablesChan := make(chan string)
+	purgeRequestsChan := make(chan bool)
+	transitionRequestsChan := make(chan *transitionRequest)
 
-		t := timer.NewSuspendableTicker(d, true)
-		collector.tickers = append(collector.tickers, t)
+	tickers := [](*timer.SuspendableTicker){}
+	addTicker := func(d time.Duration) *timer.SuspendableTicker {
+		t := timer.NewSuspendableTicker(d, false)
```
GuptaManan100, line 231: The test failures are definitely because of the changes in this PR. They work on `main`, and from what I can tell they are coming from this specific change, of the ticker not being initially suspended. This causes the `tableCheckTicker` to tick, leading to a run of `collector.checkTables` which executes `sqlShowVtTables`. These show up as extra outputs in the vtexplain tests - `      + 	0 ks_unsharded/-: show full tables like '\_vt\_%'

GuptaManan100: I can confirm, if I change that line to again suspend the ticker initially, then all the tests in `vtexplain` work.

GuptaManan100: `vtexplain` is running a vttablet tabletserver with a call like `tsv := tabletserver.NewTabletServer(topoproto.TabletAliasString(t.Alias), config, memorytopo.NewServer(""), t.Alias)`
I don't think vtexplain needs the garbage collector to be running, so we can just remove it entirely from vtexplain. 
cc - @harshit-gangal WDYT?

shlomi-noach: > so we can just remove it entirely from vtexplain.

@GuptaManan100 thank you! The tablet GC is not configurable to just turn off. How would you remove it from vtexplain? Do we need a special flag, only used in testing, to prevent tablet GC from running?

shlomi-noach: I could add this as a `tabletenv.Config` variable?

shlomi-noach: OK, just did that in https://github.com/vitessio/vitess/pull/11972/commits/398d022cc97b41d6f48b61fc0777c8579c144fba, this consistent with `EnableOnlineDDL` and `EnableTxThrottler` etc.  The tests are now good!

GuptaManan100: Yes this is exactly what I had in mind, something like `EnableOnlineDDL`

deepthi (approved):

go/vt/vttablet/tabletserver/gc/tablegc.go
```diff
@@ -654,8 +632,7 @@ func (collector *TableGC) Status() *Status {
 		Keyspace: collector.keyspace,
 		Shard:    collector.shard,
 
-		isPrimary: (atomic.LoadInt64(&collector.isPrimary) > 0),
-		IsOpen:    (atomic.LoadInt64(&collector.isOpen) > 0),
+		IsOpen: (atomic.LoadInt64(&collector.isOpen) > 0),
```
rsajwani, line 635: Why LoadInt64 here whereas we have removed atomic from other places where we are reading  isOpen ? do we not need synchronization anywhere else ?

go/vt/vttablet/tabletserver/gc/tablegc.go
```diff
@@ -210,101 +194,113 @@ func (collector *TableGC) Open() (err error) {
 		delete(collector.lifecycleStates, schema.EvacTableGCState)
 	}
 
+	ctx := context.Background()
+	ctx, collector.cancelOperation = context.WithCancel(ctx)
+	go collector.operate(ctx)
```
rsajwani, line 199: moving operate from init to Open. Does this means GC will start later then what it is currently now. Do you see any issue because of that delay?

rsajwani (approved):

go/vt/vttablet/tabletserver/gc/tablegc.go
```diff
@@ -210,101 +194,113 @@ func (collector *TableGC) Open() (err error) {
...
-					}
-				}()
-			}
+		case <-ctx.Done():
+			log.Info("TableGC: done operating")
+			return
 		case <-tableCheckTicker.C:
 			{
-				_ = collector.checkTables(ctx)
+				log.Info("TableGC: tableCheckTicker")
```
GuptaManan100, line 256: Won't this log every 5 seconds on a primary tablet?

GuptaManan100: Same comments for the other logs on tickers

GuptaManan100 (approved):

go/vt/vttablet/tabletserver/gc/tablegc.go
shlomi-noach: That's because `isOpen` is only changed in `Open()` and `Close()`, which are both using `stateMutex.Lock()` to exclude each other. So those two functions don't need to `LoadInt64`. But `Status()` is not protected with that same mutex. The tradeoff is that it must read `isOpen` with `LoadInt64` because it might run at the same time `Open()` or `Close()` are running.
An alternative is to add `stateMutex.Lock()` to `Status()`, but I prefer not to.

go/vt/vttablet/tabletserver/gc/tablegc.go
shlomi-noach: I don't see an issue, because `Operate()` works on intervals anyhow. There is no guarantee when exactly the cycles begin. The default cycle interval is `1h`. So the precise timing of when the cycle begins is insignificant.

go/vt/vttablet/tabletserver/gc/tablegc.go
shlomi-noach: > Won't this log every 5 seconds on a primary tablet?

The default interval is `1h`, and I see no reason why in production it would be any less then several minutes. In testing, where we set it to `5s` - yes. But then, do we care?

shlomi-noach: I removed the log entry for `log.Info("TableGC: purgeReentranceTicker")`, which runs once per `1m`. Still very spacious IMO and shouldn't be a problem logging once per minute, but now removed. The rest of logs are either once per hour, or upon something actually being done.

shlomi-noach:
I've made the necessary changes, just waiting to verify that all questions have been satisfied.

shlomi-noach:
all righty, merging!

PR #11978, opened 2022-12-15
Add `VStreamerCount` stat to `vttablet`

timvaillancourt:
## Description

This PR adds the ~~`VStreamersActive`~~ `VStreamerCount` stats metric to `vttablet` to make the # of active streams more observable. There may be a way to infer this using rates from a combination of `VStreamersCreated` and other metrics but I was unable to do so myself (in Prometheus/Grafana) after many attempts so this feels like a rough edge

The length of the VStream engine `streamers` map is used as the number of active streams. This is done without a lock but I'm happy to add one

## Checklist

-   [x] Documentation: https://github.com/vitessio/website/pull/1291

### DIFF
--- a/go/vt/vttablet/tabletserver/vstreamer/engine.go
+++ b/go/vt/vttablet/tabletserver/vstreamer/engine.go
@@ -97,6 +97,7 @@ type Engine struct {
 	rowStreamerNumPackets     *stats.Counter
 	rowStreamerWaits          *servenv.TimingsWrapper
 	errorCounts               *stats.CountersWithSingleLabel
+	vstreamersActive          *stats.GaugeFunc
 	vstreamersCreated         *stats.Counter
 	vstreamersEndedWithErrors *stats.Counter
 	vstreamerFlushedBinlogs   *stats.Counter
@@ -138,6 +139,7 @@ func NewEngine(env tabletenv.Env, ts srvtopo.Server, se *schema.Engine, lagThrot
 		errorCounts:               env.Exporter().NewCountersWithSingleLabel("VStreamerErrors", "Tracks errors in vstreamer", "type", "Catchup", "Copy", "Send", "TablePlan"),
 		vstreamerFlushedBinlogs:   env.Exporter().NewCounter("VStreamerFlushedBinlogs", "Number of times we've successfully executed a FLUSH BINARY LOGS statement when starting a vstream"),
 	}
+	vse.vstreamersActive = env.Exporter().NewGaugeFunc("VStreamersActive", "Count of vstreamers active", vse.getStreamersActive)
 	env.Exporter().NewGaugeFunc("RowStreamerMaxInnoDBTrxHistLen", "", func() int64 { return env.Config().RowStreamer.MaxInnoDBTrxHistLen })
 	env.Exporter().NewGaugeFunc("RowStreamerMaxMySQLReplLagSecs", "", func() int64 { return env.Config().RowStreamer.MaxMySQLReplLagSecs })
 	env.Exporter().HandleFunc("/debug/tablet_vschema", vse.ServeHTTP)
@@ -381,6 +383,10 @@ func (vse *Engine) setWatch() {
 		vse.vschemaUpdates.Add(1)
 		return true
 	})
+}
+
+func (vse *Engine) getStreamersActive() int64 {
+	return int64(len(vse.streamers))
 }
 
 func getPacketSize() int64 {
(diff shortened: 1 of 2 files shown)


mattlord:
Just a note that we'll need to document this new metric here on the vreplication metrics page: https://github.com/vitessio/website/pull/1267

@timvaillancourt we should try and have you push directly to that PR. I can help get you setup with any access you need and walk you through it as needed.

go/vt/vttablet/tabletserver/vstreamer/engine.go
```diff
@@ -383,6 +385,10 @@ func (vse *Engine) setWatch() {
 	})
 }
 
+func (vse *Engine) getStreamersActive() int64 {
+	return int64(len(vse.streamers))
+}
```
mattlord, line 390: Slice length is not concurrency safe so we should use the mutex. Also, IMO a better name:
```
func (vse *Engine) getVStreamerCount() int64 {
	vse.mu.Lock()
	defer vse.mu.Unlock()
 	return int64(len(vse.streamers))
 }
```

mattlord (commented):
Thank you! I also noticed this missing metric info today: https://vitess.slack.com/archives/C0PQY0PTK/p1671132033807179?thread_ts=1670952842.374699&cid=C0PQY0PTK

I had to look at the tablets' `/debug/status` endpoints to see where the vstreamer was running. 

I would note that this is not *only* about vtgate<->vttablet vstreams but also vttablet<->vttablet vstreams such as are used for VReplication workflows after the copy phase. Just something to keep in mind as we think about this.

go/vt/vttablet/tabletserver/vstreamer/engine.go
timvaillancourt: @mattlord sounds good 👍

I might be worrying about introducing a lock too much but could I hypothetically use `atomic.AddUint64()`/`atomic.LoadUint64()` instead and incr/decr it from `.Stream()` instead and is there a preference?

timvaillancourt:
@mattlord thanks for the review! One question re: the mutex added and website PR on the way tomorrow 👍

timvaillancourt:
Hrm, re-requesting review from Matt dropped the other reviewers. My apologies 🤷

go/vt/vttablet/tabletserver/vstreamer/engine.go
mattlord: I don't think that's needed here, because the engine doesn't change THAT often (not typically starting/ending hundreds or even thousands of vstreams on a tablet in a second, or even minute). @rohit-nayak-ps, what do you think?

mattlord (approved):
LGTM! ❤️

timvaillancourt:
> @timvaillancourt we should try and have you push directly to that PR. I can help get you setup with any access you need and walk you through it as needed.

@mattlord thanks! Website PR created here: https://github.com/vitessio/website/pull/1291

go/vt/vttablet/tabletserver/vstreamer/engine.go
```diff
@@ -138,6 +138,7 @@ func NewEngine(env tabletenv.Env, ts srvtopo.Server, se *schema.Engine, lagThrot
 		errorCounts:               env.Exporter().NewCountersWithSingleLabel("VStreamerErrors", "Tracks errors in vstreamer", "type", "Catchup", "Copy", "Send", "TablePlan"),
 		vstreamerFlushedBinlogs:   env.Exporter().NewCounter("VStreamerFlushedBinlogs", "Number of times we've successfully executed a FLUSH BINARY LOGS statement when starting a vstream"),
 	}
+	env.Exporter().NewGaugeFunc("VStreamerCount", "Current number of vstreamers", vse.getVStreamerCount)
```
rohit-nayak-ps, line 141: Can we instead use a `stats.Gauge` as a new field `vstreamerCount` in `vstreamer.Engine`? Then in `vse.Stream()`: `vse.vstreamerCount.Add(1)` while creating a stream and `vse.vstreamerCount.Add(-1)`. `stats.Gauge` is an `atomicInt64`. We then avoid the mutex requirement.

timvaillancourt: @rohit-nayak-ps good idea, I'll make that change 👍

timvaillancourt: @rohit-nayak-ps / @mattlord changes made in https://github.com/vitessio/vitess/pull/11978/commits/5d889e0f2ca1a71303052171daa712e7b6ba271f

mattlord (approved):
LGTM (again)! 😄

rohit-nayak-ps (approved):
lgtm

Thanks! this is a useful new metric.

Issue #11977, opened 2022-12-15
Detailed backup stats


maxenglander:
# Description

I would like to have better instrumentation on backups, in particular backups generated by `vtbackup`. While backup stats are exposed via `servenv` since https://github.com/vitessio/vitess/pull/11388 (which is great!) ideally I would like more fine-grained stats on:

 * Download last backup.
 * Decompress last backup.
 * Catchup to primary with replication.
 * Compress new backup.
 * Upload new backup.

# Use case

I want to make backups (and restores) as fast as possible. Having more fine-grained stats will guide me to put energies in the right place.

maxenglander:
There are other backup plugins that would benefit from detailed stats, but think this is far enough along that we can close this out.

PR #11388, opened 2022-09-28
expose vtbackup stats at --port /metrics

maxenglander:
## Description

As far as I can tell `vtbackup` does not currently expose metrics. It would be awesome if it did. In particular it would be great to have the following timings:

 * How long it takes to download the last backup.
 * How long it takes to start and stop MySQL (MySQL startup can be slow sometimes, e.g. during InnoDB initialization).
 * How long it takes to connect to the primary.
 * How long it takes to download the binary log.
 * How long it takes to apply the binlog.
 * How long it takes to perform and upload the new backup.

This PR modifies `vtbackup` command so that a server is started on `--port`. The server is similar to that launched by other VT components, and includes a `/metrics` route. With this PR this route include two useful metrics which are managed by the `mysqlctl` package: 

 * `vtbackup_restore_duration_seconds`
 * `vtbackup_backup_duration_seconds`

Keeping this PR small to lay the groundwork. The other metrics can come in a later PR if we're OK with this overall approach.

### Use cases

PlanetScale makes ongoing internal and public efforts to improve backup and restore performance. It would be great to have detailed metrics on current performance so that we can make informed decisions on where to put our energy.

PR #11979, opened 2022-12-16
vtbackup, mysqlctl: detailed backup and restore metrics

maxenglander:
# Description

Addresses #11977.

I would like to have better instrumentation on backups, in particular backups generated by `vtbackup`. While backup stats are exposed via `servenv` since https://github.com/vitessio/vitess/pull/11388 (which is great!) ideally I would like more fine-grained stats on:

 * Download last backup.
 * Decompress last backup.
 * Catchup to primary with replication.
 * Compress new backup.
 * Upload new backup.

# Design

There were a couple design goals that shaped the way the code is written.

## No breaking interface changes

_After discussion with Deepthi we decided to introduce a breaking change after all._

~A quick Google/GitHub search didn't reveal any out-of-tree `backupstorage` plugins, but the way the backup/restore APIs are laid out makes it seem like they were designed to support out-of-tree plugins.~

~I tried to keep that in mind when writing this PR, in particular by not making any changes that would require anyone using out-of-tree plugins to make code changes when they upgrade to a Vitess version with these changes.~

## Separate policy and mechanism

It wouldn't be great if every out-of-tree `backupstorage` generated metrics in ways that conflicted with each other or varied widely from way Vitess users are used to consuming in-tree metrics.

In this PR, I tried to create a minimal stats mechanism that can be used by in-tree and out-of-tree code, but where the policies for stats (metric names and labels, stats sink, etc.) are kept in-tree and under the control of the Vitess user.

This approach seems similar in spirit to what is already being done with `BackupParams.Logger` and `RestoreParams.Logger`.

# Changes

This PR adds several new metrics:

 * `vtbackup_duration_by_phase_seconds` with `phase` label
 * `{vtbackup,vttablet}_backup_bytes` with `component`, `implementation`, and `operation` labels
 * `{vtbackup,vttablet}_backup_count` with `component`, `implementation`, and `operation` labels
 * `{vtbackup,vttablet}_backup_duration_nanoseconds` with `component`, `implementation`, and `operation` labels
 * `{vtbackup,vttablet}_restore_bytes` with `component`, `implementation`, and `operation` labels
 * `{vtbackup,vttablet}_restore_count` with `component`, `implementation`, and `operation` labels
 * `{vtbackup,vttablet}_restore_duration_nanoseconds` with `component`, `implementation`, and `operation` labels

It also deprecates these older backup/restore metrics:

 * `{vtbackup,vttablet}_backup_duration_seconds`
 * `{vtbackup,vttablet}_restore_duration_seconds`

# Notes

Changes to `vtbackup`:

 * Add new stat `duration_seconds` metric which reports durations of additional phases not covered by `mysqlctl`.
 * Reports phases like `initmysqld`, `initialbackup`, `restorelastbackup`, `catchupreplication`, etc.

Changes to `mysqlctl`:

 * Add `backup_bytes`, `backup_count`, `backup_duration_nanoseconds`, `restore_bytes`, `restore_count`, `restore_duration_nanoseconds` metrics.
 * These metrics each have three labels: `component`, `implementation`, and `operation`.
 * The idea is that we can re-use these metric across different components (`-` = unscoped, top-level, `backupstorage`, `backupengine`, etc.) across different implementations (`s3`, `file`, etc.), and across different operations (`backup`, `restore`, `read`, `compress`, `encrypt`).

Other notes:

 * An alternative design would be to just have each engine/plugin declare its own stats, or have a set of global stats that each engine/plugin/package can import.
 * Why `nanoseconds` on those two new metrics? Because as we're reporting on read/write times for individual files, if all the files are small and take less than a second to process then they end up reporting as zero.
 * ~Why `backupengine.(Parameterizable)`? I wasn't sure how safe it would be to break any of the APIs like `BackupEngine` and `BackupStorage`. Figured it was better to introduce changes this way until I get some guidance.~

# Samples

Sample metrics generated by running `vtbackup` from this branch against the [local example](https://github.com/vitessio/vitess/tree/main/examples/local) cluster. Processed with `jq` and sorted for readability.

```
{
  "BackupBytes": {
    "BackupEngine.Builtin.Source:Read": 4777,
    "BackupEngine.Builtin.Compressor:Write": 4616,
    "BackupEngine.Builtin.Destination:Write": 162,
    "BackupStorage.File.File:Write": 163
  },
  "BackupCount": {
    "-.-.Backup": 1,
    "BackupEngine.Builtin.Source:Open": 161,
    "BackupEngine.Builtin.Source:Close": 322,
    "BackupEngine.Builtin.Compressor:Close": 161,
    "BackupEngine.Builtin.Destination:Open": 161,
    "BackupEngine.Builtin.Destination:Close": 322
  },
  "BackupDurationNanoseconds": {
    "-.-.Backup": 4188508542,
    "BackupEngine.Builtin.Source:Open": 10649832,
    "BackupEngine.Builtin.Source:Read": 55901067,
    "BackupEngine.Builtin.Source:Close": 960826,
    "BackupEngine.Builtin.Compressor:Write": 278358826,
    "BackupEngine.Builtin.Compressor:Close": 79358372,
    "BackupEngine.Builtin.Destination:Open": 16456627,
    "BackupEngine.Builtin.Destination:Write": 11021043,
    "BackupEngine.Builtin.Destination:Close": 17144630,
    "BackupStorage.File.File:Write": 10743169
  },
  "DurationByPhaseSeconds": {
    "InitMySQLd": 2,
    "RestoreLastBackup": 6,
    "CatchUpReplication": 1,
    "TakeNewBackup": 4
  },
  "RestoreBytes": {
    "BackupEngine.Builtin.Source:Read": 1095,
    "BackupEngine.Builtin.Decompressor:Read": 950,
    "BackupEngine.Builtin.Destination:Write": 209,
    "BackupStorage.File.File:Read": 1113
  },
  "RestoreCount": {
    "-.-.Restore": 1,
    "BackupEngine.Builtin.Source:Open": 161,
    "BackupEngine.Builtin.Source:Close": 322,
    "BackupEngine.Builtin.Decompressor:Close": 161,
    "BackupEngine.Builtin.Destination:Open": 161,
    "BackupEngine.Builtin.Destination:Close": 322
  },
  "RestoreDurationNanoseconds": {
    "-.-.Restore": 6204765541,
    "BackupEngine.Builtin.Source:Open": 10542539,
    "BackupEngine.Builtin.Source:Read": 104658370,
    "BackupEngine.Builtin.Source:Close": 773038,
    "BackupEngine.Builtin.Decompressor:Read": 165692120,
    "BackupEngine.Builtin.Decompressor:Close": 51040,
    "BackupEngine.Builtin.Destination:Open": 22715122,
    "BackupEngine.Builtin.Destination:Write": 41679581,
    "BackupEngine.Builtin.Destination:Close": 26954624,
    "BackupStorage.File.File:Read": 102416075
  },
  "backup_duration_seconds": 4,
  "restore_duration_seconds": 6
}
```

# Performance

At @deepthi suggestion I compared performance of backups on `main` versus this branch.

I set up the commerce example cluster, and created a table with ~20 GiB of data, then ran:

```
$ vtctlclient BackupShard -- --allow_primary --concurrency=1 commerce/0
```

Multiple times with this branch and `main`, comparing the values of `vttablet_backup_duration_seconds`.

| # | `main` branch (7fc1b48d9) | this branch (d5364e6) |
|-|-|-|
| 1 | 21 | 21 |
| 2 | 20 | 23 |
| 3 | 22 |  21 | 
| 4 | 21 | 22 |
| 5 | 21 | 22 | 

Assuming the differences aren't due to vagaries of CPU and disk usage on my Mac M1, this branch adds a roughly 3.8% performance overhead.

# Out-of-scope

This PR doesn't add new metrics to all backup engines or storage engines. Would like to get buy-in on this approach (or a different one) first,  and then expand whatever approach we adopt in follow-on PRs to cover additional backup engines & storages.

# Checklist

 * [x] Implement.
 * [x] Write tests.
 * [x] [Docs](https://github.com/vitessio/website/pull/1364)


go/cmd/vtbackup/vtbackup.go
```diff
@@ -120,6 +121,11 @@ var (
 	detachedMode     bool
 	keepAliveTimeout = 0 * time.Second
 	disableRedoLog   = false
+	durationByPhase  = stats.NewGaugesWithSingleLabel(
+		"duration_seconds",
+		"How long it took vtbackup to perform a each phase of operation (in seconds).",
```
maxenglander, line 126: Because vtbackup also picks up stats from `mysqlctl`, there will be some overlap between this metric and those metrics. I think that's OK, personally, but if we want I can take more care to avoid any overlap.

go/stats/timings.go
```diff
@@ -180,7 +180,18 @@ func (t *Timings) Label() string {
 	return t.label
 }
 
-var bucketCutoffs = []int64{5e5, 1e6, 5e6, 1e7, 5e7, 1e8, 5e8, 1e9, 5e9, 1e10}
+var bucketCutoffs = []int64{
```
maxenglander, line 183: This isn't needed. Can revert although I kind of like it.

maxenglander: I guess these aren't strictly seconds...it depends on the unit of the metric. Will revert.

go/vt/mysqlctl/backup.go
```diff
@@ -92,9 +92,6 @@ var (
 	// backupCompressBlocks is the number of blocks that are processed
 	// once before the writer blocks
 	backupCompressBlocks = 2
-
-	backupDuration  = stats.NewGauge("backup_duration_seconds", "How long it took to complete the last backup operation (in seconds)")
```
maxenglander, line 96: Not deleted, just moved to `backupstats` package.

go/vt/mysqlctl/backup.go
```diff
@@ -116,6 +113,10 @@ func registerBackupFlags(fs *pflag.FlagSet) {
 // - shuts down Mysqld during the backup
 // - remember if we were replicating, restore the exact same state
 func Backup(ctx context.Context, params BackupParams) error {
+	if params.Stats == nil {
+		params.Stats = backupstats.DefaultBackupStats()
```
maxenglander, line 117: The whole approach of this PR is to bundle up stats in an interface that can be passed downwards into deeper parts of the backup system, rather than having different engines and plugins maintain their own stats or import a set of global stats.

Having `Stats` be part of `BackupParams` means that vtbackup could pass in different metrics than other packages that use `mysqlctl`. Also will be nice because we can pass in mock stats to make testing easy.

go/vt/mysqlctl/backup.go
```diff
@@ -125,6 +126,20 @@ func Backup(ctx context.Context, params BackupParams) error {
 		return vterrors.Wrap(err, "unable to get backup storage")
 	}
 	defer bs.Close()
+
+	params.Logger.Infof("Checking if we should parameterize backup storage %s?", backupstorage.BackupStorageImplementation)
+	if p, ok := bs.(backupstorage.Parameterizer); ok {
```
maxenglander, line 131: I didn't have to do this. I could have added new params or methods to the `BackupStorage` or `FileHandle` interfaces. But I wasn't sure how careful we need to be about making changes to those APIs, so figured best to do things in a non-breaking way until I get some guidance.

go/vt/mysqlctl/backup.go
```diff
@@ -125,6 +126,20 @@ func Backup(ctx context.Context, params BackupParams) error {
 		return vterrors.Wrap(err, "unable to get backup storage")
 	}
 	defer bs.Close()
+
+	params.Logger.Infof("Checking if we should parameterize backup storage %s?", backupstorage.BackupStorageImplementation)
```
maxenglander, line 130: TODO Delete

go/vt/mysqlctl/backup.go
```diff
@@ -125,6 +126,20 @@ func Backup(ctx context.Context, params BackupParams) error {
...
+	if p, ok := bs.(backupstorage.Parameterizer); ok {
+		params.Logger.Infof("Yes we should.")
```
maxenglander, line 132: TODO delete

go/vt/mysqlctl/backup.go
```diff
@@ -134,9 +149,13 @@ func Backup(ctx context.Context, params BackupParams) error {
 	if err != nil {
 		return vterrors.Wrap(err, "failed to find backup engine")
 	}
-
 	// Take the backup, and either AbortBackup or EndBackup.
-	usable, err := be.ExecuteBackup(ctx, params, bh)
+	beParams := params.Copy()
```
maxenglander, line 153: This is a bit awkward, and copying structs makes me uncomfortable. Is there a way to lint for exhaustive field copying?

go/vt/mysqlctl/backupstats/defaults.go
```diff
@@ -0,0 +1,52 @@
...
+	workLabels = []string{"Component", "Implementation", "Work"}
+
+	backupDuration       *stats.Gauge
+	backupWorkDurations  *stats.CountersWithMultiLabels
+	restoreDuration      *stats.Gauge
+	restoreWorkDurations *stats.CountersWithMultiLabels
+)
+
+func init() {
+	backupDuration = stats.NewGauge(
```
maxenglander, line 17: These stats get reported no matter what, which doesn't quite line up with the way the rest of this PR is meant to works. Rethinnk this.

go/vt/mysqlctl/builtinbackupengine.go
```diff
@@ -702,15 +705,36 @@ func (be *BuiltinBackupEngine) backupFile(ctx context.Context, params BackupPara
...
 		} else {
-			compressor, err = newBuiltinCompressor(CompressionEngineName, writer, params.Logger)
+			compressor, err = newBuiltinCompressor(CompressionEngineName, timedWriter, params.Logger)
 		}
 		if err != nil {
 			return vterrors.Wrap(err, "can't create compressor")
 		}
-		writer = compressor
+
+		// The compression duration includes the time it takes to write
```
maxenglander, line 721: I didn't have to do this. We could just report compression time without trying to factor out that because of the way we're chaining `io.Writer` together that that time includes disk writing (or S3 uploading) time. In other words we could leave it as an exercise for the Grafana use to subtract one metric from another to make meaning out of them.

I prefer this, but don't feel strongly about it.

go/vt/mysqlctl/builtinbackupengine.go
```diff
@@ -913,18 +937,19 @@ func (be *BuiltinBackupEngine) restoreFile(ctx context.Context, params RestorePa
 		}
 	}()
 
+	dst := bufio.NewWriterSize(dstFile, writerBufferSize)
+
+	var reader io.Reader
 	bp := newBackupReader(name, 0, source)
 	go bp.ReportProgress(builtinBackupProgress, params.Logger)
-
```
maxenglander, line 918: Not sure if all these changes are necessary. TODO.

go/vt/mysqlctl/builtinbackupengine_test.go
```diff
@@ -100,7 +100,7 @@ func TestExecuteBackup(t *testing.T) {
 	oldDeadline := setBuiltinBackupMysqldDeadline(time.Second)
 	defer setBuiltinBackupMysqldDeadline(oldDeadline)
 
-	bh := filebackupstorage.FileBackupHandle{}
+	bh := filebackupstorage.NewBackupHandle(nil, "", "", false)
```
maxenglander, line 103: Not strictly necessary, but makes a bit easier to avoid nil dereferences.

go/vt/mysqlctl/compression.go
```diff
@@ -251,7 +251,6 @@ func newBuiltinCompressor(engine string, writer io.Writer, logger logutil.Logger
 		compressor = zst
 	default:
 		err = fmt.Errorf("%w value: %q", errUnsupportedCompressionEngine, engine)
-		return compressor, err
```
maxenglander, line 254: TODO revert

go/vt/mysqlctl/filebackupstorage/file.go
```diff
@@ -49,79 +53,112 @@ func init() {
 	servenv.OnParseFor("vttablet", registerFlags)
 }
 
-// FileBackupHandle implements BackupHandle for local file system.
-type FileBackupHandle struct {
-	fbs      *FileBackupStorage
+// fileBackupHandle implements BackupHandle for local file system.
+type fileBackupHandle struct {
```
maxenglander, line 57: I like this but will revert it so that I'm being consistent about not breaking any APIs.

go/vt/mysqlctl/filebackupstorage/file_test.go
```diff
@@ -31,9 +33,9 @@ import (
 
 // setupFileBackupStorage creates a temporary directory, and
 // returns a FileBackupStorage based on it
-func setupFileBackupStorage(t *testing.T) *FileBackupStorage {
+func setupFileBackupStorage(t *testing.T) backupstorage.BackupStorage {
```
maxenglander, line 36: Revert this too to avoid any API breakages

go/vt/mysqlctl/backupengine.go
```diff
@@ -141,14 +182,18 @@ func registerBackupEngineFlags(fs *pflag.FlagSet) {
 //
 // This must only be called after flags have been parsed.
 func GetBackupEngine() (BackupEngine, error) {
-	name := backupEngineImplementation
+	name := GetBackupEngineImplementation()
```
maxenglander, line 185: Don't think I need this. Revert.

go/vt/mysqlctl/backupstats/component.go
```diff
@@ -0,0 +1,36 @@
+/*
```
maxenglander, line 1: TODO document all the new interfaces and structs. Add a `doc.go`.

go/ioutil/reader.go
```diff
@@ -0,0 +1,82 @@
+/*
```
maxenglander, line 1: TODO document all the new interfaces and structs. Add a `doc.go`.

go/vt/mysqlctl/backupstats/stats.go
```diff
@@ -0,0 +1,93 @@
...
+	return newScopedStats(uss.nsCountersWithLabels, nil, scopes...)
+}
+
+func newScopedStats(nsCountersWithLabels *stats.CountersWithMultiLabels, labels []string, scopes ...Scope) Stats {
+	if labels == nil {
+		labels = make([]string, 3)
+	}
+
+	for _, scope := range scopes {
+		labels[scope.Type] = scope.Value
```
maxenglander, line 74: TODO add an out-of-bounds check. Probably fine to just log a warning and not panic, since the worst that can happen is screwed up metrics.  Alternately can return an `error` from the function and let that bubble up.

go/vt/mysqlctl/backupstats/stats.go
```diff
@@ -0,0 +1,93 @@
...
+	copy(labels, ss.labels)
+	return newScopedStats(ss.nsCountersWithLabels, labels, scopes...)
+}
+
+func (ss *scopedStats) AddDuration(d time.Duration) {
+	ss.nsCountersWithLabels.Add(ss.labels, int64(d.Nanoseconds()))
+}
+
+func (ss *scopedStats) SetDuration(d time.Duration) {
+	ss.nsCountersWithLabels.Reset(ss.labels)
```
maxenglander, line 91: This is gross and also this function is unused in the current code. Maybe better to panic or return an error?

go/vt/vttablet/tabletmanager/restore.go
```diff
@@ -191,6 +192,7 @@ func (tm *TabletManager) restoreDataLocked(ctx context.Context, logger logutil.L
 		Shard:               tablet.Shard,
 		StartTime:           logutil.ProtoToTime(request.BackupTime),
 		DryRun:              request.DryRun,
+		Stats:               backupstats.RestoreStats(),
```
maxenglander, line 195: If `Stats` is not set then `mysqlctl` will use `backupstats.NopStats`. This way any out-of-tree code can opt in to the new stats, or not (and not have to worry about nil deference errors).

go/vt/mysqlctl/backupstats/stats.go
```diff
@@ -0,0 +1,147 @@
...
+			"How much time has been spent on restore operations (in nanoseconds).",
+			labels,
+		)
+	})
+	return newScopedStats(restoreCount, restoreDurationNs, nil)
+}
+
+// NopStats creates a no-op Stats suitable for tests and for backwards
+// compoatibility.
+func NopStats() Stats {
```
maxenglander, line 114: Is `NoStats` a better. name?

go/vt/mysqlctl/backupstorage/params.go
```diff
@@ -0,0 +1,47 @@
...
+
+// Parameterizer gives BackupStorage components a way to opt in to receiving
+// Params without breaking the current API.
+type Parameterizer interface {
+	WithParams(Params) BackupStorage
+}
+
+// NopParams gives BackupStorage components way to log and generate stats
+// without doing nil checking.
+func NopParams() Params {
```
maxenglander, line 42: Is `EmptyParams` or `NoParams` a better name?

go/test/fuzzing/tablet_manager_fuzzer.go
```diff
@@ -22,7 +22,7 @@ import (
 	"vitess.io/vitess/go/mysql/fakesqldb"
 	"vitess.io/vitess/go/sqltypes"
 	"vitess.io/vitess/go/vt/dbconfigs"
-	"vitess.io/vitess/go/vt/mysqlctl/fakemysqldaemon"
+	"vitess.io/vitess/go/vt/mysqlctl"
```
maxenglander, line 25: Moved `FakeMysqlDaemon` into `mysql` package so that I can use it in non-blackbox tests of `Backup` and `Restore`.

go/vt/mysqlctl/backupstats/stats.go
```diff
@@ -0,0 +1,166 @@
...
+		labelValues = make([]string, len(durationNs.Labels()))
+		for i := 0; i < len(labelValues); i++ {
+			labelValues[i] = unscoped
+		}
+	}
+
+	return &scopedStats{count, durationNs, labelValues}
+}
+
+func (s *scopedStats) Scope(scopes ...Scope) Stats {
```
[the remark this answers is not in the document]
rsajwani: i am bit confused how is index() working over here. at the will copyOfLabelValues > 3 in any case ?

go/vt/mysqlctl/backup.go
```diff
@@ -125,6 +127,19 @@ func Backup(ctx context.Context, params BackupParams) error {
 		return vterrors.Wrap(err, "unable to get backup storage")
 	}
 	defer bs.Close()
+
```
rsajwani, line 130: why not define scope here 
scope := Stats.Scope(
			stats.Component(stats.BackupStorage),
			stats.Implementation(backupstorage.BackupStorageImplementation))
and use it #133 and #154

go/vt/mysqlctl/backup.go
```diff
@@ -134,9 +149,14 @@ func Backup(ctx context.Context, params BackupParams) error {
...
-
+	// Scope stats to selected backup engine.
+	beParams := params.Copy()
+	beParams.Stats = params.Stats.Scope(
+		stats.Component(stats.BackupEngine),
+		stats.Implementation(backupEngineImplementation),
+	)
 	// Take the backup, and either AbortBackup or EndBackup.
-	usable, err := be.ExecuteBackup(ctx, params, bh)
+	usable, err := be.ExecuteBackup(ctx, beParams, bh)
```
rsajwani, line 159: why copy , why not pass reference of &param

go/vt/mysqlctl/backup.go
```diff
@@ -156,7 +176,8 @@ func Backup(ctx context.Context, params BackupParams) error {
 	}
 
 	// The backup worked, so just return the finish error, if any.
-	backupDuration.Set(int64(time.Since(startTs).Seconds()))
+	stats.DeprecatedBackupDurationS.Set(int64(time.Since(startTs).Seconds()))
+	params.Stats.Scope(stats.Operation("backup")).TimedIncrement(time.Since(startTs))
```
rsajwani, line 180: .scope return stats, do we not need to assigned it back ??

go/ioutil/writer.go
```diff
@@ -0,0 +1,98 @@
...
+
+// Duration reports the total time spend on Write calls so far.
+func (twc *timedWriteCloser) Duration() time.Duration {
+	return twc.total
+}
+
+// Write calls the inner WriteCloser, increments the total Duration, and calls
+// any registered callbacks with the amount of time spent on this Write call.
+func (twc *timedWriteCloser) Write(p []byte) (n int, err error) {
+	t := time.Now()
```
rsajwani, line 65: if implementation of writecloser and writer is same then should we create a helper method and call the helper at both places. that way all future changes are done at one place. Easy to manage.

maxenglander:
Hey @vmg I updated the benchmarks following your suggestions.

 * I downloaded [this Wikipedia dataset](https://dumps.wikimedia.org/enwiki/20221220/enwiki-20221220-externallinks.sql.gz) (contains mostly URLs), loaded it into MySQL 8, and uploaded the InnoDB (`.ibd`) file [here on archive.org](https://archive.org/details/enwiki-20221220-externallinks.ibd).
 * I updated the benchmark suite to download and compress that file. No longer printing out any usage info. Users may override the URL, but this isn't advertised, and is mainly intended for debugging purposes.
 * Added a `SetBytes` metric.
 * Reporting `compression-ratio/op` metric. Chose this awkward name because the `ReportMetric` docs say that units end in `/op` by convention. Happy to rename this based on your advice though.

Not sure if it's a good idea to [host this file on archive.org](https://archive.org/details/enwiki-20221220-externallinks.ibd). Maybe @rsajwani will have thoughts on a better place to host this file.

go/ioutil/reader.go
```diff
@@ -0,0 +1,98 @@
...
+)
+
+// TimedReadCloser tracks how much time is spent on Read calls.
+type TimedReadCloser interface {
+	io.ReadCloser
+	// Duration reports the total duration of time spent on Read calls.
+	Duration() time.Duration
+}
+
+// TimedReader tracks how much time is spent on Read calls.
```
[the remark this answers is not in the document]
maxenglander: It's slightly different. One is for `TimedReader`, the other is for `TimedReadCloser`.

go/vt/mysqlctl/backup.go
maxenglander: Hey @rsajwani the `stats` objects used on L133 and L154 have different scopes, so we cannot re-use them.

go/vt/mysqlctl/backup.go
maxenglander: The `stats` object passed to `be.ExecuteBackup` has a different scope than the one passed in to this function.

go/vt/mysqlctl/backup.go
maxenglander: Assign it back to `params`?

go/vt/mysqlctl/backupstats/stats.go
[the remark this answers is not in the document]
maxenglander: I'm going to change this to be less confusing. Basically the idea is that `Type` should be (or have) a number that is `>= 0` and  `< len(labels)`.

maxenglander:
Thanks for the review @rsajwani I made some code changes based, and left questions/comments on your other feedback. Ready for another review.

go/ioutil/timer.go
```diff
@@ -0,0 +1,59 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+/*
+This file contains the meter struct, which contains contains
```
rsajwani, line 18: vvv minor: contains contains

go/vt/mysqlctl/backup.go
rsajwani: ah missed that .. go it ..

go/vt/mysqlctl/backup.go
rsajwani: yeah

rsajwani: actually I re-read the code.. we are ok here..

go/ioutil/timer.go
maxenglander: Fixed!

go/vt/mysqlctl/backupstats/stats.go
```diff
@@ -0,0 +1,199 @@
...
+
+import (
+	"sync"
+	"time"
+
+	"vitess.io/vitess/go/stats"
+	vtstats "vitess.io/vitess/go/stats"
+)
+
+// Stats is a reporting interface meant to be shared amount backup and restore
```
deepthi, line 27:
```suggestion
// Stats is a reporting interface meant to be shared among backup and restore
```

go/vt/mysqlctl/backupstats/stats.go
```diff
@@ -0,0 +1,199 @@
...
+// components.
+//
+// This interface is meant to give those components a way to reports stats
```
deepthi, line 30:
```suggestion
// This interface is meant to give those components a way to report stats
```

go/vt/mysqlctl/backupstats/stats.go
```diff
@@ -0,0 +1,199 @@
...
+	backupCount       *stats.CountersWithMultiLabels
+	backupDurationNs  *stats.CountersWithMultiLabels
+	restoreBytes      *stats.CountersWithMultiLabels
+	restoreCount      *stats.CountersWithMultiLabels
+	restoreDurationNs *stats.CountersWithMultiLabels
+)
+
+// BackupStats creates a new Stats for backup operations.
+//
+// It registers two stats with the Vitess stats package.
```
deepthi, line 81:
```suggestion
// It registers three stats with the Vitess stats package.
```

deepthi: Or just say "the following stats", then the comment won't become incorrect when we add the next metric.

deepthi: stats vs metrics? Metrics is probably more correct.

go/vt/mysqlctl/backupstats/stats.go
```diff
@@ -0,0 +1,199 @@
...
+			"How much time has been spent on backup operations (in nanoseconds).",
+			labels,
+		)
+	})
+	return newScopedStats(backupBytes, backupCount, backupDurationNs, nil)
+}
+
+// RestoreStats creates a new Stats for restore operations.
+//
+// It registers two stats with the Vitess stats package.
```
deepthi, line 112: Same comment as for `BackupStats`.

deepthi (commented):
This is great work. These metrics have been missing forever.
Can you please create a website PR as well?
Once feedback is addressed we can merge this and you can carry on with adding metrics to other implementations.
Apologies for how long this review took.

deepthi:
One more thing.. Can you run a decent size backup with and without these changes and record any variation in time taken? I'm wondering if the reader/writer wrappers add any overhead.

go/vt/mysqlctl/backupstorage/params.go
```diff
@@ -0,0 +1,47 @@
...
+type Params struct {
+	Logger logutil.Logger
+	Stats  backupstats.Stats
+}
+
+// Parameterizer gives BackupStorage components a way to opt in to receiving
+// Params without breaking the current API.
+type Parameterizer interface {
+	WithParams(Params) BackupStorage
+}
```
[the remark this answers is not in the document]
maxenglander: > What happens when we add more params?

Hm I'm not sure. I think if we add new fields to the `Params` struct it shouldn't cause any compiler errors? Renaming and removing params would cause errors of course. 

We wouldn't be able to change the `Parameterizer` interface, which makes it kind of brittle.

> We have to weigh whether it is better to allow them to continue without metrics, or to force them to implement metrics. I'd say we force them to implement metrics

If we are OK forcing people to update their out-of-tree plugins if/when this is released I am fine with that too!

I guess the question then becomes, if we force the change now, does it make sense to add `WithParams` to the `BackupStorage` interface, or do something else instead? 

To me using `WithParams` feels a bit awkward if are making a breaking change. 

Here are some thoughts about other options:

### Add `Params` to  `BackupStorage` interface

This isn't elegant, but it's straight-forward. 

Biggest downside in my mind is it will require the most code changes to backup storage implementations to pass `Params` downwards to `BackupHandle`, but that can be done in multiple PRs to make it manageable.

Changes would look like this:

```
type BackupStorage interface {
        - ListBackups(ctx context.Context, dir string) ([]BackupHandle, error)
	+ ListBackups(ctx context.Context, params Params, dir string) ([]BackupHandle, error)
	- StartBackup(ctx context.Context, dir, name string) (BackupHandle, error)
	+ StartBackup(ctx context.Context, params Params, dir, name string) (BackupHandle, error)
	- RemoveBackup(ctx context.Context, dir, name string) error
	+ RemoveBackup(ctx context.Context, params Params, dir, name string) error
	- Close() error
	+ Close(params Params) error
}
```

### Change `Context` to `BackupStorageContext`

Introduce a new object `BackupContext` that extends `Context`. Same as last option, just different aesthetic appeal.

```
+ type BackupStorageContext interface {
+        context.Context
+        Logger() logutil.Logger
+        Stats() backupstats.Stats
+ }
+
type BackupStorage interface {
        - ListBackups(ctx context.Context, dir string) ([]BackupHandle, error)
	+ ListBackups(ctx BackupStorageContext, dir string) ([]BackupHandle, error)
	- StartBackup(ctx context.Context, dir, name string) (BackupHandle, error)
	+ StartBackup(ctx BackupStorageContext, dir, name string) (BackupHandle, error)
	- RemoveBackup(ctx context.Context, dir, name string) error
	+ RemoveBackup(ctx BackupStorageContext, dir, name string) error
	- Close() error
	+ Close(ctx BackupStorageContext) error
}
```

### Introduce `BackupStorageFactory`

Change `map[string]BackupStorage` to `map[string]BackupStorageFactory`.

```
+type BackupStorageFactory = func(Params) BackupStorage
+
- var BackupStorageMap = make(map[string]BackupStorage)
+ var BackupStorageMap = make(map[string]BackupStorageFactory)
```

Add factories to each backup storage implementation. I do like that doing this attaches `Params` to `BackupStorage` and should mean less work to make `Params` available to `BackupHandle`, but overall I think the little amount of effort we save doesn't justify this kind of fancy approach.

maxenglander:
>  Can you run a decent size backup with and without these changes and record any variation in time taken?

Good idea.

I set up the commerce example cluster, and created a table with ~20 GiB of data, then ran:

```
$ vtctlclient BackupShard -- --allow_primary --concurrency=1 commerce/0
```

Multiple times with this branch and `main`, comparing the values of `vttablet_backup_duration_seconds`.

| # | `main` branch (7fc1b48d9) | this branch (d5364e6) |
|-|-|-|
| 1 | 21 | 21 |
| 2 | 20 | 23 |
| 3 | 22 |  21 | 

Will add this info to PR description as well.

go/vt/mysqlctl/backupstats/stats.go
maxenglander: Fixed!

maxenglander:
Hi @deepthi since the last review I:

 * Removed `Parameterizer` and moved `WithParams` into the `BackupStorage` interface. Left `TODO` in every backup storage except for `file` to actually make use of `backupstorage.Params`. This was the least amount of work for me, but happy to take a different approach if you prefer.
 * Added docs in https://github.com/vitessio/website/pull/1354. Note that as of now the website makes no reference to `vtbackup`. I created [a separate PR](https://github.com/vitessio/website/pull/1351) to introduce `vtbackup` into the docs.
 * Compared backup performance on `main` vs. this PR. Await your input on whether the overhead is acceptable or whether we need to dig in and improve it.

deepthi:
I'm also surprised at the amount of overhead. Given how close we are to branching off release-16.0, I'd like us to defer merging this until that is done. Once this is merged, we need to
- implement metrics for other backup storage methods. s3 is the most popular one and the one we should benchmark.
- investigate why the overhead is so much. I'm not too concerned about the file backup storage, but would be more worried if we saw the same kind of overhead with s3.
- run the benchmark on something other than a development laptop. implementing s3 and merging that will allow us to take the code and run it on PS infrastructure to get a good benchmark

So the way forward is:
- Wait until Feb 6, 2023 and merge this PR
- Implement metrics for s3 and merge them
- Benchmark and improve if necessary before v17

I'd be curious to know how much overhead we would already incur with s3 with the subset of metrics we would already get with this PR. IIUC, we'd already be getting the majority of metrics listed in this PR. All except `File:Read` and `File:Write`.

maxenglander:
@deepthi that sounds like a good plan to me :) 

I think the cause of overhead is probably this pattern:

```
timedDest := ioutil.NewMeteredWriteCloser(dest, destStats.TimedIncrementBytes)
```

It causes `destStats.TimedIncrementBytes` to be called every time we call `timedDest.Write`. That in turn increments multiple stats in the `stats` package. I think we can take a different approach, e.g. `defer destStats.TimedIncrementBytes(timedDest.TotalBytes)` or do some kind of buffered/throttled call to the `stats` package.

go/cmd/vtbackup/vtbackup.go
```diff
@@ -312,12 +322,14 @@ func takeBackup(ctx context.Context, topoServer *topo.Server, backupStorage back
...
 			return fmt.Errorf("backup failed: %v", err)
 		}
+		durationByPhase.Set("InitialBackup", int64(time.Since(backupParams.BackupTime).Seconds()))
 		log.Info("Initial backup successful.")
 		return nil
 	}
 
 	backupDir := mysqlctl.GetBackupDir(initKeyspace, initShard)
 	log.Infof("Restoring latest backup from directory %v", backupDir)
+	restoreAt := time.Now()
```
rsajwani, line 332: Just an idea : why not time.Now.Unix ?

maxenglander: It never occurred to me! I have seen the `s := time.Now() ... time.Since(s).Seconds()` in use elsewhere in the `mysqlctl` package, so I just went with that:

https://github.com/vitessio/vitess/blob/0a173bed6f08eceba38e06269e004ce5c76ee96d/go/vt/mysqlctl/backup.go#L149

I'll go ahead and make the switch to `time.Now().Unix()`.

maxenglander: Hm actually it looks like `time.Since` expects a `time.Time`, and `Unix()` returns an `int64` so it doesn't fit well into the pattern here.

rsajwani: hmm, i think its ok .. time.unix make integer manipulation easy . in terms of functional sense it will not make difference.

go/ioutil/meter.go
```diff
@@ -0,0 +1,58 @@
...
+import "time"
+
+// meter contains time-and-byte-tracking functionality.
+type meter struct {
+	fs       []func(b int, d time.Duration)
+	bytes    int64
+	duration time.Duration
+}
+
+// Bytes reports the total bytes read in calls to f so far.
```
rsajwani, line 33: f ?

maxenglander: Hm this refers to the argument `f` in `func measure`, but it's not very helpful here. Let me reword this.

maxenglander, line 33:
```suggestion
// Bytes reports the total bytes processed in calls to measure().
```

go/ioutil/meter.go
```diff
@@ -0,0 +1,58 @@
...
+func (mtr *meter) Bytes() int64 {
+	return mtr.bytes
+}
+
+// Duration reports the total time spend in calls to f so far.
```
maxenglander, line 38:
```suggestion
// Duration reports the total time spent in calls to measure().
```

go/ioutil/meter.go
```diff
@@ -0,0 +1,58 @@
...
+func (mtr *meter) Duration() time.Duration {
+	return mtr.duration
+}
+
+// measure tracks the measure it takes to execute f. Time is accumulated into total,
```
maxenglander, line 43:
```suggestion
// measure tracks the time spent and bytes processed by f. Time is accumulated into total,
```

maxenglander (commented):
Improve code comments in ioutil/meter

go/vt/mysqlctl/backup.go
rsajwani: just thinking out loud here.. can we have tow stats , enginestats and storagestats in backupparam.. so you won't end up copy

maxenglander: I think if we did that approach we might end up needing three stats, one for the engine, one for the storage, and one for the controlling function (`mysqlctl.Backup`).

Overall I think I am OK with the copy, and one thing I didn't know about Golang when I wrote the comment above is that this kind of struct creation is exhaustive:

	a1 := A{
		"hello",
		true,
		3,
	}

It won't let you omit any struct fields. So I at least feel good about that. The only risk now is if someone swaps the order of two struct fields that have the same type 😬

go/vt/mysqlctl/backupstats/constants.go
```diff
@@ -0,0 +1,17 @@
+/*
```
rsajwani, line 1: why do we need this file?

go/vt/mysqlctl/backup.go
```diff
@@ -85,8 +89,7 @@ var (
 	// once before the writer blocks
 	backupCompressBlocks = 2
 
-	backupDuration  = stats.NewGauge("backup_duration_seconds", "How long it took to complete the last backup operation (in seconds)")
-	restoreDuration = stats.NewGauge("restore_duration_seconds", "How long it took to complete the last restore operation (in seconds)")
+	titleCase = cases.Title(language.English).String
```
rsajwani, line 92: we need to document these in release notes about the deprecation and new counters..

go/vt/mysqlctl/backupstats/deprecated.go
```diff
@@ -0,0 +1,35 @@
...
+
+package backupstats
+
+import "vitess.io/vitess/go/stats"
+
+var (
+	// DeprecatedBackupDurationS is a deprecated statistic that will be removed
+	// in the next release. Use backup_duration_nanoseconds instead.
+	DeprecatedBackupDurationS = stats.NewGauge(
+		"backup_duration_seconds",
```
rsajwani, line 25: we need to add it to release notes

go/vt/mysqlctl/filebackupstorage/file.go
```diff
@@ -58,6 +62,23 @@ type FileBackupHandle struct {
 	errors   concurrency.AllErrorRecorder
 }
 
+func NewBackupHandle(
```
rsajwani, line 65: remind me why we have this method. Is it so we can assign default with noparam (backward compatibility)?

doc/releasenotes/17_0_0_summary.md
```diff
@@ -3,11 +3,26 @@
 ### Table of Contents
```
rsajwani, line 4: just wondering do you wanna talk about the newly added metrics here as well?

rsajwani: just like you have given example in website doc.... just floating an idea here

go/vt/mysqlctl/filebackupstorage/file.go
maxenglander: Yeah that's pretty much why. Makes it less likely to introduce nil dereference errors.

go/vt/mysqlctl/backup.go
maxenglander: Have added to release notes

go/vt/mysqlctl/backupstats/deprecated.go
maxenglander: Done

doc/releasenotes/17_0_0_summary.md
maxenglander: Added some more details!

maxenglander (commented):
@rsajwani thanks for all the helpful suggestions. I think I addressed all of your feedback, ready for another look!

rsajwani (approved):
LGTM. Thanks Max. This is awesome work.

deepthi (approved):

Issue #11810, opened 2022-11-24
set transaction isolation level extention to read queries


harshit-gangal:
The PR https://github.com/vitessio/vitess/pull/11673 added support for `set transaction isolation level` on a transaction.
This should be extended to read queries as well. 
As an example: an application that sets isolation to `read-uncommitted` to get the best possible read performance, even when using autocommit (i.e. not in a transaction) should be able to use this.

PR #11987, opened 2022-12-19
support transaction isolation modification through reserved connection system settings

harshit-gangal:
## Description

This PR moves the transaction isolation system variables (transaction_isolation or set transaction isolation level) from vitess aware to reserved connection.
Earlier it used to only work with transactions. With this change, support will be added for read queries.

## Related Issue(s)

- Closes https://github.com/vitessio/vitess/issues/11810
- Related to https://github.com/vitessio/vitess/pull/11673 and https://github.com/vitessio/vitess/pull/11804

## Checklist

### DIFF
--- a/go/test/endtoend/vtgate/reservedconn/sysvar_test.go
+++ b/go/test/endtoend/vtgate/reservedconn/sysvar_test.go
@@ -427,18 +427,36 @@ func TestSysVarTxIsolation(t *testing.T) {
 	require.NoError(t, err)
 	defer conn.Close()
 
+	// will run every check twice to see that the isolation level is set for all the queries in the session and
+
 	// default from mysql
 	utils.AssertMatches(t, conn, "select @@transaction_isolation", `[[VARCHAR("REPEATABLE-READ")]]`)
+	// ensuring it goes to mysql
+	utils.AssertContains(t, conn, "select @@transaction_isolation, connection_id()", `REPEATABLE-READ`)
+	// second run, to ensuring the setting is applied on the session and not just on next query after settings.
+	utils.AssertContains(t, conn, "select @@transaction_isolation, connection_id()", `REPEATABLE-READ`)
 
 	// setting to different value.
 	utils.Exec(t, conn, "set @@transaction_isolation = 'read-committed'")
-	utils.AssertMatches(t, conn, "select @@transaction_isolation", `[[VARCHAR("READ-COMMITTED")]]`)
+	utils.AssertMatches(t, conn, "select @@transaction_isolation", `[[VARCHAR("read-committed")]]`)
+	// ensuring it goes to mysql
+	utils.AssertContains(t, conn, "select @@transaction_isolation, connection_id()", `read-committed`)
+	// second run, to ensuring the setting is applied on the session and not just on next query after settings.
+	utils.AssertContains(t, conn, "select @@transaction_isolation, connection_id()", `read-committed`)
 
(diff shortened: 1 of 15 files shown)


go/vt/vtgate/safe_session.go
```diff
@@ -597,10 +597,10 @@ func (session *SafeSession) SetPreQueries() []string {
 	first := true
 	for _, k := range keys {
 		if first {
-			preQuery.WriteString(fmt.Sprintf("set @@%s = %s", k, sysVars[k]))
+			preQuery.WriteString(fmt.Sprintf("set %s = %s", k, sysVars[k]))
```
GuptaManan100, line 600: Why is this change needed?

GuptaManan100 (commented):
Everything else looks good, but isn't this like a breaking change? Previously setting the transaction isolation would only apply to the next query, but now it applies to the session, if I understand correctly. Shouldn't this be added to the release notes?

go/vt/vtgate/safe_session.go
harshit-gangal: TLDR; previously it was broken. 

@@ without any scope represents `session` scope. Only for the case of `transaction_isolation` it means the next transaction. This gets applied only once. 
With the implementation of `settings pool` these settings will not be applied again and the connection will be reused,  which means the next query the connection will see, will have the session level isolation level and not the next transaction isolation level.
So, the user expectation will turn out to be wrong.

The RFC already states that the next transaction isolation statements will be converted to the session level setting.

go/test/endtoend/vtgate/reservedconn/sysvar_test.go
```diff
@@ -427,18 +427,36 @@ func TestSysVarTxIsolation(t *testing.T) {
...
 	utils.AssertMatches(t, conn, "select @@transaction_isolation", `[[VARCHAR("REPEATABLE-READ")]]`)
+	// ensuring it goes to mysql
+	utils.AssertContains(t, conn, "select @@transaction_isolation, connection_id()", `REPEATABLE-READ`)
+	// second run, to ensuring the setting is applied on the session and not just on next query after settings.
+	utils.AssertContains(t, conn, "select @@transaction_isolation, connection_id()", `REPEATABLE-READ`)
 
 	// setting to different value.
 	utils.Exec(t, conn, "set @@transaction_isolation = 'read-committed'")
-	utils.AssertMatches(t, conn, "select @@transaction_isolation", `[[VARCHAR("READ-COMMITTED")]]`)
+	utils.AssertMatches(t, conn, "select @@transaction_isolation", `[[VARCHAR("read-committed")]]`)
```
deepthi, line 441: Why are some values uppercase and some lowercase?
Line 437 - `REPEATABLE-READ`. But everything else is lowercase.

deepthi: Looks like it may be an artifact of how the test input is provided. To avoid confusion, my suggestion is to make them all uniform.

deepthi (approved):
One small suggestion, rest LGTM

GuptaManan100 (approved):
I don't have any other comments than the ones already answered.

go/test/endtoend/vtgate/reservedconn/sysvar_test.go
harshit-gangal: I will raise a separate PR to handle it.

Issue #11965, opened 2022-12-14
Update vtgate advertised mysql version


harshit-gangal:
Currently, VTGate advertises 5.7.9 as the MySQL version to the clients. 
The MySQL 5.7 version is going EOL in October 2023. 
Mysql 8.0 was launched in April 2018 and it has been 4+ years. Any new installations are happening on the 8.0 versions. There are some MySQL clients/frameworks that rely on VTGate's advertised MySQL version.

In Vitess CI, we have also upgraded our environments to use the latest version of MySQL (8.0.31 released in October 2022).

The plan is to upgrade the advertised version to `8.0.31` and old installations can change the default by providing the `mysql_server_version` flag on the VTGate.

Issue #11948, opened 2022-12-13
Update the MySQL version used in the Docker lite images


frouioui:
The MySQL version we are using in our `vitess/lite` images is pretty old (`8.0.23`) we should upgrade it to a more recent and stable version: `8.0.31`.

- [ ] Find the implications of this change. Run some upgrade/downgrade tests between v15 and v16 on docker. and also with the operator.
- [ ] Use those findings to warns users in the release notes.
- [ ] We also need to change the default docker image we publish from 5.7 to 8.0.

deepthi:
We also need to change the default docker image we publish from 5.7 to 8.0.

frouioui:
Linking https://github.com/vitessio/vitess/issues/7812 to this issue as the enhancement proposed by the two issues is the same.

PR #11989, opened 2022-12-19
vtgate advertised mysql server version to 8.0.31

harshit-gangal:
## Description

This PR updates the MySQL server version advertised by VTGate from 5.7.9 to 8.0.31

## Related Issue(s)

- Closes https://github.com/vitessio/vitess/issues/11965
- Depends on #11948 

## Checklist


go/vt/servenv/buildinfo.go
```diff
@@ -94,7 +94,7 @@ func (v *versionInfo) MySQLVersion() string {
 	if mySQLServerVersion != "" {
 		return mySQLServerVersion
 	}
-	return "5.7.9-vitess-" + v.version
+	return "8.0.31-vitess-" + v.version
```
dbussink, line 97: I think we also need to change https://github.com/vitessio/vitess/blob/ece6501975604ef5d0be3ab256145047e13d9c00/go/mysql/collations/local.go#L38-L44 together with this?

Otherwise we get a reported version of 8.0, but still default collation settings for MySQL 5.7?

The code might warrant a refactor there so that we don't have to update this in multiple places somehow?

dbussink:
We have another default version number here that we need to change together with this I think:

https://github.com/vitessio/vitess/blob/ece6501975604ef5d0be3ab256145047e13d9c00/go/vt/sqlparser/parser.go#L46-L47

deepthi:
This should be done after https://github.com/vitessio/vitess/issues/11948 is fixed.

go/vt/servenv/buildinfo.go
harshit-gangal: done.

GuptaManan100 (approved):

GuptaManan100:
Approving to merge after the pre-requisites are met

deepthi:
Website location to update: https://vitess.io/docs/16.0/overview/supported-databases/
We currently say
```
The VTGate proxy server advertises its version as MySQL 5.7.
```

harshit-gangal:
> This should be merged only after #11948 is fixed.

To my understanding, this should not be considered a blocker. Currently, VTGate emits 5.7 and we still release vitess/lite with 8.0.23. what will impact when VTGate emits 8.0.31?

harshit-gangal:
Docs PR: https://github.com/vitessio/website/pull/1337

deepthi:
> To my understanding, this should not be considered a blocker. Currently, VTGate emits 5.7 and we still release vitess/lite with 8.0.23. what will impact when VTGate emits 8.0.31?

This is actually a good point and I'm glad you raised it.
We need to keep the vtgate server version in sync with the default version we ship with `vitess/lite`. Currently the default `vitess/lite` image ships with 5.7. We need to make that 8.0 along with making the vtgate version change from 5.7 -> 8.0.
There's no reason why that cannot be done as part of this PR, but it is something we'll need to announce to the community.

As for the minor version mismatch, I don't know that it really matters. In theory, it is possible for frameworks to send queries based on MySQL minor version, and if they do then making vtgate present as 8.0.31 when the backing MySQL is 8.0.23 may be problematic. In practice, I have no idea whether any of the popular frameworks are doing this. So keeping these versions in sync is only a precaution to avoid breaking people who are running with defaults.

harshit-gangal:
> If you want to make the default image change, all we need to do is change the symlink for `docker/lite/Dockerfile`.

done

harshit-gangal:
While looking at this in regards to docker image, we pass in the advertised version 
```
/vt/run.sh "8.0.31-vitess"
```

Docker image changes looks unrelated to the change in the PR. I will revert all the docker image changes made in this PR.

systay (approved):

harshit-gangal:
blocked for merge. waiting on https://github.com/vitessio/vitess/pull/12054

frouioui (approved):

Issue #11904, opened 2022-12-07
Bug Report: v12 and v13 releases for tabletgateway don't perform alias nil check in tx_conn.go queryService function


pbibra:
### Overview of the Issue

discussion here: https://vitess.slack.com/archives/C0PQY0PTK/p1670351761839309

When performing COMMITs using `tabletgateway`, we run into the following error: 

```
tablet: <nil> is either down or nonexistent"
```

This nil check does not exist in v12 and v13 releases and should probably be backported: https://github.com/slackhq/vitess/commit/18234019398bac30dfe4040f529cb5a7fe68bc00#diff-33ceb6465a70a93307b16c53d48aa46ec05feabbfa39f5de52522f5eb8f52835R277

### Reproduction Steps

1. Use Vitess 12 
2. Set `gateway_implementation` to `tabletgateway`
3. Send a `COMMIT` transaction to a VTGate
4. Should get the `tablet: <nil> is either down or nonexistent"` error. The `queryService` function in `tx_conn.go` tries to fetch `QueryServiceByAlias()->TabletConnection()` which is a function only implemented by `TabletGateway` and the new `healthcheck` respectively.

### Binary Version

```sh
v12.0.5
```

### Operating System and Environment details

```sh
n/a
```

systay:
v12 and v13 are no longer supported

PR #11992, opened 2022-12-19
[release-12] backport query service nil alias check to v12

pbibra:

## Description

v12 users using the `tabletgateway` may run into the following error when running COMMIT type transactions:

```
tablet: <nil> is either down or nonexistent"
```

This is because the `QueryServiceByAlias` implementation in `TabletGateway` does not account for a nil alias. This was fixed when the `discoverygateway` was deprecated so backporting that here.

## Related Issue(s)

https://github.com/vitessio/vitess/issues/11904

## Checklist


GuptaManan100 (approved):

harshit-gangal:
v12.0 is out of Vitess Maintainance cycle. Closing this PR.

pbibra:
v12 is out of maintenance

PR #11994, opened 2022-12-20
go/vt/mysqlctl: add compression benchmarks

maxenglander:
Relates to #7802.

These benchmarks are proving useful in internal discussions around improving the performance of backups. Sharing here in case others find them useful. 

This benchmark downloads and compresses [this ~60GB InnoDB file](https://archive.org/details/enwiki-20221220-externallinks.ibd), which was built by loading [this Wikipedia dataset](https://dumps.wikimedia.org/enwiki/20221220/enwiki-20221220-externallinks.sql.gz) into MySQL 8.

```
$ go test -bench=BenchmarkCompress ./go/vt/mysqlctl -run=NONE -timeout=12h -benchtime=1x -v
goos: darwin
goarch: arm64
pkg: vitess.io/vitess/go/vt/mysqlctl
BenchmarkCompressLz4Builtin
    compression_benchmark_test.go:310: downloading data from https://www.dropbox.com/s/raw/smmgifsooy5qytd/enwiki-20080103-pages-articles.ibd.tar.zst
BenchmarkCompressLz4Builtin-8                  1        11737493087 ns/op        577.98 MB/s             2.554 compression-ratio
BenchmarkCompressPargzipBuiltin
BenchmarkCompressPargzipBuiltin-8              1        31083784040 ns/op        218.25 MB/s             2.943 compression-ratio
BenchmarkCompressPgzipBuiltin
BenchmarkCompressPgzipBuiltin-8                1        13325299680 ns/op        509.11 MB/s             2.910 compression-ratio
BenchmarkCompressZstdBuiltin
BenchmarkCompressZstdBuiltin-8                 1        18683863911 ns/op        363.09 MB/s             3.150 compression-ratio
BenchmarkCompressZstdExternal
BenchmarkCompressZstdExternal-8                1        10795487675 ns/op        628.41 MB/s             3.093 compression-ratio
BenchmarkCompressZstdExternalFast4
BenchmarkCompressZstdExternalFast4-8           1        7139319009 ns/op         950.23 MB/s             2.323 compression-ratio
BenchmarkCompressZstdExternalT0
BenchmarkCompressZstdExternalT0-8              1        4393860434 ns/op        1543.97 MB/s             3.093 compression-ratio
BenchmarkCompressZstdExternalT4
BenchmarkCompressZstdExternalT4-8              1        4389559744 ns/op        1545.49 MB/s             3.093 compression-ratio
PASS
cleaning up "/var/folders/96/k7gzd7q10zdb749vr02q7sjh0000gn/T/ee7d47b45ef09786c54fa2d7354d2a68.dat"
```

System info:

 * `Version: 16.0.0-SNAPSHOT (Git revision 90baaedac2ce6aef89977846c44c313a371ef234 branch 'maxeng-zstd-faster') built on Tue Dec 20 00:00:48 EST 2022 by maxenglander@planetscale.local using go1.19.4 darwin/arm64`
 * `Darwin planetscale.local 21.6.0 Darwin Kernel Version 21.6.0: Thu Sep 29 20:13:46 PDT 2022; root:xnu-8020.240.7~1/RELEASE_ARM64_T8101 arm64`
 * image

### DIFF
--- a/go/vt/mysqlctl/compression_benchmark_test.go
+++ b/go/vt/mysqlctl/compression_benchmark_test.go
@@ -30,6 +30,16 @@ type (
 		dataLocalPath string
 		dataURL       string
 	}
+
+	countingReader struct {
+		count int64
+		r     io.Reader
+	}
+
+	countingWriter struct {
+		count int64
+		w     io.Writer
+	}
 )
 
 const (
@@ -56,35 +66,58 @@ func setupBenchmarkCompressEnv(args benchmarkCompressArgs) benchmarkCompressEnv 
 }
 
 func (bce *benchmarkCompressEnv) benchmark() {
+	var numUncompressedBytes, numCompressedBytes int64
+
 	bce.b.StopTimer()
 	bce.b.ResetTimer()
 
 	for i := 0; i < bce.b.N; i++ {
+		// Create compressor.
 		logger := logutil.NewMemoryLogger()
-		compressor := bce.compressor(logger)
+		cWriter := &countingWriter{w: io.Discard}
+		compressor := bce.compressor(logger, cWriter)
+
 		reader, err := bce.reader()
 		require.Nil(bce.b, err, "Failed to get data reader.")
+		cReader := &countingReader{r: reader}
 
 		// Only time the part we care about.
 		bce.b.StartTimer()
-		_, err = io.Copy(compressor, reader)
+		_, err = io.Copy(compressor, cReader)
 		bce.b.StopTimer()
 
 		// Don't defer closing, otherwise we can exhaust open file limit.
 		reader.Close()
 		compressor.Close()
 		require.Nil(bce.b, err, logger.Events)
+
+		// Report how many bytes we've compressed.
+		bce.b.SetBytes(cReader.count)
+
+		// Record how many bytes compressed so we can report these later.
+		numCompressedBytes += cWriter.count
+		numUncompressedBytes += cReader.count
 	}
+
+	// Report the average compression ratio (= <uncompressed>:<compressed>) achieved per-operation.
+	//
+	// Must report this outside of b.N loop because b.ReportMetric overrides
+	// any previously reported value for the same unit.
+	bce.b.ReportMetric(
+		float64(numUncompressedBytes)/float64(numCompressedBytes),
+		// By convention, units should end in /op.
+		"compression-ratio/op",
+	)
 }
 
-func (bce *benchmarkCompressEnv) compressor(logger logutil.Logger) io.WriteCloser {
+func (bce *benchmarkCompressEnv) compressor(logger logutil.Logger, writer io.Writer) io.WriteCloser {
 	var compressor io.WriteCloser
 	var err error
 
 	if bce.builtin != "" {
-		compressor, err = newBuiltinCompressor(bce.builtin, io.Discard, logger)
+		compressor, err = newBuiltinCompressor(bce.builtin, writer, logger)
 	} else if bce.external != "" {
-		compressor, err = newExternalCompressor(context.Background(), bce.external, io.Discard, logger)
+		compressor, err = newExternalCompressor(context.Background(), bce.external, writer, logger)
 	}
 
 	require.Nil(bce.b, err, "Failed to create compressor.")
@@ -155,6 +188,18 @@ func (bce *benchmarkCompressEnv) validate() {
 	if bce.builtin == "" && bce.external == "" {
 		require.Fail(bce.b, "Either builtin or external compressor must be specified.")
 	}
+}
+
+func (cr *countingReader) Read(p []byte) (nbytes int, err error) {
+	nbytes, err = cr.r.Read(p)
+	cr.count += int64(nbytes)
+	return
+}
+
+func (cr *countingWriter) Write(p []byte) (nbytes int, err error) {
+	nbytes, err = cr.w.Write(p)
+	cr.count += int64(nbytes)
+	return
 }
 
 func BenchmarkCompressLz4Builtin(b *testing.B) {
--- a/go/vt/mysqlctl/compression_benchmark_test.go
+++ b/go/vt/mysqlctl/compression_benchmark_test.go
@@ -61,7 +61,7 @@ const (
 	// built from this Wikipedia dataset:
 	//
 	//     https://dumps.wikimedia.org/archive/enwiki/20080103/enwiki-20080103-pages-articles.xml.bz2
-	defaultDataURL = "https://www.dropbox.com/s/raw/smmgifsooy5qytd/enwiki-20080103-pages-articles.ibd.tar.zst"
+	defaultDataURL = "https://github.com/vitessio/vitess-resources/releases/download/testdata-v1.0/enwiki-20080103-pages-articles.ibd.tar.zst"
 
 	// By default, don't limit how many bytes we input into compression.
 	defaultMaxBytes int64 = 0
--- a/go/vt/mysqlctl/compression_benchmark_test.go
+++ b/go/vt/mysqlctl/compression_benchmark_test.go
@@ -1,6 +1,7 @@
 package mysqlctl
 
 import (
+	"bufio"
 	"compress/gzip"
 	"context"
 	"crypto/md5"
@@ -10,8 +11,10 @@ import (
 	"net/http"
 	"os"
 	"path"
+	"strconv"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/require"
 
@@ -20,10 +23,9 @@ import (
 
 type (
 	benchmarkCompressArgs struct {
-		b          *testing.B
-		builtin    string
-		external   string
-		saveToFile bool
+		b        *testing.B
+		builtin  string
+		external string
 	}
 
 	benchmarkCompressEnv struct {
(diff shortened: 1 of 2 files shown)


rsajwani:
I don't see any issue in merging this. I think it it is useful. I believe we should include it in our release notes if we end up merging it.

go/vt/mysqlctl/compression_benchmark_test.go
```diff
@@ -0,0 +1,190 @@
...
+to obtain compressors.
+
+You must supply a file to compress by setting this environment variable:
+
+    export VT_MYSQLCTL_COMPRESSION_BENCHMARK_DATA_PATH=path/to/data.xml
+
+We recommend using a data set that compresses well, such as:
+
+ - The enwik9 dataset: http://mattmahoney.net/dc/textdata
+ - A larger Wikipedia dataset from: https://en.wikipedia.org/wiki/Wikipedia:Database_download
```
rsajwani, line 41: or we can use dd (https://www.cyberciti.biz/faq/howto-create-lage-files-with-dd-command/)
Not sure if dd does not result in too much compression..

rsajwani (approved):

go/vt/mysqlctl/compression_benchmark_test.go
maxenglander: I can add it if you think it makes sense. I initially tried `head -c[bignumber] /dev/urandom`, but ended up with too little compression. Got guidance to use realistic datasets.

maxenglander:
@rsajwani added some 16_0_0 notes, let me know if you had something else in mind.

go/vt/mysqlctl/compression_benchmark_test.go
rsajwani: make sense.. I think i like wikipedia databset.. it looked more realistic.

vmg:
Hiiii! I really enjoy benchmarking. Here are some notes of how I think good benchmarks should be written:

1. I think that a good benchmark must be reproducible. Otherwise it doesn't contribute anything to the codebase -- we can't make good decisions off its data, and most notably, we cannot use it to catch performance regressions! That's the key objective of a benchmark. We need to be able to measure how the performance of the system evolves over time.

    Obviously, to be reproducible, the inputs to the benchmark need to be constant. Manually downloading an arbitrary XML dataset from Wikipedia does not lend itself to reproducibility! We cannot commit the benchmarked data to the repository, but I think that in order to merge this code, the benchmark code needs to be able to fetch on its own the data it'll be benchmarking. The benchmark should not have accompanying documentation on how to run it; it needs to run on its own without setup, and it needs to run with the same data so its results are always comparable.

2. A good benchmark must also be representative. It needs to measure something that is important to us. In this case, I posit the question: does an XML dump from Wikipedia look like a MySQL backup from `xtrabackup`? My hunch is that it doesn't! I think a MySQL dump is going to have _a lot of binary data_ and an XML dump is gonna have exactly 0 bytes of binary data. So measuring how good is `zstd` at compressing XML is not going to tell us how good it is at compressing our backups.

    This is a challenge, because where can we find an actual backup dump hosted somewhere so that we can reliably benchmark it? I'm afraid I don't have good ideas here. Maybe there _are_ backups hosted somewhere that we could use as a reference. Maybe the right solution is synthetically generating a backup by converting a random seed into data.

    Either way, I think this _is_ important to get right. Here's why: I have a hunch that once all is said and done and measured, the most efficient approach to compressing backups in both time and size is going to be a `zstd` variant using a [MySQL-backup tuned dictionary](https://raw.githack.com/facebook/zstd/release/doc/zstd_manual.html#Chapter13). But the only way we'll ever get to verify that is if our benchmarks are actually trying to compress MySQL backups instead of the life and works of [Srinivasa Ramanujan](https://en.wikipedia.org/wiki/Srinivasa_Ramanujan) (very interesting Wikipedia page btw). So getting this right and benchmarking the right dataset is going to be annoying, but worth it in the long run.

3. Last but not least, a good benchmark must be meaningful. If you run a compression benchmark and it tells you that `zstd` compresses at, huh, `27094172323 ns/op`... That's not very actionable information. What even is an OP? Is this fast at all? I have no idea! There's clearly two issues here:
    - Nanoseconds per operation is not how you measure compression speed. It provides no information because nanoseconds are _tiny_ and operations are arbitrary. A more interesting metric would be (mega)bytes per second (i.e. _throughput_), something that the go `Benchmark` library fortunately allows us to capture, measure and report: https://pkg.go.dev/testing#B.SetBytes
    - Neither nanoseconds per operation, nor bytes per second, are a meaningful enough metric on their own. Because we're benchmarking a compression algorithm! I could win this benchmark quite easily if I just... copy data back and forth. My "compression algorithm" runs at the speed of RAM. Of course, it doesn't actually _compress_ the data, but you won't be able to tell from the benchmark, will you?
    To make meaningful decisions when comparing compression algorithms, we must evaluate their throughput _and_ their compression ratio (and this leads us to our previous point; the compression ratio is also going to be wildly different between XML and binary dumps). Once again, the `Benchmark` code in the Go standard library is comprehensive enough: https://pkg.go.dev/testing#B.ReportMetric allows us to report ratios together with our throughput.

So that's my old man advice about benchmarks! The number one blocker for this PR IMO is the reproducibility issue. Nobody is going to run these benchmarks if you need to come to this issue and look up how to fetch data with `curl`. But once we get to reproducible benchmarks, we might as well make them reproducible with meaningful data. Let's get this right so this is worth merging!

maxenglander:
Hey @vmg thanks for the detailed feedback and helpful suggestions. Feel like I leveled up just from reading this. All feels very doable.

vmg:
You're welcome! Please reach out if you get stuck trying to find a sample backup that can be used for benchmarking. I have some ideas on how we could generate them synthetically.

maxenglander:
Hey @vmg I updated the benchmarks following your suggestions.

 * I downloaded [this Wikipedia dataset](https://dumps.wikimedia.org/enwiki/20221220/enwiki-20221220-externallinks.sql.gz) (contains mostly URLs), loaded it into MySQL 8, and uploaded the InnoDB (`.ibd`) file [here on archive.org](https://archive.org/details/enwiki-20221220-externallinks.ibd).
 * I updated the benchmark suite to download and compress that file. No longer printing out any usage info. Users may override the URL, but this isn't advertised, and is mainly intended for debugging purposes.
 * Added a `SetBytes` metric.
 * Reporting `compression-ratio/op` metric. Chose this awkward name because the `ReportMetric` docs say that units end in `/op` by convention. Happy to rename this based on your advice though.

Not sure if it's a good idea to [host this file on archive.org](https://archive.org/details/enwiki-20221220-externallinks.ibd). Maybe @rsajwani will have thoughts on a better place to host this file.

vmg:
This looks great! My only concern is (ab)using `archive.org`. I don't know where else could we host this in the long term. Thanks for taking the time to greatly improve this PR though! :+1: from me!

vmg (approved):

rsajwani:
> This looks great! My only concern is (ab)using `archive.org`. I don't know where else could we host this in the long term. Thanks for taking the time to greatly improve this PR though! 👍 from me!

I was trying to see if we have any S3 location which we can use or are already using for our testing... cc: @deepthi . Do we have a S3 bucket which we use for testing and have our own testing account ... may be we can upload it to there ...

maxenglander:
FYI @rsajwani Deepthi suggested I take a look at this as an option: https://github.com/vitessio/vitess-resources

rsajwani:
> FYI @rsajwani Deepthi suggested I take a look at this as an option: https://github.com/vitessio/vitess-resources

yeah, that looks a good idea as well..

maxenglander:
Thanks to @GuptaManan100 we now have the benchmark data set in a reliable place under our control. Results from running the benchmarks with the latest commit.

```
$ go test -bench=BenchmarkCompress ./go/vt/mysqlctl -run=NONE -timeout=12h -benchtime=1x -v
goos: darwin
goarch: arm64
pkg: vitess.io/vitess/go/vt/mysqlctl
BenchmarkCompressLz4Builtin
    compression_benchmark_test.go:310: downloading data from https://github.com/vitessio/vitess-resources/releases/download/testdata-v1.0/enwiki-20080103-pages-articles.ibd.tar.zst
BenchmarkCompressLz4Builtin-8                  1        12132147170 ns/op        559.18 MB/s             2.554 compression-ratio
BenchmarkCompressPargzipBuiltin
BenchmarkCompressPargzipBuiltin-8              1        30344518890 ns/op        223.57 MB/s             2.943 compression-ratio
BenchmarkCompressPgzipBuiltin
BenchmarkCompressPgzipBuiltin-8                1        12743268729 ns/op        532.36 MB/s             2.910 compression-ratio
BenchmarkCompressZstdBuiltin
BenchmarkCompressZstdBuiltin-8                 1        18400257932 ns/op        368.69 MB/s             3.150 compression-ratio
BenchmarkCompressZstdExternal
BenchmarkCompressZstdExternal-8                1        11445409832 ns/op        592.73 MB/s             3.093 compression-ratio
BenchmarkCompressZstdExternalFast4
BenchmarkCompressZstdExternalFast4-8           1        7127384284 ns/op         951.82 MB/s             2.323 compression-ratio
BenchmarkCompressZstdExternalT0
BenchmarkCompressZstdExternalT0-8              1        3997374544 ns/op        1697.11 MB/s             3.093 compression-ratio
BenchmarkCompressZstdExternalT4
BenchmarkCompressZstdExternalT4-8              1        4154964174 ns/op        1632.75 MB/s             3.093 compression-ratio
PASS
cleaning up "/var/folders/96/k7gzd7q10zdb749vr02q7sjh0000gn/T/50bd7cc81b5edaf5bdb887a7c4ab3ba5.dat"
```

This is ready to merge (or get more feedback).

PR #11998, opened 2022-12-20
only expand when we have full information

### LINKED ISSUE #11880: Bug Report: `SELECT *` expansion only considers subquery in column list on unsharded queries
### Overview of the Issue

@yoheimuta encountered a bug where a subquery in the columns list of an inner query would be the only column expanded by the `SELECT *` expression of the outer query. This bug only exists in the Gen4 planner.

### Reproduction Steps

Using the vschema defined at `go/vt/vtexplain/testdata/test-vschema.json`:

```sql
SELECT * FROM (
  SELECT *,
  (SELECT id FROM unsharded_b LIMIT 1) AS unsharded_b_id
  FROM unsharded_a LIMIT 10
) as results
```

V3 plan:

```json
{
  "QueryType": "SELECT",
  "Original": "SELECT * FROM (SELECT *, (SELECT id FROM unsharded_b LIMIT 1) AS unsharded_b_id FROM unsharded_a LIMIT 10) as results",
  "Instructions": {
    "OperatorType": "Route",
    "Variant": "Unsharded",
    "Keyspace": {
      "Name": "main",
      "Sharded": false
    },
    "FieldQuery": "select * from (select *, (select id from unsharded_b where 1 != 1) as unsharded_b_id from unsharded_a where 1 != 1) as results where 1 != 1",
    "Query": "select * from (select *, (select id from unsharded_b limit 1) as unsharded_b_id from unsharded_a limit 10) as results",
    "Table": "unsharded_a"
  }
}
```

Gen4 plan:

```json
{
  "QueryType": "SELECT",
  "Original": "SELECT * FROM (SELECT *, (SELECT id FROM unsharded_b LIMIT 1) AS unsharded_b_id FROM unsharded_a LIMIT 10) as results",
  "Instructions": {
    "OperatorType": "Route",
    "Variant": "Unsharded",
    "Keyspace": {
      "Name": "main",
      "Sharded": false
    },
    "FieldQuery": "select results.unsharded_b_id from (select *, (select id from unsharded_b where 1 != 1) as unsharded_b_id from unsharded_a where 1 != 1) as results where 1 != 1",
    "Query": "select results.unsharded_b_id from (select *, (select id from unsharded_b limit 1) as unsharded_b_id from unsharded_a limit 10) as results",
    "Table": "unsharded_a, unsharded_b"
  },
  "TablesUsed": [
    "main.unsharded_a",
    "main.unsharded_b"
  ]
}
```

### Binary Version

```sh
Version: 22067c7a4e334a7e90850a2119135e6b1ae23279
```

```sh
N/A
```

### Log Fragments

_No response_

systay:
## Description
Before planning a query, vtgate will simplify the query. One of the simplifications is to expand `*` outputs in the query.
This is only done if we have enough schema information about the tables in the query.

The issue reported in #11880 is related to a derived table where we don't have enough information to expand the columns. Before this fix, we would just use whatever column information we had, which is not correct.

This PR changes the semantic analysis so it keeps track in more places if we can successfully expand columns or not, and will just not expand when we are missing necessary information.

## Related Issue(s)
Fixes #11880

## Checklist

### DIFF
--- a/go/vt/vtgate/semantics/derived_table.go
+++ b/go/vt/vtgate/semantics/derived_table.go
@@ -27,17 +27,18 @@ import (
 
 // DerivedTable contains the information about the projection, tables involved in derived table.
 type DerivedTable struct {
-	tableName   string
-	ASTNode     *sqlparser.AliasedTableExpr
-	columnNames []string
-	cols        []sqlparser.Expr
-	tables      TableSet
+	tableName       string
+	ASTNode         *sqlparser.AliasedTableExpr
+	columnNames     []string
+	cols            []sqlparser.Expr
+	tables          TableSet
+	isAuthoritative bool
 }
 
 var _ TableInfo = (*DerivedTable)(nil)
 
 func createDerivedTableForExpressions(expressions sqlparser.SelectExprs, cols sqlparser.Columns, tables []TableInfo, org originable) *DerivedTable {
-	vTbl := &DerivedTable{}
+	vTbl := &DerivedTable{isAuthoritative: true}
 	for i, selectExpr := range expressions {
 		switch expr := selectExpr.(type) {
 		case *sqlparser.AliasedExpr:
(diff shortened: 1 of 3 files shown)


harshit-gangal (approved):

GuptaManan100 (approved):

arthurschreiber (commented):
❤️

PR #12000, opened 2022-12-21
Remove unnecessary logging

### LINKED ISSUE #11869: Bug Report: `BUG: tablet identifier should represent single table: TableSet{0,1}`
### Overview of the Issue

When executing a subquery that contains more than a single table (e.g. by joining two tables), the planner generates the following warnings:

```
W1201 19:52:27.315947    7944 SQL_builder.go:79] error in replacing table expression in semtable: Code: INTERNAL
BUG: tablet identifier should represent single table: TableSet{0,1}
```

Note that the planner does not fail - it only outputs the error but then ignores it.

### Reproduction Steps

Use the vschema defined in `go/vt/vtgate/planbuilder/testdata/vschemas/schema.json`.

Run the following query:

```sql
select count(*) from (select 1 from music join user on user.id = music.user_id where music.id = 10) as __subquery_for_count
```

### Binary Version

```sh
Built using commit 22067c7a4e334a7e90850a2119135e6b1ae23279.
```

```sh
N/A
```

### Log Fragments

_No response_

systay:
## Description
The perfectly valid use case of derived tables that have been pushed under a `Route` lead to a warning from the planner that wasn't really helpful or informative. The solution was to simply remove this logging altogether.

## Related Issue(s)
Fixes #11869

## Checklist

### DIFF
--- a/go/vt/vtgate/planbuilder/operators/SQL_builder.go
+++ b/go/vt/vtgate/planbuilder/operators/SQL_builder.go
@@ -23,7 +23,6 @@ import (
 
 	"vitess.io/vitess/go/vt/vtgate/planbuilder/operators/ops"
 
-	"vitess.io/vitess/go/vt/log"
 	"vitess.io/vitess/go/vt/sqlparser"
 	"vitess.io/vitess/go/vt/vterrors"
 	"vitess.io/vitess/go/vt/vtgate/planbuilder/plancontext"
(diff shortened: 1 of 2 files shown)


GuptaManan100 (approved):

GuptaManan100 (commented):
Could you add a unit test for the failing query? The review checklist requires it.

arthurschreiber (approved):

Issue #11664, opened 2022-11-08
Information_schema views not correct


systay:
In relational databases, the information schema (`information_schema`) is an ANSI-standard set of read-only views that provide information about all of the tables, views, columns, and procedures in a database.

Vitess today has limited support for these views. The planner will inspect a query and try to figure out which keyspace it's trying to inspect by looking at the predicates given, and send the query to a random shard in that keyspace. If it is unable to use the predicates to figure out where to send it, Vitess will send it to the default keyspace.

There are some important short-comings:
1. Multiple keyspaces. Since a Vitess system can have multiple keyspaces, and these can be backed by mysql instances that know nothing about the other keyspaces, no one keyspace will contain information about all tables in the Vitess system. It is not possible to get query results from multiple keyspaces today.
2. Database names. From the users perspective, a keyspace is equivalent to a mysql database. A keyspace named XXX can be backed by a mysql database named YYY. Today, we return the whatever the underlying mysql information_schema has, which might not be the same as the keyspace name.
3. Other constructs, such as [views](https://github.com/vitessio/vitess/issues/11559), also needs to be present in the information schema.

`performance_schema` is very different, and we don't need to support that.

frouioui:
> Multiple keyspaces. Since a Vitess system can have multiple keyspaces, and these can be backed by mysql instances that know nothing about the other keyspaces, no one keyspace will contain information about all tables in the Vitess system. It is not possible to get query results from multiple keyspaces today.

I thought we supported multi-keyspaces queries such as: `select ks1.t.id, k2.t.id from ks1.t, ks2.t`. Do you mean that we cannot get information_schema queries results from multiple keyspaces? Or do you mean that we don't support queries like `select t1.id, t2.id from t1, t2`, where `t1` is in `ks1` and `t2` in `ks2`.

> Database names. From the users perspective, a keyspace is equivalent to a mysql database. A keyspace named XXX can be backed by a mysql database named YYY. Today, we return the whatever the underlying mysql information_schema has, which might not be the same as the keyspace name.

Do you have an example of how a keyspace named XXX can be backed by a mysql database named YYY? Do we prefix the database with `vt_`?

systay:
> I thought we supported multi-keyspaces queries such as: `select ks1.t.id, k2.t.id from ks1.t, ks2.t`. Do you mean that we cannot get information_schema queries results from multiple keyspaces? 

[...]

PR #12006, opened 2022-12-21
better querying of information_schema views

systay:
## Description
Changes how Vitess handles `information_schema` queries.

The logic is:

1. If the query is specific to a single database, by having predicates against `TABLE_SCHEMA` or similar columns, we can route the query to a single keyspace.
2. The query can be against one of the tables that does not have schema names in any of the columns, for example `INFORMATION_SCHEMA.COLLATIONS`. In this case we will select the default  keyspace if one is selected, or to an arbitrary keyspace and send the query to it.
3. Finally, if the table does contain schema specific column(s), but no predicates against them, we need to send the query to one keyspace with the given predicates, and then concatenate that with information from all other keyspaces. From these, we only want the rows that are specific to that keyspace.

So, for this simple query:

```sql
select table_name from information_schema.tables
```

The plan produced would look something like this:
image

## Related Issue(s)
#11664

Left to do:

- [x] Use the selected KS and not any arbitrary when possible
- [ ] Add `CASE` expressions to that the keyspace name is what is in the schema column output, even if the user is using a different name for the local database
- [ ] Add lots of end-to-end tests

## Checklist


go/vt/vtgate/planbuilder/testdata/filter_cases.json
```diff
@@ -4400,6 +4400,140 @@
         "Query": "select t.table_schema from information_schema.`tables` as t where t.table_schema in (select c.column_name from information_schema.`columns` as c)",
         "Table": "information_schema.`tables`"
       }
+    },
+    "gen4-plan": {
+      "QueryType": "SELECT",
+      "Original": "select t.table_schema from information_schema.tables as t where t.table_schema in (select c.column_name from information_schema.columns as c)",
```
GuptaManan100, line 4406: I really like this. In comparison, v3 looks so stupid 😆

go/vt/vtgate/planbuilder/plan_test.go
```diff
@@ -589,7 +590,14 @@ func (vw *vschemaWrapper) ForeignKeyMode() string {
 
 func (vw *vschemaWrapper) AllKeyspace() ([]*vindexes.Keyspace, error) {
```
GuptaManan100, line 591: nit: I think the name should be plural like `AllKeyspaces`

go/vt/vtgate/planbuilder/operators/route_planning.go
```diff
@@ -63,6 +62,28 @@ func transformToPhysical(ctx *plancontext.PlanningContext, in ops.Operator) (ops
...
+			}
+
+			var newSrcs []ops.Operator
+			for _, src := range union.Sources {
+				_, ok = src.(*Route)
+				if !ok {
+					return op, rewrite.SameTree, nil
+				}
+
+				// if we got here we are dealing with a UNION created for an info_schema table
```
GuptaManan100, line 78: I don't quite understand this. Why is it necessary that the union is for `info_schema` tables. Could it not just a union of other routes?

go/vt/vtgate/planbuilder/operators/info_schema_planning.go
```diff
@@ -0,0 +1,361 @@
...
+	return col, isDbNameCol(col), isTableNameCol(col)
+}
+
+func isDbNameCol(col *sqlparser.ColName) bool {
+	return col.Name.EqualString("table_schema") || col.Name.EqualString("constraint_schema") || col.Name.EqualString("schema_name") || col.Name.EqualString("routine_schema")
+}
+
+func isTableNameCol(col *sqlparser.ColName) bool {
+	return col.Name.EqualString("table_name")
+}
```
GuptaManan100, line 283: There are 2 more implementations of it in the `planbuilder` package, which have different definitions. We should merge them into one.

GuptaManan100 (commented):
overall, I like how it is turning out.

PR #12007, opened 2022-12-21
Implement the RowsColumnTypeScanType interface in the go sql driver

johanoskarsson:
## Description

I attempted to use https://github.com/jimsmart/schema to introspect the vitess schema via the vitessdriver. Much to my surprise all of the column types were returned as `interface{}`. I dug deeper and found that the Go sql code attempts to cast the row to `RowsColumnTypeScanType` and if that fails it will not be able to return the types.

`RowsColumnTypeScanType` contains just one function that I've implemented below. Example of how the mysql drivers implemented it: https://github.com/go-sql-driver/mysql/blob/4591e42e65cf483147a7c7a4f4cfeac81b21c917/fields.go#L140

Overall it's fairly straight forward I have tested it against our vitess locally. However a few of the types are unfamiliar to me, such as HEXNUM so any extra eyes would be appreciated.

## Related Issue(s)

Adding the vitess dialect to the schema project here https://github.com/jimsmart/schema/pull/21

## Checklist

## Deployment Notes

Should not require any particular deployment notes.

### DIFF
--- a/go/vt/vitessdriver/rows_test.go
+++ b/go/vt/vitessdriver/rows_test.go
@@ -18,10 +18,12 @@ package vitessdriver
 
 import (
 	"database/sql/driver"
+	"fmt"
 	"io"
 	"reflect"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
 	"vitess.io/vitess/go/sqltypes"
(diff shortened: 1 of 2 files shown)


jimsmart:
Just to add to the conversation here...

When I initially wrote https://github.com/jimsmart/schema some 5 years ago, I realised that the various database drivers out there actually have very different support for what is returned as sql.ColumType metadata.

I tried to catalogue some of that info, in a separate project — originally with the intent of further looking into the differences, and submitting issues / patches, and perhaps even opening an issue on the main Go repo to discuss / track some of the issues I found. Unfortunately, to date, I never got around to any of that.

But here is the project I built to catalogue the driver capabilities WRT sql.ColumnType — the table in the main README is manually composed, but in the subfolders all of the READMEs are automatically generated (and it is from these that the main README is summarised). These individual reports can be accessed via the 'View' link in the Report column in the project's top-level README. I only test common column types.

Note that this project, and its reports, have not been updated in ~5yrs.

https://github.com/jimsmart/drivercaps

The long and short of it, and why I mention it here, is that support for .ScanType() varies greatly, with some drivers returning their own types for certain column types — browsing the individual reports for what other db drivers return for common column types may be helpful to you as a point of reference.

systay:
@johanoskarsson So, the vitess-driver is not actively worked on. We usually suggest that users use the standard mysql connection to vitess, or alternatively connect using grpc. Have you tried connection using the normal mysql connection protocol?

I'll look over your changes as well, just wanted to share this piece of info first.

go/vt/vitessdriver/rows_test.go
```diff
@@ -135,3 +135,95 @@ func TestRows(t *testing.T) {
...
+		typeInt64,
+		typeUint64,
+		typeFloat32,
+		typeFloat64,
+		typeRawBytes,
+		typeTime,
+	}
+
+	for i := 0; i < len(wantTypes); i++ {
+		if ri.ColumnTypeScanType(i) != wantTypes[i] {
```
systay, line 223: nitpick: would you mind using `stretchr/testify` for the assertions?

go/vt/vitessdriver/rows_test.go
```diff
@@ -135,3 +135,95 @@ func TestRows(t *testing.T) {
...
+			t.Errorf("unexpected type %v, wanted %v", ri.ColumnTypeScanType(i), wantTypes[i])
+		}
+	}
+
+	_ = ri.Close()
```
systay, line 228: if it's important to close this, it should be done in a `defer`

go/vt/vitessdriver/rows_test.go
```diff
@@ -135,3 +135,95 @@ func TestRows(t *testing.T) {
...
+			{
+				Name: "field13",
+				Type: sqltypes.VarBinary,
+			},
+			{
+				Name: "field14",
+				Type: sqltypes.Datetime,
+			},
+		},
+		RowsAffected: 0,
```
systay, line 199: no need to spell out default field values

frouioui (approved):
Everything looks good to me, thank you for your contribution!

go/vt/vitessdriver/rows_test.go
```diff
@@ -135,3 +137,90 @@ func TestRows(t *testing.T) {
 
 	_ = ri.Close()
 }
+
+func TestColumnTypeScanType(t *testing.T) {
+	var r = sqltypes.Result{
+		Fields: []*querypb.Field{
+			{
```
frouioui, line 144: Can you please add a comment that describe what the expectations of the test is? This falls under our review item:

> If a test is added or modified, there should be a documentation on top of the test to explain what the expected behavior is what the test does.

johanoskarsson: Done!

harshit-gangal (approved):

johanoskarsson:
I assume the test failure is unrelated to this PR?

harshit-gangal:
can you do a merge from the main and push?

johanoskarsson:
Done, thanks!

PR #12009, opened 2022-12-22
Fix parsing and normalization of floating point types

### LINKED ISSUE #12008: Bug Report: Failure to parse create table with float with length
### Overview of the Issue

MySQL allows the following syntax to create a table:

```sql
create table a (a float(25))
```

But the Vitess parser doesn't allow the length notation. This is only allowed for `float`, not for double. 

See also https://dev.mysql.com/doc/refman/8.0/en/floating-point-types.html, but the documentation is wrong since with `float(24)` it's still a `float`, but only with `float(25)` does it become a `double`. 

### Reproduction Steps

```sql
create table a (a float(25))
```

### Binary Version

```sh
All versions
```

```sh
n/a
```

### Log Fragments

_No response_

dbussink:
Parsing floating point type definitions does not match MySQL. We fail to handle the length attribute that is allowed for `float`. 

This also fixes a number of other inconsistencies, like we don't allow `float4` (an alias for `float`) or `float8` (an alias for `double`). Lastly, we also want to normalize this in `schemadiff` since creating a table with this syntax works, but the resulting `show create table` has been normalized to either `float` or `double` depending on the size.  

## Related Issue(s)

Fixes #12008

## Checklist


systay (approved):

go/vt/sqlparser/sql.y
```diff
@@ -2202,10 +2202,28 @@ length_opt:
...
   }
+| '(' INTEGRAL ')'
+  {
+    $$ = LengthScaleOption{
+        Length: NewIntLiteral($2),
+    }
+  }
 | '(' INTEGRAL ',' INTEGRAL ')'
   {
     $$ = LengthScaleOption{
```
GuptaManan100, line 2229: nit: this can be slightly cleaned up to use the same newly created `double_length_opt`, like - 

```
float_length_opt:
double_length_opt
   {
     $$ = $1
   }
| '(' INTEGRAL ')'
   {
     $$ = LengthScaleOption{
         Length: NewIntLiteral($2),
     }
   }
```

GuptaManan100 (approved):
Rest LGTM!

go/vt/sqlparser/sql.y
GuptaManan100: I don't know if this will cause a shift-reduce conflict, but I don't expect that to be the case.

Issue #12013, opened 2022-12-23
Online DDL: migration runs to completion even if postponed and before being reviewed


shlomi-noach:
## Description

https://github.com/vitessio/vitess/pull/11910 introduced the concept of "immediate" migrations and as part of that introduced a new `reviewed_timestamp` column. As a reminder:

- any new migration (submitted e.g. via `ApplySchema`) is `insert`ed into `_vt.schema_migrations` table
- and the very initial `state` for the migration is `queued`
- the function `reviewQueuedMigrations()` reviews those newly submitted migrations and classifies and backfills some details
- and https://github.com/vitessio/vitess/pull/11910 introduces a `reviewed_timestamp` column. Once said function classifies and backfills data, it sets `reviewed_timestamp` to current timestamp (changing from `NULL`)

The next function in the state machine, `scheduleNextMigration()`, reads migrations that are `queued` and checks which of them, if any, is eligible to execute.
The bug is that  `scheduleNextMigration()` reads just _any_ `queued` migration.  This can happen in any race condition at the right (wrong) time, or if `reviewQueuedMigrations()` bails with error mid through migrations, for whatever reason.

At any case, it shouldn't read just any migration. It should only read those that have been reviewed, ie have `reviewed_timestamp IS NOT NULL`.

PR #12014, opened 2022-12-23
OnlineDDL: scheduleNextMigration() to only read reviewed migrations

shlomi-noach:
## Description

Fixes https://github.com/vitessio/vitess/issues/12013 by adding `AND reviewed_timestamp IS NOT NULL` 

## Related Issue(s)

fixes https://github.com/vitessio/vitess/issues/12013

## Checklist

### DIFF
--- a/go/vt/vttablet/onlineddl/executor.go
+++ b/go/vt/vttablet/onlineddl/executor.go
@@ -2237,30 +2237,26 @@ func (e *Executor) reviewEmptyTableRevertMigrations(ctx context.Context, onlineD
 // Non immediate operations are:
 // - A gh-ost migration
 // - A vitess (vreplication) migration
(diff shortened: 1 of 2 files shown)


mattlord (approved):

shlomi-noach:
Refactored a bit; still the main important change here is the addition of `AND reviewed_timestamp IS NOT NULL`. There was another missing piece where `reviewImmediateOperations()` did not classify correctly `REVERT` migrations. This is now fixed with a bit of refactoring for better readability, so the change looks a bit bigger than it really is.

GuptaManan100 (approved):

dbussink:
@shlomi-noach Can we also add tests to cover this? Not sure if the race can be timed / triggered or with introducing some mocked error?

shlomi-noach:
@dbussink yeah, I was wondering the same. I don't have a good story for the race condition tests. Let me think about this some more.

shlomi-noach:
meanwhile merging without adding extra tests as I still don't have a good story to those.

Issue #11975, opened 2022-12-15
Experimental foreign keys tracking issue


shlomi-noach:
Tracking issue for some PRs experimenting with better/more foreign key support. Some known limitations to using foreign keys in Vitess include:

- Online DDL (see [this blog post](https://vitess.io/blog/2021-06-15-online-ddl-why-no-fk/) for details)
- Cascading (or `set null`) changes are not logged to the binary log, and are only ever known to the internal InnoDB engine layer. As result, you may not reliably track changes to a child table in VReplication, as it relies on the binary log.
  This is then implicitly yet another issue for Online DDL
- MySQL foreign keys are limited to the scope of a server, and cannot enforce constraints across shards in a multisharded environment

In light, or in spite of the above, we want to slowly explore options for better foreign key support. As this is an exploration, we may/will submit some PRs that are only tackling subsets of the problem. It it probable that most will eventually connect to at least provide an incomplete-but-better-than-nothing support.

Some aspects/areas that may/will be affected:

- `schemadiff` library
- Online DDL, `go/vt/schema`
- `vtctl ApplySchema`
- VTGate
- VReplication
- others

derekperkins:
Super excited for these improvements!

shlomi-noach:
RFC for query-serving support for foreign keys: https://github.com/vitessio/vitess/issues/12967

PR #12016, opened 2022-12-27
schemadiff: normalize `PRIMARY KEY` definition

shlomi-noach:
## Description

In this PR we normalize and validate `PRIMARY KEY` definitions:

### 1. Normalize and validate `CREATE TABLE`

This statement:
```sql
create table t (id int primary key)
```
is normalized into:
```sql
create table t (id int, primary key(id))
```

Table definition is validated to have no more than one `PRIMARY KEY`.

### 2. Validate `ADD COLUMN`

In this statement:
```sql
alter table t add column i int primary key
```

We validate that the table does not already have a `PRIMARY KEY`. Once applied, the table definition is normalized as per the above.

### 3. Validate `MODIFY COLUMN`

In this statement:
```sql
alter table t modify column i int primary key
```

We validate that the table either:

- does not already have a `PRIMARY KEY`, or
- already has a `PRIMARY KEY (i)`, ie no change to the `PRIMARY KEY` itself.

Once applied, the table definition is normalized as per the above.

### 4. Table diffs change

The diff of these two tables:
```sql
create table t (id int primary key);
create table t (id bigint primary key);
```
is:
```sql
alter table t modify column id bigint
```
Note that there is nothing to say that `id` is a `PRIMARY KEY` even though the original definitions included that information as part of column definition options. The reason is that internally the tables are normalized as per (1) above.

---

Several tests are added to validate all the above. Otherwise there's been a massive impact to unit tests because we used the `id int primary key` generously throughout the tests, and had to rewrite dozens of those.

## Related Issue(s)

tracking: #10203

This work is needed for better foreign key support in `schemadiff`: it needs to validate that a parent table referenced by a foreign key, has the appropriate indexes over the referenced columns.  It's then important to have the `PRIMARY KEY` normalized as this PR does. See https://github.com/vitessio/vitess/issues/11975 and https://github.com/vitessio/vitess/pull/11944

Otherwise, this is just the right way of normalizing a table. `SHOW CREATE TABLE` in MySQL always extracts `PRIMARY KEY` information outside the column definition.

## Checklist


go/vt/sqlparser/ast_funcs.go
```diff
@@ -106,13 +106,13 @@ type TableOption struct {
...
-	colKeyUnique
-	colKeyUniqueKey
-	colKey
+	ColKeyNone ColumnKeyOption = iota
+	ColKeyPrimary
+	ColKeySpatialKey
+	ColKeyFulltextKey
+	ColKeyUnique
+	ColKeyUniqueKey
+	ColKey
```
shlomi-noach, line 115: The change in `go/vt/sqlparser` is to make the above constants public, so that we can introspect a `ColumnDefinition` and see whether it has a specific key option.

dbussink (approved):

deepthi (approved):
LGTM

PR #12022, opened 2022-12-28
mysqlctl is a command-line client so remove server flags

mattlord:
## Description

In working on https://github.com/vitessio/website/pull/1322 I realized that `mysqlctl` is a command-line client tool, while `mysqlctld` is a gRPC server that receives requests from a `vttablet`.

Most of the server flags were already removed from `mysqlctl` as part of the work to move from go/flag to spf13/pflag in v15. This PR removes the unnecessary ones that remained.

## Checklist

-   [x] Documentation was updated in: https://github.com/vitessio/website/pull/1322


GuptaManan100 (approved):
lgtm

GuptaManan100:
I don't think we need to go through a deprecation cycle, but should we specify this change in the release notes? If someone is using these flags, they should stop using them for this binary before upgrading.

frouioui (approved):
I agree with @GuptaManan100, let's mention this removal in the release notes.

doc/releasenotes/16_0_0_summary.md
```diff
@@ -40,6 +40,9 @@ In [PR #11097](https://github.com/vitessio/vitess/pull/11097) we introduced nati
 Orchestrator integration in `vttablet` was deprecated in the previous release and is deleted in this release.
 Consider using `VTOrc` instead of `Orchestrator`.
 
+The [`mysqlctl` command-line client](https://vitess.io/docs/16.0/reference/programs/mysqlctl/) had some leftover (ignored) server flags after the [v15 pflag work](https://github.com/vitessio/enhancements/blob/main/veps/vep-4.md). Those unused flags have now been removed. If you are using any of the following flags with `mysqlctl` in your scripts or other tooling, they will need to be removed prior to upgrading to v16:
```
frouioui, line 43: Let's just had a sub-heading here to have a nice separation with the `Orchestrator Integration Deletion` sub-section. Otherwise looks good to me.

frouioui: For instance `#### mysqlctl CLI`

mattlord: Ah, right! Good catch. Will do.

rsajwani (commented):
LGTM

Issue #11969, opened 2022-12-15
Bug Report: VDiff2 --only_pks Does Not Work


mattlord:
### Overview of the Issue

The [`--only_pks` flag](https://vitess.io/docs/16.0/reference/vreplication/vdiff2/#--only_pks) is supposed to limit the columns in the diff reports to only the PK columns. With VDiff2, however, this flag has no impact.

```
$ vtctlclient -- VDiff --only_pks --format=json customer.commerce2customer | jq
{
  "t1": {
    "ProcessedRows": 3,
    "MatchingRows": 2,
    "MismatchedRows": 1,
    "ExtraRowsSource": 0,
    "ExtraRowsSourceDiffs": null,
    "ExtraRowsTarget": 0,
    "ExtraRowsTargetDiffs": null,
    "MismatchedRowsSample": [
      {
        "Source": {
          "Row": {
            "c1": 2
          },
          "Query": ""
        },
        "Target": {
          "Row": {
            "c1": 2
          },
          "Query": ""
        }
      }
    ],
    "TableName": "t1"
  }
}
```

```
$ vtctlclient -- VDiff --v2 --only_pks --format=json customer.commerce2customer
{
	"UUID": "e60caa6e-7c29-11ed-ae82-920702940ee0"
}

$ vtctlclient -- VDiff --v2 --only_pks --format=json customer.commerce2customer show last
{
	"Workflow": "commerce2customer",
	"Keyspace": "customer",
	"State": "completed",
	"UUID": "e60caa6e-7c29-11ed-ae82-920702940ee0",
	"RowsCompared": 3,
	"HasMismatch": true,
	"Shards": "0",
	"StartedAt": "2022-12-15 03:38:11",
	"CompletedAt": "2022-12-15 03:38:11",
	"TableSummary": {
		"t1": {
			"TableName": "t1",
			"State": "completed",
			"RowsCompared": 3,
			"MatchingRows": 2,
			"MismatchedRows": 1,
			"ExtraRowsSource": 0,
			"ExtraRowsTarget": 0
		}
	},
	"Reports": {
		"t1": {
			"0": {
				"TableName": "t1",
				"ProcessedRows": 3,
				"MatchingRows": 2,
				"MismatchedRows": 1,
				"ExtraRowsSource": 0,
				"ExtraRowsTarget": 0,
				"MismatchedRowsSample": [
					{
						"Source": {
							"Row": {
								"c1": "2",
								"c2": "2"
							}
						},
						"Target": {
[...]

PR #11916, opened 2022-12-08
VDiff2: Migrate VDiff1 Unit Tests

mattlord:
## Description

This effectively ports over the [VDiff1 vtctld/wrangler unit tests](https://github.com/vitessio/vitess/blob/main/go/vt/wrangler/vdiff_test.go) to VDiff2 tablet/tabletmanager and vtctl unit tests, building on the unit test framework code added in https://github.com/vitessio/vitess/pull/11768.

## Related Issue(s)

  - Fixes: https://github.com/vitessio/vitess/issues/11151

## Checklist

-   [x] "Backport to:" labels have been added if this change should be back-ported
-   [x] Tests were added
-   [x] Documentation is not required

PR #12025, opened 2022-12-29
VDiff2: Properly Apply --only_pks Report Flag

mattlord:
## Description

This issue was discovered when working on [the VDiff2 unit tests](https://github.com/vitessio/vitess/pull/11916/files#diff-f767e045f3e0520a19788d6c09294faa7584ddb2927fcc377e8b232e429a55cfR207). The `--only_pks` VDiff client flag was not properly applied in VDiff2 (`--v2`). The specified client side flag [was not passed into the TableDiffer](https://github.com/vitessio/vitess/pull/12025/files#diff-42ed1dd669cda463ab1283c988f1c4b628f0159e33c1b8b0071a5688a11ddcd4L111-R111) which generated the row diffs when needed. This PR addresses that.

It also improves the protobuf field name. That change is safe as the old field will be ignored when unmarshalling the JSON into the struct, which is fine since it has no effect, and the new field will only affect *new* VDiffs where it would actually be used.

Lastly, I also took this opportunity to de-noise the verbose output of some related wrangler tests.

You can see a simple manual test in the issue: https://github.com/vitessio/vitess/issues/11969

## Related Issue(s)

  - Fixes: https://github.com/vitessio/vitess/issues/11969

## Checklist

-   [x] Tests were updated


deepthi (approved):

deepthi:
Are website docs for this already up to date?

mattlord:
> Are website docs for this already up to date?

Yep! The flag was documented for v2, the value simply wasn't used before this.

rohit-nayak-ps (approved):
lgtm

PR #12027, opened 2023-01-02
OnlineDDL: 'mysql' strategy, managed by the scheduler, but executed via normal MySQL statements

shlomi-noach:
## Description

This PR introduces a new [DDL strategy](https://vitess.io/docs/15.0/user-guides/schema-changes/ddl-strategies/): `mysql`. This strategy is a hybrid between `direct` (which is completely non-Online) and the various online strategies.

A migration submitted with `mysql` strategy is _managed_. That is, it gets a migration UUID. The scheduler queues it, reviews it, runs it. The user may cancel or retry it, much like all Online DDL migrations. The difference is that when the scheduler runs the migration via normal MySQL `CREATE/ALTER/DROP TABLE` statements.

The user may add `ALGORITHM=INPLACE` or `ALGORITHM=INSTANT` as they please, and the scheduler will attempt to run those as is. Some migrations will be completely blocking. See the [MySQL documentation](https://dev.mysql.com/doc/refman/8.0/en/innodb-online-ddl-operations.html). In particular, consider that for non-`INSTANT` DDLs, replicas will accumulate substantial lag.

This was requested and discussed by @derekperkins in a recent Vitess Open Hour call.

## Related Issue(s)

Tracking: #6926

## Checklist

### DIFF
--- a/go/test/endtoend/onlineddl/scheduler/onlineddl_scheduler_test.go
+++ b/go/test/endtoend/onlineddl/scheduler/onlineddl_scheduler_test.go
@@ -669,6 +669,47 @@ func TestSchemaChange(t *testing.T) {
 			})
 		})
 	}
+	// 'mysql' strategy
+	t.Run("mysql strategy", func(t *testing.T) {
+		t.Run("declarative", func(t *testing.T) {
+			t1uuid = testOnlineDDLStatement(t, createT1Statement, "mysql --declarative", "vtgate", "just-created", "", false)
+
+			status := onlineddl.WaitForMigrationStatus(t, &vtParams, shards, t1uuid, normalWaitTime, schema.OnlineDDLStatusComplete, schema.OnlineDDLStatusFailed)
+			fmt.Printf("# Migration status (for debug purposes): <%s>\n", status)
+			onlineddl.CheckMigrationStatus(t, &vtParams, shards, t1uuid, schema.OnlineDDLStatusComplete)
+			checkTable(t, t1Name, true)
+		})
+
+		t.Run("fail postpone-completion", func(t *testing.T) {
+			t1uuid := testOnlineDDLStatement(t, trivialAlterT1Statement, "mysql --postpone-completion", "vtgate", "", "", true)
+
+			// --postpone-completion not supported in mysql strategy
+			time.Sleep(ensureStateNotChangedTime)
+			onlineddl.WaitForMigrationStatus(t, &vtParams, shards, t1uuid, normalWaitTime, schema.OnlineDDLStatusFailed)
+			onlineddl.CheckMigrationStatus(t, &vtParams, shards, t1uuid, schema.OnlineDDLStatusFailed)
+		})
+		t.Run("trivial", func(t *testing.T) {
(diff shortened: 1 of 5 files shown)


go/test/endtoend/onlineddl/scheduler/onlineddl_scheduler_test.go
```diff
@@ -669,6 +669,47 @@ func TestSchemaChange(t *testing.T) {
 			})
 		})
 	}
+	// 'mysql' strategy
+	t.Run("mysql strategy", func(t *testing.T) {
+		t.Run("declarative", func(t *testing.T) {
+			t1uuid = testOnlineDDLStatement(t, createT1Statement, "mysql --declarative", "vtgate", "just-created", "", false)
+
+			status := onlineddl.WaitForMigrationStatus(t, &vtParams, shards, t1uuid, normalWaitTime, schema.OnlineDDLStatusComplete, schema.OnlineDDLStatusFailed)
```
mattlord, line 677: Curious why we specify multiple statuses in `WaitForMigrationStatus` but then only a single one in `CheckMigrationStatus`?

mattlord (approved):

go/test/endtoend/onlineddl/scheduler/onlineddl_scheduler_test.go
shlomi-noach: The reason is that we wait till either status is seen (the function will otherwise give some `30sec` for the migration to "evolve"). So there are terminal states for the function, so that it does not need to wait after meeting these states. This just reduces test time. Then, we really want to validate that we reached a particular result: typically we look for a success, ie `OnlineDDLStatusComplete`. But sometimes we intentionally seek a failure.

derekperkins (approved):
This is awesome, thanks for pushing this through!

deepthi (commented):
This is great. I like how you keep improving online DDL to take advantage of upstream MySQL improvements.
I think this is significant enough to either warrant a new issue altogether, or at least a specific comment in the tracking issue as to why we want to introduce this option. Currently there's nothing in the tracking issue that tells us anything about why this PR is being proposed.
Also website docs 😄

shlomi-noach:
> This is great. I like how you keep improving online DDL to take advantage of upstream MySQL improvements.
I think this is significant enough to either warrant a new issue altogether, or at least a specific comment in the tracking issue as to why we want to introduce this option. Currently there's nothing in the tracking issue that tells us anything about why this PR is being proposed.

:+1: 

> Also website docs smile

:+1: 

Let me first create all the necessary docs before we merge this PR

shlomi-noach:
Website docs: https://github.com/vitessio/website/pull/1342

shlomi-noach:
Fixes https://github.com/vitessio/vitess/issues/12075

I've put https://github.com/vitessio/vitess/issues/12075 up to explain why we're doing this.

shlomi-noach:
@deepthi for final approval + merge, as there's now a tracking issue as well as website docs

deepthi (approved):

PR #12028, opened 2023-01-02
*WIP* schemadiff: `Diff()` of schemas ordered by valid dependencies

shlomi-noach:
## Description

This PR extends https://github.com/vitessio/vitess/pull/12026, which is still unmerged. [Diff here](https://github.com/planetscale/vitess/compare/schemadiff-fk-keys...planetscale:vitess:schemadiff-fk-diff-order).

A `schemadiff`'s `Schema` is normalized to have tables in foreign key dependency order (parents come first, children later), as well as in view dependency order (tables first, then views that only rely on tables, then 2nd level views etc.). This makes it possible to create, or apply, the schema by iterating its entities in order.

With this PR, we also sort the `Diff()` of two schemas, such that the diff can be applied in-order to a schema and without error. At this time, the diff is sorted like so:

- `CREATE TABLE` follow the same ordering for schema creation, ie. foreign key parents come first, then their children, and recursively onwards.
- `CREATE TABLE` comes before `CREATE VIEW`, again following same logic as illustrated above.
- `DROP VIEW` uses reverse order of creation, such that we first drop the views with most dependencies, then those with fewest dependencies
- `DROP TABLE` uses reverse order of creation, such that we first `DROP` a foreign key relationship's child before `DROP`ping its parent.

There is still work to be done where it comes to modifying foreign key constraints in existing tables.

## Related Issue(s)

- Foreign keys tracking issue: https://github.com/vitessio/vitess/issues/11975
- schemadiff tracking issue: https://github.com/vitessio/vitess/issues/10203
- https://github.com/vitessio/vitess/pull/12026
- 
## Checklist


shlomi-noach:
Permalink for diff: https://github.com/planetscale/vitess/compare/211f85d63c99f38e24fe24dc6a7d8e57a5fbcc2f...planetscale:vitess:schemadiff-fk-diff-order

shlomi-noach:
The only relevant commit is https://github.com/vitessio/vitess/pull/12028/commits/66dc1ae117298e0e7c39a8ab8242446980458d66

shlomi-noach:
Replaced by https://github.com/vitessio/vitess/pull/12184 ; was simpler to create a new branch/PR than to resolve all conflicts.

PR #12030, opened 2023-01-03
[vtctldclient|docs] apply doc feedback based on website PR feedback

ajm188:
## Description

Applying @mattlord's feedback from https://github.com/vitessio/website/pull/1177

## Related Issue(s)

https://github.com/vitessio/website/pull/1177 

## Checklist

### DIFF
--- a/go/cmd/vtctldclient/command/shard_routing_rules.go
+++ b/go/cmd/vtctldclient/command/shard_routing_rules.go
@@ -35,15 +35,20 @@ var (
 	// ApplyShardRoutingRules makes an ApplyShardRoutingRules gRPC call to a vtctld.
 	ApplyShardRoutingRules = &cobra.Command{
 		Use:                   "ApplyShardRoutingRules {--rules RULES | --rules-file RULES_FILE} [--cells=c1,c2,...] [--skip-rebuild] [--dry-run]",
-		Short:                 "Applies VSchema shard routing rules.",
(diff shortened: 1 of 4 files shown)


go/cmd/vtctldclient/command/shard_routing_rules.go
```diff
@@ -35,15 +35,20 @@ var (
...
 	GetShardRoutingRules = &cobra.Command{
-		Use:                   "GetShardRoutingRules",
-		Short:                 "Displays VSchema shard routing rules.",
+		Use:   "GetShardRoutingRules",
+		Short: "Displays the currently active shard routing rules as a JSON document.",
+		Long: `Displays the currently active shard routing rules as a JSON document.
+
+See the documentation on shard level migrations[1] for more information.
+
+[1]: https://vitess.io/docs/16.0/reference/vreplication/shardlevelmigrations/`,
```
mattlord, line 51: Nice! IMO we should remove the version though, which will redirect to the current production release: https://vitess.io/docs/reference/vreplication/shardlevelmigrations/

mattlord (approved):
I think we should update the link as noted but otherwise LGTM! ❤️

go/cmd/vtctldclient/command/shard_routing_rules.go
ajm188: oh nice

Issue #11665, opened 2022-11-08
Bug Report: Unable to run ERS from VTAdmin UI


GuptaManan100:
### Overview of the Issue

Using the docker images for `v15.0.0.0:latest`, we can't run ERS from the VTAdmin UI. An error is noticed like 
![Screenshot 2022-11-08 at 9 00 13 PM](https://user-images.githubusercontent.com/35839558/200607618-e4cbd819-8027-4738-94ae-e069d3d1750d.png)

### Reproduction Steps

Run the `vtorc-vtadmin-test` in the VTOp repository. Try running ERS from the VTAdmin UI and see it fail.

### Binary Version

```sh
v15 and maybe main (needs to be tested)
```

### Operating System and Environment details

```sh
Docker images
```

GuptaManan100:
I initially thought that the issue might be because of `nginx` vs `npm serve`. In the operator we use `nginx` to run the web of VTAdmin, but we use `npm serve` in the local example.
I have tested this hypothesis and `nginx` is not the issue. I tried running the local examples with `nginx` and it works.

GuptaManan100:
Another issue could be that `npm build` might not be agnostic to the OS version it is being built on. The SHA for the `main....js` file built are actually different locally and in the docker image. This could be an issue because we run `npm build` in the image from `node` and then we create a new layer using `nginx` and copy the build files. Maybe they aren't compatible 🤷‍♂️ 
Anyways, I am out of my depth now. Maybe @ajm188 knows?

GuptaManan100:
The reason that the second rbac doesn't work is because we renamed the action from `emergency_reparent_shard` to `emergency_failover_shard`. The rbac files in the top have to be changed to reflect this change.

GuptaManan100:
Fix for VTop - https://github.com/planetscale/vitess-operator/pull/352

PR #12034, opened 2023-01-03
Fix rbac config in the vtop example

GuptaManan100:
## Description

This PR fixes the rbac config in the vtop example as pointed out in #11665 

## Related Issue(s)

- #11665 

## Checklist

### DIFF
--- a/examples/operator/101_initial_cluster.yaml
+++ b/examples/operator/101_initial_cluster.yaml
@@ -236,8 +236,8 @@ stringData:
       clusters: ["*"]
     - resource: "Shard"
       actions:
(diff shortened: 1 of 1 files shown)


deepthi (approved):

GuptaManan100:
We don't run the vitess operator test at all. When we do run it we can consider doing a PRS or ERS.

frouioui:
Should this PR close https://github.com/vitessio/vitess/issues/11665?

frouioui (approved):

GuptaManan100:
I have to fix the test on vtop too. I'll close it after that PR is merged. I am working on it now.

Issue #9518, opened 2022-01-16
RFC: Deprecate MariaDB Support in 2022 and Remove in 2023


mattlord:
### Proposal

Deprecate [official MariaDB support](https://vitess.io/docs/overview/supported-databases/) in Vitess 14.0 (~June 2022) and remove it entirely in 16.0 (~Feb 2023). This gives Vitess users approximately 2 years where they can continue to use Vitess with MariaDB and offers adequate time to transition to MySQL 8 ([MySQL 5.7 is currently scheduled for EOL in 2023](https://endoflife.software/applications/databases/mysql)).
### Reasoning

> Note: MySQL and Percona Server for MySQL are considered fully equivalent in Vitess.

The reasons for proposing this are as follows:
-  [Vitess only officially supports MariaDB 10.0-10.3 today](https://vitess.io/docs/overview/supported-databases/)
   -  MariaDB 10.3 is already nearly 4 years old ([10.7 is the latest](https://mariadb.com/kb/en/mariadb-1080-release-notes/)) and [MariaDB 10.3 will be EOL next year (2023)](https://endoflife.date/mariadb)
- MariaDB is a hard fork of MySQL and 10.X is a very different database than MySQL 8.X (which is quickly becoming the default with Vitess)
  - Even though MariaDB has often been described as a "drop in replacement" for MySQL, this is not true today with MariaDB 10.7 and MySQL 8.0
  - MySQL and MariaDB are now properly treated as independent database products in various places and in more and more distros (e.g. Ubuntu which supports MySQL 8.0) as the two diverge more and more with each passing year; e.g. at [Amazon Web Services](https://aws.amazon.com/rds/) where they support both MySQL and MariaDB separately in RDS, at [Google Cloud Services](https://cloud.google.com/) where MariaDB is not supported in CloudSQL, along with other places like [DB-Engines](https://db-engines.com/en/ranking) and [among domain experts when comparing the two at a technical level](https://twitter.com/perconabytes/status/1489932735506534400) regarding [modern SQL support](https://modern-sql.com/blog/2018-08/whats-new-in-mariadb-10.3) or [overall compatibility](http://www.tusacentral.net/joomla/index.php/mysql-blogs/232-who-is-drop-in-replacement-of)
[...]

PR #12040, opened 2023-01-04
Remove building Docker containers with MariaDB

dbussink:
In https://github.com/vitessio/vitess/issues/9518 MariaDB support was deprecated in v14 and is being EOL'd in v16, which means that at this point for the future Vitess v16 we don't need to publish containers with MariaDB anymore.

## Related Issue(s)

In https://github.com/vitessio/vitess/issues/9518 we deprecated MariaDB support in v14 and are EOLing it in v16 so we should stop publishing artifacts with it in v16 as well. 

## Checklist


mattlord:
I personally like it. I intentionally left these parts alone in https://github.com/vitessio/vitess/pull/12036 as it was not directly related to the issue being addressed in that PR and more importantly I was not sure what knock on effects this might have or what other related work may be needed. I'll let @frouioui and @deepthi comment on that and let them formally approve.

frouioui (approved):

frouioui:
Once this Pull Request is merged we will remove the auto build for `mariadb` and `mariadb103` Docker tags.

deepthi (commented):
What is the reason to remove the alpine docker file? Did you look at the history to see why it was added?

mattlord (approved):

PR #12042, opened 2023-01-05
BugFix: Fix race in `IsFlagProvided`

### LINKED ISSUE #12041: Bug Report: data race in `IsFlagProvided` function.
### Overview of the Issue

The `IsFlagProvided` function introduced in https://github.com/vitessio/vitess/pull/11881 seems to be running into a data race.

The reason is that `Visit` function is uses isn’t thread-safe -
```golang
if len(f.actual) != len(f.sortedActual) {
	f.sortedActual = sortFlags(f.actual)
}
flags = f.sortedActual
```
So if two go-routines call this same function, one can be reading `sortedActual` in line 1 and the other could be writing to it in the next line.

### Reproduction Steps

Run unit_race over and over until this fails.

### Binary Version

```sh
main
```

```sh
main
```

### Log Fragments

_No response_

GuptaManan100:
## Description

This PR fixes the race in `IsFlagProvided` as described in issue #12041. 

The way the fix is accomplished is to replace the usage of `flag.Visit` with `flag.Lookup` which just reads from a single map and doesn't write anything, making it thread-safe.

## Related Issue(s)

- Fixes #12041 

## Checklist

### DIFF
--- a/go/internal/flag/flag.go
+++ b/go/internal/flag/flag.go
@@ -72,13 +72,11 @@ func Parse(fs *flag.FlagSet) {
 
 // IsFlagProvided returns if the given flag has been provided by the user explicitly or not
 func IsFlagProvided(name string) bool {
-	found := false
-	flag.Visit(func(f *flag.Flag) {
-		if f.Name == name {
-			found = true
-		}
-	})
-	return found
+	fl := flag.Lookup(name)
+	if fl != nil {
+		return fl.Changed
+	}
+	return false
 }
 
 // TrickGlog tricks glog into understanding that flags have been parsed.
--- a/go/vt/topo/locks_test.go
+++ b/go/vt/topo/locks_test.go
@@ -58,11 +58,27 @@ func TestGetLockTimeout(t *testing.T) {
 			lockTimeoutValue:            "33s",
 			remoteOperationTimeoutValue: "22s",
 			expectedLockTimeout:         33 * time.Second,
+		}, {
+			description:                 "remote operation timeout flag specified to the default",
+			lockTimeoutValue:            "",
+			remoteOperationTimeoutValue: "15s",
+			expectedLockTimeout:         15 * time.Second,
+		}, {
+			description:                 "lock-timeout flag specified to the default",
+			lockTimeoutValue:            "45s",
+			remoteOperationTimeoutValue: "33s",
+			expectedLockTimeout:         45 * time.Second,
 		},
 	}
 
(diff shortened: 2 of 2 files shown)


go/internal/flag/flag.go
```diff
@@ -72,13 +72,11 @@ func Parse(fs *flag.FlagSet) {
...
-	found := false
-	flag.Visit(func(f *flag.Flag) {
-		if f.Name == name {
-			found = true
-		}
-	})
-	return found
+	fl := flag.Lookup(name)
+	if fl != nil {
+		return fl.Changed
```
dbussink, line 77: What are the semantics here if someone did provide a flag but gave it the default value? Does `fl.Changed` return `true` or `false` in that case? 

And if that's different from before, does that matter or not?

GuptaManan100: `fl.Changed` is `true` even when you pass a flag with the default value. That semantic is very important for us because we want to know if a flag is explicitly passed or not. We don't care if it is the default value. I had the question in mind when I implemented this, therefore I have added 2 more tests in this PR which test this scenario of passing a flag explicitly equal to its default value.

GuptaManan100: As an extra check, I had also changed the `vtctld` binary locally to print the lock timeout value and here are the results - 
```
~/vitess   flag-lookup-race ±  vtctld
E0105 12:53:08.941311   51261 main.go:55] lock-timeout value - 45s
~/vitess   flag-lookup-race ±  vtctld --remote_operation_timeout=20s
E0105 12:53:19.659368   51292 main.go:55] lock-timeout value - 20s
~/vitess   flag-lookup-race ±  vtctld --remote_operation_timeout=20s --lock-timeout=32s
E0105 12:53:27.063374   51321 main.go:55] lock-timeout value - 32s
~/vitess   flag-lookup-race ±  vtctld --remote_operation_timeout=20s --lock-timeout=45s
E0105 12:53:30.086307   51350 main.go:55] lock-timeout value - 45s
~/vitess   flag-lookup-race ±  vtctld --lock-timeout=45s
E0105 12:53:43.704695   51381 main.go:55] lock-timeout value - 45s
~/vitess   flag-lookup-race ±  vtctld --remote_operation_timeout=15s
E0105 12:53:56.462829   51412 main.go:55] lock-timeout value - 15s
```
This function works exactly as before.

dbussink (approved):

deepthi (approved):

PR #12044, opened 2023-01-05
Keywords must be unique and can't be reused

### LINKED ISSUE #12046: Bug Report: parsing reuses keywords
### Overview of the Issue

In https://github.com/vitessio/vitess/pull/12009 I added aliases for float4 / float8 but that was done wrongly. We can't alias these with the same keyword, or we break functions like SQLTypeToQueryType because we only have one version of it in the keywords list.

### Reproduction Steps

Parse a table with a `double` type and then check the inferred type. 

### Binary Version

```sh
On main since https://github.com/vitessio/vitess/pull/12009
```

```sh
n/a
```

### Log Fragments

_No response_

dbussink:
In https://github.com/vitessio/vitess/pull/12009 I added aliases for float4 / float8 but that was done wrongly. We can't alias these with the same keyword, or we break functions like SQLTypeToQueryType because we only have one version of it in the keywords list.

## Related Issue(s)

This was broken in  https://github.com/vitessio/vitess/pull/12009 but that's not in any released version yet.

Fixes #12046

## Checklist


GuptaManan100 (approved):
You need to add an issue for this PR since it's a bug fix. Otherwise LGTM.

systay (approved):

Issue #12049, opened 2023-01-06
Bug Report: VTGate does not set immediate caller id from gRPC static auth username


brendar:
### Overview of the Issue

When using gRPC static auth on VTGate, the immediate caller id is always set to `unsecure_grpc_client`, rather than the username provided by the client. This means that table ACLs on VTTablet will reject the client's queries.

### Reproduction Steps

1. Enable gRPC static auth on VTGate
   * Modify `examples/local/scripts/vtgate-up.sh` to add args
      * `--grpc_auth_mode static`
      * `--grpc_auth_static_password_file grpc_static_auth.json`
3. Enable table ACLs on VTTablet
   * Create file `examples/local/table_acl.json`
      ```json
      {
        "table_groups": [
          {
            "name": "default",
            "table_names_or_prefixes": ["%"],
            "readers": ["vitess"],
            "writers": ["vitess"],
            "admins": ["vitess"]
          }
        ]
      }
      ```
   * Modify `examples/local/scripts/vttablet-up.sh` to add args
      * `--enforce-tableacl-config`
      * `--queryserver-config-strict-table-acl`
      * `--table-acl-config table_acl.json`
3. Run `101_initial_cluster.sh`
4. Attempt to query a table using [grpcurl](https://github.com/fullstorydev/grpcurl)
   * `grpcurl -plaintext -H='username:vitess' -H='password:vitess_password' -d '{"query": { "sql": "SELECT * FROM commerce.customer" }}' localhost:15991 vtgateservice.Vitess.Execute`
   * Result:
     ```
      {
        "error": {
          "message": "target: commerce.0.primary: vttablet: rpc error: code = PermissionDenied desc = Select command denied to user 'unsecure_grpc_client' for table 'customer' (ACL check error) (CallerID: unsecure_grpc_client)",
          "code": "PERMISSION_DENIED"
        },
        "session": {
          "autocommit": true,
          "rowCount": "-1"
        }
      }
     ```

### Binary Version

```sh
main
```

### Operating System and Environment details

```sh
Not applicable
```

PR #3418, opened 2017-11-27
gRPC Auth Plugins

rafael:
### Description

* The following PR adds the capability of implementing Auth plugins on top of gRPC for Vitess.  The core of this feature uses gRPC interceptors (StreamInterceptor, UnaryInterceptor) to extract from each gRPC request the metadata needed to authenticate/authorize the request. Inspiration comes from [here](https://grpc.io/docs/guides/auth.html#extending-grpc-to-support-other-authentication-mechanisms). 
* Users which would like to add different kind of auth between the gRPC components of Vitess can implement these plugins on demand. 
* This PR also adds a basic auth plugin. With this plugin, each gRPC component can be started with the following parameters: 
 
  1. On the server side:

  ```
  vttablet ... -grpc_auth_mode static -grpc_auth_static_password_file /path/to/creds/grpc_static_auth.json
  ```
  2. On the clients:
  ```
  vtctlclient -grpc_auth_static_client_creds /path/to/creds/grpc_static_client_auth.json
  ```
  With this, servers will enforce clients to have a valid username/pass in order to make gRPC requests. 
 
* In the future, we will expand this work to have an approach that includes both `Tables` and `gRPC ` ACL's.

### Outstanding Items

- [x]  At the moment, there is a limitation in python gRPC implementation. It is not matching the golang one and it enforces to use TLS when using `WithPerRPCCredentials`. This creates a problem in the context of Vitess where we don't want to force our users to use TLS when using an auth plugin. This issue is being in discussed in the gRPC python mailing [list](https://groups.google.com/forum/#!topic/grpc-io/2KvZEaNRCmA).  One option that we are thinking is to use something similar to [Header](https://godoc.org/google.golang.org/grpc#Header) to pass these kind of credentials, however this doesn't seem to be supported in python neither.  ***BLOCKER***
- [x] Clean up python client depending on resolution on issue above. 
- [ ] Add tests for each of the go clients to assert basic auth plugin is working.

PR #12050, opened 2023-01-06
VTGate: Set immediate caller id from gRPC static auth username

brendar:
Signed-off-by: Brendan Dougherty <brendan.dougherty@shopify.com>

## Description

Currently, when using gRPC static auth on VTGate, the immediate caller id is always set to `unsecure_grpc_client`, rather than the username provided by the client. This means that table ACLs on VTTablet will reject the client's queries.

The problem appears to have been discussed when static auth was originally added [here](https://github.com/vitessio/vitess/pull/3418#discussion_r154534073), but was not changed in a followup.

This PR takes a similar approach to what was suggested and sets the authenticated username in the Context. This allows the VTGate server to retrieve the username in a similar manner to how it retrieves the certificate common name as the immediate caller id when using mTLS ([source](https://github.com/vitessio/vitess/blob/371be9e919adc7c3c277834974fb666656775f5d/go/vt/vtgate/grpcvtgateservice/server.go#L69-L95)).

## Related Issue(s)

Fixes https://github.com/vitessio/vitess/issues/12049

## Checklist

### DIFF
--- a/go/test/endtoend/vtgate/grpc_server_auth_static/main_test.go
+++ b/go/test/endtoend/vtgate/grpc_server_auth_static/main_test.go
@@ -1,0 +1,212 @@
+/*
+Copyright 2023 The Vitess Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package grpcserverauthstatic
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"os"
+	"path"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"google.golang.org/grpc"
+
+	"vitess.io/vitess/go/test/endtoend/cluster"
+	"vitess.io/vitess/go/vt/grpcclient"
+	"vitess.io/vitess/go/vt/vtgate/grpcvtgateconn"
+	"vitess.io/vitess/go/vt/vtgate/vtgateconn"
+)
+
+var (
+	clusterInstance   *cluster.LocalProcessCluster
+	vtgateGrpcAddress string
+	hostname          = "localhost"
+	keyspaceName      = "ks"
+	cell              = "zone1"
+	sqlSchema         = `
+		create table test_table (
+			id bigint,
+			val varchar(128),
+			primary key(id)
+		) Engine=InnoDB;
+`
+	grpcServerAuthStaticJSON = `
+		[
+		  {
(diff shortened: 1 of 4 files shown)


test/config.json
```diff
@@ -900,6 +900,15 @@
 			"RetryMax": 1,
 			"Tags": []
 		},
+		"vtgate_grpc_server_auth_static": {
+			"File": "unused.go",
+			"Args": ["vitess.io/vitess/go/test/endtoend/vtgate/grpc_server_auth_static"],
+			"Command": [],
+			"Manual": false,
+			"Shard": "vtgate_general_heavy",
```
brendar, line 908: I'm not sure if this is the right shard to use. Any guidance would be appreciated.

go/test/endtoend/vtgate/grpc_server_auth_static/main_test.go
```diff
@@ -0,0 +1,212 @@
...
+		}
+		vtgateGrpcAddress = fmt.Sprintf("%s:%d", clusterInstance.Hostname, clusterInstance.VtgateGrpcPort)
+
+		return m.Run()
+	}()
+	os.Exit(exitcode)
+}
+
+func TestAuthenticatedUserWithAccess(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
```
GuptaManan100, line 143: To me the tests feel self-explanatory, but could you add a couple of lines to each describing what its testing.

GuptaManan100 (approved):
Other than this, I like the changes and the tests too.

test/config.json
GuptaManan100: This shard looks as good as any. I checked the run of this shard as well. It takes about 13 minutes with this new test, so there is no risk of flakiness because of timeouts. It should be okay.

go/test/endtoend/vtgate/grpc_server_auth_static/main_test.go
brendar: Done

Phanatic (approved):
👋🏾 I work at PlanetScale and built out our Service-to-service authentication story and custom ACLs for PlanetScale databases. 

PlanetScale's usage of vtgate static auth and ACLs system is a bit unique in that we don't handle authentication for user queries in vtgate. We do that in our own query front-end service. 
VTGate static authentication is configured for service-to-service authentication between the query front-end and vtgate, ACL system is configured to pass through user roles from our credential store to vtgate.  

We do this by setting the effectiveCallerID on requests made to the vtgate gRPC service and having the same names reflected in the acl config file for a given vttablet. 

https://github.com/vitessio/vitess/blob/c0b5df6f1fb245b3e5ce56d9dda6415eaa2939f6/go/vt/vtgate/grpcvtgateservice/server.go#L107-L113

Given ☝🏾 , this change looks good to me. Since the caller identity is _still_ the EffectiveCallerId from the context and not request identity used for service-to-service communication.

brendar:
Thank you both for the reviews. Is there anything else needed or would it be possible for you to merge the PR?

harshit-gangal (approved):

deepthi:
@brendar we are planning to effectively revert this PR and backport the change to release-16.0 branch as well. See discussion https://github.com/vitessio/vitess/pull/12961#issuecomment-1521550337

PR #12057, opened 2023-01-09
stats/prometheus: normalize labels for single-label implementations

### LINKED ISSUE #12058: Bug Report: Prometheus metrics reporting doesn't snake case single label setups
### Overview of the Issue

Prometheus metrics are normally emitted using `snake_case` for labels. This doesn't happen though for the single label cases. 

### Reproduction Steps

See the following code for example:

```go
 	m1 := stats.NewCountersWithSingleLabel("ThisIsMetric1", "helpstring1", "ThisIsALabel")
 	m1.Add("labelvalue1", 420)

 	m2 := stats.NewCountersWithMultiLabels("ThisIsMetric2", "helpstring2", []string{"ThisIsALabel"})
 	m2.Add([]string{"labelvalue2"}, 420)
```

The expectation is that both metrics emit the same labels, namely `this_is_a_label` in `snake_case`. Today they create two different labels, the `stats.NewCountersWithMultiLabels` creates `snake_case` but `stats.NewCountersWithSingleLabel` creates `CamelCase`.

### Binary Version

```sh
n/a
```

```sh
n/a
```

### Log Fragments

_No response_

vmg:
## Description

Here's a bug we've found in production: when exporting Prometheus metrics, the name of the labels is not consistent when using Single-Label metrics vs Multi-Label metrics. More precisely, the Single-Label metrics API does not perform label normalization (i.e. from CamelCase to snake_case), while the Multi-Label API does perform this normalization.

The normalization into snake_case *is* sensible, as metric names in Prometheus are idiomatically written in snake case, so the lack of normalization when using single-label metrics feels like an oversight that caused our Prometheus metrics exports to report the wrong metrics.

After this change, the behavior between using a Single-Label metric and a Multi-Label metric with just one label is consistent. This is, however, a breaking change so it should be noted in the release notes.

## Related Issue(s)

Fixes #12058

## Checklist

### DIFF
--- a/go/stats/prometheusbackend/prometheusbackend_test.go
+++ b/go/stats/prometheusbackend/prometheusbackend_test.go
@@ -357,6 +357,26 @@ func testMetricsHandler(t *testing.T) *httptest.ResponseRecorder {
 	return response
 }
 
+func TestPrometheusLabels(t *testing.T) {
+	m1 := stats.NewCountersWithSingleLabel("ThisIsMetric1", "helpstring1", "ThisIsALabel")
+	m1.Add("labelvalue1", 420)
+
+	m2 := stats.NewCountersWithMultiLabels("ThisIsMetric2", "helpstring2", []string{"ThisIsALabel"})
+	m2.Add([]string{"labelvalue2"}, 420)
+
+	response := testMetricsHandler(t)
+
+	expect := []string{
+		"namespace_this_is_metric1{this_is_a_label=\"labelvalue1\"} 420",
+		"namespace_this_is_metric2{this_is_a_label=\"labelvalue2\"} 420",
+	}
+	for _, line := range expect {
+		if !strings.Contains(response.Body.String(), line) {
+			t.Fatalf("Expected result to contain %s, got %s", line, response.Body.String())
+		}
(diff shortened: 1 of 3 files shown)


dbussink (approved):

go/stats/prometheusbackend/prometheusbackend_test.go
```diff
@@ -357,6 +357,26 @@ func testMetricsHandler(t *testing.T) *httptest.ResponseRecorder {
...
+
+	response := testMetricsHandler(t)
+
+	expect := []string{
+		"namespace_this_is_metric1{this_is_a_label=\"labelvalue1\"} 420",
+		"namespace_this_is_metric2{this_is_a_label=\"labelvalue2\"} 420",
+	}
+	for _, line := range expect {
+		if !strings.Contains(response.Body.String(), line) {
+			t.Fatalf("Expected result to contain %s, got %s", line, response.Body.String())
```
deepthi, line 375: Ideally we would use `assert/require` but I see that the whole file is currently using `t.Fatalf` 😞

deepthi (approved):

PR #12059, opened 2023-01-09
Refactor sqlparser.Rewrite uses

systay:
## Description
Refactor of sqlparser.Rewrite usages.

Doing cursor.Replace() on the down path is not great.
It makes it difficult to understand what happens after - do we visit the children of the old AST object or the new one?

This change moves most uses of Rewrite so cursor.Replace() happens on the up path.
In a separate PR, I'll change the API so it's not possible to do replace on the down path.

## Checklist

### DIFF
--- a/go/vt/sqlparser/rewriter_api.go
+++ b/go/vt/sqlparser/rewriter_api.go
@@ -51,6 +51,18 @@ func Rewrite(node SQLNode, pre, post ApplyFunc) (result SQLNode) {
 	return parent.SQLNode
 }
 
+// SafeRewrite does not allow replacing nodes on the down walk of the tree walking
+// Long term this is the only Rewrite functionality we want
+func SafeRewrite(node SQLNode, down func(SQLNode) bool, up ApplyFunc) (result SQLNode) {
+	var pre func(cursor *Cursor) bool
+	if down != nil {
+		pre = func(cursor *Cursor) bool {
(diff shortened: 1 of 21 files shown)


go/vt/sqlparser/rewriter_api.go
```diff
@@ -51,6 +51,18 @@ func Rewrite(node SQLNode, pre, post ApplyFunc) (result SQLNode) {
...
 
+// SafeRewrite does not allow replacing nodes on the down walk of the tree walking
+// Long term this is the only Rewrite functionality we want
+func SafeRewrite(node SQLNode, down func(SQLNode) bool, up ApplyFunc) (result SQLNode) {
+	var pre func(cursor *Cursor) bool
+	if down != nil {
+		pre = func(cursor *Cursor) bool {
+			return down(cursor.Node())
+		}
+	}
```
GuptaManan100, line 62: What does this do? I don't follow how it prevents replacing nodes.

GuptaManan100 (approved):
Rest LGTM

frouioui (approved):

go/vt/sqlparser/rewriter_api.go
systay: The `down` function does not get a cursor, and you can only replace using `cursor.Replace`

PR #12055, opened 2023-01-09
consolidating OnlineDDL 'singleton' tests into 'scheduler' tests: part 1

shlomi-noach:
## Description

Following up on the discussion on https://github.com/vitessio/vitess/pull/11988#discussion_r1060209136

In this PR we copy all `onlineddl_singleton` tests into `onlineddl_scheduler`. A bit of refactoring was required to adapt to the subtle differences between the two.

The objective is to reduce the number of CI workflows.

Once this PR is merged, we will follow up in a next PR that completely removes `onlineddl_singleton` from the code base.

## Related Issue(s)

#6926

## Checklist

-   [ ] "Backport to:" labels have been added if this change should be back-ported
-   [ ] Tests were added or are not required
-   [ ] Documentation was added or is not required

PR #12061, opened 2023-01-10
consolidating OnlineDDL 'declarative' tests into 'scheduler' tests: part 1

shlomi-noach:
## Description

Similarly to https://github.com/vitessio/vitess/pull/12055, the `onlineddl_scheduler` tests now include testing for `declarative`, effectively making `onlineddl_declarative` redundant.

In this PR we do not yet remove `onlineddl_declarative`, because we can't atomically merge the PR while removing a `required` test from the very same PR. A followup PR will remove both `onlineddl_singleton` and `onlineddl_declarative` tests and workflows.

The objective of this change is to reduce CI workflow files.

## Related Issue(s)

#6926

## Checklist

### DIFF
--- a/go/test/endtoend/onlineddl/scheduler/onlineddl_scheduler_test.go
+++ b/go/test/endtoend/onlineddl/scheduler/onlineddl_scheduler_test.go
@@ -17,18 +17,22 @@ limitations under the License.
 package scheduler
 
 import (
+	"context"
 	"errors"
 	"flag"
 	"fmt"
 	"io"
+	"math/rand"
 	"os"
 	"path"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 
 	"vitess.io/vitess/go/mysql"
 	"vitess.io/vitess/go/textutil"
+	"vitess.io/vitess/go/vt/log"
 	"vitess.io/vitess/go/vt/schema"
 	"vitess.io/vitess/go/vt/sqlparser"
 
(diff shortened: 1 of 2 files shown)


go/test/endtoend/onlineddl/scheduler/onlineddl_scheduler_test.go
```diff
@@ -991,7 +1053,679 @@ func TestSingleton(t *testing.T) {
...
+			executeStrategy: "vtctl",
+			ddlStrategy:     string(schema.DDLStrategyOnline),
+		}
+	}
+
+	// init-cleaup
+	t.Run("init: drop table", func(t *testing.T) {
+		// IF EXISTS is not supported in -declarative
+		uuid := testOnlineDDL(t, dropIfExistsStatement, "online", "vtgate", "", "")
+		onlineddl.CheckMigrationStatus(t, &vtParams, shards, uuid, schema.OnlineDDLStatusComplete)
```
mattlord, line 1326: I'm curious why we have this and the following subtest here since they don't use `declarative`? IMO we should remove them as we're not testing the feature and we DO have subtests that confirm they fail.

Same holds true for all of the subtests which aren't using `declarativeStrategy`. Perhaps there's good reason to have them here though, which is fine. I do see that you had a comment about this for some towards the end (they're not declarative but fit best in this test).

go/test/endtoend/onlineddl/scheduler/onlineddl_scheduler_test.go
```diff
@@ -991,7 +1053,679 @@ func TestSingleton(t *testing.T) {
...
+		uuids = append(uuids, uuid)
+		onlineddl.CheckMigrationStatus(t, &vtParams, shards, uuid, schema.OnlineDDLStatusComplete)
+		onlineddl.CheckMigrationArtifacts(t, &vtParams, shards, uuid, true)
+		checkTable(t, tableName, true)
+		initTable(t)
+		testSelectTableMetrics(t)
+	})
+	// CREATE2
+	t.Run("declarative CREATE TABLE with changes where table exists", func(t *testing.T) {
+		// The exists but with different schema
```
mattlord, line 1590: Smallest of nits, but here and in the following sub-test comment we're missing `table` exists in the comment.

mattlord (approved):
I had some minor questions/comments, but nothing blocking. Thank you again! ❤️

go/test/endtoend/onlineddl/scheduler/onlineddl_scheduler_test.go
shlomi-noach: This particular execution ensures the table does not exist before we begin the actual `declarative` testing. In `onlineddl_declarative` this was not required, since we started with an empty cluster. But now that we have converged multiple tests into a single workflow, we need to perform pre-test cleanup. Specifically, the particular table actually does exist from a previous test.
I'm not using `-declarative` in this `DROP TABLE` because I want a "trusted" method before embarking on `-declarative` validations.

go/test/endtoend/onlineddl/scheduler/onlineddl_scheduler_test.go
shlomi-noach: fixed!

shlomi-noach:
ping for review :pray:

shlomi-noach:
Looking for a 2nd review

rohit-nayak-ps (approved):
lgtm

shlomi-noach:
Followup in https://github.com/vitessio/vitess/pull/12182

PR #12064, opened 2023-01-10
[15.0] CI: remove pitrtls test

deepthi:
## Description
This is a manual partial back port of #11611.
The test has been flaky, and it doesn't give us anything that isn't covered by other tests.

## Related Issue(s)
#11611 
#11881 <- picking some of the flaky test fixes from this PR as well

## Checklist


mattlord (approved):

GuptaManan100 (approved):

deepthi:
I picked a flaky test fix from `main` to see if that helps with the failing Upgrade/Downgrade test.

frouioui (approved):

go/test/endtoend/reparent/utils/utils.go
```diff
@@ -129,7 +129,7 @@ func setupCluster(ctx context.Context, t *testing.T, shardName string, cells []s
 		// the replication manager to silently fix the replication in case ERS or PRS mess up. All the
 		// tests in this test suite should work irrespective of this flag. Each run of ERS, PRS should be
 		// setting up the replication correctly.
-		"--disable_active_reparents")
```
frouioui, line 132: We should add this flag conditionally, if the binary version of `vttablet` is >= 16 then let's use `--disable-replication-manager`, otherwise, let's use `--disable_active_reparents`. This would unblock the `Upgrade Downgrade Testing Reparent Old VTTablet` CI workflow

deepthi: Ah, this flag was introduced in 15, and doesn't exist in 14. And on this branch we test upgrade/downgrade between 14 and 15.
I'm going to first drop this commit and see if the test passes. Otherwise I'll implement the conditional code and 🙏

PR #11491, opened 2022-10-13
Adding deprecate message to backup hooks

rsajwani:
Signed-off-by: Rameez Sajwani <rameezwazirali@hotmail.com>

## Description

This PR generates deprecation message for backup hooks. They are plan to be removed in v16.0

## Checklist

-   [ ] "Backport me!" label has been added if this change should be backported
-   [x] Tests were added or are not required
-   [x] Documentation was added or is not required

PR #12066, opened 2023-01-10
backup: remove deprecated hook support

deepthi:
## Description
In v15, we deprecated backup/restore hooks and provided a much better way of customizing compression/decompression of backups. This PR deletes the flags and code associated with backup/restore hooks.

## Related Issue(s)
#11491 #10558 #10670 #7802

## Checklist

-   [x] Documentation was added or is not required vitessio/website#1339


deepthi:
I have not created a separate issue for this task. If any reviewers feel that one is required, I'm happy to create it.

go/vt/mysqlctl/builtinbackupengine.go
```diff
@@ -181,8 +176,6 @@ func (fe *FileEntry) open(cnf *Mycnf, readOnly bool) (*os.File, error) {
 // ExecuteBackup runs a backup based on given params. This could be a full or incremental backup.
 // The function returns a boolean that indicates if the backup is usable, and an overall error.
 func (be *BuiltinBackupEngine) ExecuteBackup(ctx context.Context, params BackupParams, bh backupstorage.BackupHandle) (bool, error) {
-	params.Logger.Infof("Hook: %v, Compress: %v", backupStorageHook, backupStorageCompress)
```
mattlord, line 184: Might be informative to instead log the `params` but IMO fine to just remove too.

go/vt/mysqlctl/builtinbackupengine.go
```diff
@@ -301,11 +294,6 @@ func (be *BuiltinBackupEngine) executeIncrementalBackup(ctx context.Context, par
 // and an overall error.
 func (be *BuiltinBackupEngine) executeFullBackup(ctx context.Context, params BackupParams, bh backupstorage.BackupHandle) (bool, error) {
 
-	params.Logger.Infof("Hook: %v, Compress: %v", backupStorageHook, backupStorageCompress)
```
mattlord, line 304: Same comment as above for the log line.

mattlord (approved):

GuptaManan100 (approved):
Did you check if the flags are used in the vitess operator?

GuptaManan100:
Okay, I have verified, its not used in VTop.

deepthi:
> Did you check if the flags are used in the vitess operator?

I didn't check, but I knew that this flag wasn't being used 😄

go/vt/mysqlctl/builtinbackupengine.go
deepthi: I've added a log line that prints more info.
```
I0111 13:48:44.086977   11714 builtinbackupengine.go:179] Executing Backup at 2023-01-11 13:48:39.006292 -0800 PST m=+2.432865209 for keyspace/shard ks/0 on tablet vtbackup-2077360256, concurrency: 4, compress: true, incrementalFromPos:
```

go/vt/mysqlctl/builtinbackupengine.go
deepthi: This is a duplicate log line, because the exact same thing was already being logged just before we call this function. So I have left this deleted.

PR #12071, opened 2023-01-11
Copy design docs over from website

deepthi:
## Description
As part of the website docs review project, we've decided to re-organize design docs. Website should link to either RFC issues or documents from the main vitess repository.
As the first step, I've copied over a bunch of those docs over from the website, renamed and lightly edited them.
Once this PR is merged, I'll do a website PR that removes these files from there and links to these files instead.

## Related Issue(s)
vitessio/website#1254
vitessio/website#1257

## Checklist


GuptaManan100 (approved):
LGTM

frouioui (approved):
I don't have strong opinions but I think it would be better to mode design docs / RFC summaries into a subdirectory of `./doc`. My thinking is that `./doc` will otherwise grow very large and will be less reader-friendly.

Otherwise, LGTM

deepthi:
> I don't have strong opinions but I think it would be better to mode design docs / RFC summaries into a subdirectory of `./doc`. My thinking is that `./doc` will otherwise grow very large and will be less reader-friendly.
> 
> Otherwise, LGTM

Good idea. I will create a design-docs folder and move things in before merging this.

PR #12073, opened 2023-01-11
go/vt/mysqlctl: add configurable read buffer to builtin backups

### LINKED ISSUE #12069: builtin backup engine does not do any read buffering when reading from disk
The builtin backup engine reads files from disk (optionally) compresses them, and sends them to backupstorage.

Currently, the backup engine uses the `*os.File` returned by `os.Open` to read files from disk.

https://github.com/vitessio/vitess/blob/7fc1b48d9ea8f10afa597a1e3885b4fb55f3b6be/go/vt/mysqlctl/builtinbackupengine.go#L166

It then passes this to `io.Copy`.

https://github.com/vitessio/vitess/blob/7fc1b48d9ea8f10afa597a1e3885b4fb55f3b6be/go/vt/mysqlctl/builtinbackupengine.go#L718

`io.Copy` reads in [blocks of `32*1024` bytes](https://github.com/golang/go/blob/6732fcc06df713fc737cee5c5860bad87599bc6d/src/io/io.go#L389)

This is not optimal in all environments. In some environments, it's better to read bigger blocks less frequently.

It would be nice if we performed read buffering here, and even nicer if the buffer size was configurable, so it could be tuned in an optimal way for different hardware.

maxenglander:
## Description

The builtin backup engine reads files from disk, optionally compresses them, and sends them to backupstorage.

Currently, the backup engine uses the `*os.File` returned by `os.Open` to read files from disk.

https://github.com/vitessio/vitess/blob/7fc1b48d9ea8f10afa597a1e3885b4fb55f3b6be/go/vt/mysqlctl/builtinbackupengine.go#L166

It then passes this to `io.Copy`.

https://github.com/vitessio/vitess/blob/7fc1b48d9ea8f10afa597a1e3885b4fb55f3b6be/go/vt/mysqlctl/builtinbackupengine.go#L718

`io.Copy` reads in [blocks of `32*1024` bytes](https://github.com/golang/go/blob/6732fcc06df713fc737cee5c5860bad87599bc6d/src/io/io.go#L389)

This is not optimal in all environments. In some environments, it's better to read bigger blocks less frequently.

## Changes

This PR:

 * Adds support for read-buffering.
 * Makes both read and write buffering configurable via flags.
 * Decouples the write buffer sizes for writing-files-to-disk-during-restore vs. writing-to-backupstorage-during-backups.

## Evidence

Coming up with a benchmark to prove that this is useful is difficult, but anecdotally I've seen a modest performance improvement from increasing the size of the read buffer. I'm not sure if the performance improvement is coming from using the hardware more effectively, or if there are more subtle interactions happening between all the moving pieces of `mysqlctl.Backup`.

My test environment looks like this:

 * Shard with a single ~37 GiB table (mostly UUIDs).
 * Run vtbackup on...
 * An AWS r5a.xlarge (4 vcpu, 32 GiB mem).
 * With an EBS GP3 provisioned with 6K IOPS and 250 MiB/s throughput.
 * In a container, with 24 GiB mem limit, no CPU limit.
 * Backupstorage is a customized version of s3backupstorage that does some extra stuff.
 * Using `compression-engine=external`, `external-compressor="zstd -c -T4 -1"`.

Running vtbackup, the baseline backup time is **~215s**. 
When I add a read buffer of 2MiB, that goes down to **~175s** (~18% improvement).

If we feel it's needed, I can try to put together more complete & reproducible evidence.

## Related Issue(s)

Fixes #12069

## Checklist

-   [x] Documentation: https://github.com/vitessio/website/pull/1365


rsajwani:
Changes look good to me .. Few observations

- we need to update release notes with this change
- we need to update website with this change
- I am ok with this change. We are adding metrics for backup and restore time. Is it possible if we deploy that change first collect data for a week or so and then add this change. This will help us gauge if the values are helping us or not.
-

go/vt/mysqlctl/builtinbackupengine.go
```diff
@@ -128,6 +137,8 @@ func init() {
 func registerBuiltinBackupEngineFlags(fs *pflag.FlagSet) {
 	fs.DurationVar(&BuiltinBackupMysqldTimeout, "builtinbackup_mysqld_timeout", BuiltinBackupMysqldTimeout, "how long to wait for mysqld to shutdown at the start of the backup.")
 	fs.DurationVar(&builtinBackupProgress, "builtinbackup_progress", builtinBackupProgress, "how often to send progress updates when backing up large files.")
+	fs.IntVar(&builtinBackupFileReadBufferSize, "builtinbackup-file-read-buffer-size", builtinBackupFileReadBufferSize, "read files from disk in blocks of this many bytes. Golang defaults are used when set to 0.")
```
rsajwani, line 140: minor: should we make it UintVar instead?

maxenglander: Done!

rsajwani: I think you need to generate help text again ... Rest looks good

maxenglander: Doh, fixed

maxenglander:
> Is it possible if we deploy that change first collect data for a week or so and then add this change. This will help us gauge if the values are helping us or not.

That's definitely possible. This PR does not change current behavior, so in my opinion it would be fine to merge this first, but wait until after metrics PR is in prod before turning on these flags. I leave it to you to decide :)

rsajwani:
Looks good to me .. there is failure in  vreplication_across_db_versions, which is because of https://github.com/vitessio/vitess/pull/12253.

rsajwani (approved):

go/vt/mysqlctl/xtrabackupengine.go
```diff
@@ -66,6 +66,7 @@ var (
 
 const (
 	streamModeTar        = "tar"
+	writerBufferSize     = 2 * 1024 * 1024 /*2 MiB*/
```
deepthi, line 69: why was this moved from builtin to this file? It's not actually used by xtrabackupengine AFAICT.

deepthi: ah it was already being used. And it is no longer used in builtin. 👍

mattlord (approved):

PR #12081, opened 2023-01-12
Makefile: use buildx when GOOS/GOARCH are specified

maxenglander:
Right now it's now very easy to build Docker images on Mac M1, unless I'm missing something obvious. 

If I try to build `lite` images for example, it will kick off a `docker build`, which will try to build arm64/aarch64 Docker images. This usually fails when the Docker build tries to download dependencies for that architecture, and the providers of those dependencies don't support it.

Instead of trying figure out how to resolve arm64 dependencies, this PR tries to work around the problem by making it easier to create amd64 Docker images from Mac M1.

This PR changes the `make docker_*` target so that it uses `docker buildx` to create cross-platform builds when either `GOOS` or `GOARCH` are set.

So for example, on my M1 Mac, I can build images for AMD64 Linux like this:

```sh
$ GOOS=linux GOARCH=amd64 make docker_lite_mysql80
```


deepthi (approved):
LGTM, nice little fix.

dbussink (approved):

PR #12085, opened 2023-01-12
Remove Dead Legacy Workflow Management Code

mattlord:
## Description

The [legacy sharding code was removed in v15](https://github.com/vitessio/vitess/pull/10278), but this old workflow management code is _even older_ than the legacy sharding code. It did not work and was  dead code, so is being removed here.

## Related Issue(s)

  - https://github.com/vitessio/website/issues/1284

## Checklist

-   [x] Tests not required
-   [x] Documentation:


deepthi (approved):

ajm188 (approved):
lgtm!

GuptaManan100:
Folks, we missed the part that when we delete flags, we've to check if VTop is also using them. If it is, then we delete the usage there too.
I traced this from the failure in https://github.com/planetscale/vitess-operator/pull/359

GuptaManan100:
Here is the fix for VTop - https://github.com/planetscale/vitess-operator/pull/360

GuptaManan100:
This PR fixes release-15.0 too, so that upgrade-downgrade tests start working again - https://github.com/vitessio/vitess/pull/12128

PR #12089, opened 2023-01-13
Fix: VTOrc forgetting old instances

### LINKED ISSUE #12088: Bug Report: VTOrc doesn't forget old instances
### Overview of the Issue

After the refactor done in https://github.com/vitessio/vitess/pull/11881, there are cases where VTOrc doesn't forget old instances.

### Reproduction Steps

1. Start a cluster.
2. Stop a tablet without removing its topo information.
3. Start the tablet again with a different host and port information
4. Check VTOrc logs to see that it is trying to fix things with the old tablet information

### Binary Version

```sh
main
```

```sh
main
```

### Log Fragments

_No response_

GuptaManan100:
## Description

This PR fixes the bug described in #12088.
The problem was introduced in #11881  which changed the forgetting logic. Earlier we would compare the hostname and port of the tablets already stored with the ones read from the tablet records and remove the ones not available. A peculiar situation arises when we have a tablet whose hostname and port information is cleared out due to restart/reschedule on a different pod. We don't want to remove the tablet record entirely as this can change the quorum that VTOrc sees. This was fixed in #11881 by changing the forgetting logic to compare the tablet aliases instead of the hostname and port.

This however created a different issue wherein we don't delete the old records corresponding to the same tablet! When a tablet gets rescheduled, we don't delete the old record since the alias still matches the ones we read from the tablet records. The new record is also inserted and we end up with 2 records for the same tablet in the `vitess_tablet` information, one old and one new. 

This problem is fixed by this PR by adding `alias` as an explicit column and making it a unique key. So now when we insert into the table the new record, the old one automatically disappears.

I would actually like to go a step further wherein  we don't just make `alias` a unique key, but also the primary key. We should also rework the InstanceKey field to be the alias instead of identifying the tablets by their hostname and port. However, this isn't in the scope of the bug-fix and I'll do these changes in a separate PR.

## Related Issue(s)

- Fixes #12088 

## Checklist


go/vt/vtorc/logic/tablet_discovery_test.go
```diff
@@ -175,6 +176,26 @@ func TestRefreshTabletsInKeyspaceShard(t *testing.T) {
 		// We expect 1 tablet to be refreshed since that is the only one that has changed
 		verifyRefreshTabletsInKeyspaceShard(t, false, 1, tablets)
 	})
+
+	t.Run("change the port and call refreshTabletsInKeyspaceShard again", func(t *testing.T) {
+		defer func() {
```
GuptaManan100, line 181: Prior to these changes, the newly added test would fail saying that the `vitess_tablet` table had 4 tablets instead of the expected 3.

deepthi (approved):

rsajwani (approved):

Issue #12093, opened 2023-01-13
Bug Report: VReplication Does Not Surface COM_BINLOG_DUMP_GTID Errors


mattlord:
### Overview of the Issue

VReplication ends up making `COM_BINLOG_DUMP_GTID` mysql protocol requests to a vttablet's mysqld in order to stream binary log events from a source. Currently if it gets an error it logs messages about that connection between the tablet's binlog component and the mysqld being lost/closed and returns a generic EOF error to the caller/client, for example:
```
vttablet: rpc error: code = Unknown desc = stream (at source tablet) error @ c7594820-936c-11ed-b025-12af3911a716:1-71: unexpected server EOF
```
In most cases the issue *is an ephemeral error* and reconnecting resolves it. In other cases, the problem persists and all the client sees is that no progress is being made and things seem stuck with what appears to be connection related issues. 

The most common error that gets us to this stuck state is this (from the mysqld error log):
```
[Warning] [MY-011809] [Server] Cannot replicate to server with server_uuid='...' because the present server has purged required binary logs. The connecting server needs to replicate the missing transactions from elsewhere, or be replaced by a new server created from a more recent backup. To prevent this error in the future, consider increasing the binary log expiration period on the present server. The missing transactions are '...'.
```

Currently the only way to detect the situation is by looking at the relevant mysqld's error log and seeing the above message repeatedly there.

We should try and surface this error/failure in VReplication if at all possible so that it's clear that there's an unrecoverable error that the user needs to address.

### How To Repeat
```
git checkout main && make build && pushd examples/local

./101_initial_cluster.sh; ./201_customer_tablets.sh; ./202_move_tables.sh

vtctlclient Workflow -- customer.commerce2customer stop

mysql < ../common/insert_commerce_data.sql

for id in 100 101 102; do
  mysql -u root -h localhost --socket=${VTDATAROOT}/vt_0000000${id}/mysql.sock --binary-as-hex=false -e "flush binary logs; select sleep(10); purge binary logs before now()"
done

vtctlclient Workflow -- customer.commerce2customer start
sleep 10
vtctlclient Workflow -- customer.commerce2customer show

./401_teardown.sh
popd
```

The final relevant output from the `show` is:
```
"State": "Error",
"Message": "vttablet: rpc error: code = Unknown desc = stream (at source tablet) error @ c7594820-936c-11ed-b025-12af3911a716:1-71: unexpected server EOF",
```

PR #12095, opened 2023-01-13
VReplication: Propagate Binlog Stream Errors

mattlord:
## Description

This PR adds an errors channel to the binlog streamer so that we can propagate stream errors back to the caller/client. See [the bug report](https://github.com/vitessio/vitess/issues/12093) for additional details.

The key final snippet of the `show` output from the bug report's test case is now:
```
"State": "Error",
"Message": "vttablet: rpc error: code = Unknown desc = stream (at source tablet) error @ a61630ee-9379-11ed-b080-97c823585086:1-71: Cannot replicate because the master purged required binary logs. Replicate the missing transactions from elsewhere, or provision a new slave from backup. Consider increasing the master's binary log expiration period. The GTID set sent by the slave is 'a61630ee-9379-11ed-b080-97c823585086:1-71', and the missing transactions are 'a61630ee-9379-11ed-b080-97c823585086:72-83' (errno 1236) (sqlstate HY000)",
```

## Related Issue(s)

  - Fixes: https://github.com/vitessio/vitess/issues/12093

## Checklist

-   [x] Tests were added

### DIFF
--- a/go/vt/binlog/binlog_connection.go
+++ b/go/vt/binlog/binlog_connection.go
@@ -29,8 +29,6 @@ import (
 	"vitess.io/vitess/go/pools"
 	"vitess.io/vitess/go/vt/dbconfigs"
 	"vitess.io/vitess/go/vt/log"
-	vtrpcpb "vitess.io/vitess/go/vt/proto/vtrpc"
-	"vitess.io/vitess/go/vt/vterrors"
 )
 
 var (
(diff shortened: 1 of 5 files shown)


rohit-nayak-ps (approved):
Looks good! 

I verified that the VStream API now returns the new error on local example after purging logs.

shlomi-noach (approved):
LGTM, to verify, the errors channel is only used to send a single error; once that happens the vreplication stream is closed, correct?

mattlord:
> LGTM, to verify, the errors channel is only used to send a single error; once that happens the vreplication stream is closed, correct?

Yes, today there's only ever a single error sent in the channel. The other option would be to add an error to the struct with a getter and setter, but this seemed slightly more idiomatic and leaves the door open to other behavior in the future.

PR #12096, opened 2023-01-15
schemadiff: find if a schema diff has ordering constraints

shlomi-noach:
This checks if the order of diffs is required; a new function `DiffsHaveDependencies()` returns 'true' when there is _some_ ordering conflict between any two diffs

The question is of general interest, when we have a list of diffs of two schemas, and wish to apply them onto the original schema. The ordering constraint tells us whether we can just apply them in any random order, or concurrently, or whether we have to restrict ourselves to some strict order.


shlomi-noach:
Currently the only dependency diff that is tested is `ViewDependencyUnresolvedError`; in the future we will add foreign key dependencies.

shlomi-noach:
Converted to draft as I'm working on a more sophisticated solution.

shlomi-noach:
Superseded by #12551

PR #12098, opened 2023-01-15
OnlineDDL: support integer-to-enum conversion in `vitess` migrations

### LINKED ISSUE #12097: Bug Report: in OnlineDDL/vitess, converting an integer type to `enum` fails the migration
### Overview of the Issue

When a migration converts an integer column (e.g. `int`) to an `enum`, vreplication's table plan as well as vcopier attempt to insert the integer value to the target table. But that is wrong, since in the target table an integer value is perceived to be the _index_ of the enum value rather than the value.

Consider an `enum('7', '8')`. Consider that the original table has the value `7`. Attempting to insert the value `7` is an error because there is no value in index `7`. There are only indexes `1` and `2`, mapping to the string values `"7"` and `"8"`, respectively.

We should support the specific case of converting integers to enums, such that the value is converted to a string.

### Reproduction Steps

Use this table:
```sql
create table onlineddl_test (
  id int auto_increment,
  i int not null,
  primary key(id)
) auto_increment=1;
```
Populate it with values:
```sql
insert into onlineddl_test values (null, 0);
insert into onlineddl_test values (null, 1);
```
try to convert it like so using `online` strategy: 
```sql
alter table onlineddl_test change i i enum('0', '1')
```

### Binary Version

```sh
-
```

```sh
-
```

### Log Fragments

_No response_

shlomi-noach:
## Description

This PR fixes #12097 and correctly converts an integer into an `enum` value in a `vitess` strategy migration.
The solution is similar to how we map enum values, or how we convert character sets: we have new information in proto's `message Rule`. `onlineddl.VRepl` populates that information. It modifies the streamer's `SELECT` query to convert the integer to text. The vreplication table plan also gets hold of this information on converts integer types to `enum` when applying binlog events.

## Related Issue(s)

Fixes https://github.com/vitessio/vitess/issues/12097
#6926

## Checklist


go/vt/vttablet/onlineddl/vrepl/types.go
```diff
@@ -92,6 +94,11 @@ func (c *Column) IsNumeric() bool {
 	return c.NumericPrecision > 0
 }
 
+// IsIntegral returns true if the column is some form of an integer
+func (c *Column) IsIntegral() bool {
```
mattlord, line 98: Nit, but IMO `IsIntegralType()` would be good here as I think it's more explicit/descriptive and then the name used is uniform between the two packages.

proto/binlogdata.proto
```diff
@@ -171,6 +171,11 @@ message Rule {
   // SourceUniqueKeyTargetColumns represents the names of columns in target table, mapped from the chosen unique
   // key on source tables (some columns may be renamed from source to target)
   string source_unique_key_target_columns = 7;
+
+  // ConvertIntToEnum lists any columns that are converted from an integral value into an enum.
+  // such columns need to have special transofrmation of the data, from an integral format into a
```
mattlord, line 176: Such a small nit that I'm annoying myself as I type 🤣 but we use sentence capitalization and punctuation in the first part and NOT in the second. IMO we should use it for both.

mattlord (approved):
LGTM! I had two of the nittiest nits possible — up to you if/how to address them.

go/vt/vttablet/onlineddl/vrepl/types.go
shlomi-noach: Renamed

proto/binlogdata.proto
shlomi-noach: Fixed

rohit-nayak-ps (approved):
lgtm

PR #12099, opened 2023-01-16
store transaction isolation level in upper case

harshit-gangal:
## Description

This is a follow-up PR from https://github.com/vitessio/vitess/pull/11987 for comment https://github.com/vitessio/vitess/pull/11987#discussion_r1064957797. The only possible solution was to store it as upper case as how MySQL returns the value for the `transaction_isolation` system variable.

## Checklist


go/test/endtoend/vtgate/reservedconn/sysvar_test.go
```diff
@@ -433,30 +433,30 @@ func TestSysVarTxIsolation(t *testing.T) {
...
 	// ensuring it goes to mysql
 	utils.AssertContains(t, conn, "select @@transaction_isolation, connection_id()", `REPEATABLE-READ`)
-	// second run, to ensuring the setting is applied on the session and not just on next query after settings.
+	// second run, ensuring it has the same value.
 	utils.AssertContains(t, conn, "select @@transaction_isolation, connection_id()", `REPEATABLE-READ`)
 
 	// setting to different value.
-	utils.Exec(t, conn, "set @@transaction_isolation = 'read-committed'")
-	utils.AssertMatches(t, conn, "select @@transaction_isolation", `[[VARCHAR("read-committed")]]`)
+	utils.Exec(t, conn, "set @@transaction_isolation = 'READ-COMMITTED'")
```
systay, line 440: we want to store them with upper case and return them with uppercase, but we want to allow for both cases for input, right?

go/vt/sqlparser/parse_test.go
```diff
@@ -1107,25 +1107,25 @@ var (
 		output: "set /* mixed list */ @@a = 3, names 'utf8', charset 'ascii', @@b = 4",
 	}, {
 		input:  "set session transaction isolation level repeatable read",
-		output: "set @@session.transaction_isolation = 'repeatable-read'",
+		output: "set @@session.transaction_isolation = 'REPEATABLE-READ'",
```
systay, line 1110: I think we want to add more tests so we have both cases, not change the existing ones

harshit-gangal: Here the output changes not the input.

go/test/endtoend/vtgate/reservedconn/sysvar_test.go
harshit-gangal: currently, we do not change the case when `set <system_variable>` is used and store it as-is. 
This is only for the case when `set transaction isolation level <value>` is provided we have the option as we convert the statement to `set @@session.transaction_isolation = <value>`

Do you suggest that I make that change as well, to store the user-provided value in upper case for `set <system_varable>`?

GuptaManan100: Do we use the stored value for printing things, not the constants defined in sql.y?

GuptaManan100 (approved):
Apart from the ongoing discussion, everything else LGTM!

go/test/endtoend/vtgate/reservedconn/sysvar_test.go
harshit-gangal: Changed as discussed offline.

go/vt/sqlparser/parse_test.go
harshit-gangal: This is not longer valid.

go/vt/sysvars/sysvars.go
```diff
@@ -39,8 +39,18 @@ type SystemVariable struct {
...
 
 	SupportSetVar bool
+
+	Case StorageCase
 }
 
+type StorageCase int
+
+const (
+	SCSame StorageCase = iota
```
systay, line 49: nitpick: Couldn't we skip the `SC` and just use `sysvars.AlwaysUpper`,`sysvars.AlwaysLower` and `sysvars.FromUser`? really not important, if you want to merge, go ahead :)

systay (approved):

PR #12105, opened 2023-01-16
Keep the correct case for the charset for canonical string

### LINKED ISSUE #12104: Bug Report: wrong case for `sqlparser.CanonicalString` and `CONVERT(... USING...)`
### Overview of the Issue

The following query is not normalized correctly in `sqlparser.CanonicalString`:

```
select convert('abc' using utf8mb4)
```

It currently upcases the charset which is incorrect.

dbussink:
The canonical string representation needs to keep the correct case for the charset. Those are never upcased in MySQL so we should not enforce that here either.

## Related Issue(s)

Fixes #12104

## Checklist

### DIFF
--- a/go/vt/sqlparser/ast_format.go
+++ b/go/vt/sqlparser/ast_format.go
@@ -1773,7 +1773,7 @@ func (node *ConvertExpr) Format(buf *TrackedBuffer) {
 
 // Format formats the node.
 func (node *ConvertUsingExpr) Format(buf *TrackedBuffer) {
-	buf.astPrintf(node, "convert(%v using %s)", node.Expr, node.Type)
+	buf.astPrintf(node, "convert(%v using %#s)", node.Expr, node.Type)
 }
 
 // Format formats the node.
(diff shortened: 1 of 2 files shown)


shlomi-noach (approved):
Should we enforce lower casing the charset though?

dbussink:
> Should we enforce lower casing the charset though?

We don't do that anywhere in the parser at the moment at all. We could maybe, but that's a whole different effort that goes beyond just the charset here, but applies to anything that's always lowercase and would need separate logic to be enforced then.

GuptaManan100:
I don't think we should do the lower casing, because mixed case and upper case do work with MySQL - 
```mysql
mysql [localhost:5731] {msandbox} (test) > select convert('abc' using utf8mb4);
+------------------------------+
| convert('abc' using utf8mb4) |
+------------------------------+
| abc                          |
+------------------------------+
1 row in set (0.00 sec)

mysql [localhost:5731] {msandbox} (test) > select convert('abc' using UTF8mb4);
+------------------------------+
| convert('abc' using UTF8mb4) |
+------------------------------+
| abc                          |
+------------------------------+
1 row in set (0.00 sec)

mysql [localhost:5731] {msandbox} (test) > select convert('abc' using UTF8MB4);
+------------------------------+
| convert('abc' using UTF8MB4) |
+------------------------------+
| abc                          |
+------------------------------+
1 row in set (0.00 sec)
```

PR #12107, opened 2023-01-16
CI, tabletmanager throttler topo tests: polling until status received

shlomi-noach:
## Description

An attempt to fix `tabletmanager_throttler_topo` flakiness, as seen [here](https://github.com/planetscale/vitess-private/actions/runs/3913922615/jobs/6690389759#logs). In a few places where we expect to find a status "OK", we now poll for the status until "OK" is received, or until some timeout (at which time the test fails).

The flakiness is only found in GitHub CI and never in my local tests, which suggests this is a timing issue. This fix hopefully works around it.

## Checklist

### DIFF
--- a/go/test/endtoend/tabletmanager/throttler_topo/throttler_test.go
+++ b/go/test/endtoend/tabletmanager/throttler_topo/throttler_test.go
@@ -271,10 +271,7 @@ func TestInitialThrottler(t *testing.T) {
 	defer cluster.PanicHandler(t)
 
 	t.Run("validating OK response from disabled throttler", func(t *testing.T) {
-		resp, err := throttleCheck(primaryTablet, false)
-		require.NoError(t, err)
-		defer resp.Body.Close()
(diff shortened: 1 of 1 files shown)


go/test/endtoend/tabletmanager/throttler_topo/throttler_test.go
```diff
@@ -516,15 +510,8 @@ func TestRestoreDefaultQuery(t *testing.T) {
 		_, err := updateThrottlerConfig(true, false, throttlerThreshold.Seconds(), "", false)
 		assert.NoError(t, err)
 	})
-	t.Run("requesting heartbeats", func(t *testing.T) {
-		_ = warmUpHeartbeat(t)
-	})
```
shlomi-noach, line 521: There is no need to request heartbeats, since `waitForThrottleCheckStatus` does that anyway.

deepthi (approved):

PR #12111, opened 2023-01-17
BugFix: Cast expression translation by evaluation engine

### LINKED ISSUE #12110: Bug Report: Cast expression query fails but works with a limit clause
### Overview of the Issue

The following query doesn't work in Vitess - 
```
-- Doesn't work and returns "Unsupported type conversion: DATE" message:
select cast('2023-01-07 12:34:56' as date);
```

However, if we add a limit clause, then it does work - 
```
-- Works:
select cast('2023-01-07 12:34:56' as date) limit 1;
```

### Reproduction Steps

Start a vitess cluster and run the listed queries

### Binary Version

```sh
main, release-15 and release-14
```

```sh
main, release-15 and release-14
```

### Log Fragments

_No response_

GuptaManan100:
## Description

This PR fixes the bug described in #12110. The issue is that the evalengine converts the expression into a eval engine expression whose evaluation isn't supported at runtime. What this means is that planning phase succeeds but the runtime throws an error.
This isn't correct since we have an alternate way of planning of not using the evaluation engine and send the query to a vttablet like we did before evalengine existed. 

This PR fixes this bug by failing the translation itself instead of creating an expression that can't be evaluated.

## Related Issue(s)

- Fixes #12110 

## Checklist

### DIFF
--- a/go/vt/vtgate/evalengine/convert.go
+++ b/go/vt/vtgate/evalengine/convert.go
@@ -39,6 +39,10 @@ type (
 )
 
 func (c *ConvertExpr) unsupported() {
+	throwEvalError(c.returnUnsupportedError())
+}
+
+func (c *ConvertExpr) returnUnsupportedError() error {
 	var err error
 	switch {
 	case c.HasLength && c.HasScale:
@@ -48,7 +52,7 @@ func (c *ConvertExpr) unsupported() {
 	default:
 		err = vterrors.Errorf(vtrpcpb.Code_UNIMPLEMENTED, "Unsupported type conversion: %s", c.Type)
 	}
-	throwEvalError(err)
+	return err
 }
 
 func (c *ConvertExpr) eval(env *ExpressionEnv, result *EvalResult) {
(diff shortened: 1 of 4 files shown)


go/vt/vtgate/evalengine/convert.go
```diff
@@ -39,6 +39,10 @@ type (
 )
 
 func (c *ConvertExpr) unsupported() {
+	throwEvalError(c.returnUnsupportedError())
+}
+
+func (c *ConvertExpr) returnUnsupportedError() error {
 	var err error
```
harshit-gangal, line 46: nit: err variable can be removed.

harshit-gangal (approved):

GuptaManan100:
This is also a regression. It works for release-13.0 but not for 14, 15 and main.

go/vt/vtgate/evalengine/convert.go
GuptaManan100: We need it. We write to it in the switch case. Otherwise we would need to return in each switch case. I don't hate that either.

frouioui (approved):

Issue #12112, opened 2023-01-17
Feature Request: Online DDL, in-order completion of migrations


shlomi-noach:
### Feature Description

`ApplySchema` may run multiple schema changes. With Online DDL, and when `--allow-concurrent` is given, those migrations may run concurrently. These migrations may then complete at any time of their choosing. A `CREATE TABLE` migration, for example, can be expected to complete very quickly, as opposed to a long running `ALTER TABLE`.

Even when `--postpone-completion` is given, and even if we run `ALTER VITESS_MIGRATION COMPLETE ALL` when all migrations are `ready_to_complete`, even then the migrations may complete in any arbitrary order.

We wish to introduce a new DDL strategy flag, `--in-order-completion`, which enforces the ordering of migration completion. A migration running with this strategy will only complete when it is the first non-pending migration in the same migration context group.

That is, if there's any pending (`queued`, `ready`, `running`) migration whose `id` is smaller (thus, precedes our migration), our migration will stall and not complete.

Note that it _is_ possible for the migration to run if an earlier migration has `failed`. It's important to allow that to avoid hanging migrations.

The outcome of this feature is that the order of migrations as submitted in `ApplySchema` is the order in which the migrations are finalized. The feature puts a specific order to what otherwise would be an arbitrary order.

Tracking issue: https://github.com/vitessio/vitess/issues/6926

### Use Case(s)

The use case for this feature are migrations dependencies. Ideally, an `ApplySchema` should not have inter-dependencies between migrations, but it can happen.

Examples:

- On a pre-existing table, we add a new column. We also create a `VIEW` that selects from that column. In this situation, the `ALTER TABLE` must fully complete before we attempt to create the view, even though view creation is near instantaneous.
- Creating two tables with parent-child foreign key relationship. The parent table must be created first, the child table second.

With `--in-order-completion`, we assume the user has sorted out the correct order of migrations. `vitess` does not attempt to order the migrations. The task is very complex and sometimes not at all possible. Possibly omething for the future, but not required for now. `schemadiff` is able to [resolve some ordering dependencies](https://github.com/vitessio/vitess/issues/12028).

shlomi-noach:
Originally I thought the feature should be scoped to migrations in the same migration context. But I would also like to explore the possibility of cross-context in-order completion. This will be useful if, for example, you submit a single migration at a time, and still want them to run concurrently and complete in-order. Even more interesting is when you want to `REVERT` multiple migrations, concurrently, and want to complete the reverts in-order.

PR #12113, opened 2023-01-17
Online DDL: --in-order-completion ddl strategy and logic

shlomi-noach:
## Description

Read story in https://github.com/vitessio/vitess/issues/12112

This PR introduces a `--in-order-completion` DDL strategy flag. Any migration that runs under this flag will only _complete_ if there's no prior migrations still pending.

In the happy story, this means a migration will complete when all prior migrations have completed, thus, migrations complete in-order. 

In a less-than-happy story, some migrations may fail or be cancelled. This still lets an `--in-order-completion` migration complete, or else we can have infinitely hanging migrations.

The order of migrations is determined by `_vt.schema_migrations.id` column. In a multi-statement `ApplySchema`, for example, the order of queries in the `--sql` value is the order in which they're written to the database table, hence they're similarly ordered by `id`, which is auto-incrementing.

Note that `--in-order-completion` still allows concurrency. In fact, it is designed to work with concurrent migrations. The idea is that as many migrations as we wish may run concurrently, but the way they finally `complete` is in-order.

The flag applies to:

- all "immediate" migrations: `CREATE|DROP TABLE`, `CREATE|ALTER|DROP VIEW`
- including `ALTER TABLE` that uses `INSTANT` DDL
- as well as any `vitess|online` `ALTER TABLE` migration

Perhaps it's better to say where the flag does _not_ apply: it does not apply to `gh-ost` and `pt-osc` migrations.

This PR needs website docs updates as well as changelog update.

## Related Issue(s)

Closes https://github.com/vitessio/vitess/issues/12112

## Checklist


shlomi-noach:
Added release notes.

shlomi-noach:
Documentation PR: https://github.com/vitessio/website/pull/1352

go/test/endtoend/onlineddl/scheduler/onlineddl_scheduler_test.go
```diff
@@ -804,6 +811,85 @@ func TestSchemaChange(t *testing.T) {
...
+		t.Run("expect table success", func(t *testing.T) {
+			status := onlineddl.WaitForMigrationStatus(t, &vtParams, shards, t1uuid, normalWaitTime, schema.OnlineDDLStatusComplete, schema.OnlineDDLStatusFailed)
+			fmt.Printf("# Migration status (for debug purposes): <%s>\n", status)
+			onlineddl.CheckMigrationStatus(t, &vtParams, shards, t1uuid, schema.OnlineDDLStatusComplete)
+		})
+		t.Run("expect view success", func(t *testing.T) {
+			status := onlineddl.WaitForMigrationStatus(t, &vtParams, shards, v1uuid, normalWaitTime, schema.OnlineDDLStatusComplete, schema.OnlineDDLStatusFailed)
+			fmt.Printf("# Migration status (for debug purposes): <%s>\n", status)
+			onlineddl.CheckMigrationStatus(t, &vtParams, shards, v1uuid, schema.OnlineDDLStatusComplete)
+		})
```
mattlord, line 890: Same comment here. The dependency here can be used to confirm order but I think only if we require that they both actually complete.

go/vt/schema/ddl_strategy_test.go
```diff
@@ -123,6 +124,13 @@ func TestParseDDLStrategy(t *testing.T) {
 			runtimeOptions:       "",
 			isPostponeCompletion: true,
 		},
+		{
+			strategyVariable:    "online --in-order-completion",
+			strategy:            DDLStrategyOnline,
```
mattlord, line 129: Don't we also want this for the vitess strategy?

go/vt/vttablet/onlineddl/executor.go
```diff
@@ -3518,6 +3532,10 @@ func (e *Executor) reviewRunningMigrations(ctx context.Context) (countRunnning i
 	if err != nil {
 		return countRunnning, cancellable, err
 	}
+	pendingMigrationsUUIDs, err := e.readPendingMigrationsUUIDs(ctx)
```
mattlord, line 3535: Same note here about enforcing/guaranteeing submission order by adding an ORDER BY clause to the query that gets used.

go/test/endtoend/onlineddl/scheduler/onlineddl_scheduler_test.go
```diff
@@ -804,6 +811,85 @@ func TestSchemaChange(t *testing.T) {
...
+			fmt.Sprintf("drop view  if exists t3_%s", u),
+		}
+		sql := strings.Join(sqls, ";")
+		var vuuids []string
+		t.Run("drop multiple tables and views, in-order-completion", func(t *testing.T) {
+			uuidList := testOnlineDDLStatement(t, createParams(sql, ddlStrategy+" --allow-concurrent --in-order-completion", "vtctl", "", "", true)) // skip wait
+			vuuids = strings.Split(uuidList, "\n")
+			assert.Equal(t, 4, len(vuuids))
+			for _, uuid := range vuuids {
+				onlineddl.WaitForMigrationStatus(t, &vtParams, shards, uuid, normalWaitTime, schema.OnlineDDLStatusComplete, schema.OnlineDDLStatusFailed)
```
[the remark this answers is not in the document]
shlomi-noach: @mattlord the thing is that if a migratoin goes into `failed` state, that's a terminal state. There is no need to further wait for a `complete` state. This line says "wait until either 'complete' or 'failed' states are met".

go/test/endtoend/onlineddl/scheduler/onlineddl_scheduler_test.go
```diff
@@ -804,6 +811,85 @@ func TestSchemaChange(t *testing.T) {
...
+			}
+		})
+		require.Equal(t, 4, len(vuuids))
+		for i := range vuuids {
+			if i > 0 {
+				testTableCompletionTimes(t, vuuids[i-1], vuuids[i])
```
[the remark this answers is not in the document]
shlomi-noach: The onlineDDL scheduler is incapable of completing two migrations within the same second. The comparison is fair.

UUIDs are in submission order. Therefore `vuuids[0]` refer to the UUID of the first migration, `vuuids[1]` is the UUID of the 2nd migration, etc.

go/test/endtoend/onlineddl/scheduler/onlineddl_scheduler_test.go
```diff
@@ -804,6 +811,85 @@ func TestSchemaChange(t *testing.T) {
...
+		createv1 := fmt.Sprintf("create view %s as select id from %s", v1name, v2name)
+
+		sql := fmt.Sprintf("%s; %s;", createv2, createv1)
+		var vuuids []string
+		t.Run("create two views, expect both complete", func(t *testing.T) {
+			uuidList := testOnlineDDLStatement(t, createParams(sql, ddlStrategy+" --allow-concurrent --in-order-completion", "vtctl", "", "", true)) // skip wait
+			vuuids = strings.Split(uuidList, "\n")
+			assert.Equal(t, 2, len(vuuids))
+			for _, uuid := range vuuids {
+				onlineddl.WaitForMigrationStatus(t, &vtParams, shards, uuid, normalWaitTime, schema.OnlineDDLStatusComplete, schema.OnlineDDLStatusFailed)
```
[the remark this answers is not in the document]
shlomi-noach: This doesn't confirm the order. It just verifies success of both migrations. Order is verified later.

go/test/endtoend/onlineddl/scheduler/onlineddl_scheduler_test.go
shlomi-noach: We do validate that they both `complete`.  `"expect table success"` validates that the table's migration is `complete`, and `"expect view success"` validates the the view's migration is complete.
The above does not validate order. Order is validated later on.

go/vt/schema/ddl_strategy_test.go
shlomi-noach: `vitess` == `online`. They are synonyms with the intention of only using `vitess`. But changing names is hard.

go/vt/vttablet/onlineddl/executor.go
```diff
@@ -3267,22 +3267,36 @@ func (e *Executor) runNextMigration(ctx context.Context) error {
...
+
+			if conflictFound, _ := e.isAnyConflictingMigrationRunning(onlineDDL); conflictFound {
+				continue // this migration conflicts with a running one
+			}
+			if e.countOwnedRunningMigrations() >= maxConcurrentOnlineDDLs {
+				continue // too many running migrations
+			}
+			if isImmediateOperation && onlineDDL.StrategySetting().IsInOrderCompletion() {
+				// This migration is immediate: if we run it now, it will complete within a second or two at most.
+				if len(pendingMigrationsUUIDs) > 0 && pendingMigrationsUUIDs[0] != onlineDDL.UUID {
```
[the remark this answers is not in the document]
shlomi-noach: >  the rows are in submission order here?

Yes!

> If so, we should add an explicit ORDER BY clause to the sqlSelectPendingMigrations statement.

It was already added! https://github.com/vitessio/vitess/pull/12113/files#diff-d7427b60825b46773b837642e539fd8f00382fc230b717043f3aa76d3ee7ae14R386-R394

go/vt/vttablet/onlineddl/executor.go
shlomi-noach: You must have missed it, all relevant queries have `ORDER BY id`

go/vt/vttablet/onlineddl/executor.go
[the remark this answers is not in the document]
mattlord: Sorry! I was looking at the wrong snapshot.

go/test/endtoend/onlineddl/scheduler/onlineddl_scheduler_test.go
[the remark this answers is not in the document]
mattlord: If it fails, it won't be seen as a migration that finished at the beginning of the Unix epoch? This is the kind of thing a unit test could confirm. What do we expect the test failure/output to be in that case?

go/test/endtoend/onlineddl/scheduler/onlineddl_scheduler_test.go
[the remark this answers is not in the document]
mattlord: This is the same point as above so resolving this one.

go/test/endtoend/onlineddl/scheduler/onlineddl_scheduler_test.go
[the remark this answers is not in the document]
mattlord: OK. `The onlineDDL scheduler is incapable of completing two migrations within the same second` isn't something I'd seen/noticed before. I also don't see it here (unless I'm blind): https://github.com/vitessio/vitess/blob/main/doc/design-docs/OnlineDDLScheduler.md

So if that behavior is assumed/required for this feature to work IMO we should at least comment that somewhere. Maybe we already have?

go/test/endtoend/onlineddl/scheduler/onlineddl_scheduler_test.go
mattlord: Yes, you're right. Here we do `onlineddl.CheckMigrationStatus(t, &vtParams, shards, t1uuid, schema.OnlineDDLStatusComplete)` which returns an error, w/o waiting, if the current status does not match. Sorry about that.

Side note:
This logic still doesn't make much sense to me as I don't see how that is better/different than returning an error in the wait if we don't get completed status by the timeout. In my mind the wait should return an error if the status you're waiting for doesn't occur within the provided timeout. Having this split into different calls invites flakiness by the author not knowing or forgetting that wait doesn't actually wait / return an error. Technically you don't even *need* to pass Failed in that call other than to stop waiting sooner. 

I'm sorry for being a PITA but we really need to get the flakiness of these under control. It's been a year or so now with it being a constant issue so I'm being hyper/over critical here (I know that it's unpleasant, for both of us). I think we could eliminate lines of code, simplify the tests, and make them more robust for all of us with: 
```
onlineddl.WaitForMigrationStatus(t, &vtParams, shards, v1uuid, normalWaitTime, schema.OnlineDDLStatusComplete)
```

And that having a require/assert and printing the message if we don't get what we want in the provided timeout:
```
"Expected migration status of Completed not seen for UUID <uuid> within the timeout of 60s. Last seen status: Failed\n"
```

This would make the discussion above moot because we'd clearly fail the test immediately if we didn't reach that Completed status and not have to wonder what happens after (since we're testing that they all succeed and complete in order).  I'll put that on myself though for some refactors later on to clean up some of the logic where we wait for different things then we check and it's not always clear why e.g. So banking this for myself later and will resolve this thread.

go/test/endtoend/onlineddl/scheduler/onlineddl_scheduler_test.go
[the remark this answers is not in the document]
shlomi-noach: > OK. The onlineDDL scheduler is incapable of completing two migrations within the same second isn't something I'd seen/noticed before. I also don't see it here (unless I'm blind):

You know, with apologies let me retract that comment. It is true, but irrelevant and confusing. The test `testTableCompletionTimes` merely checks that timestamp1 <= timestamp2; whether there's a full second between them or not, is irrelevant and not tested. 

With that, I understand the question; given that `completed_timestamp` is in 1second resolution, how do we validate that the two migrations did, in fact, complete in a specific order? Let me look into that and hopefully I can refine the tests!

go/test/endtoend/onlineddl/scheduler/onlineddl_scheduler_test.go
shlomi-noach: > This logic still doesn't make much sense to me as I don't see how that is better/different than returning an error in the wait if we don't get completed status by the timeout.

I believe I have a good answer for that! The problem this is trying to solve is excessive CI runtime. Yes, we _could_ just wait for `complete` state. We can wait for `30sec` (or whatever it is), and if by that time we still don't see `complete`, then the test has failed.

But with GitHub CI, and with so many tests in this suite, this can accumulate and get us to tests timeouts. So, we take advantage of the following fact: if the test is `failed`, then that's a terminal state. There is no way for it to magically transition into `success`! And so if it is `failed`, we can stop waiting. That can shave off dozens of seconds. It actually does! Since there's some tests we _expect_ to fail.

And so the general idea is this:

- we run a migration
- we wait until it has reached a terminal state: either `complete` or `failed`; we know there's no point waiting beyond any of these
- we then validate what the terminal state actually was. Typically we validate that it is `complete`

Does that make more sense?

Some other tests, BTW, wait until the state is `queued` or `ready`. This happens with some of the "postponed" tests. Since these states are non-terminal, there _could be_ a scenario where a migration will then transition into `running` or `complete`. In some of these cases we do an extra sleep after reaching `queued` or `ready`, and then _again_ check to see that the state is still in `queued` or `ready`.

go/test/endtoend/onlineddl/scheduler/onlineddl_scheduler_test.go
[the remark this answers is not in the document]
shlomi-noach: @mattlord `completed_timestamp` is now `timestamp(6)`, and with that I think the comparison is now safe.

go/test/endtoend/onlineddl/scheduler/onlineddl_scheduler_test.go
mattlord: Yeah, that does make sense! I would then prefer to have the wait always check for failed status and if it sees that then fail the test via require/assert. We don't need to figure this out here, just something to consider for later.

mattlord (approved):
Thanks! ❤️

go/test/endtoend/onlineddl/scheduler/onlineddl_scheduler_test.go
shlomi-noach: Except for those cases where we _expect_ the migration to fail...

shlomi-noach:
Looking for a 2nd review/approval

deepthi:
request for review from @vitessio/query-serving

go/vt/vttablet/onlineddl/executor.go
```diff
@@ -3267,22 +3267,36 @@ func (e *Executor) runNextMigration(ctx context.Context) error {
...
+					continue
 				}
 			}
+			// This migration seems good to go
+			return onlineDDL, err
```
frouioui, line 3299: Should we update the pending migrations to say that the migration `onlineDDL.UUID` is no longer in a pending state?

frouioui (approved):

go/vt/vttablet/onlineddl/executor.go
shlomi-noach: That happens a few lines below,

```go
e.executeMigration(ctx, onlineDDL)
```

So, it's only when the migration is actually started, or executed, that it leaves the pending migrations. There can always be a failure in between, and we don't want to lose track of the migration. This is why we rely on the `migration_status` persisted in `_vt.schema_migrations`.

frouioui: Got it, makes sense. Thanks!

Issue #12101, opened 2023-01-16
Bug Report: Vreplication can stop suddenly after an error instead of retrying


arthurschreiber:
### Overview of the Issue

We've been running a `CreateLookupVindex` in a sharded keyspace, and ran into an issue where on one of the shards, multiple (but not all) of the workflows stopped unexpectedly, and were not retried.

I believe this is due to a bug in the `lastWorkflowError` functionality in the `vreplication` controller code.

When a vreplication controller encounters an error, the error is recorded using [`ct.lastWorkflowError.record(err)`](https://github.com/vitessio/vitess/blob/ece6501975604ef5d0be3ab256145047e13d9c00/go/vt/vttablet/tabletmanager/vreplication/controller.go#L270-L271).

If the error is seen as recoverable, and [`ct.lastWorkflowError.shouldRetry()`](https://github.com/vitessio/vitess/blob/ece6501975604ef5d0be3ab256145047e13d9c00/go/vt/vttablet/tabletmanager/vreplication/controller.go#L273) signals that the vreplication can be retried, the controller will attempt to restart the vreplication workflow.

The idea behind `lastWorkflowError` is to record errors from the vreplication call, and allow restarting vreplication again unless vreplication continuously fails with the same error for a duration longer than specified by the `vreplication_max_time_to_retry_on_error` flag.

Unfortunately, I believe this check is implemented incorrectly. [`vr.Replicate(ctx)`](https://github.com/vitessio/vitess/blob/ece6501975604ef5d0be3ab256145047e13d9c00/go/vt/vttablet/tabletmanager/vreplication/controller.go#L269) is a potentially long running call, and errors are only recorded (or cleared) if `vr.Replicate` returns.

We've seen the following happen:
* `vr.Replicate(ctx)` fails with some error
* the error is recorded via `ct.lastWorkflowError.record(err)`
* the error is seen as recoverable (`isUnrecoverableError(err)` returns false) and `ct.lastWorkflowError.shouldRetry()` returns true as the error happened for the first time just now
* Vreplication is retried and restarted successfully
* ... a long time passes (hours / days / weeks)
* `vr.Replicate(ctx)` fails with the same error as before
* the error is passed to `ct.lastWorkflowError.record(err)`, but as it's the same error, [no new timestamp is recorded](https://github.com/vitessio/vitess/blob/ece6501975604ef5d0be3ab256145047e13d9c00/go/vt/vttablet/tabletmanager/vreplication/last_error.go#L55-L59)
[...]

PR #12114, opened 2023-01-17
VReplication Last Error: retry error if it happens after timeout

rohit-nayak-ps:
## Description

This PR fixes a bug where a VReplication workflow doesn't restart if the same error happened far enough in the past as to be beyond the defined `--vreplication_max_time_to_retry_on_error`.

See related issue for more details.

## Related Issue(s)
https://github.com/vitessio/vitess/issues/12101

## Checklist

### DIFF
--- a/go/vt/vttablet/tabletmanager/vreplication/controller.go
+++ b/go/vt/vttablet/tabletmanager/vreplication/controller.go
@@ -78,13 +78,12 @@ func newController(ctx context.Context, params map[string]string, dbClientFactor
 	}
 
 	ct := &controller{
-		vre:               vre,
-		dbClientFactory:   dbClientFactory,
-		mysqld:            mysqld,
-		blpStats:          blpStats,
-		done:              make(chan struct{}),
-		source:            &binlogdatapb.BinlogSource{},
-		lastWorkflowError: newLastError("VReplication Controller", maxTimeToRetryError),
+		vre:             vre,
+		dbClientFactory: dbClientFactory,
+		mysqld:          mysqld,
+		blpStats:        blpStats,
+		done:            make(chan struct{}),
+		source:          &binlogdatapb.BinlogSource{},
 	}
 	log.Infof("creating controller with cell: %v, tabletTypes: %v, and params: %v", cell, tabletTypesStr, params)
 
--- a/go/vt/vttablet/tabletmanager/vreplication/controller.go
+++ b/go/vt/vttablet/tabletmanager/vreplication/controller.go
@@ -95,6 +95,7 @@ func newController(ctx context.Context, params map[string]string, dbClientFactor
 	}
 	ct.id = int32(id)
 	ct.workflow = params["workflow"]
+	ct.lastWorkflowError = newLastError(fmt.Sprintf("VReplication controller %d for workflow %q", ct.id, ct.workflow), maxTimeToRetryError)
 
 	state := params["state"]
 	blpStats.State.Set(state)
--- a/go/vt/vttablet/tabletmanager/vreplication/last_error.go
+++ b/go/vt/vttablet/tabletmanager/vreplication/last_error.go
@@ -33,11 +33,13 @@ type lastError struct {
 	name           string
 	err            error
 	firstSeen      time.Time
+	lastSeen       time.Time
 	mu             sync.Mutex
 	maxTimeInError time.Duration // if error persists for this long, shouldRetry() will return false
 }
 
 func newLastError(name string, maxTimeInError time.Duration) *lastError {
+	log.Infof("Created last error %s:%s", name, maxTimeInError)
 	return &lastError{
 		name:           name,
 		maxTimeInError: maxTimeInError,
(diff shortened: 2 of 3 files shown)


go/vt/vttablet/tabletmanager/vreplication/last_error.go
```diff
@@ -48,15 +50,27 @@ func (le *lastError) record(err error) {
 	le.mu.Lock()
 	defer le.mu.Unlock()
 	if err == nil {
+		log.Infof("resetting last error")
```
mattlord, line 53: We should include the name here as well so that we can follow the flow in the info log and tie the messages together.

go/vt/vttablet/tabletmanager/vreplication/last_error.go
```diff
@@ -33,11 +33,13 @@ type lastError struct {
 	name           string
 	err            error
 	firstSeen      time.Time
+	lastSeen       time.Time
 	mu             sync.Mutex
 	maxTimeInError time.Duration // if error persists for this long, shouldRetry() will return false
 }
 
 func newLastError(name string, maxTimeInError time.Duration) *lastError {
+	log.Infof("Created last error %s:%s", name, maxTimeInError)
```
mattlord, line 42: IMO, we should change the usage here to add more value to these log messages:
```
func newController(ctx context.Context, params map[string]string, dbClientFactory func() binlogplayer.DBClient, mysqld mysqlctl.MysqlDaemon, ts *topo.Server, cell, tabletTypesStr string, blpStats *binlogplayer.Stats, vre *Engine) (*controller, error) {
	if blpStats == nil {
		blpStats = binlogplayer.NewStats()
	}

	ct := &controller{
		vre:               vre,
		dbClientFactory:   dbClientFactory,
		mysqld:            mysqld,
		blpStats:          blpStats,
		done:              make(chan struct{}),
		source:            &binlogdatapb.BinlogSource{},
	}
	log.Infof("creating controller with cell: %v, tabletTypes: %v, and params: %v", cell, tabletTypesStr, params)

	// id
	id, err := strconv.Atoi(params["id"])
	if err != nil {
		return nil, err
	}
	ct.id = uint32(id)
	ct.workflow = params["workflow"]

        ct.lastWorkflowError = newLastError(fmt.Sprintf("VReplication controller %d for workflow %q", ct.id, ct.workflow), maxTimeToRetryError),
```

go/vt/vttablet/tabletmanager/vreplication/last_error.go
```diff
@@ -48,15 +50,27 @@ func (le *lastError) record(err error) {
...
 	defer le.mu.Unlock()
 	if err == nil {
+		log.Infof("resetting last error")
 		le.err = nil
 		le.firstSeen = time.Time{}
+		le.lastSeen = time.Time{}
 		return
 	}
 	if !vterrors.Equals(err, le.err) {
+		log.Infof("got new last error %+v for %s, was %+v", err, le.name, le.err)
```
mattlord, line 60: IMO we should consistently use sentence capitalization with log messages.

go/vt/vttablet/tabletmanager/vreplication/last_error.go
```diff
@@ -48,15 +50,27 @@ func (le *lastError) record(err error) {
...
 		le.firstSeen = time.Now()
+		le.lastSeen = time.Now()
 		le.err = err
+	} else {
+		// same error seen
+		log.Infof("got same last error for %s seen %+v, last seen at %s after %dms", le.name, le.err, le.lastSeen, int(time.Since(le.lastSeen).Milliseconds()))
```
mattlord, line 66: I think this would be better (`%q` is especially nice when there's whitespace in the value so that it's clear what the variable value is):
```
log.Infof("Got the same last error for %q: %+v ; first seen at %s and last seen %dms ago", le.name, le.err, le.firstSeen, int(time.Since(le.lastSeen).Milliseconds()))
```

go/vt/vttablet/tabletmanager/vreplication/controller.go
```diff
@@ -95,6 +95,7 @@ func newController(ctx context.Context, params map[string]string, dbClientFactor
 	}
 	ct.id = int32(id)
 	ct.workflow = params["workflow"]
+	ct.lastWorkflowError = newLastError(fmt.Sprintf("VReplication controller %d for workflow %q", ct.id, ct.workflow), maxTimeToRetryError)
```
mattlord, line 98: Why are we doing this here? Can you please add a comment to clarify? I don't think it's related to the bug we're fixing is it? Did we just realize that each workflow's controller should have a non-nil newLastError struct (regardless of whether there's ever been an error or not?

go/vt/vttablet/tabletmanager/vreplication/last_error_test.go
```diff
@@ -35,17 +35,18 @@ func TestLastError(t *testing.T) {
...
 		require.True(t, le.shouldRetry())
 		for i := 1; i < 10; i++ {
 			le.record(err2)
+			time.Sleep(1 * time.Millisecond)
 		}
 		require.True(t, le.shouldRetry())
 		time.Sleep(150 * time.Millisecond)
 		le.record(err2)
-		require.False(t, le.shouldRetry())
+		require.True(t, le.shouldRetry())
```
mattlord, line 49: Is this meant to test and confirm the bug fix / new behavior? If so, could you please add a comment as to how it does that? I'm guessing it's because `maxTimeToRetryError` is less than 150ms?

mattlord (commented):
The meat of this LGTM, I just had some questions about other changes.

mattlord (approved):
Approving since at this point my questions/requests are only code comment related. This way you can address those code comment related requests as you think best and are unblocked at that point.

go/vt/vttablet/tabletmanager/vreplication/controller.go
rohit-nayak-ps: We were already initializing it while creating the controller. I was addressing your review comments on adding more detail to the name for better observability. I removed the previous init since it was superfluous.

shlomi-noach (approved):
I feel like a couple more unit tests and comments could help clarify the behavior of LastErorr, but otherwise LGTM

shlomi-noach (approved):
thank you :heart:

shlomi-noach:
feel free to ignore `onlineddl_vrepl` for now.

PR #12119, opened 2023-01-18
Split replication stopping into permanent and temporary for replication manager

GuptaManan100:
Signed-off-by: Manan Gupta <manan@planetscale.com>

## Checklist


GuptaManan100:
Instead, we are going to remove the replication manager entirely #12149.

PR #12124, opened 2023-01-19
Views Support: Updating Views in VSchema for query serving

harshit-gangal:
## Description

This PR completed the loop. This PR adds views tracker in VTGate which will load the new views definition when it changes and updates the VTGate's VSchema views.

## Related Issue(s)

- #11559 

## Checklist


go/test/endtoend/vtgate/schematracker/sharded/st_sharded_test.go
```diff
@@ -280,3 +217,50 @@ func TestDMLOnNewTable(t *testing.T) {
...
+	// insert some data
+	_ = utils.Exec(t, conn, "alter table t8 add column new_col varchar(50)")
+	time.Sleep(2 * time.Second)
+
+	_ = utils.Exec(t, conn, "insert into t8(id8, new_col) values (1, 'V')")
+	defer utils.Exec(t, conn, "delete from t8")
+
+	// create a view with t8, having the new column.
+	_ = utils.Exec(t, conn, "create view t8_view as select * from t8")
+	time.Sleep(2 * time.Second)
```
GuptaManan100, line 260: Don't you want to create the view before adding the column? Otherwise isn't it just created when the table already has the new column?

go/vt/vtgate/vindexes/vschema.go
```diff
@@ -175,24 +175,32 @@ type KeyspaceSchema struct {
...
+
 // MarshalJSON returns a JSON representation of KeyspaceSchema.
 func (ks *KeyspaceSchema) MarshalJSON() ([]byte, error) {
-	return json.Marshal(struct {
-		Sharded  bool              `json:"sharded,omitempty"`
-		Tables   map[string]*Table `json:"tables,omitempty"`
-		Vindexes map[string]Vindex `json:"vindexes,omitempty"`
-		Error    string            `json:"error,omitempty"`
-	}{
+	ksJ := ksJSON{
```
GuptaManan100, line 188: Is there no test for this JSON output, because I don't see any corresponding test change?

go/vt/vtgate/schema/tracker.go
```diff
@@ -225,6 +297,47 @@ func (t *Tracker) updateTables(keyspace string, res *sqltypes.Result) {
...
+func (t *Tracker) updatedViewSchema(th *discovery.TabletHealth) bool {
+	viewsUpdated := th.Stats.ViewSchemaChanged
+	views, err := sqltypes.BuildBindVariable(viewsUpdated)
+	if err != nil {
+		log.Errorf("failed to read updated views from TabletHealth: %v", err)
+		return false
+	}
+	bv := map[string]*querypb.BindVariable{"viewNames": views}
+	res, err := th.Conn.Execute(t.ctx, th.Target, mysql.FetchUpdatedViews, bv, 0, 0, nil)
+	if err != nil {
```
GuptaManan100, line 309: Shouldn't we be using an RPC for it, so that it is easier to make changes to the schema and be backward compatible?

go/test/endtoend/vtgate/schematracker/sharded/st_sharded_test.go
harshit-gangal: I can add that test, in that test, view should not contain the new column.

go/vt/vtgate/vindexes/vschema.go
harshit-gangal: I tested it manually, I will add a test for it.

go/vt/vtgate/schema/tracker.go
harshit-gangal: That is the next step. A new PR will be raised for that. This feature is now tested e2e by doing the existing way of how table schema is retrieved.

go/vt/vtgate/schema/tracker.go
```diff
@@ -65,39 +67,86 @@ const defaultConsumeDelay = 1 * time.Second
...
+	// We must clear out any previous view definition before loading it here as this is called
+	// whenever a shard's primary tablet starts and sends the initial signal. Without
+	// clearing out the previous view definition we can end up with duplicate entries when the
+	// tablet is simply restarted or potentially when we elect a new primary.
+	t.clearKeyspaceViews(target.Keyspace)
+	t.updateViews(target.Keyspace, fvRes)
+	log.Infof("finished loading views for keyspace %s. Found %d views", target.Keyspace, len(fvRes.Rows))
+
 	return nil
 }
```
arthurschreiber, line 151: `loadTables` takes a lock on `t.mu`, shouldn't `loadViews` also take the same lock?

go/vt/vtgate/schema/tracker_test.go
```diff
@@ -308,3 +308,127 @@ func TestTrackerGetKeyspaceUpdateController(t *testing.T) {
...
+			sbc := sandboxconn.NewSandboxConn(tablet)
+			ch := make(chan *discovery.TabletHealth)
+			tracker := NewTracker(ch, "", true)
+			tracker.tables = nil // making tables map nil - so load keyspace does not try to load the tables information.
+			tracker.consumeDelay = 1 * time.Millisecond
+			tracker.Start()
+			defer tracker.Stop()
+
+			results := []*sqltypes.Result{{}}
+			for _, d := range tcase.deltas {
```
systay, line 386: this is pretty hard to read code. can we make it easier to understand you think?

go/vt/vtgate/schema/update_controller.go
```diff
@@ -110,6 +110,18 @@ func (u *updateController) getItemFromQueueLocked() *discovery.TabletHealth {
 					item.Stats.TableSchemaChanged = append(item.Stats.TableSchemaChanged, table)
 				}
 			}
+			for _, view := range u.queue.items[i].Stats.ViewSchemaChanged {
```
systay, line 113: a comment explaining what this loop does?

go/vt/vtgate/schema/update_controller_flaky_test.go
```diff
@@ -177,3 +177,153 @@ func TestMultipleUpdatesFromDifferentShards(t *testing.T) {
...
+		viewUpdates []string
+	}
+	type testCase struct {
+		updateViews                  []string
+		signalExpected, initExpected int
+		inputs                       []input
+		delay                        time.Duration
+		init, initFail, updateFail   bool
+	}
+	tests := []testCase{{
```
systay, line 193: more tests that are hard to grokk. can we add helpful names to the different tests to help readability?

go/mysql/schema.go
```diff
@@ -129,6 +129,12 @@ order by table_name, ordinal_position`
 	SelectFromViewsTable = `SELECT TABLE_NAME FROM _vt.views WHERE TABLE_NAME IN ::TABLE_NAME`
 
 	SelectAllViews = `SELECT TABLE_NAME, UPDATED_AT FROM _vt.views`
+
+	// FetchUpdatedViews queries fetches information about updated views
+	FetchUpdatedViews = `SELECT TABLE_NAME, VIEW_DEFINITION, CREATE_STATEMENT FROM _vt.views WHERE TABLE_NAME IN ::viewNames`
+
+	// FetchViews queries fetches all views
+	FetchViews = `SELECT TABLE_NAME, VIEW_DEFINITION, CREATE_STATEMENT FROM _vt.views`
```
mattlord, line 137: Any particular reason to use all caps for the views queries? Nothing wrong with it, it's just non-standard within Vitess and even this package.

go/test/endtoend/vtgate/schematracker/sharded/st_sharded_test.go
```diff
@@ -280,3 +219,65 @@ func TestDMLOnNewTable(t *testing.T) {
...
+
+	// insert some data
+	_ = utils.Exec(t, conn, "insert into t2 (id3, id4) values (1, 10), (2, 20), (3, 30)")
+	defer utils.Exec(t, conn, "delete from t2")
+
+	selQuery := "select sum(id4) from t2 where id4 > 10"
+
+	// create a view
+	_ = utils.Exec(t, conn, "create view test_view as "+selQuery)
+	time.Sleep(2 * time.Second)
```
mattlord, line 238: This vtgate test is already a little flaky in my experience — especially with GH Actions is otherwise struggling/slow. Using sleeps like this is likely to make it more so. What is the point of this sleep? If we're waiting for some condition to be true, we should check and wait for it with a timeout. IMO these [new] `waitForX()` functions should replace all of the sleeps.

go/vt/vtgate/schema/update_controller_flaky_test.go
```diff
@@ -177,3 +177,153 @@ func TestMultipleUpdatesFromDifferentShards(t *testing.T) {
 		})
 	}
 }
+
+func TestViewsUpdates(t *testing.T) {
```
mattlord, line 181: Do we know why these are flaky tests? I want to tackle the ones we've marked as flaky some day...

go/test/endtoend/vtgate/schematracker/sharded/st_sharded_test.go
frouioui: Agreed with @mattlord, we could run a query that selects the view in a loop with some sort of timeout. `utils.AssertMatchesWithTimeout` allows us to do that.

go/vt/vtgate/schema/tracker_test.go
```diff
@@ -308,3 +308,127 @@ func TestTrackerGetKeyspaceUpdateController(t *testing.T) {
 	assert.NotNil(t, ks2.reloadKeyspace, "ks2 needs to be initialized")
 	assert.Nil(t, ks3.reloadKeyspace, "ks3 already initialized")
 }
+
+func TestViewsTracking(t *testing.T) {
+	target := &querypb.Target{Cell: "aa", Keyspace: "ks", Shard: "-80", TabletType: topodatapb.TabletType_PRIMARY}
```
frouioui, line 313: We should add a comment as to why this test exist.

go/vt/vtgate/schema/update_controller_flaky_test.go
frouioui, line 181: Same as above, we should add a comment for this test.

go/vt/vtgate/schema/update_controller_flaky_test.go
```diff
@@ -177,3 +177,153 @@ func TestMultipleUpdatesFromDifferentShards(t *testing.T) {
...
+				}
+				d := &discovery.TabletHealth{
+					Tablet:  tablet,
+					Target:  target,
+					Serving: true,
+					Stats:   &querypb.RealtimeStats{ViewSchemaChanged: in.viewUpdates},
+				}
+				if test.delay > 0 {
+					time.Sleep(test.delay)
+				}
```
frouioui, line 311: This might be a source of flakiness

go/vt/vtgate/schema/update_controller.go
harshit-gangal: done

go/mysql/schema.go
harshit-gangal: there is nothing in particular.
changed queries to lowecase

go/vt/vtgate/schema/update_controller_flaky_test.go
harshit-gangal: added test descriptions.

go/vt/vtgate/schema/update_controller_flaky_test.go
harshit-gangal: I believe the reason it is flaky is sometimes health check updates from the shard can be consolidated into a single call if some go-routine is delayed to run the fetch call.

go/vt/vtgate/schema/update_controller_flaky_test.go
harshit-gangal: The test is to see that multiple signals happen because they are received at different time intervals.

go/test/endtoend/vtgate/schematracker/sharded/st_sharded_test.go
harshit-gangal: handled.

go/test/endtoend/utils/utils.go
```diff
@@ -232,6 +236,47 @@ func WaitForAuthoritative(t *testing.T, vtgateProcess cluster.VtgateProcess, ks,
 	}
 }
 
+// WaitForColumn waits for a table's column to be present
+func WaitForColumn(t *testing.T, vtgateProcess cluster.VtgateProcess, ks, tbl, col string) error {
```
systay, line 240: nit: since you are sending in `t *testing.T`, we don't really need to return an error, right? we can just fail in here, right?

systay (approved):

go/test/endtoend/utils/utils.go
harshit-gangal: I just want the author of the test to decide if they want to continue or fail.

frouioui (approved):

PR #12083, opened 2023-01-12
Remove Deprecated flags, code and stats.

GuptaManan100:
## Description

This PR removed a lot of deprecated flags, code and stats which were deprecated in previous Vitess releases. The following changes have been made - 

1. `Keyspace` and `TableName` fields have been removed from LogStats.
2. `QueriesProcessed`, `QueriesRoutedStats` and `QueryRowCounts` have been removed.
3. `enable_semi_sync` flag has been removed.
4. DeprecatedDBName field which wasn't been written to is also removed.

## Related Issue(s)

- Fixes #5960 
- #11749 

## Checklist

-   [x] "Backport to:" labels have been added if this change should be back-ported
-   [x] Tests were added or are not required
-   [x] Documentation was added or is not required

PR #12128, opened 2023-01-20
Remove removed flags from being used for v16+ binaries

GuptaManan100:
## Description

This PR removes the flags removed from v16 to be set on binaries from that and later releases. This change is required because we run upgrade-downgrade tests which spawn binaries from the next release too. So the flags removed in #12083 and #12085 have to be put behind a condition check to see if the spawned binary is from a version higher than 15 or not.

This PR makes these changes to fix the upgrade-downgrade tests which are currently broken as seen in https://github.com/vitessio/vitess/actions/runs/3949422838/jobs/6763433594

## Checklist

### DIFF
--- a/go/test/endtoend/cluster/vtctld_process.go
+++ b/go/test/endtoend/cluster/vtctld_process.go
@@ -58,8 +58,6 @@ func (vtctld *VtctldProcess) Setup(cell string, extraArgs ...string) (err error)
 		"--topo_global_server_address", vtctld.CommonArg.TopoGlobalAddress,
 		"--topo_global_root", vtctld.CommonArg.TopoGlobalRoot,
 		"--cell", cell,
-		"--workflow_manager_init",
-		"--workflow_manager_use_election",
 		"--service_map", vtctld.ServiceMap,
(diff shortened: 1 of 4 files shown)


go/test/endtoend/cluster/vtctld_process.go
```diff
@@ -58,15 +58,25 @@ func (vtctld *VtctldProcess) Setup(cell string, extraArgs ...string) (err error)
...
+	if err != nil {
+		return err
+	}
+	// workflow_manager_init and workflow_manager_use_election are removed in v16 and shouldn't be set on any release v16+
+	if majorVersion <= 15 {
+		vtctld.proc.Args = append(vtctld.proc.Args,
+			"--workflow_manager_init",
+			"--workflow_manager_use_election")
+	}
+
```
deepthi, line 79: None of the endtoend tests actually ever tested workflow manager. We don't need this conditional code at all.

deepthi (approved):

go/test/endtoend/cluster/vtctld_process.go
GuptaManan100: Done

harshit-gangal (approved):

Issue #12132, opened 2023-01-22
SchemaDiff: normalize a boolean column to tinyint(1)


rohit-nayak-ps:
### Feature Description

If a user specifies a column of type `boolean`  in a DDL, MySQL actually stores it as a `tinyint(1)`. This has implications for `schemadiff`: if the desired schema specified to `schemadiff` has such a column, then if we compare it to the output of `show create table` we will always get a diff. The diff will be an alter trying to convert the tinyint(1) to boolean, which will in practice be a no-op.

`schemadiff` already normalizes some datatypes, like `float4` (to `float), `real` (to double) etc. We should do the the same for `booelan`.

### Use Case(s)

This issue was detected while implementing the new schema init logic (https://github.com/vitessio/vitess/pull/11520). If the desired schema for one of the `_vt` tables had a column defined as `boolean` instead of `tinyint(1)`, it would always generate such no-op `alters` every time the primary inited.

We worked around this by changing the datatypes to `tinyint(1)`, but it is probably good to implement this normalization, since users of OnlineDDL might also experience this.

PR #12133, opened 2023-01-22
SchemaDiff: normalize boolean columns to tinyint(1), and, add TableQualifier hint to force adding a db qualifier to the alter diff.

rohit-nayak-ps:
## Description
This PR adds two changes to schemadiff:

1. Normalizes boolean columns to tinyint(1) in the normalization phase of schemadiff, as reported in https://github.com/vitessio/vitess/issues/12132.

2. For the use of functionality such as in https://github.com/vitessio/vitess/pull/11520, it is useful for the generated diff create/alter scripts to also provide the database qualifier to the table name.  

    In that PR we have the desired schema in a static file in the form `create table if not exists _vt.copy_state ...`. We get the output of `show create table _vt.copy_state`, which _does not_ have the _vt qualifier. We then ask `schemadiff` to diff these two. Currently the `alter` that is generated does not have the `_vt` qualifier because the current schema (the output of `show create table`) has no qualifier.

    This PR adds a `TableQualifier` `DiffHint` which can be used to force the addition of the `_vt` qualifier in the generated `alter`.

## Related Issue(s)
https://github.com/vitessio/vitess/issues/12132

## Checklist


go/vt/schemadiff/table.go
```diff
@@ -435,6 +435,27 @@ func (c *CreateTableEntity) normalizeColumnOptions() {
...
+		if col.Type.Type == "boolean" {
+			col.Type.Type = "tinyint"
+			col.Type.Length = &sqlparser.Literal{
+				Type: sqlparser.IntVal,
+				Val:  "1",
+			}
+
+			if col.Type.Options.Default != nil {
+				defaultVal := "0"
+				if col.Type.Options.Default.(sqlparser.BoolVal) {
```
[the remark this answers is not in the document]
rohit-nayak-ps: Good catch. Implemented your suggestions.

rohit-nayak-ps:
I added another  functionality to this PR (as explained in the Description). I can move this into its own PR if that works better.

go/vt/schemadiff/types.go
```diff
@@ -94,6 +94,11 @@ const (
 	TableCharsetCollateIgnoreAlways
 )
 
+const (
+	TableQualifierUseSpecified int = iota
+	TableQualifierUseOriginal
+)
```
dbussink, line 100: @rohit-nayak-ps Hmm, doesn't this change the default behavior if `TableQualifierUseSpecified` is the `0` option? Shouldn't these be flipped to not change the default behavior and only explicitly opt into this?

rohit-nayak-ps: Those enums are not intuitive: was going to ask for better naming. That is currently the default that is already in `schemadiff`: it uses the qualifier of the desired (`to`) schema. So essentially `UseOriginal` is the one in `from`.

dbussink (approved):

go/vt/schemadiff/diff_test.go
```diff
@@ -52,6 +53,36 @@ func TestDiffTables(t *testing.T) {
...
+		{
+			name:     "change of columns, boolean type",
+			from:     "create table t(id int primary key)",
+			to:       "create table t(id int primary key, i int, b boolean)",
+			diff:     "alter table t add column i int, add column b tinyint(1)",
+			cdiff:    "ALTER TABLE `t` ADD COLUMN `i` int, ADD COLUMN `b` tinyint(1)",
+			action:   "alter",
+			fromName: "t",
+			toName:   "t",
+		},
```
shlomi-noach, line 65: Please add another test comparing two tables, one with `bool`, the other with `tinyint(1)`, that expects an empty diff.

go/vt/schemadiff/table.go
```diff
@@ -724,6 +750,10 @@ func (c *CreateTableEntity) TableDiff(other *CreateTableEntity, hints *DiffHints
 	alterTable := &sqlparser.AlterTable{
 		Table: c.CreateTable.Table,
 	}
+	if hints.TableQualifierHint == TableQualifierDeclared {
+		alterTable.Table.Qualifier = other.Table.Qualifier
+	}
+
```
shlomi-noach, line 756: should this only apply to `AlterTableDiff` or also to `AlterViewDiff`? And to `CREATE/DROP` as well? etc. etc.?

shlomi-noach (commented):
Please see questions/comments inline

shlomi-noach (approved):

PR #12135, opened 2023-01-24
Generating copy-on-rewrite logic

systay:
## Description
Adds a new syntax tree rewrite utility. It makes it possible to share parts of the syntax tree with the original, so it should decrease memory pressure somewhat during planning. It also traverses the tree only once to do both operations, before we first needed to clone the full tree and only after could we rewrite it.

Additionally, it can report back to the user which nodes got cloned, which makes it possible to copy semantic information between the two trees.

## Checklist


go/vt/sqlparser/cow.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+	cow struct {
+		pre    func(node, parent SQLNode) bool
+		post   func(cursor *CopyOnWriteCursor)
+		cloned func(old, new SQLNode)
+		cursor CopyOnWriteCursor
+	}
+)
+
+func CopyOnRewrite(node SQLNode, pre func(node, parent SQLNode) bool, post func(cursor *CopyOnWriteCursor)) SQLNode {
```
systay, line 34: comment

go/vt/sqlparser/cow.go
```diff
@@ -0,0 +1,95 @@
...
+	c.cursor.replaced = nil
+	c.post(&c.cursor)
+	if c.cursor.replaced != nil {
+		return c.cursor.replaced, true
+	}
+	return node, false
+}
+
+// StopTreeWalk aborts the current tree walking. No more nodes will be visited, and the rewriter will exit out early
+func (c *CopyOnWriteCursor) StopTreeWalk() {
```
systay, line 76: Maybe we should use a recovered panic here instead of setting a boolean and checking it everywhere

vmg (approved):
Looks great. I discussed with @systay the unified interface for `pre`/`post` and I think this is a great design!

go/vt/sqlparser/cow.go
frouioui: I thought that `recover` was a source of performance loss

frouioui (approved):

PR #12137, opened 2023-01-24
schemadiff: TableCharsetCollateStrategy hint

shlomi-noach:
## Description

`schemadiff`: adding a new hint, named `TableCharsetCollateStrategy`, which controls how `schemadiff` treats diffs in a table's `CHARSET` and `COLLATE` option values:

- `TableCharsetCollateStrict` - default, if charsets or collations are different, that's a diff.
- `TableCharsetCollateIgnoreEmpty` - if one of the tables has no value for `CHARSET` or `COLLATE`, ignore the diff (works whether the source table or target table have empty values)
- `TableCharsetCollateIgnoreAlways` - never generate a table charset/collate diff

## Related Issue(s)

## Checklist


rohit-nayak-ps (approved):
lgtm

dbussink (approved):

PR #12138, opened 2023-01-24
testutils: use a rows parser for test comparisons

vmg:
## Description

A pretty common pattern in Vitess end2end tests is comparing the results of a query against a hardcoded string that represents a `%v`-formatted syntax of the rows we're expecting. It usually looks something like this:

```go
assert.Equals(t, "[[INT64(1) INT64(8)] [INT64(1) INT64(9)]]", fmt.Sprintf("%v", rows))
```

This is all well and good! There are several helpers that print the rows directly without having to call `Sprintf` and that compare them directly. I like the literal strings to define the expectations for row results.

However, shit gets funky when testing queries whose result from Vitess/MySQL doesn't have consistent ordering. How can we handle such comparisons when the `sql.Rows` returned by Vitess can have any ordering but our expectation string is obviously fixed?

This is how it's done at the moment:

```go
actual := fmt.Sprintf("%v", qr.Rows)
assert.Equal(mcmp.t, utils.SortString(expected), utils.SortString(actual))
```

This... doesn't seem generally safe? Is `SortString` doing something _very smart_ here? If you look at its implementation, you can see that, huh, it's doing the opposite of something very smart. It re-arranges all the individual bytes of the string alphabetically and returns that.

So, for a rows string like `"[[INT64(1) INT64(8)] [INT64(1) INT64(9)]]"`, it generates `"   (((())))114444666689IIIINNNNTTTT[[[]]]"` and compares that against the other formatted rows. The many ways in which this is not a generally safe way to compare row results are left to the reader. I bet you can come up with hundreds!

So, onto the actual content of the PR: this PR introduces a generally safe way to compare unordered sets of rows! :tada: 

It works by implementing a simple Lexer + Parser that can take a "row string" like the ones we use in our tests and converts it into an actual `sql.Row` (i.e. a `[]sql.Value` slice). Then, it performs an ordering-agnostic comparison of the `sql.Value` entries against the values returned by Vitess/MySQL. It works very well in practice! We've tested it internally at PlanetScale.

I'm hoping this will catch some actual mismatches in our end2end tests down the road.

## Checklist


go/test/utils/rows_test.go
```diff
@@ -0,0 +1,172 @@
...
+	"[[INT64(2) INT64(4)]]",
+	"[[INT64(1) INT64(1)] [INT64(1) INT64(2)] [INT64(1) INT64(3)]]",
+	"[[INT64(1) INT64(1)] [INT64(1) INT64(2)] [INT64(1) INT64(3)] [INT64(1) INT64(4)] [INT64(1) INT64(5)] [INT64(1) INT64(6)]]",
+	"[[INT64(1) INT64(1)] [INT64(1) INT64(2)] [INT64(1) INT64(3)]]",
+}
+
+func TestRowParsing(t *testing.T) {
+	for _, r := range TestRows {
+		output, err := ParseRows(r)
+		if err != nil {
```
systay, line 162: if only we had some test library that allowed us to write `NoError(t, err)`

/s

systay (approved):
This is really great. Thank you!

frouioui (approved):

vmg:
:tada:

PR #12139, opened 2023-01-24
Clean up ColumnType uses

systay:
## Description
Small PR that makes an AST type consistently used as reference instead of value.

`ColumnType` was sometimes used as a value type although only the reference type implements the SQLNode interface, which lead to unwanted behaviour such as these fields not being seen by the AST tooling (Equality, Cloning and Rewriting).

## Checklist


harshit-gangal (approved):

harshit-gangal:
needs sizegen run

frouioui (approved):

Issue #12134, opened 2023-01-23
Metrics for prepared statements


maxenglander:
As far as I can tell VTGate does not offer metrics for prepared statements. Or rather, it does but it uses the same stat used by `Execute`:

https://github.com/vitessio/vitess/blob/784d79cb58be70e25550257649f576e23010d522/go/vt/vtgate/vtgate.go#L533-L537

It would be helpful if there were a way to distinguish `Prepare` calls from other calls. Some of the Vitess users that my team supports can sometimes unknowningly back themselves into a performance corner using prepared statements, and having VTG-level visibility into prepared statement usage would help us help them.

If there is already a way to do this with VTG, please just point me the manual I should read. If not, I'll happily contribute a change if that's the way to go here.

PR #12141, opened 2023-01-24
vtgate: metrics for prepared stmts

maxenglander:
## Description

As far as I can tell VTGate does not offer metrics for prepared statements. Or rather, it does but it uses the same stat used by `Execute`:

https://github.com/vitessio/vitess/blob/784d79cb58be70e25550257649f576e23010d522/go/vt/vtgate/vtgate.go#L533-L537

It would be helpful if there were a way to distinguish `Prepare` calls from other calls.

## Changes

 * Set the `Prepare` log and stats key to `Prepare`
 * Note the change in release summary

## Related Issue(s)

Fixes https://github.com/vitessio/vitess/issues/12134

## Checklist


maxenglander:
I'm not sure how to address the `NeedsDescriptionUpdate` label.

rsajwani (commented):
changes look good to me. But i will let queryserving team to sign-off.

maxenglander:
Hey @harshit-gangal @systay pardon the ping, just checking in case you haven't seen this one yet

harshit-gangal (approved):

harshit-gangal:
This is a breaking change on the metrics part if not log.

cc: @deepthi

doc/releasenotes/17_0_0_summary.md
```diff
@@ -3,11 +3,66 @@
...
 - **[Major Changes](#major-changes)**
+  - **[Breaking Changes](#breaking-changes)**
   - **[New command line flags and behavior](#new-flag)**
     - [Builtin backup: read buffering flags](#builtin-backup-read-buffering-flags)
 
 ## <a id="major-changes"/> Major Changes
 
+### <a id="breaking-changes"/> Breaking changes
+
+#### Dedicated stats for VTGate Prepare operations
```
deepthi, line 14: This note is too long. Can it be condensed? GH imposes a limit on how much content you can have on a release. We exceeded that for v16-rc1.

deepthi:
> This is a breaking change on the metrics part if not log.
> 
> cc: @deepthi

That is fine. It was always a bug to include Prepare metrics in Execute metrics.

doc/releasenotes/17_0_0_summary.md
deepthi: I would not call this a breaking change. It is a bug fix. Yes people's metrics will show different numbers if they are using Prepare, but they won't become non-functional.

deepthi:
Also, isn't there a stats unit test you can piggyback on? We should be adding a unit test for every bug fix.

doc/releasenotes/17_0_0_summary.md
maxenglander: I won't argue you about naming, but conceivably this could break alerts for people if they make heavy use of prepared statements and are relying on Execute metrics to indirectly monitor them. People using Vitess might expect to find this kind of information in a breaking changes section, even if we (you and other maintainers) don't think of it that way. 

Will change it as you've suggested, just offering my 2c!

maxenglander:
AFAICT the failing integration tests is unrelated to these changes.

```
=== RUN   TestVSchema
    vschema_test.go:115: Query: SHOW VSCHEMA TABLES (-want +got):
          string(
        - 	"[]",
        + 	`[[VARCHAR("dual")]]`,
          )
        
        Got:[[VARCHAR("dual")]]
    vschema_test.go:131: Query: SHOW VSCHEMA TABLES (-want +got):
          strings.Join({
          	`[[VARCHAR("`,
        + 	`dual")] [VARCHAR("`,
          	`main")] [VARCHAR("vt_user")]]`,
          }, "")
        
        Got:[[VARCHAR("dual")] [VARCHAR("main")] [VARCHAR("vt_user")]]
--- FAIL: TestVSchema (0.07s)

```

doc/releasenotes/17_0_0_summary.md
deepthi: You make some good points. Yes, let's keep it in the breaking changes section.

doc/releasenotes/17_0_0_summary.md
```diff
@@ -4,12 +4,13 @@
 
 - **[Major Changes](#major-changes)**
   - **[Breaking Changes](#breaking-changes)**
-    - [Deprecated Stats](#deprecated-stats)
```
maxenglander, line 7: Moved this down under `Deprecations and Deletions`

doc/releasenotes/17_0_0_summary.md
```diff
@@ -19,14 +20,30 @@
 
 When using TLS with `vtgr`, we now default to TLS 1.2 if no other explicit version is configured. Configuration flags are provided to explicitly configure the minimum TLS version to be used. 
 
-#### <a id="deprecated-stats"/>Deprecated Stats
+#### <a id="dedicated-vtgate-prepare-stats"> Dedicated stats for VTGate Prepare operations
 
-These stats are deprecated in v17.
+Prior to v17 Vitess incorrectly combined stats for VTGate Execute and Prepare operations under a single stats key (`Execute`). In v17 Execute Prepare operations generate stats under independent stats keys.
```
deepthi, line 25:
```suggestion
Prior to v17 Vitess incorrectly combined stats for VTGate Execute and Prepare operations under a single stats key (`Execute`). In v17 Execute and Prepare operations generate stats under independent stats keys.
```

doc/releasenotes/17_0_0_summary.md
```diff
@@ -168,4 +185,13 @@ Some notes to help understand these metrics:
...
+* The deprecated `automation` and `automationservice` protobuf definitions and associated client and server packages have been removed.
+
+### <a id="deprecated-stats"/>Deprecated Stats
+
+These stats are deprecated in v17.
+
+| Deprecated stat | Supported alternatives |
+|-|-|
+| `backup_duration_seconds` | `BackupDurationNanoseconds` |
+| `restore_duration_seconds` | `RestoreDurationNanoseconds` |
```
deepthi, line 197: Was this missed in a previous PR? Am wondering because it is unrelated to the rest of this PR.

deepthi (approved):
We can merge after the last comment is addressed.

doc/releasenotes/17_0_0_summary.md
maxenglander: No they weren't missed, but I moved them further down in the release notes because it seems more appropriate. Can revert if you prefer.

deepthi (approved):

PR #12144, opened 2023-01-25
DDL: do not Reload() for 'CREATE TEMPORARY' and 'DROP TEMPORARY' statements

shlomi-noach:
## Description

In `query_executor.go`, in `execDDL()`, we differentiate two cases:

1. The query is an Online DDL
2. The query is not Online DDL

In the latter case, we always invoke `Reload()` (specifically `ReloadAtEx()`) to reload table schema. 
However, for `CREATE TEMPORARY TABLE` and `DROP TEMPORARY TABLE` this is wasteful:

- Only the connections which created or dropped the table is at all aware of the existence of the table
- `Reload()` adds an overhead which becomes significant if temporary tables are created and dropped intensively.

With this PR we skip `Reload()` for temporary tables.

## Checklist

### DIFF
--- a/go/vt/vttablet/tabletserver/query_executor.go
+++ b/go/vt/vttablet/tabletserver/query_executor.go
@@ -658,9 +658,8 @@ func (qre *QueryExecutor) execDDL(conn *StatefulConnection) (*sqltypes.Result, e
 	}
 
 	isTemporaryTable := false
-	switch stmt := qre.plan.FullStmt.(type) {
-	case sqlparser.DDLStatement:
-		isTemporaryTable = stmt.IsTemporary()
+	if ddlStmt, ok := qre.plan.FullStmt.(sqlparser.DDLStatement); ok {
+		isTemporaryTable = ddlStmt.IsTemporary()
 	}
 	if !isTemporaryTable {
 		// Temporary tables are limited to the session creating them. There is no need to Reload()
--- a/go/vt/vttablet/tabletserver/query_executor.go
+++ b/go/vt/vttablet/tabletserver/query_executor.go
@@ -657,20 +657,31 @@ func (qre *QueryExecutor) execDDL(conn *StatefulConnection) (*sqltypes.Result, e
 		}
 	}
 
-	defer func() {
-		// Call se.Reload() with includeStats=false as obtaining table
-		// size stats involves joining `information_schema.tables`,
-		// which can be very costly on systems with a large number of
-		// tables.
-		//
-		// Instead of synchronously recalculating table size stats
-		// after every DDL, let them be outdated until the periodic
-		// schema reload fixes it.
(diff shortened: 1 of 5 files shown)


go/vt/vttablet/tabletserver/query_executor.go
```diff
@@ -657,20 +657,31 @@ func (qre *QueryExecutor) execDDL(conn *StatefulConnection) (*sqltypes.Result, e
...
-
+	isTemporaryTable := false
+	switch stmt := qre.plan.FullStmt.(type) {
+	case *sqlparser.CreateTable:
+		isTemporaryTable = stmt.IsTemporary()
+	case *sqlparser.DropTable:
+		isTemporaryTable = stmt.IsTemporary()
+	}
+
+	if !isTemporaryTable {
```
harshit-gangal, line 668: In the analyzeDDL plan
we mark in the plan `NeedsReservedConn` to true, so we can use that field here and ignore the `reload`

```go
return &Plan{PlanID: PlanDDL, FullQuery: fullQuery, FullStmt: stmt, NeedsReservedConn: stmt.IsTemporary()}, nil
```

harshit-gangal (commented):
Can we add a test for this change?

shlomi-noach:
@harshit-gangal good idea; I don;t have an immediate understanding of how to test this, I'd need to dig in. If you happen to know, would you add such test?

go/vt/vttablet/tabletserver/query_executor.go
shlomi-noach: Can there be a future scenario where `NeedsReservedConn` is `true` even if the table is not `TEMPORARY`?

shlomi-noach: I've meanwhile simplified the check for `IsTemporary()`

shlomi-noach:
Fixes https://github.com/vitessio/vitess/issues/12146

harshit-gangal:
> @harshit-gangal good idea; I don;t have an immediate understanding of how to test this, I'd need to dig in. If you happen to know, would you add such test?

I will add the test

go/vt/vttablet/tabletserver/query_executor.go
harshit-gangal: I do not expect that to change and I have added a test specifically for that. 
We can keep the current simplified version as well.

shlomi-noach: let's keep the existing version, then, as it is more explicit :pray:

harshit-gangal (approved):

shlomi-noach:
Tests-wise, I see validation that `NeedsReservedConn` is `true` or is `false`; but should there be a test that validates that `Reload` does, in fact, execute or does not execute?

frouioui (approved):

PR #12145, opened 2023-01-25
fix: add WatchRecursive to k8stopo

carsonoid:
This is a rework of the k8stopo to implement of the behaviors expected by `WatchRecursive`. 

The rework here is that all watches, not just `WatchRecursive` are now more performant. Each topo server now only has a single Kubernetes API watch and calls to the `Watch` or `WatchRecursive` methods simply plug a new channel into the ongoing stream of updates. This means that no matter how many watches are set up, the load on the API is only ever one single update stream. Since we were already using an update stream to handle locally indexed data for `List` and `Get` operations there is essentially no overhead added to implement `WatchRecursive`

Signed-off-by: Carson Anderson <ca@carsonoid.net>

## Description

Add support for `topo.WatchRecursive` to the k8stopo.

It does this by watching everything and just returning the updates for resources that match.

## Checklist

### DIFF
--- a/go/vt/topo/k8stopo/server_flaky_test.go
+++ b/go/vt/topo/k8stopo/server_flaky_test.go
@@ -79,7 +79,13 @@ func TestKubernetesTopo(t *testing.T) {
 	if err != nil {
 		t.Fatal("Unable to start k3s", err)
 	}
-	defer killK3s()
+	defer (func() {
+		killK3s()
+		err := c.Wait()
+		if err != nil {
+			t.Log("k3s exited with error", err)
+		}
+	})()
 
 	// Wait for server to be ready
 	for {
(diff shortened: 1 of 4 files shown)


go/vt/topo/k8stopo/server_flaky_test.go
```diff
@@ -79,7 +79,13 @@ func TestKubernetesTopo(t *testing.T) {
 	if err != nil {
 		t.Fatal("Unable to start k3s", err)
 	}
-	defer killK3s()
+	t.Cleanup(func() {
```
carsonoid, line 82: This change makes tests less flaky by ensuring we give the OS time to fully shut down the k3s process. It's not perfect though because test timeouts can still leave a k3s process running in the background. So this is an improvement but still not perfect.

go/vt/topo/k8stopo/watchers.go
```diff
@@ -0,0 +1,146 @@
...
+func (wg *watchGroup) addWatcher(w watcher) func() {
+	// add to the watchers
+	id := uuid.NewString()
+
+	wg.watchers.Store(id, w)
+
+	return func() {
+		wg.watchers.Delete(id)
+	}
+}
```
vmg, line 66: This is not a great use for a `sync.Map`. I think `watchGroup` should be modeled as a `map[watcher]struct{}` with an adjacent `sync.Mutex`. It has the same behavior as the `sync.Map`, but you don't need to use UUIDs or return a `func()` for cleanup (you can `delete(wg, w)` directly), you don't need to use `interface{}` all over the place, etc.

Generally speaking, `sync.Map` is not a very good primitive!

vmg (changes requested):

PR #12147, opened 2023-01-25
schemadiff: validate views' referenced columns

shlomi-noach:
## Description

Up till now, `schemadiff` views validation and analysis was limited to:

- Obviously validate statement
- Infer tables (or views) reverenced by a view
- Validate those tables/views exist
- Validate no dependency loop

As of this PR, `schemadiff` also validates columns used and returned by views:

- Validate that columns in `SELECT` and in `WHERE` clause do in fact exist in tables/views referenced by the view. Error when an unknown column is encountered.
- Allow unqualified table name where there is no ambiguity. Error on unqualified name with ambiguity.
- Analyze view's column names, whether aliased or not.
- Support cascaded views.
- Support star expression, either qualified (`create view v as select t.* from t, t2 ...`) or unqualified (`create view v as select * from t1,t2`) ; infer list of columns ; support cascaded views based on star expression views.

Plenty tests added to validate behavior.

Joint with @dbussink and based on his preliminary work.

## Related Issue(s)

- Tracking issue: https://github.com/vitessio/vitess/issues/10203

## Checklist


shlomi-noach:
In a followup PR: optimize such that we only analyze columns tables/views that are referenced by some view in the first place; there is no need to analyze e.g. a table that is never read by a view. This will allow us to save some memory & cpu.

go/vt/schemadiff/diff_test.go
```diff
@@ -667,6 +667,7 @@ func TestDiffSchemas(t *testing.T) {
 					// validate schema1 unaffected by Apply
 					assert.Equal(t, schema1SQL, schema1.ToSQL())
 
+					require.NotNil(t, schema2)
```
dbussink, line 670: Is this check useful? We already check earlier that there's no error returned when `schema2` is created?

go/vt/schemadiff/schema.go
```diff
@@ -750,3 +759,258 @@ func (s *Schema) Apply(diffs []EntityDiff) (*Schema, error) {
 	}
 	return dup, nil
 }
+
+// TODO
```
dbussink, line 763: What does this `TODO` reference to?

go/vt/schemadiff/schema.go
```diff
@@ -750,3 +759,258 @@ func (s *Schema) Apply(diffs []EntityDiff) (*Schema, error) {
...
+
+func (s *Schema) ValidateViewReferences() error {
+	var allerrors error
```
dbussink, line 766: Usually use `errs` as a the name for this kinda thing to collect errors.

go/vt/schemadiff/schema.go
```diff
@@ -750,3 +759,258 @@ func (s *Schema) Apply(diffs []EntityDiff) (*Schema, error) {
...
+	availableColumns := tablesColumnsMap{}
+
+	for _, e := range s.Entities() {
+		entityColumns, err := s.getEntityColumnNames(e.Name(), availableColumns)
+		if err != nil {
+			return err
```
dbussink, line 772: Should this add to the error list instead? And then we continue through the rest of the entities? If this is a totally unexpected error, maybe good to add a comment about that and why it's safe to return immediately?

go/vt/schemadiff/schema.go
```diff
@@ -750,3 +759,258 @@ func (s *Schema) Apply(diffs []EntityDiff) (*Schema, error) {
...
+		// Found the column in the table, all good.
+		return nil
+	}
+
+	if _, ok := columnErrors[tableName]; !ok {
+		columnErrors[tableName] = make(map[string]struct{})
+	}
+
+	if _, ok := columnErrors[tableName][node.Name.Lowered()]; ok {
+		return nil
```
dbussink, line 935: Similarly, I think is a case of an error already handled earlier so it's safe to do this return?

go/vt/schemadiff/schema.go
shlomi-noach: hehe, it references the fact that is is already done... removing...

go/vt/schemadiff/diff_test.go
shlomi-noach: removed

go/vt/schemadiff/schema.go

go/vt/schemadiff/schema.go

go/vt/schemadiff/schema.go

go/vt/schemadiff/schema.go
shlomi-noach: I'm not sure what you mean here; do you mean it's safe to comment out this return?

dbussink: Ah wait never mind, I read this the wrong way around 🤦

go/vt/schemadiff/schema.go
```diff
@@ -750,3 +759,256 @@ func (s *Schema) Apply(diffs []EntityDiff) (*Schema, error) {
...
 }
+
+func (s *Schema) ValidateViewReferences() error {
+	var errs error
+	availableColumns := tablesColumnsMap{}
+
+	for _, e := range s.Entities() {
+		entityColumns, err := s.getEntityColumnNames(e.Name(), availableColumns)
+		if err != nil {
+			errs = multierr.Append(errs, err)
```
dbussink, line 770: I think it's safer here to `continue` this loop? So we don't setup the available columns since there's none for this table?

go.mod
```diff
@@ -113,6 +113,7 @@ require (
 	github.com/kr/text v0.2.0
 	github.com/nsf/jsondiff v0.0.0-20210926074059-1e845ec5d249
 	github.com/openark/golib v0.0.0-20210531070646-355f37940af8
+	go.uber.org/multierr v1.8.0
```
dbussink, line 116: Let's also update to `1.9.0` here since that's the latest?

shlomi-noach: done

go/vt/schemadiff/schema.go
shlomi-noach: agreed, done!

dbussink (approved):

go/vt/schemadiff/schema.go
```diff
@@ -750,3 +759,257 @@ func (s *Schema) Apply(diffs []EntityDiff) (*Schema, error) {
...
+				err := verifyUnqualifiedColumn(view, availableColumns, tableReferences, node.Name, unqualifiedColumnErrors)
+				errs = multierr.Append(errs, err)
+			} else {
+				err := verifyQualifiedColumn(view, availableColumns, tableAliases, node, qualifiedColumnErrors)
+				errs = multierr.Append(errs, err)
+			}
+		}
+		return true, nil
+	}, view.Select)
+	if err != nil {
```
systay, line 853: FYI - `sqlparser.Walk` will only return the error you return from inside the visitor function, and since you don't return any errors from that, no errors will ever make it out.

shlomi-noach: @systay ah, great! Thank you. I'll keep the check as it is, for safety, but good to know!

dbussink: @systay Right, but the errors are gathered here in `errs`? And those are at the end returned if the walker itself doesn't error (which is not expected).

systay: @dbussink yeah, that was my point. Not really necessary to catch the returned error from `sqlparser.Walk` since that is not how we are dealing with the errors. When I know it will never return an error, I usually write.

```
	_ = sqlparser.Walk(...)
```

OTOH - it's probably good defensive programming to do as @shlomi-noach is doing here and catching and checking the error anyway.

dbussink: @systay Right, guess as someone now knowing the details too much about how `Walk` is implemented, I wouldn't know that it can't return an error and would still write it defensively :smile:.

shlomi-noach:
Looking for a 2nd review/approval

go/vt/schemadiff/schema.go
```diff
@@ -750,3 +759,257 @@ func (s *Schema) Apply(diffs []EntityDiff) (*Schema, error) {
...
 }
+
+func (s *Schema) ValidateViewReferences() error {
+	var errs error
+	availableColumns := tablesColumnsMap{}
+
+	for _, e := range s.Entities() {
+		entityColumns, err := s.getEntityColumnNames(e.Name(), availableColumns)
+		if err != nil {
+			errs = multierr.Append(errs, err)
```
mattlord, line 770: Why not use the Vitess concurrency package here along with `vterrors.Aggregate`? You can see that used throughout the code base if you look for `.AggrError(vterrors.Aggregate)`. Between the `concurrency` and `vterrors` packages we should have whatever related functionality you need here for error handling. For example: https://github.com/vitessio/vitess/blob/390ddad257c7c9ac0ba6cfe15e2a21b0ee943e67/go/vt/wrangler/traffic_switcher.go#L146-L161

go/vt/schemadiff/errors.go
```diff
@@ -308,3 +308,31 @@ type ViewDependencyUnresolvedError struct {
...
+	View      string
+	Table     string
+	Column    string
+	NonUnique bool
+}
+
+func (e *InvalidColumnReferencedInViewError) Error() string {
+	switch {
+	case e.Column == "":
+		return fmt.Sprintf("view %s references non-existing table %s", sqlescape.EscapeID(e.View), sqlescape.EscapeID(e.Table))
```
mattlord, line 322: Nit, but non-existent is more proper in these errors.

go/vt/schemadiff/errors.go
```diff
@@ -308,3 +308,31 @@ type ViewDependencyUnresolvedError struct {
...
+	case e.Table != "":
+		return fmt.Sprintf("view %s references non existing column %s.%s", sqlescape.EscapeID(e.View), sqlescape.EscapeID(e.Table), sqlescape.EscapeID(e.Column))
```
mattlord, line 324: Is it possible that the identifiers are already escaped at this point? If so, we could unescape first before escaping.

go.mod
```diff
@@ -115,6 +115,7 @@ require (
 	github.com/openark/golib v0.0.0-20210531070646-355f37940af8
 	github.com/planetscale/log v0.0.0-20221118170849-fb599bc35c50
 	github.com/slok/noglog v0.2.0
+	go.uber.org/multierr v1.9.0
```
mattlord, line 118: This should not be necessary. See other comment.

mattlord (commented):
I think we can/should avoid importing the external `multierr` package but it otherwise LGTM.

Not adding a formal approval as we should, IMO, get @systay or @harshit-gangal to review as they are in the query serving team and working on the overall Vitess View support work. This way we can ensure that we're moving forward together and all the pieces will fit together nicely. 

@systay or @harshit-gangal , could you please review? Please note if you cannot and I can add a formal approval here if needed. Thanks!

go/vt/schemadiff/schema.go
dbussink: It looks like it does use locking which is unnecessary here. Dunno if we care about the overhead of that here? That logic seems designed for concurrent error gathering which isn't what we're doing here. 

Using `multierr` seems simpler here and it's already an indirect dependency? 

I think this logic here is temporary anyway, since once https://github.com/golang/go/issues/53435 is available with Go 1.20 we probably want to switch to that anyway.

shlomi-noach: cc @dbussink

go/vt/schemadiff/errors.go
shlomi-noach: fixed

go/vt/schemadiff/errors.go
shlomi-noach: It is not possible; those values are retrieved from the AST and are always unqualified.

go/vt/schemadiff/schema.go
shlomi-noach: heh, race condition

shlomi-noach:
Incorporated https://github.com/vitessio/vitess/issues/12189 into this PR, as it deals with view analysis

go/vt/schemadiff/schema.go
shlomi-noach: How strongly do people feel about this? Looking at the two implementation `multierr` does seem to be more fitting to our purpose, but I don't feel strongly.

mattlord: If you only need multiple errors you can use vterrors.Aggregate(), no need for the concurrency piece (as noted, geared toward N goroutines) like we do here e.g.:
https://github.com/vitessio/vitess/blob/3f55e407e8300713d81f7298cdc9677db4be492a/go/vt/vttablet/tabletmanager/vreplication/vcopier.go#L597-L621

I don't feel strongly about it though. Up to you.

shlomi-noach:
I've switched away from `multierr` and into `AllErrorsRecorder`. I want to say `multierr` seems to have the better design over `AllErrorsRecorder`, and even over the new Go 1.20 error wrapping support.

shlomi-noach:
Looking for another review/approval :pray:

shlomi-noach:
ping @vitessio/query-serving for review

shlomi-noach:
ping https://github.com/orgs/vitessio/teams/query-serving for review

GuptaManan100 (approved):
Everything else LGTM!

go.mod
```diff
@@ -187,7 +187,7 @@ require (
 	github.com/tidwall/pretty v1.2.0 // indirect
 	go.opencensus.io v0.24.0 // indirect
 	go.uber.org/atomic v1.10.0 // indirect
-	go.uber.org/multierr v1.9.0 // indirect
```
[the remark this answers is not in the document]
shlomi-noach: It was upgraded in a previous PR, that was eventually not using `multierr`. I just ran `go mod tidy` and it auto-downgraded, so... I don't know?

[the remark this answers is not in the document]
dbussink: @shlomi-noach might be because the direct dependency was dropped? A `go get -u go.uber.org/multierr` probably would add this back correctly (it should be there, so good catch @GuptaManan100).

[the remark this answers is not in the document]
shlomi-noach: This is fixed.

shlomi-noach:
Replaced by https://github.com/vitessio/vitess/pull/12565 ; will close shortly.

PR #12148, opened 2023-01-25
Remove the semgrep action

dbussink:
This job is currently always failing since the syntax is wrong. Let's remove it for now. 

## Checklist

I think we should backport to also fix this noise on those branches as well.


dbussink:
See https://github.com/vitessio/vitess/actions/workflows/semgrep.yml for all the current failures.

mattlord (approved):

deepthi (approved):

frouioui (approved):

PR #12149, opened 2023-01-25
Remove replication manager and run VTOrc in all e2e tests

GuptaManan100:
## Description

This PR removes the replication manager from vttablets. It also adds VTOrc to all the end-to-end tests with some exceptions like reparent tests that don't want VTOrc running which could shadow bugs if they were introduced. 

The flag `disable-replication-manager` is also deprecated. 

Going forward VTOrc is the only component in Vitess that is responsible for fixing replication and if the users want Vitess to manage the replication, then they must run VTOrc.

## Checklist


doc/releasenotes/16_0_0_summary.md
```diff
@@ -39,6 +39,11 @@ In [PR #11097](https://github.com/vitessio/vitess/pull/11097) we introduced nati
 - A server restored to a point in time remains in `DRAINED` tablet type, and does not join the replication stream (thus, "frozen" in time).
 - It is possible to take incremental backups from different tablets. It is OK to have overlaps in incremental backup contents. The restore process chooses a valid path, and is valid as long as there are no gaps in the backed up binary log content.
 
+### Replication manager removal and VTOrc becomes compulsory
```
deepthi, line 42:
```suggestion
### Replication manager removal and VTOrc becomes mandatory
```

doc/releasenotes/16_0_0_summary.md
```diff
@@ -39,6 +39,11 @@ In [PR #11097](https://github.com/vitessio/vitess/pull/11097) we introduced nati
...
+VTOrc is now a compulsory component of Vitess starting from v16. If the users want VTOrc to manage replication, then they must run VTOrc.
```
deepthi, line 43:
```suggestion
VTOrc is now a **required** component of Vitess starting from v16. If the users want VTOrc to manage replication, then they must run VTOrc.
```

deepthi (approved):
👍

frouioui (approved):

PR #12151, opened 2023-01-25
RFC: flow based tablet load balancer

demmer:
## Description
RFC implementation for a new balancing option to the tabletgateway's algorithm for routing queries to a set of available replicas.

_Note: This code has not yet been tested in production to verify it actually works, so we are currently just looking for feedback on the approach and whether there are any concerns about merging this._

## Related Issue(s)
The motivation described in  #12241

## Checklist

## Deployment Notes

None

### DIFF
--- a/go/vt/vtgate/balancer/balancer.go
+++ b/go/vt/vtgate/balancer/balancer.go
@@ -1,0 +1,348 @@
+/*
+Copyright 2023 The Vitess Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


deepthi:
@demmer we are requiring issues for all enhancements / bug fixes. Can you create an issue (preferably as an RFC)? Then this PR can just say `Fixes #nnnn`.

demmer:
@deepthi Done. See https://github.com/vitessio/vitess/issues/12241

Issue #12168, opened 2023-01-26
OnlineDDL: _vt.schema_migrations `AUTO_INCREMENT` can have substantial gaps


shlomi-noach:
We've noticed that `_vt.schema_migrations`' `id` column, an `AUTO_INCREMENT` can have quite significant gaps. The technical reason for these gaps is that `vitess` uses `INSERT IGNORE INTO _vt.schema_migrations`, as it allows idempotent submissions of migrations. So, if a migration is submitted with an already existing `UUID`, it is _mostly ignored_.

The side effect is that the table's `AUTO_INCREMENT` value increases. Normally, no big deal, but in production we have some retry mechanisms that, given specific circumstance, can retry submitting a migration routinely, hammering the `AUTO_INCREMENT` value; we've the number jump to some `x00,000` with just a few migrations. That's too much.

We'd like to reduce or eliminate this `AUTO_INCREMENT` impact.

PR #12169, opened 2023-01-26
OnlineDDL: avoid schema_migrations AUTO_INCREMENT gaps by pre-checking for existing migration

shlomi-noach:
## Description

Fixes https://github.com/vitessio/vitess/issues/12168
This PR prevents excessive gaps in `_vt.schema_migrations` `id` values with a small refactor:

- Instead of relying on `INSERT IGNORE`, we first `SELECT` for existing migration UUID, and only if it does not exist, we `INSERT`.
- Because this is now a two-step logic, we add a mutex around the function (the mutex is only used by `SubmitMigration` function and does not otherwise conflict with the rest of the scheduler's work)
- The logic that used to _follow_ the `INSERT IGNORE`, which addresses the case where the UUID pre-existed, is now moved to run _before_ `INSERT IGNORE`

Tests are unchanged. We expect no change in behavior.

## Related Issue(s)

https://github.com/vitessio/vitess/issues/12168

## Checklist

### DIFF
--- a/go/test/endtoend/onlineddl/scheduler/onlineddl_scheduler_test.go
+++ b/go/test/endtoend/onlineddl/scheduler/onlineddl_scheduler_test.go
@@ -237,6 +237,9 @@ func TestSchemaChange(t *testing.T) {
 	t.Run("singleton", testSingleton)
 	t.Run("declarative", testDeclarative)
 	t.Run("foreign-keys", testForeignKeys)
+	t.Run("summary: validate sequential migration IDs", func(t *testing.T) {
+		onlineddl.ValidateSequentialMigrationIDs(t, &vtParams, shards)
+	})
 }
 
 func testScheduler(t *testing.T) {
@@ -1958,10 +1961,6 @@ func testForeignKeys(t *testing.T) {
 			})
 		})
 	}
-}
-
-func TestSequentialMigrationIds(t *testing.T) {
-	onlineddl.ValidateSequentialMigrationIDs(t, &vtParams, shards)
 }
 
 // testOnlineDDLStatement runs an online DDL, ALTER statement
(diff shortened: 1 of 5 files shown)


shlomi-noach:
converted to draft, doing another necessary refactor (need to take singleron/singleton-context scenario into consideration)

shlomi-noach:
OK, did another refactor; extracted an inner function, `validateSingleton()` and some logic around it, onto a new function, `func (e *Executor) submitCallbackIfNonConflicting()`. And this logic applies to either submitting a migration or retrying a migration. It validates whether it is OK for the given migration to retry/submit given `singleton` or `singleton-context` constraints.

mattlord (commented):
LGTM!

Not breaking any test would indicate that we don't have test coverage around submissions. 🙂 I think we should add more unit tests eventually (that could also e.g. confirm that we don't have gaps anymore when we resubmit) but that's a bigger issue and not specific to this work. 

In the meantime, can we easily add anything to the endtoend tests to confirm the behavioral change? Perhaps a select at the end of a workflow to look for gaps in the migration IDs?

shlomi-noach:
> Not breaking any test would indicate that we probably don't have test coverage around submissions.

The opposite is true. We have plenty tests around submissions -- every single test, really, and in particular all `singleton` and `singleton-context` tests -- literally hundreds.

shlomi-noach:
> Perhaps a select at the end of a workflow to look for gaps in the migration IDs?

I'll look into that.

mattlord:
> > Not breaking any test would indicate that we probably don't have test coverage around submissions.
> 
> The opposite is true. We have plenty tests around submissions -- every single test, really, and in particular all `singleton` and `singleton-context` tests -- literally hundreds.

Yea, I meant code test coverage — i.e. unit tests. 🙂 Was just noting it (not a blocker).

shlomi-noach:
Added an explicit test to validate `id` values are sequential and without gaps; this test is now called at the end of multiple `endtoend` workflows.

dbussink (approved):

deepthi:
request for review from @vitessio/query-serving

frouioui (approved):

PR #12171, opened 2023-01-26
Use schema for the information_schema views

systay:
## Description
Improve `information_schema` planning by using more of the known schema information when planning these queries.

Earlier we would accept any view/table as long as it was in the `information_schema` keyspace. Now we use the schema information that is built into Vitess to assert that the table/view is known before accepting the query.

It also contains a bit of refactoring to make the code easier to understand.


go/vt/vtgate/planbuilder/operators/system_tables.go
```diff
@@ -37,60 +37,66 @@ func (r *Route) findSysInfoRoutingPredicatesGen4(predicates []sqlparser.Expr, re
...
-			if r.SysTableTableName == nil {
-				r.SysTableTableName = map[string]evalengine.Expr{}
-			}
 			r.SysTableTableName[bvName] = out
 		}
 	}
 	return nil
 }
 
 func extractInfoSchemaRoutingPredicate(in sqlparser.Expr, reservedVars *sqlparser.ReservedVars) (bool, string, evalengine.Expr, error) {
```
frouioui, line 53: Perhaps adding a comment here to explain what the return parameters are would be 👌🏻

go/vt/vtgate/planbuilder/operators/system_tables.go
```diff
@@ -37,60 +37,66 @@ func (r *Route) findSysInfoRoutingPredicatesGen4(predicates []sqlparser.Expr, re
...
-					name = sqltypes.BvSchemaName
-				} else {
-					name = reservedVars.ReserveColName(col.(*sqlparser.ColName))
-				}
-				replaceOther(sqlparser.NewArgument(name))
-				return isSchemaName, name, evalExpr, nil
-			}
+	cmp, ok := in.(*sqlparser.ComparisonExpr)
+	if !ok || cmp.Operator != sqlparser.EqualOp {
+		return false, "", nil, nil
```
frouioui, line 56: That's a nice cleanup

frouioui (approved):

harshit-gangal:
I would suggest not making it restrictive for information_schema tables and columns. MySQL keeps adding newer tables and columns and would be difficult to maintain or can create blockers for installations with newer MySQL versions.

We should fail only when table and column binding is absolutely needed for planner to move forward.

systay:
> We should fail only when table and column binding is absolutely needed for planner to move forward.

Done. We actually did this in a separate PR recently, I just merged main here and added a test that shows that missing `information_schema` tables are not a problem.

go/vt/vtgate/executor_select_test.go
```diff
@@ -459,13 +459,14 @@ func TestGen4SelectDBA(t *testing.T) {
...
-	query := "select * from INFORMATION_SCHEMA.foo"
+	query := "select * from INFORMATION_SCHEMA.TABLE_CONSTRAINTS"
 	_, err := executor.Execute(context.Background(), "TestSelectDBA",
 		NewSafeSession(&vtgatepb.Session{TargetString: "TestExecutor"}),
 		query, map[string]*querypb.BindVariable{},
 	)
 	require.NoError(t, err)
-	wantQueries := []*querypb.BoundQuery{{Sql: query, BindVariables: map[string]*querypb.BindVariable{}}}
+	expected := "select CONSTRAINT_CATALOG, CONSTRAINT_SCHEMA, CONSTRAINT_NAME, TABLE_SCHEMA, TABLE_NAME, CONSTRAINT_TYPE, `ENFORCED` from INFORMATION_SCHEMA.TABLE_CONSTRAINTS"
+	wantQueries := []*querypb.BoundQuery{{Sql: expected, BindVariables: map[string]*querypb.BindVariable{}}}
```
harshit-gangal, line 469: nit: this change is not needed after merging from `main`

harshit-gangal (commented):
add a test where the view exists in information_schema but the column does not exist. 
I think this might fail here, Should the planner allow this?

go/vt/vtgate/executor_select_test.go
systay: I like the new test better, if you don't mind 🤷

harshit-gangal (approved):

PR #12172, opened 2023-01-26
[Gen4] Fix lookup vindexes with `autocommit` enabled

arthurschreiber:
## Description

The `VindexLookup` primitive re-implemented some parts of the lookup vindex functionality, but missed to honor the `autocommit` flag. This pull request changes `VindexLookup` to re-use existing logic instead of using a separate implementation.

## Related Issue(s)

- Fixes https://github.com/vitessio/vitess/issues/12173

## Checklist

### DIFF
--- a/go/vt/vtgate/vindexes/lookup.go
+++ b/go/vt/vtgate/vindexes/lookup.go
@@ -58,6 +58,10 @@ func (ln *LookupNonUnique) AllowBatch() bool {
 	return ln.lkp.BatchLookup
 }
 
+func (ln *LookupNonUnique) AutoCommitEnabled() bool {
+	return ln.lkp.Autocommit
+}
+
 // String returns the name of the vindex.
 func (ln *LookupNonUnique) String() string {
 	return ln.name
@@ -231,6 +235,10 @@ func (lu *LookupUnique) GetCommitOrder() vtgatepb.CommitOrder {
 
 func (lu *LookupUnique) AllowBatch() bool {
 	return lu.lkp.BatchLookup
+}
+
+func (lu *LookupUnique) AutoCommitEnabled() bool {
+	return lu.lkp.Autocommit
 }
 
 // NewLookupUnique creates a LookupUnique vindex.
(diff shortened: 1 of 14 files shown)


go/vt/vtgate/vindexes/vindex.go
```diff
@@ -140,9 +140,8 @@ type (
 	LookupPlanable interface {
 		String() string
 		Query() (selQuery string, arguments []string)
+		Map(ctx context.Context, vcursor VCursor, ids []sqltypes.Value) ([]key.Destination, error)
 		MapResult(ids []sqltypes.Value, results []*sqltypes.Result) ([]key.Destination, error)
-		AllowBatch() bool
-		GetCommitOrder() vtgatepb.CommitOrder
 	}
```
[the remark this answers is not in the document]
harshit-gangal: Ignore this comment. We have to move autocommit logic as well from Map function to vindex_lookup engine primitive.

arthurschreiber:
@harshit-gangal I took a different approach and simply re-implement the autocommit logic inside the `VindexLookup` primitive.

I'm not 100% happy with the code duplication that's going on. It would be nice to refactor this further so that both the old API for lookup vindexes and the new API can share the same code / logic so that any future issues only need to be fixed in one place.

arthurschreiber:
@harshit-gangal I think this is ready for another review. With the latest commit I pushed, all test cases should also be passing now. 🤞

go/vt/vtgate/vindexes/lookup.go
```diff
@@ -104,6 +108,14 @@ func (ln *LookupNonUnique) Map(ctx context.Context, vcursor VCursor, ids []sqlty
...
 }
 
+func (ln *LookupNonUnique) Lookup(ctx context.Context, vcursor VCursor, ids []sqltypes.Value) ([]*sqltypes.Result, error) {
+	results, err := ln.lkp.Lookup(ctx, vcursor, ids, vtgatepb.CommitOrder_NORMAL)
+	if err != nil {
+		return nil, err
+	}
+	return results, err
+}
+
```
harshit-gangal, line 118: I think this is not used and hence can be removed.

harshit-gangal (approved):
One comment rest looks good.

go/vt/vtgate/vindexes/lookup.go
arthurschreiber: Ah, right, I must have committed this by accident! Thanks for catching this. 🙇‍♂️

arthurschreiber: Done! This change was removed.

systay (approved):

PR #12174, opened 2023-01-26
Add flag to select tx throttler tablet type

timvaillancourt:
## Description

This PR allows the user to specify which tablet type(s) should be monitored by the [`vttablet` Transaction Throttler](https://github.com/vitessio/vitess/tree/main/go/vt/vttablet/tabletserver/txthrottler) _([Design Doc](https://github.com/vitessio/vitess/blob/main/doc/design-docs/ReplicationLagBasedThrottlingOfTransactions.md))_. The motivation for this is to allow `rdonly` tablets to be included in throttling

This involves the addition of the `--tx-throttler-tablet-types` flag _(a comma-separated list of tablet types)_ with the default of `replica` for backwards-compatability. Only `replica` and/or `rdonly` are supported

This resolves [this code comment in `tx_throttler.go`](https://github.com/vitessio/vitess/blob/main/go/vt/vttablet/tabletserver/txthrottler/tx_throttler.go#L358-L359): 
> // TODO(erez): If this becomes necessary, we can add a configuration option that would
> // determine whether we consider RDONLY tablets here, as well.

`vttablet --help` output:
```bash
tim@Tims-MacBook-Pro vitess % ./vttablet --help 2>&1 | grep tx-throttler-tablet-types
      --tx-throttler-tablet-types strings                                A comma-separated list of tablet types. Only tablets of this type are monitored for replication lag by the transaction throttler. Supported types are replica and/or rdonly. (default replica)
```

~~Backports to v14 + v15 would be much appreciated if they merge cleanly~~

## Checklist


deepthi:
Looks like our bot isn't running right now, otherwise there would have been a comment like this https://github.com/vitessio/vitess/pull/12138#issuecomment-1402283638

Relevant points:
- New flags should use dashes, not underscores
- New flags need to be documented on the website

Re backports, we do them only for regression / production blocking bugs. This doesn't seem like either. If that isn't right and you can provide a justification for the back port, I'm happy to discuss.

timvaillancourt:
@deepthi thanks for the feedback, updating the PR based on this 👍 

> * New flags should use dashes, not underscores

The majority of the existing tx throttler flags are using underscores, would it make sense for me to update those also?

> Re backports, we do them only for regressions / production blocking bugs. This doesn't seem like either. 

Good point. Correct, this does not meet the criteria for backporting 👍

timvaillancourt:
> The majority of the existing tx throttler flags are using underscores, would it make sense for me to update those also or is an inconsistency acceptable?

For now I have updated the new flag only to dashes. It appears the existing flags have `Synonym to -tx_throttler_healthcheck_cells` aliases with dashes 👍

timvaillancourt:
Website docs PR: https://github.com/vitessio/website/pull/1368

deepthi (commented):
This looks pretty good to me. The only thing I'd ask is to add a unit test that actually tests with multiple tablet types.
e2e tests are expensive so I'm not going to ask for one. Can you confirm that you are able to run this locally and get the expected behavior?
Website PR is looking good.

deepthi:
> > The majority of the existing tx throttler flags are using underscores, would it make sense for me to update those also or is an inconsistency acceptable?
> 
> For now I have updated the new flag only to dashes. It appears the existing flags have dash-based aliases with help like: `Synonym to -tx_throttler_healthcheck_cells` 👍

You did the right thing. We eventually want to move all flags to be only dashes, but we have a long way to go until we get there. It will take a couple more releases (at least).

harshit-gangal (approved):

go/vt/vttablet/tabletserver/tabletenv/config.go
```diff
@@ -138,8 +140,9 @@ func registerTabletEnvFlags(fs *pflag.FlagSet) {
...
-	flagutil.DualFormatStringVar(fs, &currentConfig.TxThrottlerConfig, "tx_throttler_config", defaultConfig.TxThrottlerConfig, "The configuration of the transaction throttler as a text formatted throttlerdata.Configuration protocol buffer message")
+	flagutil.DualFormatStringVar(fs, &currentConfig.TxThrottlerConfig, "tx_throttler_config", defaultConfig.TxThrottlerConfig, "The configuration of the transaction throttler as a text formatted throttlerdata.Configuration protocol buffer message.")
 	flagutil.DualFormatStringListVar(fs, &currentConfig.TxThrottlerHealthCheckCells, "tx_throttler_healthcheck_cells", defaultConfig.TxThrottlerHealthCheckCells, "A comma-separated list of cells. Only tabletservers running in these cells will be monitored for replication lag by the transaction throttler.")
+	fs.Var((*topoproto.TabletTypeListFlag)(&currentConfig.TxThrottlerTabletTypes), "tx-throttler-tablet-types", "A comma-separated list of tablet types. Only tablets of this type are monitored for replication lag by the transaction throttler. Supported types are replica and/or rdonly.")
```
shlomi-noach, line 145: My opinion: we should distance ourselves from proto-format flags. I understand that these are strongly typed, but on the other hand this isn't standard practice in `golang` world or outside `vitess`, and mostly these are difficult to craft by hand, what with all the escaping.

Personally I'd go with just a comma delimited `string` like here:

- https://github.com/vitessio/vitess/blob/11fd98da90c41bbfe522ec6761d731666978ab78/go/vt/vttablet/tabletserver/gc/tablegc.go#L67
- then split it like here: https://github.com/vitessio/vitess/blob/11fd98da90c41bbfe522ec6761d731666978ab78/go/vt/vttablet/onlineddl/executor.go#L326

go/vt/vttablet/tabletserver/txthrottler/tx_throttler.go
```diff
@@ -119,6 +119,15 @@ func tryCreateTxThrottler(env tabletenv.Env, topoServer *topo.Server) (*TxThrott
 		return newTxThrottler(env, &txThrottlerConfig{enabled: false})
 	}
 
+	for _, tabletType := range env.Config().TxThrottlerTabletTypes {
+		switch tabletType {
+		case topodatapb.TabletType_REPLICA, topodatapb.TabletType_RDONLY:
+			continue
+		default:
+			return nil, fmt.Errorf("%q tablet type is not supported", tabletType)
```
shlomi-noach, line 127: Please use `vterrors.Errorf( .... )` and supply an appropriate error code.
vterrors are more formal, are used to wrap and unwrap other errors, and are easier to detect in logs.

go/vt/vttablet/tabletserver/tabletenv/config.go
shlomi-noach: Not a hill to die on for now, as the current config is also already in proto format. Feel free to keep this as is. Possibly, in the future we may change this.

ajm188: Also going to tangent here: because we're using pflag, we can use `fs.StringSliceVar` and have pflag do the CSV splitting for us (and therefore move away from this "string var then we split it for you" code that's all over vitess)

ajm188: Docs: https://pkg.go.dev/github.com/spf13/pflag#FlagSet.StringSliceVar

shlomi-noach: oh yes, definitely `StringSliceVar` over `string`. And I should do the same in existing codebase!

ajm188 (commented):
couple comments, but all minor!

go/vt/vttablet/tabletserver/tabletenv/config.go
```diff
@@ -499,6 +505,27 @@ func (c *TabletConfig) verifyTransactionLimitConfig() error {
...
+		}
+
+		switch tabletType {
+		case topodatapb.TabletType_REPLICA, topodatapb.TabletType_RDONLY:
+			currentConfig.TxThrottlerTabletTypes = append(
+				currentConfig.TxThrottlerTabletTypes,
+				tabletType,
+			)
+		default:
+			return vterrors.Errorf(vtrpc.Code_INTERNAL, "unsupported tablet type %q", tabletType)
```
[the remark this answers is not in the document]
timvaillancourt: This seems more appropriate 👍

go/vt/vttablet/tabletserver/tabletenv/config.go
```diff
@@ -142,7 +145,7 @@ func registerTabletEnvFlags(fs *pflag.FlagSet) {
...
 	flagutil.DualFormatStringVar(fs, &currentConfig.TxThrottlerConfig, "tx_throttler_config", defaultConfig.TxThrottlerConfig, "The configuration of the transaction throttler as a text formatted throttlerdata.Configuration protocol buffer message.")
 	flagutil.DualFormatStringListVar(fs, &currentConfig.TxThrottlerHealthCheckCells, "tx_throttler_healthcheck_cells", defaultConfig.TxThrottlerHealthCheckCells, "A comma-separated list of cells. Only tabletservers running in these cells will be monitored for replication lag by the transaction throttler.")
-	fs.Var((*topoproto.TabletTypeListFlag)(&currentConfig.TxThrottlerTabletTypes), "tx-throttler-tablet-types", "A comma-separated list of tablet types. Only tablets of this type are monitored for replication lag by the transaction throttler. Supported types are replica and/or rdonly.")
+	fs.StringSliceVar(&txThrottlerTabletTypes, "tx-throttler-tablet-types", []string{"replica"}, "A comma-separated list of tablet types. Only tablets of this type are monitored for replication lag by the transaction throttler. Supported types are replica and/or rdonly.")
```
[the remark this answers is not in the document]
timvaillancourt: > the default value should just be defaultConfig.TxThrottlerTabletTypes

@ajm188 that works, but it would mean the struct-field `TxThrottlerTabletTypes` needs to become `[]string`

The drawback is we'd probably need to parse strings to tablet-types twice, once in `.verifyTxThrottlerConfig()` to validate the config early, then again when we compare `topodata.TabletType` in `.StatsUpdate()` (in `*txThrottlerState`). Do you see a smoother way than that?

shlomi-noach:
@timvaillancourt did you want to pursue this PR? I will reopen it if so.

timvaillancourt:
@shlomi-noach thanks, yes! I plan to address suggestions on this PR in the next week or so 👍

timvaillancourt:
@ajm188 to address your suggestion earlier, I've moved the new flag to use `*topoproto.TabletTypeListFlag` which takes care of the conversions we were discussing 🎉 

Assuming this passes CI, I think this is now ready again for review 🙇

go/vt/vttablet/tabletserver/tabletenv/config.go
```diff
@@ -32,8 +32,13 @@ import (
 	"vitess.io/vitess/go/vt/dbconfigs"
 	"vitess.io/vitess/go/vt/log"
 	querypb "vitess.io/vitess/go/vt/proto/query"
+	topodatapb "vitess.io/vitess/go/vt/proto/topodata"
+	"vitess.io/vitess/go/vt/proto/vtrpc"
+	vtrpcpb "vitess.io/vitess/go/vt/proto/vtrpc"
```
ajm188, line 37: this is duplicated

go/vt/vttablet/tabletserver/tabletenv/config.go
```diff
@@ -151,8 +156,9 @@ func registerTabletEnvFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&currentConfig.TwoPCCoordinatorAddress, "twopc_coordinator_address", defaultConfig.TwoPCCoordinatorAddress, "address of the (VTGate) process(es) that will be used to notify of abandoned transactions.")
 	SecondsVar(fs, &currentConfig.TwoPCAbandonAge, "twopc_abandon_age", defaultConfig.TwoPCAbandonAge, "time in seconds. Any unresolved transaction older than this time will be sent to the coordinator to be resolved.")
 	flagutil.DualFormatBoolVar(fs, &currentConfig.EnableTxThrottler, "enable_tx_throttler", defaultConfig.EnableTxThrottler, "If true replication-lag-based throttling on transactions will be enabled.")
-	flagutil.DualFormatStringVar(fs, &currentConfig.TxThrottlerConfig, "tx_throttler_config", defaultConfig.TxThrottlerConfig, "The configuration of the transaction throttler as a text formatted throttlerdata.Configuration protocol buffer message")
+	flagutil.DualFormatStringVar(fs, &currentConfig.TxThrottlerConfig, "tx_throttler_config", defaultConfig.TxThrottlerConfig, "The configuration of the transaction throttler as a text formatted throttlerdata.Configuration protocol buffer message.")
```
ajm188, line 159:
```suggestion
	flagutil.DualFormatStringVar(fs, &currentConfig.TxThrottlerConfig, "tx_throttler_config", defaultConfig.TxThrottlerConfig, "The configuration of the transaction throttler as a text-formatted throttlerdata.Configuration protocol buffer message.")
```

if we're gonna make changes while we're here, this should really be hyphenated

ajm188 (commented):
one minor change and a nitpick, then lgtm

timvaillancourt:
@ajm188 good catches thanks, added those suggestions 👍

ajm188 (approved):

harshit-gangal:
I have a question here.

The purpose of the `tx throttler` is to avoid creating new transactions as the existing replicas are lagging and the new transactions would not be able to commit as they would not receive semi-sync acks.

What purpose we would solve to throttler on `rdonly` tablet types. `primary` tablets do not expect semi-sync acks from those tablets. We would be throttling user traffic when they can actually able to write without causing delays to normal working.

timvaillancourt:
> What purpose we would solve to throttler on `rdonly` tablet types. `primary` tablets do not expect semi-sync acks from those tablets. We would be throttling user traffic when they can actually able to write without causing delays to normal working.

@harshit-gangal while I don't recommend it for most, in Slack's current use-case we read from `rdonly` tablets in some situations and "care" about the staleness of those reads from an application consistency perspective (vs durability)

harshit-gangal:
> > What purpose we would solve to throttler on `rdonly` tablet types. `primary` tablets do not expect semi-sync acks from those tablets. We would be throttling user traffic when they can actually able to write without causing delays to normal working.
> 
> @harshit-gangal while I don't recommend it for most, in Slack's current use-case we read from `rdonly` tablets in some situations and "care" about the staleness of those reads from an application consistency perspective (vs durability)

have you explored the flag `discovery_high_replication_lag_minimum_serving`? The purpose is to remove unhealthy tablets from serving traffic if they exceed the lag tolerance irrespective of not meeting the minimum tablets serving criteria.

timvaillancourt:
> have you explored the flag `discovery_high_replication_lag_minimum_serving`? The purpose is to remove unhealthy tablets from serving traffic if they exceed the lag tolerance irrespective of not meeting the minimum tablets serving criteria.

@harshit-gangal yes, we're using this to remove unhealthy replicas in prod, however losing availability to replicas causes other undesireable problems

The goal we have here is to reduce the rate of writes (tell low-priority clients to backoff) when we are exceeding the rate tablets can replicate to avoid both lag and loss of availability of these tablet types

cc @ejortegau / @shlomi-noach for more thoughts

go/vt/vttablet/tabletserver/txthrottler/tx_throttler.go
```diff
@@ -369,14 +376,16 @@ func (ts *txThrottlerState) deallocateResources() {
...
-	ts.throttler.RecordReplicationLag(time.Now(), tabletStats)
+
+	// Monitor tablets for replication lag if they have a tablet
+	// type specified by the --tx_throttler_tablet_types flag.
+	for _, expectedTabletType := range *ts.config.tabletTypes {
+		if tabletStats.Target.TabletType == expectedTabletType {
+			ts.throttler.RecordReplicationLag(time.Now(), tabletStats)
+			return
+		}
+	}
```
shlomi-noach, line 390: Personal opinion, I'd have a `map` where the _key_ is the tablet type, value is `bool`, and then ask `if ts.config.myTabletTypes[tabletStats.Target.TabletType] {...}`

shlomi-noach (approved):

shlomi-noach (approved):

shlomi-noach:
The discussion took many turns by now, but I wish to comment about the use of `RDONLY` and `semi-sync` etc. I think this flag is in line with existing tablet throttler flag `--throttle_tablet_types`.
It's also worth pointing out that semi-sync has no strict dependency on lag. It is possible for all semi-sync replicas to `ACK` all binlog events, and _also_ have minutes of lag. So on that respect, too, there is no problem imO with including `RDONLY` tablets.

harshit-gangal (approved):

PR #12175, opened 2023-01-27
handle system databases other that information_schema correctly

systay:
## Description
In recent changes, we have added schema information about the `information_schema` database to the vschema. This means that we started asking the vschema about schema information even about system tables.

Since `sys`, `performance_schema` and `mysql` were not added to the vschema, vtgate started to fail planning queries against these databases complaining about the database not existing.

```sql
> SELECT variable, value, set_time, set_by FROM sys.sys_config;

ERROR 1105 (HY000): Unknown database 'sys' in vschema
```

## Related Issue(s)
The bug was introduced in https://github.com/vitessio/vitess/pull/11941
Fixes: https://github.com/vitessio/vitess/issues/12176

## Checklist


harshit-gangal (approved):

dbussink (approved):

frouioui (approved):
LGTM

Issue #12011, opened 2022-12-22
Bug Report: vtgate healthcheck cache grows unbounded


olyazavr:
### Overview of the Issue

We track the number of tablets each vtgate is aware of using the /debug/gateway metrics, and in Vitess 5, that number was stable, around 2-3k. However, with the same number of tablets, Vitess 14 vtgates never seem to forget about old de-registered tablets and that number grows boundlessly:
<img width="1413" alt="image" src="https://user-images.githubusercontent.com/5015430/209222856-6ef322a5-73c4-4cb3-aa8d-c8e11c3db8fa.png">
top is Vitess 14, bottom Vitess 5

We see this in debug/status/ as well:
<img width="1077" alt="image" src="https://user-images.githubusercontent.com/5015430/209222989-1a39bdc7-7b50-4ad2-9288-abedc37125e2.png">

These are all old backup pods that have since spun down and de-registered themselves from the topo. They do NOT appear in `vtctl ListShardTablets`

### Reproduction Steps

Create a new tablet with unique uid, then spin it down and de-reg it, see that the vtgate still has this in its healthcheck cache.
before:
<img width="1474" alt="image" src="https://user-images.githubusercontent.com/5015430/209223426-dec7cb06-ef12-4870-8588-3b98df9d1e14.png">
 
after new tablet spin up and then de-reg:
<img width="1467" alt="image" src="https://user-images.githubusercontent.com/5015430/209224392-aec6aed1-da3f-4d48-a0e8-8e1eaaaa85ca.png">

### Binary Version

```sh
Vitess 14: commit 01ec41fee8e0f0feca4d8b4857f11076adf6be4e
```

### Operating System and Environment details

```sh
CentOS8, Linux 5.4.141-hs22.el8.x86_64
```

mattlord:
Ongoing discussion of the issue is happening in the Vitess slack here: https://vitess.slack.com/archives/C0PQY0PTK/p1671735878009459

I have not been able to repeat the issue yet, and this involves a fork with some related patches in it (to start a tablet as the DRAINED type) so we're still investigating.

olyazavr:
The full path our backup tablets take is: (1) spin up as DRAINED, (2) then RESTORE, (3) then BACKUP, (4) vtctl DeleteTablet

Here are two logs for the vtgate responding to this flow: 
1. https://gist.github.com/olyazavr/f650b1449869164faccf19a7e9b0bd60
[...]

PR #12178, opened 2023-01-27
VTGate: Ensure HealthCheck Cache Secondary Maps Stay in Sync With Authoritative Map on Tablet Delete

mattlord:
## Description

The VTGate HealthCheck cache has an [authoritative map of tablets](https://github.com/vitessio/vitess/blob/release-15.0/go/vt/discovery/healthcheck.go#L267-L268) — keyed by tabletAlias — that it uses for query serving. It has a [secondary map](https://github.com/vitessio/vitess/blob/release-15.0/go/vt/discovery/healthcheck.go#L269-L272) — keyed by `<keyspace>.<shard>.<tabletType>` which has a sorted list of tablets that is used when printing the contents of the cache for the vtgate `/debug/status` endpoint and for the `show vitess_tablets` SQL statement without blocking query serving (lock free in the hot path).

The healthcheck cache implementation in the `discovery` package is responsible for keeping those two maps in sync as tablets are added and removed. There was one case, however, where it was failing to do so. That was in deleteTablet: https://github.com/vitessio/vitess/blob/release-15.0/go/vt/discovery/healthcheck.go#L432-L455

1. We incorrectly assumed that if the tablet was NOT in the authoritative map it was not in the secondary maps and did not attempt to delete it from those in order to ensure that they stay in sync.
2. The secondary maps are keyed in part by the tabletType, which can change in-between cache updates, so when deleting a tablet from the cache (by alias) we should be sure that the tablet is deleted from the secondary maps in any form it could take. 

This PR addresses both of these points.

> **Note** [several people have reported this recently, all of which I believe were using k8s](https://vitess.slack.com/archives/C0PQY0PTK/p1671735878009459). So I suspect that part of what triggered it was tablets being restarted/rescheduled and [coming up with a different `host:port` value which triggers a ReplaceTablet codepath](https://github.com/vitessio/vitess/blob/release-15.0/go/vt/discovery/topology_watcher.go#L219-L230) which does a Delete and Add. This would also have to coincide with the tablet type changing concurrently as well — e.g. from `REPLICA` to `PRIMARY` or `DRAINED` to `BACKUP`.

The new unit test case added in `TestRemoveTablet` demonstrates the issue as the test fails currently on `main` here:
```
--- FAIL: TestRemoveTablet (0.00s)
    /Users/matt/git/vitess/go/vt/discovery/healthcheck_test.go:743: 
        	Error Trace:	/Users/matt/git/vitess/go/vt/discovery/healthcheck_test.go:743
        	Error:      	Should be empty, but was [0x140000b07d0]
        	Test:       	TestRemoveTablet
        	Messages:   	wrong result, expected empty list
```

Which is this point in the test:
```
	// Now replace it, which does a Remove and Add. The tablet should
	// be removed from the cache and all its maps even though the
	// tablet type had changed in-between the initial Add and Remove.
	hc.ReplaceTablet(tablet, tablet)
	// Confirm that the old entry is gone.
	a = hc.GetTabletStats(&querypb.Target{Keyspace: "k", Shard: "s", TabletType: topodatapb.TabletType_REPLICA})
	assert.Empty(t, a, "wrong result, expected empty list")
```

## Related Issue(s)

  - Related to: https://github.com/vitessio/vitess/pull/9237
  - Fixes: https://github.com/vitessio/vitess/issues/12011

## Checklist

-   [x] Tests were added

### DIFF
--- a/go/vt/discovery/healthcheck.go
+++ b/go/vt/discovery/healthcheck.go
@@ -429,8 +429,24 @@ func (hc *HealthCheckImpl) deleteTablet(tablet *topodata.Tablet) {
 	hc.mu.Lock()
 	defer hc.mu.Unlock()
 
-	key := KeyFromTablet(tablet)
 	tabletAlias := tabletAliasString(topoproto.TabletAliasString(tablet.Alias))
+	defer func() {
+		// We want to be sure it's gone from the secondary maps
+		// even if it's already gone from the authoritative map.
+		key := KeyFromTablet(tablet)
+		// delete from map by keyspace.shard.tabletType
+		ths, ok := hc.healthData[key]
+		if !ok {
+			log.Warningf("We have no health data for target: %v", key)
+			return
+		}
+		delete(ths, tabletAlias)
+		// delete from healthy list
+		healthy, ok := hc.healthy[key]
+		if ok && len(healthy) > 0 {
+			hc.recomputeHealthy(key)
+		}
+	}()
 	// delete from authoritative map
 	th, ok := hc.healthByAlias[tabletAlias]
 	if !ok {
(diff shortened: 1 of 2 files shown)


deepthi (approved):
Excellent work! I get why we need to iterate through tablet types, that will be significantly cheaper than iterating through all the map values when it actually matters, i.e. in clusters where there are large numbers of tablets being monitored by each vtgate.

frouioui (approved):
This is really good, thank you @mattlord.

mattlord:
This should be the last missing piece related to: https://github.com/vitessio/vitess/issues/8465

PR #12187, opened 2023-01-30
Purge logs without panicking

ajm188:
## Description

After the move to `pflag` the `logutil.PurgeLogs` that happened during servenv's `OnInit` would panic, because `Init` is called before `Parse`, and the `log_dir` flag doesn't get defined when logutil [starts checking for it](https://github.com/vitessio/vitess/blob/f33b86219ddc52e17a343fdf842728c0b2eae7cc/go/vt/logutil/purge.go#L117-L133)

This reorders the calls to prevent this panic.

Basically, we need this for anyone that imports `servenv` _and_ registers the logutil flags. In most cases, we can get away with shoving this into `servenv.ParseFlags(WithArgs)`, but there are a few places that don't use that entry point, which I found with:

```
✗ diff <(go list -f '{{ .ImportPath }}| {{ join .Deps " " }} ' ./go/cmd/... | awk -F"|" '$2 ~ " vitess.io/vitess/go/vt/servenv " {print $1}' | sed 's:vitess.io/vitess/::' | sort | uniq) <(go list -f '{{ .ImportPath }}| {{ join .Deps " " }} ' ./go/cmd/... | awk -F"|" '$2 ~ " vitess.io/vitess/go/vt/servenv " {print $1}' | sed 's:vitess.io/vitess/::' | xargs -I% grep -r "ParseFlags" % | cut -d: -f1 | xargs dirname | sort | uniq)
3,6d2
< go/cmd/query_analyzer
< go/cmd/rulesctl
< go/cmd/rulesctl/cmd
< go/cmd/rulesctl/common
9d4
< go/cmd/vtadmin
12d6
< go/cmd/vtclient
17,20d10
< go/cmd/vtctldclient
< go/cmd/vtctldclient/cli
< go/cmd/vtctldclient/command
< go/cmd/vtctldclient/docgen
28d17
< go/cmd/zk
```

## Related Issue(s)

@dbussink pointed this out to me, not sure if he's filed an issue yet.

## Checklist


deepthi (approved):

deepthi:
GH won't let me see the unit_race failure log, but that needs to be fixed.

frouioui (approved):

ajm188:
> GH won't let me see the unit_race failure log, but that needs to be fixed.

yeah; this seems related to the glog/flag stuff that @mattlord was looking into the other day. i'm also looking now, coincidentally

Issue #12188, opened 2023-01-30
Bug Report: schemadiff: invalid schema when only entity is a view reading FROM DUAL


shlomi-noach:
### Overview of the Issue

In a scenario where the entire schema is:

```sql
create view v as select 1 from dual
```

We get a 
```
view `v` has unresolved/loop dependencies
```

This is because of a delicate change in https://github.com/vitessio/vitess/pull/11944 and a hidden assumption: that there's always going to be at least one table... In our case there isn't any table, and a counter was not incremented when the algorithm assumed it would be.

### Reproduction Steps

```go
func TestNewSchemaFromQueriesViewFromDual(t *testing.T) {
	queries := []string{
		"create view v20 as select 1 from dual",
	}
	_, err := NewSchemaFromQueries(queries)
	assert.NoError(t, err)
}
```

^ the above fails with error

### Binary Version

```sh
-
```

### Operating System and Environment details

```sh
-
```

PR #12189, opened 2023-01-30
schemadiff: fix scenario where no tables exist in schema and with just views reading from DUAL

shlomi-noach:
## Description

fixes https://github.com/vitessio/vitess/issues/12188

Ensures an internal counter is at least `1` after iterating list of tables; if the list of tables is empty, the counter was never incremented.

Added tests that fail before the fix and of course pass with the fix. 

## Related Issue(s)

https://github.com/vitessio/vitess/issues/12188

## Checklist

### DIFF
--- a/go/vt/schemadiff/schema.go
+++ b/go/vt/schemadiff/schema.go
@@ -263,6 +263,12 @@ func (s *Schema) normalize() error {
 	// - then we only want views that depend on 1st level views or on tables. These are 2nd level views.
 	// - etc.
 	// we stop when we have been unable to find a view in an iteration.
+
+	// it's possible that the hasn't been any table in this schema.
+	// for purposes of the algorithm, the iterationLevel must be at least 1
+	if iterationLevel < 1 {
+		iterationLevel = 1
+	}
 	for {
 		handledAnyViewsInIteration := false
 		for _, v := range s.views {
--- a/go/vt/schemadiff/schema.go
+++ b/go/vt/schemadiff/schema.go
@@ -264,8 +264,11 @@ func (s *Schema) normalize() error {
 	// - etc.
 	// we stop when we have been unable to find a view in an iteration.
 
-	// it's possible that the hasn't been any table in this schema.
-	// for purposes of the algorithm, the iterationLevel must be at least 1
+	// It's possible that there's never been any tables in this schema. Which means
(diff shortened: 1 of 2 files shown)


dbussink (approved):

shlomi-noach:
Incorporated these changes into https://github.com/vitessio/vitess/pull/12147 ; first PR to get merged - wins

go/vt/schemadiff/schema.go
```diff
@@ -263,6 +263,12 @@ func (s *Schema) normalize() error {
 	// - then we only want views that depend on 1st level views or on tables. These are 2nd level views.
 	// - etc.
 	// we stop when we have been unable to find a view in an iteration.
+
+	// it's possible that the hasn't been any table in this schema.
+	// for purposes of the algorithm, the iterationLevel must be at least 1
```
mattlord, line 268: Consistent sentence capitalization and punctuation would make this easier to read.

`It's possible that there's never been any tables in this schema.`? Is that we mean? Or `If there are currently no tables in the schema, we still need to perform one iteration run.`

go/vt/schemadiff/schema_test.go
```diff
@@ -126,6 +126,22 @@ func TestNewSchemaFromQueriesUnresolvedAlias(t *testing.T) {
...
+	assert.NoError(t, err)
+}
+
+func TestNewSchemaFromQueriesViewFromDualImplicit(t *testing.T) {
+	queries := []string{
+		"create view v20 as select 1",
+	}
+	_, err := NewSchemaFromQueries(queries)
+	assert.NoError(t, err)
+}
```
mattlord, line 143: These are related, I guess, because we have no tables (only a from dual query)?

mattlord (approved):
LGTM. Only had a couple of minor questions.

go/vt/schemadiff/schema_test.go
shlomi-noach: Yes, exactly. Added comments to the test to clarify.

go/vt/schemadiff/schema.go
shlomi-noach: Edited the text altogether.

rsajwani (approved):

frouioui (approved):

Issue #12190, opened 2023-01-30
Bug Report: Parsing bitnum values doesn't support more than 64 bit values


dbussink:
### Overview of the Issue

```
SELECT b'11101010100101010010101010101010101010101000100100100100100101001101010101010101000001'
```

is valid in MySQL and returns binary data that can be longer than a 64 bit integer. The logic for parsing bitnum today uses parsing into an integer, but that is lossy.

### Reproduction Steps

```
SELECT b'11101010100101010010101010101010101010101000100100100100100101001101010101010101000001'
```

fails, but should return:

```
0x3AA54AAAAAA24925355541
```

### Binary Version

```sh
-
```

### Operating System and Environment details

```sh
-
```

### Log Fragments

```sh
-
```

PR #12191, opened 2023-01-30
Fix parsing of bitnum values larger than 64 bit

dbussink:
We can use big.Int directly to parse this and convert it to bytes. We don't need to parse it into an integer, nor use an intermediate hex encoding either.

## Related Issue(s)

Fixes https://github.com/vitessio/vitess/issues/12190

## Checklist


systay (approved):
nice fix

GuptaManan100 (approved):

Issue #12179, opened 2023-01-27
Bug Report: tabletgateway is making more GetTablet calls to the toposerver in comparison with discoverygateway


pbibra:
### Overview of the Issue

We are currently rolling out the `tabletgateway` and noticed that after switching over, our toposerver has been receiving an increased number of `GetTablet` calls. 

We've narrowed it down to this line of code: https://github.com/vitessio/vitess/blob/f33b86219ddc52e17a343fdf842728c0b2eae7cc/go/vt/discovery/topology_watcher.go#L202

We believe this filtering is preventing the watcher from adding tablets ignored by the healthcheck to its cache: https://github.com/slackhq/vitess/blob/ff6807c7c0fb379a69086b413b6eb09a3865e8ef/go/vt/discovery/topology_watcher.go#L177
As a result, we're constantly fetching information for already seen tablets regardless of whether `refreshKnownTablets` is false. 

This is different from the behavior in the `discoverygateway` where this filtering happens when adding the tablet to the healthcheck, not when fetching its info in `loadTablets: https://github.com/vitessio/vitess/blob/bc4a9606e1899a4223b80edfb212f32af81000ac/go/vt/discovery/legacy_topology_watcher.go#L390

### Reproduction Steps

Switching over from discoverygateway to tabletgateway

### Binary Version

```sh
v12
```

### Operating System and Environment details

```sh
N/A
```

deepthi:
I think I understand what is happening. Previously the topology watcher used to cache ALL tablets, not just the ones included in the keyspace/shard filter.
Looks like you will be able to fix this by moving the `isIncluded` check into the next 2 for loops. Check before calling `AddTablet` / `ReplaceTablet` etc. 
I do recommend creating unit tests that fail with the current code and encode the expected behavior before attempting the fix.

PR #12194, opened 2023-01-31
Bug fix: Cache filtered out tablets in toplogogy watcher to avoid unnecessary GetTablet calls to topo

brirams:
## Description

This addresses https://github.com/vitessio/vitess/issues/12179 by moving the point at which the topology watcher applies the provided `TabletFilter` in order to avoid too many `GetTablet` calls to the topology server. A failing test was added to confirm that this PR fixes the referenced issue

## Related Issue(s)

https://github.com/vitessio/vitess/issues/12179

## Checklist

### DIFF
--- a/go/vt/discovery/topology_watcher.go
+++ b/go/vt/discovery/topology_watcher.go
@@ -199,9 +199,6 @@ func (tw *TopologyWatcher) loadTablets() {
 				log.Errorf("cannot get tablet for alias %v: %v", alias, err)
 				return
 			}
-			if !(tw.tabletFilter == nil || tw.tabletFilter.IsIncluded(tablet.Tablet)) {
-				return
-			}
 			tw.mu.Lock()
 			aliasStr := topoproto.TabletAliasString(alias)
 			newTablets[aliasStr] = &tabletInfo{
@@ -217,6 +214,10 @@ func (tw *TopologyWatcher) loadTablets() {
 	tw.mu.Lock()
 
 	for alias, newVal := range newTablets {
+		if !(tw.tabletFilter == nil || tw.tabletFilter.IsIncluded(newVal.tablet)) {
+			continue
+		}
+
 		// trust the alias from topo and add it if it doesn't exist
 		if val, ok := tw.tablets[alias]; ok {
 			// check if the host and port have changed. If yes, replace tablet.
@@ -236,6 +237,10 @@ func (tw *TopologyWatcher) loadTablets() {
 	}
 
 	for _, val := range tw.tablets {
+		if !(tw.tabletFilter == nil || tw.tabletFilter.IsIncluded(val.tablet)) {
+			continue
+		}
+
 		if _, ok := newTablets[val.alias]; !ok {
 			tw.healthcheck.RemoveTablet(val.tablet)
 			topologyWatcherOperations.Add(topologyWatcherOpRemoveTablet, 1)
--- a/go/vt/discovery/topology_watcher_test.go
+++ b/go/vt/discovery/topology_watcher_test.go
@@ -506,3 +506,136 @@ func TestFilterByKeyspace(t *testing.T) {
 		}
 	}
 }
+
+func TestFilterByKeypsaceSkipsIgnoredTablets(t *testing.T) {
+	ts := memorytopo.NewServer("aa")
+	fhc := NewFakeHealthCheck(nil)
+	logger := logutil.NewMemoryLogger()
+	topologyWatcherOperations.ZeroAll()
+	counts := topologyWatcherOperations.Counts()
+	f := NewFilterByKeyspace(testKeyspacesToWatch)
+	tw := NewCellTabletsWatcher(context.Background(), ts, fhc, f, "aa", 10*time.Minute, false /*refreshKnownTablets*/, 5)
+
+	counts = checkOpCounts(t, counts, map[string]int64{})
+	checkChecksum(t, tw, 0)
+
+	// Add a tablet to the topology.
+	tablet := &topodatapb.Tablet{
+		Alias: &topodatapb.TabletAlias{
+			Cell: "aa",
+			Uid:  0,
+		},
+		Hostname: "host1",
+		PortMap: map[string]int32{
+			"vt": 123,
+		},
+		Keyspace: "ks1",
+		Shard:    "shard",
+	}
+	if err := ts.CreateTablet(context.Background(), tablet); err != nil {
+		t.Fatalf("CreateTablet failed: %v", err)
+	}
+	tw.loadTablets()
+	counts = checkOpCounts(t, counts, map[string]int64{"ListTablets": 1, "GetTablet": 1, "AddTablet": 1})
+	checkChecksum(t, tw, 3238442862)
+
+	// Check the tablet is returned by GetAllTablets().
+	allTablets := fhc.GetAllTablets()
+	key := TabletToMapKey(tablet)
+	if _, ok := allTablets[key]; !ok || len(allTablets) != 1 || !proto.Equal(allTablets[key], tablet) {
+		t.Errorf("fhc.GetAllTablets() = %+v; want %+v", allTablets, tablet)
+	}
+
+	// Add a second tablet to the topology that should get filtered out by the keyspace filter
+	tablet2 := &topodatapb.Tablet{
+		Alias: &topodatapb.TabletAlias{
+			Cell: "aa",
+			Uid:  2,
+		},
+		Hostname: "host2",
+		PortMap: map[string]int32{
+			"vt": 789,
+		},
+		Keyspace: "ks3",
+		Shard:    "shard",
+	}
+	if err := ts.CreateTablet(context.Background(), tablet2); err != nil {
+		t.Fatalf("CreateTablet failed: %v", err)
+	}
+	tw.loadTablets()
+
+	counts = checkOpCounts(t, counts, map[string]int64{"ListTablets": 1, "GetTablet": 1})
+	checkChecksum(t, tw, 2762153755)
+
+	// Check the new tablet is NOT returned by GetAllTablets().
+	allTablets = fhc.GetAllTablets()
+	key = TabletToMapKey(tablet2)
+	if _, ok := allTablets[key]; ok || len(allTablets) != 1 {
+		t.Errorf("fhc.GetAllTablets() = %+v; want %+v", allTablets, tablet2)
+	}
+
+	// Load the tablets again to show that when refreshKnownTablets is disabled,
+	// only the list is read from the topo and the checksum doesn't change
+	tw.loadTablets()
+	counts = checkOpCounts(t, counts, map[string]int64{"ListTablets": 1})
+	checkChecksum(t, tw, 2762153755)
+
+	// same tablet, different port, should update (previous
+	// one should go away, new one be added)
+	origTablet := proto.Clone(tablet).(*topodatapb.Tablet)
+	origKey := TabletToMapKey(tablet)
+	tablet.PortMap["vt"] = 456
+	if _, err := ts.UpdateTabletFields(context.Background(), tablet.Alias, func(t *topodatapb.Tablet) error {
+		t.PortMap["vt"] = 456
+		return nil
+	}); err != nil {
+		t.Fatalf("UpdateTabletFields failed: %v", err)
+	}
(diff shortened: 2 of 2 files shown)


go/vt/discovery/topology_watcher_test.go
```diff
@@ -506,3 +506,136 @@ func TestFilterByKeyspace(t *testing.T) {
 		}
 	}
 }
+
+func TestFilterByKeypsaceSkipsIgnoredTablets(t *testing.T) {
+	ts := memorytopo.NewServer("aa")
```
frouioui, line 511: Can you please add a comment that describes the Why of this test?

frouioui (approved):
LGTM. Thank you @brirams!

frouioui:
Also, `golangci-lint` is failing because of:
```go

// go/vt/discovery/topology_watcher_test.go:637:2: ineffectual assignment to counts (ineffassign)
counts = checkOpCounts(t, counts, map[string]int64{"ListTablets": 1})
```

go/vt/discovery/topology_watcher_test.go
```diff
@@ -506,3 +506,136 @@ func TestFilterByKeyspace(t *testing.T) {
...
+
+	// Remove ignored tablet and check that we didn't try to remove it from the health check
+	if err := ts.DeleteTablet(context.Background(), tablet2.Alias); err != nil {
+		t.Fatalf("DeleteTablet failed: %v", err)
+	}
+	if _, err := topo.FixShardReplication(context.Background(), ts, logger, "aa", "ks3", "shard"); err != nil {
+		t.Fatalf("FixShardReplication failed: %v", err)
+	}
+	tw.loadTablets()
+	counts = checkOpCounts(t, counts, map[string]int64{"ListTablets": 1})
```
frouioui, line 637:
```suggestion
	checkOpCounts(t, counts, map[string]int64{"ListTablets": 1})
```

go/vt/discovery/topology_watcher.go
```diff
@@ -199,9 +199,6 @@ func (tw *TopologyWatcher) loadTablets() {
 				log.Errorf("cannot get tablet for alias %v: %v", alias, err)
 				return
 			}
-			if !(tw.tabletFilter == nil || tw.tabletFilter.IsIncluded(tablet.Tablet)) {
-				return
-			}
```
mattlord, line 204: I can see why we added the other two locations. However, It's not clear to me how short circuiting this anonymous goroutine/function isn't also helpful? It still avoids unnecessarily getting the mutex and modifying the newTablets structure while holding it, doesn't it?

go/vt/discovery/topology_watcher.go
```diff
@@ -217,6 +214,10 @@ func (tw *TopologyWatcher) loadTablets() {
 	tw.mu.Lock()
 
 	for alias, newVal := range newTablets {
+		if !(tw.tabletFilter == nil || tw.tabletFilter.IsIncluded(newVal.tablet)) {
```
mattlord, line 217: Nit, and I know this is what we were previously using, but IMO this is easier to read and process as a human:
```
if tw.tabletFilter != nil && !tw.tabletFilter.IsIncluded(newVal.tablet)) {
```

go/vt/discovery/topology_watcher_test.go
```diff
@@ -506,3 +506,141 @@ func TestFilterByKeyspace(t *testing.T) {
 		}
 	}
 }
+
+// TestFilterByKeypsaceSkipsIgnoredTablets confirms a bug fix for the case when a TopologyWatcher
+// has a FilterByKeyspace TabletFilter configured along with refreshKnownTablets turned off. We want
+// to ensure that the TopologyWatcher:
+//   - does not continuosly call GetTablets for tablets that do not satisfy the filter
+//   - does not add or remove these filtered out tablets from the its healtcheck
+func TestFilterByKeypsaceSkipsIgnoredTablets(t *testing.T) {
```
mattlord, line 515: Confirmed this fails on main: 
```
--- FAIL: TestFilterByKeypsaceSkipsIgnoredTablets (0.00s)
    /Users/matt/git/vitess/go/vt/discovery/topology_watcher_test.go:573: want checksum 2762153755 got 3238442862
    /Users/matt/git/vitess/go/vt/discovery/topology_watcher_test.go:585: expected GetTablet to increase by 0, got 2 -> 3
    /Users/matt/git/vitess/go/vt/discovery/topology_watcher_test.go:586: want checksum 2762153755 got 3238442862
    /Users/matt/git/vitess/go/vt/discovery/topology_watcher_test.go:603: expected GetTablet to increase by 0, got 3 -> 4
    /Users/matt/git/vitess/go/vt/discovery/topology_watcher_test.go:610: want checksum 2762153755 got 3238442862
    /Users/matt/git/vitess/go/vt/discovery/topology_watcher_test.go:620: expected GetTablet to increase by 0, got 4 -> 5
    /Users/matt/git/vitess/go/vt/discovery/topology_watcher_test.go:621: want checksum 789108290 got 0
```

go/vt/discovery/topology_watcher_test.go
```diff
@@ -506,3 +506,141 @@ func TestFilterByKeyspace(t *testing.T) {
...
+	}
+	tw.loadTablets()
+	counts = checkOpCounts(t, counts, map[string]int64{"ListTablets": 1, "GetTablet": 1, "AddTablet": 1})
+	checkChecksum(t, tw, 3238442862)
+
+	// Check the tablet is returned by GetAllTablets().
+	allTablets := fhc.GetAllTablets()
+	key := TabletToMapKey(tablet)
+	if _, ok := allTablets[key]; !ok || len(allTablets) != 1 || !proto.Equal(allTablets[key], tablet) {
+		t.Errorf("fhc.GetAllTablets() = %+v; want %+v", allTablets, tablet)
```
mattlord, line 551: We should remove `fhc.` here as it's non-standard in the code base (not that it's somehow wrong). Even in the patch, compare it to CreateTablet just above. 

The `len(allTablets) != 1` part also feels unrelated. Not that it's wrong in this case, but I think what we're testing/confirming is that the key exists and has the data we expect. So the len just only adds noise for the human reading the test AFAICT. For this and the other similar checks, we could also make the fatal log message more clearly state what assertion failed (e.g. "tablet %+v not found in tablet list").

go/vt/discovery/topology_watcher_test.go
```diff
@@ -506,3 +506,141 @@ func TestFilterByKeyspace(t *testing.T) {
...
+	tw.loadTablets()
+
+	counts = checkOpCounts(t, counts, map[string]int64{"ListTablets": 1, "GetTablet": 1})
+	checkChecksum(t, tw, 2762153755)
+
+	// Check the new tablet is NOT returned by GetAllTablets().
+	allTablets = fhc.GetAllTablets()
+	key = TabletToMapKey(tablet2)
+	if _, ok := allTablets[key]; ok || len(allTablets) != 1 {
+		t.Errorf("fhc.GetAllTablets() = %+v; want %+v", allTablets, tablet2)
```
mattlord, line 579: Same here regarding `fhc.` (it's petty, I know).

go/vt/discovery/topology_watcher_test.go
```diff
@@ -506,3 +506,141 @@ func TestFilterByKeyspace(t *testing.T) {
...
+	}
+	tw.loadTablets()
+
+	counts = checkOpCounts(t, counts, map[string]int64{"ListTablets": 1, "GetTablet": 1})
+	checkChecksum(t, tw, 2762153755)
+
+	// Check the new tablet is NOT returned by GetAllTablets().
+	allTablets = fhc.GetAllTablets()
+	key = TabletToMapKey(tablet2)
+	if _, ok := allTablets[key]; ok || len(allTablets) != 1 {
```
mattlord, line 578: If it's there we know it failed, right? Feel like the len() check is unnecessary and a distraction.

go/vt/discovery/topology_watcher_test.go
```diff
@@ -506,3 +506,141 @@ func TestFilterByKeyspace(t *testing.T) {
...
+	counts = checkOpCounts(t, counts, map[string]int64{"ListTablets": 1})
+	checkChecksum(t, tw, 2762153755)
+
+	// same tablet, different port, should update (previous
+	// one should go away, new one be added)
+	origTablet := proto.Clone(tablet).(*topodatapb.Tablet)
+	origKey := TabletToMapKey(tablet)
+	tablet.PortMap["vt"] = 456
+	if _, err := ts.UpdateTabletFields(context.Background(), tablet.Alias, func(t *topodatapb.Tablet) error {
+		t.PortMap["vt"] = 456
```
mattlord, line 594: This is a pointer so the double assignment feels unnecessary, no? I think we can remove the first assignment on like 592.

go/vt/discovery/topology_watcher_test.go
```diff
@@ -506,3 +506,141 @@ func TestFilterByKeyspace(t *testing.T) {
...
+		return nil
+	}); err != nil {
+		t.Fatalf("UpdateTabletFields failed: %v", err)
+	}
+	tw.loadTablets()
+	allTablets = fhc.GetAllTablets()
+	key = TabletToMapKey(tablet)
+
+	counts = checkOpCounts(t, counts, map[string]int64{"ListTablets": 1})
+	if _, ok := allTablets[origKey]; !ok || len(allTablets) != 1 || !proto.Equal(allTablets[origKey], origTablet) {
```
mattlord, line 604: All we're testing/confirming in this case is that the key still exists, right? If so, we should just check for that and elide the secondary checks. The log message could be simplified to clarify specifically what assertion failed. Same nit about using `fhc.` in the log messages here.

go/vt/discovery/topology_watcher_test.go
```diff
@@ -506,3 +506,141 @@ func TestFilterByKeyspace(t *testing.T) {
...
+	// confirm state of topo
+	allTablets = fhc.GetAllTablets()
+	key = TabletToMapKey(tablet)
+	if _, ok := allTablets[key]; ok || len(allTablets) != 0 {
+		t.Errorf("fhc.GetAllTablets() = %+v; don't want %v", allTablets, key)
+	}
+	key = TabletToMapKey(tablet2)
+	if _, ok := allTablets[key]; ok || len(allTablets) != 0 {
+		t.Errorf("fhc.GetAllTablets() = %+v; want %+v", allTablets, tablet2)
+	}
```
mattlord, line 632: Similar comments here that we can be more precise about what we're really testing for in the conditionals (only including the key ones) and in the failing log message.

go/vt/discovery/topology_watcher_test.go
```diff
@@ -506,3 +506,141 @@ func TestFilterByKeyspace(t *testing.T) {
...
+
+	// Remove ignored tablet and check that we didn't try to remove it from the health check
```
mattlord, line 634: Where do we check that it wasn't removed, or we didn't even attempt to, from the health check [cache]?

mattlord (commented):
Looks good! I only had some minor nits/comments/questions/suggestions. They can be addressed in a comment reply or new commit — depending on what you think. I will do another review quickly whenever I see you respond. 

Thank you for working on this, @brirams! ❤️

go/vt/discovery/topology_watcher.go
brirams: haha -- i had to apply demorgan's law to figure out what this was doing as well. will update

go/vt/discovery/topology_watcher.go
brirams: we can't avoid the first run of this goroutine because all we have during the first run is tabletAliases, which don't contain the keyspace or shard information that we need to apply filtering. On subsequent runs, the filtered out tablets will be in the topology watcher's cache as a known tablet and having `refreshKnownTablets` set to false would short circuit the `GetTablet` goroutine. are you suggesting that we should do this filtering even when `refreshKnownTablets` is set to true?

go/vt/discovery/topology_watcher_test.go
brirams: this is checked by the `checkOpCounts` below: if it had been called on the HealthCheck, we'd get a delta for that stat.

mattlord (approved):
Nice work on this @brirams! Thanks again ❤️

PR #12195, opened 2023-01-31
[vtctldserver] OnlineDDL protobuf message introduction

ajm188:
## Description

Introducing a formal protobuf type for `OnlineDDL` types to use across the existing onlineddl code, `vtctldserver` (and client), and eventually vtadmin.

This is a prerequisite to moving any of the command functionality, as it will help us ensure we're maintaining compatibility if both the new and old code is using the same underlying types.

In order to do this I needed to slightly change some of the field names, which I've (attempted) to do in a backwards-compatible way, issuing deprecation warnings anywhere we parse in an "old" way.

## Related Issue(s)

Closes https://github.com/vitessio/vitess/issues/13713

## Checklist


go/vt/schema/internal/hooks/decode.go
shlomi-noach, line 1: Can you please explain what the functions in this file do? I'm a bit lost.

go/cmd/vtctldclient/command/schema.go
```diff
@@ -287,7 +287,7 @@ func commandReloadSchemaShard(cmd *cobra.Command, args []string) error {
 func init() {
 	ApplySchema.Flags().MarkDeprecated("--allow-long-unavailability", "")
 	ApplySchema.Flags().MarkDeprecated("--skip-preflight", "Deprecated. Assumed to be always 'true'")
-	ApplySchema.Flags().StringVar(&applySchemaOptions.DDLStrategy, "ddl-strategy", string(schema.DDLStrategyDirect), "Online DDL strategy, compatible with @@ddl_strategy session variable (examples: 'gh-ost', 'pt-osc', 'gh-ost --max-load=Threads_running=100'.")
+	ApplySchema.Flags().StringVar(&applySchemaOptions.DDLStrategy, "ddl-strategy", strings.ToLower(tabletmanagerdatapb.OnlineDDL_Strategy_name[int32(tabletmanagerdatapb.OnlineDDL_DIRECT)]), "Online DDL strategy, compatible with @@ddl_strategy session variable (examples: 'gh-ost', 'pt-osc', 'gh-ost --max-load=Threads_running=100'.")
```
shlomi-noach, line 290: Can we have a helper function to simplify `strings.ToLower(tabletmanagerdatapb.OnlineDDL_Strategy_name[int32(...)])`? This kind of logic recurs a few times in the code and I find it difficult to parse.

go/test/endtoend/onlineddl/ghost/onlineddl_ghost_test.go
```diff
@@ -221,11 +223,11 @@ func TestSchemaChange(t *testing.T) {
 
 	testWithInitialSchema(t)
 	t.Run("create non_online", func(t *testing.T) {
-		_ = testOnlineDDLStatement(t, alterTableNormalStatement, string(schema.DDLStrategyDirect), "vtctl", "non_online", "")
+		_ = testOnlineDDLStatement(t, alterTableNormalStatement, "direct", "vtctl", "non_online", "")
```
shlomi-noach, line 226: Can we have a helper function that translates `OnlineDDL_DIRECT` to the string `"direct"`?

go/vt/schema/ddl_strategy.go
```diff
@@ -99,20 +76,56 @@ func ParseDDLStrategy(strategyVariable string) (*DDLStrategySetting, error) {
...
 	return setting, nil
 }
 
+func ParseDDLStrategyName(name string) (tabletmanagerdatapb.OnlineDDL_Strategy, error) {
+	if name == "" {
+		// backward compatiblity and to handle unspecified values
+		return tabletmanagerdatapb.OnlineDDL_DIRECT, nil
+	}
+
+	lowerName := strings.ToUpper(name)
```
shlomi-noach, line 96: `lowerName := strings.ToUpper(...)` is contradictory is confusing. Let's do use `ToLower()` and look for texts like `"gh-ost"` instread of `"GH-OST"`

```suggestion
	lowerName := strings.ToLower(name)
```

go/test/endtoend/onlineddl/ghost/onlineddl_ghost_test.go
shlomi-noach: Presumably based on `tabletmanagerdatapb.OnlineDDL_Strategy_name[int32(strategy)]`?

go/vt/schema/ddl_strategy_test.go
```diff
@@ -22,21 +22,40 @@ import (
...
+
+		var assertion func(t assert.TestingT, value bool, msgAndArgs ...any) bool
+		switch direct {
+		case true:
+			assertion = assert.True
+		case false:
+			assertion = assert.False
+		}
+
+		assertion(t, NewDDLStrategySetting(strategy, "").IsDirect())
```
shlomi-noach, line 57: I think this reads simpler:
```suggestion
    assert.Equal(t, direct, NewDDLStrategySetting(strategy, "").IsDirect())
```

go/vt/schema/online_ddl.go
```diff
@@ -68,42 +76,126 @@ func validateWalk(node sqlparser.SQLNode, allowForeignKeys bool) (kontinue bool,
...
-	Schema   string      `json:"schema,omitempty"`
-	SQL      string      `json:"sql,omitempty"`
-	UUID     string      `json:"uuid,omitempty"`
-	Strategy DDLStrategy `json:"strategy,omitempty"`
-	Options  string      `json:"options,omitempty"`
+	// m is used to provide concurrent access to ReadyToComplete logic. Callers
+	// should use the IsReadyToComplete and MarkReadyToComplete methods to
+	// safely modify the value of that field rather than accessing it through
+	// the embedded protobuf message directly.
+	m sync.Mutex
```
shlomi-noach, line 85: I'm wary of adding a mutex member of `OnlineDDL`. Throughout the existing code we pass `OnlineDDL` by reference and by value, and this does not play well with mutex members as their behavior can be inconsistent. `OnlineDDL` was designed to be a data-only object. I'd say we can perhaps enforce users to use `sync/atomic` `Bool` or similar?

go/vt/schema/online_ddl.go
```diff
@@ -68,42 +76,126 @@ func validateWalk(node sqlparser.SQLNode, allowForeignKeys bool) (kontinue bool,
...
+	MigrationContext string `json:"context,omitempty" mapstructure:"context,omitempty"`
+	TabletAlias      string `json:"tablet,omitempty" mapstructure:"tablet,omitempty"`
 }
 
 // FromJSON creates an OnlineDDL from json
-func FromJSON(bytes []byte) (*OnlineDDL, error) {
-	onlineDDL := &OnlineDDL{}
-	err := json.Unmarshal(bytes, onlineDDL)
-	return onlineDDL, err
+func FromJSON(data []byte) (*OnlineDDL, error) {
```
shlomi-noach, line 94: Will it not be safer to create a temporary `struct` in the likeness of the old `OnlineDDL` struct, read the JSON into that struct, then copy over the results to the "new" OnlineDDL fields?

go/vt/schema/online_ddl.go
```diff
@@ -381,6 +481,33 @@ func (onlineDDL *OnlineDDL) IsView() bool {
...
+// IsReady returns true if the given online DDL is ready to complete, as well as
+// if it was _ever_ ready to complete. It is threadsafe.
+//
+// Do not attempt to access or modify the underlying fields directly; instead
+// use this method and its companion MarkReady.
+func (onlineDDL *OnlineDDL) IsReady() (isReady, wasReady bool) {
+	onlineDDL.m.Lock()
+	defer onlineDDL.m.Unlock()
+
+	return onlineDDL.ReadyToComplete, onlineDDL.OnlineDDL.WasReadyToComplete
```
shlomi-noach, line 493: Can we use `sync/atomic` here instead of using a Mutex?

go/vt/schema/online_ddl.go
```diff
@@ -68,42 +76,126 @@ func validateWalk(node sqlparser.SQLNode, allowForeignKeys bool) (kontinue bool,
...
-	SQL      string      `json:"sql,omitempty"`
-	UUID     string      `json:"uuid,omitempty"`
-	Strategy DDLStrategy `json:"strategy,omitempty"`
-	Options  string      `json:"options,omitempty"`
+	// m is used to provide concurrent access to ReadyToComplete logic. Callers
+	// should use the IsReadyToComplete and MarkReadyToComplete methods to
+	// safely modify the value of that field rather than accessing it through
+	// the embedded protobuf message directly.
+	m sync.Mutex
+	*tabletmanagerdatapb.OnlineDDL
```
shlomi-noach, line 86: Should we use `tabletmanagerdatapb.OnlineDDL` by value and not by pointer? This will be more inline with the current usage. Again, I'm concerned about copying `OnlineDDL` values: with `*tabletmanagerdatapb.OnlineDDL` we risk two instances of `OnlineDDL` sharing data. Maybe there's nothing in the code to worry about, and if all the tests pass then I guess everything's fine. But the original design was to share nothing, and I'm concerned we might miss a leak.

shlomi-noach (commented):
@ajm188 wow, what a massive effort! 🙇  I've left several comments throughout the PR. But, I would also like to carefully question a basic premise here. I'm not in on the many intricates of `vtctld` and you would know better than me; also, you seem to already have a plan for `OnlineDDL` commands which we haven't discussed, so I definitely might just miss something.

This change is so massive first and foremost due to moving `OnlineDDL` to `proto`. The question being: why do we need `OnlineDDL` to be in `proto`? While I don't have a clear grasp of how I'd like to see `vtctldclient OnlineDDL` commands implemented, I am not sure I see the need for `OnlineDDL` to pass as a message, in either direction. All I think is needed are a command ("cancel", "complete", "show") and a UUID/"all", for most cases. I don't see that the state of an `OnlineDDL` struct should pass back or forth?

Basically, the entire existence of the `OnlineDDL` command was made as a convenience. Most of the commands can be applied via `ApplySchema` anyway. So it's mostly syntactic sugar. It was created at a time where we envisioned `VExec` to be a new mechanism with which to control all-things-internal, but we've moved away from that idea. But the command `vtctl OnlineDDL` caught on, being useful, and so we want to keep supporting it -- however I'd hate for this otherwise syntactic-sugar command to cause such a disruption in the code/design. In my comments below I raise a few concerns over some of the changes.

I'm happy to discuss further. This is such a tremendous work 👏  and again I might miss something very important, but I also want to be cautious and make sure we don't make changes that aren't strictly necessary 🙏

shlomi-noach (commented):
@ajm188 wow, what a massive effort! 🙇  I've left several comments throughout the PR. But, I would also like to carefully question a basic premise here. I'm not in on the many intricates of `vtctld` and you would know better than me; also, you seem to already have a plan for `OnlineDDL` commands which we haven't discussed, so I definitely might just miss something.

This change is so massive first and foremost due to moving `OnlineDDL` to `proto`. The question being: why do we need `OnlineDDL` to be in `proto`? While I don't have a clear grasp of how I'd like to see `vtctldclient OnlineDDL` commands implemented, I am not sure I see the need for `OnlineDDL` to pass as a message, in either direction. All I think is needed are a command ("cancel", "complete", "show") and a UUID/"all", for most cases. I don't see that the state of an `OnlineDDL` struct should pass back or forth?

Basically, the entire existence of the `OnlineDDL` command was made as a convenience. Most of the commands can be applied via `ApplySchema` anyway. So it's mostly syntactic sugar. It was created at a time where we envisioned `VExec` to be a new mechanism with which to control all-things-internal, but we've moved away from that idea. But the command `vtctl OnlineDDL` caught on, being useful, and so we want to keep supporting it -- however I'd hate for this otherwise syntactic-sugar command to cause such a disruption in the code/design. In my comments below I raise a few concerns over some of the changes.

I'm happy to discuss further. This is such a tremendous work 👏  and again I might miss something very important, but I also want to be cautious and make sure we don't make changes that aren't strictly necessary 🙏

shlomi-noach:
(No idea why the review comment submitted twice).

Just a bit more context: the `OnlineDDL` object is used internally by the tablet, specifically by the `onlineddl/Executor` to manage the state of migrations. It's only ever used within `go/vt/vttablet/onlineddl/...`. Externally, there are commands to manipulate migrations, but then those just hand off the UUID of a migration and the keyspace. These commands do not understand the meaning of `OnlineDDL`'s structure. For those commands, IMHO there is no advantage to having a proto for `OnlineDDL`.

I think the closes to exposing `OnlineDDL` is `SHOW VITESS_MIGRATIONS` that runs a `SELECT * FROM _vt.schema_migrations`. _Some_ of the columns returned by this query, map into an `OnlineDDL` struct. But by far not all of them. The majority of columns are not represented as fields in `OnlineDDL`. Hence, the command `SHOW VITESS_MIGRATIONS` (or `OnlineDDL show`) needs to somehow return all those columns anyway; having _some_ of them in a proto-ed `OnlineDDL` doesn't solve that need and so there is IMHO no advantage to having a proto for `OnlineDDL` for this command, either.

ajm188:
@shlomi-noach yep, you are exactly correct (regarding "wait, hang on, do we need this?") and i came to the same realization yesterday when I started on `onlineddl show`. What we actually need is a protobuf representation of the `_vt.schema_migrations` table, and the _only_ things that are actually useful in this PR are the status/strategy enums and parsers, which don't even need to be defined within `tabletmanagerdata`.

Wish I had gone with the whole-hog approach rather than this way as I would have realized that sooner, but c'est la vie!

I've got the protobuf working in another branch which I'll open shortly, I have some open questions about exactly how much we want to be showing

go/vt/schema/ddl_strategy.go
ajm188: ah it should actually be `upperName` because we need the ALL_CAPS to index into the enum table

ajm188:
we don't need this, instead see #13738

PR #12197, opened 2023-02-01
[vtgate planner] Routing & Merging refactor

systay:
## Description
More info to come

## Checklist


go/vt/vtgate/planbuilder/testdata/filter_cases.json
```diff
@@ -1150,7 +1150,7 @@
       "Original": "select Id from user where 1 in ('aa', 'bb')",
       "Instructions": {
         "OperatorType": "Route",
-        "Variant": "Scatter",
+        "Variant": "None",
```
systay, line 1153: We are now using the `evalengine` to check if we can at plan-time evaluate expression. If we can, and if the result is `false`, we can use the `None` opcode which is very cheap.

go/vt/vtgate/planbuilder/testdata/info_schema57_cases.json
```diff
@@ -279,9 +279,9 @@
           "Sharded": false
         },
         "FieldQuery": "select RC.CONSTRAINT_NAME, ORDINAL_POSITION from INFORMATION_SCHEMA.KEY_COLUMN_USAGE as KCU, INFORMATION_SCHEMA.REFERENTIAL_CONSTRAINTS as RC where 1 != 1",
-        "Query": "select RC.CONSTRAINT_NAME, ORDINAL_POSITION from INFORMATION_SCHEMA.KEY_COLUMN_USAGE as KCU, INFORMATION_SCHEMA.REFERENTIAL_CONSTRAINTS as RC where KCU.TABLE_SCHEMA = :__vtschemaname and KCU.TABLE_NAME = :KCU_TABLE_NAME and KCU.COLUMN_NAME = 'id' and KCU.REFERENCED_TABLE_SCHEMA = 'test' and KCU.CONSTRAINT_NAME = 'data_type_table_id_fkey' and KCU.CONSTRAINT_NAME = RC.CONSTRAINT_NAME order by KCU.CONSTRAINT_NAME asc, KCU.COLUMN_NAME asc",
+        "Query": "select RC.CONSTRAINT_NAME, ORDINAL_POSITION from INFORMATION_SCHEMA.KEY_COLUMN_USAGE as KCU, INFORMATION_SCHEMA.REFERENTIAL_CONSTRAINTS as RC where KCU.TABLE_SCHEMA = :__vtschemaname and KCU.TABLE_NAME = :KCU_TABLE_NAME and KCU.COLUMN_NAME = 'id' and KCU.REFERENCED_TABLE_SCHEMA = :__vtschemaname and KCU.CONSTRAINT_NAME = 'data_type_table_id_fkey' and KCU.CONSTRAINT_NAME = RC.CONSTRAINT_NAME order by KCU.CONSTRAINT_NAME asc, KCU.COLUMN_NAME asc",
```
systay, line 282: The old query was wrong - we should never use the given schema name. Instead we have to replace the literal value with the argument `:__vtschemaname` which is then filled in by the vttablet with the name of the underlying MySQL database.

go/vt/vtgate/planbuilder/testdata/info_schema57_cases.json
```diff
@@ -279,9 +279,9 @@
...
         "SysTableTableName": "[KCU_TABLE_NAME:VARCHAR(\"data_type_table\")]",
-        "SysTableTableSchema": "[VARCHAR(\"test\")]",
+        "SysTableTableSchema": "[VARCHAR(\"test\"), VARCHAR(\"test\")]",
```
systay, line 284: This is wrong. I'll fix it.

go/vt/vtgate/planbuilder/testdata/info_schema57_cases.json
```diff
@@ -408,16 +408,40 @@
       "QueryType": "SELECT",
       "Original": "SELECT kcu.constraint_name constraint_name, kcu.column_name column_name, kcu.referenced_table_name referenced_table_name, kcu.referenced_column_name referenced_column_name, kcu.ordinal_position ordinal_position, kcu.table_name table_name, rc.delete_rule delete_rule, rc.update_rule update_rule FROM information_schema.key_column_usage AS kcu INNER JOIN information_schema.referential_constraints AS rc ON kcu.constraint_name = rc.constraint_name WHERE kcu.table_schema = ? AND rc.constraint_schema = ? AND kcu.referenced_column_name IS NOT NULL ORDER BY ordinal_position",
       "Instructions": {
-        "OperatorType": "Route",
-        "Variant": "DBA",
-        "Keyspace": {
-          "Name": "main",
-          "Sharded": false
+        "OperatorType": "Join",
```
systay, line 411: The old plan was wrong. Given `WHERE kcu.table_schema = ? AND rc.constraint_schema = ?`, we won't know until runtime if the user wants to look at information from the same or different keyspaces/schemas, and so merging these routes into a single one is invalid.

go/vt/vtgate/planbuilder/testdata/info_schema57_cases.json
```diff
@@ -547,7 +571,7 @@
         "FieldQuery": "select fk.referenced_table_name as to_table, fk.referenced_column_name as primary_key, fk.column_name as `column`, fk.constraint_name as `name`, rc.update_rule as on_update, rc.delete_rule as on_delete from information_schema.referential_constraints as rc, information_schema.key_column_usage as fk where 1 != 1",
         "Query": "select fk.referenced_table_name as to_table, fk.referenced_column_name as primary_key, fk.column_name as `column`, fk.constraint_name as `name`, rc.update_rule as on_update, rc.delete_rule as on_delete from information_schema.referential_constraints as rc, information_schema.key_column_usage as fk where rc.constraint_schema = :__vtschemaname and rc.table_name = :rc_table_name and fk.referenced_column_name is not null and fk.table_schema = :__vtschemaname and fk.table_name = :fk_table_name",
         "SysTableTableName": "[fk_table_name:VARCHAR(\"table_name\"), rc_table_name:VARCHAR(\"table_name\")]",
-        "SysTableTableSchema": "[VARCHAR(\"table_schema\"), VARCHAR(\"table_schema\")]",
+        "SysTableTableSchema": "[VARCHAR(\"table_schema\")]",
```
systay, line 574: Since we have merged the two routes, there is no need to specify the `table_schema` schema twice.

go/vt/vtgate/planbuilder/testdata/info_schema80_cases.json
```diff
@@ -576,17 +600,37 @@
...
-        "Keyspace": {
-          "Name": "main",
-          "Sharded": false
-        },
-        "FieldQuery": "select cc.constraint_name as `name`, cc.check_clause as expression from information_schema.check_constraints as cc, information_schema.table_constraints as tc where 1 != 1",
-        "Query": "select cc.constraint_name as `name`, cc.check_clause as expression from information_schema.check_constraints as cc, information_schema.table_constraints as tc where cc.constraint_schema = :__vtschemaname and tc.table_schema = :__vtschemaname and tc.table_name = :tc_table_name",
-        "SysTableTableName": "[tc_table_name:VARCHAR(\"table_name\")]",
-        "SysTableTableSchema": "[VARCHAR(\"constraint_schema\"), VARCHAR(\"table_schema\")]",
-        "Table": "information_schema.check_constraints, information_schema.table_constraints"
+        "OperatorType": "Join",
```
systay, line 603: Here we know that the two schemas being searched for are different - `WHERE tc.table_schema = 'table_schema' AND ... cc.constraint_schema = 'constraint_schema'`. Merging these two was wrong.

go/vt/vtgate/planbuilder/testdata/reference_cases.json
```diff
@@ -22,7 +22,7 @@
       "Original": "select * from ambiguous_ref_with_source",
       "Instructions": {
         "OperatorType": "Route",
-        "Variant": "Reference",
+        "Variant": "Unsharded",
```
systay, line 25: `ambiguous_ref_with_source` exists both as an unsharded table in the `main` keyspace, and as a reference table in the `user` keyspace. The latter is a copy of the unsharded table spread out to all shards so that all joins can be local.

During route planning we have decided that we want to send the query to the unsharded `main` keyspace. The OpCode is more accurate if it's `Unsharded` for this route.

go/vt/vtgate/planbuilder/operator_transformers.go
```diff
@@ -158,22 +157,18 @@ func routeToEngineRoute(ctx *plancontext.PlanningContext, op *operators.Route) (
...
-		vindex = op.Selected.FoundVindex
-		values = op.Selected.Values
+
+	ks, _ := ctx.VSchema.DefaultKeyspace() // we don't want to fail here - maybe the op knows where it is going
+
+	rp := &engine.RoutingParameters{
+		Opcode:   op.Routing.OpCode(),
+		Keyspace: ks,
 	}
+	op.Routing.UpdateRoutingParams(ctx, rp)
```
frouioui, line 167: It's not clear to me why `UpdateRoutingParams` woud not take care of assigning the `Opcode` and `Keyspace` fields.

go/vt/vtgate/planbuilder/operator_transformers.go
```diff
@@ -204,89 +199,68 @@ func transformRoutePlan(ctx *plancontext.PlanningContext, op *operators.Route) (
...
-	if op.Selected != nil {
-		vindex = op.Selected.FoundVindex
-		values = op.Selected.Values
-	}
 	ast := upd.AST
 	replaceSubQuery(ctx, ast)
+	rp := &engine.RoutingParameters{
+		Opcode: op.Routing.OpCode(),
+	}
+	op.Routing.UpdateRoutingParams(ctx, rp)
```
frouioui, line 207: I have the same question here. Why are we not assigning the keyspace even though we previously used to set the keyspace on the `RoutingParameters`:
```go
RoutingParameters: &engine.RoutingParameters{
			Opcode:            op.RouteOpCode,
			Keyspace:          op.Keyspace,
			Vindex:            vindex,
			Values:            values,
			TargetDestination: op.TargetDestination,
		},
```

go/vt/vtgate/planbuilder/operator_transformers.go
systay: OpCode is something that all routings need to set, so instead of setting it manually in every `Routing` implementation, I do it on the outside.

The keyspace we are setting here on the outside is the currently selected one, not the one for the query. This is then overwritten in most `Routing`s, but not all.

go/vt/vtgate/planbuilder/operators/info_schema_planning.go
```diff
@@ -0,0 +1,311 @@
...
+	"vitess.io/vitess/go/vt/servenv"
+	"vitess.io/vitess/go/vt/sqlparser"
+	"vitess.io/vitess/go/vt/vterrors"
+	"vitess.io/vitess/go/vt/vtgate/engine"
+	"vitess.io/vitess/go/vt/vtgate/evalengine"
+	"vitess.io/vitess/go/vt/vtgate/planbuilder/plancontext"
+)
+
+type InfoSchemaRouting struct {
+	// The following two fields are used when Routing information_schema queries
```
frouioui, line 39: aren't they all used when routing `information_schema` queries?

go/vt/vtgate/planbuilder/operator_transformers.go
systay: Same answer here :)

The difference here is that for UPDATE/DELETE we require a keyspace, so the default one is not needed.

go/vt/vtgate/planbuilder/operators/info_schema_planning.go
```diff
@@ -0,0 +1,311 @@
...
+	Table               *QueryTable
+}
+
+func (isr *InfoSchemaRouting) UpdateRoutingParams(ctx *plancontext.PlanningContext, rp *engine.RoutingParameters) {
+	rp.SysTableTableSchema = nil
+	for _, expr := range isr.SysTableTableSchema {
+		eexpr, err := evalengine.Translate(expr, &notImplementedSchemaInfoConverter{})
+		if err != nil {
+			// we have already tried this earlier, so we can expect it to work again
+			panic("not expected: " + err.Error())
```
frouioui, line 51: how about using `VT13001` instead of panicking?

go/vt/vtgate/planbuilder/operators/info_schema_planning.go
```diff
@@ -0,0 +1,311 @@
...
+		}
+		rp.SysTableTableName[k] = eexpr
+	}
+}
+
+func (isr *InfoSchemaRouting) Clone() Routing {
+	return &InfoSchemaRouting{
+		SysTableTableSchema: slices.Clone(isr.SysTableTableSchema),
+		SysTableTableName:   maps.Clone(isr.SysTableTableName),
+	}
```
frouioui, line 71: We do not clone the `Table` field?

go/vt/vtgate/planbuilder/operators/merging.go
```diff
@@ -0,0 +1,328 @@
...
+	}
+
+	switch {
+	// if either side is a dual query, we can always merge them together
+	case a == dual:
+		return m.merge(lhsRoute, rhsRoute, routingB)
+	case b == dual:
+		return m.merge(lhsRoute, rhsRoute, routingA)
+
+		// an unsharded/reference route can be merged with anything going to that keyspace
```
frouioui, line 54:
```suggestion
	// an unsharded/reference route can be merged with anything going to that keyspace
```

go/vt/vtgate/planbuilder/operators/merging.go
```diff
@@ -0,0 +1,328 @@
...
+	case a == anyShard && sameKeyspace:
+		return m.merge(lhsRoute, rhsRoute, routingB)
+	case b == anyShard && sameKeyspace:
+		return m.merge(lhsRoute, rhsRoute, routingA)
+
+		// None routing can always be merged, as long as we are aiming for the same keyspace
```
frouioui, line 60:
```suggestion
	// None routing can always be merged, as long as we are aiming for the same keyspace
```

go/vt/vtgate/planbuilder/operators/merging.go
```diff
@@ -0,0 +1,328 @@
...
+	case a == none && sameKeyspace:
+		return m.merge(lhsRoute, rhsRoute, routingA)
+	case b == none && sameKeyspace:
+		return m.merge(lhsRoute, rhsRoute, routingB)
+
+		// infoSchema routing is complex, so we handle it in a separate method
```
frouioui, line 66:
```suggestion
	// infoSchema routing is complex, so we handle it in a separate method
```

go/vt/vtgate/planbuilder/operators/merging.go
```diff
@@ -0,0 +1,328 @@
...
+	case a == infoSchema && b == infoSchema:
+		return tryMergeInfoSchemaRoutings(routingA, routingB, m, lhsRoute, rhsRoute)
+
+		// sharded routing is complex, so we handle it in a separate method
```
frouioui, line 70:
```suggestion
	// sharded routing is complex, so we handle it in a separate method
```

go/vt/vtgate/planbuilder/operators/merging.go
```diff
@@ -0,0 +1,328 @@
...
+		refA, ok := routingA.(*AnyShardRouting)
+		if ok {
+			if altARoute := refA.AlternateInKeyspace(routingB.Keyspace()); altARoute != nil {
+				lhsRoute = altARoute
+				routingA = lhsRoute.Routing
+				sameKeyspace = true
+			}
+		}
+	}
+	if !sameKeyspace {
```
frouioui, line 153: either the condition of these two `if`s is wrong, or, we can merge the two bocks together

frouioui: by merging the two blocks we would be able to remove a level of indentation in the whole function

systay: we are changing `sameKeyspace` inside the first `if` block. In the second we are checking if we still are on different keyspaces or if the first one already fixed it

go/vt/vtgate/planbuilder/operators/reference_routing.go
```diff
@@ -0,0 +1,139 @@
...
+
+var _ Routing = (*AnyShardRouting)(nil)
+
+func (rr *AnyShardRouting) UpdateRoutingParams(ctx *plancontext.PlanningContext, rp *engine.RoutingParameters) {
+	rp.Keyspace = rr.keyspace
+}
+
+func (rr *AnyShardRouting) Clone() Routing {
+	return &AnyShardRouting{keyspace: rr.keyspace}
+}
```
frouioui, line 41: We are not cloning the `Alternates` field?

frouioui (commented):
This is great, it's an important change that makes things easier for all of us and that fills some gaps we had, thank you @systay 🙏🏻 

I left a few questions, comments and nits after this first pass

go/vt/vtgate/planbuilder/operators/info_schema_planning.go
```diff
@@ -0,0 +1,311 @@
...
+		// we are not going to be able to route based on this expression,
+		// and might as well move on
+		return false, "", nil
+	}
+	var name string
+	if isSchemaName {
+		name = sqltypes.BvSchemaName
+	} else {
+		name = reservedVars.ReserveColName(col)
+	}
```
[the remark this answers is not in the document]
systay: meh 🤷

go/vt/vtgate/engine/route.go
```diff
@@ -158,6 +158,9 @@ func (route *Route) RouteType() string {
 
 // GetKeyspaceName specifies the Keyspace that this primitive routes to.
 func (route *Route) GetKeyspaceName() string {
+	if route.Keyspace == nil {
+		return ""
+	}
 	return route.Keyspace.Name
```
harshit-gangal, line 164: in which case we will not have a keyspace for a route?

go/vt/vtgate/executor_framework_test.go
```diff
@@ -45,359 +46,11 @@ import (
...
-			"source": "TestUnsharded.zip_detail"
-		}
-	}
-}
-`
+//go:embed executorVSchema.json
+var executorVSchema string
+
+//go:embed unshardedVschema.json
+var unshardedVSchema string
```
harshit-gangal, line 53: nit: can we move this to a different package like `./vtgate/testdata/` having json in the same package is confusing if it is only used in testing.

go/vt/vtgate/planbuilder/operator_transformers.go
```diff
@@ -124,9 +124,8 @@ func transformHorizon(ctx *plancontext.PlanningContext, op *operators.Horizon, i
 		}
 
 		return planLimit(node.Limit, plan)
-	default:
-		panic("only SELECT and UNION implement the SelectStatement interface")
 	}
+	panic("only SELECT and UNION implement the SelectStatement interface")
```
harshit-gangal, line 128: nit: the existing function can return an error, so, should we replace `panic` with an `error`

go/vt/vtgate/planbuilder/operator_transformers.go
```diff
@@ -158,25 +157,32 @@ func routeToEngineRoute(ctx *plancontext.PlanningContext, op *operators.Route) (
...
 	}, nil
 }
 
+func newRoutingParams(ctx *plancontext.PlanningContext, opCode engine.Opcode) *engine.RoutingParameters {
+	ks, _ := ctx.VSchema.DefaultKeyspace()
+	if ks == nil {
+		// if we don't have a selected keyspace, any keyspace will do
+		// this is used by operators that do not set the keyspace
+		ks, _ = ctx.VSchema.AnyKeyspace()
+	}
```
harshit-gangal, line 179: should we not check here for the opcodes which are allowed to use `AnyKeyspace`?

go/vt/vtgate/planbuilder/operators/info_schema_planning.go
```diff
@@ -0,0 +1,314 @@
...
+// what keyspace the query go to, because we don't see normalized literal values
+type InfoSchemaRouting struct {
+	SysTableTableSchema []sqlparser.Expr
+	SysTableTableName   map[string]sqlparser.Expr
+	Table               *QueryTable
+}
+
+func (isr *InfoSchemaRouting) UpdateRoutingParams(_ *plancontext.PlanningContext, rp *engine.RoutingParameters) error {
+	rp.SysTableTableSchema = nil
+	for _, expr := range isr.SysTableTableSchema {
```
harshit-gangal, line 49: now that we do not merge if SysTableTableSchema is different, should SysTableTableSchema be an Expr than slice of Expr? Similarly, Do we need SysTableTableName as a map?

go/vt/vtgate/planbuilder/operators/info_schema_planning.go
```diff
@@ -0,0 +1,314 @@
...
+
+func (isr *InfoSchemaRouting) OpCode() engine.Opcode {
+	return engine.DBA
+}
+
+func (isr *InfoSchemaRouting) Keyspace() *vindexes.Keyspace {
+	// TODO: for some info schema queries, we do know which keyspace it will go to
+	// if we had this information, more routes could be merged.
+	return nil
+}
```
harshit-gangal, line 114: what we can do is, if there is only 1 keyspace then we know for sure where the query will go to. This can help merge some routes.

go/vt/vtgate/planbuilder/operators/info_schema_planning.go
```diff
@@ -0,0 +1,314 @@
...
+	// if both sides have the same schema predicate, we can safely merge them
+	case sqlparser.Equals.Exprs(isrA.SysTableTableSchema, isrB.SysTableTableSchema):
+		for k, expr := range isrB.SysTableTableName {
+			if e, found := isrA.SysTableTableName[k]; found && !sqlparser.Equals.Expr(expr, e) {
+				// schema names are the same, but we have contradicting table names, so we give up
+				return nil, nil
+			}
+			isrA.SysTableTableName[k] = expr
+		}
+		return m.merge(lhsRoute, rhsRoute, isrA)
```
harshit-gangal, line 189: isn't same schema name enough to ensure it will go to same keyspace? I think we do not need to check if table schema is same.
If they are not present in the predicate then we should check for table name

go/vt/vtgate/planbuilder/operator_transformers.go
systay: This is just setting the default keyspace. Routings will then set the keyspace they will use, if they specify if.

go/vt/vtgate/planbuilder/operators/info_schema_planning.go
systay: yep. follow up PR though, right?

go/vt/vtgate/planbuilder/operators/info_schema_planning.go
systay: yeah, that makes sense

go/vt/vtgate/engine/route.go
systay: removed this code

frouioui (approved):
Looks good to me!

go/vt/vtgate/planbuilder/operators/reference_routing.go
```diff
@@ -0,0 +1,138 @@
...
+	return sr, nil
+}
+
+func (sr *SequenceRouting) Cost() int {
+	return 0
+}
+
+func (sr *SequenceRouting) OpCode() engine.Opcode {
+	return engine.Next
+}
```
harshit-gangal, line 134: not a deal breaker but this does not belong to `reference_routing`

go/vt/vtgate/planbuilder/operators/sharded_routing.go
```diff
@@ -0,0 +1,619 @@
...
+			for _, predicate := range sqlparser.ExtractINFromOR(or) {
+				routing, err = UpdateRoutingLogic(ctx, predicate, routing)
+				if err != nil {
+					return nil, err
+				}
+			}
+		}
+	}
+
+	return routing, nil
```
harshit-gangal, line 136: nit: unrelated to the refactor. It is easier to follow.

```go
if !sr.isScatter() {
   return routing, nil
}
....
```

go/vt/vtgate/planbuilder/operators/merging.go
```diff
@@ -0,0 +1,323 @@
...
+	case *DualRouting:
+		return dual
+	case *ShardedRouting:
+		return sharded
+	case *NoneRouting:
+		return none
+	case *TargetedRouting:
+		return targeted
+	}
+	panic("switch should be exhaustive")
```
harshit-gangal, line 180: better to include `routing` type in the panic message

harshit-gangal (approved):

go/vt/vtgate/planbuilder/operators/sharded_routing.go
systay: hard agree

harshit-gangal (approved):

go/vt/vtgate/planbuilder/operators/info_schema_planning.go
harshit-gangal: Having a task in information schema work for memory will work.

go/vt/vtgate/planbuilder/operators/info_schema_planning.go
harshit-gangal: We should add a task for it

PR #12198, opened 2023-02-01
sqlparser: Add parsing support to spatial POINT expression

Ayman161803:
Signed-off-by: Ayman161803 <mohammedaymannawaz@gmail.com>
This PR adds parsing support to spatial datatype expressions of the kind ``point(x,y)``

## Related Issue(s)
#8604 

## Checklist

## Deployment Notes

cc @GuptaManan100


Ayman161803:
I have not created new test cases because the prevailing test cases cover almost all scenarios.

go/vt/sqlparser/ast.go
```diff
@@ -3083,6 +3090,7 @@ func (*ExtractValueExpr) iCallable()                   {}
 func (*UpdateXMLExpr) iCallable()                      {}
 func (*PerformanceSchemaFuncExpr) iCallable()          {}
 func (*GTIDFuncExpr) iCallable()                       {}
+func (*PointExpr) iCallable()                              {}
```
dbussink, line 3093: I suspect the static linter checks fail on this one or the previous as they are not lined out. Please run `gofmt` for these as the CI check mentions.

go/vt/sqlparser/ast_format.go
```diff
@@ -2754,3 +2754,10 @@ func (node *Variable) Format(buf *TrackedBuffer) {
 	}
 	buf.astPrintf(node, "%v", node.Name)
 }
+
+// Format formats the node.
+func (node *PointExpr) Format(buf *TrackedBuffer) {
+	buf.WriteString("point")
+	buf.astPrintf(node, "(%v, %v)", node.XCordinate, node.YCordinate)
+
```
dbussink, line 2762: Can you remove this stray newline?

Ayman161803:
Addressed formatting issues

go/vt/sqlparser/ast_format.go
```diff
@@ -2754,3 +2754,9 @@ func (node *Variable) Format(buf *TrackedBuffer) {
 	}
 	buf.astPrintf(node, "%v", node.Name)
 }
+
+// Format formats the node.
+func (node *PointExpr) Format(buf *TrackedBuffer) {
+	buf.WriteString("point")
+	buf.astPrintf(node, "(%v, %v)", node.XCordinate, node.YCordinate)
```
GuptaManan100, line 2761: Why don't you merge the first call into the second one itself like - 
```go
buf.astPrintf(node, "point (%v, %v)", node.XCordinate, node.YCordinate)
```

go/vt/sqlparser/sql.y
```diff
@@ -7704,6 +7708,7 @@ non_reserved_keyword:
 | JSON_VALID %prec FUNCTION_CALL_NON_KEYWORD
 | JSON_VALUE %prec FUNCTION_CALL_NON_KEYWORD
 | JSON_UNQUOTE %prec FUNCTION_CALL_NON_KEYWORD
+| POINT %prec FUNCTION_CALL_NON_KEYWORD
```
GuptaManan100, line 7711: Could you move this back to its original position. This list is alphabetically sorted.

GuptaManan100 (approved):
Everything else LGTM!

GuptaManan100:
@Ayman161803 This is great work! Thank-you for the contribution!

frouioui (approved):
LGTM

go/vt/sqlparser/ast_format.go
dbussink: @GuptaManan100 How would it case it in the canonical form with this style? 

@Ayman161803 I think it's good to also add a canonical format test to the `tracked_buffer_test.go` for a `point`. So we know what it looks like / what we want it to look like there as well.

GuptaManan100:
@Ayman161803 The file `ast_format_fast.go` isn't correctly formatted. I know it is generated but it still ends up with some formatting issues in the header.

go/vt/sqlparser/ast_format_fast.go
```diff
@@ -19,7 +19,6 @@ package sqlparser
 
 import (
 	"fmt"
-
```
GuptaManan100, line 22: This is the problem I think.

go/vt/sqlparser/ast_format.go
Ayman161803: @dbussink I have added test cases assuming ``POINT(x, y)`` to be the right canonical form.

Ayman161803:
@GuptaManan100 Formatted ``ast_format_fast.go`` as suggested.

@dbussink Added canonical test cases assuming ``POINT(x, y)`` to be the right canonical form.

Thank you for the reviews

GuptaManan100 (approved):

PR #12199, opened 2023-02-01
Fix CodeQL identified issues

dbussink:
This addresses two classes of issues that CodeQL found. One is a not as strict as needed regexp for a domain match, where there's no literal `.` being matched.

The other is a user controlled XSS since we run the template with arbitrary user input. Instead we validate if the throttler with that name actually exists before trying to execute the template.

On error we also don't reflect the input to avoid the same XSS.

## Related Issue(s)

Part of #12216

## Checklist

### DIFF
--- a/go/vt/throttler/throttlerlogz.go
+++ b/go/vt/throttler/throttlerlogz.go
@@ -24,6 +24,8 @@ import (
 	"strings"
 	"time"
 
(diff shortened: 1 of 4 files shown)


rohit-nayak-ps (approved):
lgtm

PR #12200, opened 2023-02-01
Allow manually kicking off CodeQL

dbussink:
This allows for running it on other branches manually to test to see if it helps address issues that it reports.

## Related Issue(s)

Wanted to see what it reports for #12199 but we can't run it today without merging things. 

## Checklist


frouioui (approved):
LGTM

Issue #12185, opened 2023-01-29
Releases: VTOrc Unusable on Many Linux Platforms


mattlord:
##  Description

We build "generic" Linux x86_64 packages as part of each release. For example (see the Assets section at the bottom): https://github.com/vitessio/vitess/releases/tag/v15.0.2

These have historically worked quite well because the [golang runtime](https://pkg.go.dev/runtime) includes its own [`POSIX`](https://en.wikipedia.org/wiki/POSIX) and system call (syscall) implementation that is statically linked in so it does not have any runtime dependencies (dynamic linking) and most importantly it does not rely on the local libc library to interface with the OS kernel. This meant that these generic Linux x86_64 packages could be used on (virtually) any x86_64 Linux system.

In [Vitess 15 we marked VTOrc as Generally Available](https://vitess.io/blog/2022-10-26-announcing-vitess-15/) (production ready) and plan to integrate this more and more over time to where it's a critical (if not required) part of any Vitess cluster. Even today, for example, it's required for the [local examples](https://vitess.io/docs/get-started/local/) to work as [`vtorc` is responsible for choosing a tablet to promote to `PRIMARY` for each shard](https://github.com/vitessio/vitess/pull/11319).

VTOrc uses [`sqlite`](https://www.sqlite.org/index.html) for its persistent storage via the [go-sqlite3 package](https://github.com/mattn/go-sqlite3). `sqlite` is written in `C` and relies on the libc library it was compiled and linked with in order to interface with the OS kernel. VTOrc then uses [`cgo`](https://pkg.go.dev/cmd/cgo) to build bindings which allow `sqlite` to be used from Vitess via the [go-sqlite3 package](https://github.com/mattn/go-sqlite3). You can see that here: https://github.com/vitessio/vitess/blob/release-15.0/Makefile#L87-L99

[glibc](https://www.gnu.org/software/libc/) is generally ABI backward-compatible so as long as the Linux system has the same or *newer* [glibc](https://www.gnu.org/software/libc/) version installed than was used when building it, then the `vtorc` binary should execute just fine (the dynamic linker will be able resolve the [glibc](https://www.gnu.org/software/libc/) symbols in the [ELF binary](https://en.wikipedia.org/wiki/Executable_and_Linkable_Format)).

[...]

PR #12202, opened 2023-02-01
Use Ubuntu 20.04 for Release Builds

mattlord:
## Description

This moves the release package build CI workflow to Ubuntu 20.04, which is the [oldest supported Linux distro available on GitHub hosted runners](https://github.com/actions/runner-images/blob/main/README.md) today.

This will get us back to what we were using with v14 and the `vtorc` binary — which links in the native `sqlite` `C` library via `cgo` — will be usable on a larger set of Linux distros. For additional details see: https://github.com/vitessio/vitess/issues/12185

We may want to investigate additional options at some point, even going so far as attempting to statically link `glibc` in `sqlite` and using that resulting archive rather than shared object, but this PR gets us back to where we were and improves the situation significantly.

## Related Issue(s)

  - Helps with: https://github.com/vitessio/vitess/issues/12185

## Checklist

### DIFF
--- a/.github/workflows/create_release.yml
+++ b/.github/workflows/create_release.yml
@@ -10,7 +10,12 @@ on:
 jobs:
   build:
     name: Create Release
-    runs-on: ubuntu-22.04
+    # Use the oldest supported Ubuntu release for vtorc as that uses
+    # sqlite and the library dynamically links in glibc with versioned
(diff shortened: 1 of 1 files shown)


GuptaManan100 (approved):
Thank-you!

rohit-nayak-ps (approved):
lgtm

PR #12204, opened 2023-02-01
fix dual table handling

### LINKED ISSUE #6429: Creating a table named "dual" breaks schema loading.
#### Overview of the Issue

If you create a table named `dual`, vttablet can no longer load schema.

The most immediate cause is that we do not backtick-quote `dual` in this query, so it fails:

https://github.com/vitessio/vitess/blob/9f2b0fd0b77a5db9f836f750f36573bee03e7f0b/go/vt/vttablet/tabletserver/schema/load_table.go#L51

At first I thought we weren't quoting there at all, but after I tried to add it (#6427), I found out that we already quote certain identifiers via this line:

https://github.com/vitessio/vitess/blob/9f2b0fd0b77a5db9f836f750f36573bee03e7f0b/go/vt/vttablet/tabletserver/schema/load_table.go#L33

which ends up calling this:

https://github.com/vitessio/vitess/blob/9f2b0fd0b77a5db9f836f750f36573bee03e7f0b/go/vt/sqlparser/ast_funcs.go#L677-L683

The problem is that Vitess does not consider `dual` to be a keyword, even though the [MySQL docs](https://dev.mysql.com/doc/refman/5.7/en/keywords.html) say it is a reserved keyword. Unfortunately, @sougou said adding `dual` as a keyword would not be trivial. We could add a hack to fix this particular code path, but this likely won't be the only thing that breaks if you try to use a table named `dual`.

#### Reproduction Steps

```sql
create table `dual` ...
```

#### Log Fragments

vttablet:

```
tracker.go:153] Tracker's vStream ended: stream error @ c65a881b-a74f-11ea-bd03-e656b7761cef:1-3,d80528a4-a74f-11ea-8c5a-c2d0e5edc1e7:1-17: No tables used (errno 1096) (sqlstate HY000) during query: select * from dual where 1 != 1, retrying in 5 seconds
```

systay:
## Description
Fixes how we route dual tables. MySQL has a little weird way of routing the `dual` table.

```sql
mysql> select 1 from dual;
+---+
| 1 |
+---+
| 1 |
+---+
1 row in set (0.00 sec)

mysql> select 1 from `dual`;
ERROR 1046 (3D000): No database selected
mysql> use test;
Reading table information for completion of table and column names
You can turn off this feature to get a quicker startup with -A

Database changed
mysql> create table `dual`(id bigint not null);

mysql> select 1 from `dual`;
Empty set (0.00 sec)

mysql> select 1 from dual;
+---+
| 1 |
+---+
| 1 |
+---+
1 row in set (0.00 sec)

mysql> select 1 from test.dual;
Empty set (0.00 sec)
```

As can be seen in the little session above, ``` `dual` ``` and ```dual``` are handled differently. When qualified by the db name, you can select from the user table. But even when standing on the correct database, doing a `select 1 from dual` will still always use the magic dual and not the user table with the same name.

Additionally, if you prefix the table with a database name, it will never resolve to the magic dual, only complain that dual does not exist.

```sql
mysql> drop table test.dual;

mysql> select 1 from test.dual;
ERROR 1146 (42S02): Table 'test.dual' doesn't exist
```

This PR fixes so that magic `dual` is only routed to if the table is not prefixed with any db name.
That means that the `dual` table is no longer in every keyspace, and will not show up when doing `SHOW VSCHEMA TABLES`.

It's now possible to use a real user table named `dual` (why anyone would want that is a separate question).

The fact that MySQL differentiates between ``` `dual` ``` and ```dual``` needs parser and AST support that we are currently lacking.

## Related Issue(s)
Fixes #6429

## Checklist

### DIFF
--- a/go/vt/vtgate/planbuilder/operators/helpers.go
+++ b/go/vt/vtgate/planbuilder/operators/helpers.go
@@ -35,7 +35,7 @@ func Compact(ctx *plancontext.PlanningContext, op ops.Operator) (ops.Operator, e
 		Compact(ctx *plancontext.PlanningContext) (ops.Operator, rewrite.TreeIdentity, error)
 	}
 
-	newOp, err := rewrite.BottomUp(op, func(op ops.Operator) (ops.Operator, rewrite.TreeIdentity, error) {
+	newOp, err := rewrite.BottomUp(op, semantics.EmptyTableSet(), TableID, func(_ semantics.TableSet, op ops.Operator) (ops.Operator, rewrite.TreeIdentity, error) {
 		newOp, ok := op.(compactable)
 		if !ok {
 			return op, rewrite.SameTree, nil
--- a/go/vt/vtgate/planbuilder/operators/rewrite/rewriters.go
+++ b/go/vt/vtgate/planbuilder/operators/rewrite/rewriters.go
@@ -80,13 +80,17 @@ func bottomUp(root ops.Operator, rootID semantics.TableSet, resolveID func(ops.O
 	newInputs := make([]ops.Operator, len(oldInputs))
 	childID := rootID
 
+	// noLHSTableSet is used to mark which operators that do not send data from the LHS to the RHS
+	// It's only UNION at this moment, but this package can't depend on the actual operators, so
+	// we use this interface to avoid direct dependencies
 	type noLHSTableSet interface{ NoLHSTableSet() }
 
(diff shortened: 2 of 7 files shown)


GuptaManan100 (approved):

frouioui (approved):

systay:
I messed up and force pushed the wrong content to this branch before merging it. The fix is now on #12292

Issue #12143, opened 2023-01-24
Remove ReadOnly from sql_init.db


rsajwani:
After we implement super_read_only , we should remove 'readonly' from sql_init.db, because we are not adding super read only which is superset of access and implicitly turn on read_only.

PR #11706, opened 2022-11-15
[Don't review] Issue#10363 ii

rsajwani:
## Checklist

-   [ ] "Backport to:" labels have been added if this change should be back-ported
-   [ ] Tests were added or are not required
-   [ ] Documentation was added or is not required

PR #12206, opened 2023-02-01
Initialize Tablet with super_read_only mode

### LINKED ISSUE #10363: Task: Add full super_read_only support in Vitess

### LINKED ISSUE #12140: Get rid of `use_super_read_only` flag from VTTtablet
Once we implement super-read-only, we don't need `use_super_read_only` flag as this will be the default behavior and things are not expected to work without it anyways. 

File : rpc_replication.go

### LINKED ISSUE #12180: [RFC] Tablet startup in `super_read_only` mode
## TL;DR
We want replicas DB to be in `super_read_only` mode (i-e `super-read-only` set to true) all the time. Only primary DB can accept writes (i-e `super_read_only` set to false). This will make sure that nobody can modify replica from anywhere, anytime. 

## Motivation
As of today all replicas comes up in `read_only` mode, however this doesn't prevent user's like `root` and `vt_dba` who have `SUPER` privileges to change the database anytime anywhere. We want to leverage GLOBAL `super_read_only` configuration in order to protect against errant GTIDs. This will make sure that apart from primary no other component or offline system can mutate DB resulting in errant GTIDs that are then lying in wait to cause later failures, as you can see in https://github.com/vitessio/vitess/pull/9312 and https://github.com/vitessio/vitess/pull/10094 

Furthermore not all the uses-cases are ensuring that we end up replica in read-only mode. During ERS there are cases where we don't set the `read-only`. With `super_read_only` change we will make sure that we cover those cases as well.

## History

In the past we have run into situation where ERS turns primary results in errantGTIDs. The reason turns out to be cases where offline processes or operators might have mutated replica schema and it make them potential for errant GTIDs. Here are some of the issues which indicated that we ran into situations like this in production systems.

https://github.com/vitessio/vitess/pull/10363
https://github.com/vitessio/vitess/pull/10094
https://github.com/vitessio/vitess/pull/9312
https://github.com/vitessio/vitess/pull/10448

## Possible Approaches

We had tried few times to implement this `super_read_only` change but we ran into different ranges of issue mainly due to de-centralized logic of setting replica to read/write mode and `withDDL` all across code base. I am listing here few changes done in the past for `super_read_only`, but later on due to regression they have be reverted.

https://github.com/vitessio/vitess/pull/11706
https://github.com/vitessio/vitess/pull/10094
https://github.com/vitessio/vitess/pull/9312
https://github.com/vitessio/vitess/pull/10448

With schema initialization changes https://github.com/vitessio/vitess/pull/11520, we believe it will be much easier now to implement `super_read_only` change.

## Proposed Solution

This change should be build on top of https://github.com/vitessio/vitess/pull/11520, where instead of using `withddl` we use declarative approach. Using declarative approach has helped us condense all our schema changes to one place. This provide us an opportunity to apply `super_read_only` changes to our DB. 

Every MySql instance will bootstrap in `super_read_only` mode. Only during init_db we switch `super_read_only` `OFF` temporarily in order to perform some mutations like creating necessary users and permission. During Reparenting we will change `super_read_only` values for primary and replica as require. Here is the quick summary of reparenting operations.

All together there are few operations which are called across reparenting. I am summarizing how they are going to change `super_read_only` status. This will help to understand individual reparenting operations `PRS`, `ERS` & `ExternallyReparenting`.

PromoteReplica --> Sets `super_read_only` to `OFF` for given Replica Tablet
DemotePrimary --> Sets `super_read_only` to `ON` for given Primary Tablet

```
InitShardPrimary
	Calls InitPrimary
        ---> Calls tm.ChangeTypeLocked with DBAAction.ReadWriteAction 
                 ---> Call SetReadOnly(false)
```

### PRS
    switch {
	case currentPrimary == nil && ev.ShardInfo.PrimaryAlias == nil:
		// Case (1): no primary has been elected ever. Initialize
		// the primary-elect tablet
		reparentJournalPos, err = pr.performInitialPromotion(ctx, ev.NewPrimary, opts)
		---> Calls InitPrimary
			---> Calls tm.ChangeTypeLocked with DBAAction.ReadWriteAction
				---> Calls sql.setReadOnly(false)
	case currentPrimary == nil && ev.ShardInfo.PrimaryAlias != nil:
		// Case (2): no clear current primary. Try to find a safe promotion
		// candidate, and promote to it.
		reparentJournalPos, err = pr.performPotentialPromotion(ctx, keyspace, shard, ev.NewPrimary, tabletMap, opts)
		---> Calls PromoteReplica
		         ---> Call tm.ChangeTypeLocked with DBAAction.ReadWriteAction
				 ---> Calls sql.setReadOnly(false)
	case topoproto.TabletAliasEqual(currentPrimary.Alias, opts.NewPrimaryAlias):
		// Case (3): desired new primary is the current primary. Attempt to fix
		// up replicas to recover from a previous partial promotion.
		reparentJournalPos, err = pr.performPartialPromotionRecovery(ctx, ev.NewPrimary)
		---> Calls tm.SetReadWrite(true)
                          ---> Calls setReadOnly(false)
	default:
		// Case (4): desired primary and current primary differ. Do a graceful
		// demotion-then-promotion.
		reparentJournalPos, err = pr.performGracefulPromotion(ctx, ev, keyspace, shard, currentPrimary, ev.NewPrimary, tabletMap, opts)
		---> Calls DemotePrimary
			---> Calls sql.setSuperReadOnly(true)
		Calls tm.promoteReplica 
			---> Calls tm.ChangeTypeLocked with DBAAction.ReadWriteAction
			         ---> Calls sql.setReadOnly(false)
           	
### ERS
	Calls InitPrimary if its uninitialize cluster , in this case PrimarAlias is empty
		---> Calls tm.ChangeTypeLocked with DBAAction.ReadWriteAction
		         ---> Calls sql.setReadOnly(false)
         ---OR---
   	Calls tm.promoteReplica 
         	---> Call tm.ChangeTypeLocked with DBAAction.ReadWriteAction
			---> Calls sql.setReadOnly(false)

         Call reparentReplica which sets the source of replication (setreplicationsource)

### ExternallyReparentShard
```
Vitess expects that the user has set the database into ReadWrite mode before calling this reparenting type.
```
I am listing here major changes which is needed for this RFC.

- [x] my.cnf contains super-read-only, to ensure all Mysql instances comes up in `super-read-only` mode.
- [x] init_db.sql is executed after sql get initialized. We switch `super_read_only` off temporarily in order to perform some mutations like creating necessary users and permission.
- [x] unit test doesn't need `super_read_only`, so we will have a separate init_db.sql file for unit tests.
- [x] All places where we have to switch `ON` read/write we will turn off `read_only`. All places where we have to switch `OFF` read/write we will turn on `super_read_only`
- [x] Adding isSuperReadOnly property in fullStatus, so that tablet state can be queried through vtctld.

rsajwani:
## Description
As of today MySQL server which is controlled by VTTablet, starts in read-only mode. This is because Vitess controls the MySQL cluster topology and chooses one tablet to be the primary for each shard, and only that one MySQL server should be read-write. All other servers in that shard are replicas, and should not be taking any writes. However this doesn't prevent users like `root` and `vt_dba` who have `SUPER` privileges from writing to the database on replicas and causing errant GTIDs. We want to leverage GLOBAL `super_read_only` configuration in order to protect against these scenarios. This will make sure that apart from primary no other component or offline system can mutate DB resulting in errant GTIDs that are then lying in wait to cause later failures, as you can see in https://github.com/vitessio/vitess/pull/9312 and https://github.com/vitessio/vitess/pull/10094 

Furthermore, it is possible that due to program bugs or unexpected failures during PRS/ERS, we end up with a replica that is not in read-only mode. With super-read-only change we will make sure that we cover those cases as well.  

## Tablet schema refactoring
This change is built on top of https://github.com/vitessio/vitess/pull/11520, where instead of using `withddl` we use a declarative approach to initializing vttablet's sidecar schema. Using a declarative approach helps us condense all our schema changes to one place, which allows us to make changes to the code to apply `super-read-only` to the DB. 
The previous approach of using `withddl` meant that schema changes were scattered all across the code, which made it extremely difficult to do the `super-read-only` change.

## Details 

This PR usese https://github.com/vitessio/vitess/pull/11520 as a building block. I am listing major changes below to help you review the PR.

- mysql57.cnf & mysql80.cnf contains super-read-only, to ensure all Mysql instances comes up in super-read-only mode.
- During init_db.sql we switch super-read-only `OFF` temporarily in order to perform some mutations like creating necessary users and permissions.
- Since VTTestServer doesn't need super-read-only, therefore we need to come up with a separate SQL initialization (init_testserver_db.sql) for it. 
- Adding `# {{custom_sql}}`  inside init_db.sql file. This is because in our tests we modify this file and add custom SQL like (change password of DBA user etc). Now that we expect setting super-read-only `ON` to execute at the end, we need these custom sqls to be added before that last line. See example: go/test/endtoend/backup/vtbackup/main_test.go
- During PRS & ERS, we make sure that only primary turns off super-read-only and rest of replica are strictly in super-read-only mode.
- Adding 'SuperReadOnly' property, which is available through VtctldClient GetFullStatus.

## For Reference 
Some related work items done in the past

https://github.com/vitessio/vitess/pull/11706
https://github.com/vitessio/vitess/pull/10094
https://github.com/vitessio/vitess/pull/9312
https://github.com/vitessio/vitess/pull/10448

## Unresolved Issues during code changes:
I have filed few issues which I will need to resolve once I check-in this PR.

- https://github.com/vitessio/vitess/issues/12143

## Related Issue(s)

Fixes #10363
Fixes #12180
Fixes #12140 

## Test Results

vtctld command to show super_read_only state

```
 vtctldclient --server localhost:15999 GetFullStatus zone1-0000000101
{
  "server_id": 493124066,
  "server_uuid": "ed6d02e6-a35b-11ed-b467-9cf2e7b0ba93",
  "replication_status": null,
  "primary_status": {
    "position": "MySQL56/ed6d02e6-a35b-11ed-b467-9cf2e7b0ba93:1-26",
    "file_position": "FilePos/vt-0000000101-bin.000001:14540"
  },
  "gtid_purged": "MySQL56/",
  "version": "8.0.31",
  "version_comment": "Homebrew",
  "read_only": false,
  "gtid_mode": "ON",
  "binlog_format": "ROW",
  "binlog_row_image": "FULL",
  "log_bin_enabled": true,
  "log_replica_updates": true,
  "semi_sync_primary_enabled": true,
  "semi_sync_replica_enabled": true,
  "semi_sync_primary_status": true,
  "semi_sync_replica_status": false,
  "semi_sync_primary_clients": 1,
  "semi_sync_primary_timeout": "1000000000000000000",
  "semi_sync_wait_for_replica_count": 1,
  "super_read_only": false
}
```

## Checklist


go/cmd/vtbackup/vtbackup.go
```diff
@@ -302,6 +303,19 @@ func takeBackup(ctx context.Context, topoServer *topo.Server, backupStorage back
 		if err := mysqld.ResetReplication(ctx); err != nil {
 			return fmt.Errorf("can't reset replication: %v", err)
 		}
+		// We need to switch off super-read-only before we create database.
+		resetFunc, err := mysqld.SetSuperReadOnly(false)
```
rsajwani, line 307: resetFunc will bring back super_read_only to its original value.

go/mysql/collations/integration/helpers_test.go
```diff
@@ -137,7 +137,7 @@ func verifyWeightString(t *testing.T, local collations.Collation, remote *remote
 }
 
 func exec(t *testing.T, conn *mysql.Conn, query string) *sqltypes.Result {
-	res, err := conn.ExecuteFetch(query, -1, true)
+	res, err := conn.ExecuteFetchWithSuperReadOnlyHandling(query, -1, true)
```
rsajwani, line 140: These test does not have vttablets. They just create mysql instance and run test against them...

go/test/endtoend/utils/mysql.go
```diff
@@ -114,7 +114,7 @@ func createInitSQLFile(mysqlDir, ksName string) (string, error) {
 		return "", err
 	}
 	defer f.Close()
-
+	_, _ = f.WriteString("SET GLOBAL super_read_only='OFF';")
```
rsajwani, line 117: In these test we create Mysql without any vttablets (i-e barecluster) and then try to create database so I have to set super_read_only to false given its not going through flow of typical vttablet.

go/vt/mysqlctl/backup.go
```diff
@@ -344,21 +342,6 @@ func Restore(ctx context.Context, params RestoreParams) (*BackupManifest, error)
 		return nil, err
 	}
```
rsajwani, line 344: given we are starting mysql above. It will always start in replica mode (super_read_only to true). I don't think this code block in any case will run for primary ...

go/vt/mysqlctl/builtinbackupengine.go
```diff
@@ -322,17 +323,22 @@ func (be *BuiltinBackupEngine) executeFullBackup(ctx context.Context, params Bac
 	}
 
 	// get the read-only flag
-	readOnly, err = params.Mysqld.IsReadOnly()
+	readOnly, err := params.Mysqld.IsReadOnly()
```
rsajwani, line 326: printing read_only for debugging purpose.

go/vt/mysqlctl/builtinbackupengine.go
```diff
@@ -322,17 +323,22 @@ func (be *BuiltinBackupEngine) executeFullBackup(ctx context.Context, params Bac
...
+	}
+	log.Infof("Flag values during full backup, read_only: %v, super_read_only:%v", readOnly, superReadOnly)
 
 	// get the replication position
 	if sourceIsPrimary {
-		if !readOnly {
-			params.Logger.Infof("turning primary read-only before backup")
-			if err = params.Mysqld.SetReadOnly(true); err != nil {
-				return false, vterrors.Wrap(err, "can't set read-only status")
+		if !superReadOnly {
```
rsajwani, line 338: Before shutting down we make sure primary is super_read_only. This is existing logic. I just change it instead of read_only to use super_read_only.

go/vt/mysqlctl/builtinbackupengine.go
```diff
@@ -385,9 +391,11 @@ func (be *BuiltinBackupEngine) executeFullBackup(ctx context.Context, params Bac
 	}
 
 	// And set read-only mode
-	params.Logger.Infof("resetting mysqld read-only to %v", readOnly)
-	if err := params.Mysqld.SetReadOnly(readOnly); err != nil {
-		return usable, err
+	if resetSuperReadOnlyFunc != nil {
```
rsajwani, line 394: I think we don't need this anymore but I am keeping it because I am not sure why on first place this logic of switching read_only was inserted. Primary should alway be non_read_only.

go/vt/mysqlctl/mysqld.go
```diff
@@ -1251,6 +1253,29 @@ func (mysqld *Mysqld) applyBinlogFile(binlogFile string, includeGTIDs mysql.GTID
 		args := []string{
 			"--defaults-extra-file=" + cnf,
 		}
+
+		// We disable super_read_only, in case it is in the default MySQL startup
+		// parameters.  We do it blindly, since this will fail on MariaDB, which doesn't
+		// have super_read_only This is safe, since we're restarting MySQL after the restore anyway
+		log.Infof("applyBinlogFile: disabling super_read_only")
+		resetFunc, err := mysqld.SetSuperReadOnly(false)
```
rsajwani, line 1261: mysqlbin is not able to replay logs if super_read_only is set, so we have to disable it temporarily until applyBin log is done.

go/test/endtoend/backup/vtbackup/main_test.go
```diff
@@ -126,13 +134,20 @@ func TestMain(m *testing.M) {
 				return 1, err
 			}
 		}
-
-		// Create database
-		for _, tablet := range []cluster.Vttablet{*primary, *replica1} {
-			if err := tablet.VttabletProcess.CreateDB(keyspaceName); err != nil {
-				return 1, err
-			}
+		vtTabletVersion, err := cluster.GetMajorVersion("vttablet")
```
rsajwani, line 137: For debugging purpose

go/test/endtoend/cluster/mysqlctl_process.go
```diff
@@ -249,6 +257,22 @@ func MysqlCtlProcessInstanceOptionalInit(tabletUID int, mySQLPort int, tmpDirect
 	return mysqlctl
 }
 
+func isSQLFlavor() (bool, error) {
```
rohit-nayak-ps, line 260: We are only using this to check if it supports super-read-only to find which init_db.sql to use, right? Maybe we can have a utility function like `GetInitDBSQLFilename()` which has the logic in the file and returns the correct init filename?

go/test/endtoend/vtorc/readtopologyinstance/main_test.go
```diff
@@ -104,8 +105,19 @@ func TestReadTopologyInstanceBufferable(t *testing.T) {
 	assert.Equal(t, primaryInstance.ReplicationIOThreadState, inst.ReplicationThreadStateNoThread)
 	assert.Equal(t, primaryInstance.ReplicationSQLThreadState, inst.ReplicationThreadStateNoThread)
 
-	// insert an errant GTID in the replica
-	_, err = utils.RunSQL(t, "insert into vt_insert_test(id, msg) values (10173, 'test 178342')", replica, "vt_ks")
+	// Insert an errant GTID in the replica
```
rohit-nayak-ps, line 108: We are changing the way to create an errant gtid, because replicas have super-read-only = "ON" now, right? Maybe it will be useful to add that to the comments as well.

go/vt/mysqlctl/replication.go
```diff
@@ -254,14 +272,50 @@ func (mysqld *Mysqld) SetReadOnly(on bool) error {
...
+		query += "'ON'"
 	} else {
-		query += "OFF"
+		query += "'OFF'"
 	}
-	return mysqld.ExecuteSuperQuery(context.TODO(), query)
+	if err := mysqld.ExecuteSuperQuery(context.TODO(), query); err != nil {
+		return nil, err
+	}
+	if superReadOnlyEnabled {
```
rohit-nayak-ps, line 314: We only return `returnFunc` if `superReadOnlyEnabled` is true. So when is `resetFunc()` ever returned as the `returnFunc`? We should add some unit tests around this function to clarify and validate the logic.

```
	if on && !superReadOnlyEnabled {
		returnFunc = resetFunc
	}
	if !on && superReadOnlyEnabled {
		returnFunc = setFunc
	}
```

go/vt/vttest/environment.go
```diff
@@ -155,6 +163,22 @@ func (env *LocalTestEnv) MySQLManager(mycnf []string, snapshot string) (MySQLMan
 	}, nil
 }
 
+func isRunningMariaDB() bool {
```
rohit-nayak-ps, line 166: It might be good to add a `IsMariaDB()` API to `mysqlctl` where it will can be implemented just as `return capabilities.isMariaDB()`

mattlord: This is already there in the `mysql` package's flavor interface:
```
// IsMariaDB returns true iff the other side of the client connection
// is identified as MariaDB. Most applications should not care, but
// this is useful in tests.
func (c *Conn) IsMariaDB() bool {
	switch c.flavor.(type) {
	case mariadbFlavor101, mariadbFlavor102:
		return true
	}
	return false
}
```

deepthi:
@GuptaManan100 
> Furthermore not all the uses-cases are ensuring that we end up replica in read-only mode. During ERS there are cases where we don't set the read-only.

What is this referring to?

go/test/endtoend/cluster/mysqlctl_process.go
```diff
@@ -236,11 +237,18 @@ func (mysqlctl *MysqlctlProcess) Connect(ctx context.Context, username string) (
 // MysqlCtlProcessInstanceOptionalInit returns a Mysqlctl handle for mysqlctl process
 // configured with the given Config.
 func MysqlCtlProcessInstanceOptionalInit(tabletUID int, mySQLPort int, tmpDirectory string, initMySQL bool) *MysqlctlProcess {
+	var initFile = path.Join(os.Getenv("VTROOT"), "/config/init_db.sql") //default value
+	if isSQL, err := isSQLFlavor(); err == nil {
+		if !isSQL {
```
rsajwani, line 242: This is to support MariaDBMigration test

go/test/endtoend/cluster/mysqlctld_process.go
```diff
@@ -145,11 +145,18 @@ func (mysqlctld *MysqlctldProcess) CleanupFiles(tabletUID int) {
 // MysqlCtldProcessInstance returns a Mysqlctld handle for mysqlctld process
 // configured with the given Config.
 func MysqlCtldProcessInstance(tabletUID int, mySQLPort int, tmpDirectory string) *MysqlctldProcess {
+	var initFile = path.Join(os.Getenv("VTROOT"), "/config/init_db.sql") //default value
+	if isSQL, err := isSQLFlavor(); err == nil {
```
rsajwani, line 149: This is to support MariaDBMigration test

go/vt/vttablet/tabletmanager/tm_init.go
```diff
@@ -76,14 +76,13 @@ const denyListQueryList string = "DenyListQueryRules"
...
-	initTags           flagutil.StringMapValue
-
+	tabletHostname       string
+	initKeyspace         string
+	initShard            string
+	initTabletType       string
+	initDbNameOverride   string
+	skipBuildInfoTags    = "/.*/"
+	initTags             flagutil.StringMapValue
 	initPopulateMetadata bool
```
rsajwani, line 86: remove redundant extra line

go/vt/vttablet/tabletmanager/tm_init.go
```diff
@@ -96,7 +95,6 @@ func registerInitFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&initDbNameOverride, "init_db_name_override", initDbNameOverride, "(init parameter) override the name of the db used by vttablet. Without this flag, the db name defaults to vt_<keyspacename>")
 	fs.StringVar(&skipBuildInfoTags, "vttablet_skip_buildinfo_tags", skipBuildInfoTags, "comma-separated list of buildinfo tags to skip from merging with --init_tags. each tag is either an exact match or a regular expression of the form '/regexp/'.")
 	fs.Var(&initTags, "init_tags", "(init parameter) comma separated list of key:value pairs used to tag the tablet")
-
```
rsajwani, line 99: remove redundant extra line

GuptaManan100:
@deepthi We only set read only mode when we demote primary. The problem in ERS is that if the primary tablet is unavailable (for some duration), then we `DemotePrimary` call doesn't succeed. If the tablet is back again, in time for `SetReplicationSource` call to be successful, we end up not converting the tablet to read-only state. 
Currently, the code to set read only (demotePrimary) is called from the `DemotePrimary` rpc and `endPrimaryTerm` from shard-sync.

go/test/endtoend/cluster/vtbackup_process.go
```diff
@@ -54,7 +54,6 @@ type VtbackupProcess struct {
 
 // Setup starts vtbackup process with required arguements
 func (vtbackup *VtbackupProcess) Setup() (err error) {
-
```
rsajwani, line 57: extra space ... fixing it

go/vt/mysqlctl/backup.go
```diff
@@ -27,13 +27,11 @@ import (
 
 	"github.com/spf13/pflag"
 
-	"vitess.io/vitess/go/vt/servenv"
-
-	"vitess.io/vitess/go/mysql"
 	"vitess.io/vitess/go/stats"
```
rsajwani, line 30: fixing import format.

go/vt/vtexplain/vtexplain_test.go
```diff
@@ -24,15 +24,15 @@ import (
...
-	"vitess.io/vitess/go/vt/vttablet/tabletserver/tabletenv/tabletenvtest"
-
 	"github.com/google/go-cmp/cmp"
 	"github.com/stretchr/testify/require"
 
 	"vitess.io/vitess/go/vt/key"
-	querypb "vitess.io/vitess/go/vt/proto/query"
 	"vitess.io/vitess/go/vt/proto/topodata"
 	"vitess.io/vitess/go/vt/topo"
+	"vitess.io/vitess/go/vt/vttablet/tabletserver/tabletenv/tabletenvtest"
```
rsajwani, line 33: fixing import format.

go/test/endtoend/backup/vtbackup/backup_only_test.go
```diff
@@ -294,17 +300,18 @@ func resetTabletDirectory(t *testing.T, tablet cluster.Vttablet, initMysql bool)
 
 func tearDown(t *testing.T, initMysql bool) {
 	// reset replication
+	for _, db := range []string{"_vt", "vt_insert_test"} {
+		_, err := primary.VttabletProcess.QueryTablet(fmt.Sprintf("drop database if exists %s", db), keyspaceName, true)
```
rsajwani, line 304: we drop database from primary and wait for replication to catch-up

go/test/endtoend/backup/vtbackup/main_test.go
```diff
@@ -89,8 +90,15 @@ func TestMain(m *testing.M) {
 		dbCredentialFile = cluster.WriteDbCredentialToTmp(localCluster.TmpDirectory)
 		initDb, _ := os.ReadFile(path.Join(os.Getenv("VTROOT"), "/config/init_db.sql"))
 		sql := string(initDb)
+		// Since password update is DML we need to insert it before we disable
```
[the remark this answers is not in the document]
rsajwani: fixed

go/test/endtoend/backup/vtctlbackup/backup_utils.go
```diff
@@ -207,9 +214,6 @@ func LaunchCluster(setupType int, streamMode string, stripes int, cDetails *Comp
 	}
 
 	for _, tablet := range []cluster.Vttablet{*primary, *replica1} {
```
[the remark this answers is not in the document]
rsajwani: I think we discussed about it.. I was redundant call anyways regardless of this change.

go/vt/mysqlctl/replication.go
rsajwani: we return 'reset' or 'set' function as returnFunc with following logic
        
        // we are asked to turn on super_read_only but original value is false
        // then return resetFunc, that can be used as defer by caller
	if on && !superReadOnlyEnabled {
		returnFunc = resetFunc
	}
        // we are asked to turn off super_read_only but original value is true
        // then return setFunc, that can be used as defer by caller
	if !on && superReadOnlyEnabled {
		returnFunc = setFunc
	}

rsajwani: for any other case return function is going to return nil .. since there is no internal state change.

rsajwani: fixed. yes there was bug :)

rsajwani:
> @deepthi We only set read only mode when we demote primary. The problem in ERS is that if the primary tablet is unavailable (for some duration), then we `DemotePrimary` call doesn't succeed. If the tablet is back again, in time for `SetReplicationSource` call to be successful, we end up not converting the tablet to read-only state. Currently, the code to set read only (demotePrimary) is called from the `DemotePrimary` rpc and `endPrimaryTerm` from shard-sync.

We don't have test case for this scenario in CI I believe.

go/test/endtoend/backup/vtbackup/backup_only_test.go
rsajwani: previous logic was we delete database from all the replica manually.

go/test/endtoend/utils/mysql.go
```diff
@@ -60,24 +61,24 @@ func CreateMysqldAndMycnf(tabletUID uint32, mysqlSocket string, mysqlPort int) (
 	return mysqlctl.NewMysqld(&cfg), mycnf, nil
 }
 
-func NewMySQLWithDetails(port int, hostname, dbName string, schemaSQL ...string) (mysql.ConnParams, func(), error) {
+func NewMySQLWithDetails(port int, hostname, dbName string, schemaSQL ...string) (mysql.ConnParams, *mysqlctl.Mysqld, func(), error) {
```
rsajwani, line 64: Modify method to return mysqld reference as well. I did this in order to write unit test for SetSuperReadOnly.

deepthi:
@rsajwani can you update the `Details` section to be current? Some of the line items seem to be obsolete.
For example, do we still need a separate sql for MariaDB? Are you still having to make changes in tests to run super-read-only off and on? I thought that isn't needed any more in the latest iteration of the code.

go/mysql/query.go
```diff
@@ -328,6 +328,39 @@ func (c *Conn) ExecuteFetchMulti(query string, maxrows int, wantfields bool) (re
 	return res, more, err
 }
 
+// ExecuteFetchWithSuperReadOnlyHandling should be used if you are executing a query
+// on any tablet regardless of tablet type.
+// This function will temporarily make the mysql instance read-write and
+// re-enable read-only mode after the query is executed if needed.
+func (c *Conn) ExecuteFetchWithSuperReadOnlyHandling(query string, maxrows int, wantfields bool) (result *sqltypes.Result, err error) {
```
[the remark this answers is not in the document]
deepthi: The body of this function does not use any unexposed funcs, they are all public, so it should be possible to move this to vttablet_process.go which is the place from which we are calling it.

go/mysql/query.go
```diff
@@ -328,6 +328,39 @@ func (c *Conn) ExecuteFetchMulti(query string, maxrows int, wantfields bool) (re
...
+				superReadOnlyEnabled = true
+				if _, err = c.ExecuteFetch("SET GLOBAL super_read_only='OFF'", 1, false); err != nil {
+					return nil, err
+				}
+			}
+		}
+	}
+
+	result, _, err = c.ExecuteFetchMulti(query, maxrows, wantfields)
+	// TODO: may be use it in defer()
```
deepthi, line 355: yes, good idea.

go/test/endtoend/recovery/unshardedrecovery/recovery.go
```diff
@@ -91,16 +92,19 @@ func TestMainImpl(m *testing.M) {
...
 		localCluster.Keyspaces = append(localCluster.Keyspaces, *keyspace)
 
+		// Create a new init_db.sql file that sets up passwords for all users.
+		// Then we use a db-credentials-file with the passwords.
+		// We need these users during backup-restore process for example vt_repl is needed during ResetReplication while taking backup.
 		dbCredentialFile = cluster.WriteDbCredentialToTmp(localCluster.TmpDirectory)
 		initDb, _ := os.ReadFile(path.Join(os.Getenv("VTROOT"), "/config/init_db.sql"))
 		sql := string(initDb)
+		// since original initDb does not have updated passwords. Therefore, update the init file with passwords SQL.
+		oldAlterTableMode := `SET GLOBAL old_alter_table = ON;`
```
deepthi, line 102: I wish we had a code comment as to why we want `old_alter_table` to be ON. I'm sure there was a good reason.

go/test/endtoend/utils/mysql.go
deepthi, line 64:
```suggestion
func NewMySQLWithMysqld(port int, hostname, dbName string, schemaSQL ...string) (mysql.ConnParams, *mysqlctl.Mysqld, func(), error) {
```

go/test/endtoend/utils/mysql.go
deepthi: let's catch the error and fail if it is not nil.

go/test/endtoend/vtorc/readtopologyinstance/main_test.go
```diff
@@ -104,8 +105,19 @@ func TestReadTopologyInstanceBufferable(t *testing.T) {
...
+	// The way to do this is to disable global recoveries, stop replication and inject an errant GTID.
+	// After this we restart the replication and enable the recoveries again.
+	err = logic.DisableRecovery()
+	require.NoError(t, err)
+	err = utils.RunSQLs(t, []string{`STOP SLAVE;`,
+		`SET GTID_NEXT="12345678-1234-1234-1234-123456789012:1";`,
+		`BEGIN;`, `COMMIT;`,
+		`SET GTID_NEXT="AUTOMATIC";`,
+		`START SLAVE;`,
```
deepthi, line 117: we'll need to remember to change `SLAVE` -> `REPLICA` after MySQL 5.7 support goes EOL.

go/vt/mysqlctl/backup.go
deepthi: Since we no longer populate those metadata tables, it is safe to remove this.

go/vt/mysqlctl/builtinbackupengine.go
deepthi: Is it even necessary to get / check / print readOnly any more? We could delete these lines of code, we are only interested in super_read_only from now on.

go/vt/mysqlctl/builtinbackupengine.go
deepthi: This code block is not gated by `sourceIsPrimary`. The previous authors are just making sure this is set to whatever it was before the backup was started. It could have been done differently, basically something like
```
if sourceIsPrimary {
    ... params.Mysqld.SetReadOnly(false) ...
    ...
}
```

go/vt/mysqlctl/mysqld.go
```diff
@@ -1251,6 +1251,29 @@ func (mysqld *Mysqld) applyBinlogFile(binlogFile string, includeGTIDs mysql.GTID
...
+		if err != nil {
+			if strings.Contains(err.Error(), mysql.ERUnknownSystemVariable.ToString()) {
+				log.Warningf("Restore: server does not know about super_read_only, continuing anyway...")
```
deepthi, line 1262: This is not the `Restore` func, so the log message needs to be adjusted.

go/vt/mysqlctl/mysqld.go
```diff
@@ -1251,6 +1251,29 @@ func (mysqld *Mysqld) applyBinlogFile(binlogFile string, includeGTIDs mysql.GTID
...
+			} else {
+				log.Errorf("Restore: unexpected error while trying to set super_read_only: %v", err)
```
deepthi, line 1264: This is not the `Restore` func, so the log message needs to be adjusted.

go/vt/vttablet/tabletmanager/rpc_replication.go
```diff
@@ -37,11 +37,11 @@ import (
 	topodatapb "vitess.io/vitess/go/vt/proto/topodata"
 )
 
-var setSuperReadOnly bool
 var disableReplicationManager bool
 
 func registerReplicationFlags(fs *pflag.FlagSet) {
-	fs.BoolVar(&setSuperReadOnly, "use_super_read_only", setSuperReadOnly, "Set super_read_only flag when performing planned failover.")
+	fs.Bool("use_super_read_only", true, "Set super_read_only flag when performing planned failover.")
+	fs.MarkDeprecated("use_super_read_only", "From v17 onwards MySQL server will always starts as super_read_only=`true`")
```
deepthi, line 44:
```suggestion
	fs.MarkDeprecated("use_super_read_only", "From v17 onwards MySQL server will always start with super_read_only=ON")
```

go/vt/vttablet/tabletmanager/rpc_replication.go
```diff
@@ -463,23 +468,17 @@ func (tm *TabletManager) demotePrimary(ctx context.Context, revertPartialFailure
...
-		if err := tm.MysqlDaemon.SetSuperReadOnly(true); err != nil {
-			if strings.Contains(err.Error(), mysql.ERUnknownSystemVariable.ToString()) {
-				log.Warningf("server does not know about super_read_only, continuing anyway...")
-			} else {
-				return nil, err
-			}
-		}
-	} else {
-		if err := tm.MysqlDaemon.SetReadOnly(true); err != nil {
+	if _, err := tm.MysqlDaemon.SetSuperReadOnly(true); err != nil {
```
deepthi, line 474: Should we not be doing this in a defer call to make sure it is always executed?

deepthi (commented):
I'm very pleased with this way this has finally turned out.
The most important thing to fix from this review is to remove the facility to bypass super_read_only from the mysql package. Most of the others are minor changes.

go/cmd/vtbackup/vtbackup.go
```diff
@@ -404,6 +417,8 @@ func takeBackup(ctx context.Context, topoServer *topo.Server, backupStorage back
 		return err
 	}
 
+	log.Infof("takeBackup: primary position is: %s", primaryPos.String())
```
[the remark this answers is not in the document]
rsajwani: I find it useful during debugging so I thought of keeping it..

go/test/endtoend/backup/vtbackup/backup_only_test.go
```diff
@@ -62,12 +63,21 @@ func TestTabletInitialBackup(t *testing.T) {
 	// Initialize the tablets
 	initTablets(t, false, false)
 
-	// Restore the Tablets
+	vtTabletVersion, err := cluster.GetMajorVersion("vttablet")
+	require.NoError(t, err)
+	// For all version above v15, each replica will start in super-read-only mode.
+	if vtTabletVersion > 16 {
```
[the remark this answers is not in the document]
rsajwani: Not this is not downgrade/upgrade. Instead of writing a separate test. I just add this in existing test to check super_read_only functionality. Basically it testing from v17 onwards all replica should come up as super_read_only.

go/test/endtoend/backup/vtbackup/backup_only_test.go
```diff
@@ -62,12 +63,21 @@ func TestTabletInitialBackup(t *testing.T) {
...
+		err := primary.VttabletProcess.CreateDB("testDB")
+		require.ErrorContains(t, err, "The MySQL server is running with the --super-read-only option so it cannot execute this statement")
+		err = replica1.VttabletProcess.CreateDB("testDB")
+		require.ErrorContains(t, err, "The MySQL server is running with the --super-read-only option so it cannot execute this statement")
```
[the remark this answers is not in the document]
rsajwani: Yes I am piggybacking on existing test. Instead of writing a separate test. I just add this in existing test to check super_read_only functionality. Basically it testing from v17 onwards all replica should come up as super_read_only.

go/test/endtoend/recovery/unshardedrecovery/recovery.go
rsajwani: yeah , I didn't research that part..

config/init_db.sql
```diff
@@ -11,6 +11,12 @@
 ###############################################################################
 # Equivalent of mysql_secure_installation
 ###############################################################################
+# We need to ensure that super_read_only is disabled so that we can execute
+# these commands. Note that disabling it does NOT disable read_only.
+# We save the current value so that we only re-enable it at the end if it was
+# enabled before.
+SET @original_super_read_only=IF(@@global.super_read_only=1, 'ON', 'OFF');
+SET GLOBAL super_read_only='OFF';
```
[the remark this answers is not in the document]
rsajwani: fixed

config/mycnf/test-suite.cnf
```diff
@@ -23,3 +23,4 @@ sql_mode = STRICT_TRANS_TABLES
 
 # set a short heartbeat interval in order to detect failures quickly
 slave_net_timeout = 4
+super-read-only = false
```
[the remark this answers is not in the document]
rsajwani: done. 
Test-suite is mainly used for vttestserver, which does not need super-read-only. Therefore, we are disabling it.

doc/releasenotes/17_0_0_summary.md
```diff
@@ -9,6 +9,9 @@
     - [Builtin backup: read buffering flags](#builtin-backup-read-buffering-flags)
   - **[New stats](#new-stats)**
     - [Detailed backup and restore stats](#detailed-backup-and-restore-stats)
+  - **[VTTablet](#vttablet)**
+    - [VTTablet: Initializing all replica DB with super_read_only](#vttablet-initialization)
```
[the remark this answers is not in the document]
rsajwani: fixed

doc/releasenotes/17_0_0_summary.md
```diff
@@ -160,3 +163,15 @@ Some notes to help understand these metrics:
  * `DurationByPhaseSeconds["CatchUpReplication"]` measures how long it took to catch-up replication after the restore phase.
  * `DurationByPhaseSeconds["RestoreLastBackup"]` measures to the duration of the restore phase.
  * `RestoreDurationNanoseconds["-.-.Restore"]` also measures to the duration of the restore phase.
+
+### <a id="vttablet"/> VTTablet
+#### <a id="vttablet-initialization"/> Initializing all replica DB with super_read_only
+In order to prevent SUPER privileged users like `root` or `vt_dba` to produce errant GTIDs anywhere anytime, all the replica DBs are initialized with the Mysql
```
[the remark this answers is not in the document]
rsajwani: done

doc/releasenotes/17_0_0_summary.md
```diff
@@ -160,3 +163,15 @@ Some notes to help understand these metrics:
...
+global variable `super_read_only` value set to `ON`. During re-parenting, we set `super_read_only` to `OFF` for the promoted primary tablet. This will allow the
```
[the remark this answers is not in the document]
rsajwani: done

doc/releasenotes/17_0_0_summary.md
```diff
@@ -160,3 +163,15 @@ Some notes to help understand these metrics:
...
+primary to accept writes. All replica except the primary will still have their global variable `super_read_only` set to `ON`. This will make sure that apart from
```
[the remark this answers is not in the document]
rsajwani: done

doc/releasenotes/17_0_0_summary.md
```diff
@@ -160,3 +163,15 @@ Some notes to help understand these metrics:
...
+the replication no other component or offline system can mutate replica DB resulting in errant GTIDs that are then lying in wait to cause later failures.
```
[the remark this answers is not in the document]
rsajwani: done

go/mysql/query.go
```diff
@@ -328,6 +328,39 @@ func (c *Conn) ExecuteFetchMulti(query string, maxrows int, wantfields bool) (re
 	return res, more, err
 }
 
+// ExecuteFetchWithSuperReadOnlyHandling should be used if you are executing a query
+// on any tablet regardless of tablet type.
+// This function will temporarily make the mysql instance read-write and
```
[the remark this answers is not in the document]
rsajwani: done

go/test/endtoend/backup/vtbackup/main_test.go
```diff
@@ -86,11 +87,16 @@ func TestMain(m *testing.M) {
 
 		// Create a new init_db.sql file that sets up passwords for all users.
 		// Then we use a db-credentials-file with the passwords.
+		// We need these users during backup-restore process for example vt_repl is needed during ResetReplication while taking backup.
 		dbCredentialFile = cluster.WriteDbCredentialToTmp(localCluster.TmpDirectory)
 		initDb, _ := os.ReadFile(path.Join(os.Getenv("VTROOT"), "/config/init_db.sql"))
 		sql := string(initDb)
+		// since original initDb does not have updated passwords. Therefore, update the init file with passwords SQL.
```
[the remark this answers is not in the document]
rsajwani: hmm .. I though I sort of stated it on #90
// We need these users/passwords during backup-restore process for example vt_repl is needed during ResetReplication while taking backup.

go/test/endtoend/cluster/mysqlctl_process.go
```diff
@@ -236,11 +237,16 @@ func (mysqlctl *MysqlctlProcess) Connect(ctx context.Context, username string) (
 // MysqlCtlProcessInstanceOptionalInit returns a Mysqlctl handle for mysqlctl process
 // configured with the given Config.
 func MysqlCtlProcessInstanceOptionalInit(tabletUID int, mySQLPort int, tmpDirectory string, initMySQL bool) *MysqlctlProcess {
+	var initFile = path.Join(os.Getenv("VTROOT"), "/config/init_db.sql") //default value
+	updatedInitFile, err := getInitDBFile()
+	if err == nil {
```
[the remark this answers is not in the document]
rsajwani: hmm I thought its ok here to drop to default init_db in case of error.

go/test/endtoend/utils/mysql.go
rsajwani: done.

go/vt/mysqlctl/builtinbackupengine.go
rsajwani: I want to keep it for debugging purpose initially. This will help me debug if there are production issues. I will remove them later as part of another work item. 
https://github.com/vitessio/vitess/issues/12143

go/vt/mysqlctl/builtinbackupengine.go
rsajwani: I will just then create parity with existing code then.

go/vt/mysqlctl/mysqld.go
rsajwani: done

go/vt/mysqlctl/mysqld.go
rsajwani: done

go/vt/vttablet/tabletmanager/rpc_replication.go
rsajwani: I am afraid if I put it in defer call then even in case of partial failures (due to which we won't able to demote) we will still  turn it to super_read_only to 'ON'. 
Furthermore if you see the defer call down handles the same scenario where in case of partial failures or any erro (FinalErr) it sets ReadOnly to 'False'. This is to make sure we don't end up changing the state in case demotePrimary fails for any reason.

deepthi: This sounds good.

go/test/endtoend/cluster/mysqlctl_process.go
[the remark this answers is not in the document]
deepthi: let's put a comment in that case.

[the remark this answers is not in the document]
deepthi: if the intent of this func is to test that you can override the config, we should fail the test if that fails.

config/mycnf/test-suite.cnf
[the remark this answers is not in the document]
deepthi: I think "does not need" isn't accurate. vttestserver uses a single MySQL for primary and replicas, so it is not possible to run it with super-read-only.
Aren't there other tests (vttablet/endtoend) which also use this cnf file?

go/vt/mysqlctl/builtinbackupengine.go
```diff
@@ -398,9 +405,9 @@ func (be *BuiltinBackupEngine) executeFullBackup(ctx context.Context, params Bac
 		return usable, vterrors.Wrap(err, "can't restart mysqld")
 	}
 
-	// And set read-only mode
-	params.Logger.Infof("resetting mysqld read-only to %v", readOnly)
-	if err := params.Mysqld.SetReadOnly(readOnly); err != nil {
+	// Resetting super_read_only back to its original value
+	params.Logger.Infof("resetting mysqld super_read-only to %v", superReadOnly)
```
deepthi, line 409:
```suggestion
	params.Logger.Infof("resetting mysqld super-read-only to %v", superReadOnly)
```

rsajwani: yeah , I choose '_', I might have missed this place.. so will fix it

systay (changes requested):
The changes in `vtexplain` are hard to understand, and seem unrelated to the rest of the changes.

go/vt/vtexplain/vtexplain_vttablet.go
```diff
@@ -506,13 +506,35 @@ func (t *explainTablet) HandleQuery(c *mysql.Conn, query string, callback func(*
...
 	result := tEnv.getResult(query)
-	emptyResult := &sqltypes.Result{}
-	if sidecardb.MatchesInitQuery(query) {
-		return callback(emptyResult)
-	}
+
 	if result != nil {
 		return callback(result)
 	}
+
```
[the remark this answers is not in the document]
rsajwani: @systay, did you go through changes in this PR (https://github.com/vitessio/vitess/pull/11520/files#diff-24ee60b7c22c2ac8452465a23da2335941f725ef220c689f814d49ac5abec28a) ... As stated, the original changes are done in sidecar db , I just refactored it ..

config/mycnf/test-suite.cnf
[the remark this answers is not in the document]
rsajwani: right now testserver is the only one using test-suite.cnf. I have changed the wordings though.

go/vt/vtexplain/vtexplain_vttablet.go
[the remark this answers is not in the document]
deepthi: @systay the way I understand it is that a previous PR introduced an out-of-package dependency that is not really needed and Rameez has cleaned that up to do this the "right" way.

go/test/endtoend/cluster/mysqlctl_process.go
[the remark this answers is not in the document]
rsajwani: fixed. It resulted in a bit of refacotring ... all the places where we are using MysqlCtlProcessInstanceOptionalInit is not using this pattern (in ordrer to return error as well )

```
mysqlctlProcess, err := MysqlCtlProcessInstanceOptionalInit(tablet.TabletUID, tablet.MySQLPort, cluster.TmpDirectory, !cluster.ReusingVTDATAROOT)
			if err != nil {
				return err
			}
			tablet.MysqlctlProcess = *mysqlctlProcess
```

go/test/endtoend/backup/vtbackup/main_test.go
```diff
@@ -112,7 +119,11 @@ func TestMain(m *testing.M) {
 			tablet.VttabletProcess.ExtraArgs = commonTabletArg
 			tablet.VttabletProcess.SupportsBackup = true
 
-			tablet.MysqlctlProcess = *cluster.MysqlCtlProcessInstance(tablet.TabletUID, tablet.MySQLPort, localCluster.TmpDirectory)
+			mysqlctlProcess, err := cluster.MysqlCtlProcessInstance(tablet.TabletUID, tablet.MySQLPort, localCluster.TmpDirectory)
```
rsajwani, line 122: this is due to refactoring done in go/test/endtoend/cluster/mysqlctl_process.go

go/test/endtoend/recovery/unshardedrecovery/recovery.go
rsajwani: https://github.com/vitessio/vitess/issues/8315, its mainly due to xtrabackup

go/test/endtoend/vault/dbcreds_secret.json
```diff
@@ -1,17 +1,23 @@
 {
+  "root": [
+    "RootPass"
+  ],
   "vt_app": [
-    "password"
+    "VtAppPass"
```
rsajwani, line 6: changing these passwords so I can reuse GetPasswordUpdateSQL. This will make test consistent as well ...

config/mycnf/default.cnf
```diff
@@ -14,9 +14,6 @@ port = {{.MysqlPort}}
 secure-file-priv = {{.SecureFilePriv}}
 {{end}}
 
-# all db instances should start in read-only mode - once the db is started and
-# fully functional, we'll push it into read-write mode
-read-only
```
deepthi, line 19: I thought we agreed that we still need to keep this because of MariaDB. Other option is to add these lines to mariadb10.cnf.

go/test/endtoend/cluster/mysqlctl_process.go
```diff
@@ -235,36 +236,63 @@ func (mysqlctl *MysqlctlProcess) Connect(ctx context.Context, username string) (
...
+		return "", err
+	}
+	flavor, _, err := mysqlctl.ParseVersionString(versionStr)
+	if err != nil {
+		return "", err
+	}
+	if flavor == mysqlctl.FlavorMySQL || flavor == mysqlctl.FlavorPercona {
+		return path.Join(os.Getenv("VTROOT"), "/config/init_db.sql"), nil
+	}
+	return path.Join(os.Getenv("VTROOT"), "go/test/endtoend/vreplication/testdata/config/init_testserver_db.sql"), nil
```
deepthi, line 269: Can you put a comment here why we are using the testserver_db.sql?

config/init_db.sql
```diff
@@ -77,3 +83,9 @@ FLUSH PRIVILEGES;
 
 RESET SLAVE ALL;
 RESET MASTER;
+
+# custom sql is used to add custom scripts like creating users/passwords. We use it in our tests
+# add custom sql here
+
+# We need to set super_read_only back to what it was before
+SET GLOBAL super_read_only=IFNULL(@original_super_read_only, 'OFF');
```
deepthi, line 91: @mattlord does this look correct to you?

go/vt/mysqlctl/builtinbackupengine.go
```diff
@@ -398,9 +405,9 @@ func (be *BuiltinBackupEngine) executeFullBackup(ctx context.Context, params Bac
...
+	if _, err := params.Mysqld.SetSuperReadOnly(superReadOnly); err != nil {
 		return usable, err
 	}
```
deepthi, line 412: Should we not move this to a defer call right after we set it at the beginning?
I realize it was already like this before this PR.

deepthi (approved):
Almost there 😄 
Approving provisionally, the remaining changes are pretty small, but should be addressed before we merge.

systay (approved):

go/test/endtoend/utils/utils.go
```diff
@@ -288,3 +287,23 @@ func convertToMap(input interface{}) map[string]interface{} {
 	output := input.(map[string]interface{})
 	return output
 }
+
+func GetInitDBSQL(initDBSQL string, updatedPasswords string, oldAlterTableMode string) (string, error) {
+	// Since password update is DML we need to insert it before we disable
+	// super_read_only therefore doing the split below.
+	splitString := strings.Split(initDBSQL, "# add custom sql here")
```
mattlord, line 294: As noted before, I think we can make this a more clear placeholder. For example:
```
# {{custom_sql}}
```

And would be nicer if the string was a const as well.

go/test/endtoend/recovery/pitr/shardedpitr_test.go
```diff
@@ -509,8 +522,16 @@ func testTabletRecovery(t *testing.T, binlogServer *binLogServer, lookupTimeout,
 }
 
 func launchRecoveryTablet(t *testing.T, tablet *cluster.Vttablet, binlogServer *binLogServer, lookupTimeout, restoreKeyspaceName, shardName string) {
-	tablet.MysqlctlProcess = *cluster.MysqlCtlProcessInstance(tablet.TabletUID, tablet.MySQLPort, clusterInstance.TmpDirectory)
-	err := tablet.MysqlctlProcess.Start()
+	mysqlctlProcess, err := cluster.MysqlCtlProcessInstance(tablet.TabletUID, tablet.MySQLPort, clusterInstance.TmpDirectory)
+	if err != nil {
+		require.NoError(t, err)
+	}
```
mattlord, line 528:
```suggestion
	require.NoError(t, err)
```

go/test/endtoend/recovery/pitr/shardedpitr_test.go
```diff
@@ -416,10 +425,27 @@ func initializeCluster(t *testing.T) {
...
+	// The original init_db.sql does not have any passwords. Here we update the init file with passwords
+	sql, err = utils.GetInitDBSQL(sql, cluster.GetPasswordUpdateSQL(clusterInstance), "")
+	if err != nil {
+		require.NoError(t, err, "expected to load init_db file")
+	}
+	initDBFileWithPassword = path.Join(clusterInstance.TmpDirectory, "init_db_with_passwords.sql")
+	err = os.WriteFile(initDBFileWithPassword, []byte(sql), 0666)
+	if err != nil {
+		require.NoError(t, err, "expected to load init_db file")
+	}
```
mattlord, line 440:
```suggestion
	require.NoError(t, err, "expected to load init_db file")
```

go/test/endtoend/backup/vtbackup/backup_only_test.go
```diff
@@ -367,3 +374,39 @@ func verifyDisableEnableRedoLogs(ctx context.Context, t *testing.T, mysqlSocket
 		}
 	}
 }
+
+// This helper function wait for all replicas to catch-up the replication.
+// It does this by querying the status detail url of each replica and find the lag.
+func waitForReplicationToCatchup(tablets []cluster.Vttablet) bool {
+	endTime := time.Now().Add(time.Second * 30)
+	timeout := time.After(time.Until(endTime))
```
mattlord, line 382: Nit, but better to use a Timer or a Ticker with a context.WithTimeout. This is fine though as it's only test code.

go/test/endtoend/backup/vtbackup/backup_only_test.go
```diff
@@ -367,3 +374,39 @@ func verifyDisableEnableRedoLogs(ctx context.Context, t *testing.T, mysqlSocket
...
+		Key   string
+		Class string
+		Value string
+	}
+	// defining a struct instance
+	var statuslst []kv
+	for {
+		select {
+		case <-timeout:
+			return false
```
mattlord, line 394: We should either require.Fatal here or use require.True() with the return value in the calling location (it's currently ignored by the caller).

go/test/endtoend/backup/vtbackup/backup_only_test.go
```diff
@@ -367,3 +374,39 @@ func verifyDisableEnableRedoLogs(ctx context.Context, t *testing.T, mysqlSocket
...
+		default:
+			var replicaCount = 0
```
mattlord, line 396: Nit, but ~ `caughtUpReplicaCount` or `nonLaggingReplicaCount` would be better.

go/test/endtoend/backup/vtbackup/backup_only_test.go
```diff
@@ -294,17 +300,18 @@ func resetTabletDirectory(t *testing.T, tablet cluster.Vttablet, initMysql bool)
 
 func tearDown(t *testing.T, initMysql bool) {
 	// reset replication
+	for _, db := range []string{"_vt", "vt_insert_test"} {
+		_, err := primary.VttabletProcess.QueryTablet(fmt.Sprintf("drop database if exists %s", db), keyspaceName, true)
+		require.Nil(t, err)
+	}
+	waitForReplicationToCatchup([]cluster.Vttablet{*replica1, *replica2})
```
mattlord, line 307: As the code is today, we should use:
```
succeeded := waitForReplicationToCatchup([]cluster.Vttablet{*replica1, *replica2})
require.True(t, succeeded, "Timed out waiting for all replicas to catch up")
```

go/test/endtoend/backup/vtbackup/backup_only_test.go
```diff
@@ -367,3 +374,39 @@ func verifyDisableEnableRedoLogs(ctx context.Context, t *testing.T, mysqlSocket
...
+		default:
+			var replicaCount = 0
+			for _, tablet := range tablets {
+				status := tablet.VttabletProcess.GetStatusDetails()
+				json.Unmarshal([]byte(status), &statuslst)
+				for _, obj := range statuslst {
+					if obj.Key == "Replication Lag" && obj.Value == "0s" {
+						replicaCount++
+					}
+				}
```
mattlord, line 404:
```suggestion
				if lag, err := jsonparser.GetString("Replication Lag"); err == nil && lag == "0s" {
					replicaCount++
				}
```

go/test/endtoend/backup/vtbackup/backup_only_test.go
```diff
@@ -294,17 +300,18 @@ func resetTabletDirectory(t *testing.T, tablet cluster.Vttablet, initMysql bool)
 
 func tearDown(t *testing.T, initMysql bool) {
 	// reset replication
+	for _, db := range []string{"_vt", "vt_insert_test"} {
```
mattlord, line 303: I feel like this should be a variable, so that creation and deletion are kept in sync if and when these values change.

go/test/endtoend/vault/vault_test.go
```diff
@@ -251,10 +252,25 @@ func initializeClusterLate(t *testing.T) {
...
 
+	initDb, _ := os.ReadFile(path.Join(os.Getenv("VTROOT"), "/config/init_db.sql"))
+	sql := string(initDb)
+	// The original init_db.sql does not have any passwords. Here we update the init file with passwords
+	sql, err = utils.GetInitDBSQL(sql, cluster.GetPasswordUpdateSQL(clusterInstance), "")
+	if err != nil {
+		require.NoError(t, err, "expected to load init_db file")
+	}
+	newInitDBFile := path.Join(clusterInstance.TmpDirectory, "init_db_with_passwords.sql")
+	err = os.WriteFile(newInitDBFile, []byte(sql), 0666)
```
mattlord, line 263: Same nit/note here about broad permissions (anyone can read/write to the file). Is this needed?

mattlord (commented):
Nice work on this, @rsajwani ! I only had minor nits/comments/suggestions. We can discuss those and quickly get this thing merged already! 🙂

config/init_db.sql
deepthi: oh yeah, that definitely needs to be `ON`.

doc/releasenotes/17_0_0_summary.md
```diff
@@ -197,4 +200,15 @@ These stats are deprecated in v17.
...
 | `backup_duration_seconds` | `BackupDurationNanoseconds` |
-| `restore_duration_seconds` | `RestoreDurationNanoseconds` |
\ No newline at end of file
+| `restore_duration_seconds` | `RestoreDurationNanoseconds` |
+### <a id="vttablet"/> VTTablet
+#### <a id="vttablet-initialization"/> Initializing all replicas with super_read_only
+In order to prevent SUPER privileged users like `root` or `vt_dba` from producing errant GTIDs on replicas, all the replica MySQL servers are initialized with the MySQL
+global variable `super_read_only` value set to `ON`. During failovers, we set `super_read_only` to `OFF` for the promoted primary tablet. This will allow the
+primary to accept writes. All replicas except the primary will still have their global variable `super_read_only` set to `ON`. This will make sure that apart from
+MySQL replication no other component or offline system can write directly to a replica.
```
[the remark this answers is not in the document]
rsajwani: I used operator word

go/test/endtoend/backup/vtbackup/backup_only_test.go
rsajwani: not used anywhere else so we should be good here.

go/test/endtoend/backup/vtbackup/backup_only_test.go
rsajwani: I am finding it hard to parse through json parser . therefore I am keepign it with my original approach

go/test/endtoend/backup/vtbackup/backup_only_test.go
```diff
@@ -57,6 +57,8 @@ func TestTabletInitialBackup(t *testing.T) {
 	//    - list the backups, remove them
 	defer cluster.PanicHandler(t)
 
+	waitForReplicationToCatchup([]cluster.Vttablet{*replica1, *replica2})
```
mattlord, line 60: Just a reminder that if vtorc is used in these tests, which tablets are replicas is no longer fixed. That's a bigger issue though, not related to this PR.

go/test/endtoend/vtorc/utils/utils.go
```diff
@@ -589,7 +589,8 @@ func RunSQL(t *testing.T, sql string, tablet *cluster.Vttablet, db string) (*sql
 func RunSQLs(t *testing.T, sqls []string, tablet *cluster.Vttablet, db string) error {
 	// Get Connection
 	tabletParams := getMysqlConnParam(tablet, db)
-	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	var timeoutDuration = time.Duration(5 * len(sqls))
+	ctx, cancel := context.WithTimeout(context.Background(), timeoutDuration*time.Second)
```
mattlord, line 593: Non-blocking nit, but IMO it's more natural to do this all in the duration definition:
```
var timeoutDuration = time.Duration(5 * time.Second * len(sqls))
ctx, cancel := context.WithTimeout(context.Background(), timeoutDuration)
```

mattlord (approved):
LGTM! ❤️ I only had one insignificant nit. Up to you if you think it's worth changing.

PR #12210, opened 2023-02-01
grpcvtgateconn: add Dial and deprecate DialWithOpts

### LINKED ISSUE #12209: Bug Report: context shadowing issue with grpcvtgateconn.DialWithOpts
### Overview of the Issue

The current first few lines of this function are:

```go
// DialWithOpts allows for custom dial options to be set on a vtgateConn.
func DialWithOpts(ctx context.Context, opts ...grpc.DialOption) vtgateconn.DialerFunc {
	return func(ctx context.Context, address string) (vtgateconn.Impl, error) {
```

As a result, the outer context is completely ignored due to Go shadowing rules. In addition, the inner context is currently unused because we call `grpcclient.Dial` rather than `DialContext`.

I will send a PR to fix both problems.

### Reproduction Steps

n/a

### Binary Version

```sh
main
```

```sh
n/a
```

### Log Fragments

_No response_

mdlayher:
## Description

Resolves a linked issue relating to Go variable scopes.

## Related Issue(s)

Fixes #12209.

## Checklist

### DIFF
--- a/go/test/endtoend/vtgate/grpc_server_auth_static/main_test.go
+++ b/go/test/endtoend/vtgate/grpc_server_auth_static/main_test.go
@@ -196,7 +196,7 @@ func TestUnauthenticatedUser(t *testing.T) {
 func dialVTGate(ctx context.Context, t *testing.T, username string, password string) (*vtgateconn.VTGateConn, error) {
 	clientCreds := &grpcclient.StaticAuthClientCreds{Username: username, Password: password}
 	creds := grpc.WithPerRPCCredentials(clientCreds)
-	dialerFunc := grpcvtgateconn.DialWithOpts(ctx, creds)
+	dialerFunc := grpcvtgateconn.Dial(creds)
 	dialerName := t.Name()
 	vtgateconn.RegisterDialer(dialerName, dialerFunc)
 	return vtgateconn.DialProtocol(ctx, dialerName, vtgateGrpcAddress)
--- a/go/vt/vitessdriver/driver.go
+++ b/go/vt/vitessdriver/driver.go
@@ -89,7 +89,7 @@ func OpenWithConfiguration(c Configuration) (*sql.DB, error) {
 	}
 
 	if len(c.GRPCDialOptions) != 0 {
-		vtgateconn.RegisterDialer(c.Protocol, grpcvtgateconn.DialWithOpts(context.TODO(), c.GRPCDialOptions...))
+		vtgateconn.RegisterDialer(c.Protocol, grpcvtgateconn.Dial(c.GRPCDialOptions...))
 	}
 
 	return sql.Open(c.DriverName, json)
--- a/go/vt/vtgate/grpcvtgateconn/conn.go
+++ b/go/vt/vtgate/grpcvtgateconn/conn.go
@@ -73,11 +73,11 @@ type vtgateConn struct {
 }
 
 func dial(ctx context.Context, addr string) (vtgateconn.Impl, error) {
-	return DialWithOpts(ctx)(ctx, addr)
+	return Dial()(ctx, addr)
 }
 
-// DialWithOpts allows for custom dial options to be set on a vtgateConn.
-func DialWithOpts(ctx context.Context, opts ...grpc.DialOption) vtgateconn.DialerFunc {
+// Dial produces a vtgateconn.DialerFunc with custom options.
+func Dial(opts ...grpc.DialOption) vtgateconn.DialerFunc {
 	return func(ctx context.Context, address string) (vtgateconn.Impl, error) {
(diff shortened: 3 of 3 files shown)


dbussink:
@mdlayher Looks like there's a legit failure here:

```
4
Error: go/test/endtoend/vtgate/grpc_server_auth_static/main_test.go:199:44: cannot use ctx (variable of type "context".Context) as grpc.DialOption value in argument to grpcvtgateconn.DialWithOpts: "context".Context does not implement grpc.DialOption (missing method apply) (typecheck)
[5](https://github.com/vitessio/vitess/actions/runs/4069074021/jobs/7014865043#step:22:6)
	dialerFunc := grpcvtgateconn.DialWithOpts(ctx, creds)
[6](https://github.com/vitessio/vitess/actions/runs/4069074021/jobs/7014865043#step:22:7)
	                                          ^
[7](https://github.com/vitessio/vitess/actions/runs/4069074021/jobs/7014865043#step:22:8)
Error: Process completed with exit code 1.
```

dbussink:
> @mdlayher Looks like there's a legit failure here:

Pushed up a fix for this one.

GuptaManan100 (approved):

GuptaManan100 (commented):
Is it okay to change the function signature of a grpcvtgateconn exported function? Do we promise compatibilities here? @harshit-gangal @systay

mdlayher:
Yep I wondered the same thing, no rush!

dbussink:
> Is it okay to change the function signature of a grpcvtgateconn exported function? Do we promise compatibilities here? @harshit-gangal @systay

I did a code search for usages of this and it seems almost all forks and nothing really we'd break there I could find in a quick search. That doesn't cover of course code we don't know about.

I think though that since this would be in Vitess 16, it's not entirely unexpected to have things change slightly then. Also the change itself is simple, they can remove an argument and don't have to add anything for it so the update is simple as well. All those to me at least speak for changing it here. And arguably the old signature is exactly what was part of causing the problem anyway.

harshit-gangal:
For Backward Compatibility, we should not change the signature of `DialWithOpts` and add a new one `DialContextWithOpts`

harshit-gangal:
Also, you should add it in the release note to use this new function to connect than older one as you highlighted the issues in the bug report.

dbussink:
> For Backward Compatibility, we should not change the signature of `DialWithOpts` and add a new one `DialContextWithOpts`

It's actually the other way around. We have `DialWithOpts` which today requires a context, but that context isn't useful / used at all anymore. Not sure if there's a good standard naming convention for that then?

mdlayher:
The problem is that the current context parameter is of no use and I don't see a clear path forward to make it useful. We could document that it has no effect and deprecate the current function.

Given that the current API of the package is a single exported function https://pkg.go.dev/vitess.io/vitess/go/vt/vtgate/grpcvtgateconn (which makes the WithOpts suffix even more odd), I suggest:

```go
// Dial produces a vtgateconn.DialerFunc with custom options.
func Dial(opts ...grpc.DialOption) vtgateconn.DialerFunc
```

```go
// DialWithOpts allows for custom dial options to be set on a vtgateConn.
//
// Deprecated: the context parameter cannot be used by the returned
// vtgateconn.DialerFunc and thus has no effect. Use Dial instead.
func DialWithOpts(ctx context.Context, opts ...grpc.DialOption) vtgateconn.DialerFunc
```

Thoughts?

harshit-gangal:
> The problem is that the current context parameter is of no use and I don't see a clear path forward to make it useful. We could document that it has no effect and deprecate the current function.
> 
> Given that the current API of the package is a single exported function https://pkg.go.dev/vitess.io/vitess/go/vt/vtgate/grpcvtgateconn (which makes the WithOpts suffix even more odd), I suggest:
> 
> ```go
> // Dial produces a vtgateconn.DialerFunc with custom options.
> func Dial(opts ...grpc.DialOption) vtgateconn.DialerFunc
> ```
> 
> ```go
> // DialWithOpts allows for custom dial options to be set on a vtgateConn.
> //
> // Deprecated: the context parameter cannot be used by the returned
> // vtgateconn.DialerFunc and thus has no effect. Use Dial instead.
> func DialWithOpts(ctx context.Context, opts ...grpc.DialOption) vtgateconn.DialerFunc
> ```
> 
> Thoughts?

We should mark it as deprecated and add the reason why it should not be used. We can instead do Dial as you suggested.

mdlayher:
@harshit-gangal @dbussink updated, PTAL.

harshit-gangal:
There are CI failure due to this change. You would be needed to look at them and fix them.

mdlayher:
Lots of Azure apt package failures. Rebasing and trying again to see if there are any true failures due to this change.

harshit-gangal (approved):

dbussink (approved):

Issue #12208, opened 2023-02-01
Bug Report: Online DDL migration resubmission deadlock


shlomi-noach:
### Overview of the Issue

https://github.com/vitessio/vitess/pull/12169 introduced a deadlock scenario, found during internal testing and identified by @rbranson:

- the function `submitCallbackIfNonConflicting` acquires `e.migrationMutex`
- in case this is a resubmission, we call `submitCallbackIfNonConflicting` with `e.RetryMigration(...)` as callback
- `e.RetryMigration` itself tries to acquire same lock.

Fix incoming.

### Reproduction Steps

Resubmit same migration without `--singleton` nor `--singleton-context` ddl strategy flags.

### Binary Version

```sh
-
```

### Operating System and Environment details

```sh
-
```

PR #12211, opened 2023-02-01
Online DDL: fix deadlock by releasing mutex before calling callback

shlomi-noach:
## Description

Fixes https://github.com/vitessio/vitess/issues/12208

I mean, the code which https://github.com/vitessio/vitess/pull/12169 replaced, strictly said:

> // We wrap everything in a function because we want the mutex released. We will need to reaquire it later on
> // for other bookkeeping work.

So here we are. This PR re-introduces a similar function wrap, to release the mutex before running the callback.

## Related Issue(s)

- https://github.com/vitessio/vitess/issues/12208
- #6926

## Checklist

### DIFF
--- a/go/test/endtoend/onlineddl/scheduler/onlineddl_scheduler_test.go
+++ b/go/test/endtoend/onlineddl/scheduler/onlineddl_scheduler_test.go
@@ -803,6 +803,43 @@ func testScheduler(t *testing.T) {
 			}
 		})
 	})
+
+	t.Run("Idempotent submission, retry failed migration in singleton context", func(t *testing.T) {
+		uuid := "00000000_1111_3333_3333_444444444444"
+		ddlStrategy := ddlStrategy + " --singleton-context"
+		overrideVtctlParams = &cluster.VtctlClientParams{DDLStrategy: ddlStrategy, SkipPreflight: true, UUIDList: uuid, MigrationContext: "idempotent:1111-3333-3333"}
+		defer func() { overrideVtctlParams = nil }()
+		// create a migration and cancel it. We don't let it complete. We want it in "failed" state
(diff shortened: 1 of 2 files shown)


shlomi-noach:
The gitdiff does injustice to the change. Basically a new anonymous `func() error` wraps a big chunk of `submitCallbackIfNonConflicting()`, shifting right most of the text; on top of it, which is what confuses `git`, the function only returns an `error` type, so all `return` point look slightly different.

mattlord (approved):
I know that I'm probably annoying you to death regarding unit tests... but this may have been caught by the unit test race workflow (don't know for sure, it's just another benefit to having them).

shlomi-noach:
@mattlord I came here to say that you explicitly asked me about test coverage.

shlomi-noach:
Notes on testing:

- Looking at the `unit_race` test, there _was_ a failure, but unrelated: https://github.com/vitessio/vitess/actions/runs/4069075880/jobs/7008400822 (I promise I did not just hit the re-run before looking at the output)
- I was planning to add a "resubmit" test to test the deadlock -- and as it happens there already is one: https://github.com/vitessio/vitess/blob/4b3b37d41ec08130d162b5115904108cbb3dc14b/go/test/endtoend/onlineddl/scheduler/onlineddl_scheduler_test.go#L772-L805 ; so evidently that test is not enough and I need to think of a stronger test.

shlomi-noach:
All right, it needs to be a resubmission of a `failed` or `cancelled` migration, _and_ in either `--singleton` or `--singleton-context` strategy.

shlomi-noach:
Added a test that validates lack of deadlock. The same test fails on current `main` code, on deadlock.

dbussink (approved):

PR #12214, opened 2023-02-02
Move vtorc from go-sqlite3 to modernc.org/sqlite

dbussink:
This moves `vtorc` from the `go-sqlite3` library that uses CGO, to use `modernc.org/sqlite` which is a pure Go implementation.

`vtorc` is the only component we have to build with `CGO` but it's causing pain for releases since we need to build it against an old Linux for linking against `glibc`.

Using `modernc.org/sqlite` allows for using Go only again and makes all Vitess components buildable without `CGO`.

In
https://datastation.multiprocess.io/blog/2022-05-12-sqlite-in-go-with-and-without-cgo.html someone ran some basic benchmarks. It shows that the pure Go version can be twice as slow, but the usage of `vtorc` is very limited and we operate on small datasets, so I think the performance impact purely of a somewhat slower `sqlite` implementation is negligable.

None of this is in a hot query serving path or anything like that, so I have little concern performance wise.

## Related Issue(s)

  - Fixes: https://github.com/vitessio/vitess/issues/12185 
      - This provides a fundamental fix by removing the entire problem so we don't have to worry about how we build `vtorc` since we don't need `CGO`. 

## Checklist


GuptaManan100 (approved):
LGTM!

dbussink:
Manual backport in https://github.com/vitessio/vitess/pull/12223

Issue #12216, opened 2023-02-02
Handling CodeQL reports


dbussink:
### Overview of the Issue

We have CodeQL scanning enabled and there are a number of things it reports. Nothing seems critical or immediately exploitable, but it does validly call out some shaky handling of casting and other things. 

This is a tracking issue for these issues to handle them.

- [ ] Handling of integer casts in various places
- [ ] Self XSS around the throttler with invalid throttlers
- [ ] Incomplete regular expression for hostnames
- [ ] Log entries created from user input

### Reproduction Steps

-

### Binary Version

```sh
-
```

### Operating System and Environment details

```sh
-
```

### Log Fragments

```sh
-
```

PR #12217, opened 2023-02-02
Fix typecasting issue for workflow types

dbussink:
This addresses a number of the issues identified by CodeQL around typecasting. We use 32 bit integers for these, so let's make sure we parse and cast then as those everywhere it matters.

## Related Issue(s)

Part of https://github.com/vitessio/vitess/issues/12216

## Checklist

### DIFF
--- a/go/vt/wrangler/materializer.go
+++ b/go/vt/wrangler/materializer.go
@@ -1309,9 +1309,19 @@ func (mz *materializer) generateInserts(ctx context.Context, targetShard *topo.S
 		if mz.isPartial {
 			workflowSubType = binlogdatapb.VReplicationWorkflowSubType_Partial
 		}
+		var workflowType binlogdatapb.VReplicationWorkflowType
+		switch mz.ms.MaterializationIntent {
+		case vtctldatapb.MaterializationIntent_CUSTOM:
+			workflowType = binlogdatapb.VReplicationWorkflowType_Materialize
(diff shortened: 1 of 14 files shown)


rohit-nayak-ps (approved):
lgtm. thanks!

go/vt/wrangler/materializer.go
```diff
@@ -1309,9 +1309,19 @@ func (mz *materializer) generateInserts(ctx context.Context, targetShard *topo.S
...
+		case vtctldatapb.MaterializationIntent_CUSTOM:
+			workflowType = binlogdatapb.VReplicationWorkflowType_Materialize
+		case vtctldatapb.MaterializationIntent_MOVETABLES:
+			workflowType = binlogdatapb.VReplicationWorkflowType_MoveTables
+		case vtctldatapb.MaterializationIntent_CREATELOOKUPINDEX:
+			workflowType = binlogdatapb.VReplicationWorkflowType_CreateLookupIndex
+		}
+
 		ig.AddRow(mz.ms.Workflow, bls, "", mz.ms.Cell, mz.ms.TabletTypes,
-			int64(mz.ms.MaterializationIntent),
```
dbussink, line 1313: Replaced this with the explicit switch, since depending on enum order is really an implementation detail we should not (ab)use.

shlomi-noach (approved):

PR #12218, opened 2023-02-02
Use consistent type for vreplication id

dbussink:
The `_vt.vreplication` table has an `INT` type without `UNSIGNED`, so it maps to the Go `int32` type to have the exact same consistent range.

This means that we should use `int32` consistently for everywhere that we use the VReplication identifier.

We had int32, uint32 and int used pretty randomly for these which leads to CodeQL flagging these and how this was found.

Instead this now converts everything to int32. It's also safe to change the protobuf, since the protobuf wire format for int32, uint32, int64 & uint64 is all compatible and actually uses the same variable int length format.

The only exception would be if someone has a VReplication identifier that would not fit in `int32`, but well, this PR shows that in that case we already blow up in unexpected and unknown ways anyway. So it's better to have that explicitly error.

## Related Issue(s)

Part of https://github.com/vitessio/vitess/issues/12216

## Checklist


mattlord:
Thank you for these! I had discussed this before with @rohit-nayak-ps but we didn't have Int32 type support in the sqltypes package and I didn't hear back on that from the QueryServing team when I asked about that so it feel to the wayside.

dbussink:
> Thank you for these! I had discussed this before with @rohit-nayak-ps but we didn't have Int32 type support in the sqltypes package and I didn't hear back on that from the QueryServing team when I asked about that so it feel to the wayside.

Ah yeah, we actually do support `int32` in a bunch of places, just not all yet. I've added this in https://github.com/vitessio/vitess/pull/12218/commits/396d4691ef965a8f7a3e9a3d1852d6af4004a362 and it's a pretty straightforward addition, so that's good :smile:.

mattlord (approved):
My pedantic self loves it. ❤️ The only thing I'm unsure about are the protobuf changes... they seem like potentially breaking changes, BUT I think they're all internal-only so should be fine. Could cause upgrade/downgrade issues though potentially. @deepthi thoughts?

dbussink:
> My pedantic self loves it. ❤️ The only thing I'm unsure about are the protobuf changes... they seem like potentially breaking changes, BUT I think they're all internal-only so should be fine. Could cause upgrade/downgrade issues though potentially. @deepthi thoughts?

> Instead this now converts everything to int32. It's also safe to change the protobuf, since the protobuf wire format for int32, uint32, int64 & uint64 is all compatible and actually uses the same variable int length format.

@mattlord This is something I mentioned in the PR body, it's actually compatible :smile:. The only time it's not, is when it overflows the lower type, but as mentioned there as well, we have a different problem then :smile:.

mattlord:
> > My pedantic self loves it. ❤️ The only thing I'm unsure about are the protobuf changes... they seem like potentially breaking changes, BUT I think they're all internal-only so should be fine. Could cause upgrade/downgrade issues though potentially. @deepthi thoughts?
> 
> > Instead this now converts everything to int32. It's also safe to change the protobuf, since the protobuf wire format for int32, uint32, int64 & uint64 is all compatible and actually uses the same variable int length format.
> 
> @mattlord This is something I mentioned in the PR body, it's actually compatible 😄. The only time it's not, is when it overflows the lower type, but as mentioned there as well, we have a different problem then 😄.

👍 Let's merge this in v17 (main becomes 17-snapshot soon) and then we have plenty of time to catch anything unexpected.

go/vt/vttablet/tabletmanager/vreplication/engine.go
```diff
@@ -671,11 +676,15 @@ func (vre *Engine) transitionJournal(je *journalEvent) {
 			return
 		}
 		log.Infof("Created stream: %v for %v", qr.InsertID, sgtid)
-		newids = append(newids, int(qr.InsertID))
+		if qr.InsertID > math.MaxInt32 {
```
shlomi-noach, line 679: nice

shlomi-noach (approved):
This is great and I like how a lot of casting is now unnecessary.

Merely wondering whether we should aim for `int64`? Not expecting workflow sequences to run out of `int32` but then again, is there a particular reason not to use `int64`?

dbussink:
> Merely wondering whether we should aim for int64? Not expecting workflow sequences to run out of int32 but then again, is there a particular reason not to use int64?

Because the underlying table uses an `int` in MySQL, which is 32 bits. If we want to go to `int64`, we also should change the vreplication table to be a `begint`.

shlomi-noach:
>  we also should change the vreplication table to be a bigint.

@dbussink right! And I'm happy to do that. We can do it all in one PR. If you prefer to first merge this with `int32` it's fine and we can increase later.

shlomi-noach:
We stick to `int32` in the meantime.

Issue #12221, opened 2023-02-02
New merge VReplication workflows do not start automatically on multi-primary clusters


rohit-nayak-ps:
### Overview of the Issue

On a multi-primary cluster like Galera, creating a merge workflow, where there are multiple vreplication streams for a single workflow, results in an error and what appears to be a broken workflow which has not started.

The workflow does get created properly and a `Workflow Start` does succeed.

This code, which starts the vreplication controllers for each workflow stream, expects that the `id`s generated are consecutive.

```
for id := int(qr.InsertID); id < int(qr.InsertID)+plan.numInserts; id++ {
			if ct := vre.controllers[id]; ct != nil {
				// Unreachable. Just a failsafe.
				ct.Stop()
				delete(vre.controllers, id)
			}
			params, err := readRow(dbClient, id)
```

### Reproduction Steps

* set `@@auto_increment_increment` to greater than 1, say, 3
* Run the local example steps including Reshard without any teardowns, so that the single-sharded `customer` tablets are still running
* Run a merge workflow, for example, something like `Reshard -- --source_shards '-80,80-' --target_shards '0' Create customer.cust_merge` so that the created workflow has two source streams on the same target shard
* We get an error that looks like 
```
 rpc error: code = Unknown desc = TabletManager.VReplicationExec on xyz error: unexpected number of rows: &{[name:"id" type:INT32 
```

### Binary Version

```sh
$ vtgate --version
Version: 16.0.0-SNAPSHOT (Git revision 335f46c75b70e1b9e0998b8a6b333a6efa963471 branch 'rn-vrep-new-workflow-auto-increment-step') built on Thu Feb  2 18:56:55 CET 2023 by rohit@rohit-ubuntu using go1.19.4 linux/amd64
```

PR #12228, opened 2023-02-03
VReplication: workflows with multiple streams fail to start on multi-primary setups

rohit-nayak-ps:
## Description

New workflows are created by inserting rows in `_vt.vreplication`. These are done using the `VReplicationExec` API. For some types of workflows there could be multiple source streams and hence multiple rows inserted. For example, resharding from 2 shards to 1.The code that handles these inserts is currently using `last_insert_id()` as the id of the first stream and expects the subsequent ones to be an increment of 1.

This is not true for multi-primary setups, see Related Issue below. This PR gets the auto_increment step from the db and expects subsequent ids to have that increment.

We test the new functionality by specifying a different `@@auto_increment_increment` for a selected merge workflow in the vreplication e2e tests. I confirmed that without this logic we fail as reported in the issue.

## Related Issue(s)

  - Fixes: https://github.com/vitessio/vitess/issues/12221

## Checklist


go/test/endtoend/vreplication/vreplication_test.go
```diff
@@ -973,12 +980,19 @@ func reshard(t *testing.T, ksName string, tableName string, workflow string, sou
 				t.Fatal(err)
 			}
 		}
+		tablets := vc.getVttabletsInKeyspace(t, defaultCell, ksName, "primary")
+
+		// Test multi-master setups with auto increment steps > 1.
+		for _, tablet := range tablets {
+			autoIncrementSetQuery := fmt.Sprintf("set auto_increment_increment = %d; set @@global.auto_increment_increment = %d",
```
mattlord, line 987: Nit, but IMO we should be explicit in both cases about the intended scope. i.e.:
```
autoIncrementSetQuery := fmt.Sprintf("set @@session.auto_increment_increment = %d; set @@global.auto_increment_increment = %d",
```

go/vt/vttablet/tabletmanager/vreplication/engine.go
```diff
@@ -393,7 +394,18 @@ func (vre *Engine) exec(query string, runAsAdmin bool) (*sqltypes.Result, error)
 		}
 
 		vdbc := newVDBClient(dbClient, binlogplayer.NewStats())
-		for id := int32(qr.InsertID); id < int32(maxInsert); id++ {
+
+		// If we are creating multiple streams, for example in a merge workflow going from 2 shards to 1 shard,
+		// we will be inserting multiple rows. To get the ids of subsequent streams we need to know what
+		// the auto_increment_increment step is. In a multi-master environment, for example, we will encounter
+		// auto increment steps > 1.
```
mattlord, line 401: This is annoyingly pedantic, I know, but the length of these comment lines is much longer than the standard elsewhere in the file. Suggestion:
```
		// If we are creating multiple streams, for example in a
                // merge workflow going from 2 shards to 1 shard, we
 		// will be inserting multiple rows. To get the ids of
                // subsequent streams we need to know what the
 		// auto_increment_increment step is. In a multi-master
                // environment, for example, we will often encounter
 		// auto_increment steps > 1.
```

go/vt/vttablet/tabletmanager/vreplication/engine.go
```diff
@@ -844,6 +856,22 @@ func (vre *Engine) readAllRows(ctx context.Context) ([]map[string]string, error)
...
+func (vre *Engine) getAutoIncrementStep(dbClient binlogplayer.DBClient) (int32, error) {
+	qr, err := dbClient.ExecuteFetch("select @@auto_increment_increment", 1)
+	if err != nil {
+		return 0, err
+	}
+	if len(qr.Rows) != 1 {
+		// Handles case where underlying mysql doesn't support auto_increment_increment for any reason.
+		return 1, nil
+	}
+	autoIncrement, err := evalengine.ToInt64(qr.Rows[0][0])
```
mattlord, line 868: In light of all the recent CodeQL related work and aligning types, we should match what MySQL uses: https://dev.mysql.com/doc/refman/8.0/en/replication-options-source.html#sysvar_auto_increment_increment

MySQL uses a uint16 for this variable. Doesn't look like we've yet added `ToUint16()`, so we could do that here: https://github.com/vitessio/vitess/blob/main/go/sqltypes/value.go#L319
```
func (v Value) ToUint16() (uint16, error) {
	if !v.IsIntegral() {
		return 0, ErrIncompatibleTypeCast
	}

	u, err := strconv.ParseUint(v.RawStr(), 10, 16)
	return uint16(u), err
}
```

This will @dbussink and I some future work. 😄

go/vt/vttablet/tabletmanager/vreplication/engine_test.go
```diff
@@ -154,6 +154,7 @@ func TestEngineExec(t *testing.T) {
 
 	dbClient.ExpectRequest("use _vt", &sqltypes.Result{}, nil)
 	dbClient.ExpectRequest("insert into _vt.vreplication values(null)", &sqltypes.Result{InsertID: 1}, nil)
+	dbClient.ExpectRequest("select @@auto_increment_increment", &sqltypes.Result{}, nil)
```
mattlord, line 157: Also in these places, I think it's worth being explicit about the desired variable scope (assuming it's @@session).

go/vt/vttablet/tabletmanager/vreplication/engine.go
```diff
@@ -393,7 +394,18 @@ func (vre *Engine) exec(query string, runAsAdmin bool) (*sqltypes.Result, error)
...
+
+		// If we are creating multiple streams, for example in a merge workflow going from 2 shards to 1 shard,
+		// we will be inserting multiple rows. To get the ids of subsequent streams we need to know what
+		// the auto_increment_increment step is. In a multi-master environment, for example, we will encounter
+		// auto increment steps > 1.
+		autoIncrementStep, err := vre.getAutoIncrementStep(dbClient)
+		if err != nil {
+			return nil, err
+		}
+		firstID := int32(qr.InsertID)
```
mattlord, line 406: Confirmed that this is indeed an int32: https://github.com/vitessio/vitess/blob/main/go/vt/sidecardb/schema/vreplication/vreplication.sql#L19

mattlord (commented):
LGTM! I only had some minor nits/comments. Let me know what you think.

go/vt/vttablet/tabletmanager/vreplication/engine.go
dbussink: Yeah, this would be nice to keep things correctly typed and so we throw errors that are correct when there's a type mismatch instead of more silent weird truncation.

rohit-nayak-ps: Fixed

deepthi (approved):

PR #12229, opened 2023-02-03
Use the correct uint32 type for the GTID file position

dbussink:
This is defined as uint32 at the protocol level, so better to use that type consistently and also parse it accordingly with `ParseUint` with the right size.

## Related Issue(s)

Part of #12216 as this was flagged there.

## Checklist


shlomi-noach (approved):

GuptaManan100 (commented):
LGTM!

PR #12236, opened 2023-02-03
Prevent bootstrap of ZK from clearing previous dist binaries

adsr:
## Description

Presently [protoc bootstrap](https://github.com/vitessio/vitess/blob/b09ce8baee19144ec28e6753714a6dae74907d1d/bootstrap.sh#L268) gets wiped out by [ZK bootstrap](https://github.com/vitessio/vitess/blob/b09ce8baee19144ec28e6753714a6dae74907d1d/bootstrap.sh#L273) via [this rm](https://github.com/vitessio/vitess/blob/b09ce8baee19144ec28e6753714a6dae74907d1d/bootstrap.sh#L65) in `install_dep`. The fix in this patch is to invoke `install_dep` for ZK with subdir of `dist/` instead of the top-level `dist/`, similar to the other deps.

## Related Issue(s)

N/A

## Checklist

## Deployment Notes

N/A

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -127,16 +127,14 @@ install_zookeeper() {
   local version="$1"
   local dist="$2"
   zk="zookeeper-$version"
-  vtzk="vt-zookeeper-$version"
   # This is how we'd download directly from source:
   # wget "https://dlcdn.apache.org/zookeeper/$zk/apache-$zk.tar.gz"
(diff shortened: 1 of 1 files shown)


deepthi:
@adsr I'm trusting that you have validated this locally. We'll get another approval for this.

frouioui (approved):

vmg (approved):

PR #12237, opened 2023-02-05
Fix additional integer type handling

dbussink:
The tablet uid can be a uint32 everywhere and we can parse it like that with pflag as well. This way we don't have to cast and change the valid range and get correct messages when it's out of range.

As well we use a regular int for the port which better matches how Go itself tracks ports and removes a bunch of casts from places where we already using int.

Lastly it also fixes some small cases of how we parse to match the type we store it in.

## Related Issue(s)

Part of #12216 

## Checklist


frouioui (approved):

rohit-nayak-ps (approved):
lgtm

PR #12239, opened 2023-02-05
Examples: Add lib functions to wait for shard states

mattlord:
## Description

Before modifying a shard we need to wait for the tablets to be up, a primary to be elected, and that primary to become healthy/serving. We were skipping this last check,  but in practice it was virtually never an issue because the transition to primary after election (by `vtorc` now) was so fast. After https://github.com/vitessio/vitess/pull/11520, however, once the primary has been elected it then needs to initialize its sidecar database. This additional time, while small, now makes the local examples unreliable because we do not wait for that to happen before we e.g. execute the `ApplySchema` vtctl client command. We also did not check for any errors in vtctl client commands like that, so you would not know anything was wrong until trying to execute a later step or command and you would see that e.g. the tables did not exist. In the process, I also made the following notable changes:
>**Note** I took this opportunity to remove an annoying and unnecessary step in the `region_sharding` examples of having to specify a full path to `countries.json` — which meant that the examples did not work w/o taking that manual step — when a relative path works and we're using relative paths everywhere else in the examples.

>**Note** I realized that there was a lot of duplicated files/code between the `local` and `region_examples`, so I took the opportunity to refactor these so that they share the same files between them (this will require a docs PR now too).

## Related Issue(s)

  - Related to: https://github.com/vitessio/vitess/pull/11520

## Checklist

-   [x] Documentation: https://github.com/vitessio/website/pull/1380


frouioui (approved):

GuptaManan100 (approved):
looks good to me apart from the outstanding comments.

examples/local/lib/utils.sh
```diff
@@ -0,0 +1,72 @@
...
+#   wait_for_shard_tablets commerce 0 2
+function wait_for_shard_tablets() {
+	if [[ -z ${1} || -z ${2} || -z ${3} ]]; then
+		fail "A keyspace, shard, and number of tablets must be specified when waiting for tablets to come up"
+	fi
+	keyspace=${1}
+	shard=${2}
+	num_tablets=${3}
+
+	for _ in $(seq 0 200); do
```
rohit-nayak-ps, line 15: Since we also have a tablet id 200, this has always been a bit confusing to me! Since we are refactoring anyway, can we please use a constant here and below to be clear what this range upper value is for?

rohit-nayak-ps (approved):
Great changes, thanks!

examples/local/lib/utils.sh
```diff
@@ -0,0 +1,72 @@
+#!/bin/bash
+
+# Wait for the given number of tablets to show up
+# for the keyspace/shard in the topology server.
+# Example (wait for 2 tables in commerce/0):
+#   wait_for_shard_tablets commerce 0 2
+function wait_for_shard_tablets() {
+	if [[ -z ${1} || -z ${2} || -z ${3} ]]; then
```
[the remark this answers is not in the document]
mattlord: Yeah, I wanted to do that (note that other files, e.g. env.sh, are also the same). We would also have to change the documentation for both examples though as they currently tell users to copy just that local/region_sharding directory away and then run them:
  - https://vitess.io/docs/16.0/get-started/local/#start-a-single-keyspace-cluster
  - https://vitess.io/docs/16.0/user-guides/configuration-advanced/region-sharding/#create-and-start-the-cluster

I don't really like that either... so I could modify the docs as well. Then we could move a number of copied files into a parent directory.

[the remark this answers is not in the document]
mattlord: Done: https://github.com/vitessio/vitess/pull/12239/commits/a7bb876035298e147771ca9b87cddfdb4f60233a

examples/common/scripts/k3s-down.sh
```diff
@@ -16,11 +16,7 @@
 
 # This is an example script that stops the k3s server started by k3s-up.sh.
 
-set -e
```
frouioui, line 19: should we keep `set -e`, i don't see it in `env.sh`

examples/region_sharding/lib/utils.sh
```diff
@@ -0,0 +1,90 @@
+#!/bin/bash
+
```
frouioui, line 2: so should we be able to remove this file now that we have `examples/common/lib/utils.sh`?

frouioui (approved):
This is looking good to me! Thank you for doing this @mattlord, this is awesome work 🙌🏻

examples/common/scripts/k3s-down.sh
mattlord: Good catch! Didn't intend to do that. Will put it back.

examples/region_sharding/lib/utils.sh
mattlord: Doh! Yeah, those should already be gone. Thanks!

PR #12240, opened 2023-02-06
Support Custom SidecarDB Names on VTTablets

mattlord:
## Description

This PR adds a new **experimental** feature: *keyspace level sidecar database names*. For the *why*, please see [the feature request](https://github.com/vitessio/vitess/issues/12304).

In this PR the sidecar db name can only be set when the keyspace is first created via the `CreateKeyspace` vtctl client command using the new `--sidecar-db-name` command flag. From that point forward it is immutable (w/o modifying the topo directly outside of Vitess). If no value is specified for the keyspace then it will be set to the default of `_vt` when the keyspace is created, and as a fallback the default will also be set when the first tablet is started within the keyspace if for any reason it's not already set in the topo keyspace record.

The management of the sidecar database resides entirely within a `vttablet`. This allows each `vttablet` in a keyspace to manage its own sidecar database independently of all other keyspaces and processes in the cluster. This allows e.g. for most tablets within a Vitess cluster to use the default sidecar name of `_vt` while some can use a non-standard name of e.g. `_vt_import` or `__virtual_vitess_cluster` for whatever purposes as needed (in various cases the tablet's `mysqld` instance may be participating in two distinct vitess clusters or two distinct keyspaces). From outside of the `vttablet`, Vitess is able to assume that the standard name of `_vt` is being used (including external/3rd party tooling that e.g. may be doing things like `vtctlclient ExecuteFetchAsDba`) unless talking directly to a tablet's mysqld instance.

>**Note** there are two exceptions to the above rule today — where another process *does* need to know the sidecar database name that a `vttablet` is using (by way of the keyspace it's in). Both are in `vtgate` and are due to the fact that today `vtgate` sends raw SQL queries to a `vttablet` through the normal user query hot path (tabletserver's `QueryExecutor.Execute()`) in the following cases:
> 1. OnlineDDL's `show vitess_migrations` implementation
> 2. VTGate's Schema Tracker implementation
>
> The plan is to eliminate these two exceptions in the future by leveraging new `vttablet` RPCs that `vtgate` can then use instead of the current SQL interface. This way we have proper encapsulation and `vtgate` does not need to be aware of the implementation details (in this case the schema `vttablet` uses for the storage).

## Related Issue(s)

  - Fixes: https://github.com/vitessio/vitess/issues/12304

## Checklist

-   [x] Documentation: https://github.com/vitessio/website/pull/1402


shlomi-noach:
@mattlord will it make sense to have the sidecar db name _predictable_, possibly even as default,m without the need to add `--sidecar-db-name`?

e.g. is my keyspace is `commerce`, the sidecar DB name would be `_vt_commerce` ; in this case the tablet can first automatically search for sidecar db names as it starts; it can auto correlate the keyspace/sidecardb. It may choose to fall back to `_vt` if there's only one key space and no `_vt_<keyspace>` DB, etc.

mattlord:
> @mattlord will it make sense to have the sidecar db name _predictable_, possibly even as default, without the need to add `--sidecar-db-name`?
> 
> e.g. is my keyspace is `commerce`, the sidecar DB name would be `_vt_commerce` ; in this case the tablet can first automatically search for sidecar db names as it starts; it can auto correlate the keyspace/sidecardb. It may choose to fall back to `_vt` if there's only one key space and no `_vt_<keyspace>` DB, etc.

Just FYI, the `--sidecar-db-name` vttablet flag is gone. You now have to specify the non-default sidecar-db-name when creating the keyspace. 

This is an interesting idea and not one that I'm opposed to in any way. For me, if it's still variable then we still need to read it from the topo server and the PR is mostly the same.

shlomi-noach:
> Currently the sidecar db name can only be set

Just to clarify to myself, by "Currently" you mean "in this PR", right? I normally assume that "currently" relates to existing code in `main`.

go/vt/sidecardb/sidecardb.go
```diff
@@ -364,35 +396,44 @@ func (si *schemaInit) ensureSchema(table *sidecarTable) error {
...
 func AddSchemaInitQueries(db *fakesqldb.DB, populateTables bool) {
+	once.Do(loadSchemaDefinitions)
 	result := &sqltypes.Result{}
+	db.AddQuery(fmt.Sprintf("use %s", GetIdentifier()), result)
 	for _, q := range sidecarDBInitQueryPatterns {
 		db.AddQueryPattern(q, result)
 	}
 	for _, q := range sidecarDBInitQueries {
-		db.AddQuery(q, result)
+		db.AddQuery(fmt.Sprintf(q, GetIdentifier()), result)
```
shlomi-noach, line 436: Continued - again I feel like this is not idiomatic; we should use `ParseAndBind` instead of `fmt.Sprintf`

go/vt/sqlparser/utils.go
```diff
@@ -113,3 +113,52 @@ func NormalizeAlphabetically(query string) (normalized string, err error) {
...
+	}, nil)
+	// If we didn't modify anything, return the original query.
+	// This is particularly helpful with unit tests that
+	// execute a query which slightly differs from the parsed
+	// version: e.g. 'where id=1' becomes 'where id = 1'.
+	if modified {
+		return String(upd), nil
+	}
+	return query, nil
+}
```
shlomi-noach, line 164: :+1:  nice

go/vt/vtctl/vtctl.go
```diff
@@ -1832,6 +1833,7 @@ func commandCreateKeyspace(ctx context.Context, wr *wrangler.Wrangler, subFlags
 	baseKeyspace := subFlags.String("base_keyspace", "", "Specifies the base keyspace for a snapshot keyspace")
 	timestampStr := subFlags.String("snapshot_time", "", "Specifies the snapshot time for this keyspace")
 	durabilityPolicy := subFlags.String("durability-policy", "none", "Type of durability to enforce for this keyspace. Default is none. Possible values include 'semi_sync' and others as dictated by registered plugins.")
+	sidecarDBName := subFlags.String("sidecar-db-name", sidecardb.DefaultName, "Vitess sidecar database used for internal metadata.")
```
shlomi-noach, line 1836: We might want to consider the magic value `"auto"` to auto-generate a sidecar db name, e.g. if my keyspace is `commerce` then the auto generated name would be `_vt_commerce`.

go/vt/vtgate/schema/tracker.go
```diff
@@ -259,7 +262,8 @@ func (t *Tracker) updatedTableSchema(th *discovery.TabletHealth) bool {
 		return false
 	}
 	bv := map[string]*querypb.BindVariable{"tableNames": tables}
-	res, err := th.Conn.Execute(t.ctx, th.Target, mysql.FetchUpdatedTables, bv, 0, 0, nil)
+	res, err := th.Conn.Execute(t.ctx, th.Target, fmt.Sprintf(mysql.FetchUpdatedTables, "_vt"),
```
shlomi-noach, line 265: another use of `fmt.Sprintf()` (recurring comment, I'll stop here)

go/vt/vttablet/onlineddl/executor.go
```diff
@@ -76,13 +77,13 @@ var (
...
 var vexecUpdateTemplates = []string{
-	`update _vt.schema_migrations set migration_status='val1' where mysql_schema='val2'`,
-	`update _vt.schema_migrations set migration_status='val1' where migration_uuid='val2' and mysql_schema='val3'`,
-	`update _vt.schema_migrations set migration_status='val1' where migration_uuid='val2' and mysql_schema='val3' and shard='val4'`,
+	fmt.Sprintf(`update %s.schema_migrations set migration_status='val1' where mysql_schema='val2'`, sidecardb.GetIdentifier()),
+	fmt.Sprintf(`update %s.schema_migrations set migration_status='val1' where migration_uuid='val2' and mysql_schema='val3'`, sidecardb.GetIdentifier()),
+	fmt.Sprintf(`update %s.schema_migrations set migration_status='val1' where migration_uuid='val2' and mysql_schema='val3' and shard='val4'`, sidecardb.GetIdentifier()),
 }
 
 var vexecInsertTemplates = []string{
```
shlomi-noach, line 85: you may safely ignore this variable - it is unused and to be removed shortly.

go/vt/vttablet/onlineddl/executor.go
```diff
@@ -3234,7 +3238,7 @@ func (e *Executor) runNextMigration(ctx context.Context) error {
 		if err != nil {
 			return nil, err
 		}
-		r, err := e.execQuery(ctx, sqlSelectReadyMigrations)
+		r, err := e.execQuery(ctx, fmt.Sprintf(sqlSelectReadyMigrations, sidecardb.GetIdentifier()))
```
shlomi-noach, line 3241: can we ensure the query always executes on the correct schema (ie the keyspace's sidecardb) rather than re-inject the schema name in all queries?

shlomi-noach (commented):
Wow, monumental work!

I think a recurring thing is that we now re-generate queries programmaticly in all places (I notice `onlineddl/executor.go` ; I feel this is cumbersome, can we instead ensure the connection uses the correct schema? We should aim to minimize the programmatic manipulations of queries, if possible. Maybe that's easier said than done, I'm not sure.

go/vt/vttablet/tabletserver/tabletserver.go
```diff
@@ -730,6 +731,12 @@ func (tsv *TabletServer) Execute(ctx context.Context, target *querypb.Target, sq
 		return nil, vterrors.New(vtrpcpb.Code_INTERNAL, "[BUG] transactionID and reserveID must match if both are non-zero")
 	}
 
+	// Rewrite any sidecar database usage if needed.
+	sql, err = sqlparser.ReplaceTableQualifiers(sql, sidecardb.DefaultName, sidecardb.GetName())
+	if err != nil {
+		return nil, err
+	}
+
 	return tsv.execute(ctx, target, sql, bindVariables, transactionID, reservedID, nil, options)
```
harshit-gangal, line 740: this will happen for all incoming queries in the query serving path.

proto/topodata.proto
```diff
@@ -304,6 +304,16 @@ message Keyspace {
...
+  // ThrottlerConfig has the configuration for the tablet
+  // server's lag throttler, and applies to the entire
+  // keyspace, across all shards and tablets.
+  ThrottlerConfig throttler_config = 9;
+
+  // SidecarDBName is the name of the Vitess sidecar database
+  // used for various system metadata that is stored in each
+  // tablet's mysqld instance.
+  string sidecar_db_name = 10;
 }
```
harshit-gangal, line 317: should adding `ThrottlerConfig` be part of this PR?

go/vt/vttablet/onlineddl/executor.go
```diff
@@ -284,7 +285,13 @@ func (e *Executor) execQuery(ctx context.Context, query string) (result *sqltype
...
 	}
 	defer conn.Recycle()
-	return conn.Exec(ctx, query, math.MaxInt32, true)
+
+	// Replace any provided sidecar DB qualifiers with the correct one.
+	uq, err := sqlparser.ReplaceTableQualifiers(query, sidecardb.DefaultName, sidecardb.GetName())
+	if err != nil {
+		return nil, err
+	}
+	return conn.Exec(ctx, uq, math.MaxInt32, true)
```
shlomi-noach, line 294: This is a quick and powerful way to address the qualifier! I love it.

rohit-nayak-ps:
What about defining an interface with the required query Exec function signatures? We implement two concretes and at runtime select the correct one: 
1. sidecardb==`_vt`: will be a passthrough to the same function called today
1. different sidecardb name: this will first `ReplaceTableQualifiers()`

This way we don't incur any performance penalty for those using the recommended `_vt`. 

If `ReplaceTableQualifiers()` introduces unacceptable delays in the any path, we could see if introducing a cache where we store the original query and replaced query in a map helps

mattlord:
> What about defining an interface with the required query Exec function signatures? We implement two concretes and at runtime select the correct one:
> 
> 1. sidecardb==`_vt`: will be a passthrough to the same function called today
> 2. different sidecardb name: this will first `ReplaceTableQualifiers()`
> 
> This way we don't incur any performance penalty for those using the recommended `_vt`.
> 
> If `ReplaceTableQualifiers()` introduces unacceptable delays in the any path, we could see if introducing a cache where we store the original query and replaced query in a map helps

Good idea! I implemented this: https://github.com/vitessio/vitess/pull/12240/commits/c8cd425b64601a4a34a7d4e1816529d44ecd2735

Did you have something else in mind?

go/vt/sidecardb/sidecardb.go
```diff
@@ -103,7 +109,9 @@ func validateSchemaDefinition(name, schema string) (string, error) {
 	return normalizedSchema, nil
 }
 
-func initSchemaFiles() {
+// loadSchemaDefinitions loads the embedded schema definitions
+// into a slice of sidecarTables for processing.
+func loadSchemaDefinitions() {
 	sqlFileExtension := ".sql"
 	err := fs.WalkDir(schemaLocation, ".", func(path string, entry fs.DirEntry, err error) error {
```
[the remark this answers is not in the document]
mattlord: I don't disagree but it's also not related to this PR and I want to start limiting changes/refactoring I do that's not really part of this feature.

arthurschreiber:
Oh, this is super interesting to see! I think in `vtcombo`, this will allow us to have separate, per keyspace sidecar databases, instead of a single sidecar database that's incorrectly shared by all keyspaces. ❤️

go/vt/sqlparser/utils_test.go
```diff
@@ -180,3 +181,85 @@ func TestQueryMatchesTemplates(t *testing.T) {
...
+		{
+			name: "simple delete",
+			in:   "delete from _vt.foo where id = 1",
+			out:  "delete from foo where id = 1",
+		},
+		{
+			name: "simple set",
+			in:   "set names 'binary'",
+			out:  "set names 'binary'",
+		},
```
shlomi-noach, line 252: Can you please add tests with:
- a derived table 
- an inner select from another table

go/vt/vttablet/tabletserver/throttle/throttler.go
```diff
@@ -401,6 +397,14 @@ func (throttler *Throttler) Open() error {
...
 	}
 	ctx := context.Background()
+	// The query needs to be dynamically built because the sidecar database name
+	// is not known when the TabletServer is created, which in turn creates the
+	// Throttler.
+	throttler.metricsQuery.Store(sqlparser.BuildParsedQuery(defaultReplicationLagQuery, sidecardb.GetIdentifier()).Query) // default
+	if throttleMetricQuery != "" {
+		throttler.metricsQuery.Store(throttleMetricQuery) // override
+	}
+	throttler.initConfig()
```
shlomi-noach, line 407: This looks valid :+1:

go/vt/vttablet/onlineddl/executor.go
```diff
@@ -187,11 +188,14 @@ type Executor struct {
 	tickReentranceFlag            int64
 	reviewedRunningMigrationsFlag bool
 
-	ticks             *timer.Timer
-	isOpen            int64
-	schemaInitialized bool
```
shlomi-noach, line 192: :+1:

go/vt/vttablet/onlineddl/executor.go
```diff
@@ -187,11 +188,14 @@ type Executor struct {
...
+	ticks  *timer.Timer
+	isOpen int64
 
-	initVreplicationDDLOnce sync.Once
```
shlomi-noach, line 194: :+1:

go/vt/vttablet/tabletmanager/tm_init.go
```diff
@@ -486,6 +487,34 @@ func (tm *TabletManager) createKeyspaceShard(ctx context.Context) (*topo.ShardIn
...
+		ctx, unlock, lockErr := tm.TopoServer.LockKeyspace(ctx, tablet.Keyspace, "Setting sidecar database name")
+		if lockErr != nil {
+			return nil, vterrors.Wrap(lockErr, "createKeyspaceShard: cannot GetOrCreateShard shard")
+		}
+		err = tm.TopoServer.UpdateKeyspace(ctx, ks)
+		if err != nil {
+			unlock(&lockErr)
+			return nil, vterrors.Wrap(err, "createKeyspaceShard: cannot GetOrCreateShard shard")
+		}
+		unlock(&lockErr)
```
shlomi-noach, line 510: looks like `unlock(&lockErr)` gets called no matter what. Perhaps write once, right after `UpdateKeyspace()`?

go/vt/vttablet/tabletmanager/vreplication/controller_plan.go
```diff
@@ -228,12 +254,23 @@ func buildDeletePlan(del *sqlparser.Delete) (*controllerPlan, error) {
 }
 
 func buildSelectPlan(sel *sqlparser.Select) (*controllerPlan, error) {
-	switch sqlparser.ToString(sel.From) {
+	tableExpr, ok := sel.From[0].(*sqlparser.AliasedTableExpr)
```
shlomi-noach, line 257: I'm just thinking, could there be a scenario where `From` is an empty slice? e.g. `FROM DUAL` gets truncated? I know we won't be running `FROM DUAL` in `vreplication`; but I'd feel safer if we tested for the length of the slice.

shlomi-noach (approved):
This looks good, with a few comments left inline. I've reviewed throttler, online ddl, vreplication and general other stuff, but I definitely did not review query-serving related changes.

I know you've been working long on this PR and hope you can merge it soon!

go/vt/sqlparser/utils_test.go
mattlord: Done!

go/vt/vttablet/tabletmanager/tm_init.go
mattlord: Good catch. Cleaned up.

go/vt/vttablet/tabletmanager/vreplication/controller_plan.go
mattlord: Nice catch! Yes, will add some protection to this and any similar cases in these build functions.

go/vt/sidecardb/identifier_cache.go
```diff
@@ -0,0 +1,121 @@
...
+		}
+	}
+	return instance
+}
+
+// NewTestIdentifierCache returns an initialized cache struct. This is
+// NOT a singleton and allows you to override the global instance used
+// outside of unit tests.
+// NOTE: This should ONLY be used in unit tests and NOT in production!
+func NewTestIdentifierCache(loadFunc func(context.Context, string) (string, error)) *IdentifierCache {
```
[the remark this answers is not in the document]
mattlord: I tried, but it doesn't work. Looks like go will not let you overwrite a pointer to a non-exported variable. I like that, actually, as it enforces the singleton pattern here.

I added a Destroy() function instead though as that could potentially come in handy down the line.

mattlord:
> Great work on this!
> 
> Looks good. I left a couple of optional comments.

Thank you for the review and helpful feedback! I've implemented changes based on your comments here: https://github.com/vitessio/vitess/pull/12240/commits/a2d378a15f95738808d666f79d917a46a9f9eb64

go/test/endtoend/vtgate/schematracker/sharded_prs/st_sharded_test.go
```diff
@@ -156,7 +172,7 @@ func TestMain(m *testing.M) {
 			return 1
 		}
 
-		err = waitForVTGateAndVTTablet()
+		err = clusterInstance.WaitForVTGateAndVTTablets(5 * time.Minute)
```
harshit-gangal, line 175: this looks like too much time to wait to fail the CI run eventually.

go/vt/sidecardb/identifier_cache.go
```diff
@@ -0,0 +1,119 @@
...
+
+	vtrpcpb "vitess.io/vitess/go/vt/proto/vtrpc"
+	"vitess.io/vitess/go/vt/sqlparser"
+	"vitess.io/vitess/go/vt/vterrors"
+)
+
+// This provides a read through cache of sidecar database identifiers;
+// loading the values from an opaque backend database using a provided
+// load function.
+type IdentifierCache struct {
```
harshit-gangal, line 33: nit: go comment to start with the struct name.

examples/local/101_initial_cluster.sh
```diff
@@ -19,6 +19,15 @@
...
 
+# This is done here as a means to test the experimental custom
+# sidecar database name work in a wide variety of scenarios as
+# the local examples are used to test many features locally.
+# This is NOT here to indicate that you should normally use a
+# non-default (_vt) value or that it is somehow a best practice
+# to do so. In production, you should ONLY use a non-default
+# sidecar database name when it's truly needed.
+SIDECAR_DB_NAME=${SIDECAR_DB_NAME:-"__vitess_metadata"}
+
```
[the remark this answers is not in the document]
mattlord: OK, I can change the default to `_vt` then and keep the ability to set it via the ENV variable.

go/mysql/endtoend/schema_change_test.go
```diff
@@ -84,19 +85,23 @@ func TestChangeSchemaIsNoticed(t *testing.T) {
 
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
+			clearQuery := sqlparser.BuildParsedQuery(mysql.ClearSchemaCopy, sidecardb.GetIdentifier()).Query
+			insertQuery := sqlparser.BuildParsedQuery(mysql.InsertIntoSchemaCopy, sidecardb.GetIdentifier()).Query
+			detectQuery := sqlparser.BuildParsedQuery(mysql.DetectSchemaChange, sidecardb.GetIdentifier()).Query
```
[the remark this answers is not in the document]
mattlord: Good point. I'll update that.

go/test/endtoend/vtgate/schematracker/sharded_prs/st_sharded_test.go
mattlord: I didn't come up with that time, I merely migrated the code to a utility function. That time came from here: https://github.com/vitessio/vitess/blame/c89fa57b6330a4a878aea82350f8525b50ad4b77/go/test/endtoend/vtgate/schematracker/sharded_prs/st_sharded_test.go#L196

go/vt/sidecardb/identifier_cache.go
mattlord: OK, will update.

go/vt/sidecardb/identifier_cache.go
```diff
@@ -0,0 +1,119 @@
...
+	sidecarDBIdentifiers sync.Map
+
+	// The callback used to load the values from the database into
+	// the cache.
+	load func(context.Context, string) (string, error)
+}
+
+const (
+	ErrIdentifierCacheUninitialized  = "sidecar database identifier cache is not initialized"
+	ErrIdentifierCacheNoLoadFunction = "the load from database function has not been set"
```
harshit-gangal, line 48: looks like these are used locally and do not need to be public.

go/vt/sidecardb/identifier_cache.go
```diff
@@ -0,0 +1,119 @@
...
+	return identifierCache.Load().(*IdentifierCache), created
+}
+
+func GetIdentifierCache() (*IdentifierCache, error) {
+	if identifierCache.Load() == nil {
+		return nil, vterrors.New(vtrpcpb.Code_INTERNAL, ErrIdentifierCacheUninitialized)
+	} else {
+		return identifierCache.Load().(*IdentifierCache), nil
+	}
+}
```
harshit-gangal, line 73: It would be easier at other places if we can merge these 2 methods.

go/vt/sqlparser/utils.go
```diff
@@ -113,3 +113,52 @@ func NormalizeAlphabetically(query string) (normalized string, err error) {
...
+	newQualifier := NewIdentifierCS(newdb)
+
+	modified := false
+	upd := Rewrite(in, func(cursor *Cursor) bool {
+		switch node := cursor.Node().(type) {
+		case TableName:
+			if !node.Qualifier.IsEmpty() &&
+				node.Qualifier.String() == oldQualifier.String() &&
+				node.Qualifier.String() != newQualifier.String() {
+				node.Qualifier = newQualifier
```
harshit-gangal, line 141: In the start of the function, it is already checked that olddb and newdb are not same. So, `node.Qualifier.String() != newQualifier.String()` check is not needed.

go/vt/sqlparser/utils.go
```diff
@@ -113,3 +113,52 @@ func NormalizeAlphabetically(query string) (normalized string, err error) {
...
+				cursor.Replace(node)
+				modified = true
+			}
+		case *ShowBasic: // for things like 'show tables from _vt'
+			if !node.DbName.IsEmpty() &&
+				node.DbName.String() == oldQualifier.String() &&
+				node.DbName.String() != newQualifier.String() {
+				node.DbName = newQualifier
```
harshit-gangal, line 149: similarly here

go/vt/vttablet/tabletserver/vstreamer/vstreamer.go
```diff
@@ -643,7 +644,8 @@ func (vs *vstreamer) buildJournalPlan(id uint64, tm *mysql.TableMap) error {
 		return err
 	}
 	defer conn.Close()
-	qr, err := conn.ExecuteFetch("select * from _vt.resharding_journal where 1 != 1", 1, true)
+	qr, err := conn.ExecuteFetch(fmt.Sprintf("select * from %s.resharding_journal where 1 != 1",
+		sidecardb.GetIdentifier()), 1, true)
```
harshit-gangal, line 648: should this use `sqlparser.BuildParsedQuery`

go/vt/vttablet/tabletserver/vstreamer/rowstreamer.go
```diff
@@ -119,7 +119,7 @@ func (rs *rowStreamer) Stream() error {
 		return err
 	}
 	defer conn.Close()
-	if _, err := conn.ExecuteFetch("set names binary", 1, false); err != nil {
+	if _, err := conn.ExecuteFetch("set names 'binary'", 1, false); err != nil {
```
harshit-gangal, line 122: why is this change needed?

go/vt/vttablet/tabletserver/schema/engine.go
```diff
@@ -141,7 +143,7 @@ func (se *Engine) syncSidecarDB(ctx context.Context, conn *dbconnpool.DBConnecti
 
 	var exec sidecardb.Exec = func(ctx context.Context, query string, maxRows int, useDB bool) (*sqltypes.Result, error) {
 		if useDB {
-			_, err := conn.ExecuteFetch(sidecardb.UseSidecarDatabaseQuery, maxRows, false)
+			_, err := conn.ExecuteFetch(fmt.Sprintf("use %s", sidecardb.GetIdentifier()), maxRows, false)
 			if err != nil {
```
harshit-gangal, line 147: this might use `sqlparser.BuildParsedQuery`

go/vt/vttablet/tabletmanager/tm_init.go
```diff
@@ -486,6 +487,33 @@ func (tm *TabletManager) createKeyspaceShard(ctx context.Context) (*topo.ShardIn
...
+			return nil, vterrors.Wrap(lockErr, "createKeyspaceShard: cannot GetOrCreateShard shard")
+		}
+		err = tm.TopoServer.UpdateKeyspace(ctx, ks)
+		unlock(&lockErr)
+		if err != nil {
+			return nil, vterrors.Wrap(err, "createKeyspaceShard: cannot GetOrCreateShard shard")
+		}
+		if lockErr != nil {
+			return nil, vterrors.Wrap(lockErr, "createKeyspaceShard: cannot GetOrCreateShard shard")
+		}
```
harshit-gangal, line 512: if there is lock error we will not bring up the tablet.
lock error can be multiple tablets trying to update the same thing here and lock timeout.
I think with lock timeout we should add a retry.


what I am trying to say here is. Some thoughts should be when multiple tablets come at the same time, will this cause any issue or not

go/vt/vtgate/planbuilder/show.go
```diff
@@ -265,7 +268,17 @@ func buildShowVMigrationsPlan(show *sqlparser.ShowBasic, vschema plancontext.VSc
...
-	sql := "SELECT * FROM _vt.schema_migrations"
+	sdbidc, err := sidecardb.GetIdentifierCache()
+	if err != nil {
+		return nil, err
+	}
+	sidecarDBID, err := sdbidc.Get(ks.Name)
+	if err != nil {
+		return nil, vterrors.Wrapf(err, "failed to read sidecar database identifier for keyspace %q from the cache",
+			ks.Name)
+	}
```
harshit-gangal, line 279: in vtgate planner or executor, we through errors with an error code
example: `vterrors.VT13001`

go/vt/vtgate/planbuilder/show.go
```diff
@@ -265,7 +268,17 @@ func buildShowVMigrationsPlan(show *sqlparser.ShowBasic, vschema plancontext.VSc
...
+
+	sql := fmt.Sprintf("SELECT * FROM %s.schema_migrations", sidecarDBID)
```
harshit-gangal, line 281: should use the sqlparser method here as well

harshit-gangal (commented):
some nits, mostly looks good to me.

go/vt/sidecardb/identifier_cache.go
mattlord: Merge what two methods? GetIdentifierCache() and Get()? I don't think I like that, although it would be easier for the caller in some cases (specifically where all you want to do is Get a single value) -- in general I think it makes it a little awkward. I think you're right though in that there's no reason NOT to add a convenience function for those specific cases:
```
// GetIdentifierForKeyspace is a convenience function that returns
// the sidecar database identifier for the provided keyspace.
func GetIdentifierForKeyspace(keyspace string) (string, error) {
	cache, err := GetIdentifierCache()
	if err != nil {
		return "", err
	}
	return cache.Get(keyspace)
}
```

I will do that.

go/vt/vttablet/tabletserver/vstreamer/vstreamer.go
mattlord: Good catch! I somehow missed this one.

go/vt/vttablet/tabletserver/vstreamer/rowstreamer.go
harshit-gangal: then, its clearly a parsing bug. :(

go/vt/vttablet/tabletmanager/tm_init.go
mattlord: If you have a non-serving keyspace, a keyspace with no tablets, and then you simultaneously bring up 10 tablets let's say... it's possible. I'll work on something for this.

go/vt/sidecardb/identifier_cache.go
harshit-gangal: My thinking was when someone calls into `GetIdentifierCache` then `IdentifierCache` gets created if not already done.

harshit-gangal: This way we do not have to take care of the places where they have to be initialized first before using it.

mattlord:
> some nits, mostly looks good to me.

Thanks for the helpful comments @harshit-gangal ! I believe that I've addressed all of them here: https://github.com/vitessio/vitess/pull/12240/commits/da4ee794ee600368882c138b10ba371233488c64

go/vt/sidecardb/identifier_cache.go
mattlord: This doesn't work, as the caller needs to provide the load function (which in practice usually means they need to have a topo server handy).

go/vt/vttablet/tabletserver/vstreamer/rowstreamer.go
mattlord: That may be, but the "correct" syntax is to use quotes anyway so I did not find it relevant to my PR and instead used the quotes.

https://dev.mysql.com/doc/refman/8.0/en/set-names.html

go/vt/sidecardb/identifier_cache.go
mattlord: The way it is now, the user does not need to know anything about the implementation used (only the creator does). I like this abstraction and IMO it simplifies things for the user.

harshit-gangal (approved):

PR #12246, opened 2023-02-06
onlineddl_vrepl suite: fix auto_increment flakyness

shlomi-noach:
## Description

This PR fixes a recurring CI error with `onlineddl_vrepl_suite` (not to be confused with another flakyness we have in `onlineddl_vrepl`). The flakyness fixed here has to do with `AUTO_INCREMENT` values, and which depends on the MySQL version. I simplified the tests by just saying something like "here's a starting point at `auto_increment=8`, gonna throw a bunch of stuff at the table, delete or whatnot, and expect resulting `auto_increment` to be two digit. That's it. Previously, we tried to hit the exact resulting value. But the value does not matter. What matters is that e.g. when rows are deleted, and table is duplicated, we don't fall back to `auto_increment=8`.

## Related Issue(s)

#6926

## Checklist


dbussink (approved):

rohit-nayak-ps (approved):
lgtm

PR #12247, opened 2023-02-06
evalengine: new evaluation framework

vmg:
## Description

The `evalengine` in Vitess was bitrotting because I didn't get a chance to maintain it these past few months. Back at it. I've learned a lot of things about MySQL evaluation while working on PlanetScale Boost, which makes extensive use of Vitess' evaluation engine, so it's time to put them in practice by improving it!

### Removed lazy evaluation

First and foremost: the way we were doing evaluation was not correct! I implemented a very complex evaluation system for Vitess where the `EvalResult` was eagerly initialized but _lazily_ evaluated. This added _a lot of complexity_, but the visible effect is that the errors for bad expressions in our evaluation engine matched MySQL's quite closely. This was, however, only superficial. I now understand that the reason why we required lazy evaluation of expressions was because we were not eager enough when typechecking the evaluated AST, so we were missing errors in logical branches of the evaluation that MySQL catches early and that we could only catch by delaying evaluation.

**I've fixed this!** The previous evalengine did four phases, arranged as such:

- During translation:
    - translate
    - simplify
- During evaluation:
    - typechecking
    - evaluation

The new framework now behaves like this:

- During translation:
    - translate
    - typechecking
    - simplify
- During evaluation:
    - evaluation

Not only is this more correct, it is also much more efficient, particularly for complex expressions, because we now only perform the typechecking step _once_ during translation, not once every time we evaluate the expression.

### Static type-checking

The reason why we can now move the typechecking step to translation is because I now understand better the implications of type-checking an SQL expression and the way Vitess API's work. Our previous implementation had a very complex and very expensive type-checking step mostly for one reason: **bind variables**. It was really hard to validate the correctness of an expression that had bvars as placeholders, particularly when the bind variables could contain arbitrary amounts of values (i.e. SQL tuples). But as it turns out, this is not the case in practice! Vitess actually distinguishes between normal bind variables (`Argument` in the SQL parser) and tuple bind variables (`ListArgument`), and our evaluation engine was not keeping this distinction in mind.

By updating our translator and type-checker to be aware of these two different bvar modes, the type checking can be performed eagerly now! And furthermore, it is greatly simplified, because Vitess' SQL parser is smart enough to only allow `ListArgument` (i.e. tuple bind variables) in places where it makes semantic sense in the SQL expression (e.g. in the right-hand side of an `IN`, but not on the left hand side), so there's a wide array of logical checks that the evaluation engine was performing superfluously, because our parser had already validated them. **All these checks have been greatly simplified and they run much faster and during translation**.

...Which brings up to my next point: I've split it from this PR, but based on the re-design and refactoring I've performed here, it is now possible to statically compile _some_ SQL expressions for evaluation! Namely, any dynamic expression that **does not use bind variables** and knows the types of its input columns ahead of time can be optimized _a lot_. This will give a very significant performance boost to some of the existing VTGate use cases and (selfishly) to PlanetScale Boost, since it does never use bind variables during evaluation. But that's a matter for another PR.

### Dynamic `EvalResult`

As part of the lazy evaluation design, which I've removed in this PR, our `EvalResult` type did some very un-ergonomic design choices. Most notably, to prevent using interface types which are always heap-allocated in Go, it used a very large hand-crafted union. This design was a good performance choice in its early iterations, but as we've started making the `evalengine` more complete and supporting more and more MySQL types, the union-as-a-struct approach has become both unmaintainable and very expensive. We just have too many fields in `EvalResult`!

As part of the new evaluation framework, I've replaced the `EvalResult` with an internal `eval` interface, so now the results of every evaluation step are always concrete types that implement the `eval` interface (e.g. `evalInt64`, `evalBytes`, `evalTuple`, etc).

This has several consequences:

- Evaluation is **actually faster** despite the fact that it performs more individual heap allocations (this is from preliminary benchmarks, I'm going to expand on this on a follow-up PR).
- Evaluation uses **less memory**! Again, despite the fact that it performs more individual allocations, the average allocation is just so much smaller than an `EvalResult` that memory usage is down as a whole.
- It opens up our design space to **use a memory arena for evaluation**. This will be handled in a separate PR. I have working branches that use both a manual slab arena and the new experimental `runtime/arena` in Go 1.20, and they're non-invasive. The performance results are very exciting!!!

### Better support for custom types

Last, but not least, and actually the original goal of this PR, is adding support for more custom SQL types, something which was becoming unmanageable with the old `EvalResult` design. With the new way `eval` objects work, it is now trivial to add new Complex Types such as `sqltype.JSON`, which will be the goal of my next PR.

### TL;DR

- **Faster!**
- **No external API changes to the `evalengine`**
- Type-checking is now performed during translation, not evaluation
- More accurate error returns compared to MySQL
- Evaluation results are now dynamically allocated as interfaces
- Enables static compilation (follow-up PR)
- Enables arena usage (follow-up PR)
- Enables JSON support (follow-up PR)

cc @systay @frouioui 

## Checklist


go/vt/vtgate/evalengine/eval.go
```diff
@@ -0,0 +1,315 @@
...
+	case v.IsUnsigned():
+		var uval uint64
+		uval, err := v.ToUint64()
+		if err != nil {
+			return nil, vterrors.Errorf(vtrpcpb.Code_INVALID_ARGUMENT, "%v", err)
+		}
+		return newEvalUint64(uval), nil
+	default:
+		uval, err := strconv.ParseUint(v.RawStr(), 10, 64)
+		if err == nil {
```
dbussink, line 263: The logic in `NewLiteralIntegralFromBytes` checks if it fits in an `int64`, do we need something similar here to be consistent with that? Or does that not matter?

go/vt/vtgate/evalengine/api_hash.go
```diff
@@ -0,0 +1,123 @@
...
+		case v.IsSigned():
+			i, err = v.ToInt64()
+		case v.IsUnsigned():
+			var uval uint64
+			uval, err = v.ToUint64()
+			i = int64(uval)
+		default:
+			return 0, vterrors.Errorf(vtrpcpb.Code_INTERNAL, "coercion should not try to coerce this value to a signed int: %v", v)
+		}
+		return HashCode(uint64(i)), err
```
dbussink, line 74: I did some small simplification of this code in https://github.com/vitessio/vitess/pull/12238 to make it nicer for the CodeQL checks, I think that applies here too? Or does that change have issues potentially (if it has, we should add tests if it does, since that does pass all tests). Afaik it's just fewer casts but in the end `HashCode()` itself is a cast to `uintptr` anyway.

vmg: Ayup, those casts are correct. I've just cherry-picked them into this PR. Could you please update #12238 once this is merged? Thanks!

go/vt/vtgate/evalengine/eval_tuple.go
```diff
@@ -0,0 +1,41 @@
...
+	vtrpcpb "vitess.io/vitess/go/vt/proto/vtrpc"
+	"vitess.io/vitess/go/vt/vterrors"
+)
+
+type evalTuple struct {
+	t []eval
+}
+
+func (e *evalTuple) hash() (HashCode, error) {
+	return 0, vterrors.Errorf(vtrpcpb.Code_UNIMPLEMENTED, "cannot hash TUPLE")
```
systay, line 30: why? can't we just merge the individual hashes using some clever algo with prime numbers?

vmg: This is a wrong error message; the hash method for tuple should never be called because tuple-equality comparison is always done sequentially in the evalengine (to follow MySQL's approach), so the hashes of two tuples should never be compared. `Code_UNIMPLEMENTED` is definitely wrong here. I'll update this to make it more explicit.

go/vt/vtgate/evalengine/eval.go
vmg: No, the `NewLiteralXXX` functions do extra checking to ensure the SQL Parser or people manually using the API do not overflow the numeric types. This is an internal API, however, and MySQL always silently handles these overflows.

systay (approved):

go/vt/vtgate/evalengine/fn_string.go
```diff
@@ -0,0 +1,284 @@
...
+	return sqltypes.VarChar, f1
+}
+
+func (c *builtinCollation) eval(env *ExpressionEnv) (eval, error) {
+	arg, err := c.arg1(env)
+	if err != nil {
+		return nil, err
+	}
+
+	col := collations.Local().LookupByID(evalCollation(arg).Collation)
```
frouioui, line 229: can `col` ever be `nil` if the given collation does not exist?

frouioui (approved):
This is looking good to me, thanks @vmg 🙌🏻 

I had a concern about the different `panic` that we are introducing. Wouldn't it be better to return errors instead of exiting vtgate?

In the description of the PR we can read that this change makes the evalengine more efficient and faster, do you have an idea of how much faster the code would be with this change?

go/vt/vtgate/evalengine/translate_test.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2021 The Vitess Authors.
```
GuptaManan100, line 2: Should we change the copyright of existing file?

go/vt/vtgate/evalengine/fn_string.go
vmg: No, the collations are always tested before translation. There are no missing collations in the engine.

go/vt/vtgate/evalengine/api_types.go
```diff
@@ -0,0 +1,171 @@
...
+	switch num := num.(type) {
+	case *evalInt64:
+		if num.i < 0 {
+			return 0, vterrors.Errorf(vtrpcpb.Code_INVALID_ARGUMENT, "negative number cannot be converted to unsigned: %d", num.i)
+		}
+		return uint64(num.i), nil
+	case *evalUint64:
+		return num.u, nil
+	}
+	panic("unreachable")
```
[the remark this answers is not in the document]
vmg: Sure!

go/vt/vtgate/evalengine/api_hash_test.go
```diff
@@ -49,12 +49,27 @@ func TestHashCodesRandom(t *testing.T) {
...
+	require.NoError(t, err)
+
+	t.Logf("%v %v", hash1, hash2)
+}
+
+func TestHash1(t *testing.T) {
+	hash1, err := NullsafeHashcode(sqltypes.NewVarChar("5989907846286453570"), collations.CollationBinaryID, sqltypes.Float64)
+	require.NoError(t, err)
+	t.Logf("%d", hash1)
+}
```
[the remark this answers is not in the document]
vmg: These are not actual tests, I was just verifying some values manually. I'll remove them :sweat_smile:

go/vt/vtgate/evalengine/fn_string.go
frouioui: Got it!

GuptaManan100 (approved):

go/vt/vtgate/evalengine/translate_test.go
vmg: I think so? It's always safe to bump our own copyrights as long as we're actively working on the specific files (which we are in this PR!).

vmg:
> In the description of the PR we can read that this change makes the evalengine more efficient and faster, do you have an idea of how much faster the code would be with this change?

I'm gonna follow up with some interesting benchmarks as I start working with static compilation. From the internal Boost benchmarks, the speed-ups were not significant (besides the reduced memory usage, which is always good!), but the main goal of this PR was correctness _and_ laying the groundwork to allow us to evaluate these expressions statically, which is where the numbers will start looking good!

go/vt/vtgate/evalengine/translate_test.go
deepthi: actually we should not change the copyright year. if there are substantial changes you _could_ add the new copyright year, but that is not necessary.

PR #12248, opened 2023-02-06
Use consistent type port and replication state

dbussink:
These are used as int32 in the protobuf (and enum maps to int32 for replication state). This means we should also keep using int32 internally if easily possible.

## Related Issue(s)

Part of #12216 

## Checklist

### DIFF
--- a/go/mysql/flavor.go
+++ b/go/mysql/flavor.go
@@ -393,7 +393,7 @@ func (c *Conn) SetReplicationPositionCommands(pos Position) []string {
 // as the new replication source (without changing any GTID position).
 // It is guaranteed to be called with replication stopped.
 // It should not start or stop replication.
-func (c *Conn) SetReplicationSourceCommand(params *ConnParams, host string, port int, connectRetry int) string {
+func (c *Conn) SetReplicationSourceCommand(params *ConnParams, host string, port int32, connectRetry int) string {
 	args := []string{
 		fmt.Sprintf("MASTER_HOST = '%s'", host),
 		fmt.Sprintf("MASTER_PORT = %d", port),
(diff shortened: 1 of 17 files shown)


go/mysql/flavor.go
```diff
@@ -457,23 +457,23 @@ func parseReplicationStatus(fields map[string]string) ReplicationStatus {
...
-		status.ReplicationLagSeconds = uint(parseUint)
+		status.ReplicationLagSeconds = uint32(parseUint)
 	}
-	parseUint, _ = strconv.ParseUint(fields["Master_Server_Id"], 10, 0)
-	status.SourceServerID = uint(parseUint)
-	parseUint, _ = strconv.ParseUint(fields["SQL_Delay"], 10, 0)
-	status.SQLDelay = uint(parseUint)
+	parseUint, _ = strconv.ParseUint(fields["Master_Server_Id"], 10, 32)
+	status.SourceServerID = uint32(parseUint)
+	parseUint, _ = strconv.ParseUint(fields["SQL_Delay"], 10, 32)
```
rsajwani, line 475: nit: I would suggest to create a constant for '32' .. e.g const int32Size = 32

dbussink: @rsajwani That seems overkill to me for stuff like this. It's about this specific value and we have `ParseInt` and related functions in many places.

rsajwani: that was the idea.. its used at alot of place.. a minor change in future will result us to change it everywhere .. strconv already have const define like these.. But I am ok if you fee its overkill

rsajwani (approved):
LGTM. There is a test failure which doesn't look related to this change but still kindly make sure if passes before you push.

go/mysql/flavor.go
dbussink: Integer type sizes will never change :smile:.

dbussink:
> LGTM. There is a test failure which doesn't look related to this change but still kindly make sure if passes before you push.

That's fixed in https://github.com/vitessio/vitess/pull/12253

mattlord (approved):

PR #12249, opened 2023-02-06
Change SQL error codes to use explicit type

dbussink:
This ErrorCode type maps to a uint16 to match what MySQL has as the range and parses things accordingly as well. With the explicit type we know better when we use the specific error code.

## Related Issue(s)

Fixes some of the issues from #12216 which were identified here because of the pretty loose typing.

## Checklist

### DIFF
--- a/go/mysql/client_test.go
+++ b/go/mysql/client_test.go
@@ -37,7 +37,7 @@ import (
 )
 
 // assertSQLError makes sure we get the right error.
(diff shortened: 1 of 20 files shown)


frouioui (approved):
This is looking good to me, thanks for this refactoring!

mattlord (approved):
❤️

PR #12054, opened 2023-01-09
Update the MySQL version used by our Docker images

frouioui:
## Description

This PR upgrades the default MySQL version used by the `vitess/lite:latest` and `vitess/lite:mysql80` images.

- `vitess/lite:latest`: `5.7` -> `8.0.31`
- `vitess/lite:mysql80`: `8.0.23` -> `8.0.31`

## Related Issue(s)

- Fixes #11948

## Checklist

-   [x] "Backport to:" labels have been added if this change should be back-ported
-   [x] Tests were added or are not required
-   [x] Documentation was added or is not required

PR #12252, opened 2023-02-06
Revert default MySQL 80 version to `8.0.30`

frouioui:
## Description

It was found that MySQL 8.0.31 is not supported by the latest version of xtrabackup leading to invalid upgrade paths from `v15.x.x` to `v16.x.x`.

While we wait for percona to release its 8.0.31 version, this PR partially reverts some recent changes made in https://github.com/vitessio/vitess/pull/12054 and https://github.com/vitessio/vitess/pull/11989, the new MySQL 80 version that will be used is `8.0.30`.

## Checklist


rsajwani:
changes look good to me but the test are failing apparently for some other reason ...

mattlord (approved):

GuptaManan100 (approved):
Everything else LGTM

PR #12253, opened 2023-02-06
Use a working MariaDB mirror for CI

dbussink:
The current mirror listed seems to be missing this, so let's point it at a working one.

## Checklist


rsajwani (approved):

mattlord (approved):
Thanks! ❤️

PR #12254, opened 2023-02-06
Subquery bugfix

systay:
## Checklist


go/vt/vtgate/planbuilder/operators/rewrite/rewriters.go
```diff
@@ -73,12 +74,22 @@ func TopDown(in ops.Operator, rewriter BreakableFunc) (ops.Operator, error) {
...
 }
 
-func bottomUp(root ops.Operator, rewriter Func) (ops.Operator, TreeIdentity, error) {
+func bottomUp(root ops.Operator, rootID semantics.TableSet, resolveID func(ops.Operator) semantics.TableSet, rewriter Func) (ops.Operator, TreeIdentity, error) {
 	oldInputs := root.Inputs()
 	anythingChanged := false
 	newInputs := make([]ops.Operator, len(oldInputs))
+	childID := rootID
+
+	type noLHSTableSet interface{ NoLHSTableSet() }
```
frouioui, line 83: Adding a comment that explains why we are introducing this interface might help our future self

frouioui (approved):
lgtm

go/vt/vtgate/planbuilder/testdata/filter_cases.json
```diff
@@ -6696,5 +6696,69 @@
...
+        },
+        "ProjectedIndexes": "-2",
+        "TableName": "unsharded_`user`_unsharded",
+        "Inputs": [
+          {
+            "OperatorType": "Join",
+            "Variant": "Join",
+            "JoinColumnIndexes": "R:0,L:0",
+            "TableName": "unsharded_`user`",
+            "Inputs": [
```
harshit-gangal, line 6720: for the input to semi join it receives.

u1.bar, 1

what does -2 in projected indexes represent?

on semi join we will have 
u1.bar, 1, 1 and take the last column `2` where `-` stands for right.

Is my understanding correct?

harshit-gangal (approved):

go/vt/vtgate/planbuilder/testdata/filter_cases.json
systay: The semi join has the join between `unsharded` and `user` as the input, and it then checks with the last route if each row in the input can be returned or not. The data sent from the LHS to the RHS is `u1.bar`, which we can see here:

```
        "JoinVars": {
          "u1_bar": 0
        },
```

`"ProjectedIndexes": "-2"` on the semi-join means that we will output the second column from the left. Semi-joins never return data from the RHS, so these numbers will always be negative. We can check the second row on the join columns `"JoinColumnIndexes": "R:0,L:0",`, which tells us that the second column is the first column coming from the LHS of the inner join, which is the literal `1`.

Does that make sense?

arthurschreiber (approved):

PR #12256, opened 2023-02-06
backport partial fix from 11786

tanjinx:
## Description

An outer join query used to work in v12 but got a planner error: `unknown plan type for DISTINCT *planbuilder.filter`
a sample query:
```
 SELECT DISTINCT
       wie.id,
       wie.c2,
       wie.c3,
       wie.status,
       wie.data,
       wie.error,
       wie.created
     FROM wb_tbl wie
     INNER JOIN
       wb_tbl wie_f
       ON (
         wie.c2 = wie_f.c2
         AND wie.c3 = wie_f.c3
          AND wie_f.created >= "2023-01-24 00:00:00.000"
       )
     LEFT JOIN
       wb_tbl wie_g
       ON (
         wie.c2 = wie_g.c2
         AND wie.c3= wie_g.c3
         AND wie.created < "2023-01-24 00:00:00.000"
         AND wie.id < wie_g.id
       )
     WHERE
       wie_g.id IS NULL
       AND wie.c2 =10000
      AND wie.status = "completed"
      ORDER BY wie.created DESC, wie.id DESC LIMIT 10;
```

## Related Issue(s)

https://github.com/vitessio/vitess/issues/12259

## Checklist

### DIFF
--- a/go/vt/vtgate/planbuilder/physical/route_planning.go
+++ b/go/vt/vtgate/planbuilder/physical/route_planning.go
@@ -81,17 +81,32 @@ func CreatePhysicalOperator(ctx *plancontext.PlanningContext, opTree abstract.Lo
 	case *abstract.Concatenate:
 		return optimizeUnion(ctx, op)
 	case *abstract.Filter:
-		src, err := CreatePhysicalOperator(ctx, op.Source)
-		if err != nil {
-			return nil, err
-		}
-		return &Filter{
-			Source:     src,
-			Predicates: op.Predicates,
-		}, nil
+		return optimizeFilter(ctx, op)
 	default:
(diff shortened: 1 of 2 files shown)


harshit-gangal:
The changes look good. 
Kindly add the failing query to the test.

frouioui (approved):
This is looking good, thank you! Please add the query you added in the description of your PR into the plan tests and we're good.

tanjinx:
unfortunately the fix is incomplete. I tried a revised/simplified query and still hit the error. 
```
create table test_tbl (c1 int, c2 int, c3 int, c4 int);

select distinct t1.c1
from test_tbl t1
left join test_tbl t3
on t1.c2 = t3.c2
where t3.c1 is NULL;

                [unknown plan type for DISTINCT *planbuilder.filter]
        --- FAIL: TestPlan/filter_cases.txt/3946_Gen4:_ (0.00s)

```

Issue #12255, opened 2023-02-06
Fix help-text for backup_storage_number_blocks


rsajwani:
This is what it is right now 

`
fs.IntVar(&backupCompressBlocks, "backup_storage_number_blocks", backupCompressBlocks, "if backup_storage_compress is true, backup_storage_number_blocks sets the number of blocks that can be processed, at once, before the writer blocks, during compression (default is 2). It should be equal to the number of CPUs available for compression.")
`

'at once' in above text should be either `parallel` or `concurrent`

PR #12258, opened 2023-02-06
fix help text for backup_storage_number_blocks

rsajwani:

## Related Issue(s)

#12255

## Checklist


go/flags/endtoend/vtbackup.txt
```diff
@@ -10,7 +10,7 @@ Usage of vtbackup:
       --backup_storage_block_size int                   if backup_storage_compress is true, backup_storage_block_size sets the byte size for each block while compressing (default is 250000). (default 250000)
       --backup_storage_compress                         if set, the backup files will be compressed. (default true)
       --backup_storage_implementation string            Which backup storage implementation to use for creating and restoring backups.
-      --backup_storage_number_blocks int                if backup_storage_compress is true, backup_storage_number_blocks sets the number of blocks that can be processed, at once, before the writer blocks, during compression (default is 2). It should be equal to the number of CPUs available for compression. (default 2)
+      --backup_storage_number_blocks int                if backup_storage_compress is true, backup_storage_number_blocks sets the number of blocks that can be processed, in parallel, before the writer blocks, during compression (default is 2). It should be equal to the number of CPUs available for compression. (default 2)
```
mattlord, line 13: Since we're here, I think that we should capitalize the sentence describing it. I'm also assuming you don't mean physical CPUs / sockets here. It's also really long and IMO unnecessary repeating the flag name over and over. Suggestion:
```
If backup_storage_compress is true, backup_storage_number_blocks sets the number of blocks that can be processed in parallel before the writer blocks during compression. This value should be equal to the number of vCPUs available. (default 2)
```

mattlord (commented):
Is this text change in response to code changes? If so, could you add a link to the PR(s)? Or perhaps the wording was misleading?

go/flags/endtoend/vtbackup.txt
rsajwani: vCPU is assuming we will be running in hypervisor. If I read details of this api, it actually defaults to GOMAXPROCS, which is basically number of logical CPU (https://pkg.go.dev/runtime#GOMAXPROCS). That is number of the processor's cores multiplied by the number of threads per core. 

Furthermore as per (klauspost)[https://github.com/klauspost/pgzip]

You should at least have a block size of 100k and at least a number of blocks that match the number of cores your would like to utilize, but about twice the number of blocks would be the best.

How about we say 
`This value should be equal to at least the number of cores available.`

mattlord: vCPU has nothing to do with hypervisors. They are equal to the number of hardware threads, with the OS kernel treating each one as a virtual CPU with sockets being physical CPUs and typically NUMA regions. 

If you prefer hardware threads, that’s also fine. My laptop e.g. has 1 CPU with 8 cores, each cote with 2 threads. So 16 vCPUs or hardware threads but 1 (physical) CPU.

rsajwani: so basically you are saying vCPU and logical CPU are same , coz I alway use term logical CPU for this :) ... my bias is because of google search :) ..

mattlord (approved):

deepthi (approved):

Issue #12261, opened 2023-02-07
Bug Report: cannot update _vt.schema_migrations table schema if zero values present in `requested_timestamp`


shlomi-noach:
### Overview of the Issue

Historically, there have been zero values in `_vt.schema_migrations.requested_timestamp` column; this was due t oa bug, fixed in https://github.com/vitessio/vitess/commit/3012060515af8f1997e1e8e28511115025b4372a#diff-d7427b60825b46773b837642e539fd8f00382fc230b717043f3aa76d3ee7ae14R103

Unfortunately, those values violate the default `sql_mode`, which includes `no_zero_date` and `no_zero_in_date`.

If we make changes to `_vt.schema_migrations` table, via the `schemadiff` mechanism, and if zero values are found in `requested_timestamp` column, then those schema changes will fail, and the tablet will not be able to initialize properly.

### Reproduction Steps

deploy recent version, with an old migration that has `0000-00-00 00:00:00` value for `requested_timestamp`.

### Binary Version

```sh
-
```

### Operating System and Environment details

```sh
-
```

PR #12262, opened 2023-02-07
Allow zero (in) date when setting up internal _vt schema

shlomi-noach:
## Description

Fixes https://github.com/vitessio/vitess/issues/12261

Historically, there have been zero values in `_vt.schema_migrations.requested_timestamp` column; this was due to a bug, fixed in https://github.com/vitessio/vitess/commit/3012060515af8f1997e1e8e28511115025b4372a#diff-d7427b60825b46773b837642e539fd8f00382fc230b717043f3aa76d3ee7ae14R103

Unfortunately, those values violate the default `sql_mode`, which includes `no_zero_date` and `no_zero_in_date`.

If we make changes to `_vt.schema_migrations` table, via the `schemadiff` mechanism, and if zero values are found in `requested_timestamp` column, then those schema changes will fail, and the tablet will not be able to initialize properly.

In this PR, we take a general approach allowing zero in date and zero date in the session running the schema changes. This isn't specific to `_vt.schema_migrations` or to any specific column. It ensures we can preserve an already existing situation.

## Related Issue(s)

  - Fixes: https://github.com/vitessio/vitess/issues/12261
  - Related to: https://github.com/vitessio/vitess/issues/12294

## Checklist


rohit-nayak-ps (approved):
Thanks for catching and fixing this!

lgtm

rohit-nayak-ps (changes requested):
A few unit tests need fixing to handle the new queries to get/set sql_mode.

shlomi-noach:
v15 "backport": https://github.com/vitessio/vitess/pull/12263 ; it's not a real backport because of how schema management has diverged, so it's its own fix.

shlomi-noach:
@rohit-nayak-ps what am I doing wrong with the tests here? My fix does not seem to produce the correct result. I'm getting a `No such field in RowNamedValues`

rohit-nayak-ps (approved):

shlomi-noach:
Thank you @rohit-nayak-ps for fixing the tests!

go/vt/sidecardb/sidecardb.go
```diff
@@ -202,6 +208,31 @@ func Init(ctx context.Context, exec Exec) error {
...
+	if err != nil {
+		return nil, vterrors.Errorf(vtrpcpb.Code_UNKNOWN, "could not read sql_mode: %v", err)
+	}
+
+	resetSQLModeFunc := func() {
+		restoreSQLModeQuery := fmt.Sprintf("set @@session.sql_mode='%s'", sqlMode)
+		_, _ = si.exec(si.ctx, restoreSQLModeQuery, 0, false)
+	}
+
+	changeSQLModeQuery := fmt.Sprintf("set @@session.sql_mode=REPLACE(REPLACE('%s', 'NO_ZERO_DATE', ''), 'NO_ZERO_IN_DATE', '')", sqlMode)
```
mattlord, line 229: I think this should be more general. IMO, the point here is that any data that may currently exist in the table — however it got there (keep in mind that `SQL_MODE` is a session level setting) — should not prevent us from reaching the desired schema state. 

I think that we should become fully permissive here (`set sql_mode=''`) as we are not concerned with the data in the table at this point but rather the schema (and things like this do not violate the schema).

go/vt/sidecardb/sidecardb.go
```diff
@@ -202,6 +208,31 @@ func Init(ctx context.Context, exec Exec) error {
 	return nil
 }
 
+// We need to allow zero dates for existing sidecar database tables which may happen to
+// actually have zero (in) date values. setSQLMode gets the current sql_mode, change it to a more relaxed value,
+// defer restoring it to the original value.
+func (si *schemaInit) setSQLMode() (func(), error) {
```
mattlord, line 214: I think this should be `setPermissiveSQLMode()` that effectively removes all restrictions.

mattlord (changes requested):
LGTM! However, I think this should be address the more general issue of `SQL_MODE` in the context of what this package is supposed to achieve.

go/vt/sidecardb/sidecardb.go
rohit-nayak-ps: I wasn't sure if doing that would suppress schema-related errors. So it is guaranteed that `sql_mode` only applies to data validations?

shlomi-noach: Trying to think of other possible scenarios -- no zero for auto-increment? What else could fail the migration?
I'm happy to set an empty `sql_mode`.

> as we are not concerned with the data in the table at this point but rather the schema 

Just pointing out the the bug this PR solves has to do with the data, not with the schema. We';re solving an issue where a row has a timestamp value of `0000-00-00 00:00:00`

mattlord: The DDL defines the table's schema. The `sidecardb` packages purpose is to ensure that the database is using that schema. `SQL_MODE` impacts the MySQL server's behavior beyond the defined schema — the SQL syntax MySQL supports and the data validation checks it performs beyond what the table defines. In our case, disabling the `SQL_MODE` settings ensures that we can apply the correct schema regardless of what data is *already* in the table. 0 dates are merely one example today.

mattlord: To be clear, I can't think of any specific other failure that we'd encounter today. My point is more general and forward looking (we should address the fundamental issue rather than the specific case encountered today).

shlomi-noach: fixed

go/vt/sidecardb/sidecardb.go
shlomi-noach: fixed

go/vt/sidecardb/sidecardb.go
```diff
@@ -202,6 +208,30 @@ func Init(ctx context.Context, exec Exec) error {
 	return nil
 }
 
+// We need to allow zero dates for existing sidecar database tables which may happen to
+// actually have zero (in) date values. setPermissiveSQLMode gets the current sql_mode, change it to a more relaxed value,
+// defer restoring it to the original value.
```
mattlord, line 213: Suggestion:
```
// setPermissiveSQLMode gets the current sql_mode for the session, removes any
// restrictions, and returns a function to restore it back to the original session value.
// We need to allow for the recreation of any data that currently exists in the table, such
// as e.g. allowing any zero dates that may already exist in a preexisting sidecar table.
```

go/vt/sidecardb/sidecardb_test.go
```diff
@@ -27,11 +27,18 @@ import (
...
 	defer db.Close()
 	AddSchemaInitQueries(db, false)
 	db.AddQuery("use dbname", &sqltypes.Result{})
+	sqlMode := sqltypes.MakeTestResult(sqltypes.MakeTestFields(
+		"sql_mode",
+		"varchar"),
+		"ONLY_FULL_GROUP_BY,STRICT_TRANS_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO,NO_ENGINE_SUBSTITUTION",
+	)
+	db.AddQuery("select @@session.sql_mode as sql_mode", sqlMode)
+	db.AddQueryPattern("set @@session.sql_mode=.*", &sqltypes.Result{})
```
mattlord, line 41: We shouldn't need this as it's done above in `AddSchemaInitQueries()`.

mattlord (approved):
LGTM. Thank you! ❤️ 

I had a couple of minor comments that you can address as you see best.

go/vt/sidecardb/sidecardb_test.go
shlomi-noach: fixed

go/vt/sidecardb/sidecardb.go
shlomi-noach: fixed

shlomi-noach:
unit tests still fail with "query not found"

rohit-nayak-ps:
> unit tests still fail with "query not found"

Fixed

PR #12263, opened 2023-02-07
Bugfix/Backport to v15: Fix schema migrations requested_timestamp zero values

shlomi-noach:
Fixes https://github.com/vitessio/vitess/issues/12261

The fix  in https://github.com/vitessio/vitess/pull/12262 cannot be backported because of how internal schema management was redesigned in https://github.com/vitessio/vitess/pull/11520

The solution in this PR is to fix the actual values. They should not be zero. They should have some timestamp value, which we borrow from `added_timestamp` column.

An `UPDATE` query runs on PRIMARY tablet initialization, just after table is created and before the rest of DDLs kick. So it runs once in the lifetime of the tablet. It's a full table scan query, but `_vt.schema_migrations` is never too large.

## Checklist

### DIFF
--- a/go/vt/vttablet/onlineddl/schema.go
+++ b/go/vt/vttablet/onlineddl/schema.go
@@ -393,6 +393,12 @@ const (
 			AND cleanup_timestamp IS NULL
 			AND completed_timestamp IS NULL
 	`
+	sqlFixRequestedTimestamp = `UPDATE _vt.schema_migrations
+		SET
+			requested_timestamp = added_timestamp
+		WHERE
+			requested_timestamp < added_timestamp;
+	`
 	sqlSelectMigration = `SELECT
 			id,
 			migration_uuid,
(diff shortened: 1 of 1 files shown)


mattlord:
I don't think we want this as:
  1. It's a temporary fix relatively late in the major release (15.0.3 is next) and there may not be a later release to undo it
      - We don't require incrementally stepped upgrades to each patch release within a major release (should be able to go from 15.0 to 15.4)
  2. We'd have to go all the way back to 13.0 with this

I'm not sure we need to do anything for the older releases — since I'm not even sure that OnlineDDL was  GA/production-ready ([in v14](https://vitess.io/blog/2022-06-28-announcing-vitess-14/#online-ddl)) back when the original bug that led to the 0 date values for the `schema_migration.request_submitted` field existed? —  _**but if so**_, I feel like we should instead temporarily set a permissive `SQL_MODE` in `WithDDL.unify()`, similar to what we do in VReplication here: https://github.com/vitessio/vitess/blob/release-15.0/go/vt/vttablet/tabletmanager/vreplication/vreplicator.go

I do agree that we should implement a more general solution in the new `sidecardb` package to prevent similar things in the future.

shlomi-noach:
The `upstream` fix is https://github.com/vitessio/vitess/pull/12262

GA was production ready in `v14`, per https://vitess.io/blog/2022-06-28-announcing-vitess-14/

I'm open to all options.

shlomi-noach:
We need a resolution whether this should or should not go into v15 :pray:

frouioui:
Hello @shlomi-noach, does this PR must go onto this week's `v15.0.3` patch release?

rohit-nayak-ps (approved):
Echoing @deepthi and @shlomi-noach's comments, this looks like a practical low impact solution to solve potential blockers for existing 15.0 clusters.

frouioui (approved):

PR #12267, opened 2023-02-07
fixing dependency bot alert

rsajwani:

Fix vulnerability define [here](https://github.com/planetscale/vitess/security/dependabot/8)
Seeing this warning when we push any PR
```
git push -u origin vtadmin-package-vulner                 
Enumerating objects: 9, done.
Counting objects: 100% (9/9), done.
Delta compression using up to 10 threads
Compressing objects: 100% (4/4), done.
Writing objects: 100% (5/5), 558 bytes | 558.00 KiB/s, done.
Total 5 (delta 3), reused 0 (delta 0), pack-reused 0
remote: Resolving deltas: 100% (3/3), completed with 3 local objects.
remote: 
remote: Create a pull request for 'vtadmin-package-vulner' on GitHub by visiting:
remote:      https://github.com/planetscale/vitess/pull/new/vtadmin-package-vulner
remote: 
remote: GitHub found 1 vulnerability on planetscale/vitess's default branch (1 high). To find out more, visit:
remote:      https://github.com/planetscale/vitess/security/dependabot/8
```
I did npm install to fix this issue.

## Related Issue(s)

[alert](https://github.com/planetscale/vitess/security/dependabot/8)

## Checklist

### DIFF
--- a/web/vtadmin/package-lock.json
+++ b/web/vtadmin/package-lock.json
@@ -4417,7 +4417,7 @@
         "boolbase": "^1.0.0",
         "css-what": "^3.2.1",
         "domutils": "^1.7.0",
-        "nth-check": "^1.0.2"
+        "nth-check": "^2.0.1"
       }
     },
     "node_modules/@svgr/plugin-svgo/node_modules/css-tree": {
@@ -27569,7 +27569,7 @@
             "boolbase": "^1.0.0",
             "css-what": "^3.2.1",
             "domutils": "^1.7.0",
-            "nth-check": "^1.0.2"
+            "nth-check": "^2.0.1"
           }
         },
         "css-tree": {
--- a/web/vtadmin/package-lock.json
+++ b/web/vtadmin/package-lock.json
@@ -4451,15 +4451,6 @@
       "integrity": "sha512-iV3XNKw06j5Q7mi6h+9vbx23Tv7JkjEVgKHW4pimwyDGWm0OIQntJJ+u1C6mg6mK1EaTv42XQ7w76yuzH7M2cA==",
       "dev": true
     },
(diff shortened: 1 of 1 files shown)


frouioui (approved):

dbussink (changes requested):
I don't think this is the right way to fix it. It looks like the dependency resolving was manually modified which isn't correct. 

We either need to explicitly force this dependency to a later version using something like https://docs.npmjs.com/cli/v8/configuring-npm/package-json#overrides, or move away from `create-react-app`. 

This specific `nth-check` issue is not really a pressing security issue (see https://github.com/facebook/create-react-app/issues/11174), but the bigger concern is that the project seems to be entirely unmaintained. So we should probably switch away from that for that reason and use something like Vite instead.

Also cc @notfelineit since in another PR we also talked about upgrading things in general, so maybe we should bite the bullet and move to Vite as well then all in 1 go.

dbussink:
> We either need to explicitly force this dependency to a later version using something like https://docs.npmjs.com/cli/v8/configuring-npm/package-json#overrides

https://github.com/vitessio/vitess/pull/12270 uses the override to get rid of the broken dependency. We need to either do that, or move away from `create-react-app` if we want to address this alert. 

> I did npm install to fix this issue.

I don't see how this is possible. The changes here show that for an existing package, the dependency was overwritten which is only possible if you manually change the `package-lock.json`. It's also visible that when you remove `package-lock.json` and then `npm install` again, the changes in this PR are gone.

rsajwani:
@dbussink , should I close this PR then given you mention this is not the right way to resolve this dependency

rsajwani:
closing this PR in favor of https://github.com/vitessio/vitess/pull/12270

PR #12271, opened 2023-02-08
Download mariadb from vitess-resources

frouioui:
## Description

We are now downloading MariaDB from [vitess-resources v4](https://github.com/vitessio/vitess-resources/releases/tag/v4.0).

## Checklist

### DIFF
--- a/go/test/endtoend/vreplication/cluster_test.go
+++ b/go/test/endtoend/vreplication/cluster_test.go
@@ -251,7 +251,7 @@ func downloadDBTypeVersion(dbType string, majorVersion string, path string) erro
 	} else if dbType == "mysql" && majorVersion == "8.0" {
 		versionFile = "mysql-8.0.28-linux-glibc2.17-x86_64-minimal.tar.xz"
 		url = "https://dev.mysql.com/get/Downloads/MySQL-8.0/" + versionFile
-	} else if dbType == "mariadb" && majorVersion == "10.10" {
+	} else if dbType == "mariadb" && majorVersion == "10.6" {
 		versionFile = "mariadb-10.6.12-linux-systemd-x86_64.tar.gz"
 		url = "https://github.com/vitessio/vitess-resources/releases/download/v4.0/" + versionFile
 	} else {
--- a/go/test/endtoend/vreplication/cluster_test.go
+++ b/go/test/endtoend/vreplication/cluster_test.go
@@ -251,8 +251,8 @@ func downloadDBTypeVersion(dbType string, majorVersion string, path string) erro
 	} else if dbType == "mysql" && majorVersion == "8.0" {
 		versionFile = "mysql-8.0.28-linux-glibc2.17-x86_64-minimal.tar.xz"
 		url = "https://dev.mysql.com/get/Downloads/MySQL-8.0/" + versionFile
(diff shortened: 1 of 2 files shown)


dbussink (approved):

systay (approved):

GuptaManan100 (approved):

go/test/endtoend/vreplication/cluster_test.go
```diff
@@ -252,8 +252,8 @@ func downloadDBTypeVersion(dbType string, majorVersion string, path string) erro
 		versionFile = "mysql-8.0.28-linux-glibc2.17-x86_64-minimal.tar.xz"
 		url = "https://dev.mysql.com/get/Downloads/MySQL-8.0/" + versionFile
 	} else if dbType == "mariadb" && majorVersion == "10.10" {
-		versionFile = "mariadb-10.10.2-linux-systemd-x86_64.tar.gz"
-		url = "https://mirror.rackspace.com/mariadb/mariadb-10.10.2/bintar-linux-systemd-x86_64/" + versionFile
+		versionFile = "mariadb-10.6.12-linux-systemd-x86_64.tar.gz"
```
arthurschreiber, line 255:
```suggestion
		versionFile = "mariadb-10.10.3-linux-systemd-x86_64.tar.gz"
```

dbussink: @arthurschreiber We're deliberate here to use the latest LTS so it's not EOL as quickly as `10.10` will be (only 1 year support for that one).

frouioui: Hey @arthurschreiber, we want to use `10.6.12` as it offers a long term support of 5 years, as opposed to 1 year for `10.10.3`

arthurschreiber: Ah, ok. It's just confusing because the line before explicitly checks for `10.10`, but then we end up using `10.6` here.

dbussink: @arthurschreiber ah, that's a good catch, I think we should update that check then.

dbussink: Didn't realize here that the version is specified somewhere else actually used in the test and not here:

```
go/test/endtoend/vreplication/vreplication_test.go:     sourceKsOpts["DBTypeVersion"] = "mariadb-10.10"
```

frouioui: Good catch, i'll fix that!

arthurschreiber: https://github.com/vitessio/vitess/blob/c7605a147fa548c7bf1c0623a23688fd41b1ded5/go/test/endtoend/vreplication/vreplication_test.go#L350 Is the only place where this is coming from, I believe.

arthurschreiber (approved):

frouioui:
Reading through @mattlord comment https://github.com/vitessio/vitess/pull/12036#issuecomment-1371007779, I have reverted the change to use `10.10.3` instead of `10.6.12`.

mattlord (approved):
Thanks, @frouioui! I was going to bring this issue up today and come to some resolution so you saved me a bunch of time.❤️

PR #12275, opened 2023-02-08
schemadiff/OnlineDDL: ExtractConstraintOriginalName understands auto-generated names

shlomi-noach:
## Description

When altering a table that has constraints, e.g. `CHECK` constraints, vitess/OnlineDDL has to:

- generate new and unique names (because constraint names are unique across the schema)
- generate those names deterministically (so that it's the same name on all shards) 
- and, as a gesture, attempt to preserve the original name as suffix.

`schemadiff`, on the other hand, offers diffing tables with a constraint check stratety:

- either require constraint names to be identical between two tables (`ConstraintNamesStrict`)
- or completely ignore the names (`ConstraintNamesIgnoreAll`)
- or ignore the vitess-generated part and only compare the "original name" part (`ConstraintNamesIgnoreVitess`)

Last pieces of information:
- MySQL the user does not have to specify constraint names
- and if they don't, MySQL auto-generates the constraint names.
- but **also** if you have an auto-generated name and you rename the table, the constraint name is renamed to reflect the new table's name.

OK, those are a LOT of conditionals. Bottom line: we do not allow fully-auto-generated names. The user may define them, but we modify them on first sight; we strip out the table-name part of the auto generated name.

Which complicates things because now what is the "original name" of the constraint we attempt to preserve? By our own convention, it's the part that says `chk_1` or something; whatever follows the table name.

Long story short: this PR now improves behavior of `ConstraintNamesIgnoreVitess` by being able to compare an auto-generated name with a vitess-generated name.

Nice to have. Small feature. Brain exploding with conditionals.

## Checklist


shlomi-noach:
ping for review

rohit-nayak-ps (approved):
lgtm

PR #12286, opened 2023-02-09
fix linter warnings

systay:
## Description
Recently added code started producing linter warnings. This PR fixes that.

Also updates the CI checks to be aligned with the pre-commit checks.

## Related Issue(s)
Issue introduced in #12247

### DIFF
--- a/.github/workflows/static_checks_etc.yml
+++ b/.github/workflows/static_checks_etc.yml
@@ -181,7 +181,7 @@ jobs:
 
       - name: Run golangci-lint
         if: steps.skip-workflow.outputs.skip-workflow == 'false' && steps.changes.outputs.go_files == 'true'
-        run: $(go env GOPATH)/bin/golangci-lint run -E revive go/... || exit 1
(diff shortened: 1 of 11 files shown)


harshit-gangal (approved):

frouioui (approved):

GuptaManan100 (approved):
LGTM

vmg:
Are we sure that we can lint all of `./go/...` with `-E revive`? I thought the reason why CI didn't run with `revive` on is because we're far from lint-clean on this. Let's see what the actual run looks like :sweat_smile:

vmg:
OK, so I was right, there are just too many errors in the codebase because `revive` is very specific about the way it likes to name things. After speaking with @systay and @frouioui we've decided to disable `revive` enforcement by default (in CI and the post-commit hooks).

For those following at home, `revive` is just a naming linter... It doesn't provide particularly useful feedback in my experience. Either way, y'all still welcome to run it manually, as it's still included in our linting setup config. Simply run `golangci-lint -E revive ./go/your_code/...` to get feedback on your naming.

PR #12288, opened 2023-02-09
Online DDL: remove legacy "stowaway table" logic

shlomi-noach:
## Description

The "stowaway table" was part of the previous `vitess` cut-over mechanism, described in https://vitess.io/blog/2022-04-06-online-ddl-vitess-cut-over/

However, the logic has changed in https://github.com/vitessio/vitess/pull/11460. We no longer use a stowaway table in the cut-over mechanism. The stowaway table was the place where we would store our original table at cut-over phase. But because the cut-over was not atomic, we had to journal that changed. That was done with the `stowaway_table` column in `_vt.schema_migrations`. We would check this value to be able to rollback a halfway executed migration.

Anyway, now we use atomic cutover. We have a "sentry table" now. But this isn't merely a change of name. The sentry table is ephemeral. That is, it's a real table, but we don't care what happens to it. It's temporary in use. There's no situation where we have to resurrect it.

This PR cleans up all the stowaway table logic, which is unused now. To clarify how unused it is, the only scenario where anyone would still need this logic is if they had a failover during a cut-over phase, and at the same time chose to upgrade their vitess cluster. Which is unreasonable.

## Related Issue(s)

- #6926
- #11460 

## Checklist

### DIFF
--- a/go/vt/sidecardb/schema/onlineddl/schema_migrations.sql
+++ b/go/vt/sidecardb/schema/onlineddl/schema_migrations.sql
@@ -58,7 +58,6 @@ CREATE TABLE IF NOT EXISTS _vt.schema_migrations
     `reverted_uuid`                   varchar(64)      NOT NULL DEFAULT '',
     `is_view`                         tinyint unsigned NOT NULL DEFAULT '0',
     `ready_to_complete`               tinyint unsigned NOT NULL DEFAULT '0',
-    `stowaway_table`                  tinytext         NOT NULL,
     `vitess_liveness_indicator`       bigint           NOT NULL DEFAULT '0',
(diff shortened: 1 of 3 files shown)


mattlord (commented):
Shouldn't we remove it from the schema here too? https://github.com/vitessio/vitess/blob/main/go/vt/sidecardb/schema/onlineddl/schema_migrations.sql#L61

shlomi-noach:
> Shouldn't we remove it from the schema here too?
Yes, indeed. Done.

mattlord (approved):

deepthi (approved):
I assume no tests directly tested this internal detail.

PR #12296, opened 2023-02-09
Addition of flag to disable automated recovery in VTOrc at startup

### LINKED ISSUE #11958: Feature Request: option to disable/enable Automated Recovery permanently
### Feature Description

Currently as we see , VTOrc does not do any filtering to decide if cluster is OK to do automated recovery or not , original orchestrator lets you set filters , so you can skip some cluster. If there are not ready , or you don't feel confidence yet. 

We think it would be better to have at least disabling and enabling option for VTOrc as i understand VTOrc does not want to do filtering.

https://github.com/vitessio/vitess/blob/397fbe2eca8500ca7a36f26ed33c99c0c67184cb/go/vt/vtorc/inst/cluster.go#L31 

### Use Case(s)

Sometimes if you have your own backup/provisioning/recovery solution integrated with Vitess , you might not feel confident about autorecovery for some cluster or you might have doing some long running config change tests on slave and would want to permanently disable it.

GuptaManan100:
## Description

This PR adds a flag to VTOrc which allows the users to disable the global recoveries from VTOrc on start-up. VTOrc will only monitor the tablets and won't fix anything. The problems can be seen at the api endpoints `/api/replication-analysis` and `/api/problems`. 

To make VTOrc start running recoveries, the users will have to enable the global recoveries using the api endpoint `/api/enable-global-recoveries`

## Related Issue(s)

- Fixes #11958 

## Checklist

### DIFF
--- a/go/vt/vtorc/logic/orchestrator.go
+++ b/go/vt/vtorc/logic/orchestrator.go
@@ -17,6 +17,7 @@
 package logic
 
 import (
+	"fmt"
 	"os"
 	"os/signal"
 	"sync"
@@ -373,7 +374,7 @@ func ContinuousDiscovery() {
 		log.Infof("Disabling global recoveries. VTOrc won't run any fixes until global recoveries are enabled")
 		err := DisableRecovery()
 		if err != nil {
-			log.Errorf("Unable to disable recoveries - %v", err)
+			panic(fmt.Sprintf("Unable to disable recoveries - %v", err))
 		}
 	}
 	for {
--- a/go/vt/vtorc/logic/orchestrator.go
+++ b/go/vt/vtorc/logic/orchestrator.go
@@ -366,6 +366,16 @@ func ContinuousDiscovery() {
 	servenv.OnTermSync(closeVTOrc)
 
 	log.Infof("continuous discovery: starting")
+	// Disable global recoveries according to the flags
+	// We are guaranteed that the VTOrc database has already been initialized,
+	// so this is the correct place to disable recoveries.
+	if config.DisableGlobalRecoveries {
+		log.Infof("Disabling global recoveries. VTOrc won't run any fixes until global recoveries are enabled")
+		err := DisableRecovery()
+		if err != nil {
+			log.Errorf("Unable to disable recoveries - %v", err)
+		}
+	}
 	for {
 		select {
 		case <-healthTick:
--- a/doc/releasenotes/17_0_0_summary.md
+++ b/doc/releasenotes/17_0_0_summary.md
@@ -1,0 +1,18 @@
+## Summary
+
+### Table of Contents
+
+- **[Major Changes](#major-changes)**
+  - **[New Command Line Flags and Behavior](#new-flag)**
+    - [VTOrc Flag to Disable Global Recoveries](vtorc-disable-recovery-flag)
+
+
+## <a id="major-changes"/>Major Changes
+
+### <a id="new-flag"/>New Command Line Flags and Behavior
+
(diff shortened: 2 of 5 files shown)


go/vt/vtorc/logic/orchestrator.go
```diff
@@ -366,6 +366,16 @@ func ContinuousDiscovery() {
...
 
 	log.Infof("continuous discovery: starting")
+	// Disable global recoveries according to the flags
+	// We are guaranteed that the VTOrc database has already been initialized,
+	// so this is the correct place to disable recoveries.
+	if config.DisableGlobalRecoveries {
+		log.Infof("Disabling global recoveries. VTOrc won't run any fixes until global recoveries are enabled")
+		err := DisableRecovery()
+		if err != nil {
+			log.Errorf("Unable to disable recoveries - %v", err)
```
rsajwani, line 376: should we not panic here instead of keep going on and do recoveries even we have been told not to...

go/vt/vtorc/logic/orchestrator.go
```diff
@@ -366,6 +366,16 @@ func ContinuousDiscovery() {
 	servenv.OnTermSync(closeVTOrc)
 
 	log.Infof("continuous discovery: starting")
+	// Disable global recoveries according to the flags
+	// We are guaranteed that the VTOrc database has already been initialized,
+	// so this is the correct place to disable recoveries.
+	if config.DisableGlobalRecoveries {
+		log.Infof("Disabling global recoveries. VTOrc won't run any fixes until global recoveries are enabled")
+		err := DisableRecovery()
```
rsajwani, line 374: inside disableRecovery... do we need INSERT IGNORE or just INSERT is enough?

rsajwani:
general question: Just calling it out here. I thought about and does see any issue but  given anyone can enable and disable the flag. Should we worry about any concurrency here ?

GuptaManan100:
@rsajwani What do you mean by concurrency issues? enabling and disabling a flag will require a restart, so I am not sure what concurrency issues there are?

go/vt/vtorc/logic/orchestrator.go
GuptaManan100: Yes, that sounds like a good idea. I'll change it to a panic.

GuptaManan100:
Website docs - https://github.com/vitessio/website/pull/1383

frouioui (approved):
LGTM

serhatcetinkaya:
@GuptaManan100 why didn't you merge the PR ? we were waiting for this to be able to use vt-orc

GuptaManan100:
We had a discussion and the Vitess team doesn't think that it's a good idea to add this flag. We don't want to support this method of running VTOrc wherein it is used only for fault detection. You can cherry-pick the changes in this PR into your private fork if you want to use VTOrc that way.

serhatcetinkaya:
as I see the tool supports disabling it with an API call (`/api/disable-global-recoveries`). to be honest I didn't understand, how come the team doesn't want to support this with a flag but have an endpoint to disable it. does this mean that api endpoint will be removed in future ?

https://github.com/vitessio/vitess/issues/11958#issuecomment-1352558792

GuptaManan100:
@serhatcetinkaya That endpoint is meant for debugging usage and also for temporarily turning off global recoveries. 

The only use case for having a flag to do the same thing that the API already does, is to want to persist the turning off of recoveries across VTOrc restarts. This too is only required if you want to permanently run VTOrc in a mode of only monitoring the cluster. This isn't something we want to support. We haven't had any requests from any other users for this mode of running VTOrc either.

Issue #12269, opened 2023-02-08
Bug Report: bad count result of cross shard outer join


eseokoh:
### Overview of the Issue

count of cross shard outer join shows incorrect result. 

### Reproduction Steps

1. Deploy the following vschema:
```json
{
  "sharded": true,
  "vindexes": {
    "hash": {
      "type": "xxhash"
    },
  "tables": {
    "x0": {
      "column_vindexes": [
        {
          "column": "a",
          "name": "xxhash"
        }
      ]
    },
    "x1": {
      "column_vindexes": [
        {
          "column": "c",
          "name": "xxhash"
        }
      ]
    }
  }
}
```
2. Deploy the following schema:
```sql
create table x0(a int primary key, b int);
create table x1(c int primary key, b int);
insert into x0 (a, b) values (1, 11), (2, null), (3, 0);
insert into x1 (c, b) values (1, 11);
```
3. Run Select: 

```sql
select x0.a from x0 left outer join x1 on x0.b = x1.b;

+---+
| a |
+---+
| 1 |
| 2 |
| 3 |
+---+
3 rows in set (0.01 sec)
```

As you see, this is good, however the followings are bad.

```sql
select count(x0.a) from x0 left outer join x1 on x0.b = x1.b;
+-------------+
| count(x0.a) |
+-------------+
|           1 |
+-------------+
1 row in set (0.01 sec)
```

```sql
select count(*) from x0 left outer join x1 on x0.b = x1.b;

+----------+
| count(*) |
+----------+
|        1 |
+----------+
1 row in set (0.09 sec)
```

### Binary Version

```sh
14.0.4
```

### Operating System and Environment details

```sh
N/A
```

### Log Fragments

```sh
N/A
```

PR #12298, opened 2023-02-09
Fix aggregation on outer joins

harshit-gangal:
## Description

This PR fixes an issue with aggregation on the left join by adding coalesce on the result from the right-hand side table.

## Related Issue(s)

- Fixes https://github.com/vitessio/vitess/issues/12269

## Checklist

-   [X] "Backport to:" labels have been added if this change should be back-ported 14.0 15.0 16.0

### DIFF
--- a/go/test/endtoend/vtgate/queries/aggregation/aggregation_test.go
+++ b/go/test/endtoend/vtgate/queries/aggregation/aggregation_test.go
@@ -395,4 +395,5 @@ func TestAggregateLeftJoin(t *testing.T) {
 	mcmp.AssertMatchesNoOrder("SELECT t1.shardkey FROM t1 LEFT JOIN t2 ON t1.t1_id = t2.id", `[[INT64(1)] [INT64(0)]]`)
 	mcmp.AssertMatches("SELECT count(t1.shardkey) FROM t1 LEFT JOIN t2 ON t1.t1_id = t2.id", `[[INT64(2)]]`)
 	mcmp.AssertMatches("SELECT count(*) FROM t1 LEFT JOIN t2 ON t1.t1_id = t2.id", `[[INT64(2)]]`)
+	mcmp.AssertMatches("SELECT sum(t1.shardkey) FROM t1 LEFT JOIN t2 ON t1.t1_id = t2.id", `[[DECIMAL(1)]]`)
 }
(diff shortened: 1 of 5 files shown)


go/test/endtoend/vtgate/queries/aggregation/aggregation_test.go
```diff
@@ -384,3 +384,16 @@ func TestAggregateRandom(t *testing.T) {
 
 	mcmp.AssertMatches("SELECT /*vt+ PLANNER=gen4 */ t1.shardKey, t1.name, count(t2.id) FROM t1 JOIN t2 ON t1.value != t2.shardKey GROUP BY t1.t1_id", `[[INT64(1) VARCHAR("name 1") INT64(2)] [INT64(2) VARCHAR("name 2") INT64(2)]]`)
 }
+
+func TestAggregateLeftJoin(t *testing.T) {
+	mcmp, closer := start(t)
```
frouioui, line 389: According to the PR's review checklist we should add a comment as to why this test was added and what it tests, even though the name is quite informative already.

frouioui (approved):

go/vt/vtgate/planbuilder/horizon_planning.go
```diff
@@ -381,7 +381,13 @@ func generateAggregateParams(aggrs []operators.Aggr, aggrParamOffsets [][]offset
...
 						Operator: sqlparser.MultOp,
 						Left:     aggrExpr,
-						Right:    curr,
+						Right: &sqlparser.FuncExpr{
+							Name: sqlparser.NewIdentifierCI("coalesce"),
+							Exprs: sqlparser.SelectExprs{
+								&sqlparser.AliasedExpr{Expr: curr},
+								&sqlparser.AliasedExpr{Expr: sqlparser.NewIntLiteral("1")},
+							},
+						},
```
GuptaManan100, line 390: To me this looks like something we would only need for outer joins, right? Is is worth adding that check? Most of the planner tests that are there in the file are doing inner joins, could you add one for left-join too?

GuptaManan100 (approved):

go/vt/vtgate/planbuilder/horizon_planning.go
harshit-gangal: It is non-trivial to track outer joins here. I will add a test for left-join if it does not exist.

PR #12299, opened 2023-02-09
Add parsing support for linestring constructor

Ayman161803:
Signed-off-by: Ayman161803 <mohammedaymannawaz@gmail.com>

## Description
This PR adds parsing support to ``LINESTRING( POINT(x,y), POINT(x,y) ... )`` constructor

Note: Linestring takes in more than one point as argument. Test cases have been altered accordingly. Parsing ``Linestring(Point(x,y))`` with just one point argument gives the following error

image

## Related Issue(s)
#8604 

## Checklist


go/vt/sqlparser/ast.go
```diff
@@ -2710,6 +2710,11 @@ type (
 		YCordinate Expr
 	}
 
+	//LineString represents LineString(POINT(x,y), POINT(x,y), ..) expression
+	LineStringExpr struct {
+		PointParams []*PointExpr
```
systay, line 2715: A `LineString()` function only takes point values as arguments, but these could come from various sources. You could use column values, user defined variables, or other functions that return `point` values. I don't think we can hard code these arguments to `PointExpr`

```sql
mysql> create table geom (pt1 point, pt2 point);

mysql> select linestring(pt1, pt2) from geom;
```

☝️ this is a small example showing what I mean.

My suggestion is to use `Expr` instead of `*PointExpr`

systay:
The error you are seeing for a single point value is not a syntax error, and should not be stopped by the parser. I'd suggest letting the parser accept it, and MySQL or evalengine will fail it with the proper error message

go/vt/sqlparser/ast.go
Ayman161803: That makes a lot of sense. I have updated yacc grammar to accept ``expression_list`` instead of ``point_expression_list``, adding a test case to show that the example that u gave gets parsed now.

Ayman161803:
> The error you are seeing for a single point value is not a syntax error, and should not be stopped by the parser. I'd suggest letting the parser accept it, and MySQL or evalengine will fail it with the proper error message

I have altered ``sql.y`` taking this into consideration. The ``Linestring`` constructor will now accept just one point argument too as shown in [this](https://github.com/vitessio/vitess/blob/3a522ff039b706438fbeb6a67ef76fee19fcc9e8/go/vt/sqlparser/parse_test.go#L188) test case

GuptaManan100 (changes requested):
I checked the MySQL docs, and it does seem that having a lineString with a single point value is allowed by the parser, albeit an execution error - 
```MySQL
mysql [localhost:8026] {msandbox} (test) > create table x1(location GEOMETRYCOLLECTION DEFAULT (LINESTRING(POINT(4, 5))));
Query OK, 0 rows affected (0.24 sec)

mysql [localhost:8026] {msandbox} (test) > select linestring(point(1,1));
ERROR 3037 (22023): Invalid GIS data provided to function linestring.
```

In Vitess too, we should allow line string to have a single point.

go/vt/sqlparser/testdata/select_cases.txt
```diff
@@ -9491,10 +9491,10 @@ OUTPUT
 select count(test.t1.b) from t1
 END
 INPUT
-select st_difference((convert(st_polygonfromwkb(linestring(point(1,1))) using gb18030)), st_geomcollfromwkb(point(1,1)));
+select st_difference((convert(st_polygonfromwkb(linestring(point(1,1), point(2,2))) using gb18030)), st_geomcollfromwkb(point(1,1)));
 END
 OUTPUT
-select st_difference(convert(st_polygonfromwkb(linestring(point(1, 1))) using gb18030), st_geomcollfromwkb(point(1, 1))) from dual
```
GuptaManan100, line 9497: Nevermind, it is already accepting single values. Just revert this change, so that we have a test for that too.

go/vt/sqlparser/ast_format.go
```diff
@@ -2759,3 +2759,16 @@ func (node *Variable) Format(buf *TrackedBuffer) {
...
+// Format formats the node.
+func (node *LineStringExpr) Format(buf *TrackedBuffer) {
+	buf.literal("linestring(")
+	if len(node.PointParams) > 0 {
+		var prefix string
+		for _, n := range node.PointParams {
+			buf.astPrintf(node, "%s%v", prefix, n)
+			prefix = ", "
+		}
+	}
```
GuptaManan100, line 2772: The code on how to print a list of Expressions is already present in the format function of `Exprs`. So we can simply use - 
```golang
// Format formats the node.
func (node *LineStringExpr) Format(buf *TrackedBuffer) {
	buf.astPrintf(node, "linestring(%v)", node.PointParams)
}
```

go/vt/sqlparser/testdata/select_cases.txt
Ayman161803: Reverted

go/vt/sqlparser/ast_format.go
Ayman161803: Updated ``Format`` function accordingly

GuptaManan100 (approved):
LGTM now! Good work!

systay (approved):

PR #12309, opened 2023-02-10
Fix additional typecasting alerts

dbussink:
This should solve the remaining valid typecasting alerts from CodeQL.

## Related Issue(s)

Part of #12216 

## Checklist


go/mysql/collations/remote/collation.go
```diff
@@ -120,11 +120,11 @@ func (c *Collation) Collate(left, right []byte, isPrefix bool) int {
 	c.sql.WriteString(c.suffix)
 	c.sql.WriteString(")")
 
-	var cmp int64
+	var cmp int
```
mattlord, line 123: Isn't int a variable type that is based on the word size of the machine/binary? Meaning that on e.g. an x86 machine it would be an int32 and on an x86_64 machine it would be an int64?

mattlord (commented):
I'm not sure that we want to use `int` here. The rest looks good to me.

go/mysql/collations/remote/collation.go
dbussink: @mattlord yep, but that's already the return value of this method. So that behavior was already problematic. Fwiw, I don't think we even support 32 bit at this point anymore (it's bound to have other issues I suspect).

dbussink: Btw, I don't think it matters in practice for this case since `STRCMP` always returns `-1`, `0` or `1`.

mattlord: Right, but the "return this as MySQL would" does not seem right to me as it's not a concrete type before compilation.

dbussink: Before this change we'd break this theoretically if `STRCMP` was able to return a value in the range of `int64` on 32 bit platforms. With this change we'd get an error in that case which is preferable (although it still doesn't happen anyway). And it means CodeQL doesn't think this is potentially vulnerable either so I think this change is what we'd want here.

mattlord: Do we mean that it returns either a 32 or 64 bit integer as MySQL would?

mattlord: After further discussion, I think this is good. The goal is to avoid typecasting and we need to use an `int` in the end. In this case we're letting go manage the underlying concrete type as needed instead of casting in the source code.

mattlord (approved):

frouioui (approved):

Issue #12314, opened 2023-02-11
Remove deprecated usage of x509.ParseCRL


dbussink:
Go 1.19 provides `x509.ParseRevocationList` and has deprecated `x509.ParseCRL`. We should switch to the new function to parse CRL files.

PR #12315, opened 2023-02-11
Switch to using new Go 1.19 CRL parser

dbussink:
`x509.ParseCRL` is deprecated, we should use `x509.ParseRevocationList` instead which is new in Go 1.19.

## Related Issue(s)

Fixes https://github.com/vitessio/vitess/issues/12314

## Checklist

### DIFF
--- a/go/vt/tlstest/tlstest.go
+++ b/go/vt/tlstest/tlstest.go
@@ -337,12 +337,18 @@ func RevokeCertAndRegenerateCRL(root, parent, name string) {
 	if err != nil {
 		log.Fatal(err)
 	}
-	crlList, err := x509.ParseCRL(data)
+
+	block, _ := pem.Decode(data)
+	if block == nil || block.Type != "X509 CRL" {
+		log.Fatal("failed to parse CRL PEM")
+	}
+
+	crlList, err := x509.ParseRevocationList(block.Bytes)
 	if err != nil {
 		log.Fatal(err)
 	}
 
-	revoked := crlList.TBSCertList.RevokedCertificates
+	revoked := crlList.RevokedCertificates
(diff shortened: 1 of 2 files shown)


go/vt/tlstest/tlstest.go
```diff
@@ -337,12 +337,18 @@ func RevokeCertAndRegenerateCRL(root, parent, name string) {
 	if err != nil {
 		log.Fatal(err)
 	}
-	crlList, err := x509.ParseCRL(data)
+
+	block, _ := pem.Decode(data)
+	if block == nil || block.Type != "X509 CRL" {
+		log.Fatal("failed to parse CRL PEM")
+	}
```
dbussink, line 344: `PEM` decoding is not done implicitly with `x509.ParseRevocationList` so this needs to be done here explicitly in the tests.

go/vt/vttls/crl.go
```diff
@@ -83,7 +82,7 @@ func loadCRLSet(crl string) ([]*pkix.CertificateList, error) {
 			continue
 		}
 
-		parsedCRL, err := x509.ParseCRL(block.Bytes)
+		parsedCRL, err := x509.ParseRevocationList(block.Bytes)
```
dbussink, line 85: `PEM` decoding here is already done explicitly so there's no user facing changes in the usage (still only `PEM` formatted CRLs are supported).

dbussink:
FYI @hkdsun as original contributor of the CRL support. To be clear, at the time this was the right API since Go 1.19 didn't exist yet, but we should still switch now to the newer API.

frouioui (approved):

deepthi (approved):

PR #12316, opened 2023-02-12
Component: vtgate Auto Update tables in "Status For Vtgate" page

jwangace:
## Description
Currently, for each component that has **/debug/status** page, the page is static and does not refresh containing tables. Vtgate page is particularly useful, as a daily user, I'd like to see those tables to be refreshed, this is quite useful when doing deployments && updates.

This PR, added **refreshTableContentIfTableHasID()** javascript function, it access "/debug/status" in the background(every 5 seconds) and looks for those tables ONLY if they have **id** html attribute defined, and only updates those tables.

Because I only care about the status page of Vtgate, I only added the id html attribute on those relevant tables, but it can be applied to any component which has a static page.

## Related Issue(s)
There is None, this is rather a new feature.

## Checklist

### DIFF
--- a/go/vt/servenv/status.go
+++ b/go/vt/servenv/status.go
@@ -81,6 +81,26 @@ var statusHTML = `<!DOCTYPE html>
 <html>
 <head>
 <title>Status for {{.BinaryName}}</title>
+<script src="https://ajax.googleapis.com/ajax/libs/jquery/3.6.3/jquery.min.js"></script>
+<script>
+function refreshTableContentIfTableHasID() {
+  $.get("/debug/status", function(data) {
+    var data = $.parseHTML(data);
+    $("table").each(function() {
+  	  var findTableID = $(this).attr("id");
+  	  if (typeof findTableID !== "undefined") {
+  	    var newTable = $(data).filter("#" + findTableID);
+  	    if (newTable.length > 0) {
+  	  	  $(this).empty().append(newTable.html());
+  	    }
+  	  }
+    });
+  });
+}
+$(document).ready(function() {
+	setInterval(refreshTableContentIfTableHasID, 5000);
+});
+</script>
 <style>
 body {
 font-family: sans-serif;
--- a/go/vt/servenv/status.go
+++ b/go/vt/servenv/status.go
@@ -75,6 +75,9 @@ var (
 	blockProfileRate = 0
 
 	globalStatus = newStatusPage("")
+	// tableRefreshRate defines the time interval in milliseconds of tables get updated,
+	// and only update tables those have "refreshRequired" class
+	tableRefreshRate = 10000
 )
 
 var statusHTML = `<!DOCTYPE html>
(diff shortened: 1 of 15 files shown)


GuptaManan100 (commented):
I really like the idea of auto-refresh. My only comment is to not overload the id field with information. Instead we should create a class (maybe call it `refreshRequired` or something else), and refresh tables that have that class. Using id doesn't seem like a great solution, but it isn't obvious to other developers that adding an id element to a table will cause it to be refreshed.

frouioui:
> I really like the idea of auto-refresh. My only comment is to not overload the id field with information. Instead we should create a class (maybe call it refreshRequired or something else), and refresh tables that have that class. Using id doesn't seem like a great solution, but it isn't obvious to other developers that adding an id element to a table will cause it to be refreshed.

I approve this comment, good idea!

go/vt/servenv/status.go
```diff
@@ -81,6 +81,26 @@ var statusHTML = `<!DOCTYPE html>
...
+  	    if (newTable.length > 0) {
+  	  	  $(this).empty().append(newTable.html());
+  	    }
+  	  }
+    });
+  });
+}
+$(document).ready(function() {
+	setInterval(refreshTableContentIfTableHasID, 5000);
+});
```
frouioui, line 102: I am not sure whether we should make the interval configurable or not. Not everyone will want to overload their VTGate with a request every 5 seconds.

frouioui (commented):
I like the idea of refreshing the page. Thanks for contributing!

jwangace:
> > I really like the idea of auto-refresh. My only comment is to not overload the id field with information. Instead we should create a class (maybe call it refreshRequired or something else), and refresh tables that have that class. Using id doesn't seem like a great solution, but it isn't obvious to other developers that adding an id element to a table will cause it to be refreshed.
> 
> I approve this comment, good idea!

Thank you both @GuptaManan100 and @frouioui for your comments, really helpful. I agree adding class "refreshRequired" is a good idea, it makes things more self-explanatory, definitely can add this class.
However, technically I still think we need to attach an id to each table to make it easier to identify each table, still possible to do that without an id, for example, for this table:
```
<table>
  <tr>
    <th colspan="5">HealthCheck Tablet Cache</th>
  </tr>
  <tr>
   <th>Cell</th>
   <th>Keyspace</th>
   <th>Shard</th>
   <th>TabletType</th>
   <th>TabletHealth</th>
  </tr>
...
</table>
```
The only signature is either the head **HealthCheck Tablet Cache** or the table <th> tags, which is a per table different, this potentially would make javascript code turn ugly pretty quickly.
For now I don't have a good idea but I will put some thoughts on it.

go/vt/servenv/status.go
jwangace: totally make sense, do you mind giving me suggestions on which place is good for this config and what's the good value as a default?

jwangace: I updated this as a variable, and set it to 10 seconds as default, this should be a value that won't overload vtgate.

PR #12321, opened 2023-02-12
Internal refactor: LastError as a public struct

shlomi-noach:
## Description

`lastError` is a mechanism introduced into `VReplication` in https://github.com/vitessio/vitess/pull/10429 ; the idea is that we can have a retry-able operation that waits for some error to clean up, or times out if the error persists.

Up till now this was a local class in `vreplication` package. I have a use case outside VReplication (well, related, it's Online DDL). Anyway, the mechanism is really quite generic and useful, and so I'm externalizing the structure and moving it to `go/vt/vterrors`.

There's no real code changes here other than using the `name` field of `LastError` in place of hard coded `"Vreplication"` text.

This refactor includes the recent changes made in https://github.com/vitessio/vitess/pull/12114

## Related Issue(s)

- #10429
- #12114

## Checklist

### DIFF
--- a/go/vt/vterrors/last_error.go
+++ b/go/vt/vterrors/last_error.go
@@ -86,7 +86,7 @@ func (le *LastError) ShouldRetry() bool {
 		// within the max time range
 		return true
 	}
-	log.Errorf("%s: the same error was encountered continuously since %s, it is now assumed to be unrecoverable; any affected workflows will need to be manually restarted once error '%s' has been addressed",
+	log.Errorf("%s: the same error was encountered continuously since %s, it is now assumed to be unrecoverable; any affected operations will need to be manually restarted once error '%s' has been addressed",
 		le.name, le.firstSeen.UTC(), le.err)
 	return false
 }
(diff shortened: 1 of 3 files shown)


rohit-nayak-ps (approved):
lgtm

go/vt/vterrors/last_error.go
```diff
@@ -87,7 +86,7 @@ func (le *lastError) shouldRetry() bool {
 		// within the max time range
 		return true
 	}
-	log.Errorf("VReplication encountered the same error continuously since %s, we will assume this is a non-recoverable error and will not retry anymore; the workflow will need to be manually restarted once error '%s' has been addressed",
-		le.firstSeen.UTC(), le.err)
+	log.Errorf("%s: the same error was encountered continuously since %s, it is now assumed to be unrecoverable; any affected workflows will need to be manually restarted once error '%s' has been addressed",
```
mattlord, line 89: Should we say operations instead of workflows?

mattlord (approved):

go/vt/vterrors/last_error.go
shlomi-noach: fixed.

shlomi-noach:
ignoring `Upgrade Downgrade Testing Query Serving` test

shlomi-noach:
Will backport this maually

Issue #12326, opened 2023-02-13
Bug Report: regression, VReplication errors lost their type, converted to errorString


shlomi-noach:
### Overview of the Issue

In Vreplication, the function `isUnrecoverableError()` needs to identify specific errors from which it is impossible to recover. This is done by casting the error to `SqlErr` and checking for specific SQL code.

https://github.com/vitessio/vitess/blob/c4292b7c560766275e22be8105cf95d9d5afe7b8/go/vt/vttablet/tabletmanager/vreplication/utils.go#L123-L135

However, at this time, we find the error cannot be casted to `SqlErr` and is in fact `errorString`, which is the type you get when you e.g. `return fmt.Errorf("task error: %s", err)`

We need to retrace the changes and ensure to use proper wrapping (such as `vterrors.Wrapf`) as well as add unwrapping functionality for `isUnrecoverableError()` to use. 

The impact of this issue is that VReplication does not bail out on clearly terminal scenarios, and instead keeps on retrying until some timeout (default `10min`). This is still logically correct. But in CI we don't have that much time to spend.

### Reproduction Steps

-

### Binary Version

```sh
-
```

### Operating System and Environment details

```sh
-
```

shlomi-noach:
The regression is caused by https://github.com/vitessio/vitess/pull/10828, where a bunch of `fmt.Errorf(...)` wrapping were introduced. e.g.:

https://github.com/vitessio/vitess/pull/10828/files#diff-a1cffc790e352be31a3f600180d968b6d96f3bd90acf4bfa0a49e3a66611558cR1033-R1065

systay:
There is a helper method `go/mysql/sql_error.go:NewSQLErrorFromError()` that will do a bit of magic to extract a SQLError from inside a given error. I think we should use that more!

PR #12325, opened 2023-02-12
Fixing `onlineddl_vrepl` flakiness, and adding more tests

shlomi-noach:
## Description

This PR depends on:

- https://github.com/vitessio/vitess/issues/12318
- https://github.com/vitessio/vitess/issues/12320
- https://github.com/vitessio/vitess/issues/12321
- https://github.com/vitessio/vitess/issues/12323
- https://github.com/vitessio/vitess/pull/12327

And on top of these:

- Uses topo-bases throttler
- Breaks down throttler/PRS test into more distinct parts
- Adds new `--postpone-completion`/PRS test
- Fixes reading of VReplication status (reduces test time by `60sec`)

**Only makes sense to review once all the above are merged**

## Related Issue(s)

- #6926
- https://github.com/vitessio/vitess/issues/12317
- https://github.com/vitessio/vitess/issues/12319
- https://github.com/vitessio/vitess/issues/12322

## Checklist

-   [ ] "Backport to:" labels have been added if this change should be back-ported
-   [ ] Tests were added or are not required
-   [ ] Documentation was added or is not required

PR #12327, opened 2023-02-13
(re)Formalize SQLError in VReplication, add underlying wrap/unwrap functionality

shlomi-noach:
## Description

Fixes https://github.com/vitessio/vitess/issues/12326

This PR re-formalizes SQLError in vreplication in two ways:

- Uses `vterrors.Wrapf(...)` as opposed to `fmt.Errorf(...)`
- As fallback method, uses `NewSQLErrorFromError()` as suggested by @systay in https://github.com/vitessio/vitess/issues/12326#issuecomment-1427426676

While here:

- Added `Unwrap` and `UnwrapAll` functions in `vterrors`
- Using said functions in `NewSQLErrorFromError()` to first (attempt to) get the original error object, before resorting to parsing.

Added a bunch of unit tests.

## Related Issue(s)

Fixes https://github.com/vitessio/vitess/issues/12326
Precondition for https://github.com/vitessio/vitess/pull/12323
Precondition for https://github.com/vitessio/vitess/pull/12325
Formalizes changes made in https://github.com/vitessio/vitess/pull/10828

## Checklist


systay (approved):

go/vt/vterrors/vterrors.go
```diff
@@ -244,6 +245,26 @@ func Wrapf(err error, format string, args ...any) error {
...
+	}
+	return false, err
+}
+
+// UnwrapAll attempts to recursively unwrap the given error, and returns the most underlying cause
+func UnwrapAll(err error) error {
+	wasWrapped := true
+	for wasWrapped {
+		wasWrapped, err = Unwrap(err)
+	}
```
mattlord, line 264: I feel like we should probably add a limit here to prevent never getting out of here some odd reason.

go/vt/vttablet/tabletmanager/vreplication/utils.go
```diff
@@ -124,10 +124,13 @@ func isUnrecoverableError(err error) bool {
 	if err == nil {
 		return false
 	}
-	sqlErr, isSQLErr := err.(*mysql.SQLError)
+	sqlErr, isSQLErr := mysql.NewSQLErrorFromError(err).(*mysql.SQLError)
```
mattlord, line 127: I'm not sure how this is better? It should already be an SQLError. Won't isSQLErr always be true in this case?

go/vt/vttablet/tabletmanager/vreplication/utils.go
```diff
@@ -124,10 +124,13 @@ func isUnrecoverableError(err error) bool {
...
 	if !isSQLErr {
 		return false
 	}
+	if sqlErr.Num == mysql.ERUnknownError {
+		return false
+	}
```
mattlord, line 133: I think this is only needed here because isSQLErr will now always be true, no?

mattlord (commented):
The wrapping parts makes sense to me, but unless I'm missing something we should remove the `utils.go` changes.

go/vt/vttablet/tabletmanager/vreplication/utils.go
shlomi-noach: Answered in main comment

go/vt/vttablet/tabletmanager/vreplication/utils.go
shlomi-noach: Answered in main comment

go/vt/vterrors/vterrors.go
shlomi-noach: The logic is legit IMHO and the unit tests make good coverage. Please give this another thought. Would you want to see code that counts to `50` and then fails? Would that code look clearer or even more confusing? Please let me know and I'll do whatever you think.

shlomi-noach:
> but unless I'm missing something we should remove the utils.go changes.

The addition to `utils.go`, ie the use of `NewSQLErrorFromError()`, is a fallback technique, designed to prevent future regressions. The function `NewSQLErrorFromError()` parses the text of an error, and reconstructs a `SqlError` object if the text seems to indicate a `SQLError`. I cannot design a `endtoend` or ` unit` test that validates that any error path does indeed produce a real `SqlError`; it's impossible to validate that no one ever wraps an error with `fmt.Errorf()`; that's because we can't anticipate any and all errors, or any and all code paths. And some errors actually will _not_ be `SqlError`.

So the idea of this safeguard is very appealing and I think we should keep it.


Now, as it happens, the current implementation `NewSQLErrorFromError()` _always returns a `SqlError`_, even if the error really has nothing to do with SQL. That's the implementation and I chose not to deal with it in this PR. I've already opened 6 different PRs for tackling the flaky tests, going down quite a few rabbit holes. So I wanted to stop at some point. I confirm to the existing behavior, and maybe in a future PR I'll tackle the `NewSQLErrorFromError()` logic.

At any case, if the error has nothing to do with SQL, you get a `error` object, which you need to cast to `SqlError`, that has `sqlErr.Num == mysql.ERUnknownError`. That's the existing logic and I agree it's confusing. But I still prefer to not try and fix everything in this PR.

go/mysql/sql_error.go
```diff
@@ -87,6 +87,9 @@ func NewSQLErrorFromError(err error) error {
 		return nil
 	}
 
+	// in case this error is wrapped by vterrors, extract the inner most error
+	err = vterrors.UnwrapAll(err)
```
rohit-nayak-ps, line 91: This is breaking an existing test. So we may be expecting the current functionality elsewhere. Tagging @GuptaManan100 since the test was added fairly recently by him.

```
  server_flaky_test.go:1290: 
2023-02-15T10:11:35.1823236Z         	Error Trace:	/home/runner/work/vitess/vitess/go/mysql/server_flaky_test.go:1290
2023-02-15T10:11:35.1823826Z         	Error:      	"Row count exceeded 10000 (errno 1317) (sqlstate 70100) during query: error" does not contain "wrapped: Row count exceeded 10000"
2023-02-15T10:11:35.1824147Z         	Test:       	TestErrorCodes/wrapped:_Row_count_exceeded_10000
```

shlomi-noach: FWIW I intend to do more work on `NewSQLErrorFromError` in a separate PR. At any case happy to accommodate.

GuptaManan100: Sweeeet! Thank-you!

rohit-nayak-ps (approved):
lgtm

shlomi-noach:
@mattlord since you had comments, can you please take another look?

go/vt/vterrors/vterrors.go
mattlord: Yeah, what to do when this fails due to hitting the limit would be unknown. We have some protection from infinite recursion and would panic: https://github.com/golang/go/commit/757e0de89f80e89626cc8b7d6e670c0e5ea7f192

So I'm totally fine with it as-is.

mattlord (approved):

mattlord:
> > but unless I'm missing something we should remove the utils.go changes.
> 
> The addition to `utils.go`, ie the use of `NewSQLErrorFromError()`, is a fallback technique, designed to prevent future regressions. The function `NewSQLErrorFromError()` parses the text of an error, and reconstructs a `SqlError` object if the text seems to indicate a `SQLError`. I cannot design a `endtoend` or ` unit` test that validates that any error path does indeed produce a real `SqlError`; it's impossible to validate that no one ever wraps an error with `fmt.Errorf()`; that's because we can't anticipate any and all errors, or any and all code paths. And some errors actually will _not_ be `SqlError`.
> 
> So the idea of this safeguard is very appealing and I think we should keep it.
> 
> Now, as it happens, the current implementation `NewSQLErrorFromError()` _always returns a `SqlError`_, even if the error really has nothing to do with SQL. That's the implementation and I chose not to deal with it in this PR. I've already opened 6 different PRs for tackling the flaky tests, going down quite a few rabbit holes. So I wanted to stop at some point. I confirm to the existing behavior, and maybe in a future PR I'll tackle the `NewSQLErrorFromError()` logic.
> 
> At any case, if the error has nothing to do with SQL, you get a `error` object, which you need to cast to `SqlError`, that has `sqlErr.Num == mysql.ERUnknownError`. That's the existing logic and I agree it's confusing. But I still prefer to not try and fix everything in this PR. There's a dozen or more files that use this pattern and I wanted to keep this PR focused.

Makes sense (I read through what that code does and how too). Thanks!

shlomi-noach:
Not sure what changed that a bunch of `vtgate_*` CI tests consistently fail. I merged `main` but to no effect.

shlomi-noach:
The failing tests are due to the change

```go
	err = vterrors.UnwrapAll(err)
```

I'm digging in.

shlomi-noach:
OK i'm reverting the `err = vterrors.UnwrapAll(err)` change, it is not very important to this PR, will handle it in a later PR. It seems to affect too many tests.

shlomi-noach:
I see why the `Unwrap` addition in `sql_error.go` caused the tests to fail. There's logic that attempts to convert a `vterrors` error code into SQL code. `Unwrap` stripped down the `vterrros` information, thus said info was lost.

shlomi-noach:
Will cherry pick this to v16 manually, since I've merged `main` in the process.

PR #12328, opened 2023-02-13
SidecarDB Init: don't fail on schema init errors

rohit-nayak-ps:
## Description

The initial implementation of the new mechanism for initializing the sidecar database in https://github.com/vitessio/vitess/pull/11520 causes a failure if there is an error while applying a DDL to create or alter even a single sidecardb table.  Such a failure causes the tablet initialization to fail. Hence if there is an issue with one of the queries to create or upgrade a table, say due to an incompatible MySQL version we would end up with tablets not reaching the serving state. 

Now, if this happens during a Vitess version upgrade, it would not be catastrophic because the upgrade process would be stopped when the first replica being upgraded fails to launch.

However if, for any reason, the entire cluster rolls over to a new Vitess version in the presence of these failures, we would have a catastrophic incident where we could not serve queries.  This can also happen if we upgrade the MySQL version on all tablets for the same Vitess version.

For example, we recently encountered a user cluster where a schema upgrade (using `withddl`) had caused a cluster to fail when the underlying MySQL was upgraded. This was due to a data inconsistency: one of the sidecar tables had zero dates and a table alter which added an unrelated column failed on newer versions of MySQL.  

While such issues might happen very rarely, the possible existence of such critical issues is a strong reason not to fail tablet launches if we encounter errors during the sidecar init process.

This PR ignores errors in the `sidecardb` module when we execute `create` or `alter` queries on sidecardb tables.  Quietly failing errors or just logging is also not ideal since it means some modules or some functionality of modules can fail due to an incorrect schema of a table used by that module and debugging them can be time consuming (Note that we had the same issue the `withDDL` approach). 

To surface these issues we have added two stats which are available via the `debug/vars` endpoint: 
  * `SidecarDBDDLErrorCount` which is the number of errors encountered
  * `SidecarDBDDLErrors` which is a list of the corresponding golang errors 
 
A future PR will add functionality to VTAdmin to create visual alerts based on these stats.

## Related Issue(s)
https://github.com/vitessio/vitess/issues/10133

## Checklist

### DIFF
--- a/go/test/endtoend/tabletmanager/replication_manager/tablet_test.go
+++ b/go/test/endtoend/tabletmanager/replication_manager/tablet_test.go
@@ -164,11 +164,12 @@ func waitForSourcePort(ctx context.Context, t *testing.T, tablet cluster.Vttable
 	return fmt.Errorf("time out before source port became %v for %v", expectedPort, tablet.Alias)
 }
 
-func getSidecarDbDDLQueryCount(tablet *cluster.VttabletProcess) (int64, error) {
+func getSidecarDBDDLQueryCount(tablet *cluster.VttabletProcess) (int64, error) {
 	vars := tablet.GetVars()
-	val, ok := vars["SidecarDbDDLQueryCount"]
+	key := "SidecarDBDDLQueryCount"
+	val, ok := vars[key]
 	if !ok {
-		return 0, fmt.Errorf("SidecarDbDDLQueryCount not found in debug/vars")
+		return 0, fmt.Errorf("%s not found in debug/vars", key)
 	}
 	return int64(val.(float64)), nil
 }
@@ -178,7 +179,7 @@ func TestReplicationRepairAfterPrimaryTabletChange(t *testing.T) {
 	err := waitForSourcePort(ctx, t, replicaTablet, int32(primaryTablet.MySQLPort))
 	require.NoError(t, err)
 
-	sidecarDDLCount, err := getSidecarDbDDLQueryCount(primaryTablet.VttabletProcess)
+	sidecarDDLCount, err := getSidecarDBDDLQueryCount(primaryTablet.VttabletProcess)
 	require.NoError(t, err)
 	// sidecar db should create all _vt tables when vttablet started
 	require.Greater(t, sidecarDDLCount, int64(0))
@@ -197,7 +198,7 @@ func TestReplicationRepairAfterPrimaryTabletChange(t *testing.T) {
 	err = waitForSourcePort(ctx, t, replicaTablet, int32(newMysqlPort))
 	require.NoError(t, err)
 
-	sidecarDDLCount, err = getSidecarDbDDLQueryCount(primaryTablet.VttabletProcess)
+	sidecarDDLCount, err = getSidecarDBDDLQueryCount(primaryTablet.VttabletProcess)
 	require.NoError(t, err)
 	// sidecardb should find the desired _vt schema and not apply any new creates or upgrades when the tablet comes up again
 	require.Equal(t, sidecarDDLCount, int64(0))
--- a/go/test/endtoend/tabletmanager/replication_manager/tablet_test.go
+++ b/go/test/endtoend/tabletmanager/replication_manager/tablet_test.go
@@ -24,6 +24,8 @@ import (
 	"testing"
 	"time"
 
+	"vitess.io/vitess/go/vt/sidecardb"
+
 	"github.com/stretchr/testify/require"
 
 	"vitess.io/vitess/go/test/endtoend/cluster"
@@ -166,7 +168,7 @@ func waitForSourcePort(ctx context.Context, t *testing.T, tablet cluster.Vttable
 
 func getSidecarDBDDLQueryCount(tablet *cluster.VttabletProcess) (int64, error) {
 	vars := tablet.GetVars()
-	key := "SidecarDBDDLQueryCount"
+	key := sidecardb.StatsKeyQueryCount
 	val, ok := vars[key]
 	if !ok {
 		return 0, fmt.Errorf("%s not found in debug/vars", key)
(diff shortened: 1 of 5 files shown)


go/vt/sidecardb/sidecardb.go
```diff
@@ -71,10 +73,43 @@ func (t *sidecarTable) String() string {
...
+// failOnSchemaInitError decides whether we fail the schema init process when we encounter an error while
+// applying a table schema upgrade ddl or continue with the next table.
+// If true, tablets will not launch. The cluster will not come up until the issue is resolved.
+// If false, the init process will continue trying to upgrade other tables. So some functionality might be broken
+// due to an incorrect schema, but the cluster should come up and serve queries.
+// This is an operational trade-off: if we always fail it could cause a major incident since the entire cluster will be down.
+// If we are more permissive, it could cause hard-to-detect errors, because a module
+// doesn't load or behaves incorrectly due to an incomplete upgrade. Errors however will be reported and if the
+// related stats endpoints are monitored we should be able to diagnose/get alerted in a timely fashion.
+const failOnSchemaInitError = false
```
deepthi, line 96: For now this is purely internal and there's no way to make this true, correct?

go/vt/sidecardb/sidecardb.go
```diff
@@ -71,10 +73,43 @@ func (t *sidecarTable) String() string {
...
 
 func init() {
 	initSchemaFiles()
-	ddlCount = stats.NewCounter("SidecarDbDDLQueryCount", "Number of create/upgrade queries executed")
+	ddlCount = stats.NewCounter("SidecarDBDDLQueryCount", "Number of create/upgrade queries executed")
```
deepthi, line 100: nit: why `create/upgrade`? why not just `Number of queries ...`?

deepthi (commented):
Needs a proper description and link to tracking issue / previous PR.

go/vt/sidecardb/sidecardb.go
rohit-nayak-ps: Yes, I implemented it as such just to make the design decision very explicit in code.

go/vt/sidecardb/sidecardb.go
rohit-nayak-ps: fixed

rohit-nayak-ps:
> Needs a proper description and link to tracking issue / previous PR.

Duh, sorry, was multi-tasking and forgot to save the description in another tab.

go/test/endtoend/tabletmanager/replication_manager/tablet_test.go
```diff
@@ -164,11 +164,12 @@ func waitForSourcePort(ctx context.Context, t *testing.T, tablet cluster.Vttable
 	return fmt.Errorf("time out before source port became %v for %v", expectedPort, tablet.Alias)
 }
 
-func getSidecarDbDDLQueryCount(tablet *cluster.VttabletProcess) (int64, error) {
+func getSidecarDBDDLQueryCount(tablet *cluster.VttabletProcess) (int64, error) {
 	vars := tablet.GetVars()
-	val, ok := vars["SidecarDbDDLQueryCount"]
+	key := "SidecarDBDDLQueryCount"
```
frouioui, line 169: Should this key be a constant? I see we are using it in several places.

frouioui (approved):

go/vt/sidecardb/sidecardb.go
```diff
@@ -71,10 +73,43 @@ func (t *sidecarTable) String() string {
...
+	ddlErrorCount = stats.NewCounter("SidecarDBDDLErrorCount", "Number of errors during sidecar schema upgrade")
+	ddlErrorHistory = history.New(maxDDLErrorHistoryLength)
+	stats.Publish("SidecarDBDDLErrors", stats.StringMapFunc(func() map[string]string {
+		mu.Lock()
+		defer mu.Unlock()
+		result := make(map[string]string, len(ddlErrorHistory.Records()))
+		for _, e := range ddlErrorHistory.Records() {
+			d := e.(*ddlError)
+			result[d.tableName] = d.err.Error()
+		}
```
mattlord, line 110: To be safe, we should check that it was a valid cast. We don't want to crash the vttablet process for new error handling.

go/vt/sidecardb/sidecardb.go
```diff
@@ -357,17 +406,31 @@ func (si *schemaInit) ensureSchema(table *sidecarTable) error {
 		}
 		_, err := si.exec(ctx, ddl, 1, true)
 		if err != nil {
-			log.Errorf("Error running ddl %s for table %s during sidecar database initialization %s: %+v", ddl, table, err)
-			return err
+			ddlErr := vterrors.Wrapf(err,
+				"Error running ddl %s for table %s during sidecar database initialization", ddl, table)
```
mattlord, line 410: Same note here about capitalizing DDL.

go/vt/sidecardb/sidecardb_test.go
```diff
@@ -40,50 +48,72 @@ func TestAllSidecarDB(t *testing.T) {
...
 		}
+		// return errors for the table creation ddls applied for tables specified in schemaError
+		showCreateTableQueryPrefix := "show create table"
+		if !strings.Contains(strings.ToLower(query), showCreateTableQueryPrefix) {
+			for _, e := range schemaErrors {
+				if strings.Contains(query, e.tableName) {
+					return nil, fmt.Errorf(e.errorValue)
+				}
+			}
+		}
```
mattlord, line 82: IMO it's worth avoiding this imprecise string matching. We could instead parse it, walk the AST, have a case where the type is not SHOW, then if the table identifier == the table name in the test we return the error. I can help with this if you want.

mattlord (approved):
I only had some minor nits and suggestions. I will let you make the final call on those.

go/test/endtoend/tabletmanager/replication_manager/tablet_test.go
rohit-nayak-ps: Fixed

PR #12335, opened 2023-02-13
Auto refresh tables on vtgate Status Page

jwangace:
## Description
Currently, for each component that has **/debug/status** page, the page is static and does not refresh containing tables. Vtgate page is particularly useful, as a daily user, I'd like to see those tables to be refreshed, this is quite useful when doing deployments && updates.

This PR, added **refreshTablesHaveClassrefreshRequired()** javascript function, it access "/debug/status" in the background(every 10 seconds by default) and looks for those tables ONLY if they have **class of refreshRequired** attribute, and only updates those tables.

Because I only care about the status page of Vtgate, I only added the class attribute on those relevant tables, but it can be applied to any component which has a static page.

## Related Issue(s)
There is None, this is rather a new feature.

## Checklist


frouioui:
For review tracking purposes: This PR is a follow-up of https://github.com/vitessio/vitess/pull/12316

go/vt/servenv/status.go
```diff
@@ -75,6 +75,9 @@ var (
 	blockProfileRate = 0
 
 	globalStatus = newStatusPage("")
+	// tableRefreshRate defines the time interval in milliseconds of tables get updated,
+	// and only update tables those have "refreshRequired" class
+	tableRefreshRate = 10000
```
frouioui, line 80: Regarding the comment I wrote on https://github.com/vitessio/vitess/pull/12316#discussion_r1104055991, I am still unsure if we should make this value configurable through a CLI flag. @GuptaManan100, wdyt?

frouioui (approved):
This is looking good to me overall, though let's confirm this with a second review.

go/vt/servenv/status.go
GuptaManan100: I think it makes sense to have this as a flag. The default being a very large value so that the current behaviour remains unchanged and is opt-in.

GuptaManan100 (commented):
Everything else LGTM other than the comment ☝️

go/vt/servenv/status.go
jwangace: Thanks @GuptaManan100 and @frouioui, so I guess this is the right place to go? https://github.com/vitessio/vitess/blob/main/go/vt/servenv/servenv.go#L92

Also, I would love to hear your opinions on what you think a default value is. TBH I think 10 seconds is relatively a conservative number, I probably would set it to 5 in our vitess clusters.

jwangace:
> Everything else LGTM other than the comment ☝️

I moved that variable to flag

go/vt/servenv/status.go
frouioui: I think both ways are fine, we let the user configure it anyway if they're not happy with the default.

frouioui: `servenv.go` looks like a good place for me to add the flag 👍🏻

frouioui: However, watch out, the `TestHelpOutput` test in `go/flags/endtoend` will fail since you are adding a flag to several binaries. What the test does is to execute `<all vitess binaries> --help` and compare the output with the `.txt` files stored at the root of the same package. You will need to update the `.txt`s files with the new list of flags.

frouioui: However, watch out, the `TestHelpOutput` test in `go/flags/endtoend` will fail since you are adding a flag to several binaries. What the test does is to execute `<all vitess binaries> --help` and compare the output with the `.txt` files stored at the root of the same package. You will need to update the `.txt`s files with the new list of flags.

GuptaManan100: I think the default should be very high like `10000000` seconds, so that the behaviour is disabled by default. If someone wants the refresh, then they can do it be adding the flag. WDYT @frouioui?

frouioui: @GuptaManan100 i like the idea of having it disabled by default but in that case let's use a value such as `0` that means "no refresh"

GuptaManan100:
@dbussink Has some concerns about adding a dependency on jQuery. He is currently out of office. We can wait for his review too once he is back.

jwangace:
> @dbussink Has some concerns about adding a dependency on jQuery. He is currently out of office. We can wait for his review too once he is back.

sounds good, I can rewrite it with Vanilla javascript if adding jQuery is not preferred.

jwangace:
Updates on the PR:
- Replace JQuery with Vanilla javascript
- Set default as 0 and not refreshing
- Update *.txt files for flags

frouioui (approved):
LGTM

GuptaManan100 (approved):
Me and @frouioui like the changes and they look good to us! We have decided to go ahead and merge this PR. If @dbussink has any comments or changes to request, we can always address them in a follow up PR, especially since we are very early in the v17 release cycle.

jwangace:
> Me and @frouioui like the changes and they look good to us! We have decided to go ahead and merge this PR. If @dbussink has any comments or changes to request, we can always address them in a follow up PR, especially since we are very early in the v17 release cycle.

Good news and thank you all for your assistance, it's also a good chance for me to learn the process of working with Vitess! 
Just have a question regarding that failed End2End test, it seems that adding tags requires particular order? if so I will update them.

frouioui:
> Good news and thank you all for your help, it's also a good experience for me to learn the process of working with Vitess, and had a taste of how to work on creating PRs, etc.

Thank you for your contribution! ❤️ 

> Just have a question regarding that failed End2End test, it seems that adding tags requires particular order? if so I will update them.

That's correct, the output `.txt` file needs to be in the same order as if you were running `<binary> --help`.

frouioui:
Hey @jwangace, there are two workflows that are stuck:
- Run endtoend tests on Cluster (tabletmanager_tablegc) mysql57
- Run endtoend tests on Cluster (vtorc) mysql57

Do you mind force pushing the same commit again to re-trigger the CI?

jwangace:
Hey @frouioui I checked 3 failed ones, and it seems not related, like one of them:
```
log: exiting because of error: write /tmp/plannedreparent.test.fv-az577-265.runner.log.ERROR.20230221-075853.60932: no space left on device
```
let me know anything I can help.

frouioui:
@jwangace, our CI is highly flaky and there are some fixes currently in progress. I merged your PR, thank you for your contribution ❤️

Issue #12121, opened 2023-01-19
Bug Report: vitessdriver does not correctly INSERT NULL []byte values


simon-engledew:
### Overview of the Issue

Using the Golang vitess database driver (`vitess.io/vitess/go/vt/vitessdriver`) when attempting to insert a row with a `[]byte(nil)` value, `vendor/vitess.io/vitess/go/sqltypes/bind_variables.go` incorrectly inserts the value as an empty string. `''`.

This can be fixed by modifying `bind_variables.go` slightly:

https://github.com/simon-engledew/vitess/blob/6beaa908a547d0b5b39262cfa3a216f6f792f414/go/sqltypes/bind_variables.go#L144-L153

### Reproduction Steps

Create a table with a `VARBINARY DEFAULT NULL`.

Set up a [connection in go](https://pkg.go.dev/vitess.io/vitess/go/vt/vitessdriver) and run:

```
db.Exec("INSERT INTO some_table VALUES (?)", []byte(nil))
```

The value in the database will be `''` instead of `NULL`.

### Binary Version

From go.mod:

```sh
vitess.io/vitess v0.15.2
```

### Operating System and Environment details

```sh
Go version 1.19
```

harshit-gangal:
Could you create an upstream PR for this change with a unit test?

simon-engledew:
Sure! Here is a unit test and a fix:

https://github.com/vitessio/vitess/pull/12363

simon-engledew:
It looks like this behaviour is intentional! 🙇‍♂️ Thanks for looking into the issue. 😂

PR #12363, opened 2023-02-14
fix(vitessdriver): correctly handle nil bytes

simon-engledew:
## Description

This fixes an issue with the vitess driver where it incorrectly deserializes NULL as []bytes{}.

## Related Issue(s)

https://github.com/vitessio/vitess/issues/12121

## Checklist

### DIFF
--- a/go/sqltypes/bind_variables.go
+++ b/go/sqltypes/bind_variables.go
@@ -132,6 +132,9 @@ func StringBindVariable(v string) *querypb.BindVariable {
 
 // BytesBindVariable converts a []byte to a bind var.
 func BytesBindVariable(v []byte) *querypb.BindVariable {
+	if v == nil {
+		return NullBindVariable
+	}
 	return &querypb.BindVariable{Type: VarBinary, Value: v}
 }
 
--- a/go/vt/vitessdriver/driver_test.go
+++ b/go/vt/vitessdriver/driver_test.go
@@ -549,6 +549,43 @@ func TestDatetimeQuery(t *testing.T) {
 	}
 }
 
+func TestEmptyValues(t *testing.T) {
+	var testcases = []struct {
+		desc   string
+		in     []driver.NamedValue
+		out    map[string]*querypb.BindVariable
+		outErr string
+	}{{
+		desc: "empty bytes",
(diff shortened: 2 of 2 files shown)


frouioui:
Added the `Skip CI` label to not overload the CI until this PR is ready for review.

vmg:
I'm not sure these are the semantics we want. This is clearly a case of Go's type system being a bit rough around the edges (very similar things happen when comparing interfaces with nil), but in Go, a `nil` byte slice and an empty byte slice _are the same_. This is, `bytes.Equals([]byte(nil), []byte("")) == true`. Having this distinction in Vitess' bind variable code seems like it will end up generating more bugs than it fixes. If you need a SQL NULL, you should bind a Go `nil`. If you pass in either `[]byte(nil)` or `[]byte("")`, it's much more reasonable to expect an empty `VARBINARY`.

My two cents. I don't think there's a right or wrong answer here but I think:

1. The behavior of nil slices in Go (as a language) is pretty shitty
2. The behavior that our bind variables have right now is _intuitive_ and in accordance with the way the language behaves.

So I think we should just maintain the existing behavior.

simon-engledew:
@vmg sorry, the `[]byte(nil)` is a red herring - I've updated the tests.

The problem is that the driver is incorrectly reading a `NULL` NULLABLE column as an empty byte slice. This behaviour is different to most sql drivers.

vmg:
@simon-engledew: I think there's some confusion here... If you change the value to `nil` in the test, the `if v == nil` check you added to `BytesBindVariable` is never reached. You can verify that with a debugger locally!

Can you double check? The only way to trigger your check in `BytesBindVariable` is with `[]byte(nil)`.

simon-engledew:
Ah yes, my mistake! ❤️ I see what you are saying, sorry. 🙇‍♂️

That makes sense. 🙌  The codebase I was trying to switch to the driver uses Gorm V1, so what is bound is outside of my control. The test suite fails spectacularly without this change and (mostly) passes without it.

Given I hit a couple more issues with the driver after tweaking this I can just stick with the mysql protocol. 🤔

simon-engledew:
Thanks very much for taking the time to look into my issue! ❤️

PR #12274, opened 2023-02-08
evalengine: it's time for JSON!

vmg:
## Description

Yeah! Took a lot of cleanups, but we're finally here. First phase of the JSON implementation for our evalengine.

This PR includes support for:

- Handling the `JSON` data type
- `JSON_EXTRACT()`
- `JSON_UNQUOTE()`
- the `->` and `->>` operators
- `JSON_OBJECT()`
- `JSON_ARRAY()`
- `CAST(native_mysql_type as JSON)` and its reverse `CAST(json_object AS native_mysql_type)`
- `TO_BASE64()` and `FROM_BASE64()` (bonus!)
- `JSON_DEPTH()`
- `JSON_LENGTH()`

These are most of the read-only JSON operations which I expect to see in user queries. JSON transformation operators are supported in the underlying JSON library but I haven't wired them up yet.

Some implementation notes:

Using `encoding/json` for this was a non-starter, because it doesn't have the same semantics as MySQL's JSON implementation and it has terrible ergonomics and performance for dealing with unstructured/unknown data. To make this work, I used a modified version of [`fastjson`](https://github.com/valyala/fastjson) (MIT licensed) which I vendored in `internal/json` like we did for the `decimal` library. The performance of this is pretty good and it provides an abstract syntax tree to manipulate the JSON quite efficiently. I'm pleased about carrying the dependency.

The modifications that the library required are all related to the way MySQL handles JSON object keys: most notably, it _keeps them sorted lexicographically_ (to speed up insertions/removals without having to use a map) and, for duplicate keys, it keeps the last one seen for the object.

On top of this, many of the JSON APIs in MySQL have JSON Path arguments (including the core `->` operator), and by JSON Path I mean, _not_ JSON Path. It's something else. It superficially looks like JSON path but it has nothing to do with it. I had to cook a custom parser based on the grammar declared in the documentation (https://dev.mysql.com/doc/refman/8.0/en/json.html#json-path-syntax), which came up looking nice and tidy because I've somehow spent a lot of time writing parsers lately. It also has the benefit of integrating tightly with our JSON parser implementation for very efficient matching.

## Checklist

-   [ ] "Backport to:" labels have been added if this change should be back-ported
-   [ ] Tests were added or are not required
-   [ ] Documentation was added or is not required

PR #12369, opened 2023-02-14
evalengine: virtual machine

vmg:
## Description

In case you missed it, I introduced a new evaluation framework (in https://github.com/vitessio/vitess/pull/12247) last week -- what could be called `evalengine` v3. That re-engineering had multiple goals, namely:

- Improve the correctness of the way perform evaluations by making them non-lazy
- Improve the performance of AST-based evaluation
- Allow us to implement custom data types to enable our new JSON implementation (done in https://github.com/vitessio/vitess/pull/12274 !)
- Finally, enabling a more static type-checking pass of the full SQL AST that bring us closer to compilation (this PR!)

So, after two weeks of preparation work, I'm ready to show my big plan for `evalengine` performance.

The goal of this PR is implementing a Virtual Machine that can execute SQL expressions inside of Vitess in a very efficient way.

For those new to programming language design, there are roughly 3 ways to execute a _dynamic_ language at runtime. In increasing level of complexity _and_ performance:

1. An AST-based evaluator, where the syntax of the language is parsed into an AST and evaluation is performed by recursively walking each node of the AST and computing the results. _(this is the way the `evalengine` works right now!)_
2. A bytecode VM, where the AST is compiled into binary bytecode that can be evaluated by a virtual machine -- a piece of code that simulates a CPU, but with higher-level instructions. _(this is what we're trying to do here!)_
3. A JIT compiler, in which the bytecode is compiled directly into the host platform's native instructions, so it can be executed directly by the CPU without being interpreted by a Virtual Machine. _(we'll talk about this later!)_

Now you're probably thinking: does this make sense performance-wise? OK, maybe you're not thinking that. It's a rhetorical question with the aim is explaining the following intuition: SQL expressions are incredibly dynamic (when it comes to typing), very high level (when it comes to each primitive operation), and with very little control flow (when it comes to evaluation -- SQL expressions don't really loop, and conditionals are rare; their flow is always lineal!). This can lead us to believe that there's no performance to be squeezed from translating our AST-based evaluation engine into bytecode. The AST is already well suited for high level operations and type-switching!

This is only _superficially_ true. Lots of programming languages are highly dynamic and they manage to run in bytecode VMs much more efficiently than with an AST interpreter (Ruby's transition from its original AST interpreter in MRI to YARV comes to mind). What's the secret here?

Mostly, the secret is [Efficient Interpretation using Quickening](https://publications.sba-research.org/publications/dls10.pdf) (_Stefan Brunthaler_) and variations of it. The idea is dynamic code is very hard to execute efficiently, and the way to optimize it in practice is rewriting the bytecode from more generic instructions (e.g. a sum operator that needs to figure out the types of the two operands to know how to sum them) into specific static instructions which are specialized for the types they operate on (e.g. a sum operator that knows that both operands are integers).

To do that, a quickening VM needs to figure out _at runtime_ the types of the expressions being evaluated and incrementally rewriting the bytecode into instructions that operate on them. This is hard! But we can take this idea even further, and make it more performant and _simpler_: as of https://github.com/vitessio/vitess/pull/12247, our evaluation engine knows how to deterministically type-check any SQL expression based on the types of its input! See where are we going with this? All we need are the _fields_ of the underlying SQL database, and **we get to compile any SQL expression into a highly specialized static form which doesn't need to type-switch on any of its arguments**.

So that's what we're doing on this PR: we convert `evalengine` AST expressions into _statically typed bytecode_, which can perform the same evaluation as the AST but knowing all the types ahead of time. Yes, this results in very fast code. Graph time!

image

<details>
<summary>Raw benchmark data</summary>

```
goos: linux
goarch: amd64
pkg: vitess.io/vitess/go/vt/vtgate/evalengine
cpu: AMD Ryzen 7 2700X Eight-Core Processor         
BenchmarkCompilerExpressions/complex_arith/eval=ast-16           1407079               859.3 ns/op            80 B/op          9 allocs/op
BenchmarkCompilerExpressions/complex_arith/eval=ast-16           1360402               861.7 ns/op            80 B/op          9 allocs/op
BenchmarkCompilerExpressions/complex_arith/eval=ast-16           1367221               874.0 ns/op            80 B/op          9 allocs/op
BenchmarkCompilerExpressions/complex_arith/eval=vm-16            7015935               154.5 ns/op             0 B/op          0 allocs/op
BenchmarkCompilerExpressions/complex_arith/eval=vm-16            7830362               165.2 ns/op             0 B/op          0 allocs/op
BenchmarkCompilerExpressions/complex_arith/eval=vm-16            7685808               150.4 ns/op             0 B/op          0 allocs/op
BenchmarkCompilerExpressions/comparison_i64/eval=ast-16         10571341               123.2 ns/op             8 B/op          1 allocs/op
BenchmarkCompilerExpressions/comparison_i64/eval=ast-16          9295422               117.6 ns/op             8 B/op          1 allocs/op
BenchmarkCompilerExpressions/comparison_i64/eval=ast-16         10396491               122.7 ns/op             8 B/op          1 allocs/op
BenchmarkCompilerExpressions/comparison_i64/eval=vm-16          29770506                36.60 ns/op            0 B/op          0 allocs/op
BenchmarkCompilerExpressions/comparison_i64/eval=vm-16          34390213                35.68 ns/op            0 B/op          0 allocs/op
BenchmarkCompilerExpressions/comparison_i64/eval=vm-16          34351640                60.08 ns/op            0 B/op          0 allocs/op
BenchmarkCompilerExpressions/comparison_u64/eval=ast-16          9401823               126.0 ns/op            16 B/op          1 allocs/op
BenchmarkCompilerExpressions/comparison_u64/eval=ast-16         10032682               126.6 ns/op            16 B/op          1 allocs/op
BenchmarkCompilerExpressions/comparison_u64/eval=ast-16          9627816               125.8 ns/op            16 B/op          1 allocs/op
BenchmarkCompilerExpressions/comparison_u64/eval=vm-16          33149958                35.51 ns/op            0 B/op          0 allocs/op
BenchmarkCompilerExpressions/comparison_u64/eval=vm-16          33283934                34.86 ns/op            0 B/op          0 allocs/op
BenchmarkCompilerExpressions/comparison_u64/eval=vm-16          34757577                35.18 ns/op            0 B/op          0 allocs/op
BenchmarkCompilerExpressions/comparison_dec/eval=ast-16          3904896               319.5 ns/op            64 B/op          3 allocs/op
BenchmarkCompilerExpressions/comparison_dec/eval=ast-16          3724894               315.4 ns/op            64 B/op          3 allocs/op
BenchmarkCompilerExpressions/comparison_dec/eval=ast-16          3638520               320.2 ns/op            64 B/op          3 allocs/op
BenchmarkCompilerExpressions/comparison_dec/eval=vm-16           4912896               230.3 ns/op            40 B/op          2 allocs/op
BenchmarkCompilerExpressions/comparison_dec/eval=vm-16           4803223               232.7 ns/op            40 B/op          2 allocs/op
BenchmarkCompilerExpressions/comparison_dec/eval=vm-16           4843416               232.4 ns/op            40 B/op          2 allocs/op
BenchmarkCompilerExpressions/comparison_f/eval=ast-16            4004112               290.4 ns/op            16 B/op          2 allocs/op
BenchmarkCompilerExpressions/comparison_f/eval=ast-16            3932679               297.2 ns/op            16 B/op          2 allocs/op
BenchmarkCompilerExpressions/comparison_f/eval=ast-16            4412119               268.7 ns/op            16 B/op          2 allocs/op
BenchmarkCompilerExpressions/comparison_f/eval=vm-16            16343871                63.32 ns/op            0 B/op          0 allocs/op
BenchmarkCompilerExpressions/comparison_f/eval=vm-16            18287685                61.96 ns/op            0 B/op          0 allocs/op
BenchmarkCompilerExpressions/comparison_f/eval=vm-16            19428633                66.09 ns/op            0 B/op          0 allocs/op
PASS
ok      vitess.io/vitess/go/vt/vtgate/evalengine        51.304s
```
</details>

Here we have a performance comparison of 5 different queries (ranging from very complex to very simple) between three implementations:

1. **old**, which is the original `evalengine` before https://github.com/vitessio/vitess/pull/12247 was merged
2. **ast**, which is the `evalengine` as of right now (yes I did a very good job optimizing the AST evaluator, thanks for the kind feedback!)
3. **vm**, which is the result of this PR.

The results are stark: the pre-compiled SQL expressions when ran in the VM are up to 20x times faster than the original code, and most interestingly: as a side effect of the static typing, evaluating expressions in the VM **does not allocate memory**.

### An efficient and maintainable Virtual Machine in Go

Implementing a VM usually involves a lot of complexity. You have to write a compiler that processes the input expression AST and generates the corresponding binary instructions (you have to come up with an encoding even!) and _afterwards_ you have to implement the actual VM, which decodes each instruction and performs the corresponding operation. And you have to constantly keep these in sync!

Historically, a bytecode VM has always been implemented the same way: a big-ass switch statement. You decode an instruction, and switch on the type to jump to the operation that needs to be performed. This is how (in theory), bytecode VMs beat AST evaluators: because there are no recursive function calls; the program's execution happens linearly via jumps.

This design, however, has always had shortcomings. If you would like an in-depth explanation of these shortcomings, Mike Pall, the author of LuaJIT, elaborates more on [this ML post](http://lua-users.org/lists/lua-l/2011-02/msg00742.html). But allow me to summarize: Besides the fact that the VM's instructions need to be kept in-sync with the compiler, the actual performance of this main VM loop is not great in practice because compilers usually struggle when compiling _massive functions_. They spill registers all over the place on each arm of the switch, because it's hard to tell which arms are hot and which ones are cold. With all the pushing and popping, the jump into the switch's arm often looks more like a function call! 

...And this applies to C compilers, by the way. It's safe to assume that these problems are the same when we implement the VM in Go, and it my testing, they are actually _much worse_ because the Go compiler sucks. For starters, most of the time the different arms of the switch statement are jumped to via _binary search_ instead of a jump table. Switch jump table optimization was implemented last year (https://go-review.googlesource.com/c/go/+/357330), but it is _fiddly_, and there's no way to enforce it. You have to tweak the way the VM's instructions are encoded carefully to ensure that you're jumping in the VM's main loop. 

So, if switch-based VM loops are not the state of the art, what _is_ the state of the art for writing fast interpreters in Go? Well, it turns out that there's nobody doing fast interpreters in Go right now (at least nobody I can find). Most of the dynamic languages I've found implemented in Go have terrible performance. So we must innovate!

The most interesting approach for machines implemented in C or C++ is continuation-style evaluation loops, as seen [in this report that implements this technique for parsing Protocol Buffers](https://blog.reverberate.org/2021/04/21/musttail-efficient-interpreters.html). This involves implementing all the opcodes for the VM as freestanding functions that operate on the VM, with the return of the function being _the next step of the computation_. It does sound like something expensive and, huh, recursive, but the trick is that newer versions of LLVM allow us to mark functions as _forcefully_ tail-called (see: https://en.wikipedia.org/wiki/Tail_call), so the resulting code is not recursively calling the VM loop but instead _jumping_ between the operations and using the free-standing functions as an abstraction to control register placement and spillage.

Of course this is not something we can do in Go because, well, the Go compiler is allergic to optimization. It _can_ sometimes emit tail calls, but it needs to be tickled in just the right way, and this is something that we cannot enforce at all in this implementation. This got me thinking: what if we have free-standing functions for each instruction, but instead of tail-calling, we forcefully return control to the evaluation loop after each one? If our compiled bytecode is not bytecode but instead **a slice of function pointers to each instruction**, this has many appealing properties.

1. The VM becomes trivial! It's just 5 lines of code, and it doesn't have to worry about optimizing any large switch statements. It's just repeatedly calling functions one after the other!

```go
func (vm *VirtualMachine) execute(p *Program) (eval, error) {
	code := p.code
	ip := 0

	for ip < len(code) {
		ip += code[ip](vm)
		if vm.err != nil {
			return nil, vm.err
		}
	}
	if vm.sp == 0 {
		return nil, nil
	}
	return vm.stack[vm.sp-1], nil
}
```

2. The compiler becomes trivial too, because there is no bytecode! Instead, the compiler emits the individual instructions directly by pushing "callbacks" into a slice. There are no instruction opcodes to keep track off, no encoding to perform and nothing to keep in sync with the VM! Developing the compiler means developing the VM simultaneously!

```go
func (c *compiler) emitPushNull() {
	c.ins = append(c.ins, func(vm *VirtualMachine) int {
		vm.stack[vm.sp] = nil
		vm.sp++
		return 1
	})
}
```

3. ...but wait, if there's no instruction encoding, then we cannot have instructions with arguments. This is a bit of a showstopper... Except it isn't, because the Go compiler actually supports _closures_! We can emit any instruction we want and the Go compiler will automatically capture its arguments inside the callback pointer. We don't have to think about how to encode our arguments, and in fact, our arguments can be as complex as they need to be: the resulting callback will contain a copy of them created by the Go compiler. It's essentially a poor man's JIT, and it works amazingly well in practice, both performance-wise and for ergonomics. Check out this compiler method that generates an instruction to push a `TEXT` SQL object from the input rows into the stack:

```go
func (c *compiler) emitPushColumn_text(offset int, col collations.TypedCollation) {
	c.ins = append(c.ins, func(vm *VirtualMachine) int {
		vm.stack[vm.sp] = newEvalText(vm.row[offset].Raw(), col)
		vm.sp++
		return 1
	})
}
```

Both the offset in the input `rows` array _and_ the collation for the text are statically baked into the generated instruction!

### Next Steps

Wrapping up: This is a novel design for a VM that is both extremely efficient ("almost JIT" thanks to the Go compiler) and very easy to extend and maintain, coupled with an architecture that uses incremental static typing of SQL to generate very efficient compiled programs that run natively in Go and evaluate arbitrary SQL expressions without dynamic memory allocations.

The code in this PR is still an early work in progress: I've only implemented support for all artithmetic operations (a surprising amount there are!) and all comparison operators -- the bare minimum to be able to run meaningful benchmarks with complex SQL expressions and compare them with the old implementations.

My goal is to incrementally update this PR until the expression coverage is good enough that it makes sense to merge it. Since the compiled expressions can be run transparently instead of a normal AST evaluation, I would like to ship the optimizing compiler as an experimental feature in Vitess 17, to stabilize it into full coverage throughout Vitess 18, and then to eventually remove the existing AST evaluator in Vitess 19.

Amibitious? Maybe! I foresee some roadblocks, particularly regarding de-optimization, but I'll talk about those as I get closer to feature completness.

---

#### Addendum: JIT Compilation

Inquiring minds may be wondering: what's next? Are we doing JIT compilation next? The answer is no. Although this design for a compiler and VM looks like an exceptional starting point for implementing a full JIT compiler _in theory_, in practice the trade-off between optimization and complexity doesn't make sense. JIT compilers are important for programming languages where their bytecode operations can be optimized into a very low level of abstraction (e.g. where an "add" operator only has to perform a native x64 `ADD`). In these cases, the overhead of dispatching instructions becomes so dominant that replacing the VM's loop with a block of JITted code makes a significant performance difference. However, for SQL expressions, and even after our quickening pass, most of the operations remain extremely high level (things like "match this JSON object with a path" or "add two fixed-width decimals together"). The overhead of instruction dispatch, as measured in these benchmarks, is less than 20% (and can possibly be optimized further in the VM's loop). 20% is not the number you're targetting before you start fucking around with raw assembly for a JIT. So at this point my intuition is that JIT compilation would be a needlessly complex dead optimization.

## Related Issue(s)

Also implements a number of missing functions from the evalengine listed in https://github.com/vitessio/vitess/issues/9647

## Checklist


go/vt/vtgate/evalengine/arena.go
```diff
@@ -0,0 +1,81 @@
...
+	a.aFloat64 = a.aFloat64[:0]
+	a.aDecimal = a.aDecimal[:0]
+	a.aBytes = a.aBytes[:0]
+}
+
+func (a *Arena) newEvalDecimalWithPrec(dec decimal.Decimal, prec int32) *evalDecimal {
+	if cap(a.aDecimal) > len(a.aDecimal) {
+		a.aDecimal = a.aDecimal[:len(a.aDecimal)+1]
+	} else {
+		a.aDecimal = append(a.aDecimal, evalDecimal{})
```
dbussink, line 25: Ah, nice to just fall back to whatever Go's growth strategy is to not invent our own here.

go/vt/vtgate/evalengine/arithmetic.go
```diff
@@ -35,18 +35,18 @@ func addNumericWithError(left, right eval) (eval, error) {
 	v1, v2 := makeNumericAndPrioritize(left, right)
 	switch v1 := v1.(type) {
 	case *evalInt64:
-		return intPlusIntWithError(v1.i, v2.(*evalInt64).i)
+		return mathAdd_ii(v1.i, v2.(*evalInt64).i)
```
dbussink, line 38: Do we want to use non standard Go function names here with underscores?

go/vt/vtgate/evalengine/arithmetic.go
```diff
@@ -158,120 +148,175 @@ func makeNumericAndPrioritize(left, right eval) (evalNumeric, evalNumeric) {
 	return i1, i2
 }
 
-func intPlusIntWithError(v1, v2 int64) (eval, error) {
+func mathAdd_ii(v1, v2 int64) (eval, error) {
+	result, err := mathAdd_ii0(v1, v2)
+	return newEvalInt64(result), err
```
dbussink, line 153: I guess we're ok with the small difference here that we do return a `newEvalInt64` value on error now (with the value `0`) vs. returning `nil` before? 

It does add an allocation in the error path, but that's then rare enough?

go/vt/vtgate/evalengine/expr_compare.go
```diff
@@ -111,50 +111,41 @@ func (compareNullSafeEQ) compare(left, right eval) (boolean, error) {
...
+func compareAsSameNumericType(l, r sqltypes.Type) bool {
+	if sqltypes.IsIntegral(l) && sqltypes.IsIntegral(r) {
 		return true
 	}
-	if sqltypes.IsFloat(ltype) && sqltypes.IsFloat(rtype) {
+	if sqltypes.IsFloat(l) && sqltypes.IsFloat(r) {
 		return true
 	}
-	if ltype == sqltypes.Decimal && rtype == sqltypes.Decimal {
+	if l == sqltypes.Decimal && r == sqltypes.Decimal {
```
dbussink, line 129: We have `sqltypes.IsDecimal()` these days for consistency :smile:.

go/vt/vtgate/evalengine/expr_compare.go
```diff
@@ -111,50 +111,41 @@ func (compareNullSafeEQ) compare(left, right eval) (boolean, error) {
...
 		return true
 	}
 	return false
 }
 
-func needsDecimalHandling(l, r eval) bool {
-	ltype := l.SQLType()
-	rtype := r.SQLType()
+func compareAsDecimal(ltype, rtype sqltypes.Type) bool {
 	return ltype == sqltypes.Decimal && (sqltypes.IsIntegral(rtype) || sqltypes.IsFloat(rtype)) ||
```
dbussink, line 136: Similar here, we can use `sqltypes.IsDecimal()`.

go/vt/vtgate/evalengine/vm.go
```diff
@@ -0,0 +1,61 @@
+package evalengine
+
+import (
+	"errors"
+
+	"vitess.io/vitess/go/sqltypes"
+)
+
+const MaxStack = 32
```
dbussink, line 9: Is this value based on anything or mostly pulled out of thin air? Dunno how deep the stack can get on evaluating common expressions?

GuptaManan100:
Removing `Benchmark me` label because we want to run the benchmarks for v16.0.0. Please add it again when you want to run benchmarks.

go/vt/vtgate/evalengine/api_types.go
```diff
@@ -176,16 +175,10 @@ func ToNative(v sqltypes.Value) (any, error) {
 	return out, err
 }
 
-func NormalizeValue(v sqltypes.Value, coll collations.ID) string {
+func NormalizeValue(v sqltypes.Value, coll collations.ID) sqltypes.Value {
```
systay, line 178: as far as I can see, only tests use this method. is that intentional?

go/vt/vtgate/evalengine/arena.go
```diff
@@ -0,0 +1,101 @@
+package evalengine
```
systay, line 1: license missing

go/vt/vtgate/evalengine/arena.go
```diff
@@ -0,0 +1,101 @@
...
+
+import (
+	"vitess.io/vitess/go/mysql/collations"
+	"vitess.io/vitess/go/sqltypes"
+	"vitess.io/vitess/go/vt/vtgate/evalengine/internal/decimal"
+)
+
+type Arena struct {
```
systay, line 9: comment please?

go/vt/vtgate/evalengine/api_types.go
dbussink: @systay Yeah, it's a test only thing. I think we should move it then to the test file. Let me do that.

go/vt/vtgate/evalengine/compiler_asm.go
```diff
@@ -0,0 +1,2150 @@
...
+}
+
+func (asm *assembler) emit(f frame, instruction string, args ...any) {
+	if asm.log != nil {
+		asm.log.Instruction(instruction, args...)
+	}
+	asm.ins = append(asm.ins, f)
+}
+
+func (asm *assembler) Add_dd() {
```
systay, line 82: mainly curious - why capitalised method names?

go/vt/vtgate/evalengine/compiler_asm.go
```diff
@@ -0,0 +1,2158 @@
...
+		j.to = len(asm.ins)
+	}
+}
+
+func (asm *assembler) adjustStack(offset int) {
+	asm.stack.cur += offset
+	if asm.stack.cur < 0 {
+		panic("negative stack position")
+	}
+	if asm.stack.cur > asm.stack.max {
```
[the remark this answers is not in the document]
vmg: This can happen any time! This is the function that statically keeps track of how large the stack can get while executing the program. `stack.cur` is the current stack size after this adjustment, and `stack.max` is the largest stack size we've seen so far. If the adjustment has made the stack larger than before, we need to update the new max. It's not an error at all!

systay (approved):
Approving with the understanding that we'll continue working on this and not enable it until we feel safe doing so

dbussink (approved):
I think the baseline we have here now is good. We need to further iterate on this, but right now it's not wired up to anything to execute it and it's easier to iterate further in smaller PRs then. 

So I think we should merge for now.

Issue #12371, opened 2023-02-15
Bug Report: wrong case for sqlparser.CanonicalString and CAST(... AS ...)


shlomi-noach:
### Overview of the Issue

The following query is not normalized correctly in `sqlparser.CanonicalString`:

```
select cast('17' as signed)
```

It currently upcases the type `signed` into `SIGNED` which is technically valid, but inconsistent with MySQL's normalization, which is to always lower case the type.

### Reproduction Steps

-

### Binary Version

```sh
-
```

### Operating System and Environment details

```sh
-
```

PR #12372, opened 2023-02-15
sqlparser/schemadiff: normalize CAST type case

shlomi-noach:
## Description

Fixes https://github.com/vitessio/vitess/issues/12371

Normalizing the `CAST(...AS <type>)` in `CannonicalString()` ; previously uppercased, but that's incompatible with MySQL's behavior. Now keeping the original case.

## Related Issue(s)

- fixes https://github.com/vitessio/vitess/issues/12371
- similar to https://github.com/vitessio/vitess/pull/12105

## Checklist


vmg (approved):

shlomi-noach:
Ignoring `Upgrade Downgrade Testing Query Serving (Schema) ` and `onlineddl_vrepl` CI failures.

PR #12373, opened 2023-02-15
delete deprecated vtctld healthcheck flags and associated realtime stats code

### LINKED ISSUE #11973: Delete deprecated vtctl flags
- [ ] `vtctl_healthcheck_topology_refresh`
- [ ] `vtctl_healthcheck_retry_delay`
- [ ] `vtctl_healthcheck_timeout`

ajm188:
## Description

These flags were deprecated in the previous version, and then I just followed the rabbit trail of compiler errors as I deleted stuff :) 

## Related Issue(s)

- Closes #11973.

## Checklist


frouioui (approved):
LGTM

frouioui:
the reference `.txt`s files used by `TestHelpOutput` will need to be updated

ajm188:
ah shoot, yeah forgot those

go/flags/endtoend/vtctld.txt
```diff
@@ -149,5 +149,4 @@ Usage of vtctld:
       --v Level                                                          log level for V logs
   -v, --version                                                          print binary version
       --vmodule moduleSpec                                               comma-separated list of pattern=N settings for file-filtered logging
-      --vtctl_healthcheck_topology_refresh duration                      refresh interval for re-reading the topology (default 30s)
```
deepthi, line 152: Hmm, I thought deprecated flags should not show up in the help text.

go/vt/vtctld/vtctld.go
```diff
@@ -131,19 +127,8 @@ func InitVtctld(ts *topo.Server) error {
...
-		if err != nil {
-			log.Errorf("Failed to get the list of known cells, failed to instantiate the healthcheck at startup: %v", err)
-		} else {
-			healthCheck = vtctl.NewHealthCheck(ctx, ts, localCell, cells)
-		}
-	}
-
 	// Serve the REST API
-	initAPI(context.Background(), ts, actionRepo, healthCheck)
+	initAPI(context.Background(), ts, actionRepo, nil /* no healthcheck */)
```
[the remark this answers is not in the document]
ajm188: sure, but it involves a ton more cleanup

go/vt/vtctld/api.go
```diff
@@ -490,39 +466,14 @@ func initAPI(ctx context.Context, ts *topo.Server, actions *ActionRepository, he
...
-			if healthcheck == nil {
-				return nil, fmt.Errorf("healthcheck not initialized")
-			}
-
-			heatmap, err := heatmapData(healthcheck, keyspace, cell, tabletType, metric)
-			if err != nil {
-				return nil, fmt.Errorf("couldn't get heatmap data: %v", err)
-			}
-			return heatmap, nil
+			return nil, fmt.Errorf("healthcheck not initialized")
```
arthurschreiber, line 476: 🤔 This change doesn't seem quite correct to me. Does this mean that the `tablet_statuses` endpoint will now always fail with a `healthcheck not initialized` failure?

go/vt/vtctld/api.go
```diff
@@ -537,26 +488,7 @@ func initAPI(ctx context.Context, ts *topo.Server, actions *ActionRepository, he
...
-		tabletAlias := topodatapb.TabletAlias{
-			Cell: cell,
-			Uid:  uid,
-		}
-		tabletStat, err := healthcheck.GetTabletHealthByAlias(&tabletAlias)
-		if err != nil {
-			return nil, fmt.Errorf("could not get tabletStats: %v", err)
-		}
-		return tabletStat, nil
+		return nil, fmt.Errorf("healthcheck not initialized")
```
arthurschreiber, line 491: Same here. If the endpoint is always returning an error, why would we even need the endpoint still? Can't we just remove it and have it return a 404 instead?

go/vt/vtctld/api.go
ajm188: almost certainly yes; but, technically, it can also return with a "invalid tablet type" failure, which is a slightly different signal to the end user

ajm188: that said, that really only happens due to user input, so i'd be totally fine with the argument "yeah, but a user shouldn't do that so there's functionally no difference if we delete the endpoint" and immediately 404-ing

go/vt/vtctld/api.go
ajm188: sure, we can do that instead. @deepthi wdyt?

deepthi: 404 sounds good to me

ajm188: this meant i had to change the function so i'd have access to the `ResponseWriter` to properly do the 404, but other than that, should be straightforward

deepthi (approved):

PR #12375, opened 2023-02-15
Schema RPC to fetch table/view definition

harshit-gangal:
## Description

This PR adds a new RPC to the vttablet query service for VTGate to invoke to retrieve the updated view or table definition. 

## Related Issue(s)

- #11559 

## Checklist


go/vt/vttablet/tabletserver/query_executor.go
```diff
@@ -1227,10 +1227,63 @@ func (qre *QueryExecutor) recordUserQuery(queryType string, duration int64) {
...
-func (qre *QueryExecutor) generateBindVarsForViewDDLInsert(createView *sqlparser.CreateView) map[string]*querypb.BindVariable {
+func generateBindVarsForViewDDLInsert(createView *sqlparser.CreateView) map[string]*querypb.BindVariable {
 	bindVars := make(map[string]*querypb.BindVariable)
 	bindVars["table_name"] = sqltypes.StringBindVariable(createView.ViewName.Name.String())
 	bindVars["view_definition"] = sqltypes.StringBindVariable(sqlparser.String(createView.Select))
 	bindVars["create_statement"] = sqltypes.StringBindVariable(sqlparser.String(createView))
 	return bindVars
 }
+
+func (qre *QueryExecutor) GetSchemaDef(tableType querypb.TableType, tableNames []string) (map[string]string, error) {
```
deepthi, line 1238:
```suggestion
func (qre *QueryExecutor) GetSchemaDefinitions(tableType querypb.TableType, tableNames []string) (map[string]string, error) {
```

go/vt/vttablet/tabletserver/query_executor.go
```diff
@@ -1227,10 +1227,63 @@ func (qre *QueryExecutor) recordUserQuery(queryType string, duration int64) {
...
+	switch tableType {
+	case querypb.TableType_VIEWS:
+		return qre.getViewDef(tableNames)
+	case querypb.TableType_TABLES:
+		return nil, vterrors.Errorf(vtrpcpb.Code_INVALID_ARGUMENT, "invalid table type %v", tableType)
+	}
+	return nil, vterrors.Errorf(vtrpcpb.Code_INVALID_ARGUMENT, "invalid table type %v", tableType)
+}
+
+func (qre *QueryExecutor) getViewDef(tableNames []string) (map[string]string, error) {
```
deepthi, line 1248:
```suggestion
func (qre *QueryExecutor) getViewDefinitions(tableNames []string) (map[string]string, error) {
```

deepthi: shouldn't it be `viewNames` instead of `tableNames`?

proto/query.proto
```diff
@@ -973,3 +973,27 @@ message TransactionMetadata {
   int64 time_created = 3;
   repeated Target participants = 4;
 }
+
+
+// TableType represents the type of table requested.
+enum TableType {
+  VIEWS = 0;
+  TABLES = 1;
+  ALL_TABLES_VIEWS = 2;
```
deepthi, line 982: can be just `ALL`?
that way if we add any other type of objects in the future, the name doesn't become obsolete.

deepthi (commented):
This comment in schema.go needs to be updated
```
// This file provides a few utility variables and methods, mostly for tests.
// The assumptions made about the types of fields and data returned
// by MySQl are validated in schema_test.go. This way all tests
// can use these variables and methods to simulate a MySQL server
// (using fakesqldb/ package for instance) and still be guaranteed correct
// data.
```
We are now using those queries in production, not just tests.

go/vt/vttablet/tabletserver/query_executor.go
harshit-gangal: everything is a table only the type is different. `VIEW / BASE TABLE` i.e. how MySQL differentiates them.

go/vt/vtgate/schema/tracker.go
```diff
@@ -300,21 +300,11 @@ func (t *Tracker) updateTables(keyspace string, res *sqltypes.Result) {
...
-	res, err := th.Conn.Execute(t.ctx, th.Target, mysql.FetchUpdatedViews, bv, 0, 0, nil)
+	res, err := th.Conn.GetSchema(t.ctx, th.Target, querypb.SchemaTableType_VIEWS, viewsUpdated)
 	if err != nil {
 		t.tracked[th.Target.Keyspace].setLoaded(false)
 		// TODO: optimize for the views that got errored out.
 		log.Warningf("error fetching new views definition for %v", viewsUpdated, err)
-		code := vterrors.Code(err)
-		if code == vtrpcpb.Code_UNAUTHENTICATED || code == vtrpcpb.Code_PERMISSION_DENIED {
-			log.Warning(aclErrorMessageLog)
-		}
```
frouioui, line 317: What's the reason for removing this?

go/vt/vttablet/endtoend/views_test.go
```diff
@@ -145,7 +148,30 @@ func TestDropViewDDL(t *testing.T) {
...
+	viewSchemaDef, err := client.GetSchema(querypb.SchemaTableType_VIEWS)
+	require.NoError(t, err)
+	require.Zero(t, len(viewSchemaDef))
+
+	client.UpdateContext(callerid.NewContext(
+		context.Background(),
+		&vtrpcpb.CallerID{},
+		&querypb.VTGateCallerID{Username: "dev"}))
+
+	defer client.Execute(qDelAllRows, nil)
```
frouioui, line 169: nit: shoudn't we do this after `client.Execute` and before the `require.NoError`

go/vt/vtgate/schema/tracker.go
harshit-gangal: GetSchema does not go through ACL check, so tracker will not see such error message.

frouioui: Got it

frouioui (approved):

GuptaManan100 (approved):

PR #12381, opened 2023-02-16
Fix `SHOW VSCHEMA TABLES` tests using v17 vtgate that expected `dual`

frouioui:
## Description

This PR is a follow up of the work done in https://github.com/vitessio/vitess/pull/12292. We removed the `dual` table from the output of `SHOW VSCHEMA TABLES`, this PR updates the tests that are used in upgrade/downgrade accordingly.

## Checklist


harshit-gangal (commented):
I do not think it is important to run this test in upgrade downgrade.
changing he expectation based on version is basically against the idea of having an upgrade downgrade test.

harshit-gangal (approved):

systay (approved):

PR #12385, opened 2023-02-17
go/vt/mysqlctl{/backupstats}: throttle backupstats for perf

maxenglander:
## Description

https://github.com/vitessio/vitess/pull/11979 added detailed backup/restore metrics, but appeared to add ~4% overhead in a local test. I'm no longer able to reproduce the findings of that test.

In spite of that, this PR attempts to reduce any overhead by throttling stats calls. 

Takes inspiration from [logutil.ThrottledLogger](https://github.com/vitessio/vitess/blob/d90c6f54b7c4349bc78302bc2f49202e547e2491/go/vt/logutil/throttled.go#L29), but does not drop any stats. Instead batches and and combines unflushed stats and then flushes them either when `Flush()` is called by the user, or a subsequent stats call occurs after a user-specified time interval.

## Performance

A repeat of the performance test in https://github.com/vitessio/vitess/pull/11979.

I set up the commerce example cluster, and created a table with ~30 GiB of data  by repeatedly inserting `REPEAT(UUID(),10)`, and then ran...

```
$ vtctlclient Backup -- --allow_primary <tablet>
```

...multiple times, collecting the value of `vttablet_backup_duration_seconds`. 

Did this with the `main` commit before backupstats was introduced, the `main` commit when backupstats  was introduced, and the `HEAD` of this PR.

| # | `main` (7fc1b48d) | `main` (d5364e61) | this PR (65ac8fe2) |
|-|-|-|-|
| 1 | 89 | 90 | 89 |
| 2 | 89 | 90 | 89 |
| 3 | 90 | 89 | 88 | 
| 4 | 91 | 89 | 90 |
| 5 | 89 | 89 | 91 | 
| Δ | N/A | -0.22% | -0.22% |

In the last PR I found that d5364e61 added a bit of overhead 7fc1b48d.

> adds a roughly 3.8% performance overhead.

In this PR, it's not evident that d5364e61 adds any overhead. It's also not evident that 65ac8fe2 is any improvement to d5364e61.

I'm not sure it makes sense to merge this code, since it's not clear that there's anything to fix.

One possible explanation for getting misleading results in the last PR. I used `BackupShard` in the last PR whereas I'm using `Backup` against a specific tablet in this PR. Correct me if I'm wrong but I think that `BackupShard` picks a random replica tablet. Maybe small differences between replicas could result in different performance results from runs against randomly selected replicas.

## Related Issue(s)

https://github.com/vitessio/vitess/issues/11977
https://github.com/vitessio/vitess/pull/11979

## Checklist

-   [x] Documentation ~was added or~ is not required

### DIFF
--- a/go/vt/mysqlctl/backup.go
+++ b/go/vt/mysqlctl/backup.go
@@ -85,6 +85,10 @@ var (
 	backupCompressBlocks = 2
 
 	titleCase = cases.Title(language.English).String
+
+	// backupStatsEmitInterval is the maximum frequency at which backup stats
+	// are emitted.
+	backupStatsEmitInterval = 10 * time.Second
 )
 
 func init() {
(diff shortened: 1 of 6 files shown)


deepthi:
If there is no difference in performance, I'd like us to hold off on merging this PR.
Maybe when you implement the same metrics for s3, you can re-do the benchmarks, and let's see if this makes a difference?

maxenglander:
Returning this PR to draft until we have compelling evidence that it's needed.

PR #12387, opened 2023-02-17
Skip `TestReparentDoesntHangIfPrimaryFails` in vttablet v16 and above

frouioui:
## Description
This PR is a follow up of the recent work done by @rohit-nayak-ps around vttablet sidecar (https://github.com/vitessio/vitess/pull/11520). In the latter work, we removed `Create`/`Alter` `ReparentJournal`, this PR fixes a test that uses this SQL query. The test will not be ran during upgrade tests when vttablet is on version 16.

## Related Issue(s)

- Follow up of #11520

## Checklist


GuptaManan100 (commented):
Why is this required?

GuptaManan100 (approved):
Nevermind, I just read the description

dbussink (approved):

PR #12389, opened 2023-02-17
Add CpuUsage to RealtimeStats

timvaillancourt:
## Description

This PR adds optional support for setting the existing, but unused(?) [`CpuUsage` field in `RealtimeStats`](https://github.com/vitessio/vitess/blob/main/go/vt/proto/query/query.pb.go#L4987-L4988).

This support is enabled via the new `vttablet` flag `--enable-system-health-monitor` and is implemented as a generic monitor (`systemHealthCollector`) for system health metrics, currently CPU usage only.

Unfortunately this adds a new dependency [`github.com/shirou/gopsutil/cpu`](https://github.com/shirou/gopsutil/cpu) _([BSD License](https://github.com/shirou/gopsutil/blob/master/LICENSE))_ as a cross-platform _(Linux, BSDs, Windows)_ library for gathering CPU usage - I hope that is not a problem.

The metric collection is on a fixed 1000ms ticker - please let me know if this should be configurable. The collection runs on "serving" nodes only. 

 Code style, naming, etc suggestions/nits are welcome! 🙇 

## Related Issue(s)

Feature Request: https://github.com/vitessio/vitess/issues/12579

## Checklist

-   [ ] Testing on slackhq/vitess fork


deepthi:
this should be reviewed by @vitessio/query-serving. unfortunately I can't remove review requests from code owners. maybe time to refine that file..

timvaillancourt:
> this should be reviewed by @vitessio/query-serving

Thanks! I'll setup a website/docs PR once the approach once we have agreement on naming/etc 👍

go.mod
```diff
@@ -112,6 +112,7 @@ require (
 	github.com/kr/pretty v0.3.1
 	github.com/kr/text v0.2.0
 	github.com/nsf/jsondiff v0.0.0-20210926074059-1e845ec5d249
+	github.com/shirou/gopsutil v3.21.11+incompatible
```
dbussink, line 115: Any reason for not using the latest version here? Looks like this one is from late 2021 and there's a bunch more recent ones.

go/vt/vttablet/tabletserver/system_health.go
```diff
@@ -0,0 +1,140 @@
...
+*/
+
+package tabletserver
+
+import (
+	"context"
+	"sync"
+	"time"
+
+	"github.com/shirou/gopsutil/cpu"
```
dbussink, line 24: I think we want to use `github.com/shirou/gopsutil/v3/cpu` here to use the `v3` API? See https://github.com/shirou/gopsutil#v3-migration

go.mod
dbussink: The import likely needs to be changed to the following for this (and to also fix the other comment):

```
github.com/shirou/gopsutil/v3 v3.23.1
```

go/vt/vttablet/tabletserver/system_health.go
```diff
@@ -0,0 +1,140 @@
...
+const (
+	systemHealthMonitorCPUSampleWindow = time.Millisecond * 200
+	systemHealthMonitorInterval        = time.Millisecond * 1000
+)
+
+// systemHealthCollector is a collector for system health.
+type systemHealthCollector struct {
+	config          *tabletenv.TabletConfig
+	cpuSampleWindow time.Duration
+	cpuUsage        sync2.AtomicFloat64
```
dbussink, line 40: I think we should move away from using `sync2` now that Go has built in easier to use atomics. We also should switch all existing `sync2` usage but that's a separate change then.

go.mod
timvaillancourt: @dbussink no good reason, this is what `go mod tidy` chose maybe due to not supporting version subdirs

Updating to the version suggested 👍

go/vt/vttablet/tabletserver/system_health.go
timvaillancourt: @dbussink ahh ok, I thought maybe `sync2` was doing something special 😄

Would you prefer plain `sync/atomic` to be used here?

harshit-gangal:
@timvaillancourt The use case for sending CPU usage is not provided. Also, I could not find an issue linked to this PR to understand the same. 
Kindly provide the same to know the reason for the change.

go/vt/vttablet/tabletserver/system_health.go
dbussink: `sync2` is something that has existed since before Go 1.19 came out which added much nicer usable atomics compared to what the Go provided before. See https://golang.org/doc/go1.19#atomic_types

So we haven't done the work yet to migrate, but I've opened #12391 to see what the fallout of that is going to be.

go/vt/vttablet/tabletserver/state_manager.go
```diff
@@ -531,23 +542,25 @@ func (sm *stateManager) unserveCommon() {
...
+	log.Info("Started system health monitor close")
+	sm.systemHealth.Close()
+	log.Info("Finished system health monitor close. Started online ddl executor close")
 	sm.ddle.Close()
-	log.Infof("Finished online ddl executor close. Started table garbage collector close")
+	log.Info("Finished online ddl executor close. Started table garbage collector close")
 	sm.tableGC.Close()
-	log.Infof("Finished table garbage collector close. Started lag throttler close")
+	log.Info("Finished table garbage collector close. Started lag throttler close")
 	sm.throttler.Close()
```
harshit-gangal, line 552:
```go
	sm.systemHealth.Open()
	sm.tableGC.Open()
	sm.ddle.Open()
```
We follow the order of Open and Close.

as open happened before  `tableGC` and `ddle`, the close should happen after `tablegc` and `ddle`

go/vt/vttablet/tabletserver/system_health.go
timvaillancourt: Sounds good 👍. One thing it lacks that `sync2` does well is support for `float64` values 🤔

dbussink: > Sounds good 👍. One thing it lacks that `sync2` does well is support for `float64` values 🤔

I wouldn't really classify it as `well` though, since there's a reason why Go doesn't have that, see https://github.com/golang/go/issues/21996.

But you can use the  same trick as for the single other float64 case I found in https://github.com/vitessio/vitess/pull/12391.

go/vt/vttablet/tabletserver/system_health_test.go
```diff
@@ -0,0 +1,74 @@
...
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+
+	"vitess.io/vitess/go/vt/vttablet/tabletserver/tabletenv"
+)
+
+func TestSystemHealthCollector(t *testing.T) {
+	monitor := &systemHealthCollector{
```
frouioui, line 30: We should add a comment here to describe what the test does and what's its expected behavior.

> If a test is added or modified, there should be a documentation on top of the test to explain what the expected behavior is what the test does.

timvaillancourt:
> @timvaillancourt The use case for sending CPU usage is not provided.

@harshit-gangal 👋, I've created this feature request to explain the intent of this PR: https://github.com/vitessio/vitess/issues/12579

timvaillancourt:
Bump to remove stale status. Hoping to fix this PR up soon 👍

deepthi:
@timvaillancourt is this ready for re-review?

timvaillancourt:
> @timvaillancourt is this ready for re-review?

@deepthi yes, I think so, although nothing is ready to use this information yet.

As an update: based on discussions in #feat-handling-overload with a few orgs, we may not use this metric for backpressure now as conn pool capacity has proven to be a more reliable metric than CPU. So this leaves only one future use case, which I believe was the intent of `CpuUsage` existing in `RealtimeStats` in the first place: considering CPU usage in load balancing

Related, in a recent discussion @harshit-gangal brought up a great point regarding how well this metric will work on Kube/Docker-based Vitess deployments. This isn't something we've validated as we don't have Vitess on a Kube platform. If anyone is available to test this on Kube, that would be very helpful. I'll run a few more tests on our EC2 hardware to validate this next week 👍

PR #12391, opened 2023-02-20
Move to Go 1.19 atomics

### LINKED ISSUE #12395: Move to Go 1.19 atomics
Vitess has had a `sync2` package for a while with easier to use atomics. Since Go 1.19, the standard library also provides easier to use atomics. 

We should switch to using the Go 1.19 version so that we can clean up the code. The Go standard library code also has additional benefits such as giving guarantees for proper alignment across platforms.

Additionally, there's more things we can move such as using the semaphore from the `x/sync` package in favor of our own hand crafted one.

dbussink:
This change moves to Go 1.19 atomics in favor of the version we had in the `sync2` package. The version in that package predates the availability of the new atomics, but for the future we want to use the Go provided ones.

On top of that, we can also use the semaphore from `x/sync` instead of a custom one in `sync2`. This cleans up a bunch of additional code from `sync2`.

Upsides are removing the mostly duplicate code from the standard library (albeit with slight different name) and that the stdlib implementation also guarantees things we can't, such as proper alignment for the atomic variables.

## Related Issue(s)

Fixes #12395

## Checklist


go/vt/vtgate/buffer/buffer_helper_test.go
```diff
@@ -121,7 +121,7 @@ func waitForState(b *Buffer, want bufferState) error {
 func waitForPoolSlots(b *Buffer, want int) error {
 	start := time.Now()
 	for {
-		got := b.bufferSizeSema.Size()
+		got := b.bufferSize
```
shlomi-noach, line 124: I'm not even sure what the old code means; the semaphore acts both as an atomic primitive _and_ indicates the size?

go/vt/vttablet/tabletmanager/vdiff/engine.go
```diff
@@ -169,12 +168,12 @@ func (vde *Engine) openLocked(ctx context.Context) error {
 	return nil
 }
 
-var openRetryInterval = sync2.NewAtomicDuration(1 * time.Second)
+var openRetryInterval = 1 * time.Second
```
shlomi-noach, line 171: nice

go/vt/vttablet/tabletserver/tabletserver.go
```diff
@@ -1953,32 +1957,32 @@ func (tsv *TabletServer) QueryPlanCacheWait() {
...
 // SetThrottleMetricThreshold changes the throttler metric threshold
 func (tsv *TabletServer) SetThrottleMetricThreshold(val float64) {
-	tsv.lagThrottler.MetricsThreshold.Set(val)
+	tsv.lagThrottler.StoreMetricsThreshold(val)
 }
 
 // ThrottleMetricThreshold returns the throttler metric threshold
 func (tsv *TabletServer) ThrottleMetricThreshold() float64 {
-	return tsv.lagThrottler.MetricsThreshold.Get()
+	return time.Duration(tsv.lagThrottler.MetricsThreshold.Load()).Seconds()
```
shlomi-noach, line 1985: This one is incorrect. By default, the metrics threshold indicates seconds of replication lag. But it is also customizable to indicate any floating point value, per `metrics_query`. So, for example, it could be the number of threads running (an integer threshold), or the number of writes per sec (float).

shlomi-noach: The correct behavior is to just return the value, without casting-to-Duration-and-extracting-Seconds()

shlomi-noach (changes requested):
Very thorough! I have one request for change, see lag throttler. Otherwise this looks good

go/vt/vtgate/buffer/buffer_helper_test.go
dbussink: We had code that indeed checked this size but the `x/sync` Go semaphore doesn't expose that (as it's a total weight, not a real size). 

That's why this had to be added as a separate field. Agreed it's a bit wonky but I didn't want to go and change behavior here.

shlomi-noach (approved):

go/vt/vttablet/tabletserver/tabletserver.go
dbussink: @shlomi-noach Ah ok, it does have to be a float and there's no float atomics (see https://github.com/golang/go/issues/21996), so now it uses the bit conversions. That's also what `sync2` was already doing anyway.

shlomi-noach: right! Sadly we need to do the conversion. Floats in computers are a pain.

go/vt/vtctl/schematools/reload.go
```diff
@@ -67,8 +68,15 @@ func ReloadShard(ctx context.Context, ts *topo.Server, tmc tmclient.TabletManage
 			defer wg.Done()
 
 			if concurrency != nil {
-				concurrency.Acquire()
-				defer concurrency.Release()
+				if concurrency.Acquire(ctx, 1) != nil {
+					// We timed out waiting for the semaphore. This is best-effort, so just log it and move on.
```
frouioui, line 72: Should we print out the error here?

frouioui (approved):
Thank you @dbussink 🙏🏻

go/vt/vtctl/schematools/reload.go
dbussink: We can, I can add it. It would be a context cancelled or expired message always but that might be helpful still?

frouioui: If the error is always the same, then I think the warning log is sufficient as the error itself is not going to bring much more help.

vmg (approved):
LGTM. There's a pice of @shlomi-noach's feedback you haven't addressed yet btw

go/vt/vttablet/tabletserver/tabletserver.go
vmg: This hasn't been addressed yet!

dbussink: Doh, fixed all cases except this one 🤦. Done!

Issue #12392, opened 2023-02-20
Bug Report: Additional incorrect casing for sqlparser.CanonicalString


dbussink:
### Overview of the Issue

Similar to issues like #12371, there's more cases like that. This issue tracks auditing similar cases instead of depending on finding these adhoc.

### Reproduction Steps

```
CAST(`info` -> '$.field' AS unsigned array))
```

right now erroneously upcases `ARRAY` but there's other cases too.

### Binary Version

```sh
-
```

### Operating System and Environment details

```sh
-
```

### Log Fragments

```sh
-
```

dbussink:
- [ ] `CAST(info -> '$.field' AS unsigned array))` upcases `ARRAY` incorrectly.
- [ ] `GROUP_CONCAT(c ORDER BY c ASC SEPARATOR 'foo')` upcases `'FOO`' incorrectly.
- [ ] `SELECT 1 FROM t1 INTO OUTFILE 'test/t1.txt'` upcases `'TEST/T1.TXT'` incorrectly.
- [ ] `SELECT X'4d7953514c'` upcases hex characters incorrectly.
- [ ] `ALTER VSCHEMA CREATE VINDEX lookup_vdx USING lookup WITH owner=user, table=name_user_idx, from=name, to=user_id` upcases the parameters like `OWNER=USER` incorrectly.
- [ ] `ALTER VITESS_MIGRATION '9aecb3b4_b8a9_11ec_929a_0a43f95f28a3' throttle` upcases the UUID and last statement incorrectly.
- [ ] Time functions like `current_timestamp` are upcased incorrectly.

PR #12393, opened 2023-02-20
Fix additional casting bugs in the sqlparser

dbussink:
This fixes a bunch of additional canonical casing bugs in the parser. 

## Related Issue(s)

Fixes https://github.com/vitessio/vitess/issues/12392

## Checklist

### DIFF
--- a/go/vt/sqlparser/ast_format.go
+++ b/go/vt/sqlparser/ast_format.go
@@ -254,7 +254,7 @@ func (node *AlterVschema) Format(buf *TrackedBuffer) {
 func (node *AlterMigration) Format(buf *TrackedBuffer) {
 	buf.astPrintf(node, "alter vitess_migration")
 	if node.UUID != "" {
-		buf.astPrintf(node, " '%s'", node.UUID)
+		buf.astPrintf(node, " '%#s'", node.UUID)
 	}
 	var alterType string
 	switch node.Type {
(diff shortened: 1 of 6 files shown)


go/vt/sqlparser/ast_format.go
```diff
@@ -1512,10 +1512,14 @@ func (node *FuncExpr) Format(buf *TrackedBuffer) {
 // Format formats the node
 func (node *GroupConcatExpr) Format(buf *TrackedBuffer) {
 	if node.Distinct {
-		buf.astPrintf(node, "group_concat(%s%v%v%s%v)", DistinctStr, node.Exprs, node.OrderBy, node.Separator, node.Limit)
+		buf.astPrintf(node, "group_concat(%s%v%v", DistinctStr, node.Exprs, node.OrderBy)
 	} else {
-		buf.astPrintf(node, "group_concat(%v%v%s%v)", node.Exprs, node.OrderBy, node.Separator, node.Limit)
+		buf.astPrintf(node, "group_concat(%v%v", node.Exprs, node.OrderBy)
 	}
+	if node.Separator != "" {
```
dbussink, line 1519: This works because also if the separate is set to `''` explicitly, that means we have that as the encoded string here already so we print it again. 

Before we would treat it combined as one thing, but we'd upcase the whole thing, breaking this.

shlomi-noach (approved):

PR #11356, opened 2022-09-26
VTOrc Cleanup - Configs, APIs and old UI

GuptaManan100:
## Description

Since VTOrc was forked from `Orchestrator`, it inherited a lot of configurations that don't make sense for the Vitess use-case.
All of such configurations have been removed in this PR.

VTOrc ignores the configurations that it doesn't understand. So old configurations can be kept around on upgrading and won't cause any issues.
They will just be ignored.

For all the configurations that are kept, flags have been added for them and the flags are the desired way to pass these configurations going forward.
The config file will be deprecated and removed in upcoming releases. The following is a list of all the configurations that are kept and the associated flags added.

|          Configurations Kept          |           Flags Introduced            |
|:-------------------------------------:|:-------------------------------------:|
|            SQLite3DataFile            |         `--sqlite-data-file`          |
|          InstancePollSeconds          |        `--instance-poll-time`         |
|    SnapshotTopologiesIntervalHours    |    `--snapshot-topology-interval`     |
|    ReasonableReplicationLagSeconds    |    `--reasonable-replication-lag`     |
|             AuditLogFile              |        `--audit-file-location`        |
|             AuditToSyslog             |         `--audit-to-backend`          |
|           AuditToBackendDB            |          `--audit-to-syslog`          |
|            AuditPurgeDays             |       `--audit-purge-duration`        |
|      RecoveryPeriodBlockSeconds       |  `--recovery-period-block-duration`   |
| PreventCrossDataCenterPrimaryFailover |    `--prevent-cross-cell-failover`    |
|        LockShardTimeoutSeconds        |        `--lock-shard-timeout`         |
|      WaitReplicasTimeoutSeconds       |       `--wait-replicas-timeout`       |
|     TopoInformationRefreshSeconds     | `--topo-information-refresh-duration` |
|          RecoveryPollSeconds          |      `--recovery-poll-duration`       |

The config file still takes precedence over the flags that can now be specified.

Apart from configurations, some flags from VTOrc have also been removed -
- `sibling`
- `destination`
- `discovery`
- `skip-unresolve`
- `skip-unresolve-check`
- `noop`
- `binlog`
- `statement`
- `grab-election`
- `promotion-rule`
- `skip-continuous-registration`
- `enable-database-update`
- `ignore-raft-setup`
- `tag`

[...]

PR #12396, opened 2023-02-20
Clean up deprecated flag from test setup

dbussink:
This was missed in https://github.com/vitessio/vitess/pull/11356 where this flag was deprecated and removed from most other places.

## Related Issue(s)

Missed in https://github.com/vitessio/vitess/pull/11356

## Checklist


GuptaManan100 (commented):
It wasn't missed. I had kept it to make sure we keep supporting the flag. But now that v16 release has been branched, I can even remove that flag. I can do these changes all together in a PR. There are other deprecated flags that can be removed too.

PR #12404, opened 2023-02-21
Remove a bunch of debug logging

dbussink:
This removes a bunch of debug logging that is set for higher log levels. The problem is that many of these are logging user provided values directly which is something CodeQL also alerts on.

If someone needs this kind of additional logging level, I think it suffices to modify the code locally to test something rather than always carrying these high log level entries.

## Related Issue(s)

Part of #12216 where CodeQL identified a bunch of user provided log input issues. 

## Checklist


frouioui (approved):

GuptaManan100 (approved):
LGTM

rsajwani (commented):
LGTM

PR #12407, opened 2023-02-21
sqlparser: Tools cleanup

vmg:
## Description

As discussed with @systay, I started working on having all `sqlparser` AST nodes by reference. It's hairy! But I've done some cleanups to the tooling around the parser that make this kind of refactorings easier.

I'd like to merge them straight away because they're generally useful.

- All the `sqlparser` codegen is now done using `go:generate` and centralized in a single file
- All codegen for `sqlparser` and for `sizegen` is now resilient to partial errors. **This is huge!** Now the codegen works even if there are existing errors in the already-generated files (something that happens very often when performing refactorings). This was such a pain in the ass before that made these refactorings really troublesome.
- The AST helper generator can now check that all AST elements are passed by reference. This check is disabled by default because, huh, it doesn't pass yet. :)

## Checklist


go/tools/asthelpergen/asthelpergen.go
```diff
@@ -92,54 +93,41 @@ func newGenerator(mod *packages.Module, sizes types.Sizes, named *types.Named, g
 }
 
 func findImplementations(scope *types.Scope, iff *types.Interface, impl func(types.Type) error) error {
+	const OnlyReferences = false
```
dbussink, line 96: this can be lowercase?

go/tools/common/common.go
```diff
@@ -17,19 +17,38 @@ limitations under the License.
...
+			filePath := e.Pos[:idx]
+			_, fileName := path.Split(filePath)
+			if !skip(fileName) {
+				errors = append(errors, e.Error())
+			}
 		}
 	}
-	return failed
+	if len(errors) > 0 {
+		return fmt.Errorf("found %d error(s) when loading Go packages:\n\t%s", len(errors), strings.Join(errors, "\n\t"))
```
dbussink, line 40: Guess we're almost on Go 1.20 where we could use `errors.Join` instead.  Not as important here though as this is not something where the caller would want to dig into the actual different separate error messages, as it's all printed to the command line for the human running this afaik?

go/vt/sqlparser/ast_format_fast.go
```diff
@@ -19,7 +19,6 @@ package sqlparser
 
 import (
 	"fmt"
-
```
dbussink, line 22: The import formatting might complain about this line removal? We should fix the generation though to create a format that `goimports` thinks is ok.

dbussink (commented):
> All codegen for sqlparser and for sizegen is now resilient to partial errors. This is huge! Now the codegen works even if there are existing errors in the already-generated files (something that happens very often when performing refactorings). This was such a pain in the ass before that made these refactorings really troublesome.

It was also a pain on merge conflicts, so another reason this is great :tada:.

go/vt/sqlparser/ast_format_fast.go
vmg: Ah yes, this is not correct and I'm not sure why it changed. The `static` linter in CI is complaining about it.

dbussink: @vmg I don't think this is something you changed, but an already existing long standing problem with the code generation for this file. I thought it might be easy to fix while you're in here already.

go/tools/asthelpergen/asthelpergen.go
vmg: Constants in Go are always UpperCase or the linter will complain!

dbussink: @vmg Huh? The first letter is about exporting it or not? You can for sure have constants that start with a lowercase?

vmg:
Fixed the spacing issues we were seeing in the output. :ok_hand:

dbussink:
> Fixed the spacing issues we were seeing in the output. 👌

Looks like some stuff was missed?

```
go/tools/sizegen/sizegen.go:520:34: cannot use file (variable of type *jen.File) as string value in argument to codegen.SaveJenFile (typecheck)
[5](https://github.com/vitessio/vitess/actions/runs/4241476823/jobs/7371745246#step:22:6)
			if err := codegen.SaveJenFile(file, fullPath); err != nil {
```

go/tools/asthelpergen/asthelpergen.go
```diff
@@ -92,54 +92,41 @@ func newGenerator(mod *packages.Module, sizes types.Sizes, named *types.Named, g
...
 	for _, name := range scope.Names() {
 		obj := scope.Lookup(name)
 		if _, ok := obj.(*types.TypeName); !ok {
 			continue
 		}
 		baseType := obj.Type()
 		if types.Implements(baseType, iff) {
-			err := impl(baseType)
-			if err != nil {
+			if OnlyReferences {
```
frouioui, line 104: Doesn't this always evaluate to `false` considering `OnlyReferences` is a constant?

frouioui (approved):

go/vt/sqlparser/generate.go
```diff
@@ -0,0 +1,7 @@
+package sqlparser
```
GuptaManan100, line 1: We need the license header

GuptaManan100 (approved):
Everything else LGTM!

go/tools/asthelpergen/asthelpergen.go
GuptaManan100: I think the intent is to make it always `true` after we fix the current structs to only implement the `SQLNode` as pointers.
Currently switching it to `true` would panic the generator.

Makefile
```diff
@@ -159,20 +159,17 @@ install-testing: build
...
 	make -C go/vt/vtctl/vtctldclient
 
-parser:
-	make -C go/vt/sqlparser
+sqlparser:
+	go generate ./go/vt/sqlparser/...
+
+parser: sqlparser
+visitor: sqlparser
+asthelpergen: sqlparser
```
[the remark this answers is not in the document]
GuptaManan100: I personally would prefer to have them around. If someone wants to only generate the rewriter, then they would continue to run `make asthelpergen`. It doesn't seem all that great asking people to run `make sqlparser`. We might generate everything under the hood, but the aliases are still good to have.

go/tools/asthelpergen/asthelpergen.go
vmg: Yes, we can't enable `onlyReferences` until we actually refactor our SQL AST to have only references! The functionality is here so we can check our implementation when it's eventually ready.

Makefile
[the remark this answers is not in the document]
vmg: Sure!

PR #12409, opened 2023-02-21
Create Views allowed for same keyspace

harshit-gangal:
## Description

This PR enables view support on information schema queries by restricting the create views to refer only single keyspace.
The views are now created at the MySQL level so all information schema, show query, describe, and explain query can work.

## Related Issue(s)

- https://github.com/vitessio/vitess/issues/11559

## Checklist

### DIFF
--- a/go/vt/vtgate/planbuilder/testdata/view_cases.json
+++ b/go/vt/vtgate/planbuilder/testdata/view_cases.json
@@ -19,23 +19,9 @@
     }
   },
   {
-    "comment": "create view",
+    "comment": "create view with different keyspaces",
     "query": "create view user.view_a as (select id from unsharded union select id from unsharded_auto) union (select id from unsharded_auto union select name from unsharded)",
-    "plan": {
-      "QueryType": "DDL",
-      "Original": "create view user.view_a as (select id from unsharded union select id from unsharded_auto) union (select id from unsharded_auto union select name from unsharded)",
-      "Instructions": {
-        "OperatorType": "DDL",
-        "Keyspace": {
-          "Name": "user",
-          "Sharded": true
-        },
-        "Query": "create view view_a as select id from unsharded union select id from unsharded_auto union select id from unsharded_auto union select `name` from unsharded"
-      },
-      "TablesUsed": [
-        "user.view_a"
-      ]
-    }
+    "plan": "VT12001: unsupported: Select query does not belong to the same keyspace as the view statement"
   },
   {
     "comment": "create view with authoritative columns",
--- a/go/vt/vttablet/endtoend/views_test.go
+++ b/go/vt/vttablet/endtoend/views_test.go
@@ -31,7 +31,6 @@ import (
 )
 
 var qSelAllRows = "select table_schema, table_name, create_statement from _vt.views"
-var qDelAllRows = "delete from _vt.views"
 
 // Test will validate create view ddls.
 func TestCreateViewDDL(t *testing.T) {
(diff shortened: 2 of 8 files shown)


go/vt/vttablet/endtoend/views_test.go
```diff
@@ -153,6 +152,77 @@ func TestDropViewDDL(t *testing.T) {
 	require.Zero(t, qr.Rows)
 }
 
+// TestViewDDLWithInfrSchema will validate information schema queries with views.
+func TestViewDDLWithInfrSchema(t *testing.T) {
```
frouioui, line 156: We should perhaps add an E2E tests where we try to add a view with two different keyspaces in the select query and look for `VT12001` error

frouioui (approved):

go/vt/vttablet/endtoend/views_test.go
harshit-gangal: we have a unit test in plan_test that covers it.
views_test is vttablet e2e test, the failure will happen in vtgate while planning.

go/vt/vtgate/planbuilder/testdata/view_cases.json
```diff
@@ -19,23 +19,9 @@
...
-          "Name": "user",
-          "Sharded": true
-        },
-        "Query": "create view view_a as select id from unsharded union select id from unsharded_auto union select id from unsharded_auto union select `name` from unsharded"
-      },
-      "TablesUsed": [
-        "user.view_a"
-      ]
-    }
+    "plan": "VT12001: unsupported: Select query does not belong to the same keyspace as the view statement"
```
GuptaManan100, line 24: Maybe we should move this to unsupported cases now. There are people who use that list to see what all won't work with Vitess and this seems like a good addition to that.

go/vt/vttablet/tabletserver/health_streamer.go
```diff
@@ -380,7 +382,13 @@ func getChangedTableNames(ctx context.Context, conn *connpool.DBConn) ([]string,
 	}
 	alloc := func() *sqltypes.Result { return &sqltypes.Result{} }
 	bufferSize := 1000
-	err := conn.Stream(ctx, mysql.DetectSchemaChange, callback, alloc, bufferSize, 0)
+
+	schemaChangeQuery := mysql.DetectSchemaChange
+	// If views are enabled, then views are tracked/handled separately and schema change does not need to track them.
+	if hs.viewsEnabled {
+		schemaChangeQuery = mysql.DetectSchemaChangeOnlyBaseTable
+	}
```
GuptaManan100, line 390: Can't we always use `mysql.DetectSchemaChangeOnlyBaseTable`?

GuptaManan100 (approved):

go/vt/vtgate/planbuilder/testdata/view_cases.json
frouioui: That's a good idea

go/vt/vttablet/tabletserver/health_streamer.go
harshit-gangal: There is a possibility of users creating views on MySQL and providing that in vschema as table, so schema tracking should be able to populate column list for those when vitess managed views are not enabled.

go/vt/vtgate/planbuilder/testdata/view_cases.json
harshit-gangal: there is already a case in unsupported that fails with this error.

Issue #12411, opened 2023-02-21
Feature Request: extend numeric_static_map vindex to support a hash fallback


maxenglander:
## Description

It would be useful to have the ability to fallback to a hash lookup when there are keys missing from the JSON mapping of `numeric_static_map`. 

E.g.

```
{
  "sharded": true,
  "vindexes": {
    "nsm_idx": {
      "type": "numeric_static_map",
      "params": {
        "fallback_type": "xxhash"
      }
    }
  }
}
```

When `fallback_type` is set, `numeric_static_map` would internally fall back to using the specified hashing function to map requested keys not present in the JSON mapping.

## Use Cases

This would be helpful for use cases like placing large tenants on specific shards, and falling back to hashing for tenants that don't need special handling.

dbussink:
> "use_fallback": true

What identifies in this case what we fall back to? Could you add what the full vschema would look like with how the fallback is defined?

maxenglander:
Hey @dbussink currently in [this draft implementation](https://github.com/vitessio/vitess/pull/12414/files#diff-2706bb924ecf40a03b2fedc1ba40aba4980f022665692a0efb631ee4805d19a0R85-R97) the fallback is hardcoded as `xxhash`. I figured that's simplest for now, and would allow for expansion later on. 

E.g. eventually the draft implementation could be amended like this in a backwards-compatible way:

```
if useFallback != "true" {
  /* assume it's a hashing vindex defined elsewhere in the vschema, look for it and use it */
}
```

I'm open to making this more flexible now if that's your preference. For now I've clarified in this feature request that I'd be fine with a hardcoded hashing function.

deepthi:
Why not replace `use_fallback` with something like `fallback_type: xxhash`? The eliminates hard-coding and ambiguity.

maxenglander:
@deepthi I'm totally fine with that. My thinking was to keep the minimize code complexity and edge cases by limiting options initially, and be able to expand later as the demand for new params/hashes arises.

I'll update the FR  inline with your suggestion.

Issue #12413, opened 2023-02-21
Feature Request: support non-string vindex params


maxenglander:
### Feature Description

Currently the protobuf definition for the vschema supports `map[string]string` params for Vindexes.

It would be great if this was more flexible, and could support boolean, numeric, and nested objects.

### Use Case(s)

This is motivated by https://github.com/vitessio/vitess/issues/12412, which proposes extending `numeric_static_map` vindex with a nested JSON parameter, e.g.:

```
{
  "sharded": true,
  "vindexes": {
    "nsm_idx": {
      "type": "numeric_static_map",
      "params": {
        "json": {
          "1": 2,
          "2": 3,
          "3": 4
        }
      }
    }
  }
}
```

Issue #12412, opened 2023-02-21
Feature Request: extend numeric_static_map vindex to support inline JSON


maxenglander:
### Feature Description

Currently `numeric_static_map` vindex supports defining a JSON file with a mapping of keys to shards. In some environments, it is awkward to try to distribute a JSON file everywhere it needs to be read.

This feature request proposes extending `numeric_static_map` vindex to support inline JSON, e.g. like so:

```
{
  "sharded": true,
  "vindexes": {
    "nsm_idx": {
      "type": "numeric_static_map",
      "params": {
        "json": {
          "1": 2,
          "2": 3,
          "3": 4
        }
      }
    }
  }
}
```

### Use Case(s)

The current `json_path` is a bit difficult to support in our environment, where we offer Vitess-as-a-service to customers, and we need to persist a JSON file in a dedicated location (e.g. cloud object storage, and then distribute this file everywhere it needs to be read). It would simplify things for us a lot to be able to define this mapping directly in the VSchema.

dbussink:
> This feature request proposes extending `numeric_static_map` vindex to support inline JSON, e.g. like so:

I think this also applies to other cases where we have an external JSON file, such as `region_json`. I think we should aim here for allowing embedding then in general if we want to add that?

maxenglander:
I like that, updated. I grepped for `os.ReadFile` and `Unmarshal` in the vindexes package and didn't find other vindexes that have an external JSON file. Let me know if you know of others, or if you meant we should aim even more broadly (e.g. outside of the vindexes package).

deepthi:
How big could these inline JSON objects get? It will be good to review the error handling code for when we write the SrvVSchema to the topo. That could several times the size of the provided VSchema.

maxenglander:
For my immediate use case being able to accommodate a map with 50 or so entries would be plenty. Does capping this to 1-2KiB seem reasonable?

I can test out submitting large payloads to SrvVSchema to see what kinds of errors it returns (if any).

PR #12414, opened 2023-02-21
go/vt/vtgate: add json and use_fallback opts to numeric_static_map vdx

maxenglander:
## Description

`numeric_static_map` supports a JSON file. In some environments it may be more convenient to set this in the VSchema directly, rather than make the JSON file available everywhere it needs to be loaded.

In addition, it would be useful to have the ability to fallback to a hash lookup when there are keys missing from the json mapping.

## Related Issue(s)

 * Fixes https://github.com/vitessio/vitess/issues/12411
 * Addresses https://github.com/vitessio/vitess/issues/12412

One thing @aquarapid and I are not happy about is having to embed a JSON document inside a proto string type. If possible, we would like to amend the proto definition for vindexes to support nested, typed options. Have created https://github.com/vitessio/vitess/issues/12413 to have that discussion, and can amend this PR depending on the outcome of that.

## Checklist

-   [ ] Documentation: https://github.com/vitessio/website/pull/1415

### DIFF
--- a/go/vt/vtgate/vindexes/numeric_static_map.go
+++ b/go/vt/vtgate/vindexes/numeric_static_map.go
@@ -42,8 +42,9 @@ type NumericLookupTable map[uint64]uint64
 // NumericStaticMap is similar to vindex Numeric but first attempts a lookup via
 // a JSON file.
 type NumericStaticMap struct {
-	name   string
-	lookup NumericLookupTable
+	name    string
+	hashVdx Hashing
+	lookup  NumericLookupTable
 }
 
 func init() {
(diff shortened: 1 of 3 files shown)


go/vt/vtgate/vindexes/numeric_static_map.go
```diff
@@ -52,19 +53,53 @@ func init() {
...
+
+	var hashVdx Hashing
+
+	if s, ok := params["fallback_type"]; ok {
+		if err != nil {
+			return nil, err
+		}
+		// We just use xxhash blindly;  we could make this configurable.
+		vindex, err := CreateVindex(s, name+"_hash", map[string]string{})
+		if err != nil {
```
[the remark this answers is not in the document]
maxenglander: Done!

harshit-gangal (approved):

go/vt/vtgate/vindexes/numeric_static_map.go
```diff
@@ -52,19 +53,52 @@ func init() {
...
+		lt, err = parseNumericLookupTable([]byte(jsonStr))
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	var hashVdx Hashing
+
+	if s, ok := params["fallback_type"]; ok {
+		if err != nil {
```
systay, line 87: don't we want to check the `err` variable outside of this `if` statement?

go/vt/vtgate/vindexes/numeric_static_map.go
```diff
@@ -121,22 +155,32 @@ func (vind *NumericStaticMap) Hash(id sqltypes.Value) ([]byte, error) {
 		return nil, err
 	}
 	lookupNum, ok := vind.lookup[num]
-	if ok {
+	if !ok {
+		// Not in lookup, us fallback hash
```
systay, line 159: nitpick: `us` -> `use`

systay (approved):
nitpicks aside, LGTM!

go/vt/vtgate/vindexes/numeric_static_map.go
maxenglander: This check looks completely wrong, not sure what I was thinking here

harshit-gangal (approved):

harshit-gangal:
You would need to update vindex page `https://vitess.io/docs/16.0/reference/features/vindexes/`

harshit-gangal:
> 

It will be in 17.0 only.

PR #12415, opened 2023-02-21
Examples: Correct VTAdmin Discovery File Path And Add Check

mattlord:
## Description

VTAdmin failing to discover the Vitess cluster in the local examples was [reported in the Vitess Slack](https://vitess.slack.com/archives/C02JQQ88Z1C/p1676993782703629).

This was caused by the relative path usage missed as part of the refactoring done in: https://github.com/vitessio/vitess/pull/12239

And this was easily missed because there was no check done to confirm proper discovery and thus the local examples CI tests did not fail.

In this PR we correct the static discovery file path AND add a check to ensure that the discovery was successful and if not, we exit with an error (which would cause the tests to fail).

>**Note** While testing/confirming this fix in the CI with manual test re-runs, I saw [another timing issue](https://github.com/vitessio/vitess/actions/runs/4237884075/jobs/7364570381) related to the specific healthy shard indicator that the scripts were using vs. what `vtctld` used in checking that a shard has a healthy serving primary tablet when e.g. executing `ApplySchema` (this was also reported [in the Vitess Slack here](https://vitess.slack.com/archives/C0PQY0PTK/p1676977726069479)). I [changed the example scripts' healthy shard check](https://github.com/vitessio/vitess/pull/12415/commits/48ef9ca9cfb61c989e772a3aae914d22822b3a48) to match [with what vtctld uses](https://github.com/vitessio/vitess/blob/d473a7b0c209790be5ed26a2cc1449db094d2986/go/vt/topo/shard.go#L181-L183). After that commit, [the tests never failed](https://github.com/vitessio/vitess/actions/runs/4238921222).

## Related Issue(s)

  - Fixes: https://github.com/vitessio/vitess/issues/12416
  - Follow-up to: https://github.com/vitessio/vitess/pull/12239

## Checklist

### DIFF
--- a/examples/common/lib/utils.sh
+++ b/examples/common/lib/utils.sh
@@ -52,18 +52,19 @@ function wait_for_healthy_shard_primary() {
 	fi
 	keyspace=${1}
 	shard=${2}
-	healthy_indicator="PRIMARY: Serving"
+	unhealthy_indicator='"primary_alias": null'
 	wait_secs=180
 
 	for _ in $(seq 1 ${wait_secs}); do
-		if curl -s "http://$(vtctldclient GetTablets --keyspace "${keyspace}" --shard "${shard}" | grep -i 'primary' | awk '{print $5}')/debug/status_details" | grep -qi "${healthy_indicator}"; then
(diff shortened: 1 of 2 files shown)


notfelineit (approved):
Nice, love the added check 🥳

GuptaManan100 (approved):

Issue #12419, opened 2023-02-21
Bug Report: MariaDB support broken in vdiff2


mikehamrick:
### Overview of the Issue

When attempting to perform an import from an importkeyspace of `10.3.36-MariaDB-log` we see repeated errors in the Vitess logs in the form of:

>  Error retrying vdiffs: You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near '>>'$.core_options.auto_retry' = 'true'' at line 1 (errno 1064) (sqlstate 42000) during query: select * from _vt.vdiff where state = 'error' and options->>'$.core_options.auto_retry' = 'true'

It appears that the `column->>path` JSON operator used in the SQL here is [not supported](https://jira.mariadb.org/browse/MDEV-13594) by MariaDB.

### Reproduction Steps

I don't have a complete list of reproduction steps here sadly. It's from an import of an external MariaDB keyspace. Hopefully the error message alone will be enough to find the issue!

### Binary Version

```sh
2023-02-07.e775d05
```

### Operating System and Environment details

```sh
N/A
```

mattlord:
[We do not support MariaDB in Vitess v16](https://github.com/vitessio/vitess/issues/9518) (where [VDiff2 is GA and the default](https://github.com/vitessio/vitess/pull/12084)) and later, BUT this is a pretty easy fix — as `col1->>'$.field1'` is equivalent to `JSON_UNQUOTE(JSON_EXTRACT(col1, '$.field1'))` — and doing so allows us to fully support VDiff2 with MariaDB imports, even when running the diffs against the reverse workflows.

deepthi:
@mattlord to me this looks like
- there's an external primary vttablet against MariaDB
- because it is a primary we create all the engines (including VDiff)
- that engine periodically checks for failed VDiff attempts
- the query for fetching those fails

So even before anyone tries to actually run a VDiff this will continuously error.

mattlord:
@deepthi correct, you would see an error log message about this every 30 seconds on that tablet. It wouldn't cause anything to fail, but you would see that in the logs.

mattlord:
Fixed by: https://github.com/vitessio/vitess/pull/12420

PR #12084, opened 2023-01-12
Mark VDiff V2 as GA in v16

mattlord:
## Description

[VDiff v2](https://vitess.io/docs/16.0/reference/vreplication/vdiff2/) is now considered [feature complete](https://github.com/vitessio/vitess/issues/10494) and has been experimental for 1+ releases now so we are marking it as Generally Available or production-ready in v16. For now we will hold off on deprecating v1 — with an eventual removal coming 1+ releases after the deprecation — but the deprecation should probably be done in v17.

## Related Issue(s)

  - https://github.com/vitessio/vitess/issues/10494

## Checklist

-   [x] "Backport to:" labels have been added if this change should be back-ported
-   [x] Tests are not required
-   [x] Documentation: https://github.com/vitessio/website/pull/1347

PR #12420, opened 2023-02-22
VReplication: Use MariaDB Compat JSON Functions

mattlord:
## Description

This is needed to better support importing from MariaDB to Vitess/MySQL as we leverage MySQL's JSON support more and more in VReplication.

While [we do not fully support MariaDB in Vitess v16](https://github.com/vitessio/vitess/issues/9518) and later (where [VDiff2 is GA and the default](https://github.com/vitessio/vitess/pull/12084)), we DO want to support easy and reliable imports from MariaDB to Vitess and MySQL. This is also a pretty easy and straightforward fix — as `col1->>'$.field1'` is equivalent to `JSON_UNQUOTE(JSON_EXTRACT(col1, '$.field1'))` — and doing so allows us to fully support VDiff2 with MariaDB imports, even when running the diffs are run against the reverse workflows (after `SwitchTraffic` is done).

## Related Issue(s)

  - Fixes: https://github.com/vitessio/vitess/issues/12419

## Checklist

### DIFF
--- a/go/vt/vttablet/tabletmanager/vdiff/engine_test.go
+++ b/go/vt/vttablet/tabletmanager/vdiff/engine_test.go
@@ -236,7 +236,7 @@ func TestEngineRetryErroredVDiffs(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			vdiffenv.dbClient.ExpectRequest("select * from _vt.vdiff where state = 'error' and options->>'$.core_options.auto_retry' = 'true'", tt.retryQueryResults, nil)
+			vdiffenv.dbClient.ExpectRequest("select * from _vt.vdiff where state = 'error' and json_unquote(json_extract(options, '$.core_options.auto_retry')) = 'true'", tt.retryQueryResults, nil)
 
 			// Right now this only supports a single row as with multiple rows we have
 			// multiple controllers in separate goroutines and the order is not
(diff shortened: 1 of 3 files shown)


mattlord:
The OnlineDDL VReplication test failures are a known unrelated issue that is being worked on.

GuptaManan100 (approved):
LGTM

mattlord:
> > The OnlineDDL VReplication test failures are a known unrelated issue that is being worked on.
> 
> @mattlord Is this something we could relatively easily test in the MariaDB migration job that we have? That it also tests a vdiff both ways?

I double checked... we *do* already do this today: https://github.com/vitessio/vitess/blob/a5d0cdc479b41155db8275b4e58fd7baf84c30d8/go/test/endtoend/vreplication/vreplication_test.go#L773

What wouldn't have worked on the reverse workflow is the auto retry on error (because the query used to find VDiffs that should be retried failed due to the invalid syntax). But you could still manually [re]start the workflow:
  - Stop (clear the error and mark it as stopped): https://vitess.io/docs/16.0/reference/vreplication/vdiff/#stopping-a-vdiff
  - Resume: https://vitess.io/docs/16.0/reference/vreplication/vdiff/#resume-a-previous-vdiff

Or of course you could run another VDiff, which is what you always had to do with VDiff1 if you encountered an error. 

So again, I think in practice this is a pretty minor issue but again, one that I think is still definitely worth doing.

rohit-nayak-ps (approved):
lgtm

dbussink (approved):

PR #12421, opened 2023-02-22
Change `GetSchema` RPC to return `CreateView` instead of `SelectStmt`

frouioui:
## Description

This PR changes `GetSchema` RPC to return `CreateView` instead of `SelectStmt`.

## Related Issue(s)

- Part of #11559 

## Checklist


harshit-gangal (approved):

GuptaManan100 (approved):
I understand the changes, but just want to confirm why are we dropping the ViewStatement field. Is it just because we don't need it and is like a duplicated thing already part of the create view statement?

harshit-gangal:
> I understand the changes, but just want to confirm why are we dropping the ViewStatement field. Is it just because we don't need it and is like a duplicated thing already part of the create view statement?

Yes, create view has more information than the select statement part of it. If we need additional information about the view, it is present in create view statement that can be used.

PR #12423, opened 2023-02-22
Cleanup vtorc ssl package

dbussink:
This is not used for anything, so let's remove it. The only references to the code here was in the tests for itself.

## Checklist


frouioui (approved):

GuptaManan100 (approved):
LGTM! Thank-you!

PR #12424, opened 2023-02-22
Add configuration option for minimum TLS version

dbussink:
We were hardcoding this to TLS 1.0 which is not considered safe anymore these days and was called out by CodeQL as a security issue.

Instead we should default to TLS 1.2 which we do everywhere else, and allow for an option to change it if needed so that we're safe by default here.

I've also reviewed the documentation about `vtgr`, but we don't make any mention there for any of the SSL/TLS related configuration options, so I don't think there's a good place to add that yet? It's also marked still as experimental. 

I did add a release notes entry with the change that the default is now TLS 1.2. 

## Related Issue(s)

This is the last remaining issue identified in #12216 

## Checklist

### DIFF
--- a/doc/releasenotes/17_0_0_summary.md
+++ b/doc/releasenotes/17_0_0_summary.md
@@ -14,6 +14,10 @@
 
 ### <a id="breaking-changes"/>Breaking Changes
 
+#### <a id="vtgr-default-tls-version"/>Default TLS version changed for `vtgr`
+
+When using TLS with `vtgr`, we now default to TLS 1.2 if no other explicit version is configured. Configuration flags are provided to explicitly configure the minimum TLS version to be used. 
+
 #### <a id="deprecated-stats"/>Deprecated Stats
 
 These stats are deprecated in v17.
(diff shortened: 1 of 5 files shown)


doc/releasenotes/17_0_0_summary.md
```diff
@@ -14,6 +14,8 @@
 
 ### <a id="breaking-changes"/>Breaking Changes
 
+ - When using TLS with `vtgr`, we now default to TLS 1.2 if no other explicit version is configured. Configuration flags are provided to explicitly configure the minimum TLS version to be used. 
+
```
frouioui, line 18: I think this should be in its own subheading (`####`) under `Breaking Changes`. It's easier for readers to read when they're skimming through the summary at the top of the file.

frouioui (approved):

doc/releasenotes/17_0_0_summary.md
dbussink: @frouioui What should the title be then? Any example I could look at?

GuptaManan100: We could use something like `#### <a id="tls-version"/>TLS Version Change in Vtgr`

GuptaManan100 (commented):
Everything else looks good to me.

GuptaManan100 (approved):

PR #12428, opened 2023-02-22
Add 128 bit hashing for the evalengine

dbussink:
This adds support for 128 bit hashes. With 128 bit hashes, we have enough entropy so we can remove the need to compare values after  hashing.

It only adds the new function with tests, usage will be added in a future change. We've been using it internally in PlanetScale already with no issues. 

## Related Issue(s)

Extension after #12247 

## Checklist


vmg (approved):

go/mysql/collations/uca_test.go
```diff
@@ -870,8 +876,16 @@ func (c *ConsistentCollation) Collate(left, right []byte, isPrefix bool) int {
 		c.t.Errorf("ConsistentCollation(%s): expected WeightString %q / %v == %q / %v to be %v", c.Name(), left, w1, right, w2, equal)
 	}
 
-	h1 := c.Hash(left, 0)
-	h2 := c.Hash(right, 0)
+	var hasher vthash.Hasher
+
+	hasher.Reset()
```
frouioui, line 881: isn't the `hasher` already reseted after its creation on line `879`?

go/vt/vtgate/evalengine/internal/decimal/helpers.go
```diff
@@ -0,0 +1,7 @@
+package decimal
+
```
frouioui, line 2: The header is missing

go/vt/vthash/hash.go
```diff
@@ -1,7 +1,5 @@
-//go:build amd64 || arm64 || mips64 || mips64le || ppc64 || ppc64le || riscv64 || s390x
-
 /*
-Copyright 2021 The Vitess Authors.
+Copyright 2022 The Vitess Authors.
```
frouioui, line 2: nit: 2023

go/vt/vthash/metro/metro.go
```diff
@@ -0,0 +1,186 @@
+/*
+Copyright 2022 The Vitess Authors.
```
frouioui, line 2: nit: 2023, same on its test file

frouioui (approved):

PR #12432, opened 2023-02-22
Remove `are-you-alive` example

frouioui:
## Description

This PR removes an old and unused example: `are-you-alive`. Removing this will allow us to not have to maintain it.

## Checklist


dbussink (approved):
Yay removing code!

GuptaManan100 (approved):

PR #12437, opened 2023-02-22
Adding check for v16 in backup tests

rsajwani:
## Checklist


GuptaManan100 (approved):

.github/workflows/upgrade_downgrade_test_backups_e2e_next_release.yml
```diff
@@ -191,4 +191,4 @@ jobs:
         mkdir -p /tmp/vtdataroot
         set -x
         source build.env
-        eatmydata -- go run test.go -skip-build -keep-data -docker=false -print-log -follow -tag upgrade_downgrade_backups
```
GuptaManan100, line 194: I looked at the code and `keep-data` defaults to true. So you can't just remove the flag. You actually need `-keep-data=false`. That  is what I have done in https://github.com/vitessio/vitess/pull/12462

GuptaManan100 (changes requested):

.github/workflows/upgrade_downgrade_test_backups_e2e_next_release.yml
GuptaManan100: Strictly it is not required since the tests are passing on this PR, but if we actually want to turn it off, then we need to do it this way.

harshit-gangal: initially it was false as default, we should move back to that. not for this PR

harshit-gangal (approved):

frouioui (approved):

GuptaManan100 (approved):

PR #12442, opened 2023-02-23
schemadiff: AlterTableAlgorithmStrategy

shlomi-noach:
## Description

`schemadiff` now adds an `ALGORITHM` hint. The algorithm has an operational effect on the MySQL engine, but no semantic effect on the table. It only applies for `AlterTable` diffs.

## Related Issue(s)

## Checklist


GuptaManan100 (approved):
LGTM

rohit-nayak-ps (approved):
lgtm

PR #12447, opened 2023-02-23
GetSchema rpc to streaming api

harshit-gangal:
## Description

This PR converts `GetSchema` RPC to a streaming api.

## Related Issue(s)

 - #11559 

## Checklist


go/vt/vtgate/schema/tracker.go
```diff
@@ -134,21 +134,18 @@ func (t *Tracker) loadViews(conn queryservice.QueryService, target *querypb.Targ
...
+	// whenever a shard's primary tablet starts and sends the initial signal.
+	// This is needed clear out any stale view definitions.
 	t.clearKeyspaceViews(target.Keyspace)
-	t.updateViews(target.Keyspace, fvRes)
-	log.Infof("finished loading views for keyspace %s. Found %d views", target.Keyspace, len(fvRes))
 
-	return nil
+	return conn.GetSchema(t.ctx, target, querypb.SchemaTableType_VIEWS, nil, func(schemaRes *querypb.GetSchemaResponse) error {
+		t.updateViews(target.Keyspace, schemaRes.TableDefinition)
+		log.Infof("finished loading views for keyspace %s. Found %d views", target.Keyspace, len(schemaRes.TableDefinition))
```
GuptaManan100, line 146: This log line is now incorrect. You can have the view definitions coming in separate packets, so you don't want to print 2 lines saying it finished loading the views for a keyspace and reporting different number of views each time, while the correct count is the sum of both of them

GuptaManan100 (approved):
Rest LGTM

go/vt/vtgate/schema/tracker.go
harshit-gangal: I thought about that. moving it out.

frouioui (approved):

go/vt/vtgate/schema/tracker.go
```diff
@@ -134,20 +134,23 @@ func (t *Tracker) loadViews(conn queryservice.QueryService, target *querypb.Targ
...
+	// whenever a shard's primary tablet starts and sends the initial signal.
+	// This is needed clear out any stale view definitions.
 	t.clearKeyspaceViews(target.Keyspace)
-	t.updateViews(target.Keyspace, fvRes)
-	log.Infof("finished loading views for keyspace %s. Found %d views", target.Keyspace, len(fvRes))
 
+	var noOfViews int
+	err := conn.GetSchema(t.ctx, target, querypb.SchemaTableType_VIEWS, nil, func(schemaRes *querypb.GetSchemaResponse) error {
+		t.updateViews(target.Keyspace, schemaRes.TableDefinition)
+		noOfViews += len(schemaRes.TableDefinition)
```
ajm188, line 147: this is an extreme nitpick (and especially if this was existing code please ignore it) -- to me, "no." only works as an abbreviation with the period included (which you obviously can't have here), how about `numViews` instead?

ajm188 (approved):
lgtm, I have one extremely annoying quibble

Issue #12450, opened 2023-02-23
Bug Report: vitess migration may be caught in endless cut-over retries


shlomi-noach:
### Overview of the Issue

It is possible for a vitess migration to get into a seemingly endless sequence of failed cut-over attempts. An example log follows:

```log
I0223 10:06:04.217356 2580892 executor.go:4037] updateMigrationStage: uuid=5790ab5c_b361_11ed_b4e5_0a43f95f28a3, stage=starting cut-over
I0223 10:06:04.266126 2580892 executor.go:4037] updateMigrationStage: uuid=5790ab5c_b361_11ed_b4e5_0a43f95f28a3, stage=sentry table created: _vt_HOLD_aeaf349ab36111edac600a43f95f28a3_20230224100604
I0223 10:06:04.273038 2580892 executor.go:4037] updateMigrationStage: uuid=5790ab5c_b361_11ed_b4e5_0a43f95f28a3, stage=waiting for post-sentry pos: MySQL56/324e040a-b349-11ed-bf20-0a43f95f28a3:1-8327
E0223 10:06:09.278869 2580892 engine.go:823] error waiting for pos: MySQL56/324e040a-b349-11ed-bf20-0a43f95f28a3:1-8327, last pos: MySQL56/324e040a-b349-11ed-bf20-0a43f95f28a3:1-8326: context deadline exceeded, wait time: 4.998974716s: possibly no tablets are available to stream in the source keyspace for your cell and tablet_types setting
W0223 10:06:09.278953 2580892 rpc_server.go:72] TabletManager.VReplicationWaitForPos(id:4 position:"MySQL56/324e040a-b349-11ed-bf20-0a43f95f28a3:1-8327")(on zone1-0000000100 from ) error: error waiting for pos: MySQL56/324e040a-b349-11ed-bf20-0a43f95f28a3:1-8327, last pos: MySQL56/324e040a-b349-11ed-bf20-0a43f95f28a3:1-8326: context deadline exceeded, wait time: 4.998974716s: possibly no tablets are available to stream in the source keyspace for your cell and tablet_types setting
I0223 10:06:09.302441 2580892 executor.go:4130] updateMigrationMessage: uuid=5790ab5c_b361_11ed_b4e5_0a43f95f28a3, message=rpc error: code = DeadlineExceeded desc = context deadline exceeded
[...]

PR #12451, opened 2023-02-23
OnlineDDL: mitigate scenario where a migration sees recurring cut-over timeouts

shlomi-noach:
## Description

Issue details in https://github.com/vitessio/vitess/issues/12450

This PR offers a quick mitigation like so:

- Increase of the `const` value of `vreplicationCutOverThreshold` to `10sec`. The absolute value of `5sec` is just too low for the combined delays incurred by even the smallest replication lag, plus wait interval times. It's too marginal.

In a future PR we will make the threshold dynamic, and collaborative with the throttler's threshold, if enabled. But in the meantime `10sec` is a substantial improvement for most situations.

Additionally, this PR now `DROP`s the temporary sentry table used in a cut-over step, as soon as the cut-over is completed (or failed), so that we don't get a proliferation of tables in the schema. If the table is successfully dropped, then we also remove it from the `artifacts` column, so that the column does not get bloated. If, for any reason, the table cannot be dropped, we still complete the cut-over step, and we let the existing `gcartifacts()` and table GC mechanisms to eventually reclaim the table. Noteworthy that the sentry table is empty, and never populated, and so `DROP`ping it is a fast and safe operation.

## Related Issue(s)

- Fixes https://github.com/vitessio/vitess/issues/12450
- https://github.com/vitessio/vitess/issues/6926

## Checklist


notfelineit (approved):
the in-code comments are so great!

shlomi-noach:
`onlineddl_vrepl_stress` is failing and probably related. Checking.

shlomi-noach:
Ping for review

go/vt/vttablet/tabletserver/query_executor.go
```diff
@@ -66,7 +66,7 @@ type QueryExecutor struct {
 
 const (
 	streamRowsSize         = 256
-	maxQueryBufferDuration = 10 * time.Second
+	maxQueryBufferDuration = 15 * time.Second
```
shlomi-noach, line 69: This is the single most important change in this PR, an increase in timeout.

go/vt/vttablet/onlineddl/executor.go
```diff
@@ -138,7 +138,7 @@ const (
 	emptyHint                                = ""
 	readyToCompleteHint                      = "ready_to_complete"
 	databasePoolSize                         = 3
-	vreplicationCutOverThreshold             = 5 * time.Second
+	vreplicationCutOverThreshold             = 10 * time.Second
```
shlomi-noach, line 141: This is the single most important change in this PR, an increase in timeout.

PR #12452, opened 2023-02-23
evalengine: use 128 bit hashing internally

vmg:
## Description

Follow up on https://github.com/vitessio/vitess/pull/12428. Now that we have support for 128-bit hashing, we should also use it internally for all hashing operations.

cc @dbussink 

## Checklist

### DIFF
--- a/go/vt/vtgate/evalengine/eval_tuple.go
+++ b/go/vt/vtgate/evalengine/eval_tuple.go
@@ -18,15 +18,10 @@ package evalengine
 
 import (
 	"vitess.io/vitess/go/sqltypes"
-	"vitess.io/vitess/go/vt/vthash"
 )
 
 type evalTuple struct {
 	t []eval
-}
-
-func (e *evalTuple) Hash(h *vthash.Hasher) {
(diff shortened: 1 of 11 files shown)


go/vt/vtgate/evalengine/eval_tuple.go
```diff
@@ -18,16 +18,15 @@ package evalengine
...
 )
 
 type evalTuple struct {
 	t []eval
 }
 
-func (e *evalTuple) Hash() (HashCode, error) {
-	return 0, vterrors.Errorf(vtrpcpb.Code_INTERNAL, "should never attempt to hash a TUPLE")
+func (e *evalTuple) Hash(h *vthash.Hasher) {
+	panic("should never attempt to hash a TUPLE")
```
dbussink, line 29: Hmm, should we still have an error for potentially unexpected cases like this? So we don't panic / crash the whole vtgate if we hit a bug like this even if it's not expected.

dbussink (approved):

frouioui (approved):

GuptaManan100 (approved):

PR #12453, opened 2023-02-23
examples schema: demi-idempotency via CREATE TABLE IF NOT EXISTS

shlomi-noach:
## Description

This PR changes all `CREATE TABLE` into `CREATE TABLE IF NOT EXISTS` in the various `examples` schemas.
This allows the user to take down a cluster and bring it up again with no error (currently the `create table` statements fail the bootstrap script).

## Checklist

### DIFF
--- a/examples/backups/create_commerce_schema.sql
+++ b/examples/backups/create_commerce_schema.sql
@@ -1,11 +1,11 @@
-create table product(
+create table if not exists product(
     sku varbinary(128),
     description varbinary(128),
     price bigint,
     primary key(sku)
 ) ENGINE=InnoDB;
 
-create table customer_seq(id int, next_id bigint, cache bigint, primary key(id)) comment 'vitess_sequence';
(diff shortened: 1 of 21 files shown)


GuptaManan100 (approved):
I like the idea

shlomi-noach:
Anecdotally, this is the process I use for development. `101` cluster is my go-to unsharded cluster setup; as I make code changes, I take down the cluster, deploy new changes (e.g. add some specific logging), run the new code on the existing cluster (so I don't need to re-init everything) and so forth.

frouioui (approved):
Sounds good to me!

deepthi:
> This allows the user to take down a cluster and bring it up again with no error

The idea is that all vitess processes can be taken down but the MySQL databases still exist, and you restart everything on the same data.

shlomi-noach:
> The idea is that all vitess processes can be taken down but the MySQL databases still exist, and you restart everything on the same data.

Right, but the `examples` did not comply with that idea.

PR #12454, opened 2023-02-23
[main] Create Views allowed for same keyspace (#12409)

harshit-gangal:
## Description
This is a forward port of #12409.


frouioui (approved):

GuptaManan100 (approved):
We'll have to merge `main` for the `onlineddl_vrepl` tests to pass. I think the flakiness is fixed on `main`

PR #12455, opened 2023-02-23
WIP: sqlparser: ParseAndBindWithQualifier() convenience function

shlomi-noach:
## Description

There is a use case where it's common to parse a query where we both want to specify a qualifier (schema name or table name) as well as bind variables. It's a bit complicated to mix both types when parsing&binding queries.

This PR offers a new, dedicated function called `ParseAndBindWithQualifier()`. See unit test addition for usage samples. cc @mattlord 

## Checklist

### DIFF
--- a/go/vt/sqlparser/parsed_query.go
+++ b/go/vt/sqlparser/parsed_query.go
@@ -218,3 +218,23 @@ func ParseAndBind(in string, binds ...*querypb.BindVariable) (query string, err 
 	}
 	return parsed.GenerateQuery(bindVars, nil)
 }
+
+// ParseAndBind is a one step sweep that binds variables to an input query, in order of placeholders.
+// It is useful when one doesn't have any parser-variables, just bind variables.
+// Example:
+//
+//	query, err := ParseAndBind("select * from tbl where name=%a", sqltypes.StringBindVariable("it's me"))
+func ParseAndBindWithQualifier(in string, qualifier string, binds ...*querypb.BindVariable) (query string, err error) {
+	vars := make([]any, len(binds))
+	for i := range binds {
+		vars[i] = fmt.Sprintf(":var%d", i)
+	}
+	allVars := append([]any{qualifier}, vars...)
+	parsed := BuildParsedQuery(in, allVars...)
+
+	bindVars := map[string]*querypb.BindVariable{}
+	for i := range binds {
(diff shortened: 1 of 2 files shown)


go/vt/sqlparser/parsed_query_test.go
```diff
@@ -203,3 +203,38 @@ func TestParseAndBind(t *testing.T) {
...
+		in        string
+		qualifier string
+		binds     []*querypb.BindVariable
+		out       string
+	}{
+		{
+			in:        "drop table `%s`",
+			qualifier: "tbl",
+			out:       "drop table `tbl`",
+		},
```
mattlord, line 218: We should test w/o escaping in the in query too. Both when escaping is required (e.g. `test-name`) and when it's not. I think it's problematic to include the escaping in the incoming query, as it's up to the parser to provide the escaping when needed. I see specifying the escaping for the identifier in the raw query string as being equivalent to specifying quotes when we know the bind placeholder is for a string: `'%a'`. I suspect that when the escaping IS needed, we'll have double escaping here:
```
drop table ``test-name``
```

go/vt/sqlparser/parsed_query.go
```diff
@@ -218,3 +218,23 @@ func ParseAndBind(in string, binds ...*querypb.BindVariable) (query string, err
 	}
 	return parsed.GenerateQuery(bindVars, nil)
 }
+
+// ParseAndBind is a one step sweep that binds variables to an input query, in order of placeholders.
+// It is useful when one doesn't have any parser-variables, just bind variables.
+// Example:
+//
+//	query, err := ParseAndBind("select * from tbl where name=%a", sqltypes.StringBindVariable("it's me"))
```
mattlord, line 226: We should update the comment. 🙂

go/vt/sqlparser/parsed_query_test.go
shlomi-noach: Yeah, I feel like we should get @vitessio/query-serving 's official opinion. No strong opinions on my side. I understand both approaches, they both make sense.

shlomi-noach:
Likely to close this PR. It's a proposition I'm not very happy about. We're thinking of alternative ways.

shlomi-noach:
closing as irrelevant

PR #12457, opened 2023-02-23
Fix Vtop example and release script

### LINKED ISSUE #12456: Bug Report: operator example is probably broken on main
### Overview of the Issue

In #12054 we changed the docker images to default to MySQL 8.0. However, the example still has
```
mysql56Compatible: vitess/lite:v16.0.0-rc1
```
We need to change this to 
```
mysql80Compatible: vitess/lite:v16.0.0-rc1
```

Also, the operator.yaml file is outdated. We also need to fix the yaml files to not pass the `enforceSemiSync` field when it is removed.

### Reproduction Steps

Run the operator example from
https://github.com/vitessio/vitess/tree/main/examples/operator/

### Binary Version

```sh
main
```

```sh
any
```

### Log Fragments

_No response_

GuptaManan100:
## Description
Since the default images are now MySQL 8.0, we need to update the operator example to use the right property for the image.

This PR also fixes the yaml files to not use `enforceSemiSync` anymore and updates the operator.yaml to the latest version, and fixes the release script to also change the vtadmin image version automatically.
Further the vtop example script is fixed to use `vtctldclient` inplace of `vtctlclient`.

## Related Issue(s)
Fixes #12456

## Checklist

### DIFF
--- a/examples/operator/101_initial_cluster.yaml
+++ b/examples/operator/101_initial_cluster.yaml
@@ -15,7 +15,7 @@ spec:
     vtbackup: vitess/lite:latest
     vtorc: vitess/lite:latest
     mysqld:
-      mysql56Compatible: vitess/lite:latest
+      mysql80Compatible: vitess/lite:latest
     mysqldExporter: prom/mysqld-exporter:v0.11.0
   cells:
   - name: zone1
--- a/examples/operator/101_initial_cluster.yaml
+++ b/examples/operator/101_initial_cluster.yaml
@@ -87,15 +87,12 @@ spec:
           databaseInitScriptSecret:
             name: example-cluster-config
             key: init_db.sql
-          replication:
-            enforceSemiSync: false
           tabletPools:
           - cell: zone1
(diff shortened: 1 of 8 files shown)


frouioui (approved):

examples/operator/101_initial_cluster.yaml
```diff
@@ -15,7 +15,7 @@ spec:
     vtbackup: vitess/lite:latest
```
zaphinath, line 15: Testing this branch and the vtctld image ends up in a crashloop backoff with the following error: 
unknown flag: --enable_realtime_stats

zaphinath (changes requested):

examples/operator/101_initial_cluster.yaml
GuptaManan100: That will only happen if you are using an older version of `operator` and not the latest version.

harshit-gangal (approved):

PR #12440, opened 2023-02-23
examples: Fix Vtop example and release script

deepthi:
## Description
Since the default images are now MySQL 8.0, we need to update the operator example to use the right property for the image.

This PR also fixes the image we use for vtadmin deployment and fixes the release script to also change that automatically.
Further the vtop example script is fixed to use `vtctldclient` inplace of `vtctlclient`.

## Related Issue(s)
Fixes #12438 

## Checklist

-   [ ] "Backport to:" labels have been added if this change should be back-ported
-   [x] Tests were added or are not required
-   [x] Documentation was added or is not required

Issue #12438, opened 2023-02-22
Bug Report: operator example is probably broken on release-16.0


deepthi:
### Overview of the Issue

In #12054 we changed the docker images to default to MySQL 8.0. However, the example still has
```
mysql56Compatible: vitess/lite:v16.0.0-rc1
```
We need to change this to 
```
mysql80Compatible: vitess/lite:v16.0.0-rc1
```

### Reproduction Steps

Run the operator example from
https://github.com/vitessio/vitess/blob/release-16.0/examples/operator/

### Binary Version

```sh
16.0.0-rc1
```

### Operating System and Environment details

```sh
any
```

PR #12458, opened 2023-02-23
Fix vtctldclient command in vtop test script and the fix operator example

GuptaManan100:
## Description

This PR fixes a `vtctldclient` command which was incorrectly pushed in the PR https://github.com/vitessio/vitess/pull/12440 while fixing the issue https://github.com/vitessio/vitess/issues/12438

This PR also updates the operator.yaml file and removes `enforceSemiSync` property since it is deleted.
These changes were missed in https://github.com/vitessio/vitess/pull/12440 though they were included in https://github.com/vitessio/vitess/pull/12457

## Related Issue(s)

- https://github.com/vitessio/vitess/issues/12438

## Checklist


GuptaManan100:
Just as a note, the test in the script `vtop_example.sh` were working before these changes too, but these changes are still good to make since the operator.yaml file includes the crds which are updated.

examples/operator/operator.yaml
```diff
@@ -74,8 +74,35 @@ spec:
                         - kind
                         - name
                       type: object
+                      x-kubernetes-map-type: atomic
```
rsajwani, line 77: for my knowledge.. Is it some auto generated file ? and what are all these changes?

GuptaManan100: It is auto generated, but on vitess-operator. Whenever it changes, we have to copy them here, but I forgot to do that this time. 😭 
Anyways these changes came from upgrading the controller-gen version from `controller-gen.kubebuilder.io/version: v0.6.2` to `controller-gen.kubebuilder.io/version: v0.11.3`

test/vtop_example.sh
```diff
@@ -252,7 +252,7 @@ function verifyVtadminSetup() {
 function verifyVTOrcSetup() {
   # Set the primary tablet to readOnly using the vtctld and wait for VTOrc to repair
   primaryTablet=$(getPrimaryTablet)
-  vtctldclient SetReadOnly "$primaryTablet"
+  vtctldclient SetWritable "$primaryTablet" false
```
harshit-gangal, line 255: Do we still have SetReadWrite command as well?

harshit-gangal (approved):

test/vtop_example.sh
GuptaManan100: No we don't. In vtctldclient we only have `SetWritable`.
In vtctlclient we have `SetReadOnly` and `SetReadWrite`

frouioui (approved):

rsajwani (commented):
LGTM

PR #12460, opened 2023-02-23
Fix shorthand flag for recursive delete in zk tool

adsr:
## Description

`zk rm` appears to be broken due to a mistake in pflag setup.

The pflag library which we recently migrated to prohibits shorthand flags greater than a single character[1]. As currently written, the `zk` tool panics in `cmdRm` due to the shorthand `rf`. ~The patch fixes by changing this shorthand from `rf` to `F`.~ Fortunately pflag supports combined flags. Since we already have `r` and `f` defined, we don't need a separate `rf` flag, so we can simply remove this line.

[1] https://github.com/spf13/pflag/blob/85dd5c8bc61cfa382fecd072378089d4e856579d/flag.go#L868-L872

## Related Issue(s)

N/A

## Checklist

### DIFF
--- a/go/cmd/zk/zkcmd.go
+++ b/go/cmd/zk/zkcmd.go
@@ -477,19 +477,15 @@ func cmdTouch(ctx context.Context, subFlags *pflag.FlagSet, args []string) error
 
 func cmdRm(ctx context.Context, subFlags *pflag.FlagSet, args []string) error {
 	var (
-		force             bool
-		recursiveDelete   bool
-		forceAndRecursive bool
+		force           bool
+		recursiveDelete bool
 	)
 	subFlags.BoolVarP(&force, "force", "f", false, "no warning on nonexistent node")
 	subFlags.BoolVarP(&recursiveDelete, "recursivedelete", "r", false, "recursive delete")
-	subFlags.BoolVarP(&forceAndRecursive, "forceandrecursive", "rf", false, "shorthand for -r -f")
 
 	if err := subFlags.Parse(args); err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


deepthi:
@ajm188 our tests [here](https://github.com/vitessio/vitess/tree/main/go/flags/endtoend) don't capture sub flags. This seems like a small enough change that we could merge it without a test. WDYT?

ajm188:
> @ajm188 our tests [here](https://github.com/vitessio/vitess/tree/main/go/flags/endtoend) don't capture sub flags. This seems like a small enough change that we could merge it without a test. WDYT?

I agree with this; those tests would need a pretty heavy overhaul to handle subcommands and it doesn't feel worth it for this change.

That said ... why do we need this at all? With `pflag`, we just remove that `forceandrecursive` and pass the shorthands for `recursive` and `force` together. That is, if we delete the line this PR changes, `zk rm -rf` will do exactly what @adsr wants.

adsr:
> That said ... why do we need this at all? With pflag, we just remove that forceandrecursive and pass the shorthands for recursive and force together. That is, if we delete the line this PR changes, zk rm -rf will do exactly what @adsr wants.

This sgtm. Updated PR.

ajm188:
> This sgtm. Updated PR.

Can you also update the description and related prose?

adsr:
Updated PR desc

ajm188 (approved):
lgtm

ajm188 (approved):
lgtm

deepthi (approved):

PR #12468, opened 2023-02-24
Fix scalar aggregation engine primitive for column truncation

harshit-gangal:
## Description

This PR fixes the issue on scalar aggregation primitive where the additional column added by the planner was not removed from the final result sent back to the client.

```
user query: select count(distinct col) from user
planned query: select col, weight_string(col) from user group by col, weight_string(col) order by col asc
```

Scaler Aggregation after aggregating the column count missed truncating the `weight_string(col)` column.

This was only missed for OLTP workload, for OLAP the logic was present.

## Checklist

-   [X] Did the new or modified tests pass consistently locally and on the CI

### DIFF
--- a/go/test/endtoend/vtgate/queries/aggregation/aggregation_test.go
+++ b/go/test/endtoend/vtgate/queries/aggregation/aggregation_test.go
@@ -398,3 +398,29 @@ func TestAggregateLeftJoin(t *testing.T) {
 	mcmp.AssertMatches("SELECT count(*) FROM t1 LEFT JOIN t2 ON t1.t1_id = t2.id", `[[INT64(2)]]`)
 	mcmp.AssertMatches("SELECT sum(t1.shardkey) FROM t1 LEFT JOIN t2 ON t1.t1_id = t2.id", `[[DECIMAL(1)]]`)
 }
+
+func TestScalarAggregate(t *testing.T) {
+	// disable schema tracking to have weight_string column added to query send down to mysql.
+	clusterInstance.VtGateExtraArgs = append(clusterInstance.VtGateExtraArgs, "--schema_change_signal=false")
+	require.NoError(t,
+		clusterInstance.RestartVtgate())
+
+	// update vtgate params
+	vtParams = clusterInstance.GetVTParams(keyspaceName)
+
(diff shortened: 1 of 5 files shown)


frouioui (approved):

harshit-gangal:
will manually backport to 14.0, 15.0, and main. post code-freeze

go/vt/vtgate/planbuilder/testdata/onecase.json
```diff
@@ -1,9 +1,65 @@
 [
   {
     "comment": "Add your test case here for debugging and run go test -run=One.",
-    "query": "",
-    "plan": {
-
+    "query": "select count(distinct foo) from user where id in (1,2,3)",
+    "v3-plan": {
```
GuptaManan100, line 5: onecase.json shouldn't have any commits.

go/test/endtoend/vtgate/queries/aggregation/aggregation_test.go
```diff
@@ -398,3 +398,30 @@ func TestAggregateLeftJoin(t *testing.T) {
...
+	require.NoError(t,
+		clusterInstance.RestartVtgate())
+
+	// update vtgate params
+	vtParams = clusterInstance.GetVTParams(keyspaceName)
+
+	defer func() {
+		// roll it back
+		clusterInstance.VtGateExtraArgs = append(clusterInstance.VtGateExtraArgs, "--schema_change_signal")
+		require.NoError(t,
```
GuptaManan100, line 415: won't we end up with flags like `vtgate ... --schema_change_signal=false --schema_change_signal`. Is that what we want?

GuptaManan100 (approved):

go/test/endtoend/vtgate/queries/aggregation/aggregation_test.go
harshit-gangal: the last flag setting will retain.

GuptaManan100: Cool 👍

rsajwani (commented):
LGTM

PR #12481, opened 2023-02-26
Fix `vtctldclient`'s Root command to return an error on unknown command

### LINKED ISSUE #12480: vtctldclient exits 0 when given invalid command
### Overview of the Issue

When given an invalid subcommand (either due to something completely out of left field, or just a typo), `vtctldclient` should should display usage, an error message, and exit non-zero. Currently it only displays usage.

### Reproduction Steps

`vtctldclient foobarbaz >/dev/null || echo $?` shows nothing, but should show a number other than zero.

### Binary Version

```sh
v15.0.2
v16-rc1
```

```sh
all OSes
```

### Log Fragments

_No response_

ajm188:
## Description

Add a Run func to `vtctldclient`'s Root command to return an error on unknown command.

Cobra can't (and shouldn't!) be distinguish between "we didn't find a subcommand because the user made an error" and "we didn't find a subcommand because the user is trying to invoke a parent command with positional arguments", so we specify a run function (what cobra invokes when executing the command) on the root (top-level parent) and check if there are any positional arguments, which in our case we assume means the user was trying to run a subcommand that doesn't exist.

## Related Issue(s)

Closes #12480.

## Checklist

### DIFF
--- a/go/cmd/vtctldclient/command/root.go
+++ b/go/cmd/vtctldclient/command/root.go
@@ -19,6 +19,7 @@ package command
 import (
 	"context"
 	"errors"
+	"fmt"
 	"io"
 	"strconv"
 	"time"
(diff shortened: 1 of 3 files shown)


GuptaManan100 (approved):

GuptaManan100 (commented):
Can we add a test to verify that it indeed returns an error?

ajm188:
> Can we add a test to verify that it indeed returns an error?

Done!

GuptaManan100:
Sweet! 🥇

go/cmd/vtctldclient/command/root.go
```diff
@@ -78,6 +79,26 @@ var (
...
+		// we've structured things can only be a subcommand name, **and**
+		//
+		// (2) Cobra was unable to find a subcommand with that name for which to
+		// call a Run or RunE function.
+		//
+		// From this we conclude that the user was trying to either run a
+		// command that doesn't exist (e.g. "vtctldclient delete-my-data") or
+		// has misspelled a legitimate command (e.g. "vtctldclient StapReplication").
+		// If we think this has happened, return an error, which will get
+		// displayed to the user in main.go along with the usage.
```
mattlord, line 94: Not something that I'm suggesting we bring in here but we could at some point use the "did you mean X" functionality that cobra provides via the `command.SuggestionsMinimumDistance = X` option. Maybe you explored it and there were downsides?

mattlord (approved):
Thanks!

go/cmd/vtctldclient/command/root.go
ajm188: i never explored it but i definitely want to at some point!

Issue #12483, opened 2023-02-27
Bug Report: Regression in Query Support


GuptaManan100:
### Overview of the Issue

The query - 
``` MySQL
SELECT * FROM unsharded_a AS t1  JOIN (SELECT trim((SELECT MAX(name) FROM unsharded_a)) AS name) AS t2 WHERE t1.name >= t2.name ORDER BY t1.name ASC LIMIT 1;
```

This query used to work in release-13 of Vitess but stops working in release-14 onwards.

The error message we get is - 
```
"unsupported: '*' expression in cross-shard query"
```

This is however incorrect, since this query is run in an unsharded keyspace.

### Reproduction Steps

1. Setup an unsharded cluster
2. Setup a schema such that `uncharted` table has a `name` column
3. Run the query

### Binary Version

```sh
main, release-16, release-15 and release-14
```

### Operating System and Environment details

```sh
all
```

PR #12484, opened 2023-02-27
BugFix: Unsharded query using a derived table and a dual table

GuptaManan100:
## Description

This PR fixes the query support for the query described in the bug https://github.com/vitessio/vitess/issues/12483

## Related Issue(s)

- Fixes https://github.com/vitessio/vitess/issues/12483

## Checklist


frouioui (approved):

arthurschreiber (approved):

harshit-gangal (commented):
This fix is good. But, we should not rely too much on the planner shortcut of single unsharded.
The planner should be able to handle and plan this query even if it goes through full planning.

dbussink (approved):

PR #12494, opened 2023-02-27
streamlog: make generic

vmg:
## Description

The `logstats` package is a good example of a primitive that is used throughout the Vitess codebase and that still uses `any` for polymorphism. It's un-ergonomic and it has a lot of type checks that won't be needed if we switch to use Generics.

## Checklist

### DIFF
--- a/go/vt/vtgate/querylogz.go
+++ b/go/vt/vtgate/querylogz.go
@@ -17,8 +17,6 @@ limitations under the License.
 package vtgate
 
 import (
-	"fmt"
-	"io"
 	"net/http"
 	"strconv"
 	"strings"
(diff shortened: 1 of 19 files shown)


go/vt/vtgate/querylogz.go
```diff
@@ -100,21 +98,12 @@ func querylogzHandler(ch chan any, w http.ResponseWriter, r *http.Request) {
...
 			}
-			stats, ok := out.(*logstats.LogStats)
-			if !ok {
-				err := fmt.Errorf("unexpected value in %s: %#v (expecting value of type %T)", QueryLogger.Name(), out, &logstats.LogStats{})
-				_, _ = io.WriteString(w, `<tr class="error">`)
-				_, _ = io.WriteString(w, err.Error())
-				_, _ = io.WriteString(w, "</tr>")
-				log.Error(err)
-				continue
-			}
```
vmg, line 117: :fire: this error handling. Let the type system do its job!

harshit-gangal (approved):

dbussink (approved):

PR #12502, opened 2023-02-27
Flakes: Use new healthy shard check in vreplication e2e tests

mattlord:
## Description

I have noticed that e.g. the [`vreplication_basic` CI workflow](https://github.com/vitessio/vitess/blob/main/.github/workflows/cluster_endtoend_vreplication_basic.yml) has become less reliable since [the SidecarDB init work](https://github.com/vitessio/vitess/pull/11520) was merged. The seen failures were always on the [`TestVreplicationCopyThrottling`](https://github.com/vitessio/vitess/blob/1d0b76e1898f113796f0e95ecdb8a149b2dd04cb/go/test/endtoend/vreplication/vreplication_test.go#L208) test. The reason is because we now needed a more reliable check for a healthy shard.

This is needed because checking that there's a primary tablet for the shard in vtgate's healtcheck — which is what [`vtgate.WaitForStatusOfTabletInShard()`](https://github.com/vitessio/vitess/blob/43e4222e87347583ecb72276d15bf309b003cda5/go/test/endtoend/cluster/vtgate_process.go#L204-L220) does — is no longer a reliable indicator that the shard has a healthy serving primary, because now after a primary is elected it needs to initialize its sidecar database and wait for those DDLs to replicate via semi-sync replication before it becomes serving and can proceed to perform normal functions. So this delay could cause test flakiness if you required a healthy shard before continuing with the test as was the case in [`TestVreplicationCopyThrottling`](https://github.com/vitessio/vitess/blob/1d0b76e1898f113796f0e95ecdb8a149b2dd04cb/go/test/endtoend/vreplication/vreplication_test.go#L208).

This PR migrates all of the places in the `vreplication` endtoend tests where we were using this to wait for a healthy shard:
`vtgate.WaitForStatusOfTabletInShard(fmt.Sprintf("%s.%s.primary", keyspace, shard), 1)` to instead using the new `WaitForHealthyShard` cluster package helper:
```
err := cluster.WaitForHealthyShard(vc.VtctldClient, keyspace, shard)
require.NoError(t, err)
```

>**Note** After these changes, the [`vreplication_basic` CI workflow](https://github.com/vitessio/vitess/blob/main/.github/workflows/cluster_endtoend_vreplication_basic.yml) e.g. has passed every time: https://github.com/vitessio/vitess/actions/runs/4297831822

## Related Issue(s)

  - Follow-up to: https://github.com/vitessio/vitess/pull/11520

## Checklist


deepthi (approved):

deepthi:
Looks like a good change. Do we have a sense for how long the schema init is taking? IIRC, in local testing it was only 1-2 seconds.

frouioui (approved):

mattlord:
> Looks like a good change. Do we have a sense for how long the schema init is taking? IIRC, in local testing it was only 1-2 seconds.

Yeah, 1-3 seconds seems normal locally. But that is plenty of time for races. And sometimes we encounter very interesting timings on Actions runners. 🥲

Issue #12504, opened 2023-02-28
Bug Report: TableGC purge race condition and an endless attempt to purge non-existing table


shlomi-noach:
### Overview of the Issue

The [Table GC (table garbage collector)](https://vitess.io/docs/16.0/user-guides/schema-changes/table-lifecycle/) has a race condition when it comes to purging `_vt_PURGE_*` tables. The following scenario is possible:

- The server runs on a MySQL `8.0.23`  or above - which means MySQL supports "fast drop table" and TableGC is aware of that. Which means TableGC is supposed to skip the `PURGE` state.
- Some table gets named as `_vt_PURGE_*` for some reason. Either it's a leftover from a previous configuration, or because it's explicitly created like so [here](https://github.com/vitessio/vitess/blob/1d0b76e1898f113796f0e95ecdb8a149b2dd04cb/go/vt/vttablet/onlineddl/executor.go#L3812).
- `checkTables()` finds the table [and adds it](https://github.com/vitessio/vitess/blob/1d0b76e1898f113796f0e95ecdb8a149b2dd04cb/go/vt/vttablet/tabletserver/gc/tablegc.go#L423) to in-memory `purgingTables` map
- Also, `shouldTransition` [is `true` for said table](https://github.com/vitessio/vitess/blob/1d0b76e1898f113796f0e95ecdb8a149b2dd04cb/go/vt/vttablet/tabletserver/gc/tablegc.go#L403)
- In a goroutine, called by a ticker, we iterate `purgingTables` and [find that table](https://github.com/vitessio/vitess/blob/1d0b76e1898f113796f0e95ecdb8a149b2dd04cb/go/vt/vttablet/tabletserver/gc/tablegc.go#L268).
- In that same `purge()` function we also [attempt to delete rows from the table](https://github.com/vitessio/vitess/blob/1d0b76e1898f113796f0e95ecdb8a149b2dd04cb/go/vt/vttablet/tabletserver/gc/tablegc.go#L505-L526).
- But meanwhile, because `PURGE` isn't one of the working states, the table is transitioned into `EVAC` or `DROP` state.
- The attempt to delete rows from the table fails with error: the table does not exist. That's the race condition.
- But then, an error is returned and the purge goroutine [exits](https://github.com/vitessio/vitess/blob/1d0b76e1898f113796f0e95ecdb8a149b2dd04cb/go/vt/vttablet/tabletserver/gc/tablegc.go#L269-L272)
[...]

PR #12505, opened 2023-02-28
TableGC: fix PURGE race condition

shlomi-noach:
## Description

See https://github.com/vitessio/vitess/issues/12504 for problem description

This PR fixes the race condition by eliminating the execution of one code path as follows:

If `PURGE` isn't a handled state (e.g. running on MySQL `8.0.23` and above), then we now never populate the in-memory `purgingTables` map. It's left empty at all times.

As such, the goroutine that runs on `purgeRequestsChan` has nothing to do, as it attempts to pick a table out of said in-memoty map, which is empty. So that code path is eliminated.

`checkTables()` is now the only owner for any existing `_vt_PURGE_*` tables, should these ever exist. It's fine if they do exist, and `checkTables()` will transition them into the next state.

Speaking of which, the fact the Online DDL Executor used to explicitly rename tables into `_vt_PURGE_*` is strictly speaking OK, not a bug, but then again it's also wasteful. going forward, we expect `PURGE` state to go extinct (as we disable it by default in `8.0.23`). So why bother creating a table in `PURGE` state.

So in this PR we also change the Online DDL Executor behavior: instead of renaming an artifact table into a `_vt_PURGE...`, it renames it into a `_vt_HOLD...` with a timestamp that is in the past. This way, next iteration of TableGC will transition the table into next state. To clarify, `_vt_PURGE` is also fine, and Table GC would transition that to the next state, in very much the same way. It's just that it's surprising to see a `_vt_PURGE` table when `PURGE` is not a supported  state. I think it's now cleaner.

Lastly, added an `Info` log entry upon `Open()` that clarifies the MySQL version and the supported states.

## Related Issue(s)

Fixes https://github.com/vitessio/vitess/issues/12504

## Checklist


shlomi-noach:
All `Upgrade Downgrade` tests failing because of `golang` version. To be ignored.

dbussink (approved):

go/vt/vttablet/onlineddl/executor.go
```diff
@@ -3799,17 +3799,21 @@ func (e *Executor) deleteVReplicationEntry(ctx context.Context, uuid string) err
 }
 
 // gcArtifactTable garbage-collects a single table
-func (e *Executor) gcArtifactTable(ctx context.Context, artifactTable, uuid string, t time.Time) (string, error) {
+func (e *Executor) gcArtifactTable(ctx context.Context, artifactTable, uuid string, timestampInThePast time.Time) (string, error) {
```
deepthi, line 3802: nit: it seems weird to rename this function parameter. Those are the semantics the caller is using, this function simply generates the rename statement with whatever time is given to it.

deepthi (approved):
Just one comment, the rest LGTM

go/vt/vttablet/onlineddl/executor.go
shlomi-noach: renamed back to original generic name `t`

PR #12519, opened 2023-02-28
`ApplyVSchemaDDL`: escape Sequence names when writing the VSchema

hkdsun:
## Description

We use `ALTER VSCHEMA ... ADD AUTO_INCREMENT ...` statements to provision new sharded tables with their Sequence table references, however we ran into a bug because our sequence tables are stored in keyspaces that contain dashes (e.g. <code>\`app-name-global\`.sequence_table</code>).

The issue is that the `ApplyVSchemaDDL` function does not escape the value of the  `Sequence` field below even though it's capable of parsing user input correctly:

```go
		table.AutoIncrement = &vschemapb.AutoIncrement{
			Column:   alterVschema.AutoIncSpec.Column.String(),
			Sequence: sequenceFqn,
		}
```

Not surprising, since we're going through a cycle of parsing user input (from a SQL statment) and then trying to write something out to the VSchema's `Table.AutoIncrement.Sequence` field with the (new-ish?) understanding that it will be parsed using `sqlparser` again. As [the VSchema docs mention](https://vitess.io/docs/15.0/reference/features/vschema/), "If necessary, the reference to the sequence table lookup.user_seq can be escaped using backticks."

We would _really_ love to have this be backported and have a `v14.0.5` minor version be released with this fix.

## Checklist

## Deployment Notes

There should be no functional change in correct existing behaviour. The failing case will be fixed by the changes.

### DIFF
--- a/go/vt/topotools/vschema_ddl.go
+++ b/go/vt/topotools/vschema_ddl.go
@@ -17,7 +17,6 @@ limitations under the License.
 package topotools
 
 import (
-	"fmt"
 	"reflect"
 
 	"vitess.io/vitess/go/vt/sqlparser"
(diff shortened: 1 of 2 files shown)


deepthi:
@vitessio/query-serving please review

hkdsun:
I've just pushed a new commit to solve this more elegantly.

One might think "this should have always been written as such" 😄

Could I please get a review? And any word on whether this could be backported? Our team at Shopify would have to otherwise plan a forked release fairly soon.

Thanks everyone

deepthi (approved):
Elegant is correct 😄 
LGTM, and I think we can backport.
We'll wait to get at least 1 more approval from the query-serving team.

PR #12525, opened 2023-03-01
Use `MarshalVT`/`UnmarshalVT` instead of `proto.Marshal`/`proto.Unmarshal`.

arthurschreiber:
## Description

Use the more efficient `MarshalVT`/`UnmarshalVT` functions instead of `proto.Marshal`/`proto.Unmarshal`.

The only places I didn't switch over are `go/vt/servenv/grpc_codec.go` (because that is used by the grpc communication for `etcd-client`) and `go/vt/topo/decode.go` (which I can switch over if anyone cares about that).

## Related Issue(s)

N/A

## Checklist

### DIFF
--- a/go/vt/tableacl/tableacl.go
+++ b/go/vt/tableacl/tableacl.go
@@ -113,7 +113,7 @@ func (tacl *tableACL) init(configFile string, aclCB func()) error {
 		return err
 	}
 	config := &tableaclpb.Config{}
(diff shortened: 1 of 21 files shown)


vmg (approved):
LGTM. I don't think this is gonna fix the corruption but you're seeing in prod, @arthurschreiber, but it'll be a nice optimization.

frouioui (approved):

Issue #12529, opened 2023-03-01
Bug Report: Failing `vitess/base` docker build due to git error


frouioui:
### Overview of the Issue

Since the recent `go1.20.1` the `vitess/base` Docker build, on both `main` and `release-16.0` branches, started failing with the following `git` error:
```
fatal: detected dubious ownership in repository at '/vt/src/vitess.io/vitess'
To add an exception for this directory, call:

        git config --global --add safe.directory /vt/src/vitess.io/vitess
```

The `go build` command has a flag that chooses whether we want to use version control or not for the build: `-buildvcs`. Its default value is set to `auto`, however, in this context it was unable to switch to `false`. In the Docker build logs we can see the following:
```
Use -buildvcs=false to disable VCS stamping.
error obtaining VCS status: exit status 128
```

A workaround for this issue was to manually set `-buildvcs=false`. A second workaround is to add `/vt/src/vitess.io/vitess` as a safe `git` directory in our `git` config.

### Reproduction Steps

Run the following command on `main` or `release-16.0` at the root of the repo:

```sh
docker build --platform linux/amd64 -f docker/base/Dockerfile  -t myBaseTest .
```

### Binary Version

```sh
main, release-16.0
```

### Operating System and Environment details

```sh
Docker
```

### Log Fragments

```sh
2023-03-01T09:12:06Z #9 0.419 git config --global --add safe.directory /vt/src/vitess.io/vitess
2023-03-01T09:12:06Z #9 0.420 fatal: detected dubious ownership in repository at '/vt/src/vitess.io/vitess'
2023-03-01T09:12:06Z #9 0.420 To add an exception for this directory, call:
2023-03-01T09:12:06Z #9 0.420
2023-03-01T09:12:06Z #9 0.420 git config --global --add safe.directory /vt/src/vitess.io/vitess
2023-03-01T09:12:06Z #9 0.433 Wed Mar 1 09:12:06 UTC 2023: Building source tree
2023-03-01T09:12:06Z #9 0.487 fatal: not in a git directory
2023-03-01T09:12:09Z #9 2.786 go: downloading golang.org/x/term v0.5.0
2023-03-01T09:12:09Z #9 2.897 go: downloading golang.org/x/text v0.7.0
2023-03-01T09:12:11Z #9 5.163 go: downloading golang.org/x/sync v0.1.0
2023-03-01T09:12:12Z #9 6.425 go: downloading golang.org/x/net v0.7.0
[...]

PR #12530, opened 2023-03-01
Fix dubious ownership of git directory in `vitess/base` Docker build

frouioui:
## Description

This Pull Request adds the `git config --global --add safe.directory /vt/src/vitess.io/vitess` to our `vitess/base` Dockerfile in order to trust the `git` directory.

## Related Issue(s)

Fixes https://github.com/vitessio/vitess/issues/12529

## Checklist

### DIFF
--- a/docker/base/Dockerfile
+++ b/docker/base/Dockerfile
@@ -33,6 +33,9 @@ ARG BUILD_GIT_REV
 USER root
 COPY . /vt/src/vitess.io/vitess
 
+# Trust dubious ownership of git directory
+RUN git config --global --add safe.directory /vt/src/vitess.io/vitess
+
 # Build Vitess
 RUN make build
 


docker/base/Dockerfile
```diff
@@ -33,6 +33,9 @@ ARG BUILD_GIT_REV
 USER root
 COPY . /vt/src/vitess.io/vitess
 
+# Trust dubious ownership of git directory
+RUN git config --global --add safe.directory /vt/src/vitess.io/vitess
+
```
arthurschreiber, line 38: Alternatively, you could specify the owner on the `COPY` command like this:

```suggestion
COPY --chown=root . /vt/src/vitess.io/vitess
```

deepthi: I was wondering why we don't run into this problem with `vitess/lite` and these comments explain why.
`vitess/lite` has the following
```
# Re-copy sources from working tree.
COPY --chown=vitess:vitess . /vt/src/vitess.io/vitess
```
So another +1 to this recommended solution.

frouioui: Thanks all for this recommendation @mattlord @deepthi @arthurschreiber ❤️ I pushed a fix in fc7d57e110.

arthurschreiber (approved):
Looking great! Can you also update the PR description to match? 🙇

deepthi (approved):

PR #12534, opened 2023-03-01
Fix initialization code to also stop replication to prevent crash

### LINKED ISSUE #10880: Bug Report: Semi-sync not setup correctly on VTTablet startup

### LINKED ISSUE #12533: Bug Report: Vttablet restart when MySQL replication is running crashes
### Overview of the Issue

Restarting a vttablet when MySQL replication is running crashes the vttablet.
This happens only when `--restore_from_backup` flag is not provided or set to false.

The logs say - 
```
I0301 22:52:17.809336   63118 tm_init.go:871] Getting a new durability policy for semi_sync
I0301 22:52:17.809355   63118 replication.go:633] Setting semi-sync mode: primary=false, replica=true
I0301 22:52:17.809717   63118 query.go:80] exec SET GLOBAL rpl_semi_sync_master_enabled = 0, GLOBAL rpl_semi_sync_slave_enabled = 1
I0301 22:52:17.810129   63118 query.go:80] exec RESET SLAVE ALL
E0301 22:52:17.810274   63118 query.go:82] ExecuteFetch(RESET SLAVE ALL) failed: This operation cannot be performed with running replication threads; run STOP SLAVE FOR CHANNEL '' first (errno 3081) (sqlstate HY000) during query: RESET SLAVE ALL
I0301 22:52:17.810310   63118 tm_state.go:101] In tmState.Open()

F0301 22:52:17.853669   63118 vttablet.go:128] failed to parse --tablet-path or initialize DB credentials: ExecuteFetch(RESET SLAVE ALL) failed: This operation cannot be performed with running replication threads; run STOP SLAVE FOR CHANNEL '' first (errno 3081) (sqlstate HY000) during query: RESET SLAVE ALL
MysqlDaemon.SetReplicationSource failed
```

This bug was originally introduced by https://github.com/vitessio/vitess/pull/10881

### Reproduction Steps

Start a cluster.
Restart a running vttablet.

### Binary Version

```sh
main, release-16, release-15
```

```sh
all
```

### Log Fragments

_No response_

GuptaManan100:
## Description

This PR fixes the bug described in #12533 by fixing the initialization code to also stop replication before it tries to `reset slave  all`. The old code assumed that replication is already stopped when a new vttablet was coming up, but that assumption is incorrect. We should stop replication irrespective before we try to change the primary hostname, and port.

This PR is a follow-up to: https://github.com/vitessio/vitess/pull/10881 which was originally created to resolve #10880.

## Related Issue(s)

 - Fixes #12533 

## Checklist

### DIFF
--- a/go/test/endtoend/reparent/newfeaturetest/reparent_test.go
+++ b/go/test/endtoend/reparent/newfeaturetest/reparent_test.go
@@ -96,3 +96,16 @@ func TestSingleReplicaERS(t *testing.T) {
 	// Also check the writes succeed after failover
 	utils.ConfirmReplication(t, tablets[0], []*cluster.Vttablet{})
 }
+
+// TestTabletRestart tests that a running tablet can be  restarted and everything is still fine
+func TestTabletRestart(t *testing.T) {
+	defer cluster.PanicHandler(t)
+	clusterInstance := utils.SetupReparentCluster(t, "semi_sync")
+	defer utils.TeardownCluster(clusterInstance)
+	tablets := clusterInstance.Keyspaces[0].Shards[0].Vttablets
+
+	utils.StopTablet(t, tablets[1], false)
+	tablets[1].VttabletProcess.ServingStatus = "SERVING"
+	err := tablets[1].VttabletProcess.Setup()
+	require.NoError(t, err)
+}
--- a/go/vt/vttablet/tabletmanager/tm_init.go
+++ b/go/vt/vttablet/tabletmanager/tm_init.go
@@ -882,7 +882,7 @@ func (tm *TabletManager) initializeReplication(ctx context.Context, tabletType t
 		log.Warningf("primary tablet in the shard record does not have mysql hostname specified, possibly because that tablet has been shut down.")
 		return nil, nil
 	}
-	if err := tm.MysqlDaemon.SetReplicationSource(ctx, currentPrimary.Tablet.MysqlHostname, currentPrimary.Tablet.MysqlPort, false /* stopReplicationBefore */, true /* startReplicationAfter */); err != nil {
+	if err := tm.MysqlDaemon.SetReplicationSource(ctx, currentPrimary.Tablet.MysqlHostname, currentPrimary.Tablet.MysqlPort, true /* stopReplicationBefore */, true /* startReplicationAfter */); err != nil {
 		return nil, vterrors.Wrap(err, "MysqlDaemon.SetReplicationSource failed")
 	}
 
--- a/go/vt/vtctl/grpcvtctldserver/endtoend/init_shard_primary_test.go
+++ b/go/vt/vtctl/grpcvtctldserver/endtoend/init_shard_primary_test.go
@@ -60,6 +60,7 @@ func TestInitShardPrimary(t *testing.T) {
 
 	tablet2.FakeMysqlDaemon.ExpectedExecuteSuperQueryList = []string{
 		// These come from tablet startup
+		"STOP SLAVE",
 		"RESET SLAVE ALL",
 		"FAKE SET MASTER",
 		"START SLAVE",
@@ -73,6 +74,7 @@ func TestInitShardPrimary(t *testing.T) {
 	tablet2.FakeMysqlDaemon.SetReplicationSourceInputs = append(tablet2.FakeMysqlDaemon.SetReplicationSourceInputs, fmt.Sprintf("%v:%v", tablet1.Tablet.Hostname, tablet1.Tablet.MysqlPort))
 
 	tablet3.FakeMysqlDaemon.ExpectedExecuteSuperQueryList = []string{
+		"STOP SLAVE",
 		"RESET SLAVE ALL",
 		"FAKE SET MASTER",
 		"START SLAVE",
--- a/go/vt/vttablet/tabletmanager/tm_init_test.go
+++ b/go/vt/vttablet/tabletmanager/tm_init_test.go
@@ -381,6 +381,7 @@ func TestCheckPrimaryShip(t *testing.T) {
 	fakeMysql := tm.MysqlDaemon.(*mysqlctl.FakeMysqlDaemon)
 	fakeMysql.SetReplicationSourceInputs = append(fakeMysql.SetReplicationSourceInputs, fmt.Sprintf("%v:%v", otherTablet.MysqlHostname, otherTablet.MysqlPort))
 	fakeMysql.ExpectedExecuteSuperQueryList = []string{
+		"STOP SLAVE",
 		"RESET SLAVE ALL",
 		"FAKE SET MASTER",
 		"START SLAVE",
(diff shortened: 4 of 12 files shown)


go/test/endtoend/reparent/newfeaturetest/reparent_test.go
```diff
@@ -96,3 +96,16 @@ func TestSingleReplicaERS(t *testing.T) {
...
+	defer cluster.PanicHandler(t)
+	clusterInstance := utils.SetupReparentCluster(t, "semi_sync")
+	defer utils.TeardownCluster(clusterInstance)
+	tablets := clusterInstance.Keyspaces[0].Shards[0].Vttablets
+
+	utils.StopTablet(t, tablets[1], false)
+	tablets[1].VttabletProcess.ServingStatus = "SERVING"
+	err := tablets[1].VttabletProcess.Setup()
+	require.NoError(t, err)
+}
```
mattlord, line 111: This fails on `main`?

go/vt/vttablet/tabletmanager/tm_init.go
```diff
@@ -882,7 +882,7 @@ func (tm *TabletManager) initializeReplication(ctx context.Context, tabletType t
 		log.Warningf("primary tablet in the shard record does not have mysql hostname specified, possibly because that tablet has been shut down.")
 		return nil, nil
 	}
-	if err := tm.MysqlDaemon.SetReplicationSource(ctx, currentPrimary.Tablet.MysqlHostname, currentPrimary.Tablet.MysqlPort, false /* stopReplicationBefore */, true /* startReplicationAfter */); err != nil {
+	if err := tm.MysqlDaemon.SetReplicationSource(ctx, currentPrimary.Tablet.MysqlHostname, currentPrimary.Tablet.MysqlPort, true /* stopReplicationBefore */, true /* startReplicationAfter */); err != nil {
```
mattlord, line 885: Looks like this was introduced in https://github.com/vitessio/vitess/pull/10881 then (so 15.0.0 and later). We should note that in the description and in the Related Issues. Suggestion there:
```
  - Follow-up to: https://github.com/vitessio/vitess/pull/10881
```

That change was made to address: https://github.com/vitessio/vitess/issues/10880

mattlord (approved):
Approving as it's definitely an improvement. I'm not certain, however, that this eliminates the general race condition as a human, tablet repair, vtorc, etc could still start replication in between right? I feel like perhaps we should continue retrying to achieve desired config and error after a number of attempts.

mattlord:
From the issue, does the process end here?
```
F0301 22:52:17.853669   63118 vttablet.go:128] failed to parse --tablet-path or initialize DB credentials: ExecuteFetch(RESET SLAVE ALL) failed: This operation cannot be performed with running replication threads; run STOP SLAVE FOR CHANNEL '' first (errno 3081) (sqlstate HY000) during query: RESET SLAVE ALL
MysqlDaemon.SetReplicationSource failed
```
If so, then I would guess we have a nil pointer dereference or something that flows from this. That would seem like a separate issue to address, as ideally we should retry to reach the desired state during init and then shutdown gracefully if we fail to. Otherwise I think we may still be still prone to the process simply “disappearing” w/o a trace or clue because of race conditions around this.

go/test/endtoend/reparent/newfeaturetest/reparent_test.go
GuptaManan100: Yes, it does.

go/vt/vttablet/tabletmanager/tm_init.go
GuptaManan100: Okay 👍

GuptaManan100:
@mattlord It's not really a race. Its just that this part of code didn't stop replication before it started changing the primary information. It assumed that replication would be stopped already, which is an incorrect assumption.
As far as the question for a race with someone fixing the replication (vtorc, or manual) is concerned, I don't think that is an issue because of how `SetReplicationSource` is written. It accumulates all the commands it has to run and then runs them all one after the other.
```go
        cmds := []string{}
	if replicationStopBefore {
		cmds = append(cmds, conn.StopReplicationCommand())
	}
	// Reset replication parameters commands makes the instance forget the source host port
	// This is required because sometimes MySQL gets stuck due to improper initialization of
	// master info structure or related failures and throws errors like
	// ERROR 1201 (HY000): Could not initialize master info structure; more error messages can be found in the MySQL error log
	// These errors can only be resolved by resetting the replication parameters, otherwise START SLAVE fails.
	// Therefore, we have elected to always reset the replication parameters whenever we try to set the source host port
	// Since there is no real overhead, but it makes this function robust enough to also handle failures like these.
	cmds = append(cmds, conn.ResetReplicationParametersCommands()...)
	smc := conn.SetReplicationSourceCommand(params, host, port, int(replicationConnectRetry.Seconds()))
	cmds = append(cmds, smc)
	if replicationStartAfter {
		cmds = append(cmds, conn.StartReplicationCommand())
	}
	return mysqld.executeSuperQueryListConn(ctx, conn, cmds)
```

I think that the idea of having a retry is still a good one. We could also just ignore the error, instead of exiting on it and let VTOrc repair the replication later.

This is what we have in `Start` - 
```
_, err = tm.initializeReplication(ctx, tm.Tablet().Type)
	tm.tmState.Open()
	return err
```
Maybe we should ignore that error, because that error gets propagated up the stack eventually doing `log.Exitf`. An error in initializing replication shouldn't really shutdown the tablet. I am still in two-minds about that though. WDYT?

GuptaManan100:
@mattlord Me and @deepthi had a chat about this PR today morning. She also feels that we shouldn't be ignoring an error during start-up. Also, we both think that a retry for initializing replication isn't required especially since VTOrc has already become compulsory. 

Other than that, I have pushed fixes to the tests so this PR should be good to go. If something does turn up, please feel free to resolve it as you both see fit since I'll be off for about 11 days.

deepthi (approved):

PR #12539, opened 2023-03-02
OnlineDDL: optimizing --singleton-context conflict check

shlomi-noach:
## Description

An optimization PR.

Intro: When we submit an Online DDL migration, and with `--singleton-context`, the submission is only accepted if there's no pending migration with a different migration context than the submitted migration. i.e., the migration is only ever accepted if all other pending migrations (queued, ready, running) have the exact same migration-context as the submitted migration.

Sometimes we will submit a large number of migrations, e.g. when importing a schema. Say we have `200` `CREATE TABLE` statements. Whether one by one, or whether by `ApplySchema --sql="many; migrations; here;"`, the Online DDL executor is asked to `SubmitMigration()` for each and every DDL,, one at a time.

For each such migration, the executor scans the existing list of pending migrations to see if there's a conflicting migration.

Problem: for the first migration it scans and finds `0` pending migration. For the 2nd migration is scans and find `1` pending migration, ... for the 200th migration it scans and finds `199` pending migrations. And for each found pending migration, it reads that migration to check for conflict. This has `O(n^2)` complexity and more importantly `O(n^2)` database reads.

This PR offers a simple optimization. We're still at `O(n^2)` complexity, the algorithm isn't changed that much. It's just that were able to determine that a pending migration does not, in fact, conflict with the submitted migration, sonner than before, and without having to actually read the migration.

This is done by pre-fetching `migration_context` column; if the value is the same as our migration's, then by definition there's no conflict and no need to read each and every pending migration.

In the "happy path", this does reduce database access to `O(n)`, ie we only need to scan the pending migrations `n` times (although, we still traverse `O(n^2)` elements in-memory). The happy path is actually common, and so this is a win.

On a local dev env, which is otherwise quiet and idle, this reduces overall submission time as follows:

- Without the optimization, submitting `100` migrations takes `4.5` seconds, submitting another `100` migrations (still in same migration context) takes `8sec`) totaling some `12.5` sec. Notice the 2nd batch is significantly slower than first batch.
- With the optimization,  submitting `100` migrations takes about `2.1 - 2.4` seconds, submitting another `100` migrations (still in same migration context) takes similarly `2.1 - 2.4` seconds, totaling about `5sec`. Notice the 2nd batch has same runtime as first batch.

As the number of DDLs per migration-context increases, so does the gain of the optimization.

## Checklist

### DIFF
--- a/go/vt/vttablet/onlineddl/executor.go
+++ b/go/vt/vttablet/onlineddl/executor.go
@@ -4544,19 +4544,29 @@ func (e *Executor) submitCallbackIfNonConflicting(
 		e.migrationMutex.Lock()
 		defer e.migrationMutex.Unlock()
 
-		pendingUUIDs, err := e.readPendingMigrationsUUIDs(ctx)
+		rs, err := e.execQuery(ctx, sqlSelectPendingMigrations)
 		if err != nil {
 			return err
 		}
+		rows := rs.Named().Rows
+
 		switch {
 		case onlineDDL.StrategySetting().IsSingleton():
 			// We will reject this migration if there's any pending migration
-			if len(pendingUUIDs) > 0 {
-				return vterrors.Errorf(vtrpcpb.Code_FAILED_PRECONDITION, "singleton migration rejected: found pending migrations [%s]", strings.Join(pendingUUIDs, ", "))
+			if len(rows) > 0 {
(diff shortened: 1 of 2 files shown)


dbussink (approved):

rohit-nayak-ps (approved):
lgtm

PR #12543, opened 2023-03-02
handle filter on top of UNION

systay:
## Description
When planning for filtering on top of a derived table that contains an UNION, the planner was just giving up. This change is not the most optimal fix, since we should be able to push the filtering into the route, but at least this plan works.

We need to forward port this fix to `release-16.0` and `main`.

## Related Issue(s)
-  https://github.com/vitessio/vitess/issues/9214

## Checklist

### DIFF
--- a/go/vt/vtgate/planbuilder/abstract/concatenate.go
+++ b/go/vt/vtgate/planbuilder/abstract/concatenate.go
@@ -47,6 +47,20 @@ func (c *Concatenate) TableID() semantics.TableSet {
 
 // PushPredicate implements the Operator interface
 func (c *Concatenate) PushPredicate(expr sqlparser.Expr, semTable *semantics.SemTable) (LogicalOperator, error) {
+	op, err := c.tryPush(expr, semTable)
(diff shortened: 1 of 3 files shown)


frouioui:
We don't have a `Forwardport to: xxx` for release branch, the `Backport to: ` will do the same job.

frouioui (approved):

dbussink (approved):

harshit-gangal (dismissed):
I tried running the query reported on the issue. It still throws unsupported error
`unsupported: JOIN not supported between derived tables`

harshit-gangal:
> I tried running the query reported on the issue. It still throws unsupported error `unsupported: JOIN not supported between derived tables`

specifically this on https://github.com/vitessio/vitess/issues/9214#issuecomment-973937494

harshit-gangal (approved):

frouioui:
We decided there is no need to forwardport this change to `main` and `release-16.0`, the bug is not on these branches.

PR #12546, opened 2023-03-03
Flakes: Address Common Unit Test Races

mattlord:
## Description

The [unit race workflow](https://github.com/vitessio/vitess/blob/main/.github/workflows/unit_race.yml) was [failing pretty regularly in the CI](https://github.com/vitessio/vitess/actions/workflows/unit_race.yml) (~20% of the time), and the two failures that I would regularly see were:
  1. `testVDiffEnv` related races — most often in the wrangler unit tests, as they  have the widest use
      - These were simply NOT concurrency-safe
  2. `glog.MaxSize` related races — most often in the wrangler unit tests as they create a LOT of loggers, primarily ConsoleLoggers and MemoryLoggers
      - The way we modified the `glog.MaxSize` global exported variable in our `--log_rotate_max_size` flag handling was not concurrency-safe

I was able to quickly repeat these failures locally as well using the method below. After the changes in this PR, I was able to run it successfully 100 times in a row:
```
cnt=1
while true; do
    echo -n "Attempt# ${cnt} :: "
    go test -timeout 30s -parallel 6 -count 1 -race vitess.io/vitess/go/vt/wrangler 2>>/tmp/results
    if [[ $? -ne 0 ]]; then
         echo "Failed on attempt ${cnt}"
         break
    fi
    let cnt+=1
    sleep 1
done
```
```
Attempt# 1 :: ok  	vitess.io/vitess/go/vt/wrangler	9.320s
...
Attempt# 100 :: ok  	vitess.io/vitess/go/vt/wrangler	9.503s
```

It was failing in the CI about 1 in 4-5 attempts. In this PR it [passed 19 out of 20 times, 18 times in a row](https://github.com/vitessio/vitess/actions/runs/4324961110/jobs/7566860171) — which is at least a big improvement.

## Checklist

### DIFF
--- a/go/vt/log/log.go
+++ b/go/vt/log/log.go
@@ -22,6 +22,9 @@ limitations under the License.
 package log
 
 import (
+	"strconv"
+	"sync/atomic"
+
 	"github.com/golang/glog"
 	"github.com/spf13/pflag"
 )
@@ -78,5 +81,32 @@ var (
 // calls this function, or call this function directly before parsing
 // command-line arguments.
 func RegisterFlags(fs *pflag.FlagSet) {
-	fs.Uint64Var(&glog.MaxSize, "log_rotate_max_size", glog.MaxSize, "size in bytes at which logs are rotated (glog.MaxSize)")
+	flagVal := logRotateMaxSize{
+		val: "1887436800", // glog.MaxSize value (which is not concurrency safe)
+	}
+	fs.Var(&flagVal, "log_rotate_max_size", "size in bytes at which logs are rotated (glog.MaxSize)")
+}
+
+// logRotateMaxSize implements pflag.Value and is used to
+// try and provide thread-safe access to glog.MaxSize.
+type logRotateMaxSize struct {
+	val string
+}
+
+func (lrms *logRotateMaxSize) Set(s string) error {
+	maxSize, err := strconv.ParseUint(s, 10, 64)
+	if err != nil {
+		return err
+	}
+	atomic.StoreUint64(&glog.MaxSize, maxSize)
+	lrms.val = s
+	return nil
+}
+
+func (lrms *logRotateMaxSize) String() string {
+	return lrms.val
+}
+
+func (lrms *logRotateMaxSize) Type() string {
+	return "uint64"
 }
--- a/go/vt/log/log.go
+++ b/go/vt/log/log.go
@@ -22,6 +22,7 @@ limitations under the License.
 package log
 
 import (
+	"fmt"
 	"strconv"
 	"sync/atomic"
 
@@ -82,7 +83,7 @@ var (
 // command-line arguments.
 func RegisterFlags(fs *pflag.FlagSet) {
 	flagVal := logRotateMaxSize{
-		val: "1887436800", // glog.MaxSize value (which is not concurrency safe)
+		val: fmt.Sprintf("%d", atomic.LoadUint64(&glog.MaxSize)),
 	}
 	fs.Var(&flagVal, "log_rotate_max_size", "size in bytes at which logs are rotated (glog.MaxSize)")
 }
--- a/go/vt/vtctl/vdiff_env_test.go
+++ b/go/vt/vtctl/vdiff_env_test.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"math/rand"
 	"sync"
+	"testing"
 
 	"vitess.io/vitess/go/sqltypes"
 	"vitess.io/vitess/go/vt/grpcclient"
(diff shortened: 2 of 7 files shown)


go/vt/log/log.go
```diff
@@ -78,5 +81,32 @@ var (
 // calls this function, or call this function directly before parsing
 // command-line arguments.
 func RegisterFlags(fs *pflag.FlagSet) {
-	fs.Uint64Var(&glog.MaxSize, "log_rotate_max_size", glog.MaxSize, "size in bytes at which logs are rotated (glog.MaxSize)")
+	flagVal := logRotateMaxSize{
+		val: "1887436800", // glog.MaxSize value (which is not concurrency safe)
+	}
+	fs.Var(&flagVal, "log_rotate_max_size", "size in bytes at which logs are rotated (glog.MaxSize)")
```
mattlord, line 87: Simply passing the literal as the default — rather than the `glog.MaxSize` var — is likely enough. If you prefer I can do that instead:
```
fs.Uint64Var(&glog.MaxSize, "log_rotate_max_size", 1887436800, "size in bytes at which logs are rotated (glog.MaxSize)")
```

go/vt/log/log.go
```diff
@@ -78,5 +81,32 @@ var (
 // calls this function, or call this function directly before parsing
 // command-line arguments.
 func RegisterFlags(fs *pflag.FlagSet) {
-	fs.Uint64Var(&glog.MaxSize, "log_rotate_max_size", glog.MaxSize, "size in bytes at which logs are rotated (glog.MaxSize)")
+	flagVal := logRotateMaxSize{
+		val: "1887436800", // glog.MaxSize value (which is not concurrency safe)
```
dbussink, line 85: Instead of copying the value here, can't we use `glob.MaxSize` itself here? Maybe with an atomic load then I guess?

dbussink (approved):

go/vt/log/log.go
ajm188: does this work?

```go
flagVal := logRotateMaxSize{val: fmt.Sprintf("%d", glog.MaxSize)}
fs.Var(&flagVal, ...)
```

that way if a glog version were to ever happen to change the value (unlikely, but nonzero), we'd get the change for free

go/vt/log/log.go
ajm188: oh, lol, come on GitHub PR UI https://github.com/vitessio/vitess/pull/12546#discussion_r1125686188

so, uhh, yeah I agree with @dbussink here 😂

mattlord: That was the first thing I tried and it didn't help.

go/vt/log/log.go
mattlord: It works, yes. But was still a little racy, even when using an atomic load there.

go/vt/log/log.go
mattlord: Giving it a go in the CI now (will do 20 test runs again): https://github.com/vitessio/vitess/pull/12546/commits/0c5be06fa97184e86118c9086fe4cf75e9de4bd5

IIRC, the race moved to pflag.Parse/TryParse around flag.TrickGlog but let's see.

go/vt/vtctl/vdiff_env_test.go
```diff
@@ -94,6 +80,17 @@ func newTestVDiffEnv(sourceShards, targetShards []string, query string, position
 	}
 	env.wr = wrangler.NewTestWrangler(env.cmdlog, env.topoServ, env.tmc)
 
+	dialerName := fmt.Sprintf("VDiffTest-%d", rand.Intn(1000000000))
```
ajm188, line 83: if we update this to take a `*testing.T` (or better, `testing.TB`), you can parameterize this to `fmt.Sprintf("VDiffTest-%s", t.Name())` and avoid any chance of collision entirely

go/vt/wrangler/wrangler_env_test.go
```diff
@@ -53,39 +54,34 @@ type testWranglerEnv struct {
...
-		tablets:    make(map[int]*testWranglerTablet),
 		topoServ:   memorytopo.NewServer("zone1"),
 		cell:       "zone1",
 		tabletType: topodatapb.TabletType_REPLICA,
 		tmc:        newTestWranglerTMClient(),
 	}
 	env.wr = New(logutil.NewConsoleLogger(), env.topoServ, env.tmc)
+	env.tmc.tablets = make(map[int]*testWranglerTablet)
+
+	dialerName := fmt.Sprintf("WranglerTest-%d", rand.Intn(1000000000))
```
ajm188, line 73: ditto here

ajm188 (approved):
LGTM, one small (potential) improvement, take it or leave it! and thanks!

go/vt/vtctl/vdiff_env_test.go
mattlord: Good idea! I did that here (kept the rand int part for too): https://github.com/vitessio/vitess/pull/12546/commits/2d7e509e104c4e0a640a96d01ebf3398cf0133c5

Thanks!

PR #12551, opened 2023-03-06
schemadiff: diffing two schemas generates a rich SchemaDiff object

shlomi-noach:
## Description

This is a `schemadiff` enhancement. Previously, when diffing two schemas (via `Schema.Diff(other)`) we'd return a list of `Diff` objects, each of which would represent a single change, .e.g an `ALTER TABLE`, `DROP VIEW`, etc. The diffs order was partly heuristic (`DROP VIEW` -> `DROP TABLE` -> `ALTER TABLE` -> `CREATE TABLE` -> `ALTER VIEW` -> `CREATE VIEW`) and partly evaluated (order views by dependencies). However, the order was not necessarily applicable.

This PR introduces a new `SchemaDiff` function which returns a new `SchemaDiff` object:

```golang
func (s *Schema) SchemaDiff(other *Schema, hints *DiffHints) (*SchemaDiff, error)
```

`SchemaDiff` is a rich object which includes all diffs between the two schemas, but also:

- Diffs are guaranteed to be sorted in lienarized, applicable order, if possible: you'd be able to apply the diffs one by one, and keep the schema consistent at all times while doing so, resolving any dependencies between the diffs. This includes `VIEW` and `FOREIGN KEY` dependencies.
- There are scenarios where the declarative diffs between two schema have no single linearized sequence off diffs. This happens because we only attempt a single `ALTER` operation per table/view. In the future, we may allow a breakdown into granular diffs.
- The diffs are arranged into equivalence classes, where diffs in the same equivalence class are said to depend on each other.
- "depend on each other" has a formal definition of increasing severity or strictness. Two diffs may not have any dependency, or may have an unknown dependency (the current algorithm does not lead to any such result), an in-order completion dependency, or a strictly sequential dependency. This information is made available by `SchemaDiff`. This lets the user/automation that gets `SchemaDiff` to know whether some migrations may run [`--in-order`](https://github.com/vitessio/vitess/pull/12113), or if one migration must strictly wait for another to complete.
- The algorithm follows MySQL's capabilities; it attempts the least strict dependency type between any two diffs, where possible. An in-order dependency type is known to be applicable in Online DDL's `vitess` strategy.

The implementation relies heavily on the various `Validate()` functionalities. The order of diffs begins with a heuristic order, which is likely to be the final order in most cases; but it may permute that order in between dependent diffs, if the heuristic order is not linearly valid. 

Note: `VIEW` column dependencies support is provided in unmerged https://github.com/vitessio/vitess/pull/12147 ; https://github.com/vitessio/vitess/pull/12147 works, but we're looking to refactor it and reuse an internal implementation. The tests are prepared and commented out.

## Related Issue(s)

- https://github.com/vitessio/vitess/pull/12113
- Supersedes https://github.com/vitessio/vitess/pull/12028
- WIP: https://github.com/vitessio/vitess/pull/12147
- tracking: https://github.com/vitessio/vitess/issues/10203

## Checklist


shlomi-noach:
Converting to Draft while adapting to merger of https://github.com/vitessio/vitess/pull/12565

shlomi-noach:
Good merger with #12565, and the tests we were waiting on exposed a bug, now fixed.

shlomi-noach:
Ping for review :pray: @dbussink in particular

go/mysql/collations/mysqldata.go
```diff
@@ -119,7 +119,7 @@ var tounicode_dec8_swedish_ci = [...]uint16{
...
...x61, 0x62, 0x63, 0x64, 0x65, 0x66, 0x67, 0x68, 0x69, 0x6a, 0x6b, 0x6c, 0x6d, 0x6e, 0x6f, 0x70, 0x71, 0x72, 0x73, 0x74, 0x75, 0x76, 0x77, 0x78, 0x79, 0x7a, 0x7b, 0x7c, 0x7d, 0x7e, 0x7f, 0x80, 0x81, 0x82, 0x83, 0x84, 0x85, 0x86, 0x87, 0x88, 0x89, 0x8a, 0x8b, 0x8c, 0x8d, 0x8e, 0x8f, 0x90, 0x91, 0x92, 0x93, 0x94, 0x95, 0x96, 0x97, 0x98, 0x99, 0x9a, 0x9b, 0x9c, 0x9d, 0x9e, 0x9f, 0xa0, 0xa1, 0xa2, 0xa3, 0xa8, 0xa5, 0x0, 0xa7, 0x0, 0xa9, 0xaa, 0xab, 0x0, 0x0, 0x0, 0x0, 0xb0, 0xb1, 0xb2, 0xb3, 0x0, 0xb5, 0xb6, 0xb7, 0x0, 0xb9, 0xba, 0xbb, 0xbc, 0xbd, 0x0, 0xbf, 0xc0, 0xc1, 0xc2, 0xc3, 0xc4, 0xc5, 0xc6, 0xc7, 0xc8, 0xc9, 0xca, 0xcb, 0xcc, 0xcd, 0xce, 0xcf, 0x0, 0xd1, 0xd2, 0xd3, 0xd4, 0xd5, 0xd6, 0x0, 0xd8, 0xd9, 0xda, 0xdb, 0xdc, 0x0, 0x0, 0xdf, 0xe0, 0xe1, 0xe2, 0xe3, 0xe4, 0xe5, 0xe6, 0xe7, 0xe8, 0xe9, 0xea, 0xeb, 0xec, 0xed, 0xee, 0xef, 0x0, 0xf1, 0xf2, 0xf3, 0xf4, 0xf5, 0xf6, 0x0, 0xf8, 0xf9, 0xfa, 0xfb, 0xfc, 0x0, 0x0, 0xfd}}, eightbit.UnicodeMapping{From: 0x152, To: 0x178, Range: []uint8{0xd7, 0xf7, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0xdd}}}
```
shlomi-noach, line 122: The diff here is due to `gofmt -s -w ...`. the file is not formatted correctly in `main`.

go/mysql/collations/mysqldata.go
shlomi-noach, line 1: The diff here is due to `gofmt -s -w ...`. the file is not formatted correctly in `main`.

go/vt/schemadiff/schema_diff.go
```diff
@@ -0,0 +1,247 @@
...
+		depType: depType,
+	}
+}
+
+func (d *DiffDep) hashKey() string {
+	return d.diff.CanonicalStatementString() + "/" + d.depDiff.CanonicalStatementString()
+}
+
+/*
+The blow is adapted from https://yourbasic.org/golang/generate-permutation-slice-string/
```
mattlord, line 55: `The below`, I'm assuming.

go/vt/schemadiff/schema_diff.go
```diff
@@ -0,0 +1,247 @@
...
+	depType DiffDepType
+}
+
+func NewDiffDep(diff EntityDiff, depDiff EntityDiff, depType DiffDepType) *DiffDep {
+	return &DiffDep{
+		diff:    diff,
+		depDiff: depDiff,
+		depType: depType,
+	}
+}
```
mattlord, line 48: This isn't concurrency-safe, right? That seems fine but IMO we should comment on that.

go/vt/schemadiff/schema_diff_test.go
```diff
@@ -0,0 +1,657 @@
...
+			"create table t2 (id int primary key, ts timestamp);",
+			"create view v1 as select id from t1",
+		}
+		toQueries = []string{
+			"create table t1 (id int primary key, info int not null, i int);",
+			"create table t2 (id int primary key);",
+			"create view v1 as select id, info from t1",
+			"create view v2 as select id from t2",
+		}
+	)
```
mattlord, line 40: I feel like we could probably benefit from additional test cases? Maybe not though. Your call.

go/vt/schemadiff/schema_diff_test.go
```diff
@@ -0,0 +1,657 @@
...
+		res := ""
+		for _, diff := range diffs {
+			res = res + diff.CanonicalStatementString() + ";"
+		}
+		return res
+	}
+	{
+		iteration := 0
+		allPerms := map[string]bool{}
+		allDiffs := schemaDiff.allDiffs()[:]
```
mattlord, line 67: Curious why you use `[:]` here?

go/vt/schemadiff/schema_diff_test.go
```diff
@@ -0,0 +1,657 @@
...
+			name: "add two fulltext keys",
+			toQueries: []string{
+				"create table t1 (id int primary key, info int not null);",
+				"create table t2 (id int primary key, ts timestamp, v varchar, fulltext key ftk1 (v), fulltext key ftk2 (v));",
+				"create view v1 as select id from t1",
+			},
+			expectDiffs: 2,
+			expectDeps:  1,
+			sequential:  true,
+			entityOrder: []string{"t2", "t2"},
```
mattlord, line 168: I remember why from our previous discussions, but probably worth a comment why we expect this.

go/vt/schemadiff/schema_diff_test.go
```diff
@@ -0,0 +1,657 @@
...
+				assert.Equalf(t, len(tc.entityOrder), len(orderedDiffs), "expected %d diffs/entities per %v", len(tc.entityOrder), tc.entityOrder)
+				diffEntities := []string{}
+				for _, diff := range orderedDiffs {
+					diffEntities = append(diffEntities, diff.EntityName())
+				}
+				assert.Equalf(t, tc.entityOrder, diffEntities, "diffs: %v", strings.Join(diffStatementStrings, ";\n"))
+			}
+			for _, diff := range orderedDiffs {
+				s := diff.CanonicalStatementString()
+				// Internal sanity, while we're here: see that the equivalence relation has entries for all diffs.
```
mattlord, line 651: FWIW, I think as part of the inclusive language effort we're supposed to avoid sanity-check related language.

go/vt/schemadiff/table_test.go
```diff
@@ -390,6 +390,13 @@ func TestCreateTableDiff(t *testing.T) {
 			from: "create table t1 (`id` int primary key, i int)",
 			to:   "create table t2 (`id` int, i int, primary key (id))",
 		},
+		{
+			name:  "change primary key column",
+			from:  "create table t1 (id int primary key, info int not null);",
+			to:    "create table t1 (the_id int primary key, info int not null);",
+			diff:  "alter table t1 drop primary key, drop column id, add column the_id int first, add primary key (the_id)",
+			cdiff: "ALTER TABLE `t1` DROP PRIMARY KEY, DROP COLUMN `id`, ADD COLUMN `the_id` int FIRST, ADD PRIMARY KEY (`the_id`)",
```
mattlord, line 398: I would have expected to rename the column. I guess not? I imagine you have valid reasons.

mattlord (approved):
LGTM! I only had minor nits/comments/questions.

Thank you for all of the unit tests! ❤️

go/mathutil/equivalence_relation.go
```diff
@@ -0,0 +1,144 @@
...
+	return class1 == class2, nil
+}
+
+// OrderedClasses returns the list of classes, increasing
+func (r *EquivalenceRelation) OrderedClasses() []int {
+	classes := []int{}
+	for class := range r.classElementsMap {
+		classes = append(classes, class)
+	}
+	sort.Ints(classes)
```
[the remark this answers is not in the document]
shlomi-noach: I don't think it will, because I don't have a slice in the first place; I have a `map`, and I want to return a slice of the sorted map's keys. So I need to do the hard work of turning that into a slice, and from there, `Ints()` is the trivial sorting mechanism.

go/vt/schemadiff/table_test.go
shlomi-noach: Clarified in code why this is actually the expected behavior (there's a _hint_ switch to control what behavior you want)

go/vt/schemadiff/schema_diff_test.go
shlomi-noach: Changed to `integrity`

go/vt/schemadiff/schema_diff_test.go
shlomi-noach: Clarified in code comment.

go/vt/schemadiff/schema_diff_test.go
shlomi-noach: I'm curious too! Removed.

go/vt/schemadiff/schema_diff_test.go
shlomi-noach: Added multiple tests.

go/vt/schemadiff/schema_diff.go
shlomi-noach: Correct! Clarified in code comment on `SchemaDiff` struct.

go/vt/schemadiff/schema_diff.go
shlomi-noach: Fixed

go/vt/schemadiff/schema.go
```diff
@@ -763,15 +761,130 @@ func (s *Schema) apply(diffs []EntityDiff) error {
...
+	schemaDiff := NewSchemaDiff(s)
+	schemaDiff.loadDiffs(diffs)
+
+	// Utility function to see whether the given diff has dependencies on diffs that operate on any of the given named entities,
+	// and if so, record that dependency
+	checkDependencies := func(diff EntityDiff, dependentNames []string) (dependentDiffs []EntityDiff, relationsMade bool) {
+		for _, dependentName := range dependentNames {
+			dependentDiffs = schemaDiff.diffsByEntityName(dependentName)
+			for _, dependentDiff := range dependentDiffs {
+				// so, 'diff' is an entity that has changed. But now we also see that one of the
```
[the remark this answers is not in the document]
shlomi-noach: Rewrote and fixed grammar.

go/mathutil/equivalence_relation.go
```diff
@@ -0,0 +1,144 @@
...
+
+// Add adds a single element to the set. The element is associated with its own unique class
+func (r *EquivalenceRelation) Add(element string) {
+	if _, ok := r.elementClassMap[element]; ok {
+		// element already exists
+		return
+	}
+	r.elementClassMap[element] = r.classCounter
+	r.classElementsMap[r.classCounter] = []string{element}
+	r.classCounter++
```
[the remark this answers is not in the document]
shlomi-noach: Right. As the equivalence relation goes, it's only important that any new element gets its own distinct class-number. So `r.classCounter++` ensures the next element is added onto its own distinct class.

The users of the equivalence relation, though, are interested in the order, just so that we have kind of a stable sorting mechanism.

go/vt/schemadiff/schema.go
```diff
@@ -457,7 +451,11 @@ func (s *Schema) Diff(other *Schema, hints *DiffHints) (diffs []EntityDiff, err
 	for _, e := range s.Entities() {
 		if _, ok := other.named[e.Name()]; !ok {
 			// other schema does not have the entity
-			dropDiffs = append(dropDiffs, e.Drop())
+			// Entities are sorted in foreign key CREATE TABLE valid order (create parents first ,then children).
```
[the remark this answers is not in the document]

go/mysql/collations/mysqldata.go
```diff
@@ -119,7 +119,7 @@ var tounicode_dec8_swedish_ci = [...]uint16{
 	0x00f8, 0x00f9, 0x00fa, 0x00fb, 0x00fc, 0x00ff, 0x0000, 0x0000,
```
dbussink, line 119: @shlomi-noach Any reason why the collations data was updated in this PR?

go/vt/schemadiff/schema.go
```diff
@@ -763,15 +761,130 @@ func (s *Schema) apply(diffs []EntityDiff) error {
...
 	if err := dup.apply(diffs); err != nil {
 		return nil, err
 	}
 	return dup, nil
 }
 
+// SchemaDiff calulates a rich diff between this schema and the given schema. It is stronger than Diff().
+// On top of returning the list of diffs that can take this schema into the given schema, this function also
+// evaluates the dependencies between those diffs, if any.
+func (s *Schema) SchemaDiff(other *Schema, hints *DiffHints) (*SchemaDiff, error) {
```
dbussink, line 773: Do we want to deprecate / remove `Diff`? I guess it uses it internally, but should we then rename `Diff` to `diff` to not export it?

go/vt/schemadiff/table.go
```diff
@@ -231,6 +246,11 @@ func (d *RenameTableEntityDiff) IsEmpty() bool {
 	return d.Statement() == nil
 }
 
+// EntityName implements EntityDiff
+func (d *RenameTableEntityDiff) EntityName() string {
+	return d.from.Name()
```
dbussink, line 251: Is this deliberately the `from` name? Or should it be the `to` name? Or does it not matter here?

go/mysql/collations/mysqldata.go
shlomi-noach: @dbussink  yes! Per https://github.com/vitessio/vitess/pull/12551#pullrequestreview-1382840070, this is just a `gofmt` issue.

rohit-nayak-ps: I think the only change is to fix the lint warnings, removing the redundant `eightbit.UnicodeMapping{}` type.

go/vt/schemadiff/schema.go
shlomi-noach: Yes, I intend to do that, but in a separate PR, to have a grace period of a couple weeks where we can use either.

go/mysql/collations/mysqldata.go
shlomi-noach: @dbussink  yes! Per https://github.com/vitessio/vitess/pull/12551#pullrequestreview-1382840070, this is just a `gofmt` issue.

go/vt/schemadiff/table.go
shlomi-noach: While this isn't strictly in-use right now, it is deliberately `from` name. This is because the diff operates on the `from` table -- so that table is the one being affected by the operation. This is in line with the rest of diff types (with the exception of `CREATE` where the affected table is `to`).

PR #12560, opened 2023-03-06
Examples, Flakes: Wait for Shard's VReplication Engine to Open

mattlord:
## Description

The [local examples CI workflow](https://github.com/vitessio/vitess/blob/main/.github/workflows/local_example.yml) is still flaky — [failing ~ 35% of the time](https://github.com/vitessio/vitess/actions/workflows/local_example.yml) — and it always fails here (or in the same way in later VReplication steps):
```
+ for shard in "customer/0"
+ true
+ command mysql --no-defaults -h 127.0.0.1 -P 15306 customer/0 -e 'show tables'
+ break
+ ./202_move_tables.sh
E0306 18:54:21.968491   13432 main.go:96] E0306 18:54:21.967743 traffic_switcher.go:208] buildTrafficSwitcher failed: rpc error: code = Unknown desc = TabletManager.VReplicationExec on zone1-0000000200 error: vreplication engine is closed: vreplication engine is closed
E0306 18:54:21.969504   13432 main.go:105] remote error: rpc error: code = Unknown desc = TabletManager.VReplicationExec on zone1-0000000200 error: vreplication engine is closed: vreplication engine is closed
MoveTables Error: rpc error: code = Unknown desc = TabletManager.VReplicationExec on zone1-0000000200 error: vreplication engine is closed: vreplication engine is closed
```
You can see an example here (see any of the previous 4 failed runs): https://github.com/vitessio/vitess/actions/runs/4346133011

The CI test [specifically waits for the shard to be able to successfully serve a query through the vtgate](https://github.com/vitessio/vitess/blob/main/test/local_example.sh#L37-L49), but for some reason that doesn't always work as expected in the CI; I have not been able to repeat it locally — so perhaps a bash version difference or some other missing context variable OR perhaps the vtgate _is_ able to serve queries but the VReplication engine takes much longer to open (GitHub runners being very slow at times is a common factor in our tests). Given that users can also run into this when running the steps as a sequence: `./101...; ./201...; ./202...; ./203...` I added a VReplication engine check to the `wait_for_healthy_shard` library function.

With the changes in this PR, the workflow ran successfully 10 times in a row in the CI with each workflow run containing 3 subtests: 1 each for consul, etcd, and k8s topos, so the base test really passed 30 times in a row: https://github.com/vitessio/vitess/actions/runs/4347322200

## Checklist

### DIFF
--- a/examples/common/lib/utils.sh
+++ b/examples/common/lib/utils.sh
@@ -24,10 +24,10 @@ function wait_for_shard_tablets() {
 	if [[ -z ${1} || -z ${2} || -z ${3} ]]; then
 		fail "A keyspace, shard, and number of tablets must be specified when waiting for tablets to come up"
 	fi
-	keyspace=${1}
-	shard=${2}
-	num_tablets=${3}
-	wait_secs=180
+	local keyspace=${1}
+	local shard=${2}
+	local num_tablets=${3}
+	local wait_secs=180
 
 	for _ in $(seq 1 ${wait_secs}); do
 		cur_tablets=$(vtctldclient GetTablets --keyspace "${keyspace}" --shard "${shard}" | wc -l)
@@ -50,10 +50,10 @@ function wait_for_healthy_shard_primary() {
 	if [[ -z ${1} || -z ${2} ]]; then
 		fail "A keyspace and shard must be specified when waiting for the shard's primary to be healthy"
 	fi
-	keyspace=${1}
-	shard=${2}
-	unhealthy_indicator='"primary_alias": null'
-	wait_secs=180
+	local keyspace=${1}
+	local shard=${2}
+	local unhealthy_indicator='"primary_alias": null'
+	local wait_secs=180
 
 	for _ in $(seq 1 ${wait_secs}); do
                 if ! vtctldclient --server=localhost:15999 GetShard "${keyspace}/${shard}" | grep -qi "${unhealthy_indicator}"; then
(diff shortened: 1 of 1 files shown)


examples/common/lib/utils.sh
```diff
@@ -67,21 +67,49 @@ function wait_for_healthy_shard_primary() {
...
+                fail "A keyspace and shard must be specified when waiting for the shard primary tablet's VReplication engine to open"
+        fi
+        local keyspace=${1}
+        local shard=${2}
+        local wait_secs=90
+
+        for _ in $(seq 1 ${wait_secs}); do
+                if vtctlclient --server=localhost:15999 Workflow -- "${keyspace}" listall &>/dev/null; then
+                        break
+                fi
```
frouioui, line 87: `listall` here exits with `1` or another value than `0` if the list is not valid?

frouioui (approved):

examples/common/lib/utils.sh
mattlord: I'm not sure what you mean by "the list is not valid", but:
```
$ vtctlclient Workflow -- commerce listall; echo $?
Workflow Error: rpc error: code = Unknown desc = no primary found for shard commerce/0
1

$ vtctlclient Workflow -- commerce listall; echo $?
Workflow Error: rpc error: code = Unknown desc = TabletManager.VReplicationExec on zone1-0000000100 error: vreplication engine is closed: vreplication engine is closed
1

$ vtctlclient Workflow -- foobar listall; echo $?
Workflow Error: rpc error: code = Unknown desc = node doesn't exist: /vitess/global/keyspaces/foobar/shards/
1

$ vtctlclient Workflow -- commerce listall; echo $?
No workflows found in keyspace commerce
0
```

rohit-nayak-ps (approved):
lgtm

PR #12565, opened 2023-03-07
schemadiff: validate views' referenced columns via semantics

shlomi-noach:
An altrnative to https://github.com/vitessio/vitess/pull/12147, that uses `go/vt/vtgate/semantics` to validate view column references.

## Description

Up till now, `schemadiff` views validation and analysis was limited to:

- Obviously validate statement
- Infer tables (or views) reverenced by a view
- Validate those tables/views exist
- Validate no dependency loop

As of this PR, `schemadiff` also validates columns used and returned by views:

- Validate that columns in `SELECT` and in `WHERE` clause do in fact exist in tables/views referenced by the view. Error when an unknown column is encountered.
- Allow unqualified table name where there is no ambiguity. Error on unqualified name with ambiguity.
- Analyze view's column names, whether aliased or not.
- Support cascaded views.
- Support star expression, either qualified (`create view v as select t.* from t, t2 ...`) or unqualified (`create view v as select * from t1,t2`) ; infer list of columns ; support cascaded views based on star expression views.

Plenty tests added to validate behavior.

Joint with @dbussink and based on his preliminary work.

## Related Issue(s)

- Tracking issue: https://github.com/vitessio/vitess/issues/10203

## Checklist


shlomi-noach:
Here's how this PR compares with #12147:

https://github.com/planetscale/vitess/compare/planetscale:schemadiff-validate-view-columns-deps...planetscale:schemadiff-validate-view-columns-semantics

go/vt/schemadiff/schema.go
```diff
@@ -761,3 +769,151 @@ func (s *Schema) Apply(diffs []EntityDiff) (*Schema, error) {
...
+	}
+	return columnNames
+}
+
+// getViewColumnNames returns the names of aliased columns returned by a given view.
+func (s *Schema) getViewColumnNames(v *CreateViewEntity, schemaInformation *declarativeSchemaInformation) (
+	columnNames []*sqlparser.IdentifierCI,
+	err error,
+) {
+	err = sqlparser.Walk(func(node sqlparser.SQLNode) (kontinue bool, err error) {
```
systay, line 879: I don't think you want to `Walk` here, since it will traverse down into subqueries and derived tables and capture those columns as well.

You can just iterate over the `v.Select.GetColumns()` slice and achieve the same result.

go/vt/schemadiff/schema.go
```diff
@@ -761,3 +769,151 @@ func (s *Schema) Apply(diffs []EntityDiff) (*Schema, error) {
...
+		switch node := node.(type) {
+		case *sqlparser.StarExpr:
+			if tableName := node.TableName.Name.String(); tableName != "" {
+				for _, col := range schemaInformation.Tables[tableName].Columns {
+					name := sqlparser.CloneRefOfIdentifierCI(&col.Name)
+					columnNames = append(columnNames, name)
+				}
+			} else {
+				dependentNames, err := getViewDependentTableNames(v.CreateView)
```
systay, line 888: FYI - information about used tables can be had from the semantic table that is the output of `semantics.Analyze`.

It's in the `Tables []TableInfo` on `SemTable`

go/vt/schemadiff/schema.go
```diff
@@ -761,3 +769,151 @@ func (s *Schema) Apply(diffs []EntityDiff) (*Schema, error) {
...
+							columnNames = append(columnNames, name)
+						}
+					}
+				}
+			}
+			if len(columnNames) == 0 {
+				return false, &InvalidStarExprInViewError{View: v.Name()}
+			}
+		case *sqlparser.AliasedExpr:
+			if node.As.String() != "" {
```
systay, line 906: `sqlparser.AliasedExpr` has the `ColumnName()` method that does something very similar to this

go/vt/schemadiff/semantics.go
```diff
@@ -0,0 +1,62 @@
...
+	tbl := &vindexes.Table{
+		Name:                    sqlparser.NewIdentifierCS(tableName),
+		Columns:                 []vindexes.Column{},
+		ColumnListAuthoritative: true,
+		Keyspace:                semanticKS,
+	}
+	si.Tables[tableName] = tbl
+}
+
+// addColumn adds a fake column with n otype. It assumes the table already exists
```
systay, line 56: typo nitpick:  `with n otype` -> `with no type`

go/vt/schemadiff/schema.go
```diff
@@ -761,3 +769,151 @@ func (s *Schema) Apply(diffs []EntityDiff) (*Schema, error) {
...
+		return nil, &EntityNotFoundError{Name: entityName}
+	}
+	// The entity is either a table or a view
+	switch entity := entity.(type) {
+	case *CreateTableEntity:
+		return s.getTableColumnNames(entity), nil
+	case *CreateViewEntity:
+		return s.getViewColumnNames(entity, schemaInformation)
+	}
+	return nil, &EntityNotFoundError{Name: entityName}
```
systay, line 863: nitpick: this is not so much an `EntityNotFoundError` and more of a `WrongEntityTypeError`

shlomi-noach: Agreed. It's really an unexpected error, and we should never get there, because today, and for the foreseeable future, we only deal with `Table` and `View` entities. Converted to
```go
vterrors.Errorf(vtrpc.Code_INTERNAL, "unexpected entity type for %v", entityName)
```

go/vt/schemadiff/semantics.go
shlomi-noach: fixed

go/vt/schemadiff/schema.go
shlomi-noach: but, this is chicken and egg: I'm running the above in order to _build_ the `FakeSI` to feed `semantics`...

go/vt/schemadiff/schema.go
shlomi-noach: done, working!

go/vt/schemadiff/schema.go
shlomi-noach: done. This works!

shlomi-noach:
@systay I've addressed all comments (4 implemented as suggested, 1 feedback).

systay (approved):

shlomi-noach:
are any more reviews required from @vitessio/query-serving ?

go/vt/schemadiff/errors.go
```diff
@@ -334,3 +334,39 @@ type ViewDependencyUnresolvedError struct {
...
+func (e *InvalidColumnReferencedInViewError) Error() string {
+	switch {
+	case e.Column == "":
+		return fmt.Sprintf("view %s references non-existent table %s", sqlescape.EscapeID(e.View), sqlescape.EscapeID(e.Table))
+	case e.Table != "":
+		return fmt.Sprintf("view %s references non-existent column %s.%s", sqlescape.EscapeID(e.View), sqlescape.EscapeID(e.Table), sqlescape.EscapeID(e.Column))
+	case e.Ambiguous:
+		return fmt.Sprintf("view %s references unqualified but non unique column %s", sqlescape.EscapeID(e.View), sqlescape.EscapeID(e.Column))
+	default:
+		return fmt.Sprintf("view %s references unqualified but non-existent column %s", sqlescape.EscapeID(e.View), sqlescape.EscapeID(e.Column))
```
harshit-gangal, line 354: the error message looks very fragile based on what struct fields are set.

harshit-gangal: can we also create the message ahead of time, when the struct is created?

dbussink: I think that’s actually the wrong pattern. The core is the structured error and the string format is just a detail here. 

The structured error removes the issue of string being brittle because anyone wants more details doesn’t look at the string at all.

shlomi-noach: > the error message looks very fragile based on what struct fields are set.

I can rewrite that to be more clear

shlomi-noach: I was able to simplify `InvalidColumnReferencedInViewError` by removing unused `Table` columns and subsequent error message generation cases.

dbussink: @harshit-gangal allocating the string is also wasteful since when using structured errors it is never used. So I don’t think we should do that.

shlomi-noach:
PR seems to suffer a cyclic import path.

go/vt/schemadiff/semantics.go
```diff
@@ -0,0 +1,62 @@
...
+// to make it more simple and accessible to schemadiff's logic.
+type declarativeSchemaInformation struct {
+	semantics.FakeSI
+}
+
+func newDeclarativeSchemaInformation() *declarativeSchemaInformation {
+	return &declarativeSchemaInformation{
+		semantics.FakeSI{
+			Tables: make(map[string]*vindexes.Table),
+		},
```
harshit-gangal, line 41: FakeSI is for testing, better to take the field from there and create your own full struct.

harshit-gangal (approved):
2 comments. Overall looks good.

go/vt/schemadiff/errors.go
harshit-gangal: At the MySQL protocol level, error strings are transferred over the wire, so we might be generating them eventually.

dbussink: @harshit-gangal this is never used for MySQL level errors. But even then, we’d allocate when needed and not always so then it’s imho still an improvement. 

It also improved the ergonomics imho since it’s now easier to create a structured error then.

go/vt/schemadiff/semantics.go
shlomi-noach: fixed

go/vt/schemadiff/schema.go
```diff
@@ -761,3 +770,147 @@ func (s *Schema) Apply(diffs []EntityDiff) (*Schema, error) {
...
+			}
+			switch semErr.Code {
+			case semantics.AmbiguousColumn:
+				if colName := extractColName(semErr.Args()[0]); colName != "" {
+					return &InvalidColumnReferencedInViewError{
+						View:      view.Name(),
+						Column:    colName,
+						Ambiguous: true,
+					}
+				}
```
frouioui, line 828: nit: should we extract this if into a method that takes `colName`, `view.Name()`, `Ambiguous (true or false)`, and `semErr`? We're using almost the same snippet in the `case` below.

frouioui (approved):

shlomi-noach:
Depends on https://github.com/vitessio/vitess/pull/12663 to solve import loop

shlomi-noach:
Merged `resolve-semantics-engine-deps`, introduced [here](https://github.com/vitessio/vitess/pull/12663).

shlomi-noach:
only merge this after https://github.com/vitessio/vitess/pull/12663 is merged.

Issue #12567, opened 2023-03-07
Feature Request: improvements to logrotate


maxenglander:
## Description

There are a few different feature requests, and I thought it might be a bit annoying to give each of these their own GH issue, so putting them all in one for now. Happy to split them out though. I think all of these ideas can be introduced in a way that is backwards compatible for folks who rely on the current behavior of the container.

 - [ ] make gid/uid configurable
 - [ ] add `mysql` binary to the container
 - [ ] make `vtdataroot` is configurable
 - [ ] make log paths, sizes, and rotate params configurable
 - [ ] make standard mysql logs discoverable

## make gid/uid configurable

**problem**

Right now the Dockerfile is creating a `vitess` user and group of 1000 and 2000 respectively. https://github.com/vitessio/vitess/blob/97b151e4b8f96547efb642531c0953de86f065ca/docker/k8s/logrotate/Dockerfile#L30-L31

Meanwhile the `lite` images do not specify a UID/GID, and they end having a UID/GID of 999, at least for me on my Mac M1:

```
$ docker run -it --entrypoint=/bin/bash vitess/lite:latest
vitess@508dc6ec8fbd:/$ cat /etc/passwd | grep vitess
vitess:x:999:999::/home/vitess:/bin/sh
```

The discrepancy in these values makes it difficult to run `docker/lite` and `logrotate` together in a K8s Pod. If we try to set the security context to run the `logrotate` container as user, it will be able to create `logrotate.status`, but won't be able to rotate files owned by user 999. If we set the security context to run as user 999, problem switches the other way.

**recommendation**

Not sure what the intent is with having the `logrotate` UID/GID be 1000/2000. Rather than change this and potentially break things for people who rely on that currently, we can introduce build args `GID` and `UID`. Or we could just break things for people and make these the same as what's in `vitess/lite`.

## add mysql client binary to the container

The current `logrotate.conf` file uses `/usr/bin/mysql` to do some postrotation stuff. But `/usr/bin/mysql` isn't on the container! Not sure how other people are using this currently, but I think it would be helpful if we added the client binary to the container.

https://github.com/vitessio/vitess/blob/97b151e4b8f96547efb642531c0953de86f065ca/docker/k8s/logrotate/logrotate.conf#L1

## make `vtdataroot` is configurable

[...]

PR #12568, opened 2023-03-07
docker/k8s/logrotate: add flexibility and support non-wildcard logs

maxenglander:
## Description

Proof-of-concept PR to show how to make `logrotate` container more flexible. These changes aim to be backwards compatible with the existing behavior of the container.

 * Makes gid/uid configurable so the container can be integrated with `vitess/lite` containers and not conflict over file ownership.
 * Adds mysql client binary to the container so that `postrotate` works out-of-the-box.
 * Adds support for configuring the wildcard log path, `size` and `rotate` params.
 * Adds support for individually configuring the path, `size` and `rotate` params of audit, error, general and slow logs.
 * Adds support for disabling `set -e`, so that if logrotate exits for some reason, it doesn't restart the container. Maybe some people want the container to restart, others might not.
 * Adds support for dynamically discovering the audit, error, general and slow log paths via MySQL global vars.
 * Adds log-specific postrotate scripts for audit, error, general and slow log paths.
 * Make the sleep interval configurable.

This is a lot of bash! If we don't like that can rewrite this in Golang.

## Related Issue(s)

https://github.com/vitessio/vitess/issues/12567

## Checklist


derekperkins:
> so that if logrotate exits for some reason, it doesn't kill the pod it lives in

To clarify, that would just restart the logrotate container, not the entire pod, but the general point is still valid

derekperkins:
This is a lot of bash... Is there a reason to handle flag parsing instead of just using env vars? If we could eliminate that, I think that would reduce a significant amount of code

maxenglander:
> To clarify, that would just restart the logrotate container

🤦 One day I'll be able to keep straight how K8s works. You're right, probably nobody is going to want to put a readiness probe on the logrotate container.

>  Is there a reason to handle flag parsing instead of just using env vars?

No real reason other than force of habit, I agree with this suggestion will make the change.

derekperkins:
While I hacked together the last very minimal logrotate, I don't know that I'm the best person to review this. I'm not the best at bash, and I know very little about logrotate. It appears that the defaults here don't break any historical usage, but I haven't tested it. If you pushed a test image somewhere, I'd be happy to try it out on our canary keyspace

maxenglander:
Hey @derekperkins thanks for the offer to test out the image. I've built an image like this:

```
$ git --no-pager log -1
commit 676b2e02a3cb38f4ec7711ac80ed090abf36e51f (HEAD -> maxeng-flexible-logrotate, planetscale/maxeng-flexible-logrotate)
Author: Max Englander <max@planetscale.com>
Date:   Mon Mar 20 13:27:13 2023 -0400

    remove cli args just use env vars
    
    Signed-off-by: Max Englander <max@planetscale.com>
$ pwd                  
/Users/maxenglander/dev/vitessio/vitess/docker/k8s/logrotate
$ docker buildx build --platform linux/amd64 --push -t maxenglander/logrotate:676b2e02a3cb38f4ec7711ac80ed090abf36e51f .
[+] Building 0.0s (0/0)                        
```

The image is public [here in Dockerhub](https://hub.docker.com/layers/maxenglander/logrotate/676b2e02a3cb38f4ec7711ac80ed090abf36e51f/images/sha256-6f6429df2ccbb1685e898c9a4c9409e2b01437fd19845b5d32a085bc442ade9a?context=repo). Let me know if you can't access it or it's not the right architecture for your env.

I've tested a different internal image which was built identically except for UID/GID changes:

```
docker buildx build --platform linux/amd64 --build-arg UID=999 --build-arg GID=999 -t [redacted]/logrotate:676b2e02a3cb38f4ec7711ac80ed090abf36e51f .
```

Separately, I'll ask around for another maintainer with some bash/logrotate expertise to also take a look here.

Issue #12570, opened 2023-03-08
RFC: Support Automatic Read Write Seperation


earayu:
# Background
Currently, Vitess allows users to specify the tablet type to access using the `use` command, such as `use dbName@PRIMARY`, `use dbName@REPLICA`. However, this requires users to modify their program to decide which type of tablet to use. If Vitess could make this decision automatically, it would make Vitess easier to use and better balance the load on different MySQL nodes.

# Feature
A parameter `read_write_separation_strategy=enable/disable` is provided, and when users do not specifically specify which type of tablet to access, VTGate will automatically route read requests to REPLICA or RDONLY based on this parameter.

In addition, a new Hint has been added, which allows users to route a SQL statement to a specific type of tablet:
```
/*vt+ NODE_TYPE=PRIMARY */
/*vt+ NODE_TYPE=REPLICA */
/*vt+ NODE_TYPE=RDONLY */
```

# Solution
By parsing the user's SQL, we can determine whether it is a simple query statement, a DML statement, or a special statement that needs to be executed on the PRIMARY node, such as `SELECT LAST_INSERT_ID()`.
If it is a simple query statement, we can set `defaultTabletType=REPLICA/RDONLY` before execution.
There are also other factors to consider, such as whether a transaction is used or whether a temporary table is used. These will be explained below.

# Usage
VTGate has added a new startup parameter: `read_write_separation_strategy=enable/disable`. Its function and usage are similar to `ddl_strategy`, with the default value being `disable`. Users can also set its value using the set command, for example: `SET @@read_write_separation_strategy=enable`. Similar to `ddl_strategy`, it will store the variable value in the session and affect subsequent SQL statements in the same connection.
As for the Hint, it will only affect the current SQL statement being executed.

# Forwarding Rules

- [ ]  Forward to read-only instances or primary instance（Automatic）
	- [x]  SELECT statements outside of transactions.
	- [x]  COM_STMT_EXECUTE command.
- [ ]  Always forward to primary instance
	- [x]  INSERT, UPDATE, DELETE, SELECT FOR UPDATE.
	- [x]  All DDL operations (create/drop tables/databases, alter table structures, permissions, etc.).
	- [x]  Requests in all transactions.
	- [x]  Requests that use temporary tables.
	- [x]  Requests that use GET_LOCK/RELEASE_LOCK/IS_USED_LOCK/RELEASE_ALL_LOCKS/IS_FREE_LOCK.
	- [x]  SELECT last_insert_id().
[...]

PR #12571, opened 2023-03-08
support read write seperation

earayu:
## Description
Currently, Vitess allows users to specify the tablet type to access using the use command, such as use dbName@PRIMARY, use dbName@REPLICA. However, this requires users to modify their program to decide which type of tablet to use. If Vitess could make this decision automatically, it would make Vitess easier to use and better balance the load on different MySQL nodes.

## Related Issue(s)

[* RFC：https://github.com/vitessio/vitess/issues/12570](https://github.com/vitessio/vitess/issues/12570)

## Checklist


deepthi:
Hello! You seem to have put a significant amount of time into this. For future reference, we (maintainers) prefer to discuss the proposal first, especially when it is something radical like this, before a PR is created in order to avoid wasted time if we decide not to accept the proposal.

earayu:
> Hello! You seem to have put a significant amount of time into this. For future reference, we (maintainers) prefer to discuss the proposal first, especially when it is something radical like this, before a PR is created in order to avoid wasted time if we decide not to accept the proposal.

Sure, you are right. 
I believe 'read write seperation' is a requirement that many needs, this PR could be a Proof of Concept. I am new to the Vitess commuity, so I am not sure what's the best way to propose a new feature. But I've attached a RFC for this feature: https://github.com/vitessio/vitess/issues/12570 
Maybe we can talk about the feature first

go/vt/sqlparser/analyzer_sql_type.go
```diff
@@ -0,0 +1,134 @@
+/*
+Copyright 2021 The Vitess Authors.
```
harshit-gangal, line 2: 2021 -> 2023

harshit-gangal:
The RFC says comment directive is `NODE_TYPE` but the implementation has `ROLE`. I would like to suggest to use `TABLET_TYPE`.

Another interesting this is that at the table level it is already supported.
what I means is that if you do `select 1 from tbl@replica` this query will go to `REPLICA` tablet.

PR #12574, opened 2023-03-08
Internal refactor: SQLError

shlomi-noach:
## Description

This is an internal refactor and reimplementation of how we generate SQLError.

Background: sometimes we get `SQLError` wrapped, or sent over the wire, or analyzed and recreated as other errors -- and at the end of all this we may still want to be able to treat them as the original `SQLError` that they were. Other times, we want to translate a vitess error into a `SQLError`.

For this, we have the ever so useful `NewSQLErrorFromError`. However, there are anti-patterns to its current implementation:

- It _always_ returns a `SQLError` object, even if the original error has nothing to do with `SQLError`.
  - in which case the error's num and state are `ERUnknownError` and `SSUnknownSQLState`, respectively. This really means "this is not, in fact, a SQL error".
- Function signature returns an `err`, so the caller needs to cast back as `SQLError`.

Example to an anti-pattern usage:

```go
	sqlErr, isSQLErr := mysql.NewSQLErrorFromError(err).(*mysql.SQLError)
	if !isSQLErr {
		return false
	}
	if sqlErr.Num == mysql.ERUnknownError {
		return false
	}
```

again, now annotated:

```go
	sqlErr, isSQLErr := mysql.NewSQLErrorFromError(err).(*mysql.SQLError)
	if !isSQLErr { // This is wasteful, because the result is in fact always a SQLError
		return false
	}
	if sqlErr.Num == mysql.ERUnknownError { // which semantically means: this is not, in fact, a SQLError
		return false
	}
```

This PR addresses the two issues above, as follows:

#### NewSQLErrorFromError(err)

The function `NewSQLErrorFromError()` now returns `(sqlError *SQLError, ok bool)`

- if `ok` is `false`, then the error was not a `SQLError`; it could be `nil`, it should not be used.
- if `ok` is `true``, then we get a formal `SQLError`, no need to cast it. And we know the num and state are actually real.

Most usage now uses this new function and signature. Example for new usage:

```golang
	sqlErr, ok := mysql.NewSQLErrorFromError(err)
	if !ok {
		return false
	}
	switch sqlErr.Num {
	...
        }
```

#### ForceNewSQLErrorFromError(err)

This new function:
```golang
// ForceNewSQLErrorFromError returns a SQLError obejct at all costs. It first attempts to extract
// sensible information from the error, via NewSQLErrorFromError(). But if that's unsuccessful, it
// returns a SQLError with _undefined_ sql num&state.
func ForceNewSQLErrorFromError(err error) *SQLError {
```

Is a backwards-compatible function, for the minority cases of code that absolutely requires the result to be `SQLError`, even if num&state are undefined.

## Related Issue(s)

Story evolved while working on https://github.com/vitessio/vitess/pull/12327

## Checklist


shlomi-noach:
Sadly, this was wishful thinking. The attempt to formalize the result type as `SQLError` breaks due to `golang`'s behavior with `nil` interfaces vs. `nil` implementing values. This leads to `if err != nil` to return `true` even when we return a `nil` value as `SQLError`.

shlomi-noach:
Closing as this refactor does not achieve the desired intention.

Issue #12336, opened 2023-02-14
Bug Report: OnlineDDL VReplication Throttle Via Vstream CI Test Flakiness


mattlord:
### Overview of the Issue

In local runs I see this test fail every so often: https://github.com/vitessio/vitess/blob/d15766ef21dd85615d2db66db7a920e32f036128/go/test/endtoend/onlineddl/vrepl/onlineddl_vrepl_test.go#L391-L422

When it does fail, it always fails here: https://github.com/vitessio/vitess/blob/a6529e1df6ea6e4e7b7ecfa65f985f5f1db0803d/go/test/endtoend/onlineddl/vrepl/onlineddl_vrepl_test.go#L419

With a message like this:
```
--- FAIL: TestSchemaChange/throttled_and_unthrottled_migration_via_vstreamer (9.11s)
Error:      	"2023-02-12 16:50:54" is not greater than or equal to "2023-02-12 16:50:55"
```

It's always one second in the past. 

This can happen when the test runs at the second boundary as we're using second based precision timestamps on the database side. It's the same basic issue that I had seen and addressed in the VDiff2 tests here: https://github.com/vitessio/vitess/pull/11054

### Reproduction Steps

If you run the test locally enough times, you'll see it (it's a timing issue):
```
go test -timeout 30m -v -failfast -run TestSchemaChange vitess.io/vitess/go/test/endtoend/onlineddl/vrepl
```

### Binary Version

```sh
Version: 17.0.0-SNAPSHOT
```

### Operating System and Environment details

```sh
N/A (FWIW I was running it locally on macOS arm64)
```

shlomi-noach:
> This can happen when the test runs at the second boundary as we're using second based precision timestamps on the database side. 

I'm still confused on how this might happen? The migration _first starts_, ie `started_timestamp` is updated to `now()`. Then, it may or may not throttle. It's impossible for throttling to take place before the migration started. So how does the throttling timestamp lower that `started_timestamp`?

shlomi-noach:
`time_throttled` is set in `_vt.vreplication` in:

https://github.com/vitessio/vitess/blob/a6529e1df6ea6e4e7b7ecfa65f985f5f1db0803d/go/vt/vttablet/tabletmanager/vreplication/vreplicator.go#L534-L535

Then, the value is copied over to `_vt.schema_migrations` in:

[...]

PR #12582, opened 2023-03-09
OnlineDDL: Don't Update Migration Started Timestamp on Completion

mattlord:
## Description

We were updating the migration's *started* timestamp on completion. I believe that's what was causing: https://github.com/vitessio/vitess/issues/12336

You can see where it happened 4 times in a row (the workflow retries a second time and the workflow failed 2 times in a row) here: https://github.com/vitessio/vitess/actions/runs/4370884719/jobs/7646225949

That same test is now being run X times in a row here to see if it is still flaky after this change: https://github.com/vitessio/vitess/actions/runs/4371207016

## Related Issue(s)

  - Fixes: https://github.com/vitessio/vitess/issues/12336

## Checklist

### DIFF
--- a/go/vt/vttablet/onlineddl/executor.go
+++ b/go/vt/vttablet/onlineddl/executor.go
@@ -4743,12 +4743,10 @@ func (e *Executor) onSchemaMigrationStatus(ctx context.Context,
 	case schema.OnlineDDLStatusComplete:
 		{
 			progressPct = progressPctFull
-			_ = e.updateMigrationStartedTimestamp(ctx, uuid)
 			err = e.updateMigrationTimestamp(ctx, "completed_timestamp", uuid)
 		}
 	case schema.OnlineDDLStatusFailed:
 		{
-			_ = e.updateMigrationStartedTimestamp(ctx, uuid)
 			err = e.updateMigrationTimestamp(ctx, "completed_timestamp", uuid)
 		}
 	}


shlomi-noach (changes requested):
@mattlord sadly this is not the issue. The function `updateMigrationStartedTimestamp` runs this query:

```sql
UPDATE _vt.schema_migrations SET
			started_timestamp =IFNULL(started_timestamp,  NOW(6)),
			liveness_timestamp=IFNULL(liveness_timestamp, NOW(6))
		WHERE
			migration_uuid=%a
```			

The query only ever updates `started_timestamp` if it's `NULL`, and thereafter keeps it untouched.

mattlord:
> @mattlord sadly this is not the issue. The function `updateMigrationStartedTimestamp` runs this query:
> 
> ```sql
> UPDATE _vt.schema_migrations SET
> 			started_timestamp =IFNULL(started_timestamp,  NOW(6)),
> 			liveness_timestamp=IFNULL(liveness_timestamp, NOW(6))
> 		WHERE
> 			migration_uuid=%a
> ```
> 
> The query only ever updates `started_timestamp` if it's `NULL`, and thereafter keeps it untouched.

Ah, right. Thanks! I can see that it has still failed (and required a retry in 2 of the workflow runs). Closing this PR then.

PR #12583, opened 2023-03-09
CI: extend timeouts in onlineddl_vrepl due to slow CI runners

shlomi-noach:
## Description

We're seeing flakiness in `onlineddl_vrepl`, and this time on very basic tests. Example: failure on `successful_online_alter,_vtctl` in https://github.com/vitessio/vitess/actions/runs/4360811846/jobs/7624093912

This is just a very very basic test: submit a migration via `vtctl`, see that it completes. The test failed because after `20sec`, a migration (one out of two shards) was still running.

These migrations take 4-5 seconds on a local dev box. `20sec` used to be enough in CI. But recently, GitHub's CI runners performance is so bad that this is not enough.

This PR just increases some timeouts.

## Checklist

### DIFF
--- a/go/test/endtoend/onlineddl/vrepl/onlineddl_vrepl_test.go
+++ b/go/test/endtoend/onlineddl/vrepl/onlineddl_vrepl_test.go
@@ -46,8 +46,8 @@ var (
 	onlineDDLThrottlerAppName = "online-ddl"
 	vstreamerThrottlerAppName = "vstreamer"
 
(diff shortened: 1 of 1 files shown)


mattlord (approved):

frouioui:
@shlomi-noach I would probably backport this to our supported release branches, WDYT?

frouioui (approved):

shlomi-noach:
@frouioui backporting to `16`, `15`, 14` !

PR #12584, opened 2023-03-09
CI: increase overall test timeouts for all OnlineDDL tests

shlomi-noach:
## Description

Due to recent GitHub CI runners slowness, we're seeing some tests time out after `20min` of running. These tests normally run for `5-6 min` on a local dev env, and `20min` used to give good margins. Not anymore.

This PR increases all Online DDL related tests timeouts to `30min`.

## Checklist

### DIFF
--- a/test/config.json
+++ b/test/config.json
@@ -252,7 +252,7 @@
 		},
 		"onlineddl_ghost": {
 			"File": "unused.go",
-			"Args": ["vitess.io/vitess/go/test/endtoend/onlineddl/ghost"],
+			"Args": ["vitess.io/vitess/go/test/endtoend/onlineddl/ghost", "-timeout", "30m"],
 			"Command": [],
 			"Manual": false,
 			"Shard": "onlineddl_ghost",
@@ -261,7 +261,7 @@
 		},
 		"onlineddl_vrepl": {
 			"File": "unused.go",
-			"Args": ["vitess.io/vitess/go/test/endtoend/onlineddl/vrepl"],
+			"Args": ["vitess.io/vitess/go/test/endtoend/onlineddl/vrepl", "-timeout", "30m"],
 			"Command": [],
 			"Manual": false,
 			"Shard": "onlineddl_vrepl",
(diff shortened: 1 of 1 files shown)


frouioui:
Same question as in https://github.com/vitessio/vitess/pull/12583#issuecomment-1461406455. I think we should backport this to `release-14.0`, `release-15.0`, and `release-16.0`, WDYT?

frouioui (commented):
I am seeing the different workflows you are changing have a step called `Run cluster endtoend test` which as a timeout of 45 minutes by default:
https://github.com/vitessio/vitess/blob/3ff3c476a0b86172ce1be40d5d3af93a94013acd/.github/workflows/cluster_endtoend_onlineddl_revert.yml#L119

Should we also increase this variable to leave more time for the workflows to complete?

Othewise, LGTM

frouioui (approved):

frouioui:
@shlomi-noach, I looked at it a bit deeper, and `test.go` has a `-timeout` flag that defines the timeout for each test with a default value of 30 minutes. I think it is ok to remove the `-timeout 30m` argument you're adding, and leave that to `test.go`, but I don't have strong opinions.

> These affect all tests -- should we bump the timeout for everything? I don't mind, just making sure.

All the shards you are changing, expect `schemadiff_vrepl`, are composed of only one test, so I think it is okay to leave the CI timeout as it is (45 minutes). For shard `schemadiff_vrepl`, there are two tests: `sidecardb` and `schemadiff_vrepl`, I don't think the sum of the two will go above 45 minutes, but if it does we can change the timeout later.

frouioui:
> Should we also increase this variable to leave more time for the tests to complete?

So I take back what I said, I think let's not increase it :)

shlomi-noach:
@frouioui oh, I think I understand better -- you're saying that `30min` is the default, and that, previously `"-timeout", "20m"` overrode that to get a shorter timeout?

I didn't know that.

At any case, I prefer to keep the `timeout` explicit in `test/confg.json`. This way we don't lose information and intention.

frouioui:
@shlomi-noach, this is what I meant yes. But yes that makes sense to keep the timeout explicit in the `config.json` file, that's why I don't really have strong opinions, I like both ways :)

systay (approved):

dbussink (approved):

Issue #12540, opened 2023-03-02
Bug Report: query result based on cross shard aggregation is bad


eseokoh:
### Overview of the Issue

Sometimes query based on a cross shard aggregation shows a bad result, where not all shards are hit.

### Reproduction Steps

0. Prepare for a keyspace with 2 shards

1. Deploy the following vschema:
```sql
alter vschema on aggt0 add vindex xxhash (shardkey) using xxhash;
```

2. Deploy the following schema:
```sql
 create table aggt0 (
 shardkey bigint primary key,
 a int,
 b int
);

insert into aggt0 (shardkey, a, b) values (0, 100, 10), (10, 200, 20);
```

Please notice that `shardkey` of `0` goes to `80-` shard and that of `10` does to `-80`. 

3. Run Query

```sql
mysql> select A.a, A.b, (A.a / A.b) as d from (select sum(a) as a, sum(b) as b from aggt0) A;
+------+------+---------+
| a    | b    | d       |
+------+------+---------+
|  300 |   30 | 10.0000 |
+------+------+---------+
1 row in set (0.13 sec)
```

This is absolutely good.

However, when the subquery hit not both but one shard, it may show a bad result.
Please repeat the query until you have the bad result.

- the expected result
```sql
mysql> select A.a, A.b, (A.a / A.b) as d from (select sum(a) as a, sum(b) as b from aggt0 where a = 100) A;
+------+------+---------+
| a    | b    | d       |
+------+------+---------+
|  100 |   10 | 10.0000 |
+------+------+---------+
1 row in set (0.00 sec)
```

- the unexpected result
```sql
mysql> select A.a, A.b, (A.a / A.b) as d from (select sum(a) as a, sum(b) as b from aggt0 where a = 100) A;
+------+------+------+
| a    | b    | d    |
+------+------+------+
|  100 |   10 | NULL |
+------+------+------+
1 row in set (0.01 sec)
```

### Binary Version

```sh
14.0.4
```

### Operating System and Environment details

```sh
unrelated
```

### Log Fragments

```sh
N/A
```

PR #12607, opened 2023-03-10
[gen4 planner] Make sure to not push down expressions when not possible

frouioui:
## Description

This PR fixes https://github.com/vitessio/vitess/issues/12540 by making sure to not push down to MySQL complex expressions that involved aggregations.

If we have a derived table and it's aggregating results for us, we can't use these results until the are fully aggregated, which happens on the vtgate level.

```sql
select dt.a,
       dt.b,
       (dt.a / dt.b)
from (select sum(a) as a,
             sum(b) as b
      from user) dt
```

In the query above, since we have to spread out the aggregation across all `user` shards, we can't push down the division between the two sums until after summing up the individual shard results.

## Related Issue(s)

Fixes https://github.com/vitessio/vitess/issues/12540

## Checklist

### DIFF
--- a/go/vt/vtgate/executor_select_test.go
+++ b/go/vt/vtgate/executor_select_test.go
@@ -3887,7 +3887,7 @@ func TestSelectAggregationRandom(t *testing.T) {
 	rs, err := executor.Execute(context.Background(), "TestSelectCFC", session,
 		"select /*vt+ PLANNER=gen4 */ A.a, A.b, (A.a / A.b) as c from (select sum(a) as a, sum(b) as b from user) A", nil)
 	require.NoError(t, err)
-	assert.Equal(t, `[[INT64(10) INT64(1) INT64(10)]]`, fmt.Sprintf("%v", rs.Rows))
+	assert.Equal(t, `[[INT64(10) INT64(1) DECIMAL(10.0000)]]`, fmt.Sprintf("%v", rs.Rows))
 }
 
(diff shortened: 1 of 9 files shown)


systay (changes requested):
I don't this is correct. The query we are trying to fix should not be using the `AggregateRandom` expression here in the first place

frouioui (commented):
LGTM

arthurschreiber (approved):
❤️

go/vt/vtgate/executor_select_test.go
```diff
@@ -3856,6 +3856,40 @@ func TestSelectAggregationData(t *testing.T) {
...
+		sbc.SetResults([]*sqltypes.Result{sqltypes.MakeTestResult(
+			sqltypes.MakeTestFields("a|b|c", "int64|int64|int64"),
+			"null|null|null",
+		)})
+	}
+
+	conns[0].SetResults([]*sqltypes.Result{sqltypes.MakeTestResult(
+		sqltypes.MakeTestFields("a|b|c", "int64|int64|int64"),
+		"10|1|10",
+	)})
```
harshit-gangal, line 3881: the mock result should be changed to return only 2 columns from the shards.
`sum(a)` and `sum(b)`

frouioui: Fixed via e9a203f00f

harshit-gangal (approved):

